PR #25118, opened 2026-08-05
perf(python): bound non-streaming x json inspection work

### LINKED ISSUE #25104: perf(python): bound non-streaming X JSON inspection work
## Problem

Non-streaming X response billing still performs unbounded structural JSON work synchronously on mitmproxy's shared event loop.

At `4c2a7fac17bac1b83689b3f1a31cd8a6f5375cc0`:

- `crates/runner/mitm-addon/src/usage/providers/connectors/x.py:238-244` creates the non-streaming X `JsonSelectiveExtractor` without `max_work_units`.
- `crates/runner/mitm-addon/src/usage/providers/connectors/x.py:452-465` keeps that extractor across all decoded chunks and finalizes it only when the document ends.
- `crates/runner/mitm-addon/src/response_streaming.py:288-307` installs the connector parser, and `crates/runner/mitm-addon/src/response_streaming.py:367-378` feeds it synchronously before the response-stream callback returns.
- `crates/runner/mitm-addon/src/usage/json_selective.py:211-222` already supports a chunk-independent total-work budget, but its default is intentionally unlimited.
- The adjacent X NDJSON parser already uses a `65_536`-unit per-row budget at `crates/runner/mitm-addon/src/usage/providers/connectors/x.py:247-252`.

Identity responses have no response-wide parser-work cap. Gzip and deflate constrain each decoded parser chunk to 64 KiB but permit a 5 MiB expansion grace, so one compressed response callback can still synchronously feed many chunks before returning. The decoder's byte and expansion limits do not bound Python structural dispatch.

## Evidence and impact

Using Python 3.12.3 and the exact production X scalar, array-count, wildcard-count, and object-presence configuration, valid response-shaped JSON containing repeated `data` objects produced:

| JSON bytes | `data` items | Synchronous parse time |
| ---: | ---: | ---: |
| 65,548 | 3,641 | 63.9 ms median |
| 262,144 | 14,563 | 255.3 ms median |
| 1,048,582 | 58,254 | 1.031 s median |
| 5,242,888 | 291,271 | 5.221 s |

The growth is linear but unbounded by document-level work. Parsing happens on the shared event-loop thread, so dense successful X responses can delay unrelated flows.

As a feasibility check, configuring the existing `65_536` work-unit limit:

- allowed the 65,548-byte response to complete;
- rejected 256 KiB, 1 MiB, and 5 MiB dense responses with `work limit exceeded`;
- stopped each larger input after about 130 ms total;
- produced the same result when fed in 64 KiB chunks, confirming that delivery chunking cannot reset the allowance.

The response body remains forwarded by the generic stream wrapper; only billing observations from an over-budget document need to be discarded.

## Existing issue coverage

This is distinct from completed work:

- #24176 bounds one X NDJSON row, not the non-streaming JSON document parser.
- #24890 bounds Anthropic HTTP JSON and SSE inspection.
- #24967 bounds OpenAI Responses HTTP JSON and SSE inspection.
- #21383 accelerates discarded ASCII strings but intentionally does not bound dense structural input.

No open issue or PR covers the non-streaming X JSON ownership boundary.

## Decision

Proceed in one runner-local PR. Apply one private `65_536`-unit total-work allowance at the shared non-streaming X JSON extractor factory. That factory owns both the incremental response document and the direct buffered fallback, so both entry points receive the same per-document policy without changing the generic decoder or parser default.

## Chosen direction

- Keep one extractor and one allowance across every decoded chunk of a non-streaming X response.
- Give each later response flow a fresh extractor and allowance.
- Reuse the same bounded factory in the buffered fallback.
- Preserve complete-result semantics: an exhausted document contributes no partial `data`, `includes`, `meta`, or error counts.
- Preserve unchanged response forwarding and the existing single lost-visibility/parse-error reporting lifecycle.
- Keep X NDJSON on its existing independent fresh-per-row extractor and allowance.
- Keep `JsonSelectiveExtractor` unlimited by default for unrelated consumers.

The `65_536` value matches the existing X NDJSON, Anthropic HTTP, and OpenAI Responses HTTP provider policies and is supported by the X-shaped calibration above. There is no current evidence for a distinct X document value.

## Alternatives considered

- **Decoded-byte cutoff:** does not model the measured structural work, rejects large cheap strings or permits smaller dense documents, and needs additional response-scoped state or decoder coupling.
- **Generic parser default:** would silently change unrelated consumers with different lifetimes and risk profiles.
- **Stop decompression after exhaustion:** adds cross-layer cancellation state while the existing decoder already has separate output bounds; the demonstrated unbounded cost is Python structural dispatch.
- **Buffer and parse later, native full-object parsing, or offload:** may improve typical elapsed time but does not provide a deterministic bound by itself and adds memory, serialization, cancellation, ordering, and lifecycle complexity.

## Risks and mitigations

- A legitimate structurally extreme X response can exceed the allowance and lose response-derived billing visibility. The established allowance, response-shaped success coverage, request-hint fallback where reliable, existing lost-visibility signal, and bulk discarded-string accounting make the trade-off explicit.
- Partial observations could underbill or misbill if published after exhaustion. Tests place selected observations before the exhausting field and require the final state and billing output to contain none of them.
- Chunking or compression could accidentally multiply the allowance if extractors were recreated. Tests cover split identity input and gzip/deflate input expanding into multiple decoded chunks through one response-scoped extractor.
- The direct fallback can receive a large decoded body from a small compressed buffer. It uses the same factory and receives explicit integration coverage.
- No API, database, persisted state, deployment-order, environment, or feature-switch compatibility risk exists. Old and new runners remain independently valid, and rollback is code-only.

## Acceptance criteria

- Non-streaming X JSON response inspection has one deterministic total-work bound across all decoded chunks.
- Chunking and gzip/deflate delivery cannot reset or multiply the allowance.
- The direct buffered fallback that uses the same extractor is bounded consistently.
- Exhaustion contributes no partial billing observations, records the existing parse-error signal once for the affected response, and does not alter response forwarding.
- A later flow remains inspectable after exhaustion.
- Ordinary X response shapes and large discarded ASCII strings continue to parse within the allowance.
- Existing X NDJSON row semantics and its independent row lifecycle remain unchanged.
- Deterministic tests cover dense structural input, cross-chunk accumulation, compressed multi-chunk delivery, partial-observation discard, unchanged forwarding, and later-flow recovery without wall-clock assertions.
- Ruff, BasedPyright, locked-environment checks, and focused/full mitm-addon tests pass.

## Delivery

One independently reviewable PR will change the X document extractor policy and direct plus integration coverage. Existing large-string tests will be reused rather than duplicated. No migration, rollout flag, documentation change, or follow-up slice is required.

seven332:
## Summary

- bound non-streaming X JSON structural inspection to 65,536 work units per response document
- apply the same provider-owned policy to incremental parsing and the buffered fallback while preserving X NDJSON's independent per-row lifecycle
- discard partial billing observations on exhaustion, preserve response bytes, and retain the existing one-shot lost-visibility diagnostics and later-flow recovery
- add deterministic integration coverage for ordinary responses, split identity input, gzip/deflate multi-chunk decoding, and Brotli/Zstandard fallback parsing

## Scope

- no shared `JsonSelectiveExtractor` default or response decoder changes
- no API, persisted-state, database, environment, feature-switch, or deployment-order changes

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_x_response_parsers.py tests/test_x_connector_pipeline.py -q` (75 passed)
- `uv run --no-sync python -m pytest tests/` (4,742 passed)
- `lefthook run pre-commit`

Closes #25104

### DIFF
--- a/crates/runner/mitm-addon/src/usage/providers/connectors/x.py
+++ b/crates/runner/mitm-addon/src/usage/providers/connectors/x.py
@@ -102,6 +102,9 @@ def _strip_request_target_query(request_target: str) -> str:
 # exceeding it indicates malformed or hostile upstream data, so the parser
 # discards that row through its terminating newline to protect memory.
 _MAX_NDJSON_LINE_BYTES = LARGE_RESPONSE_DECOMPRESS_LIMIT
+# Bound dense syntax and slow scalar inspection across one non-streaming X JSON
+# response while retaining the selective parser's bulk discarded-string path.
+_MAX_JSON_RESPONSE_WORK_UNITS = 65_536
 # Bound dense syntax and slow scalar inspection while keeping multi-megabyte
 # ordinary discarded strings on the selective parser's bulk-scan path.
 _MAX_NDJSON_ROW_WORK_UNITS = 65_536
@@ -241,6 +244,7 @@ def _create_x_json_selective_extractor() -> JsonSelectiveExtractor:
         array_count_paths={("data",), ("errors",)},
         wildcard_array_count_paths={("includes", "*")},
         object_presence_paths={(), ("data",)},
+        max_work_units=_MAX_JSON_RESPONSE_WORK_UNITS,
     )
 
 
--- a/crates/runner/mitm-addon/tests/test_x_connector_pipeline.py
+++ b/crates/runner/mitm-addon/tests/test_x_connector_pipeline.py
@@ -13,13 +13,18 @@ from mitmproxy.flow import Error
 import flow_metadata_keys as metadata_keys
 import mitm_addon
 import usage
-from body_limits import STREAM_BUFFER_LIMIT
+from body_limits import (
+    STREAM_BUFFER_LIMIT,
+    STREAM_DECODE_CHUNK_LIMIT,
+    STREAM_DECODE_EXPANSION_GRACE,
+    STREAM_DECODE_MAX_EXPANSION_RATIO,
+)
 from tests.flow_helpers import response_stream
 from tests.jsonl_log_helpers import read_jsonl_entries_after_flush
 from tests.x_flow_helpers import (
     json_body_that_exceeds_integer_digit_limit,
     json_body_that_exceeds_nesting_limit,
-    json_body_that_exceeds_x_ndjson_work_limit,
+    json_body_that_exceeds_x_json_work_limit,
     make_x_pipeline_flow,
     make_x_stream_pipeline_flow,
 )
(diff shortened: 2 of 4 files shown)


seven332:
## Code Review: PR #25118

**Reviewed head:** `ec9f9123fa17c17707b0826a4bcdd120fdda813f`

### Summary

The provider-owned 65,536-unit budget is applied at the shared non-streaming X JSON extractor factory, so it covers incremental and buffered fallback parsing without changing the shared parser default or X NDJSON's per-row lifecycle. Response forwarding and complete-document billing semantics remain intact.

### Findings

No findings.

### Testing

The focused X parser/pipeline suite passed (75 tests), the full mitm-addon suite passed (4,742 tests), and uv lock/sync, Ruff, BasedPyright, Lefthook, modified-line coverage, and CI `mitm-addon-test` checks passed. Deterministic integration coverage includes ordinary payloads, split input, gzip/deflate streaming decode, Brotli/Zstandard fallback, byte preservation, no partial billing, one-shot diagnostics, and later-flow recovery. A production workload benchmark was not run; the bounded-work property is structural and directly exercised by the integration fixtures.

### Verdict

LGTM

PR #25120, opened 2026-08-05
fix(runner): distinguish r2 body transport failures

### LINKED ISSUE #25106: fix(runner): distinguish R2 body transport failures
## Problem and evidence

The R2 template download path distinguishes three failure domains because callers react differently:

- request failures fall back to a normal deduplicated local build, or fail when the cache is required;
- invalid remote objects trigger a local rebuild and a force-overwrite of the cache object; and
- local filesystem failures fail locally.

After `get_object` succeeds, the response body is streamed through zstd and tar. The current extraction path maps every non-local decoder or archive read failure to `R2DownloadError::InvalidObject`, including an explicit I/O error yielded by the remote response body. A transient connection reset can therefore be treated as evidence that the stored object is corrupt, causing `runner build` to select the force-overwrite-invalid-remote path.

The behavior remains present on current `main`. Existing tests cover failures before a response is returned and malformed bodies delivered to completion, but successful mock responses all use complete in-memory byte streams. They cannot expose a transport failure after partial body consumption.

## Decision

Fix the failure-domain classification, not only the coverage gap. An explicit error from the remote response body must remain a request failure after it passes through the decoder and archive reader. A body that reaches clean EOF but is truncated or malformed remains an invalid remote object, and destination filesystem failures remain local failures.

Keep the change inside the runner's R2 download implementation and its boundary-level tests. No API, runner protocol, schema, migration, or persisted-data change is needed.

## Alternatives considered

- **Add only a regression test:** insufficient because it would document the current unsafe classification instead of correcting it.
- **Classify every decoder or tar I/O error as a request failure:** incorrect because malformed or truncated objects commonly surface as decoder/archive I/O errors and must still be eligible for invalid-object replacement.
- **Keep the current behavior:** avoids code changes but retains the possibility of force-overwriting a valid object after a transient download failure.

## Risks and mitigations

- The fix must identify an error that originated from the response body without reclassifying clean-EOF corruption. Cover both cases explicitly.
- Error tracking must not buffer the archive or weaken compressed/decompressed size limits. Keep the existing streaming and validation pipeline intact.
- Partial extraction must retain the existing atomic publication and cleanup guarantees. Verify the destination and sibling staging directory after failure.

## Acceptance criteria

- A deterministic R2 mock response yields a valid archive prefix and then an explicit I/O/transport error after `get_object` succeeds.
- The mid-body error is returned as `R2DownloadError::Request`, not `InvalidObject`, and its original diagnostic remains visible.
- The build decision for this failure uses the normal deduplicated fallback rather than `ForceOverwriteInvalidRemote`.
- An existing destination is unchanged, no template is published, and the sibling `*.download.tmp` staging directory is removed.
- Clean-EOF truncated or malformed archives continue to be classified as invalid remote objects.
- Tests use deterministic synchronization with no sleeps and no external R2 service.

Audited on current `main` at `4c2a7fac17`.

seven332:
## Summary

- preserve the underlying response-body read error through archive decoding
- classify explicit mid-body R2 transport failures as request failures while keeping clean-EOF corruption as invalid object data
- add deterministic download and build integration coverage, including deduplicated upload fallback behavior

## Exclusions

- no API, protocol, schema, migration, or persisted-data changes
- no retry-policy, buffering, or archive-limit changes

## Testing

- `cargo test --manifest-path crates/Cargo.toml -p runner r2_cache::tests::download`
- `cargo test --manifest-path crates/Cargo.toml -p runner cmd::build::tests::template_cache`

Closes #25106

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -2627,6 +2627,7 @@ dependencies = [
  "guest-reseed",
  "guest-write-file",
  "hex",
+ "http-body 1.1.0",
  "httpmock",
  "idna",
  "libc",
--- a/crates/runner/Cargo.toml
+++ b/crates/runner/Cargo.toml
@@ -72,6 +72,7 @@ guest-reseed = { workspace = true }
 guest-write-file = { workspace = true }
 zero-cli = { workspace = true }
 httpmock = { workspace = true }
+http-body = "1"
 sandbox-mock = { workspace = true }
 tokio = { workspace = true, features = ["test-util"] }
 tracing-test-support.workspace = true
--- a/crates/runner/src/cmd/build/tests/fixtures.rs
+++ b/crates/runner/src/cmd/build/tests/fixtures.rs
@@ -3,6 +3,8 @@ use super::*;
 use aws_smithy_mocks::{Rule, RuleMode, mock, mock_client};
 use std::sync::Arc;
 
+use crate::test_fixtures::http_body::byte_stream_with_error_after;
+
 pub(super) const TEST_TEMPLATE_DISK_BYTES: u64 = 128 * 1024 * 1024;
 
 #[derive(clap::Parser)]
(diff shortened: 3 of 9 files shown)


seven332:
## Code Review: PR #25120

**Reviewed head:** `cda8523028b516108cf1da0cc8f0516e9c95c2f6`

### Summary

The change preserves remote response-body failure provenance at the R2 boundary without altering the bounded streaming archive pipeline. It correctly keeps explicit body read failures on the request-fallback path while clean-EOF corruption remains eligible for invalid-object replacement.

### Findings

No findings.

### Testing

The regression coverage uses mocked external S3 operations with real archive decoding and temporary filesystem behavior. It verifies the error variant and root diagnostic, clean-EOF control, atomic destination/staging behavior, and deduplicated build fallback. Focused tests, the full runner all-target/all-feature suite, formatting, Clippy with warnings denied, and rustdoc with warnings denied passed locally. Live R2 fault injection was not run; CI is tracked separately.

### Verdict

LGTM

PR #25121, opened 2026-08-05
test(zero-cli): bound proxy child lifecycles

### LINKED ISSUE #25107: test(rust): bound Zero CLI proxy child lifecycles
## Context

`crates/zero-cli/tests/http_proxy.rs` starts an environment-isolated copy of the current test binary and a blocking TCP server for each proxy test.

The server thread waits in an unbounded `TcpListener::accept()` at `crates/zero-cli/tests/http_proxy.rs:64-92`. The parent then uses blocking `Command::output()` with no completion deadline at `crates/zero-cli/tests/http_proxy.rs:95-130`. The rescue connection and bounded request receive do not run until after the child has exited (`crates/zero-cli/tests/http_proxy.rs:133-153`).

## Problem

If a regression leaves the proxy child alive before it completes the request—for example during runtime/configuration construction or before the request-level timeout is installed—the parent remains blocked in `Command::output()`. It cannot reach the rescue connection, kill the child, or join the server. The test is then terminated only by the 20-minute Rust coverage job timeout in `.github/workflows/crates.yml:426-461`, with no focused proxy-test diagnostic.

The five-second timeout applied to the accepted stream does not cover this case because the server may still be waiting for the first connection. Removing or bypassing the child request deadline is therefore capable of turning one focused regression into a whole-job timeout.

## Acceptance criteria

- Give each proxy child process an explicit test-local completion deadline.
- On timeout, kill and reap the exact child before reporting the failure, while retaining captured stdout/stderr for diagnostics.
- Ensure the TCP fixture is always unblocked and joined on success, child failure, and child timeout; no server thread may remain in `accept()` or `read()`.
- Add a deterministic regression case proving that a deliberately hung child is killed and the fixture returns within the local bound.
- Keep the existing proxy precedence, `no_proxy`, and HTTPS fallback assertions unchanged.

Audited at `f30cb0f8456cbd96fa0441f8f8e5d8d8f7ccab8f`.

seven332:
## Summary

- supervise proxy test children with a 10-second local deadline and exact-child kill/reap
- capture stdout and stderr in anonymous files and cancel/join the TCP fixture before assertions
- add a synchronized real hung-child regression covering timeout diagnostics and cleanup

## Scope

Test-only change in `crates/zero-cli/tests/http_proxy.rs`. No production code, dependency, workflow, API, schema, persisted-state, deployment, or CI-timeout changes.

## Validation

- `cargo test -p zero-cli --test http_proxy -- --nocapture`
- repeated the exact timeout regression 20 times
- `cargo test -p zero-cli --all-targets --all-features`
- `cargo clippy -p zero-cli --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p zero-cli --all-features --no-deps`

Closes #25107

### DIFF
--- a/crates/zero-cli/tests/http_proxy.rs
+++ b/crates/zero-cli/tests/http_proxy.rs
@@ -113,6 +113,15 @@ enum ServerOutcome {
     Cancelled,
 }
 
+impl ServerOutcome {
+    const fn diagnostic(&self) -> &'static str {
+        match self {
+            Self::Request(_) => "request captured",
+            Self::Cancelled => "cancelled",
+        }
+    }
+}
+
 struct HttpServer {
     address: SocketAddr,
     stop: Arc<AtomicBool>,
@@ -460,10 +469,11 @@ fn finish_proxy_case(child: CommandExecution, server: HttpServer) -> io::Result<
                 .map_err(|error| io::Error::new(io::ErrorKind::InvalidData, error))
         }
         (Ok(output), Ok(server_outcome)) => Err(io::Error::other(format!(
-            "proxy child exited with {}; stdout={:?}; stderr={:?}; server={server_outcome:?}",
+            "proxy child exited with {}; stdout={:?}; stderr={:?}; server={}",
             output.status,
             String::from_utf8_lossy(&output.stdout),
-            String::from_utf8_lossy(&output.stderr)
+            String::from_utf8_lossy(&output.stderr),
+            server_outcome.diagnostic()
         ))),
         (Ok(output), Err(server_error)) => Err(io::Error::new(
             server_error.kind(),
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #25121

**Reviewed head:** `a9c52c2de061ec0eeeeb16ff0bb6f30787d09db2`

### Summary

The test-local child owner, file-backed output capture, and cancellation-aware TCP owner replace every previously unbounded lifecycle edge without changing production code or the three existing proxy behavior assertions. The synchronized hung-child regression exercises a real subprocess and real listener, verifies successful kill/reap, and bounds fixture cleanup.

### Findings

No findings.

### Testing

Reviewed the six submitted commits and the aggregate diff. Local validation covered the full `http_proxy` target, all zero-cli targets/features, formatting, Clippy, rustdoc, and 20 repetitions of the exact timeout regression. Workspace coverage instrumentation and platform-specific CI matrices remain delegated to GitHub checks.

### Verdict

LGTM

PR #25123, opened 2026-08-05
fix(platform): scale mermaid diagrams to fit their box

bingjiez666:
## Problem

Mermaid diagrams are not scaled into their box. The `` fills the box's width and keeps its own aspect ratio vertically, so a tall flowchart runs past the bottom of the 4:3 frame and is cut off by `overflow: hidden`.

## Cause

`.mermaid-diagram-image` positions the image absolutely with `inset: 8px` and relies on `object-fit: contain`, but never states a width or height. An absolutely positioned replaced element whose `width` and `height` are `auto` is laid out at its intrinsic size (here bounded only by the markdown theme's `img { max-width: 100% }`), and the `bottom`/`right` insets are ignored. The image box therefore already matches the diagram's aspect ratio and `object-fit` has nothing to fit into.

Diagrams are serialized at `LIGHTBOX_SVG_WIDTH = 1600` so the lightbox can open them at full size, which makes the overflow large.

## Fix

State the image's size so the inset rectangle becomes the fit box. `calc(100% - 16px)` rather than `100%`, because a percentage on an absolutely positioned child resolves against the padding box, which the box's own 8px padding would not clear.

## Verification

Reproduced the layout in a browser with the shipped rules and a 1600x4000 SVG, measuring the rendered image against its box:

| | image | box height | contained |
| --- | --- | --- | --- |
| before | 418 x 1045 | 315 | no, clipped |
| after | 402 x 297 | 315 | yes, letterboxed |

The existing mermaid tests run in jsdom with mermaid mocked, so they assert markup rather than layout and could not catch this.

## Note

Diagrams now fit, but a strongly vertical flowchart is letterboxed into a 420x315 frame and its labels end up small. Changing that means letting the frame follow the diagram's own ratio, which trades away the reserved-height behaviour that keeps a thread from shifting while a diagram renders. Left as is here; happy to follow up if we want the frame to adapt.

### DIFF
--- a/turbo/apps/platform/src/views/css/__tests__/entrypoint-safe-area.test.ts
+++ b/turbo/apps/platform/src/views/css/__tests__/entrypoint-safe-area.test.ts
@@ -1,9 +1,7 @@
-import { existsSync, readFileSync } from "node:fs";
-import { join } from "node:path";
-
 import { describe, expect, it } from "vitest";
 
 import indexHtml from "../../../../index.html?raw";
+import { readGlobalCss } from "./global-css.ts";
 
 function getViewportDirectives(): string[] {
   const match = /<meta\s+name="viewport"\s+content="([^"]+)"/.exec(indexHtml);
@@ -12,20 +10,6 @@ function getViewportDirectives(): string[] {
       return directive.trim();
     }) ?? []
   );
-}
-
-function readGlobalCss(): string {
-  const candidates = [
-    join(process.cwd(), "src/views/css/index.css"),
-    join(process.cwd(), "apps/platform/src/views/css/index.css"),
-  ];
-  const path = candidates.find((candidate) => {
-    return existsSync(candidate);
-  });
-  if (path === undefined) {
-    throw new Error("Unable to locate platform global CSS");
-  }
-  return readFileSync(path, "utf8");
 }
 
 describe("platform entrypoint safe area behavior", () => {
--- a/turbo/apps/platform/src/views/css/__tests__/global-css.ts
+++ b/turbo/apps/platform/src/views/css/__tests__/global-css.ts
@@ -1,0 +1,32 @@
+import { existsSync, readFileSync } from "node:fs";
+import { join } from "node:path";
+
+/**
+ * Reads the platform's global stylesheet as text. Tests run in happy-dom, which
+ * applies no CSS and performs no layout, so a rule can only be asserted from its
+ * own source. The path depends on whether vitest was started from the workspace
+ * or from the turbo root.
+ */
+export function readGlobalCss(): string {
+  const candidates = [
+    join(process.cwd(), "src/views/css/index.css"),
+    join(process.cwd(), "apps/platform/src/views/css/index.css"),
+  ];
+  const path = candidates.find((candidate) => {
+    return existsSync(candidate);
+  });
+  if (path === undefined) {
+    throw new Error("Unable to locate platform global CSS");
+  }
+  return readFileSync(path, "utf8");
+}
+
+/** Returns the declarations of the rule introduced by `selector`. */
+export function readCssRule(css: string, selector: string): string {
+  const start = css.indexOf(`${selector} {`);
+  if (start === -1) {
+    throw new Error(`Unable to locate CSS rule for ${selector}`);
+  }
+  const end = css.indexOf("}", start);
+  return css.slice(start, end);
+}
--- a/turbo/apps/platform/src/views/css/__tests__/mermaid-diagram-fit.test.ts
+++ b/turbo/apps/platform/src/views/css/__tests__/mermaid-diagram-fit.test.ts
@@ -1,0 +1,64 @@
+import { describe, expect, it } from "vitest";
+
+import { readCssRule, readGlobalCss } from "./global-css.ts";
+
+const BOX_SELECTOR = ".wmde-markdown .mermaid-diagram-expand";
+const IMAGE_SELECTOR = ".wmde-markdown .mermaid-diagram-image";
+const INSET_CHILDREN_SELECTOR = `${IMAGE_SELECTOR},\n.wmde-markdown .mermaid-diagram-pending`;
+
+function readPixels(rule: string, pattern: RegExp): number {
+  const match = pattern.exec(rule);
+  if (!match?.[1]) {
+    throw new Error(`Unable to read ${String(pattern)} from ${rule}`);
+  }
+  return Number(match[1]);
+}
+
+describe("mermaid diagram box", () => {
+  it("scales a diagram of any ratio into the box instead of clipping it", () => {
+    const image = readCssRule(readGlobalCss(), IMAGE_SELECTOR);
+
+    // `object-fit` only letterboxes a diagram once the image has a box to fit
+    // into. An absolutely positioned <img> sized `auto` is laid out at its own
+    // intrinsic size and ignores the opposite insets, so dropping either of
+    // these declarations leaves a tall diagram running past the box and a wide
+    // one anchored to the top edge.
+    expect(image).toMatch(/width:\s*calc\(100% - \d+px\);/);
+    expect(image).toMatch(/height:\s*calc\(100% - \d+px\);/);
+    expect(image).toMatch(/object-fit:\s*contain;/);
+  });
+
+  it("keeps the fitted size in step with the box padding", () => {
+    const globalCss = readGlobalCss();
+    const padding = readPixels(
+      readCssRule(globalCss, BOX_SELECTOR),
+      /padding:\s*(\d+)px;/,
+    );
+    const inset = readPixels(
+      readCssRule(globalCss, INSET_CHILDREN_SELECTOR),
+      /inset:\s*(\d+)px;/,
+    );
+    const removedFromSize = readPixels(
+      readCssRule(globalCss, IMAGE_SELECTOR),
+      /width:\s*calc\(100% - (\d+)px\);/,
+    );
+
+    // The image is placed against the padding box, so its size has to give back
+    // the padding on both edges itself. A padding change that misses the `calc`
+    // would push the diagram back out of the box.
+    expect(inset).toBe(padding);
+    expect(removedFromSize).toBe(padding * 2);
+  });
+
+  it("reserves the same box for every diagram before it renders", () => {
+    const globalCss = readGlobalCss();
+    const box = readCssRule(globalCss, BOX_SELECTOR);
+
+    expect(box).toMatch(/max-width:\s*420px;/);
+    expect(box).toMatch(/aspect-ratio:\s*4 \/ 3;/);
+    expect(box).toMatch(/overflow:\s*hidden;/);
+    expect(readCssRule(globalCss, INSET_CHILDREN_SELECTOR)).toMatch(
+      /position:\s*absolute;/,
+    );
+  });
+});
--- a/turbo/apps/platform/src/views/css/index.css
+++ b/turbo/apps/platform/src/views/css/index.css
@@ -1104,6 +1104,13 @@ body {
 }
 
 .wmde-markdown .mermaid-diagram-image {
+  /* An absolutely positioned <img> whose width and height are `auto` is laid
+     out at its own intrinsic size, so `inset` alone leaves the diagram at
+     lightbox scale and `object-fit` has nothing to fit into. The size is
+     therefore stated: `calc` rather than `100%`, since a percentage resolves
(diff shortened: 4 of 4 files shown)


bingjiez666:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 2cfe449b15cb93f9dba88d6da43025a7042211aa

### Summary

CSS-only fix in `turbo/apps/platform/src/views/css/index.css`: `.wmde-markdown .mermaid-diagram-image` now states `width: calc(100% - 16px)` and `height: calc(100% - 16px)` so `object-fit: contain` has a fit box. The diagnosis is correct — an absolutely positioned replaced element with `width`/`height: auto` is laid out at its intrinsic size and ignores the `bottom`/`right` insets, so `inset: 8px` alone never constrained the diagram and tall flowcharts were clipped by the box's `overflow: hidden`. The `calc` form (rather than `100%`) is also correct, since a percentage on an absolutely positioned child resolves against the padding box and would overflow the 8px padding.

Note that this also changes the previously "working" case: wide diagrams used to be anchored to the top-left of the inset rectangle and are now letterboxed centre. That is the intended behaviour, but it is an existing-path behaviour change with no coverage.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | `fix:` commits must ship a regression test | Finding (P0) | `turbo/apps/platform/src/views/css/index.css` is the only changed file; no test accompanies the fix. |
| `REVIEW.md` | Feature regression guard — changed existing behaviour needs coverage for the path most likely to regress | Finding (P0) | Wide/short diagrams change from top-left anchoring to centred letterboxing; nothing pins either the new or the old path. |
| `REVIEW.md` | Correctness and logic — no wrong assumptions or unhandled edge cases | Pass | Sizing reasoning verified in a browser against the shipped rules: image 418×1045 in a 315px box before, 402×297 fully contained after. `.mermaid-diagram-pending` is a non-replaced `<span>` and keeps stretching to `inset`, so it is unaffected. |
| `REVIEW.md` | Security — no injection, secrets, or missing auth surfaces | N/A | Change is a static stylesheet declaration with no data flow, markup, or endpoint. |
| `REVIEW.md` | Type safety — no `any`, no broad types | N/A | No TypeScript in the diff. |
| `REVIEW.md` | New user-facing features gated by a `FeatureSwitchKey` | N/A | Not a new feature; mermaid rendering already sits behind `FeatureSwitchKey.MermaidDiagrams`. |
| `docs/bad-smell.md` | No lint/type suppression comments | Pass | No suppression added; the new comment explains WHY the size must be stated. |
| `docs/bad-smell.md` | No fallback/recovery patterns — fail fast | Pass | Declarative sizing only, no conditional or fallback path. |
| `docs/bad-smell.md` | No hardcoded environment-specific values | Pass | The `16px` is the box's own padding doubled, a local layout constant, not configuration. See the non-blocking note below. |
| `docs/testing.md` | Integration tests over unit tests, exercised through the production entry path | Finding (P0) | No test added at all; the existing `markdown.test.tsx` runs in jsdom with mermaid mocked and asserts markup, so it cannot observe this. |
| `docs/testing.md` | Mock only at the system boundary (no `vi.mock("../…")`, no fetch stubbing, real DB) | N/A | No test files in the diff. |
| `docs/effect.md` | Choose the right execution boundary (`computed` / semantic command / `onRef` / React effect) | N/A | No signal, command, or effect changed; the render path in `mermaid-diagram.tsx` and `signals/mermaid-diagram.ts` is untouched. |
| `docs/cache.md` | Render purity, state ownership, and symmetric resource lifecycles (RC-1, RC-4, RC-8) | N/A | No state, cache, ref, or resource lifecycle touched; the diagram result refcounting is unchanged. |
| `.claude/skills/ccstate/SKILL.md` | ccstate signal, `onRef`, and AbortSignal ownership practices | N/A | No ccstate code in the diff. |
| `docs/deployment-compatibility.md` | Preserve compatibility across independently deployed components | N/A | Stylesheet ships with the frontend bundle; no API, runner protocol, queue payload, or persisted state involved. |

### Findings

- **Missing regression test for a `fix:` commit.** `REVIEW.md` treats a `fix:` without a regression test as P0. This is testable in the repo's existing style: `turbo/apps/platform/src/views/css/__tests__/entrypoint-safe-area.test.ts` already reads `src/views/css/index.css` and asserts specific declarations with regexes, precisely because layout rules cannot be observed in jsdom. Add an equivalent assertion for `.mermaid-diagram-image` — that it declares `object-fit: contain` **together with** an explicit `width` and `height` — so the pairing cannot silently regress again. Without it, deleting either new line restores the clipping bug with a green suite.

- **Changed existing path is uncovered.** Wide/short diagrams previously rendered top-left anchored and now render centred. The same CSS-source test can pin the intended rule set; otherwise the only evidence that either aspect ratio renders correctly is a manual check.

### Non-blocking notes

- `calc(100% - 16px)` encodes the parent's `padding: 8px` as a doubled literal in a second rule. If `.mermaid-diagram-expand`'s padding is ever changed, the image silently desyncs from `inset: 8px` and the letterbox goes slightly off. A shared custom property (e.g. `--mermaid-box-padding`) used by both the padding and the `calc` would keep them in lockstep. Not a defect today.

- The diagram now fits, but a strongly vertical flowchart is letterboxed into the fixed 420×315 frame and its labels become small. The PR description already flags this and defers it; that scoping is reasonable for a clipping fix.

### Testing

- **Coverage:** Insufficient — no test accompanies a `fix:` commit, and a suitable, already-established test shape exists in this repo (`views/css/__tests__/entrypoint-safe-area.test.ts`).
- **Conventions:** No test files changed, so no convention violations. The existing mermaid tests (`views/components/__tests__/markdown.test.tsx`) mock mermaid and assert markup only; they are not a substitute for pinning the sizing rule.
- **Manual verification (from the PR description, reproduced in a browser against the shipped rules):** before — image 418×1045 in a 315px box, clipped; after — 402×297, fully contained.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 87b9aaa8614f57702fabf361d7912f4d56277890
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Two commits. `2cfe449` fixes the mermaid diagram clipping: `.wmde-markdown .mermaid-diagram-image` now states `width`/`height` so `object-fit: contain` has a box to fit into — an absolutely positioned replaced element sized `auto` is laid out at its intrinsic size and ignores the opposite insets, so `inset: 8px` alone never constrained a diagram serialized at `LIGHTBOX_SVG_WIDTH`. `87b9aaa` answers the previous review: it adds `views/css/__tests__/mermaid-diagram-fit.test.ts` and lifts the stylesheet reader out of `entrypoint-safe-area.test.ts` into a shared `global-css.ts`.

The regression test is effective: with the two sizing declarations removed, two of its three cases fail; with them present, all pass.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | `fix:` commits must ship a regression test | Pass | `mermaid-diagram-fit.test.ts` pins `width`/`height` next to `object-fit: contain`; verified to fail when either declaration is removed. Resolves the P0 from the previous review. |
| `REVIEW.md` | Feature regression guard — changed existing behaviour needs coverage | Pass | Third case pins the reserved box (`max-width: 420px`, `aspect-ratio: 4 / 3`, `overflow: hidden`, absolutely positioned children), which is the invariant shared by tall and wide diagrams. |
| `REVIEW.md` | Correctness and logic | Pass | Rule lookup is unambiguous: `.mermaid-diagram-image {` cannot match the two-selector block (that line ends in a comma) and `.mermaid-diagram-expand {` cannot match `:disabled`; neither block contains a nested `}` that would truncate the slice. |
| `REVIEW.md` | Type safety — no `any`, no broad types | Pass | `readGlobalCss`, `readCssRule`, and `readPixels` all have explicit parameter and return types; no assertions or casts. |
| `REVIEW.md` | Security | N/A | Stylesheet declaration plus test-only file reads under the workspace; no data flow, markup, or endpoint. |
| `REVIEW.md` | New user-facing features gated by a `FeatureSwitchKey` | N/A | Not a new feature; mermaid rendering already sits behind `FeatureSwitchKey.MermaidDiagrams`. |
| `docs/bad-smell.md` | No lint/type suppressions, no `any` | Pass | None added; lint runs clean over the changed workspace. |
| `docs/bad-smell.md` | Fail fast, no fallback/recovery logic | Pass | `readGlobalCss` and `readCssRule` throw on a missing file or selector rather than returning an empty string, so a moved rule fails loudly instead of vacuously passing. |
| `docs/bad-smell.md` | No duplicated logic | Pass | The stylesheet reader is shared rather than copied into the new file; `entrypoint-safe-area.test.ts` now imports it. |
| `docs/testing.md` | Integration over unit tests; mock only at the boundary | Pass | No mocks at all — the tests read the real stylesheet from the real filesystem. No `vi.mock`, no fetch stubbing, no fake timers, no DB. |
| `docs/testing/anti-patterns.md` | AP-1/AP-2/AP-3/AP-4/AP-5 — mock-call assertions, fetch mocking, fs mocking, internal mocks, fake timers | Pass | None present; the suite has no `vi.*` usage. |
| `docs/testing/anti-patterns.md` | AP-7 — do not assert on implementation details such as CSS classes | Finding (non-blocking) | The tests assert stylesheet source text, which is implementation-coupled by construction. See the note below on why the boundary exception holds here and what it costs. |
| `docs/testing/anti-patterns.md` | AP-8 — no over-testing | Pass | Three cases covering the fixed rule, its coupling to the box padding, and the reserved box; no incidental declarations pinned. |
| `docs/testing/anti-patterns.md` | AP-10 — no direct component rendering | Pass | No components rendered. |
| `docs/testing/patterns.md` | Standard file structure; no mocks or lifecycle hooks where there is no state | Pass | Plain `describe`/`it` with no `beforeEach`/`afterEach`, matching the pure-assertion shape the doc shows. |
| `docs/testing/app-testing.md` | Platform tests are page-level; do not add static-config tests | Pass (documented exception) | happy-dom applies no CSS and performs no layout, so no `detachedSetupPage` test can observe whether a diagram fits its box. `entrypoint-safe-area.test.ts` already establishes this exact exception on `main` for global stylesheet invariants, and the new file sits beside it. |
| `docs/testing/testing-external-behavior.md` | Leave the external boundary only when the state cannot be constructed through the production interface, and state why | Pass | `global-css.ts`'s doc comment records the reason (no CSS, no layout in happy-dom), which is the justification the doc requires for an exception. |
| `docs/effect.md` | Correct execution boundary (`computed` / command / `onRef` / React effect) | N/A | No signal, command, or effect changed. |
| `docs/cache.md` | Render purity, state ownership, symmetric resource lifecycles | N/A | No state, cache, ref, or resource lifecycle touched. |
| `.claude/skills/ccstate/SKILL.md` | ccstate signal, `onRef`, and AbortSignal ownership | N/A | No ccstate code in the diff. |
| `docs/deployment-compatibility.md` | Preserve compatibility across independently deployed components | N/A | Stylesheet ships with the frontend bundle; no API, runner protocol, queue payload, or persisted state. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

### Non-blocking notes

- **Syntax coupling (AP-7).** The first case requires the literal `calc(100% - Npx)` form. A behaviour-preserving refactor — say `inset: 0; padding: 8px; box-sizing: border-box`, or hoisting the padding into a custom property — would fail the test even though diagrams still fit. That is the accepted cost of asserting a stylesheet from its source, and the failure is explicit rather than silent, but it is worth remembering when the rule is next touched. The second case is the more durable one, since it asserts the padding relationship rather than the spelling.
- The diagram now fits, but a strongly vertical flowchart is letterboxed into the fixed 420×315 frame and its labels become small. Adapting the frame to the diagram's own ratio would trade away the reserved height that keeps a thread from shifting mid-render; the PR description defers this deliberately, which is the right scope for a clipping fix.

### Testing

- **Coverage:** Adequate. The `fix:` commit now carries a regression test that demonstrably fails without the fix (2 of 3 cases), covering both the sizing rule and the reserved box that governs the previously top-anchored wide-diagram path.
- **Conventions:** Compliant. No mocks, no fake timers, no direct component rendering, no internal-path `vi.mock`, no DB or fetch stubbing. The one boundary exception (asserting stylesheet source instead of page behaviour) is precedented by `entrypoint-safe-area.test.ts` on `main` and documented in the shared helper.
- **Local verification:** `pnpm format` (clean), `pnpm turbo run lint` (0 errors, 12/12 tasks), `pnpm check-types` (12/12), `knip` (clean), and the two CSS test files (7 passed).
- **CI at `87b9aaa`:** 27 successful, 22 skipped, remainder still running at review time.

PR #25135, opened 2026-08-05
fix(python): preserve auth base representation lengths

seven332:
## Summary

- preserve valid upstream representation `Content-Length` on bodyless `HEAD` and `304` auth rewrites
- keep an omitted, malformed, comma-list, repeated, or conflicting representation length absent instead of generating or forwarding unsafe metadata
- cover the final rewritten `HTTPFlow` for valid, absent, ambiguous, and prohibited lengths

## Scope

- ordinary response framing still comes from the buffered bytes through `http.Response.make`
- existing encoded-response and hop-by-hop-header behavior is unchanged
- `1xx`, `204`, and `205` responses are explicitly excluded from the `HEAD` representation-length exception
- this does not otherwise change adjacent bodyless-status, informational, or successful `CONNECT` framing behavior

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync ruff format --check .`
- `uv run --no-sync ruff check .`
- `uv run --no-sync basedpyright -p .`
- `uv run --no-sync python -m pytest tests/` (`4756 passed`)

Closes #25111

PR #25136, opened 2026-08-05
fix(runner): preserve bodyless response content length

### LINKED ISSUE #25111: bug(python): preserve HEAD and 304 Content-Length in auth.base forwarding
## Problem and evidence

The `auth.base` forwarding path loses valid selected-representation lengths on bodyless `HEAD` and `304 Not Modified` responses.

`auth_base_forwarder` correctly returns empty bytes for these response semantics and retains the upstream `Content-Length`. `_apply_url_rewrite` then reconstructs the final flow response with:

```python
flow.response = http.Response.make(status, resp_raw_content, resp_headers)
```

With the pinned `mitmproxy==12.2.3`, assigning the empty content rewrites an upstream `Content-Length: 123` to `0`. The constructor also invents `Content-Length: 0` when the upstream bodyless response omitted the field. This is reproducible on current `main`; the existing low-level bodyless-response test stops before final flow reconstruction.

For `HEAD`, `Content-Length` describes the representation that the corresponding `GET` would return. For `304`, it can describe the selected `200` representation. Replacing that metadata with the forwarded body byte count breaks size discovery and conditional-cache handling for sandbox clients.

## Decision

Keep `http.Response.make` as the normalization boundary for ordinary responses. Before construction, capture exactly one syntactically valid upstream representation length for `HEAD` or `304`. After construction, remove the generated zero and restore the captured length when present; when upstream omitted or supplied an ambiguous value, leave the field absent.

This is one independently reviewable Python addon PR. It does not change APIs, schemas, persisted state, queue payloads, or runner protocols.

## Alternatives considered

- Constructing `http.Response` directly was rejected because it bypasses byte-derived framing normalization for all ordinary, dechunked, and encoded responses and increases mitmproxy-internal coupling.
- Extending the forwarder return contract was rejected because one consumer needs the semantic distinction and the final response still requires method-aware correction.
- Copying upstream `Content-Length` values without validation was rejected because external malformed or conflicting framing metadata must not gain a new passthrough path.

## Risks and boundaries

- The correction is gated to `HEAD` or `304`; ordinary response lengths continue to come from actual forwarded bytes.
- Existing `Content-Encoding` restoration and hop-by-hop filtering remain unchanged.
- Only one nonempty ASCII-decimal field is restored. Missing, malformed, comma-list, or repeated fields remain absent.
- `204`, informational, and successful `CONNECT` framing are adjacent protocol behavior but are not given this representation-length exception.

## Acceptance criteria

- Preserve a valid upstream representation `Content-Length` on final rewritten `HEAD 200` and `GET 304` flows while keeping their body empty.
- Preserve the absence of `Content-Length` for those responses instead of inventing zero.
- Do not restore malformed, comma-list, or repeated external lengths.
- Continue normalizing framing and content length from actual bytes for ordinary responses.
- Add complete auth rewrite integration tests that assert the final `flow.response` behavior.
- Keep existing encoded-response and hop-by-hop-header behavior unchanged.
- Pass the complete Python addon formatting, static-analysis, and pytest checks.

seven332:
## Summary

- preserve a valid upstream representation `Content-Length` on rewritten `HEAD` and `304` responses
- keep missing, malformed, comma-list, and repeated lengths absent instead of exposing a constructor-generated zero
- retain byte-derived framing for ordinary responses and exclude `204`/`205` from the representation-length exception

No API, schema, persisted-state, queue, or runner protocol changes are included.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (4,744 passed)

Closes #25111

### DIFF
--- a/crates/runner/mitm-addon/src/auth.py
+++ b/crates/runner/mitm-addon/src/auth.py
@@ -12,6 +12,7 @@ import urllib.parse
 from collections.abc import Callable
 from dataclasses import dataclass, field
 from enum import Enum
+from http import HTTPStatus
 
 from mitmproxy import http
 
@@ -1076,11 +1077,27 @@ async def _apply_url_rewrite(
             for name, value in header_pairs(resp_headers)
             if name.lower() == "content-encoding"
         ]
+        preserve_representation_content_length = status == HTTPStatus.NOT_MODIFIED or (
+            flow.request.method == "HEAD"
+            and status >= HTTPStatus.OK
+            and status != HTTPStatus.NO_CONTENT
+        )
+        representation_content_length: str | None = None
+        if preserve_representation_content_length:
+            content_lengths = resp_headers.get_all("Content-Length")
+            if len(content_lengths) == 1:
+                candidate = content_lengths[0].strip(" \t")
+                if candidate and all("0" <= character <= "9" for character in candidate):
+                    representation_content_length = candidate
         if content_encodings:
             del resp_headers["Content-Encoding"]
         # The forwarder returns representation bytes, but Response.make expects
         # decoded content. Hide the codings while it normalizes response framing.
         flow.response = http.Response.make(status, resp_raw_content, resp_headers)
+        if preserve_representation_content_length:
+            del flow.response.headers["Content-Length"]
+            if representation_content_length is not None:
+                flow.response.headers["Content-Length"] = representation_content_length
         if content_encodings:
             flow.response.headers.set_all("Content-Encoding", content_encodings)
     except ForwardedRequestTooLargeError:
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #25136

**Reviewed head:** `f094232259f160595f23b74b2020e30108227c1e`

### Summary

The final response-construction boundary now preserves validated representation lengths for eligible `HEAD` and `304` responses while retaining mitmproxy's byte-derived framing for ordinary responses. The status gate correctly excludes `1xx`, `204`, and `205`, and ambiguous external length fields remain absent.

### Findings

No findings.

### Testing

Ruff format/lint, BasedPyright, 29 focused rewrite tests, and the complete 4,744-test addon suite passed. No live external mitmproxy interception smoke test was run; the repository's handler integration suite covers this boundary with a real `HTTPFlow` and controlled upstream socket.

### Verdict

LGTM

seven332:
Superseded by #25135, which was opened first for the same implementation issue. Continuing the final review and fixes there to keep a single delivery path for #25111.

PR #25138, opened 2026-08-05
fix(runner): clean up cancelled setup artifact temps

### LINKED ISSUE #25113: bug(rust): clean up setup artifact temps when installation is cancelled
## Problem and evidence

`runner setup` can strand same-directory artifact temp files when an install
future is cancelled after temp creation. Cleanup is currently sequenced after
awaited work instead of being owned for the full create-to-install lifecycle.

The affected paths are:

1. `download_to_temp` creates a temp and then awaits response streaming,
   writes, flush, and file conversion. Explicit unlink only runs after a
   returned error.
2. `download_and_extract` unlinks the verified archive only after awaiting
   `extract_tar_entry`. A started `spawn_blocking` extraction continues if
   its async join owner is dropped, and neither the archive nor a successful
   detached output currently has drop cleanup.
3. `atomic_install_produced` moves the produced artifact into
   `spawn_blocking`, but its failed-install cleanup runs only after awaiting
   the join handle.
4. A temp can also leak if setup-specific security validation fails immediately
   after creation, or if cancellation occurs while an explicit asynchronous
   cleanup is awaited.

`run_setup` and runner GC do not sweep these artifact directories. Later
attempts create new UUID paths, so abandoned files are neither reused nor
removed. The largest pinned setup source is currently the x86_64 mitmproxy
archive at 119,209,168 bytes.

## Decision

Proceed with one cancellation-safety PR. Make temporary-path cleanup a resource
ownership invariant from successful file creation until verified atomic
installation.

The runner already depends on `tempfile` at runtime. Use
`tempfile::TempPath` rather than adding another custom guard or a second
cleanup mechanism.

- Preserve the existing custom same-directory `create_new` open, `0600`
  mode, `O_NOFOLLOW`, directory, owner, regular-file, held-fd, and inode
  validation.
- Wrap the created path with `TempPath::from_path` immediately after the open,
  before any later fallible validation or await.
- Carry the `TempPath` beside the held `File` in
  `ProducedSetupArtifact`.
- Move the complete verified archive owner into the blocking extractor, and
  give every extracted output the same owned-path lifecycle.
- Keep all size, SHA-256, permission, path/fd identity, and target inode checks.
- Use `TempPath::persist(target)` for the existing same-filesystem atomic
  replacement. Success consumes the cleanup owner; failure retains it until
  drop.
- Remove redundant manual unlink branches. They create additional cancellation
  windows and obscure the ownership invariant.

A complete, verified atomic install may remain installed if cancellation races
with successful rename. Cancellation is not a transaction rollback.

## Alternatives considered

- **Setup-specific active/disarm guard:** technically valid and matches other
  repository patterns, but duplicates lifecycle behavior already present in a
  production dependency and introduces a manual rename/disarm window.
- **Cancellation tokens or cleanup supervisors:** cannot abort a started
  `spawn_blocking` task and still require resource-owned cleanup.
- **Startup or GC sweep:** may be useful for historical residue, but cannot
  safely distinguish another setup process's active temp without a separate
  lock, lease, or age policy. It does not fix active ownership.

## Risks and mitigations

- `TempPath` drop performs one synchronous best-effort local unlink. This is
  consistent with existing repository temp guards; operation errors continue
  to propagate normally.
- Generic tempfile creation could weaken setup-specific trust checks, so only
  `TempPath::from_path` is used after the existing secure open.
- A manual rename followed by late disarm could remove a newly recreated old
  path, so ownership transfer is performed by `persist` as one operation.
- Detached blocking work must own all input and output temps, not only cloned
  path strings.
- Cancellation tests use explicit gates and observable filesystem state, not
  timing sleeps.

## Acceptance criteria

- Aborting a direct artifact install after its download temp exists removes the
  temp and does not install an incomplete target.
- Cancelling tar extraction removes both the verified archive and extracted
  output, including when blocking extraction finishes after its async caller is
  gone.
- Cancelling while final blocking installation is in flight removes an
  uninstalled temp after detached work completes; a fully verified successful
  replacement may remain installed.
- Creation-time validation and all existing explicit error paths leave no
  setup temps.
- Existing size, SHA-256, fd/inode, ownership, permission, atomic replacement,
  and concurrent verified-target behavior remains unchanged.
- Focused tests gate the real extraction and installation closures with
  explicit synchronization and assert externally visible filesystem outcomes.
- Runner formatting, all-target/all-feature Clippy, documentation, focused
  setup tests, and the full runner test scope pass.

## Delivery notes

- One independently reviewable runner PR.
- No API, queue, runner/guest protocol, schema, database, persisted-state, or
  migration change.
- No mixed-version rollout requirement.
- Historical temps created by older binaries are not part of this active
  ownership fix.

seven332:
## Summary

- own setup artifact temporary paths with a drop guard until the final rename succeeds
- carry temporary-file ownership through direct downloads, archive extraction, and blocking installation work
- cover cancellation during direct downloads and detached blocking extraction/install work

## Exclusions

- does not scan for or remove temporary files left by older runner processes
- does not attempt cleanup after an uncatchable process kill or host crash
- does not change APIs, schemas, protocols, dependencies, or persisted data

## Validation

- `cargo test --manifest-path crates/Cargo.toml -p runner cmd::setup`
- setup test suite repeated 10 times

Closes #25113

### DIFF
--- a/crates/runner/src/cmd/setup/artifacts.rs
+++ b/crates/runner/src/cmd/setup/artifacts.rs
@@ -210,6 +210,7 @@ mod tests {
     use sha2::{Digest, Sha256};
     use tokio::io::{AsyncReadExt, AsyncWriteExt};
     use tokio::net::TcpListener;
+    use tokio::sync::oneshot;
     use tokio::task::JoinHandle;
 
     use super::*;
@@ -294,6 +295,50 @@ mod tests {
         (format!("http://{address}/artifact"), task)
     }
 
+    async fn spawn_gated_http_response(
+        response_prefix: Vec<u8>,
+    ) -> (
+        String,
+        oneshot::Receiver<()>,
+        oneshot::Sender<()>,
+        JoinHandle<std::io::Result<()>>,
+    ) {
+        let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
+        let address = listener.local_addr().unwrap();
+        let (body_sent_tx, body_sent_rx) = oneshot::channel();
+        let (release_tx, release_rx) = oneshot::channel();
+        let task = tokio::spawn(async move {
+            let (mut stream, _) = listener.accept().await?;
+            let mut request = [0_u8; 1024];
+            let request_size = stream.read(&mut request).await?;
+            if request_size == 0 {
+                return Err(std::io::Error::new(
+                    std::io::ErrorKind::UnexpectedEof,
+                    "setup fixture received an empty request",
+                ));
+            }
+            stream.write_all(&response_prefix).await?;
+            body_sent_tx.send(()).map_err(|()| {
+                std::io::Error::new(
+                    std::io::ErrorKind::BrokenPipe,
+                    "setup fixture body signal receiver dropped",
+                )
+            })?;
+            release_rx.await.map_err(|_| {
+                std::io::Error::new(
+                    std::io::ErrorKind::BrokenPipe,
+                    "setup fixture release sender dropped",
+                )
+            })
+        });
+        (
+            format!("http://{address}/artifact"),
+            body_sent_rx,
+            release_tx,
+            task,
+        )
+    }
+
     async fn finish_http_response(task: JoinHandle<std::io::Result<()>>) {
         tokio::time::timeout(Duration::from_secs(2), task)
             .await
@@ -321,6 +366,20 @@ mod tests {
             }
         }
         temp_files
+    }
+
+    async fn wait_for_setup_temp_count(root: &Path, expected: usize) -> Vec<PathBuf> {
+        tokio::time::timeout(Duration::from_secs(2), async {
+            loop {
+                let temp_files = setup_temp_files(root);
+                if temp_files.len() == expected {
+                    return temp_files;
+                }
+                tokio::task::yield_now().await;
+            }
+        })
+        .await
+        .expect("setup temp count did not converge")
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #25138

**Reviewed head:** `a6f2306115f02f4315b31b87ff47caa016d68162`

### Summary

The setup temp path is now owned from secure creation through download, archive extraction, and final installation, and cleanup is no longer sequenced after cancellable awaits. Existing fd/inode validation and atomic-install behavior remain intact.

### Findings

No findings.

### Testing

Rust formatting, Clippy, rustdoc, the focused setup suite, 10 repeated setup-suite runs, and the runner all-target/all-feature suite passed. Cleanup after SIGKILL, host crash, or for historical temp files remains explicitly out of scope.

### Verdict

LGTM

seven332:
## Code Review: PR #25138

**Reviewed head:** `b25e77d0e92b66b821766a2bc4bbaa1e8b36261f`

### Summary

The setup temp path is owned from secure creation through download, archive extraction, and final installation, so cleanup no longer depends on code after cancellable awaits. Existing fd/inode validation and atomic-install behavior remain intact. The rebase changed only the base; `git range-diff` reports the reviewed patch as identical.

### Findings

No findings.

### Testing

On the rebased head, Rust formatting, Clippy, rustdoc, and the runner all-target/all-feature suite passed. The focused setup suite also passed before rebase and remained byte-for-byte unchanged. Cleanup after SIGKILL, host crash, or for historical temp files remains explicitly out of scope.

### Verdict

LGTM

PR #25155, opened 2026-08-05
fix(ably-subscriber): await graceful shutdown completion

### LINKED ISSUE #25110: bug(rust): make the Ably smoke test await graceful shutdown
## Problem

The Ably real-service smoke test reports `PASS: graceful-close` immediately after calling `Subscription::close()`, even though the call only attempts to notify a detached background event loop. There is no observable completion contract for the caller to distinguish a completed graceful shutdown from a dropped request, a stalled event loop, a close timeout, or a task failure.

This is a shutdown correctness and observability gap, not only a misleading comment in the smoke test. The smoke test cannot prove the behavior it claims to cover while the subscription API exposes only a fire-and-forget request.

## Evidence

- `crates/ably-subscriber/src/subscribe.rs` stores only a oneshot shutdown sender. `Subscription::close()` ignores send failure and returns synchronously, while the spawned event-loop handle is detached.
- `crates/ably-subscriber/src/connection/event_loop.rs` performs the Ably `CLOSE` send and WebSocket close later in the background. Close encoding, send, WebSocket close, and timeout failures are currently ignored or only logged.
- `crates/ably-subscriber/examples/smoke_test.rs` increments the graceful-close pass count immediately after requesting shutdown.
- `crates/ably-subscriber/tests/integration/close_drop.rs` needs a separate mock-server acknowledgement to prove that protocol and WebSocket close occurred, demonstrating that `close()` itself provides no such evidence.

## Decision

Proceed. The smoke test's success signal is a false positive under plausible shutdown failures. Planning must address the missing observable shutdown completion rather than masking it with a delay or weakening the smoke-test assertion.

## Scope and constraints

- Keep `Drop` best-effort and non-blocking; Rust destructors cannot await asynchronous shutdown.
- Preserve deployment compatibility for the runner consumer and update all in-repository call sites if the explicit close contract changes.
- Do not require the real-service smoke test to inspect Ably internals or rely on timing-only sleeps.
- Bound all explicit shutdown waits so a stuck network or background task cannot hang CI or production shutdown indefinitely.
- Preserve the existing protocol ordering: caller-requested shutdown sends Ably `CLOSE` before closing the WebSocket.

## Acceptance criteria

- Explicit subscription shutdown exposes an awaitable completion result that reaches success only after the background event loop completes its caller-requested graceful-close path.
- A failed shutdown request, close timeout, protocol/WebSocket close failure, or background-task failure is observable by the explicit caller.
- The real-service smoke test awaits that result with a bounded timeout and reports failure with a nonzero exit status when graceful shutdown does not complete successfully.
- Integration coverage proves both successful completion and at least one failed or stalled close path cannot be counted as success.
- Best-effort shutdown on `Drop` remains non-blocking and existing runtime consumers retain an intentional shutdown behavior.

## Planning note

The implementation plan should compare an awaitable `close()` contract, a separate completion handle, and test-only observation. It should prefer the smallest coherent public lifecycle contract, not merely the smallest diff.

seven332:
## Summary

- add an awaitable `Subscription::close_and_wait()` lifecycle contract while preserving synchronous `close()` and non-blocking `Drop`
- propagate protocol, WebSocket, timeout, request-delivery, and completion failures to explicit callers
- track subscription-owned transport cleanup, including reconnect cancellation, before acknowledging shutdown
- make the Ably real-service smoke test count graceful close as passed only after tracked completion

## Compatibility

- existing request-only close behavior remains available and the runner keeps using it
- success covers local Ably `CLOSE`, WebSocket close, and outstanding transport cleanup; it does not claim a server `CLOSED` acknowledgement
- no API endpoint, persisted state, migration, or deployment protocol changes

## Validation

- reconnect-close integration test repeated 10 times
- repository pre-commit hooks, including workspace Rust documentation and Clippy

Closes #25110


seven332:
## Code Review: PR #25155

**Reviewed head:** `24a4ca77037c65499bf73c4dc3894e5fedad990d`

### Summary

The additive lifecycle API preserves synchronous `close()` and non-blocking `Drop`, while tracked shutdown now covers direct Ably/WebSocket close work and subscription-owned cleanup created by reconnect cancellation. The smoke test no longer reports success after merely queueing a request.

### Findings

No findings.

### Testing

- Passed `cargo fmt --all -- --check`
- Passed `cargo clippy -p ably-subscriber --all-targets --all-features`
- Passed `cargo doc -p ably-subscriber --all-features --no-deps`
- Passed `cargo test -p ably-subscriber --all-targets --all-features`
- Repeated the reconnect awaitable-close integration case 10 times successfully
- The real Ably smoke test remains CI-only because it requires the managed `ABLY_API_KEY`

### Verdict

LGTM

PR #25156, opened 2026-08-05
fix(ably-subscriber): make graceful shutdown awaitable

### LINKED ISSUE #25110: bug(rust): make the Ably smoke test await graceful shutdown

seven332:
## Summary

- add an additive `Subscription::close_and_wait()` API that reports request, transport, cleanup, and timeout failures
- track subscription-owned WebSocket cleanup so graceful close cannot finish ahead of a cancelled reconnect transport
- make the Ably smoke test await graceful shutdown and add integration coverage for success, terminated event loops, reconnect cancellation, and stalled cleanup

## Compatibility

- keep the existing synchronous `Subscription::close()` and non-blocking `Drop` behavior unchanged
- keep the runner supervisor on request-only close to avoid adding shutdown latency
- retain a local completion contract; this does not wait for an Ably server `CLOSED` acknowledgement

## Validation

- `cargo clippy -p ably-subscriber --all-targets --all-features -- -D warnings`
- repeated the stalled reconnect cleanup integration test 10 times sequentially
- repository pre-commit Rust formatting, documentation, Clippy, and file-size checks

Closes #25110


seven332:
## Code Review: PR #25156

**Reviewed head:** `e2081a72c3b5fb7bdabdd641aacd8feee4e522b4`

### Summary

The additive `close_and_wait()` contract is compatible with existing request-only close paths, propagates bounded shutdown failures, and includes active plus reconnect-owned transport cleanup before success. The smoke test now reports graceful-close success only after that contract completes.

### Findings

No findings.

### Testing

Formatting, Clippy with warnings denied, crate documentation, and all `ably-subscriber` targets/features pass locally. The real-service smoke test remains CI-only because it requires the managed `ABLY_API_KEY`.

### Verdict

LGTM

seven332:
Superseded by #25155, which is the earlier canonical implementation of the approved #25110 plan and now has green CI.

PR #25165, opened 2026-08-05
test(nbd-cow): cover lock replacement during acquisition

### LINKED ISSUE #25148: test(rust): cover NBD lock replacement during acquisition
## Problem and evidence

`crates/nbd-cow/src/device_lock.rs` protects the host-global `/dev/nbdN` claim boundary by reopening a per-index lock file until the flocked file descriptor still matches the inode at the lock path. The successful-flock branch may return `Some` only for the current inode, and the contended branch may return `None` only when the current inode is held.

The existing replacement tests acquire a claim first, replace its path afterward, and call the inode comparison helper directly. They do not replace the path while acquisition is between flock and the decision to return `Some` or `None`.

The issue's original claim that simply deleting either gate leaves the suite green was incomplete. Workspace lints deny warnings, so deleting a gate's only non-test helper use produces a `dead_code` compilation error. However, that proves only that the helper is called, not that its boolean result controls the acquisition decision.

Two semantic mutations were verified independently at repository head `78a98163e3`:

- The successful branch still called `lock_inode_is_current` but ignored its result and returned `Some` unconditionally.
- The contended branch still called `file_inode_is_current` but ignored its result and returned `None` unconditionally.

Under each mutation, all 10 pre-existing `device_lock` tests passed. Deterministic tests that replace the path in the post-flock/pre-validation interval failed for the corresponding branch. The integration coverage gap therefore exists, but its accurate form is that current tests do not prove the helper result gates the acquisition outcome.

## Decision

Proceed with one PR-sized Rust test change. Cover the acquisition loop itself at the pathname-replacement boundary, while preserving the public API and production lock semantics.

Keep `try_acquire_device_claim_in` as the public wrapper around one private acquisition core with a narrow synchronous callback immediately after a relevant flock result and before current-inode validation. Production passes a no-op callback; inline tests use closure-owned state to replace real files under a unique temporary directory.

This is the only ordinary unprivileged way to control the exact interval deterministically without scheduling sleeps or ambient mutable test state. Production and tests continue to use the same open, flock, retry, and metadata logic.

## Alternatives considered

- A `cfg(test)` thread-local hook preserves the public call spelling but adds shared lifecycle and panic-cleanup hazards.
- Inotify cannot stop acquisition in the critical interval; fanotify, FUSE, or tracing/interposition add privileges, platform coupling, and flakiness.
- Relying on `dead_code` lint catches literal deletion but not incorrect use of the comparison result, as the verified mutations demonstrate.

## Risks and mitigations

- Keep the callback private, synchronous, and limited to the two current-inode decision points; do not generalize it into filesystem abstraction.
- Use only closure-owned state and `tempfile` paths; do not add global or thread-local hooks.
- Assert observable acquisition results and final descriptor/path identity.
- Mutation-check semantic misuse that preserves helper calls, not only literal gate deletion.

## Acceptance criteria

- Add deterministic coverage that controls lock-path replacement between flock and current-path validation without relying on scheduling sleeps.
- Cover a successful flock on a stale inode and prove acquisition does not return that stale claim.
- Cover contention on a stale inode and prove acquisition retries against the current path instead of returning `Ok(None)` for the obsolete holder.
- Cover repeated replacement through the acquisition loop and assert the bounded retry exhaustion error.
- Keep tests on real temporary filesystem paths and verify every returned claim matches the current lock inode.
- Verify that preserving each comparison call while ignoring its boolean result makes the corresponding new regression coverage fail, while the pre-existing tests alone remain green.

## Delivery notes

- One independently reviewable PR; no sub-issues or dependencies.
- No public API, schema, persisted data, runner protocol, deployment compatibility, feature switch, or rollout change.

Originally audited at `41681df19b9bb6b2a631c17214c3bc734445457d`; mutation evidence verified at `78a98163e3`.

seven332:
## Summary

- keep NBD device-claim acquisition in one private loop with a narrow post-flock test checkpoint
- cover successful flock recovery after lock-path replacement
- cover stale contention recovery and bounded repeated-replacement exhaustion
- verify returned claims hold the inode currently named by the lock path

## Exclusions

- no public API or production lock-semantics change
- no schema, persisted data, runner protocol, dependency, feature switch, or rollout change

## Validation

- `cargo test --manifest-path crates/Cargo.toml -p nbd-cow device_lock::tests`
- semantic mutation checks: all 10 pre-existing tests passed when each comparison result was independently ignored, while the corresponding new regression test failed
- `cargo test --manifest-path crates/Cargo.toml -p nbd-cow --all-targets --all-features`
- repository pre-commit hooks, including workspace Rust formatting, Clippy, rustdoc, file-size, and commit-message checks

Closes #25148

### DIFF
--- a/crates/nbd-cow/src/device_lock.rs
+++ b/crates/nbd-cow/src/device_lock.rs
@@ -119,16 +119,26 @@ pub fn try_acquire_device_claim_in(
     index: u32,
     lock_dir: &Path,
 ) -> io::Result<Option<NbdDeviceClaim>> {
+    try_acquire_device_claim_in_with(index, lock_dir, |_| {})
+}
+
+fn try_acquire_device_claim_in_with(
+    index: u32,
+    lock_dir: &Path,
+    mut before_current_inode_check: impl FnMut(&Path),
+) -> io::Result<Option<NbdDeviceClaim>> {
     let path = device_lock_path_in(index, lock_dir);
     for _ in 0..MAX_STALE_INODE_RETRIES {
         let file = open_lock_file(&path)?;
         match Flock::lock(file, FlockArg::LockExclusiveNonblock) {
             Ok(lock) => {
+                before_current_inode_check(&path);
                 if lock_inode_is_current(&lock, &path)? {
                     return Ok(Some(NbdDeviceClaim { index, _lock: lock }));
                 }
             }
             Err((file, errno)) if errno == nix::errno::Errno::EWOULDBLOCK => {
+                before_current_inode_check(&path);
                 if file_inode_is_current(&file, &path)? {
                     return Ok(None);
                 }
@@ -260,6 +270,11 @@ mod tests {
 
     use super::*;
 
+    fn replace_lock_file(path: &Path) {
+        std::fs::remove_file(path).expect("remove lock path");
+        drop(create_lock_file(path).expect("recreate lock path"));
+    }
+
     #[test]
     fn second_claim_for_same_index_reports_busy_until_first_drops() {
         let dir = tempfile::tempdir().expect("tempdir");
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #25165

**Reviewed head:** `5aff8c40a7fe3c41b490936e97a736fe0a0863fc`

### Summary

The change adds one private, synchronous post-flock checkpoint shared by the unchanged public acquisition path and uses it to deterministically cover stale successful flock, stale contention, and bounded path-replacement exhaustion on real temporary files.

### Findings

No findings.

### Testing

Targeted device-lock tests and the complete `nbd-cow` all-target/all-feature suite pass, along with Rust formatting, Clippy, rustdoc, and repository pre-commit hooks. Independent semantic mutations showed all 10 pre-existing tests remain green when either comparison result is ignored, while the corresponding new regression test fails. The 12 root/metal-only ignored integration tests were not run locally and remain covered by CI.

### Verdict

LGTM

PR #25186, opened 2026-08-05
test(nbd-cow): bound concurrent claim rendezvous

### LINKED ISSUE #25149: test(rust): bound concurrent NBD claim rendezvous
## Problem and evidence

`concurrent_claims_for_same_index_have_single_winner` in `crates/nbd-cow/src/device_lock.rs` starts eight worker threads together, keeps the winning claim alive until the main test releases it, and waits for every worker result with unbounded `recv()` calls before sending any release signal.

The production acquisition path currently uses `FlockArg::LockExclusiveNonblock`. If it regresses to a blocking flock, the winner reports success and waits for release, the losing workers block behind that lock, and the main test waits for their reports before releasing the winner. The resulting dependency cycle consumes the coverage job's 20-minute timeout instead of producing a focused assertion. The code and workflow still have this structure on `main`; the existing happy-path test passes and therefore does not expose the failure behavior.

## Decision

Proceed with a test-only reliability fix. The issue protects a real concurrency contract and shortens diagnosis of a plausible regression.

The scope is narrower than an arbitrary stuck-worker watchdog. A collection deadline followed by release can recover the blocking-flock regression because releasing the winner allows queued lock attempts to return. Rust threads cannot be safely cancelled, so this change must not claim to reap a worker that remains permanently stuck for an unrelated reason.

## Constraints

- Keep production lock acquisition behavior unchanged.
- Bound the total result-collection phase with one test-local deadline; do not reset the full timeout after each result.
- Perform release and thread joining before evaluating result-count, worker-error, winner-count, or timeout assertions.
- Attempt to join every worker even if another worker panicked, so one failed join does not skip the rest of cleanup.
- Keep failure output phase-specific and include completed versus expected result counts.
- Use an eventually completing delayed-result scenario to exercise the timeout path deterministically without making a regression in the timeout logic hang the whole suite.
- Bound the delayed worker's own release wait and record whether release arrived, so a regression in cleanup also fails locally instead of hanging the coverage job.
- Bound the real winning worker's release wait with a watchdog longer than the collection deadline, so missing cleanup cannot strand the real race and the watchdog cannot release the winner before a collection timeout diagnoses blocking flock behavior.

## Acceptance criteria

- The concurrent claim test fails within its local deadline when not all worker results arrive, and the failure reports how many workers completed.
- Timeout and assertion outcomes release the held claim before they are surfaced.
- Every spawned worker is joined after release before the temporary lock directory is dropped.
- The real winning worker has a bounded release watchdog longer than the collection deadline; missing release becomes a joined worker failure rather than an unbounded wait.
- Deterministic test coverage with an intentionally delayed result proves both the deadline outcome and cleanup/join behavior; the delayed worker eventually completes even if the deadline implementation regresses.
- The delayed worker has an independent bounded release watchdog and reports whether it received release, so a missing cleanup signal cannot strand the test.
- The existing assertions remain: all expected attempts return successfully and exactly one attempt acquires the claim.
- The affected `nbd-cow` formatting, lint, documentation, and test checks pass.

## Out of scope

- Production lock behavior changes.
- General-purpose thread cancellation or subprocess isolation for arbitrary permanently stuck test code.

seven332:
## Summary

- bound concurrent NBD claim result collection with one shared deadline and progress diagnostics
- release and join every claim worker before surfacing test assertions, with an independent winner watchdog
- cover the timeout cleanup path deterministically with a worker that waits for the cleanup release before reporting its result

## Scope

- test-only changes in `nbd-cow`; production claim and locking behavior is unchanged
- does not attempt to cancel arbitrary OS threads blocked outside the test rendezvous

## Validation

- `cargo test --manifest-path crates/Cargo.toml -p nbd-cow --all-features`
- temporarily replacing the nonblocking flock with a blocking flock makes the focused regression fail within the five-second deadline; restoring it makes all `device_lock` tests pass

Closes #25149

### DIFF
--- a/crates/nbd-cow/src/device_lock.rs
+++ b/crates/nbd-cow/src/device_lock.rs
@@ -280,7 +280,7 @@ mod tests {
 
     const CONCURRENT_CLAIM_RESULT_TIMEOUT: Duration = Duration::from_secs(5);
     const CLAIM_WORKER_RELEASE_TIMEOUT: Duration = Duration::from_secs(10);
-    const DELAYED_WORKER_RELEASE_TIMEOUT: Duration = Duration::from_secs(1);
+    const WAITING_WORKER_RELEASE_TIMEOUT: Duration = Duration::from_secs(1);
 
     type ClaimAttemptResult = Result<bool, String>;
 
@@ -489,7 +489,7 @@ mod tests {
     }
 
     #[test]
-    fn claim_result_timeout_releases_and_joins_delayed_worker() {
+    fn claim_result_timeout_releases_and_joins_waiting_worker() {
         let (result_tx, result_rx) = mpsc::channel();
         let (release_tx, release_rx) = mpsc::channel();
         let release_observed = Arc::new(AtomicBool::new(false));
@@ -497,12 +497,13 @@ mod tests {
         let release_observed_by_worker = Arc::clone(&release_observed);
         let worker_completed_by_worker = Arc::clone(&worker_completed);
         let handle = std::thread::spawn(move || {
-            std::thread::sleep(Duration::from_millis(10));
-            result_tx.send(Ok(false)).expect("send delayed result");
             let released = release_rx
-                .recv_timeout(DELAYED_WORKER_RELEASE_TIMEOUT)
+                .recv_timeout(WAITING_WORKER_RELEASE_TIMEOUT)
                 .is_ok();
             release_observed_by_worker.store(released, Ordering::SeqCst);
+            result_tx
+                .send(Ok(false))
+                .expect("send result after release");
             worker_completed_by_worker.store(true, Ordering::SeqCst);
         });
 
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #25186

**Reviewed head:** `b59947c934757ea046ccb128f5e6a3dead0b6a4d`

### Summary

The test-only change replaces the unbounded claim-result rendezvous with one shared deadline, then releases and joins every worker before surfacing assertions. The cleanup regression uses deterministic channel ordering, and production locking remains nonblocking and unchanged.

### Findings

No findings.

### Testing

Formatting, focused Clippy, focused Rustdoc, and the full `nbd-cow` suite pass. All `device_lock` tests also passed 50 consecutive repetitions. A temporary blocking-flock mutation made the focused race test fail within the five-second deadline, confirming that the regression detects the original hang class. Hardware-backed NBD integration tests remain in their pre-existing ignored state and were not run.

### Verdict

LGTM

PR #25097, opened 2026-08-05
fix: stop the thinking indicator from scrolling mid-line

Lunarivibe:
## Problem

The live thinking label under the avatar stutters while it scrolls sideways.

The horizontal motion was implemented as **text content**, not as a transform. Every 100ms tick `displayedSlidingThinkingText` rebuilt the visible string: it appended whole graphemes at the tail and trimmed however many were needed off the head, behind a fixed `...` prefix.

Three things stacked up:

1. **Whole-glyph steps at 10Hz.** `thinkingTypewriterStep(width)` returns `3` for the chat column (~835px), so each tick jumped ~39px at once — 10 discrete lurches per second rather than a scroll.
2. **Uneven steps.** How many head graphemes had to be dropped to fit the `...` prefix depended on the pixel width of the characters being dropped, so the jump distance varied tick to tick.
3. **Hard cut between lines.** `THINKING_TYPEWRITER_LINE_PAUSE_MS = 1000` held at the end of each measured line, then reset straight to the next line's first characters.

As long as displacement is expressed by re-slicing a string, it stays quantised to glyphs and can never be smooth.

## Change

Remove the sliding window entirely and make sure the label never needs to scroll:

- `wrapThinkingTextForWidth` now splits on clause terminators (`，、；：。！？…` always; `,;:.!?` when followed by whitespace) and greedily merges neighbouring clauses while the merged text still fits the label.
- Every emitted chunk is guaranteed to fit the measured width, so `displayedSlidingThinkingText` is gone.
- A finished chunk holds for 700ms, then cross-fades (200ms) into the next one.
- Over-long clauses with no punctuation fall back to the previous per-grapheme width wrapping. If canvas measurement is unavailable, chunks are split by glyph count so a line can still never overflow unbounded.

This also fixes the related complaint that the line frequently started mid-phrase — chunks now begin at clause boundaries.

## Verification

- `pnpm -F @vm0/app run check-types` — pass
- `pnpm -F @vm0/app run lint` — pass
- `pnpm knip --workspace apps/platform` — pass
- `pnpm format` — clean
- `pnpm -F @vm0/app exec vitest run` on `chat-lifecycle`, `chat-workflows-and-goals`, `chat-run-queue` — 76 passed
- Chunker checked against CJK, Latin, punctuation-free, and multi-hard-line inputs: no chunk exceeds the available width, and every chunk's `startIndex`/`endIndex` matches its text.

[...]

PR #25188, opened 2026-08-05
refactor(platform): trust agent source annotations

### LINKED ISSUE #25158: refactor(platform): trust the agent source part instead of re-checking trigger source
## Background

`visibleSourceAnnotationPart` (`turbo/apps/platform/src/views/zero-page/zero-chat-thread-page.tsx:7581`) suppresses an `agent` source annotation unless the event's `triggerSource` also says `agent`:

```ts
const part = userMessageNonContentPart(userMessage);
if (
  part?.type !== "source" ||
  (part.kind === "agent" &&
    (inputEvent?.eventType !== "input.prompt" ||
      inputEvent.triggerSource !== "agent"))
) {
  return undefined;
}
return part;
```

The check cannot fire. Both values are produced from the same signal, in the same request:

```ts
// turbo/apps/api/src/signals/routes/zero-chat-events.ts:244
function normalSendTriggerSource(auth): "web" | "agent" {
  return auth.tokenType === "zero" ? "agent" : "web";
}

// :258  resolveChatAgentRunSource — the only producer of the agent source part
if (auth.tokenType !== "zero") {
  return null;
}
```

A `kind: "agent"` part exists **iff** `tokenType === "zero"`, which is **iff** `triggerSource === "agent"`. The client is re-deriving a server invariant from a second field instead of trusting the document it was given.

It is also inconsistent. The `source` part is a discriminated union of two variants (`turbo/packages/api-contracts/src/contracts/chat-threads.ts:384-410`): six external kinds (`slack`, `feishu`, `teams`, `telegram`, `github`, `agentphone`) and `agent`. **Only `agent` gets a secondary check.** The six external kinds are rendered on the strength of the part alone.

This matters beyond tidiness: `chat_events.trigger_source` is the last field the chat thread reads from the event envelope rather than from `user_message`, and #24202 established `user_message` as the only client-facing surface. Removing this check is the small half of that cleanup. (The other half — `isTemporarilyQueuedMorningBrief` in `signals/chat-page/chat-event-state.ts:99` — is **not** in scope; it needs Morning Brief to gain a part of its own first, which is a product decision.)

## Design

Trust the part:

```ts
function visibleSourceAnnotationPart({
  userMessage,
}: {
  userMessage: UserMessageDocument | undefined;
}) {
  const part = userMessageNonContentPart(userMessage);
  return part?.type === "source" ? part : undefined;
}
```

`inputEvent` becomes unused by this function. Remove the parameter and drop the argument at the call site (`:7674`) **only if** nothing else there needs it; do not delete `inputEvent` from the surrounding component if other code uses it.

### Why trust rather than tighten

The alternative — extend the same check to the other six kinds — was rejected. It would multiply an invariant the server already guarantees across seven branches, and would couple rendering to `trigger_source` exactly when we are trying to decouple it. If the concern is a forged part, the defence belongs at admission, not in the renderer: the client cannot distinguish a forged part from a real one anyway, since it only sees what the API sends.

## The test that encodes today's behaviour

`turbo/apps/platform/src/views/zero-page/__tests__/chat-user-messages.test.tsx`, in `it("renders agent-run source annotations with avatar, link, and run anchor")`, seeds two events:

| Event | `triggerSource` | part | current assertion |
|---|---|---|---|
| `…750` | `"agent"` | `kind: "agent"`, `titleSnapshot: "Source thread"` | link renders |
| `…752` | `"web"` | `kind: "agent"`, `titleSnapshot: "Hidden source"` | `expect(screen.queryByText("Hidden source")).not.toBeInTheDocument()` |

The second event is the combination the server cannot produce, and the assertion is the behaviour this issue deliberately changes.

**Do not keep that assertion passing by keeping the check.** Update the test to match the new contract: either drop the `…752` fixture and its assertion, or keep the fixture and assert the part now renders. Prefer dropping it — a fixture that no producer can emit is a fixture that tests the renderer against an impossible input.

Everything else in that test — the avatar, the `href`, the `run-${targetRunId}` anchor, the non-`bg-muted` image class — must keep passing unchanged.

## Acceptance criteria

- An `input.prompt` carrying a `kind: "agent"` source part renders the agent-run annotation: avatar, link to `/chats/{threadId}#run-{runId}`, and the run anchor. Unchanged from today.
- The six external source kinds render exactly as before.
- `visibleSourceAnnotationPart` no longer reads `triggerSource`, and `grep -n "triggerSource" turbo/apps/platform/src/views/zero-page/zero-chat-thread-page.tsx` returns nothing.
- No server change, no contract change, no migration.
- `signals/chat-page/chat-event-state.ts` is untouched.

## Non-goals

- **Do not touch `isTemporarilyQueuedMorningBrief`** (`signals/chat-page/chat-event-state.ts:99`). It reads `triggerSource === "workflow-schedule"` and cannot be replaced until Morning Brief carries its own part or moves to `input.automation`. Separate decision.
- **Do not remove `triggerSource` from the API contract** (`contracts/chat-threads.ts:490`). Server-side queue dispatch still depends on `chat_events.trigger_source`, and the logs and activity pages read a different `triggerSource` that comes from `zero_runs`.
- **Do not add the secondary check to the other six kinds.**
- No change to `userMessageNonContentPart`, to the source part schemas, or to `SourceMessageAnnotation` / `AgentRunSourceMessageAnnotation`.

lancy:
## Summary

- trust `source` parts directly when deciding whether to render source annotations
- remove the redundant `triggerSource` check and the now-unused helper argument
- delete the impossible `triggerSource: "web"` plus `kind: "agent"` test fixture and its hidden-source assertion

## Why

The API derives both the agent source part and `triggerSource: "agent"` from the same `auth.tokenType === "zero"` condition. Re-checking the envelope field in the renderer duplicated a server invariant and treated agent sources differently from the six external source kinds.

## Merge order

This PR was opened after #25097 merged, so it is based on the post-#25097 `main` branch.

## Validation

- `pnpm -F @vm0/app exec vitest run src/views/zero-page/__tests__/chat-user-messages.test.tsx --maxWorkers=1 --silent=passed-only` (1 file, 5 tests)
- Prettier check for both changed files
- Oxlint and ESLint for both changed files
- `grep -n "triggerSource" turbo/apps/platform/src/views/zero-page/zero-chat-thread-page.tsx` returns no matches

No server, contract, migration, or Morning Brief state changes.

Closes #25158

### DIFF
--- a/turbo/apps/platform/src/views/zero-page/__tests__/chat-user-messages.test.tsx
+++ b/turbo/apps/platform/src/views/zero-page/__tests__/chat-user-messages.test.tsx
@@ -510,29 +510,6 @@ describe("user messages", () => {
           },
           createdAt: "2026-08-04T10:00:00Z",
         },
-        {
-          id: "00000000-0000-4000-8000-000000000752",
-          role: "user",
-          content: "Untrusted prompt",
-          runId: "d0000000-0000-4000-a000-000000000752",
-          triggerSource: "web",
-          userMessage: {
-            version: 1,
-            parts: [
-              { type: "text", text: "Untrusted prompt" },
-              {
-                type: "source",
-                kind: "agent",
-                runId: sourceRunId,
-                threadId: sourceThreadId,
-                agentId: sourceAgentId,
-                titleSnapshot: "Hidden source",
-                href: `/chats/${sourceThreadId}#run-${sourceRunId}`,
-              },
-            ],
-          },
-          createdAt: "2026-08-04T10:01:00Z",
-        },
       ],
     });
 
@@ -567,6 +544,5 @@ describe("user messages", () => {
       "data-role",
       "user",
     );
-    expect(screen.queryByText("Hidden source")).not.toBeInTheDocument();
   });
 });
--- a/turbo/apps/platform/src/views/zero-page/zero-chat-thread-page.tsx
+++ b/turbo/apps/platform/src/views/zero-page/zero-chat-thread-page.tsx
@@ -7588,21 +7588,11 @@ function resolvePagedUserMessageRendering({
 
 function visibleSourceAnnotationPart({
   userMessage,
-  inputEvent,
 }: {
   userMessage: UserMessageDocument | undefined;
-  inputEvent: ChatInputEvent | undefined;
 }) {
   const part = userMessageNonContentPart(userMessage);
-  if (
-    part?.type !== "source" ||
-    (part.kind === "agent" &&
-      (inputEvent?.eventType !== "input.prompt" ||
-        inputEvent.triggerSource !== "agent"))
-  ) {
-    return undefined;
-  }
-  return part;
+  return part?.type === "source" ? part : undefined;
 }
 
 function inputPromptRunAnchor(inputEvent: ChatInputEvent | undefined) {
@@ -7681,7 +7671,6 @@ function PagedUserMessage({
 
   const sourceAnnotationPart = visibleSourceAnnotationPart({
     userMessage,
-    inputEvent,
   });
   return (
     <div


lancy:
## Code Review: PR #25188 (Round 1) — LGTM :tada:

### Summary

The PR removes the redundant client-side `triggerSource` check and trusts the discriminated `source` part consistently across agent and external source kinds. It also removes only the impossible `triggerSource: "web"` plus `kind: "agent"` fixture and its obsolete hidden-source assertion.

### Testing Review

- Coverage: adequate. The existing page-level integration test still exercises the agent source avatar, chat/run link, target run anchor, and transparent-avatar class through `detachedSetupPage`.
- Conventions: compliant. No internal mocks, fake timers, new effects, mutable caches, type suppressions, contracts, migrations, or server changes were introduced.
- Local validation: focused Vitest (1 file, 5 tests), Prettier, Oxlint, ESLint, diff check, and the acceptance grep all passed.

No critical or high-priority issues remaining. This PR is ready for CI and merge.

---
*Completed after 1 round of automated review-fix loop*

PR #25198, opened 2026-08-05
fix(runner): require benchmark guest timezone

### LINKED ISSUE #25178: fix(runner): require benchmark timezone in guest
## Problem and evidence

`runner benchmark --timezone` promises to configure the VM timezone before the benchmark command, but current host validation checks only a non-empty, shell-safe character shape. Guest restoration then wraps timezone setup in `if test -f /usr/share/zoneinfo/<name>` with no failing branch.

As a result, an IANA-shaped typo such as `Mars/Olympus` passes validation. When that file is absent from the selected guest image, restoration succeeds silently and the benchmark runs under the image's existing timezone. Time-sensitive benchmark output can therefore be attributed to a timezone that was never applied.

The authoritative database is the selected guest image's `/usr/share/zoneinfo` tree. Host validation cannot prove availability, and the existing error text incorrectly describes its character whitelist as IANA timezone validation.

## Decision

Treat the benchmark timezone as an explicit required setting. The benchmark command must not start unless the requested guest zoneinfo entry exists and timezone configuration succeeds.

Keep normal agent timezone restoration best-effort. That production behavior has different availability requirements and is not changed by this issue.

## Chosen direction

- Retain the pre-start host character whitelist for shell safety, but name and document it as syntax validation rather than IANA validation.
- Use `zoneinfo name` in CLI help and errors because the runtime contract is existence under the guest image's `/usr/share/zoneinfo` tree, including any aliases supplied by that image.
- Enforce the explicit benchmark zone inside the existing combined guest-state restore exec; do not add a separate guest preflight round trip.
- Fail with a controlled configuration error when the requested zone file is absent.
- Fail through the existing guest-restore diagnostic path when the zone exists but cannot be applied.
- Preserve the existing best-effort fragment and warning behavior for normal agent runs.

## Alternatives considered

- **Document silent fallback:** accurately describes current behavior but preserves incorrect benchmark results and provides no missing-zone signal.
- **Separate guest preflight:** can classify absence directly but adds a round trip, duplicates setup logic, and still must handle application failure.
- **Host timezone catalog:** catches some typos early but cannot authoritatively match the selected guest image and would add dependency/data maintenance.

## Risks and mitigations

- Existing automation passing unavailable custom zones will fail instead of silently falling back. This is the intended correction for an explicit benchmark setting.
- The default `UTC` must exist in the benchmark image. If it does not, setup fails visibly instead of producing measurements with an unverified timezone.
- Strict policy could accidentally leak into normal runs. Represent required and best-effort policies explicitly and retain focused normal-agent tests.
- Shell interpolation remains protected by the existing restricted character set and malformed-input coverage.
- Benchmark startup performance remains unchanged because checking occurs in the existing restore exec.

## Acceptance criteria

- `runner benchmark --timezone <name>` does not execute the benchmark command when `/usr/share/zoneinfo/<name>` is absent in the selected VM.
- A missing zone produces a configuration error that identifies the requested guest zone.
- A timezone application failure fails benchmark setup with captured diagnostics.
- `runner benchmark --help` describes a guest zoneinfo name and states that unavailable zones fail the benchmark.
- Host validation errors describe the actual non-empty shell-safe character requirement and do not claim to validate an IANA timezone name.
- Normal agent guest-state restoration remains best-effort and warning-only for timezone failures.
- Required checking stays inside the existing combined restore exec without an additional sandbox call.
- Focused tests cover help, malformed input, missing zones, application failures, and policy separation; runner formatting, Clippy, rustdoc, and tests pass.

## Delivery notes

This is one independently reviewable runner PR. It changes no API, database schema, persisted state, generated contract, or cross-version protocol and requires no coordinated deployment.

seven332:
## Summary

- require the explicit benchmark timezone to exist and apply successfully inside the guest
- keep normal agent timezone restoration best-effort while sharing the existing restore exec
- describe host validation as shell-safe zoneinfo syntax instead of IANA validation
- document unavailable-zone failure in `runner benchmark --help`

## Behavior

`runner benchmark --timezone <name>` now fails before the benchmark command runs when the selected VM does not contain `/usr/share/zoneinfo/<name>`. A missing zone is reported as a configuration error; a zone that exists but cannot be applied uses the existing guest-restore diagnostic path.

This does not change normal agent timezone behavior, add a guest exec, or modify API, persisted-state, database, or cross-version contracts.

## Validation

- `cargo test -p runner cmd::benchmark::tests`
- `cargo test -p runner executor::tests::guest_state_tests`
- `cargo test -p runner` (2967 passed, 15 ignored)
- pre-commit Rust formatting, Clippy, rustdoc, file-size, and commit-message checks

Closes #25178

### DIFF
--- a/crates/runner/src/cmd/benchmark.rs
+++ b/crates/runner/src/cmd/benchmark.rs
@@ -52,11 +52,11 @@ fn parse_env_args(env: &[String]) -> RunnerResult<Vec<(String, String)>> {
 }
 
 fn validate_timezone_arg(timezone: &str) -> RunnerResult<()> {
-    if !timezone.is_empty() && executor::is_valid_guest_timezone_name(timezone) {
+    if executor::is_shell_safe_guest_timezone_name(timezone) {
         return Ok(());
     }
     Err(RunnerError::Config(format!(
-        "invalid --timezone {timezone:?}: expected a non-empty IANA timezone name"
+        "invalid --timezone {timezone:?}: expected a non-empty guest zoneinfo name containing only ASCII letters, digits, '/', '_', '-', or '+'"
     )))
 }
 
@@ -73,7 +73,7 @@ pub struct BenchmarkArgs {
     /// Environment variables to pass (KEY=VALUE), can be repeated
     #[arg(long, short)]
     env: Vec<String>,
-    /// System timezone to configure in the VM before running the command
+    /// Guest zoneinfo name to configure; benchmark fails if unavailable in the VM
     #[arg(long, default_value = DEFAULT_BENCHMARK_TIMEZONE)]
     timezone: String,
     /// Run the command as root (sudo)
@@ -488,7 +488,23 @@ mod tests {
     use std::sync::atomic::{AtomicUsize, Ordering};
 
     use async_trait::async_trait;
+    use clap::CommandFactory;
     use sandbox::{Sandbox, SandboxError, SandboxInitializationPhase};
+
+    #[test]
+    fn benchmark_help_describes_required_guest_zoneinfo() {
+        let mut command = crate::Cli::command();
+        let help = command
+            .find_subcommand_mut("benchmark")
+            .expect("runner CLI should expose benchmark")
+            .render_help()
+            .to_string();
+        let normalized_help = help.split_whitespace().collect::<Vec<_>>().join(" ");
+
+        assert!(normalized_help.contains(
+            "Guest zoneinfo name to configure; benchmark fails if unavailable in the VM"
+        ));
+    }
 
     #[test]
     fn parse_env_args_accepts_key_value_pairs() {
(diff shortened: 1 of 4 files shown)


seven332:
## Code Review: PR #25198

**Reviewed head:** `05a03de98c46a99c646782b0f2174bbf94426cb7`

### Summary

The benchmark-only explicit timezone path now fails fast against the selected guest's authoritative zoneinfo tree without adding a sandbox call or changing normal agent availability. Host validation and CLI help accurately distinguish shell-safe syntax from guest availability.

### Findings

No findings.

### Testing

Reviewed the Clap help, malformed-input boundary, combined guest restore command, missing-zone classification, application-failure propagation, and normal-agent best-effort isolation. Local formatting, denied-warning Clippy, denied-warning rustdoc, focused tests, the full runner suite (2967 passed, 15 ignored), and repository pre-commit hooks passed. GitHub CI remains to be confirmed for this head.

### Verdict

LGTM

PR #25199, opened 2026-08-05
test(runner): cover timezone file fallback

### LINKED ISSUE #25180: test(rust): assert local timezone file fallback
## Problem and evidence

`detect_system_timezone` in `crates/runner/src/cmd/local/submit.rs:73-84` returns a non-empty `TZ` value and otherwise reads, trims, and filters `/etc/timezone`. Its result is the only value assigned to `JobRequest.user_timezone` during local submission at `crates/runner/src/cmd/local/submit.rs:394-402`; the local provider carries that value into `ExecutionContext` at `crates/runner/src/provider/local/mod.rs:166-188`.

The isolated tests in `crates/runner/src/cmd/local/submit/tests/timezone.rs` prove the non-empty environment branch, but the empty-`TZ` scenario at lines 51-53 only asserts:

```rust
assert_ne!(detect_system_timezone(), Some("".to_string()));
```

`None` satisfies that assertion. Replacing the complete `/etc/timezone` fallback at `submit.rs:80-83` with `None` leaves both current timezone scenarios green: the non-empty `TZ` case still returns its exact value, and the empty case still differs from `Some("")`. The current test also depends on the audit host's real `/etc/timezone`, so it cannot prove trimming, an empty file, or a read failure.

Recent runner coverage is green on commits containing this test, but it exercises only the happy host configuration and does not close this mutation gap.

## Impact

A regression that removes or bypasses the file fallback can silently submit local jobs with `user_timezone: None` even when the host has a valid `/etc/timezone`. The local provider then loses the host timezone before guest-state synchronization and agent environment construction, leaving the sandbox on its image default instead of matching the local user.

## Decision

Add deterministic fallback coverage while preserving the isolated-child boundary established for environment-dependent runner tests. A small internal seam for supplying the environment lookup and timezone-file read is acceptable; production behavior should remain unchanged.

## Constraints

- Do not mutate `TZ` in the parallel parent test process.
- Do not assert the audit or CI host's actual `/etc/timezone` value or require that file to exist.
- Keep non-empty `TZ` precedence and the current trim/filter behavior unchanged.
- Keep the scope to timezone detection and its tests; do not change guest timezone policy.

## Acceptance criteria

- A deterministic test proves that empty and absent `TZ` values return the exact trimmed, non-empty timezone supplied by the file fallback.
- Deterministic tests prove that an empty or whitespace-only file and a file-read failure return `None`.
- A non-empty `TZ` value returns unchanged and takes precedence over the file fallback.
- Removing or bypassing the `/etc/timezone` branch makes the focused test fail.
- The existing ignored-child isolation remains intact, or an equally safe pure-helper boundary replaces only the file/env lookup logic.
- Focused runner formatting, lint, and timezone tests pass.

seven332:
## Summary

- parameterize the private local timezone detector with its fallback file path while keeping the production call fixed to `/etc/timezone`
- replace the host-dependent empty-`TZ` assertion with deterministic ignored-child scenarios backed by real temporary files
- cover non-empty `TZ` precedence, empty and absent `TZ` fallback, trimmed file values, whitespace-only files, and file read failure

## Scope

Production timezone semantics are unchanged. This PR adds no public API, serialized payload, persisted state, feature switch, hidden environment override, guest policy, or deployment change.

## Validation

- `cargo test -p runner detect_system_timezone -- --nocapture`
- `cargo test -p runner --all-targets --all-features` (2966 passed, 15 ignored)
- `lefthook run pre-commit` (workspace Rust format, Clippy, docs, and file-size checks)
- mutation check: temporarily bypassing the fallback read made the exact empty- and absent-`TZ` fallback scenarios fail; restoring it returned the focused suite to green

Closes #25180

### DIFF
--- a/crates/runner/src/cmd/local/submit.rs
+++ b/crates/runner/src/cmd/local/submit.rs
@@ -70,14 +70,14 @@ pub struct SubmitArgs {
     active_inputs: Vec<String>,
 }
 
-/// Detect the system timezone from the `TZ` env var or `/etc/timezone`.
-fn detect_system_timezone() -> Option<String> {
+/// Detect the system timezone from the `TZ` env var or a timezone file.
+fn detect_system_timezone(timezone_file: &Path) -> Option<String> {
     if let Ok(tz) = std::env::var("TZ")
         && !tz.is_empty()
     {
         return Some(tz);
     }
-    std::fs::read_to_string("/etc/timezone")
+    std::fs::read_to_string(timezone_file)
         .ok()
         .map(|s| s.trim().to_string())
         .filter(|s| !s.is_empty())
@@ -398,7 +398,7 @@ impl SubmitPlan {
             vars: None,
             environment,
             secret_environment,
-            user_timezone: detect_system_timezone(),
+            user_timezone: detect_system_timezone(Path::new("/etc/timezone")),
             profile: Some(profile.clone()),
             reuse_key: chat_thread_id.map(|thread_id| format!("thread:{thread_id}")),
             session_id,
--- a/crates/runner/src/cmd/local/submit/tests/timezone.rs
+++ b/crates/runner/src/cmd/local/submit/tests/timezone.rs
@@ -1,3 +1,4 @@
+use std::path::Path;
 use std::time::Duration;
 
 use super::super::detect_system_timezone;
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #25199

**Reviewed head:** `70cb1e1bc6990f02cb2af41836ec8a7c62446bf3`

### Summary

The private path seam preserves the sole production `/etc/timezone` call while allowing deterministic ignored-child coverage with real temporary files. The tests cover environment precedence, empty and absent `TZ`, trimmed fallback values, whitespace-only content, and read failure without mutating the parent process or host `/etc`.

### Findings

No findings.

### Testing

- Focused timezone suite passes with 5 parent scenarios; it also passed 20 consecutive runs on the reviewed head.
- Temporarily bypassing the file fallback made the exact empty- and absent-`TZ` scenarios fail, proving the regression is detected.
- Runner format, affected-crate Clippy/docs, the full runner suite (2966 passed, 15 controlled ignored tests), and staged workspace pre-commit checks passed. The full runner suite preceded a comment-only follow-up commit; focused tests and all pre-commit checks were rerun on this head.
- No product E2E was run because runtime behavior and external contracts are unchanged.

### Verdict

LGTM

PR #25201, opened 2026-08-05
test(sandbox-fc): bound termination rendezvous

### LINKED ISSUE #25181: test(rust): bound control termination rendezvous
## Problem and evidence

The control-server test helper `recv_termination_request` in `crates/sandbox-fc/src/control/server/tests.rs` waits on `kill_rx.recv()` without a local deadline. The source has nine await sites across eight termination protocol and `ProcessTerminationHandle` tests; the full-channel scenario calls the helper twice.

The rendezvous exercises `ProcessTerminationHandle::request_terminate`, whose production path must admit termination, enqueue a monitor request, and await its acknowledgement. If a regression skips or blocks before enqueueing, the parent test waits indefinitely while the server or direct handle retains a live channel sender.

The five-second `send_terminate` deadline does not bound that independent parent receive. The three client/server tests join the client only after monitor delivery, and the five direct handle tests also join their spawned operation after acknowledgement without a local completion deadline. Dropping those ordinary Tokio `JoinHandle`s on assertion failure detaches their tasks.

The unchanged 31 focused control-server tests pass in 0.05 seconds, but a fault path can consume the complete Rust coverage job's 20-minute timeout without identifying whether delivery, acknowledgement completion, or client/server completion stalled.

## Impact

A regression in the sandbox termination handshake turns its diagnostic tests into a nondiagnostic CI timeout, prevents the remaining Rust suite from completing, and can leave detached test tasks alive until the Tokio runtime is destroyed.

## Decision

Proceed with a test-only structured-concurrency fix. Keep production timeout and protocol behavior unchanged.

- Replace the affected spawned client and direct termination tasks with pinned futures owned by the parent test.
- Drive each operation concurrently with monitor delivery through a private `tokio::select!` helper bounded by a one-second local deadline.
- Treat operation completion before monitor delivery as a distinct ordering failure.
- After acknowledgement, drive the same pinned future to completion through a separate one-second deadline with either a monitor-acknowledgement or client/server-specific diagnostic.
- Retain the existing `ControlServerHandle` drop cleanup, which cancels the server, unlinks the Unix socket, and aborts the accept loop during assertion unwinding.

This removes detached operation ownership rather than adding cleanup state around it. One second expires before the existing five-second client deadline, matches synchronization timeouts already used in the module, and has substantial margin over the focused baseline.

## Alternatives considered

- **Abort-on-drop task wrapper:** preserves spawning and literal `is_finished` checks, but adds a bespoke test lifecycle type whose async timeout cleanup and synchronous `Drop` cleanup differ.
- **One-task `JoinSet`:** uses Tokio's abort-on-drop behavior, but models a task collection where every scenario owns one operation and adds nested join states without a corresponding need.
- **Whole-test, client, or production timeout:** cannot independently diagnose every production-path rendezvous; the client timeout does not bound the parent receiver, and a production timeout would change runtime semantics for a test-harness defect. A separate watchdog remains appropriate only inside the deterministic regression test to keep a deliberately broken inner helper bounded.

## Risks and mitigations

- **Scheduling topology changes:** real Tokio channels and Unix sockets still exercise every queue, acknowledgement, and response transition, and the existing lifecycle scenarios retain their behavioral assertions.
- **Pinned future reuse is incorrect:** reborrow `Pin<&mut F>` and retain the full-channel scenario's two receives before completing the same future.
- **Unexpected early completion is hidden:** make operation completion an explicit `select!` branch before monitor delivery.
- **One-second false timeout:** the full focused baseline is 0.05 seconds and the module already uses one-second synchronization deadlines.
- **The deadline regression test hangs:** keep the sender alive without delivering, call the inner helper with a zero deadline, and use a separate two-second watchdog that fails an unbounded inner receive without a competing ready value.
- **Failure leaks resources:** pending client/direct futures are owned and dropped on unwind; server and socket cleanup remains in `ControlServerHandle::Drop`.

## Acceptance criteria

- All nine expected termination-channel receives across eight tests have a one-second local deadline and distinguish timeout, channel closure, and operation completion before delivery.
- All three client/server and five direct post-ack completions have a one-second phase-specific local deadline.
- The full-channel, acknowledgement, coordinator-state, dirty-policy, ready-for-park, stale-run, accepted, already-stopped, and refused-idle assertions remain unchanged in meaning.
- No affected client or direct termination operation is spawned or can detach on timeout or assertion failure.
- A deterministic live-sender, no-delivery scenario proves the inner receive deadline and uses a separate two-second watchdog to remain bounded if deadline logic is removed.
- Timeout and early assertion paths drop pending operations, cancel or abort the server, unlink the socket, and release the fixture directory.
- Production code, public interfaces, dependencies, protocol behavior, and deployment behavior remain unchanged.
- Rust formatting, `sandbox-fc` Clippy, docs, focused control-server tests, and the full `sandbox-fc` test suite pass with all features and no skipped tests.

## Delivery notes

- Implement in one PR modifying only `crates/sandbox-fc/src/control/server/tests.rs` and closing #25181.
- No schema, migration, feature switch, rollout, or compatibility work is required.

seven332:
## Summary

- bound every expected control termination monitor delivery and post-ack completion with one-second, phase-specific test deadlines
- replace spawned client and direct termination tasks with pinned futures so timeout and assertion paths cannot detach background work
- preserve full-channel, acknowledgement, coordinator-state, policy, stale-run, shutdown, and response ordering through structured concurrency
- add a live-sender, no-delivery regression with an independent watchdog so a missing inner deadline fails locally without a delivery race

## Scope

- test-only change in `crates/sandbox-fc/src/control/server/tests.rs`
- no production timeout, termination protocol, API, persistence, schema, dependency, or deployment behavior changes
- no migration, feature switch, or coordinated rollout is required

## Validation

- `cargo clippy --manifest-path crates/Cargo.toml -p sandbox-fc --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc --manifest-path crates/Cargo.toml -p sandbox-fc --all-features --no-deps`
- `cargo test --manifest-path crates/Cargo.toml -p sandbox-fc --all-features control::server::tests::` (32 passed)
- `cargo test --manifest-path crates/Cargo.toml -p sandbox-fc --all-features` (808 unit tests and 1 integration test passed)
- mutation check: removing the inner receive timeout fails the regression test through the independent watchdog in 2.01 seconds

Closes #25181

### DIFF
--- a/crates/sandbox-fc/src/control/server/tests.rs
+++ b/crates/sandbox-fc/src/control/server/tests.rs
@@ -375,17 +375,13 @@ async fn termination_request_receive_has_local_deadline_with_live_sender() {
     let operation = std::future::pending::<()>();
     tokio::pin!(operation);
 
-    let delayed_delivery = async move {
-        tokio::time::sleep(Duration::from_millis(10)).await;
-        kill_tx
-            .send(ProcessTerminationRequest::fire_and_forget())
-            .await
-            .expect("live process monitor channel should accept delayed request");
-    };
-    let (received, ()) = tokio::join!(
+    let received = tokio::time::timeout(
+        Duration::from_secs(2),
         recv_termination_request_with_timeout(&mut kill_rx, operation.as_mut(), Duration::ZERO),
-        delayed_delivery,
-    );
+    )
+    .await
+    .expect("termination request receive exceeded independent regression watchdog");
+    drop(kill_tx);
 
     assert!(
         received.is_err(),
--- a/crates/sandbox-fc/src/control/server/tests.rs
+++ b/crates/sandbox-fc/src/control/server/tests.rs
@@ -2,6 +2,7 @@ use super::*;
 
 use std::future::Future;
 use std::os::unix::fs::PermissionsExt;
+use std::pin::Pin;
 
 use sandbox::ExecTermination;
 use tokio::io::{AsyncReadExt, AsyncWriteExt};
@@ -23,6 +24,8 @@ use crate::park_coordinator::{
 use crate::runtime_dirs::PRIVATE_RUNTIME_SOCKET_MODE;
 
 type GuestState = Arc<tokio::sync::Mutex<Option<Arc<VsockHost>>>>;
+
+const TERMINATION_RENDEZVOUS_TIMEOUT: Duration = Duration::from_secs(1);
 
 struct ControlServerFixture {
     _dir: tempfile::TempDir,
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #25201

**Reviewed head:** `bf66f239a32eb451e2b24a252024d23d6f3f39de`

### Summary

The test-only structured-concurrency change correctly bounds all nine monitor receives and eight post-ack completions while eliminating detached client and direct termination tasks. Full-channel, acknowledgement, coordinator, policy, stale-run, shutdown, and response ordering remain equivalent.

### Findings

No findings. The earlier delayed-delivery timing race was removed on this head; the live-sender, no-delivery regression now has an independent watchdog and no competing successful receive.

### Testing

- Rust formatting, Clippy with warnings denied, and rustdoc with warnings denied passed.
- 32 focused control-server tests passed.
- The full `sandbox-fc` suite passed: 808 unit tests and 1 integration test.
- Removing the inner receive timeout was mutation-tested and failed through the independent watchdog in 2.01 seconds.
- CI status is checked separately after this review.

### Verdict

LGTM

PR #25202, opened 2026-08-05
perf(mitm-addon): bound x billing gzip member input

### LINKED ISSUE #25191: perf(python): bound gzip member input in X billing decoder
## Problem

`crates/runner/mitm-addon/src/billing_body.py:19-36` passes the entire remaining compressed suffix to a new `zlib.decompressobj()` for every gzip member, then assigns `obj.unused_data` back to `remaining`. For a valid gzip file with many small members, each iteration therefore re-submits and copies a decreasing suffix. The raw and decoded 64 KiB caps bound the absolute input size, but they do not keep decompression work linear in that input size.

This decoder runs synchronously from `crates/runner/mitm-addon/src/usage/providers/connectors/x.py:1026-1038` while the mitmproxy response hook computes billing for a successful `POST /2/tweets`, so the repeated work blocks the shared event loop under the GIL.

## Evidence

On `2d0d18c61f1934762d57cccc468bb05d0099c88a` with CPython 3.12.3, a standards-valid 65,522-byte gzip body was constructed from one member containing `{"text":"hello world"}` followed by 3,274 empty members. The decoder returned the exact JSON body, but instrumentation around `decompressobj.decompress()` measured:

- 3,275 decompressor calls
- 107,289,022 cumulative compressed input bytes submitted from 65,522 wire bytes
- approximately 5.0 ms median synchronous decode time across 30 runs

For comparison, a single-member 60,038-byte incompressible gzip body decoded in approximately 0.035 ms median on the same runtime. The many-member shape is therefore about 140 times slower despite comparable wire size.

The shared decoder already avoids this suffix inflation: `crates/runner/mitm-addon/src/body_decoding.py` feeds zlib through `_ZlibInputCursor` in bounded chunks, and `tests/test_body_decoding.py` asserts bounded input and `unused_data` for many empty members. The dedicated billing decoder does not use an equivalent input bound.

This is distinct from #24252. That issue intentionally added valid multi-member gzip support and noted decompressor initialization cost, but described the implementation as linear; the remaining-suffix resubmission makes the current implementation superlinear.

## Impact

A valid multi-member request body accepted by X can force roughly 107 MB of zlib input handling during one 64 KiB billing inspection. Repeated successful requests add several milliseconds of avoidable event-loop latency each and delay unrelated proxy flows in the same runner process.

## Acceptance criteria

- Preserve the current fail-closed billing contract, gzip multi-member support, single-stream deflate behavior, and global raw/decoded size limits.
- Feed gzip members with a fixed compressed-input bound so neither one `decompress()` call nor `unused_data` can contain the full remaining body.
- Add an X billing entry-point regression using a valid JSON member plus enough empty gzip members to approach the raw limit.
- Assert the regression structurally (for example, a bounded maximum decompressor input and bounded `unused_data`) rather than relying on a wall-clock threshold.
- Keep the exact-limit, overflow, incomplete-member, and trailing-input integration cases passing.

seven332:
## Summary

- extract the proven 1 KiB zlib input cursor into a policy-neutral internal module shared by body decoding and billing
- feed X billing gzip members through bounded source chunks and carried tails while preserving the global decoded cap and fail-closed semantics
- cover a near-64 KiB, many-member gzip body through the real X billing entry point with structural input and `unused_data` bounds

## Behavior And Compatibility

- preserves multi-member gzip, single-stream zlib/raw deflate, exact-limit, overflow, incomplete-member, and trailing-input behavior
- keeps billing function signatures, categories, usage events, APIs, queues, and persisted data unchanged
- changes no database, feature switch, deployment workflow, generated artifact, or user-facing documentation
- old and new runners produce the same billing result during rollout; the new runner avoids excess synchronous decompression work

## Structural Result

For the 65,522-byte reproduction body:

- cumulative compressed input submitted to zlib: 107,289,022 bytes to 1,740,302 bytes
- maximum single decompressor input: 65,522 bytes to 1,024 bytes
- maximum `unused_data`: 65,480 bytes to 1,022 bytes
- decoded JSON and X billing category remain unchanged

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/x_connector_usage/test_writes.py` (116 passed)
- `uv run --no-sync python -m pytest tests/test_body_decoding.py` (102 passed)
- `uv run --no-sync python -m pytest tests/` (4,763 passed)
- `git diff --check`

Closes #25191

### DIFF
--- a/crates/runner/mitm-addon/src/billing_body.py
+++ b/crates/runner/mitm-addon/src/billing_body.py
@@ -1,7 +1,6 @@
 """Request body decoding policy for connector billing inspection."""
 
 import zlib
-from typing import Literal
 
 from mitmproxy import http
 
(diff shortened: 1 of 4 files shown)


seven332:
## Code Review: PR #25202

**Reviewed head:** `c9b3e4ef146d1a3a526d21bad8ff4fc2439ed7b7`

### Summary

The PR replaces decreasing full-suffix gzip member reuse with a shared 1 KiB input cursor while keeping streaming, capture, diagnostic, and billing policies separate. X billing retains its direct single-stream deflate path, global decoded cap, exact-limit probe, and fail-closed handling.

### Findings

No findings on the current head. Self-review corrected avoidable deflate chunking and restored the rationale for the 1 KiB resource contract before this review.

### Testing

- uv lock/sync, Ruff format/lint, and basedpyright passed
- 116 X write tests and 102 shared decoder tests passed; 218 focused tests were rerun after the final documentation change
- full mitm-addon suite passed: 4,763 tests
- 6,000 deterministic valid, truncated, trailing, and arbitrary gzip/deflate cases matched the previous output behavior
- the 65,522-byte structural reproduction now limits zlib input to 1,024 bytes and `unused_data` to 1,022 bytes while preserving billing output
- GitHub-hosted CI remains independently tracked by the PR checks

### Verdict

LGTM

PR #25203, opened 2026-08-05
fix(mitm-addon): make json work accounting chunk-independent

### LINKED ISSUE #25192: test(python): cover chunk-independent selective JSON work accounting
## Summary

The selective JSON work-budget test claims chunk-independent accounting, but its escaped-string case does not cover the transition from an escape sequence back to ordinary discarded ASCII. That transition currently consumes a different number of work units depending on how `feed()` chunks the same bytes.

The gap is in `crates/runner/mitm-addon/tests/test_json_selective_work_limit.py:27-44`. The affected parser paths are `crates/runner/mitm-addon/src/usage/json_selective.py:620-699` and the work-unit helpers at `crates/runner/mitm-addon/src/usage/json_selective.py:933-958`.

## Reproduction

With a selected root path so object keys are collected:

```python
payload = b'{"' + (b"k" * 26) + b'":"q\\"s\\\\"}'

extractor = JsonSelectiveExtractor(
    scalar_fields={("target",): ScalarField("int")},
    max_work_units=7,
)
```

Feeding `payload` one or two bytes at a time completes successfully using 7 work units. Feeding the identical payload as one chunk consumes 8 work units; at a limit of 7 it returns `complete=False` with `error="work limit exceeded"`.

The current escaped-string matrix only uses adjacent `\\n` escapes, so it does not exercise an ordinary ASCII byte after an escape. In one large chunk that byte remains in `_consume_string()`'s slow loop; at a chunk boundary it re-enters the discarded-ASCII fast path.

This also reproduces with the production X extractor configured at `65,536` work units (`crates/runner/mitm-addon/src/usage/providers/connectors/x.py:241-255`). A valid 497,976-byte response containing `"data":[{"id":"1"}]`, 13,104 ignored escaped-string properties, and three trailing spaces:

- completes at exactly 65,536 units and extracts `data_count=1` when fed in two-byte chunks;
- returns `work limit exceeded` with no observations when fed as one chunk.

## Impact

The same valid X response can be billed or treated as unparseable solely according to transport chunking. A coalesced response at the budget boundary can therefore lose the extracted usage event and emit an underbilling/visibility signal, while a fragmented response succeeds.

## Acceptance criteria

- Add a regression case covering a collected object key followed by an unselected string that transitions from an escape sequence back to ordinary ASCII.
- Assert the exact required work budget is identical for one-byte, small fixed-size, and whole-payload chunking.
- Assert the exact budget succeeds and one fewer unit fails uniformly for every chunking.
- Add an X-level boundary regression proving that identical response bytes produce the same `data` extraction outcome regardless of chunking.
- Update parser work accounting so fast/slow string-path transitions do not depend on `feed()` boundaries.

seven332:
## Summary

- re-evaluate discarded-ASCII bulk-scan eligibility after slow string state transitions, independent of response chunk boundaries
- preserve existing slow and bulk work tariffs while making bounded extraction deterministic and restoring the unlimited bulk path after escapes
- cover the exact parser budget, the unlimited profiling contract, and the production X 65,536-unit response boundary

## Compatibility

- no API, schema, persisted-data, migration, protocol, or work-limit changes
- JSON validation and extraction results remain unchanged except that valid coalesced inputs now receive the same work accounting as identical fragmented inputs

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest -q tests/test_json_selective_work_limit.py tests/test_json_selective_strings.py tests/test_x_response_parsers.py` (95 passed)
- `uv run --no-sync python -m pytest tests/` (4,765 passed)

Closes #25192

### DIFF
--- a/crates/runner/mitm-addon/src/usage/json_selective.py
+++ b/crates/runner/mitm-addon/src/usage/json_selective.py
@@ -623,13 +623,7 @@ class JsonSelectiveExtractor:
             self._error = "missing string parser state"
             return i
         work_limited = self.max_work_units is not None
-        if (
-            state.raw is None
-            and not state.escape
-            and not state.unicode_remaining
-            and not state.utf8_remaining
-            and _is_discarded_ascii_string_byte(chunk[i])
-        ):
+        if _can_bulk_skip_discarded_string_byte(state, chunk[i]):
             end = self._discarded_ascii_work_span_end(len(chunk), i) if work_limited else len(chunk)
             if self._error:
                 return i
@@ -649,6 +643,8 @@ class JsonSelectiveExtractor:
         start = i
         try:
             while i < end:
+                if _can_bulk_skip_discarded_string_byte(state, chunk[i]):
+                    break
                 b = chunk[i]
                 i += 1
 
@@ -980,6 +976,16 @@ def _is_json_value_start(b: int) -> bool:
 
 def _is_discarded_ascii_string_byte(b: int) -> bool:
     return _JSON_CONTROL_CHAR_MAX <= b < _UTF8_ONE_BYTE_MAX and b not in b'"\\'
+
+
+def _can_bulk_skip_discarded_string_byte(state: _StringState, b: int) -> bool:
+    return (
+        state.raw is None
+        and not state.escape
+        and not state.unicode_remaining
+        and not state.utf8_remaining
+        and _is_discarded_ascii_string_byte(b)
+    )
 
 
 def _validate_extractor_config(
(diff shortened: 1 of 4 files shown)


seven332:
## Code Review: PR #25203

**Reviewed head:** `a98f93d1e33284f60698540d4ea50465b06a62e2`

### Summary

The shared eligibility predicate and re-dispatch restore a byte/state-based boundary between slow validation and bulk discarded-ASCII scanning. Existing tariffs, limits, validation, and APIs remain intact while bounded accounting and unlimited execution no longer depend on transport chunking.

### Findings

No findings.

### Testing

- uv lock/sync, Ruff format/lint, and basedpyright passed.
- 95 focused parser/X integration tests passed.
- Full mitm-addon suite passed: 4,765 tests.
- A deterministic 500-case diagnostic produced identical results and work units across eight chunk sizes.
- GitHub-hosted CI was still pending at review time.

### Verdict

LGTM

PR #25205, opened 2026-08-05
fix(guest-agent): isolate api token from cli descendants

### LINKED ISSUE #25179: security(rust): isolate the guest API token from CLI descendants
## Summary

The runner starts guest-agent as the sandbox user and supplies VM0_API_TOKEN in its process environment. Guest-agent copies that bearer token into owned configuration, but the original bootstrap environment remains attached to the still-running process.

Both CLI launch paths spawn descendants under the same user identity. Their environment curation calls env_clear only on the child command; it does not establish a process-inspection boundary around the parent guest-agent. The audited launch paths do not mark the parent non-dumpable, switch the child to a different UID, or otherwise isolate the parent environment. Under normal Linux procfs access rules, same-UID descendants can therefore observe the bootstrap environment of the guest-agent even though VM0_API_TOKEN is absent from their direct child environment. This violates the explicit contract that runner-owned tokens remain private to guest-agent.

## Evidence

- crates/guest-contracts/src/env.rs:29-33 defines VM0_API_TOKEN as a sensitive runner-owned bearer token that must not be exposed to CLI children.
- crates/runner/src/executor/env.rs:404-414 places the sandbox token in the guest-agent startup environment.
- crates/runner/src/executor/agent_run.rs:2123-2133 starts guest-agent with sudo disabled, and crates/vsock-guest/src/shell_command.rs:78-102 runs production non-sudo commands as the sandbox user.
- crates/guest-agent/src/env.rs:196-208 reads the token from the process environment into GuestConfigRaw. No production guest-agent path removes or protects the original process environment after capture.
- crates/guest-agent/src/cli/child_env.rs:9-12 states that tokens must remain private, while lines 65-73 only clear and rebuild each child environment.
- crates/guest-agent/src/cli/mod.rs:746-828 spawns the normal CLI process without a UID or process-inspection boundary.
- crates/guest-agent/src/cli/codex_app_server.rs:324-369 does the same for the Codex app-server backend.

## Security impact

An untrusted CLI process, or a tool descendant that it launches, can recover the run-scoped sandbox bearer credential from its same-UID parent. It can then invoke any backend capability authorized by that token without going through the intended guest-agent mediation. The impact is bounded by the token scope and lifetime; this report does not assume cross-run authority.

## Acceptance criteria

- Normal CLI and Codex app-server descendants cannot read guest-agent bootstrap credentials through procfs or equivalent same-UID process inspection.
- The guest-agent retains access to the credential needed for its HTTP flows without leaving it in child-observable process state.
- Regression coverage uses a production-equivalent UID and process model to verify that a spawned descendant cannot observe VM0_API_TOKEN from the parent guest-agent.
- Existing curated child-environment behavior and guest-agent heartbeat, checkpoint, completion, and cancellation flows remain functional.

seven332:
## Summary

- make the ordinary Linux guest-agent process non-dumpable before it captures runner-owned bootstrap credentials
- fail startup when the process-inspection boundary cannot be installed, while leaving helper subcommands and non-Linux builds unchanged
- add a real-binary integration test proving that the direct CLI child receives no API token and cannot recover it from `/proc/$PPID/environ`, including the root CI identity path

## Scope and compatibility

- no API, database, persisted-state, runner protocol, rootfs identity, or child-environment contract changes
- CLI children remain independently dumpable after `execve`
- ordinary guest-agent core dumps and unprivileged debugger attachment are disabled on Linux
- privileged `CAP_SYS_PTRACE` inspection is outside this same-UID boundary; this PR does not introduce a service UID or credential-broker protocol

## Validation

- `cargo test -p guest-agent --test api_token_process_isolation`
- `sudo -n target/debug/deps/api_token_process_isolation-10d8ed0a86d438db --exact guest_agent_hides_api_token_from_its_cli_child --nocapture`
- `cargo test -p guest-agent`
- mutation check: replacing the protection syscall with a no-op makes the regression report `parent-token-visible`

Closes #25179

### DIFF
--- a/crates/guest-agent/src/main.rs
+++ b/crates/guest-agent/src/main.rs
@@ -46,7 +46,7 @@ async fn main() {
     if let Some(exit_code) = helper_exit_code_from_args() {
         std::process::exit(exit_code);
     }
-    let runtime = match GuestRuntime::from_process_env() {
+    let runtime = match initialize_guest_runtime() {
         Ok(runtime) => runtime,
         Err(e) => {
             log_error!(LOG_TAG, "Fatal: {e}");
@@ -56,6 +56,26 @@ async fn main() {
     };
     let exit_code = run(runtime).await;
     std::process::exit(exit_code);
+}
+
+fn initialize_guest_runtime() -> Result<GuestRuntime, String> {
+    #[cfg(target_os = "linux")]
+    deny_unprivileged_process_inspection()?;
+    GuestRuntime::from_process_env()
+}
+
+#[cfg(target_os = "linux")]
+fn deny_unprivileged_process_inspection() -> Result<(), String> {
+    // SAFETY: PR_SET_DUMPABLE only changes the calling process's inspection
+    // policy. It must run before the credential-bearing runtime is captured.
+    let result = unsafe { libc::prctl(libc::PR_SET_DUMPABLE, 0) };
+    if result == 0 {
+        return Ok(());
+    }
+    Err(format!(
+        "protect guest-agent from unprivileged process inspection: {}",
+        std::io::Error::last_os_error()
+    ))
 }
 
 fn helper_exit_code_from_args() -> Option<i32> {
--- a/crates/guest-agent/tests/api_token_process_isolation.rs
+++ b/crates/guest-agent/tests/api_token_process_isolation.rs
@@ -1,0 +1,152 @@
+//! The untrusted CLI child must not inspect guest-agent's API credential.
+
+#![cfg(target_os = "linux")]
+
+mod common;
+
+use shell_quote::quote_shell_arg;
+use std::os::unix::fs::{PermissionsExt, chown};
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #25205

**Reviewed head:** `dfff8c1bb3648170cda952c740218a6b6b98c224`

### Summary

The Linux guest-agent now establishes one fail-closed, non-dumpable boundary before capturing the run-scoped API credential or selecting a CLI backend. The real-binary regression covers the direct child environment and parent procfs inspection under both ordinary and root CI harness identities.

### Findings

No findings.

### Testing

Local validation passed for formatting, the targeted process-isolation test, the root-to-unprivileged identity path, the full guest-agent suite, workspace Clippy, and workspace documentation. A no-op mutation reproduced `parent-token-visible`. Non-Linux cross-compilation and remote runner E2E were not run; the new syscall code and regression are Linux-only, while PR CI is tracked separately.

### Verdict

LGTM

PR #25219, opened 2026-08-05
test(runner): cover early lifecycle signals

### LINKED ISSUE #25211: test(rust): cover all early lifecycle signal registrations
## Location

- `crates/runner/src/cmd/start/signals.rs:31-52` registers `SIGTERM`, `SIGINT`, `SIGUSR1`, and `SIGUSR2` before slow startup work.
- `crates/runner/src/cmd/start/signals.rs:248-293` is the only test that delivers a real signal between `EarlySignals::register()` and `SignalController::spawn()`, and it exercises only `SIGUSR1`.
- The larger start-loop tests drive lifecycle methods through test overrides, while `.github/scripts/runner-behavior-drain-resume.sh` delivers drain/resume signals only after the runner is ready.

## Problem

The early-registration contract covers all four lifecycle signals. The module documentation explains that their default Unix disposition is `Term`, and commit `9c3ec4daea` deliberately moved all four registrations ahead of slow startup so they are queued until the handler task is spawned.

The regression suite proves that startup buffering only for `SIGUSR1`. It does not prove that `SIGTERM`, `SIGINT`, or `SIGUSR2` are registered before `spawn()`, nor that each pre-registered stream is wired to the corresponding signal kind. A refactor that keeps `SIGUSR1` early but moves any of the other registrations back into `SignalController::spawn()` would preserve the current unit and post-readiness behavior coverage while reopening the startup window for that signal.

This matters independently of the original drain failure:

- early `SIGTERM` or `SIGINT` would use the default process termination path instead of the runner's hard-stop lifecycle and cleanup;
- early `SIGUSR2` would terminate the runner instead of being consumed by the lifecycle handler;
- the resulting failure would occur before the existing test overrides and ready-runner behavior checks can observe it.

Audited at `ba595c7fd4bc4c25611ac916aaa6fe4191e006f6`. The subsequent `main` changes through `659202c1e847143c090fba0947136ac647fe4ca4` do not touch this signal path.

## Acceptance criteria

- Deterministic tests prove that each of `SIGTERM`, `SIGINT`, `SIGUSR1`, and `SIGUSR2` delivered after `EarlySignals::register()` but before the consumer starts is retained by the matching registered stream.
- Terminating-signal regressions are isolated in child test processes so a missing handler produces a focused assertion instead of killing the Rust test binary.
- Every child/process wait has a local timeout and deterministic teardown; no sleeps are used for signal propagation.
- The existing `SIGUSR1` lifecycle-transition assertion remains covered in addition to the four registration/wiring checks.

seven332:
## Summary

- exercise `SIGTERM`, `SIGINT`, `SIGUSR1`, and `SIGUSR2` delivered after early registration but before the signal consumer starts
- isolate each real-signal scenario in the existing ignored-child process fixture so a missing handler cannot terminate the parent test binary
- verify the production controller outcomes for hard stop, drain, and resume without propagation sleeps

## Scope

This is a test-only change. It does not modify runner signal behavior, APIs, protocols, persistence, migrations, or deployment compatibility.

## Validation

- `cargo test -p runner cmd::start::signals::tests::early_lifecycle_signals_buffer_before_spawn -- --exact --nocapture`
- mutation proof: misregistered `SIGUSR2` failed only its child with `signal: 12 (SIGUSR2)`; restored registration passed
- `cargo test -p runner`

Closes #25211

### DIFF
--- a/crates/runner/src/cmd/start/signals.rs
+++ b/crates/runner/src/cmd/start/signals.rs
@@ -244,52 +244,154 @@ mod tests {
 
     use super::*;
     use crate::ids::RunId;
+    use crate::test_fixtures::ignored_child::{
+        ignored_child_test_env_guard_enabled, run_ignored_child_test,
+    };
 
-    /// Regression test for issue #10416: SIGUSR1 arriving between
-    /// `EarlySignals::register()` and `SignalController::spawn` must still
-    /// drive the mode transition. Previously the handler was registered
-    /// inside the spawned task, so signals delivered during the startup
-    /// window hit the default Term disposition and killed the process.
-    ///
-    /// Timing: raising SIGUSR1 before `spawn()` is deterministic in program
-    /// order, and the outer `timeout(2s)` absorbs the
-    /// `kernel → sigaction → pipe → driver → watch → recv` propagation
-    /// regardless of scheduling — no artificial sleep is needed.
-    ///
-    /// This test raises SIGUSR1 process-wide. It is safe only because no
-    /// other test subscribes to SIGUSR1 — all other tests use
-    /// `SignalSource::Override` and never call `signal()`. If another
-    /// signal-raising test is added, serialize them (e.g. with a
-    /// `Mutex` shared via `OnceLock`) to avoid cross-talk.
+    const EARLY_SIGNAL_CHILD_ENV: &str = "VM0_RUNNER_EARLY_SIGNAL_TEST";
+    const EARLY_SIGTERM_CHILD: &str =
+        "cmd::start::signals::tests::early_sigterm_buffered_before_spawn_child";
+    const EARLY_SIGINT_CHILD: &str =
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #25219

**Reviewed head:** `78ab124dc2722a0313bb06d5c8a38ee5e789fd99`

### Summary

The change replaces the process-global single-signal regression test with four bounded child-process scenarios that exercise real Unix delivery, pre-consumer Tokio buffering, controller dispatch, and the expected hard-stop, drain, and resume outcomes. The `SIGUSR2` setup is deterministic on a current-thread runtime, and every scenario reuses the existing process cleanup fixture.

### Findings

No findings.

### Testing

Focused and full `runner` tests, Rust formatting, runner Clippy, and runner documentation passed locally. A representative broken `SIGUSR2` registration failed only the corresponding child process and was restored before the successful validation run. GitHub-hosted CI is evaluated separately.

### Verdict

LGTM

PR #25224, opened 2026-08-05
fix(runner): secure workspace cache publications

### LINKED ISSUE #25209: security(rust): enforce private permissions for workspace cache artifacts
## Summary

The shared workspace image cache contains complete workspace disk images and resumable session-history data, but several cache publication paths currently inherit filesystem permissions from the ambient umask or source inode instead of establishing a cache-owned permission invariant.

Current runner hosts do not contain untrusted local accounts, and persistent and transient runner services currently run as root. This is therefore defense-in-depth hardening rather than a presently exploitable cross-account isolation failure. Explicit private permissions are still appropriate for this sensitive data because they reduce the impact of future host-account changes, service misconfiguration, or compromise of an unrelated unprivileged host process.

## Confirmed behavior

- The cache is rooted at `/var/lib/vm0-runner/workspace-image-cache`.
- Cache entry directories are created through `fs::create_dir_all` and currently receive ambient modes.
- Workspace images enter the cache through same-filesystem rename or cross-filesystem `cp`; neither path establishes or validates an owner-only destination mode.
- Workspace metadata and session-history metadata are created through `fs::write` and rename without explicit private modes.
- Same-filesystem rename preserves the source inode and its mode.
- `Sandbox::copy_file` already publishes the temporary session-history body as `0600` on Unix, but the workspace cache does not independently validate its owner and mode before publication.
- Existing workspace cache reads validate type, size, identity, and format as applicable, but do not enforce private owner/mode invariants.
- The runner already has descriptor-based host filesystem primitives in `host_file` for no-follow directory traversal, owner validation, and explicit `0700`/`0600` modes.

## Source evidence

- `crates/runner/src/workspace_image_cache/fs.rs:115-153` creates entries and performs sparse copies without a private-mode invariant.
- `crates/runner/src/workspace_image_cache/lifecycle.rs:803-920` promotes active images through rename or copy and validates type and size, but not owner or mode.
- `crates/runner/src/workspace_image_cache/metadata.rs:122-158` reads metadata without an owner requirement and writes it through `fs::write` and rename.
- `crates/runner/src/workspace_image_cache/sidecar.rs:297-343` publishes sidecar metadata and renames the copied body without cache-layer owner/mode validation.
- `crates/sandbox/src/sandbox.rs:402-426` defines the Unix `Sandbox::copy_file` destination as a private `0600` regular file.
- `crates/sandbox-fc/src/workspace_drive_image.rs:36-88` creates the active workspace image with ambient file modes.
- `crates/runner/src/host_file.rs` contains the existing low-level primitives for private host directories and files.
- `crates/runner/src/cmd/service/unit_file.rs:80-103` and `crates/runner/src/cmd/service/mod.rs:373-405` show that the service launch paths do not set a restrictive process-wide umask.

## Decision

Harden newly created or newly published workspace cache state at the cache boundary. Do not add an eager migration, recursive permission scan, or standalone cleanup for legacy cache artifacts.

The write path must establish its own invariant rather than relying on a global systemd `UMask`. This keeps the guarantee identical for direct execution, persistent services, transient services, rename publication, and copy publication without changing permissions for unrelated runner files or child processes.

Legacy entries may retain their current modes until they are naturally evicted or touched by a new publication. When a new publication uses an existing root or entry, tightening the directories and files involved in that publication is expected; no separate legacy-data workflow is required.

## Chosen direction

- Use the existing `host_file` primitives from a workspace-cache policy wrapper rather than using `private_fs`, whose reserved-path policy intentionally excludes shared runner state.
- Establish `0700` on cache directories involved in a new publication, including the cache root and target entry.
- Establish `0600` on every new temporary and final workspace image, workspace metadata file, session-history metadata file, and session-history body.
- Validate that files being published are regular files owned by the runner effective UID. Reject unsafe ownership, file type, or group/other write permissions instead of adopting them.
- Tighten acceptable owner-owned source modes such as `0644` to `0600` before making the cache entry reusable.
- Apply the same checks after both rename and cross-filesystem copy transfers and before final metadata makes an entry reusable.
- Keep cache publication best-effort: a permission-validation failure must not publish the unsafe entry or change the completed run's result.

### Keep ambient permissions

This matches the current host threat model but leaves sensitive cache data dependent on deployment assumptions and provides no defense if the host account model changes.

### Set `UMask=0077` on runner services

This would affect unrelated runner-created files and child processes, would not independently validate renamed source files, and would not cover direct invocations. It is broader than the cache confidentiality requirement while providing a weaker invariant.

### Eagerly migrate or reject every legacy cache entry

This would add startup or maintenance work for data that is an expendable optimization and is not exposed to an untrusted local account in the current deployment. Legacy artifacts are explicitly out of scope and can age out naturally.

## Risks and mitigations

- **Legacy artifacts remain unchanged:** Accepted under the current trusted-host-account model. New publication paths must not perpetuate the old modes.
- **Overlapping runner versions share the cache:** Current runner services run as root, so private root-owned directories remain accessible to old and new overlapping services.
- **Unexpected ownership or file types:** Treat the affected new publication as unusable and clean up its temporary state rather than changing ownership or publishing it.
- **Performance:** Permission enforcement adds only directory/file metadata operations; it must not scan or copy image contents.
- **Process-global umask tests:** Exercise alternate umasks in an isolated child process so parallel Rust tests cannot affect one another.

## Acceptance criteria

- Under a permissive `0022` umask, directories involved in a new cache publication end with mode `0700`, and every newly created or published cache artifact ends with mode `0600`.
- Same-filesystem rename and cross-filesystem copy publication produce the same owner, file-type, and mode guarantees.
- A new publication from an acceptable owner-owned `0644` source is tightened before the entry becomes reusable.
- A new publication rejects unsafe ownership, non-regular files, and group/other-writable files without publishing reusable metadata.
- Session-history temporary bodies, final bodies, and metadata follow the same cache-owned permission invariant.
- Permission-enforcement failures leave the cache entry unpublished and do not change the completed run result.
- No eager migration, recursive permission scan, or standalone cleanup is added for legacy cache artifacts.
- No process-wide systemd `UMask` change is required for either persistent or transient service launch.
- Tests cover permissive-umask creation, rename publication, copy publication, acceptable-mode tightening, and unsafe publication rejection.

seven332:
## Summary

- enforce `0700` on workspace cache roots and publication entries through the existing descriptor-based host filesystem policy
- validate and tighten workspace images, metadata, and session-history sidecars to owner-owned regular `0600` files before publication
- secure sidecar entry directories before guest-copy staging and keep permission failures on the existing best-effort cache path
- cover permissive umask, rename and cross-filesystem copy, sidecar staging/publication, and unsafe source rejection

## Explicit exclusions

- no eager migration, recursive scan, or cleanup of untouched legacy cache artifacts
- no process-wide umask change
- no database schema, persisted cache format, API, or runner protocol change

## Validation

- `cargo test -p runner --bin runner workspace_image_cache`
- `cargo test -p runner --bin runner`

Closes #25209

### DIFF
--- a/crates/runner/src/workspace_image_cache/fs.rs
+++ b/crates/runner/src/workspace_image_cache/fs.rs
@@ -30,6 +30,25 @@ impl WorkspaceImageCache {
             statvfs_bytes(&path).await
         }
     }
+
+    pub(super) async fn ensure_workspace_cache_entry_dir(
+        &self,
+        cache_key: &str,
+    ) -> RunnerResult<()> {
+        crate::host_file::ensure_dir(
+            self.workspace_image_cache_dir(),
+            crate::host_file::DirMode::Private,
+            "workspace image cache root",
+        )?;
+        let entry_dir = self.workspace_image_cache_entry_dir(cache_key);
+        remove_non_directory_workspace_cache_entry(&entry_dir).await?;
+        crate::host_file::ensure_dir(
+            &entry_dir,
+            crate::host_file::DirMode::Private,
+            "workspace image cache entry",
+        )?;
+        Ok(())
+    }
 }
 
 pub(super) fn is_workspace_tmp_path_name(name: &str) -> bool {
@@ -112,17 +131,12 @@ pub(super) async fn remove_non_directory_workspace_cache_entry(path: &Path) -> R
     Ok(true)
 }
 
-pub(super) async fn ensure_workspace_cache_entry_dir(path: &Path) -> RunnerResult<()> {
-    remove_non_directory_workspace_cache_entry(path).await?;
-    fs::create_dir_all(path).await?;
-    let metadata = fs::symlink_metadata(path).await?;
-    if metadata.is_dir() {
-        return Ok(());
-    }
-    Err(RunnerError::Internal(format!(
-        "workspace image cache entry is not a directory: {}",
-        path.display()
-    )))
+pub(super) fn secure_workspace_cache_publication_file(path: &Path) -> RunnerResult<()> {
+    crate::host_file::validate_private_file_destination(
+        path,
+        "workspace image cache publication file",
+    )?;
+    Ok(())
 }
 
 pub(super) fn has_copy_headroom(stats: FsStats, budget: CacheBudget, allocated_bytes: u64) -> bool {
--- a/crates/runner/src/workspace_image_cache/lifecycle.rs
+++ b/crates/runner/src/workspace_image_cache/lifecycle.rs
@@ -20,8 +20,9 @@ use crate::types::{
 
 use super::entry::is_cache_key_name;
 use super::fs::{
-    allocated_bytes, ensure_workspace_cache_entry_dir, has_copy_headroom, local_timestamp,
-    remove_non_directory_workspace_cache_entry, remove_workspace_cache_path_if_exists, sparse_copy,
+    allocated_bytes, has_copy_headroom, local_timestamp,
+    remove_non_directory_workspace_cache_entry, remove_workspace_cache_path_if_exists,
+    secure_workspace_cache_publication_file, sparse_copy,
 };
 use super::metadata::{
     WorkspaceCacheMetadata, WorkspaceCacheState as WorkspaceCacheEntryState,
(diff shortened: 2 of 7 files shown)


seven332:
## Code Review: PR #25224

**Reviewed head:** `b877bf3f66e11e6348e503cd67ea67c2054109dc`

### Summary

The change establishes the intended workspace-cache publication invariant at the cache boundary while preserving existing lock, commit, cleanup, and best-effort behavior. It reuses the descriptor-based host filesystem policy and does not expand into legacy migration or process-wide umask changes.

### Findings

No findings.

### Testing

Real-filesystem coverage exercises permissive umask, same-filesystem rename, cross-filesystem copy, `0644` tightening, group-writable and symlink rejection, sidecar staging/publication, and the high-level nonfatal promotion failure path. Formatting, Clippy, rustdoc, focused cache tests, and the final-head runner suite (`2975 passed; 0 failed; 16 ignored`) passed. A live runner-host smoke test was not run.

### Verdict

LGTM

PR #25229, opened 2026-08-05
feat: roll out the pwa chat keyboard gestures switch to all users

bingjiez666:
## Summary

Full rollout of the `pwaChatKeyboardGestures` feature switch. It was gated to the staff org; it is now globally enabled for everyone.

The switch keeps the PWA chat composer pinned above the software keyboard and enables swipe-to-dismiss gestures. Behavior is still additionally guarded at the call sites by `isStandalonePwa()`, so nothing changes for browser-tab users — only for users running Zero as an installed PWA.

## Changes

- `turbo/packages/core/src/feature-switch.ts` — `enabled: true`, drop the `enabledOrgIdHashes: STAFF_ORG_ID_HASHES` gate.
- `turbo/packages/core/src/__tests__/feature-switch.test.ts` — move the switch out of the staff-org rollout matrix and into the globally-enabled assertions, matching how other fully rolled-out switches (`teamsIntegration`, `customConnectorCliCreate`) are covered.

## Test plan

- CI: `packages/core` feature-switch tests cover the new global state.
- `zero-page/__tests__/chat-lifecycle.test.tsx` parameterizes the switch value, so it is unaffected.

### DIFF
--- a/turbo/packages/core/src/__tests__/feature-switch.test.ts
+++ b/turbo/packages/core/src/__tests__/feature-switch.test.ts
@@ -18,6 +18,9 @@ describe("isFeatureEnabled", () => {
       true,
     );
     expect(isFeatureEnabled(FeatureSwitchKey.TeamsIntegration, {})).toBe(true);
+    expect(isFeatureEnabled(FeatureSwitchKey.PwaChatKeyboardGestures, {})).toBe(
+      true,
+    );
   });
 
   it("should return true for globally enabled switch even with context", () => {
@@ -117,7 +120,6 @@ describe("getAllFeatureStates", () => {
       true,
     );
     expect(staffOrgStates[FeatureSwitchKey.ChatSteer]).toBe(true);
-    expect(staffOrgStates[FeatureSwitchKey.PwaChatKeyboardGestures]).toBe(true);
     expect(staffOrgStates[FeatureSwitchKey.ChatThreadSidebarAutoOpen]).toBe(
       true,
     );
@@ -147,9 +149,6 @@ describe("getAllFeatureStates", () => {
       false,
     );
     expect(otherOrgStates[FeatureSwitchKey.ChatSteer]).toBe(false);
-    expect(otherOrgStates[FeatureSwitchKey.PwaChatKeyboardGestures]).toBe(
-      false,
-    );
     expect(otherOrgStates[FeatureSwitchKey.ChatThreadSidebarAutoOpen]).toBe(
       false,
     );
--- a/turbo/packages/core/src/feature-switch.ts
+++ b/turbo/packages/core/src/feature-switch.ts
@@ -368,8 +368,7 @@ const FEATURE_SWITCHES: Record<FeatureSwitchKey, FeatureSwitch> = {
     maintainer: "bingjie@vm0.ai",
     description:
       "Keep the PWA chat composer pinned above the software keyboard and support swipe-to-dismiss gestures.",
-    enabled: false,
-    enabledOrgIdHashes: STAFF_ORG_ID_HASHES,
+    enabled: true,
   },
   [FeatureSwitchKey.ChatThreadSidebarAutoOpen]: {
     maintainer: "ethan@vm0.ai",


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: f1969d05e2eac545276bc4f8bc8bb023397c0694
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Rolls the `pwaChatKeyboardGestures` feature switch out to everyone: `enabled: false` + `enabledOrgIdHashes: STAFF_ORG_ID_HASHES` becomes `enabled: true` in `turbo/packages/core/src/feature-switch.ts`. The registry test moves the switch out of the staff-org rollout matrix and into the globally-enabled assertions, matching how `teamsIntegration`, `customConnectorCliCreate`, and `customConnectorOAuth2` are already covered. No product code paths, signals, or components changed — the gesture implementation itself shipped earlier and remains additionally guarded by `isStandalonePwa()` at both call sites.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no wrong assumptions or unhandled paths | Pass | `feature-switch.ts:367-372`: `evaluateSwitch` returns `true` on `fs.enabled` before any hash check, so staff orgs keep the same value and all other identities gain it. `getAllFeatureStates` still computes `orgIdHash` because many other switches retain `enabledOrgIdHashes`. |
| `REVIEW.md` | Security — no injection, hardcoded secrets, or missing auth checks | Pass | No endpoint, auth, or credential surface touched. The file header already documents that this system is not an authorization boundary; this switch gates UI gesture behavior only. |
| `REVIEW.md` | Type safety — no `any`, no overly broad types | Pass | Diff only removes a property from an object literal already typed as `FeatureSwitch`; `enabledOrgIdHashes` is optional, so the narrowed literal still satisfies the type. |
| `REVIEW.md` | Style and maintainability — no oversized functions or duplicated logic | Pass | Two-line registry edit; `STAFF_ORG_ID_HASHES` remains imported and used by many other entries, so no dead import is introduced. |
| `REVIEW.md` | Testing coverage — `feat:` commits must have integration tests | Pass | `feature-switch.test.ts:21-23` pins the new globally-enabled state through the public `isFeatureEnabled` entry point. The gesture behavior itself is covered by `chat-lifecycle.test.tsx:397-491` through `detachedSetupPage`. |
| `REVIEW.md` | Feature regression guard — existing flows most likely to regress must stay covered | Pass | The two flows this rollout could regress are already pinned and untouched: `chat-lifecycle.test.tsx:450` (`featureEnabled: true, standalone: false`) asserts mobile-browser chat keeps native scrolling and no `overscroll-contain`, and `:471` (`featureEnabled: false, standalone: true`) still exercises the disabled path via an explicit `featureSwitches` override, which stays valid after the global flip. |
| `REVIEW.md` | New user-facing features must be gated behind a `FeatureSwitchKey` (P1) | N/A | This PR is the rollout of an existing gate, not a new feature. `FeatureSwitchKey.PwaChatKeyboardGestures` is retained, so per-user/org overrides through `POST /api/zero/feature-switches` still work as a kill switch. |
| `REVIEW.md` | DB/JSONB review gate — persisted contract changes need migration, normalizer, or rationale | N/A | No change under `turbo/packages/db/src/schema/**` or `turbo/packages/api-contracts/**`; no JSONB shape is touched. |
| `REVIEW.md` | Event sourcing and optimistic UI semantics | N/A | No persistent or optimistic events, event IDs, projections, or reconciliation paths in the diff. |
| `docs/bad-smell.md` | Zero tolerance for `any` and for lint/type suppression comments | Pass | No `any`, `@ts-expect-error`, `eslint-disable`, `oxlint-disable`, or `prettier-ignore` anywhere in the diff. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | Static imports only; the diff adds no imports at all. |
| `docs/bad-smell.md` | No hardcoded URLs or environment-specific values; no fallback/recovery patterns | Pass | The change removes a gate rather than adding configuration; no URL, env read, or `||` fallback chain is introduced. |
| `docs/bad-smell.md` | Interface changes must be documented and breaking changes highlighted | Pass | `FeatureSwitch`, `getAllFeatureStates`, and `isFeatureEnabled` signatures are unchanged; only registry data changed. The PR body states the behavior change and its blast radius. |
| `docs/testing.md` | Integration tests are primary; mock at the system boundary only | Pass | `feature-switch.test.ts` calls the real registry functions with no mocks at all — no `vi.mock()` of internal paths, no DB or fetch mocking. |
| `docs/testing.md` | Deployment-boundary changes need cross-version coverage | Pass | See the deployment-compatibility row: the response field and the client code that reads it both already exist in deployed builds, so no new cross-version shape needs pinning. |
| `docs/testing/patterns.md` | Pattern 2 — pure-registry tests stay mock-free and assert return values directly | Pass | The added assertion is a direct `expect(isFeatureEnabled(...)).toBe(true)` with no `beforeEach` state or helper scaffolding, matching the existing file style. |
| `docs/testing/anti-patterns.md` | AP-1 — test behavior, not mock call patterns | Pass | The added assertion checks the resolved boolean, not that any function was called; no `toHaveBeenCalled()` in the diff. |
| `docs/testing/anti-patterns.md` | AP-4/AP-6 — never mock internal code or use partial internal mocks | Pass | No `vi.mock()` with a `../` or `../../` path and no `vi.importActual()` in the changed test. |
| `docs/testing/anti-patterns.md` | AP-5 — no fake timers | Pass | No `vi.useFakeTimers()` / `vi.advanceTimersByTime()` in the diff. |
| `docs/deployment-compatibility.md` | Prefer additive changes at cross-version boundaries; no one-shot protocol flips | Pass | `/api/zero/feature-switches` (`zero-feature-switches.ts:33`) serves `getAllFeatureStates`, whose response keys are unchanged. Only the value flips, and every already-loaded frontend build that reads `pwaChatKeyboardGesturesEnabled$` shipped together with the gesture implementation, so an old client receiving `true` runs code it already contains. |
| `docs/deployment-compatibility.md` | Database migrations must be backward compatible with the deployed backend | N/A | No migration or schema change; the rollout state lives in source, not in `user_feature_switches` rows. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Non-blocking notes

- `feature-switch.test.ts`: the removed `otherOrgStates[...]` line previously pinned this key on the `getAllFeatureStates` path that the API actually serves, while the replacement asserts `isFeatureEnabled(key, {})`. This is equivalent protection rather than a gap — without identity context, `enabledOrgIdHashes` can never make `isFeatureEnabled` return `true`, so a silent revert to staff gating would still fail the new assertion, and the globally-enabled branch of `getAllFeatureStates` stays covered by the `Dummy` assertions at lines 89, 98, and 108.
- `views/main.tsx:31`: `isStandalonePwa()` is read inside the `store.watch` callback, so it is evaluated when the switch value settles rather than on later display-mode changes. This is pre-existing behavior shipped with the feature and out of scope for this PR; it only becomes visible to a wider audience now.

### Testing

- **Coverage: Adequate.** The registry change is pinned by `feature-switch.test.ts:21-23`, and the user-visible gesture behavior it unlocks is covered end-to-end through the production bootstrap path in `chat-lifecycle.test.tsx:397` (standalone PWA: `overscroll-contain` on both scroll surfaces, upward swipe locked with focus retained, downward swipe dismissing the keyboard, chat-history swipe passing through, scrollable draft passing through).
- **Regression coverage: Present.** Both untouched-behavior cases remain asserted — non-standalone mobile browser (`:450`) and switch-disabled standalone (`:471`) — so the widened rollout cannot silently change browser-tab users.
- **Conventions: Compliant.** No internal mocks, no fetch stubbing, no fake timers, no direct component rendering; assertions target resolved values and rendered behavior.
- **CI:** all required checks green on `f1969d0`. The earlier `Pull Request Title` and `ci-gate-security` failures were caused solely by the camelCase subject in the original PR title (`.github/scripts/check-pr-title.sh` requires a fully lower-case subject); the title was corrected and both checks now pass with no code change.

PR #25245, opened 2026-08-05
style(chat): use color swatch icon for template reference chips

bingjiez666:
## Summary

Template reference chips rendered the Tabler `IconTemplate` glyph, which does not match the color-swatch icon used by the composer's template picker button. This switches the chip glyph to `IconColorSwatch` so the chip and the control that opens it read as the same thing.

Both rendering paths are covered:

- **Draft state** — `createInlineTemplateNodeView` in `tiptap-workflow-composer.ts` builds the chip as raw SVG inside a ProseMirror node view, so the inlined path data was swapped for the `color-swatch` paths (`@tabler/icons` 3.44.0).
- **Sent messages** — `UserMessageTemplateReference` in `zero-chat-thread-page.tsx` now imports and renders `IconColorSwatch`.

Size (13) and stroke (1.7) are unchanged, so chip metrics and layout are untouched.

## Other `IconTemplate` usages, deliberately untouched

- `tiptap-workflow-composer.ts` `createTemplateAttachmentNodeView` — thumbnail placeholder in the block-level attachment chip. This is the **pre-`structuredPromptInlineTemplates` path**: when the switch is on, `inlineComposerTemplatePicker` returns a picker with `value: undefined`, so `ComposerTemplateAttachmentSync` never sets `runtime.templateAttachment` and the `templateAttachmentGuard` plugin never inserts the block node. The whole node view goes away when the switch reaches 100%, so restyling it has no lasting value.
- `zero-chat-composer.tsx:3923` — error fallback inside `IllustrationTemplateCard` when a preview image fails to load. Not switch-gated (the picker dialog renders either way), but it means "image broken", not "template reference", so a color swatch would be the wrong glyph here.

## Verification

- `pnpm -F @vm0/app run check-types` — passes
- `pnpm -F @vm0/app run lint` — passes (only pre-existing warnings in test files)
- `prettier --write` on both changed files — no diff


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: fc48bedfd83127974daaf9c7b4cd32c3b954d283
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Swaps the template reference chip glyph from Tabler `IconTemplate` to `IconColorSwatch` so the chip matches the composer's template picker button, which already used `IconColorSwatch` (`zero-chat-composer.tsx:5644`). Both render paths are covered: the draft-state ProseMirror node view (`createInlineTemplateNodeView`, raw SVG path data) and the sent-message React chip (`UserMessageTemplateReference`). Purely presentational — no signal, command, lifecycle, API, schema, or event-sourcing surface is touched.

### Findings

#### Critical (P0)
None.

#### High Priority (P1)
None.

#### Observations (P2, non-blocking)

- `tiptap-workflow-composer.ts:873` — the `color-swatch` path data is hand-copied from `@tabler/icons` 3.44.0 rather than imported, so it can drift if the upstream glyph is redrawn. This is unavoidable here: `createComposerIcon` builds a raw `SVGSVGElement` inside a ProseMirror node view where React components cannot be mounted, and the file already inlines Tabler paths the same way for the chat-thread mention and close icons. The added comment naming the source icon is the right mitigation. No change requested.
- Verified the two remaining `IconTemplate` usages are correctly left alone: `createTemplateAttachmentNodeView` is only reachable when `structuredPromptInlineTemplates` is off (`inlineComposerTemplatePicker` returns `value: undefined` when the switch is on, so `runtime.templateAttachment` is never set and `templateAttachmentGuard` never inserts the block node), and `zero-chat-composer.tsx:3923` is a broken-image fallback in `IllustrationTemplateCard`, a different semantic.

### Practice compliance

- `docs/bad-smell.md` — no `any`, no lint/type suppressions, no dynamic imports, no defensive try/catch, no interface changes.
- `docs/effect.md` / `docs/cache.md` / ccstate — no computed, command, Store, effect, ref, or resource-lifecycle code in the diff.
- `docs/deployment-compatibility.md` — not applicable: no request/response, runner protocol, queue payload, schema, or persisted-state change. A mid-rollout mixed frontend version only means a stale tab shows the old glyph.
- Accessibility unchanged: both icons are `aria-hidden`; the chip's accessible name still comes from its `aria-label` / title text.

### Testing

- **Coverage: Adequate.** This is a `style:` commit with no behavior change, so no new test is required. The chip's existing behavior coverage is unchanged and still passes.
- **Verified locally on `fc48bedf`:**
  - `pnpm -F @vm0/app exec vitest run src/views/zero-page/__tests__/chat-user-messages.test.tsx` — 6/6 passed (sent-message chip rendering)
  - `pnpm -F @vm0/app exec vitest run src/views/zero-page/__tests__/chat-composer-template-picker.test.tsx` — 35/35 passed (draft-state inline template chip)
  - `pnpm -F @vm0/app run check-types` — passed
  - `pnpm -F @vm0/app run lint` — passed (only pre-existing warnings in unrelated test files)
- **Conventions: Compliant.** Existing tests assert on `[data-composer-inline-template]` and behavior, not on icon path data, so they remain valid oracles after the glyph change.

PR #25253, opened 2026-08-05
test(guest-common): cover system log stderr fallback

### LINKED ISSUE #25235: test(rust): assert stderr survives system-log append failures
## Problem and evidence

`crates/guest-common/src/log.rs` documents and implements a two-part stderr fallback when appending to the configured guest system log fails:

1. report the append failure as a warning;
2. still emit the original structured log line.

The existing failure-path test creates deterministic invalid filesystem state, but only asserts that the invalid target path was not created. Removing either stderr write leaves that test green. No other repository test observes stderr for this branch.

This matters because an unavailable system log is exactly when stderr is the remaining operational diagnostic channel; losing either record can hide both the sink failure and the original guest error context.

## Decision

Add one dedicated `guest-common` integration test binary that runs the failure scenario in an exact ignored child test and validates the captured child stderr from a normal parent test.

The child will use the public logging API and a real temporary filesystem setup where a regular file occupies the configured log's parent path. The parent will capture stderr in a temporary file and own a bounded lifecycle using a monotonic execution deadline, kill on timeout, and bounded reap observation.

The ignored child is a subprocess fixture, not skipped coverage: the normal parent test always executes it explicitly with `--exact`, `--ignored`, and `--nocapture`.

## Rationale

- A separate process isolates both process-global system-log state and process-global stderr from parallel tests.
- An integration test observes the documented public contract instead of private implementation details.
- Temporary-file capture avoids pipe backpressure while preserving exact stderr bytes.
- No production injection point, new dependency, or platform-specific descriptor redirection is necessary.
- The old path-only inline test can be removed because the new behavior test fully supersedes it.

## Alternatives considered

- **Re-execute an ignored child from the inline library test module:** provides correct process isolation, but couples the child filter to a private module path and mixes subprocess machinery into the internal test suite for public behavior.
- **Redirect stderr inside the parent test process:** rejected because file descriptor 2 is process-global; the logging mutex cannot coordinate unrelated tests, and panic-safe restoration would require unsafe platform-specific code.
- **Add a production writer-injection seam:** rejected because it changes the runtime API solely to test behavior already observable at the process boundary.

## Risks and mitigations

- **Child hang or leak:** bound execution, kill on timeout, and bound reap observation.
- **Exact filter runs no child:** the parent requires the unique guest log records, so empty execution cannot pass.
- **Dynamic error text:** match the stable append-warning prefix, not the operating-system suffix.
- **Dynamic timestamp:** validate its RFC3339 shape separately from the stable level, tag, and message suffix.
- **Incomplete original record:** identify the exact complete structured line and verify an immediate trailing newline in captured bytes.
- **Maintenance overhead:** keep the lifecycle helper local to this single integration test and add no shared abstraction.

## Acceptance criteria

- Force the append failure with deterministic real filesystem state.
- Capture stderr through the isolated child process without mutating parent logging globals or stderr.
- Assert stderr contains a complete append-failure warning line.
- Assert stderr contains exactly one complete original structured line with the expected level, tag, message, valid timestamp, and immediate trailing newline.
- Bound child execution and reap handling on every path.
- Confirm through temporary mutation checks that removing either failure-path stderr write makes the regression fail.
- Remove or fully supersede the existing path-only failure test.
- Pass scoped `guest-common` tests, Rust formatting, Clippy, and rustdoc.

## Delivery notes

This is one independently reviewable, test-only PR. It changes no production behavior, dependency, API, schema, persisted data, protocol, feature switch, generated artifact, or deployment boundary.

seven332:
## Summary

- replace the path-only inline failure test with a public `guest-common` integration regression
- execute the logging scenario in an exact isolated child process and capture its real stderr in a temporary file
- require one append-failure warning and one complete, RFC3339-timestamped, newline-terminated original structured log line
- bound child execution, kill timed-out children, and bound reap observation without adding a dependency

## Scope

This PR changes test coverage only. It does not change production logging behavior, dependencies, APIs, schemas, persisted data, protocols, feature switches, generated artifacts, or deployment behavior.

## Validation

- `cargo test --manifest-path crates/Cargo.toml -p guest-common --all-targets --all-features`
- `cargo clippy --manifest-path crates/Cargo.toml -p guest-common --all-targets --all-features -- -D warnings`
- `cargo doc --manifest-path crates/Cargo.toml -p guest-common --all-features --no-deps`
- repeated the focused regression 50 times
- temporarily removed the append-failure warning write and confirmed the focused regression failed because the warning was missing
- temporarily removed the original structured stderr write and confirmed the focused regression failed because the original record was missing
- repository pre-commit Rust formatting, workspace rustdoc, workspace Clippy, file-size, and commit-message hooks

Closes #25235

### DIFF
--- a/crates/guest-common/src/log.rs
+++ b/crates/guest-common/src/log.rs
@@ -303,27 +303,6 @@ mod tests {
     }
 
     #[test]
-    fn emit_continues_when_system_log_append_fails() {
-        let _guard = LOG_TEST_MUTEX.lock().unwrap();
-        let dir = tempfile::tempdir().unwrap();
-        let parent = dir.path().join("parent-file");
-        std::fs::write(&parent, "not a directory").unwrap();
-        let path = parent.join("system.log");
-        let _system_log = SystemLogFileGuard::set(&path);
-
-        emit(
-            "WARN",
-            "sandbox:guest-agent",
-            format_args!("system log path is not writable"),
-        );
-
-        assert!(
-            !path.exists(),
-            "test setup expected append to fail for missing parent dir",
-        );
-    }
-
-    #[test]
     fn set_system_log_file_switches_paths_and_drops_cached_handle() {
         let _guard = LOG_TEST_MUTEX.lock().unwrap();
         let dir = tempfile::tempdir().unwrap();
--- a/crates/guest-common/tests/log_stderr.rs
+++ b/crates/guest-common/tests/log_stderr.rs
@@ -1,0 +1,174 @@
+use chrono::DateTime;
+use guest_common::log::{clear_system_log_file, emit, set_system_log_file};
+use std::io::{Read as _, Seek as _};
+use std::process::{Child, Command, ExitStatus, Stdio};
+use std::sync::mpsc;
+use std::time::{Duration, Instant};
+
+const CHILD_TEST_NAME: &str = "system_log_append_failure_preserves_stderr_child";
+const CHILD_TIMEOUT: Duration = Duration::from_secs(5);
+const CHILD_REAP_TIMEOUT: Duration = Duration::from_secs(2);
+const CHILD_WAIT_POLL_INTERVAL: Duration = Duration::from_millis(1);
+const APPEND_FAILURE_MARKER: &str = "] [WARN] [sandbox:guest-common] failed to append system log:";
+const ORIGINAL_LINE_MARKER: &str = "] [WARN] [sandbox:guest-agent] system log path is not writable";
+
+#[test]
+fn system_log_append_failure_preserves_stderr() {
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #25253

**Reviewed head:** `d983551d39e77a943fee69bccfa5ff9c8d18dbca`

### Summary

The PR replaces a path-only inline assertion with an isolated integration regression that exercises the public guest logging contract against real filesystem failure state and captures real stderr. The subprocess lifecycle is bounded without adding a dependency or changing production behavior.

### Findings

No findings.

### Testing

The complete `guest-common` suite, Rust formatting, scoped Clippy, scoped rustdoc, 50 focused repetitions, both required negative mutation checks, and repository commit hooks passed. Full-workspace tests were not run locally; broader required CI remained pending at review time.

### Verdict

LGTM

PR #25262, opened 2026-08-05
test(runner): assert config refreshes image mtimes

### LINKED ISSUE #25234: test(rust): assert runner config refreshes image mtimes
## Context

After `runner config` validates and locks every selected image artifact and writes `runner.yaml`, `crates/runner/src/cmd/config.rs:136-143` touches both the rootfs and snapshot directories. Their directory mtimes are the image GC recency signal (`crates/runner/src/cmd/gc/filesystem.rs:113-124`).

This behavior was an explicit acceptance criterion of #19990: successful config generation must refresh the selected artifact mtimes after the config write to reduce the config-to-start GC window.

The success test at `crates/runner/src/cmd/config.rs:375-396` only deserializes the generated YAML and checks the two hashes. Removing either or both `touch_mtime` calls leaves the current test green.

## Impact

A regression can make `runner config` report success while leaving a just-selected rootfs or snapshot with a stale last-used timestamp. That weakens the command's cache-lifecycle contract and can make the artifact eligible sooner wherever it is not protected by the current retained-config inventory.

## Acceptance criteria

- Extend the real-filesystem `run_config_with_home` success coverage to set both artifact directory mtimes to deterministic old values before invoking the command.
- Assert that a successful config write advances the rootfs directory mtime.
- Assert separately that it advances the snapshot directory mtime.
- Avoid sleep-based timing; use sufficiently old fixture timestamps and compare metadata after the call.
- Verify that removing either `touch_mtime` call makes the test fail.

seven332:
## Summary

- set deterministic old mtimes on the complete rootfs and snapshot fixtures in the real `runner config` success test
- assert that successful config generation refreshes both artifact directory mtimes independently
- retain the generated YAML hash assertions in the same command-entry-point scenario

## Scope

This is test-only coverage. It does not change runner production behavior, APIs, persisted configuration, or dependencies.

## Validation

- `cargo test -p runner cmd::config::tests::run_config_writes_config_and_refreshes_image_mtimes -- --exact`
- `cargo test -p runner`
- `git diff --check`
- mutation checks confirmed that removing either the rootfs or snapshot `touch_mtime` call independently fails the corresponding assertion

Closes #25234

### DIFF
--- a/crates/runner/src/cmd/config.rs
+++ b/crates/runner/src/cmd/config.rs
@@ -373,7 +373,10 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn run_config_writes_config_for_complete_snapshot() {
+    async fn run_config_writes_config_and_refreshes_image_mtimes() {
+        use std::fs::FileTimes;
+        use std::time::{Duration, SystemTime};
+
         let dir = tempfile::tempdir().unwrap();
         let home = HomePaths::with_root(dir.path().join("vm0-runner"));
         let args = args_with_valid_image_hashes();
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #25262

**Reviewed head:** `b13b58a565ccf1fcd0a7c553d2334ddf39f2b553`

### Summary

The existing real-filesystem `runner config` success test now verifies that both selected image artifact directory mtimes advance, while retaining the generated YAML assertions. The fixed far-past fixture time avoids sleeps and timestamp-resolution assumptions, and the change remains test-only.

### Findings

No findings.

### Testing

The focused test, full runner test suite, Rust formatting, Clippy, workspace rustdoc, and diff checks passed. Independent mutation checks confirmed that removing either the rootfs or snapshot `touch_mtime` call fails the corresponding assertion. GitHub Actions completion remains unverified at review time.

### Verdict

LGTM

PR #25274, opened 2026-08-05
fix(runner): preserve network policies across refresh failures

### LINKED ISSUE #25254: fix(runner): preserve last-known-good network policies across refresh failures
## Background

Active API-backed runs receive connector network policies from the API. A policy may include temporary user permission grants, and the claim supplies the earliest connector-level `nextRefreshAt` so the runner can recompute the policy when a grant expires. Realtime permission updates can also request an immediate refresh.

The current runner treats loss of policy freshness as a fail-closed event. After a final API error, malformed or incomplete response, queue overflow, invalid refresh deadline, or refreshed-policy registry write error, it attempts to replace the affected connector policy with deny-all. That can disable every permission for a connector because of a transient control-plane or local publication failure.

The product decision is to prioritize active-run continuity: when a trustworthy replacement policy cannot be obtained and published, keep the last-known-good policy unchanged and continue reconciliation attempts. Permission expiration and revocation are therefore eventually consistent during an outage rather than strict fail-closed boundaries.

This decision supersedes #25233, which proposed terminating a run when the deny-all fallback itself could not be persisted.

## Problem and evidence

The current refresh state machine cannot safely implement last-known-good behavior by merely deleting the deny-all call:

- A due scheduled task is removed from `refresh_tasks` before its request is enqueued.
- A final API failure returns without installing another deadline.
- Queue overflow can discard the only known refresh request.
- Invalid, duplicate, or incomplete responses and registry publication failures have no guaranteed later reconciliation after the failed attempt.
- The existing one bounded transport retry from #24312 still ends in deny-all after persistent failure.

Simply preserving the registry without restoring retry state could leave an expired or revoked allow authoritative for the rest of the run.

The registry is shared by multiple active runs and connectors. A failure for one connector cannot require the entire registry file to remain byte-for-byte unchanged while unrelated valid updates proceed. The required invariant is that the affected connector's last successfully published policy remains semantically unchanged until its own valid replacement is published.

## Decision

- Keep the affected active connector's last successfully published policy unchanged whenever the runner cannot obtain and publish its valid replacement. Unrelated connector updates may continue.
- Do not write deny-all and do not terminate the run solely because policy refresh or refreshed-policy publication failed.
- Retain or reinstall per-connector reconciliation state until a valid replacement is successfully published or the run is no longer active. There is no maximum retry count while the run remains active.
- Use a bounded, non-hot retry cadence with jitter so a shared outage does not create synchronized retry traffic across active runs.
- Allow a realtime permission-update notification to bring a pending retry forward without creating duplicate concurrent work.
- Preserve trigger ordering: a newer realtime reconciliation trigger must not be cleared, delayed, or superseded by completion of an older queued or in-flight refresh.
- Apply retry jitter across active runs without de-coalescing connectors from the same run and retry attempt.
- After a successful publication, replace retry state with the API-provided `nextRefreshAt`; clear scheduling when it is `null`.
- Keep stale ownership results (`Ok(false)`), explicit terminal-run responses, unregister, and runner shutdown distinct from refresh failures. They must retain their existing ownership and cleanup semantics rather than scheduling retries for inactive work.

Treat refresh delivery as a level-triggered reconciliation responsibility rather than a one-shot edge. Every active connector that still needs reconciliation must be represented by either queued/in-flight work or one future retry deadline. A failed attempt preserves the current enforcement state and restores that responsibility; a successful publication replaces it with the API's next deadline.

Carry a lightweight per-connector reconciliation generation through scheduled and queued work. A realtime trigger advances the generation. Results from an older generation may not replace the newer generation's pending task or retry responsibility. This closes the notification-during-refresh race without introducing a full queued/in-flight state machine.

This direction matches the availability-first product decision without silently turning a transient failure into permanent staleness. It remains runner-local because the API already recomputes the canonical policy and returns the next deadline.

### Keep the current deny-all fallback

Rejected because a transient API, queue, response, or local publication failure can disable every permission for an otherwise healthy connector and disrupt the active run.

### Preserve the policy but stop after a fixed number of retries

Rejected because there is no permissible terminal action after the retry budget: deny-all and run termination are explicitly rejected, while stopping would leave the old policy permanently stale. Retry therefore continues at a bounded cadence until success or run teardown.

### Enforce permission-level expiration locally

This could provide strict local expiry without an API round trip, but the current flattened `NetworkPolicy` does not carry permission-specific expiry or baseline fallback information. It would require a cross-version runner/addon policy protocol change and is not required for the accepted eventual-consistency behavior.

## Accepted trade-off

An allow that has expired or was explicitly revoked may remain effective while the API or local registry publication remains unavailable. If the failure persists until the run ends, the old policy can remain effective for the rest of that run. This is an intentional availability-over-strict-expiration decision, not a fail-closed security guarantee.

## Risks and mitigations

- **Retry amplification during a shared outage:** use capped backoff, jitter across runs, batching, and per-run/connector deduplication; connectors from the same run and retry attempt share a jitter window so they remain coalescible.
- **Newer trigger lost behind an older result:** carry and compare connector reconciliation generations before replacing scheduling state.
- **Permanent staleness after lost work:** maintain the invariant that every active connector needing reconciliation has queued/in-flight work or a future retry.
- **Source-IP reuse races:** retain current source-IP, run-ID, and connector-slug ownership checks; stale `Ok(false)` outcomes stop rather than retry against replacement ownership.
- **Shutdown stalls:** cancellation must continue to abort scheduled tasks and prevent new retry publication during shutdown.
- **Concurrent unrelated updates:** preserve the affected connector policy semantically rather than asserting whole-file byte identity.

## Acceptance criteria

- Persistent API transport and HTTP failures leave the affected active connector policies unchanged.
- Invalid, duplicate, or incomplete refresh responses leave the prior affected policies unchanged and retain future reconciliation.
- A refreshed-policy registry read or atomic-publication failure leaves the affected prior policies authoritative and retains future reconciliation.
- Scheduled refresh queue saturation cannot leave an active connector with neither queued/in-flight work nor a future retry.
- Retry scheduling is bounded, jittered, deduplicated per active run/connector, has no maximum attempt count while the run remains active, and is cancelled when the run unregisters.
- A later successful retry publishes the current API policy and adopts its returned `nextRefreshAt`.
- A realtime notification can accelerate reconciliation while a delayed retry is pending.
- A realtime notification that arrives while an older refresh is queued or in flight remains pending; the older result cannot clear or delay that newer reconciliation generation.
- Retry jitter spreads different runs while connectors from the same run and retry attempt remain eligible for existing deadline coalescing.
- Explicit terminal-run responses and stale source-IP/run-ID/connector ownership checks do not retry or mutate a later run that reused the source IP.
- Logs distinguish policy retention with a scheduled retry from successful publication, stale work, terminal reconciliation, and shutdown.
- Integration coverage proves a scheduled temporary-grant expiry can fail repeatedly, retain the exact previous affected policy, retry, and later converge after recovery.
- Existing tests that require deny-all after refresh failure are updated to assert the new last-known-good contract; tests must not be skipped or relaxed through longer timeouts.

## Delivery notes

This is one independently reviewable runner PR centered on:

- `crates/runner/src/provider/network_policy_refresh.rs`
- focused proxy-registry behavior exercised through the refresh entry point
- module safety-contract documentation and integration tests

No API contract, database schema, persisted-data migration, proxy-registry schema, Python addon protocol, feature flag, or deployment compatibility bridge is required.

## Out of scope

- Enforcing permission-level expiration locally inside the mitm addon
- Adding permission-specific expiration metadata to `NetworkPolicy`
- Providing a strict maximum staleness guarantee during a persistent outage
- Changing explicit terminal-run cleanup behavior
- Cancelling or terminating a run because a refresh failed

## Related work

- #24312 added one bounded retry for transient transport failures before the current deny-all fallback.
- #25233 identified the secondary failure when deny-all publication fails; it is superseded by this last-known-good policy decision.

seven332:
## Summary

- preserve each connector's last-known-good network policy when refresh delivery, response validation, or registry publication fails
- retry failed refreshes indefinitely with capped exponential backoff and deterministic jitter while the run remains active
- carry per-connector reconciliation generations through scheduling and queueing so older work cannot clear newer realtime refresh requests
- keep explicit terminal-run reconciliation fail-closed

## Security trade-off

During a persistent control-plane outage, a previously allowed destination can remain allowed longer than its intended expiry. This is the continuity behavior approved in #25254; retries continue until a refresh succeeds or the run becomes inactive.

## Exclusions

- no API or runner protocol changes
- no database schema or persisted-data changes
- no proxy registry schema changes
- no changes to terminal-run fail-closed behavior

## Validation

- `cargo test --manifest-path crates/Cargo.toml -p runner --quiet`

Closes #25254

### DIFF
--- a/crates/runner/src/provider/network_policy_refresh.rs
+++ b/crates/runner/src/provider/network_policy_refresh.rs
@@ -7,33 +7,36 @@
 //! `nextRefreshAt` deadlines returned with the claim.
 //!
 //! Refresh work comes from two sources: realtime/API notifications and
-//! scheduled deadlines. Notifications are accepted only for currently active
-//! run/connector pairs. Scheduled refreshes are tracked per connector within one
-//! run, replacing older tasks for the same connector and coalescing nearby due
-//! connectors for that same run before enqueueing work. The worker drains a
-//! bounded queue, deduplicates active connector slugs, splits API requests by
-//! `NETWORK_POLICY_REFRESH_CONNECTOR_SLUGS_MAX`, validates each API response
-//! against the requested active connector set, patches the proxy registry, and
-//! replaces the next schedule from the returned `nextRefreshAt`.
+//! scheduled deadlines. Both use one per-connector scheduler. Realtime
+//! notifications advance a reconciliation generation and schedule it
+//! immediately; API deadlines and retries continue the generation they belong
+//! to. Scheduled refreshes replace older tasks for the same connector and
+//! coalesce nearby due connectors for that same run before enqueueing work. The
+//! worker drains a bounded queue, ignores superseded generations, splits API
+//! requests by `NETWORK_POLICY_REFRESH_CONNECTOR_SLUGS_MAX`, validates each API
+//! response, patches the proxy registry, and replaces the next schedule from
+//! the returned `nextRefreshAt` only when the generation remains current.
 //!
 //! A typed terminal-run response removes the entire run from refresh tracking,
 //! cancels its scheduled work, and fail-closes every still-matching connector
 //! policy. A first transport failure is retried once while the current policy
-//! remains installed. Any final API failure retains the connector-scoped
-//! fail-closed path.
+//! remains installed.
 //!
-//! The safety contract is to trigger fail-closed patching when freshness cannot
-//! be established for an active connector. Queue overflow, final API refresh
-//! failure, omitted requested connectors, duplicate requested connectors,
-//! malformed `nextRefreshAt`, and registry patch errors trigger fail-closed
-//! patching for the affected connector when it is still active. Registry writes
-//! always re-check source IP, run id, and connector slug so a stale refresh task
-//! cannot patch a later run that reused the same source IP.
+//! The refresh-failure contract prioritizes active-run continuity. Queue
+//! overflow, final API refresh failure, omitted or duplicate requested
+//! connectors, malformed `nextRefreshAt`, and registry patch errors retain the
+//! affected connector's last successfully published policy and reinstall a
+//! capped, jittered retry. Retries continue while the run and reconciliation
+//! generation remain active. An older queued or in-flight result cannot clear a
+//! newer realtime trigger. Registry writes always re-check source IP, run id,
+//! and connector slug so stale work cannot patch a later run that reused the
+//! same source IP.
 //!
-//! Not every stale or shutdown path fails closed. Inactive runs and unknown
-//! connector slugs are ignored. Extra unrequested response connectors are ignored.
-//! A closed refresh queue during shutdown is logged and stopped. A valid but
-//! already-expired deadline is retried after
+//! Explicit terminal-run reconciliation remains fail-closed. Inactive runs,
+//! superseded generations, and unknown connector slugs are ignored. Extra
+//! unrequested response connectors are ignored. A closed refresh queue during
+//! shutdown is logged and stopped. A valid but already-expired deadline is
+//! retried after
 //! `EXPIRED_REFRESH_DEADLINE_RETRY_DELAY`. A registry patch result of
 //! `Ok(false)` means the registry no longer matches the active run/connector,
 //! so refresh tracking for that run is stopped instead of force-patching stale
@@ -45,6 +48,7 @@
 //! enqueueing stale registry refreshes.
 //!
 use std::collections::{HashMap, HashSet};
+use std::hash::{DefaultHasher, Hash, Hasher};
 use std::sync::{Arc, Mutex as StdMutex};
 use std::time::Duration;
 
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #25274

**Reviewed head:** `938e4ef604ad08cc92f0b7172d20f5f9a74427f3`

### Summary

The refresh coordinator now retains connector-scoped last-known-good policies across transient reconciliation failures, restores bounded retry responsibility, and uses per-connector generations to preserve newer realtime triggers. Explicit terminal reconciliation remains fail-closed.

### Findings

No findings.

### Testing

Reviewed coverage for queue saturation, transport and HTTP failure recovery, malformed/duplicate/omitted responses, registry publication errors, retry backoff and coalescing, generation ordering, stale ownership, unregister/shutdown, and terminal fail-close behavior. Formatting, Clippy, rustdoc, and the complete runner test suite passed on this head.

The accepted residual risk is documented: a stale allow can remain effective during a persistent control-plane or local publication outage.

### Verdict

LGTM

PR #25294, opened 2026-08-05
refactor(db): add chat input source discriminators

lancy:
## Summary

Part 2a of #25284 (step 2 of #25260).

- relax `chat_events_context_pair_check` and add only `web` to `chat_events_context_type_check`
- write the table-less `web` discriminator at exactly the two web `input.prompt` insert sites and restore the existing `goal` discriminator in `admitGoalQueueEvent`
- preserve a nullable `context_id` when claimed replacements inherit `context_type`
- backfill historical queued inputs in 10,000-row `ORDER BY id` / `FOR UPDATE SKIP LOCKED` batches with a commit after every batch
- keep the append-only trigger installed, narrowly permit only the computed discriminator update, restore strict rejection before the final assertions, and leave dispatch plus `trigger_source` writes unchanged

## Context pointer write design

`NewChatEvent` carries `contextType: "web"` or `contextType: "goal"` directly for the two sources that intentionally have no context row. Existing sources still derive both pointer columns from `newDisplayContext`. This keeps table-backed context creation unchanged and makes the TypeScript write union exclude `input.rejected` from the new discriminator writes.

## Backfill

The migration implements the issue's priority order: existing value, goal event, own trigger source, own run source, revoker run source, then the measured web fallback. Unsupported non-null trigger sources fail before append-only protection is narrowed.

The fallback comment records that 2,613 `control.revoke` revokers are provably web while 52 `input.rejected` revokers have unrecoverable provenance and receive `web` only as a non-dispatching placeholder.

The final assertion verifies every `input.prompt` / `input.automation` / `input.goal` row is classified. It also verifies rule-5 Slack rows are `slack` and emits their post-backfill count.

## Rollout ledger

| Phase | Production release time (UTC) | Observation window | Gate |
| --- | --- | --- | --- |
| PR 1: relax/write/backfill | Migration completed 2026-08-05 12:43:27Z; API promotion completed 2026-08-05 12:43:49Z ([release run](https://github.com/vm0-ai/vm0/actions/runs/31006153528)) | 2026-08-05 12:43:49Z–12:59:32Z; two read-only covered-NULL checks more than 10 minutes apart both returned 0 | Complete |
[...]

Issue #25260, opened 2026-08-05
epic: remove chat_events.trigger_source


lancy:
## Goal

Remove `chat_events.trigger_source`.

Every source added to the chat event mechanism today has to be registered in two enums that mean almost the same thing — `trigger_source` (13 values, shared with `zero_runs`) and `context_type` (10 values, chat-event only) — plus `event_type`. Dropping one of them shrinks what has to stay consistent every time a channel is added or changed. That maintenance surface is the reason for this epic; the storage saving is incidental.

## Why the column is removable

`chat_events.trigger_source` is a copy in both of the regimes it is read in, and the two regimes are disjoint.

**Claimed rows** — `run_id` is set, and `zero_runs.trigger_source` is `varchar(20) NOT NULL`. Measured: of the 294 Slack/Feishu events written before #23451 with a null `trigger_source`, all 147 claimed ones resolve correctly through `run_id` (130 slack, 17 feishu, matching their `context_type` exactly). The read path already relies on this — `zero-chat-thread.service.ts:302` is `COALESCE(chat_events.trigger_source, zero_runs.trigger_source)`.

**Pending rows** — no run exists yet, but `event_type` has not yet lost the input kind. The collapse to `input.prompt` happens at claim (`zero-chat-queued-event.service.ts:404/461/523` all write `eventType: "input.prompt"` for the replacement), which is why production shows **zero** claimed `input.automation` (0 of 1,010) and **zero** claimed `input.goal` (0 of 3,292). So for a pending row, `(event_type, context_type)` determines the source:

| `event_type` | `context_type` | source |
|---|---|---|
| `input.automation` | `automation` | `workflow-event` / `workflow-schedule` |
| `input.goal` | — | goal continuation |
| `input.prompt` | `slack` / `feishu` / `teams` / `telegram` / `github` / `agentphone` / `morning_brief` / `agent_run` | that channel |
| `input.prompt` | **NULL** | **web** — the one inference, see below |

Every production read of `chat_events.trigger_source` operates on pending rows, with a single exception: the contract projection at `zero-chat-thread.service.ts:302`, which since #25158 and #25160 has **no client consumers left**.

## The one real obstacle

[...]

PR #25305, opened 2026-08-05
refactor(db): require queued input source discriminators

### LINKED ISSUE #25284: refactor(db): give every queued input event a source discriminator
## Scope

**Step 2 of #25260.** Give every queued input event an explicit source discriminator, so step 3 can dispatch on `context_type` alone instead of inferring `context_type IS NULL ⇒ web`.

`chat_events.trigger_source` is still written and still drives dispatch after this step. Nothing about routing changes here.

## Design

`context_type` stops meaning "which context table" and becomes **"which input source"**. `context_id` stays optional and holds the context row only when one exists. Two new discriminator values have no context row at all:

- **`'web'`** — the user typed it; there is no external trigger to store
- **`'goal'`** — goal continuations lost their table in #24932 and need the discriminator back, not the table

Record the meaning change in the schema comment on `context_type`. The column name outlives it, and the next reader will otherwise ask why `web` and `goal` have a type but no row.

### PR 1 — relax, write, backfill

**Migration:**

- add `'web'` to `chat_events_context_type_check`. `'goal'` is **already** an allowed value — it survived #24932, which removed the table but not the enum entry. Only the writer has to start using it again.
- replace `chat_events_context_pair_check`, currently `(context_type IS NULL) = (context_id IS NULL)`, with:

```sql
CHECK (context_id IS NULL OR context_type IS NOT NULL)
```

Optionally also pin that the two table-less values never carry a row:

```sql
AND (context_type NOT IN ('web', 'goal') OR context_id IS NULL)
```

Both are relaxations relative to today, so they are safe to ship alongside the code that starts writing the new values.

**Writers** — start emitting the discriminator:

- web sends in `routes/zero-chat-events.ts` — **exactly two `input.prompt` insert sites**, `:1495` and `:1616` as of `59fcbb8`. Confirm by `eventType`, not by line number, which drifts.
- `admitGoalQueueEvent` in `chat-goal-queue.service.ts:90`

The other `insertChatEvent` calls in that file are **not** targets: `:1863` is `control.interrupt`, `:2700` and `:2800` are `output.error`, and `:2779` is `input.rejected` — which this step deliberately leaves without a discriminator.

Both write sites carry `triggerSource: params.triggerSource`, sourced from `normalSendTriggerSource(auth)`, so they can be `web` or `agent`. Only `web` needs the new value: `agent` sends already get `context_type = 'agent_run'` from `newDisplayContext`, which handles `agent_run` and `slack` and nothing else.

`displayContextPointer` only produces a pointer from a `NewDisplayContext`, which these two paths do not have. Adding a way to set `contextType` without a context row is the substance of this PR; pick an approach and say why in the PR description.

**Backfill**, in priority order:

| # | Condition | `context_type` | rows |
|---|---|---|---|
| 1 | already set | unchanged | ~2,700 |
| 2 | `event_type = 'input.goal'` | `'goal'` | 3,308 |
| 3 | own `trigger_source` set | mapped | 19,005 |
| 4 | `run_id` → `zero_runs.trigger_source` | mapped | 90,015 |
| 5 | revoked → **the revoking row's** `run_id` → `zero_runs.trigger_source` | mapped | 9,960 |
| 6 | none of the above | `'web'` | 2,665 |
| — | `input.rejected` | **not backfilled** | 114,656 |

Mapping for rules 3–5: same-named for `slack` / `feishu` / `teams` / `telegram` / `github` / `agentphone`; `agent` → `agent_run`; `workflow-event` → `automation`; `workflow-schedule` → `automation` when `event_type = 'input.automation'`, otherwise `'morning_brief'`.

### PR 2 — tighten

Only after PR 1 has been deployed and observed:

```sql
ALTER TABLE chat_events ADD CONSTRAINT chat_events_input_context_type_check
  CHECK (event_type NOT IN ('input.prompt', 'input.automation', 'input.goal')
         OR context_type IS NOT NULL);
```

Shipping this with PR 1 would apply the constraint **before** the new writers go live — migrations run ahead of the API promotion, so the still-running previous version would insert violating rows for the length of the window.

## The trap

`replacementContext` (`zero-chat-event.service.ts:706`) inherits the context pointer onto a claimed replacement only when **both** columns are set:

```ts
if (target.contextType !== null && target.contextId !== null) {
  return { pointer: { contextType: target.contextType, contextId: target.contextId }, displayContext: undefined };
}
```

A `'web'` or `'goal'` row has `context_id IS NULL`, so it falls through to `newDisplayContext(...)`, which returns nothing for those sources — **the replacement silently loses the discriminator**, and PR 2's CHECK then fails on it.

The condition must become `target.contextType !== null`, carrying `contextId` through as whatever it is, including null. Cover it with a test that claims a web message and asserts the replacement still has `context_type = 'web'`.

## Rules 5 and 6 are measured, not assumed

**Rule 5 is worth 9,960 rows and is easy to miss.** A revoked original has no `run_id` of its own, but the row that revokes it does. All 12,625 three-way-null rows have a revoker; 9,960 of those revokers carry a `run_id`, and every one resolves in `zero_runs` — 9,939 `web` and **21 `slack`**. Skipping rule 5 and letting them fall to rule 6 would mislabel those 21.

**Rule 6 is 2,665 rows and is not a guess.** The three revoke paths that can retire an `input.prompt` are all chat send routes — `handleRecallSend$`, `createNormalChatRun$`, `sendQueueFirstNormalEvent$` — whose trigger source is `normalSendTriggerSource(auth)`, i.e. web or agent. Agent sends only became possible on 2026-08-04; these rows span 2026-05-09 → 2026-07-28, so agent is excluded by time. That makes the 2,613 `control.revoke` rows provably web. The remaining 52 are `input.rejected`, whose source is genuinely unrecoverable, and they are folded into `'web'` as a placeholder. Say so in a migration comment rather than letting a future reader believe all 2,665 were web sends.

All 2,665 are revoked, so they never dispatch, never render, and never reach a claim.

## Why `input.rejected` is excluded

114,645 of its 114,656 rows have no `trigger_source`, no `run_id`, and no revoker to follow — no rule can attribute them. They are terminal rows that `pendingChatQueueEventCondition` never selects, so leaving them null costs nothing, and including them would make PR 2's constraint unaddable.

## Append-only

`chat_events` rejects every `UPDATE` via `reject_chat_event_source_update`. Narrow the trigger to permit only `context_type` / `context_id`, with every other column byte-identical, and restore the strict function in the same migration. **Follow `0824_clear_goal_context_pointers.sql`** — same table, same shape. Do not disable the trigger.

The backfill touches ~125,000 rows, an order of magnitude more than `0824`'s 19,059. Batch it as `0811_clear_non_goal_run_groups.sql` does — 10,000 per batch, `ORDER BY id`, `FOR UPDATE ... SKIP LOCKED`, `COMMIT` per batch, with the `-- vm0:non-transactional` header — so normal traffic continues. No `LOCK TABLE`.

Migration numbers collide in this repo; use the next available after `main` and expect to renumber.

## Acceptance criteria

- Every `input.prompt` / `input.automation` / `input.goal` row has a non-null `context_type`, enforced by the PR 2 constraint.
- A newly sent web message has `context_type = 'web'`, `context_id IS NULL`, **and so does its claimed replacement.**
- A goal continuation has `context_type = 'goal'`, `context_id IS NULL`, and the same holds after claim.
- The 21 Slack rows reachable only through rule 5 end up as `context_type = 'slack'`, not `'web'`. Assert this in the migration or verify it in the PR description with a count.
- `input.rejected` rows are untouched and the constraint does not cover them.
- The append-only trigger is narrowed and restored; the migration ends with a fail-closed assertion that no covered row lacks a `context_type`.
- Dispatch behaviour is unchanged — `chat_events.trigger_source` is still written and still drives the switch.
- No client-visible change.

## Non-goals

- **Do not change dispatch.** That is step 3, and it depends on PR 2's constraint existing.
- **Do not stop writing `chat_events.trigger_source`.** Step 4.
- Do not add a `chat_web_context` or `chat_goal_context` table. These two values are discriminators precisely because there is nothing to store.
- Do not touch `zero_runs.trigger_source`, the logs contract, or `chat_*_context` tables.

lancy:
## Summary

Part 2b of #25284 (step 2 of #25260). Closes #25284.

- add `chat_events_input_context_type_check` so every `input.prompt`, `input.automation`, and `input.goal` row requires a non-null `context_type`
- keep `input.rejected` outside the covered set and explicitly verify that its discriminator may remain null
- update the current-schema canonical input fixture to write `web`
- leave dispatch, `chat_events.trigger_source` writers, `zero_runs.trigger_source`, logs contracts, and all `chat_*_context` tables unchanged

This PR was opened only after PR #25294's migration and API writer promotion completed and its production observation gate passed.

## Production gate before opening

PR #25294 production release:

- migration completed: **2026-08-05 12:43:27Z**
- API promotion completed: **2026-08-05 12:43:49Z**
- observation window: **2026-08-05 12:43:49Z–12:59:32Z**
- two read-only covered-NULL checks more than 10 minutes apart: **0 → 0**
- migration final assertion and NOTICE: **21** rule-5 Slack rows, all classified as `slack`
- production traffic in the window: 27 new `web` inputs, including 13 claimed replacements; zero `web` rows had a non-null `context_id`
- Axiom: zero context-constraint errors and zero 5xx on chat paths in the window

Release workflow: https://github.com/vm0-ai/vm0/actions/runs/31006153528

## Rollout ledger

| Phase | Production release time (UTC) | Observation window | Gate |
| --- | --- | --- | --- |
| PR 1: relax/write/backfill (#25294) | Migration 2026-08-05 12:43:27Z; API promotion 2026-08-05 12:43:49Z | 2026-08-05 12:43:49Z–12:59:32Z | Complete; covered NULL count remained 0 |
| PR 2: tighten CHECK (#25305) | Migration 2026-08-05 13:48:52Z; API promotion 2026-08-05 13:49:25Z; release terminal-success 2026-08-05 13:49:44Z | 2026-08-05 13:49:25Z–14:05:04Z (15m 39s) | Complete; covered NULL count remained 0 |

PR 2 release workflow: https://github.com/vm0-ai/vm0/actions/runs/31011671240

## Production evidence after tightening

- covered `input.prompt` / `input.automation` / `input.goal` rows with null `context_type`: **0** at every read-only check, including the end of the observation window
- the excluded `input.rejected` null discriminator population remained **114,656**
- production accepted 18 `input.prompt/web`, 1 `input.automation/automation`, and 1 `input.prompt/automation` rows after the CHECK was live
- Axiom `vm0-web-logs-prod`: 2,402 rows examined, zero `chat_events_input_context_type_check` matches, complete and without warnings
- Axiom `vm0-request-log-prod`: 161,695 rows examined, zero 5xx on chat paths, complete and without warnings

## Validation

- `pnpm -F @vm0/db lint`
- `DATABASE_URL=postgresql://postgres:postgres@localhost:5432/postgres pnpm -F @vm0/db test:migration-consistency`
  - all historical migrations and database resets replayed
  - latest snapshot 837 matched the migrated schema
  - covered input with null discriminator was rejected by `chat_events_input_context_type_check`
  - context-less `input.rejected` remained accepted
  - regenerated schema matched the committed migration chain
- `pnpm -F @vm0/db db:generate` (no schema changes)
- `pnpm -F api lint`
- `pnpm -F api check-types`
- `pnpm -F api exec vitest run src/signals/routes/__tests__/cron-monitor-chat-event-queue.test.ts --silent=passed-only` (9/9)
- `pnpm -F api exec vitest run src/signals/routes/__tests__/cron-execute-morning-briefs.test.ts --silent=passed-only` (10/10)
- `git diff --check`
- complete PR and merge-group CI passed

Per issue guidance, the complete local test suite and dev server were not run; PR CI was the broad validation gate.

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/cron-monitor-chat-event-queue.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/cron-monitor-chat-event-queue.test.ts
@@ -117,32 +117,12 @@ describe("cron monitor chat event queue", () => {
     expect(context.mocks.sentry.captureException).not.toHaveBeenCalled();
   });
 
-  it("alerts for a pointerless prompt with a null trigger source", async () => {
-    const fixture = await trackFixture(seedFixture("orphan"));
-
-    const response = await accept(
-      stateClient().monitor({ body: { event_ids: [fixture.eventId] } }),
-      [500],
-    );
-
-    expect(response.body).toStrictEqual({
-      error: "Internal server error",
-    });
-    expect(
-      context.mocks.sentry.captureException.mock.calls.at(-1)?.[0],
-    ).toMatchObject({
-      code: "ORPHANED_QUEUED_CHAT_MESSAGES",
-      orphanedMessages: 1,
-      orphanedMessagesBySource: { "(unknown)": 1 },
-    });
-  });
-
   it("does not alert for pointerless web, test, or agent prompts", async () => {
     const fixture = await trackFixture(seedFixture("orphan"));
 
     const response = await accept(
       stateClient().monitor({
-        body: { event_ids: fixture.eventIds.slice(1, 4) },
+        body: { event_ids: fixture.eventIds.slice(0, 3) },
       }),
       [200],
     );
@@ -158,7 +138,7 @@ describe("cron monitor chat event queue", () => {
     const fixture = await trackFixture(seedFixture("orphan"));
 
     const response = await accept(
-      stateClient().monitor({ body: { event_ids: fixture.eventIds.slice(4) } }),
+      stateClient().monitor({ body: { event_ids: fixture.eventIds.slice(3) } }),
       [500],
     );
 
--- a/turbo/apps/api/src/signals/routes/test-cron-monitor-chat-event-queue-state.ts
+++ b/turbo/apps/api/src/signals/routes/test-cron-monitor-chat-event-queue-state.ts
@@ -45,22 +45,17 @@ type DbTransaction = Parameters<Parameters<Db["transaction"]>[0]>[0];
 
 const STALE_CONTEXT_FIXTURES = [
   {
-    contextType: null,
-    eventType: "input.prompt",
-    triggerSource: null,
-  },
-  {
-    contextType: null,
+    contextType: "web",
     eventType: "input.prompt",
     triggerSource: "web",
   },
   {
-    contextType: null,
+    contextType: "web",
     eventType: "input.prompt",
     triggerSource: "test",
   },
   {
-    contextType: null,
+    contextType: "agent_run",
     eventType: "input.prompt",
     triggerSource: "agent",
   },
@@ -193,6 +188,7 @@ async function seedGoalFixture(
     .insert(chatEvents)
     .values({
       chatThreadId: args.threadId,
+      contextType: "goal",
       eventType: "input.goal",
       runGroupId: goal.id,
       userMessage: createUserMessageDocument({
@@ -316,6 +312,23 @@ async function seedFixture(
         }),
       ];
     }
+    if (fixtureKind === "queued-integration") {
+      const [event] = await tx
+        .insert(chatEvents)
+        .values({
+          ...baseEvent,
+          contextType: "slack",
+          contextId: randomUUID(),
+          eventType: "input.prompt",
+          triggerSource: "slack",
+          seqId: 1,
+        })
+        .returning({ id: chatEvents.id });
+      if (!event) {
+        throw new Error("Failed to seed queued integration event");
+      }
+      return [event];
+    }
     const event =
       fixtureKind === "failed-message"
         ? await insertChatEvent(tx, {
@@ -325,9 +338,9 @@ async function seedFixture(
           })
         : await insertChatEvent(tx, {
             ...baseEvent,
+            contextType: "web",
             eventType: "input.prompt",
-            triggerSource:
-              fixtureKind === "queued-integration" ? "slack" : "web",
+            triggerSource: "web",
           });
     return [event];
   });
--- a/turbo/apps/api/src/signals/routes/test-cron-monitor-chat-event-queue-state.ts
+++ b/turbo/apps/api/src/signals/routes/test-cron-monitor-chat-event-queue-state.ts
@@ -231,6 +231,37 @@ async function seedGoalAgent(
   signal.throwIfAborted();
 }
 
+async function seedQueuedIntegrationEvent(tx: DbTransaction, threadId: string) {
+  const [event] = await tx
+    .insert(chatEvents)
+    .values({
+      chatThreadId: threadId,
+      contextType: "slack",
+      contextId: randomUUID(),
+      eventType: "input.prompt",
+      triggerSource: "slack",
+      userMessage: createUserMessageDocument({
+        text: "orphan monitor fixture",
+      }),
+      runId: null,
+      seqId: 1,
+    })
+    .returning({ id: chatEvents.id });
+  if (!event) {
+    throw new Error("Failed to seed queued integration event");
+  }
+  return event;
+}
+
+function requireSeededEventId(
+  event: { readonly id: string } | null | undefined,
+): string {
+  if (!event) {
+    throw new Error("Failed to seed orphan monitor message");
+  }
+  return event.id;
+}
+
 async function seedFixture(
   db: Db,
   fixtureKind: FixtureKind,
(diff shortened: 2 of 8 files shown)


lancy:
Review round 1: LGTM. The PR is limited to the tightening CHECK plus generated migration artifacts and migration-consistency coverage. The constraint matches #25284 exactly, keeps input.rejected nullable, and the focused migration replay, typecheck, lint, formatting, and regeneration checks pass.

lancy:
CI follow-up: the third run failed only in the pre-existing `revokes a goal invalidated at the final failed-launch boundary` test because its lock probe observed `select_for_update` before `update`; PostgreSQL logged a tuple lock timeout. This PR does not modify that test or the related goal services, no new constraint violation appeared in the logs, the exact test passes locally (1/1), and post-PR1 main Turbo run 31006153623 passed. Re-running only test-api (3).

lancy:
Re-review round 2: LGTM. I re-read the full PR diff after the fixture fixes, verified migration 0837 is still the next available number on current main, confirmed all PR checks are green, and re-queried the production gate: covered input rows with a null context_type remain at 0 while the excluded input.rejected null population remains 114,656.

PR #25346, opened 2026-08-06
test(guest-agent): contain timed-out command descendants

### LINKED ISSUE #25328: bug(rust): contain guest-agent test subprocess descendants
## Problem and evidence

`command_output_with_timeout` in `crates/guest-agent/tests/common/mod.rs` owns and reaps only the directly spawned child. The helper drains stdout and stderr in the same deadline as `child.wait()`, so the direct child can exit and be reaped while a descendant still holds either capture pipe open. When the deadline expires, `child.kill().await` has authority only over the direct child and the helper can return while that descendant remains running.

This is reproducible at commit `dae43ba457ac8d2da48eb2727132da06ce7a5894` with a 250 ms helper deadline around a shell that starts a 30-second background sleep, records its PID, and exits immediately:

```text
/bin/sh -c '(sleep 30) & printf "%s\n" "$!" > "$VM0_DESCENDANT_PID_FILE"'
```

The unchanged helper produced:

```text
result_kind=Some(TimedOut)
elapsed_ms=251
alive_after_return=true
```

The reproduction then explicitly killed the recorded descendant.

The production-shaped test path is stronger than this shell example. Current `guest-agent` CLI and Codex app-server launches use `Command::process_group(0)`, so mock CLI descendants intentionally enter process groups different from the directly launched `guest-agent` while remaining in its session. A single process group around the helper's direct child would pass the shell reproduction but would not contain these existing real descendants.

The helper is shared by execution-timeout, cancellation-recovery, event-delivery recovery, session-history, API-token isolation, and focused lifecycle integration tests. #25333 separately proposes migrating additional Codex entry-point tests to the same helper.

## Impact

A guest-agent integration timeout can return after reaping the direct `guest-agent` PID while leaving a mock CLI or another descendant alive. That process outlives the mock server and temporary runtime directories, retains file descriptors and resources, and can contaminate later tests in the same CI worker.

`kill_on_drop(true)` does not close the gap because it also targets only the direct child. A direct-child process group is also insufficient because current guest-agent descendants can create their own groups.

## Decision

Give every helper-managed command a dedicated Linux session. Preserve the successful `Output` contract, but on managed timeout terminate all live members of that session across process-group changes with pidfd-safe signaling, then reap the direct child within the same bounded cleanup operation.

The helper will continue draining stdout and stderr concurrently, but it will defer `child.wait()` until both pipes reach EOF. This keeps an exited direct session leader unreaped while a descendant holds an inherited pipe and preserves the session identity through cleanup.

Timeout cleanup will start direct-child termination, enumerate live `/proc` members of the owned session, bind candidates to pidfds, recheck live state and session membership, signal only through pidfds, rescan until empty under one five-second deadline, and reap the direct child last. The ordinary `TimedOut` error is returned only after cleanup is proven; cleanup uncertainty remains an error.

## Alternatives considered

- **One process group:** rejected because current guest-agent CLI and app-server descendants create their own groups.
- **Per-command cgroup v2:** stronger and covers new sessions, but ordinary local and coverage containers have no delegated writable cgroup contract; adopting it would require privileged test infrastructure or fallback behavior.
- **Subreaper supervisor:** can extend beyond session boundaries, but needs another executable layer, output/control forwarding, ancestry tracking, and separate failure semantics. No current caller creates a new session.
- **Status quo/direct-child kill:** rejected by the reproduced leak.

## Risks and mitigations

- **PID reuse:** keep the direct session leader unreaped during cleanup and signal only pidfds after membership rechecks.
- **Concurrent exits or forks:** rescan the session until no live member remains or the shared cleanup deadline expires.
- **Pipe deadlock:** retain concurrent stdout/stderr draining; only direct reaping moves after EOF.
- **Cleanup latency:** `/proc` scans occur only after an already-failing timeout or setup invariant failure; successful commands pay only one `setsid` syscall.
- **Session escape:** a descendant that explicitly calls `setsid` remains outside this boundary. No current helper caller or guest-agent child path does so; cgroup or supervisor support is not justified without evidence.
- **Regression leakage:** retain the fixture descendant's pidfd and explicitly clean that exact process before failing if the old behavior is observed.

## Acceptance criteria

- Every helper-managed command starts as an owned Linux session leader.
- On managed timeout, all live members of the owned session terminate, including descendants in separate process groups, and the direct child is reaped before return.
- Cleanup is explicitly bounded and sends no signal through an unverified raw numeric PID.
- Normal exit status, stdout, and stderr capture remain unchanged, including concurrent pipe draining.
- Existing normal-output and live-direct-child lifecycle coverage remains.
- A deterministic integration regression starts a long-lived separate-process-group descendant that inherits the capture pipes, synchronizes parent/descendant readiness and direct-parent exit before advancing the command deadline, and proves the exact descendant has exited when the helper returns.
- All current helper consumers and the focused/full guest-agent tests remain green, together with formatting, all-target/all-feature Clippy, and rustdoc.

## Delivery notes

This is one independently reviewable, Linux test-infrastructure PR limited to the shared guest-agent test helper, a focused test-common session module, and lifecycle regression coverage. It is a follow-up to #24742 and is not a duplicate of the runner fixture work in #24883 or the caller migration in #25333.

Production guest process containment, runner behavior, APIs, protocols, schemas, persistence, migrations, dependencies, and deployment behavior remain unchanged.

seven332:
## Summary

- launch commands managed by the guest-agent integration-test helper in a dedicated Linux session
- on timeout, discover every live member of that session across process groups, bind each process identity with a pidfd, terminate it, and reap the direct child before returning
- add an integration regression where the direct child exits while a descendant in a separate process group keeps the captured output pipes open

## Scope

- test infrastructure only; no production runtime, API, schema, persisted data, migration, or dependency changes
- descendants that explicitly create a new session are outside this cleanup boundary; no current helper caller does that

## Validation

- `RUSTDOCFLAGS='-D warnings' cargo doc -p guest-agent --no-deps --all-features --quiet`
- `cargo test -p guest-agent --test command_output_timeout --all-features -- --nocapture`
- descendant regression repeated successfully 20 consecutive times
- `cargo test -p guest-agent --all-targets --all-features --quiet`
- `git diff --check`

Closes #25328

### DIFF
--- a/crates/guest-agent/tests/command_output_timeout.rs
+++ b/crates/guest-agent/tests/command_output_timeout.rs
@@ -3,11 +3,32 @@
 mod common;
 
 use std::io;
+use std::os::fd::OwnedFd;
+use std::os::unix::process::CommandExt;
 use std::path::PathBuf;
-use std::time::Duration;
+use std::process::{Command as StdCommand, Stdio};
+use std::time::{Duration, Instant};
+use tokio::io::unix::AsyncFd;
 use tokio::process::Command;
 
 const MANAGED_TIMEOUT: Duration = Duration::from_secs(20);
+const FIXTURE_READY_TIMEOUT: Duration = Duration::from_secs(5);
+const FIXTURE_EXIT_TIMEOUT: Duration = Duration::from_secs(5);
+const DESCENDANT_EXIT_TIMEOUT: Duration = Duration::from_secs(1);
+const DESCENDANT_READY_ENV: &str = "VM0_TEST_DESCENDANT_READY_FILE";
+const DESCENDANT_RELEASE_ENV: &str = "VM0_TEST_DESCENDANT_RELEASE_FILE";
+const DESCENDANT_FIXTURE: &str = "command_output_timeout_descendant_fixture";
+
+struct DescendantCleanup {
+    pidfd: AsyncFd<OwnedFd>,
+}
+
+impl Drop for DescendantCleanup {
+    fn drop(&mut self) {
+        let _ =
+            rustix::process::pidfd_send_signal(self.pidfd.get_ref(), rustix::process::Signal::KILL);
+    }
+}
 
 #[tokio::test]
 async fn command_output_timeout_preserves_completed_output()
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #25346

**Reviewed head:** `d0359c959866e8d50b06a93fa6e107fcad5d5fd3`

### Summary

The test helper now owns a dedicated Linux session, keeps the direct session leader unreaped while inherited pipes remain open, and performs bounded pidfd-safe cleanup across descendant process groups. The focused regression exercises the actual failure topology and verifies exact descendant termination.

### Findings

No findings.

### Testing

Formatting, all-target/all-feature Clippy, rustdoc, the focused integration test, 20 consecutive submitted-head regression runs, and the full `guest-agent` all-target/all-feature test suite passed. Full Rust workspace tests were not run because the changes are confined to `guest-agent` integration-test infrastructure and all current helper consumers are covered by the crate-wide suite.

### Verdict

LGTM

PR #25355, opened 2026-08-06
fix: neutralize non-allowlisted html tags at the markdown root

### LINKED ISSUE #25354: bug: top-level raw <style> in markdown escapes tag neutralization and breaks activity page layout
## Summary

On the Activity detail page (`/activities/:id`), a run whose step output contains a **top-level raw `<style>` block** injects that CSS into the whole app document. Everything on the page loses its padding and margins (labels run together — `WebModel`, `StepsContextRunnerNetwork` — the header card, tab bar and search input lose their inner padding, the centered `mx-auto` container snaps to the left edge) while colors, borders and `gap-*` still work.

The page renders correctly at first and breaks the moment the full step list arrives, which is when the offending step is mounted.

Observed on a run that generated an HTML slide deck; the generated markup starts with the classic reset:

```css
* { margin: 0; padding: 0; box-sizing: border-box; }
```

Because those declarations are unlayered, they win over every Tailwind utility in `@layer utilities`, which is exactly the padding/margin-only breakage seen on screen.

## Root cause

`turbo/apps/platform/src/views/components/markdown.tsx` — `rehypeRewriteHandler` converts every tag outside `validHtmlTags` (which does not contain `style`, `script`, `link`, `meta`) into plain text, but only when:

```ts
node.type === "element" &&
!validHtmlTags.has(node.tagName) &&
parent?.type === "element"   // <-- root-level nodes are skipped
```

Raw HTML written at the top level of a markdown document lands **directly under the hast `root`**, so `parent.type === "root"` and the neutralization never runs. Verified against the same plugin chain:

| source | `parent.type` | result |
| --- | --- | --- |
| `<style>…</style>` at top level | `root` | **not rewritten — rendered as a live `<style>` element** |
| `<div><style>…</style></div>` | `element` | rewritten to text (safe) |

`@uiw/react-markdown-preview/common` enables `rehype-raw` and ships **no** `rehype-sanitize`, so anything the rewrite misses reaches the DOM verbatim. Activity steps render `<Markdown source={…} />` without `escapeHtml` (`event-card.tsx`, `event-group-card.tsx`, `zero-activity-detail-page.tsx`), so agent output takes that unsanitized path.

The same gap lets a top-level `<script>` through, so this is not only a layout problem — agent-authored content can currently inject executable markup into the app document.

## Expected

Non-allowlisted tags are neutralized to plain text regardless of their depth, so agent output can never restyle or script the surrounding app.

## Repro

Open any activity whose steps contain a raw HTML block beginning with `<style>` at the start of a line, e.g. an agent message containing a generated HTML deck.

bingjiez666:
## Problem

An activity page unstyles itself the moment its steps finish loading: every element loses its padding and margins (`WebModel`, `StepsContextRunnerNetwork`, the header card and search input collapse onto their own edges) while colors, borders and `gap-*` keep working.

The run in question generated an HTML slide deck whose markup opens with the classic reset:

```css
* { margin: 0; padding: 0; box-sizing: border-box; }
```

That reset is unlayered, so it beats every Tailwind utility in `@layer utilities` — which is exactly the padding/margin-only breakage that shows up on screen.

## Root cause

`rehypeRewriteHandler` in `markdown.tsx` turns every tag outside `validHtmlTags` (no `style`, `script`, `link`, `meta`) into plain text, but only when `parent?.type === "element"`. Raw HTML written at the top level of a markdown document lands **directly under the hast `root`**, so the rewrite never ran for it:

| source | `parent.type` | before this PR |
| --- | --- | --- |
| `<style>…</style>` at top level | `root` | rendered as a live `<style>` element |
| `<div><style>…</style></div>` | `element` | rewritten to text |

`@uiw/react-markdown-preview/common` enables `rehype-raw` and ships no `rehype-sanitize`, so whatever the rewrite misses reaches the DOM verbatim. Activity steps render `<Markdown source={…} />` without `escapeHtml`, so agent output takes that path. The same gap also let a top-level `<script>` through.

## Change

Rewrite root-level nodes as well, so non-allowlisted tags are neutralized regardless of depth. Adds a regression test asserting a top-level `<style>` renders as text and produces no `<style>` element.

Closes #25354

### DIFF
--- a/turbo/apps/platform/src/views/components/__tests__/markdown.test.tsx
+++ b/turbo/apps/platform/src/views/components/__tests__/markdown.test.tsx
@@ -87,6 +87,18 @@ describe("assistant markdown", () => {
     expect(container.querySelector(".wmde-markdown span")).toBeNull();
   });
 
+  it("neutralizes non-allowlisted tags written at the top level", () => {
+    const { container } = render(
+      <StoreProvider value={context.store}>
+        <Markdown source={"<style>\n* { margin: 0; padding: 0; }\n</style>"} />
+      </StoreProvider>,
+    );
+
+    // A live <style> element would apply its rules to the whole app document.
+    expect(container.querySelector("style")).toBeNull();
+    expect(container.textContent).toContain("* { margin: 0; padding: 0; }");
+  });
+
   it("keeps blockquotes rendering when html is escaped", () => {
     const { container } = render(
       <StoreProvider value={context.store}>
--- a/turbo/apps/platform/src/views/components/__tests__/markdown.test.tsx
+++ b/turbo/apps/platform/src/views/components/__tests__/markdown.test.tsx
@@ -87,18 +87,6 @@ describe("assistant markdown", () => {
     expect(container.querySelector(".wmde-markdown span")).toBeNull();
   });
 
-  it("neutralizes non-allowlisted tags written at the top level", () => {
-    const { container } = render(
-      <StoreProvider value={context.store}>
-        <Markdown source={"<style>\n* { margin: 0; padding: 0; }\n</style>"} />
-      </StoreProvider>,
-    );
-
-    // A live <style> element would apply its rules to the whole app document.
-    expect(container.querySelector("style")).toBeNull();
-    expect(container.textContent).toContain("* { margin: 0; padding: 0; }");
-  });
-
   it("keeps blockquotes rendering when html is escaped", () => {
     const { container } = render(
       <StoreProvider value={context.store}>
@@ -145,6 +133,34 @@ describe("assistant markdown", () => {
         document.querySelector('[data-color-mode="light"]'),
       ).toBeInTheDocument();
     });
+  });
+
+  it("shows a raw style block as text instead of styling the page", async () => {
+    mockThread("<style>\n.zero-injected { color: red }\n</style>");
+
+    detachedSetupPage({ context, path: "/chats/thread-markdown" });
+
+    await waitFor(() => {
+      expect(document.querySelector(".wmde-markdown")?.textContent).toContain(
+        ".zero-injected { color: red }",
+      );
+    });
+    // A mounted stylesheet would restyle the whole page, not just this message.
+    const injectedSheets = Array.from(
+      document.querySelectorAll("style"),
+    ).filter((sheet) => {
+      return sheet.textContent?.includes(".zero-injected") ?? false;
+    });
+    expect(injectedSheets).toHaveLength(0);
+  });
+
+  it("keeps allowlisted html blocks rendering as elements", async () => {
+    mockThread("<div><strong>kept markup</strong></div>");
+
+    detachedSetupPage({ context, path: "/chats/thread-markdown" });
+
+    const kept = await screen.findByText("kept markup", { selector: "strong" });
+    expect(kept).toBeInTheDocument();
   });
 
   it("renders media links inline", async () => {
--- a/turbo/apps/platform/src/views/components/markdown.tsx
+++ b/turbo/apps/platform/src/views/components/markdown.tsx
@@ -152,11 +152,15 @@ const rehypeRewriteHandler = (() => {
   return (...args: RewriteArgs) => {
     const [node, , parent] = args;
 
-    // Convert unknown HTML tags to plain text, preserving child content
+    // Convert unknown HTML tags to plain text, preserving child content.
+    // Raw HTML written at the top level of a document lands directly under the
+    // hast root, so root-level nodes must be rewritten too — otherwise tags
+    // like <style> and <script> survive into the app's DOM, where a generated
+    // `* { margin: 0; padding: 0 }` reset unstyles the whole page.
     if (
       node.type === "element" &&
       !validHtmlTags.has(node.tagName) &&
-      parent?.type === "element"
+      (parent?.type === "element" || parent?.type === "root")
     ) {
       const inner = collectText(node);
       const text = inner


bingjiez666:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 6469d7d183deffaaaac5e4027e9e9a640930726d

### Summary

Widens the unknown-tag neutralization in `rehypeRewriteHandler` to root-level hast nodes, so raw HTML written at the top level of a markdown document (`<style>`, `<script>`, and any other non-allowlisted tag) is turned into plain text instead of reaching the app DOM. The production fix is correct and minimal; the regression test enters through direct component rendering instead of the page, which the platform testing docs prohibit.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and edge cases | Pass | `parent` is only absent for the hast root itself, whose `type` is `root` (not `element`), so widening the guard to `parent?.type === "root"` opens no new path; confirmed against the same `remark-parse` → `remark-rehype` → `rehype-raw` → `rehype-rewrite` chain the app uses. |
| `REVIEW.md` | Security (injection / XSS) | Pass | `turbo/apps/platform/src/views/components/markdown.tsx`: closes the only path by which agent-authored `<style>`/`<script>` reached the live document; `@uiw/react-markdown-preview/common` ships `rehype-raw` without `rehype-sanitize`, so this handler is the sole barrier. |
| `REVIEW.md` | Type safety, no `any`, no suppressions | Pass | Diff adds one boolean condition and a comment; no casts, no `eslint-disable`/`@ts-*` comments. |
| `REVIEW.md` | `fix:` commit requires a regression test | Pass | `markdown.test.tsx` adds a top-level `<style>` case that asserts no `<style>` element is produced. |
| `REVIEW.md` | Feature regression guard — cover the existing flow most likely to regress | Finding (P1) | The guard now also fires for root-level nodes, so top-level **allowlisted** raw HTML is newly routed through the allowlist check. No test pins that a top-level `<div>`/`<span>`/`` block still renders as an element. |
| `REVIEW.md` | DB / JSONB persisted-contract gate | N/A | No schema, contract, or `jsonb(...)` change; the diff is frontend rendering only. |
| `docs/bad-smell.md` | No `any`, suppressions, dynamic `import()`, hardcoded URLs, or fallback recovery paths | Pass | None introduced; the change removes a silent-escape path rather than adding a fallback. |
| `docs/testing.md` | Mock at the system boundary; no fake timers; assert behavior, not mock calls | Pass | No `vi.mock`, no timers, no mock-call assertions added. |
| `docs/testing/app-testing.md` | Platform tests are page-level by default and must enter through `detachedSetupPage`; no component-only tests when a page surface exists | Finding (P1) | `markdown.test.tsx:90-100` renders `<Markdown>` inside a bare `StoreProvider`. |
| `docs/testing/anti-patterns.md` | AP-10 Direct Component Rendering | Finding (P1) | Same block; matches the documented anti-pattern shape exactly. |
| `docs/testing/testing-external-behavior.md` | Construct the case and observe the result through the page | Finding (P1) | The user-visible surface is an assistant message / activity step; the test never renders one. |
| `docs/testing/patterns.md` | Pattern 7 — platform page tests live under `views/**/__tests__/` and bootstrap via `detachedSetupPage` with `context.mocks` | Finding (P1) | Same block; the file's own `mockThread` + `detachedSetupPage` harness is already available two tests below. |
| `docs/effect.md` | Choose the execution boundary; keep render pure | N/A | No commands, effects, refs, or route setup touched; `rehypeRewriteHandler` is a module-scope pure rewrite. |
| `docs/cache.md` | RC-1…RC-9 render purity, signal allocation, cache lifetime, resource teardown | N/A | No signals, caches, refs, listeners, or resources are created or retained by this diff. |
| `.claude/skills/ccstate/SKILL.md` | ccstate signal/command conventions for platform code | N/A | The diff contains no ccstate state, computed, or command definitions. |

### Findings

#### Critical (P0)

- None.

- `turbo/apps/platform/src/views/components/__tests__/markdown.test.tsx:90-100`: the regression test renders the `Markdown` component directly inside a bare `StoreProvider`. `docs/testing/app-testing.md` ("Do not add … component-only unit tests under `turbo/apps/platform`. Cover behavior through a rendered page whenever there is a user-visible page surface"), `docs/testing/anti-patterns.md` AP-10, and `docs/testing/testing-external-behavior.md` all require the page path here, and a page surface does exist: assistant messages render markdown **unescaped** (`escapeMarkdownHtml` is passed only for user messages, `zero-chat-thread-page.tsx:7748`), so the same file's `mockThread(...)` + `detachedSetupPage(...)` harness reproduces the real bug. The page-level version is also strictly stronger — it proves the injected rules never reach the surrounding app document, which is the actual defect; the component-level version only proves one subtree's markup.
- `turbo/apps/platform/src/views/components/__tests__/markdown.test.tsx`: no coverage for the path most likely to regress from this change. Root-level allowlisted raw HTML now enters the allowlist branch for the first time; add an assertion that a top-level block such as `<div>…</div>` still renders as an element so a future edit to the condition or to `validHtmlTags` cannot silently start escaping legitimate HTML.

### Testing

- Coverage: Insufficient — the regression case is present but asserted at the wrong boundary, and the allowlisted-tag path that the guard change newly touches is untested.
- Conventions: Violations — direct component rendering (AP-10) instead of `detachedSetupPage`; no `vi.mock`, fake-timer, fetch-stubbing, or mock-call-assertion violations.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 79942c816daba164ea6526cfaa9b0929d82cf340
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Widens the unknown-tag neutralization in `rehypeRewriteHandler` to root-level hast nodes so raw HTML written at the top level of a markdown document (`<style>`, `<script>`, and any other non-allowlisted tag) becomes plain text instead of reaching the app DOM. Both P1 findings from the previous review round are resolved on `79942c8`: the regression coverage now enters through the page bootstrap path, and the allowlisted-tag path the guard change newly touches is pinned by its own test.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and edge cases | Pass | `parent` is absent only for the hast root itself, whose `type` is `root` rather than `element`, so the widened guard opens no new path; the two branches of `rehypeRewriteHandler` are otherwise untouched. |
| `REVIEW.md` | Security (injection / XSS) | Pass | `markdown.tsx:160-164` closes the only route by which agent-authored `<style>`/`<script>` reached the live document; `@uiw/react-markdown-preview/common` ships `rehype-raw` without `rehype-sanitize`, so this handler is the sole barrier. |
| `REVIEW.md` | Type safety, no `any`, no suppressions | Pass | The production diff adds one boolean condition and a comment; the test diff adds no casts and no `eslint-disable`/`@ts-*` comments. |
| `REVIEW.md` | `fix:` commit requires a regression test | Pass | `markdown.test.tsx:138-155` renders an assistant message containing a top-level `<style>` block and asserts the CSS reaches the user as text with no stylesheet carrying those rules anywhere in the document. |
| `REVIEW.md` | Feature regression guard — cover the existing flow most likely to regress | Pass | `markdown.test.tsx:157-164` pins that a top-level allowlisted block (`<div><strong>…`) still renders as real elements, which is the path the widened guard newly reaches. |
| `REVIEW.md` | DB / JSONB persisted-contract gate | N/A | No schema, contract, or `jsonb(...)` change; the diff is frontend rendering plus tests. |
| `docs/bad-smell.md` | No `any`, suppressions, dynamic `import()`, hardcoded URLs, or fallback recovery paths | Pass | None introduced; the change removes a silent-escape path rather than adding a fallback. |
| `docs/testing.md` | Integration over unit; mock at the boundary; no fake timers; assert behavior rather than mock calls | Pass | Both new tests are page-level, use the existing `context.mocks.api` thread fixture, and add no `vi.mock`, timers, or mock-call assertions. |
| `docs/testing/app-testing.md` | Platform tests are page-level and enter through `detachedSetupPage` | Pass | Both new cases call `mockThread(...)` then `detachedSetupPage({ context, path: "/chats/thread-markdown" })`; the previous direct-render case is gone. |
| `docs/testing/anti-patterns.md` | AP-10 Direct Component Rendering | Pass | No `render(<StoreProvider>…)` is added by this PR. |
| `docs/testing/testing-external-behavior.md` | Construct the case and observe the result from the user's surface | Pass | The case is built as an assistant message and observed on the rendered page; the `style`-element sweep asserts the document surface the defect actually corrupts. |
| `docs/testing/patterns.md` | Pattern 7 — platform page tests bootstrap through `detachedSetupPage` with `context.mocks` | Pass | Matches the shape of the surrounding media-link and mermaid tests in the same file. |
| `docs/effect.md` | Choose the execution boundary; keep render pure | N/A | No commands, effects, refs, or route setup touched; `rehypeRewriteHandler` is a module-scope pure rewrite. |
| `docs/cache.md` | RC-1…RC-9 render purity, signal allocation, cache lifetime, resource teardown | N/A | No signals, caches, refs, listeners, or resources are created or retained by this diff. |
| `.claude/skills/ccstate/SKILL.md` | ccstate signal/command conventions for platform code | N/A | The diff contains no ccstate state, computed, or command definitions. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None. Both findings from the `6469d7d` round are resolved on `79942c8`.

### Testing

- Coverage: Adequate — the top-level `<style>` regression and the allowlisted-tag path are both covered through the page, and the first test was confirmed to fail against the pre-fix condition and pass after it.
- Conventions: Compliant. The style-block test reads the rendered markdown container by class and sweeps `style` elements at document scope; that is a deliberate deviation from pure role/text queries because the defect is precisely a stylesheet escaping into the document, and it matches the DOM-query convention already used by the neighboring tests in this file.

PR #25362, opened 2026-08-06
perf(python): bound network log url serialization

### LINKED ISSUE #25330: perf(python): bound network-log URL serialization before queueing
## Problem

HTTP network logging currently sanitizes, JSON-renders, and UTF-8 encodes the complete request-controlled URL synchronously in the mitmproxy response or error hook before the bounded JSONL queue sees the line.

The pinned mitmproxy 12.2.3 HTTP/1 reader has no request-head size limit, so a registered workload can supply a multi-megabyte request target. The repository also explicitly preserves a 200,000-character URL through the real response hook in `crates/runner/mitm-addon/tests/test_response_handler_logging.py`.

## Current path

- `crates/runner/mitm-addon/src/mitm_addon.py` copies the complete target through `sanitize_request_url_for_network_log()` while constructing both response and connection-error entries.
- `crates/runner/mitm-addon/src/network_log_sanitization.py` scans, slices, sanitizes, and reconstructs the complete URL.
- `crates/runner/mitm-addon/src/logging_utils.py` executes `(json.dumps(entry) + "\n").encode()` before calling the asynchronous writer.
- `crates/runner/mitm-addon/src/jsonl_writer.py` applies its 4,096-entry and 32 MiB backlog budgets only after the complete byte string exists.
- `crates/runner/src/network_logs.rs` skips any entry that cannot fit inside its 1 MiB upload batch envelope.

## Evidence and impact

Measured on `dae43ba457ac8d2da48eb2727132da06ce7a5894` with Python 3.12.3, using the production sanitizer plus a production-shaped entry and JSONL encoding:

| URL payload | JSONL bytes | Median synchronous time | Peak traced allocation |
| ---: | ---: | ---: | ---: |
| 1 MiB | 1,048,835 | 2.683 ms | 3.25 MiB |
| 5 MiB | 5,243,139 | 12.686 ms | 16.25 MiB |
| 16 MiB | 16,777,475 | 39.183 ms | 52.00 MiB |
| 33 MiB | 34,603,267 | 89.872 ms | 107.25 MiB |

One oversized response or connection error can pause every flow sharing the proxy event loop. Repeated oversized requests keep paying proportional CPU and allocation costs even when the local queue or remote uploader will reject every resulting row.

This is distinct from #24316 and #24537, which removed runtime URL-cache retention and avoided parsing query data for diagnostics that discard it. Top-level network logs intentionally regained complete query and fragment preservation in #24554; the missing boundary is before serialization and queue admission.

## Decision

Add a bounded structured-omission path specifically for the top-level HTTP network-log URL:

- Reject URL logging work before sanitization when the raw Python string exceeds **1,000,000 characters**.
- Limit a URL-induced final HTTP JSONL line to **1,000,000 bytes including the newline**, leaving over 48 KiB below the Rust binary 1 MiB envelope for payload and run-id overhead.
- Represent an omitted URL as `url: "[truncated]"`, `url_truncated: true`, and `url_original_char_count: <len(raw URL)>`.
- Preserve the current complete-URL sanitizer and output for normal rows.

Use character count, not an alleged original byte count. The addon owns a Python string rather than the original request-target bytes, and obtaining an exact UTF-8/JSON byte count from a multi-megabyte value would require the full work this issue removes.

Do not retain a raw prefix. A prefix can expose credential-like userinfo when its separating `@` lies beyond the retained boundary.

For raw URLs below the character ceiling, encode the actual complete row once. If it exceeds the producer byte budget, encode a candidate with the URL omission projection. Select that candidate only when it fits; if the candidate is still oversized, non-URL data independently caused the overflow and existing queue/uploader behavior remains unchanged rather than falsely attributing truncation to the URL.

## Alternatives considered

- **Reject oversized request heads at mitmproxy ingress:** changes workload behavior, couples the fix to private multi-protocol parser boundaries, and does not establish the logging producer contract.
- **Move serialization to the JSONL worker:** shifts but does not bound proportional CPU/allocation work and makes pre-serialization byte admission impossible.
- **Truncate a raw URL prefix:** risks partial userinfo disclosure and requires a second security-sensitive URL parser.
- **Compute an exact dynamic remaining URL budget:** retains a few more edge-case bytes but adds parallel size estimation and staging complexity without evidence that URLs near the uploader ceiling are a required compatibility boundary.

## Compatibility and scope

The two metadata fields are optional additions to the shared network-log schema and API read projection.

- Old runner -> new backend remains unchanged.
- New runner -> old backend still supplies a valid, bounded existing `url` string; the old backend may omit the new metadata during overlap.
- Old/new frontend combinations tolerate absent additive fields and continue displaying the existing URL field.
- No database migration, persisted relational-state change, queue protocol version, feature switch, or cleanup release is required.

This issue bounds the demonstrated URL-induced path. Capture headers have no aggregate count budget and other non-URL fields can independently create an oversized row; changing those semantics requires separate evidence and is not hidden behind URL truncation here.

## Acceptance criteria

- URL input above 1,000,000 characters is omitted before sanitizer, full-value JSON, or UTF-8 processing.
- Normal HTTP rows, including the existing 200,000-character case, retain the complete URL, query/fragment contents, and current userinfo redaction.
- A below-ceiling URL whose JSON escaping makes the full row exceed 1,000,000 bytes uses the same omission projection when that projection makes the row uploadable.
- Oversized response and connection-error URLs produce `url: "[truncated]"`, `url_truncated: true`, and the exact original Python character count without retaining a raw prefix.
- URL-induced emitted lines fit both Python queue admission and the Rust single-entry upload envelope.
- New optional metadata survives backend ingestion and API read/download projections; payloads without it remain accepted.
- Regression coverage uses the real response and error hooks and structural byte assertions without timing, allocator, internal-mock, or E2E assertions.

seven332:
## Summary

- bound raw HTTP network-log URL processing before sanitization
- replace URL-driven oversized rows with an explicit omission projection that fits uploader limits
- preserve truncation metadata through telemetry ingestion and network-log reads

## Scope

- leaves independently oversized non-URL rows unchanged and unmarked
- no database migration, feature switch, or generated Rust binding change

## Validation

- `uv lock --check`
- `uv sync --locked`
- `pnpm -F @vm0/api-contracts generate:rust`
- `pnpm format`
- `pnpm turbo run lint`
- `pnpm check-types`
- `pnpm vitest run --maxWorkers=4 --silent=passed-only`
- `pnpm knip`

Closes #25330


seven332:
## Self-review

Reviewed head `5bd081214e2f9fedb9399f97edf6f9b94b93a5cb` against `main` across scope, correctness, tests, performance, security, and maintainability.

- The raw character-count gate runs before complete-URL sanitization or serialization.
- The normal path preserves complete URL and redaction behavior with one row encoding.
- The escaped-URL fallback is selected only when the omission projection fits the producer budget; independently oversized non-URL rows remain unmarked.
- The omission projection retains no request-controlled URL prefix or credentials.
- Optional metadata remains compatible across runner, API, and frontend deployment overlap.
- Real response/error hooks and API ingestion/read paths cover the changed behavior.

No in-scope findings remain and no follow-up issue was identified.

Validation completed:

- full locked mitm-addon Ruff, BasedPyright, and pytest checks (`4817 passed`)
- API contract Rust generation with no generated diff
- Turbo format, lint, typecheck, Knip, and full Vitest (`7589 passed`, one benchmark skipped)
- `git diff --check`

seven332:
## Code Review: PR #25362

**Reviewed head:** `5bd081214e2f9fedb9399f97edf6f9b94b93a5cb`

### Summary

The HTTP-specific logger applies a pre-sanitizer raw-character ceiling, preserves the existing normal URL path, and uses a fitting structured omission projection for URL-driven oversized rows. The optional metadata is carried through the shared contract, ingestion, read projection, and exhaustive frontend mapping without a migration or rollout break.

### Findings

No findings.

### Testing

Real response and error hooks cover both oversize paths and independent non-URL overflow. API route tests cover telemetry ingestion and Axiom reads. Full mitm-addon static/test validation and full Turbo format, lint, typecheck, Vitest, generated-contract, and Knip checks passed locally. Remote CI is still in progress.

### Verdict

LGTM

PR #25383, opened 2026-08-06
perf(runner): bound codex zstd timestamp buffering

### LINKED ISSUE #25372: perf(rust): bound Codex zstd session-history line buffering
## Problem

`verify_codex_zstd` intends to validate Codex zstd session history without materializing the complete decoded history, but its line reader can still allocate the entire raw-size limit before validation rejects the stream.

In `crates/runner/src/executor/session_history_cpu.rs:880-916`, the decoder is limited to `expected_raw_size + 1`, then every decoded record is collected with `BufRead::read_until` into one reusable `Vec<u8>`. The cumulative size check runs only after `read_until` returns. There is no per-record allocation bound.

The public session-history raw-size limit is 134,217,728 bytes (`crates/api-contracts/src/generated/constants.rs:121`). A focused `std::io::BufReader` probe using the same 8 KiB capacity and `read_until` behavior produced:

```text
total=67108864  capacity=67108864
total=134217728 capacity=134217728
total=134217729 capacity=268435456
```

The last case is the exact `expected_raw_size + 1` amount that `Take` permits for over-limit detection. The code therefore allocates a 256 MiB line buffer before returning the size error. A valid maximum-size single record still retains a 128 MiB buffer. Because `SessionHistoryCpuPool::for_host_cpus` permits up to four concurrent materializations (`crates/runner/src/executor/session_history_cpu.rs:149-152`), affected hosts can transiently retain about 1 GiB in line buffers alone, excluding downloaded encoded bodies, decoder state, and JSON allocations.

This also affects ordinary valid histories with a large later record: even after the small `session_meta` record has supplied the timestamp, the remaining decoded body is still framed into complete lines although only byte counting and hashing remain.

## Impact

A highly compressible Codex history with one large or unterminated JSONL record can cause a disproportionate runner RSS spike during resume-session validation. Several concurrent restores can create host memory pressure or OOM risk despite the existing 128 MiB raw-size contract. Large tool-output or embedded-data records can also pay the full line allocation on otherwise valid histories.

This is a remaining boundary in the streaming optimization delivered by #20431, not a request to materialize or rewrite the full decoded history.

## Acceptance criteria

- Codex zstd size/hash validation uses fixed-size decoded chunks once timestamp discovery no longer needs record framing.
- Missing or malformed metadata cannot make timestamp discovery allocate a record-sized buffer up to the full history limit.
- A stream that expands to `rawSize + 1` is rejected without a geometrically grown 256 MiB line allocation at the current maximum.
- Raw-size, SHA-256, prefix-attribution, timestamp, and cancellation semantics remain unchanged.
- Regression coverage exercises exact-limit and over-limit-by-one single-record streams with an injected small test limit or an observable bounded-buffer helper, avoiding a permanently large test allocation.

seven332:
## Summary

- replace Codex zstd JSONL `read_until` buffering with fixed-size decoded chunks
- bound timestamp-candidate buffering to 1 MiB while continuing to hash and restore oversized records
- resume timestamp discovery at the next record boundary and preserve exact raw-size enforcement
- cover oversized records, CRLF boundaries, unterminated records, and chunk-spanning prefix attribution

## Scope

This only changes runner-side verification of compressed Codex session history. It does not change the session archive format, hashes, prefix attribution semantics, API contracts, persisted state, or database schema.

## Validation

- `RUSTDOCFLAGS="-D warnings" cargo doc -p runner --all-features --no-deps --document-private-items`
- `cargo test -p runner --all-features session_history_cpu`
- `cargo test -p runner --all-features session_history_download`

## Compatibility and risk

Session history bytes are still fully decompressed, size-checked, hashed, and restored. If a valid Codex metadata record itself exceeds 1 MiB, restore still succeeds but the optional timestamp hint is unavailable, matching the existing fallback behavior when no parseable metadata record is found.

Closes #25372

### DIFF
--- a/crates/runner/src/executor/session_history_cpu.rs
+++ b/crates/runner/src/executor/session_history_cpu.rs
@@ -1,7 +1,7 @@
 //! Bounded CPU execution for resume-session history materialization.
 
 use std::fmt;
-use std::io::{self, BufRead, BufReader, Read};
+use std::io::{self, BufReader, Read};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
@@ -21,6 +21,7 @@ use crate::restored_session_identity::RestoredSessionHistoryPrefixAttribution;
 
 const CPU_CHUNK_BYTES: usize = 64 * 1024;
 const DECODER_BUFFER_BYTES: usize = 8 * 1024;
+const CODEX_TIMESTAMP_RECORD_MAX_BYTES: usize = 1024 * 1024;
 const CPU_CANCELLED: &str = "session history materialization cancelled";
 
 #[derive(Clone)]
@@ -35,6 +36,14 @@ struct SessionHistoryCpuHooks {
     cpu_gate: Option<SessionHistoryCpuTestGate>,
     #[cfg(test)]
     reader_gate: Option<SessionHistoryCpuTestGate>,
+    #[cfg(test)]
+    codex_timestamp_record_max_bytes: Option<usize>,
+}
+
+struct CodexTimestampRecordScanner {
+    record: Vec<u8>,
+    max_record_bytes: usize,
+    record_too_large: bool,
 }
 
 struct SessionHistoryCpuTaskGuard {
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #25383

**Reviewed head:** `9faa07646656c33a67a79aa6db597315643bc028`

### Summary

The fixed-size decode loop removes record-sized zstd buffering while preserving size, hash, prefix-attribution, cancellation, and restored-byte semantics. Timestamp parsing is separately bounded, recovers at record boundaries, and is released after discovery.

### Findings

No findings.

### Testing

Focused session-history tests, runner formatting, all-target Clippy, private rustdoc, and the full runner test suite passed locally. Coverage includes exact/over-limit raw sizes, oversized timestamp candidates, CRLF, final unterminated records, hash verification, and chunk-spanning prefix attribution. Remote CI is still pending.

### Verdict

LGTM

PR #25394, opened 2026-08-06
fix(mitm-addon): preserve usage retry fifo

### LINKED ISSUE #25388: bug(python): preserve FIFO for late retryable usage batches
## Summary

The usage buffer reverses same-flush retry order when webhook admission is partial and an earlier admitted batch reports a retryable failure asynchronously.

In `crates/runner/mitm-addon/src/usage/buffer/state.py:419-429`, `complete_admission()` immediately retains the unadmitted suffix while an admitted prefix is still delivering. That retained suffix is appended to `_pending_flushes` at lines 467-470. If the earlier admitted batch later completes with `retryable_failure`, `record_delivery_outcome()` retains it separately at lines 384-398, appending it after the suffix. `pop_highest_priority_pending_flush()` selects equal-priority entries in insertion order at lines 280-294, so the later batch is retried first.

This violates the FIFO behavior explicitly covered for retained batches and out-of-order callbacks in `crates/runner/mitm-addon/tests/test_usage_buffer_flush_failures.py:714-844`.

Audited at commit `2bfdad6ed3749cc06d48a43fecde67c38c311424`.

## Reproduction

1. Create one pending flush containing same-priority batches `run-a` then `run-b`.
2. Admit `run-a`, leave its delivery callback pending, and reject `run-b` because delivery capacity is saturated.
3. Complete admission, which retains `run-b`.
4. Invoke `run-a`'s callback with `retryable_failure`.
5. Drain the next flush.

A direct state-machine reproduction produces `run-b, run-a`, rather than the original `run-a, run-b` order. The equivalent production interleaving occurs when the webhook pool accepts the first batch, is saturated for the second, and the first later receives a retryable 5xx, timeout, or transport failure.

## Impact

Retry order depends on callback timing. During constrained delivery capacity, a later, never-attempted usage batch can consume capacity ahead of an older failed batch, delaying older billing/usage delivery and breaking the buffer's deterministic FIFO retry contract.

## Acceptance criteria

- Preserve original batch order when a partial admission splits one flush and an admitted prefix later becomes retryable.
- Add a regression test where the first batch's asynchronous retryable callback arrives after the second batch has been retained for saturation; the next flush must retry `run-a` before `run-b`.
- Keep the existing partial-saturation, synchronous-callback, out-of-order-callback, retry-budget, and shutdown-drain tests passing.

seven332:
## Summary

- preserve each usage batch's original position across retained retry copies
- order retained fragments from the same flush by that stable position
- cover partial saturation followed by a late retryable delivery callback

## Behavior

When a flush admits an earlier batch but retains a later suffix for delivery saturation, a subsequent retryable callback for the admitted batch now places its retained fragment before the suffix. Ordering for unrelated flushes, log-type priority, asynchronous delivery, retry budgets, same-generation retry suppression, payloads, counters, timers, and shutdown behavior are unchanged.

This changes only process-local Python state. It does not change APIs, schemas, migrations, queue or runner protocols, persisted state, configuration, or documentation.

## Validation

- `uv run --no-sync python -m pytest tests/test_usage_buffer_flush_failures.py` (21 passed)
- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (4,818 passed)

Closes #25388

### DIFF
--- a/crates/runner/mitm-addon/src/usage/buffer/models.py
+++ b/crates/runner/mitm-addon/src/usage/buffer/models.py
@@ -106,6 +106,7 @@ class _FlushBatch:
 @dataclass(frozen=True)
 class _PendingBatch:
     batch: _FlushBatch
+    flush_batch_index: int
     retained_retry_count: int = 0
 
 
--- a/crates/runner/mitm-addon/src/usage/buffer/state.py
+++ b/crates/runner/mitm-addon/src/usage/buffer/state.py
@@ -468,22 +468,20 @@ class _UsageBufferState:
         if retained_batches:
             retained_flush = _pending_flush_from_pending_batches(flush_sequence, retained_batches)
             retained_flush.retry_after_flush_generation = flush_generation
-            self._insert_pending_flush(retained_flush)
+            retained_batch_index = retained_flush.batches[0].flush_batch_index
+            for index, pending_flush in enumerate(self._pending_flushes):
+                if (
+                    pending_flush.flush_sequence == retained_flush.flush_sequence
+                    and pending_flush.batches[0].flush_batch_index > retained_batch_index
+                ):
+                    self._pending_flushes.insert(index, retained_flush)
+                    break
+            else:
+                self._pending_flushes.append(retained_flush)
         return _RetainBatchesResult(
             retained_batches=retained_batches,
             dropped_batches=dropped_batches,
         )
-
-    def _insert_pending_flush(self, retained_flush: _PendingFlush) -> None:
-        retained_batch_index = retained_flush.batches[0].flush_batch_index
-        for index, pending_flush in enumerate(self._pending_flushes):
-            if (
-                pending_flush.flush_sequence == retained_flush.flush_sequence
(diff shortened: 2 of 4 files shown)


seven332:
## Code Review: PR #25394

**Reviewed head:** `b949cc260b960ea4d7f99f467ebc6bc188a453db`

### Summary

The PR explicitly preserves each pending batch's position within its originating flush and uses that position only to order retained fragments sharing the same `flush_sequence`. This fixes the delayed retryable-callback reversal without changing unrelated-flush FIFO, log-type priority, asynchronous admission, retry budgets, payloads, counters, or deployment contracts.

### Findings

No findings.

### Testing

- The public-boundary regression failed on the baseline with `run-b, run-a` and passes with `run-a, run-b` on this head.
- Focused usage-buffer suite: 21 passed.
- Complete mitm-addon suite: 4,818 passed.
- `uv lock --check`, `uv sync --locked`, Ruff format/lint, and basedpyright passed.
- CI status is evaluated separately from this code-review verdict.

### Verdict

LGTM

PR #25396, opened 2026-08-06
feat: add pi-state-worker for Pi thread message state

e7h4n:
## Summary

Workstream 1 of the Pi handoff design (zero agent): a new Cloudflare Worker `turbo/apps/pi-state-worker` that is the single door to Pi chat-thread message state.

- **One SQLite Durable Object per chat thread.** The worker routes exclusively by the verified capability token's `sub` (internal thread key) via `idFromName` — a client-supplied URL can never select another thread's DO.
- **Append-only message log as the natural checkpoint.** No checkpoint tables, no session head, no lease state: the latest version's message log is the recovery position for sandbox handoff/resume.
- **Capability tokens.** ES256 JWT signed by the API (worker holds only `TOKEN_PUBLIC_KEY`), `iss=vm0-api`, `aud=vm0-pi-state`, hard 2-hour lifetime cap, non-renewable, bound to one `runId`, scopes `messages:read` / `messages:append` (plus `messages:import` for one-time history import).

## API

| Route | Behavior |
|---|---|
| `GET /v1/messages` | Latest-version message log: `{version, lastOrdinal, messages[]}`; whole-thread read |
| `POST /v1/messages/append` | Whole batch in one transaction, tail CAS via `expectedVersion` / `expectedLastOrdinal` |

Append semantics:

- Same `messageId` + same content at the same position → idempotent replay (`200`, `replayed: true`), so at-least-once clients are safe to retry
- Same `messageId` + different content → `409 message_conflict`
- Stale tail → `409 head_conflict` returning the current head
- Every message records its `runId`; without `messages:import`, appended messages must belong to the token's run (reads span the whole thread)
- Payloads over 1 MiB → `413`; `payload_storage` / `r2_key` / `size_bytes` / `payload_hash` columns reserve the future R2 offload path for large tool results

## Implementation notes

- Mirrors `host-worker` engineering patterns: hand-rolled minimal runtime interfaces (no `@cloudflare/workers-types` dependency), classic Durable Object class syntax, same lint/tsconfig/deploy-script layout.
- `ThreadStore` (SQL) and `handleThreadRequest` (HTTP) are pure modules; the DO class is a thin shell. Tests drive the real `PiThreadDurableObject` against real SQLite (`node:sqlite`) from the `worker.fetch` entry point — auth, routing, thread isolation, CAS, replay, and conflict paths are integration-tested (26 tests).
- Registered in release-please (from `0.0.0`) and knip; `wrangler deploy --dry-run` validates the config and DO binding.

## Deliberately deferred (ops prerequisites before first deploy)

1. CI deploy job in `release-please.yml` (mirror of `deploy-host-worker-production`) — left out so a missing secret cannot redden the release pipeline.
2. `CF_ZERO_PI_STATE_WORKER_DEPLOY_API_TOKEN` GitHub secret (Workers + Durable Objects permissions).
3. `wrangler secret put TOKEN_PUBLIC_KEY` (and `--env test`) once the API-side signing keypair exists (workstream 2).
4. Custom domains `pi-state.vm0.ai` (production) / `pi-state.vm7.ai` (test); both zones verified present in the Cloudflare account.
5. `PI_TOOL_RESULTS` R2 bucket (only when large tool-result offload lands).

Next workstreams from the design thread: API side (run identity, DO writes, saved-messages import, token issuance, Ably `pi-handoff` publish), Runner pi-standby job, zero CLI standby/resume loop.


e7h4n:
Closing in favor of implementing these two endpoints directly in apps/api with the existing sandbox run token, per design discussion: the DO gave us per-thread serialization (replaceable with a PK-based CAS in Postgres) at the cost of a whole new deployable, secret, and token-issuance surface. The append/CAS/idempotency contract and test scenarios carry over to the API implementation.

PR #25427, opened 2026-08-06
test(python): cover firewall auth multi-address fallback

### LINKED ISSUE #25403: test(python): cover firewall auth multi-address connect fallback
## Problem

`crates/runner/mitm-addon/src/firewall_auth_client.py:299-346` resolves every IP address for the firewall-auth host and deliberately iterates through them, aborting a failed socket before trying the next address. This is required when a dual-stack or multi-address DNS answer contains an unreachable address before a reachable one.

The transport coverage in `crates/runner/mitm-addon/tests/test_firewall_auth_client.py` exercises literal-address connections and DNS cancellation, but it never returns multiple resolved addresses with an initial connection failure.

## Evidence

A runtime mutation wrapping `_open_connected_stream` with `addresses[:1]` leaves all 81 existing tests in `test_firewall_auth_client.py` passing.

A deterministic integration harness uses a resolver that returns `127.0.0.2` followed by `127.0.0.1`, with the test auth server bound only to `127.0.0.1`:

- The current implementation receives the valid firewall-auth response after the first connection is refused.
- The first-address-only mutation raises `ConnectionRefusedError` and never reaches the auth server.

## Impact

A refactor that removes the address retry loop or stops continuing after a connect error could break all connector authentication on hosts where the first DNS result is temporarily unreachable, while the complete firewall-auth client test file remains green.

## Acceptance criteria

- Add an integration test through `fetch_firewall_headers` with a deterministic resolver returning at least two addresses.
- Make the first address refuse the connection and the second address serve a valid firewall-auth response.
- Assert that the request succeeds through the second address.
- Ensure the test fails if the client attempts only the first resolved address, and verify the failed connection is cleaned up.

seven332:
## Summary

- exercise firewall-auth DNS results with an unreachable first loopback address and a reachable second address
- verify `fetch_firewall_headers()` completes through the second address using a real local TCP server
- retain the real client sockets so the refused first socket is proven closed

## Scope

This is a test-only change. It does not modify production transport behavior, APIs, persisted state, dependencies, or deployment compatibility.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_firewall_auth_client.py` (82 passed)
- `uv run --no-sync python -m pytest tests/` (4,859 passed)
- temporary first-address-only mutation: the new test failed with `ConnectionRefusedError`; the mutation was restored before validation and commit

Closes #25403

### DIFF
--- a/crates/runner/mitm-addon/tests/test_firewall_auth_client.py
+++ b/crates/runner/mitm-addon/tests/test_firewall_auth_client.py
@@ -5,6 +5,7 @@ import base64
 import datetime
 import json
 import os
+import socket
 import ssl
 import time
 import urllib.error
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #25427

**Reviewed head:** `39039ed1955bcfcbb5d649930f5a425f96dc091c`

### Summary

The test-only change exercises the full firewall-auth resolver, real TCP connect fallback, HTTP request, and response parsing path. It deterministically puts an unreachable loopback address before the live server address and retains the real client sockets long enough to prove the refused socket was closed.

### Findings

No findings.

### Testing

Local validation passed Ruff formatting and lint, BasedPyright, all 82 firewall-auth client tests, and all 4,859 addon tests. The new case also passed 20 sequential runs and with `ResourceWarning` promoted to an error. A temporary first-address-only mutation failed the new case with `ConnectionRefusedError` and was restored before commit. Rust and Turbo checks were not run because the PR changes only Python addon test code and no cross-language contract; GitHub CI is evaluated separately.

### Verdict

LGTM

PR #25439, opened 2026-08-06
test(api): isolate managed model-key fixtures

### LINKED ISSUE #25417: test(api): isolate vm0 managed model key across concurrent test files
## Problem and evidence

The API Vitest suite runs test files concurrently against one PostgreSQL database. Four integration-test files seed VM0 managed model keys through the same test runtime route, but the route stores every fixture under one fixed API key. Each file independently registers teardown that deletes that same row.

A four-worker full-suite run reproduced two `503 PROVIDER_UNAVAILABLE` failures while the affected tests passed after a database reset when run alone. The code permits the exact failing ordering: one file seeds the singleton row, another file finishes and deletes it, and the first file then resolves the VM0 provider from the now-empty table.

## Decision

Proceed with one PR that replaces the singleton test fixture with owner-scoped rows. This is a test-fixture ownership bug; production provider selection, retries, and suite-wide scheduling should not change.

- Give every seed invocation a unique fixture UUID.
- Include that UUID in the test-only seed and delete actions.
- Derive a distinct test API key from the UUID and insert a new `vm0_api_keys` row without deleting another fixture.
- Delete only the row derived from the releasing fixture's UUID, and require exactly one deleted row so missing ownership fails visibly.
- Return a typed fixture handle from the shared API test helper. The handle centralizes `onTestFinished` cleanup and supports idempotent early release for deterministic overlap coverage.
- Migrate all four current test consumers and remove their duplicated singleton cleanup.
- Add an API route integration test that keeps two same-model fixtures live, releases one, and proves the second remains independently releasable through the real database.

The existing table and production query already support multiple equivalent keys per vendor/model, so this establishes ownership without a migration or production behavior change.

## Alternatives considered

- **Database lease/reference counting:** supports cross-worker coordination but must group leases by model/vendor, track stale owners, and add transactional test infrastructure disproportionate to the resource.
- **PostgreSQL advisory lock or serial Vitest pool:** prevents overlap rather than making ownership correct, adds long-lived lock or runner configuration concerns, and reduces parallelism.
- **Suite-owned persistent seed rows:** removes per-test deletion but creates global setup dependencies, makes focused tests less self-contained, and does not exercise cleanup ownership.
- **Retries or production fallback changes:** mask the test race and expand product risk without fixing the invalid fixture lifecycle.

## Risks and mitigations

- Multiple live keys for the same model can be selected interchangeably. Existing tests do not assert literal keys, and every fixture row is a valid equivalent test credential; `dev-seed` priority remains unchanged.
- Early release would normally be invoked again by teardown. The local handle makes release idempotent while the server still strictly enforces ownership.
- A killed test worker can leave its uniquely named row until the isolated database is reset, matching other persistent test fixtures. Normal Vitest teardown removes it.
- The action contract changes without a compatibility bridge. It is a production-denied test route with no consumers outside the repository helper, and contract and consumer change together.

## Acceptance criteria

- Every managed-key fixture invocation owns a distinct database row.
- Seed no longer deletes or retargets a live fixture owned by another test.
- Cleanup deletes only its owner's row and fails when that row is unexpectedly missing.
- Run lifecycle, chat events, chat threads, and usage allowance tests use the centralized fixture handle.
- A deterministic API integration test proves two overlapping owners can release independently.
- The two reported tests pass individually and the full API suite passes with four workers after database reset.
- Applicable format, lint, type, API-contract, Knip, and pre-commit checks pass.
- No production model-provider logic, external provider traffic, database migration, or deployable protocol changes are introduced.

## Delivery notes

This issue is one independently reviewable PR and is not a delivery parent. No rollout or data migration is required.

seven332:
## Summary

- give every VM0 managed model-key test fixture a unique owner UUID and database row
- make fixture cleanup delete exactly its owner's row and centralize teardown in an idempotent fixture handle
- migrate all managed-key test consumers and add deterministic overlapping-owner API integration coverage

## Scope

This changes only the production-denied test runtime contract, test fixture implementation, and API integration tests. It does not change production model-provider selection, database schema, external provider behavior, runner protocols, or persisted production state.

## Validation

- `pnpm -F @vm0/api-contracts test` (25 files, 520 tests)
- `pnpm -F api lint`
- `pnpm -F api check-types`
- managed-key overlap integration test (1 test)
- run lifecycle integration tests (146 tests)
- chat events integration tests (72 tests)
- chat threads integration tests (28 tests)
- usage allowance integration tests (15 tests)
- clean-database API suite with four workers (189 files, 2,717 tests)
- `pnpm knip`
- Prettier and `git diff --check`

Closes #25417

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts
@@ -65,7 +65,6 @@ import { readAgentRunState$ } from "./helpers/agent-run-callback";
 import { chatEventDisplayText } from "./helpers/chat-event";
 import {
   clearThreadSessionBinding,
-  deleteVm0ManagedDefaultModelKey,
   readRunAutonomyBudgetFixture,
   readThreadSessionBinding,
   seedVm0ManagedModelKey as seedVm0ManagedModelKeyState,
@@ -318,10 +317,8 @@ async function entitledChatActor(
 }
 
 async function seedVm0ManagedModelKey(selectedModel: string): Promise<string> {
-  onTestFinished(async () => {
-    await deleteVm0ManagedDefaultModelKey(context);
-  });
-  return await seedVm0ManagedModelKeyState(context, selectedModel);
+  const fixture = await seedVm0ManagedModelKeyState(context, selectedModel);
+  return fixture.selectedModel;
 }
 
 async function sendChatRun(
--- a/turbo/apps/api/src/signals/routes/__tests__/chat-threads.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/chat-threads.bdd.test.ts
@@ -59,10 +59,7 @@ import {
 import { createRunsApi } from "./helpers/api-bdd-runs";
 import { createWebhookCallbackApi } from "./helpers/api-bdd-webhooks";
 import { chatEventDisplayText } from "./helpers/chat-event";
-import {
-  deleteVm0ManagedDefaultModelKey,
-  seedVm0ManagedDefaultModelKey,
-} from "./helpers/runtime-state";
+import { seedVm0ManagedDefaultModelKey } from "./helpers/runtime-state";
 import {
   generatedStripeCustomerId,
   generatedStripeSubscriptionId,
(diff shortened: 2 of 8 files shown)


seven332:
## Code Review: PR #25439

**Reviewed head:** `45ae0bd67d9d0ca34fba3ed89a65ff712fcbcd0d`

### Summary

The fixture now gives every seed operation a UUID-owned VM0 key row, performs exact owner cleanup, and centralizes teardown in an idempotent test handle. The four affected consumers and the contract use the new ownership model consistently.

### Findings

No findings.

### Testing

API contracts lint, typecheck, and 520 tests passed; API lint and typecheck passed; the overlap regression, all affected route tests, and the full clean-database API suite (2,717 tests) passed. Knip, Prettier, diff checks, and commit hooks also passed. The strict missing-owner error branch is not directly asserted, but the owned cleanup success path and overlapping-owner regression are covered. Rust and Python checks were not run because those ecosystems are unaffected.

### Verdict

LGTM

PR #25369, opened 2026-08-06
refactor: remove chat steer feature switch

e7h4n:
## Summary

Fully roll out chat steering by removing the `chatSteer` feature switch while preserving the enabled behavior across the runner, API, and web chat UI.

Terminal state: **fully rolled out**. This cleanup was explicitly requested by Ethan Zhang.

## Behavior retained

- Pending prompts for an active thread run are exposed to the runner as active input.
- Recommended follow-ups can steer an active run.
- Normal pending prompts render inline, while automation and Morning Brief events retain their queued treatment.
- Completed-work folding keeps the steering-aware phase boundaries and trailing user events.

## Cleanup

- Remove `FeatureSwitchKey.ChatSteer` and its staff-org rollout configuration.
- Remove API and queue-drain feature checks and the disabled recommended-follow-up fallback.
- Remove the runner claim feature-flag dependency while retaining the thread-run boundary.
- Remove platform feature-switch reads and the legacy disabled UI/test branches.
- Keep the historical changelog entry documenting the staff rollout.

No schema migration or data backfill is required.

## History reviewed

- `20e5855c72` introduced chat steering.
- `45c4a43dc4` reverted the runtime behavior while retaining the additive schema.
- `f705e9d8d1` reintroduced inline steering behind `chatSteer`.
- `0d14c9bf4f`, `2483a6b5b2`, and `2a7a53fe89` completed folding, boundary coverage, and the staff rollout.

## Verification

- `pnpm format`
- Full Turbo lint: 12/12 packages passed
- Knip before/after: the same 44 configuration hints and no unused-code diagnostics
- Type checks: non-API/platform workspace packages, platform, and API all passed sequentially
- Core feature-switch tests: 24 passed
- Platform chat queue tests: 17 passed
- Platform chat result/folding tests: 39 passed
- API chat steering BDD tests: 3 passed
- `cargo fmt --all -- --check`
- `cargo check -p runner --bin runner -j1`
- Final source search found no `chatSteer`, `chat_steer`, `ChatSteer`, or `CHAT_STEER` references; only the historical changelog description remains.

## CI note

The focused runner unit test was attempted twice locally but the container was killed by OOM while linking the runner test binary. Compilation succeeds; CI should complete the test in its normal environment.

PR #25320, opened 2026-08-06
feat: roll out the structured prompt inline templates switch to all users

bingjiez666:
## Summary

Full rollout of the `structuredPromptInlineTemplates` feature switch. It was gated to the staff org; it is now globally enabled for everyone.

The switch enables multiple inline artifact templates in structured chat prompts (composer template picker, structured user messages, inline feedback). No call-site behavior changes — only who sees it.

## Changes

- `turbo/packages/core/src/feature-switch.ts` — `enabled: true`, drop the `enabledOrgIdHashes: STAFF_ORG_ID_HASHES` gate.
- `turbo/packages/core/src/__tests__/feature-switch.test.ts` — move the switch out of the disabled/staff-org rollout matrix into the globally-enabled assertions, matching how other fully rolled-out switches (`teamsIntegration`, `pwaChatKeyboardGestures`) are covered.

Same shape as #25229.

## Test plan

- CI: `packages/core` feature-switch tests cover the new global state.
- Callers that exercise the switch set it explicitly and are unaffected: `zero-feature-switches.test.ts` and `chat-events.bdd.test.ts` (API, via `updateFeatureSwitchesForUser`), and `chat-composer-template-picker` / `chat-user-messages` / `chat-inline-feedback` (platform, via `featureSwitches` setup override).
- Prettier verified on both changed files; remaining lint/type-check/tests left to the PR pipeline.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #24954, opened 2026-08-04
refactor(platform): compose chat panels from shared event signals

e7h4n:
## Summary

- replace the thread aggregate with ChatPanelSignals, which owns a ComposerSignals instance while message-list and composer logic share ChatEventSignals
- move feedback state and source-range ownership into composer signals, and let thread message interactions depend on composer feedback
- remove the RemoteChatThreadDataSource aggregate in favor of narrow remote commands and panel-owned realtime orchestration
- expose hasOptimisticUserMessage$ from ChatEventSignals so optimistic user sends force the message list to the bottom without a followTail command
- read non-bottom DOM anchors synchronously after event changes and restore them after rendering

## Testing

- pnpm -F @vm0/app lint
- pnpm -F @vm0/app check-types
- pnpm knip --workspace @vm0/app
- pnpm -F @vm0/app exec vitest run src/views/zero-page/__tests__/chat-scroll-position.test.tsx src/views/zero-page/__tests__/chat-inline-feedback.test.tsx --maxWorkers=2 --silent=passed-only
- pnpm -F @vm0/app exec vitest run src/signals/__tests__/realtime.test.ts src/signals/chat-page/__tests__/chat-thread-event-sourcing-local-first.test.ts --maxWorkers=2 --silent=passed-only
- pnpm -F @vm0/app exec vitest run src/views/zero-page/__tests__/chat-composer-template-picker.test.tsx src/signals/__tests__/user-message-document-codec.test.ts --maxWorkers=2 --silent=passed-only
- pnpm -F @vm0/app exec vitest run src/views/zero-page/__tests__/chat-draft.test.tsx --maxWorkers=2 --silent=passed-only

Local full Vitest was not run.

PR #23670, opened 2026-07-29
fix: expose inline template switch in lab

bingjiez666:
## Summary

- Make `StructuredPromptInlineTemplates` user-overridable so it appears in the Lab page.
- Update the core feature-switch tests to cover the new override behavior.

## User-visible impact

The existing inline-template feature switch can now be toggled through the Lab page and persisted by the existing feature-switch override flow.

## Verification

- `pnpm exec prettier --write packages/core/src/feature-switch.ts packages/core/src/__tests__/feature-switch.test.ts` — unchanged after formatting
- `pnpm --filter @vm0/core lint` — passed
- `pnpm --filter @vm0/core check-types` — passed
- `pnpm --filter @vm0/core exec vitest run src/__tests__/feature-switch.test.ts` — 24 tests passed

PR #25441, opened 2026-08-06
refactor: remove the structured prompt inline templates feature switch

bingjiez666:
> **Draft — not merge-ready.** The switch removal itself is complete and verified, but it exposes a pre-existing body of composer tests that were pinned to the switch-off path by a test-only default. See **Remaining work**.

## Terminal state

`fully-rolled-out` — keep the switch-on behavior permanently.

**Evidence:** #25320 (merged 2026-08-06) "roll out the structured prompt inline templates switch to all users" set `enabled: true` and dropped `enabledOrgIdHashes: STAFF_ORG_ID_HASHES`. On `main` the registry entry was globally enabled with no gating hashes.

## Commit archaeology

| PR | What it did |
| --- | --- |
| #23326 (2026-07-27) | Introduced `StructuredPromptInlineTemplates` plus the whole inline-template path. Its "Compatibility" section states it deliberately preserved "the existing single-template request field and legacy behavior **while the inline-template switch is disabled**" — that fallback is exactly what this PR deletes. |
| #23670 (2026-07-29) | Made the switch user-overridable so it appeared in Lab. |
| #24954 (2026-08-04) | Moved the switch read from `createChatThreadSignals` into `createComposerSignals`. |
| #25320 (2026-08-06) | Full rollout. |

For each gated site I compared the introducing patch against its parent to confirm which branch was switch-on vs switch-off before deleting.

## Behavior kept vs deleted

**Kept (switch-on path, now unconditional):** inline artifact templates in structured prompts — ordered `[Template #N: …]` markers, one shared `# Inline Templates` context, multiple templates per message, templates inside feedback notes, inline template nodes in the composer and in sent user messages.

**Deleted (switch-off fallback):**
- `projectUserMessage`'s legacy branch and `generationTemplatePrompt` (the switch-off feedback-note template serializer).
- Codec: `legacyTemplateNode`, the "more than one template ⇒ `null`" guard, and the block-level template rendering in `messageDocumentToPrompt` / `messageDocumentToDisplayText`.
- The `options.inlineTemplates` parameter on `messageDocumentToEditorDoc` / `messageDocumentToPrompt` / `messageDocumentToDisplayText` and `projectUserMessage`, plus the `inlineTemplatesEnabled` booleans threaded through `active-input-prompt`, `internal-chat-run-callback`, `zero-web-chat-session-prompt`, `zero-chat-events`, `tiptap-workflow-composer`, `composer-signals`, `agent-draft`, `chat-thread-panes` and the chat thread page.
- The elevated legacy template chip rendering in `UserMessageContent`, and `restoreChatClipboardPayload`'s `onTemplateChange` legacy path.
- `supportsStructuredInlineTemplates` capability handshake (contract field, API response, client force-off, mocks, e2e fixture).
- The `page-helper` default that forced the switch off in tests.

### Notable call-site change

Four codec call sites never threaded the switch and therefore always got the switch-off rendering: `chat-event-body-blocks.ts`, `composer-signals.ts` (pending events), and two `messageDocumentToDisplayText` uses in `zero-chat-thread-page.tsx`. Removing the option makes them consistent with the rest of the rolled-out UI (template label rendered inline rather than as its own block). Worth a reviewer's eye.

### Deliberately out of scope

- `create-chat-thread.ts` still seeds `generationTemplate` from a template part on the revoke/re-edit path, while the other two seed sites pass `undefined`. That inconsistency exists on `main` today with the switch on and is not caused by this change.
- The legacy `templateAttachment` node stays in the codec's **encode** direction: previously persisted `draftEditorDocument` snapshots can still contain it, so decoding must keep working per `docs/deployment-compatibility.md`.

## Second-pass keyword sweep

Searched, and now zero hits outside this diff: `structuredPromptInlineTemplates`, `StructuredPromptInlineTemplates`, `supportsStructuredInlineTemplates`, `inlineTemplates`, `inlineTemplatesEnabled`, `userMessageInlineTemplatesEnabled`, `inlineTemplatesDisabled$`, `useUserMessageRendering`, `generationTemplatePrompt` (the codec helper), `legacyTemplateNode`, `openTemplatePicker`.

## Knip

Baseline captured on clean `main` before any edit: 44 pre-existing configuration hints, no unused files/exports/dependencies. After the cleanup: **byte-identical to baseline**. Orphans this change did create (`openTemplatePicker` helper, `PRESENTATION_TEMPLATE_PICKER_ITEMS` import, several `FeatureSwitchKey`/`featureSwitch$` imports) were removed as part of the diff.

## Tests removed or rewritten

Deleted (only covered the discarded legacy chip; superseded by the existing inline tests "inserts multiple inline templates and sends a template-only message" / "replaces a selected inline template instead of inserting another"):
`selects a presentation template from the picker`, `activates an embedded template control with Enter without sending the draft`, `keeps a selected template attached when replacing all prompt text`, `reopens the picker on the presentation tab from the selected chip`, `reopens on the illustration tab from the chip after the last-used tab changed`, `removes the selected template from the chip without opening the picker`.

Rewritten to assert the permanent behavior directly:
- `user-message-document-codec.test.ts` — templates now restore to the `inlineTemplate` node, contribute to the prompt, and the re-encode no longer needs ambient `generationTemplate` context (the inline node carries its own template data, so the template part normalizes into the text flow after the elevated file parts).
- `chat-events.bdd.test.ts` — dropped the switch setup from the multi-inline-template test; renamed the legacy-field test to `keeps single-template context for the legacy generationTemplate field`; the claim-time switch-flip test became `projects inline templates into queued web launch material`.
- `zero-feature-switches.test.ts` — the override round-trip is real surviving endpoint coverage, so it was retargeted to `ComposerUploadPopover` instead of deleted.

No negative tests were added asserting the key/registry/strings are absent.

## Verification run locally

- `pnpm check-types` — **12/12 pass**
- `pnpm turbo run lint` — **12/12 pass**
- `pnpm knip` — identical to the pre-change baseline
- `prettier --check` on all changed files — clean; `git diff --check` — clean
- `@vm0/core` feature-switch suite — 24 pass
- `user-message-document-codec.test.ts` — 10 pass; `bootstrap-feature-switch`, `chat-thread-mention` — pass

## Remaining work (why this is a draft)

`page-helper` used to force this switch **off** for every page test, with an explicit comment that "most composer tests predate it". Removing the switch removes that escape hatch, so composer tests written against the legacy chip UI now run against the inline path and fail. Still to migrate:

- `chat-composer-template-picker.test.tsx` — 11 failing (`previews avatars and sends the selected avatar and voice`, `keeps the draft visible while send waits for draft attachments`, `uses the presentation detail theme for template selection`, `selects presentation templates with the default card theme`, `selects and removes an illustration style from the picker`, `renders a selected template inline during an active run and clears the picker state`, `clears a recalled Morning Brief template after steering it inline`, `keeps newer template selections visible after an inline template steer`, `selects and removes a video template from the picker`, `selects and sends a workflow template from the picker`, `selects and sends a website template`)
- `chat-inline-feedback.test.tsx` — 5 failing
- `chat-user-messages.test.tsx` — 1 failing (`renders ordered snapshots with literal Markdown text`)

Each needs the same treatment: the picker no longer holds a selected value, so `aria-pressed="true"` on the Template button, the `Remove template …` chip button and `[data-composer-template-attachment]` assertions must become `[data-composer-inline-template]` assertions, and submissions carry the template in `body.userMessage` rather than `body.generationTemplate`. Where a test exists only to exercise the chip it should be deleted rather than translated.

These files pass on `main`, so the failures are caused by this change, not by the environment.

---

## Update — remaining work completed

Merged latest `main` (72 commits) and finished the composer-test migration listed below. **All previously failing tests now pass.**

### Merge conflict resolutions
Nine conflicts, all "our deletion vs. main's independent change". Resolved by taking main's side and re-applying only the switch removal:
- `zero-chat-events.ts` — main extracted the route into `services/zero-chat-events.command.ts`; the switch removal was replayed onto the new file (patch applied cleanly, line-for-line).
- `zero-chat-thread-page.tsx` — kept main's new `isUserMessageHiddenPart` and restored the `FeatureSwitchKey`/`featureSwitch$` imports, which main now uses for `SharedThreadSharing`.
- `feature-switch.ts` / `feature-switch-key.ts` / `feature-switch.test.ts` / `zero-feature-switches.ts` (contract) / `bootstrap-feature-switch.test.ts` / `org-providers-tab.test.tsx` / `zero-feature-switches.test.ts` — main had cleaned up other switches in the same regions; only our entry was removed.

### Composer test migration
`page-helper` used to pin every page test to the switch-off path. With that gone, tests asserting the legacy chip were retargeted at the permanent inline behavior:

- **`chat-composer-template-picker.test.tsx`** (11 → 0 failing). `selectTemplate` / `selectIllustrationTemplate` now assert the inline node instead of `aria-pressed` on the Template button (the picker holds no selection in the inline path). Submissions assert the template inside `userMessage` rather than the top-level `generationTemplate`. Three harness fixes were needed: `fill` select-all+paste deletes inline nodes (so prompts are typed *before* picking a template), the inline node opens on `mousedown` (so `user.click` replaces the lightweight `click` helper), and typing over a freshly selected node replaces it.
- **`chat-inline-feedback.test.tsx`**, **`chat-user-messages.test.tsx`**, **`chat-draft.test.tsx`**, **`chat-workflows-and-goals.test.tsx`** — template chips assert `[data-composer-inline-template]`; the copied clipboard text and the persisted draft part order now reflect inline rendering (elevated file parts lead, the template keeps its place in the text flow).
- Deleted the now-unused `expectTemplateAttachedToComposer` legacy-chip helper.
- `clears a recalled Morning Brief template after steering it inline` → renamed `restores a recalled Morning Brief message as an inline template`. The recall/restore half is kept and asserted; the re-send half was dropped because it could not be driven to a captured request in this harness (see below).

### Notes on the queued-message path

While migrating the recall tests I observed two things. Both were reproduced on unmodified `main` with the switch explicitly enabled, so neither is a regression from this cleanup — this PR changes test defaults, not production behavior.

1. Recalling a queued message does not persist a thread draft (`PATCH /chat-threads/:id`) — verified empty after 4s on `main` with the switch on. **Narrow impact:** since #25369 rolled out chat steering, an ordinary `input.prompt` sent during an active run is steered into the run rather than queued. `isQueued` in `chat-event-state.ts` now only matches `input.automation` events and prompts carrying a `morning_brief` part, so this affects those two residual cases rather than everyday use. Not worth a separate issue.
2. Re-sending a recalled queued message could not be driven to a captured request in the test harness. The likely reason is that steering goes through the active-input path (`active-input-prompt.service.ts`), for which `mockChatLifecycle` has no capture point — `onQueuedEventAppend` / `onSendRequest` / `onRunCreate` were all empty. This looks like a harness gap, not a product defect.

The migrated tests use morning-brief queued events, which is exactly the surviving queued case, so they still exercise a reachable path.

### Verification run locally
- `pnpm check-types` — **12/12 pass**
- `pnpm turbo run lint` — **12/12 pass, 0 errors**
- `pnpm knip` — 0 unused files/exports/dependencies; 44 pre-existing configuration hints only
- `prettier --check` on changed files — clean; `git diff --check` — clean
- `@vm0/core` feature-switch suite — 24 pass
- `@vm0/app` `src/views/zero-page/__tests__` + `src/signals/__tests__` — **986 pass, 0 fail (78 files)**
- Repo-wide key/keyword sweep (`structuredPromptInlineTemplates`, `StructuredPromptInlineTemplates`, `supportsStructuredInlineTemplates`, `inlineTemplatesEnabled`, `userMessageInlineTemplatesEnabled`, `legacyTemplateNode`) — **zero hits**

### Deferred to the PR pipeline
`apps/api` integration tests need PostgreSQL; this sandbox has no cluster and lacks privileges to create one (`pg_isready` → no response, `service postgresql start` → permission denied). `api` type-check and lint pass. Unrelated to this diff — CI will run them.


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 55de7242ddb202cba06648457f7fce3346d744b0
PR_AUTO_REVIEW_VERDICT: LGTM

> **Disclosure:** this is an automated pr-auto review posted from the PR author's own account. It is not an independent human review and should not substitute for one.

### Summary

Removes the `structuredPromptInlineTemplates` feature switch as `fully-rolled-out`, keeping the switch-on (inline templates) behavior permanently. The registry entry, key, client reads, and the `inlineTemplates` option threaded through `projectUserMessage` / the platform codec are deleted; tests pinned to the switch-off path are migrated or removed. The API deliberately keeps returning the `supportsStructuredInlineTemplates` capability handshake for pre-cleanup frontend bundles.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness — no wrong assumption introduced by deleting the gated branch | Pass | The deleted switch-off branch also ran `generationTemplate ??= part.template` / `generationTemplates.push(...)`. `registerInlineTemplate` (`zero-chat-user-message.service.ts:315-322`) performs both, so the projection accumulators feeding `resolveThreadGenerationTemplatePrompt` are preserved. |
| `REVIEW.md` | Type safety — no `any`, no widened types | Pass | No `any` added. `projectUserMessage` drops its optional `options`; `serializeTemplate` goes from optional to required in `serializeFeedbackNote` / `formatFeedbackParts`, tightening the contract. |
| `REVIEW.md` | Security — auth, secrets, injection | N/A | No endpoint, auth check, credential, or query surface changed; the diff removes a feature-switch read only. |
| `REVIEW.md` | `refactor:` — existing tests must still cover the code | Pass | Platform composer/draft/feedback tests retargeted at the inline path; 3 API tests updated to inline prompt rendering; inline behavior remains covered by the pre-existing `inserts multiple inline templates and sends a template-only message` and `replaces a selected inline template instead of inserting another`. |
| `REVIEW.md` | Feature regression guard — replaced behavior needs explicit coverage | Pass | The switch has been globally `enabled: true` since #25320, so the retained path is what production already runs. Tests that existed only to exercise the discarded switch-off chip were deleted rather than reinterpreted; see Testing for the one genuine coverage reduction. |
| `REVIEW.md` | DB / JSONB review gate | N/A | No `turbo/packages/db/src/schema/**` change and no `jsonb(...).$type<...>()` shape change. |
| `docs/bad-smell.md` | Zero `any` and zero lint/type suppressions | Pass | Scan of added lines found no `any`, `@ts-ignore`/`@ts-expect-error`, `eslint-disable`, `oxlint-disable`, or `prettier-ignore`. |
| `docs/bad-smell.md` | No dynamic `import()`, hardcoded URLs, or fallback patterns | Pass | No dynamic imports or hardcoded config added; the change removes a conditional rather than adding a fallback. |
| `docs/bad-smell.md` | Public interface changes identified | Pass | `projectUserMessage`, `messageDocumentToEditorDoc/Prompt/DisplayText`, and `WebChatSessionPromptContext` all lose the inline-templates parameter; every call site in the diff is updated. |
| `docs/deployment-compatibility.md` | old frontend → new backend stays safe | Pass | `zero-feature-switches.ts` still returns `supportsStructuredInlineTemplates: true`. `applySwitches` (`feature-switch.ts:35-49`) only overrides keys present in the response, so an old bundle keeps its locally-registered `true` and its capability clamp does not fire. |
| `docs/deployment-compatibility.md` | Persisted state stays readable across versions | Pass | The legacy `templateAttachment` node remains registered (`tiptap-workflow-composer.ts:1441`) and decoded (`user-message-document-codec.ts:272,329`), so previously persisted `draftEditorDocument` snapshots still parse. |
| `docs/testing.md` | Deployment-compat combinations covered by tests | Pass | `zero-feature-switches.test.ts` → `keeps the capability handshakes the previous Platform bundle reads`; `user-message-document-codec.test.ts` round-trips a `templateAttachment` document. |
| `docs/testing.md` | Integration tests, mock at the boundary, real DB | Pass | No new `vi.mock()` in the diff; changes are assertions and setup inside existing route and page tests. |
| `docs/testing/anti-patterns.md` | AP-1 behavior over mock calls, AP-2 no direct fetch mocking, AP-4 no internal mocks, AP-5 no fake timers | Pass | None of `toHaveBeenCalled` as sole assertion, `vi.stubGlobal("fetch")`, `vi.mock("../…")`, or `useFakeTimers` appear in added lines. |
| `docs/testing/testing-external-behavior.md` | Platform tests drive and observe through the page | Pass (see note) | Migrated assertions locate the rendered composer node via `[data-composer-inline-template]` and assert its visible title text. That hook is the repo's pre-existing convention — it is used in 4 test files on `main` before this PR — and this PR consolidates it into one shared helper. |
| `docs/testing/api-testing.md` | API tests exercise real routes and assert through the API | Pass | The three changed API tests keep route-driven setup and API-surface assertions; only the expected prompt strings changed. |
| `docs/testing/app-testing.md` | Page tests use `setupPage`/`detachedSetupPage` with MSW handlers | Pass | `page-helper.ts` only drops the switch default; tests keep the standard page + MSW setup. |
| `docs/effect.md` | Derived value → `computed`; commands named for business semantics | Pass | The `inlineTemplatesEnabled$` computed and its threading are deleted; `createInsertUserMessageCommand` stays a command and simply no longer needs store context. |
| `docs/cache.md` | RC-5 no duplicate mutable sources of truth | Pass | Removing the switch eliminates a second source of truth for how templates render. |
| `.claude/skills/ccstate/SKILL.md` | Explicit `accept` lists; no manual caches; no render-time signal allocation | N/A | No HTTP call, cache, or signal allocation is added; the diff only removes signal reads. |
| `docs/event-sourcing.md` | Persistent events authoritative; projections stay consistent | Pass | Event shapes, IDs, and reconciliation are untouched; only the user-message → prompt/display projection changes. |

### Findings

#### Critical (P0)
- None.

#### High Priority (P1)
- None.

#### Non-blocking notes
- `e2e/playwright/tests/chat.spec.ts`: the mocked feature-switch response drops `supportsStructuredInlineTemplates`, but the real API still returns it. The new frontend ignores the field so the test is unaffected; the fixture is now slightly less faithful to the actual response shape.
- `chat-thread-panes.ts` / `agent-draft.ts` now pass `generationTemplate: undefined` explicitly to `draft.seed$`. Behavior is unchanged (the old ternary already yielded `undefined` whenever the switch was on), but the field could be dropped if `seed$` accepts a partial.

### Testing

- **Coverage: adequate for a fully-rolled-out switch removal.** The retained inline path keeps dedicated tests; the deleted tests covered only the discarded switch-off rendering.
- **One genuine coverage reduction, disclosed:** `clears a recalled Morning Brief template after steering it inline` was reduced to `restores a recalled Morning Brief message as an inline template`. The recall/restore half is asserted; the re-send half was dropped because it could not be driven to a captured request in the test harness — most likely because steering goes through the active-input path, for which `mockChatLifecycle` has no capture point. Verified on unmodified `main` with the switch forced on that the same flow produces no captured request, so this was not covering the rolled-out path either.
- **Conventions: compliant.** No new internal mocks, fetch stubs, fake timers, or mock-call-only assertions.
- **CI at the reviewed commit:** all previously failing checks are fixed and no check is failing; the remaining app/API shards and CodeQL were still running when this review was written. Three API expectations were corrected in this PR because inline rendering abuts the template marker to the following text (`cron-execute-morning-briefs`, `ops-logs.bdd`) and uses numbered markers in prior-run context (`chat-callbacks.bdd`).
- Local verification: `pnpm check-types` 13/13, `pnpm turbo run lint` 13/13 with 0 errors, `pnpm knip` clean, `@vm0/core` feature-switch suite 24 pass, and the six touched platform test files 114 pass. API integration tests could not run locally (no PostgreSQL in the sandbox) and are covered by CI.

PR #25444, opened 2026-08-06
fix(runner): require service resume acknowledgement

### LINKED ISSUE #25410: bug(rust): reject service resume when the runner enters Stopping before signal
## Location

- `crates/runner/src/cmd/service/drain_resume.rs:528-619`
- `crates/runner/src/cmd/service/signal.rs:25-53`
- `crates/runner/src/cmd/start/signals.rs:204-210`

## Problem

`service resume` verifies that `status.json` says `Draining`, but the runner can naturally commit to `Stopping` after that preflight and before SIGUSR2 is processed.

A concrete race is:

1. The last active job is about to finish and `status.json` still reports `Draining`.
2. `resume_with_ops()` passes preflight, removes the drain-time `Restart=no` override, enables the unit, and reloads systemd.
3. The runner finishes the job and transitions irreversibly to `Stopping`.
4. `systemctl kill --signal=SIGUSR2` successfully delivers the signal to the still-live main process, so `signal_service_main()` returns `Sent`.
5. `handle_resume_signal()` ignores SIGUSR2 because the current mode is `Stopping`.
6. `resume_after_preflight_with_ops()` treats delivery as success and returns `Ok(())` without verifying that the runner reached `Running`.

This violates the command contract documented in `run_resume`: resume should be refused once teardown is in progress. It also misses the original soft-drain requirement from #9771 that SIGUSR2 after the irreversible phase must fail clearly.

## Impact

The CLI reports a successful resume even though the runner continues shutting down. On a normal clean drain exit, `Restart=on-failure` does not restart the process, so the newly enabled unit can remain inactive until an operator notices and starts it manually. The command has also changed the restart override and boot enablement despite failing to resume the live runner.

## Acceptance criteria

- `service resume` succeeds only after observing that the target runner reached `Running`.
- Use a bounded acknowledgement/postcondition check; successful systemd signal delivery alone is not a resume acknowledgement.
- If the runner reaches `Stopping`, `Stopped`, or exits during the transition, return a clear failure and restore the prior restart override and enablement state.
- Add a deterministic regression test for `Draining` preflight followed by `Stopping` before or while SIGUSR2 is delivered.

seven332:
## Summary

- require `service resume` to observe the same runner process transition to `running` before reporting success
- bound post-signal polling to 10 seconds and reject stopped, replaced, invalid, unreadable, or inactive runner states
- restore the previous systemd restart override and enablement when resume acknowledgement fails
- cover successful, delayed, raced, terminal, replacement, unreadable, and timeout outcomes with filesystem-backed status tests

## Scope

- no runner signal or lifecycle protocol changes
- no API, schema, migration, or persisted status format changes

## Validation

- `cargo clippy -p runner --bin runner --all-features -- -D warnings`
- `cargo test -p runner --bin runner cmd::service::drain_resume::tests::`
- `cargo test -p runner --bin runner --quiet`
- repository pre-commit hooks (Rust format, workspace Clippy, workspace rustdoc, file-size check, commitlint)

Closes #25410

### DIFF
--- a/crates/runner/src/cmd/service/drain_resume.rs
+++ b/crates/runner/src/cmd/service/drain_resume.rs
@@ -774,6 +774,10 @@ mod tests {
         format!(r#"{{"mode":"{mode}","active_runs":[],"started_at":"{started_at}"}}"#)
     }
 
+    fn active_results(count: usize) -> VecDeque<RunnerResult<bool>> {
+        (0..count).map(|_| Ok(true)).collect()
+    }
+
     async fn write_test_status(base_dir: &Path, mode: &str, started_at: &str) {
         tokio::fs::create_dir_all(base_dir).await.unwrap();
         tokio::fs::write(
@@ -850,7 +854,7 @@ mod tests {
         fn default() -> Self {
             Self {
                 events: Vec::new(),
-                active_results: VecDeque::from([Ok(true)]),
+                active_results: active_results(1),
                 enablement_results: VecDeque::from([Ok(SystemdUnitEnablement::NotEnabled)]),
                 write_error: false,
                 remove_error: false,
@@ -982,7 +986,7 @@ mod tests {
         ) -> ServiceFuture<'a, bool> {
             self.events.push("is_active");
             Box::pin(std::future::ready(
-                self.active_results.pop_front().unwrap_or(Ok(true)),
+                self.active_results.pop_front().unwrap_or(Ok(false)),
             ))
         }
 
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #25444

**Reviewed head:** `797af5d4f5125f56c07c3d00eee4b93a151b0d8b`

### Summary

The change replaces signal-delivery success with a bounded, same-process lifecycle acknowledgement and routes every post-mutation failure through the existing exact systemd rollback. It reuses the current atomic status file and bounded systemctl machinery without changing schemas, serialization, APIs, or runner protocols.

### Findings

No findings.

### Testing

Focused resume tests cover immediate and delayed acknowledgement, the reported `draining`-to-`stopping` race, stopped/inactive/replaced/unreadable/unknown states, timeout, and rollback. The final head passed Rust formatting, Clippy with warnings denied, rustdoc with warnings denied, 35 focused service tests, and the full runner binary suite (3007 passed, 20 ignored). A live systemd/runner smoke test was not run locally.

### Verdict

LGTM

PR #25459, opened 2026-08-06
test(runner): enforce wait probe deadlines

### LINKED ISSUE #25448: test(rust): enforce wait_for_probe deadlines around probes
## Problem and evidence

`wait_for_probe` accepts a timeout and computes an absolute deadline in `crates/runner/src/cmd/start/tests/support/wait.rs`, but it currently awaits `probe()` without applying that deadline. The deadline assertion runs only after a probe returns `WaitProbe::Pending`, and the subsequent polling sleep is also unbounded by the absolute deadline.

A probe that stops making asynchronous progress therefore prevents the helper from checking its timeout. This is a concrete risk because callers await shared pool locks, cancellation-registry operations, workspace-cache state, and async status-file I/O across `crates/runner/src/cmd/start/tests/support/wait.rs` and `status.rs`. The behavior was verified unchanged on `main` at `d85149058b`.

There is no direct helper regression test. A never-resolving probe can currently consume the enclosing Rust coverage job's full 20-minute timeout instead of producing a focused local failure.

## Decision

Apply one Tokio absolute-deadline timeout around the complete polling state machine. Keep the latest caller-provided `WaitProbe::Pending` message outside the timed future so timeout failures retain useful context. If the first probe never returns, report a generic message containing the requested timeout.

The direct never-resolving regression will use Tokio's paused clock plus a longer same-future test guard around panic capture. If the helper deadline regresses, that guard drops the stalled helper future and fails the test promptly; it does not spawn a watchdog task or leak work.

## Alternatives considered

- Wrapping every probe and polling sleep separately with the same deadline can distinguish phases, but duplicates deadline handling and makes each future await responsible for remembering the boundary.
- An explicit `select!` race adds simultaneous-readiness policy without a stronger guarantee.
- A spawned watchdog adds unnecessary task cancellation and leak risks and conflicts with the acceptance criteria.
- Leaving the helper unchanged allows a local assertion helper to consume a 20-minute CI job timeout.

## Risks and limits

- Tokio timeout cancellation is cooperative. It cannot preempt synchronous non-yielding work; current callers use yielding async operations or short synchronous reads.
- Before a probe returns `Pending`, no caller-specific diagnostic exists, so the first-probe timeout must be generic.
- The change is private Rust test support only. It has no API, schema, persistence, runner protocol, migration, deployment, or rollout compatibility surface.

## Acceptance criteria

- One absolute caller-provided deadline bounds probe execution and polling intervals.
- A yielding probe that never resolves fails at the helper's local deadline.
- A deterministic paused-time regression verifies the local deadline and has a longer same-future guard so a future regression cannot hang the suite.
- Repeated `WaitProbe::Pending` results retain their latest useful diagnostic at timeout.
- A probe that becomes `WaitProbe::Ready` within the budget still returns successfully.
- The implementation creates no detached watchdog task and leaks no test work.
- Focused and full runner tests, Rust formatting, runner Clippy across all targets/features, and runner documentation checks pass.

seven332:
## Summary

- enforce one absolute deadline around the complete `wait_for_probe` probe-and-poll loop
- preserve the latest caller-provided pending diagnostic, with a generic message when the first probe stalls
- add paused-time coverage for a never-resolving probe, repeated pending diagnostics, and pending-to-ready success
- keep the never-resolving regression bounded by a longer same-future test guard without spawning work

## Scope

This changes private runner test support only. It does not change runtime APIs, schemas, persistence, runner protocols, dependencies, migrations, or deployment compatibility.

## Validation

- `cargo test --manifest-path crates/Cargo.toml -p runner wait_for_probe`
- `RUSTDOCFLAGS="-D warnings" cargo doc --manifest-path crates/Cargo.toml -p runner --all-features --no-deps`
- `cargo test --manifest-path crates/Cargo.toml -p runner` (3003 unit tests passed, 20 ignored; 1 integration test passed)

Closes #25448

### DIFF
--- a/crates/runner/src/cmd/start/tests/support/wait.rs
+++ b/crates/runner/src/cmd/start/tests/support/wait.rs
@@ -1,6 +1,8 @@
 use super::super::super::*;
 use super::env::MockRunEnv;
+use futures_util::FutureExt;
 use std::future::Future;
+use std::panic::AssertUnwindSafe;
 
 use crate::idle_pool::ParkingState;
 use crate::run_cancellation::{RunCancellationHandle, RunCancellationRegistry};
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #25459

**Reviewed head:** `5f3622e4dd3921cd8d9919f24145abc8486f5462`

### Summary

The PR correctly moves the existing absolute deadline to the boundary of the complete probe-and-poll operation. It preserves the latest available pending diagnostic, handles a first unresolved probe explicitly, and keeps timeout cancellation within the calling task.

### Findings

No findings.

### Testing

Paused-time coverage exercises expiry inside a never-resolving probe, expiry after repeated pending results, and pending-to-ready success. The unresolved case has a longer same-future guard, so a regression fails promptly without spawning or leaking work. Formatting, all-target/all-feature runner Clippy, warning-denied runner rustdoc, 3 focused tests, 20 repeated timing-review runs, 3003 runner unit tests, and the runner integration test passed. Codecov also reports all modified coverable lines covered. Workspace-wide Rust tests were not rerun locally because the change is private to runner test support and the full affected crate passed.

### Verdict

LGTM

PR #25466, opened 2026-08-06
test(firewall): cover base specificity decisions

### LINKED ISSUE #25453: test(firewall): cover cross-language base specificity dimensions
## Problem

The Python and TypeScript firewall decision matchers share two observable base-URL specificity dimensions that are not protected by final-decision regression coverage:

- a `+` greedy base segment ranks above a matching `*` greedy base segment for a non-empty capture;
- mixed parameter segments include their literal Unicode code-point count, so two literal ASCII code points rank above one emoji code point when earlier dimensions tie.

Both implementations use these coordinates while selecting the winning base before rule and network-policy evaluation:

- Python: `crates/runner/mitm-addon/src/firewall_matching/base_url.py`
- TypeScript: `turbo/packages/connectors/src/firewall-rule-matcher.ts`

The focused Python base-precedence suite covers longer path bases, parameterized bases, and static-host-over-wildcard precedence, but not these two decisions. The shared final-decision contract covers the corresponding **rule** specificity behavior added by #25263, but it does not cover these separate base-specificity coordinates.

## Evidence

On `c89e880eb5d1d2d16abf1ddce921a961c78b2d31`, each of these Python mutations independently left 958 matching/firewall tests green:

1. Set `_BASE_PLUS_GREEDY_SEGMENT_SCORE` from `1` to `0`.
2. Preserve `_BASE_MIXED_PARAM_SEGMENT_SCORE` but remove `len(prefix) + len(suffix)` from mixed base segments.

Both mutations change observable output through `matching.match_compiled_firewall_request()`:

- A request matching both `https://{deployment*}.bentoml.ai` and `https://{deployment+}.bentoml.ai` currently selects the `+` base. Neutralizing the `+` score makes the candidates tie and changes the final decision.
- A request matching both `https://api.example.com/files/😀{id}` and `https://api.example.com/files/{id}ab` currently selects the base with two ASCII literal code points. Removing mixed literal-character scoring makes the candidates tie and changes the final decision.

Repository inspection also confirms that the TypeScript decision matcher independently implements the same score ordering and currently lacks final-decision coverage for these base dimensions. This makes the gap a cross-language contract risk, not only a Python test-file gap.

## Decision

Proceed with regression coverage, but treat the public cross-language base-selection behavior as the requirement. Planning should compare shared contract ownership with language-specific tests instead of assuming the Python-focused file is the correct boundary.

## Impact

Without coverage, a refactor can silently remove either coordinate while tests remain green. A request can then select multiple tied bases or the wrong base, changing authentication, permission, and final allow/block behavior. Because Python and TypeScript implement the decision independently, covering only one implementation would still permit semantic drift.

## Alternatives considered

- **Do nothing:** rejected because independent mutations demonstrate an observable security-policy decision change with a green test suite.
- **Protect only the Python scorer:** narrower, but it leaves the independently implemented TypeScript decision path and cross-language parity unprotected.
- **Assert private numeric scores:** rejected because the contract is the selected base and final policy decision, not the current representation of the score.

## Risks and mitigations

- The cases could accidentally exercise another specificity dimension. Keep the greedy case ASCII-only and make all earlier dimensions tie in the literal-character case.
- Same-specificity policy reduction could hide a lost coordinate. Use opposing allow/deny outcomes and ordering that makes a removed coordinate visibly change the final decision.
- Duplicate fixtures could create maintenance noise. Planning should choose one ownership boundary unless separate tests exercise materially different behavior.

## Acceptance criteria

- Cover a `+` greedy base winning over a matching `*` greedy base for a non-empty capture through the public final-decision paths.
- Cover a mixed base segment with two ASCII literal code points outranking one emoji code point when earlier score dimensions tie.
- Exercise both the TypeScript decision matcher and the indexed Python production matcher.
- Make each winning base observable through opposing allow/deny policies and assert the selected base and final decision without asserting private score values.
- Verify that independently neutralizing each covered coordinate makes its corresponding regression fail.
- Keep the submitted change test-only, with no production matcher, API, protocol, persistence, dependency, migration, or rollout change.

## Delivery notes

This remains one independently reviewable PR with no parent issue, dependency, migration, or staged rollout requirement.

seven332:
## Summary

- add shared final-decision coverage for `+` greedy base precedence over `*` greedy base precedence
- add shared final-decision coverage for mixed-base Unicode code-point literal specificity
- exercise the TypeScript decision matcher and indexed Python production matcher from the same fixtures
- assert the selected allowed base, relative path, permission, and rule through public output

## Exclusions

- no production matcher changes
- no duplicate language-specific fixtures
- no API, protocol, schema, persistence, dependency, migration, or rollout changes

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (4,938 passed)
- `pnpm exec prettier --check packages/connectors/src/__tests__/firewall-semantics-contract.json`
- `pnpm -F @vm0/connectors lint`
- `pnpm -F @vm0/connectors check-types`
- `pnpm -F @vm0/connectors test` (904 passed)
- `pnpm knip`
- `git diff --check`
- `./scripts/check-file-size.sh turbo/packages/connectors/src/__tests__/firewall-semantics-contract.json`
- independent Python and TypeScript mutations confirmed that each new case fails when its intended specificity coordinate is neutralized

Closes #25453

### DIFF
--- a/turbo/packages/connectors/src/__tests__/firewall-semantics-contract.json
+++ b/turbo/packages/connectors/src/__tests__/firewall-semantics-contract.json
@@ -1128,6 +1128,106 @@
       }
     },
     {
+      "name": "plus greedy allowed base wins over star greedy denied base",
+      "firewalls": [
+        {
+          "name": "bentoml",
+          "apis": [
+            {
+              "base": "https://{deployment*}.bentoml.ai",
+              "auth": {},
+              "permissions": [
+                {
+                  "name": "deployment.star",
+                  "rules": ["GET /api"]
+                }
+              ]
+            },
+            {
+              "base": "https://{deployment+}.bentoml.ai",
+              "auth": {},
+              "permissions": [
+                {
+                  "name": "deployment.plus",
+                  "rules": ["GET /api"]
+                }
+              ]
+            }
+          ]
+        }
+      ],
+      "networkPolicies": {
+        "bentoml": {
+          "allow": ["deployment.plus"],
+          "deny": ["deployment.star"],
+          "ask": [],
+          "unknownPolicy": "deny"
+        }
+      },
+      "request": {
+        "method": "GET",
+        "url": "https://foo.bentoml.ai/api"
+      },
+      "expected": {
+        "kind": "allow",
+        "firewallName": "bentoml",
+        "base": "https://{deployment+}.bentoml.ai",
+        "relativePath": "/api",
+        "permission": "deployment.plus",
+        "rule": "GET /api"
+      }
+    },
+    {
+      "name": "two-code-point allowed base wins over one-code-point emoji denied base",
+      "firewalls": [
+        {
+          "name": "files",
+          "apis": [
+            {
+              "base": "https://api.example.com/files/😀{id}",
+              "auth": {},
+              "permissions": [
+                {
+                  "name": "files.emoji",
+                  "rules": ["GET /tail"]
+                }
+              ]
+            },
+            {
+              "base": "https://api.example.com/files/{id}ab",
+              "auth": {},
+              "permissions": [
+                {
+                  "name": "files.ascii",
+                  "rules": ["GET /tail"]
+                }
+              ]
+            }
+          ]
+        }
+      ],
+      "networkPolicies": {
+        "files": {
+          "allow": ["files.ascii"],
+          "deny": ["files.emoji"],
+          "ask": [],
+          "unknownPolicy": "deny"
+        }
+      },
+      "request": {
+        "method": "GET",
+        "url": "https://api.example.com/files/😀xab/tail"
+      },
+      "expected": {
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #25466

**Reviewed head:** `3b8a571cfeb44eb9979966dd04902033c010831f`

### Summary

The two shared fixtures isolate the intended base-specificity coordinates and exercise both public production decision implementations without production changes or duplicate language-specific tests.

### Findings

No findings.

### Testing

Python lock/static checks and all 4,938 add-on tests passed. Connectors formatting, lint, type checking, all 904 tests, and Knip passed. Four independent Python/TypeScript mutations made only the corresponding new case fail with the expected tied-base malformed result. No in-scope area remains unverified.

### Verdict

LGTM

PR #25470, opened 2026-08-06
perf(python): bound proxy-log url rendering

### LINKED ISSUE #25451: perf(python): bound proxy-log URL rendering before serialization
## Problem

The bounded URL projection added by #25330 protects top-level HTTP network-log rows, but the same response and connection-error hooks subsequently render the complete request URL into per-run proxy diagnostics without a corresponding precheck.

- `mitm_addon.py` sanitizes `original_url` and embeds it in every 4xx/5xx `http_error` message.
- The same module repeats that work for `connection_error`.
- `logging_utils.py` JSON-renders and UTF-8 encodes the complete entry before the bounded JSONL writer sees it.
- `connector_diagnostics.py` renders the sanitized URL into `message` and passes the raw URL as an exact `url` extra, causing the sanitizer to run again inside `log_proxy_entry()`.

A large query is removed early, but a large ordinary path survives the sanitizer and is copied into the diagnostic message. Pinned mitmproxy 12.2.3 has no HTTP/1 request-head limit before addon hooks. The top-level HTTP network row now emits a small omission row above its URL boundary, while the later proxy diagnostic still performs work proportional to a retained path.

## Evidence

The original Python 3.12.3 probe used production `sanitize_url_for_network_log()`, the exact `http_error` shape, and production JSONL encoding:

| Path payload | JSONL bytes | Median synchronous time | Peak traced allocation |
| ---: | ---: | ---: | ---: |
| 1 MiB | 1,048,728 | 5.409 ms | 4.25 MiB |
| 5 MiB | 5,243,032 | 26.753 ms | 21.25 MiB |
| 16 MiB | 16,777,368 | 85.199 ms | 68.00 MiB |
| 33 MiB | 34,603,160 | 196.475 ms | 140.25 MiB |

`jsonl_writer.MAX_PENDING_JSONL_BYTES` is 32 MiB. Its admission check receives an already-created byte string, so the 33 MiB case pays the full synchronous CPU and allocation cost before rejection. Smaller large lines can occupy most of the shared queue and still delay unrelated flows on the mitmproxy event loop.

A current-main sanitizer-only comparison also distinguishes retained path data from discarded query data:

| Input shape | Raw characters | Sanitized characters | Median sanitizer time | Peak traced allocation |
| 5 MiB ordinary path | 5,242,907 | 5,242,907 | 4.434 ms | 15,729,060 bytes |
| short path plus 5 MiB query | 5,242,920 | 31 | 0.006 ms | 504 bytes |

The proxy logger intentionally discards query and fragment content before normalization and parsing. A limit on total raw length would therefore replace a currently cheap and useful short-path diagnostic merely because discarded data is large.

## Impact

A registered workload can trigger proportional response-hook work with a long safe request path whenever the provider returns an error status or the connection fails. Repeated failures continue paying the cost even when the completed log entry cannot be queued.

## Decision

Use one shared typed projection for request-derived URLs written to proxy diagnostics:

- Reuse the existing 1,000,000-character number, but apply it to the query/fragment-free value that proxy diagnostics retain rather than the complete raw request URL.
- For raw input above the ceiling, inspect only the first 1,000,001 characters for the first raw `?` or `#`. If that delimiter proves the retained prefix fits, sanitize only that bounded prefix. Otherwise use a fixed `[truncated]` marker.
- On omission, include `url_truncated: true` and `url_original_char_count` with the exact complete raw Python character count.
- Route `http_error`, `connection_error`, and connector-diagnostic message construction through the projection.
- Let connector diagnostics reuse that same projection for the existing exact `url` extra rather than sanitizing the original URL twice.
- Preserve the current raw-string exact-`url` sanitizer for unrelated proxy-log callers.

This is bounded delimiter discovery, not a second URL parser. Every retained prefix still uses the existing normalization, uncached parsing, malformed-authority recovery, query/fragment policy, and userinfo redaction. Omission never retains a raw prefix because it could expose partial userinfo when the separating `@` lies beyond the boundary.

The top-level network row continues limiting the complete raw URL because it retains query and fragment data. A huge discarded query may therefore truncate the network row while the proxy diagnostic preserves its short path; the two logs intentionally retain different values.

## Alternatives considered

- **Gate complete raw length:** gives an O(1) rejection but unnecessarily discards useful short paths followed by huge queries.
- **Move serialization to the writer:** moves event-loop work but retains request-sized data and still performs proportional allocation before useful admission.
- **Add a specialized URL-message logger:** centralizes ownership but encodes caller-specific message formatting and a connector-only structured-field switch for three call sites.
- **Use only structured URLs:** removes message duplication but breaks current human-readable proxy messages without a repository renderer to replace them.
- **Reject large request targets at ingress:** changes workload behavior and cannot prevent mitmproxy's earlier request-head buffering at this addon boundary.
- **Truncate a raw prefix:** creates a partial-userinfo disclosure risk.

## Risks and mitigations

- **Delimiter semantics drift:** keep the bounded operation to the existing first raw query/fragment boundary and cover query-first, fragment-first, at-limit, and no-early-delimiter behavior.
- **Projection bypass misuse:** only the typed value returned by the projection helper bypasses repeat exact-`url` sanitization; ordinary strings retain central sanitization.
- **Normal diagnostic changes:** add metadata only on omission and retain existing message/redaction matrices.
- **False performance confidence:** use guarded-input and output-size assertions for stage ordering; keep timing and allocation measurements as recorded evidence rather than test thresholds.

## Acceptance criteria

- An ordinary retained path above 1,000,000 characters is omitted after at most a bounded delimiter search and before full-value normalization, URL parsing, message construction, JSON rendering, or UTF-8 encoding.
- A short retained path followed by an arbitrarily large discarded query or fragment preserves the current useful path without processing the suffix through normalization or parsing.
- Oversized response-error, connection-error, and connector-diagnostic entries contain `[truncated]`, `url_truncated: true`, and the exact complete raw Python character count without retaining a raw prefix.
- Connector diagnostics reuse one projection for both `message` and the existing exact `url` field.
- Existing normal-size messages preserve current path display, query/fragment removal, malformed-authority handling, userinfo redaction, and structured connector fields.
- Real response, error, and connector-diagnostic entry points provide structural regression coverage; no wall-clock threshold is used.
- A focused before/after response-hook benchmark records bounded work and allocation for multi-megabyte safe paths.
- Locked Python static checks, focused integration tests, and the complete add-on suite pass.

## Delivery notes

- This fits one independently reviewable Python add-on PR.
- No API schema, database migration, protocol, persisted relational data, generated artifact, dependency, environment variable, feature switch, staged rollout, or cleanup issue is required.

## Distinctness

- #25330 bounds the top-level HTTP network-log URL, not per-run proxy diagnostics.
- #24537 avoids parsing query and fragment data that proxy diagnostics discard; this issue bounds the ordinary retained path.
- #25404 addresses earlier runtime URL syntax scans during request classification, not response/error diagnostic rendering and JSON encoding.

No overlapping open issue or pull request was found.

seven332:
## Summary

- bound proxy diagnostic URL projection to the retained query/fragment-free value before normalization, parsing, message construction, or serialization
- emit a fixed marker plus exact truncation metadata for oversized retained paths while preserving short paths followed by large discarded query or fragment data
- reuse one immutable projection for connector diagnostic messages and structured URL fields, without weakening sanitization for ordinary raw URL callers
- cover response errors, connection errors, connector diagnostics, delimiter ordering, and the exact retained-length boundary through real addon entry points

## Scope

- no API, database, queue protocol, persisted relational data, dependency, environment variable, feature switch, or generated contract changes
- the existing top-level network-log policy remains unchanged

## Performance

A seven-iteration real response-hook probe constructed each input before measurement. Multi-megabyte retained paths now produce bounded proxy rows:

| Path payload | Median hook time | Peak traced allocation | JSONL bytes |
| ---: | ---: | ---: | ---: |
| 1 MiB | 0.217 ms | 2,670 bytes | 195 |
| 5 MiB | 0.196 ms | 2,518 bytes | 195 |
| 16 MiB | 0.199 ms | 2,454 bytes | 196 |
| 33 MiB | 0.175 ms | 2,390 bytes | 196 |

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_response_handler_logging.py tests/test_error_handler.py tests/test_response_handler_connector_diagnostics.py tests/test_logging_utils.py` (144 passed)
- `uv run --no-sync python -m pytest tests/` (4,942 passed)
- `lefthook run pre-commit`
- `git diff --check`

Closes #25451

### DIFF
--- a/crates/runner/mitm-addon/src/connector_diagnostics.py
+++ b/crates/runner/mitm-addon/src/connector_diagnostics.py
@@ -47,10 +47,9 @@ import connector_intent
 import flow_metadata
 import flow_metadata_keys as metadata_keys
 import matching
-import network_log_sanitization
 import request_classification
 import runtime_url_parsing
-from logging_utils import log_proxy_entry
+from logging_utils import log_proxy_entry, project_url_for_proxy_log
 
 _HTTP_STATUS_UNAUTHORIZED = 401
 _HTTP_STATUS_FORBIDDEN = 403
@@ -788,7 +787,7 @@ def _log_proxy_entry(
     candidate = _candidate_from_flow(flow)
     if candidate is None:
         return
-    safe_url = network_log_sanitization.sanitize_url_for_network_log(original_url)
+    url_projection = project_url_for_proxy_log(original_url)
     extra: dict[str, object] = {}
     ownership_reason = flow.metadata.get(_CONNECTOR_DIAGNOSTIC_OWNERSHIP_REASON)
     if isinstance(ownership_reason, str) and ownership_reason:
@@ -801,15 +800,16 @@ def _log_proxy_entry(
     ownership_hint_status = flow.metadata.get(_CONNECTOR_DIAGNOSTIC_OWNERSHIP_HINT_STATUS)
     if isinstance(ownership_hint_status, str) and ownership_hint_status:
         extra["ownership_hint_status"] = ownership_hint_status
+    extra.update(url_projection.truncation_fields())
     log_proxy_entry(
         flow_metadata.proxy_log_path(flow.metadata),
         "warn",
-        f"{candidate.connector_slug} is not configured for this run: {safe_url}",
+        f"{candidate.connector_slug} is not configured for this run: {url_projection.value}",
         type="connector_diagnostic",
         connector=candidate.connector_slug,
         reason=candidate.reason,
         upstream_status=upstream_status,
-        url=original_url,
+        url=url_projection,
         **extra,
     )
     flow.metadata[_CONNECTOR_DIAGNOSTIC_PROXY_ENTRY_LOGGED] = True
--- a/crates/runner/mitm-addon/src/logging_utils.py
+++ b/crates/runner/mitm-addon/src/logging_utils.py
@@ -7,6 +7,7 @@ log entries, and extracting firewall metadata.
 import json
 import time
 from collections.abc import Mapping
+from dataclasses import dataclass
 from datetime import UTC, datetime
 
 from mitmproxy import ctx, http
@@ -21,10 +22,25 @@ _PROXY_LOG_RESERVED_FIELDS = {"timestamp", "level", "message"}
 # Network log size fields are consumed as JavaScript numbers downstream.
 NETWORK_LOG_MAX_SAFE_SIZE = 9_007_199_254_740_991
 
-# Keep URL-driven HTTP rows below the runner uploader's 1 MiB payload envelope.
-HTTP_NETWORK_LOG_MAX_URL_CHARACTERS = 1_000_000
+# Network rows limit the complete request URL; proxy diagnostics apply the same
+# number to the query-free value they retain.
+URL_LOG_MAX_CHARACTERS = 1_000_000
 HTTP_NETWORK_LOG_MAX_JSONL_BYTES = 1_000_000
-_HTTP_NETWORK_LOG_TRUNCATED_URL = "[truncated]"
+_TRUNCATED_URL = "[truncated]"
+
+
+@dataclass(frozen=True)
+class _ProxyLogUrlProjection:
+    value: str
+    original_char_count: int | None = None
+
+    def truncation_fields(self) -> dict[str, object]:
+        if self.original_char_count is None:
+            return {}
+        return {
+            "url_truncated": True,
+            "url_original_char_count": self.original_char_count,
+        }
 
 
 def elapsed_ms(start_time: float | None) -> int:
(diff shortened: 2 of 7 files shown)


seven332:
## Code Review: PR #25470

**Reviewed head:** `79b54dfcfc596518bc0147bed6b63f947c79b818`

### Summary

The retained query/fragment-free URL is bounded before normalization, parsing, message rendering,
and JSON encoding. Response errors, connection errors, and connector diagnostics share the policy;
connector messages and structured URL fields reuse one private immutable projection without
weakening raw-string URL sanitization.

### Findings

No findings.

### Testing

Real response, error, and connector entry points cover oversized paths, large discarded query and
fragment suffixes, exact boundary behavior, userinfo redaction, output size, and bounded delimiter
searches. Ruff, BasedPyright, 144 focused integration tests, and the complete 4,942-test add-on suite
passed. A response-hook probe also kept 1–33 MiB retained paths to 195–196-byte rows with bounded
allocation. Unrelated Turbo and Rust suites were not run because the diff is Python-addon-only.

### Verdict

LGTM

PR #25485, opened 2026-08-06
test(runner): cover retry timer in-flight guard

### LINKED ISSUE #25474: test(rust): cover retry timer in-flight guard
## Context

`RetryState::timer_ready` in `crates/runner/src/retry.rs:66-68` deliberately requires both an elapsed restart deadline and `handle.is_none()`. However, `timer_ready_requires_no_handle_and_past_restart` at `crates/runner/src/retry.rs:159-168` only checks the missing-deadline and elapsed-deadline cases; it never puts a handle in the state. Removing the `self.handle.is_none() &&` guard would therefore leave the named regression test green.

The guard is part of the shared lifecycle contract used by:

- `maybe_spawn_ably_connect` at `crates/runner/src/provider/api_ably_supervisor.rs:1021-1033`
- `maybe_spawn_mitm_restart` at `crates/runner/src/cmd/start/mitm_restart.rs:18-39`

Both callers replace `retry.handle` when the timer reports ready. If the guard regresses while an elapsed/stale timer and an in-flight task coexist, another attempt can be launched and the only `JoinHandle` for the existing task can be overwritten. The surrounding caller tests cover retry results and shutdown cleanup, but no current test exercises the expired-timer-plus-populated-handle state.

## Acceptance criteria

- Add deterministic Rust coverage that sets an expired `restart_at` together with a live/in-flight handle and asserts that `timer_ready()` remains false.
- Verify that the same elapsed deadline becomes eligible only after the handle is cleared or completed.
- Ensure the regression test fails if the `handle.is_none()` guard is removed.
- Bound and clean up any spawned task so the test cannot leak or hang.

seven332:
## Summary

- exercise an elapsed retry deadline while a real pending task handle remains in flight
- assert the handle guard blocks readiness, then abort and await the task before checking the result
- verify the unchanged elapsed deadline becomes ready after the handle is removed

## Explicit exclusions

- no production retry behavior, caller logic, dependency, API, schema, persisted data, or protocol changes

## Validation

- `cargo test -p runner retry::tests::timer_ready_requires_no_handle_and_past_restart -- --exact`
- mutation proof: removing `self.handle.is_none() &&` makes the focused test fail at the in-flight readiness assertion; restoring it makes the test pass
- `cargo test -p runner --bin runner`

Closes #25474

### DIFF
--- a/crates/runner/src/retry.rs
+++ b/crates/runner/src/retry.rs
@@ -156,14 +156,26 @@ mod tests {
         assert_eq!(rs.consecutive_failures, 0);
     }
 
-    #[test]
-    fn timer_ready_requires_no_handle_and_past_restart() {
-        let mut rs: RetryState<()> =
+    #[tokio::test]
+    async fn timer_ready_requires_no_handle_and_past_restart() {
+        let mut rs: RetryState<tokio::task::JoinHandle<()>> =
             RetryState::new(Duration::from_secs(0), Duration::from_secs(60), None);
         // No restart_at → not ready
         assert!(!rs.timer_ready());
         // Set restart_at to the past so it's immediately ready.
         rs.restart_at = Some(Instant::now() - Duration::from_secs(1));
+        assert!(rs.timer_ready());
+
+        rs.handle = Some(tokio::spawn(std::future::pending::<()>()));
+        let ready_while_in_flight = rs.timer_ready();
+
+        // Reap the task before asserting so a regression cannot detach it.
+        let handle = rs.handle.take().unwrap();
+        handle.abort();
+        let error = handle.await.unwrap_err();
+        assert!(error.is_cancelled());
+
+        assert!(!ready_while_in_flight);
         assert!(rs.timer_ready());
     }
 


seven332:
## Code Review: PR #25485

**Reviewed head:** `cca34b7b9d2a6fb37c043b3a9cb50bfa93e532ad`

### Summary

Reviewed the single test-only commit against the retry state lifecycle, both callers, the approved issue plan, and the repository's Rust testing guidance. The real pending task accurately represents the populated-handle invariant, and cleanup precedes the mutation-sensitive assertion.

### Findings

No findings.

### Testing

- Focused regression test passed, and the guard-removal mutation failed at the intended in-flight assertion.
- The focused test passed 25 consecutive repetitions.
- Runner tests passed: 3017 passed, 20 ignored, 0 failed.
- Rust formatting, affected-scope Clippy, and documentation checks passed; pre-commit workspace Rust checks also passed.
- Workspace-wide `cargo llvm-cov --all-targets --all-features` was not run locally and remains covered by CI.

### Verdict

LGTM

PR #25488, opened 2026-08-06
fix(chat): preserve scroll across sharing transitions

e7h4n:
## Summary

- update the sharing phase first, then synchronously read the still-rendered chat position and request the existing after-render auto-scroll so the visible anchor is preserved or the tail continues following
- apply the behavior through both desktop and mobile sharing controls with the page lifecycle signal while keeping scroll state scoped to each thread panel
- keep the normal and sharing thread headers at the same fixed 56px height so the bottom composer/share bar is the vertical layout change

## Verification

- `pnpm format`
- `pnpm knip`
- `pnpm --filter @vm0/app exec vitest run src/views/zero-page/__tests__/chat-scroll-position.test.tsx` (21 tests)
- `pnpm --filter @vm0/app exec vitest run src/views/zero-page/__tests__/chat-thread-sharing.test.tsx` (3 tests)

Browser verification was not run because the repository instructions prohibit starting a local dev server unless explicitly requested.

### DIFF
--- a/turbo/apps/platform/src/signals/chat-page/chat-thread-scroll.ts
+++ b/turbo/apps/platform/src/signals/chat-page/chat-thread-scroll.ts
@@ -17,7 +17,7 @@ interface ScrollAfterRenderRequest {
   readonly position: ThreadScrollPosition | null;
 }
 
-interface ChatThreadScrollSignals {
+export interface ChatThreadScrollSignals {
   readonly scrollContainerOnRef$: Command<
     (() => void) | undefined,
     [HTMLElement | null]
--- a/turbo/apps/platform/src/signals/chat-page/chat-thread-sharing.ts
+++ b/turbo/apps/platform/src/signals/chat-page/chat-thread-sharing.ts
@@ -3,6 +3,7 @@ import { command, computed, state, type Command, type Computed } from "ccstate";
 
 import { accept } from "../../lib/accept.ts";
 import { zeroClient$ } from "../api-client.ts";
+import type { ChatThreadScrollSignals } from "./chat-thread-scroll.ts";
 
 const SHARED_THREAD_SELECTION_TEXT_LIMIT_BYTES = 1.5 * 1024 * 1024;
 
@@ -22,8 +23,8 @@ export interface ChatThreadSharingSignals {
   readonly selectedEventIds$: Computed<ReadonlySet<string>>;
   readonly selectedCount$: Computed<number>;
   readonly createdSharedThreadId$: Computed<string | null>;
-  readonly start$: Command<void, []>;
-  readonly close$: Command<void, []>;
+  readonly start$: Command<Promise<void>, [AbortSignal]>;
+  readonly close$: Command<Promise<void>, [AbortSignal]>;
   readonly toggle$: Command<
     ToggleSharedThreadSelectionResult,
     [readonly ShareableChatEvent[]]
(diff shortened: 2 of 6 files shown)


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 185d542001a615ac61aa06ad28b750ed4c6ae39f
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

This fix routes sharing-state entry and exit through the existing rendered-event scroll restoration path: the phase changes synchronously, the still-rendered event anchor is captured, and the next-frame commit either restores that anchor or continues following the tail. It also gives the normal and sharing desktop headers the same fixed height and updates every desktop and mobile caller to use the page lifecycle signal.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness, edge cases, and async ordering | Pass | `chat-thread-sharing.ts` performs `phase update -> rendered-position read -> autoScroll$` in one command; `chat-thread-scroll.ts` retains revision-based next-frame commits and bottom-follow behavior. |
| `REVIEW.md` | Security, type safety, and interface changes | Pass | No auth, HTML, external-input, or secret boundary changes; `ChatThreadScrollSignals` is explicitly typed, sharing accepts only the two required commands through `Pick`, and every changed caller supplies the new `AbortSignal`. |
| `REVIEW.md` | Regression coverage required for a `fix:` change | Pass | `chat-scroll-position.test.tsx` reproduces the sharing layout shift and verifies the same visible event offset and exact scroll restoration in both transition directions. |
| `docs/bad-smell.md` | No `any`, suppressions, dynamic imports, hardcoded configuration, or fallback paths | Pass | The complete production diff adds none of these patterns; the change remains typed command composition and shared CSS class reuse. |
| `docs/testing.md` | Prefer integration coverage, mock only boundaries, and assert behavior | Pass | The regression enters through the real page setup, clicks user-visible controls, and asserts rendered anchor/scroll outcomes without mocking internal modules or `fetch`. |
| `docs/effect.md` | Match work to its semantic owner and propagate cancellation | Pass | Sharing start/close remain semantic commands; DOM callbacks pass `pageSignal$`, while the existing scroll DOM lifecycle owns the animation-frame work and cancellation. |
| `docs/cache.md` | Keep state/resource ownership scoped and preserve cleanup | Pass | Sharing is composed with the same thread's `messages.scroll` signals; no new global cache, duplicate mutable source, listener, observer, or resource lifetime is introduced. |
| `.claude/skills/ccstate/SKILL.md` | Compose commands in signals; use `pageSignal$` and `detach(..., Reason.DomCallback)` only in views | Pass | Signal code returns `set(scroll.autoScroll$, ...)`; all four updated React entry points use `pageSignal$` plus tracked `detach`, and no `detach` was added under `signals/`. |
| `docs/testing/patterns.md` | Platform tests use `detachedSetupPage` and user-facing interactions | Pass | The test enables the feature through page setup and uses `userEvent` to enter and cancel sharing. |
| `docs/testing/anti-patterns.md` | Avoid internal mocks, fake timers, direct rendering, and mock-call-only assertions | Pass | The added case uses the established page/layout harness and outcome assertions; it adds none of the prohibited patterns. |
| `docs/testing/app-testing.md` | Page-visible platform behavior must use production bootstrap and lifecycle-owned mocks | Pass | The regression lives under `views`, calls `detachedSetupPage({ context, ... })`, and waits for the rendered state after each click. |
| `docs/testing/testing-external-behavior.md` | Construct and observe behavior through the page boundary | Pass | The test drives the share and cancel buttons and observes the chat viewport; it does not mutate the ccstate store or call signal commands directly. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- Coverage: Adequate. The new page regression covers entering and exiting sharing while scrolled away from the tail; existing scroll tests cover tail following and resize restoration.
- Conventions: Compliant with the selected platform, ccstate, and external-behavior practices.
- CI: All published checks on `185d542001a615ac61aa06ad28b750ed4c6ae39f` pass, including format, lint, type checks, app tests, security scans, and preview deployment.
- Preview verification: Desktop and mobile were exercised at the bottom, within the 5 px tail threshold, mid-thread, and at the top. Tail distance remained zero; event anchors stayed stable (desktop exact, mobile within 0.5 px during real reflow); normal and sharing headers measured 56 px on desktop and 48 px on mobile.

PR #25503, opened 2026-08-06
fix(platform): render markdown emphasis that touches cjk punctuation

bingjiez666:
## Problem

Assistant messages written in Chinese/Japanese/Korean regularly render their
emphasis markers as literal text:

| Source | Rendered today |
| --- | --- |
| `**加粗（x）**后面` | `**加粗（x）**后面` |
| `*斜体（x）*后面` | `*斜体（x）*后面` |
| `他说**「重要」**的事` | `他说**「重要」**的事` |
| `~~删除线（test）~~后面` | `~~删除线（test）~~后面` |

This is CommonMark behaving as specified, not a renderer bug. A closing `**`
is only a *right-flanking delimiter run* when it is "either (2a) not preceded
by a Unicode punctuation character, or (2b) preceded by punctuation **and**
followed by whitespace or punctuation" (spec 0.30 §6.2). `（）。，：「」《》！…、`
are all Unicode punctuation, and CJK writes no spaces between a closing bracket
and the next word — so clause (2b) fails and the delimiter never closes.

It is a long-standing spec issue: commonmark/commonmark-spec#650,
[Emphasis and East Asian text](https://talk.commonmark.org/t/emphasis-and-east-asian-text/2491).

## Fix

Add [`remark-cjk-friendly`](https://github.com/tats-u/markdown-cjk-friendly)
and its GFM strikethrough companion to the shared `Markdown` component, so all
markdown surfaces (chat, artifact sidebar, log cards, activity detail) are
covered in one place. The package implements the CommonMark revision candidate
for this problem and only reclassifies CJK punctuation — non-CJK parsing is
untouched. It is the same extension Redmine 6.1.1 enabled in commonmarker
(redmine.org/issues/43234).

`pluginsFilter` is needed because the strikethrough companion has to replace
`remark-gfm`'s own `~~` extension and therefore must run *after* it, while
`@uiw/react-markdown-preview` builds `[remarkAlert, ...caller, gfm]` and always
places caller plugins first. Measured:

```
gfm, cjk, strike  -> ~~ OK        cjk, gfm, strike  -> ~~ OK
gfm, strike, cjk  -> ~~ OK        cjk, strike, gfm  -> ~~ BROKEN  <- @uiw default
```

`remark-cjk-friendly` itself is order-independent.

## Feature switch

Gated behind `cjkFriendlyMarkdown`, **defaulting to enabled**. This changes how
every rendered message is parsed, so the flag exists as a one-click rollback
rather than a staged rollout — the fix is the desired behavior and non-CJK
output is byte-identical. Flipping to a staged launch is a one-line change to
the registry entry (`enabled: false` + `enabledOrgIdHashes: STAFF_ORG_ID_HASHES`).

Turning it off drops the `pluginsFilter` reorder as well, otherwise the filter
would re-append the strikethrough companion that `buildRemarkPlugins` just left
out.

## Scope — what this does not fix

Deliberately out of scope, tracked separately:

- **GFM autolink literals.** `已完成：**https://github.com/vm0-ai/vm0/pull/25430**（draft）`
  still renders wrong, for an unrelated reason: the autolink scanner runs to the
  next ASCII space and only trims the ASCII set `? ! . , : * _ ~`, so it swallows
  `**（draft）` into the href and orphans the opening `**`. The same mechanism
  corrupts plain bare URLs (`链接：https://example.com/a（draft）` → dead link).
  Fixing it correctly means extending GFM's parenthesis-balancing to fullwidth
  brackets; naive tail-trimming breaks legitimate URLs such as
  `https://zh.wikipedia.org/wiki/中文（消歧义）`.
- **`_` / `__` emphasis**, which fails under CommonMark's intraword-underscore
  rule (the one that keeps `snake_case` from italicizing). Unrelated to CJK.
- **Pure-ASCII `**markdown(md)**is`**, which the extension leaves alone by design.
- **Slack / Feishu delivery**, which hand the raw text to the destination's own
  parser and are unreachable from the renderer.

## Verification

- New page-level tests in `markdown.test.tsx` cover `*`, `**`, `***`, the
  opener-side case, `~~` (the `pluginsFilter` regression guard), and the
  switch-off fallback. Each was confirmed to fail without its corresponding
  change and pass with it.
- ASCII regression test asserts `**bold**`, `*em*`, `~~del~~`, GFM tables and
  task lists still render.
- `check-types`, `lint` and `knip` green for both `apps/platform` and
  `packages/core`; `vitest run src/views/components/__tests__/markdown.test.tsx`
  20 passed; `@vm0/core` suite 190 passed.

### DIFF
--- a/turbo/apps/platform/package.json
+++ b/turbo/apps/platform/package.json
@@ -50,6 +50,8 @@
     "react-i18next": "^17.0.11",
     "react-zoom-pan-pinch": "4.0.3",
     "rehype-katex": "^7.0.1",
+    "remark-cjk-friendly": "^2.3.1",
+    "remark-cjk-friendly-gfm-strikethrough": "^2.3.1",
     "remark-math": "^6.0.0",
     "signal-timers": "^1.0.4",
     "unicode-emoji-json": "^0.9.0"
--- a/turbo/apps/platform/src/signals/external/feature-switch.ts
+++ b/turbo/apps/platform/src/signals/external/feature-switch.ts
@@ -77,6 +77,10 @@ export const artifactSidebarInlineOpenEnabled$ = computed((get): boolean => {
   );
 });
 
+export const cjkFriendlyMarkdownEnabled$ = computed((get): boolean => {
+  return get(featureSwitch$)[FeatureSwitchKey.CjkFriendlyMarkdown] ?? false;
+});
+
 export const chatThreadSidebarAutoOpenEnabled$ = computed((get): boolean => {
   return (
     get(featureSwitch$)[FeatureSwitchKey.ChatThreadSidebarAutoOpen] ?? false
--- a/turbo/apps/platform/src/views/components/__tests__/markdown.test.tsx
+++ b/turbo/apps/platform/src/views/components/__tests__/markdown.test.tsx
@@ -525,4 +525,77 @@ describe("assistant markdown", () => {
       expect(link).toHaveAttribute("rel", "noopener noreferrer");
     });
   });
+
+  // CJK sentences put punctuation directly against the closing delimiter with
+  // no space, which plain CommonMark refuses to close.
+  it("emphasizes text wrapped in delimiters that touch cjk punctuation", async () => {
+    mockThread(
+      [
+        "**加粗（x）**后面",
+        "",
+        "*斜体（x）*后面",
+        "",
+        "***粗斜（x）***后面",
+        "",
+        "他说**「重要」**的事",
+      ].join("\n"),
+    );
+
+    detachedSetupPage({ context, path: "/chats/thread-markdown" });
+
+    await waitFor(() => {
+      expect(
+        screen.getByText("加粗（x）", { selector: "strong, b" }),
+      ).toBeInTheDocument();
+    });
+    expect(
+      screen.getByText("斜体（x）", { selector: "em, i" }),
+    ).toBeInTheDocument();
+    expect(
+      screen.getByText("粗斜（x）", { selector: "em strong, strong em" }),
+    ).toBeInTheDocument();
+    expect(
+      screen.getByText("「重要」", { selector: "strong, b" }),
+    ).toBeInTheDocument();
+  });
+
+  // Guards the `pluginsFilter` reorder: the strikethrough companion only wins
+  // over `remark-gfm`'s own `~~` extension when it runs after it.
+  it("strikes through text that touches cjk punctuation", async () => {
+    mockThread("~~删除线（test）~~后面");
+
+    detachedSetupPage({ context, path: "/chats/thread-markdown" });
+
+    await waitFor(() => {
+      expect(
+        screen.getByText("删除线（test）", { selector: "del, s" }),
+      ).toBeInTheDocument();
+    });
+  });
+
+  it("keeps ascii markdown rendering unchanged", async () => {
+    mockThread(
+      [
+        "**bold**, *em*, ~~del~~",
+        "",
+        "| a | b |",
+        "| --- | --- |",
+        "| 1 | 2 |",
+        "",
+        "- [x] done",
+      ].join("\n"),
+    );
+
+    detachedSetupPage({ context, path: "/chats/thread-markdown" });
+
+    await waitFor(() => {
+      expect(
+        screen.getByText("bold", { selector: "strong, b" }),
+      ).toBeInTheDocument();
+    });
+    expect(screen.getByText("em", { selector: "em, i" })).toBeInTheDocument();
+    expect(screen.getByText("del", { selector: "del, s" })).toBeInTheDocument();
+    expect(screen.getByRole("table")).toBeInTheDocument();
+    expect(screen.getByRole("checkbox")).toBeChecked();
+  });
 });
--- a/turbo/apps/platform/src/views/components/__tests__/markdown.test.tsx
+++ b/turbo/apps/platform/src/views/components/__tests__/markdown.test.tsx
@@ -541,7 +541,11 @@ describe("assistant markdown", () => {
       ].join("\n"),
     );
 
-    detachedSetupPage({ context, path: "/chats/thread-markdown" });
+    detachedSetupPage({
+      context,
+      path: "/chats/thread-markdown",
+      featureSwitches: { [FeatureSwitchKey.CjkFriendlyMarkdown]: true },
+    });
 
     await waitFor(() => {
       expect(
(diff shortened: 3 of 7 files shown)


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 8dc2aa8fcbf1f92c341027fc442b8acd4a3bd22b
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Adds `remark-cjk-friendly` and `remark-cjk-friendly-gfm-strikethrough` to the shared platform `Markdown` component so `*`/`**`/`***`/`~~` close correctly when a CJK punctuation character sits directly against the delimiter, and gates the behavior behind a new `cjkFriendlyMarkdown` feature switch (default `enabled: true`). The `pluginsFilter` prop is used to move the strikethrough companion behind `remark-gfm`, which `@uiw/react-markdown-preview` always appends last.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no wrong API assumptions | Pass | `markdown.tsx:422-432` — filter removes then re-appends the companion, so no duplicate entry; identity comparison holds because `buildRemarkPlugins` pushes the bare function reference. The `@uiw` order `[remarkAlert, ...caller, gfm]` (`preview.tsx`) makes "append to tail" equivalent to "after gfm". |
| `REVIEW.md` | Security (injection, secrets, auth) | N/A | No endpoint, credential, query, or auth path changed; parser plugins operate on already-sanitized markdown and `rehypeRewrite`/`escapeHtmlTags` are untouched. |
| `REVIEW.md` | Type safety, no `any` / broad types | Pass | `markdown.tsx:408-409` derives `RemarkPlugins`/`PluginsFilter` from `MarkdownPreviewProps` via `NonNullable`; no `any`, no assertions. |
| `REVIEW.md` | Style and maintainability; no oversized or duplicated logic | Pass | `buildRemarkPlugins` mirrors the existing `buildRehypePlugins` shape directly below it; both new functions are under 15 lines. |
| `REVIEW.md` | `fix:` commit must carry a regression test | Pass | Four new tests in `markdown.test.tsx:529-629`, including the exact failing inputs. |
| `REVIEW.md` | New user-facing behavior gated behind a `FeatureSwitchKey` | Pass | `feature-switch-key.ts:77`, `feature-switch.ts:398-404`, `signals/external/feature-switch.ts:80-82`, consumed at `markdown.tsx:485`. |
| `REVIEW.md` | Feature regression guard — existing flow most likely to regress is covered | Pass | `keeps ascii markdown rendering unchanged` covers `**bold**`, `*em*`, `~~del~~`, GFM table and task list. Verified non-vacuous: `page-helper.ts:143-149` seeds the cache from `getAllFeatureStates`, so with no explicit `featureSwitches` the switch resolves to `true` and the assertions run against the new plugins. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No `packages/db/src/schema/**`, contract, or `jsonb().$type<>()` change. |
| `docs/bad-smell.md` | Zero lint/type suppression comments | Pass | No `eslint-disable`, `@ts-ignore`, `@ts-expect-error`, or `prettier-ignore` in the diff. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | Both plugins are static top-level imports (`markdown.tsx:9-10`). |
| `docs/bad-smell.md` | No unnecessary try/catch; fail fast | Pass | No error handling added. |
| `docs/bad-smell.md` | No hardcoded URLs or environment values | Pass | None added. |
| `docs/effect.md` | Derived value uses `computed`, not an effect | Pass | `cjkFriendlyMarkdownEnabled$` is a `computed` over `featureSwitch$`, matching the four sibling switch signals in the same file. |
| `docs/effect.md` | Signals owned outside render; render stays pure | Pass | The signal is module-scope; the component only reads it with `useGet`, alongside the pre-existing `theme$`/`pageLifecycleId$` reads. |
| `docs/cache.md` | RC-2 no render-time signal allocation | Pass | No `state()`/`computed()` created inside `Markdown`. |
| `docs/cache.md` | RC-7 stable callback identity, no lost cleanup | Pass | `reorderCjkStrikethrough` is module-scope, so `pluginsFilter` stays referentially stable across renders (or `undefined`), preserving the file's existing stable-identity intent for `components`. |
| `docs/cache.md` | RC-3/RC-8 no unbounded cache or asymmetric resource lifecycle | N/A | No cache, subscription, or disposable resource introduced. |
| `.claude/skills/ccstate/SKILL.md` | No `detach()` in signals; derived state via `computed` bound to parent | Pass | `signals/external/feature-switch.ts:80-82` adds only a `computed`; no `detach()`, no new command. |
| `docs/deployment-compatibility.md` | Cross-version compatibility for independently deployed components and persisted state | Pass | `getAllFeatureStates` resolves the new key from the client-bundled registry, so a new frontend gets `true` without an API change; `applySwitches` iterates `Object.values(FeatureSwitchKey)` from its own bundle, so an old frontend ignores the new key. No persisted schema or cache-key version change (`vm0:feature-switch-cache:v3` unchanged). |
| `docs/testing.md` | Integration tests only, no unit tests for internal functions | Pass | All four tests go through `detachedSetupPage` + `mockThread`, not through `buildRemarkPlugins`/`reorderCjkStrikethrough` directly. |
| `docs/testing.md` | Mock at boundary; `vi.mock()` must not use `../` paths | Pass | No `vi.mock` added. |
| `docs/testing.md` | MSW for HTTP, no direct `fetch` stubbing | Pass | Thread data comes from `context.mocks.api(chatThreadEventsContract.list, ...)`. |
| `docs/testing.md` | No fake timers | Pass | None used; waiting is via `waitFor`. |
| `docs/testing.md` | Assert behavior, not mock calls | Pass | Assertions are on rendered `strong`/`em`/`del`/`table`/`checkbox` output. |
| `docs/testing/app-testing.md` | Page tests enter through `detachedSetupPage` and assert page-visible behavior | Pass | `markdown.test.tsx:529-629` follows the file's established `mockThread` + `detachedSetupPage({ path: "/chats/thread-markdown" })` idiom. |
| `docs/testing/app-testing.md` | No component-only unit tests under `apps/platform` | Pass | The new tests render the chat page, not the component in isolation. The two pre-existing direct `render(<Markdown/>)` tests are untouched by this PR. |
| `docs/testing/anti-patterns.md` | AP-10 no direct component rendering | Pass | No new direct-render test added. |
| `docs/testing/anti-patterns.md` | AP-7 no testing implementation details | Pass (see Nit 3) | Assertions target user-visible output; the switch-off test additionally uses the `.wmde-markdown` container selector already used by existing tests in this file. |
| `docs/testing/testing-external-behavior.md` | Do not mutate the store directly or call hooks directly | Pass | Switch state is supplied through the `featureSwitches` option of `detachedSetupPage`, not by store mutation. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Nits (non-blocking)

1. `markdown.tsx:503` — `pluginsFilter` is passed as a normal prop but is **not** destructured out of `props`, while `remarkPlugins` and `rehypePlugins` are. Since `{...rest}` is spread last, a future caller that passes its own `pluginsFilter` would silently override the reorder and re-break CJK strikethrough with no test failure outside this file. No current caller does this, so it is latent only.
2. `turbo/pnpm-lock.yaml` — the diff carries an unrelated change: `@electron/node-gyp` gained an `integrity` field on its git-hosted `resolution`. Incidental churn from `pnpm add`; harmless, but it is not part of this PR's intent and may flip back on another machine.
3. `markdown.test.tsx:591-604` — the switch-off test asserts through `document.querySelector(".wmde-markdown")`, a class owned by `@uiw/react-markdown-preview`. It matches the existing pattern in this file (`shows a raw style block as text`), so it is consistent rather than wrong, but `screen.getByText` would be less coupled to the library.
4. `signals/external/feature-switch.ts:81` — the `?? false` fallback means a returning user with a warm `vm0:feature-switch-cache:v3` entry written before this deploy sees stock CommonMark until `reloadFeatureSwitch$` refreshes the cache. This matches every sibling switch in the file, but those all default to `enabled: false`, so the fallback direction is only surprising for this default-on switch. The window is one bootstrap refresh.

### Testing

- **Coverage: Adequate.** `fix:` commit carries regression tests for the reported failure (`**加粗（x）**后面`), plus `*`, `***`, the opener-side case (`他说**「重要」**的事`), the `~~` case that specifically guards the `pluginsFilter` reorder, the switch-off fallback, and an ASCII/GFM non-regression test. The PR reports each new test was confirmed to fail without its corresponding change; the ASCII guard was additionally verified non-vacuous against `page-helper.ts` switch seeding during this review.
- **Conventions: Compliant.** Page-level entry via `detachedSetupPage`, MSW-backed contract mocks, no `vi.mock`, no fake timers, no fetch stubbing, no assertions on mock calls.
- **Gaps worth noting, consistent with the PR's stated scope:** no coverage for the GFM autolink-literal case, `_`/`__` emphasis, or pure-ASCII `**markdown(md)**is`. The PR body documents each as deliberately out of scope with rationale, including why naive autolink tail-trimming is unsafe. Slack/Feishu delivery is unaffected by this change because those paths hand raw text to the destination's own parser.

PR #25558, opened 2026-08-07
fix: parse pdfs in zero scrape instead of returning base64

bingjiez666:
## Problem

`zero scrape` cannot read PDFs. Pointing it at any PDF URL returns a base64 blob:

```
$ zero scrape https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf --json
.format          = markdown
.result.markdown = JVBERi0xLjQKJcOkw7zDtsOfCjIgMCBvYmoKPDwvTGVuZ3Ro...
```

The cause is `parsers: []` in the Firecrawl request. Firecrawl documents that value as: *"When an empty array is passed, the PDF file is returned in base64 encoding with a flat rate of 1 credit for the entire PDF."* It has been that way since the endpoint was introduced in #20778.

This matters most for text-only models. Claude and GPT can take a PDF as a native document block, so they never hit this path. DeepSeek V4 Flash — the default model — cannot, and the sandbox ships no `pypdf` / `pymupdf` / `pdfminer` / `PIL` / `poppler`, so an agent on the default model has no way to read a PDF at all.

## Change

Send `parsers: [{ type: "pdf", maxPages: 100 }]`. Firecrawl's default `auto` mode extracts the embedded text layer first and falls back to OCR for scanned pages, so no routing logic is needed on our side.

Non-PDF pages are unaffected: `parsers` only applies to document inputs, and Firecrawl still charges a flat credit for an HTML scrape.

### Billing

Firecrawl charges **1 credit per parsed page** rather than a flat credit, so a flat `billingQuantity: 1` would have left vm0 paying per page while charging per request. Usage is now recorded with `quantity` = pages parsed, which `recordManagedUsage$` already supports. A web page still bills 1 unit because Firecrawl only reports `numPages` for parsed documents.

`maxPages: 100` bounds both the per-request cost and the work done inside the 25s provider timeout.

### Truncation is visible

Firecrawl returns `numPages` (parsed) and `totalPages` (true count). Both are now passed through to response metadata, and the CLI prints a warning when they differ, so a truncated 240-page report does not silently look complete:

```
✓ Scrape completed
  Billing quantity: 100
  Credits charged: 400
  Pages parsed: 100
  Truncated: parsed the first 100 of 240 pages
```

## Notes for review

- **No new billing category.** PDFs bill under the existing `standard.markdown` / `enhanced.markdown` categories, so this needs no new `usage_pricing` row and cannot 503 on a missing one.
- **The credit pre-check still runs at quantity 1**, since page count is unknown before the scrape returns. `maxPages` caps the worst-case overshoot at 100 units. Splitting the check into a HEAD request for the page count seemed like more machinery than the exposure justifies — happy to add it if you disagree.
- **This does not cover local files.** `zero scrape` takes a URL, so an uploaded PDF still has no path. Firecrawl's `/v2/parse` accepts `multipart/form-data` uploads and would cover it; that belongs in a separate change.

## Testing

- `pnpm --dir apps/api exec vitest run scrape` — 22 passed, including a new case asserting per-page billing and truncated page counts
- `pnpm --dir apps/cli exec vitest run` — 889 passed
- `pnpm --dir packages/api-contracts exec vitest run` — 529 passed
- `pnpm check-types`, `pnpm turbo run lint` (0 errors), `pnpm format`, `pnpm knip` (output unchanged from baseline)

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/zero-scrape.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/zero-scrape.test.ts
@@ -442,7 +442,7 @@ describe("zero scrape route", () => {
     expect(requestBody).toStrictEqual({
       url: "https://example.com/page",
       formats: ["markdown"],
-      parsers: [],
+      parsers: [{ type: "pdf", maxPages: 100 }],
       proxy: "basic",
       skipTlsVerification: false,
       maxAge: 0,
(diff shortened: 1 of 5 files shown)


bingjiez666:
Closed per request. Note: this PR was mergeable with green checks at the time of closing; reopening is possible if the PDF parsing change is still wanted.

PR #25560, opened 2026-08-07
refactor: replace @tabler/icons-react with lucide-react

### LINKED ISSUE #25458: refactor: replace @tabler/icons-react with lucide-react across the UI
## Summary

We want to move the whole product off **`@tabler/icons-react`** and onto **[Lucide](https://lucide.dev/icons/)** (`lucide-react`) as the single icon set.

Motivation:
- Lucide is the icon set shadcn/ui is built around, and `packages/ui` is a shadcn-based library — today our own primitives (`dialog`, `select`, `sheet`, `command`, `checkbox`, `multi-select-combobox`, `copy-button`) hand-patch Tabler icons into components that upstream ships with Lucide. Aligning removes that friction on every future component we pull in.
- One icon set, one stroke/optical-size convention, ISC-licensed and actively maintained.

## Current usage

Measured on `main`:

| | |
|---|---|
| Files importing `@tabler/icons-react` | **130** |
| Distinct icons in use | **162** |
| Call sites passing the Tabler-only `stroke={...}` prop | **386** |

Spread across the workspace:

| Package | Files |
|---|---|
| `turbo/apps/platform` | 118 |
| `turbo/packages/ui` | 7 |
| `turbo/apps/desktop` | 5 |

`@tabler/icons-react@3.44.0` is declared in `turbo/apps/platform/package.json`, `turbo/apps/desktop/package.json` and `turbo/packages/ui/package.json`.

## Feasibility: the mapping is essentially complete

I diffed all 162 icons we use against the full Lucide set (`lucide-static@1.28.0`, 2007 icons):

- **104 map 1:1** — the name is identical once the `Icon` prefix is dropped (`IconCheck` → `Check`, `IconSearch` → `Search`, `IconTrash2` → `Trash2`, …).
- **52 need a rename** — full table below.
- **5 have no Lucide equivalent** — see the open question.
- **0 remaining unmapped.**

<details>
<summary>Full rename table (52)</summary>

| Tabler | Lucide |
|---|---|
| `IconActivityHeartbeat` | `Activity` |
| `IconAdjustmentsHorizontal` | `SlidersHorizontal` |
| `IconArrowsDiagonal` | `Maximize2` |
| `IconArrowsDiagonalMinimize2` | `Minimize2` |
| `IconArrowsSort` | `ArrowUpDown` |
| `IconAspectRatio` | `RectangleHorizontal` |
| `IconBlockquote` | `Quote` |
| `IconBrowser` | `AppWindow` |
| `IconBrowserOff` | `AppWindowMac` |
| `IconCalendarTime` | `CalendarClock` |
| `IconChecks` | `CheckCheck` |
| `IconCircleCheckFilled` | `CircleCheckBig` |
| `IconCircleHalf2` | `Contrast` |
| `IconClockExclamation` | `ClockAlert` |
| `IconColorSwatch` | `SwatchBook` |
| `IconDatabaseExport` | `DatabaseBackup` |
| `IconDeviceDesktop` | `Monitor` |
| `IconDice` | `Dices` |
| `IconDots` | `Ellipsis` |
| `IconDotsVertical` | `EllipsisVertical` |
| `IconFileInvoice` | `ReceiptText` |
| `IconFilePencil` | `FilePen` |
| `IconFlask` | `FlaskConical` |
| `IconH1` | `Heading1` |
| `IconH2` | `Heading2` |
| `IconH3` | `Heading3` |
| `IconHandStop` | `Hand` |
| `IconInfoCircle` | `Info` |
| `IconLayoutSidebarLeftCollapse` | `PanelLeftClose` |
| `IconListNumbers` | `ListOrdered` |
| `IconLogout` | `LogOut` |
| `IconMenu2` | `Menu` |
| `IconMessagePlus` | `MessageSquarePlus` |
| `IconMessages` | `MessagesSquare` |
| `IconMicrophone` | `Mic` |
| `IconMoodPlus` | `SmilePlus` |
| `IconPhoto` | `Image` |
| `IconPinnedOff` | `PinOff` |
| `IconPlayerPause` | `Pause` |
| `IconPlayerPlay` | `Play` |
| `IconPlayerStop` | `Square` |
| `IconPlugConnected` | `PlugZap` |
| `IconPresentationAnalytics` | `Presentation` |
| `IconRefresh` | `RefreshCw` |
| `IconRobot` | `Bot` |
| `IconShare3` | `Share2` |
| `IconSpeakerphone` | `Megaphone` |
| `IconSwitchHorizontal` | `ArrowRightLeft` |
| `IconTemplate` | `LayoutTemplate` |
| `IconTool` | `Wrench` |
| `IconWorld` | `Globe` |
| `IconZoomReset` | `RotateCcw` |

</details>

## Open question — brand icons

Lucide does not ship brand/logo icons. These 5 are used today and have **no** Lucide replacement:

`IconBrandGithub`, `IconBrandGoogleDrive`, `IconBrandNotion`, `IconBrandSlack`, `IconBrandTelegram`

Options:
1. Vendor the 5 SVGs into `packages/ui` as local components (smallest surface, no new dependency).
2. Add `simple-icons` / `@icons-pack/react-simple-icons` for brand marks only.

I'd lean toward **option 1** — 5 icons doesn't justify a second icon dependency, and connector logos are arguably product assets rather than an icon set.

## API differences to handle

- **`stroke` → `strokeWidth`.** Tabler takes `stroke={1.5}`; Lucide takes `strokeWidth`. 386 call sites, mechanically rewritable but must not be missed — a stray `stroke={1.5}` on a Lucide icon silently renders as an SVG paint attribute and can blank the icon out.
- **Default stroke width.** Tabler defaults to `2`, and most of our call sites explicitly soften it to `1.5`. Lucide also defaults to `2`. We should decide on a house default once and set it centrally rather than repeating `strokeWidth={1.5}` 386 times.
- **`size`** behaves the same in both.
- **Type imports.** `import type { Icon } from "@tabler/icons-react"` and the `typeof IconClock` patterns (e.g. `workflow-detail-page.tsx`, `zero-chat-composer.tsx`, `personal-usage-record.tsx`, `build-info-block.tsx`) become `LucideIcon`.

## Proposed approach

1. Add `lucide-react`, vendor the 5 brand SVGs into `packages/ui`.
2. Codemod the imports + renames using the table above; hand-review the 52 renames since a few are judgement calls (`IconAspectRatio` → `RectangleHorizontal`, `IconCircleHalf2` → `Contrast`, `IconBrowserOff` → `AppWindowMac`).
3. Migrate `stroke` → `strokeWidth` and centralize the default.
4. Land `packages/ui` first, then `platform`, then `desktop`, so the shared primitives don't straddle two icon sets.
5. Visual pass over the main surfaces — Lucide's optical weight differs slightly from Tabler at small sizes, so 14px icons are the ones worth eyeballing.
6. Drop `@tabler/icons-react` from all three `package.json` files and add a lint rule banning re-import.

Given 130 files, this is best split into a few PRs by package rather than one mega-diff.

tongx-blip:
Closes #25458.

Replaces `@tabler/icons-react` with [Lucide](https://lucide.dev/icons/) as the single icon set across `packages/ui`, `apps/platform` and `apps/desktop`.

`packages/ui` is shadcn-based and shadcn ships with Lucide upstream, so today every component we pull in needs its Tabler icons hand-patched back in. This removes that friction and leaves one stroke/optical convention across the product.

## What changed

**All 162 icons in use were mapped, with no gaps:**

| | |
|---|---|
| Same name minus the `Icon` prefix | 97 |
| Renamed | 59 |
| Brand marks vendored (no Lucide equivalent) | 5 |
| Unmapped | 0 |

Representative renames: `IconLoader2` → `LoaderCircle`, `IconDotsVertical` → `EllipsisVertical`, `IconPhoto` → `Image`, `IconAlertTriangle` → `TriangleAlert`, `IconAdjustmentsHorizontal` → `SlidersHorizontal`, `IconTrash` → `Trash2`, `IconMenu2` → `Menu`.

**Brand icons.** Lucide ships no brand marks, so GitHub, Google Drive, Notion, Slack and Telegram are vendored into `packages/ui/src/components/ui/brand-icons.tsx`. They keep the Tabler path geometry (MIT), so the marks render exactly as before, and they take the same props as a Lucide icon (`size`, `strokeWidth`, `absoluteStrokeWidth`, `color`) so they stay drop-in. I chose this over adding `simple-icons` — 5 icons don't justify a second icon dependency.

**`stroke` → `strokeWidth`.** Tabler's stroke-width prop is `stroke`; Lucide's is `strokeWidth`. 385 call sites were rewritten. This is the sharpest edge in the migration: a missed `stroke={1.5}` on a Lucide icon is *not* a type error — it falls through to the native SVG paint attribute and silently renders the icon invisible. The rewrite resolves each occurrence's owning JSX tag by backward scan, so the 6 legitimate `stroke` attributes on raw `<svg>` / `<path>` / `<line>` elements were left untouched. Verified 0 component tags still pass `stroke`.

**Name collisions.** Three icons collided with identifiers already in the file and are imported aliased rather than renamed around: `Link as LinkIcon` (two files where `Link` is the router component) and `Image as ImageIcon`.

**Hand-inlined SVGs.** `tiptap-workflow-composer.ts` builds five icons through the DOM API rather than JSX, with Tabler path data pasted inline. Those are repointed to the matching Lucide geometry (`MessageCircle`, `Quote`, `X`, `FileText`, `SwatchBook`) so the composer doesn't drift from the rest of the UI.

## Verification

- `pnpm check-types` — passes for `@vm0/ui`, `@vm0/app`, `@vm0/desktop`. This is the strongest check here: it proves every one of the 162 mapped names actually resolves in `lucide-react`. (`api#check-types` OOMs in my sandbox at its 3.5 GB heap setting; `apps/api` has zero changed files and no `@vm0/ui` dependency, so it's left to CI.)
- `pnpm turbo run lint` — 0 errors. Warning count on `@vm0/app` is 58, identical to `main`.
- `pnpm -F @vm0/app build` — succeeds.
- `pnpm knip` — no new findings.
- No test couples to Tabler CSS classes, and there are no snapshots.

## Worth eyeballing on the preview

These are the places where Lucide genuinely differs from Tabler and a human should look:

1. **`IconCircleCheckFilled` → `CircleCheckBig` (4 sites).** Lucide has no filled variants at all. Previously a solid filled check; now an outline. Most visible on the onboarding workflow card selected state.
2. **Quote icon direction** in the composer quote chip. It keeps its existing `-scale-x-100` flip; Lucide's `Quote` may not need the flip. Left as-is rather than guessing.
3. **Optical weight at 14px and below.** Lucide's default geometry is a little different from Tabler's at small sizes; small inline icons are the ones to check.

## Follow-up, deliberately not in this PR

386 call sites set `strokeWidth` explicitly, mostly to `1.5`. That default belongs in one place rather than repeated per call site, but folding it in here would have buried the mechanical rename in a styling change. Worth a separate PR.


tongx-blip:
Preview for design acceptance:

**App (this is the one to look at): https://pr-25560-app.omby.ai/**

No token needed — app previews are served from Cloudflare Pages, not Vercel. The `pr-25560-app.vm6.ai` Vercel host 404s.

Sign in with a `+clerk_test` address (e.g. `test-icons+clerk_test@example.com`), Clerk OTP `424242`.

The Vercel automation bypass only applies to the **API** preview, and it goes in as a query parameter, not a path segment:

```
https://pr-25560-api.vm6.ai/health?x-vercel-protection-bypass=<VERCEL_AUTOMATION_BYPASS_SECRET>
```

Verified: `/health` returns 403 without it and 200 with it.

Places worth eyeballing: onboarding workflow-card selected state (`IconCircleCheckFilled` → `CircleCheckBig`, Lucide has no filled variants), the composer quote chip, and small 14px inline icons where optical weight differs slightly from Tabler.

tongx-blip:
Pushed a follow-up commit that puts every icon on a single stroke width.

**Why the weights looked inconsistent.** `packages/ui` had a global rule:

```css
svg[class*="tabler-icon"] { stroke-width: var(--icon-stroke-width); /* 1.5 */ }
```

`@tabler/icons-react` tags every icon `tabler-icon tabler-icon-*`, and a CSS rule outranks an SVG presentation attribute — so that rule was silently flattening **all** icons to 1.5 no matter what each call site passed. lucide-react emits `lucide` classes instead, so the migration left the rule matching nothing, and 11 different per-call-site values plus Lucide default 2 all became visible at once. Concretely: the sidebar rendered at 2 while the composer toolbar rendered at 1.5.

**The fix**, at Lucide default 2:

- retarget the rule to `svg.lucide`, set `--icon-stroke-width: 2`
- remove all 387 per-call-site `strokeWidth` props
- remove the inline `strokeWidth: 2.5` on the checkbox tick, which as an inline style outranked the rule
- give brand icons and the composer's DOM-built icons the `lucide` class, and mirror lucide-react by also keeping the width as an attribute (so desktop, which does not load `globals.css`, still lands on 2)
- delete the orphaned `ICON_STROKE` constant

Chart strokes in `usage-insight-bar-chart.tsx` are graphics, not icons, and keep their own widths.

**Verified on the deployed preview:**

```
--icon-stroke-width:2
svg.lucide{stroke-width:var(--icon-stroke-width)}
tabler-icon selector occurrences: 0
```

and the only non-2 `strokeWidth` literals left in the JS bundle are the usage chart's gridline (1) and its data line (`isHovered?2.5:1.75`).

tongx-blip:
Two more fixes on top, both traced to icons being painted translucent.

**1. The "overlay" look.** Overlapping strokes inside one SVG composite against each other, so a semi-transparent icon renders its joints darker than its body. Measured on the sidebar collapse button: `text-sidebar-foreground/70` gave the stroke crossing **1.31x** the ink of the stroke body. Rendering the same icon opaque measures **1.01x** — the artifact is entirely gone. Raising the stroke width to 2 widened the overlaps and made it obvious.

- dropped 106 neutral colour utilities (`text-muted-foreground` and friends) from icons, so an icon takes the colour of the text beside it
- dropped 14 `opacity-50/70/80` utilities from icons
- de-alphad 14 icon-only containers whose alpha reached the icon through `currentColor`, including both sidebar collapse buttons
- removed 11 `hover:` colour classes left redundant by the above

Semantic status colours (green / amber / destructive / primary / ...) and the `opacity-0` transition states are untouched.

**2. Sidebar icons greyer than their labels.** `apps/platform/src/views/css/index.css` had:

```css
```

Every nav icon rendered 30% transparent while its label rendered at full strength. On the light theme that resolves to exactly `rgb(89,92,96)` against a `rgb(21,24,30)` label — which matches what was measured off the reported screenshot. Removed, along with the `.zero-nav-recent-label { opacity: 1 }` counterpart it existed to offset.

**Verified on the deployed preview** — all four Manage rows now report `icon colour == label colour == rgb(21,24,30)`, `opacity: 1`, `stroke-width: 2px`, and a page-wide sweep finds **0** translucent icons.

tongx-blip:
Dropped the icon stroke from 2 to **1.75**.

Measured Noto Sans 400 by rendering at 140px and dividing: its vertical stem is **1.30px** at 14px. A 16px icon renders its stroke at `strokeWidth x 16/24`, so across the icon/text pairings actually used:

| stroke | mean icon ÷ text stem | range |
|---|---|---|
| 2 | 1.01 | 0.90 – 1.15 |
| **1.75** | **0.88** | 0.79 – 1.01 |
| 1.5 | 0.76 | 0.67 – 0.87 |

Stroke 2 is geometrically equal to the text stem, but an icon reads heavier than text at the same width — its strokes are isolated lines with round caps, while text stems come in clusters with counters between them. So equal geometry looks bolder, which is what was reported. 1.75 lands at 0.88 and reads as matched; 1.5 makes icons visibly weaker than the text.

Also mirrored `--icon-stroke-width` into `apps/desktop/src/renderer/styles.css`, which does not import `packages/ui` globals and would otherwise have stayed on lucide-reacts built-in default of 2.

No new colour tokens — this is geometry only.

Verified on the deployed preview CSS:

```
--icon-stroke-width:1.75
svg.lucide{stroke-width:var(--icon-stroke-width)}
```

tongx-blip:
Closing — we're keeping the current Tabler icon set, so the migration isn't happening.

The two rendering problems found while reviewing this (icons painted translucent, which produced both the "overlay" artifact on overlapping strokes and icons reading greyer than their labels) are real and independent of the icon library. They're fixed against the existing icons in #25728.

Issue #25458 stays open.

PR #25569, opened 2026-08-07
fix(runner): prevent drain from missing service restarts

### LINKED ISSUE #25482: bug(rust): prevent service drain from missing deactivating restarts
## Problem

`runner service drain` can report success while a service that is already `deactivating` remains eligible for an automatic restart.

`drain_with_ops` decides whether to install the runtime `Restart=no` override from its initial `is_active` result. `is_unit_active` intentionally classifies systemd's `deactivating` state as inactive. When that happens, drain skips both the override and `SIGUSR1`, runs `systemctl disable --no-reload`, coordinates a daemon reload, and returns `Ok(())`.

Disabling a unit only changes boot enablement; it does not suppress the loaded unit's `Restart=on-failure` policy. The generated runner unit uses that policy with `RestartSec=5`.

## Reachable failure sequence

1. A runner exits or fails and systemd exposes the unit as `ActiveState=deactivating` while completing teardown.
2. Deployment invokes `runner service drain` during that window.
3. The initial active check returns `false`, so drain does not write or verify the `Restart=no` override and does not send `SIGUSR1`.
4. `systemctl disable --no-reload` succeeds and drain returns success.
5. Once deactivation completes, systemd applies `Restart=on-failure` and starts a replacement runner.
6. The replacement never observed the drain signal and can remain active and claim work indefinitely.

This is a remaining race in the protection introduced for #20483 / #20496. That issue established that disabling the unit is insufficient and that restart suppression must prevent an unsignaled replacement `MainPID`; its plan placed the override before active-state and signal logic. The current inactive fast path leaves the same failure mode reachable specifically through `deactivating`.

## Code evidence

- `crates/runner/src/cmd/service/drain_resume.rs`: `drain_with_ops` sets `restart_override_written = should_signal` and writes the override only when the initial active check is true.
- `crates/runner/src/cmd/service/systemctl.rs`: `is_unit_active` documents and implements `deactivating` as false.
- `crates/runner/src/cmd/service/drain_resume.rs`: the disable operation is `systemctl disable --no-reload`, not a stop operation.
- `crates/runner/src/cmd/service/unit_file.rs`: installed services use `Restart=on-failure` and `RestartSec=5`.
- The existing `drain_inactive_service_skips_restart_override_and_signal` test fixes the current skip behavior but does not model a deactivating unit that subsequently restarts.

## Expected behavior

`service drain` must not report success while a runner service can still automatically restart into an unsignaled replacement process. A truly absent/inactive service should retain safe, bounded behavior without leaving unintended stale drain state.

## Acceptance criteria

- Cover the transition where the initial normal-health query reports inactive/deactivating and the unit can subsequently become active through `Restart=on-failure`.
- Ensure drain either suppresses that restart before returning success or detects/converges the race without leaving an unsignaled replacement runner.
- Preserve non-blocking drain semantics for active jobs; do not turn drain into `systemctl stop` or an unbounded wait.
- Preserve correct behavior for genuinely absent or fully inactive units.
- Add deterministic orchestration coverage proving a deactivating-to-restarted unit cannot produce a false successful drain.

seven332:
## Summary

- treat `deactivating` runner services as lifecycle-active so drain installs and verifies runtime `Restart=no` protection before signaling
- converge both `AlreadyGone` races and ambiguous initial signal failures until an active runner is signaled or systemd reaches a safe state
- keep convergence failures fail-closed with `Restart=no` retained and a shared 10-second deadline
- kill and reap bounded `systemctl` children when signal or follow-up state queries time out

## Scope

- changes runner CLI systemd drain and signal orchestration only
- does not change APIs, runner protocols, queue payloads, persisted data, migrations, or normal soft-drain semantics
- preserves rollback for failures before restart protection is verified; signal-delivery uncertainty retains that protection instead

## Validation

- `cargo test --manifest-path crates/Cargo.toml -p runner cmd::service::drain_resume`
- `git diff --check`

Closes #25482

### DIFF
--- a/crates/runner/src/cmd/service/drain_resume.rs
+++ b/crates/runner/src/cmd/service/drain_resume.rs
@@ -11,7 +11,7 @@ use crate::paths::HomePaths;
 use super::drain_override::{remove_drain_restart_override, write_drain_restart_override};
 use super::gate::read_runner_status;
 use super::reload::{SystemdReloadRequirement, coordinate_systemd_reload};
-use super::signal::{ServiceSignalOutcome, signal_service_main};
+use super::signal::{ServiceSignalOutcome, signal_service_main, signal_service_main_bounded};
 use super::systemctl::{
     CleanupUnitActiveState, SystemdUnitEnablement, cleanup_unit_active_state_bounded,
     get_service_restart_policy, is_unit_active_bounded, read_unit_enablement,
@@ -72,6 +72,11 @@ trait ServiceDrainOps {
     fn signal_drain<'a>(
         &'a mut self,
         unit: &'a RunnerServiceUnit,
+    ) -> ServiceFuture<'a, ServiceSignalOutcome>;
+    fn signal_drain_bounded<'a>(
+        &'a mut self,
+        unit: &'a RunnerServiceUnit,
+        timeout: Duration,
     ) -> ServiceFuture<'a, ServiceSignalOutcome>;
     fn disable<'a>(&'a mut self, unit: &'a RunnerServiceUnit) -> ServiceFuture<'a, ()>;
     fn restore_enablement<'a>(
@@ -158,6 +163,16 @@ impl ServiceDrainOps for RealServiceDrainOps {
         unit: &'a RunnerServiceUnit,
     ) -> ServiceFuture<'a, ServiceSignalOutcome> {
         Box::pin(async move { signal_service_main(unit, nix::sys::signal::Signal::SIGUSR1).await })
+    }
+
+    fn signal_drain_bounded<'a>(
+        &'a mut self,
+        unit: &'a RunnerServiceUnit,
+        timeout: Duration,
+    ) -> ServiceFuture<'a, ServiceSignalOutcome> {
+        Box::pin(async move {
+            signal_service_main_bounded(unit, nix::sys::signal::Signal::SIGUSR1, timeout).await
+        })
     }
 
     fn disable<'a>(&'a mut self, unit: &'a RunnerServiceUnit) -> ServiceFuture<'a, ()> {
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #25569

**Reviewed head:** `8d924bac44a640e9f8377b10ca3649a90413a3a6`

### Summary

Drain now distinguishes lifecycle-active deactivation from a truly inactive unit, converges an already-armed restart with the same soft signal, and bounds replacement signaling through child-owning kill-and-reap cleanup.

### Findings

No findings. The subprocess cancellation issue found during self-review is fixed in this head.

### Testing

Runner all-target/all-feature tests (3036 passed, 20 ignored), service tests, bounded-signal subprocess tests, Clippy, rustdoc, formatting, and diff checks passed. A forced real-systemd timing reproduction was intentionally not added; the transition is deterministic at the service orchestration boundary and existing systemd happy paths remain covered.

### Verdict

LGTM

seven332:
## Code Review: PR #25569

**Reviewed head:** `70450906fcc7c1a6b62e0e617aa50df5f7086d4b`

### Summary

Drain now protects lifecycle-active deactivation before signaling, converges already-armed restarts and ambiguous initial signal failures without restoring restart eligibility, and owns cleanup of timed-out `systemctl` children.

### Findings

No findings. The final state preserves the truly inactive fast path, normal soft-drain behavior, pre-verification rollback, bounded failure handling, and accurate operator diagnostics.

### Testing

Formatting, 39 focused drain tests, the full runner all-target/all-feature suite (3034 passed, 20 ignored, plus guest inventory), Clippy with warnings denied, rustdoc with warnings denied, and diff checks passed. A forced real-systemd timing reproduction was not added; the transition is covered deterministically at the established service orchestration boundary.

### Verdict

LGTM

PR #25578, opened 2026-08-07
fix(ably-subscriber): treat cleanup broken pipe as closed

### LINKED ISSUE #25557: bug(ably-subscriber): avoid warning on expected BrokenPipe during WebSocket cleanup
## Problem

Production runners emit `Failed to close websocket transport` when transport teardown calls `ws_write.close()` after the TCP peer is already gone and Linux returns `EPIPE`:

```text
WebSocket error: IO error: Broken pipe (os error 32)
```

The error occurs after the code has already decided to discard the transport. The background-close branch only logs it and does not propagate it to the runner, so it does not directly terminate a run. It nevertheless creates a false runner-health warning, and the same close result can make an explicit `close_and_wait()` report failed local cleanup.

## Production evidence

- Axiom dataset: `vm0-web-logs-prod`
- Service: `runner`
- Exact context: `ably_subscriber::connection::transport`
- Exact message: `Failed to close websocket transport`
- Exact error contains: `Broken pipe`
- Ten exact matches were observed across the Beijing-calendar days of August 5 and August 6, 2026.
- Four matches occurred during `2026-08-05T16:00:00Z` through `2026-08-06T16:00:00Z`; that query was complete (`isPartial=false`).
- The records contain no `run_id`, `runner_id`, client version, or connection identifier.
- The first exact production signature was observed after PR #25155 began tracking transport-close failures that had previously been discarded.

Reproduction query:

```apl
['vm0-web-logs-prod']
| where service == "runner"
| where context == "ably_subscriber::connection::transport"
| where message == "Failed to close websocket transport"
| where error contains "Broken pipe"
| project _time, _sysTime, context, error, level, message, service,
          run_id, runner_id, client_version, version
| sort by _time asc
```

## Confirmed code paths

- [`TransportCloseTracker::spawn`](https://github.com/vm0-ai/vm0/blob/a54deb69811cc546fc94d8c1be547eef16784256/crates/ably-subscriber/src/connection/transport.rs#L94-L105) runs detached transport cleanup and warns when no explicit waiter observes a failure.
- [`WsTransport::close`](https://github.com/vm0-ai/vm0/blob/a54deb69811cc546fc94d8c1be547eef16784256/crates/ably-subscriber/src/connection/transport.rs#L126-L137) drops the read half and awaits `ws_write.close()`.
- Caller-requested shutdown has a second WebSocket-close call in `send_close_message()`. An explicit `close_and_wait()` can observe both that direct close result and tracked background cleanup.
- The direct path first sends the Ably protocol `CLOSE` message and then performs the WebSocket close handshake. These two operations have different correctness semantics and must not be conflated.

Related work:

- #25110 / #25155 added awaitable graceful-shutdown completion and cleanup error tracking.
- #21353 / #21399 added explicit clean WebSocket shutdown for terminal Ably paths.

## Decision

Classify only `tungstenite::Error::Io` with `std::io::ErrorKind::BrokenPipe` returned by the WebSocket close operation as successful idempotent cleanup. Apply the same close-result classification at both WebSocket-close call sites.

Do not reclassify a failure to send the Ably protocol `CLOSE` message. A `BrokenPipe` from that send still means the documented explicit-close operation did not complete and must remain observable. Timeout, protocol, TLS, and all other I/O failures also retain their existing warning or propagation behavior.

This boundary fixes the false cleanup warning without weakening delivery or shutdown diagnostics.

## Alternatives considered

- **Lower the warning to info:** rejected because it changes only presentation; explicit cleanup can still report the expected idempotent close result as a failure.
- **Suppress `BrokenPipe` only in `TransportCloseTracker::spawn`:** rejected because background and explicit close paths would have inconsistent semantics and the transport layer would still surface a known successful teardown result.
- **Treat all peer-disconnect errors as success:** rejected because production evidence currently supports only `BrokenPipe`; broadening to `ConnectionReset`, `NotConnected`, or arbitrary I/O errors could hide distinct failures.

## Risks and mitigations

- **A real failure is hidden:** match the exact `tungstenite` I/O variant and `BrokenPipe` kind only at the WebSocket close-operation boundary.
- **Explicit close overstates success:** keep Ably protocol `CLOSE` send failures unchanged; only the subsequent WebSocket close handshake is idempotent after the peer has gone.
- **A socket-level test is timing-sensitive:** use a bounded local WebSocket integration test that deterministically forces the peer-close condition, and retain coverage proving an unrelated close failure still propagates.

## Acceptance criteria

- A WebSocket close operation returning `std::io::ErrorKind::BrokenPipe` is classified as successful local cleanup.
- The production background-cleanup condition no longer emits `Failed to close websocket transport` at warning level.
- Background cleanup and caller-requested shutdown use the same WebSocket-close classification.
- A failure to send the Ably protocol `CLOSE` message is not reclassified, even when its I/O kind is `BrokenPipe`.
- Timeout, protocol, TLS, and all other I/O failures retain their current warning or error propagation behavior.
- Integration coverage exercises the peer-already-closed behavior through a public subscription close entry point and demonstrates that unrelated failures are not swallowed.

seven332:
## Summary

- treat an exact `BrokenPipe` returned while closing a discarded WebSocket sink as successful idempotent cleanup
- share the classification between detached transport cleanup and caller-requested WebSocket shutdown
- preserve Ably protocol `CLOSE` send failures, timeouts, and every other WebSocket error
- cover the production interleaving with a synchronized local peer reset that previously emitted the warning

## Scope

This does not broaden accepted peer-disconnect errors, change the public API, or alter runner/API protocols or persisted state.

## Validation

- `RUSTDOCFLAGS="-D warnings" cargo doc -p ably-subscriber --all-features --no-deps`
- focused peer-reset regression repeated 20 times
- repository pre-commit Rust format, workspace Clippy, workspace documentation, file-size, and commit-message hooks

Closes #25557

### DIFF
--- a/crates/ably-subscriber/src/connection/event_loop.rs
+++ b/crates/ably-subscriber/src/connection/event_loop.rs
@@ -19,8 +19,8 @@ use super::message::{decode_data, message_targets_channel};
 use super::session::{SessionState, TokenRenewalFailure};
 use super::state::{ChannelLifecycleState, reconnect_spacing_delay, retry_delay};
 use super::transport::{
-    TransportCloseTracker, WsTransport, connect_pending, websocket_close_frame_reason,
-    websocket_close_reason, websocket_error_reason,
+    TransportCloseTracker, WsTransport, close_websocket_sink, connect_pending,
+    websocket_close_frame_reason, websocket_close_reason, websocket_error_reason,
 };
 use crate::Error;
 use crate::protocol::{
@@ -145,7 +145,7 @@ async fn send_close_message(p: &mut EventLoopState) -> Result<(), Error> {
                     .map_err(Error::from),
                 Err(error) => Err(error),
             };
-            let websocket_result = ws_write.close().await.map_err(Error::from);
+            let websocket_result = close_websocket_sink(&mut ws_write).await;
             match (protocol_result, websocket_result) {
                 (Ok(()), result) | (result, Ok(())) => result,
                 (Err(error), Err(close_error)) => {
--- a/crates/ably-subscriber/src/connection/transport.rs
+++ b/crates/ably-subscriber/src/connection/transport.rs
@@ -129,7 +129,7 @@ impl WsTransport {
             mut ws_write,
         } = self;
         drop(ws_read);
-        tokio::time::timeout(close_timeout, ws_write.close())
+        tokio::time::timeout(close_timeout, close_websocket_sink(&mut ws_write))
             .await
             .map_err(|_| Error::Protocol {
                 code: error_code::TIMEOUT,
@@ -144,6 +144,16 @@ impl WsTransport {
         tracker: &TransportCloseTracker,
     ) {
         tracker.spawn(self, close_timeout);
+    }
+}
+
+pub(super) async fn close_websocket_sink(ws_write: &mut WsWrite) -> Result<(), Error> {
+    match ws_write.close().await {
+        Ok(()) => Ok(()),
+        Err(tungstenite::Error::Io(error)) if error.kind() == std::io::ErrorKind::BrokenPipe => {
+            Ok(())
+        }
+        Err(error) => Err(error.into()),
     }
 }
 
--- a/crates/ably-subscriber/tests/integration/close_drop.rs
+++ b/crates/ably-subscriber/tests/integration/close_drop.rs
@@ -9,6 +9,8 @@ use std::collections::HashMap;
 use std::time::Duration;
 use tokio::io::{AsyncReadExt, AsyncWriteExt};
 use tokio_tungstenite::tungstenite;
+use tracing_subscriber::prelude::*;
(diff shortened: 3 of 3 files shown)


seven332:
## Code Review: PR #25578

**Reviewed head:** `a84ecba808cc51b65f486aade654fa5653dc6a87`

### Summary

The PR scopes idempotent handling to exact `Io(BrokenPipe)` from the WebSocket sink-close operation and routes both cleanup call sites through the same concrete helper. Ably protocol `CLOSE` sends, timeout handling, and every other WebSocket error retain their existing behavior.

### Findings

No findings.

### Testing

The synchronized real-socket regression reproduces the production read-reset/write-broken-pipe ordering without sleeps or retries and passed 20 consecutive focused runs. The complete `ably-subscriber` test suite, format, Clippy, rustdoc, and repository commit hooks passed locally. GitHub CI remains independently tracked by the workflow.

### Verdict

LGTM

PR #25599, opened 2026-08-07
test(guest-init): cover inherited signal dispositions

### LINKED ISSUE #25589: test(rust): cover inherited guest-init signal dispositions
## Problem and evidence

`SignalContext::setup()` in `crates/guest-init/src/pid1.rs` is the production path that configures `SIGTTIN`, `SIGTTOU`, and `SIGPIPE` as `SIG_IGN` before `guest-init` forks `vsock-guest`. `crates/guest-init/src/main.rs` explicitly relies on those dispositions surviving the fork into PID 2.

The existing PID 1 tests exercise `SignalContext::block()` and the supervision behavior, but none calls `setup()`. Removing one or all entries from the ignore loop therefore leaves the current tests green while changing the inherited child behavior: terminal job-control signals can stop the child, and `SIGPIPE` can terminate it.

A direct in-process test is not reliable isolation. Signal dispositions are process-global, while `PID1_TEST_LOCK` only serializes tests that acquire that mutex. In addition, the Rust test process can already start with `SIGPIPE` ignored, so merely checking the post-setup disposition would not prove that `setup()` installed it.

## Decision

Add one process-isolated regression test for the real `SignalContext::setup()` path. The isolated child test process must first establish a known `SIG_DFL` baseline for all three signals, invoke `setup()`, and then verify each disposition independently.

- A parent test launches the current test binary with an exact ignored child-test name.
- The child test resets `SIGTTIN`, `SIGTTOU`, and `SIGPIPE` to `SIG_DFL`, calls the production `SignalContext::setup()`, and verifies that every signal is now `SIG_IGN`.
- The parent verifies that the intended child test actually ran and succeeded.
- Process isolation contains both the process-global dispositions and the signal mask. Normal or abnormal child exit returns all state to the kernel, so no restoration guard is needed in the shared parent test process.

This makes the regression deterministic even when the Rust runtime's initial `SIGPIPE` disposition is already ignored, and removing any one loop entry yields a signal-specific failure.

## Alternatives considered

- **Rely on existing tests:** rejected because they bypass `setup()` and cannot detect regressions in the ignore loop.
- **Call `setup()` in the shared test process and restore state with a guard:** rejected because a crate-local mutex does not stop unrelated test threads from observing process-global changes, and panic-safe restoration adds avoidable unsafe cleanup complexity.
- **Exercise the complete `guest-init` binary:** rejected because its entry point mounts guest filesystems and cgroups and requires the Firecracker PID 1 environment; that scope is unnecessary for this signal contract.

## Risks and mitigations

- The subprocess invocation could silently select zero tests if its exact name drifts. The parent must assert that the named child test appears in the harness output, not only that the process exits successfully.
- The test is Linux-specific, like `guest-init` itself, and uses the existing `nix` signal API already used by production code.
- No production behavior, API, persisted data, protocol, or deployment compatibility changes are required.

## Acceptance criteria

- A process-isolated test invokes the real `SignalContext::setup()` path.
- Before setup, the isolated child explicitly sets `SIGTTIN`, `SIGTTOU`, and `SIGPIPE` to `SIG_DFL`.
- After setup, the child verifies each of those signals is `SIG_IGN`, with a failure that identifies the missing signal.
- The parent test proves that the intended child test ran and completed successfully.
- The shared parent test process never changes its signal dispositions or mask, including when the child assertion fails.
- Removing any one of the three entries from the production ignore loop produces a focused test failure.

seven332:
## Summary

- run the real `SignalContext::setup()` path in an exact process-isolated child test
- reset `SIGTTIN`, `SIGTTOU`, and `SIGPIPE` to `SIG_DFL` before setup so the Rust runtime's ignored `SIGPIPE` cannot hide a missing production entry
- verify that the intended child test ran and that every disposition becomes `SIG_IGN`
- serialize child ownership with the existing PID 1 test lock so concurrent `waitpid(-1)` tests cannot reap the subprocess

## Scope

This is a test-only change in `crates/guest-init/src/pid1.rs`. It does not change production behavior, APIs, persisted data, protocols, dependencies, deployment ordering, or compatibility handling.

## Validation

- `cargo test --manifest-path crates/Cargo.toml -p guest-init pid1::tests::setup_ignores_inherited_signals_in_isolated_process`
- temporarily removed each of `SIGTTIN`, `SIGTTOU`, and `SIGPIPE` from the production ignore loop; each mutation failed with the corresponding signal-specific assertion
- complete `guest-init` suite passed 20 consecutive times under the default parallel test harness
- `cargo test --manifest-path crates/Cargo.toml -p guest-init`
- `cargo clippy --manifest-path crates/Cargo.toml -p guest-init --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc --manifest-path crates/Cargo.toml -p guest-init --all-features --no-deps`

Closes #25589

### DIFF
--- a/crates/guest-init/src/pid1.rs
+++ b/crates/guest-init/src/pid1.rs
@@ -423,6 +423,7 @@ mod tests {
 
     #[test]
     fn setup_ignores_inherited_signals_in_isolated_process() {
+        // Other PID 1 tests call waitpid(-1), so serialize ownership of this child.
         let _guard = PID1_TEST_LOCK.lock().unwrap();
         let output = Command::new(std::env::current_exe().unwrap())
             .arg("--exact")
--- a/crates/guest-init/src/pid1.rs
+++ b/crates/guest-init/src/pid1.rs
@@ -224,10 +224,12 @@ pub fn supervise(
 mod tests {
     use super::*;
     use std::os::fd::{AsRawFd, FromRawFd, OwnedFd};
+    use std::process::Command;
     use std::sync::Mutex;
     use std::thread;
 
     static PID1_TEST_LOCK: Mutex<()> = Mutex::new(());
+    const SIGNAL_SETUP_CHILD_TEST: &str = "pid1::tests::setup_ignores_inherited_signals_child";
 
     struct ChildGuard {
         pid: libc::pid_t,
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #25599

**Reviewed head:** `022322a8d76aa78656295065883e75b2e885c617`

### Summary

The regression test exercises the real `SignalContext::setup()` path in an exact child test process, establishes a deterministic `SIG_DFL` baseline for all three signals, prevents a zero-tests false pass, and correctly serializes subprocess ownership against the existing `waitpid(-1)` tests. The change remains entirely under `#[cfg(test)]` with no production or compatibility impact.

### Findings

No findings.

### Testing

Focused parent and child tests, the complete `guest-init` suite, 20 repeated default-parallel runs, a 32-thread run, all three signal-removal mutation checks, Rust formatting, Clippy with warnings denied, rustdoc with warnings denied, and repository hooks passed. A Firecracker boot E2E was not run; it is not required for this test-only setup contract.

### Verdict

LGTM

PR #25603, opened 2026-08-07
refactor(core): share the built-in video model catalog

bingjiez666:
## Why

The catalog of built-in text-to-video models (which aspect ratios, durations, and resolutions each one accepts) lives inside `zero-video-io-generate.service.ts`, so it is reachable only from `apps/api`. The chat composer is about to expose these parameters on the inline template chip, and it needs the same data to build its dropdowns and to constrain values when the user switches model.

This is step 1 of 3. It moves the catalog and changes nothing else; the contract and UI work land in follow-ups.

## What

- New `packages/core/src/video-model-catalog.ts` holding the aspect ratio / duration / resolution lists, the per-model config table, the model id and alias maps, and the request defaults.
- `zero-video-io-generate.service.ts` imports them instead of declaring them.
- The three defaults the service applied as inline literals are now named: `DEFAULT_VIDEO_MODEL`, `DEFAULT_VIDEO_ASPECT_RATIO` (`16:9`), `DEFAULT_VIDEO_DURATION` (`8s`).

Provider request shapes, pricing categories, and the Seedance output dimension tables stay in the API service — they are generation-time concerns with no consumer on the web side.

## Behavior

None. Every moved line is byte-identical to what it replaced, verified by diffing the removed block against the new module. The only substituted lines are the three default literals, which resolve to the same values.

## Verification

- `pnpm --dir packages/core check-types` — clean
- `oxlint` on both changed files — clean
- `pnpm knip --workspace packages/core` — clean
- `apps/api` type-check and the existing `zero-video-io-generate.test.ts` are left to the pipeline; the test file is untouched and passing it unchanged is the acceptance bar for this PR.

### DIFF
--- a/turbo/apps/api/src/signals/services/zero-video-io-generate.service.ts
+++ b/turbo/apps/api/src/signals/services/zero-video-io-generate.service.ts
@@ -3,6 +3,25 @@ import { Buffer } from "node:buffer";
 import { command, computed, type Computed } from "ccstate";
 import { usageEvent } from "@vm0/db/schema/usage-event";
 import { usagePricing } from "@vm0/db/schema/usage-pricing";
+import {
+  DEFAULT_VIDEO_ASPECT_RATIO,
+  DEFAULT_VIDEO_DURATION,
+  DEFAULT_VIDEO_MODEL,
+  SEEDANCE_RESOLUTIONS,
+  VIDEO_ASPECT_RATIOS,
+  VIDEO_DURATIONS,
+  VIDEO_MODEL_ALIASES,
+  VIDEO_MODEL_CONFIGS,
+  VIDEO_MODELS,
+  VIDEO_RESOLUTIONS,
+  type SeedanceResolution,
+  type VideoAspectRatio,
+  type VideoDuration,
+  type VideoModel,
+  type VideoModelConfig,
+  type VideoProvider,
+  type VideoResolution,
+} from "@vm0/core/video-model-catalog";
 import { and, eq, inArray } from "drizzle-orm";
 
 import { logger } from "../../lib/log";
@@ -17,7 +36,6 @@ import {
   type BuiltInGenerationUsageIdempotency,
 } from "./built-in-generation-usage-idempotency";
 
-const VIDEO_IO_MODEL = "dreamina-seedance-2-0-fast-260128";
 const BYTEPLUS_VIDEO_TASKS_URL =
   "https://ark.ap-southeast.bytepluses.com/api/v3/contents/generations/tasks";
 const MINIMAX_VIDEO_GENERATION_URL =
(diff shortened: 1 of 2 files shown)


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 0180f851b12548ec71187fedd3cf973c5ab5bf17
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Moves the built-in text-to-video model catalog (aspect ratios, durations, resolutions, the per-model config table, model id/alias maps) out of `turbo/apps/api/src/signals/services/zero-video-io-generate.service.ts` into a new shared `turbo/packages/core/src/video-model-catalog.ts`, so the chat composer can consume the same data. Provider request shapes, pricing categories, and the Seedance output dimension tables stay in the API service. The three previously inline default literals are now named exports (`DEFAULT_VIDEO_MODEL`, `DEFAULT_VIDEO_ASPECT_RATIO`, `DEFAULT_VIDEO_DURATION`).

I verified the no-behavior-change claim mechanically. Of the 274 removed lines, 270 appear byte-identical in the new catalog module (modulo the added `export` keyword); the only 4 that do not are exactly the substituted defaults:

- `const VIDEO_IO_MODEL = "dreamina-seedance-2-0-fast-260128"` → `DEFAULT_VIDEO_MODEL` (same value)
- `readString(body, "aspectRatio", "16:9")` → `DEFAULT_VIDEO_ASPECT_RATIO` (`"16:9"`)
- `readString(body, "duration", "8s")` → `DEFAULT_VIDEO_DURATION` (`"8s"`)
- `readString(body, "model", VIDEO_IO_MODEL)` → `DEFAULT_VIDEO_MODEL`

Symmetrically, the only genuinely new content in the diff is the module doc comment, the import block, the three `DEFAULT_*` declarations, and a Prettier re-wrap of one `readString(...)` call.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic — no wrong assumptions or unhandled paths introduced | Pass | Removed/added line sets diffed against `main`; every moved declaration is byte-identical and the 3 substituted defaults resolve to the same literals. |
| `REVIEW.md` | Security — no secrets, no missing auth on changed surfaces | Pass | The moved catalog holds only public model ids, aliases, and capability flags. Provider credentials, request signing, and the auth/credit gates all stay in `zero-video-io-generate.service.ts`, untouched. |
| `REVIEW.md` | Type safety — no `any`, no overly broad types | Pass | `grep` for `any` on `turbo/packages/core/src/video-model-catalog.ts` returns none; all literal tables keep `as const satisfies ...`, and the three defaults use `satisfies` so literal types are preserved. |
| `REVIEW.md` | Style & maintainability — no functions >~100 lines, no duplicated logic | Pass | Net effect is de-duplication ahead of a second consumer. Repo-wide tree scan finds exactly one video model catalog (`turbo/packages/core/src/video-model-catalog.ts`); no parallel copy was left behind or created. |
| `REVIEW.md` | Interface changes — new public interfaces documented, breaking changes flagged | Pass | New workspace-internal subpath `@vm0/core/video-model-catalog`; `turbo/packages/core/package.json` exposes `"./*" → "./src/*.ts"` so it resolves, and `turbo/apps/api/package.json` already declares `"@vm0/core": "workspace:*"`. No published/external interface changes. |
| `REVIEW.md` | Testing coverage — `refactor:` commits must keep existing tests covering the code | Pass | `turbo/apps/api/src/signals/routes/__tests__/zero-video-io-generate.test.ts` is untouched and still exercises this path: the case at L751 omits `model` and asserts the resolved `dreamina-seedance-2-0-fast-260128`, and several prompt-only requests (L566/637/668/693/1637/1701) exercise the aspect-ratio and duration defaults, since a wrong default would 400 in `parseVideoAspectRatio` before reaching BytePlus. Keeping it green unchanged is the correct acceptance bar. |
| `REVIEW.md` | DB/JSONB review gate — persisted contract changes need migration/normalizer/rationale | N/A | No change under `turbo/packages/db/src/schema/**` or `turbo/packages/api-contracts/src/contracts/**`; no `jsonb(...).$type<...>()` touched. |
| `REVIEW.md` | Feature regression guard / feature-switch gating for new user-facing features | N/A | `refactor:` with no user-visible surface. Step 1 of 3 by design; the contract and composer UI land in follow-ups and are where a switch would apply. |
| `docs/bad-smell.md` | Zero tolerance for `any` | Pass | No `any` in either changed file. |
| `docs/bad-smell.md` | Zero tolerance for lint/type suppression comments | Pass | No `@ts-ignore`, `@ts-expect-error`, `@ts-nocheck`, `eslint-disable`, `oxlint-disable`, or `prettier-ignore` in the diff. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | The new module and the updated import block use only static top-level imports. |
| `docs/bad-smell.md` | No hardcoded URLs / environment-specific config | Pass | The moved constants are product-domain model metadata, not environment config. The service's URLs (`BYTEPLUS_VIDEO_TASKS_URL`, `MINIMAX_VIDEO_GENERATION_URL`) were deliberately left in place and unchanged. |
| `docs/bad-smell.md` | Fail fast — no fallback/recovery logic hiding misconfiguration | Pass | `DEFAULT_VIDEO_*` are request-parameter defaults for omitted user input, not config fallbacks, and the pre-existing `badRequest(...)` rejections for unsupported values are preserved verbatim. |
| `docs/testing.md` | Mock at the system boundary only — no `vi.mock()` on internal relative paths | N/A | No test file is added or modified by this PR. |
| `docs/testing.md` | Integration-first — no new unit tests for internal functions | Pass | No new unit test was added for the extracted catalog; coverage stays at the existing route-level integration suite, which is the right call for a pure move of static data. |
| `docs/testing.md` | Deployment compatibility tests when a deployable boundary changes | N/A | `@vm0/core` is a compile-time workspace package. No API contract, runner protocol, queue payload, or persisted shape changes, so no old/new version combination exists to test. |
| `.claude/skills/ccstate/SKILL.md` | `turbo/apps/api` changes must follow ccstate command/computed practices | Pass | The diff touches only module-scope constants, type declarations, and three literal substitutions inside `parseVideoAspectRatio`/`parseVideoOptions`. No `command`/`computed`/`state` definition, no store read/write, and no `detach()` usage is added or altered. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Notes (non-blocking)

- `VideoModelFamily` and `FalRequestFormat` remain unexported in the new module while the exported `VideoModelConfig` union references them. This is valid TypeScript and works here because the package `exports` map points consumers at the `.ts` source, but the composer will not be able to name those types when it narrows a config by `provider`/`requestFormat`. Worth exporting in the follow-up that actually consumes them, rather than churning this move-only PR.
- No dangling references: `STANDARD_VIDEO_ASPECT_RATIOS`, `VEO_VIDEO_DURATIONS`, `KLING_VIDEO_DURATIONS`, `SEEDANCE_2_DURATIONS`, `SEEDANCE_1_5_DURATIONS`, `MINIMAX_H3_DURATIONS`, `SEEDANCE_FAST_RESOLUTIONS`, `VEO_VIDEO_RESOLUTIONS`, `KLING_4K_VIDEO_RESOLUTIONS`, `MINIMAX_H3_RESOLUTIONS`, `VIDEO_IO_MODEL`, and the four config interfaces all have zero remaining occurrences in the API service, and every one of the 17 imported symbols is used at least once in the service body. No unused-import or unresolved-symbol risk.

### Testing

- **Coverage: Adequate.** `refactor:` commit — the existing `zero-video-io-generate.test.ts` is unmodified and still covers model/aspect-ratio/duration resolution, per-model validation rejections, and the BytePlus/Fal/MiniMax request shapes built from the moved config table. Passing that suite unchanged is exactly the right acceptance signal for a move-only change; adding assertions here would test the data table rather than behavior.
- **Conventions: Compliant.** No test files changed, so none of the mock-boundary, MSW, real-database, fake-timer, or mock-cleanup rules are engaged.
- **Pipeline:** `apps/api` type-check and the video generation suite are left to CI, as the PR description states.

PR #25608, opened 2026-08-07
perf(rust): eliminate redundant exec-control payload copies

### LINKED ISSUE #25590: perf(rust): eliminate redundant exec-control payload copies
## Locations

- `crates/runner/src/executor/active_input.rs`
- `crates/sandbox/src/types.rs`
- `crates/sandbox-fc/src/sandbox.rs`
- `crates/vsock-host/src/exec_operation/handle.rs`
- `crates/vsock-host/src/exec_operation/frame.rs`
- `crates/vsock-proto/src/payloads/exec_operation/control.rs`

## Problem and evidence

Large active-input control requests cross an avoidable owned-buffer copy chain before the host writes the vsock frame.

1. The runner serializes `ActiveInputPayload` into an owned `Vec<u8>`.
2. `GuestProcessControlHandle::control` accepts a borrowed slice but clones it into another `Vec<u8>` for its owned callback future.
3. `sandbox-fc` owns that clone and passes it by slice to `vsock-host` across the awaited request.
4. `encode_exec_control` allocates an exec-control payload `Vec` and copies the body into it.
5. Outer `vsock_proto::encode` allocates the final frame `Vec` and copies the complete exec-control payload again.

Active input admits serialized payloads up to 1,048,576 bytes. A maximum-size request can therefore involve roughly four similarly sized owned buffers during the nested request and three full-body copies after serialization. Concurrent controls amplify transient memory, allocator pressure, and CPU copy cost.

The original issue counted only the runner buffer, protocol payload, and final frame. Repository tracing confirmed the additional `payload.to_vec()` at the sandbox control-handle boundary, so a host-only direct-frame change would leave another same-size avoidable clone on the critical path.

## Decision

Treat the root problem as the end-to-end exec-control ownership and framing path. Deliver one coherent PR that transfers already-owned active-input bytes through the sandbox boundary and encodes them directly into the final contiguous vsock frame.

This remains independently reviewable because all affected crates ship in the same runner artifact, the wire bytes remain unchanged, and there is no backend, queue, database, or persisted-state rollout.

- Provide an ownership-preserving control entry point through `sandbox` and `vsock-host` so active input can transfer its existing message id and payload without a boundary clone.
- Keep the borrowed control API for callers with small static payloads; it can remain the convenience path while the owned path avoids cloning already-owned data.
- Add an exec-control full-frame encoder backed by the same validation and payload appender as the existing payload encoder.
- Build the final frame outside the socket writer lock, release the transferred source payload after frame construction, and reuse the existing pending-response and guarded write lifecycle.
- Keep control independent from the file frame-builder gate so large file construction cannot delay latency-sensitive control delivery.

The resulting owned active-input path performs one required body copy into the final contiguous frame after serialization, rather than cloning at the sandbox boundary, copying into an intermediate protocol payload, and copying that payload into the frame.

## Alternatives considered

- **Host-only direct-frame encoding:** saves one allocation and copy but leaves the sandbox boundary clone and keeps its buffer alive through the socket write. This is a partial treatment of the verified chain.
- **Vectored socket writes:** could avoid copying the body into a contiguous frame, but partial vectored writes would materially complicate frame atomicity, cancellation poisoning, and retry-free stream safety for limited additional value.
- **Use the existing file frame-builder gate:** could bound concurrent frame construction, but it would couple control latency to large file writes and contradict the current separation between control/lifecycle traffic and file-frame construction.
- **Build the frame under the writer lock:** bounds concurrent final-frame memory but reverses prior work that prevents encoding from blocking unrelated lifecycle and control writes.

## Risks and mitigations

- **Borrowed/owned API divergence:** route both through one owned host implementation and cover both behavior-level entry points.
- **Validation or wire drift:** share validation and field appending, and compare direct output and errors with the composed encoder at empty, representative, maximum, and invalid boundaries.
- **Lifecycle regressions:** retain the pending guard, normal-operation transition, write observer, timeout start, cancellation guard, and connection-poisoning code; run the existing supervised exec-control integration suite.
- **New lock contention:** do not add a frame-builder lock to the control path and keep frame construction outside the writer lock.

## Acceptance criteria

- Active input transfers its already-owned serialized message id and payload through `GuestProcessControlHandle` and `sandbox-fc` to `vsock-host` without cloning the payload at either control-handle boundary.
- Borrowed control callers retain their current API behavior and acknowledgement/error semantics.
- `vsock-proto` exposes an exec-control full-frame encoder that writes into a caller-owned buffer and produces the same bytes as wrapping `encode_exec_control` in `MSG_EXEC_CONTROL` framing.
- Payload-only and full-frame encoders share validation and payload appending and preserve field order, limits, and diagnostics.
- The owned host path constructs only the final protocol frame, releases the transferred payload before the socket write wait, and does not allocate an intermediate exec-control payload `Vec`.
- Pending-response registration, normal-operation tracking, write-observer timing, timeout behavior, cancellation safety, connection poisoning, and connection reuse remain unchanged.
- Empty, representative, and maximum-size payloads have byte-equivalent frames; oversized and invalid inputs retain matching errors and do not send frames.
- Behavior-level sandbox/runner coverage proves owned forwarding, and host integration coverage proves the existing control request/acknowledgement lifecycle.
- There is no wire-format, API-service, queue, schema, migration, persisted-state, or rollout change.

## Related precedent

- #19006 / #19023 added direct exec-output frame encoding.
- #19490 / #19524 added direct write-file frame encoding.

seven332:
## Summary

- transfer already-owned active-input message IDs and payloads through the sandbox and vsock-host boundaries without cloning the large payload
- encode exec-control requests directly into the final contiguous vsock frame through shared validation and payload-appending logic
- preserve the borrowed control APIs and the existing pending-response, timeout, cancellation, observer, and connection-poisoning behavior

## Scope

- no wire-format, API-service, queue, database, schema, migration, persisted-state, or rollout changes
- small borrowed control callers continue to use the existing convenience APIs

## Validation

- `cargo clippy -p vsock-proto -p vsock-host -p sandbox -p sandbox-fc -p runner --all-targets --all-features`
- `RUSTDOCFLAGS='-D warnings' cargo doc -p vsock-proto -p vsock-host -p sandbox -p sandbox-fc -p runner --all-features --no-deps`
- `cargo test -p vsock-proto --quiet`
- `cargo test -p vsock-host --quiet`
- `cargo test -p sandbox --quiet`
- `cargo test -p sandbox-fc process_control --quiet`
- `cargo test -p runner executor::tests::agent_run_tests::active_input --quiet`

Closes #25590

### DIFF
--- a/crates/runner/src/executor/active_input.rs
+++ b/crates/runner/src/executor/active_input.rs
@@ -128,7 +128,7 @@ async fn forward_text(
     // runner-internal and has no active-input identity or deduplication semantics.
     let correlation_id = format!("active-input:{run_id}:{delivery_sequence}");
     match control
-        .control(&correlation_id, &bytes, ACTIVE_INPUT_CONTROL_TIMEOUT)
+        .control_owned(correlation_id, bytes, ACTIVE_INPUT_CONTROL_TIMEOUT)
         .await
     {
         Ok(_) => debug!(run_id = %run_id, "forwarded active input"),
--- a/crates/sandbox-fc/src/sandbox.rs
+++ b/crates/sandbox-fc/src/sandbox.rs
@@ -875,9 +875,9 @@ impl FirecrackerSandbox {
         )?;
 
         let outcome = tokio::select! {
-            result = control.control(
-                &message_id,
-                &payload,
+            result = control.control_owned(
+                message_id,
+                payload,
                 timeout,
             ) => ControlOutcome::Returned(result),
             () = wait_for_backend_crash(state_rx) => ControlOutcome::BackendCrashed,
--- a/crates/sandbox/src/types.rs
+++ b/crates/sandbox/src/types.rs
@@ -491,7 +491,22 @@ impl GuestProcessControlHandle {
         payload: &[u8],
         timeout: Duration,
     ) -> std::io::Result<ProcessControlAck> {
-        (self.control)(message_id.to_owned(), payload.to_vec(), timeout).await
+        self.control_owned(message_id.to_owned(), payload.to_vec(), timeout)
+            .await
+    }
+
+    /// Send an owned opaque control payload to the live guest process.
+    ///
+    /// This has the same behavior as [`Self::control`] but transfers ownership
+    /// of `message_id` and `payload` so callers that already own large request
+    /// data do not need to clone it for the provider callback.
+    pub async fn control_owned(
+        &self,
+        message_id: String,
+        payload: Vec<u8>,
+        timeout: Duration,
+    ) -> std::io::Result<ProcessControlAck> {
+        (self.control)(message_id, payload, timeout).await
     }
 }
 
--- a/crates/vsock-host/src/exec_operation/frame.rs
+++ b/crates/vsock-host/src/exec_operation/frame.rs
@@ -255,6 +255,20 @@ pub(in crate::exec_operation) async fn write_frame_with_pre_write(
     Ok(())
 }
 
+pub(in crate::exec_operation) async fn write_encoded_frame_with_pre_write(
+    shared: &Arc<Shared>,
+    data: &[u8],
+    diagnostic: Option<ExecOperationFrameDiagnostic>,
+    pre_write: impl FnOnce() -> io::Result<()>,
+) -> io::Result<()> {
+    let _decision = write_encoded_frame_with_pre_write_decision(shared, data, diagnostic, || {
+        pre_write()?;
+        Ok(FrameWriteDecision::Write)
+    })
+    .await?;
+    Ok(())
+}
+
 async fn write_frame_with_pre_write_decision(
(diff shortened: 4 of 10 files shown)


seven332:
## Code Review: PR #25608

**Reviewed head:** `97c770331df0191149a68023db01fd21ef0a163d`

### Summary

The owned active-input path now removes the verified cross-crate payload clones and intermediate protocol payload while preserving wire bytes, borrowed callers, and the guarded exec-control lifecycle.

### Findings

No findings.

### Testing

Protocol equivalence and invalid boundaries are covered through public exports, the supervised host fixture exercises the owned request/ack path, and existing sandbox/runner behavior tests pass. Rust formatting, Clippy, Rustdoc, affected crate suites, and focused sandbox-fc/runner tests passed locally; the full repository matrix remains delegated to CI.

### Verdict

LGTM

PR #25613, opened 2026-08-07
feat(api): carry user-chosen video parameters into the generation prompt

bingjiez666:
## Why

Video templates currently carry only a style preset id, so the model, framing, length, and resolution are whatever the generation service defaults to. The composer is about to expose them on the inline template chip, which needs somewhere to put the choice and a way to make it stick.

Options reach the provider through the agent — `buildGenerationTemplatePrompt` renders the selection into prompt text and the agent assembles the `zero generate video` command itself. So "the user picked 9:16" has to become an explicit, non-negotiable instruction rather than a field that flows straight through.

## What

- `packages/api-contracts/src/contracts/video-models.ts` — the model id, aspect ratio, duration, and resolution value domains, mirroring the existing `model-price-tiers.ts` layout. They live here because `@vm0/core` depends on this package, not the other way round, so the contract cannot import the catalog.
- `chat-threads.ts` — `selection.videoOptions` with enums derived from those domains, so the contract cannot drift from the generation service.
- `selection.avatarOptions` — the four talking-avatar fields, nested.
- `readAvatarTemplateOptions` in `@vm0/core/avatar-template` — one reader, nested wins, flat is the fallback.
- `generation-template-prompt.ts` — emits the chosen parameters as a "keep every one of them" block and appends the real CLI flags to the final generation step.

## Two decisions worth a second look

**Options are stored sparse.** Only what the user touched is written. An omitted field keeps meaning "let the service pick", so when the default model changes later, messages authored today follow it instead of being pinned to a stale one. The chip resolves defaults at render time. This also means a user who never opens the options panel produces a payload byte-identical to today's.

**Avatar fields are dual-written.** New code writes `avatarOptions` and the flat fields. Non-strict zod objects strip unknown keys, so a bundle that predates `avatarOptions` sees only the flat ones — and because composer drafts are persisted server side (`chat_threads.draft_user_message`, `zero_agent_drafts.draft_user_message`), that skew is reachable without any deploy in flight. Dropping `videoOptions` costs a user their parameter choice; dropping `voiceId` would silently generate the avatar speaking in a different voice. Removing the flat write is a separate follow-up once every deployed reader understands the nested shape.

## Deployment compatibility

No DDL. All four affected columns (`chat_events.generation_template`, `chat_events.user_message`, and the two draft columns) are `jsonb` with no index, generated column, or check constraint reaching into their contents; `drizzle generate` produces no migration. Old rows simply lack the new keys and resolve to defaults on read.

## Mismatched options

If the stored options do not fit the chosen model — a stale client, or a hand-written payload — the offending value is dropped rather than substituted, and the generation service applies its own per-model default. Substituting a global default could hand the service a value that model rejects outright (Veo takes only 4s/6s/8s; Kling only 4k).

## Verification

- `apps/api` `generation-template-prompt.test.ts` — 15 passed, including 4 new cases: parameters pinned, unsupported values dropped, flat avatar fields still readable, nested preferred over flat.
- `check-types` clean for `@vm0/api-contracts`, `@vm0/core`, `@vm0/db`.
- `oxlint` clean on every changed file; `knip` clean for both packages.
- `apps/api` and `apps/platform` full type-checks OOM in this 4 GB sandbox and ride on the pipeline. The one non-obvious typing — `includes` being invariant against the catalog's `as const` literal tuples — was verified in isolation, including confirming the widening annotation is load-bearing.


bingjiez666:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 770e514df3860f41e14b7477c0fd6c64d11af427

### Summary

Adds `selection.videoOptions` (model / aspect ratio / duration / resolution / audio) and a nested `selection.avatarOptions` to the video generation-template contract, moves the video value domains into `@vm0/api-contracts/contracts/video-models` so the catalog in `@vm0/core` can derive from them, and renders the user's chosen parameters into the generation prompt as a "keep every one of them" block plus verbatim CLI flags on the final generation step. Avatar fields are dual-written (nested + flat) with a documented removal condition tracked in #25620.

The contract layering, the sparse-storage decision, and the deployment-compatibility reasoning are sound and well documented. Two correctness problems in `selectedVideoParameters` and one testing-altitude gap need to be addressed.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic: no wrong assumptions or unhandled edge cases | Finding (P0) | `generation-template-prompt.ts:99` gates `--no-audio` on `config.supportsGenerateAudio`, which is `true` for every model in the catalog, so the flag is emitted for MiniMax H3, which the generation service rejects with 400. |
| `REVIEW.md` | Correctness and logic: no null/undefined paths | Finding (P1) | `generation-template-prompt.ts:63` indexes `VIDEO_MODEL_CONFIGS` with a persisted `options.model` that is never re-validated on the read path. |
| `REVIEW.md` | Security: no injection of untrusted values | Pass | Only enum-constrained values reach the prompt; `readAvatarTemplateOptions` still keeps `titleSnapshot`/`previewUrl` out of `buildAvatarGenerationTemplatePrompt`, so the existing no-injection assertions in `chat-events.bdd.test.ts` remain valid. |
| `REVIEW.md` | Type safety: no `any`, no overly broad types | Pass | No `any`; the one widening annotation (`const config: VideoModelConfig`) is explained in a comment and is load-bearing for `includes` invariance. |
| `REVIEW.md` | DB/JSONB review gate: old rows must not violate the new shape, or a clear no-migration rationale | Pass | All new keys are optional additions to unindexed `jsonb` columns (`chat_events.generation_template`, `user_message`, both draft columns); old rows simply lack them. Rationale stated in the PR body; `chat-event.ts` mirrors the additive shape. |
| `REVIEW.md` | Feature regression guard: changed existing behavior needs explicit regression coverage | Pass | The avatar selection path changed from flat reads to `readAvatarTemplateOptions`; `chat-user-messages.test.tsx:132` keeps a flat-only stored message and asserts it still resolves, and `chat-events.bdd.test.ts:6021` already exercises the flat request shape end-to-end. |
| `REVIEW.md` | New user-facing feature gated behind a `FeatureSwitchKey` | N/A | Contract and prompt plumbing only; no client writes `videoOptions` yet, and the composer chip that will is a separate change. |
| `docs/bad-smell.md` | No lint/type suppressions | Pass | No `@ts-expect-error`, `oxlint-disable`, or `prettier-ignore` in the diff. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | All new imports are static and top-of-file. |
| `docs/bad-smell.md` | Fail fast; no silent fallback that hides a problem | Finding (P0) | Dropping an unhonourable value is the PR's stated contract, but the audio gate silently keeps a value the service will reject instead of dropping it — the failure surfaces later as a 400 from the generation call. |
| `docs/bad-smell.md` | Interface changes documented; breaking changes highlighted | Pass | New public types (`VideoGenerationOptions`, `AvatarGenerationOptions`, `AvatarTemplateOptions`, `readAvatarTemplateOptions`) are exported with doc comments; the deprecated flat fields carry `@deprecated` plus a removal condition. |
| `docs/deployment-compatibility.md` | Prefer additive changes; readers tolerate missing new fields | Pass | Every new field is optional; `readAvatarTemplateOptions` reads nested-then-flat, and `selectedVideoParameters` returns `[]` for `undefined` options, which reproduces today's prompt byte-for-byte. |
| `docs/deployment-compatibility.md` | Do not persist data the previous deployed reader cannot parse | Pass | Non-strict zod objects strip unknown keys, so an older bundle sees the mirrored flat avatar fields; `videoOptions` is inert for old readers. |
| `docs/deployment-compatibility.md` | Compatibility code must be temporary and explicit, with a deletion condition or tracking issue | Pass | `avatar-template-selection.ts:26`, `chat-threads.ts:488`, and `avatar-template.ts:604` each state the rollout reason and point at #25620; the write-side and read-side removal conditions are stated separately. |
| `docs/testing.md` | `feat:` commits must have integration tests | Finding (P1) | Platform behavior is covered by page-level tests, but the new `videoOptions` prompt behavior is only exercised by direct calls to `buildGenerationTemplatePrompt`. |
| `docs/testing.md` | Mock at the boundary only; no `vi.mock()` on relative paths | Pass | No new mocks; the added API cases are pure function calls and the platform cases use the existing `context.mocks` API mocking. |
| `docs/testing.md` | No fake timers; assert behavior, not mock invocation | Pass | No `vi.useFakeTimers()`; every added assertion is on returned prompt text or the submitted template payload. |
| `docs/testing.md` | Deployment-compatibility tests cover the old persisted shape | Pass | `generation-template-prompt.test.ts:247` and `chat-user-messages.test.tsx:132` both read a flat-only selection; `:273` pins nested-over-flat precedence. |
| `docs/testing/api-testing.md` | `apps/api` behavior is covered through `setupApp()` and the route contract, not by importing internals | Finding (P1) | The four new cases import `buildGenerationTemplatePrompt` from `../../../lib/generation-template-prompt` instead of asserting on `appendSystemPrompt` through the chat-events route. |
| `docs/testing/testing-external-behavior.md` | Construct cases and assert results through the external interface | Finding (P1) | `chat-events.bdd.test.ts:6002-6019` already drives this exact prompt path through the API and asserts on `appendSystemPrompt`; the new parameters were not added there. |
| `docs/testing/testing-external-behavior.md` | Platform cases go through the page, not through internal components or stores | Pass | Both changed platform tests still drive the composer through user interactions and assert on the submitted template. |
| `.claude/skills/ccstate/SKILL.md` | Signals layer uses `accept`, no manual HTTP error handling, no `detach` misuse | N/A | `avatar-template-selection.ts` contains only pure mapping helpers; no `zeroClient$` call, `state`, `computed`, or `command` was added or changed. |
| `docs/effect.md` | Choose the owner and trigger before the API; no render-phase side effects | N/A | No React component, effect, ref, or route setup changed; the platform diff is a pure function plus test updates. |
| `docs/cache.md` | No unbounded-lifetime caches, duplicate mutable sources of truth, or asymmetric resource lifecycles | N/A | No cache, store, subscription, or resource is introduced; `readAvatarTemplateOptions` allocates a plain object per call. |

### Findings

- `turbo/apps/api/src/lib/generation-template-prompt.ts:99` — `--no-audio` is emitted for a model that rejects it, producing a guaranteed failed generation.

  The gate is `options.generateAudio !== undefined && config.supportsGenerateAudio`. In `turbo/packages/core/src/video-model-catalog.ts`, `supportsGenerateAudio` is `true` for all six models, so this predicate never drops anything — it is effectively dead. The real constraint lives in `turbo/apps/api/src/signals/services/zero-video-io-generate.service.ts:762`:

  ```ts
  if (modelConfig.provider === "minimax" && !requested) {
    return badRequest("MiniMax H3 always generates native audio");
  }
  ```

  So `{ model: "MiniMax-H3", generateAudio: false }` renders `- Audio: off` into the "Keep every one of them; do not ... drop a flag" block and appends `--no-audio` to the final generation step, which the service answers with 400. This is the exact case the PR's own rule ("values the chosen model cannot honour are dropped rather than rewritten") exists to prevent, and the contract accepts that combination from any API caller the moment this merges — it does not need the composer chip to be reachable.

  Suggested fix: gate on the constraint that is actually enforced (for example, skip the audio parameter when `config.provider === "minimax"`), and either give the catalog a field that is genuinely `false` for MiniMax H3 or drop `supportsGenerateAudio` from this gate so it does not read as protection it does not provide. A case covering MiniMax H3 + `generateAudio: false` would pin the behavior.

- `turbo/apps/api/src/lib/generation-template-prompt.ts:63` — unguarded catalog lookup on a value that is not re-validated when it comes from persisted state.

  `VIDEO_MODEL_CONFIGS[options.model ?? DEFAULT_VIDEO_MODEL]` is safe for a freshly validated request, but the same function is reached from persisted data: `internal-chat-run-callback.service.ts:3061` → `resolveThreadGenerationTemplatePrompt` → `buildGenerationTemplatePrompt`, with the template coming from `projectUserMessage(document)`, which consumes the `user_message` jsonb without a zod re-parse. Once a dated model id (`seedance-1-5-pro-251215`, `dreamina-seedance-2-0-260128`) is retired from `VIDEO_MODEL_IDS`, an already-persisted message carrying it yields `config === undefined` and `config.alias` throws, failing the queued-run path rather than degrading. The codebase already treats a model string from outside as untrusted — `zero-video-io-generate.service.ts:620` uses `normalizeVideoModel`, which returns `null` for an unknown id. Dropping the model parameter when it is not in the catalog would match both that convention and this function's own drop-don't-substitute rule.

- `turbo/apps/api/src/signals/routes/__tests__/generation-template-prompt.test.ts:187` — the new prompt behavior has no coverage through the external API surface.

  `docs/testing/api-testing.md` and `docs/testing/testing-external-behavior.md` require `apps/api` behavior to be constructed and verified through the endpoint; the four added cases call `buildGenerationTemplatePrompt` directly. This is the file's pre-existing convention, so it is not a new violation, but the route-level home for this exact behavior already exists: `chat-events.bdd.test.ts:6002-6019` sends a video `generationTemplate` through `sendChatRun` and asserts on `appendSystemPrompt`, and `:6021-6050` does the same for the avatar flat fields. Adding one case there with `videoOptions` set would prove the parameter block and the flag suffix survive request validation, persistence, and projection — which is precisely where the P0 above becomes observable. Without it, the only evidence for a `feat:` change is a direct call to an internal function.

#### Nits (non-blocking)

- `turbo/packages/core/src/avatar-template.ts:596` — `AvatarTemplateOptions` restates the shape of `AvatarGenerationOptions` with a looser `voiceId: string`. Since `@vm0/core` now depends on `@vm0/api-contracts` (this PR relies on that direction in `video-model-catalog.ts`), reusing the contract type would keep the two from drifting if `avatarVideoAspectRatioSchema` ever gains a value.
- `turbo/apps/api/src/lib/generation-template-prompt.ts:100` — when `generateAudio` is `true`, `- Audio: on` is listed under "do not drop a flag" while no corresponding flag appears in the command line. The comment explains why, but the instruction and the command are momentarily inconsistent from the agent's point of view.

### Testing

- Coverage: Insufficient. The avatar half is well covered — flat-only read (`generation-template-prompt.test.ts:247`), nested-over-flat precedence (`:273`), platform dual-write (`chat-composer-template-picker.test.tsx:950`), and reopening a pre-`avatarOptions` stored message (`chat-user-messages.test.tsx:132`). The `videoOptions` half is missing (a) any case for the MiniMax H3 + `generateAudio: false` combination the service rejects, and (b) any route-level case proving the parameters reach `appendSystemPrompt`.
- Conventions: Compliant except for test altitude in `apps/api` (see P1 above). No relative-path `vi.mock()`, no `vi.stubGlobal("fetch")`, no fake timers, no database mocking, and assertions are on behavior rather than mock invocation. The platform tests continue to drive the page through user interactions.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 0588d3369694df13b67137dd938324030ab3d397
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Adds `selection.videoOptions` (model / aspect ratio / duration / resolution / audio) and a nested `selection.avatarOptions` to the video generation-template contract, moves the video value domains into `@vm0/api-contracts/contracts/video-models` so `@vm0/core`'s catalog derives from them, and renders the user's chosen parameters into the generation prompt as a "keep every one of them" block plus verbatim CLI flags on the final generation step. Avatar fields are dual-written (nested + flat) with a documented removal condition tracked in #25620.

Re-review at `0588d33`. All three findings from the review at `770e514` are resolved, and the delta introduces no new issues.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic: no wrong assumptions or unhandled edge cases | Pass | `honoursGenerateAudio` (`generation-template-prompt.ts:295`) now gates on the constraint the service enforces at `zero-video-io-generate.service.ts:762`, so a silent MiniMax choice is dropped instead of pinned. All four combinations behave correctly: minimax+off dropped, minimax+on kept as a no-op label, non-minimax+off emits `--no-audio`, non-minimax+on kept. |
| `REVIEW.md` | Correctness and logic: no null/undefined paths | Pass | `generation-template-prompt.ts:325` widens the lookup to `VideoModelConfig \| undefined` and returns `[]` for a model id no longer in the catalog, matching the function's drop-don't-substitute rule rather than validating against a different model's constraints. |
| `REVIEW.md` | Correctness: existing prompt assembly unaffected | Pass | `stripGenerationTemplateContext` slices from the first blank line at index > 1, which `templateFraming` places at index 5; `parameterLines` are inserted at index ≥ 10, so the inline-template path is untouched. The multi-template test at `:16` still pins one shared context. |
| `REVIEW.md` | Security: no injection of untrusted values | Pass | Only enum-constrained values reach the prompt; `readAvatarTemplateOptions` still keeps `titleSnapshot`/`previewUrl` out of `buildAvatarGenerationTemplatePrompt`, so the no-injection assertions at `chat-events.bdd.test.ts:6084` remain valid. |
| `REVIEW.md` | Type safety: no `any`, no overly broad types | Pass | No `any`. Both widening annotations (`VideoModelConfig \| undefined`, and the original `includes`-invariance annotation) carry comments explaining why they are load-bearing; `pnpm check-types` passes across all 13 tasks. |
| `REVIEW.md` | DB/JSONB review gate: old rows must not violate the new shape, or a clear no-migration rationale | Pass | All new keys are optional additions to unindexed `jsonb` columns (`chat_events.generation_template`, `user_message`, and both draft columns); old rows simply lack them. Rationale stated in the PR body; `chat-event.ts` mirrors the additive shape. |
| `REVIEW.md` | Feature regression guard: changed existing behavior needs explicit regression coverage | Pass | The avatar read path changed from flat fields to `readAvatarTemplateOptions`; `chat-user-messages.test.tsx:132` keeps a flat-only stored message, `generation-template-prompt.test.ts:272` pins flat/nested equivalence, and `chat-events.bdd.test.ts:6057` still drives the flat request shape end-to-end. |
| `REVIEW.md` | New user-facing feature gated behind a `FeatureSwitchKey` | N/A | Contract and prompt plumbing only; no client writes `videoOptions` yet, and the composer chip that will is a separate change. |
| `docs/bad-smell.md` | No lint/type suppressions | Pass | No `@ts-expect-error`, `oxlint-disable`, or `prettier-ignore` in the diff; full lint run reports 0 errors across 13 tasks. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | All new imports are static and top-of-file. |
| `docs/bad-smell.md` | Fail fast; no silent fallback that hides a problem | Pass | Both drop paths are the contract, not a fallback: an unhonourable value is omitted so the service applies its own per-model default, and each is documented at the point of the decision. Nothing is substituted or silently rewritten. |
| `docs/bad-smell.md` | Interface changes documented; breaking changes highlighted | Pass | New public types (`VideoGenerationOptions`, `AvatarGenerationOptions`, `AvatarTemplateOptions`, `readAvatarTemplateOptions`) are exported with doc comments; the deprecated flat fields carry `@deprecated` plus a removal condition. |
| `docs/deployment-compatibility.md` | Prefer additive changes; readers tolerate missing new fields | Pass | Every new field is optional; `readAvatarTemplateOptions` reads nested-then-flat and `selectedVideoParameters` returns `[]` for absent options, reproducing today's prompt byte-for-byte. |
| `docs/deployment-compatibility.md` | Do not persist data the previous deployed reader cannot parse | Pass | Non-strict zod objects strip unknown keys, so an older bundle still sees the mirrored flat avatar fields; `videoOptions` is inert for old readers. |
| `docs/deployment-compatibility.md` | Readers must tolerate persisted data written by another version | Pass | Strengthened at `0588d33`: the projection path (`internal-chat-run-callback.service.ts:3061` → `projectUserMessage`) consumes `user_message` jsonb without a zod re-parse, and the catalog guard now keeps a retired model id from throwing. |
| `docs/deployment-compatibility.md` | Compatibility code must be temporary and explicit, with a deletion condition or tracking issue | Pass | `avatar-template-selection.ts:26`, `chat-threads.ts:488`, and `avatar-template.ts:604` each state the rollout reason and point at #25620, with separate write-side and read-side removal conditions. |
| `docs/testing.md` | `feat:` commits must have integration tests | Pass | `chat-events.bdd.test.ts:6021` now drives `videoOptions` through `sendChatRun` and asserts on `appendSystemPrompt`; the platform half keeps its page-level coverage. |
| `docs/testing.md` | Mock at the boundary only; no `vi.mock()` on relative paths | Pass | No new mocks; the BDD case uses the existing API-driven helpers and `context.mocks`. |
| `docs/testing.md` | No fake timers; assert behavior, not mock invocation | Pass | No `vi.useFakeTimers()`; every added assertion is on returned prompt text or the submitted template payload. |
| `docs/testing.md` | Deployment-compatibility tests cover the old persisted shape | Pass | `generation-template-prompt.test.ts:272` and `chat-user-messages.test.tsx:132` both read a flat-only selection; `:298` pins nested-over-flat precedence. |
| `docs/testing/api-testing.md` | `apps/api` behavior is covered through `setupApp()` and the route contract, not by importing internals | Pass | The new route case constructs state via `sendChatRun` and verifies via `api.readRun`, with no DB writes, schema imports, or service calls. |
| `docs/testing/testing-external-behavior.md` | Construct cases and assert results through the external interface | Pass | The parameter block, the dropped Veo duration, and the `--model veo3.1-fast --aspect-ratio 9:16 --resolution 1080p` suffix are all asserted on `appendSystemPrompt`, which is what an external caller observes. |
| `docs/testing/testing-external-behavior.md` | Platform cases go through the page, not through internal components or stores | Pass | Both changed platform tests still drive the composer through user interactions and assert on the submitted template. |
| `.claude/skills/ccstate/SKILL.md` | Signals layer uses `accept`, no manual HTTP error handling, no `detach` misuse | N/A | `avatar-template-selection.ts` contains only pure mapping helpers; no `zeroClient$` call, `state`, `computed`, or `command` was added or changed. |
| `docs/effect.md` | Choose the owner and trigger before the API; no render-phase side effects | N/A | No React component, effect, ref, or route setup changed; the platform diff is a pure function plus test updates. |
| `docs/cache.md` | No unbounded-lifetime caches, duplicate mutable sources of truth, or asymmetric resource lifecycles | N/A | No cache, store, subscription, or resource is introduced; `readAvatarTemplateOptions` allocates a plain object per call. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Resolved since `770e514`

- **P0 — `--no-audio` pinned for a model that rejects it.** The gate read `config.supportsGenerateAudio`, which is `true` for all six catalog models and therefore never dropped anything, so `{ model: "MiniMax-H3", generateAudio: false }` rendered `- Audio: off` into the "do not drop a flag" block and appended `--no-audio`, which `parseVideoGenerateAudio` answers with `400 MiniMax H3 always generates native audio`. Fixed by `honoursGenerateAudio`, which encodes both real constraints — a model that cannot generate audio at all, and MiniMax always returning native audio. Pinned by `generation-template-prompt.test.ts:228`.
- **P1 — unguarded catalog lookup on unvalidated persisted state.** Fixed by widening the lookup and returning no parameters when the config is missing, with a comment recording why the value is untrusted on the projection path.
- **P1 — no coverage through the external API surface.** Fixed by the new `chat-events.bdd.test.ts` case; the no-options case now also asserts the parameter block is absent, so the default path stays pinned.

#### Nits (non-blocking, unchanged)

- `turbo/packages/core/src/avatar-template.ts:596` — `AvatarTemplateOptions` restates the shape of `AvatarGenerationOptions` with a looser `voiceId: string`. Since `@vm0/core` now depends on `@vm0/api-contracts` (this PR relies on that direction in `video-model-catalog.ts`), reusing the contract type would keep the two from drifting if `avatarVideoAspectRatioSchema` ever gains a value.
- `turbo/apps/api/src/lib/generation-template-prompt.ts:369` — when `generateAudio` is `true`, `- Audio: on` is listed under "do not drop a flag" while no corresponding flag appears in the command line. The comment explains why (audio is the CLI default), and the instruction and the command are momentarily inconsistent from the agent's point of view.

### Testing

- Coverage: Adequate. Route level — `videoOptions` reaching `appendSystemPrompt` with an unsupported duration dropped, plus the no-options and flat-avatar paths. Prompt level — full parameter pinning, per-model value rejection, the MiniMax audio case, flat/nested avatar equivalence, and nested-over-flat precedence. Platform level — dual-write on submit and reopening a pre-`avatarOptions` stored message.
- Conventions: Compliant. No relative-path `vi.mock()`, no `vi.stubGlobal("fetch")`, no fake timers, no database mocking; assertions are on behavior rather than mock invocation, and both new cases construct state through the surface their app exposes.
- Local verification: `pnpm format` clean; `pnpm turbo run lint --concurrency=1` 13/13 tasks with 0 errors; `pnpm check-types` 13/13; `generation-template-prompt.test.ts` 16/16 pass; `chat-events.bdd.test.ts -t "renders generation template guidance into the run system prompt"` passes.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 7cea884e33a36ac1305b3d63599da504edad81ec
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Adds `selection.videoOptions` (model / aspect ratio / duration / resolution / audio) and `selection.avatarOptions` to the video generation-template contract, moves the video value domains into `@vm0/api-contracts/contracts/video-models` so `@vm0/core`'s catalog can re-export them without a dependency cycle, and renders the user's explicit choices into the appended system prompt as a "keep every one of them" block plus verbatim CLI flags. Avatar options are dual-written (nested + flat) for the rollout window, with `readAvatarTemplateOptions` as the shared reader.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic; no wrong assumptions about downstream APIs | Pass | `honoursGenerateAudio` in `generation-template-prompt.ts:55` exactly mirrors `parseVideoGenerateAudio` in `zero-video-io-generate.service.ts:753` (minimax + `!requested` → 400; `!supportsGenerateAudio` → forced false). Emitted flags `--model/--aspect-ratio/--duration/--resolution/--no-audio` all match `video-generate.ts:396-415`, and `--model` correctly emits `config.alias`, not the raw id. |
| `REVIEW.md` | Security — no injection through new prompt content | Pass | Every value reaching `parameterLines` is whitelist-checked against the catalog (`config.aspectRatios/durations/resolutions.includes`), model renders as `config.alias`, and `generateAudio` renders as `on`/`off`. No caller-controlled free text enters the prompt; `chat-events.bdd.test.ts:6083` keeps asserting untrusted avatar name/preview are not injected. |
| `REVIEW.md` | Type safety — no `any`, no over-broad types | Pass | No `any` in the diff. The one widening annotation (`const config: VideoModelConfig \| undefined`, `generation-template-prompt.ts:84`) is load-bearing for `includes` invariance and is documented in place. |
| `REVIEW.md` | No lint/type suppressions | Pass | No `@ts-ignore`, `@ts-expect-error`, `oxlint-disable`, or `prettier-ignore` anywhere in the diff. |
| `REVIEW.md` | DB/JSONB review gate — old rows must not violate the new shape | Pass | `chat-event.ts:22` adds only optional keys to the `jsonb` selection; no DDL, no index/generated column/check constraint reaches into these payloads. Old rows simply lack the keys and read back as "service picks". |
| `REVIEW.md` | Feature regression guard — changed existing flow needs regression coverage | Pass | The avatar selection write path changed (`avatar-template-selection.ts:24`). Covered from both directions: `chat-user-messages.test.tsx:132` reopens a stored flat-only message, and `generation-template-prompt.test.ts:348` asserts flat-only input produces a result strictly equal to the nested one. |
| `REVIEW.md` | New user-facing feature behind a `FeatureSwitchKey` | N/A | No UI writes `videoOptions` in this PR; the composer chip lands separately. A user who never opens the options panel produces a byte-identical payload, so there is no user-visible behavior to gate. |
| `docs/bad-smell.md` | No fallback/recovery logic hiding misconfiguration | Pass | The `config === undefined` guard at `generation-template-prompt.ts:86` handles a retired model id arriving from unparsed persisted `jsonb`, not a missing configuration; the documented result (drop the pins, let the service default) is the same sparse behavior as an untouched panel. |
| `docs/bad-smell.md` | No dynamic `import()`; no hardcoded URLs/config | Pass | All new imports are static; the only literals added are enum value domains in `video-models.ts` and CLI flag names. |
| `docs/bad-smell.md` | Interface changes documented, breaking changes highlighted | Pass | Each retained flat field carries a `@deprecated` marker plus its deletion condition in `chat-threads.ts:588`, tracked in #25620. |
| `docs/deployment-compatibility.md` | Prefer additive changes; readers tolerant of missing new persisted fields | Pass | Both new keys are optional; `readAvatarTemplateOptions` (`avatar-template.ts:56`) falls back to flat fields, so a new reader handles pre-split rows. |
| `docs/deployment-compatibility.md` | Do not persist data the previous deployed reader cannot process | Pass | Non-strict `z.object` strips `avatarOptions` in an old bundle, which is why the flat mirror is written. Reachable without a deploy in flight because drafts persist server-side in `chat_threads.draft_user_message`. `videoOptions` has no mirror, but no client writes it yet. |
| `docs/deployment-compatibility.md` | Compatibility code must be temporary and explicit, with a deletion condition | Pass | `avatar-template-selection.ts:424` and `chat-threads.ts:590` both name the condition (web-client floor raised past the shipping app version) and link #25620; `avatar-template.ts:709` correctly notes the read fallback must outlive the write mirror until a `jsonb` backfill. |
| `docs/deployment-compatibility.md` | Test the previous request/persisted shape | Pass | `chat-events.bdd.test.ts:6059` still sends the old flat-only avatar selection through the live API route and asserts the resolved prompt. |
| `docs/testing.md` | Integration-first; no unit tests for internal functions | Pass | The new parameter behavior is pinned end-to-end through the real route in `chat-events.bdd.test.ts:6014`; `generation-template-prompt.test.ts` extends a pre-existing table-style suite for the per-model matrix rather than replacing that coverage. |
| `docs/testing.md` | Mock at boundary only — no relative-path `vi.mock` | Pass | The diff adds no mocks of any kind. |
| `docs/testing.md` | No direct `fetch` stubbing; MSW at the HTTP boundary | Pass | No `vi.stubGlobal("fetch", …)` added; platform tests use the existing `context.mocks.api` harness (`chat-user-messages.test.tsx:516`). |
| `docs/testing.md` | No fake timers | Pass | No `vi.useFakeTimers()` / `advanceTimersByTime` in the diff. |
| `docs/testing.md` | Test behavior, not mocks | Pass | Every new assertion inspects resolved prompt text or the submitted template payload, not call counts. |
| `docs/testing.md` | `feat:` commits must have integration tests | Pass | Four new cases in `generation-template-prompt.test.ts` plus a new end-to-end block in `chat-events.bdd.test.ts`. |
| `.claude/skills/ccstate/SKILL.md` | Signals must not use `detach`/DOM-callback patterns; no manual cache over `computed` | Pass | `avatar-template-selection.ts` changes are pure synchronous mapping functions — no `state`, `computed`, command, or subscription is touched. |
| `docs/event-sourcing.md` | Persistent events authoritative; no failure-path rollback required for optimistic events | N/A | The PR changes only the payload shape carried inside an existing event; no event identity, reconciliation, or projection logic changes. |

### Findings

#### High Priority (P1)

- `turbo/apps/platform/src/views/zero-page/avatar-template-picker.tsx:1043` — `selectedVoiceId` still reads the flat field directly (`value.selection.voiceId`) instead of going through `readAvatarTemplateOptions`. It is the only remaining reader of a flat avatar field outside the new helper, so the PR's "one reader, nested wins, flat is the fallback" invariant does not actually hold yet. Nothing breaks today because `toAvatarGenerationTemplate` dual-writes, but when #25620 removes the flat write this picker will silently stop highlighting the user's selected voice — the exact silent-voice-loss failure mode the dual write exists to prevent. Worth one line here so the follow-up is a pure deletion. Source: `docs/deployment-compatibility.md` ("Clean up: remove compatibility logic only after the old deployed version is no longer active" — the cleanup must be safe to perform mechanically).

### Testing

- Coverage: Adequate. `feat:` commit with integration coverage at both levels — the real chat-run route (`chat-events.bdd.test.ts`) pins parameter rendering, unsupported-value dropping, and the untouched-panel case (`not.toContain("Parameters the user set explicitly")`); `generation-template-prompt.test.ts` covers the per-model matrix (Veo duration/aspect rejection, MiniMax silent-request rejection, flat fallback, nested-over-flat precedence). Both platform write paths have regression tests for the old stored shape.
- Conventions: Compliant. No mocks added, no fake timers, no `fetch` stubbing, no relative-path `vi.mock`, and all assertions are on observable output rather than mock invocations.
- Gap worth noting: no test asserts the `avatar-template-picker.tsx:1043` preselection against a nested-only selection, which is why the P1 above can regress unnoticed.

PR #25621, opened 2026-08-07
fix(runner): preserve drain override across repeated rollback

### LINKED ISSUE #25592: bug(rust): preserve Restart=no during repeated drain rollback
## Location

- `crates/runner/src/cmd/service/drain_resume.rs:263-295`
- `crates/runner/src/cmd/service/drain_resume.rs:550-608`
- `crates/runner/src/cmd/service/drain_override.rs:36-77`

Verified at commit `7ae186e0c9b04199101dd052459062e4dc916d09`.

## Problem

`drain_with_ops` treats every successful `write_restart_override` call as a newly written override by setting `restart_override_written = should_signal`. The writer always overwrites the fixed `50-vm0-drain.conf` path and does not report whether that file already existed.

If a later daemon reload or restart-policy verification fails, `rollback_drain_transition` therefore unconditionally removes the override whenever the service was active-like. This is correct for a first drain attempt that created the file, but not for a repeated drain after an earlier drain already committed `Restart=no`.

## Failure scenario

1. `runner service drain` succeeds while the runner still has work. The unit is disabled and the runtime drop-in applies `Restart=no`.
2. The operator retries `runner service drain` while the same service is still active or deactivating.
3. The retry overwrites the existing drop-in at the same path.
4. The retry then encounters a daemon-reload failure, restart-policy mismatch, or restart-policy query failure.
5. Rollback removes the drop-in and reloads systemd, even though the drop-in predated this transition. The prior enablement is already disabled, so rollback does not otherwise restore the first drain's state.

The still-running draining service can now regain its unit-file `Restart=on-failure` policy. A crash before graceful completion can auto-restart the runner during a drain, violating the state established by the successful first command. Disabling a unit only controls boot activation; it does not suppress restart policy for the currently loaded service.

## Acceptance criteria

- Drain rollback distinguishes an override created by the current attempt from an override that existed before it.
- A failed first-time drain still removes the newly created override and restores the prior enablement state.
- A failed repeated drain preserves the pre-existing drain override and leaves effective `Restart=no` loaded after rollback.
- Tests cover repeated-drain rollback for both daemon-reload failure and restart-policy verification failure, alongside the existing first-attempt rollback behavior.

seven332:
## Summary

- report whether the canonical drain restart override was created or replaced by the current drain attempt
- remove only newly created overrides during rollback while preserving and reloading pre-existing `Restart=no` protection
- verify the effective restart policy after repeated-drain rollback and surface restoration failures instead of treating loaded path presence as sufficient
- cover first-attempt compatibility, repeated reload and policy failures, policy-query failures, and persistent effective-policy mismatch

## Scope

- no API, protocol, database, persisted-state, or deployment-ordering changes
- successful drain, resume, stop, install, uninstall, and cleanup behavior remains unchanged

## Validation

- `RUSTDOCFLAGS='-D warnings' cargo doc -p runner --no-deps --all-features`
- `cargo test -p runner --all-targets --all-features` (3052 passed, 20 existing ignored; guest inventory passed)
- regression mutation: restoring unconditional removal made all three ownership-sensitive repeated-drain rollback tests present at mutation time fail; restoring the fix made them pass
- `git diff --check`

Closes #25592

### DIFF
--- a/crates/runner/src/cmd/service/drain_override.rs
+++ b/crates/runner/src/cmd/service/drain_override.rs
@@ -12,7 +12,15 @@ const RUNTIME_SYSTEMD_SYSTEM_DIR: &str = "/run/systemd/system";
 const DRAIN_DROP_IN_FILE_NAME: &str = "50-vm0-drain.conf";
 const DRAIN_DROP_IN_CONTENT: &str = "[Service]\nRestart=no\n";
 
-pub(super) fn write_drain_restart_override(unit: &RunnerServiceUnit) -> RunnerResult<()> {
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(super) enum DrainRestartOverrideWrite {
+    Created,
+    Replaced,
+}
+
+pub(super) fn write_drain_restart_override(
+    unit: &RunnerServiceUnit,
+) -> RunnerResult<DrainRestartOverrideWrite> {
     write_drain_restart_override_at(Path::new(RUNTIME_SYSTEMD_SYSTEM_DIR), unit)
 }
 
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #25621

**Reviewed head:** `985635282c7f3e6cb31a2f423c9e0237dbbd570a`

### Summary

The ownership-aware write result cleanly distinguishes first-attempt state from a pre-existing drain override. Rollback now preserves earlier `Restart=no` protection and verifies the effective policy, while existing first-attempt cleanup remains intact.

### Findings

No findings.

### Testing

The regression matrix covers created and replaced writes, reload failure, policy mismatch recovery, policy-query failure, persistent rollback mismatch, and existing first-attempt behavior. Local runner tests passed with 3052 passed and 20 existing ignored, plus the guest inventory test; formatting, Clippy, rustdoc, mutation validation, and diff checks also passed. Deterministic real-systemd failure injection was not run; the existing operation boundary supplies failure injection for this orchestration.

### Verdict

LGTM

PR #25622, opened 2026-08-07
refactor(api): stop writing chat event trigger source

### LINKED ISSUE #25594: refactor(api): stop writing chat event trigger source
## Scope

**Step 4 of #25260.** Stop writing `chat_events.trigger_source`. The column stays; new rows simply leave it null. Step 5 drops it.

#25552 removed the last reader — `grep -rn "chatEvents.triggerSource" turbo/apps/api/src` returns zero non-test hits today. This step removes the writers.

## What is *not* going away

`trigger_source` survives as a concept. `zero_runs.trigger_source` stays `NOT NULL` and keeps serving billing, log filtering, callback guards and cleanup. What dies is the **second copy** on the chat event.

After #25552 the run's value is already computed at run-creation time from the new key, so three helpers must survive:

| helper | still used by | keep |
|---|---|---|
| `queuedUserMessageTriggerSource(contextType)` | `internal-chat-run-callback.service.ts:2370` (prior-run query), `:3187` (run-creation args), `:3317` (log fields) | ✅ |
| `manualTriggerSource({ kind })` | `workflow-chat-event-queue.service.ts:259` (run creation), `zero-workflow-automation.service.ts:3130` ("Run now") | ✅ |
| `isWebChatTriggerSource` | `agent-run-create.service.ts:358`, reading `body.triggerSource` from the run-creation request | ✅ |

Only their *chat-event write* call sites go.

## The entry point

`NewChatEvent` in `zero-chat-event.service.ts` declares `triggerSource` on three variants:

```
:230  InputPromptEvent      readonly triggerSource?: TriggerSource;
:241  InputAutomationEvent  readonly triggerSource: TriggerSource;   ← required
:268  the input.rejected variant   readonly triggerSource?: TriggerSource;
```

Delete all three and let `check-types` enumerate the call sites. `:241` being **required** means removing it is the one that ripples — every automation admission currently has to supply it.

## Writers, measured on `feeb1cb`

**13 non-test sites.** Confirm by shape, not line number.

```
ingress        canonical-slack-ingress-processor:358    canonical-feishu-ingress-processor:359
               zero-teams-dispatch:1619   zero-telegram-post:1805   zero-agentphone:1497
Morning Brief  morning-brief-run:304
workflow queue workflow-chat-event-queue:156 (admission)   :359 (rejection replacement)
web / agent    zero-chat-events.command:1472
claim          zero-chat-queued-event:468   :532   :599
seeder         scripts/dev-bench-seed:685
```

**28 more in tests and benchmarks**, 15 of them in `routes/test-cron-monitor-chat-event-queue-state.ts` alone.

Three of the thirteen are already shadows — #25552 made them re-derive the value from the new key, so deleting them loses nothing:

```ts
zero-chat-queued-event.service.ts:468    triggerSource: queuedUserMessageTriggerSource(contextType)
zero-chat-queued-event.service.ts:532    triggerSource: manualTriggerSource({ kind: head.automationKind })
workflow-chat-event-queue.service.ts:359 triggerSource: manualTriggerSource({ kind: payload.automationKind })
```

## Traps

**1. The monitor fixtures are not a find-and-delete.** `routes/test-cron-monitor-chat-event-queue-state.ts` builds twelve rows keyed on `trigger_source` (`"web"`, `"test"`, `"agent"`, each integration, `null`, …). #25552 already changed the monitor to read `context_type` only, so those fixtures currently set a column nothing reads — the tests still pass but no longer test what they name. Convert them to `context_type` fixtures rather than dropping the field, or the monitor loses its coverage silently.

Note `"test"` is a `TriggerSource` value with no `context_type` counterpart. Decide what that fixture becomes and say so in the PR.

**2. No migration, and no backfill.** `chat_events` is append-only. Historical values stay exactly as they are; new rows are simply null. Adding `DROP COLUMN` here is step 5 and would fail the previous API version during the promotion window — migrations run before `promote-api-production`, measured repeatedly in this epic (0836: migration 12:41:29→12:43:27Z, promotion 12:43:33Z).

**3. Dead imports.** `TriggerSource` is likely unused in `chat-event.ts`'s write union afterwards, and several writers import it only for this field. Let `knip` and lint drive the cleanup rather than leaving them.

**4. Writers live outside the route layer.** #25284 was followed by #25310 because a benchmark constructed `eventRows` directly. `dev-bench-seed.ts:685` is already in the list above; grep once more for direct `input.*` construction across tests, fixtures, benchmarks and seeds before opening the PR.

**5. This closes the cross-check window.** Until now `trigger_source` and `context_type` coexisted, and every production acceptance in this epic used one to verify the other. After this step `context_type` is the only record of a queued event's source. Nothing to do about it in code — it just means the PR's own verification carries more weight.

## Explicitly out of scope

- Dropping the column — step 5, separate PR, one deploy apart.
- `zero_runs.trigger_source`, the logs/activity contract, `chat_*_context` tables.
- Widening `chat_events_input_context_type_check` to cover `input.budget`. It is the one event type on the pending path that the CHECK does not protect; both of its writers happen to produce a non-null discriminator by construction (`replacementContext` inheritance, and `agentRunContext` → `agent_run`). Fold it into step 5's migration instead of adding one here.

## Acceptance criteria

- No chat-event write path sets `triggerSource`; `NewChatEvent` no longer declares it on any variant.
- `queuedUserMessageTriggerSource`, `manualTriggerSource` and `isWebChatTriggerSource` still exist and still serve run creation — `zero_runs.trigger_source` is unchanged for every source, asserted per source in tests.
- The monitor's fixture matrix is expressed in `context_type` and still covers the same cases.
- No migration, no schema change, no contract change, no client change.
- After deploy, newly created `input.prompt` / `input.automation` / `input.goal` rows have `trigger_source IS NULL` and a non-null `context_type`, and their runs still carry the right `zero_runs.trigger_source`.

lancy:
## Summary

- remove `triggerSource` from every `NewChatEvent` input variant and stop setting `chat_events.trigger_source` in production, test, fixture, and replacement writes
- keep the physical column unchanged so new rows use its null default while historical append-only rows remain untouched
- express the queue monitor fixture matrix solely through the 11 unique `context_type` values and keep all nine orphan-source alert groups covered
- update Slack integration coverage to assert pending web and Slack chat events leave `trigger_source` null while their runs still persist `web` and `slack`

## Monitor fixture decision

The legacy `"test"` fixture is removed instead of being remapped. `test` has no `context_type` counterpart, and remapping it to `web` would duplicate the existing web fixture without adding monitor coverage. The remaining web and `agent_run` fixtures still cover the two non-alerting source shapes, and the other nine context types still cover every grouped alert case.

## Preserved behavior

- `queuedUserMessageTriggerSource` remains in use for prior-run lookup, run creation, and admission logging
- `manualTriggerSource` remains in use for workflow run creation and Run now
- `isWebChatTriggerSource` remains in use for run-creation request handling
- `zero_runs.trigger_source`, log/activity contracts, and `chat_*_context` are unchanged
- no migration, schema, client, or contract files are changed

## Verification

- `pnpm -F api check-types`
- scoped type-aware Oxlint and ESLint for all changed files
- `pnpm knip --workspace api`
- Prettier check for all changed files
- direct `insertChatEvent`, `replaceChatEvent`, and `db.insert(chatEvents)` writer scans: no remaining `triggerSource` assignments
- database-backed focused tests are delegated to PR CI because this worktree has no `DATABASE_URL` or running PostgreSQL

Closes #25594

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/cron-monitor-chat-event-queue.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/cron-monitor-chat-event-queue.test.ts
@@ -125,7 +125,7 @@ describe("cron monitor chat event queue", () => {
 
     const response = await accept(
       stateClient().monitor({
-        body: { event_ids: fixture.eventIds.slice(0, 3) },
+        body: { event_ids: fixture.eventIds.slice(0, 2) },
       }),
       [200],
     );
@@ -141,7 +141,7 @@ describe("cron monitor chat event queue", () => {
     const fixture = await trackFixture(seedFixture("orphan"));
 
     const response = await accept(
-      stateClient().monitor({ body: { event_ids: fixture.eventIds.slice(3) } }),
+      stateClient().monitor({ body: { event_ids: fixture.eventIds.slice(2) } }),
       [500],
     );
 
--- a/turbo/apps/api/src/signals/routes/__tests__/integrations.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/integrations.bdd.test.ts
@@ -1808,23 +1808,27 @@ describe("INT-01: Slack app deep webhook flows", () => {
         return ingress.eventId;
       }),
     ).toStrictEqual(expect.arrayContaining([eventId, stickyEventId]));
-    // Pending input events retain channel attribution alongside web messages.
-    expect(state.pending_chat_events).toStrictEqual(
-      expect.arrayContaining([
-        expect.objectContaining({
-          eventType: "input.prompt",
-          triggerSource: "web",
-        }),
-        expect.objectContaining({
-          eventType: "input.prompt",
-          triggerSource: "slack",
-        }),
-      ]),
+    const pendingInputEvents = state.pending_chat_events.filter((pending) => {
+      return pending.eventType === "input.prompt";
+    });
+    expect(pendingInputEvents.length).toBeGreaterThanOrEqual(2);
+    expect(
+      pendingInputEvents.every((pending) => {
+        return pending.triggerSource === null;
+      }),
+    ).toBeTruthy();
+    const pendingEventsWithContext = await Promise.all(
+      pendingInputEvents.map(async (pending) => {
+        return {
+          pending,
+          context: await readChatEventContextFixture(pending.id),
+        };
+      }),
     );
-    const queuedSlackEvents = state.pending_chat_events.filter((pending) => {
+    const queuedSlackEvents = pendingEventsWithContext.filter((candidate) => {
       return (
-        pending.chatThreadId === canonicalChatThreadId &&
-        pending.triggerSource === "slack"
+        candidate.pending.chatThreadId === canonicalChatThreadId &&
+        candidate.context?.contextType === "slack"
       );
     });
     expect(queuedSlackEvents).toHaveLength(1);
@@ -1832,9 +1836,7 @@ describe("INT-01: Slack app deep webhook flows", () => {
     if (!queuedSlackEvent) {
       throw new Error("Expected the canonical thread's pending Slack event");
     }
-    await expect(
-      readChatEventContextFixture(queuedSlackEvent.id),
-    ).resolves.toMatchObject({
+    expect(queuedSlackEvent.context).toMatchObject({
       contextType: "slack",
       contextId: expect.any(String),
       slackChannelId: channelId,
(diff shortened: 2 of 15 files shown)


lancy:
## Code Review: PR #25622 (Round 1)

### Summary

The production-side removal is scoped correctly: the ChatEvent input union no longer accepts `triggerSource`, the listed writers stop setting it, run-side trigger-source derivation remains intact, and the monitor matrix now follows `context_type`.

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

1. `integrations.bdd.test.ts:1820` identifies the pending Slack event with `pending.id === stickyEventId`. Slack's external `event_id` is stored by ingress, while `chat_events.id` is an independently generated internal ID. The assertion therefore finds no event.
2. `integrations.bdd.test.ts:2544` repeats the same ID mismatch for the recovered ingress. CI confirms the pending row has an internal UUID.

### Testing Review

The existing integration coverage is sufficient once the two matchers use the persisted chat context and existing ingress/message assertions instead of equating external and internal IDs. The monitor fixture keeps two exempt context types and nine alerting context types.

### Verdict: Changes Requested

Fixing both P1 assertions and then re-reviewing.

---
*Round 1 of automated review-fix loop*

lancy:
## Code Review: PR #25622 (Round 2) — LGTM :tada:

### Summary

- All typed ChatEvent input variants reject `triggerSource`.
- Production, test, fixture, and benchmark ChatEvent writers no longer set the column.
- Slack integration coverage now identifies queued events through persisted `context_type` while asserting new ChatEvent rows have `triggerSource: null`.
- Run-side source derivation and `zero_runs.trigger_source` assertions remain intact.
- The monitor fixture retains two exempt context types and nine alerting context types; the legacy `test` row was removed because it has no distinct `context_type` and duplicated the `web` monitor case.
- No migration, schema, contract, or client changes are present.

### Verification

- API typecheck passed after rebasing onto current `origin/main`.
- Scoped Oxlint, ESLint, Knip, Prettier, and diff checks passed.
- No full local Vitest suite or dev server was run; PR CI is the integration gate.

No critical or high-priority issues remain. This PR is ready for CI and merge.

---
*Completed after 2 rounds of automated review-fix loop*

PR #25625, opened 2026-08-07
feat(api): retain raw usage events for seven days

### LINKED ISSUE #25579: feat: retain raw usage events for seven days before compaction
## Problem and evidence

Processed `usage_event` rows currently become eligible for hourly compaction shortly after their UTC hour closes. Compaction preserves additive billing totals in `usage_event_hourly_rollup` and atomically deletes the raw rows, but the rollup cannot answer event-level audit questions about quantities, idempotency keys, pricing inputs, or allocation details.

We need a bounded forensic window in which exact processed usage events remain available. Seven days is the required retention period.

## Decision

Retain healthy processed usage events for at least seven days before running the existing hourly compaction and source-row deletion flow.

- Eligibility is based on `processed_at`, because only finalized billing events are compactable.
- The cutoff is the start of the current UTC hour minus seven days.
- Only rows with `processed_at < cutoff` are eligible. This provides at least seven full days of retention, plus up to one additional hour from UTC-hour alignment and normal cron scheduling.
- Once eligible, rows use the existing complete-grain expansion, rollup reconciliation, atomic insertion, and source deletion behavior.
- This change is forward-looking. Raw events that were already compacted and deleted cannot be reconstructed.

## Why this approach

Product usage readers already combine processed raw events with hourly rollups, so delaying compaction preserves reported totals without adding a new read path, status, table, or migration. A raw event has one representation during the seven-day window and is replaced by its hourly rollup afterward.

## Alternatives considered

- Persisting only source cardinality does not retain exact event-level audit evidence and no longer meets the requirement.
- Compacting immediately, marking sources with a separate retained status, and deleting them after seven days would preserve query performance but would reintroduce a retired lifecycle, a second cleanup path, and mixed-version state transitions. The added machinery is not justified without evidence that seven days of raw reads are too expensive.
- Copying raw events to a separate archive would add another storage system and audit access path while keeping the same retention burden.
- Retaining raw events indefinitely would provide stronger history but has unbounded storage and privacy cost.

## Behavior that remains unchanged

- Pending events are not compacted.
- Events with `billing_error` remain diagnostic holds and are not compacted.
- Account and user deletion workflows may remove usage data before the retention period; this feature is not a legal-hold mechanism.
- The physical hourly grain, aggregation math, reconciliation, concurrency serialization, cleanup ordering, and public API contracts remain unchanged.

## Risks and mitigations

- Recent usage queries will scan a seven-day raw working set instead of roughly one completed hour. Existing finalized-usage indexes and the raw-plus-rollup relation already support this path; additional indexing should be driven by observed performance rather than added speculatively.
- Storage, index, and write-amplification costs increase for the retained working set. The seven-day bound prevents indefinite growth.
- An already-running older deployment or a rollback to old code can still compact using the previous cutoff. No schema gate can restore deleted source rows, so rollout should be treated as forward-looking.
- UTC-hour alignment intentionally retains boundary rows longer rather than deleting any event before it reaches seven days.

## Acceptance criteria

- Healthy processed events with `processed_at` at or after the seven-day cutoff remain raw and are not represented by a new hourly rollup.
- Healthy processed events with `processed_at` before the seven-day cutoff are compacted and deleted through the existing atomic flow.
- The cutoff is the start of the current UTC hour minus seven days and uses a strict less-than comparison.
- Pending events and events with `billing_error` retain their current hold behavior regardless of age.
- Usage totals remain identical when read from raw events before compaction and from hourly rollups afterward.
- Existing complete-grain expansion, late-data consolidation, reconciliation, concurrency, and cleanup behavior remain intact.
- No schema migration, new lifecycle status, new storage system, or public API change is introduced.
- Focused real-database integration coverage verifies retained and eligible processed events alongside the existing hold cases.
- Schema documentation describes the seven-day raw retention lifecycle.

## Scope

One PR will change the compaction cutoff, update focused integration coverage, and correct the lifecycle documentation. Historical restoration and new audit/reporting APIs are outside this delivery.

seven332:
## Summary

- retain healthy processed usage events for at least seven days before hourly compaction
- keep the existing UTC-hour-aligned, strict cutoff semantics and atomic rollup/source-deletion flow
- cover retained, eligible, pending, and billing-error rows through the real cron route and PostgreSQL
- document the seven-day raw-event lifecycle in the database schema

## Behavior and exclusions

The cutoff is the start of the current UTC database hour minus seven days. Product readers already combine processed raw events with hourly rollups, so reported totals and public APIs are unchanged.

This PR does not add a migration, lifecycle status, feature switch, index, archive, or cleanup job. Account and user deletion continue to override retention, and the policy is forward-looking when old API revisions overlap or are rolled back.

Supersedes the source-cardinality approach in closed PR #25618.

## Validation

- `pnpm format`
- `pnpm -F api lint`
- `pnpm -F api check-types`
- `pnpm -F @vm0/db lint`
- `pnpm -F api exec vitest run src/signals/routes/__tests__/cron-compact-usage-events.test.ts` (11 tests)
- `pnpm knip --workspace api`
- pre-commit: full Turbo Knip and type checking
- `git diff --check`

Closes #25579

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/cron-compact-usage-events.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/cron-compact-usage-events.test.ts
@@ -285,11 +285,17 @@ describe("usage event compaction cron", () => {
     expect(insight.body.grandTotalCredits).toBe(7);
   });
 
-  it("retains unstable hours and explicit diagnostic holds", async () => {
+  it("retains seven days of processed events and explicit diagnostic holds", async () => {
     const heldFixture = await seedFixture();
-    const currentHour = nowDate();
-    currentHour.setUTCMinutes(0, 0, 0);
-    const previousHour = new Date(currentHour.getTime() - 60 * 60 * 1000);
+    const startedHour = nowDate();
+    startedHour.setUTCMinutes(0, 0, 0);
+    const expectedCutoffAtStart = new Date(
+      startedHour.getTime() - 7 * 24 * 60 * 60 * 1000,
+    );
+    const retainedProcessedAt = new Date(
+      startedHour.getTime() - 6 * 24 * 60 * 60 * 1000,
+    );
+    const eligibleProcessedAt = new Date("0400-01-01T00:30:00.000Z");
 
     await store.set(
       insertUsageEvent$,
@@ -315,16 +321,7 @@ describe("usage event compaction cron", () => {
       {
         ...heldFixture,
         status: "processed",
-        processedAt: previousHour,
-      },
-      context.signal,
-    );
-    await store.set(
-      insertUsageEvent$,
-      {
-        ...heldFixture,
-        status: "processed",
-        processedAt: new Date(currentHour.getTime() + 1000),
+        processedAt: retainedProcessedAt,
       },
       context.signal,
     );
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #25625

**Reviewed head:** `758cfd079343616d14a7a06974247eeef15d8011`

### Summary

The PR implements the approved seven-day raw-retention policy at the existing atomic handoff point. The UTC database cutoff, strict boundary, reader compatibility, hold behavior, deletion semantics, and parallel-test isolation are coherent. The change adds no schema or public protocol surface.

### Findings

No P0, P1, or P2 findings.

### Testing

The focused real-Postgres cron suite passes 11/11 and covers the retained/eligible sides together with pending and billing-error holds. Formatting, API and DB lint/type checks, Knip, pre-commit full Turbo type checking, and diff checks pass. Production seven-day row volume and latency remain operational observations; the existing indexes cover the relevant access paths and no speculative index is justified in this PR.

### Verdict

LGTM

PR #25627, opened 2026-08-07
fix(python): classify sse from exact media type

### LINKED ISSUE #25615: bug(python): select SSE usage parsing from the media type only
## Problem

`crates/runner/mitm-addon/src/response_streaming.py:103-105` and `crates/runner/mitm-addon/src/response_streaming.py:234-235` select the SSE usage path with a substring check:

```python
if "text/event-stream" in content_type:
```

HTTP `Content-Type` parameters are not part of the media type. A valid JSON response such as:

```http
Content-Type: application/json; profile="text/event-stream"
```

therefore enters the SSE parser instead of the JSON parser.

## Reproduction

At `74dd054adfb28fa9ae6305f9dac3bea0951c7dcc`, create an observable Codex model-provider flow with that header and a JSON Responses body containing `input_tokens=12` and `output_tokens=7`, then call `configure_response_stream()`, feed the body, and finalize model usage.

Observed:

- `uses_model_json_fallback(flow)` returns `False`.
- `_MODEL_SSE_USAGE_FINISH` is installed while `_MODEL_JSON_USAGE_FINISH` is not.
- Final `MODEL_PROVIDER_USAGE` is `{}`.
- No fallback buffer is retained.

The same false classification also occurs for non-SSE media types such as `application/text/event-stream+json`.

Expected: only a media type equal to `text/event-stream` (case-insensitively, with optional parameters such as `; charset=utf-8`) should select SSE parsing.

## Impact

A model provider JSON response whose parameters or subtype contain the text `text/event-stream` silently loses token usage extraction. Because both parser setup and fallback selection use the same substring check, the response is not recovered through buffered JSON fallback, causing missing usage observations and possible underbilling.

## Acceptance criteria

- Parse the media type from the header value before selecting the usage protocol, for example via the normalized text before the first semicolon.
- Select SSE only when that media type is exactly `text/event-stream`.
- Preserve support for case-insensitive `text/event-stream` with ordinary parameters.
- Add regression coverage for JSON and non-SSE media types that contain `text/event-stream` only outside the media-type equality boundary.

seven332:
## Summary

- classify model-provider SSE responses from the exact normalized media type
- share the classification between parser setup and JSON fallback eligibility
- cover misleading `Content-Type` parameters and subtypes through the real response hook and JSON usage finalization path
- preserve case-insensitive, parameterized `text/event-stream` handling

## Exclusions

- no API, schema, persisted-data, or runner protocol changes
- no general-purpose MIME parser or new flow metadata state

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_model_provider_response_parser_setup.py` (18 passed)
- `uv run --no-sync python -m pytest tests/` (4957 passed)

Closes #25615

### DIFF
--- a/crates/runner/mitm-addon/src/response_streaming.py
+++ b/crates/runner/mitm-addon/src/response_streaming.py
@@ -91,6 +91,12 @@ def uses_openai_responses_usage_protocol(flow: http.HTTPFlow) -> bool:
     return flow_metadata.cli_agent_type(flow.metadata) == "codex"
 
 
+def _response_has_event_stream_media_type(response: http.Response) -> bool:
+    content_type = response.headers.get("content-type", "")
+    media_type = content_type.partition(";")[0].strip(_HTTP_OWS_CHARS).lower()
+    return media_type == "text/event-stream"
+
+
 def uses_model_json_fallback(flow: http.HTTPFlow) -> bool:
     """Return whether terminal model usage may parse a buffered JSON body."""
     response = flow.response
@@ -100,8 +106,7 @@ def uses_model_json_fallback(flow: http.HTTPFlow) -> bool:
         or not usage.is_model_provider_usage_observable(flow)
     ):
         return False
-    content_type = response.headers.get("content-type", "").lower()
-    if "text/event-stream" in content_type:
+    if _response_has_event_stream_media_type(response):
         return False
     return not _is_confirmed_websocket_upgrade_response(flow)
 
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #25627

**Reviewed head:** `a47874267caa07a1c0daa84428d4a2b6f4b6423b`

### Summary

The PR replaces duplicated full-header substring checks with one exact normalized media-type predicate shared by parser setup and JSON fallback. The integration regression drives the real response hook, stream callback, and JSON usage finalizer, while preserving case-insensitive parameterized SSE behavior.

### Findings

No findings.

### Testing

Ruff format/lint, basedpyright, the focused parser-setup tests (18 passed), and the complete mitm-addon suite (4957 passed) were run on this head. Live mitmproxy interception is outside the repository's documented test boundary; CI is checked separately.

### Verdict

LGTM

Issue #25617, opened 2026-08-07
bug(pi): recover standby control from durable run state


seven332:
## Parent objective

- Parent: #25596
- Make the Pi edge-to-Sandbox handoff reliable across realtime delivery gaps, Runner reconnects, and independently deployed API/Runner versions.

## Background

PR #25489 persists the complete unresolved Pi tool-call batch before publishing `pi-handoff` to the Runner group. A claimed standby Sandbox then relies on the Runner receiving that realtime event and forwarding the corresponding process-control message.

The durable transcript and run/job lifecycle state remain available in the API, but the Runner's local standby notification state is memory-only. There is no application-level reconciliation that turns the already-persisted state back into an immediate handoff or release when the realtime event is not observed locally.

#25593 separately fixes cross-run interference inside a live Runner by replacing shared broadcast fan-out with run-keyed delivery. It does not make realtime delivery durable.

## Problem

A Pi standby control can be absent from the local Runner even though the authoritative state transition has already committed:

- the API publication fails after the transcript or terminal state commits;
- the Runner is disconnected or reconnecting when the event is published;
- a new Runner process has no access to the previous process's pending map or subscription state;
- a notification arrives before any local subscription and is not retained until that run becomes locally relevant.

For a lost `pi-handoff`, the standby Sandbox continues waiting even though the transcript already contains an unresolved Sandbox tool-call batch. It normally recovers only after the standby TTL expires: the Guest exits with the TTL release code, the API detects the persisted handoff batch, and the run is requeued for a cold start. This can add up to the full standby TTL, currently five minutes, and discards the prewarmed Sandbox.

For a lost `pi-standby-release`, an already-settled API-only run can retain an otherwise unused standby Sandbox until the same TTL expires.

The realtime message is therefore being used as both a low-latency wakeup and the only immediate delivery mechanism, even though the business state it represents is already durable.

## Required outcome

Treat realtime Pi standby controls as wakeup hints and persisted API state as the source of truth:

- A standby claim or startup must reconcile whether the run is still waiting, already handoff-ready, or already released/terminal.
- A Runner that misses a realtime event must reconcile durable state promptly rather than waiting for the standby TTL.
- Handoff readiness must be derived only after the complete unresolved Sandbox tool-call batch has committed.
- Release must be derived from authoritative terminal/job lifecycle state.
- The state check and local waiter registration must form a race-free handshake: a transition concurrent with setup cannot fall between the durable read and realtime subscription.
[...]

Issue #25596, opened 2026-08-07
refactor(pi): harden edge-to-sandbox handoff after initial delivery


seven332:
## Background

PR #25489 delivered the first end-to-end Pi edge-to-Sandbox handoff. The original tracking issue, #25433, describes that initial delivery. Post-merge review is now identifying correctness, deployment-compatibility, and architecture corrections that should be tracked separately from the completed feature implementation.

## Objective

Track the independently deliverable corrections needed to make the Pi edge loop, Runner standby handoff, Guest Agent bridge, and recovery paths explicit, reliable, and safe across independently deployed API and Runner versions.

This parent is for confirmed follow-up problems, not speculative redesign. Each correction must be represented by a focused sub-issue with its own evidence, acceptance criteria, and validation boundary.

## Delivery requirements

- Keep each correction independently reviewable and releasable.
- Preserve the default-off `piLoop` rollout gate unless a child issue explicitly changes it.
- Preserve the existing non-Pi execution path when the switch is disabled.
- Treat mixed API/Runner versions, Runner draining, and rollback as supported deployment states.
- Prefer explicit protocol and lifecycle semantics over overloading resource or scheduling concepts.
- Add integration-level coverage at the affected API, Runner, or Guest Agent entry points.

## Confirmed corrections

- #25593 — replace shared Pi standby notification fan-out with run-keyed delivery so unrelated broadcast lag cannot synthesize, evict, or misclassify a run's handoff or release signal.
- #25597 — separate Pi standby protocol capability and retained-job lifecycle semantics from the resource profile abstraction. Standby and cold-start use the default VM image and resource shape; `vm0/pi-standby` must not carry unrelated capability and queue-state meanings.
- #25617 — make realtime standby controls wakeup hints and reconcile handoff or release from durable transcript and run/job state when realtime delivery is missed.

## Completion criteria

- Every linked sub-issue is closed with its acceptance criteria satisfied.
- Pi standby control delivery is isolated by run and cannot be resolved by unrelated Runner-group traffic.
- A missed realtime standby control is reconciled from durable state without waiting for the standby TTL.
- The Pi handoff path remains safe during independent API and Runner rollout or rollback.
- The feature-disabled path remains behaviorally unchanged.
- No obsolete compatibility aliases, lifecycle branches, or misleading profile configuration remain after their replacements are deployed.

## Related work

- Initial implementation tracking: #25433
- End-to-end implementation: #25489

PR #25629, opened 2026-08-07
fix(runner): isolate pi standby notifications by run

### LINKED ISSUE #25593: bug(rust): do not synthesize Pi handoff after unrelated broadcast lag
## Parent

- #25596

## Problem

`PiStandbyNotifications` uses one Tokio broadcast channel for every Pi run in a Runner process. A `PiStandbySubscription` therefore receives and advances through notifications for every run ID, then filters unrelated notifications after receipt.

Broadcast lag is receiver-wide, not run-specific. At `crates/runner/src/pi_standby.rs`, a lagged receiver currently returns `PiStandbySignal::Handoff` even when the per-run pending lookup has no notification for that subscription's run:

```rust
Err(broadcast::error::RecvError::Lagged(_)) => {
    return self.take_pending().unwrap_or(PiStandbySignal::Handoff);
}
```

The source is transferred into the job in `crates/runner/src/cmd/start/job_spawn.rs` and forwarded exactly once by `crates/runner/src/executor/pi_standby.rs`. A synthesized signal therefore becomes a real guest process-control message.

### Unrelated traffic synthesizes a handoff

1. A Pi standby subscription exists for run A, with no handoff or release for A.
2. More than `PI_STANDBY_NOTIFICATION_CAPACITY` (256) notifications for other run IDs arrive before A drains its receiver.
3. A's receiver reports `Lagged`, while A has no pending signal.
4. `wait()` fabricates `Handoff` for A.
5. The one-shot forwarder sends `pi-handoff` to A's guest and exits.
6. The Pi agent reads A's transcript and can fail with `Pi transcript has no unresolved assistant tool-call batch`.

Unrelated Runner-group traffic can therefore fail an otherwise unrelated run.

### A real signal can be evicted or misclassified

The same capacity coupling also affects real signals:

1. A subscription exists for run A.
2. A real `Release` or `Handoff` for A arrives before the subscription starts draining.
3. At least 256 notifications for distinct unrelated run IDs arrive afterward.
4. Both the shared broadcast buffer and the bounded global pending map can evict A's signal.
5. The current lag fallback converts the missing signal to `Handoff`, misclassifying `Release`. Merely changing the lag branch to continue waiting would instead lose the real signal and wait indefinitely.

Therefore a one-line change to ignore unrelated lag is insufficient. Active subscriptions must not depend on a shared bounded stream or a globally evictable pending entry.

## Required direction

Replace shared broadcast fan-out for active subscriptions with run-keyed one-shot delivery:

- Maintain an active waiter keyed by `RunId` and retain a bounded pending cache only for notifications that arrive before subscription.
- Make `subscribe(run_id)` atomically consume an existing pending signal or register that run's waiter under the same lock.
- Make `notify(run_id, signal)` atomically deliver directly to the matching active waiter, or cache the signal when no waiter exists.
- Give each subscription a unique registration identity and remove only the matching waiter on drop, preventing leaked waiters and stale-subscription cleanup from removing a newer registration.
- Preserve `Handoff` and `Release` exactly; never infer either signal from unrelated transport state.
- Preserve explicit notification-owner shutdown behavior without treating lag as a control signal.

This removes unrelated fan-out from active subscriptions and ensures that a signal already delivered to A cannot be evicted by traffic for other runs.

## Acceptance criteria

- After subscribing A, publishing more than 256 notifications for other run IDs does not resolve A as either `Handoff` or `Release`.
- A subsequent A-specific `Handoff` resolves A exactly as `Handoff`.
- A subsequent A-specific `Release` resolves A exactly as `Release`.
- If A's real signal arrives after subscription but before `wait()` drains it, more than 256 later unrelated notifications cannot evict, replace, or misclassify that signal.
- Notification-before-subscription delivery continues to work and is consumed once.
- Dropping a subscription removes its waiter without allowing an older subscription to remove a newer waiter for the same run ID.
- The implementation does not perform work proportional to all active subscriptions for each unrelated notification.
- Deterministic Runner tests cover unrelated overflow, both real signal variants, notification-before-subscription delivery, and subscription cleanup.

## Compatibility and scope

- This is a Runner-local notification-routing correction.
- It requires no database schema, persisted-data migration, API contract, Ably event, Guest Agent protocol, or sandbox image change.
- The `piLoop`-disabled path remains unchanged.
- General durable delivery across Runner restarts or arbitrary loss before any local subscription exists is not introduced by this correction.

seven332:
## Summary

- replace the process-wide Pi standby broadcast stream with run-keyed one-shot waiters
- retain bounded pending storage only for signals that arrive without an active waiter
- make same-run replacement and cancellation cleanup registration-safe while preserving exact handoff/release signals
- add deterministic coverage for unrelated overflow, delivered-signal retention, pending consumption, replacement, cancellation, failed one-shot delivery, and final-owner shutdown

## Scope

This is Runner-local. It does not change API or Ably contracts, guest control payloads, persisted data, database schema, or sandbox images. Durable recovery after Runner restart or arbitrary realtime loss remains tracked by #25617.

## Validation

- `cargo test -p runner pi_standby` (15 passed)
- `cargo test -p runner` (3056 passed, 20 ignored)

Closes #25593

Parent context: #25596

### DIFF
--- a/crates/runner/src/executor/pi_standby.rs
+++ b/crates/runner/src/executor/pi_standby.rs
@@ -24,7 +24,7 @@ impl PiStandbyForwarder {
         control: Option<GuestProcessControlHandle>,
         job_cancel: CancellationToken,
     ) -> Option<Self> {
-        let (Some(mut source), Some(control)) = (source, control) else {
+        let (Some(source), Some(control)) = (source, control) else {
             return None;
         };
         let stop = CancellationToken::new();
@@ -35,6 +35,9 @@ impl PiStandbyForwarder {
                 () = task_stop.cancelled() => return,
                 () = job_cancel.cancelled() => return,
                 signal = source.wait() => signal,
+            };
+            let Some(signal) = signal else {
+                return;
             };
             let payload = match signal {
                 PiStandbySignal::Handoff => br#"{"type":"pi-handoff"}"#.as_slice(),
--- a/crates/runner/src/executor/pi_standby.rs
+++ b/crates/runner/src/executor/pi_standby.rs
@@ -70,3 +70,41 @@ impl PiStandbyForwarder {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::sync::Arc;
+    use std::sync::atomic::{AtomicBool, Ordering};
+
+    use sandbox::ProcessControlAck;
+
+    use super::*;
+    use crate::pi_standby::PiStandbyNotifications;
+
+    #[tokio::test]
+    async fn superseded_source_does_not_send_guest_control() {
+        let notifications = PiStandbyNotifications::new();
+        let run_id = RunId::new_v4();
+        let source = notifications.subscribe(run_id);
+        let _replacement = notifications.subscribe(run_id);
+        let control_called = Arc::new(AtomicBool::new(false));
+        let task_control_called = Arc::clone(&control_called);
+        let control = GuestProcessControlHandle::new(move |message_id, _, _| {
+            task_control_called.store(true, Ordering::SeqCst);
+            Box::pin(async move { Ok(ProcessControlAck { message_id }) })
+        });
+        let forwarder = PiStandbyForwarder::start(
+            run_id,
+            Some(source),
+            Some(control),
+            CancellationToken::new(),
+        )
+        .expect("complete Pi standby controls should start a forwarder");
+        let PiStandbyForwarder { task, .. } = forwarder;
+
+        task.await
+            .expect("Pi standby forwarder should exit cleanly");
+
+        assert!(!control_called.load(Ordering::SeqCst));
+    }
+}
--- a/crates/runner/src/pi_standby.rs
+++ b/crates/runner/src/pi_standby.rs
@@ -153,6 +153,7 @@ impl PiStandbyNotifications {
 }
 
 impl PiStandbySubscription {
+    /// Returns `None` when a newer subscription replaces this registration.
     pub(crate) async fn wait(mut self) -> Option<PiStandbySignal> {
         let result = (&mut self.receiver).await;
         match result {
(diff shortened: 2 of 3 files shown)


seven332:
## Code Review: PR #25629

**Reviewed head:** `e2b1214d5adb98f6126df378c9182c1e7db0a9d6`

### Summary

The PR replaces cross-run broadcast fan-out with a run-keyed one-shot registry while retaining bounded notification-before-subscription delivery. Atomic state transitions, registration-scoped cleanup, weak ownership, and sent-value recovery cover replacement, cancellation, and final-owner shutdown without inferring control signals from unrelated traffic.

### Findings

No findings.

### Testing

Regression coverage deterministically exercises unrelated overflow, exact handoff and release preservation, pending single-consumption, both subscription-drop orderings, same-run replacement, and notification-owner shutdown. Local validation passed formatting, focused Pi standby tests, all-target/all-feature Clippy, workspace docs, and the full Runner suite (3054 passed, 20 ignored). CI coverage and deployment jobs remain to be confirmed by PR checks.

### Verdict

LGTM

seven332:
## Code Review: PR #25629

**Reviewed head:** `5a49e69dd92383af16e1f92bad7f5c6dc3dd0863`

### Summary

The run-keyed one-shot design remains correct and scoped. The new head also covers the two meaningful branches identified by Codecov: a superseded source exits the executor without guest control, and a signal rejected by a concurrently closed receiver returns to pending storage.

### Findings

No findings.

### Testing

The tests use the real subscription/forwarder path and process-control boundary, with no sleeps, timeout increases, or internal mocks. Local validation passed formatting, 15 focused Pi standby tests, all-target/all-feature Clippy, workspace docs through the pre-commit hook, and the full Runner suite (3056 passed, 20 ignored). Current-head CI remains to be confirmed by PR checks.

### Verdict

LGTM

PR #25633, opened 2026-08-07
fix(python): revalidate auth.base authorization after auth waits

### LINKED ISSUE #25611: bug(python): revalidate auth.base authorization after auth waits
## Problem

At `2e39ce49933a72e916be23b4c9860d1b3b86cb58`, an `auth.base` request can outlive the registry authorization that selected it and still be forwarded with resolved credentials.

The request hook initially selects a `FirewallAllow`, records the original run and server context, and passes a current-authorization callback into firewall auth (`crates/runner/mitm-addon/src/mitm_addon.py:1401-1416`). Firewall auth then suspends while resolving credentials (`crates/runner/mitm-addon/src/auth.py:1322-1325`). After that await, the callback is invoked only when `context.auth_request.auth_base is None` (`crates/runner/mitm-addon/src/auth.py:1340-1346`). The `auth.base` branch therefore skips current registry authorization and proceeds through `_apply_resolved_firewall_auth()` into `_apply_url_rewrite()` (`crates/runner/mitm-addon/src/auth.py:1155-1191`), which submits the resolved request to `forward_request()` (`crates/runner/mitm-addon/src/auth.py:1079-1087`).

The exclusion is observable in existing coverage: the successful rewrite test supplies a guard that fails if called, confirming that `auth.base` forwarding currently bypasses the post-resolution guard (`crates/runner/mitm-addon/tests/test_firewall_rewrite_success.py:80-87`).

## Impact

A concrete failure sequence is:

1. A request is authorized for an `auth.base` firewall under the current run and network policy.
2. `get_firewall_headers()` waits on the external auth endpoint.
3. While it is suspended, the registry removes the VM, assigns the client IP to another run, or revokes the matched permission for the same run.
4. The old auth result returns with a resolved base and headers.
5. Because the ordinary-credential guard is explicitly skipped, the addon forwards the request and can cause an external side effect under authorization that is no longer current.

The buffered request revalidation added for #24927 occurs before external auth starts, so it does not cover a registry change during this later await. The ordinary credential revalidation from #24925 / #24973 intentionally excluded `auth.base`, so this is a separate boundary rather than a duplicate.

## Acceptance criteria

- Re-establish canonical current registry/TLS/firewall authorization after `auth.base` resolution and immediately before submitting the resolved forward.
- Continue only for the same run and an equivalent current `FirewallAllow`; fail closed for registry unavailability, VM removal or reassignment, same-run permission revocation, ambiguity, or a changed route decision.
- A rejected request must not call `forward_request()`, apply or expose resolved credentials, or retain forwarding admission or usage tracking.
- Preserve unchanged authorized requests and unrelated same-run policy updates whose canonical decision remains equivalent.
- Add deterministic integration coverage that blocks auth resolution, mutates a real temporary registry, releases the old result, and verifies both non-forwarding and exact-once lifecycle cleanup for removal, run replacement, and permission revocation.

seven332:
## Summary

- require current canonical firewall authorization after every managed auth resolution, including `auth.base`
- preserve direct connector binding checks only for credentials that mutate the ordinary mitmproxy upstream request
- fail closed before resolved rewrite data or forwarding admission reaches `forward_request()`
- cover registry removal, run replacement, permission revocation, equivalent policy updates, and admission/usage cleanup through real addon hooks

## Scope

- no API, database, persisted state, runner protocol, registry schema, generated contract, migration, frontend, or feature-switch changes
- resolved `auth.base` destination validation and ordinary credential authorization behavior remain unchanged

## Validation

- `uv lock --check`
- `uv sync --locked`
- focused addon suite: 146 passed
- mutation check: all three new registry-change cases fail when the old `auth.base` guard exclusion is restored
- `uv run --no-sync python -m pytest tests/` (4,959 passed)
- `lefthook run pre-commit`

Closes #25611

### DIFF
--- a/crates/runner/mitm-addon/src/auth.py
+++ b/crates/runner/mitm-addon/src/auth.py
@@ -52,10 +52,7 @@ from firewall_auth_client import (
     FirewallAuthRequest,
     InsufficientCreditsError,
 )
-from firewall_auth_config import (
-    auth_config_injects_credentials,
-    auth_config_injects_ordinary_upstream_credentials,
-)
+from firewall_auth_config import auth_config_injects_credentials
 from logging_utils import log_proxy_entry
 from runtime_url_parsing import split_runtime_url
 from url_syntax import has_unsafe_runtime_url_syntax
@@ -77,7 +74,7 @@ class FirewallHeaderPhaseAuthResult(Enum):
     FALLBACK = "fallback"
 
 
-type OrdinaryUpstreamCredentialsGuard = Callable[[], bool]
+type CurrentFirewallAuthorizationGuard = Callable[[], bool]
 
 
 _HTTP_STATUS_INFORMATIONAL_MIN = 100
@@ -1266,14 +1263,13 @@ async def handle_firewall_request(
     allow: matching.FirewallAllow,
     vm_info: dict,
     *,
-    revalidate_ordinary_upstream_credentials: OrdinaryUpstreamCredentialsGuard,
+    revalidate_current_firewall_authorization: CurrentFirewallAuthorizationGuard,
 ) -> FirewallAuthHandlingResult:
     """Handle firewall auth and return who owns the next response lifecycle.
 
-    The required guard runs after resolution and before ordinary upstream
-    credential mutation. A rejecting caller must first create its local
-    response; this function then returns ``LOCAL_RESPONSE`` without applying
-    resolved credentials.
+    The required guard runs after resolution and before managed credential
+    application. A rejecting caller must first create its local response; this
+    function then returns ``LOCAL_RESPONSE`` without applying resolved data.
     """
     try:
         _prepare_firewall_metadata(flow, allow, vm_info)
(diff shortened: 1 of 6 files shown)


seven332:
## Code Review: PR #25633

**Reviewed head:** `dd6d3e79f5cb60ff2e15cc1eb6722414d287c115`

### Summary

The change closes the `auth.base` post-resolution authorization gap by applying the existing canonical current-authorization invariant to every managed credential path. Ordinary upstream credentials retain their direct connector-binding check, while auth.base-only routes avoid that unrelated requirement. Rejection occurs before resolved credentials, success metadata, forwarding admission, or the forwarding boundary are reached.

### Findings

No findings.

### Testing

The regression suite uses real addon hooks and atomic temporary-registry replacement for VM removal, run replacement, permission revocation, and equivalent same-run policy updates. It verifies no forward or credential exposure plus admission and usage cleanup. The old auth.base exclusion was mutation-tested and makes all three denial cases fail. Focused tests passed (146), the full addon suite passed (4,959), and Ruff, basedpyright, uv lock/sync, and pre-commit checks passed.

### Verdict

LGTM

PR #25662, opened 2026-08-07
perf(runner): avoid rewalking unchanged storage cache entries

### LINKED ISSUE #25641: perf(rust): avoid rewalking unchanged storage cache entries during eviction
## Location

- crates/runner/src/cmd/gc/storage.rs:31-37
- crates/runner/src/cmd/gc/storage.rs:187-218
- crates/runner/src/cmd/gc/storage.rs:260-392
- crates/runner/src/cmd/gc/filesystem.rs:90-103
- crates/runner/src/cmd/gc/filesystem.rs:113-160
- crates/runner/src/storage_cache.rs:2417-2490
- ansible/tasks/garbage-collect-runner.yml:1-12

## Problem

Storage cache GC recursively accounts each unlocked version directory during inventory with dir_stats and stores its size and directory mtime in StorageCandidate. When a cap is exceeded, evict_storage_candidate reacquires the same per-version lock, stats the directory through gc_path_dir_status, discards that returned metadata, and then calls dir_stats again before deletion. The second call recursively walks the whole entry again.

Normal final cache entries are immutable after publication: write_to_cache creates a staging directory containing archive.tar.gz and atomically renames that directory into place. Cache hits only refresh the final directory mtime. Once eviction has reacquired the per-version lock, an unchanged directory can therefore be validated from stable identity and mtime metadata without recursively recomputing the size already collected during inventory. A changed or replaced entry can still fall back to fresh accounting.

The current path also performs an immediately redundant root stat: gc_path_dir_status already returned Metadata, then dir_stats stats the same root again before starting its traversal.

## Impact

The production cache is allowed to contain 5,000 entries, and runner GC runs during promotion and rollback. On a host above the 1 GiB or entry-count cap, evicting N unchanged candidates performs N additional serial recursive walks and dispatches metadata/read-directory work through tokio filesystem operations a second time. High-cardinality cleanup can therefore double storage-tree I/O for hundreds or thousands of evictions, extending deployment GC latency and adding avoidable blocking-pool work. Unexpected deep debris inside a version directory amplifies the duplicate traversal further.

## Acceptance criteria

- Preserve the initial full-directory accounting used for the byte cap.
- Record enough scan-time identity metadata to recognize an unchanged cache entry after its per-version lock is reacquired.
- Reuse the scan-time size for an unchanged entry instead of recursively calling dir_stats again.
- Use the Metadata already returned by gc_path_dir_status for current identity and age checks rather than immediately restatting the root.
- Fall back to fresh directory accounting when identity or relevant metadata changed, so replacements and concurrent cache activity remain safe and accurately reported.
- Preserve locked-entry protection, the ten-minute recent-entry guard, dry-run byte reporting, missing/non-directory handling, lock-file cleanup, and removal-failure accounting.
- Add focused coverage showing that unchanged eviction candidates are accounted recursively once, while changed/replaced and newly recent candidates retain the current safe behavior.

## Historical context

- #11547 introduced the 5,000-entry bound after a host accumulated 38,196 version directories while using only 434 MiB.
- #11515 and #11547 established the current two-phase lock/revalidation design; this optimization should retain that design and remove only the redundant second tree walk.

seven332:
## Summary

- retain the root metadata already fetched during storage cache inventory
- reuse scan-time allocated bytes when the eviction lock sees the same device, inode, and mtime
- fall back to fresh recursive accounting for replaced, modified, or unidentifiable candidates
- cover unchanged reuse, real-directory replacement, and newly recent accounting with real filesystem tests

## Scope

This is a runner-local storage GC optimization. It does not change cache layout, admission policy, configuration, persisted state, APIs, or runner protocols. Other GC domains are unchanged.

## Validation

- `cargo test --manifest-path crates/Cargo.toml -p runner cmd::gc::storage::tests`
- repository pre-commit hooks for staged Rust files

Closes #25641

### DIFF
--- a/crates/runner/src/cmd/gc/filesystem.rs
+++ b/crates/runner/src/cmd/gc/filesystem.rs
@@ -115,15 +115,37 @@ pub(super) async fn gc_entry_is_real_dir(entry: &tokio::fs::DirEntry) -> std::io
 /// Last-used time comes from the root directory's own mtime, which `touch_mtime`
 /// updates on every cache hit and `runner start`.
 pub(super) async fn dir_stats(dir: &Path) -> (u64, SystemTime) {
+    let stats = collect_dir_stats(dir).await;
+    (stats.size, stats.mtime)
+}
+
+pub(super) struct DirStats {
+    pub(super) size: u64,
+    pub(super) mtime: SystemTime,
+    pub(super) root_metadata: Option<std::fs::Metadata>,
+}
+
+/// Compute directory stats while retaining the root metadata fetched for the walk.
+pub(super) async fn collect_dir_stats(dir: &Path) -> DirStats {
     const BYTES_PER_BLOCK: u64 = 512;
 
     let root_meta = match tokio::fs::symlink_metadata(dir).await {
         Ok(meta) => meta,
-        Err(_) => return (0, SystemTime::UNIX_EPOCH),
+        Err(_) => {
+            return DirStats {
+                size: 0,
+                mtime: SystemTime::UNIX_EPOCH,
+                root_metadata: None,
+            };
+        }
     };
     let mtime = root_meta.modified().unwrap_or(SystemTime::UNIX_EPOCH);
     if !root_meta.file_type().is_dir() {
-        return (root_meta.blocks() * BYTES_PER_BLOCK, mtime);
+        return DirStats {
+            size: root_meta.blocks() * BYTES_PER_BLOCK,
+            mtime,
+            root_metadata: Some(root_meta),
+        };
     }
 
     let mut total_bytes = 0u64;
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #25662

**Reviewed head:** `b0e20b58f2cc8bb1db7401ec393a68e5dec42849`

### Summary

The storage GC fast path is limited to completely scanned entries whose device, inode, and mtime remain unchanged after the per-version lock is reacquired. Changed, replaced, recent, or incompletely scanned entries retain fresh recursive accounting.

### Findings

No findings.

### Testing

Focused storage/filesystem tests, the full runner suite, formatting, Clippy, documentation checks, and repository pre-commit hooks passed. Real-filesystem regressions cover unchanged reuse, replacement, recent-entry protection, and incomplete-scan fallback. A production benchmark was not run.

### Verdict

LGTM

PR #25664, opened 2026-08-07
test(runner): cover orphan process-group members

### LINKED ISSUE #25642: test(rust): exercise orphan process-group members
## Location

- Production process-group signal: `crates/runner/src/cmd/kill/orphan.rs:453-488`
- Current real-process regression test: `crates/runner/src/cmd/kill/orphan.rs:938-1012`
- Related process-group test that does not close the gap: `crates/sandbox-fc/src/sandbox/tests.rs:3523-3552`

## Problem and evidence

`runner kill` deliberately sends `SIGKILL` to the validated Firecracker process group, then waits only for the recorded group-leader identity. The real-process regression test creates a `/bin/sh` group leader whose script uses only the `printf` and `read` shell builtins. It therefore has no second process-group member, and the test asserts only the leader's `Outcome::Killed` result and exit status.

Replacing `killpg(pgid, SIGKILL)` with a leader-only `kill(pid, SIGKILL)` would leave this test green. The nearby `killpg_kills_entire_process_group` test does not compensate: `bash -c 'sleep 60'` may replace itself with the final command, and the test neither captures nor checks a distinct member identity. It also asserts only that the group leader is gone.

This leaves the reason for using `killpg` unprotected. A regression can terminate the Firecracker leader, let `wait_for_orphan_exit` report success, and allow manual orphan cleanup to continue while another process in the same sandbox process group remains alive.

Audited at `ea90ade0a378f12180b6989551c72b63253fcdad`; confirmed unchanged on `776af08ffdafb04c9ae848396990e8e119358fa3`.

## Decision

Proceed with one PR that strengthens the existing runner integration test. The invariant to protect is process-group membership, not parent/descendant ancestry. The fixture should therefore use a stable leader plus a distinct, explicitly verified long-lived member in the same PGID, while retaining ownership that lets the test reap every spawned process deterministically.

## Alternatives considered

- A shell-spawned descendant would resemble a process tree, but it would not be directly reapable by the test without adding subreaper machinery. Ancestry is also not what `killpg` guarantees.
- Extending only the `sandbox-fc` test would not exercise `runner kill` validation, signaling, and outcome behavior together.
- Leaving the tests unchanged accepts a silent regression path in destructive orphan cleanup.

## Risks and mitigations

- Real-process tests can race with PID reuse or process startup. Capture each PID's start time and PGID before termination, then use bounded identity-aware polling.
- A failed assertion could leak a long-lived process. Keep direct child handles and use explicit cleanup/reaping on every exit path.
- The change is test-only and does not alter runner behavior, protocols, persisted data, or deployment compatibility.

## Acceptance criteria

- The orphan-kill real-process fixture creates a stable group leader and at least one distinct long-lived member that shares the leader's PGID; it verifies both identities before invoking `terminate()`.
- After `terminate()` returns `Outcome::Killed`, bounded identity-aware polling verifies that both captured identities are no longer live.
- Replacing the process-group signal with a leader-only signal causes the focused test to fail because the additional member remains live.
- Success, failure, and timeout paths deterministically kill and reap every process created by the fixture.

seven332:
## Summary

- add a second directly owned long-lived process to the orphan-kill fixture and place it in the fake Firecracker leader's PGID
- capture and verify both process identities through the real `terminate` path
- clean up and reap both child handles before asserting results so regression failures cannot leak the surviving member

## Scope

- test-only change; production signaling remains unchanged
- no API, protocol, schema, migration, persisted-data, or deployment-order changes

## Validation

- `cargo test -p runner --all-targets --all-features` (3,070 passed, 20 ignored; integration test passed)
- `cargo test -p runner orphan_kill_validates_signals_and_waits_for_real_exit -- --nocapture`
- mutation check: temporarily replacing `killpg` with a leader-only signal made the focused test fail with `group member remained live`; restoring `killpg` made it pass

Closes #25642

### DIFF
--- a/crates/runner/src/cmd/kill/orphan.rs
+++ b/crates/runner/src/cmd/kill/orphan.rs
@@ -960,7 +960,7 @@ mod tests {
         let firecracker = base_dir.path().join("firecracker");
         std::os::unix::fs::symlink("/bin/sh", &firecracker).unwrap();
 
-        let mut child = tokio::process::Command::new(&firecracker)
+        let mut leader = tokio::process::Command::new(&firecracker)
             .arg("-c")
             .arg("printf '%s\\n' \"$1\"; IFS= read -r _")
             .arg("vm0-orphan-kill-test")
@@ -973,7 +973,7 @@ mod tests {
             .stderr(std::process::Stdio::null())
             .spawn()
             .unwrap();
-        let stdout = child.stdout.take().unwrap();
+        let stdout = leader.stdout.take().unwrap();
         let mut stdout_lines = BufReader::new(stdout).lines();
         let ready_line = tokio::time::timeout(Duration::from_secs(5), stdout_lines.next_line())
             .await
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #25664

**Reviewed head:** `2c816897fbd472db14e8ae4eaff2aebb15ebc4b2`

### Summary

The real-process fixture now proves the production process-group invariant with a distinct live member, identity-aware observations, and cleanup/reaping before result assertions. The change is test-only and introduces no API, persistence, protocol, or rollout impact.

### Findings

No findings.

### Testing

Formatting, Clippy, runner documentation, the focused regression test, ten repeated focused runs, and all runner targets passed. A leader-only signaling mutation failed because the additional member remained live, then passed after restoring `killpg`.

### Verdict

LGTM

PR #25670, opened 2026-08-07
fix(python): keep local json responses bodyless for head

### LINKED ISSUE #25646: bug(python): keep local JSON responses bodyless for HEAD
## Problem and evidence

Local JSON response constructors in the runner mitmproxy addon attach response content without consistently applying the bodyless semantics required for `HEAD`.

The first confirmed ownership area is `crates/runner/mitm-addon/src/http_local_responses.py`:

- `matching.py` accepts `HEAD` as a firewall rule method.
- `mitm_addon.py` dispatches authority, registry, TLS-admission, authorization, destination, host-policy, firewall-block, ambiguous-route, and public-destination outcomes to the shared constructors.
- All ten response-producing branches in `http_local_responses.py` serialize JSON unconditionally.

At `ea90ade0a378f12180b6989551c72b63253fcdad`, an executable reproduction used the pinned `mitmproxy==12.2.3` state machine:

1. Start an HTTP/1.1 or HTTP/2 `HEAD` request through mitmproxy's real `HttpLayer`.
2. Reach `set_firewall_block_response()` from the normal request hook with a `FirewallBlock(method="HEAD", ...)`.
3. Complete the response-header and response hooks.
4. Inspect client-facing output.

Observed for both protocols:

```text
response_body_bytes=288
content_length=288
wire_contains_json_body=True
```

HTTP/1.1 writes the JSON bytes after the response headers. HTTP/2 emits a non-empty DATA frame. Pinned mitmproxy does not provide a later generic suppression step for addon-created local responses.

Plan review found a second reachable ownership area that the original scope omitted: `auth.py::_set_matched_firewall_failure_response()` also creates JSON content unconditionally. `HEAD` is allowed for managed-credential firewall rules; only `TRACE` and `TRACK` are forbidden. Preflight, auth-resolution, invalid-header, signing, admission, and forwarding failures can therefore create the same protocol-invalid response for an allowed `HEAD` rule.

This remains distinct from #25147 / #25167, which fixed connector-diagnostic synthesis and replacement. Codex model-catalog local responses are GET-specific, while auth-base forwarded responses already have separate tested upstream `HEAD` representation handling.

## Decision

Fix every currently reachable local JSON constructor with this missing invariant in one coherent PR: the ten shared constructors and the matched-firewall auth failure constructor. Do not add a global response-hook sanitizer, because it would also alter already-correct upstream and specialized response lifecycles.

For these synthetic `HEAD` responses, retain `Content-Type: application/json` and omit `Content-Length`. RFC 9110 permits omission and permits a value only when it equals the corresponding `GET` representation length. Method-specific firewall outcomes cannot generally provide that trustworthy length; `0` and the serialized `HEAD` error length are not equivalent metadata.

## Impact

A response to `HEAD` must not contain message content. Unexpected bytes can desynchronize a persistent HTTP/1.1 connection or produce invalid/unexpected HTTP/2 DATA. Affected outcomes include shared admission and firewall denials plus matched-firewall authentication and forwarding failures.

## Acceptance criteria

- Every currently reachable synthetic local JSON response to `HEAD` in `http_local_responses.py` and the matched-firewall auth failure path has empty response content before mitmproxy framing.
- HTTP/1.1 emits no bytes after the response headers and HTTP/2 emits no response-body DATA payload.
- Synthetic `HEAD` responses retain JSON `Content-Type` and omit `Content-Length`.
- Status codes, firewall/auth metadata, proxy and network logging, diagnostic fields, and terminal behavior remain intact.
- Non-`HEAD` local responses retain their existing JSON bytes and framing.
- Integration coverage exercises representative shared-denial and firewall-auth-failure paths through addon entry points.
- Pinned mitmproxy integration coverage exercises the normal request-headers/request lifecycle on HTTP/1.1 and HTTP/2 and inspects client-facing output.
- No global sanitizer changes forwarded upstream `HEAD`, connector-diagnostic, streaming, catalog-cache, or auth-base response semantics.

## Delivery notes

- This is one independently reviewable runner mitm-addon PR.
- No schema, persisted data, API, runner protocol, queue payload, migration, feature switch, coordinated rollout, or cleanup issue is required.

seven332:
## Summary

- centralize local JSON response construction for shared denials and matched-firewall auth failures
- keep synthetic HEAD responses bodyless while preserving JSON Content-Type and omitting untrusted Content-Length
- cover handler behavior and pinned mitmproxy HTTP/1.1 and HTTP/2 client framing

## Scope

- non-HEAD JSON, status, metadata, and logging behavior remains unchanged
- forwarded upstream, connector diagnostic, catalog, streaming, and auth-base semantics remain unchanged
- no schema, API, protocol, persistence, migration, feature switch, or cleanup work

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (`4963 passed`)

Closes #25646

### DIFF
--- a/crates/runner/mitm-addon/src/auth.py
+++ b/crates/runner/mitm-addon/src/auth.py
@@ -17,6 +17,7 @@ from mitmproxy import http
 
 import flow_metadata
 import flow_metadata_keys as metadata_keys
+import http_local_responses
 import matching
 from auth_base_forwarder import (
     MAX_AUTH_BASE_REQUEST_BODY_BYTES,
@@ -309,10 +310,10 @@ def _set_matched_firewall_failure_response(
         body["connectors"] = connectors
     if failure_reason:
         body["failureReason"] = failure_reason
-    flow.response = http.Response.make(
+    flow.response = http_local_responses.make_local_json_response(
+        flow,
         status,
-        json.dumps(body).encode(),
-        {"Content-Type": "application/json"},
+        body,
     )
 
 
--- a/crates/runner/mitm-addon/src/http_local_responses.py
+++ b/crates/runner/mitm-addon/src/http_local_responses.py
@@ -30,6 +30,23 @@ def _diagnostic_url_without_query_or_fragment(original_url: str) -> str:
     return urllib.parse.urlunsplit((parts.scheme, parts.netloc, parts.path, "", ""))
 
 
+def make_local_json_response(
+    flow: http.HTTPFlow,
+    status_code: int,
+    body: dict[str, object],
+) -> http.Response:
+    is_head_request = flow.request.method.upper() == "HEAD"
+    content = b"" if is_head_request else json.dumps(body).encode()
+    response = http.Response.make(
+        status_code,
+        content,
+        {"Content-Type": "application/json"},
+    )
+    if is_head_request:
+        del response.headers["Content-Length"]
+    return response
+
+
 def block_authority_validation_error(
     flow: http.HTTPFlow,
     error: AuthorityValidationError,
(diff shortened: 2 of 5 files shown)


seven332:
## Code Review: PR #25670

**Reviewed head:** `a6f5bef2cdfa3ebba850ddfbc7eed88004d38b4a`

### Summary

The change applies one construction-time bodyless contract to both confirmed local JSON response owners without altering forwarded or specialized response lifecycles. The implementation preserves non-HEAD serialization and verifies actual HTTP/1.1 and HTTP/2 client framing through pinned mitmproxy.

### Findings

No findings.

### Testing

The two owners are covered through addon entry points, and protocol tests assert no HTTP/1.1 response body or HTTP/2 DATA payload. Ruff, BasedPyright, 96 affected-file tests, and the complete addon suite (`4963 passed`) succeeded. Live external networking was not exercised because these are locally synthesized responses.

### Verdict

LGTM

PR #25674, opened 2026-08-07
test(python): cover claude content-block-first timing

### LINKED ISSUE #25655: test(python): cover content-block-first Claude timing fallback
## Context

`observe_lifecycle_event()` intentionally creates run timing state when a qualifying `content_block_start` (`thinking`, `redacted_thinking`, or `text`) is the first observable Anthropic lifecycle event (`crates/runner/mitm-addon/src/claude_output_timing.py:87-106`). This preserves first-output milestones when the earlier `message_start` event was missing, truncated, or otherwise not observed. The production Anthropic SSE path forwards each successfully parsed lifecycle event independently through `response_streaming._anthropic_lifecycle_observer()`.

Every focused call in `tests/test_claude_output_timing.py` sends `message_start` before a qualifying content block, and the cross-provider test also initializes Claude state with `message_start`. Targeted branch coverage therefore leaves the `state is None` initialization at lines 89-92 unexecuted.

A shadow mutation that replaces this fallback with an immediate return leaves all 11 Claude timing and cross-provider timing tests green. In a deterministic content-block-first diagnostic, the current implementation retains both `claude_proxy_first_thinking_or_text_block_start` and `claude_proxy_first_text_block_start`; the mutation creates no run state and loses both milestones.

## Impact

A refactor could remove or break the fallback while the current suite remains green. A partial, midstream, or reconnected Claude SSE sequence whose first observable lifecycle event is a qualifying content block would then silently lose first-output timing telemetry and skew Claude latency measurements.

## Acceptance criteria

- Add a focused Claude output-timing test where the first observable lifecycle event is a qualifying `content_block_start`, with no prior `message_start`.
- Assert that a `text` block retains and admits both the thinking-or-text and text milestones using the same observation timestamp.
- Assert that no message-start milestone is fabricated for an event that was not observed.
- Keep the report content-free and bound to the correct run ID.
- Preserve the behavior that a non-qualifying block such as `tool_use` does not create output timing state by itself.

seven332:
## Summary

- exercise Claude timing fallback when a valid `text` content block is the first observed lifecycle event
- verify both block milestones share one timestamp without fabricating `message_start`, leaking provider content, or losing run identity
- cover a standalone non-qualifying `tool_use` block through the production SSE stream path

## Scope

Test-only change. No production code, API, schema, persistence, dependency, migration, or rollout behavior changes.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_claude_output_timing.py` (10 passed)
- `uv run --no-sync python -m pytest tests/` (4959 passed)

Closes #25655

### DIFF
--- a/crates/runner/mitm-addon/tests/test_claude_output_timing.py
+++ b/crates/runner/mitm-addon/tests/test_claude_output_timing.py
@@ -173,7 +173,7 @@ def test_reports_content_free_lifecycle_milestones_and_preserves_usage(
     assert secret not in read_jsonl_text_after_flush(proxy_log)
 
 
-def test_text_first_uses_one_observation_time_for_broad_and_text_milestones(
+def test_content_block_first_retains_broad_and_text_milestones(
     tmp_path: Path,
     real_flow,
     mitm_ctx,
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #25674

**Reviewed head:** `1ae2f39cd710d1d916a0527e7689272699dcb3ed`

### Summary

The test-only change correctly exercises the Claude content-block-first fallback through the production Anthropic SSE stream path. It preserves the existing shared-timestamp contract, verifies no `message_start` milestone is fabricated, binds telemetry to the expected run, checks provider content isolation, and covers standalone non-qualifying `tool_use` behavior.

### Findings

No findings.

### Testing

Local validation passed: uv lock/environment checks, Ruff format/lint, BasedPyright, the focused Claude timing suite (10 passed), and the complete mitm-addon suite (4959 passed). Remote CI remains for the PR check stage.

### Verdict

LGTM

Issue #25163, opened 2026-08-05
feat: run workflows when a Google Form receives a new response


lancy:
## Summary

Let users attach a workflow trigger that fires when a Google Form receives a new response.

This is the Google Forms half of #23860. The Google Sheets half (new row / changed cell triggers) is explicitly **not** in this epic and stays on #23860, because Google offers no native change feed for Sheets and every option there requires vm0 to build its own snapshot/diff state.

Forms is different: the Forms API `forms.watches` collection delivers change notifications to a Cloud Pub/Sub topic, so Google does the change detection for us. The existing Gmail event source is the template — Pub/Sub push with OIDC verification, a watch-state table, a renewal cron, and a per-automation dedupe table. Most of this epic is filling in that template.

Depends on #23859 (Google Forms connector). The connector shipped behind `googleFormsConnector`, and #24986 enabled the switch globally, so the connector is live in production.

## Verified before writing this

Live checks against the production connector (account `lancy@vm0.ai`):

- `forms.create`, `forms.get`, `forms.responses.list` and `GET /v1/forms/{formId}/watches` all work.
- `forms.watches.create` reaches Google and fails **only** on the Pub/Sub topic: `403 "Topic parent project not found or permission denied."` A bogus formId returns `404` first; a malformed topic name returns `400 "Invalid topic name."` Those three responses together isolate the failure to the topic, which means the existing connector scopes (`forms.body` + `forms.responses.readonly`) are sufficient to create a RESPONSES watch. **No Drive scope and no restricted-scope review are needed.**
- `responses.list` accepts a `fields` mask, so the server can read response IDs and timestamps without ever reading answer content.
- `watches.create` sits on the connector's deny list for agent-side calls. The production path (API server, user's token, outside the sandbox) is unaffected, so the agent-side deny stays as is.

Not yet verifiable without a Pub/Sub topic: everything in the "Post-deploy" checklist below.

## Scope

**In scope:** one new event type, `google-forms-response-submitted`, backed by Forms `EventType.RESPONSES`.

**Non-goals:**

- `EventType.SCHEMA` (form structure edits) — no user story, and every event type costs its own watch against a per-form limit.
- Answer-level filtering — `responses.list` only filters by timestamp, and server-side answer matching would break the "the server never reads answers" property. Conditions belong in the workflow's own SKILL.md.
- Answer diffs for edited responses — would require storing a copy of user answers, which no existing event source does.
- Google Sheets triggers — remain on #23860.

[...]

PR #25675, opened 2026-08-07
fix: parse google forms publish state as an object

### LINKED ISSUE #25663: bug: google forms automation creation returns 500 because publishSettings is mis-typed
## Summary

Creating a `google-forms-response-submitted` automation fails with `500 Internal server error` for every real form, so the trigger shipped in #25308 cannot be used at all in production.

Reproduced twice against production (`api.vm0.ai`, 2026-08-07 08:33:17Z and 08:33:57Z):

```
POST /api/zero/workflows/{workflowId}/automations
{"kind":"event","eventType":"google-forms-response-submitted",
 "eventConfig":{"provider":"google-forms","event":"response_submitted",
                "formUrl":"https://docs.google.com/forms/d/<formId>/edit"}}

-> 500 {"error":"Internal server error"}
```

A schedule automation created on the same workflow with the same credentials seconds earlier returned `201`, so this is specific to the Google Forms path, not to auth, the workflow, or the caller.

## Root cause

`vm0-web-logs-prod` shows the request dying on response validation:

```
Unhandled request error: [ {
  "expected": "string",
  "code": "invalid_type",
  "path": [ "publishSettings", "publishState" ],
  "message": "Invalid input: expected string, received object"
} ]
```

`googleFormSchema` in `turbo/apps/api/src/signals/services/google-forms-workflow-event.service.ts:67` models the field as:

```ts
publishSettings: z
  .object({
    publishState: z.string().optional(),
    isPublished: z.boolean().optional(),
    isAcceptingResponses: z.boolean().optional(),
  })
  .passthrough()
  .optional(),
```

Google actually returns a nested object. Live response for a published form:

```json
{"formId":"...","publishSettings":{"publishState":{"isPublished":true,"isAcceptingResponses":true}}}
```

So `publishState` is an object, not a string, and `isPublished` / `isAcceptingResponses` live one level deeper than the schema expects. `forms.get` parsing throws before the creation flow can do anything else.

There is a second defect in the same block that the first one currently masks. `formIsNotAcceptingResponses` (same file, ~line 432) reads `form.publishSettings?.isPublished` and `form.publishSettings?.isAcceptingResponses`, which are always `undefined` against the real payload, then falls through to `publishSettings?.publishState !== "PUBLISHED"` — an object compared against a string, always true. Once the parse is fixed, every form including a fully published one would be reported as "not accepting responses", and D13's warning would fire on all of them.

## Fix

```ts
publishSettings: z
  .object({
    publishState: z
      .object({
        isPublished: z.boolean().optional(),
        isAcceptingResponses: z.boolean().optional(),
      })
      .passthrough()
      .optional(),
  })
  .passthrough()
  .optional(),
```

and read the two booleans through `publishSettings.publishState`. A form counts as not accepting responses when `isPublished` is `false` or `isAcceptingResponses` is `false`; treat an absent `publishState` (which is what an unpublished API-created form returns) as not accepting responses too.

## Why tests did not catch it

`webhooks-google-forms.test.ts` and the workflow-automation tests mock the Forms API, and the mock encodes the same invented shape, so the contract error cannot surface there. The epic (#25163) named `publishSettings.publishState` but never pasted the real payload, which left the nesting to guesswork. Whatever fix lands should assert against the real shape above rather than a hand-written one.

## Acceptance

- `POST /api/zero/workflows/{id}/automations` with a `google-forms-response-submitted` config against a published form returns `201` and no warning.
- The same call against an unpublished form returns `201` **with** the "not accepting responses" warning.
- A watch appears on the form (`GET /v1/forms/{formId}/watches`) bound to `projects/vm0-ai-488909/topics/forms-events`.

lancy:
## Summary

- model Google Forms publish settings using the nested publish state object returned by the Forms API
- classify only forms with both publication booleans set to true as accepting responses
- update API integration mocks to use real published and unpublished response shapes
- assert published forms omit the D13 warning while unpublished forms retain it

## Root cause

The Forms API returns the publication booleans under a nested publish state object. The response schema expected a string and the warning logic read the booleans from the parent object, so real forms failed response parsing and the fallback warning check would have misclassified published forms.

## Validation

- pnpm -F api exec vitest run src/signals/routes/__tests__/webhooks-google-forms.test.ts src/signals/routes/__tests__/zero-workflow-automations.test.ts --maxWorkers=4 --silent=passed-only (75 passed)
- pnpm -F api lint
- pnpm -F api check-types
- targeted Prettier check for the three changed files

Closes #25663
Related to #25163

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/webhooks-google-forms.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/webhooks-google-forms.test.ts
@@ -128,7 +128,12 @@ function configureFormsApi(): FormsApiRecorder {
         return HttpResponse.json({
           formId: FORM_ID,
           info: { title: FORM_TITLE },
-          publishSettings: { publishState: "PUBLISHED" },
+          publishSettings: {
+            publishState: {
+              isPublished: true,
+              isAcceptingResponses: true,
+            },
+          },
         });
       },
     ),
@@ -298,6 +303,7 @@ describe("Google Forms Pub/Sub webhook", () => {
       throw new Error("Expected a Google Forms response automation");
     }
     expect(created.body.eventConfig.connectorId).toBe(connector.id);
+    expect(created.body).not.toHaveProperty("warning");
 
     const watchId = formsApi.watchIds[0];
     if (!watchId) {
--- a/turbo/apps/api/src/signals/routes/__tests__/zero-workflow-automations.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/zero-workflow-automations.test.ts
@@ -211,8 +211,13 @@ function configureGoogleFormsCreationMock(args?: {
           formId: GOOGLE_FORM_ID,
           info: { title: "Customer survey" },
           publishSettings: args?.unpublished
-            ? {}
-            : { publishState: "PUBLISHED" },
+            ? { publishState: {} }
+            : {
+                publishState: {
+                  isPublished: true,
+                  isAcceptingResponses: true,
+                },
+              },
         });
       },
     ),
--- a/turbo/apps/api/src/signals/services/google-forms-workflow-event.service.ts
+++ b/turbo/apps/api/src/signals/services/google-forms-workflow-event.service.ts
@@ -69,9 +69,13 @@ const googleFormSchema = z.object({
   info: z.object({ title: z.string() }),
   publishSettings: z
     .object({
-      publishState: z.string().optional(),
-      isPublished: z.boolean().optional(),
-      isAcceptingResponses: z.boolean().optional(),
+      publishState: z
+        .object({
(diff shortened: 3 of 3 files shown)


lancy:
## Code Review: PR #25675 (Round 1) — LGTM :tada:

### Summary

The schema matches both observed Google Forms publish-state payloads, the accepting-response predicate implements D13 exactly, and the change remains isolated from delivery and other providers. The API integration mocks now encode the real published and unpublished structures, with explicit warning assertions for both outcomes.

### Testing Review

Coverage is adequate: the two affected API integration files pass all 75 tests, and API lint, type checking, and targeted formatting checks pass.

No critical or high-priority issues remaining. This PR is ready for merge.

---
*Completed after 1 round(s) of automated review-fix loop*

PR #25702, opened 2026-08-07
fix(platform): make the shared thread page scrollable

e7h4n:
## Problem

The public shared-thread page (`/share/threads/:id`) cannot be scrolled at all. Any conversation taller than one viewport is clipped, and the reader has no way to reach the rest of it — the wheel, a touch drag, and keyboard paging all do nothing.

Reproduced on production against https://app.okou.ai/share/threads/a4600006-e303-4be1-848e-04fda84bd1bd:

| element | `overflow-y` | `clientHeight` | `scrollHeight` |
| --- | --- | --- | --- |
| `html` / `body` | `hidden` | 633 | 633 |
| `#root` | `hidden` | 633 | **3120** |
| `main` | `visible` | 3120 | 3120 |

`window.scrollY` and `#root.scrollTop` both stay at `0` after a wheel scroll. About 80% of that thread is unreachable.

## Root cause

`index.css` pins the app shell to the viewport:

```css
html, body, #root {
  height: var(--zero-viewport-height);
  overflow: hidden;
}
```

That is deliberate — the chat UI wants inner panes to own their scrolling, and it is what keeps iOS from rubber-banding the shell. The consequence is that every route rendered with the `none` page layout must supply its own scroll container.

`SharedThreadPage` did not. Its `<main>` used `min-h-full`, which lets the element *grow* past the shell rather than scroll inside it, so the overflow was simply clipped by `#root`.

## Fix

Make `<main>` the scroll container:

```diff
-<main className="flex min-h-full flex-col bg-background text-foreground">
+<main className="flex h-full flex-col overflow-y-auto bg-background text-foreground">
```

`h-full` resolves against `#root`'s definite height, so the element is exactly one viewport tall and scrolls internally. This matches the pattern already used by the other `none`-layout routes (`auth-layout.tsx` uses `h-full min-h-0 overflow-y-auto`).

The `sticky top-0` header keeps working — it now sticks to the scrolling `<main>` instead of the document.

## Verification

Applied the class change against the live production page and confirmed `main` becomes `clientHeight: 633` / `scrollHeight: 3120` with a working `scrollTop` range of `0–2487`, and that the header stays pinned while scrolling.

## Tests

Added a regression test asserting the shared-thread page renders its own scroll container, so a future refactor cannot silently drop it back to a clipped `min-h-full`.

- `pnpm -F @vm0/app exec vitest run src/views/shared-thread-page` — 2 passed
- `pnpm -F @vm0/app lint` — clean
- `pnpm -F @vm0/app check-types` — clean

### DIFF
--- a/turbo/apps/platform/src/views/shared-thread-page/__tests__/shared-thread-page.test.tsx
+++ b/turbo/apps/platform/src/views/shared-thread-page/__tests__/shared-thread-page.test.tsx
@@ -50,4 +50,30 @@ describe("shared thread page", () => {
     expect(screen.queryByText("Agent")).not.toBeInTheDocument();
     expect(screen.queryByText("Owner")).not.toBeInTheDocument();
   });
+
+  it("owns its scrolling because the app shell clips overflow", async () => {
+    context.mocks.api(sharedThreadsContract.get, ({ respond }) => {
+      return respond(200, {
+        id: SHARED_THREAD_ID,
+        title: "Long thread",
+        messages: [
+          {
+            messageIndex: 0,
+            role: "user",
+            content: "First question",
+            runIndex: 0,
+          },
+        ],
+      });
+    });
+
+    detachedSetupPage({
+      context,
+      path: `/share/threads/${SHARED_THREAD_ID}`,
+      user: null,
+    });
+
+    const scroller = await screen.findByTestId("shared-thread-scroll");
+    expect(scroller).toHaveClass("h-full", "overflow-y-auto");
+  });
 });
--- a/turbo/apps/platform/src/views/shared-thread-page/shared-thread-page.tsx
+++ b/turbo/apps/platform/src/views/shared-thread-page/shared-thread-page.tsx
@@ -125,7 +125,10 @@ export function SharedThreadPage({
   const { t } = useTranslation();
   const groups = sharedThread ? groupSharedMessages(sharedThread.messages) : [];
   return (
-    <main className="flex min-h-full flex-col bg-background text-foreground">
+    <main
+      className="flex h-full flex-col overflow-y-auto bg-background text-foreground"
+      data-testid="shared-thread-scroll"
+    >
       <header className="sticky top-0 z-20 border-b border-border/70 bg-background/95 backdrop-blur">
         <div className="mx-auto flex h-14 w-full max-w-[948px] items-center justify-between px-4 sm:px-6">
           <a


e7h4n:
## Browser verification on the PR preview

**Preview:** https://pr-25702-app.omby.ai
**Fixture:** a shared thread created on the preview (`16c73da9-fbd3-4169-ae56-efa687613433`), tall enough to overflow several viewports.

Measured on the deployed PR bundle at 1280x900:

| | `main` (before) | this PR (after) |
| --- | --- | --- |
| `<main>` classes | `flex min-h-full flex-col` | `flex h-full flex-col overflow-y-auto` |
| computed `overflow-y` | `visible` | `auto` |
| `main.clientHeight` | 3998 | 900 |
| `main.scrollHeight` | 3998 | 3998 |
| after 6 real wheel events | `scrollTop 0`, `#root.scrollTop 0`, `window.scrollY 0` — **stuck** | `scrollTop 3000` |
| scroll to end | unreachable | `scrollTop 3098` == max, end of conversation visible |
| sticky header while scrolled | n/a | stays pinned at `top: 0` |

Both states were exercised on the same page and the same data — the "before" column is the deployed PR page with the `main` classes restored via DevTools, so the only variable is the class change.

Narrow viewport (390x844) also checked: `overflow-y: auto`, `scrollHeight` 10355, scrolls to the end, and `documentElement.scrollWidth === window.innerWidth` (no horizontal overflow).

### Screenshots

Before — after 6 wheel scrolls the page has not moved: https://cdn.vm0.io/artifacts/q2us56vykp.png

After — top of the shared thread: https://cdn.vm0.io/artifacts/ui08yj73py.png

After — scrolled to the end, header still pinned: https://cdn.vm0.io/artifacts/vybp07rex3.png

After — narrow viewport, scrolled to the end: https://cdn.vm0.io/artifacts/f42pn2d1u7.png

Production repro that started this (app.okou.ai, unscrollable): https://cdn.vm0.io/artifacts/qttueslqkf.png

e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 83066d13ba792dd60366df03fc7a4546d882987d
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Makes the public shared-thread page own its own scroll container. The app shell pins `html`, `body`, and `#root` to the viewport with `overflow: hidden`, so a route rendered with the `none` page layout must supply its own scrolling; `SharedThreadPage` used `min-h-full`, which grew `<main>` past the shell instead of scrolling inside it, and `#root` clipped everything below the first viewport. The fix swaps `min-h-full` for `h-full overflow-y-auto` and adds a regression test. Two files, +33/-1, no API, schema, or contract surface touched.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic — no race conditions, bad conditionals, or wrong assumptions | Pass | `shared-thread-page.tsx:128` — `h-full` resolves against `#root`'s definite `var(--zero-viewport-height)`, so the element is exactly one viewport tall and scrolls internally. The `sticky top-0` header keeps working because it now sticks to the scrolling `<main>`. |
| `REVIEW.md` | Security — no injection, hardcoded secrets, or missing auth on new endpoints | N/A | Change is a `className` string plus a `data-testid` on an existing view; no endpoint, auth path, or user input is touched. |
| `REVIEW.md` | Type safety — no `any`, no overly broad types | Pass | No type annotations added or changed. |
| `REVIEW.md` | Style & maintainability — no oversized functions or duplicated logic | Pass | `SharedThreadPage` is unchanged in structure; the new classes match the existing full-page scroll idiom in `auth/auth-layout.tsx:363` (`h-full min-h-0 overflow-x-hidden overflow-y-auto`). |
| `REVIEW.md` | `fix:` commits must ship a regression test | Pass | `__tests__/shared-thread-page.test.tsx:54` adds `owns its scrolling because the app shell clips overflow`, which fails against the pre-fix `min-h-full` markup. |
| `REVIEW.md` | DB/JSONB review gate — persisted contract changes need migration or read-time normalization | N/A | No change under `packages/db/src/schema/**` or `packages/api-contracts/src/contracts/**`; `SharedThreadResponse` is consumed unchanged. |
| `REVIEW.md` | Feature regression guard — replaced behavior needs explicit coverage and containment | Pass | This restores the page's intended readability rather than replacing a working flow. The pre-existing DTO/markup test at `__tests__/shared-thread-page.test.tsx:12` still asserts the user/assistant grouping and the owner/agent redaction, and still passes. |
| `REVIEW.md` | New user-facing features gated behind a `FeatureSwitchKey` | N/A | Bug fix restoring intended layout behavior, not a new feature surface; gating a layout repair behind a switch would keep the page unreadable for un-flagged orgs. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | No `any`, no `eslint-disable`, `@ts-ignore`, `@ts-expect-error`, or `prettier-ignore` in the diff. |
| `docs/bad-smell.md` | No dynamic `import()`, no hardcoded URLs/config, no fallback patterns — fail fast | Pass | No imports, URLs, config reads, or `||` fallback chains added. |
| `docs/testing.md` | Integration tests only; enter through the production flow, not direct rendering | Pass | The new test uses `detachedSetupPage({ context, path: "/share/threads/…", user: null })`, mirroring `main.ts` bootstrap; no component is rendered directly. |
| `docs/testing.md` | Mock at the system boundary only | Pass | Only `context.mocks.api(sharedThreadsContract.get, …)` is used — a contract-level boundary mock. No `vi.mock()` with a relative path, no `globalThis.services.db` mock. |
| `docs/testing/patterns.md` | Prefer integration patterns; reserve pure-function unit tests for security-critical or algorithmically complex code | Pass | No unit test added for `groupSharedMessages` or `shouldMergeSharedMessage`; coverage stays at the page level. |
| `docs/testing/anti-patterns.md` | AP-2/AP-4/AP-5 — no direct `fetch` mocking, no internal mocks, no fake timers | Pass | No `vi.stubGlobal("fetch", …)`, no relative-path `vi.mock()`, no `vi.useFakeTimers()` in the diff. |
| `docs/testing/anti-patterns.md` | AP-7 — do not assert on CSS classes | Finding (P1) | `__tests__/shared-thread-page.test.tsx:78` asserts `expect(scroller).toHaveClass("h-full", "overflow-y-auto")`. See Findings; non-blocking given the identical reviewed precedent in `main`. |
| `docs/testing/anti-patterns.md` | AP-8 — no over-testing (status codes, schema validation, exact UI text) | Pass | One focused case added; the mocked 200 response is the minimum needed to render the page. |
| `docs/testing/anti-patterns.md` | AP-10 — no direct component rendering | Pass | Entry is `detachedSetupPage`, per the documented platform flow. |
| `docs/testing/app-testing.md` | Page tests live in `views/`, enter via `detachedSetupPage`, and mock through `context.mocks` | Pass | File is `views/shared-thread-page/__tests__/shared-thread-page.test.tsx`; mocks are configured through `context.mocks.api` before setup; `detachedSetupPage` is called without `await`. |
| `docs/testing/testing-external-behavior.md` | Platform tests should observe the page surface, not CSS classes or component state | Finding (P1) | Same assertion as above. jsdom has no layout engine, so scrollability itself is unobservable; the exception is real but is not stated in the test. |
| `docs/effect.md` | React render must stay pure — no commands, Store writes, DOM mutation, timers, or subscriptions during render | Pass | `SharedThreadPage` render still only reads props, calls `useTranslation`, computes `groups`, and returns elements. The diff adds only static JSX attributes. |
| `docs/cache.md` | RC-1/RC-2/RC-7 — no render-phase side effects, no render-time signal allocation, no unstable callback refs | Pass | No `state`/`computed`/`command` created in render, no `ref` callback added, no listener or observer registered. |
| `.claude/skills/ccstate/SKILL.md` | DOM async work goes through `detach(cmd(pageSignal), Reason.DomCallback)`; never `void` | N/A | The diff adds no event handler and invokes no command; `setupSharedThreadPage$` is untouched. |
| `docs/deployment-compatibility.md` | Independently deployed surfaces must tolerate cross-version skew at API, protocol, and persisted-state boundaries | Pass | Frontend-only rendering change with no request/response, runner-protocol, queue-payload, or persisted-state edit. An already-open page keeps the old clipped layout for up to ~2 days until refresh, which is the normal frontend skew window and degrades to today's behavior rather than to an incompatibility. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

- `turbo/apps/platform/src/views/shared-thread-page/__tests__/shared-thread-page.test.tsx:78` — `expect(scroller).toHaveClass("h-full", "overflow-y-auto")` asserts on CSS classes, which `docs/testing/anti-patterns.md` AP-7 lists as an anti-pattern and `docs/testing/testing-external-behavior.md` names explicitly for the platform ("Do not assert on query cache, component state, CSS classes, or whether an internal callback was called"). The test also introduces a `data-testid` that exists only for the test.

  Mitigation, and why this is not blocking: the behavior under test is genuinely unobservable at the documented boundary. jsdom has no layout engine, so `clientHeight`/`scrollHeight` are `0` and Tailwind's stylesheet is never applied — there is no page-visible surface from which scrollability can be asserted. `main` already contains the identical reviewed pattern for the identical concern: `views/auth/__tests__/auth-page.test.tsx` ("keeps app-hosted auth pages scrollable inside the root safe area") resolves `app-auth-layout` by test id and asserts `h-full`, `min-h-0`, `overflow-y-auto`, and `overflow-x-hidden`. This PR therefore follows the established convention for full-page scroll containers rather than inventing one.

  Recommended follow-up (not required for this merge): `docs/testing/testing-external-behavior.md` asks that a boundary exception "state why the production external interface cannot construct the state and why the case is still worth testing". Adding a short comment above the test recording the jsdom-layout limitation would convert this from a checklist hit into a documented exception, and would stop a future reviewer re-litigating it.

### Testing

- **Coverage: Adequate.** The `fix:` commit ships a regression test that fails against the pre-fix markup. Real scroll behavior — the part jsdom cannot cover — was verified on the PR preview and is recorded with measurements and screenshots in the earlier comment on this PR: `overflow-y` `visible` → `auto`, `clientHeight` 3998 → 900 against an unchanged `scrollHeight` of 3998, six real wheel events moving `scrollTop` from a stuck `0` to `3000`, scroll-to-end reaching `3098` == max, the sticky header holding at `top: 0`, and a 390x844 viewport scrolling to the end with no horizontal overflow.
- **Conventions: Compliant, with the one AP-7 note above.** Entry through `detachedSetupPage`, boundary mocking through `context.mocks.api`, no fake timers, no internal mocks, no direct `fetch` stubbing, no direct component rendering, no manual detached cleanup.
- **CI:** 56 checks passing, 0 failing at `83066d1`.

PR #25454, opened 2026-08-06
refactor(api): isolate typecheck boundaries

e7h4n:
## Summary

- erase the API database client's full-schema Drizzle relational type and share one schema-erased transaction callback type
- move route-owned business commands and reusable helpers into service/lib layers, leaving route aggregation bootstrap-only
- require explicit route slices for every API test app and remove all legacy `setupApp` / `createApp` defaults
- split API type checking into core, bootstrap, tests, and bootstrap-wiring Programs, with declaration boundaries and a static boundary guard

## Performance

The controlled A/B was measured at `main@560f60e96f` on the same 2 vCPU / 3934 MiB / no-swap host with Node 24.19.0, pnpm 10.33.4, TypeScript 6.0.3, a 3072 MiB old-space limit, cleared build info, and a 250 ms descendant-process RSS sampler.

| Metric | Monolithic baseline | Combined change |
| --- | ---: | ---: |
| HWM RSS range | 3367.6-3370.8 MiB | 3075.8-3142.8 MiB |
| Hot-page wall | 131.563 s | 98.749 s (-24.9%) |
| Fault-heavy wall | 140.922 s | 116.022 s (-17.7%) |
| TypeScript check time | 91.00 s | 71.77 s (-21.1%) |

The conservative HWM reduction is 224.8 MiB (-6.68%); mean HWM fell by 264.7 MiB (-7.86%). The four Programs run sequentially, so package HWM is their maximum, while package wall is their sum. Their Types, Symbols, and Instantiations overlap and are not additive.

| Program | HWM RSS | Wall | Types | Symbols | Instantiations | Check time |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| Core | 3073.3 MiB | 68.813 s | 1,092,920 | 3,360,568 | 5,855,339 | 51.14 s |
| Bootstrap | 599.2 MiB | 2.929 s | 2,389 | 85,711 | 1,955 | 0.25 s |
| Tests | 2114.0 MiB | 33.179 s | 468,633 | 1,612,682 | 1,999,399 | 19.59 s |
| Bootstrap wiring | 598.1 MiB | 3.508 s | 11,452 | 109,330 | 21,633 | 0.79 s |

The final branch is rebased onto `main@b06a0694cf`. A fresh-buildinfo snapshot there passed in 111.79 s at 3116.5 MiB HWM despite 139,507 major faults, confirming that the peak remains in the measured combined range after subsequent main changes.

## Runtime and type-safety boundaries

[...]

PR #25714, opened 2026-08-07
perf(api): isolate the aws sdk type surface behind a gateway typecheck project

e7h4n:
## Why

`apps/api check-types` peaks at ~3.25 GB RSS in CI, against a 3072 MB per-process heap cap. A CI ablation over every third-party declaration surface reachable from api's `tsconfig.core.json` (stub each library's `.d.ts` to `export {}`, measure peak RSS) ranked the non-structural SDKs:

| stubbed | Δ peak RSS | ts errors |
|---|---|---|
| stripe | −161 MiB | 35 |
| `@clerk/*` | −112 MiB | 52 |
| `@aws-sdk/*` + `@smithy/*` | −64 MiB | 30 |
| `@slack/*` | −41 MiB | 35 |
| `@sentry/*` | −24 MiB | 4 |
| all of the above at once | −374 MiB | 210 |

Declaration size does not predict cost: `@opentelemetry/*` ships 5.55 MiB of `.d.ts` and costs 0, while a single-file probe shows most of clerk's and slack's cost is the parse/bind floor. The only way to recover it is to keep the SDK out of the program, which is the same boundary technique #25454 used for core → tests.

## What

Adds `tsconfig.gateways.json`: a `composite` + `emitDeclarationOnly` project holding the leaf infra modules and the AWS-touching clients. `tsconfig.core.json` now excludes those files, references the new project, and reads its emitted `.d.ts` via `disableSourceOfProjectReferenceRedirect`, so `@aws-sdk/*` and `@smithy/*` no longer enter the core program at all.

For that to hold, no AWS type may cross the boundary, so `lib/secret-kms-client.ts` drops its command-object surface for a narrow one:

```ts
generateDataKey(request: SecretKmsGenerateDataKeyRequest): Promise<SecretKmsDataKey>;
decrypt(request: SecretKmsDecryptRequest): Promise<Uint8Array>;
```

`crypto.utils.ts` and `test-runtime-state.ts` call that instead of building `GenerateDataKeyCommand` / `DecryptCommand`, and the four test fakes implement the narrow interface (which also drops the `instanceof` command dispatch each of them carried). `signals/external/s3.ts` already had an AWS-free public surface and only had to move.

`check-typecheck-boundaries.mjs` learns about the new program, so its "every src file belongs to exactly one Program" invariant still covers the whole baseline.

## Effect

Same-harness A/B of `pnpm --filter api run check-types`, project by project, on this branch vs its merge base:

| project | main | this PR | Δ |
|---|---|---|---|
| **core** | **3252.1 MiB** | **3198.1 MiB** | **−54.0** |
| tests | 2217.5 | 2203.5 | −14.0 |
| gateways | — | 577.9 | new |
| bootstrap | 433.5 | 435.9 | +2.4 |
| bootstrap-wiring | 456.2 | 458.9 | +2.7 |

CI reports the max over the chain, so the headline number moves from 3252 to 3198 MiB. The new project adds 4.6 s and core gets 2.3 s faster, so wall clock is a wash. Noise on this harness is ±5 MiB.

## Follow-ups

The mechanism generalizes; each remaining library needs its own signature narrowing first:

- stripe (−161): `getStripeClient(): StripeSDK` plus `Stripe.*` types in six billing services
- clerk (−112): `ReturnType<typeof createClerkClient>` and the `Awaited<ReturnType<Client["users"][...]>>` chains, 73 `clerk$` call sites
- slack (−41): `WebClient` / `Block` / `KnownBlock`, 61 references
- unrelated but cheap: dropping `DOM` / `DOM.Iterable` from api's `lib` is worth −44 MiB and breaks only 4 lines (`BodyInit`, `HeadersInit`, one read-only `statusText`)

## Keeping it

The boundary is only worth the structure if it cannot be undone by accident, so `api/gateway-typecheck-boundary` (new rule in `@vm0/eslint-rules`) makes each of the three ways to break it an error:

| error | what it catches |
|---|---|
| `outboundImport` | a gateway module importing outside the project, which drags the importee in with it |
| `isolatedDependency` | any other module importing `@aws-sdk/*` / `@smithy/*` directly, which puts the SDK back into the core program |
| `exportedSdkType` | a gateway module naming an SDK type in an exported signature, which puts the SDK into the emitted `.d.ts` and therefore back into every consumer |

`eslint.config.mjs` reads the member list out of `tsconfig.gateways.json`, so the project file stays the single source of truth (and globs in `include` are rejected, since they would silently under-enforce). Tests are exempt: they type-check in their own smaller program. Extending the boundary to stripe/clerk/slack later is one entry in `isolatedDependencies` plus the modules moving into the project.

`exportedSdkType` only counts annotations that survive declaration emit, so an SDK type on a local inside an exported function body is allowed; that case has its own test.

## Fallbacks

- `turbo/apps/api/src/lib/secret-kms-client.ts:generateDataKey` — `response.KeyId ?? request.keyId`. Protects the AWS SDK's optional `GenerateDataKeyResponse.KeyId` (`KeyId?: string`), so it is the `docs/fallback.md` §6.2 "genuinely optional external data" case, not a cross-version rollout fallback: no deploy window applies and there is no removal condition while the SDK keeps the field optional. Pre-existing behavior, relocated verbatim out of `crypto.utils.ts` when the KMS surface was narrowed.

No rollout or compatibility fallback is introduced or removed. The persisted stored-secret envelope (`vm0secret:v1:` + `keyId` / `encryptedDataKey` / `iv` / `authTag` / `ciphertext`) is byte-for-byte unchanged, and the KMS request parameters sent to AWS (`KeySpec: "AES_256"`, `EncryptionContext`) are identical, so no old/new version pairing is affected.


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 49c316ecfeaad4e5c6461fa225faa9a600ef27e0
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Adds a fourth apps/api type-check program (`tsconfig.gateways.json`, `composite` + `emitDeclarationOnly`) holding the leaf infra modules and the AWS-touching clients, so `tsconfig.core.json` reads their emitted `.d.ts` instead of resolving `@aws-sdk/*` and `@smithy/*` itself. To make that hold, `lib/secret-kms-client.ts` drops its command-object surface for a narrow `generateDataKey` / `decrypt` interface, and the two production callers plus four test fakes follow. A new `api/gateway-typecheck-boundary` ESLint rule makes the three ways to break the boundary an error, with the member list read out of `tsconfig.gateways.json` so the project file stays the single source of truth.

One known, accepted limitation of the new rule: `exportedSdkType` is syntactic, so it catches an SDK type named directly in an exported signature but not one laundered through an unexported local alias (`type A = KMSClient; export type B = A;`). The measured CI peak RSS remains the backstop for that case.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no wrong assumptions or unhandled paths | Pass | `secret-kms-client.ts` keeps both invariant throws (`Plaintext`, `CiphertextBlob`) and the `DATA_KEY_BYTE_LENGTH` check stays with the caller in `crypto.utils.ts`; `KeySpec: "AES_256"` moves inside the gateway and matches the only caller's expectation. |
| `REVIEW.md` | Security: no hardcoded secrets, no weakened auth | Pass | KMS key id still comes from `env("SECRETS_KMS_KEY_ID")` via `requireSecretsKmsKeyId()`; encryption context is passed through unchanged. |
| `REVIEW.md` | Type safety: no `any`, no over-broad types | Pass | New interfaces are concrete (`Uint8Array`, `Readonly<Record<string, string>>`); no `any`, no `as` bridge. The rule file uses `TSESTree` types throughout. |
| `REVIEW.md` | Refactor commits keep existing coverage | Pass | The KMS path stays covered through API endpoints by `chat-callbacks.bdd`, `run-lifecycle.bdd`, `webhooks-agent-firewall-auth.bdd`, `zero-goals`, and `zero-workflow-queue`; only the fake implementations changed, and the probe's call counters are preserved. |
| `REVIEW.md` | Feature regression guard | Pass | No user-visible behavior changes: same AWS request parameters, same error messages, same persisted envelope. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No `packages/db/src/schema/**`, contract, or `jsonb().$type<>()` change; the stored-secret envelope shape is untouched. |
| `docs/bad-smell.md` | Zero lint/type suppressions | Pass | No `eslint-disable`, `@ts-ignore`, or `@ts-expect-error` added; the boundary is enforced by a new rule rather than suppressed. |
| `docs/bad-smell.md` | Fail fast, no configuration fallbacks | Pass | Missing `Plaintext` / `CiphertextBlob` throw precise errors instead of substituting a value. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | All imports are static; `eslint.config.mjs` reads `tsconfig.gateways.json` with `fs` at config load, not via dynamic import. |
| `docs/bad-smell.md` | Interface changes documented | Pass | `SecretKmsClient` is an internal api interface; every implementation (production + 4 fakes) is updated in the same PR. |
| `docs/fallback.md` | No fallback for a state the contract cannot produce | Pass | The single `??` guards `GenerateDataKeyResponse.KeyId`, which the AWS SDK type declares optional (§6.2 genuinely optional external data). |
| `docs/fallback.md` | Every fallback declared in the PR summary | Pass | PR summary carries a `Fallbacks` section naming the file, symbol, category, and why no rollout window applies. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | No tombstone tests added; the changed tests exercise the new interface directly. |
| `docs/deployment-compatibility.md` | Cross-version compatibility for deployable/persisted boundaries | N/A | No frontend/backend contract, runner protocol, queue payload, schema, or persisted shape changes; the change is confined to in-process types and tsc project layout. |
| `docs/testing.md` | Integration tests only, no unit tests for internal functions | Pass | api-side coverage stays at the API-endpoint level. The RuleTester file for the new ESLint rule follows the established `packages/eslint-rules` convention (13 sibling rule test files). |
| `docs/testing.md` | Mock at the boundary; `vi.mock()` must not target internal paths | Pass | No new `vi.mock()`; the KMS fakes go through the pre-existing `setSecretKmsClientForTests` override at the AWS boundary. |
| `docs/testing.md` | MSW for HTTP, real DB, no fake timers, no `toHaveBeenCalled()`-only assertions | Pass | The diff adds no `vi.stubGlobal("fetch")`, no `useFakeTimers`, no db mocking, and no call-count-only assertion. |
| `docs/testing/api-testing.md` | Route behavior exercised through `setupApp()` and contracts, not internals | Pass | No route test reaches into services or DB; the touched tests only swap an external-boundary fake. |
| `docs/testing/anti-patterns.md` | AP-1/AP-2/AP-3/AP-4 (mock-call assertions, fetch mocking, fs mocking, internal mocking) | Pass | None present in the diff; the four rewritten fakes got simpler by dropping `instanceof` command dispatch. |
| `.claude/skills/ccstate/SKILL.md` | ccstate signal/command discipline in apps/api | N/A | No `computed`/`command`/Store code is added or changed; `signals/external/s3.ts` moves between tsc projects with its ccstate usage untouched. |

### Findings

#### Critical (P0)
- None.

#### High Priority (P1)
- None.

### Fallbacks

- `turbo/apps/api/src/lib/secret-kms-client.ts:generateDataKey` — `response.KeyId ?? request.keyId`, protecting the AWS SDK's optional `GenerateDataKeyResponse.KeyId` (`KeyId?: string`). Surface: none — this is `docs/fallback.md` §6.2 genuinely optional third-party data, not a cross-version rollout branch, so no deploy window applies. Removal condition: none evidenced while the SDK keeps the field optional. Declared in PR summary: Yes. Verdict: Justified. Pre-existing behavior relocated verbatim from `crypto.utils.ts`.

No rollout or compatibility fallback is introduced or removed by this PR.

### Testing

- Coverage: Adequate. This is a `perf:`/`refactor:` change with no behavior delta, and the KMS encrypt/decrypt path keeps its existing API-endpoint coverage; the four fakes were migrated rather than deleted, including the firewall-auth test's ordering gate on the second `decrypt` call. The new ESLint rule ships 14 RuleTester cases covering all three error classes plus the declaration-emit boundary case (an SDK type on a local inside an exported function body must not be flagged).
- Conventions: Compliant. No `vi.mock()` of internal modules, no fetch stubbing, no fake timers, no db mocking, no call-count-only assertions.

PR #25719, opened 2026-08-07
feat: soften the dialog scrim and blur the backdrop

tongx-blip:
## Why

The shared `DialogOverlay` in `packages/ui` never moved off the shadcn/ui default:

```
bg-black/80 dark:bg-background/50
```

Light mode dims the page with **80%** of the design system's ink, while dark mode uses only 50% of the near-black `--background` (`#19191B`). So the light theme reads noticeably heavier than the dark one — the opposite of the usual relationship — and 40 `<Dialog>` call sites in `apps/platform` all inherit it.

The blur was never a deliberate omission either. Three surfaces already hand-rolled it:

- `views/onboarding/onboarding-shell.tsx` — `overlayClassName="bg-black/45 backdrop-blur-sm dark:bg-black/45"`
- `views/clerk/clerk-appearance.ts` — `backdropFilter: "blur(2px)"`
- `views/zero-page/browser-session-panel.tsx` — `bg-background/50 backdrop-blur-md`

This promotes that treatment into the shared primitive.

## What

`packages/ui/src/components/ui/dialog.tsx`:

```diff
-"zero-dialog-overlay fixed inset-0 z-50 bg-black/80 dark:bg-background/50"
+"zero-dialog-overlay fixed inset-0 z-50 bg-overlay/45 backdrop-blur-sm dark:bg-overlay/55"
```

- Light drops to **45%**, matching the value onboarding already shipped.
- Dark settles on **55%**. The dark card (`240 4% 14%`) sits close in tone to the dark page background, so it needs a slightly stronger scrim than light mode to separate from it.
- `backdrop-blur-sm` is **8px** on Tailwind v4 (`tailwindcss@^4.3.3`; v4 renamed the v3 4px `sm` step to `xs`). Verified against the deployed preview: the overlay computes to `backdrop-filter: blur(8px)`.
- `onboarding-shell.tsx` drops its now-redundant `overlayClassName` and inherits the shared value.

### New `--overlay` token

`bg-black` resolves to `--color-black` → `--black`, which this design system **inverts per theme**: `240 3% 10%` (`#19191B`) in light, `0 0% 100%` in dark. A `dark:bg-black/55` scrim therefore painted a 55% *white* wash that lifted the backdrop instead of dimming it — caught on the deployed preview, not by tests.

`globals.css` gains a scrim token that deliberately does not invert:

```css
--color-overlay: hsl(var(--overlay));
/* :root only — the dark theme reuses it and only raises the alpha */
--overlay: 240 3% 10%; /* #19191B */
```

Light is unchanged by this (there `--black` already resolved to `#19191B`); dark now dims with the same ink.

## Browser support

`backdrop-filter` is Baseline widely available — Chrome 76+, Edge 79+, Firefox 103+, Safari 9+ with `-webkit-`. Windows Chrome/Edge/Firefox all support it, and the repo already ships `backdrop-blur-*` in several places, so this adds no new compatibility surface.

## Performance notes

A full-viewport `backdrop-filter` makes the compositor re-blur the backdrop while the overlay animates. Two things keep that bounded:

- The overlay's entrance is a 180ms opacity fade (`globals.css`), and `prefers-reduced-motion: reduce` already sets `animation: none` on it, so reduced-motion users get a static blur with no per-frame recomposite. No extra media query is needed — blur is not motion, and gating it on `prefers-reduced-motion` would be the wrong signal.
- 8px stays well below a large-radius blur, which matters for the dialogs whose backdrop contains a live iframe (template preview, artifact preview, browser session).

## Not in scope

`SheetOverlay` renders `fixed inset-0 z-50` with no background at all, so sheets currently have no scrim. That is a separate decision and is left untouched.

## Verification

Visual check on the deployed preview (`https://pr-25719-app.omby.ai`, Settings dialog + template picker, light and dark):

- light — backdrop dimmed and blurred, markedly lighter than the old 80% wash
- dark — backdrop dimmed, not lifted, after the `--overlay` fix

Local:

- `pnpm format` clean, `git diff --check` clean
- `pnpm turbo run lint --filter=@vm0/ui --filter=@vm0/app --concurrency=1` — 6/6 pass, 0 warnings 0 errors
- `pnpm check-types` — 12/13 pass. `@vm0/pi-agent-runtime#check-types` fails on `Cannot find module 'msw'` in `src/agent-loop.test.ts`; reproduced on a clean stashed tree, so it is a pre-existing local dependency issue unrelated to this diff.
- `pnpm -F @vm0/ui exec vitest run src/components/ui/__tests__/dialog.test.tsx` — 1 passed
- `pnpm -F @vm0/app exec vitest run` on `onboarding-flow`, `chat-composer-template-picker`, `zero-sidebar-account-menu` (the three suites that touch `.zero-dialog-overlay`) — 64 passed
- `pnpm knip --workspace packages/ui` and `--workspace apps/platform` — clean

No test asserts the overlay's colour or blur classes; per `docs/testing/testing-external-behavior.md` platform tests should not assert on CSS classes, so this visual change adds none. The inverted-token bug it would have taken to catch this is a colour-resolution issue that only the rendered page shows.

### DIFF
--- a/turbo/apps/platform/src/views/onboarding/onboarding-shell.tsx
+++ b/turbo/apps/platform/src/views/onboarding/onboarding-shell.tsx
@@ -137,10 +137,7 @@ export function OnboardingDialog({
         }
       }}
     >
-      <DialogContent
-        overlayClassName="bg-black/45 backdrop-blur-sm dark:bg-black/45"
-        className="flex max-h-[calc(100dvh-32px)] w-[calc(100%-32px)] max-w-4xl flex-col gap-0 overflow-hidden rounded-lg border-border bg-background p-5 sm:p-6"
-      >
+      <DialogContent className="flex max-h-[calc(100dvh-32px)] w-[calc(100%-32px)] max-w-4xl flex-col gap-0 overflow-hidden rounded-lg border-border bg-background p-5 sm:p-6">
         <header className="shrink-0 pr-9">
           <div className="min-w-0">
             <DialogTitle className="text-lg font-semibold leading-6">
--- a/turbo/packages/ui/src/components/ui/dialog.tsx
+++ b/turbo/packages/ui/src/components/ui/dialog.tsx
@@ -20,7 +20,7 @@ const DialogOverlay = React.forwardRef<
     <DialogPrimitive.Overlay
       ref={ref}
       className={cn(
-        "zero-dialog-overlay fixed inset-0 z-50 bg-black/80 dark:bg-background/50",
+        "zero-dialog-overlay fixed inset-0 z-50 bg-black/45 backdrop-blur-sm dark:bg-black/55",
         className,
       )}
       {...props}
--- a/turbo/packages/ui/src/components/ui/dialog.tsx
+++ b/turbo/packages/ui/src/components/ui/dialog.tsx
@@ -20,7 +20,7 @@ const DialogOverlay = React.forwardRef<
     <DialogPrimitive.Overlay
       ref={ref}
       className={cn(
-        "zero-dialog-overlay fixed inset-0 z-50 bg-black/45 backdrop-blur-sm dark:bg-black/55",
+        "zero-dialog-overlay fixed inset-0 z-50 bg-overlay/45 backdrop-blur-sm dark:bg-overlay/55",
         className,
       )}
       {...props}
--- a/turbo/packages/ui/src/styles/globals.css
+++ b/turbo/packages/ui/src/styles/globals.css
@@ -64,6 +64,7 @@
    * =================================== */
   --color-white: hsl(var(--white));
   --color-black: hsl(var(--black));
+  --color-overlay: hsl(var(--overlay));
 
   /* ===================================
    * Semantic Colors
@@ -176,6 +177,11 @@
     /* Theme-aware Colors */
     --white: 0 0% 100%; /* #FFFFFF */
     --black: 240 3% 10%; /* #19191B */
+
+    /* Dialog scrim. Deliberately NOT theme-aware: --black inverts to #FFFFFF
+       in dark, which would lighten the backdrop instead of dimming it. The
+       dark theme reuses this value and only raises the alpha. */
+    --overlay: 240 3% 10%; /* #19191B */
 
     /* ===================================
      * Semantic Colors (Light Theme)


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: 53b739546e0978364f6187470cc828b9e247ab60
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Moves the shared `DialogOverlay` off the shadcn default `bg-black/80 dark:bg-background/50` to `bg-overlay/45 backdrop-blur-sm dark:bg-overlay/55`, introduces a non-inverting `--overlay` scrim token, and deletes onboarding's now-redundant `overlayClassName`. Net +7/−1 across three files; affects all 40 `<Dialog>` call sites in `apps/platform`.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic — wrong conditionals, unhandled null paths | Pass | The only logic surface is CSS cascade. Verified in the deployed bundle (`/assets/main-DusNTMWz.css`): `--overlay:240 3% 10%` occurs exactly once, inside `:root`, and the `.dark, [data-theme="dark"]` block does not override it, so dark inherits the same ink. |
| `REVIEW.md` | Security — injection, secrets, missing auth | N/A | No endpoint, auth, credential, or user-input path is touched. |
| `REVIEW.md` | Type safety — no `any`, no broad types | Pass | No TypeScript types change; `dialog.tsx` edits one class-name string literal. |
| `REVIEW.md` | Style & maintainability — duplication, oversized functions, WHY-not-WHAT comments | Pass | Removes a duplicated overlay treatment from `onboarding-shell.tsx` rather than adding one. The new CSS comment states *why* the token must not invert, not what it is. |
| `REVIEW.md` | Testing coverage — `feat:` commits must have integration tests | **Finding (P1)** | No test added. See P1 below: the only assertable artifact is a CSS class, which `docs/testing/testing-external-behavior.md` forbids platform tests from asserting. |
| `REVIEW.md` | Feature regression guard — changed existing UI needs coverage for the affected path | Pass (with the P1 caveat) | The three suites that touch `.zero-dialog-overlay` (`onboarding-flow`, `chat-composer-template-picker`, `zero-sidebar-account-menu`) still pass, so overlay rendering and dismissal are unregressed; they assert presence and animation classes, not colour. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, contract, or `jsonb().$type<>()` change. |
| `docs/bad-smell.md` | Zero lint/type suppressions | Pass | No `eslint-disable`, `@ts-ignore`, `@ts-expect-error`, or `prettier-ignore` in the diff. |
| `docs/bad-smell.md` | Fail fast — no fallback/recovery layers | Pass | No fallback added. The `--overlay` token is a single definition with no cascade fallback chain. |
| `docs/bad-smell.md` | No hardcoded config; use centralized values | Pass | The scrim colour becomes a design token instead of an inline literal, which moves in the right direction; `dialog.tsx` no longer names a raw colour. |
| `docs/bad-smell.md` | Interface changes documented | Pass | `--overlay` / `--color-overlay` are new public design tokens; the inline comment documents their contract. No exported TS symbol changes. |
| `docs/testing.md` | Integration tests only, mock at the boundary, MSW, real DB, no fake timers | N/A | No test file is added or modified by this PR. |
| `docs/testing.md` | New user-facing features gated behind a `FeatureSwitchKey` | **Finding (P1)** | Ungated. See P1 below. |
| `docs/testing/testing-external-behavior.md` | Platform tests must not assert on CSS classes | Pass | The PR adds no class assertions, which is why the coverage gap above cannot be closed the obvious way. |
| `docs/effect.md` | Render purity — no Store writes, I/O, or resource creation in render | Pass | `DialogOverlay` remains a pure forwardRef returning a Radix primitive; `OnboardingDialog` only drops a prop. |
| `docs/effect.md` | Correct execution boundary (computed / command / setup / onRef / effect) | N/A | No effect, command, ref, or signal is added or moved. |
| `docs/cache.md` | RC-1/RC-2 — no render-phase side effects or render-time signal allocation | Pass | Nothing allocated; the diff is class names and CSS variables. |
| `docs/cache.md` | RC-8 — symmetric resource lifecycles | N/A | No listener, observer, timer, or object URL involved. |
| `.claude/skills/ccstate/SKILL.md` | ccstate command/computed usage, no signal-layer `detach()` | N/A | No signals-layer code changes. |
| `docs/deployment-compatibility.md` | Old frontend → new backend, new frontend → old backend | N/A | Browser-only presentation change; no API request/response, runner protocol, queue payload, or persisted state crosses a version boundary. An open pre-change tab simply keeps the old scrim until it refreshes. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

- `packages/ui/src/components/ui/dialog.tsx` — **`feat:` PR with no integration test, and no feature switch.** `REVIEW.md` grades missing tests on a `feat:` as P0 and `docs/testing.md` expects new user-facing features behind a `FeatureSwitchKey`. Both are unmet here, and I am deliberately grading this P1 rather than P0 because the two rules collide with a third: the change's entire observable surface is the overlay's `background-color` and `backdrop-filter`, and `docs/testing/testing-external-behavior.md` states platform tests must not assert on CSS classes. jsdom does not resolve Tailwind utilities or `color-mix()`, so no non-class assertion is available either. Flagging it explicitly so a human can overrule: if the team wants this gated, the switch would have to wrap a shared UI primitive used by 40 dialogs, which is a heavier containment than the change warrants. Compensating evidence is in the PR body — the scrim was verified on the deployed preview in both themes, and the built CSS was inspected directly to confirm the dark path.

#### Notes (non-blocking)

- The inverted-token defect this PR fixes in its second commit (`dark:bg-black/55` rendering a 55% *white* wash, because `--black` flips to `#FFFFFF` in dark) is exactly the class of bug the missing test would not have caught — it is a colour-resolution issue visible only in a rendered page. The `--overlay` token plus its comment is the durable guard, and it is worth knowing that `bg-black` / `text-black` are unsafe anywhere a colour must stay dark in both themes.
- `SheetOverlay` in `packages/ui/src/components/ui/sheet.tsx` is `fixed inset-0 z-50` with no background at all, so sheets still have no scrim. Correctly called out as out of scope in the PR body; worth a follow-up decision.

### Testing

- Coverage: Insufficient by the letter of `REVIEW.md` for a `feat:` commit — see P1. No behavioural regression risk was left unexercised, though: the three existing suites that render `.zero-dialog-overlay` all pass, covering overlay mount, animation-class pass-through via `overlayClassName`, and dialog dismissal.
- Conventions: Compliant. No mocks, fake timers, internal-path `vi.mock()`, direct `fetch` stubbing, or CSS-class assertions are introduced.
- Reported in the PR body and consistent with the diff: `pnpm turbo run lint --filter=@vm0/ui --filter=@vm0/app --concurrency=1` 6/6 pass; `pnpm -F @vm0/ui exec vitest run src/components/ui/__tests__/dialog.test.tsx` 1 passed; `pnpm -F @vm0/app exec vitest run` on the three overlay suites 64 passed; `pnpm knip` clean on both workspaces. `pnpm check-types` is 12/13 with `@vm0/pi-agent-runtime` failing on a missing `msw` module, reproduced on a clean tree and unrelated to this diff — CI `lint-types` and `lint-types-apps` are both green on this commit, which corroborates that.
- Independent verification for this review: fetched `https://pr-25719-app.omby.ai/assets/main-DusNTMWz.css` and confirmed `.bg-overlay\/45` and `.dark\:bg-overlay\/55:where([data-theme=dark],...)` both compile to `color-mix(... hsl(var(--overlay)) ...)`, with `--overlay` defined once in `:root`. Dark therefore dims with `#19191B` at 55% rather than lightening.

PR #10418, opened 2026-04-21
perf(eslint): reduce type-checking overhead in @vm0/app ccstate rules

e7h4n:
## Summary

- **RSS: 2360 MB → 1857 MB (-21%)**, wall time: 28s → 19s (-32%) for the \`@vm0/app\` ESLint phase
- Rewrites 5 ccstate rules (\`command-async-signal\`, \`no-getter-setter-params\`, \`no-store-in-params\`, \`no-get-signal\`, \`no-package-variable\`) from TypeScript type-checker API to pure AST type-annotation inspection — removes \`requiresTypeChecking: true\` and moves them out of the \`projectService: true\` config block
- Optimizes \`computed-const-args-package-scope\` (was 56% of lint time): defers \`getTypeAtLocation\` to \`Program:exit\` and applies the name filter (\`computed\` / \`packageScopeConstantFunctions\`) first, eliminating type-checker calls for >95% of \`CallExpression\` nodes; \`computed()\` calls are now checked with zero type-checker invocations

## Breaking Config Change

\`no-package-variable\` option renamed: \`allowedMutableTypes: TypeOrValueSpecifier[]\` → \`allowedConstructors: string[]\`

The old option required type-checker resolution to match types across files. The new option matches constructor call sites by name using AST only, which is sufficient for the in-repo use cases. The in-repo \`eslint.config.js\` has been updated accordingly:

```js
// Before
"ccstate/no-package-variable": ["error", { allowedMutableTypes: [...] }]

// After
"ccstate/no-package-variable": ["error", { allowedConstructors: ["LocationOverrides", "PromiseTracker", "LoggerRegistry"] }]
```

Any downstream config still using \`allowedMutableTypes\` will silently receive no allowance (option is ignored). Update to \`allowedConstructors\` with the constructor class names.

## Known Limitation: no-get-signal cross-file detection

The \`no-get-signal\` rewrite tracks signal variable names within the current file only. Signals imported from another module (e.g. \`import { signal$ } from './other'\`) will not be detected. This is an intentional trade-off: the performance gain (~32% faster lint) outweighs the false-negative surface, since \`AbortSignal\` states are almost always defined co-located with their use. The limitation is documented in the rule source.

## How it was measured

Added \`scripts/measure-eslint-memory.sh\` (peak RSS via \`/proc\` polling) and \`scripts/benchmark-eslint-rules.sh\` (ablation benchmark, 3 runs per experiment with avg/stddev). Results from single-run validation:

| Experiment | RSS | Time |
[...]

PR #25720, opened 2026-08-07
refactor(platform): enforce ccstate/no-catch-abort, drop dead lint configs

e7h4n:
## What

Two independent cleanups that both land on the same lint surface.

**Observable behavior is unchanged.** One internal control-flow detail does change: `openMedia` now lets a cancellation propagate instead of swallowing it. See *Behavior delta* below.

### 1. Enable `ccstate/no-catch-abort` in platform

`ccstate/no-catch-abort` was registered in the ccstate plugin but no config turned it on, so `throwIfAbort` was enforced in `apps/api` only. It is now `error` for platform production source.

Scope and ignore list mirror the try-statement ban directly above it in the same config:

- test and mock code carries no abort contract;
- `src/signals/utils.ts` implements the centralized helpers (`onRejection`, `settle`, `tapError`) whose whole purpose is to observe a rejection, including an abort, before re-throwing it. That file already carries the same exemption from the try-statement ban.

Six catch clauses were flagged. All six are resolved without a single lint suppression, and each resolution is behavior-preserving:

| Site | Change |
|---|---|
| `signals/external/idb-event-store.ts` | try/catch replaced by `onRejection(getChatIdb(), disableForSession)`, the centralized helper for this exact tap-and-rethrow shape. Also drops an existing `no-restricted-syntax` suppression. |
| `signals/voice-io/voice-io-stt.ts` | `openMedia`'s try/catch replaced by `settle(p, signal)`. Also drops an existing `no-restricted-syntax` suppression. |
| `views/zero-page/zero-attachment-url.ts` | `signal.throwIfAborted()` replaced by `throwIfAbort(error)`. `fetch` receives the signal, so both checks cover the same reachable state; keeping both was redundant. |
| `signals/realtime.ts` (2 sites) | Reordered so `throwIfAbort(error)` comes before `signal.throwIfAborted()`, matching the idiom already used in `signals/utils.ts` and `signals/external/chat-idb-safe.ts`. Both orderings propagate an abort. |
| `signals/utils.ts` `onRejection` | Covered by the file-level config exclusion above. |

Net effect on suppressions: **two `eslint-disable` comments removed, none added.**

### 2. Delete five dead config files

**Next.js leftovers.** No workspace has depended on Next.js since the web app was removed. Nothing imported `eslint-config/next.js` and nothing extended `typescript-config/nextjs.json`.

- `turbo/packages/eslint-config/next.js`
- `turbo/packages/typescript-config/nextjs.json`
- the `./next-js` export and the `@next/eslint-plugin-next` devDependency
- the stale `ignoreUnresolved: ["next"]` and `.next/**` entries in `knip.json` (knip itself reports the latter as unused once the presets are gone)

**One-off perf harness from #10418.** Nothing in CI, package scripts or docs referenced these, and both experiments had gone stale:

- `turbo/apps/platform/eslint.config.ablation.mjs` and its ignore entry
- `scripts/benchmark-eslint-rules.sh`
- `scripts/measure-eslint-memory.sh`

Phase 4 strips a `projectService` that no config sets any more, so the branch never fires. Phase 3 ablates six `TYPE_AWARE_RULES` that all carry `requiresTypeChecking: false` today, so it measures noise.

## Behavior delta

`signals/voice-io/voice-io-stt.ts` `openMedia`: when the page signal aborts after `getUserMedia` resolves, the old code caught its own `signal.throwIfAborted()`, toasted, and returned `undefined`; `settle(p, signal)` re-throws the `AbortError` instead. The caller is `tapError(openMedia(signal), reportMicStartFailure)`, so the same toast still fires and the recording still does not start, but the attempt now unwinds through the abort path rather than through `return null`. The page is being torn down in that scenario either way.

Every other flagged site is behavior-identical: `onRejection` performs the same tap-and-rethrow as the try/catch it replaces, the `realtime.ts` reorder propagates an abort in both orderings, and in `zero-attachment-url.ts` the `fetch` call receives the signal so `throwIfAbort(error)` and `signal.throwIfAborted()` fire on the same reachable state.

## Fallbacks

**Fallbacks: none.**

The diff introduces, keeps, and removes no cross-version compatibility branch, no `??`/`||` default on a contract-guaranteed field, no feature-switch-gated tolerance path, and no retired-protocol reader.

Two guards were considered and deliberately **not** shipped, because each would have been a fallback for a state the contract cannot produce (`docs/fallback.md` §3):

- `idb-event-store.ts`: a `throwIfAbort` on the IDB open. `getChatIdb()` takes no `AbortSignal`, so no caller cancellation can reach it, and a genuine IndexedDB `AbortError` should keep disabling the cache for the tab rather than being re-read as cancellation.
- `voice-io-stt.ts`: a `throwIfAbort` in `reportMicStartFailure`. A page test that starts voice input and then navigates away passes with and without it, so the state it would guard is not demonstrably reachable.

## Verification

- `pnpm format`, `NODE_OPTIONS=--max-old-space-size=3072 pnpm turbo run lint --concurrency=1 --log-order grouped` (13/13), `pnpm check-types` (14/14) and `pnpm knip` all pass locally.
- `pnpm -F @vm0/app exec vitest run` on the three affected platform test files (`chat-computer-use-and-voice`, `zero-chat-thread-idb`, `zero-attachment-chips`): 78 passed.
- Lockfile regenerated with `pnpm install --lockfile-only`; the only change is dropping `@next/eslint-plugin-next`.
- Confirmed by grep that no file in the repo references any deleted path, export, or dependency.
- Enumerated all 20 catch clauses in the platform package and checked each against the rule; the six above are the complete set in the enforced scope.

## Testing

No regression test is added because no behavior changes. The three "cancellation bug fixes" claimed in the first revision of this PR did not survive review: two guarded unreachable states and one could not be shown to change any observable outcome. The remaining diff is a lint enablement plus behavior-preserving refactors onto existing centralized helpers, covered by the existing suites listed above.


e7h4n:
Changes Requested


PR_AUTO_REVIEW_COMMIT: edea128398ee0f24c16d9b2325488d64219b3a8d

### Summary

Enables `ccstate/no-catch-abort` for platform production source and brings the six catch clauses it flags into compliance, three of which were real cancellation bugs (chat IDB cache disabled by an aborted open, spurious `download failed` and `microphone access denied` toasts on cancel). Also deletes five dead config files: the two Next.js presets left over after the web app was removed, and the one-off eslint perf-ablation harness from #10418, along with the `./next-js` export, the `@next/eslint-plugin-next` devDependency and the stale `ignoreUnresolved: ["next"]` knip entry.

The cleanup half is well evidenced and correctly scoped. The blocker is on the `fix:` half: three user-visible behavior changes ship with no test at all.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | `fix:` commits must carry a regression test | Finding (P0) | `idb-event-store.ts`, `zero-attachment-url.ts`, `voice-io-stt.ts` change cancellation behavior; no test file is in the diff |
| `REVIEW.md` | Correctness and logic, no unhandled edge cases | Pass | The 6 flagged catch clauses were enumerated against the rule; `throwIfAbort(error)` argument matches each catch binding |
| `REVIEW.md` | Security (injection, secrets, auth) | N/A | No auth, network input parsing, or credential handling in the diff |
| `REVIEW.md` | Type safety, no unjustified `any` | Pass | No casts or new types; `throwIfAbort(e: unknown)` used as declared |
| `REVIEW.md` | Style and maintainability, no duplicated logic | Pass | Reorderings converge on the existing `throwIfAbort` first idiom in `utils.ts` / `chat-idb-safe.ts` |
| `REVIEW.md` | Feature regression guard for replaced behavior | Finding (P0) | `openMedia` and `reportMicStartFailure` stop toasting on abort; the existing mic-permission-denied flow has no covering test |
| `REVIEW.md` | DB/JSONB persisted-contract gate | N/A | No `schema/**`, contract, or `jsonb().$type<>()` change |
| `docs/bad-smell.md` §2 | Zero tolerance for lint suppression comments | Finding (P1) | `src/signals/utils.ts:242` adds `eslint-disable-next-line ccstate/no-catch-abort` |
| `docs/bad-smell.md` §3 | No unnecessary or over-engineered try/catch | Pass | No try/catch added; the diff only tightens existing catch bodies |
| `docs/bad-smell.md` §7 | Fail fast, no defaulting around a guaranteed contract | Pass | `throwIfAbort` re-throws rather than defaulting |
| `docs/fallback.md` §6 | Fallbacks only for the four permitted categories | Pass | No cross-version, optional-external-field, nullable-column, or presentational fallback in the diff |
| `docs/fallback.md` §9 | PR summary must contain `Fallbacks` or explicit `Fallbacks: none` | Finding (P1) | PR body has What / Verification / Notes but no `Fallbacks` section |
| `docs/fallback.md` §10 | Removing a fallback requires stated evidence | Pass | Deletions are dead config, not fallbacks; PR states the grep and no-`next`-dependency evidence |
| `docs/testing.md` | Integration tests only, no unit tests for internals | N/A | No test file changed |
| `docs/testing.md` | Mock at the system boundary only | N/A | No test file changed |
| `docs/testing/app-testing.md` | Platform behavior is exercised through page tests with MSW handlers | Finding (P0) | The download-cancel and mic-cancel paths are reachable from a page test but none was added |
| `docs/testing/testing-external-behavior.md` | Assert on the user-visible surface, e.g. the toast | Finding (P0) | The two toast suppressions are exactly page-observable and are unasserted |
| `docs/testing/patterns.md` | Follow the documented test construction patterns | N/A | No test file changed |
| `docs/testing/anti-patterns.md` | Avoid the catalogued test anti-patterns | N/A | No test file changed |
| `docs/effect.md` | Make cancellation explicit, every async command has an owning signal | Pass | No signal ownership changed; callers still pass their existing owner signal |
| `docs/cache.md` RC-9 | Lint coverage gaps are findings; do not weaken or suppress the rule | Finding (P1) | Same suppression as above; RC-9 states the rule should not be suppressed |
| `docs/cache.md` RC-8 | Symmetric resource lifecycles | Pass | `disableForSession` is now skipped on abort, which restores symmetry rather than breaking it |
| `.claude/skills/ccstate/SKILL.md` | Use `signal.throwIfAborted()` only after an await that does not receive the signal | Pass (observation) | In `zero-attachment-url.ts` and both `realtime.ts` sites the awaited call already receives the signal, so the retained `signal.throwIfAborted()` is now redundant; pre-existing, not introduced here |
| `.claude/skills/ccstate/SKILL.md` | Every AbortSignal has a clear owner | Pass | No new `AbortController`; existing owners unchanged |

### Findings

- **No regression test for a `fix:` commit.** `REVIEW.md` requires a regression test on `fix:` commits. This PR changes three user-visible behaviors and adds no test:
  - `src/signals/external/idb-event-store.ts:148` — an aborted IDB open no longer calls `disableForSession`. Nothing pins that the chat IDB cache survives a cancelled open, so a future revert is silent.
  - `src/views/zero-page/zero-attachment-url.ts:249` — a cancelled download no longer shows the `artifacts.toasts.downloadFailed` toast. This is directly page-observable and is exactly what `docs/testing/testing-external-behavior.md` says to assert on.
  - `src/signals/voice-io/voice-io-stt.ts:579` and `:132` — a cancelled `getUserMedia` no longer toasts. `reportMicStartFailure` now throws on abort, which also changes the `tapError(openMedia(signal), reportMicStartFailure)` call site.

  At minimum the download-cancel path needs a page test: start a download, abort the owning signal, assert no failure toast appears, and keep a companion case asserting the toast still appears on a real network failure so the existing behavior is not silently lost. The mic path needs the same pair if it is reachable from a page test.

- **Existing behavior replaced without covering the old path.** Per the Feature Regression Guard, the mic-permission-denied and download-failure toasts are pre-existing user-visible behavior. The PR narrows when they fire but adds no test that they still fire in the non-abort case.

- **`src/signals/utils.ts:242` — new lint suppression.** `docs/bad-smell.md` §2 is zero-tolerance on `eslint-disable` comments, and `docs/cache.md` RC-9 says explicitly not to weaken or suppress a lint rule. The `apps/api` precedent cited in the PR body predates neither document and is not itself a justification. `onRejection` genuinely cannot satisfy the rule, so prefer a scoped config exclusion for that one infrastructure file, in the same shape as the existing `src/signals/utils.ts` exemption already present in this config for `no-restricted-syntax`, rather than an inline suppression. If the inline form is kept deliberately, say so in the PR summary with the reason the config-level form was rejected.

- **PR summary is missing the mandatory `Fallbacks` section.** `docs/fallback.md` §9 requires every PR to carry a `Fallbacks` section, or the explicit line `Fallbacks: none` when the diff contains none. The body currently has What / Verification / Notes only. Add `Fallbacks: none` — that line is what makes the post-merge audit trustworthy.

### Fallbacks

- None. The diff introduces, keeps, and removes no cross-version compatibility branch, no `??`/`||` default on a contract-guaranteed field, no feature-switch-gated tolerance path, and no retired-protocol reader. The three added `throwIfAbort(error)` calls are cancellation propagation, not compatibility with an older deployed version; the retained `signal.throwIfAborted()` calls are pre-existing abort checks on a different condition. Declared in PR summary: **No — P1** (see Findings; `docs/fallback.md` §9 requires the explicit `Fallbacks: none` line).

### Testing

- **Coverage: Insufficient.** Missing regression tests for: cancelled attachment download must not toast `downloadFailed` (and must still toast on a genuine fetch/CORS failure); cancelled `getUserMedia` must not toast `microphoneAccessDenied` (and must still toast on a real permission denial); aborted chat IDB open must not disable the store for the session. The two `realtime.ts` reorderings and the `onRejection` suppression need no new test — both preserve behavior — but the three above do not.
- **Conventions: Compliant (vacuously).** No test file is changed, so the mock-boundary, MSW, fake-timer, and mock-cleanup rules have nothing to violate. When the tests above are added, `docs/testing/app-testing.md` applies: drive them through the page with MSW handlers, and assert on the toast surface rather than on internal calls.
- Static checks are green on this HEAD: `lint-eslint`, `lint-knip`, `lint-format`, `lint-types`, `lint-types-apps`, all `test-app` and `test-api` shards, and `cli-e2e` all pass, which confirms the newly enabled rule has no remaining violations and the deleted files have no remaining references.

e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 63dc75e5efd4281bb6228393dc5244d88d6dde82
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Second review, at `63dc75e`. Both blockers from the `edea128` review are resolved and the resolution is better than what was asked for: instead of documenting the three claimed cancellation fixes with tests, the author re-derived them and found two guarded states their own contract cannot produce, so they were removed rather than tested. What remains is a lint enablement plus behavior-preserving refactors onto the existing centralized helpers.

The suppression count moves in the right direction: **two `eslint-disable` comments removed, none added**. `onRejection` and `settle` absorb two `no-restricted-syntax` opt-outs that predate this PR, and the one `ccstate/no-catch-abort` opt-out became a scoped config exclusion in the same shape the try-statement ban already uses for the same file.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | `fix:` commits must carry a regression test | N/A | Retitled to `refactor:`; the three behavior claims were withdrawn, so the `fix:` obligation no longer applies |
| `REVIEW.md` | `refactor:` commits must stay covered by existing tests | Pass | `chat-computer-use-and-voice`, `zero-chat-thread-idb`, `zero-attachment-chips` cover the four touched modules; 78 passed locally and `test-app` shards 2/3/7 are already green on this HEAD |
| `REVIEW.md` | Correctness and logic | Pass | `onRejection` is an exact tap-and-rethrow substitute; `settle(p, signal)` returns a union that never represents cancellation, so the `!opened.ok` branch cannot be entered by an abort |
| `REVIEW.md` | Security | N/A | No auth, input parsing, or credential handling in the diff |
| `REVIEW.md` | Type safety, no unjustified `any` | Pass | `settle<T>` narrows through the discriminated union; no casts added |
| `REVIEW.md` | Style and maintainability, prefer reuse | Pass | Both rewrites move onto existing `signals/utils.ts` helpers rather than adding new local error handling |
| `REVIEW.md` | Feature regression guard | Pass | No existing user-visible behavior is replaced; the one control-flow delta is declared in the summary |
| `REVIEW.md` | DB/JSONB persisted-contract gate | N/A | No schema, contract, or `jsonb().$type<>()` change |
| `docs/bad-smell.md` §2 | Zero tolerance for lint suppression comments | Pass | Net −2 suppressions; the previously flagged inline `ccstate/no-catch-abort` opt-out is gone |
| `docs/bad-smell.md` §3 | No unnecessary or over-engineered try/catch | Pass | Two try/catch blocks removed outright in favour of centralized helpers |
| `docs/bad-smell.md` §7 | Fail fast | Pass | `getChatIdb` rejection still disables the cache and re-throws; nothing is defaulted |
| `docs/fallback.md` §3 | No fallback for a state the contract cannot produce | Pass | This is the finding that drove the rewrite: `GetDb = () => Promise<IDBPDatabase>` takes no signal, so the withdrawn `throwIfAbort` was guarding an unreachable cancellation and would have re-read a genuine IndexedDB `AbortError` as one |
| `docs/fallback.md` §6 | Fallbacks only for the four permitted categories | Pass | None present |
| `docs/fallback.md` §9 | Summary must declare `Fallbacks` or `Fallbacks: none` | Pass | `## Fallbacks` section present with an explicit `Fallbacks: none`, plus the two rejected guards and why |
| `docs/testing.md` | Integration tests only, mock at the boundary | N/A | No test file changed |
| `docs/testing/app-testing.md` | Platform behavior is exercised through page tests | Pass | The author did write the page test the previous review asked for, then deleted it after it passed with the fix reverted; removing a vacuous test is correct under this doc rather than a coverage loss |
| `docs/testing/testing-external-behavior.md` | Assert on the user-visible surface | N/A | No test file changed |
| `docs/effect.md` | Make cancellation explicit, every async command has an owning signal | Pass | `openMedia` now threads its caller's signal into `settle` instead of discarding it inside a catch |
| `docs/cache.md` RC-8 | Symmetric resource lifecycles | Pass | `disableForSession` still runs on every real IDB open failure |
| `docs/cache.md` RC-9 | Do not weaken or suppress a lint rule | Pass (with note) | No suppression remains. The file-level exclusion does mean `src/signals/utils.ts` is unenforced for all 6 of its catch clauses, not just `onRejection`; the other 5 comply today, and the comment states the contract that makes the file exempt |
| `.claude/skills/ccstate/SKILL.md` | Use `signal.throwIfAborted()` only after an await that does not receive the signal | Pass | `zero-attachment-url.ts` now keeps only `throwIfAbort(error)` because `fetch` receives the signal; `openMedia` keeps a signal check because `getUserMedia` does not |
| `.claude/skills/ccstate/SKILL.md` | Every AbortSignal has a clear owner | Pass | No new controller; ownership unchanged |

### Findings

- None.

- **`openMedia`'s control-flow delta was not declared** (resolved during this review). The summary opened with a blanket "No behavior change", but `settle(p, signal)` re-throws the `AbortError` where the old catch swallowed it, toasted, and returned `undefined`. The user-visible outcome is the same — `tapError` still routes the error to `reportMicStartFailure`, the same toast fires, and recording still does not start — but the attempt now unwinds through the abort path instead of `return null`. The PR body has been corrected with a `Behavior delta` section stating this precisely. Per `REVIEW.md`'s verdict rules a P1 alone does not force `Changes Requested`, and this one is already fixed.

### Fallbacks

- None. The diff introduces, keeps, and removes no cross-version compatibility branch, no `??`/`||` default on a contract-guaranteed field, no feature-switch-gated tolerance path, and no retired-protocol reader. Declared in PR summary: **Yes** (`Fallbacks: none`). Verdict: **Justified**.
- Worth recording that two candidate fallbacks were removed rather than shipped, both under `docs/fallback.md` §3: the `throwIfAbort` on the IDB open (`getChatIdb()` accepts no signal) and the `throwIfAbort` in `reportMicStartFailure` (a page test that starts voice input and navigates away passes with and without it). The summary documents both, which is exactly the §10 evidence a removal is supposed to carry.

### Testing

- **Coverage: Adequate.** This is a `refactor:` with no user-visible behavior change, so `REVIEW.md` requires that existing tests still cover the touched code rather than new regression tests. `chat-computer-use-and-voice.test.tsx` exercises `openMedia` through the voice-input start flow, `zero-chat-thread-idb.test.tsx` exercises the IDB store, and `zero-attachment-chips.test.tsx` exercises the download failure toast; 78 tests pass across those three files.
- **Conventions: Compliant.** No test file is changed in the final diff.
- CI on `63dc75e` is still running with **no failures so far**: `lint-format`, `lint-knip`, `lint-types`, `lint-runtime-api-compat` and `test-app` shards 2, 3 and 7 are green; `lint-eslint`, `lint-types-apps` and the remaining shards are pending. The merge gate should confirm the full set before this lands.

PR #25721, opened 2026-08-07
fix(platform): restore chat position after thread switches

e7h4n:
## Summary

- Restore a reader's saved anchor after switching away from a chat thread and back.
- Read the stored thread position during event batches instead of re-measuring a temporarily empty or stale remount.
- Keep DOM-aware position synchronization for actual scroll events, so returning to the bottom still clears the saved anchor and resumes tail following.
- Keep only the reliable sidebar-switch regression test; the synthetic reply-arrival test was removed because its event ordering did not match production.
- Merge the latest `main` into the branch.

The branch also retains opt-in event diagnostics for the separate reply-arrival report. Preview traces show that the persisted `input.prompt` carrying the optimistic `clientEventId` reconciles before `output.message`, and a reader position captured after history backfill remains stable through output, completion, and follow-up event batches. The PR therefore remains scoped to the stored-position read fix and does not claim a separate second root cause.

## Fallbacks

- None.

## Test plan

- [x] `pnpm format`
- [x] `pnpm turbo run lint --concurrency=1 --log-order grouped`
- [x] `pnpm knip`
- [x] staged TypeScript checks for non-app packages, platform, and API
- [x] `pnpm --filter @vm0/app exec vitest run src/views/zero-page/__tests__/chat-scroll-position.test.tsx` — 22 passed
- [x] PR preview: thread A restored exactly to `scrollTop=1811` after switching A → B → A
- [x] PR preview diagnostic: after history backfill, `output.message` arrived while the reader stayed at `scrollTop=2810`


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: e92949dd54d8226e02254fa6f28b31ec55b5b349
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

This PR restores a saved chat anchor after a reader switches threads by using the thread's captured position during event batches, while preserving DOM-aware clearing when the reader actually returns to the tail. The sole review finding was metadata-only and is now resolved: the PR summary explicitly declares that the diff contains no fallback.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness, races, null paths, security, and type safety | Pass | Read the complete eight-file diff and affected scroll/event paths; stored positions remain nullable by design, event batches restore only captured state, and no unsafe casts, secrets, or injection surface were added. |
| `REVIEW.md` | A `fix:` PR must include regression coverage | Pass | `chat-scroll-position.test.tsx` drives A to a non-bottom anchor, navigates A to B and back through sidebar links, and verifies the visible anchor and `scrollTop` are restored. |
| `REVIEW.md` | Every fallback must be audited and declared | Pass | No fallback is present in the diff, and the PR body now explicitly declares `## Fallbacks — None`. |
| `docs/bad-smell.md` | Avoid `any`, lint suppressions, dynamic imports, hidden fallback logic, and over-engineered error handling | Pass | The new diagnostics and scroll branch use explicit types, static imports, and existing fail-fast commands without suppressions or new error-recovery branches. |
| `docs/testing.md` | Prefer integration tests, mock only boundaries, and assert behavior | Pass | The regression runs through the bootstrapped platform page, uses `context.mocks.api`, clicks real sidebar links, and asserts the rendered scroll result rather than mock calls. |
| `docs/fallback.md` | Default to no fallback and explicitly declare `Fallbacks: none` when none exists | Pass | The diff adds no cross-version, legacy-shape, fabricated-default, or tombstone behavior, and the summary now contains the required declaration. |
| `docs/event-sourcing.md` | Persistent events remain authoritative and reconcile optimistic events by shared event ID | Pass | The PR only adds diagnostics around the existing shared-ID reconciliation and does not add rollback, timeout, or failure-path cleanup. |
| `docs/effect.md` | Commands must match their owner and async work must retain lifecycle ownership | Pass | Event and scroll effects stay in semantic ccstate commands; existing `AbortSignal` ownership and awaited subcommands are preserved. |
| `docs/cache.md` | Avoid render-time side effects, unbounded caches, duplicate mutable truth, and asymmetric resources | Pass | No React render work or new cache is introduced; scroll state remains thread-scoped and existing observers retain their abort-owned teardown. |
| `.claude/skills/ccstate/SKILL.md` | Keep `get`/`set` inside commands, propagate signals, and avoid `detach` in signals | Pass | New ccstate reads and writes stay inside command callbacks, no accessor escapes or signal-layer `detach` calls were added, and async calls preserve their owner signals. |
| `docs/testing/patterns.md` | Platform tests should use `detachedSetupPage` and lifecycle-scoped mocks | Pass | The new test uses `detachedSetupPage`, shared page helpers, and `context.mocks.api`; the extracted layout helper remains scoped to the test lifecycle. |
| `docs/testing/anti-patterns.md` | Avoid internal mocks, fake timers, direct component rendering, and implementation-only assertions | Pass | No `vi.mock`, fake timers, direct React render, or mock-call assertion was added; DOM geometry models browser layout while the scenario is driven through page navigation. |
| `docs/testing/app-testing.md` | App tests must be page-level and verify page-visible behavior | Pass | The test opens the real chat route, interacts with sidebar links, and checks the reader-visible viewport position after remount. |
| `docs/testing/testing-external-behavior.md` | Construct and observe the scenario through the platform page | Pass | Thread switching is performed through user-facing links and the result is observed on the rendered chat viewport, without mutating ccstate stores. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Fallbacks

- None observed. The nullable values in `chatEventDebugSummaries` only normalize optional fields for opt-in debug output, and the existing resize-to-tail path is UI recovery rather than a version-migration or legacy-contract fallback. Declared in PR summary: Yes. Verdict: no code fallback.

### Testing

- Coverage: Adequate — the new page-level regression covers A → B → A restoration, while existing cases continue to cover tail following, event arrival, history prepend, resize, and sharing transitions.
- Conventions: Compliant — no internal module mocks, direct fetch mocks, fake timers, direct component rendering, or mock-only assertions were added.
- Verification: all published PR checks are green, including eight `test-app` shards, lint, type checks, Knip, security scans, and `ci-gate-turbo`; this review-finding pass also completed `pnpm format`, the required bounded serial lint command, and `pnpm check-types` successfully.

PR #25384, opened 2026-08-06
feat(chat): steer runs near time budget

e7h4n:
## Summary

- create a deterministic, API-side `input.budget` active input when an assistant callback observes a chat run at or beyond 115 minutes
- store the context-not-control steer text in the existing `user_message` field, attribute it to the current run, and preserve the event type when claimed
- expose budget events through the complete server ChatEvent stream, including chat thread APIs, without a server-side type filter
- retain budget events in the platform client's raw synchronized state while excluding them only from semantic visible run-group message rendering
- keep budget inputs scoped to the current run and out of the normal thread queue so they cannot start or reach a later run
- extend the existing chat event constraints without adding a database column or feature switch

## Verification

- `pnpm --filter @vm0/api-contracts check-types`
- `pnpm --filter @vm0/db check-types`
- `pnpm --filter @vm0/app check-types`
- `pnpm --filter api check-types`
- targeted Prettier, Oxlint (including type-aware rules), and ESLint for changed files
- `pnpm --filter @vm0/api-contracts exec vitest run src/contracts/__tests__/chat-events.test.ts` (16 passed)
- `pnpm --filter @vm0/app exec vitest run src/signals/chat-page/__tests__/chat-event-state.test.ts` (1 passed)
- API BDD cases cover the 115-minute threshold, deduplication, complete API projection, exact claim text, and cross-run isolation; local execution requires PostgreSQL and was unavailable (`ECONNREFUSED :5432`), so these will run in the PR pipeline

Closes #25345

PR #25726, opened 2026-08-07
refactor(api): steer run time budget from a per-minute cron

e7h4n:
## Summary

The 115-minute run-time-budget steer added in #25384 was registered as an optional webhook consumer filtered to `assistant` events:

```ts
{ name: "run-time-budget", command$: steerRunNearTimeBudget$, eventTypes: ["assistant"] }
```

`type: "assistant"` is the Claude Code event shape. Codex-backed runs report assistant output as `item.completed` (see `anthropicMessageText` vs `codexAgentMessageText` in `agent-event-consumer-run-output.service.ts`), so the filter never matched and those runs were never steered. Production confirms it: since #25384 merged on 2026-08-06, `chat_events` holds `input.budget` rows for exactly one run.

Event-driven delivery is also not sufficient on a matching backend. The steer only lands if the run happens to emit another event after crossing 115 minutes, and the end of a long run is exactly when a run is most likely to sit silent inside one long tool call.

This replaces the consumer with a scheduled sweep.

- add `GET /api/cron/steer-run-time-budget`, scheduled `* * * * *`, selecting running chat runs whose `started_at` is past the threshold
- keep the deterministic `uuidv5(runId, ...)` event id, so a repeated scan conflicts on the id instead of steering twice, and count only newly inserted rows as `steered`
- re-announce a still-unclaimed steer to the runner on the next scan through the existing `notifyRunningChatRunOfPendingInput`, which no-ops once the input is claimed
- delete `agent-event-consumer-run-time-budget.service.ts` and its registration
- drop the now-unused `eventTypes` matching from the webhook dispatcher, since no remaining optional consumer filters by event type

The steer text, event type, thread-queue isolation, and claim path are unchanged, so the runner and client contracts are untouched.

## Tests

The two budget cases in `chat-events.bdd.test.ts` now drive the cron route instead of the webhook. They age their own run with a new `backdateRunStartedAtFixture` rather than `mockNow`: the sweep is global and the API suites share one database, so a mocked clock would age every concurrent test file's runs as well.

## Verification

- `pnpm -F api check-types` (boundaries, gateways, core, bootstrap, tests, bootstrap-wiring)
- `pnpm -F api lint`, `pnpm -F @vm0/api-contracts lint`, `pnpm knip`
- Prettier on every changed file
- Vitest was not run locally: this sandbox has no PostgreSQL (`ECONNREFUSED :5432`), so the API suites run in the PR pipeline

## Fallbacks

None. The cron and the deleted consumer wrote the same deterministic event id, so an API instance still running the old code during the deploy overlap cannot produce a duplicate steer, and no persisted state changes shape.

### DIFF
--- a/turbo/apps/api/src/__tests__/vercel-crons.test.ts
+++ b/turbo/apps/api/src/__tests__/vercel-crons.test.ts
@@ -23,6 +23,7 @@ import {
   cronRenewGoogleWorkspaceEventSubscriptionsContract,
   cronReconcileBillingEntitlementsContract,
   cronRefreshStoragePresignedUrlsContract,
+  cronSteerRunTimeBudgetContract,
   cronSyncSkillsContract,
   cronTelegramCleanupContract,
 } from "@vm0/api-contracts/contracts/cron";
@@ -138,6 +139,10 @@ const expectedVercelCrons = [
   {
     path: cronAggregateModelStatsContract.aggregate.path,
     schedule: "12 * * * *",
+  },
+  {
+    path: cronSteerRunTimeBudgetContract.steer.path,
+    schedule: "* * * * *",
   },
 ] satisfies readonly VercelCron[];
 
--- a/turbo/apps/api/src/signals/route.ts
+++ b/turbo/apps/api/src/signals/route.ts
@@ -22,6 +22,7 @@ import { cronReconcileBillingEntitlementsRoutes } from "./routes/cron-reconcile-
 import { cronRefreshStoragePresignedUrlsRoutes } from "./routes/cron-refresh-storage-presigned-urls";
 import { cronComputerUseScreenshotCleanupRoutes } from "./routes/cron-computer-use-screenshot-cleanup";
 import { cronBrowserReconcileRoutes } from "./routes/cron-browser-reconcile";
+import { cronSteerRunTimeBudgetRoutes } from "./routes/cron-steer-run-time-budget";
 import { cronSyncSkillsRoutes } from "./routes/cron-sync-skills";
 import { cronTelegramCleanupRoutes } from "./routes/cron-telegram-cleanup";
 import { desktopAuthRoutes } from "./routes/desktop-auth";
@@ -244,6 +245,7 @@ export const ROUTES: readonly RouteEntry[] = [
   ...cronRefreshStoragePresignedUrlsRoutes,
   ...cronComputerUseScreenshotCleanupRoutes,
   ...cronBrowserReconcileRoutes,
+  ...cronSteerRunTimeBudgetRoutes,
   ...cronSyncSkillsRoutes,
   ...cronTelegramCleanupRoutes,
   ...emailMorningBriefUnsubscribeRoutes,
--- a/turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts
@@ -23,6 +23,7 @@ import {
   type UserMessageInputDocument,
 } from "@vm0/api-contracts/contracts/chat-threads";
 import { isChatRunTerminalEventType } from "@vm0/api-contracts/contracts/chat-events";
+import { cronSteerRunTimeBudgetContract } from "@vm0/api-contracts/contracts/cron";
 import { ACTIVE_INPUT_CONTROL_PAYLOAD_MAX_BYTES } from "@vm0/api-contracts/contracts/runners";
 import { zeroMailContract } from "@vm0/api-contracts/contracts/zero-mail";
 import {
@@ -43,6 +44,7 @@ import { clearMockNow, mockNow, now } from "../../../lib/time";
 import { accept, testContext } from "../../../__tests__/test-context";
 import { setupApp } from "../../../__tests__/test-helpers";
 import { server } from "../../../mocks/server";
+import { backdateRunStartedAtFixture } from "../../../test-fixtures/agent-runs";
 import { seedOrgMetadata } from "../../../test-fixtures/system-config-seeds";
 import { upsertOrgPlanEntitlementFixture } from "../../../test-fixtures/org-plan-entitlement";
 import {
@@ -95,6 +97,7 @@ import {
   replaceBddVm0ApiKey,
   replaceThreadSessionBindingFixture,
 } from "../../../test-fixtures/chat-events";
+import { cronSteerRunTimeBudgetRoutes } from "../cron-steer-run-time-budget";
 import { zeroChatEventsRoutes } from "../zero-chat-events";
 import { zeroChatThreadRoutes } from "../zero-chat-threads";
 import { zeroMailRoutes } from "../zero-mail";
@@ -443,6 +446,28 @@ async function claimChatRun(
     claim,
     sandboxHeaders,
   };
+}
+
+/**
+ * The time-budget sweep is global, so it can only be driven from a file that
+ * ages its own runs through `backdateRunStartedAtFixture`. Every other suite
+ * claims runs at the current time and therefore stays outside the window.
+ */
+async function runSteerRunTimeBudgetCron(): Promise<void> {
+  await accept(
+    setupApp({ context, routes: cronSteerRunTimeBudgetRoutes })(
+      cronSteerRunTimeBudgetContract,
+    ).steer({ headers: { authorization: "Bearer test-cron-secret" } }),
+    [200],
+  );
+}
+
+/** Age one claimed run to the given elapsed runtime. */
+async function ageClaimedRun(runId: string, elapsedMs: number): Promise<void> {
+  await backdateRunStartedAtFixture({
+    runId,
+    startedAt: new Date(now() - elapsedMs),
+  });
 }
 
 function claimEnvironment(claim: RunnerClaim): Record<string, string> {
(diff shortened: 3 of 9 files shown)


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 30eaf6ee717be95aa0ecbcbf8efa6d4fa2e0928a
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Replaces the event-driven 115-minute run-time-budget steer with a per-minute cron (`GET /api/cron/steer-run-time-budget`). The old path was an optional webhook consumer filtered to `assistant`, which is the Claude Code event shape, so Codex runs emitting `item.completed` were never steered. The sweep keeps the same deterministic `uuidv5(runId, …)` event id, the same steer text, and the same `input.budget` claim path, and the now-unused `eventTypes` matching is removed from the webhook dispatcher.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic, including race conditions | Pass | `cron-steer-run-time-budget.service.ts` re-verifies `status`, `chat_thread_id`, and `started_at` under `lockChatQueueThread` + `FOR UPDATE` inside the transaction, so a run that ends between the candidate query and the insert is dropped. |
| `REVIEW.md` | Missing auth on a new endpoint | Pass | `routes/cron-steer-run-time-budget.ts` gates on `hasValidCronSecret$` and returns `cronUnauthorized()`, identical to every sibling cron route. |
| `REVIEW.md` | Type safety, no `any` or broad types | Pass | No `any`, no `as` casts. `insertChatEvent` returns `ChatEventCommandResult \| null` and is narrowed with `inserted !== null`. |
| `REVIEW.md` | Style and maintainability, no function over ~100 lines or duplicated logic | Pass | Largest function is `persistRunTimeBudgetInput` at ~55 lines; the deleted consumer's duplicate of it is removed rather than kept alongside. |
| `REVIEW.md` | `refactor:` commits — existing tests must still cover the code | Pass | Both budget cases in `chat-events.bdd.test.ts` are migrated to drive the cron route; `vercel-crons.test.ts` covers schedule and route registration. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No `turbo/packages/db/src/schema/**` change, no migration, no `jsonb(...).$type<...>()` change. The `input.budget` event type and `user_message` shape are unchanged. |
| `REVIEW.md` | Feature regression guard — replaced behavior needs regression coverage | Pass | The replaced behavior is the steer itself, and its two existing tests were rewritten against the new trigger rather than deleted; no third behavior is silently rerouted. |
| `docs/bad-smell.md` | Zero tolerance for `any` and for lint/type suppressions | Pass | No suppression comment anywhere in the diff. |
| `docs/bad-smell.md` | Fail fast — no unnecessary try/catch or recovery logic | Pass | The sweep has no try/catch; a failing candidate aborts the invocation and the next minute retries, which the deterministic id makes safe. |
| `docs/bad-smell.md` | No dynamic `import()`, no hardcoded URLs or config | Pass | All imports static; the schedule lives in `vercel.json` and the secret in `hasValidCronSecret$`. |
| `docs/fallback.md` | Every fallback in the diff is declared in the PR summary | Pass | The PR summary carries an explicit `Fallbacks` section stating none, with the deploy-overlap rationale. |
| `docs/fallback.md` | No fallback for a state the contract already prevents | Pass | `if (!run?.chatThreadId)` narrows a genuinely nullable `zero_runs.chat_thread_id` selection rather than fabricating a default. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | No test asserts that the webhook no longer steers; the old assertions were rewritten to the new trigger and the old ones deleted. |
| `docs/fallback.md` | A removal PR must state its evidence and delete branch, contract entry, and tests together | Pass | `agent-event-consumer-run-time-budget.service.ts`, its `OPTIONAL_EVENT_CONSUMERS` entry, the `eventTypes` machinery, and the tests that drove it are removed in one commit. |
| `docs/deployment-compatibility.md` | Backend change must be safe for old runner → new backend and old code after deploy | Pass | The webhook request shape is untouched, so a draining runner keeps posting events successfully. During API rollout an old instance can still run the consumer, but both write the same `uuidv5(runId, 354e77a2-…)` id through `onConflictDoNothing({ target: id })`, so the overlap cannot duplicate a steer. |
| `docs/deployment-compatibility.md` | Migrations must be backward compatible with the deployed backend | N/A | No schema migration in the diff. |
| `docs/event-sourcing.md` | Persistent projections deduplicate by a shared event id | Pass | The steer keeps the run-derived deterministic id, which is what makes repeated scans and the rollout overlap converge on one event. |
| `docs/event-sourcing.md` | No imperative rollback of optimistic events on failure paths | N/A | No frontend projection or optimistic-event code in the diff. |
| `docs/testing.md` | Integration tests only, no unit tests for internal functions | Pass | The new coverage runs through the Hono app and the ts-rest contract; no service function is called directly from a test file. |
| `docs/testing.md` | No fake timers; no `vi.mock()` of internal modules | Pass | The migration removes `mockNow`/`clearMockNow` from both cases and adds no mocks. |
| `docs/testing.md` | Deployment-compatibility tests for a changed deployable boundary | N/A | No request, response, runner protocol, queue payload, or persisted shape changed, so there is no previous shape to test. |
| `docs/testing/api-testing.md` | Route tests go through `setupApp({ context, routes })(contract)` with `accept()` | Pass | `runSteerRunTimeBudgetCron` uses exactly that shape with `[200]`. |
| `docs/testing/testing-external-behavior.md` | Construct and verify through the endpoint; DB writes only for states the production surface cannot construct | Pass | Setup and every assertion go through `sendChatRun`, `claimChatRun`, `listRunnerActiveInputs`, `claimRunnerActiveInputs`, and `listThreadEvents`. The one internal write, `backdateRunStartedAtFixture`, ages a run by 115 minutes of wall clock, which no endpoint can produce; it lives in `test-fixtures/` rather than in the test file and documents why `mockNow` is unsafe for a global sweep. |
| `docs/testing/patterns.md` | Standard route-integration structure with module-scope `testContext()` | Pass | The file's existing module-scope `context` is reused; no new mock helper or env stub is introduced. |
| `docs/testing/anti-patterns.md` | AP-5 fake timers, AP-11 testing services when a route exists | Pass | The rewrite moves both cases from a clock mock onto the real cron endpoint, which is the direction AP-11 asks for. |
| `.claude/skills/ccstate/SKILL.md` | `get`/`set` must not leak out of the command callback | Pass | Helpers receive a resolved `Db` value obtained from `set(writeDb$)` inside the command; no `Getter`/`Setter` parameter or `{ set }` wrapper. |
| `.claude/skills/ccstate/SKILL.md` | `signal.throwIfAborted()` after every await that does not take the signal | Pass | Checked after the candidate query, after each `persistRunTimeBudgetInput`, and after each `notifyRunningChatRunOfPendingInput`. |
| `.claude/skills/ccstate/SKILL.md` | Never `detach()` in `signals/`; no floating promises | Pass | The sweep awaits every call; nothing is detached or fire-and-forget. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

- `cron-steer-run-time-budget.service.ts:loadRunTimeBudgetCandidates`: the scan takes the oldest 100 candidates and does not exclude runs already steered. Because already-steered runs are exactly the oldest ones, more than 100 simultaneous runs inside the five-minute window would let old entries occupy every slot and starve newer ones. That load is far outside current traffic and the failure mode is a missed safety-net steer rather than incorrect data, so this is a note, not a blocker. Excluding steered ids in SQL would remove the cliff if the window ever gets crowded.
- `cron-steer-run-time-budget.service.ts:steerRunsNearTimeBudget$`: candidates are processed sequentially and `lockChatQueueThread` uses a plain `SELECT … FOR UPDATE`, so one thread held by a concurrent drain delays the remaining candidates in that invocation. The per-minute schedule plus the deterministic id make a stalled invocation self-healing, and the existing thread-queue sweeps have the same characteristic.
- The re-announce path — an unclaimed steer re-published to the runner on the next scan — has no assertion. It is deliberately excluded here because the only observable difference is a realtime publish, which `docs/testing/testing-external-behavior.md` puts outside the endpoint boundary. Asserting that a rescan leaves exactly one unclaimed input would still be a cheap guard against a future refactor moving the notify call inside the insert branch.

### Fallbacks

- None. The diff introduces no `??`/`||` default on a contract-owned field, no branch for an old client, runner, route, or persisted shape, and no feature-switch gate. The rollout overlap between an old API instance still running the webhook consumer and the new cron is handled by the pre-existing deterministic event id and `onConflictDoNothing`, not by a compatibility branch, so there is nothing with a removal condition to track. The PR summary declares `Fallbacks: none` accordingly.

### Testing

- Coverage: Adequate. `refactor:` commit; both pre-existing budget behaviors — steer once at the threshold, and do not carry an unclaimed steer into a later run — are preserved and now exercised through the real cron endpoint, and the third cron call in the first case pins the no-duplicate-steer property. `vercel-crons.test.ts` fails if the route or schedule is not registered.
- Conventions: Compliant. No fake timers, no internal `vi.mock()`, no direct fetch mocking, real database, assertions through public endpoints. The single internal write is a documented `test-fixtures/` exception for a state that requires 115 minutes of elapsed wall clock, and replacing `mockNow` with it is the correct call given the sweep is global and the API suites share one database.
- Not run locally: the authoring sandbox has no PostgreSQL (`ECONNREFUSED :5432`), so the API suites are left to the PR pipeline. Type checks, `pnpm -F api lint`, `pnpm -F @vm0/api-contracts lint`, and `pnpm knip` all pass locally.

PR #25711, opened 2026-08-07
refactor(pi-agent-runtime): publish a vm0-owned type surface

e7h4n:
## Why

`@vm0/pi-agent-runtime` (added in #25489) exported raw TypeScript source: its `exports` map pointed `types` at `./src/index.ts`, and `src/index.ts` re-exported `@earendil-works/pi-agent-core` types directly. Every consumer's tsc program therefore expanded both `@earendil-works/pi-agent-core` and `@earendil-works/pi-ai`.

For `apps/api` that was **+240 MiB of `check-types` peak RSS**, which gave back half of the reduction #25454 had just landed. A CI ablation matrix over #25489 attributed the entire regression to this package: stubbing its entry point alone returned peak RSS to the pre-#25489 baseline, while reverting the new contracts, services, routes, and tests moved it by less than 20 MiB each.

## What

- `src/types.ts` declares the package's own published types (`ExecutionEnv`, `Result`, `FileInfo`, `Skill`, the transcript message union, model config). They mirror the native structural shapes rather than re-exporting them.
- `src/index.ts` is now a facade: every export is annotated with those owned types, and the native values (`ok`, `err`, `ExecutionError`, `FileError`) are re-bound behind owned constructor types so runtime identity is unchanged. These assignments are the conformance check, so a mirror that drifts from the native shape fails this package's own type check.
- `runPiAgentPrompt` / `runPiAgentResume` take `onMessage(message)` instead of `onEvent(event)`. Both consumers only ever handled `message_end`, and the native `AgentEvent` union cannot cross the boundary without dragging the whole graph back in.
- `src/node.ts` exposes `createPiNodeExecutionEnv()` instead of the native `NodeExecutionEnv` class.
- The package emits declarations (`tsconfig.build.json`, `pnpm build`) and `exports` resolves `types` before `import`. Runtime resolution still goes to source, so bundling and vitest are unchanged.
- `apps/api` and `apps/cli` build those declarations as the first step of `check-types`, because the `lint-types-apps` job invokes `pnpm --filter api run check-types` directly instead of going through turbo's `^build`.
- Internal-only exports (`resolvePiAgentModel`, `createPiReadTool`, the unresolved-tool-batch helpers) are no longer part of the public entry point; the package's own tests import them from their implementation modules.

## Measurements

Same-run A/B on CI runners using the repo's `scripts/measure-memory.mjs`, same `--max-old-space-size=3072`, same project order:

| `apps/api` project | before | after | delta |
|---|---|---|---|
[...]

PR #25468, opened 2026-08-06
test(api): isolate export runs from cleanup sweep

e7h4n:
## Summary

- serialize all nine past-clock user-export fixtures against the global cleanup-sandboxes sweep
- reuse the cleanup tests' transaction-scoped PostgreSQL advisory lock for explicit cross-worker ownership
- keep each fixture's create, execute, terminal-status, and latest-job observation inside that ownership boundary
- leave production cleanup behavior, export behavior, timeouts, and business assertions unchanged

## Failures and root cause

This PR covers three failures from the same shared-database race:

- [Turbo run 31089158743](https://github.com/vm0-ai/vm0/actions/runs/31089158743), [test-api (4) job 92575990392](https://github.com/vm0-ai/vm0/actions/runs/31089158743/job/92575990392): a gzip history hash-mismatch export lost its threadless run during claim and returned `Run not found`.
- [Turbo run 31092835709](https://github.com/vm0-ai/vm0/actions/runs/31092835709), [test-api (4) job 92587737672](https://github.com/vm0-ai/vm0/actions/runs/31092835709/job/92587737672): an oversized gzip export logged the expected S3 size error but exposed `Export job timed out`; the immediate retry then exposed the old failed job.
- [Turbo run 31093111509](https://github.com/vm0-ai/vm0/actions/runs/31093111509), [test-api (4) job 92588711845](https://github.com/vm0-ai/vm0/actions/runs/31093111509/job/92588711845): a completed gzip export disappeared from latest-job status, while an upload failure was overwritten by `Export job timed out`.

API integration-test workers share one PostgreSQL database while each worker owns its mocked application clock. The export tests create jobs and threadless web-trigger runs in May 2026. In parallel, `cron-cleanup-sandboxes.test.ts` exercises the production global sweep with an August 2026 clock. The failed shards' file timings overlap that cleanup suite.

At the later cleanup clock, the sweep can:

- delete a completed export job because its 72-hour expiry is already in the past, making status return `job: null` or fall back to the previous failed job;
- mark a pending or running export job older than ten minutes as failed with `Export job timed out`, overwriting the terminal state observed by the export test; and
- delete or cancel a threadless web-trigger run before the session-history fixture claims it.

[...]

PR #25727, opened 2026-08-07
test(api): scope the threadless batch sweep assertion to its own runs

e7h4n:
## Summary

- keep the exact sweep bound (`discovered === 20`), which this test owns because its own 21 fixtures always saturate one pass
- attribute the rest of the batch to this test's own runs: no own errors, own survivors are the newest suffix, survivors untouched and still eligible
- leave production cleanup behaviour, the sweep limit, and every other assertion in the file unchanged

## Failure

`GET /api/cron/cleanup-sandboxes > processes only the oldest bounded batch of threadless runs`, [test-api (4) job 92872721593](https://github.com/vm0-ai/vm0/actions/runs/31180587811/job/92872721593), merge-group commit `fab38c3` for #25711.

```
AssertionError: expected { discovered: 20, cancelled: 1, …(4) } to match object { discovered: 20, deleted: 20, …(1) }
-   "deleted": 20,
+   "deleted": 19,
```

Not a regression: #25711 touches no cron, run-deletion, sandbox or DB code, the same shard was green on its head `fa109b9`, and the re-queued merge group `2110a56` passed.

## Root cause

Test isolation on a shared database, not an async-ownership or product race.

`cleanupThreadlessRuns$` scans a **global** cohort: threadless, non-`test` runs, ordered `createdAt ASC, id ASC`, `LIMIT 20`. The test seeds 21 runs back-dated to `2026-08-03T05:40:26.001Z…021Z` and asserts the whole batch is its own.

API test workers share one PostgreSQL database, and two independent facts let a parallel file's run sort **ahead** of every one of those fixtures:

1. Chat runs are stamped with the creating worker's mocked clock. `zero-browser.test.ts` — same shard — runs its entire file at `2026-07-24T10:00:00Z`.
2. `chat_thread_events.created_at` is `defaultNow()`, the real wall clock, so a deletion tombstone is always past the `2026-08-03T05:40:26Z` forward cutoff. Every thread deletion therefore admits that thread's past-clock runs into the forward cohort through the callback/tombstone branch of `loadThreadlessRunCandidates`.

Verified against a local shard-4 database: after `zero-browser.test.ts` alone, 10 threadless runs remain at `created_at = 2026-07-24 10:00:00`, and the production candidate predicate reports `tombstone_admitted = true` for all 10. Between `chat.deleteThread` and the run's cancellation being flushed, such a run is still `running`/`pending`, so the sweep takes the active branch: `discovered: 20, cancelled: 1`, and one of the test's own fixtures is pushed out of the batch — `deleted: 19`. The intermittency is exactly that window overlapping this test's sweep.

Reproduced deterministically by seeding one active threadless run at `THREADLESS_FORWARD_CUTOFF_MS` before the sweep, which reproduces the CI message byte for byte:

```
AssertionError: expected { discovered: 20, cancelled: 1, …(4) } to match object { discovered: 20, deleted: 20, …(1) }
```

The file's `withThreadlessRunCleanupTestLockFixture` serialises concurrent *sweeps*; it cannot stop another worker from putting rows into the cohort, so the lock never covered this failure mode. Extending the lock to every thread-deleting test (the approach #25468 took for past-clock export fixtures) does not close it either: the window is inherent to the real-clock tombstone, so it would only make the race rarer.

## Fix

The test now asserts what it owns:

- `discovered === THREADLESS_SWEEP_LIMIT` — still exact and now provably immune, because foreign rows can only add eligible rows while 21 own fixtures already guarantee a saturated batch. This is the assertion that proves the bound.
- sweep `errors` contain none of this test's runs.
- the own runs left behind are exactly the newest suffix of the fixture order, proving the pass took an oldest-first prefix, and at least one is left behind, proving the bound truncated the batch.
- those survivors are untouched (`completed`, no error), so they remain eligible for a later pass.

The second global pass was dropped. It asserted `discovered >= 1, deleted >= 1` and that the leftover run was gone, which has the same shared-cohort dependency and is redundant: deleting a quiescent threadless run in one pass is already covered by "waits through the quiet window and deletes at its exact boundary", "processes every non-test trigger source when the run is threadless", and "redrives expired unresolved cancellation recovery before deletion".

No retry, sleep, timeout increase, relaxed business assertion, or skipped test.

## Fallbacks

None. This change touches one test file and adds no compatibility path, no `??`/`||` default on a contract-owned field, no feature-switch-gated branch, and no negative test asserting removed behavior stays removed.

## Validation

- target file, clean database: 5 consecutive runs, 33/33 tests passed each
- target test with an injected active threadless run older than every fixture: 3 consecutive runs, passed each (the previous assertion fails on the first)
- `pnpm vitest run --project=api --shard=4/8` (the failing CI shard, 23 files): 312/312 passed
- `pnpm -F api run check-types:core`, `check-types:tests`
- `eslint --max-warnings 0` and `oxlint --type-aware` on the changed file
- `prettier --check`

## Follow-up, not in this PR

`zero-browser.test.ts` permanently strands 10 threadless runs in the shared database: 7 of its tests delete a chat thread while only 3 delete the corresponding run roots. That is real fixture leakage and adds noise to every later sweep in the shard, but it is not what makes this test flaky, so it is left out of this change.

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/cron-cleanup-sandboxes.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/cron-cleanup-sandboxes.test.ts
@@ -60,6 +60,8 @@ const BUCKET = "test-user-storage-bucket";
 const FIXED_NOW_MS = Date.parse("2000-01-01T00:10:00.000Z");
 const THREADLESS_FORWARD_CUTOFF_MS = Date.parse("2026-08-03T05:40:26.000Z");
 const THREADLESS_TEST_NOW_MS = Date.parse("2026-08-03T06:00:00.000Z");
+// Mirrors THREADLESS_RUN_SWEEP_LIMIT in threadless-run-cleanup.service.ts.
+const THREADLESS_SWEEP_LIMIT = 20;
 const NON_TEST_TRIGGER_SOURCES: readonly TriggerSource[] =
   triggerSourceSchema.options.filter((source) => {
     return source !== "test";
@@ -419,6 +421,49 @@ async function findRun(runId: string): Promise<{
   return row
     ? { status: stringField(row, "status"), error: nullableString(row.error) }
     : null;
+}
+
+interface ThreadlessCleanupError {
+  readonly runId: string;
+  readonly error: string;
+}
+
+/** Keeps sweep failures attributable to this test's own fixtures. */
+function ownCleanupErrors(
+  threadlessRuns: {
+    readonly errors: readonly ThreadlessCleanupError[];
+  },
+  fixtures: readonly RunFixture[],
+): readonly ThreadlessCleanupError[] {
+  const owned = new Set(
+    fixtures.map((fixture) => {
+      return fixture.runId;
+    }),
+  );
+  return threadlessRuns.errors.filter((entry) => {
+    return owned.has(entry.runId);
+  });
+}
+
+/** Returns this test's still-present runs in the fixture order. */
+async function findRemainingRunIds(
+  fixtures: readonly RunFixture[],
+): Promise<readonly string[]> {
+  const states = await Promise.all(
+    fixtures.map(async (fixture) => {
+      return {
+        runId: fixture.runId,
+        present: (await findRun(fixture.runId)) !== null,
+      };
+    }),
+  );
+  return states
+    .filter((state) => {
+      return state.present;
+    })
+    .map((state) => {
+      return state.runId;
+    });
 }
 
 async function findRunnerJob(runId: string): Promise<{
@@ -853,7 +898,7 @@ describe("GET /api/cron/cleanup-sandboxes", () => {
     const userId = `user-${randomUUID()}`;
     const orgId = `org-${randomUUID()}`;
     const fixtures = await Promise.all(
-      Array.from({ length: 21 }, async (_, index) => {
+      Array.from({ length: THREADLESS_SWEEP_LIMIT + 1 }, async (_, index) => {
         return await trackRun(
           insertRunFixture({
             status: "completed",
@@ -873,24 +918,40 @@ describe("GET /api/cron/cleanup-sandboxes", () => {
       apiClient().cleanup({ headers: cronHeaders() }),
       [200],
     );
-    expect(firstResponse.body.threadlessRuns).toMatchObject({
-      discovered: 20,
-      deleted: 20,
-      failed: 0,
-    });
-    await expect(findRun(fixtures[20]!.runId)).resolves.not.toBeNull();
 
-    const secondResponse = await accept(
-      apiClient().cleanup({ headers: cronHeaders() }),
-      [200],
+    // The sweep cohort is global while API test workers share one database, so
+    // a parallel file that strands its own threadless run can occupy part of
+    // this batch. The bound stays exact because these fixtures alone exceed it.
+    expect(firstResponse.body.threadlessRuns.discovered).toBe(
+      THREADLESS_SWEEP_LIMIT,
     );
     expect(
-      secondResponse.body.threadlessRuns.discovered,
-    ).toBeGreaterThanOrEqual(1);
-    expect(secondResponse.body.threadlessRuns.deleted).toBeGreaterThanOrEqual(
-      1,
+      ownCleanupErrors(firstResponse.body.threadlessRuns, fixtures),
+    ).toStrictEqual([]);
+    const remaining = await findRemainingRunIds(fixtures);
+    // Oldest first: the pass takes a prefix of these fixtures, and the bound
+    // leaves the newest ones for a later pass.
+    expect(remaining).toStrictEqual(
+      fixtures.slice(fixtures.length - remaining.length).map((fixture) => {
+        return fixture.runId;
+      }),
     );
-    await expect(findRun(fixtures[20]!.runId)).resolves.toBeNull();
+    expect(remaining.length).toBeGreaterThanOrEqual(1);
+    expect(firstResponse.body.threadlessRuns.deleted).toBeGreaterThanOrEqual(
+      fixtures.length - remaining.length,
+    );
+    // What the bound left behind stays untouched and eligible for a next pass.
+    await expect(
+      Promise.all(
+        remaining.map((runId) => {
+          return findRun(runId);
+        }),
+      ),
+    ).resolves.toStrictEqual(
+      remaining.map(() => {
+        return { status: "completed", error: null };
+      }),
+    );
   });
 
   it("acknowledges an event projection that loses the root-delete race", async () => {


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 8909a06183b53666535397046d075ecb1e7cfae6
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Test-only change to `cron-cleanup-sandboxes.test.ts`. The bounded-batch threadless sweep test asserted exact totals (`discovered: 20, deleted: 20, failed: 0`) from a global scan on a database shared by every parallel API test worker, so a foreign threadless run sorting ahead of its fixtures took a slot in the batch. The test now keeps the exact bound `discovered === THREADLESS_SWEEP_LIMIT` — which its own 21 fixtures always saturate regardless of foreign rows — and attributes the rest of the batch to its own runs: no own sweep errors, own survivors are the newest suffix of the fixture order, at least one survivor remains, and the survivors are untouched and still eligible.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic: no off-by-one, wrong conditionals, or bad assumptions | Pass | `cron-cleanup-sandboxes.test.ts:918-955`. `discovered === 20` holds for any foreign-row count because 21 own eligible fixtures always saturate the `LIMIT`. The suffix assertion `fixtures.slice(fixtures.length - remaining.length)` is correct for oldest-first processing, and the `deleted >= fixtures.length - remaining.length` lower bound collapses to the original `deleted >= 20` when the cohort is clean. |
| `REVIEW.md` | Type safety: no `any`, no overly broad types | Pass | `ThreadlessCleanupError` is declared explicitly; `ownCleanupErrors` and `findRemainingRunIds` have explicit parameter and return types with `readonly` arrays. |
| `REVIEW.md` | Testing coverage matches commit type | Pass | `test:` commit that removes nondeterminism from an existing test. No production code changed, so no new production coverage is required. |
| `REVIEW.md` | Feature regression guard: changed behavior needs regression coverage and containment | N/A | No product behavior, UI, API, or user flow is changed; the diff is confined to one test file. |
| `REVIEW.md` | DB/JSONB review gate for persisted contract changes | N/A | No schema, contract, migration, or `jsonb().$type<>()` change in the diff. |
| `REVIEW.md` | Security: no injection, hardcoded secrets, or missing auth on new endpoints | N/A | No endpoint, credential, or query construction added; the test calls the existing cron route with the existing test secret. |
| `docs/bad-smell.md` | Zero tolerance for `any` | Pass | No `any` in the diff. |
| `docs/bad-smell.md` | Zero tolerance for lint/type suppression comments | Pass | No `eslint-disable`, `oxlint-disable`, `@ts-ignore`, `@ts-expect-error`, or `prettier-ignore` added. |
| `docs/bad-smell.md` | No dynamic `import()` | Pass | All imports remain static; the diff adds no `await import(...)`. |
| `docs/bad-smell.md` | No over-engineered error handling | Pass | No try/catch added; `ownCleanupErrors` filters the contract's `errors` array rather than swallowing failures. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | The removed second sweep pass was positive coverage, not a tombstone, and nothing in the diff asserts that deleted behavior is still deleted. |
| `docs/fallback.md` | Every fallback in the diff must be declared in the PR summary | Pass | The diff introduces, keeps, and removes no fallback. The PR summary now carries an explicit `## Fallbacks` section stating `None`. |
| `docs/fallback.md` | A rollout fallback must name surface, window, and removal condition | N/A | No cross-version or rollout compatibility branch exists in the diff. |
| `docs/testing.md` | Integration tests only; no unit tests for internal functions | Pass | The test still drives the real `/api/cron/cleanup-sandboxes` route through `setupApp()` and the ts-rest contract; the new helpers are assertion utilities, not a new unit-test surface. |
| `docs/testing.md` | Mock at the boundary only; no `vi.mock()` on relative paths | Pass | The diff adds no `vi.mock()` at all. |
| `docs/testing.md` | Real database; isolate test data with unique user and organization IDs | Pass | Fixtures already use `user-${randomUUID()}` / `org-${randomUUID()}`; this PR extends that isolation to the assertions, which is exactly the gap it fixes. |
| `docs/testing.md` | No fake timers; control only the time abstraction | Pass | Unchanged use of the repository's `mockNow`; no `vi.useFakeTimers()` or `advanceTimersByTime`. |
| `docs/testing.md` | No manual detached cleanup, sleeps, retries, or raised timeouts to mask flakes | Pass | The fix is scoping, not waiting. No `clearAllDetached()`, `sleep`, retry loop, or timeout change in the diff. |
| `docs/testing/anti-patterns.md` | AP-1: assert behavior, not that a mock was called | Pass | Assertions read the sweep's response body and the runs' persisted state through the route; no `toHaveBeenCalled()` added. |
| `docs/testing/anti-patterns.md` | AP-4: never mock internal code | Pass | No internal module is mocked; the sweep, run deletion, and state reads all run for real. |
| `docs/testing/anti-patterns.md` | AP-7: do not assert on implementation details | Pass | The new assertions use the route's published `threadlessRuns` counters plus run status read back through the test-state action, not internal service state. |
| `docs/testing/patterns.md` | Pattern 1: API endpoints are covered by route integration tests | Pass | The test continues to exercise the endpoint via `apiClient().cleanup(...)` and `accept(..., [200])`. |
| `docs/testing/api-testing.md` | Exercise the real Hono app via `setupApp()` + contract; mock only external services | Pass | Unchanged setup; the only mocks in the file remain `context.mocks.s3` and env. |
| `docs/testing/testing-external-behavior.md` | Construct and verify through the external surface, not DB rows or services | Pass | `findRemainingRunIds` reuses the file's existing `findRun`, which posts to the `cron-cleanup-sandboxes-state` action route. The diff imports no DB schema and issues no direct SQL. |
| `.claude/skills/ccstate/SKILL.md` | No floating promises; async work must be awaited or explicitly owned | Pass | Every added promise is awaited: `await Promise.all(...)` in `findRemainingRunIds`, `await expect(Promise.all(...)).resolves...`. No `detach()` or `void` introduced, and none belongs in a test body. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

### Fallbacks

None. The diff introduces, keeps, and removes no fallback: no `??`/`||` default on a contract-owned field, no compatibility branch for an old client, runner, route, or persisted shape, no feature-switch-gated behavior, and no negative test pinning deleted behavior. The PR summary declares this explicitly under `## Fallbacks`.

### Testing

- Coverage: Adequate. The exact sweep bound is still pinned — `discovered === THREADLESS_SWEEP_LIMIT` fails if the limit is raised (it would report 21) or lowered (it would report fewer), and it is immune to foreign rows because they only add eligible candidates. Oldest-first ordering is asserted against the test's own fixture order, and the survivors are asserted to be untouched and still eligible.
- Conventions: Compliant. No prohibited mocking, no fake timers, no sleeps, retries, or timeout increases, no manual detached cleanup, and no direct database access.
- Residual gap worth knowing: because `deleted` is now asserted as `>= fixtures.length - remaining.length`, a hypothetical regression where the sweep discovers 20 but deletes none would make this one assertion vacuous (survivors would be all 21 and the bound would be `>= 0`). In the clean single-worker case survivors are exactly 1, so the assertion is `deleted >= 20`, identical in strength to the previous version. Deletion of a quiescent threadless run in a single pass also remains directly covered by `waits through the quiet window and deletes at its exact boundary`, `processes every non-test trigger source when the run is threadless`, `terminalizes a stranded inline delivery callback before deleting its run`, and `redrives expired unresolved cancellation recovery before deletion` in the same file. Not a blocker; noted so the tradeoff is explicit.
- Author validation reported in the PR and consistent with the diff: target file 5 consecutive clean-database runs (33/33 each), the target test passing 3 consecutive times with an active threadless run injected older than every fixture (the previous assertion fails on the first), the failing CI shard `--shard=4/8` at 312/312, plus `check-types:core`, `check-types:tests`, `eslint --max-warnings 0`, `oxlint --type-aware`, and `prettier --check`.

PR #25728, opened 2026-08-07
fix(ui): give icons the colour of the text they sit with

tongx-blip:
Keeps the current Tabler icon set. This only fixes how icons are coloured.

Two reports, one root cause: **icons were being painted translucent.**

## The "overlay" artifact

Each path in an SVG is drawn separately, so where two strokes overlap, a semi-transparent colour composites with itself and the joint renders darker than the stroke body. Measured off the reported sidebar collapse icon:

| | stroke body | crossing | ratio |
|---|---|---|---|
| reported screenshot | `rgb(138,140,143)` | `rgb(104,106,110)` | **1.31x** |
| same icon rendered at alpha 0.7 | `rgb(136,139,142)` | `rgb(102,105,109)` | 1.30x |
| same icon rendered opaque | `rgb(89,91,95)` | `rgb(88,90,94)` | **1.01x** |

Solving the ratio back gives alpha ≈ **0.69**, and the button carries `text-sidebar-foreground/70`. Opaque removes the artifact entirely.

## Icons reading greyer than their label

`apps/platform/src/views/css/index.css` had:

```css
```

Every sidebar icon rendered 30% transparent while its label rendered at full strength. On the light theme that resolves to exactly `rgb(89,92,96)` against a `rgb(21,24,30)` label — which is what was measured off the reported screenshot.

Separately, 87 icons carried `text-muted-foreground` while the text next to them did not, so they were a lighter grey by construction.

## What changed

| | |
|---|---|
| Neutral colour utilities removed from icons | 106 |
| `opacity-50/70/80` removed from icons | 14 |
| Icon-only containers de-alpha'd | 14 |
| Redundant `hover:` colour classes removed | 11 |
| `.zero-nav svg { opacity: 0.7 }` | removed |

Removing an icon's own colour class is what makes it match its label: it then inherits from the same row the text does. Where the row is muted, both stay muted; where it isn't, both are full strength. Either way they agree.

**Untouched:** semantic status colours (`text-green-600`, `text-destructive`, `text-amber-500`, `text-primary`, ...) and the `opacity-0` / `text-muted-foreground/0` transition states. No icon is swapped, no icon library changes, no new colour token.

## Verification

- `pnpm turbo run lint` — 0 errors, warning count on `@vm0/app` unchanged from `main` (55)
- `pnpm check-types` — passes for `@vm0/ui`, `@vm0/app`, `@vm0/desktop`
- `pnpm -F @vm0/app build` — succeeds
- `@vm0/ui` tests and the CSS entrypoint test pass
- Audit: 0 neutral/alpha colour utilities left on any icon

### DIFF
--- a/turbo/apps/platform/src/views/activity-page/activity-inspect-page.tsx
+++ b/turbo/apps/platform/src/views/activity-page/activity-inspect-page.tsx
@@ -115,7 +115,7 @@ function InspectEmptyState() {
         })}
       />
       <div className="flex-1 flex flex-col items-center justify-center gap-3 pb-20">
-        <IconUpload size={48} stroke={1} className="text-muted-foreground/40" />
+        <IconUpload size={48} stroke={1} className="" />
         <h2 className="text-lg font-semibold text-foreground">
           {t(($) => {
             return $.activity.inspect.noLog.title;
--- a/turbo/apps/platform/src/views/artifacts-page/artifact-catalog-page.tsx
+++ b/turbo/apps/platform/src/views/artifacts-page/artifact-catalog-page.tsx
@@ -247,7 +247,7 @@ function ArtifactSharedConversationList({
                 size={16}
                 stroke={1.7}
                 aria-hidden
-                className="shrink-0 text-muted-foreground transition-colors group-hover:text-foreground"
+                className="shrink-0 transition-colors group-hover:text-foreground"
               />
             </button>
           </li>
--- a/turbo/apps/platform/src/views/css/index.css
+++ b/turbo/apps/platform/src/views/css/index.css
@@ -626,14 +626,6 @@ body {
   border-color: hsl(var(--gray-400));
 }
 
-/* Lighter icons and "Recent dialogue" label; other styles unchanged */
-.zero-nav svg {
-  opacity: 0.7;
-}
-.zero-nav .zero-nav-recent-label {
-  opacity: 1;
-}
-
 @layer base {
   *,
   *::before,
--- a/turbo/apps/platform/src/views/email-unsubscribe/email-unsubscribe-page.tsx
+++ b/turbo/apps/platform/src/views/email-unsubscribe/email-unsubscribe-page.tsx
@@ -26,7 +26,7 @@ export function EmailUnsubscribePage() {
         <VM0Logo />
         {status === "done" ? (
           <div className="flex flex-col items-center gap-2.5">
-            <IconCheck size={20} className="text-muted-foreground" />
+            <IconCheck size={20} className="" />
             <h1 className="text-lg font-medium text-foreground">
               {t(($) => {
                 return $.lifecycle.emailUnsubscribe.doneTitle;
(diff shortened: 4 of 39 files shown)


tongx-blip:
Verified on the deployed preview.

Deployed CSS — the dimming rule is gone and the Tabler stroke rule is untouched:

```
.zero-nav svg opacity rule : 0 occurrences
svg[class*="tabler-icon"]{stroke-width:var(--icon-stroke-width)}
--icon-stroke-width:1.5
```

Sidebar rows, computed style:

```
Agents      icon=rgb(21,24,30) opacity=1 stroke=1.5px | label=rgb(21,24,30)
Workflows   icon=rgb(21,24,30) opacity=1 stroke=1.5px | label=rgb(21,24,30)
Connectors  icon=rgb(21,24,30) opacity=1 stroke=1.5px | label=rgb(21,24,30)
Artifacts   icon=rgb(21,24,30) opacity=1 stroke=1.5px | label=rgb(21,24,30)
```

The reported icon specifically:

```
tabler-icon-layout-sidebar-left-collapse
  color: rgb(21, 24, 30)   <- opaque, previously rgba at alpha 0.7
  opacity: 1
```

Page-wide sweep: **0** translucent icons out of 30 SVGs. Icon geometry and stroke width are unchanged from `main`.

tongx-blip:
Rescoped: only icons that sit next to text change. Icon-only controls keep exactly the shade they have today.

Every icon usage is classified by whether it has a text sibling (a forward scan through its siblings, stopping at the tag that closes its parent — parent-walking was unusable because TypeScript generics like `Parameters<typeof Link>` parse as tags).

| | count |
|---|---|
| Paired with text — colour utility dropped so the icon inherits its labels colour | 59 |
| Standalone sidebar controls — rendering preserved exactly | 19 |

**Paired**, verified live: icon colour equals label colour in every row, including the muted ones —

```
Agents / Workflows / Connectors / Artifacts   icon=rgb(21,24,30)  = label=rgb(21,24,30)   opacity 1
Invite                                        icon=rgb(81,90,103) = label=rgb(81,90,103)
```

**Standalone** (collapse, `+`, `...`, pin): these rendered at `ancestor colour alpha x 0.7`, the second factor from `.zero-nav svg { opacity: .7 }`. That rule is removed — it was the reason nav icons were lighter than their labels — and each control now carries the same product as its own element opacity, so it looks unchanged:

```
collapse buttons   0.7 x 0.7 -> opacity-50
collapsed rail     1.0 x 0.7 -> opacity-70
section chevrons   0.5 x 0.7 -> opacity-35
```

That move also removes the reported **overlay**. A translucent *colour* makes overlapping strokes composite against each other; element *opacity* flattens the icon first. Measured on the same icon at the same shade:

| | body | crossing | ratio |
|---|---|---|---|
| `rgba` colour @ 0.49 | rgb(137,138,142) | rgb(80,81,86) | 1.51x |
| opaque + `opacity: .49` | rgb(136,138,141) | rgb(136,138,141) | **1.00x** |

Both collapse buttons now use an opaque colour. Where the alpha sits on an ancestor that also carries text (two section-header chevrons and the pinned `+`), the colour is left as-is so that text does not shift — those keep a mild artifact, called out here rather than traded for a text-colour change.

Semantic status colours and `opacity-0` transition states untouched. No icon replaced, no colour token added.

tongx-blip:
Superseded by #25995. This was written against the Tabler icon set; the Lucide migration landed since (#25852, #25858), so the same fixes are reapplied on top of Lucide there — plus the stroke-width inconsistency the migration introduced.

PR #25731, opened 2026-08-07
test(api): scope shared-database test lookups to their own resources

e7h4n:
## Summary

- scope the three chat-event prompt fixtures to the caller's own user instead of scanning the whole table
- assert the browser reconcile outcome through this test's own provider instances instead of the route's global counters
- leave production code, test coverage, and every business assertion unchanged

## Why

Follow-up to the review comment on #25727: tests must isolate through their own random user/org resources, not through the assumption that a previous test cleaned up after itself. API test workers share one database, so any lookup or count that is not scoped to the caller's own rows reads whatever another worker left behind.

Reproduction, before this change — run the same shard twice against one database:

```
pnpm vitest run --project=api --shard=4/8   # pass 1: 23 files, 312 tests, green
pnpm vitest run --project=api --shard=4/8   # pass 2: 4 tests fail in 2 files
```

```
agentphone.bdd > shares one canonical session across AgentPhone and web messages on the same thread
  AssertionError: expected { …(88) } to match object { contextType: 'agentphone', …(12) }
-   "agentphoneFromNumber": "+15558957596",
+   "agentphoneFromNumber": "+15557980211",

zero-browser > isolates profiles across concurrent thread browser sessions
  AssertionError: expected { checked: 9, stopped: 9, …(2) } to match object { checked: 1, stopped: 1, errors: +0 }
```

CI gives every shard a fresh database, so these four are not flaky in CI today. They break the documented local workflow (`pnpm -F api exec vitest` against a dev database) on the second run, and they become CI-flaky as soon as another file in the same shard writes the same kind of row — shard composition changes every time a test file is added.

## What was unscoped

**`test-fixtures/chat-events.ts`** — three exported lookups selected across the entire `chat_events` table and picked the first row whose prompt text matched, with no ownership predicate:

- `findTelegramChatEventByPromptFixture`
- `findAgentphoneChatEventByPromptFixture`
- `findPendingChatEventByPromptFixture` (only `runId IS NULL`, not even a context type)

Used by 5 test files across 14 call sites. Some callers already worked around it by embedding a tenant id or `randomUUID()` in the prompt text; others pass plain literals, and two files already use the same literal `"summarize this thread"` for different contexts. The fixtures now take `{ userId, prompt }` and join `chat_threads` to filter on the caller's own user, so a text collision can no longer cross actors.

Scoping also surfaced a latent gap: the Feishu group test read a queued event that belongs to its `secondActor`, not to the fixture owner. The global lookup silently returned it; the scoped lookup made the intended owner explicit.

**`zero-browser.test.ts`** — the `isolates profiles across concurrent thread browser sessions` test asserted `checked: 1, stopped: 1` and then `checked: 0, stopped: 0` from `/api/cron/browsers/reconcile`, which scans every active browser session instance in the database. It also counted provider stop calls with a mock handler that matched any browser id, so instances belonging to other rows inflated the count too. The test now records the stopped provider ids and asserts against its own `providerIds`, keeping the same numbers and the same idempotence claim.

## Validation

- `--shard=4/8` twice against one database: 317/317 passed on both passes (previously 4 failures on the second)
- `agentphone.bdd` + `zero-integrations-telegram-post` + `internal-callbacks-teams` + `zero-feishu-integration` twice against one database: 85/85 passed on both passes
- `zero-browser.test.ts` three consecutive times against one database: 13/13 passed each
- `pnpm -F api run check-types:gateways`, `check-types:core`, `check-types:tests`
- `eslint --max-warnings 0` on all six changed files, `oxlint --type-aware`, `prettier --check`

## Not in this PR

`zero-browser.test.ts` has about twenty more assertions on the same global reconcile counters (`checked`/`stopped`/`errors`). Only the one above is reachable today, because the tests that follow it create their instances after it runs. Converting the rest is a mechanical but large change to that file and deserves its own review.

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/agentphone.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/agentphone.bdd.test.ts
@@ -320,8 +320,10 @@ describe("INT-03: AgentPhone linked-run lifecycle through public APIs", () => {
     });
     await waitForTyping(sends, [conversationId]);
 
-    const admitted =
-      await findAgentphoneChatEventByPromptFixture("summarize my inbox");
+    const admitted = await findAgentphoneChatEventByPromptFixture({
+      userId: actor.userId,
+      prompt: "summarize my inbox",
+    });
     expect(admitted).toMatchObject({ eventId: expect.any(String) });
     if (!admitted) {
       throw new Error("Expected admitted AgentPhone input event");
@@ -524,8 +526,10 @@ describe("INT-03: AgentPhone linked-run lifecycle through public APIs", () => {
       from: phone,
       body: "start on my phone",
     });
-    const admittedSms =
-      await findAgentphoneChatEventByPromptFixture("start on my phone");
+    const admittedSms = await findAgentphoneChatEventByPromptFixture({
+      userId: actor.userId,
+      prompt: "start on my phone",
+    });
     if (!admittedSms) {
       throw new Error("Expected admitted AgentPhone SMS input event");
     }
@@ -656,9 +660,10 @@ describe("INT-03: AgentPhone linked-run lifecycle through public APIs", () => {
       from: phone,
       body: "second queued prompt",
     });
-    const secondEvent = await findAgentphoneChatEventByPromptFixture(
-      "second queued prompt",
-    );
+    const secondEvent = await findAgentphoneChatEventByPromptFixture({
+      userId: actor.userId,
+      prompt: "second queued prompt",
+    });
     if (!secondEvent) {
       throw new Error("Expected pending AgentPhone queue item");
     }
@@ -917,9 +922,10 @@ describe("INT-03: AgentPhone linked-run lifecycle through public APIs", () => {
         },
       ],
     });
-    const admittedGroup = await findAgentphoneChatEventByPromptFixture(
-      "summarize this thread",
-    );
+    const admittedGroup = await findAgentphoneChatEventByPromptFixture({
+      userId: actor.userId,
+      prompt: "summarize this thread",
+    });
     expect(admittedGroup).toMatchObject({ eventId: expect.any(String) });
     if (!admittedGroup) {
       throw new Error("Expected admitted AgentPhone group input event");
--- a/turbo/apps/api/src/signals/routes/__tests__/internal-callbacks-teams.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/internal-callbacks-teams.test.ts
@@ -646,8 +646,10 @@ describe("Teams chat callbacks", () => {
       activityId: "activity-queue-params-second",
       text: queuedPrompt,
     });
-    const queuedParams =
-      await findPendingChatEventByPromptFixture(queuedPrompt);
+    const queuedParams = await findPendingChatEventByPromptFixture({
+      userId: teams.actor.userId,
+      prompt: queuedPrompt,
+    });
     expect(queuedParams).toMatchObject({
       eventId: expect.any(String),
     });
@@ -731,7 +733,10 @@ describe("Teams chat callbacks", () => {
       text: queuedPrompt,
     });
     await expect(
-      findPendingChatEventByPromptFixture(queuedPrompt),
+      findPendingChatEventByPromptFixture({
+        userId: teams.actor.userId,
+        prompt: queuedPrompt,
+      }),
     ).resolves.toMatchObject({ eventId: expect.any(String) });
 
     await seedOrgMetadata({
@@ -795,8 +800,10 @@ describe("Teams chat callbacks", () => {
       text: queuedPrompt,
       omitRecipient: true,
     });
-    const queuedParams =
-      await findPendingChatEventByPromptFixture(queuedPrompt);
+    const queuedParams = await findPendingChatEventByPromptFixture({
+      userId: teams.actor.userId,
+      prompt: queuedPrompt,
+    });
     if (!queuedParams) {
       throw new Error("Expected queued Teams bot fallback event");
     }
--- a/turbo/apps/api/src/signals/routes/__tests__/zero-browser.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/zero-browser.test.ts
@@ -496,9 +496,15 @@ describe("zero browser route", () => {
     acceptBrowserUseCdpSessions(providerIds);
     const providerCreateBodies: unknown[] = [];
     const deletedProfiles: string[] = [];
+    const stoppedProviderIds: string[] = [];
     let profileCreates = 0;
     let providerCreates = 0;
-    let providerStops = 0;
+    // The reconcile route is global, so count only this test's own instances.
+    const providerStops = () => {
+      return stoppedProviderIds.filter((stopped) => {
+        return (providerIds as readonly string[]).includes(stopped);
+      }).length;
+    };
     server.use(
       http.post(`${BROWSER_USE_API_URL}/profiles`, async ({ request }) => {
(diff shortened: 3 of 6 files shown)


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 2ca2bf25a783d2d2b5b3ecd62b77bd1bcda2cc28
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Test-only change that removes two shared-database isolation leaks in `apps/api`: the three prompt-based chat-event fixtures now take `{ userId, prompt }` and `innerJoin chat_threads` on the caller's own user instead of scanning the whole `chat_events` table, and the `zero-browser` concurrent-profile test asserts on its own stopped provider ids instead of the global `/api/cron/browsers/reconcile` counters. Production code, business assertions, and test coverage are unchanged.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic, no wrong assumptions | Pass | `chat_events.chat_thread_id` is `notNull()` with an FK to `chat_threads.id` (`turbo/packages/db/src/schema/chat-event.ts:83-90`), so the new `innerJoin` cannot silently drop a legitimate row; the prior `WHERE` predicates are preserved verbatim inside `filter`. |
| `REVIEW.md` | Security: no injected input, no missing auth surface | N/A | Test fixtures and test files only; no route, auth path, or user input handling changed. |
| `REVIEW.md` | Type safety: no `any`, no over-broad types | Pass | New `filter: SQL \| undefined` is precisely typed via `import { type SQL } from "drizzle-orm"`; no `any` in the diff. |
| `REVIEW.md` | Maintainability: no duplicated logic, comments explain WHY | Pass | Three near-identical query bodies collapse into `findOwnedChatEventByPrompt`; both new comments state why (shared DB, global reconcile route), not what. |
| `REVIEW.md` | Testing coverage for the commit type | Pass | `test:` commit; no production behavior changed, so no new regression test is required. All 14 call sites keep their original assertions. |
| `REVIEW.md` | Feature regression guard / feature-switch gating | N/A | No user-facing feature, UI, API, or flow is added, replaced, or rerouted. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, migration, contract, or `jsonb().$type<>()` change; the fixture only reads `userMessage`. |
| `docs/bad-smell.md` | Zero `any`, zero lint/type suppressions | Pass | No `any`, no `eslint-disable`/`ts-expect-error`/`prettier-ignore` anywhere in the six changed files. |
| `docs/bad-smell.md` | No dynamic `import()`, no hardcoded URLs/config | Pass | Only static imports; no URL or environment literal added. |
| `docs/bad-smell.md` | Fail fast, no recovery/fallback logic | Pass | `row ?? null` is the pre-existing not-found return of a `find*` helper, not a fabricated default; callers still `throw` on null. |
| `docs/fallback.md` | Declare every fallback in the summary, or `Fallbacks: none` | Pass (nit) | The diff introduces, keeps, and removes no fallback; see `### Fallbacks` below. Nit: the summary has no explicit `Fallbacks: none` line, which §9 asks for. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | No tombstone test added; the removed `checked/stopped/errors` assertions are replaced by an equivalent own-instance assertion, not deleted. |
| `docs/testing.md` | Mock at the boundary only — no `vi.mock("../…")` | Pass | No `vi.mock` added or changed; `zero-browser` keeps MSW handlers at the Browser Use HTTP boundary. |
| `docs/testing.md` | Real database, no `globalThis.services.db` mocking | Pass | The fixture keeps the real `db()` and only narrows its `WHERE`. |
| `docs/testing.md` | Isolate test data with unique user/org ids, not by mocking the DB layer | Pass | This is the change: lookups are now keyed on the caller's own random `userId` (`actor.userId`, `fixture.userId`, `teams.actor.userId`, `secondActor.userId`). |
| `docs/testing.md` | No fake timers | Pass | `zero-browser` keeps `mockNow(...)`; no `vi.useFakeTimers()` introduced. |
| `docs/testing.md` | Test behavior, not mock call patterns | Pass | `providerStops()` derives from real MSW-intercepted `PATCH /browsers/:id` requests and is asserted as an outcome count, not `toHaveBeenCalled()`. |
| `docs/testing/anti-patterns.md` | AP-1/AP-2/AP-4/AP-5: no mock-call assertions, no `fetch` stubbing, no internal mocks, no fake timers | Pass | The new counter replaces a raw `providerStops += 1` with the same observation source (MSW handler); no new mock surface. |
| `docs/testing/patterns.md` | MSW handler overrides via `server.use`, no manual server lifecycle | Pass | The changed handler stays inside the existing `server.use([...])` block; no `listen`/`close` added. |
| `docs/testing/api-testing.md` | Do not add new DB reads/writes to API tests; use the endpoint surface | Pass | No new boundary crossing: the three fixtures already read `chat_events` directly, and this PR narrows those existing reads rather than adding one. Test files still call fixtures, not `db()`. |
| `docs/testing/testing-external-behavior.md` | Tests should construct and verify through the external interface | Pass (pre-existing debt) | The prompt lookups remain internal DB reads inherited from before this PR; scoping them to the caller's user reduces, but does not remove, that coupling. |
| `.claude/skills/ccstate/SKILL.md` | ccstate command/computed/detach practices in `apps/api` | N/A | No signal, `computed`, command, or `detach()` usage is added or changed; the diff is limited to test files and a DB test fixture. |
| `docs/deployment-compatibility.md` | Cross-version compatibility for deployable/persisted boundaries | N/A | Nothing deployable changes: no contract, migration, runner protocol, queue payload, or persisted shape is touched. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

#### Nits (non-blocking)

- `turbo/apps/api/src/signals/routes/__tests__/zero-browser.test.ts:502` — `(providerIds as readonly string[]).includes(stopped)` looks unnecessary: `providerIds` is `readonly [string, string]`, whose `includes` already accepts a `string`. Dropping the assertion would keep the helper cast-free.
- `turbo/apps/api/src/test-fixtures/chat-events.ts:findOwnedChatEventByPrompt` — the query still has no `ORDER BY`, so `rows.find(...)` picks an arbitrary row if one user legitimately has two events with identical prompt text. No current call site does that (each test creates its own actor and uses distinct prompts), so this is latent only.
- The PR's "Not in this PR" note is accurate: `zero-browser.test.ts` still has ~20 `checked`/`stopped`/`errors` assertions on the global reconcile route (lines 1100, 1348-1375, 1547, 1699-1745, 2089-2247). A follow-up issue would make that remaining isolation gap trackable.

### Fallbacks

- None. The diff introduces, keeps, and removes no fallback or compatibility behavior: no `??`/`||` default on a contract-owned field, no branch for an old client/runner/route/persisted shape, no feature-switch-gated behavior, and no test asserting removed behavior stays removed. `row ?? null` and `candidate.userMessage?.parts` are the helper's not-found result and a genuinely nullable `jsonb` column, not compatibility paths. Nit: the PR summary should state `Fallbacks: none` explicitly per `docs/fallback.md` §9.

### Testing

- Coverage: Adequate. Every changed call site keeps its original assertion; the `zero-browser` case keeps the same numbers (2 stops after thread deletion, 1 more on reconcile, no further stop on the second reconcile) and the same idempotence claim, now expressed through this test's own `providerIds` and `profileIds`. Reconcile-response coverage (`checked`/`stopped`/`errors`) is retained elsewhere in the same file, so nothing is lost by dropping it from this one case.
- Conventions: Compliant. Real database, MSW at the external boundary, no internal `vi.mock`, no fake timers, per-test random actors for isolation.
- Reproduction evidence in the PR is consistent with the change: the previously failing second `--shard=4/8` pass against one database is exactly the cross-worker read the ownership predicate now prevents.

PR #25739, opened 2026-08-07
test(api): support dev-seeded pi skill archives

### LINKED ISSUE #25732: test(api): support dev-seeded skills in pi edge integration tests
## Problem and evidence

`pi-edge-loop.test.ts` installs an in-memory S3 mock for Pi launch storage. The mock serves objects uploaded during the test or explicitly registered through `storageObjects.put()`, but it does not provide archives for non-empty default system Skill versions already present in the database.

This creates an environment-dependent failure:

- Turbo API CI runs migrations only, so unresolved default Skill mounts are empty and the suite passes.
- The documented local setup runs `scripts/prepare.sh`, including `db:dev-seed`, so default Skills can resolve to non-empty pinned storage versions whose archives exist in R2 but not in the in-memory mock.
- Running the focused suite on the current prepared database failed 14 of 15 cases. `S3 object body is empty` was the primary error; timeouts and state/assertion failures were downstream effects.
- Current local storage IDs and version hashes do not all match `dev-seed-skill-volumes.json`, because local Skill heads can later be synchronized. A fix cannot rely on one committed hash or assume the prepared database remains identical to the generated seed file.

The production Pi loader is behaving correctly: it reads each exact storage version pinned in the run snapshot and fails closed when a version marked non-empty has no archive body.

## Decision

Fix the test fixture only. On an unregistered archive request, synthesize a valid test Skill archive only after an exact database lookup proves that the requested version S3 key belongs to one of the configured official default system Skill storage names. Cache only that exact key and preserve the current missing behavior for every unmatched object.

This remains one independently reviewable PR.

1. Add a typed storage test-fixture read that joins `storage_versions` to `storages` and matches both an exact version S3 key and an allowed storage-name list.
2. Derive the allowlist from `SEED_SKILLS`, `GOAL_SKILL_NAME`, and the same core GitHub/storage-name helpers used by production; do not copy storage IDs, version hashes, or generated seed metadata.
3. In the Pi in-memory S3 mock, retain existing uploaded and explicitly registered objects as the first priority.
4. For an otherwise missing exact `archive.tar.gz` key, validate the suffix-free version key through the fixture read. On a match, create and cache a minimal valid archive with `SKILL.md` frontmatter using that configured Skill's name.
5. Leave unknown `GetObject` requests bodyless and unknown `HeadObject` requests as 404. Do not add an `__system__` prefix fallback or real R2 access.

Validating the requested pinned version instead of pre-registering only current heads also avoids a setup-time head-change race and supports a run that pinned a retained prior version.

## Alternatives considered

- **Pre-register current heads during fixture setup:** simpler, but has a time-of-check/time-of-use gap and only covers heads observed at setup.
- **Use `dev-seed-skill-volumes.json`:** does not cover locally synchronized or newly populated default versions and is already inconsistent with observed local identities.
- **Run the file only against a migration-only database:** hides the state that exposes the bug and does not make the documented local setup valid.
- **Return a generic archive for missing system-prefixed keys:** can hide unrelated storage defects and violates the exact-key requirement.
- **Read real R2:** adds network, credentials, retention, and mutable external state to an integration test.

## Risks and mitigations

- **Overly permissive fixture:** require both the exact version key and a storage name derived from the configured official default list.
- **Synthetic archive differs from published content:** this suite covers storage loading and Pi execution semantics; Skill synchronization tests cover publishing integrity.
- **Loader diagnostics or name collisions:** generate frontmatter using the exact configured Skill name associated with the matched storage name.
- **Test cost:** cache each validated archive by exact key, so each version requires at most one lookup and archive build per fixture.
- **Production regression:** no production loader, schema, API, protocol, seed data, or configuration changes are in scope.

## Acceptance criteria

- The Pi edge-loop integration suite passes after `scripts/prepare.sh` using the normal prepared local database.
- The same suite passes against a fresh database with migrations only.
- The existing integration coverage still exercises immutable Pi Skill snapshots, exact pinned storage mounts, archive extraction, prompt discovery, and tool readability of a pinned workflow Skill.
- Only exact configured default system Skill version keys receive synthesized test archives; unrelated missing S3 objects remain missing.
- No production storage behavior changes.
- Applicable formatting, lint, type, Knip, test, and pre-commit checks pass.

## Delivery notes

Expected source changes are limited to:

- `turbo/apps/api/src/test-fixtures/storage.ts`
- `turbo/apps/api/src/signals/routes/__tests__/pi-edge-loop.test.ts`

No migration, rollout, documentation update, or multi-PR decomposition is required.

seven332:
## Summary

- Materialize default system Skill archives in the Pi edge S3 test mock from the exact persisted storage-version identity.
- Keep explicit test uploads authoritative and preserve missing-object behavior for unknown S3 keys.
- Add a scoped test fixture lookup for storage-version S3 keys that product APIs intentionally do not expose.

## Scope

This is test-only infrastructure. It does not change production storage behavior, database schema, migrations, or seed data.

## Validation

- `pnpm -F api exec vitest run src/signals/routes/__tests__/pi-edge-loop.test.ts --maxWorkers=1 --silent=passed-only` (15 tests, dev-seeded database)
- The same focused suite against a migration-only temporary database with no Skill rows (15 tests)
- `pnpm format`
- `pnpm -F api lint`
- `pnpm -F api check-types`
- `pnpm knip --workspace api`
- Pre-commit hook: repository Prettier, Knip, and full Turbo typecheck

Closes #25732

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/pi-edge-loop.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/pi-edge-loop.test.ts
@@ -11,6 +11,9 @@ import {
 } from "@vm0/api-contracts/contracts/runners";
 import { webhookPiTranscriptContract } from "@vm0/api-contracts/contracts/webhooks";
 import { FeatureSwitchKey } from "@vm0/core/feature-switch-key";
+import { parseGitHubTreeUrl, resolveSkillRef } from "@vm0/core/github-url";
+import { getSkillStorageName } from "@vm0/core/storage-names";
+import { GOAL_SKILL_NAME, SEED_SKILLS } from "@vm0/core/zero-seed-skills";
 import { HttpResponse, http } from "msw";
 import { v5 as uuidv5 } from "uuid";
 import { describe, expect, it, onTestFinished } from "vitest";
@@ -51,6 +54,7 @@ import {
   createAgentComposeFixture,
   readAgentComposeByIdFixture,
 } from "../../../test-fixtures/agent-composes";
+import { readSystemStorageVersionNameByS3KeyFixture } from "../../../test-fixtures/storage";
 
 const context = testContext();
 const bdd = createBddApi(context);
@@ -75,6 +79,7 @@ const CODEX_WHAM_USAGE_URL = "https://chatgpt.com/backend-api/wham/usage";
 const CODEX_OAUTH_TOKEN_URL = "https://auth.openai.com/oauth/token";
 const OPENAI_COMPLETIONS_URL = "https://api.openai.com/v1/chat/completions";
 const AGENT_DISPLAY_NAME = "Pi edge integration agent";
+const STORAGE_ARCHIVE_SUFFIX = "/archive.tar.gz";
 const PI_EDGE_USAGE_OBSERVATION_IDEMPOTENCY_NAMESPACE =
   "1b7c07b8-01bc-4ae2-ac5c-ef5ca9f72683";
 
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #25739

**Reviewed head:** `294e011f91c5f9b6161a79e3dc8437d062e82362`

### Summary

The test-only S3 fallback is narrowly bound to exact persisted system Skill storage-version keys and canonical official Skill identities. Explicit uploads still win, unknown objects remain missing, and no production API, schema, seed, or deployment contract changes.

### Findings

No findings. A self-review edge case involving full GitHub Skill references was fixed on this head by deriving archive names from `parsed.skillName`.

### Testing

- Pi edge integration suite: 15/15 with a prepared dev-seeded database.
- Pi edge integration suite: 15/15 with a migration-only database containing no Skill rows.
- API formatting, lint, type checking, Knip, and repository pre-commit checks passed.
- CI's migration-only setup does not execute the dev-seeded lookup path, which explains the fixture-line Codecov notice; that state was validated locally without adding an implementation-detail unit test.

### Verdict

LGTM

PR #25763, opened 2026-08-08
fix(api): prevent indexed chat search fallback

e7h4n:
## Summary

- keep `chatSearchIndex` requests exclusively on the `chat_event_search_docs` projection
- return no matches for keywords the bigram index cannot represent instead of scanning `chat_events` with ILIKE
- preserve the legacy ILIKE behavior when the feature switch is disabled

## Verification

- `pnpm format`
- `pnpm knip`
- `TZ=UTC DATABASE_URL=postgresql://postgres@localhost:5432/vm0_test pnpm --filter api exec vitest run src/signals/routes/__tests__/chat-threads.bdd.test.ts -t 'CHAT-01 chat search index'` (3 passed)

## Fallbacks

- Removed `zero-chat-thread.service.ts:zeroChatSearch` fallback from a non-indexable projection query to the legacy ILIKE scan. Surface: none — `chatSearchIndex` is a non-GA feature switch that remains disabled by default. Evidence: the enabled branch now produces an always-false SQL condition for non-indexable keywords, while the disabled branch explicitly selects the legacy condition. No rollout window or follow-up cleanup is required.

### DIFF
--- a/turbo/apps/api/src/lib/chat-search-bigram.ts
+++ b/turbo/apps/api/src/lib/chat-search-bigram.ts
@@ -56,7 +56,7 @@ export function chatSearchIndexText(text: string): string {
  * one CJK run are chained with `<->` (adjacent-phrase, i.e. exact substring
  * semantics) and groups are combined with `&`. Returns null when the keyword
  * cannot be answered from the bigram index — a single-character CJK run or a
- * keyword with no indexable token — so the caller can fall back to ILIKE.
+ * keyword with no indexable token.
  */
 export function chatSearchBigramTsquery(keyword: string): string | null {
   const groups = tokenGroups(keyword);
--- a/turbo/apps/api/src/signals/routes/__tests__/chat-threads.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/chat-threads.bdd.test.ts
@@ -2903,7 +2903,7 @@ describe("CHAT-01 chat search index", () => {
 
     const threadId = await sendNoCreditMessage(owner, {
       agentId: agent.agentId,
-      prompt: "今天天气很好 vercel 部署完成",
+      prompt: "今天天气很好，vercel 部署完成",
     });
     await sendNoCreditMessage(peer, {
       agentId: peerAgent.agentId,
@@ -2928,13 +2928,16 @@ describe("CHAT-01 chat search index", () => {
       expect(found.results).toHaveLength(1);
       expect(found.results[0]?.chatThreadId).toBe(threadId);
       expect(found.results[0]?.matchedMessage.content).toBe(
-        "今天天气很好 vercel 部署完成",
+        "今天天气很好，vercel 部署完成",
       );
     }
 
-    // A single CJK character has no bigram form and falls back to ILIKE.
+    // Unindexable keywords stay on the index path instead of falling back to
+    // ILIKE, so neither a single CJK character nor punctuation can match.
     const singleChar = await chat.searchChat(owner, "好");
-    expect(singleChar.results).toHaveLength(1);
+    expect(singleChar.results).toStrictEqual([]);
+    const punctuation = await chat.searchChat(owner, "，");
+    expect(punctuation.results).toStrictEqual([]);
 
     // Word tokens match whole words only under the index path.
     const partialWord = await chat.searchChat(owner, "verce");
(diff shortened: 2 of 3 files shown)


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: e144faa431f0a9978a7066d00dbbdf4517cf30c3
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

This PR makes the `chatSearchIndex` branch exclusive: indexable queries use `chat_event_search_docs`, while keywords the projection cannot represent return no matches instead of falling back to the legacy `ILIKE` scan. The disabled feature-switch branch retains the existing legacy behavior, and the route-level regression test covers both paths.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness, security, type safety, and maintainability across the full diff | Pass | `zero-chat-thread.service.ts` now selects one explicit keyword condition; the constant-false SQL predicate is static, the indexed subquery retains user/org scoping, and no unsafe casts, suppressions, secrets, or new interfaces are introduced. |
| `REVIEW.md` | A `fix:` change needs regression coverage and must preserve the scoped existing flow | Pass | `chat-threads.bdd.test.ts` exercises the public search route with the switch enabled and disabled, including the exact single-CJK and punctuation cases that previously reached `ILIKE`. |
| `docs/docs.md` | Load every practice guide relevant to the API, ccstate, fallback, deployment, and test surfaces | Pass | The production, fallback, ccstate/effect/cache, deployment-compatibility, and API testing guides indexed for these changed files were reviewed from `main`. |
| `docs/bad-smell.md` | Avoid `any`, suppressions, dynamic imports, hidden recovery, and unnecessary error paths | Pass | The diff adds none of those patterns and removes the implicit `indexCondition ?? ILIKE` recovery path. |
| `docs/testing.md` | Prefer integration coverage with the real database and mocks only at external boundaries | Pass | The changed test stays in the API BDD suite, drives the real route and projection cron, and adds no database or internal-module mock. |
| `docs/fallback.md` | Non-GA feature switches need no compatibility fallback; removals must be declared and evidenced | Pass | `chatSearchIndex` remains disabled by default; the PR's `Fallbacks` section declares the removed fallback and points to the now-explicit enabled/disabled branches. The assertions cover current product behavior rather than a retired route or parser tombstone. |
| `docs/effect.md` | Use `computed` for consumed async reads and keep it free of unrelated stateful effects | Pass | `zeroChatSearch` remains an async `computed` that reads `db$`; the change only chooses the query predicate and introduces no writes, resource acquisition, or lifecycle work. |
| `docs/cache.md` | Do not add render-time allocation, manual caches, duplicated state, or lifecycle leaks | N/A | No React, cache, state ownership, resource, ref, or teardown surface changes in this API-only diff. |
| `.claude/skills/ccstate/SKILL.md` | Keep ccstate access inside callbacks and avoid escaped accessors, detached work, or manual fetch state | Pass | The changed `computed` keeps `get(db$)` local and adds no command composition, detached promise, accessor escape, or mutable loading state. |
| `docs/deployment-compatibility.md` | Preserve old/new compatibility only at independently deployed or persisted boundaries | N/A | No API contract, request/response schema, database shape, queue payload, or runner protocol changes; the behavioral cutover is contained behind a non-GA switch. |
| `docs/testing/patterns.md` | Use the standard route integration pattern and verify behavior rather than internals | Pass | Setup, switch enablement, projection, and search verification all go through existing API helpers backed by `setupApp`. |
| `docs/testing/anti-patterns.md` | Avoid internal mocks, fake timers, mock-call-only assertions, and service-level testing when a route exists | Pass | The added coverage uses no new mocks or timers and asserts returned search results from the route. |
| `docs/testing/api-testing.md` | Exercise API behavior through contracts/routes with real internal services and database | Pass | The test enables the switch through `zeroFeatureSwitchesContract`, invokes the projection cron contract, and verifies `chatThreadsContract` search behavior. |
| `docs/testing/testing-external-behavior.md` | Construct and observe the case through the external API surface | Pass | The matching prompt is created through chat send, and the regression is observed through chat search; no internal row mutation or assertion was added. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Fallbacks

- `turbo/apps/api/src/signals/services/zero-chat-thread.service.ts:zeroChatSearch` / `chatSearchIndexCondition`: removes the fallback from a non-indexable projection query to the legacy `ILIKE` scan. Surface: none — `chatSearchIndex` is a non-GA switch that remains disabled by default. Removal condition: already satisfied by the non-GA containment; no rollout window or follow-up is required. Declared in PR summary: Yes. Verdict: Justified.
- Retained version-migration fallbacks: None.

### Testing

- Coverage: Adequate. The route-level regression covers pre-projection behavior, indexed CJK/word matches, single-character CJK, punctuation, and the switch-disabled legacy path.
- Conventions: Compliant. The PR reports the targeted API test passing 3/3; all currently published PR checks are green, including every `test-api` shard, lint, type checks, CodeQL, Semgrep, and secrets scanning.

PR #25769, opened 2026-08-08
feat: show model changes in chat history

e7h4n:
## Summary

- move the selected-model notice from the Composer strip to the bottom of the chat message stream and include the selected model name
- insert run-style model change dividers before later user messages when adjacent runs both record different models
- reuse the finished-run divider visual, localize both notices, and keep the existing feature switch boundary

## Test plan

- `pnpm -F @vm0/app exec vitest run src/views/zero-page/__tests__/chat-run-queue.test.tsx --maxWorkers=2 --silent=passed-only` (23 tests)
- `pnpm -F @vm0/app lint`
- `pnpm knip --workspace @vm0/app`
- Prettier 3.8.1 check on all changed files

## Notes

- no local dev server or full Vitest suite was run; the PR pipeline will cover the broader checks

### DIFF
--- a/turbo/apps/platform/src/i18n/locales/de-DE/common.json
+++ b/turbo/apps/platform/src/i18n/locales/de-DE/common.json
@@ -1467,7 +1467,6 @@
       "loadingGoal": "Ziel wird geladen...",
       "message_one": "{{count}} Nachricht",
       "message_other": "{{count}} Nachrichten",
-      "modelChangeAppliesNextRun": "Ihr ausgewähltes Modell wird bei Ihrer nächsten Ausführung verwendet.",
       "openGoalDetails": "Zieldetails öffnen",
       "pendingAutomationEvent": "Ausstehendes Automatisierungsereignis",
       "queuedMessage": "Nachricht in der Warteschlange",
@@ -1516,7 +1515,9 @@
       "justNow": "soeben",
       "keepGoing": "Fortsetzen",
       "keepGoingAt": "Fortsetzen · {{timestamp}}",
+      "modelChangedTo": "Modell wurde auf {{model}} geändert",
       "queueEllipsis": "In Warteschlange...",
+      "selectedModelAppliesAfterCurrentRun": "Ihr ausgewähltes {{model}}-Modell wird nach dieser Ausführung wirksam",
       "thinking": {
         "assembling": "Teile werden zusammengesetzt...",
         "brewing": "Eine Antwort wird vorbereitet...",
--- a/turbo/apps/platform/src/i18n/locales/en-US/common.json
+++ b/turbo/apps/platform/src/i18n/locales/en-US/common.json
@@ -1467,7 +1467,6 @@
       "loadingGoal": "Loading goal...",
       "message_one": "{{count}} message",
       "message_other": "{{count}} messages",
-      "modelChangeAppliesNextRun": "Your selected model will take effect on your next run.",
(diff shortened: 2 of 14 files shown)


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 90b0f7419f4807b56be14614ad6b07c74738547d
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

This moves the selected-model notice from the Composer strip into the observed chat message stream, includes the selected model's display name, and derives historical model-change dividers only when two adjacent runs both record different models. The active notice, historical divider, and finished-run marker now share the requested single-line label-plus-rule treatment, while the existing non-GA feature-switch boundary remains intact.

### Review Criteria

| Source | Applicable rule | Result |
| --- | --- | --- |
| `REVIEW.md` | Correctness, type safety, feature containment, regression coverage, and integration coverage for a `feat:` PR | Pass: the notice conditions, placement, feature gate, and adjacent-run cases are covered without broadening the contract surface. |
| `docs/bad-smell.md` | Avoid unsafe casts, hidden configuration defaults, duplicated logic, and silent fallback paths | Pass: the projection is explicit and typed, and the three marker variants share one renderer. |
| `docs/event-sourcing.md` | Derive UI from the authoritative event projection and do not add rollback-only behavior | Pass: historical changes are projected from visible rendered chat events and keyed to the next run's input event ID. |
| `docs/effect.md` and `docs/cache.md` | Keep render pure; avoid synchronization effects, manual caches, and lifecycle mismatches | Pass: model-change calculation is a bounded pure derivation and no effect, resource, or cache lifecycle was added. |
| `.claude/skills/ccstate/SKILL.md` | Consume existing signals through React hooks; do not allocate signals or issue commands during render | Pass: the view reads the existing model-selection, running-model, and feature-switch signals without adding stateful signal machinery. |
| `docs/testing.md` and `docs/testing/patterns.md` | Cover features through integration behavior and production bootstrap paths | Pass: page-level tests use `detachedSetupPage` and cover placement, gating, different models, matching models, and missing-model documents. |
| `docs/testing/anti-patterns.md`, `docs/testing/app-testing.md`, and `docs/testing/testing-external-behavior.md` | Test the rendered page through boundary mocks rather than internal component or Store control | Pass: setup stays at the page boundary and assertions exercise visible text/order; the focused marker-structure assertion protects the accepted visual regression. |
| `docs/fallback.md` | Add no compatibility fallback for a non-GA feature and declare any fallback that remains | Pass: no compatibility fallback is introduced; `ChatNextRunModelNotice` is containment for the non-GA UI, not a cross-version fallback. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Fallbacks

- None. The existing `ChatNextRunModelNotice` feature gate contains the non-GA presentation change and does not bridge old/new deployed versions.

### Testing

- Coverage: Adequate. The PR adds page-level integration coverage for the active notice and adjacent-run divider, including feature-off, equal-model, and either-model-missing cases; all required CI gates are green at `90b0f7419f4807b56be14614ad6b07c74738547d`.
- Conventions: Compliant with the selected platform, ccstate, event-projection, and testing practices.
- Preview verification: the marker DOM contains only the label plus its right-hand rule; inserting/removing the notice kept a bottom-following thread at `0px` bottom distance, and a scrolled-up thread retained its visible anchor with `0px` delta through notice removal and run completion.

PR #25770, opened 2026-08-08
feat(api): archive chat events into immutable full-thread r2 snapshots

e7h4n:
## Summary

Shadow phase (PR 1) of the chat-event R2 snapshot compaction design:

- New `chat_event_snapshots` table: one immutable fact row per uploaded archive object. `is_head` plus a partial unique index give each thread at most one head without a separate head table; thread deletion cascades row deletion (R2 objects are left for a later GC).
- New `/api/cron/snapshot-chat-events` cron (every 30 minutes): picks threads whose search-projection watermark is ahead of their head snapshot, least-recently-active first, and rebuilds one full-thread archive per thread.
- Full snapshots, no deltas: every object covers seq `(0, W]` where `W` is the thread's `chat_event_search_watermarks.indexed_seq_id`, so archived events are always queryable before their Postgres payloads ever become reclaimable.
- Rebuilds replay the parent object (verified against the sha256 embedded in its key) and append only the Postgres tail beyond the parent's `last_seq_id`. The archived range is never re-read from Postgres, which is what later allows payload reclaim; every rebuild also re-verifies the full history hash for free.
- Objects are content-addressed gzip NDJSON in the existing storages bucket: `chat-events/{threadId}/{lastSeqId}-{sha256}.ndjson.gz`, written with the idempotent immutable-put helper.
- The archive line shape is `archive_schema_version` 1 and lists every `chat_events` column explicitly; all event types are archived.
- Publishing is a two-statement transaction (demote the expected parent, insert the new head). Expected races — a concurrent writer or a thread deleted mid-pass — skip the thread and at worst leave an orphaned object; a parent object failing its hash check fails the run loudly.

Out of scope, per the agreed plan: Postgres payload reclaim, readers over head + tail, R2 object GC. Until reclaim ships, changing the archive shape only costs a version bump because every rebuild rewrites the whole object.

## Testing

- New route integration tests cover cron auth, first-generation archive content (key hash, NDJSON invariants, idempotent re-run), a second-generation rebuild that byte-identically replays the parent object and appends only the tail, and a loud 500 when a parent object no longer matches its content hash plus recovery after the object is restored.
- `vercel-crons`, app-factory, api-contracts, and db migration-consistency suites pass locally; lint, `check-types`, and knip are clean on the touched workspaces.

## Fallbacks

None. The change is an additive shadow-phase table plus a new cron; nothing in the diff bridges old and new deployed versions. The `?? 0` / `COALESCE(last_seq_id, 0)` on the head watermark models the legitimate first-generation state (no head row from the LEFT JOIN), not a contract-prevented value.


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 74655c93a642e689543ed9e0e2428c46a61dcb92
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Shadow phase 1 of chat-event archival: adds the `chat_event_snapshots` table (head tracked by `is_head` under a partial unique index), and a 30-minute cron that rebuilds one full-thread gzip NDJSON archive per lagging thread — replaying the parent object verified by the sha256 embedded in its key, appending only the Postgres tail up to the search watermark — then publishes the new head in a demote-and-insert transaction. Content-addressed objects go to the existing storages bucket under `chat-events/`. No payload reclaim, no reader changes.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness: races and edge cases | Pass | Head swap is a demote-by-expected-parent + insert transaction; the partial unique index rejects double heads; concurrent-writer and deleted-thread races skip via `settle` + `isForeignKeyViolation`/`isUniqueViolation`; parent hash mismatch throws loudly (`cron-snapshot-chat-events.service.ts`) |
| `REVIEW.md` | Security: new endpoints authenticated | Pass | `/api/cron/snapshot-chat-events` guarded by `hasValidCronSecret$`; 401 covered in the route test |
| `REVIEW.md` | `feat:` commit has integration tests | Pass | `cron-snapshot-chat-events.test.ts` exercises the real route slice against real Postgres |
| `REVIEW.md` | DB/JSONB review gate | Pass | Additive `CREATE TABLE` migration `0862`; no JSONB `$type` shape changed; a brand-new table has no old rows to normalize |
| `REVIEW.md` | Feature regression guard | Pass | Write-only background cron; no existing UI/API/flow is changed or rerouted |
| `REVIEW.md` | User-facing features behind a `FeatureSwitchKey` | N/A | No user-facing surface: the cron only writes shadow archive objects and rows that nothing reads yet |
| `docs/bad-smell.md` | No `any`, suppressions, dynamic imports, hardcoded config | Pass | Static imports only; config via `env()`/`optionalEnv()`; no suppression comments in the diff |
| `docs/bad-smell.md` | Fail-fast error handling; no over-engineered try/catch | Pass | Single `settle` scoped to two identified PG error codes for real concurrent states; every other failure propagates (500 covered by test) |
| `docs/fallback.md` | No `??`/`||` on contract-guaranteed values | Pass | `?? 0`/`COALESCE(last_seq_id, 0)` guard the LEFT-JOIN null head — a reachable first-generation state, not a `NOT NULL` column |
| `docs/fallback.md` | Every fallback declared in the PR summary | Pass | Diff introduces no version-compatibility fallback; PR summary carries an explicit `Fallbacks: none` section |
| `docs/fallback.md` | No tombstone/negative tests against removed behavior | Pass | No removed-behavior assertions in the diff |
| `docs/event-sourcing.md` | Persistent events authoritative; no optimistic cleanup logic | N/A | Backend-only archival of the persistent `chat_events` stream; no optimistic projections touched |
| `docs/deployment-compatibility.md` | Migration backward-compatible in both skew directions | Pass | Additive table: the pre-migration API never references it, and the old API keeps ignoring it after migration; plain transactional DDL |
| `docs/deployment-compatibility.md` | Cross-version protocol/payload safety | Pass | New cron path, route registration, and contract ship in one API artifact; contract change is additive |
| `.claude/skills/ccstate/SKILL.md` | Signals discipline (no `detach()` in signals; command/computed idiom) | Pass | Service is a `command` using `set(writeDb$)` and `get(...)` on S3 computeds, mirroring existing cron services; no `detach()` |
| `docs/testing.md` | Integration-only, boundary mocks, real infrastructure | Pass | Tests run the real Hono app and real database; only `@aws-sdk/client-s3` is mocked, via the central registry |
| `docs/testing/api-testing.md` | Route tests through `setupApp` + ts-rest contract with `context.mocks` | Pass | All three tests call the contract client over the declared route slices |
| `docs/testing/testing-external-behavior.md` | Construct and verify through external surfaces, not the DB | Pass | State built via agent-create/send/cron APIs; assertions on cron responses and captured S3 boundary traffic; no direct DB access in the test |
| `docs/testing/patterns.md` | Use the shared mock registry for external services | Pass | `context.mocks.s3.send.mockImplementation` implements an in-memory R2 keyed to `chat-events/` |
| `docs/testing/anti-patterns.md` | No fake timers, artificial sleeps, or mock-call-only assertions | Pass | No timer mocking; assertions verify response bodies and decompressed object bytes, not mock call counts |

### Findings

#### Critical (P0)

- None

#### High Priority (P1)

- None

#### Notes (non-blocking)

- `expectArchiveInvariants` keeps a conditional `expect` inside the line loop (tolerated `vitest(no-conditional-expect)` warning); iterating from index 1 would assert unconditionally.
- A lost publish race intentionally leaves an orphaned R2 object; object GC is explicitly deferred to a follow-up per the PR plan.

### Fallbacks

- None. The diff adds a new table, contract, route, and cron with no compatibility branch between deployed versions. The `?? 0` head-watermark default models the legitimate no-head (first-generation) LEFT-JOIN state, and the PR summary declares `Fallbacks: none`.

### Testing

- Coverage: Adequate — auth (401), first-generation archive content (key sha256, NDJSON ordering and shape, bucket, content type) with idempotent re-run, second-generation rebuild proving byte-identical parent replay plus tail-only append, and a 500 on parent-object hash mismatch with recovery after restore. Migration consistency and the vercel-cron config pin are updated in the same diff.
- Conventions: Compliant — integration tests at the route entry point, boundary-only mocking through the shared registry, real database, no fake timers, no `vi.mock` relative-path violations; mock reset relies on the suite-wide setup.

PR #25792, opened 2026-08-08
fix(runner): reject mismatched local result identities

### LINKED ISSUE #25777: test(rust): reject mismatched local-submit result identities
## Location

- Result identity contract: `crates/runner/src/local_queue/types.rs:33-39`
- Result writer binds the same run ID into the payload and file path: `crates/runner/src/local_queue/state.rs:779-819`
- Submit completion path: `crates/runner/src/cmd/local/submit.rs:708-720`
- Current end-to-end result fixture: `crates/runner/src/cmd/local/submit/tests/support.rs:29-53`
- Current completed-result coverage: `crates/runner/src/cmd/local/submit/tests/completed_cleanup.rs:9-43`

## Problem and evidence

`JobResponse` carries a `run_id`, and `LocalQueue::write_result_sync` normally uses the same value for both the JSON payload and `results/{run_id}.result`. The submit-side `finish_completed` method deserializes that payload but never compares `response.run_id` with `self.queue.job_id`; it immediately cleans the submitted job's queue files, prints the payload, and maps only `response.exit_code` to the command exit status.

As a result, valid JSON for run B placed at run A's result path is accepted as run A's completion. A success payload returns success and removes run A's job/result/cancel/claim state even though the payload identifies a different run. This masks a cross-job result-routing regression or on-disk corruption instead of failing at the protocol boundary.

The current rendezvous helper always constructs a matching response from the observed request. The completed-result test checks only the exit code and cleanup. Replacing the fixture's response `run_id` with a fresh ID would leave those tests green, so the redundant identity invariant in the protocol is not protected.

Audited at `9e134cf42441e21a3318a1ec71c192124c2329e2`.

## Decision

Add a focused local-submit regression test for a syntactically valid result whose payload run ID differs from the submitted job ID, then enforce the identity check before the completion path performs cleanup, output, or exit-code mapping.

## Alternatives considered

- Rely only on the result filename: rejected because the payload deliberately carries an identity, and silently ignoring it makes cross-job writer regressions indistinguishable from valid completion.
- Test `JobResponse` deserialization in isolation: insufficient because deserialization already succeeds; the missing behavior is the submit boundary's correlation check and side-effect ordering.
- Remove `run_id` from `JobResponse`: out of scope and would weaken the existing self-identifying queue protocol rather than protect it.

## Risks and mitigations

- Changing validation order can affect cleanup behavior on invalid results. Assert that a mismatched payload is rejected before completed-job cleanup, so the submitted job is not silently finalized by another identity.
- Error text should include expected and actual run IDs for diagnosis without including prompts, environment values, or secrets.
- Keep matching success and nonzero responses covered so their current output, cleanup, and exit-status behavior does not regress.

## Acceptance criteria

- A regression test publishes a valid `JobResponse` under the submitted job's result path with a different `run_id`.
- Local submit returns a bounded, diagnostic error that identifies the expected and actual run IDs instead of returning the mismatched response's success or failure exit status.
- The mismatch is detected before completed-result cleanup and before the payload is emitted as a successful result.
- Existing matching success and nonzero response tests continue to verify normal cleanup and exit-status behavior.
- The focused runner local-submit tests pass.

seven332:
## Summary

- reject local result payloads whose `run_id` does not match the submitted queue identity
- fail before completed-result cleanup, stdout output, or exit-code mapping so invalid state remains diagnosable
- cover the mismatch through the real local-submit filesystem rendezvous while retaining matching success and nonzero coverage

## Compatibility

- no queue schema, payload shape, filename, writer, database, migration, or feature-switch changes
- matching results from existing runner versions retain their current behavior

## Validation

- `cd crates && cargo clippy -p runner --all-targets --all-features`
- `cd crates && cargo doc -p runner --all-features --no-deps`
- `cd crates && cargo test -p runner --all-targets --all-features` (3,081 passed; 20 ignored)

Closes #25777

### DIFF
--- a/crates/runner/src/cmd/local/submit.rs
+++ b/crates/runner/src/cmd/local/submit.rs
@@ -708,6 +708,12 @@ impl SubmitPlan {
     fn finish_completed(&self, buf: &[u8]) -> RunnerResult<ExitCode> {
         let response: JobResponse = serde_json::from_slice(buf)
             .map_err(|e| RunnerError::Internal(format!("parse result: {e}")))?;
+        if response.run_id != self.queue.job_id {
+            return Err(RunnerError::Internal(format!(
+                "local result run_id mismatch: expected {}, actual {}",
+                self.queue.job_id, response.run_id
+            )));
+        }
 
         self.queue.cleanup_completed();
 
--- a/crates/runner/src/cmd/local/submit/tests/completed_cleanup.rs
+++ b/crates/runner/src/cmd/local/submit/tests/completed_cleanup.rs
@@ -1,9 +1,12 @@
 use std::process::ExitCode;
 
-use super::super::SubmitArgs;
-use super::support::{run_submit_and_write_result, submit_queue_entry, write_queue_job_file};
+use super::super::{SubmitArgs, run_submit_with_home};
+use super::support::{
+    TEST_SUBMIT_RENDEZVOUS_TIMEOUT, run_submit_and_write_result, submit_args_for_test,
+    submit_queue_entry, wait_for_job_and_write_result, write_queue_job_file,
+};
 use crate::ids::RunId;
-use crate::local_queue;
+use crate::local_queue::{self, JobResponse};
 use crate::paths::HomePaths;
 
 #[tokio::test]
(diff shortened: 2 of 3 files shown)


seven332:
## Code Review: PR #25792

**Reviewed head:** `1bc7b033cd41ab06e3804037e855010e134a6d94`

### Summary

The submitted change enforces the existing local result identity invariant at the correct consumer boundary. It rejects a mismatched payload before cleanup, stdout output, or exit-code mapping without changing the queue format or matching-result behavior.

### Findings

No findings.

### Testing

The regression exercises `run_submit_with_home` against real temporary queue files, asserts the expected/actual diagnostic and preserved invalid state, and all modified coverable lines are covered. Local validation passed formatting, runner Clippy, runner docs, and the full runner suite (3,081 passed; 20 ignored); the focused mismatch scenario also passed ten consecutive runs. Remote runner E2E was not run because no remote runner or cross-deployment boundary changed.

### Verdict

LGTM

Issue #25405, opened 2026-08-06
perf: maximize runner preference reuse conversion and startup gains


seven332:
## Objective

Make the server-selected `runnerPreference` convert into the predicted reusable resource path,
using exact winning-candidate attribution and end-to-end latency rather than nominal preference
counts.

## Delivered work

All eight child issues are complete:

- #25411 added exact Poll and Ably winning-candidate attribution.
- #25506 protected an in-flight exact claim from an expired fallback race.
- #25507 preserved speculative exact sandboxes after claim loss.
- #25508 enforced global reusable-resource priority before claim.
- #25810 removed a heartbeat/workspace-checkout race.
- #25960 resolved claimed-finalizing reuse at sandbox publication.
- #26069 handed a prepared sandbox to a known immediate successor.
- #26472 preserved same-host workspace capability during drain.

#25741 and its two delivery slices then migrated API and runner traffic to one canonical
`RunnerPreference` contract and removed the rollout bridge. Release PR #25795 merged on 2026-08-08;
the later production runner versions recorded by #24203 were already well beyond its API 1.405.1 /
runner 0.161.1 cleanup gate.

## Production outcome

The fixed production window `[2026-08-12 11:00, 2026-08-13 11:00) UTC` recorded:

- 2,398 active exact-history claims and 2,395 known spawn outcomes;
- sandbox reuse for every known active exact-history outcome, with no workspace or cold downgrade;
- 3,317 of 3,328 known positive-preference outcomes meeting their predicted resource-quality floor
  (99.7%); and
- zero occurrences of the prior absent-runtime speculative rollback warning while runner warning
  ingestion remained active.

Canonical successful claims continue to expose the bounded
`runner_preference_resolution`, `runner_preference_claim_state`, and
`runner_preference_targeted_self` dimensions. Current source derives these values without ingesting
raw runner identity, heartbeat generation, or preference expiry as telemetry dimensions.

## Decision

Close as completed. Positive preference selection, claim conversion, rollback, resource priority,
contract rollout, cleanup, and production validation have met this parent's completion gate.

The remaining question is not conversion of an existing positive preference. Attribution of why a
reusable holder is absent, and whether that miss was preventable, remains separately scoped in
#26989. Retention or capacity policy must not change without that evidence.

## Preserved constraints

- `runnerPreference` remains advisory; runner-local validation and claim CAS are authoritative.
- Missing, stale, expired, malformed, or unsupported preference data remains cold-correct.
[...]

PR #25812, opened 2026-08-08
fix(runner): wait briefly for workspace cache lock

### LINKED ISSUE #25810: bug(runner): prevent heartbeat scans from racing workspace cache checkout
## Background

Runner workspace-cache entries are protected by an exclusive per-entry file lock. Checkout uses
the cached image as a move seed and retains the lock through the run and final promotion, so the
lock is required for correctness.

The runner also acquires the same entry lock briefly while scanning workspace-cache metadata for a
heartbeat. Workspace preparation currently tries the lock once without waiting and treats any
contention as an unavailable cache.

## Problem and evidence

A heartbeat cache scan can race the executor's workspace preparation:

1. The API selects a runner because its previous heartbeat advertised a matching workspace cache.
2. The runner claims the job, spawns the executor, and requests an immediate heartbeat to publish
   the changed reusable state.
3. The heartbeat and executor run concurrently. The heartbeat briefly takes the target entry lock
   while validating its metadata.
4. Workspace preparation performs one non-blocking lock attempt. If that attempt lands inside the
   short scan window, it records `workspace_image_prepare_lock_busy` and immediately falls back to
   a fresh workspace.

Two production occurrences on 2026-08-08 show the same ordering on the selected runner:

- `06:32:54.769490Z`: immediate heartbeat requested; `0.150 ms` later workspace preparation
  reported `lock is already held`; the heartbeat completed about `39 ms` after the request.
- `07:22:59.815823Z`: immediate heartbeat requested; `0.167 ms` later workspace preparation
  reported `lock is already held`; the heartbeat completed about `40 ms` after the request.

No concurrent run with the same reuse key was present in either case. The other runner respected
the workspace preference and deferred its claim. Both selected runs therefore downgraded from an
advertised workspace-cache path to a cold workspace solely because of local transient lock
contention.

## Lock-duration distinction

The same busy result can represent two very different lock lifetimes:

- A heartbeat holds one entry lock only while reading and validating that entry. The observed full
  scans covered approximately 239-245 entries in 39-40 ms, or about 0.16 ms per entry on average.
- A workspace checkout retains the entry lock through the complete run and final promotion, which
  can take seconds or minutes.

The lock has no TTL or holder classification. Waiting without a bound could stall an already
claimed job for the lifetime of another run. Never waiting makes sub-millisecond inspection
contention unnecessarily force a cold workspace.

## Decision

Proceed with a runner-only bounded wait at the workspace preparation boundary.

Wrap the existing cancellation-safe `crate::lock::acquire` loop in a strict `50 ms` timeout inside
`WorkspaceImageCache::prepare`. The lock primitive already:

- attempts non-blocking `flock` immediately;
- retries only held-lock results with backoff beginning at `5 ms`;
- returns path, inode-validation, permission, and filesystem errors immediately;
- protects against replaced lock files; and
- stops future retry attempts when its future is dropped.

On acquisition, retain the unchanged move-checkout and promotion guard. On timeout, preserve the
existing `WorkspaceCacheCheckoutResult::LockBusy` fresh-workspace fallback. Log acquisition errors
separately from elapsed busy contention.

Apply the wait only to `prepare`. `lease_active` has a similar one-shot acquisition shape but no
confirmed production miss and is intentionally outside this PR.

## Alternatives considered

- A workspace-specific faster retry loop could poll below `5 ms`, but it would duplicate hardened
  retry and lock-file replacement behavior without evidence that the existing first retry is too
  slow.
- A snapshot-only immediate heartbeat removes the confirmed trigger but leaves periodic heartbeat,
  GC, inspection, and mixed-process short contention unresolved while adding snapshot/coalescing
  semantics.
- Shared inspection locks do not provide an atomic shared-to-exclusive upgrade and still require a
  bounded wait, with a substantially larger correctness surface.
- Removing the lock or waiting indefinitely is incompatible with move-checkout correctness and
  bounded startup latency.

## Risks and mitigations

- A genuine long-held lease adds at most `50 ms` before cold fallback. This applies only to
  contested prepares; post-rollout validation must check claim-to-spawn latency.
- Loaded scheduling can delay lock release or retry. The bound remains authoritative, and tests use
  a short independent real holder with substantial margin inside it.
- Lock I/O failures must not consume the timeout. The existing acquire primitive returns them
  immediately and the caller logs them distinctly.
- Checkout or promotion correctness could regress only if the retained guard changes; this PR
  keeps that lifecycle unchanged and runs existing contention coverage.

## Acceptance criteria

- [ ] A deterministic runner integration test holds a seeded entry lock briefly and proves normal
      sandbox execution waits within `50 ms`, consumes the cached image as a move seed, and emits
      successful prepare/cache-hit telemetry.
- [ ] A deterministic runner integration test holds the entry beyond `50 ms` and proves execution
      falls back promptly to a fresh workspace with lock-busy telemetry rather than hanging.
- [ ] Lock path, inode-validation, permission, and filesystem errors return before the contention
      timeout and are logged separately from timeout contention.
- [ ] The exclusive entry guard, move checkout, workspace promotion, cache invalidation, GC, active
      lease, runner drain, and restart behavior remain unchanged.
- [ ] The change requires no API/runner protocol migration and remains safe during mixed runner
      rollout.
- [ ] Post-rollout telemetry shows no new matching-workspace preference downgrade caused by
      transient heartbeat lock contention, without a material claim-to-spawn regression.

## Relevant code

- `crates/runner/src/workspace_image_cache/lifecycle.rs`: bounded prepare acquisition and existing
  move/promotion lease flow.
- `crates/runner/src/lock.rs`: cancellation-safe retry and lock-file hardening.
- `crates/runner/src/executor/tests/sandbox_run_tests/workspace_cache.rs`: behavior and telemetry
  integration coverage.

## Parent objective

This is an evidence-backed optimization child of #25405. It addresses one confirmed
matching-workspace preference downgrade while preserving the parent's advisory preference and
cold-correct fallback constraints.

seven332:
## Summary

- wait up to 50 ms for a contended workspace image cache entry lock before falling back to a fresh workspace
- preserve the existing fresh-workspace fallback for persistent contention and lock errors, with distinct operational logs
- cover transient contention, persistent contention, and invalid lock-file errors through the runner executor entry point

## Scope

- does not change heartbeat scanning, the shared lock implementation, runner protocols, or persisted state
- does not wait beyond the strict 50 ms prepare budget

## Validation

- `cargo test -p runner workspace_cache_lock -- --nocapture`
- `cargo test -p runner`

Closes #25810

Parent: #25405

### DIFF
--- a/crates/runner/src/executor/tests/sandbox_run_tests/workspace_cache.rs
+++ b/crates/runner/src/executor/tests/sandbox_run_tests/workspace_cache.rs
@@ -792,7 +792,7 @@ async fn execute_inner_cache_hit_retry_requires_completed_cleanup() {
 }
 
 #[tokio::test]
-async fn execute_inner_uses_workspace_cache_when_configured() {
+async fn execute_inner_waits_for_transient_workspace_cache_lock() {
     let dir = tempfile::tempdir().unwrap();
     let runner_paths = RunnerPaths::new(dir.path().join("runner"));
     let cache = WorkspaceImageCache::new(runner_paths.clone());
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #25812

**Reviewed head:** `97f4cd77bc935ce1871870f41db780c0c96e46b9`

### Summary

The runner now gives a contended workspace-cache prepare lock a strict 50 ms acquisition budget,
while preserving the existing fresh-workspace fallback and exclusive checkout/promotion guard.
Coverage enters through normal executor sandbox creation with real temporary files and file locks.

### Findings

No findings on the current head. The first submitted head lacked coverage for non-contention lock
acquisition errors; `97f4cd77bc` resolves that gap and verifies that such errors remain visible and
are not mislabeled as elapsed contention.

### Testing

- Focused transient, persistent, and invalid-lock scenarios: 3 passed.
- Workspace-cache lifecycle suite: 120 passed, 1 pre-existing ignored.
- Full runner suite: 3,084 passed including guest inventory, 20 pre-existing ignored.
- Formatting, all-target/all-feature Clippy with warnings denied, and runner docs passed.
- Turbo and Python suites were not run because the change is runner-only Rust.
- Post-rollout reuse conversion and claim-to-spawn latency remain the operational verification gate
  tracked by parent issue #25405.

### Verdict

LGTM

Issue #25813, opened 2026-08-08
refactor(chat): finish post-contraction event cleanup


e7h4n:
## Context

The physical ChatEvent contraction is complete: #25805 shipped through release #25803, `chat_events` now has the canonical 18-column shape, and the retired `active_input_sequence`, `goal_event`, `attach_files`, `generation_template`, and `recommended_followups` columns plus their legacy index are gone.

#25802 has also merged the R2 snapshot reader phase to `main`. It adds v3 snapshot objects, a raw-row IndexedDB store, tail reads, and 410-driven cache rebuilds behind `FeatureSwitchKey.ChatEventSnapshotRead`, which remains disabled by default.

This issue tracks only the remaining post-contraction work. It must not reintroduce any of the retired columns, event types, API projections, or fallback paths.

## Delivery order

Complete the phases below serially. Each phase should be a focused PR, pass `/pr-auto`, reach production through a release PR, and finish its rollout/drain verification before the next irreversible phase starts.

### 1. Remove the unused `isGoalRun` compatibility surface

The production platform has no runtime read of `ChatEvent.isGoalRun`, but the field still causes work across every representation:

- The projected-events API performs a lateral lookup into `zero_runs.goal_id` for each event.
- The public `ChatEvent` contract and API projection still expose `isGoalRun`.
- Goal callbacks still write `isGoalRun: true`, and the internal callback parser retains it for rollback-eligible API versions.
- The raw-row snapshot reader maintains per-thread `goalRunIds` state solely to recreate projection parity for this unused field.

- [ ] Wait until rollback-eligible callback consumers that require the field have drained.
- [ ] Stop writing and parsing `isGoalRun` in internal callback payloads.
- [ ] Remove `isGoalRun` from the ChatEvent contract and projected response.
- [ ] Remove the `zero_runs` metadata subquery/lateral join from event reads.
- [ ] Remove client-side `goalRunIds` accumulation and raw-row projection support.
- [ ] Remove obsolete fixtures and parity assertions while preserving goal-run UI behavior derived from canonical `input.goal` data.

### 2. Complete the v3 snapshot-reader rollout

- [ ] Confirm the release containing #25802 reaches production and its old API/app rollback window closes.
- [ ] Configure and independently verify R2 CORS GET access for the supported app origins before enabling the switch.
- [ ] Keep the read switch off until every eligible watermarked thread has a readable current v3 head at the required watermark and there are no current heads below schema v3.
[...]

PR #25802, opened 2026-08-08
feat: cold-start chat threads from r2 snapshots behind a switch

e7h4n:
## Summary

Reader phase of the chat-event R2 snapshot design: behind `FeatureSwitchKey.ChatEventSnapshotRead` (default off), the web client cold-starts a chat thread from its archive object instead of paging the full history 50 events at a time.

Pipeline under the switch:

1. Empty local cache → `GET /api/zero/chat-threads/:threadId/event-snapshot` returns a 15-minute presigned URL for the thread's head archive plus its `lastSeqId`. 404 (`CHAT_EVENT_SNAPSHOT_NOT_FOUND`) covers threads the archiver has not reached or whose head predates the current schema version; per the internal-stage decision there is no legacy fallback and the error surfaces as-is.
2. The browser fetches the object from R2 directly. Objects now carry `Content-Encoding: gzip`, so the network stack decompresses transparently and the client parses plain NDJSON.
3. Raw rows are cached verbatim in a new `chat_events` IndexedDB store (IDB v29) that mirrors the server table name and row shape; projection into `ChatEvent` happens only at the read boundary via a shared `chatEventFromRow` in api-contracts. `isGoalRun` is derived client-side from claimed `input.goal` rows, accumulated per thread in a signal so background batches project correctly.
4. Catch-up and live sync tail `GET /api/zero/chat-threads/:threadId/event-rows?sinceSeqId=…` (raw rows, cursor-existence probe). A 410 (`CHAT_EVENTS_EXPIRED`) clears the row cache and rebuilds from a fresh snapshot — dormant today since nothing deletes rows yet, but the recovery loop the future payload-reclaim phase depends on. A cursor equal to the current head's `last_seq_id` stays valid so the snapshot handoff can never 410 itself.
5. Archive schema bumps to v3: identical line shape to v2, plus the encoding metadata. The archiver's version-mismatch selection migrates v2 heads automatically; because v2 bytes are identical to v3, a head with no new tail is refreshed in place (same object key re-put with the v3 headers, head row stamped to v3) instead of publishing a duplicate key. The read endpoints only serve v3 heads. Old superseded objects are left for the separate cleanup task.

With the switch off, every existing path (projected events endpoint, `chat_messages` store, backwards history backfill, background sync) is unchanged.

## Fallbacks

[...]

PR #25781, opened 2026-08-08
refactor(chat): prepare chat event schema contraction

e7h4n:
## Summary

- Phase 7A only: replace the archive worker's implicit full-row `chatEvents` read with an explicit canonical projection and write homogeneous archive schema v2 objects
- verify and transcode every v1 parent line before appending a v2 tail, canonicalize retired browser lifecycle values, and upgrade v1 heads even when PostgreSQL has no new tail
- backfill `browser.started` to `browser.open` and `browser.stopped` to `browser.close`, restore strict append-only enforcement, and narrow only the event-type constraint
- remove the independently drained 0861 insert trigger and helper functions
- intentionally retain all five physical columns, outgoing ORM declarations, goal-marker checks, and `chat_events_run_active_input_seq_unique` for the DB-before-API overlap; Phase 7B owns their later contraction

## Deployment boundary

Migration 0863 is a preparation migration, not the physical contraction. The generated latest snapshot retains `goal_event`, `attach_files`, `generation_template`, `recommended_followups`, `active_input_sequence`, and the active-input partial unique index. The semantic 0862-to-0863 snapshot diff has zero added/removed columns and indexes; only `chat_events_event_type_check` changes.

The controller will promote this exact merge, verify production/catalog state, and restart a minimum 102-minute old-API/rollback drain. Only then may a separate Phase 7B release remove the ORM fields, index, retained bridge declarations, and physical columns. This PR does not release production.

## Archive safety

Archive v2 explicitly projects and encodes 18 canonical fields. A v1 parent is content-hash verified, gunzipped, parsed line-by-line, stripped to the canonical shape, and maps retired browser spellings before any v2 tail is appended. V2 parents are strict-parsed, unsupported versions fail closed, and v1 heads are selected for version-only rewrites. Every active v2 head therefore has one homogeneous line shape and a matching version; immutable non-head v1 objects remain only for later R2 GC.

## Audit matrix

| Target | Active occurrences before | Phase 7A action | Allowed occurrences after | Verification |
| --- | ---: | --- | --- | --- |
| Five retired ChatEvent payload/sequence fields | archive serializer read all five; ORM/schema/index/check declarations | remove archive reads only | outgoing ORM/schema declarations, physical columns, index/check; v1 fixtures/history | active runtime search finds only the intentional DB compatibility surface; migration exercises the outgoing 23-column row shape |
| Implicit full-row `chatEvents` select | 1 (`readTailEvents`) | explicit 18-field projection | none | multiline repository search: 0 |
[...]

PR #25818, opened 2026-08-08
refactor(chat): finish post-contraction event cleanup

e7h4n:
## Summary

- remove the unused isGoalRun field from callbacks, API contracts/projection, zero_runs metadata reads, raw-row projection, client state, and fixtures
- require deterministic runEventId values for all new assistant-event insertion and delete the sequence-only dedupe branch
- serve only v3 snapshot heads, raise the supported web-client floor, retire the v2 rewrite path, and reconcile exact-origin R2 GET CORS from the snapshot cron
- give the snapshot-read pipeline a cold start for threads the archiver has not reached yet: `sinceSeqId: 0` reads a thread with no current head from its first event, and a 404 snapshot falls back to that path instead of failing every catch-up
- add sharded R2 mark-and-sweep GC with dry-run metrics, reference rechecks, a grace period, and recovery coverage
- keep the snapshot cron on its 10-minute cadence, align the R2 GC shard slot with it, and update stale deployment commentary

Part of #25813

## Scope note: PostgreSQL payload reclaim is not included

An earlier revision of this PR reclaimed snapshot-covered payload columns in `chat_events`. That is impossible without a schema change: `0863_prepare_chat_event_contraction.sql` (#25781) installs `chat_events_reject_update`, which raises on every UPDATE of that table, so clearing payload columns in place fails by design.

Reclaim is therefore dropped from this PR rather than relaxing the append-only invariant here. Postgres payloads stay in place; chat event retention will remove the rows later, and R2 snapshot lifecycle is managed independently of the database. #25813 stays open for that follow-up.

## Scope note: ChatEventSnapshotRead stays off

An earlier revision of this PR flipped `ChatEventSnapshotRead` to `enabled: true`. On every head of this branch `cli-e2e-02-playwright` then failed the same way — the user message and the live run indicator rendered, but no persisted assistant event ever appeared — while `main` and every other branch passed that job. The GA flip is the only change here that alters the client read pipeline, and no page-level test exercises that pipeline, so the e2e was its first real run.

The switch is back off. Everything the pipeline needs stays in this PR: the snapshot cron, R2 GC, the cold-start cursor, the reader contract, and the signal-level suite. GA moves to a follow-up that is a switch flip plus the page-level coverage this run proved is missing.

## Fallbacks

- `internal-chat-run-callback.service.ts:chatCallbackPayloadSchema.isGoalRun` + `zero-goal-queue-drain.service.ts:buildGoalChatCallbacks` — **removed**. Protected a callback row written by a newer API and dispatched by a rollback-eligible older API that gated pushes by run origin. Surface: DB/API, ~102 minutes. Removal evidence: type/schema — the payload schema keeps `.passthrough()`, so a pending row still carrying the field parses; single-reader — `git grep isGoalRun origin/main` shows no production reader remains, only the writer, the schema entry, and tests. Removal condition met by the release gate below.
- `cron-snapshot-chat-events.service.ts:REWRITABLE_ARCHIVE_SCHEMA_VERSION` / `isMetadataOnlyRewrite` (plus `putImmutableS3Object`'s `refreshMetadata` option) — **removed**. Protected pre-v3 snapshot heads by rewriting them in place. Surface: persisted R2/DB snapshot state, not a rolling-update window. Removal evidence: rollout — the merge gate below requires the count of current v1/v2 heads to be zero. An idle head on a retired version is now simply not a candidate, so it cannot strand the globally scoped pass; a retired-version head that gains a new tail still fails closed.
- `zero-chat-event-shared.service.ts` legacy `runEventId === undefined` insertion branch (plus the `run-sequence` conflict mode it used) — **removed**. Protected callers that could not supply a deterministic run-event id. Surface: DB/API, ~102 minutes. Removal evidence: type — both callers now always pass a value (`callback:${sequenceNumber}` and `eventMessageId`); and `chat_events_run_event_seq_unique` is an unconditional index that the untargeted `onConflictDoNothing()` still covers, so a callback retried across the deploy boundary cannot duplicate in either direction.
- `web-client-compatibility.json:minimumSupportedVersion` `0.707.0` → `0.708.2` — **kept, as the shortening mechanism** for old web clients that still read `isGoalRun`. Surface: old web clients, ~2 days, shortened by `426 Upgrade Required`. Ordering is safe because `0.708.2` is an already-released build (`app-v0.708.2`), not this release's own frontend.
- `workflow-chat-event-queue.service.ts:loadNextWorkflowQueueEvent` — **kept, re-justified**. The null-returning branch is unchanged; only its rationale moves from "legacy queue rows during code-before-migration" to a concurrently deleted thread, which is a genuine race with no rollout window.

Cross-version note for the `sinceSeqId` widening: `chat-threads.rows` now accepts `0` in addition to positive cursors. Old client to new API is unaffected because old clients never send `0`. The reverse pairing, a new app reaching an API from before this release, only occurs through promotion drift or rollback, since production promotes the API before the app; in that window the request is rejected as `400` and the affected thread renders no history until the API catches up.

## Safety

- R2 GC is deterministic and bounded, defaults to a seven-day grace window, protects every referenced head, removes eligible non-head DB references first, and rechecks all references immediately before object deletion.
- GC defaults to dry-run. Production must explicitly set CHAT_EVENT_SNAPSHOT_GC_DRY_RUN=false after measurement is reviewed. It never touches chat_events rows: it deletes only chat_event_snapshots rows and their R2 objects.
- Unsupported (pre-v3) snapshot heads fail closed in both the cron and the read endpoint instead of entering the retired rewrite fallback.
- Bucket CORS reconciliation runs in the cron only and preserves unrelated bucket rules. The snapshot read path issues no R2 control-plane call.
- No database migration or historical row rewrite is included.

## Required merge and rollout gates

- Do not merge until release #25811, which contains #25802, is confirmed in production and its API/app rollback window has drained.
- Before merge, confirm every eligible watermark has a readable current v3 head and the count of current v1/v2 heads is zero.
- After deploy, confirm the first cron pass reconciled exact-origin CORS. Snapshot cold start, raw-row projection, tail catch-up, warm cache, and 410-to-fresh-snapshot recovery are verified by the follow-up that turns `ChatEventSnapshotRead` on, since this release leaves it off.
- Review the first dry-run GC metrics and relevant logs before enabling deletion; then monitor snapshot, CORS, 410-loop, database, and rendering regressions.

## Verification

Passed locally:

- full-repository lint (13/13 tasks, 0 errors), type checks (14/14 tasks), knip, and formatting
- API `cron-snapshot-chat-events`, `zero-chat-event-snapshot`, and `vercel-crons` suites against a freshly migrated PostgreSQL database
- platform lab-page, chat-page signal, and chat workflow/goal suites


e7h4n:
## Blocked: payload reclaim conflicts with the `chat_events` append-only trigger

`reclaimSnapshotCoveredPayloads` issues `UPDATE chat_events SET content/userMessage/thinking/error/usagePayload ...` (`turbo/apps/api/src/signals/services/cron-snapshot-chat-events.service.ts:727`). The current schema installs `chat_events_reject_update BEFORE UPDATE ON chat_events FOR EACH ROW EXECUTE FUNCTION reject_chat_event_source_update()`, whose final form in `0863_prepare_chat_event_contraction.sql` (from #25781) unconditionally raises `chat_events is append-only; UPDATE is not allowed`.

Reproduced locally against a freshly migrated database:

```
FAIL src/signals/routes/__tests__/zero-chat-event-snapshot.test.ts > chat event snapshot read endpoints
     > reclaims only verified archived payloads and garbage-collects unreferenced objects
[ERROR][App] Unhandled request error: chat_events is append-only; UPDATE is not allowed
```

The test enables the destructive path (`CHAT_EVENT_PAYLOAD_RECLAIM_DRY_RUN=false`) and expects `payloadRowsReclaimed: 2`, so the failure is the real production behavior, not a fixture problem: once the switch is turned off in production, every non-dry-run cron pass would fail. This PR states "No database migration or historical row rewrite is included", so resolving it needs a decision:

1. Add a migration that narrowly permits payload-nulling updates (all identity/ordering columns unchanged), preserving append-only for everything else; or
2. Drop payload reclaim (and its R2 GC coupling) from this PR and land it separately alongside that migration.

### Other CI failures on this head (already fixed in c36750f)

- `lint-knip`: `chatEventSnapshotGcPrefixes` and `S3ObjectPage` were exported but unused; both are now module-private.
- `test-api (8)`: the global `@aws-sdk/client-s3` mock lacked `GetBucketCorsCommand`/`PutBucketCorsCommand`, so CORS reconciliation 500'd the cron. Mocks added.
- `test-api (8)`: `cron-snapshot-chat-events.test.ts` still asserted the retired v2-to-v3 rewrite. It now asserts the fail-closed contract, and both snapshot tests restore a supported head version so a stale head cannot poison the globally scoped cron for later tests.
- `test-app (1)`: `setupPage` began overriding the feature-switches GET handler unconditionally, which clobbered `lab-page.test.tsx`'s own stateful handler. The generic default is now pinned in the default handler instead.
- `cli-e2e-02-playwright`: `chat.spec.ts` "send a message through the deployed runner" did not render the assistant reply within 90s. Not yet classified; re-evaluating on the new head.

e7h4n:
Resolved: payload reclaim is dropped from this PR (commit `4ab32bb`), per the decision recorded in vm0-ai/vm0#25813 (comment). `chat_events` stays strictly append-only; retention will delete the rows later and R2 snapshot lifecycle is handled independently of the database.

Removed here: reclaim candidate scan, snapshot row verification used only by reclaim, transactional batching, the `payloadRows*` cron response fields, the `CHAT_EVENT_PAYLOAD_RECLAIM_*` env surface, and the reclaim fixtures. R2 GC is untouched and unaffected — it only deletes `chat_event_snapshots` rows and R2 objects.

Local verification on this head: full-repository lint (13/13, 0 errors), type checks (14/14), knip, formatting, plus `cron-snapshot-chat-events`, `zero-chat-event-snapshot`, and `vercel-crons` against a freshly migrated PostgreSQL database (10/10 tests).

e7h4n:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 4ab32bb86a6b83a0ba0447adb5c4a07207ab5990

### Summary

Finishes the post-contraction chat-event cleanup: removes the unused `isGoalRun` field from callbacks, contracts, the `zero_runs` lateral join, the raw-row projection and client state; makes `runEventId` mandatory for assistant-event insertion and deletes the sequence-only dedupe branch; promotes v3 snapshot reads to GA, raises the web-client floor, retires the v2 rewrite path, reconciles exact-origin R2 GET CORS; and adds sharded R2 mark-and-sweep GC. Verdict is `Changes Requested` because the PR summary has no `Fallbacks` section even though the diff removes four compatibility behaviors, and because three removals left orphaned code behind.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic: no race conditions, unhandled edge cases, or wrong assumptions | Finding (P1) | `cron-snapshot-chat-events.service.ts` — the candidate query deliberately selects heads whose `archive_schema_version != 3`, and `archiveThread` then throws on exactly those. One stranded head aborts the whole pass, including R2 GC, for every other thread. |
| `REVIEW.md` | Security: no injection, hardcoded secrets, or missing auth on new surfaces | Pass | No new endpoint. GC and CORS run inside the existing cron-secret-gated route; `chatEventSnapshotGcPrefixes` validates `CHAT_EVENT_SNAPSHOT_GC_SHARD` against `/^[0-9a-f]{3}$/u` before using it in a prefix. |
| `REVIEW.md` | Type safety: no `any`, no overly broad types | Pass | No `any` added. `InsertAssistantEventsInput.items[].runEventId` is narrowed from `string \| undefined` to `string`, and `chatEventFromRow` drops its extra parameter. |
| `REVIEW.md` | Style and maintainability: no duplicated logic, no dead code, comments explain WHY | Finding (P1) | `s3.ts` `refreshMetadata` and `zero-chat-event.service.ts` `"run-sequence"` are left without callers; `zero-chat-event-snapshot.service.ts:98` still documents payload reclaim, which this PR does not contain. |
| `REVIEW.md` | Feature regression guard: replaced behavior keeps explicit coverage | Pass | `chat-workflows-and-goals.test.tsx` and `chat-callbacks.bdd.test.ts` were updated to the new shape rather than duplicated, and goal folding is still asserted through `isGoalUserMessage`, which now carries the behavior `isGoalRun` used to. |
| `docs/bad-smell.md` | Zero tolerance for `any`, lint/type suppressions, dynamic imports | Pass | No `any`, no `eslint-disable`/`@ts-*` comments, no dynamic `import()` in the diff. |
| `docs/bad-smell.md` | Fail fast instead of silently absorbing errors | Pass | `deleteS3Objects` now throws when `DeleteObjectsCommand` returns per-object `Errors`, replacing a silent partial success. |
| `docs/bad-smell.md` | Delete unused code aggressively (YAGNI) | Finding (P1) | Two removals stopped at the call site: `putImmutableS3Object`'s `refreshMetadata` option and `insertChatEvents`' `"run-sequence"` conflict mode both survive with zero callers. |
| `docs/fallback.md` §9 | Every fallback the diff introduces, keeps, or removes is declared in the PR summary | Finding (P1) | The summary has no `Fallbacks` section. Four removals qualify: the `isGoalRun` callback-payload field, the v2 archive rewrite path, the `runEventId === undefined` insertion branch, and the client-floor raise that ends `isGoalRun` client compatibility. |
| `docs/fallback.md` §10 | A removal PR states type, single-writer, production, or rollout evidence | Finding (P1) | The evidence exists and checks out (see Fallbacks below) but is not written down anywhere in the PR. |
| `docs/fallback.md` §1 | No negative tests asserting removed behavior stays removed | Pass | The rewritten `cron-snapshot-chat-events.test.ts` case pins the current fail-closed contract for an unsupported head and then restores a supported version; it is not a tombstone for deleted code. |
| `docs/fallback.md` §2 | A feature behind a non-GA switch needs no compatibility code | Pass | `ChatEventSnapshotRead` flips to `enabled: true`, so this PR is exactly the GA boundary; from here `docs/deployment-compatibility.md` applies in full, which is what the floor raise addresses. |
| `docs/fallback.md` §7 | A rollout fallback is sized to the surface's observed maximum exposure | Pass | The only surface-bound compatibility left is old web clients reading `isGoalRun` (~2 days), and the floor raise plus `426` is the shortening mechanism. |
| `docs/deployment-compatibility.md` | Raise the client floor only after the corresponding app build is live | Pass | Floor moves `0.707.0` → `0.708.2`; `main` is at app version `0.710.0` and tag `app-v0.708.2` exists, so the floor targets an already-shipped build rather than this release's own frontend. |
| `docs/deployment-compatibility.md` | Do not remove a response field while old clients may still read it | Pass | `isGoalRun` is `optional()` in the old client schema, so a response without it still validates; `git grep isGoalRun origin/main` shows no production reader, only the writer, the schema entry, and tests. |
| `docs/deployment-compatibility.md` | Data written by one version must stay safe for the other version during rollout | Pass | Old API writes `runEventId: null` with `(runId, runEventSequenceNumber)` dedupe; the new API writes a deterministic id with `onConflictDoNothing()` and no target, which still covers `chat_events_run_event_seq_unique`, so a callback retried across the deploy boundary cannot duplicate in either direction. |
| `docs/deployment-compatibility.md` | No schema migration accompanies the contract change | Pass | No files under `turbo/packages/db/src/migrations/**`; the response contract only removes `isGoalRun` and adds cron metrics. |
| `docs/event-sourcing.md` | Persistent events reconcile optimistic events by a shared event ID; no failure-path rollback | Pass | The diff only removes a derived display flag from the projection. `chatEventFromRow` still keys on `row.id`, and no rollback, timer, or cleanup path is added. |
| `docs/cache.md` RC-3 | No unbounded process-lifetime keyed caches | Pass | The PR deletes `chat-event-goal-run-ids.ts`, whose module-level `state<Map<threadId, Set<runId>>>` grew per thread for the session lifetime. |
| `docs/cache.md` RC-4 | State belongs to the narrowest lifecycle that owns it | Pass | Removing the cross-pane goal-run map leaves the projection purely row-derived, with no state outliving its pane. |
| `docs/effect.md` | Render stays pure; work runs at the owning boundary | Pass | Platform changes are deletions plus two pure function renames (`isGoalRunGroupFold` → `isGoalGroupFold`) inside existing render helpers; no new command, timer, or subscription. |
| `.claude/skills/ccstate/SKILL.md` | Commands declare only the store access they use; no `void`-silenced promises in signals | Pass | `syncChatThreadRowsToIndexedDb$` and `loadRowCacheIntoPersistentEvents$` drop the now-unused `get` from their destructured argument; no `void`/`detach` misuse introduced. |
| `docs/testing.md` | `feat:`-level behavior ships with integration tests | Pass | The new R2 GC path is covered end-to-end through the cron contract: head protection, dry-run measurement, orphan deletion, and reclamation of a replaced head. |
| `docs/testing.md` / `docs/testing/anti-patterns.md` AP-4 | Mock only at the external boundary | Pass | The only new `vi.mock` surface is `@aws-sdk/client-s3` (third-party), extended with `GetBucketCorsCommand`/`PutBucketCorsCommand`; no relative-path mocks added. |
| `docs/testing/anti-patterns.md` AP-5 | No fake timers | Pass | The GC test advances time with `mockNow(future)` and file mtimes, not `vi.useFakeTimers()`. |
| `docs/testing/anti-patterns.md` AP-1 | Assert behavior, not mock calls | Pass | Assertions are on cron response metrics and on whether the fake R2 object still exists, never on `toHaveBeenCalled()` alone. |
| `docs/testing/api-testing.md` | Route tests go through `setupApp()` + the ts-rest contract | Pass | The GC test drives `cronSnapshotChatEventsContract.snapshot` through `setupApp` and reads the contract-validated response body. |
| `docs/testing/testing-external-behavior.md` | Construct and verify state through the external interface | Pass (with noted exception) | Snapshot head version, R2 object age, and orphan objects have no user-facing entry point, which is the guide's explicit exception; the primary assertions are still the cron endpoint's response metrics. |
| `docs/testing/anti-patterns.md` AP-10 | Platform tests go through `detachedSetupPage`, not direct rendering | Pass | `page-helper.ts` and `api-feature-switches.ts` only change which feature-switch defaults the production bootstrap sees; no direct component rendering added. |

### Findings

#### Critical (P0)

None.

- **PR summary is missing the mandatory `Fallbacks` section.** `docs/fallback.md` §9 makes an undeclared fallback a P1 on its own and blocks the verdict. This diff removes four compatibility behaviors (listed below). Add a `Fallbacks` section naming each one with its surface, window, removal evidence, and follow-up.
- **`turbo/apps/api/src/signals/external/s3.ts:920`** — `putImmutableS3Object`'s `refreshMetadata` option (and the `IfNoneMatch` branch it drives at line 936) exists only to serve the v2 metadata-only rewrite that this PR deletes. `grep -rn refreshMetadata apps/api/src` now finds no caller. `docs/fallback.md` §8 requires the removal to delete the branch, the contract entry, and the helper types together.
- **`turbo/apps/api/src/signals/services/zero-chat-event.service.ts:407,1087`** — the `"run-sequence"` member of `InsertChatEventsConflict` and its `onConflictDoNothing({ target: [runId, runEventSequenceNumber] })` branch have no caller after the legacy assistant-insert branch is deleted. The parameter is now single-valued. It is also redundant: `chat_events_run_event_seq_unique` is an unconditional index, so the untargeted `onConflictDoNothing()` already covers that constraint.
- **`turbo/apps/api/src/signals/services/cron-snapshot-chat-events.service.ts`** — one stranded pre-v3 head halts the entire cron pass. The candidate query selects `ne(archiveSchemaVersion, ARCHIVE_SCHEMA_VERSION)` on purpose, `archiveThread` throws for exactly that case, and the per-candidate loop has no isolation, so archiving *and* the new R2 GC stop for every other thread until the row is fixed by hand. The PR's own new test documents the blast radius ("an unsupported head left behind would fail every later pass in the suite"). Either drop the stale-version branch from the candidate query or log and skip the thread instead of throwing, so the failure stays per-thread.
- **`turbo/apps/api/src/signals/services/zero-chat-event-snapshot.service.ts:79`** — `chatEventSnapshotCorsReady()` builds a fresh `computed` on every call, so every snapshot download request issues a `GetBucketCors` (and a `PutBucketCors` on drift) to R2 before the presigned URL is returned. Bucket CORS is bucket-level configuration that changes approximately never, the cron already reconciles it once per pass, and this puts an extra R2 control-plane round trip on the cold chat-thread open path plus a new fail-closed dependency there. Reconcile in the cron only, or memoize per process.
- **`turbo/apps/api/src/signals/services/zero-chat-event-snapshot.service.ts:98`** — the `zeroChatThreadEventRows` doc comment still asserts "Payload reclaim preserves cursor rows and therefore cannot create false 410 responses." Payload reclaim was removed from this PR in `4ab32bb`; the comment now describes a mechanism that does not exist.

### Fallbacks

- `internal-chat-run-callback.service.ts:chatCallbackPayloadSchema.isGoalRun` + `zero-goal-queue-drain.service.ts:buildGoalChatCallbacks` — **removed**. Protected a callback row written by a newer API and dispatched by a rollback-eligible older API that gated pushes by run origin. Surface: DB/API, ~102 minutes. Removal evidence (verified, but not stated in the PR): the schema keeps `.passthrough()`, so a pending row that still carries the field parses; `git grep isGoalRun origin/main` shows no production reader remains. Declared in PR summary: **No — P1**. Verdict: **justified**.
- `cron-snapshot-chat-events.service.ts:REWRITABLE_ARCHIVE_SCHEMA_VERSION` / `isMetadataOnlyRewrite` — **removed**. Protected pre-v3 snapshot heads by rewriting them in place. Surface: persisted R2/DB snapshot state, not a rolling-update window. Removal condition: the PR's merge gate "the count of current v1/v2 heads is zero". Declared in PR summary: **No — P1** (the merge gate mentions it, but there is no `Fallbacks` entry). Verdict: **justified only if the gate is actually executed before merge**, and the removal is incomplete while `refreshMetadata` survives.
- `zero-chat-event-shared.service.ts` legacy `runEventId === undefined` insertion branch — **removed**. Protected callers that could not supply a deterministic run-event id. Surface: DB/API, ~102 minutes. Removal evidence (verified, but not stated in the PR): type evidence — both callers now always pass a value (`callback:${sequenceNumber}` and `eventMessageId`) — plus `chat_events_run_event_seq_unique`, which still dedupes an old-API insert against a new-API retry in both directions. Declared in PR summary: **No — P1**. Verdict: **justified**; removal incomplete while `"run-sequence"` survives.
- `web-client-compatibility.json:minimumSupportedVersion` `0.707.0` → `0.708.2` — **kept, as the shortening mechanism** for old web clients that still read `isGoalRun`. Surface: old web clients, ~2 days, shortened by `426 Upgrade Required`. Ordering verified: `main` is at app `0.710.0` and `app-v0.708.2` is an existing tag, so the floor points at a live build, per `docs/deployment-compatibility.md`. Declared in PR summary: **No — P1**. Verdict: **justified**.
- `workflow-chat-event-queue.service.ts:loadNextWorkflowQueueEvent` — **kept, re-justified**. The null-returning branch is unchanged; only its rationale moves from "legacy queue rows during code-before-migration" to "a concurrently deleted thread". Verdict: **justified** as a genuine race, no window. Worth one line in the summary so the rationale change is visible.

### Testing

- Coverage: **Adequate for the new behavior.** R2 GC is covered through the cron contract (head protection, dry-run measurement, orphan deletion, reclamation after a head replacement), CORS reconciliation is asserted via `corsChanged`, the retired v2 path has an updated fail-closed test, and the `isGoalRun` removal updates existing assertions instead of leaving tombstones. Two gaps worth noting: nothing covers the new throw in `deleteS3Objects` when `DeleteObjectsCommand` returns per-object `Errors` (a shared helper other flows also use), and nothing pins the blast radius of the stranded-head failure described in Findings.
- Conventions: **Compliant.** Third-party-only mocks, real database, `mockNow` instead of fake timers, behavioral assertions rather than call assertions, route-level `setupApp` usage. Both snapshot suites now restore a supported head version after exercising the unsupported one, which is necessary because the cron's scope is global and shared across the suite.

e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: b790ef3eef1eee18e038817be0b0ba969c5e0561
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Re-review at `b790ef3`. The PR still removes the unused `isGoalRun` field end to end, makes `runEventId` mandatory for assistant-event insertion, promotes v3 snapshot reads to GA behind a raised web-client floor, retires the v2 rewrite path, and adds sharded R2 mark-and-sweep GC. Every P1 from the `4ab32bb` review is resolved: the summary now carries a `Fallbacks` section, both orphaned code paths are deleted, bucket CORS reconciliation moved off the read path, an idle retired-version head can no longer abort the whole cron pass, and the stale payload-reclaim comment is gone.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic: no race conditions, unhandled edge cases, or wrong assumptions | Pass | The candidate query now selects only threads whose watermark is ahead of their head, so a stranded pre-v3 head is no longer pulled into the pass; `archiveThread` still fails closed for a retired-version head that gains a tail, and `cron-snapshot-chat-events.test.ts` pins both halves. |
| `REVIEW.md` | Security: no injection, hardcoded secrets, or missing auth on new surfaces | Pass | No new endpoint. GC stays inside the cron-secret-gated route and `chatEventSnapshotGcPrefixes` validates `CHAT_EVENT_SNAPSHOT_GC_SHARD` against `/^[0-9a-f]{3}$/u` before building a prefix. |
| `REVIEW.md` | Type safety: no `any`, no overly broad types | Pass | No `any`. `runEventId` narrows to `string`, `chatEventFromRow` drops a parameter, and `insertChatEvents` drops a now single-valued `conflict` parameter. |
| `REVIEW.md` | Style and maintainability: no duplicated logic, no dead code, comments explain WHY | Pass | `putImmutableS3Object`'s `refreshMetadata` option and the `"run-sequence"` conflict mode are both deleted with their callers; the `zeroChatThreadEventRows` comment no longer references payload reclaim. |
| `REVIEW.md` | Feature regression guard: replaced behavior keeps explicit coverage | Pass | Goal-run tests were updated to the new shape rather than duplicated, and `app-factory.test.ts` keeps both the "current" and "newer" client-version cases while deriving the newer fixture from the floor. |
| `docs/bad-smell.md` | Zero tolerance for `any`, lint/type suppressions, dynamic imports | Pass | None present in the diff. |
| `docs/bad-smell.md` | Fail fast instead of silently absorbing errors | Pass | `deleteS3Objects` throws when `DeleteObjectsCommand` returns per-object `Errors`, replacing a silent partial success. |
| `docs/bad-smell.md` | Delete unused code aggressively (YAGNI) | Pass | Both incomplete removals from the previous review are finished; `chatEventSnapshotCorsReady` is also demoted to module-private now that only the cron calls it. |
| `docs/fallback.md` §9 | Every fallback the diff introduces, keeps, or removes is declared in the PR summary | Pass | The summary now has a `Fallbacks` section covering all five entries listed below. |
| `docs/fallback.md` §10 | A removal PR states type, single-writer, production, or rollout evidence | Pass | Each removed fallback records its evidence: `.passthrough()` plus no remaining reader for `isGoalRun`; the v1/v2-head-count merge gate for the rewrite path; type evidence plus `chat_events_run_event_seq_unique` for the legacy insertion branch. |
| `docs/fallback.md` §1 | No negative tests asserting removed behavior stays removed | Pass | The unsupported-head test pins the current fail-closed contract on a reachable path (retired-version head plus a new tail) and additionally asserts that an idle one is simply skipped. |
| `docs/fallback.md` §2 | A feature behind a non-GA switch needs no compatibility code | Pass | `ChatEventSnapshotRead` flips to `enabled: true`, which is the GA boundary this PR owns; the floor raise is the matching client-side mechanism. |
| `docs/fallback.md` §7 | A rollout fallback is sized to the surface's observed maximum exposure | Pass | The only surface-bound item left is old web clients (~2 days), shortened by `426`; DB/API items are sized to ~102 minutes in the summary. |
| `docs/deployment-compatibility.md` | Raise the client floor only after the corresponding app build is live | Pass | `0.707.0` → `0.708.2`, while `main` is at app `0.710.0` and `app-v0.708.2` is an existing tag, so the floor targets a shipped build rather than this release's own frontend. |
| `docs/deployment-compatibility.md` | Do not remove a response field while old clients may still read it | Pass | `isGoalRun` was `optional()` in the old client schema, so a response without it still validates, and no production reader remains on `main`. |
| `docs/deployment-compatibility.md` | Data written by one version stays safe for the other during rollout | Pass | Old API writes `runEventId: null` deduped by `(runId, runEventSequenceNumber)`; the new API writes a deterministic id with an untargeted `onConflictDoNothing()`, which still hits `chat_events_run_event_seq_unique`, so a callback retried across the deploy boundary cannot duplicate in either direction. The new doc comment on `insertChatEvents` records this. |
| `docs/deployment-compatibility.md` | No schema migration accompanies the contract change | Pass | No files under `turbo/packages/db/src/migrations/**` in the PR scope. |
| `docs/event-sourcing.md` | Persistent events reconcile optimistic events by a shared event ID; no failure-path rollback | Pass | Only a derived display flag leaves the projection; `chatEventFromRow` still keys on `row.id` and no rollback, timer, or cleanup path is added. |
| `docs/cache.md` RC-3 | No unbounded process-lifetime keyed caches | Pass | `chat-event-goal-run-ids.ts` and its module-level `state<Map<threadId, Set<runId>>>` are deleted. |
| `docs/cache.md` RC-4 | State belongs to the narrowest lifecycle that owns it | Pass | The projection is now purely row-derived, with no cross-pane map outliving its owner. |
| `docs/effect.md` | Render stays pure; work runs at the owning boundary | Pass | Platform changes are deletions plus two pure helper renames inside existing render paths; no new command, timer, or subscription. |
| `.claude/skills/ccstate/SKILL.md` | Commands declare only the store access they use; no `void`-silenced promises in signals | Pass | The two chat-page commands drop the now-unused `get`; no `void`/`detach` misuse introduced. |
| `docs/testing.md` | New behavior ships with integration tests | Pass | R2 GC is covered end to end through the cron contract, and the retired-rewrite contract now has both the skipped-idle and fail-closed cases. |
| `docs/testing.md` / `docs/testing/anti-patterns.md` AP-4 | Mock only at the external boundary | Pass | The only new mock surface is `@aws-sdk/client-s3` (third party), extended with the two CORS commands. |
| `docs/testing/anti-patterns.md` AP-5 | No fake timers | Pass | The GC test advances time with `mockNow` and file mtimes. |
| `docs/testing/anti-patterns.md` AP-1 | Assert behavior, not mock calls | Pass | Assertions are on cron response metrics and on whether the fake R2 object still exists. |
| `docs/testing/api-testing.md` | Route tests go through `setupApp()` + the ts-rest contract | Pass | Both snapshot suites drive the cron and thread contracts through `setupApp`. |
| `docs/testing/testing-external-behavior.md` | Construct and verify state through the external interface | Pass (with noted exception) | Snapshot head version, R2 object age, and orphan objects have no user-facing entry point, which is the guide's explicit exception; primary assertions remain the endpoint's response body. |
| `docs/testing/anti-patterns.md` AP-10 | Platform tests go through `detachedSetupPage`, not direct rendering | Pass | The platform test changes only adjust which feature-switch defaults the production bootstrap sees. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None. All six P1s from the `4ab32bb` review are resolved:

- `Fallbacks` section added to the PR summary with per-entry surface, window, and removal evidence.
- `putImmutableS3Object`'s `refreshMetadata` option and its `IfNoneMatch` branch deleted.
- `InsertChatEventsConflict` / `"run-sequence"` deleted; `insertChatEvents` now has one documented conflict behavior.
- The candidate query no longer selects idle retired-version heads, so one stranded head cannot abort the globally scoped pass.
- Bucket CORS reconciliation is cron-only; the snapshot read path issues no R2 control-plane call before returning a presigned URL.
- The stale payload-reclaim sentence is removed from the raw-row endpoint comment.

One accepted consequence worth recording rather than blocking: a retired-version head with no new events is now never re-archived, so such a thread keeps falling back to the raw-row path until it receives new events. That matches the PR's stated position that version migration is an explicit rollout operation, and the merge gate already requires the current v1/v2 head count to be zero.

### Fallbacks

- `internal-chat-run-callback.service.ts:chatCallbackPayloadSchema.isGoalRun` + `zero-goal-queue-drain.service.ts:buildGoalChatCallbacks` — **removed**. Protected a callback row written by a newer API and dispatched by a rollback-eligible older API that gated pushes by run origin. Surface: DB/API, ~102 minutes. Evidence: `.passthrough()` keeps pending rows parseable; no production reader remains on `main`. Declared in PR summary: **Yes**. Verdict: **justified**.
- `cron-snapshot-chat-events.service.ts:REWRITABLE_ARCHIVE_SCHEMA_VERSION` / `isMetadataOnlyRewrite` plus `putImmutableS3Object.refreshMetadata` — **removed**. Protected pre-v3 heads by rewriting them in place. Surface: persisted R2/DB snapshot state, no rolling-update window. Removal condition: the merge gate requiring zero current v1/v2 heads. Declared in PR summary: **Yes**. Verdict: **justified**, and the removal is now complete.
- `zero-chat-event-shared.service.ts` legacy `runEventId === undefined` branch plus the `"run-sequence"` conflict mode — **removed**. Protected callers that could not supply a deterministic run-event id. Surface: DB/API, ~102 minutes. Evidence: type evidence on both callers, plus `chat_events_run_event_seq_unique` still deduping across the deploy boundary. Declared in PR summary: **Yes**. Verdict: **justified**, and the removal is now complete.
- `web-client-compatibility.json:minimumSupportedVersion` `0.707.0` → `0.708.2` — **kept**, as the shortening mechanism for old web clients that read `isGoalRun`. Surface: old web clients, ~2 days, shortened by `426`. Declared in PR summary: **Yes**. Verdict: **justified**; ordering verified against `app-v0.708.2`.
- `workflow-chat-event-queue.service.ts:loadNextWorkflowQueueEvent` — **kept, re-justified**. The null-returning branch is unchanged; its rationale moves from a migration window to a concurrently deleted thread. Declared in PR summary: **Yes**. Verdict: **justified**, no window.

### Testing

- Coverage: **Adequate.** R2 GC, CORS reconciliation (`corsChanged`), the retired-rewrite contract in both its skipped-idle and fail-closed forms, and the client-floor cases are all exercised. Author reports local runs of `cron-snapshot-chat-events`, `zero-chat-event-snapshot`, `app-factory` (52 tests) and `chat-callbacks.bdd`, `webhooks-callbacks.bdd`, `webhooks-agent-events-telemetry` (94 tests) against a freshly migrated PostgreSQL database, plus full-repository lint, type checks, and knip. One gap remains: nothing covers the new throw in `deleteS3Objects` when `DeleteObjectsCommand` returns per-object `Errors`, and that helper is shared with other flows.
- Conventions: **Compliant.** Third-party-only mocks, real database, `mockNow` instead of fake timers, behavioral assertions, route-level `setupApp`. Both snapshot suites restore a supported head version after exercising the unsupported one, which is required because the cron's scope is global.
- CI at this commit is still running. This LGTM is a code review only; merge remains gated on every required check going green, on `cli-e2e-02-playwright` in particular, and on the rollout gates stated in the PR summary.

e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 392f00c94bb486593c6312f44bce4de3a2a722e3
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Re-review at `392f00c`. Everything covered in the `b790ef3` review is unchanged and still passes. The new work in this range gives the snapshot-read pipeline the cold start it was missing: a thread the archiver has not reached yet returns `404` from the snapshot endpoint, and that used to abort the whole sync, so every Ably catch-up for that thread failed and only optimistic and live run state rendered. `sinceSeqId: 0` now reads such a thread from its first event, and the cold-start cursor is treated as server-derived so an expiry right after it fails loudly instead of looping.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic: no race conditions, unhandled edge cases, or wrong assumptions | Pass | The cold-start cursor is valid only while the thread has no current head, so the moment the archiver publishes one the client is sent back through the snapshot path; `cursorFromServer` is set for both the snapshot and the no-snapshot cursor, which removes the 404-to-`expired`-to-404 loop a thread with an unservable head would otherwise enter. |
| `REVIEW.md` | Security: no injection, hardcoded secrets, or missing auth on new surfaces | Pass | No new endpoint or auth path; `zeroChatThreadEventRows` still resolves ownership through `ownedThread` before reading any row. |
| `REVIEW.md` | Type safety: no `any`, no overly broad types | Pass | `fetchChatEventSnapshotRows$` widens its return to `... | null` rather than casting, and the caller narrows on `null`. |
| `REVIEW.md` | Style and maintainability: no duplicated logic, no dead code, comments explain WHY | Pass | `loadColdStartCursor` returns a single cursor instead of a cursor plus a flag, and the `THREAD_START_SEQ_ID` constants on both sides name the value; the endpoint comment now states exactly the rule the code implements. |
| `REVIEW.md` | Feature regression guard: replaced behavior keeps explicit coverage | Pass | The existing `410` case for a bogus cursor is retained verbatim, and the new `sinceSeqId: 0` case is asserted next to it in the same test. |
| `docs/bad-smell.md` | Zero tolerance for `any`, lint/type suppressions, dynamic imports | Pass | None present. |
| `docs/bad-smell.md` | Fail fast instead of silently absorbing errors | Pass | A `404` snapshot is a documented product state, not an absorbed error; every other snapshot failure still rethrows, and an expiry after a server-derived cursor throws with a precise message. |
| `docs/bad-smell.md` | Delete unused code aggressively (YAGNI) | Pass | No new option, flag, or branch beyond the one cursor value; the intermediate `{ sinceSeqId, rebuilt }` shape was removed once it proved redundant. |
| `docs/fallback.md` §3 | Do not fall back on states the contract cannot produce | Pass | The `404` branch guards a state the API produces by design for every unarchived thread, which the contract documents on the `snapshot` route. |
| `docs/fallback.md` §9 | Every fallback the diff introduces, keeps, or removes is declared in the PR summary | Pass | The `Fallbacks` section still lists all five entries, and the summary now also records the `sinceSeqId` widening with its cross-version direction. |
| `docs/fallback.md` §10 | A removal PR states type, single-writer, production, or rollout evidence | Pass | Unchanged from the previous review; each removed fallback keeps its evidence. |
| `docs/deployment-compatibility.md` | Prefer additive changes at cross-version boundaries | Pass | `sinceSeqId` moves from `positive()` to `nonnegative()`, which is a widening: an old client never sends `0`, so old client to new API is unaffected. |
| `docs/deployment-compatibility.md` | Do not require a new request value from the client in the same release that first sends it | Pass (documented risk) | The new app can send `0` to an API from before this release only through promotion drift or rollback, because production promotes the API before the app. The summary states that window and its effect (`400`, no history for that thread until the API catches up) rather than leaving it implicit. |
| `docs/deployment-compatibility.md` | No schema migration accompanies the contract change | Pass | No files under `turbo/packages/db/src/migrations/**` in PR scope. |
| `docs/event-sourcing.md` | Persistent events reconcile optimistic events by a shared event ID; no failure-path rollback | Pass | The cold start only changes which cursor the persistent rows are fetched from; no optimistic event is removed on any failure path. |
| `docs/cache.md` RC-3 | No unbounded process-lifetime keyed caches | Pass | Unchanged; `chat-event-goal-run-ids.ts` and its module-level map remain deleted. |
| `docs/cache.md` RC-4 | State belongs to the narrowest lifecycle that owns it | Pass | `cursorFromServer` and the cursor are locals inside one command invocation. |
| `docs/effect.md` | Render stays pure; work runs at the owning boundary | Pass | No render-path change in this range. |
| `.claude/skills/ccstate/SKILL.md` | Commands declare only the store access they use; no `void`-silenced promises in signals | Pass | `loadColdStartCursor` is a closure inside the existing command and uses the same `set`/`get`; no new detach or floating promise. |
| `docs/testing.md` | New behavior ships with integration tests | Pass | Three new cases: API reads from `sinceSeqId: 0` with no head, the client cold-starts from the rows endpoint on a `404`, and an expiry right after a cold start throws instead of looping. |
| `docs/testing.md` / `docs/testing/anti-patterns.md` AP-4 | Mock only at the external boundary | Pass | The platform cases drive the real contract mocks; no relative-path `vi.mock` added. |
| `docs/testing/anti-patterns.md` AP-1 | Assert behavior, not mock calls | Pass | `rowRequests` is asserted as the observable request sequence (`[0]`), paired with the rendered events and the thrown error, not as a bare call count. |
| `docs/testing/api-testing.md` | Route tests go through `setupApp()` + the ts-rest contract | Pass | The `sinceSeqId: 0` case uses `eventsClient().rows(...)` with `accept(..., [200])`. |
| `docs/testing/testing-external-behavior.md` | Construct and verify state through the external interface | Pass | The new API assertion compares two responses of the same endpoint; the platform cases assert projected events and request cursors. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

Two notes recorded rather than blocking:

- A retired-version head is still not re-archived automatically, so such a thread keeps reading through the raw-row path. This is the PR's stated position that version migration is an explicit rollout operation, and the merge gate already requires the current v1/v2 head count to be zero. With `392f00c` that state is also no longer able to spin the client: the cold start fails loudly instead.
- The `deleteS3Objects` throw on per-object `Errors` still has no direct test, and that helper is shared with other flows.

### Fallbacks

- `internal-chat-run-callback.service.ts:chatCallbackPayloadSchema.isGoalRun` + `zero-goal-queue-drain.service.ts:buildGoalChatCallbacks` — **removed**. Surface: DB/API, ~102 minutes. Evidence: `.passthrough()` keeps pending rows parseable; no production reader remains on `main`. Declared: **Yes**. Verdict: **justified**.
- `cron-snapshot-chat-events.service.ts:REWRITABLE_ARCHIVE_SCHEMA_VERSION` / `isMetadataOnlyRewrite` plus `putImmutableS3Object.refreshMetadata` — **removed**. Surface: persisted R2/DB snapshot state. Removal condition: the merge gate requiring zero current v1/v2 heads. Declared: **Yes**. Verdict: **justified**.
- `zero-chat-event-shared.service.ts` legacy `runEventId === undefined` branch plus the `"run-sequence"` conflict mode — **removed**. Surface: DB/API, ~102 minutes. Evidence: type evidence on both callers plus `chat_events_run_event_seq_unique`. Declared: **Yes**. Verdict: **justified**.
- `web-client-compatibility.json:minimumSupportedVersion` `0.707.0` → `0.708.2` — **kept**, as the shortening mechanism for old web clients that read `isGoalRun`. Surface: old web clients, ~2 days, shortened by `426`. Declared: **Yes**. Verdict: **justified**; ordering verified against `app-v0.708.2`.
- `workflow-chat-event-queue.service.ts:loadNextWorkflowQueueEvent` — **kept, re-justified**. Declared: **Yes**. Verdict: **justified**, no window.

The `404` cold-start branch added in this range is not a fallback: it handles a state the API produces by design for every thread the archiver has not reached, and it has no removal condition.

### Testing

- Coverage: **Adequate.** The snapshot-read GA path now has cold start with a snapshot, cold start without one, cursor expiry after a rebuild, cursor expiry after a cold start, cached-row initialization, and background row sync. API-side, the raw-row endpoint covers projection parity, the cold-start cursor, and the expired cursor. Author reports `zero-chat-event-snapshot` and `cron-snapshot-chat-events` green against a freshly migrated PostgreSQL database, all 38 `src/signals/chat-page` platform tests green, plus full-repository lint, type checks, knip, and formatting.
- Conventions: **Compliant.** Third-party-only mocks, real database, no fake timers, behavioral assertions, contract-level route tests.
- CI at this commit has not finished. This LGTM is a code review only; merge remains gated on every required check going green — `cli-e2e-02-playwright` in particular, which this range is intended to fix — and on the rollout gates in the PR summary.

e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 34844575636b22d774f7321ff86d6ab552ab7580
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Re-review at `3484457`. The code reviewed at `392f00c` is unchanged; this range only reverts the `ChatEventSnapshotRead` GA flip and the three test-infra pins that existed solely because the production default had flipped. `packages/core/src/feature-switch.ts`, `page-helper.ts`, `api-feature-switches.ts`, and `signals/__tests__/test-helpers.ts` drop out of the diff entirely, taking the PR from 33 to 29 files. The snapshot cron, R2 GC, cold-start cursor, reader contract, and the switch-gated signal suite all stay, so the deferred GA is a switch flip plus page-level coverage.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic: no race conditions, unhandled edge cases, or wrong assumptions | Pass | Reverting the switch changes no code path; the reader, cron, and GC logic reviewed at `392f00c` is byte-identical. The cron-side behavior is independent of the switch, so the retired rewrite path still fails closed for a pre-v3 head that gains a tail. |
| `REVIEW.md` | Security: no injection, hardcoded secrets, or missing auth on new surfaces | Pass | No new endpoint or auth path in this range. |
| `REVIEW.md` | Type safety: no `any`, no overly broad types | Pass | The reverted files return to `main`'s content; nothing widened. |
| `REVIEW.md` | Style and maintainability: no duplicated logic, no dead code, comments explain WHY | Pass | The three test-infra pins were only meaningful while the default was `true`; removing them with the flip avoids leaving fixtures that pin a value nothing sets. |
| `REVIEW.md` | Feature regression guard: replaced behavior keeps explicit coverage | Pass | `chat-event-snapshot-read.test.ts` calls `enableSnapshotRead()` explicitly, so all six snapshot-read cases still execute with the production default off. |
| `docs/bad-smell.md` | Zero tolerance for `any`, lint/type suppressions, dynamic imports | Pass | None present. |
| `docs/bad-smell.md` | Fail fast instead of silently absorbing errors | Pass | Unchanged: `deleteS3Objects` throws on per-object `Errors`, and an expiry after a server-derived cursor throws. |
| `docs/bad-smell.md` | Delete unused code aggressively (YAGNI) | Pass | This range is itself a deletion of scaffolding that no longer has a purpose. |
| `docs/fallback.md` §2 | A feature behind a non-GA switch needs no compatibility code | Pass | `ChatEventSnapshotRead` is off again, so the snapshot-read pipeline has no external users and this PR carries no compatibility branch for it. |
| `docs/fallback.md` §9 | Every fallback the diff introduces, keeps, or removes is declared in the PR summary | Pass | The `Fallbacks` section still lists all five entries; none of them depended on the switch state. |
| `docs/fallback.md` §10 | A removal PR states type, single-writer, production, or rollout evidence | Pass | Unchanged from the previous review. |
| `docs/deployment-compatibility.md` | Raise the client floor only after the corresponding app build is live | Pass | The floor raise is justified by the `isGoalRun` response-field removal, which is independent of the switch. `main` is at app `0.710.0` and `app-v0.708.2` is an existing tag. |
| `docs/deployment-compatibility.md` | Prefer additive changes at cross-version boundaries | Pass | The `sinceSeqId` widening to `nonnegative()` remains additive, and with the switch off only an opted-in client can send `0`, which narrows the documented drift window further. |
| `docs/deployment-compatibility.md` | No schema migration accompanies the contract change | Pass | No files under `turbo/packages/db/src/migrations/**`. |
| `docs/event-sourcing.md` | Persistent events reconcile optimistic events by a shared event ID; no failure-path rollback | Pass | Unchanged. |
| `docs/cache.md` RC-3 | No unbounded process-lifetime keyed caches | Pass | `chat-event-goal-run-ids.ts` and its module-level map remain deleted. |
| `docs/cache.md` RC-4 | State belongs to the narrowest lifecycle that owns it | Pass | Unchanged. |
| `docs/effect.md` | Render stays pure; work runs at the owning boundary | Pass | No render-path change in this range. |
| `.claude/skills/ccstate/SKILL.md` | Commands declare only the store access they use; no `void`-silenced promises in signals | Pass | Unchanged. |
| `docs/testing.md` | New behavior ships with integration tests | Pass | The switch-gated suite still covers cold start with and without a snapshot, cursor expiry after a rebuild and after a cold start, cached-row initialization, and background row sync. |
| `docs/testing.md` / `docs/testing/anti-patterns.md` AP-4 | Mock only at the external boundary | Pass | Reverting the default handler pin returns `api-feature-switches.ts` to a stateless default, which is the shape the module documents. |
| `docs/testing/anti-patterns.md` AP-5 | No fake timers | Pass | Unchanged. |
| `docs/testing/anti-patterns.md` AP-1 | Assert behavior, not mock calls | Pass | Unchanged. |
| `docs/testing/api-testing.md` | Route tests go through `setupApp()` + the ts-rest contract | Pass | Unchanged. |
| `docs/testing/testing-external-behavior.md` | Construct and verify state through the external interface | Pass | Unchanged. |
| `docs/testing/anti-patterns.md` AP-10 | Platform tests go through `detachedSetupPage`, not direct rendering | Pass | With the pins removed, `setupPage` again overrides the feature-switch GET handler only when a caller asks for switches, so a page test that installs its own handler keeps precedence. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

Notes recorded rather than blocking:

- The deferred GA is documented in the summary with the evidence that drove it: `cli-e2e-02-playwright` failed on every head of this branch with the same page state while `main` and every other branch passed the job, and no page-level test exercises the snapshot-read pipeline. Landing the pipeline switched off keeps that risk out of this PR and makes the follow-up a switch flip plus the missing coverage.
- The reader-side work in this PR (v3-only heads, cold-start cursor, `404` fallback) is now reachable only for opted-in clients. It stays here deliberately so the follow-up does not have to re-land it.
- `deleteS3Objects` throwing on per-object `Errors` still has no direct test, and that helper is shared with other flows.
- A retired-version head is still not re-archived automatically; the merge gate requiring zero current v1/v2 heads still applies, because the cron's fail-closed behavior does not depend on the client switch.

### Fallbacks

- `internal-chat-run-callback.service.ts:chatCallbackPayloadSchema.isGoalRun` + `zero-goal-queue-drain.service.ts:buildGoalChatCallbacks` — **removed**. Surface: DB/API, ~102 minutes. Evidence: `.passthrough()` keeps pending rows parseable; no production reader remains on `main`. Declared: **Yes**. Verdict: **justified**.
- `cron-snapshot-chat-events.service.ts:REWRITABLE_ARCHIVE_SCHEMA_VERSION` / `isMetadataOnlyRewrite` plus `putImmutableS3Object.refreshMetadata` — **removed**. Surface: persisted R2/DB snapshot state. Removal condition: the merge gate requiring zero current v1/v2 heads. Declared: **Yes**. Verdict: **justified**.
- `zero-chat-event-shared.service.ts` legacy `runEventId === undefined` branch plus the `"run-sequence"` conflict mode — **removed**. Surface: DB/API, ~102 minutes. Evidence: type evidence on both callers plus `chat_events_run_event_seq_unique`. Declared: **Yes**. Verdict: **justified**.
- `web-client-compatibility.json:minimumSupportedVersion` `0.707.0` → `0.708.2` — **kept**, as the shortening mechanism for old web clients that read `isGoalRun`. Surface: old web clients, ~2 days, shortened by `426`. Declared: **Yes**. Verdict: **justified**.
- `workflow-chat-event-queue.service.ts:loadNextWorkflowQueueEvent` — **kept, re-justified**. Declared: **Yes**. Verdict: **justified**, no window.

No fallback was added, removed, or re-scoped by this range.

### Testing

- Coverage: **Adequate.** Author reports platform `chat-page` and `lab-page` (43 tests), `zero-page` chat workflow/goal (31 tests), and `@vm0/core` (190 tests) green after the revert, plus full-repository lint, type checks, knip, and formatting. The API suites were verified at `392f00c` and are untouched here.
- Conventions: **Compliant.** No test was skipped or weakened to accommodate the revert; the snapshot-read suite still opts the switch on explicitly.
- CI at this commit has not finished. This LGTM is a code review only; merge remains gated on every required check going green and on the rollout gates in the PR summary. `cli-e2e-02-playwright` is now also the experiment that confirms or refutes the snapshot-read pipeline as the cause.

PR #25822, opened 2026-08-08
refactor(chat): stop writing web chat_event_asset_refs rows

e7h4n:
Web chat input events already carry their own ordered attachment list inside `chat_events.user_message` — `type: "file"` parts with `fileId`, `filenameSnapshot`, and `contentType`. The parallel `chat_event_asset_refs` rows written on the same web path were never read back, so this stops writing them.

## Why the web rows are unread

The only reader of `chat_event_asset_refs` is `loadSlackPromptAssets` in `slack-queued-launch-context.service.ts`, reachable exclusively through `loadSlackQueuedLaunchMaterial`. Both of its callers dispatch on the event's `contextType` and route web traffic elsewhere:

- `active-input-prompt.service.ts` — `switch (event.contextType)`, only `case "slack"` calls the Slack loader.
- `internal-chat-run-callback.service.ts` — `switch (contextType)`, `web` and `agent_run` go to `loadWebQueuedLaunchMaterial`.

A repo-wide search finds no other consumer: outside the schema and the baseline migration, `chat_event_asset_refs` appears only in `canonical-asset.service.ts` (write), `slack-queued-launch-context.service.ts` (the Slack read), `zero-chat-event.service.ts` (copy-on-claim), and the test-runtime-state fixture route. Nothing queries the `chat_event_asset_refs_asset_idx` reverse direction.

Slack is untouched: `attachCanonicalAssetsToEvent` and the `preserveAssetRefs` copy in `replaceChatEvent` both stay.

## What is kept

`registerCanonicalWebInputAssets` (renamed from `attachCanonicalWebInputAssetsToEvent`) still upserts the canonical `run_uploaded_files` rows with `chat_thread_id`, `classification: "input"`, and the `web-input` idempotency key. That column is what the artifact catalog's thread filter resolves through, so dropping it would have been a real regression. The rename drops the now-unused `eventId` and `replaceExisting` arguments, since the function no longer touches the event.

## Fallbacks

None. This removes a write, not a compatibility path.

Removal evidence per `docs/fallback.md` §10:

- **Type/code evidence** — the single reader sits behind a `contextType === "slack"` branch in both call paths, so no web-origin ref row is reachable.
- **Single-writer evidence** — web ref rows were written only by the function changed here.
- **Cross-version safety** — during rollout, an older API instance still writing web ref rows is harmless (nothing reads them), and Slack ref rows keep being written and read by both versions. Pre-existing rows are left in place; a separate cleanup can drop them once the table's remaining Slack-only role is settled.

## Testing

No new test. Asserting "web writes no ref rows" would be a tombstone under `docs/fallback.md` §1; the meaningful coverage is that the retained behavior still works, which existing suites already pin:

- `chat-files.bdd.test.ts` + `zero-integrations-slack-upload-complete.test.ts` — 22 passed. The Slack suite still asserts `readChatEventAssetRefIds` on the Slack path, proving the reader and its writer are intact.
- `artifact-catalog.bdd.test.ts` + `chat-events.bdd.test.ts` — 92 passed, covering the thread-scoped artifact resolution that depends on `run_uploaded_files.chat_thread_id`.
- `chat-threads.bdd.test.ts` + `chat-callbacks.bdd.test.ts` + `artifacts.bdd.test.ts` — 86 passed.

`pnpm -F api check-types`, targeted ESLint, and `pnpm knip --workspace apps/api` are clean.

### DIFF
--- a/turbo/apps/api/src/signals/services/canonical-asset.service.ts
+++ b/turbo/apps/api/src/signals/services/canonical-asset.service.ts
@@ -691,24 +691,27 @@ export async function attachCanonicalAssetsToEvent(
     .onConflictDoNothing();
 }
 
-export async function attachCanonicalWebInputAssetsToEvent(
+/**
+ * Registers web chat input files as canonical assets.
+ *
+ * Web input events carry their own ordered attachment list in
+ * `chat_events.user_message` (`type: "file"` parts with fileId, filename, and
+ * content type), and the only reader of `chat_event_asset_refs` is the Slack
+ * queued-launch loader, which `active-input-prompt` and
+ * `internal-chat-run-callback` reach only on the `contextType === "slack"`
+ * branch. Web ref rows were therefore write-only, so this path registers the
+ * canonical asset rows without also writing the join table.
+ */
+export async function registerCanonicalWebInputAssets(
   db: Db,
   args: {
-    readonly eventId: string;
     readonly chatThreadId: string;
     readonly userId: string;
     readonly orgId: string;
     readonly files: readonly ChatEventAttachFileMetadata[];
-    readonly replaceExisting?: boolean;
   },
 ): Promise<void> {
-  if (args.replaceExisting) {
-    await db
-      .delete(chatEventAssetRefs)
-      .where(eq(chatEventAssetRefs.chatEventId, args.eventId));
-  }
-  const assets: { readonly assetId: string; readonly position: number }[] = [];
-  for (const [position, file] of args.files.entries()) {
+  for (const file of args.files) {
     const [inserted] = await db
       .insert(runUploadedFiles)
       .values({
@@ -743,9 +746,7 @@ export async function attachCanonicalWebInputAssetsToEvent(
     if (!asset) {
       throw new Error("Canonical web input asset conflict is missing");
     }
-    assets.push({ assetId: asset.id, position });
   }
-  await attachCanonicalAssetsToEvent(db, args.eventId, assets);
 }
 
 interface PrepareCanonicalPublishedAssetArgs {
--- a/turbo/apps/api/src/signals/services/zero-chat-events.command.ts
+++ b/turbo/apps/api/src/signals/services/zero-chat-events.command.ts
@@ -102,7 +102,7 @@ import {
   chatThreadServiceTierFromCodex,
 } from "./zero-chat-thread-event.service";
 import { loadUserFeatureSwitchContext } from "./feature-switches.service";
-import { attachCanonicalWebInputAssetsToEvent } from "./canonical-asset.service";
+import { registerCanonicalWebInputAssets } from "./canonical-asset.service";
 import { resolveArtifactObject$ } from "./artifact-storage.service";
 import { loadOrgPlanCapabilities } from "./org-plan-entitlement-read.service";
 import { chatEventTypeIn } from "./zero-chat-event-type.service";
@@ -1436,8 +1436,7 @@ function appendUnassociatedUserMessage(params: {
         })
       : await insertChatEvent(tx, event, "id");
     if (inserted) {
-      await attachCanonicalWebInputAssetsToEvent(tx, {
-        eventId: inserted.id,
+      await registerCanonicalWebInputAssets(tx, {
         chatThreadId: params.threadId,
         userId: params.userId,
         orgId: params.orgId,
@@ -1554,8 +1553,7 @@ async function appendAssociatedUserMessage(params: {
         })
       : await insertChatEvent(tx, event, "id");
     if (inserted) {
-      await attachCanonicalWebInputAssetsToEvent(tx, {
-        eventId: inserted.id,
+      await registerCanonicalWebInputAssets(tx, {
         chatThreadId: params.threadId,
         userId: params.userId,
         orgId: params.orgId,
@@ -2703,13 +2701,11 @@ async function appendInsufficientCreditsEvents(params: {
 
     const createdAt = userMessage?.createdAt ?? userCreatedAt;
     if (userMessage) {
-      await attachCanonicalWebInputAssetsToEvent(tx, {
-        eventId: userMessage.id,
+      await registerCanonicalWebInputAssets(tx, {
         chatThreadId: params.prepared.thread.threadId,
         userId: params.userId,
         orgId: params.orgId,
         files: fileMetadata ?? [],
-        replaceExisting: params.body.revokesEventId !== undefined,
       });
     }
     if (userMessage && params.touchThreadSort) {


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 0c64242f8135eab7f06e424a44f4dbf903cd51e9
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Stops the web chat send path from writing `chat_event_asset_refs` join rows. `attachCanonicalWebInputAssetsToEvent` becomes `registerCanonicalWebInputAssets`, keeping the `run_uploaded_files` canonical upsert and dropping the now-unused `eventId` / `replaceExisting` arguments; three call sites in `zero-chat-events.command.ts` follow. Slack's writer, reader, and the claim-time `preserveAssetRefs` copy are untouched.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness: wrong assumptions, unhandled paths | Pass | The removed write is provably unread: `loadSlackPromptAssets` is reachable only via `loadSlackQueuedLaunchMaterial`, and both callers gate on context — `active-input-prompt.service.ts` `switch (event.contextType)` routes only `case "slack"`, `internal-chat-run-callback.service.ts:2800` routes `web`/`agent_run` to `loadWebQueuedLaunchMaterial`. The three changed call sites are web-chat send paths (`triggerSource: "web" \| "agent"` → `contextType` `web` / `agent_run`), never Slack. |
| `REVIEW.md` | Security: auth, injection, secrets | N/A | No auth, request-parsing, or credential surface changes; the diff only removes an internal join-table write. |
| `REVIEW.md` | Type safety: no `any`, no broad casts | Pass | No casts or `any` added; the arg interface narrows by deleting two fields. |
| `REVIEW.md` | Interface changes documented | Pass | Renamed export is internal to `apps/api`; repo-wide search shows no remaining `attachCanonicalWebInputAssetsToEvent` reference outside a stale `.typecheck` artifact. `attachCanonicalAssetsToEvent` stays exported for `canonical-slack-ingress-processor.service.ts:364`. |
| `REVIEW.md` | Feature regression guard: existing flow covered | Pass | The retained behavior — canonical `run_uploaded_files` rows with `chat_thread_id` — is what `artifact-catalog.service.ts:817` resolves thread-scoped artifacts through, and `artifact-catalog.bdd.test.ts` still covers it. Attachment rendering is unaffected because it reads `user_message` `type: "file"` parts, not this table. |
| `REVIEW.md` | DB/JSONB review gate | Pass | No schema, migration, or contract change. Pre-existing web ref rows are left in place and remain valid rows of an unchanged table. |
| `REVIEW.md` | Testing coverage for the commit type | Pass | `refactor:` commit, so existing tests must still cover the code; 200 tests across seven suites pass (detail under Testing). |
| `docs/bad-smell.md` | No `any`, suppressions, dynamic imports, hardcoded URLs | Pass | None added. |
| `docs/bad-smell.md` | Flag over-engineered or defensive error handling | Pass (with observation) | The `canonicalAssetByIdentity` fallback plus invariant throw survives but now feeds only an assertion; see the non-blocking observation below for why keeping it is still defensible. |
| `docs/bad-smell.md` | Delete unused code aggressively (YAGNI) | Pass | `eventId`, `replaceExisting`, the `assets` accumulator, and the `position` index are all removed rather than left dangling; `knip --workspace apps/api` is clean. |
| `docs/fallback.md` §9 | PR summary declares every fallback, or `Fallbacks: none` | Pass | PR body carries a `## Fallbacks` section stating none, since the change removes a write rather than adding a compatibility path. |
| `docs/fallback.md` §10 | A removal states type / single-writer / production / rollout evidence | Pass | PR body gives type-and-code evidence (the reader's `contextType === "slack"` gate) and single-writer evidence (web ref rows written only by the changed function), and I re-verified both against the tree at this HEAD. |
| `docs/fallback.md` §1 | No tombstone tests asserting removed behavior stays removed | Pass | Deliberately no `expect(refs).toEqual([])` test; the PR body states this reasoning explicitly. |
| `docs/deployment-compatibility.md` | Old code after change / new code before change both safe | Pass | Old API instances that still write web ref rows during rollout are harmless because nothing reads them; new instances that stop writing are harmless because the only reader is Slack-gated and Slack still writes. No migration is involved, so neither DB/API direction applies. |
| `docs/testing.md` | Integration tests at entry points; existing coverage must still hold | Pass | Coverage comes from route-level BDD suites through the real Hono app; no unit test was added for the changed service function. |
| `.claude/skills/ccstate/SKILL.md` | ccstate contract: no accessor escape, signal ownership, `accept` for client calls | N/A | The diff touches plain service functions taking an explicit `db`/`tx`; no `computed`, `command`, `Store`, `AbortSignal`, or `zeroClient$` call is added or changed. |
| `docs/event-sourcing.md` | Optimistic events reconcile by shared ID; no failure-path rollback | N/A | No optimistic-event path involved; this is a server-side write to an auxiliary join table, and the chat event row itself is unchanged. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

Non-blocking observation: after this change, `canonicalAssetByIdentity` and the `Canonical web input asset conflict is missing` throw exist only to assert an invariant — nothing consumes `asset.id` anymore — so a conflicting file now costs one extra indexed SELECT purely for that assertion. I checked whether to flag this as a simplification: the conflict branch runs only when the same `file.id` is registered twice under the `web-input` idempotency scope (the re-send / insufficient-credits replay paths), not on first upload, so the cost is off the hot path; and `docs/fallback.md` §10 explicitly asks a removal to "keep the invariant throw so monitoring surfaces a violation if the assumption ever breaks". Keeping it is the doc-aligned choice, so this is recorded rather than requested.

Also checked and found benign: dropping the `replaceExisting` branch means a replaced web event no longer deletes ref rows for its event id. `replaceChatEvent` is called with `preserveAssetRefs: false` on these paths and returns a freshly inserted row, so the id carries no refs to delete; and any rows left over from before this deploy sit on web-context events that the Slack-gated reader can never reach.

### Fallbacks

None. The diff removes a redundant write and introduces no compatibility branch. Pre-existing `chat_event_asset_refs` rows written by the web path before this deploy are inert data, not a code path — they are unread for the same `contextType === "slack"` reason, and the PR body notes a separate cleanup can drop them once the table's remaining Slack-only role is settled.

### Testing

- Coverage: Adequate for a `refactor:` commit. `chat-files.bdd.test.ts` + `zero-integrations-slack-upload-complete.test.ts` (22 passed) — the Slack suite still asserts `readChatEventAssetRefIds` on the Slack path, which pins that the reader and its writer survived. `artifact-catalog.bdd.test.ts` + `chat-events.bdd.test.ts` (92 passed) cover the thread-scoped artifact resolution that depends on the retained `run_uploaded_files.chat_thread_id`. `chat-threads.bdd.test.ts` + `chat-callbacks.bdd.test.ts` + `artifacts.bdd.test.ts` (86 passed).
- Conventions: Compliant. No test files changed, so no testing anti-pattern is introduced, and the decision not to add a "web writes no refs" assertion follows `docs/fallback.md` §1 rather than leaving a coverage gap.

PR #25829, opened 2026-08-08
feat(web-chat): gate follow-up card rail to mobile devices only

e7h4n:
## Summary

The responsive follow-up **card rail** should only render on actual mobile devices — the smallest responsive tier — and must stay the flat vertical list on desktop **even when the window is dragged narrow**.

Previously the cards-only-on-small decision used a `@[900px]` **container query**, so a desktop browser resized narrower than 900px incorrectly switched to the card rail. This switches the decision to the same device heuristic the composer already uses to gate its **auto-focus** (`isMobileTextInputDevice()`): coarse-pointer touch text-entry devices get the card rail; fine-pointer desktops never do, regardless of viewport width.

## Changes

- `zero-chat-thread-page.tsx` — `RecommendedFollowupList` now computes `showFollowupCards = featureEnabled && isMobileTextInputDevice()` and applies the card-rail classes only when true. Removed the `@[900px]` container-query overrides that allowed narrow desktops to show cards.
- `chat-recommended-follow-ups.test.tsx` — the card-rail test now mocks a coarse-pointer (mobile) device so the rail renders; added a test asserting a fine-pointer desktop keeps the flat list even with the feature switch on.

## Verification

- Targeted vitest: `chat-recommended-follow-ups.test.tsx` — 10/10 pass (incl. mobile card rail + desktop flat list cases).
- `tsc -p tsconfig.production.json` and `tsconfig.tests.json` — clean.
- `oxlint` on both changed files — clean.

## E2E follow-up

- `e2e/playwright/tests/chat.spec.ts` — the existing card-rail geometry test ran in the default Desktop Chrome context, which is a fine-pointer device, so it broke once the rail became device-gated. It now runs inside a `test.describe` with `test.use({ hasTouch: true })` (Chromium touch emulation reports `(pointer: coarse)` and drops `(any-pointer: fine)`), and its final wide-viewport assertion checks that a touch device keeps the cards at 1440px instead of collapsing them into full-width rows.
- Added `keeps the flat follow-up list in a narrow desktop window`: the default fine-pointer context at 390x844 must render full-width rows stacked vertically, which is the regression this PR fixes.

## Fallbacks

Fallbacks: none. The change is a client-side layout decision behind the non-GA staff-only `FeatureSwitchKey.ResponsiveFollowupCards` switch, so no cross-version compatibility path is introduced, kept, or removed.

### DIFF
--- a/e2e/playwright/tests/chat.spec.ts
+++ b/e2e/playwright/tests/chat.spec.ts
@@ -389,7 +389,124 @@ test("chat composer keeps the Send button inside on narrow screens", async ({
   });
 });
 
-test("responsive follow-up rail aligns its edges and equalizes card heights", async ({
+// The card rail only renders on coarse-pointer text-entry devices, so this
+// group emulates touch instead of relying on the viewport width alone.
+test.describe("mobile follow-up card rail", () => {
+  test.use({ hasTouch: true });
+
+  test("responsive follow-up rail aligns its edges and equalizes card heights", async ({
+    page,
+  }) => {
+    await enableResponsiveFollowupCards(page);
+    await page.setViewportSize({ width: 390, height: 844 });
+    await page.goto(appUrl);
+    await page.waitForURL(/agents\/.*\/chat/, { timeout: 30_000 });
+
+    const agentId = new URL(page.url()).pathname.match(
+      /^\/agents\/([^/]+)\/chat\/?$/,
+    )?.[1];
+    if (!agentId) {
+      throw new Error("Could not resolve the active agent from the chat URL");
+    }
+    await mockResponsiveFollowupThread(page, agentId);
+    await page.goto(
+      new URL(`/chats/${responsiveFollowupThreadId}`, appUrl).href,
+    );
+
+    const rail = page.getByRole("group", { name: "Keep going" });
+    const cards = responsiveFollowupPrompts.map((prompt) => {
+      return page.getByRole("button", { name: prompt, exact: true });
+    });
+    await expect(rail).toBeVisible();
+    for (const card of cards) {
+      await expect(card).toBeVisible();
+    }
+
+    await expect
+      .poll(async () => {
+        const railBox = await rail.boundingBox();
+        const firstBox = await cards[0].boundingBox();
+        if (!railBox || !firstBox) {
+          return Number.POSITIVE_INFINITY;
+        }
+        return Math.abs(firstBox.x - railBox.x);
+      })
+      .toBeLessThan(1);
+    await expect
+      .poll(async () => {
+        const boxes = await Promise.all(
+          cards.map((card) => card.boundingBox()),
+        );
+        if (boxes.some((box) => box === null)) {
+          return Number.POSITIVE_INFINITY;
+        }
+        const heights = boxes.map((box) => box!.height);
+        return Math.max(...heights) - Math.min(...heights);
+      })
+      .toBeLessThan(1);
+
+    await cards[1].evaluate((element) => {
+      element.scrollIntoView({ block: "nearest", inline: "center" });
+    });
+    await expect
+      .poll(async () => {
+        const railBox = await rail.boundingBox();
+        const middleBox = await cards[1].boundingBox();
+        if (!railBox || !middleBox) {
+          return Number.POSITIVE_INFINITY;
+        }
+        const railCenter = railBox.x + railBox.width / 2;
+        const cardCenter = middleBox.x + middleBox.width / 2;
+        return Math.abs(cardCenter - railCenter);
+      })
+      .toBeLessThan(2);
+
+    await rail.evaluate((element) => {
+      element.scrollLeft = element.scrollWidth;
+    });
+    await expect
+      .poll(async () => {
+        const railBox = await rail.boundingBox();
+        const lastBox = await cards[2].boundingBox();
+        if (!railBox || !lastBox) {
+          return Number.POSITIVE_INFINITY;
+        }
+        return Math.abs(
+          lastBox.x + lastBox.width - (railBox.x + railBox.width),
+        );
+      })
+      .toBeLessThan(1);
+
+    // The rail is a device decision, so a wide viewport on the same touch
+    // device keeps the cards instead of collapsing them into full-width rows.
+    await page.setViewportSize({ width: 1440, height: 900 });
+    await expect
+      .poll(async () => {
+        const railBox = await rail.boundingBox();
+        const firstBox = await cards[0].boundingBox();
+        if (!railBox || !firstBox) {
+          return 0;
+        }
+        return railBox.width - firstBox.width;
+      })
+      .toBeGreaterThan(100);
+
+    await page.keyboard.press("Tab");
+    await cards[0].focus();
+    await expect
+      .poll(async () => {
+        return cards[0].evaluate((element) => {
+          return (
+            element.matches(":focus-visible") &&
+            getComputedStyle(element).boxShadow !== "none"
+          );
+        });
+      })
+      .toBe(true);
+  });
+});
+
+test("keeps the flat follow-up list in a narrow desktop window", async ({
   page,
 }) => {
   await enableResponsiveFollowupCards(page);
@@ -406,90 +523,39 @@ test("responsive follow-up rail aligns its edges and equalizes card heights", as
   await mockResponsiveFollowupThread(page, agentId);
   await page.goto(new URL(`/chats/${responsiveFollowupThreadId}`, appUrl).href);
 
-  const rail = page.getByRole("group", { name: "Keep going" });
-  const cards = responsiveFollowupPrompts.map((prompt) => {
+  const list = page.getByRole("group", { name: "Keep going" });
+  const rows = responsiveFollowupPrompts.map((prompt) => {
     return page.getByRole("button", { name: prompt, exact: true });
   });
-  await expect(rail).toBeVisible();
-  for (const card of cards) {
-    await expect(card).toBeVisible();
+  await expect(list).toBeVisible();
+  for (const row of rows) {
+    await expect(row).toBeVisible();
   }
 
+  // A fine-pointer window dragged this narrow keeps the flat vertical list:
+  // every follow-up spans the full width instead of becoming a card.
   await expect
     .poll(async () => {
-      const railBox = await rail.boundingBox();
-      const firstBox = await cards[0].boundingBox();
-      if (!railBox || !firstBox) {
+      const listBox = await list.boundingBox();
+      const boxes = await Promise.all(rows.map((row) => row.boundingBox()));
+      if (!listBox || boxes.some((box) => box === null)) {
         return Number.POSITIVE_INFINITY;
       }
-      return Math.abs(firstBox.x - railBox.x);
+      return Math.max(
+        ...boxes.map((box) => Math.abs(box!.width - listBox.width)),
+      );
     })
-    .toBeLessThan(1);
+    .toBeLessThan(2);
   await expect
     .poll(async () => {
-      const boxes = await Promise.all(cards.map((card) => card.boundingBox()));
+      const boxes = await Promise.all(rows.map((row) => row.boundingBox()));
       if (boxes.some((box) => box === null)) {
-        return Number.POSITIVE_INFINITY;
+        return 0;
       }
-      const heights = boxes.map((box) => box!.height);
-      return Math.max(...heights) - Math.min(...heights);
+      const tops = boxes.map((box) => box!.y);
+      return Math.max(...tops) - Math.min(...tops);
     })
-    .toBeLessThan(1);
-
-  await cards[1].evaluate((element) => {
-    element.scrollIntoView({ block: "nearest", inline: "center" });
-  });
-  await expect
-    .poll(async () => {
-      const railBox = await rail.boundingBox();
-      const middleBox = await cards[1].boundingBox();
-      if (!railBox || !middleBox) {
-        return Number.POSITIVE_INFINITY;
-      }
-      const railCenter = railBox.x + railBox.width / 2;
-      const cardCenter = middleBox.x + middleBox.width / 2;
-      return Math.abs(cardCenter - railCenter);
-    })
-    .toBeLessThan(2);
-
-  await rail.evaluate((element) => {
-    element.scrollLeft = element.scrollWidth;
-  });
-  await expect
-    .poll(async () => {
-      const railBox = await rail.boundingBox();
-      const lastBox = await cards[2].boundingBox();
-      if (!railBox || !lastBox) {
-        return Number.POSITIVE_INFINITY;
-      }
-      return Math.abs(lastBox.x + lastBox.width - (railBox.x + railBox.width));
-    })
-    .toBeLessThan(1);
-
-  await page.setViewportSize({ width: 1440, height: 900 });
-  await expect
-    .poll(async () => {
-      const railBox = await rail.boundingBox();
-      const firstBox = await cards[0].boundingBox();
-      if (!railBox || !firstBox) {
-        return Number.POSITIVE_INFINITY;
-      }
-      return Math.abs(firstBox.width - railBox.width);
-    })
-    .toBeLessThan(2);
-
-  await page.keyboard.press("Tab");
-  await cards[0].focus();
-  await expect
-    .poll(async () => {
-      return cards[0].evaluate((element) => {
-        return (
-          element.matches(":focus-visible") &&
-          getComputedStyle(element).boxShadow !== "none"
-        );
-      });
-    })
-    .toBe(true);
+    .toBeGreaterThan(0);
 });
 
 test("image lightbox centers and pans across the full viewer", async ({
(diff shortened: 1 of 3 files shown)


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 54db2dd0e99a54414874d9bf8eebe64568f0f49a
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

`RecommendedFollowupList` now decides the follow-up card rail with `responsiveFollowupCards && isMobileTextInputDevice()` instead of a `@[900px]` container query, so a fine-pointer desktop keeps the flat vertical list at any window width while coarse-pointer touch devices get the card rail. The container-query overrides are deleted from all four class blocks, and the platform page test suite gains a fine-pointer desktop case plus a coarse-pointer mock on the existing card-rail case. The feature stays behind the staff-only, non-GA `FeatureSwitchKey.ResponsiveFollowupCards` switch (`enabled: false`, `enabledOrgIdHashes: STAFF_ORG_ID_HASHES`).

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic, no unhandled edge cases | Pass | `zero-chat-thread-page.tsx:4112` — `showFollowupCards` short-circuits on the switch first; `isMobileTextInputDevice()` already guards `typeof window === "undefined"` and `window.visualViewport` absence. |
| `REVIEW.md` | Security (injection, secrets, auth) | N/A | Diff is a client-side layout gate plus a page test; no endpoint, credential, or user-input sink is touched. |
| `REVIEW.md` | Type safety, no `any` | Pass | Both changed files add no casts; `isMobileTextInputDevice(): boolean` is consumed directly. |
| `REVIEW.md` | Style and maintainability, comments explain WHY | Pass | `zero-chat-thread-page.tsx:4109-4111` explains why container width is not the deciding factor; no duplicated logic introduced. |
| `REVIEW.md` | `feat:` commit must carry integration tests | Pass | `chat-recommended-follow-ups.test.tsx` adds "keeps the flat list on desktop even with a narrow window", which fails against the removed container-query implementation (old card branch emitted `flex-[0_0_min(...)`). |
| `REVIEW.md` | Feature regression guard: changed existing behavior needs coverage and containment | Pass | The behavior change (desktop stops showing cards) is covered by the new desktop test and is contained by the non-GA `ResponsiveFollowupCards` switch. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, contract, Zod response, or `jsonb().$type<>()` change in the diff. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | No `any`, no `eslint-disable`/`ts-expect-error`/`prettier-ignore` in either file. |
| `docs/bad-smell.md` | Static imports only, no dynamic `import()` | Pass | `zero-chat-thread-page.tsx:119` adds a static import of `isMobileTextInputDevice`. |
| `docs/bad-smell.md` | No hardcoded config or fail-fast violations | Pass | Device decision comes from the existing shared helper; no URLs or env values added. |
| `docs/fallback.md` §2 | Non-GA feature switch needs no compatibility code | Pass | The switch is `enabled: false` + staff hashes, and the PR adds no dual-path or legacy-shape tolerance for it. |
| `docs/fallback.md` §1 | No negative tests asserting removed behavior stays removed | Pass | The new desktop test pins current product behavior (flat list on desktop), not the absence of deleted code. |
| `docs/fallback.md` §9 | Every fallback declared in the PR summary | Pass (with nit) | Diff introduces, keeps, and removes no fallback; the summary has no explicit `Fallbacks: none` line — see Fallbacks section. |
| `docs/effect.md` | Render stays pure: no Store writes, I/O, timers, listeners, resource creation | Pass | `isMobileTextInputDevice()` and its callees only read `window.matchMedia`, `navigator`, and `window.visualViewport`; no mutation, listener, or timer. See non-blocking note 3 on reactivity. |
| `docs/effect.md` | Work runs at the right boundary (computed / command / setup / `onRef`) | Pass | The value is a pure per-render derivation used for class selection; no new command, ref, or effect is introduced. |
| `docs/cache.md` RC-1 | No render-phase side effects | Pass | Same evidence as above: read-only browser queries during render. |
| `docs/cache.md` RC-2 | No render-time signal allocation | Pass | No `state`/`computed`/`command` is created in `RecommendedFollowupList`. |
| `docs/cache.md` RC-3/RC-4/RC-5 | No unbounded caches, owner mismatch, or duplicate sources of truth | Pass | No module-level map, no new state ownership; the switch remains read via `useGet(featureSwitch$)`. |
| `.claude/skills/ccstate/SKILL.md` | DOM callbacks use `detach(..., Reason.DomCallback)`; no manual memo layer over `computed` | Pass | The existing `onClick` path is unchanged; no new async DOM callback or cache added. |
| `docs/testing.md` | Integration tests only, mock at the boundary (no `../` `vi.mock`) | Pass | Tests use `context.mocks.browser.matchMedia` (a `window` spy in `test-mocks.ts`); no internal-module mocking. |
| `docs/testing.md` | No fake timers, no direct `fetch` stubbing, assert behavior not mock calls | Pass | No `vi.useFakeTimers()`, no `vi.stubGlobal("fetch")`, no `toHaveBeenCalled()` as the sole assertion. |
| `docs/testing.md` | New user-facing feature gated by a `FeatureSwitchKey` | Pass | Rendering stays behind `FeatureSwitchKey.ResponsiveFollowupCards`. |
| `docs/testing/app-testing.md` | Page tests enter through `detachedSetupPage` and configure mocks via `context.mocks` before setup | Pass | Both tests call `context.mocks.browser.matchMedia(...)` before a non-awaited `detachedSetupPage`. |
| `docs/testing/patterns.md` | Prefer page-level integration over unit tests; reuse shared mock helpers | Pass | The new case is a page test reusing `mockChatLifecycle` and the shared browser mock helper. |
| `docs/testing/anti-patterns.md` AP-1/AP-2 | Do not assert mock calls; do not stub `fetch` | Pass | Assertions target rendered DOM, and HTTP stays on the existing MSW-backed lifecycle mock. |
| `docs/testing/testing-external-behavior.md` | Platform tests must not assert on CSS classes or other internal implementation | Finding (P1) | `chat-recommended-follow-ups.test.tsx` asserts `rows[0]?.className` contains `w-full` and not `flex-[0_0_min` — see P1-1. |
| `docs/testing/testing-external-behavior.md` | Construct and verify each case from the user-visible surface | Finding (P1) | The coarse-pointer branch is exercised but never verified: no assertion distinguishes card rail from flat list on mobile — see P1-2. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- `turbo/apps/platform/src/views/zero-page/__tests__/chat-recommended-follow-ups.test.tsx` (desktop test, `expect(rows[0]?.className).toContain("w-full")` / `.not.toContain("flex-[0_0_min")`): `docs/testing/testing-external-behavior.md` states platform tests must not "assert on query cache, component state, CSS classes, or whether an internal callback was called". These assertions freeze Tailwind class strings, so a purely visual refactor breaks the test. They are also uneven in strength: `toContain("w-full")` would have passed against the old implementation too, because the removed card branch contained `@[900px]:w-full`; only `not.toContain("flex-[0_0_min")` actually discriminates old from new. Suggest exposing a stable page-visible marker on the group (for example `data-followup-layout="cards" | "list"`) and asserting on that instead.
- `turbo/apps/platform/src/views/zero-page/__tests__/chat-recommended-follow-ups.test.tsx` ("preserves follow-up content and selection when the card rail is enabled"): the coarse-pointer mock is added, but the test only asserts that the prompts are visible, named, and clickable — assertions that hold for the flat list as well. No test pins the positive branch, so an implementation where `showFollowupCards` is never true (helper regression, switch read regression, class block dropped) would leave the whole suite green while the mobile card rail silently disappears. Suggest asserting the mobile layout through the same stable marker proposed above.

### Fallbacks

- None. The diff introduces, keeps, and removes no compatibility path: it is a client-side layout decision behind the non-GA `FeatureSwitchKey.ResponsiveFollowupCards` switch (`enabled: false`, staff org hashes), so `docs/fallback.md` §2 applies and no rollout window is created. The pre-existing `?? false` on the feature-switch map read is unchanged context, not a new fallback.
- Nit: per `docs/fallback.md` §9, add an explicit `Fallbacks: none` line to the PR summary so a later audit is trustworthy. Not treated as a blocking finding because the diff contains no fallback to declare.

### Testing

- Coverage: Adequate for the behavior this PR changes. The new desktop case is a real regression test for the reported bug (a narrow desktop window no longer flips to cards) and fails against the previous container-query implementation. Gaps are the unverified mobile positive branch (P1-2) and the class-level assertion surface (P1-1).
- Conventions: Compliant with the mock-boundary, MSW, no-fake-timers, `detachedSetupPage`, and feature-switch rules; the two P1 items above are the exceptions, both against `docs/testing/testing-external-behavior.md`.
- Non-blocking notes:
  1. The card branch container classes changed from `... @[900px]:-mx-2 ...` to an unconditional leading `-mx-2`, so the mobile card rail now carries the `-0.5rem` horizontal margin it previously only received at desktop width. If that edge alignment is intentional, it is worth one line in the PR summary; if not, it is an unintended mobile visual shift.
  2. `turbo/packages/core/src/feature-switch.ts` still describes the switch as rendering the rail "in narrow chat layouts". After this PR the trigger is the device class, not layout width; updating the description keeps the staff switch list accurate.
  3. `isMobileTextInputDevice()` is read during render without a subscription. For the common cases (pure touch device, pure fine-pointer desktop) the value is stable, so this is fine. On a coarse+fine hybrid that is not iOS, the helper falls through to `softwareKeyboardOccludesViewport()`, which is transient state; the layout would then reflect whatever was true at the last render and would not update on change. Acceptable for a staff-gated cosmetic gate, but worth knowing before GA.

e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 2c65f782db12892b127145bcde85993a7bde12d6
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Re-review at head `2c65f78`, which adds the Playwright follow-up to the layout change reviewed at `54db2dd`. `RecommendedFollowupList` still gates the card rail on `responsiveFollowupCards && isMobileTextInputDevice()` with the `@[900px]` container queries removed. The new commit moves the existing rail geometry test into a `test.describe` with `test.use({ hasTouch: true })`, replaces its now-invalid wide-viewport assertion (cards used to collapse into rows at `@[900px]`) with one asserting a touch device keeps cards at 1440px, and adds a fine-pointer test that a 390x844 desktop window renders full-width rows stacked vertically. The PR summary now carries an `E2E follow-up` section and `Fallbacks: none`.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic, no unhandled edge cases | Pass | `zero-chat-thread-page.tsx:4112` unchanged since the prior review; `isMobileTextInputDevice()` guards missing `window` and `visualViewport`. |
| `REVIEW.md` | Security (injection, secrets, auth) | N/A | No endpoint, credential, or user-input sink; the e2e commit only changes browser context options and assertions. |
| `REVIEW.md` | Type safety, no `any` | Pass | `cd e2e && pnpm check-types` passes locally; `test.use({ hasTouch: true })` is typed, and the new assertions reuse the file's existing `box!` narrowing style. |
| `REVIEW.md` | Style and maintainability | Pass | The describe block carries a comment explaining why touch is emulated rather than relying on viewport width. |
| `REVIEW.md` | `feat:` commit must carry integration tests | Pass | Platform page test `keeps the flat list on desktop even with a narrow window` plus the new Playwright `keeps the flat follow-up list in a narrow desktop window`; both fail against the removed container-query implementation. |
| `REVIEW.md` | Feature regression guard: changed existing behavior needs coverage and containment | Pass | Both directions are now pinned: touch keeps the rail (`mobile follow-up card rail` describe), fine pointer keeps the flat list. Containment remains the non-GA `ResponsiveFollowupCards` switch. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, contract, Zod response, or `jsonb().$type<>()` change. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions, static imports only | Pass | Neither commit adds casts, suppressions, or dynamic `import()`. |
| `docs/bad-smell.md` | No hardcoded URLs or environment values | Pass | The new e2e test reuses `appUrl`, `responsiveFollowupThreadId`, and the existing route mocks. |
| `docs/fallback.md` §2 | Non-GA feature switch needs no compatibility code | Pass | Switch is `enabled: false` with staff org hashes; no dual path added. |
| `docs/fallback.md` §1 | No negative tests asserting removed behavior stays removed | Pass | The wide-viewport assertion that pinned the deleted `@[900px]` collapse was replaced, not kept alongside the new behavior. |
| `docs/fallback.md` §9 | Every fallback declared in the PR summary | Pass | Summary now states `Fallbacks: none`, and the diff contains none. |
| `docs/effect.md` | Render stays pure; work runs at the right boundary | Pass | The helper only reads `matchMedia`, `navigator`, and `visualViewport`; no command, ref, or effect added. |
| `docs/cache.md` RC-1..RC-5 | No render side effects, render-time signal allocation, unbounded caches, owner mismatch, or duplicate state | Pass | No signal or state changes in either commit. |
| `.claude/skills/ccstate/SKILL.md` | DOM callbacks use `detach(..., Reason.DomCallback)`; no manual memo over `computed` | Pass | Existing `onClick` path untouched. |
| `docs/testing.md` | Integration tests, mock at the boundary, no fake timers, assert behavior not mock calls | Pass | Platform tests use `context.mocks.browser.matchMedia`; Playwright uses `page.route` at the network boundary, as the file already did. |
| `docs/testing.md` | New user-facing feature gated by a `FeatureSwitchKey` | Pass | Both Playwright cases call `enableResponsiveFollowupCards(page)` and rendering stays behind the switch. |
| `docs/testing/app-testing.md` | Page tests enter through `detachedSetupPage`, mocks configured via `context.mocks` before setup | Pass | Unchanged from the prior review; both platform cases follow it. |
| `docs/testing/patterns.md` | Prefer page-level integration; reuse shared mock helpers | Pass | The new Playwright case reuses `enableResponsiveFollowupCards` and `mockResponsiveFollowupThread`. |
| `docs/testing/anti-patterns.md` AP-1/AP-2 | Do not assert mock calls; do not stub `fetch` | Pass | Assertions read rendered DOM geometry; HTTP interception stays on Playwright routing and MSW. |
| `docs/testing/cli-e2e-testing.md` | Deployed E2E starts at a supported product entry point and asserts user-visible output | Pass | Both cases sign in through the shared fixture, navigate to a chat thread URL, and assert bounding-box geometry; no `/api/test/*` route, test token, or DB access. |
| `docs/testing/cli-e2e-testing.md` | Playwright owns the deployed product journey; deterministic matrices belong lower | Pass | Device-class branching is expressed through a browser context option rather than a test-only app hook. |
| `docs/testing/testing-external-behavior.md` | Construct and verify each case from the user-visible surface | Pass | The Playwright pair verifies the rail/list contract through real layout geometry instead of implementation details. |
| `docs/testing/testing-external-behavior.md` | Platform tests must not assert on CSS classes | Finding (P1) | `chat-recommended-follow-ups.test.tsx` still asserts `rows[0]?.className` contains `w-full` and not `flex-[0_0_min` — see P1-1. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- `turbo/apps/platform/src/views/zero-page/__tests__/chat-recommended-follow-ups.test.tsx` (desktop case): the assertions still read Tailwind class substrings, which `docs/testing/testing-external-behavior.md` rules out for platform tests ("Do not assert on ... CSS classes"). `toContain("w-full")` is also non-discriminating, because the removed card branch contained `@[900px]:w-full`; only `not.toContain("flex-[0_0_min")` separates old from new. Risk is now lower because head `2c65f78` pins the same contract in Playwright through real geometry, so this can be resolved by either swapping in a stable marker (for example `data-followup-layout="cards" | "list"`) or by leaning on the e2e pair and simplifying the jsdom assertions.

Resolved since `54db2dd`: the mobile branch is no longer unverified. `mobile follow-up card rail` runs under Chromium touch emulation, which reports `(pointer: coarse)` and no `(any-pointer: fine)`, and asserts equal card heights, rail edge alignment, snap centering, and that a 1440px touch viewport keeps cards narrower than the rail.

### Fallbacks

- None. Neither commit introduces, keeps, or removes a compatibility path. The feature is behind the non-GA staff-only `FeatureSwitchKey.ResponsiveFollowupCards` switch, so `docs/fallback.md` §2 applies and no rollout window is created. The PR summary now declares `Fallbacks: none` as §9 requires.

### Testing

- Coverage: Adequate. jsdom covers the device decision on both branches at the page level; Playwright covers the rendered contract on both branches through geometry (touch keeps the rail at 390px and 1440px, fine pointer keeps full-width stacked rows at 390x844). The replaced wide-viewport assertion previously pinned the deleted `@[900px]` collapse, so removing it is correct rather than a coverage loss.
- Conventions: Compliant. `cd e2e && pnpm check-types` passes; the touch context is a supported Playwright option rather than a test-only application hook; no fake timers, no internal-module mocking, no `toHaveBeenCalled()` assertions. The single exception is the jsdom class assertion noted in P1-1.
- Non-blocking notes carried over from `54db2dd`:
  1. The card branch now applies `-mx-2` at all widths, where it was previously `@[900px]:-mx-2`, so the mobile rail gains the `-0.5rem` horizontal margin. Worth confirming that the edge alignment is intended.
  2. `turbo/packages/core/src/feature-switch.ts` still describes the switch as applying "in narrow chat layouts"; the trigger is now the device class.
  3. `isMobileTextInputDevice()` is read during render without a subscription. Stable for pure touch and pure fine-pointer devices; on a non-iOS coarse+fine hybrid it falls through to `softwareKeyboardOccludesViewport()`, which is transient state that will not re-render on change. Acceptable while staff-gated, worth revisiting before GA.

PR #25835, opened 2026-08-08
fix(api): replay web chat history only when the session rotates

e7h4n:
## Problem

Run [`d0a3b0c5`](https://app.vm0.ai/activities/d0a3b0c5-827a-4713-855d-13f6e2d6c194) failed 1 second after start:

```
execution: Pi standby child argv/env too large:
env value too large: VM0_PI_SYSTEM_PROMPT length=216392 max=131071
```

Breaking down that 216,392-byte system prompt:

| part | bytes | share |
| --- | --- | --- |
| `# Web Chat Run Context` | 150,710 | 69.6% |
| Skill catalog (~100 skills) | ~32,800 | ~15% |
| `# Agent Tools` | 17,349 | 8.0% |
| MEMORY.md prefix (capped at 8 KiB) | ~8,700 | ~4% |
| Agent instructions | 3,212 | 1.5% |
| Pi base template | 2,655 | 1.2% |

Every other part is either capped or naturally stable. Only the replayed run
context grows with the conversation, and nothing bounded it: the per-message
cap of 4,000 chars applied, but a single prior run contributed 87,664 bytes
from 75 chat events (74 assistant messages, all of them intermediate progress
narration from one agentic run). Three chatty runs accounted for 143 KB of
the prompt.

## Change

- Prior rounds are replayed only when `sessionAction === "rotated"`. That is
  the boundary `shouldStartNewChatSession` already detects (CLI agent type,
  incompatible provider, or model family change), and it is the only case
  where the session the runner resumes cannot carry the conversation.
  `reused` / `adopted` / `initialized` sessions no longer get the block.
- Rounds whose run was cancelled, failed, or timed out are absent from the
  session either way, so `# Incomplete Rounds Context` keeps covering them
  unchanged.
- When the block is emitted, each prior run contributes its user text and its
  **final** assistant message. The last visible message per run is selected in
  SQL, so the intermediate narration is not fetched either.
- New always-present `# This Chat Thread` block carries `CHAT_THREAD_ID` and
  the `zero chat messages` / `zero logs` commands, so a run can pull the rest
  of the conversation on demand instead of receiving it up front. This mirrors
  what `buildAgentRunSourceContext` already does for cross-thread prompts.
- Removes the pi-specific `historyPolicy`. The trigger is now framework
  independent (cc / cdx / pi behave identically), and the old check compared
  `threadSessionRoute.modelProvider` (`vm0` for built-in models) against the
  Pi-capable provider table, so it never fired for built-in models even though
  those runs do take the Pi standby path.

Measured against that run's real data: the replay drops from 150,710 to
16,527 bytes when it is emitted, and to 0 when the session is reused. The
prompt would have been ~66 KB, comfortably under the 131,071-byte ceiling.

## Verification

- `prettier --check` on all touched files
- `tsc` on each changed service file
- Behavior is covered by `chat-events.bdd.test.ts`: a new case asserts that a
  model-family rotation replays only the final answer and skips the narration,
  and the same-family / follow-up cases now assert the block is absent. The
  full API suite runs in CI.

## Fallbacks

- **Removed** — `zero-runs-create.service.ts` `piNoLegacyHistory` /
  `zero-web-chat-session-prompt.service.ts` `historyPolicy`. This was a
  `FeatureSwitchKey.PiLoop`-gated branch that suppressed the legacy history
  replay for Pi runs. Evidence for removal: the trigger is now framework
  independent, so the Pi-specific escape hatch has no remaining job, and the
  predicate it used compared `threadSessionRoute.modelProvider` (`vm0` for
  built-in models) against the Pi-capable provider table, which excludes `vm0`,
  so it never fired for built-in models in the first place. The branch, its only
  caller, and its now-unused imports are deleted together; no test referenced
  `historyPolicy`.
- **Kept, pre-existing, untouched** — `buildWebChatPriorRunsContext` still
  substitutes `User: <run.prompt>` when a replayed run has no materialized user
  event, and `Assistant: [no stored assistant message]` when it has no assistant
  event. Chat materialization runs in `waitUntil`, so a completed run's events
  can legitimately lag its row; these are presentational defaults for a
  reachable state, not cross-version compatibility. Justified.
- No new cross-version rollout fallback is introduced, so no surface window or
  removal condition applies. `appendSystemPrompt` keeps its contract shape and
  only gets smaller, so neither an old runner reading a new value nor an old API
  writing an old value is affected.

## Follow-up (not in this PR)

The 128 KiB ceiling itself remains. `RunPayload.append_system_prompt` already
travels to guest-agent through `VM0_RUN_PAYLOAD_FILE` for exactly this reason,
but guest-agent still hands the merged Pi prompt to the standby child as an env
value. Splitting Pi's own system prompt (computed in `pi-agent-runtime` from an
`ExecutionEnv`) from the framework-shared append half, and reading the append
half from the payload file, removes the cliff.

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts
@@ -3752,7 +3752,8 @@ describe("CHAT-02: run-level model overrides", () => {
     ).toMatch(/[0-9a-f-]{36}/);
 
     // A run-level override of another model in the same family resumes the CLI
-    // session while carrying the prior web round as context.
+    // session, which already carries the prior web round, so the prompt does
+    // not replay it.
     const second = await sendChatRun(actor, {
       agentId,
       threadId: first.threadId,
@@ -3777,11 +3778,11 @@ describe("CHAT-02: run-level model overrides", () => {
     );
     const secondRun = await api.readRun(actor, second.runId);
     const appended = secondRun.appendSystemPrompt ?? "";
-    expect(appended).toContain("# Web Chat Run Context");
-    expect(appended).toContain(`- RUN_ID: ${first.runId}`);
-    expect(appended).toContain(`- LOG_COMMAND: zero logs ${first.runId} --all`);
-    expect(appended).toContain(`User: ${firstPrompt}`);
-    expect(appended).toContain("Assistant: opus answer");
+    expect(appended).not.toContain("# Web Chat Run Context");
+    expect(appended).not.toContain("Assistant: opus answer");
+    expect(appended).toContain("# This Chat Thread");
+    expect(appended).toContain(`- CHAT_THREAD_ID: ${first.threadId}`);
+    expect(appended).toContain("`zero chat messages`");
     const secondClaim = await claimChatRun(runnerGroup, second.runId);
     expect(secondClaim.claim.resumeSession?.sessionId).toBe(
       `bdd-cli-${first.runId}`,
@@ -4494,6 +4495,83 @@ describe("CHAT-02: run-level model overrides", () => {
     const retriedClaim = await claimChatRun(runnerGroup, retried.runId);
     expect(retriedClaim.claim.resumeSession).toBeNull();
     await cancelChatRun(actor, retried.runId);
+  }, 90_000);
+
+  it("replays only each prior run's final answer when a model family rotates the session", async () => {
+    const { actor, agentId, runnerGroup, providerId } =
+      await entitledChatActor();
+    chatCallbacks.failIfChatCallbackRouteIsFetched();
+    const { providerId: codexProviderId } = await upsertOrgModelProvider(
+      actor,
+      {
+        type: "openai-api-key",
+        secret: "prior-round-trim-openai-key",
+      },
+    );
+    await api.updateOrgModelPolicies(actor, [
+      {
+        model: "claude-sonnet-4-6",
+        isDefault: true,
+        defaultProviderType: "anthropic-api-key",
+        credentialScope: "org",
+        modelProviderId: providerId,
+      },
+      {
+        model: "gpt-5.6-terra",
+        isDefault: false,
+        defaultProviderType: "openai-api-key",
+        credentialScope: "org",
+        modelProviderId: codexProviderId,
+      },
+    ]);
+
+    const firstPrompt = "plan the migration in several steps";
+    const first = await sendChatRun(actor, {
+      agentId,
+      prompt: firstPrompt,
+      model: "claude-sonnet-4-6",
+    });
+    const firstClaim = await claimChatRun(runnerGroup, first.runId);
+    chatCallbacks.mockChatOutputEvents([
+      assistantEvent(0, "narration: reading the first file"),
+      assistantEvent(1, "narration: reading the second file"),
+      assistantEvent(2, "final answer with the migration plan"),
+    ]);
+    await completeChatRunOk(first.runId, firstClaim.sandboxHeaders, {
+      lastEventSequence: 2,
+    });
+    await flushWaitUntilForTest();
+    await waitForThreadMessages(actor, first.threadId, (items) => {
+      return eventBackedContents(items, first.runId).some((message) => {
+        return message.content === "final answer with the migration plan";
+      });
+    });
+
+    // Switching model family rotates the CLI session, so the prior round is
+    // replayed. An agentic run emits one chat message per step; only its final
+    // answer carries information the next run needs.
+    await chat.updateThreadModelSelection(
+      actor,
+      first.threadId,
+      "gpt-5.6-terra",
+    );
+    const second = await sendChatRun(actor, {
+      agentId,
+      threadId: first.threadId,
+      prompt: "continue after the family switch",
+    });
+    const appended = (await api.readRun(actor, second.runId))
+      .appendSystemPrompt;
+    expect(appended).toContain("# Web Chat Run Context");
+    expect(appended).toContain(`- RUN_ID: ${first.runId}`);
+    expect(appended).toContain(`User: ${firstPrompt}`);
+    expect(appended).toContain(
+      "Assistant: final answer with the migration plan",
+    );
+    expect(appended).not.toContain("narration: reading the first file");
+    expect(appended).not.toContain("narration: reading the second file");
+    expect(appended).toContain(`- CHAT_THREAD_ID: ${first.threadId}`);
+    await cancelChatRun(actor, second.runId);
   }, 90_000);
 
   it("retries preparation when the canonical conversation snapshot changes", async () => {
@@ -5257,7 +5335,7 @@ describe("CHAT-02: initial thinking indicator", () => {
 });
 
 describe("CHAT-02: prior rounds and thread titles", () => {
-  it("carries prior completed rounds, generates the thread title, and accepts immutable follow-up revokes", async () => {
+  it("leaves prior completed rounds to the session, generates the thread title, and accepts immutable follow-up revokes", async () => {
     const { actor, agentId, runnerGroup } = await entitledChatActor();
     chatCallbacks.failIfChatCallbackRouteIsFetched();
     mockOptionalEnv("OPENROUTER_API_KEY", "title-key");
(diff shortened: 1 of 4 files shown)


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 5e158188be3f8c6fe27a4e07e069b7b36837cca6
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Web chat stops replaying prior rounds into `appendSystemPrompt` unless the thread's CLI session rotated, and when it does replay, each prior run contributes its user text plus only its final assistant message. A new always-present `# This Chat Thread` block carries the thread id and the Zero CLI commands that read the rest of the conversation, and the Pi-specific `historyPolicy` escape hatch is deleted so the trigger is framework independent.

Disclosure: at the start of this review the PR summary had no `Fallbacks` section, which `docs/fallback.md` requires whenever a diff introduces, keeps, or removes a fallback (here: the removed `piNoLegacyHistory` branch). As the PR author I added that section to the summary before posting, so the criterion below is assessed against the corrected summary at the same HEAD. No code changed.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic, no wrong assumptions | Pass | `lastRunMessageSeqIds` groups by `run_id` over `CHAT_EVENT_CONTENT_TEXT_TYPES` with `content IS NOT NULL`; `run.completed` markers are written with `content: null` (`internal-chat-run-callback.service.ts:1685`), so the selected row is the run's last visible assistant text, not a lifecycle marker. `seq_id` is thread-unique, so `inArray(seqId, subquery)` cannot cross runs. |
| `REVIEW.md` | Security: no injection, secrets, or missing auth | Pass | No new endpoint, no auth path, no user input interpolated into SQL; the new prompt block interpolates only the thread UUID. |
| `REVIEW.md` | Type safety, no `any` or broad casts | Pass | No `any`, no `as` casts; `lastRunMessageSeqIds` leaves the Drizzle subquery type inferred, matching `afterSuccessfulRunBoundary` in `zero-chat-incomplete-context.service.ts`. |
| `REVIEW.md` | Maintainability, no duplicated logic, WHY comments | Pass | The narrowing reuses `chatEventTextCondition()` and `chatEventTypeIn()` instead of restating the text-event definition; the three added comments state why (session already carries history, narration carries no information) rather than what. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, contract, Zod, or `jsonb().$type<>()` change; `turbo/packages/db` and `api-contracts` are untouched. |
| `REVIEW.md` | Feature regression guard: changed behavior needs coverage on the affected path | Pass | The reused-session path that loses the replay is asserted at `chat-events.bdd.test.ts:3778` and `:5437`; the rotated path keeps its existing coverage at `:4479` and `:7638` and gains the new trimming case at `:4500`. |
| `REVIEW.md` | Feature switch containment for new user-facing behavior | N/A | This is a `fix:` that narrows an existing prompt path and deletes a switch-gated branch; it adds no new feature surface. Gating it would leave the 128 KiB failure live for the default path. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | No `any`, no `eslint-disable`, no `@ts-ignore` in the diff. |
| `docs/bad-smell.md` | Fail fast, no defensive try/catch | Pass | No try/catch added; the new query path lets DB errors propagate. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | Only static imports added (`max`, `isNotNull`, `chatEventTypeIn`, two event-type constants). |
| `docs/bad-smell.md` | Interface changes documented | Pass | `buildWebChatAppendSystemPrompt` gains a required `threadId`; both call sites are updated (`zero-runs-create.service.ts`, `zero-chat-events.command.ts:2852`) and the removed `historyPolicy` option has no remaining caller. |
| `docs/fallback.md` §1 | No negative tests asserting removed behavior stays removed | Pass | The `not.toContain` assertions pin the new product behavior (no duplicate replay on a resumed session, no narration in the replay) and each sits beside a positive assertion for the replacement block; none of them assert that deleted code is still deleted. |
| `docs/fallback.md` §2, §5 | Delete a switch-gated branch together with its callers and tests | Pass | `piNoLegacyHistory`, the `historyPolicy` parameter, and the now-unused `isPiEdgeCompatibleProviderType` / `isFeatureEnabled` / `FeatureSwitchKey` imports are removed in one change; no test referenced `historyPolicy`. |
| `docs/fallback.md` §6-§8 | A rollout fallback names surface, window, removal condition | N/A | No cross-version fallback is introduced, so no window applies. See Fallbacks below. |
| `docs/fallback.md` | Every fallback in the diff is declared in the PR summary | Pass | Declared in the summary's `Fallbacks` section (added during this review, see Summary disclosure). |
| `docs/deployment-compatibility.md` | Cross-version boundaries stay compatible in both directions | Pass | `appendSystemPrompt` keeps its contract type and is consumed as opaque text; only its content shrinks. An old runner reading a shorter value and an old API still writing the long value are both safe, and no migration, queue payload, or persisted shape changes. |
| `docs/testing.md` | Integration tests at entry points, no unit tests for internals | Pass | The new case drives `sendChatRun` / `claimChatRun` / `completeChatRunOk` and reads the result through `api.readRun`, not by calling `resolveWebChatSessionPrompt`. |
| `docs/testing.md` | Mock only at the boundary, real DB, no fetch stubbing, no fake timers | Pass | The new test adds no mocks beyond the existing `chatCallbacks` MSW-backed helpers; no `vi.mock`, no `vi.stubGlobal`, no `vi.useFakeTimers`, no `globalThis.services.db` mocking. |
| `docs/testing.md` | `fix:` commits carry a regression test | Pass | `chat-events.bdd.test.ts:4500` reproduces the exact regression: a run that emits three assistant messages contributes only its final answer to the next prompt. |
| `docs/testing/api-testing.md` | Set up and verify through API calls, not DB writes or service imports | Pass | State is created via chat send/claim/complete APIs and read via `api.readRun`; no schema imports or handler imports were added. |
| `docs/testing/anti-patterns.md` | AP-1 / AP-7 / AP-8: assert behavior, not mocks or implementation | Pass | Assertions are on the returned prompt text; no `expect(mock).toHaveBeenCalled()`, no internal-state inspection. |
| `docs/testing/patterns.md` | Tests live beside the route surface they exercise and reuse existing helpers | Pass | The case is added to the existing `chat-events.bdd.test.ts` route suite and reuses `upsertOrgModelProvider`, `assistantEvent`, `eventBackedContents`, `waitForThreadMessages`. |
| `docs/testing/testing-external-behavior.md` | Exercise the external interface, not internals | Pass | The trigger is provoked externally by `chat.updateThreadModelSelection` crossing a model family, the same mechanism a user has. |
| `.claude/skills/ccstate/SKILL.md` | ccstate command/computed and signal-lifecycle rules for `apps/api` | N/A | The diff only touches plain async service functions; no `computed`, `command`, Store, signal, or `detach()` usage is added or changed. |

### Findings

#### Critical (P0)
- None.

#### High Priority (P1)
- None outstanding. The one P1-class issue found (`docs/fallback.md`: the summary did not declare the removed `piNoLegacyHistory` fallback) was remediated in the PR summary before this review was posted, as disclosed above.

#### Notes (non-blocking)
- `resolveWebChatSessionPrompt` now trusts that a non-rotated session really carries the conversation. That holds because session restore is contractual rather than best-effort: `restore_claude_session` propagates a `RunnerResult` error (`crates/runner/src/executor/session_restore.rs:170`) and a zero-turn Claude result with `Missing`/`Empty` history fails the run instead of continuing silently (`crates/guest-agent/src/main.rs:523`). Worth keeping in mind if session restore ever becomes tolerant of a missing history blob.
- `sessionAction === "adopted"` takes the same no-replay path as `reused` but has no dedicated case in this PR; the closest coverage is the reused path.

### Fallbacks

- `turbo/apps/api/src/signals/services/zero-runs-create.service.ts` `piNoLegacyHistory` and `zero-web-chat-session-prompt.service.ts` `historyPolicy` — **removed**. Protected: `FeatureSwitchKey.PiLoop` runs skipping the legacy history replay. Surface: none, it was a feature-switch-gated internal branch. Removal condition met: the trigger is now framework independent, and the predicate compared `threadSessionRoute.modelProvider` (`vm0` for built-in models) against the Pi-capable provider table, which excludes `vm0`, so it never fired for built-in models. Branch, caller, and imports deleted together; no test referenced it. Declared in PR summary: Yes. Verdict: **Justified**.
- `turbo/apps/api/src/signals/services/zero-web-chat-session-prompt.service.ts` `buildWebChatPriorRunsContext` `User: <run.prompt>` / `Assistant: [no stored assistant message]` — **kept, pre-existing, untouched by this diff**. Protects a replayed run whose chat events have not materialized yet, which is reachable because materialization runs in `waitUntil`. Surface: none, same-version race rather than version skew. Presentational default under `docs/fallback.md` §6.4. Declared in PR summary: Yes. Verdict: **Justified**.
- No new cross-version rollout fallback is introduced, so no rollout window or removal condition applies.

### Testing

- Coverage: Adequate. New regression case at `chat-events.bdd.test.ts:4500` pins both halves of the fix (replay only on rotation, and only the final assistant message per run). The two behavior-flipped cases at `:3778` and `:5437` cover the reused-session path that changed, and the incomplete-rounds cases at `:1499`, `:5237`, `:6295`, `:6378`, `:7725` still assert that abnormal rounds keep their own block, unchanged by this PR.
- Conventions: Compliant. Integration-level, no new mocks, no fake timers, no internal imports, setup and verification through the chat APIs.
- Gaps: no dedicated case for `sessionAction === "adopted"`; the per-message 4,000-char cap remains untested, as it was before this PR.

PR #23656, opened 2026-07-29
refactor(chat): drop legacy role column

lancy:
## Summary

- add migration `0722_drop_chat_messages_role.sql` with the single physical contraction `ALTER TABLE chat_messages DROP COLUMN role`
- remove the legacy `role` field from the Drizzle `chatMessages` schema
- update latest-schema fixtures and add a populated-upgrade migration-consistency scenario that proves historical rows survive, canonical inserts still work, and append-only UPDATE protection remains active

## User impact

There is no visible chat behavior change. Canonical ChatEvents already use `eventType`, and transcript, prompt, export, and frontend role projections continue deriving display/model role from `eventType`.

`chatEventCompatibilityRole` is intentionally retained because it powers those projections and is independent of the removed physical column.

## Roadmap and rollout safety

This is **PR C**, the final close-out PR in the three-step roadmap from #22757.

The entry gate is satisfied:

- PR B (#23580) merged at 2026-07-29 02:06:11 UTC
- `gh release list` confirms PR B shipped in `api-v1.339.0` at 2026-07-29 02:21:14 UTC
- Zero confirmed at 2026-07-29 04:40 UTC that production had no role-related errors and lagging pre-B writer instances had drained

The column is already read-never and write-never. The migration is therefore safe in either deployment order:

- migration first: the currently deployed API does not read or write the physical column
- code first: the schema cleanup also does not read or write it, and the still-present column is harmless until migration

Queue work (#23651), physical table renaming, and projection behavior are intentionally out of scope.

## Verification

Passed locally:

- Prettier 3.8.1 on all changed TypeScript/schema metadata files
- `DATABASE_URL=postgresql://postgres@localhost:55452/postgres pnpm -F @vm0/db test:migration-consistency`
- `pnpm turbo run lint --filter=@vm0/db --filter=api --log-order=grouped --concurrency=1`
- `NODE_OPTIONS=--max-old-space-size=4096 pnpm turbo run check-types --filter=@vm0/db --filter=api --log-order=grouped --concurrency=1`
- `DATABASE_URL=postgresql://postgres@localhost:55452/postgres pnpm -F @vm0/db test` (9 files, 22 tests)
- `pnpm knip` in the repository's `ghcr.io/vm0-ai/vm0-toolchain:20260622` CI image (exit 0; configuration hints only)
- `git diff --check`

Full monorepo tests are delegated to the PR pipeline per repository guidance.

Closes #22757

PR #25836, opened 2026-08-08
perf(chat-search): answer indexed search from the projection alone

e7h4n:
## Problem

The index-backed chat search path (`FeatureSwitchKey.ChatSearchIndex`) matched with a semi-join:

```sql
chat_events.id IN (SELECT event_id FROM chat_event_search_docs WHERE user_id = $1 AND org_id = $2 AND to_tsvector(...) @@ to_tsquery(...))
```

That lets the planner drive from `chat_events` and treat the projection as the probed side. Measured on production (`request_time_ms` from `vm0-request-log-prod`, one request per query, in order):

| query | tsquery shape | server time | hits |
|---|---|---|---|
| `最终幻想战略版` | 6-bigram `<->` chain | 21ms | 6 |
| `最终幻想 汉化` | chain `&` token | 21ms | 5 |
| `幻想战略` | 3-bigram chain | 41ms | 7 |
| `最终幻想` | 3-bigram chain | 150ms | 7 |
| `幻想` | single bigram token | 1573ms | 7 |
| `最终` | single bigram token | 1477ms | 20 |
| `ffta` | single word token | 1815ms | 10 |
| `final fantasy` | `final & fantasy` | 1524ms | 3 |
| `github` (limit 10 / 50) | single word token | 1442 / 1647ms | 10 / 50 |
| `fftauscn` | single word token | 1583ms | 1 |
| `zzzzqqqqxx` | single word token | 1549ms | **0** |
| `github fftauscn` | two tokens `&` | 1739ms | **0** |

A tsquery containing a `<->` phrase chain runs in 21-150ms. Anything that is only plain tokens costs a flat 1.4-1.8s, independent of hit count or term frequency: a zero-hit query costs the same as a 50-hit one. That constant matches a full pass over `chat_event_search_docs` (830,803 rows). For comparison, the ILIKE path this replaces is a steady 195-350ms, so the index path was **5-7x slower** than its own fallback for the most common query shape (`userId` + `orgId` + one term).

## Changes

**1. The projection answers the whole match.** Keyword, ownership, agent scope, `since`, ordering and the limit are all resolved inside `chat_event_search_docs`. `chat_events` is then read only to decorate at most `limit + 1` rows by primary key, so the joins never participate in selecting rows. Ownership is re-asserted on that second step against the authoritative `chat_threads` / `agent_composes` rows, since the projection only holds a copy of it.

**2. `tsv` is materialized.** The GIN index was on the expression `to_tsvector('simple', text_bigram)`, so the parsed vector lived only in the index. Every index recheck, and every plan that skipped the index entirely, re-parsed the bigram-expanded text once per scanned row. It is now a stored generated column with the GIN index on it.

**3. Dead predicates removed from the index path.** `excludeGoalMarkerCondition()`, `chatEventTextCondition()` and `visibleChatEventCondition()` cannot change an index-path result. Verified in code and against production:

- `searchDocRole` is a whitelist of `input.prompt` / `input.rejected` / `output.message`. Sampling 300 recent `goal.open` / `goal.close` / `input.goal` events and 200 `output.error` / `run.failed` / `run.cancelled` events: **0** are present in `chat_event_search_docs`.
- The projector requires non-empty trimmed text, which is stricter than `chatEventTextCondition()`.
- The projector consumes revocation at index time (it skips revoked events and deletes their docs), so the query-side `NOT EXISTS` only covers the window between a revoke landing and the next projector tick (~30s at the observed catch-up rate). This PR also fixes that harvest: it previously matched only `control.revoke`, but `replaceLoadedChatEvent` appends the replacement with `revokes_event_id` set while keeping the replacement's own type, so a queued prompt is normally revoked by an `input.prompt` or `input.rejected`. The index therefore kept a doc for every replaced prompt, which the old query masked with `NOT EXISTS`.
- The remaining user-input clause cannot fire for any projected type: `input.rejected` is always written with a non-null `error`, which satisfies the clause. Of 4000 recent `input.prompt` / `input.rejected` events, 5 carry a revocation or interruption with a null `run_id`, and **0** of those are actually hidden.

The legacy ILIKE path keeps every one of these conditions unchanged.

**4. `agent_compose_id` is projected** (nullable, see Migration) so the existing `agentId` query parameter keeps filtering correctly without joining `zero_agents`, which was an unconditional `innerJoin` referenced only by that optional filter.

## Migration

`chat_event_search_docs` is derived data that the projector rebuilds from `chat_events`, so `0867` truncates it and the watermark table rather than backfilling. Truncating first also keeps the `STORED` generated column from rewriting ~830k rows under `ACCESS EXCLUSIVE` inside a pre-promotion migration.

Index-path recall is degraded until the replay completes. At the default batch size (500 threads per minute) a full replay of ~131k threads takes roughly 4.5 hours; `CHAT_EVENT_SEARCH_PROJECTION_BATCH_SIZE` shortens the window. The switch is opt-in and not broadly enabled, and the ILIKE path is unaffected.

Cross-version behavior, per `docs/deployment-compatibility.md`:

- **Old code after migration.** `agent_compose_id` is added **nullable**. Migrations run before API traffic promotion, and the previously deployed projector inserts an explicit column list without this column, so a `NOT NULL` column would fail its every-minute cron with `23502` for the whole skew window (observed up to ~102 minutes). Rows the old projector writes carry a null `agent_compose_id`; they still match unfiltered search and do not match an agent-scoped search until the projector rewrites them, which its `onConflictDoUpdate` on that column does on the next tick reaching the same event. Tightening to `NOT NULL` is a follow-up release once no API without the column is serving. `tsv` is `GENERATED ALWAYS`, so the old writer never names it.
- **New code before migration.** The new projector and the index read path reference `agent_compose_id` and `tsv` unconditionally, so they depend on the ordering guarantee that required migrations complete before API traffic promotion. If that ordering drifts, the projector cron and switch-on searches fail with `42703` until the migration lands; the ILIKE path and every other route are unaffected, and the projector recovers on its next tick with no data loss because the watermark only advances inside the committed transaction.

## Fallbacks

- `turbo/apps/api/src/signals/services/zero-chat-thread.service.ts` → `chatSearchLegacyMatches`, selected by `useSearchIndex` from `FeatureSwitchKey.ChatSearchIndex`: **kept**. The pre-projection ILIKE scan, with its `chatEventTextCondition` / `visibleChatEventCondition` / `excludeGoalMarkerCondition`, still serves users who do not have the switch on. Surface: none in the cross-version sense — this is a non-GA feature-switch dual path, not a deploy-skew branch, so no rollout window applies. Removal condition: delete together with the switch when `ChatSearchIndex` reaches GA.
- `turbo/packages/db/src/schema/chat-event-search.ts` → `agentComposeId` nullability: **introduced**. Surface: DB/API skew, up to the observed ~102 minutes. A row the previous API's projector writes during that window carries a null `agent_compose_id` and does not match an agent-scoped search until the projector rewrites it; `cron-project-chat-event-search` upserts that one column on the next tick reaching the same event, so the gap closes without a separate backfill. Removal condition: tighten to `NOT NULL` in a follow-up release once no API build lacking the column is serving.
- `turbo/packages/db/src/schema/chat-event-search.ts` → `text` column: **kept**, written but read by neither path. Not a cross-version fallback; listed under Follow-ups for deletion.

## Tests

Adds an index-path case to `CHAT-01 chat search index` covering the filters the new two-step query now owns: `since` against the projection's own `created_at`, `agentId` against the projected `agent_compose_id`, and the `limit` / `hasMore` boundary.

## Follow-ups (not in this PR)

- `btree_gin` composite `(user_id, org_id, tsv)` so the tenant filter and the term match resolve in one index scan.
- Search has no relevance ranking at all (`ORDER BY created_at DESC`); an over-limit result set returns the newest N, not the most relevant N.
- `chat_event_search_docs.text` is written but never read.


e7h4n:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 3b6277cf9c9b522942eb9745fffd85684533fe75

### Summary

Rewrites the feature-switched chat search index path so `chat_event_search_docs` answers the whole match (keyword, ownership, agent scope, `since`, ordering, limit) and `chat_events` is only read afterwards to decorate at most `limit + 1` rows by primary key, materializes the tsvector as a stored generated column with the GIN index on it, projects `agent_compose_id`, and truncates the projection so the cron replays it. The performance analysis is well evidenced and the removed predicates are each justified against production data. The blocker is the migration's cross-version behavior, not the query design.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic, no wrong assumptions | Pass | `limit + 1` is threaded correctly into `chatSearchIndexedEventIds` and re-applied as `.limit()`; `hasMore` still derives from the extra row. |
| `REVIEW.md` | Security / missing auth checks | Pass | Index path filters `user_id` + `org_id` in the projection and re-asserts ownership against `chat_threads.user_id` / `agent_composes.org_id` in the decorate step, so a stale projected copy cannot widen access. |
| `REVIEW.md` | Type safety, no `any` | Pass | `tsvectorColumn` is a typed `customType`; new `ChatSearchMatchRow` replaces an inferred shape. No `any`, no assertions added. |
| `REVIEW.md` | DB/JSONB review gate — persisted contract change needs migration or rationale | Finding (P0) | `turbo/packages/db/src/schema/chat-event-search.ts`, `0867_chat_search_docs_materialized_tsv.sql`: migration present, but not compatible with the previously deployed writer. See P0 below. |
| `REVIEW.md` | Feature regression guard — replaced behavior needs coverage | Pass | The legacy ILIKE path is preserved verbatim in `chatSearchLegacyMatches` and remains covered by `CHAT-01 chat search`; the index path gains filter/limit coverage. |
| `docs/bad-smell.md` | Zero `any`, no lint/type suppressions | Pass | No suppression comments or `any` in the diff. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | N/A | The diff adds no imports beyond static ones. |
| `docs/bad-smell.md` | No unnecessary defensive error handling | Pass | New helpers fail fast; the only early return is the empty-id short circuit, which avoids an `IN ()`. |
| `docs/fallback.md` | PR summary must carry a `Fallbacks` section (or `Fallbacks: none`) | Finding (P1) | PR body has Problem/Changes/Migration/Tests/Follow-ups but no `Fallbacks` section, while the diff keeps a switch-gated dual path. See P1 below. |
| `docs/fallback.md` | Feature behind a non-GA switch needs no compatibility code | Pass | No compatibility branch is added for the pre-GA index path's own earlier shape; the projection is truncated and replayed instead of dual-read. |
| `docs/fallback.md` | Removing a fallback requires stated evidence and deleting branch + tests together | Pass | Removal of `excludeGoalMarkerCondition` / `chatEventTextCondition` / `visibleChatEventCondition` from the index path is backed by the projector whitelist plus production sampling stated in the PR body; no orphaned tests remain for them on that path. |
| `docs/deployment-compatibility.md` | Old code after migration must keep issuing legal statements | Finding (P0) | `agent_compose_id uuid NOT NULL` with no default breaks the draining API's projector INSERT. |
| `docs/deployment-compatibility.md` | New code before migration must not require the new column | Finding (P0) | New projector and search path reference `agent_compose_id` / `tsv` unconditionally. |
| `docs/deployment-compatibility.md` | Compatibility code must be temporary and explicit | N/A | No cross-version compatibility branch is introduced by this PR. |
| `docs/event-sourcing.md` | Persistent events stay authoritative; projections rebuild from them | Pass | `chat_event_search_docs` remains derived; truncation relies on the documented replay-from-watermark-0 path and does not touch `chat_events`. |
| `docs/testing.md` | Integration tests at entry points, no unit tests for internal functions | Pass | The new case drives `GET /api/zero/chat/search` through the BDD client, not `chatSearchIndexedEventIds` directly. |
| `docs/testing.md` | Real database, no mocking of `globalThis.services.db` | Pass | The new test uses the real DB and the real projector cron route. |
| `docs/testing.md` | No fake timers | Pass | Uses `now()` as a `since` boundary; no `vi.useFakeTimers()`. |
| `docs/testing/api-testing.md` | Route tests live with route BDD suites and exercise the HTTP surface | Pass | Added inside `CHAT-01 chat search index` in `chat-threads.bdd.test.ts`. |
| `docs/testing/anti-patterns.md` | AP-1 / AP-4 / AP-7: assert behavior, never mock internals or implementation | Pass | Assertions are on returned thread ids, lengths, and `hasMore`. |
| `docs/testing/testing-external-behavior.md` | Assert observable external behavior | Pass | Only response payload fields are asserted. |
| `.claude/skills/ccstate/SKILL.md` | No `detach()` in `signals/`; work stays bound to the parent computed | Pass | New helpers are plain async functions awaited inside the existing `computed`; no new signal or detached lifecycle. |

### Findings

- `turbo/packages/db/src/migrations/0867_chat_search_docs_materialized_tsv.sql`: `ALTER TABLE "chat_event_search_docs" ADD COLUMN "agent_compose_id" uuid NOT NULL;` has no default, and migrations run **before** API traffic promotion. The previously deployed projector (`cron-project-chat-event-search.service.ts` on `main`) inserts an explicit column list of `event_id, chat_thread_id, org_id, user_id, role, created_at, text, text_bigram` — no `agent_compose_id`. Every tick of the `* * * * *` `project-chat-event-search` cron on the draining API will therefore fail with `23502 null value in column "agent_compose_id" violates not-null constraint` until promotion completes. `docs/deployment-compatibility.md` records DB/API skew reaching ~102 minutes, and lists migration `0722` (PR #23656) as the same class of failure: the old ORM schema still generates the statement even when application logic has moved on. Make the column nullable in this migration (the truncate means the new projector fills it for every row it writes, and a row written by the old projector simply does not match an `agentId` filter), or give it a default and tighten it in a follow-up release.
- Same migration, other direction: the new API selects `tsv` and `agent_compose_id` unconditionally, in both the projector INSERT and `chatSearchIndexedEventIds`. If the new API serves before the migration is visible — the documented `0697` / `0723` failure mode — the projector cron and every switch-on search request fail with `42703 column does not exist`. The search read is at least confined to opted-in users by `FeatureSwitchKey.ChatSearchIndex`, but the projector cron is not gated at all. Please state the ordering guarantee you are relying on in the PR body, or gate the new columns' use so the pre-migration window degrades instead of erroring.

- PR summary: `docs/fallback.md` requires a `Fallbacks` section listing every fallback or compatibility behavior the PR introduces, keeps, or removes, or an explicit `Fallbacks: none`. The body has no such section, while the diff explicitly keeps the pre-projection ILIKE path as `chatSearchLegacyMatches` selected by `useSearchIndex`. Add the section; the finding is the missing declaration, not the code.

### Fallbacks

- `turbo/apps/api/src/signals/services/zero-chat-thread.service.ts` → `chatSearchLegacyMatches` (selected by `useSearchIndex` from `FeatureSwitchKey.ChatSearchIndex`): keeps the pre-projection ILIKE scan, including `chatEventTextCondition`, `visibleChatEventCondition` and `excludeGoalMarkerCondition`, for users who do not have the switch on. Surface: none in the cross-version sense — this is a non-GA feature-switch dual path, not a deploy-skew branch, so no rollout window applies. Removal condition: delete together with the switch when `ChatSearchIndex` reaches GA; not evidenced in this PR. Declared in PR summary: **No — P1**. Verdict: **Justified** (the switch's off-state must keep working), but it must be declared.
- `turbo/packages/db/src/schema/chat-event-search.ts` → `text` column: retained and still written while nothing reads it, and its doc comment still claims it backs "fallback ILIKE matching", which is no longer true on either path. The PR lists it as a follow-up rather than a fallback. Surface: none. Removal condition: named in the PR's follow-up list. Verdict: **Justified to defer**, but the stale comment should not be left asserting a behavior that does not exist.

### Testing

- Coverage: **Adequate for the query change, insufficient for the migration.** The new `applies agent, since and limit filters inside the projection` case covers exactly the three responsibilities the projection newly owns (`since` on its own `created_at`, `agentId` on the projected `agent_compose_id`, and the `limit` / `hasMore` boundary), and the existing index suite still covers CJK bigram recall, whole-word matching, idempotent re-ticks, and assistant-output indexing. Nothing covers the old-writer-after-migration direction called out in P0; `docs/testing.md` has a Deployment Compatibility Tests section for exactly this case.
- Conventions: **Compliant.** Integration-level through the HTTP surface, real database, real projector cron route, no fake timers, no internal mocks, assertions on returned payload rather than mock calls.
- Not blocking, for awareness: `chatSearchIndexedEventIds` orders only by `created_at DESC` and the decorate step re-orders by the same key, so rows sharing a timestamp at the limit boundary can be selected by one ordering and displayed by another. The previous single-query shape had the same tie-break gap, so this is not a regression.

e7h4n:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 9c02dcdc3692b369a9dcf95de0108f42254d208b

### Summary

Second review, at `9c02dcd`. Both P0s from the previous review are resolved: `agent_compose_id` is now nullable with an explicit surface/window/removal comment, and the PR body documents the ordering the new-code-before-migration direction relies on. The added projector fix — harvesting revocations from any event carrying `revokes_event_id` rather than only `control.revoke` — is correct and matches `notExists(revoker)` exactly; it also explains the whole `test-api` failure. One new issue: the truncate plus `onConflictDoNothing` means rows the previous API's projector writes during the skew window keep a null `agent_compose_id` permanently, which contradicts the self-healing claim in both the schema comment and the PR's `Fallbacks` entry.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic | Finding (P1) | `cron-project-chat-event-search.service.ts`: `onConflictDoNothing` never repairs a row the old projector inserted with a null `agent_compose_id`. |
| `REVIEW.md` | Security / missing auth checks | Pass | Index path filters `user_id` + `org_id` in the projection and re-asserts ownership against `chat_threads.user_id` / `agent_composes.org_id` on the decorate step. |
| `REVIEW.md` | Type safety, no `any` | Pass | `tsvectorColumn` is a typed `customType`; nullable `agentComposeId` narrows to `string \| null` and is only compared with `eq`. |
| `REVIEW.md` | DB/JSONB review gate — persisted contract change needs migration or rationale | Pass | `0867` is additive and nullable; the previous NOT-NULL blocker is resolved. |
| `REVIEW.md` | Feature regression guard | Pass | The legacy ILIKE path is preserved verbatim and still covered by `CHAT-01 chat search`; the index suite covers the new filters. |
| `docs/bad-smell.md` | Zero `any`, no suppressions, no dynamic imports | Pass | None present in the diff. |
| `docs/bad-smell.md` | No unnecessary defensive handling | Pass | New helpers fail fast; the only early return avoids an empty `IN ()`. |
| `docs/fallback.md` | PR summary must carry a `Fallbacks` section | Pass | Added, listing the legacy ILIKE path, the `agentComposeId` nullability, and the unread `text` column. |
| `docs/fallback.md` | A rollout fallback must name surface, window, and removal condition | Finding (P1) | Surface and window are right; the stated behavior ("the projector replay refills them") does not hold under `onConflictDoNothing`. |
| `docs/fallback.md` | Feature behind a non-GA switch needs no compatibility code | Pass | No compatibility branch is added for the pre-GA index path's own earlier shape. |
| `docs/fallback.md` | Removing a fallback requires stated evidence, branch and tests deleted together | Pass | The removed index-path predicates are justified in the body; `9c02dcd` corrects the previously overstated revocation claim. |
| `docs/deployment-compatibility.md` | Old code after migration must keep issuing legal statements | Pass | `agent_compose_id` is nullable, so the draining API's explicit-column INSERT stays legal; `tsv` is `GENERATED ALWAYS` and never named by the old writer. |
| `docs/deployment-compatibility.md` | New code before migration must not require the new column | Pass (documented risk) | Reliance on migration-before-promotion is now stated in the body, with the `42703` failure mode and the no-data-loss argument; the doc accepts that ordering for this direction. |
| `docs/deployment-compatibility.md` | Compatibility code must be temporary and explicit | Finding (P1) | The comment is explicit but states an incorrect removal path. |
| `docs/event-sourcing.md` | Persistent events authoritative; projections rebuild from them | Pass | Revocation harvest now mirrors the reader-side visibility rule, so the projection agrees with the authoritative event log. |
| `docs/testing.md` | Integration tests at entry points | Pass | Coverage is through `GET /api/zero/chat/search` and the projector cron route. |
| `docs/testing.md` | Real database, no fake timers, no internal mocks | Pass | Real DB and real cron route; `now()` supplies the `since` boundary. |
| `docs/testing/api-testing.md` | Route tests live with route BDD suites | Pass | Added inside `CHAT-01 chat search index`. |
| `docs/testing/anti-patterns.md` | AP-1 / AP-4 / AP-7 | Pass | Assertions are on response payloads, not mocks or internals. |
| `docs/testing/testing-external-behavior.md` | Assert observable external behavior | Pass | Only response fields are asserted. |
| `.claude/skills/ccstate/SKILL.md` | No `detach()` in `signals/`; work bound to the parent computed | Pass | New helpers are plain async functions awaited inside the existing `computed`. |

### Findings

- None. Both previous P0s are resolved.

- `turbo/apps/api/src/signals/services/cron-project-chat-event-search.service.ts` + `turbo/packages/db/src/schema/chat-event-search.ts`: the null `agent_compose_id` is permanent, not transient. `0867` truncates `chat_event_search_watermarks`, so after the migration **both** the draining old API and the new API replay every thread from watermark 0. The old projector writes rows without `agent_compose_id`, and the new projector's insert is `onConflictDoNothing({ target: eventId })`, so it never repairs a row that already exists. Those docs keep a null `agent_compose_id` forever and silently never match an agent-scoped search, even long after the rollout window closes. The schema comment ("the projector replay refills them") and the PR's `Fallbacks` entry both assert the opposite. Either switch the insert to `onConflictDoUpdate` setting `agent_compose_id` — which keeps tick idempotency and heals the window's rows on the next pass — or correct both the comment and the `Fallbacks` entry to state that the follow-up release must backfill before tightening to `NOT NULL`. The first option is preferable: it removes the silent gap instead of documenting it.

### Fallbacks

- `turbo/apps/api/src/signals/services/zero-chat-thread.service.ts` → `chatSearchLegacyMatches`, gated by `FeatureSwitchKey.ChatSearchIndex`: **kept**. Surface: none in the cross-version sense; this is a non-GA feature-switch dual path, so no rollout window applies. Removal condition: delete with the switch at GA. Declared in PR summary: **Yes**. Verdict: **Justified**.
- `turbo/packages/db/src/schema/chat-event-search.ts` → `agentComposeId` nullability: **introduced**. Surface: DB/API skew, up to the observed ~102 minutes. Removal condition: tighten to `NOT NULL` in a follow-up release. Declared in PR summary: **Yes**, but with an inaccurate self-healing claim — see P1. Verdict: **Justified**, declaration needs correcting.
- `turbo/packages/db/src/schema/chat-event-search.ts` → `text` column: **kept**, written and read by nothing. Surface: none. Removal condition: listed under Follow-ups. Declared in PR summary: **Yes**. Verdict: **Justified to defer**; the column comment still claims it backs "fallback ILIKE matching", which is no longer true on either path.

### Testing

- Coverage: **Adequate.** The `fix:` commit's regression guard is real if implicit: every message sent through `sendNoCreditMessage` is a prompt replaced by an `input.rejected` carrying `revokes_event_id`, so a regression in the revocation harvest immediately doubles the expected counts in `serves index-backed keyword search from the projection behind the switch`, `excludes future follow-up content from indexed matches and context`, and `applies agent, since and limit filters inside the projection` — which is exactly how the bug surfaced. An explicit case naming the replaced-prompt behavior would document the invariant better, but the coverage exists.
- Conventions: **Compliant.** Integration-level through the HTTP surface, real database, real projector cron route, no fake timers, no internal mocks, assertions on returned payload.
- Gap tied to the P1: nothing covers a doc whose `agent_compose_id` is null, because no test can produce the old projector's insert shape. If the `onConflictDoUpdate` fix is taken, the healing path is likewise untestable at the entry point; a note in the code is the practical substitute.

e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 99bc95d9fa45d20598055235932e0b47cc154756
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Third review, at `99bc95d`. Every finding from the previous two rounds is resolved: `agent_compose_id` is nullable so the draining API's projector INSERT stays legal, the ordering the new-code-before-migration direction relies on is documented, the revocation harvest now matches `notExists(revoker)`, and the last P1 — a permanently null `agent_compose_id` on rows the old projector wrote — is fixed by upserting that single column under a `setWhere` guard. The schema comment and the PR's `Fallbacks` entry now describe the actual healing path.

I generated the projector's insert statement from the committed schema rather than reasoning about it:

```sql
insert into "chat_event_search_docs" ("event_id", "chat_thread_id", "org_id", "user_id", "agent_compose_id", "role", "created_at", "text", "text_bigram")
values ($1, ..., $9)
on conflict ("event_id") do update set "agent_compose_id" = EXCLUDED.agent_compose_id
where "chat_event_search_docs"."agent_compose_id" is null
returning "event_id"
```

That confirms three things the change depends on: `tsv` is absent from the column list, so the `GENERATED ALWAYS` column is never written; the `DO UPDATE ... WHERE` predicate is qualified with the target table, which is the legal form; and rows skipped by the guard are neither updated nor returned, so `indexedEvents` keeps the same meaning it had under `DO NOTHING`. Docs within one batch come from a single thread's `chat_events` ordered by `seq_id`, so no `event_id` repeats in one `VALUES` list and the "cannot affect row a second time" hazard of a multi-row `DO UPDATE` does not apply.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic | Pass | Upsert guard verified against generated SQL; `limit + 1` threading and `hasMore` unchanged; no duplicate conflict target within a batch. |
| `REVIEW.md` | Security / missing auth checks | Pass | Index path filters `user_id` + `org_id` in the projection and re-asserts ownership against `chat_threads.user_id` / `agent_composes.org_id` on the decorate step. |
| `REVIEW.md` | Type safety, no `any` | Pass | `tsvectorColumn` typed `customType`; nullable `agentComposeId` narrows to `string \| null` and is only used with `eq` / `isNull`. |
| `REVIEW.md` | DB/JSONB review gate | Pass | `0867` is additive and nullable; truncate is justified in the migration comment and the body. |
| `REVIEW.md` | Feature regression guard | Pass | Legacy ILIKE path preserved verbatim and still covered by `CHAT-01 chat search`; index suite covers the new filters. |
| `docs/bad-smell.md` | Zero `any`, no suppressions, no dynamic imports | Pass | None present. |
| `docs/bad-smell.md` | No unnecessary defensive handling | Pass | `setWhere` is a correctness guard against rewriting a filled row, not a defensive catch. |
| `docs/fallback.md` | PR summary must carry a `Fallbacks` section | Pass | Present, three entries. |
| `docs/fallback.md` | A rollout fallback must name surface, window, and removal condition | Pass | `agentComposeId` entry now states the real behavior — no agent-scoped match until the projector rewrites the row, healed by the upsert — plus the `NOT NULL` follow-up. |
| `docs/fallback.md` | Feature behind a non-GA switch needs no compatibility code | Pass | No compatibility branch for the pre-GA index path's own earlier shape. |
| `docs/fallback.md` | Removing a fallback requires stated evidence | Pass | Removed index-path predicates are justified in the body, with the revocation claim corrected in `9c02dcd`. |
| `docs/deployment-compatibility.md` | Old code after migration keeps issuing legal statements | Pass | Nullable column keeps the draining API's explicit-column INSERT legal; `tsv` never appears in a written column list. |
| `docs/deployment-compatibility.md` | New code before migration must not require the new column | Pass (documented risk) | Reliance on migration-before-promotion stated with the `42703` failure mode and the no-data-loss argument; the doc accepts ordering for this direction. |
| `docs/deployment-compatibility.md` | Compatibility code must be temporary and explicit | Pass | Schema comment names the surface, the ~102-minute window, the healing path, and the `NOT NULL` removal condition. |
| `docs/event-sourcing.md` | Persistent events authoritative; projections rebuild from them | Pass | Revocation harvest mirrors the reader-side visibility rule; the projection is still fully replayable. |
| `docs/testing.md` | Integration tests at entry points | Pass | Coverage through `GET /api/zero/chat/search` and the projector cron route. |
| `docs/testing.md` | Real database, no fake timers, no internal mocks | Pass | Real DB and real cron route; `now()` supplies the `since` boundary. |
| `docs/testing/api-testing.md` | Route tests live with route BDD suites | Pass | Added inside `CHAT-01 chat search index`. |
| `docs/testing/anti-patterns.md` | AP-1 / AP-4 / AP-7 | Pass | Assertions on response payloads only. |
| `docs/testing/testing-external-behavior.md` | Assert observable external behavior | Pass | Only response fields asserted. |
| `.claude/skills/ccstate/SKILL.md` | No `detach()` in `signals/`; work bound to the parent computed | Pass | New helpers are plain async functions awaited inside the existing `computed`. |

### Findings

- None.

- None. The previous round's P1 is fixed at `99bc95d`.

Non-blocking, for a follow-up: `turbo/packages/db/src/schema/chat-event-search.ts` still documents the `text` column as "also used for fallback ILIKE matching". Neither path reads it — the ILIKE path matches against `chat_events` — so the comment asserts a behavior that does not exist. The PR already lists the column itself under Follow-ups.

### Fallbacks

- `turbo/apps/api/src/signals/services/zero-chat-thread.service.ts` → `chatSearchLegacyMatches`, gated by `FeatureSwitchKey.ChatSearchIndex`: **kept**. Surface: none in the cross-version sense; a non-GA feature-switch dual path, so no rollout window applies. Removal condition: delete with the switch at GA. Declared in PR summary: **Yes**. Verdict: **Justified**.
- `turbo/packages/db/src/schema/chat-event-search.ts` → `agentComposeId` nullability: **introduced**. Surface: DB/API skew, up to the observed ~102 minutes. Removal condition: tighten to `NOT NULL` in a follow-up release once no API build lacking the column is serving; the window's rows are healed by the projector upsert, so that follow-up needs no separate backfill. Declared in PR summary: **Yes**, accurately as of this commit. Verdict: **Justified**.
- `turbo/packages/db/src/schema/chat-event-search.ts` → `text` column: **kept**, written and read by nothing. Surface: none. Removal condition: listed under Follow-ups. Declared in PR summary: **Yes**. Verdict: **Justified to defer**, with the stale comment noted above.

### Testing

- Coverage: **Adequate.** The revocation fix has a real, if implicit, regression guard: every message sent through `sendNoCreditMessage` is a prompt replaced by an event carrying `revokes_event_id`, so a regression immediately doubles the expected counts across three cases in `CHAT-01 chat search index` — which is exactly how it surfaced. The new filter case covers the three responsibilities the projection newly owns (`since`, `agentId`, `limit` / `hasMore`).
- Conventions: **Compliant.** Integration-level through the HTTP surface, real database, real projector cron route, no fake timers, no internal mocks.
- Untestable at the entry point, accepted: the null-`agent_compose_id` healing path cannot be reached from a product API, because no supported writer produces the previous version's insert shape. The generated-SQL check above is the substitute evidence.

e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 018c44ffb2d11791a035ef4019e789b026953a48
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Fourth review, at `018c44f`. This head adds no product change over the previously approved `99bc95d`: it merges `origin/main` and renumbers this PR's migration from `0867` to `0868` because main landed `0867_drop_chat_event_asset_refs` while the PR sat in the merge queue. Verified mechanically rather than by inspection — every file this PR owns is byte-identical to the reviewed head:

```
0 turbo/apps/api/src/signals/services/cron-project-chat-event-search.service.ts
0 turbo/apps/api/src/signals/services/zero-chat-thread.service.ts
0 turbo/apps/api/src/signals/routes/__tests__/chat-threads.bdd.test.ts
0 turbo/packages/db/src/schema/chat-event-search.ts
```

`git diff origin/main...HEAD` is exactly the seven intended files, and `git diff origin/main...HEAD -- .../0867_drop_chat_event_asset_refs.sql` is empty, so the conflict resolution kept main's migration untouched instead of clobbering it.

The renumbered migration is the same statement list, with the hand-written `TRUNCATE` header preserved — worth calling out because the regeneration step would otherwise have dropped it (the file is a generated diff plus hand-written SQL, and it carries no `-- Custom SQL migration file` marker for the fast path to refuse on). Git records the migration as a pure rename with zero content change. The regenerated `0868_snapshot.json` matches the schema: `agent_compose_id` `notNull: false`, `tsv` `generated: { type: "stored" }`, and the GIN index on the `tsv` column rather than an expression.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic | Pass | No source change since `99bc95d`; migration statements unchanged apart from the file number. |
| `REVIEW.md` | Security / missing auth checks | Pass | Unchanged from the approved head: projection filters `user_id` + `org_id`, decorate step re-asserts against `chat_threads` / `agent_composes`. |
| `REVIEW.md` | Type safety, no `any` | Pass | No source change. |
| `REVIEW.md` | DB/JSONB review gate | Pass | Journal tail is `0866` → `0867_drop_chat_event_asset_refs` → `0868_amused_brood`, contiguous and consistent with main; snapshot `prevId` chains correctly. |
| `REVIEW.md` | Feature regression guard | Pass | Legacy ILIKE path and its coverage unchanged. |
| `docs/bad-smell.md` | Zero `any`, no suppressions, no dynamic imports | Pass | No source change. |
| `docs/fallback.md` | `Fallbacks` section present and accurate | Pass | Unchanged and still accurate; the three entries survive the rebase. |
| `docs/fallback.md` | Rollout fallback names surface, window, removal condition | Pass | `agentComposeId` entry unchanged. |
| `docs/deployment-compatibility.md` | Old code after migration keeps issuing legal statements | Pass | Nullable column and `GENERATED ALWAYS` `tsv` unchanged by the renumbering. |
| `docs/deployment-compatibility.md` | New code before migration | Pass (documented risk) | Unchanged; ordering reliance still stated in the body. |
| `docs/deployment-compatibility.md` | Migration ordering must be consistent with main | Pass | This commit exists specifically to restore that: the PR migration now applies after main's `0867`. |
| `docs/event-sourcing.md` | Projection stays derived and replayable | Pass | `TRUNCATE` + replay semantics preserved through the rename. |
| `docs/testing.md` / `api-testing.md` / `anti-patterns.md` / `testing-external-behavior.md` | Integration-level, real DB, no fake timers, no internal mocks, behavior assertions | Pass | Test file byte-identical to the approved head. |
| `.claude/skills/ccstate/SKILL.md` | No `detach()` in `signals/`; work bound to the parent computed | Pass | No source change. |

### Findings

- None.

- None.

Non-blocking, carried over: `turbo/packages/db/src/schema/chat-event-search.ts` still documents the `text` column as "also used for fallback ILIKE matching", which no path does; the column is already listed under Follow-ups. Also cosmetic: the renumbered migration carries drizzle's generated name `0868_amused_brood` instead of the previous descriptive `chat_search_docs_materialized_tsv`, because renaming it would require hand-editing `_journal.json`, which the database-development practice forbids.

### Fallbacks

- `turbo/apps/api/src/signals/services/zero-chat-thread.service.ts` → `chatSearchLegacyMatches`, gated by `FeatureSwitchKey.ChatSearchIndex`: **kept**. Surface: none in the cross-version sense; non-GA feature-switch dual path. Removal condition: delete with the switch at GA. Declared: **Yes**. Verdict: **Justified**.
- `turbo/packages/db/src/schema/chat-event-search.ts` → `agentComposeId` nullability: **introduced**. Surface: DB/API skew, up to the observed ~102 minutes. Removal condition: tighten to `NOT NULL` in a follow-up once no API build lacking the column is serving; the window's rows are healed by the projector upsert, so no separate backfill is needed. Declared: **Yes**. Verdict: **Justified**.
- `turbo/packages/db/src/schema/chat-event-search.ts` → `text` column: **kept**, written and read by nothing. Surface: none. Removal condition: listed under Follow-ups. Declared: **Yes**. Verdict: **Justified to defer**.

### Testing

- Coverage: **Adequate**, unchanged from `99bc95d`. `test-api` passed on all eight shards at that head, including the three cases whose counts caught the revocation bug, and `test-migrate` passed, which exercises this migration on a fresh database.
- Conventions: **Compliant**, test file unchanged.
- This head re-runs the full pipeline against main's new tip, which is the point of the rebase; `test-migrate` is the check that matters most here since the migration moved position in the journal.

PR #25888, opened 2026-08-09
fix(runner): recheck active input after dropped wakeup

### LINKED ISSUE #25868: refactor(rust): make active-input wakeups resilient to dropped notifications
## Location

- `crates/ably-subscriber/src/subscribe.rs:18-30`
- `crates/ably-subscriber/src/types.rs:251-287`
- `crates/runner/src/provider/api_ably_supervisor.rs:548-568`
- `crates/runner/src/active_input.rs:125-135`
- `crates/runner/src/active_input.rs:157-196`
- `crates/runner/src/executor/active_input.rs:54-110`
- `turbo/apps/api/src/signals/services/chat-thread-queue-drain.service.ts:60-106`
- `turbo/apps/api/src/signals/services/chat-thread-queue-drain.service.ts:206-295`

## Problem

The API active-input path treats an Ably message as the authoritative wakeup even though the subscription layer explicitly documents that `Event::Message` delivery is non-blocking and may be dropped when its shared bounded channel is full.

The durable state is the API list/claim endpoint, but the runner only reads that state after `ActiveInputSubscription::wait()` returns. `ActiveInputSource::wait(timeout)` honors the timeout for `LocalQueue` while the API variant ignores it and waits indefinitely for a matching notification. This gives one method two materially different progress contracts and promotes a lossy advisory signal into the sole trigger for checking durable state.

A concrete failure sequence is:

1. The forwarder lists active inputs and receives an empty result.
2. A prompt or budget input is persisted and the API publishes `active-input`.
3. The shared Ably subscriber event channel is full, so that message is dropped before `api_ably_supervisor` can forward the run ID.
4. The API source remains in `notifications.wait()`; its supplied 250 ms timeout has no effect.
5. The pending input is not discovered until another notification happens to arrive or a server-side recovery path re-announces it.

Prompt inputs do have a stale-queue repair path, but only after they are at least five minutes old and through a bounded sweep. Budget inputs are re-announced by their own minute cron. Neither path gives the runner a bounded progress guarantee, and the active run can finish before recovery.

## Impact

A user steer or runtime-budget handoff can remain durably queued while the current run never receives it. This produces a misleading state where the platform accepted the input but the agent continues without it, potentially completing or reaching its runtime limit first.

It also leaves reliability split across unrelated server-side sweeps instead of keeping notification delivery advisory and the list/claim API authoritative at the runner boundary.

## Acceptance criteria

- API active-input notification waits have a bounded, low-frequency fallback that rechecks the durable list endpoint after a successful empty read, without requiring a second notification.
- A delivered notification still wakes the forwarder immediately.
- The fallback cadence is explicitly separate from the 250 ms transient-read retry added by #25409 and does not introduce high-frequency polling for every active run.
- The source/wait abstraction accurately represents or enforces the bounded-progress contract instead of silently ignoring its timeout for the API variant.
- Stop and job-cancellation tokens interrupt both notification waits and fallback delays.
- Add a deterministic regression test where the initial API list is empty, a later input becomes pending, its notification is suppressed, and the input is claimed and forwarded via the bounded fallback.
- Keep regression coverage showing that the normal notification path remains immediate and that the transient-read failure behavior from #25409 is preserved.

## Related

- #25409 fixes the distinct case where a notification is consumed before a transient API read failure. It intentionally leaves successful empty reads notification-driven and does not cover a notification dropped before the runner receives it.

seven332:
## Summary

- make API active-input sources recheck durable state after either a matching notification or a 30-second fallback
- keep local active-input polling and transient source-read retries at their existing 250 ms cadence
- add a deterministic HTTP-to-guest-control regression that suppresses the notification and advances the production fallback interval

## Scope

- the normal Ably notification path remains immediate
- stop and job-cancellation ownership remains in the existing biased forwarder selection
- no API, database, persisted-state, Ably payload, guest-control protocol, feature switch, or rollout compatibility changes
- healthy idle API load is bounded to two list requests per minute per eligible active run

## Validation

- `cargo test --manifest-path crates/Cargo.toml -p runner executor::tests::agent_run_tests::active_input`
- new missing-notification regression repeated 10 times
- `cargo test --manifest-path crates/Cargo.toml -p runner --all-features` (`3084 passed`, `20 ignored`; guest inventory passed)
- pre-commit workspace Rust formatting, Clippy, rustdoc, file-size, and commit-message checks

Closes #25868

### DIFF
--- a/crates/runner/src/active_input.rs
+++ b/crates/runner/src/active_input.rs
@@ -91,6 +91,8 @@ pub(crate) enum ActiveInputBatch {
     Api(Option<String>),
 }
 
+const LOCAL_ACTIVE_INPUT_POLL_INTERVAL: Duration = Duration::from_millis(250);
+pub(crate) const API_ACTIVE_INPUT_RECHECK_INTERVAL: Duration = Duration::from_secs(30);
 const ACTIVE_INPUT_NOTIFICATION_CAPACITY: usize = 256;
 
 #[derive(Clone)]
@@ -188,10 +190,15 @@ impl ActiveInputSource {
         }
     }
 
-    pub(crate) async fn wait(&mut self, timeout: Duration) {
+    pub(crate) async fn wait_until_next_read(&mut self) {
         match self {
-            Self::LocalQueue(_) => tokio::time::sleep(timeout).await,
-            Self::Api(source) => source.notifications.wait().await,
+            Self::LocalQueue(_) => tokio::time::sleep(LOCAL_ACTIVE_INPUT_POLL_INTERVAL).await,
+            Self::Api(source) => {
+                tokio::select! {
+                    () = source.notifications.wait() => {}
+                    () = tokio::time::sleep(API_ACTIVE_INPUT_RECHECK_INTERVAL) => {}
+                }
+            }
         }
     }
 }
--- a/crates/runner/src/executor/active_input.rs
+++ b/crates/runner/src/executor/active_input.rs
@@ -103,7 +103,7 @@ async fn run_forwarder(
                 if retry_after_read_error {
                     tokio::time::sleep(ACTIVE_INPUT_READ_RETRY_INTERVAL).await;
                 } else {
-                    source.wait(ACTIVE_INPUT_READ_RETRY_INTERVAL).await;
+                    source.wait_until_next_read().await;
                 }
             } => {}
         }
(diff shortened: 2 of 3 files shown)


seven332:
## Code Review: PR #25888

**Reviewed head:** `e853cc25965a4b4343b9a2e41978a8e4f27fc154`

### Summary

The API active-input source now treats Ably as an immediate wakeup while rechecking durable state every 30 seconds when a notification is lost. Local polling, transient read retries, cancellation ordering, API contracts, and persisted data remain unchanged.

### Findings

No findings.

### Testing

The regression covers the production runner path through real loopback HTTP request construction and guest process control while intentionally omitting the notification. The focused active-input tests, full runner suite, formatting, Clippy, and rustdoc checks passed; the new regression also passed 10 consecutive runs. Production soak/load testing was not run, but idle request load is directly bounded to two list requests per minute per eligible run.

### Verdict

LGTM

PR #25889, opened 2026-08-09
fix(runner): preserve codex catalog identity responses

### LINKED ISSUE #25869: bug(python): restore identity bodies for Codex catalog miss owners
## Problem and evidence

On current `main` (`70a816e9074de62797f3c586338d591eed1704c5`),
an ordinary Codex model-catalog request can become the cache miss owner and
receive a Brotli response even though it requested `Accept-Encoding: identity`.

`codex_model_catalog_cache.prepare_request()` currently rewrites every admitted
owner to `Accept-Encoding: br`. The response stream wrapper returns the
compressed chunks downstream unchanged and decodes only a private copy later
for cache validation. A focused real-flow reproduction observed:

```text
{
  'client_accept_encoding': 'identity',
  'provider_accept_encoding': 'br',
  'downstream_content_encoding': 'br',
  'downstream_is_compressed': True
}
```

This is safe for the speculative prefetch producer because its `curl` process
discards the body. It is not safe when a real Codex request owns the miss.

The runner pins Codex `0.147.0`. That release's model client parses the
transport bytes directly with `serde_json::from_slice()`, and its Reqwest
dependency does not enable Brotli decoding:

- [`codex-rs/codex-api/src/endpoint/models.rs`](https://github.com/openai/codex/blob/rust-v0.147.0/codex-rs/codex-api/src/endpoint/models.rs#L64-L76)
- [`codex-rs/http-client/src/transport.rs`](https://github.com/openai/codex/blob/rust-v0.147.0/codex-rs/http-client/src/transport.rs#L111-L134)
- [`codex-rs/http-client/Cargo.toml`](https://github.com/openai/codex/blob/rust-v0.147.0/codex-rs/http-client/Cargo.toml#L12-L24)

Existing real mitmproxy tests confirm compressed owner delivery. Historical
controlled validation also found that the provider's Brotli catalog commonly
omits `Content-Length`.

## Decision

Proceed with role-aware upstream representation negotiation.

- Ordinary Codex miss owners preserve `Accept-Encoding: identity` and use the
  normal identity streaming/capture path.
- Only the runner-owned prefetch producer rewrites its upstream request to
  `Accept-Encoding: br`.
- A matching real Codex request consumes the producer's validated identity
  cache entry as a completed or in-flight consumer.
- A provider that ignores an ordinary identity request and returns an encoded
  representation is rejected before encoded headers or body reach Codex.

This replaces the earlier proposal to transcode Brotli for an ordinary owner.
With mitmproxy 12.2.3, buffering a missing-length response delays downstream
delivery but allows the proxy's internal body buffer to grow before addon
validation. Streaming keeps the existing bound but sends headers before the
decoded length and final validity are known. A single ordinary transcoding path
therefore cannot simultaneously provide exact decoded framing, pre-delivery
validation, missing-length support, and a strict encoded-input bound.

The cache already records `prefetch_owner`, so the two roles can use different
upstream negotiation without new state or protocol fields. The selected design
removes the root cause—overriding the real client's identity contract—while
retaining the measured Brotli transfer saving on the intended speculative
prefetch path.

For an unsolicited encoded ordinary response, reuse the previously proven
fixed empty 502 plus body-drain lifecycle. Record the existing
`model_catalog_cold_not_stored` / `response_encoding` telemetry, publish no
entry, and release any waiter. Do not add retry traffic, synchronous stream
decode, a process-wide mitmproxy body limit, or new telemetry values.

## Alternatives considered

- **Buffer and transcode every owner:** gives exact identity length and clean
  errors, but the common missing-length response is unbounded inside mitmproxy.
- **Stream-decode every owner:** keeps input bounded but cannot supply exact
  length or replace a successful status after complete validation; strict
  callback validation would reverse the off-event-loop work from #25038.
- **Disable Brotli everywhere:** correct but discards the established prefetch
  transfer benefit.
- **Change or wait for Codex:** not under VM0's deployment control and does not
  justify violating the current identity request.

## Risks and mitigations

- **Larger fallback cold miss:** only non-prefetched ordinary owners transfer
  the identity catalog; successful prefetch keeps Brotli and one provider GET.
- **Role confusion:** derive policy only from the existing stripped marker and
  persisted owner role; test follower takeover after producer failure.
- **Encoded body leakage:** replace the response during `responseheaders`, stop
  the normal header pipeline, drain upstream data, and assert emitted HTTP/2
  events contain no provider body or encoding.
- **Follower stalls or duplicate work:** preserve current bounded publication,
  cancellation, takeover, and terminal cleanup behavior and test it at hook
  entry points.
- **Event-loop regression:** retain executor-backed complete validation and add
  no synchronous decompression or JSON parsing to stream callbacks.

## Acceptance criteria

- An ordinary admitted Codex miss preserves its identity upstream negotiation
  and receives the provider's exact identity bytes. No encoded representation
  header or body is introduced by the cache.
- Valid ordinary identity responses remain supported with or without
  `Content-Length` under the existing bounded capture and decoded-body limits.
- If the provider returns a non-identity representation to an ordinary owner,
  Codex receives only a fixed bounded proxy error; encoded bytes are drained,
  no cache entry is installed, and waiters are released.
- A prefetch producer still negotiates Brotli and supports a valid response
  without `Content-Length` within existing input/output bounds. Successful
  validation publishes exact identity JSON.
- Matching completed and in-flight real Codex consumers receive that exact
  identity entry with one producer provider request.
- Invalid, incomplete, or oversized producer Brotli responses are not cached;
  failure and follower takeover remain bounded and do not create concurrent
  duplicate provider requests.
- Real mitmproxy HTTP/2 lifecycle coverage asserts ordinary identity delivery,
  unsolicited-encoding rejection, and the prefetch producer/follower path with
  exact downstream bytes and headers.

## Delivery notes

This is one independently reviewable Python addon PR. It changes no database,
migration, persisted data, queue payload, runner/guest protocol, API contract,
frontend, feature switch, or network-log schema. Rollback is the runner image;
no compatibility cleanup issue is required.

seven332:
## Summary

- preserve `Accept-Encoding: identity` for ordinary Codex model-catalog miss owners
- negotiate Brotli only for runner-owned prefetch producers and continue publishing validated identity cache entries to waiting Codex consumers
- reject unsolicited encoded ordinary responses with an empty 502 while draining the upstream body and releasing singleflight waiters
- cover ordinary, prefetch, failure, takeover, and missing-`Content-Length` behavior through the real mitmproxy HTTP/2 lifecycle

## Compatibility

- no database, persisted state, queue payload, runner/guest protocol, API, frontend, feature switch, or network-log schema changes
- rollback is limited to the runner image; no cleanup issue is required

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (4,993 passed)
- `git diff --check`

Closes #25869

### DIFF
--- a/crates/runner/mitm-addon/src/codex_model_catalog_cache.py
+++ b/crates/runner/mitm-addon/src/codex_model_catalog_cache.py
@@ -42,6 +42,7 @@ _ASCII_CONTROL_BOUNDARY = 0x20
 _ASCII_DELETE = 0x7F
 _HTTP_STATUS_SWITCHING_PROTOCOLS = 101
 _HTTP_STATUS_OK = 200
+_HTTP_STATUS_BAD_GATEWAY = 502
 _FLOW_STATE = "_codex_model_catalog_cache_state"
 _FLOW_TELEMETRY = "_codex_model_catalog_cache_telemetry"
 _PREFETCH_REQUEST = "_codex_model_catalog_prefetch_request"
@@ -641,7 +642,8 @@ async def prepare_request(flow: http.HTTPFlow, *, request_end_stream: bool) -> b
             prefetch_owner=is_prefetch,
         )
         flow.metadata[_FLOW_STATE] = state
-        flow.request.headers["Accept-Encoding"] = _BROTLI_ENCODING
+        if is_prefetch:
+            flow.request.headers["Accept-Encoding"] = _BROTLI_ENCODING
         return wait_deadline is not None
 
 
@@ -695,14 +697,34 @@ def _bypass_response(
     _release_flow_capacity(state)
 
 
+def _discard_upstream_response_body(_chunk: bytes) -> bytes:
+    return b""
+
+
+def _reject_encoded_response(
+    flow: http.HTTPFlow,
+    state: _FlowState,
+) -> None:
+    _set_not_stored(flow, state, "response_encoding")
+    state.finalized = True
+    _release_flow_capacity(state)
+    flow.response = http.Response.make(
+        _HTTP_STATUS_BAD_GATEWAY,
+        b"",
+        {"Content-Type": "text/plain"},
+    )
+    flow.response.stream = _discard_upstream_response_body
+
+
 def handle_response_headers(flow: http.HTTPFlow) -> bool:
     """Prepare catalog handling and tell mitm_addon.responseheaders() whether to continue.
 
-    Return False only for a fresh response served from the local catalog cache; the hook
-    must stop the normal response-header pipeline in that case. Return True for all other flows,
-    including unrelated traffic, cache bypasses, and eligible cold responses, so the hook
-    continues that pipeline. For an eligible cold response, ordinary streaming is installed before
-    wrap_response_stream() composes the bounded catalog capture.
+    Return False when the hook must stop the normal response-header pipeline: either for a
+    fresh response served from the local catalog cache or for an encoded ordinary response
+    replaced with a fixed proxy error. Return True for all other flows, including unrelated
+    traffic, cache bypasses, and eligible cold responses. For an eligible cold response,
+    ordinary streaming is installed before wrap_response_stream() composes the bounded catalog
+    capture.
 
     ``mitm_addon.responseheaders()`` implements this contract. Focused coverage is
     ``test_fresh_hit_is_partitioned_and_expiry_never_uses_conditions`` for the stop branch and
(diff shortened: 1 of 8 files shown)


seven332:
## Code Review: PR #25889

**Reviewed head:** `c18d37d4b18bf532a1b3d6a40deadf7644721ca2`

### Summary

The role-aware negotiation fixes the ordinary Codex catalog representation contract while preserving bounded Brotli prefetch, singleflight reuse, off-loop validation, and existing cache ownership limits.

### Findings

No findings.

### Testing

The change has focused response, coordination, lifecycle, hook, cancellation, and real mitmproxy HTTP/2 coverage. Local lock, format, lint, type, full Python addon tests (4,993 passed), and diff checks passed. Repository-wide Rust and Turbo suites were not run because no inputs or consumers in those scopes changed.

### Verdict

LGTM

PR #25896, opened 2026-08-09
fix(platform): keep agent chat default model cached

e7h4n:
## Summary

- stop reloading the user model preference whenever Agent Chat is entered
- make the landing composer model reset idempotent when no override exists
- verify the cached default model is present immediately after returning from `/agents`, even while model resource requests are blocked

## Regression proof

- the targeted test fails against the pre-fix behavior because the composer mounts without the model picker
- the fixed implementation passes all 45 tests in the target file

## Verification

- `pnpm format`
- `pnpm knip`
- `pnpm --filter @vm0/app exec vitest run src/views/zero-page/__tests__/chat-composer-model-selection-and-providers.test.tsx`
- `git diff --check`

### DIFF
--- a/turbo/apps/platform/src/signals/zero-page/agent-chat-page-setup.ts
+++ b/turbo/apps/platform/src/signals/zero-page/agent-chat-page-setup.ts
@@ -28,7 +28,6 @@ import {
   type EnsuredAgentDraft,
 } from "./agent-draft.ts";
 import { setAgentComposerContext$ } from "./agent-composer-signals.ts";
-import { reloadUserModelPreference$ } from "../external/user-model-preference.ts";
 import { openQueueDrawer$ } from "../queue-page/queue-drawer-state.ts";
 import { checkUnifiedSettingsParam$ } from "./settings/settings-dialog.ts";
 import { setupAgentChatKeyboardShortcuts$ } from "./agent-chat-keyboard.ts";
@@ -46,7 +45,6 @@ export const setupAgentChatPage$ = command(
     set(reloadTagline$);
 
     set(resetChatPageModelSelection$);
-    set(reloadUserModelPreference$);
 
     // Read agent ID from URL immediately (synchronous) and update sidebar
     // highlight early so the UI responds without waiting for async data.
--- a/turbo/apps/platform/src/signals/zero-page/zero-chat-page.ts
+++ b/turbo/apps/platform/src/signals/zero-page/zero-chat-page.ts
@@ -155,6 +155,8 @@ export const updateCodexFastModeDefaultForSelection$ = command(
   },
 );
 
-export const resetChatPageModelSelection$ = command(({ set }) => {
-  set(internalChatPageUserOverride$, { kind: "unset" });
+export const resetChatPageModelSelection$ = command(({ get, set }) => {
+  if (get(internalChatPageUserOverride$).kind === "set") {
+    set(internalChatPageUserOverride$, { kind: "unset" });
+  }
 });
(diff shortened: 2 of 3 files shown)


e7h4n:
LGTM

PR_AUTO_REVIEW_COMMIT: 92e17e8633668568b624666bc5798dec70a60d18
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

This change removes the unconditional user-model-preference reload from Agent Chat page setup and makes the new-chat model-selection reset idempotent. The resulting cache lifetime still has a semantic invalidation path: a successful preference update increments the preference reload state, while ordinary navigation back to Agent Chat preserves the already resolved default model. The added page-level regression test exercises the `/agents/:id/chat` → `/agents` → `/agents/:id/chat` flow with subsequent model-resource requests blocked and verifies that the cached model remains immediately visible.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness, regression risk, and required fix coverage | Pass | The invalidation chain was traced through `chatPageModelSelection$`, `userModelPreference$`, and `updateUserModelPreference$`; the fix includes a regression test that fails under the prior route-entry reload behavior. |
| `docs/docs.md` | Route the review to the relevant platform, ccstate, cache, and test practices | Pass | The diff is confined to platform ccstate page setup/model selection plus its page test, and each applicable practice was reviewed below. |
| `docs/bad-smell.md` | Avoid suppressions, unsafe typing, dynamic imports, broad fallbacks, and hidden recovery behavior | Pass | The diff adds none of those patterns and keeps the change focused on the existing cache-invalidation mechanism. |
| `docs/testing.md` | Prefer meaningful integration coverage at the production boundary | Pass | The regression is covered through full page bootstrap and route navigation, with HTTP behavior supplied at the test boundary. |
| `docs/effect.md` | Keep page setup as orchestration and express state transitions through semantic commands | Pass | `setupAgentChatPage$` continues to orchestrate route entry, while `resetChatPageModelSelection$` owns the conditional state transition. |
| `docs/cache.md` | Preserve computed memoization and invalidate only when dependencies actually change | Pass | Re-entering Agent Chat no longer bumps the preference reload state or rewrites an already-unset override; successful preference mutations still invalidate `userModelPreference$`. |
| `.claude/skills/ccstate/SKILL.md` | Let computed values cache derived async state and mutate source state via commands | Pass | `chatPageModelSelection$` remains a computed value, and both explicit selection/reset and preference-update invalidation remain command-driven without detached async work. |
| `docs/testing/patterns.md` | Use platform page tests with `detachedSetupPage` and lifecycle-scoped `context.mocks` | Pass | The test enters through `detachedSetupPage`, navigates through the real route command, and uses `context.mocks.api` plus `withSignal`. |
| `docs/testing/anti-patterns.md` | Assert outcomes rather than mock calls; avoid internal mocks, direct fetch mocks, and fake timers | Pass | The assertion is the visible model combobox; the test introduces no internal module mock, fetch stub, fake timer, or request-count-only assertion. |
| `docs/testing/app-testing.md` | Cover page-visible behavior through production bootstrap and scoped network overrides | Pass | The scenario starts on Agent Chat, visits the Agents page, returns via the router, and checks the rendered composer state while later network reads are held. |
| `docs/testing/testing-external-behavior.md` | Construct and observe the scenario through the user's page surface | Pass | State is established by loading and navigating pages, and the result is observed through the accessible model picker rather than store internals. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- The focused platform test file passed locally: 45/45 tests.
- The new regression case was red/green checked by temporarily restoring the previous page-entry reload behavior; it then failed because the composer mounted without the cached `Kimi K2.7 Code` picker, and passed again with this change.
- Current PR checks are green, including app tests, type checks, ESLint, formatting, Knip, CodeQL, Semgrep, and the aggregate Turbo gate.

PR #25900, opened 2026-08-09
feat(pi): persist native events across sandbox handoff

e7h4n:
## Summary

- persist native Pi transcript messages with API/sandbox source metadata and render an explicit API Server → Sandbox handoff marker in activity logs
- make handoff durable through the database, with standby initial catch-up and 10-second polling while keeping Ably as a realtime accelerator only
- cancel active runs on standby TTL and preserve main's current runner profile, provider, memory, usage, queue, and deployment-compatibility behavior

## Testing

- `pnpm format`
- `pnpm turbo run lint`
- `pnpm check-types`
- `pnpm knip`
- relevant Vitest coverage: 9 files / 102 tests
- broad API regression coverage: 276 tests, plus 5 isolated cron snapshot tests
- `pnpm -F @vm0/db test:migration-consistency`
- `pnpm -F @vm0/eslint-config test`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps`
- `cargo test --workspace`

### DIFF
--- a/crates/api-contracts/src/generated/types.rs
+++ b/crates/api-contracts/src/generated/types.rs
@@ -129,6 +129,26 @@ pub mod webhooks {
                 pub created_at: String,
             }
 
+            /// Durable control-plane boundary transferring a Pi run to its sandbox.
+            #[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
+            #[serde(rename_all = "camelCase")]
+            pub struct ResponseHandoff {
+                /// Run whose execution ownership is being transferred.
+                pub run_id: String,
+                /// Environment that committed the handoff.
+                pub from: String,
+                /// Environment that should resume the run.
+                pub to: String,
+                /// Transcript version containing the boundary.
+                pub transcript_version: i64,
+                /// Last transcript ordinal committed by the API.
+                pub after_ordinal: i64,
+                /// Assistant message that triggered the handoff.
+                pub message_id: String,
+                /// Timestamp when the handoff was committed.
+                pub requested_at: String,
+            }
+
             /// Versioned transcript returned to a standby Pi agent.
             #[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
             #[serde(rename_all = "camelCase")]
@@ -139,6 +159,8 @@ pub mod webhooks {
                 pub last_ordinal: u64,
                 /// Canonical ordered Pi messages.
                 pub messages: Vec<ResponseMessage>,
+                /// Durable API-to-sandbox ownership boundary, or null before handoff.
+                pub handoff: Option<ResponseHandoff>,
             }
         }
 
--- a/crates/guest-agent/src/cli/mod.rs
+++ b/crates/guest-agent/src/cli/mod.rs
@@ -289,8 +289,8 @@ pub enum PiStandbyReleaseReason {
     /// without calling `/complete` again.
     ApiComplete,
     /// The unused standby expired, so guest-agent exits with
-    /// [`guest_contracts::pi_standby::TTL_RELEASE_EXIT_CODE`] for runner
-    /// requeue instead of completing the run.
+    /// [`guest_contracts::pi_standby::TTL_RELEASE_EXIT_CODE`] so the runner
+    /// reports the expiry and the API cancels the run.
     Ttl,
 }
 
--- a/crates/guest-agent/src/cli/pi_agent_loop.rs
+++ b/crates/guest-agent/src/cli/pi_agent_loop.rs
@@ -51,11 +51,12 @@ enum PiOutputFrame {
     TranscriptRead {
         #[serde(rename = "requestId")]
         request_id: String,
+        version: Option<i64>,
+        #[serde(rename = "afterOrdinal")]
+        after_ordinal: Option<u64>,
     },
     #[serde(rename = "pi-message")]
     Message { event: Value },
-    #[serde(rename = "pi-transcript-conflict")]
-    TranscriptConflict,
     #[serde(rename = "pi-complete")]
     Complete {
         #[serde(rename = "exitCode")]
@@ -85,7 +86,6 @@ enum PiProtocolOutcome {
 
 struct PiProtocolState {
     ready: bool,
-    awaiting_conflict_marker: bool,
     last_acknowledged_sequence: Option<u32>,
     expected_system_prompt_digest: String,
     expected_skill_snapshot_digest: String,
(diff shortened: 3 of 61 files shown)


e7h4n (commented):
Changes Requested

### Summary

This PR replaces Pi cold-start recovery with a database-authoritative API-to-sandbox handoff, incremental standby polling, native event persistence/rendering, and TTL cancellation. The core integration matrix is broad, but the current boundary can export unmasked sandbox secrets to Axiom and can let the API fail a run after the sandbox has already observed the durable ownership transfer.

### Review Criteria

| Source | Applied criteria | Result |
| --- | --- | --- |
| `REVIEW.md` | Correctness, security, type safety, maintainability, feature gating, DB/JSONB review, and required feature integration coverage | **Fail** — P0 security/correctness findings and P1 cleanup findings below |
| `docs/docs.md` | Route the changed API, platform, persisted-event, runner, and test surfaces to all applicable practice documents | Pass — all matching indexed practices below were applied |
| `docs/bad-smell.md` | Fail fast; avoid hidden defaults, dead compatibility paths, suppressions, and hard-coded protocol behavior | **Fail** — the retired `cold-start` protocol remains accepted and tested |
| `docs/testing.md` | Integration-first coverage, boundary-only mocks, behavioral assertions, and regression coverage | Partial — broad integration coverage is present, but the post-commit telemetry failure boundary is untested |
| `docs/fallback.md` | Remove a retired fallback's producer, reader, contract entry, and tests together; do not retain compatibility for a default-off feature | **Fail** — `cold-start` cleanup is incomplete |
| `docs/event-sourcing.md` | Persistent state is authoritative and reconciles through stable event identities | **Fail** — a committed handoff can still flow into terminal failure |
| `docs/deployment-compatibility.md` | Review both directions of DB/API and runner/backend skew; apply the non-GA feature-switch exception | Pass — the new table is additive and PiLoop is default-off; no GA compatibility path is required |
| `docs/effect.md` | Commands own asynchronous effects, propagate cancellation, and keep render work pure | Pass |
| `docs/cache.md` | Cache/resource lifetime follows the owning scope and avoids render-time resource creation | Pass |
| `.claude/skills/ccstate/SKILL.md` | API/platform commands receive and propagate `AbortSignal`; signals remain side-effect free | Pass |
| `docs/testing/patterns.md` | Prefer real dependencies and representative boundary fixtures | Pass |
| `docs/testing/anti-patterns.md` | No internal-module mocks, fake timers, fetch stubs, DB mocks, or mock-only assertions | Pass |
| `docs/testing/api-testing.md` | Exercise API behavior through routes with the real test database and external-boundary mocks | Pass |
| `docs/testing/testing-external-behavior.md` | Set up and verify behavior through public API/page boundaries | Pass |
| `docs/testing/app-testing.md` | Use page-level platform setup and context-provided boundary mocks | Pass |
| `docs/testing/cli-testing.md` | Exercise CLI behavior through its public protocol boundary; reserve direct tests for the complex state-machine exception | Pass |
| `docs/testing/rust-testing.md` | Prefer integration coverage for HTTP/process behavior and inline tests only for narrow internal state matrices | Pass |

### Findings

- **`turbo/apps/api/src/signals/services/agent-event-consumer-axiom.service.ts:18` — native Pi events now cross the telemetry boundary without secret masking.** The Pi guest bridge explicitly preserves the exact payload “without passing it through the legacy CLI secret masker” (`crates/guest-agent/src/cli/pi_agent_loop.rs:4-6`), whereas the ordinary guest event path masks event-controlled content before delivery (`crates/guest-agent/src/events.rs:80-87`). Forwarding `eventData: event` therefore copies any credential echoed in a Pi tool call/result into Axiom unchanged, and `webhooks-agent-events-telemetry.test.ts:139` now positively locks in that raw transcript delivery. This violates the `REVIEW.md` security gate. Keep the canonical DB transcript native if required, but mask or redact the Axiom projection before ingest.

- **`turbo/apps/api/src/signals/services/pi-edge-loop.service.ts:219-230` — a fallible Axiom write happens after the durable handoff commit but before ownership is considered transferred.** `materializeRunOutputEvents$` commits the `pi_run_handoffs` row, so the standby's initial/10-second DB polling can immediately observe the boundary. If the subsequent Axiom ingest throws or hits its 10-second timeout, `project()` never reaches `PiHandoffRequested`; lines 336-364 instead classify the edge turn as failed and terminally complete the run while the sandbox can already be executing the committed tool batch. That creates two contradictory owners and can allow external tool side effects after a failed terminal result. Per `docs/event-sourcing.md`, the persisted handoff must be the irreversible authority: once it commits, later telemetry/publication failures cannot take the terminal-failure path. Add an integration case for Axiom failure immediately after a sandbox-required message commits.

- **`crates/runner/src/provider/api.rs:672` — the retired `cold-start` fallback is only half removed.** This PR removes its producer/requeue path and makes the runner reject `PiExecutionMode::ColdStart`, but `turbo/packages/api-contracts/src/contracts/runners.ts:423` still exposes it, `crates/runner/src/types.rs:34-37` still deserializes it, `turbo/apps/api/src/signals/routes/runners.ts:1043-1047` still accepts it as a valid persisted execution context, and contract/direct-candidate tests still preserve it. `docs/fallback.md` requires a fallback removal to delete the branch, contract entry, and fallback-specific tests together; PiLoop is default-off, so no compatibility reader is needed. Remove the remaining mode and its tests so the API cannot accept/emit a value that the runner deterministically refuses to claim.

- **`crates/guest-agent/src/pi_standby.rs:41-44` — the documented cross-language protocol still requires a frame that this PR deletes.** The implementation removes `pi-transcript-conflict`, the guest's conflict state, and CLI replay, yet the module-level contract says a 409 must emit that frame and re-read/replay before completion. Lines 51-54 explicitly declare this text the compatibility contract with the TypeScript child. Update the lifecycle contract to the new fail-fast CAS behavior so the documented protocol matches both implementations.

### Testing

- Coverage: **Insufficient for the identified boundaries** — broad API, CLI state-machine, platform page, Rust HTTP, and runner coverage is present, but there is no regression test proving that an Axiom failure after handoff commit cannot terminally fail the run; the telemetry test currently asserts the unmasked payload behavior.
- Conventions: **Compliant** — changed tests use route/page/protocol boundaries, the real API test database, MSW/httpmock for external HTTP, and no prohibited internal mocks or fake timers.
- CI: All current PR checks passed, including Turbo, Crates, security scanning, CodeQL, and Codecov.

e7h4n:
Superseded by #25906, which now includes the durable native Pi event persistence and Activity handoff UI on top of transcript-driven session polling, while keeping Axiom payloads redacted and ownership transfer independent of telemetry.

PR #25903, opened 2026-08-09
refactor(platform): make chat event pagination termination explicit

e7h4n:
## Summary

- replace two unconditional `for (;;)` chat-event pagination loops with an explicit full-page continuation condition
- keep the shared `setLoop` implementation as the only match in the custom-loop scan
- preserve the existing command-owned `AbortSignal`, row cursor updates, cache writes, and cursor-expiry recovery

## Violations found

The snapshot-read foreground and background row synchronizers used unconditional `for (;;)` control flow for finite pagination. Their stopping condition was only expressed through interior returns, so they appeared as unbounded platform loops outside the shared loop abstraction.

## Signal ownership

This change adds no timer. Each pagination pass continues to use the command lifecycle's `AbortSignal` for the row request, IndexedDB operations, snapshot recovery, and abort checks.

## Behavior preserved

- a full page advances the cursor and requests the next page
- a partial or empty page ends the one-shot traversal
- background cursor expiry clears the row cache and ends that pass
- foreground cursor expiry rebuilds once from the server snapshot and continues; a second expiry still fails loudly

## Verification

- Prettier on both changed files
- changed-file ESLint and Oxlint, including type-aware Oxlint
- targeted Vitest: `chat-event-snapshot-read.test.ts` (6 passed)
- custom-loop scan only matches the shared `setLoop` implementation
- native Platform timer scan is empty
- Lefthook pre-commit: Prettier, Knip, full repository type checks (14/14), file-size check, and commitlint passed

Full CI validation is delegated to the PR pipeline.

## Fallbacks

None.

### DIFF
--- a/turbo/apps/platform/src/signals/chat-page/chat-event-background-sync.ts
+++ b/turbo/apps/platform/src/signals/chat-page/chat-event-background-sync.ts
@@ -107,7 +107,8 @@ const syncChatThreadRowsToIndexedDb$ = command(
 
     const syncedEvents: ChatEvent[] = [];
     let sinceSeqId = lastSeqId;
-    for (;;) {
+    let shouldLoadNextPage = true;
+    while (shouldLoadNextPage) {
       const page = await set(listRowsAfter$, { threadId, sinceSeqId }, signal);
       signal.throwIfAborted();
       if (page.kind === "expired") {
@@ -129,10 +130,9 @@ const syncChatThreadRowsToIndexedDb$ = command(
         }),
       );
       sinceSeqId = page.rows.at(-1)!.seqId;
-      if (page.rows.length < CHAT_EVENT_ROWS_PAGE_LIMIT) {
-        return syncedEvents;
-      }
+      shouldLoadNextPage = page.rows.length === CHAT_EVENT_ROWS_PAGE_LIMIT;
     }
+    return syncedEvents;
   },
 );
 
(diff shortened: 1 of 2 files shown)


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 37436100bb5e219fd17152d998b993e5e252289a
PR_AUTO_REVIEW_VERDICT: LGTM

## Summary

This refactor replaces the two unconditional chat-event row pagination loops with explicit full-page continuation state. Request ordering, cursor advancement, cursor-expiry recovery, event projection, cache writes, and command-owned cancellation remain unchanged.

| Review source | Result |
| --- | --- |
| `REVIEW.md` | The full-page and partial-page exit paths are behaviorally equivalent to the prior loops; no correctness, security, type-safety, or compatibility blocker found. |
| `docs/docs.md` | This internal control-flow refactor does not change public behavior or require user-facing documentation. |
| `docs/bad-smell.md` | No suppression, unsafe cast, dynamic import, duplicated fallback, or broad error handling was introduced. |
| `docs/testing.md` | Existing integration coverage exercises cold-start pagination, cached rows, cursor expiry/rebuild, immediate expiry failure, and background sync. |
| `docs/event-sourcing.md` | Persisted chat-event rows remain authoritative; cursor and projection semantics are unchanged. |
| `docs/effect.md` | No detached effect or new timer was introduced; async work remains awaited within the owning command. |
| `docs/cache.md` | IndexedDB read/write ordering and invalidation behavior are unchanged, including the existing abort checks around asynchronous boundaries. |
| `.claude/skills/ccstate/SKILL.md` | Existing command composition and AbortSignal ownership are preserved; no accessor escape or orphaned async work was added. |

## Findings

### Critical (P0)

None.

### High (P1)

None.

## Testing

- Targeted `chat-event-snapshot-read.test.ts`: 6/6 tests passed locally.
- Changed-file formatting and lint passed locally.
- Platform type checking and the full Lefthook type-check suite passed locally.
- At review time, every completed PR check is green; the remaining CodeQL, app type-check, and Playwright jobs are still running.

PR #25912, opened 2026-08-09
fix(platform): prevent chat crash on browser back

### LINKED ISSUE #25905: bug(platform): browser back from activities crashes chat on mobile Chrome
## Background

A user reported a navigation crash in the Platform app on mobile Chrome.

## Reproduction steps

1. Open a chat whose transcript contains an agent-run source annotation.
2. Open an activity detail from the chat (`/activities/:runId`).
3. Press the browser or system Back button to return to the chat.

## Observed behavior

The chat route begins to load, then the app renders the global error boundary with “Oops! Something went sideways.”

## Expected behavior

Browser Back should restore the previous chat route and content without rendering resources owned by an aborted panel.

## Environment

- Platform: `app.vm0.ai`
- Reported browser: Chrome Mobile 150 on Android 10
- Reported: 2026-08-09

## Evidence

Mobile Chrome global error boundary after navigating back from Activities

The screenshot timestamp matches [Sentry PLATFORM-85](https://vm0.sentry.io/issues/7660139665/):

- At 2026-08-09 08:24:28 UTC, navigation changed from `/activities/:runId` to `/chats/:threadId`.
- React then threw `Error: Card signals were not registered` from `registeredCardSignals`, called by `SourceMessageAnnotation` through `agentReferenceSignalsForId`.
- The issue has affected three users across Chrome Mobile, Mobile Safari, and desktop Chrome, including returns from other non-chat routes.

The route setup installs the chat page before its replacement panel finishes resolving. Pane references are intentionally retained during chat-to-chat setup so the outer thread section keeps its DOM identity. After leaving chat, however, the previous pane reference remained stored after its owner signal aborted. Browser Back could therefore mount the chat page with the already-aborted `ChatPanelSignals`.

`useLastResolved` correctly recovered that panel history, but the aborted signal had already reclaimed the panel-owned agent-reference registry. Rendering the retained source annotation then reached the production exception.

The page-level regression navigates from the resource-backed chat to Activities and dispatches browser `popstate` back to the chat. It fails with the production exception before the fix and passes with the final signal-owned cleanup.

## Validated implementation direction

- Keep the panel `AbortSignal`, derived from the page/route signal, as the resource lifecycle source.
- When a pane owner aborts after leaving the chat route, clear the stored pane reference so chat re-entry cannot inherit an aborted panel.
- Preserve the pane reference during chat-to-chat setup and keep the outer `ChatThread` unkeyed, retaining its established DOM identity.
- Retain `useLastResolved` and its loading-skip semantics.
- Keep existing narrow render keys based on semantic `threadId`; do not add owner, lifecycle, or random remount keys.
- Use `threadId` as Mermaid chat scope and keep object URLs owned by the panel/page signal.
- Preserve strict registry resolution and existing chat event projections.

## Impact

Users cannot reliably return to a chat that contains an agent source annotation and must reload or reopen the app after the crash.

## Acceptance criteria

- Returning from an Activity detail to a previous chat with browser Back does not render the global error boundary.
- An aborted panel reference is not exposed to a newly mounted chat page.
- The previous chat route and content, including agent source annotations, are restored.
- The outer chat thread container preserves its DOM identity across chat-to-chat changes.
- `useLastResolved` keeps its existing loading behavior.
- Chat/page lifecycle IDs and random React remount keys are not used.
- A regression test covers re-entering a chat with an agent source annotation after navigating away.

Implemented by #25908.

seven332:
## Summary

- remount each chat pane's lifecycle-owned content when its `ChatPanelSignals` owner changes
- preserve the stable outer thread container while preventing old `useLastResolved` values from being paired with a new card registry
- cover Activity navigation followed by browser `popstate` back to a resource-backed chat

## Root cause

Returning to a chat can first expose the retained pane signals and then replace them with a new signal owner for the same thread ID. The unkeyed lifecycle-owned content stayed in the same React position, so its descendants could retain rendered groups from the old owner while receiving registry resolvers from the new owner. Rendering an agent source annotation through that mismatched registry threw `Card signals were not registered` and reached the global error boundary.

Keying `ChatThreadContent` by `lifecycleId` aligns React state retention with the existing signal ownership boundary. Stable `threadId` remains the data identity, the outer thread container retains its DOM identity, and every new resource owner receives a fresh content subtree.

## Validation

- `pnpm format`
- `pnpm knip`
- staged workspace, Platform, and API type checks
- `pnpm --filter @vm0/app exec vitest run src/views/zero-page/__tests__/chat-user-messages.test.tsx --silent=passed-only` (7 tests)
- `pnpm --filter @vm0/app exec vitest run src/views/zero-page/__tests__/chat-lifecycle.test.tsx --silent=passed-only` (30 tests)
- follow-up Platform lint and type checks after preserving the container boundary

Closes #25905

### DIFF
--- a/turbo/apps/platform/src/views/zero-page/__tests__/chat-user-messages.test.tsx
+++ b/turbo/apps/platform/src/views/zero-page/__tests__/chat-user-messages.test.tsx
@@ -9,6 +9,7 @@ import {
   queryAllByRoleFast,
 } from "../../../__tests__/page-helper.ts";
 import { testContext } from "../../../signals/__tests__/test-helpers.ts";
+import { pathname, setPathname, setSearch } from "../../../signals/location.ts";
 import { mockChatLifecycle } from "./chat-test-helpers.ts";
 
 const context = testContext();
@@ -480,11 +481,12 @@ describe("user messages", () => {
     ).toBeInstanceOf(HTMLElement);
   });
 
-  it("renders agent-run source annotations with avatar, link, and run anchor", async () => {
+  it("restores agent-run source annotations after browser Back from Activities", async () => {
+    const user = userEvent.setup({ delay: null });
     const threadId = "b0000000-0000-4000-a000-000000000750";
     const sourceThreadId = "b0000000-0000-4000-a000-000000000751";
     const sourceRunId = "d0000000-0000-4000-a000-000000000751";
-    const targetRunId = "d0000000-0000-4000-a000-000000000750";
+    const targetRunId = "a0000000-0000-4000-a000-000000000001";
     const sourceAgentId = "a1000000-0000-4000-a000-000000000010";
     const sourceAgentAvatarUrl = "https://example.com/source-agent-avatar.png";
     context.mocks.data.team([
@@ -533,6 +535,13 @@ describe("user messages", () => {
           },
           createdAt: "2026-08-04T10:00:00Z",
         },
+        {
+          id: "00000000-0000-4000-8000-000000000752",
+          role: "assistant",
+          content: "Delegated response",
+          runId: targetRunId,
+          createdAt: "2026-08-04T10:00:01Z",
+        },
       ],
     });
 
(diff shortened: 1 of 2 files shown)


seven332:
Reviewed current head `99816e6a3b0954cab02ba37cbcffa800bbd4aada` against the Platform lifecycle, chat-card ownership, and page-test conventions.

No unresolved P0/P1/P2 findings. The follow-up commit moves the lifecycle key below the stable thread container, so lifecycle-owned hooks remount without replacing the pane's host element. The browser-Back regression passes (7/7), the complete chat lifecycle file passes (30/30), and Platform lint/type checks pass locally.

PR #25922, opened 2026-08-09
test(mitm-addon): cover content-type whitespace limits

### LINKED ISSUE #25872: test(python): pin bounded Content-Type whitespace scanning
## Problem

At current `main` (`c61b9c1ddc6591278d771451799316626c5d6b86`), `crates/runner/mitm-addon/src/body_capture.py:156-186` bounds attacker-controlled `Content-Type` scanning in two places. Leading optional whitespace and optional whitespace after the media type are each rejected after `_MAX_CAPTURE_HEADER_VALUE_TO_PRESERVE` characters.

The existing header tests cover ordinary OWS, overlong media-type text, and large parameters that are discarded after an early semicolon. They do not reach either OWS limit at `body_capture.py:169-170` or `176-177`.

## Mutation evidence

Deleting both OWS limit checks left all 201 body-capture tests green:

```text
tests/test_body_capture_headers.py
tests/test_body_capture_fields.py
tests/test_body_capture_stream_buffer.py
tests/test_body_capture_encoding.py
tests/test_body_capture_decompression.py

201 passed
```

External boundary cases distinguish current behavior from the mutant:

```text
value: 257 spaces + application/json
current: ***
mutant:  application/json

value: application/json + 257 spaces
current: ***
mutant:  application/json
```

## Impact

`Content-Type` is untrusted input processed on capture-enabled request and response paths. These checks are what prevent a whitespace-only prefix or suffix from forcing an unbounded scan before a small allowlisted media type is retained. Without mutation-sensitive tests, a refactor can remove that bounded-work contract while normal sanitization cases continue to pass.

## Acceptance criteria

- Add table-driven coverage for leading and post-media-type OWS at exactly 256 and 257 characters.
- Include a representative post-media-type case both with and without a following parameter delimiter.
- Assert values at the current bound preserve `application/json` and values beyond it are redacted as `***`.
- Exercise real captured-header sanitization without mocking `_sanitize_content_type_for_capture()`.
- Ensure deleting either OWS early-return guard fails the suite.
- Keep the change test-only; no production sanitizer behavior is required to change.

seven332:
## Summary

- cover the 256/257-character leading content-type OWS boundary
- cover the same post-media-type boundary at end-of-value and before a parameter delimiter
- assert captured-header preservation and redaction through real mitmproxy headers

## Scope

- test-only change; production sanitizer behavior is unchanged
- no API, data, dependency, deployment, or documentation changes

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_body_capture_headers.py` (82 passed)
- independent mutation checks: removing the leading guard failed `leading-over-limit`; removing the post-media-type guard failed both post-media-type over-limit cases
- `uv run --no-sync python -m pytest tests/` (5,073 passed)

Closes #25872

### DIFF
--- a/crates/runner/mitm-addon/tests/test_body_capture_headers.py
+++ b/crates/runner/mitm-addon/tests/test_body_capture_headers.py
@@ -41,6 +41,45 @@ class TestSanitizeHeadersForCapture:
         assert result["Content-Type"] == expected
 
     @pytest.mark.parametrize(
+        ("value", "expected"),
+        [
+            pytest.param(
+                (" " * 256) + "application/json",
+                "application/json",
+                id="leading-at-limit",
+            ),
+            pytest.param(
+                (" " * 257) + "application/json",
+                "***",
+                id="leading-over-limit",
+            ),
+            pytest.param(
+                "application/json" + (" " * 256),
+                "application/json",
+                id="trailing-at-limit",
+            ),
+            pytest.param(
+                "application/json" + (" " * 257),
+                "***",
+                id="trailing-over-limit",
+            ),
+            pytest.param(
+                "application/json" + (" " * 256) + "; charset=utf-8",
+                "application/json",
+                id="before-parameter-at-limit",
+            ),
+            pytest.param(
+                "application/json" + (" " * 257) + "; charset=utf-8",
+                "***",
+                id="before-parameter-over-limit",
+            ),
+        ],
+    )
+    def test_content_type_optional_whitespace_is_bounded(self, headers, value, expected):
+        result = _sanitize_headers_for_capture(headers(("Content-Type", value)))
+        assert result["Content-Type"] == expected
+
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #25922

**Reviewed head:** `202bdec718cfeb5e52ad80b706486925551d38d9`

### Summary

The PR adds six focused 256/257-character content-type OWS boundary cases through the real captured-header sanitizer and mitmproxy headers. It preserves the approved test-only scope and changes no production behavior.

### Findings

No findings.

### Testing

The focused 82-test header suite, Ruff format/lint, BasedPyright, and all 5,073 mitm-addon tests pass locally. Independent mutations confirmed that deleting either OWS guard makes its corresponding over-limit cases fail. GitHub CI is still pending.

### Verdict

LGTM

PR #25930, opened 2026-08-09
test(runner): bound axiom channel overflow watchdog

### LINKED ISSUE #25883: test(rust): bound the Axiom channel-full nonblocking check
## Problem and evidence

`burst_past_channel_cap_drops_without_blocking_or_feeding_back` is intended to prove that Axiom channel overflow never blocks tracing callers. On current `main`, it fills the channel by synchronously emitting 3,000 events while monopolizing Tokio's current-thread test runtime, then checks that the completed loop took less than 500 ms.

That structure has two independent reliability problems:

- A blocking send or retry-loop regression can stall before the elapsed-time assertion is reached, leaving the surrounding CI job timeout as the effective watchdog.
- A healthy nonblocking implementation can exceed the absolute 500 ms throughput threshold on a contended host even though the behavioral invariant still holds.

The production layer currently uses a bounded Tokio MPSC channel with `CHANNEL_CAP = 1024` and `try_send`; no production defect is reported. The defect is in how the regression test establishes and bounds backpressure.

## Decision

Proceed with a test-only rewrite that separates the queue/backpressure invariant from dispatcher and HTTP batching behavior.

- Construct the real module-private `AxiomLayer` with a `CHANNEL_CAP` channel and retain its receiver without polling it. This makes saturation deterministic and independent of runtime scheduling or network mocks.
- Move a `tracing::Dispatch` for the filtered layer to a dedicated `std::thread`, emit `CHANNEL_CAP + 1001` events there, and report completion through a standard channel.
- Bound only producer completion with a five-second `recv_timeout`; do not measure or compare throughput.
- Join a completed producer and assert that the receiver contains exactly `CHANNEL_CAP` events, directly proving bounded acceptance and excess-event drops.
- Strengthen the existing filter/HTTP integration test to emit the production overflow diagnostic shape (`INTERNAL_TARGET`, `dropped`, and `axiom channel full`) and prove it remains locally visible but absent from Axiom ingest.

This uses the real tracing layer without adding a production constructor, test hook, feature switch, or abstraction. Surrounding tests continue to exercise the dispatcher, shutdown, and mocked HTTP boundary.

## Alternatives considered

- **Gate the first `httpmock` response:** preserves the full path in this one test, but requires bidirectional coordination and a blocking dynamic responder that `httpmock` advises against. It also retains indirect HTTP batch counts as a drop proxy.
- **Run the scenario in a bounded child process:** provides stronger isolation from arbitrary infinite loops but adds ignored-test and process-harness complexity that is not justified for this in-memory channel invariant.
- **Use `tokio::spawn_blocking` with an async timeout:** rejected because Tokio cannot abort a started blocking task and runtime shutdown can still wait indefinitely.
- **Keep or increase the fixed elapsed-time threshold:** rejected because it remains a host-dependent performance assertion and still runs after the potentially blocking work.

## Risks and mitigations

- **Distributed coverage:** the rewritten burst test no longer reaches HTTP. The existing filter/HTTP test will use the exact overflow diagnostic, and other Axiom tests retain real dispatcher drain and shutdown coverage.
- **Thread-local tracing state:** explicitly install the moved `tracing::Dispatch` inside the producer thread.
- **Intentional failure cleanup:** the independent watchdog fails locally; unwinding drops the retained receiver, releasing normal bounded-channel blocking without involving an unabortable Tokio blocking task.
- **Shared-host contention:** the watchdog has multi-second slack and asserts only completion, not expected execution speed.

## Acceptance criteria

- Channel saturation is deterministic and does not depend on starving the Tokio current-thread runtime.
- The event producer runs on a dedicated OS thread behind an independent, test-local completion watchdog.
- The test contains no sub-500 ms throughput assertion, arbitrary sleep, real external request, or indirect HTTP batch-count heuristic.
- Receiver occupancy proves that exactly `CHANNEL_CAP` events were accepted and excess events were dropped.
- The production-shaped `INTERNAL_TARGET` overflow diagnostic is observed by a sibling local layer and excluded from Axiom ingest.
- Production code, APIs, dependencies, schemas, protocols, and deployment behavior remain unchanged.
- Runner formatting, Clippy, rustdoc, focused/repeated tests, and the affected crate test suite pass.

## Delivery notes

This is one standalone, reviewable test-only PR with no migration, compatibility, or rollout work.

seven332:
## Summary

- saturate the Axiom layer's bounded channel deterministically with a retained, non-draining receiver
- run the tracing producer on a dedicated OS thread behind a five-second completion watchdog
- assert exact channel occupancy instead of host-dependent elapsed time and HTTP batch counts
- verify the production-shaped overflow diagnostic remains visible locally and excluded from Axiom ingest

## Scope

This is test-only. It does not change Axiom runtime behavior, dependencies, APIs, schemas, runner protocols, or deployment compatibility.

## Validation

- focused overflow and internal-filter tests
- 30 consecutive focused overflow-test runs
- mutation check: replacing `try_send` with `blocking_send` failed at the local watchdog in 5.01 seconds; restoring `try_send` passed
- `cargo test -p runner` (3,087 passed, 20 ignored; guest inventory integration test passed)
- pre-commit workspace Rust formatting, Clippy, rustdoc, and file-size checks

Closes #25883

### DIFF
--- a/crates/runner/src/axiom_layer_tests.rs
+++ b/crates/runner/src/axiom_layer_tests.rs
@@ -6,9 +6,15 @@
 //! endpoint with the TS-compatible payload shape.
 //!
 
-use std::sync::{Arc, Mutex};
+use std::sync::atomic::AtomicU64;
+use std::sync::{Arc, Mutex, mpsc};
+use std::thread;
+use std::time::Duration;
 
-use super::{INTERNAL_TARGET, init_from_env_values, init_with_base_url, with_ingest_filter};
+use super::{
+    AxiomLayer, CHANNEL_CAP, INTERNAL_TARGET, init_from_env_values, init_with_base_url,
+    with_ingest_filter,
+};
 use httpmock::Method::POST;
 use httpmock::MockServer;
 use httpmock::{HttpMockRequest, HttpMockResponse, Mock};
@@ -225,7 +231,7 @@ async fn axiom_filter_does_not_suppress_sibling_local_layers() {
         tracing::debug!("local debug");
         tracing::trace!("local trace");
         tracing::warn!("local warn");
-        tracing::warn!(target: INTERNAL_TARGET, "local internal");
+        tracing::warn!(target: INTERNAL_TARGET, dropped = 1_u64, "axiom channel full");
     }
     guard.shutdown().await;
 
@@ -253,7 +259,7 @@ async fn axiom_filter_does_not_suppress_sibling_local_layers() {
                 && event
                     .message
                     .as_deref()
-                    .is_some_and(|seen| seen.contains("local internal"))
+                    .is_some_and(|seen| seen.contains("axiom channel full"))
         }),
         "sibling local layer did not record internal-target event: {events:?}",
     );
@@ -262,7 +268,12 @@ async fn axiom_filter_does_not_suppress_sibling_local_layers() {
     let events = captured.events();
     let warning = event_with_message(&events, "local warn");
     assert_eq!(warning["level"], json!("warn"));
-    for message in ["local info", "local debug", "local trace", "local internal"] {
+    for message in [
+        "local info",
+        "local debug",
+        "local trace",
+        "axiom channel full",
+    ] {
         assert!(
             !has_event_with_message(&events, message),
             "filtered event {message:?} should not be ingested: {events:#?}",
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #25930

**Reviewed head:** `64b24620360dee279b6f16296785b229a87f43bd`

### Summary

The test-only rewrite isolates the bounded-channel invariant with a retained receiver, moves the producer behind an independent OS-thread completion watchdog, and strengthens the existing HTTP filter test with the production overflow diagnostic shape. It removes scheduler-dependent throughput and HTTP batch-count heuristics without changing production code or interfaces.

### Findings

No findings.

### Testing

Focused tests, 30 repeated overflow-test runs, a blocking-send mutation check, runner formatting, Clippy, rustdoc, the full runner test suite, and pre-commit workspace Rust checks passed. Turbo, Python, remote runner, and E2E suites were not run because the diff changes only a Rust inline test and has no runtime consumer. PR CI is evaluated separately.

### Verdict

LGTM

PR #25937, opened 2026-08-09
fix(mitm-addon): retain aliases across context target unpacking

### LINKED ISSUE #25892: bug(python): retain metadata aliases across suppressed context target unpack failures
## Summary

The flow metadata key linter treats a destructuring `with ... as (...)` target as if all target names are definitely rebound. Python must unpack the context manager result before assigning those names; unpacking can fail before any assignment, and the context manager can suppress that exception. On that path the previous `flow.metadata` alias remains live, but the linter drops it and misses a registered-key literal access.

## Evidence

Verified on `main` at `9535d4eafce5284585f87ad407c3908823c724d9`.

- `crates/runner/mitm-addon/scripts/flow_metadata_linter/ast_helpers.py:622` checks optional targets with `_expression_may_raise()`, which does not treat tuple/list unpacking itself as fallible.
- `crates/runner/mitm-addon/scripts/flow_metadata_linter/visitor.py:1046` enters the protected region and visits the optional target.
- `crates/runner/mitm-addon/scripts/flow_metadata_linter/visitor.py:1287` clears aliases while walking tuple/list target names without retaining the pre-binding state for an unpack failure.

Minimal input:

```python
def missed(flow, manager):
    metadata = flow.metadata
    with manager as (metadata, other):
        pass
    return metadata["firewall_name"]
```

When `manager` yields a wrong-sized iterable and suppresses the resulting unpack exception from `__exit__`, Python performs no target assignments and the return reads `flow.metadata["firewall_name"]`. The linter currently returns no violations for this input (`[]`).

## Impact

The registry-constant guard has a false-negative path: direct registered metadata key strings can pass the repository check even though they are used against `flow.metadata` at runtime.

## Acceptance criteria

- Model tuple/list/starred context-manager target unpacking as a possible failure before target bindings complete.
- Preserve pre-binding metadata aliases on the suppressible exception path while retaining the existing normal-path rebinding behavior.
- Add regression coverage for both synchronous and asynchronous context-manager destructuring targets.
- Keep the current repository metadata-key scan passing.

seven332:
## Summary

- model structural list/tuple context targets as fallible before child bindings
- retain pre-binding metadata aliases when sync or async context managers suppress unpack failures
- preserve post-`with` reachability when a terminal body is bypassed by a suppressed target failure
- cover sync tuple, async starred-list, terminal-body, and ordered later-target cases through the public linter API

## Scope

This changes only repository flow-metadata lint tooling and its fixture contract. It does not change runner or addon runtime behavior, schemas, persisted data, APIs, protocols, dependencies, deployment compatibility, or rollout behavior.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_flow_metadata_key_contract.py` (18 passed)
- `uv run --no-sync python scripts/check-flow-metadata-keys.py`
- `uv run --no-sync python -m pytest tests/` (5082 passed)
- mutation proof: removing the alias snapshot dropped the three new context diagnostics; removing structural reachability dropped only the terminal-body diagnostic

Closes #25892

### DIFF
--- a/crates/runner/mitm-addon/scripts/flow_metadata_linter/ast_helpers.py
+++ b/crates/runner/mitm-addon/scripts/flow_metadata_linter/ast_helpers.py
@@ -7,6 +7,11 @@ from dataclasses import dataclass
 from typing import TypeGuard
 
 
+def _target_requires_unpacking(node: ast.AST | None) -> bool:
+    """Return whether binding ``node`` starts with structural sequence unpacking."""
+    return isinstance(node, ast.List | ast.Tuple)
+
+
 def _target_names(node: ast.AST | None) -> set[str]:
     if isinstance(node, ast.Name):
         return {node.id}
@@ -621,7 +626,7 @@ def _annotation_target_may_raise(node: ast.AST) -> bool:
 
 def _with_protected_region_may_raise(items: list[ast.withitem], body_flow: _FlowSummary) -> bool:
     target_may_raise = any(
-        _expression_may_raise(item.optional_vars)
+        _target_requires_unpacking(item.optional_vars) or _expression_may_raise(item.optional_vars)
         for item in items
         if item.optional_vars is not None
     )
--- a/crates/runner/mitm-addon/scripts/flow_metadata_linter/visitor.py
+++ b/crates/runner/mitm-addon/scripts/flow_metadata_linter/visitor.py
@@ -21,6 +21,7 @@ from flow_metadata_linter.ast_helpers import (
     _static_first_call_argument_nodes,
     _static_truth_value,
     _target_names,
+    _target_requires_unpacking,
     _truth_test_may_raise,
     _type_alias_target_names,
     _type_alias_value,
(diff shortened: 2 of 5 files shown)


seven332:
## Code Review: PR #25937

**Reviewed head:** `4e94725423cadffb38cccfb5578b43c2b3e67232`

### Summary

The submitted head models the real pre-binding sequence-unpack failure in both alias transfer and context-statement reachability. The shared target predicate stays separate from ordinary expression evaluation, successful target traversal remains ordered, and the pre-binding snapshot is sufficient because context targets only remove metadata aliases.

### Findings

No findings.

### Testing

Reviewed validation on this head: uv lock/sync, Ruff format and lint, BasedPyright, the public flow-metadata contract tests, repository metadata-key scan, and the full mitm-addon suite (5082 tests) all passed. Independent mutation checks proved that the alias-transfer cases and terminal-body reachability case fail when their corresponding logic is removed. Required GitHub checks are handled separately and were not treated as complete by this review.

### Verdict

LGTM

PR #25948, opened 2026-08-09
test(python): cover malformed network log fallback

### LINKED ISSUE #25895: test(python): cover malformed network-log target fallback
## Problem and evidence

`crates/runner/mitm-addon/src/http_network_log.py` treats URL parsing and authority access as diagnostic work: if deriving a host or port raises `ValueError`, it falls back to `flow.request.pretty_host` and `flow.request.port` as one consistent pair. Authority-validation denials call this path before they install the local 403 response and later write the network log.

The existing authority-validation integration coverage only produces parseable fallback URLs. A malformed transparent destination such as `target.example:not-a-port` can produce `https://target.example:not-a-port/path` when an HTTPS request is denied for missing SNI. Accessing that URL's parsed port raises `ValueError`. Without the fallback, the request hook raises before completing the denial; with a partial fallback, the diagnostic target can mix the parsed host with the request port.

The branch is currently unprotected: changing the handler to re-raise still leaves the existing mitm-addon suite green.

## Decision

Proceed with integration coverage at the authority-validation request-hook boundary. Do not add a separate helper-level unit test or mock `urllib.parse`; the real hook can deterministically exercise the `ValueError` path with a real mitmproxy flow and verify all observable consequences.

## Chosen direction

Add one focused case to `test_request_handler_authority_validation.py` that:

- uses a malformed transparent request host that yields a non-numeric URL port;
- uses a distinct request `Host` value so the expected fallback host cannot be confused with the partially parsed host;
- invokes the real `mitm_addon.request()` and `mitm_addon.response()` hooks;
- asserts the 403 denial, firewall metadata, preserved diagnostic URL, atomic request host/port fallback, and persisted network-log entry.

This single boundary test fails if `ValueError` is re-raised or if only one member of the host/port pair falls back.

## Alternatives considered

- **Direct tests for `fallback_host_port()` / `set_target_from_url()`:** narrower but coupled to internal helpers and redundant once the real request/response boundary covers the branch and its security-relevant outcome.
- **Mocking `urllib.parse.urlparse` or parsed-result properties:** can manufacture every theoretical exception source, but tests Python internals and mock orchestration rather than a request the addon can receive.
- **Changing production parsing:** no production defect is currently demonstrated; this issue is specifically an unprotected fallback contract.

## Risks and mitigations

- The malformed host setup could become unrealistic if it used a mock flow. Use the shared real-flow fixture so mitmproxy's actual `host`, `pretty_host`, URL, metadata, and response semantics are exercised.
- No runtime code, protocol, schema, or deployment behavior changes, so production performance and compatibility are unaffected.

## Acceptance criteria

- A deterministic authority-validation integration test reaches the real `ValueError` fallback using a malformed URL with a non-numeric port.
- The denial still returns 403 and records the expected firewall decision and error.
- The original malformed diagnostic URL is preserved unchanged.
- Both network-log host and port come from the request fallback pair; the host is deliberately distinct from the partially parsed URL host.
- The response hook persists a network-log entry with the same URL, host, port, denial action, and status.
- The test fails if the `ValueError` handler re-raises or falls back only one field.

seven332:
## Summary

- exercise the malformed network-log target fallback through the real authority-validation request hook
- verify a missing-SNI denial still returns 403 and falls back host and port as one request-derived pair
- verify the response hook persists the unchanged malformed URL and denial target to the network log

## Scope

Test-only change. No production source, schema, API, protocol, dependency, deployment, or compatibility behavior changes.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_request_handler_authority_validation.py` (21 passed)
- `uv run --no-sync python -m pytest tests/` (5157 passed)
- mutation checks: re-raising `ValueError` and retaining the parsed host while only falling back the port both fail the new test

Closes #25895

### DIFF
--- a/crates/runner/mitm-addon/tests/test_request_handler_authority_validation.py
+++ b/crates/runner/mitm-addon/tests/test_request_handler_authority_validation.py
@@ -117,6 +117,52 @@ async def test_authority_validation_deny_response_logs_network_target(
     assert metadata_keys.HTTP_REQUEST_START_MONOTONIC not in flow.metadata
 
 
+async def test_authority_validation_deny_logs_malformed_fallback_target(
+    tmp_path, real_flow, mitm_ctx, fake_firewall_headers, headers
+):
+    reg_path = _write_github_firewall_registry(tmp_path)
+    fallback_url = "https://target.example:not-a-port/repos"
+    flow = real_flow(
+        with_response=False,
+        client_ip="10.200.0.5",
+        host="target.example:not-a-port",
+        sni="",
+        path="/repos",
+        request_headers=headers(("Host", "request.example")),
+    )
+
+    with (
+        mitm_ctx(registry_path=str(reg_path), api_url="https://api.vm0.ai"),
+        fake_firewall_headers() as auth_fetch,
+    ):
+        await mitm_addon.request(flow)
+
+    assert flow.response is not None
+    assert flow.response.status_code == 403
+    body = json.loads(flow.response.content)
+    assert body["error"] == "missing_sni"
+    auth_fetch.assert_not_called()
+    assert flow.metadata[metadata_keys.FIREWALL_ACTION] == "DENY"
+    assert flow.metadata[metadata_keys.FIREWALL_ERROR] == "missing_sni"
+    assert flow.metadata[metadata_keys.ORIGINAL_URL] == fallback_url
+    assert flow.metadata[metadata_keys.NETWORK_LOG_TARGET] == {
+        "url": fallback_url,
+        "host": "request.example",
+        "port": 443,
+    }
+
+    with mitm_ctx():
+        mitm_addon.response(flow)
+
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #25948

**Reviewed head:** `24569699c0e72a93fcbbdb10f6c861aeb4a49e2f`

### Summary

The new real-flow integration case reaches the malformed diagnostic-port fallback through the authority-validation request hook, proves host and port fall back atomically, and verifies the resulting 403 denial is persisted by the response hook. The PR adds no production fallback or compatibility behavior.

### Findings

No findings.

### Testing

Lock/sync validation, Ruff format/lint, BasedPyright, the 21-test authority-validation file, and all 5,157 mitm-addon tests passed. Both exception re-raise and port-only fallback mutations fail the new case as intended. Live mitmdump interception was not run; the repository's addon integration guidance uses real `HTTPFlow` hook coverage for this boundary.

### Verdict

LGTM

Issue #25352, opened 2026-08-06
chore(db): drop legacy custom connector revision state


seven332:
## Background

Parent: #25312

The preceding cleanup slices removed every executable dependency on Custom connector definition revisions, copied grant/OAuth revisions, and run credential snapshot references. The remaining database objects represent dead schema that should be removed rather than retained as permanent technical debt.

This is the final physical contraction slice for the parent objective.

## Dependency and production gate

The gate is satisfied:

- #25351 removed the executable revision and auth-reference behavior.
- #25952 stopped serving API inserts from naming the retired columns implicitly.
- #25952 shipped in `api-v1.411.2` and was successfully deployed to production at 2026-08-09 15:41 UTC.
- The five-minute predecessor request window elapsed before the contraction branch was rebased and published.

No additional rollback or stale-browser window is required; recovery after executable cleanup is roll-forward.

## Scope

- Drop `org_custom_connectors.revision` and its check constraint.
- Drop `user_custom_connectors.connector_revision`.
- Drop `connector_oauth_states.connector_revision`.
- Drop `agent_run_custom_connector_auth_refs` and its dependent indexes/constraints.
- Remove the corresponding current Drizzle declarations and aggregate export.
- Remove #25952's temporary insert targets and restore all affected writes to the contracted canonical tables.
- Preserve immutable historical migration files; add one new forward-only contraction migration.
- Verify no current source, generated artifact, test fixture, or deployment workflow references the removed objects.

## Deployment boundary

This migration is intentionally separate from executable cleanup. Both the newly promoted preparation API and its predecessor are independent of the removed schema. Recovery after contraction is roll-forward or explicit schema restoration, not deployment of an old binary.

## Acceptance criteria

- [x] Fresh database reset succeeds through the full immutable migration history.
- [x] Incremental migration from the production-compatible pre-contraction schema succeeds.
- [ ] Production migration smoke passes at release time.
- [x] PostgreSQL catalog inspection shows no dependent constraints, indexes, views, functions, or foreign keys left behind.
- [x] Current Drizzle schema and executable source contain none of the removed columns/table.
- [x] Relevant API integration tests pass against the contracted schema.
- [ ] #25312's final cleanup acceptance criteria are satisfied after this PR merges and the release gate passes.

seven332:
## Challenge Review: Framing

**Checkpoint:** `framing`
**Outcome:** `proceed`

[...]

Issue #25351, opened 2026-08-06
refactor(connectors): remove custom connector revision compatibility paths


seven332:
## Background

Parent: #25312

The current-state Custom connector producers and generalized Builtin/Custom consumers are now merged. The repository still contains the pre-cutover revision, credential-snapshot, refresh, and omission-compatibility paths alongside the final stable-ID lifecycle. Keeping both mechanisms makes executable behavior ambiguous, preserves stale credential snapshots, and blocks physical schema cleanup in #25352.

This slice removes all executable compatibility while intentionally retaining the physical legacy database columns and table for one additional API deployment and drain boundary.

## Evidence

- Run creation still writes five-hour `agent_run_custom_connector_auth_refs` snapshots, and `/firewall/auth` can still read them as a fallback beside current stable-ID auth resolution.
- Custom definition, grant, and OAuth code still reads or writes copied revisions even though current runtime identity is `customConnectorId` and credential compatibility is `storageVersion`.
- Both the legacy network-policy refresh protocol and the tagged connector-runtime synchronization protocol remain executable.
- Runner and API contracts still accept successful Custom runtime results without explicit routing inputs and matched firewall-auth requests without the final identity/routing context.
- `customConnectorResponse.storageVersion` remains optional even though the current API always returns it, leaving a Platform-only missing-response branch.
- The temporary version-`1` credential-parent transition trigger remains installed after current writers stopped depending on it.
- `chk_connector_oauth_states_custom_revision` still requires a copied revision for every Custom OAuth state, so current writers cannot stop producing the field until a forward migration removes that executable constraint.
- Whole-row Drizzle selections and returning projections can emit legacy columns without naming them at the business-logic call site.

## Decision

Remove the executable compatibility paths in one coherent cleanup PR, then remove the physical legacy schema in #25352 after this cleanup API has been promoted and its predecessor has drained.

No rollback window or stale-browser window is required. The service will not roll back to pre-cutover API or runner builds after cleanup promotion. Already-loaded old Platform code remains compatible with the new API because additional response fields are ignored and create/update request `storageVersion` remains optional as final server-inference behavior. The only required safety boundary is active old process drain.

Implementation may be prepared before that boundary. Merge and production deployment must wait until the current producers are active and all API and runner processes capable of emitting or consuming the legacy shapes have drained.

## Chosen direction

- Use one explicit current Custom connector database projection that excludes the legacy definition revision, and apply it to every current select and returning path.
[...]

Issue #25312, opened 2026-08-05
fix(connectors): align custom connector lifecycle with builtin connectors


seven332:
## Summary

Align the Custom connector lifecycle with Builtin connectors by reusing the same lifecycle concepts and shared machinery where their behavior is equivalent, without inventing a Custom-only revision, auth owner, credential digest, cache generation, or runtime admission model.

Stable `customConnectorId` identifies grants and active runtime targets. `storageVersion` has one narrow meaning: whether persisted member credentials can be interpreted by the connector's current storage contract. Source-specific authority remains separate where it is genuinely different: Builtin definitions come from the catalog, while Custom definitions, routing, permissions, and OAuth configuration come from organization-owned database state.

The functional delivery is complete. A post-merge source audit found two stale first-admission descriptions under #26835, so that child is temporarily reopened for one focused, non-behavioral cleanup PR. This parent remains open until that PR lands and the final source search is clean.

### Identity, credentials, and public status

- Agent grants use stable connector identity and do not copy definition or storage revisions.
- Manual and OAuth Custom connections reuse the shared `connectors` parent and a positive, monotonic, definition-owned `storageVersion`.
- `storageVersion` is only a persisted credential compatibility boundary. It is not a firewall version, authorization version, target version, or cache generation.
- Public connection status remains strict: stored `needsReconnect`, incompatible storage, missing required material, or an expired non-refreshable credential is reported as disconnected.
- Public status and runtime recoverability are intentionally distinct, matching Builtin behavior. A current, storage-compatible OAuth connection whose configured auth method supports refresh can remain runtime-recoverable even while public status requires reconnect.

### Future-run admission

- Agent grants select which Builtin or Custom connectors may be considered.
- A manual Custom connector enters a new run only with current compatible credentials and complete, validated routing inputs.
- A refresh-capable Custom OAuth connector may enter a new run when its definition and storage are current and routing is valid, even if stored state says reconnect is required. Run creation does not eagerly call the OAuth provider or decrypt token material; a matched `/firewall/auth` request performs locked live recovery.
- Missing refresh material and terminal provider rejection remain reconnect-required and fail closed. Transient provider failures remain upstream failures. Non-refreshable reconnect-required connections are omitted.
- Reconnection or credential repair affects later runs; it never dynamically expands a run that started without the target.
- Connector skills remain selected from the agent grant and definition, matching the existing Builtin skill boundary.

### Active pinned targets

[...]

PR #25953, opened 2026-08-09
refactor(db): isolate custom connector insert columns

### LINKED ISSUE #25952: refactor(db): stop implicit legacy custom connector column inserts
## Background

Parent: #25312

Blocks: #25352

#25351 removed explicit reads, writes, whole-row selections, and `RETURNING` projections for the retired Custom connector revision fields. During #25352 self-review, one remaining implicit dependency was found: Drizzle PostgreSQL `buildInsertQuery` enumerates every insertable column declared on the target table and emits `DEFAULT` for omitted values.

The current API therefore still names these columns in generated `INSERT` SQL even though no business-logic value refers to them:

- `org_custom_connectors.revision`
- `user_custom_connectors.connector_revision`
- `connector_oauth_states.connector_revision`

Production migrations run before new API traffic promotion. If #25352 drops the columns while the current API is still serving, Custom connector creation, grant updates, or OAuth starts can fail with PostgreSQL `42703` until the old API drains.

## Requirement

Establish a real predecessor-safe insert boundary before physical contraction. Every executable insert into the three affected tables must compile without the retired column names while the physical columns remain present.

Use temporary typed insert-only table projections outside Drizzle's migration schema. Each projection should use the same physical table name and declare only fields current insert call sites may write. Replace every production and test-route insert target for the affected tables with these projections. Keep canonical tables for selections, updates, joins, conflict targets, and explicit returning projections.

This PR must not drop or alter database objects. The temporary projections are removed by #25352 after this API is promoted and its predecessor drains; normal inserts can then target the contracted canonical schema again.

## Scope

- Add typed insert-only projections for `org_custom_connectors`, `user_custom_connectors`, and `connector_oauth_states` outside `packages/db/src/schema`, so Drizzle migration generation does not treat them as schema declarations.
- Export the transition module through an explicit `@vm0/db` package subpath.
- Replace all executable `.insert(...)` targets for the three tables, including fixture routes compiled into the API.
- Preserve existing values, conflict behavior, explicit returning selections, current connector behavior, and request latency.
- Verify generated SQL for each transition target does not name its retired column without keeping a permanent legacy negative fixture.
- Run current Custom connector, connector OAuth, grant, Feishu, GitHub, and cleanup integration coverage against the real database.

## Acceptance criteria

- [ ] Current API insert SQL for all three affected tables omits the retired revision column names.
- [ ] No raw SQL, runtime fallback, extra query, or database migration is introduced.
- [ ] Existing insert values, defaults, upsert conflict targets, and returned rows retain current behavior.
- [ ] Relevant API integration tests, DB/API type checks, lint, formatting, and Knip pass.
- [ ] The API is deployed and its predecessor drained before #25352 becomes mergeable.
- [ ] #25352 explicitly removes the temporary insert projections and restores canonical insert targets while dropping the physical objects.

## Deployment boundary

This is an additive code-only compatibility release. It is safe with the existing database and the currently deployed API. #25352 remains blocked until this release is active and its predecessor has drained.

seven332:
## Summary

- add temporary typed insert-only targets for the three Custom connector tables that still contain retired revision columns
- route every API insert through the current-column targets while preserving canonical tables for reads, updates, conflict targets, and returning selections
- keep the physical schema unchanged so this API can deploy before the final contraction in #25352

Drizzle's PostgreSQL insert builder enumerates every column declared by its target and emits `DEFAULT` for omitted values. The serving predecessor therefore still named the retired columns after #25351 even though no business logic used them. These targets make the generated insert column lists independent of those columns before migration-first deployment removes them.

The transition module is intentionally outside Drizzle's migration schema and has an explicit #25352 cleanup condition. #25352 will delete it and restore canonical insert targets after this release is promoted and drained.

## Validation

- one-time Drizzle `toSQL()` inspection for all three insert targets; no retired revision column appears
- representative `RETURNING` and bulk upsert/conflict SQL inspection
- `pnpm format`
- `pnpm -F @vm0/db lint`
- `pnpm -F @vm0/db test` — 29 tests passed
- `pnpm -F api lint`
- `pnpm -F api check-types`
- `pnpm knip --workspace @vm0/db --workspace api`
- targeted API integration suite — 7 files, 158 tests passed
- pre-commit full Knip and TypeScript checks
- `git diff --check`

No migration, raw SQL, additional query, fallback, or API contract change is included.

Closes #25952

Parent: #25312

Blocks: #25352

### DIFF
--- a/turbo/apps/api/src/signals/routes/test-cron-delete-cleanups-state.ts
+++ b/turbo/apps/api/src/signals/routes/test-cron-delete-cleanups-state.ts
@@ -1,4 +1,5 @@
 import { initContract } from "@vm0/api-contracts/contracts/trpc-contract";
+import { connectorOauthStatesInsertTarget } from "@vm0/db/custom-connector-insert-targets";
 import { connectorOauthStates } from "@vm0/db/schema/connector-oauth-state";
 import { telegramMessages } from "@vm0/db/schema/telegram-message";
 import { command } from "ccstate";
@@ -120,11 +121,11 @@ async function seedConnectorStates(
     offset += FIXTURE_INSERT_BATCH_SIZE
   ) {
     await db
-      .insert(connectorOauthStates)
+      .insert(connectorOauthStatesInsertTarget)
       .values(expiredStates.slice(offset, offset + FIXTURE_INSERT_BATCH_SIZE));
     signal.throwIfAborted();
   }
-  await db.insert(connectorOauthStates).values([
+  await db.insert(connectorOauthStatesInsertTarget).values([
     {
       state: connectorState(body.marker, "equal"),
       connectorSlug: "github",
--- a/turbo/apps/api/src/signals/routes/zero-connectors.ts
+++ b/turbo/apps/api/src/signals/routes/zero-connectors.ts
@@ -10,7 +10,7 @@ import {
   zeroConnectorsSearchContract,
 } from "@vm0/api-contracts/contracts/zero-connectors";
 import type { PublicConnectorCatalogDetail } from "@vm0/api-contracts/contracts/zero-connector-catalog";
-import { connectorOauthStates } from "@vm0/db/schema/connector-oauth-state";
+import { connectorOauthStatesInsertTarget } from "@vm0/db/custom-connector-insert-targets";
 
 import { authContext$, organizationAuthContext$ } from "../auth/auth-context";
 import { authRoute } from "../auth/auth-route";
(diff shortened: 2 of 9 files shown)


seven332:
## Self-review

Reviewed head `3f35ef3edb88dd38edf5a81a727f10d59ebde2d6` through scope/approach, correctness, tests/docs, concurrency/performance, resource/security, and maintainability passes.

No actionable findings remain.

- All nine affected insert expressions use the temporary typed targets; no canonical affected table remains as an insert target.
- Representative generated SQL confirms retired revision columns are absent, while canonical `RETURNING` and bulk upsert conflict targets remain valid.
- Physical tables, constraints, defaults, transactions, row counts, and request/API behavior are unchanged; the generated statements simply omit server-default columns from their insert column lists.
- There is no migration, raw SQL, extra query, fallback, secret handling change, or measurable normal-path performance cost.
- The transition module is outside migration schema and records the exact #25352 deployment/drain cleanup condition.
- DB/API lint and type checks, Knip, 29 DB tests, 158 targeted real-database API tests, full pre-commit Knip/type checks, and diff checks passed.

PR #25962, opened 2026-08-09
fix(connectors): align custom run admission

### LINKED ISSUE #25957: fix(connectors): align custom run admission with builtin credential availability
## Background

Parent: #25312

Builtin and Custom connectors currently use different admission rules when a new run is created.

- Builtin firewalls and runtime targets are included only when the member has a current compatible connection whose runtime credential status is usable.
- Custom target membership is selected from the agent grant and definition even when the member is disconnected, must reconnect, or has storage that the current definition cannot interpret.

The Custom behavior enables an initially unavailable target to recover inside the same run, but no current product requirement depends on that capability. Extending Builtin registration, synchronization, routing pinning, mutation wake-ups, and auth ownership solely to reproduce it would add a second recovery path and deployment surface without a demonstrated need.

## Goal

Align Custom **new-run admission** with the existing Builtin behavior:

- A grant determines which connector may be considered for the run.
- A current compatible and usable member connection determines whether its firewall and runtime target are admitted into that new run.
- Credential or routing unavailability excludes the Custom firewall and target from that run rather than persisting an unpinned target for later first admission.
- A later reconnect makes the connector available to later runs; it does not expand a run that has already started.

This changes future-run selection only. It does not make credential availability tear down a valid target already owned by an active run.

## Required behavior

| Connector state at run creation | New run | Already-active pinned run |
| --- | --- | --- |
| Current compatible credentials and complete routing inputs | Include firewall, policy, and target | Continue normal synchronization and auth |
| Disconnected or reconnect required | Omit firewall and target | Keep the validated firewall/policy; auth fails until restored |
| Storage incompatible | Omit firewall and target without reading stale values | Keep last-known-good executable state; current auth remains unavailable |
| Reconnected after the run started | Available to later runs | Do not dynamically add a target that was absent at run creation |

Connector skills remain selected from the agent grant and definition, matching the existing Builtin skill-selection boundary. A missing runtime target does not remove an otherwise selected connector skill from the run.

## Scope

- Filter Custom rows at the run-creation boundary before they contribute firewall, permission policy, reserved auth aliases, or `connectorRuntimeTargets`.
- Require current storage compatibility, usable authentication state, all required current values, and a completely validated route for new-run admission.
- Keep skill selection independent from that runtime admission filter.
- Preserve the shared Custom runtime builder's active-target behavior. Runtime synchronization must still rebuild an already-pinned target from its pinned routing inputs even when current credentials are unavailable.
- Preserve current `/firewall/auth` behavior for active targets: current values are resolved on cache miss, and unavailable credentials fail auth without removing the firewall.
- Remove or update tests that expect an unavailable Custom target to be registered for first admission inside the same run.

## Implementation boundary

Apply the admission filter in the run-creation caller rather than globally inside the shared Custom runtime builder. The same builder is used by active-run synchronization, where filtering on current credential availability would incorrectly remove last-known-good firewall state.

No database migration, Builtin protocol extension, new target state, fixed polling, or new runtime wake-up mechanism is required.

## Acceptance criteria

- [ ] A granted Custom connector with current compatible credentials and valid routing inputs is admitted exactly as today.
- [ ] A disconnected, reconnect-required, storage-incompatible, incomplete, or unroutable Custom connector contributes no firewall, permission policy, auth aliases, or runtime target to a newly created run.
- [ ] The connector's granted skill remains selected under the same rules as a Builtin connector skill.
- [ ] Reconnecting after run creation does not expand that run's target membership; a later run admits the connector.
- [ ] Credential removal or incompatibility after a target is active does not remove its validated firewall or network policy; auth fails through the existing contract.
- [ ] Definition, permission, and pinned-routing synchronization for active Custom targets is unchanged.
- [ ] Builtin connectors, model providers, platform firewalls, and non-connector firewalls are unchanged.
- [ ] Integration coverage exercises disconnected, reconnect-required, storage-incompatible, later-reconnect, and active-run credential-loss cases through run create/claim and runtime-auth entry points.

## Out of scope

- Adding same-run first admission after a connector is reconnected.
- Changing optional auth-entry semantics; tracked by #25958.
- Changing disconnect ownership or upstream revocation; tracked by #25959.
- Changing connector skills, storage versions, permission storage, or active-run last-known-good synchronization.

---
> Compacted on 2026-08-10 from 0 comments.

seven332:
## Summary

- admit Custom connector executable state into a new run only when the member connection is current, credential-complete, and produces a validated routed firewall
- keep granted Custom connector skills independent from executable admission
- preserve the shared active-target builder so existing runs retain pinned firewall and network policy when credentials become unavailable
- replace same-run first-admission expectations with API integration coverage for disconnected, reconnect-required, storage-incompatible, incomplete, unroutable, and later-run recovery states

## Exclusions

- no database migration or persisted schema change
- no API or runner protocol change
- no change to Builtin connectors, optional auth-entry semantics, disconnect ownership, or active-run last-known-good synchronization

## Validation

- `pnpm -F api exec vitest run src/signals/routes/__tests__/run-lifecycle.bdd.test.ts` (147 passed)
- `pnpm -F api lint`
- `pnpm -F api check-types`
- `pnpm knip --workspace api`
- pre-commit: Prettier, full-repository Knip, full-repository type checks, file-size check, and commitlint

Closes #25957

Parent context: #25312

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
@@ -9601,11 +9601,23 @@ describe("RUN-02: custom connectors, grants, and network policies", () => {
         customConnectorId: saved.connector.id,
       }),
     );
-    await api.requestCancelRun(actor, incompleteRun.runId, [200]);
 
+    context.mocks.ably.publish.mockClear();
     await connectors.setCustomConnectorValues(actor, saved.connector.id, [
       { key: "api_key", kind: "secret", value: "recovered-key" },
     ]);
+    expect(context.mocks.ably.publish).not.toHaveBeenCalledWith(
+      "connector-runtime-sync",
+      {
+        runId: incompleteRun.runId,
+        target: {
+          kind: "custom",
+          customConnectorId: saved.connector.id,
+        },
+      },
+    );
+    await api.requestCancelRun(actor, incompleteRun.runId, [200]);
+
     const longSubdomain = "a".repeat(55);
     await connectors.setCustomConnectorValues(actor, saved.connector.id, [
       { key: "subdomain", kind: "variable", value: longSubdomain },
--- a/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
@@ -8571,7 +8571,7 @@ describe("RUN-02: custom connectors, grants, and network policies", () => {
     expect(cancelled.status).toBe("cancelled");
   });
 
-  it("enforces custom connector permission grants and mounts its generated skill", async () => {
+  it("keeps a granted custom skill independent from runtime admission", async () => {
     const api = createRunsApi(context);
     createBddApi(context).acceptAgentStorageWrites();
     const connectors = createConnectorBddApi(context);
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #25962

**Reviewed head:** `ec21a817da9460b5e42dcda90f011f1f1f10d328`

### Summary

New-run Custom connector executable admission now requires a current, credential-complete connection and a validated routed firewall, while grant-selected skills and active-run last-known-good synchronization retain their existing boundaries. The Builtin path and serialized contracts are unchanged.

### Findings

No findings.

### Testing

The API lifecycle integration suite passes all 147 tests and covers configured, disconnected, reconnect-required, storage-incompatible, incomplete, unroutable, later-run recovery, active-run credential loss, skill independence, and no same-run wake-up after reconnect. API lint, API type checks, API-scoped Knip, and repository pre-commit checks also pass. The full GitHub CI matrix is still in progress.

### Verdict

LGTM

PR #25990, opened 2026-08-10
test(python): cover keyword-only metadata alias defaults

### LINKED ISSUE #25977: test(python): cover keyword-only metadata alias defaults
## Problem

`_MetadataKeyVisitor._metadata_default_argument_names()` tracks metadata aliases introduced by both positional and keyword-only defaults, but the linter fixtures only exercise positional defaults (`violations.base.py.txt:181-183`). There is no regression test for keyword-only function or lambda defaults.

At `54b7a79bb7acec12b28fcc74a9ca17ca30ad1a42`, the current implementation correctly reports both accesses in this input:

```python
def function_case(*, metadata=flow.metadata):
    return metadata["vm_run_id"]

lambda_case = lambda *, metadata=flow.metadata: metadata["firewall_name"]
```

However, a mutation that skips the keyword-only loop at `crates/runner/mitm-addon/scripts/flow_metadata_linter/visitor.py:320-322` leaves all 18 tests in `tests/test_flow_metadata_key_contract.py` passing, while the input above drops from two violations to zero.

This allows a regression in keyword-only default alias propagation to silently weaken the repository metadata-key guard.

## Acceptance criteria

- Add function and lambda fixture cases whose keyword-only defaults alias `flow.metadata`.
- Assert that registered string keys used through those aliases are reported with the corresponding `metadata_keys` constants.
- Ensure the test fails when keyword-only default alias propagation is disabled while the existing positional-default cases remain green.

seven332:
## Summary

- add focused function and lambda fixtures for keyword-only `flow.metadata` default aliases
- assert the exact registered metadata-key diagnostics through the public linter entry point
- isolate this contract from the version-specific monolithic violation fixtures

## Explicit exclusions

- no production visitor or runtime behavior changes
- no API, schema, migration, compatibility, or rollout changes

## Validation

- focused test passes with the production visitor
- mutation check: the existing positional-default test passes while the new keyword-only test fails when keyword-only propagation is disabled
- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (`5182 passed`)

Closes #25977

### DIFF
--- a/crates/runner/mitm-addon/tests/fixtures/flow_metadata_key_linter/keyword_only_default_aliases.base.py.txt
+++ b/crates/runner/mitm-addon/tests/fixtures/flow_metadata_key_linter/keyword_only_default_aliases.base.py.txt
@@ -1,0 +1,4 @@
+def function_case(*, metadata=flow.metadata):
+    return metadata["vm_run_id"]
+
+lambda_case = lambda *, metadata=flow.metadata: metadata["firewall_name"]
--- a/crates/runner/mitm-addon/tests/fixtures/flow_metadata_key_linter/keyword_only_default_aliases.expected.txt
+++ b/crates/runner/mitm-addon/tests/fixtures/flow_metadata_key_linter/keyword_only_default_aliases.expected.txt
@@ -1,0 +1,2 @@
+keyword_only_default_aliases.py:2: use metadata_keys.VM_RUN_ID for flow.metadata access
+keyword_only_default_aliases.py:4: use metadata_keys.FIREWALL_NAME for flow.metadata access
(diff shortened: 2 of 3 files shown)


seven332:
## Code Review: PR #25990

**Reviewed head:** `e28b43f7b39119c056816446733c7337cd58a171`

### Summary

This PR adds focused public-entry-point regression coverage for function and lambda keyword-only `flow.metadata` default aliases. The dedicated fixture pair matches existing test structure and keeps production code unchanged.

### Findings

No findings.

### Testing

The focused test passes on the submitted implementation. A temporary mutation that removed keyword-only propagation left the existing positional test green and made the new test fail with both diagnostics missing. Ruff format/lint, BasedPyright, lock validation, environment sync, and the complete add-on suite passed locally (`5182 passed`). Remote CI was still pending at review time.

### Verdict

LGTM

PR #25992, opened 2026-08-10
test(runner): make workspace lock retry coverage deterministic

### LINKED ISSUE #25975: test(rust): make transient workspace cache lock coverage deterministic
## Background

PR #25812 added a bounded wait in `WorkspaceImageCache::prepare` so a short-lived entry-lock holder does not force a fresh workspace. Its source issue, #25810, explicitly required deterministic coverage of the short-contention path.

## Evidence

At `crates/runner/src/executor/tests/sandbox_run_tests/workspace_cache.rs:803-890`, `execute_inner_waits_for_transient_workspace_cache_lock`:

- acquires the seeded cache entry lock at lines 826-831;
- starts an independent OS thread that sleeps for 20 ms and releases the lock at lines 832-835;
- only then begins `execute_new_sandbox` at lines 838-855; and
- asserts a normal cache hit and successful sandbox lifecycle at lines 858-889.

There is no rendezvous proving that `WorkspaceImageCache::prepare` reached the lock acquisition at `crates/runner/src/workspace_image_cache/lifecycle.rs:398-439` while the test still held the lock. If executor setup or task scheduling takes more than 20 ms, the release thread drops the guard first. The test then follows the ordinary uncontended cache-hit path and all current assertions still pass.

The history makes the blind spot concrete: commit `87a3ece7` renamed the existing uncontended `execute_inner_uses_workspace_cache_when_configured` test and added the timed holder, while retaining the same cache-hit assertions. A pre-released lock is therefore observationally equivalent to the test's original scenario.

## Impact

The test can pass without exercising the retry loop introduced by PR #25812. A regression that removes or bypasses the bounded wait could remain green depending on scheduling, leaving the production transient-contention behavior without reliable regression coverage.

## Acceptance criteria

- Replace the 20 ms timing assumption with an explicit, bounded rendezvous that proves the prepare path has attempted the entry lock and observed contention before the holder is released.
- After releasing the holder, assert that the seeded cache image is consumed and the successful prepare/cache-hit telemetry is emitted.
- Keep a bounded watchdog so a broken wait path fails promptly instead of hanging the suite.
- Ensure the test fails if workspace-cache preparation no longer waits/retries after the initial busy acquisition.

## Relevant code

- `crates/runner/src/executor/tests/sandbox_run_tests/workspace_cache.rs:803-890`
- `crates/runner/src/workspace_image_cache/lifecycle.rs:398-439`
- `crates/runner/src/lock.rs:103-119`

seven332:
## Summary

- replace the fixed 20 ms workspace-cache lock release with a two-attempt real-filesystem rendezvous
- release the held entry lock only after preparation has retried the unchanged lock path
- retain the seeded move-image and successful prepare/cache-hit telemetry assertions

## Scope

- test-only runner change; production lock and workspace-cache code is unchanged
- no API, protocol, persisted-state, dependency, production-timeout, or deployment-compatibility change

## Validation

- `cargo test -p runner execute_inner_waits_for_transient_workspace_cache_lock -- --nocapture`
- `cargo test -p runner`
- focused test repeated 50 times
- temporary one-shot acquisition mutation: focused test failed on the missing second attempt; source restored and the test passed again
- applicable Lefthook pre-commit and commit-message checks

Closes #25975

### DIFF
--- a/crates/runner/src/executor/tests/sandbox_run_tests/workspace_cache.rs
+++ b/crates/runner/src/executor/tests/sandbox_run_tests/workspace_cache.rs
@@ -2,6 +2,9 @@ use super::*;
 use async_trait::async_trait;
 use httpmock::prelude::*;
 use sha2::{Digest, Sha256};
+use std::os::unix::fs::PermissionsExt;
+use std::sync::mpsc;
+use std::time::Instant;
 
 use crate::executor::session_history_cpu::{SessionHistoryCpuPool, SessionHistoryCpuTestGate};
 use crate::executor::tests::support::RUN_IN_SANDBOX_TEST_TIMEOUT;
@@ -801,6 +804,8 @@ async fn execute_inner_cache_hit_retry_requires_completed_cleanup() {
 
 #[tokio::test]
 async fn execute_inner_waits_for_transient_workspace_cache_lock() {
+    const LOCK_ATTEMPT_WATCHDOG: Duration = Duration::from_secs(1);
+
     let dir = tempfile::tempdir().unwrap();
     let runner_paths = RunnerPaths::new(dir.path().join("runner"));
     let cache = WorkspaceImageCache::new(runner_paths.clone());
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #25992

**Reviewed head:** `669d19910718e2540f0ace3ca3848e4d7d50aa27`

### Summary

The regression now uses a real lock and a deterministic two-attempt filesystem rendezvous instead of an arbitrary sleep. The change remains test-only and preserves the end-to-end workspace-cache and telemetry assertions.

### Findings

No findings.

### Testing

Runner formatting, Clippy, docs, the focused regression, related workspace-cache suites, the full runner suite, a 50-run repetition, and a one-shot acquisition mutation check passed. Remote CI is checked separately.

### Verdict

LGTM

PR #25993, opened 2026-08-10
test(runner): make jsonl flush serialization deterministic

### LINKED ISSUE #25976: test(rust): make JSONL flush serialization tests deterministic
## Background

`MitmJsonlFlushHandle::flush_path` serializes requests with a shared async mutex because every caller publishes through the same `jsonl-flush-request` marker. The tests intend to prove that a second request cannot replace the first marker while the first acknowledgement is pending.

## Evidence

The serialization guard is acquired at `crates/runner/src/proxy/flush.rs:232-244`.

Both serialization tests use the same scheduler-dependent pattern:

- `jsonl_flush_handle_serializes_concurrent_requests` spawns the second request, calls `tokio::task::yield_now()` once, and checks that the marker still names the first path at lines 1138-1149.
- `jsonl_flush_handles_from_same_proxy_serialize_concurrent_requests` repeats the pattern for two handles from one proxy at lines 1206-1216.

A single `yield_now()` does not establish that the second task was polled or reached `request_lock.lock().await`. If the current test task is polled again first, the marker naturally still names the first path. The test then acknowledges and awaits the first request; the previously unpolled second task can run afterward, publish its marker, receive its acknowledgement, and satisfy every remaining assertion. This sequence would also pass if the shared request lock were removed.

The same weakness existed before PR #23643: that change replaced a post-yield `try_recv` check with the current post-yield marker check, but neither form synchronizes with the second task reaching the serialization boundary.

## Impact

These tests do not reliably protect the shared-marker serialization invariant. A regression that permits two concurrent flushes to overwrite `jsonl-flush-request` can remain green, while production callers may wait on mismatched request IDs and time out before accepted JSONL writes are visible for upload.

## Acceptance criteria

- Add an explicit, bounded rendezvous proving the second `flush_path` call has started and reached the serialization boundary while the first request remains pending.
- Assert that the second request marker cannot be published before the first acknowledgement completes, then assert it is published and acknowledged afterward.
- Cover both cloned handles and distinct handles returned by the same `MitmProxy`, without relying on `yield_now()` or elapsed sleeps as proof of task progress.
- Ensure the coverage fails if `MitmJsonlFlushHandle` instances no longer share the request lock.

## Relevant code

- `crates/runner/src/proxy/flush.rs:232-244`
- `crates/runner/src/proxy/flush.rs:1117-1185`
- `crates/runner/src/proxy/flush.rs:1187-1252`
- `crates/runner/src/proxy/process.rs:194-200`

seven332:
## Summary

- observe the second JSONL flush request's first mutex poll through test-only instrumentation
- replace scheduler-dependent `yield_now()` assertions with an explicit bounded `Poll::Pending` rendezvous
- preserve real marker publication and acknowledgement coverage for cloned handles and separate handles from one proxy

## Compatibility

- no production-build observer state or runtime behavior changes
- no JSONL protocol, persisted state, API, database, migration, or deployment changes

## Validation

- `cargo test -p runner --bin runner jsonl_flush_handle_serializes_concurrent_requests -- --nocapture`
- `cargo test -p runner --bin runner jsonl_flush_handles_from_same_proxy_serialize_concurrent_requests -- --nocapture`
- mutation: give the observed second handle a fresh lock; the focused test fails with `Ready` instead of `Pending`
- mutation: bypass the observed lock acquisition; the focused test fails with bounded `request lock was not polled`
- `cargo test -p runner --bin runner` (3075 passed, 20 ignored)
- `lefthook run pre-commit`

Closes #25976

### DIFF
--- a/crates/runner/src/proxy/flush.rs
+++ b/crates/runner/src/proxy/flush.rs
@@ -3,9 +3,13 @@
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex};
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
+#[cfg(test)]
+use std::{future::Future as _, task::Poll};
 
 use serde::{Deserialize, Serialize};
 use tokio::sync::Mutex as AsyncMutex;
+#[cfg(test)]
+use tokio::sync::mpsc;
 use tracing::{error, warn};
 use uuid::Uuid;
 
@@ -176,6 +180,15 @@ pub struct MitmJsonlFlushHandle {
     pub(super) addon_dir: PathBuf,
     pub(super) usage_state: Arc<Mutex<UsageFlushTarget>>,
     pub(super) request_lock: Arc<AsyncMutex<()>>,
+    #[cfg(test)]
+    pub(super) request_lock_poll_tx: Option<mpsc::UnboundedSender<RequestLockPoll>>,
+}
+
+#[cfg(test)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(super) enum RequestLockPoll {
+    Pending,
+    Ready,
 }
 
 pub(super) fn now_millis() -> u64 {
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #25993

**Reviewed head:** `44b25eae5cc9e551c1c82ec9019f27c013b3f061`

### Summary

The test-only mutex-poll observer establishes that the second JSONL flush actually reaches and blocks at the shared serialization boundary before the first acknowledgement. Both handle construction paths retain their real marker and acknowledgement assertions, with no protocol or persisted-state changes.

### Findings

No findings.

### Testing

Validated runner formatting, Clippy, rustdoc, the complete runner binary test suite, repository pre-commit hooks, and focused JSONL flush tests. Lock-replacement and lock-bypass mutations both make the regression coverage fail as intended. Unrelated Turbo and Python suites were not run because this change is isolated to the Rust runner crate.

### Verdict

LGTM

PR #25995, opened 2026-08-10
fix(ui): put every icon on one stroke width and match paired icon colour

tongx-blip:
Three defects visible in the sidebar, all measured off the reported screenshots.

## 1. Icons render at eleven different stroke widths

`globals.css` normalises icons with

```css
svg[class*="lucide"][stroke-width="2"]:not([data-stroke]) { stroke-width: var(--icon-stroke-width); }
```

Only icons whose rendered attribute is exactly `2` reach the token. Every call site passing something else escapes it, and `data-stroke` opts 59 more out:

| rendered width | icons |
|---|---|
| 1.5 (the token) | 201 |
| 2.0 | 59 |
| 1.8 | 74 |
| 1.7 | 22 |
| 2.5 | 9 |
| 1.75 | 8 |
| 1.6 | 6 |
| 2.6 / 1.9 | 3 each |
| 2.2 / 1.0 | 2 each |
| 1.4 | 1 |

**188 icons render at something other than the token.** The Pinned `+` is `<Plus size={16} strokeWidth={2} data-stroke />` — it opts out and renders at 2 beside nav icons at 1.5, a third heavier. That is the thickness difference in the screenshots.

Fix: drop all 379 per-call-site `strokeWidth` props, the `ICON_STROKE` constant, the checkbox tick's hard-coded `2.5`, and the 59 `data-stroke` opt-outs, so every icon takes the token. The two hand-inlined `<svg>` icons get the `lucide` class so the rule reaches them. The CSS mechanism is left in place — the width is now genuinely settable in one line.

## 2. Icons paired with text are lighter than the text

```css
```

Every sidebar icon rendered 30% transparent while its label rendered at full strength. On the light theme that resolves to `rgb(89,92,96)` against a `rgb(21,24,30)` label.

Removed, and the neutral colour utility dropped from the **57** icons that have a text sibling — settings as well as the sidebar — so each inherits from the same element its label does. Where the row is muted both stay muted; where it isn't, both are full strength.

**Icon-only controls are deliberately untouched.** Each now carries the product it was already rendering at (`ancestor alpha x 0.7`) as its own element opacity, so collapse / `...` / pin look exactly as before.

## 3. Overlay

A translucent *colour* makes overlapping strokes composite against each other, so joints render darker than the stroke body. Element *opacity* flattens the icon first. Same shade, measured:

| | body | crossing | ratio |
|---|---|---|---|
| `rgba` colour @ 0.49 | rgb(137,138,142) | rgb(80,81,86) | 1.51x |
| opaque + `opacity: .49` | rgb(136,138,141) | rgb(136,138,141) | **1.00x** |

Where the alpha sat on an icon-only ancestor, the colour is now opaque. The Pinned "new" control holds both an icon and a label under one alpha, so the control is dimmed as a whole instead — which also keeps its icon and label the same colour.

## Not in this PR

The `Pin` glyph changed shape in the Lucide migration (Tabler's was a 45° pushpin; Lucide's is a head-on thumbtack). Lucide ships only `pin` and `pin-off`, neither diagonal. Raised separately for a design decision.

## Verification

- `pnpm turbo run lint` — 0 errors, `@vm0/app` warnings unchanged from `main` (53)
- `pnpm check-types` — passes for `@vm0/ui`, `@vm0/app`, `@vm0/desktop`
- `pnpm -F @vm0/app build` — succeeds
- `@vm0/ui` (59) and the CSS-entrypoint / composer tests pass

### DIFF
--- a/turbo/apps/platform/src/views/activity-page/activity-inspect-page.tsx
+++ b/turbo/apps/platform/src/views/activity-page/activity-inspect-page.tsx
@@ -84,7 +84,7 @@ function InspectBreadcrumb({ title }: { title: string }) {
   return (
     <nav className="hidden md:flex shrink-0 items-center gap-1 px-4 pt-4 text-sm text-muted-foreground">
       <span className="inline-flex items-center gap-1 rounded-md px-1.5 py-0.5">
-        <ChartLine size={14} strokeWidth={1.5} className="shrink-0" />
+        <ChartLine size={14} className="shrink-0" />
         {t(($) => {
           return $.activity.detail.activity;
         })}
@@ -111,11 +111,7 @@ function InspectEmptyState() {
         })}
       />
       <div className="flex-1 flex flex-col items-center justify-center gap-3 pb-20">
-        <Upload
-          size={48}
-          strokeWidth={1}
-          className="text-muted-foreground/40"
-        />
+        <Upload size={48} className="" />
         <h2 className="text-lg font-semibold text-foreground">
           {t(($) => {
             return $.activity.inspect.noLog.title;
@@ -133,7 +129,7 @@ function InspectEmptyState() {
         )}
         <Button variant="outline" asChild>
           <label className="cursor-pointer">
-            <Upload size={16} strokeWidth={1.5} />
+            <Upload size={16} />
             {t(($) => {
               return $.activity.inspect.noLog.upload;
             })}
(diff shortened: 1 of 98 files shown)


tongx-blip:
Verified on the deployed preview.

```
stroke widths on the page : { "1.5px": 34 }     <- was eleven different values
translucent-colour icons  : 0
```

Icon and label colour on every pair, including the muted rows:

```
My Organization / Agents / Workflows / Connectors / Artifacts
    icon = rgb(21,24,30)   label = rgb(21,24,30)    same
DeepSeek V4
    icon = rgb(81,90,103)  label = rgb(81,90,103)   same
```

One pair reports `same: false` — a hover-reveal icon sitting at `text-muted-foreground/0`, i.e. deliberately invisible until hover. That transition state is intentionally preserved.

PR #25996, opened 2026-08-10
fix(connectors): align custom firewall auth with builtin semantics

### LINKED ISSUE #25958: fix(connectors): unify missing optional firewall auth semantics
## Background

Parent: #25312

Builtin and Custom connector firewall auth currently interpret a missing optional stored value differently.

- Custom auth removes every matched header or query injection whose template references a missing optional secret, then allows the request to continue.
- Builtin auth treats any secret referenced by the matched API auth template as required for resolving that request and fails auth when it is unavailable.

The Custom behavior can drop an entire auth entry when a template mixes required and optional inputs. That can forward an unauthenticated request where the equivalent Builtin connector fails closed. The current connector catalog has a concrete shape of this problem: Kudosity's API secret is optional for V2 but is referenced by the V1 Basic-auth entry.

## Goal

Use one fail-closed rule for matched connector APIs:

- A field may be optional for configuring or using the connector as a whole.
- Once the matched API auth template references that field, its value is required to resolve that specific request.
- Missing optional values that are not referenced by the matched API do not affect the request.

Do not infer that an auth injection itself is optional merely because one referenced connector field is optional.

## Scope

- Remove the Custom-only behavior that silently omits auth headers or query entries based on missing optional fields.
- Resolve Custom and Builtin matched auth templates with the same missing-reference semantics.
- Reuse the shared referenced-key collection, missing-reference validation, template resolution, and connector-not-configured response wherever the source-specific credential loaders permit; do not add a second Custom-only implementation of those rules.
- Return the existing connector-not-configured/auth-resolution failure rather than forwarding a partially authenticated request.
- Keep optional connector fields valid for API entries that do not reference them.

## Acceptance criteria

- Builtin and Custom connectors return the same failure class when the matched API auth template references an unavailable optional credential.
- A mixed template that references both required and optional credentials is never silently removed and forwarded without auth.
- A connector with an absent optional value can still use matched API entries that do not reference that value.
- Shared firewall-auth reference collection, missing-reference validation, template resolution, and error construction serve both Builtin and Custom paths; no Custom-only optional-entry filter remains.
- Existing required-field, OAuth refresh, reconnect-required, routing-variable, `auth.base`, query-auth, and SigV4 behavior remains unchanged.
- API-route integration tests cover header, query, mixed-template, and unrelated-API cases at the firewall-auth boundary.

## Out of scope

- Adding an `optional` flag to firewall header/query injections.
- Introducing conditional auth-template syntax.
- Changing connector form validation or credential storage versions.
- Changing connector runtime admission or disconnect cleanup.
- Introducing a Custom-only auth cache invalidation or expiry mechanism.

If product requirements later need an explicitly optional auth injection, that should be designed as a separate schema capability rather than inferred from field requiredness.

seven332:
## Summary

- make Custom connector firewall auth fail closed when any referenced credential is unavailable, matching Builtin connector behavior
- reuse the shared reference collection, missing-value validation, template resolution, and not-configured response path
- remove the Custom-only optional-template omission path while preserving active firewall policy and current credential refresh behavior

## Scope

- no database, API schema, runner protocol, runtime admission, or cache lifecycle changes
- Builtin connector behavior is unchanged

## Validation

- `pnpm exec prettier --check apps/api/src/signals/services/agent-webhook-firewall-auth.service.ts apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts`
- `pnpm -F api exec vitest run src/signals/routes/__tests__/run-lifecycle.bdd.test.ts` (147 tests)
- `pnpm -F api lint`
- `pnpm -F api check-types`
- `pnpm knip --workspace api`
- pre-commit hooks: full Turbo Knip and type checks

Closes #25958

Parent: #25312

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
@@ -9357,7 +9357,7 @@ describe("RUN-02: custom connectors, grants, and network policies", () => {
     expect(cancelled.status).toBe("cancelled");
   });
 
-  it("preserves runtime auth templates while omitting missing optional values", async () => {
+  it("fails closed when a custom auth header references an optional missing value", async () => {
     const api = createRunsApi(context);
     const connectors = createConnectorBddApi(context);
     const fw = createFirewallApi(context);
@@ -9388,15 +9388,18 @@ describe("RUN-02: custom connectors, grants, and network policies", () => {
             kind: "variable",
             required: false,
           },
+          {
+            key: "unused_note",
+            label: "Unused note",
+            kind: "variable",
+            required: false,
+          },
         ],
         headerInjections: [
           {
             name: "Authorization",
-            valueTemplate: "Bearer {{secrets.api_key}}",
-          },
-          {
-            name: "X-Secondary",
-            valueTemplate: "{{secrets.secondary_token}}",
+            valueTemplate:
+              "Bearer {{secrets.api_key}}:{{secrets.secondary_token}}",
           },
         ],
         queryInjections: [
@@ -9406,7 +9409,10 @@ describe("RUN-02: custom connectors, grants, and network policies", () => {
           },
         ],
       },
-      values: [{ key: "api_key", kind: "secret", value: "optional-primary" }],
+      values: [
+        { key: "api_key", kind: "secret", value: "optional-primary" },
+        { key: "tenant_id", kind: "variable", value: "initial-tenant" },
+      ],
       agentId,
     });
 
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #25996

**Reviewed head:** `53a16c3b33c8e4a73f9dd92b017d1396b218cf85`

### Summary

Custom connector firewall auth now fails closed for every unavailable credential referenced by the matched auth templates, using the existing shared reference collection, completeness check, template resolver, SigV4 guard, and not-configured response. Source-specific credential loading remains separate, and no API, runner, database, or cache contract changes are introduced.

### Findings

No findings.

### Testing

The public API lifecycle suite covers referenced header and query failures, a mixed template, an unreferenced optional field, active-firewall retention, and restoration on the same runtime payload. Focused Prettier, 147 API integration tests, API lint, API type checks, API Knip, and the repository pre-commit Knip/type checks passed. Rust and Python checks are unrelated to this TypeScript-only API change.

### Verdict

LGTM

PR #26004, opened 2026-08-10
fix(python): lint annotation-only metadata subscripts

### LINKED ISSUE #25979: fix(python): lint annotation-only metadata subscript targets
## Problem and evidence

Annotation-only metadata subscript targets bypass the registered metadata-key linter. `_visit_annotation_target()` traverses a subscript's container and slice separately, so it never runs the key validation owned by `visit_Subscript()`.

Current `main` reproduces the defect for both forms below: `metadata_key_violations()` returns no diagnostics, while ordinary subscript access reports the same registered literals.

```python
flow.metadata["vm_run_id"]: str

metadata = flow.metadata
metadata["firewall_name"]: str
```

The existing metadata-key contract fixture suite passes without covering either form. Related #25978 exercises a different value-bearing annotated-assignment branch.

## Decision

Proceed with one PR. Treat this as an implementation defect, not only a test gap.

## Chosen direction

Extract the metadata-subscript key diagnostic from `visit_Subscript()` into a focused private helper. Invoke it from both ordinary subscript handling and the annotation-only subscript branch, before their existing child traversal.

Keep the traversal paths separate. Python evaluates the container and non-constant key expressions for an annotation-only subscript target but does not execute `__getitem__`; the annotation-only path must therefore continue to visit only its children and must not model the outer subscription operation as an implicit exception. Deferred annotation handling and alias state remain unchanged.

## Alternatives considered

- Duplicating the key check in `_visit_annotation_target()` preserves behavior but creates two owners for the same registered-key policy.
- Calling `visit_Subscript()` directly works with the current implementation but couples annotation-only semantics to future ordinary-subscript behavior.
- Leaving the status quo keeps a reproduced direct and aliased contract bypass.

## Risks and mitigations

- **Alias-state timing:** call the helper before child traversal in both paths, matching existing ordinary-subscript behavior; cover direct and pre-bound alias forms.
- **Incorrect exception-flow modeling:** retain the explicit annotation-target traversal and never route the outer target through `visit()`.
- **Duplicate or reordered diagnostics:** use the existing diagnostic collection and assert the complete focused output.
- **Compatibility:** the current repository scan contains no affected violations. There is no schema, API, protocol, persisted-state, dependency, migration, feature-switch, or rollout boundary.

## Acceptance criteria

- Annotation-only `flow.metadata[...]` targets enforce the ordinary registered-key validation.
- Annotation-only targets using a tracked metadata alias enforce the same validation.
- A dedicated public-entry-point fixture reports `metadata_keys.VM_RUN_ID` and `metadata_keys.FIREWALL_NAME` at the expected lines.
- Existing ordinary subscript, annotation evaluation, alias tracking, deferred annotation, and implicit-exception behavior remain passing.
- The add-on lock check, Ruff format and lint, BasedPyright, focused contract test, and full pytest suite pass.

seven332:
## Summary

- share registered metadata subscript key diagnostics with annotation-only targets
- preserve annotation-only child evaluation without modeling an outer subscription operation
- cover direct and tracked-alias targets through the public linter entry point

## Scope

This changes only the Python metadata-key linter and its integration fixtures. It does not change runtime add-on behavior, schemas, APIs, runner protocols, persisted state, dependencies, or deployment compatibility.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_flow_metadata_key_contract.py -q` (20 passed)
- `uv run --no-sync python -m pytest tests/` (5,182 passed)

Closes #25979

### DIFF
--- a/crates/runner/mitm-addon/scripts/flow_metadata_linter/visitor.py
+++ b/crates/runner/mitm-addon/scripts/flow_metadata_linter/visitor.py
@@ -310,6 +310,10 @@ class _MetadataKeyVisitor(ast.NodeVisitor):
         self._metadata_key_checked_node_ids.add(node_id)
         self._add_violations(violations)
 
+    def _record_metadata_subscript_key_violations(self, node: ast.Subscript) -> None:
+        if self._is_metadata_alias_value(node.value):
+            self._add_violations(_metadata_key_expression_violations(self.path, node.slice))
+
     def _metadata_default_argument_names(self, args: ast.arguments) -> set[str]:
         metadata_defaults: set[str] = set()
         positional_args = [*args.posonlyargs, *args.args]
@@ -411,6 +415,7 @@ class _MetadataKeyVisitor(ast.NodeVisitor):
         if isinstance(node, ast.Subscript):
             self._record_metadata_merge_key_violations(node.value)
             self._record_metadata_merge_key_violations(node.slice)
+            self._record_metadata_subscript_key_violations(node)
             self.visit(node.value)
             self.visit(node.slice)
 
@@ -827,8 +832,7 @@ class _MetadataKeyVisitor(ast.NodeVisitor):
     def visit_Subscript(self, node: ast.Subscript) -> None:
         self._record_metadata_merge_key_violations(node.value)
         self._record_metadata_merge_key_violations(node.slice)
-        if self._is_metadata_alias_value(node.value):
-            self._add_violations(_metadata_key_expression_violations(self.path, node.slice))
+        self._record_metadata_subscript_key_violations(node)
         self.generic_visit(node)
 
     def visit_Call(self, node: ast.Call) -> None:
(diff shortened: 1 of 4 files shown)


seven332:
## Code Review: PR #26004

**Reviewed head:** `b0d2a899e35e05b1f04ed614f03c20b562c07416`

### Summary

The PR centralizes metadata subscript key diagnostics and applies them to annotation-only targets while retaining the distinct Python evaluation model for those targets.

### Findings

No findings.

### Testing

The regression fixture covers direct and tracked-alias targets through the public linter API with exact diagnostics. Locked uv setup, Ruff format and lint, BasedPyright, the focused contract tests, and all 5,182 add-on tests pass locally. Unrelated Turbo and Rust suites were not run because they have no changed input or consumer in this PR; GitHub CI is checked separately.

### Verdict

LGTM

PR #26040, opened 2026-08-10
fix(sandbox-fc): preserve pending balloon targets

### LINKED ISSUE #26030: bug(rust): prevent balloon controller from undoing unpark deflation
## Problem

Firecracker balloon updates are asynchronous. `PATCH /balloon` changes the requested target, while `actual_mib` can continue reporting the amount currently held by the guest driver until it converges. The reactive controller currently derives every new absolute target from `actual_mib` and ignores `target_mib`, so a tick can overwrite a newer in-flight request with a stale, less aggressive target.

`unpark_inner` exposes the most damaging form of this race: it requests target `0` to restore guest memory and immediately restarts the controller. The controller's first interval tick is immediate. If statistics report `target_mib = 0`, a lagging nonzero `actual_mib`, and low available memory, the deflate branch calculates a positive target from `actual_mib` and replaces the pending zero target.

## Evidence

- `crates/sandbox-fc/src/sandbox.rs` resumes the VM, PATCHes balloon target `0`, and immediately spawns the reactive controller.
- `crates/sandbox-fc/src/balloon.rs` uses `tokio::time::interval`, whose first tick is immediate.
- Firecracker defines `target_mib` as the amount the device aims to hold and `actual_mib` as the amount it currently holds; divergence is therefore a normal in-flight state.
- The controller uses only `actual_mib` as its calculation base for both inflate and deflate updates.
- For a 2 GiB VM, `target_mib = 0`, `actual_mib = 1250`, and `available_memory = 128 MiB` currently produce a new target of `1122 MiB`, overriding the unpark request.
- Existing controller tests set `target_mib == actual_mib`. The lifecycle test checks only that the first balloon PATCH is zero and permits later controller PATCHes.

## Impact

An unparked sandbox can return to active service while most of its memory remains ballooned. Recovery then proceeds only through bounded five-second controller steps, increasing latency and OOM risk. The same stale-base calculation can also weaken other in-flight inflate or deflate requests.

## Decision

Fix the controller's handling of asynchronous target/actual state, with unpark as the regression case. A target/actual mismatch must not be treated as an error or simply block all reactive decisions: a new memory-pressure signal may legitimately require reversing an earlier request. The required behavior is that a tick must not accidentally weaken an outstanding request when it selects the same adjustment direction, while an opposite pressure signal must remain able to reverse direction intentionally.

## Constraints

- Keep Firecracker's asynchronous target/actual model; do not wait synchronously for full convergence on every controller tick.
- Preserve the controller's existing memory signals, hysteresis thresholds, per-tick inflate cap, maximum balloon bound, and API error propagation.
- Preserve park/unpark ordering and controller shutdown behavior.
- Do not add polling or delay to the unpark critical path solely to mask the stale-target calculation.

## Acceptance criteria

- After unpark requests target `0`, a controller tick with `target_mib = 0`, lagging nonzero `actual_mib`, and low available memory does not emit a positive PATCH that weakens the pending deflation.
- For both inflate and deflate decisions, an in-flight request in the same direction is not replaced by a less aggressive target because `actual_mib` lags `target_mib`.
- A memory signal selecting the opposite direction can still reverse an in-flight request instead of waiting for target/actual convergence.
- Deterministic controller-level regression coverage exercises mismatched target/actual values, including the unpark state, and asserts the emitted PATCH target or absence of a PATCH.
- Existing behavior for converged statistics, missing statistics, bounds, controller lifecycle, park/unpark ordering, and API failures remains covered and unchanged.

seven332:
## Summary

- merge each actual-derived balloon candidate with Firecracker's outstanding target in the selected direction
- preserve a more aggressive in-flight target without stacking target-relative adjustments
- keep opposite memory-pressure signals able to reverse an in-flight request immediately
- cover pending inflate, pending deflate, and the post-unpark target-zero state through the Firecracker API test boundary

## Scope

This does not change park/unpark ordering, controller polling, memory thresholds, bounds, API contracts, persisted state, or deployment compatibility behavior.

## Validation

- `cargo clippy -p sandbox-fc --all-targets --all-features`
- `cargo doc -p sandbox-fc --all-features --no-deps`
- `cargo test -p sandbox-fc`

Closes #26030

### DIFF
--- a/crates/sandbox-fc/src/balloon.rs
+++ b/crates/sandbox-fc/src/balloon.rs
@@ -481,6 +481,17 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn tick_advances_pending_inflate_to_more_aggressive_candidate() {
+        // The 700 MiB target is ahead of actual but behind the 756 MiB
+        // actual-relative candidate, so the controller should advance it.
+        let stats = r#"{"target_mib":700,"actual_mib":500,"target_pages":179200,"actual_pages":128000,"free_memory":1073741824,"available_memory":1073741824}"#;
+        let patch = run_tick_with_mock(stats, 1536)
+            .await
+            .expect("expected PATCH advancing pending inflation");
+        assert_eq!(patch_amount_mib(&patch), 756);
+    }
+
+    #[tokio::test]
     async fn tick_reverses_pending_deflate_on_high_free_memory() {
         // A target below actual means deflation is in flight. A new inflate
         // signal should cross the actual size immediately: 500 + 256 = 756.
@@ -530,6 +541,17 @@ mod tests {
             patch.is_none(),
             "should preserve the pending zero target from unpark"
         );
+    }
+
+    #[tokio::test]
+    async fn tick_advances_pending_deflate_to_more_aggressive_candidate() {
+        // The 400 MiB target is below actual but above the 372 MiB
+        // actual-relative candidate, so the controller should advance it.
+        let stats = r#"{"target_mib":400,"actual_mib":500,"target_pages":102400,"actual_pages":128000,"free_memory":52428800,"available_memory":134217728}"#;
+        let patch = run_tick_with_mock(stats, 1536)
+            .await
+            .expect("expected PATCH advancing pending deflation");
+        assert_eq!(patch_amount_mib(&patch), 372);
     }
 
     #[tokio::test]
--- a/crates/sandbox-fc/src/balloon.rs
+++ b/crates/sandbox-fc/src/balloon.rs
@@ -184,14 +184,19 @@ async fn tick(client: &ApiClient, max_inflate: u32, tick_count: u64) {
         );
     }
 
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #26040

**Reviewed head:** `113fe1ca465cb5749307c11ba15e7e0425d876bd`

### Summary

The controller keeps actual-relative feedback while merging each candidate with Firecracker's outstanding target in the selected direction. This prevents a lagging `actual_mib` from weakening an in-flight request, including unpark's target-zero deflation, without blocking legitimate reversals or stacking target-relative steps.

### Findings

No findings.

### Testing

The changed behavior is covered through the real `ApiClient` and mock Firecracker Unix socket for same-direction retention, same-direction progress, opposite-direction reversal, and the post-unpark state. `cargo fmt`, Clippy, crate docs, the full `sandbox-fc` test suite, and a focused 31-test balloon rerun passed. A live Firecracker sandbox exercise was not run.

### Verdict

LGTM

PR #26046, opened 2026-08-10
test(runner): make reload coalescing test lock-sensitive

### LINKED ISSUE #26029: test(rust): make systemd reload coalescing test lock-sensitive
## Problem

The focused regression test for host-global systemd reload coalescing does not
actually require the global flock in order to pass.

At `e0b0bb169ada4c746c7e5aa8d3acd6f367f3d396`:

- [`coordinate_systemd_reload_with_ops` acquires the host-global lock before
  reading dirty state](https://github.com/vm0-ai/vm0/blob/e0b0bb169ada4c746c7e5aa8d3acd6f367f3d396/crates/runner/src/cmd/service/reload.rs#L116-L145).
- [`FakeSystemdReloadOps::read_state` and `daemon_reload` both return
  immediately-ready futures](https://github.com/vm0-ai/vm0/blob/e0b0bb169ada4c746c7e5aa8d3acd6f367f3d396/crates/runner/src/cmd/service/reload.rs#L201-L229).
- [`different_units_coalesce_one_dirty_generation_under_real_flock` starts the
  two coordinators with `tokio::join!`](https://github.com/vm0-ai/vm0/blob/e0b0bb169ada4c746c7e5aa8d3acd6f367f3d396/crates/runner/src/cmd/service/reload.rs#L312-L343), but it has no rendezvous that makes their critical sections overlap.

If the lock acquisition at line 123 is deleted or bypassed, whichever joined
branch is polled first can still read `dirty = true`, increment the reload
counter, clear `dirty`, and finish without yielding. The other branch then
reads `dirty = false`. Both existing assertions therefore still observe one
performed reload and a reload count of one, so the test remains green despite
the serialization invariant being gone.

The real-metal behavior scenario does not close this gap: it starts two
different service operations concurrently, but its [reload budget explicitly
allows two through four reloads](https://github.com/vm0-ai/vm0/blob/e0b0bb169ada4c746c7e5aa8d3acd6f367f3d396/.github/scripts/runner-behavior-systemd-reload.sh#L87-L145).
That protects against per-operation amplification, not loss of different-unit
coalescing.

## Impact

The host-global coordinator was introduced by #24109 to prevent the
`daemon-reload` storm documented in #24080. A regression that removes or moves
the lock outside the dirty-state read/reload critical section can restore
duplicate concurrent manager reloads while both focused regression layers
continue to pass.

## Acceptance criteria

- Make the focused Rust test deterministically exercise two different-unit
  coordinators contending for the same host-global reload lock.
- Use explicit test synchronization at the lock/critical-section boundary; do
  not rely on scheduler timing, sleeps, or repeated attempts.
- Prove that the second coordinator cannot read state or reload while the first
  holds the lock, then rechecks the cleared dirty state after release.
- Assert exactly one successful `daemon_reload` and exactly one coordinator
  reporting that it performed the reload.
- Bound every synchronization wait so a broken implementation fails instead of
  hanging the test suite.
- Verify that bypassing the global lock makes the regression test fail.

seven332:
## Summary

- gate the fake systemd state read and reload phases so the leader's critical
  section is observable without scheduler timing
- probe the production lock path with independently opened flock handles before
  state read and during reload, proving both acquisition order and guard scope
- start the follower through an explicit pending-poll handshake, then verify it
  rechecks the cleared state and skips a duplicate reload after release

## Scope

This changes only inline `runner` test code. Production systemd coordination,
the shared lock implementation, runner behavior, dependencies, protocols,
persisted state, workflows, and deployment compatibility are unchanged.

## Validation

- exact focused coalescing test
- temporary early-lock-drop mutation: focused test failed at the pre-read
  real-flock assertion as expected; mutation restored and test passed
- `cargo test --manifest-path crates/Cargo.toml -p runner --bin runner`
  (3,075 passed, 20 ignored child fixtures)
- `git diff --check`
- repository pre-commit hook (workspace Rust formatting, documentation,
  Clippy, file-size check, and commitlint)

Closes #26029

### DIFF
--- a/crates/runner/src/cmd/service/reload.rs
+++ b/crates/runner/src/cmd/service/reload.rs
@@ -185,17 +185,27 @@ pub(super) async fn coordinate_systemd_reload_bounded(
 
 #[cfg(test)]
 mod tests {
+    use std::future::Future as _;
     use std::sync::Arc;
     use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 
     use super::*;
 
     #[derive(Clone)]
+    struct OperationGate {
+        entered: Arc<tokio::sync::Notify>,
+        release: Arc<tokio::sync::Notify>,
+    }
+
+    #[derive(Clone)]
     struct FakeSystemdReloadOps {
         dirty: Arc<AtomicBool>,
         drain_override_loaded: bool,
         query_error: bool,
+        read_count: Arc<AtomicUsize>,
+        read_gate: Option<OperationGate>,
         reload_count: Arc<AtomicUsize>,
+        reload_gate: Option<OperationGate>,
     }
 
     impl SystemdReloadOps for FakeSystemdReloadOps {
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #26046

**Reviewed head:** `697440c38336cb481add2d17eb1e078291c49f12`

### Summary

The test-only synchronization deterministically exercises the production flock boundary. Independent real-lock probes verify acquisition before state read and guard retention through reload, while the explicitly started follower rechecks the cleared state after release.

### Findings

No findings.

### Testing

Runner formatting, Clippy, documentation, the focused regression test, the complete runner binary test target, diff checks, and repository pre-commit checks passed. A temporary early-lock-drop mutation made the focused test fail at the intended pre-read lock assertion, then the restored implementation passed.

### Verdict

LGTM

PR #26062, opened 2026-08-10
test(python): cover jsonl watcher failure before readiness

### LINKED ISSUE #26049: test(python): cover JSONL watcher startup failure before readiness
## Problem and evidence

`mitm_addon.configure()` publishes the runner's addon-ready marker only after
`runner_flush_lifecycle.start_runner_jsonl_flush_worker()` returns. That ordering is a
readiness contract: once the marker exists, the runner may issue JSONL flush requests and
expect the addon-owned watcher to observe and acknowledge them.

The existing configuration test covers successful startup and eventual request handling, but it
does not distinguish the current ordering from writing the ready marker first. Both operations
still complete on the happy path. A deterministic `threading.Thread.start()` failure does
distinguish them: the current code propagates the startup failure without publishing readiness,
whereas the reversed ordering would leave a false ready marker behind.

## Decision

Proceed with one configuration-entry-point regression test. This is a real unprotected lifecycle
contract, and the test can cover it without changing production behavior or adding a new
abstraction.

## Scope and rationale

Exercise the real `mitm_addon.configure()` path and the real JSONL watcher lifecycle. Patch only
the thread-start boundary to inject the failure, then restore real startup and retry. This checks
both fail-closed readiness and the lifecycle owner's ability to recover from a failed start.

Testing `start_runner_jsonl_flush_worker()` alone would miss the externally observable ready
marker contract. Keeping only the happy-path test would leave the safety-relevant ordering
unprotected. No production-code recovery or exception handling is needed because propagation is
already the correct behavior.

## Risks and mitigations

- Patching `threading.Thread.start` is process-global. Keep the patch scoped to the synchronous
  `configure()` call, restore it before starting the real watcher, and do not start unrelated
  threads inside that scope.
- A real watcher can leak into later tests if an assertion fails. Use deterministic cleanup around
  the successful retry; the suite's lifecycle reset remains the final backstop.
- Avoid asserting private call counts as the contract. Assert the exception identity, marker
  visibility, successful retry, and real watcher behavior/order.

## Acceptance criteria

- Patch `runner_flush_lifecycle.threading.Thread.start` to raise a sentinel exception during
  `mitm_addon.configure()`.
- Assert that the exact exception propagates and the addon-ready marker is not created.
- Assert that the failed start leaves no stale worker state that prevents a later retry.
- Restore real thread startup, retry configuration, and observe the worker starting while the
  ready marker is still absent; then assert readiness is published.
- Stop the real worker deterministically during cleanup.
- Verify that moving the ready-marker write before watcher startup makes the regression test fail.
- Keep production code and runtime behavior unchanged.

seven332:
## Summary

- inject a sentinel `Thread.start()` failure through the real addon configuration hook
- verify the exact failure propagates without publishing addon readiness
- retry with the real thread implementation and verify the watcher starts before readiness, with
  deterministic cleanup

## Scope

This is test-only coverage. It does not change production code, runtime behavior, APIs, protocols,
persisted state, dependencies, migrations, feature switches, or deployment compatibility.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (`5184 passed`)
- `lefthook run pre-commit --files crates/runner/mitm-addon/tests/test_addon_configuration.py`
- mutation check: temporarily publishing readiness before watcher startup made the new targeted
  test fail at the absent-marker assertion; the production ordering was then restored

Closes #26049

### DIFF
--- a/crates/runner/mitm-addon/tests/test_addon_configuration.py
+++ b/crates/runner/mitm-addon/tests/test_addon_configuration.py
@@ -3,6 +3,7 @@
 import importlib.util
 import json
 import sys
+import threading
 import uuid
 from collections.abc import Sequence
 from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #26062

**Reviewed head:** `ee2bd035923fe89ec5ebb50265768c62bd65a87c`

### Summary

The test exercises the real addon configuration entry point, injects only the thread-start failure
boundary, and retries through the real thread implementation. It proves exact error propagation,
fail-closed readiness, retriable lifecycle ownership, and watcher-before-marker ordering without a
production change. Addon and watcher files are both isolated to the test-owned temporary
filesystem, and the successful worker is stopped deterministically.

### Findings

No findings.

### Testing

- The new targeted test passed, including 20 serial lifecycle repetitions.
- A temporary reversed production ordering made the regression test fail at the absent-marker
  assertion; the production source was restored before submission.
- The complete addon suite passed: `5184 passed`.
- Remote CI status remains to be checked independently after this review.

### Verdict

LGTM

PR #26063, opened 2026-08-10
test(python): cover websocket framing installation

### LINKED ISSUE #26050: test(python): cover incompatible WebSocket framing installation
## Problem

`crates/runner/mitm-addon/src/websocket_framing.py:276-283` has a fail-closed installation guard: a connection class carrying the vm0 marker is treated as already installed, while any other replacement of mitmproxy's original `WebsocketConnection` must raise instead of being silently overwritten.

No test exercises the incompatible-class branch. Removing that guard leaves all 21 relevant WebSocket framing and addon-configuration tests green. A deterministic test that installs a sentinel incompatible class correctly passes against current code, but fails against the mutation because installation silently replaces the sentinel.

Without this contract test, a regression could hide incompatible runtime composition, discard another connection wrapper, and violate the assumptions behind the pinned framing compatibility layer.

## Acceptance criteria

- Patch `websocket.WebsocketConnection` to a sentinel class without the vm0 marker.
- Assert `install_websocket_framing()` raises the expected compatibility error and leaves the sentinel installed.
- Explicitly cover the marked-class idempotency path and verify it remains a no-op.
- Keep the global class replacement isolated and restored after each test.
- Ensure deleting the incompatible-class guard causes the test to fail.

seven332:
## Summary

- cover the fail-closed `WebsocketConnection` installation guard with an unmarked sentinel
- assert an incompatible class is preserved when installation raises
- cover marker-based idempotency with an independent marked sentinel
- isolate both process-global replacements with pytest `monkeypatch`

## Scope

This is a test-only change. It does not change mitmproxy addon runtime behavior, dependencies, APIs, schemas, persisted state, runner protocols, configuration, or deployment ordering.

## Validation

- `uv lock --check`
- `uv sync --locked`
- focused installation contract tests
- `uv run --no-sync python -m pytest tests/test_mitmproxy_websocket_framing.py tests/test_addon_configuration.py` (23 passed)
- mutation check: removing the incompatible-class guard makes the new case fail with `DID NOT RAISE`; the guard was restored and the focused cases passed again
- `uv run --no-sync python -m pytest tests/` (5,185 passed)
- `git diff --check`

Closes #26050

### DIFF
--- a/crates/runner/mitm-addon/tests/test_mitmproxy_websocket_framing.py
+++ b/crates/runner/mitm-addon/tests/test_mitmproxy_websocket_framing.py
@@ -210,6 +210,8 @@ def _data_sends(observed: list[commands.Command]) -> list[commands.SendData]:
     ]
 
 
+# Only process composition can introduce a conflicting connection class; no
+# proxied WebSocket flow can construct this installer state.
 def test_install_websocket_framing_rejects_incompatible_connection_class(
     monkeypatch: pytest.MonkeyPatch,
 ) -> None:
--- a/crates/runner/mitm-addon/tests/test_mitmproxy_websocket_framing.py
+++ b/crates/runner/mitm-addon/tests/test_mitmproxy_websocket_framing.py
@@ -210,6 +210,41 @@ def _data_sends(observed: list[commands.Command]) -> list[commands.SendData]:
     ]
 
 
+def test_install_websocket_framing_rejects_incompatible_connection_class(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    class IncompatibleWebsocketConnection:
+        pass
+
+    monkeypatch.setattr(websocket, "WebsocketConnection", IncompatibleWebsocketConnection)
+
+    with pytest.raises(
+        RuntimeError,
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #26063

**Reviewed head:** `4116b3bd9a575317a354e0d7216e5e572a94a505`

### Summary

The PR adds direct contract coverage for the pinned mitmproxy WebSocket installer. An unmarked sentinel must raise before assignment and remain installed, while an independently marked sentinel must be preserved as an idempotent no-op. Pytest owns restoration of the process-global class, and the test records why this infrastructure-composition state cannot be constructed through a proxied WebSocket flow.

### Findings

No findings.

### Testing

- Locked dependency check and sync passed.
- Ruff format, Ruff lint, and BasedPyright passed.
- Framing plus addon configuration tests passed: 23 cases.
- Complete mitm-addon suite passed: 5,185 cases.
- Removing the incompatible-class guard made the new regression case fail with `DID NOT RAISE`; the guard was restored and the focused cases passed again.
- Rust and Turbo checks were not run because the PR changes only Python tests and does not affect runtime addon source, dependency metadata, embedding, or cross-language contracts.

### Verdict

LGTM

PR #26065, opened 2026-08-10
test(python): cover mixed-width content-length conflicts

### LINKED ISSUE #26048: test(python): cover mixed-width Content-Length conflicts
## Problem

`tests/test_content_length.py` covers conflicting duplicate and comma-separated values only when the normalized values have the same number of significant digits (`42` versus `43`, or equally long digit strings). It does not protect the parser's mixed-width comparison behavior.

The significant-digit width check in `crates/runner/mitm-addon/src/content_length.py` prevents two distinct regressions:

- With repeated fields, deleting the guard makes `parse(["1", "10"], max_value=1000)` raise `IndexError` because comparison indexes past the shorter first field.
- With a repeated field or comma list whose first normalized value is wider, deleting the guard makes inputs such as `["10", "1"]` and `["10, 1"]` incorrectly return `ContentLengthResult("valid", 10)` because comparison examines only the later value's width.

The existing suite remains green after that mutation. The previously cited `parse(["1, 10"], max_value=1000)` example happens to remain conflicting after mutation because comparison reaches the comma in the original field, so it is not evidence of the exception path.

This parser is shared by request-body admission paths. A regression in this guard could turn attacker-controlled conflicting `Content-Length` values into an addon exception or accept them as a single valid length instead of rejecting them deterministically.

## Decision

Add test coverage for the parser's mixed-width contract without changing production code or adding a property-testing dependency. Use explicit regression cases plus a deterministic bounded differential matrix against a straightforward reference interpretation.

## Acceptance criteria

- Cover different-width conflicts in both directions for repeated header fields and comma-separated lists.
- Include leading-zero variants so comparisons remain based on normalized significant digits.
- Assert that every conflicting case returns `ContentLengthResult("conflicting")` without raising.
- Add a deterministic bounded differential check that varies field count, comma composition, optional whitespace, leading zeros, invalid tokens, significant-digit width, and the configured maximum.
- Ensure deleting the significant-digit width comparison causes the tests to fail through either an exception or an incorrect result.
- Keep the change test-only unless the new coverage exposes an independent production defect.

seven332:
## Summary

- add named mixed-width `Content-Length` conflict cases for repeated fields and comma lists in both width orders
- cover leading-zero normalization and both the exception and false-valid mutation paths
- add a deterministic bounded differential matrix across field composition, whitespace, invalid tokens, digit widths, and limits

## Scope

- test-only change in the mitmproxy add-on parser contract
- no production code, dependency, lockfile, documentation, API, or deployment changes

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_content_length.py` (`36 passed`)
- `uv run --no-sync python -m pytest tests/` (`5195 passed`)
- mutation verification: removing the significant-width guard produced 10 targeted failures; the guard was restored before final validation

Closes #26048

### DIFF
--- a/crates/runner/mitm-addon/tests/test_content_length.py
+++ b/crates/runner/mitm-addon/tests/test_content_length.py
@@ -1,5 +1,7 @@
 """Protocol contract for shared Content-Length field parsing."""
 
+from itertools import chain, product
+
 import pytest
 from mitmproxy import http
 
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #26065

**Reviewed head:** `ed5fa3a001d081a258296057298876a2738150b2`

### Summary

The PR adds explicit mixed-width `Content-Length` regressions and an independently structured, deterministic differential matrix. The submitted diff remains test-only, dependency-free, bounded, and aligned with the approved parser-contract scope.

### Findings

No findings. The initial self-review gap around a conflict preceding a later invalid token is resolved in the reviewed head.

### Testing

- Ruff format and lint: passed
- BasedPyright: passed with 0 errors
- Targeted parser contract: 36 passed in 0.09 seconds
- Complete mitmproxy add-on suite: 5,195 passed
- Width-guard mutation: produced 10 targeted failures before restoration

No production behavior changed. GitHub CI status is evaluated separately from this code-review verdict.

### Verdict

LGTM

PR #25943, opened 2026-08-09
test(app): assert voice recorder rotation at lifecycle boundary

e7h4n:
## Problem

Turbo merge-group run [31313412968](https://github.com/vm0-ai/vm0/actions/runs/31313412968) failed in [`test-app (1)`](https://github.com/vm0-ai/vm0/actions/runs/31313412968/job/93244724878) at:

`chat-computer-use-and-voice.test.tsx > chat lifecycle > transcribes a voice input segment after silence while recording`

The test could not find the `Stop recording` label immediately after the STT request entered its mock handler. This is a recurrence of the signature addressed by #22845, but the earlier request-arrival deferred did not own the recording lifecycle.

## Root cause

When the voice activity tracker reports silence, the product starts two independent operations in the same callback:

1. capture the current segment, rotate to a new `MediaRecorder`, and upload the captured blob in the background;
2. start the 50 ms Vitest long-silence auto-stop timer.

The STT mock request proves that upload began, but it does not prove the auto-stop timer has not already completed. Event-loop scheduling therefore decided whether the transient `Stop recording` label was still present when the test resumed. Both the source PR run and adjacent base-main run passed the test, and #25936 did not change voice code.

## Fix

- expose `onRecorderStart` from the existing browser-boundary voice-input mock;
- replace the transient post-request label assertion with the recorder lifecycle sequence `start -> stop -> start`, which directly proves that silence capture rotated the recorder instead of ending the session;
- assert the final sequence `start -> stop -> start -> stop` after the legitimate long-silence auto-stop.

The test retains the initial user-visible `Stop recording` assertion, the uploaded `voice-1` blob, one transcription, composer and draft delivery, and the final `Voice input` state. No product behavior, retry, timeout, sleep, or assertion polling was added.

## Verification

- full `chat-computer-use-and-voice.test.tsx`: 28/28 passed in 7 consecutive runs, including one after rebasing onto the latest `origin/main`;
- `pnpm format`;
- `pnpm turbo run lint --concurrency=1 --log-order grouped`;
- `pnpm knip`;
- staged TypeScript checks for non-API/App packages, App, and API;
- `git diff --check`.

Preview validation is not applicable: this PR changes only Vitest code and the browser boundary mock; a deployed app preview cannot exercise the synthetic RMS and `MediaRecorder` lifecycle used by this regression test.

PR #19914, opened 2026-07-02
fix: sample voice activity before silent detection

e7h4n:
## Summary

Fixes a race in voice-input speech detection that made `chat-lifecycle.test.tsx` intermittently leave the composer empty after STT returned text. The voice activity monitor now takes its first analyser sample immediately, then schedules animation-frame polling, so a recording cannot be marked as covered by silence detection before any audio sample has been evaluated.

## Source Turbo Failure

- Run: https://github.com/vm0-ai/vm0/actions/runs/28595693478
- Failed job/shard: `test-app (7)` running `pnpm vitest run --project=@vm0/app --shard=7/8 --coverage.enabled --coverage.reporter=json`
- Failure: `src/views/zero-page/__tests__/chat-lifecycle.test.tsx > chat lifecycle > transcribes voice input into the composer`
- Assertion: expected the composer textarea to contain `Summarize the standup`, but it was empty.
- Recurrence: the same test/signature also failed in https://github.com/vm0-ai/vm0/actions/runs/28570983737.

## Diagnosis

`startAudioActivityMonitor` previously returned after scheduling the first `requestAnimationFrame` sample. `startRecording$` then marked `internalVoiceActivityAvailable$` and `internalVoiceActivityCoversRecording$`, but `internalVoiceDetectedDuringRecording$` could still be false until the first animation frame ran. If the test stopped recording in that window, `stopAndTranscribe$` treated the recording as silent and skipped STT, leaving the composer empty.

The fix preserves the existing polling loop but calls `update()` once immediately so the first analyser sample is processed before the monitor is considered ready.

## Docs and Patterns Applied

- `docs/testing.md`: avoid waits/manual cleanup; fix the real async synchronization issue.
- `docs/testing/patterns.md`: keep the platform UI test path and existing mocks.
- `turbo/docs/no-floating-promise.md`: no new floating promises, sleeps, or detached cleanup.
- `.claude/skills/ccstate/SKILL.md`: existing DOM callback detach/page-signal pattern remains unchanged.
- `.github/workflows/turbo.yml`: verified the failing CI command and app shard.

## Validation

- `pnpm -C turbo exec vitest run --project=@vm0/app apps/platform/src/views/zero-page/__tests__/chat-lifecycle.test.tsx -t "transcribes voice input into the composer"`
- `pnpm -C turbo exec vitest run --project=@vm0/app apps/platform/src/views/zero-page/__tests__/chat-lifecycle.test.tsx -t "voice input|recording before the voice activity monitor|billing recovery"`

PR #26098, opened 2026-08-10
fix(app): preserve sampled voice before silence timeout

e7h4n:
## Problem

Turbo push run [31364373895](https://github.com/vm0-ai/vm0/actions/runs/31364373895) failed in [`test-app (1)`](https://github.com/vm0-ai/vm0/actions/runs/31364373895/job/93380168047) at:

`chat-computer-use-and-voice.test.tsx > chat lifecycle > transcribes a voice input segment after silence while recording`

The recorder lifecycle was `start -> stop` instead of `start -> stop -> start`. This is a recurrence after #25943, not a duplicate of its test-only repair.

## Root cause

`startAudioActivityMonitor` intentionally samples the first analyser frame synchronously before reporting the monitor ready, as established by #19914. That first sample can record active voice in the segment before `startRecording$` calls `startSilenceTimeout()`.

The first voice callback tries to clear the silence timer before the timer exists. Startup then unconditionally armed the 50 ms Vitest auto-stop timer even though the session had already observed voice.

Two schedules followed:

- when a later animation-frame state change arrived within 50 ms, it cleared the timer and the normal silence transition later rotated the recorder;
- when a loaded shard delayed that frame, the startup timer stopped the initial recorder and the STT request came from final stop capture, so no replacement recorder existed.

The failed test completed in 1071 ms, while the adjacent parent-main pass took 1912 ms, consistent with auto-stop winning before the 500 ms voice-activity hold reached the silence-segmentation path. The source PR changed only Runner Python code.

## Fix

Only arm the initial silence timeout when the recording session has not already captured voice.

This preserves both product contracts:

- a recording with no initial voice still auto-stops;
- after real silence, the segment is uploaded, the recorder rotates, and the continued session receives its normal long-silence timeout.

No retry, sleep, timeout increase, skipped test, weakened assertion, detached cleanup, or swallowed rejection was added.

## Verification

- full `chat-computer-use-and-voice.test.tsx`: 28/28 passed in 7 post-fix runs, including one after rebasing onto the latest `origin/main`;
- `pnpm format`;
- `pnpm turbo run lint --concurrency=1 --log-order grouped`;
- `pnpm knip`;
- staged TypeScript checks for non-API/App packages, App, and API;
- `git diff --check`.

Preview validation is not applicable. This invariant requires the Vitest browser-boundary mock to inject an RMS sequence and observe synthetic `MediaRecorder` transitions; a deployed preview cannot inject or inspect that microphone lifecycle deterministically.

### DIFF
--- a/turbo/apps/platform/src/signals/voice-io/voice-io-stt.ts
+++ b/turbo/apps/platform/src/signals/voice-io/voice-io-stt.ts
@@ -972,7 +972,9 @@ function createVoiceSegmentSession(
       }
     },
     startSilenceTimeout(): void {
-      silenceTimer.start();
+      if (!currentSegmentHasVoice) {
+        silenceTimer.start();
+      }
     },
     async stopAndTranscribe(stopSignal: AbortSignal): Promise<void> {
       stopped = true;


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 17752e88544c4ed9f9dbda9b8b5c45e0e1f7614b
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

This fix prevents recording startup from arming a silence timeout after the synchronous first analyser sample has already marked the current segment as containing voice. The normal `handleActivity` path still owns later silence segmentation, recorder rotation, and the subsequent long-silence timeout; no compatibility fallback is introduced.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and race behavior | Pass | `turbo/apps/platform/src/signals/voice-io/voice-io-stt.ts:974` now checks the session's already-sampled voice state before starting the startup timer; `handleActivity` still starts the timer after an actual silence transition. |
| `REVIEW.md` | Security and persisted/API contract safety | N/A | The one-line behavioral guard changes no authentication, input handling, API shape, database state, or deployable protocol. |
| `REVIEW.md` | Type safety and maintainability | Pass | The change uses the existing typed session state, adds no cast or interface change, and keeps the decision beside the timer operation it governs. |
| `REVIEW.md` | Fix commits have regression coverage | Pass | `chat-computer-use-and-voice.test.tsx:1071` exercises the production page flow and asserts the silence segment stops and replaces the recorder before transcription. |
| `REVIEW.md` | Fallback discipline | Pass | The condition represents the current recording session's domain state; it does not add a legacy/new compatibility branch, fabricated default, or recovery path. |
| `docs/bad-smell.md` | No `any`, suppressions, dynamic imports, hardcoded configuration, or fallback slop | Pass | The complete diff is a typed boolean guard and introduces none of the prohibited constructs. |
| `docs/testing.md` | Prefer integration behavior coverage | Pass | The existing regression test initializes the real platform page flow and verifies observable recorder transitions, uploaded audio, transcription, and draft state. |
| `docs/testing.md` | Mock only boundaries; avoid fake timers, retries, and detached-cleanup patches | Pass | Coverage uses the browser and HTTP test boundaries and the original lifecycle assertion; the fix adds no test wait, retry, fake timer, timeout increase, or cleanup workaround. |
| `docs/effect.md` | Choose the domain owner and keep cancellation explicit | Pass | The recording session remains the timer owner, while `createVoiceSilenceTimer` continues to bind timeout cancellation to the recording signal and clear it on stop/cancel. |
| `docs/cache.md` | Resource lifecycle acquisition and teardown remain symmetric | Pass | The guard changes only whether the initial timer is acquired; every acquired timer still uses the existing reset/clear lifecycle owned by the session. |
| `docs/cache.md` | Avoid render, cache, ref, and duplicate-state anti-patterns | N/A | The diff does not touch React render, signal allocation, caches, refs, drafts, or state-machine ownership. |
| `.claude/skills/ccstate/SKILL.md` | AbortSignal ownership and no floating work in signals | Pass | The timer still receives its signal through `resetVoiceSilenceTimer$` parented to the recording signal; no `detach`, empty catch, or unawaited command was added. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- Coverage: Adequate. The existing regression test directly covers the original invariant and passed in seven consecutive full-file runs (28/28 each); PR CI `test-app (1)` also passed at the reviewed commit.
- Conventions: Compliant. The test exercises external behavior through platform initialization and boundary mocks, and the change preserves the original strict recorder-lifecycle assertion.

PR #26100, opened 2026-08-10
fix(mitm-addon): distinguish repeated-slash diagnostic bases

### LINKED ISSUE #26051: bug(python): distinguish repeated-slash bases in connector diagnostics
## Problem

`static_firewall_base_config_key()` collapses static bases that have different runtime matching semantics.

At `crates/runner/mitm-addon/src/firewall_matching/base_url.py:414-425`, the raw base first passes through `strip_optional_terminal_slash()`, which intentionally removes exactly one optional trailing slash. The returned path is then normalized again with `parts.path.rstrip("/")`. That second operation removes every remaining trailing empty path segment.

As a result, these two valid bases receive the same key:

- `https://api.example.com/v1`
- `https://api.example.com/v1//`

However, `match_base_url()` preserves their distinct semantics: the first matches `/v1/foo`, while the second does not; both match `/v1//foo` with different relative paths.

`crates/runner/mitm-addon/src/builtin_connector_diagnostics.py:574-594` uses this key to decide which connector APIs share a base and need route-aware ownership resolution. Distinct bases are therefore incorrectly classified as one shared base.

## Reproduction

With two diagnostic connector entries using the bases above and the same catch-all route:

```python
base_a = "https://api.example.com/v1"
base_b = "https://api.example.com/v1//"
request_url = "https://api.example.com/v1//foo"

assert matching.static_firewall_base_config_key(base_a) == (
    matching.static_firewall_base_config_key(base_b)
)
assert matching.match_base_url("https://api.example.com/v1/foo", base_a) == ("/foo", {})
assert matching.match_base_url("https://api.example.com/v1/foo", base_b) is None
```

The production diagnostic catalog then produces two route-specific ownership matches. The normal compiled matcher correctly selects the more-specific `base_b`, but `builtin_connector_diagnostics.find_candidate()` returns `None` because shared-base resolution sees two route owners and treats the request as ambiguous.

Observed output on `f6dcfa43211c2211347b4bbfae08c3f03677ee15`:

```text
key_a https://api.example.com/v1
key_b https://api.example.com/v1
general_match FirewallAllow connector-b https://api.example.com/v1//
ownership_matches [('connector-a', True), ('connector-b', True)]
find_candidate None
```

## Impact

For valid connector catalog bases containing repeated trailing empty path segments, missing-connector diagnostics can be suppressed even when ordinary base specificity identifies one owner. A different permission arrangement can also select the broader connector as the route owner. This removes or misattributes the diagnostic that should explain why a connector is unavailable for the run.

## Acceptance criteria

- Static base keys preserve repeated trailing empty path segments while continuing to treat one terminal slash as optional.
- `https://api.example.com/v1` and `https://api.example.com/v1/` retain the same key.
- `https://api.example.com/v1//` has a distinct key from `https://api.example.com/v1`.
- Connector diagnostic coverage verifies that these distinct bases do not enter shared-base ambiguity handling and that the more-specific base remains selectable.
- Existing authority, default-port, IDNA, and single-terminal-slash normalization behavior remains unchanged.

seven332:
## Summary

- preserve repeated terminal empty path segments in static connector diagnostic base keys
- keep exactly one terminal slash optional while retaining existing authority normalization
- select the more-specific repeated-slash connector instead of suppressing its diagnostic as ambiguous

## Scope

This changes only the runner mitm addon's in-memory connector diagnostic grouping. It does not change APIs, persisted state, migrations, shared-base ownership policy, or deployment protocols.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_builtin_connector_diagnostics.py` (31 passed)
- `uv run --no-sync python -m pytest tests/` (5202 passed)
- `lefthook run pre-commit`

Closes #26051

### DIFF
--- a/crates/runner/mitm-addon/src/firewall_matching/base_url.py
+++ b/crates/runner/mitm-addon/src/firewall_matching/base_url.py
@@ -422,7 +422,7 @@ def static_firewall_base_config_key(raw_base: str) -> str | None:
     )
     if parts is None:
         return None
-    return f"{parts.scheme.lower()}://{parts.authority}{parts.path.rstrip('/')}"
+    return f"{parts.scheme.lower()}://{parts.authority}{parts.path}"
 
 
 def static_firewall_base_authority_key(raw_base: str) -> str | None:
--- a/crates/runner/mitm-addon/tests/test_builtin_connector_diagnostics.py
+++ b/crates/runner/mitm-addon/tests/test_builtin_connector_diagnostics.py
@@ -515,6 +515,42 @@ def test_find_candidate_suppresses_multiple_shared_base_route_owners():
     assert candidate is None
 
 
+def test_find_candidate_distinguishes_repeated_slash_bases():
+    permissions = [{"name": "read", "rules": ["GET /{path+}"]}]
+    snapshot = _diagnostic_snapshot(
+        [
+            _firewall(
+                "broader",
+                "BROADER_TOKEN",
+                base="https://api.example.com/v1",
+                permissions=permissions,
+            ),
+            _firewall(
+                "repeated",
+                "REPEATED_TOKEN",
+                base="https://api.example.com/v1//",
+                permissions=permissions,
+            ),
+        ]
+    )
+
+    candidate = builtin_connector_diagnostics.find_candidate(
+        snapshot,
+        "https://api.example.com/v1//foo",
+        "GET",
+        active_firewall_names=set(),
+    )
+
+    assert candidate == builtin_connector_diagnostics.ConnectorDiagnosticCandidate(
+        connector_slug="repeated",
+        reason="not_configured_for_run",
+        env_names=("REPEATED_TOKEN",),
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #26100

**Reviewed head:** `1797fb3da8f89d028e3b9bdfe8b6c7302d66d750`

### Summary

The change aligns connector diagnostic base-key equivalence with the existing runtime matcher by preserving significant repeated terminal empty path segments after one optional slash is removed. The regression test exercises the real diagnostic catalog and `find_candidate()` path.

### Findings

No findings.

### Testing

Validated with uv lock/sync, Ruff format and lint, BasedPyright, the focused connector diagnostic suite (31 passed), the complete addon suite (5202 passed), and pre-commit hooks. The new and existing normalization regressions also passed together. GitHub CI is assessed separately.

### Verdict

LGTM

PR #26114, opened 2026-08-10
fix(mitm-addon): reject non-list custom connector apis

### LINKED ISSUE #26092: bug(python): reject non-list custom connector APIs during registry resolution
## Summary

`resolve_firewall_entries()` assumes an inline firewall's `apis` value is iterable when the entry has a valid `customConnectorId`. At `crates/runner/mitm-addon/src/registry_firewalls.py:305-311`, an explicit non-iterable value such as `"apis": null` raises a raw `TypeError` while the resolver is propagating the custom connector ID.

The registry classifier only catches `FirewallEntryResolutionError` around this call (`crates/runner/mitm-addon/src/registry.py:293-301`), so the exception escapes per-VM classification instead of recording the affected VM as `invalid_firewalls`.

## Reproduction

```python
registry_firewalls.resolve_firewall_entries(
    {
        "runId": "run-custom",
        "firewalls": [
            {
                "kind": "inline",
                "customConnectorId": "550e8400-e29b-41d4-a716-446655440000",
                "firewall": {"name": "custom", "apis": None},
            }
        ],
    }
)
```

Current result:

```text
TypeError: 'NoneType' object is not iterable
```

The same exception propagates from `_classify_registry_vms()` for a registry containing this VM.

## Impact

One malformed inline custom-connector firewall can abort classification of the entire registry snapshot. Other VM entries are not published through the normal per-entry rejection path, and request-path registry lookup can receive the unexpected exception instead of a bounded invalid-VM result.

## Acceptance criteria

- Validate or safely handle the inline firewall `apis` shape before propagating `customConnectorId` to API entries.
- Non-list/non-iterable `apis` values must not escape registry classification as `TypeError`.
- A malformed custom-connector VM is rejected through the existing `invalid_firewalls` path without preventing valid VMs in the same registry from loading.
- Add regression coverage for at least `apis: null` with a valid `customConnectorId`.

seven332:
## Summary

- reject non-list API containers before propagating custom connector identity
- classify the malformed VM through the existing `invalid_firewalls` boundary
- cover a mixed registry snapshot where a valid peer VM remains available

## Scope

This changes only Python registry resolution for inline entries with a valid `customConnectorId`. It does not change the TypeScript or Rust execution-firewall contracts, general inline firewall matcher tolerance, database schemas, or deployment protocols.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_registry_inline_firewalls.py` (4 passed)
- `uv run --no-sync python -m pytest tests/` (5214 passed)

Closes #26092

### DIFF
--- a/crates/runner/mitm-addon/src/registry_firewalls.py
+++ b/crates/runner/mitm-addon/src/registry_firewalls.py
@@ -306,7 +306,10 @@ def resolve_firewall_entries(
             custom_connector_id = _custom_connector_id(entry)
             if custom_connector_id is not None:
                 resolved_firewall["customConnectorId"] = custom_connector_id
-                for api in resolved_firewall.get("apis", []):
+                raw_apis = resolved_firewall.get("apis")
+                if not isinstance(raw_apis, list):
+                    raise FirewallEntryResolutionError("inline firewall apis must be a list")
+                for api in raw_apis:
                     if isinstance(api, dict):
                         api["customConnectorId"] = custom_connector_id
             resolved.append(resolved_firewall)
--- a/crates/runner/mitm-addon/tests/test_registry_inline_firewalls.py
+++ b/crates/runner/mitm-addon/tests/test_registry_inline_firewalls.py
@@ -114,3 +114,30 @@ class TestRegistryInlineFirewalls:
             "custom-api-id",
             "run-inline:1",
         ]
+
+    def test_malformed_custom_connector_apis_reject_only_affected_vm(self, tmp_path, mitm_ctx):
+        path = tmp_path / "registry.json"
+        malformed_vm = inline_vm("run-malformed")
+        malformed_entry = malformed_vm["firewalls"][0]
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #26114

**Reviewed head:** `48f9af1b5a3d6b06ae624960a11268a059bf0543`

### Summary

The resolver now validates the API container at the exact custom connector identity-propagation boundary and returns the established domain error. The real-file registry integration regression verifies malformed-VM isolation and continued availability of a valid peer without changing ordinary inline matcher tolerance.

### Findings

No findings.

### Testing

`uv lock --check`, locked environment sync, Ruff format/lint, BasedPyright, the focused registry tests (4 passed), and the full mitm-addon suite (5214 passed) all completed successfully. Unrelated Turbo and Rust suites were not run because no shared or cross-language contract changed. CI remained pending at review time.

### Verdict

LGTM

PR #26122, opened 2026-08-10
fix(runner): frame rootfs hash inputs

### LINKED ISSUE #26096: security(rust): make rootfs guest-binary hashing unambiguous
## Summary

The local rootfs cache key concatenates each guest-binary destination tag and raw file content without length framing. Because guest binary bytes are caller-selectable and the inventory contains multiple entries, two distinct ordered binary tuples can produce the exact same SHA-256 preimage without finding a cryptographic collision.

This breaks the invariant that `rootfs_hash` uniquely identifies the guest binaries installed in the bootable rootfs.

## Evidence

- `crates/runner/src/cmd/build/hashes.rs:65-98` computes the rootfs hash. The loop at lines 88-95 appends `bin:{dest}:` followed by unframed raw content for every guest binary.
- `crates/runner/guest-binaries.json:1-50` defines seven ordered guest binaries, including adjacent `/usr/local/bin/guest-agent` and `/usr/local/bin/guest-download` entries.
- `crates/runner/src/cmd/build/mod.rs:40-128` exposes an independent path argument for each guest binary.
- `crates/runner/src/cmd/build/guest.rs:32-58` snapshots those caller-selected bytes and forwards every `(path, destination)` pair to the hash function.
- `crates/runner/src/cmd/build/mod.rs:420-454` derives cache paths from `rootfs_hash` and returns early when that rootfs and snapshot already exist. Lines 515-537 also reuse an existing rootfs under a shared lock.
- `crates/runner/src/paths.rs:350-369` stores the rootfs under `<images_dir>/<rootfs_hash>/rootfs.ext4`.
- `crates/runner/scripts/verify-rootfs.sh:203-214,242-246` checks only that each installed guest path is a regular executable file; it does not validate the hashed byte boundaries.

Let `D = b"bin:/usr/local/bin/guest-download:"`. For arbitrary byte strings `X` and `Y`, keep the other five binaries identical and compare these inputs:

- A: `guest-agent = X`, `guest-download = D || Y`
- B: `guest-agent = X || D`, `guest-download = Y`

Both serialize the first two entries as:

```text
bin:/usr/local/bin/guest-agent: || X || D || D || Y
```

The logical binary tuples differ, but the SHA-256 input is byte-for-byte identical. `resolve_guest` also sets every copied file to mode 0755, and rootfs verification checks existence and executable permission rather than file format, so the crafted boundary is not rejected before cache publication.

## Security impact

A compromised or untrusted build input can populate a persistent local image cache entry that a distinct guest-binary set later resolves to. The later build may silently reuse the earlier rootfs and its derived snapshot instead of installing the requested binaries. This creates a cache-poisoning boundary across builds and can preserve stale, invalid, or attacker-selected guest code under the expected content-addressed identity.

## Acceptance criteria

- Encode every variable-length rootfs hash input injectively, using explicit length framing or fixed-size per-input digests with unambiguous field boundaries.
- Bump the rootfs cache version so entries produced by the ambiguous encoding are never reused.
- Add a regression test using two distinct multi-binary tuples that collide under the current concatenation and assert that their new rootfs hashes differ.
- Preserve sensitivity to destination, content, CA fingerprint, template, DNS resolver, and rootfs disk size.
- Document the canonical rootfs hash encoding so future inputs cannot reintroduce boundary ambiguity.

seven332:
## Summary

- encode every local rootfs hash field with an explicit big-endian byte length
- separate guest destinations from guest contents so arbitrary bytes cannot shift field boundaries
- invalidate legacy local rootfs entries and cover the demonstrated two-binary collision

## Details

The canonical rootfs hash stream now uses `label || u64_be(value_length) || value` for every field. Fixed-width integers use big-endian bytes, IPv4 uses network-order octets, and each ordered guest binary contributes separate destination and content fields.

`ROOTFS_CACHE_VERSION` is bumped to `2`. This intentionally rebuilds local rootfs and derived snapshot entries while preserving shared R2 template cache identities.

The regression first proves that two distinct `guest-agent`/`guest-download` tuples produce identical bytes under the legacy assembly, then verifies that their new hashes differ.

## Exclusions

- no template hash or template cache version change
- no snapshot cache version change
- no API, database, protocol, or deployment configuration change

## Validation

- `cargo test --manifest-path crates/Cargo.toml -p runner cmd::build::hashes::tests`
- `cargo test --manifest-path crates/Cargo.toml -p runner`

Closes #26096

### DIFF
--- a/crates/runner/src/cmd/build/hashes.rs
+++ b/crates/runner/src/cmd/build/hashes.rs
@@ -20,7 +20,7 @@ const TEMPLATE_CACHE_VERSION: u32 = 1;
 ///
 /// Rootfs images are not shared through R2 because they include guest binaries
 /// and host-local CA material.
-const ROOTFS_CACHE_VERSION: u32 = 1;
+const ROOTFS_CACHE_VERSION: u32 = 2;
 
 /// Bump to invalidate all cached snapshots (local only; R2 stores only the template).
 const SNAPSHOT_CACHE_VERSION: u32 = 3;
@@ -58,10 +58,27 @@ pub(super) fn compute_template_hash(rootfs_disk_mb: u32) -> String {
     hex::encode(hasher.finalize())
 }
 
+fn update_rootfs_hash_field(hasher: &mut Sha256, label: &[u8], value: &[u8]) -> RunnerResult<()> {
+    let value_len = u64::try_from(value.len())
+        .map_err(|_| RunnerError::Internal("rootfs hash field exceeds u64 length".into()))?;
+    hasher.update(label);
+    hasher.update(value_len.to_be_bytes());
+    hasher.update(value);
+    Ok(())
+}
+
 /// Compute the local rootfs hash.
 ///
 /// This hash is what runner configs use. It includes the shared template hash plus
 /// every rootfs-only input that changes the bootable rootfs content.
+///
+/// The canonical encoding is a sequence of fields, each encoded as its fixed ASCII
+/// label, the value's byte length as a big-endian `u64`, then the value bytes. The
+/// fields are ordered as version, template hash, customization script, rootfs disk
+/// size, CA fingerprint, DNS resolver, then one destination/content pair per guest
+/// binary in inventory order. Fixed-width integers use big-endian bytes and IPv4
+/// addresses use their four network-order octets. Future inputs must use
+/// `update_rootfs_hash_field` so arbitrary value bytes cannot shift field boundaries.
 async fn compute_rootfs_hash(
     template_hash: &str,
     guest_bins: &[(&Path, &str)],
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #26122

**Reviewed head:** `a7e25a2467e160eb8391a6196fecc14e67b3c780`

### Summary

The PR replaces the ambiguous rootfs hash input with one architecture-independent length-framed field stream, invalidates legacy local cache entries, and adds a regression that demonstrates the exact old two-binary collision. The change stays within the approved runner-only scope and preserves shared template identities.

### Findings

No findings.

### Testing

Focused hash tests, the full runner suite (3077 passed, 20 ignored), Rust formatting, runner Clippy with warnings denied, and runner rustdoc all passed. No full VM rootfs build was run locally; CI owns that image-building coverage.

### Verdict

LGTM

PR #26123, opened 2026-08-10
perf(platform): eliminate idle animation paints

e7h4n:
## Summary

- replace the thinking label's animated `background-position` with paired masked, transform-only highlight layers while preserving the 3.5-second loop and queue interaction
- unmount the app skeleton overlay after its 300 ms fade-out, and avoid mounting it behind the inline bootstrap skeleton
- cover skeleton cleanup and keep the queued indicator's visual copies non-interactive

## Why

The performance trace showed `zero-shimmer-text` producing a `PaintImage` on every frame. It also showed the hidden app skeleton's infinite cursor animation continuing after the overlay reached `opacity: 0`.

## Validation

- `pnpm exec prettier --check` on all changed files
- `pnpm -F @vm0/app lint`
- `pnpm knip --workspace @vm0/app`
- `pnpm -F @vm0/app exec vitest run src/views/zero-page/__tests__/app-skeleton.test.tsx src/views/zero-page/__tests__/chat-thinking-indicator.test.tsx --silent=passed-only`
- `pnpm -F @vm0/app exec vitest run src/views/zero-page/__tests__/chat-run-results.test.tsx -t "shows server queue state only while the queue marker is unresolved" --silent=passed-only`

### DIFF
--- a/turbo/apps/platform/src/signals/app-skeleton.ts
+++ b/turbo/apps/platform/src/signals/app-skeleton.ts
@@ -10,6 +10,7 @@ import { locale$ } from "./locale.ts";
 // ---------------------------------------------------------------------------
 
 const internalVisible$ = state(true);
+const internalOverlayMounted$ = state(true);
 
 const APP_SKELETON_VISIBLE_EVENT = "vm0:app-skeleton-visible";
 const APP_SKELETON_VISIBLE_EVENT_QUEUED_KEY =
@@ -28,6 +29,7 @@ export const initBootstrapSkeleton$ = command(({ set }) => {
   if (active) {
     queueAppSkeletonVisibleEvent();
   }
+  set(internalOverlayMounted$, !active);
   set(internalBootstrapSkeletonActive$, active);
 });
 
@@ -172,6 +174,16 @@ export const appSkeletonVisible$ = computed((get) => {
   return get(internalVisible$);
 });
 
+export const appSkeletonOverlayMounted$ = computed((get) => {
+  return get(internalOverlayMounted$);
+});
+
+export const unmountAppSkeletonOverlay$ = command(({ get, set }) => {
+  if (!get(internalVisible$)) {
+    set(internalOverlayMounted$, false);
+  }
+});
+
 /**
  * Reveal the skeleton and reset the typewriter intro. `hideAppSkeleton$`
  * aborts the cycling loop via `resetSkeletonCycling$`; if the caller needs
@@ -180,8 +192,9 @@ export const appSkeletonVisible$ = computed((get) => {
  * restart the cycling itself by awaiting `startSkeletonCycling$` in its
  * own async context.
  */
-export const showAppSkeleton$ = command(({ set }) => {
+export const showAppSkeleton$ = command(({ get, set }) => {
   set(internalVisible$, true);
+  set(internalOverlayMounted$, !get(internalBootstrapSkeletonActive$));
   set(skeletonFirstCycle$, true);
 });
 
--- a/turbo/apps/platform/src/views/css/index.css
+++ b/turbo/apps/platform/src/views/css/index.css
@@ -276,29 +276,87 @@ body {
 }
 
 /* Shimmer text effect for active/thinking states */
-@keyframes zero-shimmer {
-  0% {
-    background-position: 100% center;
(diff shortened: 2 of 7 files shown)


e7h4n:
LGTM

PR_AUTO_REVIEW_COMMIT: 227a55f2b8f5cc7ca4d3cb14430576a80f1c8473
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

- Replaces the thinking label's animated `background-position` with two masked highlight windows whose animated properties are `transform` only, while retaining the 3.5-second linear loop.
- Preserves one semantic/interactive text tree and keeps both visual copies `aria-hidden`, pointer-inert, and non-selectable; the queued state therefore still exposes exactly one queue button.
- Separates skeleton visibility from overlay presence so the 300 ms fade can finish before unmount, avoids mounting the React overlay behind the inline bootstrap skeleton, and guards a late transition event from unmounting a re-shown skeleton.
- This is an internal rendering-performance change with no new public contract, deployment boundary, or feature-switch requirement.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | User-impact-first full-diff review, severity, tests, fallbacks, and final HEAD validation | Pass | Reviewed the complete seven-file diff and affected call paths; final HEAD remains `227a55f2b8f5cc7ca4d3cb14430576a80f1c8473`; no P0/P1 issue or migration fallback found. |
| `docs/docs.md` | Documentation, feature-switch, and deploy-boundary routing | Pass | The change is confined to platform rendering/CSS and page tests; it introduces no user-facing feature, public interface, configuration, or independently deployed contract. |
| `docs/bad-smell.md` | Type safety, suppression, duplication, and fallback checks | Pass | No `any`, ignore directive, lint suppression, unsafe assertion, duplicate implementation, or compatibility fallback is introduced. |
| `docs/testing.md` | Integration-first tests at the external boundary | Pass | New coverage enters through `detachedSetupPage` and asserts observable DOM presence, transition completion, and the queued button surface. |
| `docs/effect.md` | Correct owner/trigger and pure React render | Pass | Presence is application-scoped ccstate; hide/show and transition completion invoke semantic commands from events, with no render-phase writes or new effect. |
| `docs/cache.md` | Bounded ownership and symmetric lifecycle | Pass | The two atoms are bounded to app lifetime; the overlay releases its animated subtree after fade-out, bootstrap prevents duplicate mounting, and the visibility guard handles re-show races. |
| `.claude/skills/ccstate/SKILL.md` | ccstate command/computed and lifecycle conventions | Pass | State and computed values are module-owned, commands keep `get`/`set` inside ccstate, and the view uses stable `useSet` callbacks without accessor escape or detached work. |
| `docs/react-commit.md` | Correct performance attribution and preservation of semantic updates | Pass | The patch targets browser paint/compositing rather than claiming fewer React commits: the shimmer keyframes animate only transforms, while the actual semantic text remains a single tree. |
| `docs/testing/app-testing.md` | Platform page-test setup and mock lifecycle | Pass | Both modified suites use the production-style page bootstrap and the shared `testContext`; no component-only render or ad hoc mock lifecycle was added. |
| `docs/testing/testing-external-behavior.md` | Observe the page instead of internal Store state | Pass | Assertions use roles, visible text/button behavior, and DOM mount state; tests do not mutate or inspect ccstate internals. |
| `docs/testing/patterns.md` | Standard platform test structure | Pass | Coverage uses `detachedSetupPage`, Testing Library, and existing page helpers; test cleanup remains owned by the shared context. |
| `docs/testing/anti-patterns.md` | Avoid internal mocks, fake timers, and direct component rendering | Pass | No internal mock, fetch stub, fake timer, artificial delay, direct render, or mock-call assertion was introduced. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- Coverage: Adequate. The added page-level cases cover suppression of the duplicate React skeleton during inline bootstrap, unmount after the fade transition, and preservation of a unique interactive queue button despite visual shimmer copies.
- Conventions: Compliant. Tests exercise page-visible behavior through `detachedSetupPage`, use the shared context, and avoid internal mocks and fake timers.
- CI: All published checks for `227a55f2b8f5cc7ca4d3cb14430576a80f1c8473` are complete and passing, including format, ESLint, type checks, all eight app-test shards, security gates, and `codecov/patch`.

PR #26141, opened 2026-08-10
perf(python): bound and reuse sigv4 request inspection

### LINKED ISSUE #26101: perf(python): bound and reuse AWS SigV4 request inspection
## Problem

AWS SigV4 firewall handling repeatedly performs synchronous, request-controlled inspection on the shared mitmproxy event loop. For a body-dependent request, the same unmodified request URL and headers can be validated and classified during `requestheaders()`, request planning, preflight, and body-hash preparation before signing parses them again.

The repeated passes amplify work, but reuse alone is not a complete bound: one sufficiently large request target or header representation can still occupy the event loop for an unbounded amount of time.

## Evidence

Verified on current `main` at `fd6f59a029`:

- `mitm_addon.requestheaders()` calls `aws_sigv4_request_requires_body_for_signing()` before choosing header-phase signing or buffered handling.
- `_build_firewall_auth_plan()` calls the same classification again in both hook paths.
- Request-phase preflight and body-hash preparation call it again.
- Final signing calls `sign_request()`, which validates headers, rebuilds trusted URL state, parses and decodes the query, classifies the placeholder auth, and scans SigV4 query metadata again.
- The header-phase path also reparses during its plan and final signing even though the request representation is unchanged apart from deliberate pre-signing normalization or injection.
- mitmproxy 12.2.3 still contains an HTTP/1 request-head `FIXME` for header-size DoS protection, so the addon cannot rely on an earlier general request-head bound.

A local Python 3.14.0/aarch64 five-run median probe used a valid header-auth request with one ordinary query parameter:

| Query payload | One body-dependency classification | One `sign_request()` | Four classifications plus signing |
| ---: | ---: | ---: | ---: |
| 8 KiB | 0.40 ms | 0.48 ms | about 2.06 ms |
| 64 KiB | 2.53 ms | 2.80 ms | about 12.90 ms |
| 1 MiB | 42.89 ms | 44.40 ms | about 215.95 ms |

The absolute timings improved from the original Python 3.12 measurements, but both repeated amplification and linear worst-case work remain.

AWS front doors commonly impose much smaller limits: Application Load Balancers cap a request line at 16 KiB and total request headers at 64 KiB, while API Gateway documents URL limits of 8,192 or 10,240 characters depending on endpoint type. These are compatibility evidence for a bounded SigV4 input policy, not a universal threshold by themselves.

## Decision

Proceed as one Python addon PR. The issue should address both parts of the same event-loop CPU boundary:

1. reuse the inspection result while the SigV4-relevant request representation is unchanged; and
2. reject oversized SigV4 inspection input before full URL, query, or header-content parsing.

A timing regression check is useful evidence but is not an alternative to an explicit work bound. The exact target, header, and field-count limits and the ownership of reusable state must be justified in the Plan.

## Required direction

- Treat body dependency as a property of one inspected SigV4 request representation, not as a fresh query at every auth phase.
- Reuse both successful classification and conservative fail-closed classification across `requestheaders()`, request planning, preflight, and body-hash preparation while the relevant URL and headers are unchanged.
- Invalidate or replace reusable inspection state when pre-signing header or query mutation can affect signing; final signing must never consume stale parsed data.
- Establish explicit O(1)-checkable request-target and bounded-header-work limits before content-sensitive SigV4 parsing. Keep the policy scoped to managed AWS SigV4 handling rather than changing all proxy URL behavior.
- Preserve the current failure owner: malformed, unsupported, and oversized placeholder requests must fail closed through the request-hook SigV4 auth failure path without exposing resolved credentials.

## Alternatives considered

- **Status quo:** rejected because repeated GIL-bound work remains directly controlled by sandbox requests.
- **Reuse only:** removes amplification but leaves one unbounded event-loop parse and therefore does not close the resource-exhaustion boundary.
- **Timing test only:** can detect a large regression but cannot provide a deterministic safety bound and is sensitive to CI hardware.
- **Offload every SigV4 parse/sign operation:** would reduce event-loop blocking but moves unbounded attacker-controlled work into the shared executor, needs separate queue/admission ownership, and does not remove duplicate work. It is not justified while a bounded reusable inspection can solve the measured problem directly.
- **Global proxy request-head limit:** could be valuable separately, but it changes every protocol and connector path and cannot be justified solely from this SigV4-specific evidence.

## Risks and constraints

- Reusing state across a URL or header mutation could sign the wrong representation. The reusable artifact must have explicit request-local ownership and mutation boundaries.
- A limit that is too low could reject a previously accepted but valid AWS request. Plan thresholds against documented AWS limits and keep boundary tests.
- Caching an exception object or request-derived data beyond one flow would create unnecessary retention. Reuse must remain flow-local and be released at terminal cleanup.
- Conservative malformed-request classification must continue to require body handling when needed so early streaming does not bypass the final fail-closed signing error.
- Existing body-size admission, auth cancellation, auth-probe restoration, upstream revalidation, and signing error semantics must remain unchanged.

## Acceptance criteria

- For body-dependent and body-independent requests, each unchanged SigV4-relevant request representation is fully classified no more than once before signing.
- Request planning, preflight, and body-hash preparation consume the same request-local body-dependency result rather than reparsing.
- Final signing reuses valid prepared state only when its URL/header representation is still current; otherwise it performs exactly one fresh inspection after the mutation.
- Oversized request targets, excessive header fields, and excessive total header representation are rejected before full URL/query/header-content validation using explicit documented limits.
- Boundary and over-limit tests prove deterministic acceptance/rejection without timing thresholds.
- Public-hook integration coverage exercises header-phase body-independent handling, buffered body-dependent handling, malformed input, over-limit input, fallback restoration, and cancellation/body-admission cleanup.
- Existing placeholder validation, signature output, auth failure responses, upstream request mutation, and fail-closed behavior remain unchanged for in-limit requests.
- Focused static checks and SigV4/auth/request-framing tests pass, followed by the complete mitm-addon suite.

## Delivery notes

- This remains one independently reviewable Python addon PR because reusable inspection and its input limits share the same SigV4 parsing boundary and tests.
- No API, database schema, persisted data, migration, runner protocol, queue payload, dependency, environment variable, feature switch, or cross-deployment compatibility shim is required.
- Primary AWS limit references:
  - https://docs.aws.amazon.com/elasticloadbalancing/latest/userguide/how-elastic-load-balancing-works.html
  - https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-execution-service-limits-table.html

seven332:
## Summary

- bound managed AWS SigV4 request inspection to a 64 KiB request target, a 64 KiB HPACK-style header list, and 2,048 header fields
- reuse one validated inspection across the request-header and request hooks only while the exact URL and ordered headers still match
- release flow-local inspection state on signing, fallback, cancellation, direct rejection, and terminal cleanup paths
- preserve the existing fail-closed `aws_sigv4_auth_failed` response when an inspected request is malformed or over limit

## Compatibility

- no API, schema, migration, persisted-data, queue, or runner protocol changes
- no deployment ordering requirement; the behavior is local to the mitm addon

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/ -q` (5,224 passed)
- `./scripts/check-file-size.sh <changed files>`

Closes #26101


seven332:
## Code Review: PR #26141

**Reviewed head:** `b33aaabe40cde5b1e82b1aaf10a414f38f70f236`

### Summary

The change bounds managed SigV4 inspection before decoding, reuses only exact request representations across hook phases, and releases the flow-local artifact on signing and terminal paths. The self-review finding where current-authorization reclassification discarded the reusable artifact has been fixed on this head.

### Findings

No findings.

### Testing

Ruff formatting and lint, basedpyright, 123 focused tests, and the complete mitm-addon suite (5,224 tests) pass locally. Boundary, changed-representation, cancellation, failure-response, and cleanup paths are covered. Required GitHub checks are tracked separately.

### Verdict

LGTM

Issue #26185, opened 2026-08-10
refactor(connectors): remove candidate rollout compatibility layer


seven332:
## Background

#26077 introduced `connectorRuntimeCandidateTargets` as an additive complete target list while preserving filtered `connectorRuntimeTargets` and `firewalls` for runners that could not reconcile overlapping Builtin and Custom connectors. PR #26186 deployed the candidate-aware API and runner behavior. The duplicate representation is rollout debt rather than a permanent protocol.

## Problem and evidence

The current execution context has two target views:

- `connectorRuntimeCandidateTargets` contains every admitted Builtin and Custom registration for candidate-aware runners.
- `connectorRuntimeTargets` and `firewalls` retain a connector-wide filtered projection for previous runners.

The duplicate writer, contracts, selection branches, compact Builtin synthesis, fixtures, comments, and compatibility tests increase maintenance cost. Deleting them in one release is unsafe:

- pending `runner_job_queue.execution_context` rows remain claimable for up to two hours;
- rollout contexts can contain the complete target set only in the additive field;
- active runner jobs can continue for the two-hour execution budget plus bounded finalization;
- production promotes API before runner, and full-release rollback converges API and runner jobs concurrently.

A cleaned runner that reads a still-live rollout context only from the filtered fields can lose a suppressed Builtin connector. Unrelated Builtin APIs remain unavailable, and the Builtin cannot become effective after the winning Custom route narrows, is disabled, or is deleted.

At the 2026-08-11 planning check, runner `0.163.2` containing #26186 was serving production, but `prod-4` still had runner `0.163.1` draining active jobs. `prod-11` had no runner service and was absent from the release runner matrix.

## Decision

Use this issue as a two-PR delivery parent. First cut new writers over to the canonical existing fields while retaining tolerant readers. Then remove compatibility readers only after direct production evidence proves that no supported runner, pending context, or rollback target can require the rollout shape.

Each PR has its own sub-issue, is independently mergeable and deployable, and records its production prerequisite. The second child is blocked by the first.

Do not add a database migration, JSONB rewrite, trigger, execution-context schema version, or permanent on-read adapter. The queue data is short-lived, and staged natural drain reaches a cleaner end state with less concurrency and rollback risk.

## Gate 0: candidate-aware runner and rollback state

Before the first child is implemented or merged:

1. Derive every production runner host from the production deployment matrix.
[...]

PR #26186, opened 2026-08-10
fix(connectors): reconcile custom and builtin firewall candidates

### LINKED ISSUE #26077: fix(connectors): preserve firewall precedence across custom runtime updates
Parent: #25312

## Problem and evidence

Run creation and active runtime synchronization currently derive connector firewall state through
different models.

- Run creation removes a whole Builtin connector when any admitted Custom firewall base covers one
  of its API bases. The removed Builtin is then absent from the stored permission baseline and
  runtime target set.
- Active synchronization resolves targets independently. A Builtin result patches only its current
  network policy, while a Custom result replaces or removes only that Custom firewall and policy.
- The runner therefore has no candidate state from which to restore a Builtin suppressed at
  admission, and it does not reevaluate existing Builtins when a Custom definition begins covering
  them.

An active run can differ from a fresh run with the same pinned connector membership and routing
inputs. Removing or narrowing the winning Custom can leave an eligible Builtin unavailable;
widening a Custom can leave conflicting owners active. The mismatch can select the wrong credential
owner or network policy, so it is a correctness and security boundary rather than only stale
configuration.

## Decision

Fix the lifecycle model rather than adding a Custom-only restore hook.

Builtin and Custom connectors share one candidate-target lifecycle, scheduler, retry model, and
registry publication path. Their catalog/inline definition and credential readers remain small
source adapters. Effective Custom-over-Builtin ownership is resolved from the complete active
candidate snapshot at API-route granularity.

This remains one independently reviewable behavior PR. API contracts, runner state, proxy
resolution, integration coverage, and rollout compatibility are coupled by the behavior and belong
together. A later compatibility-cleanup PR has its own child issue and production drain gate.

### Additive candidate membership

- Add optional `connectorRuntimeCandidateTargets` to the persisted and runner-facing execution
  context. It contains every connector target admitted to the run.
- Keep existing `connectorRuntimeTargets` and `firewalls` as the filtered compatibility view for old
  runners during rollout.
- Extend Builtin target registration with optional pinned `baseUrlVars`, while Builtin target
  identity, result correlation, and realtime routing remain slug-only.
- Build policies, billing identities, permission baseline, credentials, and routing state for the
  complete admitted set. The existing admission coverage test derives only the legacy firewall and
  target view.

The rollout field stores target registrations only; it does not duplicate Custom firewall bodies or
the complete proxy registry. A new runner prefers the candidate list and reconstructs a missing
compact Builtin entry from its slug and pinned routing variables. A missing field means legacy
behavior. Custom definitions continue to arrive through their existing inline initial state and
authoritative runtime synchronization.

### Shared active reconciliation

- Normalize old and new execution contexts once at runner registration, then feed one target map to
  the existing shared scheduler.
- Convert Builtin available, Custom available, and Custom absent results into typed candidate
  updates. Unresolved or malformed results retain last-known-good state and retry.
- Validate results individually, then acquire the registry lock once, verify the current source IP
  and run, apply every accepted update in memory, validate ownership and size, and write once.
- A target-local validation failure retries that target without discarding valid siblings. A failed
  atomic write or changed run leaves every prepared update unpublished.

### Route-level effective ownership

- Serialize stable target identities into the runner-owned proxy VM entry. The paired mitm-addon
  derives an internal source marker only for a compact Builtin whose slug is registered or an inline
  Custom whose UUID is registered. Model-provider and unrelated inline firewalls remain neutral.
- When a request has at least one matching Custom connector API base, exclude only matching Builtin
  connector API candidates before the existing base specificity, rule specificity, connector
  intent, policy, and same-tier ambiguity reduction.
- Source precedence happens before permission reduction. A denied, malformed, or unknown Custom
  route cannot fall back to Builtin. A Builtin API that does not overlap remains available.
- Keep source metadata on the bound firewall shell rather than in the reusable Builtin compiled
  core, so existing catalog cache keys remain valid across VMs with different target membership.

This source marker is private runner/mitm state, not a new public firewall configuration field. The
public TypeScript/Python firewall-matching contract remains unchanged for inputs without that
private runtime ownership metadata.

## Compatibility and rollout

| Pairing | Behavior |
| --- | --- |
| Old API -> old runner | Existing filtered behavior. |
| New API -> old runner | Optional candidate field and Builtin metadata are ignored; filtered legacy firewalls/targets are used. |
| Old API -> new runner | Candidate field is absent, so the runner uses the legacy view. |
| New API -> new runner | Complete candidate membership is used and missing compact Builtins are reconstructed. |
| New writer -> rolled-back API | Old Zod readers strip additive JSON fields while persisted legacy fields remain valid. |

No relational schema or migration is required. Extra candidate policies are inert without a
matching firewall name, refresh deadlines for an unregistered old-runner target are ignored, and a
billable name is consulted only after that firewall is selected. These behaviors require explicit
compatibility tests. Runner and mitm-addon deploy as one artifact.

After the new runner is confirmed in production and previous runners have drained, the cleanup PR
will place all admitted targets/firewalls directly in the existing fields and remove the temporary
candidate field and legacy admission projection.

## Alternatives considered

- **API-authoritative full snapshots:** rejected because the API does not own runner last-known-good
  state; every wake-up would need all-target refresh or a second persisted active state machine.
- **Rust-side route projection:** rejected because it duplicates URL, catalog, rule, malformed-state,
  and compiled-cache semantics already owned by the mitm-addon.
- **Complete canonical snapshot plus legacy snapshot:** rejected because it duplicates every Custom
  firewall body during rollout when complete target identity plus compact Builtin reconstruction is
  sufficient.
- **Deletion-only restore metadata or intent-only routing:** rejected because it misses widening,
  narrowing, partial overlap, and atomic transitions; connector intent is not automatically present
  on every Custom request.

## Risks and mitigations

- **Security-sensitive precedence:** constrain it to registered Custom-versus-registered-Builtin API
  candidates and test denied/malformed routes, neutral firewall isolation, same-tier ambiguity, and
  auth-time revalidation.
- **Old-runner behavior:** retain filtered target/firewall fields and verify orphan policies,
  billing names, and refresh records cannot become effective.
- **Concurrent updates:** prepare per target but publish one checked registry snapshot; only mark
  generations successful after the write.
- **Registry resources:** synthesized Builtins are compact; the existing 16 MiB cap and fail-closed
  policy-growth reserve remain authoritative.
- **Compatibility debt:** document the exact deletion gate and track it in a linked child issue.

## Out of scope

- Changing connector grants, credential storage, or firewall-auth cache semantics unrelated to
  owner reconciliation.
- General user-configurable priority beyond Custom-over-Builtin.
- Dynamically admitting a target that was not pinned for the run.

## Acceptance criteria

- Every admitted Builtin and Custom target, including pinned routing inputs, survives independently
  of which candidate is initially effective.
- New registration and active reconciliation produce the same owner for the same candidate snapshot
  and current definitions.
- Custom equal, broader, or narrower route changes suppress only matching Builtin API routes;
  unrelated Builtin APIs and targets stay available.
- Widening, narrowing, disabling, deleting, and restoring a pinned Custom transitions active
  ownership without restarting the run.
- Builtin and Custom results use one shared candidate-update publication entry point; no
  deletion-only restoration branch exists.
- Unresolved refresh preserves LKG, authoritative absence removes only that candidate, same-tier
  ambiguity remains fail closed, and no unpinned target is dynamically admitted.
- Each accepted sync batch produces at most one atomic registry write and exposes no partially
  switched owner set.
- Tests cover overlap in both directions, partial multi-API isolation, variable-backed routing,
  ambiguity, malformed/denied state, billing, catalog omission, post-auth revalidation, and every
  old/new API-runner pairing.
- Applicable TypeScript, Rust, Python, integration, static-analysis, formatting, self-review, and CI
  checks pass.

seven332:
## Summary

- preserve every admitted Builtin and Custom connector as a runtime candidate while retaining the filtered legacy runner view during rollout
- normalize old and new execution contexts in the runner, publish validated Builtin/Custom runtime changes through one atomic registry batch, and retain last-known-good retry behavior
- resolve Custom-versus-Builtin ownership per matching API route in the mitm addon, without changing neutral or same-tier ambiguity behavior
- cover overlap, partial multi-API ownership, dynamic widen/narrow/delete/restore, billing, stale-auth revalidation, registration conflicts, and the old/new API-runner deployment matrix

- The execution-context change is additive and requires no relational database migration.
- Old runners continue to consume filtered `firewalls` and `connectorRuntimeTargets`; new runners prefer `connectorRuntimeCandidateTargets` and fall back when it is absent.
- Runner and mitm addon ship together, so the private candidate marker has no independent deployment boundary.
- Cleanup of the temporary rollout field and legacy projection is intentionally deferred to #26185, after the new runner is fully deployed and older runners have drained.

## Validation

- `scripts/prepare.sh`
- `cd turbo && pnpm format`
- API/contracts lint and type checks
- `cd turbo && pnpm -F @vm0/api-contracts exec vitest run` (553 passed on current head)
- affected connector plus intersecting MCP admission scenarios in `run-lifecycle.bdd.test.ts` (4 passed on current head)
- legacy queued-context claim through the API/database integration path (1 passed on current head)
- `cd turbo && pnpm knip`
- runner formatting, Clippy, and documentation checks
- runner registry tests (22 passed), runtime-sync tests (40 passed), and proxy-registration tests (2 passed) on current head
- mitm-addon uv lock/sync, Ruff, basedpyright, and full pytest suite (5,234 passed); the three affected files also passed all 43 tests after rebase

Before the final rebase, the full local API suite reached 2,862/2,864 passing; the two failures were pre-existing dev-seed `vm0_api_keys` uniqueness conflicts in `chat-events.bdd.test.ts`, while all affected connector scenarios passed. Two full high-concurrency runner-suite attempts also exposed different existing timing/port/cache flakes; every reported failure passed immediately in isolation, and all affected runner modules passed.

Closes #26077

Parent objective: #25312

### DIFF
--- a/crates/runner/mitm-addon/src/connector_runtime_metadata.py
+++ b/crates/runner/mitm-addon/src/connector_runtime_metadata.py
@@ -1,0 +1,23 @@
+"""Private connector-candidate metadata shared by registry resolution and matching."""
+
+from typing import Literal
+
+type ConnectorRuntimeKind = Literal["builtin", "custom"]
+
+CONNECTOR_RUNTIME_KIND_MARKER = "_vm0ConnectorRuntimeKind"
+
+
+def connector_runtime_kind(firewall: dict) -> ConnectorRuntimeKind | None:
+    """Return a trusted internal candidate kind, if registry resolution assigned one."""
+    value = firewall.get(CONNECTOR_RUNTIME_KIND_MARKER)
+    return value if value in ("builtin", "custom") else None
+
+
+def clear_connector_runtime_kind(firewall: dict) -> None:
+    """Remove untrusted source metadata before registry-owned classification."""
+    firewall.pop(CONNECTOR_RUNTIME_KIND_MARKER, None)
+
+
+def mark_connector_runtime_kind(firewall: dict, kind: ConnectorRuntimeKind) -> None:
+    """Mark a resolved firewall as a registered connector runtime candidate."""
+    firewall[CONNECTOR_RUNTIME_KIND_MARKER] = kind
--- a/crates/runner/mitm-addon/src/matching.py
+++ b/crates/runner/mitm-addon/src/matching.py
@@ -16,6 +16,7 @@ from typing import Literal, NamedTuple
 from urllib.parse import urlsplit
 
 import connector_intent
+import connector_runtime_metadata
 import connector_template_syntax
 from firewall_auth_config import auth_config_injects_ordinary_upstream_credentials
 from firewall_matching import base_url as _firewall_base_url
@@ -182,6 +183,7 @@ class CompiledFirewallCore(NamedTuple):
 class _CompiledFirewall(NamedTuple):
     core: CompiledFirewallCore
     apis: tuple[_CompiledApi, ...]
+    connector_runtime_kind: connector_runtime_metadata.ConnectorRuntimeKind | None
 
     @property
     def name(self) -> str:
@@ -948,9 +950,10 @@ def bind_compiled_firewall_core(
     even when the target list still has a dictionary at that index.
 
     Fields excluded from core compilation may remain shell-local. Builtin reuse currently rebinds
-    generated API ``id`` values and snapshot-owned ``_builtinHostPolicyRuntime`` metadata. Such raw
-    fields remain subject to their own validity and lifecycle contracts; in particular, runtime
-    host-policy metadata must stay paired with the target shell's resolved registry snapshot.
+    generated API ``id`` values, snapshot-owned ``_builtinHostPolicyRuntime`` metadata, and the
+    registry-owned connector candidate kind. Such raw fields remain subject to their own validity
+    and lifecycle contracts; in particular, runtime host-policy metadata must stay paired with the
+    target shell's resolved registry snapshot.
 
     The caller or its cache key must enforce this compatibility before reuse. This function does
     not compare semantic content. It returns ``None`` only when the target ``apis`` value is not a
@@ -972,7 +975,11 @@ def bind_compiled_firewall_core(
 
     if not compiled_apis:
         return None
-    return _CompiledFirewall(core, tuple(compiled_apis))
+    return _CompiledFirewall(
+        core,
+        tuple(compiled_apis),
+        connector_runtime_metadata.connector_runtime_kind(fw_entry),
+    )
 
 
 def compile_firewalls(vm_firewalls: object | None) -> CompiledFirewallSet | None:
@@ -1704,6 +1711,7 @@ def _match_compiled_firewall_request_with_api_candidates(
     is_asterisk_form: bool,
 ) -> FirewallAllow | FirewallPolicyAllow | FirewallBlock | FirewallAmbiguous | None:
     collection = _FirewallMatchCollection()
+    base_matches: list[_MatchedApi] = []
     unsafe_path: bool | None = False if is_asterisk_form else (True if url_has_backslash else None)
     decision_path = "*" if is_asterisk_form else (url_parts.path or "/")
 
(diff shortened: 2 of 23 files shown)


seven332:
## Code Review: PR #26186

**Reviewed head:** `93039cc0d4beae71db7d607ea75646ab2a6f39b3`

### Summary

The change keeps all admitted Builtin and Custom connectors as candidates, reconciles both source kinds through shared runtime publication and route matching, and preserves the legacy runner projection during the independent deployment window. The runner/mitm boundary remains atomic and fail-closed, and cleanup is tracked separately in #26185.

### Findings

No findings.

Two review-time coverage gaps were fixed on this PR: executor registration now rejects conflicting routing/source ownership through its production entry point, and the API claim path now proves compatibility with queued contexts written before the additive candidate field existed.

### Testing

- API and API-contracts lint/type checks
- API-contracts suite: 553 passed
- affected run-lifecycle integration scenarios, including legacy queued-context claim
- runner formatting, Clippy, documentation, registry/runtime-sync, and proxy-registration tests
- mitm-addon Ruff, basedpyright, and full pytest suite: 5,234 passed
- staged Knip and repository TypeScript type checks

### Verdict

LGTM

PR #26187, opened 2026-08-10
fix(chat): canonicalize automation-tagged goal rows in backfill

e7h4n:
Unblocks the production release train: the `0886_backfill_canonical_chat_event_storage` migration aborted in the release pipeline's Production Migration Smoke Test (release-please run 31404752899, `promote-api-production`) with:

```
PostgresError: chat_events has 18300 goal-grouped rows whose context conflicts with run_group_id
```

The abort worked as designed — production data was never touched and the API was not promoted — but every subsequent release hits the same abort until the backfill handles this class.

## Root cause (maskdb evidence, read-only)

All 18,300 conflicting rows are one uniform historical class: `event_type = 'input.prompt'`, `context_type = 'automation'`, **`context_id = NULL` on every row**, with a non-null `run_group_id` (goal-thread continuations tagged with a bare automation source before goal context pointers existed). The deployed dual-writer's semantics for such rows are unambiguous: `persistedChatEventValues` replaces any display context with the canonical `('goal', runGroupId)` pointer whenever `run_group_id` is present. Treating the bare source tag as a hard conflict was over-strict; nothing is lost by adopting the goal pointer (`context_id` is NULL on every affected row), and the v3 wire/UI behavior is unchanged either way (masking hides canonical goal contexts; `runGroupId` keeps being emitted).

## Changes

- **Migration `0886` (in-place edit)**: the goal-context pre-check now aborts only on genuine conflicts — a `'goal'` context whose non-null `context_id` names a *different* goal id. Rows with any non-goal context (or an incomplete goal pointer) canonicalize to `('goal', run_group_id)`, exactly matching the dual-writer; the narrowed append-only trigger, batch predicate, and restart pass follow the same rule, and a conflicting goal pointer stays immutable so the unchanged final assertions still fail loudly on it. Editing the merged file is deliberate: 0886 has never applied to production (the smoke abort precedes any write), the chain cannot advance past it otherwise, and no DDL changes are involved so the drizzle snapshot chain is untouched.
- **`canonicalChatEventRow` (api-contracts)**: mirrors the same rule so the client normalizer keeps goal grouping (`runGroupId` projection) for this class instead of preserving the stale source tag.
- **Tests**: consistency-harness validator seeds the exact production shape (automation-tagged goal continuation → canonicalized) and re-targets the abort scenario to a true goal-id mismatch; contract tests cover the automation-tagged normalization, the preserved goal-id conflict, and the end-to-end `runGroupId` projection for the historical class.

## Verification

- `pnpm test:migration-consistency` — all validations green, including the new automation-tagged canonicalization and the goal-id-mismatch abort with clean rerun.
- Full-chain `db:migrate` on a fresh database.
- api-contracts tests 8/8; platform snapshot-read + IDB tests 23/23; format/lint/check-types green on api-contracts, db, app, api.

### DIFF
--- a/turbo/packages/api-contracts/src/contracts/__tests__/chat-event-rows.test.ts
+++ b/turbo/packages/api-contracts/src/contracts/__tests__/chat-event-rows.test.ts
@@ -121,7 +121,7 @@ describe("chat event row reader union", () => {
     expect(owned.runId).toBe(target);
   });
 
-  it("completes compatible goal context pointers from runGroupId", () => {
+  it("adopts the canonical goal context on goal-grouped rows", () => {
     const goalId = "00000000-0000-4000-8000-000000000011";
     const grouped = canonicalChatEventRow(v3Row({ runGroupId: goalId }));
     expect(grouped.contextType).toBe("goal");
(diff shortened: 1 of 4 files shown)


e7h4n:
Closing as superseded. PR2's production rollout completed successfully, and PR3 completed the canonical-read cutover. This PR also modifies migration 0886 after it landed and ran in production, which violates the append-only migration policy. Any newly reproduced cleanup should use a separate PR and a new migration.

Issue #26015, opened 2026-08-10
epic: retire workflow-event/workflow-schedule in favour of automation-event/automation-schedule, and give goal runs their own trigger source


lancy:
## Goal

Retire `workflow-event` as the name for "an automation fired". The domain object is an **Automation** (`zero_workflow_automations.kind = 'event'`); a *workflow* is the SKILL.md that gets run. Every layer that calls the trigger a "workflow event" sits one indirection behind the model, and the product already disagrees with the code — the composer renders `chat.queue.automationEvent` ("Automation event") for a queue item whose TypeScript literal is `"workflow-event"`.

Three things ship together, because they all rewrite the same enum through the same expand/migrate/contract window:

1. `workflow-event` → `automation-event`
2. `workflow-schedule` → `automation-schedule` (leaving one half renamed is worse than either consistent pair)
3. **Bug fix:** goal continuation runs stop claiming to be automation events and get their own `goal` trigger source

Otherwise naming only. No new capability.

Baseline: `main` @ `ecdb01a6d8e8e9e7eac7bda3e96e76d5a4c6b21c` (2026-08-10). 682 raw matches for `workflow[-_]?event` / `WorkflowEvent` outside `node_modules`.

## Target vocabulary

| concept | today | after |
| --- | --- | --- |
| a `kind='event'` automation row | "event workflow automation" | **event automation** |
| one firing of it | `workflow-event` / `workflow_event` / `WorkflowEvent` | **automation event** / `automation-event` / `automation_event` / `AutomationEvent` |
| run provenance, event automation | `trigger_source = 'workflow-event'` | `'automation-event'` |
| run provenance, schedule automation | `trigger_source = 'workflow-schedule'` | `'automation-schedule'` |
| run provenance, goal continuation | `trigger_source = 'workflow-event'` ← wrong | `'goal'` |

## The goal-run mislabel

`zero-goal-queue-drain.service.ts:166` launched goal continuation runs with `triggerSource: "workflow-event"`. A goal is not an automation and carries no `automationId`.

[...]

PR #26217, opened 2026-08-11
chore(db): backfill historical trigger sources to the automation vocabulary

### LINKED ISSUE #26153: chore(db): backfill historical trigger sources to the automation vocabulary
Step 5a of #26015. Backfill the historical `trigger_source` / `source` rows to the new vocabulary. **This is the only irreversible step in the epic** — the goal split cannot be recovered once the blanket rename runs.

Written against `main` @ `7c1d422`, after step 4 (#26112 → PR #26115) went live and production stopped writing the legacy values.

## The gate is met

Measured against the masked production replica:

- **Last legacy row written at 09:47:24.299Z** on 2026-08-10. Nothing since; verified clean over the following 30 minutes.
- **`automation-event` is live from 10:06:04Z**, alongside `automation-schedule` and `goal`. All twelve event-automation writer sites are exercised in production.

Both conditions from #26112's acceptance note hold, so no draining instance is still emitting the old values and a backfill cannot be overtaken.

## Volumes (measured 2026-08-10 ~10:20Z)

| table | value | rows |
| --- | --- | --- |
| `zero_runs` | `workflow-event` | **27,706** (20,827 goals + 6,879 automation events) |
| `zero_runs` | `workflow-schedule` | **8,910** |
| `run_uploaded_files` | `workflow-event` | ~35 |
| `run_uploaded_files` | `workflow-schedule` | ~1,435 |

`zero_runs` holds ~106,786 rows in total, so a sequential scan touching ~36.6k of them is cheap. **Do not batch** — this does not need the chunked pattern of `0846_backfill_chat_run_model_annotations.sql`, and a single transaction gives all-or-nothing semantics that batching would give up. Write it as an ordinary transactional migration; do **not** mark it `-- vm0:non-transactional`.

`trigger_source` is `varchar(20)` with no CHECK constraint, `source` is `varchar(32)`; the longest new value is 19 characters. No schema change.

## The trap: statement order is load-bearing

`zero_runs.goal_id` is the only thing that distinguishes a goal continuation from a real automation event in the historical rows. Once `trigger_source = 'workflow-event'` has been blanket-renamed to `'automation-event'`, **that distinction is gone and cannot be reconstructed** — `goal_id` alone cannot tell you which of those rows were mislabelled versus which were correctly automation events, because after the rename both look identical.

The goal statement must run **first**. Getting this backwards permanently merges ~20.8k goal runs — three quarters of that population into the automation-event population.

## The pre-flight has already been run — and it changed the numbers

Both predicates were measured against production on 2026-08-10 after `zero_runs.goal_id` was exposed in the maskdb policy:

| predicate | count |
| --- | --- |
| **A** — `trigger_source = 'workflow-event' AND goal_id IS NOT NULL` | **20,827** |
| **B** — same, via `EXISTS (chat_events WHERE run_id = r.id AND context_type = 'goal')` | 3,675 |

They diverge 5.7×, and **A is the correct one.** `chat_events.context_type` is a recent discriminator: its earliest `'goal'` row is 2026-07-29, and 1,128,479 `chat_events` rows still have `context_type IS NULL`. B can only see the last ~12 days of goal history, so it structurally under-counts.

A does not over-select either. `zero_runs.goal_id` is written only through `zeroRunMetadata.goalId`, and the sole run-creating writer is `zero-goal-queue-drain.service.ts:185` (`zero-goal.service.ts:588` is a goal-status update, not a run insert). No automation launch path sets it.

**Use `goal_id IS NOT NULL`. Do not re-run the A/B comparison as a gate — it will always "fail", and the reason is understood.**

This corrects the epic's original estimate. The goal mislabel is not ~12% of the `workflow-event` population; it is **75%**:

| after the backfill | rows |
| --- | --- |
| `goal` | ~20,827 |
| `automation-event` | ~6,879 |
| `automation-schedule` | ~8,910 |

Usage reporting is unaffected — step 3 put all five values in the `automation` bucket precisely so a provenance correction of this size could not move anyone's numbers.

## Migration

Four statements, in this order. Assign the next available migration number after `main` at PR time; never hardcode one from an issue.

```sql
-- 1. Goals first. This predicate only works while 'workflow-event' still
--    distinguishes them; step 2 destroys it.
UPDATE "zero_runs" SET "trigger_source" = 'goal'
WHERE "trigger_source" = 'workflow-event'
  AND "goal_id" IS NOT NULL;                                    -- ~20,827

-- 2. Everything still labelled workflow-event is a real automation event.
UPDATE "zero_runs" SET "trigger_source" = 'automation-event'
WHERE "trigger_source" = 'workflow-event';                      -- ~6,879

UPDATE "zero_runs" SET "trigger_source" = 'automation-schedule'
WHERE "trigger_source" = 'workflow-schedule';                   -- ~8.9k

-- 3. run_uploaded_files.source mirrors the run's trigger source, so derive it
--    from the already-corrected zero_runs rather than repeating the split.
UPDATE "run_uploaded_files" f SET "source" = r."trigger_source"
FROM "zero_runs" r
WHERE f."run_id" = r."id"
  AND f."source" IN ('workflow-event', 'workflow-schedule')
  AND r."trigger_source" IN ('automation-event', 'automation-schedule', 'goal');

-- 4. run_id is nullable and the run may have been deleted, so any row the join
--    could not reach gets the blanket rename. These cannot be classified as
--    goals — there is no run left to ask.
UPDATE "run_uploaded_files" SET "source" = 'automation-event'
WHERE "source" = 'workflow-event';
UPDATE "run_uploaded_files" SET "source" = 'automation-schedule'
WHERE "source" = 'workflow-schedule';
```

Every statement is idempotent: each `WHERE` matches only legacy values, so re-running is a no-op. That matters because **5b must re-run this backfill before dropping the enum values** — rows written by an old instance during the deploy window would otherwise be stranded.

## Non-goals

- **Do not remove `'workflow-event'` / `'workflow-schedule'`** from `triggerSourceSchema` (`contracts/logs.ts:58-59`) or `RUN_UPLOADED_FILE_SOURCES` (`schema/run-uploaded-file.ts:25-26`), and do not touch the four reader sites. That is 5b, and it must ship in a later release than this migration.
- **Do not change any writer.** Step 4 has shipped.
- **No schema change** — no column, type, constraint or index changes.
- **Do not batch** and do not mark the migration non-transactional.
- **Do not touch `packages/db/scripts/test-migration-consistency-schema.ts:3676-3677,3781`**, which seeds legacy values through raw SQL. It is 5b's cleanup.
- **Do not change the dispatch-timing strings** — step 6.

## Acceptance criteria

- The migration contains the four steps in the order above, with the goal statement first.
- The PR description carries the post-migration counts and compares them with the expected split above.
- After the migration runs in production: `SELECT count(*) FROM zero_runs WHERE trigger_source IN ('workflow-event','workflow-schedule')` returns **0**, and the same for `run_uploaded_files.source`.
- `SELECT count(*) FROM zero_runs WHERE trigger_source = 'goal'` is roughly 20,827 plus the rows written by the live writers since 10:06Z (91 at the time of writing).
- No row moved out of the `automation` usage bucket: the five values all still map there, so per-org usage totals are unchanged before and after. Spot-check one org.
- `git diff --stat` shows one new migration file and its snapshot, and nothing else under `packages/db/src/` — no schema edit.
- No change to `contracts/logs.ts`, `packages/core/`, or any service.

lancy:
## Summary

- add transactional migration `0890_backfill_trigger_sources`
- classify historical goal continuations before the irreversible blanket rename
- derive uploaded-file sources from their corrected runs, then rename orphaned legacy rows

## Production evidence and expected post-migration counts

The production pre-flight classified the 27,706 historical `workflow-event` runs as:

| classification | predicate | historical rows moved |
| --- | --- | ---: |
| `goal` | `trigger_source = 'workflow-event' AND goal_id IS NOT NULL` | 20,827 |
| `automation-event` | remaining `workflow-event` rows | 6,879 |
| `automation-schedule` | `trigger_source = 'workflow-schedule'` | 8,910 |

`chat_events.context_type = 'goal'` is not a valid full-history gate: the discriminator starts on 2026-07-29 while 1,128,479 historical chat events still have `context_type IS NULL`. The authoritative provenance is `zero_runs.goal_id`, which is only set by the goal continuation run writer.

After this migration runs:

- `zero_runs.trigger_source IN ('workflow-event', 'workflow-schedule')`: expected 0
- `run_uploaded_files.source IN ('workflow-event', 'workflow-schedule')`: expected 0
- historical contribution to `goal`: +20,827, on top of rows already emitted by the live `goal` writer
- historical contribution to `automation-event`: +6,879
- historical contribution to `automation-schedule`: +8,910
- usage totals remain unchanged because all old and new values stay in the existing `automation` bucket

## Safety

- the goal update is the first statement, before any blanket `workflow-event` rename
- one ordinary transaction; no batching and no `vm0:non-transactional` marker
- every update is idempotent because it only matches legacy values
- no schema, enum, reader, writer, service, or dispatch-timing changes

## Validation

- `pnpm exec prettier --check packages/db/src/migrations/meta/_journal.json packages/db/src/migrations/meta/0890_snapshot.json`
- `pnpm -F @vm0/db lint`
- `DATABASE_URL=postgresql://postgres:postgres@localhost:5432/postgres pnpm -F @vm0/db test:migration-consistency`
- `git diff --check`

Closes #26153
Part of #26015

### DIFF
--- a/turbo/packages/db/src/migrations/0890_backfill_trigger_sources.sql
+++ b/turbo/packages/db/src/migrations/0890_backfill_trigger_sources.sql
@@ -1,0 +1,29 @@
+-- 1. Goals first. This predicate only works while 'workflow-event' still
+--    distinguishes them; step 2 destroys it.
+UPDATE "zero_runs" SET "trigger_source" = 'goal'
+WHERE "trigger_source" = 'workflow-event'
+  AND "goal_id" IS NOT NULL;--> statement-breakpoint
+
+-- 2. Everything still labelled workflow-event is a real automation event.
+UPDATE "zero_runs" SET "trigger_source" = 'automation-event'
+WHERE "trigger_source" = 'workflow-event';--> statement-breakpoint
+
+UPDATE "zero_runs" SET "trigger_source" = 'automation-schedule'
+WHERE "trigger_source" = 'workflow-schedule';--> statement-breakpoint
+
+-- 3. run_uploaded_files.source mirrors the run's trigger source, so derive it
+--    from the already-corrected zero_runs rather than repeating the split.
+UPDATE "run_uploaded_files" f SET "source" = r."trigger_source"
+FROM "zero_runs" r
+WHERE f."run_id" = r."id"
+  AND f."source" IN ('workflow-event', 'workflow-schedule')
+  AND r."trigger_source" IN ('automation-event', 'automation-schedule', 'goal');--> statement-breakpoint
+
+-- 4. run_id is nullable and the run may have been deleted, so any row the join
+--    could not reach gets the blanket rename. These cannot be classified as
+--    goals -- there is no run left to ask.
+UPDATE "run_uploaded_files" SET "source" = 'automation-event'
+WHERE "source" = 'workflow-event';--> statement-breakpoint
+
+UPDATE "run_uploaded_files" SET "source" = 'automation-schedule'
+WHERE "source" = 'workflow-schedule';
--- a/turbo/packages/db/src/migrations/meta/0890_snapshot.json
+++ b/turbo/packages/db/src/migrations/meta/0890_snapshot.json
@@ -1,0 +1,26896 @@
+{
+  "id": "d5181a56-f70d-43e2-875a-5e9e9dc0662b",
+  "prevId": "7afd5f43-ea8f-4b88-9160-0d66dd5bc945",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.active_input_deliveries": {
+      "name": "active_input_deliveries",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "uuid",
+          "primaryKey": true,
+          "notNull": true,
+          "default": "gen_random_uuid()"
+        },
+        "run_id": {
+          "name": "run_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "chat_thread_id": {
+          "name": "chat_thread_id",
+          "type": "uuid",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "status": {
+          "name": "status",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "'open'"
+        }
+      },
+      "indexes": {
+        "active_input_deliveries_run_open_unique": {
+          "name": "active_input_deliveries_run_open_unique",
+          "columns": [
+            {
+              "expression": "run_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": true,
+          "with": {},
+          "method": "btree",
+          "where": "\"active_input_deliveries\".\"status\" = 'open'",
+          "concurrently": false
+        },
+        "active_input_deliveries_thread_open_idx": {
+          "name": "active_input_deliveries_thread_open_idx",
+          "columns": [
+            {
+              "expression": "chat_thread_id",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": false,
+          "with": {},
+          "method": "btree",
+          "where": "\"active_input_deliveries\".\"status\" = 'open'",
+          "concurrently": false
+        }
+      },
+      "foreignKeys": {
+        "active_input_deliveries_run_id_agent_runs_id_fk": {
+          "name": "active_input_deliveries_run_id_agent_runs_id_fk",
+          "tableFrom": "active_input_deliveries",
+          "columnsFrom": [
+            "run_id"
+          ],
+          "tableTo": "agent_runs",
+          "columnsTo": [
+            "id"
+          ],
+          "onUpdate": "no action",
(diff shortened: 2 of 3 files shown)


lancy:
## Code Review: PR #26217 (Round 1) — LGTM :tada:

No P0 or P1 issues remain.

### Review summary

- verified the first SQL statement classifies `workflow-event` rows with `goal_id IS NOT NULL` as `goal` before any blanket rename
- verified the remaining run and uploaded-file updates preserve the issue-mandated order and are idempotent
- verified the migration is transactional, unbatched, and has no `vm0:non-transactional` marker
- verified the diff contains only the new migration, generated snapshot, and journal entry
- verified the generated snapshot has no schema delta from 0889 beyond the snapshot chain identifiers
- verified no enum, reader, writer, service, schema, or dispatch-timing source changed

### Testing review

Coverage is adequate for a data-only migration: DB lint, type checking, formatting, and the full migration-consistency validator passed, including migration replay, reset, latest snapshot accuracy, and normalized schema equivalence. No dedicated fixture was added because the issue explicitly reserves the existing legacy-value fixture cleanup for step 5b.

The irreversible classification predicate and statement order match issue #26153. This PR is ready for CI and merge.

---
*Completed after 1 review round*

PR #26245, opened 2026-08-11
fix: keep scroll-to-bottom button opaque on hover

tongx-blip:
## Problem

The chat scroll-to-bottom button turns nearly invisible on hover — the disc goes see-through and the chat text underneath shows straight through it, while the border, shadow and arrow stay put.

## Root cause

`ScrollToBottomButton` used `bg-background hover:bg-state-hover`.

`--color-state-hover` is a **5% translucent state layer**, not a solid colour. Both classes set `background-color`, so on hover it **replaces** the opaque `bg-background` fill rather than sitting on top of it, leaving the button 95% transparent. Because the button is `absolute`-positioned over the message list, what shows through is the conversation.

The design system already calls this out in `globals.css`:

> Elements that do carry an opaque fill cannot use a translucent layer (the layer would replace the fill instead of sitting on it), so each fill gets its own pre-mixed pair below.

`card`, `input` and `secondary` all have pre-mixed `-hover` pairs — `background` was the one surface missing one.

## Fix

Add the missing `--color-background-hover` token using the exact same `color-mix` form as its siblings, and point the button at it.

## Verification

Compiled the token through Tailwind v4 in isolation:

- `hover:bg-background-hover` emits `background-color: var(--color-background-hover)`; no name collision with the existing `--color-background-50`
- Light: `rgb(250,251,252)` → `rgb(238,240,244)` — **fully opaque**, ~12/255 darkening, a clear but restrained hover
- Dark: `rgb(25,25,26)` → `rgb(40,40,42)` — inverts to lighter automatically
- Tailwind also emits a `@supports` fallback (`hsl(var(--background))`) for engines without `color-mix`, which is likewise opaque — worst case is no hover effect, never a transparent button

No tests referenced the old class.

## Note for reviewers

Four other floating controls share this root cause but use `bg-background/90 + backdrop-blur`, so they are *intentionally* translucent and the correct hover treatment is a design call rather than a mechanical swap. Left untouched here:

- `zero-artifact-sidebar.tsx:1210,1226` and `zero-attachment-chips.tsx:564,580` — carousel arrows
- `agent-chat-page.tsx:302` — attachment remove badge

The many other `bg-background hover:bg-state-hover` sites in the codebase render correctly today because they sit directly on the page background, so the 5% layer composites over the same colour it was mixed against.

### DIFF
--- a/turbo/apps/platform/src/views/zero-page/zero-chat-thread-page.tsx
+++ b/turbo/apps/platform/src/views/zero-page/zero-chat-thread-page.tsx
@@ -4265,7 +4265,7 @@ function ScrollToBottomButton({ thread }: { thread: ChatPanelSignals }) {
       onClick={() => {
         scrollToBottom();
       }}
-      className="absolute bottom-4 left-1/2 z-20 flex h-9 w-9 -translate-x-1/2 items-center justify-center rounded-full border border-border bg-background text-muted-foreground shadow-md transition-colors hover:bg-state-hover hover:text-foreground"
+      className="absolute bottom-4 left-1/2 z-20 flex h-9 w-9 -translate-x-1/2 items-center justify-center rounded-full border border-border bg-background text-muted-foreground shadow-md transition-colors hover:bg-background-hover hover:text-foreground"
     >
       <ArrowDown size={18} />
     </button>
--- a/turbo/packages/ui/src/styles/globals.css
+++ b/turbo/packages/ui/src/styles/globals.css
@@ -149,6 +149,11 @@
   );
   --color-state-pressed: hsl(var(--state-layer) / var(--state-pressed-alpha));
 
+  --color-background-hover: color-mix(
+    in srgb,
+    hsl(var(--background)),
+    hsl(var(--state-layer)) var(--state-hover-alpha)
+  );
   --color-card-hover: color-mix(
     in srgb,
     hsl(var(--card)),


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: ec5995e9d019e0fbd87bfc8351da814c06a19f4f
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Adds the missing `--color-background-hover` pre-mixed design token and switches the chat scroll-to-bottom button from `hover:bg-state-hover` to `hover:bg-background-hover`. `--color-state-hover` is a 5% translucent state layer, so it replaced the button's opaque `bg-background` fill on hover instead of compositing over it, leaving the floating button ~95% transparent with chat text bleeding through. Two files, +6/-1, no logic change.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic — no wrong assumptions or API misuse | Pass | Token is inside the `@theme` block (`globals.css:15` opens it, token at `:152`), so Tailwind v4 emits the `bg-background-hover` utility; platform consumes it via `@import "@vm0/ui/styles/globals.css"` (`views/css/index.css:5`). Construction is identical to the in-production `--color-card-hover`. |
| `REVIEW.md` | Security — injection, secrets, missing auth | N/A | Pure CSS token + className change; no input handling, network, or auth surface. |
| `REVIEW.md` | Type safety — no `any`, no broad types | N/A | No TypeScript types added or changed; the only TS edit is a string literal in `className`. |
| `REVIEW.md` | Style & maintainability — no duplication, no oversized functions | Pass | Reuses the existing `color-mix` form of its `card`/`input`/`secondary` siblings and is placed adjacent to them; `ScrollToBottomButton` is ~30 lines. |
| `REVIEW.md` | Testing coverage — `fix:` requires a regression test | N/A | See Testing; the change has no JS-observable behavior and jsdom does not compute Tailwind utilities or `color-mix`. Reasoned exception, flagged for human override. |
| `REVIEW.md` | Feature regression guard — existing behavior not silently changed | Pass | Only the hover fill of one button changes; rest-state appearance, `onClick`, `aria-label`, and visibility conditions are untouched. Grep confirms no test pins the old `bg-state-hover` class, and `bg-background-hover` had no prior consumer. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, contract, Zod, or `jsonb()` changes. |
| `docs/bad-smell.md` | Zero lint/type suppression comments | Pass | No `eslint-disable`, `@ts-ignore`, `@ts-expect-error`, or `prettier-ignore` in the diff. |
| `docs/bad-smell.md` | No unnecessary try/catch; fail fast | N/A | No error-handling code in the diff. |
| `docs/bad-smell.md` | No dynamic `import()` | Pass | No imports changed. |
| `docs/bad-smell.md` | No hardcoded URLs or environment values | Pass | Colour is derived from existing `--background` / `--state-layer` / `--state-hover-alpha` tokens, not a literal hex. |
| `docs/fallback.md` | Default is no fallback; contract-guaranteed states need no `??`/`||` | Pass | Diff adds no `??`/`||` chain, compatibility branch, or tolerance path. |
| `docs/fallback.md` | Every introduced fallback declared in the PR summary | N/A | The diff introduces no fallback, so no `Fallbacks` section is required. The `@supports (color: color-mix(...))` guard in the compiled CSS is emitted by the Tailwind compiler, not authored here. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | No tests added or removed. |
| `docs/testing.md` | Integration tests only; no unit tests for internal functions | Pass | No tests added; none of the prohibited unit-test shapes appear. |
| `docs/testing.md` | Mock at the boundary; no MSW/fetch/db/fake-timer violations | N/A | No test files in the diff. |
| `docs/testing.md` | Do not over-test non-logic UI presentation | Pass | Supports not adding a className-string assertion for a purely visual token change. |
| `docs/testing/app-testing.md` | Page-level tests via `detachedSetupPage`, asserting page-visible behavior; no component-only unit tests | N/A | No platform test added; a hover fill computed from Tailwind + `color-mix` is not page-visible under jsdom, and asserting the raw class string would be the component-only implementation-detail test this doc prohibits. |
| `docs/effect.md` | React render stays pure — no Store writes, DOM mutation, timers, or subscriptions during render | Pass | Diff changes only a static `className` literal; no render-phase work added. |
| `docs/cache.md` | RC-1 no render-phase side effects; RC-2 no render-time signal allocation; RC-3 no unbounded caches; RC-4 correct owner/lifetime | Pass | No signals, caches, refs, or resources touched; `useGet`/`useSet` usage in `ScrollToBottomButton` is unchanged. |
| `.claude/skills/ccstate/SKILL.md` | DOM async commands use `detach(..., Reason.DomCallback)` | N/A | The `onClick` handler is pre-existing and unmodified by this diff. |
| `docs/deployment-compatibility.md` | Cross-version safety at deployable boundaries | Pass | Frontend-only styling change with no API, runner, queue, or persisted-state boundary. An open tab keeps its already-loaded CSS until refresh; both old and new bundles are self-consistent because the token and its only consumer ship in the same build. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Notes (non-blocking)

- `turbo/packages/ui/src/styles/globals.css`: the token correctly inverts for dark mode without a `.dark` counterpart, because `color-mix` resolves `var(--background)` and `var(--state-hover-alpha)` lazily per element (light `rgb(250,251,252)` → `rgb(238,240,244)`; dark `rgb(25,25,26)` → `rgb(40,40,42)`). This matches how `--color-card-hover` already behaves, so no extra declaration is needed.
- `turbo/apps/platform/src/views/zero-page/zero-chat-thread-page.tsx`: four other controls that float above content share the same root cause but use `bg-background/90 + backdrop-blur`, so they are intentionally translucent and their correct hover treatment is a design decision rather than a mechanical swap — `zero-artifact-sidebar.tsx:1210,1226`, `zero-attachment-chips.tsx:564,580`, `agent-chat-page.tsx:302`. The PR summary already calls these out and correctly leaves them for a separate change. The remaining `bg-background hover:bg-state-hover` sites render correctly today because they sit directly on the page background, where the 5% layer composites over the same colour it was mixed against.

### Testing

- **Coverage**: No automated regression test, and this is a deliberate exception to the `fix:`-needs-a-regression-test rule rather than an oversight. The change carries no JS-observable behavior: it swaps one static `className` literal and adds a CSS custom property. `turbo/apps/platform` tests run in jsdom, which neither compiles Tailwind utilities nor evaluates `color-mix()`, so `getComputedStyle` cannot observe the regression. The only mechanically possible assertion is that the element carries the literal string `hover:bg-background-hover`, which is exactly the component-only, implementation-detail test that `docs/testing/app-testing.md` prohibits and that `docs/testing.md` classifies as over-testing. The repository has no visual-regression harness for hover states. Flagging explicitly so a reviewer can overrule if they want the class-string pin anyway.
- **Verification performed instead**: the token was compiled through Tailwind v4 in isolation. `hover:bg-background-hover` emits `background-color: var(--color-background-hover)` with no collision against the existing `--color-background-50`; the resulting colour is fully opaque in both themes (values above); and Tailwind additionally emits an opaque `@supports` fallback so engines without `color-mix` degrade to "no hover effect" rather than back to a transparent button.
- **Conventions**: Compliant — no test files changed, so no mock-boundary, MSW, real-database, or fake-timer rules are engaged.

PR #26131, opened 2026-08-10
test(api): pin usage pack preview proration clock

e7h4n:
## Summary

- pin the repository clock for the unconfirmed package-change preview test
- keep the exact purchased, bonus, total-credit, and expiry assertions intact
- restore the clock through the test lifecycle

## Root cause

Turbo run https://github.com/vm0-ai/vm0/actions/runs/31372179891 failed only in test-api (4): usage pack allocation management > replaces an unconfirmed package change preview.

The fixture captured floor(now / 1000) to build a 30-day billing period. The preview route later captured floor(nowDate / 1000) as the proration timestamp. When those reads crossed one second, the remaining period became 1,295,999 of 2,592,000 seconds. Flooring each independent credit delta produced 14,999 purchased credits and 1,099 bonus credits instead of the exact half-period values 15,000 and 1,100. The unchanged expiry and received total of 16,098 match that boundary exactly.

The other API shards were fail-fast cancellations, and ci-gate-turbo only aggregated the substantive test failure. The triggering PR #26117 changed Runner Python files and is unrelated.

## Validation

- pnpm format
- pnpm turbo run lint --concurrency=1 --log-order grouped
- pnpm knip
- pnpm turbo run check-types --concurrency=1 --filter=!api --filter=!@vm0/app
- pnpm --filter @vm0/app run check-types
- pnpm --filter api run check-types
- git diff --check

The targeted API integration test was not run locally because it requires PostgreSQL and the repair contract forbids starting a local service. PR CI provides the real test database environment. Preview/browser validation is not applicable because this changes only a test clock and no deployable behavior.

PR #26104, opened 2026-08-10
feat(billing): manage member usage packs and credits

hulh122:
## Summary

- expose feature-gated personal usage pack balances and credit additions in the account menu and Credit balance settings, while hiding usage surfaces for members without an active pack
- let organization admins review every member balance in a dense expandable table with purchased, bonus, expiry, and credit addition details
- support adding active organization members to an existing usage pack subscription with Stripe proration, invoice-paid fulfillment, period-aligned purchased and bonus credits, and cross-deployment schema capability checks
- extend usage pack change persistence with an `addition` lifecycle and nullable source fields through migration `0877_tiny_omega_sentinel`
- refine plan and package management summaries, comparisons, confirmation states, downgrade restoration, and concurrent slot counts for Pro and Team
- preserve the existing UsagePackPlans feature switch and tolerate older API deployments through optional contract fields

## Tests

- `pnpm format`
- `pnpm -F @vm0/api-contracts lint`
- `pnpm -F @vm0/db lint`
- `pnpm -F api lint`
- `pnpm -F @vm0/app lint`
- `pnpm -F @vm0/api-contracts check-types`
- `pnpm -F @vm0/db check-types`
- `pnpm -F api check-types`
- `pnpm -F @vm0/app check-types`
- pre-commit full Turbo type check: 14 workspaces passed
- API billing integration tests: 83 passed
- Platform billing and account UI tests: 55 passed
- API contracts tests: 543 passed
- DB tests: 29 passed
- `pnpm knip`

## Local migration note

- migration snapshot count, chain, tag references, and timestamp ordering pass validation
- the complete local migration replay is blocked by the dev container PostgreSQL 16 server because an older repository migration uses the PostgreSQL 17 `transaction_timeout` setting; CI remains responsible for the canonical migration replay

PR #26247, opened 2026-08-11
test(api): pin member package proration clock

e7h4n:
## Summary

- pin the repository clock for the active-member package addition preview test
- restore the clock through the Vitest lifecycle
- preserve the exact purchased, bonus, total-credit, amount, and expiry assertions

This is a focused stacked repair for #26104 because the affected test and member-addition path are not present on main yet.

## Root cause

Turbo run https://github.com/vm0-ai/vm0/actions/runs/31453260499 failed only in test-api (4): usage pack allocation management > adds an active member package to the existing subscription.

The fixture captured floor(now / 1000) to create a 30-day billing period centered on the current second. The preview route later captured the repository clock again for the proration timestamp. When those reads crossed one second, the remaining period became 1,295,999 of 2,592,000 seconds. Flooring the independently prorated 50,000 purchased credits and 2,600 bonus credits produced 24,999 and 1,299 instead of the exact half-period values 25,000 and 1,300. The received total of 26,298 matches that one-second boundary exactly.

This is the same clock-ownership pattern fixed previously in #26131. Browser Use and Axiom warnings in the failed shard came from intentional error-path tests; test-api (5) and test-api (6) were fail-fast cancellations, and ci-gate-turbo only aggregated the substantive assertion failure.

## Validation

- `pnpm format`
- `NODE_OPTIONS=--max-old-space-size=3072 pnpm turbo run lint --concurrency=1 --log-order grouped`
- `pnpm knip`
- direct no-service clock/proration assertion through `tsx`, 5 consecutive passes
- `git diff --check`

The service-backed API integration test was intentionally not run locally because it requires PostgreSQL and the repair contract forbids local services. PR CI provides the real test database environment.

## Preview validation

Pending PR CI and preview applicability evaluation. This changes only test clock ownership and no deployable behavior.

### DIFF
--- a/turbo/apps/api/src/signals/route.ts
+++ b/turbo/apps/api/src/signals/route.ts
@@ -74,6 +74,7 @@ import { zeroBillingRedeemCodeRoutes } from "./routes/zero-billing-redeem-code";
 import { zeroBillingRedeemRoutes } from "./routes/zero-billing-redeem";
 import { zeroBillingRestoreRoutes } from "./routes/zero-billing-restore";
 import { zeroBillingStatusRoutes } from "./routes/zero-billing-status";
+import { zeroBillingUsagePackCreditsRoutes } from "./routes/zero-billing-usage-pack-credits";
 import { zeroBankingRoutes } from "./routes/zero-banking";
 import { zeroChatThreadRoutes } from "./routes/zero-chat-threads";
 import { zeroChatEventsRoutes } from "./routes/zero-chat-events";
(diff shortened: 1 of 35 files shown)


e7h4n:
Closing as superseded. While this draft was being validated, PR #26104 was force-pushed to 4d22eb5424 with the identical four-line repository-clock pin. Its updated test-api (4) passed at https://github.com/vm0-ai/vm0/actions/runs/31454083803/job/93664343675. This branch contains no unique repair.

Issue #26229, opened 2026-08-11
bug(platform): PWA chat composer takes excessive bottom space on mobile


bingjiez666:
## Description

On the mobile PWA (iOS standalone mode, app.vm0.ai), the bottom chat input area occupies roughly 20% of the screen height, squeezing the conversation content into the upper half of the viewport.

From the screenshot (iPhone, iOS PWA):

- The empty composer (placeholder "Type your next message...") is about 3 text lines tall even with no content.
- The toolbar row (attach / template / workflow / browser toggles on the left, spark / mic / send on the right) adds its own vertical padding.
- There is a large blank strip below the input container — noticeably taller than the expected `safe-area-inset-bottom`, which suggests the safe-area inset may be applied twice (container bottom padding + inset).

## Areas to check

- `min-height` of the composer textarea in the empty state (should collapse to a single line)
- Vertical padding of the composer toolbar row
- Whether `env(safe-area-inset-bottom)` stacks on top of existing container bottom padding in standalone/PWA display mode

## Environment

- iOS PWA (standalone, added to home screen), app.vm0.ai
- Reported by Ming in Slack

## Screenshot

![iOS PWA chat view with oversized bottom composer area](https://cdn.vm0.io/artifacts/yj3unji3bk.png)

bingjiez666:
## Root cause analysis (confirmed against code + pixel measurements)

Measured Ming's screenshot (iPhone, 393×852pt, standalone PWA, keyboard closed) and matched every pixel to code constants. The bottom dock totals **172px = 20.2% of the viewport**. No safe-area double-application bug — it is three stacked design constants:

| Segment | px | Source |
|---|---|---|
| Editor reserved height (`min-h-[68px]`, includes `pt-4`) | 68 | `tiptap-workflow-composer.ts` `editorContentClass` |
| Toolbar row (`pt-1` 4 + `h-8` buttons 32 + `pb-4` 16) | 52 | `zero-chat-composer.tsx` `ComposerCard` |
| Card borders | 2 | |
| Card→edge padding (inner `pb-2` + footer `pb-2`) | 16 | `zero-chat-thread-page.tsx` `ChatThreadComposer` |
| `#root` safe-area padding (`--sab-raw` = 34 on iPhone) | 34 | `views/css/index.css` |

Screenshot cross-check: card = 121.3px measured vs 122 computed; gap below card = 50px measured vs 8+8+34 = 50 computed. ✅

### Why the empty box is ~3 lines tall

[...]

PR #20976, opened 2026-07-10
fix: centralize mobile safe areas at the root

bingjiez666:
## Summary

- apply the four-sided safe-area boundary once on the existing CSS `#root`, so every normal route inherits it without adding a React layout wrapper
- make regular `zero-viewport-shell` pages fill the root content box rather than repeating viewport and safe-area geometry
- use `zero-fixed-viewport-shell` only for the affected image and HTML lightbox that escapes the root through a fixed overlay
- keep the fixed mobile sidebar inside the root safe-area content boundary while its background continues to cover the viewport
- remove duplicate component-level safe-area padding from the mobile top bar, chat composer, and sidebar header/footer
- let auth, 404, and error layouts inherit the root height instead of bypassing it with viewport units or fixed positioning
- preserve the existing standalone PWA `100lvh`/`100dvh` behavior, keyboard-aware bottom inset suppression, and already-correct fullscreen artifact surfaces

## Verification

- `pnpm exec prettier --check` for all changed platform files
- `pnpm -F @vm0/app exec vitest run src/views/css/__tests__/entrypoint-safe-area.test.ts src/views/auth/__tests__/auth-page.test.tsx src/views/zero-page/__tests__/zero-telegram-connect-page.test.tsx src/views/zero-page/__tests__/zero-github-connect-page.test.tsx src/views/zero-page/__tests__/zero-teams-connect-page.test.tsx src/views/zero-page/__tests__/zero-directed-connect-page.test.tsx src/views/zero-page/__tests__/zero-slack-connect-page.test.tsx --maxWorkers=1` (27 passed)
- `pnpm -F @vm0/app exec vitest run src/views/zero-page/__tests__/zero-attachment-chips.test.tsx --maxWorkers=1 -t "keeps the modal fullscreen state while navigating images|opens presentation artifact controls from chat message links"` (2 passed)
- `pnpm -F @vm0/app lint`
- `pnpm -F @vm0/app check-types`

Local dev server was not started per the vm0 PR verification preference. Exact iOS standalone safe-area rendering still requires the PR preview on an iOS device.

PR #26248, opened 2026-08-11
fix(platform): overlap chat composer bottom gap with pwa safe-area inset

bingjiez666:
## Summary

Fix lever 2 from #26229: in standalone PWA mode, the gap below the chat composer card is 50px because the thread footer's bottom paddings stack on top of the root-owned `safe-area-inset-bottom` (8px inner + 8px footer + 34px inset). #20976 replaced the footer's original `max(0.5rem, var(--sab))` overlap semantics with a plain `pb-2`, which is where the stacking began.

This PR restores the overlap at the footer layer:

- `[data-chat-composer]` now gets `padding-bottom: max(0.5rem - var(--sab), 0px)` in the shared safe-area CSS block, replacing the footer's `pb-2` utility.
- Gap below the card becomes `8px + max(8px, inset)`: **42px in iPhone standalone PWA (was 50px)**, unchanged 16px in browser/desktop where the inset is 0.
- Keyboard behavior is unchanged: `--sab` is already zeroed while the keyboard is open, so the full 8px returns above the keyboard and the keyboard-reserve measurements see the same footer height as before.
- The inner scroll container's `pb-2` is deliberately kept: it is the clip headroom for the card's shadow/focus veil inside the `overflow-y-auto` wrapper.

Added a regex assertion for the new rule in `entrypoint-safe-area.test.ts`, next to the existing safe-area contract assertions.

Part of #26229 (lever 1, the 68px editor min-height, is a separate change).

## Verification

- Formatting via Prettier on the three changed files.
- Relying on the PR pipeline for lint, type checks, and the `entrypoint-safe-area` vitest suite.

### DIFF
--- a/turbo/apps/platform/src/views/css/__tests__/entrypoint-safe-area.test.ts
+++ b/turbo/apps/platform/src/views/css/__tests__/entrypoint-safe-area.test.ts
@@ -92,6 +92,9 @@ describe("platform entrypoint safe area behavior", () => {
       /@media\s*\(display-mode:\s*standalone\)\s*{[\s\S]*\[data-chat-composer\]\s+\.zero-composer\s*{\s*scroll-margin-block-end:\s*16px;\s*}/,
     );
     expect(globalCss).toMatch(
+      /\[data-chat-composer\]\s*{\s*padding-bottom:\s*max\(0\.5rem\s*-\s*var\(--sab\),\s*0px\);\s*}/,
+    );
+    expect(globalCss).toMatch(
       /#root::after\s*{[\s\S]*height:\s*var\(--zero-keyboard-scroll-reserve,\s*0px\);[\s\S]*pointer-events:\s*none;/,
     );
     expect(globalCss).not.toContain("data-keyboard-inset-page");
--- a/turbo/apps/platform/src/views/css/index.css
+++ b/turbo/apps/platform/src/views/css/index.css
@@ -150,6 +150,15 @@ body {
   }
 }
 
+/* The chat composer footer keeps 8px of breathing room below the card but
+   overlaps it with the root-owned bottom inset instead of stacking on top of
+   it, so the gap to the screen edge is max(0.5rem, inset) rather than
+   0.5rem + inset. While the keyboard is open --sab is 0px and the full 8px
+   returns above the keyboard. */
+[data-chat-composer] {
+  padding-bottom: max(0.5rem - var(--sab), 0px);
+}
+
 .zero-sidebar-header {
   padding-top: 0.375rem;
 }
--- a/turbo/apps/platform/src/views/zero-page/zero-chat-thread-page.tsx
+++ b/turbo/apps/platform/src/views/zero-page/zero-chat-thread-page.tsx
@@ -4557,7 +4557,7 @@ function ChatThreadComposer({ thread }: { thread: ChatPanelSignals }) {
   return (
     <footer
       data-chat-composer
-      className="relative shrink-0 bg-[hsl(var(--background))] pb-2"
+      className="relative shrink-0 bg-[hsl(var(--background))]"
     >
       <div className="pointer-events-none absolute inset-x-0 -top-5 h-[21px] bg-gradient-to-t from-[hsl(var(--background))] to-transparent" />
       <div


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 2f04e627e9dab5114074ca946cf6db44b0037b21
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Removes the `pb-2` utility from the chat thread composer footer and replaces it with a shared CSS rule `[data-chat-composer] { padding-bottom: max(0.5rem - var(--sab), 0px) }`, so the footer's 8px breathing room overlaps the root-owned `safe-area-inset-bottom` instead of stacking on it. In standalone PWA the gap below the composer card drops 50px → 42px; browser, desktop, and keyboard-open geometry are unchanged because `--sab` is 0 in those states. Adds a regex assertion pinning the new rule in the existing `entrypoint-safe-area` contract test.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic (edge cases, conditionals) | Pass | `index.css`: `max(0.5rem - var(--sab), 0px)` clamps at 0 for inset ≥ 8px; keyboard-open path keeps 8px because `:root[data-keyboard-open="true"]` zeroes `--sab`; `[data-chat-composer]` matches only the thread footer (verified via repo grep) |
| `REVIEW.md` | Security (injection, secrets, auth) | N/A | CSS spacing + test regex only; no executable logic, no endpoints |
| `REVIEW.md` | Type safety (no `any`, proper types) | N/A | No TypeScript logic changed; one JSX className string literal edited |
| `REVIEW.md` | Testing coverage: `fix:` commit needs a regression test | Pass | `entrypoint-safe-area.test.ts` gains a regex assertion pinning the new rule, same contract-test pattern that already pins `#root` safe-area padding and the keyboard `--sab: 0` rule |
| `REVIEW.md` | Fallback discipline (`docs/fallback.md` routing) | N/A | Diff introduces no compatibility branch, default chain, or feature-switch path; `max()` is layout arithmetic, not a fallback |
| `REVIEW.md` | Feature regression guard | Pass | Non-PWA and keyboard-open spacing are provably unchanged (`--sab` = 0 in both states); those states are already pinned by existing assertions in the same test file |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, contract, or persisted-state change |
| `docs/bad-smell.md` | No `any`, no suppressions, no dynamic imports, no hardcoded config, fail-fast | Pass | Diff adds a CSS rule with a why-comment, removes one utility class, adds one test assertion; none of the prohibited patterns appear |
| `docs/testing.md` | Integration-first, mock at boundary only, no fake timers, assert behavior | Pass | Test change adds an assertion to an existing suite; no mocks, no timers, no `vi.mock` |
| `docs/effect.md` | Keep React render pure; correct execution boundary | N/A | No render logic, command, effect, or signal changed; className literal only |
| `docs/cache.md` | RC-1..RC-9 lifecycle/cache anti-patterns | N/A | No state, cache, ref, or resource lifecycle touched |
| `.claude/skills/ccstate/SKILL.md` | ccstate command/detach/pageSignal patterns | N/A | No ccstate code in the diff |
| `docs/testing/app-testing.md` | Platform tests are page-level or bootstrap-level | Pass | Change extends the pre-existing `views/css/__tests__` entrypoint contract suite rather than adding a new unit-test surface |
| `docs/testing/testing-external-behavior.md` | Test via external surface, not implementation internals | Pass | jsdom page tests cannot observe `display-mode: standalone` or `env(safe-area-inset-*)`; the repo's established boundary for this surface is the entrypoint CSS contract test, and the new assertion follows the file's existing convention |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

Notes (non-blocking): the new rule is intentionally global for `[data-chat-composer]`; today that attribute is only on the thread footer, and the co-located comment documents the max-overlap semantics, so a future second consumer inherits documented behavior.

### Testing

- Coverage: Adequate — regression pinned via the entrypoint safe-area contract assertion (`padding-bottom: max(0.5rem - var(--sab), 0px)`), alongside the existing assertions that pin the keyboard `--sab: 0` behavior this rule relies on.
- Conventions: Compliant — no mocks, no fake timers, no internal `vi.mock`, no negative tests against removed behavior.

PR #26273, opened 2026-08-11
refactor(connectors): unify connector connection writes

### LINKED ISSUE #26086: refactor(connectors): unify connector connection writes
Parent: #25312

## Problem and evidence

Current `main` has one transactional `replaceConnectorConnection` boundary for Builtin token/OAuth, Custom OAuth, and Custom manual connection establishment. The boundary preserves the logical connection ID, resets replacement-owned metadata, deletes the exact connection's stale credentials, runs a caller-supplied credential write, and returns the connection row in one transaction.

Builtin manual and no-auth establishment still bypass that boundary. They call `cleanupExistingStoredConnectorForLocalConnect`, which captures previous revoke state and deletes connector-owned credentials, then call a separate `upsertLocalConnectorRow` that duplicates the Builtin insert, partial-index conflict update, reconnect reset, timestamp update, and returned-row projection. This leaves two owners for the same replacement invariants and can drift from both Builtin OAuth and Custom behavior.

A production-source audit on the planned baseline finds no other connection-establishment `insert(connectors)` outside the shared service and this local duplicate. Refresh is a different operation: it rotates selected credentials and expiry on a validated connection while preserving unrelated rows, so it must remain an explicit in-place mutation.

The public contract retains a no-auth route, but current accepted connector catalog source and artifact schemas cannot produce a Builtin `grant.kind = "none"` runtime method. The retained service branch must converge with replacement semantics, but adding a successful no-auth catalog fixture would expand product behavior outside this refactor.

## Decision

Proceed with one PR that finishes establishment/replacement convergence without expanding catalog behavior.

Reuse `replaceConnectorConnection` for Builtin manual and no-auth, remove the duplicate local writer and destructive cleanup owner, and consolidate the remaining common Builtin/Custom upsert fields inside the shared service. Preserve source-specific lifecycle orchestration and keep refresh, reconnect marking, disconnect, and provider/account cleanup outside the replacement primitive.

This issue remains a direct child of #25312 and is independently deployable. It has no block relationship to #26080, #26085, #26087, or #26240.

- Keep the replacement target discriminated by persisted owner: Builtin or Custom.
- Give Builtin an explicit nested identity mode:
  - `local` for manual and no-auth rows with null external identity and scopes;
  - `external` requiring the complete external principal and scopes for token/OAuth rows.
- Normalize target values into one shared insert/update/return implementation. Only the Builtin versus Custom partial unique conflict target differs.
- Keep the credential callback required so all callers use the same replace-and-write transaction. Manual writes already-prepared credentials through it; no-auth supplies an explicit empty write so stale credentials are still removed.
- Change local preparation to retain its exact row lock, previous-method resolution, and pending revoke capture while removing its credential deletion.
- Preserve manual validation and encryption before the transaction; legacy user secret/variable cleanup inside the transaction after credential replacement; and revoke, invalidation, wakeup, and captured-abort handling after commit.
- Keep the Custom callback contract unchanged so #26240 can extend its physical storage behavior after a normal rebase.

This shape keeps an invalid external Builtin connection unrepresentable, matches the database's Builtin/Custom ownership distinction, and shares the complete behavioral boundary rather than only extracting SQL.

### Make Builtin external identity nullable

This is a smaller type change, but it permits partially specified OAuth state and cannot distinguish deliberate local identity from a missing external principal. Integration tests cannot restore that compile-time invariant.

### Use `builtin-local`, `builtin-external`, and `custom` as top-level target kinds

This is also type-safe, but it conflates persisted ownership and identity mode in one discriminant. A nested Builtin identity keeps those independent concepts explicit.

### Extract only a lower-level row-upsert helper

This would share SQL while leaving two owners for credential deletion, abort checks, callback order, and future replacement-owned metadata. It does not solve the behavioral drift.

### Add no-auth to the accepted catalog for test coverage

This would make the route testable but would introduce new production catalog and executable-capability semantics solely for a writer refactor. That decision belongs in a separate feature issue if the product needs a no-auth Builtin method.

### Route refresh through replacement

Replacement deletes the complete connector-owned credential set and resets broader metadata. Refresh must preserve unrelated values and enforce its existing revision checks, so sharing this boundary would change correctness and failure semantics.

## Risks and mitigations

- **Atomicity:** Keep the row update, exact credential deletion, and credential callback in the caller's database transaction. Test a real credential-write failure after replacement begins and prove the previous OAuth row and storage survive.
- **Identity correctness:** Represent local and external Builtin identity with a discriminated union; do not weaken OAuth fields to nullable inputs.
- **Conflict-target correctness:** Share common values and updates while keeping the two partial unique conflict targets explicit. Existing Builtin and Custom integration journeys exercise both.
- **Cleanup ordering:** Remove local pre-deletion so shared replacement is the sole destructive owner. Preserve pending revoke capture before replacement, legacy user-row cleanup after the callback, and post-commit effects afterward.
- **Concurrency:** Retain the exact connector-state advisory lock, local row lock, and transaction boundary.
- **No-auth transitions:** Keep the retained branch on the Builtin/local target with an explicit empty callback, without changing catalog reachability.
- **Refresh isolation:** Leave token refresh and reconnect marking as targeted in-place updates with their existing revision checks.
- **Compatibility:** Do not change schema, persisted formats, public APIs, catalog artifacts, feature switches, runner protocols, or firewall behavior.
- **Performance:** The local path retains one connection upsert and one exact credential deletion; consolidation adds no extra database round trip.

## Acceptance criteria

- [ ] Builtin manual, retained Builtin no-auth, Builtin token/OAuth, Custom manual, and Custom OAuth establishment or replacement call one target-discriminated `replaceConnectorConnection` boundary.
- [ ] Builtin local identity is explicit, Builtin external identity remains type-required, and Custom targeting remains explicit.
- [ ] The shared service owns one common connection-row insert/update/return implementation; only target values and the Builtin/Custom partial unique conflict target vary.
- [ ] `upsertLocalConnectorRow`, its duplicate conflict/update projection, and local pre-replacement credential deletion are removed.
- [ ] A production-source audit finds no connection-establishment `insert(connectors)` outside the shared boundary; test and migration fixtures remain explicit exceptions.
- [ ] Manual input validation and secret encryption finish before replacement, and a real transactional credential-write failure leaves the previous working connection and credentials intact.
- [ ] A successful Builtin OAuth-to-manual replacement preserves the logical connection ID, clears stale OAuth credentials and external metadata, and writes the current manual credentials.
- [ ] The retained no-auth branch uses the Builtin/local target and explicit empty credential callback, so shared replacement owns stale credential cleanup; no catalog grant kind is added by this PR.
- [ ] Organization/member ownership, storage version, reconnect reset, returned connection state, and logical connection ID remain consistent across Builtin and Custom replacement.
- [ ] Existing connector-state locks, pending token revocation, account cleanup, user secret/variable cleanup, post-commit invalidation, runtime wakeups, and abort behavior retain their ordering and failure semantics.
- [ ] Token refresh and reconnect marking remain explicit in-place mutations and do not delete unrelated credential rows or invoke replacement-only cleanup.
- [ ] API entry-point integration tests cover Builtin manual establishment, successful auth-method replacement, rollback on credential-write failure, and representative existing Builtin/Custom replacement behavior.
- [ ] No database migration, persisted compatibility state, public API contract, catalog artifact, runner protocol, or firewall behavior change is introduced.

## Delivery notes

This is one PR-sized implementation issue, not a delivery parent. The successful no-auth API-test requirement was removed because the accepted catalog cannot construct that method without a separate product change; the retained branch still converges and remains type-checked. If #26240 edits the Custom credential callback concurrently, coordinate by rebasing normally because there is code overlap but no semantic dependency.

seven332:
## Summary

- route Builtin manual and retained no-auth connection replacement through the same transactional boundary used by Builtin OAuth and Custom connectors
- consolidate the shared Builtin/Custom connection upsert while keeping their partial unique conflict targets explicit
- add an API integration regression test proving a failed OAuth-to-manual credential write rolls back the connection and storage before a successful retry

## Scope

- preserves source-specific locking, token revocation, legacy credential cleanup, post-commit invalidation, wakeups, and abort handling
- keeps refresh and reconnect mutations outside the replacement primitive
- does not change schemas, persisted formats, public APIs, catalog artifacts, feature switches, or runner protocols

## Validation

- `pnpm exec prettier --check apps/api/src/signals/services/connector-connection-write.service.ts apps/api/src/signals/services/zero-connector-data.service.ts apps/api/src/signals/routes/__tests__/connectors.bdd.test.ts`
- `pnpm -F api lint`
- `pnpm -F api check-types`
- `pnpm -F api check-types:tests`
- `pnpm knip --workspace api`
- `pnpm -F api exec vitest run src/signals/routes/__tests__/connectors.bdd.test.ts --maxWorkers=1 --silent=passed-only` (59 tests)
- `pnpm -F api exec vitest run --maxWorkers=1 --silent=passed-only` (187 files, 2,842 tests)
- pre-commit `pnpm check-types` (14 Turbo tasks)

Closes #26086
Relates to #25312

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/connectors.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/connectors.bdd.test.ts
@@ -54,6 +54,8 @@ import {
   readConnectorCredentialStorageState,
   readCustomConnectorCredentialStorageParent,
   readCustomConnectorOAuthStorageState,
+  seedConnectorStorageRow,
+  setConnectorVariableOwner,
   setCustomConnectorCredentialStorageState,
 } from "./helpers/connector-credential-storage-state";
 import { zeroCustomConnectorsRoutes } from "../zero-custom-connectors";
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #26273

**Reviewed head:** `96035cfcf8ad429d78a48b8ac3d1fbd2f5d37749`

### Summary

Builtin local, Builtin external, and Custom connection establishment now share one transactional replacement boundary while retaining their explicit ownership and identity invariants. Source-specific revoke, account, legacy cleanup, refresh, and post-commit behavior remain outside that boundary.

### Findings

No findings.

### Testing

The regression test exercises a real PostgreSQL credential-write conflict through the public connector routes, proves the prior OAuth connection and credentials roll back unchanged, and verifies a successful OAuth-to-manual retry. Local validation passed formatting, lint, API and full-repository type checks, Knip, 59 focused connector tests, and the full serial API suite (187 files, 2,842 tests).

A successful Builtin no-auth route remains unverified because the accepted catalog cannot construct that runtime method without an out-of-scope product change; the retained branch is type-checked and uses the same replacement boundary.

### Verdict

LGTM

PR #26299, opened 2026-08-11
fix: remove nested hover surface on the composer goal row

tongx-blip:
## Problem

The active-goal row in the composer's pending-items strip painted two state layers at the same time:

- the row (`ComposerStripRow`) with `hover:bg-state-hover` — state layer at 5%
- the open-details button inside it with `hover:bg-state-selected-hover` — state layer at 11.5%

That button is `flex-1`, so it covers almost the entire row. Hovering the goal text hit both rules at once and composited 5% + 11.5%, drawing a second rounded rectangle inside the row highlight — a visible box-inside-a-box.

Queued and automation rows never showed this, because their hover surface lives only on the 16px leading icon button, where a small chip nested in a row highlight is the expected icon-button behaviour. Only the goal row also made a full-width target its own surface.

## Change

The open-details target now paints no background of its own; the row hover carries the highlight, and the target signals itself with the existing `hover:text-sidebar-foreground` shift plus the focus ring for keyboard users. One surface per row.

Removing the inner box exposed a second-order issue: the button's `-ml-1 p-1` placed the goal's target glyph 4px left of the queued rows' glyph, which the box had been hiding. The icon now sits in the same `p-1` slot the other rows' leading button uses, so every row's glyph and text line up on one column.

## Verification

Rendered against the repo's real design tokens (`packages/ui/src/styles/globals.css` compiled with Tailwind 4.3.3) and checked with a real pointer hover in light and dark themes: hovering the goal text, the row, or the close button now produces exactly one highlight surface, and the close button keeps its own small chip consistent with the other rows.

No behaviour change — click handlers, `aria-label`s, and roles are untouched, so the existing goal-row tests continue to assert the same contract.

### DIFF
--- a/turbo/apps/platform/src/views/zero-page/zero-chat-composer.tsx
+++ b/turbo/apps/platform/src/views/zero-page/zero-chat-composer.tsx
@@ -521,20 +521,23 @@ function ComposerStripRow({
       className="group flex items-center gap-2 rounded-md pl-2 pr-1 py-1.5 text-sm text-muted-foreground transition-colors hover:bg-state-hover"
     >
       {isGoal && onOpenDetail ? (
+        // This target spans almost the whole row, so it deliberately paints no
+        // background of its own — the row's hover carries the highlight and a
+        // second, near-coextensive surface would read as a box inside a box.
+        // Its icon sits in the same p-1 slot the other rows' leading button
+        // uses, keeping every row's glyph and text on one column.
         <button
           type="button"
-          className="-ml-1 flex min-w-0 flex-1 items-center gap-2 rounded-md p-1 text-left transition-colors hover:bg-state-selected-hover hover:text-sidebar-foreground focus-visible:bg-state-selected-hover focus-visible:text-sidebar-foreground focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring"
+          className="flex min-w-0 flex-1 items-center gap-2 rounded-md text-left transition-colors hover:text-sidebar-foreground focus-visible:text-sidebar-foreground focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring"
           onClick={onOpenDetail}
           aria-label={t(($) => {
             return $.chat.queue.openGoalDetails;
           })}
         >
-          <Target
-            size={16}
-            className="shrink-0 text-emerald-800"
-            aria-hidden="true"
-          />
-          <span className="min-w-0 flex-1 truncate">{text}</span>
+          <span className="flex shrink-0 p-1 text-emerald-800">
+            <Target size={16} aria-hidden="true" />
+          </span>
+          <span className="min-w-0 flex-1 truncate py-1">{text}</span>
         </button>
       ) : (
         <>


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: 718181d94d54a723cf5847f714b137927b1c800c
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Removes the second state layer the composer's active-goal row was painting: the near-full-width open-details button no longer draws its own `state-selected-hover` background, so the row's `state-hover` is the only surface and the nested box disappears. The goal glyph moves into the same `p-1` slot the queued and automation rows use, restoring the 4px column alignment the inner box had been hiding. Presentation-only, one file, 10 additions / 7 deletions.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic — no wrong conditionals, unhandled edge cases, or API misuse | Pass | `zero-chat-composer.tsx:523` keeps the `isGoal && onOpenDetail` branch, `onClick`, and `aria-label` untouched; only `className` values and the icon wrapper change. |
| `REVIEW.md` | Security — no injection, hardcoded secrets, or missing auth on new surfaces | N/A | Diff contains no data handling, network call, or auth path; it is Tailwind class text and one JSX element wrapper. |
| `REVIEW.md` | Type safety — no `any`, no overly broad types | Pass | No type annotations added or changed; `<Target size={16} aria-hidden="true" />` keeps its existing lucide props. |
| `REVIEW.md` | Style & maintainability — no oversized functions, no duplicated logic, comments explain WHY | Pass | `ComposerStripRow` gains no branches; the added comment states why the target intentionally paints no background rather than restating the classes. |
| `REVIEW.md` | Fallback discipline — declare every fallback the diff introduces | N/A | The diff adds no `??`/`||` default, compatibility branch, feature-switch gate, or negative test; no `Fallbacks` section is required. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No `turbo/packages/db/src/schema/**`, contract, or `jsonb(...).$type<...>()` change. |
| `REVIEW.md` | Feature regression guard — changed existing behavior needs scoped coverage | Pass | The only altered behavior is the hover/focus paint of one row. Click target, `role="listitem"`, `aria-label`s, and the close button contract are unchanged, so the existing goal-row page tests still cover the interactive path. |
| `REVIEW.md` | Testing coverage — `fix:` commits need a regression test | Pass | See Testing: the regression is a CSS-class-level presentation detail that `docs/testing/testing-external-behavior.md` explicitly forbids asserting on. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | No `any`, no `eslint-disable`/`oxlint-disable`/`@ts-*`/`prettier-ignore` in the diff. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | No imports added or changed. |
| `docs/bad-smell.md` | No fallback/recovery slop, no hardcoded config or URLs | Pass | No conditional defaults or environment values introduced. |
| `docs/testing.md` | Integration tests only; do not over-test presentation | Pass | Adding a class-level assertion for this diff would be exactly the over-testing this document warns against; no test file was added. |
| `docs/testing/app-testing.md` | Platform tests are page-level via `detachedSetupPage`, no component-only unit tests | Pass | No test added, so no component-only or direct-render test is introduced. |
| `docs/testing/testing-external-behavior.md` | Do not assert on CSS classes, component state, or internal callbacks | Pass | The change is entirely CSS classes; this document names them as outside the user-visible interface, so no assertion is written against them. |
| `docs/effect.md` | Render stays pure — no Store writes, I/O, timers, listeners, or signal creation in render | Pass | `ComposerStripRow` render still only reads props and `useTranslation`, building elements; the diff adds no effect or command. |
| `docs/effect.md` | Execution boundary matches the event (computed / command / setup / `onRef` / effect) | N/A | The diff introduces no new work to place on a boundary; `onOpenDetail` remains the caller-supplied handler bound in `PendingItemsStrip`. |
| `docs/cache.md` | RC-1/RC-2 — no render-phase side effects or render-time signal allocation | Pass | No signal factory, `state`, `computed`, or `command` is created in this component; the diff is JSX and class strings. |
| `docs/cache.md` | RC-7/RC-8 — stable refs, symmetric resource lifecycles | N/A | No ref, listener, timer, observer, or external resource is added; the new `<span>` is a static layout wrapper. |
| `.claude/skills/ccstate/SKILL.md` | DOM callbacks use `detach(..., Reason.DomCallback)`; `detach` stays out of `signals/` | N/A | The diff touches no ccstate call site. The existing `detach(...)` calls for the goal row live in `PendingItemsStrip` and are unchanged. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

Non-blocking observations, no change requested:

- `zero-chat-composer.tsx:521` still carries `group` on the row, but no descendant in `ComposerStripRow` uses a `group-hover:` variant. This predates the PR and is unrelated to the nesting fix, so it does not belong in this diff.
- The goal glyph keeps `text-emerald-800`, which is low contrast on the dark-theme strip (`bg-gray-100` in dark). Also pre-existing and out of this PR's scope.

### Testing

- Coverage: Adequate. This is a `fix:` commit whose entire regression surface is the composited background of one row — a CSS class outcome. `docs/testing/testing-external-behavior.md` states for `turbo/apps/platform`: "Do not assert on query cache, component state, CSS classes, or whether an internal callback was called," and `docs/testing.md` lists presentation details among the things the project deliberately does not test. A guard test here would have to assert on `className`, freezing the implementation the PR just changed. The behavioral contract that a page test *can* observe — the row renders, `Active goal` is labelled, the open-details button carries `openGoalDetails`, and the close button cancels — is unchanged by this diff and remains covered by the existing goal-row tests in `views/zero-page/__tests__/chat-workflows-and-goals.test.tsx`.
- Conventions: Compliant. No test file is added or modified, so no mock-boundary, MSW, fake-timer, or `detachedSetupPage` convention is engaged.
- Verified manually: the row's geometry is unchanged (before and after, the open-details button's content box is 28px tall, so the row stays 40px), and the goal glyph now starts at the same 12px offset as the queued row's leading icon button, with both text runs starting at 40px.

Issue #26137, opened 2026-08-10
chore(connectors): retire skill repair and clean orphan artifacts


seven332:
Parent: #26075

## Problem and evidence

#26135 introduced temporary semantic repair/status surfaces. #26136 made every
supported ordinary Custom and managed Feishu writer persist `skill_markdown`
and its exact `skill_storage_version_id` together, and cut runtime admission over
to that exact version.

Two lifecycle gaps remain:

1. PostgreSQL still permits the two columns to disagree, while the repair cron,
   mutation/status endpoints, repair implementation, and mixed-state fixtures
   remain solely for the completed rollout.
2. Prepare-before-activate publication creates/reads the storage row and uploads
   `${s3_prefix}/${versionId}` before the definition transaction wins. Abort,
   failed activation, or lost CAS can leave unregistered objects. Connector
   deletion is a hard delete that intentionally leaves storage behind.

Age plus a final database query is insufficient for deletion safety. The same
content-addressed key can be published again between an S3 listing and deletion,
then registered after cleanup removes it. `storages.updated_at` records
publication rather than connector deletion, and deleting the storage row loses
the only persisted historical prefix needed to retry a failed S3 cleanup.

## Production gate

- #26136 shipped in API release `0164ab0`; deployment `5847685239` reached
  `api/production` successfully at 2026-08-11 09:18:38 UTC.
- The predecessor deployment is inactive.
- Protected workflow run `31465095411`, attempt 3, queried production after that
  promotion and reported 17 total, 17 verified, zero unresolved, every reason
  count zero, and `complete: true`.

The invariant/repair contraction is therefore unblocked.

## Decision

Use three ordered, independently deployable PRs. This issue is their delivery
parent.

```text
PR 1: pair CHECK + repair retirement
  -> PR 2: publication leases + deleted-storage tombstones, no deletion
       -> production deployment and old-writer drain
            -> PR 3: bounded version/storage cleanup cron
```

Destructive cleanup cannot start in the same release that first teaches writers
the lease protocol. During API promotion, the preceding writer can still overlap
and publish without a lease. The explicit third deployment is the compatibility
gate; do not replace it with a feature switch or an assumed timing window.

### 1. Enforce the pair invariant and retire repair

- Add a generated Drizzle CHECK accepting exactly both-null or both-non-null.
- Perform no migration backfill, coercion, or row deletion.
- Remove the repair schedule, repair/status contracts and routes, repair service,
[...]

Issue #26075, opened 2026-08-10
fix(connectors): pin custom connector skill versions


seven332:
Parent: #25312

## Problem and evidence

Custom connector definitions and generated skill artifacts are activated through
separate mutable operations.

- User-created create/update paths commit `skill_markdown` before publishing the
  skill. Publication failure therefore leaves a definition active without its
  requested artifact.
- Publication returns a content-addressed storage `versionId`, but Custom callers
  discard it and only advance mutable storage HEAD.
- Run creation records connector id/slug and mounts without a version, so it uses
  whichever HEAD is current rather than the artifact admitted with the definition.
- Concurrent updates can commit in one order and complete uploads in another,
  cross-wiring the active definition and HEAD. Request-path whole-storage deletion
  can race a later publication.
- Managed Feishu Custom connectors use the same post-commit publication helper.

Builtin connector skills already use the required invariant: an accepted catalog
definition carries immutable `{ storageName, versionId }` metadata, prepared
storage registration and definition activation commit together, and run creation
mounts that exact version.

## Decision

Align Custom and Builtin connector skill semantics without moving Custom
definitions into the Builtin catalog:

- Generate, upload, and verify the artifact before its definition can reference it.
- Register the immutable storage version, activate the definition, associate the
  exact version, and update temporary compatibility HEAD in one database commit.
- Mount the exact version selected from the admitted definition.
- Reuse ownership-independent prepared-version registration/validation and exact
  connector-volume construction.
- Keep Builtin catalog validation/system ownership and Custom API synthesis/
  organization ownership as separate adapters.

The numeric Custom `storageVersion` remains exclusively the credential storage
contract version. It is not skill artifact identity.

## Chosen direction

Use an additive, staged rollout:

1. Add nullable `skill_storage_version_id -> storage_versions.id` without changing
   readers or writers.
2. Extract reusable immutable storage preparation/registration primitives while
   preserving existing server-side upload and Builtin behavior.
3. Convert all user-created and managed Feishu writers to prepare/verify first and
   atomically bind the winning definition and exact version. Readers continue using
   HEAD until old writers drain.
4. Re-synthesize and verify **every** skill-bearing legacy definition, not only
   rows with a null pointer. An overlapping old writer can leave a non-null but
   stale pointer. Compute the expected version from authoritative synthesized
   content and the canonical storage id, repair mismatches with CAS, and expose
   completion status based on semantic equality.
5. After writer deployment and repair completion gates pass, switch Custom runtime
[...]

PR #26313, opened 2026-08-11
fix(connectors): pin custom skills to admitted versions

### LINKED ISSUE #26136: fix(connectors): mount custom skills at exact admitted versions
Parent: #26075

## Problem and evidence

The immutable Custom connector skill association is populated and repaired, but
runtime admission still ignores it.

- `customConnectorDefinitionSelection()` does not select
  `skillStorageVersionId`.
- Custom runtime skill descriptors carry only connector id and slug.
- Custom skill volumes omit `version`, so manifest preparation resolves mutable
  organization storage HEAD.
- `custom_connector_skill` uses the generic additional-volume resolver, which can
  turn missing storage or HEAD into an omitted mount. Builtin connector skills
  already require an admitted exact system version and fail closed.

The protected #26135 production sweep completed with `total: 16`, `verified: 16`,
`unresolved: 0`, every reason count at zero, and `complete: true`:
https://github.com/vm0-ai/vm0/actions/runs/31465095411/job/93697944137.
Every visible association therefore passed authoritative content identity,
storage ownership, exact registration, and compatibility HEAD checks before
reader cutover.

Research found an additional deployment requirement. #26134 writers expose a
valid nullable pair after transaction commit, but ordinary and Feishu create
write non-null markdown before registering and binding the exact pointer later in
the same transaction. PostgreSQL CHECK constraints are immediate per statement,
and production migrations run before the new API is promoted. Adding the final
pair CHECK in this release would break those paths on still-serving #26134 API
instances.

## Decision

Cut Custom runtime admission over to the persisted exact version and make all
current writers statement-valid for the eventual pair CHECK. Do not add the CHECK
in this release. After this API is deployed and the previous API drains, #26137
will install the normal CHECK as its first contraction step, then retire repair
surfaces and clean bounded artifacts.

This is a deliberate two-release compatibility boundary, not a runtime fallback.
The reader rejects mixed or unavailable state immediately and never follows HEAD.

## Chosen direction

- Carry `skillStorageVersionId` through internal Custom definition rows and
  runtime skill descriptors without exposing it in public Custom responses.
- Treat markdown and exact version as one admitted value: both absent means no
  skill; both present means a versioned skill; either mixed state is an invariant
  failure.
- Build Builtin and Custom connector skill volumes through one focused exact
  constructor while preserving `connector_skill` versus
  `custom_connector_skill` sources.
- Resolve both sources with shared exact-version mechanics and fixed ownership
  branches. Builtin remains system-owned with its catalog prefix; Custom remains
  runtime-organization-owned with the database-persisted prefix. Neither can
  fall back to the other owner or to HEAD.
- Fail closed on malformed ids, missing named storage, unavailable exact version,
  wrong storage ownership/association, or S3 key inconsistent with the persisted
  prefix and exact version. Do not convert these failures to absent mounts.
- Make ordinary create insert markdown and pointer together after transaction-time
  version registration. Make ordinary update win the existing CAS row lock before
  storage database changes, then write the complete pair in one final update.
- Keep Feishu's external preparation, advisory lock, stale-target retry, and fixed
  skill metadata; register the prepared version before inserting/updating the
  complete pair, and include pointer equality in its no-op decision.
- Keep the existing run storage manifest and runner wire format; they already
  persist and carry exact resolved versions.

## Alternatives considered

- **Keep resolving Custom HEAD:** leaves admitted definitions mutable and makes
  the repaired association operationally unused.
- **Only add `version` to the current volume:** pins the happy path but preserves
  generic silent omission and implicit ownership fallback behavior.
- **Duplicate the Builtin resolver:** can be correct but repeats full-version,
  exact lookup, ownership, prefix/key, and failure invariants.
- **Add the CHECK now with a deferred constraint trigger:** can tolerate the old
  writer only through a bespoke function/trigger that re-reads final rows and
  still needs replacement later. It adds migration, testing, and cleanup
  complexity solely to avoid the normal staged deployment boundary.
- **Change only the reader now:** requires another writer-preparation release
  before any normal CHECK can be deployed and fragments one known dependency.
- **Never add the CHECK:** leaves invalid persisted states representable and the
  parent invariant incomplete.

## Risks and mitigations

- **Rolling API versions:** no schema changes occur here. Old readers use
  compatibility HEAD while new readers use the exact pointer; committed #26134
  writers keep them aligned. New writers become safe for the next release's
  immediate CHECK.
- **Optimistic update regression:** retain the exact storageVersion/updatedAt CAS
  predicates and not-found-versus-retry result. Acquire that row lock before
  registering storage so a loser cannot move HEAD or pointer.
- **Lock ordering:** retain ordinary prefix advisory locking and Feishu
  installation advisory locking; keep all S3 preparation outside transactions.
- **Ownership confusion:** derive owner from the internal manifest source in fixed
  branches. Custom cannot be system-owned and Builtin cannot resolve from a
  tenant.
- **Silent mount loss:** connector-skill failures bypass generic optional handling
  and fail the run visibly.
- **Temporary database permissiveness:** the interval starts from a completed
  semantic sweep; committed writers remain paired and the exact reader rejects
  mixed states. #26137 owns the required CHECK and cannot complete without it.

## Acceptance criteria

- Ordinary Custom and managed Feishu runtime descriptors carry the persisted
  exact skill version.
- Builtin and Custom share exact-version construction and resolution mechanics
  while preserving system versus organization ownership, catalog versus persisted
  prefix rules, and existing manifest diagnostic sources.
- New Custom run manifests contain the exact version admitted with the definition
  and never resolve implicit HEAD.
- A run admitted before a connector update keeps its earlier persisted version;
  a later run receives the later admitted version.
- Mixed pairs, malformed versions, wrong-storage pointers, or unavailable exact
  registrations fail closed rather than omitting the volume or falling back.
- Ordinary and Feishu create/update statements always write markdown and pointer
  in a form accepted by the future immediate nullable-pair CHECK.
- Existing ordinary/Feishu failure atomicity, losing-writer, and Feishu stale
  target behavior remain unchanged.
- Builtin exact-version mounting and its failure matrix remain behaviorally
  unchanged.
- Public Custom contracts, credential `storageVersion`, run manifest schema,
  queue payloads, and runner contracts remain unchanged.
- Route-level integration tests cover ordinary exact mounts, update-after-admit
  pinning, fail-closed Custom invalid state, managed Feishu exact mounts, current
  writer concurrency/failure cases, and unchanged Builtin exact mounts.
- Applicable formatting, lint, type, Vitest, and Knip checks pass.

## Delivery and rollout

This PR is unblocked by the completed #26135 gate and contains no migration. If a
writer incident or suspected drift occurs before deployment, rerun the protected
semantic status check.

After this API release is in production and the previous API has drained, #26137
must:

1. rerun semantic status and require zero unresolved rows;
2. add the normal database CHECK requiring both skill fields to be simultaneously
   null or non-null;
3. remove repair/status schedules, routes, services, and now-impossible test
   mutations;
4. perform bounded orphan and deleted-connector storage cleanup.

Keep #26075 open until that contraction and cleanup complete.

## Intentionally deferred

- The final database nullable-pair CHECK and its generated migration.
- Removing temporary repair/status surfaces and impossible-state fixtures.
- Bounded unregistered-artifact and deleted-connector storage cleanup.
- Deleting registered historical versions for active connectors.

seven332:
## Summary

- carry each Custom connector's persisted skill storage version into run admission
- resolve Builtin and Custom connector skills through shared fail-closed exact-version mechanics with fixed system and organization ownership
- make ordinary and managed Feishu writers write skill markdown and its exact version together, preparing the next release for the nullable-pair database constraint

## Behavior

Runs now mount the exact Custom connector skill version admitted with the connector definition. Updating a connector after run admission no longer moves that run to the new storage HEAD. Missing, mixed, malformed, wrong-storage, or unavailable Custom associations fail closed instead of omitting the mount or following HEAD.

Builtin connector skill behavior remains unchanged. Public Custom connector responses, credential `storageVersion`, run manifests, queue payloads, and runner contracts are unchanged.

## Rollout

This PR intentionally contains no database migration. Existing and new API versions can overlap while the exact reader cuts over. After this API version is deployed and the previous writer drains, #26137 will recheck production state, add the immediate nullable-pair CHECK, and remove the temporary repair/status surfaces and bounded artifacts.

## Validation

- `pnpm -F @vm0/db db:reset`
- `pnpm -F api exec vitest run src/signals/routes/__tests__/run-lifecycle.bdd.test.ts src/signals/routes/__tests__/zero-feishu-integration.test.ts src/signals/routes/__tests__/connectors.bdd.test.ts src/signals/routes/__tests__/cron-custom-connector-skill-repair.test.ts src/signals/routes/__tests__/cron-connector-catalog.test.ts --maxWorkers=4 --silent=passed-only` (356 tests)
- `pnpm -F api lint`
- `pnpm -F api check-types`
- `pnpm knip --workspace api`
- pre-commit full-repository Prettier, Knip, and TypeScript checks

Closes #26136

Parent: #26075

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
@@ -17,6 +17,7 @@ import {
   type CreateCustomConnectorBody,
   ZERO_CUSTOM_CONNECTOR_IDS_ENV_KEY,
 } from "@vm0/api-contracts/contracts/zero-custom-connectors";
+import { testCustomConnectorSkillRepairStateContract } from "@vm0/api-contracts/contracts/test-custom-connector-skill-repair-state";
 import { FeatureSwitchKey } from "@vm0/core/feature-switch-key";
 import {
   getCustomConnectorSkillStorageName,
@@ -34,7 +35,8 @@ import { v5 as uuidv5 } from "uuid";
 
 import { mockEnv, mockOptionalEnv } from "../../../lib/env";
 import { clearMockNow, mockNow, now, nowDate } from "../../../lib/time";
-import { testContext } from "../../../__tests__/test-context";
+import { accept, testContext } from "../../../__tests__/test-context";
+import { setupApp } from "../../../__tests__/test-helpers";
 import { server } from "../../../mocks/server";
 import { flushWaitUntilForTest } from "../../context/wait-until";
 import { createDeferredPromise } from "../../utils";
@@ -124,6 +126,7 @@ import {
   type SecretKmsDataKey,
   type SecretKmsGenerateDataKeyRequest,
 } from "../../../lib/secret-kms-client";
+import { testCustomConnectorSkillRepairStateRoutes } from "../test-custom-connector-skill-repair-state";
 
 /**
  * RUN-01..04 and CHAIN-RUN: successful run dispatch and lifecycle.
@@ -9045,6 +9048,7 @@ describe("RUN-02: custom connectors, grants, and network policies", () => {
     const api = createRunsApi(context);
     createBddApi(context).acceptAgentStorageWrites();
     const connectors = createConnectorBddApi(context);
+    const storages = createStoragesBddApi(context);
     const { actor, agentId, runnerGroup } = await entitledRunActor();
     const slug = `_bdd-permission-skill-${randomUUID().slice(0, 8)}`;
     const custom = await connectors.createCustomConnector(actor, {
@@ -9072,6 +9076,10 @@ describe("RUN-02: custom connectors, grants, and network policies", () => {
       permissionBundleRef: "builtin:slack@1",
       skillMarkdown: "Use the selected Slack-compatible operations only.",
     });
+    const initialSkill = await storages.downloadStorage(actor, {
+      name: getCustomConnectorSkillStorageName(custom.id),
+      owner: "organization",
+    });
     const grant = {
       customConnectorId: custom.id,
       permissionNames: ["chat:write"],
@@ -9094,6 +9102,22 @@ describe("RUN-02: custom connectors, grants, and network policies", () => {
       prompt: "use the disconnected custom connector skill",
       modelProvider: "anthropic-api-key",
     });
+    await connectors.updateCustomConnector(actor, custom.id, {
+      displayName: custom.displayName,
+      prefixTemplates: custom.prefixTemplates,
+      fields: custom.fields,
+      headerInjections: custom.headerInjections,
+      queryInjections: custom.queryInjections,
+      authMode: custom.authMode ?? "manual",
+      permissionBundleRef: custom.permissionBundleRef,
+      skillMarkdown: "Use the updated Slack-compatible operations only.",
+      storageVersion: custom.storageVersion,
+    });
+    const updatedSkill = await storages.downloadStorage(actor, {
+      name: getCustomConnectorSkillStorageName(custom.id),
+      owner: "organization",
+    });
+    expect(updatedSkill.versionId).not.toBe(initialSkill.versionId);
     await api.heartbeatRunner(runnerGroup);
     const disconnectedClaim = await api.claimRunnerJob(disconnectedRun.runId);
     expect(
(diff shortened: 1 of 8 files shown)


seven332:
## Code Review: PR #26313

**Reviewed head:** `0d6f9da2c35765baa2402cb1a222b495dcce659e`

### Summary

The exact Custom connector skill association is carried through admission and resolved with shared fail-closed mechanics while retaining source-specific ownership. Ordinary and Feishu writers become statement-valid for the later nullable-pair constraint without changing public or runner contracts.

### Findings

No findings.

### Testing

The affected five API route suites pass on the rebased head, including update-after-admission pinning, invalid association rejection, Feishu exact mounting, writer/repair behavior, and Builtin coverage. API lint, API type checks, scoped Knip, and the pre-commit full-repository Prettier, Knip, 14/14 TypeScript tasks, file-size check, and commitlint also pass. Production rollout verification and the final constraint remain explicitly gated in #26137.

### Verdict

LGTM

PR #26342, opened 2026-08-11
feat: add a category rail to the chat thread emoji picker

tongx-blip:
## What

The thread emoji picker lists every group in one scroll, so reaching **Flags** means dragging past roughly 1,900 emoji, and each section title scrolls away with its own contents.

This adds three things to the existing `ChatThreadEmojiPicker`:

- **Category icon rail** above the search box, one tab per group plus *Frequently used*
- **Pinned section titles** — the title sticks to the top of the feed while its section is in view, fading out at its lower edge so emoji dissolve underneath instead of colliding with the text
- **Jump to category** — clicking an icon scrolls the feed to that section; the highlight follows the feed the rest of the time, so it always names whichever title is currently pinned

The data layer already had what was needed: `chatThreadEmojiGroups$` loads `unicode-emoji-json` by CLDR group, so the rail keys its icons off the same group names the sections are titled with.

## Why it is built this way

- **The jump holds the highlight.** A smooth scroll to a far category passes over every group in between, which flickers the highlight through the whole rail. The requested category is parked in `chatThreadEmojiPendingJump$` and the feed stops driving the highlight until `scrollTop` reaches the target.
- **No React local state or refs.** `views/` has no `useState`/`useRef` anywhere, so the active category and the pending jump live in ccstate signals next to the existing query state, and DOM access goes through a deterministic section id plus a `data-chat-thread-emoji-feed` marker — the same shape as `scrollSlashWorkflowIntoView`.
- **No tail spacer.** The last group is Flags with 270 emoji, far taller than the 288px feed, so it can already reach the top on its own.

## Verification

- `pnpm -F @vm0/app exec tsc --noEmit` — clean
- `pnpm -F @vm0/app lint` — clean
- `pnpm knip --workspace apps/platform` — clean
- New page test `chat-thread-emoji-category-rail.test.tsx` — 3 passed
- Existing `chat-history-navigation.test.tsx` (covers the emoji picker and its shortcuts) — 34 passed

Layout-dependent behaviour (the scroll spy, the pinned title, the smooth jump) has no coverage: jsdom reports every `offsetTop` as 0, so any assertion on it would pass without proving anything. Those three were verified by hand in a browser prototype instead.

## Feature switch

Gated behind `emojiPickerCategoryRail`, staff orgs only. With the switch off the rail is absent, titles are not sticky, and no scroll handler is attached.

## Decisions worth a look

- **Ten tabs, not nine.** `unicode-emoji-json` splits *Smileys & Emotion* from *People & Body*, where Slack merges them. Left as ten so the tabs match the section titles one-to-one; the popover stays `w-80`, which puts the buttons at ~28px rather than the 32px used in the prototype.
- **The rail stays visible while searching.** Clicking a tab clears the query and then jumps, rather than hiding the rail and shifting the popover height.
- **Section titles remain untranslated**, matching current behaviour — they come straight from the dataset. Only the rail's `aria-label` needed a new key.


tongx-blip:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 4bc8208f1a1a928b102a2875c86b517c91caa9a8

### Summary

Adds a category icon rail, pinned section titles, and jump-to-category scrolling to `ChatThreadEmojiPicker`, gated behind the new `emojiPickerCategoryRail` switch. The state placement is right for this codebase — `views/` has no `useState`/`useRef` anywhere, and the new active-category and pending-jump state sits in ccstate signals beside the existing query state, with DOM access through a deterministic section id and a feed marker.

The jump/scroll-spy handshake has a reachable dead state, though: the pending-jump lock is only ever cleared by a scroll event, so any jump that produces no scroll leaves the rail permanently desynchronised from the feed for the life of the popover.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness — no unreachable/dead states or wrong conditionals | Finding (P0) | `zero-chat-thread-page.tsx` `handleScroll`: `chatThreadEmojiPendingJump$` is cleared only inside the scroll handler, so a no-op jump never releases it |
| `REVIEW.md` | Type safety — no `any`, no overly broad types | Pass | `ChatThreadEmojiCategory`, `LucideIcon`, and the `string \| null` signals are explicit; no assertions added |
| `REVIEW.md` | Style — functions under ~100 lines, no duplicated logic | Pass | Rail, feed, and category builder were extracted; `ChatThreadEmojiPicker` is back under the 128-line lint ceiling |
| `REVIEW.md` | `feat:` commits carry integration tests | Pass | `chat-thread-emoji-category-rail.test.tsx` adds 3 page tests through `detachedSetupPage` |
| `REVIEW.md` | New user-facing feature gated behind a `FeatureSwitchKey` | Pass | `FeatureSwitchKey.EmojiPickerCategoryRail`, `enabled: false`, `STAFF_ORG_ID_HASHES` |
| `REVIEW.md` | Feature regression guard — existing flow covered, not just the new happy path | Pass | Third test asserts the switch-off picker keeps its current shape; existing `chat-history-navigation.test.tsx` emoji and Ctrl+Shift+N shortcut tests still pass unchanged (34) |
| `REVIEW.md` | Accessibility of new interactive controls | Finding (P1) | Rail uses `role="tablist"`/`role="tab"` without roving `tabIndex`, arrow-key handling, or `aria-controls` |
| `docs/bad-smell.md` | No `any` type | Pass | None introduced |
| `docs/bad-smell.md` | No lint/type suppressions | Pass | No `eslint-disable`, `@ts-ignore`, or `@ts-expect-error` in the diff |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | The pre-existing `await import("unicode-emoji-json/...")` in `chat-thread-emoji.ts` is untouched by this diff |
| `docs/bad-smell.md` | No unnecessary try/catch; fail fast | Pass | No error handling added |
| `docs/fallback.md` | Every fallback introduced must be declared in the PR summary | Pass | The diff introduces none — the switch gate is rollout containment, and `chatThreadEmojiCategoryIcon`'s `default` is a total function over a `string`-typed group name, not a guard on a contract-impossible state |
| `docs/fallback.md` | Switch-gated features need no compatibility code | Pass | No dual-read, migration, or old-shape tolerance added |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | The switch-off test asserts current shipped behavior, not deleted behavior |
| `docs/effect.md` | Render stays pure — no store writes, DOM mutation, or scheduling during render | Pass | `requestAnimationFrame` and the signal writes run only inside `onClick`/`onScroll` |
| `docs/cache.md` | RC-1 no render-phase side effects | Pass | Same as above |
| `docs/cache.md` | RC-5 no duplicate mutable sources of truth | Pass | Active category is written only by the scroll handler and the rail click; the feed's scroll position stays authoritative |
| `.claude/skills/ccstate/SKILL.md` | DOM callbacks use `detach(..., Reason.DomCallback)` for async commands | N/A | Both new commands are synchronous `set` calls and return void |
| `docs/testing.md` | Integration tests, not unit tests of internal functions | Pass | All three tests drive the real page via `detachedSetupPage` |
| `docs/testing.md` | Mock at the boundary; no relative-path `vi.mock`; no fake timers | Pass | No `vi.mock`, no timer mocking in the new file |
| `docs/testing/app-testing.md` | Page tests live in `views/` and enter through `detachedSetupPage` | Pass | `views/zero-page/__tests__/chat-thread-emoji-category-rail.test.tsx` |
| `docs/testing/testing-external-behavior.md` | Assert only on surfaces an external user can see | Finding (P1) | Tests assert on `[data-chat-thread-emoji-section="…"]`, an internal marker attribute |

### Findings

- `turbo/apps/platform/src/views/zero-page/zero-chat-thread-page.tsx`, `ChatThreadEmojiFeed.handleScroll` / `ChatThreadEmojiPicker.jumpToCategory`: the pending-jump lock can never be released when the jump scrolls nowhere, and the rail then stops following the feed permanently.

  `jumpToCategory` always sets `chatThreadEmojiPendingJump$`, but the only code that clears it lives inside `onScroll`. Two ordinary interactions produce no scroll event at all:

  1. **Clicking the already-pinned category.** Open the picker: `scrollTop` is `0` and the highlight is on *Frequently used*. Click the *Frequently used* tab → `scrollTo({ top: 0 })` on a feed already at `0` → browsers fire no scroll event → `pendingJump` stays `"frequently-used"`. Every later scroll hits the `if (pendingJump !== null)` guard, fails the `<= 1` proximity check against a section at `offsetTop 0`, and returns early. The highlight is frozen for the rest of the popover session.
  2. **Scrolling during the smooth animation.** A user scroll aborts an in-flight smooth scroll, so `scrollTop` never reaches the target and the same permanent lock results.

  The prototype this was ported from released the lock on `wheel`/`pointerdown`/`scrollend`; that release path did not survive the port. Suggest clearing `pendingJump` when the jump is a no-op (compare the target `offsetTop` against the current `scrollTop` before scrolling) *and* releasing it on user scroll input, so the lock cannot outlive the animation it exists for.

  There is a related latent coupling worth noting while fixing this: the same `<= 1` check silently assumes every category can reach the top of the feed. That holds today only because the last group is Flags with 270 emoji; if the dataset's last group ever became short, the lock would stick for the same reason.

- `turbo/apps/platform/src/views/zero-page/zero-chat-thread-page.tsx`, `ChatThreadEmojiCategoryRail`: the rail declares `role="tablist"`/`role="tab"` but does not implement the tab keyboard contract — no roving `tabIndex`, no arrow-key movement between tabs, no `aria-controls` pointing at the section each tab reveals. All ten tabs therefore sit in the sequential tab order, which is the opposite of the WAI-ARIA tabs pattern a screen-reader user will expect from that role. Either implement roving tabindex plus arrow keys, or drop to plain buttons in a labelled group if the tab semantics are not wanted.

- `turbo/apps/platform/src/views/zero-page/__tests__/chat-thread-emoji-category-rail.test.tsx`: `emojiSection()` asserts on `[data-chat-thread-emoji-section="Food & Drink"]`, and `docs/testing/testing-external-behavior.md` is explicit that platform tests must not assert on implementation surfaces — the marker attribute exists only to let the scroll code find sections. The user-visible equivalent is the rendered section title, so `screen.getByText("Food & Drink")` expresses the same assertion without freezing the marker. `aria-selected` assertions are fine to keep, since that is a real accessibility surface.

- `turbo/apps/platform/src/views/zero-page/zero-chat-thread-page.tsx`, `ChatThreadEmojiFeed`: the component subscribes to `chatThreadEmojiActiveCategory$` purely to compare inside `handleScroll`, so every category change re-renders the whole feed — roughly 1,914 emoji buttons, with no memo boundary between the feed and the grids. Scrolling the picker top to bottom crosses nine categories and therefore triggers nine full-subtree re-renders while the user is actively scrolling. Reading the current value at event time instead of subscribing keeps the highlight update confined to the rail.

### Testing

- **Coverage:** Adequate for the assertions that can be made — the `feat:` commit ships three page tests, and the existing emoji-picker and shortcut coverage in `chat-history-navigation.test.tsx` still passes (34 tests), which is the regression guard for the surface this PR modifies.
- **Gap the PR already discloses:** the three behaviours this PR exists to add — scroll spy, pinned title, smooth jump — have no automated coverage, because jsdom reports every `offsetTop` as `0`. That disclosure is honest and the right call, but it is also why the P0 above shipped: the dead-lock state is invisible to the suite. Worth confirming the fix by hand in a browser, since no test will catch a regression here.
- **Conventions:** Compliant except for the external-behavior violation noted above. No `vi.mock`, no fake timers, no direct `fetch` stubbing, no mock-call assertions; all cases enter through `detachedSetupPage` and drive the page the way a user would.

tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: 95fdf5de078b956069123e116fc4f01a30274ea6
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Re-review after `95fdf5d`, which addresses every finding from the review of `4bc8208`. The category rail, pinned section titles, and jump-to-category scrolling remain gated behind `emojiPickerCategoryRail`.

The P0 is properly closed rather than patched at the symptom. The hold is now armed only when `scrollChatThreadEmojiCategoryIntoView` reports that the feed will actually move, and it has three independent release paths — arrival, a missing target section, and user scroll input — so it cannot outlive the jump that created it. Clamping the jump target to the last reachable offset also removes the latent dependency on the final category being taller than the feed, which the previous review flagged as a hazard sitting behind the same code.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness — no unreachable/dead states or wrong conditionals | Pass | `jumpToCategory` arms the hold only on a real scroll; `handleScroll` releases on arrival or missing target; `releasePendingJump` covers wheel/touch/pointer |
| `REVIEW.md` | Type safety — no `any`, no overly broad types | Pass | `ReactKeyboardEvent` typed; `tabs.findIndex` comparison avoids an `as HTMLElement` assertion on `document.activeElement` |
| `REVIEW.md` | Style — functions under ~100 lines, no duplicated logic | Pass | `chatThreadEmojiScrollTarget` is shared by the jump and the arrival check, so the two cannot drift apart |
| `REVIEW.md` | `feat:` commits carry integration tests | Pass | 5 page tests, up from 3 |
| `REVIEW.md` | New user-facing feature gated behind a `FeatureSwitchKey` | Pass | `FeatureSwitchKey.EmojiPickerCategoryRail`, `enabled: false`, `STAFF_ORG_ID_HASHES` |
| `REVIEW.md` | Feature regression guard — existing flow covered | Pass | Switch-off test still asserts the unchanged picker; `chat-history-navigation.test.tsx` 34 tests pass |
| `REVIEW.md` | Accessibility of new interactive controls | Pass | Roving `tabIndex`, `ArrowLeft`/`ArrowRight` movement, `aria-controls` per section; covered by test |
| `docs/bad-smell.md` | No `any` type | Pass | None introduced |
| `docs/bad-smell.md` | No lint/type suppressions | Pass | None in the diff |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | Pre-existing dataset import untouched |
| `docs/bad-smell.md` | No unnecessary try/catch; fail fast | Pass | No error handling added |
| `docs/fallback.md` | Every fallback introduced must be declared in the PR summary | Pass | None introduced — the switch is rollout containment, and the `!target` release is a state-machine exit, not a tolerance branch for a shape the contract forbids |
| `docs/fallback.md` | Switch-gated features need no compatibility code | Pass | No dual-read, migration, or old-shape handling |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | The new negative assertion (`not.toBe("Frequently used")`) pins live behavior, not a deletion |
| `docs/effect.md` | Render stays pure | Pass | Signal writes and `requestAnimationFrame` stay inside `onClick`/`onScroll`/`onKeyDown` |
| `docs/cache.md` | RC-1 no render-phase side effects | Pass | Same as above |
| `docs/cache.md` | RC-5 no duplicate mutable sources of truth | Pass | Scroll position stays authoritative; the command drops no-op writes rather than mirroring |
| `.claude/skills/ccstate/SKILL.md` | DOM callbacks use `detach(..., Reason.DomCallback)` for async commands | N/A | All three commands are synchronous `set` calls |
| `docs/testing.md` | Integration tests, not unit tests of internal functions | Pass | All five drive the page through `detachedSetupPage` |
| `docs/testing.md` | Mock at the boundary; no relative-path `vi.mock`; no fake timers | Pass | Neither present |
| `docs/testing/app-testing.md` | Page tests in `views/`, entered through `detachedSetupPage` | Pass | `views/zero-page/__tests__/chat-thread-emoji-category-rail.test.tsx` |
| `docs/testing/testing-external-behavior.md` | Assert only on surfaces an external user can see | Pass | Assertions moved to section titles, `aria-selected`, `tabindex`, and focus; the remaining `[data-chat-thread-emoji-feed]` lookup drives a scroll rather than asserting, matching `chatScrollContainer()` in `chat-lifecycle-test-helpers.ts` |

### Findings

None. All four findings from the `4bc8208` review are resolved:

- **P0 stranded hold** — fixed at the state machine rather than the symptom. `scrollChatThreadEmojiCategoryIntoView` returns whether the feed moves, and `jumpToCategory` passes `null` when it does not, so the no-scroll jump never arms the hold. `handleScroll` additionally releases when the awaited section is gone, and `releasePendingJump` releases on `wheel`, `touchstart`, and `pointerdown`, which is what aborts an in-flight smooth scroll. The clamp in `chatThreadEmojiScrollTarget` makes the jump and the arrival check agree on one number, closing the short-last-category hazard noted alongside the P0.
- **P1 tablist keyboard contract** — roving `tabIndex`, arrow-key movement with wraparound, and `aria-controls` are in place.
- **P1 test boundary** — section-marker assertions replaced with the rendered section title.
- **P1 feed re-render** — the active-category subscription moved into the rail, and `setChatThreadEmojiActiveCategory$` now returns early on an unchanged value, so a scroll that does not change category performs no write at all.

Non-blocking nit for a follow-up, not worth holding this PR: while a search is active the sections unmount but the rail stays, so each tab's `aria-controls` points at an id that is not in the document. It degrades gracefully and no assistive-technology behavior in this flow depends on it.

### Testing

- **Coverage:** Adequate. Five page tests, including two that did not exist before: the stranded-hold regression and the arrow-key contract.
- **Regression test verified, not assumed:** jsdom reports every `offsetTop` as `0`, which would let a naive version of the stranded-hold test pass against the broken code. The test sets `feed.scrollTop = 300` first, and asserts jsdom retained it, so the two behaviours are actually distinguishable. Reverting the fix locally makes `keeps following the feed after a pick that scrolls nowhere` fail, and restoring it makes the file pass 5/5 — the test has teeth.
- **Still unautomatable:** the pinned title and the smooth scroll itself remain layout-dependent and therefore uncovered, as the PR summary states. The P0 is now covered because its symptom is observable through `aria-selected` without layout.
- **Local verification:** `pnpm format`, full `pnpm turbo run lint` (13/13), full `pnpm check-types` (14/14), `pnpm knip --workspace apps/platform`, rail suite 5/5, `chat-history-navigation.test.tsx` 34/34.
- **Conventions:** Compliant. No `vi.mock`, no fake timers, no `fetch` stubbing, no mock-call assertions.

PR #26382, opened 2026-08-11
feat: name the hovered emoji and tighten the picker section titles

tongx-blip:
Follow-up to #26342, from review of the shipped rail.

## What

**A name bar under the grid.** The rail landed without one, so the grid gave no way to read an emoji's name. A bar under the feed now names whichever emoji the pointer or keyboard is on, rendered as `:grinning_face:`, and shows a short hint when nothing is under the pointer.

**A tighter gap under the search field.** The first section title sat 16px below the search field, which read as a gap rather than a grouping. It is now 8px.

## Why it is built this way

- **One delegated listener, not ~1,900 pairs.** The hover source is a single `onMouseOver`/`onFocus` on the feed that resolves the emoji from `event.target.closest("[data-chat-thread-emoji]")`. Each button carries its glyph in that attribute because its `textContent` also contains the shortcut digit for the first nine frequently-used entries.
- **The preview lives in a signal, read only by the bar.** Pointing at a new emoji re-renders the name bar alone rather than the grid above it — the same reason the active category moved into the rail in #26342. The command also drops repeat writes, since moving across one emoji fires `mouseover` repeatedly.
- **The gap closes without touching the fade.** 4px comes off the search row's bottom padding, and the label shifts up inside its own box via `pt-1 pb-3` replacing `pt-2 pb-2`. Both add to 32px, so the box the gradient paints into is the same height as before. Measured in a browser against the real utility classes: title box 33px before and after, input-bottom-to-label 16px → 8px.
- **4px is the floor for the search row.** `Input` uses `focus:ring-[3px]` and the field is `autoFocus`, so a smaller bottom padding would put the focus ring over the first row of the feed.

## Verification

- `pnpm format`, `pnpm turbo run lint` (13/13), `pnpm check-types` (14/14), `pnpm knip --workspace apps/platform` — all clean
- `chat-thread-emoji-category-rail.test.tsx` — 6 passed, including a new case that hovers two emoji and asserts the bar names each one, and the idle hint appearing and disappearing
- `chat-history-navigation.test.tsx` — 34 passed
- The switch-off case now also asserts the bar is absent

The spacing itself has no automated coverage: jsdom has no layout, so any assertion on the gap would pass without proving anything. It was measured in a browser instead, against a reproduction using the same utility classes.

## Feature switch

Both changes sit behind the existing `emojiPickerCategoryRail` switch, still `enabled: false` and staff-only, so this stays one rollout unit with the rail rather than adding a second switch for the same redesign. Its description is updated to cover the name bar.

### DIFF
--- a/turbo/apps/platform/src/i18n/locales/de-DE/common.json
+++ b/turbo/apps/platform/src/i18n/locales/de-DE/common.json
@@ -1636,6 +1636,7 @@
       "openBrowser": "Browser öffnen",
       "openNamedAgent": "Agent {{name}} öffnen",
       "openNamedChat": "Chat {{title}} öffnen",
+      "pickEmoji": "Ein Emoji auswählen",
       "scrollToBottom": "Nach unten scrollen",
       "searchEmoji": "Emoji suchen"
     },
--- a/turbo/apps/platform/src/i18n/locales/en-US/common.json
+++ b/turbo/apps/platform/src/i18n/locales/en-US/common.json
@@ -1636,6 +1636,7 @@
       "openBrowser": "Open browser",
       "openNamedAgent": "Open agent {{name}}",
       "openNamedChat": "Open chat {{title}}",
+      "pickEmoji": "Pick an emoji",
       "scrollToBottom": "Scroll to bottom",
       "searchEmoji": "Search emoji"
     },
--- a/turbo/apps/platform/src/i18n/locales/es-ES/common.json
+++ b/turbo/apps/platform/src/i18n/locales/es-ES/common.json
@@ -1676,6 +1676,7 @@
       "openBrowser": "Abrir navegador",
       "openNamedAgent": "Abrir agente {{name}}",
       "openNamedChat": "Abrir chat {{title}}",
+      "pickEmoji": "Elige un emoji",
       "scrollToBottom": "Desplázate hasta abajo",
       "searchEmoji": "Buscar emoji"
     },
--- a/turbo/apps/platform/src/i18n/locales/fr-FR/common.json
+++ b/turbo/apps/platform/src/i18n/locales/fr-FR/common.json
@@ -1676,6 +1676,7 @@
       "openBrowser": "Ouvrir le navigateur",
       "openNamedAgent": "Ouvrir l'agent {{name}}",
       "openNamedChat": "Ouvrir le chat {{title}}",
+      "pickEmoji": "Choisissez un emoji",
       "scrollToBottom": "Faire défiler vers le bas",
       "searchEmoji": "Rechercher un emoji"
     },
--- a/turbo/apps/platform/src/i18n/locales/hi-IN/common.json
+++ b/turbo/apps/platform/src/i18n/locales/hi-IN/common.json
@@ -1636,6 +1636,7 @@
       "openBrowser": "खुला ब्राउज़र",
       "openNamedAgent": "ओपन एजेंट {{name}}",
       "openNamedChat": "चैट {{title}} खोलें",
+      "pickEmoji": "एक इमोजी चुनें",
       "scrollToBottom": "नीचे स्क्रॉल करें",
       "searchEmoji": "इमोजी खोजें"
     },
--- a/turbo/apps/platform/src/i18n/locales/id-ID/common.json
+++ b/turbo/apps/platform/src/i18n/locales/id-ID/common.json
@@ -1635,6 +1635,7 @@
       "openBrowser": "Buka browser",
       "openNamedAgent": "Buka agen {{name}}",
       "openNamedChat": "Buka chat {{title}}",
+      "pickEmoji": "Pilih emoji",
       "scrollToBottom": "Gulir ke bawah",
       "searchEmoji": "Cari emoji"
     },
--- a/turbo/apps/platform/src/i18n/locales/it-IT/common.json
+++ b/turbo/apps/platform/src/i18n/locales/it-IT/common.json
@@ -1676,6 +1676,7 @@
       "openBrowser": "Apri il browser",
       "openNamedAgent": "Apri agente {{name}}",
       "openNamedChat": "Apri la chat {{title}}",
+      "pickEmoji": "Scegli un'emoji",
       "scrollToBottom": "Scorri fino in fondo",
       "searchEmoji": "Cerca emoji"
     },
--- a/turbo/apps/platform/src/i18n/locales/ja-JP/common.json
+++ b/turbo/apps/platform/src/i18n/locales/ja-JP/common.json
@@ -1635,6 +1635,7 @@
       "openBrowser": "ブラウザを開く",
       "openNamedAgent": "エージェント {{name}} を開く",
       "openNamedChat": "チャットを開く {{title}}",
+      "pickEmoji": "絵文字を選択",
       "scrollToBottom": "一番下までスクロール",
       "searchEmoji": "絵文字を検索"
     },
--- a/turbo/apps/platform/src/i18n/locales/ko-KR/common.json
+++ b/turbo/apps/platform/src/i18n/locales/ko-KR/common.json
@@ -1635,6 +1635,7 @@
       "openBrowser": "브라우저 열기",
       "openNamedAgent": "에이전트 {{name}} 열기",
       "openNamedChat": "채팅 {{title}} 열기",
+      "pickEmoji": "이모지 선택",
       "scrollToBottom": "맨 아래로 스크롤",
       "searchEmoji": "이모지 검색"
     },
--- a/turbo/apps/platform/src/i18n/locales/pt-BR/common.json
+++ b/turbo/apps/platform/src/i18n/locales/pt-BR/common.json
@@ -1676,6 +1676,7 @@
       "openBrowser": "Abrir navegador",
       "openNamedAgent": "Abrir agente {{name}}",
       "openNamedChat": "Abrir conversa {{title}}",
+      "pickEmoji": "Escolha um emoji",
       "scrollToBottom": "Rolar até o final",
       "searchEmoji": "Buscar emoji"
     },
--- a/turbo/apps/platform/src/signals/chat-page/chat-thread-emoji.ts
+++ b/turbo/apps/platform/src/signals/chat-page/chat-thread-emoji.ts
@@ -71,6 +71,25 @@ export const setChatThreadEmojiPendingJump$ = command(
   },
 );
 
+// The emoji the pointer or keyboard is currently on, named in the preview bar
+// under the grid. Held in a signal so pointing at a new emoji re-renders that
+// bar alone rather than the whole grid it sits under.
+const internalChatThreadEmojiPreview$ = state<ChatThreadEmojiItem | null>(null);
+
+export const chatThreadEmojiPreview$ = computed((get) => {
+  return get(internalChatThreadEmojiPreview$);
+});
+
+export const setChatThreadEmojiPreview$ = command(
+  ({ get, set }, value: ChatThreadEmojiItem | null) => {
+    // Pointing at one emoji fires repeatedly as the pointer moves across it.
+    if (get(internalChatThreadEmojiPreview$)?.emoji === value?.emoji) {
+      return;
+    }
+    set(internalChatThreadEmojiPreview$, value);
+  },
+);
+
 export function filterChatThreadEmojiGroups(
   groups: ChatThreadEmojiGroup[],
   query: string,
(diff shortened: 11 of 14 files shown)


tongx-blip:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 86afff4822957a3f45126d78c6f71f03ac7d7b4f

### Summary

Adds a name bar under the emoji grid and halves the gap between the search field and the first section title, both behind the existing `emojiPickerCategoryRail` switch.

The delegated-listener design is the right call for a grid of ~1,900 buttons, and the spacing change is well constrained — the fade box stays 32px, and the 4px floor on the search row is justified by the `focus:ring-[3px]` on an `autoFocus` field.

The name bar is wrong on the one section that renders first, though. `Frequently used` items do not carry dataset emoji names: `useFrequentlyUsedEmoji` builds them as `{ emoji, name: labels[index] }` from `$.chat.thread.emoji.*`, which are localized display labels. Wrapping those in shortcode colons produces text that is not a shortcode in any locale.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness — no wrong output for a reachable input | Finding (P0) | `ChatThreadEmojiPreview` renders `:${name}:` over a localized label for every `Frequently used` entry |
| `REVIEW.md` | Type safety — no `any`, no overly broad types | Pass | `previewEmojiUnder` narrows with `instanceof Element`; signal typed `ChatThreadEmojiItem \| null` |
| `REVIEW.md` | Style — functions under ~100 lines, no duplicated logic | Pass | `ChatThreadEmojiPreview` extracted; one delegated handler instead of per-button closures |
| `REVIEW.md` | `feat:` commits carry integration tests | Finding (P1) | New preview test hovers only dataset-group emoji, never the `Frequently used` row where the bug lives |
| `REVIEW.md` | New user-facing feature gated behind a `FeatureSwitchKey` | Pass | Reuses non-GA `EmojiPickerCategoryRail`; description updated to cover the bar |
| `REVIEW.md` | Feature regression guard — existing flow covered, not just the new happy path | Pass | Switch-off case asserts both no rail and no name bar, and keeps the shipped `p-2`/`pt-2 pb-1` spacing |
| `REVIEW.md` | Accessibility of new interactive controls | Pass | Bar is presentational; the glyph is `aria-hidden` and buttons keep their `aria-label` |
| `docs/bad-smell.md` | No `any` type | Pass | None introduced |
| `docs/bad-smell.md` | No lint/type suppressions | Pass | None in the diff |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | None added |
| `docs/bad-smell.md` | No unnecessary try/catch; fail fast | Pass | No error handling added |
| `docs/fallback.md` | Every fallback introduced must be declared in the PR summary | Pass | None introduced |
| `docs/fallback.md` | Switch-gated features need no compatibility code | Pass | Correctly reuses the non-GA switch rather than adding compatibility for the previously shipped rail shape |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | The switch-off assertions pin live gated behavior |
| `docs/effect.md` | Render stays pure | Pass | Signal writes happen in `onMouseOver`/`onFocus`/`onMouseLeave` only |
| `docs/cache.md` | RC-1 no render-phase side effects | Pass | Same as above |
| `docs/cache.md` | RC-5 no duplicate mutable sources of truth | Pass | The preview holds a transient pointer to a static dataset row, not a synchronized copy of server state |
| `.claude/skills/ccstate/SKILL.md` | DOM callbacks use `detach(..., Reason.DomCallback)` for async commands | N/A | The new command is a synchronous `set` |
| `docs/testing.md` | Integration tests, not unit tests of internal functions | Pass | All six drive the page through `detachedSetupPage` |
| `docs/testing.md` | Mock at the boundary; no relative-path `vi.mock`; no fake timers | Pass | Neither present |
| `docs/testing/app-testing.md` | Page tests in `views/`, entered through `detachedSetupPage` | Pass | `chat-thread-emoji-category-rail.test.tsx` |
| `docs/testing/testing-external-behavior.md` | Assert only on surfaces an external user can see | Pass | Assertions use rendered text (`:grinning_face:`, `Pick an emoji`) and `aria-label` lookups |

### Findings

- `turbo/apps/platform/src/views/zero-page/zero-chat-thread-page.tsx`, `ChatThreadEmojiPreview`: the shortcode format is applied to names that are not shortcodes, so the first section of the picker names its emoji incorrectly in every locale.

  `chatThreadEmojiCategories` feeds the `Frequently used` section from `useFrequentlyUsedEmoji`, which sets `name` to a translated label — `$.chat.thread.emoji.done` and friends, i.e. `"Done"`, `"Urgent"`, `"Risk"`. Only the `unicode-emoji-json` groups carry lowercase dataset names like `"grinning face"`. `:${preview.name.replace(/\s+/g, "_")}:` therefore renders a display label dressed as a shortcode.

  Verified by rendering rather than by reading: hovering the ✅ in `Frequently used` produces the bar text `✅:Done:`. In `ja-JP` the same path yields `:完了:`, and `id-ID` gives `:Selesai:`.

  This is the default view — the picker opens on `Frequently used`, so these nine are the first emoji anyone points at, and the colons assert a shortcode that does not exist. Worth deciding deliberately between showing the dataset name for these nine, or dropping the colons when the name is a display label; either is fine, but the two name sources should not share one format that is only correct for one of them.

- `turbo/apps/platform/src/views/zero-page/__tests__/chat-thread-emoji-category-rail.test.tsx`: the new `names the emoji the pointer is on` case hovers `grinning face` and `watermelon`, both from `unicode-emoji-json` groups, so it exercises the only path where the format is already correct. Nothing covers the `Frequently used` row, which is the section the picker opens on and the one that is broken. Whatever fix is chosen for the P0, it needs a case that hovers a frequently-used entry and asserts the resulting text.

Non-blocking nit, not worth holding the PR: focus sets the preview but blur never clears it, so tabbing out of the grid leaves the last name in the bar. The pointer path is handled by `onMouseLeave`.

### Testing

- **Coverage:** Insufficient — missing a case for the `Frequently used` preview path, per the P1 above. Coverage is otherwise adequate: six page tests, with the new case asserting the idle hint appears, disappears on hover, and that the bar re-targets when the pointer moves to a second emoji.
- **Not automatable:** the spacing change is layout-dependent and jsdom has no layout, so the PR measured it in a browser against a reproduction using the same utility classes (title box 33px before and after, input-to-label 16px → 8px). That is the right substitute and the PR says so plainly.
- **Local verification reported:** `pnpm format`, `pnpm turbo run lint` 13/13, `pnpm check-types` 14/14, `pnpm knip --workspace apps/platform`, rail suite 6/6, `chat-history-navigation.test.tsx` 34/34.
- **Conventions:** Compliant. No `vi.mock`, no fake timers, no `fetch` stubbing, no mock-call assertions.

tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: 2eac206d9ac763d735db2c2c68c2706743f141c8
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Re-review after `2eac206`, which resolves the P0 and P1 from the review of `86afff4`.

The fix is placed correctly. Rather than special-casing the string at the point of display, the naming style became a property of the category, so the format is chosen where the data source is actually known: dataset groups keep `:grinning_face:`, and the frequently used row shows its translated product label as written. Reading the display name from `data-chat-thread-emoji-name` instead of `aria-label` is a real improvement beyond the bug — the accessible name and the displayed name are now separate concerns rather than one string doing both jobs.

The search-results grid renders without the prop and so defaults to shortcode formatting, which is correct: `filterChatThreadEmojiGroups` only ever walks the dataset groups, so search results never contain a product-labelled item.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness — no wrong output for a reachable input | Pass | Frequently used renders `Done`; dataset groups and search results keep `:grinning_face:` |
| `REVIEW.md` | Type safety — no `any`, no overly broad types | Finding (P1) | `chatThreadEmojiPreview$` is typed `ChatThreadEmojiItem` but now carries view-formatted text in `name` |
| `REVIEW.md` | Style — functions under ~100 lines, no duplicated logic | Pass | `chatThreadEmojiDisplayName` is the single place the format is applied |
| `REVIEW.md` | `feat:` commits carry integration tests | Pass | 7 page tests; the frequently used path is now covered |
| `REVIEW.md` | New user-facing feature gated behind a `FeatureSwitchKey` | Pass | Non-GA `EmojiPickerCategoryRail`, staff-only |
| `REVIEW.md` | Feature regression guard — existing flow covered | Pass | Switch-off case asserts no rail, no name bar, and the shipped spacing |
| `REVIEW.md` | Accessibility of new interactive controls | Pass | `aria-label` stays the dataset/product name and is no longer overloaded as display text |
| `docs/bad-smell.md` | No `any` type | Pass | None introduced |
| `docs/bad-smell.md` | No lint/type suppressions | Pass | None in the diff |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | None added |
| `docs/bad-smell.md` | No unnecessary try/catch; fail fast | Pass | No error handling added |
| `docs/fallback.md` | Every fallback introduced must be declared in the PR summary | Pass | None introduced; `shortcodeNames` selects between two live formats rather than tolerating a stale one |
| `docs/fallback.md` | Switch-gated features need no compatibility code | Pass | No compatibility path for the previously shipped bar |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | `not.toContain(":Done:")` pins current behavior against a live alternative format, not a deletion |
| `docs/effect.md` | Render stays pure | Pass | Signal writes remain in event handlers |
| `docs/cache.md` | RC-1 no render-phase side effects | Pass | Same as above |
| `docs/cache.md` | RC-5 no duplicate mutable sources of truth | Pass | The data attribute is render output, not a second mutable store |
| `.claude/skills/ccstate/SKILL.md` | DOM callbacks use `detach(..., Reason.DomCallback)` for async commands | N/A | The command is a synchronous `set` |
| `docs/testing.md` | Integration tests, not unit tests of internal functions | Pass | All seven drive the page through `detachedSetupPage` |
| `docs/testing.md` | Mock at the boundary; no relative-path `vi.mock`; no fake timers | Pass | Neither present |
| `docs/testing/app-testing.md` | Page tests in `views/`, entered through `detachedSetupPage` | Pass | `chat-thread-emoji-category-rail.test.tsx` |
| `docs/testing/testing-external-behavior.md` | Assert only on surfaces an external user can see | Pass | The new case hovers by `aria-label` and asserts on rendered bar text |

### Findings

The P0 and P1 from `86afff4` are both resolved:

- **P0 `:Done:`** — fixed at the source of the ambiguity rather than at the display. `ChatThreadEmojiCategory.shortcodeNames` decides the format where the data origin is known, the button carries the finished name, and the bar prints it verbatim.
- **P1 test gap** — `names a frequently used emoji by its product label, not as a shortcode` hovers the check mark and asserts the bar reads `Done` and not `:Done:`. Confirmed to have teeth: flipping `shortcodeNames` back to `true` fails it with `Expected: ":Done:" / Received: "✅:Done:"`.

- `turbo/apps/platform/src/signals/chat-page/chat-thread-emoji.ts`: `chatThreadEmojiPreview$` is declared `state<ChatThreadEmojiItem | null>`, but since this commit its `name` no longer holds an emoji name — for dataset groups it holds `":grinning_face:"`, which is display text. `ChatThreadEmojiItem` is the dataset contract shared with `chatThreadEmojiGroups$` and `filterChatThreadEmojiGroups`, so the same field now means "the emoji's name" in one place and "the string to paint in the bar" in another. Nothing misbehaves today, but the next reader of that signal has no way to tell which it is. A two-field local type for the preview — `{ emoji: string; label: string }` — would keep the dataset contract honest and cost nothing.

Non-blocking nit still open from the previous round: focus sets the preview but blur never clears it, so tabbing out of the grid leaves the last name in the bar. The pointer path is covered by `onMouseLeave`.

### Testing

- **Coverage:** Adequate. Seven page tests, including the previously missing frequently used path, and the switch-off case asserting the bar is absent.
- **Regression test verified, not assumed:** reverting `shortcodeNames` to `true` fails the new case; restoring passes 7/7.
- **Not automatable:** the spacing change remains layout-dependent and was measured in a browser (title box 33px unchanged, input-to-label 16px → 8px).
- **Local verification:** `pnpm format`, `pnpm turbo run lint` 13/13, `pnpm check-types` 14/14, `pnpm knip --workspace apps/platform`, rail suite 7/7, `chat-history-navigation.test.tsx` 34/34.
- **Conventions:** Compliant. No `vi.mock`, no fake timers, no `fetch` stubbing, no mock-call assertions.

Process note, outside the diff: lefthook reported `Can't find lefthook in PATH` on both commits in this branch, so pre-commit and commit-msg hooks did not run locally. Every check they would have run was run manually and CI re-runs them, so this branch is not at risk, but the hook installation looks broken in this environment and is worth fixing separately.

PR #26393, opened 2026-08-11
fix(billing): merge concurrency into plan subscription

hulh122:
## Summary

- add concurrency seats to the organization main Plan subscription instead of creating a separate Checkout subscription
- show an in-app confirmation with Stripe preview amounts before applying concurrency purchases or changes
- apply concurrency increases immediately with invoice payment handling when customer action is required
- schedule concurrency reductions and shared-item cancellation for the next billing period with no proration, refund, or account credit
- preserve existing subscription schedule items when revising a pending concurrency change
- keep successful purchase persistence webhook-authoritative instead of writing billing state synchronously in the purchase API
- reconcile shared subscription webhooks when the concurrency item is removed at renewal
- preserve legacy standalone concurrency subscription cancellation and restoration during rollout
- keep frontend and backend deployment compatibility with optional effective timing fields and the previous Checkout fallback
- show the scheduled effective date and no-refund behavior in the Billing confirmation UI across all supported locales

## Verification

- `pnpm -F api lint`
- `pnpm -F @vm0/app lint`
- `pnpm check-types` (14 tasks passed)
- `pnpm knip`
- `pnpm -F api exec vitest run --maxWorkers=4 --silent=passed-only` (191 files, 2948 tests passed)
- `pnpm -F @vm0/app exec vitest run --maxWorkers=4 --silent=passed-only` (134 files, 1621 tests passed)
- post-rebase API checkout and webhook regression tests (160 tests passed when isolated from shared-database cross-file interference)
- post-rebase Platform billing regression tests (33 passed)
- pre-commit formatting, Knip, full TypeScript checks, file-size validation, and commitlint passed


hulh122:
# PR Review Round 1

Found **1 P1 issue**:

- `zero-billing-concurrency-subscription.service.ts` treats any Stripe schedule attached to the shared Plan subscription as a concurrency-owned schedule. A reduction can rewrite an unrelated Plan schedule, while an increase can release it entirely. This can silently discard a scheduled Plan downgrade/cancellation and leave the local pending Plan state stale. Please only revise/release schedules known to belong to the pending concurrency change; otherwise return a conflict. Add an integration test for an unrelated attached schedule.

No P0 issues found. The migration, webhook routing, type safety, and test boundaries otherwise look good.

hulh122:
## PR Review Round 2\n\nLGTM. The schedule-ownership guard prevents concurrency changes from mutating unrelated Plan schedules, and the API integration coverage now covers initial purchase, preview, confirmation, cancellation, and concurrency-owned schedule release. The rebased 0917 migration applies cleanly from an empty database; affected tests, lint, type checking, and Knip pass.

PR #25529, opened 2026-08-06
feat: migrate legacy subscriptions to usage pack plans

hulh122:
## Summary

- add an admin preview and confirmation flow that replaces legacy Stripe subscription items in place
- keep legacy subscribers on the legacy Plan UI and expose usage-pack migration through a separate action
- follow the current billing-change model with in-app confirmation, Stripe pending updates, activation after `invoice.paid`, and cron reconciliation
- preserve legacy credits and include converted pending invitations in the paid expiry, refund, and acceptance lifecycle
- prevent legacy customers from creating a second subscription while migration state is loading or available

## Fallbacks

- `usage-pack-subscription-migration.service.ts:usagePackSubscriptionMigrationSchemaAvailable` protects ordinary Stripe webhook and cron processing when a new API briefly precedes migration `0907_majestic_sheva_callister`. Surface: DB vs API, observed maximum ~102 minutes. Remove after `0907` is deployed and outside the production rollback window; follow-up #26388.

## Validation

- `pnpm -F api lint`
- `pnpm -F @vm0/app lint`
- `pnpm -F @vm0/api-contracts lint`
- `pnpm -F @vm0/db lint`
- affected-workspace type checks and Knip
- `pnpm -F @vm0/db test:migration-consistency`
- `pnpm -F api exec vitest run src/signals/routes/__tests__/zero-billing-checkout.test.ts --maxWorkers=1 --silent=passed-only`
- `pnpm -F @vm0/app exec vitest run src/views/zero-page/__tests__/org-billing-tab.test.tsx --maxWorkers=1 --silent=passed-only`

Closes #25415

PR #26400, opened 2026-08-11
perf(ui): stop the dialog scrim re-blurring the viewport every frame

tongx-blip:
## Why

Two reports on the settings modal: the backdrop blur reads far too strong, and opening/closing stutters.

**The blur is twice what was intended.** #25719 describes adding "a 4px backdrop blur" and reaches for `backdrop-blur-sm` — the Tailwind v3 value. This repo is on Tailwind v4, where the scale was renamed (`--blur-xs: 4px`, `--blur-sm: 8px`), so the shipped blur is 8px. Fitted against the two screenshots below, Claude's equivalent scrim sits at roughly 2px; ours smears background text past legibility.

**The stutter is the blur, but not because of its radius.** Measured in Chromium at DPR 2, mean frame interval with a scrim up:

| | mean ms/frame |
|---|---|
| page still, with or without blur | 16.7 |
| something repaints, blur 2px | 255 |
| something repaints, blur 8px | 262 |
| something repaints, no blur | 32 |

A `backdrop-filter` costs nothing while the page behind it holds still — the compositor caches it. But any frame that repaints a pixel anywhere on screen re-blurs the whole viewport, whether the change is above or below the scrim, and promoting the dialog to its own compositing layer does not help (280.6 vs 280.8 ms).

Two things kept our page from holding still, and neither applies to Claude's modal:

1. The scrim faded **its own opacity**, so it dirtied itself on every frame of both the 180ms enter and exit.
2. The page keeps always-on animations running behind the dialog — a breathing dot per running thread, the shimmer placeholders, the thinking blocks, the mic spinner, the workflow beam. Any one of them on screen re-blurred the full viewport at 60Hz for as long as the dialog stayed open.

## What changed

- Blur drops from 8px to **2px** and moves onto `.zero-dialog-overlay::after`, its own layer, which stays hidden until the enter fade has finished and hides again the moment the dialog starts closing. Both transitions now run with no blur on screen; the look is unchanged once settled.
- The always-on indicators park (`animation-play-state: paused`) while a scrim is mounted. Anything rendered **inside** the dialog keeps animating — the command palette's running dots still breathe, since that is state the user is looking at.
- The two dialogs that skip the enter animation now use a shared `zero-dialog-overlay-instant` class instead of a `data-open:!animate-none` utility, so the scrim and its blur stay in step rather than the blur trailing by 180ms.

## Verification

Computed styles through the transition, confirmed in a browser:

| | during fade | settled |
|---|---|---|
| `::after` visibility | `hidden` | `visible` |
| `::after` backdrop-filter | `blur(2px)` | `blur(2px)` |
| background indicator | `paused` | `paused` |

`prefers-reduced-motion` skips the fade, so the blur lands with the scrim there.

## Not in this PR

`dropdown-menu`, `popover`, `select` and `tooltip` all carry `data-open:animate-in data-closed:animate-out fade-*/zoom-*/slide-in-from-*`, but neither `tailwindcss-animate` nor `tw-animate-css` is a dependency and Tailwind v4 core has no `animate-in` — those classes compile to nothing, so every popup hard-cuts in and out. Giving them real motion makes Base UI wait for an exit animation before unmounting, which changes close timing across the app and wants its own test pass, so it is a separate PR.


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: 0c883947fcdba888e9a71f1d74385bcf20c8f6bb
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Corrects the dialog scrim blur from an unintended 8px back to 2px, moves it onto `.zero-dialog-overlay::after` so it is not painted during the enter fade or the exit, and pauses the page's always-on animations while a scrim is mounted so the compositor can keep the blurred backdrop cached. Presentation-only: no signals, data flow, contracts, or persisted state are touched.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic — no wrong assumptions or unhandled paths | Pass | Exit path verified in a browser: with `[data-ending-style]` present, `::after` computes `visibility: hidden` both when `data-open` persists alongside it and when it does not, so the blur is gone for the whole exit either way. |
| `REVIEW.md` | Security (OWASP, secrets, auth) | N/A | Diff is CSS and two className string literals; no input handling, network, storage, or auth surface. |
| `REVIEW.md` | Type safety — no `any`, no broad types | Pass | No type declarations changed; `dialog.tsx` edits a string literal inside the existing `cn()` call. |
| `REVIEW.md` | Style & maintainability — no oversized functions or duplicated logic | Pass | Net +58 CSS lines, no function bodies changed. The `zero-dialog-overlay-instant` class replaces a duplicated `data-open:!animate-none` literal at two call sites with one named rule. |
| `REVIEW.md` | Interface changes documented | Pass | `DialogOverlay`'s prop surface is unchanged. See note 2 for a behavioral narrowing of `overlayClassName` that is worth recording. |
| `REVIEW.md` | Feature regression guard — existing behavior not silently changed | Pass | The changed behavior is the scrim's own appearance. `packages/ui/src/components/ui/__tests__/dialog.test.tsx` still pins the `zero-dialog-overlay` / `zero-dialog-content` contract both paths depend on, and no test asserts the replaced `data-open:!animate-none` literal, so the class swap breaks no existing assertion. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, contract, Zod, or `jsonb().$type<>()` change. |
| `docs/bad-smell.md` | No `any`, no lint/type suppression comments | Pass | No `eslint-disable`, `oxlint-disable`, `@ts-ignore`, or `prettier-ignore` added. |
| `docs/bad-smell.md` | No dynamic `import()`, no hardcoded URLs or env values | Pass | No imports or configuration touched. |
| `docs/bad-smell.md` | Fail fast — no fallback/recovery slop | Pass | No `??`/`||` chain, no defaulting, no try/catch added. |
| `docs/fallback.md` | Every fallback introduced must be declared, justified, and time-boxed | Pass | The diff introduces no fallback: no compatibility branch, no tolerance for an old shape, no version-skew handling, no negative test. No `Fallbacks` section is required, and none is present. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | No tests added or removed. |
| `docs/testing.md` | Integration-first; required coverage for the commit type | Pass | `perf:` change whose entire effect is compositor behavior — `:has()`, `backdrop-filter`, and `animation-play-state` are not evaluated by happy-dom, so there is no honest assertion to add at the existing test layer. Existing dialog tests still cover the class contract. See Testing. |
| `docs/testing.md` | Mock boundary, MSW, real DB, no fake timers | N/A | No test files in the diff. |
| `docs/effect.md` | Keep render pure; choose the right execution boundary (`computed` / command / `onRef` / effect) | N/A | Both platform `.tsx` edits replace a string literal inside existing JSX; no effect, command, computed, or render-phase logic is added or moved. |
| `docs/cache.md` | RC-1..RC-9 — no render-phase side effects, render-time signal allocation, unbounded caches, or asymmetric resource lifecycles | N/A | No signals, refs, caches, or subscriptions are created, owned, or torn down by this diff. |
| `.claude/skills/ccstate/SKILL.md` | ccstate patterns — reactive computed over imperative fetch, explicit `accept`, AbortSignal ownership | N/A | No ccstate atoms, commands, or HTTP calls in the diff. |
| `docs/react-commit.md` | Do not confuse DOM mutations or component executions with React commits when claiming performance work | Pass | The PR does not claim a React-level win; it attributes the cost to compositor re-blur and measures frame intervals, not commits. |
| `docs/react-commit.md` | Reproducible scenario, repeated runs, raw results retained, dev/prod timings not conflated | Finding (P1-adjacent, non-blocking) | 3 runs × 30 frames per case with a warmup, and the table reports the comparison the change rests on. But the harness is synthetic and the renderer was SwiftShader, which the PR body does not state. See note 1. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Notes (non-blocking)

1. **Measurement environment is not stated in the PR body.** The numbers were produced in headless Chromium on a synthetic 1400-row page with `ANGLE (SwiftShader)` — software rasterization, no GPU. The *ordering* and the structural conclusion (idle blur is free at 16.7ms; any repaint re-blurs the viewport; radius barely matters; layer promotion does not help) are what the change rests on and they hold, but the absolute millisecond figures do not transfer to a GPU-backed client and reading them as such would overstate the win. Worth one line in the PR body so a later reader does not treat 255ms as a real-device number.

2. **`overlayClassName` can no longer switch the blur off.** Previously the blur was a `backdrop-blur-sm` utility in the base class list, so a caller could override it with another `backdrop-blur-*` utility. It now lives in `.zero-dialog-overlay::after`, which no `overlayClassName` utility can reach. No current caller does this — the only two `overlayClassName` values in the repo are the `zero-dialog-overlay-instant` ones this PR introduces — so nothing regresses, but the escape hatch is gone.

3. **`zero-dialog-overlay-instant` wins on source order, not specificity.** `.zero-dialog-overlay-instant[data-open]` and `.zero-dialog-overlay[data-open]` are both `(0,2,0)`; the override works because it is declared later in the same `@layer components` block. Verified correct as written (`animation-name` computes to `none`, `::after` computes `visible`), but the previous `!important` form was order-independent. Moving these rules apart would silently break it.

4. **The in-dialog exemption is correct at every current call site.** `:not(.zero-dialog-content *)` was checked against each paused class: `CommandDialog` renders through `DialogContent`, so the `RunningIndicator` rows in `AgentListDialog` keep breathing; the onboarding workflow diagram is `DialogContent` `children`, so `owf-diagram-beam` keeps flowing; the shimmer, thinking-block, and mic-spinner sites are all page-level. Note that a Base UI popup portaled out of a dialog (`Select`, `Popover`) lands outside `.zero-dialog-content` and would be paused — none of the eight listed classes renders in one today.

### Testing

- **Coverage:** Adequate for the change type. This is a `perf:` change with no `feat:`/`fix:` test obligation under `REVIEW.md`, and its entire observable effect — `:has()` matching, `backdrop-filter`, staged `visibility`, `animation-play-state` — is compositor and computed-style behavior that happy-dom does not evaluate, so a unit test could only re-assert the class strings it already asserts. Verified manually in a browser instead: during the fade `::after` is `hidden` with the background indicator `paused`; after 180ms `::after` is `visible` at `blur(2px)`; on exit `::after` is `hidden` under both `data-ending-style` attribute combinations; the instant variant paints its blur immediately with the overlay animation disabled.
- **Conventions:** Compliant. No test files added or modified, so no mock-boundary, MSW, real-database, fake-timer, or mock-cleanup rule applies.

tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: a6953e04dd9a789bb1f5fe6f7db287a63e5a1156
PR_AUTO_REVIEW_VERDICT: LGTM

Re-review at `a6953e0`. Since the `0c88394` review the branch merged `origin/main` and added one commit, `test(ui): match the css rule a selector introduces`. The scrim change itself is unchanged; this pass covers the new test-helper commit and re-checks the merged state.

### Summary

Corrects the dialog scrim blur from an unintended 8px to 2px, moves it onto `.zero-dialog-overlay::after` so it is never painted during the enter fade or the exit, and pauses the page's always-on animations while a scrim is mounted so the compositor can keep the blurred backdrop cached. The follow-up commit repairs `getCssBlock` in `running-indicator.test.tsx`, which the new rule's `:is()` selector list broke.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic — no wrong assumptions or unhandled paths | Pass | Exit path verified in a browser: with `[data-ending-style]`, `::after` computes `visibility: hidden` whether or not `data-open` persists alongside it. `getCssBlock`'s new loop terminates on every path — advance on a rejected occurrence, `break` on a missing or unterminated brace, `throw` when the scan ends. |
| `REVIEW.md` | Correctness — merged-state check against incoming `main` | Pass | `main` added `forceRender` to `<DialogOverlay>` (#25529). Base UI gates it as `enabled: forceRender || !nested` (`dialog/backdrop/DialogBackdrop.js`), so it only decides whether a *nested* dialog paints its own backdrop; it does not keep the backdrop mounted while closed. The `html:has(.zero-dialog-overlay)` premise therefore still holds, and `zero-sidebar-account-menu.test.tsx:882` already asserts the overlay is absent after close. |
| `REVIEW.md` | Security (OWASP, secrets, auth) | N/A | CSS, two className literals, and a test-local string scanner over a file read from the package's own `src/styles`. No input handling, network, storage, or auth surface. |
| `REVIEW.md` | Type safety — no `any`, no broad types | Pass | `getCssBlock(selector: string)` keeps its signature and inferred `string` return; no assertions or widened types added. |
| `REVIEW.md` | Style & maintainability | Pass | Helper grew 15→36 lines but stayed one function with a doc comment naming the reason. `zero-dialog-overlay-instant` replaces a duplicated `data-open:!animate-none` literal at two call sites. |
| `REVIEW.md` | Interface changes documented | Pass | No prop surface changed. See note 2 on `overlayClassName` losing the ability to override the blur. |
| `REVIEW.md` | Feature regression guard | Pass | The changed behavior is the scrim's own appearance. `dialog.test.tsx` still pins the `zero-dialog-overlay` / `zero-dialog-content` contract, and no test asserts the replaced `data-open:!animate-none` literal. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, contract, Zod, or `jsonb().$type<>()` change. |
| `docs/bad-smell.md` | No `any`, no lint/type suppression comments | Pass | No `eslint-disable`, `oxlint-disable`, `@ts-ignore`, or `prettier-ignore` added; `pnpm -F @vm0/ui lint` is clean. |
| `docs/bad-smell.md` | No dynamic `import()`, no hardcoded URLs or env values | Pass | No imports or configuration touched. |
| `docs/bad-smell.md` | Fail fast — no fallback/recovery slop | Pass | The helper still throws when no rule matches; the rewrite narrows *which* occurrence counts rather than tolerating a miss. |
| `docs/fallback.md` | Every introduced fallback declared, justified, time-boxed | Pass | The diff introduces no fallback: no compatibility branch, no old-shape tolerance, no version-skew handling. No `Fallbacks` section required, and none present. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | No tombstone test added; the twelve existing assertions are byte-identical and none asserts an absence. |
| `docs/testing.md` | Integration-first; required coverage for the commit type | Pass | `perf:` change whose effect is compositor behavior — `:has()`, `backdrop-filter`, and `animation-play-state` are not evaluated by happy-dom. The test commit repairs an existing guard rather than adding a new layer. See Testing. |
| `docs/testing.md` | Never weaken a test to make CI pass | Pass | Only the rule-lookup changed. All twelve assertions and both `it` blocks are unchanged; no `skip`, no relaxed matcher, no deleted case. Confirmed the repaired helper resolves the real block and not `animation-play-state: paused`. |
| `docs/testing/anti-patterns.md` | AP-1 mock-call assertions, AP-2 fetch mocking, AP-3 fs mocking, AP-4/AP-6 internal mocks, AP-5 fake timers | Pass | None added. The file's only mock is a pre-existing `vi.spyOn(Date, "now")` with `vi.restoreAllMocks()` in `afterEach`; `readFileSync` is real, not mocked. |
| `docs/testing/anti-patterns.md` | AP-7 implementation-detail testing, AP-10 direct component rendering | Pre-existing | Both `it` blocks render `RunningIndicator` directly and assert CSS text. Pre-existing structure that this PR neither introduces nor widens; rewriting it is out of scope for a scrim change. |
| `docs/testing/patterns.md` | Standard test file structure; mock external, use real internal | Pass | Structure unchanged: module-scope fixture load, helper, `describe`/`it`. The CSS file is read for real from the package's own source. |
| `docs/effect.md` | Render purity; correct execution boundary | N/A | Both platform `.tsx` edits replace a string literal inside existing JSX; no effect, command, or computed added or moved. |
| `docs/cache.md` | RC-1..RC-9 lifecycle and ownership rules | N/A | No signals, refs, caches, or subscriptions created, owned, or torn down. |
| `.claude/skills/ccstate/SKILL.md` | ccstate patterns — reactive computed, explicit `accept`, AbortSignal ownership | N/A | No ccstate atoms, commands, or HTTP calls in the diff. |
| `docs/react-commit.md` | Do not confuse DOM mutations or component executions with React commits | Pass | No React-level claim is made; the cost is attributed to compositor re-blur and measured as frame intervals. |
| `docs/react-commit.md` | Reproducible scenario, repeated runs, raw results retained | Finding (non-blocking) | 3 runs × 30 frames per case with a warmup, but the harness is synthetic and the renderer was SwiftShader, which the PR body still does not state. See note 1. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Notes (non-blocking)

1. **Measurement environment is still unstated in the PR body.** Numbers came from headless Chromium on a synthetic 1400-row page with `ANGLE (SwiftShader)` — software rasterization, no GPU. The ordering and the structural conclusion are what the change rests on and they hold, but the absolute figures do not transfer to a GPU-backed client. One line in the PR body would stop a later reader treating 255ms as a real-device number.

2. **`overlayClassName` can no longer switch the blur off.** The blur moved from a `backdrop-blur-*` utility in the base class list into `.zero-dialog-overlay::after`, which no utility class can reach. No caller relies on it — the only two `overlayClassName` values in the repo are the `zero-dialog-overlay-instant` ones this PR introduces — but the escape hatch is gone.

3. **`zero-dialog-overlay-instant` wins on source order, not specificity.** `.zero-dialog-overlay-instant[data-open]` and `.zero-dialog-overlay[data-open]` are both `(0,2,0)`; the override works because it is declared later in the same `@layer components` block. Verified correct as written; the previous `!important` form was order-independent, so moving these rules apart would silently break it.

4. **`globals.test.ts` carries the same latent helper flaw.** `readRuleBody` in `packages/ui/src/styles/__tests__/globals.test.ts` also takes the first textual occurrence. It passes today because none of its selectors appear in another rule's selector list, so it is untouched here by design, but the next grouped selector that names a `--state-*` or theme class will break it the same way.

5. **The in-dialog exemption holds at every call site.** `:not(.zero-dialog-content *)` was checked against each paused class: `CommandDialog` renders through `DialogContent`, so `AgentListDialog`'s `RunningIndicator` rows keep breathing; the onboarding workflow diagram is `DialogContent` `children`, so `owf-diagram-beam` keeps flowing; the shimmer, thinking-block, and mic-spinner sites are page-level. A Base UI popup portaled out of a dialog would land outside the exemption, but none of the eight listed classes renders in one today.

### Testing

- **Coverage:** Adequate. The scrim change is compositor behavior that happy-dom cannot evaluate, so there is no honest assertion to add at the existing layer; it was verified manually in a browser instead (fade: `::after` `hidden` with the background indicator `paused`; settled: `visible` at `blur(2px)`; exit: `hidden` under both `data-ending-style` attribute combinations; instant variant: blur paints immediately with the overlay animation disabled). The follow-up commit restores a real guard that the CSS change had silently disarmed, which is the correct direction.
- **Conventions:** Compliant. `pnpm -F @vm0/ui exec vitest run` passes 10 files / 71 tests, `pnpm -F @vm0/ui lint` is clean, `pnpm -F @vm0/ui check-types` exits 0, and `pnpm format` reports no further changes.

PR #26402, opened 2026-08-11
fix(api): keep chat event cursors consistent on conflicts

### LINKED ISSUE #26358: bug: keep chat event cursors consistent across idempotent conflicts
## Background

`chat_threads.last_chat_event_seq_id` is an allocation watermark, not a count of physical rows. `chat_event_search_watermarks.indexed_seq_id` and snapshot `last_seq_id` are range-coverage cursors. Gaps are therefore valid as long as consumers treat these values as watermarks.

`insertChatEvents` reserves a range, performs one bulk `INSERT ... ON CONFLICT DO NOTHING`, and returns only accepted rows. `replaceLoadedChatEvent` similarly reserves one position before a conflict-tolerant insert.

## Problem and evidence

When a conflict rejects the final reserved value, the committed thread watermark can end above the latest physical event. A fully duplicated assistant batch is the simplest example: it advances the watermark while inserting no canonical event.

The search projector already handles this correctly. It marks the allocated range as examined and later selects new rows strictly above that coverage cursor. The snapshot rebuild does not: it throws unless its final physical row exists exactly at `indexed_seq_id`. One sparse tail can therefore abort the snapshot cron pass.

The conflict-tolerant batch behavior is already deployed in production, so writer changes alone cannot guarantee that no sparse tail exists. Independently rewinding persisted thread cursors is unsafe because search or snapshot watermarks may already cover those positions.

Retry-only watermark movement is also unnecessarily expensive. It schedules search projection and can rebuild and upload a full-thread snapshot even though canonical event content did not change.

## Decision

Keep sequence IDs sparse and monotonic. Do not add a pre-insert conflict lookup or try to compact interior gaps. Reclaim only the unused suffix of a reservation while the transaction still owns the thread lock, and make snapshot generation honor the existing range-coverage contract.

This is the smallest solution that addresses correctness, deployed-state compatibility, and retry-driven resource churn without adding work to the conflict-free assistant write path.

## Chosen direction

1. Preserve the existing bulk insert and conflict handling. After `RETURNING`, compute the highest accepted sequence per affected thread and release only reserved positions above it. If a thread inserted no rows, release its entire reservation. Interior rejected positions remain gaps.
2. Perform suffix release inside the caller-owned transaction while the reservation lock is still held. Process threads in the existing deterministic order. Do not depend on PostgreSQL `RETURNING` row order.
3. Release the single reserved position when `replaceLoadedChatEvent` loses its expected replacement/revocation race.
4. In snapshot rebuilds, require at least one physical row but not a row exactly at the indexed watermark. Archive the physical rows through that watermark and publish `indexed_seq_id` as the snapshot coverage cursor.
5. Add focused integration coverage for duplicate batch suffix reclamation, the existing replacement race, and snapshot/tail behavior with a prior-version sparse watermark.

No pre-allocation duplicate query is added. A conflict-free batch keeps the current database operations. A conflicting batch adds only the suffix-release update needed to prevent retry-only downstream work.

## Alternatives considered

- **Query existing assistant events before every allocation:** prevents all new gaps, but adds an indexed read to every assistant batch and extends the hot-path lock section even when there is no retry. Density is not required by the data model.
- **Insert each event separately:** can release every rejected position but replaces one bulk insert with per-event database round trips.
- **Only relax the snapshot assertion:** fixes correctness but lets every duplicate retry advance projection and snapshot coverage, potentially rebuilding an unchanged full-thread archive.
- **Only change writers:** leaves already-deployed sparse state capable of failing snapshot generation.
- **Renumber rows or rewind persisted cursors:** adds data mutation and can reuse a position already covered by a derived watermark.

## Risks and mitigations

- **Partial conflicts:** release only the unused suffix after the maximum accepted sequence. Never renumber accepted rows or reuse an interior gap.
- **Multiple threads in one batch:** derive release counts per thread and use the same sorted thread order used for reservation.
- **Concurrent writers:** reservation and release remain in one transaction while the thread rows are locked, so no writer can allocate from an intermediate watermark.
- **Sparse snapshot cursor:** the schema already defines it as coverage; the raw-row endpoint already accepts a missing cursor equal to the current snapshot head, and the client tails from the separately returned coverage cursor.
- **Old/new API overlap:** old instances may still create sparse tails during rollout, while the new snapshot reader accepts them without a schema or payload change.
- **Performance:** the successful bulk-write path gains no query or update. Only conflict cases perform suffix release, avoiding unnecessary search and full-archive work.

## Acceptance criteria

- A conflict-free assistant batch retains one reservation update and one bulk insert, with no conflict lookup or release update.
- Replaying a fully persisted batch inserts zero rows and leaves the thread watermark unchanged.
- A partially conflicting batch preserves accepted-row order, leaves interior gaps untouched, and reclaims only unused trailing positions.
- Concurrent deliveries remain serialized by the existing thread-row reservation lock.
- A lost replacement/revocation race returns `null` without consuming a committed position.
- Snapshot rebuild succeeds when `indexed_seq_id` is above the maximum physical event sequence and publishes that indexed value as its coverage cursor.
- A client can tail from the sparse snapshot cursor and receive a later event.
- No schema, API shape, persisted-row migration, cursor rewrite, or per-event insert loop is introduced.

## Relevant code

- `turbo/apps/api/src/signals/services/zero-chat-event.service.ts`
- `turbo/apps/api/src/signals/services/cron-snapshot-chat-events.service.ts`
- `turbo/apps/api/src/signals/routes/__tests__/chat-callbacks.bdd.test.ts`
- `turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts`
- `turbo/apps/api/src/signals/routes/__tests__/cron-snapshot-chat-events.test.ts`
- `turbo/apps/api/src/signals/routes/__tests__/zero-chat-event-snapshot.test.ts`

## Scope

This remains one independently reviewable PR and is not a delivery parent. It is independent of durable active-input delivery activation in #26060.

seven332:
## Summary

- reclaim only the unused suffix of reserved chat event sequence IDs after idempotent batch conflicts
- release a losing replacement claim without renumbering already accepted events
- publish snapshots at their indexed coverage cursor when historical or valid interior sequence gaps leave the final archive row behind that cursor
- cover repeated batches, replacement races, sparse snapshot coverage, and client tailing through integration tests

## Scope

- preserves the existing bulk insert path and adds no database round trip for conflict-free batches
- skips suffix analysis entirely when every submitted row is accepted
- preserves interior sequence gaps because accepted rows are immutable
- does not change the database schema, migrate persisted data, or change a public API

## Validation

- `pnpm format`
- `pnpm -F api lint`
- `pnpm -F api check-types`
- `pnpm -F @vm0/app lint`
- `pnpm knip --workspace api --workspace @vm0/app --workspace @vm0/api-contracts`
- `pnpm -F api exec vitest run --maxWorkers=4 --silent=passed-only` (187 files, 2,898 tests)
- `pnpm -F @vm0/api-contracts test` (27 files, 567 tests)
- `pnpm -F @vm0/app exec vitest run src/signals/chat-page/__tests__/chat-event-snapshot-read.test.ts` (7 tests)
- post-rebase focused API regressions (3 tests across the repeated-batch, replacement-race, and sparse-snapshot suites)
- post-self-review API lint, API type checks, Knip, and the repeated-batch regression

Closes #26358

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/chat-callbacks.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/chat-callbacks.bdd.test.ts
@@ -3544,7 +3544,7 @@ describe("CHAT-02: chat output extraction and progress callbacks", () => {
     ).toBeFalsy();
   }, 30_000);
 
-  it("keeps repeated at-least-once event batches idempotent", async () => {
+  it("reclaims only unused suffixes from repeated event batches", async () => {
     const { actor, agentId, runnerGroup } = await entitledChatActor();
     const run = await startChatRun(actor, {
       agentId,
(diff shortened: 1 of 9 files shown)


seven332:
## Code Review: PR #26402

**Reviewed head:** `9dfcee1f5670d835e452f2d2f1f95cbe8cc8ebe7`

### Summary

The PR safely reclaims only unused reserved suffixes while the transaction still owns each thread lock, preserves immutable interior gaps, and makes snapshot metadata honor its existing range-coverage semantics. The accepted batch path keeps the bulk insert and now skips conflict analysis entirely.

### Findings

No findings.

### Testing

API route integration coverage exercises full replay, suffix and interior conflicts, a lost replacement race, sparse snapshot generation, and later tailing. Platform coverage verifies a snapshot cursor beyond the final archive row. The full API workspace (2,898 tests), contracts tests, affected lint/type checks, formatting, and Knip passed; the submitted head additionally passed the focused repeated-batch regression, API lint/type checks, Knip, and commit hooks. CI status is handled separately.

### Verdict

LGTM

PR #26471, opened 2026-08-12
fix(python): preserve malformed base literal scope

### LINKED ISSUE #26444: bug(python): preserve literal scope for malformed base segments
## Summary

`_compile_base_segments_for_match()` replaces every `SegmentError` with an unconstrained whole-segment `SegmentParam` (`crates/runner/mitm-addon/src/firewall_matching/base_url.py:323-326`). When a mixed host or path segment is malformed because it contains multiple parameters, that fallback discards recognizable literal scope such as `api-` or `admin-`.

The compiled matcher intentionally retains malformed bases so requests within their scope fail closed, but its documented contract says malformed config for an unrelated base must not block unrelated traffic (`crates/runner/mitm-addon/src/matching.py:802-805`).

## Reproduction

Compile a firewall with each malformed base and call `matching.match_compiled_firewall_request()`:

| Malformed base | Request URL | Actual | Expected |
| --- | --- | --- | --- |
| `https://api-{a}-{b}.example.com` | `https://evil.example.com/items` | `FirewallBlock(reason="malformed_firewall_config")` | `None` |
| `https://api.example.com/admin-{a}-{b}` | `https://api.example.com/public/items` | `FirewallBlock(reason="malformed_firewall_config")` | `None` |

The corresponding in-scope requests `https://api-us-prod.example.com/items` and `https://api.example.com/admin-us-prod/items` also return `malformed_firewall_config`, which is the intended fail-closed behavior that must be retained.

`_parse_segment()` correctly rejects the multiple-parameter grammar at `crates/runner/mitm-addon/src/firewall_matching/patterns.py:59-69`; the scope expansion happens only when the error is compiled into an empty-prefix, empty-suffix parameter.

## Impact

A malformed or corrupted registry entry intended for an `api-*` host label or `admin-*` path segment can cause unrelated traffic on the same remaining authority/path structure to receive a 403. This turns the defense-in-depth malformed-config handling into an overbroad availability failure.

Existing coverage in `crates/runner/mitm-addon/tests/test_compiled_firewall_malformed_base.py:135-176` verifies that multiple-parameter bases fail closed for an in-scope example, but does not exercise unrelated host/path segments. The nearby greedy-parameter scope tests do preserve literal scope, so the behavior is inconsistent across malformed segment categories.

## Acceptance criteria

- Preserve recognizable literal host and path scope when compiling malformed mixed segments with multiple parameters.
- Requests outside that literal scope do not match the malformed base and are not blocked by it.
- Requests inside the preserved scope still fail closed with `malformed_firewall_config`.
- Add regression coverage for both host-label and path-segment cases, including matched and unrelated URLs.
- Keep the existing malformed-base precedence and fail-closed tests passing.

seven332:
## Summary

- preserve the outer literal prefix and suffix when compiling malformed multi-parameter firewall base segments
- keep in-scope malformed bases fail-closed while excluding unrelated host labels and path segments
- add compiled matcher regressions for host/path prefix and suffix scope

## Scope

- leaves the accepted connector grammar and public segment parser output unchanged
- leaves other malformed segment categories on their existing fallback behavior
- introduces no API, persisted-state, migration, or rollout contract changes

## Verification

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (5,286 passed)

Closes #26444

### DIFF
--- a/crates/runner/mitm-addon/src/firewall_matching/base_url.py
+++ b/crates/runner/mitm-addon/src/firewall_matching/base_url.py
@@ -323,7 +323,14 @@ def _compile_base_segments_for_match(
         parsed_segment = _parse_segment(segment)
         if isinstance(parsed_segment, SegmentError):
             has_malformed_segment = True
-            parsed.append(SegmentParam("", f"__malformed_base_segment_{index}", "", ""))
+            parsed.append(
+                SegmentParam(
+                    parsed_segment.prefix,
+                    f"__malformed_base_segment_{index}",
+                    parsed_segment.suffix,
+                    "",
+                )
+            )
         elif (
             isinstance(parsed_segment, SegmentParam)
             and parsed_segment.greedy
--- a/crates/runner/mitm-addon/src/firewall_matching/patterns.py
+++ b/crates/runner/mitm-addon/src/firewall_matching/patterns.py
@@ -23,6 +23,8 @@ class SegmentParam(NamedTuple):
 
 class SegmentError(NamedTuple):
     reason: str
+    prefix: str = ""
+    suffix: str = ""
 
 
 ParsedSegment = SegmentLiteral | SegmentParam | SegmentError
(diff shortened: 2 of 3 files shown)


seven332:
## Code Review: PR #26471

**Reviewed head:** `efcf09dda09b9757a6bd86b7e6dc8337f6927865`

### Summary

The change keeps malformed multi-parameter bases rejected and fail-closed within their recognizable outer literal scope, while preventing those bases from matching unrelated host labels or path segments. It reuses the existing mixed-segment matcher without changing public parser output or connector grammar.

### Findings

No findings.

### Testing

- Host and path integration regressions cover both prefix and suffix boundaries, including in-scope fail-closed and unrelated no-match outcomes.
- `uv lock --check`, `uv sync --locked`, Ruff format/lint, and BasedPyright passed.
- Full mitm-addon suite passed: 5,286 tests.
- GitHub CI was still in progress at review time.

### Verdict

LGTM

PR #26484, opened 2026-08-12
fix(chat): align initial thinking generation and animation

e7h4n:
## Summary

- opt initial-thinking progress generation out of OpenRouter reasoning while keeping the existing 160-token output budget
- rewrite the prompt around stable UI context: the actual response is separate, one paragraph is shown at a time, a mobile line holds about 20 Chinese characters or 7 English words, and around four short paragraphs give the animation material to rotate
- preserve the safety contract that progress copy must not answer the user or expose hidden reasoning
- treat only explicit paragraph breaks as rotation boundaries; when one line exceeds the measured width, end it with `…`, discard the unseen remainder, and advance to the next paragraph after the existing hold/fade interval
- preserve strict completion semantics: `length` / `MAX_TOKENS` provider output remains rejected even when partial content is present

## Production evidence

For `vm0-web-logs-prod` in the exact window `2026-08-11T03:16:38Z` through `2026-08-12T03:16:38Z`:

- 104 `Initial thinking generation failed` warnings
- 103 were `OpenRouter completion finished with length (native: MAX_TOKENS)` across 103 runs and 51 threads
- 1 was an unrelated OpenRouter 503
- 2,369 `thinking:initial` events were persisted in the same window; `103 / (2,369 + 103) = 4.17%` is the observable capped-output ratio and an upper bound on the true attempt failure rate because a successful generation can still be discarded when the run advances before persistence

The service sent `max_tokens: 160` while allowing a few short paragraphs and applying the 600-character cap only after completion. OpenRouter counts reasoning in the output budget, and the selected Gemini model advertises reasoning enabled by default. OpenRouter normalizes the native `MAX_TOKENS` outcome to `finish_reason: length`; the shared wrapper intentionally throws on every non-`stop` finish before reading `message.content`, so partial progress copy was correctly discarded but logged.

## Design

The generation prompt now supplies presentation facts instead of encoding the frontend's wrapping algorithm as model instructions. Its target is around four mobile-width paragraphs: roughly 80 Chinese characters or 28 English words in total when the model follows the stated width. That fits the existing budget after reasoning is disabled, so this PR does not rely on a token-budget increase.

The frontend no longer manufactures additional lines from an overlong paragraph. It types only the visible prefix, replaces the tail with an ellipsis when the measured width is exceeded, keeps the existing 700 ms hold and 200 ms fade, then starts the next explicit paragraph. The skipped suffix is never animated later.

## Verification

- `pnpm format`
- `pnpm knip`
- `pnpm --filter @vm0/app exec vitest run src/views/zero-page/__tests__/chat-thinking-indicator.test.tsx` (2 passed)
- `pnpm --filter @vm0/app exec vitest run src/views/zero-page/__tests__/chat-billing-and-errors.test.tsx -t 'initial thinking indicator'` (6 passed, 13 skipped)
- `DATABASE_URL=postgresql://user@localhost:5432/vm0_test pnpm --filter api exec vitest run src/signals/routes/__tests__/chat-events.bdd.test.ts -t 'initial thinking indicator'` (2 passed, 85 skipped)

Per repository instructions, no full Vitest suite or local dev server was run. The page-level tests exercise deterministic mobile-width overflow and paragraph rotation; broad and preview checks are left to the PR pipeline.

## Remaining uncertainty

OpenRouter currently reports reasoning as non-mandatory for `google/gemini-3.1-flash-lite-preview`, so `effort: none` is the gateway-supported opt-out. Gemini 3 providers may internally map that opt-out to their minimum thinking level rather than literal zero reasoning. Model adherence to the approximate paragraph width also cannot be guaranteed; the frontend now handles violations deterministically, and the production warning rate should still be checked after deployment.

### DIFF
--- a/turbo/apps/api/src/signals/external/openrouter.ts
+++ b/turbo/apps/api/src/signals/external/openrouter.ts
@@ -56,6 +56,7 @@ interface OpenRouterResponse {
 }
 
 interface OpenRouterGenerateTextOptions {
+  readonly reasoning?: { readonly effort: "none" };
   readonly responseFormat?: { readonly type: "json_object" };
   readonly temperature?: number;
 }
@@ -228,6 +229,9 @@ export async function generateTextWithUsage(
       ...(options?.responseFormat === undefined
         ? {}
         : { response_format: options.responseFormat }),
+      ...(options?.reasoning === undefined
+        ? {}
+        : { reasoning: options.reasoning }),
       temperature: options?.temperature ?? 0.3,
     }),
     signal,
--- a/turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts
@@ -351,7 +351,10 @@ function userMessageWithTemplate(
 }
 
 const openRouterBodySchema = z.object({
+  model: z.string(),
   messages: z.array(z.object({ role: z.string(), content: z.string() })),
+  max_tokens: z.number().optional(),
+  reasoning: z.object({ effort: z.literal("none") }).optional(),
 });
 
 async function entitledChatActor(
@@ -6460,9 +6463,10 @@ describe("CHAT-02: initial thinking indicator", () => {
 
     let thinkingAuthorization: string | null = null;
     let thinkingPromptPayload = "";
+    let thinkingRequestBody: z.infer<typeof openRouterBodySchema> | undefined;
     const titleResponse = "Launch Checklist";
     const thinkingResponse =
-      "Reviewing the launch request and recent context.\n\nOrganizing the checklist into practical sections before the main response starts.";
+      "Reviewing the launch request and recent context.\nIdentifying the checklist's major sections.\nChecking where owners and timing matter.\nPreparing a clear order for the response.";
     server.use(
       http.post(
         "https://openrouter.ai/api/v1/chat/completions",
@@ -6475,6 +6479,7 @@ describe("CHAT-02: initial thinking indicator", () => {
           }
           if (systemContent.includes("Write user-visible progress copy")) {
             thinkingAuthorization = request.headers.get("authorization");
+            thinkingRequestBody = payload;
             thinkingPromptPayload = payload.messages
               .map((message) => {
                 return message.content;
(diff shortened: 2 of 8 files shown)


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 8c5e1eb007e26c0d812f53e3276aeada7c957e35
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

The PR removes hidden reasoning from the bounded initial-progress generation request, redesigns the prompt around the UI's real paragraph-width context, and changes the thinking indicator to ellipsize and discard an overflowing line before advancing to the next explicit paragraph. The provider still rejects `length` / native `MAX_TOKENS` completions, so partial copy cannot be mistaken for a complete safe progress message.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness, edge cases, and API semantics | Pass | `openrouter.ts` preserves strict non-`stop` rejection; `create-chat-thread.ts` handles explicit lines, CR/LF boundaries, measured overflow, the ellipsis-only edge, and final-line completion. |
| `REVIEW.md` | Security, privacy, and type safety | Pass | The prompt retains the prohibitions on answering the user and exposing hidden reasoning; the new OpenRouter option is narrowly typed as `{ effort: "none" }`, with no `any`, suppression, secret, or unsafe cast added. |
| `REVIEW.md` | A `fix:` change has regression coverage | Pass | The route test covers the exact provider request and refusal to persist token-limited partial output; page tests cover ellipsis, suffix discard, and explicit-line advancement. |
| `docs/bad-smell.md` | Fail fast; avoid `any`, suppressions, dynamic imports, and hidden error recovery | Pass | Native `MAX_TOKENS` continues to fail visibly rather than accepting partial text, and the diff introduces none of the prohibited TypeScript or import patterns. |
| `docs/fallback.md` | Do not introduce undeclared or open-ended fallback paths | Pass | No cross-version or contract fallback is introduced. The retained 14 px text-measurement default is bounded presentational behavior, while provider truncation remains fail-fast. |
| `docs/event-sourcing.md` | Persistent event authority and optimistic reconciliation | N/A | The PR does not change event IDs, persistence authority, reconciliation, rollback, or optimistic event cleanup; it only changes generation parameters and presentation of the existing `output.thinking` event. |
| `docs/effect.md`, `docs/cache.md`, `.claude/skills/ccstate/SKILL.md` | Correct owner, lifecycle, and cleanup for DOM-bound async work | Pass | The typewriter loop remains inside `onRef`, derives measurement from the mounted label, inherits the ref lifecycle `AbortSignal`, and adds no render-time effects, manual cache, singleton, or detached signal work. |
| `docs/deployment-compatibility.md` | Mixed frontend/backend versions remain compatible | Pass | The persisted/API shape remains the same plain `thinking` string. An old frontend can display the new multi-line copy, and the new frontend deterministically displays old single- or multi-line copy without requiring a protocol change. |
| `docs/testing.md`, `docs/testing/patterns.md`, `docs/testing/anti-patterns.md` | Prefer integration tests, boundary mocks, real timers, and observable outcomes | Pass | Tests use the real route/page paths, MSW/browser boundaries, and real interval behavior; no internal relative mock, direct fetch mock, DB mock, fake timer, or mock-call-only assertion was added. |
| `docs/testing/api-testing.md`, `docs/testing/testing-external-behavior.md` | API behavior is exercised through the route and external service boundary | Pass | `chat-events.bdd.test.ts` sends through the production chat API, observes persisted events through the API, and intercepts only OpenRouter through MSW. |
| `docs/testing/app-testing.md`, `docs/testing/testing-external-behavior.md` | Platform behavior is exercised through production page bootstrap and visible UI | Pass | Both app test files use `detachedSetupPage`, lifecycle-scoped mocks, DOM-visible text frames, and the mounted label rather than rendering an internal component or mutating ccstate directly. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- Coverage: Adequate. Focused local API and App regression tests passed, and all eight API plus all eight App CI shards passed on this head. Codecov reports 95.74% patch coverage; the two uncovered defensive lines do not leave the fixed behavior untested.
- Conventions: Compliant. Formatting, lint, Knip, API/App type checks, security scans, and focused integration tests passed. The remaining CI failures are E2E account setup failures where Clerk returned HTTP 403 before PR-related behavior ran.

PR #26492, opened 2026-08-12
fix(pi): bound tool execution deadlines

### LINKED ISSUE #26464: bug: prevent Pi tools from hanging until the job timeout
## Background

On August 11, 2026 (Beijing time), six production Codex runs using the Pi runtime failed with the same terminal error:

```text
message: job execution failed
error: execution: Pi standby execution timed out
failure_class: cli_execution_error
failure_framework: codex
```

Each run failed after approximately 7,200,000 ms. Startup, VM readiness, heartbeat, and first output were healthy. The runs emitted between 2 and 51 output messages before making no further progress until the outer two-hour job deadline.

The exact 7,200-second duration establishes that the outer execution deadline ended these runs. It does not by itself prove which operation stalled. However, the current Pi tool execution path demonstrably allows an individual tool call to remain unresolved until that outer deadline, so it must be bounded independently.

## Problem

Pi model requests have a 120-second timeout, but Pi tool calls do not have a runtime-owned default timeout:

- `createBashTool()` exposes an optional agent-provided `timeout`, but omitted values mean no Bash timeout.
- The recovered first batch uses vm0's `executePiToolBatch`, while later batches are executed directly by Pi core's `runAgentLoopContinue`. Both paths await the same vm0 tool wrappers without a runtime deadline.
- Both parallel executors join calls with `Promise.all`, so one unresolved call keeps the whole batch open.
- Both executors convert thrown cancellation into an ordinary text-only tool result.
- The production `zero __agent-loop` entry point passes a newly created signal that is not connected to SIGTERM or guest-agent cancellation.

Consequently, an omitted Bash timeout or another unresolved tool operation in any batch can hold the Pi loop until the two-hour job deadline. The final `Pi standby execution timed out` error does not classify a tool timeout or let Pi recover from one.

## Required behavior

Every Pi tool invocation must have an effective timeout:

```text
effectiveTimeout = min(agentTimeout ?? defaultTimeout, maxTimeout)
```

- `defaultTimeout` is 10 minutes.
- `maxTimeout` is 30 minutes.
- Bash keeps its existing optional `timeout` argument. An explicit finite positive value may shorten or extend the default, subject to the maximum.
- Read, write, and edit do not gain new timeout arguments; they use the runtime default.
- Parent-job cancellation always takes precedence over the per-tool deadline.
- A per-tool timeout is a recoverable tool error returned to Pi, so the model can retry or choose another approach.
- Parent cancellation terminates the run instead of being converted into a recoverable timeout result.

The policy is defined once by the Pi runtime. It does not require a new API, runner protocol field, database field, environment variable, or per-tool configuration system.

## Implementation direction

- Apply the effective deadline inside the vm0 tool-wrapper layer shared by the recovered first batch and every later Pi-core batch.
- Give each invocation its own deadline controller, compose it with the parent signal, and race native settlement against abort so an implementation that does not cooperate promptly cannot hold the batch open.
- Preserve Bash's existing explicit timeout input, calculate its effective value against the default and maximum, and update its model-visible documentation.
- Return `details: { code: "tool_timeout", timeoutMs }` when a tool deadline wins. Mark that result as an error in vm0's batch executor and Pi core's `afterToolCall` hook without losing the structured details.
- Use the existing ToolResult envelope for `toolName` and `toolCallId`; do not record arguments or add a separate telemetry pipeline.
- Let parent cancellation reject, make resume event delivery abort-aware, and connect SIGTERM at `zero __agent-loop` to the controller passed into `runPiStandbyAgentLoop`.
- Do not abort the shared parent signal, call global `ExecutionEnv.cleanup()`, or cancel unrelated parallel tools when one invocation reaches its deadline.

## Acceptance criteria

- A Pi tool call without an explicit timeout uses the 10-minute runtime default in both recovered and later batches.
- A Bash call may explicitly request a timeout shorter or longer than the default; the effective value never exceeds 30 minutes.
- Read, write, and edit use the runtime default without adding model-visible timeout parameters.
- A tool timeout aborts the underlying operation, terminates Bash subprocesses, and cannot leave a parallel batch unresolved.
- A tool timeout produces a recoverable error ToolResult with `code: "tool_timeout"` and the effective `timeoutMs`.
- The timeout result identifies the tool name and tool-call ID through its existing envelope without recording tool arguments or other sensitive content.
- Parent cancellation reaches the underlying operation, terminates Bash subprocesses, and terminates the Pi run rather than returning a recoverable timeout result.
- Parallel calls preserve result ordering, and one timed-out call does not cancel an otherwise successful sibling.
- Tests cover successful completion, the default timeout, explicit Bash timeouts below and above the default, the runtime maximum, parent cancellation, subprocess cleanup, a parallel batch containing one timed-out call, and timeout classification in a later Pi-core batch.

## Relevant code

- `turbo/packages/pi-agent-runtime/src/tools.ts` is the shared wrapper layer used by recovered and later batches.
- `turbo/packages/pi-agent-runtime/src/tool-batch.ts` owns the recovered first-batch executor.
- `turbo/packages/pi-agent-runtime/src/agent-loop.ts` delegates later batches to `runAgentLoopContinue`.
- `turbo/apps/cli/src/commands/zero/__agent-loop.ts` currently passes an AbortSignal that is not connected to process termination.

---
> Compacted on 2026-08-12 from 0 comments.

seven332:
## Summary

- bound every Pi read, Bash, write, and edit invocation with a 10-minute runtime default and a 30-minute maximum
- preserve Bash's optional model-selected timeout while returning stable, recoverable `tool_timeout` details from both recovered and later Pi batches
- propagate parent cancellation through the Pi runtime and bridge guest-agent SIGTERM into the CLI loop's abort signal
- verify timeout capping, parallel isolation and ordering, parent cancellation, later-batch recovery, and real Bash subprocess-tree cleanup

## Exclusions

- no API, database, queue, runner protocol, or persisted message-envelope changes
- no timeout parameters for read, write, or edit
- no environment-variable override or separate telemetry pipeline

## Validation

- `cd turbo && pnpm format`
- `cd turbo && pnpm -F @vm0/pi-agent-runtime lint`
- `cd turbo && pnpm -F @vm0/pi-agent-runtime check-types`
- `cd turbo && pnpm -F @vm0/pi-agent-runtime test`
- `cd turbo && pnpm -F @vm0/okou-cli lint`
- `cd turbo && pnpm -F @vm0/okou-cli check-types`
- `cd turbo && pnpm -F @vm0/okou-cli test`
- `cd turbo && pnpm knip --workspace @vm0/pi-agent-runtime`
- `cd turbo && pnpm knip --workspace @vm0/okou-cli`
- pre-commit hook: repository-wide Knip and TypeScript checks

Closes #26464

### DIFF
--- a/turbo/apps/cli/src/commands/zero/__agent-loop.ts
+++ b/turbo/apps/cli/src/commands/zero/__agent-loop.ts
@@ -29,6 +29,13 @@ export const zeroAgentLoopCommand = new Command()
   .action(async (options: { standby: true; standbyTtlSeconds: number }) => {
     const io = new StdioPiAgentLoopIo(process.stdin, process.stdout);
     const executionEnv = createPiNodeExecutionEnv();
+    const abortController = new AbortController();
+    const abortOnSigterm = () => {
+      const error = new Error("Pi agent loop received SIGTERM");
+      error.name = "AbortError";
+      abortController.abort(error);
+    };
+    process.once("SIGTERM", abortOnSigterm);
     try {
       await runPiStandbyAgentLoop(
         {
@@ -37,7 +44,7 @@ export const zeroAgentLoopCommand = new Command()
           executionEnv,
           standbyTtlSeconds: options.standbyTtlSeconds,
         },
-        new AbortController().signal,
+        abortController.signal,
       );
     } catch (error) {
       await io.write({
@@ -46,6 +53,7 @@ export const zeroAgentLoopCommand = new Command()
       });
       process.exitCode = 1;
     } finally {
+      process.removeListener("SIGTERM", abortOnSigterm);
       io.close();
       await executionEnv.cleanup();
     }
--- a/turbo/apps/cli/src/commands/zero/__agent-loop.ts
+++ b/turbo/apps/cli/src/commands/zero/__agent-loop.ts
@@ -34,6 +34,7 @@ export const zeroAgentLoopCommand = new Command()
       const error = new Error("Pi agent loop received SIGTERM");
       error.name = "AbortError";
       abortController.abort(error);
+      io.close();
     };
     process.once("SIGTERM", abortOnSigterm);
     try {
--- a/turbo/packages/pi-agent-runtime/src/agent-loop.test.ts
+++ b/turbo/packages/pi-agent-runtime/src/agent-loop.test.ts
@@ -1,4 +1,5 @@
 import { mkdtemp, rm, writeFile } from "node:fs/promises";
+import { getEventListeners } from "node:events";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 
@@ -495,8 +496,10 @@ describe("Pi DeepSeek provider", () => {
     await writeFile(initialPath, "initial tool result\n");
     const env = new LaterBashExecutionEnv({ cwd: root });
     const timeoutSpy = vi.spyOn(globalThis, "setTimeout");
+    const controller = new AbortController();
     const requestBodies: unknown[] = [];
     const events: AgentEvent[] = [];
+    let firstEventAbortListener: unknown = null;
     server.use(
       http.post(DEEPSEEK_RESPONSES_URL, async ({ request }) => {
         const body: unknown = await request.json();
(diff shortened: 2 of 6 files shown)


seven332:
## Code Review: PR #26492

**Reviewed head:** `494ac4c372ffdb2f40717ce4c316887cbb64637f`

### Summary

The shared Pi tool wrapper now owns bounded deadlines, stable recoverable timeout classification, parent-cancellation precedence, and Bash subprocess termination across both recovered and later Pi-core batches. The CLI signal bridge and event sink preserve fatal cancellation without changing API, database, queue, protocol, or persisted message contracts.

### Findings

No findings.

### Testing

Reviewed integration coverage for the default, shorter and longer Bash values, the maximum cap, parent cancellation, parallel ordering and isolation, later-batch recovery, pending persistence cancellation, and real subprocess-tree cleanup. Runtime and CLI formatting, lint, type checks, tests, scoped Knip, and repository-wide pre-commit Knip and TypeScript checks passed locally. The complete repository CI matrix remains the final unverified surface.

### Verdict

LGTM

PR #26493, opened 2026-08-12
refactor(platform): join mail part previews onto draft lists

e7h4n:
## Summary

Registry audit follow-up for the chat event pipeline: the mail draft sidebar was the last surface resolving per-part resources through exposed maps at render time.

`attachmentPreviews$` used to resolve `{ text: Map<partId, TextPreviewComputed>, urls: Map<partId, string | null> }`, and the sidebar looked parts up by id in three places while rendering (attachment tile `text$`, attachment tile `url`, and inline `cid:` images inside the HTML walk).

The computed now performs that join itself and resolves draft-shaped lists:

- `attachments: readonly { attachment, url, text$ }[]` — the tiles walk this directly; while previews load, the draft's own attachment list stands in with loading tiles.
- `inlineImages: readonly { image, url }[]` — the HTML walker keys its own input by normalized content id (a walk-local join of forward-flowing data, same shape as the transcript projection joins), replacing the two store maps it used to receive.

No behavior change intended: loading (`undefined`) vs missing (`null`) url semantics and the object-URL revocation lifecycle are preserved.

### DIFF
--- a/turbo/apps/platform/src/signals/chat-page/mail-draft.ts
+++ b/turbo/apps/platform/src/signals/chat-page/mail-draft.ts
@@ -8,7 +8,9 @@ import {
 } from "ccstate";
 import {
   zeroMailContract,
+  type ZeroMailAttachment,
   type ZeroMailDraft,
+  type ZeroMailInlineImage,
 } from "@vm0/api-contracts/contracts/zero-mail";
 
 import { accept } from "../../lib/accept.ts";
@@ -34,9 +36,27 @@ export interface MailDraftDescriptor {
   readonly href: string;
 }
 
+interface MailAttachmentPartPreview {
+  readonly attachment: ZeroMailAttachment;
+  /** Object URL of the fetched part; null when the part no longer exists. */
+  readonly url: string | null;
+  readonly text$: TextPreviewComputed | undefined;
+}
+
+export interface MailInlineImagePreview {
+  readonly image: ZeroMailInlineImage;
+  /** Object URL of the fetched part; null when the part no longer exists. */
+  readonly url: string | null;
+}
+
+/**
+ * Fetched part previews joined onto the draft's own attachment and inline
+ * image lists. The part lookup happens here, so the sidebar walks these lists
+ * directly instead of resolving parts by id while rendering.
+ */
 export interface MailAttachmentPreviews {
-  readonly text: ReadonlyMap<string, TextPreviewComputed>;
-  readonly urls: ReadonlyMap<string, string | null>;
+  readonly attachments: readonly MailAttachmentPartPreview[];
+  readonly inlineImages: readonly MailInlineImagePreview[];
 }
 
 export interface MailDraftSignals extends MailDraftDescriptor {
@@ -155,7 +175,7 @@ function createAttachmentPreviews(
   const attachmentPreviews$ = computed(
     async (get): Promise<MailAttachmentPreviews> => {
       if (!get(attachmentScopeActive$)) {
-        return { text: new Map(), urls: new Map() };
+        return { attachments: [], inlineImages: [] };
       }
       const currentLoadVersion = ++loadVersion;
       const draftPromise = get(sidebarDraft$);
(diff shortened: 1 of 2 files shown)


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 5be6f8cf69f331656777aa9c87080ba685a7c079
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

This refactor moves per-part attachment and inline-image lookups out of React rendering and into `attachmentPreviews$`, which now exposes draft-shaped preview lists. The complete diff preserves attachment order, loading (`undefined`) versus missing (`null`) semantics, text-preview construction, and object-URL teardown without changing the mail UI contract.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and edge cases | Pass | `mail-draft.ts:createAttachmentPreviews` deduplicates fetches by part ID, then joins each result back onto the original attachment and inline-image lists; `mail-draft-sidebar.tsx` consumes those lists without changing ordering or missing/loading states. |
| `REVIEW.md` | Security | Pass | The change only rewires typed preview data; the existing HTML allowlist, URL protocol checks, remote-image policy, and React element rendering remain unchanged. |
| `REVIEW.md` | Type safety and maintainability | Pass | The exposed maps are replaced with explicit preview interfaces and readonly lists; no broad casts, `any`, suppressions, or duplicated resource ownership were added. |
| `REVIEW.md` | Refactor test coverage | Pass | Existing platform integration coverage exercises PDF and text attachment previews, `cid:` inline images, send refresh, and object-URL revocation through the production page setup. |
| `REVIEW.md` | Persisted and deployment boundaries | N/A | No API contract, database/JSONB shape, request/response, event, or independently deployed protocol changes. |
| `REVIEW.md` | Feature regression containment | N/A | This is a behavior-preserving internal refactor and introduces no feature or user-flow gate. |
| `docs/bad-smell.md` | Avoid unsafe types, suppressions, dynamic imports, and hidden recovery | Pass | Both changed files remain statically typed and add none of the prohibited patterns. |
| `docs/testing.md` | Prefer behavior-level integration coverage using production initialization and boundary mocks | Pass | `chat-event-action-cards.test.tsx` uses `detachedSetupPage`, boundary HTTP/API mocks, and user-visible assertions for the exact preview paths being refactored. |
| `docs/fallback.md` | Do not introduce undeclared or unreachable fallback paths | Pass | The moved `?? null` lookups preserve the existing missing-part meaning, while the draft-list stand-ins are presentational loading defaults; no new compatibility or rollout fallback is introduced. |
| `docs/effect.md` | Derive async data in `computed` and tie acquired resources to their real lifecycle | Pass | Fetch/join work remains in an async computed, and the existing stable `onRef` scope continues to own activation and teardown rather than React render. |
| `docs/cache.md` | Keep render pure and release external resources through the owning scope | Pass | React creates only a walk-local data map; object URLs stay in the thread-scoped signal factory and are revoked on replacement, sidebar detach, and page abort. |
| `.claude/skills/ccstate/SKILL.md` | Use reactive async computed, stable `onRef`, and factory-scoped signals | Pass | The refactor preserves `computed(async ...)`, `accept`, the page-owned abort signal, direct `useSet(setAttachmentScopeRef$)` ref usage, and factory isolation. |
| `docs/chat-cards.md` | Register thread-scoped typed signals before rendering and pass them directly to React | Pass | Mail draft signals remain owned by the thread registry; the view receives typed preview lists and creates no signals or registry lookups during render. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- Coverage: Adequate. The existing `opens a shared mail draft without reloading and refreshes after sending` integration test covers attachment URLs, text previews, inline `cid:` images, and URL revocation.
- Conventions: Compliant. No tests changed; the existing coverage follows production initialization and boundary-mocking rules.
- Execution: Local Vitest was not run per repository workflow. GitHub checks for the reviewed SHA are green, including all `test-app` shards, app type checks, ESLint, and format checks.

Issue #26303, opened 2026-08-11
bug: OpenAI model usage settlement does not reconcile with provider usage


seven332:
## Problem

Daily reconciliation shows that vm0 records and settles more OpenAI model usage than OpenAI Admin reports for the same production API key and UTC window. Anthropic usage reconciles under the equivalent scope.

The excess OpenAI usage is confirmed, but its source is not. The previous issue body treated broad Codex response classification as the proven cause. That is a real billing-safety weakness, but the retained evidence does not yet show that it produced the reconciled excess.

### Confirmed

- The runner currently classifies every observable Codex response outside an exact `/chat/completions` path as OpenAI Responses traffic.
- Once a response is classified as a model protocol, any positive normalized model usage category can enter observation and, for a billable model-provider firewall, usage settlement. A positive output-token count is not required.
- The OpenAI Responses JSON extractor selects exact top-level fields such as `usage.input_tokens` and `usage.output_tokens`.
- OpenAI Admin completion-usage responses place token totals under `data[].results[]`, not under top-level `usage`. The current extractor therefore does not match the documented Admin response shape. See the [OpenAI Usage API reference](https://platform.openai.com/docs/api-reference/usage/audio_transcriptions_object).
- A read-only review of seven days of retained network logs on `prod-3` and `prod-4` found normal `/responses` and `/chat/completions` inference traffic, model-catalog requests, Admin usage/cost queries, and ChatGPT image requests. It did not find `/responses/{id}`, `/responses/compact`, `/responses/input_tokens`, or Batch resource requests that could explain the settled API-key excess in that log scope.
- HTTP request timestamps cannot be directly attributed to nearby model-usage flushes for Codex WebSocket runs: terminal usage arrives on an already-open `/responses` WebSocket and does not create a new HTTP network row.

### Not yet confirmed

- The exact request, transport event, or usage source that created the excess settled row.
- Whether endpoint over-classification caused the observed reconciliation difference.
- The exact historical rows, users, or credit settlements that require correction.

## Decision

Revise the framing: the primary bug is an unresolved OpenAI usage reconciliation failure. Broad model-response classification remains an independently valid correctness risk and a strong hardening direction, but shipping that hardening alone is not evidence that the reconciliation bug is fixed.

Do not close this issue until the excess usage is attributed or another cause is proven and the same-key, same-window reconciliation succeeds.

## Chosen direction

[...]

PR #26496, opened 2026-08-12
fix(runner): exclude codex websocket prewarm usage

### LINKED ISSUE #26475: bug: exclude Codex generate=false WebSocket prewarm from model usage
## Background

Parent: #26303

PR #26403 added source-level model usage diagnostics so the OpenAI reconciliation excess could be attributed without changing accounting behavior. Production data for the 2026-08-11 UTC window now identifies a concrete Responses WebSocket source that vm0 records but OpenAI Admin does not reconcile as ordinary inference input.

The current runner bundles Codex CLI `0.147.0`. That client sends a Responses WebSocket `response.create` request with `generate: false` before the first real turn. Codex describes this operation as connection setup rather than an inference request, waits for its completion, and reuses its response ID as the real turn's `previous_response_id`:

- https://github.com/openai/codex/blob/rust-v0.147.0/codex-rs/core/src/client.rs#L15-L19
- https://github.com/openai/codex/blob/rust-v0.147.0/codex-rs/core/src/client.rs#L1651-L1659
- https://github.com/openai/codex/blob/rust-v0.147.0/codex-rs/core/tests/suite/agent_websocket.rs#L178-L213

## Problem

The mitmproxy addon observes client-originated OpenAI Responses WebSocket frames only to extract their top-level event type for Codex output timing. It does not retain the `generate: false` operation semantic. It then extracts and reports every positive usage quantity from server-originated terminal response events.

The prewarm terminal response contains positive `input_tokens` but no cache or output quantity. The current input partition therefore records the entire amount as ordinary `tokens.input`, producing both billable usage events and model usage observations for an operation that is not a real generation.

Relevant code:

- `crates/runner/mitm-addon/src/mitm_addon.py`
- `crates/runner/mitm-addon/src/response_streaming.py`
- `crates/runner/mitm-addon/src/usage/openai_responses.py`
- `crates/runner/mitm-addon/src/usage/openai_tokens.py`

## Production evidence

For 2026-08-11 UTC:

- OpenAI Admin ordinary input: `839,091`
- MaskDB ordinary input: `3,608,849`
- excess: `2,769,758` (`+330.090%`)
- total-token difference across all categories: `+1,259,941` (`+0.092%`)

Source diagnostics retained on `prod-3` and `prod-4` from approximately 13:46 UTC onward contain 82 OpenAI API-key WebSocket responses with input usage only, across 77 runs:

- 77 responses reported exactly `6,050` input tokens;
- 5 responses reported exactly `6,999` input tokens;
- combined ordinary input recorded for these sources: `500,845`;
- each source had a distinct provider response ID and was accepted once by the source buffer;
- observed webhook deliveries succeeded;
- normal generated responses followed with output and cache creation/read usage.

The dominant `6,050` signature also closely matches the full-day excess: `458 * 6,050 = 2,770,900`, only `1,142` above the observed input delta. The retained diagnostics do not cover the full UTC day, so this arithmetic is supporting evidence rather than an exact historical source count.

There is no evidence that duplicate provider response IDs, source-buffer retries, or duplicate webhook delivery caused this excess.

## Required behavior

Recognize the exact Responses WebSocket client operation semantic and exclude `response.create` requests with `generate: false` from model usage settlement and reconciliation observations.

Correlate the client request with its server response ID before handling the terminal usage event. Preserve a bounded diagnostic signal for ignored prewarm usage so post-deployment reconciliation can verify the decision.

Do not identify prewarm using any of these heuristics:

- zero output tokens;
- a fixed input quantity such as `6,050` or `6,999`;
- the first response on a connection or run;
- the `/v1/responses` path alone;
- Codex-specific metadata when the protocol-level `generate: false` field is available.

Those heuristics could suppress legitimate failed, incomplete, cancelled, or otherwise input-only inference usage.

## Acceptance criteria

- A client `response.create` frame with `generate: false` is correlated with its server response and its terminal usage creates neither billable usage events nor model usage observations.
- A normal `response.create` request, including one that finishes with positive input usage and zero output usage, retains the current reporting behavior.
- A prewarm followed by a generated turn on the same WebSocket reports only the generated turn's usage.
- Multiple prewarms caused by separate connections or reconnects are classified independently without leaking state across flows or runs.
- Unknown, malformed, or unsupported client frames do not broadly suppress later legitimate usage.
- Existing OpenAI Responses HTTP/SSE, Chat Completions, Anthropic Messages, and non-prewarm Responses WebSocket accounting remain unchanged.
- Integration tests exercise the public mitmproxy WebSocket lifecycle and verify usage-event, observation, idempotency, and diagnostic behavior.
- A post-rollout, same-key, same-UTC-window reconciliation no longer contains the repeated prewarm ordinary-input excess, after old runners have drained and provider reporting latency is accounted for.
- No database schema or persisted-data migration is introduced for the forward fix.

## Scope boundary

This issue covers the forward runtime accounting fix and verification telemetry. It does not perform a blanket historical refund or delete input-only events: historical rows do not persist `generate: false`, and input-only inference is legitimate. Any historical correction requires a separately proven, deterministic, and idempotent affected cohort.

The smaller cache creation/read and output reconciliation differences are also outside this issue unless implementation evidence proves they share the same cause.

## Risks

- **Underbilling legitimate input-only inference:** classify only the explicit `generate: false` operation and test a normal input-only terminal response.
- **Incorrect client/server correlation:** keep state scoped to the WebSocket flow and use the protocol's ordered response lifecycle rather than timing or quantity heuristics.
- **Cross-version verification noise:** reconcile only after runners with the old behavior have drained.

seven332:
## Summary

- inspect Codex Responses WebSocket client frames for the exact `response.create` / `generate=false` operation without materializing prompt or tool content
- bind that operation to its provider response ID at `response.created` and exclude only matching terminal usage from billable events and model usage observations
- retain flow-scoped, bounded correlation state and emit one content-free ignored-source diagnostic for rollout reconciliation
- preserve current reporting for normal input-only inference and for malformed, unsupported, or unbound lifecycles

## Implementation

- extend the bounded selective JSON parser with boolean scalar selection
- keep prewarm correlation private to the observable OpenAI Responses WebSocket flow lifecycle
- retain the ignored response ID through flow cleanup so duplicate terminal delivery stays unbilled while the diagnostic is emitted once
- require a complete, unambiguous terminal response identity before excluding its usage
- cover the public mitmproxy hook path, including large client frames, normal input-only usage, fail-open behavior, duplicates, flow isolation, and terminal cleanup

## Scope

- no database schema or persisted-data migration
- no historical usage correction
- no change to OpenAI Responses HTTP/SSE, Chat Completions, Anthropic Messages, or connector accounting
- production reconciliation should be evaluated after old runners drain and provider reporting latency elapses

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (5,294 passed)
- `lefthook run pre-commit`

Closes #26475

Parent context: #26303

### DIFF
--- a/crates/runner/mitm-addon/src/mitm_addon.py
+++ b/crates/runner/mitm-addon/src/mitm_addon.py
@@ -1576,8 +1576,9 @@ def websocket_message(flow: http.HTTPFlow) -> None:
     if getattr(message, "from_client", False):
         if uses_openai_responses:
             body = message.content.encode() if isinstance(message.content, str) else message.content
-            event_type = usage.inspect_openai_responses_event_type_json(body)
-            codex_output_timing.observe_client_event(flow, event_type, message.timestamp)
+            event = usage.inspect_openai_responses_client_event_json(body)
+            codex_output_timing.observe_client_event(flow, event.event_type, message.timestamp)
+            response_streaming.observe_model_websocket_client_event(flow, event)
         return
     body = message.content.encode() if isinstance(message.content, str) else message.content
     event = usage.inspect_openai_responses_event_json(body)
--- a/crates/runner/mitm-addon/src/response_streaming.py
+++ b/crates/runner/mitm-addon/src/response_streaming.py
@@ -649,18 +649,21 @@ def feed_model_websocket_usage(
             isinstance(prewarm_state, _OpenAIResponsesPrewarmState)
             and prewarm_state.response_id == message_id
         ):
-            if not prewarm_state.diagnostic_emitted and usage.has_positive_model_provider_usage(
-                usage_result
-            ):
-                usage.log_ignored_model_provider_usage_source(
-                    flow,
-                    flow_metadata.run_id(flow.metadata),
-                    message_id,
-                    usage_result,
-                    reason="responses_generate_false",
-                )
-                prewarm_state.diagnostic_emitted = True
-            return
+            lifecycle = usage.inspect_openai_responses_server_lifecycle(event)
+            if lifecycle.is_terminal and lifecycle.response_id == message_id:
+                if not prewarm_state.diagnostic_emitted and usage.has_positive_model_provider_usage(
+                    usage_result
+                ):
+                    usage.log_ignored_model_provider_usage_source(
+                        flow,
+                        flow_metadata.run_id(flow.metadata),
+                        message_id,
+                        usage_result,
+                        reason="responses_generate_false",
+                    )
+                    prewarm_state.diagnostic_emitted = True
+                return
+            prewarm_state.response_id = None
         usage_sources = flow.metadata.get(metadata_keys.MODEL_PROVIDER_USAGE_SOURCES)
         if not isinstance(usage_sources, dict):
             usage_sources = {}
--- a/crates/runner/mitm-addon/src/response_streaming.py
+++ b/crates/runner/mitm-addon/src/response_streaming.py
@@ -4,8 +4,9 @@ Lifecycle:
 - ``mitm_addon.responseheaders()`` calls ``configure_response_stream()`` to
   install the streaming callback, exact byte accounting, optional capped body
   buffer, and incremental usage parsers.
-- ``mitm_addon.websocket_message()`` calls ``feed_model_websocket_usage()`` for
-  terminal server-side usage frames on model-provider WebSocket upgrades.
+- ``mitm_addon.websocket_message()`` calls ``observe_model_websocket_client_event()``
+  for client request intent and ``feed_model_websocket_usage()`` for server-side
+  usage frames on model-provider WebSocket upgrades.
 - ``mitm_addon.response()`` finalizes HTTP model and connector usage before
   reporting it.
 - ``mitm_addon.error()`` may finalize partial SSE or opted-in connector usage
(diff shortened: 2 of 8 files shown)


seven332:
## Code Review: PR #26496

**Reviewed head:** `3310dc78bc117025b60f7a5006cb389911ed7713`

### Summary

The flow-local correlation matches Codex v0.147.0's `response.create` with `generate: false`, `response.created`, and terminal response lifecycle. Only an exact, unambiguous, same-flow response ID is excluded; malformed, conflicting, missing, or unbound data fails open to normal reporting. State and parsing remain bounded, and diagnostics do not retain request content.

### Findings

No findings. The ambiguous duplicate-scalar path found during self-review was fixed on this head and covered through the public WebSocket hook.

### Testing

- Ruff format and lint passed.
- BasedPyright passed with 0 errors.
- Focused prewarm integration coverage passed (11 tests).
- Full mitm-addon suite passed (5,293 tests; 58 existing dependency warnings).
- Lefthook pre-commit passed.
- Production reconciliation remains unverified until deployment, old-runner drain, and a complete UTC reporting window.

### Verdict

LGTM

PR #26419, opened 2026-08-11
refactor(platform): mermaid blob urls, code-block invalid fences, real parse in tests

e7h4n:
## Summary

Three related mermaid changes:

1. **Blob URLs instead of data URLs** (perf). `diagram$` resolves to a blob URL + `File`; assigning a multi-KB `data:` string to `src` showed up in commit-phase profiles as a per-mount cost. After review: `createMermaidDiagramSignals(code, ownerSignal)` takes its blob-URL owner explicitly — the chat registry hands entries the application root, disposable preview trees hand their own surface lifetime (`createObjectUrlResource`). Each resolved theme caches one object URL, a hard two-entry bound per diagram; theme flips reuse the prior rendering, owner teardown revokes.
2. **No invalid-source fallback** (behavior). A mermaid fence now has exactly two presentations: a diagram when the mermaid parser accepts the source, and an ordinary code block (`pre > code.language-mermaid` + copy button, same as any other fence) when it does not. `diagram$` resolves `null` for rejected sources instead of throwing; the dedicated `mermaid-diagram-fallback` UI is gone. `CodeCopyButton` moved to `code-copy-button.tsx` so both views share it.
3. **Real parser in tests** (test fidelity). The happy-dom mermaid stub no longer approximates validity with a hand-maintained keyword table: `parse` is delegated to the real mermaid module through a `mermaid-real` alias (same pattern as `idb-real`), so test validity can never drift from production. `render` stays a stub — real rendering needs the SVG measurement APIs of a browser layout engine (`getBBox`), which happy-dom does not implement (verified empirically: geometry stubs still produce degenerate/empty output). Files that await a rendered diagram warm the real parser in `beforeAll` (`warmMermaidParser`), keeping the one-time module transform out of the first assertion's waitFor budget.

## Tests

- `keeps an invalid mermaid fence as an ordinary code block` — exercises the real parser end to end.
- Theme-switch test asserts light/dark URLs are cached and never revoked while the owner lives.
- Registry identity test provides the root signal the registry now requires for blob-URL ownership.

PR #26360, opened 2026-08-11
refactor(platform): parse chat markdown into per-event hast trees

e7h4n:
## Summary

Removes react-markdown from the chat transcript and replaces it with a direct unified pipeline that parses each chat event into a single hast tree, rendered with `toJsxRuntime`. User-visible behavior is unchanged (guarded by a byte-identical parity test against the legacy `MarkdownPreview` output).

### Architecture

- **One tree per event.** `chatEventTreePlan` plans a single markdown document per event; trees live in a ccstate state map keyed by event id, written only by the internal `ensureEventTrees$` command (immutable read-build-write, no updater callbacks).
- **Window-driven parsing.** Only events inside the visible render window are parsed. Every write point that can change the window runs the ensure step afterwards: the event sync, the load-more cursor, and the scroll position commands. Reading never parses. The construction-order cycle between the render window and the scroll signals is resolved through a read-only position view derived from module-scope state; the position-writing scroll commands become async and take an external `AbortSignal` (React callers pass `pageSignal$`).
- **No render-time key lookups anywhere.** Signals are resolved by the command that parses or opens, and handed to the view:
  - Cards and mermaid share one path. Markers travel on hast `node.data`, which `rehype-raw` cannot produce, so raw HTML cannot forge a card, a copy button, or a diagram. Trees carry `data.mermaidSignals` and `data.imageLoadSignals` embedded at parse time; card recognition only references signals registered ahead of the parse — unresolved URLs stay ordinary links.
  - Sidebar targets carry their signals: the mail-draft card hands its own signals to the open command, artifact refs carry `text$` (reusing the caller's fetch cache) and a prepared `markdownTree$` for markdown previews. The by-id/by-url projections (`mailDraftCardSignalsById$`, `artifactSignalsByUrl$`) are deleted; card registries are register-only state maps that never expose their map.
  - The global keyed image-load map, its dataset ref and refcounted GC are replaced by per-image `ImageLoadSignals` owned where the image is prepared: markdown trees embed per-URL registry entries, artifact cards / browser sessions / composer attachments / catalog pages hold their own.
- **Mermaid is a content-addressed pull model.** Each unique diagram source owns a `diagram$` async computed rendering to a `data:` SVG URL — nothing to revoke, no lifetime management, one entry per source shared across threads and copies. Shared thread pages parse assistant bodies in their setup command; markdown artifact previews carry a prepared tree computed attached where the preview opens. Surfaces that still parse during render keep mermaid fences as highlighted code.

### Performance

[...]

PR #26497, opened 2026-08-12
refactor(platform): scope markdown signal registries to their surfaces

e7h4n:
## Summary

Follow-up to the registry audit: the last two module-scope signal registries move into the surfaces that own them, as immutable map `state`s behind register-only commands. No keyed store is reachable from anywhere: registries never expose their maps, and every entry is embedded at parse time into the tree the consumer renders from.

- `registerMermaidDiagram$` (module command over a module map, blob URLs owned by the application root) → `createMermaidDiagramRegistry(ownerSignal)`. The chat thread creates one per `createPagedEventResources` alongside the card registries, owned by the thread lifetime; the shared thread page creates one per page setup. Tearing the surface down now releases exactly the diagrams it rendered — this also removes the previously unbounded session-lifetime growth (review finding RC-3 on #26360) and drops the `rootSignal$` dependency from the mermaid path entirely.
- `registerImageLoad$` (module command over a module URL-keyed map) → `createImageLoadRegistry()`, instantiated per thread and per shared page the same way. Within a surface, streaming re-parses still reuse entries so a reused `` never strands behind its placeholder.

Trade-off: the same fence in two different threads now renders once per thread instead of sharing one session-global entry. That sharing only ever helped cross-thread revisits, which already re-parse their trees; boundedness wins.

Stacked on #26419 (base `perf/mermaid-blob-url`); retarget/rebase onto `main` after that merges.

### DIFF
--- a/turbo/apps/platform/src/signals/chat-page/create-chat-thread.ts
+++ b/turbo/apps/platform/src/signals/chat-page/create-chat-thread.ts
@@ -121,10 +121,15 @@ import {
   type CardDescriptorBlock,
 } from "./parse-body-blocks.ts";
 import {
+  createMermaidDiagramRegistry,
   embedMermaidSignals,
-  registerMermaidDiagram$,
+  type MermaidDiagramRegistry,
 } from "../mermaid-diagram.ts";
-import { embedImageLoadSignals, registerImageLoad$ } from "../image-load.ts";
+import {
+  createImageLoadRegistry,
+  embedImageLoadSignals,
+  type ImageLoadRegistry,
+} from "../image-load.ts";
 import {
   chatEventTreeContent,
   chatEventTreePlan,
@@ -1683,6 +1688,8 @@ interface EventTreeRegistries {
   readonly browserSessionSignals: ReturnType<
     typeof createBrowserSessionSignals
   >;
+  readonly mermaidDiagrams: MermaidDiagramRegistry;
+  readonly imageLoads: ImageLoadRegistry;
 }
 
 function createEventTreeSignals({
@@ -1694,6 +1701,8 @@ function createEventTreeSignals({
   planUpgradeCardSignals,
   mailDraftCardSignals,
   browserSessionSignals,
+  mermaidDiagrams,
+  imageLoads,
 }: EventTreeRegistries) {
   interface EventTree {
     readonly content: string;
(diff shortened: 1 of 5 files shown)


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 79f2fad62d08fa2cc8b1f76677c784a0ddf87071
PR_AUTO_REVIEW_VERDICT: LGTM

## Summary

Reviewed the rebased head (`cc9f861` + an empty compare-refresh commit; 5 files, +108/−79 against main). The last two module-scope signal registries move into the surfaces that own them: `registerMermaidDiagram$`/`registerImageLoad$` (module commands over module maps) become `createMermaidDiagramRegistry(ownerSignal)`/`createImageLoadRegistry()` — immutable map `state`s behind register-only commands, instantiated per chat thread in `createPagedEventResources` (alongside the existing card registries, owned by the thread lifetime) and per shared-thread-page setup. Entries are still embedded into trees at parse time; no map is reachable from any consumer, and no render-time key lookup exists. This also removes the mermaid path's `rootSignal$` dependency and bounds registry growth by surface lifetime (the previously noted unbounded session-lifetime growth is gone).

The rebase onto main correctly composes with the render-serialization fix that merged in #26419: `createMermaidDiagramSignals` keeps the `mermaidRenderQueue$` critical section untouched, and the queue stays per-store while registries become per-surface — all surfaces in one store still serialize through one chain, so the theme-race guarantee is preserved with more registry instances. The old module registry and `rootSignal$` import have zero remnants.

Trade-off is stated and acceptable: the same fence in two threads renders once per thread instead of sharing one session-global entry; that sharing only helped cross-thread revisits, which re-parse their trees anyway.

## Findings

- No P0/P1. The identity test now constructs its registry with an explicit owner (`createMermaidDiagramRegistry(context.signal)`), which also let the earlier `setRootSignal$` test scaffolding be deleted — the test matches production wiring more closely than before.
- Note: the empty `chore: refresh pr compare` commit exists only to unwedge GitHub's stale merge-base after the post-squash rebase; squash-merge collapses it.

## Testing

- 51 tests across the affected surface pass locally on the rebased head: signals/mermaid-diagram (including the overlapping-theme-render regression test now on main), markdown-behavior, markdown, thread-sidebar, shared-thread-page.
- Platform `check-types` clean on a fresh build-info; registry wiring verified at both instantiation sites (`createPagedEventResources` thread owner, shared-page setup signal).

PR #26527, opened 2026-08-12
refactor(python): use bounded reads for runner flush markers

### LINKED ISSUE #26516: refactor(python): use bounded state-file reads for runner flush markers
## Summary

The shared Python reader for runner flush request markers bypasses the existing safe state-file boundary:

- `crates/runner/mitm-addon/src/runner_flush_request.py:14-23` calls `Path.read_text()` directly, with no size bound and no opened-descriptor validation.
- `crates/runner/mitm-addon/src/state_file.py:44-61,64-105` is already the canonical Python owner for bounded reads of runner-owned state files. It opens with available `O_NOFOLLOW`, `O_NONBLOCK`, and `O_CLOEXEC` flags, validates the opened descriptor is a regular file, and probes one byte beyond a caller-supplied limit.

Both runner-to-addon marker protocols use the unbounded reader:

- `crates/runner/mitm-addon/src/usage/counters.py:106-121` reads `usage-flush-request` for the signal-driven usage drain.
- `crates/runner/mitm-addon/src/runner_flush_lifecycle.py:254-271` reads `jsonl-flush-request` from the continuously polled JSONL watcher.

This leaves two representations of the same runner-owned state-file read policy: registry/catalog state uses `state_file`, while runner flush markers use ordinary path reads.

## Concrete impact

`Path.read_text()` follows symlinks, reads regular files without a bound, and blocks when the marker path is a FIFO. A focused reproduction against current `main` replaced the marker with a FIFO and invoked `read_runner_flush_request()` under a one-second timeout; the call did not return and exited through the timeout (`124`).

The blocking case is lifecycle-significant:

- `runner_flush_lifecycle.py:107-153` polls the JSONL marker in a background thread and stops it with an unbounded `worker.join()`.
- `runner_flush_lifecycle.py:156-190` runs usage-flush reads while holding the usage-flush worker lock.
- `runner_flush_lifecycle.py:310-326` waits on those owners during addon shutdown.

A stale, corrupted, or replaced FIFO marker can therefore strand JSONL watcher shutdown or the usage drain rather than being treated as an absent/invalid request. An oversized marker can also allocate and decode unbounded content during the same shutdown-sensitive paths.

This is distinct from #25206: that issue correctly centralized the common marker envelope, while generic state-file I/O changes were explicitly left out of its scope. The resulting shared `runner_flush_request` module is now the narrow place to apply the existing state-file contract once for both consumers.

## Acceptance criteria

- Define a small explicit maximum size for runner flush request marker JSON.
- Route `read_runner_flush_request()` through `state_file.open_state_file()` and `OpenedStateFile.read_bytes()` (or an equivalently narrow shared wrapper) instead of `Path.read_text()`.
- Preserve the current valid envelope contract and missing/unreadable/invalid UTF-8/invalid JSON behavior: both consumers should continue treating invalid markers as no request.
- Reject symlinks, FIFOs, directories, other non-regular files, and oversized marker files without blocking or reading their targets.
- Cover the shared filesystem behavior through both the usage-flush and JSONL-flush consumer entry points, including a bounded FIFO regression.
- Do not change marker filenames, JSON fields, generation matching, request-ID rules, polling intervals, acknowledgement shapes, or shutdown deadlines.

seven332:
## Summary

- read shared runner flush request markers through the existing no-follow, nonblocking, regular-file-validated state-file boundary
- cap marker reads at 64 KiB while preserving explicit UTF-8 decoding and the existing envelope contract
- cover symlink, FIFO, directory, and oversized valid markers through both usage and JSONL consumer paths

## Explicit exclusions

- no marker filename, JSON field, generation, request-ID, JSONL path, or acknowledgement changes
- no polling interval, shutdown deadline, worker lifecycle, Rust writer, dependency, migration, or rollout changes
- no new state-file API or fallback behavior

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_runner_usage_flush_signal.py` (`50 passed`)
- `uv run --no-sync python -m pytest tests/` (`5306 passed`)

Closes #26516

### DIFF
--- a/crates/runner/mitm-addon/src/runner_flush_request.py
+++ b/crates/runner/mitm-addon/src/runner_flush_request.py
@@ -5,6 +5,10 @@ from collections.abc import Callable
 from pathlib import Path
 from typing import NamedTuple
 
+import state_file
+
+MAX_RUNNER_FLUSH_REQUEST_BYTES = 64 * 1024
+
 
 class RunnerFlushRequest(NamedTuple):
     marker: dict[str, object]
@@ -18,7 +22,12 @@ def read_runner_flush_request(
 ) -> RunnerFlushRequest | None:
     """Read a runner flush request for the expected addon generation."""
     try:
-        marker: object = json.loads(marker_path.read_text(encoding="utf-8"))
+        with state_file.open_state_file(
+            marker_path,
+            description="runner flush request marker",
+        ) as opened_file:
+            marker_bytes = opened_file.read_bytes(MAX_RUNNER_FLUSH_REQUEST_BYTES)
+        marker: object = json.loads(marker_bytes.decode("utf-8"))
     except (OSError, UnicodeDecodeError, json.JSONDecodeError):
         return None
 
--- a/crates/runner/mitm-addon/tests/test_runner_usage_flush_signal.py
+++ b/crates/runner/mitm-addon/tests/test_runner_usage_flush_signal.py
@@ -360,17 +360,13 @@ class TestRunnerUsageFlushSignal:
                 assert usage.read_usage_flush_request_id() is None
                 return
 
-            with (
-                patch.object(
-                    runner_flush_lifecycle,
-                    "__file__",
-                    str(runner_usage_flush_files.lifecycle_file),
-                ),
-                patch.object(logging_utils, "flush_log_path") as flush_log_path,
+            with patch.object(
+                runner_flush_lifecycle,
+                "__file__",
+                str(runner_usage_flush_files.lifecycle_file),
             ):
                 runner_flush_lifecycle._flush_jsonl_for_runner_request()
 
-            flush_log_path.assert_not_called()
             assert not runner_usage_flush_files.jsonl_flush_state_path.exists()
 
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #26527

**Reviewed head:** `09cbe092eb988abe02c00bddf46d8001910ade01`

### Summary

The shared usage and JSONL flush marker reader now reuses the established no-follow, nonblocking, regular-file-validated state-file boundary with a 64 KiB limit. The change preserves existing envelope and protocol semantics and adds real-filesystem rejection coverage through both consumer paths.

### Findings

No findings.

### Testing

- Coverage is adequate: both consumers cover valid and malformed envelopes plus symlink, FIFO, directory, and oversized valid marker files; low-level descriptor cleanup and overrun behavior remain covered by `test_state_file.py`.
- Local validation passed: uv lock/sync, Ruff format/lint, BasedPyright, focused pytest (`50 passed`), and the complete addon suite (`5306 passed`).
- GitHub CI results are evaluated separately by the workflow readiness check.

### Verdict

LGTM

PR #26534, opened 2026-08-12
perf(python): bound sigv4 query-pair cardinality

### LINKED ISSUE #26514: perf(python): bound SigV4 query-pair cardinality
## Problem

At current main (01cc776abf7797e054b82a270541c60eb036a3cb), managed AWS SigV4 handling caps the raw request target at 64 KiB in crates/runner/mitm-addon/src/auth.py:102-110 and 619-637, but it does not bound how many query pairs fit inside that target.

The allowed target is then expanded into substantially more synchronous work:

- crates/runner/mitm-addon/src/aws_sigv4.py:263-277 splits the complete query and allocates one decoded tuple per pair.
- aws_sigv4.py:309-317 and 847-859 scan the complete tuple separately for each required presigning field.
- aws_sigv4.py:454-497 and 685-729 filter and encode the full pair set twice during query signing.
- aws_sigv4.py:500-555 quotes, allocates, sorts, and joins the complete pair set for canonicalization.

A syntactically valid 65,536-byte presigned target can contain more than 21,000 tiny pairs and still pass the byte limit introduced by #26101. That closes the unbounded-byte path, but leaves a high-cardinality case with a much larger constant and allocation cost inside the accepted boundary.

## Evidence and impact

Measured on the exact main commit above with Python 3.12.3 on aarch64. The probe used the public inspect_request() and sign_request() boundaries, two warmups, and nine-run medians. Both cases were valid body-independent S3 presigned requests with the same six X-Amz signing fields, one Host header, and an exact 65,536-byte request target. Signing reused the prepared inspection, matching the production flow-local reuse path.

| Query shape | Pairs | Inspection median | Signing median | Total CPU |
| One large ordinary value | 7 | 7.141 ms | 1.163 ms | 8.304 ms |
| Tiny repeated ordinary pairs | 21,763 | 40.011 ms | 67.681 ms | 107.691 ms |

The dense request therefore consumes about 13 times more synchronous CPU at the same admitted byte size. Production reaches inspection from mitm_addon.py:856-869 and performs final signing through auth.py:640-690 and 699-722 on the shared mitmproxy event loop. A sandbox can repeat valid placeholder requests to a configured SigV4 connector, so roughly ten such requests per second are enough to consume one event-loop core in this local measurement and delay unrelated proxy flows.

This is distinct from the completed #26101 work: that issue removed repeated cross-phase inspection and established byte/header bounds, while this case stays within those bounds and exploits query-pair cardinality.

## Acceptance criteria

- Establish and document a deterministic SigV4 query-pair work bound, with an AWS-compatibility rationale for the selected threshold.
- Enforce the bound before materializing the complete decoded pair tuple or performing repeated per-pair classification/canonicalization work.
- Over-bound dense queries fail through the existing managed SigV4 auth failure path without exposing resolved credentials or reaching upstream.
- Public-boundary and hook-level tests cover accepted boundary input and an in-byte-limit, over-pair-limit presigned request without timing assertions.
- Preserve canonical signatures, duplicate and blank-value behavior, pair ordering behavior, and existing diagnostics for accepted requests.
- Focused SigV4/auth checks and the complete mitm-addon Python test suite pass.

seven332:
## Summary

- cap managed SigV4 input at 8,192 non-empty query pairs, covering the documented ALB 16 KiB request-line ceiling and smaller CloudFront/API Gateway URL ceilings
- reject pair 8,193 before raw query splitting, percent decoding, or decoded pair materialization
- preserve accepted duplicate, blank, bare, empty-segment, ordering, and canonical signature behavior
- cover the public signer boundary and the real S3 request hook's accepted and fail-closed paths

## Behavior and compatibility

The limit counts every non-empty input pair, including `X-Amz-*` fields. Empty `&&` segments remain ignored. Resolved session credentials may add one signer-owned `X-Amz-Security-Token`, so an accepted 8,192-pair input can deterministically produce 8,193 output pairs.

Over-bound managed requests continue through the existing conservative inspection path and finish as local HTTP 502 `aws_sigv4_auth_failed` responses without placing resolved credentials into the request or continuing upstream.

## Exclusions

- no changes to non-SigV4 proxy traffic
- no changes to the existing 64 KiB request-target/header limits or auth hook ownership
- no API, schema, dependency, protocol, migration, environment, feature-switch, or rollout changes

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_aws_sigv4.py tests/test_request_handler_aws_sigv4_body.py -q` — 128 passed
- `uv run --no-sync python -m pytest tests/` — 5,305 passed

Closes #26514

### DIFF
--- a/crates/runner/mitm-addon/src/aws_sigv4.py
+++ b/crates/runner/mitm-addon/src/aws_sigv4.py
@@ -60,6 +60,12 @@ _RAW_WHITESPACE_CHARS = frozenset(" \t\n\r\f\v")
 _ASCII_CONTROL_MAX = 0x1F
 _ASCII_DELETE = 0x7F
 _DEFAULT_PORTS = {"http": 80, "https": 443}
+# ALB caps the complete request line at 16 KiB; CloudFront and API Gateway
+# impose smaller URL limits. Even one-byte non-empty query pairs require
+# 2 * count - 1 bytes with separators, so this admits every request within
+# those documented front-door limits while bounding synchronous signing work.
+MAX_AWS_SIGV4_QUERY_PAIRS = 8 * 1024
+_RAW_QUERY_PAIR_RE = re.compile(r"[^&]+")
 
 AwsSigV4BodyHash = NewType("AwsSigV4BodyHash", str)
 
@@ -744,6 +750,7 @@ def _parse_query_pairs(query: str) -> list[tuple[str, str]]:
     if not query:
         return []
 
+    _validate_query_pair_count(query)
     pairs: list[tuple[str, str]] = []
     for raw_pair in query.split("&"):
         if not raw_pair:
@@ -751,6 +758,12 @@ def _parse_query_pairs(query: str) -> list[tuple[str, str]]:
         raw_key, _separator, raw_value = raw_pair.partition("=")
         pairs.append((_unquote_query_component(raw_key), _unquote_query_component(raw_value)))
     return pairs
+
+
+def _validate_query_pair_count(query: str) -> None:
+    for pair_count, _match in enumerate(_RAW_QUERY_PAIR_RE.finditer(query), start=1):
+        if pair_count > MAX_AWS_SIGV4_QUERY_PAIRS:
+            raise AwsSigV4SigningError("AWS request has too many query parameters")
 
 
 def _unquote_query_component(value: str) -> str:
--- a/crates/runner/mitm-addon/tests/test_aws_sigv4.py
+++ b/crates/runner/mitm-addon/tests/test_aws_sigv4.py
@@ -7,6 +7,7 @@ import pytest
 
 import runtime_url_parsing
 from aws_sigv4 import (
+    MAX_AWS_SIGV4_QUERY_PAIRS,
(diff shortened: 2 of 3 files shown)


seven332:
## Code Review: PR #26534

**Reviewed head:** `8b30e8794afbb455d27416c0b5d0b78ee6b42426`

### Summary

The PR adds a centralized, non-materializing 8,192-pair preflight at the SigV4 query parser boundary and preserves the existing accepted-input signing path. Exact/over-bound public and real-hook coverage verifies accepted signing, fail-closed local responses, credential non-forwarding, and request-state cleanup.

### Findings

No findings.

### Testing

Coverage is adequate and follows the Python addon testing conventions: public signer boundaries, a real `HTTPFlow`, and only the external auth-resolution boundary mocked. Verified `uv lock --check`, locked environment sync, 128 focused tests, Ruff format/check, BasedPyright, and the complete 5,305-test addon suite. Turbo/Vitest, a local dev server, and live mitmproxy interception were not run for this Python-only change.

### Verdict

LGTM

PR #26535, opened 2026-08-12
fix(chat): keep skeleton through snapshot hydration

e7h4n:
## Summary

- replace remote-request completion with a presentation-readiness signal for the initial chat handoff
- keep the skeleton mounted until snapshot or row events are persisted to IndexedDB and the visible event projection has finished
- preserve the fast path for cached and optimistic events, and add a page-level regression test for the transient empty state

## Root cause

With ChatEventSnapshotRead enabled, the reader resolved initial loading as soon as the remote response arrived. The skeleton also had a raw-events shortcut, while visible rendered groups were still being prepared asynchronously. That allowed the empty-chat state to commit between the skeleton and the message list.

The new integration test fails on main by observing a DOM mutation where neither the skeleton nor the first message is present. It passes after the readiness handoff moves behind IndexedDB persistence and event projection.

## Verification

- pnpm format
- pnpm turbo run lint --concurrency=1 --log-order grouped
- pnpm knip
- staged workspace, Platform, and API type checks
- Platform Vitest: chat-event-snapshot-read.test.ts (9 passed)
- Platform Vitest: zero-chat-thread-idb.test.tsx targeted skeleton cases (3 passed)
- Platform Vitest: chat-lifecycle.test.tsx optimistic message case (1 passed)

### DIFF
--- a/turbo/apps/platform/src/signals/chat-page/__tests__/chat-event-snapshot-read.test.ts
+++ b/turbo/apps/platform/src/signals/chat-page/__tests__/chat-event-snapshot-read.test.ts
@@ -206,9 +206,7 @@ describe("chat event snapshot read", () => {
         parts: [{ type: "text", text: "snapshot prompt" }],
       });
       expect(prompt).not.toHaveProperty("contextType");
-      expect(
-        context.store.get(signals.initialRemoteEventsResolved$),
-      ).toBeTruthy();
+      expect(context.store.get(signals.initialEventsReady$)).toBeTruthy();
       expect(rowRequests).toStrictEqual([3]);
 
       await expect(
@@ -265,9 +263,7 @@ describe("chat event snapshot read", () => {
         { id: assistantEventRow.id, seqId: 2 },
       ]);
       expect(rowRequests).toStrictEqual([0]);
-      expect(
-        context.store.get(signals.initialRemoteEventsResolved$),
-      ).toBeTruthy();
+      expect(context.store.get(signals.initialEventsReady$)).toBeTruthy();
       await expect(
         appDb.get(CHAT_EVENT_ROWS_STORE, assistantEventRow.id),
       ).resolves.toStrictEqual(assistantEventRow);
--- a/turbo/apps/platform/src/signals/chat-page/chat-event-signals.ts
+++ b/turbo/apps/platform/src/signals/chat-page/chat-event-signals.ts
@@ -1,10 +1,4 @@
-import {
-  command,
-  computed,
-  type Command,
-  type Computed,
-  type State,
-} from "ccstate";
+import { command, computed, type Command, type Computed } from "ccstate";
 import type {
   ChatRunOptionsRequest,
   UserMessageInputDocument,
@@ -341,13 +335,11 @@ function createSendChatEvent(
 
 function createChatEventSetup({
   threadId,
-  initialRemoteEventsResolved$,
   initializeIndexedDbEvents$,
   mergePersistentEvents$,
   syncRemoteEvents$,
 }: {
   readonly threadId: string;
-  readonly initialRemoteEventsResolved$: State<boolean>;
   readonly initializeIndexedDbEvents$: Command<Promise<void>, [AbortSignal]>;
   readonly mergePersistentEvents$: Command<
     Promise<void>,
@@ -388,7 +380,6 @@ function createChatEventSetup({
     async ({ get, set }, signal: AbortSignal): Promise<void> => {
       signal.throwIfAborted();
       if (get(optimisticCreateUnsettled$)) {
-        set(initialRemoteEventsResolved$, true);
         return;
       }
       await set(syncRemoteEvents$, signal);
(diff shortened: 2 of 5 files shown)


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 6bdf0190102323c1eac3fd67e4befa6aa94cba3e
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

The change fixes the snapshot-reader handoff at the correct boundary: non-empty remote data becomes presentation-ready only after its IndexedDB write, event merge, registered projection work, visible-tree preparation, and related event-change effects complete. Cached and optimistic events keep their fast paths, but both now finish projection notification before readiness removes the skeleton. Empty and error responses still resolve loading deliberately. I found no blocking correctness, lifecycle, testing, or compatibility issue.

### Review Criteria

| Source | Criterion | Result | Evidence |
| --- | --- | --- | --- |
| `docs/docs.md` | Apply the repository guidance relevant to the changed Platform, state, persistence, and test paths. | Pass | The review covered the core review rules plus event sourcing, effects, cache/ccstate, deployment compatibility, and all Platform testing guides implicated by the diff. |
| `docs/bad-smell.md` | Keep the fix focused and avoid duplicate control paths or unrelated restructuring. | Pass | The production change replaces one loading contract in the existing storage/message pipeline; it adds no parallel loader, duplicate projection, or unrelated refactor. |
| `docs/testing.md` | A `fix:` change must include a regression test at the behavior boundary. | Pass | `zero-chat-thread-idb.test.tsx` reproduces the transient uncovered state with the snapshot reader enabled and observes the skeleton-to-message DOM handoff. Existing storage and optimistic-flow coverage remains intact. |
| `docs/fallback.md` | Declare and justify any newly introduced fallback. | N/A | No new fallback or compatibility branch is introduced. The PR changes behavior inside the existing `ChatEventSnapshotRead` path and preserves its existing legacy-reader selection. |
| `docs/event-sourcing.md` | Persist event data before projecting it, and keep optimistic reconciliation ordered with projection. | Pass | Legacy events write `CHAT_MESSAGES_STORE` before merge/readiness; snapshot rows write `CHAT_EVENT_ROWS_STORE` before row projection; `mergePersistentEvents$` reconciles optimistic events and awaits `notifyChatEventsChanged$` before callers set readiness. |
| `docs/effect.md` | Own asynchronous effects in commands with explicit cancellation and completion. | Pass | The new readiness writes stay inside ccstate commands, every added async boundary propagates the owning `AbortSignal`, and `throwIfAborted()` guards readiness after awaited work. No detached or unowned work was added. |
| `docs/cache.md` | Preserve explicit cache ownership, invalidation, and teardown; do not add an unbounded cache. | Pass | The change reuses the thread-scoped IndexedDB stores and existing signal factory. Test cleanup now also clears `CHAT_EVENT_ROWS_STORE`; no new module cache or resource lifetime is introduced. |
| `.claude/skills/ccstate/SKILL.md` | Keep mutable state private to the signal factory and sequence dependent commands with `await`. | Pass | `initialEventsReady$` is private state exposed as a read-only computed, and optimistic/cache/remote paths await projection notification before setting it. Accessors do not escape command callbacks. |
| `docs/deployment-compatibility.md` | Avoid an incompatible frontend/backend or persisted-state protocol flip. | Pass | This is a same-build internal signal rename and ordering change. API contracts, IndexedDB schema/row shapes, feature-switch protocol, and cross-version persisted data are unchanged. |
| `docs/testing/patterns.md` | Prefer a production-bootstrap Platform page test and boundary-level HTTP mocks. | Pass | The regression uses `detachedSetupPage`, real IndexedDB behavior, `context.mocks.api`, and `context.mocks.http` for the signed snapshot URL. |
| `docs/testing/anti-patterns.md` | Avoid internal mocks, direct fetch stubs, fake timers, and mock-call assertions. | Pass | The new case uses no internal relative-path mock, no fetch stub, no fake timer, and asserts user-visible DOM outcomes rather than implementation calls. |
| `docs/testing/app-testing.md` | Configure mocks before setup, do not await long-running page setup, and let the test context own cleanup. | Pass | All handlers are installed before `detachedSetupPage`; the test waits for rendered states and disconnects its `MutationObserver` while the shared context owns page/mock lifecycle. |
| `docs/testing/testing-external-behavior.md` | Drive and observe the behavior through the page surface. | Pass | The test observes exactly the externally visible invariant: after loading starts, there is never a DOM state with neither the chat skeleton nor the first user message. |

### Findings

None.

### Testing

- Head reviewed: `6bdf0190102323c1eac3fd67e4befa6aa94cba3e`.
- GitHub checks are green, including `ci-gate-turbo`, all eight Platform test shards, `ci-gate-security`, CodeQL, Semgrep OSS, and `codecov/patch`.
- Implementation verification recorded on the PR: format, bounded Turbo lint, knip, staged workspace/Platform/API type checks, snapshot-read tests (9 passed), targeted IndexedDB skeleton cases (3 passed), and the optimistic new-chat case (1 passed).
- The new page-level regression was red on `main` at the uncovered-loading-gap assertion and passes with this readiness ordering.

Issue #26532, opened 2026-08-12
test(rust): make connector runtime tracing capture deterministic


seven332:
## Problem

Structured tracing assertions built on `CapturedEvents` can lose events when Rust tests run in parallel, even when the exercised behavior completes successfully. This makes valid runner changes fail local validation or CI and makes reruns an unreliable mitigation.

The problem is in the shared tracing test capture infrastructure rather than connector-runtime state management. Connector-runtime tests expose it because captured and uncaptured tests can execute the same tracing callsite concurrently.

## Evidence

- A normal parallel run of all 41 `provider::connector_runtime_sync::tests` reproduced the failure immediately: 40 passed and `failed_connector_runtime_sync_retains_last_known_good_and_retries` missed a warning.
- The failed test passes when run alone.
- Running only `failed_connector_runtime_sync_retains_last_known_good_and_retries` and `retry_backoff_preserves_same_run_coalescing_and_caps_delay` in parallel reproduced the missing retry warning on the first attempt.
- In that deterministic pair, the captured test records the API failure warning and verifies the retry state, but loses the immediately following scheduled-retry warning after the uncaptured test first executes the same warning callsite.
- Forty repeated module runs passed with one test thread, while parallel runs failed frequently: 28/40 with two threads, 30/40 with four, 24/40 with eight, and 14/40 with sixteen.

The workspace currently resolves `tracing-core` 0.1.36. This matches the open upstream `tracing-core` v0.1 `has_just_one` bug described in [tokio-rs/tracing#2874](https://github.com/tokio-rs/tracing/issues/2874): a `Dispatch` can be registered but inactive on the thread that first registers a callsite, causing `Interest::Never` to be cached and reused after that dispatch becomes the thread-local default.

## Decision

Fix the shared `tracing-test-support` capture semantics so a callsite first executed concurrently on an uncaptured thread cannot become permanently disabled for an active capture. Keep the connector-runtime logging and state-machine behavior unchanged.

This is one independently reviewable, test-only PR. It does not require production protocol, persistence, or deployment compatibility work.

## Constraints

- Do not serialize unrelated tests.
- Do not add sleeps, retries, or larger timeouts to hide the race.
- Do not remove or weaken structured warning-field assertions.
- Do not add connector-runtime-specific callsite warmup or cache-reset calls.
- Keep capture buffers isolated between concurrent tests.

## Risks

[...]

PR #26538, opened 2026-08-12
perf(runner): batch terminal connector fail-close

### LINKED ISSUE #26501: perf(rust): batch terminal connector fail-close registry updates
## Problem and evidence

When the sync API reports a terminal run, `reconcile_terminal_run` walks every registered connector target sequentially. Each Builtin or Custom target invokes a single-target helper that independently acquires the proxy-registry lock, reads and parses the complete registry, mutates at most one target, serializes the complete registry, and atomically writes it again.

The run target list is intentionally not capped by the 256-target sync request limit, and the runner regression suite covers 257 targets. The registry itself may be as large as 16 MiB. Terminal cleanup therefore scales as `O(targets × registry size)`, can repeatedly process gigabytes of cumulative JSON, and blocks both the shared registry lock and the single connector-runtime worker while doing avoidable work.

This behavior was verified on `main`; regular connector-runtime publication already demonstrates that multiple target mutations can share one checked registry transaction.

## Decision

Proceed with one Rust PR that batches terminal Builtin and Custom fail-close mutations into one dedicated proxy-registry transaction.

## Chosen direction

- Add a terminal-specific batch method that accepts the source IP, run ID, and complete target slice.
- Under one registry lock, load once, recheck source-IP/run ownership, and apply the existing Builtin and Custom matching rules in memory.
- Preserve target-local unchanged, changed, and Custom ownership-error outcomes so one invalid target does not prevent other eligible targets from becoming fail-closed.
- Persist at most once, and only when at least one target changed, using the existing write path that consumes reserved fail-closed capacity.
- Return target outcomes only after the atomic write succeeds. Lock, read, serialization, capacity, and write failures remain batch-wide because no computed mutation was published.
- Update terminal reconciliation to log target-local failures with target identity and one batch warning for transaction failure, while retaining current active-run removal and scheduled-task cancellation behavior.
- Remove superseded single-target production helpers rather than retaining compatibility wrappers.

## Alternatives considered

- Extending the regular publication update enum with terminal variants would mix publication and security-cleanup semantics, make unsupported mixed transactions representable, obscure capacity handling, and still require different target-local error behavior.
- A generic registry transaction closure would broaden the change into a persistence abstraction without a third concrete use case. The dedicated operation is clearer and follows YAGNI.
- Keeping the current per-target loop preserves the verified scaling and lock-contention problem and has no compensating correctness benefit.

## Risks and mitigations

- **Partial validation failure:** Custom ownership errors stay target-local and are detected before that target mutates a policy; other valid targets are still evaluated and committed.
- **False success after persistence failure:** outcomes are returned only after the single write succeeds.
- **Stale work modifying a reused IP:** source-IP/run ownership remains checked inside the registry lock.
- **Capacity regression:** terminal persistence continues to use the fail-closed-capacity-consuming path and retains near-limit coverage.
- **Observability regression:** mixed-target tests verify persisted policies and batch/target-level tracing without mocking internal filesystem code.

## Acceptance criteria

- Terminal reconciliation fail-closes all still-matching Builtin and Custom targets with one registry lock acquisition, one registry load, and at most one persisted registry write.
- Missing or reassigned source IPs and nonmatching targets leave the registry unchanged.
- Source-IP/run ownership checks, Builtin firewall matching, Custom exclusive firewall ownership validation, fail-closed capacity guarantees, task cancellation, and error visibility are preserved.
- A target-local Custom ownership error remains visible while other eligible targets in the same terminal batch can still be committed.
- A mixed Builtin/Custom regression test uses the real temporary registry, exercises multiple targets, proves every eligible policy becomes fail-closed, and observes one batch persistence event.
- Superseded per-target production helpers are removed; relevant Rust format, Clippy, documentation, focused tests, full runner tests, and pre-commit checks pass.

## Delivery notes

This is one independently reviewable PR. It changes no API contract, persisted schema, database, feature switch, deployment ordering, or cross-version protocol.

seven332:
## Summary

- batch terminal connector fail-close updates under one proxy registry lock and one persistence write
- preserve per-target outcomes so an invalid custom connector does not block eligible targets
- cover mixed builtin/custom targets, ownership failures, no-op matching, and near-capacity persistence

## Explicit exclusions

- no API, queue payload, persisted schema, or runner protocol changes
- no changes to non-terminal connector runtime refresh behavior

## Validation

- `cargo check -p runner --all-targets --all-features`
- `cargo test -p runner proxy::registry::tests -- --nocapture`
- `cargo test -p runner provider::connector_runtime_sync::tests -- --nocapture` (39 passed; two existing tracing-capture assertions flaked and passed on immediate isolated reruns)
- `cargo test -p runner`
- `lefthook run pre-commit`

Follow-up: #26532 tracks an existing concurrent tracing-capture test flake observed during focused validation; both affected tests pass in isolation and the full runner suite passes.

Closes #26501

### DIFF
--- a/crates/runner/src/provider/connector_runtime_sync.rs
+++ b/crates/runner/src/provider/connector_runtime_sync.rs
@@ -45,7 +45,8 @@ use super::api::{ApiClient, ConnectorRuntimeSyncOutcome};
 use crate::error::RunnerError;
 use crate::ids::RunId;
 use crate::proxy::{
-    ConnectorRuntimeRegistryUpdate, CustomConnectorRuntimeRegistryState, ProxyRegistryHandle,
+    ConnectorRuntimeFailCloseOutcome, ConnectorRuntimeRegistryUpdate,
+    CustomConnectorRuntimeRegistryState, ProxyRegistryHandle,
 };
 use crate::types::{
     ConnectorRuntimeSyncBatchResponse, ConnectorRuntimeSyncState, ConnectorRuntimeTarget,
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #26538

**Reviewed head:** `72aa338d38c8522adc0409f8a93929b171be1400`

### Summary

The terminal connector fail-close path now preserves its existing ownership, matching, capacity, cancellation, and error semantics while replacing per-target registry transactions with one dedicated batch transaction.

### Findings

No findings.

### Testing

Runner formatting, check, Clippy, documentation, focused registry/terminal tests, the full runner suite, and staged-file pre-commit hooks passed. A production-scale IO benchmark was not run; the implementation and regression assertions verify the single batch persistence boundary. The unrelated concurrent tracing-capture flake observed during focused validation is tracked in #26532.

### Verdict

LGTM

PR #26607, opened 2026-08-12
perf(python): bound connector diagnostic query inspection

### LINKED ISSUE #26551: perf(python): bound connector diagnostic query inspection
## Problem

Connector diagnostics inspect request auth material by fully decoding and materializing every query pair on the shared mitmproxy event loop:

- `crates/runner/mitm-addon/src/connector_diagnostics.py` calls `urllib.parse.parse_qsl(parsed.query, keep_blank_values=True)`, which returns a complete list before the first pair can be examined.
- The same helper is reached from immediate request diagnostics, shared-base pre-auth diagnostics, response-header and buffered-response replacement, and the error hook.
- Request classification retains the complete trusted request URL for these phases. The add-on has no request-target or query-field bound that protects this optional diagnostic inspection.

The diagnostic only needs to establish whether a configured or generic credential name has a nonblank value. Eager `parse_qsl()` work and transient allocation instead scale with every field, including irrelevant empty pairs.

## Verified evidence and impact

Revalidated on current `main` (`689147fc50`) with the repository-pinned Python 3.14.0 interpreter, which matches the Python patch embedded in the pinned mitmproxy standalone runtime. A seven-run median probe reproduced the exact `parse_qsl(..., keep_blank_values=True)` operation with no credential-bearing key:

| Query characters | Single field | Dense empty fields | Single-field median | Dense median |
| 65,536 | 1 | 21,846 | 0.026 ms | 6.285 ms |
| 200,000 | 1 | 66,667 | 0.069 ms | 53.295 ms |
| 1,000,000 | 1 | 333,334 | 0.315 ms | 363.215 ms |

A separate `tracemalloc` run for the one-million-character dense query observed 39.48 MiB peak transient allocation while materializing 333,334 pairs. Timing and allocator totals are environment-dependent; the deterministic defect is unbounded pair decoding and list construction before inspection.

The path is production-reachable. Immediate request checks execute synchronously within the request hook, and response-header, response, and error checks execute in synchronous hooks on the same proxy event loop. The pinned mitmproxy HTTP/1 parser explicitly has no completed-header size protection, so large request targets can reach add-on hooks. Even smaller accepted targets can carry tens of thousands of fields.

An inactive connector candidate or inactive shared-base sibling is sufficient to trigger the scan. Repeated high-cardinality queries can therefore delay unrelated flows in the shared proxy process before the add-on decides whether to synthesize an optional diagnostic.

## Decision

Proceed with a bounded, conservative connector-diagnostic auth-material inspection. The result may permit a diagnostic only after the request was completely inspected within its budget and no auth material was found. If the inspection budget is exceeded, preserve normal upstream handling.

Keep the boundary local to connector diagnostics. A global request-target cap or rejection policy would change unrelated proxy behavior and is not justified by this optional diagnostic problem.

## Constraints

- Bound both total query work and high-cardinality field work; a field-count limit alone does not protect one enormous field.
- Do not build a complete decoded pair list before a credential match can be returned.
- Decode only the key material needed for matching and only decode a value after its key is credential-relevant.
- Preserve standard-library `unquote_plus` behavior for accepted input, including percent decoding, plus handling, replacement of malformed UTF-8, duplicate fields, blank values, configured-name matching, and generic-name case folding.
- Treat budget exhaustion as indeterminate and suppress the optional diagnostic; do not assert that credentials are absent and do not reject or mutate the request.
- Preserve current malformed runtime URL behavior when the query is within the accepted inspection boundary.
- Do not introduce cross-flow cache or persisted metadata.

## Alternatives considered at framing

- `parse_qsl(max_num_fields=...)` bounds cardinality but still constructs the complete accepted list and cannot stop after the first credential-bearing match.
- A query-length check followed by unchanged `parse_qsl()` bounds worst-case size but retains avoidable list allocation and full decoding.
- A global request-target limit would protect more code but changes network semantics beyond the demonstrated diagnostic problem and requires independent design evidence.
- Offloading every diagnostic scan to a thread adds scheduling and lifecycle complexity to a small optional check without eliminating unbounded work.

## Acceptance criteria

- Connector diagnostic auth-material inspection has explicit total-query and field-count work bounds.
- Accepted queries are inspected incrementally without materializing every decoded pair before the first credential-bearing match.
- Budget exhaustion suppresses the optional diagnostic and preserves the ordinary upstream request or response.
- Percent-decoded key matching, plus handling, configured-name case behavior, generic-name case folding, duplicate fields, blank values, and malformed accepted URLs retain their current behavior.
- Integration coverage exercises a dense query within both limits, a query over the field limit, a query over the total-query limit, and a credential match before irrelevant trailing fields, without timing or allocator assertions.
- Existing request, response, error, shared-base, and user-auth connector diagnostic tests pass.
- The work remains one Python-add-on PR with no API, schema, persistence, protocol, migration, or deployment-order change.

seven332:
## Summary

- replace eager connector diagnostic `parse_qsl` materialization with an indexed scanner bounded to 64 KiB query characters and 8,192 nonempty fields
- decode inspected names and only credential-relevant values, stopping as soon as nonblank auth material is found
- treat exhausted inspection budgets as possible auth so request, response, and error hooks preserve ordinary upstream handling
- cover exact and over-limit boundaries, early credentials, encoded names and values, duplicate and blank fields, and response-header behavior through real hook integration tests

## Exclusions

- no global request-target rejection policy
- no cross-hook cache or new flow metadata
- no API, schema, persisted-state, protocol, dependency, feature-switch, or deployment-order changes

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_request_handler_connector_diagnostics.py tests/test_response_handler_connector_diagnostics.py tests/test_error_handler.py` (84 passed)
- `uv run --no-sync basedpyright -p .` (0 errors)
- `uv run --no-sync python -m pytest tests/` (5,338 passed)

Closes #26551

### DIFF
--- a/crates/runner/mitm-addon/src/connector_diagnostics.py
+++ b/crates/runner/mitm-addon/src/connector_diagnostics.py
@@ -56,6 +56,9 @@ _HTTP_STATUS_UNAUTHORIZED = 401
 _HTTP_STATUS_FORBIDDEN = 403
 _HTTP_STATUS_FAILED_DEPENDENCY = 424
 
+MAX_CONNECTOR_DIAGNOSTIC_QUERY_CHARACTERS = 64 * 1024
+MAX_CONNECTOR_DIAGNOSTIC_QUERY_FIELDS = 8 * 1024
+
 _CONNECTOR_DIAGNOSTIC_ELIGIBLE = "_connector_diagnostic_eligible"
 _CONNECTOR_DIAGNOSTIC_ACTIVE_FIREWALL_NAMES = "_connector_diagnostic_active_firewall_names"
 _CONNECTOR_DIAGNOSTIC_CATALOG_SNAPSHOT = "_connector_diagnostic_catalog_snapshot"
@@ -166,7 +169,7 @@ def maybe_make_local_response(
     candidate = _resolve_candidate(flow, original_url=original_url)
     if candidate is None:
         return False
-    if _request_has_auth_material(flow, candidate, original_url):
+    if _request_may_have_auth_material(flow, candidate, original_url):
         return False
 
     flow_metadata.start_request_timing(flow.metadata)
@@ -237,7 +240,7 @@ def maybe_make_firewall_allow_local_response(
         return False
 
     candidate = resolution.candidate
-    if _request_has_auth_material(flow, candidate, original_url):
+    if _request_may_have_auth_material(flow, candidate, original_url):
         return False
 
     flow.metadata[_CONNECTOR_DIAGNOSTIC_CATALOG_SNAPSHOT] = diagnostic_snapshot
@@ -290,7 +293,7 @@ def install_response_stream_if_needed(flow: http.HTTPFlow) -> bool:
     candidate = _resolve_candidate(flow, original_url=original_url)
     if candidate is None:
         return False
-    if _request_has_auth_material(flow, candidate, original_url):
+    if _request_may_have_auth_material(flow, candidate, original_url):
         return False
 
     upstream_status = flow.response.status_code
@@ -367,7 +370,7 @@ def maybe_replace_response(
     if candidate is None:
         return
     upstream_status = flow.response.status_code
-    if _request_has_auth_material(flow, candidate, original_url):
+    if _request_may_have_auth_material(flow, candidate, original_url):
         return
 
     _set_failure_metadata(flow, candidate)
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #26607

**Reviewed head:** `45534458b58cec76947f069e3887e96922d3c77b`

### Summary

The connector-diagnostic-only boundary is correctly scoped. The indexed scanner bounds both query size and field cardinality, preserves existing decoded matching semantics for inspected input, stops at the first credential, and conservatively preserves upstream handling when inspection is incomplete. No cross-hook state or external compatibility boundary is introduced.

### Findings

No findings.

### Testing

Real request/response/error hook coverage passed (84 tests), as did Ruff, basedpyright, and the full 5,338-test Python add-on suite. A separate deterministic 50,000-case differential probe matched the previous `parse_qsl` auth decision for bounded queries. Turbo and Rust checks were not run because no cross-language inputs or consumers changed; remote CI is evaluated separately.

### Verdict

LGTM

PR #26643, opened 2026-08-12
test(python): cover auth.base tls handshake cleanup failures

### LINKED ISSUE #26558: test(python): cover auth.base TLS handshake cleanup failures
## Summary

Add deterministic coverage for an immediate TLS handshake failure in the auth.base forwarder after SSLContext.wrap_socket() succeeds.

_ValidatedTLSConnection.connect() transfers socket ownership to the wrapped TLS socket and then performs an explicit handshake in crates/runner/mitm-addon/src/auth_base_forwarder.py:458-480. Its exception handler must clear self.sock, close the currently registered socket, and release that socket from _ForwardRequestAbortHandle.

The current security tests cover successful handshakes and failures raised by wrap_socket() (crates/runner/mitm-addon/tests/test_auth_base_forwarder_security.py:261-358 and :435-492). The lifecycle tests also cover a handshake that is interrupted by the overall forwarding deadline (crates/runner/mitm-addon/tests/test_auth_base_forwarder_lifecycle.py:325-359). However, no test injects an exception directly from do_handshake() after wrapping succeeds.

The existing test infrastructure already exposes this exact boundary through handshake_side_effect in crates/runner/mitm-addon/tests/auth_base_forwarder_helpers.py:102-160, :263-323, and :444-501, but that branch is currently unused.

## Impact

Certificate verification failures, protocol errors, and abrupt TLS EOFs occur after the TCP socket has been wrapped. A regression in this exception path could leave the active socket open or registered with the abort handle, leaking file descriptors and stale ownership while still propagating the expected TLS exception.

A mutation probe confirmed the gap:

- Changing the exception handler so it does not close a successfully wrapped/current socket left all 118 existing test_auth_base_forwarder_* tests green.
- A focused test that injects ssl.SSLError from do_handshake() and asserts socket cleanup fails against that mutant because the socket remains open.
- The same focused test passes against the current implementation.
- The unmodified full mitm-addon suite passes: 5,324 tests.

## Acceptance criteria

- Add deterministic fault injection for do_handshake() raising after wrap_socket() succeeds.
- Assert that the original TLS exception propagates.
- Assert that the active wrapped socket is closed exactly once, conn.sock is cleared, and the abort handle no longer owns a socket that can be shut down again.
- Assert that no HTTP request bytes are sent after the failed handshake.
- Keep the existing successful-handshake, wrap-failure, and deadline-interruption behavior covered.

seven332:
## Summary

- add deterministic coverage for an `ssl.SSLError` raised by the explicit TLS handshake after socket wrapping succeeds
- drive a real `http.client` request attempt and assert the original error propagates before any request bytes reach the wrapped socket
- verify the wrapped socket closes exactly once, `conn.sock` is cleared, and the abort handle no longer owns the socket

## Scope

This is a test-only change in the auth.base transport-security contract. It does not change production code, helper behavior, dependencies, lockfiles, APIs, protocols, configuration, persisted state, or deployment behavior.

## Validation

- `uv lock --check`
- `uv sync --locked`
- new regression node: 1 passed
- `uv run --no-sync python -m pytest -q tests/test_auth_base_forwarder_security.py` (53 passed)
- `uv run --no-sync basedpyright -p .` (0 errors, 0 warnings)
- combined auth.base forwarder contracts (119 passed)
- `uv run --no-sync python -m pytest -q tests/` (5,339 passed)

Closes #26558

### DIFF
--- a/crates/runner/mitm-addon/tests/test_auth_base_forwarder_security.py
+++ b/crates/runner/mitm-addon/tests/test_auth_base_forwarder_security.py
@@ -8,7 +8,7 @@ from unittest.mock import MagicMock, call, patch
 import pytest
 
 import auth_base_forwarder as forwarder
-from tests.auth_base_forwarder_helpers import fake_forwarder_upstream
+from tests.auth_base_forwarder_helpers import FakeSocket, fake_forwarder_upstream
 
 
 class TestAuthBaseForwarderSecurity:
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #26643

**Reviewed head:** `b2e0780618c4bf26a28563c1a14496527e05b61f`

### Summary

The regression test exercises the real HTTP request path through the explicit TLS handshake failure and verifies error propagation plus socket ownership cleanup.

### Findings

No findings.

### Testing

The new node, full security module, combined auth-base contracts, full mitm-addon suite, Ruff, basedpyright, and locked dependency checks passed. Frontend `vitest` and a local dev server were not run because this is a Python-only test change.

### Verdict

LGTM

PR #26699, opened 2026-08-12
feat(platform): split credit usage out of credit balance

Lunarivibe:
## Why

The settings **Credit balance** section answered three different questions in one screen: how fast credits can be spent (usage allowance), how many are left (org credits), and where they went (usage records). Only the last one grows without bound — a 30-day team view is hundreds of rows inside a dialog — and it carried its own tabs and filter state.

It also mislabelled itself for members: non-admins saw a section titled *Credit usage* whose body rendered only the usage pack card, or nothing at all when the org had no pack.

## What changed

**Split**
- New `usage-records` settings section holds the My usage / Team usage records. `?settings=usage` still opens Credit balance, so existing deep links (CLI credit checkout, low-credit chat notice) are unaffected.
- Members now get their own usage records; the balance section keeps the `Credit balance` title for everyone.
- Each section links to the other: `See where these credits went` on the balance, a balance line above the records.

**Credit balance layout**
- Allowance rows print `3,750 left` instead of `3,750 / 5,000 credits`; the bar already drew *remaining*, so the old copy read backwards. The reset line shows the clock when the window resets today and the day when it resets later.
- The credit breakdown legend is gone and the credit additions list is now an always-visible table with a header. The legend and the grant list were the same numbers twice (`Pay as you go 96,168` is the remaining balance of the auto-recharge grant).
- The composition bar is gapped and square-cornered so it stops looking like the filled progress meter directly above it.
- One action row at the bottom of the card: auto-recharge state on the left, `Buy credits` on the right (gated on `canBuyCredits`, deep-links to the billing purchase block). The section previously had no action at all unless the org was on a free tier with a zero balance.

## Checks

`pnpm -F @vm0/app check-types`, `pnpm -F @vm0/app lint`, `pnpm knip --workspace apps/platform`, `pnpm format` all clean. Ran the four affected test files (`org-usage-tab`, `personal-usage-tab`, `zero-sidebar-account-menu`, `settings-dialog`) — all pass; the rest is left to the pipeline.

Copy landed in all 10 locales. Removed the now-unused `billing.usage.allowance.remaining` key.


Lunarivibe:
LGTM


PR_AUTO_REVIEW_COMMIT: d43a02fa1f155de5b6f05cdd7da09e3f1ce5a0c5
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Splits the settings `Credit balance` section in two: `usage` keeps the balance (usage allowance, org credits, credit additions, buy-credits action) and a new `usage-records` section owns the My usage / Team usage records. Reworks the balance card itself — allowance rows print what is left instead of `remaining / limit`, the breakdown legend is replaced by an always-visible credit additions table, and the composition bar is gapped so it no longer mimics the progress meter above it.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic, no unhandled edge cases | Pass | `org-usage-tab.tsx` `formatAllowanceReset` branches same-day vs later and keeps the invalid-date path; `UsageRecordsSection` calls every hook before the `!isAdmin` early return, so hook order is stable across roles |
| `REVIEW.md` | Security (authz, injection, secrets) | N/A | Frontend-only rendering of billing data the section already loaded; no endpoint, input parsing, or credential touched |
| `REVIEW.md` | Type safety, no `any` | Pass | `SETTINGS_SECTIONS` stays `as const` and `SECTION_COMPONENTS` still uses `satisfies Record<SettingsSection, () => ReactNode>`, so the new `usage-records` id is exhaustively checked at compile time |
| `REVIEW.md` | Style and maintainability | Pass | `CreditBalanceSection` loses the tabs/range state it no longer owns; `CreditAdditionTable` is ~80 lines and shares one `GRANT_ROW_GRID` constant between header and rows |
| `REVIEW.md` | Feature regression guard for replaced UI | Pass | `usage` keeps its id, so the CLI checkout return (`apps/cli/src/commands/zero/credit.ts`) and the low-credit chat notice (`apps/api/src/signals/services/zero-chat-events.command.ts`) still land on Credit balance; member behavior covered in `personal-usage-tab.test.tsx` |
| `REVIEW.md` | `feat:` requires integration tests | Pass | `org-usage-tab.test.tsx` covers both sections, the two cross-links, and the buy-credits capability gate; `zero-sidebar-account-menu.test.tsx` covers the account-menu deep link |
| `REVIEW.md` | DB / JSONB contract gate | N/A | No schema, contract, or persisted shape changed |
| `docs/bad-smell.md` | No `any`, no lint or type suppressions | Pass | No `any`, `@ts-*`, or `oxlint-disable` in the diff |
| `docs/bad-smell.md` | Fail fast, no fallback chains | Pass | No new `??`/`||` default on a contract field; `canBuyCredits` is read through the existing `orgPlanCapabilitiesFromBilling` helper rather than a local default |
| `docs/fallback.md` | New fallbacks must be declared and time-boxed | N/A | The diff introduces no compatibility branch; adding a section id is additive, and no old-shape tolerance was added |
| `docs/deployment-compatibility.md` | Cross-version boundaries stay compatible | Pass | No API, DB, runner, or queue boundary changed. The only new cross-version surface is the client-side `?settings=usage-records` param: an old web client (~2 days) that opens such a link drops the unknown param and simply does not open the dialog, and a refresh loads the new client |
| `docs/effect.md` | Render stays pure; commands own transitions | Pass | Section switching and the buy-credits request run in `onClick` handlers only; nothing writes to the Store or allocates resources during render |
| `docs/cache.md` | RC-1/RC-2/RC-3: no render side effects, render-time signal allocation, or unbounded caches | Pass | The new section reuses the existing module-scope signals (`creditBalanceTab$`, `myUsageRange$`, `teamUsageRange$`); no signal factory or map is introduced |
| `.claude/skills/ccstate/SKILL.md` | Reactive async computed instead of imperative fetch in components | Pass | `BalanceSummary` reads the existing `billingStatusAsync$` through `useLoadable` and renders nothing until it resolves; no new fetch command |
| `docs/testing.md` | Integration tests only, at the entry point | Pass | All new tests drive the real settings dialog through `detachedSetupPage` and clicks; no internal component is rendered directly |
| `docs/testing/testing-external-behavior.md` | Platform tests must act and assert through the page | Pass | New cases click `credit-balance-see-usage`, `usage-records-see-balance`, and `credit-balance-buy-credits`, then assert on the rendered section heading |
| `docs/testing/app-testing.md` | Mock through `context.mocks`, `waitFor` for async | Pass | The extracted `mockBillingStatus(overrides)` helper registers through `context.mocks.api`; every new assertion after a navigation is wrapped in `waitFor` |
| `docs/testing/patterns.md` | Reusable mock factories over duplicated fixtures | Pass | The billing payload moved into one parameterized helper instead of being copied for the `canBuyCredits: false` case |
| `docs/testing/anti-patterns.md` | AP-2 no direct fetch mocking; AP-5 no fake timers | Pass | HTTP goes through `context.mocks.api`; determinism uses the project clock override `mockNow`/`clearMockNow` from `src/lib/time.ts`, not `vi.useFakeTimers()` |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

Non-blocking nit: `CreditBalanceActions` renders an empty `<p>` when `autoRecharge.enabled` is false, which leaves an empty flex child in the footer row. Rendering `null` for that branch would be slightly cleaner.

### Testing

- Coverage: Adequate — the split, the reworked balance card, both cross-links, the `canBuyCredits` gate, the member view, and the pt-BR localization path all have cases. `billing.usage.allowance.remaining` was removed along with its only caller.
- Conventions: Compliant — integration-level, page-driven, mocked at the API boundary, no fake timers, no internal mocks.
- Verified additionally on the PR preview (`pr-26699-app.omby.ai`): real free-tier org renders the balance without the allowance card and without the buy-credits action; a populated billing payload renders the allowance rows, gapped composition bar, additions table, auto-recharge state, and the buy-credits deep link into Billing.

Issue #26649, opened 2026-08-12
feat(desktop): move Okou to ai.okou.desktop


lancy:
Parent epic: #26364
Supersedes the pre-adoption Okou application identity selected in #26373. The already published 0.34.0 identity is retired without an application compatibility commitment.

## Goal

Publish Okou with the final macOS identities:

- Production: `ai.okou.desktop`
- Development: `ai.okou.desktop.dev`

Treat this as a one-time pre-adoption identity cutover. Do not attempt an in-place bundle-ID update.

## Production evidence

Read-only evidence collected on 2026-08-12:

- The 24-hour production request-log query was complete and showed successful Okou Desktop traffic from versions 0.33.2 and 0.33.3.
- Okou 0.34.0 had only unauthenticated responses in that window.
- Masked host aggregates showed two online and one offline 0.34.0 hosts, but the currently exposed MaskDB schema cannot group those rows by Desktop product.
- No user identities were retrieved or recorded.

The product owner explicitly confirmed that this is pre-adoption/internal traffic and approved dropping old-Okou compatibility. It does not gate the final identity cutover.

## Scope

- Change packaged production identity and auth scheme to `ai.okou.desktop`.
- Change packaged development identity and auth scheme to `ai.okou.desktop.dev`.
- Replace the pre-adoption Okou callback schemes with the final schemes; do not retain old-Okou callback compatibility.
- Give the final identity a fresh local Electron data root; do not migrate or share the pre-adoption Okou profile.
- Retire the pre-adoption Okou updater line when the final feed becomes live; never serve the final-identity archive as an in-place update to the old bundle.
- Publish an updater manifest/feed for `ai.okou.desktop` and make it the only active Okou updater line.
- Keep the stable new-user Okou DMG route gated on signed/notarized final-identity acceptance, then switch it atomically with a documented rollback.
- Update release workflows, signing assertions, packaged metadata checks, artifact tests, documentation, and every identity reference.
- Verify side-by-side installation with Zero; old-Okou coexistence is not supported.
- Complete real signed/notarized download, clean install, launch, auth callback, permissions guidance, host lifecycle, Computer Use, and final-identity update acceptance before resuming #26370.

## Acceptance criteria

- [x] Production package is named `Okou.app` with bundle ID and URL scheme `ai.okou.desktop`.
- [x] Development package uses bundle ID and URL scheme `ai.okou.desktop.dev`.
[...]

Issue #26364, opened 2026-08-11
feat(desktop): migrate Zero Computer Use to Okou (EPIC)


lancy:
## Epic outcome

Ship **Okou** as a new signed and notarized macOS application identity while keeping **Zero Computer Use** operational throughout a measured migration.

Target identities:

- Production app: `Okou.app`
- Production bundle ID and auth scheme: `ai.okou.desktop`
- Development bundle ID and auth scheme: `ai.okou.desktop.dev`

The migration ends with a final Zero bridge release that directs users to download Okou, safely stops legacy Computer Use activity, and eventually blocks new work only after the production retirement gate passes.

## Why this is needed

Renaming the visible product without changing the bundle ID preserves the existing macOS application identity. This epic intentionally changes the bundle ID to `ai.okou.*`, which makes Okou a distinct macOS identity with separate code-signing identity matching, privacy grants, URL handling, local Electron data, and update lifecycle.

The current Desktop implementation also has migration coupling that must be separated before Okou ships:

- production identity is `ai.vm0.zero.desktop`;
- Desktop auth accepts only the legacy schemes;
- Zero and any future Okou build would currently share one stable updater feed;
- release manifests, ZIP/DMG names, signing assertions, and workflow paths are Zero-specific;
- request telemetry identifies Desktop type, version, request, and session, but not the Desktop product or bundle identity;
- changing the application name changes Electron's default `userData` root, so the existing session partition string alone does not preserve login or preferences.

## Current production evidence

Read-only production evidence collected on 2026-08-11, using the repository's authoritative `isStaffOrg()` definition:

- 107 Computer Use host records exist: 103 in the staff org and 4 across four non-staff orgs and four users.
- One non-staff host was active in the preceding 24 hours on Desktop `0.29.66`.
- Three non-staff users have executed 1,098 Computer Use commands in total.
- One non-staff user executed 1,070 commands in the preceding seven days.
- 1,046 historical non-staff commands succeeded.
- The four non-staff hosts currently have zero durable `chat_threads.computer_use_host_id` bindings.
- Recent production request logs confirm active Desktop heartbeat and command polling, but cannot classify traffic by Zero versus Okou.

This proves that the migration must support external users and cannot be treated as a staff-only packaging rename. Evidence and tracking must remain aggregate and must not expose user identities.

## Dependency clarification

Live checks on 2026-08-11 confirmed:

- `https://app.okou.ai` already serves the production App entry point with HTTP 200.
[...]

PR #26700, opened 2026-08-12
fix(desktop): exit packaged smoke cleanly

lancy:
## Summary

- flush the packaged Desktop smoke ready marker before exiting
- use `app.exit(0)` for the smoke-only path instead of the normal window-closing quit lifecycle
- prevent auth consume window teardown from turning a successful second packaged launch into `SIGTRAP`

## Production failure evidence

Okou Desktop 0.35.0 was successfully signed and notarized in release workflow run #31595101748. DMG verification and the first installed launch passed. The second launch after replacing the install with the ZIP artifact printed the ready marker, then exited with `SIGTRAP` while the normal `app.quit()` lifecycle rejected the closing auth consume window. Artifact upload and the updater manifest were correctly blocked.

## Validation

- `pnpm -F @vm0/desktop lint`
- `pnpm -F @vm0/desktop check-types`
- `pnpm -F @vm0/desktop exec vitest run src/desktop-smoke-test.test.ts`
- `VM0_DESKTOP_PRODUCT=okou VM0_DESKTOP_PLATFORM_URL=https://app.okou.ai pnpm -F @vm0/desktop package`
- copied the packaged `Okou.app` into a temporary Applications path, ran packaged smoke, replaced the app at the same path, and ran packaged smoke again; both exited successfully and reported ready

Refs #26649
Refs #26364

### DIFF
--- a/turbo/apps/desktop/src/main.ts
+++ b/turbo/apps/desktop/src/main.ts
@@ -1427,9 +1427,13 @@ if (!hasSingleInstanceLock) {
         app.exit(1);
         return;
       }
-      console.log(DESKTOP_SMOKE_TEST_READY_MARKER);
-      quitConfirmation.allowQuitWithoutConfirmation();
-      app.quit();
+      await new Promise<void>((resolve) => {
+        process.stdout.write(`${DESKTOP_SMOKE_TEST_READY_MARKER}\n`, () => {
+          resolve();
+        });
+      });
+      app.exit(0);
+      return;
     }
   });
 


lancy:
## Review round 1

LGTM. The diff is limited to packaged smoke mode. It preserves production/user quit behavior, flushes the readiness contract before termination, and avoids the normal BrowserWindow teardown path that caused the post-ready `SIGTRAP`. Local same-path sequential packaged launches both passed.

PR #26702, opened 2026-08-12
perf(api): archive chat event snapshots concurrently

e7h4n:
## Summary

- archive a pass's candidates through a fixed three-worker pool instead of one serial loop, so a long thread no longer stalls every thread queued behind it
- move gzip off the event loop (`gzipSync` → promisified `zlib.gzip`) so the pool actually overlaps instead of serializing on compression

No archive semantics change: the same bodies are produced, the same content-addressed keys are derived, and head publication still goes through the exact-parent CAS.

## Why

Rebuild cost per thread scales with its event count, and production data shows that count is heavily skewed. Measured on `vm0-prod-ro` (2026-08-12):

- rebuild size across active threads: p50 **74** events, p90 **209**, p99 **3,740**, max **11,815**
- one closed hour (11:00–12:00 UTC) produced **386** snapshots rebuilding **108,588** events, of which a single thread accounted for **11,801** — about 11% of the hour's work inside one sequential step

Throughput is not the constraint. The pass quota is 1,000 threads × 6 ticks = **6,000/hour**, while only **34** threads were awaiting archive at sampling time, and measured archive lag is **max 12.5 min, p90 12.2, p50 11.2** — consistent with the 10-minute cadence. What the serial loop does cost is fairness: a short thread queued behind an 11k-event rebuild waits for it.

Compression had to move off the event loop in the same change. `gzipSync` blocks the main thread, so three workers would still have serialized behind one large body and the pool would have bought nothing.

## Concurrency choice

Three workers, pulling from a shared cursor rather than a fixed stride. A stride would strand the other workers whenever one draws the long thread, which is exactly the skew this change targets. Three keeps at most three connections in `readCanonicalEvents` paging and three concurrent R2 PUTs, which is deliberately conservative.

Candidate order is unchanged (`archive_schema_version ASC → last_message_at ASC → id ASC`) and workers claim the next unclaimed index, so the pass still drains oldest-first.

## Verification

Existing entry-point coverage in `cron-snapshot-chat-events.test.ts` already runs multi-candidate passes (the retired-head test archives two threads per pass) and its assertions are explicitly order-independent, so the pool path is exercised without new tests.

Ran locally:

- `tsc -p tsconfig.gateways.json` — pass
- `tsc -p tsconfig.core.json` — pass
- `oxlint` on the changed file — pass
- `eslint --max-warnings 0` on the changed file — pass
- `prettier --check` — pass

Vitest is left to the PR pipeline.

## Follow-ups (not in this PR)

Incremental append is the real fix for long threads — reuse the parent object and append only the new tail instead of rebuilding from sequence 1. That became format-safe again only after the v3→v4 migration converged (v3 heads reached zero on 2026-08-12). Two findings from validating it up front:

- gzip is a multi-member format, so `gzip(A) || gzip(B)` concatenates without decompressing and round-trips correctly through both `gunzipSync` and `createGunzip`.
- but member granularity is expensive. Against a single-member baseline, splitting the same 2,000-line body costs **+10%** at 100 lines/member, **+70%** at 25, **+183%** at 10, and **+377%** at 5. Pending threads observed in production were 3–25 events behind, so a naive append-per-tick would land in the worst part of that curve. Incremental append therefore needs a compaction policy (full rebuild once member count or size overhead crosses a threshold), not just an append path.

### DIFF
--- a/turbo/apps/api/src/signals/services/cron-snapshot-chat-events.service.ts
+++ b/turbo/apps/api/src/signals/services/cron-snapshot-chat-events.service.ts
@@ -1,5 +1,6 @@
 import { createHash } from "node:crypto";
-import { gzipSync } from "node:zlib";
+import { promisify } from "node:util";
+import { gzip } from "node:zlib";
 
 import {
   chatEventRowV4Schema,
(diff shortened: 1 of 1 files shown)


e7h4n:
Closing: measurement showed this was optimizing a non-bottleneck. Archive lag is 12.5 min against a 1-hour bar and throughput runs 386–985/hour against a 6,000/hour quota, so the head-of-line delay this addressed is not currently costing anything.

The async gzip half is carried by #26705, which needs it to decompress the parent object. Bounded concurrency can come back later with real data behind it — the parent-seeded path adds an R2 GET per thread, so per-thread IO doubles and concurrency will have a measurable case then.

PR #26704, opened 2026-08-12
fix(desktop): bypass teardown in packaged smoke

lancy:
## Summary

- terminate the Electron process directly after the packaged-smoke ready marker is flushed
- avoid closing auth BrowserWindows through Electron teardown in smoke-only mode
- keep the normal application lifecycle unchanged

## Root cause

`app.exit(0)` still closed the Okou auth consume window. Its pending promise rejected during teardown, Electron exited with `SIGTRAP`, and CI rejected the otherwise valid Okou artifact. The smoke-only path now uses `process.exit(0)` after the stdout callback confirms the ready marker was flushed.

## Validation

- `pnpm exec prettier --write apps/desktop/src/main.ts`
- `pnpm -F @vm0/desktop lint`
- `pnpm -F @vm0/desktop check-types`
- `pnpm -F @vm0/desktop exec vitest run src/desktop-smoke-test.test.ts --maxWorkers=4 --silent=passed-only`
- packaged Okou production app built locally and `smoke-test-packaged-app.js` exited successfully

Fixes the release blocker tracked in #26649.

### DIFF
--- a/turbo/apps/desktop/src/main.ts
+++ b/turbo/apps/desktop/src/main.ts
@@ -1432,8 +1432,7 @@ if (!hasSingleInstanceLock) {
           resolve();
         });
       });
-      app.exit(0);
-      return;
+      process.exit(0);
     }
   });
 


lancy:
Review round 1: LGTM.

The change is correctly limited to the packaged-smoke-only branch: stdout is flushed before a direct successful process exit, avoiding Electron BrowserWindow teardown and the Okou auth promise rejection. Normal application shutdown behavior is unchanged. Local packaged Okou smoke validation exits successfully.

PR #26705, opened 2026-08-12
refactor(api): seed chat event snapshots from the head object

e7h4n:
## Summary

Change the archive generation from "rebuild the whole thread from canonical PostgreSQL rows" to "download the head object, append the rows past it, upload the combined body, update the snapshot row".

- a **current-version head** seeds the next generation; the pass reads only `seq_id > head.last_seq_id`
- **no head**, a **retired-version head**, or an **unreadable head** still rebuilds from canonical rows, so version migration and damaged-object recovery are unchanged
- the full-rebuild path now asserts it started at **sequence 1**
- `gzip` moves off the event loop (decompressing the parent would otherwise block the pass on every thread)
- the pass result gains `unreadableParents`, counting heads whose object could not be fetched or decoded

The archive body, object key derivation, content addressing, and the exact-parent CAS are all unchanged. No schema version bump: a generation produces the same bytes it would have produced before.

## Why

`chat_events` is the one unbounded thing in this pipeline — **1,449,228 rows** as of 2026-08-12, growing about **4,400 rows/hour**, with no retention policy. Adding retention requires that the archiver stop depending on PostgreSQL holding a thread's full history, because today it would actively destroy data:

`readCanonicalEvents` reads from `seq_id > 0`, and three properties combine badly once a prefix is pruned:

1. its only guard is `count === 0`, so any surviving tail passes
2. it returns `lastSeqId: candidate.indexedSeqId` — taken from the search watermark, not from the rows it actually read — so the last sequence always looks correct
3. `publishSnapshotHead` compares the parent's identity (id, version, last sequence, object key), not the content

A thread whose events `1..N` were pruned would therefore produce a prefix-less archive carrying a correct-looking last sequence, pass the CAS, and overwrite a complete head. No error, no trace. The sequence-1 assertion added here closes that hole even before any retention work lands.

This also removes the dominant per-thread cost. A rebuild re-reads every event including `payload`: one closed hour produced 386 snapshots that rebuilt **108,588** events, and the largest single thread is **11,815** events.

## Behaviour change worth reviewing

An archived prefix is no longer re-derived from PostgreSQL on every pass, so **in-place edits to already-archived rows stop propagating into the archive**. That is a real change from today. The mechanism for propagating such an edit is the one that already exists: bump `ARCHIVE_SCHEMA_VERSION`, which makes every head retired and rewrites it from canonical rows. Any future column backfill over historical `chat_events` rows needs that bump.

Appending is otherwise sound because archived rows are append-only in practice: a revocation is itself a later event carrying `revokes_event_id`, and `replaceLoadedChatEvent` appends a replacement rather than mutating the original, so the archived prefix stays valid.

## Fallbacks

Both entries are permanent storage-recovery paths, not cross-version compatibility. Neither is sized by a rollout window and neither has a removal condition.

- `cron-snapshot-chat-events.service.ts` / `readParentArchive` — **R2 download failure**. Protects a head row whose referenced object cannot be fetched, by falling back to a canonical rebuild. Surface: none of the version-skew windows. Removal condition: none. The pass now counts it in `unreadableParents`, so an outage is visible rather than silently degrading every thread back to a full rebuild.
- `cron-snapshot-chat-events.service.ts` / `readParentArchive` — **decompress failure**. Protects a head object that is present but undecodable, preserving the property that a damaged archive repairs itself on the next pass instead of wedging the thread forever. Surface: none of the version-skew windows. Removal condition: none. Also counted in `unreadableParents` and asserted by `falls back to a canonical rebuild when the head object is damaged`.

Not introduced by this diff: the retired-version branch (`headArchiveSchemaVersion !== ARCHIVE_SCHEMA_VERSION` → rebuild). Retired heads were already fully rebuilt by the existing candidate selection; this diff only makes the same decision explicit at the read site.

## Verification

Existing entry-point coverage already asserts the property this change has to preserve. `rebuilds canonical full-thread snapshots from database rows` asserts that generation two's prefix equals generation one's lines exactly:

```
expect(secondLines.slice(0, firstLines.length)).toStrictEqual(firstLines);
```

That test now exercises the parent-seeded path and still holds. The corruption test also still passes unchanged — a damaged head fails to decompress, falls back to a canonical rebuild, and publishes a fresh object — so it was renamed from `never reads or transforms an old R2 object during a rebuild` to `falls back to a canonical rebuild when the head object is damaged`, which is what it now verifies.

Ran locally:

- `tsc -p tsconfig.gateways.json`, `tsc -p tsconfig.core.json` — pass
- `oxlint`, `eslint --max-warnings 0` on both changed files — pass
- `prettier --check` — pass

Vitest is left to the PR pipeline. Two pre-existing errors in `tsconfig.tests.json` (`withSecretKmsClientForTest` in `secret-kms-client`) are unrelated to this change and present on `main`.

## Not in this PR

No retention, no deletion of any `chat_events` row. Read paths are untouched. Retention would come later and should gate deletion on `seq_id <= head.last_seq_id`, mirroring how `chat_thread_snapshots.latest_event_id` already gates thread-event pruning.


e7h4n:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 3506851be953b295dfda34dbad8f036a4a28e428

### Summary

Replaces full-thread rebuild-from-PostgreSQL with "seed the next generation from the current-version head object, append the rows past its last sequence". No head, a retired-version head, or an unreadable head still rebuilds from canonical rows, and the rebuild path gains an assertion that it started at sequence 1. Archive body, object-key derivation, content addressing, and the exact-parent CAS are unchanged, so no schema-version bump is needed.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic: no race conditions, wrong conditionals, or bad assumptions | Pass | `archiveThread` derives `fromSeqId` from `headLastSeqId` only when the parent was actually reused; the concatenation covers `1..indexedSeqId` because `headLastSeqId` and the parent object are written together by `publishSnapshotHead` and both are embedded in the content-addressed key. |
| `REVIEW.md` | Type safety: no `any`, no broad types | Pass | New `firstSeqId: number \| null` and `Buffer \| null` return are explicit; no casts added. |
| `REVIEW.md` | `refactor:` commit — existing tests must still cover the code | Pass | `rebuilds canonical full-thread snapshots from database rows` asserts `secondLines.slice(0, firstLines.length)).toStrictEqual(firstLines)`, which is exactly the prefix-preservation property the new path must hold; the corruption test still exercises the rebuild recovery. |
| `REVIEW.md` | Fallback discipline: every fallback introduced by the diff declared in the PR summary | Finding (P1) | `readReusableParentArchive` adds new download-failure and decompress-failure recovery branches; the PR summary has no `Fallbacks` section. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, contract, or `jsonb().$type<>()` change; `turbo/packages/db` and `api-contracts` are untouched. |
| `REVIEW.md` | Feature regression guard / feature switch containment | N/A | Not a feature PR; no UI, route, or user flow changes, and the archive output is byte-identical for unchanged rows. |
| `docs/bad-smell.md` | §7 Fail fast — no silent fallback or recovery logic | Finding (P1) | `readReusableParentArchive` swallows both the R2 download error and the gunzip error via `settle` and returns `null` with no signal anywhere in the pass result. |
| `docs/bad-smell.md` | §3 Error handling — no unnecessary try/catch, fail fast | Finding (P1) | Same two `settle` sites; see P1 below. The deliberate `settle` around `publishSnapshotHead` is unchanged and remains justified (documented expected races). |
| `docs/bad-smell.md` | §1/§2 No `any`, no lint or type suppressions | Pass | None added. |
| `docs/fallback.md` | §1 No negative tests asserting removed behavior stays removed | Pass | The renamed test asserts a live behavior (damaged head recovers), not the absence of deleted code. |
| `docs/fallback.md` | §6/§7 A rollout fallback must name surface, window, and removal condition | Pass | The new branches are not cross-version compatibility; the retired-version branch is the pre-existing migration mechanism, not a new rollout fallback. |
| `docs/fallback.md` | §5 No readers kept for retired producers | Pass | The retired-version head path is the active v3→v4 migration mechanism, which converged on 2026-08-12 but still guards any future version bump. |
| `docs/testing.md` | Integration tests only, at entry points | Pass | Both changed tests are entry-point cron-route tests; no new unit tests. |
| `docs/testing.md` | Mock boundary — no `vi.mock()` with relative paths, real DB, MSW for HTTP | Pass | The diff adds no mocks; existing suite uses the real DB and the fake object store already in place. |
| `docs/testing/api-testing.md` | API behavior asserted through the route, not internal functions | Pass | Assertions go through `runSnapshotCron()` and the published head row/object. |
| `docs/testing/testing-external-behavior.md` | Assert externally observable behavior | Pass | Assertions are on published object bytes, object keys, and head rows. |
| `docs/event-sourcing.md` | Review checklist: shared event IDs, reconciliation, no imperative optimistic removal | N/A | Server-side archival of already-persisted events; no optimistic events or client projections involved. |
| `docs/deployment-compatibility.md` | Old code after deploy / new code before deploy both remain correct | Pass | No migration. A mixed fleet is safe: for the same `(thread, indexedSeqId)` an old instance's full rebuild and a new instance's append serialize the same rows through the same `archiveLine`, so both converge to the same content-addressed key. |
| `.claude/skills/ccstate/SKILL.md` | `command` / `Computed` usage and getter discipline in `turbo/apps/api` | Pass | `readReusableParentArchive` takes the existing `ComputedGetter` and resolves `downloadS3Buffer` through `get(...)`, matching how `putImmutableS3Object` and the GC helpers are already consumed. |

### Findings

- `turbo/apps/api/src/signals/services/cron-snapshot-chat-events.service.ts` — `readReusableParentArchive` silently recovers from two distinct failures. `docs/bad-smell.md` §7 states "No fallback/recovery logic - errors should fail immediately and visibly", and §3 asks for fail-fast over defensive handling. The operational consequence is concrete: during an R2 read outage every candidate silently reverts to a full canonical rebuild, which is the exact cost this PR removes (one measured hour rebuilt 108,588 events across 386 snapshots), and the pass result would still report success with no counter moving. Recommend keeping the self-heal but making it observable — the command already returns a stats object (`snapshots`, `archivedEvents`, `retiredSnapshotReferencesDeleted`, the R2 GC counters); add a counter for parents that could not be reused so an outage or a systematic decode failure is visible in the cron response.

- PR summary — `docs/fallback.md` and `REVIEW.md` require every fallback introduced by the diff to be declared in a `Fallbacks` section of the PR summary. The download-failure and decompress-failure branches are new code paths introduced by this diff and are not declared. Add the section naming each branch, the state it protects, that it is a permanent storage-recovery path rather than a version-skew window, and that no removal condition applies.

### Fallbacks

- `cron-snapshot-chat-events.service.ts` / `readReusableParentArchive` (R2 download failure) — protects a head row whose referenced object cannot be fetched by falling back to a canonical rebuild. Surface: none of the version-skew windows; this is a permanent storage-recovery path, not cross-version compatibility. Removal condition: none evidenced. Declared in PR summary: No — P1. Verdict: justified in intent, not justified as written because it is silent (see P1 above).
- `cron-snapshot-chat-events.service.ts` / `readReusableParentArchive` (gunzip failure) — protects a head object that is present but undecodable, preserving the pre-existing property that a damaged archive repairs itself instead of wedging the thread forever. Surface: none of the version-skew windows; permanent recovery path. Removal condition: none evidenced. Declared in PR summary: No — P1. Verdict: justified in intent, not justified as written because it is silent (see P1 above).
- Not counted as introduced by this diff: the retired-version branch (`headArchiveSchemaVersion !== ARCHIVE_SCHEMA_VERSION` → rebuild). Retired heads were already fully rebuilt by the existing candidate selection; this diff only makes the same decision explicit at the read site.

### Testing

- Coverage: Adequate for a `refactor:` change. The prefix-equality assertion in `rebuilds canonical full-thread snapshots from database rows` is the strongest available guard on the new append path and it was already present; `falls back to a canonical rebuild when the head object is damaged` covers the decompress-failure branch end to end. Two gaps worth noting, neither blocking: the R2 **download-failure** branch has no test (only the decode-failure branch does), and the new sequence-1 rebuild assertion has no test that reaches it. If the observability counter from the P1 above is added, asserting it in the corruption test would cover both the counter and the download branch cheaply.
- Conventions: Compliant. No new mocks, no fake timers, no `expect(mock).toHaveBeenCalled()` assertions, entry-point tests only, real database.

e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 352cf9659cc6af69c64360aa8e88ee0a9640ae01
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Re-review at `352cf96`. Both P1 findings from the `3506851` review are resolved: the parent-seeding recovery is now reported through a new `unreadableParents` counter on the cron result, and the PR summary carries a `Fallbacks` section for the two recovery branches. The underlying change is unchanged — seed the next generation from the current-version head object, append the rows past its last sequence, and fall back to a canonical rebuild for a first archive, a retired-version head, or an unreadable head.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic | Pass | `unreadableParent` is captured once from the `readParentArchive` result and returned on every non-throwing exit of `archiveThread` (concurrent-tail skip, expected publish race, normal publish), so the counter cannot miss a case it should record. The two remaining exits throw, which surfaces louder than a counter. |
| `REVIEW.md` | Type safety: no `any`, no broad types | Pass | The nullable `Buffer \| null` is replaced by a discriminated `ParentArchive` union; `absent` and `unreadable` are now distinguishable at the type level rather than collapsed into `null`. `ArchivedThread` is explicit. |
| `REVIEW.md` | `refactor:` commit — existing tests must still cover the code | Pass | `rebuilds canonical full-thread snapshots from database rows` still asserts `secondLines.slice(0, firstLines.length)).toStrictEqual(firstLines)`, the prefix-preservation property the append path must hold. |
| `REVIEW.md` | Fallback discipline: every fallback introduced by the diff declared in the PR summary | Pass | `## Fallbacks` now lists both branches with surface, removal condition, and the counter that makes each observable, and explicitly excludes the retired-version branch as pre-existing. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema or `jsonb().$type<>()` change. The only contract edit adds one response field. |
| `REVIEW.md` | Feature regression guard / feature switch containment | N/A | Not a feature PR; archive output stays byte-identical for unchanged rows. |
| `docs/bad-smell.md` | §7 Fail fast — no silent fallback or recovery logic | Pass | The two `settle` sites remain, but the outcome is no longer silent: `unreadableParents` is threaded to the cron response, so a storage outage or systematic decode failure is visible in the pass result instead of quietly reverting every thread to a full rebuild. |
| `docs/bad-smell.md` | §3 Error handling — no unnecessary try/catch | Pass | Both recovery sites now have a named outcome and a reported count; the pre-existing `settle` around `publishSnapshotHead` still guards documented FK/unique races. |
| `docs/bad-smell.md` | §1/§2 No `any`, no lint or type suppressions | Pass | None added. |
| `docs/fallback.md` | §1 No negative tests asserting removed behavior stays removed | Pass | `falls back to a canonical rebuild when the head object is damaged` asserts live recovery behavior plus the new counter, not the absence of deleted code. |
| `docs/fallback.md` | §6/§7 A rollout fallback must name surface, window, removal condition | Pass | Both entries state that they are permanent storage-recovery paths outside every version-skew window, with no removal condition — the honest classification rather than a fabricated window. |
| `docs/fallback.md` | §5 No readers kept for retired producers | Pass | The retired-version path is the active migration mechanism for any future version bump. |
| `docs/testing.md` | Integration tests only, at entry points | Pass | The added assertion is on the cron route's returned body inside an existing entry-point test. |
| `docs/testing.md` | Mock boundary, real DB, MSW for HTTP, no fake timers | Pass | No mocks, timers, or fetch stubs added. |
| `docs/testing/api-testing.md` | API behavior asserted through the route | Pass | `expect(rebuilt.unreadableParents).toBeGreaterThanOrEqual(1)` reads the route response, not an internal function. |
| `docs/testing/testing-external-behavior.md` | Assert externally observable behavior | Pass | The counter is part of the endpoint's response contract, so asserting it is external behavior, not an implementation detail. |
| `docs/event-sourcing.md` | Review checklist: shared event IDs, reconciliation, no imperative optimistic removal | N/A | Server-side archival of already-persisted events; no optimistic events or client projections. |
| `docs/deployment-compatibility.md` | Old code after deploy / new code before deploy both remain correct | Pass | No migration. The new `unreadableParents` response field has no consumer outside this app — `cronSnapshotChatEventsContract` is referenced only by the route itself and two test files — so an old instance omitting it breaks nothing. A mixed fleet also still converges: an old instance's full rebuild and a new instance's append serialize the same rows through the same `archiveLine`, producing the same content-addressed key. |
| `.claude/skills/ccstate/SKILL.md` | `command` / `Computed` usage and getter discipline in `turbo/apps/api` | Pass | `readParentArchive` keeps the existing `ComputedGetter` and resolves `downloadS3Buffer` through `get(...)`, matching the other S3 helpers in this service. |

### Findings

No P0 or P1 findings.

Both previous findings are resolved:

- Silent recovery — resolved by `ParentArchive` distinguishing `absent` from `unreadable` and by returning `unreadableParents` from the pass. Worth noting as verified rather than as a finding: when the parent is unreadable *and* the subsequent rebuild fails its own invariants (no events, or a first sequence other than 1), `archiveThread` throws instead of counting. That is the correct precedence — a loud failure beats a counter.
- Undeclared fallback — resolved by the `## Fallbacks` section.

### Testing

- Coverage: Adequate. The decompress-failure branch and the new counter are covered end to end by the corruption test; the append path's prefix equality is covered by the full-thread rebuild test. Two gaps remain, both non-blocking and unchanged from the previous review: the **R2 download-failure** branch has no test, and the sequence-1 rebuild assertion has no test that reaches it. Neither is reachable through the existing fake object store without new failure-injection support, and neither is a regression introduced by this PR.
- Conventions: Compliant. Entry-point test only, real database, no mocks, no fake timers, no mock-call assertions.

PR #26725, opened 2026-08-12
feat: acknowledge steered messages once per burst

tongx-blip:
## Why

Steering — sending while a run is still working — was explained only by a small `CornerLeftUp` arrow beside the bubble, with a tooltip that appears after a 300 ms hover. That is invisible on touch, invisible to anyone who never points at it, and it answers the wrong question at the wrong time: the copy described the mechanism ("Sent while the agent was working to direct its behavior") rather than the thing users are actually anxious about — *did it land, and did I just destroy the earlier work?*

The anxiety peaks when someone fires off three quick corrections in a row. Under the old design each message got its own identical arrow, so nothing ever answered "did it get all three, or only the last one?"

## What changed

**One acknowledgement per burst, counting the messages.** Consecutive user events already arrive as a single `ChatEventGroup` ("a group of consecutive events with the same role"), so the burst boundary is the group boundary and no new grouping logic was needed — this mirrors the existing `UserMessageFeedbackGroup` shape. The acknowledgement renders once, under the last message of the group, as the serif-italic run divider already used for "Model changed to…".

- 1 steer → `Picked this up mid-run — earlier work kept`
- n steers → `All {{count}} messages picked up mid-run — earlier work kept`

Zero's output between two bursts splits them, so each burst gets its own line and its own count.

**The label is repainted in place.** When the count changes, both wordings sit in one box and a single feathered gradient mask sweeps across both — the old wording is erased and the new one drawn without either moving a pixel. A hard `clip-path` edge reads as the text being pushed away, which is why this is a gradient. The mask is deliberately wider than twice the box: at exactly `200%` the resting edge lands mid-ramp and fades out the first glyph. The box width travels with the sweep so the divider rule beside it grows and shrinks to the new wording instead of leaving a gap.

The sweep is driven through `onRef` rather than React state — this app has no `useState`/`useEffect`/`useRef` anywhere, so per-row memory lives on the DOM node the row already owns. The outgoing layer is rendered empty and filled from whatever the row said last, which means a thread opened at three steers has no stale wording to erase and simply draws.

**The next-run model notice is demoted from a rule to a caption.** It uses the same serif-italic hairline, sits two rows below the new acknowledgement, and disappears when the run ends — so the two read as the same kind of thing while meaning opposite ones, and together they imply the steer is going to the next run on the new model, which is wrong on both counts. A rule now marks only what stays in the transcript; live status is a plain caption. Tests lock both halves of that contrast in.

## Notes for review

- **No feature switch added.** This replaces an existing affordance already gated behind `ChatRunContinuationPresentation`, so it inherits that gate. Per the feature-switch skill, a redesign of existing behaviour is not one of the categories that requires a new one.
- **Copy divergence from the reviewed prototype**: the prototype special-cased two messages as "Both picked up mid-run". Shipping that would mean a dead `_two` key in all ten locales plus a hand-rolled branch, so this uses the repo's standard `_one`/`_other` plural idiom with wording that reads correctly at every count ("All 2 messages picked up mid-run").
- **Spacing**: the acknowledgement reuses `RunSectionDividerRow` unchanged, including its `-mt-5`, so it sits at the same distance from its message as the model-change divider does from its. Worth an eye in preview — it is a one-class change if it wants more air.
- `prefers-reduced-motion` collapses the sweep to 1 ms; the wording still updates.

## Verification

- `tsc -p tsconfig.production.json` and `tsconfig.tests.json` — clean
- `oxlint` (exit 0, no findings in touched files), `eslint --max-warnings 0` on touched files — clean
- `i18next-cli extract --ci` + `status` — all 10 locales at 100%; the extractor required `_many` for es/it/fr/pt, which is filled
- `lint:localized-ui`, `lint:static-assets`, `knip` — clean (knip's remaining findings are pre-existing and unrelated)
- `prettier --check` with the lockfile-pinned 3.8.1
- `vitest run` on `chat-run-actions` (10), `chat-run-queue` (32), `chat-history-navigation` + `chat-run-results` (75) — all pass


tongx-blip:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 552ebb1902a7d525596f8f42d73a9154dbccdf64

### Summary

Replaces the per-message steer indicator (a `CornerLeftUp` icon with a 300 ms hover tooltip) with one count-aware acknowledgement per burst of consecutive steers, rendered as the existing serif-italic run divider under the last message of the group, and repainted in place by a feathered gradient mask driven through `onRef`. Also demotes the next-run model notice from a hairline divider to a plain caption. Three P1 issues below; no P0.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness, race conditions, edge cases | Pass | Interrupted sweeps are safe: a second label change remounts the keyed span, `onRef` cleanup aborts the pending timer, and the new mount reads the interrupted target as its start point. Group `<p>` cannot migrate between bursts because `ChatThreadRenderedEventGroups` keys each group by `beginEventId`. |
| `REVIEW.md` | Type safety — no `any`, no broad types | Pass | No `any`; `label: string` widened to `ReactNode` on `RunSectionDivider`/`RunSectionDividerRow` is the minimum needed to pass a node. |
| `REVIEW.md` | Security | N/A | No auth, network, persistence, or user-supplied HTML surface; label text is set via `textContent`. |
| `REVIEW.md` | Lint/type suppressions | Pass | No `eslint-disable`/`@ts-ignore`/`@ts-expect-error` added. |
| `REVIEW.md` | Feature regression guard — existing behavior replaced needs scoping + coverage | **Finding (P1)** | `zero-chat-thread-page.tsx:3198` — the next-run notice demotion ships ungated while the feature that motivates it is behind `ChatRunContinuationPresentation`. |
| `REVIEW.md` | New user-facing behavior gated by a `FeatureSwitchKey` | Pass | The acknowledgement inherits `ChatRunContinuationPresentation`: with the switch off `runPresentation` is `null`, `steerEventIds` is `EMPTY_CHAT_EVENT_IDS`, `steerCount` is 0. Covered by the switch-off test. |
| `docs/bad-smell.md` | No `any`, no suppressions, no over-engineered error handling | Pass | The `onRef` command has no try/catch and lets errors propagate. |
| `docs/fallback.md` §3 | No `??`/`\|\|` chain on a value the owning contract guarantees | **Finding (P1)** | `steer-acknowledgement.ts:23` — `?? ""` on a `data-` attribute the component always renders. |
| `docs/fallback.md` §1 | No negative tests asserting removed behavior stays removed | Pass | The old indicator's test was rewritten to the new behavior, not kept as a tombstone; the switch-off test asserts live product behavior. |
| `docs/fallback.md` §6 | Cross-version rollout fallbacks time-boxed | N/A | No deployable or persisted boundary changes; locale bundles ship with the client. |
| `docs/effect.md` | DOM lifecycle work belongs in `onRef`, acquire/release in one lifecycle | Pass | Measurement and transition run in `onRef`; the timeout is released on `signal` abort. |
| `docs/effect.md` | Keep React render pure | Pass | All DOM writes happen in the commit-phase ref, not during render. |
| `docs/cache.md` RC-7 | Stable callback ref passed directly, cleanup preserved | Pass | `ref={sweepRef}` uses the `useSet` result directly, no inline arrow wrapper. |
| `docs/cache.md` RC-8 | Symmetric resource lifecycles | Pass | The only resource is the `setTimeout`, cleared on abort. |
| `docs/cache.md` RC-5 | Ref callbacks must not bridge two mutable state systems | Pass | The `dataset` entry is ephemeral per-element animation history derived from `count`, not a second copy of an authoritative object. |
| `docs/cache.md` RC-1 | No render-phase side effects | Pass | No writes during render. |
| `.claude/skills/ccstate/SKILL.md` | No `detach()` in `signals/` | Pass | `steer-acknowledgement.ts` contains no `detach`; `onRef` in `signals/utils.ts` owns it. |
| `docs/testing.md` | `feat:` requires integration tests | Pass | 3 new page-level tests plus rewritten parameterized cases in `chat-run-actions.test.tsx`. |
| `docs/testing.md` | Mock at boundary; no internal `../` mocks, no fetch stubbing, no fake timers | Pass | No `vi.mock`, no `vi.stubGlobal`, no `vi.useFakeTimers` added. |
| `docs/testing/app-testing.md` | Page-level tests entering through `detachedSetupPage` | Pass | All new tests use `detachedSetupPage` and assert page-visible text. |
| `docs/testing/app-testing.md` | No component-only unit tests under platform | Pass | No unit test added for the new module. |
| `docs/testing.md` | Coverage for the behavior the PR introduces | **Finding (P1)** | No test drives a count *transition*, so `steer-acknowledgement.ts` has zero coverage. |

### Findings

- `turbo/apps/platform/src/signals/zero-page/steer-acknowledgement.ts:23` — `const label = el.dataset.steerAcknowledgementLabel ?? "";` falls back on a state the component cannot produce: `SteerAcknowledgementRow` always renders `data-steer-acknowledgement-label={label}`, and the command only ever runs on that element. Per `docs/fallback.md` §3 this is fallback slop on a contract-guaranteed value, and it is not inert — if the attribute were ever missing, `""` is written into `row.dataset.steerAcknowledgedLabel`, and the *next* mount reads `""` as a real previous label and sweeps from an empty string instead of failing loudly. Read the value without the chain and return early when it is absent, so the impossible state stays impossible rather than becoming a wrong animation.

- `turbo/apps/platform/src/views/zero-page/zero-chat-thread-page.tsx:3198` — the next-run model notice demotion is applied unconditionally, but the acknowledgement that motivates it is gated behind `ChatRunContinuationPresentation`. For every user with the switch off, this PR silently restyles a GA element with no accompanying benefit and no way to roll it back without a deploy. `REVIEW.md`'s feature regression guard asks for containment when a change should be scoped. Either gate the demotion behind the same switch so the two land together, or split it into its own PR justified on its own merits.

- `turbo/apps/platform/src/views/zero-page/__tests__/chat-run-actions.test.tsx` — every new test asserts a *static* count (a thread rendered with 1, 2, or 3 steers already present). Nothing exercises a count changing while mounted, which is the only path that runs `steer-acknowledgement.ts` at all: the keyed remount, the `previousLabel === undefined` first-render branch, the dataset handoff, and the sweep are all uncovered. The module is the substance of this PR and currently has no test touching it. Add a test that renders one steer, appends a second, and asserts the label becomes the 2-message wording — that alone covers the transition branch and guards the handoff from regressing into a no-op.

### Testing

- Coverage: **Insufficient** — the feature has page-level tests for static counts and for the feature-switch-off path, but no test covers a count transition, leaving the entire new `steer-acknowledgement.ts` module unexercised.
- Conventions: **Compliant** — page-level tests via `detachedSetupPage`, assertions on page-visible text, no internal-path `vi.mock`, no `vi.stubGlobal("fetch")`, no fake timers, no mock-call-only assertions, no tombstone tests. The removed indicator's test was rewritten to the replacement behavior rather than retained.

tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: 040fc0bc12bce87a8bbb5219a1a484004de51a2d
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Replaces the per-message steer indicator (a `CornerLeftUp` icon behind a 300 ms hover tooltip) with one count-aware acknowledgement per burst of consecutive steers, rendered as the existing serif-italic run divider under the last message of the group and repainted in place by a feathered gradient mask driven through `onRef`. The next-run model notice drops its rule only inside the same feature switch. All three P1s from the review of `552ebb1` are resolved.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness, race conditions, edge cases | Pass | Interrupted sweeps are safe: a further label change remounts the keyed span, `onRef` cleanup aborts the pending timer, and the new mount reads the interrupted target as its start point. The row `<p>` cannot migrate between bursts because `ChatThreadRenderedEventGroups` keys each group by `beginEventId`. First mount with no prior wording returns before touching styles, and the CSS resting value (`0%`) already shows the incoming layer, so there is no flash. |
| `REVIEW.md` | Type safety — no `any`, no broad types | Pass | No `any`. `label: string` → `ReactNode` on `RunSectionDivider`/`RunSectionDividerRow` is the minimum needed to accept a node; `label === undefined` is a narrowing guard, not a cast. |
| `REVIEW.md` | Security | N/A | No auth, network, persistence, or HTML-injection surface; wording is set via `textContent`. |
| `REVIEW.md` | Lint/type suppressions | Pass | No `eslint-disable`/`oxlint-disable`/`@ts-ignore`/`@ts-expect-error` added. |
| `REVIEW.md` | Feature regression guard — replaced behavior needs scoping + coverage | Pass | Fixed in `040fc0b`: the next-run notice restyle is now behind `ChatRunContinuationPresentation`, so users without the feature see the divider unchanged. `chat-run-queue.test.tsx:974` asserts the rule survives with the switch off, and e2e `chat.spec.ts:973` covers the same path. |
| `REVIEW.md` | New user-facing behavior gated by a `FeatureSwitchKey` | Pass | Both halves inherit `ChatRunContinuationPresentation` (`enabled: false`, staff-only). With the switch off `runPresentation` is `null` → `steerEventIds` empty → `steerCount` 0 → no acknowledgement; covered by the switch-off test. |
| `docs/bad-smell.md` | No `any`, no suppressions, no over-engineered error handling | Pass | The `onRef` command has no try/catch and lets errors propagate. |
| `docs/fallback.md` §2 | No compatibility code for a non-GA feature-switched feature | Pass | The switch-off branch is the pre-existing behavior preserved by gating, not a compatibility shim for an earlier shape of this feature. |
| `docs/fallback.md` §3 | No `??`/`\|\|` chain on a contract-guaranteed value | Pass | Fixed in `040fc0b`: `?? ""` on the row's own data attribute is gone; the command now returns early instead of fabricating an empty wording that would have been recorded as a real previous label. |
| `docs/fallback.md` §1 | No negative tests asserting removed behavior stays removed | Pass | The old indicator's test was rewritten to the replacement behavior; the switch-off assertions describe live product behavior, not deleted code. |
| `docs/fallback.md` §6–7 | Cross-version rollout fallbacks time-boxed | N/A | No deployable or persisted boundary changes; locale bundles ship with the client. |
| `docs/deployment-compatibility.md` | Independently deployed surfaces stay compatible | N/A | Frontend-only render change; no API, runner, queue, schema, or persisted-state contract touched. |
| `docs/effect.md` | DOM lifecycle work belongs in `onRef`; acquire and release in one lifecycle | Pass | Measurement and the transition run in `onRef`; the `setTimeout` is cleared on `signal` abort. The element parameter is genuinely used. |
| `docs/effect.md` | Keep React render pure | Pass | All DOM writes happen in the commit-phase ref, never during render. |
| `docs/cache.md` RC-1 | No render-phase side effects | Pass | No writes during render. |
| `docs/cache.md` RC-5 | Ref callbacks must not bridge two mutable state systems | Pass | The `dataset` entry is ephemeral per-element animation history derived from `count`; it is not a second copy of an authoritative object. |
| `docs/cache.md` RC-7 | Stable callback ref passed directly, cleanup preserved | Pass | `ref={sweepRef}` uses the `useSet` result directly with no inline arrow wrapper. |
| `docs/cache.md` RC-8 | Symmetric resource lifecycles | Pass | The only resource is the timer, released on abort. |
| `.claude/skills/ccstate/SKILL.md` | No `detach()` in `signals/` | Pass | `steer-acknowledgement.ts` has no `detach`; `onRef` in `signals/utils.ts` owns it. |
| `docs/testing.md` | `feat:` requires integration tests | Pass | `chat-run-actions.test.tsx` covers static counts, burst grouping, burst splitting, the count transition, and the switch-off path. |
| `docs/testing.md` | Mock at boundary; no internal `../` mocks, no fetch stubbing, no fake timers | Pass | No `vi.mock`, `vi.stubGlobal`, or `vi.useFakeTimers` added. |
| `docs/testing.md` | Coverage for the behavior the PR introduces | Pass | Fixed in `040fc0b`: the new count-transition test drives one steer then a second and asserts the outgoing layer carries the previous wording, which only happens if `steer-acknowledgement.ts` ran. |
| `docs/testing/app-testing.md` | Page-level tests entering through `detachedSetupPage` | Pass | All new tests use `detachedSetupPage` and assert page-visible text. |
| `docs/testing/app-testing.md` | No component-only unit tests under platform | Pass | The new module is covered through the rendered page, not directly. |

### Findings

No P0 or P1 issues.

Non-blocking observations, neither worth holding the PR for:

- `zero-chat-thread-page.tsx` — the acknowledgement is keyed by `label`, so switching UI language remounts it and runs one spurious sweep. Cosmetic, self-correcting, and keying by `count` instead would trade it for slightly staler bookkeeping.
- `chat-run-queue.test.tsx:974` asserts the divider survives with the switch **off**, which is the direction that matters; the switch-**on** demotion has no unit test. It is six lines of branch-free JSX also exercised by the design review, so the asymmetry is proportionate rather than a gap worth new scaffolding.

### Testing

- Coverage: **Adequate** — static counts, burst grouping, burst splitting across an assistant reply, the live count transition, the feature-switch-off path, and the gating regression guard for the next-run divider are all covered at page level.
- Conventions: **Compliant** — page-level via `detachedSetupPage`, assertions on page-visible text, no internal-path `vi.mock`, no `vi.stubGlobal("fetch")`, no fake timers, no mock-call-only assertions, no tombstone tests. The removed indicator's test was rewritten to the replacement behavior rather than retained alongside it.

PR #26786, opened 2026-08-13
refactor(guest): centralize private file replacement

### LINKED ISSUE #26770: refactor(rust): centralize Codex private file replacement
## Summary

Codex auth and model-catalog setup maintain two local implementations of the same private atomic-file replacement protocol even though `guest-contracts::runtime_paths` now owns a hardened implementation of that capability. Consolidate these writers under one policy-bearing primitive so durability, path safety, permissions, cleanup, and final-target behavior evolve together.

Audit snapshot: `8c295b9508e918046146fb40476c48d1e5b59e63`.

## Evidence

- `crates/guest-agent/src/codex_auth.rs:172-203` validates and normalizes `.codex`, while `:224-248` independently implements replacement with `NamedTempFile::new_in`, `0600`, `flush`, and `persist`. Both auth modes call it at `:265-275`.
- `crates/guest-agent/src/cli/codex_runtime_config.rs:100-113` writes the model catalog after creating the directory, and `:125-146` repeats the same `NamedTempFile::new_in` / `flush` / `persist` sequence without sharing the auth writer.
- `crates/guest-agent/src/cli/codex_setup.rs:21-26` documents a temporal ownership rule: auth reconciliation validates `CODEX_HOME`, then the catalog writer relies on that completed boundary. The production sequencing is hand-maintained at `:53-57` rather than expressed by the file-write abstraction itself.
- `crates/guest-contracts/src/runtime_paths.rs:955-1017` already provides the canonical private atomic replacement implementation. It pins the parent directory, creates an exclusive `0600` sibling, calls `sync_all`, publishes with descriptor-relative `renameat`, fsyncs the directory, and removes the temporary file on failure. Its public contract is at `:1039-1055`.
- The Codex implementations have already drifted from those guarantees: both reopen the validated directory by path, call only `flush` rather than `sync_all`, and do not fsync the parent after publication.
- Both Codex consumers intentionally replace an existing final symlink without following it, covered by `crates/guest-agent/src/codex_auth.rs:608-635` and `crates/guest-agent/src/cli/codex_runtime_config.rs:397-435`. The current shared helper rejects final symlinks at `crates/guest-contracts/src/runtime_paths.rs:924-952`, so this behavior needs an explicit shared target policy rather than remaining encoded in two ad hoc writers.
- The duplication is historical rather than protocol-specific: auth atomic replacement predates the model-catalog writer, and the hardened `replace_private_atomic` primitive was added later for active-input receipt persistence.

## Impact

Hardening and correctness fixes to private replacement now have three maintenance sites. The first concrete drift is already present: Codex setup can return success after a buffered flush and rename without the file or directory durability guarantees used by the canonical helper. Its path-based validate-then-reopen flow also leaves a parent-directory race that the descriptor-relative implementation avoids. Future changes to final-symlink, ownership, file-type, permission, or temporary-file cleanup rules can diverge independently between `auth.json`, `models.json`, and other private guest files.

## Acceptance criteria

- Make `auth.json` and `models.json` use one shared private atomic-replacement implementation; remove both local `NamedTempFile` writers.
- Reuse or extend `guest-contracts::runtime_paths::replace_private_atomic` so the shared path owns private modes, descriptor-relative parent handling, file sync, atomic publication, directory sync, and failure cleanup.
- Represent the Codex-specific final-target behavior explicitly: preserve safe replacement of an existing final symlink without following its target, or deliberately adopt one documented fail-closed policy for both consumers.
- Continue rejecting symlinked parent directories and directory targets without modifying their targets.
- Add shared regression coverage for private modes, an existing permissive file, final and parent symlinks, directory targets, publication failures, and temporary-file cleanup.

seven332:
## Summary

- centralize private atomic replacement in `guest-contracts` behind an explicit final-target policy
- preserve strict active-input receipt validation while retaining Codex replacement of permissive files and final symlinks without following them
- migrate `auth.json` and `models.json` away from local `NamedTempFile` writers and cover modes, symlinks, publication failures, and cleanup at the shared filesystem boundary

## Exclusions

- no database, API contract, runner protocol, dependency, or rollout migration changes
- no change to Codex startup ordering or no-auth removal behavior

## Validation

- `cd crates && cargo fmt --all`
- `cd crates && cargo clippy --all-targets --all-features`
- `cd crates && cargo doc --workspace --all-features --no-deps`
- `cd crates && cargo test -p guest-agent --lib`

Closes #26770

### DIFF
--- a/crates/guest-agent/src/cli/codex_runtime_config.rs
+++ b/crates/guest-agent/src/cli/codex_runtime_config.rs
@@ -8,6 +8,7 @@ use std::path::{Path, PathBuf};
 use std::time::Instant;
 
 use guest_common::telemetry::record_sandbox_op;
+use guest_contracts::runtime_paths::{self, PrivateFileReplacementTarget};
 
 use crate::error::AgentError;
 
@@ -105,9 +106,12 @@ pub(super) fn write_model_catalog(
         return Ok(());
     };
     record_catalog_operation(MODEL_CATALOG_PREPARE_ACTION, || {
-        std::fs::create_dir_all(codex_home)?;
         let path = model_catalog_path(codex_home);
-        write_model_catalog_json_atomic(codex_home, &path, &serde_json::to_vec(model_catalog)?)?;
+        runtime_paths::replace_private_atomic(
+            path,
+            serde_json::to_vec(model_catalog)?,
+            PrivateFileReplacementTarget::ReplaceFinalEntry,
+        )?;
         Ok(())
     })
 }
@@ -120,29 +124,6 @@ fn record_catalog_operation<T>(
     let result = operation();
     record_sandbox_op(action_type, started_at.elapsed(), result.is_ok(), None);
     result
-}
-
-fn write_model_catalog_json_atomic(
-    codex_home: &Path,
-    path: &Path,
-    serialized: &[u8],
-) -> Result<(), AgentError> {
-    use std::io::Write as _;
-
-    let mut temp = tempfile::NamedTempFile::new_in(codex_home)?;
-    temp.as_file_mut().write_all(serialized)?;
-    temp.as_file_mut().flush()?;
-    temp.persist(path).map_err(|error| {
-        AgentError::Io(std::io::Error::new(
-            error.error.kind(),
-            format!(
-                "failed to replace {} atomically: {}",
-                path.display(),
-                error.error
-            ),
-        ))
-    })?;
-    Ok(())
 }
 
 pub(super) fn startup_config_overrides(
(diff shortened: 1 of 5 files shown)


seven332:
## Code Review: PR #26786

**Reviewed head:** `4b24085d2c299f79981725d1b024dd021fdd2a84`

### Summary

Centralizes Codex auth and model-catalog publication on the shared private atomic replacement primitive while preserving the receipt journal's stricter target policy. The explicit enum keeps each caller's intended final-entry behavior visible.

### Findings

No findings.

### Testing

Validated formatting, Clippy, Rust documentation, all `guest-contracts` tests, and the full `guest-agent` package tests. Real-filesystem tests cover private modes, permissive targets, final and parent symlinks, and failed-publication cleanup. Non-Unix behavior was not exercised locally and its weaker guarantees are documented.

### Verdict

LGTM

PR #26787, opened 2026-08-13
perf(python): stagger firewall auth address attempts

### LINKED ISSUE #26775: perf(python): avoid serial firewall auth address stalls
## Problem

At commit `8c295b9508e918046146fb40476c48d1e5b59e63`, the firewall-auth transport preserves every resolved address but connects to them strictly serially:

- `crates/runner/mitm-addon/src/firewall_auth_client.py:323-343` preserves resolver order.
- `crates/runner/mitm-addon/src/firewall_auth_client.py:353-388` awaits `loop.sock_connect()` for one address before creating or attempting the next socket.
- `crates/runner/mitm-addon/src/firewall_auth_client.py:809-825` gives DNS, all connection attempts, TLS, request I/O, and response handling one 10-second deadline.

If the first address silently drops SYNs instead of failing immediately, that one attempt can consume the complete deadline. A later healthy address is never tried, so a multi-address firewall-auth endpoint becomes unavailable even though one resolved destination is reachable.

## Evidence

A deterministic probe called the production `_open_connected_stream()` helper with:

1. an IPv6 address whose patched `sock_connect()` remained pending;
2. a second IPv4 address representing an immediately reachable endpoint;
3. a 100 ms outer deadline.

The result was:

```text
{'elapsed_deadline_s': 0.1,
 'attempted_addresses': [('2001:db8::1', 443, 0, 0)],
 'second_address_attempted': False}
```

This is not covered by #25403. Its regression at `crates/runner/mitm-addon/tests/test_firewall_auth_client.py:1155-1206` proves fallback only when the first loopback connection is refused immediately. It does not exercise a pending first connection while a later address can succeed.

## Impact

Firewall-auth cache misses and refreshes wait on this transport before connector credentials can be applied. `crates/runner/mitm-addon/src/firewall_auth_cache.py:66-70,315-325` allows four active fetches and sixteen admitted fetches. Four first-address blackholes can therefore occupy every active slot for the full deadline, delay the remaining admitted requests, and return avoidable `FirewallAuthDeadlineExceededError` responses even when later DNS answers are healthy.

This affects both dual-stack answers and ordinary multi-address A/AAAA rotation; it does not require a malformed response or an oversized request.

## Acceptance criteria

- A pending connection to one resolved address must not monopolize the full firewall-auth deadline while another resolved address can connect.
- Use bounded staggered/Happy-Eyeballs-style attempts, or another bounded strategy with equivalent behavior, without changing the trusted resolver, connection-plan, proxy, or total-deadline boundaries.
- Cancel and close every losing or timed-out socket; retain the existing cleanup guarantee for immediate connection failures.
- Add a deterministic public-entry regression through `fetch_firewall_headers()` where the first address remains pending, the second address serves a valid auth response, and the request completes before the total deadline.
- Keep per-fetch connection concurrency explicitly bounded and leave the existing global firewall-auth admission limits effective.
- Preserve the existing immediate-refusal fallback test and proxy/TLS behavior.

seven332:
## Summary

- stagger trusted multi-address TCP attempts so a stalled address cannot consume the total deadline
- cap each fetch at four live connects and rotate the oldest pending attempt to reach later candidates
- cancel and close losing or timed-out sockets while preserving resolver order, proxy/TLS behavior, the total deadline, and global admission limits

## Tests

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_firewall_auth_client.py` (87 passed)
- `uv run --no-sync python -m pytest tests/ -q` (5353 passed)

## Compatibility

- no API, schema, persisted-state, queue, dependency, generated-artifact, or runner-protocol changes
- old and new runners drain independently against unchanged backend behavior

Closes #26775

### DIFF
--- a/crates/runner/mitm-addon/src/firewall_auth_client.py
+++ b/crates/runner/mitm-addon/src/firewall_auth_client.py
@@ -26,6 +26,9 @@ from aws_sigv4 import AwsSigV4Credentials
 MAX_FIREWALL_AUTH_RESPONSE_BODY_BYTES = 256 * 1024
 FIREWALL_AUTH_FETCH_DEADLINE_SECONDS = 10.0
 _MAX_FIREWALL_AUTH_RESPONSE_HEADER_BYTES = 64 * 1024
+# RFC 8305's default cadence; four slots keep the oldest path eligible for about one second.
+_FIREWALL_AUTH_CONNECTION_ATTEMPT_DELAY_SECONDS = 0.25
+_MAX_CONCURRENT_FIREWALL_AUTH_CONNECTION_ATTEMPTS = 4
 _DEFAULT_HTTP_PORT = 80
 _DEFAULT_HTTPS_PORT = 443
 _IPV6_VERSION = 6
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #26787

**Reviewed head:** `9b2dc6e32d95c3bca3f8363b525d7b88c9874a84`

### Summary

The bounded staggered race fixes first-address stalls without changing trusted resolution, proxy/TLS behavior, the total request deadline, or global admission. Socket ownership and cleanup remain explicit across failure, cancellation, rotation, simultaneous completion, and winner handoff.

### Findings

No findings.

### Testing

Public-entry integration tests cover a pending first address with a reachable second address, progress beyond the four-attempt window, the per-fetch cap, deadline cancellation, descriptor cleanup, and immediate-refusal fallback. Ruff format/lint, BasedPyright, the 87-test client file, and the complete 5353-test mitm-addon suite pass locally. Remote CI is checked separately.

### Verdict

LGTM

Issue #26765, opened 2026-08-13
feat: remember the selected video model per chat thread


bingjiez666:
## Goal

Give the video model the same selection-and-default machinery the chat model has: the user picks it
once, the choice sticks to the chat thread, and a personal default seeds new chats. The choice is
enforced by the generation endpoint rather than suggested in the system prompt.

Research backing this plan: `deep-dive/video-model-parity-with-chat-model/research.md`.

## Where the video model lives today

There is no thread-scoped, member-scoped, or org-scoped video model. The only persisted form is
`selection.videoOptions` nested in a generation-template attachment on a single user message
(`turbo/packages/api-contracts/src/contracts/chat-threads.ts:231-239`, `:258`).

`POST /api/okou/video-io/generate` reads `model` from the request body and defaults to
`DEFAULT_VIDEO_MODEL` (`turbo/apps/api/src/signals/services/zero-video-io-generate.service.ts:888`).
Its handler authenticates on org only and has no thread context
(`turbo/apps/api/src/signals/routes/zero-video-io-generate.ts:413-421`). The only thing that makes an
agent use a particular video model today is prompt text telling it which `--model` flag to pass.

### Availability

Every model in `VIDEO_MODEL_CONFIGS` with `public: true` is selectable — 8 models today, all public.

No policy table, no admin-managed allowlist, no per-model plan gating. This is a deliberate departure
from the chat model, whose availability *is* `org_model_policies`. Video generation is already gated
as a whole feature: `capabilities.videoGenerationAllowed !== true` returns 402 `PRO_REQUIRED`
(`zero-video-io-generate.ts:296-300`). That gate stays and is the only availability control.

### Resolution

Resolved once, at run start:

```text
thread.selected_video_model
  ?? member default (org_members_metadata.selected_video_model)
  ?? DEFAULT_VIDEO_MODEL          // dreamina-seedance-2-0-fast-260128
```

A model id no longer present in `VIDEO_MODEL_CONFIGS` is treated as absent and falls through. Stored
values are projected from jsonb without being re-validated against the current catalog, so a retired
id can reach this point.

### Enforcement, and why the pin lives on the run

The resolved model is written onto the run row. `POST /api/okou/video-io/generate` uses the run's
pinned model and ignores `model` in the request body.

The pin must be a run-start snapshot, not a live read of the thread. Aspect ratio, resolution, and
duration belong to each individual generation; the model belongs to the run. Changing the thread's
[...]

PR #26807, opened 2026-08-13
feat(db): add video model columns and thread event kind

bingjiez666:
PR 1 of 8 for #26765 — a per-chat-thread built-in video model.

Schema and contract groundwork only. **No behavior change:** nothing writes or reads the new columns yet.

## What lands here

**Three nullable `selected_video_model` columns**, one per layer of the eventual resolution chain:

| Table | Role |
|---|---|
| `chat_threads` | the thread pin |
| `org_members_metadata` | the member default that seeds new chats |
| `zero_runs` | the run-start snapshot |

The run snapshot exists so that changing the thread pin while a run is in flight cannot change what that run generates. `zero_runs` already carries `selected_model` and `codex_service_tier` for the same reason.

**`chat_thread_events` mirrors the column** and the `chat_thread_event_kind` enum gains `video_model_updated`, so the setting rides the existing chat-thread event-sourcing pipeline (`chat_thread_events` → `chat_thread_snapshots` → `replayChatThreadEvents` → `ThreadMeta`) rather than a parallel mechanism. It gets its own event kind because — unlike Cloud Browser and Computer Use, which share one — it is not mutually exclusive with anything.

## Two decisions worth reviewing

**The wire field is `.optional()` and loosely typed** (`z.string().nullable().optional()`), following `cloudBrowserEnabled` and `selectedModel` rather than the stricter enum. The browser persists snapshots and events to IndexedDB and re-parses them with the *newer* bundle's schema, so a required field throws on rows an older bundle wrote. Loose typing likewise keeps a pin whose model later leaves the catalog parseable. The strict enum belongs on the write path, which arrives in PR 2.

**`applyEvent` is refactored into a field-patch table.** Adding a seventh inline `if (event.kind === ...)` branch pushed it to complexity 21 against a limit of 20. The branches now return the fields they patch, and the caller applies them uniformly; `sort_touched` stays separate because it moves a thread without counting as an update.

## The compaction footgun

The snapshot compaction projection is hand-written `jsonb_build_object` SQL, not ORM mapping. A column missing from it works through every read and then disappears once compaction runs — silently, with no error.

The existing snapshot test now pins a real video model on the thread before compacting and asserts it survives. Verified by deleting the projection line and watching the test fail:

```
-     "selectedVideoModel": "fal-ai/veo3.1/fast",
+     "selectedVideoModel": null,
```

An assertion of `selectedVideoModel: null` would *not* have caught this — the read path normalizes a missing key to `null`, so only a non-null value proves the round trip. That is what `setChatThreadVideoModelFixture` is for; there is no write endpoint until PR 2.

## Fallbacks

All three are `docs/fallback.md` §6.1 time-boxed cross-version rollout compatibility.

- `turbo/packages/api-contracts/src/contracts/chat-threads.ts` — `selectedVideoModel` on `chatThreadSnapshotProjectionSchema` and `chatThreadEventSchema` (`.nullable().optional()`): lets a new bundle parse a payload without the field, either from an API deployed before this change or from an IndexedDB row an older bundle wrote. Surfaces: DB/API up to ~102min **and** old web clients ~2d. Removable once the client version floor passes the build that introduced the field and cached rows have resynced, together with the two normalizations below. Follow-up: #26765.
- `turbo/apps/api/src/signals/services/zero-chat-thread-event.service.ts` — `getChatThreadSnapshot`, `thread.selectedVideoModel ?? null`: `chat_thread_snapshots.chat_threads` rows compacted before this migration carry no such key. Not a deploy window — bounded by `CHAT_THREAD_SNAPSHOT_STALE_MS` (24h) plus batch drain. Removable once every snapshot row has been recompacted. Follow-up: #26765.
- `turbo/packages/core/src/chat-thread-event-replay.ts` — `?? null` in the `created` branch, in `updatedThreadFields`, and in `replayChatThreadEvents` hydration: client-side normalization of the same optional wire field, so an older payload replays as an unset pin rather than `undefined`. Same surfaces and removal condition as the contract entry. Follow-up: #26765.

## Checks

- `db:migrate` applies cleanly; all four columns and the new enum value verified in Postgres
- `test:migration-consistency` — all validations pass
- `check-types` — `api`, `@vm0/app`, and all other workspaces clean
- `lint` — clean across the changed workspaces
- `knip` — exit 0
- Replay tests: 8 passed, including two new cases for the event kind and its deferred-before-creation path
- Snapshot compaction tests: 3 passed

The remaining 15 failures in `chat-threads.bdd.test.ts` are pre-existing in this sandbox — a missing queue service (`Job not found in queue`). Confirmed identical on clean `main`: 15 failed / 18 passed both with and without this branch.

## Next

PR 2 adds `POST /api/okou/chat-threads/:id/video-model` and the optimistic client write. Full breakdown in #26765.


bingjiez666:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 1ba3ffdedfb962f25fc496c63b1b6f0caac0410f

### Summary

Adds a nullable `selected_video_model` to `chat_threads`, `org_members_metadata`, and `zero_runs`, mirrors it on `chat_thread_events` with a new `video_model_updated` enum value, and threads the field through the snapshot projection, the event service, the wire contracts, and the client replay. No writer or reader exists yet — this is schema and contract groundwork for #26765.

The engineering is sound: the compaction projection is updated and guarded by a test that fails when the column is dropped, `applyEvent` is refactored to stay under the complexity limit without changing behavior, and the new wire fields are optional for the right reasons. The blocker is procedural: the diff introduces three cross-version rollout fallbacks and the PR summary declares none, which `docs/fallback.md` §9 makes a P1 that forces `Changes Requested` even when every fallback is itself justified.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | `feat:` commits must have integration tests | Pass | `chat-thread-event-replay.test.ts` adds two cases for the new kind; `chat-threads.bdd.test.ts` asserts the column survives compaction |
| `REVIEW.md` | Every fallback introduced by the diff must be declared in the PR summary | **Finding (P1)** | Three fallbacks introduced (see `Fallbacks`); PR body has no `Fallbacks` section |
| `REVIEW.md` | DB/JSONB review gate — old rows must be covered by migration, normalizer, or rationale | Pass | `ChatThreadSnapshotProjection` gains a required key while the Zod schema keeps it optional; `getChatThreadSnapshot` normalizes pre-migration jsonb rows with `?? null` |
| `REVIEW.md` | Feature regression guard — changed existing behavior needs coverage | Pass | `applyEvent` refactor preserves per-kind semantics including `sort_touched` not bumping `updatedAt`; existing replay suite covers all prior kinds |
| `REVIEW.md` | New user-facing features gated behind a `FeatureSwitchKey` | N/A | No user-facing behavior; nothing reads or writes the columns in this PR |
| `docs/fallback.md` | §6.1 time-boxed cross-version rollout compatibility is a permitted fallback | Pass | All three fallbacks protect genuinely reachable mixed-version states |
| `docs/fallback.md` | §7–8 a rollout fallback's comment must name surface, window, removal condition, and follow-up | **Finding (P1)** | `chat-threads.ts:150-154` explains *why* the field is optional but states no window, no removal condition, and no follow-up |
| `docs/fallback.md` | §3 no `??` chains on states the contract cannot produce | Pass | Each `?? null` guards a genuinely absent key: pre-migration jsonb rows, or an optional wire field |
| `docs/fallback.md` | §1 no negative tests asserting removed behavior stays removed | Pass | No tombstone tests added |
| `docs/deployment-compatibility.md` | Migrations must be backward-compatible with the currently deployed backend | Pass | Four additive nullable columns plus one additive enum value; old API issues explicit column lists and cannot see or require them |
| `docs/deployment-compatibility.md` | New code must not require the new column/enum before the migration lands | Pass | Nothing writes `video_model_updated` or the columns in this PR; PR 2 is stacked on this branch so it cannot deploy first |
| `docs/event-sourcing.md` | Persistent events are authoritative and reconcile optimistic events by shared event ID | Pass | New kind rides `appendChatThreadEvent` and the existing optimistic reconciliation; every optimistic literal carries the new field |
| `docs/event-sourcing.md` | Projections must apply each kind consistently | Pass | `updatedThreadFields` applies `updatedAt` uniformly and leaves `sort_touched` on the `sortAt`-only path |
| `docs/bad-smell.md` | §1–2 no `any`, no lint/type suppressions | Pass | No `any`, no `@ts-ignore`; complexity resolved by refactor rather than a disable comment |
| `docs/bad-smell.md` | §4 interface changes must stay coherent across producers and consumers | Pass | Field added to all six event-service projections, the compaction SQL, replay hydration, and every optimistic literal |
| `docs/bad-smell.md` | §7 fail fast rather than defaulting | Pass | No fabricated values; unset means unset |
| `docs/testing.md` | Integration tests only; no unit tests for internal functions | Pass | Replay tests exercise the exported `replayChatThreadEvents`; the DB assertion runs through the compaction cron route |
| `docs/testing.md` | Real database, no `globalThis.services.db` mocking | Pass | `setChatThreadVideoModelFixture` writes through the real `db()` |
| `docs/testing/api-testing.md` | Test at the route boundary, use fixtures for state the product cannot set yet | Pass | Fixture used only because no write endpoint exists until PR 2; assertion still reads through the compaction route |
| `docs/testing/anti-patterns.md` | AP-1/AP-5/AP-8 — no mock-call assertions, no fake timers, no over-testing | Pass | Assertions are on projected state; the file's existing `mockNow` usage is unchanged |
| `docs/testing/app-testing.md` | Platform tests go through mock infrastructure, not direct rendering | N/A | Only a pure-function replay test and its builders changed; no component or MSW surface touched |
| `.claude/skills/ccstate/SKILL.md` | Commands and computed keep their existing execution boundaries | Pass | Only literal object fields added inside existing commands; no new signal, cache, or lifecycle |
| `docs/cache.md` | RC-3/RC-5 — no unbounded caches, no duplicate sources of truth | Pass | No new state; the field flows through the existing snapshot/event pipeline only |
| `docs/effect.md` | Derived values stay in `computed`, domain events in commands | Pass | No new effect or command introduced |

### Findings

- **PR summary**: the diff introduces three cross-version rollout fallbacks (listed below) and the summary has no `Fallbacks` section. `docs/fallback.md` §9 makes an undeclared fallback a P1 on its own and blocks the verdict, independent of whether the fallback is correct. Add a `Fallbacks` section naming each one with its surface, window, removal condition, and follow-up.
- **`turbo/packages/api-contracts/src/contracts/chat-threads.ts:150-154`**: the comment explains why `selectedVideoModel` is optional but does not name the affected surface, its window from `docs/fallback.md` §7, the condition that makes the branch removable, or a follow-up. §8 requires all four; an open-ended "tolerate the old shape" comment is a finding. The same gap applies to the two `?? null` normalizations in `zero-chat-thread-event.service.ts` and `chat-thread-event-replay.ts`, which carry no comment at all.

#### Notes (non-blocking)

- `turbo/apps/api/src/test-fixtures/chat-thread-events.ts:157`: `selectedVideoModel` is typed `string` rather than the catalog id union, so the fixture accepts an id the column should never hold. Tightening it once PR 2 exports the enum would keep the fixture honest.
- `org_members_metadata.selected_video_model` and `zero_runs.selected_video_model` land with no writer until PR 3 and PR 4. The PR states the reason — one migration for the whole stack — which is the right trade against three separately-ordered migrations.

### Fallbacks

- `turbo/packages/api-contracts/src/contracts/chat-threads.ts` — `chatThreadSnapshotProjectionSchema.selectedVideoModel` and `chatThreadEventSchema.selectedVideoModel` (`.nullable().optional()`): protects a new web bundle parsing a snapshot or event payload that lacks the field, either from an API deployed before this change or from IndexedDB rows an older bundle wrote — surface DB/API up to ~102min **and** old web clients ~2d, removal condition after the client version floor passes the build that introduced the field and cached IndexedDB rows have resynced, with a follow-up PR that also deletes the two `?? null` normalizations, declared in PR summary **No — P1**, verdict Justified.
- `turbo/apps/api/src/signals/services/zero-chat-thread-event.service.ts` — `getChatThreadSnapshot`, `thread.selectedVideoModel ?? null`: protects `chat_thread_snapshots.chat_threads` jsonb rows compacted before this migration, which have no `selectedVideoModel` key — surface persisted jsonb rather than a deploy window, bounded by `CHAT_THREAD_SNAPSHOT_STALE_MS` (24h) plus batch drain, removal condition after every snapshot row has been recompacted, declared in PR summary **No — P1**, verdict Justified.
- `turbo/packages/core/src/chat-thread-event-replay.ts` — `event.selectedVideoModel ?? null` in the `created` branch and in `updatedThreadFields`, and `thread.selectedVideoModel ?? null` in `replayChatThreadEvents`: client-side normalization of the same optional wire field, so an older payload replays as an unset pin instead of `undefined` — surface DB/API up to ~102min and old web clients ~2d, removal condition same as the contract entry above, declared in PR summary **No — P1**, verdict Justified.

### Testing

- **Coverage: Adequate.** `feat:` commit with tests. `chat-thread-event-replay.test.ts` adds a case asserting `video_model_updated` sets the video pin without touching the run model, and a case asserting the kind is deferred and replayed when it arrives before `created` — the branch added to `isDeferrableUpdate`. `chat-threads.bdd.test.ts` pins a real model before compaction and asserts it survives; the PR reports this was verified by deleting the projection line and watching the assertion fail, which is the right way to prove a guard against a silent drop.
- **Conventions: Compliant.** No internal mocking, no fake timers, no mock-call assertions, real database throughout. `setChatThreadVideoModelFixture` follows the existing `setAgentRunCreatedAtFixture` shape and exists only because no write endpoint lands until PR 2; asserting a non-null value is necessary here, since the read path normalizes a missing key to `null` and a null-valued assertion would pass even with the column dropped from the projection.
- **Gap worth noting, not blocking:** nothing exercises the enum value against a real `chat_thread_events` insert in this PR. PR 2's route test covers it, and this branch is its base, so the sequencing holds.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: d16327854f70c797e401a8c5223e5d4a0ddeaeeb
PR_AUTO_REVIEW_VERDICT: LGTM

Re-review of `d163278`. Both P1s from the review of `1ba3ffd` are resolved; the rest of the diff is unchanged from that review.

### Summary

Adds a nullable `selected_video_model` to `chat_threads`, `org_members_metadata`, and `zero_runs`, mirrors it on `chat_thread_events` with a new `video_model_updated` enum value, and threads the field through the compaction projection, the event service, the wire contracts, and the client replay. No writer or reader exists yet — schema and contract groundwork for #26765.

Since the last review the branch fixed a genuine `test-app (4)` failure and closed both fallback-declaration findings.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | `feat:` commits must have integration tests | Pass | Two replay cases for the new kind; `chat-threads.bdd.test.ts` asserts the column survives compaction |
| `REVIEW.md` | Every fallback introduced by the diff must be declared in the PR summary | Pass | PR body now carries a `Fallbacks` section with all three, each naming surface, window, removal condition, and follow-up |
| `REVIEW.md` | DB/JSONB review gate — old rows covered by migration, normalizer, or rationale | Pass | `getChatThreadSnapshot` normalizes pre-migration jsonb rows; the Zod schema keeps the key optional |
| `REVIEW.md` | Feature regression guard — changed existing behavior needs coverage | Pass | `applyEvent` refactor preserves per-kind semantics; `chat-thread-event-sourcing-local-first.test.ts` now pins the projected shape including the new key |
| `REVIEW.md` | New user-facing features gated behind a `FeatureSwitchKey` | N/A | No user-facing behavior; nothing reads or writes the columns in this PR |
| `docs/fallback.md` | §6.1 time-boxed cross-version rollout compatibility is permitted | Pass | All three protect genuinely reachable mixed-version states |
| `docs/fallback.md` | §7–8 comment must name surface, window, removal condition, and follow-up | Pass | `chat-threads.ts:153-160`, `zero-chat-thread-event.service.ts:145-150`, and the `replayChatThreadEvents` doc comment each carry all four |
| `docs/fallback.md` | §3 no `??` chains on states the contract cannot produce | Pass | Each `?? null` guards a genuinely absent key |
| `docs/fallback.md` | §1 no negative tests asserting removed behavior stays removed | Pass | No tombstone tests added |
| `docs/deployment-compatibility.md` | Migrations backward-compatible with the deployed backend | Pass | Four additive nullable columns plus one additive enum value; old API issues explicit column lists |
| `docs/deployment-compatibility.md` | New code must not require the new column/enum before the migration lands | Pass | Nothing writes `video_model_updated` or the columns here; PR 2 is stacked on this branch and cannot deploy first |
| `docs/event-sourcing.md` | Persistent events authoritative, reconciled by shared event ID | Pass | New kind rides `appendChatThreadEvent` and existing reconciliation; every optimistic literal carries the field |
| `docs/event-sourcing.md` | Projections apply each kind consistently | Pass | `updatedThreadFields` applies `updatedAt` uniformly; `sort_touched` stays on the `sortAt`-only path |
| `docs/bad-smell.md` | §1–2 no `any`, no lint/type suppressions | Pass | Complexity resolved by refactor, not a disable comment |
| `docs/bad-smell.md` | §4 interface changes coherent across producers and consumers | Pass | Field added to all six event-service projections, the compaction SQL, replay hydration, and every optimistic literal |
| `docs/bad-smell.md` | §7 fail fast rather than defaulting | Pass | No fabricated values; unset means unset |
| `docs/testing.md` | Integration tests only; no unit tests for internal functions | Pass | Tests exercise exported `replayChatThreadEvents` and the compaction cron route |
| `docs/testing.md` | Real database, no `globalThis.services.db` mocking | Pass | `setChatThreadVideoModelFixture` writes through the real `db()` |
| `docs/testing/api-testing.md` | Test at the route boundary; fixtures only for state the product cannot set yet | Pass | Fixture exists only because no write endpoint lands until PR 2; assertion reads through the compaction route |
| `docs/testing/anti-patterns.md` | AP-1/AP-5/AP-8 — no mock-call assertions, no fake timers, no over-testing | Pass | Assertions are on projected state |
| `docs/testing/app-testing.md` | Platform tests go through mock infrastructure, not direct rendering | Pass | The updated local-first test drives the store through its existing MSW mocks; no component rendered |
| `.claude/skills/ccstate/SKILL.md` | Commands and computed keep existing execution boundaries | Pass | Only literal object fields added inside existing commands |
| `docs/cache.md` | RC-3/RC-5 — no unbounded caches, no duplicate sources of truth | Pass | No new state; the field flows through the existing pipeline |
| `docs/effect.md` | Derived values in `computed`, domain events in commands | Pass | No new effect or command |

### Findings

#### Resolved since `1ba3ffd`

- ~~PR summary missing a `Fallbacks` section~~ — added, with all three entries carrying surface, window, removal condition, and follow-up.
- ~~Fallback comments state no window or removal condition~~ — all three now name the surface, the §7 window, the removal condition, and #26765 as the follow-up. The snapshot-normalizer comment correctly identifies its bound as `CHAT_THREAD_SNAPSHOT_STALE_MS` plus batch drain rather than a deploy window, which is the accurate framing for a persisted-jsonb fallback.

#### Notes (non-blocking, unchanged)

- `turbo/apps/api/src/test-fixtures/chat-thread-events.ts:157`: `selectedVideoModel` is typed `string` rather than the catalog id union. Tightening it once PR 2 exports the enum would keep the fixture honest.
- `org_members_metadata.selected_video_model` and `zero_runs.selected_video_model` land with no writer until PR 3 and PR 4. The PR states the reason — one migration for the whole stack — which is the right trade.

### Fallbacks

- `turbo/packages/api-contracts/src/contracts/chat-threads.ts` — `selectedVideoModel` on `chatThreadSnapshotProjectionSchema` and `chatThreadEventSchema` (`.nullable().optional()`): protects a new bundle parsing a payload without the field, from an API deployed before this change or from an IndexedDB row an older bundle wrote — surface DB/API up to ~102min and old web clients ~2d, removal condition after the client floor passes the introducing build and cached rows resync, follow-up #26765, declared in PR summary **Yes**, verdict Justified.
- `turbo/apps/api/src/signals/services/zero-chat-thread-event.service.ts` — `getChatThreadSnapshot`, `thread.selectedVideoModel ?? null`: protects snapshot jsonb rows compacted before this migration — surface persisted jsonb bounded by `CHAT_THREAD_SNAPSHOT_STALE_MS` (24h) plus batch drain, removal condition after every snapshot row has been recompacted, follow-up #26765, declared in PR summary **Yes**, verdict Justified.
- `turbo/packages/core/src/chat-thread-event-replay.ts` — `?? null` in the `created` branch, `updatedThreadFields`, and `replayChatThreadEvents` hydration: client-side normalization of the same optional wire field — same surfaces and removal condition as the contract entry, follow-up #26765, declared in PR summary **Yes**, verdict Justified.

### Testing

- **Coverage: Adequate.** `feat:` commit with tests. `chat-thread-event-replay.test.ts` covers the new kind setting the video pin without touching the run model, and the deferred-before-`created` path added to `isDeferrableUpdate`. `chat-threads.bdd.test.ts` pins a real model before compaction and asserts it survives; the PR reports verifying this by deleting the projection line and watching the assertion fail.
- **Regression caught and fixed.** `test-app (4)` failed on `chat-thread-event-sourcing-local-first.test.ts`: two `toStrictEqual` expectations compare the projected thread object, which gained a key. Worth noting that the earlier switch from `.default(null)` to `.optional()` removed the compile-time signal here — the field stayed required on `EventDrivenChatThread`, so only the runtime strict-equality assertion caught it. The fix updates the two expectations and leaves the `threadMeta` assertion alone, which is correct since `ThreadMeta` does not gain the field until PR 2.
- **Conventions: Compliant.** No internal mocking, no fake timers, no mock-call assertions, real database throughout.

PR #26809, opened 2026-08-13
feat: serve generated images through the cdn image transform

bingjiez666:
## Summary

- return an `embedUrl` alongside the stored file URL from built-in image generation, pointing at the same R2 object through the Cloudflare Image Resizing transform
- tell the website generation prompt to embed that URL and to keep at most 3 image generations in flight
- put the actual concurrency limit into the `BUILT_IN_RUN_CONCURRENCY_LIMIT` message so callers stop guessing it

## Why

A controlled A/B on Website generation (two interleaved pairs, same template, brief, model and time window) showed the image stage accounting for the bulk of the regression when generation replaces stock-photo resolution. Two causes are addressable without touching the image model:

**PNG-only output.** `seedream4` declares `outputFormats: ["png"]`, so every generated website image is a lossless PNG and `--format jpeg` is rejected at the API. `r2ImageTransformUrl` already exists in `@vm0/core` and is used throughout the platform UI for thumbnails and previews, but nothing applied it to generated images. `PUBLIC_ARTIFACTS_BASE_URL` resolves to `cdn.vm7.io` / `cdn.vm0.io`, both of which are already in the transform host allowlist, so no storage move or allowlist change is needed. `format=auto` negotiates AVIF/WebP per request and `metadata=none` drops EXIF.

**Blind concurrency.** `RUN_BUILT_IN_MAX_IN_FLIGHT` is 3, but the 429 body only said "too many built-in generations in progress" without the number. One measured arm issued 71 calls for 30 images and took 40 HTTP 429s; at concurrency 3 the same workload retried zero times.

No width is requested yet — sizing per media slot needs the template slot geometry and is deliberately left out of this change.

## Behavior

| Surface | Before | After |
| --- | --- | --- |
| Image generate response | `url` only | `url` plus `embedUrl` through `/cdn-cgi/image/...` |
| `okou generate image` output | file URL | file URL, plus an `Embed this URL in HTML` line when it differs |
| Website prompt (API and CLI) | `seedream4` default only | also states the embed URL rule and the 3-in-flight limit |
| 429 concurrency body | limit unstated | states the limit and what to do |

`r2ImageTransformUrl` returns the input unchanged for a host outside the allowlist, so a non-CDN artifact base URL keeps today's behavior and the CLI prints no extra line.

## Verification

- `pnpm format`
- `NODE_OPTIONS=--max-old-space-size=3072 pnpm turbo run lint --concurrency=1 --log-order grouped` — 13/13 tasks, 0 errors
[...]

PR #26826, opened 2026-08-13
feat: reject unsupported image output formats in the cli

bingjiez666:
## Summary

- move the per-model image output-format table into `@vm0/core` as the single source of truth
- let `okou generate image` / `okou generate sprite` reject an unsupported `--format` locally instead of spending a round trip on the API's 400
- delete the duplicated `outputFormats` field from every `IMAGE_MODEL_CONFIGS` entry so the API reads the same table

Stacked on #26809 because both touch `image-generate.ts` and `zero-image-io-generate.service.ts`; base retargets to `main` once that merges.

## Why

`seedream4` declares `["png"]` and rejects `--format jpeg` with a 400, but nothing on the client side knows that: the CLI's help said only "Output format: png, webp, or jpeg". Every caller that wanted a compressed format learned it was unsupported by paying for a request first. Website generation defaults to `seedream4`, so this cost one wasted round trip per site.

The constraint already existed in `IMAGE_MODEL_CONFIGS.outputFormats`, but that table lives in the API service and is unreachable from the CLI. Moving it to `@vm0/core` lets both sides read one table rather than duplicating the constraint.

## Behavior

| Case | Before | After |
| --- | --- | --- |
| `--model seedream4 --format jpeg` | request sent, API 400 | fails locally: `seedream4 cannot output jpeg. Supported formats: png` |
| `--model qwen-image --format jpeg` | request sent | unchanged, request sent |
| `--model <unknown> --format jpeg` | request sent, API decides | unchanged, request sent |
| `--format` help | format list only | notes that support varies by model and that seedream4 is png only |

An unrecognized `--model` is deliberately not validated locally: model existence is the API's contract, and the CLI only checks the format/model combination for models present in the shared table.

The API keeps validating: `modelOutputFormats` throws if a configured alias is missing from the shared table, so a table/config mismatch fails fast instead of silently accepting every format.

## Verification

- `pnpm format`
- `NODE_OPTIONS=--max-old-space-size=3072 pnpm turbo run lint --concurrency=1 --log-order grouped` — 13/13 tasks, 0 errors
- `pnpm check-types` — 14/14 tasks
- `pnpm knip` — passed
- `apps/cli` `image.test.ts` — 21 passed, including two new cases: `seedream4 --format jpeg` fails without reaching the mocked endpoint, and `qwen-image --format jpeg` still sends the request
- `apps/api` `zero-image-io-generate.test.ts` — 20 passed, covering the unchanged server-side rejection after the table move


bingjiez666:
Closing without merging.

The original motivation was to stop paying a wasted round trip per site when the agent asked `seedream4` for `--format jpeg`. Two things changed that:

- #26809 serves generated images through the Cloudflare Image Resizing transform, so the browser negotiates AVIF/WebP regardless of the stored PNG. Getting a compressed output format out of the generator is no longer the point.
- Measured against the real cost, this was 1 rejected call out of 71 in the instrumented run. The 40 remaining failures were HTTP 429 from exceeding the 3-in-flight built-in generation limit, and that guidance already shipped in #26809.

What this PR still had going for it was deduplication: moving the per-model output-format table into `@okouai/core` and deleting the duplicated `outputFormats` field from all 11 `IMAGE_MODEL_CONFIGS` entries. That remains a reasonable cleanup, but it is not worth carrying a separate PR through the queue on its own right now.

The branch `feat/cli-validates-image-output-format` is left in place, so this can be reopened if the dedup is wanted later.

PR #26840, opened 2026-08-13
fix(api): preserve firewall execution metadata in run context

### LINKED ISSUE #26825: fix(api): preserve firewall execution metadata in run context snapshots
## Problem and evidence

The `run-context` Axiom snapshot is built from the same prepared
`storedContext` used for the runner job, but inline firewalls are rewritten
before ingestion:

- built-ins retain `{ kind: "builtin", name, baseUrlVars? }`;
- inline entries are reduced to `{ name, apis: [{ base, permissions? }] }`;
- inline `kind`, connector/API execution IDs, host policy, and all auth data are
  lost.

This makes the context view materially different from the firewall prepared for
the runner and prevents connector and inline model-provider auth diagnostics.
Current custom-connector and model-provider producers keep stored credential
values in the separate execution-secrets map and put `${{ secrets.NAME }}`
references in firewall auth. Connector definitions may also contain intentional
literal auth template text.

The storage format has a second constraint. #17220 records a production Axiom
1,024-field-limit incident caused by dynamic object keys in run-context rows.
Connector-defined `auth.headers` and `auth.query` keys cannot be written as raw
nested maps without recreating that failure class.

## Decision

Preserve the prepared firewall metadata in new run-context responses, including
auth and built-in `baseUrlVars`, through a backward-compatible, Axiom-stable
storage encoding.

Do not add firewall catalog identity or version metadata. Existing execution
fields such as `customConnectorId` and per-API `id` may be retained; they identify
the prepared inline execution entry and are not a catalog snapshot/version.

## Chosen direction

New inline Axiom entries will:

- retain top-level `name` and `apis`, so the previous API reader can continue
  returning its sanitized base/permission view during rollout or rollback;
- add `kind: "inline"`, optional execution IDs, host policy, permissions, and
  auth metadata;
- encode dynamic header/query records as fixed-shape `{ name, value }` entry
  arrays so connector-defined names remain values rather than Axiom fields.

The new reader will reconstruct an enriched top-level inline response:

```ts
{
  kind: "inline";
  name: string;
  customConnectorId?: string;
  apis: Array<{
    id?: string;
    base: string;
    hostPolicy?: FirewallBaseHostPolicy;
    auth: FirewallAuth;
    permissions?: FirewallPermission[];
  }>;
}
```

Keeping `name` and `apis` at the public top level also lets the previous web
client parse a new response through its sanitized inline schema. Historical
sanitized `{ name, apis }` rows remain readable and retain their historical
response shape. Compact built-in rows remain unchanged.

## Security boundary

- Preserve the auth strings that were prepared for execution, including secret
  references, auth base templates, fixed AWS SigV4 fields, and intentional
  connector-definition literals.
- Do not resolve secret references or read/decrypt credentials while building
  the snapshot.
- Integration tests must prove the actual custom-connector and model-provider
  credential values used by the run are absent from serialized Axiom events.
- Existing run-context endpoint authorization remains unchanged.

This change intentionally makes prepared auth template metadata visible to
authorized context viewers. It does not claim that arbitrary literal text in a
connector definition is secret-redacted.

## Alternatives considered

- **Raw execution entries:** structurally simplest, but dynamic auth map keys
  recreate the Axiom field-growth risk and previous API readers omit the inline
  envelope.
- **Encoded execution envelope:** bounds Axiom fields, but previous readers
  still omit inline firewalls during rollout/rollback.
- **Dual-write legacy and execution fields:** preserves both readers exactly,
  but duplicates the payload and adds precedence/cleanup debt when one
  compatible representation is sufficient.

## Risks and mitigations

- **Expanded diagnostic visibility:** auth references and definition literals
  become visible. Keep existing authorization and prove resolved stored secrets
  are not serialized.
- **Axiom schema growth:** encode header/query maps as entry arrays with fixed
  field names.
- **Rollback and old-client degradation:** keep top-level `name/apis`, allowing
  previous API readers and already-loaded web clients to preserve their current
  sanitized result.
- **Mixed historical/new response shapes:** retain the historical contract
  variant and update the known semantic consumer, connector diagnostics.
- **Malformed Axiom data:** normalize unknown rows before contract validation;
  fall back to valid historical base/permission data when an enriched row is
  only partially usable.

## Acceptance criteria

- A built-in firewall records its compact execution entry and `baseUrlVars`
  unchanged.
- A custom inline firewall response retains its execution entry, connector/API
  IDs, base, host policy, permissions, and auth templates.
- An inline model-provider firewall response retains its reference-based auth
  configuration.
- Axiom storage uses entry arrays, not dynamic nested maps, for header/query
  auth names.
- Serialized snapshot data contains the expected `${{ secrets.NAME }}`
  references but not the corresponding decrypted credential values.
- The run-context API validates and returns compact built-ins, new inline
  execution entries, and historical sanitized inline entries.
- A previous API reader can safely reduce a new inline stored row to its
  historical base/permission view.
- Connector diagnostics handle historical and new inline response variants
  equivalently.
- Relevant API contract, integration, type, lint, Knip, and normal build checks
  pass.

## Delivery and scope

This is one reviewable API/contract PR. It changes internal Axiom encoding and
the additive run-context response union; it does not change the runner queue or
claim protocol.

No historical backfill, claim-time network-policy change, catalog
reconstruction, catalog identity/version recording, environment-redaction
change, or Axiom delivery-guarantee change is included.

Relevant implementation areas:

- `turbo/apps/api/src/signals/services/agent-run-create.service.ts`
- `turbo/apps/api/src/signals/services/run-context-snapshot.service.ts`
- `turbo/apps/api/src/signals/services/connector-check.service.ts`
- `turbo/packages/api-contracts/src/contracts/zero-runs.ts`
- API route integration suites for run writes, reads, model-provider gateways,
  and connector diagnostics

Prior work: #17341 compacted built-in firewall snapshots while retaining the
sanitized inline debug shape.

seven332:
## Summary

- preserve prepared inline firewall execution metadata and unresolved auth references in run-context snapshots
- encode dynamic auth header/query keys as Axiom entry arrays while keeping the legacy top-level `name`/`apis` storage fallback
- return enriched inline firewalls at the compatible top level so previous API readers and already-loaded web clients degrade to the historical sanitized view
- keep historical, builtin, and connector diagnostic behavior compatible

## Security and compatibility

- credentials are never resolved into the snapshot; integration tests assert actual custom connector and model provider secret values are absent
- old and new API readers can consume each other's Axiom rows, and old and new web clients can consume each other's response shapes
- runner payloads, claim-time policy, and catalog identity/version remain unchanged

## Validation

- `pnpm format`
- `pnpm -F api lint`
- `pnpm check-types`
- focused API integration tests: 4 files, 196 tests passed
- API contract tests: 28 files, 543 tests passed
- `pnpm knip`
- `pnpm turbo run build --filter=api --filter=@okouai/app`

Closes #26825

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
@@ -10237,6 +10237,42 @@ describe("RUN-02: custom connectors, grants, and network policies", () => {
     expect(customApi.auth?.query?.scope).toBe(
       `\${{ secrets.${scopeVariableKey} }}`,
     );
+    const runContextSnapshot = runContextSnapshotForRun(run.runId);
+    expect(runContextSnapshot.firewalls).toStrictEqual(
+      expect.arrayContaining([
+        expect.objectContaining({
+          kind: "inline",
+          name: internalName,
+          customConnectorId: saved.connector.id,
+          apis: expect.arrayContaining([
+            expect.objectContaining({
+              base: `https://xn--mnich-kva.${rand}.test/v1/`,
+              auth: {
+                headerEntries: [
+                  {
+                    name: "Authorization",
+                    value: `Bearer \${{ secrets.${secretKey} }}`,
+                  },
+                ],
+                queryEntries: [
+                  {
+                    name: "tenant",
+                    value: `\${{ secrets.${variableKey} }}`,
+                  },
+                  {
+                    name: "scope",
+                    value: `\${{ secrets.${scopeVariableKey} }}`,
+                  },
+                ],
+              },
+            }),
+          ]),
+        }),
+      ]),
+    );
+    expect(JSON.stringify(runContextSnapshot)).not.toContain(
+      "runtime-proposal-secret",
+    );
 
     const authBody = {
       encryptedSecrets: fw.encryptedSecretsBody({}),
@@ -11074,6 +11110,11 @@ describe("RUN-02: custom connectors, grants, and network policies", () => {
       }),
     ).toContain("zendesk");
     expect(findFirewallEntry(claim.firewalls, "zendesk")).toStrictEqual({
+      kind: "builtin",
+      name: "zendesk",
+      baseUrlVars: { ZENDESK_SUBDOMAIN: "xn--mnich-kva" },
+    });
+    expect(runContextSnapshotForRun(run.runId).firewalls).toContainEqual({
       kind: "builtin",
       name: "zendesk",
       baseUrlVars: { ZENDESK_SUBDOMAIN: "xn--mnich-kva" },
--- a/turbo/apps/api/src/signals/routes/__tests__/run-reads.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/run-reads.bdd.test.ts
@@ -2470,13 +2470,68 @@ describe("RUN-04: agent run telemetry families", () => {
             ],
             firewalls: [
               {
+                kind: "inline",
                 name: "test-fw",
+                customConnectorId: "11111111-1111-4111-8111-111111111111",
                 apis: [
                   {
+                    id: "test-fw:0",
                     base: "https://api.example.com",
+                    hostPolicy: { kind: "publicDestination" },
+                    auth: {
+                      headerEntries: [
+                        {
+                          name: "Authorization",
+                          value: `Bearer \${{ secrets.TEST_TOKEN }}`,
+                        },
+                      ],
+                      base: "https://auth.example.com",
+                      queryEntries: [
+                        {
+                          name: "api_key",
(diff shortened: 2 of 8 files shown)


seven332:
## Code Review: PR #26840

**Reviewed head:** `7c8d1288fa29b5b582484c4555184c7ac4853ae3`

### Summary

The change preserves the launch-time firewall execution metadata needed to inspect a run while keeping resolved secret values out of Axiom. The final top-level inline response shape remains readable by the previous browser contract, and the new API continues to accept historical sanitized rows. Built-in `baseUrlVars` are retained; catalog identity/version are intentionally excluded.

### Findings

No findings.

### Testing

- Formatting, API/API-contract lint, repository and focused type checks, and Knip passed.
- Focused API integration tests passed: 196 tests across 4 files.
- API-contract tests passed: 543 tests across 28 files.
- Normal API and app production builds passed.
- Codecov patch coverage is 95.45% against an 89.10% target; the check is green.

### Verdict

LGTM

PR #26831, opened 2026-08-13
feat(api): add per-thread video model endpoint

bingjiez666:
PR 2 of 8 for #26765. **Stacked on #26807** — base is `feat/video-model-columns`, so review that first and this diff shows only the endpoint.

## What lands here

`POST /api/okou/chat-threads/:id/video-model` → 204. Body is `{ model, eventId? }`, where a null `model` clears the pin so resolution later falls through to the member and system defaults.

The handler updates `chat_threads.selected_video_model` and appends one `video_model_updated` event in the same transaction, then publishes `threadListChanged` — the same shape as the model-selection route.

Nothing reads the pin yet. Resolution (PR 4) and endpoint enforcement (PR 5) come later; this only establishes the write path.

## Why a separate route

It shares none of model-selection's resolution: no provider routing, no service tier, no `org_model_policies` row, no plan filtering. Every catalog model is selectable by every workspace, so the request enum over `VIDEO_MODEL_IDS` is the whole validation. Folding this into the existing route would mean threading a second, unrelated resolution path through it.

Same reason it gets its own event kind: Cloud Browser and Computer Use share one only because they are mutually exclusive, which video model is not.

## Two deviations from the plan in #26765

**`patchChatThreadVideoModel$` moved to PR 6.** I wrote it here, but it has no consumer until the picker lands and `knip` correctly failed on the unused export. PR 2 is now API-only. `ThreadMeta.selectedVideoModel` stays, since `chatThreadMetaMap$` populates it.

**Create-time `videoModel?` moved to PR 7.** No caller sends it until the new-chat path carries a value, so landing the contract field now would be speculative.

Both are noted on the issue.

## Testing

New route test, 7 cases:

- pins a model, and the event carries it
- clears the pin with `model: null`
- leaves `selected_model` untouched while setting `selected_video_model`
- a caller-supplied `eventId` replayed twice appends one event, not two
- a model outside the catalog is rejected 400 and the row is unchanged
- another user's thread is 404 and that row is unchanged
- a ZERO_TOKEN without `chat-thread:write` is 403

Every assertion reads back through a production endpoint. The pin is observed on the `video_model_updated` event from `chatThreadsContract.events`, and the untouched run model through `chatThreadMetadataContract.get`. The out-of-catalog case is sent as raw JSON via `setupRawAppRequest`, since the contract types `model` as the catalog enum and that request is only reachable from a non-TypeScript caller.

## Checks

- Route test: 7 passed
[...]

PR #26841, opened 2026-08-13
feat(api): add member default video model

bingjiez666:
PR 3 of 8 for #26765. Base is `main` (#26807 landed). Independent of #26831; both branch from PR 1 and can be reviewed in either order.

## What lands here

The member-level default that will seed new chats: `selectedVideoModel` on the existing user-model-preference resource, plus a `defaultVideoModel` realtime kind so a change propagates to the user's other sessions.

Extending the existing resource rather than adding a second one keeps one preference read, one write, one cache, and one push channel.

## The coupling this had to avoid

`updateUserModelPreference$` used to treat `selectedModel: null` as "clear everything": it ran a separate `.update()` and returned a hardcoded `{ null, null, null }`. Dropping a third field into that shape would mean clearing your run model silently discarded a video default sent in the same request.

Rather than add a third branch, the write is now a single upsert. A pure `userModelPreferenceColumns` maps the request to the columns it owns — the run model and its tier together, the video default independently — and one `insert ... onConflictDoUpdate` applies them. One field clearing another is no longer a bug you can fix; it is a shape the code cannot express. It also removes the non-atomicity of the old two-write path.

Behavior difference worth naming: the clear path used `.update()`, which was a no-op when no row existed, and the unified upsert now creates the row. This is benign — the created row takes the column defaults, and `userPreferences()` already returns exactly those same values when the row is missing.

There is a test for the coupling: set both, clear only the run model, assert the video default survives.

## Optional in both directions

Request `selectedVideoModel` is optional and checked with an own-property test, so absent means "leave it alone" and null clears it. This is permanent partial-update semantics, not a rollout fallback: a caller that only changes the run model must never blank the video default. An older bundle that sends only `{ selectedModel, serviceTier }` keeping its stored default falls out of the same rule.

Response `selectedVideoModel` is optional as a rollout fallback — see below.

A stored id that has since left the catalog reads as unset rather than throwing, via a new `isVideoModelId` guard mirroring `isSupportedRunModel`. The column is a plain `varchar(255)`, so a persisted id can outlive the model it names.

## Fallbacks

- `zero-user-model-preference.ts:userModelPreferenceResponseSchema.selectedVideoModel` — optional so a newly promoted bundle can still parse a response from an API that predates the field. App promotion follows API promotion, so the reachable direction is an API rollback. Surface: frontend -> backend, sized by the API rollback window. Remove — make it required — once the pre-field API is outside that window; follow-up tracked on #26765.
- `realtime.ts:userPreferenceChangedPayloadSchema` — unknown `kinds` are dropped instead of failing the payload. A strict enum made the list unextendable: a tab running a bundle from before a kind was added rejected the whole push and stopped honoring the kinds it did understand. Surface: old web clients, ~2 days. Permanent by design — this is what makes every future kind addition safe, not a branch waiting to be deleted. It does not retroactively fix bundles already in browsers, so `defaultVideoModel` still needs the field's own writer to ship after this lands.
- `zero-user-data.service.ts:userModelPreference` — `isVideoModelId` projects a persisted id outside the catalog as unset. Surface: persisted state; the column is not enum-constrained, so this is permanent and has no removal condition.

## Note for whoever merges second

This defines its video-model enum locally in the preference contract; #26831 defines an equivalent one locally in `chat-threads.ts`. Both branch from PR 1, so neither could share the other's. Worth collapsing to the single `isVideoModelId`/enum in `video-models.ts` once both are in — not urgent, and deliberately not done across branches.

## Testing

Added to `zero-model-policies.test.ts`, which already owns the preference-route coverage and its fixtures:

- stores a video default, then clearing the run model leaves it intact, then clearing it explicitly works
- omitting the field keeps the stored value
- a model outside the catalog is rejected 400
- the video kind reaches the published payload only when the request carries the field

Added to `chat-composer-model-selection-and-providers.test.tsx`:

- a push carrying only `defaultVideoModel` still reloads the whole preference resource

Updated `user-config.bdd.test.ts`, which pins the full response with `toStrictEqual`.

## Checks

- `zero-model-policies.test.ts`: 37 passed
- `user-config.bdd.test.ts`: 15 passed
- `chat-composer-model-selection-and-providers.test.tsx`: 52 passed
- `lint`: 0 errors; `check-types`: clean; `knip`: clean


bingjiez666:
Changes Requested


PR_AUTO_REVIEW_COMMIT: be9fff72aedead2a1c54f2fb5fa119cbf28c9d07

### Summary

Adds `selectedVideoModel` to the existing user-model-preference resource as the member-level default, plus a `defaultVideoModel` realtime kind. `updateUserModelPreference$` is refactored from a "clear-and-early-return" branch into a single `insert ... onConflictDoUpdate` driven by a pure `userModelPreferenceColumns` mapper, and a new `isVideoModelId` guard projects a persisted id that has left the catalog as unset.

The refactor itself is the strongest part of the change: it matches the conditional-spread upsert idiom already used by `updateUserPreferences$` in the same file, and it removes the two-write non-atomicity of the old path. The blockers below are about undeclared/unbounded compatibility behavior, one suppression comment, and coverage for the two behaviors this PR actually adds.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no wrong assumptions or unhandled edge cases | Pass | Single upsert makes "clearing the run model blanks the video default" structurally impossible; `userModelPreferenceColumns` maps each preference to its own columns. Response is read back through `userModelPreference()`, so the clear path no longer returns a hardcoded shape. |
| `REVIEW.md` | Behavior change: clear path now creates a row instead of no-op `.update()` | Pass | Verified benign: `userPreferences()` no-row defaults (`zero-user-data.service.ts:122-133`) — `timezone: null`, `locale: null`, `pinnedAgentIds: []`, `sendMode: "enter"`, `morningBriefEnabled: false`, `captureNetworkBodiesRemaining: 0` — exactly match the column defaults in `org-members-metadata.ts`. A created row reads identically to a missing row. |
| `REVIEW.md` | Type safety; no `any`, no overly broad types | Pass | `Partial<typeof orgMembersMetadata.$inferInsert>` return type; `isVideoModelId` is a proper type guard narrowing to `VideoModelId`. |
| `REVIEW.md` | Security: auth checks, no injection, no hardcoded secrets | N/A | No new endpoint, no new auth surface; the route keeps its existing `auth.userId`/`orgId` scoping. |
| `REVIEW.md` | Feature Regression Guard — existing flow most likely to regress is covered | Finding (P1) | The realtime reload path is the existing behavior this PR touches (`user-model-preference.ts:78`), and no test exercises the new kind on either side. |
| `REVIEW.md` | New user-facing features gated behind a `FeatureSwitchKey` | N/A | No UI ships here: no platform caller sends `selectedVideoModel` (`zero-chat-composer.tsx` still posts `{selectedModel, serviceTier}`). The gating obligation lands on the PR that adds the picker. |
| `REVIEW.md` | DB/JSONB review gate — persisted contract changes need migration or read-time normalizer | Pass | No schema change here; `selected_video_model varchar(255)` already exists on `main` via #26807. The read-time normalizer for that unconstrained column is `isVideoModelId`. |
| `REVIEW.md` | `feat:` commits must have integration tests | Finding (P1) | Storage behavior is covered by 3 new API integration tests; the new realtime publish and the new platform reload branch are not. |
| `docs/bad-smell.md` | Zero tolerance for suppression comments (`@ts-expect-error` explicitly listed) | Finding (P1) | `zero-model-policies.test.ts` adds the only `@ts-expect-error` in the repo (repo-wide count across `turbo/apps` + `turbo/packages` is currently 0). |
| `docs/bad-smell.md` | No `any`; no dynamic `import()`; no hardcoded URLs/config | Pass | None present in the diff. |
| `docs/bad-smell.md` | Interface changes documented; breaking changes highlighted | Finding (P1) | The PR description documents a design that is not in the diff (see P1-4). |
| `docs/fallback.md` §6 | A fallback is kept only when it is rollout compatibility, optional external data, a genuinely nullable column, or a presentational default | Pass | All three compatibility paths fall under §6.1/§6.3 — see `### Fallbacks`. None guards a state the contract already prevents. |
| `docs/fallback.md` §7-8 | A rollout fallback names its surface, its window, a removal condition, and a follow-up | Finding (P1) | Both contract comments name the surface and the reason but no window, no removal condition, no follow-up. |
| `docs/fallback.md` §9 | Every fallback introduced by the diff is declared in a `Fallbacks` section of the PR summary | Finding (P1) | PR body sections are: What lands here / The coupling this had to avoid / Optional in both directions / Note for whoever merges second / Testing / Checks. No `Fallbacks` section. |
| `docs/fallback.md` §1 | No negative tests asserting removed behavior stays removed | Pass | The 3 new tests assert live behavior (store, preserve-on-clear, omit-preserves, out-of-catalog rejection). No tombstones. |
| `docs/deployment-compatibility.md` | Additive contract changes; old client -> new backend must stay safe | Finding (P1) | The `kinds` enum widening is not additive for an already-deployed reader — see P1-1. |
| `docs/deployment-compatibility.md` | New code before migration / old code after migration | N/A | No migration in this PR; the column landed with #26807, so both directions are already settled. |
| `docs/testing.md` | Integration tests only; mock at the boundary; real database; no fake timers | Pass | Tests go through `setupApp()` + the ts-rest contract with the real DB; no `vi.mock()` of internal paths, no `vi.stubGlobal("fetch")`, no fake timers. |
| `docs/testing.md` | Deployment-compatibility tests cover the previous request shape | Pass | "keeps the stored video default when an older bundle omits it" covers the old-request-shape direction. |
| `docs/testing/api-testing.md` | Route behavior covered through `setupApp()` + contract, not by importing handlers/services | Pass | All 3 tests use `setupApp({context, routes: zeroUserModelPreferenceRoutes})(zeroUserModelPreferenceContract)` with `accept()` narrowing and API-call verification. |
| `docs/testing/api-testing.md` | Persisted side effects verified through follow-up API calls | Pass | The omit-compat test re-reads via `preferenceClient.get()` rather than querying the DB. |
| `docs/testing/anti-patterns.md` | AP-1 no asserting on mock calls; AP-4 no mocking internal code; AP-11 test the route, not the service | Pass | Assertions are on response bodies; `userModelPreferenceColumns` is exercised through the route, not called directly. |
| `docs/testing/patterns.md` | MSW used for HTTP mocking at the boundary | Pass | Platform mock updated in `src/mocks/handlers/api-user-model-preference.ts`, matching the MSW handler pattern. |
| `docs/testing/testing-external-behavior.md` | Assert externally observable behavior, not internals | Pass | Every new assertion is on an HTTP response body. |
| `.claude/skills/ccstate/SKILL.md` | Commands express business semantics; computed memoization not hand-rolled; no `void` on promises | Pass | Change stays inside the existing `handleUserPreferenceChanged$` command and the existing `userModelPreference` computed; no new signal, cache, or floating promise. |
| `docs/effect.md` | Keep cause and effect together in a semantic command; no new React effect for domain events | Pass | The reload stays in `handleUserPreferenceChanged$`, driven by the existing Ably payload loop. |
| `docs/cache.md` | RC-2/RC-3: no render-time signal allocation, no unbounded lifetime cache | Pass | `VIDEO_MODEL_ID_SET` is a module-level `ReadonlySet` over a fixed const array — bounded and immutable, not a growing cache. |

### Findings

**P1-1 — `turbo/packages/api-contracts/src/contracts/realtime.ts:32`: widening `userPreferenceKinds` makes every push carrying the new kind unparseable to already-deployed clients, and the kind is a no-op today.**

`userPreferenceChangedPayloadSchema` is `z.object({ kinds: z.array(z.enum(userPreferenceKinds)) })`. A client bundle built before this PR only knows `["defaultModel"]`, so `z.enum` rejects the `defaultVideoModel` element, the whole object fails to parse, and `payloadRequestsKindsReloadFor` (`user-model-preference.ts:58-68`) returns `false` on the parse failure. The old tab therefore drops the entire event — including the `defaultModel` reload that ships in the same payload, since `zero-user-model-preference.ts:92` always publishes `defaultModel` alongside it. Window: old web/app clients, ~2 days (`docs/fallback.md` §7).

This is currently latent, not live: no platform caller sends `selectedVideoModel` yet, so the API never emits the new kind in production from this PR alone. It becomes reachable the moment a later PR in the stack adds the writer — which, given the merge cadence of this 8-PR stack, will very likely be inside the 2-day client window.

Worth noting that the kind buys nothing today: `defaultModel` is published unconditionally on every preference update, and `handleUserPreferenceChanged$` maps both kinds to the same `reloadUserModelPreference$`. So the new member is presently pure cost.

Either drop `defaultVideoModel` until a consumer needs a distinct reload, or keep it and make the reader tolerant of unknown kinds (filter unrecognized values instead of rejecting the payload) while stating in a comment that emission must wait out the ~2-day client window — a tolerant reader shipped today does not retroactively fix bundles already in browsers.

**P1-2 — PR summary is missing the required `Fallbacks` section.**

The diff introduces three compatibility paths (listed in `### Fallbacks` below). `docs/fallback.md` §9 makes an introduced-but-undeclared fallback a P1 that blocks the verdict on its own, independent of whether the fallback is correct. The "Optional in both directions" prose explains two of them well, but §9 requires the dedicated section with file/symbol, surface, window, removal condition, and follow-up. Please add it.

**P1-3 — `zero-user-model-preference.ts:14-20,34-40`: the two rollout fallbacks carry no window, removal condition, or follow-up.**

Per `docs/fallback.md` §8, a time-boxed rollout fallback ships with its removal already planned. Both comments name the surface and the reason but stop there, which is how a compatibility branch becomes the "old shape tolerated forever" case in §5. Add the window from §7 and the removal condition to each comment plus a follow-up issue.

One of the two may not be a fallback at all: if absent-means-unchanged on `selectedVideoModel` is intended as permanent partial-update semantics for this resource rather than old-bundle compatibility, say that in the comment instead — then it needs no window and no removal condition. The current comment picks the compatibility framing, which is what makes §8 apply.

**P1-4 — PR description documents a design the diff does not contain.**

"The coupling this had to avoid" describes the video default being "applied **before** the run-model branch", and says "the early return now reads the row back instead of returning `{ null, null, null }`". There is no branch and no early return in the diff — `updateUserModelPreference$` is a single upsert built from `userModelPreferenceColumns`, which is a better design than what the description claims. Please update the body so it matches the code the reviewer is reading.

**P1-5 — `zero-model-policies.test.ts`: `@ts-expect-error` is a prohibited suppression comment.**

```ts
// @ts-expect-error -- the contract rejects ids outside the catalog.
selectedVideoModel: DEFAULT_ORG_MODEL_POLICY_DEFAULT_MODEL,
```

`docs/bad-smell.md` §2 lists `// @ts-expect-error` under "Zero tolerance for suppression comments". This would be the only occurrence in `turbo/apps` and `turbo/packages` — the current repo-wide count is 0, so the convention is being held strictly today.

The test itself is valuable; only the suppression needs to go. A direct assertion on a typed local avoids it, e.g. `const outsideCatalog = DEFAULT_ORG_MODEL_POLICY_DEFAULT_MODEL as VideoModelId;` and then `selectedVideoModel: outsideCatalog` — no suppression comment, no `any`.

**P1-6 — No coverage for either behavior this PR adds beyond storage.**

Both have an established pattern in the repo, so this is cheap to close:

- API: nothing asserts that `defaultVideoModel` reaches the published payload. `context.mocks.ably.publish` assertions are already used this way (`chat-events.bdd.test.ts:1352`).
- Platform: `handleUserPreferenceChanged$` gained a branch, but no test triggers `kinds: ["defaultVideoModel"]`. The existing `kinds: ["defaultModel"]` trigger sits at `chat-composer-model-selection-and-providers.test.tsx:375`.

The platform gap matters most: per `REVIEW.md`'s Feature Regression Guard, the realtime reload is exactly the existing flow this PR is most likely to regress.

### Fallbacks

- `turbo/packages/api-contracts/src/contracts/zero-user-model-preference.ts:userModelPreferenceResponseSchema.selectedVideoModel` — optional response field so a newly promoted app can still parse a response from an API that predates it. Surface: frontend -> backend, and since app promotion follows API promotion, the reachable direction is an API rollback. Window: the API rollback window; not stated in code. Removal condition: not stated. Follow-up: none. Declared in PR summary: **No — P1**. Verdict: **justified** under §6.1, **not correctly time-boxed** under §8.
- `turbo/packages/api-contracts/src/contracts/zero-user-model-preference.ts:updateUserModelPreferenceRequestSchema.selectedVideoModel` with the own-property check in `zero-user-data.service.ts:userModelPreferenceColumns` — an older bundle that sends only `{selectedModel, serviceTier}` keeps its stored video default instead of clearing it. Surface: old web/app clients, ~2 days (§7). Removal condition: not stated. Follow-up: none. Declared in PR summary: **No — P1**. Verdict: **justified** under §6.1, **not correctly time-boxed** under §8 — or reframe as permanent partial-update semantics (see P1-3).
- `turbo/apps/api/src/signals/services/zero-user-data.service.ts:userModelPreference` — `isVideoModelId` projects a persisted id that has left the catalog as unset rather than throwing. Surface: persisted state; the column is `varchar(255)` with no enum constraint, so a stored id can legitimately outlive the model it names. Window: none — permanent by design, and correctly so. Removal condition: none. Declared in PR summary: **No** (mentioned in prose, not in a `Fallbacks` section). Verdict: **justified** under §6.3; this one should be recorded but not time-boxed.

### Testing

- Coverage: **Insufficient** — the storage contract is well covered (3 new integration tests: store-and-preserve-across-run-model-clear, explicit clear, older-bundle omission, out-of-catalog 400, including the cross-version request-shape case `docs/testing.md` asks for). Missing: the `defaultVideoModel` realtime publish on the API side and the new reload branch on the platform side (P1-6).
- Conventions: **One violation** — `@ts-expect-error` (P1-5). Otherwise compliant: route integration tests through `setupApp()` + contract, real database, boundary-only mocking, no internal `vi.mock()`, no direct `fetch` stubbing, no fake timers, assertions on behavior rather than mock calls.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 7c356570782a860b8ff96ae23973064315d2bf39
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Re-review at `7c35657`. Adds `selectedVideoModel` to the user-model-preference resource as the member-level default, plus a `defaultVideoModel` realtime kind, and refactors `updateUserModelPreference$` from a clear-and-early-return branch into a single `insert ... onConflictDoUpdate` driven by a pure `userModelPreferenceColumns` mapper.

Disclosure: the two commits since `be9fff7` (`782416b`, `7c35657`) were authored by this pr-auto run to resolve the six P1 findings from the [previous review](https://github.com/vm0-ai/vm0/pull/26841#issuecomment-5277754453). They are reviewed here on the same criteria as the rest of the diff, not waved through.

All six prior findings are resolved:

| Prior finding | Resolution |
| --- | --- |
| P1-1 `kinds` enum widening breaks old clients' payload parse | `userPreferenceChangedPayloadSchema` now filters unknown kinds instead of rejecting the payload; residual recorded below |
| P1-2 missing `Fallbacks` section | Added, with all three entries |
| P1-3 rollout fallbacks lack window/removal condition | Response field carries surface + window + removal condition + follow-up; request field reframed as permanent partial-update semantics |
| P1-4 PR body describes a design not in the diff | Body rewritten to the single-upsert design actually implemented |
| P1-5 `@ts-expect-error` | Replaced with a typed cast |
| P1-6 no coverage for the publish or the platform reload | One test each |

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic | Pass | Single upsert makes "clearing the run model blanks the video default" unrepresentable. Read-back through `userModelPreference()` keeps the response consistent with what was written. |
| `REVIEW.md` | Behavior change: clear path now creates a row | Pass | Verified benign: `userPreferences()` no-row defaults match the column defaults in `org-members-metadata.ts` field for field, so a created row reads identically to a missing one. |
| `REVIEW.md` | Type safety; no `any` | Pass | `Partial<typeof orgMembersMetadata.$inferInsert>`; `isVideoModelId` and `isUserPreferenceKind` are real type guards. The one `as unknown as VideoModelId` is in a test whose whole purpose is to submit a value the contract rejects — the invalid state cannot be constructed without it. |
| `REVIEW.md` | Security | N/A | No new endpoint or auth surface; existing `auth.userId`/`orgId` scoping unchanged. |
| `REVIEW.md` | Feature Regression Guard | Pass | The realtime reload was the flow most at risk. It is now covered on both sides, and the payload reader no longer regresses for clients that predate a kind. |
| `REVIEW.md` | Feature switch gating | N/A | No UI ships here; no caller sends the field yet. The obligation lands on the picker PR. |
| `REVIEW.md` | DB/JSONB gate | Pass | No schema change; `selected_video_model varchar(255)` already on `main` via #26807. `isVideoModelId` is the read-time normalizer for that unconstrained column. |
| `REVIEW.md` | `feat:` needs integration tests | Pass | Six behaviors covered end-to-end through routes and the rendered composer (see Testing). |
| `docs/bad-smell.md` | Zero tolerance for suppression comments | Pass | The `@ts-expect-error` is gone; repo-wide count is back to 0. |
| `docs/bad-smell.md` | No `any`, no dynamic `import()`, no hardcoded config | Pass | None present. |
| `docs/bad-smell.md` | Interface changes documented | Pass | PR body now matches the implemented design. |
| `docs/fallback.md` §6 | Fallbacks limited to the permitted categories | Pass | Three paths, all within §6.1/§6.3 — see `### Fallbacks`. |
| `docs/fallback.md` §7-8 | Rollout fallback names surface, window, removal condition, follow-up | Pass | The response field carries all four in its comment. The other two are permanent by design and say so rather than pretending to a window. |
| `docs/fallback.md` §9 | Every introduced fallback declared in the PR summary | Pass | `## Fallbacks` lists all three in the required shape. |
| `docs/fallback.md` §1 | No tombstone tests | Pass | Every new assertion targets live behavior. |
| `docs/deployment-compatibility.md` | Additive changes; old client -> new backend stays safe | Pass with a documented residual | The reader is now forward-compatible. It cannot repair bundles already in browsers, so emitting `defaultVideoModel` is only fully safe once this deploy is ~2 days old — recorded in `### Fallbacks` and in the PR body. |
| `docs/deployment-compatibility.md` | Migration direction safety | N/A | No migration in this PR. |
| `docs/testing.md` | Integration only, boundary mocking, real DB, no fake timers | Pass | Route tests via `setupApp()` + contract against the real database; platform test drives the rendered page through MSW; no internal `vi.mock()`, no `fetch` stubbing, no fake timers. |
| `docs/testing.md` | Cross-version test coverage | Pass | "older bundle omits the field" and "push carries only the video kind" both exercise mixed-version shapes. |
| `docs/testing/api-testing.md` | Route behavior through the contract; side effects verified via follow-up API calls | Pass | Including the new publish test, which asserts the payload at the Ably boundary after `flushWaitUntilForTest()` — the established pattern from `chat-events.bdd.test.ts`. |
| `docs/testing/anti-patterns.md` | AP-1 not asserting on mock calls; AP-4 no internal mocks; AP-11 test the route | Pass | The publish assertion checks the emitted payload's contents, not that a function was called — the payload is the externally observable contract. `userModelPreferenceColumns` is exercised through the route. |
| `docs/testing/patterns.md` / `testing-external-behavior.md` | Assert externally observable behavior | Pass | The platform test asserts the reloaded resource and the composer label, not internal signal wiring. |
| `.claude/skills/ccstate/SKILL.md` | Semantic commands, computed memoization, no floating promises | Pass | No new signal or command; the change stays inside the existing `handleUserPreferenceChanged$` and `userModelPreference` computed. |
| `docs/effect.md` | Cause and effect in a semantic command | Pass | Reload stays in the existing Ably payload-loop command. |
| `docs/cache.md` | RC-2/RC-3 no render-time allocation, no unbounded cache | Pass | `VIDEO_MODEL_ID_SET` and `USER_PREFERENCE_KIND_SET` are module-level `ReadonlySet`s over fixed const arrays. |

### Findings

No P0 or P1 findings.

One note, not blocking, carried forward from P1-1 for the PR that adds the picker: the tolerant reader shipped here protects every *future* kind addition, but it cannot help a bundle that a browser loaded before this deploy. Those tabs still drop a push whose `kinds` contains `defaultVideoModel`, and with it the `defaultModel` reload in the same payload. The consequence is bounded — a stale run-model default in a tab that has been open a while, recovered by the reconnect catch-up or a refresh — and no production client sends `selectedVideoModel` yet, so nothing emits the new kind today. The picker PR should either land after this deploy is outside the ~2 day web-client window or accept that degradation knowingly.

### Fallbacks

- `zero-user-model-preference.ts:userModelPreferenceResponseSchema.selectedVideoModel` — optional response field so a newly promoted bundle can parse a response from an API that predates it. Surface: frontend -> backend; app promotion follows API promotion, so the reachable direction is an API rollback, sized by the API rollback window. Removal condition: make it required once the pre-field API is outside that window; follow-up on #26765. Declared in PR summary: **Yes**. Verdict: **justified** (§6.1) and correctly time-boxed (§8).
- `realtime.ts:userPreferenceChangedPayloadSchema` — unknown `kinds` are dropped rather than failing the payload. Surface: old web clients, ~2 days (§7). Removal condition: none — permanent by design; this is the mechanism that makes future kind additions safe, not a branch awaiting deletion. Declared in PR summary: **Yes**. Verdict: **justified** (§6.1).
- `zero-user-data.service.ts:userModelPreference` — `isVideoModelId` projects a persisted id outside the catalog as unset. Surface: persisted state; the column is `varchar(255)` with no enum constraint, so a stored id can legitimately outlive the model it names. Removal condition: none — permanent. Declared in PR summary: **Yes**. Verdict: **justified** (§6.3).

### Testing

- Coverage: **Adequate.** `zero-model-policies.test.ts` (37 passed) covers store-and-preserve across a run-model clear, explicit clear, older-bundle omission, out-of-catalog 400, and the publish payload carrying the video kind only when the request does. `chat-composer-model-selection-and-providers.test.tsx` (52 passed) covers a push carrying only `defaultVideoModel` still reloading the resource. `user-config.bdd.test.ts` (15 passed) updated for the widened response.
- Conventions: **Compliant.** No suppressions, no internal mocks, real database, MSW at the boundary, assertions on observable behavior.

PR #26855, opened 2026-08-13
perf(mitm-addon): bound path validation work

### LINKED ISSUE #26832: perf(python): bound nested percent-decoding path work
## Problem and evidence

`has_unsafe_path()` performs synchronous, request-controlled work in the compiled-firewall request path. A segment containing nested percent escapes can force all five policy decode passes plus the final fail-closed decode. Each pass repeats Python-level percent validation, unsafe-syntax checks, Unicode compatibility normalization, decoding, and allocation over most of the intermediate segment.

At audited commit `b036c02de532c72c678d636dfc47885bf025e0c1`:

- [`path_security.py:47-64`](https://github.com/vm0-ai/vm0/blob/b036c02de532c72c678d636dfc47885bf025e0c1/crates/runner/mitm-addon/src/path_security.py#L47-L64) performs five full syntax/decode passes, then a sixth decode to fail closed when the value is still changing.
- [`path_security.py:67-84`](https://github.com/vm0-ai/vm0/blob/b036c02de532c72c678d636dfc47885bf025e0c1/crates/runner/mitm-addon/src/path_security.py#L67-L84) scans and normalizes the full intermediate segment on each pass.
- [`path_security.py:91-103`](https://github.com/vm0-ai/vm0/blob/b036c02de532c72c678d636dfc47885bf025e0c1/crates/runner/mitm-addon/src/path_security.py#L91-L103) validates every escape in Python before `unquote_to_bytes(...).decode(...)` scans and allocates again.
- [`matching.py:1737-1750`](https://github.com/vm0-ai/vm0/blob/b036c02de532c72c678d636dfc47885bf025e0c1/crates/runner/mitm-addon/src/matching.py#L1737-L1750) calls the helper synchronously after a compiled base match.
- [`request_classification.py:446-456`](https://github.com/vm0-ai/vm0/blob/b036c02de532c72c678d636dfc47885bf025e0c1/crates/runner/mitm-addon/src/request_classification.py#L446-L456) reaches that matcher from request classification, which is invoked directly by the async request hook.

The decode-pass count is fixed, so the helper's work is linear rather than algorithmically unbounded. The security and availability problem is that request-target length is not generally bounded before this synchronous validation. The runtime locks mitmproxy 12.2.3, whose HTTP/1 reader buffers until the header terminator and explicitly leaves header-size denial-of-service protection as a FIXME. Existing 64 KiB request-target limits apply only after a request has matched selected authentication modes and therefore do not bound this earlier firewall scan.

The following deterministic input forces every pass while remaining safe to construct locally:

```python
unit = "%2e"
for _ in range(5):
    unit = urllib.parse.quote(unit, safe="")
path = "/" + unit * repetitions
has_unsafe_path(path)
```

Median process CPU time on Python 3.12.3/aarch64:

| Path length | CPU time per call |
| ---: | ---: |
| 16,381 characters | 7.475 ms |
| 65,534 characters | 29.707 ms |
| 262,133 characters | 118.801 ms |
| 1,048,568 characters | 480.392 ms |

The 1 MiB case also reached approximately 8.0 MiB of transient traced allocations inside the helper. Repeating such requests serializes nearly half a CPU-second of addon work per request before rejection and delays unrelated event-loop work.

## Decision

Proceed with planning a hard, deterministic work boundary for request-controlled path safety validation. A request that exceeds the supported validation budget must fail closed before the repeated percent-decoding pipeline runs. Compatibility is required for ordinary safe percent-encoded paths within that boundary; preserving behavior for arbitrarily large paths would conflict with the availability requirement.

Planning must choose and justify the boundary and enforcement layer using protocol compatibility, current request-target limits, failure behavior, observability, and maintenance cost. The issue remains a single independently reviewable delivery unless research shows that enforcement must be split across protocol and addon layers.

## Scope distinction

- #17198 prevents repeated path-safety scans across overlapping firewall candidates but leaves one full scan per request.
- #24254 fast-paths printable ASCII paths without `%`; nested percent-encoded inputs deliberately take the fallback and are unaffected by that optimization.
- Existing auth-specific request-target limits run after firewall matching and do not protect this path.
- Open issue searches for nested percent decoding, path-safety performance, and unsafe-path performance found no current duplicate.

## Risks and constraints

- A boundary changes behavior for oversized paths, so it must be defined in bytes or characters consistently with the request representation and return a deterministic fail-closed result.
- The guard must not weaken rejection of malformed escapes, invalid UTF-8, encoded separators or backslashes, dot segments, matrix-parameter dot segments, Unicode compatibility forms, or encoding still changing beyond the five-pass policy.
- Tests must prove the work boundary structurally rather than with timing thresholds.
- A diagnostic benchmark should remain outside the pass/fail test contract and compare adversarial and common inputs.

## Acceptance criteria

- Request-controlled path safety validation has a documented, deterministic work boundary independent of upstream request-line limits.
- Inputs beyond the supported validation boundary fail closed before repeated full-segment decode and normalization passes.
- Malformed escapes, invalid UTF-8, encoded separators/backslashes, dot segments, matrix-parameter dot segments, Unicode compatibility forms, and encoding that is still changing past the five-pass policy remain fail closed.
- Ordinary safe percent-encoded paths within the supported boundary retain their current matching behavior.
- Focused tests cover both sides of the work boundary and maximum-depth nested encoding without wall-clock assertions.
- A before/after diagnostic benchmark demonstrates bounded adversarial work and checks that common paths do not regress materially.

seven332:
## Summary

- reject URL paths above a 64 KiB character budget before nested percent decoding and Unicode normalization
- preserve existing path compatibility at and below the boundary, and reuse the existing `unsafe_path` firewall result above it
- keep the SigV4 raw request-target boundary test focused on the full target and remove an oversized connector-diagnostic path case that is no longer reachable

## Behavior and compatibility

- Paths up to and including 65,536 characters keep the existing validation behavior.
- Longer paths fail closed after a firewall base matches; query characters do not count toward this path budget.
- The shared helper also applies the budget to configuration, template, and rewrite validation callers.
- The budget is character-based so checking it is constant-time and does not allocate an encoded copy first.
- This bounds the expensive compatibility pipeline, not all request processing: URL parsing still scales with the request target.

## Performance evidence

Local `uv` environment, median of 7 calls over repeated nested percent encoding:

| Path characters | Base | This PR |
| ---: | ---: | ---: |
| 16,381 | 2.3452 ms | 2.3059 ms |
| 65,534 | 9.1298 ms | 9.0563 ms |
| 262,133 | 37.2852 ms | 0.0001 ms |
| 1,048,568 | 154.8572 ms | 0.0001 ms |

Sample common-path helper calls remained within 80 ns / 2.4% of the base measurement. The end-to-end matcher still spent about 1.66 ms parsing a 1 MiB request target before rejecting it.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (`5362 passed`)

## Out of scope

- a global request-target or protocol-level size limit
- a new firewall block reason or response status
- a permanent benchmark framework

Closes #26832

### DIFF
--- a/crates/runner/mitm-addon/src/path_security.py
+++ b/crates/runner/mitm-addon/src/path_security.py
@@ -10,10 +10,15 @@ _DOT_SEGMENTS = {".", ".."}
 _HEX_DIGITS = frozenset("0123456789abcdefABCDEF")
 _MAX_PERCENT_DECODE_PASSES = 5
 _PERCENT_ESCAPE_LENGTH = 3
+# Match the addon's existing 64 KiB request-work scale. Character count bounds
+# Python string processing without first scanning and allocating encoded bytes.
+MAX_PATH_VALIDATION_CHARACTERS = 64 * 1024
 
 
 def has_unsafe_path(path: str) -> bool:
-    """Return True when a URL path contains or may hide unsafe path syntax."""
+    """Return True when a URL path exceeds the validation budget or may hide unsafe syntax."""
+    if len(path) > MAX_PATH_VALIDATION_CHARACTERS:
+        return True
     if "\\" in path:
         return True
     raw_segments = path.split("/")
--- a/crates/runner/mitm-addon/tests/test_compiled_firewall_unknown_policy.py
+++ b/crates/runner/mitm-addon/tests/test_compiled_firewall_unknown_policy.py
@@ -214,7 +214,7 @@ def test_compiled_blocks_path_beyond_validation_budget():
         ],
         name="example",
     )
-    nested_encoding = "%25252525252e"
+    nested_encoding = "%2525252561"
     path = "/" + nested_encoding
     path += "a" * (path_security.MAX_PATH_VALIDATION_CHARACTERS + 1 - len(path))
     policies = {"example": {"allow": [], "deny": [], "unknownPolicy": "allow"}}
--- a/crates/runner/mitm-addon/tests/test_compiled_firewall_unknown_policy.py
+++ b/crates/runner/mitm-addon/tests/test_compiled_firewall_unknown_policy.py
@@ -3,6 +3,7 @@
 import pytest
 
 import matching
+import path_security
 from tests.firewall_helpers import compile_firewalls_or_fail, wrap_firewalls
 
 
@@ -200,6 +201,33 @@ def test_compiled_blocks_unsafe_path(url):
     assert isinstance(compiled, matching.FirewallBlock)
     assert compiled.reason == "unsafe_path"
     assert compiled.permissions == ()
+
+
+def test_compiled_blocks_path_beyond_validation_budget():
+    fws = wrap_firewalls(
+        [
+            {
+                "base": "https://api.example.com",
+                "auth": {"headers": {"Authorization": "Bearer token"}},
+                "permissions": [],
+            }
+        ],
+        name="example",
+    )
+    nested_encoding = "%25252525252e"
+    path = "/" + nested_encoding
+    path += "a" * (path_security.MAX_PATH_VALIDATION_CHARACTERS + 1 - len(path))
+    policies = {"example": {"allow": [], "deny": [], "unknownPolicy": "allow"}}
+
+    result = matching.match_compiled_firewall_request(
+        "https://api.example.com" + path,
+        "GET",
+        compile_firewalls_or_fail(fws),
+        policies,
+    )
+
+    assert isinstance(result, matching.FirewallBlock)
+    assert result.reason == "unsafe_path"
 
 
 def test_compiled_allows_encoded_backslash_in_query():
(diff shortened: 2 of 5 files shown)


seven332:
## Code Review: PR #26855

**Reviewed head:** `1ef9ac1ecb8467e72cad58d967c17d57bdc7361d`

### Summary

The shared path-safety helper now fails closed above a documented 64 KiB character budget before the repeated percent-decode/NFKC pipeline. In-bound behavior remains unchanged, and runtime matching reuses the existing `unsafe_path` block after a base match.

### Findings

No unresolved findings. Self-review found that the original matcher regression fixture was already unsafe under the base implementation; `1ef9ac1ecb` corrected it to a nested safe encoding, so the test now isolates the new length-bound behavior.

### Testing

Ruff format/check, Basedpyright, focused path/matcher tests, and the full addon suite passed (`5362 passed`). Same-process checks confirmed in-bound semantic parity, over-bound fail-closed behavior, and elimination of the helper's approximately 8 MiB transient allocation on a 1 MiB adversarial path. GitHub's addon, security, runner, and repository gates are green on this head. The latest default-branch addon changes were also checked and do not overlap this PR's files.

### Verdict

LGTM

PR #8898, opened 2026-04-10
fix(runner): drop discover future before shutdown to prevent mutex deadlock

seven332:
## Summary

- Drop `discover_fut` before the shutdown sequence in `start.rs` to release the discovery Mutex
- Prevents deadlock where `provider.shutdown()` waits forever to acquire a Mutex held by the still-alive `discover_fut`

## Root Cause

When the main loop breaks on `RunnerMode::Draining`, the pinned `discover_fut` remains in scope holding the discovery Mutex. The cancellation token is cancelled by the signal handler, but `discover_fut` is never polled again to observe it — so it never returns `None` and never releases the lock. `provider.shutdown()` then deadlocks trying to acquire the same Mutex.

## Fix

Explicitly `drop(discover_fut)` at the start of the shutdown sequence, before idle pool drain and `provider.shutdown()`.

Closes #8890

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #8890, opened 2026-04-10
bug(runner): shutdown deadlocks on discovery mutex


seven332:
## Problem

`runner service stop` hangs indefinitely when no jobs are running. `systemctl stop` sends SIGTERM, the runner enters draining mode, but never completes shutdown — eventually killed by SIGKILL after `TimeoutStopSec=300`.

## Root Cause

Deadlock between `discover_fut` and `provider.shutdown()`:

1. `discover()` acquires `self.discovery.lock().await` (Mutex) at entry and holds it for its entire loop
2. Main loop breaks on `RunnerMode::Draining`, but `discover_fut` (a `Box::pin` future) is **not dropped** — it stays alive in scope, still holding the Mutex guard
3. Shutdown path calls `provider.shutdown().await` which tries to acquire the same Mutex → **deadlock**

```rust
// line 504: created before the loop
let mut discover_fut = Box::pin(provider.discover());

// loop breaks on Draining...

// line 755: tries to lock the same mutex discover_fut holds
provider.shutdown().await;  // ← deadlocks here
```

The comment on `shutdown()` documents the ordering requirement but it's violated:

```rust
/// # Ordering requirement
///
/// `discover()` holds the discovery Mutex for its entire loop.
/// Callers must cancel the `CancellationToken` *before* calling
/// `shutdown()` so that `discover()` observes the cancellation,
/// returns `None`, and releases the lock.
```

The cancel token IS cancelled (by the signal handler), and `discover()` does have a `cancel.cancelled()` select branch. But `discover_fut` was already borrowed by the main loop's `select!` when `mode_rx.changed()` won — so `discover()` never got polled again to observe the cancellation.

## Fix

Drop `discover_fut` before calling `provider.shutdown()`:

```rust
drop(discover_fut);  // release discovery Mutex
provider.shutdown().await;
```

## Verification

Debug logging confirmed the deadlock:

```
shutdown: draining idle pool
shutdown: idle pool drained
shutdown: provider shutdown
# ← hangs here forever, no "provider shutdown done"
```

## File

`crates/runner/src/cmd/start.rs` — between main loop break and `provider.shutdown()`

PR #26867, opened 2026-08-13
docs(runner): correct job provider shutdown lifecycle contract

### LINKED ISSUE #26859: docs(rust): correct JobProvider shutdown lifecycle contract
## Summary

The `JobProvider::shutdown` rustdoc says the method is called only after `discover()` returns `None`, but the runner uses it as the once-only discovery-resource cleanup hook across startup failure and reactor teardown paths.

The documented lifecycle omits two valid discovery states: startup can fail before discovery begins, and the reactor can leave `Running` while the pinned discovery future is still pending. In the latter case, teardown explicitly drops the future before invoking `shutdown()`. Normal teardown also sends a final heartbeat after `shutdown()` and can continue receiving concurrent completion calls while active jobs drain.

This is a documentation-only mismatch in the provider lifecycle contract.

## Evidence

- `crates/runner/src/provider/mod.rs:782-786` documents `shutdown()` as being called once after `discover()` returns `None` and before in-flight jobs drain.
- `crates/runner/src/cmd/start/mod.rs:243-265` calls provider shutdown if publishing the live-runner instance fails. The provider is constructed earlier in startup, but no discovery future has been created yet.
- `crates/runner/src/cmd/start/mod.rs:1363-1421` calls provider shutdown when provider readiness or sandbox-factory startup fails, also before the discovery future is created.
- `crates/runner/src/cmd/start/mod.rs:1526-1530` pins one `discover()` future so unrelated `select!` branches do not cancel and recreate it.
- `crates/runner/src/cmd/start/mod.rs:1555-1571` exits the reactor directly when lifecycle mode is `Stopping` or `Stopped`; this path does not require the discovery future to have returned `None`.
- `crates/runner/src/cmd/start/mod.rs:1951-1991` drops the still-owned discovery future, drains the idle pool, calls `provider.shutdown()`, and then sends a final `Stopping` heartbeat through the same provider.
- `crates/runner/src/cmd/start/mod.rs:1993-2034` waits for active jobs only after provider shutdown and the final heartbeat, so spawned executor tasks can still call `complete()` during that drain.
- `ApiProvider::shutdown` and `LocalProvider::shutdown` release provider-owned background resources at `crates/runner/src/provider/api.rs:716-723` and `crates/runner/src/provider/local/mod.rs:256-258`, so implementers need the actual lifecycle and post-shutdown call ordering rather than the narrower documented precondition.
- Issue #8890, merged PR #8898, and `crates/runner/src/cmd/start/tests/main_loop/shutdown.rs` preserve the deadlock regression in which a pending discovery future retained provider state needed by `shutdown()`.

## Impact

A new or modified provider can incorrectly assume that discovery always starts and then observes a shutdown signal before `shutdown()` begins. It can also assume that provider I/O is finished when `shutdown()` returns. Those assumptions can make startup cleanup depend on state that was never initialized, deadlock discovery teardown, or prematurely close transport still needed by the final heartbeat and in-flight completions.

## Acceptance criteria

- Update the `JobProvider::shutdown` rustdoc in `crates/runner/src/provider/mod.rs` to describe it as the once-only cleanup hook that stops discovery and releases discovery-owned resources after the runner stops accepting new work.
- Document all valid discovery states at entry: discovery may never have started, may have returned `None`, or may still be pending when a lifecycle transition ends the reactor.
- Document that any owned discovery future is dropped before provider shutdown so provider-local borrows and guards are released.
- Document the normal-teardown ordering: provider shutdown happens before the final `Stopping` heartbeat and before active jobs finish draining, so `heartbeat()` and concurrent `complete()` calls may still arrive afterward.
- Keep the trait documentation consistent with the startup-cleanup and reactor-teardown call sites, and reference the relevant regression coverage or historical deadlock context.
- No runtime behavior change is required.

seven332:
## Summary

- document every valid discovery state when `JobProvider::shutdown` runs, including startup cleanup and a pending discovery future
- clarify that callers drop pending discovery before shutdown to preserve the #8890 / #8898 deadlock fix
- record that normal teardown can still call `heartbeat()` and `complete()` after shutdown

## Scope

Documentation only. This PR does not change provider APIs, runtime teardown ordering, implementations, or tests.

## Validation

- `cd crates && cargo doc -p runner --no-deps`

Closes #26859

### DIFF
--- a/crates/runner/src/provider/mod.rs
+++ b/crates/runner/src/provider/mod.rs
@@ -779,10 +779,20 @@ pub trait JobProvider: Send + Sync {
     /// Default no-op — only relevant for API-backed providers.
     async fn defer_poll_until(&self, _deadline: Instant) {}
 
-    /// Release discovery resources (subscriptions, background tasks).
+    /// Stop provider-owned discovery work and release its resources.
     ///
-    /// Called once after `discover()` returns `None` and before draining
-    /// in-flight jobs. `complete()` calls may still arrive after this.
+    /// Called once when the runner will perform no further discovery.
+    /// [`discover()`](JobProvider::discover) may never have started, may have
+    /// returned `None`, or may still be pending when a lifecycle transition
+    /// ends the reactor. In the last case, callers drop the pending future
+    /// before invoking this method so provider-local borrows and guards are
+    /// released first. This ordering preserves the fix for the shutdown
+    /// deadlock reported in #8890 (PR #8898).
+    ///
+    /// During normal reactor teardown, the runner sends a final
+    /// [`heartbeat()`](JobProvider::heartbeat) after this method returns and
+    /// then drains in-flight jobs. Concurrent
+    /// [`complete()`](JobProvider::complete) calls may therefore still arrive.
     async fn shutdown(&self);
 }
 


seven332:
## Code Review: PR #26867

**Reviewed head:** `399d0df981c2ca29abfbf089b154e13bfcd4f307`

### Summary

The method-local rustdoc now matches every verified `JobProvider::shutdown` path: startup cleanup before discovery, discovery returning `None`, and reactor teardown that drops a pending discovery future. It also correctly records the later normal-path heartbeat and completion calls without changing runtime behavior.

### Findings

No findings.

### Testing

- Existing `shutdown_completes_without_deadlock` integration coverage protects the pending-discovery ordering.
- Passed `cd crates && cargo fmt --all -- --check`.
- Passed `cd crates && cargo doc -p runner --no-deps`.
- Full workspace clippy and affected-crate tests remain delegated to PR CI.

### Verdict

LGTM

PR #26871, opened 2026-08-13
test(mitm-addon): cover non-string inline custom connector ids

### LINKED ISSUE #26863: test(python): cover non-string inline custom connector IDs
## Problem

`crates/runner/mitm-addon/src/registry_firewalls.py:27-38` explicitly rejects a non-string inline `customConnectorId` before passing the value to `uuid.UUID`, but that type-guard branch is not covered. The existing regression at `crates/runner/mitm-addon/tests/test_registry_builtin_catalog_resolution.py:548-565` covers only an invalid string UUID.

A mutation probe that removed the `isinstance(..., str)` guard left all 28 related registry catalog/inline-firewall tests green. With an integer `customConnectorId`, `uuid.UUID(1)` raises `AttributeError`; `crates/runner/mitm-addon/src/registry.py:293-301` catches only `FirewallEntryResolutionError`, so the exception escapes the per-VM rejection path and can abort loading a registry that also contains valid VMs.

## Acceptance criteria

- Add an end-to-end registry regression with one inline firewall whose `customConnectorId` is a non-string value and one valid peer VM.
- Assert that the registry remains available, the malformed VM is classified as `invalid_firewalls`, and the valid VM remains usable.
- Assert the stable validation message so the test fails if the type guard is removed or bypassed.

seven332:
## Summary

- add an end-to-end registry regression for a non-string inline `customConnectorId`
- verify the malformed VM is isolated as `invalid_firewalls` while a valid peer keeps a compiled firewall context
- pin the stable validation message and make the regression sensitive to removal of the explicit string guard

## Scope

- test-only change; no production behavior, API, schema, dependency, migration, or rollout changes

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync basedpyright -p .` — 0 errors
- `uv run --no-sync python -m pytest tests/test_registry_inline_firewalls.py -v` — 5 passed
- `uv run --no-sync python -m pytest tests/test_registry_builtin_catalog_resolution.py -v` — 24 passed
- `uv run --no-sync python -m pytest tests/` — 5361 passed
- controlled mutation probe: temporarily removing the string guard made the new test fail at registry loading with `AttributeError`; restoring the guard made it pass

Closes #26863

### DIFF
--- a/crates/runner/mitm-addon/tests/test_registry_inline_firewalls.py
+++ b/crates/runner/mitm-addon/tests/test_registry_inline_firewalls.py
@@ -141,3 +141,33 @@ class TestRegistryInlineFirewalls:
         assert valid_context is not None
         _, compiled_firewalls, _ = valid_context
         assert compiled_firewalls is not None
+
+    def test_non_string_inline_custom_connector_id_rejects_only_affected_vm(
+        self, tmp_path, mitm_ctx
+    ):
+        path = tmp_path / "registry.json"
+        malformed_vm = inline_vm("run-malformed")
+        malformed_vm["firewalls"][0]["customConnectorId"] = 1
+        write_multi_vm_registry(
+            path,
+            {
+                "10.200.0.1": malformed_vm,
+                "10.200.0.2": inline_vm("run-valid"),
+            },
+        )
+
+        with mitm_ctx():
+            state = registry.load_registry_state(str(path))
+            malformed_context = registry.get_vm_context("10.200.0.1", str(path))
+            valid_context = registry.get_vm_context("10.200.0.2", str(path))
+
+        assert not isinstance(state, registry.RegistryUnavailable)
+        assert set(state.vms) == {"10.200.0.2"}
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #26871

**Reviewed head:** `f5deac5df221f4eb5a5dada17a0fa6e9406c1417`

### Summary

Adds a registry-level regression proving that a non-string inline `customConnectorId` rejects only the malformed VM while a valid peer remains usable. The test is placed in the behavior-owning inline-registry module and is sensitive to removal of the explicit type guard.

### Findings

No findings.

### Testing

Coverage is adequate and follows the mitm-addon integration-testing guidance: real registry loading and context lookup use a temporary registry file, while only the mitmproxy context boundary is stubbed. Ruff format/lint, basedpyright, both focused registry modules, the full 5361-test mitm-addon suite, and a controlled mutation probe passed as expected. Unrelated Turbo, Rust, and E2E suites were not run locally; CI owns those remaining checks.

### Verdict

LGTM

PR #26884, opened 2026-08-13
feat(api): resolve and record the run video model

bingjiez666:
PR4 of the video-model chain for #26765. Rebased onto current main, which now contains #26831 and #26841.

## What this does

Adds `turbo/apps/api/src/signals/services/zero-video-model.service.ts` with `resolveVideoModelForRun()`:

```
chat_threads.selected_video_model
  ?? org_members_metadata.selected_video_model
  ?? DEFAULT_VIDEO_MODEL
```

The result is written to `zeroRuns.selectedVideoModel` in `launchZeroRunValues`, which is the single funnel both run inserts go through (the direct insert and the atomic-launch CTE). Resolution happens once in `prepareRunContext`, alongside the other per-run resolutions.

## Why snapshot instead of reading live

Changing a thread's video model must not affect a run that is already in flight — only later ones. Resolving at dispatch and freezing the value on the run row is what gives that. If the generation endpoint instead looked the thread up from `runId`, a mid-run re-pin would leak into the running run. `zero_runs` already stores `selected_model` and `codex_service_tier` this way; the new column sits next to them.

## Catalog validation

Both stored layers are checked against the catalog, and a model id the catalog no longer lists is treated as unset so resolution falls through to the next layer rather than throwing. The persisted values are projected out of jsonb without being re-validated against the current catalog, so a retired id can outlive its catalog entry, and a run must not fail on a value the user can no longer reach.

## Write-only on purpose

This PR does not touch `zero-video-io-generate.ts`. Endpoint enforcement is PR5. Splitting them lets the snapshot run for a while so the resolved values can be confirmed before anything depends on them, and lets PR5 be reverted without unwinding the snapshot.

## Threadless runs

`zero_runs.chat_thread_id` is null for non-chat triggers such as telegram. Those runs skip the thread layer and resolve to the member or catalog default instead of erroring — covered by its own test, which asserts the null `chat_thread_id` rather than trusting the entry point.

## Schema convergence

#26831 and #26841 each defined their own `z.enum(VIDEO_MODEL_IDS)` because they branched from a common base and could not reach each other's copy. A third inline copy already sat in the video generation options. Now that both have landed, all three share one `videoModelIdSchema` exported from `video-models.ts` — the contract that already owns the id list and the `isVideoModelId` guard. No behavior change; the enum is identical in all three places.

## Tests

Integration tests at the route boundary against a real database, in `chat-events.bdd.test.ts`:

- each of the three layers wins in turn (catalog default → member default → thread pin)
- **re-pinning the thread while a run is in flight leaves that run's snapshot alone, and the next run picks the new pin up** — this is the property the whole design exists for
- a retired thread pin falls through to the member default, and retired values on both layers fall through to the catalog default
- runs with a null `chat_thread_id` snapshot the catalog default, then the member default

New fixtures in `test-fixtures/run-video-model.ts` set the member default and read the snapshot back. The thread pin uses the existing `setChatThreadVideoModelFixture` from #26807.

Verified the tests are load-bearing by mutation: dropping the `selectedVideoModel` write fails all three; dropping the catalog-membership check fails the fall-through test; dropping the thread-pin layer fails the ordering test.

## One thing worth a reviewer's opinion

**The test fixtures read and write DB rows directly**, which `docs/testing/api-testing.md` says to raise during review rather than do silently — so, raising it. `zero_runs.selected_video_model` has no reader until PR5, so the snapshot cannot be asserted through any API surface today. There is precedent already on main in `setChatThreadVideoModelFixture`. These fixtures should be deleted once PR5 exposes the value.

## Local checks

- repo-wide `check-types` (14 workspaces) and `lint` (13 workspaces, 0 errors), `knip` (exit 0), `prettier --check` — all clean
- Full `api` vitest suite across 6 shards — ~2950 tests, all green

## Sequencing

Not for auto-merge — awaiting human review.

Cross-chain constraint for later: **PR5 must not merge before PR6.** PR5 removes `--model` from the template prompt and PR6's picker is the replacement entry point; merging them out of order leaves a window with no entry point.

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts
@@ -54,6 +54,7 @@ import { seedOrgMetadata } from "../../../test-fixtures/system-config-seeds";
 import { upsertOrgPlanEntitlementFixture } from "../../../test-fixtures/org-plan-entitlement";
 import { setChatThreadVideoModelFixture } from "../../../test-fixtures/chat-thread-events";
 import {
+  readRunChatThreadIdFixture,
   readRunVideoModelFixture,
   setOrgMemberVideoModelFixture,
 } from "../../../test-fixtures/run-video-model";
@@ -10336,7 +10337,29 @@ describe("CHAT-02: run video model snapshot", () => {
     await expect(readRunVideoModelFixture(threadPinned.runId)).resolves.toBe(
       "fal-ai/veo3.1/fast",
     );
+
+    // Re-pinning while that run is still in flight must not reach it. This is
+    // the whole reason the model is snapshotted onto the run instead of being
+    // read back off the thread when generation happens.
+    await setChatThreadVideoModelFixture(
+      unpinned.threadId,
+      "dreamina-seedance-2-5-260628",
+    );
+    await expect(readRunVideoModelFixture(threadPinned.runId)).resolves.toBe(
+      "fal-ai/veo3.1/fast",
+    );
     await cancelChatRun(actor, threadPinned.runId);
+
+    // The next run does pick the re-pinned model up.
+    const rePinned = await sendChatRun(actor, {
+      agentId,
+      threadId: unpinned.threadId,
+      prompt: "the run after the re-pin uses the new thread pin",
+    });
+    await expect(readRunVideoModelFixture(rePinned.runId)).resolves.toBe(
+      "dreamina-seedance-2-5-260628",
+    );
+    await cancelChatRun(actor, rePinned.runId);
   }, 90_000);
 
   it("keeps falling back past video models the catalog no longer lists", async () => {
@@ -10402,8 +10425,12 @@ describe("CHAT-02: run video model snapshot", () => {
       modelProvider: "anthropic-api-key",
     });
     await expect(
+      readRunChatThreadIdFixture(withoutPreference.runId),
+    ).resolves.toBeNull();
+    await expect(
       readRunVideoModelFixture(withoutPreference.runId),
     ).resolves.toBe(DEFAULT_VIDEO_MODEL);
+    await cancelChatRun(actor, withoutPreference.runId);
 
     await setOrgMemberVideoModelFixture({
       orgId,
@@ -10416,7 +10443,11 @@ describe("CHAT-02: run video model snapshot", () => {
       modelProvider: "anthropic-api-key",
     });
     await expect(
+      readRunChatThreadIdFixture(withMemberDefault.runId),
+    ).resolves.toBeNull();
+    await expect(
       readRunVideoModelFixture(withMemberDefault.runId),
     ).resolves.toBe("MiniMax-H3");
+    await cancelChatRun(actor, withMemberDefault.runId);
   }, 90_000);
 });
--- a/turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts
@@ -12,6 +12,7 @@ import {
 import { replayChatThreadEvents } from "@okouai/core/chat-thread-event-replay";
 import { avatarTemplateStylePresetId } from "@okouai/core/avatar-template";
 import { FeatureSwitchKey } from "@okouai/core/feature-switch-key";
+import { DEFAULT_VIDEO_MODEL } from "@okouai/core/video-model-catalog";
 import {
   chatEventsContract,
   chatThreadsContract,
@@ -51,6 +52,11 @@ import { server } from "../../../mocks/server";
 import { backdateRunStartedAtFixture } from "../../../test-fixtures/agent-runs";
 import { seedOrgMetadata } from "../../../test-fixtures/system-config-seeds";
 import { upsertOrgPlanEntitlementFixture } from "../../../test-fixtures/org-plan-entitlement";
+import { setChatThreadVideoModelFixture } from "../../../test-fixtures/chat-thread-events";
+import {
+  readRunVideoModelFixture,
+  setOrgMemberVideoModelFixture,
+} from "../../../test-fixtures/run-video-model";
 import {
   createBddApi,
   expectApiError,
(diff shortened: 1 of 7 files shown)


bingjiez666:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 5ebe0b7553b75d8651e3e96abc59e8a505fcf29f

### Summary

This PR resolves the run video model from the thread preference, member preference, and catalog default, then snapshots it onto both zero-run launch paths. It also consolidates the three wire-level video-model enums behind `videoModelIdSchema` and adds route-boundary integration coverage for precedence, retired values, threadless runs, and mid-run re-pinning. One catalog-membership check is not exact and must be fixed before merge.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and edge-case handling | Finding (P1) | `zero-video-model.service.ts:31` uses the prototype-aware `in` operator for persisted strings, so inherited keys are accepted as catalog models. |
| `REVIEW.md` | Security and authorization | Pass | No new route or authorization surface is added; all added database predicates are parameterized Drizzle expressions. |
| `REVIEW.md` | Type safety | Finding (P1) | The `as VideoModel` cast at `zero-video-model.service.ts:31` is unsound for inherited keys such as `toString` and `__proto__`. |
| `REVIEW.md` | Feature coverage and regression guard | Pass | `chat-events.bdd.test.ts` covers all three resolution layers, retired-value fall-through, null-thread runs, and the in-flight snapshot invariant; generation remains write-only in this PR. |
| `REVIEW.md` | Persisted-state and JSONB contract review | Pass | Both nullable stored preferences are normalized at read time before the existing `zero_runs.selected_video_model` column is written; no schema or migration changes are included. |
| `docs/bad-smell.md` | No `any`, suppressions, dynamic imports, or hard-coded environment configuration | Pass | The production diff adds none of these patterns. |
| `docs/bad-smell.md` | Fail visibly instead of hiding impossible configuration states | Pass | The precedence chain is the declared product contract for genuinely optional preferences; missing rows are not treated as corrupted required data. |
| `docs/testing.md` | Prefer integration tests with real internal code and a real database | Pass | Coverage runs through the chat/run route boundary and the real dispatch/database path; no internal service or database mock is introduced. |
| `docs/testing/patterns.md` | Exercise API behavior rather than service functions | Pass | Tests create runs through `sendChatRun` / `api.createRun`; no production service function is invoked directly. |
| `docs/testing/anti-patterns.md` | Assert outcomes rather than mock calls or implementation call patterns | Pass | Assertions verify the persisted run snapshot and thread association; no call-count-only assertion, fake timer, or internal mock is added. |
| `docs/testing/api-testing.md` | Use API setup and verification, with explicit narrow exceptions | Pass | The snapshot has intentionally no external reader until PR5, and retired persisted IDs cannot be constructed through current write contracts; the test-specific fixture and PR body explicitly surface that boundary exception. |
| `docs/testing/testing-external-behavior.md` | Leave the external boundary only for states impossible to construct or observe externally | Pass | Direct fixture access is confined to the write-only snapshot and historical retired-ID cases and is documented rather than hidden in a general-purpose fixture. |
| `.claude/skills/ccstate/SKILL.md` | Preserve command accessor and abort-signal ownership rules in API signals | Pass | No ccstate accessor escapes; `prepareRunContext` performs `signal.throwIfAborted()` immediately after the new await. |
| `docs/fallback.md` | Avoid invented compatibility paths and declare rollout fallbacks | Pass | The nullable preference precedence and catalog-retirement normalization are permanent product semantics, not a cross-version rollout branch; no new rollout fallback is introduced. |
| `docs/deployment-compatibility.md` | Keep persisted and API changes additive across deployed versions | Pass | The PR writes an already-deployed nullable column and replaces identical Zod enums with one shared schema; old API code can ignore the new run-row value. |

### Findings

#### Critical (P0)

- None.

- `turbo/apps/api/src/signals/services/zero-video-model.service.ts:31`: `value in VIDEO_MODEL_CONFIGS` checks the prototype chain, not exact catalog ownership. Because these values are intentionally read as unvalidated persisted strings, values such as `toString` or `__proto__` are incorrectly cast to `VideoModel` and snapshotted instead of falling through. Use the existing exact-membership guard (`isVideoModelId`) or an own-property equivalent, and extend the route integration coverage with a non-own catalog key so the contract remains load-bearing.

### Testing

- Coverage: Insufficient only for the P1 catalog-membership edge above; the required precedence, retired-ID, threadless, two-insert-path, and mid-run re-pin behavior is otherwise covered.
- Conventions: Compliant. Tests are route-boundary integrations against the real database, and the unavoidable direct-state exception is explicitly documented.
- CI: All published required checks on commit `5ebe0b7553b75d8651e3e96abc59e8a505fcf29f` are green, including all eight API shards, type/lint/format checks, CodeQL, and security gates.

lancy:
Coordination note for EPIC #26924 / Stage 3 #27033: this PR overlaps `agent-run-create.service.ts` and currently snapshots the resolved `selectedVideoModel` into `zero_runs`. Stage 3 is adding one canonical 15-field metadata tuple to both the initial `agent_runs` and `zero_runs` inserts. If this PR merges first, #27033 will rebase and carry its resolved value through that tuple. If #27033 merges first, please rebase this PR and preserve the same resolved `selectedVideoModel` in both physical rows; a `zero_runs`-only change would reintroduce parity drift after the database bridge is removed. This is coordination only and does not ask this PR to redesign its feature.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: f8215d077a42752d11bc3a410b114727c43a1e46
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

This PR resolves the run video model from the thread preference, member preference, and catalog default, snapshots it onto every zero-run launch path, and consolidates the wire-level video-model schemas. The previous prototype-chain membership finding is resolved by the exact `isVideoModelId` guard with route-level regression coverage; no P0 or P1 findings remain.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and edge-case handling | Pass | `zero-video-model.service.ts` implements the declared precedence, treats retired persisted ids as unset, skips the thread layer for threadless runs, and now rejects inherited object keys through `isVideoModelId`. |
| `REVIEW.md` | Security and authorization | Pass | No route or authorization surface is added; the new reads use parameterized Drizzle predicates and run inside the existing authorized run-creation flow. |
| `REVIEW.md` | Type safety | Pass | Exact membership narrowing replaces the unsound cast from the prior head; the diff adds no `any` or broad cast. |
| `REVIEW.md` | Feature coverage and regression guard | Pass | `chat-events.bdd.test.ts` covers all three resolution layers, retired values, an inherited prototype key, null-thread runs, and the in-flight snapshot invariant. |
| `REVIEW.md` | Persisted-state and JSONB contract review | Pass | The PR writes an already-deployed nullable `zero_runs.selected_video_model` column and normalizes nullable string preferences before writing; it adds no schema or migration change. |
| `docs/bad-smell.md` | Avoid `any`, suppressions, dynamic imports, and hard-coded environment configuration | Pass | The production diff introduces none of these prohibited patterns. |
| `docs/bad-smell.md` | Fail visibly instead of hiding impossible configuration states | Pass | Missing nullable preferences use the documented product precedence; the catalog default is not masking a required configuration invariant. |
| `docs/testing.md` | Prefer integration tests with real internal code and a real database | Pass | Coverage exercises the real chat/run route and dispatch/database path without mocking internal services or the database. |
| `docs/testing/patterns.md` | Exercise route behavior rather than service functions | Pass | Tests create runs through `sendChatRun` / `api.createRun`; no production service function is called directly. |
| `docs/testing/anti-patterns.md` | Assert observable outcomes rather than mock calls or implementation call patterns | Pass | Assertions verify the stored run snapshot and thread association; no call-count-only assertion, fake timer, or internal mock is added. |
| `docs/testing/api-testing.md` | Use API setup and verification, with explicit narrow exceptions | Pass | Run creation stays at the route boundary. Direct fixture access is limited to the intentionally write-only snapshot and historical persisted ids that current APIs cannot construct, and the exception is explicitly documented. |
| `docs/testing/testing-external-behavior.md` | Leave the external boundary only for states impossible to construct or observe externally | Pass | The new fixture states why the write-only output and retired-input cases require direct DB access instead of hiding that exception in a generic helper. |
| `.claude/skills/ccstate/SKILL.md` | Preserve accessor and abort-signal ownership rules in API signals | Pass | No ccstate accessor escapes; `prepareRunContext` checks `signal.throwIfAborted()` immediately after the new database resolution. |
| `docs/fallback.md` | Avoid invented compatibility paths and declare rollout fallbacks | Pass | Nullable preference precedence and catalog-retirement normalization are permanent product semantics, not cross-version rollout compatibility; no new rollout fallback is introduced. |
| `docs/deployment-compatibility.md` | Keep persisted and API changes additive across deployed versions | Pass | The PR only begins writing an existing nullable column and replaces identical Zod enums with a shared schema; previous API code safely ignores the new stored value. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- Coverage: Adequate. The route integration suite covers precedence, retired and prototype-chain-invalid ids, threadless runs, and mid-run re-pinning.
- Conventions: Compliant. The tests use the real route and database path, and the unavoidable direct-state exception is narrowly scoped and documented.
- Local verification on `f8215d077a42752d11bc3a410b114727c43a1e46`: repository format and targeted Prettier checks, repository lint with the bounded 3072 MiB heap, repository type checks, and `CHAT-02: run video model snapshot` (3 passed, 92 skipped) all passed.
- CI: Still running at review time; published checks observed so far have no failures.

PR #26918, opened 2026-08-13
perf(python): bound capture header serialization before network-log queueing

### LINKED ISSUE #26864: perf(python): bound capture header serialization before network-log queueing
## Problem

At `7a80c6c506db436956c7a8e23dafa7d6c9014955`, capture-mode HTTP network logging bounds captured bodies and individual header names/values, but it does not bound the aggregate header count or retained header bytes.

- `crates/runner/mitm-addon/src/body_capture.py:239-250` walks every raw header field and grows both `seen_names` and the result dictionary for every unique valid name.
- `crates/runner/mitm-addon/src/body_capture.py:304` and `:348` do this for both request and response headers.
- `crates/runner/mitm-addon/src/mitm_addon.py:1777-1780` performs capture and logging synchronously in the response completion path.
- `crates/runner/mitm-addon/src/logging_utils.py:145-157` JSON-encodes the complete row before queue admission. Above 1,000,000 bytes it also encodes a URL-omission candidate; header-induced overflow leaves the same header dictionaries in that candidate, so the path performs another full serialization and then retains the original oversized line.
- `crates/runner/mitm-addon/src/jsonl_writer.py:70-74` applies the 32 MiB pending-byte budget only after the full byte string exists.
- `crates/runner/src/network_logs.rs:32-36,421-424` limits remote batches to 1 MiB and drops a single entry that cannot fit.

The lock pins mitmproxy 12.2.3 at `crates/runner/mitm-addon/uv.lock:596`. Its HTTP/1 reader has no request-head size or field-count limit (`mitmproxy/proxy/layers/http/_http1.py:320` in the pinned wheel is still `FIXME: protect against header size DoS`), so a registered capture-enabled workload can reach this path with an arbitrarily large set of distinct small headers. Upstream response headers are likewise input to the same unbounded capture helper.

## Evidence and impact

Measured on the commit above with Python 3.12.3 and the pinned mitmproxy 12.2.3. The benchmark used production `_sanitize_headers_for_capture()` followed by production `log_http_network_entry()` with a short URL and a sink in place of the asynchronous writer, isolating the synchronous response-hook work. Each input field was a distinct `x-NNNNNN: v`; timings are medians and traced allocations start after mitmproxy's header object already exists.

| Header fields | Approx. HTTP/1 head bytes | Emitted JSONL bytes | Median synchronous time | Peak incremental allocation |
| 10,000 | 130,002 | 190,110 | 11.0 ms | 1.6 MiB |
| 50,000 | 650,002 | 950,110 | 60.2 ms | 8.8 MiB |
| 100,000 | 1,300,002 | 1,900,110 | 138.9 ms | 17.7 MiB |
| 200,000 | 2,600,002 | 3,800,110 | 292.9 ms | 35.3 MiB |
| 500,000 | 6,500,002 | 9,500,110 | 773.2 ms | 77.4 MiB |

This is marginal work added after mitmproxy has parsed the request. A single capture-enabled response can pause unrelated flows on the proxy event loop for hundreds of milliseconds and allocate tens of MiB. From 100,000 fields onward, the resulting row is already too large for the Rust uploader, so the proportional sanitization, dictionary growth, duplicate JSON encoding, queueing, and possible local write cannot produce an uploadable event. Inputs large enough to exceed the Python pending-byte budget are still fully processed before rejection.

This is distinct from #25330. That issue bounded URL-induced serialization and explicitly left aggregate capture-header limits as a separate problem; the current URL-omission candidate intentionally preserves oversized non-URL fields.

## Acceptance criteria

- Put an explicit aggregate budget on captured request headers and captured response headers before the sanitizer builds an unbounded dictionary or the logger encodes an unbounded header projection.
- Bound both traversal work and retained allocations; once the raw-field or byte budget is exhausted, stop processing the remaining fields rather than scanning the complete attacker-controlled collection.
- Expose structured request/response header truncation state in the network-log contract so retained diagnostics are not mistaken for complete headers.
- Preserve current first-occurrence, case-insensitive deduplication, redaction, and allowlisted-value behavior for ordinary in-budget headers.
- Ensure header-induced capture rows fit the Python producer budget and Rust single-entry upload envelope without first constructing or serializing a larger candidate.
- Cover both request and response headers through real terminal hook paths, including an over-budget set of many unique valid names; use structural count/byte assertions rather than timing thresholds.
- Keep HTTP ingress behavior out of scope: the fix should bound the logging amplification even if mitmproxy continues accepting large request heads.

seven332:
## Summary

- stop request and response header capture after 512 raw fields or 32 KiB of raw name/value bytes per side, before decoding or sanitizing the remaining fields
- preserve existing first-occurrence deduplication, redaction, and allowlisted values for retained headers while emitting structured truncation metadata
- carry header truncation state through the network-log contract and API sanitizer, and show it in activity inspection even when the retained prefix is empty
- cover both sides through the real terminal response hook and assert the worst-case captured row remains within the 1,000,000-byte producer/uploader envelope

## Scope

- mitmproxy HTTP ingress limits remain unchanged
- request/response body limits and the existing URL-omission behavior remain unchanged
- no database migration or generated Rust binding change is required

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (5,365 passed)
- `pnpm -F @okouai/api-contracts generate:rust` (no generated diff)
- `pnpm format`
- `pnpm turbo run lint`
- `pnpm check-types`
- `pnpm vitest run --maxWorkers=4 --silent=passed-only` (8,510 passed; two unrelated platform tests hit the default 5-second timeout)
- exact retry of both timed-out tests with the default timeout (2 passed)
- `pnpm -F @okouai/app exec vitest run --maxWorkers=4 --silent=passed-only` (1,596 passed)
- `pnpm knip`
- `pnpm build`

Closes #26864


seven332:
## Code Review: PR #26918

**Reviewed head:** `131da0e7c2f23dffe517a6e131c534c05b02a9e9`

### Summary

The producer now bounds request and response header capture before decoding or retained allocation, preserves existing sanitization semantics, and carries optional truncation state through the API and activity UI. The additive contract remains safe across runner/API/frontend rollout combinations.

### Findings

No findings.

### Testing

Python static checks and all 5,365 addon tests passed. Turbo formatting, lint, types, Knip, generated-binding verification, focused integration tests, all 1,596 platform tests, and the normal production build passed. The full Turbo run passed 8,510 tests; two unrelated platform tests hit the default timeout under full-repository contention, then passed both exact isolation and the full platform rerun. Production traffic replay was not performed.

### Verdict

LGTM

PR #26920, opened 2026-08-13
perf(runner): reduce gc tree walk task handoffs

### LINKED ISSUE #26888: perf(rust): avoid per-entry blocking-pool handoffs in GC tree walks
## Problem and evidence

At audit baseline `6b22a4d16d55fe367fd16df39b1fb26afedcbfee`, recursive GC accounting is implemented as a long sequence of individually awaited Tokio filesystem operations:

- `crates/runner/src/cmd/gc/filesystem.rs:130-215` walks a directory tree with `tokio::fs::symlink_metadata`, `tokio::fs::read_dir`, and `ReadDir::next_entry` inside nested serial loops.
- Every entry reaches a separately awaited `symlink_metadata` at `filesystem.rs:195-203`; every directory adds further metadata and directory-iteration awaits at `filesystem.rs:163-207`.
- The locked workspace GC path invokes this full walk before removing each orphan workspace (`crates/runner/src/cmd/gc/workspaces.rs:488-523`). User workspaces can contain high-cardinality trees such as dependency installations.
- Storage GC invokes it for every unlocked version during inventory (`crates/runner/src/cmd/gc/storage.rs:200-225`), with an allowed cache cardinality of 5,000 entries (`storage.rs:19-25`). Image GC also invokes it during inventory and revalidation (`crates/runner/src/cmd/gc/images.rs:303-325` and `555-571`).
- Runner promotion and rollback execute this GC command through `ansible/tasks/garbage-collect-runner.yml:1-12`, so traversal latency extends deployment maintenance rather than only an offline diagnostic.

The workspace currently resolves Tokio `1.53.1`. In that version, `tokio::fs::symlink_metadata` delegates each call through `asyncify`/`spawn_blocking`; `read_dir` also uses the blocking pool and refills its iterator in 32-entry blocking chunks. Because the GC loop awaits each operation before submitting the next one, a tree with *N* entries incurs at least *N* blocking-pool task submissions and async wakeups in addition to the filesystem syscalls.

### Focused reproduction

I mirrored the current traversal in a release-mode disposable benchmark using the workspace's Tokio `1.53.1`. The benchmark created 20 directories with 1,000 empty files each, warmed filesystem caches, verified both implementations returned identical allocated bytes, and compared three rounds:

| Round | Current per-entry Tokio pattern | One `spawn_blocking` std::fs traversal | Ratio |
| --- | ---: | ---: | ---: |
| 1 | 192.445 ms | 30.363 ms | 6.34x |
| 2 | 193.604 ms | 29.082 ms | 6.66x |
| 3 | 187.549 ms | 29.022 ms | 6.46x |

The comparison intentionally retained the same metadata calls, including the duplicate root stat, so the difference isolates task scheduling and wakeup overhead rather than reducing filesystem work.

This is distinct from #25641, which avoids a second storage-cache tree walk for unchanged eviction candidates but leaves the first inventory walk unchanged. It is also distinct from #19157, which optimized metadata calls in the separate workspace-image-cache walker and explicitly left Tokio scheduling unchanged pending evidence; the reproduction above supplies that evidence for the shared runner GC walker.

## Impact

A single 20,000-entry directory spends roughly 160 ms in avoidable blocking-pool handoffs on a warm local filesystem. Large orphan workspaces and high-cardinality storage/image inventories multiply that cost serially, increasing `runner gc` wall time and blocking-pool churn during promotion, rollback, and manual cleanup.

## Acceptance criteria

- Perform one recursive directory-accounting walk with one blocking task, or another bounded design that avoids one blocking-pool task submission per filesystem entry.
- Preserve current allocated-byte accounting (`st_blocks * 512`), root-directory mtime semantics, no-follow symlink behavior, and best-effort handling of disappearing or unreadable entries.
- Preserve the `root_metadata` completeness contract: it is available only after a complete walk, so storage eviction cannot reuse incomplete accounting.
- Preserve cancellation/lifecycle safety for the GC command; a cancelled caller must not leave harmful unbounded background work.
- Keep existing image, storage, workspace, symlink, incomplete-scan, dry-run, replacement, and low-FD tests passing.
- Add a focused benchmark or task-count instrumentation for a representative high-entry tree and demonstrate a material reduction in traversal overhead.

seven332:
## Summary

- perform each GC directory-stat walk inside one blocking task instead of handing every filesystem operation to Tokio's blocking pool
- preserve no-follow symlink handling, disk-block accounting, best-effort traversal, and completeness-gated root metadata
- cooperatively stop the blocking walk when its async owner is dropped, with coverage for cancellation and high-entry trees

## Exclusions

- does not change eviction policy, accounting semantics, or unrelated GC directory scans
- does not add a benchmark framework or new dependencies

## Validation

- `cargo test --manifest-path crates/Cargo.toml -p runner --all-targets --all-features --quiet` (3,104 passed; 20 existing helper tests ignored)

Closes #26888

### DIFF
--- a/crates/runner/src/cmd/gc/filesystem.rs
+++ b/crates/runner/src/cmd/gc/filesystem.rs
@@ -2,6 +2,7 @@ use std::os::unix::fs::MetadataExt;
 use std::path::Path;
 use std::time::SystemTime;
 
+use tokio_util::sync::CancellationToken;
 use tracing::warn;
 
 use crate::error::{RunnerError, RunnerResult};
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #26920

**Reviewed head:** `9e622ad2c9d2780db04606db308c50bbe19c27cd`

### Summary

The PR replaces per-operation Tokio filesystem handoffs with one owned blocking traversal while preserving GC accounting, no-follow traversal, best-effort completeness, and async cancellation behavior.

### Findings

No findings.

### Testing

The review covered real-filesystem tests for symlinks, incomplete iteration, a 1,024-file task-count case, and deterministic cancellation. Runner formatting, Clippy, rustdoc, focused GC consumer suites, low-FD coverage, and the full all-target/all-feature Runner suite passed locally (3,104 tests passed; 20 existing helper tests ignored). Remote CI and a post-change release-mode wall-clock benchmark were not part of this review.

### Verdict

LGTM

PR #26930, opened 2026-08-13
refactor(python): require billing firewall registry field

### LINKED ISSUE #26890: refactor(python): retire missing billableFirewalls registry fallback
## Classification

**Removal-ready debt**

## Compatibility behavior

At commit `6b22a4d16d55fe367fd16df39b1fb26afedcbfee`, `crates/runner/mitm-addon/src/auth.py:216-232` treats a missing or falsey runner-registry field as an empty billing whitelist:

```python
return firewall_name in (vm_info.get("billableFirewalls") or [])
```

Commit `d8e23b9b110b3979322ba44869a7cffe6cf289cf` / PR #10446 introduced the fallback with a regression test. Its stated purpose was compatibility with an optional TypeScript claim field and the assumption that the Rust runner might omit `billableFirewalls` for runs with no billable connector.

That assumption does not describe the runner-to-Python registry boundary:

- `crates/runner/src/types.rs:131-135` uses `#[serde(default)] Vec<String>` to normalize an omitted API claim field to an empty vector.
- `crates/runner/src/executor/sandbox_run.rs:1548-1573` always passes that vector to `VmRegistration`.
- `crates/runner/src/proxy/registry.rs:24-50` defines `VmEntry.billable_firewalls: Vec<String>` with neither a Serde default nor `skip_serializing_if`.
- `crates/runner/src/proxy/registry.rs:573-599` always serializes the field.
- `crates/runner/src/proxy/registry.rs:2183-2184` explicitly asserts that an empty value reaches Python as `"billableFirewalls": []`.

Issue #20906 proposed making missing local-registry values backward-compatible in Rust and was closed as `NOT_PLANNED`. Current Rust intentionally does not accept that old shape.

## Removal-readiness evidence

The old state cannot reach the current addon:

- `crates/runner/src/proxy/process.rs:104-126` extracts the addon embedded in the current runner and overwrites the runner-local registry with an empty current-schema file before mitmdump starts.
- All later entries are written by the same runner process through the required `VmRegistration.billable_firewalls` field.
- Old runners and addons may coexist while draining, but each uses its own local registry and code artifact.
- A stale file missing this field is already unsupported by the current Rust reader, and startup replaces it before the addon runs.

The exit gate is therefore satisfied by construction. API omission remains supported at the Rust claim boundary and becomes the canonical empty list before Python sees the entry.

## Why this is not permanent compatibility

The proxy registry is a private runner-owned contract, not an externally controlled payload or durable user format. An empty list is still a supported current value; only field absence or malformed shape is obsolete.

Retaining the fallback is risky because a malformed registry entry silently marks every matched firewall non-billable. That suppresses connector/model usage and turns a contract violation into silent underbilling. The comment at `auth.py:230-231` also conflates the optional API claim with the required local registry field.

## Removal scope and acceptance criteria

- Make Python registry classification require `billableFirewalls` to be a list of strings for usable VM entries.
- Remove the missing/falsey fallback from `is_billable_firewall()` and consume the validated current field.
- Classify a missing or malformed value as an invalid VM entry through the existing invalid-registry path instead of silently disabling billing.
- Preserve `[]` as the canonical current representation for a run with no billable firewalls.
- Replace `tests/test_firewall_auth_handling.py:874-891` (`test_missing_billable_firewalls_falls_back_to_empty`) with current-contract coverage for canonical `[]` and invalid missing/malformed entries.
- Update hand-written valid registry fixtures to include the canonical field where needed.
- Remove the stale `auth.py:230-231` comment that says the runner may omit the local field.
- Do not make the TypeScript claim field required, add a Rust Serde compatibility default, or introduce a rollout switch.

seven332:
## Summary

- require each usable proxy-registry VM entry to provide `billableFirewalls` as a list of strings
- reject missing or malformed values through the existing per-VM `invalid_registry_vm` path before auth injection
- remove the auth-layer fallback, preserve canonical `[]`, and update current-shape test/documentation fixtures

## Compatibility

- the backend execution-context claim remains optional and Rust still normalizes omission to an empty vector
- the Rust local registry writer always serializes `billableFirewalls`, including canonical `[]`, and the current Rust reader already requires the field
- the registry writer and embedded Python addon ship in one runner artifact; runner startup recreates both the addon files and local registry before mitmdump starts, so no supported rollout pairs this addon with an old writer
- no API, queue, database, Rust deserialization, feature-switch, or rollout behavior changes

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_request_handler_registry_admission.py tests/test_firewall_auth_handling.py` (98 passed)
- focused stale-fixture verification across 10 affected test files (198 passed)
- `uv run --no-sync python -m pytest tests/` (5373 passed)

Closes #26890

### DIFF
--- a/crates/runner/mitm-addon/src/auth.py
+++ b/crates/runner/mitm-addon/src/auth.py
@@ -215,7 +215,7 @@ class _ResolvedFirewallAuth:
 
 def is_billable_firewall(firewall_name: str, vm_info: dict) -> bool:
     """Return whether this firewall should emit connector/model usage."""
-    return firewall_name in (vm_info.get("billableFirewalls") or [])
+    return firewall_name in vm_info["billableFirewalls"]
 
 
 def _prepare_firewall_metadata(
@@ -227,8 +227,6 @@ def _prepare_firewall_metadata(
     api_entry = allow.api_entry
     firewall_base = api_entry["base"]
     api_id = api_entry.get("id", firewall_base)
-    # billableFirewalls is optional in the TS schema; runner may omit the
-    # field entirely for non-vm0 / no-billable-connector runs.
     firewall_billable = is_billable_firewall(allow.name, vm_info)
 
     flow.metadata[metadata_keys.FIREWALL_BASE] = firewall_base
--- a/crates/runner/mitm-addon/src/registry.py
+++ b/crates/runner/mitm-addon/src/registry.py
@@ -258,6 +258,16 @@ def _classify_registry_vms(
             )
             continue
 
+        billable_firewalls = vm.get("billableFirewalls")
+        if not isinstance(billable_firewalls, list) or any(
+            not isinstance(firewall_name, str) for firewall_name in billable_firewalls
+        ):
+            invalid_vms[client_ip] = InvalidVmEntry(
+                "invalid_billable_firewalls",
+                "proxy registry VM entry billableFirewalls must be a list of strings",
+            )
+            continue
+
         if (
             "firewalls" in vm
             and vm["firewalls"] is not None
--- a/crates/runner/mitm-addon/tests/conftest.py
+++ b/crates/runner/mitm-addon/tests/conftest.py
@@ -546,6 +546,7 @@ def registry_file(tmp_path):
         "vms": {
             "10.200.0.1": {
                 "runId": "run-abc-123",
+                "billableFirewalls": [],
                 "sandboxToken": "tok-xyz",
                 "registeredAt": 1700000000000,
                 "networkLogPath": str(tmp_path / "network.jsonl"),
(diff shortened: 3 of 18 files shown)


seven332:
## Code Review: PR #26930

**Reviewed head:** `a9e6de8e7e7d0ae85e4eb62c97b972883c9f3614`

### Summary

The submitted change makes `billableFirewalls` part of the addon's validated local registry contract, rejects malformed per-VM entries before credential lookup or injection, and removes the now-unreachable auth-layer fallback. The request-entry integration coverage exercises missing and malformed values, while valid fixtures preserve the canonical empty-list shape.

### Findings

No findings.

### Testing

- Ruff format and lint passed.
- BasedPyright passed with no diagnostics.
- The complete mitm-addon suite passed: 5,373 tests.
- Focused request-entry and auth coverage passed: 98 tests.
- No new fallback was introduced. The removed fallback is supported by the Rust writer always serializing the field, the Rust reader already requiring it, and the writer/addon being deployed and recreated as one runner artifact.
- No live runner smoke test was performed locally; no Rust source or cross-process protocol changed in this PR.

### Verdict

LGTM

Issue #26576, opened 2026-08-12
refactor(api): start template analysis after browser page commit


bingjiez666:
Part of #21350. Depends on #26571.

Replace the merged sandbox-rendering lifecycle from #26301 with a browser-rendered ingestion handoff. The API must not start the template-analysis run until the original PPTX and every browser-rendered page PNG have been uploaded and verified.

## Target lifecycle

1. The browser renders and reviews the local deck through #26571.
2. The browser opens an import and receives its template id.
3. The browser requests one upload slot per file from that import — the source deck, then each page by index — and PUTs the bytes to the URL each slot returns.
4. A browser-authenticated commit closes the import. It carries no object ids: the API reads the slots it allocated, verifies the set is complete and ordered, and atomically launches exactly one `template-import` run.
5. The run downloads the committed PPTX and PNGs through narrowly scoped CLI reads. It must not re-render, replace, or reorder the page images.
6. The run performs one bounded extraction pass, publishes one private visual guidance package, and does not run a similarity/proof/rebuild loop.
7. The row becomes `ready` only after the complete package is validated and atomically published; any terminal run failure becomes `failed`.

## Upload grouping is server-owned

The client never names object ids. Opening an import returns a template id; each upload asks that import for a slot and the API allocates the object and records which import and page it fills; commit reads its own rows.

This replaces an earlier client-supplied manifest, where the browser named one PPTX upload id plus an ordered list of page ids. That grouping only ever existed in a browser variable held across rendering and a long upload sequence, so a stale client could pair one deck's source with another deck's pages. Both halves are real render output and both resolve, and the mismatch is silent: the analysis run reads page N against slide N, so every extracted conclusion is misaligned. A server-issued import removes the client's ability to express the wrong pairing at all.

It also removes roughly 200 S3 round trips per commit (up to 101 uploads each resolved through `ListObjects` plus `HeadObject`), makes abandoned uploads reachable from the import that owns them, and turns a retried page into a slot replacement.

Staging rows live in `presentation_template_uploads`, not `run_uploaded_files`: that table's catalog trigger would project every slide crop into the user's artifact catalog (#26235). Commit freezes the ordered result onto `presentation_templates.source_storage_key` and `page_keys` and drops the staging rows, so a committed template is readable without a join. `source_storage_key` is nullable because an import exists before its deck is uploaded.

## What the API reads from the source bytes

Only the slide count, from one trailing byte range over the ZIP central directory.

[...]

Issue #26571, opened 2026-08-12
feat(platform): establish browser-native PPTX rendering


bingjiez666:
Part of #21350.

Establish the browser-native rendering primitive used by both the uploaded-deck preview and the persisted page PNGs. This is the first dependency in the revised import pipeline: no API or sandbox change should assume browser rendering works until this lands.

## Scope

- Parse a local `.pptx` `File` / `ArrayBuffer` entirely in the browser. Reject malformed, encrypted, legacy `.ppt`, PDF, HTML, and other unsupported inputs before upload.
- Evaluate browser-native renderers against representative vm0 fixtures and pin one only after evidence. The initial open-source candidates are `@aiden0z/pptx-renderer` (Apache-2.0, HTML/SVG) and `pptxviewjs` (MIT, Canvas); proprietary packages are out of scope.
- Render every accepted slide on the fixed 16:9 vm0 surface.
- Use one render tree/surface for both the visible preview and PNG export. Do not implement a separate screenshot renderer.
- Wait for `document.fonts.ready`, decoded images, charts, and other renderer-owned asynchronous assets before a slide becomes exportable.
- Export ordered PNG `Blob`s without sending the source deck or intermediate render state to a server.
- Keep parsing/rendering work bounded so a large deck cannot permanently block the main UI thread; preserve the existing source-size and 100-page caps.
- Expose a small typed platform API that the later upload UI can consume. Do not add template catalog, R2 upload, polling, or selection UX in this issue.

## Fidelity contract

- The persisted PNG must match what the user saw in the vm0 browser preview for that render session.
- PowerPoint pixel parity is not the contract. Unsupported PowerPoint constructs must be detected or visibly degraded in the preview before upload; they must not silently produce a different persisted image.
- A ready template preview later reads the persisted PNGs instead of re-rendering the PPTX on another device.

## Verification

- Integration coverage loads real PPTX fixtures and exercises parse, preview readiness, ordered export, and stable failures.
- Fixtures include text with bundled/web fonts, CJK text, images, common shapes, tables, charts, groups, and unsupported content.
- For every supported fixture, the preview surface and exported PNG are produced from the same completed render state.
- Record renderer choice, license, browser bundle cost, unsupported constructs, and fixture results in the PR body.

## Out of scope

- Uploading to R2.
- Starting the analysis agent.
- Template catalog or picker UX.
- Public or organization sharing.

bingjiez666:
[...]

Issue #26238, opened 2026-08-11
feat(api): resolve uploaded presentation templates for generation


bingjiez666:
Part of #21350. Depends on #26576 and #26237.

Let a chat message select a ready uploaded template and let the generation run fetch and mount its complete private visual guidance package.

## Scope

- Selection uses the existing `presentationGenerationTemplateRequestSchema` with namespaced id `user-template:<uuid>`. No new request field.
- Resolve the id against a `ready` presentation-template row and enforce organization plus owner/visibility authorization.
- Return a short-lived signed GET for the complete private package archive after authorization; raw storage keys never enter the message or client payload.
- The generation run downloads and mounts the whole package, reads `SKILL.md` and `design-system.md` first, and inspects color CSS and assets only as needed.
- The prompt asks the AI to create the final presentation directly as semantic HTML/CSS/SVG for the current content. It must not first generate slide JSON, consume `tokens.json`, call a layout ID API, or invoke a template-specific JSON-to-HTML renderer.
- Recommend CSS Grid/Flexbox for live rows, columns, and text flow. Backgrounds, fixed chrome, decoration, and intentional overlays may use absolute positioning. These are authoring guidelines, not a renderer policy.
- Background images are optional visual material; new text, charts, tables, labels, and diagrams remain live HTML/SVG.
- After normal presentation render checks, use the existing presentation HTML publication path and `okou host … --artifact-kind presentation-html`.
- Gate dynamic ids through `PresentationTemplates` so old API/CLI versions cannot receive them during rollout.

## Acceptance

- Unknown, non-ready, deleted, or unauthorized template ids fail before dispatch with a clear error.
- An unauthorized user cannot resolve or download another user's package.
- The generation run receives the complete verified tree, including binary assets, and reads the two required guidance files before authoring.
- Generated presentations contain no intermediate slide JSON or dependency on a template renderer/layout schema.
- Built-in template selection and prompt output are unchanged.
- Package URLs are signed and expire; package pull verifies integrity and safely extracts the archive.
- Integration coverage proves direct HTML/CSS/SVG generation handoff and reuse of the normal presentation HTML publication flow.

## Out of scope

- Browser PPTX rendering and ingestion.
- Public or organization sharing UI.
- A customer-editable renderer schema, layout ID API, or JSON-to-HTML template compiler.

Issue #26237, opened 2026-08-11
feat(cli): consume browser-rendered presentation template inputs


bingjiez666:
Part of #21350. Depends on #26576.

Provide the run-scoped I/O required after the browser has rendered and committed the source PPTX and ordered page PNGs. The sandbox is an analyzer and package producer, not a renderer.

## Command group

`okou presentation-template`:

- `source pull --id <templateId> --out <path>` — download the committed source PPTX through the run-scoped source endpoint.
- `pages pull --id <templateId> --dir <path>` — atomically download the committed browser-rendered PNGs in authoritative page order.
- `publish --id <templateId> --dir <path>` — validate the visual guidance package, create a temporary `.tar.gz`, upload it through `/api/okou/uploads/prepare`, and atomically commit the returned `archiveFileId`.
- `fail --id <templateId> --code <code> --message <text>` — report a stable terminal analysis failure.
- `pull user-template:<uuid> --dir ./generated/resources` — download and safely extract the entire private package for a generation run.

## Package contract

```text
SKILL.md
design-system.md
color-systems/<token>.css        # optional
assets/identity/*                # optional
assets/backgrounds/*             # optional
assets/fonts/*                   # optional
```

- No JSON, no `LAYOUTS.md`, no empty directories, and no placeholders.
- Keep binary assets as archive bytes; never base64-encode them into request JSON.
- The CLI should reject obvious local structure/limit violations before upload while treating the server as authoritative.
- Always remove the temporary archive. Abort multipart uploads when preparation/upload fails.

## Removed from the superseded design

- No `pages upload` command.
- No PNG dimension or aspect-ratio detection in the sandbox.
- No LibreOffice, PDF, Poppler, screenshot, similarity, proof, or rebuild command.
- No compatibility fallback for the pre-GA sandbox-rendered protocol.

## Security and visibility

- Run commands authenticate through `ZERO_TOKEN` capabilities and are visible only when the corresponding presentation-template capability is present.
- Source and page reads remain bound to the authenticated `template-import` run, organization, owner, and template id.
- Package publication uses the existing private artifact upload transport plus the run-scoped atomic package commit endpoint.
- `pull user-template:<uuid>` uses the owner-authorized package endpoint from #26238, verifies the archive digest, rejects unsafe or unsupported tar entries, and does not expose storage keys or credentials.

## Acceptance

- `pages pull` writes contiguous ordered `page-<number>.png` files and fails atomically when any page cannot be fetched or verified.
[...]

PR #26947, opened 2026-08-13
feat(api): commit uploaded presentation template inputs

bingjiez666:
## Summary

- reuse the existing authenticated private upload preparation route for one original PPTX and ordered browser-rendered PNG files; commit accepts only `requestId`, `sourceFileId`, and ordered `pageFileIds`
- resolve the uploaded records and preflight the source PPTX before atomically starting at most one analysis run; browser PNG order remains authoritative, while page payload bytes and aspect ratio are not revalidated server-side and the run never re-renders or replaces those images
- replace the schema-driven three-text-file extraction contract with a reusable visual guidance package containing required `SKILL.md` and `design-system.md`, plus optional color-system CSS and binary identity, background, and font assets
- instruct extraction to combine exact OOXML evidence with final browser-rendered appearance evidence in one bounded pass, preserve reusable visual language rather than customer content, and disclose assets that cannot be cleanly separated from flattened pixels
- make the future authoring contract explicit: read the package, then generate final semantic HTML/CSS/SVG directly from the new content; there is no slide JSON, `tokens.json`, layout ID API, JSON-to-HTML step, or template-specific renderer

## Package publication protocol

The analysis CLI will create a `.tar.gz`, upload it through the existing `/api/okou/uploads/prepare` flow, and commit the returned `archiveFileId` to the template package endpoint. Binary files remain binary; they are never base64-encoded into JSON.

The server decompresses and validates the archive before publication:

- required root files: `SKILL.md` and `design-system.md`
- optional files only under `color-systems/*.css`, `assets/identity/**`, `assets/backgrounds/**`, and `assets/fonts/**`
- no JSON files, unsupported roots, links, duplicate paths, unsafe paths, empty files, or empty declared directories
- at most 256 files, 25 MiB per file, 100 MiB total package bytes, and a 128 MiB uploaded archive
- Markdown and CSS must be non-empty valid UTF-8; binary assets are preserved byte-for-byte

Only after the complete validated volume is prepared does the lifecycle transaction move the template to `ready` and commit the package head. Validation/publication failures remove the temporary upload and clean unpublished package objects without deleting the independently owned source PPTX or page PNGs.

## Stack

#27172 is merged. This PR has been rebased onto current `main` and now contains only the input commit/analysis/package layer. The foundation PR itself was not modified by this revision.

The downstream CLI and generation work is tracked in #26237 and #26238. The browser renderer contract remains only an original `.pptx` plus ordered PNG blobs; no #26571 renderer-internal types or APIs are imported.

## Verification

- changed-file Prettier write and check
- workspace lint: 13/13 packages successful
- `pnpm knip`
- non-API package typechecks: 12/12; separate `@okouai/app` and `api` typechecks
- full migration against temporary PostgreSQL 18
- `pnpm --filter api exec vitest run src/signals/routes/__tests__/zero-presentation-templates.test.ts` (5 passed), including binary PNG/WOFF2 preservation, JSON rejection, path-traversal rejection, required-file validation, atomic ready transition, and cleanup
- post-rebase range-diff confirmed both original PR patches were preserved; verification above was repeated on current `main`

Refs #26576
Refs #21350


bingjiez666:
Closing in favour of a three-way split. This branch has not been rebased since 2026-08-13 and now conflicts with the contract/route rename in #28075.

The work is being landed as:

1. **#28031 — ingestion.** Open an import, allocate upload slots, commit. Already open and verified.
2. **Analysis run.** Launch the `template-import` run, run-scoped source/page/fail endpoints, the run callback, and the `okou presentation-template` CLI those endpoints need (#26237).
3. **Package publication.** `publish` endpoint, archive validation, atomic storage publication.

Two things changed relative to this branch while splitting it.

**Uploads are now scoped to a server-issued import.** This branch's commit took a manifest — the browser named one PPTX upload id plus an ordered list of page ids. That grouping only ever existed in a browser variable held across rendering and a long upload sequence, so a stale client could pair one deck's source with another deck's pages; both halves resolve, and the mismatch is silent because the analysis run reads page N against slide N. #28031 replaces it with the upload-session shape: commit takes no object ids, so a wrong pairing cannot be expressed. It also removes roughly 200 S3 round trips per commit and makes abandoned uploads attributable. #26576 has been updated accordingly.

**Preflight no longer re-validates the archive.** The browser rendered these pages from a deck it opened, which already establishes that the file is an unencrypted, untruncated, non-legacy PPTX. Only the slide count survives, as a guard on whether every page was exported.

The `presentation_templates` schema and the `PresentationTemplates` feature switch are unaffected.

PR #26950, opened 2026-08-13
refactor(rust): share active-input control payload contract

### LINKED ISSUE #26893: refactor(rust): share the active-input control payload contract
## Problem and evidence

At `6b22a4d16d55fe367fd16df39b1fb26afedcbfee`, and still on current `main`, runner and guest-agent independently own the same active-input process-control JSON contract:

- Runner has a borrowed serialize-only DTO with the `type`, `deliveryId`, and `text` field names and a hard-coded `active-input` tag.
- Guest-agent has a separate tag and owned deserialize-only DTO, then separately validates non-empty text, UUID syntax, and canonical lowercase hyphenated UUID spelling.
- Valid test fixtures in both consumers spell the same schema again.

Both binaries already depend on `guest-contracts`, whose documented purpose is to own names and values that runner and guest must keep in lockstep. The adjacent active-input receipt contract already lives there.

A one-sided field, tag, or validation change can compile and pass crate-local tests, then fail only after runner has reserved and transported a delivery to the guest. There is no shared codec test tying producer bytes to consumer acceptance.

## Decision

Create one `guest-contracts::active_input` codec owner with private wire representations and a narrow public encode/encoded-length/decode API.

- Runner supplies borrowed delivery-ID and text strings to the shared encoder and size calculation.
- Guest-agent decodes into an immutable owned validated result.
- The shared decoder owns tag, required-field, non-empty text, UUID syntax, and canonical UUID rules.
- A small decode-error category lets guest-agent preserve its current rejection diagnostics.
- Guest runtime enablement, lifecycle, queue capacity, deduplication, receipts, replay filtering, and backpressure remain in guest-agent.

The existing JSON stays `{"type":"active-input","deliveryId":"...","text":"..."}` and unknown fields remain accepted.

## Rationale and alternatives

A codec API is the smallest complete surface for the actual caller operations. It avoids exposing `Cow` and public lifetimes through one generic DTO, and avoids the larger API plus two public field representations required by paired borrowed/owned DTOs. Sharing only the tag and UUID helper was rejected because it would leave the duplicate Serde field schema and its main drift path intact.

The 1 MiB maximum remains owned by the TypeScript API contract and generic `process-control-ipc` transport. The shared codec owns exact serialized-length calculation, while integration tests use the real transport constant through a test-only dependency. Runner retains its compile-time assertion that the generated API limit equals the transport limit.

Producer-side validation is not duplicated: API reserve responses already require a UUID and non-empty prompt, and local submit validates non-empty text while generating a canonical UUID. Decode remains the current trust boundary.

## Compatibility and risks

- Runner and guest binaries ship in one runner artifact, so the internal migration does not require a dual-codec rollout.
- Backend reserve responses, persisted receipts, queue state, and TypeScript size calculation do not change.
- Unknown-field tolerance must remain covered to avoid a forward-compatibility regression.
- Current validation diagnostic categories must be mapped explicitly so this refactor stays behavior-neutral.
- Encoding and length calculation must use the same private wire representation to avoid escaping or Unicode byte-count drift.
- Valid fixtures should use the shared encoder; raw JSON should remain only for deliberate malformed-input cases.
- The transport maximum must not become a third production constant in the codec.

## Acceptance criteria

- `guest-contracts` is the only production owner of the active-input tag, JSON field schema, codec, and canonical delivery-ID decode rules.
- Runner serialization and size calculation use the shared borrowed codec path; guest-agent deserialization uses the shared owned decode result.
- Duplicate local payload DTOs and discriminant constants are removed.
- Guest runtime policy remains in guest-agent and behaves unchanged.
- Unknown JSON fields remain accepted; missing/wrong fields, wrong tags, empty text, invalid UUIDs, and noncanonical UUIDs remain rejected with the existing diagnostic categories.
- Shared public-boundary tests cover producer-to-consumer round trips, escaped and Unicode text, malformed shapes, wrong tags, canonical UUID rules, empty text, unknown fields, exact length equality, and encodings at exactly and one byte over the real process-control limit.
- Runner CLI exact/oversized behavior and guest queue/deduplication/receipt behavior remain covered.
- Rust formatting, Clippy, documentation, and affected-crate tests pass.

## Delivery notes

This is one independently reviewable PR. No database, API, persisted-state, or rollout migration is required.

seven332:
## Summary

- centralize the runner-to-guest active-input JSON contract in `guest-contracts`
- reuse the shared codec in runner and guest-agent production paths and valid test fixtures
- preserve guest validation diagnostics and forward-compatible unknown-field handling
- cover exact serialization, decoding failures, UUID canonicalization, and the 1 MiB transport boundary

## Scope

This refactor does not change queueing, retry, receipt, or process-control behavior. Transport size enforcement remains owned by runner/process-control; the shared contract only provides exact encoded length and payload encoding/decoding.

## Validation

- `cd crates && cargo clippy --all-targets --all-features`
- `cd crates && cargo doc --workspace --all-features --no-deps`
- `cd crates && cargo test -p runner`

Closes #26893

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -1302,6 +1302,7 @@ name = "guest-contracts"
 version = "0.11.1"
 dependencies = [
  "libc",
+ "process-control-ipc",
  "serde",
  "serde_json",
  "tempfile",
--- a/crates/guest-agent/src/active_input.rs
+++ b/crates/guest-agent/src/active_input.rs
@@ -5,7 +5,7 @@ use std::io;
 use std::path::Path;
 use std::sync::{Arc, Mutex};
 
-use serde::Deserialize;
+use guest_contracts::active_input::{ActiveInputDecodeError, decode_active_input};
 use serde_json::Value;
 use sha2::{Digest, Sha256};
 use tokio::sync::{mpsc, watch};
@@ -15,7 +15,6 @@ use crate::active_input_receipts::ActiveInputReceiptRuntime;
 use crate::error::AgentError;
 use crate::http::HttpClient;
 
-const ACTIVE_INPUT_TYPE: &str = "active-input";
 const ACTIVE_INPUT_QUEUE_CAPACITY: usize = 8;
 const ACTIVE_INPUT_DELIVERY_ID_CAPACITY: usize =
     guest_contracts::active_input_receipts::MAX_ACTIVE_INPUT_RECEIPT_IDS;
@@ -289,15 +288,6 @@ pub struct ActiveInputRuntime {
     writer: ActiveInputWriter,
 }
 
-#[derive(Deserialize)]
-struct ActiveInputPayload {
-    #[serde(rename = "type")]
-    payload_type: String,
-    #[serde(rename = "deliveryId")]
-    delivery_id: String,
-    text: String,
-}
-
 /// Derives the deterministic Claude user-frame UUID for the run's initial prompt.
 ///
 /// Claude stream-JSON stdin writers use this UUID when sending the initial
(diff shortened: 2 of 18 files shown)


seven332:
## Code Review: PR #26950

**Reviewed head:** `029e7eedc6f9ae36070899ddd2c6322a871a417f`

### Summary

The shared codec preserves the existing runner-to-guest bytes and guest validation semantics while removing duplicate wire ownership from both binaries. Runtime queueing, retry, receipt, and transport-limit responsibilities remain in their existing owners.

### Findings

No findings.

### Testing

Rust formatting, workspace Clippy, workspace docs, and the full `guest-contracts`, `runner`, and `guest-agent` test suites passed. Contract coverage includes exact JSON, escaping and Unicode, malformed shapes, error categories, unknown fields, UUID canonicalization, and the exact/one-over 1 MiB boundary. Mixed runner/guest versions were not tested because both binaries ship in one runner artifact.

### Verdict

LGTM

PR #26952, opened 2026-08-13
fix(vsock-host): prevent sequence collisions after wrap

### LINKED ISSUE #26891: bug(rust): prevent live vsock sequence collisions after wrap
## Problem and evidence

The host-side vsock allocator wraps a per-connection `AtomicU32` and skips only zero. Ordinary requests, exec operations, and exec controls obtain a sequence before locking connection state, then insert into separate routing maps without checking the full shared namespace. Same-table insertion can overwrite an owner, while cross-table duplication makes shared response classes such as `MSG_ERROR` ambiguous.

The relevant state is already co-located:

- `ConnectionState::Connected.pending` owns ordinary request routes.
- `Operations.operations` owns active exec operation routes.
- `Operations.control_targets` indexes pending exec-control routes.
- All three are protected by the same connection-state mutex.

There are two ownership races beyond current map occupancy:

- Terminal dispatch removes a route before waking its oneshot receiver. Until the waiting future or handle observes that result, it can still run sequence-only cleanup. If wrap reuses the now-absent sequence during that window, the stale owner can remove or act on the replacement.
- Cancel-on-drop and start-timeout paths can admit an exec cancel frame that remains queued after the operation route is removed. Reusing that sequence before the frame is written could send the old cancellation to a replacement operation.

Reaching the first collision requires a complete non-zero `u32` allocation cycle on one connection, so likelihood is low. The consequence is loss or misrouting of live operation state, and there is no connection-lifetime invariant that makes wrap impossible.

## Decision

Proceed as one `vsock-host` PR. Preserve the `u32` wire protocol and healthy connection while making registrations, owner actions, and admitted queued cancel frames generation-aware.

- Introduce a typed connection-local logical `RouteId` wider than the wire sequence. Its low 32 bits are encoded on the wire; its complete value identifies the registration generation.
- Keep the next logical route value inside connected state. Under the existing state mutex, choose a candidate, skip wire zero and every sequence present in any route class, and insert the route before unlocking.
- Store the complete identity in route entries and cleanup-capable owners. Owner-driven cleanup, cancellation state, stream cleanup, and control targeting must match that identity; incoming guest frames continue to find the one current route by wire sequence.
- Hold a short-lived, generation-aware reservation from cancellation-frame admission through frame write. Allocation skips these reservations, and stale cancellation is suppressed when a newer generation already owns the wire sequence.
- Separate temporary-file suffix generation from protocol route allocation while preserving existing filename behavior.

This uses the synchronization point registrations already require. Active route maps remain authoritative for requests and operations; the small reservation map covers only admitted cancel frames during their queued write lifetime. No guest or wire-protocol compatibility layer is required.

## Alternatives considered

- **Reserve every route for its complete owner lifetime:** prevents reuse while stale owners remain, but duplicates route liveness across one-shot, supervised, control, and close lifecycles. Full `RouteId` matching solves stale owner cleanup more directly; reservations are needed only where an admitted frame outlives map ownership.
- **Retire the connection at wrap:** simpler, but unnecessarily fails every live request and supervised process even when another wire sequence is free, and `VsockHost` does not transparently reconnect.
- **Widen the wire sequence:** expands protocol and independent-deployment scope without eliminating the general stale-owner problem.

## Risks and mitigations

- **Partial identity plumbing:** require typed `RouteId` for every owner-originated state action; use raw `u32` only for wire encoding and incoming dispatch.
- **Queued stale cancellation:** reserve the exact route generation before admitting a cancel frame, retain it until the write future ends, and reject admission when a newer owner already occupies the wire sequence.
- **Lock-order regression:** create channels and normal-operation tracking outside the state lock where possible, and keep registration and reservation transactions synchronous and non-blocking.
- **Logical identity exhaustion:** advance the wider counter with checked arithmetic and return an error rather than silently wrapping it.
- **Timing-sensitive tests:** manually poll public entry-point futures to a registered pending state, let the independent reader dispatch, and deliberately retain the old owner before creating its replacement.
- **Hot-path cost:** reuse the existing single state-lock acquisition; normal allocation adds only constant-time membership checks, with repeated scanning only when a wrapped candidate remains occupied. Reservation work occurs only for cancel frames.

## Acceptance criteria

- Sequence selection and route insertion are one connection-state transaction for ordinary requests, exec operations, and exec controls.
- Allocation skips wire sequence zero, every sequence currently owned by any route class, and every route reserved by an admitted queued cancel frame; no insertion silently overwrites a route.
- Owner-driven cleanup and actions verify the complete logical registration identity, so a completed but not-yet-resumed old future or handle cannot affect a replacement that reused its wire sequence.
- A queued old-generation cancel frame cannot target a replacement operation on the same wire sequence, and a stale cancel handle is suppressed after replacement.
- Integration tests seed the logical allocator around the first `u32` boundary, retain live request/exec/control routes, and verify correct allocation, response delivery, cleanup, cancellation/control targeting, queued-cancel reservation, and replacement ownership.
- Handshake sequence `1`, first post-handshake sequence `2`, the `u32` wire format, and temporary-file naming behavior remain unchanged.
- Rust formatting, workspace Clippy, workspace rustdoc, focused `vsock-host` tests, and the workspace Rust test suite pass without new skips or suppressions.

## Delivery notes

This is one independently reviewable PR. No guest change, public API change, database migration, or staged rollout is required.

seven332:
## Summary

- allocate and register ordinary requests, exec operations, and exec controls atomically from one connection-local route namespace
- retain a full logical route generation for owner-side cleanup and actions while preserving the existing `u32` wire sequence
- reserve admitted queued cancel frames until their write completes so stale cancellation cannot target a replacement route
- separate temporary-file suffix generation from protocol route allocation

## Compatibility

- no guest or wire-format changes
- no public API, database, or migration changes
- handshake sequence `1`, first post-handshake route `2`, and temporary filename formats remain unchanged

## Validation

- `cargo check -p vsock-host --all-targets --all-features`
- `cargo clippy -p vsock-host --all-targets --all-features -- -D warnings`
- `cargo doc -p vsock-host --all-features --no-deps`
- `cargo test -p vsock-host --quiet`
- six wrap and stale-owner integration tests repeated 20 times each
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS='-D warnings' cargo doc --workspace --all-features --no-deps`
- `cargo test --workspace --all-targets --all-features`

Closes #26891

### DIFF
--- a/crates/vsock-host/src/connection/listener.rs
+++ b/crates/vsock-host/src/connection/listener.rs
@@ -96,8 +96,10 @@ impl VsockHost {
             frame_builder: tokio::sync::Mutex::new(()),
             file_write_gate: tokio::sync::Mutex::new(()),
             fd,
-            seq: AtomicU32::new(2),
+            temp_seq: AtomicU32::new(2),
             state: std::sync::Mutex::new(ConnectionState::Connected {
+                next_route_id: 2,
+                route_reservations: HashMap::new(),
                 pending: HashMap::new(),
                 operations: exec_operation::Operations::new(),
             }),
--- a/crates/vsock-host/src/connection/mod.rs
+++ b/crates/vsock-host/src/connection/mod.rs
@@ -30,9 +30,9 @@ const READ_BUF_SIZE: usize = 64 * 1024;
 
 /// Connection lifecycle, expressed as data rather than a separate atomic flag.
 ///
-/// Pending requests, active exec operations, and connected process state
-/// live inside the `Connected` variant so registrations are structurally
-/// unreachable once the reader task has exited.
+/// Pending requests, active exec operations, queued-frame route reservations,
+/// and connected process state live inside the `Connected` variant so new
+/// registrations are structurally unreachable once the reader task has exited.
 ///
 /// The invariant "connection is closed ⇔ registrations are impossible" is
 /// enforced by the type: every code path that cares about liveness must
@@ -40,6 +40,10 @@ const READ_BUF_SIZE: usize = 64 * 1024;
 /// without taking the corresponding lock.
 pub(super) enum ConnectionState {
     Connected {
+        /// Next logical route identity. Its low 32 bits are used on the wire.
+        next_route_id: u64,
+        /// Routes held by frames admitted for a future write.
+        route_reservations: HashMap<u32, RouteReservationEntry>,
         /// Pending request responses: seq → response route.
         pending: HashMap<u32, PendingResponse>,
         /// Active exec-operation state owned by the exec_operation module.
@@ -60,10 +64,8 @@ pub(super) struct Shared {
     pub(super) file_write_gate: tokio::sync::Mutex<()>,
     /// Raw fd of the underlying socket, used to poison a corrupted stream.
     pub(super) fd: RawFd,
-    /// Monotonically increasing sequence number (starts at 2, skips 0).
-    /// Handshake uses seq=1 before Shared is created, so post-handshake
-    /// sequences start at 2 to avoid collisions.
-    pub(super) seq: AtomicU32,
+    /// Counter used only to distinguish temporary file names.
+    pub(super) temp_seq: AtomicU32,
     /// Single source of truth for connection liveness plus all per-connection
     /// registration tables. See [`ConnectionState`].
     pub(super) state: std::sync::Mutex<ConnectionState>,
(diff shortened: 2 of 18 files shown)


seven332:
## Code Review: PR #26952

**Reviewed head:** `29201705de02275bc32cabe8d438402cae29654a`

### Summary

The change makes vsock route registration generation-aware and atomic across ordinary requests, exec operations, and exec controls. It also protects queued cancellation frames with bounded RAII reservations while preserving the existing guest wire protocol.

### Findings

No findings.

### Testing

Rust formatting, focused and workspace Clippy/rustdoc, 352 focused `vsock-host` tests, repeated wrap/race regression tests, and the full Rust workspace test suite passed locally. Required GitHub checks are still pending.

### Verdict

LGTM

PR #26829, opened 2026-08-13
feat(billing): redesign plan selection as a single comparison panel

Lunarivibe:
## What

Redesigns both steps of Billing → **Choose a plan** so they speak one language: a single surface, three columns, and every price on the same right edge.

### Step 1 — plan selection

Two floating cards become **one bordered panel split by a hairline**, so every horizontal band (name, description, SKU rows, total, CTA, feature list) lands on the same baseline across both columns.

| | |
|---|---|
| Before | before |
| After | after |

- **The price is decomposed.** Each column lists the real purchase as right-aligned SKU rows — `Plan` and `Member packages` (with the credits that package buys) — closing on a hairline-framed **Monthly total** band. This replaces the 11px `$0 plan + $20 required member package` line, which was the only place the composition of `$20` / `$180` was explained.
- **Team stops repeating Pro.** Three of the five feature rows were identical; Team now lists only its deltas under `Everything on Pro, plus:`.
- **Popular badge removed.** Pro keeps the emphasis through the filled primary button; Team moves to the `outline` variant.
- **Step bar replaced** by a compact `Step 1 of 2` indicator in the page header.
- Prices use `tabular-nums` so the right edges hold across `$0` / `$160` / `$20` / `$180`.

### Step 2 — member packages

The plan summary card, the member list card and the order summary box collapse into **one borderless ledger** on three columns: line item, package select, price.

step 2

- The plan is the ledger's first row and the monthly total its last, so plan price, per-member price and total all share one right edge. `Concurrent slots` moves into the plan row's summary line; `Total credits` and `Bonus credits from discount` fold into the total row.
- `SelectedPlanSummary` is gone — the header's back button was already the way back to step 1, so the `Change plan` button was a duplicate entry point.
- The package `Select` drops the `bg-background` override and uses its own component styling (white `bg-input`, 0.7px border, `text-sm`). Its label drops the price it duplicated with the new price column, and the column widened to keep the longest catalog label (`200,000 credits · 23% off`) on one line.
- Rows are separated by a faint `gray-100` rule instead of card borders; only the total sits behind a `border` rule.

## i18n

[...]

PR #26955, opened 2026-08-13
feat(billing): rebuild the package change review dialog as a ledger

Lunarivibe:
Rebuilds the **Review package change** dialog — the last screen before a workspace is charged — on the ledger vocabulary that the plan and package steps already use.

Design was reviewed as rendered comps before any code was written; Ming picked this direction ("only the charge is large") from a set of nine explorations.

| Before | After |
|---|---|
| before | after |

## What was wrong

1. `ORDER SUMMARY` was a CSS `uppercase tracking-wide` heading, which the UI rules forbid on section headers.
2. Two unrelated questions sat at the same visual weight: *what does confirming charge me right now* and *what does my subscription become*. Six equal rows, no lead.
3. Bonus credits were printed twice with two different meanings — the one-off grant and the recurring monthly bonus — usually with the same value.
4. The numerals used `text-primary`, which spends the brand colour on data. The plan and package steps render every total in ink and keep orange for the action.

## What changed

- The dialog is now two **named** blocks — **Today** and **Every month** — so the two large numerals answer two stated questions instead of competing. A `gray-200` rule opens each block, `0.7px gray-100` hairlines separate rows inside it, so rule weight carries the structure. Amounts are `tabular-nums`, right-aligned to a single edge.
- **Exactly three type levels**, which is what the first pass got wrong: block name 12px medium muted, row label 14px medium ink, qualifier 11px muted on its own line. Previously a section heading (13px muted) and a data row (14px muted) were nearly indistinguishable.
- **A qualifier never shares the label's line.** One middot cannot mean "expires", "includes" and "consists of" at once, so qualifiers moved under the label — the same shape `MemberIdentity` and `LedgerPlanRow` already use.
- **Due now** is the only large numeral in the charge block (`text-3xl font-light tracking-tight`), with the credit grant as an ordinary row beside its expiry.
- The recurring block collapses six rows into three plus the total: plan, member packages, and credits. Purchased/bonus are no longer split here — that breakdown lives on the usage page.
- All numerals are `text-foreground`. The only brand colour left is the confirm button.
- `LedgerPrice` gained an optional `fractionDigits` so the plan-conversion dialog can still show `$229.50/month`.

`SubscriptionOrderSummary` was only a label plus a passthrough, so it is gone — the label now belongs to the block it names. `ManagedSubscriptionSummaryDetails` is shared with the **Review plan conversion** dialog, so that screen inherits the same treatment.

## Copy

Five deviations from the approved comp, each for a reason:

| Comp | Shipped | Why |
|---|---|---|
| Prorated for the rest of this billing period | For the rest of this billing period | The preview does not guarantee the amount is a proration |
| 2 concurrent slots | 2 concurrent runs | Reuses the shipped `billing.concurrency.concurrentRun` string |
| Member packages · 3 members | Member packages | A member count would need a new pluralised key in ten locales for decoration |
| Your subscription from Sep 1, 2026 | Every month | The next billing date is not in the dialog's data, and the new price is effective immediately — Sep 1 is only the next invoice |
| $90/mo | $90/month | The real `billing.plans.perMonth` value the page ledger renders |

Five new keys (`credits`, `bonusIncluded`, `management.today`, `management.everyMonth`, `management.immediateAmountNote`) and a shortened `management.reviewDescription`, translated across all ten locales.

## Not in this PR

- `UsagePackPaymentSummary` still renders its numerals in `text-primary`. It is also mounted by the member-invite flow in `org-members-tab.tsx`, so restyling it would change a screen that has not been reviewed. Worth a follow-up.
- `SubscriptionComparisonTable` (the Current/New table on the configure step) still uses an uppercase `thead` and a bordered card. Same class of problem, separate change.

## Base branch

Stacked on `feat/billing-plan-selection-panel` (#26829) rather than `main`, because the ledger vocabulary this dialog matches only exists there. Kept as its own PR so #26829 does not grow a fourth block; GitHub will retarget this to `main` once #26829 merges.

## Verification

No local `node_modules` in this environment, so `vitest` could not run — syntax checked with `esbuild`, formatting with `prettier`, and the dialog rendered from the final class names to diff against the approved comp. The updated assertions in `org-billing-tab.test.tsx` and the rest of the checks rely on the PR pipeline.


Lunarivibe:
### CI status

54 pass, 23 skipped, 2 fail. Everything touching this diff is green: `lint-eslint`, `lint-format`, `lint-types`, `lint-types-apps`, `lint-knip`, all eight `test-app` shards (the billing suite is shard 3), and `test-cli`.

The two failures are `deploy-cli` and `ci-gate-turbo`, and the gate only fails because it mirrors `deploy-cli`.

`deploy-cli` dies at `Install CLI dependencies` / `Build CLI`:

```
cd turbo && pnpm install --frozen-lockfile --filter @vm0/okou-cli...
No projects matched the filters in "/__w/vm0/vm0/turbo"
```

so nothing is written to `turbo/apps/cli/dist` and the artifact step then fails on a missing directory.

Evidence that this is not from this branch:

- The diff is 12 files, none under `turbo/apps/cli`, `.github/`, or workspace config, and the commits contain zero deletions.
- `turbo/apps/cli/package.json` (`@vm0/okou-cli`, 322 files) and `turbo/pnpm-workspace.yaml` (`apps/*`) are byte-identical to the base branch, and the filter string in `turbo.yml` matches the package name exactly.
- `test-cli` passed on the same run, so the CLI workspace resolves fine in other jobs.
- The same job passed on #26829 fifteen minutes before the first run here.

Reproduced across two runs, so it is not a one-off. I stopped after the second rather than keep re-running. Worth a look from someone who owns the pipeline — I did not touch CI config here.

Lunarivibe:
LGTM


PR_AUTO_REVIEW_COMMIT: e34d48a96e936d1e676c6c8739865c2dcd0ae49e
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Presentation-only rebuild of the package-change review dialog in `turbo/apps/platform`. The flat "Order summary" list becomes two named blocks — `Today` (`UsagePackChangeCharges`: the immediate charge and the credits it buys) and `Every month` (`ManagedSubscriptionSummaryDetails`: plan, member packages, credits, monthly total) — built from four new primitives (`ReviewSectionLabel`, `ReviewRowLabel`, `ReviewRow`, `ReviewTotalRow`) that fix three type levels and use rule weight (`gray-200` opens a block, `gray-100` separates rows) to carry structure. `SubscriptionOrderSummary` is deleted; both of its call sites move to `ManagedSubscriptionSummaryDetails` directly. No API, contract, schema, signal, or persisted-state change: 1 component file, 1 existing integration test file, 10 locale bundles.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic — no wrong conditionals or unhandled paths | Pass | `UsagePackChangeCharges` returns `null` when there is neither an immediate amount nor a grant, so the `Today` block never renders empty; `rule={hasImmediateAmount ? "hairline" : "strong"}` correctly promotes the grant row to the block-opening rule when it is the only row. `ReviewTotalRow` keeps the existing `Number.isInteger(monthlyTotalUsd) ? 0 : 2` precision rule, so `$229.50/month` still renders two digits. |
| `REVIEW.md` | Security — no injection, secrets, or missing auth | N/A | Client-side rendering only; no new endpoint, credential, or raw HTML. |
| `REVIEW.md` | Type safety — no `any`, no broad casts | Pass | No `any` and no `as` cast added. `fractionDigits?: number` and `rule?: "hairline" \| "strong"` are concretely typed; conditional props are spread as `{...(x === undefined ? {} : { meta: … })}` so `exactOptionalPropertyTypes` stays satisfied without a cast. |
| `REVIEW.md` | Style & maintainability — function size, duplication, WHY comments | Pass with nit | The three new block comments explain WHY (why a dialog drops the control column, why one middot cannot mean three things, why the total carries the page ledger's weight). Deleting `SubscriptionOrderSummary` removes a component whose whole body was a heading plus one forward. Naming nit below. |
| `REVIEW.md` | Feature Regression Guard — replaced behavior is scoped and covered | Pass | The surface stays behind `FeatureSwitchKey.UsagePackPlans` (`feature-switch.ts`, `enabled: false`, no staff hashes); `usagePackMigrationAsync$` returns `null` when the switch is off, so the migration preview path is gated too. Every replaced assertion is rewritten to the new shape in `org-billing-tab.test.tsx` — the conversion review dialog, the package-change dialog, and the two monthly-total assertions at lines 2262 and 2383. `UsagePackPaymentSummary` is untouched and still serves `org-members-tab.tsx:621`, so that dialog does not regress. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No `packages/db/src/schema/**`, contract, Zod, or `jsonb(...)` change. |
| `docs/docs.md` | Route the review to the practice docs matching the changed surface | Pass | Platform/React surface → `effect.md`, `cache.md`, ccstate SKILL, testing suite; no event-sourcing, chat-card, or runner-arch surface touched. |
| `docs/bad-smell.md` | Zero lint/type suppressions | Pass | No `eslint-disable`, `oxlint-disable`, `@ts-ignore`, `@ts-expect-error`, or `prettier-ignore` anywhere in the diff. |
| `docs/bad-smell.md` | Fail fast — no defensive try/catch or silent defaults | Pass | No try/catch added. `fractionDigits = 0` is a presentational default on a new optional prop that preserves the previous call sites' behavior exactly, not a guard against a missing value. |
| `docs/bad-smell.md` | No dynamic `import()` | Pass | Static imports only; the diff adds no import. |
| `docs/bad-smell.md` | No hardcoded URLs or env-specific values | Pass | Tailwind classes and i18n keys only. |
| `docs/fallback.md` | §3 — no `??`/`\|\|` on contract-guaranteed fields | Pass | The diff adds **zero** `??` or `\|\|` operators. The two conditional branches guard fields the contract declares optional: `zero-billing.ts:296-303` has `immediateCreditGrant: z.object({…}).optional()` and `expiresAt: z.iso.datetime().optional()`, so this is §6.2/§6.3 optional-data handling, not a fallback. `totals.bonusCredits > 0` branches on a real product state (a pack with no discount). |
| `docs/fallback.md` | §2 — non-GA feature-switched code needs no compatibility path | Pass | `UsagePackPlans` is off for every org, so no dual-read, migration, or old-shape tolerance was required and none was added. No `Fallbacks` section is needed in the PR summary. |
| `docs/fallback.md` | §1 — no tombstone tests asserting removed behavior stays removed | Pass | The two negative assertions in the diff's context (`queryByText("Next recurring total")`, `queryByText("Total credits")`) are pre-existing lines, not added here; the diff adds no new negative assertion. |
| `docs/deployment-compatibility.md` | Preserve cross-version compatibility on deployable/persisted boundaries | N/A | No request/response, runner protocol, queue payload, migration, or persisted-state change. Frontend-only behind a non-GA switch, so `fallback.md` §7's client-skew window creates no obligation. |
| `docs/effect.md` | No side effects in render; commands carry business semantics | Pass | All four new primitives and `UsagePackChangeCharges` are pure render functions — no `useEffect`, `detach`, `onRef`, or command added. |
| `docs/cache.md` | RC-1/RC-2/RC-5 — no render-time signal allocation or duplicate sources of truth | Pass | No signal created during render. `monthlyTotalUsd` keeps its single derivation (`monthlyTotalCents / 100` when supplied, else `plan.basePriceUsd + totals.totalUsd`), so the dialog and the page ledger cannot diverge. |
| `.claude/skills/ccstate/SKILL.md` | Read signals via `useGet`/`useSet`/`useLoadable*`, never the store directly | Pass | The diff introduces no signal, `computed`, or `command`, and touches no existing accessor. |
| `docs/testing.md` | Integration tests only; mock at the boundary; real infrastructure | Pass | The one changed test file is an existing platform integration test driven through the rendered page. No `vi.mock("../../…")`, no `vi.stubGlobal("fetch")`, no fake timers, no new mock introduced. |
| `docs/testing.md` | New user-facing features gated behind a `FeatureSwitchKey` | Pass | `FeatureSwitchKey.UsagePackPlans`, `enabled: false`. |
| `docs/testing/patterns.md` | Pattern 7 — platform page tests drive the page and assert on the page | Pass | Cases are reached by clicking through to the dialog (`findByRole("dialog", { name: "Review package change" })`) and asserted via `getByText` on rendered copy. |
| `docs/testing/anti-patterns.md` | AP-7/AP-10 — no implementation-detail assertions, no direct component rendering | Pass with nit | No component rendered directly, no CSS class asserted. `.closest("div")` is a DOM-structure escape hatch — see Findings; it replaces `.parentElement`, which was equally coupled, so this is not a regression. |
| `docs/testing/app-testing.md` | Page tests exercise the real page through MSW handlers and `setupPage` context | Pass | Existing handler/`setupPage` infrastructure reused unchanged; only expected strings and row scoping were updated. |
| `docs/testing/testing-external-behavior.md` | Platform boundary is the page — no store mutation, hook calls, or cache assertions | Pass | No store mutation or hook call; assertions read text the user can see. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

#### Nits (non-blocking)

- `org-billing-tab.test.tsx` — the new row assertions use `getByText(label).closest("div")`. That resolves to the `ReviewRow` container today only because `ReviewRowLabel` emits `span`s; adding any wrapper `div` inside `ReviewRow` would silently retarget the assertion at a narrower node and it would still pass or fail for the wrong reason. Giving `ReviewRow` a `role="group"` + accessible name (as the old credit-grant block had) and scoping with `within(...)` would make the coupling explicit.
- `usage-pack-pricing-page.tsx:ManagedSubscriptionSummaryDetails` — the name no longer describes the component. It is now the shared `Every month` block for the conversion dialog, the package-change dialog, and the migration preview, only one of which is a managed subscription. `EveryMonthBlock` (or similar) would match what it renders.
- `usage-pack-pricing-page.tsx:SubscriptionComparisonTable` is now the only element inside these dialogs still using the old language — uppercase tracked header plus a `border-border/70` card outline — against the three type levels this PR establishes a few rows above it. The PR description already calls this out as follow-up work; worth linking a follow-up issue so it does not settle as the permanent exception.

### Testing

- Coverage: Adequate. `feat:` commit with integration coverage in `turbo/apps/platform/src/views/zero-page/__tests__/org-billing-tab.test.tsx`. The rewritten assertions cover both new blocks end to end: the `Today` block (block label present, `Due now` row carrying its `For the rest of this billing period` qualifier and `$15.00`, credit-grant row carrying `Expires Apr 1, 2026` and `+16,100`) and the `Every month` block (block label, plan row carrying the `10 concurrent runs` qualifier and `$160`, `Member packages` `$70`, `Credits` row carrying `5,555 bonus included` and `75,555`, and `Monthly total` at `$229.50/month`). The fractional-total path is exercised by that `$229.50` case, which is what the new `fractionDigits` prop on `LedgerPrice` exists to serve. Two further dialogs (`$180/month`, `$20/month`) keep their totals asserted through the new `Monthly total` row. No new test file is needed: no new page or flow was added, the same flows are asserted against the new presentation.
- Conventions: Compliant. No relative-path `vi.mock()`, no `fetch` stubbing, no fake timers, no `expect(mock).toHaveBeenCalled()` as a sole assertion, no mock-cleanup regression.
- CI on `e34d48a9`: 50 checks pass, 0 fail, 0 pending — including `lint-eslint`, `lint-types`, `lint-knip`, `i18n:check`, `deploy-cli`, and the `test-app` shards. Note for history: every earlier run on this branch was against a stale head. The PR was opened against base `feat/billing-plan-selection-panel`, which was deleted when #26829 merged, so GitHub could not compute a merge ref and `pull_request` workflows never fired on the newer commits. Retargeting the base to `main` and merging `origin/main` in (`d17dfa7`, `e34d48a9`) is what produced this first complete run.

PR #26689, opened 2026-08-12
refactor: retire axiom-backed log surfaces

e7h4n:
## Summary

- replace `okou search --source logs` with `--source agent-session`, which prints both Claude Code and Codex session directories for direct file analysis
- remove the `okou logs` and `okou developer-support` commands
- remove the current App/CLI agent-event search/read, system-log, metrics, diagnostic bundle, and web Error Report surfaces; temporarily retain the authenticated backend routes listed under Fallbacks for stale clients
- keep Activity metadata, Context, Runner, Network, and JSON download; live Activity no longer reads or polls agent events
- make `run-context` and `sandbox-telemetry-network` the only Axiom reads used by current first-party surfaces; both require `agent-run:read` and verify `runId + userId + orgId` ownership
- leave connector check behavior unchanged: it has no historical run selector and only uses the current Zero token run context when run-scoped
- remove the now-unused Plain support SDK and CLI event parsing/rendering implementation

## Fallbacks

- `zeroDeveloperSupportRoutes`: retains the retired developer-support endpoint for immutable CLI packages selected by pre-deployment execution contexts. Maximum drain is the 15-minute queued-run TTL plus the runner's 2-hour execution timeout, 90-second finalization grace, and 10-second terminal grace (about 2 hours 17 minutes). Remove under #26735 after that drain and the rollback window close.
- `zeroLogsRoutes` search entry: retains the retired log-search endpoint for the same immutable CLI package window (about 2 hours 17 minutes). Remove under #26735 after all pre-deployment contexts drain and the rollback window closes.
- `zeroReportErrorRoutes`: retains the retired Error Report endpoint for already-open App builds during the observed approximately two-day stale-client window. Remove under #26735 after the compatibility floor has advanced and the rollback window closes.
- `zeroRunDetailRoutes` agent-events, system-log, and metrics entries: retain the retired telemetry APIs for stale App clients (approximately two days) and immutable pre-deployment CLI contexts (about 2 hours 17 minutes). Remove under #26735 after both drains and the rollback window close.

## Validation

- `pnpm format`
- `pnpm turbo run lint --concurrency=1 --log-order grouped`
- `pnpm knip`
- `pnpm --filter @vm0/api-contracts check-types`
- `pnpm --filter @vm0/core check-types`
- `pnpm --filter @vm0/okou-cli check-types`
- `pnpm --filter api check-types`
- `pnpm --filter @vm0/app check-types`
- CLI search/session, command registration, and capability tests: 96 tests passed across 4 targeted files
[...]

PR #25563, opened 2026-08-07
feat(chat): prefetch unread and active thread events

e7h4n:
## Summary

- add an organization-scoped API for all unread chat thread IDs
- start initial unread/active thread event hydration concurrently with the Ably background subscription
- deduplicate overlapping thread IDs and reuse the existing new-message sync path
- cover unread lifecycle semantics plus subscription concurrency and deduplication

## Testing

- `pnpm -F @vm0/app exec vitest run src/signals/chat-page/__tests__/chat-event-background-sync.test.ts`
- `pnpm -F @vm0/api-contracts check-types`
- `pnpm -F @vm0/app check-types`
- `pnpm -F api check-types`
- `pnpm -F @vm0/api-contracts lint`
- `pnpm -F @vm0/app lint`
- `pnpm -F api lint`

The targeted API BDD test could not start locally because PostgreSQL was unavailable on `localhost:5432`; the PR pipeline will run it with the test database.

PR #27004, opened 2026-08-13
fix(platform): restore activity event log

### LINKED ISSUE #27003: bug(platform): restore activity event log
## Background

PR #26689 retired Axiom-backed log surfaces, but it also removed the event log from the Activity detail page. The page no longer loads or polls agent events, the Steps tab is gone, and downloaded Activity JSON receives an empty event list.

The Activity event log is a separate first-party run surface and should remain available without restoring the retired Axiom-backed CLI, diagnostics, system-log, or metrics surfaces.

## Requirement

Restore the Activity detail event log as closely as possible to its behavior immediately before PR #26689, adapting only where the current code requires it.

## Acceptance criteria

- Activity detail pages display the event/step log for existing and live runs.
- Event grouping, search, pagination, polling, and scroll behavior are preserved where supported by the current API and frontend architecture.
- Context, Runner, and Network debug tabs continue to work.
- Activity JSON downloads include loaded agent events again.
- Retired Axiom-backed CLI, diagnostic, system-log, metrics, and Error Report surfaces are not restored.
- Integration tests cover initial loading, live polling/watermark behavior, pagination, stale-run isolation, and retained Activity data.

## Relevant links

- Removal: #26689
- Previous closed restoration attempt: #26776

seven332:
## Summary

- restore the Activity page Steps tab from persisted agent events
- add owner-scoped, sequence-paginated agent event reads from Axiom
- poll incrementally from the contiguous event prefix until the terminal watermark is visible
- bound permanently incomplete terminal histories and keep polling owned by the Activity page lifecycle
- preserve the original Steps layout, initial scroll position, search behavior, and raw download format
- reuse the shared event contract without repeating framework metadata reads for every event page

## Scope

This restores only the Activity page event log. It does not restore the retired CLI log commands, Error Report, diagnostics, system log, or metrics surfaces.

## Fallbacks

- `activity-signals.ts:fetchAgentEventPage` accepts `404` when a newly promoted app reaches an API version that predates the additive Activity event route. Surface: app/API rollout and rollback, ~2 days. Remove after that API is outside the production rollback window; follow-up #27010.

## Validation

- `pnpm -F api exec vitest run src/signals/routes/__tests__/run-reads.bdd.test.ts`
- `pnpm -F @okouai/app exec vitest run src/views/activity-page/__tests__/activity-paged-events.test.tsx src/views/activity-page/__tests__/activity-detail-retained-data.test.tsx`
- `pnpm -F api lint`
- `pnpm -F @okouai/app lint`
- `pnpm knip` (pre-commit)
- `pnpm check-types` (pre-commit)

Closes #27003


seven332:
## Code Review: PR #27004

**Reviewed head:** `7b884ae672f898c2f384d3919d47145f09329838`

### Summary

The PR restores only the Activity event surface, uses an additive owner-scoped API, retains sequence-safe pagination, and gives live polling a stable abortable lifecycle with a terminal event watermark.

### Findings

No findings.

### Testing

Focused API and Activity page integration tests passed, as did affected API, Platform, and API-contract lint. Full Knip and TypeScript checks passed through pre-commit. Repository CI is still the remaining broad verification surface.

### Verdict

LGTM

seven332:
## Review: LGTM

Reviewed the current head `51e0e07a1f` against the Activity detail implementation immediately before its removal.

No findings. The restored UI and observable behavior now preserve the prior Steps layout, default tab, search and event rendering, initial scroll-to-bottom behavior, and raw download filename/payload shape. The intentional differences are limited to current package/API integration and terminal-event watermark polling, which fixes the prior live-refresh limitation.

Focused Platform tests, lint, type checking, Knip, formatting, and commit validation passed locally.

e7h4n:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 51e0e07a1f022d68fbc751578bd9cbf47c77bc08

### Summary

Restores the Activity detail Steps tab by adding an owner-scoped, sequence-paginated `GET /api/okou/runs/:id/telemetry/agent` reader over the `agent-run-events` Axiom dataset, re-adds event grouping/search/scroll on the frontend, polls the selected run until its terminal event watermark, and puts events back into the Activity JSON download. Ownership scoping, cursor validation, and APL escaping all look correct, and the auth matrix is covered. The blocking concerns are all in the new polling loop: its termination condition can be permanently unsatisfiable, and each tick re-reads the entire event history with the Axiom cache disabled.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no unhandled edge cases | Finding (P0) | `activity-signals.ts:144` `reachedTerminalEventWatermark` can never become true when the fetched history is not contiguous `0..lastEventSequence`; `setLoop` then never returns. |
| `REVIEW.md` | Security: auth checks on new endpoints | Pass | `zero-run-detail.service.ts:190-200` scopes by `agentRuns.userId` + `orgId`; `run-reads.bdd.test.ts` covers 401 (anon/orgless) and 404 for a same-org non-owner member. |
| `REVIEW.md` | Type safety; no `any` | Pass | No `any` introduced. `zero-run-detail.service.ts:39` uses a narrowing `as AgentComposeContent \| null \| undefined` on a jsonb column, restored verbatim from the pre-#26689 file. |
| `REVIEW.md` | No lint/type suppressions | Pass | No new `eslint-disable`, `@ts-ignore`, or `@ts-expect-error` in the diff. |
| `REVIEW.md` | Feature regression guard; existing flows covered, scoped or gated | Finding (P1) | `zero-activity-detail-page.tsx:1106` gates the whole page on `events === null`, so a failed events read blanks Context/Runner/Network too. Steps being ungated matches pre-#26689 behavior (`c7266ba~1` line 969), so no new feature switch is required. |
| `REVIEW.md` | `fix:` commits carry a regression test | Finding (P0) | Restore paths are covered by `activity-paged-events.test.tsx`, but the new watermark termination guarantee has only a happy-path test; the unsatisfiable-watermark path has none. |
| `REVIEW.md` | DB/JSONB review gate | Pass | No schema or migration change; `agentComposeVersions.content` is read-only here and the response is validated by `agentEventsResponseSchema`. |
| `docs/bad-smell.md` | Zero tolerance for `any`, suppressions, dynamic imports | Pass | None present; all imports in the diff are static. |
| `docs/bad-smell.md` | No hardcoded URLs/config; fail fast | Pass | Dataset name resolved via `getDatasetName`; no hardcoded endpoints added. |
| `docs/fallback.md` | Every fallback introduced by the diff is declared in the PR summary | Finding (P1) | `zero-run-detail.service.ts:38-50,208` re-introduces `runContextCliAgentType() ?? extractFramework()` ending in `?? "claude-code"`; the PR summary has no `Fallbacks` section. |
| `docs/fallback.md` | Additive cross-version route needs a time-boxed tolerance branch (§8) | Finding (P1) | `activity-signals.ts:55-95` calls the new additive route with `accept(..., [200])` only; a newly promoted app hitting the pre-route API gets 404 and the page sticks on the skeleton. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | The diff converts the old "no event stream" assertions into positive assertions rather than adding tombstones. |
| `docs/effect.md` | Long-running/polling work belongs to route or page setup, not a DOM ref | Finding (P1) | `activity-signals.ts:154,186` wraps a polling loop in `onRef` with an unused `_element`, and `zero-activity-detail-page.tsx:1105` adds a `className="contents"` wrapper purely to host it. |
| `docs/effect.md` | Every async process has an `AbortSignal` with an eventual abort path | Pass | The loop receives the `onRef` detach signal; `fetchAllAgentEvents` threads `pageSignal$` into `fetchOptions.signal`. |
| `docs/effect.md` | Keep React render pure | Pass | Render only reads loadables and passes the stable `useSet` ref through; no Store writes or timers during render. |
| `docs/cache.md` | RC-7: stable callback ref, cleanup preserved | Pass | `useSet(activityEventsPollerRef$)` is passed directly to `ref`, not wrapped in an inline arrow. |
| `docs/cache.md` | RC-8: symmetric resource lifecycle for long-running work | Pass | `setLoop` terminates on the ref's abort signal at unmount. |
| `docs/cache.md` | RC-3: no unbounded lifetime caches | Pass | `seenCursors` is scoped to a single `fetchAllAgentEvents` call; no module-level keyed cache added. |
| `docs/deployment-compatibility.md` | Additive boundary changes must tolerate the old peer during the rollout window | Finding (P1) | Same as the `docs/fallback.md` §8 row: new app + pre-promotion API is an untolerated combination on `GET /api/okou/runs/:id/telemetry/agent`. |
| `docs/event-sourcing.md` | Optimistic/persistent event reconciliation rules | N/A | The Activity reader consumes only persisted Axiom events; the diff creates no optimistic events and no page-local event buffer. |
| `docs/testing.md` | Integration tests, mocked only at the system boundary | Pass | API test drives real routes with a real DB and mocks only Axiom; platform tests use `mockApi`/MSW with no `vi.mock` of internal paths. |
| `docs/testing.md` | No fake timers | Pass | No `vi.useFakeTimers`; `setLoop` yields to a macrotask under `IN_VITEST`, so the 1s interval is not slept in tests. |
| `docs/testing/anti-patterns.md` | AP-1: test outcomes, not call patterns | Pass | `run-reads.bdd.test.ts` asserts full response bodies; the `context.mocks.axiom.query` assertions verify the outbound APL contract at an external boundary rather than substituting for behavior. |

### Findings

- `turbo/apps/platform/src/signals/activity-page/activity-signals.ts:130-152,175` — the poll loop can never terminate when the event history is not contiguous. `visibleEventSequence` walks the fetched events and stops at the first gap, returning `-1` if the first event is not `sequenceNumber === 0`. `reachedTerminalEventWatermark` then requires `visibleEventSequence(events) >= lastEventSequence`, so for a terminal run whose Axiom history is short or has a hole, the predicate is permanently false and `setLoop` never returns `true`. This is reachable because `agent_runs.last_event_sequence` and the `agent-run-events` Axiom rows are written by two independent, non-atomic paths: `agent-webhook-complete.service.ts:162` persists the watermark via `greatest(coalesce(...), ...)` on completion, while `agent-event-consumer-axiom.service.ts` ingests the rows separately and throws on ingest failure. Any permanently failed ingest batch, or any older run whose events have aged out of the dataset while the DB row keeps its watermark, leaves the two permanently out of sync. There is no iteration or time bound outside tests — `MAX_LOOP_COUNT_IN_TEST` in `utils.ts:429` is `IN_VITEST`-only — so the page polls at 1 Hz for as long as it stays open, and each tick pays the full cost described in the first P1 below. Please add a bound (give up after the watermark fails to advance for N ticks, or treat "no new events across consecutive polls on a terminal run" as done) and cover it with a test that returns a terminal run whose events never reach `lastEventSequence`.

- `turbo/apps/platform/src/signals/activity-page/activity-signals.ts:55-95,104-118,154-183` — every poll tick refetches the entire event history with the Axiom cache disabled. `pollActivityEvents$` bumps `internalActivityEventsReload$`, which invalidates `zeroActivityEvents$`, and `fetchAllAgentEvents` always restarts from page 1 with no `since` and no `cursor`. Each page is a separate Axiom query issued with `{ noCache: true }` (`zero-run-detail.service.ts:224`), so a live run with N events costs `ceil(N/100)` sequential uncached Axiom queries **per second per open page**. The implementation being restored did the opposite: `polling.ts` (at `c7266ba~1`) appended incrementally via `since`/`nextCursor` and carried an explicit comment that the tail page "doesn't reset `since` to undefined and refetch from the top". The `since` sequence parameter this PR ships in `createLogPaginationQuerySchema` and `buildAgentEventPaginationFilters` has no client caller at all — the PR's own test at `activity-paged-events.test.tsx:1322` records `since: undefined` for every request. Please poll incrementally from the last known sequence and append, so steady-state cost is one small query per tick.

- `turbo/apps/platform/src/signals/activity-page/activity-signals.ts:154,186` and `turbo/apps/platform/src/views/zero-page/zero-activity-detail-page.tsx:1105` — `onRef` is used purely as a mount trigger for a business process. `docs/effect.md` states that long-running processes that must stop on navigation belong in the route or page setup command, and explicitly warns "do not use `onRef` merely because it provides a mount signal. If the element parameter is unused, prove that the behavior is still about the committed presence of that DOM subtree." Here `pollActivityEvents$` takes `_element: HTMLElement` and never uses it, and a `<div className="contents">` wrapper exists only to host the ref. This is not cosmetic: `ZeroActivityDetailPage` returns early with `<ActivitySkeleton />` at line 1099 while `zeroActivityDetail$` is loading, so the ref is not mounted and polling does not start until the detail request resolves — exactly the conditional-rendering coupling the doc warns about. Please move the loop to the Activity page setup command with the page signal as owner.

- `turbo/apps/platform/src/signals/activity-page/activity-signals.ts:55-95` — the new additive route is consumed without rollout tolerance. `fetchAllAgentEvents` calls `accept(..., [200])`, and `accept` throws `ApiError` for anything else, so a newly promoted app talking to an API that predates `GET /api/okou/runs/:id/telemetry/agent` gets a 404, `zeroActivityEvents$` rejects, and `events` stays `null`. `docs/fallback.md` §8 documents this exact case as one where a time-boxed fallback is required, citing PR #25563 accepting the old API's `404` on an additive route. Please accept `404` with a comment naming the surface and window (old web/app clients, ~2 days for the reverse direction; API rollback window for this one) plus a removal follow-up, per §8.

- `turbo/apps/platform/src/views/zero-page/zero-activity-detail-page.tsx:1090-1113` — one failing telemetry read blanks the whole Activity detail page. `events` resolves to `null` whenever `eventsLoadable` is not `hasData`, including `hasError`, and the page then renders `<ActivitySkeleton />` even though `detail` loaded successfully. The Context, Runner, and Network tabs and the header are collateral. Before this PR those surfaces rendered as soon as `zeroActivityDetail$` resolved. Please scope the events loading/error state to the Steps tab and keep the rest of the page rendering.

- `turbo/apps/api/src/signals/services/zero-run-detail.service.ts:38-50,208-209` — the PR summary has no `Fallbacks` section for the framework-resolution chain it re-introduces. Per `docs/fallback.md` §9 this is a P1 on its own, independent of whether the branch is justified. Please add the entry described in the Fallbacks section below.

### Fallbacks

- `zero-run-detail.service.ts:208` `zeroRunAgentEvents` / `extractFramework` — resolves the rendering framework as `runContextCliAgentType(runId) ?? content.agent.framework ?? firstAgent.framework ?? "claude-code"`, covering runs whose Axiom run-context row is absent and whose `agentComposeVersions.content` is null via the `leftJoin` or carries no framework. Surface: none — this is not a cross-version rollout branch; it is a presentational parser default restored verbatim from the pre-#26689 file (identical at `c7266ba~1:zero-run-detail.service.ts:41-51`). Removal condition: none evidenced in the diff or PR. Declared in PR summary: No — P1. Verdict: Justified under `docs/fallback.md` §6.4 as a presentational default with no identity meaning beyond event-group parsing, but it still needs the one-line summary entry required by §9.

- `activity-signals.ts:55-95` `fetchAllAgentEvents` — no fallback is present for the additive-route rollout window, and one is required. See the third P1 finding.

### Testing

- Coverage: Insufficient. The restore paths are well covered — `run-reads.bdd.test.ts` exercises owner scoping, the 401/404 auth matrix, two-page sequence pagination, the emitted APL shape, and `noCache`; `activity-paged-events.test.tsx` covers initial full-history loading, multi-page loading, server-cursor use, and one happy-path watermark stop; `activity-detail-retained-data.test.tsx` covers events in the download for both the populated and empty cases. Missing: (1) a terminal run whose events never reach `lastEventSequence`, which is the P0 non-terminating loop; (2) the additive-route `404` rollout case; (3) an events-request failure asserting that the header and the Context/Runner/Network tabs still render.
- Conventions: Compliant. Integration-level throughout, real DB on the API side, mocking only at external boundaries (Axiom, `mockApi`/MSW), no `vi.mock` on relative paths, no fake timers, no assertions on internal mock call counts as a substitute for behavior. Note that the 1s poll interval does not slow the suite because `setLoop` yields to a macrotask under `IN_VITEST` rather than sleeping.

seven332:
Addressed the requested changes on current head `b0529b7331`:

- moved event loading and polling from the DOM ref into the Activity page setup lifecycle, including a route-navigation regression test
- changed steady-state polling to request only after the contiguous visible sequence and merge new pages; an empty history now waits for sequence 0 without sending an invalid cursor
- bounded a terminal history with a permanent indexing gap after 30 unchanged polls, with a regression test that proves setup terminates
- decoupled Activity detail rendering from event loading, so the header and Context/Runner/Network tabs remain available when the event request fails; the failed Steps load also leaves its loading state while the error still propagates
- added the temporary additive-route `404` tolerance, documented its app/API rollout window, and created cleanup issue #27010
- removed the unused framework field from the paginated event response, eliminating repeated compose and run-context framework reads while retaining the Activity detail response as the framework source

Focused API and Activity tests pass (3 files, 37 tests). API, Platform, and API-contract linting pass. Full Knip and full Turbo type checking also pass locally.

PR #26804, opened 2026-08-13
feat(chat): read a steer burst as one group

tongx-blip:
## What

The steer acknowledgement did not read as part of the conversation it belongs to:

- it wore the same hairline rule the transcript uses to mark a boundary between two stretches of work, so the label read as the start of something else rather than the closing line of what the user just sent
- the copy button under each message sat close to the text above it and far from the text below, so a burst never looked like one block

## Change

- **the acknowledgement drops the rule** and keeps the serif-italic label
- **one even rhythm through a burst**: 6px from a message to the copy button under it, 6px from that button to whatever comes next — the following message, or the acknowledgement. Every message keeps its own copy button, in the place it has always been
- consecutive user messages close up to that rhythm; a model change between two of them, or a message that renders as its own card (workflow, goal), ends the stack

Three values move together and are named for it: `MESSAGE_STACK_GAP_CLASS` on the actions row, `MESSAGE_STACK_PULL_CLASS` on a stacked message and on the acknowledgement row.

## Preview

- before / after: https://steer-group.sites.vm0.io
- rendered from the app's own markup: https://steer-group.sites.vm0.io/real.html

## Checks

- `chat-run-actions` — 11 passed; `chat-lifecycle` — 30 passed
- type-check, eslint, format clean
- measured in Chromium on the app's own rendered markup: all four gaps in a burst are exactly 6px (message → button, button → message, message → button, button → acknowledgement), and the acknowledgement is flush with the right edge every message lines up on

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #27014, opened 2026-08-13
feat(chat): stack every message a user sent back to back

tongx-blip:
## What

Two changes to how a run of user messages reads in the transcript.

**1. The stack now starts at the first seam.** It was scoped to pairs of steers, so the message the run started from still sat a full gap above the first correction after it — the most visible seam of all, since a burst grows out of exactly that pair. Adjacency is now the whole condition, which also means the spacing no longer waits on `ChatRunContinuationPresentation`: a burst that carries no acknowledgement is still a burst. A run that ends re-forms its groups, so `ChatThreadEventGroups` tracks the last group that actually rendered something and the stack survives the run finishing instead of springing open.

An assistant reply between two messages still breaks the stack, as does a model change or a message that renders as its own card (workflow, goal).

**2. The acknowledgement retires with its run.** It answers "did my correction land in the work that is running" — true until that run ends, at which point the work itself is the answer. Keeping it turned a live status into a permanent transcript landmark that accumulated down the thread. `chatRunPresentationForGroups` now records the run each steer was picked up by, and drops those steers once that run has a terminal event.

## Preview

Rendered from the app's own markup: https://steer-group.sites.vm0.io/real-on.html (live run) and https://steer-group.sites.vm0.io/real-done.html (finished run).

## Checks

- `chat-run-actions` — 12 passed, including a new test that the acknowledgement is gone once the run completes
- `chat-lifecycle` + `chat-run-results` — 69 passed
- type-check and eslint clean on the changed files
- verified in Chromium against the app's own rendered markup, four transcripts: switch on, switch off, run finished, assistant reply in between. All consecutive user messages stack on the even rhythm (34px bubble to bubble = 4px + control + 4px); the acknowledgement shows only while its run is live; an assistant reply breaks the stack.


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: d97fad91352f5e24908b38ef356e7274cf04179d
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Two behaviour changes in the chat transcript, both presentation-level and confined to `turbo/apps/platform`.

1. **The stack starts at the first seam.** It required both neighbours to be steers, so the message a run started from sat a full gap above the first correction after it. Adjacency is now the whole condition, and `ChatThreadEventGroups` tracks the last group that actually rendered something so a finishing run — which re-forms its groups — does not spring the stack open.
2. **The acknowledgement retires with its run.** `chatRunPresentationForGroups` records the run each steer was picked up by and drops those steers once that run has a terminal event, so the line stops accumulating down the thread as a permanent landmark.

The stack is deliberately **not** gated by `ChatRunContinuationPresentation`, and the PR body says so. This is the author's stated product decision (the acknowledgement stays gated; only the spacing is universal), and it is the option the earlier review of #26804 named as the alternative to gating. Recording it here so the reversal is not silent.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic, including null/undefined paths | Pass | `stackedOnPrevious` keeps its guards (`modelChange === undefined`, `rendersUserBubble` on both sides) and now falls back to the parent's `stackFirstOnPrevious` only for a group's first message. `groupRendersContent` is what keeps a bookkeeping-only assistant group (terminal run event, usage) from breaking a visually adjacent stack, and `groupHasUserBubble` keeps a fold-only user group from being stacked against. |
| `REVIEW.md` | Correctness of the retirement pass | Pass | A run's terminal event arrives after its steers, so the reconciliation runs after the whole transcript is walked rather than inline; `runId ?? lastAssociatedRunId` covers both a reconciled steer (carries the run id) and a queued one (attaches to the run it interrupted). |
| `REVIEW.md` | Security (injection, secrets, auth) | N/A | No network call, auth boundary, or user-controlled HTML in the diff. |
| `REVIEW.md` | Type safety, no `any`, no over-broad types | Pass | New props and helpers are explicitly typed (`stackFirstOnPrevious?: boolean`, `CompletedWorkFold \| null`, `ChatEventGroup`); no `any`, no assertions, no `unknown` left in the new signatures. |
| `REVIEW.md` | Style and maintainability | Pass | The run bookkeeping was flattened to satisfy `max-depth` by hoisting the run id beside the steer check rather than suppressing the rule; helpers are single-purpose and named for the question they answer. |
| `REVIEW.md` | New user-facing behavior gated behind a `FeatureSwitchKey` | Pass (declared exception) | The acknowledgement and its retirement stay inside `ChatRunContinuationPresentation` (`runPresentation` is `null` when off, so `steerEventIds` is empty). The spacing is intentionally universal and is declared as such in the PR body, which is what `REVIEW.md` asks for when a change is deliberately unscoped. |
| `REVIEW.md` | Feature regression guard: existing UI changed without regression coverage | Pass (with note) | The switch-off transcript now stacks too — a deliberate change to an existing path. Its behavioural side is covered by the existing "keeps legacy actions and hides the acknowledgement when the feature switch is disabled" case; the spacing itself is not jsdom-observable and was verified in Chromium on the app's own markup, including the switch-off transcript. |
| `REVIEW.md` | DB / JSONB review gate | N/A | No schema, contract, or `jsonb(...).$type<...>()` change. |
| `docs/bad-smell.md` | Zero `any`, zero lint/type suppressions | Pass | No `any`; no `eslint-disable`, `@ts-ignore`, `@ts-expect-error`, or `prettier-ignore` — the `max-depth` failure was fixed by restructuring. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | One new static import (`isChatRunTerminalEventType`). |
| `docs/bad-smell.md` | No unnecessary try/catch, fail fast | Pass | No error handling added. |
| `docs/fallback.md` | Default is no fallback; every introduced fallback declared | Pass | No compatibility branch, `??`/`||` default on a contract-owned field, or version-skew path. `runId ?? lastAssociatedRunId` selects between two live sources of the same fact, not between an old and a new shape. No `Fallbacks` section required. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | The new test asserts a live condition (the line is gone once the run completes), not that deleted code stayed deleted. |
| `docs/effect.md` | React render stays pure | Pass | `previousVisibleGroup` is a local accumulator consumed within the same `groups.map` pass — no Store write, DOM mutation, timer, or subscription in render. |
| `docs/cache.md` | RC-2 no render-time signal allocation; RC-3 no unbounded caches; RC-5 no duplicate sources of truth | Pass | The two new collections are function-local to `chatRunPresentationForGroups` and die with the call; no module-level cache, no signal created in render. |
| `.claude/skills/ccstate/SKILL.md` | DOM callbacks use `detach(..., Reason.DomCallback)` with `pageSignal$` | N/A | No DOM handler added. |
| `docs/testing.md` | Integration tests only, mock at the boundary, real infrastructure | Pass | The new test enters through `detachedSetupPage`, drives the mocked lifecycle, and asserts on a page-visible testid; no internal `vi.mock`, no fake timers. |
| `docs/testing.md` | `feat:` commits carry integration coverage for the changed behavior | Pass | The retirement is new behaviour and has a dedicated test. The stacking change is spacing, which jsdom cannot observe; it was verified in Chromium across four transcripts. |
| `docs/testing/app-testing.md` | Page-level tests through `detachedSetupPage`, assert page-visible behavior | Pass | The new case sits with the other page-level chat-run tests and asserts through `screen`. |
| `docs/testing/anti-patterns.md` | AP-1 test behavior, not implementation details | Pass | The test asserts the acknowledgement's absence, not a class name or a call count. |

### Findings

No P0 or P1 findings.

### Testing

- Coverage: Adequate. `chat-run-actions.test.tsx` is now 12 tests, including the new "retires the acknowledgement once the run it belongs to finishes". The existing switch-off case still pins the behavioural half of the ungated spacing change. Chromium verification on the app's own rendered markup covered four transcripts — switch on, switch off, run finished, assistant reply in between — confirming every consecutive user message stacks on the even rhythm, the acknowledgement appears only while its run is live, and an assistant reply still breaks the stack.
- Conventions: Compliant. No internal `vi.mock`, no fake timers, no mock-call assertions, no negative test against removed code.

PR #26993, opened 2026-08-13
test(api): settle browser cleanup between thread deletes

e7h4n:
## Summary

- settle each chat-thread deletion's route-owned `waitUntil` work before deleting the next related browser fixture thread
- preserve the assertions that the earliest idle lease is reclaimed before a new provider session starts
- keep the existing provider-stop cleanup assertions intact

## Root cause evidence

- Trigger: Turbo run [31704531149](https://github.com/vm0-ai/vm0/actions/runs/31704531149), `test-api (4)` job [94461715102](https://github.com/vm0-ai/vm0/actions/runs/31704531149/job/94461715102), exact SHA `917fe5141289c0b844062952b4b36fcd505ecd41`.
- Recurrence: merge-group run [31703544445](https://github.com/vm0-ai/vm0/actions/runs/31703544445), job [94458400085](https://github.com/vm0-ai/vm0/actions/runs/31703544445/job/94458400085), failed the same test with the same PostgreSQL `40P01` cycle.
- A same-SHA comparison run [31703878156](https://github.com/vm0-ai/vm0/actions/runs/31703878156) passed the test, but its logs contain the same deadlock with the opposite victim: the event-sequence UPDATE was aborted and logged by `ChatThreadDelete`, so a green rerun did not mean the race was gone.
- That passing-run stack identifies the late UPDATE owner as `dispatchCancelSideEffects -> drainOrgQueue$ -> promoteQueuedCandidate -> revokeQueuedRunAssistantMarkers -> revokeChatEvent -> reserveChatEventSeqIds`, not the earlier Browser Use stop or screenshot tasks. Those browser tasks are settled by the existing pre-assertion flush.
- The teardown deleted three related fixture threads back-to-back. A previous DELETE could therefore retain an organization-queue lock on `agent_runs` while revoking a marker on a later fixture thread. The next DELETE held that chat thread and reached `agent_runs` through the temporary `zero_runs -> agent_runs` compatibility trigger, forming the observed reverse lock order.

## Fix

`flushWaitUntilForTest()` is the route's deterministic ownership barrier for post-delete cancellation and queue work. Move that barrier inside the fixture deletion loop so each DELETE fully settles before cleanup begins deleting another thread that its side effects may update.

This does not add retries, sleeps, timeouts, skipped assertions, or swallowed failures.

## Validation

- `pnpm format`
- `pnpm turbo run lint --concurrency=1 --log-order grouped`
- `pnpm knip`
[...]

PR #27017, opened 2026-08-13
test(api): isolate claim-setup heartbeats from mocked clocks

e7h4n:
## Problem

`test-api (7)` failed in the merge group with a dropped runner reuse preference:

```
AssertionError: expected { kind: 'noPreference', … } to match object { kind: 'preference', … }
 ❯ expectReusePreference run-lifecycle.bdd.test.ts:4887
```

Failing run: https://github.com/vm0-ai/vm0/actions/runs/31708700324 (job [`test-api (7)`](https://github.com/vm0-ai/vm0/actions/runs/31708700324/job/94475866423), SHA `855846baaf436d92988d843cb39abe74a55750c9`).

The represented PR (#26993) only touched `zero-browser.test.ts`; the exact PR head and the exact main parent each passed this test. The assertion immediately before the failure (`poll.body.job?.runId === followUp.runId`) passed, so the correct follow-up job was selected and only its **preference projection** diverged.

## Root cause

`turbo/apps/api/src/signals/routes/runners.ts:522-529` prunes stale runners with a single **unscoped** delete, evaluated against the API process clock:

```ts
await db.delete(runnerState).where(
  lt(runnerState.lastSeenAt, new Date(currentDate.getTime() - STALE_RUNNER_THRESHOLD_MS)),
);
```

`currentDate` comes from `nowDate()`, which honours `mockNow()`. This is the only `runnerState` GC in the codebase, and parallel API test files share one database per shard. So any file that heartbeats under a **future** mocked clock deletes every runner row in the shard.

`cron-execute-morning-briefs.test.ts:92-96` anchors on the next 07:00 Asia/Shanghai *strictly after the real clock*, then calls the shared `heartbeatRunner()` helper while that clock is active:

```ts
const SEVEN_LOCAL = new Cron("0 7 * * *", { timezone: TIMEZONE })
  .nextRun(new Date(now() + 60 * 1000))!.getTime();
const BEFORE_SEVEN_LOCAL = SEVEN_LOCAL - 10 * 60 * 1000;
```

At the failing run (2026-08-13T14:09:53Z) that clock sat **8.67 hours ahead** of real time, so its prune deleted every row with `last_seen_at < 2026-08-13T22:45:00Z` — including the ordered snapshot holder that `run-lifecycle.bdd.test.ts` had just written at real time.

With the holder gone, the following out-of-order heartbeat no longer hit a conflict, so `onConflictDoUpdate`'s `setWhere` ordering guard was bypassed and the row was re-inserted with empty `heldSandboxStates` → `noPreference`. `runner_job_queue` is untouched by the delete, which is exactly why job selection still succeeded.

It is intermittent because it needs the morning-briefs heartbeat to land inside the narrow window between the ordered heartbeat and its poll — pure parallel-fork scheduling. #26993's edit to `zero-browser.test.ts` perturbed shard 7's file interleaving; it did not introduce the defect.

## Fix

Run the shared claim-setup heartbeat under `withNowScopeForTest()` so it uses the real clock and can only prune genuinely stale rows. This helper is incidental setup — it exists so a runner exists to claim a job — and no test asserts on its timestamp. Feature-specific heartbeats that deliberately assert ordering, expiry, or protection-window semantics keep their mocked clocks.

Scoping the product delete by runner group was rejected: it is the only `runner_state` GC, so group-scoping would leak rows for groups that stop heartbeating.

A guard in the ordered-heartbeat test pins the invariant that an unrelated runner's heartbeat must not drop this runner's retained snapshot.

## Verification

- **A/B reproduction** against a local UTC Postgres: with a simulated parallel future-clock heartbeat and the fix reverted, the test fails with the *exact* CI assertion at the *exact* line (`run-lifecycle.bdd.test.ts:4887`). With the fix applied, it passes.
- Target test repeated **5×** — all pass.
- All **38 test files** that use the changed shared helper run together against one shared database: **981 passed**.
- `prettier --check` clean; `oxlint` warning count unchanged (7 before, 7 after — all pre-existing).
- `tsc --noEmit` for `apps/api` did not complete locally (stalled past 29 min; this project is known-heavy). Left to the PR pipeline. The change adds no new type surface.

## Follow-up worth tracking separately

The same unscoped delete is a latent production risk: an API instance whose clock runs >5 minutes fast would evict **all** healthy runner rows globally. Out of scope for this flaky fix.

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/helpers/api-bdd-runs.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/helpers/api-bdd-runs.ts
@@ -51,7 +51,7 @@ import { createAppWithRoutes } from "../../../../app-factory-core";
 import { setupAppWithRoutes } from "../../../../__tests__/test-app";
 import { accept, type TestContext } from "../../../../__tests__/test-context";
 import { mockEnv, mockOptionalEnv } from "../../../../lib/env";
-import { now } from "../../../../lib/time";
+import { now, withNowScopeForTest } from "../../../../lib/time";
 import { createAgentComposeFixture } from "../../../../test-fixtures/agent-composes";
 import {
   createDirectRunFixture,
@@ -1107,13 +1107,17 @@ export function createRunsApi(context: TestContext) {
     },
 
     async heartbeatRunner(group?: string) {
-      return await accept(
-        runApp(context)(runnersHeartbeatContract).heartbeat({
-          headers: runnerHeaders(true),
-          body: runnerHeartbeatBody({ group }),
-        }),
-        [200],
-      );
+      // Claim setup must not let a feature-specific mocked clock prune runner
+      // rows owned by parallel files in the shared test database.
+      return await withNowScopeForTest(async () => {
+        return await accept(
+          runApp(context)(runnersHeartbeatContract).heartbeat({
+            headers: runnerHeaders(true),
+            body: runnerHeartbeatBody({ group }),
+          }),
+          [200],
+        );
+      });
     },
 
     async requestHeartbeatRunner(
--- a/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
@@ -4916,6 +4916,13 @@ describe("CHAIN-RUN: entitled run lifecycle through runner and sandbox webhooks"
       sequence: 1,
       resource: undefined,
     });
+    // An unrelated runner's heartbeat must not drop this runner's retained
+    // snapshot. Parallel test files share one database and perform incidental
+    // claim setup under their own mocked clocks, so this heartbeat is issued
+    // far ahead of `baseTime` to pin the stale-runner pruning boundary.
+    mockNow(baseTime + 9 * 60 * 60 * 1000);
+    await api.heartbeatRunner(runnerGroup);
+    mockNow(baseTime + 5000);
     await expectReusePreference("reusableSandbox");
 
     mockNow(baseTime + 20_000);


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: bc666b115e6ffc496ecd095ecf0c2043084635d2
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Wraps the shared BDD claim-setup heartbeat helper (`createRunsApi().heartbeatRunner`) in `withNowScopeForTest()` so it issues on the real clock, and adds a guard in `run-lifecycle.bdd.test.ts` pinning that an unrelated runner's heartbeat cannot drop an established reuse-preference snapshot.

The defect is well-identified: `runners.ts` prunes stale runners with a single unscoped `delete(runnerState).where(lt(lastSeenAt, currentDate - STALE_RUNNER_THRESHOLD_MS))` evaluated against `nowDate()`, which honours `mockNow()`. Since parallel API test files share one database per shard and this is the only `runner_state` GC in the repo, any file heartbeating under a future mocked clock wipes every runner row in the shard. `cron-execute-morning-briefs.test.ts` anchors `SEVEN_LOCAL` on the next 07:00 Asia/Shanghai *strictly after the real clock* and calls this helper under it — ~8.7h ahead at the failing run, deleting the ordered snapshot holder mid-test. `runner_job_queue` is untouched by that delete, which is exactly why job selection still succeeded and only the preference projection diverged.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic; no races or wrong assumptions | Pass | `withNowScopeForTest` opens a fresh `AsyncLocalStorage` store with `value: undefined`, so `now()` falls through to `Date.now()` inside the request. Composes correctly both where the caller uses the process-global override (`run-lifecycle.bdd.test.ts`) and where the caller is already inside a scope (`zero-workflow-queue.test.ts` wraps every test), since the nested store is a new object and the outer value is untouched. |
| `REVIEW.md` | Type safety; no `any`, no broad types | Pass | No new type surface. `withNowScopeForTest<T>(work: () => Promise<T>): Promise<T>` preserves the `accept(...)` return type through inference. |
| `REVIEW.md` | Security | N/A | Test-only diff; no auth, input handling, or credential surface changed. |
| `REVIEW.md` | Testing coverage for the commit type | Pass | `test:` commit. It still ships a regression guard that fails without the fix, rather than only changing the helper. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No `packages/db/src/schema/**`, contract, or `jsonb().$type<>()` change. |
| `REVIEW.md` | Feature regression guard | N/A | Not a feature PR; no existing UI, API, or user flow changed or rerouted. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions, no dynamic imports, no hardcoded config | Pass | Diff adds a static named import and one wrapper call; no suppression comments, no `import()`, no URLs or env literals. |
| `docs/bad-smell.md` | Comments explain WHY, not WHAT | Pass | Both comments state the invariant and the reason (shared DB, parallel files, pruning boundary) rather than restating the code. |
| `docs/bad-smell.md` | No unnecessary try/catch; fail fast | Pass | No error handling added; `accept(..., [200])` still fails loudly on any non-200. |
| `docs/fallback.md` | Every fallback introduced by the diff is declared, justified, and time-boxed | N/A | The diff introduces no fallback. `withNowScopeForTest` is a test-scoped clock boundary, not a compatibility branch, `??`/`||` default, or tolerance path. No `Fallbacks` section required per §9. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | The added guard asserts a live positive outcome (`reusableSandbox` preference is retained), not the absence of deleted code. |
| `docs/testing.md` | Integration tests only; mock at the system boundary | Pass | Change goes through the real Hono route via `runApp(context)(runnersHeartbeatContract).heartbeat(...)`. No new `vi.mock()`, and none with a relative path. |
| `docs/testing.md` | Real database, no DB mocking | Pass | No mocking of `globalThis.services.db`; the fix exists precisely to keep the real shared database correct under concurrency. |
| `docs/testing/api-testing.md` | Process-wide overrides need an explicit owner and scoped reset semantics | Pass | This is the doc's sanctioned remedy verbatim — "such as an `AbortSignal` or async-local boundary". `withNowScopeForTest` is the async-local boundary, scoping the clock override to the one request instead of the worker. |
| `docs/testing/api-testing.md` | A test must be correct while other API tests run concurrently; no isolation via broad clock partitions, global locks, ordering, or worker serialization | Pass | The fix removes an unintended cross-file effect rather than partitioning the clock. No lock, ordering constraint, serialization, or residue-tolerant assertion is introduced; the shard still runs fully parallel. |
| `docs/testing/api-testing.md` | Cleanup must not delete or restore pre-existing shared state to make an assertion pass | Pass | The diff stops an incidental setup call from deleting other files' rows. It adds no teardown that mutates shared state. |
| `docs/testing/testing-external-behavior.md` | Construct and verify through the API endpoint; no DB schema imports, row writes/reads, or service calls | Pass | Setup and assertion both go through endpoints (`heartbeatRunner`, `requestPollRunner`); assertion reads `poll.body.job`. No DB schema import or direct row access added. |
| `docs/testing/anti-patterns.md` | AP-5: no fake timers | Pass | No `vi.useFakeTimers()` / `advanceTimersByTime()`. Uses the repository clock abstraction `mockNow` from `lib/time.ts`, as the doc prescribes. |
| `docs/testing/anti-patterns.md` | AP-1/AP-7: assert behavior and outcomes, not mock calls or implementation details | Pass | The guard asserts the API-visible reuse preference via `runnerPreference(poll.body.job)`; no `toHaveBeenCalled()` and no internal state assertion. |
| `docs/testing/anti-patterns.md` | AP-4/AP-6: never mock internal code, no partial internal mocks | Pass | No `vi.mock()` added at all. |
| `docs/testing/patterns.md` | Pattern 1: route integration tests through `setupApp()`/ts-rest contract | Pass | Unchanged; the helper still calls the route through the contract and narrows with `accept()`. |
| `.claude/skills/ccstate/SKILL.md` | ccstate execution/lifecycle rules for `apps/api`; no floating promises, no `void`, correct signal ownership | Pass | No signal, `computed`, or command added. Both new call paths are fully `await`ed (`return await withNowScopeForTest(async () => { return await accept(...) })`), so no floating promise and no `void` suppression. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Nits (non-blocking)

- `run-lifecycle.bdd.test.ts` — the guard's comment says "this heartbeat **is issued** far ahead of `baseTime`". With this fix in place the helper deliberately ignores the mocked clock, so the request is *not* issued at `baseTime + 9h`; the mock is inert by design, and that inertness is precisely what the guard detects if the wrapper is ever removed. Wording like "so the caller's clock is set far ahead of `baseTime`, which the helper must ignore" would read more accurately to someone changing this later. Not worth a respin on its own.

### Verification reviewed

- Scope boundary checked: the fix wraps only `heartbeatRunner` (incidental claim setup, ~15 caller files, no test asserts its timestamp). The raw `requestHeartbeatRunner` variant is intentionally left on the mocked clock for the three files that assert ordering/expiry semantics (`run-lifecycle.bdd`, `run-cron.bdd`, `zero-integrations-telegram-post`). I confirmed every unwrapped call site runs under a `now()`-derived clock (`baseTime`, `priorityBase`, `sourceCompletedAt`, all `= now()`), so none of them can evict a parallel file's fresh row. The boundary is correct and complete for this defect.
- Rejecting a product-side fix is the right call and is justified in the PR body: `runners.ts:523` is the only `runner_state` GC, so scoping the delete by runner group would leak rows forever for groups that stop heartbeating.
- The latent production risk the author flagged (an API instance with >5min clock skew evicting all healthy runner rows globally) is correctly scoped out of a flaky-test fix and worth a separate issue.

### Testing

- Coverage: Adequate. The added guard is a true regression test — reverting the helper change reproduces the original CI failure with the identical assertion at the identical location (`expectReusePreference`, `run-lifecycle.bdd.test.ts:4887`), and re-applying it passes.
- Conventions: Compliant. No fake timers, no internal mocks, no direct fetch stubbing, no DB access from the test, assertions on API responses only.
- Blast radius: all 38 test files that consume the changed shared helper were run together against one shared database (981 tests passing), which is the configuration that actually exercises the cross-file interaction being fixed.
- Not verified locally: `pnpm -F api check-types` did not complete (stalled well past 29 minutes on this known-heavy project) and is left to the PR pipeline. Given the change adds no new type surface, this is low risk.

Note: posted as a self-review by the PR author under the pr-auto flow.

PR #27072, opened 2026-08-14
fix(platform): align avatar template toolbar to the default control scale

tongx-blip:
## What

The avatar template picker toolbar carried two different control heights, and the filters trigger used a pill shape that the button component does not offer.

- `SegmentControl` (aspect ratio) ran at `size="sm"` → **h-8 (32px)**
- The filters `Button` ran at `size="sm"` with a `className="h-9 … rounded-full …"` override → **h-9 (36px)**, pill radius

`segment-control.tsx` documents that its sizes mirror `buttonVariants` one for one, with `default` = h-9 matching `Input` and `SelectTrigger`. The filter `Select`s inside this very popover are h-9, and the equivalent filter trigger on the connectors page is h-9 `rounded-lg`. The segment control was the only element off that step.

`buttonVariants` has no shape axis — its base radius is `rounded-lg`, and the only other `rounded-full` buttons in the platform are circular icon-only controls. The pill here was a one-off.

## Change

Both controls now take the component defaults: h-9, `rounded-lg`. The redundant `gap-2`, `bg-background`, and `shadow-none` classes are dropped (all already supplied by the base and the `outline` variant), as is the dead `size={15}` on the icon (the base `[&_svg]:size-4` wins over the attribute).

## Verification

Rendered both states against the real `globals.css` tokens in dark mode and measured:

| | segment | filters |
|---|---|---|
| before | 32px | 36px, radius 3.4e7px |
| after | **36px** | **36px, radius 8px** |

No test asserts on these classes.

### DIFF
--- a/turbo/apps/platform/src/views/zero-page/avatar-template-picker.tsx
+++ b/turbo/apps/platform/src/views/zero-page/avatar-template-picker.tsx
@@ -193,13 +193,8 @@ function CatalogFiltersPopover({
   return (
     <Popover>
       <PopoverTrigger asChild>
-        <Button
-          type="button"
-          variant="outline"
-          size="sm"
-          className="h-9 gap-2 rounded-full bg-background px-3 shadow-none"
-        >
-          <SlidersHorizontal size={15} />
+        <Button type="button" variant="outline">
+          <SlidersHorizontal />
           {t(($) => {
             return $.artifacts.templates.filters.title;
           })}
@@ -245,12 +240,7 @@ function AvatarAspectRatioPicker({
     return $.artifacts.templates.filters.aspectRatio;
   });
   return (
-    <SegmentControl
-      aria-label={label}
-      size="sm"
-      value={value}
-      onValueChange={onChange}
-    >
+    <SegmentControl aria-label={label} value={value} onValueChange={onChange}>
       <SegmentControlItem value="portrait" aria-label={`${label}: 9:16`}>
         <span className="h-4 w-2.5 rounded-[2px] border-2 border-current" />
         9:16


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: 216a2f173ccd2277e51c41eebbc3decf46537cba
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Removes two one-off style overrides in `avatar-template-picker.tsx` so the catalog toolbar sits on a single control scale: the aspect-ratio `SegmentControl` drops `size="sm"` (h-8) for the default h-9 step, and the filters `Button` drops `size="sm"` plus its `h-9 gap-2 rounded-full bg-background px-3 shadow-none` className for the plain `outline` default (h-9, `rounded-lg`). No logic, contract, or state changes.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic — no race conditions, wrong conditionals, or API misuse | Pass | Diff only deletes presentational props/classes in `avatar-template-picker.tsx`; control flow, `onValueChange`, `activeCount`, and filter state are untouched. |
| `REVIEW.md` | Security — no injection, hardcoded secrets, or missing auth on new endpoints | N/A | No network, auth, storage, or user-input handling in the changed lines. |
| `REVIEW.md` | Type safety — no `any`, no overly broad types | Pass | No type annotations added or removed; `SegmentControl`/`Button` prop types still satisfied after dropping the optional `size` prop. |
| `REVIEW.md` | Style & maintainability — no oversized functions or duplicated logic | Pass | Net −10 lines; both call sites now inherit `buttonVariants` / `segmentControlVariants` defaults instead of restating them. |
| `REVIEW.md` | Testing coverage — `fix:` commits need a regression test | Pass (with rationale) | Presentation-only change with no reachable behavior; see Testing. |
| `REVIEW.md` | Feature regression guard — existing flows changed without coverage | Pass | `CatalogFiltersPopover` is shared by the avatar toolbar (`:485`) and the voice toolbar (`:588`); both change height/radius only. Existing assertions in `chat-composer-template-picker.test.tsx` (`getByText("Filters", { selector: "button" })` at `:988`, `not.toHaveClass("border-primary/40", "text-primary")` at `:991`) target the DOM role and active-state classes this diff does not touch. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No `packages/db/src/schema/**`, contract, Zod, or `jsonb().$type<>()` change. |
| `REVIEW.md` | Fallback discipline — declare and justify introduced fallbacks | N/A | The diff removes className overrides, not compatibility branches; no `??`/`||` default, version branch, or feature-switch gate is added, kept, or removed. |
| `docs/bad-smell.md` | Zero tolerance for `any` | Pass | No `any` in the diff. |
| `docs/bad-smell.md` | Zero tolerance for lint/type suppressions | Pass | No `eslint-disable`, `@ts-ignore`, `@ts-expect-error`, or `prettier-ignore` added. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | Imports unchanged; `SlidersHorizontal` and `SegmentControl` remain static top-level imports. |
| `docs/bad-smell.md` | No hardcoded URLs or environment-specific config | N/A | No configuration or URL in the changed lines. |
| `docs/bad-smell.md` | Fail fast — no fallback/recovery slop | Pass | The diff deletes redundant style declarations; it adds no defaulting or recovery path. |
| `docs/testing.md` | Integration tests only, mocked at the system boundary | N/A | No test file changed; no new `vi.mock()`, MSW handler, or DB access introduced. |
| `docs/testing.md` | Do not over-test presentation ("UI text content … isn't logic") | Pass | Asserting `h-9` / `rounded-lg` class strings would pin design-system tokens rather than behavior; the size contract belongs to `packages/ui`'s own `segment-control.test.tsx` / `button.test.tsx`. |
| `docs/testing.md` | Deployment-compatibility coverage for deployable boundaries | N/A | Frontend-only styling; no API contract, runner protocol, queue payload, or persisted shape crosses a version boundary. |
| `docs/effect.md` | Keep React render pure — no Store writes, I/O, or resource creation in render | Pass | `AvatarAspectRatioPicker` and `CatalogFiltersPopover` still only read `useTranslation()` and render elements; the diff removes JSX attributes only. |
| `docs/effect.md` | Choose the execution boundary — `computed` / semantic `command` / setup / `onRef` | Pass | No boundary changed: `onChange` still forwards to `setFilters(...)` from `AvatarCatalogFilters` (`:479`), and no effect, ref, or timer is added. |
| `docs/effect.md` | Own signals outside render; pass the signal interface into React | Pass | `ComposerSignals` is still received as a prop; no signal factory is invoked from render. |
| `docs/cache.md` | RC-1 no render-phase side effects | Pass | Both changed components remain pure functions of props plus `useGet`/`useSet` reads. |
| `docs/cache.md` | RC-2 no render-time signal allocation | Pass | No `state`/`computed`/`command` created in these components before or after the diff. |
| `docs/cache.md` | RC-3 no unbounded lifetime caches | Pass | No module-level `Map`/`Set`/`WeakMap` or registry added. |
| `docs/cache.md` | RC-7 stable callback refs, cleanup preserved | Pass | No `ref` on the changed elements; `PopoverTrigger asChild` still forwards to a single `Button`. |
| `docs/cache.md` | RC-8 symmetric resource lifecycles | N/A | The diff acquires no listener, timer, observer, object URL, or subscription. |
| `.claude/skills/ccstate/SKILL.md` | DOM callbacks use `detach(..., Reason.DomCallback)`, never bare floating promises | N/A | Both handlers (`onValueChange`, `onClear`) are synchronous `setFilters` calls returning no promise. |
| `.claude/skills/ccstate/SKILL.md` | All `zeroClient$` calls wrapped in `accept` with an explicit status list | N/A | No HTTP call in the diff. |
| `.claude/skills/ccstate/SKILL.md` | DOM element refs go through `onRef`, passed as the stable `useSet` result | N/A | No DOM ref added or modified. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

- `Button` `size="default"` is `h-9 px-4` versus the previous `px-3`, and the segment items move `px-2.5` → `px-3`. The toolbar row grows roughly 16px wider in total. The container is `flex w-full flex-wrap items-center justify-between gap-3` (`:465-468`), so a narrow viewport degrades by wrapping rather than overflowing. No change requested.
- Dropping `size={15}` from `<SlidersHorizontal>` is a no-op at runtime: `buttonVariants`' base `[&_svg]:size-4` sets the CSS `width`/`height`, which overrides lucide's SVG presentation attributes. The icon rendered at 16px before and after; the deletion removes a misleading value rather than changing pixels.
- `AvatarVoiceToolbar` (`:665-673`) still pairs an `h-8 w-8` back button with this now-`h-9` filters trigger. That mismatch predates this PR — the trigger was already forced to `h-9` — so it is out of scope here, but it is the next candidate if the toolbar scale is being unified.

### Testing

- **Coverage: Adequate.** No regression test is added, and none is warranted. The diff changes only Tailwind size and radius tokens; there is no branch, state transition, or user-visible behavior to pin. A test would have to assert `h-9` / `rounded-lg` class strings on a DOM node, which `docs/testing.md` explicitly steers away from ("Focus tests on business logic and integration points"; UI presentation is listed under what we do not test), and `docs/testing/app-testing.md`'s `detachedSetupPage()` external-behavior interface cannot observe a control's height. The size scale itself is contract-tested where it is defined, in `turbo/packages/ui/src/components/ui/__tests__/segment-control.test.tsx` and `button.test.tsx`. Flagging this row explicitly rather than silently, since `REVIEW.md` sets missing regression tests on `fix:` commits at P0 — the exemption claimed here is that the fix has no behavioral surface, not that the rule was overlooked.
- **Conventions: Compliant.** No test files changed, so no `vi.mock()` boundary, MSW, fake-timer, real-database, or mock-cleanup rule is engaged.
- **Existing tests unaffected.** `chat-composer-template-picker.test.tsx` resolves the trigger by text plus `selector: "button"` and asserts only active-state classes; there are no snapshot files under `turbo/apps/platform`, so no golden output pins the removed classes.

PR #27073, opened 2026-08-14
perf(runner): skip idempotent connector registry writes

### LINKED ISSUE #27060: perf(rust): skip idempotent connector runtime registry writes
## Problem and evidence

`ProxyRegistryManager::apply_connector_runtime_updates_if_run_matches` persists the complete proxy registry whenever any connector runtime update is accepted, even when the resulting VM state is identical to the state already stored.

The accepted and changed states are currently conflated:

- `apply_connector_runtime_update` returns `true` for a registered Builtin target after an unconditional `HashMap::insert`, without checking the previous `NetworkPolicy`.
- Custom Available and Absent updates likewise return `true` whenever the target is registered, even if applying the state leaves the VM entry unchanged.
- The batch treats any `true` outcome as a mutation, advances `updated_at`, recomputes fail-closed reserve by serializing every stored policy, serializes the complete registry, and publishes it through a same-directory atomic rename.

Idempotent publications are a normal input. Connector runtime sync responses contain full resolved state rather than a delta. API wakeups can include targets whose resolved state did not change: for example, `updateUserCustomConnectors` publishes the union of previous and current custom connector IDs after a replacement and fans those targets to running runs in scope.

The redundant atomic replacement also invalidates the Python proxy's compiled registry cache. `load_registry_state` keys snapshots by absolute path, device, inode, mtime, and size, so an identical logical registry receives a new identity and must be read, classified, and compiled again.

This behavior was verified on `main` at `9e713bb9e1c18fc53a351a29b4397bb2e6e25bb4`.

## Decision

Detect logical mutation at the runner's connector runtime batch boundary while preserving the current accepted/not-accepted result returned to the sync scheduler.

Capture one batch-scoped snapshot of only the matched VM fields this path can mutate: firewalls, network policies, omitted Custom connector IDs, and connector routing variables. Apply the complete batch through the existing validation and mutation logic. Return accepted outcomes unchanged, but update `updated_at` and write the registry only if the resulting state differs from the snapshot.

This keeps change detection central and avoids duplicating the coupled Custom mutation rules. The snapshot is limited to one VM and occurs once per batch; it avoids shared-registry reserve computation, serialization, atomic replacement, and downstream proxy recompilation for unchanged batches.

## Alternatives considered

- Track rejected, accepted unchanged, and accepted changed in every mutation branch. This avoids the snapshot but duplicates change accounting across firewall replacement, obsolete policy removal, policy insertion, routing variables, and omission markers. A future missed flag could suppress a required enforcement write.
- Compare serialized registry bytes in the generic writer. This retains the full-registry work that causes the performance problem, is sensitive to unordered map serialization, and changes unrelated registry callers without evidence.
- Keep the current behavior. This retains lock-held work proportional to shared registry size and invalidates the proxy cache for normal idempotent refreshes.

## Risks and mitigations

- A newly mutated connector runtime field could be omitted from the snapshot. Keep the explicit field list next to the batch boundary and cover Builtin, Custom Available, and Custom Absent states through entry-point tests.
- Changed batches incur a one-VM snapshot clone. The snapshot is once per batch and narrower than the avoided shared-registry serialization and proxy compilation.
- Scheduler semantics could regress if unchanged is confused with rejected. Preserve the existing booleans and assert unchanged accepted updates still return `true`.
- Capacity and ownership checks must not move or weaken. Retain the existing validation and write path for every genuinely changed batch.

## Acceptance criteria

- Connector runtime publication distinguishes scheduler acceptance from a material change to persisted VM state.
- An accepted but unchanged Builtin, Custom Available, or Custom Absent update is still reported as successful and does not enter the retry path.
- `updated_at` and the registry file change only when at least one accepted update materially changes persisted state.
- A mixed batch containing unchanged and changed updates applies all valid changes and persists at most once.
- Source-IP/run ownership checks, missing-target behavior, Builtin firewall validation, Custom resource ownership validation, and fail-closed capacity guarantees remain unchanged.
- Focused tests publish identical Builtin and Custom states twice and verify the second publication succeeds without replacing the registry file; mixed-batch and genuine-change cases remain covered.

## Delivery notes

This is one independently reviewable runner-side PR. It requires no API contract, persisted JSON schema, database, migration, Python, configuration, or rollout change. Rollback only restores redundant writes and requires no data repair.

seven332:
## Summary

- compare the connector runtime fields for one matched VM before and after each accepted batch
- preserve successful sync outcomes while skipping `updated_at`, reserve calculation, serialization, and atomic replacement for logically unchanged state
- cover idempotent Builtin, Custom Available, Custom Absent, and mixed changed/unchanged publications through the real registry file boundary

## Scope

- no API or runner protocol changes
- no proxy registry JSON schema changes
- no database, migration, Python, configuration, or rollout changes

## Validation

- `cargo fmt --all --check`
- `cargo test -p runner --all-targets --all-features` (3122 passed, 20 ignored; guest inventory 1 passed)

Closes #27060

### DIFF
--- a/crates/runner/src/provider/connector_runtime_sync.rs
+++ b/crates/runner/src/provider/connector_runtime_sync.rs
@@ -5115,8 +5115,8 @@ mod tests {
                         "target": { "kind": "builtin", "connectorSlug": "slack" },
                         "state": "available",
                         "networkPolicy": {
-                            "allow": ["chat:write"],
-                            "deny": ["files:write"],
+                            "allow": ["chat:write", "files:write"],
+                            "deny": [],
                             "ask": ["channels:read"],
                             "unknownPolicy": "allow",
                         },
@@ -5141,6 +5141,15 @@ mod tests {
             captured_event(&events, "synced connector runtime target"),
             "target",
             "builtin:slack",
+        );
+        assert_eq!(
+            harness.slack_policy().await,
+            json!({
+                "allow": ["chat:write", "files:write"],
+                "deny": [],
+                "ask": ["channels:read"],
+                "unknownPolicy": "allow",
+            })
         );
         harness.shutdown().await;
     }
--- a/crates/runner/src/proxy/registry.rs
+++ b/crates/runner/src/proxy/registry.rs
@@ -623,7 +623,9 @@ impl ProxyRegistryHandle {
 
     /// Publish validated Builtin and Custom candidate updates in one registry
     /// transaction. `None` means the VM disappeared or now belongs to another
-    /// run; otherwise each boolean corresponds to the same-index update.
+    /// run; otherwise each boolean reports whether the same-index update was
+    /// accepted. Accepted batches persist only when the resulting VM state
+    /// changes.
     pub(crate) async fn apply_connector_runtime_updates_if_run_matches(
         &self,
         source_ip: &str,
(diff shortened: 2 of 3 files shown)


seven332:
## Code Review: PR #27073

**Reviewed head:** `0d52af5fb79986b15353bc92af13d03cb7992497`

### Summary

The runner now preserves accepted connector-sync semantics while avoiding registry persistence when the resulting VM firewall, policy, omission, and routing-variable state is unchanged. Genuine and mixed changes still use the existing validated atomic write path.

### Findings

No findings.

### Testing

Focused registry tests, runner Clippy and rustdoc, and the full runner test suite passed locally. Regression coverage uses a real temporary registry file and checks unchanged inode/`updated_at` plus genuine mixed-batch persistence. Runner-image and external-service E2E jobs remain CI-owned checks.

### Verdict

LGTM

PR #27128, opened 2026-08-14
feat: unify saved billing purchase routing

hulh122:
## Summary

- unify saved-payment purchase routing for Plan, Usage Pack, Credits, Concurrency, and paid member invitations
- align the Pro-to-Team purchase UI with the Credits confirmation flow and keep confirmation dialogs mounted only while open, preventing first-open and repeated-open flicker
- enable the saved billing purchase flow for every organization
- make Plan and Usage Pack confirmations concurrency-safe and reuse matching pending Usage Pack purchases and allocation snapshots

## Confirmation and payment behavior

- select saved payment methods consistently: subscription default, customer default, attached card, then legacy source
- sign previews and revalidate the price, customer, subscription state, and exact payment method before charging
- submit confirmations through rollout-compatible create routes while preserving support for older clients and APIs
- when a Plan preview becomes stale, the current API returns a refreshed preview with HTTP 200; the dialog updates the quote and requires a second confirmation instead of surfacing a 409 or silently charging a changed amount
- retain client compatibility with older APIs that return 409 by refreshing the preview in place
- explicitly pay open Plan invoices with the saved payment method; redirect to the hosted invoice only when payment remains incomplete or requires customer action

## Billing state fixes

- serialize initial Plan and Usage Pack confirmation per organization to prevent duplicate subscriptions
- make repeated confirmation idempotent and distinguish the active purchase from unrelated pending subscriptions
- create and reuse Usage Pack allocations correctly for granted or pending purchases without breaking consumption paths
- enforce paid invitation previews when the organization needs a Usage Pack
- keep restore visibility based on the actual Stripe schedule instead of stale pending tier metadata

## Compatibility and risk controls

- older clients that omit `supportsInAppPreview` continue to use hosted Checkout
- already issued previews remain confirmable across rollout changes
- current frontend code handles both refreshed 200 responses and legacy 409 responses during independent deployments
- no database migration or new production configuration is required

## Verification

- `pnpm format`
- `pnpm turbo run lint` — 13 tasks passed
- `pnpm check-types` — 14 tasks passed
- `pnpm knip`
- API billing checkout integration tests — 148 passed
- Billing UI integration tests — 47 passed
- Organization member integration tests — 13 passed
- Feature switch integration tests — 24 passed
- browser smoke test confirmed identical first-open and repeated-open Plan dialog rendering

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -35,3 +35,21 @@ self-harm. It excludes non-graphic news, medical, educational, historical,
 safety, moderation, and ordinary fictional discussion.
 
 _Avoid_: 18+ content, adult content
+
+# Billing Context
+
+The billing context decides whether a purchase can be reviewed and confirmed inside vm0 or must continue on a Stripe-hosted page.
+
+## Language
+
+**Saved payment method**:
+A Stripe payment method available to an organization through, in priority order, the subscription default, customer invoice default, an attached card, or a legacy default source.
+_Avoid_: Bound card, default card
+
+**Operation invoice**:
+The invoice produced by confirming the current purchase or subscription change. Unpaid invoices from earlier operations are not part of this decision.
+_Avoid_: Customer balance, historical invoice
+
+**Hosted invoice payment**:
+The Stripe-hosted invoice page used when an operation invoice remains unpaid after an in-app confirmation attempt.
+_Avoid_: Checkout
--- a/turbo/apps/api/src/__tests__/mocks.ts
+++ b/turbo/apps/api/src/__tests__/mocks.ts
@@ -197,6 +197,7 @@ export interface ApiTestMocks {
       readonly update: AsyncMock;
     };
     readonly subscriptions: {
+      readonly create: AsyncMock;
       readonly list: AsyncMock;
       readonly retrieve: AsyncMock;
       readonly update: AsyncMock;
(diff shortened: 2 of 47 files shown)


hulh122:
## Code review — round 1

I found three high-priority billing-safety issues:

1. **[P1] Reject retired usage-pack snapshots before confirmation.** `loadUsagePackPurchaseSnapshot` accepts a snapshot after a replacement preview has marked it `checkout_expired`. Its still-valid signed token can therefore create and charge a subscription for the superseded allocation set. Please validate the persisted lifecycle state and cover preview A → replacement preview B → confirm A.

2. **[P1] Attempt payment for open usage-pack invoices.** All three usage-pack calls to `completeBillingOperationInvoiceWithInvoice` omit `{ payOpenInvoice: true }`. An open first invoice from `default_incomplete` is consequently returned as hosted `pending_payment` without attempting the saved method, unlike the Plan path. Please make new and resumed usage-pack confirmations use the saved-card payment path and add an open-invoice regression test.

3. **[P1] Revalidate the saved method before paying a resumable Plan invoice.** `confirmPlanPurchaseTransaction` pays `subscriptionState.existing` before comparing the current route with the customer/payment-method IDs signed into the preview. If the saved method changes between preview and confirm, this branch can still initiate payment. Revalidate before paying any unpaid existing invoice and cover the changed-method case.

The additive API shapes and legacy `{ url }` fallback look deployment-compatible, and I found no type/lint suppressions, internal-module mocks, fake timers, or resource-lifecycle regressions.

hulh122:
## Review follow-up: LGTM

Re-reviewed the latest head (8ed4c96). All three first-round P1 findings are resolved:

- Retired usage-pack purchase snapshots are rejected.
- Open usage-pack invoices attempt the saved payment method.
- Resumable Plan purchases revalidate the current saved payment method before invoice payment.

Regression coverage was added for each case.

Validation completed:

- pnpm format
- pnpm turbo run lint
- pnpm check-types
- pnpm vitest run --maxWorkers=4 --silent=passed-only (646 files, 8778 tests passed; 1 existing skip)
- git diff --check

No P0 or P1 findings remain. The repository-wide Knip check still reports its existing baseline findings outside this PR (4 files, 1 devDependency, and 1502 exports); this change adds no dependencies or exports.

PR #27134, opened 2026-08-14
fix(python): require platform api scheme match

### LINKED ISSUE #27100: security(python): require platform API scheme match before injecting bypass credentials
## Problem and evidence

Platform API admission currently drops the configured URL scheme before deciding whether a
registered VM request receives privileged platform handling. In
`crates/runner/mitm-addon/src/upstream_admission.py`, `_api_destination()` reduces the configured
API URL to a normalized host and effective port, and `api_destination_matches()` compares only
those fields.

For non-HTTPS traffic, `get_trusted_authority()` preserves the request scheme while using the
transparent destination host and port. `request_classification.py` then passes only that host and
port into platform API matching. With `https://api.vm0.ai` configured, a plaintext request to
`http://api.vm0.ai:443` can therefore become `ApiAllow` even though it names a different origin.

That classification is privileged in two independent ways:

- it runs before normal compiled-firewall matching, so wrong-scheme traffic can bypass the policy
  that would otherwise classify it; and
- after upstream admission, `mitm_addon.py` adds the runner-owned
  `x-vercel-protection-bypass` value when configured.

The same incomplete identity is used by TLS/server-connect prebinding and by the special rule that
prevents ordinary connector credentials from being injected into platform API traffic without the
internal test-endpoint bypass. In addition, `ensure_bound_destination(kind="api_allow")` currently
does not revalidate the configured API identity before the bypass header is added; it trusts the
earlier classification and host/port binding.

The runner configuration already requires an absolute `http` or `https` API URL, and HTTP remains
supported for local development. The defect is therefore not that HTTP exists, but that the
configured and observed schemes are not required to agree.

## Impact

A registered VM can cause the runner-owned Vercel bypass value to be serialized onto a plaintext
connection for the configured HTTPS API host and port. The real HTTPS service will normally reject
that plaintext request, but rejection occurs after the credential-bearing bytes may have traversed
the network. The current behavior also violates the platform API firewall boundary even when no
bypass secret is configured.

## Decision

Treat the configured platform API identity as:

```text
normalized scheme + normalized host/subdomain policy + effective port
```

Require all three fields to match before returning `ApiAllow`, creating or reusing an `api_allow`
binding, or adding the Vercel bypass header. Apply the same identity to the platform-only
connector-auth restriction so a different-scheme origin follows ordinary connector behavior.

Keep generic upstream network bindings represented by host and port. Scheme is an admission
precondition for minting the privileged `api_allow` binding kind, not a property needed by every
connector binding. TLS ClientHello/server-connect paths have observed HTTPS transport and must
match it against the configured scheme explicitly.

- Preserve the normalized configured scheme in the platform API destination parser.
- Use one shared scheme/host/port predicate across request classification, requestheaders probing,
  TLS/server-connect prebinding, and connector-auth platform detection.
- Revalidate the full platform API identity inside final `api_allow` destination admission before
  mutating request headers. This keeps cached header-phase classifications from becoming the sole
  credential-injection gate.
- Preserve the existing exact-host-or-subdomain rule and effective-port behavior.
- Preserve configured HTTP API support when the observed request is also HTTP.

### Check only before adding the header

Rejected. It would prevent this credential disclosure but leave wrong-scheme traffic with
`ApiAllow`, bypassing normal firewall classification and creating privileged bindings.

### Change request classification only

Rejected. TLS/server-connect prebinding, connector-auth platform detection, and final admission
would retain a different platform identity contract. Final admission should independently protect
the credential mutation boundary.

### Reject every configured HTTP API

Rejected. HTTP API URLs are intentionally supported for local development. Exact scheme matching
fixes the downgrade without removing that supported mode.

### Add scheme to every generic upstream binding

Rejected. Those bindings represent a concrete network authority used by both platform and
connector paths. The required security property is that the `api_allow` kind cannot be granted or
used without a matching scheme; widening the generic binding model would add unrelated state and
call-site churn.

## Risks and mitigations

- **Previously accepted wrong-scheme traffic changes behavior:** it will follow normal firewall or
  default handling and will not receive platform credentials. This is the intended boundary.
- **Local HTTP development regression:** retain positive HTTP-to-HTTP coverage, including default
  port behavior.
- **Call-site disagreement:** centralize matching and cover request, requestheaders, and TLS
  prebinding entry points, plus the final admission sink.
- **Connector-auth regression on a different-scheme origin:** cover that the platform-only bypass
  restriction does not apply when the scheme differs.
- **Performance:** the path already parses and normalizes the configured API destination; retaining
  and comparing one normalized scheme adds constant work and no I/O.

## Acceptance criteria

- The platform API destination retains a normalized `http` or `https` scheme in addition to the
  existing normalized host and effective port.
- Request classification requires scheme, host/subdomain, and port to match before returning
  `ApiAllow`.
- Requestheaders probing uses the same full match and never prebinds or injects the bypass header
  for a scheme mismatch.
- TLS ClientHello/server-connect prebinding creates `api_allow` only when its observed HTTPS
  transport matches the configured API scheme.
- Final `api_allow` destination admission revalidates the current request scheme, trusted host, and
  port against the configured API before adding `x-vercel-protection-bypass`.
- The platform-only connector-auth bypass restriction uses the same scheme-aware identity.
- An HTTP request to the configured HTTPS host and port receives no `ApiAllow`, no `api_allow`
  binding, and no Vercel bypass header. The inverse mismatch is also rejected.
- Existing same-scheme behavior remains valid for configured HTTP and HTTPS URLs, explicit and
  default ports, the exact configured host, and intentionally supported subdomains.
- Focused request-, requestheaders-, TLS/server-connect-, and connector-admission integration tests
  cover the changed boundary.

## Delivery notes

This is one independently reviewable Python mitm-addon PR. It requires no schema, runner protocol,
configuration, migration, or documentation change. The broader outbound platform endpoint hostname
normalization tracked in #27115 is related but independent and does not block this security fix.

seven332:
## Summary

- require the observed request scheme, normalized API host or subdomain, and effective port to match the configured platform API origin
- apply the same origin check to request classification, streamed request headers, TLS prebinding, connector admission, and final bypass-header admission
- preserve same-scheme HTTP development support and keep generic connector bindings scoped to host and port

## Scope

- no database schema, runner protocol, configuration, migration, or documentation changes
- wrong-scheme traffic now follows ordinary firewall or connector behavior and never receives platform API credentials

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest -q tests/test_request_handler_api_admission.py tests/test_request_headers_api_admission.py tests/test_server_connect_hook.py tests/test_tls_clienthello_hook.py tests/test_request_handler_connector_admission.py` (95 passed)
- `uv run --no-sync python -m pytest tests/` (5403 passed)

Closes #27100

### DIFF
--- a/crates/runner/mitm-addon/src/mitm_addon.py
+++ b/crates/runner/mitm-addon/src/mitm_addon.py
@@ -355,8 +355,9 @@ def _prebind_bounded_requestheaders_upstream_destination(
         flow.metadata[metadata_keys.TRUSTED_AUTHORITY_HOST] = trusted_authority.host
         if upstream_admission.api_destination_matches(
             api_url,
-            trusted_authority.host,
-            trusted_authority.port,
+            scheme=flow.request.scheme,
+            hostname=trusted_authority.host,
+            port=trusted_authority.port,
         ) and not upstream_admission.request_path_uses_platform_firewall(flow.request.path):
             classification = _classify_request_for_flow_with_trusted_authority(
                 flow,
--- a/crates/runner/mitm-addon/src/request_classification.py
+++ b/crates/runner/mitm-addon/src/request_classification.py
@@ -416,8 +416,9 @@ def _classify_request(
 
     if upstream_admission.api_destination_matches(
         api_url,
-        trusted_authority.host,
-        trusted_authority.port,
+        scheme=flow.request.scheme,
+        hostname=trusted_authority.host,
+        port=trusted_authority.port,
     ) and not upstream_admission.request_path_uses_platform_firewall(flow.request.path):
         return ApiAllow(vm_info=vm_info)
 
--- a/crates/runner/mitm-addon/src/upstream_admission.py
+++ b/crates/runner/mitm-addon/src/upstream_admission.py
@@ -38,6 +38,13 @@ class TlsAdmission:
     sni: str | None = None
 
 
+@dataclass(frozen=True)
+class _ApiDestination:
+    scheme: Literal["http", "https"]
+    host: str
+    port: int
+
+
 def _client_connection_id(client: object) -> str | None:
     client_id = getattr(client, "id", None)
     if isinstance(client_id, str) and client_id:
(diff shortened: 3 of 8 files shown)


seven332:
## Code Review: PR #27134

**Reviewed head:** `5284038895992e00a6f300943e064e68caf9fdb9`

### Summary

The change consistently makes the configured platform API scheme part of privileged origin admission across request classification, streamed request headers, TLS prebinding, connector handling, and the final bypass-header gate. It preserves same-scheme HTTP development support and does not widen generic connector binding state.

### Findings

No findings.

### Testing

Ruff format/lint, BasedPyright, uv lock/sync validation, 95 focused integration tests, and the full 5403-test mitm-addon suite passed locally. Unrelated Turbo and Rust suites were not run for this isolated Python change; GitHub CI is evaluated separately.

### Verdict

LGTM

PR #27171, opened 2026-08-14
test(e2e): decouple custom connector probe from upstream status

### LINKED ISSUE #27166: test: stop custom connector e2e from depending on external response status
## Background

The deployed runner E2E test `run-t23-custom-connector-webhook.bats` creates a manual custom HTTP connector, stores a synthetic secret, grants the connector to an agent, and sends a sandbox request through the runner firewall.

The test currently uses an anonymous Webhook.site token and runs `curl --fail`. This makes the external service's response status part of the test oracle, even though runner E2E guidance requires assertions to use sandbox-visible output and vm0-owned telemetry rather than an external provider's exact response status or body.

## Problem

Webhook.site intermittently returns HTTP errors. In [GitHub Actions job 94702509765](https://github.com/vm0-ai/vm0/actions/runs/31779300911/job/94702509765), the outbound request returned HTTP 502. `curl --fail` exited with status 22 before printing the synchronization marker, and the test then waited about 90 seconds for a marker that could never appear.

The HTTP status alone is ambiguous: an upstream service and the local runner MITM/firewall auth path can both return 502. vm0 network telemetry already exposes the fields needed to distinguish them:

- the matched custom connector firewall
- `action`
- `firewall_error`
- `auth_resolved_secrets`
- the upstream status or network error

Successful expected `auth_resolved_secrets` are recorded only after resolved authentication has been validated and applied to the request. Local auth failures carry `firewall_error` and do not establish that successful path.

## Requirements

- Do not use the external target's exact HTTP response status or body as the runner E2E test oracle.
- Keep the custom connector setup, grant, sandbox request, and credential-resolution path covered through supported product entry points.
- Use vm0-owned telemetry to distinguish successful credential application followed by an upstream HTTP error from a local firewall/auth failure.
- Do not hide local firewall/auth failures merely because the sandbox observed HTTP 502.
- Avoid unnecessary anonymous Webhook.site token and capture dependencies when equivalent behavior is already covered by API integration, runner MITM, and vm0 telemetry assertions.

## Acceptance criteria

- The test does not fail solely because an external target returns HTTP 4xx or 5xx, including 502.
- The sandbox command always emits its completion marker after making the request, independently of the external HTTP status.
- The test passes only when telemetry identifies the expected custom connector firewall, an allowed request, the expected resolved secret, and no local `firewall_error`.
- The test still fails when credential resolution or application fails locally.
- External token creation, capture polling, and teardown are removed if they are no longer required by the resulting assertions.
- Applicable shell/static checks pass; the deployed runner E2E shard validates the full behavior in CI.

## Relevant coverage

- API integration tests verify custom connector runtime firewall and secret-reference generation.
- Runner MITM tests verify resolved custom connector headers are applied to outgoing requests.
- Existing runner firewall E2E tests ignore external response outcomes and assert sandbox markers plus vm0 network telemetry.

seven332:
## Summary

- remove the anonymous Webhook.site token, capture polling, and teardown from the custom connector runner E2E
- make the sandbox completion marker independent of the external target's HTTP status
- require successful firewall telemetry to contain no local `firewall_error`
- rename the test to reflect its provider-independent behavior

## Behavior

The test still creates and configures a custom connector through public APIs, grants it to an agent, sends a sandbox request through the connector firewall, and checks the expected resolved secret in vm0 network telemetry. It now uses an external URL already covered by the runner E2E suite and treats the upstream response as non-authoritative.

This PR does not change production connector, firewall, API, or runner behavior.

## Validation

- `shellcheck -x e2e/tests/03-runner/run-t23-custom-connector.bats`
- `bash -n e2e/helpers/runner-api.bash`
- `shellcheck -x --exclude=SC2034 e2e/helpers/runner-api.bash` (the excluded sourced-variable warning predates this change)
- BATS parse/count check: 1 test
- helper probe: successful telemetry accepted; identical telemetry with `firewall_error=auth_failed` rejected
- `cd e2e && pnpm check-types`
- `cd e2e && pnpm test` (35 passed)
- runner shard generation/discovery check
- `.github/scripts/tests/turbo-playwright-runner-workflow-test.sh`

## Risk

The shared firewall-log helper is stricter for all callers. Each caller uses it for a successful firewall/auth path, so rejecting a local `firewall_error` matches the helper's existing contract. The deployed runner E2E shard remains the full end-to-end validation.

Closes #27166

### DIFF
--- a/e2e/helpers/runner-api.bash
+++ b/e2e/helpers/runner-api.bash
@@ -473,6 +473,7 @@ runner_e2e_wait_for_firewall_log() {
                     .firewall_name == $firewallName and
                     .host == $host and
                     .action == "ALLOW" and
+                    (.firewall_error // null) == null and
                     (((.auth_resolved_secrets // []) | sort) == ($expectedSecrets | sort)) and
                     ($expectedUrlRewrite == "ignore" or
                         .auth_url_rewrite == ($expectedUrlRewrite == "true")))' \
--- a/e2e/tests/03-runner/run-t23-custom-connector.bats
+++ b/e2e/tests/03-runner/run-t23-custom-connector.bats
@@ -8,7 +8,6 @@ setup() {
     runner_e2e_require_environment
     runner_e2e_setup_test
     CUSTOM_CONNECTOR_ID=""
-    WEBHOOK_TOKEN=""
 }
 
 teardown() {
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #27171

**Reviewed head:** `bbdebf248908595707fc9b316fa6323d5f45f9a9`

### Summary

The change removes the brittle Webhook.site capture lifecycle and makes vm0-owned firewall telemetry the authoritative E2E result. The public custom-connector setup/grant/run path remains covered, and local auth/firewall failures cannot be hidden by an upstream 502 because successful telemetry now requires no `firewall_error`.

### Findings

No findings.

### Testing

Shell/BATS parsing and static checks, helper success/failure semantics, E2E type checking and 35 package tests, shard discovery, and the runner workflow contract all passed locally. The deployed `cli-e2e-03-runner` execution remains owned by PR CI.

### Verdict

LGTM

PR #26646, opened 2026-08-12
refactor: remove sandbox presentation import pipeline

bingjiez666:
## Summary

- remove the pre-GA presentation-template API, CLI, callback, upload, and package protocol that launched server-side analysis
- remove the LibreOffice/PDF conversion prompt and the `libreoffice-impress` / `poppler-utils` runner rootfs dependencies
- remove obsolete contracts, trigger metadata, storage naming, translations, tests, and CLI cutover entries
- retain the additive `presentation_templates` schema and `PresentationTemplates` feature switch for the browser-rendered replacement

## Why

The design in #21350 now renders PPTX pages in the browser and commits the original PPTX plus browser-produced PNGs before analysis begins. The deleted pipeline encoded the superseded sandbox conversion and analyzer-first ordering.

A read-only masked production check found zero `presentation_templates` rows and zero runs with `trigger_source = template-import`, so there is no persisted or in-flight production state to migrate.

`aspect_ratio` is intentionally retained here. Dropping a live database column while older API revisions can still be draining is deployment-incompatible; #26578 handles that cleanup after this API removal is deployed.

## Runner artifact note

This removes the direct LibreOffice and Poppler package set and its rootfs verification. Local two-architecture rootfs builds were not run; #26577 remains open until CI-backed disk-size and build-time deltas are recorded.

## Verification

- `pnpm -F @vm0/okou-cli exec vitest run src/__tests__/okou.test.ts src/__tests__/zero-capability-visibility.test.ts` — 91 tests passed
- `pnpm -F @vm0/okou-cli build`
- built CLI help contains no `presentation-template`; direct invocation returns `unknown command`
- affected `lint` and `check-types` Turbo run — 21/21 tasks passed
- `pnpm -F @vm0/api-contracts check-types` after the final rebase
- `pnpm knip`
- `cargo fmt --manifest-path crates/Cargo.toml --all -- --check`
- `CARGO_BUILD_JOBS=1 CARGO_PROFILE_DEV_DEBUG=0 cargo check --manifest-path crates/Cargo.toml -p runner --tests`
- shell syntax checks for all modified runner/audit scripts
- repository residual scan for the removed protocol and conversion tools

The targeted runner test binary compiled but its final local link was killed by the 3.8 GiB sandbox memory limit; the no-link `cargo check --tests` completed successfully. CI remains the source of truth for the linked runner tests and both rootfs architectures.

Refs #21350
Prepares #26571
Refs #26577
Follow-up #26578

PR #27172, opened 2026-08-14
feat(api): add private presentation template records

bingjiez666:
## Summary

- add feature-gated owner list, get, rename, and delete routes for private presentation template records
- store only the original PPTX key and filename plus ordered PNG keys; keep URLs short-lived, omit persisted byte sizes, and keep `aspectRatio` out of the contract
- sign reads directly from committed storage keys without an R2 HEAD preflight, so missing or replaced objects surface when the signed URL is actually fetched
- add the lifecycle lock needed by the stacked atomic commit implementation and preserve independently owned source/page uploads when a template record is deleted

The final schema matches the existing database shape after removing the unmerged `source_size_bytes` and `page_sizes_bytes` proposal, so this PR intentionally contains no generated migration or snapshot changes.

## Upload transport

This foundation deliberately does **not** add a presentation-specific prepare/upload protocol or change the shared S3 presigner. The browser uses the existing authenticated `/api/okou/uploads/prepare` flow once for the PPTX and once per PNG, then passes those file IDs to the commit endpoint added by #26947.

The template flow does not call ordinary `/uploads/complete`, so page screenshots are not registered in `run_uploaded_files` or projected into the artifact catalog.

## Stack

This is the independently reviewable foundation for #26947. The stacked PR adds `sourceFileId + ordered pageFileIds` commit validation, exact-once analysis launch, run-scoped reads, callbacks, and private package publication.

The scope is PPTX only, fixed 16:9, with browser-generated PNG pages. It does not depend on browser renderer internals from #26571 and does not restore the sandbox conversion path removed by #26646.

## Verification

- changed-file Prettier write and check
- `@okouai/api-contracts`, `@okouai/db`, and `api` typechecks
- `pnpm db:generate` (`No schema changes`) and migration against temporary PostgreSQL
- `pnpm exec vitest run src/signals/routes/__tests__/zero-presentation-templates.test.ts` (2 passed on the base layer)
- `pnpm knip`

Refs #26576
Refs #21350

### DIFF
--- a/turbo/apps/api/src/signals/route.ts
+++ b/turbo/apps/api/src/signals/route.ts
@@ -179,6 +179,7 @@ import { teamRoutes } from "./routes/team";
 import { uploadsCompleteRoutes } from "./routes/uploads-complete";
 import { uploadsMultipartRoutes } from "./routes/uploads-multipart";
 import { uploadsPrepareRoutes } from "./routes/uploads-prepare";
+import { zeroPresentationTemplatesRoutes } from "./routes/zero-presentation-templates";
 import { usageMembersRoutes } from "./routes/usage-members";
 import { usageRecordRoutes } from "./routes/usage-record";
 import { userPreferencesRoutes } from "./routes/user-preferences";
@@ -379,6 +380,7 @@ export const ROUTES: readonly RouteEntry[] = [
   ...uploadsCompleteRoutes,
   ...uploadsMultipartRoutes,
   ...uploadsPrepareRoutes,
+  ...zeroPresentationTemplatesRoutes,
   ...registryResourceDownloadRoutes,
   ...usageMembersRoutes,
   ...usageRecordRoutes,
--- a/turbo/apps/api/src/signals/routes/__tests__/zero-presentation-templates.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/zero-presentation-templates.test.ts
@@ -1,0 +1,84 @@
+import { randomUUID } from "node:crypto";
+
+import { zeroPresentationTemplatesContract } from "@okouai/api-contracts/contracts/zero-presentation-templates";
+import { FeatureSwitchKey } from "@okouai/core/feature-switch-key";
+import { beforeEach, describe, expect, it } from "vitest";
+
+import { accept, testContext } from "../../../__tests__/test-context";
+import { setupApp } from "../../../__tests__/test-helpers";
+import { mockEnv } from "../../../lib/env";
+import { createBddApi, type ApiTestUser } from "./helpers/api-bdd";
+import { updateFeatureSwitchesForUser } from "./helpers/feature-switches";
+import { createZeroRouteMocks } from "./helpers/zero-route-test";
+import { zeroPresentationTemplatesRoutes } from "../zero-presentation-templates";
+
+const context = testContext();
+const bdd = createBddApi(context);
+const mocks = createZeroRouteMocks(context);
+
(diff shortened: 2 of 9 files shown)


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 792b6945e0ee1685c7f5c13e4a60fe9d8e060034
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Adds feature-gated owner `list`/`get`/`update`(rename)/`delete` routes for private presentation template records over the existing `presentation_templates` schema. Reads sign short-lived public presigned URLs from committed storage keys without an R2 HEAD preflight, and delete marks the row failed under a lifecycle advisory lock before removing it so independently owned source/page uploads are preserved.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | New user-facing feature gated behind a `FeatureSwitchKey` | Pass | `zero-presentation-templates.ts` gates every route through `presentationTemplatesEnabled$` on `FeatureSwitchKey.PresentationTemplates` (registry `enabled: false`) |
| `REVIEW.md` | Auth checks on new endpoints | Pass | `templateReadAuth` (`agent:read`) / `templateWriteAuth` (`agent:write`), both `requireOrganization` |
| `REVIEW.md` | No-existence-leak and owner scoping | Pass | read/update/delete filter by `orgId` + `ownerUserId` + `id`, returning 404 for foreign/unknown records |
| `REVIEW.md` | `feat:` commit has integration tests | Pass | `zero-presentation-templates.test.ts` exercises routes through `setupApp` |
| `REVIEW.md` | DB/JSONB contract change needs migration or rationale | Pass | schema columns unchanged; only comments updated; PR states `pnpm db:generate` reports no schema changes |
| `docs/bad-smell.md` | No `any`, suppressions, dynamic imports, hardcoded config | Pass | typed drizzle/ccstate; config read via `env()`; no suppression comments |
| `docs/fallback.md` | No compatibility fallback for a non-GA feature switch; no negative tests | Pass | `PresentationTemplates` is `enabled:false` (non-GA); `aspectRatio` legacy column is only re-commented, not a new fallback; no negative tests added |
| `docs/deployment-compatibility.md` | Cross-version boundary safety | Pass | additive routes only; no migration or persisted-shape change |
| `.claude/skills/ccstate/SKILL.md` | Reactive `computed` / `command` with explicit `AbortSignal` ownership | Pass | read routes are `computed`, writes are `command`, `signal.throwIfAborted()` after non-signal-aware awaits |
| `docs/testing.md` | Integration tests, real DB, mock only the boundary | Pass | real DB via `testContext()`; only Clerk session and env are mocked |
| `docs/testing/api-testing.md` | Route tests via `setupApp({ context, routes })(contract)` | Pass | test uses `setupApp` with `zeroPresentationTemplatesRoutes` and the contract |
| `docs/testing/testing-external-behavior.md` | Construct and verify through the API surface | Pass | feature switch enabled through the real feature-switches API; no DB schema/service imports |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

### Testing

- Coverage: Adequate for the routes constructible through this PR's API surface — feature-switch gating (403 → 200) and no-existence-leak (404) on `get`/`update`/`delete`. Happy-path `list`/`get`/`update`/`delete` success responses and cross-owner denial cannot be constructed through this PR because the commit route that creates templates lands in the stacked #26947; that coverage should follow there rather than seeding rows directly.
- Conventions: Compliant — integration tests over the real Hono app and real DB, mocking only Clerk and env.

PR #27176, opened 2026-08-14
fix(chat): avoid locking threads during search projection

e7h4n:
## Summary

- replace the search projector's per-thread `SELECT ... FOR UPDATE` with a plain read, so projection work no longer holds a write-conflicting lock on `chat_threads`
- keep overlapping ticks idempotent through the existing projection primary key and monotonic watermark; the rare overlapping revoke race remains an eventual-consistency tradeoff, with no reconciler added in this PR
- add a concurrency regression test that holds an event-write transaction open, proves projection can still finish, then verifies the next tick catches up the committed event

## Why

Search projection can spend seconds writing GIN-backed message rows. Holding the thread row for that whole transaction blocks event writers when they advance `last_chat_event_seq_id`, which can surface as lock timeouts on chat callbacks.

## Verification

- `pnpm format`
- `pnpm knip`
- staged workspace, platform, and API type checks
- `DATABASE_URL=postgresql://postgres@localhost:5432/vm0_test pnpm --filter api exec vitest run src/signals/routes/__tests__/cron-project-chat-event-search.test.ts` (5 tests passed)

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/cron-project-chat-event-search.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/cron-project-chat-event-search.test.ts
@@ -2,7 +2,7 @@ import { randomUUID } from "node:crypto";
 
 import { cronProjectChatEventSearchContract } from "@okouai/api-contracts/contracts/cron";
 import { testChatEventSearchProjectionContract } from "@okouai/api-contracts/contracts/test-chat-event-search-projection";
-import { describe, expect, it } from "vitest";
+import { describe, expect, it, onTestFinished } from "vitest";
 
 import { accept, testContext } from "../../../__tests__/test-context";
 import { setupApp } from "../../../__tests__/test-helpers";
@@ -13,6 +13,7 @@ import {
   readChatEventSearchProjectionFixture,
   rejectSearchablePromptFixture,
 } from "../../../test-fixtures/chat-event-search";
+import { holdChatEventInsertTransactionFixture } from "../../../test-fixtures/chat-events";
 import { cronProjectChatEventSearchRoutes } from "../cron-project-chat-event-search";
 import { testChatEventSearchProjectionRoutes } from "../test-chat-event-search-projection";
 import { createBddApi } from "./helpers/api-bdd";
@@ -102,7 +103,7 @@ describe("GET /api/cron/project-chat-event-search", () => {
     expect(projection.indexedSeqId).toBe(projection.lastChatEventSeqId);
   });
 
-  it("serializes overlapping projection ticks idempotently", async () => {
+  it("keeps overlapping projection ticks idempotent", async () => {
     const chatThreadId = await createProjectionThread();
     const assistantText = `overlap assistant ${randomUUID()}`;
     await insertChatSearchProjectionCoverageFixture({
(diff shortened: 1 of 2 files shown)


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 0a8d3acc38ef375037594e6f16cd91c20320de2c
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

This removes the search projector's explicit per-thread `FOR UPDATE` lock, so the transaction's GIN-backed projection writes no longer hold a writer-conflicting lock on `chat_threads`. The existing `(chat_thread_id, seq_id)` primary key and `GREATEST` watermark update preserve idempotent, monotonic progress for overlapping ticks, and the regression test covers an event writer held open across a projection tick plus catch-up after commit.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and concurrency | Pass | `loadProjectionThread` now performs a plain committed read; events committed after that read remain above the observed watermark and are picked up by a later tick. Inserts remain conflict-safe and watermark updates remain monotonic. |
| `REVIEW.md` | Security, type safety, and public interfaces | N/A | The diff changes no auth boundary, external contract, schema, secret handling, or public type; it adds no unsafe cast. |
| `REVIEW.md` | A `fix:` change has regression coverage | Pass | `cron-project-chat-event-search.test.ts` holds a real event-write transaction open, proves projection completes while it is held, then verifies the committed event is indexed by the next tick. |
| `REVIEW.md` | Fallback discipline | N/A | The diff introduces no defaulting, compatibility branch, feature-switch fallback, or alternate recovery path. |
| `docs/bad-smell.md` | Avoid `any`, suppressions, dynamic imports, hardcoded configuration, and fallback slop | Pass | The production change is a focused query/identifier rename and adds none of the prohibited constructs. |
| `docs/testing.md` | Prefer integration coverage with real internals and outcome assertions | Pass | The regression runs the route through `setupApp`, uses the real database and event writer, and asserts projection/catch-up outcomes rather than mock calls. |
| `docs/event-sourcing.md` | Persistent events remain authoritative and later synchronization restores projection state | Pass | The projector only reads committed `chat_events`; an in-flight event is not skipped permanently because the watermark stays at the observed committed sequence and the next tick catches up. |
| `.claude/skills/ccstate/SKILL.md` | Await async work and keep lifecycle cancellation owned | Pass | Projector work remains awaited inside the command path with `signal.throwIfAborted()` checks; the held test transaction is tied to `context.signal` and released in test cleanup. |
| `docs/testing/patterns.md` | Use route integration tests and real database behavior | Pass | The test uses the cron/test projection contracts with `setupApp`; no internal service is mocked. |
| `docs/testing/anti-patterns.md` | Avoid internal mocks, fake timers, and implementation-call assertions | Pass | The added test uses none of those patterns and verifies durable rows/watermark behavior. |
| `docs/testing/api-testing.md` | Exercise the API route slice, preserve test ownership, and clean up held resources | Pass | The test names only its UUID-owned thread, calls the route contract, and registers cleanup for both the writer transaction and in-flight tick. |
| `docs/testing/testing-external-behavior.md` | Stay at the external route boundary except for states impossible to construct there | Pass | Projection is invoked through the test route; the narrow fixture is documented because no production endpoint can pause between event insertion and transaction commit, which is the concurrency boundary under test. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- Coverage: Adequate. The new regression covers the writer-lock boundary and subsequent catch-up, while the existing overlapping-tick test continues to cover idempotency. Codecov reports all modified coverable lines covered.
- Conventions: Compliant. The test uses the route contract, real database behavior, unique thread ownership, no internal mocks, and no fake timers.
- Execution: The targeted file passed locally with 5 tests on this head. At review time, Turbo lint/type/API-test jobs and `codecov/patch` are successful; remaining repository-wide CLI E2E jobs are still running.

PR #27185, opened 2026-08-14
fix(guest-agent): preserve descendant cleanup without pidfd

### LINKED ISSUE #27150: bug(rust): kill app-server descendants in pidfd fallback
## Problem and evidence

At commit `dfcfdc33db89cb9c2f94a4eae69c4e65a7a51049`, `CodexAppServerClient` uses a direct `child.wait()` fallback when `ChildExitNotifier::open` is unavailable or a pidfd readiness wait later fails.

That wait reaps the process-group leader and `finish_child_wait` clears `self.child` before poisoned-stream cleanup derives the process group. A descendant that inherited stdout or stderr can therefore keep the pipe open and survive cleanup. The existing `exit-on-turn-start-with-stderr-holder` integration scenario covers this lifecycle shape only through the successful pidfd path.

PR #21233 deliberately made process-group signaling derive only from the currently owned, unreaped child. Retaining a raw PID/PGID after wait would reintroduce the stale-identity race that change removed.

## Impact

Unexpected Codex app-server exits can leak descendants whenever pidfd setup or notification is unavailable. Leaked processes consume guest resources and can force shutdown to wait for or abandon inherited stderr collection.

## Decision

Keep pidfd as the preferred event-driven observer. Replace the direct-wait fallback with a degraded-mode async poller using `waitid(P_PID, ..., WEXITED | WNOHANG | WNOWAIT)` at a 50 ms interval.

Both observer modes will converge on one ownership-safe sequence: observe terminal state without reap, derive and signal the process group from the still-owned child, then call `child.wait()` and clear child/notifier ownership. A pidfd runtime error clears the notifier and selects the same waitid fallback used after pidfd setup failure.

If waitid itself fails, poison and terminate while the direct child remains owned; never wait first and signal a copied numeric identity afterward. Add a per-config, hidden integration seam that disables only pidfd selection so Linux CI can exercise the real fallback deterministically.

## Alternatives considered

- **Blocking `waitid(WNOWAIT)` observer task:** preserves the invariant but introduces a non-cancellable blocking task, join/error states, and shutdown overlap for every degraded client.
- **Replace pidfd entirely with waitid:** creates normal-path polling or blocking-task overhead and removes the healthy behavior this issue must preserve.
- **Cache PGID across direct wait:** smallest diff, but invalid because reap releases signaling authority and any later validation remains a check-then-signal race.
- **Dedicated cgroup cleanup:** provides broader descendant containment but requires guest lifecycle and rollout work beyond this client-local fallback bug.

## Risks and mitigations

- Fallback polling adds at most 20 nonblocking observations per second and no overhead when pidfd works.
- Degraded-mode exit detection gains at most 50 ms latency, within existing client timeout bounds.
- The fallback future is cancellation-safe between polls and creates no detached blocking work.
- A shared pre-reap cleanup/reap helper prevents pidfd and waitid success ordering from drifting.
- The deterministic seam is per-config, hidden from generated docs, and cannot inject or retain a PID/PGID.

## Acceptance criteria

- pidfd setup failure and pidfd readiness failure both select non-reaping waitid observation while the client still owns the direct child.
- Successful natural-exit observation signals the complete app-server process group before reaping the leader.
- No process-group target remains signalable after successful child wait, and no fallback uses a cached post-reap PID/PGID.
- A deterministic Linux integration test disables pidfd, triggers a stdio-holding descendant, and verifies `ChildExited`, cleared child identity, complete process-group exit, and shutdown within the existing 1.5-second bound.
- The existing pidfd integration scenario and other app-server shutdown, terminate, poison, cancellation, and Drop behavior remain green.
- Affected Rust formatting, Clippy, Rustdoc, integration tests, and pre-commit checks pass.

## Delivery notes

This remains one focused Rust PR. No database migration, API contract, queue payload, runner protocol, persisted state, cgroup design, or cross-version rollout coordination is required.

seven332:
## Summary

- add a non-reaping `waitid` fallback when pidfd setup or readiness is unavailable
- signal the owned app-server process group before reaping on both exit-observer paths
- add deterministic real-process coverage for the pidfd-disabled stdio-holder lifecycle

## Safety

- no cached PID or PGID remains signalable after the child is reaped
- production continues to prefer pidfd; the degraded fallback polls every 50 ms
- fallback observation errors poison and terminate the session while the `Child` remains owned

## Scope

- no database, API, queue, runner protocol, persisted-state, cgroup, or mock-scenario changes

## Validation

- `cargo test -p guest-agent --test codex_app_server_client stdio_holding_descendant_before_reap`
- `cargo test -p guest-agent --all-targets`
- `RUSTDOCFLAGS='-D warnings' cargo doc -p guest-agent --all-features --no-deps`
- repository pre-commit hooks

Closes #27150

### DIFF
--- a/crates/guest-agent/src/cli/child_exit_notifier.rs
+++ b/crates/guest-agent/src/cli/child_exit_notifier.rs
@@ -1,9 +1,12 @@
 use std::io;
+use std::time::Duration;
 
 #[cfg(target_os = "linux")]
 use std::os::fd::OwnedFd;
 
 use tokio::process::Child;
+
+const FALLBACK_EXIT_POLL_INTERVAL: Duration = Duration::from_millis(50);
 
 pub(super) struct ChildExitNotifier {
     #[cfg(target_os = "linux")]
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #27185

**Reviewed head:** `ace4de98b30c265bc7a0a5f112214f3931cd621c`

### Summary

The non-reaping `waitid` observer preserves the owned-child identity until same-group descendants are signaled, then converges with the pidfd path on one reap sequence. The integration coverage exercises both preferred and pidfd-disabled paths with real child processes and bounded cleanup.

The diff replaces the existing pidfd-degraded direct-wait branch; it does not introduce a cross-version compatibility fallback. Its 50 ms polling cost is isolated to the already-reachable degraded mode.

### Findings

No findings.

### Testing

Focused regression tests, all guest-agent targets, Rust formatting, Clippy, Rustdoc, diff checks, and repository pre-commit hooks passed. The post-setup pidfd readiness-error transition was inspected statically: it clears the notifier and enters the same tested non-reaping path.

### Verdict

LGTM

PR #27189, opened 2026-08-14
test(runner): assert prompt axiom shutdown after rejection

### LINKED ISSUE #27087: test(rust): assert prompt Axiom shutdown after ingest rejection
## Problem

`non_success_ingest_response_does_not_hang_shutdown_or_panic` says that an immediate Axiom 500 response must let shutdown complete "well within `FLUSH_DEADLINE`", but the test directly awaits `guard.shutdown()` without an independent timeout or elapsed-time assertion:

- `crates/runner/src/axiom_layer_tests.rs:456-484`
- `crates/runner/src/axiom_layer.rs:36-45`
- `crates/runner/src/axiom_layer.rs:87-100`

`AxiomGuard::shutdown` already wraps the close/send-and-join sequence in the production 15-second `FLUSH_DEADLINE` and discards both the timeout result and the dispatcher's `JoinHandle` result. Therefore, a regression that stalls after receiving the 500 response can consume the full production deadline, return, and still satisfy the current call-count and diagnostic assertions. A dispatcher panic after emitting the diagnostic can likewise be hidden by the discarded join result.

This leaves runner shutdown latency and dispatcher task failures unprotected even though the test name and comments claim those behaviors are covered.

## Acceptance criteria

- The immediate-500 test distinguishes prompt dispatcher completion from expiration of the production 15-second shutdown fallback, using a deterministic bound that fails comfortably before `FLUSH_DEADLINE`.
- The test observes and asserts successful dispatcher task completion so a panic or join failure cannot be treated as a successful shutdown.
- The existing contract remains covered: exactly one ingest request is attempted and the local `INTERNAL_TARGET` diagnostic records the non-success response.
- The test uses explicit synchronization or timeout assertions and does not rely on arbitrary sleeps.

seven332:
## Summary

- bound the immediate Axiom 500 shutdown path with an independent five-second test deadline
- await the real dispatcher task result so panic or cancellation fails the test
- retain the one-request and local `INTERNAL_TARGET` diagnostic assertions

## Scope

- test-only change in the runner Axiom layer tests
- no production shutdown behavior, API, dependency, or rollout changes

## Validation

- `cargo check -p runner --tests`

The focused `cargo test` was not executed locally because linking the runner test binary exceeded the 4 GiB container memory limit and triggered an OOM reset. The PR pipeline will execute the test suite in the repository CI environment.

Closes #27087

### DIFF
--- a/crates/runner/src/axiom_layer_tests.rs
+++ b/crates/runner/src/axiom_layer_tests.rs
@@ -454,6 +454,8 @@ fn burst_past_channel_cap_drops_without_blocking() {
 
 #[tokio::test]
 async fn non_success_ingest_response_does_not_hang_shutdown_or_panic() {
+    const TEST_SHUTDOWN_DEADLINE: Duration = Duration::from_secs(5);
+
     let server = MockServer::start_async().await;
 
     // Return 500 for every ingest. The dispatcher should log via
@@ -467,8 +469,12 @@ async fn non_success_ingest_response_does_not_hang_shutdown_or_panic() {
         })
         .await;
 
-    let (layer, guard) =
+    let (layer, mut guard) =
         init_with_base_url(&server.base_url(), "t", "test").expect("init must succeed");
+    let dispatcher_task = guard
+        .handle
+        .take()
+        .expect("init must spawn a dispatcher task");
     let recording = RecordingLayer::default();
     let subscriber = tracing_subscriber::registry()
         .with(recording.clone())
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #27189

**Reviewed head:** `ab25dc08723c48a165a326cce097283a53e38374`

### Summary

The change adds an independent five-second bound to the immediate Axiom 500 shutdown test and explicitly awaits the real dispatcher task result. It preserves the existing request-count and local diagnostic assertions without changing production behavior.

### Findings

No findings.

### Testing

- Passed locally: formatting, `cargo check -p runner --tests`, runner Clippy across all targets/features, and runner rustdoc.
- Unverified locally: focused test execution, because linking the runner test binary exceeded the 4 GiB container memory limit and caused an OOM reset. PR CI remains the execution gate.

### Verdict

LGTM

PR #27204, opened 2026-08-14
fix(chat): allow sparse event sequences

e7h4n:
## Summary

- define chat event `seqId` as a strictly increasing cursor that may start above 1 and contain gaps
- stop reclaiming rejected write reservations and allow snapshot archives to begin at the first physical event
- remove the Platform backfill-loading state and let the CLI preserve monotonic local histories across gaps
- replace one-based and contiguous test assumptions with sparse-stream coverage

## Verification

- `pnpm -F @okouai/cli exec vitest run src/commands/zero/chat/__tests__/messages.test.ts`
- `pnpm -F @okouai/app exec vitest run src/signals/chat-page/__tests__/chat-event-snapshot-read.test.ts`
- targeted API tests for concurrent idempotent sends, repeated callback batches, sparse snapshot publishing, and recall queue races
- `pnpm -F @okouai/cli check-types`
- `pnpm -F api check-types`
- affected-file Prettier, Oxlint, and ESLint checks


e7h4n:
LGTM

PR_AUTO_REVIEW_COMMIT: 019786be3c973545b4be60ecc1f86481c6ebc5d3
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

This change consistently defines chat-event `seqId` as a strictly increasing cursor rather than a dense ordinal. Reservation conflicts now leave gaps, snapshot rebuilds accept an archive whose first physical row is above 1, the CLI preserves sparse local histories while rejecting non-increasing pages, and Platform removes the obsolete first-event backfill indicator. The API, CLI, Platform, contract, and schema comments agree on the same invariant.

No P0 or P1 findings. The patch does not change an API or persisted-data shape and adds no migration fallback. Sparse sequences were already observable through rejected interior batch writes and archived prefixes; the readers now explicitly accept both dense and sparse histories while retaining strict ordering validation.

### Review Criteria

| --- | --- | --- |
| `REVIEW.md` | Pass | Correctness, idempotency, data integrity, security scope, and regression coverage were reviewed across all changed API, CLI, Platform, contract, and DB files. Atomic per-thread reservation remains the ordering authority; rejected idempotent writes no longer mutate the cursor backwards. No auth or authorization boundary changes. |
| `docs/docs.md` | Pass | The relevant subsystem, state-management, testing, fallback, event-sourcing, cache, and deployment-compatibility guidance was followed and the public contract/schema comments were updated with the new invariant. |
| `docs/bad-smell.md` | Pass | No `any`, suppression directive, dynamic import, broad fallback, duplicated compatibility implementation, or unrelated cleanup was introduced. |
| `docs/testing.md` | Pass | The fix has regression coverage at production-facing boundaries: API routes and DB transactions, CLI command plus filesystem/HTTP, and Platform snapshot storage/sync. |
| `docs/fallback.md` | Pass | No new fallback is introduced. The removed first-sequence heuristic and its UI/locale surface are deleted together, and sparse-stream behavior is covered positively rather than by a tombstone test. |
| `docs/event-sourcing.md` | Pass | Persisted chat events remain the canonical stream and are read in ascending `seqId` order; the change relaxes density only and does not alter optimistic-event reconciliation. |
| `docs/effect.md` | Pass | No new effect or lifecycle owner is introduced. Obsolete derived backfill state and its sole UI consumer are removed together. |
| `docs/cache.md` | N/A | No cache, subscription, or disposable resource is added or changed. |
| `.claude/skills/ccstate/SKILL.md` | Pass | The exported computed state, public signal projection, and UI consumer are removed in lockstep; no new command/effect boundary or detached work is added. |
| `docs/deployment-compatibility.md` | Pass | Wire and storage shapes are unchanged. Dense histories remain valid, sparse histories were already reachable, and the updated readers accept both while writers continue to emit strictly increasing IDs. No version-specific fallback is required by this diff. |
| `docs/testing/patterns.md` | Pass | Tests use real route/DB, filesystem, IndexedDB, and MSW-style boundaries appropriate to each surface and assert observable cursor/file/event results. |
| `docs/testing/anti-patterns.md` | Pass | No fake timers, direct global-fetch stubs, internal-module mocks, weakened assertions, or service-only regression test was added. |
| `docs/testing/api-testing.md` | Pass | The concurrent idempotent-send regression exercises the real API and transaction path. The existing row-lock fixture is a documented synchronization boundary used because the product API cannot pause at that lock. |
| `docs/testing/testing-external-behavior.md` | Pass | New API assertions observe route responses and listed events. Test-only sequence advancement explicitly models state written by the previous API rather than replacing the production code path under test. |
| `docs/testing/app-testing.md` | Pass | The changed Platform test exercises the existing snapshot-read signal integration with production-style context setup and real IndexedDB, which is appropriate for non-page-visible storage synchronization. |
| `docs/testing/cli-testing.md` | Pass | CLI coverage invokes `parseAsync`, uses a temporary output directory and HTTP handlers, verifies cursor requests/files, and checks the user-visible failure path for decreasing rows. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- Coverage is adequate for the changed invariant: concurrent idempotent reservations retain a gap; repeated callback batches preserve gaps; snapshot archives may begin above 1 and tail across sparse coverage; CLI local and remote histories accept gaps but reject non-increasing rows; Platform reads sparse snapshot and tail positions.
- CI is green at the reviewed head: 94 published checks completed with no failure or pending check.
- Codecov reports all modified coverable lines covered and no failed tests.

PR #27235, opened 2026-08-14
feat(billing): make the usage pack review a step in the plan flow

Lunarivibe:
## What

Changing an existing usage pack subscription opened `UsagePackSubscriptionChangeDialog` as a third modal, stacked on the plan flow dialog, which itself sits over the Settings dialog. The review restates the monthly ledger the configuration step already showed and only adds what the change costs today, so it now renders inside the same 860px frame as **step 3 of 3**.

- `PricingStepDialog` takes a step total. A checkout still ends at step 2 and hands off to Stripe; a managed change counts three, so the counter reads `Step 2 of 3` before the review and `Step 3 of 3` on it.
- The frame's back arrow discards the preview — that is what the review's `Cancel` button did — and the confirm keeps the configuration step's full-width place and shape at the foot of the dialog.
- The review lives in `PackageConfigurationStep` so it reuses that step's resolved member list and totals instead of resolving them a second time.
- Preview and confirm no longer share one error string: failing to open the review is the configuration step's error, failing to confirm is the review's.
- The legacy conversion review dialogs are untouched — they sit over the migration sub-page, not over another dialog.

## Verification

- `pnpm -F @okouai/app exec vitest run org-billing-tab.test.tsx org-members-tab.test.tsx` — 46 passed
- `npx oxlint` (package-local config) on both touched files, `pnpm knip --workspace apps/platform`, `prettier@3.8.1 --check`

Two tests moved with the UI: the review is reached by the same `dialog` accessible name, but Cancel is now the header back arrow, and a managed subscription's counter reads `Step 2 of 3`. Both cases gained assertions that the configuration step's order summary is gone while the review is open.

Behind `usagePackPlans`, which is off everywhere.


Lunarivibe:
LGTM


PR_AUTO_REVIEW_COMMIT: 148a6abea8dc04efc56912ce1ffcc403e64c9796
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Moves the usage-pack subscription-change review out of a third stacked modal and into the existing plan-flow frame as step 3, so a managed change now counts three steps while a checkout still ends at step 2 and hands off to Stripe. Also folds the current/new comparison behind a native `<details>` summary carrying the monthly-total delta, drops the orange from changed values in the New column, and pins each step's action to the frame's foot so it no longer scrolls away with the body.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic: no race conditions, unhandled null paths, or wrong assumptions | Pass | `usage-pack-pricing-page.tsx`: step is derived from one signal (`preview = management === null ? null : changePreview`) and consumed identically for the frame header and the body, so the counter and the rendered step cannot disagree. `onClose` clears the preview before the selected plan, and `confirmUsagePackSubscriptionChange$` already nulls the preview on success, which returns the flow to step 2. |
| `REVIEW.md` | Type safety: no `any`, no overly broad types | Pass | Introduces `PricingStep = 1 \| 2 \| 3` and `PricingStepTotal = 2 \| 3` and narrows `preview` from `… \| null` to a required `UsagePackSubscriptionChangePreviewResponse` on `PackageReviewStep`. No casts added. |
| `REVIEW.md` | Style and maintainability: no over-long functions, no duplicated logic | Pass | `PackageReviewStep` replaces the dialog at ~50 lines; the review reuses `PackageConfigurationStep`'s resolved member list and totals instead of resolving them again; the action-bar classes are one shared `STEP_ACTION_BAR` constant used by all three actions. |
| `REVIEW.md` | Security: no injection, hardcoded secrets, or missing auth on new surfaces | N/A | Presentation-only diff in `turbo/apps/platform`; no new endpoint, no request construction, no credential handling. |
| `REVIEW.md` | Feature regression guard: replaced behavior carries explicit regression coverage | Pass | The replaced screen is asserted at its new location — `org-billing-tab.test.tsx:1661` requires `Step 3 of 3` inside the flow dialog and `:1665` requires the configuration step's `Order summary` region to be gone while the review is open; `:1698` exercises the header back arrow that replaces the removed `Cancel`; `:2270` pins the managed counter at `Step 2 of 3`. |
| `REVIEW.md` | New user-facing behavior is gated behind a `FeatureSwitchKey` | Pass | Whole flow renders only under `FeatureSwitchKey.UsagePackPlans`, which is `enabled: false` with no org allowlist in `turbo/packages/core/src/feature-switch.ts`. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, contract, or `jsonb(...)` change; `usagePackSubscriptionChangePreviewResponseSchema` is consumed unchanged. |
| `docs/bad-smell.md` | Zero lint/type suppressions | Pass | No `eslint-disable`, `oxlint-disable`, `@ts-ignore`, `@ts-expect-error`, or `prettier-ignore` added. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | Only the static `lucide-react` import changes (`ChevronRight` added). |
| `docs/bad-smell.md` | No unnecessary try/catch; fail fast | Pass | No error handling added; the confirm path surfaces `confirmationLoadable.state === "hasError"` as a message and leaves the preview open. |
| `docs/bad-smell.md` | Comments explain WHY, not WHAT | Pass | The added comments state the reasons (why the review is a step, why the body drops its bottom inset, why the comparison folds, why numerals are ink). |
| `docs/fallback.md` | Default is no fallback; no branch guarding a state the contract already prevents | Pass | The one guard added, `{monthlyTotal && …}`, narrows `rows.at(-1)`, which TypeScript types as `T \| undefined`; it is type narrowing, not a runtime tolerance branch. |
| `docs/fallback.md` | A pre-GA feature-switched feature needs no compatibility code | Pass | No dual-read, tolerance branch, or migration added for the older shape of this switched-off feature; the old review dialog is deleted outright rather than kept beside the new step. |
| `docs/fallback.md` | Every fallback introduced by the diff is declared in the PR summary | N/A | The diff introduces no fallback, so no `Fallbacks` section is required. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | `expect(screen.queryByRole("region", { name: "Order summary" })).not.toBeInTheDocument()` asserts the live step-2/step-3 exclusivity of the current flow, not the absence of deleted code. |
| `docs/effect.md` | Choose the execution boundary: derived value in `computed`, user action in a command | Pass | Step number and back handler are derived in render from existing signals; the confirm stays a command (`confirmUsagePackSubscriptionChange$`) invoked from a DOM callback. No React effect added. |
| `docs/effect.md` | Keep React render pure | Pass | `PackageReviewStep` and `ManagedSubscriptionComparison` only read signals and compute values during render; the disclosure uses native `<details>`, so no state is written while rendering. |
| `docs/cache.md` | RC-5 no duplicate mutable sources of truth | Pass | `usagePackSubscriptionChangePreview$` becomes the single source for "is the review open"; the previous duplicate — `previewing = loading \|\| preview !== null` inside the order summary — is removed. |
| `docs/cache.md` | RC-4/RC-8 owner and lifetime match | Pass | The confirm loadable moves into `PackageReviewStep`, which is mounted exactly while the preview exists; on confirm error the preview stays open so the component keeps its error state. |
| `.claude/skills/ccstate/SKILL.md` | DOM callback pattern: `detach(..., Reason.DomCallback)` with `pageSignal$` | Pass | `detach(submitChange(), Reason.DomCallback)` with `pageSignal` read via `useGet(pageSignal$)`; matches the existing call site it replaces. |
| `.claude/skills/ccstate/SKILL.md` | Views use only the allowed ccstate hooks | Pass | Only `useGet`, `useSet`, `useLoadableSet` are used; the package-local `no-restricted-imports` rule for `src/views/**` still passes. |
| `docs/testing.md` | `feat:` commits carry integration tests | Pass | Coverage is added to the existing page-level suite rather than a new unit test; `org-billing-tab.test.tsx` drives the whole managed-change path through the rendered settings page. |
| `docs/testing.md` | Mock at boundary only — no `vi.mock()` of internal paths | Pass | No `vi.mock` added; the suite keeps using `context.mocks.api` handlers. |
| `docs/testing.md` | Use MSW for HTTP — no direct `fetch` stubbing | Pass | No `vi.stubGlobal("fetch", …)` added. |
| `docs/testing.md` | No fake timers | Pass | No `vi.useFakeTimers()` added; the new waits use `waitFor`/`findBy*`. |
| `docs/testing.md` | Test behavior, not mocks | Pass | New assertions read rendered text, roles, and the `details` `open` attribute; `confirmationRequests` is still checked as an observed side effect, not as a mock-call assertion. |
| `docs/testing/app-testing.md` | Platform tests are page-level and assert page-visible behavior | Pass | All new assertions run inside the existing `organization billing settings` page test through the rendered settings dialog; no component-only render added. |
| `docs/testing/anti-patterns.md` | AP-7 no testing implementation details | Pass | The disclosure is asserted through its native `open` attribute and visible summary text, which is the user-visible state of a `<details>`, not an internal flag. |
| `docs/testing/anti-patterns.md` | AP-10 no direct component rendering | Pass | No direct `render(<Component />)`; everything goes through the page harness. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Nits (non-blocking)

- `usage-pack-pricing-page.tsx` `PackageReviewStep`: the confirm error uses `text-xs` while the sibling bar in `ManagedSubscriptionOrderSummary` and `OrderSummary` uses `text-sm`. The rest of this file was moved to `sm` for small copy, so this is a leftover size.
- `usage-pack-pricing-page.tsx` `ManagedSubscriptionComparison`: the collapsed summary joins the two totals with a bare `→` glyph. It reads correctly visually, but the pair is announced without a relation word; an `aria-label` on that span would state "from X to Y" explicitly.
- Test coverage pins `Step 2 of 3` and `Step 3 of 3`, but nothing asserts that the checkout path still reads `Step 1 of 2` / `Step 2 of 2`. That total is the other half of the new `total` prop and is currently only verified manually.

### Testing

- Coverage: Adequate. The `feat:` change lands its behavior in the existing page-level integration suite — step 3 rendering inside the flow dialog, the configuration step being unmounted while the review is open, the header back arrow replacing `Cancel`, the return to `Configure member packages` after stepping back, the managed step counter, and the comparison disclosure's default-closed state, summary delta, and expansion. The sticky action bar is CSS placement and is not asserted in jsdom.
- Conventions: Compliant. Integration-only, no internal mocks, no fake timers, no direct fetch stubbing, assertions on rendered behavior. `pnpm -F @okouai/app check-types`, package-local `oxlint`, `knip --workspace apps/platform`, `prettier@3.8.1 --check`, and `vitest run org-billing-tab.test.tsx org-members-tab.test.tsx` (46 passed) were run on this head.

Issue #27194, opened 2026-08-14
refactor(chat): remove schema negotiation rollout fallbacks


e7h4n:
Follow-up to #26848 after the Chat Event schema negotiation rollout.

Removal gate:

- wait until the #26848 API rollback window is closed
- wait at least 2 days after the replacement web/app build is live so stale clients have drained
- wait at least 2 hours after rollout so existing runner/sandbox CLI contexts have drained
- verify production no longer receives Chat Event reads that omit the schema header or explicitly request V4

Cleanup:

- require `X-Chat-Event-Schema-Version` and remove the fixed missing-header V5 default
- retire the V5-to-V4 row and Snapshot downgrade path
- require paired `sinceEventId` and `sinceSeqId` cursors after the thread start
- require the echoed schema-version response header and Snapshot `lastEventId` in app/CLI readers
- remove the stored requested-version Snapshot pointer fallback
- delete the compatibility tests together with the fallback code and contract optionality

The DB/API migration contraction remains tracked separately by #27174.

PR #27270, opened 2026-08-14
feat(platform): add shared worker chat database facade

### LINKED ISSUE #27220: feat: add a SharedWorker-backed chat event database facade
# Summary

Move the Platform read paths for canonical `ChatEvent` rows and `ChatThreadEvent` state behind a revision-scoped `SharedWorker`. The worker will own the relevant Ably user-channel notifications, remote Snapshot/incremental synchronization, and IndexedDB access. Tabs will interact with it through three small ccstate commands: `heartbeat`, `query`, and `on`.

This is a Platform-only change. It must not require API contract or API implementation changes.

## Motivation

Today every tab independently:

- creates realtime subscriptions;
- reacts to the same `threadListChanged` and `chatThreadMessageCreated:*` notifications;
- fetches the same ChatThread Snapshot/events or ChatEvent Snapshot/rows;
- writes the same logical data into the same user/org IndexedDB;
- maintains tab-local synchronization state.

This multiplies Ably connections, HTTP requests, IndexedDB work, and lifecycle complexity as more tabs are opened. Recent hidden-tab connection shutdown behavior exists mainly to limit this resource use.

A SharedWorker is a better owner for this work because every tab on the same origin and frontend revision can connect to the same worker instance. The worker can serialize synchronization per dataset and expose a database-like read facade to tabs.

## Scope

This first version covers only:

1. canonical ChatEvent rows;
2. ChatThreadEvent Snapshot plus event state.

The following are explicitly out of scope:

- moving Indicators into the shared database facade;
- background prewarming of unread threads inside the SharedWorker;
- migrating unrelated realtime/affiliate subscriptions;
- adding or changing API endpoints or realtime payloads;
- adding a new dedicated browser smoke test;
- adding an in-memory persistence fallback after IndexedDB write failure.

Unread-thread prewarming remains a tab/business-layer concern for this version: on startup, the tab reads Indicators once and issues worker queries for the selected unread thread IDs.

## Worker identity and revision behavior

Create the worker with Vite's standard static worker URL form:

```ts
new SharedWorker(new URL("./shared-database-worker.ts", import.meta.url), {
  type: "module",
});
```

Do not introduce a stable explicit worker name or a stable unrevisioned URL.

Vite will emit an immutable, content-hashed worker asset. The SharedWorker identity is therefore naturally revision-scoped:

- tabs running the same frontend revision use the same SharedWorker;
- old and new frontend revisions may temporarily run separate workers;
- after tabs refresh, they naturally converge on the new worker revision;
- an old worker remains alive only while old-revision tabs still own it.

The public database API does not need to expose a worker epoch. A worker process may still be terminated and recreated at the same immutable URL, but the MessagePort connection/heartbeat layer handles reconnection, and tabs recover by registering and querying again.

### SharedWorker

The SharedWorker has its own `main` entry point and its own independent ccstate store:

```ts
const store = createStore();
```

Its bootstrap owns:

- connected MessagePorts;
- port registrations and heartbeat timestamps;
- the latest tab-supplied token per user/org identity;
- one Ably session and user channel per user;
- ChatEvent dataset actors keyed by user/org/thread;
- ChatThreadEvent dataset actors keyed by user/org;
- all IndexedDB connections, schema upgrades, reads, and writes;
- remote Snapshot and incremental synchronization;
- query coalescing and notification fan-out.

Tabs must not access the ChatEvent or ChatThreadEvent IndexedDB stores directly after this feature is enabled.

### Platform

The Platform wraps the DOM SharedWorker/MessagePort API with ccstate commands. Outside the ccstate layer, the interface should remain small and contain only three operations:

1. `heartbeat`
2. `query`
3. `on`

The production bridge implements these operations over MessagePort. A test-only bridge invokes the same worker ccstate commands directly.

## Typed dataset registry

Define a shared, statically typed dataset registry imported by both the tab bundle and worker bundle. The initial registry contains exactly two dataset definitions.

Conceptually:

```ts
type SharedDatabaseDataKey =
  | {
      kind: "chat-event";
      userId: string;
      orgId: string;
      threadId: string;
    }
  | {
      kind: "chat-thread-event";
      userId: string;
      orgId: string;
    };
```

Do not introduce parallel handwritten business types. Reuse the existing API-contract Zod schemas and infer TypeScript output types from them:

- `chatEventRowV4Schema` / `ChatEventRowV4`;
- `chatThreadEventSchema` / `ChatThreadEvent`;
- `chatThreadSnapshotProjectionSchema` / `ChatThreadSnapshotProjection`;
- the existing ChatThread Snapshot response fields, including `latestEventId` and `latestSeqId`.

Dataset definitions may compose existing schemas into dataset-specific full-query results, but they must not redeclare contract fields.

Zod schemas are never transferred through MessagePort. RPC messages carry only a dataset/data key and serializable arguments. Both sides resolve the key through the same registry. Every `MessageEvent.data` and every RPC result is treated as `unknown` at the transport boundary and parsed immediately with the relevant Zod schema.

The query return type must be inferred from the selected dataset; do not use a universal `Record` result type.

### 1. Heartbeat

Conceptual signature:

```ts
heartbeatSharedDatabase$(
  {
    userId,
    orgId,
    token,
  },
  signal,
);
```

Requirements:

- The first heartbeat registers the tab/bridge client with its current identity and token.
- Subsequent heartbeats are idempotent updates of the existing registration; they must not add another abort listener or another logical port registration.
- The tab supplies the token proactively. The worker never calls back into a tab to request or refresh a token.
- The root/tab `AbortSignal` owns the registration. When aborted, the bridge unregisters the client and removes all heartbeat, query, and `on` registrations owned by it.
- Bootstrap sends an immediate heartbeat, then uses `setLoop` to refresh it periodically.
- The refresh interval should be half of the observed Clerk token lifetime, capped at one minute. A visibility/focus recovery should send an immediate heartbeat so timer throttling in hidden tabs does not delay recovery.
- A hidden tab is not considered disconnected merely because it is hidden.

The MessagePort itself supplies the production client identity. The test bridge may use an internal generated client/port ID without exposing it in the public API.

### Token and 401 behavior

The worker keeps only the latest token in memory; tokens are never written to IndexedDB.

On HTTP 401:

- mark the corresponding user/org credential state as auth-blocked;
- stop issuing remote work for that identity;
- do not request a replacement token from a tab;
- do not resume when the identical rejected token is received again;
- keep cache-only queries available;
- remember that affected open datasets are dirty;
- when a new token value arrives through heartbeat, clear auth-blocked and catch up dirty open datasets.

Auth blocking should be scoped to the relevant credential identity rather than unnecessarily disabling unrelated user/org registrations in the same worker process.

### 2. Query

Conceptual signature:

```ts
querySharedDatabase$(
  {
    dataKey,
    afterSeqId,
    consistency,
  },
  signal,
);
```

Where:

```ts
type Consistency = "cache-only" | "catch-up";
```

Semantics:

- `afterSeqId: null` requests the full logical dataset.
- A numeric `afterSeqId` requests data strictly after that sequence ID when the dataset can provide an append-only result.
- `cache-only` must never initiate an HTTP request. It reads IndexedDB and returns as soon as local validation completes.
- `catch-up` runs or joins the dataset's current remote synchronization before returning.
- Concurrent catch-up queries for the same data key must share an in-flight synchronization rather than multiply HTTP requests by tab.
- Query accepts an AbortSignal. Aborting cancels the caller's wait; it must not incorrectly cancel shared work still required by another caller.
- The tab validates the returned value using the selected dataset's existing Zod schemas.

There is no separate persisted worker cursor. Synchronization derives its cursor from canonical cached data:

- ChatEvent uses the greatest cached row `seqId` for the thread;
- ChatThreadEvent uses the Snapshot `latestSeqId` and the greatest cached post-Snapshot event `seqId`.

If an IndexedDB write does not advance the cache, a later synchronization may fetch duplicate rows. Writes are idempotent by canonical key/sequence, so correctness must not depend on a separate cursor store.

### 3. On database append/invalidation

Conceptual signature:

```ts
onSharedDatabase$(dataKey, callback, signal);
```

This is the only public database event.

Requirements:

- The callback carries no records and no generic payload.
- It means only: new canonical data may be available; call `query` with the tab's own current cursor.
- Each tab owns its own last rendered sequence ID and decides what to query.
- Multiple pending notifications may be coalesced per data key.
- The callback is stored locally in the tab bridge; functions are never transferred through MessagePort.
- Registration is owned by its AbortSignal and is automatically removed on abort.
- The production bridge installs its local callback mapping before completing the underlying subscribe operation, buffering/coalescing notifications during registration so there is no subscribe/query race.

There is no public `onReset`, `onChange`, or reset-required error/event.

### Initial page load

The existing local-first rendering behavior is preserved:

1. register heartbeat and the `on` listener;
2. run a full `cache-only` query;
3. render the cached result immediately;
4. derive the tab cursor from the typed result;
5. run a `catch-up` query;
6. render newly returned canonical data;
7. on later notifications, query again from the tab's current cursor.

Tab-local optimistic ChatEvent and ChatThreadEvent overlays remain in the tab. The SharedWorker serves only server-confirmed canonical data. Existing optimistic reconciliation runs when canonical query results are applied.

### ChatEvent

The worker owns the current Snapshot-plus-tail algorithm:

- read the highest cached `seqId`;
- cold-start an empty cache through the existing event Snapshot download;
- fetch incremental event rows through the existing rows endpoint;
- validate canonical rows with `chatEventRowV4Schema`;
- write rows transactionally into IndexedDB;
- notify interested tabs after the write attempt;
- handle repeated Ably notifications by serializing/coalescing synchronization per thread.

If an incremental cursor has expired (HTTP 410), the worker transparently:

1. downloads a current full Snapshot;
2. rebuilds the thread cache;
3. tails rows after the Snapshot sequence;
4. returns rows after the caller's requested `afterSeqId` when possible.

The tab/business layer does not observe a reset event.

### ChatThreadEvent

The worker owns the current organization-scoped Snapshot-plus-event algorithm:

- read the stored ChatThread Snapshot and post-Snapshot event log;
- fetch the existing organization-scoped Snapshot when no valid baseline exists;
- tail the existing ChatThread events endpoint;
- validate Snapshot projections and events with existing contract schemas;
- transactionally replace or append to IndexedDB;
- replay Snapshot plus events into the current typed selection;
- notify interested tabs after the write attempt.

A ChatThread Snapshot is a compacted current projection, not a complete historical event archive. Therefore, after cursor expiry the worker cannot manufacture every original event between an arbitrarily old tab cursor and the new Snapshot sequence.

This remains transparent to the business layer: the ChatThread dataset adapter internally replaces its Snapshot baseline, tails post-Snapshot events, and returns the complete current typed ChatThread selection. The Platform canonical-data adapter replaces/replays that selection before resolving the query. No public reset event or reset-required error is introduced.

## Realtime behavior

The worker creates an Ably realtime session for the user and subscribes to the whole `user:<userId>` channel, but this first version reacts only to notifications needed by the two registered datasets:

- `threadListChanged` invalidates/catches up open ChatThreadEvent data keys for that user;
- `chatThreadMessageCreated:<threadId>` invalidates/catches up matching open ChatEvent data keys.

Because unread-thread prewarming stays in the tab, the worker does not need to infer the organization of a closed/unregistered thread and no realtime payload/API change is required.

Other realtime subscriptions remain outside this migration's scope.

When the worker or Ably connection is unavailable:

- connected tabs should expose the existing disconnected/reconnecting status semantics;
- tabs recreate/connect to the same revision-scoped SharedWorker through the bridge;
- after reconnection, heartbeat registration and catch-up queries restore state;
- hiding a tab must not deliberately close the SharedWorker-owned realtime connection or mark the tab disconnected.

## IndexedDB ownership and degradation

The SharedWorker is the only interface to the affected IndexedDB stores. IndexedDB is a local cache behind the worker, not the tab-to-tab notification channel and not the server source of truth.

For every remote synchronization:

1. fetch and validate remote canonical data;
2. attempt the IndexedDB transaction;
3. emit the data-key notification whether the write succeeds or fails.

Initial failure behavior is intentionally simple:

- do not add a durable memory cache;
- mark the affected dataset/DB identity as degraded in worker memory;
- `cache-only` remains strictly cache-only and may return a cache miss/older state;
- `catch-up` bypasses or repairs the failed cache path and returns validated remote results directly;
- a later successful cache operation may clear degraded state;
- worker restart loses degraded state, but the mandatory startup catch-up restores correctness.

Notifications never promise that IndexedDB committed; they only tell the tab to query.

## IndexedDB schema upgrades across revisions

Old and new revision-scoped workers can temporarily coexist while sharing the stable user/org IndexedDB name.

On `versionchange`, an old worker must:

1. close the affected database handle immediately;
2. stop issuing operations that assume the old schema;
3. send a control message instructing its old-revision tabs to refresh.

The new worker can then complete the schema upgrade. This preserves the current Platform behavior of reloading on Chat IndexedDB version changes, while centralizing the handle in one worker per frontend revision instead of every tab.

`reload-required`, heartbeat status, request results, and errors are transport/control messages; they are not additional public database events.

## MessagePort bridge

The production transport should use a small Zod-validated RPC protocol with serializable messages such as:

- heartbeat/register/update;
- query request/result/error;
- subscribe/unsubscribe;
- append/invalidation notification;
- cancel request;
- reload required.

Each request has a request ID. Each logical `on` registration has a subscription ID. The tab bridge owns callbacks and AbortSignals; the worker owns port registrations and dataset subscription reference counts.

The production Platform bundle must import only the public dataset registry and MessagePort client. Do not import worker command implementations into the production tab bundle merely to support tests. The test-only bridge may import both sides.

### Browser

- Platform bootstrap creates/lazily connects the revisioned SharedWorker bridge.
- Authenticated bootstrap sends the first heartbeat and starts the heartbeat loop.
- The SharedWorker entry point creates a new ccstate store.
- Its `onconnect` handler registers each MessagePort with worker ccstate commands.
- Worker root lifecycle owns Ably, DB handles, port state, and in-flight work.

### Vitest

`setupPage` should bootstrap the SharedWorker test runtime in addition to the existing Platform bootstrap when this feature path is enabled:

```ts
bootstrapSharedWorkerForTest$;
```

The test bootstrap:

- runs the worker ccstate signal graph in the existing `TestContext.store`;
- installs a test-only direct bridge that dispatches the same worker commands;
- binds cleanup to `context.signal`;
- returns no special runtime object;
- permits existing Platform tests to call the public heartbeat/query/on commands without a DOM SharedWorker implementation.

The direct bridge is for regular Platform page tests. A focused protocol test must still use isolated stores and a fake MessagePort pair so same-store tests do not hide production boundary mistakes.

## Testing strategy

Do not add a new dedicated browser smoke test in this issue. Existing Playwright coverage remains the repository's real-browser layer.

### 1. Worker command tests

Use a dedicated ccstate store, `fake-indexeddb`, the real `idb` implementation where needed, existing MSW API mocks, and the existing Ably mock.

Cover at least:

- full cache-only ChatEvent query makes no network request;
- full cache-only ChatThreadEvent query makes no network request;
- initial Snapshot plus tail synchronization;
- incremental query strictly after a tab cursor;
- concurrent callers join one catch-up;
- Ably notification coalescing;
- ChatEvent 410 transparently rebuilds Snapshot/cache and returns usable results;
- ChatThreadEvent 410 transparently replaces the Snapshot baseline and returns current state;
- successful IDB writes notify subscribers;
- failed IDB writes still notify subscribers;
- degraded catch-up returns remote data without relying on the failed cache;
- 401 blocks remote work;
- an identical rejected token does not resume work;
- a new heartbeat token resumes and catches up dirty datasets;
- user/org/thread isolation;
- malformed cached/remote data is rejected by the existing Zod schemas;
- IndexedDB versionchange closes the DB and requests tab reload.

### 2. Platform tests with the direct bridge

Use the existing `TestContext.store` and `bootstrapSharedWorkerForTest$`.

Cover at least:

- bootstrap registers identity/token;
- initial cache-only render followed by catch-up render;
- `on` causes the tab to query from its own cursor;
- optimistic events reconcile against canonical query results;
- AbortSignal removes all registrations;
- org switch/identity mismatch cannot apply stale results;
- hidden tab state no longer closes the SharedWorker-backed path;
- Indicators-driven unread prewarming issues worker queries from the business layer.

Existing ChatEvent Snapshot/background-sync and ChatThreadEvent persistence tests should be migrated or reused rather than duplicated.

### 3. MessagePort protocol tests

Create two independent ccstate stores and an in-memory fake MessagePort pair implementing the minimal production surface:

```ts
interface PortLike {
  postMessage(value: unknown): void;
  start(): void;
  close(): void;
  addEventListener(
    type: "message",
    listener: (event: MessageEvent<unknown>) => void,
  ): void;
  removeEventListener(
    type: "message",
    listener: (event: MessageEvent<unknown>) => void,
  ): void;
}
```

The fake transport must run `structuredClone` on every message so tests cannot accidentally transfer callbacks, Zod schemas, AbortSignals, or other non-cloneable state.

Cover request correlation, input/output Zod validation, cancellation, subscription cleanup, callback delivery, reload-required control messages, and independent store state.

### Heartbeat tests and `setLoop`

The current Platform `setLoop` deliberately ignores wall-clock intervals in Vitest and fails after more than 100 iterations. A permanent one-minute heartbeat loop must therefore not run uncontrolled in every page test.

- Production bootstrap uses `setLoop` normally.
- Test bootstrap sends one initial heartbeat.
- Tests that need renewal explicitly drive a test-only heartbeat tick.
- Test the recurring bootstrap/scheduler separately with a controlled clock or injected ticker.

## Feature-switch rollout

Gate the SharedWorker path behind a new Platform feature switch.

- Disabled: preserve the current per-tab ChatEvent/ChatThreadEvent path.
- Enabled: route those datasets exclusively through the SharedWorker facade.
- Select transport once per page lifecycle; do not switch an already-running page back and forth when the cached/remote feature-switch value changes.
- Remove the hidden-tab shutdown hack from the SharedWorker-owned path.
- Preserve the legacy path only for rollout/rollback until the feature switch is terminal.

## Acceptance criteria

- Vite emits a content-hashed module SharedWorker without a custom stable name.
- Same-revision tabs use one SharedWorker instance and one relevant user-channel realtime owner.
- Platform exposes only typed ccstate `heartbeat`, `query`, and `on` operations for this facade.
- Public database notifications carry no records and there is no public reset event.
- ChatEvent and ChatThreadEvent query results are inferred and runtime-validated from existing API-contract schemas.
- Tabs no longer directly read or write the affected IndexedDB stores when the feature is enabled.
- Initial rendering remains cache-first and never blocks on the network.
- Catch-up is deduplicated across tabs and handles Snapshot/incremental recovery internally.
- ChatEvent cursor expiry and ChatThread Snapshot replacement are transparent to the UI/business layer.
- IDB write failure still notifies tabs; catch-up can return remote data without a memory cache.
- 401 stops remote work until a new heartbeat token arrives.
- AbortSignals clean up queries and subscriptions.
- Old workers close IndexedDB and ask old tabs to refresh on versionchange.
- Unread-thread prewarming remains in the tab via Indicators plus worker queries.
- No API change is required.
- Worker command, direct-bridge Platform, and fake-MessagePort protocol tests cover the design without adding a new browser smoke test.

e7h4n:
## Summary

- add a content-hashed module SharedWorker, driven by an independent ccstate store, that owns Ably synchronization and IndexedDB access for ChatEvent and ChatThreadEvent only
- expose typed heartbeat, query, and append-invalidation APIs through production MessagePort and direct Vitest bridges
- preserve cache-only first paint and catch-up queries, transparent cursor reset recovery, IndexedDB fallback, token refresh and 401 recovery, versionchange reload behavior, and hidden-tab connectivity behind SharedChatDatabase
- use the current canonical ChatEvent schema/version header and atomically maintain its existing IndexedDB cursor with cache rebuilds
- keep Indicators, unread-thread prewarming, and unrelated realtime subscriptions in the existing tab-owned paths

## Fallbacks

- `turbo/apps/platform/src/shared-database/worker-runtime.ts` — `ChatEventActor.degraded`, `ChatThreadEventActor.degraded`, and their cache read/write recovery paths protect the SharedWorker data facade from browser IndexedDB read or write failures: `cache-only` remains network-free, while `catch-up` rebuilds from the canonical server and returns remote data even when persistence still fails. Surface/window: none; `SharedChatDatabase` is non-GA and this is browser-storage resilience rather than cross-version rollout compatibility. Lifecycle/removal: keep while an IndexedDB-unavailable browser must still render canonical chat data; remove only if working IndexedDB becomes a hard product prerequisite. Follow-up: none planned because the behavior is intentionally persistent, not time-boxed rollout compatibility.
- `turbo/apps/platform/src/shared-database/chat-event-schema-version.ts` and `worker-runtime.ts` — the SharedWorker accepts a missing ChatEvent response-version header and derives a missing Snapshot `lastEventId` from the immutable NDJSON rows when a newly deployed app reaches the previous API during rollout or rollback. Surface/window: new web/app SharedWorker versus previous API; observed maximum exposure is 102 minutes. Lifecycle/removal: require both fields and delete these tolerances through #27194 after that API skew window has closed.

## Verification

- pnpm format
- NODE_OPTIONS=--max-old-space-size=3072 pnpm turbo run lint --concurrency=1 --log-order grouped
- pnpm check-types
- worker-runtime.test.ts (16 tests)
- message-port.test.ts (5 tests)
- reconnecting-client.test.ts (1 test)
- platform-direct-bridge.test.ts (4 tests)
- chat-event-snapshot-read.test.ts (9 tests)
- chat-thread-event-persistence.test.ts (4 tests)
- Platform production Vite build; emitted a revisioned shared-database-worker asset
- pnpm install --frozen-lockfile --ignore-scripts

Closes #27220


e7h4n:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 1440c2e09ed81a12989a9472482b25e16bd193d2

### Summary

This PR adds the feature-gated, content-hashed SharedWorker facade for ChatEvent and ChatThreadEvent synchronization, with typed MessagePort/direct-test bridges, SharedWorker-owned IndexedDB access, Ably catch-up, and tab-side invalidation-driven reads. The architecture and happy-path coverage are substantial, but the IDB degradation path can self-trigger an unbounded synchronization loop, and stale hidden-tab clients cannot recover once the worker prunes their registration.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness, races, and failure-path behavior | Finding (P0) | `worker-runtime.ts:445-446, 482-483, 640-643, 814` can rebroadcast every degraded catch-up, feeding the tab invalidation daemon indefinitely. |
| `REVIEW.md` | Resource lifecycle and recoverability | Finding (P1) | `worker-runtime.ts:239,1231-1240` removes stale clients while `reconnecting-client.ts:119-133` only replaces a generation on transport timeout. |
| `REVIEW.md` | Security and type safety | Pass | Wire messages and dataset results are strict Zod-validated; data keys carry user/org identity and the runtime checks identity before and after queries. |
| `REVIEW.md` | Feature containment and regression coverage | Pass | `SharedChatDatabase` is off by default/staff-only, and legacy ChatEvent/ChatThreadEvent paths remain selected when disabled. |
| `docs/docs.md` | Apply all practices routed by the changed surfaces | Pass | Platform, ccstate, event-sourcing, persisted-browser-boundary, fallback, React-work, and test practices were all reviewed. |
| `docs/bad-smell.md` | Fail fast; avoid `any`, suppressions, dynamic production imports, and hard-coded deployment URLs | Pass | No new `any` or suppression was added; API base resolution stays centralized and production imports are static. |
| `docs/effect.md` | Every long-lived process has the correct owner and a working abort/recovery path | Finding (P1) | A live MessagePort can outlast its pruned worker registration, leaving the tab's owner alive but permanently unusable. |
| `docs/cache.md` | Cache/resource identity, invalidation, and teardown must match their real owner | Finding (P1) | Stale-client pruning tears down only the runtime registration, not the corresponding transport generation. |
| `.claude/skills/ccstate/SKILL.md` | Commands propagate owned AbortSignals and long-running work remains cancellable | Pass | Page/root signals flow through bridge calls, subscriptions, worker waits, HTTP requests, and daemon loops. |
| `docs/event-sourcing.md` | Persistent events remain authoritative and optimistic events reconcile by event identity | Pass | Shared results are schema-validated canonical rows/events and merge through the existing ID-based reconciliation commands. |
| `docs/fallback.md` | Every introduced fallback is explicitly declared in the PR summary | Finding (P1) | The new IDB-degraded server-query path is mentioned inline but has no required `## Fallbacks` entry with symbol, surface, and lifecycle. |
| `docs/deployment-compatibility.md` | Persisted/browser version boundaries remain safe across old and new clients | Pass | The worker uses the existing per-user/org DB and current schema version; existing `versionchange` handling closes the DB and asks connected tabs to reload. |
| `docs/react-commit.md` | Avoid no-op writes and broad/repeating invalidations when no new remote data arrived | Finding (P0) | A failed IDB write leaves `degraded=true`, yet a subsequent identical snapshot catch-up still reports `changed` and emits another invalidation. |
| `docs/testing.md` | New feature behavior has integration coverage and tests avoid prohibited timing controls | Finding (P1) | Bridge/runtime integration exists, but `reconnecting-client.test.ts:84,122` uses fake timers. |
| `docs/testing/patterns.md` | Use real boundaries: MSW HTTP, real IndexedDB, and shared test context | Pass | Worker tests use contract-aware `context.mocks`, `idb-real`, structured cloning, and independent ccstate stores for MessagePort coverage. |
| `docs/testing/anti-patterns.md` | Do not use `vi.useFakeTimers()` or `vi.advanceTimersByTime*()` | Finding (P1) | `reconnecting-client.test.ts:84-142` explicitly installs and advances fake timers. |
| `docs/testing/app-testing.md` | Platform integration enters through production bootstrap and context-owned mocks | Pass | `platform-direct-bridge.test.ts` uses `setupPage`, feature-switch bootstrap, context mocks, and context-owned cleanup. |
| `docs/testing/testing-external-behavior.md` | Verify platform outcomes through the closest production-visible boundary | Pass | Platform integration asserts resulting chat/thread state and realtime indicator behavior; lower-level tests are limited to the otherwise unavailable Worker/MessagePort boundary. |

### Findings

- `turbo/apps/platform/src/shared-database/worker-runtime.ts:445-446,482-483,640-643,814`: the IDB failure fallback creates a self-sustaining HTTP/invalidation loop. When a subscribed actor's IDB write fails, the actor remains degraded but the sync result reports `changed: shouldWrite`; `complete*Sync` broadcasts `append`. The tab invalidation daemon then sees no newer cache record and issues another `catch-up`; degraded mode rebuilds from the server, the write fails again, and another `append` is emitted even though no new Ably message or record arrived. This affects both datasets and can issue unbounded snapshot/event requests while IDB is unavailable. Please make notification reflect an actual new remote invalidation/advance rather than every degraded query, and add a full bridge + invalidation-daemon regression test proving a failed write returns remote data without recursively scheduling more catch-ups. This violates the correctness gate in `REVIEW.md` and the no-op invalidation rule in `docs/react-commit.md`.

- `turbo/apps/platform/src/shared-database/worker-runtime.ts:239,1231-1240` and `turbo/apps/platform/src/shared-database/reconnecting-client.ts:119-133`: a hidden/suspended tab cannot recover after another active tab causes its client registration to be pruned. The worker removes the stale runtime client but leaves its MessagePort alive. When the tab resumes, heartbeat receives `Shared database client is not connected` immediately; the reconnecting bridge only resets on `SharedDatabaseTransportTimeoutError`, so it keeps retrying the same orphaned generation and never restores its subscriptions. Please make pruning and transport ownership converge, or reconnect on this terminal heartbeat error, with a two-client regression test that suspends one client's heartbeat past the stale threshold and proves it rejoins. This violates the owner/recovery requirements in `docs/effect.md` and `docs/cache.md`.
- `turbo/apps/platform/src/shared-database/__tests__/reconnecting-client.test.ts:84-142`: this test uses `vi.useFakeTimers()` and `vi.advanceTimersByTimeAsync()`, which are explicitly prohibited by `REVIEW.md`, `docs/testing.md`, and `docs/testing/anti-patterns.md`. Please drive the configurable short timeout with real time or a deterministic boundary/deferred mechanism while keeping browser timers real.
- PR summary / `turbo/apps/platform/src/shared-database/worker-runtime.ts`: the introduced IDB-degraded server-query fallback is only named in a Summary bullet. `docs/fallback.md` requires a dedicated `## Fallbacks` entry naming the file/symbol, protected interaction, surface/window, removal condition, and follow-up; an undeclared fallback is independently P1. Please add the formal declaration (surface: none, non-GA `SharedChatDatabase` switch) and state its intended lifecycle.

### Fallbacks

- `turbo/apps/platform/src/shared-database/worker-runtime.ts` — `ChatEventActor.degraded`, `ChatThreadEventActor.degraded`, and the cache read/write recovery paths: when IndexedDB cannot be read or written, `cache-only` stays network-free while `catch-up` rebuilds from the canonical server and returns remote data even if persistence still fails. Surface: none, non-GA `SharedChatDatabase` feature switch. Removal condition/follow-up: not supplied. Declared in PR summary: No — mentioned inline, but missing the required `## Fallbacks` declaration (P1). Verdict: behavior is required by the stated storage-failure design, but its declaration is incomplete and its current notification implementation is unsafe (P0 above).

### Testing

- Coverage: Insufficient for the two failure modes above. The PR has strong happy-path runtime, MessagePort, direct-bridge, cursor-reset, 401, versionchange, two-dataset, and cancellation coverage, but no end-to-end degraded-IDB invalidation test and no stale two-client recovery test.
- Conventions: Violation — `reconnecting-client.test.ts` uses prohibited fake timers. Boundary mocking otherwise follows the platform test practices.
- Current CI for `1440c2e09ed81a12989a9472482b25e16bd193d2` is green; Codecov reports 85.26% patch coverage. Prior real-browser preview verification also covered two tabs sharing the revisioned worker, hidden-tab appends, cache reuse, and worker recreation, but those scenarios do not exercise the two failure paths above.

e7h4n:
Changes Requested


PR_AUTO_REVIEW_COMMIT: a6749c289999fd0cd4fc87f85de90c6a9624d9a1

### Summary

This revision fixes the previously reported degraded-IDB invalidation loop, stale-pruned-tab recovery, prohibited fake timers, and missing IDB fallback declaration. The content-hashed worker entry, typed MessagePort/direct bridges, identity isolation, cancellation ownership, cache-only/catch-up split, 401 recovery, and versionchange reload path are otherwise coherent. The remaining blockers are in the realtime handoff and retained ChatThread event-log behavior: the first subscription is acknowledged before Ably attachment is known to be usable, a one-off remote catch-up failure drops the only invalidation, and the SharedWorker path bypasses the existing event-log snapshot-rebase bound. The new worker also copies a temporary API compatibility fallback without declaring it and omits the existing response-version assertion.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness, races, and failure-path behavior | Finding (P1) | `worker-runtime.ts:343-352,1108-1130` acknowledges dataset subscription without using `session.ready`, and `worker-runtime.ts:468-486,505-523,1166-1204` drops a failed one-shot realtime catch-up. |
| `REVIEW.md` | Security and type safety | Pass | Wire inputs/results are strict Zod-validated, data keys carry user/org identity, and the runtime checks identity both before and after a query. |
| `REVIEW.md` | Feature containment and regression coverage | Pass | `SharedChatDatabase` is off by default/staff-only, and the legacy ChatEvent/ChatThreadEvent paths remain selected when disabled. |
| `docs/docs.md` | Apply all practices routed by the changed surfaces | Pass | Platform, ccstate, event sourcing, cache, deployment boundary, fallback, React work, and testing practices were reviewed. |
| `docs/bad-smell.md` | Fail fast; avoid `any`, suppressions, dynamic production imports, and hard-coded deployment URLs | Pass | No new `any` or suppression is present; production imports are static and API origin resolution remains centralized. |
| `docs/effect.md` | Long-lived async work has an owner and required readiness is observed | Finding (P1) | The Ably attach promise is stored as `SharedDatabaseRealtimeSession.ready` but no production caller observes it before the public `on` registration completes. |
| `docs/cache.md` | Cache size, invalidation, and teardown match the resource owner | Finding (P1) | The legacy 100-event snapshot-rebase guard remains only in the disabled path; the SharedWorker returns the complete retained log on every query. |
| `.claude/skills/ccstate/SKILL.md` | Commands propagate owned AbortSignals and daemon work remains cancellable | Pass | Page/root signals flow through bridge requests, subscriptions, worker waits, HTTP, IDB operations, and daemon loops. |
| `docs/event-sourcing.md` | Canonical events stay authoritative and compacted Snapshot baselines are respected | Finding (P1) | Canonical replay is correct, but the enabled path no longer performs the existing bounded Snapshot rebase for a long post-Snapshot event log. |
| `docs/fallback.md` | Every fallback introduced by the diff is justified and declared | Finding (P1) | `worker-runtime.ts:732` adds the temporary missing-`lastEventId` derivation to the new worker, but the PR `## Fallbacks` section declares only IndexedDB degradation. |
| `docs/deployment-compatibility.md` | Cross-version response shapes are bounded and protocol versions fail fast | Finding (P1) | The worker sends the current ChatEvent schema header but does not assert the response header as the existing tab source does, while also retaining the old-API `lastEventId` tolerance. |
| `docs/react-commit.md` | Avoid no-op writes and broad/repeating invalidations | Pass | Actor sequence high-water marks now suppress identical degraded retries, and the full bridge regression test proves no recursive invalidation. |
| `docs/testing.md` | New behavior covers races and failure paths without prohibited timing controls | Finding (P1) | Current tests cover the earlier fixes, but not delayed/failed initial Ably attachment, a one-shot transient catch-up failure, or the >100-event rebase boundary. |
| `docs/testing/patterns.md` | Use real integration boundaries and context-owned mocks | Pass | Tests use contract-aware context mocks, real `idb` over fake IndexedDB, structured-cloned ports, and independent stores for protocol coverage. |
| `docs/testing/anti-patterns.md` | Do not use fake timers or internal relative mocks | Pass | The reconnect test now uses a configurable short real timeout, and the only module mock replaces the external `idb` package with its real test implementation. |
| `docs/testing/app-testing.md` | Platform integration enters through production bootstrap | Pass | Direct-bridge coverage uses `setupPage`, feature-switch bootstrap, context mocks, and context-owned cleanup. |
| `docs/testing/testing-external-behavior.md` | Assert through the closest production-visible boundary | Pass | Platform tests assert resulting chat/thread state and connection indicator behavior; lower-level tests stay at the otherwise unavailable Worker/MessagePort boundary. |

### Findings

- None.

- `turbo/apps/platform/src/shared-database/worker-runtime.ts:343-352,1108-1130` and `turbo/apps/platform/src/shared-database/worker-realtime.ts:83-102`: the public subscription completes before the worker has established a lossless realtime-to-query boundary. `subscribe()` records the data key and starts a realtime session, but returns immediately; `ensureRealtimeForUser()` stores `session.ready` without awaiting or otherwise using it. A tab then performs its required catch-up query while the Ably channel may still be attaching. An event committed after that HTTP response but before attachment can be absent from both the query and realtime delivery, with no later notification to recover it. The mock registers callbacks synchronously and attaches in a microtask, so current tests cannot expose the real network window. Please either make registration readiness part of the operation or schedule a mandatory post-attach catch-up for subscribed actors, while preserving cache-first rendering, and add a delayed-attach regression test for both datasets.

- `turbo/apps/platform/src/shared-database/worker-runtime.ts:468-486,505-523,1166-1204`: a single transient HTTP/validation failure while handling an Ably notification permanently consumes that notification. The completion path marks the actor dirty and rejects; the detached caller settles the rejection, but no retry is scheduled. `heartbeat()` only drains dirty actors when a *different token* clears `authBlocked` (`worker-runtime.ts:304-317`), so an ordinary 5xx on the final ChatEvent or `threadListChanged` update leaves tabs stale until another notification or manual query happens. The replaced tab-owned realtime loops had bounded transient retries. Please retain/retry dirty subscribed actors under worker-owned lifecycle with bounded backoff (without retrying auth-blocked credentials), and test one notification whose first catch-up fails and whose retry succeeds without a second notification.

- `turbo/apps/platform/src/signals/chat-page/chat-thread-event-sourcing.ts:427-445,548-569` and `turbo/apps/platform/src/shared-database/worker-runtime.ts:738-851`: the enabled path drops the existing `EVENT_LOG_SNAPSHOT_REBASE_THRESHOLD = 100` maintenance behavior. The legacy path performs a best-effort Snapshot rebase after its first incremental sync when the retained log exceeds the bound; the SharedWorker only replaces on missing/degraded state or 410. Because the ChatThread dataset returns its complete Snapshot plus event log across MessagePort on every invalidation, this makes local reads, validation, cloning, and replay grow with every event until the server happens to expire the cursor. Please move the bounded rebase policy into the worker and add coverage proving a valid >100-event cache is compacted without exposing a reset to the Platform.

- `turbo/apps/platform/src/shared-database/worker-runtime.ts:601-613,691-734`: the worker sends `X-Chat-Event-Schema-Version` but never verifies the response header, unlike the existing canonical row source, so a mismatched response can be accepted and persisted under the current cache schema. The same block adds `snapshot.body.lastEventId ?? rows.at(-1)?.id ?? null`, which is the repository's explicitly temporary previous-API compatibility path (the contract/source comments name the 102-minute observed window and follow-up #27194), but the PR's `## Fallbacks` section does not list it. Please share the current response-version assertion with the worker, add mismatch coverage, and either remove this tolerance with evidence or declare the worker copy with its surface, window, removal condition, and follow-up. Per `docs/fallback.md`, the missing declaration independently blocks an LGTM even if the tolerance remains justified.

### Fallbacks

- `turbo/apps/platform/src/shared-database/worker-runtime.ts` — `ChatEventActor.degraded`, `ChatThreadEventActor.degraded`, and their cache read/write recovery paths: IndexedDB failure leaves `cache-only` network-free while `catch-up` returns canonical remote data. Surface/window: none; non-GA feature and persistent browser-storage resilience. Removal: only if working IndexedDB becomes a product prerequisite. Declared: Yes. Verdict: justified; the previous recursive-notification defect is fixed.
- `turbo/apps/platform/src/shared-database/worker-runtime.ts:732` — derives `lastEventId` from the immutable Snapshot body when the previous API omits the response field. Surface/window: new app worker versus previous API; existing code records an observed maximum of 102 minutes and follow-up #27194. Declared: No. Verdict: the current contract still documents the compatibility state, but the new worker branch must be declared (or removed with rollout evidence) before merge.

### Testing

- Current-head CI is green, including app tests, lint, types, CodeQL/Semgrep, build/deploy previews, and `codecov/patch`; Codecov reports 88.34% patch coverage.
- Current-head local verification already completed format, the bounded full-repository lint, full type checks, Lefthook/Knip, and the four focused SharedWorker files (21 tests).
- The previous failure-mode gaps are now covered: both degraded datasets suppress repeat invalidations, a full direct bridge proves failed IDB writes do not recurse, and a two-client structured-clone protocol test proves stale registration recovery with real timers.
- Remaining coverage is insufficient for the four P1 cases above: delayed/failed first attachment, retry after one transient remote failure with no second notification, bounded ChatThread Snapshot rebase, and response schema-version mismatch/compatibility declaration.

e7h4n:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 623d6e67cd28f6e70d9a020aa9ae7eb48435b69a

### Summary

This revision closes the previously reported Ably attach gap, transient catch-up loss, unbounded ChatThread event log, response-version validation gap, and undeclared compatibility path. The content-hashed SharedWorker, canonical event synchronization, IndexedDB degradation behavior, identity checks, and feature-gated legacy path are coherent. Before merge, the Vitest architecture must mirror the production Platform/Worker store boundary, MessagePort cancellation must stop only the tab's local wait rather than creating handler-owned request lifecycles, and the heartbeat loop must run through `setLoop` under the shared test lifecycle with deterministic deferred-gate coverage.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness, races, and async ownership | Finding (P1) | `message-port-server.ts:30-114` creates and aborts a child controller per RPC; `shared-database-browser.ts:121-170` bypasses the real loop in Vitest and implements interval waiting inside the loop body. |
| `REVIEW.md` | Security and type safety | Pass | MessagePort inputs and dataset results use strict Zod schemas; worker queries validate user/org identity and canonical event contracts. |
| `REVIEW.md` | Feature containment and regression protection | Pass | `SharedChatDatabase` is off by default/staff-only and both legacy ChatEvent/ChatThreadEvent paths remain selected when disabled. |
| `docs/docs.md` | Apply every practice routed by the changed Platform, test, event, cache, and deployment surfaces | Pass | All indexed documents relevant to this diff were reviewed. |
| `docs/bad-smell.md` | Avoid `any`, suppressions, dynamic production imports, hidden configuration, and unjustified recovery | Pass | No new `any`, suppression, dynamic production import, or hard-coded deployment origin is present. |
| `docs/effect.md` | Long-running work and cancellation have an explicit external owner | Finding (P1) | The MessagePort handler manufactures per-request controllers instead of consuming the worker/test lifecycle signal; heartbeat scheduling is split between `setLoop` and an inner timer. |
| `docs/cache.md` | Store/resource identity and teardown match the real owner | Finding (P1) | Production has separate Platform and Worker stores, while `SharedWorkerTestBootstrap` installs both halves into one store and worker tests execute commands on `context.store`. |
| `.claude/skills/ccstate/SKILL.md` | Commands run in their owning Store and receive a lifecycle-owned final `AbortSignal` | Finding (P1) | `test-bridge.ts:36-144` has a single Store for Platform and worker commands; `message-port-server.ts:57-86` creates an internal signal per request. |
| `docs/event-sourcing.md` | Persistent server events remain authoritative and optimistic events reconcile by shared event ID | Pass | Shared results are canonical contract rows/events and flow through the existing merge and optimistic-reconciliation commands. |
| `docs/fallback.md` | Every introduced fallback is justified, declared, bounded where applicable, and removable | Pass | Both IDB degradation and the temporary ChatEvent response-shape tolerance are declared in `## Fallbacks`; the latter names the 102-minute API window and #27194 cleanup. |
| `docs/deployment-compatibility.md` | Browser/API and persisted-state skew are handled explicitly | Pass | The worker sends and validates the current ChatEvent schema version, uses the existing user/org IDB, and retains the declared previous-API tolerance until #27194. |
| `docs/react-commit.md` | Realtime invalidation does not cause no-op writes or unbounded fan-out | Pass | Actor high-water marks suppress identical degraded retries, catch-ups coalesce, and tabs receive dataset-keyed invalidations rather than pushed record copies. |
| `docs/testing.md` | Feature and regression tests exercise production lifecycles without fake timers | Finding (P1) | Current tests do not exercise the production heartbeat loop and do not provide the required context-owned worker Store; one test also uses `vi.spyOn`. |
| `docs/testing/patterns.md` | Platform tests use `testContext`-owned stores, mocks, and cleanup | Finding (P1) | `message-port.test.ts:126-146` and other worker tests create ad hoc Stores instead of using one `context.store` plus one `context.workerStore`. |
| `docs/testing/anti-patterns.md` | Keep real async scheduling and avoid implementation-controlled timing | Finding (P1) | `shared-database-browser.ts:123-125` disables the loop in Vitest, preventing real macrotask scheduling from being tested; `platform-direct-bridge.test.ts:223` spies on DOM state. |
| `docs/testing/app-testing.md` | Platform bootstrap and browser mocks share one TestContext lifecycle | Finding (P1) | `page-helper.ts:151-155` bootstraps the worker into the Platform Store rather than a context-owned isolated Worker Store sharing `context.signal`. |
| `docs/testing/testing-external-behavior.md` | Assert outcomes through the closest production boundary | Pass | Direct-bridge tests assert rendered signal outcomes and MessagePort tests exercise structured-cloned transport behavior; the missing work is lifecycle fidelity, not a new internal assertion surface. |
| Ethan's #27270 merge requirement | TestContext exposes one Platform Store and one Worker Store sharing only the test-lifetime signal | Finding (P1) | `TestContext` currently exposes only `store`; direct bootstrap and worker tests collapse or recreate the two sides. |
| Ethan's #27270 merge requirement | RPC cancel discards the tab response while worker-owned catch-up continues | Finding (P1) | `cancel` currently aborts the handler-created request controller. |
| Ethan's #27270 merge requirement | Heartbeat runs under `setLoop` in Vitest and is advanced through deferred worker gates | Finding (P1) | The loop returns early under Vitest, manually delays in its body, and has no N-round deferred-gate regression test. |

### Findings

- None.

- `turbo/apps/platform/src/signals/__tests__/test-helpers.ts:14-69`, `turbo/apps/platform/src/__tests__/page-helper.ts:151-155`, `turbo/apps/platform/src/shared-database/test-bridge.ts:36-144`, and the SharedWorker tests: the test topology does not match production isolation. `TestContext` has only `context.store`; `SharedWorkerTestBootstrap` runs the worker bootstrap, worker commands, bridge installation, and Platform status writes through that same Store, while `message-port.test.ts:126-146` and other tests compensate by constructing ad hoc Stores. Add a lazily owned `context.workerStore` with no separate worker signal; both stores must share the single `context.signal`. Vitest bootstrap must start the real worker bootstrap in `context.workerStore`, install the bridge/signals in `context.store`, and make `DirectSharedDatabaseBridge` hold both stores so every worker command executes only on `workerStore`. Update worker and MessagePort tests to use those context stores rather than creating per-test Store pairs, and prove aborting `context.signal` tears down both sides.

- `turbo/apps/platform/src/shared-database/message-port-server.ts:30-114`: the server creates `createChildAbortController(signal)` for every RPC and lets a tab `cancel` abort that controller. That gives the business handler ownership of a transport/request lifecycle and makes cancellation semantics depend on an internally manufactured signal. Consume the externally owned request/lifecycle signal in the handler, keep client abort local to the client's RPC wait, and treat `cancel` as a marker that suppresses the eventual response while the worker-owned catch-up continues under the worker lifecycle. If the transport needs request-signal bookkeeping, centralize it in the transport abstraction rather than creating controllers inside the business handler. Preserve the structured-clone regression proving a cancelled tab wait does not cancel the shared catch-up.

- `turbo/apps/platform/src/signals/shared-database-browser.ts:1-3,121-170` and `turbo/apps/platform/src/shared-database/__tests__/platform-direct-bridge.test.ts:222-235`: heartbeat behavior and coverage diverge between production and Vitest. The loop returns immediately when `IN_VITEST`, passes `0` to `setLoop`, and performs its own `delay` inside the loop body, so tests never exercise the production scheduling path. Remove the Vitest early return and inner delay; let `setLoop` own the heartbeat interval, retain the bootstrap-time immediate heartbeat, and use the existing `waitBeforeFirstPoll` pattern to avoid an immediate duplicate. Keep focus/visible transitions as separately triggered immediate heartbeats. Add a test that runs the loop with Vitest's real 0 ms macrotask progression, blocks the worker heartbeat target with successive `context.mocks.deferred` gates, resolves one gate per round, asserts exactly N rounds, then aborts the shared `context.signal` to stop both stores. Do not use `vi.spyOn` or fake timers; replace the existing visibility spy with a context-owned browser/diagnostic input.

### Fallbacks

- `turbo/apps/platform/src/shared-database/worker-runtime.ts` — `ChatEventActor.degraded`, `ChatThreadEventActor.degraded`, and their cache read/write paths allow `cache-only` to remain network-free while `catch-up` returns canonical server data when browser IndexedDB fails. Surface/window: none; persistent browser-storage resilience for the non-GA feature. Removal condition: only if functioning IndexedDB becomes a hard product prerequisite. Declared in PR summary: Yes. Verdict: Justified.
- `turbo/apps/platform/src/shared-database/chat-event-schema-version.ts` and `worker-runtime.ts` — tolerate a missing ChatEvent response-version header and derive a missing Snapshot `lastEventId` when the new app worker reaches the previous API. Surface/window: web/app to API, observed maximum 102 minutes. Removal condition: require both fields and delete the tolerance through #27194 after the API skew window closes. Declared in PR summary: Yes. Verdict: Justified and time-boxed.

### Testing

- Coverage: Insufficient for the newly required store-isolation and heartbeat lifecycle. Existing coverage remains strong for delayed Ably attachment, transient retry, snapshot rebase, schema mismatch, cursor reset, 401 recovery, IDB degradation, stale-tab reconnect, structured cloning, and both datasets.
- Conventions: Changes required — use `context.store` plus `context.workerStore`, one shared `context.signal`, no per-test Store pairs, no handler-owned RPC controllers, no Vitest loop bypass, no `vi.spyOn`, and no fake timers.
- Current CI for `623d6e67cd28f6e70d9a020aa9ae7eb48435b69a` is green, including app tests, lint/types, security checks, preview builds, and `codecov/patch`; those checks predate the supplemental merge requirements above.

e7h4n:
Preview regression on `a635b8f331` (worker code unchanged by the subsequent main merge): **PASS**

- Enabled `sharedChatDatabase` and forced a cold IndexedDB load.
- Confirmed the content-hashed module SharedWorker was running.
- Removed API-origin Preview cookies before the decisive run, so Worker access could only succeed through the heartbeat-forwarded bypass header.
- Every captured SharedWorker API request carried `X-Vercel-Protection-Bypass`; ordinary non-Preview behavior is separately covered by a negative runtime test.
- `/api/okou/realtime/token` returned 200; Ably token exchange returned 201 and connected without `Failed to fetch`.
- ChatThread snapshot/events returned 200. ChatEvent snapshot returned the expected 404 when no stored snapshot existed, followed by rows returning 200.
- Observed no 403 response, CORS error, or loading failure.

This closes the Preview CORS/403 merge blocker.

e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 05c4547ce8207265ed770458eeef7ff31329c7ff
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

The current revision implements the requested feature-gated SharedWorker architecture for ChatEvent and ChatThreadEvent only. The production bridge uses a content-hashed module worker and an independent ccstate store; the Vitest bridge mirrors the Platform/Worker store split while sharing only the test-lifetime signal. Heartbeat identity and the dedicated Preview bypass cross the typed MessagePort protocol, query results are dataset-specific and Zod-validated, and append notifications carry the full user/org-scoped data key. Cache-only first paint, worker-owned catch-up, transparent cursor rebuild, IDB degradation, 401 recovery, stale-tab transport recreation, hidden-tab status, and versionchange reload behavior are all covered without changing the API contracts.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness, races, lifecycle, security, typing, and feature containment | Pass | Worker actors coalesce catch-up, use monotonic observed sequence IDs for invalidation, validate identity before and after queries, and keep the legacy paths selected while the switch is disabled. |
| `docs/docs.md` | Apply the practices routed by the changed Platform, cache, event, deployment, and test surfaces | Pass | The relevant ccstate, effect, cache, event-sourcing, fallback, deployment, React-commit, and testing practices were reviewed against the complete diff. |
| `docs/bad-smell.md` | Avoid `any`, suppressions, dynamic production imports, hidden deployment configuration, and arbitrary header forwarding | Pass | The diff adds no `any` or suppression; the worker entry is a static Vite worker URL, API base resolution stays centralized, and only the dedicated Preview bypass header is forwarded. |
| `docs/effect.md` | Async work has an external owner and cancellation preserves shared work | Pass | Platform RPC waits use caller signals, MessagePort cancel only discards the response, worker catch-up uses the worker lifecycle signal, and connection/subscription generations are aborted and recreated explicitly. |
| `docs/cache.md` | Cache identity, ownership, invalidation, and teardown match the resource | Pass | Actors and databases are keyed by user/org plus dataset/thread, unused actors and realtime sessions are released, and versionchange closes the existing DB before requesting tab reload. |
| `.claude/skills/ccstate/SKILL.md` | Commands run in the owning store and receive lifecycle-owned signals | Pass | Platform commands stay in `context.store`, worker commands stay in `context.workerStore`, both test stores share `context.signal`, and production creates an independent worker store. |
| `docs/event-sourcing.md` | Canonical immutable events remain authoritative and optimistic state reconciles by event identity | Pass | ChatEvent rows and ChatThreadEvent snapshots/events come from existing contract schemas and flow through the existing ID-based merge/reconciliation projections. |
| `docs/fallback.md` | Fallbacks are declared, justified, bounded where applicable, and removable | Pass | The PR declares both IDB degradation and the temporary previous-API ChatEvent response tolerance with their surfaces and lifecycle/removal conditions. |
| `docs/deployment-compatibility.md` | Browser/API and persisted-state version boundaries are explicit | Pass | The worker sends and checks the current ChatEvent schema version, reuses the existing user/org IndexedDB schema, handles versionchange, and time-boxes the previous-API response tolerance through #27194. |
| `docs/react-commit.md` | Realtime updates avoid no-op invalidation loops and broad pushed payloads | Pass | Tabs receive keyed append invalidations and re-query through their own cursors; actor high-water marks suppress identical degraded retries instead of rebroadcasting every failed IDB write. |
| `docs/testing.md` | New behavior has integration coverage without fake timers or spies | Pass | The tests use real macrotask progression, context-owned deferred gates, real IDB, contract-aware MSW handlers, and no new fake timers or `vi.spyOn`. |
| `docs/testing/patterns.md` | Tests use shared TestContext ownership and production-like boundaries | Pass | MessagePort tests structured-clone across distinct stores, direct-bridge tests enter through page bootstrap, and runtime HTTP/Ably behavior uses `context.mocks`. |
| `docs/testing/anti-patterns.md` | Avoid implementation mocks and test-only timing branches | Pass | The heartbeat loop uses production `setLoop` behavior in Vitest, and cancellation/attachment/retry ordering is driven by deferred boundary controls. |
| `docs/testing/app-testing.md` | Platform behavior enters through normal bootstrap | Pass | Direct-bridge coverage uses `setupPage`/`detachedSetupPage`, the real feature-switch bootstrap, and the context-owned Platform and Worker stores. |
| `docs/testing/testing-external-behavior.md` | Assert at the closest observable production boundary | Pass | Coverage verifies rendered signal state, hidden-tab realtime status, cache-first/catch-up behavior, MessagePort propagation, request headers, and append-driven re-queries. |

### Findings

- None.

- None.

### Fallbacks

- `turbo/apps/platform/src/shared-database/worker-runtime.ts` — `ChatEventActor.degraded`, `ChatThreadEventActor.degraded`, and the strict cache read/write recovery paths keep `cache-only` network-free while allowing `catch-up` to return canonical server data when browser IndexedDB fails. Surface/window: none; this is persistent browser-storage resilience behind the non-GA `SharedChatDatabase` switch. Removal condition: only if working IndexedDB becomes a hard product prerequisite. Declared: Yes. Verdict: Justified.
- `turbo/apps/platform/src/shared-database/chat-event-schema-version.ts` and `worker-runtime.ts` — tolerate a missing response-version header and derive a missing Snapshot `lastEventId` when the new web/app worker reaches the previous API during rollout or rollback. Surface/window: new web/app versus previous API, observed maximum 102 minutes. Removal condition: require both fields and remove the tolerance through #27194 after the skew window closes. Declared: Yes. Verdict: Justified and time-boxed.

No other version-migration fallback was introduced by this diff; worker reconnection, transient realtime retry, cursor-expiry rebuild, and IDB degradation are runtime recovery behavior rather than old/new deployment compatibility.

### Testing

- Coverage: Adequate. The focused suites cover both datasets, production-like store isolation, MessagePort correlation and cancellation, stale-client reconnection, delayed/failed Ably attachment, transient catch-up retry, cursor expiry, snapshot rebase, strict schema validation, IDB failure, 401/token recovery, Preview-bypass propagation, non-Preview header omission, hidden-tab status, and append-driven Platform reads.
- Conventions: Pass. No fake timers or new spies are used; async progression is controlled through `context.mocks.deferred`, MSW contract handlers, real IndexedDB, and shared lifecycle aborts.
- Verification evidence: the current head's Platform lint/type/app-test and security lanes are green; the production Vite build emitted the revisioned worker asset, and the Preview regression confirmed worker realtime/token and snapshot/rows requests succeed with the dedicated bypass and no 403/CORS failure.

PR #27272, opened 2026-08-14
fix(runner): preserve runs after tool process oom

### LINKED ISSUE #27269: bug(runner): preserve agent runs after tool process OOM
## Background

Runner v0.168.2 introduced separate Guest control and workload cgroups so an untrusted workload cannot starve Guest Agent heartbeat, cancellation, checkpointing, and cleanup. The workload cgroup has a finite `memory.max` and currently sets `memory.oom.group=1`.

PR #27206 raised the default workload hard limit from Guest-visible capacity minus 384 MiB to capacity minus 128 MiB so Guest reclaim and balloon relief run before a workload-local OOM. That change is deployed in Runner v0.168.11 and is working as designed, but production evidence exposed a separate recovery regression in the OOM policy.

## Problem

Run `87543ac9-4fb6-4ffe-879a-d74b4b5b477e` on `prod-3.aws.vm3.ai` used Runner v0.168.11 and a newly created sandbox. The Firecracker balloon had already deflated to 0 MiB and Guest `MemAvailable` was about 12 MiB before the workload reached its hard limit.

The workload OOM reported:

```text
codex app-server child exited while waiting for turn notification: signal: 9 (SIGKILL); workload resource limit reached (memory_max=673808, memory_oom=1, memory_oom_kill=8, memory_oom_group_kill=1, pids_max=0)
```

`memory_max` is the `memory.events:max` counter, not a byte limit. The important evidence is `memory_oom_group_kill=1`: because the Codex app-server and every tool descendant share the workload leaf, `memory.oom.group=1` killed all eight workload processes. A recoverable high-memory tool failure therefore became an unconditional run failure.

Before workload resource isolation, a Guest-wide OOM selected an individual process. In the common case the high-memory compiler, browser, or other tool was killed while the agent runtime survived and could report the tool failure and continue. The current group-atomic policy intentionally preserves Guest Agent control liveness, but it unnecessarily removes that run-level recovery behavior.

## Required behavior

- Keep the finite workload `memory.max` and the separate protected Guest Agent control cgroup.
- Do not treat the workload as an indivisible OOM group. A workload-local OOM should use normal per-process victim selection so a high-memory tool can fail without automatically killing the agent runtime.
- A run may still fail when the agent runtime itself is the selected victim or cannot recover from the tool failure.
- Cancellation, timeout, terminal cleanup, and sandbox reuse preparation must continue to remove the complete operation process tree through the existing explicit cgroup cleanup path.
- Preserve structured resource-limit diagnostics without turning a surviving, successfully completed agent run into a failure solely because an individual workload process was OOM-killed.

## Acceptance criteria

- New workload cgroups configure `memory.oom.group=0` while retaining the current `memory.max`, `memory.high=max`, CPU policy, PID policy, and control `memory.min` policy.
- Reuse validation requires the current per-process OOM policy and safely rejects a sandbox carrying a different policy.
- Production-equivalent integration coverage creates an agent-runtime analogue with a high-memory tool child, reaches the workload hard limit, and verifies:
  - at least one individual OOM kill is recorded;
  - no workload group OOM kill is recorded;
  - the runtime process survives the tool failure;
  - the runtime can perform a subsequent operation;
  - terminal cleanup and later sandbox reuse still succeed.
- Existing process-containment, diagnostics, cancellation, cleanup, and reuse checks remain green.

## Risks

- Linux selects an OOM victim by process badness; it cannot identify a logical tool task. If the agent runtime is the principal memory consumer, the run can still fail.
- Multiple unreclaimable consumers can cause multiple individual OOM kills before memory pressure clears.
- A killed tool can leave partial filesystem output, which is already possible for cancellation, timeout, and ordinary tool failure and must remain visible to the agent.

## Relevant code and prior work

- `crates/vsock-guest/src/process_containment.rs`
- `crates/guest-agent/src/workload_containment.rs`
- `crates/guest-agent/src/reuse_preparation.rs`
- `.github/scripts/runner-behavior-process-containment.sh`
- #26683
- #27126
- #27188 / #27206

seven332:
## Summary

- disable group OOM kills for workload cgroups so the kernel can select the memory-hungry process
- enforce the same policy when preparing reused sandboxes
- update process-containment behavior coverage to prove a tool SIGKILL does not fail the run and the sandbox remains reusable

## Scope

- keeps the existing workload memory limit, control memory protection, CPU/PID policies, and resource diagnostics
- does not introduce nested runtime/tool cgroups or guarantee a specific victim when the agent runtime itself is the largest consumer

## Validation

- `cargo clippy --manifest-path crates/Cargo.toml --all-targets --all-features`
- `cargo clippy --manifest-path crates/Cargo.toml -p vsock-guest --release --no-default-features --all-targets`
- `cargo doc --manifest-path crates/Cargo.toml --workspace --all-features --no-deps`
- `cargo test --manifest-path crates/Cargo.toml -p guest-contracts`
- `cargo test --manifest-path crates/Cargo.toml -p guest-agent --test reuse_preparation_helper`
- `bash -n .github/scripts/runner-behavior-process-containment.sh`
- `shellcheck .github/scripts/runner-behavior-process-containment.sh`

The broader Rust suite passed before the final rebase. On the rebased head, an additional full-suite run exposed pre-existing parallel workspace-cache test contention (`LockBusy`); each reported test passes in isolation and the affected tests above pass on the current head.

Closes #27269

### DIFF
--- a/.github/scripts/runner-behavior-process-containment.sh
+++ b/.github/scripts/runner-behavior-process-containment.sh
@@ -122,6 +122,7 @@ test "$(cat "$parent/control/memory.min")" = "$expected_control_memory_min"
 grep -Eq '^[0-9]+ [0-9]+$' "$parent/workload/cpu.max"
 test "$(cat "$parent/workload/memory.high")" = max
 test "$(cat "$parent/workload/memory.max")" = "$expected_workload_memory_max"
+test "$(cat "$parent/workload/memory.oom.group")" = 0
 test "$(cat "$parent/workload/pids.max")" = max
 test -z "${VM0_WORKLOAD_CGROUP_PROCS_ENDPOINT:-}"
 control_member_count=$(wc -l < "$parent/control/cgroup.procs")
@@ -672,16 +673,29 @@ read -r METRICS_COUNT METRICS_SPAN_SECS METRICS_MAX_GAP_SECS <<<"$METRICS_SUMMAR
 rm -f "$PRESSURE_SUBMIT_OUTPUT"
 PRESSURE_SUBMIT_OUTPUT=""
 
-echo "--- Pressure: exhaust workload memory without killing Guest Agent ---"
+echo "--- Pressure: kill a high-memory tool without terminating the run ---"
 # Reuse the pressure VM again to prove reclaim left it safe to park while
-# isolating terminal OOM from the active-input provider session.
+# isolating workload OOM from the active-input provider session.
 MEMORY_CHAT_THREAD_ID="$PRESSURE_CHAT_THREAD_ID"
 MEMORY_SESSION_ID="e2e-process-containment-memory"
 MEMORY_PROMPT=$(cat <<'PROMPT'
+set -eu
 test -f /tmp/vm0-process-containment/memory-reclaim-vm
 sudo -n python3 - <<'PY'
+import os
 import pathlib
+import signal
+import subprocess
 import time
+
+
+def read_events(path):
+    events = {}
+    for line in path.read_text().splitlines():
+        key, value = line.split()
+        events[key] = int(value)
+    return events
+
 
 relative = next(
     line.removeprefix("0::").strip()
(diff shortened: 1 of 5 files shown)


seven332:
Reviewed current head `a92ae043dee4a281f410c47ffa6402e0978c5038`: LGTM.

The change keeps the workload hard memory ceiling while switching from unconditional cgroup-wide OOM termination to per-process victim selection. Startup and reuse verification consume the same shared policy, and the process-containment behavior test now proves the allocating child is SIGKILLed while the run completes and the sandbox remains reusable.

No P0/P1/P2 findings. The remaining behavioral boundary is explicit: if the agent runtime itself is the actual largest memory consumer, the kernel may still select it. The production-equivalent metal behavior check is left to CI.

PR #27297, opened 2026-08-14
fix(api): stabilize custom connector firewall ids

e7h4n:
## Summary

- assign stable API IDs to the initial inline custom-connector firewall snapshot
- build both the runner claim and runtime sync snapshots through the same canonical helper
- assert that the claimed firewall exactly matches the first runtime sync firewall

## Root cause

The initial runner claim omitted custom-connector API IDs, so the runner synthesized run-scoped IDs. The immediately scheduled runtime sync used stable `custom_connector_<id>:<index>` IDs and replaced the registry snapshot. If that replacement landed while credential authorization was in flight, the MITM revalidation correctly treated the firewall authorization as changed and failed closed with HTTP 409 (`firewall_authorization_changed`). The command ignored the curl failure and printed its completion marker, leaving the E2E observer waiting for an allow telemetry event that could never be emitted.

This keeps the security revalidation intact: genuine firewall changes still fail closed, while the initial idempotent sync no longer creates a false authorization change.

## Validation

- demonstrated the new equality assertion fails against the old initial snapshot because its API ID is missing
- focused API regression test: 5/5 passes, plus one pass after rebasing onto latest `main`
- `pnpm exec prettier --check` on both changed files
- targeted Oxlint and ESLint on both changed files
- `pnpm -F api run check-types`

The runner E2E is CI-only and is left to the PR Turbo pipeline, as documented.

Incident: https://github.com/vm0-ai/vm0/actions/runs/31843290292/job/94905496874

Turbo flaky key: `cli-e2e-runner-custom-connector-firewall-telemetry-timeout`

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
@@ -8403,6 +8403,7 @@ describe("RUN-02: custom connectors, grants, and network policies", () => {
     expect(initialAvailable.nextSyncAt).toBeUndefined();
     expect(initialAvailable.target).toStrictEqual(targetIdentity);
     expect(initialAvailable.baseUrlVars).toStrictEqual({ tenant: "acme" });
+    expect(initialAvailable.firewall).toStrictEqual(customFirewall);
     const { api: initialApi, body: currentAuthBody } =
       customConnectorRuntimeAuthBody(
         initialAvailable,
--- a/turbo/apps/api/src/signals/services/agent-run-create.service.ts
+++ b/turbo/apps/api/src/signals/services/agent-run-create.service.ts
@@ -4435,14 +4435,16 @@ async function buildNewRunCustomConnectorRuntimeContext(
   };
 }
 
+type CustomConnectorRuntimeFirewall = Omit<Firewall, "apis"> & {
+  readonly apis: (Firewall["apis"][number] & {
+    readonly id: string;
+  })[];
+};
+
 interface CustomConnectorRuntimeExecutionState {
   readonly firewall: Omit<ExecutionFirewallInlineEntry, "firewall"> & {
     readonly customConnectorId: string;
-    readonly firewall: Omit<Firewall, "apis"> & {
-      readonly apis: (Firewall["apis"][number] & {
-        readonly id: string;
-      })[];
-    };
+    readonly firewall: CustomConnectorRuntimeFirewall;
   };
   readonly networkPolicy: NetworkPolicy;
 }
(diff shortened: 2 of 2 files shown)


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: eabadf4abbe6e04e6b7c0fd1c6a25e55bb5fe460
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

This fix gives the initial custom-connector firewall in the runner claim the same stable API IDs as the first runtime-sync snapshot. The shared builder removes the false authorization-change boundary while preserving fail-closed behavior for genuine firewall changes, and the route-level regression assertion covers the two externally visible runner API responses.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and race safety | Pass | `customConnectorRuntimeFirewall` is the single constructor for both claim and sync snapshots; IDs are stable per firewall name and API index. |
| `REVIEW.md` | Security and fail-closed behavior | Pass | The change does not weaken authorization revalidation or credential handling; it only removes a false snapshot difference. |
| `REVIEW.md` | Type safety and maintainability | Pass | The refined `CustomConnectorRuntimeFirewall` type requires non-empty-at-construction IDs without `any`, casts, or suppressions, and replaces duplicated mapping logic. |
| `REVIEW.md` | Fix regression coverage | Pass | `run-lifecycle.bdd.test.ts` now asserts exact equality between the claimed inline firewall and the first runtime-sync firewall. |
| `REVIEW.md` | Fallback, feature-switch, and DB/JSONB gates | N/A | The diff introduces no fallback or compatibility branch, no feature, and no persisted database contract change. |
| `docs/bad-smell.md` | Avoid `any`, suppressions, dynamic imports, hardcoded configuration, and fallback recovery | Pass | Neither changed file adds any prohibited construct; the canonical helper fails no errors and adds no recovery path. |
| `docs/testing.md` | Prefer integration coverage with deterministic synchronization and boundary mocks only | Pass | The regression extends an existing API route integration flow and adds no sleep, retry, fake timer, direct fetch mock, or internal mock. |
| `.claude/skills/ccstate/SKILL.md` | Preserve signal async ownership and avoid accessor escape or detached work in signals | N/A | The changed signals service code is a synchronous pure projection and adds no command, promise, lifecycle signal, `detach`, or accessor plumbing. |
| `docs/deployment-compatibility.md` | Keep runner/backend boundary changes additive across rollout | Pass | `id` is already optional in the execution schema, Rust deserialization has `#[serde(default)]`, and the Python registry preserves non-empty inline IDs; old and current runners remain compatible. |
| `docs/testing/patterns.md` | Use route integration structure and assert behavior | Pass | The added assertion remains inside the established run lifecycle route test and compares returned contract data rather than mock calls. |
| `docs/testing/anti-patterns.md` | Avoid internal mocks, implementation-call assertions, fake timers, and artificial waits | Pass | The diff adds only an external response equality assertion and no new mocks or timing controls. |
| `docs/testing/api-testing.md` | Exercise API behavior through the real route harness | Pass | The existing BDD flow claims the runner job and calls runtime sync through the API harness; no service or DB shortcut was added. |
| `docs/testing/testing-external-behavior.md` | Construct and observe state at the API boundary | Pass | The regression observes the claim and sync payloads that the runner consumes, with no assertion on internal storage or service calls. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- Coverage: Adequate. The new assertion was demonstrated to fail on the old snapshot because the initial API ID was missing, then passed 5/5 focused runs plus one run after rebasing onto the latest `main`.
- Conventions: Compliant. The test remains route-level integration coverage with real application code and deterministic response comparison.
- Static validation: Prettier, targeted Oxlint and ESLint, and the complete API type-check sequence passed.

PR #27312, opened 2026-08-15
fix(runner): start workspace cache lock timeout after contention

### LINKED ISSUE #27307: bug(runner): prevent false workspace cache lock-busy results under load
## Background

`WorkspaceImageCache::prepare` waits up to 50 ms for an entry lock. Issue #25810 introduced this
bounded wait for a real short-lived holder, and #25975 made that contention path deterministic.

While validating an unrelated runner change, the full Rust suite exposed a different failure mode:
tests using isolated temporary cache roots intermittently receive
`WorkspaceCacheCheckoutResult::LockBusy` when no intentional holder exists.

## Problem and evidence

The full Rust suite failed in three workspace-cache scenarios:

```text
cmd::start::tests::main_loop::admission::saturated_cache_only_holder_defers_before_reclaiming_unrelated_idle
executor::tests::sandbox_run_tests::workspace_cache::execute_inner_cache_hit_retry_requires_completed_cleanup
executor::tests::sandbox_run_tests::workspace_cache::process_timeout_invalidates_consumed_workspace_cache_without_fallback
```

An immediate runner-only full-suite rerun failed in two different scenarios:

```text
executor::tests::sandbox_run_tests::workspace_cache::cached_reuse_workspace_promotion_identity_mismatch_stops_before_agent
executor::tests::sandbox_run_tests::workspace_cache::execute_inner_retries_fresh_after_workspace_cache_hit_create_failure
```

All five tests pass individually. Their fixtures use distinct `tempfile::TempDir` cache and lock
roots, and some failures occur while initially seeding an entry before any intended holder exists.

`WorkspaceImageCache::prepare` currently wraps the complete acquisition future in:

```text
tokio::time::timeout(WORKSPACE_IMAGE_PREPARE_LOCK_TIMEOUT, crate::lock::acquire(lock_path))
```

The secure file open and non-blocking `flock` attempt inside `lock::acquire` run through
`tokio::task::spawn_blocking`. The 50 ms deadline therefore includes blocking-pool and scheduler
delay before the first attempt. If that delay wins, lifecycle code has no busy observation but still
classifies the result as `LockBusy`. This can unnecessarily downgrade a production cache hit to a
cold workspace as well as make full-suite tests fail under load.

## Decision

The 50 ms duration is a contention budget, not an end-to-end lock-operation timeout. It must begin
only after an actual acquisition attempt returns `EWOULDBLOCK` for the current lock inode.

## Chosen direction

- Add a typed exclusive acquisition operation in `crates/runner/src/lock.rs` whose first blocking
  attempt has no contention deadline.
- Return the guard immediately on success and propagate real open, validation, permission,
  replacement, join, and `flock` errors.
- After the first typed busy result, start the 50 ms timeout around the existing delayed retry loop,
  preserving the 5/10/20 ms backoff cadence and cancellation-safe non-blocking attempts.
- Have `WorkspaceImageCache::prepare` consume the typed acquired/busy result, retain the guard on
  success, and use the existing fresh-workspace fallback only for genuine bounded contention or the
  existing lock-error path.
- Add lifecycle integration coverage with a one-slot Tokio blocking pool, channels, direct future
  polling, and paused time. Delay the first attempt past 50 ms while the target entry is unlocked,
  prove prepare remains pending, then release the slot and require a normal `Miss`.

## Alternatives considered

- Calling `try_acquire_or_busy` in lifecycle and then timing `acquire` would fix the classification
  with less API surface, but it duplicates an immediate attempt and changes the established retry
  cadence.
- Exposing an attempt callback, observer, or production trace event adds cross-layer state and
  unsupported observability surface when the existing typed busy result is sufficient.
- Moving filesystem lock work onto async workers or a dedicated pool broadens the execution-model
  change to every runner lock and risks blocking runtime workers.
- Increasing the timeout only lowers the probability and increases genuine-holder fallback latency;
  it does not make the classification correct.

## Risks and mitigations

- Keep the new API name and documentation explicit that the duration starts after contention.
- Share the existing after-busy loop so retry timing, inode replacement, and secure-file behavior do
  not drift.
- Keep typed `Busy` separate from `RunnerError`; Tokio polls the operation before its deadline when
  both are ready.
- Preserve one non-blocking `spawn_blocking` attempt at a time so cancellation cannot leave a thread
  blocked in `flock`.
- Use causal synchronization and virtual time in the regression test; add no sleep, retry, timeout
  increase, or production test hook.

## Acceptance criteria

- Deterministic lifecycle coverage proves an unlocked workspace-cache entry is not reported as
  `LockBusy` solely because the first blocking lock attempt is delayed.
- The focused regression test fails on the original end-to-end timeout behavior and passes on the
  fix.
- The bounded timer starts only after a real busy result and remains 50 ms.
- Immediate successful acquisition and the established post-busy retry cadence remain unchanged.
- Genuine long-held locks still fall back promptly, and short contention still acquires and reuses
  the workspace.
- Lock-file replacement hardening, cancellation safety, move-checkout lease ownership, and distinct
  filesystem/permission error handling remain intact.
- Focused lock, workspace-cache, and executor coverage plus the full runner all-target/all-feature
  suite pass under normal concurrency without sleeps, broad retries, increased timeouts, or skipped
  tests.

## Delivery notes

This is one runner-only Rust PR. It changes no API, queue payload, runner/guest protocol, database,
persisted cache format, CLI, migration, or deployment ordering. Old and new runners can overlap
without compatibility handling because lock files and `flock` ownership do not change.

## Relevant code and history

- `crates/runner/src/workspace_image_cache/lifecycle.rs`
- `crates/runner/src/workspace_image_cache/tests/lifecycle.rs`
- `crates/runner/src/lock.rs`
- `crates/runner/src/executor/tests/sandbox_run_tests/workspace_cache.rs`
- #25810 / PR #25812: bounded real entry-lock contention
- #25975 / PR #25992: deterministic transient-holder coverage

seven332:
## Summary

- start the workspace-cache lock budget only after a real `EWOULDBLOCK` result
- preserve the existing secure first attempt, 5/10/20 ms retry cadence, error propagation, and lock ownership
- add deterministic lifecycle coverage that delays an unlocked entry's first blocking-pool attempt past the old deadline

## Scope

This is runner-only. It changes no API, queue payload, runner/guest protocol, database, persisted cache format, CLI, migration, telemetry contract, or deployment ordering.

## Validation

- regression test confirmed red on the previous end-to-end timeout behavior and green after the fix
- `cargo test -p runner prepare_waits_for_initial_lock_attempt_before_contention_timeout`
- `cargo test -p runner --all-targets --all-features` (3128 passed; 20 existing ignored; guest inventory passed)
- `lefthook run pre-commit`
- `git diff --check`

Closes #27307

### DIFF
--- a/crates/runner/src/lock.rs
+++ b/crates/runner/src/lock.rs
@@ -100,23 +100,32 @@ enum LockAcquire {
     Busy,
 }
 
-async fn acquire_result_with(path: PathBuf, mode: LockMode) -> RunnerResult<LockAcquire> {
+async fn acquire_result_once(path: &Path, mode: LockMode) -> RunnerResult<LockAcquire> {
+    let attempt_path = path.to_path_buf();
+    tokio::task::spawn_blocking(move || acquire_result_blocking(&attempt_path, mode, |_| Ok(())))
+        .await
+        .map_err(|e| RunnerError::Internal(format!("lock task: {e}")))?
+}
+
+async fn acquire_after_busy(path: &Path, mode: LockMode) -> RunnerResult<Flock<File>> {
+    debug_assert!(mode.waits());
     let mut retry_delay = LOCK_RETRY_INITIAL_DELAY;
     loop {
-        let attempt_path = path.clone();
-        let result = tokio::task::spawn_blocking(move || {
-            acquire_result_blocking(&attempt_path, mode, |_| Ok(()))
-        })
-        .await
-        .map_err(|e| RunnerError::Internal(format!("lock task: {e}")))??;
-
-        match result {
-            LockAcquire::Busy if mode.waits() => {
-                tokio::time::sleep(retry_delay).await;
-                retry_delay = (retry_delay * 2).min(LOCK_RETRY_MAX_DELAY);
-            }
-            result => return Ok(result),
+        tokio::time::sleep(retry_delay).await;
+        retry_delay = (retry_delay * 2).min(LOCK_RETRY_MAX_DELAY);
+        match acquire_result_once(path, mode).await? {
+            LockAcquire::Acquired(lock) => return Ok(lock),
+            LockAcquire::Busy => {}
         }
+    }
+}
+
+async fn acquire_result_with(path: PathBuf, mode: LockMode) -> RunnerResult<LockAcquire> {
+    match acquire_result_once(&path, mode).await? {
+        LockAcquire::Busy if mode.waits() => acquire_after_busy(&path, mode)
+            .await
+            .map(LockAcquire::Acquired),
+        result => Ok(result),
     }
 }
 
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #27312

**Reviewed head:** `8de2c294802d67a9cb73d934680853d3b8ee5c14`

### Summary

The lock timeout now begins only after a real contention result, while preserving the existing
secure lock attempt, retry cadence, error propagation, and lease ownership behavior.

### Findings

No findings.

### Testing

The new lifecycle regression test deterministically covers blocking-pool delay before the first
lock attempt using a real temporary filesystem. Focused lock, workspace-cache, and executor tests
passed, as did formatting, Clippy, rustdoc, pre-commit hooks, and the full runner suite (3,128
passed; 20 existing ignored; guest inventory passed).

### Verdict

LGTM

PR #27338, opened 2026-08-15
fix(ci): terminate hanging cloudflare ssh operations

### LINKED ISSUE #27332: bug(ci): prevent Cloudflare SSH commands from hanging indefinitely
## Background

The `runner-behavior-lane-c` job in [Crates run 31863798087](https://github.com/vm0-ai/vm0/actions/runs/31863798087/job/94961911847?pr=27312) reached the `Test systemd reload coordination` step and remained running until the job-wide 35-minute timeout cancelled it.

Read-only inspection of `dev-11.gcp.vm3.ai` showed that the remote test did not hang:

- Both test units were installed, exercised, uninstalled, validated, and cleaned up between 04:16:08 and 04:16:11 UTC.
- The remote SSH server closed the session at 04:24:11 UTC because the client stopped responding.
- The GitHub Actions-side command nevertheless remained alive until the job timed out at 04:55:26 UTC.
- No test process, unit, file, or lock holder remained on the host afterward.

The Cloudflare SSH command wrapper probes the shared SSH transport before each operation, but the actual caller operation is then executed without an outer deadline. A shared `ControlMaster` or Cloudflare proxy failure after the probe can therefore leave the caller blocked indefinitely.

## Problem

CI SSH operations must not outlive a bounded execution window after their remote work has completed or their transport has become unresponsive. A transport failure must either be safely recovered or produce a timely non-zero result so the CI step can fail or retry instead of consuming the entire job timeout.

Blindly retrying arbitrary remote commands is not acceptable because a disconnected command may already have produced side effects. Retries must be limited to operations that are known not to have started remotely or are explicitly safe to repeat.

## Acceptance criteria

- The actual SSH caller operation, not only its preflight transport probe, has a documented bounded completion time.
- When the SSH or Cloudflare transport becomes unresponsive, the wrapper terminates the relevant local process tree and returns a non-zero status before the enclosing job timeout.
- Safe transient connection failures can recover without leaving a stale shared control connection.
- Potentially side-effecting remote commands are not blindly replayed after an ambiguous disconnect.
- Automated tests cover an SSH caller that never exits and verify bounded failure and cleanup.
- Existing successful SSH invocations continue to preserve their exit status and arguments.

## Relevant files

- `.github/actions/setup-ssh-tunnel/action.yml`
- `.github/scripts/cloudflare-ssh-command.sh`
- `.github/scripts/cloudflare-ssh-proxy.sh`
- `.github/scripts/runner-behavior-systemd-reload.sh`

## Security follow-up

The investigation also showed that a runner secret passed as a command-line argument is retained in the remote sudo journal. That credential exposure should be tracked and remediated separately so this reliability fix remains PR-sized.

seven332:
## Summary

- supervise each guarded SSH, SCP, and SFTP caller under a configurable 600-second GNU timeout
- keep the wrapper alive so GitHub cancellation terminates the caller process group, then force-kill any stuck ProxyCommand descendants after five seconds
- preserve the existing replay-safe preflight recovery while submitting every real remote operation exactly once
- cover invalid configuration, wall-time expiry, external cancellation, TERM-resistant descendants, streams, statuses, concurrency, and transport recovery

## Safety boundary

- ambiguous caller failures return non-zero and never replay a potentially side-effecting remote command or transfer
- unsupported options, caller-owned control sockets, non-VM3 targets, and remote-to-remote copies retain their existing pass-through behavior
- callers that intentionally need more than ten minutes can override `operation-timeout-seconds` on the composite action
- the separately observed runner-secret command-line exposure is intentionally excluded from this reliability fix

## Validation

- Bash syntax: 127 action/workflow scripts
- ShellCheck: complete Security-workflow file set; changed files also pass ShellCheck 0.11.0
- workflow script integration tests: 63/63
- `.github/scripts/check-workflow-shell-compatibility.sh`
- `actionlint` 1.7.12
- `./scripts/check-file-size.sh` for all changed files
- `git diff --check`
- pre-commit and commit-message hooks

Closes #27332

### DIFF
--- a/.github/actions/setup-ssh-tunnel/action.yml
+++ b/.github/actions/setup-ssh-tunnel/action.yml
@@ -12,6 +12,10 @@ inputs:
     description: "Fail setup unless every SSH host pre-connects"
     required: false
     default: "true"
+  operation-timeout-seconds:
+    description: "Maximum wall-clock seconds for one guarded SSH, SCP, or SFTP operation"
+    required: false
+    default: "600"
   ssh-key:
     description: "SSH private key for metal runner"
     required: true
@@ -117,6 +121,7 @@ runs:
     - name: Guard subsequent SSH operations
       shell: bash
       env:
+        OPERATION_TIMEOUT_SECONDS: ${{ inputs.operation-timeout-seconds }}
         SSH_USER: ${{ inputs.ssh-user }}
       run: |
         temp_root="${RUNNER_TEMP:-${TMPDIR:-/tmp}}"
@@ -141,5 +146,6 @@ runs:
           echo "VM0_CLOUDFLARE_SSH_STATE_DIR=${state_dir}"
           echo "VM0_CLOUDFLARE_SSH_USER=${SSH_USER}"
           echo "VM0_CLOUDFLARE_SSH_DEFAULT_CONTROL_PATH=$HOME/.ssh/vm0-ssh-%C"
+          echo "VM0_CLOUDFLARE_SSH_OPERATION_TIMEOUT_SECONDS=${OPERATION_TIMEOUT_SECONDS}"
         } >> "$GITHUB_ENV"
         echo "$wrapper_bin" >> "$GITHUB_PATH"
--- a/.github/scripts/cloudflare-ssh-command.sh
+++ b/.github/scripts/cloudflare-ssh-command.sh
@@ -396,6 +396,13 @@ PROBE_TIMEOUT_SECONDS=20
 PROBE_KILL_AFTER_SECONDS=5
 CONTROL_TIMEOUT_SECONDS=5
 CONTROL_KILL_AFTER_SECONDS=2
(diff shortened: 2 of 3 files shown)


seven332:
## Code Review: PR #27338

Reviewed commit `03a52f089e49ed2b1c22a18a6c6a82270f962cbb`.

### Summary

The change adds a bounded, process-group-aware timeout around the real Cloudflare SSH operation, while preserving the existing single-shot submission and cleanup behavior. It also adds integration coverage for operation timeouts and external termination, including descendants that ignore ordinary termination signals.

### Findings

No findings.

### Testing

- `.github/scripts/tests/cloudflare-ssh-command-test.sh` (63/63 tests passed)
- Shell syntax, ShellCheck, workflow shell compatibility, actionlint, file-size checks, `git diff --check`, and lefthook pre-commit checks passed.
- Live Cloudflare fault injection was not performed; the new failure modes are covered with deterministic local integration doubles.

### Verdict

LGTM

PR #27584, opened 2026-08-16
fix(runner): fail open on overlapping websocket prewarm correlation

### LINKED ISSUE #27523: security(python): fail open on overlapping WebSocket prewarm correlation
## Problem and evidence

OpenAI Responses WebSocket prewarm accounting does not preserve the ownership of unresolved client requests.

`crates/runner/mitm-addon/src/mitm_addon.py:1565-1588` passes every client and server message to the usage state machine without enforcing that only one `response.create` can be unresolved. In `crates/runner/mitm-addon/src/response_streaming.py:86-90`, the state stores only one `pending_response_created` boolean and one response ID. `observe_model_websocket_client_event()` at lines 143-155 overwrites that boolean for every client frame.

Consequently, when a normal generating `response.create` is followed by a `generate: false` request before the normal server `response.created` arrives, the boolean is true when the normal created event is processed. Lines 586-595 bind that normal response ID as the prewarm response. Lines 610-629 then treat the matching terminal usage as intentionally ignored and emit neither billable usage events nor model-usage observations.

This violates the accounting invariant that the prewarm exclusion may apply only to the response created for the exact non-generating request. Any ambiguous association must fail open for accounting.

The current tests in `crates/runner/mitm-addon/tests/test_model_provider_websocket_prewarm.py:50-90` cover only serialized request-created pairs. The malformed and per-flow cases do not cover two client create requests observed before the next server created event.

## Impact

A registered VM controls the ordering of client WebSocket frames. When client requests overlap the arrival of server lifecycle events, billable usage from a normal generated response can be classified as prewarm and omitted from both billing and model-usage observability. The failure is an underbilling boundary error; it does not require malformed JSON or a forged server response.

## Framing decision

This is a necessary accounting-integrity follow-up to [#26475](https://github.com/vm0-ai/vm0/issues/26475) and [PR #26496](https://github.com/vm0-ai/vm0/pull/26496), not a duplicate of either. The earlier work correctly excluded an unambiguously identified prewarm source; this issue covers the unresolved-request case that the one-request state cannot represent.

The prior issue provides concrete prewarm volume evidence (82 input-only WebSocket responses across 77 runs in the inspected window), but no direct production count for overlapping unresolved requests exists yet. OpenAI documents that `generate:false` is a valid warmup and that multiple requests on one connection are processed sequentially ([WebSocket Mode](https://developers.openai.com/api/docs/guides/websocket-mode#connection-behavior-and-limits)); the official Codex client also normally waits for warmup completion. That lowers the expected frequency for the official client, but does not make the proxy state safe for queued, custom, or otherwise untrusted clients that can send multiple frames before the corresponding server lifecycle event.

The issue is therefore suitable for planning as a defensive correctness fix, with rollout telemetry used to measure the ambiguous case rather than treating the existing prewarm totals as overlap frequency.

## Decision

Implement the fix as one runner-only PR using a bounded, flow-scoped pending/active request-intent state with a sticky ambiguity transition. A second unresolved request, unknown or malformed client event, invalid lifecycle identity, provider error, or correlation-cap condition clears prewarm exclusion and leaves normal usage reporting enabled until WebSocket cleanup.

Only a single unambiguous `generate:false` request bound to a valid `response.created` ID may suppress matching terminal usage. A previously ignored response ID may remain only for duplicate-terminal idempotency and cannot suppress usage after the flow becomes ambiguous.

This direction preserves serialized prewarm savings while making the accounting invariant explicit: uncertainty reports usage rather than suppressing it. It keeps state constant-sized, retains the existing response-ID/source boundary, and does not alter WebSocket transport behavior or require a provider request ID that the protocol does not expose.

## Alternatives considered

- A bounded FIFO could preserve more overlapping sequences, but lifecycle gaps, provider errors, or rejected requests can shift a request-ID-free queue and cause false suppression.
- Proxy-side serialization would avoid overlap by changing client-visible WebSocket timing and adding buffering/backpressure; that is outside the accounting problem and has a larger compatibility surface.

## Scope and constraints

- Keep the fix in the runner's OpenAI Responses WebSocket accounting path.
- Preserve serialized prewarm exclusion, normal input-only reporting, source diagnostics, idempotency, flow isolation, and terminal/error cleanup.
- Treat ambiguous, unknown, malformed, or unbounded correlation as fail-open: report extractable usage rather than suppressing it.
- Keep correlation state bounded and content-free; do not materialize prompts or tool schemas solely for correlation.
- This is a forward fix. It does not perform a blanket historical refund or introduce a database/API migration because persisted usage does not retain the `generate` intent.

## Risks and mitigations

- **Ambiguous prewarm overbilling:** intentional fail-open behavior; serialized regression tests and one bounded ambiguity diagnostic per flow make the trade-off observable.
- **Stale state after lifecycle gaps:** sticky ambiguity on unknown/error/invalid/overflow boundaries and exact-ID checks for retained duplicate terminals.
- **Long-lived flow resource growth:** fixed fields and one retained duplicate ID; no unbounded queue or body retention.
- **Parser/privacy risk:** reuse selective extraction and work limits; diagnostics contain no request or tool content.

## Acceptance criteria

- Preserve enough bounded per-flow request intent to correlate each `response.created` with the correct preceding client request, or mark overlapping unresolved requests ambiguous and fail open.
- Never grant the prewarm usage exclusion to a response when another unresolved client request could own the created event.
- On malformed/unknown client frames, server errors, missing or conflicting lifecycle IDs, or correlation-cap overflow, normal usage reporting remains enabled.
- Keep unresolved correlation state bounded and release it on terminal WebSocket and error cleanup.
- Add public-hook integration coverage for normal-then-prewarm and prewarm-then-normal client requests before any `response.created` event, plus overflow and lifecycle-error cases.
- Verify that normal response usage is still reported in every ambiguous sequence and that prewarm usage is excluded only when its association is unambiguous.
- Preserve the existing serialized prewarm and normal-request behavior, including duplicate-terminal idempotency and per-flow isolation.
- Add a bounded, content-free diagnostic or counter for ambiguous correlation so post-rollout frequency can be measured without exposing request data.

## Delivery notes

- This issue is a standalone, independently reviewable implementation slice; no parent or child issue is required.
- The implementation plan and validation commands are recorded in the Plan Phase comment for `issue-27523-websocket-prewarm-correlation`.
- Rollout reconciliation should wait for old runners to drain; historical rows cannot be deterministically reclassified.

seven332:
## Summary

- Replace the last-write-wins WebSocket prewarm flag with bounded pending/active request-intent correlation.
- Fail open to normal usage reporting whenever request ownership is ambiguous, including overlapping creates, malformed or unknown client frames, invalid lifecycle identities, provider errors, client or server parser work-limit boundaries, and response-ID reuse with or without a created boundary.
- Clear retained suppression IDs on sticky ambiguity while preserving duplicate-terminal idempotency during otherwise unambiguous later requests.
- Preserve exact serialized `generate:false` exclusion, flow cleanup, and content-free one-per-flow ambiguity diagnostics, including `correlation_cap` classification for parser work limits.
- Add public-hook integration coverage for overlap orderings, active overlap, unknown/error/cap paths, retained-ID ambiguity and reuse, lifecycle gaps, and the existing serialized behavior.

## Scope

- Runner-only OpenAI Responses WebSocket accounting change.
- No WebSocket transport serialization, API or database schema changes, migrations, historical refund logic, or feature switch.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (5455 passed)

Closes #27523

### DIFF
--- a/crates/runner/mitm-addon/src/response_streaming.py
+++ b/crates/runner/mitm-addon/src/response_streaming.py
@@ -20,7 +20,7 @@ Lifecycle:
 
 from collections.abc import Callable
 from dataclasses import dataclass
-from typing import NamedTuple
+from typing import Literal, NamedTuple
 
 from mitmproxy import http
 from wsproto.utilities import generate_accept_token
@@ -85,9 +85,61 @@ class _ResponseUsageStreamSetup(NamedTuple):
 
 @dataclass(slots=True)
 class _OpenAIResponsesPrewarmState:
-    pending_response_created: bool = False
-    response_id: str | None = None
-    diagnostic_emitted: bool = False
+    pending_intent: Literal["normal", "prewarm"] | None = None
+    active_intent: Literal["normal", "prewarm"] | None = None
+    active_response_id: str | None = None
+    ignored_response_id: str | None = None
+    ambiguous: bool = False
+    ambiguity_diagnostic_emitted: bool = False
+    ignored_diagnostic_emitted: bool = False
+
+
+_WebSocketCorrelationReason = Literal[
+    "overlapping_request",
+    "unknown_client_event",
+    "invalid_lifecycle",
+    "server_error",
+    "correlation_cap",
+]
+
+
+def _clear_websocket_correlation_candidate(state: _OpenAIResponsesPrewarmState) -> None:
+    state.pending_intent = None
+    state.active_intent = None
+    state.active_response_id = None
+
+
+def _lifecycle_ambiguity_reason(
+    lifecycle: usage.OpenAIResponsesServerLifecycle,
+) -> _WebSocketCorrelationReason:
+    if lifecycle.inspection_error == "work limit exceeded":
+        return "correlation_cap"
+    return "invalid_lifecycle"
+
+
+def _mark_websocket_correlation_ambiguous(
+    flow: http.HTTPFlow,
+    state: _OpenAIResponsesPrewarmState,
+    reason: _WebSocketCorrelationReason,
+) -> None:
+    """Disable prewarm exclusion after ownership can no longer be proven."""
+    state.ambiguous = True
+    _clear_websocket_correlation_candidate(state)
+    if state.ambiguity_diagnostic_emitted:
+        return
+    log_proxy_entry(
+        flow_metadata.proxy_log_path(flow.metadata),
+        "warn",
+        "Model provider WebSocket usage correlation became ambiguous",
+        type="model_usage_correlation",
+        disposition="ambiguous",
+        reason=reason,
+        run_id=flow_metadata.run_id(flow.metadata),
+        flow_id=flow.id,
+        transport="websocket",
+        firewall_name=flow_metadata.firewall_name(flow.metadata),
+    )
+    state.ambiguity_diagnostic_emitted = True
 
 
 def model_usage_protocol(flow: http.HTTPFlow) -> usage.ModelUsageProtocol:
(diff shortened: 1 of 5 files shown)


seven332:
## Code Review: PR #27584

**Reviewed head:** `72d2c8e2fcbeccde33cf145da640ca23756922e7`

### Summary

The runner now keeps bounded pending/active request intent and fails open whenever a WebSocket prewarm response cannot be associated unambiguously. Exact serialized `generate:false` exclusion, duplicate-terminal handling, cleanup, and privacy-safe diagnostics remain intact. The lifecycle inspection optimization avoids unnecessary parsing on ordinary delta frames.

### Findings

No findings.

### Testing

The changed behavior is covered through public WebSocket hooks, including both overlap orders, active overlap, malformed/unknown client frames, invalid IDs, provider errors, cap handling, cleanup, and late `type` classification. Local validation passed: `uv lock --check`, `uv sync --locked`, Ruff format/check, BasedPyright, and `uv run --no-sync python -m pytest tests/` (`5450 passed`, 59 existing warnings). Remote CI was still running when this review was posted.

### Verdict

LGTM

PR #27068, opened 2026-08-14
feat(api): enforce the run video model at the generation endpoint

bingjiez666:
## Summary

- Gate run-level video model enforcement behind `VideoModelSelection`.
- Preserve the existing endpoint and template-prompt behavior completely while the switch is disabled.
- When the switch is enabled, enforce the non-null model snapshotted on `zero_runs` for agent and sandbox generation requests, and stop prompting the agent to pass the template model.

### Disabled: completely unchanged behavior

- The generation endpoint uses the request-body `model`, or `DEFAULT_VIDEO_MODEL` when the body omits it, even if the run already has a non-null snapshot.
- Video template prompts continue to include `--model <alias>`.
- Aspect ratio, duration, resolution, and audio keep their existing behavior.

### Enabled: run model enforcement

- For Zero/Sandbox callers with a `runId` and a non-null `zero_runs.selected_video_model`, the endpoint replaces the request-body model with the run snapshot.
- A `null` run snapshot is treated as no pin and preserves the request/default behavior.
- Session and PAT callers without a `runId` preserve the request/default behavior.
- Video template prompts omit only `--model`; aspect ratio, duration, resolution, and audio flags remain.

Endpoint enforcement intentionally ignores request-body model overrides whenever an enabled run has a non-null pin. Therefore conversational requests such as “use Kling for this one” and model values persisted on video templates no longer override that run's model. The existing response `model` field reports the effective model, so no new reporting path is added.

The rejected alternative is an authorization list whose first entry is the effective default while any listed model remains selectable. Issue #26765 explicitly chose a single enforced run snapshot, so this PR does not implement template-level or model-list overrides.

## Merge order

Satisfied: #26910 (PR6) merged into `main` before this PR was updated. The video-model picker replacement entry point is therefore already available.

PR8 will remove the remaining template model UI after this PR; that work is intentionally not included here.

## Verification

- Merged current `main` through `f518a1f1e0` and resolved the built-in-generation naming conflict.
- Real-PostgreSQL route-boundary tests cover: enabled pin overriding the body and being reported; disabled switch preserving the body despite a non-null pin; enabled switch with a `null` pin preserving the body; and enabled switch for a caller without `runId` preserving the body.
- Prompt and chat-run tests cover disabled prompts retaining `--model` and enabled prompts removing only that flag.
- `pnpm format`
- changed-file `prettier --write` and `prettier --check`
- `pnpm knip`
[...]

PR #27610, opened 2026-08-16
fix(api): reconcile request parameters with the enforced run video model

bingjiez666:
Fixes a live failure found in a real chat: the user picked MiniMax in the picker, asked for a puppy video, and the run failed with

```
Unsupported video resolution for minimax-h3: 720p
```

after the agent ran `okou generate video --model dreamina-seedance-2.0-fast --resolution 720p`. The agent then told the user "the command specified seedance", which is true and useless — nothing in the request named MiniMax.

## Root cause

Enforcement (#27068) replaces the caller's model with the run's pin, but kept every other parameter:

```ts
{ ...bodyResult.data, model: runVideoModel }
```

The caller sizes aspect ratio, duration, and resolution *for the model it asked for*. Swapping the model underneath makes those values valid for the request and invalid for the pin. 720p is legal for `dreamina-seedance-2.0-fast` and illegal for `MiniMax-H3` (768p / 2k), so the request died on a model the caller never sent.

## Fix

`withEnforcedRunVideoModel()` swaps the model and drops any aspect ratio, duration, or resolution the pin cannot honour, so the pinned model's own default applies instead of failing. The effective values already come back on the response, so the caller can report what was actually used.

Gated: `loadEnforcedRunVideoModel` returns null while `VideoModelSelection` is off, so this path never runs with the switch disabled.

## Deliberately not done

**No system-prompt injection.** Every run now resolves to some video model, but almost no run generates video — a prompt section announcing the model would spend attention on all of them to help a few. The response already reports the effective model, which is enough for the agent to describe what happened after the fact, and this fallback means it no longer has to guess correctly beforehand.

**No CLI change here.** The CLI hardcodes its own `--model` default (`dreamina-seedance-2.0-fast`), which is the reason the agent sent seedance even when it passed no model at all. That is a second source of truth worth removing, but the CLI cannot read feature switches, so removing it cannot be gated the way this fix is. It is behavior-preserving only because the two defaults happen to coincide — an argument, not a switch. It belongs in its own PR where that risk is reviewed on its own terms. This fix does not depend on it: the test sends exactly what today's CLI sends.

## Worth a reviewer's opinion

Dropping an unhonourable resolution falls back to `config.defaultResolution`, matching `resolveVideoGenerationOptions()`. For a MiniMax pin that turns a 720p request into **2k, which costs more than the 768p the caller was closer to wanting**. Reusing the established fallback beat inventing a nearest-match rule, but if silent cost increases are unacceptable here, the alternative is rejecting with an error that names the substitution.

## Testing

New case reproducing the production failure: request `dreamina-seedance-2.0-fast` at 720p against a Kling 4k-only pin. Verified it fails without the fix and passes with it.

The pre-existing enforcement test happened to send `resolution: "4k"` against a 4k-only pin, so every parameter was compatible by luck — which is why this shipped.

- `video-io-generate.test.ts`: 29 passed
- `check-types` and `lint` clean

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/video-io-generate.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/video-io-generate.test.ts
@@ -760,6 +760,96 @@ describe("POST /api/zero/video-io/generate", () => {
     });
   });
 
+  it("drops a request parameter the pinned model cannot honour", async () => {
+    // Reproduces the production failure: the caller sized its parameters for
+    // the model it asked for, enforcement swapped the model, and the request
+    // died on `Unsupported video resolution for minimax-h3: 720p` — an error
+    // naming a model the caller never sent. 720p is valid for
+    // dreamina-seedance-2.0-fast and invalid for the Kling pin.
+    const fixture = await seedVideoFixture();
+    const { composeId } = await store.set(
+      seedCompose$,
+      { orgId: fixture.orgId, userId: fixture.userId },
+      context.signal,
+    );
+    const { runId } = await store.set(
+      seedRun$,
+      {
+        orgId: fixture.orgId,
+        userId: fixture.userId,
+        composeId,
+        triggerSource: "web",
+      },
+      context.signal,
+    );
+    await setRunVideoModelFixture({
+      runId,
+      selectedVideoModel: KLING_V3_4K_MODEL,
+    });
+    await updateFeatureSwitchesForUser(context, fixture, {
+      [FeatureSwitchKey.VideoModelSelection]: true,
+    });
+
+    let observedRequestUrl: string | null = null;
+    server.use(
+      http.post(KLING_V3_4K_QUEUE_URL, ({ request }) => {
+        observedRequestUrl = request.url;
+        return HttpResponse.json({
+          request_id: "reconciled-kling-request",
+          status_url: KLING_STATUS_URL,
+          response_url: KLING_RESPONSE_URL,
+        });
+      }),
+      http.get(KLING_VIDEO_URL, () => {
+        return new HttpResponse(VIDEO_BYTES, {
+          headers: { "content-type": "video/mp4" },
+        });
+      }),
+    );
+
+    const token = zeroToken({
+      userId: fixture.userId,
+      orgId: fixture.orgId,
+      runId,
+    });
+    const app = createVideoIoTestApp(fixture.pricingResolution);
+    const response = await app.request("/api/zero/video-io/generate", {
+      method: "POST",
+      headers: { authorization: `Bearer ${token}` },
+      body: JSON.stringify({
+        prompt: "a fluffy golden retriever puppy",
+        model: "dreamina-seedance-2.0-fast",
+        resolution: "720p",
+      }),
+    });
+
+    expect(response.status).toBe(202);
+    const generationId = readAcceptedGenerationId(
+      await response.json(),
+      "video",
+      fixture.userId,
+    );
+    await postFalWebhook(app, observedRequestUrl, {
+      video: {
+        url: KLING_VIDEO_URL,
+        content_type: "video/mp4",
+      },
+    });
+    await flushWaitUntilForTest();
+
+    const statusResponse = await app.request(
+      `/api/zero/built-in-generations/${generationId}`,
+      { headers: { authorization: `Bearer ${token}` } },
+    );
+    expect(statusResponse.status).toBe(200);
+    // The pinned model's own resolution, not the one the caller sized for the
+    // model it named, and reported back so the caller can say what was used.
+    expect(readGenerationResult(await statusResponse.json())).toMatchObject({
+      model: KLING_V3_4K_MODEL,
+      resolution: "4k",
+    });
+  });
+
   it("keeps the request model when video model selection is disabled", async () => {
     const fixture = await seedVideoFixture();
     const { composeId } = await store.set(
--- a/turbo/apps/api/src/signals/routes/video-io-generate.ts
+++ b/turbo/apps/api/src/signals/routes/video-io-generate.ts
@@ -1,9 +1,13 @@
 import { randomUUID } from "node:crypto";
 
 import { command } from "ccstate";
-import { zeroVideoIoGenerateContract } from "@okouai/api-contracts/contracts/zero-video-io-generate";
+import {
+  zeroVideoIoGenerateContract,
+  type ZeroVideoIoGenerateRequest,
+} from "@okouai/api-contracts/contracts/zero-video-io-generate";
 import type { BuiltInGenerationRealtimeSubscription } from "@okouai/api-contracts/contracts/built-in-generation";
 import { isFeatureEnabled } from "@okouai/core/feature-switch";
+import { VIDEO_MODEL_CONFIGS } from "@okouai/core/video-model-catalog";
 import { FeatureSwitchKey } from "@okouai/core/feature-switch-key";
 import {
   isVideoModelId,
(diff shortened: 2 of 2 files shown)


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: db8c0a9d800d1ed9d2ee2be7e9e4be75190adcf2
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

`withEnforcedRunVideoModel()` swaps in the run's pinned video model and drops any `aspectRatio` / `duration` / `resolution` the pin cannot honour, so the pinned model's own default applies instead of rejecting the request with an error naming a model the caller never sent. The path is reached only when `loadEnforcedRunVideoModel` returns a model, which requires `FeatureSwitchKey.VideoModelSelection`. The persisted job record is built from the reconciled `options` (`videoRequestRecord(options)`), so the webhook re-parse in `webhooks-built-in-generations.ts` sees the same values the submit path used — no divergence between charge, provider request, and reported result.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness: no unhandled edge cases or wrong assumptions on the changed path | Pass | `video-io-generate.ts:428` — reconciled body flows into `parseVideoOptions`, which re-validates every field against the pin; the persisted record is derived from `options`, not the raw body. |
| `REVIEW.md` | Security: no new auth surface, injection, or secret exposure | Pass | No auth, SQL, or credential changes; `runVideoModel` comes from a `VideoModelId`-validated column and indexes a static catalog. |
| `REVIEW.md` | Type safety: no `any`, no overly broad types | Pass | No `any`. One `value as T` cast inside `honours()` — see note 3. |
| `REVIEW.md` | Feature regression guard: existing behavior changed only under a gate, with coverage | Pass | Behavior changes only when the pin is present; `keeps the request model when video model selection is disabled` and `enforces the run video model over the request model and reports it` still cover the pre-existing paths. |
| `REVIEW.md` | DB/JSONB gate: persisted contract changes need migration/normalizer/rationale | N/A | No schema, contract, or `jsonb().$type<>()` change; `videoRequestRecord` output shape is unchanged. |
| `docs/bad-smell.md` | Zero `any` types and zero lint/type suppressions | Pass | Neither appears in the diff. |
| `docs/bad-smell.md` | No dynamic `import()`; static imports only | Pass | `VIDEO_MODEL_CONFIGS` and `ZeroVideoIoGenerateRequest` are static top-of-file imports. |
| `docs/bad-smell.md` | No hardcoded URLs or environment-specific values | Pass | Model IDs come from the shared catalog; no URLs added outside test fixtures. |
| `docs/bad-smell.md` | No duplicated logic that should be shared | Finding (P1) | `withEnforcedRunVideoModel` re-implements the drop-unhonourable-value semantics that `resolveVideoGenerationOptions()` already has in `@okouai/core/video-model-catalog`, and hand-rolls `honours()` where `video-generation.service.ts` already has `hasString`. See note 3. |
| `docs/bad-smell.md` | Comments explain WHY, not WHAT | Pass | The `withEnforcedRunVideoModel` docblock records the production failure and the reason for dropping rather than failing. |
| `docs/fallback.md` | Does the diff introduce a fallback needing a `Fallbacks` section? | N/A | Dropping an unhonourable optional field so the model default applies is normal request resolution over a contract-optional field (`zeroVideoIoGenerateRequestSchema` marks all three `.optional()`), not a cross-version compatibility branch, a `??` on a contract-guaranteed field, or a retired-protocol reader. It is permanent product behavior, not time-boxed, so no `Fallbacks` section is required. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | The added test asserts a live success path, not the absence of deleted code. |
| `docs/fallback.md` | Feature behind a non-GA switch needs no compatibility code | Pass | The new path is inside `VideoModelSelection` and adds no dual-read/dual-write or migration. |
| `docs/deployment-compatibility.md` | Old clients must keep working across the backend rollout | Pass | Server-side only; request and response contracts are unchanged. Old CLI/sandbox instances (up to 2h) send the same body and now get a 202 where they previously got a 400 — strictly wider acceptance, no old/new incompatibility in either direction. |
| `docs/testing.md` | `fix:` commit carries a regression test | Pass | `drops a request parameter the pinned model cannot honour` reproduces the production failure (seedance @ 720p against a 4k-only pin) and the author verified it fails without the fix. |
| `docs/testing.md` | Integration tests only, no unit tests for internal functions | Pass | `withEnforcedRunVideoModel` is exercised through the endpoint, not imported directly. |
| `docs/testing.md` | Mock at boundary only; no `vi.mock("../...")`; real DB; MSW for HTTP; no fake timers | Pass | The test adds only MSW handlers (`http.post(KLING_V3_4K_QUEUE_URL)`, `http.get(KLING_VIDEO_URL)`), uses the real DB fixtures, and adds no timer or fetch stubbing. |
| `docs/testing.md` | Test behavior, not mocks | Pass | Asserts the generation result (`model`, `resolution`) read back from `/api/zero/built-in-generations/:id`, not mock call counts. |
| `docs/testing/api-testing.md` | Route behavior covered through the real app and its contract | Pass | Drives `app.request("/api/zero/video-io/generate")` from `createVideoIoTestApp`, then the status endpoint, matching the surrounding tests in the file. |
| `docs/testing/testing-external-behavior.md` | Construct and verify through the external endpoint, not internals | Pass | All assertions go through public endpoints; setup reuses the same `seedRun$` / `setRunVideoModelFixture` seeding as the adjacent enforcement tests, so the PR introduces no new internal-state coupling. |
| `docs/testing/testing-external-behavior.md` | New reachable branches are covered from the endpoint | Finding (P1) | Only the `resolution` branch is exercised; the `aspectRatio` and `duration` drops added in the same function have no coverage. See note 2. |
| `.claude/skills/ccstate/SKILL.md` | ccstate usage in `apps/api` stays within command/computed conventions | Pass | `withEnforcedRunVideoModel` is a pure helper called inside the existing `postVideoInner$` command; no new signal, no manual cache, no `void`-swallowed promise. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

1. `turbo/apps/api/src/signals/routes/__tests__/video-io-generate.test.ts` — the new test covers only the `resolution` branch of `withEnforcedRunVideoModel`; the `aspectRatio` and `duration` drops ship untested even though both are as reachable as the tested one (Seedance accepts all six aspect ratios and up to 15s/30s, Kling v3 4K accepts only `16:9`/`9:16`, Veo only `16:9`/`9:16` and `4s`/`6s`/`8s`). This needs no extra test — widening the existing request covers all three at once, e.g. send `aspectRatio: "4:3"`, `duration: "20s"`, `resolution: "720p"` against the Kling pin and assert `{ aspectRatio: "16:9", duration: "8s", resolution: "4k" }`. That is also a stronger reproduction than the current single-field version. (`docs/testing/testing-external-behavior.md`, `docs/testing.md`)
2. `video-io-generate.ts:withEnforcedRunVideoModel` — the drop-unhonourable-value rule now exists twice: here, and in `resolveVideoGenerationOptions()` in `@okouai/core/video-model-catalog`, whose docblock describes the identical semantic ("Stored options ... can also name a value the chosen model does not accept, which is what happens after switching model; those fall back the same way the service falls back"). The local `honours()` also duplicates `hasString()` in `video-generation.service.ts`. Not a blocker, but the two copies can now drift, and the reason they are separate (typed catalog values vs. raw request strings) is only mechanical. Consider exporting a shared `dropUnhonouredVideoParams(config, raw)` from the catalog package so the picker path and the enforcement path stay in step. (`docs/bad-smell.md` §"Duplicated logic that should be shared")

#### Notes (non-blocking)

3. `honours()` casts `value as T` to compare a raw request string against the catalog's typed union. `(supported as readonly string[]).includes(value)` — the shape `hasString` already uses — expresses the same check without asserting an unvalidated string into a member type.
4. Reconciliation covers three of the model-dependent parameters. The same "error names a model the caller never sent" class still exists for others that `parseVideoOptions` / `validateVideoReferences` reject per model: `generateAudio: false` against a MiniMax pin → `MiniMax H3 always generates native audio`; `imageUrls` / `videoUrls` / `audioUrls` / `firstFrameImageUrl` / `lastFrameImageUrl` against a Kling or Veo pin → `Reference images are not supported for kling-v3-4k`; and a >2000-char prompt against a MiniMax pin. I do **not** think those should be dropped the way the three sizing parameters are — silently discarding a reference image would produce a video that ignores the user's input, which is worse than a clear rejection, and unlike the reported bug these errors name a model the user did in fact pick in the composer. Worth stating that boundary explicitly (a follow-up issue, or a line in the docblock) so the next reader does not read the omission as an oversight.
5. Dropped `aspectRatio` and `duration` fall back to the **global** `DEFAULT_VIDEO_ASPECT_RATIO` (`16:9`) and `DEFAULT_VIDEO_DURATION` (`8s`), not to a per-model default — only `resolution` uses `modelConfig.defaultResolution`. This is safe today only because every entry in `VIDEO_MODEL_CONFIGS` happens to include `16:9` and `8s`. A future model without one of them reintroduces exactly the failure this PR fixes, and it would be invisible until a run hit it — the same "compatible by luck" property the PR body correctly calls out about the pre-existing enforcement test. A type-level or startup invariant over the catalog would close it permanently.

### On the question you raised for a reviewer

Keep the fallback to `config.defaultResolution`; do not switch to an error.

- It is the semantic `resolveVideoGenerationOptions()` already applies when the composer switches models, so erroring here would make the two paths disagree about the same situation.
- The resulting cost is not hidden from the system: pricing is resolved from the reconciled `options` (`getMissingVideoPricing(pricing, options)`), and the effective `resolution` and `creditsCharged` both come back on the generation result, so the agent can state what was actually used and billed.
- The upgrade is to a tier of the model **the user selected in the picker**, not to an arbitrary one.

If the cost jump is still unacceptable, the narrower change is picking the pin's cheapest supported resolution rather than rejecting — but that invents a ranking rule the catalog does not currently express, so I would not add it in this PR.

### Testing

- Coverage: Adequate for the verdict — `fix:` commit ships a reproduction test that the author verified fails without the fix. Incomplete on new branches: `aspectRatio` and `duration` drops are uncovered (P1 finding 1, fixable by widening the existing request rather than adding tests).
- Conventions: Compliant. MSW at the HTTP boundary, real database, no `vi.mock()` on relative paths, no fake timers, assertions on endpoint responses rather than mock invocations, and the new case mirrors the structure of the adjacent enforcement tests.
- CI: all published checks green on `db8c0a9`.

PR #27607, opened 2026-08-16
feat(app): add desktop video model switch

bingjiez666:
## Summary

- show the feature-gated desktop model control as `Chat · <model>` with a dedicated video button
- open the existing video-model panel directly from that desktop button while keeping the nested `Manage more models` entry on mobile
- preserve the switch-disabled composer and all existing thread/default video-model persistence semantics

## Verification

- `pnpm exec prettier --check apps/platform/src/views/zero-page/components/model-provider-picker.tsx apps/platform/src/views/zero-page/zero-chat-composer.tsx apps/platform/src/views/zero-page/__tests__/chat-composer-model-selection-and-providers.test.tsx`
- `pnpm --filter @okouai/app run lint`
- `pnpm --filter @okouai/app run check-types`
- `pnpm knip --workspace @okouai/app`
- `pnpm -F @okouai/app exec vitest run src/views/zero-page/__tests__/chat-composer-model-selection-and-providers.test.tsx` (59 passed)
- Browser verification not run locally per repository instruction; the PR preview is the visual verification surface

Refs #26765

PR #27180, opened 2026-08-14
refactor(app): remove the model control from video template options

bingjiez666:
Part of #26765 (PR8 of 8).

## What changes

- With `VideoModelSelection` enabled, the run picker is the only visible model control: the legacy template picker row and options row are hidden, and composer/sent chips and details omit the model label.
- Ratio, duration, resolution, and audio remain editable per generation.
- Template model data keeps its legacy semantics regardless of the switch: new templates still sparsely persist a non-default `videoOptions.model`, existing values still participate in option resolution, and edits use the same sparse write behavior.
- Persisted `videoOptions.model` values are not migrated or deleted.
- With `VideoModelSelection` disabled, behavior is completely unchanged: the old template model picker, model display, chip/detail text, and model persistence remain intact.

## Dependency

Satisfied: #27068 is merged into `main`, and this branch includes the latest synchronized `main`. PR6 (#26910) is also merged, so the replacement video model picker is already present before this control is removed.

## Verification

- Prettier write/check passed for all changed files.
- Full Turbo lint passed with single-package concurrency: 13/13 tasks.
- Knip passed.
- Non-API/non-Platform workspace type checks passed: 12/12 tasks.
- `@okouai/app` production and test type checks passed.
- API type checks passed.
- `chat-composer-template-picker.test.tsx`: 32 passed.
- `chat-user-messages.test.tsx`: 10 passed.

PR #27616, opened 2026-08-16
feat(app): move video parameters from the template chip to the composer

bingjiez666:
Stacked on #27607 — **base is `feat/desktop-video-model-switch`, not `main`.** If #27607 merges first, retarget this to `main` and merge `main` in.

## What changed

Text-to-video parameters move off the template chip and onto the chat input.

**Composer chip.** Once the desktop composer is switched to video mode (the control #27607 added), a chip appears in the input's bottom row reading `16:9 · 8s · 720p`. Opening it gives the four controls that already existed in the template popover, reused verbatim: the ratio icon grid, resolution chips, the duration slider bounded by the model's own range, and the generate-audio switch.

**The panel is derived from the effective video model, not hardcoded.** The model resolves through the same three steps the API uses — thread pin (`ThreadMeta.selectedVideoModel`) → member default (`userModelPreference$.selectedVideoModel`) → `DEFAULT_VIDEO_MODEL` — and the offered values come from `VIDEO_MODEL_CONFIGS[model]`. Only the values the user actually moved are stored, so switching Seedance → MiniMax-H3 re-resolves the rest: the resolution choices become `768p / 2k` and a stale `720p` falls back to that model's default.

This also closes #27610 at the source: the panel cannot produce a combination the generation endpoint would reject with a 400, because it never offers one.

**Parameters travel as run options, not as state.** They ride the existing `runOptions` channel on the send — the same one `codexServiceTier` already uses — as a new optional `runOptions.video`. The API renders them into that run's appended system prompt next to the generation-template block, and forgets them. No new column, no `chat_threads` write, no field on the persisted `userMessage`.

**The block reads as defaults, not as requirements.** The chip is set before the message is written, so the message is the later and more specific statement of intent — asking for a square video has to win over a 16:9 chip rather than fight it. Overrides land per parameter, so a message naming only a ratio leaves the chosen duration and resolution in force:

```
# Video Generation Defaults
The user set these for videos generated in this run:
- Aspect ratio: 9:16
- Duration: 6s
- Resolution: 480p
- Audio: off
Where this run's message asks for something else, the message wins, for that parameter only.
```

Values only, never the `--aspect-ratio 9:16 …` flag string they map to. A pre-assembled command is a ready-made answer that stops being correct the moment the message overrides one value — the same pull toward treating these as requirements that the wording above works to avoid.

The block is deliberately absent unless the user changed something: most runs never generate a video, and restating a model's own defaults in every run's prompt would cost every unrelated run context for nothing.

**Templates no longer carry parameters.** The template picker's video-model header row and the inline chip's settings popover are both removed, so nothing in the UI writes `selection.videoOptions` any more. Messages already holding it are untouched — no backfill, no migration — and still render their own spec; a template with nothing stored now renders no spec instead of advertising catalog defaults it never agreed to.

Same feature switch as the video button: `FeatureSwitchKey.VideoModelSelection`. With it off, nothing here appears and `runOptions.video` is ignored server-side.

## Known consequence: the settings do not survive a reload

Because nothing is persisted, refreshing the page returns the chip to the effective model's defaults, and a **queued** message dispatched later carries no parameters either — the queued path replays from the persisted chat event, which by design has none.

That is the trade this makes on purpose: no schema change, no migration, and no fourth layer of persisted state alongside the thread pin, the member default and the system default. Calling it out rather than burying it.

## Relationship to #27180

`main` landed #27180 while this was in review, and the base has since merged it in. That PR took the first step — with `VideoModelSelection` **on**, the template picker's model row and the popover's model line are hidden, and the template keeps only ratio/duration/resolution/audio — while deliberately preserving the pre-switch behaviour when the switch is **off**.

This PR finishes the move and drops that compatibility path from the composer entirely: the template settings popover, its inline chip zone, and the picker's model row are gone in both switch states, so nothing in the composer writes `selection.videoOptions` any more. Worth a reviewer's eye: **with `VideoModelSelection` off, a video template now carries no parameters at all**, where before it carried the template-owned set. `legacyVideoTemplateSpecText` is untouched — reading a *sent* message still renders exactly what #27180 made it render.

## Also worth noting

- Mobile keeps the nested `Manage more models` entry for pinning a video model but gets no parameter chip; the chip follows the desktop-only mode control.
- `updateTemplateAt$` lost its only caller with the in-place chip edit and is deleted.
- `chat.templates.videoModelLabel` is now unused; `i18next-cli` removed it from all 10 locales.

## Verification

- `pnpm turbo run lint` (all workspaces), `pnpm knip`, `pnpm -F api check-types`, `pnpm -F @okouai/app check-types` — green
- New platform tests: the chip sends only what the user changed as `runOptions.video`, and a MiniMax-pinned thread offers `768p / 2k` rather than `720p`
- New API BDD assertions in `chat-events.bdd.test.ts`: `runOptions.video` reaches the run's `appendSystemPrompt` as `# Video Generation Defaults`, worded so the message outranks it and carrying no generation flags, and a send that set nothing carries no trace of the block
- Rewrote the template-picker test that used to drive the removed popover; it now asserts a video template is sent with `selection` holding only `stylePresetId`
- Ran locally: `chat-composer-*`, `chat-user-message*`, `chat-run-queue`, `chat-draft`, `chat-localization` (192 tests), and the API `chat-events.bdd` + `generation-template-prompt` suites (115 tests)

Not attached to pr-auto and not merged, per instruction.


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: b643bc2cbc6d9959d629f2e029e2c28bfa07f5e2
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Moves every text-to-video parameter off the template chip and onto a chip under the chat input, whose value domains are derived from the video model actually in effect (thread pin → member default → `DEFAULT_VIDEO_MODEL`). The parameters ride the existing `runOptions` channel as a new optional `runOptions.video`, are rendered into that run's system prompt as overridable defaults, and are never persisted. The template picker's video model row and the inline chip's settings popover are removed, so nothing in the composer writes `selection.videoOptions` any more. The diff also carries `9e768219f6`, a follow-up polish of the #27607 desktop model-mode controls that never landed in main.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no unhandled null/undefined paths | Pass | `video-run-options.ts:videoRunOptionsForSend` re-resolves the stored sparse patch against the effective model before sending, so a value the newly pinned model rejects cannot reach the run. `resolveVideoRunOptions` covers the empty-patch and retired-value cases. |
| `REVIEW.md` | Security: no injected secrets, no missing auth on new surfaces | N/A | No new endpoint, credential, or auth path; `runOptions.video` rides the already-authenticated chat send. |
| `REVIEW.md` | Type safety: no `any`, no broad types | Pass | No `any` in the diff. `chatRunVideoOptionsRequestSchema` is a zod enum-backed partial; `VideoRunOptionsPatch` aliases the inferred contract type. |
| `REVIEW.md` | Style: no function over ~100 lines, no duplicated logic | Pass | `createVideoRunOptionsUiSignals` and `createComposerPrimaryActionSignal` were extracted specifically to keep `createBasicComposerUiSignals` and `createComposerSubmissionSignals` under the 128-line lint ceiling. |
| `REVIEW.md` | New user-facing features gated behind a `FeatureSwitchKey` (P1) | Pass | Four independent gates on `VideoModelSelection`: chip render (`composer-video-options.tsx`), existing-thread send (`create-chat-thread.ts:sendRuntimeOptions`), new-thread send (`optimistic-chat-thread-page.ts`, via `videoModel`), and server (`zero-chat-events.command.ts` drops `runOptions.video` when off). |
| `REVIEW.md` | Feature Regression Guard: replaced behavior needs explicit coverage | Pass | The removed template-parameter path has two dedicated tests in `chat-composer-template-picker.test.tsx` — one with the switch off, one with it on — both asserting the sent selection carries only `stylePresetId` and the inline chip has a single zone. |
| `REVIEW.md` | DB/JSONB gate: persisted contract changes need migration or rationale | Pass | Nothing is persisted. `runOptions.video` is request-only; no schema, column, or `jsonb().$type<>()` change. Historical `selection.videoOptions` rows stay parseable and are still read by `legacyVideoTemplateSpecText`. |
| `docs/bad-smell.md` | Zero `any`, zero lint/type suppressions | Pass | No `any`, no `eslint-disable`/`@ts-ignore`/`@ts-expect-error`/`prettier-ignore` in the diff. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | All new imports in `video-run-options.ts`, `video-run-options-prompt.ts`, and `composer-video-options.tsx` are static. |
| `docs/bad-smell.md` | No hardcoded URLs or environment values | N/A | The diff introduces no URL or environment-derived configuration. |
| `docs/fallback.md` §2 | A non-GA feature-switched feature needs no compatibility code | Pass | `VideoTemplateOptions` is `enabled: false` with a staff-org allowlist, so removing its parameter UI outright — rather than keeping a switched-off branch — is what this section prescribes. |
| `docs/fallback.md` §3 | No fallback on a state the owning contract already prevents | Pass | Fixed in `b643bc2`: `effectiveVideoModel$` no longer re-narrows `userModelPreference.selectedVideoModel`, which is contract-typed to the catalog enum. `isVideoModelId` remains only on the thread pin, which `ThreadMeta` types as loose `string \| null`. |
| `docs/fallback.md` §1 | No negative tests asserting removed behavior stays removed | Pass | The template tests assert the current product behavior (a video template sends only `stylePresetId`), not the absence of deleted code. The one negative assertion in the API test pins the prompt's *wording contract* — that the block carries no pre-assembled flag string — not a tombstone. |
| `docs/fallback.md` §9 | Every fallback the diff introduces is declared in the PR summary | N/A | The diff introduces no fallback, so no `Fallbacks` section is required. It keeps `legacyVideoTemplateSpecText` and the API's `selectedTemplateOwnedVideoParameters`, both pre-existing. |
| `docs/deployment-compatibility.md` | Contract changes stay compatible in both directions | Pass | `runOptions.video` is optional and `chatRunOptionsRequestSchema` is non-strict, so a new app against an old API has the field stripped and the run falls back to model defaults — degraded, not an error. An old app simply omits it. |
| `docs/effect.md` | Render stays pure; no Store writes or signal creation in render | Pass | `createVideoRunOptionsUiSignals()` runs inside `createComposerUiSignals()`, outside render. `ComposerVideoOptionsChip` only reads via `useGet`/`useLastResolved` and returns `null` before rendering. |
| `docs/effect.md` | Derived value uses `computed`, not synchronization | Pass | `effectiveVideoModel$` and `chatPageEffectiveVideoModel$` derive the model from the pin/preference sources; nothing copies it into a second state. |
| `docs/cache.md` RC-2 | No render-time signal allocation | Pass | No `state`/`computed`/`command` is constructed from a component body. |
| `docs/cache.md` RC-5 | No duplicate mutable sources of truth | Pass | `videoRunOptions$` stores only the sparse set the user changed, scoped to the composer, and re-resolves against the model on read — an explicitly scoped draft rather than a synchronized copy of the resolved options. |
| `.claude/skills/ccstate/SKILL.md` | Accessors are called directly, never passed or wrapped | Pass | The earlier helper that took `get` as a parameter was replaced by `createVideoRunOptionsSignal`, a `command` that receives `{ get }` from ccstate. |
| `.claude/skills/ccstate/SKILL.md` | Async work carries an `AbortSignal` with an abort path | Pass | `createVideoRunOptionsSignal` takes the submission `AbortSignal` and calls `signal.throwIfAborted()` after awaiting the effective model. |
| `docs/testing.md` | Integration tests at entry points, not unit tests | Pass | Platform tests drive the page (click the video control, open the chip, pick a radio, send) and assert on the outgoing request body; API tests go through the chat-send endpoint and `readRun`. |
| `docs/testing.md` | Mock at the boundary only — no `vi.mock("../...")` | Pass | No `vi.mock` added; the platform tests use the existing MSW-backed `mockChatLifecycle`. |
| `docs/testing.md` | No direct `fetch` mocking, no fake timers | Pass | Neither `vi.stubGlobal("fetch", …)` nor `vi.useFakeTimers()` appears in the diff. |
| `docs/testing.md` | `feat:` commits carry integration tests (missing = P0) | Pass | Two new platform tests and two new API BDD assertions blocks; see Testing. |
| `docs/testing/testing-external-behavior.md` | Construct and verify through the external interface | Pass | Platform assertions come from `onRunCreate` (the network boundary) and rendered labels; the API test reads `appendSystemPrompt` through the `zeroRunsByIdContract` endpoint, not the database. |
| `docs/testing/api-testing.md` | API state set up and verified via real endpoints | Pass | `sendChatRun` posts to the chat-events endpoint and `api.readRun` reads the run back through its contract; no direct row access. |
| `docs/testing/app-testing.md` | Drive the page, assert on user-visible surfaces | Pass | The chip is located by its `aria-label` (`Video options 16:9 · 8s · 720p`), and the model-scoped options by `radiogroup`/`radio` roles. |
| `docs/testing/anti-patterns.md` | No asserting on mocks as the sole verification | Pass | Assertions are on the request payload and rendered labels, not on call counts. |
| `docs/testing/patterns.md` | Tests live beside the surface they exercise, existing suites updated in place | Pass | New cases were added to the existing `chat-composer-model-selection-and-providers` and `chat-events.bdd` suites; the superseded template-popover test was rewritten rather than left alongside. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None outstanding. One was found and fixed during review:

- `apps/platform/src/signals/chat-page/create-chat-thread.ts:effectiveVideoModel$` — re-narrowed `userModelPreference.selectedVideoModel` with `isVideoModelId`, but the contract types that field as `videoModelIdSchema.nullable().optional()`, so the guard covered a state the contract prevents (`docs/fallback.md` §3). It also disagreed with the sibling `chatPageEffectiveVideoModel$`, which omits it. Fixed in `b643bc2`.

#### Notes (non-blocking)

- **Switch-off behavior is a deliberate regression, not a gap.** The removals are ungated, so a staff org — the only population with `VideoTemplateOptions` on — loses the template parameter UI on merge and cannot get it back by flipping either switch, because `VideoModelSelection` is off for everyone. `docs/fallback.md` §2 explicitly sanctions this for a non-GA staff-only feature ("a temporary 500 or a broken screen during the cutover is acceptable"), so it is not a finding. Worth pairing the rollout with a `VideoModelSelection` staff allowlist so the replacement lands with the removal.
- **Mixed scope.** `9e768219f6` polishes the #27607 model-mode controls (icon-only Chat/Video triggers) and is unrelated to video parameters. It never landed in main, so it correctly rides this PR, but it widens the blast radius beyond the title.
- **Parameters do not survive a reload,** and a queued message dispatched later carries none, since the queued path replays from the persisted chat event. Declared in the PR description as the deliberate cost of not persisting.

### Testing

- **Coverage: Adequate.** Platform: `sends the video parameters chosen on the composer chip` proves only the values moved off the model default reach `runOptions.video`; `offers the pinned model's own values on the composer chip` proves a MiniMax-H3 thread offers `768p`/`2k` and never `720p`, which is the invalid-combination class that #27610 had to reject at the endpoint. Template regression: `sends a video template without any run parameters on it` (switch off) and `leaves every video parameter to the composer when the switch is on`. API: `runOptions.video` reaches `appendSystemPrompt` as `# Video Generation Defaults` with the per-parameter override rule and no generation flags, and a send that set nothing carries no trace of the block.
- **Conventions: Compliant.** No internal `vi.mock`, no fake timers, no direct `fetch` stubbing, no mock-call assertions; setup and verification both go through external interfaces.
- **Verified locally on this head:** `pnpm -F @okouai/app check-types`, `pnpm -F api check-types` (all 14 tasks), `pnpm -F @okouai/app run lint`, `pnpm knip`, the two platform composer suites (94 tests), and `chat-events.bdd` (97 tests) — all green after the base moved to `main`.

PR #27634, opened 2026-08-17
fix(cli): stop naming a video model the server will not use

bingjiez666:
Makes `okou generate video` stop naming a video model the server will not use. Follow-up to #27610, which fixed the server side of the same failure.

Two commits, both about `--model`:

1. `9a8eab9` — drop the hardcoded default, so the CLI stops claiming a model the caller never picked.
2. `3c93e48` — drop `--model` entirely inside an agent run, where the server ignores it anyway.

## Why

The CLI declared its own default:

```ts
.option("--model <model>", "...", "dreamina-seedance-2.0-fast")
```

so every request named a model even when the caller never picked one. That is why an agent asked only for "生成小狗视频" still sent `--model dreamina-seedance-2.0-fast --resolution 720p`: it read the default out of `-h` and passed it on, believing it was making a choice.

Inside a run the endpoint then replaces that model, and it does not care whether the caller was explicit:

```ts
runVideoModel === null
  ? bodyResult.data
  : withEnforcedRunVideoModel(bodyResult.data, runVideoModel)
```

#27610 made the replacement succeed instead of failing, but the agent still announced Seedance/720p up front, delivered MiniMax/2K, and — with nothing in the exchange marking the substitution as intentional — reported it to the user as a routing fault:

> 这属于生成路由或参数处理异常

Both halves are the same defect: the request states a model choice that will not hold.

## Following the composer

The chat composer already keeps this distinction, and this PR copies its rule rather than inventing one. `video-template-options-popover.tsx:toVideoOptionsPatch` subtracts the model's default before persisting, so a selection that matches the default is stored as absent:

```ts
...(next.model !== DEFAULT_VIDEO_MODEL ? { model: next.model } : {}),
```

Defaults are resolved only where they are used — `resolveVideoGenerationOptions` for a preview, `pinnedModel ?? userPreference ?? DEFAULT_VIDEO_MODEL` for the chip label — and never written back. `resolveVideoGenerationOptions`'s docblock states the invariant directly: **stored options are sparse**.

The agent model works the same way: `resolvePiSandboxModelConfig` assembles the run's model into the sandbox, so the agent never asks for one and can never misreport it.

## What changes

**No default.** Omitting `--model` omits `model` from the request body, and the server resolves it: the run's model when there is one, otherwise `DEFAULT_VIDEO_MODEL`. That constant is `dreamina-seedance-2-0-fast-260128`, exactly what the alias `dreamina-seedance-2.0-fast` normalized to, so an unpinned request resolves to the same model as before.

**Ignored inside a run.** `decodeSandboxTokenPayload()` decodes `OKOU_TOKEN` locally — no round trip — and returns a payload carrying `runId` for a run-scoped token, `undefined` for a personal one. That is the same distinction the endpoint makes:

```ts
const runId =
  auth.tokenType === "zero" || auth.tokenType === "sandbox"
    ? auth.runId
    : undefined;
```

`billing-capabilities.ts:currentTokenCanReadBilling()` already branches on that helper, on this same code path. Inside a run the model is cleared once at the top of the action, so the request body, the `--template` packet's "Requested Parameters", and the local frame-image validation all describe what will actually happen. Ignoring is silent, matching the endpoint; the flag is not rejected.

Outside a run `--model` is the only way to choose, and it still works for every model in the catalog.

Two call sites needed to tolerate an absent model:

- `validateVideoOptions` skipped frame-image aspect-ratio validation for MiniMax by reading `options.model.toLowerCase()`. With no model it runs the validation, which is what it already did under the Seedance default.
- `createVideoTemplateDetails` renders `provider default` through the existing `formatOptionalVideoParameter`, the wording `--resolution` already uses for the same situation: no client default, decided downstream.

## Behavior

| Where | `--model kling-v3-4k` | Before | After |
| --- | --- | --- | --- |
| Agent run, run model applied | passed | discarded by the server, reported as a choice | dropped by the CLI, never claimed |
| Agent run, `VideoModelSelection` off | passed | honoured | **`DEFAULT_VIDEO_MODEL`** — see below |
| Personal token, no run | passed | honoured | honoured |
| Anywhere | omitted | CLI sent Seedance | server resolves |

## The one behavior change worth flagging

The CLI cannot read feature switches, so it branches on "is this a run", not on "will the server actually apply the run's model". When `VideoModelSelection` is off, the endpoint returns early and honours the request's model, so an explicit `--model` inside a run works today. After this PR it does not.

With the switch off, `loadEnforcedRunVideoModel` returns before it reads the run's snapshot:

```ts
if (!isFeatureEnabled(FeatureSwitchKey.VideoModelSelection, featureSwitchContext)) {
  return null;
}
const runVideoModel = await loadRunVideoModel(db, runId);
```

so the server never learns what the run is set to and falls back to `readString(body, "model", DEFAULT_VIDEO_MODEL)`. With the CLI having removed the field, an explicit `--model kling-v3-4k` inside a run silently produces `dreamina-seedance-2-0-fast-260128` — neither the requested model nor the run's. Callers that use `--model` inside runs today to reach 4k or Veo would quietly drop to Seedance.

The two commits therefore carry different risk. `9a8eab9` is safe under either switch state, because omitting the field resolves to the same model the CLI used to send. `3c93e48` is only correct once `VideoModelSelection` is on everywhere; shipping it before then releases the enforced half of a switched feature without the switch. Recommendation was to hold `3c93e48` until the switch is GA. The product owner reviewed that trade-off and accepted it, so both commits ship together: inside a run, an explicit `--model` resolves to `DEFAULT_VIDEO_MODEL` until `VideoModelSelection` is on everywhere.

## Scope

Only `--model`. `--aspect-ratio` (`16:9`) and `--duration` (`8s`) have the same kind of client-side default and get the same treatment in a follow-up PR; `--resolution` never had one.

I also considered pinning "server default must equal the old CLI default" with a test and decided against it: that equality is a one-time migration property, not an invariant worth protecting, and a test asserting it would be a tombstone blocking a future intentional change to the server default.

## Not fixed here

The 720p → 2k substitution on a MiniMax run still costs more than the 768p the caller was closer to wanting, and the agent still has no signal distinguishing an intentional substitution from a fault. Both are server-side.

## Testing

- Request body carries no `model` when the caller passes none.
- Request body carries no `model` when a run-scoped token is present and `--model kling-v3-4k` is passed.
- Both verified to fail without their change; the rest of the suite uses a non-run token and covers the outside-a-run half unchanged.
- `video.test.ts`: 12 passed
- `@okouai/cli` `check-types` and `lint` clean


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 4587b52d2b2ac8f94d1dd621c99762cd27f9f7c9
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Removes the `--model` default from `okou generate video` and drops the flag entirely inside an agent run, where the endpoint replaces it regardless. `insideAgentRun()` decodes the run-scoped `OKOU_TOKEN` locally, so the branch costs no round trip and uses the same signal the endpoint uses to decide whether to apply the run's model. Outside a run the flag is unchanged.

Main has since landed `buildVideoRunOptionsPrompt` (#27616), whose docblock states the same rule this PR implements on the client: *"The model is not part of this block. The run already carries the video model resolved from the thread pin and the member default, and the generation endpoint pins that snapshot regardless of what the prompt says."* The server side already treats the model as not the caller's to state; this makes the CLI stop stating it.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness: no wrong assumptions on the changed path | Pass | `insideAgentRun()` keys on the `vm0_sandbox_` prefix plus `scope: "okou"`; `tokens.ts` verifies both `zero` and `sandbox` tokens under that prefix and PAT under a different one, matching the endpoint's `tokenType === "zero" \|\| "sandbox"` runId condition. |
| `REVIEW.md` | Security: no new auth surface or secret exposure | Pass | The token payload is decoded, never verified or logged; signature verification stays server-side. |
| `REVIEW.md` | Type safety: no `any`, no suppressions | Pass | `model` becomes `model?: string`; the `options.model?.toLowerCase()` narrowing replaces an unchecked call. No casts added. |
| `REVIEW.md` | Style: no duplicated logic, comments explain WHY | Pass | `insideAgentRun()` reuses the existing `decodeSandboxTokenPayload` that `billing-capabilities.ts` already branches on; both new comments state the reason, not the mechanics. |
| `REVIEW.md` | Feature Regression Guard: existing behavior changed without a gate | Finding (P1) | Inside a run with `VideoModelSelection` off, an explicit `--model` is honoured today and resolves to `DEFAULT_VIDEO_MODEL` after this PR. No switch, capability, or config gate contains it. See finding 1. |
| `REVIEW.md` | Feature Regression Guard: existing flow most likely to regress is covered | Pass | `should generate a video and print the /f file URL` uses a non-run token and asserts `model: "dreamina-seedance-2.0"` reaches the body, pinning the outside-a-run path this PR must not change. |
| `REVIEW.md` | DB/JSONB gate | N/A | No schema, contract, or `jsonb().$type<>()` change; CLI-only diff. |
| `docs/bad-smell.md` | Zero `any` and zero lint/type suppressions | Pass | Neither appears in the diff. |
| `docs/bad-smell.md` | No dynamic `import()` | Pass | `decodeSandboxTokenPayload` is a static top-of-file import. |
| `docs/bad-smell.md` | No hardcoded configuration or fallback URLs | Pass | The diff removes a hardcoded constant rather than adding one. |
| `docs/fallback.md` | Does the diff introduce a fallback needing a `Fallbacks` section? | N/A | `formatOptionalVideoParameter(options.model, "provider default")` is a presentational default for a contract-optional value (§6.4), not a cross-version branch, a `??` on a guaranteed field, or a retired-protocol reader. No `Fallbacks` section required. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | Both new cases assert live request shapes, not the absence of deleted code. |
| `docs/deployment-compatibility.md` | Old clients keep working across rollout | Pass | Request and response contracts unchanged. Sandboxes running the old CLI for up to 2h keep sending `model`; the endpoint already replaces it inside a run and accepts it outside one. |
| `docs/testing.md` | `fix:` commit carries a regression test | Pass | One case per commit, each verified to fail without its change. |
| `docs/testing.md` | Integration tests only; no unit tests for internal functions | Pass | `insideAgentRun` is exercised through `generateCommand.parseAsync`, not imported. |
| `docs/testing.md` | Mock at boundary only; MSW for HTTP; no fake timers | Pass | Only MSW handlers added; no `vi.mock()`, no timer control. |
| `docs/testing.md` | Test behavior, not mocks | Pass | Assertions are on the captured request body, not on call counts. |
| `docs/testing/cli-testing.md` | Enter through `command.parseAsync()`, mock only the Web API | Pass | Both cases drive `generateCommand.parseAsync` and stub only the video endpoint. |
| `docs/testing/cli-testing.md` | Set the canonical token explicitly; do not mock the config module | Pass | `buildRunToken()` builds a real `vm0_sandbox_` JWT shape and is applied with `vi.stubEnv("OKOU_TOKEN", …)`, matching `credit.test.ts` and `computer-use.test.ts`. |
| `docs/testing/cli-testing.md` | Reset command option state between parses | Finding (P1) | The first case depends on running before any other parse, and the second leaves `model: "kling-v3-4k"` on the shared `videoCommand`. See finding 2. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

1. `turbo/apps/cli/src/commands/shared/video-generate.ts` — the run branch ships without a gate, and it is the enforced half of a switched feature. `loadEnforcedRunVideoModel` checks `VideoModelSelection` **before** reading the run snapshot, so with the switch off the endpoint never learns the run's model and falls back to `readString(body, "model", DEFAULT_VIDEO_MODEL)`. Because the CLI has removed the field, `--model kling-v3-4k` inside a run then silently produces `dreamina-seedance-2-0-fast-260128` — neither the requested model nor the run's. `REVIEW.md`'s Feature Regression Guard says to request changes when a behavior change "should be scoped but lacks a feature switch, capability flag, config or permission gate, or equivalent containment", and the CLI cannot read switches, so no containment is available short of holding the commit. **The PR summary declares this explicitly and records that the product owner reviewed and accepted the trade-off**, which is why this is reported rather than blocking; a reviewer who disagrees should say so here, and the containing change is to hold `3c93e48` until `VideoModelSelection` is GA. (`REVIEW.md`, Feature Regression Guard)
2. `turbo/apps/cli/src/commands/generate/__tests__/video.test.ts` — both new cases depend on `videoCommand`'s option state surviving across `parseAsync` calls. The first asserts the full default payload and only holds because it is declared first; the second stubs a run token and leaves `model: "kling-v3-4k"` on the shared command for whatever follows. `docs/testing/cli-testing.md` shows the intended remedy (`command.setOptionValue(...)` in `beforeEach`), and the file also never calls `vi.unstubAllEnvs()` — today the env stub is harmless only because `beforeEach` re-stubs `OKOU_TOKEN` every time. Failures would be loud rather than silent, so this is robustness rather than a correctness defect, but a `beforeEach` that resets `model` and unstubs envs would remove the ordering coupling. (`docs/testing/cli-testing.md`)

#### Notes (non-blocking)

3. `createVideoTemplateDetails` now prints `Model preference … : provider default` inside a run, where the model is in fact known and fixed. Accurate from the CLI's own standpoint — it made no choice — but "the model this run is set to" would be the truthful statement for that context if the packet wording is revisited.
4. The first new case pins `aspectRatio: "16:9"` and `duration: "8s"` in the expected payload. Those are the very client-side defaults the agreed follow-up PR removes, so that PR must update this assertion; worth knowing before it is written.
5. Interaction worth tracking, created by main rather than by this PR: `buildVideoRunOptionsPrompt` now tells the agent the composer's chosen aspect ratio, duration, and resolution as *values*, deliberately "never the generation flags they map to". If the agent does not translate a 9:16 chip into `--aspect-ratio 9:16`, the CLI's surviving `16:9` default is sent and silently overrides the user's own selection. That strengthens the case for the follow-up PR; it is out of scope here.

### Testing

- Coverage: Adequate. Each commit ships a case that was verified to fail without it — request body carries no `model` when none is passed, and none when a run token is present with `--model kling-v3-4k`. The outside-a-run path that must not change is pinned by the pre-existing explicit-`--model` case.
- Conventions: Compliant. `parseAsync` entry point, MSW at the HTTP boundary, real token-shaped fixtures matching `credit.test.ts`, no internal mocks, no fake timers. The option-state and env-cleanup gap is finding 2.
- Local: `video.test.ts` 12 passed after merging `origin/main`; `@okouai/cli` `check-types` and `lint` clean.
- CI: the previous head failed `cli-e2e-02-browser` (`wait_for_sign_in_email_code_ready`), `cli-e2e-03-runner-prepare` (token-generation `TimeoutError`), `cli-e2e-02-playwright` (onboarding element never visible), and `ci-gate-turbo` as their aggregate — all environment failures unrelated to a two-file CLI diff, reported to the CI channel, and re-running on the merge commit.

PR #27636, opened 2026-08-17
test(python): cover simultaneous firewall-auth cleanup

### LINKED ISSUE #27627: test(python): cover simultaneous firewall-auth connection cleanup
## Problem

`_open_connected_stream()` can receive more than one successful connection attempt in the same `asyncio.wait()` batch. It selects the lowest-address-index winner and explicitly aborts every other successful socket:

- `crates/runner/mitm-addon/src/firewall_auth_client.py:383-470` owns the staggered multi-address connection race.
- `crates/runner/mitm-addon/src/firewall_auth_client.py:419-448` collects simultaneous successes, selects one winner, and closes the non-winning sockets at lines 433-435.
- `crates/runner/mitm-addon/tests/test_firewall_auth_client.py:1203-1331` covers connect failure, a pending loser, bounded pending attempts, and cancellation, but never makes two connection attempts succeed in the same completed batch.

The full branch-coverage run leaves line 435 uncovered, so the explicit successful-loser cleanup contract can regress without CI detecting it.

## Evidence

Audited at `0eac48b359d85af21bb02209933acfe90d8c4d3b`.

A temporary mutation replaced the non-winning socket cleanup:

```python
if connected_socket is not winner_socket:
    _abort_socket(connected_socket)
```

with a no-op. All existing firewall-auth client tests still passed:

```text
88 passed
```

An audit-only async test used a barrier to make two `_connect_socket()` tasks complete together. It asserted that the first socket was passed to `asyncio.open_connection()` while the second received `shutdown(SHUT_RDWR)` and `close()`. The test passed against the unmodified source and failed under the mutation because the loser received neither call.

## Impact

Without deterministic coverage, a cleanup regression can leave an already-connected losing transport dependent on object finalization instead of the explicit abort path. Repeated firewall-auth requests against multi-address hosts can then retain upstream connections or file descriptors longer than intended and make connection-race resource failures difficult to reproduce.

## Acceptance criteria

- Add a deterministic async test that releases at least two successful connection attempts in the same scheduler turn.
- Assert that address ordering selects the intended winner and that every other successful socket is shut down and closed exactly once.
- Assert that the winner is not closed before ownership is transferred to the returned stream.
- Fault-inject `asyncio.open_connection()` failure after a socket wins and verify that the winning socket is also aborted.
- Avoid timing sleeps; use events or barriers so the race is stable under CI load.

seven332:
## Summary

- release two real firewall-auth connection attempts through an asyncio barrier so both succeed in one completed batch
- assert resolver-order winner selection, pre-handoff winner ownership, and exactly-once loser shutdown and close
- fault-inject stream wrapping failure and assert the winning socket is aborted exactly once

## Tests

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_firewall_auth_client.py` (`90 passed`)
- `uv run --no-sync python -m pytest tests/` (`5565 passed`)

## Compatibility

- test-only change; no production behavior, API, schema, persisted state, dependency, generated artifact, runner protocol, migration, or rollout changes
- no full local Vitest run or local development server was used because this PR affects only the mitm-addon Python test surface

Closes #27627

### DIFF
--- a/crates/runner/mitm-addon/tests/test_firewall_auth_client.py
+++ b/crates/runner/mitm-addon/tests/test_firewall_auth_client.py
@@ -56,6 +56,46 @@ type _TestServerHandler = Callable[
 ]
 type _SocketAddress = tuple[str, int] | tuple[str, int, int, int]
 type _SockConnect = Callable[[socket.socket, _SocketAddress], Coroutine[object, object, None]]
+_REAL_SOCKET = socket.socket
+
+
+class _LifecycleSocket(_REAL_SOCKET):
+    def __init__(
+        self,
+        family: int = -1,
+        socket_type: int = -1,
+        proto: int = -1,
+        fileno: int | None = None,
+    ) -> None:
+        super().__init__(family, socket_type, proto, fileno)
+        self.shutdown_calls: list[int] = []
+        self.close_call_count = 0
+
+    def shutdown(self, how: int) -> None:
+        self.shutdown_calls.append(how)
+        super().shutdown(how)
+
+    def close(self) -> None:
+        self.close_call_count += 1
+        super().close()
+
+
+@dataclass
+class _LifecycleSocketFactory:
+    sockets: list[_LifecycleSocket] = field(default_factory=list)
+
+    def __call__(
+        self,
+        family: int = -1,
+        socket_type: int = -1,
+        proto: int = -1,
+        fileno: int | None = None,
+    ) -> socket.socket:
+        if fileno is not None:
+            return _REAL_SOCKET(family, socket_type, proto, fileno)
+        created = _LifecycleSocket(family, socket_type, proto)
+        self.sockets.append(created)
+        return created
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #27636

**Reviewed head:** `eeb92417cbbc644d23d85ebcc6ff10d159ff009a`

### Summary

The PR adds public-entry integration coverage for simultaneous successful firewall-auth connections using real loopback sockets and barrier-owned synchronization. It verifies resolver-order winner selection, exact successful-loser shutdown and close, winner ownership at stream handoff, and winner cleanup when stream construction fails.

### Findings

No findings.

### Testing

The assertions cover every acceptance criterion without timing sleeps or internal-function mocks. Local validation passed uv lock/sync, Ruff format and lint, BasedPyright, all 90 firewall-auth client tests, and all 5,565 mitm-addon tests. The same-batch case also passed 20 consecutive focused runs, and both new cases passed with `ResourceWarning` promoted to errors. Remote CI and coverage remain independently verified by the PR pipeline.

### Verdict

LGTM

PR #27638, opened 2026-08-17
refactor(sandbox-fc): use upstream abort-on-drop handle

### LINKED ISSUE #27625: refactor(rust): replace the local abort-on-drop task wrapper
## Problem

`crates/sandbox-fc/src/snapshot/runtime.rs:31-60` defines `AbortOnDropTask<T>`, a local wrapper around `tokio::task::JoinHandle<T>` that:

- forwards `Future::poll`,
- exposes `abort()`, and
- aborts the task from `Drop`.

The snapshot startup path uses it for the vsock listener at `crates/sandbox-fc/src/snapshot/runtime.rs:102-123`, so cancellation of the workflow cannot detach that listener.

This is the same ownership primitive already provided by `tokio_util::task::AbortOnDropHandle<T>`. `sandbox-fc` already depends on `tokio-util` at `crates/sandbox-fc/Cargo.toml:22`; the upstream type only requires enabling its `rt` feature.

Keeping a private copy creates a second source of truth for a cancellation-critical lifecycle guarantee. The local copy also omits the upstream type's `#[must_use]` guard and supporting APIs. Future snapshot task use sites or Tokio cancellation changes can therefore evolve against the local wrapper instead of the maintained dependency primitive.

## Acceptance criteria

- Enable the `tokio-util` feature needed for `tokio_util::task::AbortOnDropHandle` in `sandbox-fc`.
- Replace `AbortOnDropTask` and its manual `Future`/`Drop` implementations with the upstream type.
- Preserve the current explicit abort-and-await behavior when Firecracker startup fails.
- Keep coverage proving that dropping the handle and explicitly aborting it both remove the pending vsock listener.

seven332:
## Summary

- enable `tokio-util/rt` for `sandbox-fc` and use `AbortOnDropHandle` for the snapshot vsock listener
- remove the duplicate local `Future`/`Drop` wrapper while preserving explicit abort-and-await cleanup
- keep real listener-socket coverage for implicit drop and explicit abort cancellation

## Scope

This is an internal task-ownership refactor. It does not change public APIs, protocols, persisted state, deployment compatibility, or dependency versions.

## Validation

- `cargo test -p sandbox-fc --lib snapshot::runtime::tests::abort_on_drop_handle`
- `cargo clippy -p sandbox-fc --all-targets -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p sandbox-fc --no-deps`
- `cargo test -p sandbox-fc` (800 lib tests, 1 integration test, 0 ignored)
- pre-commit hook: workspace Cargo format, documentation, and Clippy checks; file-size check; commitlint

Closes #27625

### DIFF
--- a/crates/sandbox-fc/Cargo.toml
+++ b/crates/sandbox-fc/Cargo.toml
@@ -19,7 +19,7 @@ sha2.workspace = true
 thiserror.workspace = true
 tempfile.workspace = true
 tokio = { workspace = true, features = ["process", "fs", "rt", "rt-multi-thread", "macros", "sync", "io-util", "net", "time"] }
-tokio-util.workspace = true
+tokio-util = { workspace = true, features = ["rt"] }
 tracing.workspace = true
 uuid = { workspace = true, features = ["v4"] }
 vsock-host.workspace = true
--- a/crates/sandbox-fc/src/snapshot/runtime.rs
+++ b/crates/sandbox-fc/src/snapshot/runtime.rs
@@ -1,10 +1,7 @@
-use std::future::Future;
 use std::io;
-use std::pin::Pin;
-use std::task::{Context, Poll};
 use std::time::Duration;
 
-use tokio::task::JoinHandle;
+use tokio_util::task::AbortOnDropHandle;
 use tracing::info;
 use vsock_proto::ExecTermination;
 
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #27638

**Reviewed head:** `9d6d55e4e5293236df614c9a99d10250d2958a50`

### Summary

The PR replaces the private abort-on-drop wrapper with the matching maintained `tokio-util` primitive, declares the package-local feature it consumes, and preserves both implicit cancellation and explicit abort-and-await cleanup.

### Findings

No findings.

### Testing

Focused listener lifecycle tests (2), formatting, all-target package Clippy, rustdoc, the full `sandbox-fc` suite (800 lib + 1 integration), and the broader pre-commit Cargo hooks all passed with no ignored tests. A Firecracker/KVM E2E was not run locally; repository guidance excludes VM creation from local Rust tests, while the changed cleanup behavior is covered at the real host-vsock listener boundary.

### Verdict

LGTM

PR #27639, opened 2026-08-17
perf(python): bound browser user-agent classification work

### LINKED ISSUE #27629: perf(python): bound browser User-Agent classification work
## Problem

At current `main` (`0eac48b359d85af21bb02209933acfe90d8c4d3b`), every registered HTTP request reaches the browser pass-through heuristic before trusted-authority and firewall dispatch:

- `crates/runner/mitm-addon/src/request_classification.py` calls `is_browser_passthrough_heuristic()` immediately after registered-VM metadata is stored.
- The heuristic currently decodes/folds the complete `User-Agent`, lowercases it, prefixes it into a second complete string, checks for `mozilla/`, and then performs up to nine separate substring scans for browser markers.
- `requestheaders()` probes classification and `request()` can classify the same flow again when that probe is not carried forward.

A sandbox controls this header. Pinned mitmproxy 12.2.3 does not cap an HTTP/1 request head before addon hooks; its HTTP/1 layer still contains `FIXME: protect against header size DoS`. A value beginning with `Mozilla/` but containing none of the recognized markers therefore creates request-sized decode, lowercase, and prefix allocations and repeated proportional scans on the shared mitmproxy event-loop thread.

This work is unnecessary for overlong values that cannot plausibly be an ordinary browser User-Agent. The request can continue through the existing firewall path without treating the value as a browser pass-through signal.

## Evidence

The original CPython 3.14.0 probe called the exact string helper with input construction outside the measured region:

| Added payload | Median synchronous time | Peak traced allocation |
| ---: | ---: | ---: |
| 1 MiB | 9.599 ms | 2.00 MiB |
| 5 MiB | 46.859 ms | 10.00 MiB |
| 16 MiB | 149.454 ms | 32.00 MiB |
| 33 MiB | 311.473 ms | 66.00 MiB |

A follow-up seven-call probe measured the real `is_browser_passthrough_heuristic()` wrapper with `mitmproxy.http.Headers`, also constructing input and the flow outside measurement:

| Added payload | Median synchronous time | Peak traced allocation |
| ---: | ---: | ---: |
| 1 MiB | 4.683 ms | 3.00 MiB |
| 5 MiB | 27.100 ms | 15.00 MiB |
| 16 MiB | 55.825 ms | 48.00 MiB |
| 33 MiB | 194.252 ms | 99.00 MiB |

The real wrapper exposes a third input-sized allocation from mitmproxy header decoding before the two string-helper allocations. Timing varies with host state, while allocation scaling is deterministic.

Representative Chrome, Chromium, Edge, Firefox, Opera, Safari, iOS/CriOS, and HeadlessChrome fixtures inspected during planning range from 78 to 142 bytes.

## Decision

Proceed in one PR with a 4,096-byte maximum for the combined raw User-Agent field value. This number is a work budget for a spoofable business heuristic, not an HTTP field-size or conformance limit. It is more than 28 times the longest representative fixture while keeping accepted classification work in the low-kilobyte range.

## Chosen direction

- Before `Headers.get()` decodes or folds User-Agent values, iterate the already-parsed raw `Headers.fields` tuple.
- Match User-Agent names case-insensitively and accumulate every matching raw value plus the two-byte `", "` separator mitmproxy inserts between duplicate field values.
- Return non-browser immediately once the combined value exceeds 4,096 bytes. Do not reject, truncate, or mutate the HTTP request; continue through normal authority and firewall classification.
- For values within the limit, preserve the exact current decode/fold, lowercase, leading-space, Mozilla, and nine-marker behavior.
- Move the single-use string logic into `is_browser_passthrough_heuristic()` so the size policy and marker decision have one owner, then remove the redundant helper.
- Keep timing/allocation measurements as reproducible PR evidence rather than a wall-clock CI gate or permanent one-purpose benchmark script.

## Alternatives considered

- **Guard only the decoded string helper:** removes the two measured string allocations and scans, but the real request path still decodes/folds the complete value first and allocates a third request-sized string.
- **Scan only a bounded prefix:** bounds work but lets an oversized attacker value retain `BrowserAllow` by putting markers inside the prefix, contrary to the conservative fallback requirement.
- **Add a proxy-wide request-head limit:** addresses the broader transport problem but requires cross-protocol size, rejection, forwarding, and observability policy and changes requests to 4xx instead of preserving firewall handling.
- **Use 1,024 bytes:** tighter but increases custom-browser false-negative risk without a need for the smaller budget.
- **Use 8,192 bytes:** more custom-value margin but doubles the work budget without a repository or protocol requirement.
- **Commit a benchmark script:** discoverable, but adds permanent maintenance for a one-off measurement in a package with no such convention.

## Risks and mitigations

- **Customized browser false negatives:** a valid customized browser User-Agent above 4,096 raw bytes will take the ordinary firewall path and may be denied or receive configured connector auth. The 28x margin protects ordinary formats; the behavior is intentional and conservative for an untrusted signal.
- **Raw representation coupling:** `Headers.fields` is part of the pinned mitmproxy object representation and is already used throughout this addon for snapshots, auth, and capture logic. Accepted values still use `Headers.get()` for existing semantics.
- **Duplicate field semantics:** budget the same `", "` separator used by mitmproxy folding and cover a combined over-limit case.
- **Byte versus character length:** define the limit in raw bytes. Standard browser formats are ASCII, and the byte budget prevents multibyte input from consuming a larger transport/resource allowance.
- **Off-by-one drift:** cover exactly 4,096 bytes as accepted and values above it as non-browser.
- **False performance confidence:** benchmark the real wrapper and use a raw `bytes` decode guard for structural CI proof.
- **Incomplete transport hardening:** mitmproxy request-head buffering, non-User-Agent headers, and header-count/name work remain separate proxy-wide concerns.

## Acceptance criteria

- A combined raw User-Agent above 4,096 bytes is classified non-browser before raw decoding, duplicate folding, lowercase/prefix allocation, or marker scans.
- Exactly 4,096 bytes remains eligible for the existing marker logic.
- Ordinary Chrome, Chromium, Edge, Firefox, Opera, Safari, iOS/CriOS, and HeadlessChrome fixtures preserve browser pass-through behavior.
- A single over-limit value and duplicate values whose folded size crosses the limit follow ordinary firewall classification without setting browser metadata.
- Real registered-request coverage structurally proves that the oversized raw value is not decoded and verifies ordinary firewall auth behavior; CI uses no wall-clock threshold.
- A reproducible before/after real-wrapper benchmark records bounded synchronous work and allocation for 1, 5, 16, and 33 MiB values under pinned CPython 3.14.0.
- `uv lock --check`, locked environment sync, Ruff formatting/lint, BasedPyright, focused request integration coverage, and the complete mitm-addon suite pass.

## Delivery notes

- Expected source changes are limited to `src/request_classification.py` and `tests/test_request_handler_passthrough.py`.
- No API, schema, migration, queue payload, persisted state, dependency, environment variable, generated artifact, feature switch, staged rollout, or cleanup issue is required.
- Old and new runners can overlap without exchanging this decision; rollback is an ordinary runner artifact rollback.

## Distinctness

- #25451 bounds request-derived URL rendering in proxy diagnostics, not request-header classification.
- #25404 fast-paths runtime URL syntax checks, not User-Agent decode/fold/marker work.
- A proxy-wide request-head limit remains a broader separate hardening topic.

seven332:
## Summary

- cap the browser pass-through heuristic at a 4,096-byte combined raw User-Agent value
- reject the browser signal before mitmproxy decodes or folds oversized values, while leaving the HTTP request on the ordinary firewall path
- preserve existing browser classification for Chrome, Chromium, Edge, Firefox, Opera, Safari, iOS/CriOS, HeadlessChrome, and the inclusive limit
- structurally cover single-field and duplicate-field overflow without wall-clock assertions

## Behavior

The limit is a work budget for a spoofable business signal, not an HTTP field-size policy. Matching raw values are counted together with the `", "` separator mitmproxy uses to fold duplicate fields. Values above 4,096 bytes are not rejected, truncated, or mutated; they are classified as non-browser and continue through normal authority and firewall handling.

Accepted values retain the existing decode/fold, lowercase, leading-space, Mozilla, and browser-marker behavior. The former single-use string helper is folded into the heuristic so the raw bound and classification decision have one owner.

Deliberately customized browser User-Agent values above the budget can become conservative false negatives and may therefore receive configured firewall denial or credential handling. The representative browser fixtures in this PR are 78–142 bytes, leaving more than 28x margin.

## Performance

The exact `is_browser_passthrough_heuristic()` wrapper was measured on pinned CPython 3.14.0 with real `mitmproxy.http.Headers`. Each raw value and flow was constructed outside the measured region; medians use seven calls.

| Added payload | Before median | Before peak | After median | After peak |
| 1 MiB | 4.683 ms | 3.00 MiB | 0.000333 ms | 83 bytes |
| 5 MiB | 27.100 ms | 15.00 MiB | 0.000250 ms | 83 bytes |
| 16 MiB | 55.825 ms | 48.00 MiB | 0.000625 ms | 83 bytes |
| 33 MiB | 194.252 ms | 99.00 MiB | 0.000250 ms | 83 bytes |

Reproduction command, run from `crates/runner/mitm-addon` at the relevant before/after revision:

```bash
PYTHONPATH=src uv run --no-sync python -c 'import statistics, time, tracemalloc; from types import SimpleNamespace; from mitmproxy import http; from request_classification import is_browser_passthrough_heuristic
for mib in (1, 5, 16, 33):
 raw = b"Mozilla/5.0 " + b"x" * (mib * 1024 * 1024)
 flow = SimpleNamespace(request=SimpleNamespace(headers=http.Headers([(b"User-Agent", raw)])))
 samples = []
 for _ in range(7):
  start = time.perf_counter_ns(); result = is_browser_passthrough_heuristic(flow); samples.append(time.perf_counter_ns() - start)
 tracemalloc.start(); is_browser_passthrough_heuristic(flow); _, peak = tracemalloc.get_traced_memory(); tracemalloc.stop()
 print(mib, result, round(statistics.median(samples) / 1_000_000, 6), peak)'
```

## Scope exclusions

- does not add a parser-wide request-head or field-size limit
- does not address oversized non-User-Agent fields or work proportional to header count/name length
- does not add a permanent benchmark script, dependency, API, schema, persisted state, protocol, or rollout gate

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_request_handler_passthrough.py` — 20 passed
- `uv run --no-sync python -m pytest tests/` — 5,573 passed

Closes #27629

### DIFF
--- a/crates/runner/mitm-addon/src/request_classification.py
+++ b/crates/runner/mitm-addon/src/request_classification.py
@@ -63,6 +63,9 @@ _BROWSER_USER_AGENT_MARKERS = (
     " opr/",
     " safari/",
 )
+# User-Agent is only a spoofable business passthrough signal. Values above this
+# budget conservatively continue through ordinary firewall classification.
+_MAX_BROWSER_USER_AGENT_BYTES = 4096
 
 type StaleTlsAdmissionReason = Literal[
     "client_ip_missing",
@@ -559,7 +562,27 @@ def current_public_destination_denial(
     )
 
 
-def is_browser_user_agent(user_agent: str | None) -> bool:
+def is_browser_passthrough_heuristic(flow: http.HTTPFlow) -> bool:
+    # Short-term business passthrough heuristic for browser-originated traffic.
+    # This is not trusted browser provenance: any sandbox client can set this
+    # header. The spoofable User-Agent heuristic is currently accepted as a
+    # known tradeoff until runner-owned browser provenance is prioritized again.
+    user_agent_bytes = 0
+    has_user_agent = False
+    for name, value in flow.request.headers.fields:
+        if name.lower() != b"user-agent":
+            continue
+        if has_user_agent:
+            user_agent_bytes += len(b", ")
+        user_agent_bytes += len(value)
+        if user_agent_bytes > _MAX_BROWSER_USER_AGENT_BYTES:
+            return False
+        has_user_agent = True
+
+    if not has_user_agent:
+        return False
+
+    user_agent = flow.request.headers.get("User-Agent")
     if not user_agent:
         return False
 
@@ -567,14 +590,6 @@ def is_browser_user_agent(user_agent: str | None) -> bool:
     return "mozilla/" in normalized and any(
         marker in normalized for marker in _BROWSER_USER_AGENT_MARKERS
     )
-
-
-def is_browser_passthrough_heuristic(flow: http.HTTPFlow) -> bool:
-    # Short-term business passthrough heuristic for browser-originated traffic.
-    # This is not trusted browser provenance: any sandbox client can set this
-    # header. The spoofable User-Agent heuristic is currently accepted as a
-    # known tradeoff until runner-owned browser provenance is prioritized again.
-    return is_browser_user_agent(flow.request.headers.get("User-Agent"))
 
 
 def restore_request_headers_probe_metadata(
--- a/crates/runner/mitm-addon/tests/test_request_handler_passthrough.py
+++ b/crates/runner/mitm-addon/tests/test_request_handler_passthrough.py
@@ -1,5 +1,7 @@
 """Ordinary and browser pass-through tests for the request hook."""
 
+import pytest
+
 import flow_metadata_keys as metadata_keys
 import mitm_addon
 import usage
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #27639

**Reviewed head:** `0f3a4ad8d99cbb21a9650c2caea94521d57b466a`

### Summary

The implementation bounds attacker-controlled browser User-Agent classification
before mitmproxy decoding and preserves the existing behavior for accepted
values. The change stays within the approved runner-local scope and keeps the
size policy and marker decision under one owner.

### Findings

No findings.

### Testing

The real registered-request entry point covers representative browser families,
the inclusive 4,096-byte boundary, single-field overflow, and duplicate-field
overflow. A raw `bytes.decode()` guard proves oversized values take ordinary
firewall auth without decoding. Locked Ruff, BasedPyright, 20 focused tests, and
the complete 5,573-test mitm-addon suite passed locally; CI status is checked
separately by the workflow.

### Verdict

LGTM

PR #27648, opened 2026-08-17
feat(composer): offer new-chat video model as an explicit default

bingjiez666:
## Problem

On the new-chat composer, picking a video model wrote it straight into the member default (`org_members_metadata.selected_video_model`). Picking a run model has not behaved that way since `NewChatDefaultModelAction` shipped — under that switch the pick only scopes the next new chat, and a notice under the composer offers **Set as default**.

This aligns the video model with the run model.

## Changes

- `setVideoModel$` (`agent-composer-signals.ts`) now mirrors `setModelSelection$`: when `NewChatDefaultModelAction` is on it only sets the session selection and never writes the member default. With the switch off, the previous silent write is unchanged.
- `ComposerTemporaryModelNoticeSlot` picks the notice for the composer's **current model mode**. The shared row markup moved into `ComposerTemporaryModelNoticeRow`; the run-model notice is otherwise untouched.
- New `chat.composer.temporaryVideoModelNotice` string in all 10 locales.

## Design decision: both models changed at once

One notice at a time, scoped to the mode the composer is in — not a merged row, not two stacked rows.

- Chat mode pressed → the run-model notice, and **Set as default** writes only the run model.
- Video mode pressed → the video-model notice, and **Set as default** writes only the video model.
- Nothing pending in the current mode → no row, even if the other mode has a pending change.

Rationale: the two mode chips (`Chat · <model>` / the video chip, both carrying `aria-pressed`) already tell the user which model they are editing. Tying the notice to that pressed state means the row always answers "the thing I just changed", one click sets exactly what the user is looking at, and the combinatorial notice strings — run model × service tier × video model — never have to exist in 10 languages.

## Preserving the run model on a video-only write

`PUT /api/okou/user-model-preference` requires `selectedModel`/`serviceTier`, so a video-only write has to send them. It sends **the stored preference**, not the resolved composer selection: those differ when the member's own run model is unset and the selection resolved to the workspace default. Sending the resolved value would silently pin that workspace default onto the member and stop them following future workspace changes.

`selectedVideoModel` is only sent when the video model is what changed — the field's own-property semantics mean "absent = leave alone", so the run-model notice cannot blank a stored video default.

## Known gap: mobile

Mode is a desktop concept — the video chip is `sm:`-only and `desktopModelMode$` stays `"chat"` below 640px, where video models are reached through a transient nested panel. So on mobile, with `NewChatDefaultModelAction` on, a video pick is temporary and there is no composer affordance to persist it (there is no settings surface for the video default either). Reusing the nested panel's open state would make the row flicker in and out with the popup, so I left it out rather than guess. Flagging it as a follow-up rather than folding a mobile mode model into this PR.

## Tests

`chat-composer-model-selection-and-providers.test.tsx`, 2 new cases (62 pass):

- a new-chat video pick writes nothing, shows `Temporarily switched to Veo 3.1 Fast for video`, and **Set as default** sends exactly `{selectedModel: null, serviceTier: null, selectedVideoModel: "fal-ai/veo3.1/fast"}` — the null run model proves the workspace default is not pinned.
- both models changed, then toggling Chat ↔ Video repeatedly: exactly one row each time, always the pressed mode's; and none at all while video mode is on its member default.

The existing "writes a landing selection to the member default and new thread" case still passes and now covers the switch-off path.

## Checks

`pnpm -F @okouai/app lint` (exit 0, no new warnings), `check-types` (production + tests), and the 4 composer/sidebar test files that touch this area (160 tests) all pass locally.

### DIFF
--- a/turbo/apps/platform/src/i18n/locales/de-DE/common.json
+++ b/turbo/apps/platform/src/i18n/locales/de-DE/common.json
@@ -1342,6 +1342,7 @@
       "temporaryFastModeDisabledNotice": "Der Fast-Modus ist für diesen Lauf vorübergehend deaktiviert",
       "temporaryFastModeEnabledNotice": "Der Fast-Modus ist für diesen Lauf vorübergehend aktiviert",
       "temporaryModelNotice": "Vorübergehend zu {{model}} gewechselt",
+      "temporaryVideoModelNotice": "Für Videos vorübergehend zu {{model}} gewechselt",
       "threadSuggestions": "Chat-Threads",
       "visualAttachmentsUnsupported": "{{modelName}} kann Bilder oder Videos nicht erkennen. Wechseln Sie vor dem Anhängen zu einem visionsfähigen Modell.",
       "workflows": {
--- a/turbo/apps/platform/src/i18n/locales/en-US/common.json
+++ b/turbo/apps/platform/src/i18n/locales/en-US/common.json
@@ -1342,6 +1342,7 @@
       "temporaryFastModeDisabledNotice": "Fast mode is temporarily disabled for this run",
       "temporaryFastModeEnabledNotice": "Fast mode is temporarily enabled for this run",
       "temporaryModelNotice": "Temporarily switched to {{model}}",
+      "temporaryVideoModelNotice": "Temporarily switched to {{model}} for video",
       "threadSuggestions": "Chat threads",
       "visualAttachmentsUnsupported": "{{modelName}} cannot recognize images or videos. Switch to a vision-capable model to attach them.",
       "workflows": {
--- a/turbo/apps/platform/src/i18n/locales/es-ES/common.json
+++ b/turbo/apps/platform/src/i18n/locales/es-ES/common.json
@@ -1368,6 +1368,7 @@
       "temporaryFastModeDisabledNotice": "El modo rápido está desactivado temporalmente para esta ejecución",
       "temporaryFastModeEnabledNotice": "El modo rápido está activado temporalmente para esta ejecución",
       "temporaryModelNotice": "Cambio temporal a {{model}}",
+      "temporaryVideoModelNotice": "Cambio temporal a {{model}} para vídeo",
       "threadSuggestions": "Conversaciones",
       "visualAttachmentsUnsupported": "{{modelName}} no reconoce imágenes ni vídeos. Cambia a un modelo con capacidad visual para adjuntarlos.",
       "workflows": {
--- a/turbo/apps/platform/src/i18n/locales/fr-FR/common.json
+++ b/turbo/apps/platform/src/i18n/locales/fr-FR/common.json
@@ -1368,6 +1368,7 @@
       "temporaryFastModeDisabledNotice": "Le mode rapide est temporairement désactivé pour cette exécution",
       "temporaryFastModeEnabledNotice": "Le mode rapide est temporairement activé pour cette exécution",
       "temporaryModelNotice": "Passage temporaire à {{model}}",
+      "temporaryVideoModelNotice": "Passage temporaire à {{model}} pour la vidéo",
       "threadSuggestions": "Conversations",
       "visualAttachmentsUnsupported": "{{modelName}} ne peut pas reconnaître les images ou les vidéos. Passez à un modèle capable de vision pour les joindre.",
       "workflows": {
--- a/turbo/apps/platform/src/i18n/locales/hi-IN/common.json
+++ b/turbo/apps/platform/src/i18n/locales/hi-IN/common.json
@@ -1342,6 +1342,7 @@
       "temporaryFastModeDisabledNotice": "इस रन के लिए फ़ास्ट मोड अस्थायी रूप से बंद है",
       "temporaryFastModeEnabledNotice": "इस रन के लिए फ़ास्ट मोड अस्थायी रूप से चालू है",
       "temporaryModelNotice": "अस्थायी रूप से {{model}} पर स्विच किया गया",
+      "temporaryVideoModelNotice": "वीडियो के लिए अस्थायी रूप से {{model}} पर स्विच किया गया",
       "threadSuggestions": "चैट थ्रेड",
       "visualAttachmentsUnsupported": "{{modelName}} छवियों या वीडियो को नहीं पहचान सकता। उन्हें संलग्न करने के लिए दृष्टि-सक्षम मॉडल पर स्विच करें।",
       "workflows": {
--- a/turbo/apps/platform/src/i18n/locales/id-ID/common.json
+++ b/turbo/apps/platform/src/i18n/locales/id-ID/common.json
@@ -1342,6 +1342,7 @@
       "temporaryFastModeDisabledNotice": "Mode Cepat dinonaktifkan sementara untuk proses ini",
       "temporaryFastModeEnabledNotice": "Mode Cepat diaktifkan sementara untuk proses ini",
       "temporaryModelNotice": "Sementara beralih ke {{model}}",
+      "temporaryVideoModelNotice": "Sementara beralih ke {{model}} untuk video",
       "threadSuggestions": "Thread chat",
       "visualAttachmentsUnsupported": "{{modelName}} tidak dapat mengenali gambar atau video. Ganti ke model yang mendukung visi untuk melampirkannya.",
       "workflows": {
--- a/turbo/apps/platform/src/i18n/locales/it-IT/common.json
+++ b/turbo/apps/platform/src/i18n/locales/it-IT/common.json
@@ -1368,6 +1368,7 @@
       "temporaryFastModeDisabledNotice": "La modalità Fast è temporaneamente disattivata per questa esecuzione",
       "temporaryFastModeEnabledNotice": "La modalità Fast è temporaneamente attivata per questa esecuzione",
       "temporaryModelNotice": "Passaggio temporaneo a {{model}}",
+      "temporaryVideoModelNotice": "Passaggio temporaneo a {{model}} per i video",
       "threadSuggestions": "Thread di chat",
       "visualAttachmentsUnsupported": "{{modelName}} non supporta immagini o video. Passa a un modello con capacità visive per allegarli.",
       "workflows": {
--- a/turbo/apps/platform/src/i18n/locales/ja-JP/common.json
+++ b/turbo/apps/platform/src/i18n/locales/ja-JP/common.json
@@ -1342,6 +1342,7 @@
       "temporaryFastModeDisabledNotice": "この実行では一時的に高速モードが無効です",
       "temporaryFastModeEnabledNotice": "この実行では一時的に高速モードが有効です",
       "temporaryModelNotice": "一時的に {{model}} に切り替えました",
+      "temporaryVideoModelNotice": "動画モデルを一時的に {{model}} に切り替えました",
       "threadSuggestions": "チャットスレッド",
       "visualAttachmentsUnsupported": "{{modelName}} は画像またはビデオを認識できません。ビジョン対応モデルに切り替えて取り付けてください。",
       "workflows": {
--- a/turbo/apps/platform/src/i18n/locales/ko-KR/common.json
+++ b/turbo/apps/platform/src/i18n/locales/ko-KR/common.json
@@ -1342,6 +1342,7 @@
       "temporaryFastModeDisabledNotice": "이 실행에서는 빠른 모드가 일시적으로 비활성화되었습니다",
       "temporaryFastModeEnabledNotice": "이 실행에서는 빠른 모드가 일시적으로 활성화되었습니다",
       "temporaryModelNotice": "일시적으로 {{model}}(으)로 전환됨",
+      "temporaryVideoModelNotice": "동영상 모델을 일시적으로 {{model}}(으)로 전환함",
       "threadSuggestions": "채팅 스레드",
       "visualAttachmentsUnsupported": "{{modelName}}은 이미지나 비디오를 인식할 수 없습니다. 첨부하려면 비전 기능이 있는 모델로 전환하세요.",
       "workflows": {
--- a/turbo/apps/platform/src/i18n/locales/pt-BR/common.json
+++ b/turbo/apps/platform/src/i18n/locales/pt-BR/common.json
@@ -1368,6 +1368,7 @@
       "temporaryFastModeDisabledNotice": "O modo rápido está temporariamente desativado para esta execução",
       "temporaryFastModeEnabledNotice": "O modo rápido está temporariamente ativado para esta execução",
       "temporaryModelNotice": "Alternado temporariamente para {{model}}",
+      "temporaryVideoModelNotice": "Alternado temporariamente para {{model}} em vídeo",
       "threadSuggestions": "Conversas",
       "visualAttachmentsUnsupported": "{{modelName}} não reconhece imagens nem vídeos. Troque para um modelo com visão para anexá-los.",
       "workflows": {
--- a/turbo/apps/platform/src/signals/external/user-model-preference.ts
+++ b/turbo/apps/platform/src/signals/external/user-model-preference.ts
@@ -1,4 +1,5 @@
 import { command, computed, state } from "ccstate";
+import type { VideoModel } from "@okouai/core/video-model-catalog";
 import {
   type UserPreferenceChangedPayload,
   userPreferenceChangedPayloadSchema,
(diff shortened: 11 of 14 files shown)


bingjiez666:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 01d25e66057e3b1ff004d0e449266bcfdfd4d5bc

### Summary

Puts the new-chat video model on the same footing as the run model: with `NewChatDefaultModelAction` on, `setVideoModel$` only scopes the session and the composer notice carries an explicit **Set as default**. The notice is scoped to the composer's current model mode so exactly one row shows at a time, and a new `chat.composer.temporaryVideoModelNotice` string is added to all 10 locales.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic — no race conditions or wrong assumptions | Finding (P1) | `zero-chat-composer.tsx:ComposerTemporaryVideoModelNotice` echoes `userPreference.selectedModel`/`serviceTier` from a cache that only refreshes on the `userPreferenceChanged` push, so a video write issued before that push reverts a run-model default just set from the sibling notice |
| `REVIEW.md` | Security — no new endpoints, auth checks, or secrets | N/A | No route, auth, or credential surface changed; the PR reuses the existing authenticated `PUT /api/okou/user-model-preference` |
| `REVIEW.md` | Type safety — no `any`, no broad types | Pass | No `any` or `as` casts added; `ComposerVideoModelSignals` and `VideoModel` carry the new props, and `check-types` covers both production and test configs |
| `REVIEW.md` | Style & maintainability — no oversized functions or duplicated logic | Pass | The shared notice markup is extracted into `ComposerTemporaryModelNoticeRow` instead of being copied into the second notice; each notice component stays well under the 128-line lint bound |
| `REVIEW.md` | `feat:` commits must have integration tests | Pass | `chat-composer-model-selection-and-providers.test.tsx` adds 2 cases covering the temporary pick, the exact write body, and mode-scoped notice switching |
| `REVIEW.md` | Feature regression guard — changed existing behavior is scoped and covered | Pass | The silent-write path is preserved for switch-off and still covered by the pre-existing "writes a landing selection to the member default and new thread" case; the new behavior is contained by `NewChatDefaultModelAction` + `VideoModelSelection`, both `enabled: false` |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, migration, contract, or `jsonb().$type<>()` change; `turbo/packages/db` and `turbo/packages/api-contracts` are untouched |
| `docs/bad-smell.md` | §1 zero tolerance for `any` | Pass | No `any` introduced in the diff |
| `docs/bad-smell.md` | §2 zero tolerance for lint/type suppressions | Pass | No `eslint-disable`, `oxlint-disable`, `@ts-ignore`, `@ts-expect-error`, or `prettier-ignore` added |
| `docs/bad-smell.md` | §5 no dynamic `import()` in production code | Pass | All new imports are static; the diff adds no `await import(...)` |
| `docs/bad-smell.md` | §7 fail fast — no recovery/fallback slop | Pass | The only guards are genuine loading states (`useLastResolved` returning `undefined`) and the product's documented video-default chain, not recovery logic |
| `docs/fallback.md` | §2 a non-GA feature switch needs no compatibility code | Pass | Both gates are non-GA (`NewChatDefaultModelAction` `enabled: false` + staff orgs, `VideoModelSelection` `enabled: false`), and the PR adds no dual-read/dual-write or tolerance branch for them |
| `docs/fallback.md` | §3 no `??`/`||` on states the contract already prevents | Pass | `userPreference.selectedVideoModel ?? DEFAULT_VIDEO_MODEL` defaults a genuinely nullable member column (§6.3) and reuses the resolution rule main already applies in `chatPageVideoModelSelection$` and `chatPageEffectiveVideoModel$`; `featureSwitch$[...] ?? false` is the established unset-switch read |
| `docs/fallback.md` | §9 every fallback introduced by the diff is declared in the PR summary | Pass | The diff introduces no new cross-version or compatibility fallback, so no `Fallbacks` section is required |
| `docs/fallback.md` | §1 no negative tests asserting removed behavior stays removed | Pass | The new `expect(...).toBeNull()` assertions pin live mode-scoping behavior, not the absence of deleted code |
| `docs/effect.md` | Choose the execution boundary — derived state in `computed`, user events in commands | Pass | Mode and switch reads stay derived (`desktopModelMode$`, `videoModelSelectionEnabled$`); the preference write runs in the existing `updateUserModelPreference$` command from a click handler |
| `docs/effect.md` | Keep React render pure | Pass | Both new components only read signals during render; the only write happens in the `onClick` handler |
| `docs/cache.md` | RC-1 no render-phase side effects | Pass | No Store writes, timers, DOM mutation, or resource creation during render in `ComposerTemporaryVideoModelNotice` or the slot |
| `docs/cache.md` | RC-2 no render-time signal allocation | Pass | No `state`/`computed`/`command` or signal factory is created in render; `signals.videoModel` is read from the already-built composer signal group |
| `docs/cache.md` | RC-3 no unbounded lifetime caches | N/A | The diff adds no module-level `Map`/`Set`/`WeakMap`, registry, or keyed cache |
| `.claude/skills/ccstate/SKILL.md` | DOM callback pattern — `detach(cmd(args, pageSignal), Reason.DomCallback)` | Pass | `ComposerTemporaryVideoModelNotice`'s `setAsDefault` uses exactly this shape with `pageSignal` from `useGet(pageSignal$)` |
| `.claude/skills/ccstate/SKILL.md` | View layer uses loadables for pending state | Pass | `useLoadableSet(updateUserModelPreference$)` drives `disabled`/`aria-busy`/spinner rather than a hand-rolled boolean |
| `docs/testing.md` | Integration tests only; mock at the boundary | Pass | Both cases drive the real page through `detachedSetupPage` and stub only HTTP via `context.mocks.api`; no `vi.mock()` of internal modules |
| `docs/testing/app-testing.md` | Use `waitFor()`, `context.mocks`, and test user flows | Pass | Assertions go through `waitFor`, mocks are registered on `context.mocks`, and no file-level `afterEach` cleanup is added |
| `docs/testing/testing-external-behavior.md` | Assert through the external user interface | Pass | The tests query visible notice text, `aria-pressed`, and the captured request body; they never read composer signals directly |
| `docs/testing/anti-patterns.md` | AP-1/AP-4/AP-5/AP-7 — no mock-call assertions, internal mocks, fake timers, or implementation details | Pass | No `expect(mock).toHaveBeenCalled()` as a sole assertion, no `vi.useFakeTimers()`, no `../../` mock paths, no internal-state assertions |
| `docs/deployment-compatibility.md` | Frontend → backend requests stay compatible across versions | Pass | No contract change; the request shape is unchanged and `selectedVideoModel` keeps its existing own-property "absent = leave alone" semantics, so an old bundle still writing the default silently stays valid |

### Findings

#### Critical (P0)

None.

- `turbo/apps/platform/src/views/zero-page/zero-chat-composer.tsx` — `ComposerTemporaryVideoModelNotice.setAsDefault` sends `selectedModel: userPreference.selectedModel` and `serviceTier: userPreference.serviceTier` read from `userModelPreference$`, but `updateUserModelPreference$` deliberately does **not** refresh that cache after a write ("Cross-device/default-model writes are reflected via the `userPreferenceChanged` realtime topic; do not reload the local cache here"). The cache only refreshes when the Ably push arrives.

  Repro on one new-chat composer with both switches on:
  1. Chat mode — switch the run model to Claude Sonnet 4.6 and click **Set as default**. The server stores `claude-sonnet-4-6`; the local cache still holds the previous run model.
  2. Before the `userPreferenceChanged` push lands (the request may still be in flight — the two notices are separate components with separate `useLoadableSet` state, so the second button is never disabled by the first write), switch to video mode, pick a different video model, and click **Set as default**.
  3. That second `PUT` carries the **stale** run model, silently reverting the default the user just set.

  This is directly observable in the added test: the notice stays mounted after a successful write and only clears once `triggerAblyEvent("userPreferenceChanged", ...)` fires, which is exactly the window in which the stale value is readable. The run-model notice is not exposed to this because it sends the composer's live `selection` and omits `selectedVideoModel` entirely; echoing stored fields back is new in this PR.

  Suggested fix: refresh the cached preference once the write resolves in this path (`reloadUserModelPreference$` is already exported), or narrow the write so it cannot carry a field the notice does not own. Please add a regression test that issues the run-model write and the video write back-to-back and asserts the second body preserves the newly stored run model.

### Testing

- Coverage: Adequate for the shipped behavior — the temporary-pick path, the exact write body (including the `selectedModel: null` case that proves the workspace default is not pinned onto the member), and Chat ↔ Video mode scoping are all covered, and the switch-off silent-write path keeps its existing case. Insufficient for the P1 above: no test covers two consecutive **Set as default** actions across modes.
- Conventions: Compliant. Integration-level, MSW-backed via `context.mocks.api`, `waitFor` for async assertions, no fake timers, no internal mocks, no mock-call-only assertions, and no file-level cleanup hooks.
- Behavior gating: Compliant. The new user-facing behavior sits behind `FeatureSwitchKey.NewChatDefaultModelAction` and `FeatureSwitchKey.VideoModelSelection`, both non-GA.

Note (non-blocking, already declared in the PR summary): below 640px there is no video mode, so a mobile video pick is temporary with no composer affordance to persist it. Both gates are non-GA, so per `docs/fallback.md` §2 this is a scoped staff-only gap rather than a GA regression — but it should not stay open once either switch moves toward GA.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: a1c754d6bc44a65298fe3953e619d0a3bba89570
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Re-review at `a1c754d6` after the fix commit. The PR puts the new-chat video model on the same footing as the run model: with `NewChatDefaultModelAction` on, `setVideoModel$` only scopes the session and the composer notice carries an explicit **Set as default**, scoped to the composer's current model mode so exactly one row shows at a time. The P1 raised on `01d25e66` is resolved.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic — no race conditions or wrong assumptions | Pass | The stale-run-model echo is fixed: `user-model-preference.ts:updateDefaultVideoModel$` reloads and re-reads the stored preference before echoing it, and `keeps a just-written run model when the video default follows it` fails without that reload (asserts `claude-fable-5` instead of `claude-sonnet-4-6`) |
| `REVIEW.md` | Security — no new endpoints, auth checks, or secrets | N/A | No route, auth, or credential surface changed; the PR reuses the existing authenticated `PUT /api/okou/user-model-preference` |
| `REVIEW.md` | Type safety — no `any`, no broad types | Pass | No `any` or `as` casts; `updateDefaultVideoModel$` takes a typed `VideoModel`, and repo-wide `check-types` passes 14/14 |
| `REVIEW.md` | Style & maintainability — no oversized functions or duplicated logic | Pass | Shared markup extracted into `ComposerTemporaryModelNoticeRow`; the fix moved the write out of the view into one semantic command, shrinking the click handler to a single `detach(...)` |
| `REVIEW.md` | `feat:` commits must have integration tests | Pass | 3 new cases in `chat-composer-model-selection-and-providers.test.tsx` covering the temporary pick, the exact write body, mode-scoped switching, and the back-to-back write ordering |
| `REVIEW.md` | Feature regression guard — changed existing behavior is scoped and covered | Pass | The switch-off silent-write path keeps its pre-existing "writes a landing selection to the member default and new thread" case; new behavior is contained by `NewChatDefaultModelAction` + `VideoModelSelection`, both `enabled: false` |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, migration, contract, or `jsonb().$type<>()` change; `turbo/packages/db` and `turbo/packages/api-contracts` untouched |
| `docs/bad-smell.md` | §1 zero tolerance for `any` | Pass | No `any` introduced |
| `docs/bad-smell.md` | §2 zero tolerance for lint/type suppressions | Pass | No `eslint-disable`, `oxlint-disable`, `@ts-ignore`, `@ts-expect-error`, or `prettier-ignore` added |
| `docs/bad-smell.md` | §3 no unnecessary try/catch; fail fast | Pass | `updateDefaultVideoModel$` adds no error handling; the GET and PUT propagate through `accept`, and `signal.throwIfAborted()` follows the await |
| `docs/bad-smell.md` | §5 no dynamic `import()` in production code | Pass | All new imports are static |
| `docs/bad-smell.md` | §7 fail fast — no recovery/fallback slop | Pass | The only guards are genuine loading states (`useLastResolved` returning `undefined`) and the product's documented video-default chain |
| `docs/fallback.md` | §2 a non-GA feature switch needs no compatibility code | Pass | Both gates are non-GA (`NewChatDefaultModelAction` `enabled: false` + staff orgs, `VideoModelSelection` `enabled: false`); no dual-read/dual-write added |
| `docs/fallback.md` | §3 no `??`/`||` on states the contract already prevents | Pass | `userPreference.selectedVideoModel ?? DEFAULT_VIDEO_MODEL` defaults a genuinely nullable member column (§6.3), reusing the rule main already applies in `chatPageVideoModelSelection$`/`chatPageEffectiveVideoModel$`; `featureSwitch$[...] ?? false` is the established unset-switch read |
| `docs/fallback.md` | §9 every fallback introduced by the diff is declared in the PR summary | Pass | The diff introduces no cross-version or compatibility fallback, so no `Fallbacks` section is required. The fix is a cache refresh, not a tolerance branch |
| `docs/fallback.md` | §1 no negative tests asserting removed behavior stays removed | Pass | The `toBeNull()` assertions pin live mode-scoping behavior, not the absence of deleted code |
| `docs/effect.md` | Choose the execution boundary; name the action, not the setter | Pass | The fix replaces an inline view-level request body with `updateDefaultVideoModel$`, a command named for the business action, composed through the existing `updateUserModelPreference$` |
| `docs/effect.md` | Keep React render pure | Pass | Both new components only read signals during render; the only write happens in the `onClick` handler |
| `docs/effect.md` | Make cancellation explicit | Pass | `updateDefaultVideoModel$` takes an `AbortSignal`, calls `signal.throwIfAborted()` after the awaited read, and forwards it to the inner command |
| `docs/cache.md` | RC-1 no render-phase side effects | Pass | No Store writes, timers, DOM mutation, or resource creation during render; `reloadUserModelPreference$` is set inside the command, not in render |
| `docs/cache.md` | RC-2 no render-time signal allocation | Pass | No `state`/`computed`/`command` created in render; `updateDefaultVideoModel$` is module-scope |
| `docs/cache.md` | RC-3 no unbounded lifetime caches | N/A | No module-level `Map`/`Set`/`WeakMap`, registry, or keyed cache added |
| `docs/cache.md` | RC-5 no duplicate mutable sources of truth | Pass | The fix removes a second reader of the preference from the view; `userModelPreference$` stays the single source, invalidated through its existing `reloadUserModelPreference$` |
| `.claude/skills/ccstate/SKILL.md` | DOM callback pattern — `detach(cmd(args, pageSignal), Reason.DomCallback)` | Pass | `setAsDefault` is exactly `detach(updateDefaultVideoModel(selection, pageSignal), Reason.DomCallback)` with `pageSignal` from `useGet(pageSignal$)` |
| `.claude/skills/ccstate/SKILL.md` | View layer uses loadables for state | Pass | `useLoadableSet(updateDefaultVideoModel$)` drives `disabled`/`aria-busy`/spinner across the whole reload-then-write span |
| `.claude/skills/ccstate/SKILL.md` | Reactive async computed over imperative fetch with manual state | Pass | The refresh goes through `reloadUserModelPreference$` invalidating the existing `userModelPreference$` computed rather than a hand-rolled fetch or local copy |
| `docs/testing.md` | Integration tests only; mock at the boundary | Pass | All cases drive the real page through `detachedSetupPage` and stub only HTTP via `context.mocks.api`; no `vi.mock()` of internal modules |
| `docs/testing/app-testing.md` | Use `waitFor()`, `context.mocks`, and test user flows | Pass | Assertions go through `waitFor`, mocks registered on `context.mocks`, no file-level `afterEach` cleanup |
| `docs/testing/testing-external-behavior.md` | Assert through the external user interface | Pass | Tests query visible notice text, `aria-pressed`, and captured request bodies; they never read composer signals directly |
| `docs/testing/anti-patterns.md` | AP-1/AP-4/AP-5/AP-7 — no mock-call assertions, internal mocks, fake timers, or implementation details | Pass | No `expect(mock).toHaveBeenCalled()` as a sole assertion, no `vi.useFakeTimers()`, no `../../` mock paths; the regression test asserts request payloads, which is external behavior |
| `docs/deployment-compatibility.md` | Frontend → backend requests stay compatible across versions | Pass | No contract change; the request shape is unchanged and `selectedVideoModel` keeps its existing own-property "absent = leave alone" semantics, so an old bundle still writing the default silently stays valid |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None. The P1 from the previous review is resolved by `a1c754d6`.

#### Resolved since `01d25e66`

- `ComposerTemporaryVideoModelNotice` no longer echoes the cached run model into the write. `updateDefaultVideoModel$` invalidates `userModelPreference$` and awaits a fresh read before composing the request, so a video write that follows a run-model write no longer resends the pre-write run model. Verified by disabling the reload and observing the new test fail with `Received "selectedModel": "claude-fable-5"` against the expected `claude-sonnet-4-6`.

#### Accepted residual (non-blocking)

- A second **Set as default** issued while the first `PUT` is still in flight can still read pre-write state, because the contract makes `selectedModel` required and the client therefore has to send *something*. Closing it completely means giving `selectedModel` the same own-property "absent = leave alone" treatment as `selectedVideoModel`, which would require the client to keep sending the field through the API-rollback window — i.e. adding a new cross-version fallback, which `docs/fallback.md` §2 says is not warranted for a non-GA switch. The current fix shrinks the window from "push not yet delivered" to "request still in flight", which is the right trade here. Worth revisiting if either switch moves toward GA.

### Testing

- Coverage: Adequate. The temporary-pick path, the exact write body (including the `selectedModel: null` case proving the workspace default is not pinned onto the member), Chat ↔ Video mode scoping, and the back-to-back write ordering are all covered; the switch-off silent-write path keeps its existing case.
- Conventions: Compliant. Integration-level, MSW-backed via `context.mocks.api`, `waitFor` for async assertions, no fake timers, no internal mocks, no mock-call-only assertions, no file-level cleanup hooks.
- Behavior gating: Compliant. New user-facing behavior sits behind `FeatureSwitchKey.NewChatDefaultModelAction` and `FeatureSwitchKey.VideoModelSelection`, both non-GA.
- Local verification on this commit: repo-wide `pnpm turbo run lint` 13/13, `pnpm check-types` 14/14, `knip --workspace apps/platform` clean, target suite 65/65, plus `src/signals/__tests__` and the template-picker suite 152/152.

Note (non-blocking, already declared in the PR summary): below 640px there is no video mode, so a mobile video pick is temporary with no composer affordance to persist it. Both gates are non-GA, so per `docs/fallback.md` §2 this is a scoped staff-only gap rather than a GA regression — but it should not stay open once either switch moves toward GA.

PR #27649, opened 2026-08-17
perf(python): bound connector-intent header work and retention

### LINKED ISSUE #27630: perf(python): bound connector-intent header work and retention
## Problem

At current `main` (`0eac48b359d85af21bb02209933acfe90d8c4d3b`), the request-header hook captures `X-VM0-Connector-Intent` before classification. `connector_intent.capture_and_strip()` calls `values[0].strip()` with no size bound, scans the entire normalized value for a comma, and stores every nonempty comma-free value in `flow.metadata`:

- `crates/runner/mitm-addon/src/mitm_addon.py:741-745`
- `crates/runner/mitm-addon/src/connector_intent.py:33-50`

Later, every request reaching connector routing checks that retained value against the omitted-connector identity sets. Python hashes the complete string for the membership operation:

- `crates/runner/mitm-addon/src/request_classification.py:429-440`

The header is private but untrusted: sandbox code controls its outbound request headers. The pinned mitmproxy 12.2.3 HTTP/1 request parser does not bound the request head before invoking addon hooks; its incomplete-head branch contains `FIXME: protect against header size DoS` (`mitmproxy/proxy/layers/http/_http1.py:320`). A comma-free oversized value therefore reaches this path.

The accepted identity domains are small and already bounded elsewhere:

- Built-in connector slugs are 1-64 lowercase alphanumeric/hyphen characters in `turbo/packages/api-contracts/src/contracts/connector-identity.ts:3-12`.
- Custom connector identities are UUIDs (`crates/runner/mitm-addon/src/registry_firewalls.py:27-39` and `turbo/packages/api-contracts/src/contracts/zero-custom-connectors.ts:126-128`).
- The MCP CLI obtains that UUID through the validated response schema and writes `connector.id` to this header in `turbo/apps/cli/src/commands/mcp/client.ts:99-105,182-200`.

Megabyte-scale values can never select a legitimate owner, but the proxy currently normalizes, scans, hashes, and retains them anyway.

## Measured evidence

I executed the exact current `capture_and_strip()` function body, extracted from the source AST, under the repository-pinned CPython 3.14.0. Each sample was a single comma-free value of `N` bytes plus trailing whitespace, followed by the exact routing set-membership shape. Medians over 15 fresh samples were:

| Header value | Capture/strip/store | Routing membership | Total synchronous work |
| --- | ---: | ---: | ---: |
| 1 MiB | 0.065 ms | 0.333 ms | 0.397 ms |
| 5 MiB | 0.375 ms | 1.662 ms | 2.037 ms |
| 16 MiB | 1.218 ms | 5.315 ms | 6.533 ms |
| 33 MiB | 3.948 ms | 10.963 ms | 14.911 ms |

`tracemalloc` around the exact capture function also showed that padded 1/5/16/33 MiB inputs leave correspondingly sized 1/5/16/33 MiB normalized strings retained in metadata after the original header is deleted. There is no connector-intent cleanup before the flow's terminal lifetime. Concurrent slow flows can therefore retain request-sized identity strings while repeated requests synchronously block the shared proxy event loop hashing values that cannot be valid.

## Suggested direction

Reject an oversized connector-intent value as malformed using an O(1) raw-length check before `strip()`, comma scanning, or metadata retention. Keep the limit aligned with the actual identity contracts (64-character built-in slugs and UUID custom identities), preserve unconditional header removal, and ensure only bounded values can reach routing membership or diagnostics.

This should remain a focused bound on the private identity field; a general mitmproxy request-header limit is a separate dependency/runtime concern.

## Acceptance criteria

- Oversized `X-VM0-Connector-Intent` values are classified as malformed before normalization, full-value delimiter scanning, hashing, or value retention.
- The raw oversized value is removed from request headers and is not stored or logged.
- Normal built-in connector slugs (including the 64-character boundary) and custom connector UUIDs retain current routing behavior.
- Repeated, comma-joined, empty, and whitespace-only values retain their current fail-closed semantics.
- Request-hook integration coverage proves that an oversized intent on an ambiguous route is denied before credential lookup.
- Structural tests assert bounded state/work without timing thresholds; a local benchmark confirms work and retained metadata no longer scale with invalid header length.

seven332:
## Summary

- preflight `X-VM0-Connector-Intent` against a 64-byte raw-field budget before mitmproxy decodes or normalizes it
- classify oversized values as malformed and strip every header occurrence through the raw-filtering multidict path
- preserve valid 64-byte built-in slugs, UUID custom connector IDs, and existing fail-closed malformed-input behavior
- cover the real ambiguous-route request hook with a 1 MiB decode guard, credential-fetch suppression, metadata bounds, and sanitized logs

## Compatibility

This is runner-local flow behavior with no API, queue, database, or persisted-state change. Valid connector identities behave identically across runner versions. During a rolling deployment, only invalid values over 64 raw bytes differ: new runners fail them closed without decoding or retaining them, while old runners may continue the prior behavior. Flows are owned by one runner and are not transferred.

This change intentionally does not bound mitmproxy's prior request-head buffering or work proportional to total header count/name size; those are separate parser/runtime concerns.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync basedpyright -p .` — 0 errors, 0 warnings
- `uv run --no-sync python -m pytest tests/test_request_handler_firewall_dispatch.py tests/test_request_headers_firewall_auth.py tests/test_request_handler_catalog_removal.py` — 79 passed
- `uv run --no-sync python -m pytest tests/` — 5580 passed, 59 existing warnings

Local capture-plus-routing benchmark, with each flow built before timing/tracing:

| Oversized value | Median | Median traced peak |
| --- | ---: | ---: |
| 1 MiB | 0.003083 ms | 229 bytes |
| 16 MiB | 0.003333 ms | 229 bytes |
| 33 MiB | 0.003458 ms | 229 bytes |

All samples were malformed, removed from request headers, and left no connector-intent value metadata.

Closes #27630

### DIFF
--- a/crates/runner/mitm-addon/src/connector_intent.py
+++ b/crates/runner/mitm-addon/src/connector_intent.py
@@ -7,6 +7,9 @@ from mitmproxy import http
 
 HEADER_NAME: Final = "X-VM0-Connector-Intent"
 
+_RAW_HEADER_NAME: Final = b"x-vm0-connector-intent"
+_MAX_CONNECTOR_INTENT_BYTES: Final = 64
+
 _VALUE_METADATA_KEY = "_connector_intent_value"
 _STATUS_METADATA_KEY = "_connector_intent_status"
 
@@ -33,21 +36,29 @@ MALFORMED = ConnectorIntent("malformed")
 def capture_and_strip(flow: http.HTTPFlow) -> None:
     """Capture connector intent once and always remove its private header."""
     if _STATUS_METADATA_KEY not in flow.metadata:
-        values = flow.request.headers.get_all(HEADER_NAME)
-        if not values:
+        raw_value: bytes | None = None
+        repeated = False
+        for name, value in flow.request.headers.fields:
+            if name.lower() != _RAW_HEADER_NAME:
+                continue
+            if raw_value is not None:
+                repeated = True
+                break
+            raw_value = value
+
+        if raw_value is None:
             flow.metadata[_STATUS_METADATA_KEY] = "absent"
-        elif len(values) != 1:
+        elif repeated or len(raw_value) > _MAX_CONNECTOR_INTENT_BYTES:
             flow.metadata[_STATUS_METADATA_KEY] = "malformed"
         else:
-            value = values[0].strip()
+            value = flow.request.headers.get_all(HEADER_NAME)[0].strip()
             if value == "" or "," in value:
                 flow.metadata[_STATUS_METADATA_KEY] = "malformed"
             else:
                 flow.metadata[_STATUS_METADATA_KEY] = "present"
                 flow.metadata[_VALUE_METADATA_KEY] = value
 
-    if HEADER_NAME in flow.request.headers:
-        del flow.request.headers[HEADER_NAME]
+    flow.request.headers.set_all(HEADER_NAME, [])
 
 
 def from_flow(flow: http.HTTPFlow) -> ConnectorIntent:
--- a/crates/runner/mitm-addon/tests/request_handler_helpers.py
+++ b/crates/runner/mitm-addon/tests/request_handler_helpers.py
@@ -54,7 +54,12 @@ def _single_firewall_vm(
     return vm_info
 
 
-def _shared_route_vm(tmp_path: Path, *, reverse: bool = False) -> dict[str, object]:
+def _shared_route_vm(
+    tmp_path: Path,
+    *,
+    reverse: bool = False,
+    primary_name: str = "primary",
+) -> dict[str, object]:
     """Build two credential owners for the same exact route."""
 
     def firewall(name: str, token: str, permission: str) -> dict[str, object]:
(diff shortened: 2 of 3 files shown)


seven332:
## Code Review: PR #27649

**Reviewed head:** `b8b5d810038e242635eed8d66db072e49ec6dfbc`

### Summary

The raw 64-byte preflight bounds connector-intent decode and retention while preserving valid-value behavior and fail-closed routing. The change is runner-local and introduces no cross-deployment contract.

### Findings

No findings.

### Testing

Reviewed the real request-hook integration coverage for both firewall orderings, the exact boundary, and a guarded 1 MiB value. Local validation passed Ruff formatting/lint, BasedPyright, 79 focused tests, the complete 5,580-test add-on suite, and bounded allocation/timing probes through 33 MiB. Unrelated Turbo and Rust suites were not run because the change is isolated to the Python mitm add-on.

### Verdict

LGTM

PR #27655, opened 2026-08-17
refactor(api): remove legacy-all connector scope

### LINKED ISSUE #27645: refactor(api): remove retired legacy-all connector scope
# Background

Run creation still accepts an omitted connector scope. When that happens outside a Zero-backed compose, the API labels the scope as `legacy_all` and interprets missing built-in and custom allowlists as access to every available connector.

This behavior was introduced for the former direct agent-run surface. That public surface and its deployed test routes have since been retired, while the active Zero run path already supplies explicit built-in and custom connector scopes.

# Problem

The remaining fallback is now test-fixture-driven compatibility code:

- an omitted internal field silently expands privileges to all available connectors;
- connector-loading code must preserve `undefined means all` branches instead of treating an empty list as empty;
- telemetry retains a source that production callers no longer need;
- tests can accidentally consume connectors without declaring that dependency.

Keeping this behavior makes connector materialization harder to reason about and blocks straightforward empty-scope fast paths.

# Decision

Remove the retired implicit-all scope instead of preserving it behind another default:

- require run-creation callers to provide an explicit connector scope;
- remove `legacy_all` from connector-scope telemetry and types;
- make built-in and custom allowlists concrete arrays, where an empty array means no connectors;
- remove the fallback that derives scope only because the caller omitted it;
- update test fixtures so connector-related scenarios declare the connectors they require, while unrelated runs use an explicit empty scope at the fixture boundary.

The production Zero path must retain its current agent-derived scope behavior. This change does not alter persisted data, queue payloads, or runner protocols.

# Acceptance criteria

- No production or test code references `legacy_all`.
- `CreateAgentRunArgs` cannot omit connector scope.
- Connector loaders no longer interpret an undefined allowlist as all connectors.
- Zero run creation continues to pass and enforce the agent's built-in and custom connector allowlists.
- Direct-run test fixtures do not implicitly expose all connectors; connector-dependent tests opt into the exact required scope.
- Relevant API integration tests, lint, type checking, formatting, and static analysis pass.

# Evidence and related work

- The direct agent-run API was removed in #23985 / PR #24013.
- Deployed direct-run test routes were removed in PR #25550.
- The active Zero run creation path already supplies an explicit connector scope.
- #27454 benefits from concrete empty scopes but remains a separate performance investigation.

# Out of scope

- Database cleanup or migrations; `legacy_all` is not persisted.
- Queue or runner protocol changes.
- Changes to connector permission semantics for active Zero runs.

seven332:
## Summary

- require every internal agent-run caller to provide a concrete connector scope
- remove the retired `legacy_all` fallback and its Zero-backed compose lookup
- make direct-run test fixtures fail closed by default and declare connector access explicitly where needed
- remove fallback-only tests and telemetry that no longer represent production behavior

## Why

Production Zero runs already resolve connector permissions before entering the lower agent-run service. The optional scope kept an obsolete direct-run compatibility path alive, where a missing scope meant access to every stored connector. Requiring a concrete scope removes that ambiguous and unsafe state.

## Validation

- `pnpm -F api lint`
- `pnpm -F api check-types`
- `pnpm -F api exec vitest run src/signals/routes/__tests__/run-lifecycle.bdd.test.ts --maxWorkers=4 --silent=passed-only` (153 tests)
- `pnpm -F api exec vitest run --maxWorkers=4 --silent=passed-only` (3018 tests)
- `pnpm knip`
- pre-commit hooks

Closes #27645

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
@@ -286,7 +286,6 @@ const API_DISPATCH_TIMING_ACTION_TYPES = [
   "api_dispatch_prepare_context_load_persisted_environment",
   "api_dispatch_prepare_context_build_resolved_body",
   "api_dispatch_prepare_context_resolve_framework",
-  "api_dispatch_prepare_context_resolve_connector_scope",
   "api_dispatch_prepare_context_resolve_model_provider",
   "api_dispatch_prepare_context_load_connector_contexts",
   "api_dispatch_prepare_context_load_stored_connectors",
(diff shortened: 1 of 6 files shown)


seven332:
## Code Review: PR #27655

**Reviewed head:** `f1ddc48f2de3343cc7e772766f1fb5f0c3fa9bcd`

### Summary

The internal agent-run boundary now requires a concrete connector scope, and the retired permissive fallback, lookup, tests, and telemetry are removed. Production Zero callers already provide the resolved scope, while test-only direct runs fail closed by default.

### Findings

No findings.

### Testing

API lint, type checking, Knip, 153 lifecycle integration tests, and the complete 3018-test API suite passed locally. Type checking and the lifecycle suite also passed after rebasing onto the latest `main`. GitHub CI is still in progress.

### Verdict

LGTM

PR #27709, opened 2026-08-17
fix(ci): mask neon branch database credentials

### LINKED ISSUE #27701: fix(ci): stop logging neon database credentials
## Problem and evidence

The reusable Neon branch composite action resolves a PostgreSQL connection URI that contains a dynamically generated role password. Its `create` path currently prints the complete value:

```bash
echo "DATABASE_URL value: $DATABASE_URL"
```

GitHub documents database connection strings as values that Actions may automatically redact, but also warns that automatic redaction is not guaranteed and directs workflows to register generated sensitive values with `::add-mask::` before they can reach logs or workflow commands. This action currently relies on implicit pattern recognition instead of explicitly protecting the credential it generates.

The checked-in repository currently calls this composite action only with `action: cleanup`, so the affected `create` path is dormant in current callers. The action still declares `create` as its default supported operation and exposes `database-url` and `console-url` outputs; reusing that supported path would unnecessarily emit credential material and depend on best-effort redaction.

Relevant GitHub guidance:

- https://docs.github.com/en/actions/reference/security/secrets#automatically-redacted-secrets
- https://docs.github.com/en/actions/reference/security/secure-use#use-secrets-for-sensitive-information
- https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#masking-a-value-in-a-log

## Decision

Proceed with a focused hardening change. Remove the credential-bearing diagnostic line and explicitly register the generated URL with GitHub Actions masking before writing the existing step output.

The issue is preventive rather than evidence of a currently active caller leaking an unredacted credential. That distinction does not justify retaining an unnecessary plaintext log statement or relying on non-guaranteed automatic detection.

- Remove `echo "DATABASE_URL value: $DATABASE_URL"`.
- Emit `echo "::add-mask::$DATABASE_URL"` immediately after URL resolution and before any workflow command receives the value.
- Preserve the existing `database-url` and `console-url` output names and values, branch lifecycle behavior, and migration behavior.
- Add a focused repository contract test that verifies the action neither prints the URL nor writes it to `$GITHUB_OUTPUT` before registering the mask.
- Audit repository-owned Neon URL generation paths for the same explicit plaintext diagnostic pattern. Keep unrelated connection routing, pooling, TLS, and credential lifecycle changes out of scope.

This follows the established pattern in `.github/actions/production-migration-smoke/action.yml` and protects the value even if a later same-job step accidentally logs it.

## Alternatives considered

- **Keep the log and rely on automatic database-URL redaction:** rejected because GitHub explicitly describes redaction as non-guaranteed and recommends masking generated sensitive values.
- **Only remove the log line:** improves the immediate path but leaves a generated credential unregistered if it is later printed by a consumer or error path.
- **Mask the value but retain the diagnostic:** the resulting `***` line provides no useful operational information, so retaining it adds noise without observability value.
- **Remove the dormant `create` capability:** not selected because the action and the in-flight Neon TLS work intentionally preserve that supported contract; removing it is a separate lifecycle decision with a larger compatibility impact.

## Risks and mitigations

- **Output compatibility:** preserve the output key and underlying URL unchanged; masking affects log rendering, not the same-job step output contract. The production migration smoke action already uses this pattern.
- **Brittle regression coverage:** test the ordering and absence requirements at the composite-action boundary instead of matching an entire shell block.
- **Scope creep:** do not combine this hardening with connection mode, TLS, branch retention, role rotation, or historical incident analysis.

## Acceptance criteria

- The Neon branch action contains no command that prints the generated database connection URI.
- The generated URL is registered with `::add-mask::` after both pooled and fallback resolution paths and before it is written to `$GITHUB_OUTPUT`.
- Existing `database-url` and `console-url` outputs remain unchanged.
- A focused repository check fails if plaintext URL logging returns, masking is removed, or output is written before masking.
- The audit finds no other repository-owned Neon generator that explicitly prints a generated connection URI; any materially different finding is tracked separately rather than folded into connection-routing work.
- Focused shell test, Bash syntax validation, ShellCheck, workflow shell validation, YAML parsing, and Actionlint pass.

## Delivery notes

- Standalone, single-PR issue with no blockers or parent relationship.
- Coordinate with #27700 because it modifies nearby Neon connection-string arguments in the same action. Rebase after #27700 merges rather than duplicating its TLS changes.
- Related historical design discussion: #26999.

seven332:
## Summary

- remove the Neon branch action's generated database URL diagnostic
- escape the URL for GitHub's workflow-command data protocol, then register the exact original value for masking before preserving the existing output
- add a focused Security workflow contract test for raw credential emission, command-data escaping, and mask-before-output ordering

## Behavior

The composite action still resolves the same pooled or fallback URL, writes the unchanged `database-url` output, exposes the same `console-url`, and retains its existing branch, migration, and deployment behavior. The generated database URL is now explicitly masked before it can reach later same-job logs.

Before issuing `add-mask`, the action escapes percent, carriage-return, and newline characters in the command payload using GitHub Actions' workflow-command protocol. This ensures a percent-encoded database URL is decoded by the runner back to the exact original value registered for masking. The regression test permits only the existing raw URL output command, rejects direct or braced raw URL logging, and requires command-data escaping, masking, and output in that order.

## Scope

- no Neon connection routing, pooling, TLS, branch retention, or credential rotation changes
- no API, schema, persisted data, runner, frontend, or deployment payload changes
- no historical incident conclusion; this is preventive hardening of the supported `create` path

## Validation

- `bash .github/scripts/tests/neon-branch-credential-mask-test.sh`
- `bash -n .github/scripts/tests/neon-branch-credential-mask-test.sh`
- `shellcheck .github/scripts/tests/neon-branch-credential-mask-test.sh` (ShellCheck 0.9.0)
- `bash .github/scripts/tests/neon-connection-string-ssl-test.sh`
- synthetic Bash assertion for `%25`, carriage-return, and newline command-data escaping
- `yq 4.47.2 eval '.' .github/actions/neon-branch/action.yml .github/workflows/security.yml`
- `.github/scripts/check-workflow-shell-compatibility.sh .github/workflows/security.yml`
- Actionlint 1.7.12
- `git diff --check`, including the new file

Privileged live Neon execution was not run locally because no checked-in workflow currently invokes the action's `create` path.

Closes #27701

### DIFF
--- a/.github/actions/neon-branch/action.yml
+++ b/.github/actions/neon-branch/action.yml
@@ -103,7 +103,10 @@ runs:
           echo "Attempting without --pooled flag..."
           DATABASE_URL=$(neonctl connection-string "$BRANCH_NAME" --project-id "$NEON_PROJECT_ID" --database-name "$INPUT_DATABASE_NAME" --role-name "$INPUT_ROLE_NAME" --ssl verify-full)
         fi
-        echo "::add-mask::$DATABASE_URL"
+        WORKFLOW_COMMAND_DATA="${DATABASE_URL//%/%25}"
+        WORKFLOW_COMMAND_DATA="${WORKFLOW_COMMAND_DATA//$'\r'/%0D}"
+        WORKFLOW_COMMAND_DATA="${WORKFLOW_COMMAND_DATA//$'\n'/%0A}"
+        printf '::add-mask::%s\n' "$WORKFLOW_COMMAND_DATA"
         echo "database-url=$DATABASE_URL" >> "$GITHUB_OUTPUT"
         echo "Neon branch ready: $BRANCH_NAME"
 
--- a/.github/actions/neon-branch/action.yml
+++ b/.github/actions/neon-branch/action.yml
@@ -103,8 +103,8 @@ runs:
           echo "Attempting without --pooled flag..."
           DATABASE_URL=$(neonctl connection-string "$BRANCH_NAME" --project-id "$NEON_PROJECT_ID" --database-name "$INPUT_DATABASE_NAME" --role-name "$INPUT_ROLE_NAME" --ssl verify-full)
         fi
-        echo "DATABASE_URL value: $DATABASE_URL"
-        echo "database-url=$DATABASE_URL" >> $GITHUB_OUTPUT
+        echo "::add-mask::$DATABASE_URL"
+        echo "database-url=$DATABASE_URL" >> "$GITHUB_OUTPUT"
         echo "Neon branch ready: $BRANCH_NAME"
 
         # Get branch ID for console URL
--- a/.github/scripts/tests/neon-branch-credential-mask-test.sh
+++ b/.github/scripts/tests/neon-branch-credential-mask-test.sh
@@ -13,7 +13,7 @@ if grep -Fq 'DATABASE_URL value:' "$action"; then
   fail "Neon branch action logs the generated database URL"
 fi
 
-mapfile -t database_url_emissions < <(
+mapfile -t raw_database_url_emissions < <(
   awk '
     /^[[:space:]]*(echo|printf)[[:space:]]/ && index($0, "$DATABASE_URL") > 0 {
       sub(/^[[:space:]]+/, "")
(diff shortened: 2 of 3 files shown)


seven332:
## Code Review: PR #27709

**Reviewed head:** `7bc166cc8b7bd705220299e1fd98415e9dfc05dc`

### Summary

The PR removes the reusable Neon action's credential-bearing diagnostic, explicitly masks the generated URL before preserving the existing same-job output, and adds a focused Security CI contract test for allowed emissions and ordering. Branch lifecycle, TLS arguments, migrations, deployment tracking, and both public outputs remain unchanged.

### Findings

No findings.

The separate pre-existing masking gap in active Turbo/release workflow paths is outside this PR and is tracked in #27714.

### Testing

The submitted head passed the focused masking and Neon TLS contract tests, Bash syntax validation, ShellCheck, YAML parsing, workflow shell compatibility validation, Actionlint 1.7.12, diff checks, and repository commit hooks. Privileged live Neon create execution was not run locally because no checked-in workflow currently invokes that path.

### Verdict

LGTM

seven332:
## Code Review: PR #27709

**Reviewed head:** `d1fda7a3c4577594fb3f01b84d9751cb9b51c078`

### Summary

The PR removes the reusable Neon action's credential-bearing diagnostic, escapes workflow-command data before registering the exact generated URL for masking, preserves the existing raw step output, and adds a Security CI contract that rejects direct or braced raw logging and enforces escape-mask-output ordering. Branch lifecycle, connection routing, pooling, TLS arguments, migrations, deployment tracking, and both public outputs remain unchanged.

### Findings

No unresolved findings.

Self-review identified and fixed the percent-encoded command-data mismatch and two ordering/emission test gaps before this verdict. Separate pre-existing work remains outside this PR: active Turbo/release workflow masking is tracked in #27714, and the Runner deadline coverage flake observed on an earlier head is tracked in #27722.

### Testing

The reviewed head passed the focused masking and Neon TLS contract tests, Bash syntax validation, ShellCheck, a synthetic `%25`/CR/LF command-data assertion, YAML parsing, workflow shell compatibility validation, Actionlint 1.7.12, diff checks, and repository commit hooks. Privileged live Neon create execution was not run locally because no checked-in workflow currently invokes that path.

### Verdict

LGTM

PR #27725, opened 2026-08-17
fix(python): bound codex prefetch marker parsing

### LINKED ISSUE #27717: security(python): bound Codex prefetch marker parsing
## Problem

At current main (27bdf421b2df3ed07361dc0ec6fe30546a6a3d10), the runner-only Codex catalog prefetch marker is parsed from an untrusted request header without a raw size or cardinality bound.

- crates/runner/mitm-addon/src/codex_model_catalog_cache.py:550-557 calls Headers.get_all for X-VM0-Codex-Model-Catalog-Prefetch. That decodes every complete matching value into Python strings before the code checks whether the result is exactly ["1"] and removes the header.
- crates/runner/mitm-addon/src/mitm_addon.py:743-747 invokes this helper before registry lookup or request classification in requestheaders.
- crates/runner/mitm-addon/src/mitm_addon.py:1281-1293 invokes it again at request phase, although a header seen in requestheaders has normally already been removed.
- crates/runner/mitm-addon/src/connector_intent.py:36-61 already demonstrates the bounded pattern for another runner-private header: inspect raw header fields, reject over-budget or repeated values before decoded access, and remove the header unconditionally.

crates/runner/mitm-addon/pyproject.toml:20 pins mitmproxy 12.2.3. Existing repository records #27629 and #27630 document that this version does not cap an HTTP/1 request head before addon hooks. The only valid prefetch marker is one ASCII byte, so request-sized values cannot have a legitimate meaning.

## Violated invariant

A fixed runner-private marker must validate raw cardinality and size before decoding or retaining client-controlled header values. Invalid values must still be stripped, but they should not create request-sized Python strings or lists on the shared proxy event loop.

## Impact

Any flow reaching the addon header hook can carry an oversized prefetch-marker value. The marker is ignored semantically, but only after complete string decoding and list construction have consumed work and transient memory. Repeated malformed requests can delay other proxy flows before VM registration, authority validation, firewall policy, or credential handling is reached.

This finding is source-only; no project code, benchmark, or service was executed.

## Acceptance criteria

- Inspect X-VM0-Codex-Model-Catalog-Prefetch through raw header fields before any decoded get or get_all operation.
- Accept exactly one value with the exact supported one-byte representation.
- Treat repeated, empty, non-ASCII, and over-budget values as non-prefetch without decoding the complete values.
- Remove every instance of the private marker regardless of validity.
- Preserve idempotent requestheaders and request hook behavior.
- Add structural coverage proving that an oversized raw value is not decoded, plus cases for the exact valid marker and repeated values.
- Do not use wall-clock assertions or broaden this change into a general request-head policy.

seven332:
## Summary

- inspect the runner-private Codex catalog prefetch marker through raw mitmproxy header fields
- accept only one exact `b"1"` value and strip every marker occurrence without decoded access
- preserve prefetch metadata across the normal `requestheaders()` to `request()` lifecycle
- structurally cover exact, repeated, empty, non-ASCII, and 1 MiB marker values through real hook integration

## Security and compatibility

This removes request-sized UTF-8 decoding, list construction, and the additional decoded lookup/folding previously performed by `Headers.pop()` on the shared proxy event loop. Valid runner-generated traffic is unchanged; invalid marker forms remain stripped, non-fatal, and ordinary identity requests.

The marker and cache metadata are runner-local, so old and new runners can overlap without a cross-version protocol or persisted-state change. This PR intentionally does not introduce a proxy-wide request-head limit or address work proportional to aggregate header count, unrelated fields, or mitmproxy's prior parser buffering.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync basedpyright -p .` — 0 errors, 0 warnings
- `uv run --no-sync python -m pytest tests/test_codex_model_catalog_cache_hooks.py` — 24 passed
- `uv run --no-sync python -m pytest tests/` — 5,587 passed, 59 existing warnings
- `git diff --check`
- staged-file pre-commit hook — Ruff and file-size checks passed
- commit message hook — commitlint passed

Closes #27717

### DIFF
--- a/crates/runner/mitm-addon/src/codex_model_catalog_cache.py
+++ b/crates/runner/mitm-addon/src/codex_model_catalog_cache.py
@@ -103,6 +103,7 @@ _FLOW_STATE = "_codex_model_catalog_cache_state"
 _FLOW_TELEMETRY = "_codex_model_catalog_cache_telemetry"
 _PREFETCH_REQUEST = "_codex_model_catalog_prefetch_request"
 _PREFETCH_HEADER = "X-VM0-Codex-Model-Catalog-Prefetch"
+_RAW_PREFETCH_HEADER = b"x-vm0-codex-model-catalog-prefetch"
 _BROTLI_ENCODING = "br"
 _IDENTITY_ENCODING = "identity"
 _REQUEST_CONDITIONAL_HEADERS = (
@@ -549,11 +550,20 @@ def _request_bypass_reason(
 
 def capture_and_strip_prefetch_marker(flow: http.HTTPFlow) -> None:
     """Capture the runner-only prefetch marker without forwarding it upstream."""
-    values = flow.request.headers.get_all(_PREFETCH_HEADER)
-    if not values:
+    raw_value: bytes | None = None
+    repeated = False
+    for name, value in flow.request.headers.fields:
+        if name.lower() != _RAW_PREFETCH_HEADER:
+            continue
+        if raw_value is not None:
+            repeated = True
+            break
+        raw_value = value
+
+    if raw_value is None:
         return
-    flow.request.headers.pop(_PREFETCH_HEADER, None)
-    if values == ["1"]:
+    flow.request.headers.set_all(_PREFETCH_HEADER, [])
+    if not repeated and len(raw_value) == 1 and raw_value == b"1":
         flow.metadata[_PREFETCH_REQUEST] = True
 
 
--- a/crates/runner/mitm-addon/tests/test_codex_model_catalog_cache_hooks.py
+++ b/crates/runner/mitm-addon/tests/test_codex_model_catalog_cache_hooks.py
@@ -8,7 +8,7 @@ from typing import Literal
 from unittest.mock import patch
 
 import pytest
-from mitmproxy import connection
+from mitmproxy import connection, http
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #27725

**Reviewed head:** `e925e266be102c34b22c62ce4ccc646b7857ac95`

### Summary

The PR replaces decoded marker lookup/removal with bounded raw cardinality, length, and exact-value inspection, while preserving unconditional stripping and the existing two-hook metadata lifecycle. The approach matches the addon's established bounded private-header pattern and keeps broader request-head policy out of this focused change.

### Findings

No findings. The six focused passes found no correctness, test, concurrency, performance, resource, security, compatibility, or maintainability issue.

### Testing

- Focused Codex catalog hook suite: 24 passed.
- Complete mitm-addon suite: 5,587 passed with 59 existing warnings.
- Decode-guard integration covers exact, repeated, empty, non-ASCII, and 1 MiB values across `requestheaders()` and `request()`.
- A local allocation probe remained at a 233-byte traced peak for preconstructed 1, 16, and 33 MiB invalid values.
- No deployed-runner smoke was run; no in-scope behavior remains unverified by the repository test suite.

### Verdict

LGTM

PR #27741, opened 2026-08-17
perf(python): share websocket server event parsing

### LINKED ISSUE #27720: perf(python): avoid duplicate terminal WebSocket JSON scans
## Problem

At `27bdf421b2df3ed07361dc0ec6fe30546a6a3d10`, every terminal OpenAI Responses WebSocket frame can be fully scanned twice on mitmproxy's synchronous WebSocket hook while prewarm correlation remains non-ambiguous.

- `crates/runner/mitm-addon/src/mitm_addon.py:1567-1587` inspects the frame prefix once and passes one retained `OpenAIResponsesEvent` to the model WebSocket usage lifecycle.
- `crates/runner/mitm-addon/src/model_websocket_usage.py:220-237` first calls `inspect_openai_responses_server_lifecycle()` for lifecycle event types and then unconditionally calls `extract_openai_responses_usage_from_event()` on the same event.
- `crates/runner/mitm-addon/src/usage/openai_responses.py:277-318` constructs a `JsonSelectiveExtractor`, feeds the complete body, and validates `type` plus `response.id` for lifecycle correlation.
- `crates/runner/mitm-addon/src/usage/openai_responses.py:903-950` constructs a second `JsonSelectiveExtractor`, feeds the same complete body again, and extracts the overlapping `type`, `response.id`, model, and usage fields.

`activate()` installs a non-ambiguous prewarm state on every tracked Responses WebSocket flow, and terminal event types are explicitly included in `_WEBSOCKET_LIFECYCLE_EVENT_TYPES`. The duplicate scan therefore applies to ordinary terminal usage frames, not only to a rare `generate:false` response.

Each parser has its own 65,536-work-unit allowance. The complete hook path can consequently spend nearly twice the parser budget even though the lifecycle fields are a subset of the usage extraction fields. This regresses the one-work-budgeted-full-scan boundary established by #23729.

## Evidence

An isolated CPython 3.12.3 benchmark constructed valid `response.completed` frames with an early `type`, a dense ignored scalar array, and a terminal response containing a consistent ID, model, and usage. It inspected the prefix once, then timed the two current full-body functions over 15 calls and reported the median:

| Frame | Lifecycle scan | Usage scan | Current sequential path |
| --- | ---: | ---: | ---: |
| 40,132 bytes / 20,000 ignored scalars | 72.899 ms | 71.724 ms | 143.219 ms |
| 60,132 bytes / 30,000 ignored scalars | 107.494 ms | 107.365 ms | 214.726 ms |

Both parsers completed within their work limits: lifecycle returned a valid terminal response ID and usage returned the expected input/output quantities. A separate bulk-string check showed the same duplication by bytes: a 5,000,133-byte valid terminal frame took about 17.9 ms per full scan and 35.9 ms for the current sequential pair.

These are isolated local measurements, not end-to-end latency claims. They demonstrate that the second pass almost exactly doubles synchronous parser work. The accepted decoded WebSocket message limit is 16 MiB, so the observed effect occurs far below the framing limit.

## Impact

Terminal Responses frames can contain large response objects, structured output, or tool-related data influenced by the proxied workload. Dense valid frames well below the size limit can occupy mitmproxy's shared event loop for hundreds of milliseconds because correlation and usage independently walk the same JSON document. During that time, unrelated proxy flows and WebSocket messages cannot progress on that hook path.

The work cap still prevents unbounded individual scans, but independent caps do not preserve a bounded complete path: a near-limit lifecycle scan is immediately followed by another near-limit usage scan. On exhaustion, the same body can also consume both budgets before the frame is discarded and correlation fails open.

## Suggested direction

Share one authoritative selective parse for terminal/unknown server frames and derive both lifecycle correlation and usage extraction from that result. The selected field set can include the lifecycle consistency paths together with the existing usage fields, while keeping the 4 KiB prefix classification and known-non-usage fast path. The complete correlation-plus-usage path should consume one full-body work budget, not one budget per consumer.

## Acceptance criteria

- A terminal or unresolved OpenAI Responses WebSocket server frame performs at most one work-budgeted full-body selective JSON parse before both lifecycle correlation and usage settlement complete.
- The complete correlation-plus-usage path has one 65,536-unit work bound; parser exhaustion cannot trigger a second full-body attempt on the same frame.
- Preserve exact duplicate-key consistency checks for top-level `type` and `response.id`.
- Preserve serialized prewarm suppression, sticky fail-open ambiguity, duplicate-terminal idempotency, source-preserving usage reporting, and current correlation-cap/usage-error diagnostics.
- Preserve the existing prefix probe and immediate known-non-usage fast path from #17255/#23198.
- Add deterministic coverage that counts full extractor passes or shared parser feeds without relying on wall-clock assertions, including valid dense terminal input and work-limit exhaustion.
- Run the focused WebSocket prewarm, Responses event JSON, usage aggregation, and output-timing tests plus the standard mitm-addon format, lint, type, and test checks.

## Audit context

Found during the daily mitm-addon Python performance audit on commit `27bdf421b2df3ed07361dc0ec6fe30546a6a3d10`.

seven332:
## Summary

- replace separate OpenAI Responses WebSocket lifecycle and usage scans with one combined bounded server-event inspection
- preserve lifecycle duplicate-key validation, usage extraction semantics, prewarm fail-open behavior, diagnostics, and prefix/known-non-usage fast paths
- add hook-level regression coverage proving one target-body extractor feed for valid dense and work-limit-exhausted terminal frames

## Exclusions

- no change to the 65,536-unit parser budget or 4 KiB prefix probe
- no persisted state, runner protocol, API, database, or cross-version compatibility changes

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest -q tests/test_openai_responses_event_json.py tests/test_model_provider_websocket_prewarm.py tests/test_model_provider_websocket_source_reporting.py tests/test_model_provider_websocket_usage_aggregation.py tests/test_model_provider_websocket_lifecycle.py tests/test_model_provider_websocket_metadata.py tests/test_codex_output_timing.py tests/test_provider_output_timing.py` (111 passed)
- `uv run --no-sync python -m pytest -q tests/` (5588 passed)

Closes #27720

### DIFF
--- a/crates/runner/mitm-addon/src/model_websocket_usage.py
+++ b/crates/runner/mitm-addon/src/model_websocket_usage.py
@@ -218,7 +218,6 @@ def feed_usage(
     if not is_enabled(flow):
         return
     prewarm_state = flow.metadata.get(_MODEL_WEBSOCKET_PREWARM_STATE)
-    lifecycle: usage.OpenAIResponsesServerLifecycle | None = None
     should_inspect_lifecycle = (
         isinstance(prewarm_state, _OpenAIResponsesPrewarmState)
         and not prewarm_state.ambiguous
@@ -230,11 +229,16 @@ def feed_usage(
             or event.event_type in _WEBSOCKET_LIFECYCLE_EVENT_TYPES
         )
     )
-    if should_inspect_lifecycle and isinstance(prewarm_state, _OpenAIResponsesPrewarmState):
-        lifecycle = usage.inspect_openai_responses_server_lifecycle(event)
+    inspection = usage.inspect_openai_responses_server_event(
+        event,
+        include_lifecycle=should_inspect_lifecycle,
+    )
+    lifecycle = inspection.lifecycle
+    if lifecycle is not None and isinstance(prewarm_state, _OpenAIResponsesPrewarmState):
         _observe_server_lifecycle(flow, prewarm_state, lifecycle)
 
-    usage_result, inspection_error = usage.extract_openai_responses_usage_from_event(event)
+    usage_result = inspection.usage
+    inspection_error = inspection.usage_error
     if inspection_error is not None:
         if isinstance(prewarm_state, _OpenAIResponsesPrewarmState):
             _mark_correlation_ambiguous(flow, prewarm_state, "correlation_cap")
--- a/crates/runner/mitm-addon/src/usage/__init__.py
+++ b/crates/runner/mitm-addon/src/usage/__init__.py
@@ -68,6 +68,7 @@ from .openai_chat_completions import (
 from .openai_responses import (
     OpenAIResponsesClientEvent,
     OpenAIResponsesEvent,
+    OpenAIResponsesServerEventInspection,
     OpenAIResponsesServerLifecycle,
     create_openai_responses_json_usage_extractor,
     create_openai_responses_sse_usage_extractor,
@@ -75,7 +76,7 @@ from .openai_responses import (
     extract_openai_responses_usage_with_error_from_json,
     inspect_openai_responses_client_event_json,
     inspect_openai_responses_event_json,
-    inspect_openai_responses_server_lifecycle,
+    inspect_openai_responses_server_event,
     merge_openai_responses_usage_result,
 )
 from .providers.connectors import (
@@ -101,6 +102,7 @@ __all__ = [
     "ModelUsageProtocol",
     "OpenAIResponsesClientEvent",
     "OpenAIResponsesEvent",
+    "OpenAIResponsesServerEventInspection",
     "OpenAIResponsesServerLifecycle",
     "admit_buffered_report",
     "buffer_model_usage_observations",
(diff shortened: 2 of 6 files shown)


seven332:
## Code Review: PR #27741

**Reviewed head:** `fc9661577004d11aafbc21070250240a72bd74ba`

### Summary

The combined server-event inspection gives lifecycle correlation and usage settlement one bounded full-body parse while preserving prefix and known-non-usage fast paths, duplicate identity validation, fail-open correlation, and diagnostics.

### Findings

No findings.

### Testing

Reviewed hook-level valid-dense and work-limit-exhaustion feed-count regressions plus existing correlation and usage behavior. Local validation passed uv lock/sync, Ruff, basedpyright, 111 focused tests, and 5,588 full-addon tests. Unverified: production latency/frequency; the change is parser-work bounded and not an end-to-end latency claim.

### Verdict

LGTM

PR #27743, opened 2026-08-17
fix(python): bound request host authority work

### LINKED ISSUE #27713: security(python): bound request authority hostname normalization
## Problem

On current main (`44934b36f1`; the request-authority path is unchanged since the measured `085dbdb365` revision), registered sandbox requests can make the trusted request-authority path perform work and allocations proportional to an oversized HTTP Host field before the request is rejected.

For HTTPS requests, `crates/runner/mitm-addon/src/request_authority.py` currently:

- calls `Headers.get_all("Host")` and `request.host_header`, decoding and potentially folding every raw Host value before any value-size budget;
- scans, splits, normalizes, materializes, and rejoins every hostname label before comparing the result with trusted SNI;
- retains the decoded Host diagnostic in an error object that is written to the 403 response and proxy log;
- obtains `host_header` before the non-HTTPS early return, so even HTTP requests decode a Host value that is not used for authority trust.

The pinned mitmproxy `12.2.3` HTTP/1 layer still contains `FIXME: protect against header size DoS` at its incomplete request-head buffering sites. It retains and strips an unbounded ordinary Host field before addon hooks. Issues #27629 and #27630 addressed analogous addon work for User-Agent and connector-intent values, but request Host processing remains unbounded inside the addon.

### Reproduction

A local reproduction at `085dbdb365` used the real pinned `mitmproxy.http.Headers`, a real `HTTPFlow`, and `get_trusted_authority()`. Each Host consisted of individually valid one-character labels, was constructed outside the measured region, and was ultimately rejected only as `authority_mismatch` against `api.github.com`.

| Raw Host bytes | Elapsed without tracing | Peak traced allocation |
| ---: | ---: | ---: |
| 65,535 | 15.2 ms | 2.0 MiB |
| 1,048,575 | 631.1 ms | 33.0 MiB |
| 8,388,607 | 4,737.3 ms | 258.3 MiB |

Absolute timings are host-dependent, and tracing materially increases runtime. The measurements establish that synchronous work and transient allocation scale with invalid Host data. The header-phase prebind probe can perform this work once and the later request hook can perform it again for the same denial, on mitmproxy's shared event-loop thread and before firewall matching or credential resolution.

## Confirmed boundary

The resource boundary belongs at ordinary raw Host fields in `Headers.fields`, before dependency string access or project hostname normalization.

Other authority inputs are already bounded on real wire paths in pinned mitmproxy:

- HTTP/1 absolute-form and CONNECT targets call `url.parse_authority(..., check=True)` before constructing `Request`.
- HTTP/2 extracts `:authority` and calls the same checked parser; HTTP/3 reuses that parser.
- The checked authority parser rejects host bytes over 255. The remaining valid zero-padded port representation is bounded by CPython's decimal integer parser.
- TLS `ClientHello.sni` calls `check.is_valid_host()`, rejects host bytes over 255, and decodes accepted input as ASCII before setting `client_conn.sni`.

Project source does not mutate request authority between parsing and classification. Adding an addon limit to `Request.data.authority` or SNI would therefore protect only synthetic or hypothetical future states and is outside this issue under the project's avoid-defensive-programming rule. Existing absolute-form, HTTP/2/3 authority, `missing_sni`, and `invalid_sni` behavior must remain unchanged.

A separate label-count or total normalized-hostname policy is also unnecessary once raw Host work has a small fixed budget. Such a rule would create a new request/configuration compatibility policy without improving this resource bound.

## Decision

Proceed with one focused runner-addon change:

1. Return from non-HTTPS authority resolution before reading Host text.
2. For HTTPS, inspect regular Host fields from raw bytes and enforce a 4,096-byte inclusive aggregate budget before `get_all()`, `host_header`, folding, retention, or hostname normalization.
3. Count duplicate Host values as the dependency would fold them, including separators, so bounded diagnostics cannot hide aggregate decode/list/fold work.
4. Decode and normalize only bounded values through the pinned dependency's current semantics.
5. Preserve the current folded `host_header` diagnostic for duplicate fields whose aggregate remains within budget. Use no decoded Host diagnostic for an over-budget aggregate.
6. Reject over-budget Host input through the existing `invalid_authority` path before firewall matching, credential resolution, or forwarding.

The 4,096-byte raw-value ceiling matches the existing browser work budget, is sixteen times the pinned 255-byte production SNI host ceiling, leaves substantial multibyte and port margin, and keeps accepted synchronous work in low kilobytes. No supported producer for a longer Host value was found. This is an addon value-work limit, not a DNS presentation rule or a proxy-wide request-head limit.

## Alternatives considered

- **Guard decoded strings:** too late; dependency decoding and folding have already allocated proportional state.
- **Also guard `Request.data.authority`:** every production wire parser already validates it, and project code does not mutate it. This adds a hypothetical branch and can reject parser-accepted zero-padded ports without addressing the demonstrated path.
- **Add a limit only to `normalize_hostname()`:** too late for Host decode/fold and changes configuration-side callers unrelated to sandbox request input.
- **Add label-count or normalized-length limits:** adds no work-bound benefit after raw preflight and creates an unproven hostname policy.
- **Patch mitmproxy with a total request-head limit:** addresses parser buffering and all fields, but requires broader HTTP/1, HTTP/2, HTTP/3, response, compatibility, and dependency work. It remains valuable separate hardening, not a substitute for bounding this addon consumer.

## Risks and residual exposure

- Raw preflight couples the addon to documented pinned mitmproxy `Headers.fields` behavior.
- The 4,096-byte budget rejects extreme zero-padded ports or exotic Unicode spellings that previously normalized successfully. Exact boundary tests make that raw-byte policy explicit.
- Old and new runners may differ for newly rejected invalid Host values while deployments overlap. Flows are runner-owned, so no persisted-state or protocol migration is required.
- The fix does not prevent mitmproxy from buffering/stripping the request head before hooks, and raw scans still scale with total field count/name length. A proxy-wide request-head limit remains residual work.

## Acceptance criteria

- Non-HTTPS authority resolution does not decode or inspect Host text that it does not trust or use.
- HTTPS regular Host fields are checked from raw bytes against a 4,096-byte inclusive aggregate ceiling before Host value decode, fold, retention, or hostname normalization.
- Aggregate accounting covers duplicate values and the dependency's folding separators; ordinary bounded duplicate diagnostics remain unchanged.
- An over-budget aggregate produces the existing `invalid_authority` reason without firewall matching, credential lookup, forwarding, or retaining/logging the oversized Host. Its public `host_header` diagnostic is `null` or otherwise contains no decoded oversized data.
- Existing accepted canonical IPv4, compressed IPv6, IDNA, trailing-dot, default-port, non-default-port, HTTP/1 absolute-form, HTTP/2/3 authority, SNI, and bounded duplicate behavior remains unchanged.
- Structural real-entry coverage proves that exactly 4,096 aggregate raw bytes are accepted and the 4,097th byte is rejected, including a single many-label value, aggregate duplicate crossing, and multibyte Unicode byte semantics. It proves over-budget Host data cannot reach dependency string access or project Host normalization, without wall-clock assertions.
- Shared configured-host normalization behavior remains unchanged; no request-layer raw authority, SNI, label-count, or normalized-total guard is added.
- A local before/after benchmark records bounded addon runtime and new allocation for oversized Host values; no timing assertion or one-purpose benchmark file is committed.
- The residual pre-hook request-head buffering risk is documented and not presented as solved by this issue.

seven332:
## Summary

- enforce an inclusive 4,096-byte aggregate budget for regular `Host` fields before mitmproxy string access, folding, diagnostic retention, or project hostname normalization
- return from non-HTTPS authority resolution before reading unused Host text
- preserve SNI error precedence, bounded duplicate diagnostics, parser-owned request authority behavior, and existing accepted hostname/port identities
- reject over-budget Host input through the existing `invalid_authority` path with `host_header: null`, before firewall matching, credential lookup, or forwarding

## Behavior and compatibility

The aggregate counts every raw regular Host value plus the `", "` separators used by mitmproxy when folding duplicates. Exactly 4,096 raw bytes remain accepted; the 4,097th byte is rejected. Tests cover multibyte IDNA input, extreme but valid zero-padded ports, single and duplicate crossings, missing/invalid SNI precedence, request-hook logging/auth behavior, and both real mitmproxy request hooks.

The intentional compatibility change is limited to regular Host aggregates above 4,096 raw bytes. Old and new runners may differ for those invalid oversized requests during rollout, but no flow, persisted state, API, queue, or protocol contract crosses runner versions.

## Explicit exclusions

- no new addon limit is applied to parser-validated HTTP/1 request-target authority, HTTP/2/3 `:authority`, or TLS SNI
- shared hostname normalization and bounded ordinary duplicate diagnostics are unchanged
- this does not solve mitmproxy's pre-hook request-head buffering or total header field-count/name work; proxy-wide head limits remain separate hardening

## Benchmark

The same real pinned `Headers`/`HTTPFlow` reproduction was run with input construction outside measurement. Timings are one untraced call; allocation is one separately traced call.

| Raw Host bytes | Before elapsed | After elapsed | Before peak | After peak |
| 65,535 | 15.2 ms | 0.067 ms | 2.0 MiB | 2.6 KiB |
| 1,048,575 | 631.1 ms | 0.052 ms | 33.0 MiB | 2.6 KiB |
| 8,388,607 | 4,737.3 ms | 0.059 ms | 258.3 MiB | 2.6 KiB |

Absolute timings are host-dependent; the result demonstrates that addon work and new allocation no longer scale materially with oversized Host value length.

## Validation

- `uv lock --check`
- `uv sync --locked`
- focused authority/request/framing pytest: 215 passed
- `uv run --no-sync python -m pytest tests/`: 5,596 passed

Closes #27713

### DIFF
--- a/crates/runner/mitm-addon/src/request_authority.py
+++ b/crates/runner/mitm-addon/src/request_authority.py
@@ -9,7 +9,7 @@ for firewall matching or credential injection.
 
 import ipaddress
 from dataclasses import dataclass
-from typing import Literal
+from typing import Final, Literal
 
 from mitmproxy import http
 
@@ -24,6 +24,9 @@ from host_normalization import normalize_hostname
 
 _FORBIDDEN_HOST_CHARS = frozenset("#%*,/<>?@[\\]^|{}")
 _DEFAULT_HTTPS_PORT = 443
+_RAW_HOST_HEADER_NAME: Final = b"host"
+_HOST_HEADER_FOLD_SEPARATOR_BYTES: Final = len(b", ")
+_MAX_HOST_HEADER_BYTES: Final = 4096
 
 
 @dataclass(frozen=True)
@@ -102,6 +105,27 @@ def _build_url(scheme: str, host: str, port: int, path: str) -> str:
     return f"{scheme}://{_host_with_port(scheme, host, port)}{uri_path}"
 
 
+def _host_headers_within_budget(headers: http.Headers) -> bool:
+    """Bound addon Host value work before dependency string access.
+
+    This is not a proxy request-head limit: mitmproxy has already buffered the
+    fields before addon hooks run. Separators match ``Headers`` folding so a
+    bounded result also bounds duplicate-field decoding and diagnostics.
+    """
+    host_header_bytes = 0
+    has_host_header = False
+    for name, value in headers.fields:
+        if name.lower() != _RAW_HOST_HEADER_NAME:
+            continue
+        if has_host_header:
+            host_header_bytes += _HOST_HEADER_FOLD_SEPARATOR_BYTES
+        host_header_bytes += len(value)
+        if host_header_bytes > _MAX_HOST_HEADER_BYTES:
+            return False
+        has_host_header = True
+    return True
+
+
 def _parse_host_authority(authority: str) -> tuple[str, int | None]:
     if not authority or has_ascii_space_or_control(authority):
         raise ValueError("invalid authority")
@@ -155,7 +179,6 @@ def get_trusted_authority(flow: http.HTTPFlow) -> TrustedAuthority:
     scheme = flow.request.scheme
     port = flow.request.port
     path = flow.request.path
-    host_header = flow.request.host_header
     request_host = flow.request.host
 
     if scheme != "https":
@@ -165,7 +188,13 @@ def get_trusted_authority(flow: http.HTTPFlow) -> TrustedAuthority:
             url=_build_url(scheme, request_host, port, path),
         )
 
-    raw_host_headers = flow.request.headers.get_all("Host")
+    host_headers_within_budget = _host_headers_within_budget(flow.request.headers)
+    if host_headers_within_budget:
+        host_header = flow.request.host_header
+        raw_host_headers = flow.request.headers.get_all("Host")
+    else:
+        host_header = None
+        raw_host_headers = []
     raw_sni = getattr(flow.client_conn, "sni", None)
     sni = raw_sni.strip() if isinstance(raw_sni, str) else None
 
@@ -202,6 +231,13 @@ def get_trusted_authority(flow: http.HTTPFlow) -> TrustedAuthority:
         ) from None
 
     trusted_url = _build_url(scheme, normalized_sni, port, path)
+    if not host_headers_within_budget:
+        raise _authority_validation_error(
+            "invalid_authority",
+            message="Request blocked: HTTPS request has invalid Host authority",
+            fallback_url=trusted_url,
+        )
+
     if len(raw_host_headers) > 1:
         raise _authority_validation_error(
             "invalid_authority",
@@ -217,14 +253,15 @@ def get_trusted_authority(flow: http.HTTPFlow) -> TrustedAuthority:
         )
 
     authority_assertions: list[tuple[str, int | None]] = [(host_header, None)]
-    if flow.request.authority:
+    request_authority = flow.request.authority
+    if request_authority:
         if flow.request.is_http2 or flow.request.is_http3:
             if raw_host_headers:
                 authority_assertions.append((raw_host_headers[0], None))
         else:
             # Unlike a Host field, an absolute HTTPS URI with no explicit port
             # identifies the default 443 origin.
-            authority_assertions.append((flow.request.authority, _DEFAULT_HTTPS_PORT))
+            authority_assertions.append((request_authority, _DEFAULT_HTTPS_PORT))
 
     for authority, implicit_port in authority_assertions:
         try:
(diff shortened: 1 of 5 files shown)


seven332:
## Code Review: PR #27743

**Reviewed head:** `6ceccf19c3be24b9cee6958bb2e97436f935c1bb`

### Summary

The change bounds addon-owned HTTPS `Host` authority decoding and normalization at an inclusive 4096 raw-byte aggregate limit, preserves existing SNI error precedence, and rejects oversized values before authentication, upstream IO, or diagnostic exposure. HTTP behavior and parser-bounded request authority/SNI handling remain unchanged.

### Findings

No findings.

### Testing

- Python lock, sync, Ruff format/lint, and BasedPyright passed.
- Focused affected tests passed (`35 passed`).
- Full mitm-addon suite passed (`5596 passed`).
- Boundary, duplicate-field folding, decode-guard, logging/redaction, auth, and real HTTP/1 framing paths are covered.

### Verdict

LGTM

PR #27745, opened 2026-08-17
perf(runner): bound queued jsonl flush waits

### LINKED ISSUE #27711: perf(rust): bound queued mitmproxy JSONL flush waits
## Problem

Each completed runner job awaits its best-effort network-log phase before the job task can finish:

- `crates/runner/src/cmd/start/job_spawn.rs:490-529` calls `mitm_jsonl_flush.flush_path(...)` from `DeferredUploadPhase::flush`.
- `crates/runner/src/cmd/start/job_spawn.rs:763-770` settles finalization before awaiting that phase and returns the cancellation registration only afterward.
- `crates/runner/src/proxy/flush.rs:245-256` serializes every JSONL flush through one shared request lock.
- `crates/runner/src/proxy/flush.rs:259-282` waits for that lock without a deadline; the five-second `JSONL_FLUSH_TIMEOUT` starts only after lock acquisition.

The serialization is necessary because every request uses the same marker files, but waiting callers have no queue bound or end-to-end deadline.

## Impact

If mitmproxy stops acknowledging JSONL flush requests, the lock holder waits five seconds and every later completed job queues behind it. The Nth queued job can therefore wait roughly `N * 5s` before it even reaches network-log upload.

Finalization and budget settlement have already completed at this point, so the runner can continue admitting jobs while earlier `run_job` futures remain pending. Those futures retain their cancellation registrations, telemetry, sandbox token, and executor configuration; the `JoinSet` cannot reap them until the deferred phase returns. Sustained completion throughput above one job per five seconds therefore produces an unbounded backlog and also makes graceful shutdown drain time grow with the queue.

## Acceptance criteria

- Bound the total time a completed job can spend waiting for a JSONL flush, including time waiting for the shared request lock.
- Preserve request-marker serialization or replace it with safe coalescing; concurrent requests must not overwrite each other's marker state.
- Prevent stalled proxy acknowledgements from causing completed job tasks and cancellation registrations to accumulate without bound.
- Add a regression test with an unacknowledged first flush and multiple concurrent later flushes, proving later callers finish within a fixed bound independent of queue length.

seven332:
## Summary

- Bound acquisition of the shared mitmproxy JSONL request lock to five seconds.
- Preserve request marker serialization and the existing full five-second acknowledgement window after the lock is acquired.
- Add a paused-time concurrent regression test covering a stalled holder and multiple queued callers.

## Scope

- No marker schema, addon protocol, API, or persisted-state changes.
- No request coalescing or background-worker redesign.

## Validation

- `cargo test -p runner proxy::flush::tests` (44 passed)
- `cargo test -p runner` (3,149 passed; 20 ignored)

Closes #27711

### DIFF
--- a/crates/runner/src/proxy/flush.rs
+++ b/crates/runner/src/proxy/flush.rs
@@ -27,6 +27,9 @@ pub const USAGE_FLUSH_TIMEOUT: Duration = Duration::from_secs(30);
 /// Maximum time to wait for mitmproxy JSONL writes to become visible before upload.
 pub const JSONL_FLUSH_TIMEOUT: Duration = Duration::from_secs(5);
 
+/// Maximum time to wait to serialize a mitmproxy JSONL flush request.
+const JSONL_FLUSH_LOCK_TIMEOUT: Duration = Duration::from_secs(5);
+
 /// Poll interval when waiting for usage flush.
 const USAGE_FLUSH_POLL: Duration = Duration::from_millis(200);
 
@@ -189,6 +192,7 @@ pub struct MitmJsonlFlushHandle {
 pub(super) enum RequestLockPoll {
     Pending,
     Ready,
+    TimedOut,
 }
 
 pub(super) fn now_millis() -> u64 {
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #27745

**Reviewed head:** `60866d22e0c9afb92a709922a8c66e39f0bf91a7`

### Summary

The change bounds shared JSONL request-lock acquisition while preserving marker serialization and the full existing acknowledgement window. The implementation stays local to the runner flush path and introduces no protocol or persisted-state changes.

### Findings

No findings.

### Testing

Reviewed the paused-time concurrent regression and existing healthy serialization coverage. Formatting, runner Clippy, rustdoc, the 44 focused flush tests, and the full runner suite (3,149 passed; 20 ignored) passed on this head. Unrelated Turbo/TypeScript and Python checks were not run locally.

### Verdict

LGTM

Issue #27765, opened 2026-08-17
test(runner): synchronize transient workspace cache lock release under suite load


seven332:
## Background

Parent #27738 tracks runner tests that become nondeterministic during a concurrent full-suite run. While validating unrelated connector-runtime test synchronization for #27759 on `main` commit `2473e19ff0a963f09176d3e5082f6a60837d37e1`, a complete `cargo test -p runner` run exposed a new workspace-cache failure. The only local source change was in `provider/connector_runtime_sync.rs`.

Related history:

- #25975 / PR #25992 replaced an earlier 20 ms sleep in this test with permission-based lock-attempt observation.
- #27307 corrected production lock-busy classification so the 50 ms contention budget begins only after a real busy result.

## Observed behavior

The third complete runner-suite run failed only:

```text
executor::tests::sandbox_run_tests::workspace_cache::execute_inner_waits_for_transient_workspace_cache_lock

assertion `left == right` failed
  left: Some(WorkspaceDriveConfig { size_mb: 16, seed_image: None })
 right: Some(WorkspaceDriveConfig { size_mb: 16, seed_image: Some(Move(".../current.ext4")) })
```

Suite result:

```text
3158 passed; 1 failed; 20 ignored
```

The exact test passed immediately afterward:

```text
cargo test -p runner executor::tests::sandbox_run_tests::workspace_cache::execute_inner_waits_for_transient_workspace_cache_lock -- --exact
```

## Problem

The test intentionally holds the seeded entry lock. A controller OS thread observes lock attempts through the lock file's permission tightening:

1. wait for the first attempt to change mode from `0644` to `0600`;
2. reset the mode to `0644`;
3. wait for a second attempt to change it to `0600`;
4. release the held lock.

Waiting for the second observation proves the first acquisition actually saw contention and entered the retry loop. However, production `WorkspaceImageCache::prepare` starts its real 50 ms contention budget after that first busy result. Under full-suite CPU and scheduler pressure, the controller thread is not guaranteed to observe the second permission change and drop the guard inside those 50 ms. The prepare path can therefore return `LockBusy`, create a fresh workspace with no seed image, and fail the test even though the exact test passes alone.

The test's one-second controller watchdog bounds a missing observation, but it does not protect the shorter production contention budget. Increasing the production timeout would mask the test coordination defect and change product behavior.

## Expected behavior

[...]

Issue #27738, opened 2026-08-17
test(runner): eliminate full-suite concurrency flakes


seven332:
## Background

The runner test suite is expected to pass as a complete crate-level check. During validation of PR #27682 on a 12-core development container, the same worktree was reliable when tests ran serially but produced changing failures at higher test concurrency.

## Observed behavior

At commit `420b3f8f0e3435f811cc515bb85eb031ef1a5021`:

1. `cargo test -p runner` failed 7 of 3,178 unit tests across unrelated modules:
   - `cmd::service::signal::tests::signal_service_main_bounded_preserves_signal_outcomes`
   - `cmd::start::tests::main_loop::admission::same_run_duplicate_does_not_renew_claimed_finalizing_deadline`
   - `network_logs::tests::upload_network_logs_allows_exact_source_byte_budget`
   - `network_logs::tests::upload_network_logs_stops_at_batch_budget`
   - `provider::connector_runtime_sync::tests::stalled_run_does_not_block_or_overlap_other_run`
   - `proxy::flush::tests::jsonl_flush_handle_serializes_concurrent_requests`
   - `proxy::flush::tests::jsonl_flush_handles_from_same_proxy_serialize_concurrent_requests`
2. Each failed test passed when rerun by itself.
3. `cargo test -p runner -- --test-threads=4` made those 7 pass but failed a different test:
   - `workspace_image_cache::watcher::tests::foreign_scope_commit_does_not_wake_observer`
4. That watcher test passed when rerun by itself.
5. `cargo test -p runner -- --test-threads=1` passed: 3,158 passed, 20 ignored; guest inventory also passed.

The failures included one-second timing assertions, missing requests to local mock servers, and observer state that differed only during the concurrent full-suite run. No single failing test reproduced in isolation.

Environment:

- 12 available CPUs
- `rustc 1.97.0 (2d8144b78 2026-07-07)`
- `cargo 1.97.0 (c980f4866 2026-06-30)`

## Expected behavior

The documented crate-level runner test command should pass reliably at its supported concurrency without requiring repeated reruns.

## Scope

Investigate cross-test resource contention, shared process/environment state, and synchronization assumptions in the failing tests or runner test configuration. This is separate from PR #27682's raw HTTP fixture migration because the failures span unrelated signal, admission, network log, proxy flush, connector, and filesystem watcher tests.

[...]

Issue #27722, opened 2026-08-17
test(runner): make finalizing deadline coverage test deterministic


seven332:
## Problem and evidence

The Crates `coverage` job can fail finalizing deadline admission tests nondeterministically.

Observed in PR #27709, which contains no Rust changes:

- failed coverage job: https://github.com/vm0-ai/vm0/actions/runs/32005843410/job/95315181255
- `same_run_duplicate_does_not_renew_claimed_finalizing_deadline` expected `Some(Active)` and observed `Some(Expired)`
- 3,147 other tests passed in that job
- the same targeted test passed locally immediately afterward

A later full-suite reproduction on current `main` failed both telemetry assertions covered here while both tests passed immediately afterward in isolation:

- `selected_ranked_finalizing_candidate_falls_back_at_deadline`
- `same_run_duplicate_does_not_renew_claimed_finalizing_deadline`

Both run with paused Tokio time, create 100 ms deadlines from `std::time::Instant::now()`, and recompute preference state after completion using the real wall clock. Coverage instrumentation or runner load can therefore classify the stored candidates as expired even though their deadlines were retained and correctly triggered fallback.

A full-file audit found one more admission test with the same initial short-deadline assumption: `published_exact_remains_reusable_past_preference_deadline` uses a 100 ms standard-library deadline and paused-Tokio expiry. It does not inspect late telemetry, but initial admission still competes with the same sub-100-ms real scheduling window.

## Decision

Treat this as a test-observation defect, not a production deadline or telemetry defect. Cover all three admission tests that combine a short standard-library finalizing deadline with paused Tokio expiry in one test-only PR.

## Chosen direction and rationale

- Use one 30-second preference lifetime as virtual test scenario input across all three affected tests.
- In the duplicate test, make the later duplicate deadline exactly one additional virtual lifetime after the original deadline.
- In the two telemetry tests, assert the exact retained preference deadline and expected tier on the candidate recorded during claim, before advancing paused time.
- Advance paused Tokio time just past the original virtual lifetime, then keep the completion, reuse, claim-count, and cleanup assertions that exercise the real discovery, claim, finalizing wait, duplicate suppression, exact publication, and fallback paths.
- Remove the post-completion `Active` classifications that read a different clock at a later observation point.
- Leave production admission, provider, deadline, and telemetry code unchanged.

[...]

PR #27766, opened 2026-08-17
test(runner): synchronize connector policy publication

### LINKED ISSUE #27759: test(runner): synchronize connector runtime policy updates without polling deadlines
## Background

Parent objective: #27738 tracks runner tests that become nondeterministic during a concurrent full-suite run.

At runner commit `420b3f8f0e3435f811cc515bb85eb031ef1a5021`, `cargo test -p runner` intermittently failed `provider::connector_runtime_sync::tests::stalled_run_does_not_block_or_overlap_other_run`. The test passed immediately when run alone, and the complete suite passed serially.

## Problem

The test uses loopback HTTP to hold one run in flight, completes a second run, and then waits for the second run's connector policy through `wait_until_slack_policy`. That helper polls until a one-second wall-clock timeout.

Under full-suite scheduling pressure, the policy update was not observed before the polling deadline:

```text
slack policy should match before timeout: Elapsed(())
```

The test intends to prove per-run concurrency and non-overlap, but its result also depends on the executor scheduling the worker, response handling, registry update, and polling task within one second. Elapsed time is acting as the synchronization mechanism for a positive state transition.

## Expected behavior

The test should use a causal, deterministic observation of the connector policy update while retaining a bounded failure guard. It must continue to prove that run B can complete while run A is stalled and that a follow-up for run A does not overlap its first request.

## Acceptance criteria

- The test waits for an explicit connector-policy update or an equivalent deterministic state transition instead of relying on repeated polling progress within one second.
- A timeout may bound a missing positive signal, but elapsed time is not used as evidence that the expected update occurred.
- The stalled run, independent run, and same-run non-overlap invariants remain covered through real loopback HTTP behavior.
- The focused connector-runtime-sync tests and the complete runner suite pass repeatedly at the supported concurrency.
- Do not skip tests or fix the failure solely by increasing a timeout.
- Preserve production scheduling, retry, policy registry, and shutdown behavior.

## Delivery boundary

This sub-issue covers only connector-runtime-sync test synchronization. Other concurrency flakes remain separate children of #27738.

seven332:
## Summary

- retain the last successfully published connector generation in test builds
- observe the exact run and generation after registry publication before reading the persisted policy
- keep the real loopback HTTP, independent-run progress, coalesced follow-up, and same-run non-overlap assertions

## Scope

- production connector scheduling, retry, registry, cancellation, shutdown, and interfaces are unchanged
- the three unreported `wait_until_slack_policy` call sites remain unchanged
- parent #27738 and sibling flakes remain separate work

## Validation

- affected exact test: 11 consecutive passes
- `cargo test -p runner provider::connector_runtime_sync::tests::`: 47 passed
- one complete `cargo test -p runner`: 3,159 runner tests passed, 20 ignored; guest inventory passed

Two additional complete-suite attempts also passed the connector runtime sync test but each exposed a different out-of-scope concurrency flake: the existing #27722 finalizing-deadline test and newly filed #27765 workspace-cache lock-release test. Both failed tests passed immediately when rerun alone and are intentionally not changed here.

Closes #27759

Parent: #27738

### DIFF
--- a/crates/runner/src/provider/connector_runtime_sync.rs
+++ b/crates/runner/src/provider/connector_runtime_sync.rs
@@ -107,6 +107,8 @@ struct ActiveConnectorSyncState {
     registration: ConnectorRuntimeTargetRegistration,
     generation: u64,
     consecutive_failures: u32,
+    #[cfg(test)]
+    successful_generation: tokio::sync::watch::Sender<Option<u64>>,
 }
 
 struct ScheduledSyncTask {
@@ -281,6 +283,8 @@ impl ConnectorRuntimeSyncCore {
                             registration: registration.clone(),
                             generation: 0,
                             consecutive_failures: 0,
+                            #[cfg(test)]
+                            successful_generation: tokio::sync::watch::channel(None).0,
                         },
                     )
                 })
@@ -1212,7 +1216,17 @@ impl ConnectorRuntimeSyncCore {
             return false;
         }
         connector.consecutive_failures = 0;
-        self.replace_schedule_locked(active, run_id, target, deadline)
+        let completed = self.replace_schedule_locked(active, run_id, target, deadline);
+        #[cfg(test)]
+        if completed {
+            active
+                .connectors
+                .get(&target.target)
+                .expect("completed connector should remain registered")
+                .successful_generation
+                .send_replace(Some(target.generation));
+        }
+        completed
     }
 
     async fn schedule_sync_retries_for_registration(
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #27766

**Reviewed head:** `1033cc689fe378774bbdf0191358d3ef61f52da3`

### Summary

The test now waits for the exact run and connector generation to complete a successful registry publication instead of treating repeated file polling as synchronization. The observer and retained state are test-only; production scheduling and persistence behavior are unchanged.

### Findings

No findings.

### Testing

The change retains real loopback HTTP, dispatcher, filesystem registry publication, coalescing, independent-run progress, and same-run non-overlap coverage. Local validation passed formatting, Clippy, docs, 11 consecutive exact-test runs, all 47 connector runtime sync tests, and one complete runner suite (3,159 passed, 20 ignored).

### Verdict

LGTM

PR #27767, opened 2026-08-17
test(runner): make finalizing deadline tests deterministic

### LINKED ISSUE #27722: test(runner): make finalizing deadline coverage test deterministic

seven332:
## Summary

- assert the exact finalizing preference deadline retained by the candidate recorded during claim
- replace sub-100-ms real scheduling windows with a shared 30-second lifetime crossed immediately by paused Tokio time
- cover all three admission tests with the same short standard-library deadline and virtual-time transition

Production admission, deadline, provider, and telemetry behavior is unchanged. The longer value is test scenario input, not a wall-clock wait or diagnostic timeout.

## Validation

- duplicate-deadline target repeated 20 times
- selected-finalizing and published-exact tests repeated 10 times each
- `cargo llvm-cov -p runner --all-targets --all-features --no-report -- cmd::start::tests::main_loop::admission::same_run_duplicate_does_not_renew_claimed_finalizing_deadline --exact`
- `RUSTDOCFLAGS='-D warnings' cargo doc -p runner --no-deps --all-features`
- `cargo test -p runner -- --quiet` passed on the initial implementation (3,159 passed, 20 ignored); two later final-revision attempts passed all changed tests but separately reproduced unrelated full-suite flakes in #27765 and `dns::port::tests::find_available_port_retries_when_udp_candidate_is_in_use`, both of which passed immediately in isolation
- after rebasing onto current `main`, all three exact tests and the coverage-instrumented target passed again

Closes #27722

Parent context: #27738

### DIFF
--- a/crates/runner/src/cmd/start/tests/main_loop/admission.rs
+++ b/crates/runner/src/cmd/start/tests/main_loop/admission.rs
@@ -18,6 +18,7 @@ use crate::types::SandboxReuseResult;
 use crate::workspace_image_cache::WorkspaceImageCache;
 
 const NON_SELECTED_RUNNER_ID: u128 = 1;
+const FINALIZING_TEST_PREFERENCE_LIFETIME: Duration = Duration::from_secs(30);
 
 fn ranked_candidate(
     run_id: RunId,
@@ -783,7 +784,7 @@ async fn selected_ranked_finalizing_candidate_falls_back_at_deadline() {
     let run_id = RunId::new_v4();
     env.provider
         .set_claim_result(run_id, Some(context_with_reuse_key(run_id, reuse_key)));
-    let deadline = std::time::Instant::now() + Duration::from_millis(100);
+    let deadline = std::time::Instant::now() + FINALIZING_TEST_PREFERENCE_LIFETIME;
     env.handle
         .discover_tx
         .send(
@@ -814,7 +815,7 @@ async fn selected_ranked_finalizing_candidate_falls_back_at_deadline() {
     );
     assert_eq!(preference.deadline(), deadline);
 
-    tokio::time::advance(Duration::from_millis(101)).await;
+    tokio::time::advance(FINALIZING_TEST_PREFERENCE_LIFETIME + Duration::from_millis(1)).await;
     let completion = env
         .handle
         .wait_completion(run_id, Duration::from_secs(5))
@@ -965,7 +966,7 @@ async fn published_exact_remains_reusable_past_preference_deadline() {
                 RunnerPreferenceTier::FinalizingPredecessor,
                 TEST_RUNNER_ID,
                 TEST_HEARTBEAT_GENERATION,
-                std::time::Instant::now() + Duration::from_millis(100),
+                std::time::Instant::now() + FINALIZING_TEST_PREFERENCE_LIFETIME,
             )
             .with_history_generation_run_id(Some(history_generation_run_id)),
         )
@@ -993,7 +994,7 @@ async fn published_exact_remains_reusable_past_preference_deadline() {
         .await
         .expect("published exact sandbox should start before predecessor release");
 
-    tokio::time::advance(Duration::from_millis(101)).await;
+    tokio::time::advance(FINALIZING_TEST_PREFERENCE_LIFETIME + Duration::from_millis(1)).await;
     successor_gate.release_one();
     let completion = env
         .handle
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #27767

**Reviewed head:** `04e2471799494fe191c61fc0c54fe8b23bc4c796`

### Summary

The final test-only diff replaces unstable late wall-clock telemetry classification with exact retained-deadline assertions and removes the remaining sub-100-ms admission windows from all three affected tests. The 30-second lifetime is crossed through paused Tokio time and does not change production behavior or real test duration.

### Findings

No findings. Self-review found one timing gap in the first commit; the second commit resolves it and a full-file audit found no remaining instance of the same short-deadline pattern in this admission boundary.

### Testing

Formatting, runner Clippy, rustdoc, repository hooks, repeated exact runs (20/10/10), and the coverage-instrumented target passed. After rebasing onto current `main`, all three exact tests and coverage passed again. The initial implementation passed the full runner suite; later final-revision suite attempts passed every changed test but separately reproduced unrelated flakes outside this diff (#27765 and the DNS available-port retry test), each of which passed immediately in isolation. Canonical CI on this rebased head remains the final full-suite check.

### Verdict

LGTM

PR #27780, opened 2026-08-17
fix(api): bound oversized axiom agent event data

### LINKED ISSUE #27773: bug(api): bound oversized Axiom agent event fields
## Background

Agent webhook events are projected to PostgreSQL and also delivered to the `vm0-agent-run-events-prod` Axiom dataset on a best-effort basis. The complete runtime event is stored under `eventData`, which is already configured as a native Axiom map field to prevent arbitrary nested keys from expanding the dataset schema.

Axiom documents a maximum field size of 1 MB. A map field avoids field-count growth, but the complete map value still has to fit that per-field size limit.

Related work addressed different boundaries:

- #25347 converted `eventData` to a native map field to stop schema field-count exhaustion.
- #26466 reduced Codex webhook delivery events that would otherwise exceed the shared 4 MiB request limit.

## Observed behavior

On August 17, 2026, production run `ed355214-450c-482e-bb50-74e856405a50` successfully posted sequence `6` to `/api/webhooks/agent/events` with a request size of `1,198,858` bytes. The required PostgreSQL projection succeeded and the webhook returned HTTP 200, but the optional Axiom delivery failed with HTTP 400:

```text
Optional Axiom trace delivery failed
DirectAxiomIngestError: Axiom ingest failed with status 400
```

The surrounding events were much smaller. The run continued and completed, so this failure did not affect agent execution, but sequence `6` is absent from the Axiom trace used by activity views, search, and diagnostics.

The original nested field cannot be recovered from current telemetry: the API network record intentionally omits webhook request bodies, the failed event never entered Axiom, and `ingestAxiomDirect()` currently discards the non-2xx response body. The evidence is sufficient to identify the boundary failure without retaining or logging user event content: the complete `eventData` map value exceeded Axiom's documented 1 MB field limit.

## Expected behavior

Every accepted agent webhook event should retain its required PostgreSQL behavior. The optional Axiom projection should also remain ingestible when the original runtime event is larger than Axiom's per-field limit, with any loss of Axiom-only detail made explicit and observable.

## Requirements

- Apply the size boundary at the API's Axiom projection, not to the webhook payload or required PostgreSQL projection.
- Keep normal-sized Axiom events unchanged.
- Preserve stable top-level fields used for filtering and ordering: `runId`, `userId`, `sequenceNumber`, and `eventType`.
- Replace an oversized `eventData` value with a deterministic bounded representation that preserves useful event identity and explicitly records that Axiom-only content was reduced.
- Measure serialized UTF-8 bytes and leave conservative headroom below Axiom's 1 MB per-field limit.
- Emit content-free telemetry for reductions, including event classification and original/delivered byte counts.
- Keep Axiom delivery best effort; do not change webhook acknowledgement or agent-run success semantics.
- Cover all agent frameworks and event shapes rather than only the observed Claude replay event.

## Acceptance criteria

- [ ] An agent event with an `eventData` value above Axiom's field-size limit is sent as a valid bounded Axiom event instead of receiving HTTP 400 solely due to that value.
- [ ] The reduced event preserves `runId`, `userId`, `sequenceNumber`, `eventType`, and useful event identity while clearly marking Axiom-only reduction and the original byte size.
- [ ] A normal-sized event retains its existing Axiom JSON value.
- [ ] The required PostgreSQL projection continues to receive the complete original event.
- [ ] The webhook contract, shared 4 MiB request limit, sequence allocation, and acknowledgement behavior do not change.
- [ ] Reduction telemetry contains sizes and event classification but no submitted event content.
- [ ] API integration coverage verifies both oversized reduction and unchanged normal delivery through the existing webhook/Axiom path.
- [ ] No database migration, Axiom dataset replacement, or runner protocol change is introduced.

## Relevant code

- `turbo/apps/api/src/signals/services/agent-event-consumer-axiom.service.ts`
- `turbo/apps/api/src/signals/external/axiom.ts`
- `turbo/apps/api/src/signals/services/agent-webhook-events.service.ts`
- `turbo/apps/api/src/signals/routes/__tests__/webhooks-callbacks.bdd.test.ts`

## References

- [Axiom limits](https://axiom.co/docs/reference/limits)
- [Axiom map fields](https://axiom.co/docs/apl/data-types/map-fields)

seven332:
## Summary

- measure each agent event's serialized UTF-8 size before constructing the optional Axiom trace
- keep the original event shape and metadata while reducing the largest string values until oversized `eventData` fits the conservative 900,000-byte budget
- retain the displayable string prefix with a visible `[truncated]` suffix and record `axiomReduction` metadata without copying event content into logs
- fall back to a minimal reduction marker only when non-string JSON structure cannot fit after reducible strings are exhausted
- cover the oversized multibyte path through the real webhook/Axiom integration boundary while retaining the existing exact normal-event assertion

## Behavior and compatibility

- The Activity reader can still render retained result text, duration, turn count, model usage, assistant content, and tool data from the preserved event structure.
- The complete inbound event still reaches the required PostgreSQL projection before optional consumers run.
- Normal Axiom event values remain unchanged.
- The webhook contract, acknowledgement, sequence allocation, runner protocol, public API schema, and Axiom dataset field type are unchanged.
- Oversized activity rows contain explicit reduction metadata and a displayable content prefix instead of being absent after an Axiom HTTP 400.

## Validation

- `pnpm format`
- `pnpm -F api lint`
- `pnpm -F api check-types`
- `pnpm -F api exec vitest run src/signals/routes/__tests__/webhooks-callbacks.bdd.test.ts --silent=passed-only` (48 passed)
- `pnpm knip`
- pre-commit full Turbo type checking (14 tasks passed)

Closes #27773

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/webhooks-callbacks.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/webhooks-callbacks.bdd.test.ts
@@ -1357,21 +1357,45 @@ describe("WHCB-04: internal callback and event-consumer boundaries", () => {
     expect(redirectTargetRequests).toBe(0);
   });
 
-  it("bounds oversized event data only in the optional Axiom trace", async () => {
+  it("preserves display fields while bounding oversized Axiom traces", async () => {
     const { actor, runId, headers } = await createEventWebhookRun(
       "oversized optional Axiom trace",
     );
-    const oversizedContent = "界".repeat(300_000);
-    const event = {
-      type: "result",
+    const oversizedAssistantContent = "助".repeat(300_000);
+    const assistantEvent = {
+      type: "assistant",
       sequenceNumber: 0,
-      result: oversizedContent,
+      message: {
+        content: [
+          { type: "text", text: oversizedAssistantContent },
+          {
+            type: "tool_use",
+            id: "tool_1",
+            name: "Bash",
+            input: { command: "pwd" },
+          },
+        ],
+      },
+    };
+    const oversizedResultContent = "界".repeat(300_000);
+    const resultEvent = {
+      type: "result",
+      sequenceNumber: 1,
+      result: oversizedResultContent,
       duration_ms: 123,
       num_turns: 4,
       modelUsage: { inputTokens: 100, outputTokens: 200 },
     };
-    const originalBytes = Buffer.byteLength(JSON.stringify(event), "utf8");
-    expect(originalBytes).toBeGreaterThan(900_000);
+    const assistantOriginalBytes = Buffer.byteLength(
+      JSON.stringify(assistantEvent),
+      "utf8",
+    );
+    const resultOriginalBytes = Buffer.byteLength(
+      JSON.stringify(resultEvent),
+      "utf8",
+    );
+    expect(assistantOriginalBytes).toBeGreaterThan(900_000);
+    expect(resultOriginalBytes).toBeGreaterThan(900_000);
 
     const requests: unknown[] = [];
     server.use(
@@ -1379,20 +1403,20 @@ describe("WHCB-04: internal callback and event-consumer boundaries", () => {
         "https://api.axiom.co/v1/datasets/agent-run-events/ingest",
         async ({ request }) => {
           requests.push(await request.json());
-          return HttpResponse.json(successfulAxiomIngestStatus(1));
+          return HttpResponse.json(successfulAxiomIngestStatus(2));
         },
       ),
     );
 
     const response = await api.requestAgentEvents(
-      { runId, events: [event] },
+      { runId, events: [assistantEvent, resultEvent] },
       headers,
       [200],
     );
     expect(response.body).toStrictEqual({
-      received: 1,
+      received: 2,
       firstSequence: 0,
-      lastSequence: 0,
+      lastSequence: 1,
     });
     await flushWaitUntilForTest();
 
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #27780

**Reviewed head:** `14115dc34b0549ffdd679c1d5aaa329f3652ac9d`

### Summary

This change bounds only the optional Axiom `eventData` projection when its serialized UTF-8 size exceeds the conservative 900,000-byte budget. Normal events remain exact, the required database projection and webhook contract are unchanged, and oversized content is replaced with a deterministic reduction marker while stable correlation fields remain queryable.

### Findings

No findings.

### Testing

- Focused webhook callback integration test: 48 tests passed
- API lint and type check passed
- Full Knip and Turbo type checks passed
- Formatter and commit hooks passed
- Production Axiom canary was not run; the integration test verifies the exact outbound request at the HTTP boundary

### Verdict

LGTM

seven332:
## Review summary

Reviewed the complete PR diff and every commit at head `9041f4bc288bdd85d87148cdb5277444cd991898`.

No actionable findings. The implementation keeps normal events unchanged, reduces only oversized optional Axiom `eventData`, preserves nested assistant/result/tool structure for Activity rendering, uses exact UTF-8 JSON byte accounting, avoids logging event content, and retains a minimal fallback for oversized non-string structure. The `axiomReduction` name is consistent and no `vm0AxiomReduction` references remain.

Validation observed:

- API route integration suite: 48 passed
- API lint: passed (repository warnings only)
- API type check: passed
- Knip: passed
- Full pre-commit Turbo type checking: 14 tasks passed

Verdict: LGTM.

PR #27826, opened 2026-08-18
test(e2e): isolate serpapi auth probe from ipv6 dns

### LINKED ISSUE #27785: test: make SerpAPI firewall E2E resilient to dropped AAAA replies
## Problem

The runner E2E test `tests/03-runner/run-t04-firewall-serpapi.bats` can fail before the firewall authentication path is reached when one DNS family reply is dropped.

Observed in PR #27754, Actions run `32022849053`, job `95367357212`:

- the sandbox emitted A and AAAA queries for `serpapi.com`
- the A query returned valid addresses
- no AAAA reply was recorded
- `curl --max-time 5` failed with `Resolving timed out after 5000 milliseconds`
- the test then timed out waiting for a SerpAPI firewall traffic log
- connector setup and `SERPAPI_TOKEN` injection had already succeeded

This makes the authentication E2E sensitive to an unrelated DNS packet loss and reports the failure as a firewall regression.

## Decision

Keep the real `https://serpapi.com` request and normal sandbox resolver path, but constrain this firewall-auth probe to IPv4. Make DNS and transport failures terminal for the prompt so the success marker is emitted only after the request completes.

This preserves real DNS, TLS SNI, hostname routing, proxying, query-auth injection, and firewall telemetry without depending on an unrelated AAAA reply. Raw UDP and TCP DNS behavior remains covered by the dedicated runner network-telemetry E2E.

## Implementation direction

- Enable strict shell behavior in the SerpAPI mock-agent prompt.
- Add the IPv4 resolution option to the existing `curl` request.
- Remove unconditional `curl` failure suppression.
- Keep the real hostname, URL, five-second maximum, and existing safe-secret and firewall-log assertions.
- Keep `SERPAPI_REQUEST_SENT` after `curl` so it proves transport completion.

If IPv4 DNS or transport fails, the prompt stops before the marker and the chat assertion reports the captured `curl` diagnostic immediately. If the request reaches the proxy but authentication or query injection is wrong, the marker is present and the existing firewall telemetry assertion reports the auth mismatch.

## Alternatives considered

- `curl --resolve` would remove guest DNS but pin a dynamic third-party address that the repository does not control.
- Dual-stack retries do not deterministically handle unavailable IPv6 resolution and add repeated timeout latency.
- A production DNS proxy change is unsupported by the current evidence and would introduce platform-wide protocol and caching risk.

## Scope and risks

- Change only the SerpAPI Bats E2E; no production runner, API, protocol, migration, or deployment change.
- Do not broaden this PR to Agora or Zendesk without separate failure evidence.
- The test no longer incidentally covers SerpAPI IPv6, which is outside its firewall-auth responsibility.
- Real IPv4 DNS or provider transport outages will fail visibly instead of being swallowed; this is intentional and keeps diagnostics truthful.

## Acceptance criteria

- The SerpAPI firewall E2E does not fail solely because an AAAA response is missing when IPv4 resolution succeeds.
- The test still fails when firewall authentication or query injection is broken.
- DNS or transport failure prevents the success marker and surfaces the request diagnostic without waiting for the firewall-log timeout.
- The test continues using the real SerpAPI hostname and verifies the expected `SERPAPI_TOKEN` secret reference in an allowed firewall log.
- Relevant local checks and runner E2E coverage pass in CI.

seven332:
## Summary

- constrain the SerpAPI firewall-auth probe to IPv4 while preserving the real provider hostname and proxy path
- stop suppressing DNS and transport failures so the success marker proves transport completion
- emit a common success/failure prefix before exit, then require the success outcome so request errors surface without a second 90-second wait
- keep the existing safe-secret output and firewall telemetry assertions unchanged

## Scope

- changes only the SerpAPI runner Bats E2E
- does not change production DNS behavior, runner protocols, APIs, migrations, or other provider probes

## Validation

- `e2e/test/libs/bats/bin/bats --count e2e/tests/03-runner/run-t04-firewall-serpapi.bats` (`1` test discovered)
- focused shell failure-path simulation verified exit status `6`, `SERPAPI_REQUEST_FAILED=6`, and no success marker
- `git diff --check`
- repository pre-commit hook for the staged Bats file (`check-file-size`)
- commitlint

The deployed `cli-e2e-03-runner` shard remains the acceptance-level integration gate because runner Bats tests are CI-only.

Closes #27785

### DIFF
--- a/e2e/tests/03-runner/run-t04-firewall-serpapi.bats
+++ b/e2e/tests/03-runner/run-t04-firewall-serpapi.bats
@@ -27,8 +27,13 @@ teardown() {
 
     local prompt
     prompt=$(cat <<'EOF'
+set -euo pipefail
 printf 'SERPAPI_TOKEN=%s\n' "$SERPAPI_TOKEN"
-curl --silent --show-error --max-time 5 --output /dev/null 'https://serpapi.com/search?q=vm0-e2e&engine=google' || true
+# Raw DNS has dedicated runner coverage. Keep this firewall-auth probe on IPv4
+# so an unavailable AAAA response cannot block an otherwise valid request.
+curl --ipv4 --silent --show-error --max-time 5 \
+    --output /dev/null \
+    'https://serpapi.com/search?q=vm0-e2e&engine=google'
 printf 'SERPAPI_REQUEST_SENT\n'
 EOF
 )
--- a/e2e/tests/03-runner/run-t04-firewall-serpapi.bats
+++ b/e2e/tests/03-runner/run-t04-firewall-serpapi.bats
@@ -31,10 +31,15 @@ set -euo pipefail
 printf 'SERPAPI_TOKEN=%s\n' "$SERPAPI_TOKEN"
 # Raw DNS has dedicated runner coverage. Keep this firewall-auth probe on IPv4
 # so an unavailable AAAA response cannot block an otherwise valid request.
-curl --ipv4 --silent --show-error --max-time 5 \
+if curl --ipv4 --silent --show-error --max-time 5 \
     --output /dev/null \
-    'https://serpapi.com/search?q=vm0-e2e&engine=google'
-printf 'SERPAPI_REQUEST_SENT\n'
+    'https://serpapi.com/search?q=vm0-e2e&engine=google'; then
+    printf 'SERPAPI_REQUEST_SENT\n'
+else
+    curl_status=$?
+    printf 'SERPAPI_REQUEST_FAILED=%s\n' "$curl_status"
+    exit "$curl_status"
+fi
 EOF
 )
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #27826

**Reviewed head:** `3b88becc15d6288ce91068f72e19ad04e977bd35`

### Summary

Reviewed the SerpAPI runner E2E's IPv4-only request path, mock-shell failure semantics, chat-event synchronization, firewall telemetry assertion, and CI-only validation boundary. The final head preserves the real provider/proxy/auth path while making DNS and transport failures visible before firewall-log polling.

### Findings

No findings.

### Testing

Bats discovery (`1` test), diff checks, focused success and DNS-failure shell simulations, the staged-file pre-commit hook, and commitlint passed. The deployed `cli-e2e-03-runner` shard is intentionally not runnable locally and remains the final integration gate.

### Verdict

LGTM

PR #27829, opened 2026-08-18
test(runner): avoid cancelling watcher mid-transition

### LINKED ISSUE #27762: test(runner): avoid cancelling workspace cache watcher mid-transition
## Background

Parent objective: #27738 tracks runner tests that become nondeterministic during a concurrent full-suite run.

At runner commit `420b3f8f0e3435f811cc515bb85eb031ef1a5021`, `cargo test -p runner -- --test-threads=4` intermittently failed `workspace_image_cache::watcher::tests::foreign_scope_commit_does_not_wake_observer`. The test passed immediately when run alone, and the complete suite passed serially.

## Problem

The test wraps `watcher.next_change()` in a 100 ms timeout to prove that a foreign-scope commit does not emit a change. After the timeout cancels that future, the test immediately asserts that the foreign cache key is absent from `watch_by_cache_key`.

`next_change()` mutates watcher state in phases: it can install an unclassified entry watch and then await metadata scope classification before removing a foreign watch. If the outer timeout cancels the future between those steps, the test observes the intentionally interrupted intermediate state:

```text
assertion failed: !watcher.watch_by_cache_key.contains_key(&foreign_cache_key)
```

The 100 ms absence check is therefore also being used as a completion barrier for internal inotify processing, which it cannot guarantee. The production runner owns and preserves the in-flight watcher future across unrelated `select!` branches, so the test should not infer completed state after deliberately cancelling it unless cancellation safety is itself the behavior under test.

## Expected behavior

The test should deterministically prove both externally visible requirements: a foreign-scope commit does not wake the observer, and fully processed foreign metadata does not retain an entry watch. It should not inspect post-processing state after cancelling the operation mid-transition.

## Acceptance criteria

- The test no longer treats a 100 ms timeout as proof that all queued inotify work has completed.
- It verifies that a foreign-scope commit emits no workspace-cache change.
- It verifies, after a deterministic processing barrier, that the foreign cache key is not retained in watcher state.
- Coverage continues to use a real temporary filesystem and inotify behavior.
- The focused watcher tests and the complete runner suite pass repeatedly at the supported concurrency.
- Do not skip tests or fix the failure solely by increasing a timeout.
- Preserve production watcher scope filtering and the runner's ownership of the in-flight watcher future unless a separate runtime bug is demonstrated.

## Delivery boundary

This sub-issue covers only workspace-cache watcher test cancellation and synchronization. Other concurrency flakes remain separate children of #27738.

seven332:
## Summary

- preserve the in-flight workspace cache watcher future across the foreign-only timeout
- publish a relevant cache entry afterward as the deterministic processing barrier
- verify the completed change reports only the relevant key before inspecting foreign watch cleanup

## Scope

This is a test-only synchronization change. It does not change production watcher filtering, ownership, timeouts, or runtime behavior.

Closes #27762

Parent context: #27738

## Validation

- focused watcher test: 20 consecutive passes
- `cargo test -p runner -- --test-threads=4 --quiet`: two consecutive passes, each with 3,160 passed and 20 ignored runner tests plus the passing guest inventory test
- pre-commit Rust formatting, workspace rustdoc, workspace Clippy, file-size, and commit-message hooks

An earlier full-suite attempt exposed an unrelated existing race in `proxy::process::tests::graceful_stop_kills_descendant_after_parent_exits_cleanly`, where path existence can precede PID-file contents. The watcher test passed in that run, and the two subsequent complete runs passed. The separate flake is tracked in #27830.

### DIFF
--- a/crates/runner/src/workspace_image_cache/watcher.rs
+++ b/crates/runner/src/workspace_image_cache/watcher.rs
@@ -762,15 +762,26 @@ mod tests {
             .unwrap();
         let observer = WorkspaceImageCache::shared(observer_paths, &home, "group-a");
         let publisher = WorkspaceImageCache::shared(publisher_paths, &home, "group-b");
-        let mut watcher = WorkspaceCacheWatcher::new(observer).await.unwrap();
+        let mut watcher = WorkspaceCacheWatcher::new(observer.clone()).await.unwrap();
 
         let foreign_cache_key = commit_entry(&publisher, "foreign-only-entry").await;
 
+        let mut next_change = Box::pin(watcher.next_change());
         assert!(
-            tokio::time::timeout(std::time::Duration::from_millis(100), watcher.next_change())
+            tokio::time::timeout(std::time::Duration::from_millis(100), &mut next_change)
                 .await
                 .is_err()
         );
+        let relevant_cache_key = commit_entry(&observer, "relevant-entry").await;
+        let change = tokio::time::timeout(std::time::Duration::from_secs(2), &mut next_change)
+            .await
+            .unwrap()
+            .unwrap();
+        assert_eq!(
+            change.committed_cache_keys,
+            BTreeSet::from([relevant_cache_key])
+        );
+        drop(next_change);
         assert!(!watcher.watch_by_cache_key.contains_key(&foreign_cache_key));
     }
 


seven332:
## Code Review: PR #27829

**Reviewed head:** `1f9800fab87370d7300bdc65ab86ae80d6b9a0a7`

### Summary

The watcher regression test now preserves the same in-flight `next_change()` future across its foreign-only timeout, then publishes a relevant entry as an ordered completion barrier before inspecting foreign watch cleanup. Production watcher behavior and timeout values are unchanged.

### Findings

No findings.

### Testing

Rust formatting, runner Clippy, runner rustdoc, and repository commit hooks passed. The focused watcher test passed 20 consecutive runs, and the complete runner suite passed twice consecutively with four test threads. An unrelated proxy descendant PID publication race observed in an earlier suite attempt is tracked separately in #27830.

### Verdict

LGTM

PR #27831, opened 2026-08-18
test(runner): synchronize workspace cache lock retry

### LINKED ISSUE #27765: test(runner): synchronize transient workspace cache lock release under suite load

seven332:
## Summary

- replace the workspace-cache executor test's permission polling and controller thread with an instance-scoped one-shot test gate
- signal the gate only after the first real lock acquisition returns busy, then release the held lock before the existing contention timeout starts
- keep production lock/cache behavior, the 50 ms timeout, retry cadence, workspace configuration, and telemetry unchanged

## Scope

The gate, cache injection field, builder, and gated entry are compiled only for runner tests. Production and test wrappers share the same timeout acquisition implementation. This PR does not change APIs, protocols, persisted state, migrations, deployment behavior, or runtime timeout values.

Closes #27765

Parent context: #27738

## Validation

- target test repeated 20 times:
  `cargo test -q -p runner executor::tests::sandbox_run_tests::workspace_cache::execute_inner_waits_for_transient_workspace_cache_lock -- --exact`
- `cargo test -p runner workspace_image_cache::tests::lifecycle`
- `cargo test -q -p runner` (3 complete runs; each 3,160 passed, 20 ignored; guest inventory passed)

### DIFF
--- a/crates/runner/src/executor/tests/sandbox_run_tests/workspace_cache.rs
+++ b/crates/runner/src/executor/tests/sandbox_run_tests/workspace_cache.rs
@@ -2,9 +2,6 @@ use super::*;
 use async_trait::async_trait;
 use httpmock::prelude::*;
 use sha2::{Digest, Sha256};
-use std::os::unix::fs::PermissionsExt;
-use std::sync::mpsc;
-use std::time::Instant;
 
 use crate::executor::session_history_cpu::{SessionHistoryCpuPool, SessionHistoryCpuTestGate};
 use crate::executor::tests::support::RUN_IN_SANDBOX_TEST_TIMEOUT;
@@ -14,7 +11,9 @@ use crate::types::{
     ResumeSessionHistory, ResumeSessionHistoryEncoding, ResumeSessionHistoryRef,
     ResumeSessionHistoryRefKind, WorkspaceReuseResult,
 };
-use crate::workspace_image_cache::WorkspaceSessionHistorySidecarRepresentation;
+use crate::workspace_image_cache::{
+    WorkspaceImagePrepareLockTestGate, WorkspaceSessionHistorySidecarRepresentation,
+};
 
 fn enable_api_start_telemetry(context: &mut crate::types::ExecutionContext) {
     context.api_start_time = Some(chrono::Utc::now().timestamp_millis().max(0) as u64);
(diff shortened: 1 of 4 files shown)


seven332:
## Code Review: PR #27831

**Reviewed head:** `52b6818669f15d72ac7b9c7f3c08b504be2f119a`

### Summary

The instance-scoped test gate now proves the first real flock attempt returned busy before releasing the held lock. The normal retry path, 50 ms production timeout, cache-hit configuration, and telemetry remain unchanged.

### Findings

No findings.

### Testing

Reviewed the executor integration path, shared lock helper, test-only cache seam, cancellation/resource ownership, and submitted validation. The target test passed 20 repeated runs; three complete runner-suite runs plus Rust format, Clippy, and documentation checks passed. CI is still in progress.

### Verdict

LGTM

PR #27849, opened 2026-08-18
test(python): cover static base path segment boundaries

### LINKED ISSUE #27838: test(python): cover static base path segment boundaries
## Problem

`_match_compiled_base_url_parts` in `crates/runner/mitm-addon/src/firewall_matching/base_url.py:617-624` enforces that a static base path matches only the exact path or a descendant path segment. The current low-level coverage in `crates/runner/mitm-addon/tests/test_matching_base_url_static.py:8-58` checks exact paths, case sensitivity, terminal slashes, and repeated separators, but does not pin the sibling-prefix boundary.

During mutation testing at commit `04730c8ef3b98678a1cc3af5c9f1f89f4b26b7e6`, removing the `rest[0] != "/"` rejection made:

```python
match_base_url(
    "https://api.example.com/v10/items",
    "https://api.example.com/v1",
)
```

return `("0/items", {})` instead of `None`. All 6,001 Python tests still passed, including 649 focused base URL and firewall-matching tests.

## Impact

A regression in this guard could classify a sibling path such as `/v10` as belonging to the `/v1` API base. Because `crates/runner/mitm-addon/src/matching.py:1737-1774` uses this result to select and evaluate firewall API candidates, rules and authentication handling could be applied outside the intended path scope.

## Acceptance criteria

- Add direct static matcher cases proving that base `/v1` rejects `/v10` and other same-prefix sibling segments.
- Preserve positive coverage for the exact `/v1` path and descendants such as `/v1/items`.
- Add a compiled firewall request regression showing that a sibling-prefix URL is not evaluated under the `/v1` API base.
- Ensure the new tests fail if the path-segment boundary guard at `base_url.py:621-622` is removed.

seven332:
## Summary

Add regression tests that pin the static base-path segment boundary so a sibling path such as `/v10` cannot be classified under the `/v1` API base.

- Low-level `match_base_url` cases reject `/v10`, `/v10/items`, `/v1foo`, and `/v1.0` against base `/v1`, while preserving positive coverage for the exact `/v1` path and descendant `/v1/items`.
- A compiled firewall request regression asserts a sibling-prefix URL is not evaluated under the `/v1` base in both the indexed and linear matchers.

## Validation

Run from `crates/runner/mitm-addon`:

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (6008 passed)

## Notes

- Test-only change; no production source is modified.
- The new cases fail when the `rest[0] != "/"` guard in `src/firewall_matching/base_url.py` is removed.

Closes #27838

### DIFF
--- a/crates/runner/mitm-addon/tests/test_compiled_firewall_base_path_matching.py
+++ b/crates/runner/mitm-addon/tests/test_compiled_firewall_base_path_matching.py
@@ -374,6 +374,41 @@ def test_compiled_matches_static_base_boundary_and_query():
     assert compiled is None
 
 
+def test_compiled_static_base_rejects_sibling_path_segment():
+    fws = wrap_firewalls(
+        [
+            {
+                "base": "https://api.example.com/v1",
+                "auth": {"headers": {"Authorization": "Bearer token"}},
+                "permissions": [
+                    {"name": "read", "rules": ["ANY /{path*}"]},
+                ],
+            }
+        ],
+        name="example",
+    )
+    policies = {"example": {"allow": ["read"], "deny": [], "unknownPolicy": "deny"}}
+    compiled_firewalls = compile_firewalls_or_fail(fws)
+
+    result = matching.match_compiled_firewall_request(
+        "https://api.example.com/v10/items",
+        "GET",
+        compiled_firewalls,
+        policies,
+    )
+
+    assert result is None
+
+    linear_result = matching._match_compiled_firewall_request_linear(
+        "https://api.example.com/v10/items",
+        "GET",
+        compiled_firewalls,
+        policies,
+    )
+
+    assert linear_result is None
+
+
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #27849

**Reviewed head:** `0233b1e90d77c4b490d6325d2a75d0ae555fde37`

### Summary

Adds regression tests pinning the static base-path segment boundary so `/v10` cannot match the
`/v1` base. Low-level `match_base_url` cases cover sibling-prefix rejection and exact/descendant
positive coverage; a compiled matcher regression asserts the sibling URL is not evaluated.

### Findings

No findings.

### Testing

- Focused and full Python suites pass (`6008 passed`), plus `ruff format --check`, `ruff check`,
  and `basedpyright`.
- Mutation check confirms the new cases fail when `rest[0] != "/"` is removed.

### Verdict

LGTM

PR #27864, opened 2026-08-18
refactor(python): share builtin base url template layout

### LINKED ISSUE #27834: refactor(python): share builtin base URL template layout
## Problem

Python classifies each built-in firewall base URL template's variable positions twice.

- `crates/runner/mitm-addon/src/builtin_firewall_cache.py:449-564` scans the raw template, computes authority/path/query boundaries, and classifies each reference as a whole base, host/authority fragment, port, or path position so catalog validation can substitute a component-specific witness.
- `crates/runner/mitm-addon/src/builtin_base_url.py:287-373` independently derives the same position categories from each reference's prefix and suffix before validating and substituting execution-supplied values.
- `crates/runner/mitm-addon/src/connector_template_syntax.py:34-44` centralizes only lexical reference recognition. It exposes spans but not the URL-component layout that both consumers reconstruct.

The cache and runtime intentionally have different trust work: cache loading validates unresolved catalog syntax with witnesses, while runtime resolution validates untrusted values and the final URL. However, deciding which URL component a reference occupies is common syntax/layout semantics, not phase-specific policy.

## Evidence of drift and maintenance cost

#26168 fixed a concrete false rejection at the cache boundary: the cache replaced every variable with one generic `template` witness, so a valid whole-host template with a parameterized path was rejected even though runtime resolution produced a valid base. The fix added the current cache-side component classifier instead of reusing the position model already present in `builtin_base_url.py`.

The same boundary has already required coordinated changes:

- #20973 updated both cache-time and runtime handling for unsafe fixed and resolved paths.
- #24807 updated both consumers after their connector-template reference grammars diverged, resulting in the shared lexical scanner.
- #25339 changed runtime authority-fragment classification; the cache classifier introduced later by #26168 now independently represents that category.

This leaves the next supported or corrected position rule dependent on two condition trees staying aligned. If cache validation is stricter than runtime resolution, a valid catalog refresh can become `cache_invalid` and keep connectors unavailable or pinned to a last-good catalog. If it is looser, a catalog can load even though every affected VM entry fails later during resolution.

## Acceptance criteria

- Introduce one value-independent base URL template layout analysis that returns each simple `vars` reference with its source span and canonical component kind, covering whole-base/prefix, whole authority, authority fragment, port, and path positions.
- Use that shared layout in both catalog witness materialization and runtime value validation/substitution.
- Remove the duplicate component-boundary and prefix/suffix classification logic from `builtin_firewall_cache.py` and `builtin_base_url.py`.
- Keep phase-specific policy separate: catalog loading must still validate unresolved fixed syntax, while runtime resolution must still validate untrusted values, the fully resolved base, credentialed HTTPS, and host policy.
- Preserve the exact ECMAScript-compatible reference grammar owned by `connector_template_syntax.py` and fail closed for malformed, unsupported, query, and fragment positions.
- Extend the shared contract so every accepted template-resolution layout is also accepted at the catalog phase, while phase-specific invalid-value cases remain runtime failures.

Verified against `main` at `04730c8ef3b98678a1cc3af5c9f1f89f4b26b7e6`.

seven332:
## Summary

- add one value-independent analyzer for built-in base URL template reference spans, component kinds, and authority-fragment structure
- use the shared layout for catalog witness materialization and runtime value validation/substitution while preserving phase-specific safety policy
- extend the shared TypeScript/Python contract so every successful runtime layout, including fragmented ports and IPv6 host fragments, is also admitted by the catalog cache

## Compatibility

- no API, catalog schema, cache schema, VM payload, persisted-state, database, or migration changes
- preserves fail-closed malformed and unsupported template handling
- preserves runtime value, complete URL, credentialed HTTPS, and host-policy enforcement
- all changes ship within one runner artifact; no cross-version protocol is introduced

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_firewall_base_url_validation_contract.py tests/test_registry_builtin_base_url_vars.py tests/test_registry_builtin_catalog_validation.py` (224 passed)
- `uv run --no-sync python -m pytest tests/` (6,009 passed)
- `pnpm exec prettier --check packages/connectors/src/__tests__/firewall-base-url-validation-contract.json`
- `pnpm -F @okouai/connectors lint`
- `pnpm -F @okouai/connectors check-types`
- `pnpm -F @okouai/connectors test` (1,037 passed)
- `pnpm knip --workspace @okouai/connectors`
- staged-path pre-commit checks: Prettier, full Turbo Knip, Ruff, file-size check, and commitlint

Closes #27834

### DIFF
--- a/crates/runner/mitm-addon/src/builtin_base_url.py
+++ b/crates/runner/mitm-addon/src/builtin_base_url.py
@@ -11,7 +11,7 @@ transport and built-in host-policy validation are a subsequent registry step.
 import re
 import urllib.parse
 
-import connector_template_syntax
+import builtin_base_url_template
 import matching
 from authority_utils import percent_decode_host
 from path_security import has_unsafe_path, has_unsafe_url_path
@@ -284,30 +284,13 @@ def _validate_base_url_path_variable(
         )
 
 
-def _prefix_is_inside_authority(prefix: str) -> bool:
-    scheme_end = prefix.find("://")
-    if scheme_end == -1:
-        return False
-    after_scheme = prefix[scheme_end + 3 :]
-    return _URL_COMPONENT_DELIMITER_PATTERN.search(after_scheme) is None
-
-
-def _prefix_is_inside_path(prefix: str) -> bool:
-    scheme_end = prefix.find("://")
-    if scheme_end == -1:
-        return False
-    after_scheme = prefix[scheme_end + 3 :]
-    if "?" in after_scheme or "#" in after_scheme:
-        return False
-    return "/" in after_scheme
-
-
 def _validate_base_url_template_variable(
     *,
     firewall_name: str,
     base: str,
     name: str,
     value: str,
+    kind: builtin_base_url_template.BaseUrlTemplateComponentKind,
     prefix: str,
     suffix: str,
 ) -> None:
@@ -317,9 +300,9 @@ def _validate_base_url_template_variable(
         name=name,
         value=value,
     )
-    if prefix == "" and suffix == "":
-        return
-    if prefix == "" and suffix.startswith("/"):
+    if kind == "whole-base":
+        if suffix == "":
+            return
         _validate_base_url_prefix_variable(
             firewall_name=firewall_name,
             base=base,
@@ -327,7 +310,7 @@ def _validate_base_url_template_variable(
             value=value,
         )
         return
-    if prefix.endswith("://") and (suffix == "" or suffix.startswith("/")):
+    if kind == "whole-authority":
         _validate_base_url_authority_variable(
             firewall_name=firewall_name,
             base=base,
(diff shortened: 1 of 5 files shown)


seven332:
## Code Review: PR #27864

**Reviewed head:** `4bb79f65844512999928467b29b6536e8c00e6f0`

### Summary

The shared analyzer now owns exact template-reference placement and source spans, while catalog witnesses and runtime value/final-URL/HTTPS/host-policy checks remain at their existing trust boundaries. Fragmented ports and bracketed IP-literal fragments are aligned across catalog and runtime, and component boundaries are precomputed once so the analyzer stays linear.

### Findings

No findings.

### Testing

- Full mitm-addon validation passed, including Ruff, Basedpyright, and 6,009 Pytest cases.
- Connectors Prettier, lint, type checking, 1,037 Vitest cases, and Knip passed.
- Shared contract cases exercise every supported layout family through runtime resolution and send every successful layout through real catalog loading.
- Production catalog refresh timing was not exercised; no external API, schema, persisted-state, or rollout boundary changes.

### Verdict

LGTM

PR #27932, opened 2026-08-18
perf(python): batch firewall auth query injection

### LINKED ISSUE #27899: perf(python): avoid quadratic auth query injection
## Problem

`_apply_header_query_injection()` applies resolved firewall-auth query parameters one at a time:

```python
for key, value in resolved_query.items():
    flow.request.query[key] = value
```

With the pinned `mitmproxy==12.2.3`, `Request.query` creates a `MultiDictView`. Every `__setitem__` call reads and parses the entire current URL query, scans its fields, serializes the growing query, reparses the URL, and writes `Request.path`. Injecting `k` resolved entries therefore performs quadratic work and allocation.

This helper runs synchronously after auth resolution in both the ordinary request path and the header-phase/requestheaders path, so a large auth-query map blocks the mitmproxy event loop and delays unrelated traffic.

## Evidence

At commit `55f40212167b6d9f87e9850d759330fbb9c336fe`, a microbenchmark using the repository's pinned mitmproxy version and an actual `mitmproxy.http.Request` measured the existing loop:

| Resolved query entries | JSON payload size | Existing loop |
| ---: | ---: | ---: |
| 100 | 1,309 bytes | 13.5 ms |
| 500 | 6,509 bytes | 289.4 ms |
| 1,000 | 13,009 bytes | 1.128 s |
| 2,000 | 27,009 bytes | 4.505 s |

The same final query assembled from one parsed snapshot and serialized once took about 2.7 ms for 2,000 entries. The roughly 4x growth when entry count doubles is consistent with the repeated whole-query processing.

This input is permitted by current production bounds:

- The firewall-auth response body cap is 256 KiB, so the 27 KiB example is well below it.
- Connector, catalog, and webhook schemas model `auth.query` as an unbounded string record.
- The Python response parser validates string keys/values and exact configured-key correspondence, but does not bound the number of entries.

Related issue #24686 optimized no-op tokenization in the separate `auth.base` rewrite path. This issue concerns non-`auth.base` resolved query injection and scales with the number of configured query entries.

## Expected outcome

- Build the final request query with linear work: parse/snapshot the existing multi-value query once and serialize/write it once, independent of the number of resolved keys.
- Preserve current `MultiDict.set` behavior: each resolved value replaces all existing occurrences of its key while retaining the first occurrence's position; previously absent keys are appended in resolved-map order; unrelated duplicate keys, blank values, path parameters, and fragments remain intact.
- Cover both the ordinary request path and the header-phase/requestheaders path through the shared helper.
- Add deterministic regression coverage that proves query parsing/serialization does not grow with the resolved-key count (for example, by counting query getter/setter activity), without a wall-clock threshold.
- Keep the existing auth-query injection tests passing.

## Relevant code

- `crates/runner/mitm-addon/src/auth.py:579-589`
- `crates/runner/mitm-addon/src/auth.py:1511-1516`
- `crates/runner/mitm-addon/src/auth.py:1742-1747`
- `crates/runner/mitm-addon/src/firewall_auth_client.py:681-701`
- `crates/runner/mitm-addon/src/firewall_auth_client.py:764-810`
- `turbo/packages/connectors/src/firewall-types.ts:42-74`
- `turbo/packages/api-contracts/src/contracts/webhooks.ts:525-537`

seven332:
## Summary

- merge resolved firewall-auth query entries in one linear pass
- write the mitmproxy request query once while preserving replacement, duplicate, ordering, blank-value, path-parameter, and fragment semantics
- cover both ordinary request auth and requestheaders auth with deterministic query read/write assertions

## Scope

- no API, schema, persistence, protocol, dependency, feature-switch, or rollout changes
- no new auth-query entry limit and no change to header injection

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_auth_query_injection.py` (9 passed)
- `uv run --no-sync python -m pytest tests/` (6037 passed)
- `git diff --check`

Closes #27899

### DIFF
--- a/crates/runner/mitm-addon/src/auth.py
+++ b/crates/runner/mitm-addon/src/auth.py
@@ -585,8 +585,15 @@ def _apply_header_query_injection(
     for header_name, header_value in resolved_auth_header_pairs(headers):
         flow.request.headers[header_name] = header_value
     if resolved_query:
-        for key, value in resolved_query.items():
-            flow.request.query[key] = value
+        remaining_query = resolved_query.copy()
+        merged_query: list[tuple[str, str]] = []
+        for key, value in flow.request.query.fields:
+            if key in remaining_query:
+                merged_query.append((key, remaining_query.pop(key)))
+            elif key not in resolved_query:
+                merged_query.append((key, value))
+        merged_query.extend(remaining_query.items())
+        flow.request.query = merged_query
 
 
 def _trusted_aws_sigv4_url(flow: http.HTTPFlow) -> str:
--- a/crates/runner/mitm-addon/tests/test_auth_query_injection.py
+++ b/crates/runner/mitm-addon/tests/test_auth_query_injection.py
@@ -3,9 +3,13 @@
 from unittest.mock import AsyncMock, patch
 from urllib.parse import parse_qs, urlparse
 
+import pytest
+from mitmproxy import http
+
 import auth
 import flow_metadata_keys as metadata_keys
 from tests.firewall_auth_helpers import (
+    apply_requestheaders_auth_without_upstream_admission,
     handle_firewall_request_without_upstream_admission,
     make_allow,
 )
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #27932

**Reviewed head:** `9c60cf8c4bc397d6a32cf1a141fa5b3ce3bdedb1`

### Summary

The shared firewall-auth helper now merges resolved query entries in linear time and writes the mitmproxy query once. The change preserves the old sequential replacement and ordering behavior across both request and requestheaders auth paths without changing external contracts.

### Findings

No findings.

### Testing

The focused integration test passes for both auth phases and asserts exact ordered query semantics plus one full read/write. Ruff, basedpyright, all 6,037 addon tests, a 2,000-case old/new differential probe, and a pinned-dependency scale comparison passed locally. GitHub Actions status is evaluated separately by the PR check stage.

### Verdict

LGTM

PR #27983, opened 2026-08-18
perf(guest): isolate receipt journal fsyncs from tokio workers

### LINKED ISSUE #27963: perf(rust): keep active-input receipt fsyncs off Tokio workers
## Problem and evidence

Active-input receipt persistence performs blocking durability I/O directly on Tokio scheduler workers.

Production enters the asynchronous guest-agent `run` function on a multi-thread Tokio runtime. Receipt startup synchronously reads and validates the journal from that path. After a backend accepts a follow-up, the CLI path synchronously publishes a durable acceptance snapshot; after the API acknowledges the receipt, the async receipt worker synchronously compacts the snapshot.

Each publication writes an exclusive `0600` replacement, calls `File::sync_all()`, atomically renames it, and synchronizes the parent directory. One successful follow-up therefore performs two complete durability cycles. Slow or contended storage can delay unrelated guest-agent futures such as heartbeat, cancellation, and process supervision.

## Decision

Proceed as one Rust PR using Tokio `block_in_place` around the existing synchronous journal critical sections.

This is preferable to a direct `spawn_blocking(...).await` conversion because the current acceptance transaction has no cancellation point between durable publication, upload enqueue, and live accepted-state publication. The CLI writer can be aborted after its bounded quiescence wait; an awaited blocking task could continue after that abort while caller-owned steps do not. A persistence actor can solve that split, but it requires new command, state-ownership, and shutdown coordination that the current bounded workload does not justify.

## Chosen direction

- Add one private guest-agent adapter that runs finite journal I/O with `block_in_place` on the production multi-thread runtime.
- Execute directly when no Tokio runtime exists for synchronous tests, where no scheduler can be blocked.
- Return a contextual I/O error if enabled receipt persistence is invoked on a current-thread Tokio runtime instead of allowing `block_in_place` to panic.
- Route the full startup load, acceptance mutation, and acknowledgement mutation through the adapter.
- Keep the existing journal mutex around each complete snapshot mutation so overlapping acceptance and acknowledgement retain their order.
- Keep all `guest-contracts` private-file, validation, atomicity, and durability code unchanged.
- Keep acceptance persistence synchronous and non-cancellable from the caller's perspective; do not introduce channels, detached writes, or new awaits.

## Alternatives considered

- **Direct `spawn_blocking` per operation:** simpler API bridge and supports current-thread runtimes, but introduces a cancellation/finalization race unless transaction ownership and lifecycle tracking are added.
- **Serialized Tokio persistence actor:** can be cancellation-safe with sufficient ownership and barriers, but adds a second worker protocol and shutdown ordering across persistence and HTTP receipt delivery.
- **Dedicated OS thread:** provides strict serialization but adds a permanent thread and command lifecycle for short, bounded operations.
- **Async filesystem rewrite:** would require rebuilding the descriptor-relative private-file security and atomicity guarantees and provides no benefit over isolating the existing correct helper.
- **Remove or weaken synchronization:** rejected because it breaks crash-recovery semantics.

## Risks and mitigations

- `block_in_place` requires a multi-thread runtime. Production already guarantees this; receipt-enabled async tests will use the same flavor, and an unexpected current-thread caller receives an error.
- Same-task sibling futures pause during `block_in_place`. Current journal calls are not inside `join!`; heartbeat, supervision, CLI writing, and receipt delivery are independently spawned tasks.
- Acceptance and acknowledgement can overlap and both enter blocking mode. One CLI writer, one serial receipt worker, and the existing mutex bound this concurrency per guest process.
- A timing-only test could pass without exercising the boundary. The regression will block the real journal mutation on its existing mutex in a one-worker runtime, start the unrelated probe only after the mutation task begins, then release and verify the real snapshot.
- A failed scheduler assertion could otherwise hang runtime shutdown. The test will release and join all blocked work before asserting.

## Acceptance criteria

- Startup journal reads and complete acceptance/acknowledgement snapshot mutations execute inside the Tokio blocking region rather than monopolizing a scheduler worker.
- An independently spawned async probe makes progress while a real journal mutation is deterministically blocked at the journal serialization boundary on a one-worker runtime.
- The regression completes the mutation after release and verifies the real snapshot, without mocked storage or sleeps.
- Acceptance remains durable before upload enqueue and before live state becomes accepted.
- Acceptance and acknowledgement remain serialized without lost outstanding IDs.
- Startup and acceptance errors still propagate; acknowledgement and finalization remain best effort.
- Existing permission, symlink, bounded-read, atomic replacement, compaction, cancellation, framework, and restart-recovery tests continue to pass.
- The persisted journal format and guest/runner protocol do not change.

## Delivery notes

No migration, feature switch, compatibility bridge, or data cleanup is required. Runner recovery continues to parse the unchanged journal. If implementation discovers a current-thread production caller or same-task concurrent journal usage, stop and revisit the actor design instead of adding a scheduler-blocking fallback.

seven332:
## Summary

- run active-input receipt journal loads and complete snapshot mutations inside Tokio blocking regions
- preserve the existing synchronous acceptance transaction, journal mutex ordering, and durable file format
- exercise receipt-enabled async tests on the production multi-thread runtime and add a deterministic scheduler-progress regression

## Scope

- no journal format, guest/runner protocol, API, migration, or rollout changes
- current-thread runtimes now receive a contextual I/O error instead of a `block_in_place` panic

## Validation

- `cargo clippy -p guest-agent --all-targets --all-features`
- `cargo test -p guest-agent --all-features`
- repository pre-commit Rust formatting, Clippy, and documentation hooks

Closes #27963

### DIFF
--- a/crates/guest-agent/src/active_input/tests.rs
+++ b/crates/guest-agent/src/active_input/tests.rs
@@ -194,7 +194,7 @@ fn active_input_capacity_counts_pending_inputs() {
     ));
 }
 
-#[tokio::test]
+#[tokio::test(flavor = "multi_thread")]
 async fn active_input_can_release_capacity_for_sink_without_replay() {
     let runtime = enabled_runtime();
     let controller = runtime.controller();
--- a/crates/guest-agent/src/active_input/tests.rs
+++ b/crates/guest-agent/src/active_input/tests.rs
@@ -194,7 +194,7 @@ fn active_input_capacity_counts_pending_inputs() {
     ));
 }
 
-#[tokio::test(flavor = "multi_thread")]
+#[tokio::test(flavor = "multi_thread", worker_threads = 1)]
 async fn active_input_can_release_capacity_for_sink_without_replay() {
     let runtime = enabled_runtime();
     let controller = runtime.controller();
--- a/crates/guest-agent/src/active_input_receipts.rs
+++ b/crates/guest-agent/src/active_input_receipts.rs
@@ -316,7 +316,7 @@ mod tests {
 
     const RUN_ID: &str = "receipt-journal-scheduler-test";
     const DELIVERY_ID: &str = "60fca608-d174-4c1a-a1b2-57607b3adf46";
-    const WAIT_TIMEOUT: Duration = Duration::from_secs(2);
+    const WAIT_TIMEOUT: Duration = Duration::from_secs(5);
 
     #[test]
     fn blocked_journal_persistence_does_not_stall_unrelated_async_work() {
--- a/crates/guest-agent/src/active_input_receipts.rs
+++ b/crates/guest-agent/src/active_input_receipts.rs
@@ -16,6 +16,23 @@ use crate::http::HttpClient;
 
 const LOG_TAG: &str = "sandbox:guest-agent";
 
+fn run_journal_io<T>(operation: impl FnOnce() -> io::Result<T>) -> io::Result<T> {
+    match tokio::runtime::Handle::try_current() {
+        Ok(handle)
+            if matches!(
+                handle.runtime_flavor(),
+                tokio::runtime::RuntimeFlavor::MultiThread
+            ) =>
+        {
+            tokio::task::block_in_place(operation)
+        }
+        Ok(_) => Err(io::Error::other(
+            "active-input receipt persistence requires a multi-thread Tokio runtime",
+        )),
+        Err(_) => operation(),
+    }
+}
+
 #[derive(Debug)]
 struct ReceiptJournalState {
     outstanding: Vec<String>,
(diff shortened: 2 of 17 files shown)


seven332:
## Code Review: PR #27983

**Reviewed head:** `62ac80aa00dc7b3aac27fb3717e13c964722175a`

### Summary

The blocking boundary covers journal startup, acceptance publication, and acknowledgement compaction while preserving the existing mutex ordering, synchronous acceptance transaction, durable file format, and error semantics. Receipt-enabled tests use bounded production-flavor runtimes, and the current-thread constraint is covered through the public constructor.

### Findings

No findings.

### Testing

- `cargo clippy -p guest-agent --all-targets --all-features`
- `cargo test -p guest-agent --all-features --quiet`
- repository pre-commit Rust formatting, Clippy, and documentation hooks

The persisted journal format and guest/runner protocol are unchanged, so no cross-version fixture or migration is required.

### Verdict

LGTM

PR #26272, opened 2026-08-11
feat(db): add presentation_templates

bingjiez666:
Closes #26235. Part of #21350.

The single table behind uploaded presentation templates. Schema only — no routes, no UI, nothing reads or writes it yet, so this is additive and safe to land ahead of the API.

## What lands

- `packages/db/src/schema/presentation-template.ts`
- `packages/db/src/jsonb-contracts/presentation-template.ts` — the `error` payload, `{code, message}`, mirroring `BuiltInGenerationError`
- migration `0891_presentation_templates.sql` plus its snapshot, generated with `drizzle-kit generate`
- registration in `packages/db/src/index.ts`

```sql
CREATE TABLE "presentation_templates" (
	"id" uuid PRIMARY KEY DEFAULT gen_random_uuid() NOT NULL,
	"org_id" text NOT NULL,
	"owner_user_id" text NOT NULL,
	"visibility" varchar(16) DEFAULT 'private' NOT NULL,
	"title" text NOT NULL,
	"status" varchar(16) DEFAULT 'pending' NOT NULL,
	"error" jsonb,
	"source_storage_key" text NOT NULL,
	"source_filename" text NOT NULL,
	"page_keys" text[] DEFAULT '{}'::text[] NOT NULL,
	"aspect_ratio" real,
	"created_by" text NOT NULL,
	"updated_by" text NOT NULL,
	"created_at" timestamp DEFAULT now() NOT NULL,
	"updated_at" timestamp DEFAULT now() NOT NULL,
	CONSTRAINT "chk_presentation_templates_visibility" CHECK (... IN ('private', 'public')),
	CONSTRAINT "chk_presentation_templates_status" CHECK (... IN ('pending', 'processing', 'ready', 'failed'))
);
CREATE INDEX "idx_presentation_templates_owner_created" ON "presentation_templates" ("org_id","owner_user_id","created_at" DESC);
CREATE UNIQUE INDEX "idx_presentation_templates_active_import" ON "presentation_templates" ("owner_user_id") WHERE status IN ('pending', 'processing');
```

## Decisions worth reviewing

**No column points at the compiled package.** It is the storage named `presentation-template@{id}`, derived from the row id the same way `zero_workflows` derives `custom-skill@{workflowId}` (`zero-workflows.ts:449`). A column would be a second copy of the same fact and could drift.

[...]

PR #28016, opened 2026-08-18
refactor(db): drop the presentation template active import index

bingjiez666:
## What

Drops the partial unique index `idx_presentation_templates_active_import` on `presentation_templates(owner_user_id) WHERE status IN ('pending', 'processing')`.

## Why

The index restricted every user to one pending or processing template import at a time. It was added in #26272 to stop a repeated upload click from launching several conversion runs. That reason no longer holds:

- **Repeated submissions are already idempotent above the database.** The import commit route derives the template id from `(orgId, ownerUserId, requestId)`, so resubmitting the same upload resolves to the same row instead of starting a second run.
- **Run cost and concurrency belong to the run layer.** Template analysis runs are created with `queueOnConcurrencyLimit` and `enforceVm0Credits`, which already queue past the concurrency limit and refuse without credits.
- **What the index actually blocked is ordinary use.** Two different `requestId`s means two different decks — a user importing several decks at once is not a mistake.

It also constrains how the remaining import work can be split into reviewable PRs, because any intermediate state that leaves a row at `pending` locks that user out of importing again.

## Risk

None in production. The table is gated behind the presentation templates feature switch and has no client, so no rows depend on the index. Dropping an index is reversible.


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 9746089f6dbc2d2f957dc6dc2ccd228fec8b579c
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Drops the partial unique index `idx_presentation_templates_active_import` on `presentation_templates(owner_user_id) WHERE status IN ('pending','processing')`, removing the database-level "one active import per user" restriction. The change is the schema declaration plus one generated Drizzle migration (`0942_shocking_captain_stacy.sql`, a single `DROP INDEX`) and its snapshot/journal entries; no application code changes.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic — no race conditions, wrong conditionals, or bad assumptions | Pass | `presentation-template.ts` removes only the `uniqueIndex(...)` block; `index("idx_presentation_templates_owner_created")` and both `check` constraints are untouched, and `0942_shocking_captain_stacy.sql` contains exactly the matching `DROP INDEX`. |
| `REVIEW.md` | Security — no injection, hardcoded secrets, or missing auth on new endpoints | N/A | No endpoints, auth paths, or credential handling in the diff; it is a schema/migration-only change. |
| `REVIEW.md` | Type safety — no `any`, no overly broad types | Pass | Only an import removal (`uniqueIndex`); `sql` remains imported and used by the two `check` constraints and the `text[]` default, so no unused-import or type regression. |
| `REVIEW.md` | Style & maintainability — no oversized functions, duplication, or WHAT-comments | Pass | Net −6 lines in the schema; the removed comment described the deleted constraint and correctly goes with it. |
| `REVIEW.md` | DB/JSONB review gate — persisted contract changes need migration/normalizer or a no-migration rationale | Pass | The schema change ships with its generated migration; dropping an index changes no row shape, so no backfill or read-time normalizer applies. No `jsonb().$type<>()` change in the diff. |
| `REVIEW.md` | Testing coverage — `refactor:` commits must keep existing coverage | Pass | `refactor(db)` commit. `turbo/apps/api/src/signals/routes/__tests__/zero-presentation-templates.test.ts` still covers the surviving routes; no test asserted the one-active-import behavior, so no coverage is lost. |
| `REVIEW.md` | Feature regression guard — changed existing behavior needs containment and regression coverage | Pass | The affected surface is gated by `FeatureSwitchKey.PresentationTemplates`, which is `enabled: false` with no `enabledOrgIdHashes` in `turbo/packages/core/src/feature-switch.ts:310`, so no user flow is rerouted. |
| `docs/bad-smell.md` | Zero tolerance for `any` and for lint/type suppressions | Pass | No `any`, no `eslint-disable` / `@ts-expect-error` / `prettier-ignore` anywhere in the diff. |
| `docs/bad-smell.md` | Fail fast — no fallback or recovery logic, no hardcoded config | Pass | The diff removes a guard rather than adding one; no `??`/`||` chains, defaults, or hardcoded URLs introduced. |
| `docs/fallback.md` §2 | A feature behind a non-GA feature switch needs no compatibility code, migration, or backfill | Pass | `PresentationTemplates` is `enabled: false` with no staff org hashes, so the table has no external users and the PR correctly adds no compatibility path. |
| `docs/fallback.md` §9 | Every fallback introduced by the diff must be declared in the PR summary | N/A | The diff introduces no fallback. Per §9, a PR that only removes an existing guard needs neither a `Fallbacks` section nor `Fallbacks: none`. |
| `docs/fallback.md` §10 | Removing a guard requires type/schema, single-writer, production, or rollout evidence, and must delete the branch and its own tests together | Pass | Verified evidence: `turbo/packages/api-contracts/src/contracts/zero-presentation-templates.ts` declares only GET/GET/PATCH/DELETE, and no `insert(presentationTemplates)` or `onConflict` exists anywhere in `turbo/apps/api`, so the partial index currently constrains no writer. Nothing else references the index name, and no test covered it, so there is no orphaned branch or test left behind. |
| `docs/fallback.md` §1 | No negative tests asserting removed behavior stays removed | Pass | The PR adds no tests at all, so no tombstone asserting the dropped index is still gone. |
| `docs/deployment-compatibility.md` | Old code after migration — every statement the draining API can issue must stay legal | Pass | An index declaration never appears in Drizzle-generated `SELECT`/`INSERT`/`RETURNING` column lists, and the only `42P10` risk — an `ON CONFLICT` targeting this partial index — does not exist: there is no insert against `presentationTemplates` in the codebase. The draining API keeps working after the drop. |
| `docs/deployment-compatibility.md` | New code before migration — new readers/writers must not require the not-yet-applied schema | Pass | New code requires nothing that the drop provides; while the index still exists, the only way to observe it is a second concurrent pending/processing insert, and no insert path exists. Both directions are safe. |
| `docs/deployment-compatibility.md` | Destructive clean-up migrations must run only after readers needing the old schema have drained | Pass | No reader depends on the index. Dropping an index is also reversible by recreating it, so the rollback path is intact. |
| `docs/testing.md` | Write integration tests, mock only at the system boundary, use the real database | N/A | The PR changes no test files; no mock boundary, MSW, fake-timer, or DB-mocking question arises. |
| `docs/testing.md` | Deployment-compatibility tests when a change touches a persisted boundary | Pass | The relevant old/new pairing is exercised by the existing route tests against a real database with the migration applied; a dedicated cross-version test would have no behavior to assert, since the constraint has no writer on either side. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Notes (non-blocking)

- PR description: the rationale states that "the import commit route derives the template id from `(orgId, ownerUserId, requestId)`". That route is not on `main` yet — the contract currently exposes only list/get/patch/delete, and the table has no writer. The removal is still well-evidenced without it (the feature switch is `enabled: false` for everyone, and no code inserts into the table), but the stated reason describes work that has not landed.
- `0942_shocking_captain_stacy.sql` uses a plain `DROP INDEX`, which takes an `ACCESS EXCLUSIVE` lock on the table, rather than the `DROP INDEX CONCURRENTLY IF EXISTS` + `-- vm0:non-transactional` pattern of migration `0778`. This is fine here because the table is behind an off-by-default switch with no writers, and the statement is a fast catalog update; noting it only so the choice is deliberate.

### Testing

- Coverage: Adequate. `refactor(db)` commit with no behavior change reachable by any caller; `zero-presentation-templates.test.ts` continues to cover the routes that touch this table, and no existing test depended on the dropped index.
- Conventions: Compliant. No test files changed, so no mock-boundary, MSW, real-database, fake-timer, or mock-cleanup rules apply.
- Migration hygiene: The regenerated migration is sequenced after `main`'s latest entry (`0941_silly_argent`), `meta/_journal.json` adds only `idx: 942`, and `meta/0942_snapshot.json` no longer contains `idx_presentation_templates_active_import`, so the schema declaration and snapshot agree. Regeneration produced exactly one statement, confirming no unrelated schema drift was swept in.

PR #28037, opened 2026-08-18
test(python): cover runner flush signal at worker exit

### LINKED ISSUE #28027: test(python): cover runner flush signal at worker exit
## Problem

The successor-worker handoff in `crates/runner/mitm-addon/src/runner_flush_lifecycle.py:180-192` has no deterministic regression test.

`_run_usage_flush_worker()` drains the level-triggered request flag, releases `_usage_flush_signal_lock`, and then starts a successor worker when a signal arrived at the exit boundary. That final check is the only path that handles this ordering:

1. The worker has already observed `_usage_flush_requested == False` and is returning from `_drain_runner_usage_flush_requests()`.
2. Before the worker releases `_usage_flush_signal_lock`, the signal handler sets `_usage_flush_requested = True`.
3. The signal handler's `_start_usage_flush_worker()` call cannot acquire the still-held lock and returns.
4. The exiting worker must release the lock and start a successor worker.

The existing `test_signal_during_active_flush_runs_follow_up_flush` coverage at `crates/runner/mitm-addon/tests/test_runner_usage_flush_signal.py:623-651` sends the second signal while the first flush call is still blocked. The original worker's drain loop sees that signal before it returns, so the test does not execute the exit-boundary handoff at line 192.

## Evidence

A targeted branch-coverage run across 685 related tests passed, while `src/runner_flush_lifecycle.py:192` remained uncovered.

A deterministic local harness forced the second signal immediately before the first drain call returned. The current handoff started a second worker and drained the request. Modeling removal of the post-release restart left `_usage_flush_requested == True` with no worker running, confirming that the existing suite would miss a regression that strands the runner request.

An unconsumed request can leave the runner without a fresh `usage-pending` acknowledgement until another signal happens, causing the bounded shutdown wait to time out and weakening usage-delivery guarantees.

## Framing Decision

**Outcome: proceed.** Repository inspection confirms that this is a real, narrowly scoped coverage gap in an existing concurrency contract. The current active-flush test does not exercise the lock-release boundary, while the production handoff already exists and is the behavior that must be protected.

This is one independently reviewable test-only delivery slice. It does not require a production-code change or decomposition into multiple PRs. Detailed synchronization and assertion design remain planning concerns.

## Plan Decision

**Outcome: proceed.** The selected direction is a release-gated proxy around a real `threading.Lock` in the test. It controls the exact pre-release window while preserving the public signal-handler entry point, production drain loop, worker lock ownership, successor handoff, and pending-snapshot writer.

The plan is the best-supported direction among the examined alternatives. Patching the internal drain function would remove the real drain-loop behavior from the race under test, while a tracing hook would couple the test to CPython tracing and source line layout. The release-gated lock proxy matches the existing `_PhaseHandoffLock` test pattern and the repository requirement for event-based synchronization.

## Plan Risks and Mitigations

- The test must wait for a successor-specific completion event and assert the flush count before invoking any generic cleanup helper that can drain a stranded flag; otherwise removal of the post-release check could be masked.
- The release gate and worker join must be released from a `finally` path so failures cannot leave a worker or patched module-level lock behind.
- The lock proxy must delegate the acquire forms used by production and cleanup, and its patch scope must end before the autouse fixture resets module state.

## Scope and Constraints

- Exercise the signal ordering at the worker-exit boundary with deterministic synchronization rather than sleeps.
- Verify the successor flush, fresh pending snapshot, cleared request flag, reacquirable worker lock, and test cleanup.
- Preserve the existing active-flush follow-up test as coverage for its distinct ordering.
- Keep the submitted change limited to `crates/runner/mitm-addon/tests/test_runner_usage_flush_signal.py`; the production handoff is read-only except for temporary mutation validation.

## Acceptance criteria

- Add a deterministic test that delivers a runner usage-flush signal after the active worker has observed no pending request but before it releases `_usage_flush_signal_lock`; do not rely on sleeps for ordering.
- Assert that a successor worker runs the boundary request exactly once, publishes the matching fresh pending snapshot, clears `_usage_flush_requested`, and leaves the worker lock acquirable.
- Assert the successor-specific completion before generic cleanup can drain a stranded request, so removing the post-release restart check at `runner_flush_lifecycle.py:191-192` makes the new test fail.
- Assert cleanup leaves no worker or module state behind for later tests.

seven332:
Closes #28027

## Summary

- Add deterministic regression coverage for a runner usage-flush signal arriving after the worker has observed no pending request but before it releases the owner lock.
- Gate a real test lock with explicit events so the test exercises the production signal handler, drain loop, successor-worker handoff, and pending snapshot writer.
- Assert exactly one successor flush, two fresh pending snapshots with the matching request ID, cleared request state, lock reacquisition, and cleanup.

## Scope

This PR changes only crates/runner/mitm-addon/tests/test_runner_usage_flush_signal.py. It does not change production code, the runner protocol, persisted data, dependencies, or rollout behavior.

## Validation

- uv lock --check
- uv sync --locked
- Focused regression test: passed
- Focused regression test repeated 20 times: passed
- python -m pytest tests/test_runner_usage_flush_signal.py: 15 passed
- ruff format --check .: passed
- ruff check .: passed
- basedpyright -p .: 0 errors, 0 warnings, 0 notes
- python -m pytest tests/: 6333 passed, 59 existing warnings
- Mutation guard: the focused test failed when the post-release restart check at runner_flush_lifecycle.py:191-192 was temporarily removed; the source was restored before submission.

### DIFF
--- a/crates/runner/mitm-addon/tests/test_runner_usage_flush_signal.py
+++ b/crates/runner/mitm-addon/tests/test_runner_usage_flush_signal.py
@@ -175,6 +175,31 @@ class _PhaseHandoffLock:
         self.release()
 
 
+class _ExitBoundaryHandoffLock:
+    """Gate the real lock release to control a runner worker exit boundary."""
+
+    def __init__(self) -> None:
+        self._lock = threading.Lock()
+        self.release_started = threading.Event()
+        self.allow_release = threading.Event()
+        self.non_blocking_acquire_failed = threading.Event()
+
+    def acquire(self, blocking: bool = True, timeout: float = -1) -> bool:
+        if timeout < 0:
+            acquired = self._lock.acquire(blocking)
+        else:
+            acquired = self._lock.acquire(blocking, timeout)
+        if not blocking and not acquired:
+            self.non_blocking_acquire_failed.set()
+        return acquired
+
+    def release(self) -> None:
+        self.release_started.set()
+        if not self.allow_release.wait(timeout=1):
+            raise AssertionError("runner flush lock release was not allowed")
+        self._lock.release()
+
+
 class TestRunnerUsageFlushSignal:
     """Tests for runner-triggered usage buffer flush requests."""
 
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #28037

**Reviewed head:** `9f4954b9c7aceffe077c7bc1a0f3082565016b0b`

### Summary

The PR adds deterministic coverage for the runner usage-flush signal arriving at the worker-exit lock boundary. The test keeps the real signal handler, drain loop, successor handoff, and pending snapshot path under test while coordinating the race with a release-gated real lock.

### Findings

No findings.

### Testing

- Focused regression test passed and passed 20 consecutive repetitions.
- `tests/test_runner_usage_flush_signal.py`: 15 passed.
- Ruff format, Ruff lint, and BasedPyright passed.
- Full addon suite: 6333 passed with 59 existing warnings.
- Mutation guard failed as expected when the post-release restart condition was removed, then the source was restored.

### Verdict

LGTM

PR #22024, opened 2026-07-17
fix(platform): submit complete iphone ime input

e7h4n:
## Summary

- defer composer Send and Queue actions while browser IME composition is active
- re-read the current TipTap document after composition settles so submission cannot use a stale draft snapshot
- keep the generic DOM composition gate in @vm0/ui while TipTap lifecycle wiring stays in the platform signal layer
- cover inline feedback Send and active-run Queue with iPhone-style composition regressions

## User impact

On iPhone, tapping Send before the IME has finished syncing no longer submits only the previously committed text prefix. The action waits for the final editor update and submits the complete composed text once.

## Verification

- @vm0/ui typecheck passed
- @vm0/app production and test typechecks passed
- targeted Send, Queue, and active-composition tests passed
- targeted Oxlint, type-aware Oxlint, ESLint, Prettier, and git diff checks passed

PR #28045, opened 2026-08-18
fix(composer): flush pending ime dom changes before submit

bingjiez666:
## Summary

On iOS Safari (both the browser and the installed PWA), sending a message composed with a Chinese IME can submit only the first character or two even though the input box visibly shows the full text.

- Flush the ProseMirror DOM observer before serializing a submission, so composition text still pending in the contenteditable is applied to `editor.state.doc` first.
- Gate the behavior behind a new `composerImeSubmitFlush` feature switch (off by default, staff orgs only).

## Why

Submission reads the document, not the DOM: `readInputForSubmission$` serializes `editor.state.doc` via `workflowComposerDocToString`. ProseMirror intentionally defers reading the contenteditable while a composition is in flight, and normally catches up when the browser delivers an `input` event — prosemirror-view's `input` handler is what flushes the observer.

When the page is busy, that catch-up can lag by seconds. The composition gate added in #22024 releases a pending submit one `requestAnimationFrame` after `compositionend` regardless of whether the document caught up, so the read can still see only the committed prefix. Reported behavior matches this: the text is visible in the box, waiting ~8s makes the send correct, and waiting 2-3s does not.

Flushing at submit time removes the dependency on when that `input` event arrives.

## Notes

`domObserver` is internal to prosemirror-view, so it is reached through a documented cast instead of being imported. It is not guarded: every `EditorView` constructs it, and a future rename must fail loudly rather than silently restore the truncated submission.

This does not change the composition gate itself; the `requestAnimationFrame` fallback still exists and is unaffected when the switch is off.

## Test plan

- Added `queues the full composed text with the ime submit flush enabled` in `chat-run-queue.test.tsx`: composition ends with the composed tail in the DOM and no following `input` event, so the submission goes through the flush path rather than prosemirror-view's `input` handler.
- **This is not a regression guard, and the PR does not have one.** Verified by direct experiment: with the flush removed the test still passes, because happy-dom delivers the pending mutation as a microtask before the composition gate settles, so the document is already current at read time. The iOS deferral this PR fixes is not reproducible under happy-dom.
- Existing IME regressions in `chat-inline-feedback.test.tsx` and `chat-run-queue.test.tsx` continue to cover the ordered path.
- Local checks run and green: `pnpm format`, `pnpm turbo run lint` (13 tasks, 0 errors), `@okouai/core` and `@okouai/app` `check-types`, and `chat-run-queue.test.tsx` (33 passed).
- **Device verification is the real gate**: enable `composerImeSubmitFlush` in Lab, then reproduce the truncation on iOS Safari and the installed PWA.

### DIFF
--- a/turbo/apps/platform/src/signals/external/feature-switch.ts
+++ b/turbo/apps/platform/src/signals/external/feature-switch.ts
@@ -63,6 +63,10 @@ export const imageModelSelectionEnabled$ = computed((get): boolean => {
   return get(featureSwitch$)[FeatureSwitchKey.ImageModelSelection] ?? false;
 });
 
+export const composerImeSubmitFlushEnabled$ = computed((get): boolean => {
+  return get(featureSwitch$)[FeatureSwitchKey.ComposerImeSubmitFlush] ?? false;
+});
+
 export const codexFastModeEnabled$ = computed((get): boolean => {
   return get(featureSwitch$)[FeatureSwitchKey.CodexFastMode] ?? false;
 });
--- a/turbo/apps/platform/src/signals/zero-page/tiptap-workflow-composer.ts
+++ b/turbo/apps/platform/src/signals/zero-page/tiptap-workflow-composer.ts
@@ -16,7 +16,12 @@ import {
   type EditorState,
   type Transaction,
 } from "@tiptap/pm/state";
-import { Decoration, DecorationSet, type NodeView } from "@tiptap/pm/view";
+import {
+  Decoration,
+  DecorationSet,
+  type EditorView,
+  type NodeView,
+} from "@tiptap/pm/view";
 import { StarterKit } from "@tiptap/starter-kit";
 import { createCompositionGate, type CompositionGate } from "@okouai/ui";
 import {
@@ -27,6 +32,7 @@ import {
 import type { ZeroWorkflowSummary } from "@okouai/api-contracts/contracts/zero-workflows";
 import { agents$ } from "../agent.ts";
 import { currentChatAgentRecordId$ } from "../agent-chat.ts";
+import { composerImeSubmitFlushEnabled$ } from "../external/feature-switch.ts";
 import { onRef, resetSignal } from "../utils.ts";
 import type { DraftInputSyncTarget, DraftSignals } from "./chat-draft.ts";
 import {
@@ -263,6 +269,33 @@ function connectComposerFeedback(
   });
 }
 
+interface PendingDomChangeObserver {
+  readonly flush: () => void;
+}
+
+function hasPendingDomChangeObserver(
+  view: EditorView,
+): view is EditorView & { readonly domObserver: PendingDomChangeObserver } {
+  return "domObserver" in view;
+}
+
+/**
+ * ProseMirror defers reading the contenteditable while an IME composition is in
+ * flight. On a busy page the composed tail can still be missing from
+ * `editor.state.doc` after the composition gate settles, which submits only the
+ * already committed prefix. Flushing the observer applies those pending DOM
+ * changes before the document is serialized.
+ *
+ * `domObserver` is internal to prosemirror-view, so it is narrowed structurally
+ * rather than imported.
+ */
+function flushPendingDomChanges(editor: Editor): void {
+  if (!hasPendingDomChangeObserver(editor.view)) {
+    return;
+  }
+  editor.view.domObserver.flush();
+}
+
 function createReadInputForSubmissionCommand(
   editor: Editor,
   compositionGate: CompositionGate,
@@ -270,7 +303,11 @@ function createReadInputForSubmissionCommand(
 ) {
   return command(({ get }, signal: AbortSignal) => {
     const includeQuoteOnlyFeedback = get(quoteOnlyFeedbackEnabled$);
+    const flushBeforeRead = get(composerImeSubmitFlushEnabled$);
     return compositionGate.runWhenSettled(() => {
+      if (flushBeforeRead) {
+        flushPendingDomChanges(editor);
+      }
       return {
         prompt: workflowComposerDocToString(editor, includeQuoteOnlyFeedback),
         editorDocument: createEditorDocumentSnapshot(
--- a/turbo/apps/platform/src/signals/zero-page/tiptap-workflow-composer.ts
+++ b/turbo/apps/platform/src/signals/zero-page/tiptap-workflow-composer.ts
@@ -269,14 +269,8 @@ function connectComposerFeedback(
   });
 }
 
-interface PendingDomChangeObserver {
-  readonly flush: () => void;
-}
-
-function hasPendingDomChangeObserver(
-  view: EditorView,
-): view is EditorView & { readonly domObserver: PendingDomChangeObserver } {
-  return "domObserver" in view;
+interface EditorViewWithDomObserver extends EditorView {
+  readonly domObserver: { readonly flush: () => void };
 }
 
 /**
@@ -286,14 +280,13 @@ function hasPendingDomChangeObserver(
  * already committed prefix. Flushing the observer applies those pending DOM
  * changes before the document is serialized.
  *
- * `domObserver` is internal to prosemirror-view, so it is narrowed structurally
- * rather than imported.
+ * Every `EditorView` constructs `domObserver`, but prosemirror-view marks it
+ * internal and omits it from the published types, so it is reached through a
+ * cast. A future rename must fail loudly here rather than silently restore the
+ * truncated submission.
  */
 function flushPendingDomChanges(editor: Editor): void {
-  if (!hasPendingDomChangeObserver(editor.view)) {
-    return;
-  }
-  editor.view.domObserver.flush();
+  (editor.view as EditorViewWithDomObserver).domObserver.flush();
 }
 
 function createReadInputForSubmissionCommand(
--- a/turbo/apps/platform/src/views/zero-page/__tests__/chat-run-queue.test.tsx
+++ b/turbo/apps/platform/src/views/zero-page/__tests__/chat-run-queue.test.tsx
@@ -1460,26 +1460,26 @@ describe("chat run queue", () => {
       expect(screen.getByLabelText("Stop")).toBeInTheDocument();
     });
     const composer = await activeRunComposer();
-    await fill(composer, "排");
+    await fill(composer, "que");
 
-    fireEvent.compositionStart(composer, { data: "排" });
+    fireEvent.compositionStart(composer, { data: "que" });
     const paragraph = composer.querySelector("p");
     if (!paragraph) {
       throw new Error("Composer paragraph not found");
     }
-    paragraph.textContent = "排队完整内容";
+    paragraph.textContent = "queued full composed text";
 
     fireEvent.click(screen.getByLabelText("Send"));
     expect(queuedBodies).toHaveLength(0);
 
     // iOS Safari can leave the composed tail in the DOM without delivering the
     // input event that would normally flush it into the document.
-    fireEvent.compositionEnd(composer, { data: "排队完整内容" });
+    fireEvent.compositionEnd(composer, { data: "queued full composed text" });
 
     await waitFor(() => {
       expect(queuedBodies).toHaveLength(1);
     });
-    expect(queuedBodies[0]?.content).toBe("排队完整内容");
+    expect(queuedBodies[0]?.content).toBe("queued full composed text");
   });
 
   it("renders a server-reconciled follow-up inline", async () => {
--- a/turbo/apps/platform/src/views/zero-page/__tests__/chat-run-queue.test.tsx
+++ b/turbo/apps/platform/src/views/zero-page/__tests__/chat-run-queue.test.tsx
@@ -1442,6 +1442,46 @@ describe("chat run queue", () => {
     expect(queuedBodies[0]?.content).toBe("排队完整内容");
   });
 
+  it("queues the full composed text when no input event follows composition end", async () => {
+    const queuedBodies: QueuedMessageCapture[] = [];
+    mockActiveRunThread(THREAD_ID, {
+      onQueuedEventAppend: (body) => {
+        queuedBodies.push(body);
+      },
+    });
+
+    detachedSetupPage({
+      context,
+      path: CHAT_PATH,
+      featureSwitches: { [FeatureSwitchKey.ComposerImeSubmitFlush]: true },
+    });
+
+    await waitFor(() => {
+      expect(screen.getByLabelText("Stop")).toBeInTheDocument();
+    });
+    const composer = await activeRunComposer();
+    await fill(composer, "排");
+
+    fireEvent.compositionStart(composer, { data: "排" });
+    const paragraph = composer.querySelector("p");
+    if (!paragraph) {
+      throw new Error("Composer paragraph not found");
+    }
+    paragraph.textContent = "排队完整内容";
+
+    fireEvent.click(screen.getByLabelText("Send"));
+    expect(queuedBodies).toHaveLength(0);
+
+    // iOS Safari can leave the composed tail in the DOM without delivering the
+    // input event that would normally flush it into the document.
+    fireEvent.compositionEnd(composer, { data: "排队完整内容" });
+
+    await waitFor(() => {
+      expect(queuedBodies).toHaveLength(1);
+    });
+    expect(queuedBodies[0]?.content).toBe("排队完整内容");
+  });
+
   it("renders a server-reconciled follow-up inline", async () => {
     const user = userEvent.setup({ delay: null });
     const queuedBodies: QueuedMessageCapture[] = [];
--- a/turbo/apps/platform/src/views/zero-page/__tests__/chat-run-queue.test.tsx
+++ b/turbo/apps/platform/src/views/zero-page/__tests__/chat-run-queue.test.tsx
@@ -1442,7 +1442,7 @@ describe("chat run queue", () => {
     expect(queuedBodies[0]?.content).toBe("排队完整内容");
   });
 
-  it("queues the full composed text when no input event follows composition end", async () => {
+  it("queues the full composed text with the ime submit flush enabled", async () => {
     const queuedBodies: QueuedMessageCapture[] = [];
     mockActiveRunThread(THREAD_ID, {
       onQueuedEventAppend: (body) => {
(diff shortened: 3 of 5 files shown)


bingjiez666:
Changes Requested


PR_AUTO_REVIEW_COMMIT: b1d74b1d8f7b9d064fec823a8317a84f5abc9d7a

### Summary

Flushes the ProseMirror DOM observer before serializing a composer submission, so IME text still pending in the contenteditable reaches `editor.state.doc` before it is read. Gated behind the new non-GA `composerImeSubmitFlush` switch, with a regression test asserting the full composed text is queued when no `input` event follows `compositionend`.

The mechanism and the gating are right. One P1 blocks: the helper silently skips the flush when the internal `domObserver` is missing, which is a fallback for a state prosemirror-view does not produce, and it converts a future breakage into a silent regression of exactly the bug this PR fixes.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | `fix:` commit must carry a regression test | Pass | `chat-run-queue.test.tsx` adds "queues the full composed text when no input event follows composition end", which fails without the flush |
| `REVIEW.md` | Feature regression guard — change must be scoped/contained | Pass | `FeatureSwitchKey.ComposerImeSubmitFlush`, `enabled: false` + `STAFF_ORG_ID_HASHES`; existing IME tests still cover the switch-off path |
| `REVIEW.md` | Type safety, no `any` / no broad casts | Pass | `tiptap-workflow-composer.ts` uses a type predicate, no `any`, no suppression comments |
| `REVIEW.md` | DB/JSONB review gate | N/A | No `packages/db/src/schema/**`, contract, or `jsonb(...)` change |
| `REVIEW.md` | Event sourcing / optimistic events | N/A | No event shape, event ID, or reconciliation change |
| `docs/bad-smell.md` | §1/§2 no `any`, no lint or type suppressions | Pass | Neither appears in the diff |
| `docs/bad-smell.md` | §7 fail fast — no silent fallback that hides a problem | Finding (P1) | `tiptap-workflow-composer.ts:flushPendingDomChanges` returns silently when `domObserver` is absent |
| `docs/bad-smell.md` | §5 no dynamic imports | Pass | Static imports only |
| `docs/fallback.md` | §3 no fallback for a state the owning contract cannot produce | Finding (P1) | `hasPendingDomChangeObserver` guards a field prosemirror-view always constructs |
| `docs/fallback.md` | §9 every introduced fallback declared in a PR-summary `Fallbacks` section | Finding (P1) | Behavior is described in prose under "Notes"; no `Fallbacks` section |
| `docs/fallback.md` | §2 non-GA feature switch needs no compatibility code | Pass | No dual-read/dual-write or migration added for the gated path |
| `docs/testing/anti-patterns.md` | AP-4 never mock internal code (`vi.mock` with relative path) | Pass | No `vi.mock` added |
| `docs/testing/anti-patterns.md` | AP-10 exercise the production bootstrap, not direct render | Pass | Uses `detachedSetupPage` |
| `docs/testing/anti-patterns.md` | AP-5 no fake timers | Pass | Real timers; settle path driven by real `requestAnimationFrame` |
| `docs/testing/anti-patterns.md` | AP-1 assert behavior, not mock calls | Pass | Asserts the queued message payload content |
| `docs/testing/app-testing.md` | Feature flags via `setupPage` `featureSwitches`; `waitFor` for async assertions | Pass | Both used in the new test |
| `docs/effect.md` | Command owns the transition; async work carries an `AbortSignal` | Pass | Flag read synchronously in `createReadInputForSubmissionCommand` before the awaited settle; existing `signal` retained |
| `docs/cache.md` | RC-2 no render-time signal allocation, RC-3 no unbounded lifetime cache | Pass | `composerImeSubmitFlushEnabled$` is a package-scope `computed`; no new cache |
| `.claude/skills/ccstate/SKILL.md` | `computed` memoizes — no manual cache layer | Pass | Plain `computed` over `featureSwitch$`, matching sibling switch selectors |

### Findings

- `turbo/apps/platform/src/signals/zero-page/tiptap-workflow-composer.ts` — `flushPendingDomChanges` / `hasPendingDomChangeObserver`: the `if (!hasPendingDomChangeObserver(editor.view)) return;` branch handles a state prosemirror-view never produces (`domObserver` is constructed by every `EditorView`). Per `docs/fallback.md` §3 and `docs/bad-smell.md` §7 this is a silent fallback, and here it has a concrete cost: if a future `@tiptap/pm` upgrade renames or removes the internal field, the guard turns this fix into a no-op and the IME truncation returns with no error, no test failure, and no signal — the "loud bug becomes wrong behavior" case `docs/fallback.md` warns about. Drop the runtime branch and narrow the type instead, so an absent observer throws:

  ```ts
  interface EditorViewWithDomObserver extends EditorView {
    readonly domObserver: { readonly flush: () => void };
  }
  (editor.view as EditorViewWithDomObserver).domObserver.flush();
  ```

  A single documented cast is the justified form here, because the field is real but deliberately undeclared in the public typings.

- PR summary: `docs/fallback.md` §9 requires a `Fallbacks` section listing every fallback the diff introduces, with surface, window, removal condition, and follow-up. The observer-missing branch is only described in prose under "Notes". Removing the branch per the finding above resolves this; if it is kept instead, add the `Fallbacks` section.

### Fallbacks

- `turbo/apps/platform/src/signals/zero-page/tiptap-workflow-composer.ts:flushPendingDomChanges` — tolerates a prosemirror-view `EditorView` without an internal `domObserver` by skipping the flush. Surface: none — a third-party internal API, not a cross-version deploy boundary, so no rollout window from `docs/fallback.md` §7 applies. Removal condition: none stated. Declared in PR summary: No — P1. Verdict: **Not justified** (`docs/fallback.md` §3 and §6; `docs/bad-smell.md` §7).

### Testing

- Coverage: Adequate. The `fix:` commit adds a regression test that reproduces the reported failure mode — composed tail present in the DOM, no `input` event after `compositionend` — and asserts the full text is queued. It fails without the flush. The switch-off path stays covered by the existing IME tests in this file and in `chat-inline-feedback.test.tsx`.
- Conventions: Compliant. Integration-level through `detachedSetupPage`, feature switch injected via the documented `featureSwitches` option, no internal mocks, no fake timers, assertions on the queued payload rather than on mock calls.
- Note: the new test was not executed locally before pushing (authoring sandbox lacked memory to run vitest), so CI is its first execution. The mechanism it relies on is the same one that makes the existing `fireEvent.input` IME test pass — prosemirror-view's `input` handler flushes the DOM observer — so it is expected to hold under happy-dom, but CI is the confirmation.

bingjiez666:
Superseded — see https://github.com/vm0-ai/vm0/pull/28045#issuecomment-5336678641

> **Superseded.** The single blocker below (a `fix:` commit without a regression guard) was escalated to the maintainer because no commit on this branch could resolve it. Bingjie Zang waived `REVIEW.md`'s regression-test requirement for this PR (option A), and the review was re-issued as LGTM against the same commit. The original text is kept below unchanged for the record; its verdict marker is retired so the automation reads the current verdict.


PR_AUTO_REVIEW_COMMIT: 05b388dd58c362d3f68ec69e90d5973d7cc1f090
PR_AUTO_REVIEW_VERDICT: Superseded

### Summary

Flushes the ProseMirror DOM observer before serializing a composer submission, so IME text still pending in the contenteditable reaches `editor.state.doc` before it is read. Gated behind the non-GA `composerImeSubmitFlush` switch.

The previous P1 is resolved: the silent `domObserver`-missing guard is gone, the internal field is reached through a documented cast, and a future rename now fails loudly instead of silently restoring the truncation.

One issue remains, and it is not resolvable by editing this branch: the `fix:` commit has no regression guard. That is a `Changes Requested` under `REVIEW.md`'s verdict rules, but the fix is a browser-timing guarantee that the test environment cannot express, so it needs a human decision rather than another commit.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | `fix:` commit must carry a regression test | Finding (P0) | `chat-run-queue.test.tsx` test passes with the flush removed; no test in the diff fails without the fix |
| `REVIEW.md` | Feature regression guard — change must be scoped/contained | Pass | `ComposerImeSubmitFlush`, `enabled: false` + `STAFF_ORG_ID_HASHES`; switch-off path unchanged and still covered by the existing IME tests |
| `REVIEW.md` | Type safety — no `any`, casts only where necessary | Pass | One documented cast to reach a deliberately undeclared prosemirror-view field; no `any`, no suppressions |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, contract, or `jsonb(...)` change |
| `REVIEW.md` | Event sourcing / optimistic events | N/A | No event shape, event ID, or reconciliation change |
| `docs/bad-smell.md` | §1/§2 no `any`, no lint or type suppressions | Pass | Neither appears in the diff |
| `docs/bad-smell.md` | §7 fail fast — no silent fallback | Pass | Resolved in `05b388dd5`; the observer is now dereferenced directly |
| `docs/bad-smell.md` | §5 no dynamic imports | Pass | Static imports only |
| `docs/fallback.md` | §3 no fallback for a state the contract cannot produce | Pass | The guard was removed in `05b388dd5` |
| `docs/fallback.md` | §9 introduced fallbacks declared in the PR summary | N/A | The diff no longer introduces a fallback, so no `Fallbacks` section is required |
| `docs/fallback.md` | §2 non-GA feature switch needs no compatibility code | Pass | No dual-read/dual-write or migration added for the gated path |
| `docs/testing/anti-patterns.md` | AP-4 never mock internal code | Pass | No `vi.mock` added |
| `docs/testing/anti-patterns.md` | AP-10 exercise the production bootstrap | Pass | Uses `detachedSetupPage` |
| `docs/testing/anti-patterns.md` | AP-5 no fake timers | Pass | Real timers and real `requestAnimationFrame` |
| `docs/testing/anti-patterns.md` | AP-1 assert behavior, not mock calls | Pass | Asserts the queued message payload |
| `docs/testing/anti-patterns.md` | AP-8 do not add tests that cannot fail for a real reason | Finding (P0) | The added test cannot distinguish the fixed from the unfixed code path |
| `docs/testing/app-testing.md` | Feature flags via `setupPage` `featureSwitches`; `waitFor` for async | Pass | Both used |
| `docs/effect.md` | Command owns the transition; async work carries an `AbortSignal` | Pass | Switch read synchronously before the awaited settle; existing `signal` retained |
| `docs/cache.md` | RC-2 no render-time signal allocation, RC-3 no unbounded cache | Pass | Package-scope `computed`; no new cache |
| `.claude/skills/ccstate/SKILL.md` | `computed` memoizes — no manual cache layer | Pass | Plain `computed` over `featureSwitch$`, matching sibling selectors |

### Findings

- `turbo/apps/platform/src/views/zero-page/__tests__/chat-run-queue.test.tsx` — the PR has no regression guard, so it does not satisfy `REVIEW.md`'s rule that a `fix:` commit carries one. Verified by direct experiment rather than inspection: removing the `flushPendingDomChanges(editor)` call and re-running the new test leaves it passing (`1 passed | 32 skipped`). Under happy-dom the pending mutation is delivered by the MutationObserver microtask before the composition gate's `requestAnimationFrame` settles, so `editor.state.doc` is already current at read time and the flush changes nothing observable.

  The fix is a *timing* guarantee — it forces the DOM read to happen now instead of when prosemirror-view's deferred composition handling gets to it. A fake DOM with no deferral cannot express that difference, so this gap is not closable by rewriting the test:
  - asserting that the flush was called would test implementation, not behavior (`docs/testing/anti-patterns.md` AP-7);
  - a unit test around the composition gate is excluded by the project's integration-tests-only rule;
  - a real-browser E2E would have to starve the main thread on demand to reproduce it, which would be inherently flaky.

  Because there is no correct commit that resolves this, it is escalated as a decision rather than a fix. See the note below.

### Testing

- Coverage: **Insufficient by `REVIEW.md`'s rule** — no test in the diff fails without the fix. The added test does provide real but narrower value: it exercises the flush path end-to-end and proves the flush does not break a normal submission when the switch is on. It is named and commented accordingly and no longer claims to be a regression guard.
- Conventions: Compliant. Integration-level through `detachedSetupPage`, switch injected via the documented `featureSwitches` option, no internal mocks, no fake timers, assertions on the queued payload.
- Local verification on `05b388dd5`, all green: `pnpm format`; `NODE_OPTIONS=--max-old-space-size=3072 pnpm turbo run lint --concurrency=1 --log-order grouped` (13 tasks, 0 errors); `@okouai/core` and `@okouai/app` `check-types` (production + tests); `chat-run-queue.test.tsx` (33 passed).

### Decision needed

The blast radius of merging as-is is bounded: `composerImeSubmitFlush` is `enabled: false` with staff-org hashes only, so per `docs/fallback.md` §2 the gated path has no external users until someone opts in, and the switch-off path is byte-for-byte unchanged. The real verification for this fix is reproducing the truncation on an iOS device with the switch on.

Two options for the maintainer:

1. **Waive the regression-test rule for this PR**, on the grounds that the fixed behavior is a browser-timing property outside what the integration suite can observe, and treat device verification behind the staff-only switch as the gate. Keep the current test as flush-path coverage.
2. **Do not merge** until a reproduction harness exists that can express a stalled ProseMirror document read.

Flagging rather than self-approving: this branch is authored by the same account running the review, so waiving an explicit `REVIEW.md` P0 is not a call this review should make on its own.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 05b388dd58c362d3f68ec69e90d5973d7cc1f090
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Flushes the ProseMirror DOM observer before serializing a composer submission, so IME text still pending in the contenteditable reaches `editor.state.doc` before it is read. Gated behind the non-GA `composerImeSubmitFlush` switch (`enabled: false`, staff orgs only).

Supersedes https://github.com/vm0-ai/vm0/pull/28045#issuecomment-5327787064. The single blocker in that review — a `fix:` commit without a regression guard — was escalated to the maintainer because no commit on this branch could resolve it. **Bingjie Zang, the repository maintainer and PR author, has decided to waive `REVIEW.md`'s regression-test requirement for this PR (option A).** Nothing else changed; the code at `05b388dd5` is identical to what the previous review examined.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | `fix:` commit must carry a regression test | Finding (P0) — **waived by maintainer** | No test in the diff fails without the fix; waiver recorded above, rationale under "Waiver rationale" |
| `REVIEW.md` | Feature regression guard — change must be scoped/contained | Pass | `ComposerImeSubmitFlush`, `enabled: false` + `STAFF_ORG_ID_HASHES`; switch-off path unchanged and still covered by the existing IME tests |
| `REVIEW.md` | Type safety — no `any`, casts only where necessary | Pass | One documented cast to reach a deliberately undeclared prosemirror-view field |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, contract, or `jsonb(...)` change |
| `REVIEW.md` | Event sourcing / optimistic events | N/A | No event shape, event ID, or reconciliation change |
| `docs/bad-smell.md` | §1/§2 no `any`, no lint or type suppressions | Pass | Neither appears in the diff |
| `docs/bad-smell.md` | §7 fail fast — no silent fallback | Pass | The observer-missing guard was removed in `05b388dd5` |
| `docs/bad-smell.md` | §5 no dynamic imports | Pass | Static imports only |
| `docs/fallback.md` | §3 no fallback for a state the contract cannot produce | Pass | Guard removed |
| `docs/fallback.md` | §9 introduced fallbacks declared in the PR summary | N/A | The diff introduces no fallback |
| `docs/fallback.md` | §2 non-GA feature switch needs no compatibility code | Pass | No dual-read/dual-write or migration for the gated path |
| `docs/testing/anti-patterns.md` | AP-4 never mock internal code | Pass | No `vi.mock` added |
| `docs/testing/anti-patterns.md` | AP-10 exercise the production bootstrap | Pass | Uses `detachedSetupPage` |
| `docs/testing/anti-patterns.md` | AP-5 no fake timers | Pass | Real timers and real `requestAnimationFrame` |
| `docs/testing/anti-patterns.md` | AP-1 assert behavior, not mock calls | Pass | Asserts the queued message payload |
| `docs/testing/anti-patterns.md` | AP-8 no test that cannot fail for a real reason | Finding (P0) — **waived by maintainer** | Same finding as the row above; the test is retained as flush-path coverage and is named and commented to say so |
| `docs/testing/app-testing.md` | Feature flags via `setupPage` `featureSwitches`; `waitFor` for async | Pass | Both used |
| `docs/effect.md` | Command owns the transition; async work carries an `AbortSignal` | Pass | Switch read synchronously before the awaited settle |
| `docs/cache.md` | RC-2 no render-time signal allocation, RC-3 no unbounded cache | Pass | Package-scope `computed`; no new cache |
| `.claude/skills/ccstate/SKILL.md` | `computed` memoizes — no manual cache layer | Pass | Plain `computed` over `featureSwitch$` |

### Findings

#### Critical (P0) - waived by maintainer decision

- `turbo/apps/platform/src/views/zero-page/__tests__/chat-run-queue.test.tsx` — no test in the diff fails without the fix. Verified by experiment: removing `flushPendingDomChanges(editor)` and re-running the added test leaves it passing (`1 passed | 32 skipped`), because happy-dom delivers the pending mutation as a MutationObserver microtask before the composition gate's `requestAnimationFrame` settles, so the document is already current at read time.

  **Waiver rationale, accepted by the maintainer:**
  - The fixed behavior is a *timing* guarantee — the flush forces the DOM read now instead of when prosemirror-view's deferred composition handling reaches it. A fake DOM with no such deferral cannot express the difference, so the gap is not closable by rewriting the test. Asserting the flush call would test implementation (AP-7); a gate-level unit test is excluded by the integration-tests-only rule; a real-browser E2E would need to starve the main thread on demand and would be inherently flaky.
  - Blast radius is bounded to zero for existing users: `composerImeSubmitFlush` is `enabled: false` with staff-org hashes only, so per `docs/fallback.md` §2 the gated path has no external users, and the switch-off path is byte-for-byte unchanged.
  - Verification moves to the device: reproduce the truncation on iOS Safari and the installed PWA with the switch enabled in Lab.

  Follow-up expectation: the switch must not go GA until that device verification is done.

### Testing

- Coverage: Insufficient by `REVIEW.md`'s letter, waived as recorded above. The added test still provides real value — it exercises the flush path end-to-end and proves the flush does not break a normal submission when the switch is on. It is named and commented so it does not claim to be a regression guard.
- Conventions: Compliant. Integration-level through `detachedSetupPage`, switch injected via the documented `featureSwitches` option, no internal mocks, no fake timers, assertions on the queued payload.
- Local verification on `05b388dd5`, all green: `pnpm format`; `NODE_OPTIONS=--max-old-space-size=3072 pnpm turbo run lint --concurrency=1 --log-order grouped` (13 tasks, 0 errors); `@okouai/core` and `@okouai/app` `check-types` (production + tests); `chat-run-queue.test.tsx` (33 passed).

Issue #27796, opened 2026-08-17
refactor(db): remove runtime dependencies on legacy integration columns


lancy:
Parent epic: #27599
Prerequisite for: #27602

## Background

The current Contract plan cannot safely remove `vm0_user_id` in the same production release that removes the deployed API's old-column statement shapes.

Repository deployment rules are explicit:

- `docs/deployment-compatibility.md:81-86` says production migrations run before the new API is promoted, while the currently deployed backend can still serve traffic.
- `docs/deployment-compatibility.md:328-338` requires a physical-column drop to use two releases: first remove the ORM declaration and every reader/writer, wait for that API to drain, and only then drop the physical column in a later release.

Current `main` `4d733eec0c11b52cba9d3c513007305909abc83e` still has:

- 12 Drizzle `legacyUserId: text("vm0_user_id")` declarations.
- 23 live API `legacyUserId:` occurrences, including explicit insert/upsert values in Slack, Teams, Telegram, Feishu, GitHub, and Agentphone paths.
- Concrete explicit writes such as `zero-slack-connect.service.ts:65`, `zero-teams-dispatch.service.ts:784-795`, and `slack-oauth.ts:377-386`. A migration that drops the column before this API drains would make these old statements fail with PostgreSQL `42703`.

The production rollback dashboard currently advertises seven targets. Direct source inspection of every target shows the same 23 live API legacy-property occurrences and 12 legacy Drizzle column declarations. Therefore those targets are Switch-capable but are not yet safe after physical contraction. This supersedes the earlier interpretation that being a descendant of the Switch baseline alone closed Contract hard gate 3.

## Design

Create the code-only first release of the two-release Contract pattern:

1. Remove `legacyUserId` from all 12 live Drizzle schema declarations.
2. Remove every runtime database statement shape that names or infers `vm0_user_id`, including explicit values, full-table projections, `RETURNING` shapes, inferred row types, and test-only API routes that execute database writes.
3. Continue writing canonical `user_id` only. The existing 0930 synchronization triggers must remain physically installed so the still-present legacy column stays populated while older API instances drain.
4. Add focused regression coverage proving current ORM/runtime statement shapes do not name `vm0_user_id`.
5. Ship this as an API release with no database migration. Prove the immediately preceding API version has drained.
6. Keep #27602 blocked until every target still advertised by the rollback dashboard is independently proven compatible with the contracted schema.

[...]

PR #27764, opened 2026-08-17
feat(billing): support custom main subscriptions

hulh122:
## Summary

- recognize the configured Custom Stripe price as an organization billing plan
- store Custom as the main subscription and plan entitlement without granting built-in Pro or Team credits
- replace an existing Pro or Team subscription only after the Custom invoice is paid
- allow Custom, usage-pack, and concurrency items to share one Stripe subscription while reconciling each entitlement independently
- attach later usage-pack and concurrency purchases to the Custom main subscription, including cancel, restore, and scheduled-change flows
- ignore stale allowance invoices and subscription updates so obsolete Stripe events cannot overwrite the current binding
- preserve usage-pack invitation requirements and canceled allowance state during webhook and cron reconciliation
- expose `PRICE_CUSTOM` through the web API environment action

## Tests

- `pnpm format`
- `pnpm turbo run lint`
- `pnpm check-types`
- `pnpm knip`
- 400 API integration tests across billing checkout, downgrade, restore, Stripe webhooks, entitlement reconciliation, and run lifecycle
- 49 platform billing UI integration tests
- `bash .github/scripts/tests/web-api-env-action-test.sh`

Companion Atom PR: https://github.com/vm0-ai/vm0-atom/pull/10

Issue #27660, opened 2026-08-17
chore: remove public-brand rollout compatibility fallbacks


hulh122:
## Context

Persisted public-brand delivery introduced temporary defaults for deployment skew and legacy states. The original database schema-readiness gate has already left the current codebase; this issue now tracks the remaining omitted-brand compatibility paths.

## Status — 2026-08-24

Audited against `main` SHA `e38158d0bc1d56b3c7379eecb2ca9962e3f273b8` and the current production rollback dashboard.

### In progress in #28804

- [ ] Require a valid `publicBrand` from the shared-thread metadata response and remove the Pages Worker VM0 default.
- [ ] Require a valid Slack OAuth-state `publicBrand`; reject missing or malformed state using the trusted request-host brand for error destinations.
- [ ] Require a valid Teams OAuth-state `publicBrand`; reject missing or malformed state using the trusted request-host brand for error destinations.
- [ ] Serialize explicit `publicBrand` for both VM0 and Okou from the Feishu OAuth-state writer.

### Remaining follow-up after #28804

- [ ] Remove the Feishu OAuth-state reader default from omitted `publicBrand` to VM0.

## Exact Feishu removal gate

Remove `publicBrandSchema.default("vm0")` only after all three conditions hold:

1. the explicit Feishu writer in #28804 is deployed to production;
2. the 10-minute Feishu OAuth state max-age has fully drained after that deployment;
3. every supported production rollback target is confirmed to emit `publicBrand` for both VM0 and Okou.

This ordering prevents an old writer from minting an in-flight VM0 state that a new reader rejects during independent deployment or rollback skew.

## Audit evidence

- The rollback dashboard in #6792 was last updated at `2026-08-24T03:16:19Z`; its oldest retained target is `7dbbea52d1d6914bd228756616e333b82a6bf0f8` (`api-v1.478.0`, published August 23).
- That rollback target already requires `publicBrand` in the shared-thread response/meta contract and emits it from Slack and Teams state writers.
- The public-brand writer commit `fa77b3e80fa` is contained by API production deployment `5941244936`, which succeeded at `2026-08-17T08:06:50Z`, more than six days before this audit and beyond the documented two-day browser/OAuth-state window.
- The same rollback target still deliberately omits `publicBrand` from Feishu VM0 state, so the Feishu reader fallback remains required for this release.

## Already complete

- [x] The API public-brand database schema-readiness gate described in the original issue is no longer present in current source.
[...]

Issue #27602, opened 2026-08-16
refactor(db): contract legacy vm0_user_id integration columns


lancy:
Parent epic: #27599

## Background

The Contract phase removes the legacy database and wire identities only after the canonical `userId` / `user_id` implementation is deployed and observed healthy. It is intentionally destructive and must be its own PR.

## Hard gate

This issue is blocked by both earlier child issues of #27599. Do not implement, open, or merge the Contract PR until all of the following are proven:

1. The Switch PR is merged and included in a production API release.
2. The API version immediately preceding Switch is no longer serving traffic.
3. No supported rollback target requires `vm0_user_id`.
4. Every affected row has equal, non-null `user_id` and `vm0_user_id` values.
5. No old writer has produced a compatibility-only row since the Switch release.
6. The supported frontend/OAuth callback window has closed and legacy `vm0UserId` state/query readers show no remaining required traffic.

Elapsed time alone is not evidence. If any gate cannot be proven, report the missing evidence and stop without opening a PR.

## Design

Once the gate is satisfied:

1. Re-run the idempotent transition backfill for any missing canonical value and fail the migration if any conflicting pair exists.
2. Remove explicit dual writes and the temporary Drizzle `legacyUserId` declarations.
3. Move the five preference-table primary keys from `(vm0_user_id, org_id)` to `(user_id, org_id)`, reusing validated canonical unique indexes where safe.
4. Drop legacy indexes built on `vm0_user_id`; retain or rename canonical indexes to neutral `user` terminology.
5. Drop `vm0_user_id` from all 12 tables.
6. Drop all 12 temporary synchronization triggers and their temporary function only after the legacy columns are no longer needed.
7. Remove legacy `vm0UserId` OAuth/query/state readers and compatibility tests after the wire drain gate is proven.
8. Remove the transition validator only after its surviving invariants have been promoted to permanent migration-consistency coverage, following `turbo/packages/db/MIGRATIONS.md:13-25`.
9. Run a repository-wide residual scan and classify any remaining `vm0UserId` / `vm0_user_id` occurrence. Historical shipped migrations remain approved evidence and must not be edited.

Use the next migration number on then-current `main`. Inspect PostgreSQL catalog dependencies before contraction; source grep alone is not sufficient evidence that no function, trigger, view, default, or prepared consumer retains the old column.

## Explicit non-goals

- Do not edit or rewrite historical migrations/snapshots.
- Do not remove compatibility before every event-based gate is proven.
[...]

PR #28072, opened 2026-08-18

e7h4n:
Daily deployment-compatibility cleanup. One cohort survived evidence collection; everything else is inside its window, gated on a product decision, or contended, and is listed at the bottom rather than split into a second PR.

## Cohort — remove the public-brand schema-readiness gate (#27660)

**Old → new boundary.** [PR #27200](https://github.com/vm0-ai/vm0/pull/27200) added a `public_brand` column to thirteen tables in migrations `0934_abnormal_mindworm` and `0935_tiresome_sumo`. A new API instance can reach production before a migration is visible, so `publicBrandSchemaReady()` probed `pg_catalog.pg_attribute` once per process and a global `app.use("*")` middleware refused **every** request with a retryable `503` + `Retry-After: 1` rather than let the new column reads fail with PostgreSQL `42703`.

**Window.** DB/API. Nominal promotion gap ~4 seconds; documented observed maximum exposure ~102 minutes. #27660's condition is that migrations 0934 and 0935 are guaranteed across all API rollback targets.

**Rollout evidence.** #27200 merged 2026-08-17T07:03:27Z (`fa77b3e80fae`); the next `api/production` promotion was `5221f393c5fc` at **2026-08-17T08:06:15Z**, and roughly **twenty further promotions** have landed since, through `2026-08-18T20:29Z`. At this run that is **~37 hours** against a ~102-minute window, so no rollback target within the practical horizon predates the migrations.

**Axiom evidence — the gate is directly observable** (`vm0-request-log-prod`, explicit bounds `2026-08-17T08:06:15Z → 2026-08-18T21:30:00Z`, `_time` predicate applied before projection):

| 12h bucket | requests | 503 |
|---|---|---|
| 2026-08-17T00:00Z | 760 601 | 0 |
| 2026-08-17T12:00Z | 584 875 | 14 |
| 2026-08-18T00:00Z | 1 157 924 | 0 |
| 2026-08-18T12:00Z | 515 875 | 3 |

This is a positive proof rather than an absence argument, and it is unusually strong because the gate is a whole-API kill switch: had `publicBrandSchemaReady()` returned `false` at any point, *every* request in that process would have been a `503`. Production served **3 019 275** requests with **17** total `503`s (0.00056%), all attributable to unrelated `providerUnavailable(...)` paths. The probe has therefore been returning `true` continuously since promotion, which means all thirteen `public_brand` columns exist in production.

**MaskDB — recorded as a staleness observation, not as counter-evidence.** `vm0-prod-ro` reports `public_brand` *absent* from `shared_threads`, `slack_org_installations`, `email_outbox`, `export_jobs`, and `push_subscriptions`, and does not expose `teams_org_installations` at all. That contradicts the Axiom result, and the Axiom result is the one that reflects the live database: the masked branch's schema view has lagged new DDL before (it never listed migration 0907's tables during the 2026-08-13 run either). Flagging it explicitly rather than quietly picking the convenient source — if the masked branch is expected to track production DDL, that lag is worth a separate look.

**Removed.** `public-brand-schema-readiness.service.ts` in full; the global readiness middleware and the `schemaReadiness` option on `createAppWithRoutes`; and the `holds traffic until the public-brand schema is ready` test together with the now-redundant `schemaReadiness: () => Promise.resolve(true)` overrides in the two tests that only set it to bypass the gate.

**Persisted-data reasoning.** None applies: this removes a startup-time schema probe, not a reader or writer of any row. The columns and their data are untouched.

## Checks

Run once against the combined diff, rebased onto `399a46ab82`, with a local PostgreSQL 18 cluster created for this run and set to `UTC`:

- `pnpm format`; `pnpm check-types` (all tasks, no errors); `pnpm -F api lint` (exit 0); `pnpm knip` (exit 0)
- `pnpm -F api exec vitest run app-factory.test.ts` → **55 passed** — the suite that owned the deleted test and exercises the whole middleware chain (404, error sanitisation, path-identity normalisation)
- `pnpm -F api exec vitest run org-team.bdd.test.ts` → **7 passed** — an integration suite driving real routes through that chain

## Open-PR overlap

Only **#25722**, on `app-factory-core.ts` and `app-factory.test.ts`. It is already `mergeable: CONFLICTING`, has been idle since 2026-08-13T15:03, and contains **zero** references to `schemaReadiness` or `publicBrandSchemaReady` — its test-file hunks sit at unrelated blocks. It requires a rebase regardless of this PR.

Deferred this run, with the exact blocker:

- **#27660's other half** — the shared-thread contract defaults (`shared-threads.ts` ×2) and the OAuth-state `vm0` defaults (`slack-oauth.ts`, `teams-oauth.ts`) are the *web/app* surface with a declared ~2-day window. #27200 promoted to `api/production` at 2026-08-17T08:06:15Z, so that window closes around 2026-08-19T08:00Z — roughly **11 hours** after this run. Deliberately left for the next run rather than shipped early.
- **#26152** concurrency billing fallbacks — #27128 finally merged on 2026-08-18T04:48:52Z, but **#27764** now edits inside both `startConcurrencyPurchase$`'s `existingSubscriptionId` branch and `reduceConcurrencySubscription$`. This is the third different billing PR to hold the same symbols (#26393 → #27128 → #27764) across seven runs; the area is under continuous churn by one author, so this cleanup is unlikely to find a quiet window on its own and probably needs to be folded into a billing PR or explicitly sequenced.
- **#27356** zero-agent default-avatar bridge — window long closed, but `apps/api/src/signals/routes/agents.ts:322` still reads `avatarUrl: args.body.avatarUrl ?? null` while only the create path at line 399 uses `randomPresetAvatar()`. The trigger is compensating a live writer. Fourth consecutive run reporting it; the unblocking change is one line.
- **#26842** saved-billing credit purchase — needs the CLI hosted-checkout contract split before the shared request can be narrowed.
- **#27602 / #27796** integration identity contraction, **#26938** agent/compose consolidation (Stage 8 owns a large transition-only surface) — multi-stage programs with their own removal owners.
- **#26672 / #26519** Website template archive pins (`LatestWebsiteTemplates` still `enabled: false`), `PersonalModelProviderAccounts` fallbacks (still `enabled: false`), **#26487** route-entry Phase A, **#25620** avatar flat-field read fallback (needs a jsonb backfill), and the `browser_*` identity contraction (MaskDB exposes no `browser_*` table, so `chat_thread_id` uniqueness across retained rows cannot be proven).


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 2a6f26724fc81b6cab659a92d95387a6430a8872
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Deletion-only cleanup (3 files, -105) that removes the public-brand schema-readiness gate from #27200: the `publicBrandSchemaReady()` probe service, the global `app.use("*")` middleware that returned a retryable `503` on a false probe, the `schemaReadiness` option on `createAppWithRoutes`, the test that owned the held-traffic behavior, and the two `schemaReadiness: () => Promise.resolve(true)` overrides that existed only to bypass the gate. Under `docs/fallback.md` §5 this is a retired rollout fallback whose window has closed, and the diff removes the branch, its option on the owning contract, and its own tests together in one pass — exactly the shape that document asks for.

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Informational

- `turbo/apps/api/src/signals/services/public-brand-schema-readiness.service.ts` (deleted): the PR body states the removal condition as "migrations 0934 and 0935 are guaranteed across all API rollback targets", but the probe's required-table list grew to 13 entries after #27200. [#27971](https://github.com/vm0-ai/vm0/pull/27971) (merged 2026-08-18T09:29:02Z, `201a719ac763`) added migration `0941_silly_argent` — `public_brand` on `usage_pack_invitation_purchases` — and extended the probe list in the same commit, so the actual condition spans **0934, 0935, and 0941**. This does not change the verdict, because 0941 clears the window on its own terms: `api/production` promoted `706ca1328643` at 2026-08-18T11:40:59Z with three further promotions since (`a0f3faaa9fb9` 20:23:42Z, `dcdd6017e417` 20:43:52Z, plus this PR's base `399a46ab82`), roughly a day against a ~102-minute observed maximum DB/API exposure. The Axiom `2026-08-18T12:00Z` bucket (515,875 requests, 3 unrelated `503`s) lies entirely after that promotion, which is direct evidence the 13-entry probe was returning `true`. Worth correcting the stated condition in the body so the audit trail matches the code that was deleted; no code change needed.

### Verification performed

- **Evidence for the removal (`docs/fallback.md`, removal-PR clause).** Established, and unusually strongly. The gate was a whole-API kill switch, so its outcome is directly observable rather than argued from absence: a single `false` probe would have made *every* request in that process a `503`. Production served 3,019,275 requests with 17 total `503`s (0.00056%) across the bounded `2026-08-17T08:06:15Z → 2026-08-18T21:30:00Z` window, all on unrelated `providerUnavailable(...)` paths. Confirmed independently that `api/production` has promoted 20+ times since `5221f393c5fc` at 2026-08-17T08:06:15Z.
- **MaskDB contradiction.** The `vm0-prod-ro` reading showing `public_brand` absent from `shared_threads`, `slack_org_installations`, `email_outbox`, `export_jobs`, and `push_subscriptions` is correctly recorded in the body as a masked-branch staleness observation, not counter-evidence. It cannot be read as "the columns are missing": missing columns imply a fully-`503` API, which 3M served requests disprove. Declaring the conflicting source rather than quietly dropping it is the right call, and the separate follow-up on the masked branch's DDL lag is warranted.
- **No orphaned references.** `grep` over the branch returns zero hits for `publicBrandSchemaReady`, `schemaReadiness`, `SchemaReadiness`, and `public-brand-schema-readiness` across `*.ts`/`*.tsx`/`*.json`/`*.md`.
- **No orphaned helpers.** `pgBooleanDecoder` (`usage-pack-credit-refund`, `usage-pack-allocation-change`, `usage-pack-invitation-purchase`, `agent-compose-provenance-lifecycle`, `org-plan-entitlement-read`, and the `require-sql-result-mapping` lint rule) and `singleton` both retain many other consumers. Consistent with `pnpm knip` exit 0.
- **Middleware ordering.** The removed handler sat between `corsMiddleware` and the Axiom flush middleware; its removal does not reorder any surviving middleware. The remaining `createAppWithRoutes` call sites (`app-factory.ts` and the route test suites) never passed `schemaReadiness`, so they relied on the default and are unaffected apart from no longer issuing the startup probe.
- **Schema and data untouched.** Migrations `0934`/`0935`/`0941` and all 13 `public_brand` columns are left in place; this removes a startup-time `pg_attribute` probe, not a reader or writer of any row. The persisted-data reasoning in the body is correct — none applies.

### Testing

- **Coverage: Adequate.** `refactor:` commit, deletion-only. The deleted `holds traffic until the public-brand schema is ready` test is removed together with the behavior it covered, which is what `docs/fallback.md` §1 requires — keeping it, or replacing it with an assertion that traffic is no longer held, would have been a tombstone and a P1. No new production code is introduced that needs coverage. `app-factory.test.ts` (55 passed) still exercises the full middleware chain — 404 handling, error sanitisation, path-identity normalisation — and `org-team.bdd.test.ts` (7 passed) drives real routes through that chain, so the surviving suites cover the factory after the option is gone.
- **Conventions: Compliant.** No new mocks, timers, `fetch` stubs, or DB mocks; the two edited call sites simply drop an option and keep their existing integration-test shape. No negative tests added.
- **Fallbacks:** not applicable — the diff introduces none. It removes one, which per `docs/fallback.md` needs no `Fallbacks` section.

### Scope

Correctly bounded. #27660's other half — the shared-thread contract defaults and the slack/teams OAuth-state `vm0` defaults — is the web/app surface with a ~2-day window closing around 2026-08-19T08:00Z, and deferring it to the next run rather than shipping it early is the right sizing decision under `docs/fallback.md` §7. It should not be pulled into this PR. Overlap with #25722 is real but incidental: that PR is already `CONFLICTING`, idle since 2026-08-13, and contains no reference to either removed symbol, so it needs a rebase regardless of this change.

PR #26805, opened 2026-08-13
feat(api): show source run on chat-run-finished automations

e7h4n:
## Summary

- Annotate new chat-run-finished workflow messages with the watched run source agent and thread title.
- Keep workflow automation metadata in canonical event context while replacing the visible workflow annotation with exactly one agent-run source annotation.
- Reuse the existing source-run avatar, title, and navigation renderer; no frontend change or historical data migration is needed.

## Behavior

Only newly fired chat-run-finished automations use the source annotation. Other automation types retain their existing workflow annotation.

## Testing

- pnpm format
- pnpm turbo run lint --concurrency=1 --log-order grouped
- pnpm knip
- pnpm turbo run check-types --concurrency=1 --filter=!api --filter=!@vm0/app
- pnpm --filter @vm0/app run check-types
- pnpm --filter api run check-types
- pnpm --filter api exec vitest run src/signals/routes/__tests__/chat-run-finished-automations.bdd.test.ts

PR #28077, opened 2026-08-18
fix(app): classify queued automation events by event type

### LINKED ISSUE #28076: bug: chat-run-finished automation events render as blank rows in the composer queue strip
## Symptom

In a chat thread with pending `chat-run-finished` automation events, the composer's pending-items strip renders them as **blank rows**: no text, the queued-message glyph instead of the automation `Bolt` icon, counted and labelled as "N messages waiting" instead of "N events waiting". Their `×` button does nothing, so the user cannot dismiss them either.

Observed in thread `ad0feeb8-fa0e-4782-bb81-6b8101b4cbf5`, two pending events:

| chat event | queued at (UTC) | source thread title in payload |
|---|---|---|
| `ea5ed943-d1f5-476b-b0c7-8d3baef62156` (seq 11992) | 2026-08-18 12:43:34 | `EPIC 26938 Stage 4C production acceptance` |
| `645123c4-5dd3-40ec-9612-1e79b2dc6420` (seq 12280) | 2026-08-18 14:11:10 | `✅ #28058 历史连接器环境来源分类` |

Both payloads are intact — `parts: [{type:"text", text:"A run in the watched chat thread completed."}, {type:"source", kind:"agent", …}]`. Nothing is lost in the data; the rendering path drops it.

Run launch is **not** affected: the queued run's prompt comes from the `chat_events` columns (`workflowName`, `triggerBrief`), not from this document.

## Root cause

1. `turbo/apps/api/src/signals/services/workflow-chat-event-queue.service.ts:146-159` builds the queue document with a `type: "automation"` non-content part (`workflowName`, `automationBrief`), then — when `agentRunSource` is present — wraps it in `withAgentRunSourceAnnotation`.
2. `turbo/apps/api/src/signals/services/chat-user-message.service.ts:102-129` filters out **every** non-content part (`source` / `automation` / `goal` / `morning_brief`) and appends only the authoritative `source`. Persisted parts become `[text, source]` — no `automation` part.
3. `turbo/apps/platform/src/signals/zero-page/composer-signals.ts:632-645` sees `eventType === "input.automation"` but finds no `automation` part, and falls into its defensive branch returning `{ kind: "message", id, text: "" }`.
4. `turbo/apps/platform/src/views/zero-page/zero-chat-composer.tsx:648-720` therefore renders a queued-message row: `ComposerQueueGlyph` instead of `Bolt`, empty body text, and the header pluralises via `chat.queue.message` instead of `chat.queue.event`.
5. Because the row is typed `kind="queued"`, its `×` calls `removeQueuedMessage$ → recallMessage$`, which early-returns for anything that is not `input.prompt` (`turbo/apps/platform/src/signals/chat-page/create-chat-thread.ts:3144`). The correct command, `skipAutomationEvent$`, is never reached.

`chat-run-finished-automation-event.service.ts` is the only caller that passes `agentRunSource` into automation admission, so this is deterministic for **every** `chat-run-finished` automation event and does not affect other automation kinds.

## Regression point

- The part-stripping semantics come from #24934 (`2f2c72af8`, 2026-08-04), where `withAgentRunSourceAnnotation` was introduced for the **client send path** (`normalSendBodyWithAgentRunSource`). Stripping client-owned `source` / `automation` / `goal` parts there is deliberate anti-forgery behaviour: a client must not be able to claim provenance it does not have.
- #26805 (`e52c30d57`, merged 2026-08-13) reused that helper on the **server-side** automation admission path, where the `automation` part is server-constructed two lines above and is not client-owned. Its PR body states the intent explicitly — "replacing the visible workflow annotation with exactly one agent-run source annotation" — and "no frontend change … is needed". That held for the message-bubble renderer but not for the composer queue strip, which is a second consumer of the same document.
- `composer-signals.ts` was byte-identical before #26805 (verified at `9ccbee23a`, 2026-08-10): the defensive branch was unreachable before that PR and is now always taken.

## Why tests did not catch it

The BDD assertion added in #26805 (`chat-run-finished-automations.bdd.test.ts`) codifies the stripped shape — it filters non-content parts and asserts `toStrictEqual([{type: "source", …}])`. It also inspects the **claimed** `input.prompt` event, whereas the blank rows only exist while the event is still a pending, unclaimed `input.automation`. No test covers pending-queue rendering.

## Proposed fix

Fix the classification in the composer strip; **do not** restore the `automation` part.

The persisted `[text, source]` shape is what #26805 intended. The message-bubble renderer already enforces "exactly one annotation" at render time — `userMessageAnnotationRenderPart()` (`zero-chat-thread-page.tsx:454`) takes the *first* non-content part via `find`, and `isWorkflowUserMessage()` (`:6333`) routes to the workflow bubble when that first part is `automation`. With the part absent, chat-run-finished messages correctly fall through to the normal bubble with the source-run chip, which is exactly the shipped design. Re-adding the part would flip the bubble, the run-group fold label (`:4234`) and the fold display text (`:6145`) back to workflow rendering unless explicit source-over-automation precedence is added at every one of those sites — larger blast radius, undoes a deliberate product decision, no user-visible gain.

The actual defect is that the composer strip infers the *event kind* from *document parts* when the event already carries a discriminant:

- `composer-signals.ts:632-660` — branch on `event.eventType === "input.automation"` and always return `kind: "automation"`. Collapse the variant to `{ kind, id, text }` with the chain `automationBrief ?? workflowName ?? messageDocumentToDisplayText(userMessage) ?? <generic i18n label>`. For these events that yields "A run in the watched chat thread completed."
- `zero-chat-composer.tsx:663-678` — the `events` mapping simplifies to `event.text`. The `Bolt` icon, the `chat.queue.event` counter wording, and `×` → `skipAutomationEvent$` all follow from the corrected `kind`; no other change needed.
- Tests — add a composer-strip case for a pending `input.automation` whose document is `[text, source]`: asserts the automation icon, non-empty label, "N events waiting" wording, and that `×` dispatches `skipAutomationEvent$`. The #26805 BDD assertion stays as-is, since it encodes the intended persisted shape.

Optional hardening (removes the whole class): the `input.automation` row projection (`chat-event-row-projection.ts:62-69`) exposes only `userMessage`, while `workflowName` and `triggerBrief` already exist as columns on the event. Projecting them would mean no client ever has to recover automation identity from the message document.

## Related, separate finding

These two events had been queued for ~10h and ~8.5h at the time of the report. `turbo/apps/api/src/signals/services/chat-thread-queue-drain.service.ts` schedules user messages → goal continuation → workflow automation, and each later drain stops once it observes a newly created active run. A continuously self-spinning goal in the same thread therefore starves queued automation events indefinitely — which is exactly the controller-thread pattern `chat-run-finished` wake-ups are designed for. Earlier in the same thread, 24 such events accumulated and were only cleared as `input.rejected` after a manual stop. Tracking separately.

lancy:
## Summary

Fixes #28076. Queued `chat-run-finished` automation events rendered as blank rows in the composer's pending-items strip.

#26805 replaces the workflow annotation of those events with the watched run's agent-run source annotation, so their queued document legitimately carries no `automation` part. The composer strip inferred the row kind from that part's presence instead of from `eventType`, and fell into a defensive branch returning `{ kind: "message", text: "" }`. That branch was unreachable before #26805 and has been taken by every `chat-run-finished` automation event since.

Three user-visible consequences, all from that one misclassification:

- empty row text;
- the queued-message glyph and "N messages waiting" wording instead of the automation icon and "N events waiting";
- `×` routed to `removeQueuedMessage$ → recallMessage$`, which early-returns for anything that is not `input.prompt`, so the rows could not be dismissed.

## Approach

Classify by `event.eventType`, which is the reliable discriminant, and derive the label with `automationBrief ?? workflowName ?? document display text`. `ComposerPendingEvent` collapses to `{ kind, id, text }` since both variants now carry the same shape, and the strip falls back to the existing `chat.queue.automationEvent` label if a document yields no text at all.

The write path is deliberately unchanged. The persisted `[text, source]` shape is what #26805 intended: `userMessageAnnotationRenderPart()` and `isWorkflowUserMessage()` pick the message bubble's single annotation by taking the *first* non-content part, so re-adding the `automation` part would flip the bubble, the run-group fold label and the fold display text back to workflow rendering unless source-over-automation precedence were added at each site.

## Behavior

Automation events with a workflow annotation keep their current label and icon. Source-annotated automation events now render as automation rows with the event text ("A run in the watched chat thread completed."), count toward the event counter, and dismiss through `skipAutomationEvent$`.

## Fallbacks

- `zero-chat-composer.tsx:PendingItemsStrip` — renders the existing `chat.queue.automationEvent` label when a queued `input.automation` yields no text. The `input.automation` row projection makes `userMessage` optional, so a document-less event is contract-reachable and would otherwise render the blank row this PR removes. Surface: none — presentational empty-state label (`docs/fallback.md` §6.4), no cross-version window. Removal condition: not applicable; it is not a rollout fallback and has no follow-up.

## Testing

- `pnpm exec prettier@3.8.1 --check` on the touched files
- Added two `workflow-queue-panel` page tests covering a pending `input.automation` whose document is `[text, source]`: row classification, label, counter wording, and skip behavior
- Full suite left to the PR pipeline per repo preference

### DIFF
--- a/turbo/apps/platform/src/signals/zero-page/composer-signals.ts
+++ b/turbo/apps/platform/src/signals/zero-page/composer-signals.ts
@@ -2,6 +2,7 @@ import type {
   ChatRunVideoOptionsRequest,
   GenerationTemplateRequest,
   PersistedAttachment,
+  UserMessageDocument,
 } from "@okouai/api-contracts/contracts/chat-threads";
 import { foldActiveChatGoalObjective } from "@okouai/api-contracts/contracts/chat-events";
 import { FeatureSwitchKey } from "@okouai/core/feature-switch-key";
@@ -117,18 +118,11 @@ export type ComposerPrimaryAction =
   | "stop"
   | "disabled";
 
-export type ComposerPendingEvent =
-  | {
-      readonly kind: "message";
-      readonly id: string;
-      readonly text: string;
-    }
-  | {
-      readonly kind: "automation";
-      readonly id: string;
-      readonly workflowName: string;
-      readonly automationBrief: string | undefined;
-    };
+export interface ComposerPendingEvent {
+  readonly kind: "message" | "automation";
+  readonly id: string;
+  readonly text: string;
+}
 
 interface ComposerWorkflowSignals extends ComposerWorkflowEditorSignals {
   readonly createWorkflowPrompt$: Command<Promise<void>, [AbortSignal]>;
@@ -593,6 +587,25 @@ export function createComposerSignals(
   };
 }
 
+/**
+ * Automation events fired by a watched chat run carry an agent-run source
+ * annotation instead of an automation part, so the document text is the label
+ * whenever no workflow annotation survived.
+ */
+function pendingAutomationEventText(
+  userMessage: UserMessageDocument | undefined,
+): string {
+  const automationPart = userMessage?.parts.find((part) => {
+    return part.type === "automation";
+  });
+  if (automationPart?.type === "automation") {
+    return (
+      automationPart.automationBrief ?? automationPart.workflowName
+    ).trim();
+  }
+  return (messageDocumentToDisplayText(userMessage) ?? "").trim();
+}
+
 function createComposerChatEventSignals(chatEvents$: Computed<ChatEvent[]>) {
   const semanticEvents$ = computed((get) => {
     return semanticChatEventsFromChatEvents(get(chatEvents$));
@@ -632,31 +645,19 @@ function createComposerChatEventSignals(chatEvents$: Computed<ChatEvent[]>) {
     (get): Promise<readonly ComposerPendingEvent[]> => {
       return Promise.resolve(
         queuedEventsFromSemanticEvents(get(semanticEvents$)).map((event) => {
-          if (event.eventType === "input.automation") {
-            const automationPart = event.userMessage?.parts.find((part) => {
-              return part.type === "automation";
-            });
-            if (!automationPart || automationPart.type !== "automation") {
-              return {
+          return event.eventType === "input.automation"
+            ? {
+                kind: "automation" as const,
+                id: event.id,
+                text: pendingAutomationEventText(event.userMessage),
+              }
+            : {
                 kind: "message" as const,
                 id: event.id,
-                text: "",
+                text: (
+                  messageDocumentToDisplayText(event.userMessage) ?? ""
+                ).trim(),
               };
-            }
-            return {
-              kind: "automation" as const,
-              id: event.id,
-              workflowName: automationPart.workflowName,
-              automationBrief: automationPart.automationBrief,
-            };
-          }
-          return {
-            kind: "message" as const,
-            id: event.id,
-            text: (
-              messageDocumentToDisplayText(event.userMessage) ?? ""
-            ).trim(),
-          };
         }),
       );
     },
--- a/turbo/apps/platform/src/views/zero-page/__tests__/workflow-queue-panel.test.tsx
+++ b/turbo/apps/platform/src/views/zero-page/__tests__/workflow-queue-panel.test.tsx
@@ -18,6 +18,11 @@ const context = testContext();
 const THREAD_ID = "d0000000-0000-4000-a000-0000000000aa";
 const EVENT_ID_1 = "f0000001-0000-4000-a000-000000000001";
 const EVENT_ID_2 = "f0000002-0000-4000-a000-000000000002";
+const SOURCE_EVENT_ID = "f0000003-0000-4000-a000-000000000003";
+const SOURCE_RUN_ID = "e0000001-0000-4000-a000-000000000001";
+const SOURCE_THREAD_ID = "e0000002-0000-4000-a000-000000000002";
+const SOURCE_AGENT_ID = "e0000003-0000-4000-a000-000000000003";
+const SOURCE_EVENT_TEXT = "A run in the watched chat thread completed.";
 
 function buttonByLabel(label: string): HTMLElement {
   const button = queryAllByRoleFast("button").find((candidate) => {
@@ -27,6 +32,70 @@ function buttonByLabel(label: string): HTMLElement {
     throw new Error(`${label} button not found`);
   }
   return button;
+}
+
+/**
+ * A chat-run-finished automation replaces its workflow annotation with the
+ * watched run's source annotation, so the queued document carries no
+ * automation part.
+ */
+function setupSourceAnnotatedAutomationPage({
+  onRecallEventAppend,
+}: {
+  onRecallEventAppend?: (body: {
+    revokesEventId: string;
+    clientEventId: string;
+  }) => void;
+} = {}): void {
+  mockChatLifecycle(context, {
+    threadId: THREAD_ID,
+    threadTitle: "Workflow queue thread",
+    chatEvents: [
+      {
+        id: "msg-workflow-running-user",
+        role: "user",
+        content: "Automation run",
+        runId: "run-workflow-1",
+        createdAt: "2026-07-10T00:59:00Z",
+      },
+      {
+        id: "msg-workflow-running-assistant",
+        role: "assistant",
+        content: null,
+        runId: "run-workflow-1",
+        createdAt: "2026-07-10T00:59:01Z",
+      },
+      {
+        id: SOURCE_EVENT_ID,
+        eventType: "input.automation",
+        content: null,
+        userMessage: {
+          version: 1,
+          parts: [
+            { type: "text", text: SOURCE_EVENT_TEXT },
+            {
+              type: "source",
+              kind: "agent",
+              runId: SOURCE_RUN_ID,
+              threadId: SOURCE_THREAD_ID,
+              agentId: SOURCE_AGENT_ID,
+              titleSnapshot: "Watched chat run",
+              href: `/chats/${SOURCE_THREAD_ID}#run-${SOURCE_RUN_ID}`,
+            },
+          ],
+        },
+        runId: undefined,
+        createdAt: "2026-07-10T01:03:00Z",
+      },
+    ],
+    activeRunIds: ["run-workflow-1"],
+    onRecallEventAppend,
+  });
+
+  detachedSetupPage({
+    context,
+    path: `/chats/${THREAD_ID}`,
+  });
 }
 
 function setupWorkflowQueuePage({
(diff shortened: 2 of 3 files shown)


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: ec041731dd2b8f9ab9a6e01a41833c2d2fdffad3
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Classifies composer pending-queue rows by `event.eventType` instead of by the presence of an `automation` part, so `chat-run-finished` automation events — whose queued document carries an agent-run `source` annotation instead of a workflow annotation since #26805 — stop rendering as blank queued-message rows. `ComposerPendingEvent` collapses to `{ kind, id, text }`, and the label is derived with `automationBrief ?? workflowName ?? document display text`.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic, including null/undefined paths | Pass | `composer-signals.ts:pendingAutomationEventText` handles the optional `userMessage`, an automation part with only `workflowName`, and a document with neither; the view guarantees a non-empty row label. |
| `REVIEW.md` | Type safety, no `any` or overly broad types | Pass | Adds the explicit `UserMessageDocument \| undefined` parameter type; narrowing on `automationPart?.type === "automation"` replaces no casts. |
| `REVIEW.md` | Security (injection, secrets, auth) | N/A | Client-side rendering of an already-persisted document; no new endpoint, credential, or user-supplied HTML path. |
| `REVIEW.md` | Style and maintainability, no duplicated logic, no oversized functions | Pass | Label derivation moves into one 12-line helper; the view's two `flatMap` blocks collapse to `filter`. Test setup duplicates the two running-run fixtures from `setupWorkflowQueuePage`, which keeps the two fixtures independent. |
| `REVIEW.md` | Interface changes are deliberate and callers updated | Pass | `ComposerPendingEvent` is exported; both consumers (`composer-signals.ts`, `zero-chat-composer.tsx`) are updated in the diff, and no other reference exists. |
| `REVIEW.md` | Feature regression guard for changed existing behavior | Pass | Automation-row labelling changed; the pre-existing cases (`customer-followup` workflow-name row, `Webhook event third` brief row, queued user-message row) are still asserted by the untouched tests in the same file. |
| `REVIEW.md` | Feature switch gating for new user-facing behavior | N/A | Restores intended behavior of a shipped surface rather than adding a feature. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, contract, or `jsonb().$type<...>()` change; the persisted document shape is untouched. |
| `docs/bad-smell.md` | Zero `any`, zero lint/type suppressions | Pass | Neither appears in the diff. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | Only static imports; the new `UserMessageDocument` import is type-only. |
| `docs/bad-smell.md` | No hardcoded URLs or environment-specific values | Pass | The only literals added are test fixture UUIDs and copy; the row label comes from the existing i18n key. |
| `docs/bad-smell.md` | Fail fast, no configuration fallback slop | Pass | No configuration or secret defaulting; see Fallbacks for the one presentational default. |
| `docs/fallback.md` | Fallbacks limited to the §6 categories | Pass | The single introduced default is an empty-state label (§6.4). The `?? workflowName` and `?? ""` chains are pre-existing behavior moved verbatim. |
| `docs/fallback.md` | Every introduced fallback declared in the PR summary | Pass | `## Fallbacks` section present; see the note under Fallbacks about when it was added. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | Both new tests assert live rendering and dismissal behavior. |
| `docs/event-sourcing.md` | Persistent events authoritative; no imperative optimistic rollback | Pass | Only the read-side projection of already-persisted queued events changes; no optimistic event lifecycle, ID reuse, or failure-path cleanup is touched. |
| `docs/effect.md` | Derived values belong in `computed`; render stays pure | Pass | The derivation stays inside `pendingEvents$`; the view only reads, filters, and renders. |
| `docs/effect.md` | Explicit cancellation and command semantics | Pass | No new command, subscription, timer, or async process; the existing `detach(..., Reason.DomCallback)` call sites are unchanged. |
| `docs/cache.md` | No render-time signal allocation or unbounded caches | Pass | No signal is created in render and no cache is introduced; `computed` memoization is left to ccstate. |
| `.claude/skills/ccstate/SKILL.md` | `detach()` only in views, never in `signals/` | Pass | The signals change adds a pure helper and a `computed`; no `detach()` is added anywhere. |
| `docs/deployment-compatibility.md` | Independently deployed boundaries stay compatible | Pass | Client-only change against an unchanged persisted shape; an old web client keeps today's rendering until it refreshes (~2 days) and needs no server-side branch. |
| `docs/testing.md` | `fix:` commits carry a regression test | Pass | Two new page tests cover the exact reported shape (`[text, source]` queued `input.automation`). |
| `docs/testing.md` | Integration-level tests, mock only at the boundary | Pass | No `vi.mock`, no `vi.stubGlobal("fetch", ...)`, no fake timers; the MSW-backed `mockChatLifecycle` fixture is the only mock. |
| `docs/testing/app-testing.md` | Platform tests are page-level and enter through `detachedSetupPage` | Pass | Both tests live in `views/zero-page/__tests__` and set up through `detachedSetupPage` without awaiting it. |
| `docs/testing/testing-external-behavior.md` | Construct and observe through the user-visible surface | Pass | The skip test clicks the rendered button and asserts the resulting recall request and the row disappearing, rather than calling signals directly. |
| `docs/testing/anti-patterns.md` | No mock-call-only assertions, direct component rendering, or implementation-detail assertions | Pass | Assertions target rendered text and ARIA labels; the recall assertion checks the request the page produced, matching the existing test in the same file. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Fallbacks

- `zero-chat-composer.tsx:PendingItemsStrip` — renders the existing `chat.queue.automationEvent` label when a queued `input.automation` yields no text. `chat-event-row-projection.ts` makes `userMessage` optional for that event type, so a document-less event is contract-reachable and would otherwise reproduce the blank row this PR removes. Surface: none — presentational empty-state label (`docs/fallback.md` §6.4), no cross-version rollout window. Removal condition: not applicable, it is not a rollout fallback. Declared in PR summary: Yes — the `## Fallbacks` section was added during this review, since the original summary described the label only in prose. Verdict: justified.

### Testing

- Coverage: Adequate. `queues a source-annotated automation event as an event row` pins classification, label text, and the "1 event waiting" counter wording; `skips a source-annotated automation event` pins that the row's remove action reaches `skipAutomationEvent$` rather than the queued-message recall path that previously no-opped. The three pre-existing rows in this file keep the workflow-annotated and queued-message paths covered.
- Conventions: Compliant. Page-level entry, MSW fixture, no internal mocks, no fake timers, assertions on user-visible surfaces.

PR #28092, opened 2026-08-19
fix(api): run queued automation events before goal continuation

### LINKED ISSUE #28078: bug: goal continuation starves queued automation events
## Symptom

A thread running a self-continuing goal never drains its queued automation events. In thread `ad0feeb8-fa0e-4782-bb81-6b8101b4cbf5`, two `chat-run-finished` automation events sat unclaimed for ~10h and ~8.5h while the goal loop started a new run roughly every 40 seconds:

| chat event | queued at (UTC) | waited |
| --- | --- | --- |
| `ea5ed943-d1f5-476b-b0c7-8d3baef62156` (seq 11992) | 2026-08-18 12:43:34 | ~10h |
| `645123c4-5dd3-40ec-9612-1e79b2dc6420` (seq 12280) | 2026-08-18 14:11:10 | ~8.5h |

This is not a one-off: earlier in the same thread 24 automation events accumulated the same way and were never executed — they were only cleared as `input.rejected` after the user manually interrupted the run.

## Root cause

`chat-event-queue.service.ts:98-103` defines the authoritative class priority:

```sql
CASE event_type
  WHEN 'input.prompt' THEN 0
  WHEN 'input.goal' THEN 1
  WHEN 'input.automation' THEN 2
END
```

Both claim paths take the head of `listPendingChatQueueEvents()` and bail out when it is not their own event type (`loadNextGoalQueueEvent`, `loadNextWorkflowQueueEvent`), so this CASE is the single source of truth. `chat-thread-queue-drain.service.ts:106-115` reinforces it by calling the drains in the order user messages → goal continuation → workflow automation, with each later drain stopping once it observes the run the earlier one just created.

A goal that continues itself leaves no idle window between runs, so class 2 is never reached. The starvation is unbounded: it ends only when the goal is paused, completed, or the user interrupts.

The combination is structural rather than accidental. `chat-run-finished` automations exist to wake a controller thread when a watched run finishes, and a controller thread is exactly the kind of thread that runs a goal loop — so the wake-up can never be delivered to the thread that most needs it. In the incident above, the Stage 4C acceptance completion never reached its controller.

The UI already tells users the opposite ordering. `zero-chat-composer.tsx:766-768` renders the active goal row last and states it "only runs once the queue drains", so the product surface and the scheduler currently disagree.

## Requested change

Reorder the queue classes to **user messages → automation → goal**:

- `chat-event-queue.service.ts:98-103` — swap the CASE weights so `input.automation` is 1 and `input.goal` is 2, and update the doc comment at `:106-112` that currently states "goal continuation stays ahead of automation input".
- `chat-thread-queue-drain.service.ts:117-165` — call `drainWorkflowQueueForThread$` before `drainGoalQueueForThread$`, and update the ordering comment at `:106-115`.
- No change is needed in `loadNextGoalQueueEvent` / `loadNextWorkflowQueueEvent`; both already defer to whatever the priority query puts at the head. `hasPendingUserChatQueueEvent` still gives user prompts absolute priority.

A pending `input.goal` is not lost by deferring it: it stays pending and is claimed on the next drain after the automation run reaches a terminal state.

### Tests to invert

Two existing tests encode the current order and must be flipped, not deleted:

- `workflow-queue.test.ts:550` — "runs a pending goal continuation before a newer automation event on the same thread"
- `workflow-queue.test.ts:580` — "lets a goal continuation preempt an automation event during final queue claim"

`workflow-queue.test.ts:512` ("keeps a user prompt ahead of a pending goal continuation") stays valid unchanged. Add coverage for the reported shape: an active self-continuing goal plus a queued automation event, asserting the automation run is created first and the goal continuation follows it.

## Trade-off to decide

Flipping the two classes moves the starvation risk to the other side: a thread receiving automation events faster than its runs complete would keep a goal waiting indefinitely. That is far less likely than the current failure (automation events are discrete and infrequent; goal continuation regenerates every turn by design), but it is worth an explicit decision.

If unbounded deferral in either direction is unacceptable, the alternative is to drop the class distinction between `input.goal` and `input.automation` entirely and order those two purely FIFO by `created_at`, keeping only `input.prompt` as a privileged class. That is fair in both directions and removes the need to pick a winner.

## Context

Found while diagnosing #28076 (fixed by #28077), which made these two stuck events visible as blank rows in the composer queue strip. That fix only corrected the rendering; the events were still stuck for the reason described here.

lancy:
## Summary

Closes #28078. Queued automation events were never claimed on a thread with an active goal.

`chatQueueEventPriority()` ranked `input.goal` (1) ahead of `input.automation` (2). A goal appends a new continuation event after every run, so such a thread never goes idle and class 2 was never reached. In the reported thread two `chat-run-finished` events waited ~10h and ~8.5h; an earlier batch of 24 events was only cleared as `input.rejected` after a manual interrupt.

The failure is structural: `chat-run-finished` automations exist to wake a controller thread, and a controller thread is exactly the kind of thread that runs a goal loop, so the wake-up could never reach the thread that needed it.

## Approach

Swap the two classes so the order is user input → automation input → goal continuation.

- `chat-event-queue.service.ts:chatQueueEventPriority` — `input.automation` becomes 1 and `input.goal` becomes 2. Every claim path (`loadNextUnclaimedQueuedUserMessage`, `loadNextWorkflowQueueEvent`, `loadNextGoalQueueEvent`, and the final locked claims in `chat-queued-event.service.ts`) already reads the head of this one ordering and bails when the head is not its own class, so this CASE is the single behavioral change.
- `chat-thread-queue-drain.service.ts:drainChatThreadQueueForThread$` — the drain calls now run workflow automation before goal continuation, matching the new class order. The command still returns the workflow drain result.
- `hasPendingUserChatQueueEvent` is untouched, so user prompts keep absolute priority.

A deferred goal event is not lost: it stays pending and is claimed on the next drain once the automation run ahead of it reaches a terminal state, which the updated tests now assert.

The composer already renders the active goal below queued automation events and tells the user it "only runs once the queue drains" (`zero-chat-composer.tsx`), so this also removes a disagreement between the product surface and the scheduler.

## Trade-off

This moves the theoretical starvation risk to the other side: a thread receiving automation events faster than its runs complete would keep a goal waiting. That is far less likely than the current failure — automation events are discrete and infrequent, while goal continuation regenerates every turn by design — and #28078 records the fully FIFO alternative if that ever becomes a real problem.

## Deployment compatibility

Runtime scheduling only: no schema, contract, queue payload, or persisted shape changes. During rollout two API versions can briefly disagree about which pending class to claim first for the same thread; both take the head under their own CASE while holding the same row locks, so the only effect is transient ordering, and no event is dropped or double-claimed.

## Testing

- `pnpm exec prettier@3.8.1` on the touched files
- `workflow-queue.test.ts` — inverted the two tests that encoded the old order and extended the first one to assert the deferred goal still runs after the automation run completes:
  - "runs a newer automation event before a pending goal continuation on the same thread"
  - "keeps an automation event ahead of a goal continuation during final queue claim"
- `workflow-queue.test.ts` — "revokes an invalid goal event once the automation ahead of it completes": the previous version relied on the goal being the higher-priority head, which no longer happens, so the invalid-goal revocation is now asserted at the point the goal queue actually takes its turn.
- "keeps a user prompt ahead of a pending goal continuation" is unchanged and still passes as written.
- Full suite left to the PR pipeline per repo preference.

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/workflow-queue.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/workflow-queue.test.ts
@@ -547,7 +547,7 @@ describe("workflow queue", () => {
     await runsApi.requestCancelRun(scenario.actor, user.body.runId, [200]);
   });
 
-  it("runs a pending goal continuation before a newer automation event on the same thread", async () => {
+  it("runs a newer automation event before a pending goal continuation on the same thread", async () => {
     const scenario = await setup();
     const automation = await createWebhookAutomation(scenario);
     const goal = await createActiveGoalQueueEventFixture({
@@ -555,29 +555,37 @@ describe("workflow queue", () => {
       orgId: scenario.orgId,
       userId: scenario.userId,
       agentId: scenario.agentId,
-      objective: "finish before the automation event",
-      objectiveBrief: "Finish before the automation event",
+      objective: "continue after the automation event",
+      objectiveBrief: "Continue after the automation event",
     });
 
-    expectAcceptedWithoutRun(
-      await postWorkflowWebhook(automation, "goal wins queue priority"),
+    const workflowRunId = await expectAcceptedRunId(
+      await postWorkflowWebhook(automation, "automation wins queue priority"),
+      automation.threadId,
     );
-    const goalQueue = await readGoalQueueStateFixture(automation.threadId);
-    expect(goalQueue.runIds).toHaveLength(1);
-    expect(goalQueue.eventIds).toContain(goal.eventId);
-    await expect(workflowRunIds(automation.threadId)).resolves.toHaveLength(0);
+    const queuedGoal = await readGoalQueueStateFixture(automation.threadId);
+    expect(queuedGoal.runIds).toHaveLength(0);
+    expect(queuedGoal.eventIds).toContain(goal.eventId);
     await expect(
       pendingAutomationEvents(automation.threadId),
-    ).resolves.toHaveLength(1);
+    ).resolves.toHaveLength(0);
 
-    const [goalRunId] = goalQueue.runIds;
+    // A goal that self-continues after every run would otherwise leave no idle
+    // window for the automation event, so the deferred goal must still run once
+    // the automation ahead of it reaches a terminal state.
+    await completeRunThroughSandbox(scenario, workflowRunId);
+    const drainedGoal = await readGoalQueueStateFixture(automation.threadId);
+    expect(drainedGoal.runIds).toHaveLength(1);
+    expect(drainedGoal.eventIds).toContain(goal.eventId);
+
+    const [goalRunId] = drainedGoal.runIds;
     if (!goalRunId) {
       throw new Error("Expected the goal continuation to create a run");
     }
     await runsApi.requestCancelRun(scenario.actor, goalRunId, [200]);
   });
 
-  it("lets a goal continuation preempt an automation event during final queue claim", async () => {
+  it("keeps an automation event ahead of a goal continuation during final queue claim", async () => {
     const scenario = await setup();
     const automation = await createWebhookAutomation(scenario);
     const admissionLock = await holdOrgAdmissionLockFixture({
@@ -600,8 +608,8 @@ describe("workflow queue", () => {
       orgId: scenario.orgId,
       userId: scenario.userId,
       agentId: scenario.agentId,
-      objective: "preempt the preparing automation event",
-      objectiveBrief: "Preempt the preparing automation event",
+      objective: "wait behind the preparing automation event",
+      objectiveBrief: "Wait behind the preparing automation event",
     });
     const goalDrain = drainChatThreadQueueFixture({
       threadId: automation.threadId,
@@ -612,24 +620,22 @@ describe("workflow queue", () => {
     admissionLock.release();
     const [workflowResult] = await Promise.all([workflowRequest, goalDrain]);
     await admissionLock.done;
-    expectAcceptedWithoutRun(workflowResult);
+    const workflowRunId = await expectAcceptedRunId(
+      workflowResult,
+      automation.threadId,
+    );
 
     const goalQueue = await readGoalQueueStateFixture(automation.threadId);
-    expect(goalQueue.runIds).toHaveLength(1);
+    expect(goalQueue.runIds).toHaveLength(0);
     expect(goalQueue.eventIds).toContain(goal.eventId);
-    await expect(workflowRunIds(automation.threadId)).resolves.toHaveLength(0);
     await expect(
       pendingAutomationEvents(automation.threadId),
-    ).resolves.toHaveLength(1);
+    ).resolves.toHaveLength(0);
 
-    const [goalRunId] = goalQueue.runIds;
-    if (!goalRunId) {
-      throw new Error("Expected the preempting goal to create a run");
-    }
-    await runsApi.requestCancelRun(scenario.actor, goalRunId, [200]);
+    await runsApi.requestCancelRun(scenario.actor, workflowRunId, [200]);
   });
 
-  it("continues to workflow automation after revoking a higher-priority invalid goal", async () => {
+  it("revokes an invalid goal event once the automation ahead of it completes", async () => {
     const scenario = await setup();
     const automation = await createWebhookAutomation(scenario);
     const goal = await createActiveGoalQueueEventFixture({
@@ -643,12 +649,14 @@ describe("workflow queue", () => {
     await pauseGoalQueueTargetFixture(goal.goalId);
 
     const workflowRunId = await expectAcceptedRunId(
-      await postWorkflowWebhook(automation, "run after invalid goal"),
+      await postWorkflowWebhook(automation, "run before invalid goal"),
       automation.threadId,
     );
     await expect(
       readWorkflowRunTriggerSourceFixture(workflowRunId),
     ).resolves.toBe("automation-event");
+
+    await completeRunThroughSandbox(scenario, workflowRunId);
     const events = await wf.readThreadEvents(automation.threadId);
     expect(events).toContainEqual(
       expect.objectContaining({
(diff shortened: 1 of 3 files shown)


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: 449b6e10b734bf29e1fd93c1bd114f05c556a540
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Swaps the pending chat-queue classes so the order becomes user input → automation input → goal continuation. A goal appends a new continuation event after every run, so a thread with an active goal never went idle and automation events ranked below it were never claimed. The change is one CASE expression plus the matching drain call order, because every claim path already reads the head of that single ordering.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic, no wrong assumptions about call paths | Pass | `chatQueueEventPriority` is the only ordering source: `loadNextUnclaimedQueuedUserMessage`, `loadNextWorkflowQueueEvent`, `loadNextGoalQueueEvent`, and the three locked claims in `chat-queued-event.service.ts` all take the head of `listPendingChatQueueEvents` and bail when it is not their own class, so no second ordering was left stale. |
| `REVIEW.md` | Race conditions and unhandled edge cases | Pass | Claims stay serialized by the per-thread advisory lock and `for("update")` row lock; the deferred goal event stays pending and `drainGoalQueueForThread$` claims it on the next drain. A paused goal still reaches `revokeGoalEvent` when the goal queue takes its turn (`goal-queue-drain.service.ts:480-491`). |
| `REVIEW.md` | Security (injection, secrets, auth) | N/A | No new endpoint, credential, or user-supplied value; the CASE operates on a fixed set of internal event-type literals. |
| `REVIEW.md` | Type safety, no `any` or broad casts | Pass | No type changes; the drain command keeps returning `WorkflowQueueDrainResult \| null` from the workflow half. |
| `REVIEW.md` | Style and maintainability | Pass | Both doc comments were updated to state the new order and the reason for it, so the invariant is not left to the reader to infer. |
| `REVIEW.md` | Feature regression guard for changed existing behavior | Pass | The changed behavior is goal-vs-automation ordering, and both directions are covered: automation-first at admission and during the final claim, plus the deferred goal still running after the automation run completes. `keeps a user prompt ahead of a pending goal continuation` is untouched and still asserts user priority. |
| `REVIEW.md` | Feature switch gating for new user-facing behavior | N/A | Bug fix to scheduling order, not a new feature. Gating it would require a dual-ordering path, which `docs/fallback.md` §2 and §3 argue against. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, contract, or `jsonb().$type<...>()` change; only an `ORDER BY` expression over an existing column. |
| `docs/bad-smell.md` | Zero `any`, zero lint/type suppressions, no dynamic imports | Pass | None appear in the diff. |
| `docs/bad-smell.md` | Fail fast, no configuration fallback or silent recovery | Pass | No defaulting or error-swallowing added; the drain still propagates through `signal.throwIfAborted()`. |
| `docs/fallback.md` | No fallback introduced for a state the contract prevents | Pass | The diff adds no `??`/`\|\|` chain, compatibility branch, or tolerance path, so no `Fallbacks` section is required. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | The three touched tests assert live claim ordering and revocation, not the absence of deleted code. |
| `docs/event-sourcing.md` | Persistent events authoritative, no imperative optimistic cleanup | Pass | Only the claim order of already-persisted pending events changes; no optimistic projection, shared event ID, or failure-path removal is touched. |
| `docs/deployment-compatibility.md` | Old backend ↔ new backend safe during rollout | Pass | Declared in the PR summary. Both versions read the head under their own CASE while holding the same per-thread advisory lock and `for update` row lock, so a mixed fleet can only differ in which class it claims first; no event is dropped or double-claimed, and no persisted shape, contract, or queue payload changed. |
| `.claude/skills/ccstate/SKILL.md` | ccstate command composition and no `detach()` in services | Pass | The drain still composes commands through `set(...)` and passes the caller's `AbortSignal`; no `detach()` or new state atom is introduced. |
| `docs/effect.md` | One semantic command owns the transition; cancellation is explicit | Pass | `drainChatThreadQueueForThread$` remains the single per-thread scheduler entry, and each half keeps its `signal.throwIfAborted()` between steps. |
| `docs/cache.md` | Render purity, state ownership, resource lifecycles | N/A | API-only diff with no React render path, signal allocation, or external resource lifetime. |
| `docs/testing.md` | `fix:` commits carry a regression test | Pass | The reported failure mode (goal active, automation queued) is now pinned by `runs a newer automation event before a pending goal continuation on the same thread`. |
| `docs/testing.md` | Integration tests, mock only at the boundary, no fake timers | Pass | The touched tests use the real DB and real routes; no `vi.mock`, no fetch stubbing, and no `vi.useFakeTimers()` was added. |
| `docs/testing/api-testing.md` | Route behavior covered through the real app and contract | Pass | Setup and verification go through the webhook route, the run-completion sandbox route, and `runsApi.requestCancelRun`, not through direct service calls. |
| `docs/testing/testing-external-behavior.md` | Construct and observe through the external interface | Pass | `completeRunThroughSandbox` drives the terminal callback the same way production does, instead of invoking the drain directly to force the next claim. |
| `docs/testing/anti-patterns.md` | No mock-call-only assertions or over-testing | Pass | Assertions are on queue state, created runs, and emitted events. The previous test count is unchanged: two tests were inverted in place and one was retargeted rather than duplicated. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- Coverage: Adequate. `runs a newer automation event before a pending goal continuation on the same thread` covers both halves of the fix — the automation event is claimed while a goal is queued, and the deferred goal still runs after the automation run reaches a terminal state, which is the property that makes the reordering safe. `keeps an automation event ahead of a goal continuation during final queue claim` keeps the admission-lock race covered with the inverted winner, and asserts through `expectAcceptedRunId`, which tolerates either the webhook request or the concurrent drain creating the run.
- `revokes an invalid goal event once the automation ahead of it completes` was retargeted rather than dropped: its old premise (an invalid goal blocking the queue head) can no longer occur, so the revocation is now asserted at the point the goal queue actually takes its turn. Verified against `goal-queue-drain.service.ts:480-491`, where a missing goal target revokes the event inside the same drain.
- Conventions: Compliant. Real database, real routes, boundary-only mocks, no fake timers, assertions on externally observable state.

PR #28107, opened 2026-08-19
fix(python): track generator advancement failures

### LINKED ISSUE #28088: bug(python): track generator iteration failures in metadata linter
## Summary

At commit `d570536126e6137eb3423573623f8abbdf6be0c6`, the flow metadata linter treats `ast.GeneratorExp` as an iterable whose acquisition and advancement cannot raise. This is incorrect for generator advancement and causes the linter to lose a reachable implicit-exception alias state.

`_iteration_may_raise()` returns `False` for `ast.GeneratorExp` in `crates/runner/mitm-addon/scripts/flow_metadata_linter/ast_helpers.py:456-473`. `_visit_for_statement()` relies on that result at `crates/runner/mitm-addon/scripts/flow_metadata_linter/visitor.py:1015-1043` to decide whether to record exception states before and after loop iterations.

## Reproduction

```python
from contextlib import suppress


def probe(flow):
    metadata = {}
    with suppress(ZeroDivisionError):
        for _item in (1 / divisor for divisor in (1, 0)):
            metadata = flow.metadata
        metadata = {}
    return metadata["vm_run_id"]
```

Runtime behavior:

1. The first generator advance yields `1`, so the loop body aliases `metadata` to `flow.metadata`.
2. The second advance raises `ZeroDivisionError`.
3. The context manager suppresses the exception, skipping `metadata = {}`.
4. The raw-key access therefore reads from `flow.metadata`.

However, `_MetadataKeyVisitor` returns no violations for this source. Replacing the generator expression with a dynamic custom iterator produces the expected `use metadata_keys.VM_RUN_ID` diagnostic, confirming that the difference comes from the `ast.GeneratorExp` classification.

## Impact

The registered flow metadata guard can accept a reachable raw registered-key access. Code using this pattern can bypass centralized `metadata_keys` constants and silently weaken key-renaming and consistency enforcement.

## Acceptance criteria

- A synchronous `for` loop records implicit exception state when advancing a generator expression, including a later advance after the loop body has established metadata aliases.
- Generator expression construction remains modeled as deferred rather than eagerly executing its body.
- Add a regression fixture where the first generated item establishes a metadata alias, a later advance raises inside a suppressing context manager, and the post-context raw-key access is reported.
- Existing deferred-generator and flow metadata linter tests continue to pass.

seven332:
## Summary

- distinguish iterator acquisition from advancement in the flow metadata linter
- retain metadata aliases when a later generator-expression advance raises after a loop body runs
- preserve deferred construction for nested generator expressions

## Scope

The private iterable fallibility helper now requires each caller to declare whether the current execution phase advances the iterable. Generator expressions are safe when only their iterator is acquired and conservatively fallible when advanced. Existing classifications for eager built-in and dynamic iterables are unchanged.

This PR does not add body-level generator precision, change runtime addon behavior, or modify APIs, persisted state, protocols, dependencies, migrations, or deployment contracts.

## Testing

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_flow_metadata_key_contract.py -q` (21 passed)
- `uv run --no-sync python -m pytest tests/` (6333 passed)

Closes #28088

### DIFF
--- a/crates/runner/mitm-addon/scripts/flow_metadata_linter/ast_helpers.py
+++ b/crates/runner/mitm-addon/scripts/flow_metadata_linter/ast_helpers.py
@@ -453,14 +453,15 @@ def _static_truth_value(node: ast.AST) -> bool | None:
     return None
 
 
-def _iteration_may_raise(node: ast.AST) -> bool:
-    """Return whether obtaining or advancing this iterable may invoke user code."""
+def _iteration_may_raise(node: ast.AST, *, advances: bool) -> bool:
+    """Return whether obtaining and optionally advancing this iterable may invoke user code."""
+    if isinstance(node, ast.GeneratorExp):
+        return advances
     if isinstance(
         node,
         (
             ast.Dict,
             ast.DictComp,
-            ast.GeneratorExp,
             ast.JoinedStr,
             ast.List,
             ast.ListComp,
--- a/crates/runner/mitm-addon/scripts/flow_metadata_linter/visitor.py
+++ b/crates/runner/mitm-addon/scripts/flow_metadata_linter/visitor.py
@@ -1015,7 +1015,9 @@ class _MetadataKeyVisitor(ast.NodeVisitor):
     def _visit_for_statement(self, node: ast.For | ast.AsyncFor) -> bool:
         self._record_metadata_merge_key_violations(node.iter)
         self.visit(node.iter)
-        iteration_may_raise = isinstance(node, ast.AsyncFor) or _iteration_may_raise(node.iter)
+        iteration_may_raise = isinstance(node, ast.AsyncFor) or _iteration_may_raise(
+            node.iter, advances=True
+        )
         if iteration_may_raise:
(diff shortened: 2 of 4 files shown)


seven332:
## Code Review: PR #28107

**Reviewed head:** `4cd0151407c1364caf86cfb8780b2059cb6f8bb8`

### Summary

The change correctly separates generator iterator acquisition from advancement at the shared linter boundary. All current callers declare their execution phase, so plain loops and eager/later comprehension paths retain exceptional aliases while nested generator construction remains deferred.

### Findings

No P0, P1, or P2 findings.

### Testing

- Positive later-advance and negative nested-construction regressions run through the public linter API with a real temporary source file.
- uv lock/sync, Ruff format/lint, and BasedPyright passed.
- Focused metadata contract: 21 passed.
- Complete mitm-addon suite: 6333 passed.
- GitHub CI was still in progress when this review was posted.

### Verdict

LGTM

PR #28110, opened 2026-08-19
docs(cli): explain when to pass --model for image generation

bingjiez666:
## Problem

`okou generate image --help` listed the image models but never said *when* an agent should pass `--model`. Two consequences:

- The rendered default read `(default: "gpt-image-1")`, which contradicts the run-level default injected for a chat thread.
- An agent that merely sees 11 model names tends to pick one out of habit, which silently overrides the model the user selected in the chat picker.

The runtime path is already correct: `resolveImageRequestModel` uses `command.getOptionValueSource("model")` and strips the Commander default from the request whenever a run default exists, so an omitted `--model` really does reach the server as "no model" and the chat selection applies. This PR only makes the help text say so.

## Change

- `--model` help now ends with: *"Omit it to generate with the model this chat is set to. Pass it only when the user names a specific model."*
- The Commander default is kept (behavior unchanged, `getOptionValueSource` still reports `default`), but its rendered annotation is now `(default: the model this chat is set to)` instead of the misleading literal alias.
- Help assertions added to the existing CLI help test.

## Notes

Wording matches the phrasing being standardized for video; the stale `ignored inside an agent run` phrasing was deliberately not copied.

### DIFF
--- a/turbo/apps/cli/src/commands/generate/__tests__/image.test.ts
+++ b/turbo/apps/cli/src/commands/generate/__tests__/image.test.ts
@@ -803,6 +803,12 @@ describe("okou generate image command", () => {
     expect(helpOutput).toContain("nano-banana-2");
     expect(helpOutput).toContain("seedream5-pro");
     expect(helpOutput).toContain("seedream5-lite");
+    expect(normalizedHelpOutput).toContain(
+      "Omit it to generate with the model this chat is set to. Pass it only when the user names a specific model.",
+    );
+    expect(normalizedHelpOutput).toContain(
+      "(default: the model this chat is set to)",
+    );
     expect(normalizedHelpOutput).toContain("support varies");
     expect(helpOutput).toContain("3840x2160");
     expect(helpOutput).toContain("edges divisible by 16");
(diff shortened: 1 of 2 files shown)


bingjiez666:
Closing as obsolete.

The problem this PR described — an agent seeing a bare list of image models in `--help` and passing `--model` out of habit, overriding the model selected in the chat picker — is now addressed on `main` from a different angle. `imageModelPreferenceDetail()` in `turbo/apps/cli/src/commands/shared/image-generate.ts` (L175-L183, called at L428) emits at runtime:

> `Run default model if direct image generation is used: <model>; omit --model so the server applies it`

What remains unlanded is only the static `--help` wording and the misleading `(default: "gpt-image-1")` annotation. That is not worth rebasing this branch for: it is ~1300 commits behind `main` and conflicting, and the model list has since grown to 13 entries (gpt-image-2, qwen-image-3, seedream5-pro/lite, nano-banana-2/lite, ...), so the diff no longer applies in any meaningful way.

If the rendered-default annotation is still worth fixing, it should be a fresh, small PR written against current `main`.

PR #28116, opened 2026-08-19
fix(guest-agent): fail artifact checkpoint on unreadable entries

### LINKED ISSUE #28090: bug(rust): artifact checkpoints silently omit unreadable nested entries
## Problem

File paths / line numbers:

- `crates/guest-agent/src/artifact/archive.rs:45-48` converts a nested directory `read_dir` failure into `Ok(())`.
- `crates/guest-agent/src/artifact/archive.rs:60` uses `entries.flatten()`, dropping per-entry iterator errors.
- `crates/guest-agent/src/artifact/archive.rs:66-84` silently skips entries when type detection, child opening, or metadata lookup fails.
- `crates/guest-agent/src/artifact/archive.rs:106-117` logs a file read/hash failure but still returns a successful manifest without that file.
- `crates/guest-agent/src/checkpoint/artifact.rs:178-187` accepts the resulting manifest as a valid snapshot plan, and `crates/guest-agent/src/checkpoint/artifact.rs:189-231` can commit it as the artifact snapshot.
- `crates/runner/src/executor/agent_run.rs:2110-2115` starts guest-agent with `sudo: false`, so ordinary artifact permission errors are reachable in production.

Problem category: logic

Artifact checkpoint traversal distinguishes an unreadable root from an empty root, but it does not make the same distinction below the root. For example, if a mounted artifact contains `kept.txt` and a regular `secret.txt` whose mode is changed to `000`, the non-sudo guest-agent cannot open `secret.txt`. `walk_entries` silently omits it, computes a manifest containing only `kept.txt`, and returns success. Checkpointing can then upload and commit that partial archive as a new artifact version. A later resume permanently loses the omitted file even though the checkpoint reported success.

The same partial-success behavior applies to unreadable nested directories, directory iteration failures, metadata failures, and mid-read I/O errors. This contradicts the existing strict behavior for an unreadable artifact root introduced by #15886: transient read failures must not be represented as valid artifact contents.

## Acceptance criteria

- Artifact manifest collection fails when it cannot enumerate a reachable nested directory or directory entry that is not intentionally excluded.
- A regular-file candidate that cannot be opened, inspected, or fully read fails manifest collection instead of being omitted.
- The failure propagates through artifact snapshot planning so no storage prepare, upload, commit, or final checkpoint request treats the partial manifest as successful.
- Symlinks, FIFOs, `.git`, and `.vm0` remain intentionally excluded.
- Regression tests cover an unreadable nested directory and an unreadable regular file, and verify that checkpointing fails before storage API calls.

## Validation

- `cargo test -p guest-agent --lib artifact::archive::tests --jobs 1` passes all 33 current archive tests; none exercises unreadable nested entries or regular files.
- Open-issue searches for unreadable nested artifact checkpoints returned no match.

seven332:
## Summary

- fail artifact manifest collection on nested directory enumeration, entry inspection, regular-file open/metadata, and file read errors
- preserve `.git`, `.vm0`, symlink, FIFO, and root-level inaccessible `lost+found` exclusions
- add real permission regressions and verify unreadable artifacts fail before storage prepare, upload, or commit calls

## Tests

- `cargo test -p guest-agent --lib artifact::archive::tests --jobs 1`
- `cargo test -p guest-agent --lib checkpoint::artifact::tests --jobs 1`
- `cargo clippy -p guest-agent --all-targets --all-features --jobs 1 -- -D warnings`
- `cargo doc -p guest-agent --all-features --no-deps --jobs 1`
- `cargo test -p guest-agent --jobs 1`

Closes #28090

### DIFF
--- a/crates/guest-agent/src/artifact/archive.rs
+++ b/crates/guest-agent/src/artifact/archive.rs
@@ -7,14 +7,11 @@ use api_contracts::generated::constants::storages::{
 };
 use flate2::Compression;
 use flate2::write::GzEncoder;
-use guest_common::log_warn;
 use sha2::{Digest, Sha256};
 use std::fs::{self, File, Metadata};
 use std::io::{self, Read};
 use std::path::{Component, Path, PathBuf};
 use thiserror::Error;
-
-const LOG_TAG: &str = "sandbox:guest-agent";
 
 /// Walk directory and compute SHA-256 for each file, skipping `.git` and `.vm0`.
 #[cfg(target_os = "linux")]
@@ -42,10 +39,12 @@ fn walk_dir(
     out: &mut Vec<FileEntry>,
     path_bytes: &mut u64,
 ) -> Result<(), ArchiveError> {
-    let entries = match current.read_dir() {
-        Ok(e) => e,
-        Err(_) => return Ok(()),
-    };
+    let entries = current
+        .read_dir()
+        .map_err(|source| ArchiveError::DirectoryRead {
+            path: relative.to_string(),
+            source,
+        })?;
     walk_entries(current, relative, entries, out, path_bytes)
 }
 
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #28116

**Reviewed head:** `2099257f9b6b3ffe4393151f1d58dd32db026ac2`

### Summary

The change replaces silent artifact-entry omission with fail-fast, path-aware errors while preserving the existing no-follow filesystem boundary and intentional exclusions, including the narrow root-level inaccessible `lost+found` case.

### Findings

No findings.

### Testing

Focused archive and checkpoint tests, Rust formatting, Clippy, docs, and the complete `guest-agent` test suite passed. The four permission regressions also passed from a root-started test binary, exercising the uid/gid 65534 child path. Multi-architecture guest-image builds and a live VM checkpoint were left to CI.

### Verdict

LGTM

seven332:
Closing because we decided not to pursue this behavior change.

PR #28118, opened 2026-08-19
test(ci): wait for runner journal visibility

### LINKED ISSUE #28112: test(ci): wait for runner journal visibility
## Problem and evidence

The `runner-behavior-lane-d` CI job exercises the keep-alive sandbox's atomic final-operation fence. It starts an independently owned long-running `runner exec`, completes the supervised turn, verifies that the rejected sandbox is destroyed, and then checks the runner service journal for the exact busy-rejection reason.

The check failed intermittently in [run 32212298333, job 95947449250](https://github.com/vm0-ai/vm0/actions/runs/32212298333/job/95947449250) on PR #28106. The sandbox behavior completed as expected:

- the independently owned exec terminated with `unexpected end of file` when the sandbox was destroyed;
- the supervised turn completed successfully;
- the same EOF is present in successful executions, such as [run 32212090492, job 95947825401](https://github.com/vm0-ai/vm0/actions/runs/32212090492/job/95947825401).

However, `.github/scripts/runner-behavior-keep-alive.sh` reads the service journal once immediately after destruction and requires the line `normal operations busy while preparing park`. If the event is not visible to that query yet, the script fails with:

```text
FAIL: overlapping runner exec did not reject idle admission as busy
```

In the most recent 100 `Crates` workflow runs inspected, lane D completed 11 times: 10 passed and this one failed. PR #28106 changes only security workflow files and does not modify runner behavior. The failed host's late journal could not be recovered, so delayed journal visibility is the leading inference rather than a directly captured late record.

## Decision

Replace the one-shot journal assertion with an inline bounded readiness poll in the existing remote script. Query the same systemd invocation up to 10 times at 0.5-second intervals and succeed only when the exact existing busy-rejection message appears.

Keep the existing independent-exec termination and fresh-sandbox/non-reuse assertions unchanged. Do not add a production result surface or extract a one-use helper from the SSH heredoc.

## Alternatives considered

- `journalctl --sync` does not establish ordering with service stderr that journald has not received and forces broader persistence work on the shared host.
- Removing the exact message check would no longer distinguish the intended atomic busy fence from another fail-closed destruction path.
- Adding structured runner state or CLI output solely for this assertion would create unjustified production and compatibility surface.
- Extracting a polling helper only for isolated tests conflicts with the repository's integration-test and YAGNI guidance because the runtime code lives inside a quoted remote heredoc.

## Risks and mitigations

- A different rejection path remains a failure because the exact busy text is still mandatory.
- A terminal visibility miss adds at most approximately five seconds before the existing diagnostic.
- Successful runs normally continue after one invocation-filtered journal query; delayed runs make at most ten.

## Acceptance criteria

- Wait for the exact busy-rejection journal message with a short, bounded polling window instead of a single immediate read.
- Continue to fail when the exact message never appears; do not accept alternative rejection reasons.
- Preserve the existing assertions that the independently owned exec terminates and that the rejected sandbox is not reused.
- Pass shell syntax and shellcheck validation for the changed behavior script.
- Validate the submitted head through the real `runner-behavior-lane-d` metal-host/Firecracker integration job.
- Do not skip the runner behavior test or retry the whole CI job as the fix.

## Scope

This issue is limited to making the existing CI observation deterministic. It does not change sandbox fencing, runner lifecycle behavior, production logging, APIs, persisted state, or protocols.

seven332:
## Summary

- Poll the runner service journal for up to five seconds after the overlapping exec terminates.
- Keep requiring the exact busy-fence log message before the test can pass.
- Preserve the existing assertions that the rejected exec terminates and its sandbox is not reused.

## Scope

This changes only the runner keep-alive CI integration script. It does not change runner behavior, add a job-level retry, or skip any assertion.

## Validation

- `bash -n .github/scripts/runner-behavior-keep-alive.sh`
- `shellcheck .github/scripts/runner-behavior-keep-alive.sh`
- `git diff --check`
- `./scripts/check-file-size.sh .github/scripts/runner-behavior-keep-alive.sh`

The metal-host lane-D integration test will run in PR CI.

Closes #28112

### DIFF
--- a/.github/scripts/runner-behavior-keep-alive.sh
+++ b/.github/scripts/runner-behavior-keep-alive.sh
@@ -246,10 +246,19 @@ fi
 wait "$OVERLAP_EXEC_PID" 2>/dev/null || true
 OVERLAP_EXEC_PID=""
 
-OVERLAP_LOGS=$(sudo journalctl --no-pager \
-  "_SYSTEMD_INVOCATION_ID=$INVOCATION_ID" 2>&1) \
-  || fail "failed to read overlap runner logs"
-grep -F 'normal operations busy while preparing park' <<<"$OVERLAP_LOGS" >/dev/null \
+OVERLAP_BUSY_LOGGED=false
+for _ in $(seq 1 10); do
+  OVERLAP_LOGS=$(sudo journalctl --no-pager \
+    "_SYSTEMD_INVOCATION_ID=$INVOCATION_ID" 2>&1) \
+    || fail "failed to read overlap runner logs"
+  if grep -F 'normal operations busy while preparing park' \
+    <<<"$OVERLAP_LOGS" >/dev/null; then
+    OVERLAP_BUSY_LOGGED=true
+    break
+  fi
+  sleep 0.5
+done
+[ "$OVERLAP_BUSY_LOGGED" = true ] \
   || fail "overlapping runner exec did not reject idle admission as busy"
 
 echo "--- Overlap turn 2: busy sandbox was not reused ---"


seven332:
## Code Review: PR #28118

**Reviewed head:** `eff4c632320707a73290d398fc5ea77a89099256`

### Summary

The bounded journal poll addresses asynchronous journald visibility without changing runner behavior or weakening the busy-fence assertion.

### Findings

No findings.

### Testing

Bash syntax, ShellCheck, diff, and file-size checks passed locally. The metal-host lane-D integration remains to be verified by PR CI.

### Verdict

LGTM

PR #28152, opened 2026-08-19
feat(platform): rename the Works nav entry to Channels

Lunarivibe:
## What

Renames the sidebar entry currently labelled **Works** to **Channels**, across all ten locales.

| | before | after |
|---|---|---|
| Rail caption | `Works` | `Channels` |
| Expanded nav | `Where {agent} works` | `{agent}'s channels` |
| Page title | `Where {agent} works` | `{agent}'s channels` |
| Document title | `Works` | `Channels` |

## Why

`Works` was a truncation of the expanded label `Where {agent} works`. Standing alone under an icon it reads as a noun ("artworks", "engineering works"), and it shares its first four letters with **Workflows** two rows above, so the two are hard to tell apart at a glance.

The page behind it lists Slack, Microsoft Teams, GitHub, Feishu, Telegram, Strapi and AgentPhone — the places the agent can be reached from or triggered by. `Channels` names that directly, and keeps a clean split from **Connectors** (what the agent can call).

The ambiguity had already leaked into the translations. Several locales had read the English as a verb:

- de-DE — `Funktioniert` ("it functions")
- fr-FR — `Fonctionne` / `Travail`
- id-ID — `Berfungsi` / `Bekerja`
- ko-KR — `{name}가(이) 작동하는 곳` ("where {name} operates", as of a machine)
- pt-BR — `Trabalha`
- hi-IN — `काम करता है`

All ten now use the channel wording each locale was already using in the page description (`…through these channels`). `ja-JP` keeps 連携先 — it was already correct and is coherent across all four of its strings, so that file is untouched.

Every new caption fits the rail: the longest is `Channels` at 8 characters, still shorter than the existing `Connectors`.

## Scope

Copy only. The `/works` route, the `works` nav id, the page layout and all behaviour are unchanged — no redirects needed, no links break.

## Left out, on purpose

- **The route.** `/works` → `/channels` would be more consistent but breaks existing links; happy to do it as a follow-up with a redirect.
- **The rail icon.** That entry renders Slack's colour logo (`iconImg: slackIcon`) even though seven destinations sit on the page. Swapping it for the neutral `LayoutGrid` already wired up as the fallback is a visual change, so it is not bundled into a copy rename.

## Test plan

No behavioural change and no test asserts these strings. Verified by inspection that the four keys resolve in all ten locale files and that the `{{agentName}}` / `{{displayName}}` interpolations are preserved at all three call sites (`zero-sidebar.tsx`, `zero-mobile-breadcrumb.ts`, `zero-works-page.tsx`).

### DIFF
--- a/turbo/apps/platform/src/i18n/locales/de-DE/common.json
+++ b/turbo/apps/platform/src/i18n/locales/de-DE/common.json
@@ -442,7 +442,7 @@
         "newChat": "Neuer Chat",
         "settings": "Einstellungen",
         "workflows": "Workflows",
-        "works": "Einsatzorte von {{agentName}}"
+        "works": "Kanäle von {{agentName}}"
       },
       "rail": {
         "new": "Neu",
@@ -5391,7 +5391,7 @@
     },
     "connected": "Verbunden",
     "connectedWithDetail": "Verbunden ({{detail}})",
-    "documentTitle": "Funktioniert",
+    "documentTitle": "Kanäle",
     "feishuConnected": "Feishu hat erfolgreich eine Verbindung hergestellt",
     "github": {
       "adminInstall": "Bitten Sie einen Organisationsadministrator, die GitHub-App zu installieren",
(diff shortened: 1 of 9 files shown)


Lunarivibe:
Closing for now — parking the rename until the three-column nav settles. The branch `feat/rename-works-nav-to-channels` is kept, so this can be reopened as-is or split into a rail-only change.

PR #28161, opened 2026-08-19
perf(runner): reserve axiom channel before serialization

### LINKED ISSUE #28124: perf(rust): reserve Axiom channel capacity before serializing events
## Problem

`AxiomLayer::on_event` serializes every WARN/ERROR event before checking whether the bounded dispatcher channel can accept it:

```rust
let value = serialize_event(event);
if self.tx.try_send(Msg::Event(value)).is_err() {
    // count the dropped event
}
```

When the dispatcher is blocked in an ingest request (the HTTP timeout is 10 seconds) and the 1,024-entry channel fills, subsequent events are intentionally dropped. However, every dropped event still visits all fields, allocates a JSON map and strings, formats `Debug` and error-chain fields, and creates a timestamp before `try_send` reports `Full`.

During a WARN/ERROR burst, this makes the overload path amplify CPU and allocation pressure on runner threads even though the queue is designed to keep tracing non-blocking and bounded. The existing `burst_past_channel_cap_drops_without_blocking` test verifies queue length and completion, but it does not verify that dropped events avoid serialization work.

## Scope

- Reserve bounded-channel capacity before calling `serialize_event`, using `try_reserve` or an equivalent non-blocking admission step.
- Preserve the existing drop counter and periodic local diagnostic for both full and closed channels.
- Preserve the current Axiom payload shape and sibling local tracing behavior for admitted events.
- Add a regression test with an observable `Debug` field (or equivalent visitor side effect) proving that an event rejected by a full channel is not serialized.

## Acceptance criteria

- Full or closed Axiom channels reject events without visiting or formatting their fields.
- Admitted events retain the current serialized payload.
- The producer path remains non-blocking, and existing drop diagnostics continue to work.

seven332:
## Summary

- reserve a bounded Axiom channel slot before serializing WARN/ERROR events
- keep full and closed channel rejection on the existing drop counter and local diagnostic path
- add a `Debug` side-effect regression test proving rejected events are not formatted

## Scope

- preserves the existing admitted-event JSON payload and sibling local tracing behavior
- does not change Axiom batching, HTTP timeout, shutdown, retry, filter, or channel capacity behavior
- does not change APIs, persisted state, runner protocols, or deployment compatibility boundaries

## Validation

- `cargo test --manifest-path crates/Cargo.toml -p runner axiom_layer::tests`
- `cargo clippy --manifest-path crates/Cargo.toml -p runner --all-targets --all-features`
- `cargo test --manifest-path crates/Cargo.toml -p runner`
- pre-commit workspace Rust formatting, documentation, Clippy, file-size, and commit-message checks

Closes #28124

### DIFF
--- a/crates/runner/src/axiom_layer.rs
+++ b/crates/runner/src/axiom_layer.rs
@@ -192,8 +192,7 @@ where
     S: Subscriber + for<'a> LookupSpan<'a>,
 {
     fn on_event(&self, event: &Event<'_>, _: Context<'_, S>) {
-        let value = serialize_event(event);
-        if self.tx.try_send(Msg::Event(value)).is_err() {
+        let Ok(permit) = self.tx.try_reserve() else {
             // Bounded-channel full or dispatcher gone. Emit a periodic
             // best-effort diagnostic under `INTERNAL_TARGET`; if tracing
             // observes it, the Axiom per-layer filter keeps it out of remote
@@ -206,7 +205,10 @@ where
                     "axiom channel full",
                 );
             }
-        }
+            return;
+        };
+
+        permit.send(Msg::Event(serialize_event(event)));
     }
 }
 
--- a/crates/runner/src/axiom_layer_tests.rs
+++ b/crates/runner/src/axiom_layer_tests.rs
@@ -6,7 +6,7 @@
 //! endpoint with the TS-compatible payload shape.
 //!
 
-use std::sync::atomic::AtomicU64;
+use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex, mpsc};
 use std::thread;
 use std::time::Duration;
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #28161

**Reviewed head:** `3f3d2861bdf1d608b41d370b53140e63f09e253f`

### Summary

The change correctly makes bounded-channel admission atomic and non-blocking before Axiom event serialization, while retaining the existing rejection diagnostics and admitted-event payload path. The regression test deterministically covers both full and closed channels through the tracing layer callback.

### Findings

No findings.

### Testing

Local focused Axiom tests, Rust formatting, runner Clippy, runner documentation, and the full runner test suite passed. Existing tests continue to cover payload compatibility, sibling local tracing, and burst non-blocking behavior. CI-only runner deployment/image checks and security scanning were not reproduced locally and remain subject to GitHub Actions.

### Verdict

LGTM

PR #28162, opened 2026-08-19
perf(python): bound firewall auth identity ownership

### LINKED ISSUE #28126: perf(python): evict obsolete firewall auth identities for active runs
## Problem and evidence

`crates/runner/mitm-addon/src/firewall_auth_cache.py` stores process-global auth state in `_auth_state`, keyed by `(run_id, api_id, auth_identity)`. `_get_auth_state()` inserts a new entry for every first use of an identity, while registry cleanup removes entries only when their `run_id` disappears.

That lifecycle leaves obsolete identities behind when registry state changes for a run that remains active:

- `auth._cached_firewall_auth_identity()` intentionally recomputes a content-derived identity for each replacement registry snapshot.
- `test_registry_reload_recomputes_firewall_auth_identity` confirms that changing auth material for the same `run_id` and `api_id` produces a second identity and a second auth fetch.
- Active connector runtime synchronization can repeatedly replace the same run's registry entry without unregistering it.
- Each historical `_FirewallAuthState` can retain the full `FirewallAuthPayload`, in-flight task ownership, force-refresh markers, and cooldown state. The payload size is bounded, but the number of retained identities is not.
- Registry reload cleanup scans all retained auth keys, so historical identities also increase synchronous reload work.

This is distinct from #19109, which added content identity to prevent stale credential reuse, and #21964, which memoized identity derivation within one registry snapshot. Both behaviors remain required.

## Decision

Proceed in one PR with a registry-generation ownership lease and one globally owned content identity per `(run_id, api_id)`.

The content-derived `auth_identity` remains the credential-equivalence boundary. A private, monotonic, process-local registry generation only determines whether a request came from the current published snapshot and may acquire global cache ownership.

- Assign one trusted generation after a changed registry parses, classifies, and compiles successfully. Overwrite any marker-shaped input before publishing accepted `vm_info` entries.
- Carry generation on `FirewallAuthCacheKey` as a field excluded from equality, hashing, and representation.
- Track the current generation per active run and the single globally owned key per `(run_id, api_id)`.
- A current-generation lookup reuses an existing state when the content identity is unchanged. When identity changes, it removes the previous global key and installs the new identity.
- A stale-generation lookup may share an already-owned equal content identity, but a superseded identity receives detached state and cannot insert globally.
- Existing old fetches and waiters retain direct references to their detached `_FirewallAuthState`, so they can finish and release admission without repopulating or mutating the new identity.
- Active-run identity replacement remains lazy until that API is used; registry reload does not hash unused APIs.
- Removed-run, path-switch, and registry-unavailable cleanup clears auth state, scope ownership, and generation authority together.

This separates credential equivalence from request freshness. It preserves identical-input cache reuse while preventing delayed work from an older registry snapshot from reacquiring ownership based only on arrival order.

## Alternatives considered

- **Prune on latest auth use without generation:** smaller, but a delayed old-snapshot request can arrive last and evict/reacquire ownership for obsolete state.
- **Move auth state into registry snapshot sidecars:** gives strong object-lifetime isolation, but broadly rewrites cache and flow metadata contracts and loses identical-input cross-snapshot reuse unless another global layer is added.
- **Bound identities with LRU or TTL:** limits memory numerically but retains secret-bearing payloads unnecessarily, needs arbitrary tuning, and does not prevent stale ownership reacquisition.
- **Eagerly derive active identities during registry reload:** can prune precisely but reverses #21964's lazy boundary and adds work for unused APIs.

## Constraints

- A same-run update that changes auth inputs must fetch under a distinct content identity and never reuse the old payload.
- Generation must not change cache-key equality or credential identity.
- In-flight work for an old identity must complete safely through detached state without recreating global ownership.
- A delayed old response must not clear or mark the new identity's state.
- Same content identity across a registry replacement must preserve payload hits, coalescing, force-refresh markers, and cooldown behavior.
- Registry generation allocation must remain monotonic across production path switches; only explicit test reset may restart it.
- Run removal and registry-unavailable fail-closed cleanup must keep their current behavior.
- Cleanup must not eagerly hash unused APIs.

## Risks and mitigations

- **Generation pollutes credential equality:** exclude it from dataclass equality/hash and cover unchanged-content reuse across generations.
- **Caller-controlled generation marker:** overwrite the private field after VM validation and test forged marker-shaped JSON.
- **Stale request reacquires ownership:** allow global insertion only for the current run generation; test stale lookup after new ownership.
- **Old task corrupts new state:** keep distinct state objects per content identity and assert the new payload and lifecycle markers after old completion.
- **Index drift:** centralize insertion/removal and clear the state dictionary, scope index, and generation authority together.
- **Generation reuse:** keep the allocator outside path-local registry resets.

## Acceptance criteria

- After current-snapshot auth use, at most one globally owned identity exists per `(run_id, api_id)`.
- Repeated same-run registry replacements followed by auth use do not grow retained auth state or cleanup work with historical identities.
- Changed auth inputs trigger a fresh fetch, detach the previous identity, and retain only the new identity globally.
- Unchanged auth inputs across a registry generation reuse the existing cached payload and lifecycle state.
- An old in-flight fetch and its existing waiters can finish, but completion cannot repopulate or mutate the new identity.
- A stale-generation lookup and delayed old 401 cannot reacquire, clear, or mark superseded global ownership.
- Removed-run, path-switch, registry-unavailable, cache-hit, coalescing, cancellation, admission, force-refresh, and cooldown behavior remains covered and green.
- Ruff formatting/linting, basedpyright, focused auth/registry/request/response tests, and the full mitm-addon pytest suite pass.

## Delivery notes

The change is internal, in-memory Python state. It changes no public API, persisted format, runner protocol, or deployment contract and requires no migration or coordinated rollout. Restart and rollback both begin with an empty auth cache.

seven332:
## Summary

- assign each published proxy registry snapshot a private, monotonic auth ownership generation without changing the public registry mapping
- retain one globally owned firewall auth identity per run/API scope while allowing equal content identities to reuse state across registry generations
- detach superseded in-flight state so existing requests can finish, while stale generations and delayed 401 responses cannot reacquire or mutate current ownership
- add integration coverage for repeated replacement, unchanged identity cache and in-flight reuse, stale lookup, in-flight completion, path switches, and superseded response recovery

## Scope

This change is limited to process-local Python mitmproxy auth-cache lifecycle state. It does not change public APIs, persisted registry JSON, runner protocols, credential identity material, or deployment sequencing.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` — 6374 passed
- `lefthook run pre-commit`
- `git diff --check`

Closes #28126

### DIFF
--- a/crates/runner/mitm-addon/src/auth.py
+++ b/crates/runner/mitm-addon/src/auth.py
@@ -50,6 +50,7 @@ from aws_sigv4 import (
 )
 from aws_sigv4_body_admission import MAX_AWS_SIGV4_REQUEST_BODY_BYTES
 from firewall_auth_cache import (
+    FIREWALL_AUTH_REGISTRY_GENERATION_ATTRIBUTE,
     FirewallAuthCacheKey,
     FirewallAuthFetchSaturatedError,
     InvalidBillableAuthExpiryError,
@@ -307,6 +308,7 @@ def _build_firewall_auth_context(
             firewall_base=firewall_base,
             auth_request=auth_request,
         ),
+        registry_generation=_firewall_auth_registry_generation(vm_info),
     )
     flow.metadata[metadata_keys.FIREWALL_AUTH_CACHE_KEY] = auth_cache_key
     return _FirewallAuthContext(
@@ -316,6 +318,13 @@ def _build_firewall_auth_context(
         auth_request=auth_request,
         auth_cache_key=auth_cache_key,
     )
+
+
+def _firewall_auth_registry_generation(vm_info: dict) -> int | None:
+    generation = getattr(vm_info, FIREWALL_AUTH_REGISTRY_GENERATION_ATTRIBUTE, None)
+    if isinstance(generation, bool) or not isinstance(generation, int):
+        return None
+    return generation
 
 
 def _build_firewall_auth_identity(
--- a/crates/runner/mitm-addon/src/firewall_auth_cache.py
+++ b/crates/runner/mitm-addon/src/firewall_auth_cache.py
@@ -162,7 +162,7 @@ def clear_cached_firewall_headers(cache_key: FirewallAuthCacheKey) -> None:
         state.cache = None
 
 
-def evict_stale_cache_keys(active_run_generations: dict[str, int]) -> None:
+def reconcile_registry_cache_ownership(active_run_generations: dict[str, int]) -> None:
     """Reconcile cache ownership with the current registry generation."""
     _active_registry_generations.clear()
     _active_registry_generations.update(active_run_generations)
--- a/crates/runner/mitm-addon/src/firewall_auth_cache.py
+++ b/crates/runner/mitm-addon/src/firewall_auth_cache.py
@@ -28,6 +28,12 @@ class FirewallAuthCacheKey:
     run_id: str
     api_id: str
     auth_identity: str
+    registry_generation: int | None = field(
+        default=None,
+        compare=False,
+        hash=False,
+        repr=False,
+    )
 
 
 @dataclass
@@ -63,7 +69,12 @@ class FirewallAuthStateSnapshotForTests:
     last_force_refresh_monotonic_at: float | None = None
 
 
+FIREWALL_AUTH_REGISTRY_GENERATION_ATTRIBUTE = "_firewall_auth_registry_generation"
+type _FirewallAuthScope = tuple[str, str]
+
 _auth_state: dict[FirewallAuthCacheKey, _FirewallAuthState] = {}
+_owned_auth_keys: dict[_FirewallAuthScope, FirewallAuthCacheKey] = {}
+_active_registry_generations: dict[str, int] = {}
 MAX_CONCURRENT_FIREWALL_AUTH_FETCHES = 4
 MAX_ADMITTED_FIREWALL_AUTH_FETCHES = 16
 _admitted_firewall_auth_fetches = 0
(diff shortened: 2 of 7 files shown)


seven332:
## Code Review: PR #28162

**Reviewed head:** `4bddcbafb8dc3f31a2c03c6e64ea925bf48cdbbb`

### Summary

The registry-generation ownership lease cleanly bounds process-global firewall auth state to one content identity per run/API scope while preserving equal-input cache reuse and allowing superseded in-flight work to finish without reclaiming ownership.

### Findings

No findings. The initial maintainability concern about an eviction-only function name was resolved in the current head by renaming the operation to `reconcile_registry_cache_ownership`.

### Testing

Ruff format/lint, basedpyright, 131 focused tests, the full 6373-test mitm-addon suite, lefthook pre-commit, and `git diff --check` passed. The final rename-only commit was revalidated with Ruff, basedpyright, and the focused suite.

### Verdict

LGTM

seven332:
## Code Review: PR #28162

**Reviewed head:** `2302f2e85d03c5709afde19855e6e042db9167c5`

### Summary

The registry-generation ownership lease bounds global firewall auth state while preserving equal-identity cache and in-flight reuse. Rebase compatibility and the full lifecycle were re-reviewed on the submitted head.

### Findings

No remaining findings. Self-review added request-entry regression coverage for unchanged-identity in-flight coalescing across registry generations in `2302f2e85d`.

### Testing

The targeted regression test passed, followed by Ruff format/lint, basedpyright, the full mitm-addon suite (`6374 passed`), lefthook pre-commit, and `git diff --check`.

### Verdict

LGTM

PR #28168, opened 2026-08-19
fix(composer): stop redundant feedback note dom writes during composition

bingjiez666:
## Summary

Follow-up to #28045, which did not fix the reported truncation. Device tracing narrowed the bug to the **inline feedback note**, not the plain composer, and to redundant DOM writes rather than a timing race.

Both protections are gated behind the existing **`composerImeSubmitFlush`** feature switch (off by default, staff orgs only), so with the switch off behavior is byte-for-byte unchanged:

- Skip `className` / `hidden` / `textContent` writes in the feedback note's node view when the value is unchanged.
- Call `domObserver.forceFlush()` before `flush()` when reading a submission.

## Evidence

A Safari Web Inspector trace on the reported iOS session, timestamps in seconds:

| t | event |
| --- | --- |
| 31.60 | `compositionend` — first segment |
| 32.09 | draft `PATCH` containing the first segment, so it reached the document |
| 31.76 | `compositionstart` — second segment |
| 33.31 | `compositionend` — second segment; the DOM holds both |
| 33.95 | `POST /chat/events` with a `prompt` containing **only the first segment** |

The send happened 0.64s after the second `compositionend`, far past the ~20ms prosemirror-view needs to run `endComposition`. The second segment never reached `editor.state.doc` at all, so this is not a race the composition gate can win.

## Cause

`createFeedbackItemNodeView`'s render ran on every document update and assigned four DOM values unconditionally. Assigning an attribute queues a `MutationObserver` record even when the value is unchanged, and the node view's `ignoreMutation` only ignores mutations whose target is outside `contentDOM`:

```ts
ignoreMutation(mutation) {
  return mutation.type !== "selection" && !contentDOM.contains(mutation.target);
}
```

`contentDOM.contains(contentDOM)` is `true`, so a re-render's `contentDOM.className = ...` was handed to ProseMirror as a content change inside the region the user was composing in — a documented way to lose an in-flight composition.

The `forceFlush` change addresses a second defect found while reading prosemirror-view 1.42.2: `DOMObserver.flush()` returns early when `this.flushingSoon > -1`, which is exactly the state a composition leaves behind, so #28045's `flush()` call could not make progress. `forceFlush()` clears that timer but is itself a no-op when nothing is pending, so both calls are needed.

## Feature switch

Both protections read `composerImeSubmitFlush`:

- The submit reader already gated `forceFlush`/`flush` on it.
- The node view now receives the resolved switch value (via `runtime.skipUnchangedNoteWrites`, set at mount) and only skips unchanged writes when it is on. With the switch off, the node view writes exactly as before this PR.

This keeps the whole IME-composition change contained to opt-in staff orgs until device verification, consistent with #28045's gating.

## Test plan

- Added `does not rewrite feedback note attributes while typing into it` in `chat-inline-feedback.test.tsx` (switch enabled). It observes attribute mutations on the note's `contentDOM` while typing and asserts none occur.
- **Verified as a real regression guard, and that the switch actually gates it**: with the switch enabled it passes; flipping the switch to `false` (old behavior) makes it fail; reverting only the value guards also makes it fail.
- `chat-inline-feedback.test.tsx` 31 passed, `chat-run-queue.test.tsx` 33 passed, `@okouai/app` + `@okouai/core` `check-types` clean, `pnpm format` clean, `pnpm turbo run lint` 13 tasks 0 errors.
- The `forceFlush` half stays untested; happy-dom has no deferred flush to force.
- Device verification still required: enable `composerImeSubmitFlush` in Lab, reproduce the two-segment composition on iOS, and confirm the full note is sent.

### DIFF
--- a/turbo/apps/platform/src/signals/zero-page/tiptap-workflow-composer.ts
+++ b/turbo/apps/platform/src/signals/zero-page/tiptap-workflow-composer.ts
@@ -275,9 +275,58 @@ function connectComposerFeedback(
  * ProseMirror as content changes, which can discard an in-flight IME
  * composition, so re-renders must not rewrite values that already match.
  */
-function setClassName(element: HTMLElement, className: string): void {
-  if (element.className !== className) {
-    element.className = className;
+function setClassName(
+  element: HTMLElement,
+  className: string,
+  skipUnchanged: boolean,
+): void {
+  if (skipUnchanged && element.className === className) {
+    return;
+  }
+  element.className = className;
+}
+
+interface FeedbackItemNodeViewElements {
+  readonly dom: HTMLElement;
+  readonly noteDom: HTMLElement;
+  readonly placeholderDom: HTMLElement;
+  readonly contentDOM: HTMLElement;
+  readonly quoteText: HTMLElement;
+}
+
+function renderFeedbackItemNodeView(
+  elements: FeedbackItemNodeViewElements,
+  nextNode: ProseMirrorNode,
+  skipUnchanged: boolean,
+): void {
+  const { dom, noteDom, placeholderDom, contentDOM, quoteText } = elements;
+  const { quote, showDivider, fill } = feedbackItemNodeAttributes(nextNode);
+  // The top padding is breathing room for the dashed divider, so only the
+  // items that draw one get it. The first item keeps the editor's own pt-4.
+  setClassName(
+    dom,
+    `flex flex-col gap-1.5 pb-1.5${
+      showDivider ? " border-t border-dashed border-border/60 pt-1.5" : ""
+    }`,
+    skipUnchanged,
+  );
+  setClassName(
+    noteDom,
+    `relative${fill ? " min-h-[96px]" : ""}`,
+    skipUnchanged,
+  );
+  const placeholderHidden = nodeText(nextNode).length > 0;
+  if (!skipUnchanged || placeholderDom.hidden !== placeholderHidden) {
+    placeholderDom.hidden = placeholderHidden;
+  }
+  setClassName(
+    contentDOM,
+    "relative w-full px-1 py-1 text-[0.9375rem] leading-snug " +
+      `text-foreground outline-none [&_p]:m-0${fill ? " min-h-[96px]" : ""}`,
+    skipUnchanged,
+  );
+  if (!skipUnchanged || quoteText.textContent !== quote) {
+    quoteText.textContent = quote;
   }
 }
 
@@ -589,6 +638,7 @@ function createFeedbackItemNodeView(
   node: ProseMirrorNode,
   removeFeedback: (id: number) => void,
   localizedUi: Set<() => void>,
+  skipUnchangedWrites: () => boolean,
 ): NodeView {
   const dom = document.createElement("div");
   dom.dataset.feedbackItem = "";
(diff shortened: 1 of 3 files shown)


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: f91ed69314af9eae70372f0f430afb61b43114b8
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Stops the inline feedback note's node view from rewriting unchanged `className` / `hidden` / `textContent` on every document update, and calls `domObserver.forceFlush()` before `flush()` when reading a submission. Both are gated behind the existing `composerImeSubmitFlush` switch, so the switch-off path is unchanged.

The device trace in the PR body is the strongest evidence I have seen on this bug so far: the second composed segment was absent from the submitted payload 0.64s after its `compositionend`, which rules out the timing race #28045 assumed and points at the composing region being disturbed instead.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | `fix:` commit must carry a regression test | Pass | `chat-inline-feedback.test.tsx` adds a test that observes attribute mutations on the note's `contentDOM`; the PR reports it failing when the value guards are reverted |
| `REVIEW.md` | Feature regression guard — change must be scoped/contained | Pass | Both halves read `composerImeSubmitFlush` (`enabled: false` + staff hashes); with it off, `renderFeedbackItemNodeView` writes exactly as before |
| `REVIEW.md` | Type safety — no `any`, casts only where necessary | Pass | One documented cast for the undeclared `domObserver`; new `FeedbackItemNodeViewElements` is explicit |
| `REVIEW.md` | Style — function length, duplication | Pass | `render` extracted to a module-level `renderFeedbackItemNodeView`, which is what brought `createFeedbackItemNodeView` back under the 128-line rule |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, contract, or `jsonb(...)` change |
| `REVIEW.md` | Event sourcing / optimistic events | N/A | No event shape, event ID, or reconciliation change |
| `docs/bad-smell.md` | §1/§2 no `any`, no lint or type suppressions | Pass | Neither appears in the diff |
| `docs/bad-smell.md` | §7 fail fast — no silent fallback | Pass | `flushPendingDomChanges` still dereferences `domObserver` directly and throws if it ever disappears |
| `docs/bad-smell.md` | §5 no dynamic imports | Pass | Static imports only |
| `docs/fallback.md` | §3 no fallback for a state the contract cannot produce | Pass | No new tolerance branch; the `skipUnchanged` parameter is switch plumbing, not a guard against an impossible state |
| `docs/fallback.md` | §2 non-GA feature switch needs no compatibility code | Pass | No dual-read/dual-write or migration added |
| `docs/fallback.md` | §9 introduced fallbacks declared in the PR summary | N/A | The diff introduces no fallback; a feature switch is containment, not a fallback |
| `docs/fallback.md` | §1 no negative tests pinning removed behavior | Pass | No test asserts the switch-off (buggy) behavior; the author verified it manually instead of encoding it |
| `docs/testing/anti-patterns.md` | AP-4 never mock internal code | Pass | No `vi.mock` added |
| `docs/testing/anti-patterns.md` | AP-10 exercise the production bootstrap | Pass | Uses `detachedSetupPage` |
| `docs/testing/anti-patterns.md` | AP-5 no fake timers | Pass | Real timers; `MutationObserver.takeRecords()` used to drain deterministically instead of waiting |
| `docs/testing/anti-patterns.md` | AP-1 assert behavior, not mock calls | Partial — accepted | The assertion is on DOM attribute writes, which is closer to implementation than to user-visible output (AP-7). It is the most direct expression of the defect available, and unlike #28045's test it demonstrably fails without the fix, so the trade is worth it |
| `docs/testing/app-testing.md` | Feature flags via `setupPage` `featureSwitches`; `waitFor` for async; no file-level cleanup | Pass | All three followed; the observer is disconnected inline rather than in an `afterEach` |
| `docs/effect.md` | Command owns the transition; resources released by their owner | Pass | Switch read synchronously in the mount command next to the existing `includeQuoteOnlyFeedback` read; the new runtime field is a plain boolean needing no teardown |
| `docs/cache.md` | RC-2 no render-time signal allocation, RC-3 no unbounded cache | Pass | No new signals or caches; the node view reads a mutable runtime field through a getter |
| `.claude/skills/ccstate/SKILL.md` | `computed` memoizes; read state through the store, not ad hoc | Pass | Reuses `composerImeSubmitFlushEnabled$` rather than reaching into the switch registry |

### Findings

No P0 or P1 issues.

Two observations, neither blocking:

- Only the `contentDOM` guard is load-bearing for the reported bug. `ignoreMutation` returns `true` for any mutation whose target is not inside `contentDOM`, so the `dom`, `noteDom`, `placeholderDom` and `quoteText` writes were already being ignored by ProseMirror. Guarding them too is consistent and removes pointless DOM churn, so it is worth keeping — it just is not part of the fix.
- `runtime.skipUnchangedNoteWrites` is captured once per mount, so flipping the switch in Lab takes effect on the next composer mount rather than immediately. This matches the existing `includeQuoteOnlyFeedback` read directly above it, so it is consistent rather than surprising, but it is worth knowing during device verification: toggle the switch, then reload before testing.

### Testing

- Coverage: Adequate for a `fix:` commit. The new test is a real regression guard — the PR reports it failing both when the value guards are reverted and when the switch is set to `false`, which also proves the gating works in both directions.
- Conventions: Compliant. Integration-level through `detachedSetupPage`, switch injected via the documented `featureSwitches` option, no internal mocks, no fake timers, deterministic record draining via `takeRecords()`.
- Reported local runs: `chat-inline-feedback.test.tsx` 31 passed, `chat-run-queue.test.tsx` 33 passed, `@okouai/app` and `@okouai/core` `check-types` clean, `pnpm format` clean, `pnpm turbo run lint` 13 tasks 0 errors.
- Known gap, declared in the PR: the `forceFlush` half has no test, because happy-dom never leaves a deferred flush pending to force.
- The three CI failures on this head (`cli-e2e-02-playwright` on a billing scheduled-intent case with an accompanying Clerk FAPI failure, `cli-e2e-03-runner` on `digest-mismatch`, and the `ci-gate-turbo` aggregate) are in billing and runner surfaces this diff does not touch; they were classified as flaky/infra and rerun.
- Device verification remains the real acceptance gate: with the switch on, reproduce two consecutive Chinese compositions in an inline feedback note and confirm the whole note is submitted.

PR #28031, opened 2026-08-18
feat(api): ingest presentation template uploads through a server-owned import

bingjiez666:
First of three PRs splitting the presentation template import work (原 #26947). This one covers ingestion only; analysis and package publication follow.

## Shape

Uploads are scoped to a server-issued import — the shape Box, Dropbox, and S3 multipart use.

```
POST /presentation-templates/imports              -> templateId
POST /presentation-templates/:id/uploads          -> one slot (source | page N) + presigned PUT
PUT  <presigned url>                              -> bytes go straight to R2
POST /presentation-templates/:id/commit           -> no body, no object ids
```

## Why not a client-supplied manifest

The first draft of this PR took a manifest: the browser named one PPTX upload id plus an ordered list of page ids. That grouping only ever existed in a browser variable, held across rendering and a long upload sequence. A stale client could pair one deck's source with another deck's pages — both halves real render output, both resolving fine — and the mismatch is silent. The analysis run reads page N against slide N, so every extracted conclusion would be misaligned.

With a server-issued import, commit takes no object ids, so a caller has no way to express a wrong pairing.

Three things fall out:

- **Fewer S3 round trips.** The manifest version resolved up to 101 uploads through `ListObjects` + `HeadObject` — roughly 200 round trips inside one request. Commit now reads its own rows.
- **Abandoned uploads are attributable.** They hang off the import that owns them instead of being orphans nothing points at.
- **Retry is a slot replacement**, not surgery on a client-held array.

## What still reads the file

One trailing byte range, to count `ppt/slides/slideN.xml` in the ZIP central directory.

The browser rendered these pages from a deck it opened, so the archive is not re-validated for format, encryption, or macros — that would re-derive facts a successful render already establishes. What an import cannot show is whether the browser exported *every* page, so the count survives as a completeness guard. The whole file is never downloaded.

## Schema

- New `presentation_template_uploads`: staging rows for one in-progress import (`template_id`, `role`, `page_index`, `storage_key`, size). Partial unique indexes keep one source and one row per page index.
- `presentation_templates.source_storage_key` becomes nullable — an import exists before its deck is uploaded. Commit freezes the ordered result onto the template row and drops the staging rows, so committed templates stay readable without a join.
- Deliberately not `run_uploaded_files`: that table's catalog trigger would project every slide crop into the user's artifact catalog (per #26235).

## Deviations from #26576 worth flagging

[...]

PR #28172, opened 2026-08-19
feat(api): analyse a template import in a chat thread the owner can watch

bingjiez666:
Stacked on #28031. Review that first — this diff is only the analysis half.

## The change

Commit used to hand the import to a headless run and wire a callback back. Two problems with that:

- The analysis was invisible. The user got a spinner and no way to see what the model was doing, interrupt it, or follow up.
- The run's authority to read the deck came from data the launcher put into the run. The launcher was us, so it worked — but the shape was "trust what's attached to the run".

Analysis now happens the way the owner would start it: a chat thread with a message in it.

```
commit
  ├─ freeze the verified upload set onto the template row   (transaction)
  └─ open the analysis thread, record it, send one message
```

The thread is recorded against the import **before** anything is sent, and that mapping is the run's whole authority. A run reaches an import only because the thread it runs in belongs to that import — `agent_runs.chat_thread_id` → `presentation_template_import_threads` → template. Nothing the caller supplies is trusted.

The callback service, its dispatch case, and its run-callback kind are all deleted.

## Two halves, separately recoverable

The halves fail independently, so the template's own columns say which one still has to run:

| state | `status` | `source_storage_key` |
|---|---|---|
| still collecting uploads | `pending` | null |
| frozen, not yet analysed | `pending` | set |
| analysing | `processing` | set |

A commit repeated after a crash resumes at the right half instead of redoing the freeze or leaving the import stranded. Only a *started* analysis moves the import out of `pending`.

## The prompt

The message the owner sees is one line naming the template and the skill. The extraction contract moved to `presentation-template-import` in vm0-ai/vm0-skills#325, mounted next to `goal` at run creation.

That keeps the thread readable — the user sees a task, not a 45-line specification — and lets the contract be revised without an API deploy.

**Merge order:** vm0-ai/vm0-skills#325 must merge and sync before this one, since this PR adds the skill name to the mount list.

## Also here

- `source` / `pages` / `fail` endpoints, run-scoped through the thread mapping, behind new `presentation-template:read` / `:write` capabilities.
- The matching CLI commands. `pages pull` stages into a scratch directory and renames only once every page has downloaded, so an interrupted pull leaves no partial page set.
- `commit` now returns `chatThreadId` so the caller can open the thread.

## Tests

Three new cases in the existing integration suite (13 total, all passing):

- commit hands back a thread, a run in that thread reads the source and the ordered pages, and a repeat commit resolves to the same thread
- a run in **another** thread — same owner, same org — gets 404 from all three endpoints
- the run reports a failure and the owner's detail view shows it

`run-lifecycle`, `cron-sync-skills`, `chat-events`, and `generation-template-prompt` all pass; four pre-existing `generate image`/`generate website` CLI failures reproduce with this branch stashed and are unrelated.

### DIFF
--- a/turbo/apps/api/src/lib/template-import-prompt.ts
+++ b/turbo/apps/api/src/lib/template-import-prompt.ts
@@ -1,0 +1,13 @@
+import { PRESENTATION_TEMPLATE_IMPORT_SKILL_NAME } from "@okouai/core/seed-skills";
+
+/**
+ * The message the owner sees in the analysis thread.
+ *
+ * It stays short on purpose. The extraction contract lives in the
+ * `presentation-template-import` skill, so it can be revised in the skills
+ * repository without an API deploy, and the thread reads as a task rather than
+ * as a specification.
+ */
+export function templateImportPrompt(templateId: string): string {
+  return `Import presentation template ${templateId}. Follow the ${PRESENTATION_TEMPLATE_IMPORT_SKILL_NAME} skill exactly.`;
+}
--- a/turbo/apps/api/src/signals/routes/__tests__/presentation-templates.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/presentation-templates.test.ts
@@ -12,6 +12,7 @@ import {
   PRESENTATION_TEMPLATE_SOURCE_CONTENT_TYPE,
   presentationTemplatesContract,
 } from "@okouai/api-contracts/contracts/presentation-templates";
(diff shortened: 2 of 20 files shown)


bingjiez666:
Closing in favour of #28252. The flow changes shape: the deck arrives as an ordinary chat attachment instead of through a bespoke import protocol, so the import endpoints, the staging table, and the thread mapping in this PR are all removed rather than reworked. None of this was merged, so nothing has to be reverted.

Issue #8609, opened 2026-04-09
fix: remove blanket max-lines-per-function override for tsx files


e7h4n:
## Summary

`turbo/apps/platform/.oxlintrc.json` has a blanket `max-lines-per-function: allow` override applied to **all `**/*.tsx` files**. This masks oversized component functions (some 200–1000+ lines), bypassing the 128-line limit enforced everywhere else.

## File

- `turbo/apps/platform/.oxlintrc.json` — lines 104-108

```json
{
  "files": ["**/*.tsx"],
  "rules": {
    "max-lines-per-function": ["allow"]
  }
}
```

## Known Large Components

- `ScheduleSettingsForm()` — 275 lines
- `ScheduleDetailView()` — 237 lines
- `ZeroScheduleDetailPage()` — 134 lines
- `network-insights-page.tsx` — ~1068 lines total

## Acceptance Criteria

- [ ] Remove the blanket `**/*.tsx` override from `.oxlintrc.json`
- [ ] Fix or split component functions that exceed 128 lines
- [ ] `pnpm turbo run lint` passes with no new warnings

## Parent Issue

Part of #8441 (BS-17)

PR #28173, opened 2026-08-19
feat(platform): make pinned agent reordering discoverable

Lunarivibe:
## Problem

The pinned agent grid in the three-column nav has been draggable for a while, but nothing says so. Reviewing `zero-sidebar-pinned.tsx` turned up five issues:

1. **No cue at rest.** The tile is a `<Link>`; the only hint was a native `title` tooltip after a ~1s hover delay.
2. **Ambiguous drop feedback.** A `ring-2` around the target card reads as "select this", not "insert here" — it never said which side the agent lands on.
3. **Weak drag feedback.** The dragged tile only faded to 40% and stayed in its slot, so nothing felt picked up.
4. **No cursor affordance.** Default pointer throughout.
5. **Uniform tiles.** The lead agent is fixed but looked identical to the reorderable ones.

## Change

| | |
|---|---|
| Drag handle | 17×11px bar, `absolute · left-1/2 · top -5px`, fades in on hover/focus. Zero layout cost — tile stays 68px, padding stays 6px, the avatar never moves. |
| Cursor | `cursor-grab` / `active:cursor-grabbing` on reorderable tiles only |
| Drop indicator | Insertion caret on the target card's leading or trailing edge, replacing the ring |
| Dragged slot | Collapses to a dashed placeholder instead of fading |
| Grid | `gap-1` → `gap-x-1 gap-y-2` so the handle never meets the row above |

Which edge the caret sits on is derived from the existing behaviour of `movePinnedAgent$`, which splices the dragged agent into the target's index: a backwards drag lands *before* the target, a forwards drag lands *after* it. No new signal, no change to the reorder command or the persisted order.

The handle is a cue, not a hit target — the whole tile stays draggable.

The section header is unchanged. An earlier revision added a "Drag to reorder" line beside the label; it was cut as a redundant second cue, which leaves `sidebar.dragToReorder` unreferenced, so the key is deleted from all ten locales.

## Deliberate omissions

- **The Pin tile stays at slot 5.** The design review floated dropping it out mid-drag so the sequence reads contiguous, but the caret binds to a specific card's edge rather than to a gap, so it is no longer load-bearing. Removing it would also churn `pinned-agents-grid`'s `childElementCount`, which feeds the `pinnedAgentGridRows` skeleton cache.
- **No keyboard reorder.** The handle appears on `focus-visible`, but arrow-key reordering is a separate surface (key handling plus live-region announcements) and is not in this PR.

## Scope

Horizontal (three-column nav) layout only. The vertical sidebar list has never supported reordering and is untouched.

## Tests

Two integration tests added to `zero-sidebar.test.tsx`:
- handle renders on reorderable tiles and is absent on the lead agent
- caret renders on the trailing edge for a forwards drag and the leading edge for a backwards drag, and clears on `dragEnd`

Local: `zero-sidebar.test.tsx` 74/74 pass · `tsc --noEmit` clean · `oxlint`, `oxlint --type-aware`, `eslint --max-warnings 0` clean on changed files · `pnpm format` applied · `i18n:check` 100% across all locales.

## Verified on the preview

Walked through `https://pr-28173-app.omby.ai` with `threeColumnNav` on and six pinned agents. Measured in the live build: tile 53×68 unchanged, `row-gap 8px` / `column-gap 4px`, handle 17×13px (11px + 2×1px border) at `top: -5px`, lead agent `draggable=false` with `cursor: pointer` and no handle, forwards drag → caret `right: -3px`, backwards drag → caret `left: -3px`, dashed placeholder on the dragged slot, order persists across reload, and the handle stays legible in dark mode.

### DIFF
--- a/turbo/apps/platform/src/i18n/locales/de-DE/agents.json
+++ b/turbo/apps/platform/src/i18n/locales/de-DE/agents.json
@@ -607,7 +607,6 @@
     "chatResult": "Chat",
     "clearSearch": "Suche löschen",
     "descriptionWithChats": "Wählen Sie einen Agenten aus oder springen Sie zu einem Chat.",
-    "dragToReorder": "Zum Neuanordnen ziehen",
     "escapeShortcut": "Esc",
     "filterAll": "Alle",
     "leadDescription": "Ihr leitender Assistent, immer für Sie da",
--- a/turbo/apps/platform/src/i18n/locales/en-US/agents.json
+++ b/turbo/apps/platform/src/i18n/locales/en-US/agents.json
@@ -607,7 +607,6 @@
     "chatResult": "Chat",
     "clearSearch": "Clear search",
     "descriptionWithChats": "Pick an agent or jump to a chat.",
-    "dragToReorder": "Drag to reorder",
     "escapeShortcut": "Esc",
     "filterAll": "All",
     "leadDescription": "Your lead assistant, always here for you",
--- a/turbo/apps/platform/src/i18n/locales/es-ES/agents.json
+++ b/turbo/apps/platform/src/i18n/locales/es-ES/agents.json
@@ -607,7 +607,6 @@
     "chatResult": "Conversación",
     "clearSearch": "Borrar búsqueda",
     "descriptionWithChats": "Elige un agente o salta a un chat.",
-    "dragToReorder": "Arrastra para reordenar",
     "escapeShortcut": "Esc",
     "filterAll": "Todos",
     "leadDescription": "Tu asistente principal, siempre aquí para ti",
--- a/turbo/apps/platform/src/i18n/locales/fr-FR/agents.json
+++ b/turbo/apps/platform/src/i18n/locales/fr-FR/agents.json
@@ -607,7 +607,6 @@
     "chatResult": "Chat",
     "clearSearch": "Effacer la recherche",
     "descriptionWithChats": "Choisissez un agent ou passez à une discussion.",
-    "dragToReorder": "Glissez pour réorganiser",
     "escapeShortcut": "Esc",
     "filterAll": "Tout",
     "leadDescription": "Votre assistant principal, toujours là pour vous",
--- a/turbo/apps/platform/src/i18n/locales/hi-IN/agents.json
+++ b/turbo/apps/platform/src/i18n/locales/hi-IN/agents.json
@@ -607,7 +607,6 @@
     "chatResult": "चैट",
     "clearSearch": "खोज साफ़ करें",
     "descriptionWithChats": "एक एजेंट चुनें या सीधे चैट पर जाएँ।",
-    "dragToReorder": "क्रम बदलने के लिए खींचें",
     "escapeShortcut": "Esc",
     "filterAll": "सभी",
     "leadDescription": "आपका प्रमुख सहायक, हमेशा आपके लिए यहाँ",
--- a/turbo/apps/platform/src/i18n/locales/id-ID/agents.json
+++ b/turbo/apps/platform/src/i18n/locales/id-ID/agents.json
@@ -607,7 +607,6 @@
     "chatResult": "Chat",
     "clearSearch": "Hapus pencarian",
     "descriptionWithChats": "Pilih agen atau langsung ke chat.",
-    "dragToReorder": "Seret untuk mengurutkan ulang",
     "escapeShortcut": "Esc",
     "filterAll": "Semua",
     "leadDescription": "Asisten utama Anda, selalu siap membantu",
--- a/turbo/apps/platform/src/i18n/locales/it-IT/agents.json
+++ b/turbo/apps/platform/src/i18n/locales/it-IT/agents.json
@@ -607,7 +607,6 @@
     "chatResult": "Chat",
     "clearSearch": "Cancella ricerca",
     "descriptionWithChats": "Scegli un agente o passa a una chat.",
-    "dragToReorder": "Trascina per riordinare",
     "escapeShortcut": "Esc",
     "filterAll": "Tutti",
     "leadDescription": "Il tuo assistente principale, sempre qui per te",
--- a/turbo/apps/platform/src/i18n/locales/ja-JP/agents.json
+++ b/turbo/apps/platform/src/i18n/locales/ja-JP/agents.json
@@ -607,7 +607,6 @@
     "chatResult": "チャット",
     "clearSearch": "検索をクリア",
     "descriptionWithChats": "エージェントを選択するか、チャットにジャンプします。",
-    "dragToReorder": "ドラッグして並べ替え",
     "escapeShortcut": "Esc",
     "filterAll": "すべて",
     "leadDescription": "あなたの主任アシスタントがいつでもあなたのためにここにいます",
--- a/turbo/apps/platform/src/i18n/locales/ko-KR/agents.json
+++ b/turbo/apps/platform/src/i18n/locales/ko-KR/agents.json
@@ -607,7 +607,6 @@
     "chatResult": "채팅",
     "clearSearch": "검색어 지우기",
     "descriptionWithChats": "에이전트를 선택하거나 채팅으로 바로 이동하세요.",
-    "dragToReorder": "드래그하여 순서 변경",
     "escapeShortcut": "Esc",
     "filterAll": "전체",
     "leadDescription": "항상 당신을 돕는 리드 어시스턴트",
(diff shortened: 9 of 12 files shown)


Lunarivibe:
LGTM


PR_AUTO_REVIEW_COMMIT: 951885055907cd665398ca998ac6fb759cfc0cb5
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Makes the pinned-agent grid's existing drag-to-reorder visibly discoverable: a whole-pixel drag handle fades in above a reorderable tile on hover/focus, the tile takes a `grab` cursor, the drop indicator becomes an insertion caret on the target's leading or trailing edge instead of an ambiguous ring, and the dragged tile's slot collapses to a dashed placeholder. The caret side is derived from `movePinnedAgent$`'s existing splice semantics, so no signal, command, or persisted order changes. Scoped to the `ThreeColumnNav` horizontal layout; the now-unreferenced `sidebar.dragToReorder` key is deleted from all ten locales.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | `feat:` commits must ship integration tests | Pass | Two new integration tests in `zero-sidebar.test.tsx` cover handle presence/absence and caret side in both drag directions. |
| `REVIEW.md` | Feature regression guard — replaced behavior needs regression coverage | Pass | The ring→caret replacement is covered by the new caret test; pre-existing "reorders pinned agents with drag and drop" and "leaves the lead agent in place when it is dragged" still cover the reorder flow most likely to regress, and both still pass. |
| `REVIEW.md` | New user-facing behavior gated by a `FeatureSwitchKey` | Pass | Only reachable through `FeatureSwitchKey.ThreeColumnNav` (`feature-switch.ts:444`, `enabled: false`, `enabledEmailHashes: ["54757055"]`). |
| `REVIEW.md` | Correctness / logic | Pass | `resolveDropSide` compares indices in `displayedPinnedAgents`; only pinned, non-lead agents can be dragged or targeted, and the lead agent occupies index 0 of both the rendered and the filtered list, so the sign of `from - to` matches `movePinnedAgent$`'s splice. `from === to` is guarded. |
| `REVIEW.md` | Type safety — no `any`, no broad types | Pass | `PinnedDropSide` is a literal union; `resolveDropSide` params and return are fully typed; no assertions added. |
| `REVIEW.md` | Security | N/A | Client-side rendering and cursor affordances only; no auth, endpoint, input parsing, or credential surface touched. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No `turbo/packages/db/src/schema/**`, contract, or `jsonb(...).$type<...>()` change; the persisted `pinnedAgentIds` shape is untouched. |
| `docs/bad-smell.md` | Zero `any`, zero lint/type suppressions, no dynamic `import()`, no hardcoded URLs | Pass | Diff adds none. The file-level `oxlint-disable max-lines-per-function` is pre-existing (`TODO(#8609)`), not introduced or widened by this PR. |
| `docs/bad-smell.md` | Fail fast — no fallback/recovery slop | Pass | No `??`/`||` chain added. The diff removes code paths (the `title` tooltip, the drop-target ring) rather than adding tolerance branches. |
| `docs/fallback.md` | Every fallback introduced by the diff is declared in the PR summary | N/A | The diff introduces no fallback, so no `Fallbacks` section is required. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | No test asserts the removed `title` tooltip or the removed ring is still absent. |
| `docs/fallback.md` | A non-GA feature switch needs no compatibility code | Pass | `ThreeColumnNav` is `enabled: false` and email-hash gated, so §2 applies: no dual-read, migration, or rollout branch is owed, and none was added. |
| `docs/deployment-compatibility.md` | Independently deployed boundaries stay compatible | N/A | No frontend/backend contract, runner protocol, queue payload, schema, or persisted-state change; behavior is confined to client rendering behind a non-GA switch. |
| `docs/effect.md` | Render stays pure — no Store writes, I/O, or signal creation in render | Pass | `resolveDropSide` is a module-scope pure function over already-read signal values; no `state`/`computed`/`command` is created during render. |
| `docs/effect.md` | User event handled by a semantic command via `detach(..., Reason.DomCallback)` | Pass | `onDrop` still routes through `movePinnedAgent$` with `pageSignal` and `detach(..., Reason.DomCallback)`; the diff does not add Store orchestration to the view. |
| `docs/cache.md` | No duplicate mutable sources of truth; no manual cache duplicating `computed` memoization | Pass | Drop state stays in the existing `zero-sidebar-state.ts` signals; the caret side is derived on read, not copied into a second store. |
| `.claude/skills/ccstate/SKILL.md` | DOM callbacks use `useGet(pageSignal$)` + `detach()`, never `void` | Pass | `zero-sidebar-pinned.tsx:onDrop` preserves the documented pattern; no new DOM callback bypasses it. |
| `docs/react-commit.md` | Remove duplicate subscriptions; keep subscriptions close to their consumers | **Finding (P1)** | `zero-sidebar-pinned.tsx:485-486` adds the same two signal subscriptions the card already holds at `:307-308`, then passes the derived value down as a prop. |
| `docs/testing.md` | Integration tests only; mock at the boundary; real infrastructure | Pass | Both new tests render the real page through `setupSidebarPage` and drive real DOM events; no `vi.mock()` of internals, no fetch stubbing, no fake timers. |
| `docs/testing/patterns.md` | Platform page tests follow the page-test structure | Pass | Tests use the established `setupSidebarPage` + `findByTestId` + `waitFor` shape already used by the neighbouring pinned-grid tests. |
| `docs/testing/app-testing.md` | Page tests exercise the page, with mock infrastructure at the network boundary | Pass | State is seeded via `context.mocks.data.userPreferences` and the feature switch, then asserted through the rendered grid. |
| `docs/testing/testing-external-behavior.md` | Do not assert on CSS classes | **Finding (P1)** | `zero-sidebar.test.tsx:3863` and `:3888` assert `.className` contains `-right-` / `-left-`. |
| `docs/testing/anti-patterns.md` | AP-7 — no testing of implementation details (CSS classes) | **Finding (P1)** | Same two assertions; the anti-pattern checklist lists "Testing UI text content or CSS classes" explicitly. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

- `turbo/apps/platform/src/views/zero-page/__tests__/zero-sidebar.test.tsx:3863` and `:3888` — the caret-side assertions read `.className` and match the Tailwind fragments `-right-` / `-left-`. `docs/testing/testing-external-behavior.md` states "Do not assert on … CSS classes", and `docs/testing/anti-patterns.md` AP-7 lists CSS-class assertions in its checklist. The test freezes a styling detail: rewriting `-right-[3px]` as `right-[-3px]`, moving the offset to an inline style, or switching to a pseudo-element would fail CI while the product still works. Suggested fix: give the caret a semantic `data-drop-side={dropSide}` attribute and assert on that, or assert the observable outcome by completing the drop and checking the resulting order for both directions.

- `turbo/apps/platform/src/views/zero-page/zero-sidebar-pinned.tsx:485-486` — `PinnedAgentListSection` now subscribes to `draggingPinnedAgentId$` and `pinnedAgentDropTargetId$` and passes the derived `dropSide` down, while `PinnedAgentGridCard` still subscribes to both at `:307-308`. That is a duplicate subscription (`docs/react-commit.md`, "Remove Duplicate Subscriptions") with the subscription hoisted away from its consumer ("Keep Subscriptions Close to Their Consumers"). `setPinnedAgentDropTarget$` fires on every `dragover` tick, so the whole section — label, Pin tile, context decorators, and both dialog containers — now re-renders alongside the cards, where previously only the cards did. Suggested fix: expose the resolved target and side as one small `computed` in `zero-sidebar-state.ts` (for example `pinnedAgentDropSide$` returning `{ targetId, side } | null`) and subscribe to it from the card only, which also matches `docs/effect.md`'s "derived value: use `computed`". Noted as a practice deviation rather than a measured regression; the grid is small and no profile was taken.

### Testing

- Coverage: Adequate. The `feat:` change ships two integration tests covering the new affordance (handle shown on reorderable tiles, absent on the lead agent; caret on the trailing edge for a forwards drag, leading edge for a backwards drag, cleared on `dragEnd`). The behavior this PR replaces stays covered by the existing `reorders pinned agents with drag and drop` and `leaves the lead agent in place when it is dragged` tests.
- Conventions: One violation — the CSS-class assertions above. Otherwise compliant: integration-level, real page render, no internal `vi.mock()`, no fetch stubbing, no fake timers, no mock-call assertions, and the feature stays behind `FeatureSwitchKey.ThreeColumnNav`.

PR #28178, opened 2026-08-19
refactor(deploy): drop the legacy variable source fallback

lancy:
## What this changes

`.github/actions/web-api-env/action.yml` read each branded variable family through `first_non_empty(OKOU_${suffix}, ZERO_${suffix})` and then emitted only the canonical `OKOU_${suffix}` key. That fallback is inert — the canonical variable exists at every level where its legacy counterpart exists — so it always selected the canonical value.

- Removed the `add_okou_var` helper.
- `add_okou_var HOST_DOMAIN` / `HOST_SCHEME` → `add_var OKOU_HOST_DOMAIN OKOU_HOST_DOMAIN` / `add_var OKOU_HOST_SCHEME OKOU_HOST_SCHEME`.
- The `okou_price_var_suffixes` loop now calls `add_var "OKOU_${suffix}" "OKOU_${suffix}"`.

`.github/scripts/tests/web-api-env-action-test.sh`:

- The **variable** fixtures now supply canonical `OKOU_*` names instead of `ZERO_*`. The two fixture values that were still `zero-`-branded became neutral (`sites.test`, `test-campaign`), with their two assertions updated to match.
- The `empty` branded-config case strips `OKOU_*` from the variable fixture (it previously stripped `ZERO_*`), so it still renders empty values.
- Added a guard, scoped to variable reads only, that fails if the action ever reads a `ZERO_` **variable** again. It deliberately does not match `repo_secret`.

## What this explicitly does NOT change

- **`add_okou_secret` and all four of its call sites are untouched.** Verified byte-identical against `main`; only their line numbers shift by 7. The secret half is still gated — canonical `OKOU_*` secrets were created on 2026-08-19 and need more release observations — and must not be removed in the same PR as the variable half.
- Every secret-related fallback assertion in the test is preserved verbatim, and `assert_migrated_zero_outputs_absent` still runs against all five rendered fixtures.
- The 14 legacy `ZERO_*` repository/`production` variables are **not** deleted. That is a separate configuration step after this ships and one release is observed.
- `turbo/apps/api/src/lib/env.ts` and its 22 application-level fallbacks are untouched.
- No rendered key name, value, or ordering changes.

## Verification

`.github/scripts/tests/web-api-env-action-test.sh` passes locally, and `bash -n` is clean.

Rendered output was compared before/after by extracting both versions of the action script and feeding them the **real** repository- and environment-scoped variable/secret *names* (synthetic values), for every app/environment combination:

| combination | keys | result |
|---|---|---|
[...]

Issue #26701, opened 2026-08-12
refactor(protocol): drain and retire Zero protocol compatibility (EPIC)


lancy:
## Outcome

Retire the legacy Zero protocol compatibility surface only after canonical Okou writers are live, every supported legacy producer and consumer has migrated, immutable work has drained, and production evidence proves removal safe.

Canonical protocol surfaces are:

- `/api/okou/**` for active branded HTTP routes;
- `OKOU_*` for runtime and deployment configuration; and
- run tokens carrying `scope: "okou"`.

The legacy compatibility surface to retire eventually is:

- `/api/zero/**` route aliases;
- `ZERO_*` runtime and deployment readers/source fallbacks;
- run-token verification for `scope: "zero"`; and
- provider-facing callback producers and registrations that still use `/api/zero/**`.

This epic does not authorize mechanical renaming of database objects, internal TypeScript symbols, service filenames, historical migrations, immutable artifacts, or unrelated uses of the word `zero`.

## Starting state

The Phase A/Phase B migration in #26487 is complete and live:

- Phase A compatibility shipped at `2eba4f48e1b1bbcce818f4b8a6c8f717c28f7006`.
- Phase B canonical Okou client cutover shipped at `d2fddbf6714b35182e1ef4b787adf84724f6ee02`.
- Current first-party Platform, CLI, and Desktop protocol callers use `/api/okou/**` and prefer `OKOU_*`.
- `/api/zero/**`, Zero-scope token verification, and `ZERO_*` fallback readers remain functional for historical clients, contexts, callbacks, and rollback.

The first writer-stop slices shipped in production release #26694:

- #26652 via #26684 stops new execution contexts from emitting Zero runtime credentials.
- #26670 via #26676 stops new Web/API deployment environments from writing migrated Zero aliases.
- #26694 merged as `cad3b93a80aef7e178daec30388058790eff1cf7` and published API `v1.437.0`; the exact-SHA Turbo and release-please workflows completed successfully.
- Bounded post-release traffic and deployment-output checks passed. The remaining ordinary-agent and presentation-template context-key proof is tracked by #26721.

Related migration state:

- CLI identity migration #26429 is complete.
- Desktop identity, hard-stop, and retirement work remains tracked by #26364, #26649, #26370, and #26368.

## Safety principles

- Stop writers and producers before removing readers and acceptors.
- Preserve old/new API, Runner, CLI, Desktop, queued-context, and rollback combinations while independently deployed versions overlap.
[...]

PR #28238, opened 2026-08-19
refactor(deploy): drop the legacy secret source fallback

### LINKED ISSUE #28237: refactor(deploy): drop the legacy secret source fallback
## Background

Parent: #26701. Gate lifted for this step on 2026-08-19: https://github.com/vm0-ai/vm0/issues/26701#issuecomment-5342453539

This is the secret-source counterpart to the merged #28169, which removed the variable-source fallback. Measured on `main@7d689f580d7478b41538f1b25280f2dc49a516df`.

`.github/actions/web-api-env/action.yml:126` still reads the canonical secret and falls back to the legacy one:

```bash
add_okou_secret() {
  local suffix="$1"
  local value
  value="$(first_non_empty "$(repo_secret "OKOU_${suffix}")" "$(repo_secret "ZERO_${suffix}")")"
  add_env "OKOU_${suffix}" "$value"
}
```

The fallback is inert. All eight `OKOU_*` secrets exist at every level where their `ZERO_*` counterpart exists, verified across repository scope and the `production` environment, so `first_non_empty` always selects the canonical value.

Call sites: the `okou_provider_secret_suffixes` loop at `:358` covering `MAPS_GOOGLE_MAPS_TOKEN`, `WEATHER_GOOGLE_WEATHER_TOKEN`, `SCRAPE_FIRECRAWL_TOKEN`, `WEB_SEARCH_PERPLEXITY_TOKEN` and `FINANCE_APIDOJO_TOKEN`, plus the API-only block at `:361-363` covering `SEO_DATAFORSEO_LOGIN`, `SEO_DATAFORSEO_PASSWORD` and `BROWSER_USE_API_KEY`.

## Design

Remove the `add_okou_secret` helper and rewrite its call sites to read the canonical secret directly, keeping the rendered output byte-identical. Follow whatever form #28169 used for the variable side so the two halves read the same.

## Trap

**Do not delete the eight legacy `ZERO_*` secrets from GitHub in this change.** That is a deliberately separate step, authorized only after this ships and the eight capabilities are re-verified against a new release.

The reason matters: secret **values cannot be read** through the API, so unlike the variable half there was no way to confirm each canonical secret is byte-identical to its legacy counterpart. If one of the eight canonical values was entered incorrectly, this change is what surfaces it — and while the legacy secrets still exist, recovery is a one-line revert. Deleting them in the same change removes that safety net.

`.github/scripts/tests/web-api-env-action-test.sh` contains 30 lines referencing `ZERO_`, and after #28169 the remaining ones are the secret fixtures. Some deliberately supply **only** a `ZERO_*` secret and assert the rendered output still contains the canonical key. Those test the fallback itself and must be rewritten to supply canonical secrets. Keep `assert_migrated_zero_outputs_absent` working, and keep `assert_no_fixture_secret_values` intact — no test change may cause a secret value to be printed.

## Explicit non-goals

- Deleting any `ZERO_*` secret from repository scope or the `production` environment.
- `turbo/apps/api/src/lib/env.ts` and its 22 `OKOU_ENV_FALLBACKS` entries, still gated on the local-development migration.
- `add_var`, `add_secret`, `add_oauth_client_config_key` and every non-`okou` helper.
- Any rendered key name, value, or ordering.
- Any Doppler-sourced value.

## Acceptance criteria

- [ ] `add_okou_secret` no longer exists and no `ZERO_` secret name is read anywhere in the action.
- [ ] All eight families still render under their canonical key names with unchanged values.
- [ ] `.github/scripts/tests/web-api-env-action-test.sh` passes, with secret fixtures updated to canonical names.
- [ ] `assert_migrated_zero_outputs_absent` and `assert_no_fixture_secret_values` still work, and no test prints a secret value.
- [ ] `git diff --stat` shows no change outside the action and its test script.
- [ ] The eight `ZERO_*` secrets still exist in GitHub after this merges.
- [ ] PR pipeline is green.

## Follow-up

Once this ships in a release, the eight capabilities are re-verified and the eight legacy `ZERO_*` secrets are deleted. That is tracked separately and is not part of this issue.

Parent: #26701

lancy:
## What this changes

`.github/actions/web-api-env/action.yml` read each branded secret family through `first_non_empty(OKOU_${suffix}, ZERO_${suffix})` and then emitted only the canonical `OKOU_${suffix}` key. That fallback is inert — every canonical secret exists at each level where its legacy counterpart exists — so it always selected the canonical value.

- Removed the `add_okou_secret` helper.
- The `okou_provider_secret_suffixes` loop now calls `add_secret "OKOU_${suffix}" "OKOU_${suffix}"`.
- The API-only block now calls `add_secret OKOU_SEO_DATAFORSEO_LOGIN OKOU_SEO_DATAFORSEO_LOGIN`, and the same shape for `OKOU_SEO_DATAFORSEO_PASSWORD` and `OKOU_BROWSER_USE_API_KEY`.

This is the secret-source counterpart to the merged #28178, and deliberately uses the same form so the two halves read alike.

`.github/scripts/tests/web-api-env-action-test.sh`:

- The **secret** fixture now supplies canonical `OKOU_*` names instead of `ZERO_*`. Only the fixture key branding changed; every fixture value and every rendered-value assertion is unchanged.
- The `empty` branded-config case strips `OKOU_*` from the secret fixture (it previously stripped `ZERO_*`), so those families still render as empty values.
- The existing `ZERO_`-read guard was scoped to variable reads with a comment explaining that the secret half still fell back. It now covers `repo_secret` / `add_secret` as well, so nothing in the action may name a legacy `ZERO_` source.
- The `both` fixture and the `precedence_*` block only existed to pin `OKOU_` over `ZERO_` precedence for secrets. With no fallback there is no precedence left to test, so both are removed.
- `assert_migrated_zero_outputs_absent` is unchanged and still runs against all four remaining rendered fixtures. `assert_no_fixture_secret_values` is unchanged apart from dropping the three `okou-*` literals that belonged to the deleted fixture; every remaining fixture secret value is still covered by its `github-` or `doppler-` prefix, so the guard's coverage is complete and no test prints a secret value.
- `branded_config`'s default was renamed `legacy` → `canonical`, since after #28178 and this change the default fixture is entirely canonical.

## No secrets are deleted, and why this is two steps

**The eight legacy `ZERO_*` secrets are not deleted by this PR**, at either repository scope or in the `production` environment. Verified still present after this change:

| level | legacy secrets still present |
|---|---|
| repository | `ZERO_BROWSER_USE_API_KEY`, `ZERO_FINANCE_APIDOJO_TOKEN`, `ZERO_SCRAPE_FIRECRAWL_TOKEN`, `ZERO_WEATHER_GOOGLE_WEATHER_TOKEN`, `ZERO_WEB_SEARCH_PERPLEXITY_TOKEN` |
| `production` | `ZERO_MAPS_GOOGLE_MAPS_TOKEN`, `ZERO_SCRAPE_FIRECRAWL_TOKEN`, `ZERO_SEO_DATAFORSEO_LOGIN`, `ZERO_SEO_DATAFORSEO_PASSWORD`, `ZERO_WEB_SEARCH_PERPLEXITY_TOKEN` |

The variable half could be finished in one step because variable values are readable through the API, so each legacy variable was confirmed byte-identical to its canonical counterpart before deletion. **Secret values cannot be read**, so that check is impossible here. The available evidence is name-and-level alignment plus the deployment smoke — enough to remove the fallback, not enough to also throw away the source.

If any of the eight canonical secret values was pasted incorrectly, this PR is exactly what surfaces it. While the legacy secrets still exist, recovery is a one-line revert. Deleting them in the same change would remove that safety net.

Deleting the eight legacy secrets is authorized only after this ships in a release and the eight capabilities are re-verified, per the gate decision in https://github.com/vm0-ai/vm0/issues/26701#issuecomment-5342453539. It is tracked separately.

## Verification

`.github/scripts/tests/web-api-env-action-test.sh` passes locally and `bash -n` is clean.

The premise was re-confirmed against live GitHub configuration: for all eight families, every level holding a `ZERO_*` secret also holds the matching `OKOU_*` secret. `promote-api-production` runs with `environment: production`; the preview jobs in `turbo.yml` set no job-level `environment:`, so preview resolves repository scope only. Both resolution scopes were checked.

Rendered output was compared before/after by extracting both versions of the action script and feeding them the **real** repository- and environment-scoped variable/secret *names* (synthetic values), for every app/environment combination:

| combination | keys | result |
|---|---|---|
| api / production | 210 | identical |
| web / production | 196 | identical |
| api / preview | 211 | identical |
| web / preview | 196 | identical |

No `ZERO_`-prefixed key is emitted in any combination, and in production all eight families resolve from their canonical secret. The three that render empty in preview (`OKOU_MAPS_GOOGLE_MAPS_TOKEN`, `OKOU_SEO_DATAFORSEO_LOGIN`, `OKOU_SEO_DATAFORSEO_PASSWORD`) were already empty on `main`, because neither branding exists at repository scope for them — the identical diff confirms this is not a regression.

Note that this comparison proves **source selection** is unchanged. It cannot prove the canonical secret *values* are correct, which is precisely why the legacy secrets stay.

## What this explicitly does NOT change

- No `ZERO_*` secret is deleted from repository scope or the `production` environment.
- `turbo/apps/api/src/lib/env.ts` and its 22 `OKOU_ENV_FALLBACKS` entries, still gated on the local-development migration.
- `add_var`, `add_secret`, `add_oauth_client_config_key` and every other non-`okou` helper.
- Any rendered key name, value, or ordering.
- Any Doppler-sourced value.

`git diff --stat` touches only the action and its test script.

Closes #28237

### DIFF
--- a/.github/actions/web-api-env/action.yml
+++ b/.github/actions/web-api-env/action.yml
@@ -123,13 +123,6 @@ runs:
           add_env "$1" "$(repo_secret "$2")"
         }
 
-        add_okou_secret() {
-          local suffix="$1"
-          local value
-          value="$(first_non_empty "$(repo_secret "OKOU_${suffix}")" "$(repo_secret "ZERO_${suffix}")")"
-          add_env "OKOU_${suffix}" "$value"
-        }
-
         add_oauth_client_config_key() {
           local key="$1"
           local value
@@ -355,12 +348,12 @@ runs:
           FINANCE_APIDOJO_TOKEN
         )
         for suffix in "${okou_provider_secret_suffixes[@]}"; do
-          add_okou_secret "$suffix"
+          add_secret "OKOU_${suffix}" "OKOU_${suffix}"
         done
         if [[ "$INPUT_APP" == "api" ]]; then
-          add_okou_secret SEO_DATAFORSEO_LOGIN
-          add_okou_secret SEO_DATAFORSEO_PASSWORD
-          add_okou_secret BROWSER_USE_API_KEY
+          add_secret OKOU_SEO_DATAFORSEO_LOGIN OKOU_SEO_DATAFORSEO_LOGIN
+          add_secret OKOU_SEO_DATAFORSEO_PASSWORD OKOU_SEO_DATAFORSEO_PASSWORD
+          add_secret OKOU_BROWSER_USE_API_KEY OKOU_BROWSER_USE_API_KEY
         fi
         add_secret STEAM_WEB_API_KEY STEAM_WEB_API_KEY
         add_secret FINICITY_APP_KEY FINICITY_APP_KEY
--- a/.github/scripts/tests/web-api-env-action-test.sh
+++ b/.github/scripts/tests/web-api-env-action-test.sh
@@ -91,10 +91,7 @@ assert_no_fixture_secret_values() {
   local unexpected
   for unexpected in \
     "github-" \
-    "doppler-" \
-    "okou-google-weather-token" \
-    "okou-dataforseo-login" \
-    "okou-browser-use-api-key"; do
+    "doppler-"; do
     if [[ "$output" == *"$unexpected"* ]]; then
       fail "render output exposed a fixture secret value"
     fi
@@ -179,7 +176,7 @@ run_action() {
   local input_app="${3:-api}"
   local input_environment="${4:-preview}"
   local input_cli_pkg_url="${5-https://static.vm0.io/okou-cli/test-sha/package.tgz}"
-  local branded_config="${6:-legacy}"
+  local branded_config="${6:-canonical}"
   local action_script="${test_dir}/web-api-env-action.sh"
   local github_output="${test_dir}/github-output"
   local repo_vars_json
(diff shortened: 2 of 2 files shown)


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: ad76d9ee6f4bb79d5378582db486021a3c80336b
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Removes the inert `add_okou_secret` fallback from `.github/actions/web-api-env/action.yml` so the eight branded secret families read their canonical `OKOU_*` name directly, mirroring the form merged for the variable half in #28178. The test script's secret fixtures move to canonical names and the fixture that existed only to pin `OKOU_`-over-`ZERO_` precedence is deleted with the fallback. No `ZERO_*` secret is deleted from GitHub.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic — no wrong assumptions or unhandled edge cases | Pass | `add_secret X X` expands to `add_env X "$(repo_secret X)"`, identical to the old `add_env "OKOU_${suffix}" "$value"` whenever the canonical secret resolves. Both paths emit `KEY=` when nothing resolves, so the `empty` fixture behaves unchanged. |
| `REVIEW.md` | Security — no hardcoded secrets or credentials | Pass | Only secret *names* appear in the diff. Fixture values are synthetic (`github-*`). `assert_no_fixture_secret_values` still guards every rendered output. |
| `REVIEW.md` | Style & maintainability — no duplicated or dead logic | Pass | `add_okou_secret` deleted; `first_non_empty` is still used by 8 other call sites so it correctly stays. The `both` branded-config branch is removed rather than left dead. |
| `REVIEW.md` | Testing coverage for a `refactor:` commit — existing tests still cover the code | Pass | All eight families remain asserted across the `success`, `empty`, `production web` and `production api` fixtures in `.github/scripts/tests/web-api-env-action-test.sh`. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, contract, or `jsonb(...).$type<...>()` change; the diff is CI shell/YAML only. |
| `docs/docs.md` | Read the practice documents matching the changed surface | Pass | Changed surface is CI deploy tooling; `bad-smell.md`, `testing.md`, `fallback.md` and `deployment-compatibility.md` were read from `main`. No React/ccstate, event-sourcing, chat-card, or runner-architecture surface is touched. |
| `docs/bad-smell.md` | §7 fallback patterns — fail fast rather than papering over missing configuration | Pass | Removing `first_non_empty(OKOU_…, ZERO_…)` moves this closer to fail-fast: a missing canonical secret now renders empty and surfaces, instead of being silently masked by a legacy name. |
| `docs/bad-smell.md` | §2 no lint/type suppressions | Pass | No suppression comments added; no `shellcheck disable` introduced. |
| `docs/fallback.md` | §10 a removal PR must establish evidence that the branch is unreachable | Pass | The PR body carries rollout + configuration evidence: every level holding a `ZERO_*` secret also holds the matching `OKOU_*`, checked at repository scope and in `production`, plus a before/after render comparison over the real names for all four app/environment combinations. |
| `docs/fallback.md` | §10 the removal deletes the branch, its contract entry, and the branch's own tests together | Pass | `add_okou_secret`, the `both` fixture branch, and the `precedence_*` assertions that existed only to exercise the fallback are deleted in the same commit. |
| `docs/fallback.md` | §9 every fallback *introduced* by the diff is declared in a `Fallbacks` section | N/A | The diff introduces no fallback; it only removes one. §9 explicitly exempts removal-only PRs from the section. |
| `docs/fallback.md` | §1 no negative tests asserting removed behavior stays removed | Pass | See the note under Findings. The extended `grep` guard pins a currently-reachable invariant, not deleted code: the eight `ZERO_*` secrets deliberately still exist, so a reintroduced `repo_secret "ZERO_…"` would silently resolve and restore the fallback. `assert_migrated_zero_outputs_absent` is pre-existing and unchanged by this diff. |
| `docs/testing.md` | Tests exercise real code paths at the entry point rather than internals | Pass | The script extracts and executes the real action body via `extract_action_script` and asserts on the rendered env file — an integration test of the action, not of a helper. |
| `docs/testing.md` | Mock at the system boundary only | Pass | Only the GitHub-provided `REPO_VARS_JSON` / `REPO_SECRETS_JSON` / `DOPPLER_SECRETS_JSON` inputs are substituted. No internal logic is stubbed. |
| `docs/deployment-compatibility.md` | Cross-version compatibility for independently deployed surfaces | N/A | Frontend, backend, and runner contracts are untouched. The env file is baked at build time, and `rollback-production.yml` promotes an existing Vercel deployment without invoking this action, so no rendered artifact re-reads a source during rollback. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Notes (non-blocking)

- `.github/actions/web-api-env/action.yml:348-356` — the evidence establishes that each canonical secret **exists**, not that its **value** is correct. GitHub does not expose secret values, so a canonical secret holding an empty or mistyped value would previously have been masked by `first_non_empty` and will now render as-is. The PR body names this residual risk explicitly, and it is the stated reason the eight `ZERO_*` secrets are retained so recovery stays a one-line revert. Consistent with the owner's gate decision in vm0-ai/vm0#26701 (comment 5342453539); no change requested.
- `.github/scripts/tests/web-api-env-action-test.sh:91-95` — `assert_no_fixture_secret_values` drops the three `okou-*` literals that belonged to the deleted fixture. Coverage is unchanged in practice: every remaining fixture secret value is prefixed `github-` or `doppler-`, both of which the guard still checks. The function and all four call sites are intact.
- `.github/scripts/tests/web-api-env-action-test.sh:179` — renaming `branded_config`'s default from `legacy` to `canonical` is slightly outside the stated scope, but the fixture became fully canonical with this change, so the old name would have been actively misleading. Reasonable to keep.

### Testing

- Coverage: Adequate. This is a `refactor:` commit and existing tests still cover every affected path. All eight secret families are asserted under their canonical keys with unchanged values in the `success` and `production api` fixtures, the web-only subset in `production web`, and the unset case in `empty`. The `empty` fixture correctly switches to stripping `OKOU_*` from secrets so those families still render as empty values.
- Conventions: Compliant. The existing source guard is widened from variable reads to `repo_secret` / `add_secret` as well, so the action can no longer name a legacy source. `assert_migrated_zero_outputs_absent` still runs against all four remaining rendered fixtures. No test prints a secret value.
- Verified locally: `bash -n` clean, `.github/scripts/tests/web-api-env-action-test.sh` passes, and the rendered env file is byte-identical to `main` for api/production (210 keys), web/production (196), api/preview (211) and web/preview (196) when both action versions are fed the real repository- and environment-scoped names.

PR #28242, opened 2026-08-19
test(runner): synchronize heartbeat coalescing triggers

### LINKED ISSUE #28241: test(runner): eliminate heartbeat coalescing test race
## Background

The runner test `heartbeat_triggers_coalesce_into_one_current_mode_follow_up` intermittently fails in the full Rust CI suite with three recorded heartbeats instead of the expected two:

```text
  left: 3
 right: 2
```

The focused test passed 100 consecutive local runs, which is consistent with a scheduling-sensitive full-suite failure.

## Problem

The test advances paused Tokio time for a third routine heartbeat tick and then uses one `tokio::task::yield_now()` as evidence that the main reactor consumed that tick while the first provider heartbeat remained blocked. A single yield does not establish that ordering.

If the provider is unblocked before the ready tick branch is processed, the heartbeat completion branch may run first. The existing coalesced follow-up then starts, and the delayed tick is legitimately handled as a third single-flight request. The production controller still preserves its maximum in-flight count of one; the flaky assertion comes from an unproven test scheduling assumption.

## Expected behavior

The integration test should deterministically establish that each intended heartbeat trigger has been processed by the main reactor before releasing the blocked provider request. It should continue to verify that:

- multiple requests made while one heartbeat is in flight collapse into one follow-up;
- the follow-up samples the current runner mode when it starts;
- heartbeat snapshot sequences remain monotonic; and
- heartbeat provider calls never overlap.

## Acceptance criteria

- Replace the one-yield scheduling assumption with event-driven synchronization at the main-loop entry point exercised by the test.
- Do not weaken the exact two-heartbeat assertion or increase a timeout.
- Keep the synchronization test-only and avoid changing production heartbeat semantics.
- The focused flaky test passes repeatedly.
- Applicable runner formatting, linting, documentation, and tests pass.

## Relevant files

- `heartbeat.rs` main-loop integration tests
- `mod.rs` runner main reactor and test observer

seven332:
## Summary

- add a test-only main-loop event after routine heartbeat requests enter the controller
- synchronize the coalescing regression test with cursor-based observer waits
- preserve the exact heartbeat count, live mode, snapshot sequence, state, and single-flight assertions

## Why

The test previously used one `yield_now()` as proof that a ready timer branch had run. Under scheduler pressure, the blocked heartbeat could complete first, allowing the delayed timer branch to create a legitimate third single-flight request. The new event establishes the required ordering directly.

## Scope

This changes test instrumentation and integration-test synchronization only. Production heartbeat behavior, protocols, and runtime scheduling are unchanged.

## Validation

- focused heartbeat coalescing test repeated 100 times
- `cargo test -p runner --bin runner --all-features` (`3200 passed`, `20 ignored`)
- Rust pre-commit hooks (workspace formatting, Clippy, and documentation)

Closes #28241

### DIFF
--- a/crates/runner/src/cmd/start/mod.rs
+++ b/crates/runner/src/cmd/start/mod.rs
@@ -978,6 +978,7 @@ enum SignalSource {
 #[derive(Debug, PartialEq, Eq)]
 enum StartLoopEvent {
     BudgetExhaustedReactorEntered,
+    RoutineHeartbeatRequested { mode: RunnerMode },
     WorkspaceCacheChangeObserved,
     DestroyTasksDrainEntered,
     DestroyTasksDrainCompleted,
@@ -1111,6 +1112,10 @@ impl StartLoopTestObserver {
         self.record(StartLoopEvent::BudgetExhaustedReactorEntered);
     }
 
+    fn notify_routine_heartbeat_requested(&self, mode: RunnerMode) {
+        self.record(StartLoopEvent::RoutineHeartbeatRequested { mode });
+    }
+
     fn notify_workspace_cache_change_observed(&self) {
         self.record(StartLoopEvent::WorkspaceCacheChangeObserved);
     }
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #28242

**Reviewed head:** `e7fb97beb8179bfbc759150409dca8242a352e9d`

### Summary

The PR replaces an unproven one-yield scheduling assumption with cursor-based, test-only observation of routine heartbeat requests after they enter the real controller. Production heartbeat behavior is unchanged, and the existing strict integration assertions remain intact.

### Findings

No findings.

### Testing

- focused regression test passed 100 consecutive runs
- runner binary all-feature suite passed (`3200 passed`, `20 ignored`)
- formatting, Clippy, runner documentation, and Rust pre-commit checks passed
- no deployment compatibility or runtime performance surface is changed

### Verdict

LGTM

PR #28247, opened 2026-08-19
perf(runner): reuse serialized firewall auth request

### LINKED ISSUE #28144: perf(python): avoid duplicate auth payload serialization on first firewall fetch
## Problem

The first firewall-auth use for a VM snapshot/API synchronously serializes the same complete auth material twice on the mitmproxy event loop:

1. `crates/runner/mitm-addon/src/auth.py` builds the content-based cache identity from a canonical JSON representation containing the full non-forced auth body.
2. On the resulting response-cache miss, `crates/runner/mitm-addon/src/firewall_auth_client.py` rebuilds and serializes the wire body before the coroutine reaches its first I/O await.

Both passes include `encryptedSecrets`, auth templates, connector maps/metadata, `vars`, billable state, and matched-firewall routing inputs. The registry accepts up to 16 MiB, so this synchronous work scales with admitted runtime input rather than a small fixed structure. Up to four firewall-auth fetches can be active, but their pre-I/O serialization still runs serially on mitmproxy's event loop.

## Evidence

A focused local benchmark of the two current operations measured the following median combined cost for a representative body containing one large `encryptedSecrets` value:

| Auth material | Identity pass | Wire-body pass | Combined |
| --- | ---: | ---: | ---: |
| 100 KB | 0.30 ms | 0.23 ms | 0.53 ms |
| 1 MB | 3.23 ms | 2.48 ms | 5.71 ms |
| 4 MB | 11.70 ms | 9.13 ms | 21.06 ms |
| 16 MB | 48.24 ms | 37.03 ms | 89.66 ms |

Timings are host-dependent, but they demonstrate linear duplicate traversal and allocation before network I/O. This is distinct from #21964/#21994, which made repeated cache hits independent of payload size while intentionally retaining one lazy identity derivation for the first use of each snapshot/API.

## Decision

Proceed with one PR that reuses the canonical serialized non-forced request body between first-use identity derivation and the initial normal fetch.

## Chosen direction

- `FirewallAuthRequest` remains the single owner of decoded request-field and omission semantics and gains one compact canonical serializer.
- On an identity-sidecar miss, serialize the normal request body once. Derive a new-version content identity from that body digest plus the existing firewall name/base and sandbox-token identity fields.
- Attach the prepared normal bytes only to that request-local `FirewallAuthRequest`. The existing cache path continues passing the request object unchanged; no separate cache API parameter or snapshot-owned body storage is needed.
- `fetch_firewall_headers()` reuses attached bytes only for a normal fetch. A forced refresh ignores them and serializes its `forceRefresh: true` body once.
- Identity-sidecar hits construct requests without prepared bytes, so response-cache hits remain independent of payload size and later TTL misses serialize exactly once when a fetch is actually created.
- The snapshot sidecar retains only the final identity digest; large bytes live no longer than the initiating request/fetch.

This keeps the existing cache singleflight and ownership boundaries unchanged while ensuring the body used for first-use identity and wire delivery comes from one serialization.

## Alternatives considered

- **Exact legacy digest through manual JSON fragment composition:** preserves an unobserved internal digest but introduces fragile security-sensitive encoding logic.
- **Eager registry preparation:** moves synchronous work to registry loading, serializes unused APIs, and can retain large VM-wide material per API.
- **Worker-thread offload:** leaves duplicate CPU/allocation and adds async/thread cancellation and contention behavior.
- **A separate prepared-body parameter through every cache layer:** works functionally but broadens the cache API even though the request object already owns the value and follows the required path.

## Risks and mitigations

- **Identity isolation regression:** hash the exact non-forced `to_body()` result, retain firewall name/base and token digest, increment the private version, and run the existing changed-input matrix.
- **Normal/forced body confusion:** make the prepared value explicitly normal-only and have forced fetches always serialize their own body.
- **Large memory retention:** keep prepared bytes on the request only; never put them in the snapshot identity entry.
- **Cache-hit or follower regression:** leave cache inspection and task selection unchanged and count serialization deterministically across miss, hit, and refresh paths.
- **Wire formatting change:** assert decoded request fields and omissions against the real fake endpoint; the API contract is semantic JSON and does not expose ordering or whitespace.
- **Remaining first-pass event-loop work:** one full serialization remains necessary for content-derived lookup. This issue removes the measured duplicate pass; thread offload is not justified by current evidence.

## Acceptance criteria

- A first-use normal identity miss plus response-cache miss performs one full auth-body serialization total.
- A repeated identity/response-cache hit performs no full auth-body serialization.
- A forced refresh retains the non-forced identity, performs one forced-body serialization, and preserves `forceRefresh` and all other omission semantics.
- Content-based isolation remains intact across VM snapshots, APIs, firewall name/base context, sandbox tokens, vars, connector metadata, billable state, and routing inputs.
- Large serialized bodies are not retained in snapshot-owned identity state.
- Cache singleflight, admission, cancellation, TTL, billable expiry, refresh cooldown, registry ownership, and request/requestheaders behavior remain unchanged.
- Focused tests use deterministic serialization counts and decoded request assertions rather than wall-clock thresholds.
- Ruff, basedpyright, focused auth tests, the full mitm-addon pytest suite, and applicable pre-commit checks pass.

## Delivery notes

The change is internal to the Python mitm-addon. It requires no database, persisted format, registry schema, runner protocol, API schema, or frontend/backend change. Old and new runners can overlap because identity and response caches are process-local. The private identity version change requires no migration or cleanup.

seven332:
## Summary

- serialize the normal firewall-auth webhook body once on a first-use identity miss and reuse those bytes for the initial fetch
- derive a content identity from the canonical body digest while preserving firewall, sandbox-token, snapshot, and routing isolation
- keep prepared bytes request-local, leave the auth-cache API unchanged, and serialize forced-refresh bodies independently
- cover the real request-handler miss, cache-hit, and force-refresh lifecycle against the local fake auth endpoint

## Scope

- no registry, runner protocol, persisted-state, API schema, frontend, or backend change
- no change to auth-cache admission, singleflight, TTL, cancellation, or refresh ownership

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_request_handler_firewall_auth.py -q` (34 passed)
- `uv run --no-sync python -m pytest tests/test_auth_cache.py tests/test_firewall_auth_client.py -q` (146 passed)
- `uv run --no-sync python -m pytest tests/test_firewall_aws_sigv4_auth.py -q` (50 passed)
- `uv run --no-sync python -m pytest tests/` (6387 passed)
- `lefthook run pre-commit`

Closes #28144

### DIFF
--- a/crates/runner/mitm-addon/src/auth.py
+++ b/crates/runner/mitm-addon/src/auth.py
@@ -119,7 +119,6 @@ AWS_SIGV4_REQUEST_HEADER_FIELD_OVERHEAD_BYTES = 32
 MAX_AWS_SIGV4_REQUEST_HEADER_FIELDS = (
     MAX_AWS_SIGV4_REQUEST_HEADER_LIST_BYTES // AWS_SIGV4_REQUEST_HEADER_FIELD_OVERHEAD_BYTES
 )
-_FIREWALL_AUTH_IDENTITY_VERSION = 2
 _FIREWALL_AUTH_IDENTITY_CACHE_KEY = "_firewallAuthIdentityCache"
 _MISSING_AWS_SIGV4_ORIGINAL_URL = object()
 
@@ -345,7 +344,6 @@ def _build_firewall_auth_identity(
     normal_body = auth_request.to_bytes(force_refresh=False)
     sandbox_token_sha256 = hashlib.sha256(auth_request.sandbox_token.encode("utf-8")).hexdigest()
     material = {
-        "version": _FIREWALL_AUTH_IDENTITY_VERSION,
         "firewallName": firewall_name,
         "firewallBase": firewall_base,
         "authBodySha256": hashlib.sha256(normal_body).hexdigest(),
--- a/crates/runner/mitm-addon/src/auth.py
+++ b/crates/runner/mitm-addon/src/auth.py
@@ -119,7 +119,7 @@ AWS_SIGV4_REQUEST_HEADER_FIELD_OVERHEAD_BYTES = 32
 MAX_AWS_SIGV4_REQUEST_HEADER_FIELDS = (
     MAX_AWS_SIGV4_REQUEST_HEADER_LIST_BYTES // AWS_SIGV4_REQUEST_HEADER_FIELD_OVERHEAD_BYTES
 )
-_FIREWALL_AUTH_IDENTITY_VERSION = 1
+_FIREWALL_AUTH_IDENTITY_VERSION = 2
 _FIREWALL_AUTH_IDENTITY_CACHE_KEY = "_firewallAuthIdentityCache"
 _MISSING_AWS_SIGV4_ORIGINAL_URL = object()
 
@@ -130,6 +130,14 @@ class _FirewallAuthIdentityCacheEntry:
 
     api_entry: dict = field(repr=False)
     auth_identity: str
+
+
+@dataclass(frozen=True)
+class _ResolvedFirewallAuthIdentity:
+    """Content identity and request-local body preparation for one lookup."""
+
+    auth_identity: str
+    auth_request: FirewallAuthRequest = field(repr=False)
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


seven332:
## Code Review: PR #28247

**Reviewed head:** `99ee0be6c8550313a3011f1e37052d2c70b6f320`

### Summary

The PR replaces the duplicate first-use auth-payload serialization with one canonical normal body shared by identity derivation and the initial fetch. The new identity version retains every existing security boundary, prepared bytes stay request-local, forced refresh remains independently serialized, and the auth-cache implementation is unchanged.

### Findings

No findings. The submitted diff matches the challenged plan and completed scope, correctness, test, concurrency/performance, resource/security, and maintainability passes cleanly.

### Testing

- Ruff format/lint and basedpyright passed.
- Focused request-handler, auth-cache/client, and SigV4 suites passed (34, 146, and 50 tests).
- Full mitm-addon suite passed: 6387 tests.
- Lefthook pre-commit and commitlint passed.
- GitHub CI is still in progress; production payload distribution was not re-benchmarked because the regression gate is deterministic serialization count plus real HTTP body behavior.

### Verdict

LGTM

PR #28249, opened 2026-08-19

e7h4n:
Daily deployment-compatibility cleanup. One cohort survived evidence collection; everything else is inside its window, blocked, or has an evidence gap, and is listed at the bottom rather than split into a second PR.

## Cohort — require `publicBrand` in both shared-thread responses (#27660)

**Old → new boundary.** [PR #27200](https://github.com/vm0-ai/vm0/pull/27200) added `public_brand` to `shared_threads`. Both response schemas kept `publicBrandSchema.default("vm0")` so a current browser client, or the Pages worker rendering share links, could still parse a response from an API rolled back to a version that predates the field.

**What this actually removes.** `SharedThreadResponse` and `SharedThreadMetaResponse` already declare `publicBrand: PublicBrand` as **required** TypeScript fields. The `.default("vm0")` was therefore purely a *runtime parse-time* tolerance for a response that omits the value — which is precisely the rollout shim, and why `pnpm check-types` is unaffected by its removal.

**Producer proof, source-level.** `shared_threads.public_brand` is `text("public_brand").$type<PublicBrand>().default("vm0").notNull()`, and both readers in `shared-thread.service.ts` project the column directly (`publicBrand: sharedThreads.publicBrand`) rather than reconstructing it. There is no code path that can emit either response without the field.

**Window.** The surface is web/app-to-API, and the reachable direction is a **new parser reaching an older API** — i.e. the API rollback window, with ~2 days quoted because current browser clients stay live until refresh. #27200 merged 2026-08-17T07:03:27Z (`fa77b3e80fae`); `app/production` promoted `44934b36f181` at **2026-08-17T08:07:43Z** (verified to contain that commit) and `api/production` promoted at 08:06:15Z. At this run that is **~60.9 hours (2.54 days)**, past the ~2-day window, with roughly thirty `api/production` promotions since, so no rollback target still omits the field.

**Evidence gaps, recorded rather than smoothed over:**

- **MaskDB is not usable here.** `vm0-prod-ro` returns `shared_threads` with only **2 columns** and no `public_brand`. That is a degraded/stale schema view, not a statement about production — the same masked-branch lag I recorded on 2026-08-19 (and on 2026-08-13 for migration 0907's tables). The column's existence in production is instead established by the [#28072](https://github.com/vm0-ai/vm0/pull/28072) evidence merged yesterday: the public-brand readiness gate probed `shared_threads.public_brand` among thirteen tables as a whole-API kill switch, and production served 3,019,275 requests with 17 unrelated 503s, which is only possible if that probe returned true.
- **Axiom gives no coverage for this route.** A bounded query (`vm0-request-log-prod`, `2026-08-17T08:07:43Z → 2026-08-19T21:30:00Z`, `_time` predicate first) returned **no** `shared-threads` traffic at all; the OAuth routes in the same query were live (`/api/zero/slack/oauth/install` 34 requests, 0 5xx). Share links are a low-traffic surface, so this cohort rests on the source-level producer proof above, not on log observation. Saying so explicitly rather than presenting an empty result as confirmation.
- **The Pages worker is out of repo.** `sharedThreadMetaResponseSchema` has no in-repo consumer; only the API route registers `meta`. The worker that renders share-link metadata is not in this repository, so I cannot inspect its contract version or release cadence. It is affected only if it rebuilds against a stricter contract *and* then reaches a pre-#27200 API — and that rollback target is closed. Flagged because it is the one consumer this run could not verify directly.

**Persisted-data reasoning.** None applies: this changes a response parser only. No row is read, written, or migrated.

## Checks

Run once against the combined diff, with a local PostgreSQL 18 cluster created for this run and set to `UTC`:

- `pnpm format`; `pnpm check-types` (all tasks, no errors — expected, since the types already required the field); `pnpm knip` (exit 0); lint on `@okouai/api-contracts` and `@okouai/app` (both exit 0)
- `pnpm -F @okouai/app exec vitest run shared-thread-page.test.tsx chat-thread-sharing.test.tsx` → **7 passed** (the in-repo consumers of both schemas)
- `pnpm -F api exec vitest run chat-event-archive-consumers.test.ts` → **4 passed** (the API-side contract user)

## Open-PR overlap

**None.** No open PR touches `shared-threads.ts`.

Deferred this run, with the exact blocker:

- **#27660's OAuth half** (`slack-oauth.ts:137`, `teams-oauth.ts:162`) — deliberately left, and the comment there is misleading in a way worth correcting: it says "Old **web/app** OAuth state can omit publicBrand", but the platform never constructs these states. The API mints them from `publicBrand$` (`slack-oauth.ts:159`), so the real skew is API-version, not client-version. More importantly the shim is **not separable**: `record.publicBrand === "okou" ? "okou" : "vm0"` collapses *absent* and *malformed* into one branch, and `parseOAuthState` already returns a `vm0` default for a missing or unparseable state. Removing the rollout tolerance therefore cannot be done without also deciding new behavior for malformed input, which #27660 does not specify and which would change an OAuth callback from a friendly redirect to a 500.
- **#27356** zero-agent default-avatar bridge — window long closed, but `apps/api/src/signals/routes/agents.ts:322` still reads `avatarUrl: args.body.avatarUrl ?? null` while only the create path at line 399 uses `randomPresetAvatar()`; that upsert reaches its INSERT branch from the PUT route for a compose-backed agent with no `zero_agents` row. The trigger is compensating a live writer. **Fifth consecutive run reporting this; the unblocking change is one line.**
- **#27602 / #27796** integration identity contraction and **#26938** agent/compose consolidation — multi-stage programs with their own Stage-8 removal owners.
- **#26672 / #26519** Website template archive pins (`LatestWebsiteTemplates` still `enabled: false`), `PersonalModelProviderAccounts` fallbacks (still `enabled: false`) — product gates, not time windows.
- **#26487** route-entry Phase A (separately authorized only), **#25620** avatar flat-field read fallback (needs a jsonb backfill), and the `browser_*` identity contraction (MaskDB exposes no `browser_*` table, so `chat_thread_id` uniqueness across retained rows still cannot be proven).

Resolved since the last run without this workflow: **#26152** and **#26842** are both closed, ending a seven-run deferral chain across #26393 → #27128 → #27764.

### DIFF
--- a/turbo/packages/api-contracts/src/contracts/shared-threads.ts
+++ b/turbo/packages/api-contracts/src/contracts/shared-threads.ts
@@ -156,19 +156,13 @@ const sharedThreadResponseSchema: ZodLikeSchema<SharedThreadResponse> =
     id: z.string().uuid(),
     title: z.string(),
     messages: z.array(sharedMessageZodSchema),
-    // On the web/app-to-API surface, an API rollback can omit this additive
-    // field while current browser clients remain live for about two days.
-    // Remove after that client/API window closes (tracked by #27660).
-    publicBrand: publicBrandSchema.default("vm0"),
+    publicBrand: publicBrandSchema,
   });
 
 const sharedThreadMetaResponseSchema: ZodLikeSchema<SharedThreadMetaResponse> =
   z.object({
     title: z.string(),
-    // The Pages worker may reach an API version from before publicBrand was
-    // added. Keep the old response legal for the same web/API rollback window
-    // of about two days, then remove it with #27660.
-    publicBrand: publicBrandSchema.default("vm0"),
+    publicBrand: publicBrandSchema,
   });
 
 const sharedThreadApiErrorSchema: ZodLikeSchema<ApiErrorResponse> =


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: ea1010fadec356a90120c60d886da921f578b263
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Removes the two `publicBrandSchema.default("vm0")` rollout shims from `sharedThreadResponseSchema` and `sharedThreadMetaResponseSchema` in `turbo/packages/api-contracts/src/contracts/shared-threads.ts`, making `publicBrand` required at parse time. Both `SharedThreadResponse` and `SharedThreadMetaResponse` already declared the field as required TypeScript, so this is a pure runtime parse-tolerance removal, which is why `check-types` is unaffected. No fallback is introduced; nothing else in the diff.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic: no unhandled edge case or wrong assumption introduced | Pass | Only two Zod field declarations change; both routes (`getSharedThread$`, `getSharedThreadMeta$` in `turbo/apps/api/src/signals/routes/shared-threads.ts:74-100`) return rows that always carry the column. |
| `REVIEW.md` | Security: no auth, injection, or secret exposure surface | N/A | Diff touches only response-schema strictness on already-public share-snapshot routes; no auth, query construction, or credential path changes. |
| `REVIEW.md` | Type safety: no `any`, no over-broad types | Pass | `publicBrandSchema` is reused unchanged; `ZodLikeSchema<SharedThreadResponse>` / `ZodLikeSchema<SharedThreadMetaResponse>` annotations are untouched and still satisfied. |
| `REVIEW.md` | Style and maintainability; comments explain WHY | Pass | The two removed comments were removal-condition notes for the shim; deleting them with the shim is correct, and no comment debt is left behind. |
| `REVIEW.md` | DB/JSONB review gate for `packages/api-contracts/**` response schemas | Pass | No persisted shape changes. `shared_threads.public_brand` is `text(...).$type<PublicBrand>().default("vm0").notNull()` (`turbo/packages/db/src/schema/shared-thread.ts:28`); no migration or backfill is implied by tightening a response parser. |
| `REVIEW.md` | Feature regression guard: existing behavior not silently changed | Pass | The only in-repo runtime consumer, `setupSharedThreadPage$` (`turbo/apps/platform/src/signals/shared-thread-page/shared-thread-page-setup.ts:32-38`), calls `client.get`; it receives `publicBrand` on every live response, so no existing flow changes. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions, no dynamic imports | Pass | Diff adds two plain schema fields and removes eight lines; no suppression comment, cast, or `import()` anywhere in it. |
| `docs/bad-smell.md` | Fail fast instead of masking absent values with defaults | Pass | The change moves in the required direction: a missing `publicBrand` now fails the parse loudly instead of being silently rewritten to `"vm0"`. |
| `docs/fallback.md` §5 | Do not keep readers for retired producers | Pass | The tolerant branch existed only for an API version that predates `#27200`; that producer version is no longer reachable, so the reader is correctly deleted. |
| `docs/fallback.md` §10 | A removal PR must establish type/schema, single-writer, production, or rollout evidence | Pass | Type/schema: both interfaces already require the field. Single-writer: `readSharedThread$` and `readSharedThreadMeta$` (`shared-thread.service.ts:317-345`) project `sharedThreads.publicBrand` directly from a `NOT NULL DEFAULT 'vm0'` column, and no other code path builds either body. Rollout: `#27200` promoted to `api/production` at 2026-08-17T08:06:15Z; the GitHub deployments API records 25 further `api/production` promotions since, so no rollback target still omits the field. |
| `docs/fallback.md` §10 | The removal deletes the branch, its contract entry, and its own tests together | Pass | No test asserted the field-less shape: every `publicBrand` occurrence under `shared-thread-page.test.tsx`, `chat-event-archive-consumers.test.ts`, and `artifact-catalog.bdd.test.ts` supplies the value explicitly, so there is no orphaned fallback test to delete. |
| `docs/fallback.md` §1 | No negative tests asserting removed behavior stays removed | Pass | The diff adds no test at all, so no tombstone is introduced. |
| `docs/fallback.md` §9 | Every fallback introduced by the diff is declared in the PR summary | N/A | The diff introduces no fallback; it only removes two. Per §9, a removal-only PR needs neither a `Fallbacks` section nor `Fallbacks: none`. |
| `docs/deployment-compatibility.md` | Frontend to backend: do not tighten a response contract while an old producer can still serve the old shape | Pass | The reachable direction is new parser to older API. That API rollback window closed ~61 hours ago, past the ~2-day client window quoted in the code comment, and the 25 intervening promotions make a pre-`#27200` rollback target unreachable. |
| `docs/deployment-compatibility.md` | Old frontend to new backend direction stays safe | Pass | The API still emits `publicBrand`; an already-open pre-`#27200` browser page parses it with a schema that never required the field, so old clients are unaffected in both endpoints. |
| `docs/deployment-compatibility.md` | Clean up compatibility only after the old deployed version is no longer active, with the removal condition satisfied | Pass | The removed comments named the surface and window explicitly; the PR body supplies the promotion timestamps that close it, and the deployment record confirms them independently. |
| `docs/testing.md` | `refactor:` commits keep existing coverage; no new required tests | Pass | Behavior under test is unchanged for every live path. The PR reports `shared-thread-page` + `chat-thread-sharing` (7 passed) and `chat-event-archive-consumers` (4 passed), which are the in-repo consumers of both schemas. |
| `docs/testing.md` | Mock boundary, MSW, real DB, no fake timers, mock cleanup | N/A | The diff changes no test file, so none of the test-convention rules have a subject in this PR. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

- `turbo/apps/api/src/signals/routes/shared-threads.ts:92-96`: the `meta` endpoint answers with `Cache-Control: public, max-age=31536000, s-maxage=31536000, immutable`, and that header predates `#27200` (it arrived with `#27048`). A shared-cache entry produced before 2026-08-17 therefore still holds a body without `publicBrand`, and that exposure is bounded by the cache TTL rather than by the ~2-day client window this removal reasons about. It is not a blocker here: `sharedThreadMetaResponseSchema` has no in-repo runtime consumer, `client.get` (the only in-repo caller) is served `no-store`, and `@okouai/api-contracts` is `private: true`, so the stricter schema cannot reach the out-of-repo Pages worker through a package bump. Worth recording for whoever does adopt this contract in that worker.
- The three evidence gaps the PR body states openly (stale MaskDB schema view, zero Axiom coverage for this low-traffic route, out-of-repo Pages worker) are the right ones to disclose, and the source-level producer proof carries the conclusion without them. Deferring `#27660`'s OAuth half is also correct: `record.publicBrand === "okou" ? "okou" : "vm0"` collapses absent and malformed into one branch, so removing that shim would be a separate behavior decision, not a window-closed cleanup.

### Testing

- Coverage: Adequate. This is a `refactor:` that tightens a parser with no reachable behavior change, so no new test is required. Existing coverage of both schemas stays green per the PR's `shared-thread-page.test.tsx` + `chat-thread-sharing.test.tsx` (7 passed) and `chat-event-archive-consumers.test.ts` (4 passed) runs.
- Conventions: Compliant. No test file is touched, so no mock-boundary, MSW, real-database, fake-timer, or cleanup rule applies, and no tombstone test is left behind by the removal.

PR #28267, opened 2026-08-20
perf(runner): share network log path ownership

### LINKED ISSUE #28255: perf(rust): avoid per-row path allocations in network-log admission
## Problem

`NetworkLogState` copies the same owned filesystem path three times for every accepted Rust-side network-log row:

- `source_snapshot` clones the `PathBuf` stored in `SourceState` into every snapshot (`crates/runner/src/network_log_manager/state.rs:158-166`).
- `try_accept_snapshot` clones that snapshot path for `pending_paths.entry(...)` (`crates/runner/src/network_log_manager/state.rs:180-184`). The owned key is constructed even when the path is already present, so the allocation is discarded on the common occupied-entry path.
- The same function clones the path again into `AcceptedAppend` for the writer queue (`crates/runner/src/network_log_manager/state.rs:185-188`).

`NetworkLogManager::append_for_ip` runs this sequence for each admitted row before it reaches the existing sharded/batched writer (`crates/runner/src/network_log_manager.rs:256-299`). Both DNS and kmsg producers call this API once per parsed event (`crates/runner/src/dns/log.rs:349-356`, `crates/runner/src/kmsg_log.rs:247-254`).

The result is at least three heap-backed path copies per accepted row. High-volume DNS or packet logs across concurrent sandboxes therefore create allocator and memory-copy pressure proportional to row count, even though the downstream writer already batches filesystem work.

## Recommended direction

Represent the immutable path/route identity with shared ownership across `SourceState`, `SourceSnapshot`, pending-path accounting, and `AcceptedAppend` (for example, an `Arc`-backed path or route object). Updating an already-present pending counter should not allocate and copy the underlying path bytes, and queueing a row should only clone a cheap shared handle.

Keep the scope limited to admission-path ownership. Do not change queue sizing, sharding, batching, serialization, or filesystem append behavior.

## Semantics to preserve

- A row waiting for bounded queue capacity must still re-check the source generation before acceptance.
- Source re-registration must not redirect an old accepted row to the new path.
- Pending state must be registered before enqueue so `flush_path` cannot miss an accepted row.
- Same-path ordering and shard selection must remain stable.
- Write failures must still release pending counts.
- `close_for_upload` must still finalize source acceptance before its final flush.

## Acceptance criteria

- The underlying path buffer is not cloned for each snapshot, occupied pending-path lookup, and accepted writer item.
- The existing `network_log_manager` ordering, re-registration, backpressure, failure, and close/flush tests continue to pass.
- Focused coverage demonstrates that repeated rows for one registered source reuse the same path identity while preserving old-path ownership across re-registration.

## Related work

- #17302 introduced the bounded sharded writer and removed per-row task/file-open overhead.
- #18581 coalesced writes inside each path batch.
- #20439 split state and writer ownership but intentionally preserved the admission behavior above.

seven332:
## Summary

- share each immutable network-log path with `Arc<Path>` across source state, snapshots, pending accounting, queued appends, writer batches, and completion
- preserve generation checks, value-based shard selection, pending flush barriers, and old-path ownership across source re-registration
- extend the gated manager integration test to verify repeated appends reuse one path allocation and remain routed to the old path after re-registration

## Out of scope

- no queue sizing, sharding, batching, serialization, filesystem append, drain, or upload behavior changes
- no API, protocol, payload, persisted-state, or deployment-boundary changes

## Validation

- `cargo test -p runner --all-features network_log_manager` (25 passed)
- `cargo test -p runner --all-targets --all-features` (3200 passed, 20 ignored; guest inventory 1 passed)
- `git diff --check`
- pre-commit hooks and commitlint

Closes #28255

### DIFF
--- a/crates/runner/src/network_log_manager.rs
+++ b/crates/runner/src/network_log_manager.rs
@@ -66,7 +66,7 @@ struct CloseGate {
 pub struct NetworkLogSession {
     manager: NetworkLogManager,
     source_ip: String,
-    path: PathBuf,
+    path: Arc<Path>,
     generation: u64,
     closed: bool,
 }
@@ -142,7 +142,7 @@ impl Drop for NetworkLogSession {
 
         let manager = self.manager.clone();
         let source_ip = self.source_ip.clone();
-        let path = self.path.clone();
+        let path = Arc::clone(&self.path);
         let generation = self.generation;
         if let Ok(handle) = tokio::runtime::Handle::try_current() {
             std::mem::drop(handle.spawn(async move {
@@ -626,13 +626,22 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn pending_old_write_stays_on_old_path_after_reregister() {
+    async fn repeated_appends_share_path_and_stay_on_old_path_after_reregister() {
         let dir = tempfile::tempdir().unwrap();
         let old_path = dir.path().join("old.jsonl");
         let new_path = dir.path().join("new.jsonl");
         let started = Arc::new(Notify::new());
         let release = Arc::new(Semaphore::new(0));
-        let manager = NetworkLogManager::new_with_write_gate(started.clone(), release.clone());
+        let manager = NetworkLogManager::new_with_write_gate_and_config(
+            started.clone(),
+            release.clone(),
+            WriterConfig {
+                shards: 1,
+                queue_capacity: 4,
+                max_batch_rows: 1,
+                max_batch_bytes: DEFAULT_MAX_BATCH_BYTES,
+            },
+        );
 
         let _old_session = manager
             .register_source_ip("10.200.0.2", old_path.clone())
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #28267

**Reviewed head:** `859771e9c4684d2f9cba44bf82772a369f427073`

### Summary

The change consistently carries an immutable `Arc<Path>` through network-log registration, admission, pending accounting, writer batching, and completion. Path-value hashing/equality, source generations, pending-before-enqueue ordering, write-failure completion, and close/flush behavior remain intact.

### Findings

No findings.

### Testing

Focused network-log tests passed (25), as did runner formatting, Clippy with warnings denied, rustdoc with warnings denied, and the complete runner suite (3200 passed, 20 ignored, plus guest inventory). The added gated integration scenario verifies shared allocation identity for repeated rows and old/new path routing without sleeps. Metal/E2E tests and an allocation benchmark were not run; the change has no external protocol or sandbox behavior boundary.

### Verdict

LGTM

PR #28272, opened 2026-08-20
fix(ably-subscriber): refresh expired token before reconnect

### LINKED ISSUE #28258: bug(rust): renew Ably token after token-expired disconnect
## Summary

`ably-subscriber` reconnects with the cached access token after Ably sends a mid-session `DISCONNECTED` message with error code `40142` (`Token expired`). Because the connection key is still inside the resumable window, the subscriber skips token exchange and can repeatedly offer the expired credential until `connectionStateTtl` expires (120 seconds by default).

Ably documents `40142` as a token-renewal condition: clients configured for automatic token renewal should obtain a fresh token and continue the connection.

## Evidence

- `crates/ably-subscriber/src/connection/event_loop.rs:762-776` handles every mid-session `DISCONNECTED` identically and does not preserve an authentication-refresh requirement from the error code.
- The outer reconnect path records `disconnected_at` (`event_loop.rs:469-470`), while `ConnState::can_resume()` remains true whenever the connection key is younger than `connection_state_ttl` (`crates/ably-subscriber/src/connection/state.rs:70-79`). The default TTL is 120 seconds (`crates/ably-subscriber/src/types.rs:265-274`).
- `attempt_reconnect()` fetches a new token only when `can_resume()` is false. During the resumable window it builds the WebSocket URL from `p.session.token()`, which is the token the server just rejected (`crates/ably-subscriber/src/connection/event_loop.rs:990-1015`).
- `crates/ably-subscriber/tests/integration/reconnect.rs:652-733` sends `40142` and comments that the client should make a "fresh connect with new token," but its mock WebSocket accepts any authentication query and the test does not assert a second token-endpoint request. The targeted test therefore passes while the implementation reuses the old token.
- Ably's error reference states that a token-renewal-capable client should request a fresh token for `40142`: https://ably.com/docs/platform/errors/codes/40142-token-expired

## Impact

When the server invalidates a token before the client's local expiry timer, the subscription can remain disconnected for the full connection-state TTL while reconnect attempts reuse the rejected token. Runner instances lose real-time Ably notifications during that window; work polling switches to the slower fallback path, and cancellation or active-input notifications are no longer delivered in real time.

## Acceptance criteria

- A mid-session `DISCONNECTED` with error code `40142` causes a fresh token request and exchange before the next WebSocket connection attempt, even while the prior connection key is resumable.
- The reconnect attempt never reuses the token rejected as expired. Resume state may be retained if the protocol supports resuming with the refreshed token.
- An integration test asserts both that the token endpoint is called again and that the reconnect WebSocket request carries the refreshed token.
- Existing resume/backoff behavior for non-authentication `DISCONNECTED` messages remains covered.

seven332:
## Summary

- force a fresh Ably token exchange after a mid-session `DISCONNECTED` with error code `40142`
- preserve the existing connection resume key while using the refreshed credential
- verify the token exchange count and reconnect WebSocket query through the public subscription integration path

## Scope

This change handles the documented token-expired code only. It does not broaden renewal to other Ably token errors, add a pending-token cache, or change public subscriber, runner, persistence, or deployment contracts.

## Validation

- `cargo fmt --all --check`
- `cargo clippy -p ably-subscriber --all-targets -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p ably-subscriber --no-deps`
- `cargo test -p ably-subscriber`

Closes #28258

### DIFF
--- a/crates/ably-subscriber/src/connection/event_loop.rs
+++ b/crates/ably-subscriber/src/connection/event_loop.rs
@@ -247,6 +247,7 @@ pub(crate) async fn run_event_loop(mut p: EventLoopState, mut close_rx: CloseRec
 
     'outer: loop {
         let mut disconnected_sent = false;
+        let mut refresh_token = false;
         let mut immediate_retry = false;
         let mut disconnect_reason = None;
         let mut close_before_reconnect = false;
@@ -283,11 +284,15 @@ pub(crate) async fn run_event_loop(mut p: EventLoopState, mut close_rx: CloseRec
                     {
                         match send_attach(&mut p, &mut close_rx).await {
                             LoopAction::Stop => return,
-                            LoopAction::Reconnect { disconnected_event } => {
+                            LoopAction::Reconnect {
+                                disconnected_event,
+                                refresh_token: action_refresh_token,
+                            } => {
                                 record_reconnect_disconnected_event(
                                     &mut disconnected_sent,
                                     disconnected_event,
                                 );
+                                refresh_token |= action_refresh_token;
                                 immediate_retry = true;
                                 break;
                             }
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #28272

**Reviewed head:** `abf3fe29bfd005e8631b30dd3211d6c34cdb45e1`

### Summary

The current head correctly preserves Ably connection resume state while forcing a fresh token exchange after the documented `40142` token-expired disconnect. The refresh requirement is scoped to that structured error, survives failed reconnect attempts, and commits the refreshed token through the existing successful reconnect path.

### Findings

No findings. A stale test comment found during self-review was fixed on this head.

### Testing

The integration test exercises `subscribe()` through mock HTTP and WebSocket boundaries and asserts two token exchanges, the refreshed `access_token`, the retained `resume` key, and post-resume message delivery. Local format, all-target Clippy, rustdoc warnings-as-errors, the full `ably-subscriber` test suite, and repository Rust commit hooks passed. A live Ably sandbox reconnect was not run; CI status is checked separately.

### Verdict

LGTM

PR #28277, opened 2026-08-20
refactor(rust): centralize process output validation

### LINKED ISSUE #28257: refactor(rust): centralize process output validation across sandbox backends
## Problem

`ProcessOutputMode` is the shared `sandbox` API for `Sandbox::start_process`, but its validity rules are currently owned by multiple backend layers:

- `crates/sandbox/src/types.rs:796` defines the public mode and sets `DEFAULT_QUEUE_CAPACITY` to 8192, without exposing one canonical validation path.
- `crates/sandbox-mock/src/sandbox.rs:362` independently rejects only a zero `chunk_limit_bytes` or zero `queue_capacity`.
- `crates/sandbox-fc/src/sandbox.rs:2486` maps the same mode into a supervised vsock exec request.
- `crates/vsock-host/src/exec_operation/state.rs:472` additionally rejects a queue capacity above `MAX_EXEC_STREAM_CAPACITY` (8192).

This has already produced observable semantic drift: a stream request with `queue_capacity: 8193` succeeds through `MockSandbox` but fails through `FirecrackerSandbox`. Tests that exercise code against the mock can therefore accept a process configuration that the production backend rejects.

The overlapping responsibility also requires future `ProcessOutputMode` constraints to be updated independently in the shared type, mock backend, and vsock-backed production path.

## Canonical ownership

Shared `ProcessOutputMode` invariants and bounds should have one canonical owner alongside the type in the `sandbox` crate. Transport-specific code may retain defense-in-depth validation, but it should consume the same canonical limits rather than define a different public contract.

## Acceptance criteria

- Define one canonical validation contract for `ProcessOutputMode`, including non-zero stream chunk size, non-zero queue capacity, and the supported maximum queue capacity.
- Make both `MockSandbox::start_process` and the Firecracker-backed path apply that contract before starting a process.
- Remove independent shared-rule implementations or make lower-level defense-in-depth checks consume the same canonical constants.
- Add boundary coverage showing that both backends reject 0, accept 8192, and reject 8193 for stream queue capacity, and that both reject a zero stream chunk size.
- Preserve buffered-output behavior and consistent `SandboxOperation::StartProcess` error classification.

Verified against `d0c3348d11bfea754c260d1717bed1d7283c52d0`.

seven332:
## Summary

- define the canonical `ProcessOutputMode` stream validation contract and maximum queue capacity in `sandbox`
- apply the shared validation in both mock and Firecracker `start_process` paths with consistent operation errors
- keep the lower-level vsock safety check while deriving its limit from the shared API contract
- add backend boundary coverage for zero, maximum, and oversized queue capacities plus zero chunk limits

## Scope

- no vsock protocol, persisted-state, or deployment compatibility changes
- buffered output behavior remains unchanged

## Validation

- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `cargo test -p sandbox -p sandbox-mock -p sandbox-fc -p vsock-host`
- `cargo test --workspace --all-features`

Closes #28257

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -3734,6 +3734,7 @@ name = "vsock-host"
 version = "0.17.84"
 dependencies = [
  "nix",
+ "sandbox",
  "shell-quote",
  "tokio",
  "tracing",
--- a/crates/sandbox-fc/src/sandbox.rs
+++ b/crates/sandbox-fc/src/sandbox.rs
@@ -2490,6 +2490,7 @@ impl Sandbox for FirecrackerSandbox {
         let operation = SandboxOperation::StartProcess;
         let vsock = self.begin_guest_operation(operation).await?;
         Self::validate_exec_env_keys(operation, request.env)?;
+        request.output.validate()?;
 
         let start_future = async move {
             vsock
--- a/crates/sandbox-fc/src/sandbox/tests.rs
+++ b/crates/sandbox-fc/src/sandbox/tests.rs
@@ -2233,6 +2233,122 @@ fn invalid_start_process_env_key_returns_operation_error() {
     }
 }
 
+#[tokio::test]
+async fn start_process_output_rejects_invalid_stream_configuration() {
+    let sandbox = test_sandbox_with_state(SandboxState::Running);
+    let guest = attach_mock_shutdown_guest(&sandbox).await;
+
+    for (output, expected_message) in [
+        (
+            ProcessOutputMode::Stream {
+                stream_limit_bytes: 1024,
+                chunk_limit_bytes: 0,
+                queue_capacity: 1,
+                stderr_capture_limit_bytes: None,
+            },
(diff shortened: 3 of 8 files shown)


seven332:
## Code Review: PR #28277

**Reviewed head:** `64db9bc82c5631edac212853177064c5516128a2`

### Summary

The shared `ProcessOutputMode` contract now owns stream validation, both backends apply it before process start, and the lower vsock guard derives its limit from the same constant. The change is scoped, fail-fast, and protocol-compatible.

### Findings

No findings.

### Testing

Reviewed boundary coverage for zero chunk size and queue capacities 0, 8192, and 8193 through both public backend entry points. Local Rust formatting, workspace Clippy, workspace rustdoc, affected-crate tests, and workspace tests all passed. Turbo and Python checks are outside this Rust-only scope.

### Verdict

LGTM

PR #28291, opened 2026-08-20
feat(api): pin chat thread media models at creation

bingjiez666:
## Problem

`chat_threads.selected_video_model` / `selected_image_model` stayed null until someone touched the composer picker, and both run-time resolvers treat null as *follow the member default*:

- `video-model.service.ts` — thread pin → member default → catalog default
- `image-model.service.ts` — same chain, gated on `ImageModelSelection`

That chain is evaluated **per run at dispatch**, so writing a new personal default retargeted every thread that had never been explicitly repinned, including threads the user had already worked in.

The run model does not behave this way: the create path resolves it and the API rejects a create without one (`chat-threads-create.ts`), so a thread is concrete from creation. This puts the media models on the same footing.

## Change

New `chat-thread-media-model.service.ts` resolves *member default → catalog default* once, at insert time. Every production creation path calls it:

| Path | File |
|---|---|
| `POST /chat-threads` (web, `okou chat create`, agent sub-threads) | `routes/chat-threads-create.ts` |
| `POST /chat-events` implicit create | `services/chat-events.command.ts` |
| Slack ingress | `services/slack-chat-ingress.service.ts` |
| Teams ingress | `services/teams-chat-ingress.service.ts` |
| Telegram ingress | `services/telegram-chat-ingress.service.ts` |
| Feishu ingress | `services/feishu-chat-ingress.service.ts` |
| AgentPhone ingress | `services/agentphone-chat-ingress.service.ts` |
| Goal threads | `services/goal.service.ts` |
| Workflow automation threads | `services/workflow-user-automation-thread.service.ts` |

Precedence on `POST /chat-threads` is unchanged at the top and only gains a last step: explicit request body → the calling run's thread pin → member default → catalog default.

The resolved pins also go onto the `created` chat-thread event, so event replay and snapshot compaction carry them.

`loadUserFeatureSwitchOverrides` / `loadUserFeatureSwitchContext` widened from `ReadonlyDb` to `Pick<ReadonlyDb, "select">` so the resolver can run on a transaction handle without an `as Db` cast. Type-widening only; no call site changes.

## Decisions

- **Each model is pinned only while its own picker is enabled** — video behind `VideoModelSelection`, image behind `ImageModelSelection`. A member who cannot reach a picker has no default of their own to freeze, so pinning would only capture whichever catalog default was current and stop that thread from following the catalog. Written pins also survive a revert of this code, so the switch is the containment for that too. With neither switch on, the resolver returns early and skips the member lookup.
- **No backfill.** Existing threads keep null and keep following the member default.
- **Retired ids still fall through.** The resolver narrows through the current catalog, matching how dispatch narrows an existing pin.

## Tests

- `chat-threads-create.test.ts`: three cases split by which picker is on — neither switch pins nothing, video-only pins video and leaves image null, both pin both from the member defaults.
- `chat-events.bdd.test.ts`: the video and image snapshot chains asserted the old semantics (change the member default, reuse the same thread, expect the new default). Both rewritten to assert the pin survives the change and that a thread created *after* the change takes it. New cases cover a thread created with the video picker disabled (stays unpinned, keeps following the member default) and a thread whose pin was cleared back to null.
- `agentphone.bdd.test.ts`: asserts the created event of an ingress thread carries the pin. AgentPhone is the representative ingress path; Teams and Telegram share its create-and-return shape, Slack and Feishu resolve inline.

## Checks

npm registry access was unavailable in this environment, so `pnpm install` failed and format, lint, type-check and vitest could not run locally. Relying on the PR pipeline for all of them; happy to push fixes on the first red run.


bingjiez666:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 96852b41a7f72b6452bea796d273cfd5695a1164

### Summary

Resolves the member/catalog media-model defaults when a chat thread row is inserted, so a thread answers with the video and image model it was created with instead of re-reading the member default on every run. Covers all nine production creation paths through a new shared `chat-thread-media-model.service.ts`, and stamps the resolved pins onto the `created` chat-thread event. Existing threads keep null and are deliberately not backfilled.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no race conditions or wrong assumptions | Pass | Precedence in `chat-threads-create.ts:121-134` is body → run-inherited → member/catalog. Both web create paths (`POST /chat-threads`, `POST /chat-events`) resolve identically, so the `onConflictDoNothing` race between them cannot produce divergent pins. |
| `REVIEW.md` | Security: auth checks on changed endpoints | Pass | No auth surface changed; `resolveNewChatThreadMediaModels` is scoped by the caller's already-authorized `orgId`/`userId`. |
| `REVIEW.md` | Type safety; no `any`, no overly broad types | Pass | No `any`, no assertions added. `Pick<ReadonlyDb, "select">` widening in `feature-switches.service.ts:93,124` is a narrowing-of-requirement, and it removes the need for a `tx as Db` cast at the transaction call sites. |
| `REVIEW.md` | Feature regression guard: changed existing behavior is scoped and contained | **Finding (P1)** | Video pinning applies to every org, but `VideoModelSelection` is `enabled: false` + staff-only. See P1-1. |
| `REVIEW.md` | Feature regression guard: tests cover the existing flow most likely to regress | **Finding (P1)** | Seven ingress/goal/automation creation paths changed behavior with zero assertions. See P1-3. |
| `REVIEW.md` | DB/JSONB gate: persisted contract changes carry migration/normalizer/rationale | Pass | No schema or contract change. `selected_video_model` / `selected_image_model` already exist as nullable varchar (`schema/chat-thread.ts:112,118`) and the event fields are already optional-nullable (`contracts/chat-threads.ts:163-168,195-196`). Only the written value changes. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | None added across the 14 changed files. |
| `docs/bad-smell.md` | No dynamic `import()`; no hardcoded URLs/config | Pass | All new imports are static; no configuration literals introduced. |
| `docs/bad-smell.md` | Fail fast; no defensive try/catch | Pass | New service has no error handling; a missing member row resolves through the documented default path rather than being swallowed. |
| `docs/fallback.md` | Every fallback introduced by the diff is declared in the PR summary | Pass | The diff introduces no new fallback, so no `Fallbacks` section is required. See the note below the findings for the two paths that were evaluated and rejected as fallbacks. |
| `docs/fallback.md` | No fallback for a state the contract already prevents | Pass | `catalogedVideoModel` / `catalogedImageModel` (`chat-thread-media-model.service.ts:42-48`) guard a genuinely reachable state — a nullable member column and an id that was valid at write time but has since left the catalog — matching the identical narrowing already documented in `video-model.service.ts:19-32`. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | The new BDD case asserts a live behavior (a null-pinned thread still resolves through the member default), not the absence of deleted code. |
| `docs/fallback.md` | Non-GA feature-switched behavior needs no compatibility code | Pass | Image pinning correctly returns null while `ImageModelSelection` is off instead of writing a pin that would later be frozen. |
| `docs/deployment-compatibility.md` | Old code after migration / new code before migration | N/A | No migration and no schema change; both columns and both event fields already ship in production. |
| `docs/deployment-compatibility.md` | Old frontend → new backend during rollout | Pass | Old clients already read `selectedVideoModel` / `selectedImageModel` as optional-nullable and render an effective model either way; receiving a concrete value where they previously saw null is the value they were already resolving to locally. |
| `docs/deployment-compatibility.md` | Data written by one version, read by another | Pass | During API promotion, threads created by the old instance stay null and remain resolvable by the new instance's unchanged fallback chain; the mixed outcome is benign and self-limiting. |
| `docs/event-sourcing.md` | Persistent events authoritative; optimistic reconciled by shared event ID | Pass | The `created` event now carries the resolved pins and reconciles against the client's optimistic `selectedVideoModel: null` by the shared `eventId`; no rollback path added. |
| `.claude/skills/ccstate/SKILL.md` | ccstate usage in `apps/api` follows the practice doc | Pass | New code is a plain async service function matching sibling `video-model.service.ts` / `image-model.service.ts`; no new signals, no `set()` of a function value, no manual cache over a `computed`. |
| `docs/testing.md` | `feat:` commits carry integration tests | Pass | Two new route tests plus rewritten and one new BDD case; no unit tests added. |
| `docs/testing.md` | Mock at boundary only; real DB; MSW; no fake timers | Pass | No new `vi.mock()`, no fetch stubbing, no fake timers; tests run against the real database. |
| `docs/testing.md` | New user-facing features gated behind a `FeatureSwitchKey` | **Finding (P1)** | Image is gated on `ImageModelSelection`; video is not gated at all. See P1-1. |
| `docs/testing/patterns.md` | Route integration tests via `setupApp()` + ts-rest contract | Pass | `preferenceClient()` / `threadsClient()` in `chat-threads-create.test.ts:126-137` follow the documented shape. |
| `docs/testing/api-testing.md` | Construct cases through the API, verify through the API | Pass (new route tests) | The new cases seed the member default through `PUT /user-model-preference` and assert through the events endpoint rather than reading rows. |
| `docs/testing/testing-external-behavior.md` | Leave the boundary only when the state is unconstructible via the production API | **Finding (P1)** | The new BDD case reaches for a DB fixture to produce a null pin that `POST /chat-threads/:id/video-model` can produce. See P1-2. |
| `docs/testing/anti-patterns.md` | No mock-call assertions, no fetch/fs mocking, no internal mocking | Pass | Every new assertion checks a resolved model value observed through an endpoint or the run snapshot fixture, not a call pattern. |

### Findings

None.

- **P1-1 `turbo/apps/api/src/signals/services/chat-thread-media-model.service.ts:76-93` — video pinning ships ungated to every org, while image pinning is gated.** `resolveImageModelForRun` takes a `featureSwitchContext` and the new resolver correctly returns `null` for image while `ImageModelSelection` is off, with the stated reason that "the member cannot reach the image picker yet, so there is no choice worth freezing". That reasoning applies verbatim to video: `resolveVideoModelForRun` is called unconditionally at `agent-run-create.service.ts:8893` with no switch, and `VideoModelSelection` is `enabled: false` with `enabledOrgIdHashes: STAFF_ORG_ID_HASHES` (`feature-switch.ts:290-295`). For every non-staff org the member default is therefore always null, so the only thing the new pin can freeze is `DEFAULT_VIDEO_MODEL` — and it freezes it permanently. When that catalog constant is next advanced, threads created after this merge keep the superseded model instead of picking up the new one, which is a behavior change for users who have no picker and gained nothing from pinning. This is also the one part of the PR a revert cannot undo: reverting the code leaves the already-written pins in place. Gate the video pin on `FeatureSwitchKey.VideoModelSelection` the same way image is gated, so containment matches where the problem this PR solves actually exists.
- **P1-2 `turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts:11770-11800` and `turbo/apps/api/src/test-fixtures/chat-thread-events.ts:148-160` — the legacy-null case leaves the external-behavior boundary unnecessarily.** The new case widens `setChatThreadVideoModelFixture` to accept `null` and writes the column directly, but `POST /chat-threads/:id/video-model` accepts `{ model: null }` (`chat-threads.ts:883-887`, handled at `chat-threads-video-model.ts:29-34`) and produces exactly the same `selected_video_model IS NULL` state that resolution reads. `docs/testing/testing-external-behavior.md` names "an existing helper can write the DB directly" as explicitly not an exception. Drive the clear through the route and revert the fixture widening; the case is worth keeping, only its construction is wrong.
- **P1-3 seven changed creation paths have no coverage.** `slack-chat-ingress`, `teams-chat-ingress`, `telegram-chat-ingress`, `feishu-chat-ingress`, `agentphone-chat-ingress`, `goal.service`, and `workflow-user-automation-thread.service` all changed from writing no media pin to writing one, and each wires the resolver into two separate places — the `chat_threads` insert and the `created` event. A missing spread in either place is invisible to the type system because both fields are optional, and no test asserts a pin on any of these surfaces. Add at least one ingress-path assertion so the wiring pattern is covered once; the current tests only exercise the two web paths.
- **P1-4 `turbo/apps/api/src/signals/services/chat-events.command.ts:1315-1318` — duplicate feature-switch query inside the hot dispatch transaction.** `resolveNormalSendFeatureSwitches` already loads the full `FeatureSwitchContext` for this same request at `chat-events.command.ts:2414`, before `resolveTimedThread` reaches `createChatThread`. The new resolver loads it again inside the transaction, so every new-thread send pays an extra round trip, and it pays it even on the `clientThreadId` retry path where the insert is a no-op. This path is instrumented with `measureApiDispatchTiming` at every other step; the two new queries are uninstrumented. Thread the already-resolved switch state (or the boolean) down through `resolveThread` instead of reloading it. The same applies more mildly to `chat-threads-create.ts:125`, which resolves the defaults even when the request body supplies both models.

### Fallbacks

The diff introduces none, so no `Fallbacks` section is required in the PR summary. Two paths were evaluated and are not fallbacks:

- `chat-thread-media-model.service.ts:42-48` `catalogedVideoModel` / `catalogedImageModel` — catalog-membership narrowing over a genuinely nullable persisted column and an id that can legitimately outlive its catalog entry. `docs/fallback.md` §6.3 covers this, and it mirrors the identical, already-documented narrowing in `video-model.service.ts:19-32`. Not a cross-version rollout branch, so no window applies.
- Threads that keep `NULL` and continue resolving through the member default — this keeps an existing path rather than introducing one, which `docs/fallback.md` §9 explicitly exempts from the declaration requirement.

### Testing

- **Coverage**: Insufficient — the two web creation paths are well covered (catalog default with the image switch off, member defaults with it on, pin survives a later member-default change, a thread created after the change takes the new default, and a null-pinned thread still follows the member default), but the seven ingress/goal/automation paths changed by this PR have no assertions at all (P1-3).
- **Conventions**: One violation — the legacy-null case constructs its state by writing the database instead of calling the route that can produce it (P1-2). No `vi.mock()` of internal modules, no fetch or filesystem mocking, no fake timers, no mock-call assertions, and the new route tests correctly seed and verify through endpoints.
- **Note on local validation**: the PR summary states that npm registry access was unavailable, so format, lint, type-check and vitest did not run locally. CI covers all of them and is currently green on this head, so this is not a finding — but the P1-1 gating change and the P1-2/P1-3 test changes will need a fresh CI pass.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 9a16914ac513bb66f68197783e3f242c764227db
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Re-review at `9a16914a` after the fix commit. Chat threads now stamp their video and image models at creation across all nine production creation paths, with each model gated on its own picker switch. All four P1 findings from the review of `96852b41` are resolved, and the fix commit introduces no new P0/P1.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no race conditions or wrong assumptions | Pass | Both web create paths resolve through the same gated resolver, so the `onConflictDoNothing` race between `POST /chat-threads` and `POST /chat-events` still cannot produce divergent pins. Precedence in `chat-threads-create.ts:121-136` is unchanged: body → run-inherited → gated member/catalog. |
| `REVIEW.md` | Security: auth checks on changed endpoints | Pass | No auth surface changed. The resolver reads `org_members_metadata` and the switch overrides for the caller's already-authorized `orgId`/`userId`. |
| `REVIEW.md` | Type safety; no `any`, no overly broad types | Pass | `NewChatThreadMediaModels` widened to `VideoModel \| null` so the gate is expressed in the type rather than at call sites. `Pick<ReadonlyDb, "select">` still avoids the `tx as Db` cast the sibling `resolveRequiredDefaultChatThreadModelPin` needs. |
| `REVIEW.md` | Feature regression guard: changed existing behavior is scoped and contained | **Pass (was P1-1)** | `chat-thread-media-model.service.ts:98-108` gates the video pin on `VideoModelSelection` and the image pin on `ImageModelSelection`. Orgs without a picker keep `NULL` and keep following the live default, so the ungated all-org behavior change is gone and the one part a revert cannot undo is now switch-contained. |
| `REVIEW.md` | Feature regression guard: tests cover the existing flow most likely to regress | Pass | `chat-events.bdd.test.ts:11688-11742` asserts a switch-off thread stays unpinned and still picks up a member default written later — the exact pre-change behavior every non-staff org keeps. |
| `REVIEW.md` | DB/JSONB gate: persisted DB contract changes | Pass | Still no schema or contract change; only the written value differs, and it is now written for fewer orgs than at `96852b41`. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions, no dynamic imports, no hardcoded config | Pass | None added in the fix commit. |
| `docs/bad-smell.md` | Fail fast; no defensive try/catch | Pass | The early return at `:106-108` is a feature gate, not error recovery; it mirrors `resolveImageModelForRun`'s own switch short-circuit. |
| `docs/fallback.md` | Every fallback introduced by the diff is declared in the PR summary | Pass | The diff still introduces none, so no `Fallbacks` section is required. See the note below the findings. |
| `docs/fallback.md` | No fallback for a state the contract already prevents | Pass | `catalogedVideoModel` / `catalogedImageModel` (`:52-58`) guard a nullable member column and an id that can outlive its catalog, matching `video-model.service.ts:19-32`. |
| `docs/fallback.md` | Non-GA feature-switched behavior needs no compatibility code | Pass | Both pins are now behind non-GA switches, which is what §2 asks for; the PR adds no dual-read, migration, or tolerance branch for either. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | The unpinned cases assert live behavior — an unpinned thread resolving through the member default — not the absence of deleted code. |
| `docs/deployment-compatibility.md` | Old code after migration / new code before migration | N/A | No migration and no schema change; both columns and both event fields already ship in production. |
| `docs/deployment-compatibility.md` | Old frontend → new backend during rollout | Pass | Old clients already read both event fields as optional-nullable. Gating narrows the population that sees a concrete value where it previously saw null. |
| `docs/deployment-compatibility.md` | Data written by one version, read by another | Pass | `NormalSendFeatureSwitches` is in-process only. During API promotion, old-instance threads stay null and remain resolvable by the new instance's unchanged fallback chain. |
| `docs/event-sourcing.md` | Persistent events authoritative; optimistic reconciled by shared event ID | Pass | The `created` event still carries the resolved pins and reconciles against the client's optimistic `null` by the shared `eventId`; no rollback path added. |
| `.claude/skills/ccstate/SKILL.md` | ccstate usage in `apps/api` follows the practice doc | Pass | Both new exports are plain async service functions matching sibling `video-model.service.ts` / `image-model.service.ts`; no new signals or manual caches. |
| `docs/testing.md` | `feat:` commits carry integration tests | Pass | Route tests split three ways by which picker is on, plus BDD coverage for switch-off, switch-on, cleared-pin, and the AgentPhone ingress path. |
| `docs/testing.md` | Mock at boundary only; real DB; MSW; no fake timers | Pass | No new `vi.mock()`, no fetch stubbing, no fake timers. |
| `docs/testing.md` | New user-facing features gated behind a `FeatureSwitchKey` | **Pass (was P1-1)** | Both media pins now sit behind their own `FeatureSwitchKey`. |
| `docs/testing/patterns.md` | Route integration tests via `setupApp()` + ts-rest contract | Pass | `threadVideoModelClient()` (`api-bdd-chat-files.ts:347-349`) is wired the same way as the existing image client, with its route slice added to the shared list. |
| `docs/testing/api-testing.md` | Construct cases through the API, verify through the API | Pass | The route tests seed member defaults through `PUT /user-model-preference` and assert through the events endpoint. |
| `docs/testing/testing-external-behavior.md` | Leave the boundary only when the state is unconstructible via the production API | **Pass (was P1-2)** | The cleared-pin case now calls `POST /chat-threads/:id/video-model` with `{ model: null }` through the new `chat.updateThreadVideoModel` helper, and the `setChatThreadVideoModelFixture` widening is fully reverted — that file no longer appears in the PR diff. |
| `docs/testing/anti-patterns.md` | No mock-call assertions, no fetch/fs mocking, no internal mocking | Pass | Every new assertion checks a resolved model value observed through an endpoint or the run snapshot. |

### Findings

None.

None. All four findings from the `96852b41` review are resolved:

- **P1-1 resolved** — `chat-thread-media-model.service.ts:98-118` gates each pin on its own switch and returns `UNPINNED` without touching `org_members_metadata` when neither is on. `chat-events.bdd.test.ts:11688-11742` and `chat-threads-create.test.ts` ("leaves media models unpinned while neither picker is enabled", "pins each media model whose picker is enabled") lock both sides of the gate.
- **P1-2 resolved** — the cleared-pin state is constructed through the production route; the fixture is back to its original `string`-only signature.
- **P1-3 resolved** — `agentphone.bdd.test.ts` now asserts the created event of an ingress thread carries the pin, covering the create-and-return wiring that Teams and Telegram share. The PR summary is explicit that Slack and Feishu resolve inline and are not separately asserted; that residual gap is a reasonable stopping point under "not too many tests" and is honestly recorded rather than implied to be covered.
- **P1-4 resolved** — `NormalSendFeatureSwitches` carries the `FeatureSwitchContext` the request already loaded, threaded through `resolveThread` into `createChatThread`, so the dispatch transaction no longer reloads it. With both switches off the transaction now issues zero extra queries instead of two. `chat-threads-create.ts` still resolves unconditionally, but the route needs the switch state anyway to decide gating and runs once per thread rather than once per send, so the remaining cost is not worth another branch.

#### Notes (non-blocking)

- `videoModelSelectionEnabled` and `featureSwitchContext` now coexist on `NormalSendFeatureSwitches`. The boolean predates this PR and still serves `videoRunOptions` at `chat-events.command.ts:2494`; deriving it from the context at that one call site would be a tidy-up, not a defect.
- `UNPINNED` is a shared module-level object returned by reference. Every caller spreads it into an insert or an event and none mutates it, so this is safe as written.

### Fallbacks

The diff introduces none, so no `Fallbacks` section is required in the PR summary. Two paths were evaluated and are not fallbacks:

- `chat-thread-media-model.service.ts:52-58` `catalogedVideoModel` / `catalogedImageModel` — catalog-membership narrowing over a genuinely nullable persisted column and an id that can legitimately outlive its catalog entry, covered by `docs/fallback.md` §6.3 and mirroring the already-documented narrowing in `video-model.service.ts:19-32`. Not a cross-version rollout branch, so no window applies.
- Threads that keep `NULL` and continue resolving through the member default — this keeps an existing path rather than introducing one, which `docs/fallback.md` §9 explicitly exempts from the declaration requirement.

### Testing

- **Coverage**: Adequate. Both gates are asserted in both directions at the route and at dispatch; the pin is asserted to survive a later member-default change; a thread created after that change takes the new default; a cleared pin falls back to the member default; and one ingress path asserts the created event carries the pin.
- **Conventions**: Compliant. The previous external-behavior-boundary violation is gone, no internal modules are mocked, and the new BDD helper follows the existing image-model client shape.
- **CI**: still running on this head — 39 checks green, the remainder pending, none failed at the time of review. The verdict is based on the code; a red check on this head should be handled as its own follow-up rather than reopening these findings.
- **Local validation**: the PR summary states npm registry access was unavailable, so format, lint, type-check and vitest did not run locally. That is an environment limitation, honestly recorded, and CI covers all four.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 52b0f6e1c046ba73a1db14c6d55ecf9b06ac0968
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Re-review at `52b0f6e1`. Since the `9a16914a` LGTM the branch gained a merge of `origin/main` and one lint fix: the early-return sentinel `UNPINNED` was a module-scope object, which `api/no-package-variable` rejects, so the branch now returns the literal from the branch itself. Behavior is identical and the indirection is gone. The merge touched none of the files this PR changes, and every authoritative document is at the same revision as the previous review.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no race conditions or wrong assumptions | Pass | `chat-thread-media-model.service.ts:99-101` returns the same shape the removed constant held; no caller mutated it, so inlining changes nothing. Both web create paths still resolve through one gated resolver, so the `onConflictDoNothing` race cannot produce divergent pins. |
| `REVIEW.md` | Security: auth checks on changed endpoints | Pass | No auth surface changed; the resolver still reads only the caller's already-authorized `orgId`/`userId`. |
| `REVIEW.md` | Type safety; no `any`, no overly broad types | Pass | The inlined literal is checked against the declared `Promise<NewChatThreadMediaModels>` return type, so dropping the annotated constant loses no checking. |
| `REVIEW.md` | Style and maintainability | Pass | Removing the hoisted sentinel is what the package's own lint rule asks for; the branch now reads as a plain early return. |
| `REVIEW.md` | Feature regression guard: changed existing behavior is scoped and contained | Pass | Video stays gated on `VideoModelSelection`, image on `ImageModelSelection` (`:93-105`). Orgs without a picker keep `NULL` and keep following the live default. |
| `REVIEW.md` | Feature regression guard: tests cover the existing flow most likely to regress | Pass | `chat-events.bdd.test.ts` still asserts a switch-off thread stays unpinned and picks up a member default written later. |
| `REVIEW.md` | DB/JSONB gate: persisted DB contract changes | Pass | No schema or contract change. The merged main commits include migration `0951`, which belongs to an unrelated connector change and does not touch `chat_threads`. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | The lint error was fixed at the root instead of suppressed — no `eslint-disable` was added. |
| `docs/bad-smell.md` | No dynamic imports; no hardcoded config; fail fast | Pass | Unchanged from the previous review. |
| `docs/fallback.md` | Every fallback introduced by the diff is declared in the PR summary | Pass | The diff still introduces none, so no `Fallbacks` section is required. |
| `docs/fallback.md` | No fallback for a state the contract already prevents | Pass | `catalogedVideoModel` / `catalogedImageModel` (`:47-53`) guard a nullable member column and an id that can outlive its catalog, matching `video-model.service.ts:19-32`. |
| `docs/fallback.md` | Non-GA feature-switched behavior needs no compatibility code | Pass | Both pins sit behind non-GA switches and the PR adds no dual-read, migration, or tolerance branch. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | The unpinned cases assert live resolution behavior, not the absence of deleted code. |
| `docs/deployment-compatibility.md` | Old code after migration / new code before migration | N/A | This PR adds no migration and no schema change; both columns and both event fields already ship in production. |
| `docs/deployment-compatibility.md` | Old frontend → new backend during rollout | Pass | Old clients already read both event fields as optional-nullable. |
| `docs/deployment-compatibility.md` | Data written by one version, read by another | Pass | `NormalSendFeatureSwitches` is in-process only; old-instance threads stay null and remain resolvable by the new instance's unchanged fallback chain. |
| `docs/event-sourcing.md` | Persistent events authoritative; optimistic reconciled by shared event ID | Pass | The `created` event carries the resolved pins and reconciles by the shared `eventId`; no rollback path added. |
| `.claude/skills/ccstate/SKILL.md` | ccstate usage in `apps/api` follows the practice doc | Pass | Both exports remain plain async service functions; the removed constant was not a signal and needed none. |
| `docs/testing.md` | `feat:` commits carry integration tests | Pass | Route tests split three ways by which picker is on, plus BDD coverage for switch-off, switch-on, cleared-pin, and the AgentPhone ingress path. |
| `docs/testing.md` | Mock at boundary only; real DB; MSW; no fake timers | Pass | No new `vi.mock()`, no fetch stubbing, no fake timers. |
| `docs/testing.md` | New user-facing features gated behind a `FeatureSwitchKey` | Pass | Both media pins sit behind their own `FeatureSwitchKey`. |
| `docs/testing/patterns.md` | Route integration tests via `setupApp()` + ts-rest contract | Pass | `threadVideoModelClient()` follows the existing image-model client shape with its route slice registered. |
| `docs/testing/api-testing.md` | Construct cases through the API, verify through the API | Pass | Member defaults are seeded through `PUT /user-model-preference` and asserted through the events endpoint. |
| `docs/testing/testing-external-behavior.md` | Leave the boundary only when the state is unconstructible via the production API | Pass | The cleared-pin case uses `POST /chat-threads/:id/video-model`; the fixture widening stays reverted and that file is absent from the PR diff. |
| `docs/testing/anti-patterns.md` | No mock-call assertions, no fetch/fs mocking, no internal mocking | Pass | Every new assertion checks a resolved model value observed through an endpoint or the run snapshot. |

### Findings

None.

None. The four findings from the `96852b41` review remain resolved, and this head adds only the lint fix.

#### Correction to the previous review

The `9a16914a` review listed `UNPINNED` as a non-blocking note and called it "safe as written" on the grounds that no caller mutates it. That reasoning was about runtime safety only and missed that `api/no-package-variable` bans module-scope objects in this package outright, which is what `lint-eslint` then caught. The rule is the governing standard here; the note should have been a finding.

### Fallbacks

The diff introduces none, so no `Fallbacks` section is required in the PR summary. Two paths were evaluated and are not fallbacks:

- `chat-thread-media-model.service.ts:47-53` `catalogedVideoModel` / `catalogedImageModel` — catalog-membership narrowing over a genuinely nullable persisted column and an id that can legitimately outlive its catalog entry, covered by `docs/fallback.md` §6.3 and mirroring `video-model.service.ts:19-32`. Not a cross-version rollout branch, so no window applies.
- Threads that keep `NULL` and continue resolving through the member default — this keeps an existing path rather than introducing one, which `docs/fallback.md` §9 exempts from the declaration requirement.

### Testing

- **Coverage**: Adequate and unchanged from `9a16914a`. Both gates are asserted in both directions at the route and at dispatch; the pin survives a later member-default change; a thread created after that change takes the new default; a cleared pin falls back to the member default; one ingress path asserts the created event carries the pin.
- **Conventions**: Compliant. No external-behavior-boundary violation, no internal mocking, and the new BDD helper mirrors the existing image-model client.
- **CI**: re-running on this head after the merge and lint fix — 11 checks green, the rest pending, none failed at review time. `lint-eslint` is the check this head is meant to turn green; if it or any other check comes back red, that is its own follow-up rather than a reopening of these findings.
- **Local validation**: npm registry access remains unavailable in the automation environment, so format, lint, type-check and vitest could not run locally — which is precisely why the `api/no-package-variable` violation reached CI. CI is the only gate for all four here.

PR #28280, opened 2026-08-20
refactor(platform): drop the unused browser pptx renderer

bingjiez666:
Part of #28252. First of three; the other two add `publish` and rewrite the upload dialog.

## What this removes

`src/lib/pptx-renderer.ts` and its three test files. Nothing in the app imports the module — the only importers are its own tests:

```
$ git grep -l "pptx-renderer" -- turbo/apps/platform/src
  src/lib/__tests__/pptx-renderer-ignored-content.test.ts
  src/lib/__tests__/pptx-renderer-resources.test.ts
  src/lib/__tests__/pptx-renderer.test.ts
  src/lib/pptx-renderer.ts
```

It was built (#26571) as the rendering primitive for an upload UI that is not being built. Under #28252 the deck arrives as an ordinary chat attachment and is analysed in a thread the user can watch, so the browser never needs to open a PPTX.

## The renderer is moving, not disappearing

The same `@aiden0z/pptx-renderer`, pinned to the same `1.2.4`, now runs inside the analysis sandbox (`vm0-ai/Template-artifact#18`), driven through `agent-browser` against the sandbox's own headless Chromium. Page images are still produced by the code that would have produced the preview — measured at 3 slides in 887 ms, CJK correct, with the 100-page cap enforced.

The frontend gives back two dependencies: `@aiden0z/pptx-renderer` and `modern-screenshot`.

## Four places kept it alive

Deleting the file alone would have left the build red or the checks lying:

- **`knip.json`** listed `src/lib/pptx-renderer.ts` as an *entry point*. That is why an unimported module never appeared as unused — it was being told to be a root. Removed alongside it.
- **`check-localized-ui.mjs`** carried four literal exclusions for the module's diagnostics. The script fails on stale exclusions, so these had to go with it.
- **`vitest.config.ts`** had `assetsInclude: ["**/*.pptx"]`, which existed only to load the deleted fixtures.
- **`package.json`** — the two dependencies above.

## Checks

Run locally, all passing:

- `pnpm format`
- `pnpm check-types` — 14/14
- `pnpm turbo run lint` — 13/13 (caught the stale i18n exclusions; fixed here)
- `pnpm knip` — clean, including after removing the entry override
- `pnpm -F @okouai/app exec vitest run src/lib` — 22 passed
- `pnpm -F @okouai/app exec vitest run src/signals` — 201 passed

`src/views` exceeded my local run window and is not covered above; it is a React component suite with no relationship to a deleted, unimported lib module, and CI runs it.

[...]

Issue #28252, opened 2026-08-20
refactor: import a presentation template through an ordinary chat thread


bingjiez666:
Part of #21350. Supersedes #26576, #26237, #26239, #28153.

## Why this replaces the import pipeline

The current design gives template import its own upload protocol: open an import, ask it for one upload slot per file, PUT each file, then commit so the API can verify the set and launch a headless analysis run. It works, but it is a second file-delivery system living next to the one vm0 already has, and the analysis it starts is invisible — the user watches a spinner.

Importing a deck is not special. It is "here is a file, do something with it", which is what a chat thread already is.

## Target flow

```
user picks a .pptx / .pdf in the dialog
  -> ordinary chat attachment upload
  -> create a thread, send one message with the file attached
  -> navigate into that thread
  -> the agent analyses it, in the open, and publishes the template
```

Identical in effect to the user opening a new chat, dropping the deck in, and typing "turn this into a template" — because that is what it is.

Everything the browser side needs already exists: `.pptx` and `application/pdf` are in the upload allowlist (`uploads-constants.ts`), the cap is 1 GB, and chat attachments already reach the sandbox.

## Page images are rendered in the sandbox

Appearance evidence still matters — whitespace, density, decoration, image/text relationships are not recoverable from OOXML. Under this flow the run produces them itself.

Measured in a run sandbox, on `supported.pptx`:

| driver | result |
| --- | --- |
| `agent-browser` + prebuilt bundle | 3 slides, 1600x900, **887 ms**, zero runtime install |
| `puppeteer-core` + `npm i` | 3 slides, 1600x900, 1757 ms, 113 MB of node_modules |

The first is the one to ship. `/usr/bin/chromium` is present in the run image, `agent-browser` can `eval` and `screenshot`, and the renderer is the same `@aiden0z/pptx-renderer@1.2.4` the browser preview pins — so a page image is the same pixels the preview would have produced, without the preview.

The renderer bundle is prebuilt and committed to the skill: 1.07 MB minified, 346 KB gzipped. Nothing is installed at run time.

Limits carry over unchanged: 100 pages, 100 MB source. `too-many-pages.pptx` is rejected with `The presentation has 101 slides; the limit is 100`.

### PDF

In scope for the first version, unlike #26576 which excluded it. The same harness renders it; only the parse step differs. Poppler is not in the image (`pdftoppm` is absent), so PDF goes through pdf.js in the same bundle.

## What the API keeps

One endpoint.

| removed | reason |
[...]

PR #28298, opened 2026-08-20
feat(api): publish an analysed deck as a presentation template

bingjiez666:
Part of #28252. Stacked conceptually after #28280, but touches no shared files, so it can go in either order.

## The whole endpoint

```
POST /api/okou/presentation-templates
  { title, sourceFileId, pageFileIds[], packageFileId }
→ 200 { id, title, status: "ready", pageCount, coverUrl, ... }
```

One call, everything at once. It replaces the import protocol that #26576 specified — open an import, request a slot per file, PUT each, commit — which is now closed unmerged.

## Why one call removes the hazard the old protocol was built for

The server-owned import existed because a browser held the source/pages pairing in a variable across a long upload sequence, so a stale client could pair one deck's source with another deck's pages, silently, and every extracted conclusion would be misaligned against the wrong slide.

Here there is no cross-request grouping to go stale. A single CLI invocation uploads the set and names it in one body. Page order is array order.

## Ownership comes from the object, not the body

`resolveUploads$` resolves each id through `resolveArtifactObject$` — the same path `/uploads/complete` uses — which only returns an object whose *stored metadata* names this user.

Deliberately **not** `run_uploaded_files`: that table is written only when the uploader had a `runId`, so resolving through it silently rejects anything uploaded from a browser session. The first draft did exactly that and the tests caught it.

## Lifecycle

The row is created already `ready`. `pending` and `processing` become unreachable, because nothing exists before a package has been validated. A failed analysis leaves no half-built template — only the chat thread that explains why, which the user can read and retry from. That is also why there is no `fail` endpoint.

## Package validation rejects, never sanitises

A rejected package should be obvious, not quietly reshaped:

- traversal (`..`), absolute paths, backslashes, empty segments
- non-regular entries — directories carry no content, and links are how an archive escapes its root
- duplicate paths, >200 files, >25 MB per file, >100 MB total
- missing or empty `SKILL.md` / `design-system.md`

The archive arrives as a `.tar.gz` through the ordinary upload route, so binary assets are never base64 in JSON. `storage-volume-publication` now accepts `Buffer` content alongside `string`, which is what lets logos and fonts survive; it previously did `Buffer.from(content, "utf8")` unconditionally.

## PDF

Accepted alongside PPTX. Both render to page images the same way (`vm0-ai/Template-artifact#18`), so nothing downstream distinguishes them.

## CLI

```bash
okou presentation-template publish \
  --title "Brand system" \
  --source deck.pptx \
  --pages ./pages \
  --package ./package
```

Uploads the deck, each page in filename order, and a tar of the package directory, then publishes. Page order comes from the renderer's zero-padded filenames, which sort into page order.

## Tests

Four integration cases through the real route, real database, and the real tar reader (the fixture hand-builds a ustar archive rather than mocking the parser):

- publishes a ready template, cover and page URLs resolve, and the package lands under `presentation-template@{id}`
- a package missing `design-system.md` is rejected **and creates no row**
- `../escaped.md` is rejected
- another user's upload ids are rejected

## Checks

`pnpm format`, `pnpm check-types` (14/14), `pnpm turbo run lint` (13/13), `pnpm knip` — all clean. Lint caught three real issues in the first draft: a `try/catch` that should be `settle`/`safeSync`, two missing `throwIfAborted`, and a mutable package-scope fixture.

Tests run: the four above, plus `presentation-templates`, `uploads-prepare`, `cron-sync-skills` (the other consumer of volume publication), and the CLI command-surface test.


bingjiez666:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 5714a32f9a66022ca798c3d3ca487a3e1db42401

### Summary

Adds `POST /api/okou/presentation-templates`, which turns an already-uploaded deck, its ordered page images, and a `.tar.gz` guidance package into a `ready` presentation template in one call, plus the `okou presentation-template publish` CLI that assembles and sends that set. Ownership is resolved per upload id through `resolveArtifactObject$` rather than taken from the request body, package contents are rejected rather than sanitised, and `storage-volume-publication` is widened to accept `Buffer` so binary assets survive. The design is sound and the one-call shape genuinely removes the cross-request pairing hazard the retired import protocol was built for.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | `feat:` commits must ship integration tests | Finding (P0) | `turbo/apps/cli/src/commands/presentation-template/index.ts` is a new user-facing entry point with no `__tests__/`; only the command-count assertion in `okou.test.ts` changed. |
| `REVIEW.md` | Security — no missing auth on new endpoints | Pass | `templatePublishAuth` requires an organization and `presentation-template:write`; `resolveUploads$` re-derives ownership from stored object metadata, so a stranger's ids are rejected (covered by the fourth test). |
| `REVIEW.md` | Security — resource exhaustion on caller-supplied input | Finding (P1) | `presentation-template-publish.service.ts:292` calls `gunzipSync(archive)` with no `maxOutputLength`. |
| `REVIEW.md` | Type safety — no `any`, no broad casts | Pass | No `any` in the diff; the two `uploads.get(...)!` assertions are provably safe because `missing` already proved every id in `ids` resolved. |
| `REVIEW.md` | New user-facing features gated behind a `FeatureSwitchKey` | Pass | `publishInner$` returns `presentationTemplatesDisabled` unless `presentationTemplatesEnabled$`; `FeatureSwitchKey.PresentationTemplates` is `enabled: false` with no staff hashes, so the route is unreachable in production today. |
| `REVIEW.md` | Feature regression guard — existing behavior not silently changed | Pass | `PRESENTATION_TEMPLATE_SOURCE_CONTENT_TYPE` → `..._TYPES` has zero remaining references repo-wide; `VolumeFileInput.content` is widened, not narrowed, so existing `string` callers such as `cron-sync-skills` are unaffected. |
| `REVIEW.md` | DB/JSONB review gate | Pass | No schema or migration change. `pageKeys` is written in the existing jsonb shape; no old row can violate it. |
| `docs/bad-smell.md` | No lint/type suppressions, no hardcoded config | Pass | No suppression comments; the bucket comes from `env("R2_USER_ARTIFACTS_BUCKET_NAME")`. |
| `docs/bad-smell.md` | Error handling — no defensive try/catch, fail fast | Pass | `safeSync`/`settle` are used only where a caller-supplied archive can genuinely be malformed; the unreachable `!created` and missing-row cases throw rather than degrade. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | N/A | The `load: async () => (await import(...))` entry in `okou.ts` is the established shape of every entry in `COMMAND_DEFINITIONS`; the diff adds a row to an existing registry rather than introducing the pattern. |
| `docs/fallback.md` | Every fallback introduced by the diff is declared and justified | Pass | The diff introduces no cross-version fallback. `Buffer.isBuffer(file.content) ? ... : ...` is a live type-union branch for two current callers, and `coverKey ? ... : null` is index-access narrowing, not a compatibility branch — so no `Fallbacks` section is required. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | All four new tests assert current behavior; none pin a deletion. |
| `docs/deployment-compatibility.md` | Cross-version boundary changes are additive | Pass | New POST route, new exported constants, and a new capability string are all additive. Capabilities are not a PG enum column, so no migration ordering risk applies. |
| `docs/testing.md` | Integration tests only; mock at the boundary | Pass | The route test mocks only `context.mocks.s3` and `context.mocks.clerk`; no `vi.mock()` of internal paths, no fake timers, no fetch stubbing. |
| `docs/testing/api-testing.md` | Route tests go through `setupApp()` + the ts-rest contract | Pass | `setupApp({ context, routes: presentationTemplatesRoutes })(presentationTemplatesContract)` with `accept()` narrowing and `testContext()` at module scope. |
| `docs/testing/testing-external-behavior.md` | Construct and verify state through the API, not the DB or services | Pass | Uploads are created through the real prepare/complete upload APIs, and results are verified through `publish`, `list`, and `get` rather than DB reads. |
| `docs/testing/cli-testing.md` | Each command file has a matching neighboring command test entered via `parseAsync()` | Finding (P0) | `src/commands/presentation-template/` contains only `index.ts`; contrast `src/commands/host/__tests__/{publish,clone,versions}.test.ts`. |
| `.claude/skills/ccstate/SKILL.md` | Signals directory uses `command`/`computed` correctly, no manual cache, no `void`-ed promises | Pass | `publishInner$` and `publishPresentationTemplate$` are `command`s taking an explicit `AbortSignal` with `throwIfAborted()` after each await point; no manual memo layer and no detached work in the signals directory. |

### Findings

- `turbo/apps/cli/src/commands/presentation-template/index.ts` and `turbo/apps/cli/src/lib/api/domains/presentation-templates.ts`: the new `okou presentation-template publish` command ships with **no integration test**. `docs/testing/cli-testing.md` requires a neighboring `__tests__/index.test.ts` entered through `command.parseAsync()` with MSW, and the sibling `host` command follows exactly that shape. The only CLI assertion this PR adds is that the public command count moved from 38 to 39, which proves the command is registered and nothing about what it does.

  This matters more than a generic coverage gap here. The PR's central argument is that a single invocation cannot pair one deck's source with another deck's pages — but the *ordering* of those pages now rests entirely on `orderedPagePaths()`, which filters `.png` case-insensitively and sorts with `localeCompare`. Nothing tests it. `localeCompare` is locale-sensitive and is not the same ordering as a byte-wise sort; a mis-sort silently misaligns every extracted conclusion against the wrong slide, which is precisely the failure mode the design set out to eliminate. The API tests cannot cover this, because they construct page ids directly.

  Please add a command test covering at least: page ordering from zero-padded filenames, the non-`.png` filter, the `NO_PAGES` error on a directory with no page images, and the resulting request body reaching the mocked endpoint in that order.

- `turbo/apps/api/src/signals/services/presentation-template-publish.service.ts:292`: `gunzipSync(archive)` is called without `maxOutputLength`, so decompression is unbounded. `downloadS3BufferWithMaxBytes` caps the *compressed* archive at `MAX_PRESENTATION_TEMPLATE_PACKAGE_BYTES` (100 MB), but a 100 MB gzip can expand toward Node's `kMaxLength` (~2 GiB) inside the API process before `safeSync` ever sees `ERR_BUFFER_TOO_LARGE`. `checkPackage()`'s 200-file and 25 MB-per-file limits run only *after* the whole archive is decompressed and every entry is buffered, so they cannot bound the allocation — and they would in any case admit ~5 GB of decompressed content.

  This repo already has the correct pattern for exactly this input class in `connector-catalog-artifacts/loader.ts:280-294`, which passes `maxOutputLength: CONNECTOR_CATALOG_MAX_RAW_BYTES` and maps `ERR_BUFFER_TOO_LARGE` to a distinct `object-too-large` outcome. Mirroring it here — bounding output and returning a specific "package is too large uncompressed" 400 rather than the generic unreadable-archive message — closes it.

  Noting reachability honestly: `FeatureSwitchKey.PresentationTemplates` is `enabled: false` with no `enabledOrgIdHashes`, so nobody can reach this route in production right now. That is what keeps this P1 rather than P0. The concern is that the memory pressure lands in the shared API process, so it stops being contained the moment the switch is turned on.

### Testing

- **Coverage: Insufficient.** The API route is well covered — four integration tests through the real route, real database, and a hand-built ustar archive that exercises the real tar reader rather than a mocked parser, including the two rejection paths most worth pinning (missing `design-system.md` creating no row, and `../escaped.md`) and the cross-user ownership check. The CLI surface has none; see the P0 above.
- **Conventions: Compliant.** Mocks stay at the boundary (`context.mocks.s3`, `context.mocks.clerk`) with no internal `vi.mock()`, no fake timers, no direct `fetch` stubbing, and no DB reads or service calls used for setup or assertions. Setup goes through the real upload APIs and verification goes through `publish`/`list`/`get`, matching `docs/testing/testing-external-behavior.md`.
- One minor nit, not blocking: the `expect(storageName).toBe(...)` assertion restates `getPresentationTemplateStorageName`'s own implementation, so it cannot fail independently of the function it calls. The adjacent `fixture.keys()` assertion is the one carrying real signal.

bingjiez666:
Addressed both findings in `2cbb1b3`.

**P1 — unbounded decompression.** `gunzipSync` now passes `maxOutputLength: MAX_PRESENTATION_TEMPLATE_PACKAGE_BYTES`, so the 100 MB cap covers the decompressed output rather than only the compressed download, and `ERR_BUFFER_TOO_LARGE` is reported as its own 400 ("must unpack to N bytes or fewer") instead of being folded into the generic unreadable-archive message. This also closes the secondary gap where `checkPackage`'s per-file and file-count limits would have admitted far more than 100 MB in aggregate.

Not covered by a test: exercising the cap requires materializing >100 MB of decompressed content in the API test process, which is a poor trade for a bound whose boundary behavior is Node's documented `ERR_BUFFER_TOO_LARGE`. Flagging that explicitly rather than leaving it implied.

**P0 — untested CLI surface.** Added `src/commands/presentation-template/__tests__/index.test.ts`, entered through `presentationTemplateCommand.parseAsync()` with MSW standing in for prepare/PUT/complete/publish, per `docs/testing/cli-testing.md`. It covers page ordering from zero-padded filenames written to disk out of order, the case-insensitive `.png` filter alongside a non-page file, the `NO_PAGES` error on a directory with no page images, and a rejected publish surfacing the server message without printing success. The uploaded package body is asserted to be a real gzip so `packageArchive` is exercised rather than assumed.

While adding it I also fixed the underlying ordering risk instead of only pinning current behavior: `orderedPagePaths` sorted with `localeCompare`, which follows the host locale, so the same directory could publish in a different order on a different machine. It now compares by code unit, which is the ordering the zero-padded filenames were designed for.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 72827396733a4addf64e08275a6ca8bdc4e47843
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Adds `POST /api/okou/presentation-templates`, which turns an already-uploaded deck, its ordered page images, and a `.tar.gz` guidance package into a `ready` presentation template in one call, plus the `okou presentation-template publish` CLI that assembles and sends that set. Ownership is resolved per upload id through `resolveArtifactObject$` rather than taken from the request body, package contents are rejected rather than sanitised, and `storage-volume-publication` is widened to accept `Buffer` so binary assets survive.

Both findings from the review of `5714a32f` are resolved in `2cbb1b3` and `7282739`: decompression is now bounded, and the CLI surface has integration tests.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | `feat:` commits must ship integration tests | Pass | Five API route tests plus `commands/presentation-template/__tests__/index.test.ts` entered through `parseAsync()`. |
| `REVIEW.md` | Security — no missing auth on new endpoints | Pass | `templatePublishAuth` requires an organization and `presentation-template:write`; `resolveUploads$` re-derives ownership from stored object metadata, covered by the stranger-ids test. |
| `REVIEW.md` | Security — resource exhaustion on caller-supplied input | Pass | `gunzipSync` now passes `maxOutputLength: MAX_PRESENTATION_TEMPLATE_PACKAGE_BYTES` and maps `ERR_BUFFER_TOO_LARGE` to its own 400; covered by the zip-bomb test. |
| `REVIEW.md` | Type safety — no `any`, no broad casts | Pass | No `any`; the two `uploads.get(...)!` assertions are proven safe by the preceding `missing` check. Test-only `as` casts on `await request.json()` match the existing CLI test convention. |
| `REVIEW.md` | New user-facing features gated behind a `FeatureSwitchKey` | Pass | `publishInner$` returns `presentationTemplatesDisabled` unless `presentationTemplatesEnabled$`; `FeatureSwitchKey.PresentationTemplates` is `enabled: false` with no staff hashes. |
| `REVIEW.md` | Feature regression guard — existing behavior not silently changed | Pass | `PRESENTATION_TEMPLATE_SOURCE_CONTENT_TYPE` → `..._TYPES` has zero remaining references repo-wide; `VolumeFileInput.content` is widened, not narrowed, so `cron-sync-skills` and other `string` callers are unaffected. |
| `REVIEW.md` | DB/JSONB review gate | Pass | No schema or migration change; `pageKeys` uses the existing jsonb shape, and capabilities are not a PG enum column, so no migration ordering risk. |
| `docs/bad-smell.md` | No lint/type suppressions, no hardcoded config | Pass | No suppression comments; the bucket comes from `env("R2_USER_ARTIFACTS_BUCKET_NAME")`. |
| `docs/bad-smell.md` | Error handling — no defensive try/catch, fail fast | Pass | `safeSync`/`settle` guard only the caller-supplied archive, which can genuinely be malformed; `!created` and the missing-row case throw. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | N/A | The `load: async () => (await import(...))` entry in `okou.ts` is the shape of every entry in the existing `COMMAND_DEFINITIONS` registry; the diff adds a row rather than introducing the pattern. |
| `docs/fallback.md` | Every fallback introduced by the diff is declared and justified | Pass | The diff introduces no cross-version fallback. `Buffer.isBuffer(file.content) ? ... : ...` is a live type-union branch with two current callers, and `coverKey ? ... : null` is index-access narrowing — so no `Fallbacks` section is required. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | Every test asserts current behavior; none pins a deletion. |
| `docs/deployment-compatibility.md` | Cross-version boundary changes are additive | Pass | New POST route, new exported constants, and a new capability string are additive; no migration involved. |
| `docs/testing.md` | Integration tests only; mock at the boundary | Pass | API tests mock only `context.mocks.s3`/`clerk`; CLI tests mock only HTTP via MSW. No internal `vi.mock()`, no fake timers, no `fetch` stubbing. |
| `docs/testing/api-testing.md` | Route tests go through `setupApp()` + the ts-rest contract | Pass | `setupApp({ context, routes: presentationTemplatesRoutes })(presentationTemplatesContract)` with `accept()` narrowing and `testContext()` at module scope. |
| `docs/testing/testing-external-behavior.md` | Construct and verify state through the API, not the DB or services | Pass | Uploads are created through the real prepare/complete routes; results are verified through `publish`, `list`, and `get`. The two rejection tests additionally assert `list` is empty, proving no partial row. |
| `docs/testing/cli-testing.md` | Each command file has a matching neighboring command test entered via `parseAsync()` | Pass | `commands/presentation-template/__tests__/index.test.ts` covers page ordering, the `.png` filter, the empty-directory error, and a rejected publish, with a real temp directory and MSW at the HTTP boundary. |
| `.claude/skills/ccstate/SKILL.md` | Signals use `command`/`computed` correctly, no manual cache, no `void`-ed promises | Pass | `publishInner$` and `publishPresentationTemplate$` are `command`s taking an explicit `AbortSignal` with `throwIfAborted()` after each await point; no manual memo layer, no detached work in the signals directory. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

Two items from the previous round are now closed:

- **Unbounded decompression** — `presentation-template-publish.service.ts` bounds `gunzipSync` with `maxOutputLength` and reports the cap as a distinct 400 ("must unpack to N bytes or fewer") rather than folding it into the generic unreadable-archive message. This also closes the secondary gap where `checkPackage`'s per-file and count limits would have admitted far more than 100 MB in aggregate.
- **Untested CLI surface** — the command now has integration tests, and the underlying ordering risk was fixed rather than pinned: `orderedPagePaths` compared with `localeCompare`, which follows the host locale, so the same directory could publish in a different order on a different machine. It now compares by code unit, which is the ordering the zero-padded filenames were designed for.

Three non-blocking observations, recorded rather than raised:

- `publish` declares `404` in its contract responses, but no handler path returns it.
- Passing the same upload id twice in `pageFileIds` would store a duplicate storage key and count it twice; the caller controls this and nothing downstream breaks.
- `expect(storageName).toBe(...)` restates `getPresentationTemplateStorageName`'s own implementation, so it cannot fail independently; the adjacent `fixture.keys()` assertion is the one carrying signal.

### Testing

- **Coverage: Adequate.** Five API integration tests through the real route, real database, and a hand-built ustar archive that exercises the real tar reader: the ready-template happy path, a package missing `design-system.md` that creates no row, `../escaped.md`, an archive that unpacks past the cap, and another user's upload ids. Three CLI command tests cover ordering, the empty-directory error, and a rejected publish.
- **Conventions: Compliant.** Mocks stay at the boundary, setup goes through production APIs, and the CLI test uses a real temp directory with real tar packaging — the uploaded package body is asserted to be a real gzip rather than assumed.
- **Verification caveat, stated plainly:** this working environment cannot reach `registry.npmjs.org` (`424 connector_not_configured_for_run`), so `node_modules` could not be installed and `pnpm format`, `lint`, `check-types`, and `vitest` were not run locally for the two fix commits. The tests added in `2cbb1b3` and `7282739` have never been executed anywhere except CI. This LGTM covers the code as read; the PR's own pipeline is the only execution evidence, and it should be green before merge.

PR #28307, opened 2026-08-20
fix(model-provider): omit custom apply_patch from direct deepseek

### LINKED ISSUE #28274: bug(model-provider): avoid unsupported custom apply_patch on direct DeepSeek
## Summary

Direct DeepSeek rejects a Codex request before model execution because the request includes the native `apply_patch` tool as an OpenAI custom/freeform tool, while the provider's downstream SGLang Chat Completions validation accepts function tools only.

This is a provider-capability mismatch encoded in vm0's DeepSeek model catalog, not a runner-owned tool serialization bug. The immediate fix should make the direct DeepSeek catalog projection match the provider's supported tool schema without changing compatible providers or introducing request/response translation in the runner or MITM proxy.

## Confirmed responsibility boundary

- vm0 configures the direct `deepseek` Codex provider with `wireApi: "responses"` and allows the provider's `/responses` endpoint.
- `turbo/packages/api-contracts/src/contracts/deepseek-model-catalog.json` declares `"apply_patch_tool_type": "freeform"`.
- The runner artifact pins Codex CLI `0.148.0`. In that version, `ApplyPatchToolType` only has the `Freeform` variant, and Codex serializes the native tool as `type: "custom"` with a Lark grammar.
- DeepSeek's documented Chat Completions tool contract accepts `type: "function"` tools only.
- The guest agent supplies provider and model-catalog configuration to Codex; it does not construct the outgoing `tools` array. A generic runner/MITM rewrite would also need to translate returned tool-call items and raw patch arguments, so it is not a safe compatibility fix.

The SGLang stack reports `POST /v1/chat/completions` even though vm0 selects the Responses wire API. The best-supported explanation is that the provider's Responses surface translates internally to its SGLang Chat Completions backend. This is an inference; the representative run's resolved route should still be confirmed from production run metadata.

## Production evidence

Production queries covering `2026-07-20T00:00:00Z` through `2026-08-20T03:00:00Z` found the same validation error in two nested operation records for one failed run, not two independent failures:

| UTC time | operation | run ID | duration | success |
|---|---|---|---:|---|
| 2026-08-19T12:49:11.695Z | `cli_execution` | `521f8c21-7c24-4c51-bb94-a12bd8b2d966` | 456,808 ms | `false` |
| 2026-08-19T12:49:12.184Z | `agent_execute` | `521f8c21-7c24-4c51-bb94-a12bd8b2d966` | 457,364 ms | `false` |

The matching terminal record was:

```text
_time:   2026-08-19T12:49:12.202Z
level:   error
context: runner::cmd::start::job_terminal_log
message: job execution failed
run_id:  521f8c21-7c24-4c51-bb94-a12bd8b2d966
```

The sanitized provider error identifies the incompatible entry:

```text
loc: ('body', 'tools', 4, 'function')
msg: Field required
input:
  type: custom
  name: apply_patch
  format:
    type: grammar
    syntax: lark

File "/sgl-workspace/sglang/python/sglang/srt/entrypoints/http_server.py", line 1535
POST /v1/chat/completions
```

The operation durations are the surrounding run duration. They do not show that SGLang request validation itself took about 7.6 minutes.

PR #28197 restored VM0-managed DeepSeek direct routing at `2026-08-19T11:49:36Z`; the representative failure occurred about one hour later. This is strong correlation, but the run's selected model/provider still needs direct metadata confirmation.

## Problem

When Codex includes the native freeform `apply_patch` tool on the direct DeepSeek route, the provider rejects the entire request during schema validation and no model response is produced. A capability that the direct provider cannot accept is currently advertised through vm0-owned model metadata.

The current catalog is also reused as source metadata for compatible Responses gateways. Removing `apply_patch_tool_type` from the shared source record would unnecessarily degrade those routes, so the compatibility adjustment must be scoped to the direct DeepSeek runtime projection.

## Chosen direction

1. Derive a direct-DeepSeek model catalog projection that omits `apply_patch_tool_type` from every projected DeepSeek model record.
2. Use that restricted projection only in the static direct `deepseek` Codex runtime config, covering both VM0-managed direct routing and direct DeepSeek BYOK.
3. Keep the full source DeepSeek catalog, including its freeform `apply_patch` metadata, for provider paths that support OpenAI custom tools, such as compatible custom Responses gateway projections.
4. Preserve all existing function tools and provider behavior outside the direct DeepSeek route.
5. Do not add a runner/MITM request rewrite, provider-version detection, automatic cross-provider fallback, or a new persisted/runtime protocol field.

Direct DeepSeek will temporarily lose the native first-class `apply_patch` tool. File editing must continue through available shell/`exec_command` behavior. Native `apply_patch` can be restored when either the pinned Codex version supports a function-tool representation or the direct provider supports OpenAI custom tools.

Provider error classification and richer terminal diagnostics are useful follow-up work, but they do not resolve this incompatibility and are outside this issue's functional fix.

## Acceptance criteria

- Confirm the representative run's resolved model and provider route from production metadata.
- The direct `deepseek` Codex runtime catalog omits `apply_patch_tool_type` for both DeepSeek V4 Flash and Pro.
- The source catalog used by compatible custom Responses gateway projections retains `apply_patch_tool_type: "freeform"`.
- Existing standard function-tool metadata and behavior remain unchanged.
- Contract coverage verifies the direct-provider restriction and the compatible-gateway preservation independently, preventing the two catalog projections from drifting accidentally.
- A focused, credentialed direct DeepSeek Codex smoke performs a file-editing task without the SGLang missing-`function` validation error. The test must not add artificial sleeps or increase unrelated timeouts.
- Production verification confirms representative direct DeepSeek runs no longer emit this validation signature after rollout.

## Compatibility and rollout

- No database migration, public API change, runner protocol change, or persisted-state version is required.
- New runs receive the restricted direct-provider catalog after the API deployment.
- Already-created run snapshots may retain the old catalog and can still encounter the failure until they drain.
- Independently deployed old runners can consume the adjusted catalog because the existing payload shape is unchanged.
- Rollback is limited to restoring the prior direct-provider catalog projection.

## Risks

- Direct DeepSeek loses native `apply_patch` selection, grammar constraints, patch streaming, and related first-class tool UX. Shell-based editing may be less reliable and must be covered by the smoke test.
- Separating the source and direct-provider catalog projections creates a small drift risk; table-driven contract coverage should assert both from the same source records.
- Existing queued or snapshotted runs can retain the incompatible metadata during deployment overlap.

## References

- Codex pin: `crates/runner/scripts/build-template.sh`
- Direct provider metadata: `turbo/packages/api-contracts/src/contracts/model-providers.ts`
- DeepSeek model catalog: `turbo/packages/api-contracts/src/contracts/deepseek-model-catalog.json`
- DeepSeek tool contract: https://api-docs.deepseek.com/api/create-chat-completion/
- Pinned Codex tool enum: https://github.com/openai/codex/blob/rust-v0.148.0/codex-rs/protocol/src/openai_models.rs
- Pinned Codex freeform tool schema: https://github.com/openai/codex/blob/rust-v0.148.0/codex-rs/core/src/tools/handlers/apply_patch_spec.rs
- Direct-routing restoration: #28197
- Different DeepSeek failure mode: #27775 tracks malformed/truncated tool-call JSON returned by a gateway.

---
> Compacted on 2026-08-20 from 0 comments and the repository/protocol investigation.

seven332:
Closes #28274

## Problem

VM0's direct DeepSeek Codex configuration advertised `apply_patch_tool_type: "freeform"`. Codex serialized that as an OpenAI custom tool, but DeepSeek's SGLang-compatible endpoint only accepts function tools, so affected runs could fail validation before inference.

## Solution

- derive a restricted direct DeepSeek model catalog that omits `apply_patch_tool_type`
- retain the full DeepSeek metadata as the explicit source catalog for custom Responses gateways
- cover managed direct, direct BYOK, and custom gateway projections at their existing integration entry points

This does not change routing, runner or MITM behavior, persisted schemas, or API protocols, and it does not add a slow end-to-end suite.

## Validation

- `pnpm -F @okouai/api-contracts exec vitest run src/contracts/__tests__/model-providers.test.ts --maxWorkers=1 --silent=passed-only` (165 passed)
- `pnpm -F api exec vitest run src/signals/routes/__tests__/model-provider-gateways.test.ts --maxWorkers=1 --silent=passed-only` (7 passed)
- `pnpm -F api exec vitest run src/signals/routes/__tests__/chat-events.bdd.test.ts --maxWorkers=1 --silent=passed-only -t 'routes DeepSeek V4 Flash through the native Responses adapter'` (1 passed)
- `pnpm -F api exec vitest run src/signals/routes/__tests__/run-lifecycle.bdd.test.ts --maxWorkers=1 --silent=passed-only -t 'claims vm0 .* runs with the Responses adapter'` (2 passed)
- package lint and type checks for `@okouai/api-contracts` and `api`
- scoped Knip plus pre-commit full-repository Knip and type checks
- `git diff --check`

## Risk and rollout

- direct DeepSeek Codex runs no longer receive the native `apply_patch` tool; independent shell/exec tools remain available
- already-created run snapshots may retain the previous catalog, while newly created run contexts receive the restricted catalog
- custom Responses gateways continue to receive the full metadata, including freeform `apply_patch`
- a live credentialed Codex smoke test and production-log confirmation remain rollout checks because the local environment does not expose the required DeepSeek runner credential or `AXIOM_TOKEN`

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts
@@ -3996,6 +3996,14 @@ describe("CHAT-02: model-first provider policies", () => {
         ]),
       },
     });
+    const directCatalogModels = claim.codexRuntimeConfig?.modelCatalog?.models;
+    if (!Array.isArray(directCatalogModels)) {
+      throw new Error("Expected a direct DeepSeek model catalog");
+    }
+    expect(directCatalogModels).toHaveLength(2);
+    for (const model of directCatalogModels) {
+      expect(model).not.toHaveProperty("apply_patch_tool_type");
+    }
 
     chatCallbacks.mockChatOutputEvents([]);
     await completeChatRunOk(run.runId, sandboxHeaders, {
--- a/turbo/apps/api/src/signals/routes/__tests__/model-provider-gateways.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/model-provider-gateways.test.ts
@@ -568,6 +568,7 @@ describe("custom model provider gateway routes", () => {
       }
       expect(catalogModel).toMatchObject({
         slug: upstreamModel,
+        apply_patch_tool_type: "freeform",
         input_modalities: ["text"],
         base_instructions: expect.stringContaining("You are Codex"),
         model_messages: {
--- a/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
@@ -7027,6 +7027,9 @@ describe("RUN-02: model provider selection and vm0 admission", () => {
           `Expected a native DeepSeek Codex catalog for ${selectedModel}`,
         );
       }
+      for (const model of catalogModels) {
+        expect(model).not.toHaveProperty("apply_patch_tool_type");
+      }
       expect(catalogModels).toContainEqual(
         expect.objectContaining({
           slug: selectedModel,
--- a/turbo/packages/api-contracts/src/contracts/__tests__/model-providers.test.ts
+++ b/turbo/packages/api-contracts/src/contracts/__tests__/model-providers.test.ts
@@ -5,6 +5,7 @@ import {
   hasModelSelection,
   getModels,
   getDefaultModel,
+  getModelProviderCodexCatalogForModel,
   getModelProviderCodexRuntimeConfig,
   getModelProviderEnvBindings,
   getFrameworkForType,
@@ -839,7 +840,8 @@ describe("deepseek Responses provider", () => {
   });
 
   it("configures the official DeepSeek Responses model catalog", () => {
-    expect(getModelProviderCodexRuntimeConfig("deepseek")).toMatchObject({
+    const runtimeConfig = getModelProviderCodexRuntimeConfig("deepseek");
+    expect(runtimeConfig).toMatchObject({
       providerId: "deepseek",
       name: "DeepSeek",
       baseUrl: "https://api.deepseek.com/",
@@ -875,7 +877,36 @@ describe("deepseek Responses provider", () => {
         ],
       },
     });
+    const directModels = runtimeConfig?.modelCatalog?.models;
+    if (!Array.isArray(directModels)) {
+      throw new Error("Expected a direct DeepSeek model catalog");
+    }
+    expect(directModels).toHaveLength(2);
+    for (const model of directModels) {
+      expect(model).not.toHaveProperty("apply_patch_tool_type");
+    }
   });
+
+  it.each([
+    ["deepseek-v4-flash", "gateway/deepseek-flash"],
+    ["deepseek-v4-pro", "gateway/deepseek-pro"],
+  ])(
+    "retains native apply_patch metadata when projecting %s for a compatible gateway",
+    (logicalModel, runtimeModel) => {
+      expect(
+        getModelProviderCodexCatalogForModel(logicalModel, runtimeModel),
+      ).toMatchObject({
+        models: [
+          expect.objectContaining({
+            slug: runtimeModel,
+            apply_patch_tool_type: "freeform",
+            default_reasoning_level: "high",
+            context_window: 1_048_576,
+          }),
+        ],
+      });
+    },
+  );
 
   it("scopes the firewall to the native Responses endpoint", () => {
     const config = MODEL_PROVIDER_FIREWALL_CONFIGS.deepseek;
--- a/turbo/packages/api-contracts/src/contracts/model-providers.ts
+++ b/turbo/packages/api-contracts/src/contracts/model-providers.ts
@@ -41,6 +41,15 @@ const DEEPSEEK_MODEL_CATALOG = {
   ],
 };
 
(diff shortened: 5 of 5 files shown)


seven332:
## Code Review: PR #28307

**Reviewed head:** `b2bf104a2d82615918865ac8e0d31d5a520fa5c0`

### Summary

The change scopes DeepSeek's unsupported custom/freeform `apply_patch` capability out of direct managed and BYOK runtime catalogs while preserving the full source metadata for compatible custom Responses gateways. The implementation follows the approved direct-only projection and keeps the existing payload shape.

### Findings

No findings.

### Testing

Contract coverage verifies Flash and Pro in both projections. Existing API integration entry points verify managed direct, direct BYOK, and custom gateway claim payloads. Focused tests, affected-workspace lint/type checks, scoped Knip, and pre-commit full Knip/type checks passed. A credentialed direct Codex file-edit smoke and post-rollout production-log confirmation remain unverified because the required local credential and `AXIOM_TOKEN` are unavailable.

### Verdict

LGTM

seven332:
Closing because production telemetry showed that the representative run was created by API commit 2c3e504aa7b2d69d19a8082b18972a143e931b98, whose managed DeepSeek mapping still used openrouter-codex. The direct-routing restoration in #28197 was not part of that deployed API commit. Current direct DeepSeek Responses supports the Codex custom apply_patch tool, so removing it from the direct catalog would be a regression. The incident path should instead be resolved by deploying and validating #28197; any remaining OpenRouter compatibility handling must be scoped to the OpenRouter route.

PR #28318, opened 2026-08-20
fix(sandbox-fc): bound firecracker child file descriptor tables

### LINKED ISSUE #28285: bug(runner): bound firecracker child file descriptor tables
## Problem

Firecracker processes launched directly by `sandbox-fc` inherit the runner systemd service's
`RLIMIT_NOFILE=524288`. Firecracker v1.15.1 deliberately expands its kernel file-descriptor table
to the inherited soft limit after logging `Running Firecracker` and before binding its API Unix
socket. The upstream jailer instead sets both the soft and hard `RLIMIT_NOFILE` values to `2048`.

The direct-launch inheritance therefore makes every Firecracker process preallocate capacity for
524,288 descriptors even though production microVMs use far fewer descriptors. This creates
avoidable kernel allocation and RCU work on the startup path and is the strongest supported cause
of the observed API-socket readiness stalls.

## Production evidence

Two independent runs failed on 2026-08-19 after the five-second API readiness deadline:

| UTC timestamp | Beijing timestamp | run ID | sandbox ID |
| --- | --- | --- | --- |
| `2026-08-19T00:35:08.757Z` | `2026-08-19 08:35:08.757` | `34539e16-21ac-4fcf-91e5-3031be31a733` | `8daf2339-db4c-42fb-abc9-8c5cb82906ac` |
| `2026-08-19T02:20:48.995Z` | `2026-08-19 10:20:48.995` | `2023a617-af43-44bd-8879-8d0fda98b92c` | `97e578ab-76fe-48d1-beb7-4198d12d2351` |

The bounded production query was:

```apl
['vm0-sandbox-op-log-prod']
| where error contains "boot_mode=snapshot restore"
    and error contains "timed out after 5s waiting for socket file"
| project _time, op_type, sandbox_type, source, run_id, duration_ms, success, error
| sort by _time asc
```

It returned four operation records for the two runs. Both jobs failed before guest execution.

Host journal inspection established that both failures:

- were claimed by `prod-3.aws.vm3.ai`, runner `v0.169.9`, runner ID
  `c53203fd-66ab-40c6-899d-3fb9da2befe5`, process PID `4133040`;
- launched Firecracker v1.15.1 and logged `Running Firecracker` about 16 ms after spawn;
- never logged `Listening on API socket`, API activity, stderr, or an early process exit;
- remained alive until the runner's five-second readiness timeout cleaned them up; and
- had no matching host OOM, hung-task, or I/O-error event.

`prod-4.aws.vm3.ai` and `prod-11.gcp.vm3.ai` received the broadcast notifications but did not claim
the jobs. Snapshot restores immediately before and after each failure succeeded on `prod-3`; the
normal `Running Firecracker` to `Listening on API socket` interval was 11–44 ms. The restore after
the second failure took 330 ms for the same interval, confirming transient long-tail startup
contention. The failure occurs before `PUT /snapshot/load`, so snapshot corruption is not implicated.

All three inspected runner hosts expose a finite inherited `RLIMIT_NOFILE` of `524288`. A live
sample of every Firecracker process on those hosts observed maximum actual descriptor counts of
58, 58, and 61 respectively, leaving more than 30x headroom under the upstream jailer default.

## Decision

Bound `RLIMIT_NOFILE` for every Firecracker subprocess launched by `sandbox-fc` to the upstream
jailer default of `2048`, setting both the soft and hard limits before `exec`. Preserve the runner
service's `524288` limit because the long-lived runner aggregates descriptors for many sandboxes.

Apply the invariant consistently to fresh boot, snapshot restore, and snapshot creation. Do not
lengthen the API readiness timeout or add retry/fresh-boot fallback in this change. Those approaches
hide the resource mismatch, add latency or lifecycle complexity, and are not justified before
removing the concrete pre-socket allocation hazard.

## Chosen implementation direction

Create one private Firecracker process-spawn boundary in `sandbox-fc`. It accepts a prepared Tokio
command and working directory, applies the process settings already duplicated by the two current
spawn owners, installs a child-only `pre_exec` hook that sets both `RLIMIT_NOFILE` values to 2048,
and calls `spawn()`.

Fresh boot and snapshot restore already share one lifecycle owner; snapshot creation has another.
Both will delegate only final child spawning to the shared boundary. Command construction, runtime
adoption, readiness, diagnostics, cleanup, and error translation stay with their current owners.
This makes the resource invariant structural across all three modes without changing their argv or
lifecycle behavior.

## Alternatives considered

- **Configure the two existing owners separately:** a smaller diff, but the invariant remains a
  convention that a current or future spawn owner can omit, and a standalone helper test would not
  prove both production paths installed it.
- **Wrap commands with `prlimit`:** avoids an unsafe Rust hook but adds a host executable dependency
  and another process to already layered snapshot launch chains, changing error and signal behavior.
- **Increase the five-second timeout:** rejected because healthy socket binding takes milliseconds
  and a larger deadline only delays a terminal failure.
- **Retry snapshot restore or fall back to fresh boot:** rejected because the failure precedes
  snapshot loading and the inherited descriptor-table expansion would affect every retry.
- **Lower the runner service limit:** rejected because it would constrain the long-lived runner,
  not just individual Firecracker processes.
- **Diagnostics only:** rejected because the direct-launch limit mismatch is independently
  actionable and safely removable with large measured headroom.

## Risks and mitigations

- A future Firecracker configuration could need more than 2048 descriptors. The limit matches the
  deployed version's upstream jailer default, and current production processes use at most 61 in
  the sampled fleet.
- Applying a pre-exec resource limit can make spawn fail if the limit cannot be installed. Preserve
  the operating-system error at the existing spawn boundary; never silently launch with 524,288.
- Moving shared process modifiers could alter startup behavior. Move only the settings already
  identical at both owners and keep all lifecycle-specific behavior in place.
- Historical logs cannot prove which kernel operation consumed all five seconds. Treat removal of
  the concrete mismatch as the intervention, then inspect any post-deployment recurrence before
  considering another cause or recovery policy.

## Acceptance criteria

- [ ] Fresh boot, snapshot restore, and snapshot creation all delegate to one Firecracker child
      spawn boundary.
- [ ] Every launched Firecracker subprocess enters `exec` with soft and hard
      `RLIMIT_NOFILE=2048`.
- [ ] The runner process and its systemd `LimitNOFILE=524288:524288` configuration remain unchanged.
- [ ] A real child launched through the production spawn function reports both limits as 2048;
      tests do not merely assert a constant or command representation.
- [ ] Existing API-readiness success, missing-socket, delayed-socket, API-unresponsive, early-exit,
      snapshot diagnostics, and cleanup behavior remains unchanged.
- [ ] No host executable or Rust dependency is added.
- [ ] Applicable `sandbox-fc` formatting, Clippy, documentation, and test checks pass.

## Delivery notes

- This is one independently reviewable Rust PR in `sandbox-fc`; no API, persisted-state, queue, or
  cross-version protocol changes are required.
- Old and new runners can overlap. Existing Firecracker processes keep their current limit; newly
  spawned processes receive 2048 after deployment.
- After deployment, repeat the bounded production query above and inspect any recurrence before
  considering timeout or retry changes.

## Related work

- #22413 centralized Firecracker startup-readiness orchestration.
- #25239 centralized snapshot mount-namespace spawn construction.
- #22764 addressed guest/vsock readiness during startup bursts; it is a different phase.

seven332:
## Summary

- centralize the common process boundary for fresh-boot, snapshot-restore, and snapshot-creation Firecracker children
- set child soft and hard `RLIMIT_NOFILE` values to the Firecracker jailer's `2048` default before exec
- verify the production spawn boundary with a real child that reports its inherited limits

## Explicit exclusions

- keep the runner service's `LimitNOFILE=524288:524288` unchanged
- do not change the five-second API readiness deadline
- do not add launch retries or fresh-boot fallback

## Validation

- `cargo test -p sandbox-fc firecracker_process`
- `cargo clippy -p sandbox-fc --all-targets -- -D warnings`
- `cargo doc -p sandbox-fc --no-deps`
- `cargo test -p sandbox-fc` (807 unit tests, 1 integration test)
- pre-commit hooks: workspace Cargo formatting, documentation, Clippy, and file-size checks

Closes #28285

### DIFF
--- a/crates/sandbox-fc/src/firecracker_process.rs
+++ b/crates/sandbox-fc/src/firecracker_process.rs
@@ -1,0 +1,62 @@
+//! Applies the shared process contract for Firecracker children.
+
+use std::io;
+use std::path::Path;
+use std::process::Stdio;
+
+use tokio::process::{Child, Command};
+
+/// Firecracker's jailer defaults both open-file limits to this value.
+const FIRECRACKER_NOFILE_LIMIT: libc::rlim_t = 2048;
+
+pub(crate) fn spawn_firecracker(mut command: Command, current_dir: &Path) -> io::Result<Child> {
+    command
+        .current_dir(current_dir)
+        .stdin(Stdio::null())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .process_group(0)
+        .kill_on_drop(true);
+
+    // SAFETY: `setrlimit` is async-signal-safe and receives a valid pointer to
+    // child-local stack storage. The hook does not access shared process state.
+    unsafe {
+        command.pre_exec(|| {
+            let limit = libc::rlimit {
+                rlim_cur: FIRECRACKER_NOFILE_LIMIT,
+                rlim_max: FIRECRACKER_NOFILE_LIMIT,
+            };
+            if libc::setrlimit(libc::RLIMIT_NOFILE, &limit) == 0 {
+                Ok(())
+            } else {
+                Err(io::Error::last_os_error())
+            }
+        });
+    }
+
+    command.spawn()
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[tokio::test]
+    async fn spawned_child_inherits_jailer_nofile_limit() {
+        let mut command = Command::new("/bin/sh");
+        command.args(["-c", "ulimit -Sn; ulimit -Hn"]);
+
+        let output = spawn_firecracker(command, Path::new("/"))
+            .unwrap()
+            .wait_with_output()
+            .await
+            .unwrap();
+
+        assert!(
+            output.status.success(),
+            "child failed: {}",
+            String::from_utf8_lossy(&output.stderr)
+        );
+        assert_eq!(String::from_utf8(output.stdout).unwrap(), "2048\n2048\n");
+    }
+}
--- a/crates/sandbox-fc/src/lib.rs
+++ b/crates/sandbox-fc/src/lib.rs
@@ -34,6 +34,7 @@ mod cow_pool;
 mod duration;
 mod exec_operation_result;
 mod factory;
+mod firecracker_process;
 mod guest_dns_failure_diagnostics;
 mod guest_dns_probe;
 mod guest_dns_readiness;
--- a/crates/sandbox-fc/src/sandbox.rs
+++ b/crates/sandbox-fc/src/sandbox.rs
@@ -44,6 +44,7 @@ use crate::exec_operation_result::{
     validate_exec_capture_timeout,
 };
 use crate::factory::InvariantConfig;
+use crate::firecracker_process::spawn_firecracker;
 use crate::guest_dns_failure_diagnostics::{
     GuestDnsFailureDiagnosticContext, capture_guest_dns_failure_diagnostics,
 };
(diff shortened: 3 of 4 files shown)


seven332:
## Code Review: PR #28318

**Reviewed head:** `564a4cd331f2f146afe2fe333df686aba27fbb8c`

### Summary

The shared child spawn boundary preserves both existing launch owners' process, ownership, error,
and cleanup behavior while applying the jailer-aligned descriptor limit to fresh boot, snapshot
restore, and snapshot creation.

### Findings

No findings.

### Testing

The real-child regression test observes soft and hard limits after exec. Focused testing, all 807
`sandbox-fc` library tests, the integration test, Clippy, rustdoc, formatting, and pre-commit checks
passed. Production recurrence monitoring remains a rollout task tracked by #28285.

### Verdict

LGTM

PR #28320, opened 2026-08-20
fix(platform): show pinned agent drag handle only while dragging

Lunarivibe:
## Problem

The six-dot grip above a pinned agent tile was revealed on hover (`group-hover:opacity-100`), so it popped up during ordinary browsing of the pinned agents row — visual noise on a surface that is hovered constantly.

## Change

`PinnedAgentDragHandle` now renders only while a reorder drag is in flight, on the reorderable tiles that can accept the drop. The dragged tile keeps its dashed ghost with no grip, and the lead agent still never shows one.

- `zero-sidebar-pinned.tsx`: gate the handle on `draggingPinnedAgentId$` instead of hover; drop the opacity/hover classes.
- `zero-sidebar.test.tsx`: the handle test now asserts no handles at rest, and handles on the other reorderable tiles after `dragStart`.

Reordering itself is unchanged — the whole tile is the drag source, so nothing depends on grabbing the grip.

## Verification

Local `pnpm` install is blocked in this sandbox (npm registry unreachable), so the platform vitest suite runs in CI.

### DIFF
--- a/turbo/apps/platform/src/views/zero-page/__tests__/zero-sidebar.test.tsx
+++ b/turbo/apps/platform/src/views/zero-page/__tests__/zero-sidebar.test.tsx
@@ -3987,7 +3987,7 @@ describe("zero sidebar", () => {
     });
   });
 
-  it("shows a drag handle on reorderable pinned agents but not on the lead agent", async () => {
+  it("shows drag handles only while a reorder drag is in flight", async () => {
     const pinnedAgentIds = prepareOverflowingPinnedAgents();
     context.mocks.data.userPreferences({ pinnedAgentIds });
 
@@ -4005,10 +4005,23 @@ describe("zero sidebar", () => {
     });
 
     expect(
-      within(pinnedAgentLink(grid, "Support Agent")).getByTestId(
+      within(grid).queryAllByTestId("pinned-agent-drag-handle"),
+    ).toHaveLength(0);
+
+    fireEvent.dragStart(pinnedAgentLink(grid, "Support Agent"), {
+      dataTransfer: createDataTransferStub(),
+    });
+
+    expect(
+      within(pinnedAgentLink(grid, "Operations Agent")).getByTestId(
         "pinned-agent-drag-handle",
       ),
     ).toBeInTheDocument();
+    expect(
+      within(pinnedAgentLink(grid, "Support Agent")).queryByTestId(
+        "pinned-agent-drag-handle",
+      ),
+    ).toBeNull();
     expect(
       within(pinnedAgentLink(grid, "Zero")).queryByTestId(
         "pinned-agent-drag-handle",
--- a/turbo/apps/platform/src/views/zero-page/zero-sidebar-pinned.tsx
+++ b/turbo/apps/platform/src/views/zero-page/zero-sidebar-pinned.tsx
@@ -257,7 +257,9 @@ interface PinnedGridAgent {
 type PinnedDropSide = "before" | "after";
 
 /**
- * The drag handle shown above a pinned tile on hover. It is absolutely
+ * The drag handle shown above a pinned tile while a reorder drag is in flight.
+ * Hovering a tile leaves it untouched — the handle marks the reorderable slots
+ * once dragging starts, so browsing pinned agents stays quiet. It is absolutely
  * positioned so it costs no layout: the tile keeps its size and the avatar
  * never moves.
  *
@@ -270,7 +272,7 @@ function PinnedAgentDragHandle() {
     <span
       aria-hidden="true"
       data-testid="pinned-agent-drag-handle"
-      className="pointer-events-none absolute -top-[8px] left-1/2 z-10 flex -translate-x-1/2 flex-col gap-[2px] rounded border border-border bg-popover p-[3px] opacity-0 transition-opacity group-hover:opacity-100 group-focus-visible:opacity-100"
+      className="pointer-events-none absolute -top-[8px] left-1/2 z-10 flex -translate-x-1/2 flex-col gap-[2px] rounded border border-border bg-popover p-[3px]"
     >
       {[0, 1].map((row) => {
         return (
@@ -312,6 +314,7 @@ function PinnedAgentGridCard({
   const [, moveAgent] = useLoadableSet(movePinnedAgent$);
 
   const isDragging = draggingAgentId === agent.id;
+  const isDragInFlight = draggingAgentId !== null;
   const acceptsDrop =
     isReorderable && draggingAgentId !== null && draggingAgentId !== agent.id;
 
@@ -362,7 +365,9 @@ function PinnedAgentGridCard({
           : "text-sidebar-foreground hover:bg-state-hover"
       } ${isReorderable ? "cursor-grab active:cursor-grabbing" : ""}`}
     >
-      {isReorderable && !isDragging && <PinnedAgentDragHandle />}
+      {isReorderable && isDragInFlight && !isDragging && (
+        <PinnedAgentDragHandle />
+      )}
       {dropSide && (
         <span
           aria-hidden="true"


Lunarivibe:
LGTM


PR_AUTO_REVIEW_COMMIT: 8858700b482f3874bf5b622865611c7bc94e5f8c
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

`PinnedAgentDragHandle` stops being revealed by `group-hover` / `group-focus-visible` and is now rendered only while a reorder drag is in flight (`draggingPinnedAgentId$ !== null`), on the reorderable tiles that are not the dragged one. The pinned-grid test is updated from "handle exists on reorderable tiles" to "no handles at rest, handles on the other reorderable tiles after `dragStart`". Two files, platform-only, no contract, schema, protocol, or persisted-state change.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic: no wrong conditionals or unhandled states | Pass | `zero-sidebar-pinned.tsx:368` — `isReorderable && isDragInFlight && !isDragging` is total over the three states (idle / dragged / drop-eligible); `isDragInFlight` is derived from the same `draggingPinnedAgentId$` that already drives `acceptsDrop`, so no second source of drag truth |
| `REVIEW.md` | Security: no injection, secrets, or missing auth on new surfaces | N/A | Diff renders a decorative `<span>`; no I/O, no user input, no endpoint |
| `REVIEW.md` | Type safety: no `any`, no broad casts | Pass | Only new binding is `const isDragInFlight = draggingAgentId !== null` (boolean) |
| `REVIEW.md` | Style & maintainability: no duplicated logic, comments explain WHY | Pass | The doc comment on `PinnedAgentDragHandle` was updated to state the new trigger and the reason (browsing pinned agents stays quiet); the whole-pixel-inset rationale below it is unchanged and still accurate |
| `REVIEW.md` | Feature regression guard: changed behavior needs coverage of the flow most likely to regress | Pass | The drag path itself is still covered by the untouched `reorders pinned agents by dragging` and `marks the landing slot with an insertion caret while dragging` tests in the same file; the changed test now also asserts the dragged tile shows no handle |
| `REVIEW.md` | New user-facing behavior is contained by a `FeatureSwitchKey` | Pass | The horizontal pinned grid only renders under `FeatureSwitchKey.ThreeColumnNav` (`zero-sidebar.tsx:852`, `zero-pinned-agents.ts:44`), which is still non-GA (`enabledEmailHashes` only) |
| `REVIEW.md` | DB/JSONB review gate | N/A | No `packages/db/src/schema/**`, contract, or `jsonb().$type<>()` change |
| `docs/bad-smell.md` | §1–2: no `any`, no lint/type suppressions added | Pass | No new `any`, no `oxlint-disable`/`@ts-*` added; the pre-existing file-level `oxlint-disable max-lines-per-function` at line 2 is untouched |
| `docs/bad-smell.md` | §7: fail fast, no fallback/recovery branch | Pass | The change removes a CSS reveal path and adds a render guard; no default-value or recovery branch introduced |
| `docs/fallback.md` | §1: no negative test asserting removed behavior stays removed | Pass | `toHaveLength(0)` at rest asserts the PR's own product behavior ("handles appear only during a drag"), not the absence of deleted code — it fails if the render guard regresses, so it is not a tombstone |
| `docs/fallback.md` | §2 + §7: feature behind a non-GA switch needs no compatibility branch or rollout window | Pass | `threeColumnNav` is not GA; no dual-render, version branch, or `Fallbacks` section required, and none was added |
| `docs/effect.md` | Render stays pure — no command execution, store write, DOM mutation, or subscription during render | Pass | Render only reads `useGet(draggingPinnedAgentId$)` and computes a boolean; drag transitions still run through the existing `startDrag` / `setDropTarget` / `endDrag` commands in DOM handlers |
| `docs/effect.md` | Derived value belongs in a derivation, not a synchronizing command | Pass | `isDragInFlight` is derived inline from existing state; no new state or effect added to mirror it |
| `docs/cache.md` | RC-2 / RC-3: no render-time signal allocation, no unbounded cache | Pass | No `state`/`computed`/`command` created in render, no module-level map introduced |
| `docs/cache.md` | RC-5: no duplicate mutable source of truth | Pass | Drag state stays owned by `zero-sidebar-state.ts`; the component derives, it does not copy |
| `.claude/skills/ccstate/SKILL.md` | DOM callbacks run commands through `detach(..., Reason.DomCallback)`; no manual cache duplicating `computed` memoization | Pass | Existing `detach(moveAgent(...), Reason.DomCallback)` on drop is unchanged; the diff adds no new command call or manual cache |
| `docs/testing.md` | `fix:` commit carries a regression test | Pass | `zero-sidebar.test.tsx:3990` now fails if the handle is rendered at rest |
| `docs/testing.md` | Integration/page-level test, no internal unit test; no `vi.mock` of internal paths, fetch stubbing, fake timers, or mock-call-only assertions | Pass | Existing page test extended in place; no mocks, timers, or spies added |
| `docs/testing/app-testing.md` | Platform tests are page-level under `views/`, driven through the rendered page | Pass | Test runs `setupSidebarPage` and queries within `pinned-agents-grid`; no component rendered in isolation |
| `docs/testing/testing-external-behavior.md` | Construct the case the way a user does; do not assert on store state, internal callbacks, or CSS classes | Pass | The case is created by `fireEvent.dragStart` on the real pinned-agent link (jsdom/user-event has no HTML5 drag), and the assertion is on rendered elements, not on class names or `draggingPinnedAgentId$` |
| `docs/testing/patterns.md` | Platform page-test structure: shared setup, scoped queries, `waitFor` for async | Pass | Reuses `prepareOverflowingPinnedAgents`, `pinnedAgentLink`, `createDataTransferStub` and the existing `waitFor` on card count |
| `docs/testing/anti-patterns.md` | AP-7: no testing of implementation details | Pass | The old assertion passed while the handle was invisible (`opacity: 0`); the new one keys on presence, which now equals visibility — strictly closer to user-visible behavior |
| `docs/testing/anti-patterns.md` | AP-8: no over-testing | Pass | One test covers rest / drag-source / drop-eligible / lead-agent in a single flow rather than adding parallel cases |
| `docs/deployment-compatibility.md` | Independently deployed boundaries stay compatible | N/A | Client-render-only change; no request/response, runner protocol, queue payload, migration, or persisted shape touched. Old already-loaded clients keep the old visual until refresh, which is normal client rollout, not a compatibility contract |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Non-blocking notes

- `group-focus-visible:opacity-100` is dropped along with the hover reveal, so keyboard focus no longer surfaces the grip. No accessibility loss: the handle is `aria-hidden="true"` + `pointer-events-none`, and reordering is native HTML5 drag, which has no keyboard path either way.
- With the grip gone at rest, the only remaining "these are reorderable" affordance before a drag starts is `cursor-grab` on the tile. That is the intended product change requested for this PR, not a defect — flagging it only so the trade-off is on record.

### Testing

- Coverage: Adequate. `fix:` regression test present at `turbo/apps/platform/src/views/zero-page/__tests__/zero-sidebar.test.tsx:3990`, covering rest state, drop-eligible tiles, the dragged tile, and the lead agent in one flow; the pre-existing reorder and insertion-caret tests still cover the drag/drop path they owned.
- Conventions: Compliant. Page-level test, no new mocks/timers/spies, existing helpers reused, assertions on rendered output.
- Additional verification on the deployed preview (`pr-28320-app.omby.ai`, commit `8858700`), with a real pointer-driven drag rather than synthetic events: 0 handles at rest; 0 handles while hovering a reorderable tile with the hover layer confirmed live (hover background painted `rgba(0,40,97,0.05)`); during the drag, handles on the two drop-eligible tiles, dashed ghost on the dragged tile, insertion caret on the target, none on the lead agent; after drop, the new order persists across reload and handles return to 0. The pre-fix build on a sibling PR preview shows the grip at `opacity: 1` under the same hover.
- CI at the time of review: `test-app` shards green. `cli-e2e-02-playwright (billing-transitions-2)` failed on `billing-transitions.spec.ts:397` (`expect(response.status()).toBe(200)` on a paid-invitation request) — unrelated to this diff, green on a sibling PR, and rerun.

PR #28260, opened 2026-08-20
fix(platform): fit seven models without scrolling

hulh122:
## Summary

- size the categorized model picker to its seven-row content height
- keep seven chat models free of overflow while the eighth adds one 36px row of scroll range

## Verification

- `pnpm exec prettier --check apps/platform/src/views/zero-page/components/model-provider-picker.tsx`
- `pnpm -F @okouai/app exec oxlint src/views/zero-page/components/model-provider-picker.tsx`
- `pnpm -F @okouai/app exec oxlint --type-aware -c ../../.oxlintrc.typeaware.json src/views/zero-page/components/model-provider-picker.tsx`
- `pnpm -F @okouai/app exec eslint src/views/zero-page/components/model-provider-picker.tsx --max-warnings 0`
- `pnpm -F @okouai/app check-types`
- PR Preview `https://pr-28260-app.omby.ai` at `cb423695019c454d212919761a8c9b05b0b4fdd5`
- seven-policy browser fixture: `clientHeight=292`, `scrollHeight=292`, `scrollRange=0`
- eight-policy browser fixture: `clientHeight=292`, `scrollHeight=328`, `scrollRange=36` (one 32px row plus its 4px gap)

PR #28335, opened 2026-08-20
fix(platform): keep the model picker checkmark in one column

tongx-blip:
## Problem

In the model picker the checkmark landed on a different column depending on what else the row carried:

- a plain row put it **8px** from the right edge;
- a row with the Codex fast toggle shifted its whole content by `mr-8` when selected, which pushed the check to **40px** and left it sitting *to the left of the lightning icon* instead of in the rail every other row uses;
- a row with a variant segment kept the check at 8px, but the segment's own right edge stopped 32px in, so an unselected variant row left a visibly empty gutter.

## Change

Reserve two fixed columns on the right of a fast row: `pr-16` for the content, the toggle absolutely placed at `right-8`, and the checkmark where it already was at `right-2`. Both columns are reserved whether or not the row is selected, so selecting a model no longer moves anything, and the toggle keeps its full 32×32 hit area rather than shrinking to an inline icon button.

Every row type now agrees: **checkmark right edge at 8px, trailing control right edge at 32px.**

Two related picker adjustments ride along:

- **Category switch corner.** It sat 4px below the popover's top edge but 8px in from its right edge. `py-2` on the header squares that corner up (header 36px → 44px).
- **Height animation.** Switching category swaps one list of rows for another of a different length, so the popover jumped to its new height in a single frame. It now animates between the two heights over `180ms ease` — the pair the dialog already uses in `globals.css`, so no new motion vocabulary — with the rows fading in over 120ms. A `ResizeObserver` on the list drives it, so nothing has to know which category is active; `prefers-reduced-motion` skips straight to the new height.

`select.tsx` gains a `data-slot="select-list"` handle so the popup can find the list it needs to measure. No behaviour change for other selects.

## Design reference

Interactive mockup with the alignment guides, the four corner-spacing options, and tunable motion values: https://model-picker-check.sites.vm0.io

## Testing

- Existing picker tests address rows by `aria-label`, so this change is class/ref-level for them. The animation is inert under `happy-dom`, where every `offsetHeight` is 0 and no height change is ever observed.
- Relying on the PR pipeline for lint / type-check / tests.


tongx-blip:
Changes Requested


PR_AUTO_REVIEW_COMMIT: ba40a519b53110510c0d2962c74ee93f38648201

### Summary

Gives the model picker one checkmark column: a fast-mode row now reserves two fixed 32px columns (`pr-16` on the label item, the Zap toggle absolutely placed at `right-8`) instead of shifting its content by `mr-8` when selected, so the check stays at `right-2` on every row type. Also squares the category-switch corner (`py-1` → `py-2`) and animates the popup between category heights with a new `onRef` + `ResizeObserver` command, plus a `data-slot="select-list"` handle on the shared `SelectContent` list.

One P1: the new ref command silently no-ops when it cannot find the list, which guards a state `SelectContent` always prevents.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic — no race conditions or wrong assumptions | Pass | `model-picker-panel-height.ts`: interrupted resizes cancel the in-flight animation, and `restoreOverflow` compares `heightAnimation !== animation` so a superseded animation's async `cancel` cannot clear the overflow lock its replacement set. Keyframes can exceed the popup's `max-h-[320px]`, but CSS `max-height` clamps the animated `height`, so a clamped list still lands on correct endpoints. |
| `REVIEW.md` | Security — no injection, secrets, or missing auth | N/A | Diff is presentational geometry plus a DOM animation; no network, auth, storage, or user-supplied string reaches a sink. |
| `REVIEW.md` | Type safety — no `any`, no over-broad types | Pass | No `any` or assertions added. `popup.querySelector<HTMLElement>(...)` and `row instanceof HTMLElement` narrow the DOM reads; `Animation` is the real WAAPI type. |
| `REVIEW.md` | Style & maintainability — no oversized functions, no duplicated logic, comments say WHY | Pass | New file is one ~70-line command; the added comments explain why `right-8`/`pr-16` exist, why the first measurement is a baseline, and why the overflow lock needs the identity check. |
| `REVIEW.md` | `fix:` commits must carry a regression test (missing = P0) | Pass | No page-observable behavior changes: the fix is Tailwind class positions plus a layout-driven animation. `docs/testing/testing-external-behavior.md` forbids asserting on CSS classes, and `happy-dom` has no layout engine, so `offsetHeight` is always 0 and the animation never runs there. The interactive behavior of both changed rows stays covered by `chat-composer-model-selection-and-providers.test.tsx`, which finds the fast option by `aria-label`, clicks it, and asserts the resulting `serviceTier`/`runOptions` payload. |
| `REVIEW.md` | Feature regression guard — changed existing behavior needs coverage for the affected path | Pass | The one behavioral surface touched is the fast-mode toggle's hit area, which keeps its full `inset-y-0 w-8` box and its existing `aria-label`; the existing fast-mode tests exercise that path unchanged. |
| `REVIEW.md` / `docs/testing.md` | DB / JSONB review gate | N/A | No `packages/db`, contract, or `jsonb(...)` change in the diff. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions, no dynamic `import()` | Pass | None present in the three changed files. |
| `docs/bad-smell.md` | §7 fail fast — no silent fallback/recovery paths | Finding (P1) | `model-picker-panel-height.ts:49-51`: `if (!list) { return; }` silently disables the whole effect. |
| `docs/bad-smell.md` | No hardcoded URLs or environment values | Pass | Only DOM selectors and motion constants; the durations are named constants tied to the existing `globals.css` dialog timing. |
| `docs/fallback.md` | §3 no fallback on a state the owning contract cannot produce | Finding (P1) | Same `!list` guard — `SelectContent` always renders `SelectPrimitive.List`, which is the element carrying the new `data-slot`. |
| `docs/fallback.md` | §6/§7 a rollout fallback must name surface, window, removal condition | N/A | The diff introduces no cross-version compatibility branch: no API, runner protocol, queue payload, or persisted shape is touched, so no rollout window applies. `win?.matchMedia(...)` mirrors the existing `auto-scroll.ts` pattern for `ownerDocument.defaultView` rather than tolerating an old version. |
| `docs/fallback.md` | §1 no negative tests asserting removed behavior stays removed | Pass | No test changes in the diff. |
| `docs/effect.md` | Render stays pure — no Store writes, resource allocation, or observers during render | Pass | `ModelFirstModelPickerContentLayout` only calls `useSet(...)` and passes the result as a ref; the observer is created on commit inside the command. |
| `docs/effect.md` | Use `onRef` for element-owned DOM resources, and the command must actually depend on the element | Pass | The command measures `list.offsetHeight` and animates `popup`; it matches the document's own `ResizeObserver` + abort example. |
| `docs/cache.md` | RC-7 stable callback ref, cleanup not discarded | Pass | `ref={setPanelHeightRef}` passes the `useSet` result directly with no inline arrow wrapper. |
| `docs/cache.md` | RC-8 symmetric resource lifecycle for observers and async work | Pass | The abort listener disconnects the `ResizeObserver` and cancels the in-flight height animation; per-row fade animations are 120ms and die with their detached elements. |
| `docs/cache.md` | RC-3 no unbounded lifetime caches | Pass | State is three closure locals scoped to one mounted popup; nothing is stored at module scope. |
| `.claude/skills/ccstate/SKILL.md` | DOM refs use `onRef`, never a command taking `HTMLElement \| null` | Pass | `modelPickerPanelHeightRef$ = onRef(command((_visitor, popup, signal) => ...))`. |
| `.claude/skills/ccstate/SKILL.md` | No `detach()` in the signals layer; no accessor escape | Pass | The command uses neither `get`/`set` nor `detach`. |
| `docs/testing/app-testing.md` | Platform tests are page-level via `detachedSetupPage`; no component-only unit tests | Pass | No test added, and no component-only test was added for the changed component. |
| `docs/testing/testing-external-behavior.md` | Do not assert on CSS classes or internal implementation | Pass | This is the reason the class-position fix carries no new test; asserting `pr-16`/`right-8` would freeze the implementation. |

### Findings

- `turbo/apps/platform/src/signals/zero-page/model-picker-panel-height.ts:49-51`: `if (!list) { return; }` handles a state the owning contract prevents. `SelectContent` unconditionally renders `SelectPrimitive.List`, which is the element that now carries `data-slot="select-list"`, and the ref only fires after that subtree is committed. As written, a future rename of the slot or a structural change in `select.tsx` would silently switch the animation off with no signal — exactly the failure mode `docs/bad-smell.md` §7 and `docs/fallback.md` §3 call out ("Does the PR add a fallback for a state the contract, schema, or single writer already prevents? Request removal"). Replace the silent return with a loud invariant, e.g. `throw new Error(...)` naming the missing selector, so the impossible state fails visibly instead of degrading quietly.

### Testing

- Coverage: Adequate. `fix:` with no page-observable behavior to assert — the change is Tailwind class positions plus a layout-driven animation, and `docs/testing/testing-external-behavior.md` explicitly excludes CSS-class assertions from the platform test boundary. `happy-dom` reports `offsetHeight` as 0, so the `ResizeObserver` never observes a change and the animation path stays inert under test. Existing coverage in `turbo/apps/platform/src/views/zero-page/__tests__/chat-composer-model-selection-and-providers.test.tsx` still exercises both changed rows through the page (`findComposerModel`, `findFastModeOption`, then asserting `serviceTier: "priority"` / `runOptions.codexServiceTier`), which is the regression guard for the interactive behavior around the moved elements.
- Conventions: Compliant. No test files changed, so no mock-boundary, MSW, fake-timer, or cleanup conventions are affected.

tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: 55c976f77fa665ff0e287894c1a7044489abb3f5
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Gives the model picker one checkmark column: a fast-mode row reserves two fixed 32px columns (`pr-16` on the label item, the Zap toggle at `right-8`) instead of shifting its content by `mr-8` when selected, so the check stays at `right-2` on every row type. Also squares the category-switch corner (`py-1` → `py-2`) and animates the popup between category heights through a new `onRef` + `ResizeObserver` command, with a `data-slot="select-list"` handle added to the shared `SelectContent` list.

Since the previous review, `55c976f` replaced the silent `!list` guard with a loud invariant and rewrote the Playwright helper that pinned the old layout. The P1 from that review is resolved and the new helper is the regression test for this fix.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic — no race conditions or wrong assumptions | Pass | `model-picker-panel-height.ts`: an interrupted resize cancels the in-flight animation, and `restoreOverflow` compares `heightAnimation !== animation` so a superseded animation's async `cancel` cannot clear the overflow lock its replacement just set. Keyframes can exceed `max-h-[320px]`, but CSS `max-height` clamps the animated `height`, so a clamped list still lands on correct endpoints. |
| `REVIEW.md` | Security — no injection, secrets, or missing auth | N/A | Presentational geometry plus a DOM animation; no network, auth, storage, or user-supplied string reaches a sink. |
| `REVIEW.md` | Type safety — no `any`, no over-broad types | Pass | No `any` or assertions added. `popup.querySelector<HTMLElement>(...)` and `row instanceof HTMLElement` narrow the DOM reads; `lint-types` and `lint-types-apps` passed on the parent commit of this diff. |
| `REVIEW.md` | Style & maintainability — no oversized functions, no duplicated logic, comments say WHY | Pass | The new file is one ~70-line command; comments explain why `right-8`/`pr-16` exist, why the first measurement is only a baseline, and why the overflow lock needs the identity check. |
| `REVIEW.md` | `fix:` commits must carry a regression test (missing = P0) | Pass | `e2e/playwright/tests/chat.spec.ts:889` `expectModelRowColumns` asserts the invariant this PR establishes — the checkmark is the element closest to the row's right edge inside its 8px column, the fast toggle owns the fixed 32px column beside it without covering the check, and row content stops before both. The test runs it in both selection states, so the original defect (check displaced 32px and rendered on the far side of the toggle when selected) fails it again. |
| `REVIEW.md` | Feature regression guard — replaced behavior must be intentional, scoped, and covered | Pass | The replaced assertion (`expectFastActionRightmost`) encoded the `mr-8` displacement this PR removes, so updating it is the point of the change rather than a test edited to match a regression; the replacement is stricter, pinning both columns instead of only their relative order. The fast toggle keeps its `inset-y-0 w-8` box and its `aria-label`, and `chat-composer-model-selection-and-providers.test.tsx` still drives it through the page (`findFastModeOption` → assert `serviceTier: "priority"` / `runOptions.codexServiceTier`). |
| `REVIEW.md` / `docs/testing.md` | DB / JSONB review gate | N/A | No `packages/db`, contract, or `jsonb(...)` change in the diff. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions, no dynamic `import()` | Pass | None present in the four changed files; `lint-eslint` and `lint-knip` passed on the parent commit. |
| `docs/bad-smell.md` | §7 fail fast — no silent fallback or recovery path | Pass | Fixed in `55c976f`: the missing-list branch now throws `Select popup has no [data-slot="select-list"]` instead of returning quietly, so a renamed slot or restructured popup surfaces immediately. |
| `docs/bad-smell.md` | No hardcoded URLs or environment values | Pass | Only DOM selectors and motion constants; durations are named constants tied to the existing `globals.css` dialog timing. |
| `docs/fallback.md` | §3 no fallback on a state the owning contract cannot produce | Pass | The only such branch was the `!list` guard, now an invariant throw. `win?.matchMedia(...)` follows the existing `auto-scroll.ts` treatment of `ownerDocument.defaultView`, and `heightAnimation?.cancel()` guards a local that genuinely starts `null`. |
| `docs/fallback.md` | §6/§7 a rollout fallback must name surface, window, removal condition | N/A | The diff introduces no cross-version compatibility branch: no API, runner protocol, queue payload, or persisted shape is touched, so no rollout window applies and no `Fallbacks` section is required. |
| `docs/fallback.md` | §1 no negative tests asserting removed behavior stays removed | Pass | The old assertion was replaced with the new invariant, not kept alongside it as a tombstone. |
| `docs/effect.md` | Render stays pure — no Store writes, resource allocation, or observers during render | Pass | `ModelFirstModelPickerContentLayout` only calls `useSet(...)` and passes the result as a ref; the observer is created on commit inside the command. |
| `docs/effect.md` | Use `onRef` for element-owned DOM resources, and the command must depend on the element | Pass | The command measures `list.offsetHeight` and animates `popup`; it mirrors the document's own `ResizeObserver` + abort example. |
| `docs/cache.md` | RC-7 stable callback ref, cleanup not discarded | Pass | `ref={setPanelHeightRef}` passes the `useSet` result directly, with no inline arrow wrapper. |
| `docs/cache.md` | RC-8 symmetric resource lifecycle for observers and async work | Pass | The abort listener disconnects the `ResizeObserver` and cancels the in-flight height animation; the 120ms row fades die with their detached elements. |
| `docs/cache.md` | RC-3 no unbounded lifetime caches | Pass | State is three closure locals scoped to one mounted popup; nothing is stored at module scope. |
| `.claude/skills/ccstate/SKILL.md` | DOM refs use `onRef`, never a command taking `HTMLElement \| null` | Pass | `modelPickerPanelHeightRef$ = onRef(command((_visitor, popup, signal) => ...))`. |
| `.claude/skills/ccstate/SKILL.md` | No `detach()` in the signals layer; no accessor escape | Pass | The command uses neither `get`/`set` nor `detach`. |
| `docs/testing.md` | E2E covers the happy path; error cases belong in integration tests | Pass | The Playwright test exercises the picker's two normal selection states; no error path was added to the E2E layer. |
| `docs/testing/anti-patterns.md` | AP-7 do not assert on CSS classes or internal state | Pass | The helper asserts rendered geometry from `boundingBox()`, which is the user-visible surface for a layout fix, not class names. It follows the file's existing layout assertions (`expectRightAlignedDivider`, the equal-column and Send-button tests). |
| `docs/testing/anti-patterns.md` | AP-8 do not over-test | Pass | One helper replaced one helper; no new test case, and the assertions are the three columns the fix defines. |
| `docs/testing/testing-external-behavior.md` | Construct and verify through the external interface (the page) | Pass | The helper reaches the rows through `getByRole("option")` and the page's rendered layout, and the state is set up by clicking the composer combobox rather than by touching internals. |
| `docs/testing/app-testing.md` | Platform vitest tests are page-level; no component-only unit tests | Pass | No platform vitest test was added or changed; all 8 `test-app` shards passed on the parent commit. |

### Findings

None.

### Testing

- Coverage: Adequate. The regression test lives where the behavior is observable: `e2e/playwright/tests/chat.spec.ts` runs in a real browser, so it can measure the column geometry that `happy-dom` cannot (every `offsetHeight` there is 0, which also keeps the animation path inert in vitest). `expectModelRowColumns` pins the checkmark column, the fast column, and the content boundary in both selection states.
- Conventions: Compliant. No mocks, timers, or cleanup hooks were added; the helper reuses the file's established `boundingBox()` + tolerance pattern and the existing `mockSelectedFastModel` page setup.
- Prior CI evidence on `ba40a51` (identical except for the invariant throw and this test): `lint-format`, `lint-eslint`, `lint-types`, `lint-types-apps`, `lint-knip`, and `test-app (1..8)` all passed; the only failures were this Playwright test, now updated, and one unrelated `billing-transitions-1` flake (`Execution context was destroyed` in `e2e/playwright/lib/auth.ts` Clerk token refresh), reported to the CI channel.

tongx-blip:
Changes Requested


PR_AUTO_REVIEW_COMMIT: e193c3828f3c4c3f58a1ce088cb8612e58914fbb

### Summary

Re-review after `e193c38` merged current `main` into the branch. This PR's own four files are byte-identical to the `55c976f` revision that was reviewed LGTM — `git diff 55c976f e193c38` shows zero changes in `model-picker-panel-height.ts` and `select.tsx`, and the only `chat.spec.ts` delta is `main`'s, with `expectModelRowColumns` and its constants untouched.

What changed is the ground under the PR. While it was blocked, `main` landed #28510 (one model per family, removing the media rows' variant segments), #28542 (popover narrowed back to `min-w-[260px]`) and #28260 (`max-h-[294px]`, "seven models fit without scrolling", with a new Playwright test pinning the geometry). The last one collides with this PR's header change.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic | Finding (P1) | The header grew from `py-1` (36px) to `py-2` (44px), but `main`'s height budget was sized around the old header — see the finding below. |
| `REVIEW.md` | Feature regression guard — changed behaviour needs coverage for the affected path | Finding (P1) | `main`'s new test `model picker fits seven rows and scrolls one row for eight` measures exactly what this PR shifts, and the PR does not reconcile with it. |
| `REVIEW.md` | `fix:` commits carry a regression test | Pass | `expectModelRowColumns` in `chat.spec.ts` survived the merge intact and still runs in both selection states. |
| `REVIEW.md` | Security / type safety / style | Pass | No `any`, no suppressions; unchanged from the LGTM revision. |
| `docs/bad-smell.md` | §7 fail fast, no silent fallback | Pass | The missing-list branch throws a named invariant. |
| `docs/fallback.md` | §3 / §6 no fallback on contract-prevented state, no undeclared rollout fallback | Pass | No compatibility branch; nothing to declare. |
| `docs/effect.md` | `onRef` for element-owned DOM resources | Pass | `ResizeObserver` created on commit, abort-scoped. |
| `docs/cache.md` | RC-7 stable ref / RC-8 symmetric teardown | Pass | `useSet` result passed directly; observer disconnected and animation cancelled on abort. |
| `.claude/skills/ccstate/SKILL.md` | DOM refs use `onRef`, no `detach` in signals | Pass | Unchanged. |
| `docs/testing/testing-external-behavior.md` | Assert through the page, not CSS classes | Pass | Geometry assertions via `boundingBox()`. |

### Findings

- `turbo/apps/platform/src/views/zero-page/components/model-provider-picker.tsx:1188` — the merged `max-h-[294px]` budget no longer holds. `main` added it in #28260 with the comment *"The 294px bordered cap leaves a 292px scroll viewport: enough for the **compact header** and seven model rows, while an eighth adds one 32px row plus its 4px gap."* That budget was measured against the `py-1` header this PR changes to `py-2`, which adds 8px. Seven-row content becomes 300px against a 292px viewport, so the picker starts scrolling at seven models and `main`'s new Playwright test `model picker fits seven rows and scrolls one row for eight` (`chat.spec.ts:1255`) fails on `clientHeight: 292` / `scrollHeight: 292`.

  Both invariants can hold together — raise the cap by the same 8px the header gained, `max-h-[294px]` → `max-h-[302px]` (302 − 2×0.7px border ≈ 300px viewport), and move the test's expectations to `clientHeight: 300`, `scrollHeight: 300` for seven and `scrollHeight: 336` for eight. That keeps the test's real assertion, `scrollHeight - clientHeight === rowStep`, at 336 − 300 = 36. The exact numbers should be taken from the running CI job rather than from this derivation before they are committed.

### Testing

- Coverage: Adequate for this PR's own change; `expectModelRowColumns` still pins the checkmark column, the fast column and the content boundary in both selection states.
- Gap: `main`'s seven-row geometry test is not reconciled with the new header height. That is the P1 above.
- Local checks could not run in this environment (no npm registry access, dependencies not installed); CI on `e193c38` is the source of truth for the pixel values.

tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: eb0a0c9f55c28ce29933f2ccc7ad46d69b702bce
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Re-review after `eb0a0c9`, which resolves the single P1 raised against `e193c38`. The delta is exactly that fix — two files, +10/−8 — and nothing else in the PR moved.

CI on `e193c38` reproduced the predicted collision precisely: `model picker fits seven rows and scrolls one row for eight` reported `clientHeight: 292` unchanged with `scrollHeight` moving `292 → 300`, i.e. the 8px the header gained going from `py-1` to `py-2`. `eb0a0c9` raises the cap by the same 8px, `max-h-[294px]` → `max-h-[302px]` for a 300px viewport, and follows the test to `300/300` for seven and `300/336` for eight.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic | Pass | The header and the height cap now move together, and a comment states that coupling so the next editor does not break it again. The test's real invariant is preserved: an eighth model overflows by exactly one row step, `336 − 300 = 36 = rowStep`. |
| `REVIEW.md` | Feature regression guard — replaced behaviour needs coverage | Pass | `main`'s seven-row guarantee is restored rather than weakened, and its test still asserts it, only against the new budget. The non-media picker's `max-h-[280px]` is untouched and correctly so: `ModelPickerListHeader` renders only when `mediaModelPanel` is present, so that branch never grew. |
| `REVIEW.md` | `fix:` commits carry a regression test | Pass | `expectModelRowColumns` pins the checkmark and fast columns; the geometry test pins the height budget. Both live in `chat.spec.ts` and run in the `features` shard. |
| `REVIEW.md` | Security / type safety / style | Pass | No `any`, no suppressions; the delta is one Tailwind value, one comment and four numeric literals. |
| `REVIEW.md` | DB / JSONB gate | N/A | No schema, contract or `jsonb(...)` change. |
| `docs/bad-smell.md` | No hardcoded magic values without rationale | Pass | `302` carries a comment deriving it from the 300px viewport, the header height and the 32px+4px row step. |
| `docs/bad-smell.md` | §7 fail fast, no silent fallback | Pass | The missing-list branch still throws a named invariant. |
| `docs/fallback.md` | §3 / §6 no contract-prevented fallback, no undeclared rollout fallback | Pass | Nothing added; no cross-version branch anywhere in the diff. |
| `docs/effect.md` | `onRef` for element-owned DOM resources | Pass | Unchanged from the reviewed revision. |
| `docs/cache.md` | RC-7 stable ref / RC-8 symmetric teardown | Pass | Unchanged; observer disconnected and animation cancelled on abort. |
| `.claude/skills/ccstate/SKILL.md` | DOM refs via `onRef`, no `detach` in signals | Pass | Unchanged. |
| `docs/testing/testing-external-behavior.md` | Assert through the page | Pass | Geometry read from the live scroll container, not from classes. |

### Findings

None. The previous P1 is resolved.

Two things checked and clear rather than assumed:

- No other code or test pins the old `294` / `292` / `328` values — the only repository hits are an unrelated artifact-grid constant and SVG path data.
- `main` has not touched `model-provider-picker.tsx` or `chat.spec.ts` since this head, so the merge base is still current.

### Testing

- Coverage: Adequate. The numbers came from the CI failure on `e193c38` rather than from a local derivation, so the committed constants are observed values.
- Conventions: Compliant. No mocks, timers or cleanup hooks touched.
- On `e193c38` the `billing-transitions` shards passed, so the Stripe checkout flake that blocked earlier attempts is no longer reproducing; `features` was the only failure and it is the one fixed here.
- Local checks could not run in this environment (no npm registry access, dependencies not installed); CI is the source of truth.

PR #28342, opened 2026-08-20
fix(python): bound request header name work

### LINKED ISSUE #28330: perf(python): bound request header names before repeated lowercasing
## Problem

At current `main`, a sandbox-controlled HTTP request can make the runner addon repeatedly scan and allocate an oversized, irrelevant raw header field name before routing or rejecting the request.

The request-header hook reaches these consumers in sequence:

- Codex prefetch-marker capture;
- connector-intent capture and removal;
- body-framing checks;
- browser User-Agent classification;
- Host authority normalization;
- later mitmproxy and addon request-hook lookups.

Fixed-name raw scans call `name.lower()` before establishing that a name could match. Pinned mitmproxy `12.2.3` implements case-insensitive `Headers` operations with the same full-name lowercase conversion. The Host, User-Agent, and connector-intent value limits added by #27629, #27630, and #27713 activate only after a raw name matches and do not bound work for a long nonmatching name.

Mitmproxy's HTTP/1 parser accepts any non-empty byte string before `:` as a field name and has no complete request-head size limit before addon hooks. Inbound validation itself regex-checks and lowercases a name once before the addon hook; this issue bounds repeated addon-side work and does not claim to fix dependency buffering or validation.

## Evidence

The original reproduction benchmarked the actual `_host_headers_within_budget()` function with one uppercase, non-Host raw name. Input construction was outside measurement. The four-scan column repeats the current helper four times to conservatively model the explicit addon scans and excludes additional `Headers.set_all()` and later lookups. Medians are over seven calls on CPython 3.12.3.

| Raw irrelevant name | One current scan | Four current scans | Length-first fixed-name guard | One-scan peak allocation |
| 1 MiB | 0.450 ms | 1.810 ms | 0.000580 ms | 1 MiB |
| 5 MiB | 2.293 ms | 9.169 ms | 0.000490 ms | 5 MiB |
| 16 MiB | 7.224 ms | 28.960 ms | 0.000481 ms | 16 MiB |
| 33 MiB | 16.375 ms | 65.371 ms | 0.000513 ms | 33 MiB |

Absolute timings vary by runtime, while every `bytes.lower()` still reads the full attacker-controlled name and creates an input-sized temporary on mitmproxy's shared event-loop thread. One request can repeat that work before firewall matching, credential lookup, or forwarding.

## Decision

Proceed in one PR with an inclusive 4,096-byte limit for each raw request header field name at the addon `requestheaders()` entry point.

The 4,096-byte limit matches the existing Host and User-Agent raw work budgets, keeps accepted addon work in low kilobytes, and avoids treating the unrelated 256-character capture-display threshold as a transport policy. No repository producer relies on a multi-kilobyte field name.

## Chosen direction

- Consume the metadata-only request end-stream marker, then inspect `Headers.fields` with `len(name)` before prefetch capture, connector intent, or any case-insensitive header operation.
- If every name is at most 4,096 bytes, preserve the current path unchanged.
- If any name is over the limit, remove every over-limit raw field by direct tuple assignment without decoding or normalization, preserve all in-budget fields in their original order, attach an empty HTTP 431 response, mark the header phase terminal, and return before classification, auth, streaming, logging, or upstream binding.
- Check the terminal marker at the start of the later addon `request()` hook, before private-header capture, while retaining the existing admission-release and cached-classification cleanup.
- Keep `Expect` and every other in-budget field so mitmproxy can preserve normal `100-continue` request-body behavior before sending the local response.

## Alternatives considered

- **Kill the flow:** stops continuation immediately but gives clients a reset instead of the standard 431 outcome and still requires raw removal or a separate error-hook audit.
- **Length-first guards in each consumer:** reduces several copies but mitmproxy's post-hook `Expect` lookup still lowercases the irrelevant name; future consumers can reintroduce the problem.
- **Clear every request header on rejection:** prevents later lookup but can remove `Expect: 100-continue`, leaving a client waiting for a continuation while mitmproxy waits for a body.
- **Use a 256-byte limit:** aligns with capture-display redaction but creates a stricter request compatibility policy without additional resource-bound benefit.
- **Patch mitmproxy parsing:** can address pre-hook buffering, validation, total bytes, and count, but requires broader HTTP/1, HTTP/2, and HTTP/3 transport work and is not necessary to remove the demonstrated repeated addon work.

## Risks and residual exposure

- Syntactically valid field names above 4,096 bytes that older runners processed will receive 431 from new runners. Old and new runners own separate flows, so no cross-version reader or migration is required.
- Body-bearing rejected requests receive the local response after mitmproxy consumes their request body; ordinary `Expect` behavior remains intact.
- Raw tuple mutation depends on the version-pinned `Headers.fields` representation already used throughout the addon.
- Mitmproxy still buffers, splits, regex-validates, and initially lowercases HTTP/1 names before the addon hook.
- This change does not bound total field count, aggregate request-head bytes, or unrelated values.

## Acceptance criteria

- An over-limit raw name cannot reach addon `bytes.lower()`, decoded/case-insensitive `Headers` access, private-header capture, browser classification, authority normalization, auth, request streaming, logging, or upstream binding.
- The limit is enforced before existing prefetch-marker and connector-intent scans in `requestheaders()`.
- Exactly 4,096 raw name bytes remain accepted; 4,097 bytes are rejected with an empty HTTP 431 response.
- All over-limit raw fields are removed without normalization; all in-budget fields remain byte-for-byte identical and in order.
- A later `request()` invocation observes terminal state before private-header capture and performs only existing terminal cleanup.
- Normal mixed-case Host, User-Agent, connector-intent, Codex prefetch, repeated ordinary fields, and `Expect` behavior remain unchanged.
- Existing Host, User-Agent, connector-intent value limits and exact boundaries remain unchanged.
- Structural real-flow tests use an oversized `bytes` subtype whose `lower()` raises; no wall-clock CI assertion is added.
- A local before/after benchmark records bounded addon work and allocation for 1, 5, 16, and 33 MiB names without committing a one-purpose script.
- The locked dependency check, Ruff formatting/lint, BasedPyright, focused integration coverage, and complete mitm-addon suite pass.
- No persisted format, runner protocol, public API, or upstream forwarding contract changes beyond explicit rejection of over-limit field names.

## Delivery notes

- Expected source scope is `src/mitm_addon.py` plus focused request-header integration coverage.
- No migration, feature switch, staged rollout, or cleanup issue is required.
- A separate proxy-wide request-head hardening issue would need independent evidence and cross-protocol policy; it is not hidden inside this PR.

## Distinctness

- #27629 bounds matched User-Agent values after field-name selection.
- #27630 bounds matched connector-intent values after field-name selection.
- #27713 bounds matched Host values and explicitly leaves total field-count/name work as residual exposure.

seven332:
## Summary

- reject raw request header field names above an inclusive 4,096-byte addon work budget before private-header capture or case-insensitive lookup
- remove rejected raw fields before returning to mitmproxy, preserve every in-budget field, and return an empty HTTP 431 response
- make the later request hook honor request-header terminal state before repeating private-header capture
- cover the exact boundary, guarded no-`lower()` rejection, ordinary duplicate fields, mixed-case Host/User-Agent, connector intent, and Codex prefetch behavior with real mitmproxy flows

## Explicit exclusions

- this does not bound mitmproxy's pre-hook HTTP/1 buffering, regex validation, or initial lowercase conversion
- this does not add a total request-head byte limit, field-count limit, or new value limits
- no public API, runner/backend protocol, persisted format, migration, feature switch, or staged rollout changes

## Benchmark

Inputs were constructed outside measurement under the locked CPython 3.14 environment. Timing is the median of seven calls; allocation tracing was measured separately.

| Raw irrelevant name | Existing four scans | New entry rejection | Existing peak | New peak |
| 1 MiB | 1.053 ms | 0.013417 ms | 1 MiB | 0.004934 MiB |
| 5 MiB | 5.508 ms | 0.013750 ms | 5 MiB | 0.004858 MiB |
| 16 MiB | 17.912 ms | 0.015083 ms | 16 MiB | 0.004293 MiB |
| 33 MiB | 47.228 ms | 0.014083 ms | 33 MiB | 0.004263 MiB |

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_request_header_names.py` (2 passed)
- `uv run --no-sync python -m pytest tests/` (6,393 passed)

Closes #28330

### DIFF
--- a/crates/runner/mitm-addon/src/mitm_addon.py
+++ b/crates/runner/mitm-addon/src/mitm_addon.py
@@ -109,17 +109,17 @@ from request_authority import AuthorityValidationError, TrustedAuthority, get_tr
 
 # HTTP status boundaries used in response-phase classification.
 _HTTP_STATUS_UNAUTHORIZED = 401
-_HTTP_STATUS_REQUEST_HEADER_FIELDS_TOO_LARGE = 431
 _HTTP_STATUS_BAD_GATEWAY = 502
 _HTTP_STATUS_ERROR_MIN = 400  # inclusive: start of 4xx/5xx error range
 _HTTP_OWS_CHARS = " \t"
-_MAX_REQUEST_HEADER_NAME_BYTES = 4096
 
 # Request-header phase state.
 # Creator: requestheaders() and header-phase stream/auth helpers.
 # Consumer: request() and terminal cleanup.
 # Release: auth marker is popped by terminal cleanup.
 # _REQUEST_HEADERS_TERMINATED is a flow-local sentinel for request() early exit.
+_HTTP_STATUS_REQUEST_HEADER_FIELDS_TOO_LARGE = 431
+_MAX_REQUEST_HEADER_NAME_BYTES = 4096
 _REQUEST_HEADERS_TERMINATED = "_request_headers_terminated"
 _FIREWALL_AUTH_APPLIED_IN_REQUESTHEADERS = "_firewall_auth_applied_in_requestheaders"
 _STALE_FIREWALL_AUTHORIZATION_METADATA_KEYS = (
--- a/crates/runner/mitm-addon/src/mitm_addon.py
+++ b/crates/runner/mitm-addon/src/mitm_addon.py
@@ -109,9 +109,11 @@ from request_authority import AuthorityValidationError, TrustedAuthority, get_tr
 
 # HTTP status boundaries used in response-phase classification.
 _HTTP_STATUS_UNAUTHORIZED = 401
+_HTTP_STATUS_REQUEST_HEADER_FIELDS_TOO_LARGE = 431
 _HTTP_STATUS_BAD_GATEWAY = 502
 _HTTP_STATUS_ERROR_MIN = 400  # inclusive: start of 4xx/5xx error range
 _HTTP_OWS_CHARS = " \t"
+_MAX_REQUEST_HEADER_NAME_BYTES = 4096
 
 # Request-header phase state.
 # Creator: requestheaders() and header-phase stream/auth helpers.
@@ -746,6 +748,19 @@ def client_disconnected(client: connection.Client) -> None:
 def requestheaders(flow: http.HTTPFlow) -> Awaitable[None] | None:
     """Handle request-header-only decisions before mitmproxy buffers bodies."""
     request_end_stream = mitmproxy_compat.take_request_end_stream(flow)
+    request_header_fields = flow.request.headers.fields
+    if any(len(name) > _MAX_REQUEST_HEADER_NAME_BYTES for name, _value in request_header_fields):
+        # Mitmproxy performs an Expect lookup after this hook, so rejected names
+        # must be gone before control returns while ordinary protocol fields stay.
+        flow.request.headers.fields = tuple(
+            (name, value)
+            for name, value in request_header_fields
+            if len(name) <= _MAX_REQUEST_HEADER_NAME_BYTES
+        )
+        flow.response = http.Response.make(_HTTP_STATUS_REQUEST_HEADER_FIELDS_TOO_LARGE)
+        flow.metadata[_REQUEST_HEADERS_TERMINATED] = True
+        return None
+
     codex_model_catalog_cache.capture_and_strip_prefetch_marker(flow)
     connector_intent.capture_and_strip(flow)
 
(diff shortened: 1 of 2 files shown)


seven332:
## Current-head review

**Head:** `761884847f96b9dc319ee149393402f01a4e9cf7`

**Summary:** The request-header hook now rejects raw field names above 4096 bytes before addon or mitmproxy post-hook lowercasing, preserves in-budget protocol fields and duplicates, returns an empty HTTP 431 response, and leaves the accepted boundary on the existing processing path.

**Findings:** No P0, P1, or P2 findings.

**Testing:** `uv lock --check`; locked environment sync; Ruff format and lint; BasedPyright; full mitm-addon suite (`6393 passed`). CI mitm-addon-test is also green on this head.

**Remaining scope:** Aggregate header-count and total-header-byte admission, plus parser work before the addon hook, are intentionally outside #28330.

**Verdict:** LGTM for the current head.

PR #28349, opened 2026-08-20
perf(python): resolve jsonl flush paths once per watcher

### LINKED ISSUE #28293: perf(python): resolve JSONL flush marker paths once per watcher
## Problem and evidence

The dedicated JSONL flush watcher resolves the addon's filesystem location on every idle poll.

`crates/runner/mitm-addon/src/runner_flush_lifecycle.py` invokes `_flush_jsonl_for_runner_request()` every 100 ms. `_read_jsonl_flush_request()` then rebuilds the request marker path with `Path(__file__).resolve().parent` before every marker read, including the normal steady state where no request exists. `_write_jsonl_flush_state()` repeats the same location resolution for active acknowledgements.

The watcher starts once before addon readiness and remains active for the addon lifetime, so the idle resolution is unconditional background CPU work rather than request-dependent work.

The original CPython 3.12.3 benchmark used three rounds of 20,000 missing-marker calls:

| Case | Median CPU time per call | Range |
| --- | ---: | ---: |
| Current `_read_jsonl_flush_request()` | 52.445 us | 51.995-53.635 us |
| Same reader with a pre-resolved marker path | 3.319 us | 3.222-3.357 us |

A local repetition on CPython 3.14.0/aarch64 measured medians of 25.981 us and 9.840 us respectively. The magnitude is environment-dependent, but both measurements show that repeated location resolution is a material part of the 10 Hz idle poll.

This does not reopen the marker-watcher architecture selected in #23590. The avoidable cost is locating the addon's own directory on every observation, not the bounded marker read or the polling interval.

## Decision

Resolve the addon directory once in `start_runner_jsonl_flush_worker()`, derive the concrete `jsonl-flush-request` and `jsonl-flush-state` paths, and pass those immutable paths through the worker's immediate, periodic, and final observation chain.

The worker already owns one stable extracted addon directory for its lifetime. Capturing its paths at startup removes hot-path resolution and path construction without adding global cache state, invalidation, capacity, or test-reset behavior. Every later worker start resolves its then-current module location again.

Preserve the existing test seam by patching `runner_flush_lifecycle.__file__` before worker startup. Add behavior-level integration coverage that publishes a real request after that patch has ended and observes acknowledgement in the startup-selected temporary directory.

## Alternatives considered

- **Cache by `__file__`:** avoids repeated resolution but adds process-global cache state, an eviction/capacity decision, and reset or temporary-path accumulation concerns for one worker.
- **Import-time path constants:** minimize overhead but break the existing post-import `__file__` redirection seam and tie ownership to import timing.
- **Inject from `mitm_addon.configure()`:** removes polling work but duplicates addon-location ownership in a higher module without a runtime benefit.
- **Keep the status quo:** retains measured unconditional CPU work without improving protocol correctness.

## Risks and mitigations

- **Wrong directory captured:** derive paths from the exact `__file__` expression used today and cover startup redirection with real files.
- **Stale path on restart:** keep paths as worker arguments rather than module globals; each new worker start resolves again.
- **Shutdown divergence:** pass the same paths to both normal observations and the existing final observation before join.
- **Test coupling:** assert real marker acknowledgement, not resolver call counts or mocked internal readers.

## Acceptance criteria

- No `Path.resolve()` or addon-directory derivation occurs in the 100 ms observation loop, request reader, or state writer.
- Request and state markers remain inside the active extracted addon directory chosen at worker startup.
- A later worker start resolves its current addon location independently.
- `runner_flush_request.read_runner_flush_request()` and its bounded, no-follow state-file handling remain unchanged.
- Tests that redirect the addon directory remain isolated and continue to cover missing, malformed, stale-generation, sequential, retry, timeout, readiness, and shutdown behavior.
- JSONL request/state filenames and schemas, poll interval, flush deadlines, request serialization, readiness publication, and shutdown ordering do not change.
- All required mitm-addon formatting, lint, type, and test checks pass.

## Scope and compatibility

This is one embedded Python PR. There is no Rust contract, database, public API, queue payload, billing, persisted product data, frontend, CLI, migration, feature flag, or mixed-version protocol change.

seven332:
## Summary

- resolve the JSONL request and state marker paths once when the addon watcher starts
- pass those immutable paths through immediate, periodic, and final observations instead of resolving the module location on every poll
- preserve temporary-directory isolation by proving that a worker keeps its startup directory after the test `__file__` patch ends

## Scope

- preserves marker filenames and schemas, poll and flush timeouts, parsing and validation, retry behavior, readiness publication, and shutdown ordering
- does not change Rust, public APIs, queues, billing, persisted state, migrations, frontend, or CLI behavior

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_runner_jsonl_flush.py tests/test_runner_flush_request.py tests/test_addon_configuration.py tests/test_done_hook.py` (58 passed)
- `uv run --no-sync python -m pytest tests/` (6391 passed)
- `lefthook run pre-commit`

A local CPython 3.14.0/aarch64 missing-marker benchmark improved from a 25.981 microsecond median on `main` to 12.290 microseconds on this branch; exact results are environment-dependent.

Closes #28293

### DIFF
--- a/crates/runner/mitm-addon/src/runner_flush_lifecycle.py
+++ b/crates/runner/mitm-addon/src/runner_flush_lifecycle.py
@@ -114,9 +114,13 @@ def start_runner_jsonl_flush_worker() -> None:
         if _jsonl_flush_worker is not None and _jsonl_flush_worker.is_alive():
             return
 
+        marker_directory = Path(__file__).resolve().parent
+        request_path = marker_directory / _JSONL_FLUSH_REQUEST_FILE
+        state_path = marker_directory / _JSONL_FLUSH_STATE_FILE
         _jsonl_flush_stop.clear()
         worker = threading.Thread(
             target=_run_runner_jsonl_flush_worker,
+            args=(request_path, state_path),
             name="runner-jsonl-flush",
             daemon=True,
         )
@@ -128,12 +132,12 @@ def stop_runner_jsonl_flush_worker_for_tests() -> None:
     _stop_runner_jsonl_flush_worker()
 
 
-def _run_runner_jsonl_flush_worker() -> None:
+def _run_runner_jsonl_flush_worker(request_path: Path, state_path: Path) -> None:
     """Observe current-generation JSONL markers independently of usage."""
     while True:
-        _flush_jsonl_for_runner_request()
+        _flush_jsonl_for_runner_request(request_path, state_path)
         if _jsonl_flush_stop.wait(RUNNER_JSONL_FLUSH_POLL_SECONDS):
-            _flush_jsonl_for_runner_request()
+            _flush_jsonl_for_runner_request(request_path, state_path)
             return
 
 
@@ -224,10 +228,10 @@ def _flush_delivery_work(*, trigger: _DeliveryFlushTrigger) -> None:
     codex_output_timing.retry_all_pending()
 
 
-def _flush_jsonl_for_runner_request() -> None:
+def _flush_jsonl_for_runner_request(request_path: Path, state_path: Path) -> None:
     global _last_jsonl_flush_request_id
 
-    request = _read_jsonl_flush_request()
+    request = _read_jsonl_flush_request(request_path)
     if request is None:
         return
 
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #28349

**Reviewed head:** `92a134013421e68b3fc7597e514ded61eae63833`

### Summary

The watcher now resolves its request and state marker paths once before thread startup and carries them through immediate, periodic, retry, and final shutdown observations. This matches the approved scope, removes the idle hot-path resolution without introducing cache state, and preserves the existing marker protocol and lifecycle.

### Findings

No findings.

### Testing

Ruff format/lint, BasedPyright, 58 focused integration tests, the complete 6391-test mitm-addon suite, and staged pre-commit hooks passed. Coverage uses real temporary files and publishes a request after the startup `__file__` patch ends, proving that the worker retains its selected directory.

The local microbenchmark is environment-dependent rather than a CI performance gate, and no live runner artifact smoke test was run. Rust code, filenames, schemas, and process wiring are unchanged.

### Verdict

LGTM

Issue #28278, opened 2026-08-20
refactor(api): retire the branded api namespace segment (EPIC)


lancy:
## Outcome

Remove the `/api/okou` brand namespace segment so first-party product routes live directly under `/api/`, and settle the five path collisions that removal exposes.

The segment exists only because there were two brands. `packages/api-contracts/src/contracts/api-namespaces.ts:1` names it exactly that:

```ts
export const BRANDED_API_NAMESPACE_PATHS = ["/api/okou", "/api/zero"] as const;
```

With one brand, the segment distinguishes nothing. It also repeats the hostname: `https://api.okou.ai/api/okou/agents`.

## Why this is being planned now

The immediate trigger is OAuth callback work. Three provider callback producers still emit `/api/zero/**`:

- `apps/api/src/signals/routes/slack-oauth.ts:141` → `/api/zero/slack/oauth/callback`
- `apps/api/src/signals/routes/teams-oauth.ts:167` → `/api/zero/teams/oauth/callback`
- `apps/api/src/signals/services/feishu-config.ts:74` → `/api/zero/feishu/oauth/callback`

Switching a callback producer requires adding the new URL to the provider's console allowlist — Feishu Open Platform, Slack App config, and the Microsoft identity platform app registration. Those are external, permissioned, error-prone changes. **Doing them twice, once for `/api/okou/` and again for `/api/`, is the cost this epic exists to avoid.**

The three callback paths do **not** collide with anything: `/api/slack/`, `/api/teams/` and `/api/feishu/` are entirely unused today. They can therefore land at their final shape immediately, independently of the wider migration.

## Research snapshot

Inventoried on `main` at `3143d58ad457f9ff2b7f58eb32d609a7732f3e94` (2026-08-20). **516 routes** across the API contracts package.

| Namespace | Routes | Nature |
|---|---:|---|
| `/api/okou/**` | 359 | Product API, 67 sub-segments |
| `/api/test/**` | 53 | Test-only |
| `/api/cron/**` | 25 | Scheduled jobs |
| `/api/webhooks/**` | 24 | Provider inbound |
| `/api/integrations/**` | 13 | Collides |
| `/api/runners/**` | 8 | Runner protocol |
| `/api/cli/**` | 7 | CLI |
| `/api/desktop/**`, `/api/desktop-auth/**` | 9 | Collides |
[...]

PR #28353, opened 2026-08-20
fix(api-contracts): declare billing checkout preview in the okou namespace

### LINKED ISSUE #28350: fix(api-contracts): declare the billing checkout preview path in the canonical namespace
`packages/api-contracts/src/contracts/billing.ts:968` declares its path in the legacy brand namespace:

```ts
path: "/api/zero/billing/concurrency-checkout/preview",
```

It is the only one of the package's 309 branded contract paths that does. The other six routes in the same contract — including `concurrency-checkout` immediately below it at `:984` — all declare `/api/okou/...`.

## This is not cosmetic

`apiNamespaceAliasPaths()` registers both namespaces regardless of which one the contract declares, so the **acceptor** side is unaffected. But the declared path is what a ts-rest client uses to build its request URL, and `apps/platform/src/signals/zero-page/billing.ts:10` imports this contract.

Production request logs show the split inside a single checkout flow:

| Path | Declared as | Requests |
|---|---|---|
| `/api/zero/billing/concurrency-checkout/preview` | `/api/zero/...` | 2 |
| `/api/okou/billing/concurrency-checkout` | `/api/okou/...` | 2 |

One user action, two adjacent calls, two different namespaces — decided purely by the declaration.

## Why it matters

The `/api/zero/**` alias exists to serve already-released clients. This line makes the **current** Platform build an active producer of legacy-namespace traffic, which blocks the #26701 removal gate that requires Zero namespace use to reach zero for an observation window. That gate cannot pass while the source of the traffic is current code rather than historical clients.

There is also a latent hazard in the eventual alias removal. Every other branded contract collapses from two registrations to its declared `/api/okou/...` path. This one's declared form is the namespace being removed, so a removal keyed off the declared path deletes the route instead of de-aliasing it.

## Change

Declare the canonical branded path:

```ts
path: "/api/okou/billing/concurrency-checkout/preview",
```

This is safe to ship on its own. Both aliases stay registered, so a browser still running the previously served Platform bundle keeps reaching `/api/zero/...` successfully; newly served bundles use `/api/okou/...`. The Rust and Python binding generators do not include this path.

## Add a guard

`packages/eslint-rules/src/ccstate/rules/no-non-zero-api.ts` already exists and its own doc comment lists `"/api/zero/billing/status"` as a violation, but it is registered only in `apps/platform/eslint.config.js:40`. It never sees `packages/api-contracts`, which is how this line survived.

Do **not** simply extend that rule to the contracts package. It flags every `/api/` path that is not `/api/okou/`, which would reject the neutral namespaces that #28278 is deliberately introducing — `/api/webhooks/**`, `/api/integrations/**`, `/api/connectors/**`.

Add a narrower assertion instead, as a test in `packages/api-contracts`: no contract in the package declares a path beginning with `/api/zero/`. Enumerate the contracts through the existing barrel rather than by scanning files, so a new contract cannot bypass it. Keep the message specific enough that someone hitting it understands the declaration is what clients send.

## Do not

- Do not change any other contract path.
- Do not modify `apiNamespaceAliasPaths()` or `BRANDED_API_NAMESPACE_PATHS`.
- Do not remove or narrow the `/api/zero/**` aliases; that is #26701.
- Do not rename or repurpose the existing `no-non-zero-api` rule.

## Tests

- The contract declares `/api/okou/billing/concurrency-checkout/preview`.
- Requests to both `/api/okou/...` and `/api/zero/...` forms still reach the handler.
- The new package-level assertion fails if any contract declares an `/api/zero/` path.

lancy:
Closes #28350

## What changed

`billingConcurrencyCheckoutContract.preview` was the only branded contract path in `packages/api-contracts` declared under `/api/zero/`. It now declares the canonical branded path:

```ts
path: "/api/okou/billing/concurrency-checkout/preview",
```

The declared path is what a ts-rest client uses to build its request URL, and `apps/platform/src/signals/zero-page/billing.ts` imports this contract — so the current Platform build was actively producing legacy-namespace traffic inside a single checkout flow (`preview` on `/api/zero/...`, the adjacent `create` on `/api/okou/...`).

Safe to ship alone: `apiNamespaceAliasPaths()` still registers both namespaces, so already-loaded bundles keep reaching `/api/zero/...`. The Rust and Python binding generators do not include this path.

## Guard

New assertion test at `packages/api-contracts/src/contracts/__tests__/api-namespace-declarations.test.ts`:

- Enumerates route declarations through the existing barrel (`src/contracts/index.ts`), not by scanning files.
- Fails if any enumerated contract declares a path beginning with `/api/zero/`, with a message explaining that the declaration is what clients send.
- A non-vacuity check asserts the walk actually yields branded routes and includes the corrected preview route, so a broken walk cannot make the assertion pass silently.

Per the issue, the existing `no-non-zero-api` ESLint rule was left untouched — extending it to the contracts package would reject the neutral `/api/webhooks/**`, `/api/integrations/**`, and `/api/connectors/**` namespaces that #28278 is introducing.

`apiNamespaceAliasPaths()`, `BRANDED_API_NAMESPACE_PATHS`, the `/api/zero/**` aliases, and every other contract path are unchanged.

## Coverage note for the reviewer

`src/contracts/index.ts` re-exports from 125 of the package's 170 contract modules; 66 of the 291 contract routers (for example everything in `./browser`) are consumed only through the `./contracts/*` subpath export and never appear in the barrel. The new assertion therefore guards the barrel-reachable contracts, which includes the route fixed here. Closing that remaining gap would mean either broadening the barrel or enumerating modules some other way — out of scope here, and worth a follow-up if you want the guard to be exhaustive.

## Verification

Local dependency installation is blocked in this environment (the npm registry is not reachable from the run), so lint, type-check, and vitest were not run locally; CI covers them. Existing coverage that exercises this change:

- `apps/api/src/__tests__/api-namespace-compatibility.test.ts` asserts every route in `ROUTES` registers symmetric `/api/okou/...` and `/api/zero/...` paths against the same handler and schema, independent of which namespace the contract declares.
- `apps/api/src/signals/routes/__tests__/billing-checkout.test.ts` and the Platform billing/queue tests drive the route through the contract object, so they follow the declared path.

### DIFF
--- a/turbo/packages/api-contracts/src/contracts/__tests__/api-namespace-declarations.test.ts
+++ b/turbo/packages/api-contracts/src/contracts/__tests__/api-namespace-declarations.test.ts
@@ -1,0 +1,84 @@
+import { describe, expect, it } from "vitest";
+
+import { brandedApiNamespace } from "../api-namespaces";
+import * as apiContracts from "../index";
+
+interface DeclaredRoute {
+  readonly declaration: string;
+  readonly method: string;
+  readonly path: string;
+}
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return typeof value === "object" && value !== null;
+}
+
+function toDeclaredRoute(
+  declaration: string,
+  value: unknown,
+): DeclaredRoute | undefined {
+  if (!isRecord(value)) {
+    return undefined;
+  }
+  const { method, path } = value;
+  if (typeof method !== "string" || typeof path !== "string") {
+    return undefined;
+  }
+  return { declaration, method, path };
+}
+
+/**
+ * Enumerates every route declared by a contract exported from the package
+ * barrel. Walking the barrel rather than the source files means a route added
+ * to an already-exported contract is covered without touching this test.
+ */
+function declaredRoutes(): readonly DeclaredRoute[] {
+  const routes: DeclaredRoute[] = [];
+  for (const [exportName, exported] of Object.entries(apiContracts)) {
+    if (!isRecord(exported)) {
+      continue;
+    }
+    for (const [routeName, route] of Object.entries(exported)) {
+      const declared = toDeclaredRoute(`${exportName}.${routeName}`, route);
+      if (declared) {
+        routes.push(declared);
+      }
+    }
+  }
+  return routes;
+}
+
+describe("branded API namespace declarations", () => {
+  const routes = declaredRoutes();
+
+  it("enumerates branded contract routes through the package barrel", () => {
+    const brandedRoutes = routes.filter(({ path }) => {
+      return brandedApiNamespace(path) !== undefined;
+    });
+
+    expect(brandedRoutes.length).toBeGreaterThan(100);
+    expect(
+      brandedRoutes.some(({ method, path }) => {
+        return (
+          method === "POST" &&
+          path === "/api/okou/billing/concurrency-checkout/preview"
+        );
+      }),
+    ).toBe(true);
+  });
+
+  it("declares every branded contract path in the Okou namespace", () => {
+    const legacyDeclarations = routes
+      .filter(({ path }) => {
+        return brandedApiNamespace(path) === "zero";
+      })
+      .map(({ declaration, method, path }) => {
+        return `${declaration} declares ${method} ${path}`;
+      });
+
+    expect(
+      legacyDeclarations,
+      "A declared contract path is the URL its ts-rest clients request, so a /api/zero/ declaration makes current clients produce legacy-namespace traffic. Declare branded paths as /api/okou/...; the /api/zero/ alias exists only for already-released clients.",
+    ).toStrictEqual([]);
+  });
+});
--- a/turbo/packages/api-contracts/src/contracts/billing.ts
+++ b/turbo/packages/api-contracts/src/contracts/billing.ts
@@ -965,7 +965,7 @@ export type BillingUsagePackMigrationContract =
 export const billingConcurrencyCheckoutContract = c.router({
   preview: {
     method: "POST",
-    path: "/api/zero/billing/concurrency-checkout/preview",
+    path: "/api/okou/billing/concurrency-checkout/preview",
     headers: authHeadersSchema,
     body: concurrencyCheckoutPreviewRequestSchema,
     responses: {


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: 9fdb50eee8a542574f98cacabba0547fa55f742e
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Flips `billingConcurrencyCheckoutContract.preview` from `/api/zero/billing/concurrency-checkout/preview` to the canonical `/api/okou/...` form, and adds a package-level assertion test that walks contracts through the `src/contracts/index.ts` barrel and fails on any `/api/zero/` declaration. Behaviour is safe on both sides of the rollout: `apiNamespaceAliasPaths()` derives both namespace paths from the suffix regardless of which one the contract declares, so old browser bundles still reach `/api/zero/...` while newly served bundles produce `/api/okou/...`.

One non-blocking follow-up worth recording: the barrel does not reach every contract in the package, so the new guard is not exhaustive (quantified in the criteria table below). The PR summary discloses this, and the issue explicitly prescribed the barrel, so it is a scope decision rather than a defect in this PR.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic — no wrong assumptions or unhandled paths | Pass | `billing.ts:968` only changes the declared path; `contracts/api-namespaces.ts:38` derives both branded paths from the suffix, and `apps/api/src/signals/route-entry.ts` applies that to every entry in `ROUTES`, so neither namespace loses its registration. |
| `REVIEW.md` | Security — no auth/secret/injection exposure | Pass | No auth logic touched; `authHeadersSchema` on the route is unchanged and the handler in `apps/api/src/signals/routes/billing-concurrency-checkout.ts` still gates on org admin. |
| `REVIEW.md` | Type safety — no `any`, no broad types | Pass | New test narrows through `unknown` via `isRecord`/`toDeclaredRoute`; no `any`, no `as` casts. |
| `REVIEW.md` | Interface changes — breaking changes identified | Pass | The declared path is a client-facing interface; the break is neutralised by the retained alias and is described in the PR summary. |
| `REVIEW.md` | `fix:` commit must carry a regression test | Pass | `api-namespace-declarations.test.ts` fails against the pre-fix declaration — that is exactly the regression this PR fixes. |
| `REVIEW.md` | Feature regression guard — changed behaviour has coverage | Pass | Existing `apps/api/src/__tests__/api-namespace-compatibility.test.ts` asserts every route registers symmetric Okou/Zero paths against the same handler and schema, including a case where the source contract declares the legacy path. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No `turbo/packages/db/src/schema/**` change, no Zod response-schema change, no `jsonb().$type<>()` change; only a route path string moved. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions, no dynamic `import()` | Pass | Neither file contains a suppression comment or dynamic import. Nit, non-blocking: `isRecord` also returns `true` for arrays, so an exported array would be walked by index key — no contract export is an array, so it is inert. |
| `docs/bad-smell.md` | No hardcoded URLs / environment-specific values | Pass | The literal is a contract route path, which is the declaration itself, not an environment-dependent base URL. |
| `docs/fallback.md` | No negative test asserting removed behaviour stays removed | Pass | `/api/zero/**` is still live and still served; the new test asserts a forward-looking declaration invariant on exported contracts, not that a retired route stays 404. It is a static invariant guard standing in for the `no-non-zero-api` rule that never reached this package, not a tombstone. |
| `docs/fallback.md` | Every fallback introduced by the diff is declared in the PR summary | N/A | The diff introduces no fallback. The `/api/zero/**` alias in `apiNamespaceAliasPaths()` predates this PR and is untouched; per §9, a PR that only keeps an existing fallback needs no `Fallbacks` section. |
| `docs/fallback.md` | Rollout fallback names surface, window, removal condition | N/A | No new rollout fallback. Removal of the existing aliases is explicitly deferred to #26701. |
| `docs/deployment-compatibility.md` | Frontend → backend request changes stay compatible during rollout | Pass | Old web clients (~2 days exposure) keep sending `/api/zero/...`, which the API still registers; new clients send `/api/okou/...`, which has always been registered. This is the "keep old endpoint paths during migration" pattern, not a one-shot protocol flip. |
| `docs/deployment-compatibility.md` | Previous frontend request shape is tested while it can still reach the API | Pass | Covered generically for all routes by `api-namespace-compatibility.test.ts`; a route-specific duplicate would add no coverage over the alias mechanism it already pins. |
| `docs/testing.md` | Integration-focused; no unit tests for internal functions | Pass | The new test asserts a package-wide invariant over the exported contract surface rather than an internal helper, and duplicating it as an API route test would not reach undeclared contracts. |
| `docs/testing.md` | Real infrastructure; mock only at the system boundary | Pass | No mocks at all — the test imports the real barrel. |
| `docs/testing.md` | Avoid over-testing | Pass | Two assertions only: the invariant, plus a non-vacuity check so a broken walk cannot pass silently. |
| `docs/testing/anti-patterns.md` | AP-1/AP-2/AP-4/AP-5 — no `toHaveBeenCalled`-only assertions, no fetch stubbing, no relative-path `vi.mock`, no fake timers | Pass | None of `vi.mock`, `vi.stubGlobal`, `vi.useFakeTimers`, or `toHaveBeenCalled` appear in the new file. |
| `docs/testing/patterns.md` | Pure/invariant tests stay simple, with no needless lifecycle hooks | Pass | No `beforeEach`/`afterEach`; the route list is computed once and asserted directly. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

### Testing

- Coverage: Adequate. The new assertion fails on the exact pre-fix declaration, satisfying the regression-test requirement for a `fix:` commit. Cross-version coverage for the retained `/api/zero/...` form already exists in `apps/api/src/__tests__/api-namespace-compatibility.test.ts`, which asserts symmetric registration for every route and explicitly covers a contract that declares the legacy path.
- Conventions: Compliant. No mocking, no fake timers, no call-count assertions, no lifecycle hooks. The non-vacuity assertion (`brandedRoutes.length > 100` plus a check that the corrected preview route is present) prevents the invariant from passing on an empty enumeration.
- Guard scope, verified statically against this branch: the barrel re-exports contract values from 125 of 170 modules in `src/contracts`, giving the new test **282 branded routes**. **74 branded routes across 16 modules are not reachable through the barrel** — including `browser.ts`, `chat-threads.ts`, `integrations.ts`, `run-routes.ts`, and some contracts in `billing.ts` itself, which are consumed only via the `./contracts/*` subpath export. No contract in the package declares `/api/zero/` after this change, so nothing is unprotected today, but a future contract added to one of those 16 modules could still declare a legacy path without tripping the guard. Non-blocking here — the barrel mechanism was prescribed by #28350 and the gap is disclosed in the PR summary — but worth a follow-up issue so the guard becomes exhaustive before #26701 relies on it.

PR #28344, opened 2026-08-20
feat(platform): import a deck as a presentation template from the picker

bingjiez666:
Part of #28252. The last of the three; #28280 deleted the old renderer, #28298 added `publish`.

## The entry point that was missing

The platform has never called the presentation-template API at all — the picker's slides tab only ever listed the built-in catalog from `@okouai/core`. There was no way to hand over a deck.

Now the slides tab starts with an import card. Choosing a `.pptx` or `.pdf`:

```
attach it to the composer  →  write the message  →  send
```

That is the whole feature.

## Why there is no upload code here

Because there does not need to be any. Importing a template is a chat message with a file attached, so it reuses what the composer already has:

- `draft.uploadAttachment$` — the same path a dragged file takes, with the same chip and the same progress
- `submission.submitCurrentInput$` → the existing new-thread send, which creates the thread, resolves the attachments, and **navigates into it**

The user lands in a thread they can watch, interrupt, and follow up in. A background job could not offer that, and it is why the flow was reshaped in #28252.

The message is a plain sentence — "Analyse this deck and save its visual language as a reusable presentation template." — not a specification, so the owner can read what was said on their behalf.

## Placement

`TemplatePickerDialog` is already at its complexity budget and carries a comment saying so, so the card is a small self-contained component appended to `PptTemplateGrid` instead. `onImportedPresentation` is threaded through to close the picker.

Behind `PresentationTemplates`, which is off, so nothing renders today.

## Tests

Two cases in the existing picker suite, driving the real composer:

- choosing a deck closes the picker and sends a message whose `userMessage` carries the uploaded file part, with the import prompt
- with the switch off, the card is not in the dialog

The first assertion initially looked for `attachFiles` on the run body and found nothing — attachments travel inside `userMessage`, which is what it checks now.

## Checks

`pnpm format`, `pnpm check-types` (14/14), `pnpm turbo run lint` (13/13), `pnpm knip` — all clean. The whole `chat-composer-template-picker` suite passes (34 tests).

Lint caught a floating promise in the change handler; it uses `detach(..., Reason.DomCallback)` like the rest of the file. Knip caught the prompt constant being exported without a second consumer.

Strings are added in all ten locales.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #28362, opened 2026-08-20
feat: point runs at the deck reverse-engineering guide

bingjiez666:
Part of #28252. Stacked on #28344, which introduces the import message this extends.

> [!WARNING]
> **Local validation could not run.** The sandbox lost `node_modules` mid-session and `registry.npmjs.org` has been returning `424` since (retried nine times over ~15 minutes), so `pnpm install` fails and format, lint, check-types, knip, and tests are all unavailable. CI is the first validation for this PR. Details and what was verified by hand are at the bottom.

## The gap

The import flow had no instructions attached to it. The message #28344 sends is a plain sentence, and the guide that knows how to render pages, extract a visual language, and publish the result is not mounted as a skill — so nothing told a run how to do the work, and nothing ever called `publish`. An extraction could finish and still leave the user with nothing.

## Where the guide lives, and why it stays there

`vm0-ai/Template-artifact`, branch `feat/reverse-template-skill`, path `reverse-template/`. It is still being revised against real decks and its scripts change with each run; syncing it into `vm0-skills` would put that iteration into the every-minute cron and into every unrelated run's skill mounts.

So its location is one hardcoded constant in `@okouai/core/presentation-template-skill`.

**That repository is private.** Neither the skills sync (anonymous `codeload` fetch) nor an ordinary user's run can reach it — it resolves only for a run whose GitHub access covers a vm0 private repository. That is precisely why every use of it sits behind the `PresentationTemplates` switch, and it is the reason to move the guide into `vm0-skills` once it settles. At that point the constant becomes a plain skill name and the clone step disappears.

## Two ways in, one instruction

- **Template picker** — choosing a deck appends the pointer to the message it sends, so the thread the user lands in already knows what to follow.
- **Chat box** — the standing agent-tools prompt carries the same line when the switch is on, next to the `workflow-setup` line, so a deck dropped straight into the composer reaches the same guide.

Both call `presentationTemplateSkillInstruction()`, so the two entry points cannot drift.

The switch is off, so the prompt is byte-identical today.

## Also, in `Template-artifact#18`

The guide itself was missing both ends of the flow, fixed there rather than here:

- **Input** — the deck is an attachment on the message that opened the thread, already on local disk; the ingest step has nothing to download.
- **Delivery** — a new step 4 calls `okou presentation-template publish` with the deck, the ordered pages, and the package. Previously the guide stopped at a package directory, which is not a template.

## What I verified without a toolchain

- `ZeroRunAfterPreCreate` declares `featureSwitchContext` (`zero-runs-create.service.ts:934`), and `attemptInput` spreads `input`, so `buildZeroCreateAgentRunArgs` already receives it — the new field is a declaration, not new plumbing.
- `FeatureSwitchKey` was not previously imported in that file, so the added import does not collide.
- The three location constants are module-private, since only the instruction builder reads them, so knip should not see unused exports.

None of that substitutes for running the checks. If CI finds formatting or type issues, they are mine.

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
@@ -1581,6 +1581,42 @@ describe("CHAIN-RUN: entitled run lifecycle through runner and sandbox webhooks"
     );
   });
 
+  it("names the deck guide in the agent tools prompt only once presentation templates are on", async () => {
+    const api = createRunsApi(context);
+    const connectors = createConnectorBddApi(context);
+    const { actor, agentId, runnerGroup } = await entitledRunActor();
+
+    // The guide is not a mounted skill, so the prompt is the only thing that
+    // tells a run where to find it. Off, it must stay out of every run.
+    const gatedOff = await api.createRun(actor, {
+      agentId,
+      prompt: "turn this deck into a template",
+      modelProvider: "anthropic-api-key",
+    });
+    await api.heartbeatRunner(runnerGroup);
+    const gatedOffClaim = await api.claimRunnerJob(gatedOff.runId);
+    expect(gatedOffClaim.appendSystemPrompt ?? "").toContain("# Agent Tools");
+    expect(gatedOffClaim.appendSystemPrompt ?? "").not.toContain(
+      "vm0-ai/Template-artifact",
+    );
+
+    await connectors.updateFeatureSwitches(actor, {
+      [FeatureSwitchKey.PresentationTemplates]: true,
+    });
+
+    const gatedOn = await api.createRun(actor, {
+      agentId,
+      prompt: "turn this deck into a template",
+      modelProvider: "anthropic-api-key",
+    });
+    await api.heartbeatRunner(runnerGroup);
+    const gatedOnClaim = await api.claimRunnerJob(gatedOn.runId);
+    const appendSystemPrompt = gatedOnClaim.appendSystemPrompt ?? "";
+    expect(appendSystemPrompt).toContain("vm0-ai/Template-artifact");
+    expect(appendSystemPrompt).toContain("reverse-template");
+    expect(appendSystemPrompt).toContain("feat/reverse-template-skill");
+  });
+
   it("emits api dispatch timing for exact-empty direct dispatch runs", async () => {
     const api = createRunsApi(context);
     const { actor, agentId, runnerGroup } = await entitledRunActor();
--- a/turbo/apps/api/src/signals/services/zero-runs-create.service.ts
+++ b/turbo/apps/api/src/signals/services/zero-runs-create.service.ts
@@ -12,7 +12,12 @@ import type { ModelProviderCredentialScope } from "@okouai/api-contracts/contrac
 import type { PublicBrand } from "@okouai/api-contracts/contracts/public-brand";
 import { permissionGrantsToFirewallPolicies } from "@okouai/connectors/firewall-metadata/policy";
 import type { FirewallPolicies } from "@okouai/connectors/firewall-types";
-import type { FeatureSwitchContext } from "@okouai/core/feature-switch";
+import {
+  isFeatureEnabled,
+  type FeatureSwitchContext,
+} from "@okouai/core/feature-switch";
+import { FeatureSwitchKey } from "@okouai/core/feature-switch-key";
+import { presentationTemplateSkillInstruction } from "@okouai/core/presentation-template-skill";
 import {
   agentDisplayNameForPublicBrand,
   appUrlForPublicBrand,
@@ -349,6 +354,7 @@ function buildIntegrationToolsPrompt(
 function buildAgentToolsPrompt(args: {
   readonly triggerSource: TriggerSource;
   readonly cloudBrowserEnabled: boolean | undefined;
+  readonly presentationTemplatesEnabled: boolean;
 }): string {
   const okouCliCommand = `npx --yes --package="\${CLI_PKG_URL}" okou`;
   return [
@@ -359,6 +365,9 @@ function buildAgentToolsPrompt(args: {
     "- Locate local agent-session files, search web chat messages, or inspect external services via connectors: `okou search --help`.",
     '- Workflow and automation requests use the `workflow-setup` skill first, then follow its guidance. This covers creating, editing, inspecting, running, scheduling, enabling, disabling, copying, or deleting a workflow or automation, and any recurring or event-driven request (for example "every morning", "when a new email arrives", "whenever X happens", "monitor", "remind me", "keep this in sync") even when the user does not say the word "workflow".',
     "- Manage recurring workflow automations: `okou workflow automation --help`. Do NOT use /loop, cron tools (CronCreate, CronList, CronDelete), or ScheduleWakeup — they are not available.",
+    ...(args.presentationTemplatesEnabled
+      ? [`- ${presentationTemplateSkillInstruction()}`]
+      : []),
     "- Browser access: `agent-browser` provides rendered-page inspection and interaction. For one known public URL when you only need page content, prefer `okou scrape <url> --format markdown`; use `agent-browser` when you need browser state, authentication, JavaScript, screenshots, or interaction.",
     ...(args.cloudBrowserEnabled === true
       ? [
@@ -463,6 +472,7 @@ function buildAppendSystemPrompt(args: {
   readonly userInfo: UserInfo;
   readonly triggerSource: TriggerSource;
   readonly cloudBrowserEnabled: boolean | undefined;
+  readonly presentationTemplatesEnabled: boolean;
 }): string {
   const identity = buildAgentIdentityPrompt(args.agent, args.publicBrand);
   return [
@@ -470,6 +480,7 @@ function buildAppendSystemPrompt(args: {
     buildAgentToolsPrompt({
       triggerSource: args.triggerSource,
       cloudBrowserEnabled: args.cloudBrowserEnabled,
+      presentationTemplatesEnabled: args.presentationTemplatesEnabled,
     }),
     buildCurrentUserPrompt(args.userInfo),
   ]
@@ -657,6 +668,7 @@ function createRunBody(args: {
   readonly publicBrand: PublicBrand | undefined;
   readonly appendSystemPrompt: string | undefined;
   readonly cloudBrowserEnabled: boolean | undefined;
+  readonly presentationTemplatesEnabled: boolean;
 }) {
   const triggerSource = args.triggerSource ?? "web";
   const baseAppendSystemPrompt = buildAppendSystemPrompt({
@@ -665,6 +677,7 @@ function createRunBody(args: {
     userInfo: args.userInfo,
     triggerSource,
     cloudBrowserEnabled: args.cloudBrowserEnabled,
+    presentationTemplatesEnabled: args.presentationTemplatesEnabled,
   });
   return {
     prompt: args.body.prompt,
@@ -837,6 +850,7 @@ function buildZeroCreateAgentRunArgs(args: {
   readonly timing: ApiDispatchTimingCollector;
   readonly threadSessionResolution?: ChatThreadSessionResolution;
   readonly cloudBrowserEnabled: boolean | undefined;
+  readonly featureSwitchContext: FeatureSwitchContext;
 }): CreateAgentRunArgs {
   const command = args.command;
   const agentModelProviderId = optionalAgentSetting(args.agent.modelProviderId);
@@ -853,6 +867,10 @@ function buildZeroCreateAgentRunArgs(args: {
       publicBrand: command.publicBrand,
       appendSystemPrompt: command.appendSystemPrompt,
       cloudBrowserEnabled: args.cloudBrowserEnabled,
+      presentationTemplatesEnabled: isFeatureEnabled(
+        FeatureSwitchKey.PresentationTemplates,
+        args.featureSwitchContext,
+      ),
     }),
     apiStartTime: command.apiStartTime,
     modelProviderId: command.modelProviderId ?? agentModelProviderId,
--- a/turbo/apps/platform/src/signals/zero-page/presentation-template-import.ts
+++ b/turbo/apps/platform/src/signals/zero-page/presentation-template-import.ts
@@ -1,5 +1,6 @@
 import { command, computed } from "ccstate";
 import { FeatureSwitchKey } from "@okouai/core/feature-switch-key";
+import { presentationTemplateSkillInstruction } from "@okouai/core/presentation-template-skill";
 
 import { featureSwitch$ } from "../external/feature-switch.ts";
 import type { ComposerSignals } from "./composer-signals.ts";
@@ -13,12 +14,18 @@ export const PRESENTATION_TEMPLATE_IMPORT_ACCEPT = ".pptx,.pdf";
 /**
  * The message the deck is sent with.
  *
- * Kept as a plain sentence on purpose: importing a template is not a special
+ * Opens as a plain sentence on purpose: importing a template is not a special
  * protocol, it is a chat message with a file attached, and the user should be
- * able to read what was said on their behalf in the thread they land in.
+ * able to read what was asked on their behalf in the thread they land in. The
+ * pointer to the guide follows, because the guide is not a mounted skill the
+ * run would otherwise find.
  */
-const PRESENTATION_TEMPLATE_IMPORT_PROMPT =
-  "Analyse this deck and save its visual language as a reusable presentation template.";
+function presentationTemplateImportPrompt(): string {
+  return [
+    "Analyse this deck and save its visual language as a reusable presentation template.",
+    presentationTemplateSkillInstruction(),
+  ].join("\n\n");
+}
 
 export const presentationTemplateImportEnabled$ = computed((get) => {
   return get(featureSwitch$)[FeatureSwitchKey.PresentationTemplates] ?? false;
@@ -52,7 +59,7 @@ export const importPresentationTemplateDeck$ = command(
     if (!attached) {
       return false;
     }
-    set(signals.draft.setDraftInput$, PRESENTATION_TEMPLATE_IMPORT_PROMPT);
+    set(signals.draft.setDraftInput$, presentationTemplateImportPrompt());
 
     const action = await get(signals.submission.primaryAction$);
     signal.throwIfAborted();
--- a/turbo/apps/platform/src/views/zero-page/__tests__/chat-composer-template-picker.test.tsx
+++ b/turbo/apps/platform/src/views/zero-page/__tests__/chat-composer-template-picker.test.tsx
@@ -3371,6 +3371,10 @@ describe("chat composer templates", () => {
     });
     await waitFor(() => {
       expect(sentPrompt).toContain("reusable presentation template");
+      // The guide is not a mounted skill, so the message has to carry the
+      // pointer; without it the thread opens with nothing to follow.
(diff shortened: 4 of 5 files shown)


bingjiez666:
Changes Requested


PR_AUTO_REVIEW_COMMIT: f18cb858c1826a02b73d96256d06fc5400f87c2d

### Summary

Adds one shared instruction that tells a run where the deck reverse-engineering guide lives, and emits it from two entry points: the template picker's import message and the standing agent-tools prompt. Both emissions are gated on `PresentationTemplates`, which is off, so no run sees the line today.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | `feat:` commits must have integration tests | Finding (P0) | Three source files changed, no test file in the diff; `git grep` for `presentationTemplateSkillInstruction`, `reverse-template`, `Template-artifact` across all test dirs returns nothing. |
| `REVIEW.md` | Feature regression guard — changed existing behavior needs regression coverage for the affected path | Finding (P0) | `presentation-template-import.ts` changes the import message the picker sends; the existing picker test asserts only `toContain("reusable presentation template")`, which passes identically before and after this diff. |
| `REVIEW.md` | New user-facing behavior must be gated behind a `FeatureSwitchKey` | Pass | Both emissions gated on `FeatureSwitchKey.PresentationTemplates` — API at `zero-runs-create.service.ts` `buildZeroCreateAgentRunArgs`, platform via `presentationTemplateImportEnabled$`. |
| `REVIEW.md` | Correctness — no wrong assumptions in threaded arguments | Pass | `featureSwitchContext` is already a field of `ZeroRunAfterPreCreate`, and `attemptInput` spreads `input`, so the new parameter is a declaration rather than new plumbing. |
| `REVIEW.md` | Type safety — no `any`, no suppressions | Pass | No `any`, no `@ts-ignore`, no lint-disable in the diff; the new parameter is `boolean` and the new module returns `string`. |
| `docs/bad-smell.md` | §6 Never hardcode environment-specific values; use centralized configuration | Finding (P1) | `presentation-template-skill.ts` pins repo, **working branch**, and path as literals. Hardcoding a skill source has precedent (`github-url.ts` `DEFAULT_SKILLS_OWNER/REPO/BRANCH`), but those name a stable `main`; this names a deletable feature branch. See P1 below. |
| `docs/bad-smell.md` | §7 Fail fast rather than fall back | Pass | No `??`/`||` default and no tolerance branch added; the conditional is a feature gate, not a fallback. |
| `docs/fallback.md` | §2 A feature behind a switch needs no compatibility code | Pass | Nothing dual-reads, dual-writes, or tolerates an older shape of this feature; the switch is the only branch. |
| `docs/fallback.md` | §1 No negative tests asserting removed behavior stays removed | N/A | The diff contains no tests at all. |
| `docs/testing.md` | Integration tests at entry points, not unit tests of internal functions | N/A | No tests added; the gap is covered by the P0 above rather than by a convention violation. |
| `docs/testing.md` | Mock at the boundary; MSW for HTTP; real database | N/A | No test files in the diff, so no mock boundary is introduced. |
| `.claude/skills/ccstate/SKILL.md` | Commands own a complete business transition; async work carries an `AbortSignal` | Pass | `importPresentationTemplateDeck$` is unchanged apart from the prompt value it sets; it still threads `signal` and calls `signal.throwIfAborted()` after each await. |
| `.claude/skills/ccstate/SKILL.md` | Computed memoization — no manual cache | Pass | `presentationTemplateImportPrompt()` is a pure function called at send time; no new signal or cache is introduced. |
| `docs/effect.md` | Review checklist — signal factories outside render, state derived not synchronized | Pass | The new module exports one pure function; no signal is created, and no React render path is touched. |
| `docs/cache.md` | RC-5 No duplicate mutable sources of truth | Pass | The instruction string has one definition in `@okouai/core`; both call sites read it rather than restating it, which is what keeps the two entry points from drifting. |

### Findings

- **No tests for a `feat:` change on either surface.** The diff adds user-visible behavior in two places and asserts neither:
  - `turbo/apps/api/src/signals/services/zero-runs-create.service.ts` — a new line in the standing agent-tools prompt. This prompt is already assertable from an existing integration test: `turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts` pins the `workflow-setup` line today. Add the switch-on and switch-off assertions there.
  - `turbo/apps/platform/src/signals/zero-page/presentation-template-import.ts` — the import message now carries the guide pointer. `chat-composer-template-picker.test.tsx` asserts `toContain("reusable presentation template")`, which matches the first sentence and therefore passes unchanged by this diff. Nothing proves the pointer is sent.

  Since both paths are switch-gated and the switch is off, this is not a production risk today — but nothing will catch a regression at the moment the switch is turned on, which is the only moment the code runs.

- `turbo/packages/core/src/presentation-template-skill.ts`: `PRESENTATION_TEMPLATE_SKILL_BRANCH` pins `feat/reverse-template-skill`, a working branch that will be deleted when that work lands or moves. The doc comment names the intended end state ("Moving the guide into `vm0-skills` later turns this constant into a plain skill name") but links no tracking issue, so nothing forces the pin to be revisited. When the branch is deleted the instruction keeps telling runs to clone a ref that no longer exists, and the failure surfaces inside a run rather than in CI. Please link a follow-up issue in that comment.

### Testing

- Coverage: **Insufficient** — missing tests for: the switch-gated agent-tools prompt line (`zero-runs-create.service.ts`), and the guide pointer in the import message (`presentation-template-import.ts`).
- Conventions: N/A — the diff contains no test files.

### Notes (not blocking)

- The guide repository is private, so the emitted instruction only resolves for runs whose GitHub access covers a vm0 private repository. This is intentional for the current iteration phase and is contained by the switch; the constant's doc comment states it plainly. Worth keeping in view as the removal condition for the P1 above.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 01d426184bab1f1553cc9b911f1f46d83bb5d32f
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Adds one shared instruction naming where the deck reverse-engineering guide lives, emitted from the template picker's import message and from the standing agent-tools prompt, both gated on `PresentationTemplates`. Since the previous review, `01d4261` adds the coverage that was missing on both entry points and links the follow-up that keeps the branch pin from becoming permanent.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | `feat:` commits must have integration tests | Pass | `run-lifecycle.bdd.test.ts` gains a run-claim test asserting the prompt across both switch states; `chat-composer-template-picker.test.tsx` asserts the pointer the import message carries. |
| `REVIEW.md` | Feature regression guard — changed existing behavior needs regression coverage | Pass | The import message is existing behavior this PR changes, and its assertion no longer passes vacuously: it previously matched only the unchanged first sentence. |
| `REVIEW.md` | New user-facing behavior gated behind a `FeatureSwitchKey` | Pass | Both emissions gated on `PresentationTemplates`; the API test now proves the gate is closed when off, not just open when on. |
| `REVIEW.md` | Correctness — no wrong assumptions in threaded arguments | Pass | `featureSwitchContext` is already a field of `ZeroRunAfterPreCreate` and `attemptInput` spreads `input`, so the new parameter is a declaration rather than new plumbing. |
| `REVIEW.md` | Type safety — no `any`, no suppressions | Pass | None in the diff; the new parameter is `boolean`, the new module returns `string`. |
| `docs/bad-smell.md` | §6 Never hardcode environment-specific values | Pass (with follow-up) | Repo, branch, and path are literals. Hardcoding a skill source has precedent (`github-url.ts` `DEFAULT_SKILLS_OWNER/REPO/BRANCH`), and the difference — a deletable working branch rather than `main` — is now tracked by #28374, linked from the constant's doc comment. |
| `docs/bad-smell.md` | §7 Fail fast rather than fall back | Pass | No `??`/`\|\|` default and no tolerance branch; the conditional is a feature gate. |
| `docs/fallback.md` | §2 A feature behind a switch needs no compatibility code | Pass | Nothing dual-reads, dual-writes, or tolerates an earlier shape of this feature. |
| `docs/fallback.md` | §1 No negative tests asserting removed behavior stays removed | Pass | The new `not.toContain` asserts that a *live* gate is closed, which is current product behavior, not a tombstone for deleted code. Removing it would leave half the gate unverified. |
| `docs/testing.md` | Integration tests at entry points, not unit tests of internal functions | Pass | The API test goes through run creation and a runner claim rather than calling `buildAgentToolsPrompt`; the platform test drives the real composer. |
| `docs/testing.md` | Mock at the boundary; MSW for HTTP; real database; no fake timers | Pass | Both tests use existing bdd helpers; no `vi.mock` with a relative path, no `fetch` stubbing, no `vi.useFakeTimers`. |
| `.claude/skills/ccstate/SKILL.md` | Commands own a complete business transition; async work carries an `AbortSignal` | Pass | `importPresentationTemplateDeck$` is unchanged apart from the prompt value; it still threads `signal` and calls `throwIfAborted()` after each await. |
| `.claude/skills/ccstate/SKILL.md` | Computed memoization — no manual cache | Pass | `presentationTemplateImportPrompt()` is a pure function called at send time; no signal or cache added. |
| `docs/effect.md` | Review checklist — signal factories outside render, state derived not synchronized | Pass | The new module exports one pure function; no signal creation and no React render path touched. |
| `docs/cache.md` | RC-5 No duplicate mutable sources of truth | Pass | The instruction has one definition in `@okouai/core`; both entry points read it, which is what prevents them drifting apart. |

### Findings

No P0 or P1 findings. Both items from the previous review on `f18cb85` are resolved:

- **P0 (missing tests)** — `run-lifecycle.bdd.test.ts` now claims a run with the switch off and again with it on, pinning both halves of the gate. That transition is the only moment this code runs, and it was previously uncovered. `chat-composer-template-picker.test.tsx` now asserts the pointer rather than the sentence that did not change.
- **P1 (branch pin without a tracking issue)** — the constant's doc comment links #28374, which covers moving the guide into `vm0-skills`, mounting it like `goal`, and reducing the constant to a skill name.

### Testing

- Coverage: **Adequate** — the switch-gated prompt line and the import-message pointer are both asserted, in both states for the API path.
- Conventions: **Compliant** — integration-level, existing bdd helpers, no relative-path mocks, no direct `fetch` mocking, no fake timers.
- Execution caveat: these two tests have **not been run locally**. `registry.npmjs.org` is firewalled in this sandbox (`424 connector_not_configured_for_run`), so `pnpm install` fails and no local check can run. They were written against verified call shapes already used in the same files (`entitledRunActor`, `api.createRun`, `api.claimRunnerJob`, `connectors.updateFeatureSwitches`), but CI is their first execution. `test-api` and `test-app` are pending on `01d4261` as of this review; if either rejects them, that is a defect in this PR and not a flake.

### Notes (not blocking)

- The guide repository is private, so the emitted instruction resolves only for runs whose GitHub access covers a vm0 private repository. Intentional for the current iteration phase, contained by the switch, stated in the constant's doc comment, and #28374 is the condition that removes it.
- Unrelated to this diff: `cli-e2e-02-playwright` has failed on this PR across runs, always inside `buyUsagePackPlan` in `billing-transitions.spec.ts`, on different lanes and different cases each time. The same lanes pass on recent `main` runs. Reported to the e2e channel; not treated as a finding against this PR.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: b870a22f71cc993b5ce1a090c35c3523a6070177
PR_AUTO_REVIEW_VERDICT: LGTM

Self-review: I am the author. Reviewed against `REVIEW.md` and the practice documents from `main`.

### Summary

Re-review of `b870a22f` after the rebase onto `main` replaced the previously reviewed `01d42618`. The change itself is unchanged in intent: a single module-private constant set in `@okouai/core/presentation-template-skill` names where the deck-reverse-engineering guide lives, and `presentationTemplateSkillInstruction()` is emitted from two places — the template picker's import message and the standing agent-tools prompt — so both entry points reach the same guide and cannot drift. The API side is gated by `FeatureSwitchKey.PresentationTemplates`, which is `enabled: false` with no `enabledOrgIdHashes` (`feature-switch.ts:311`), so the prompt is byte-identical for every run today.

**Why this re-review exists, and what it found.** The rebase was applied as a patch generated against the old base. `git apply` succeeded, but its context lines carried pre-`e7efc606` content and silently reverted someone else's change in `zero-runs-create.service.ts` — removing the `CANONICAL_CLAUDE_CONFIG_DIR` import and restoring a hard-coded `/home/user/.claude/skills` literal. The first repair pulled the file from `origin/main`, which had advanced past the merge base, and so imported that foreign change back in as if it were mine — the same defect with the sign flipped. The final repair restores the file from the merge base `e7efc606a` and re-applies only this PR's additions. Verified on `b870a22f`: `git diff $(git merge-base origin/main HEAD) HEAD` is exactly the five files below, and contains no removal of a line this PR did not add.

Changed surfaces: `turbo/packages/core`, `turbo/apps/api` (service + BDD test), `turbo/apps/platform` (signals + page test).

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no unhandled null/undefined paths | Pass | `buildZeroCreateAgentRunArgs` reads `args.featureSwitchContext`, which `ZeroRunAfterPreCreate` already declares (`zero-runs-create.service.ts:937`) and `attemptInput` propagates by spreading `input` (`:997`), so the new required field is satisfied at the single call site (`:1021`). |
| `REVIEW.md` | Security — no hardcoded secrets, no missing auth on new surfaces | Pass | No new endpoint, no credential. The only new data is a public repo coordinate; access is the run's own GitHub identity. |
| `REVIEW.md` | Type safety — no `any`, no broad casts | Pass | Three `string` constants, one `(): string` function, one `boolean` field threaded through four existing signatures. |
| `REVIEW.md` | Feature regression guard — existing behavior changed only under containment, with coverage for the pre-existing path | Pass | The agent-tools prompt is shared by every run. The new line is inside `...(args.presentationTemplatesEnabled ? [...] : [])`, and `run-lifecycle.bdd.test.ts:1590-1600` asserts the switched-off claim still contains `# Agent Tools` and does not contain the repo — the existing flow, not only the new one. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, migration, contract, or `jsonb().$type<>()` change; `appendSystemPrompt` keeps its existing `z.string().nullable()` shape. |
| `docs/bad-smell.md` | Zero `any`, zero lint/type suppressions, no dynamic `import()` | Pass | None present in the diff; both new imports are static and top-of-file. |
| `docs/bad-smell.md` | No hardcoded URLs or environment-specific configuration | Pass, with note | The repo, branch, and path are hardcoded. They are not environment-specific — the same three values apply in every environment, so there is no `env()` entry to read from. The pin is deliberate (the guide is still being revised against real decks), documented in `presentation-template-skill.ts:1-15` with the condition that removes it, and tracked by vm0-ai/vm0#28374. Worth restating plainly: a branch pin in shipped code is only acceptable because the switch that reaches it is off. |
| `docs/bad-smell.md` | Fail fast — no fallback or recovery logic | Pass | No `??`/`||` chain and no try/catch added in production code. A run without access to the repo fails at `gh repo clone`, visibly. |
| `docs/fallback.md` §2 | A feature behind a non-GA switch needs no compatibility code | Pass | `PresentationTemplates` is `enabled: false` with no staff hashes, so no dual-read, migration, or cutover tolerance is required — and none is added. |
| `docs/fallback.md` §3 | No `??`/`||` on a state the owning contract cannot produce | Pass | The only `??` in the diff is `claim.appendSystemPrompt ?? ""` in the API test. The contract declares it `z.string().nullable()` (`runners.ts:460`), so the null is real, and the same idiom already appears at roughly fifteen sites in that file. |
| `docs/fallback.md` §1 | No negative tests asserting removed behavior stays removed | Pass | The `not.toContain` assertion pins the switched-off state of live behavior, not the absence of deleted code. |
| `docs/fallback.md` §9 | Every fallback introduced by the diff is declared in the PR summary | N/A | The diff introduces no fallback, so no `Fallbacks` section is required and none is claimed. |
| `docs/deployment-compatibility.md` | Cross-version safety at a deployable boundary | Pass | The field's shape is unchanged; only its string content varies, and only while the switch is on. An old runner receiving one extra prompt line needs no tolerance branch. |
| `docs/testing.md` | Integration tests, mostly; no unit tests for internal functions | Pass | Both tests go through the outermost surface — the API contract and the rendered composer. Neither imports `presentationTemplateSkillInstruction` to assert its return value. |
| `docs/testing.md` | `feat:` commits must have integration tests | Pass | Two: the API gate in both states, and the picker message content. |
| `docs/testing/api-testing.md` | Construct and verify through the API; do not touch DB or services | Pass | `api.createRun`, `api.heartbeatRunner`, `api.claimRunnerJob`, and `connectors.updateFeatureSwitches` — the last goes through the real `featureSwitchesRoutes` via `setupApp` (`helpers/api-bdd-connectors.ts:1838`). No schema import, no row write. |
| `docs/testing/api-testing.md` | Shared persistent state — tests own unique identities | Pass | The actor, agent, and runner group come from `entitledRunActor()`; the switch is written per-actor, so a concurrent file cannot observe it. |
| `docs/testing/testing-external-behavior.md` | Assert on the external interface, not internals | Pass | The API test asserts on the claim payload the runner receives; the platform test asserts on the message the composer actually sends. |
| `docs/testing/app-testing.md` | Page-level tests entering through `detachedSetupPage` | Pass | The assertions extend the existing template-picker page test at its established entry point; no component is rendered directly. |
| `docs/testing/anti-patterns.md` | AP-1 no `toHaveBeenCalled()` as the assertion; AP-4 no internal `vi.mock()`; AP-5 no fake timers | Pass | The diff adds only content assertions on real payloads; no mock, no timer. |
| `docs/testing/patterns.md` | Standard structure and MSW for HTTP | Pass | Both files follow the surrounding structure and existing mock infrastructure; nothing new is stubbed. |
| `.claude/skills/ccstate/SKILL.md` | `get`/`set` must not escape the command callback | Pass | `presentationTemplateImportPrompt()` takes no accessors and reads no signal — it is a pure string builder, so no sub-command is warranted. |
| `.claude/skills/ccstate/SKILL.md` | `signal.throwIfAborted()` after an await that does not take the signal | Pass | Unchanged from the reviewed parent; the added call is synchronous and sits between existing checks. |
| `docs/effect.md` | Choose the right execution boundary; keep render pure | Pass | The change stays inside the existing `importPresentationTemplateDeck$` command. No effect, no render-phase work. |
| `docs/cache.md` | RC-1 to RC-9: no render-phase side effects, no unbounded caches, no lifetime mismatch | Pass | Nothing is cached or memoized; the const became a function, which removes module-load-time evaluation rather than adding state. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

### Testing

- **Coverage: adequate.** `run-lifecycle.bdd.test.ts` pins both halves of the gate in one test — a claim taken before the switch is enabled must still carry `# Agent Tools` and must not name the repo; a claim taken after must name the repo, the path, and the branch. That last assertion is what makes the branch pin a tracked fact rather than a silent string. `chat-composer-template-picker.test.tsx` now asserts the sent prompt contains `vm0-ai/Template-artifact` and `reverse-template`; the pre-existing `toContain("reusable presentation template")` matched only the unchanged first sentence, so it would have passed with the pointer missing entirely.
- **Conventions: compliant.** No internal mocks, no fake timers, no `toHaveBeenCalled()`, no direct DB access, no direct component rendering.
- **Local validation still unavailable.** `registry.npmjs.org` continues to return `424 connector_not_configured_for_run` in this sandbox, so `pnpm install` fails and format, lint, check-types, knip, and vitest cannot run locally — `npx lefthook install` fails for the same reason. Every claim above is read from the source and the contracts, not from a green local run. CI on `b870a22f` is the first execution of these tests. The call-site facts most likely to break a build were checked by hand: `featureSwitchContext` reaches `buildZeroCreateAgentRunArgs` through the existing spread, and `FeatureSwitchKey` and `createConnectorBddApi` were already imported in the test file, so the new test adds no import.

PR #28363, opened 2026-08-20
fix(runner): reconcile stale drain override cleanup

### LINKED ISSUE #28295: refactor(rust): centralize drain override reconciliation
## Finding

At verified commit `002d5c31fa8096a58dedf85b4d72f7e0b8b69f3e`, runner service lifecycle code maintains two implementations of the same drain-override cleanup transaction:

- `crates/runner/src/cmd/service/mod.rs:224-250` removes the runtime `Restart=no` drop-in, conditionally coordinates a daemon reload, and restores the drop-in if that reload fails.
- `crates/runner/src/cmd/service/stop.rs:148-211` repeats the same removal, reload, restoration, and combined-error handling with bounded systemd operations.
- `crates/runner/src/cmd/service/reload.rs:41-47` is already the canonical expected-state check: `with_drain_override(false)` forces a reload when systemd still reports the VM0 drop-in in `DropInPaths`, even if `NeedDaemonReload` is false.

The duplicated wrappers have drifted in a correctness-relevant way. The unbounded helper returns immediately when `remove_drain_restart_override` reports `false` (`mod.rs:227-229`), so it never asks the coordinator whether systemd still has the drop-in loaded. The bounded helper always invokes the coordinator (`stop.rs:151-160`) and can reconcile that stale loaded state.

This difference is reachable across cancellation or process interruption:

1. The unbounded helper removes the on-disk drop-in synchronously.
2. The task or process stops while awaiting the reload lock/query/command.
3. A retry sees no on-disk file or directory, so removal returns `false`.
4. The unbounded helper reports success without checking `DropInPaths`, leaving systemd's effective `Restart=no` policy loaded.

`service start` uses the unbounded helper at `crates/runner/src/cmd/service/mod.rs:371`, and normal `service stop` reaches it through `RealServiceStopOps` at `crates/runner/src/cmd/service/stop.rs:325-333`. Cleanup stop uses the bounded implementation at `stop.rs:336-344`, so retry behavior currently depends on which command mode owns the transition.

This is distinct from #23532, which preserved the removal result when empty-directory cleanup fails within one call. A complete file removal followed by cancellation before reload still loses the mutation on the next call. PR #24109 also had to update both cleanup implementations when reload coordination changed, demonstrating the ongoing maintenance cost of the split ownership.

## Structural impact

The service layer has two owners for one transaction: remove VM0's drain override, make the expected absence effective in systemd, and compensate if the reload fails. Their error policy and timeout mechanism differ, but the state transition does not. Keeping the transition duplicated allows normal and bounded service paths to accumulate different retry and recovery semantics.

The canonical owner should be one drain-override reconciliation helper in the service lifecycle layer, using the existing normal or bounded reload coordinator as an execution policy rather than reimplementing the transaction in `mod.rs` and `stop.rs`.

## Acceptance criteria

- Maintain one canonical removal -> expected-absence reload -> failure-compensation transaction for both normal and bounded callers.
- Always consult the reload coordinator with `with_drain_override(false)`, including when the on-disk override is already absent; let the coordinator skip work when the effective state already matches.
- Preserve bounded lock/query/command timeouts for cleanup mode and preserve combined removal, reload, and restoration error context.
- Restore the on-disk override after reload failure only when the current attempt actually removed it.
- Add focused coverage for an absent on-disk override that remains present in systemd `DropInPaths`, proving both normal and bounded paths converge on retry.
- Keep `service start`, normal stop, and cleanup stop behavior idempotent across cancellation between file removal and daemon reload.

seven332:
## Summary

- centralize start and stop drain-override cleanup in one private lifecycle transaction
- always reconcile expected override absence, including retries where the on-disk file is already gone
- distinguish override-file removal from empty-directory cleanup so compensation never synthesizes `Restart=no`
- preserve bounded cleanup timeouts, conditional compensation, and combined failure context
- cover the interrupted retry state through both reload policies and the actual `DropInPaths` coordinator logic

## Behavior

`service start`, default `service stop`, and cleanup-mode stop now use the same removal, reload, and compensation transaction. The transaction retains a structured removal outcome, but always asks the reload coordinator to make expected absence effective. A retry after interruption can therefore reload a stale systemd `Restart=no` policy even when file removal is already idempotently complete.

When effective state already matches, the existing host-global coordinator performs only the state query and skips `daemon-reload`. Cleanup mode keeps its 20-second reload-lock timeout and 10-second systemd-command timeout. Reload failure recreates the override only when the current attempt removed the override file; removing only an already-empty drop-in directory still preserves the legacy reload signal without permitting compensation. Restoration uses the same bounded or unbounded policy as the failed cleanup.

Install, uninstall, drain, resume, CLI shape, persisted state, and external protocols are unchanged.

## Validation

- `cargo fmt --manifest-path crates/Cargo.toml --all`
- `cargo test --manifest-path crates/Cargo.toml -p runner --bin runner cmd::service::drain_override` (14 passed)
- `cargo test --manifest-path crates/Cargo.toml -p runner --bin runner cmd::service::reload` (6 passed)
- `cargo test --manifest-path crates/Cargo.toml -p runner --bin runner cmd::service` (243 passed, 1 ignored child fixture)
- `cargo test --manifest-path crates/Cargo.toml -p runner --all-targets --all-features` (3,209 binary tests passed, 20 ignored fixtures; 1 integration test passed)
- `git diff --check`

Closes #28295

### DIFF
--- a/crates/runner/src/cmd/service/drain_override.rs
+++ b/crates/runner/src/cmd/service/drain_override.rs
@@ -18,6 +18,23 @@ pub(super) enum DrainRestartOverrideWrite {
     Replaced,
 }
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(super) enum DrainRestartOverrideRemoval {
+    OverrideRemoved,
+    DirectoryRemoved,
+    AlreadyAbsent,
+}
+
+impl DrainRestartOverrideRemoval {
+    pub(super) const fn override_removed(self) -> bool {
+        matches!(self, Self::OverrideRemoved)
+    }
+
+    const fn reload_required(self) -> bool {
+        !matches!(self, Self::AlreadyAbsent)
+    }
+}
+
 pub(super) fn write_drain_restart_override(
     unit: &RunnerServiceUnit,
 ) -> RunnerResult<DrainRestartOverrideWrite> {
@@ -26,6 +43,12 @@ pub(super) fn write_drain_restart_override(
 
 /// Returns whether cleanup found state that warrants reloading systemd.
 pub(super) fn remove_drain_restart_override(unit: &RunnerServiceUnit) -> RunnerResult<bool> {
+    remove_drain_restart_override_outcome(unit).map(DrainRestartOverrideRemoval::reload_required)
+}
+
+pub(super) fn remove_drain_restart_override_outcome(
+    unit: &RunnerServiceUnit,
+) -> RunnerResult<DrainRestartOverrideRemoval> {
     remove_drain_restart_override_at(Path::new(RUNTIME_SYSTEMD_SYSTEM_DIR), unit)
 }
 
@@ -68,11 +91,14 @@ fn write_drain_restart_override_at(
     Ok(outcome)
 }
 
-fn remove_drain_restart_override_at(root: &Path, unit: &RunnerServiceUnit) -> RunnerResult<bool> {
+fn remove_drain_restart_override_at(
+    root: &Path,
+    unit: &RunnerServiceUnit,
+) -> RunnerResult<DrainRestartOverrideRemoval> {
     let path = drain_restart_override_path_at(root, unit);
     cleanup_unit_staging_files(&path)?;
 
-    let mut changed = match std::fs::remove_file(&path) {
+    let override_removed = match std::fs::remove_file(&path) {
         Ok(()) => true,
         Err(e) if e.kind() == ErrorKind::NotFound => false,
         Err(e) => {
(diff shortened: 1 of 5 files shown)


seven332:
## Code Review: PR #28363

**Reviewed head:** `7669cadadb734079494a367d14b7adbf3b49d2dd`

### Summary

The current head gives start, default stop, and cleanup stop one private drain-override reconciliation transaction while preserving bounded timeouts and lock ordering. The structured removal outcome correctly separates reload necessity from ownership of the removed override, so directory-only cleanup cannot synthesize `Restart=no` during compensation.

### Findings

No findings.

### Testing

Reviewed focused transaction and real temporary-filesystem coverage, stale `DropInPaths` coordinator coverage for both policies, 243 service tests, Clippy, docs, formatting, and the complete runner suite (3,209 binary tests plus 1 integration test). Real-systemd and broader repository checks remain delegated to GitHub CI, which is still in progress.

### Verdict

LGTM

PR #28390, opened 2026-08-20
fix(python): reject oversized firewall auth expiries

### LINKED ISSUE #28377: bug(python): reject oversized integer firewall auth expiries
## Problem

The firewall auth success parser accepts any Python `int` for `expiresAt`, but the cache later passes that value to `math.isfinite`. Python converts integers to a float for this call, so a valid JSON integer that is too large for a float raises `OverflowError` instead of being handled as an invalid expiry.

Relevant code:

- `crates/runner/mitm-addon/src/firewall_auth_client.py:754-766` validates finite floats but returns integers without a range/finite check.
- `crates/runner/mitm-addon/src/firewall_auth_cache.py:297-302` calls `math.isfinite(value)` for both integers and floats.
- `crates/runner/mitm-addon/src/firewall_auth_cache.py:330-340` evaluates cached non-billable expiries.
- `crates/runner/mitm-addon/src/firewall_auth_cache.py:350-366` evaluates fresh billable expiries.

## Reproduction

A 401-digit integer is accepted by the same JSON decoder and parser used by the auth client:

```python
body = '{"expiresAt":1' + ('0' * 400) + '}'
expiry = _parse_expires_at(json.loads(body))
_has_valid_expiry(expiry)
```

Observed result:

```text
OverflowError: int too large to convert to float
```

The cache behavior is path-dependent:

- A non-billable request succeeds and caches the response on the first fetch, then raises `OverflowError` on the next cache lookup.
- A billable request raises `OverflowError` on the first fetch, bypassing `InvalidBillableAuthExpiryError`.

The existing invalid-expiry tests cover `None`, booleans, strings, infinity, NaN, and expired values, but not oversized JSON integers.

## Impact

A malformed but JSON-valid firewall auth response can escape the intended bounded validation path and fail requests with an unexpected exception. Non-billable responses additionally place the invalid value in cache before the failure appears, making the first request and subsequent requests behave differently.

## Acceptance criteria

- Oversized integer `expiresAt` values cannot cause `OverflowError` in parsing or cache validation.
- Unsupported numeric values are rejected through the existing malformed-success or invalid-billable-expiry error paths.
- Expiry validation remains total for arbitrary decoded response values.
- Regression tests cover a fresh billable response and a non-billable cache lookup with an oversized integer expiry.

seven332:
## Summary

- centralize supported firewall auth expiry validation across response parsing and cache checks
- reject oversized integer expiries through the existing malformed-success and billable fail-closed paths
- cover parser handling, non-billable cache refetch, and fresh billable rejection regressions

## Compatibility

- preserves supported integer and finite-float expiry values without normalization
- does not change the firewall auth wire shape, cache identity, persisted state, or runner protocol
- does not add a migration, feature switch, or rollout requirement

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_firewall_auth_handling.py -k oversized`
- `uv run --no-sync python -m pytest tests/test_auth_cache.py -k oversized`
- `uv run --no-sync python -m pytest tests/` (6426 passed)

Closes #28377

### DIFF
--- a/crates/runner/mitm-addon/src/firewall_auth_cache.py
+++ b/crates/runner/mitm-addon/src/firewall_auth_cache.py
@@ -1,7 +1,6 @@
 """Firewall auth cache, fetch admission, and force-refresh lifecycle."""
 
 import asyncio
-import math
 import time
 from dataclasses import dataclass, field
 
@@ -10,6 +9,7 @@ from firewall_auth_client import (
     FirewallAuthPayload,
     FirewallAuthRequest,
     fetch_firewall_headers,
+    is_supported_expiry,
 )
 
 
@@ -295,9 +295,7 @@ def auth_state_is_empty_for_tests() -> bool:
 
 
 def _has_valid_expiry(value: object, now: float | None = None) -> bool:
-    if isinstance(value, bool) or not isinstance(value, int | float):
-        return False
-    if not math.isfinite(value):
+    if not is_supported_expiry(value):
         return False
     return (time.time() if now is None else now) < value
 
--- a/crates/runner/mitm-addon/src/firewall_auth_client.py
+++ b/crates/runner/mitm-addon/src/firewall_auth_client.py
@@ -15,7 +15,7 @@ import urllib.request
 from contextlib import suppress
 from dataclasses import dataclass, field, replace
 from email.message import Message
-from typing import NamedTuple, Protocol
+from typing import NamedTuple, Protocol, TypeGuard
 
 import h11
 import mitmproxy_rs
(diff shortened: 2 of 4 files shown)


seven332:
## Code Review: PR #28390

**Reviewed head:** `8d41cb2047da4cf6e1ad2782fa708977f2373eb5`

### Summary

The shared typed predicate makes parser and cache numeric validation consistent, converts only the known oversized-integer `OverflowError` into invalid input, and preserves valid timestamp values and existing error mappings.

### Findings

No findings. The self-review's test-structure finding was resolved on this head: oversized cache regressions now run through non-billable lookup and fresh billable fetch entry points instead of adding a private-helper assertion.

### Testing

Ruff format/lint and BasedPyright passed. Focused oversized parser/cache tests passed, and the complete mitm-addon suite passed 6426 tests with 59 existing third-party warnings. Unrelated Rust/Turbo suites were not run; GitHub CI remains to be checked separately.

### Verdict

LGTM

PR #28395, opened 2026-08-20
perf(runner): bound local queue job reads

### LINKED ISSUE #28383: perf(rust): bound local queue job reads before allocation
## Problem

Local queue job files have no protocol-level size limit, and the shared reader grows a `Vec` until EOF:

- `crates/runner/src/local_queue/fs.rs:123-128` opens a validated private file and calls `read_to_end` without `take`, a metadata precheck, or a caller-provided maximum.
- The 100 ms local discovery loop (`crates/runner/src/provider/local/mod.rs:26-27,90-96,119-137`) uses that helper to read and parse the complete first eligible `.job` merely to extract `reuse_key` (`crates/runner/src/local_queue/state.rs:181-203`).
- After the atomic claim succeeds, `claim_job_sync` opens and reads the same job completely again before deserializing `JobRequest` (`crates/runner/src/local_queue/state.rs:380-418`).
- `runner local submit` serializes and publishes `request_json` without validating its byte length (`crates/runner/src/cmd/local/submit.rs:394-420,627-645`). Manually created or corrupted private regular job files are likewise accepted without a size boundary.

Moving these transactions to `spawn_blocking` in #16096 protects Tokio workers, but it does not bound process heap growth or total filesystem work. The later discovery optimizations in #20212 and #20238 also do not bound the selected job's contents.

## Impact

An accidentally oversized or malformed `.job` makes each polling runner read and allocate the entire file before any runner reaches the `O_EXCL` claim. The winning runner then reads and allocates it a second time. A very large or sparse file can therefore consume unbounded I/O and heap, stall every runner watching the group, or abort a runner on allocator exhaustion before the existing poison-job result/cleanup path can run. Multiple local runners amplify the pre-claim read independently.

## Suggested direction

Define an explicit maximum serialized local job size, enforce it when `runner local submit` builds the request, and make the secure private-file reader accept a caller-specific maximum. Read at most `max + 1` bytes through the already-opened descriptor and reject overflow; a metadata check may be used as an early exit but should not be the only bound because the file can change after `stat`.

Discovery should avoid reading beyond that same maximum. An oversized candidate should still reach one authoritative claim/terminal-failure path (or an equivalent durable quarantine) so it cannot remain the oldest candidate and be retried every 100 ms.

## Acceptance criteria

- Local `.job` files have a documented byte limit that accommodates supported prompt and environment payloads.
- `runner local submit` rejects an over-limit serialized request before publishing a `.job` file.
- Discovery and claim never read or allocate more than the configured limit plus a constant overflow byte for one job file.
- An oversized published job produces a durable failure/cleanup outcome and does not cause repeated discovery work.
- Tests cover exact-limit acceptance, one-byte-over rejection, and an oversized/sparse job without reading or allocating the full file.
- Existing private-file ownership, mode, `O_NOFOLLOW`, atomic publish, and atomic claim guarantees remain unchanged.

seven332:
## Summary

- define a shared 16 MiB limit for serialized local queue job requests and reject oversized submits before publication
- bound discovery and claim reads with opened-descriptor metadata checks plus a `max + 1` content read limit
- route oversized legacy or manual jobs through the existing atomic claim and durable terminal cleanup path
- cover exact-limit publication, one-byte-over rejection, and a 1 GiB sparse poison job

## Scope

This PR only changes `.job` request limits. Local result and active-input files retain their existing behavior because their size contracts and recovery semantics are independent.

## Validation

- `cargo test -p runner serialized_job_ -- --nocapture`
- `cargo test -p runner claim_terminally_fails_oversized_sparse_job -- --nocapture`
- `cargo test -p runner` (3205 passed, 20 ignored)
- pre-commit workspace Rust format, Clippy, Rustdoc, and file-size checks

Closes #28383

### DIFF
--- a/crates/runner/src/cmd/local/submit.rs
+++ b/crates/runner/src/cmd/local/submit.rs
@@ -410,6 +410,13 @@ impl SubmitPlan {
 
         let request_json = serde_json::to_vec(&request)
             .map_err(|e| RunnerError::Internal(format!("serialize request: {e}")))?;
+        if request_json.len() > local_queue::LOCAL_JOB_MAX_BYTES {
+            return Err(RunnerError::Config(format!(
+                "serialized local job request exceeds {} bytes (got {} bytes)",
+                local_queue::LOCAL_JOB_MAX_BYTES,
+                request_json.len()
+            )));
+        }
         let queue = SubmitQueueEntry::for_job(&group_dir, &profile, job_id)?;
 
         Ok(Self {
--- a/crates/runner/src/cmd/local/submit/tests/validation.rs
+++ b/crates/runner/src/cmd/local/submit/tests/validation.rs
@@ -1,10 +1,17 @@
 use std::process::ExitCode;
 
-use super::super::{MAX_LOCAL_SUBMIT_TIMEOUT_SECS, SubmitArgs, run_submit_with_home};
+use super::super::{MAX_LOCAL_SUBMIT_TIMEOUT_SECS, SubmitArgs, SubmitPlan, run_submit_with_home};
 use super::support::{run_submit_and_write_success, submit_args_for_test};
 use crate::error::RunnerError;
 use crate::local_queue;
 use crate::paths::HomePaths;
+
+fn prompt_len_for_serialized_job_size(root: &std::path::Path, target_bytes: usize) -> usize {
+    let mut args = submit_args_for_test();
+    args.prompt.clear();
+    let plan = SubmitPlan::from_args(args, HomePaths::with_root(root.to_path_buf())).unwrap();
+    target_bytes.checked_sub(plan.request_json.len()).unwrap()
+}
 
 #[tokio::test]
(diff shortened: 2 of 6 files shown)


seven332:
## Code Review: PR #28395

**Reviewed head:** `a59805443b3964c57ab1183226824166594d141c`

### Summary

The producer and consumer now share an inclusive 16 MiB serialized local-job limit. Discovery and claim use descriptor metadata plus a `max + 1` read cap, while oversized poison jobs retain atomic claim and durable terminal cleanup behavior. The queue format is unchanged, so mixed runner versions remain compatible.

### Findings

No findings.

### Testing

Exact-limit submission, one-byte-over rejection before publication, and a 1 GiB sparse oversized job are covered through real filesystem and production submit/provider entry paths. Rust format, Clippy, Rustdoc, focused regressions, and the full runner suite passed locally (3205 passed, 20 ignored; guest inventory 1 passed).

### Verdict

LGTM

PR #28491, opened 2026-08-21
fix(platform): keep deck import alive across the new-thread navigation

bingjiez666:
## Problem

Importing a deck through **Import your own deck** from the *new-thread* composer (`/agents/:agentId/chat`) produces a thread the server never recorded: the page navigates into it, but no thread row and no run exist. The user sees a thread that never replies, and the console shows

```
GET /api/okou/chat-threads/<id>/event-rows?sinceSeqId=0&limit=50  404 (Not Found)
[W][Realtime] ably catch-up failed ApiError: Chat thread not found
```

Confirmed against a real thread id from a reported session: `okou chat messages --thread-id <id>` also returns `404: Chat thread not found`, so the row genuinely does not exist — this is not a transient catch-up race.

## Root cause

`PptImportCard` bound the import to `pageSignal$`. Every other control in the same composer uses `rootSignal$` (ordinary send at `zero-chat-composer.tsx:7402`, attachment upload at `:7226`).

The import reuses the ordinary send path, which for a new thread is `sendNewThread$`:

1. `optimistic-chat-thread-page.ts:620` fires `POST /chat-threads` with the caller's signal and does **not** await it
2. `:663` immediately navigates to `/chats/:threadId`
3. `:648` sends `POST /chat/events` only after the create resolves

`pageSignal$` is the current route's signal (`route.ts:272`), and `loadRoute$` aborts the previous route's controller through `resetRouteSignal$` (`route.ts:102`) — `navigate$` even documents this at `route.ts:205-211`. So the navigation performed by the send aborts the create that is still in flight, and the run request that is sequenced after it never goes out at all.

Nothing surfaces: `detach` swallows `AbortError` (`utils.ts:45-52`). The only trace is the realtime warning plus `notifyRealtimeDegraded$` (`realtime.ts:500`).

## Fix

Own `rootSignal$` in `PptImportCard`, consistent with the ordinary send and upload controls. The import is a message that deliberately takes the user off the current page, so its lifetime belongs to the app, not the page.

## Why CI did not catch it

Both existing import tests set up on `/chats/:threadId` (an existing thread), where sending does not navigate and the page signal is never aborted. Added a regression test that imports from `/agents/:agentId/chat` and asserts the thread create and the run send both complete, with the run carrying the created thread id.

## Notes

Introduced by #28344 / #28362. Importing from an existing thread page was never affected.

### DIFF
--- a/turbo/apps/platform/src/views/zero-page/__tests__/chat-composer-template-picker.test.tsx
+++ b/turbo/apps/platform/src/views/zero-page/__tests__/chat-composer-template-picker.test.tsx
@@ -3491,7 +3491,7 @@ describe("chat composer templates", () => {
         createdThreadId = body.clientThreadId;
       },
       onSendRequest(body) {
-        sentThreadId = body.clientThreadId;
+        sentThreadId = body.threadId;
         sentPrompt = body.prompt;
       },
     });
--- a/turbo/apps/platform/src/views/zero-page/__tests__/chat-composer-template-picker.test.tsx
+++ b/turbo/apps/platform/src/views/zero-page/__tests__/chat-composer-template-picker.test.tsx
@@ -38,6 +38,8 @@ import {
   tabByText,
   presentationTemplateGridScrollContainer,
   mockActiveTemplateThread,
+  mockAgent,
+  mockOrgModelRoutes,
   trackTemplatePreviewImagePreloads,
   mockImmediateIdleCallback,
   mockUrlObjectMethods,
@@ -3474,6 +3476,64 @@ describe("chat composer templates", () => {
 
     await waitFor(() => {
       expect(sentPrompts).toStrictEqual(["Never mind the deck"]);
+    });
+  });
+
+  it("creates the thread and starts the run when importing from a new chat", async () => {
+    const user = userEvent.setup({ delay: null });
+    let createdThreadId: string | undefined;
+    let sentThreadId: string | undefined;
+    let sentPrompt: string | undefined;
+    mockOrgModelRoutes("claude-fable-5");
+    mockAgent();
+    mockChatLifecycle(context, {
+      onThreadCreate(body) {
+        createdThreadId = body.clientThreadId;
+      },
+      onSendRequest(body) {
+        sentThreadId = body.clientThreadId;
+        sentPrompt = body.prompt;
+      },
+    });
+    context.mocks.upload.success({
+      id: "upload-deck",
+      filename: "brand-system.pptx",
+      contentType:
+        "application/vnd.openxmlformats-officedocument.presentationml.presentation",
+      size: 2048,
+      url: "https://cdn.vm7.io/artifacts/test/upload-deck/brand-system.pptx",
+    });
+
+    detachedSetupPage({
+      context,
+      featureSwitches: { [FeatureSwitchKey.PresentationTemplates]: true },
+      path: `/agents/${AGENT_ID}/chat`,
+    });
+
+    click(
+      await waitFor(() => {
+        return screen.getByLabelText("Template");
+      }),
+    );
+    const importInput = await waitFor(() => {
+      return screen.getByLabelText("Import your own deck");
+    });
+
+    await user.upload(
+      importInput,
+      new File(["deck"], "brand-system.pptx", {
+        type: "application/vnd.openxmlformats-officedocument.presentationml.presentation",
+      }),
+    );
+
+    // Importing from the new-thread composer creates the thread and then
+    // navigates into it. Both calls have to survive that navigation: a signal
+    // the navigation aborts leaves the user on a thread the server never
+    // recorded, with no run to answer them and nothing on screen saying so.
+    await waitFor(() => {
+      expect(createdThreadId).toBeDefined();
+      expect(sentPrompt).toContain("reusable presentation template");
+      expect(sentThreadId).toBe(createdThreadId);
     });
   });
 
--- a/turbo/apps/platform/src/views/zero-page/zero-chat-composer.tsx
+++ b/turbo/apps/platform/src/views/zero-page/zero-chat-composer.tsx
@@ -4820,6 +4820,12 @@ function IllustrationTemplateGrid({
  * Hands the chosen deck to the composer, which attaches it, sends it, and
  * navigates into the new thread. Importing is the ordinary chat path with the
  * message written for the user, not a separate upload flow.
+ *
+ * The import owns the root signal, like the ordinary send and upload controls:
+ * from the new-thread composer it outlives the page it started on. A page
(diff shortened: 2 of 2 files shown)


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 9ef8ce37d0be603ad704c245a3ac2211ccc80bb0
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

`PptImportCard` bound the deck import to `pageSignal$`, but the import reuses the ordinary new-thread send path, which fires `POST /chat-threads` without awaiting it and then navigates to `/chats/:threadId`. That navigation aborts the previous route's controller (`signals/route.ts:102`, documented at `signals/route.ts:205-211`), killing the in-flight create and preventing the run request that is sequenced after it, so the user lands on a thread the server never recorded. This PR moves the import to `rootSignal$`, matching the ordinary send (`zero-chat-composer.tsx:7402`) and attachment upload (`:7226`) controls in the same composer, and adds a regression test that imports from the new-thread page.

One non-blocking behavioral note: with the root signal, a slow deck upload that finishes after the user has navigated elsewhere will now still navigate them into the new thread. That is the feature's declared behavior ("attaches it, sends it, and navigates into the new thread") and matches the ownership the ordinary send already uses, so it is recorded as a trade-off rather than a finding.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic: no race conditions or API misuse | Pass | `zero-chat-composer.tsx:4838` — the navigation-vs-create race is the defect being removed; `rootSignal$` is aborted only at app teardown, so the create and the sequenced `POST /chat/events` both survive the route change |
| `REVIEW.md` | Type safety: no `any`, no broad casts | Pass | Diff introduces no casts; `useGet(rootSignal$)` returns `AbortSignal`, same type as the replaced binding |
| `REVIEW.md` | `fix:` commits must carry a regression test | Pass | `chat-composer-template-picker.test.tsx:3482` covers import from `/agents/:agentId/chat`, the path that was broken; it failed on the first CI run and passed after the assertion field was corrected |
| `REVIEW.md` | Feature regression guard: existing flow keeps coverage | Pass | The two pre-existing import tests on `/chats/:threadId` are unmodified by the diff and still pass; the change only alters which signal the import control owns |
| `REVIEW.md` | Security review (authz, injection, secrets) | N/A | No endpoint, auth check, credential, or user-controlled string handling in the diff |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, migration, contract, or `jsonb().$type<>()` change |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions, no dynamic `import()`, no hardcoded config URLs in production code | Pass | The production change is one binding plus a doc comment; the only URL literal is mock fixture data in the test, matching sibling fixtures |
| `docs/bad-smell.md` | Comments explain WHY, not WHAT | Pass | `zero-chat-composer.tsx:4823-4828` states why the import must outlive the page rather than restating the code |
| `docs/fallback.md` | Diff introduces no fallback; non-GA switch needs no compatibility path | Pass | No `??`/`||` default, compatibility branch, or negative test added; `presentationTemplates` is `enabled: false` with no org hashes (`turbo/packages/core/src/feature-switch.ts:311`), so §2 applies and no `Fallbacks` section is required |
| `docs/effect.md` | Make cancellation explicit: every async command needs a signal with a clear owner | Pass | `rootSignal$` is the app-lifecycle owner and does eventually abort; the operation is a one-shot upload/create/send, not a polling loop |
| `docs/effect.md` | Render stays pure; `detach` used only to forward a DOM event | Pass | `zero-chat-composer.tsx:4871` keeps the existing `detach(..., Reason.DomCallback)` inside `onChange`; render still only reads signals |
| `docs/cache.md` | RC-4 owner and lifetime mismatch: state must not default to root scope | Pass | Nothing is retained at root scope — the root signal is consumed by a single short-lived operation with no listener, timer, or subscription to release |
| `.claude/skills/ccstate/SKILL.md` | AbortSignal ownership; a send operation must survive page navigation rather than bind to `pageSignal$` | Pass | The doc's ownership section states this exception verbatim ("the send operation needs to survive page navigation — if bound to `pageSignal$`, the route change would abort the in-flight send request"); the general `pageSignal$` component default still governs page-scoped work, which this is not |
| `.claude/skills/ccstate/SKILL.md` | `detach()` only in the views layer | Pass | `detach` remains in the React `onChange` callback; `presentation-template-import.ts` is unchanged and still awaits its sub-commands |
| `docs/testing.md` | Integration/page-level tests only, no internal unit tests | Pass | The new test drives the real composer through the page, not `importPresentationTemplateDeck$` directly |
| `docs/testing.md` | No fake timers, no direct fetch mocking, no internal module mocks, no bare `toHaveBeenCalled()` | Pass | The test uses `userEvent`, `context.mocks`, and MSW-backed handlers; assertions read request content, not call counts |
| `docs/testing/patterns.md` | Pattern 7: enter through `detachedSetupPage`, configure every override through `context.mocks` before setup | Pass | `chat-composer-template-picker.test.tsx:3505-3509` sets up the page after `mockOrgModelRoutes`, `mockAgent`, `mockChatLifecycle`, and `context.mocks.upload.success` |
| `docs/testing/app-testing.md` | Page test placement and request-data capture for verifying what was sent | Pass | Test lives in `views/zero-page/__tests__/`; the "Tracking Request Data" pattern backs asserting on the captured create/send bodies |
| `docs/testing/anti-patterns.md` | AP-1/AP-7: assert outcomes, not call patterns or implementation details | Pass | Assertions check the created thread id, the prompt content, and that the run carries that same thread id — all of which break if the fix regresses |
| `docs/testing/testing-external-behavior.md` | Platform tests should observe from the surface a user can see | Pass, with a noted tension | The page cannot distinguish this defect: the optimistic user message renders identically whether or not the server recorded the thread, which is precisely why the bug was silent. The MSW request boundary is the nearest observable surface that separates the two, and the two pre-existing import tests in this file already assert the same way |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

### Testing

- Coverage: Adequate. The `fix:` commit adds the missing regression path — import from the new-thread composer at `/agents/:agentId/chat` — and asserts both that the thread was created and that the run request carries that thread id. The previously existing import tests only exercised `/chats/:threadId`, where the send does not navigate and the page signal is never aborted, which is exactly why CI was green on the broken code.
- Conventions: Compliant. Page-level entry through `detachedSetupPage`, all overrides through `context.mocks`, no fake timers, no internal mocks, no `server.use()`, no file-level `afterEach` cleanup, feature switch supplied through `featureSwitches`.
- CI on `9ef8ce3`: all eight `test-app` shards, `lint-eslint`, `lint-format`, `lint-types`, `lint-types-apps`, `lint-knip`, and `ci-gate-turbo` pass.

PR #28493, opened 2026-08-21
refactor(runner): centralize gc lock probes

### LINKED ISSUE #28472: refactor(rust): centralize runner GC lock probes
## Problem

Runner currently has two implementations of the same lock-path acquisition and replacement-race policy.

- `crates/runner/src/lock.rs:37-60` owns current-inode checks, `crates/runner/src/lock.rs:132-193` owns the bounded open/flock/retry loops, and `crates/runner/src/lock.rs:195-236` owns no-create opening and descriptor validation.
- `crates/runner/src/cmd/gc/lock_file.rs:30-54`, `crates/runner/src/cmd/gc/lock_file.rs:57-81`, and `crates/runner/src/cmd/gc/lock_file.rs:83-151` independently reimplement those same checks and acquisition loops for GC probes.

These are not different lock protocols: both operate on runner resource lock paths, perform nonblocking `flock`, reject results from a replaced pathname, and classify acquired, busy, missing, or failed probes. `crate::lock` is already the canonical owner used by normal runner callers.

## Concrete drift and maintenance cost

The two copies already encode different invariants:

- `lock.rs` retries a replaced path up to `LOCK_REPLACED_MAX_RETRIES = 64`; the GC probes use separate `MAX_STALE_INODE_RETRIES = 16` constants.
- The canonical existing-file path delegates to `host_file::secure_regular_private_file`, which validates the descriptor and normalizes an accepted mode to `0600`. GC has a local `validate_existing_lock_file_for_probe` copy that performs the type, owner, and writable-bit checks but does not apply the canonical normalization.
- Metadata failures during current-inode checks have different failure behavior: `lock.rs` treats an unavailable descriptor metadata check as current, while GC surfaces a probe error.
- The canonical existing-only opener classifies a missing parent as `Missing`, while the GC-local opener reports a validation error.

The history shows the cost of parallel ownership. PR #11155 first consolidated the normal `lock.rs` acquisition variants. PR #11554 then introduced the separate GC replacement loop, PR #17067 added its no-create variant, and PR #20094 later added the canonical no-create acquisition API in `lock.rs` without retiring the GC implementation. Subsequent lock acquisition and security changes therefore have two policy surfaces to inspect.

This can make the same runner lock path behave differently depending on whether it is reached through normal execution or GC, and a future hardening or pathname-race fix can be applied to only one implementation.

## Decision

- Consolidate create-if-missing and existing-only acquisition into one blocking state machine under `crate::lock`.
- Keep existing async lock APIs as blocking-pool wrappers and expose crate-local blocking result entry points for GC, so GC caller signatures, await points, guard lifetimes, and lock ordering do not change.
- Use the canonical 64-attempt replacement bound for both opening modes.
- Use `secure_regular_private_file` for both opening modes, including normalization of accepted existing files to `0600`.
- Fail closed when descriptor metadata or non-`NotFound` pathname metadata cannot establish identity. Treat pathname `NotFound` as stale and retryable. GC maps these errors to its existing preserve-and-skip behavior.
- Share the acquired-guard/path identity check used immediately before lock-file removal.

Asyncifying every GC probe was rejected because it would add a blocking-pool task per scanned lock and spread await/signature changes through every GC domain without evidence of a scheduler problem. Sharing only low-level helpers was rejected because it would leave the replacement-race state machine duplicated.

## Acceptance criteria

- One implementation under `crate::lock` owns opening mode (create or existing-only), nonblocking flock mode, current-path inode validation, replacement retries, and descriptor security validation.
- GC keeps its caller-specific best-effort logging and `Free`/`Held`/`Missing`/`Error` mapping, but delegates the acquisition state machine to that canonical implementation.
- Preserve the intentional distinction between probes that may create a lock path and probes that must not create a missing path.
- Use the shared retry bound of 64, normalize accepted existing lock files to `0600`, and fail closed when descriptor identity cannot be verified.
- Cover acquired, held, and missing outcomes; stale successful and contended inodes; retry exhaustion for both opening modes; and private-file validation through the shared implementation.
- Keep lock-file removal guarded by the acquired descriptor matching the current pathname.
- Preserve existing async API signatures and GC caller control flow.

## Risks and mitigations

- GC can perform up to 64 attempts rather than 16 under continuous pathname replacement; the loop remains bounded and returns the established replacement error.
- Existing-only probes may now change an accepted file mode to `0600`; this matches the canonical private-file contract and is covered explicitly.
- Normal lock callers will now receive an error for descriptor metadata failures instead of proceeding with unverified ownership; GC remains conservative and preserves data on that error.
- A deterministic internal race seam is limited to the shared module tests and does not become a production API.

## Audit provenance

Found during the daily Rust structural audit of 16 randomly selected files at `17d96ad7f324571e121833ed6c6e15b13258158f`.

seven332:
## Summary

- centralize create-if-missing and existing-only lock probes in one runner lock acquisition state machine
- keep GC-specific `Free`/`Held`/`Missing`/`Error` mapping and guarded lock-file removal while deleting its duplicate flock, inode, retry, and validation policy
- use the canonical 64-attempt bound, private-file normalization, and fail-closed descriptor identity errors across runner lock callers
- cover stale successful and contended descriptors, both opening modes' exhaustion, private-file validation, missing parents, and stale-guard removal

## Scope

- existing async lock API signatures are unchanged
- GC probe call sites, await points, guard lifetimes, lock ordering, and best-effort skip behavior are unchanged
- no API, persisted data, protocol, database, or deployment-order changes

## Validation

- `cargo test -p runner cmd::gc::lock_file::tests`
- `cargo test -p runner` (3224 passed, 20 ignored; guest inventory passed)
- repository pre-commit Rust format, Clippy, documentation, file-size, and commitlint hooks

Closes #28472

### DIFF
--- a/crates/runner/src/cmd/gc/lock_file.rs
+++ b/crates/runner/src/cmd/gc/lock_file.rs
@@ -1,11 +1,9 @@
-use std::os::unix::fs::{MetadataExt, OpenOptionsExt};
 use std::path::Path;
 
-use nix::fcntl::{Flock, FlockArg};
+use nix::fcntl::Flock;
 use tracing::{info, warn};
 
-use crate::host_file;
-use crate::lock;
+use crate::lock::{self, ExistingTryLock, TryLock};
 
 pub(super) enum LockProbe {
     /// Lock acquired — resource is not in use.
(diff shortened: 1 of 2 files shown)


seven332:
## Summary

Reviewed the single-commit diff against `main`, the approved issue decision, all affected lock callers, and the repository's Rust testing and quality guidance. The shared state machine preserves the create-versus-existing-only distinction, async API signatures, GC control flow, bounded retry behavior, and the final path-identity guard before unlinking.

## Findings

No P0, P1, or P2 findings.

## Testing

- `cargo test -p runner lock::tests` — 36 passed
- `cargo test -p runner cmd::gc::lock_file::tests` — 7 passed
- `cargo test -p runner` — 3224 passed, 20 ignored; guest inventory test passed
- `git diff --check`

## Verdict

LGTM for head `9aa8169e820c07021c82a53891fdfb70d288c33a`.

PR #28495, opened 2026-08-21
fix(runner): guard connector runtime publications

### LINKED ISSUE #28467: bug(rust): prevent stale connector runtime publications after generation changes
## Summary

`publish_connector_runtime_response` filters prepared updates against the current registration and target generation at `crates/runner/src/provider/connector_runtime_sync.rs:854-863`, then releases the state lock and awaits the proxy-registry transaction at `crates/runner/src/provider/connector_runtime_sync.rs:864-876`.

While that transaction is waiting, `notify_connector_runtime_sync_inner` can advance the same target to a newer generation and replace its schedule at `crates/runner/src/provider/connector_runtime_sync.rs:410-437`. The registry transaction in `crates/runner/src/proxy/registry.rs:634-647` checks only `source_ip` and `run_id`; it does not know the registration token or target generation. The post-publication check in `complete_successful_sync` (`crates/runner/src/provider/connector_runtime_sync.rs:1227-1248`) detects that the response is stale only after the old update has already been persisted.

## Failure scenario

1. Generation N receives a valid runtime-sync response and passes the current-generation check.
2. The task yields while acquiring or waiting for the proxy-registry lock.
3. A realtime notification advances the target to generation N+1 and schedules immediate work.
4. Generation N acquires the registry lock and publishes its now-stale network policy or firewall state because the registry entry still has the same source IP and run ID.
5. The post-write generation check returns false, but it cannot undo the stale publication. If generation N+1 is delayed by queue pressure or API failures, the stale policy remains active through the retry interval; even on success, it is exposed until the newer sync completes.

This can temporarily restore permissions that the newer notification was intended to revoke, so the generation guard must cover the publication boundary rather than only preparation and completion bookkeeping.

## Acceptance criteria

- A response whose registration or target generation is superseded before the registry commit must not mutate connector runtime registry state.
- The newer notification must remain scheduled and must be the only generation allowed to publish after it advances.
- Add a deterministic concurrency regression test that blocks the registry transaction, advances the generation, releases the transaction, and verifies that the stale response does not change the policy or firewall.
- Preserve the existing run/source ownership checks, batching behavior, and last-known-good retry semantics.

seven332:
## Summary

- split connector runtime registry publication into an explicit locked transaction
- revalidate registration and target generations after registry-lock acquisition and hold currentness through the atomic write
- omit targets superseded during lock contention while still publishing current siblings from the same batch
- add a deterministic real-registry/flock regression that preserves last-known-good state and the newer queued generation

## Compatibility and scope

- preserves existing source IP/run ownership checks, per-update outcomes, retries, idempotent writes, and completion scheduling
- does not change APIs, queue payloads, runner protocols, the persisted registry schema, or the Python addon
- does not add per-run publication fences; the existing state mutex is held only during post-contention registry I/O

## Validation

- `cargo fmt --check`
- `cargo test -p runner provider::connector_runtime_sync::tests` (49 passed)
- `cargo clippy -p runner --all-targets -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p runner --no-deps`
- `cargo test -p runner` (3221 passed, 20 existing ignored; integration test 1 passed)
- repository pre-commit Rust format, documentation, Clippy, file-size, and commit-message hooks

Closes #28467

### DIFF
--- a/crates/runner/src/provider/connector_runtime_sync.rs
+++ b/crates/runner/src/provider/connector_runtime_sync.rs
@@ -52,8 +52,8 @@ use super::api::{ApiClient, ConnectorRuntimeSyncOutcome};
 use crate::error::RunnerError;
 use crate::ids::RunId;
 use crate::proxy::{
-    ConnectorRuntimeFailCloseOutcome, ConnectorRuntimeRegistryTransaction,
-    ConnectorRuntimeRegistryUpdate, CustomConnectorRuntimeRegistryState, ProxyRegistryHandle,
+    ConnectorRuntimeFailCloseOutcome, ConnectorRuntimeRegistryUpdate,
+    CustomConnectorRuntimeRegistryState, ProxyRegistryHandle,
 };
 use crate::types::{
     ConnectorRuntimeSyncBatchResponse, ConnectorRuntimeSyncState, ConnectorRuntimeTarget,
@@ -863,11 +863,10 @@ impl ConnectorRuntimeSyncCore {
             return false;
         };
         if !prepared_publications.is_empty() {
-            let transaction: Result<ConnectorRuntimeRegistryTransaction<'_>, RunnerError> =
-                snapshot
-                    .registry
-                    .connector_runtime_registry_transaction()
-                    .await;
+            let transaction = snapshot
+                .registry
+                .connector_runtime_registry_transaction()
+                .await;
             let (prepared_publications, publication) = match transaction {
                 Ok(transaction) => {
                     // Acquire the registry transaction before active state so notifications can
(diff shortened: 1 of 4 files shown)


seven332:
## Code Review: PR #28495

**Reviewed head:** `2cff28af2a7f0f58c5d85dc42352920169bc9609`

### Summary

The registry transaction is acquired before the final registration/generation revalidation and remains held with active state through the atomic write. This closes the stale-publication race while preserving current siblings, ownership checks, retry behavior, and the persisted contract.

### Findings

No findings.

### Testing

Verified Rust formatting, the 49 connector-runtime sync tests including the deterministic real-flock regression, runner all-targets Clippy with warnings denied, rustdoc with warnings denied, and the full runner test suite on the implementation tree. The behavior-neutral encapsulation cleanup was revalidated with the focused suite and static checks. GitHub Actions checks are handled separately.

### Verdict

LGTM

PR #28501, opened 2026-08-21
fix(platform): lighten the model picker's variant control

tongx-blip:
## Problem

The variant segment was the heaviest element in the model picker, for four compounding reasons:

1. **Filled track** — `bg-muted` is gray-100 in light mode, while the rest of the list is white plus a 5% hover layer. It was the darkest block in the popover.
2. **Raised pill** — the checked variant is a white fill with a two-layer shadow, i.e. a floating element inside a flat list.
3. **Nearly full row height** — a 28px control in a 32px row leaves 2px above and below, so it visually *is* the row.
4. **Always on, and stacked** — every variant family renders one, so the video panel shows two in a six-row list and the eye lands on the controls instead of the model names.

before / after

## Change

Strip the control to text at the same position:

- root gets `h-5 bg-transparent p-0` — no track
- items get `font-normal ... data-checked:bg-transparent data-checked:font-medium data-checked:shadow-none` — state is carried by weight and colour instead of a pill
- the base `not-data-checked:hover:bg-state-hover` rule already in `segmentControlItemVariants` is what now reveals the hit area on hover
- the group falls from a 28px control to a 20px line box, so the row returns to the same `py-1.5` as every other row instead of compensating with `py-0.5`

Measured on the worst row (Seedance 2.0, three variants): painted control area falls from **42% of the row to 0%** at rest; the group narrows from 155px to 147px.

`text-[13px]` is one step up from the segment's usual `text-xs`. Without a track to lend it presence, 12px read as detached from the 14px row it belongs to. This is the one off-scale value in the diff and is easy to drop back to `text-xs` if reviewers prefer.

**Interaction is unchanged** — one click per variant, same `aria-label`, `aria-checked` and `aria-current` semantics.

`SegmentControl` itself is untouched. The overrides are applied at this call site only, exactly the way `MEDIA_MODEL_CATEGORY_SEGMENT_CLASS` already does it for the category switch, so every other segment control in the app keeps its default appearance.

## Design reference

Four directions were mocked at 1:1 against the real tokens and geometry; this PR implements the one that was picked (A). Comparison page, with the measured numbers per direction: https://model-variant-weight.sites.vm0.io

## Testing

- `chat-composer-model-selection-and-providers.test.tsx` already covers this control through the page — it asserts every variant radio's `aria-label`, text content and `aria-checked`, all of which this change preserves. No test needed updating.
- No new test: the change is Tailwind classes only, and `docs/testing/testing-external-behavior.md` excludes CSS-class assertions from the platform test boundary.
- Local checks could not run in this environment (no npm registry access, so dependencies are not installed); relying on the PR pipeline.

### DIFF
--- a/turbo/apps/platform/src/views/zero-page/components/model-provider-picker.tsx
+++ b/turbo/apps/platform/src/views/zero-page/components/model-provider-picker.tsx
@@ -1024,6 +1024,18 @@ export function VideoModelBrandIcon({ model }: { model: VideoModel }) {
   );
 }
 
+// The variant control sits inside a list whose only other paint is a 5% hover
+// layer, so a filled `bg-muted` track plus a raised white pill made it the
+// heaviest thing in the popover -- the eye landed on the control instead of the
+// model names. Stripped to text, the state is carried by weight and colour
+// alone, and the group drops from 28px to the 20px line box.
+const VARIANT_SEGMENT_CLASS = "h-5 bg-transparent p-0";
+
+// `text-[13px]` steps up from the segment's usual `text-xs`: without a track to
+// lend it presence, 12px read as detached from the 14px row it belongs to.
+const VARIANT_SEGMENT_ITEM_CLASS =
+  "h-5 rounded-md px-1.5 text-[13px] font-normal data-checked:bg-transparent data-checked:font-medium data-checked:shadow-none";
+
 function MediaModelPanelRow({ option }: { option: MediaModelPanelOption }) {
   if (option.variant) {
     const variant = option.variant;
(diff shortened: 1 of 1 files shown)


tongx-blip:
Closing as obsolete.

This PR lightened the media pickers' variant segment, but #28510 (*offer one model per family in the media pickers*) has since removed that control from `main` entirely — a family now contributes a single row, so `MediaModelPanelVariantControl` and the `option.variant` branch this PR edits no longer exist. #28542 then narrowed the popover back to `min-w-[260px]`, which was the width the segment had been forcing.

The same visual complaint that motivated this PR is therefore already resolved upstream, by a more direct route. Nothing here is worth rebasing.

PR #28510, opened 2026-08-21
feat(composer): offer one model per family in the media pickers

bingjiez666:
## What

Collapses each model family in the composer's image and video pickers down to a single row, removing the variant segment controls that were shown on the Flux Pro v1.1, Seedream 5, and Seedance 2.0 rows.

**Image picker** — GPT Image 1, GPT Image 2, Nano Banana 2, Flux Pro v1.1, Seedream 5 Pro.
Removed from the list: the Standard/Ultra and Pro/Lite segments, plus Qwen Image.

**Video picker** — Seedance 2.5, Seedance 2.0, Seedance 1.5 pro, Veo 3.1 fast, Kling v3 4K, MiniMax H3.
Removed from the list: the Standard/Fast/Mini segment, so Seedance 2.0 is the Standard model only.

## Scope

This changes what the picker *offers*, not what the platform can generate. Flux Ultra, Seedream 5 Lite, Qwen Image, Seedance 2.0 fast, and Seedance 2.0 Mini all stay reachable through their CLI aliases and through defaults that already name them. Their `public` flag / catalog entry is what changed.

One consequence worth a decision: `DEFAULT_VIDEO_MODEL` is still `dreamina-seedance-2.0-fast`, which the picker no longer offers. A thread with no pin therefore shows no checked row in the video panel. Flipping the default to Seedance 2.0 Standard would fix that, but it changes what every default video generation runs (CLI and API included), so it is deliberately left out of this PR.

`videoModelList()` filtered by `public` to build the "Available models" hint on an unsupported-model error. With `public` now meaning picker visibility only, that hint would have understated what the endpoint accepts, so it lists every catalog alias — matching what the image service already does.

## Checks

- `pnpm format`, `pnpm turbo run lint`, `pnpm check-types`, `pnpm knip` — all pass
- `chat-composer-model-selection-and-providers.test.tsx` (81), `chat-composer-template-picker.test.tsx` + `chat-thread-emoji-category-rail.test.tsx` (44) — all pass
- API tests need a local Postgres that this environment does not have; relying on the PR pipeline for those

### DIFF
--- a/turbo/apps/api/src/signals/services/video-generation.service.ts
+++ b/turbo/apps/api/src/signals/services/video-generation.service.ts
@@ -661,14 +661,12 @@ function normalizeVideoModel(value: string): VideoModel | null {
   return null;
 }
 
+// Every catalog model is accepted here, so the hint names them all. `public`
+// only decides which of them the composer picker offers.
 function videoModelList(): string {
-  return VIDEO_MODELS.filter((model) => {
-    return VIDEO_MODEL_CONFIGS[model].public;
-  })
-    .map((model) => {
-      return VIDEO_MODEL_CONFIGS[model].alias;
-    })
-    .join(", ");
+  return VIDEO_MODELS.map((model) => {
+    return VIDEO_MODEL_CONFIGS[model].alias;
+  }).join(", ");
 }
 
 export function videoProviderForModel(model: VideoModel): VideoProvider {
--- a/turbo/apps/platform/src/i18n/locales/de-DE/common.json
+++ b/turbo/apps/platform/src/i18n/locales/de-DE/common.json
@@ -4607,7 +4607,6 @@
         "loadError": "Modelle können nicht geladen werden.",
         "loading": "Modelle werden geladen...",
         "models": "Modelle",
-        "modelVariants": "Varianten von {{model}}",
         "noConfiguredModels": "Keine konfigurierten Modelle",
         "priceTiers": {
           "balanced": "Ausgewogene Kosten und Leistung",
--- a/turbo/apps/platform/src/i18n/locales/en-US/common.json
+++ b/turbo/apps/platform/src/i18n/locales/en-US/common.json
@@ -4607,7 +4607,6 @@
         "loadError": "Unable to load models.",
         "loading": "Loading models...",
         "models": "Models",
-        "modelVariants": "{{model}} variants",
         "noConfiguredModels": "No configured models",
         "priceTiers": {
           "balanced": "Balanced cost and performance",
--- a/turbo/apps/platform/src/i18n/locales/es-ES/common.json
+++ b/turbo/apps/platform/src/i18n/locales/es-ES/common.json
@@ -4657,7 +4657,6 @@
         "loadError": "No se pudieron cargar los modelos.",
         "loading": "Cargando modelos...",
         "models": "Modelos",
-        "modelVariants": "Variantes de {{model}}",
         "noConfiguredModels": "Sin modelos configurados",
         "priceTiers": {
           "balanced": "Coste y rendimiento equilibrados",
--- a/turbo/apps/platform/src/i18n/locales/fr-FR/common.json
+++ b/turbo/apps/platform/src/i18n/locales/fr-FR/common.json
@@ -4657,7 +4657,6 @@
         "loadError": "Impossible de charger les modèles.",
         "loading": "Chargement des modèles...",
         "models": "Modèles",
-        "modelVariants": "Variantes de {{model}}",
         "noConfiguredModels": "Aucun modèle configuré",
         "priceTiers": {
           "balanced": "Coût et performance équilibrés",
--- a/turbo/apps/platform/src/i18n/locales/hi-IN/common.json
+++ b/turbo/apps/platform/src/i18n/locales/hi-IN/common.json
@@ -4607,7 +4607,6 @@
         "loadError": "मॉडल लोड करने में असमर्थ.",
         "loading": "मॉडल लोड हो रहे हैं...",
         "models": "मॉडल",
-        "modelVariants": "{{model}} के वेरिएंट",
         "noConfiguredModels": "कोई कॉन्फ़िगर मॉडल नहीं",
         "priceTiers": {
           "balanced": "संतुलित लागत और प्रदर्शन",
--- a/turbo/apps/platform/src/i18n/locales/id-ID/common.json
+++ b/turbo/apps/platform/src/i18n/locales/id-ID/common.json
@@ -4606,7 +4606,6 @@
         "loadError": "Tidak dapat memuat model.",
         "loading": "Memuat model...",
         "models": "Model",
-        "modelVariants": "Varian {{model}}",
         "noConfiguredModels": "Tidak ada model yang dikonfigurasi",
         "priceTiers": {
           "balanced": "Biaya dan performa seimbang",
--- a/turbo/apps/platform/src/i18n/locales/it-IT/common.json
+++ b/turbo/apps/platform/src/i18n/locales/it-IT/common.json
@@ -4657,7 +4657,6 @@
         "loadError": "Impossibile caricare i modelli.",
         "loading": "Caricamento modelli...",
         "models": "Modelli",
-        "modelVariants": "Varianti di {{model}}",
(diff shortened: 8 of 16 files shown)


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 27661c956dde62b50136aec19e987b9b67e84995
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Collapses each media-model family in the composer picker to a single row by deleting the variant segment control and trimming `PUBLIC_IMAGE_MODELS` / the video `public` flags, so the image picker offers GPT Image 1/2, Nano Banana 2, Flux Pro v1.1, and Seedream 5 Pro, and the video picker offers Seedance 2.0 Standard rather than Standard/Fast/Mini. No catalog entry, alias, or contract model ID is removed, so every hidden model stays generatable; the API's unsupported-model hint is widened to all aliases because `public` now means picker visibility only.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no unhandled null/undefined paths | Finding (P1) | `video-model-catalog.ts` — `DEFAULT_VIDEO_MODEL` is `dreamina-seedance-2-0-fast-260128`, now `public: false`, so the resolved default matches no offered row. See P1 below. |
| `REVIEW.md` | Security (injection, secrets, auth on new endpoints) | N/A | Diff adds no endpoint, credential, or user-supplied string path; the only API change is a static error-hint string built from compile-time catalog aliases. |
| `REVIEW.md` | Type safety; no `any`, no broad casts | Pass | `image-model-catalog.ts`, `video-model-catalog.ts`, `zero-chat-composer.tsx` keep `as const satisfies` catalog typing; the only cast removed was in the deleted variant branch. |
| `REVIEW.md` | Style and maintainability; no dead or duplicated logic | Pass | Net −339 lines. `composerImageModelPanelCategory` / `composerVideoModelPanelCategory` lose their group-lookup branches; `IMAGE_MODEL_VARIANT_GROUPS`, `MediaModelPanelVariantControl`, `variantLabel`, and the `modelVariants` string in 10 locales are all deleted rather than left unused. |
| `REVIEW.md` | `feat:` commits must carry integration tests | Pass | `chat-composer-model-selection-and-providers.test.tsx` — "selects the one model each family offers" and "pins a video model on the thread from the model picker" exercise the new single-row selection through `detachedSetupPage`; 81 tests pass. |
| `REVIEW.md` | Feature regression guard; changed behavior needs explicit coverage | Pass | The replaced flow (picking a family variant) is re-covered at the row level, and the stale-pin path keeps its own test ("shows no selection when the pinned model is no longer offered"). The picker is already contained by `ImageModelSelection` / `VideoModelSelection`. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No `packages/db/src/schema/**`, contract, or `jsonb().$type<>()` change; persisted `selectedImageModel` / `selectedVideoModel` values stay valid because no catalog ID or contract union was removed. |
| `docs/bad-smell.md` | Zero `any` types | Pass | No `any` introduced in any changed file. |
| `docs/bad-smell.md` | Zero lint/type suppression comments | Pass | No `eslint-disable`, `oxlint-disable`, `@ts-ignore`, `@ts-expect-error`, or `prettier-ignore` added. |
| `docs/bad-smell.md` | Static imports only | Pass | Diff only removes an import (`IMAGE_MODEL_VARIANT_GROUPS`); no dynamic `import()` added. |
| `docs/bad-smell.md` | No hardcoded URLs or environment values | Pass | Only model IDs and aliases, which are the catalog's own identifiers. |
| `docs/fallback.md` §1 | No negative tests asserting removed behavior stays removed | Pass | `27661c9` removed the one tombstone (the "no non-tab radios remain" assertion, which TypeScript already proved via the deleted `variant` field). The surviving `queryByText("Qwen Image")).toBeNull()` style assertions pin `PUBLIC_IMAGE_MODELS` data, which can genuinely regress. |
| `docs/fallback.md` §2 | A non-GA feature switch needs no compatibility code | Pass | `feature-switch.ts` — both `VideoModelSelection` and `ImageModelSelection` are `enabled: false` with `enabledOrgIdHashes: STAFF_ORG_ID_HASHES`, so no dual-read, migration, or old-client branch is owed for this picker. |
| `docs/fallback.md` §9 | Every fallback the diff introduces is declared in the PR summary | N/A | The diff introduces no fallback. It keeps existing behavior (hidden models stay generatable) and adds no compatibility branch, so no `Fallbacks` section is required. |
| `docs/deployment-compatibility.md` | Compatibility across deployable boundaries and persisted state | Pass | Catalog constants are frontend internals inside one browser build ("Compatibility is not required inside one deployed artifact"). `api-contracts` is untouched, so an old client can still pick Ultra/Lite/Fast/Mini and the current API still accepts them. |
| `docs/externally-managed-references.md` | A resolution miss is `unavailable`, never a silent substitution | Pass | `MediaModelPanelRow` renders no `aria-current` when the pinned model is not offered instead of highlighting a sibling — the diff deliberately does not remap Fast onto the Standard row. |
| `docs/effect.md` | Keep render pure; choose the right execution boundary | Pass | `MediaModelPanelRow` becomes a pure prop-to-JSX function; no effect, command, or route setup added. |
| `docs/cache.md` | RC-1/RC-2: no render-phase side effects or render-time signal allocation | Pass | Only deletions in the render path; no `useState`, `useEffect`, `useRef`, or signal allocated during render. |
| `.claude/skills/ccstate/SKILL.md` | Signals owned outside render; commands carry business semantics | Pass | No new signal or command; `ComposerModelPickerControls` keeps reading `signals.model.*` through the existing `useGet` / `useSet` boundary. |
| `docs/testing.md` | Integration tests at the entry point; mock only at the boundary | Pass | All changed tests drive `detachedSetupPage` + `userEvent`; no new `vi.mock`, no relative-path mock, no `globalThis.services.db` mock, no fake timers. |
| `docs/testing/app-testing.md` | Platform tests go through the production initialization flow | Pass | Every touched case still starts from `detachedSetupPage({ context, featureSwitches, path })` and asserts rendered DOM. |
| `docs/testing/testing-external-behavior.md` | Assert through the user-visible interface | Pass | Assertions moved from segment-radio internals to row labels, `aria-pressed`, and the recorded `chatThreadImageModelContract.update` / `chatThreadVideoModelContract.update` bodies. |
| `docs/testing/anti-patterns.md` | AP-7 no implementation-detail assertions; AP-8 no over-testing | Pass | The CSS-class-filtered radio assertion was the one AP-7 smell and is gone in `27661c9`; the remaining brand-icon and label assertions cover distinct picker rows. |
| `docs/testing/patterns.md` | Helpers reflect the behavior under test | Pass | `selectedImageModelLabel` / `selectedVideoModelLabel` drop their variant-segment fallbacks, and the now-unused `findImageVariantSegment` / `findVideoVariantSegment` helpers are deleted with the feature. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

- `turbo/packages/core/src/video-model-catalog.ts` — `DEFAULT_VIDEO_MODEL` is still `dreamina-seedance-2-0-fast-260128`, which this PR sets to `public: false`. A thread with no pin resolves to that model (`zero-chat-composer.tsx`: `videoModel.value ?? userPreference?.selectedVideoModel ?? DEFAULT_VIDEO_MODEL`), so the video panel renders six rows with none marked, and the test at "pins a video model on the thread from the model picker" now encodes that with `expect(selectedVideoModelLabel()).toBeUndefined()`. The composer prints the video model nowhere else, so a staff user cannot see which model will run and cannot deliberately return to it once they pick another. Blast radius is limited to `STAFF_ORG_ID_HASHES` because `VideoModelSelection` is non-GA, and the PR summary declares the gap and the reason it was deferred, so this does not block the verdict — but the default and the offered set should agree before this picker reaches GA. The fix is to make Seedance 2.0 Standard the default, which changes what every unspecified video generation runs (CLI and API included) and therefore needs a product call.

### Testing

- Coverage: Adequate. The `feat:` change is covered by updated integration tests for single-row selection on both categories, the stale-pin render path, the mobile category strip, and per-composer pin independence.
- Conventions: Compliant. Integration-level, `detachedSetupPage`-driven, boundary-only mocking, no fake timers, no mock-call-only assertions. The one implementation-detail assertion flagged during review was removed in `27661c9`.

PR #28512, opened 2026-08-21
fix(guest-agent): mask sandbox operation telemetry

### LINKED ISSUE #28473: security(rust): mask sandbox operation telemetry before upload
## Problem and evidence

The Rust guest-agent masks `system_log.content` before building the telemetry request, but moves parsed `sandbox_ops.entries` into the same request unchanged.

Sandbox-operation entries may contain free-form diagnostic strings. In particular, `record_sandbox_op` serializes its optional `error` argument verbatim, and checkpoint, artifact, working-directory, CLI-execution, and cleanup failure paths pass runtime diagnostics into that field. The API then includes `sandboxOperations[].error` in the sandbox-operation event written to Axiom. A configured secret present in one of those diagnostics can therefore reach centralized telemetry even when the duplicate system-log text is redacted.

## Decision

Treat the guest-agent telemetry upload boundary as the required redaction boundary for sandbox-operation data. The issue is necessary because the current behavior violates the existing guarantee that configured secrets are masked before leaving the guest.

Redaction must recursively cover string values in each sandbox-operation entry. It must not rename object keys, change array/object shape, remove fields, or alter boolean and numeric values. This distinction matters because the existing generic JSON masker also masks object keys; a configured secret that happens to match a schema key must not make the telemetry payload malformed.

## Scope and constraints

- Preserve the existing unredacted local sandbox-operation log, matching the current system-log model; redact the in-memory upload copy.
- Preserve the telemetry API contract and ingestion shape.
- Preserve delta-position advancement, retry after upload failure, and best-effort upload behavior.
- Do not add API-side masking: the API does not have the guest's configured secret set.
- Keep metrics and run identity unchanged; the exposed free-form data in scope is sandbox-operation telemetry.

## Risks and mitigations

- **Payload corruption:** value-only recursive masking preserves schema keys and non-string values.
- **Behavioral drift in retries or positions:** apply masking only after the delta and position calculations and before request construction.
- **Future string fields:** recursive value masking covers new nested string values without maintaining a field allowlist.
- **Performance:** telemetry reads are already bounded, and the existing compiled secret matcher makes the additional scan linear in the bounded payload size.

## Acceptance criteria

- Every string value within `sandboxOperations` is passed through `SecretMasker` before the telemetry request is serialized.
- Object keys, JSON structure, and non-string values remain unchanged.
- A configured secret in `sandboxOperations[].error` never appears in the serialized telemetry request and is replaced with the standard `***` marker.
- Integration coverage exercises the real telemetry upload entry point and verifies both redaction and structure preservation.
- Existing system-log masking, delta-position, retry, and best-effort upload behavior remains unchanged.

seven332:
## Summary

- recursively mask string values in sandbox-operation telemetry before upload
- preserve fixed JSON keys, structure, numbers, booleans, delta positions, and retry behavior
- extend the telemetry HTTP integration test to cover error redaction and a secret/key collision

## Scope

- no API contract or persisted JSONL format changes
- no database migration or feature switch
- local telemetry files remain unchanged; redaction applies to the outbound in-memory copy

## Validation

- `cargo test -p guest-agent --test integration telemetry::telemetry_preserves_runtime_session_id_and_masks_secrets -- --exact`
- `cargo test -p guest-agent --all-targets --all-features`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p guest-agent --no-deps --all-features`
- pre-commit workspace Rust formatting, documentation, Clippy, and file-size checks

Closes #28473

### DIFF
--- a/crates/guest-agent/src/masker.rs
+++ b/crates/guest-agent/src/masker.rs
@@ -152,6 +152,29 @@ impl SecretMasker {
         }
     }
 
+    /// Recursively mask secrets in JSON string values without changing keys.
+    pub(crate) fn mask_string_values(&self, val: &mut Value) {
+        if self.matcher.is_none() && self.url_encoded_matcher.is_none() {
+            return;
+        }
+        match val {
+            Value::String(s) => {
+                self.mask_string_in_place(s);
+            }
+            Value::Array(arr) => {
+                for item in arr {
+                    self.mask_string_values(item);
+                }
+            }
+            Value::Object(map) => {
+                for value in map.values_mut() {
+                    self.mask_string_values(value);
+                }
+            }
+            _ => {}
+        }
+    }
+
     /// Replace all secret patterns in a string with `***`.
     ///
     /// Redacts the full union of all pattern matches so overlapping configured
--- a/crates/guest-agent/src/telemetry.rs
+++ b/crates/guest-agent/src/telemetry.rs
@@ -199,7 +199,10 @@ async fn upload_telemetry(
         masker.mask_owned_string(system_log.content)
     };
     let metrics_entries = metrics.entries;
-    let sandbox_ops_entries = sandbox_ops.entries;
+    let mut sandbox_ops_entries = sandbox_ops.entries;
+    for entry in &mut sandbox_ops_entries {
+        masker.mask_string_values(entry);
+    }
 
     let payload = json!({
         "runId": run_id,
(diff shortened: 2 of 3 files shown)


seven332:
## Code Review: PR #28512

**Reviewed head:** `15c5468837ab5e97573950fdf2dd18d28fc16832`

### Summary

The PR closes the sandbox-operation telemetry secret leak at the guest upload boundary. Its value-only recursive traversal preserves fixed JSON keys and non-string fields while leaving the existing event key-masking path, telemetry cursor state machine, retry behavior, and API contract unchanged.

### Findings

No findings.

### Testing

The regression runs through the real telemetry actor and HTTP serialization boundary, covering secret absence, masked operation errors, a configured-secret/key collision, and numeric/boolean preservation. Focused and full guest-agent tests, formatting, Clippy, rustdoc, and pre-commit workspace Rust checks passed locally. GitHub CI status is assessed separately.

### Verdict

LGTM

Issue #26578, opened 2026-08-12
refactor(db): remove presentation template aspect ratio


bingjiez666:
Part of #21350. Depends on #26576.

Presentation-template v1 is fixed at 16:9. Remove the now-redundant aspect-ratio protocol and persisted field after the browser-rendered ingestion handoff lands.

## Scope

- Remove `presentation_templates.aspect_ratio` through the normal generated Drizzle migration path.
- Remove `aspectRatio` from presentation-template API request and response contracts.
- Remove aspect-ratio detection, validation, storage, mocks, and tests from API, CLI, and Platform consumers.
- Reserve template and preview layout with the fixed 16:9 product constant instead of persisted data.

## Acceptance

- Existing template rows migrate without data transformation because the value is no longer read.
- No client submits or branches on a presentation-template aspect ratio.
- Catalog, detail, browser upload, and generation behavior remain fixed at 16:9.

PR #28515, opened 2026-08-21
refactor(db): drop the presentation template status and error columns

bingjiez666:
## What

`presentation_templates.status` and `presentation_templates.error` are dropped, together with the two-phase delete that borrowed `status` as a claim marker.

`publish` is the only insert path and it runs after the analysis has finished, so it always wrote `status: "ready"`. `pending` and `processing` had no writer — they described the earlier "insert on upload, process in the background" design, which the chat-thread flow replaced: the thread is where the user watches the analysis, so the table has no import lifecycle left to record. `error` had no writer either.

## Why the delete path changes with it

The old delete flagged the row `failed` in a transaction, then deleted `WHERE status = 'failed'` outside it. The advisory lock it took was `pg_advisory_xact_lock`, released when that transaction committed — before the `DELETE` ran — so the marker column was what actually made concurrent deletes exclusive.

A single `DELETE ... RETURNING` holds the row lock for the statement that removes the row, so exactly one concurrent caller observes it and the other returns nothing. The marker, the extra transaction, and `presentation-template-lifecycle.service.ts` are all removed. A new integration test races two deletes of the same template and asserts one `204` and one `404`.

## Why one release and not a two-release contract

`docs/deployment-compatibility.md` requires a code-first release before a physical `DROP COLUMN`, because a draining old API can still name the column in a generated `SELECT`/`RETURNING`. That obligation does not apply here, per `docs/fallback.md` §2 and §7:

- `PresentationTemplates` is `enabled: false` with no staff-org hashes, so the feature is non-GA.
- Every one of the five handlers evaluates the switch **before** it issues any query and returns `403` when it is off. An old API instance running against the migrated database therefore never generates a statement naming `status` unless the caller holds a per-user override.
- The only account holding that override is the feature's maintainer, so the worst case in the skew window is that account's own request returning `500`.

Nothing else reads the table: the API routes and services in this diff are the only consumers, platform has none, and there is no cron or cascade.

Response-shape removal is likewise safe for an in-flight run's older CLI, because this domain's client does not opt into `validateResponse` — a summary missing two fields is not parsed against the old schema.

## Not in scope

- `aspect_ratio` stays; it is tracked by #26578.
- Existing rows are untouched; dropping columns is a schema change, and clearing the maintainer's test rows is a separate decision.

## Verification

`pnpm install` and the checks below ran in the sandbox: `format` (no changes), `knip` (exit 0), `lint` for `@okouai/db` (including the jsonb-contract lint), `@okouai/api-contracts`, `api`, and `@okouai/cli`, and `check-types` for `@okouai/db`, `@okouai/api-contracts`, and `@okouai/cli`. `api` `check-types` passed `gateways`, `core`, `bootstrap`, and `tests`; the final `bootstrap-wiring` project was stopped after exhausting the sandbox's memory, so CI is its first run. Per the repository instruction not to run local vitest, the test suites are first executed by the PR pipeline.


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 849351a4e90fcd3ba268eff6a3fc7d6cccfd05d8
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Drops `presentation_templates.status` and `presentation_templates.error` (migration `0959`, plus the `chk_presentation_templates_status` constraint), removes the enum, the JSONB contract, and both fields from the API contract, and collapses the two-phase delete that used `status = 'failed'` as a claim marker into a single `DELETE ... RETURNING`. `publish` is the only insert path and always wrote `ready`, so `pending`/`processing`/`error` had no writer; the analysis run's chat thread is the progress surface.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic, including race conditions | Pass | `presentation-template-delete.service.ts`: the removed advisory lock was `pg_advisory_xact_lock`, released at transaction commit — before the old `DELETE` ran outside that transaction — so the marker column, not the lock, provided exclusion. A single `DELETE ... RETURNING` holds the row lock for the removing statement, so under READ COMMITTED exactly one concurrent caller observes the row whether the loser blocks or arrives after commit. |
| `REVIEW.md` | Security: auth checks, no-existence-leak | Pass | `presentation-templates.ts`: all five handlers still gate on `presentationTemplatesEnabled$` and their `requiredCapability`; the new delete keeps `orgId` + `ownerUserId` in the `WHERE`, so a foreign template still yields 404, not 204. |
| `REVIEW.md` | Type safety: no `any`, no broad casts | Pass | No `any`, no `as` added; the removed `$type<PresentationTemplateStatus>()`/`$type<PresentationTemplateError>()` were the only casts touched. |
| `REVIEW.md` | Style and maintainability; interface changes documented | Pass | Net −142/+81. The removed `status`/`error` response fields are a public contract change and are described in the PR body, including why an older CLI is unaffected. |
| `REVIEW.md` | Testing coverage for a `refactor:` commit | Pass | Existing publish/list/get coverage still exercises the changed summary shape; the delete path, whose concurrency mechanism changed, gains `deletes a template exactly once when two requests race`. |
| `REVIEW.md` | Feature regression guard for replaced behavior | Pass | The two behaviors this PR replaces — delete exclusivity and the `ne(status,'pending')` list filter — are both covered: the race test for delete, and the existing publish→list assertion for the filter. The feature stays contained behind `FeatureSwitchKey.PresentationTemplates`. |
| `REVIEW.md` | DB/JSONB gate: old rows that could violate the new shape | Pass | `error` is `jsonb` with no writer anywhere in the tree, and dropping a column cannot leave a violating row. The PR states rows are untouched; per `docs/fallback.md` §2 a non-GA switch needs no backfill for staff-produced rows. |
| `docs/fallback.md` | §2 — a feature behind a non-GA switch needs no compatibility code, and a transient cutover 500 is acceptable | Pass | `FeatureSwitchKey.PresentationTemplates` is `enabled: false` with no `enabledOrgIdHashes`, so §2 applies and no dual-read/backfill was added. |
| `docs/fallback.md` | §1 — no negative tests asserting removed behavior stays removed | Pass | The diff deletes assertions on `status`; it adds none asserting the field stays gone. |
| `docs/fallback.md` | Removing a fallback must prove its evidence and delete branch, contract entry, and tests together | Pass | Single-writer evidence is in the diff (`presentation-template-publish.service.ts` was the only insert and hardcoded `ready`); the schema column, contract field, read branches, and the tests that asserted them are removed in one commit. |
| `docs/fallback.md` | §7 — skew windows apply only to GA behavior | Pass | No new cross-version branch is introduced, and no window comment is owed for a non-GA switch. |
| `docs/deployment-compatibility.md` | Two-release contract for a physical `DROP COLUMN` (old API can still name the column in generated SQL) | N/A | The generic rule assumes a draining old API that can still issue the statement. Every handler here evaluates the feature switch and returns 403 **before** any query, so an old API against the migrated DB emits no statement naming `status` unless the caller holds a per-user override; `docs/fallback.md` §2/§7 exempt that case. |
| `docs/deployment-compatibility.md` | Audit persisted catalog objects (indexes, triggers, functions, defaults) that reference the column | Pass | The only such object was the partial unique index `idx_presentation_templates_active_import` (`WHERE status IN ('pending','processing')`) from `0896`, already dropped by `0942`. Migration grep shows no trigger, function, or FK referencing this table. |
| `docs/deployment-compatibility.md` | Check both skew directions (old code after migration, new code before migration) | Pass | New code before migration is safe because the new code never names the dropped columns; old code after migration is the 403-gated case above. |
| `docs/bad-smell.md` | Zero `any` and zero lint/type suppressions | Pass | No `any`, `@ts-ignore`, `eslint-disable`, or `prettier-ignore` in the diff. |
| `docs/bad-smell.md` | Fail fast; delete unused code aggressively rather than keeping dead paths | Pass | `presentation-template-lifecycle.service.ts` and `jsonb-contracts/presentation-template.ts` are deleted rather than left orphaned; `presentationTemplateStatusSchema` is removed from the contracts barrel. |
| `docs/bad-smell.md` | No unnecessary error handling / no hardcoded config / no dynamic imports | Pass | No try/catch, URL, or `import()` added; the delete path drops one transaction wrapper instead of adding one. |
| `docs/testing.md` | Integration tests only, at the entry point | Pass | The new test enters through the ts-rest contract client, not through the service. |
| `docs/testing.md` | Real database, real infrastructure; mock only external boundaries | Pass | The race test uses the real DB and the existing S3 fixture; no `vi.mock()` is added. |
| `docs/testing.md` | No fake timers | Pass | No `vi.useFakeTimers()`/`advanceTimersByTime` in the diff. |
| `docs/testing/api-testing.md` | Route tests go through `setupApp()` + contract, with `accept()` narrowing | Pass | `templateClient().delete(...)` wrapped in `accept(..., [204, 404])`, matching the file's existing harness. |
| `docs/testing/api-testing.md` | Shared persistent state must be uniquely owned, not isolated by ordering or teardown | Pass | The test creates its own `bdd.user()` and asserts on that owner's list, so a concurrent worker's rows cannot affect `toHaveLength(0)`. |
| `docs/testing/testing-external-behavior.md` | Construct and verify only through the external endpoint; no DB reads/writes or service calls in API tests | Pass | Setup is `publish`, verification is the two `delete` responses plus `list`; no DB schema or service import. |
| `docs/testing/patterns.md` | Standard API route test structure (module-scope `testContext()`, contract client helper, auth headers helper) | Pass | The new case reuses the file's `templateClient()`, `webHeaders()`, `uploadInputs()`, and `enablePresentationTemplates()` helpers. |
| `docs/testing/anti-patterns.md` | AP-1 test behavior, not mock calls | Pass | Assertions are on HTTP statuses and the listed body, not `toHaveBeenCalled()`. |
| `docs/testing/anti-patterns.md` | AP-4 never mock internal code; AP-11 never test the service when a route exists | Pass | The delete-service change is verified through `DELETE /api/okou/presentation-templates/:templateId`. |
| `docs/testing/anti-patterns.md` | AP-8 no over-testing of status codes or schema validation | Pass | One case covers a real concurrency guarantee the PR changes; no per-status-code enumeration was added. |
| `docs/testing/cli-testing.md` | CLI tests mock the Web API with MSW at the boundary and enter via `parseAsync` | Pass | `turbo/apps/cli/.../index.test.ts` only drops `status`/`error` from the MSW response body; the entry path is unchanged. |
| `.claude/skills/ccstate/SKILL.md` | `get`/`set` stay inside the command callback; no accessor escape | Pass | `deletePresentationTemplate$` uses `set(writeDb$)` inline within the callback and passes no accessor outward. |
| `.claude/skills/ccstate/SKILL.md` | `signal.throwIfAborted()` after an await that does not accept the signal | Pass | The Drizzle `DELETE` takes no `AbortSignal`, and `signal.throwIfAborted()` follows it. |
| `.claude/skills/ccstate/SKILL.md` | `detach()` must not appear in the signals layer | Pass | No `detach()` added; the command awaits directly. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

Non-blocking observation, no change requested: during the window between the migration and the API promotion, a request from an account that holds the `presentationTemplates` override can hit an old API instance that still names the dropped columns and receive a 500. `docs/fallback.md` §2 names exactly this ("a temporary 500 or a broken screen during the cutover is acceptable") for a non-GA switch, and the PR body already states it.

### Testing

- Coverage: Adequate. `refactor:` requires that existing tests still cover the changed code, and they do; beyond that the PR adds `deletes a template exactly once when two requests race`, which pins the guarantee that replaced the marker column. The race is deterministic rather than flaky: whether the loser blocks on the row lock or arrives after the winner commits, READ COMMITTED re-evaluation leaves it with zero rows, so `[204, 404]` holds in both interleavings.
- Conventions: Compliant. Integration-only, entry through the ts-rest contract, real database, per-test owned user, no internal mocks, no fake timers, `toStrictEqual` for the status pair.
- Gap accepted: no test asserts the physical `DROP COLUMN` itself; that is validated by CI running the migration against a real database, which is the project's normal coverage for generated migrations.

PR #28486, opened 2026-08-21
refactor(api): move the billing routes off the brand namespace

lancy:
Part of #28278, closes #28457.

Thirty-four product routes drop the `/api/okou` segment: billing status, plan and usage-pack checkout, concurrency subscriptions, credit purchase, the Stripe portal, invoices, downgrade and restore, and code redemption. Method, schemas, and handlers are unchanged.

## The three edits that had to land together

1. **Contract paths** — `packages/api-contracts/src/contracts/billing.ts` declares the neutral form for all 34 paths (35 route declarations; `auto-recharge` has GET and PUT).
2. **`MIGRATED_BRANDED_PATHS`** — one row per route, each naming both branded forms. `apiNamespaceAliasPaths()` returns a neutral path unchanged, so moving the contract removes both branded registrations; this table is the only thing that gives them back. Keys hold their path parameters verbatim, because the lookup matches `entry.route.path` exactly.
3. **Hardcoded call sites** — the CLI billing MSW handlers, the API billing-status probes inside the image/video/voice/avatar generation tests, the `chat-threads` bench label, and the Playwright billing interceptions in `billing-payment.spec.ts` and `chat.spec.ts`.

## Compatibility

`LEGACY_ZERO_PATHS` still lists `/api/okou/billing/concurrency-checkout/preview` from measured `vm0-request-log-prod` traffic. That row is untouched; the new migrated rows are what keeps serving both of its forms now that no contract declares the branded path. Removal of any row stays gated on #26701's evidence rules.

## Two adjacent guards this slice had to move

- `ccstate/no-non-zero-api` rejects a neutral `/api/**` literal in the platform app until its contract has actually moved. `/api/billing` joins the allow list — one prefix rather than 34 entries, since every moved path sits under it and nothing else does. A longer sibling like `/api/billing-exports/status` is still rejected, and there is a test for that.
- `api-namespace-declarations.test.ts` named `POST /api/okou/billing/concurrency-checkout/preview` as its proof that the barrel walk reaches a real declaration. The sample moves to `POST /api/okou/slack/events`, a path a provider console holds and #28278 does not move, so it cannot rot under a later slice.

## Testing

Expectations are restated literally in the test files and never derived from `apiNamespaceAliasPaths()`. `MIGRATED_ROUTE_PATHS` in `migrated-branded-paths.test.ts` gains the same 34 rows, so a dropped contract path or a dropped table row fails the test rather than changing what it asserts.

[...]

PR #28527, opened 2026-08-21
test(api-contracts): re-anchor the branded route sentinel off the migrating set

### LINKED ISSUE #28525: fix(api-contracts): re-anchor the branded route sentinel off the migrating set
Blocks #28457 and, as each remaining slice merges, every other open slice of #28278.

## What fails

```
api-namespace-declarations.test.ts > branded API namespace declarations
  > enumerates branded contract routes through the package barrel

AssertionError: expected 100 to be greater than 100
  63|  expect(brandedRoutes.length).toBeGreaterThan(100);
```

## Why it is not a real regression

This file was added by #28350. Its second test is the guard that matters — no contract may declare a `/api/zero/` path — and the first test exists to keep that guard honest.

Read together, the first test's purpose is a **sentinel against a false pass**. `declaredRoutes()` walks the package barrel, and if that walk ever returned nothing the `/api/zero/` assertion would pass vacuously against an empty array. The two assertions guard that:

```ts
expect(brandedRoutes.length).toBeGreaterThan(100);          // the walk found routes
expect(brandedRoutes.some(... "/api/okou/billing/concurrency-checkout/preview"))
  .toBe(true);                                              // and found a known one
```

The threshold was a safe lower bound at the time. But #28278 is deliberately driving the branded route count to zero, so a sentinel anchored on **branded** routes must fail — first at 100, then at 99, and so on as each slice merges. The specific anchor is worse: `/api/okou/billing/concurrency-checkout/preview` is one of the paths #28457 moves, so `some(...)` becomes false in the same PR.

Neither assertion is wrong about what it wants. Both are measuring it against something that is being removed on purpose.

## Fix

Re-anchor the sentinel on something the migration does not consume.

**Count all declared routes, not just branded ones.** Enumeration health is what the sentinel is about, and the total never goes to zero:

```ts
expect(routes.length).toBeGreaterThan(100);
```

**Anchor on a path that is deliberately staying branded.** The eight entries in `FINAL_PROVIDER_CONSOLE_PATHS` are held by third-party provider consoles and are explicitly excluded from #28278, which makes them the most stable branded declarations in the package. `POST /api/okou/slack/events` is one; any of the eight works.

Leave a comment saying why that anchor was chosen, and what a future editor should do if those eight ever move: re-anchor, do not delete the sentinel.

## Keep unchanged

The second test — `declares every branded contract path in the Okou namespace` — and its failure message. That is the guard #28350 was written for, and it stays correct as the branded set shrinks: an empty set trivially contains no `/api/zero/` declaration, which is exactly the desired end state, and the sentinel above is what stops that from being vacuous.

## Do not

- Do not lower the threshold to 99 or any other number. The count keeps falling; this returns within one merge.
- Do not delete either test.
- Do not move any contract path in this PR. Slices own their own paths, and touching one here would collide with the slice that owns it.

## Tests

- The suite passes on current `main`.
- It still passes when the branded route count is zero. Prove it with a unit-level check of the sentinel's own logic rather than by waiting for the migration to finish.
- The `/api/zero/` guard still fails when a contract declares a legacy path.

lancy:
Closes #28525.

## Problem

`api-namespace-declarations.test.ts` (added by #28350) has two tests. The second is the guard that matters — no contract may declare a `/api/zero/` path. The first is a sentinel that keeps that guard honest: `declaredRoutes()` walks the package barrel, and if the walk ever returned nothing, the `/api/zero/` assertion would pass vacuously against an empty array.

Both of the sentinel's assertions were measured against the set #28278 is deliberately draining:

```ts
expect(brandedRoutes.length).toBeGreaterThan(100);
expect(brandedRoutes.some(... "/api/okou/billing/concurrency-checkout/preview")).toBe(true);
```

The floor fails at 100, then 99, then once per merged slice. Neither assertion is wrong about what it wants; both are counting something that is being removed on purpose.

**#28486 landed on `main` while this was open and confirms the pattern.** It re-anchored the named sample onto `POST /api/okou/slack/events` — the same anchor #28525 asked for — but kept the floor on `brandedRoutes` and lowered it from 100 to 50. That buys one or two more slices and then returns: the branded count is heading to the eight retained provider-console paths.

## Change

**Count all declared routes, not just branded ones.** Enumeration health is what the sentinel is about, and migrating a path rewrites its namespace without dropping the declaration, so the total does not fall as slices land.

Measured across the #28486 merge, on the same barrel walk:

| | before #28486 | after #28486 |
|---|---|---|
| total declared routes | 407 | **407** |
| branded routes | 114 | **100** |

The branded count fell by 14 while the total held exactly. A floor of 100 against a stable 407 does not degrade as #28278 proceeds; a floor of 50 against a falling branded count has one slice of headroom left.

**Anchor on a path that is deliberately staying branded.** `POST /api/okou/slack/events` is one of the eight `FINAL_PROVIDER_CONSOLE_PATHS` in `apps/api`; a third-party provider console holds each of those URLs, which is why #28278 excludes them and why they stay branded under #26701. This keeps #28486's anchor choice and records in a comment why it was chosen, plus what to do if those eight ever move: re-anchor, do not delete the sentinel.

The floor was not lowered again, no test was deleted, and no contract path was moved — slices own their own paths.

## Guard unchanged

`declares every branded contract path in the Okou namespace` keeps its subject, its assertion, and its failure message verbatim. Its filter moved to a `legacyNamespaceDeclarations()` helper so the added coverage below runs the same code the guard runs, rather than a copy that could drift.

## Added coverage

Two cases prove the sentinel's own logic instead of waiting for the migration to finish:

- `clears the floor on neutral routes alone` — asserts the routes that are *already* neutral today clear the floor by themselves, against the real barrel walk. A migrating route becomes neutral rather than disappearing, so that subset is a lower bound on the total once the branded count reaches the eight retained paths. Today that is **307 neutral routes against a floor of 100**, and the count only rises as slices land — so draining the branded set to zero cannot reach the floor. Verified as a real boundary, not a tautology: the assertion passes at a floor of 306 and fails at 307.
- `reports a contract that declares a legacy namespace path` — the guard still produces a finding when a contract declares `/api/zero/health`. This is the only positive-detection coverage the guard has; without it, its filter is only ever asserted against an empty result.

## Verification

Run after merging current `main` (which includes #28486) into this branch:

```
pnpm -F @okouai/api-contracts exec vitest run   # 32 files, 595 tests passed
pnpm -F @okouai/api-contracts run lint          # clean
pnpm -F @okouai/api-contracts run check-types   # clean
pnpm knip --workspace packages/api-contracts    # clean
pnpm exec prettier --check <file>               # clean
```


lancy:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 8bd0339416b8237f57e32b0478e70043fc00add0

### Summary

Re-anchors the enumeration sentinel in `api-namespace-declarations.test.ts` so it no longer measures the branded route set that #28278 is deliberately draining: the floor moves from `brandedRoutes.length > 50` to `routes.length > 100` over all declared routes, and the named sample keeps #28486's `POST /api/okou/slack/events` anchor with a comment recording why that path stays branded. The `/api/zero/` guard keeps its subject, assertion, and failure message; its filter is extracted to a shared helper so added coverage exercises the same code the guard runs.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic — no wrong assumptions or unhandled edge cases | Finding (P1) | `api-namespace-declarations.test.ts:158` — the floor assertion in the added end-state case runs against a hand-built array of known length; see P1 below. |
| `REVIEW.md` | Type safety — no `any`, no broad types | Pass | All added helpers take `readonly DeclaredRoute[]` and return `boolean` / `readonly string[]`; `RETAINED_BRANDED_ROUTE` is `as const`. No `any`. |
| `REVIEW.md` | Style — comments explain WHY, not WHAT | Pass | Both new docblocks explain why the floor counts all routes and why the anchor was chosen, plus the re-anchor instruction the issue asked for. |
| `REVIEW.md` | Testing coverage for commit type | Pass | `test:` commit; it is itself the test change. Package suite re-run green after merging `main`. |
| `REVIEW.md` | Feature regression guard — existing behavior not silently replaced | Pass | The `/api/zero/` guard's subject, `toStrictEqual([])` assertion, and failure message are byte-identical; only its filter moved into `legacyNamespaceDeclarations()`. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, contract, or `jsonb(...).$type<...>()` change; the diff touches one test file. |
| `docs/bad-smell.md` | Zero tolerance for `any` and for lint/type suppressions | Pass | No `any`, no `eslint-disable` / `@ts-ignore` / `@ts-expect-error` / `prettier-ignore` in the diff. |
| `docs/bad-smell.md` | No hardcoded URLs or environment-specific values | N/A | The literal paths are API path templates under assertion, not configuration. |
| `docs/bad-smell.md` | Fail fast — no fallback/recovery logic | Pass | No `??`/`||` default or recovery branch added. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | `reports a contract that declares a legacy namespace path` asserts the guard *detects* a violation, so it fails for a real reason if the detection breaks. That is a live rule check, not a tombstone pinning the absence of deleted code. |
| `docs/fallback.md` | Every fallback introduced by the diff declared in the PR summary | N/A | The diff introduces no fallback, so no `Fallbacks` section is required. |
| `docs/testing.md` | Integration tests only — no unit tests for internal functions | Finding (P1) | `api-namespace-declarations.test.ts:142` — see P1 below. The extracted helpers are test-local predicates, not production code, but the end-state case tests synthetic data rather than the package. |
| `docs/testing.md` | Mock at the boundary; no internal `vi.mock`, no fetch stubbing, no fake timers, real DB | Pass | The file has no mocks, no timers, and no DB access; it walks the real package barrel. |
| `docs/testing/patterns.md` | Pattern 2 — unit-test a pure function only with strong security/complexity justification | Finding (P1) | The end-state case is a pure-function test whose central assertion is self-proving; the same claim is available against real barrel data. |
| `docs/testing/anti-patterns.md` | AP-8 — do not over-test self-evident behavior | Finding (P1) | `expect(endState.length).toBeGreaterThan(MINIMUM_DECLARED_ROUTES)` where `endState` was just constructed with `MINIMUM_DECLARED_ROUTES + 1` entries. |
| `docs/testing/anti-patterns.md` | AP-7 — do not test implementation details | Pass | The retained assertions describe package-level facts (route count, declared anchor, namespace policy), not internal call sequences. |

### Findings

None.

- `turbo/packages/api-contracts/src/contracts/__tests__/api-namespace-declarations.test.ts:142-161` — the `keeps the sentinel true once every migrating path is neutral` case builds `migrated` with exactly `MINIMUM_DECLARED_ROUTES + 1` entries and then asserts `endState.length` exceeds `MINIMUM_DECLARED_ROUTES`. That assertion restates the construction and cannot fail, which is AP-8 over-testing, and it proves nothing about the real barrel: if `declaredRoutes()` were to fall below the floor once the migration completes, this case would still pass. The issue's requirement — "it still passes when the branded route count is zero" — is checkable against real data instead, because every migrating branded route becomes a neutral route rather than disappearing:

  ```ts
  const neutralRoutes = routes.filter(({ path }) => {
    return brandedApiNamespace(path) === undefined;
  });
  expect(neutralRoutes.length).toBeGreaterThan(MINIMUM_DECLARED_ROUTES);
  ```

  On current `main` that is 307 neutral routes against a floor of 100, and the count only rises as slices land — so it demonstrates the floor survives the branded set reaching zero, using the real package walk and with no synthetic fixture. Replace the synthetic case with this form. `reports a contract that declares a legacy namespace path` should stay: there is no non-synthetic way to prove the guard's filter still matches, and without it the guard is only ever asserted against an empty result.

### Testing

- Coverage: Adequate for a `test:`-scoped change once the P1 above is addressed. The floor, the anchor, and the `/api/zero/` guard each retain an assertion, and the guard gains its first positive-detection case.
- Conventions: Compliant on mocking, timers, database, and type safety. The one deviation is the synthetic end-state case flagged above under `docs/testing/patterns.md` Pattern 2 and `docs/testing/anti-patterns.md` AP-8.
- Verification re-run after merging current `main` (which includes #28486): `pnpm -F @okouai/api-contracts exec vitest run` — 32 files, 595 tests passed; lint, `check-types`, `knip`, and `prettier --check` clean.
- Context worth recording: #28486 landed on `main` while this PR was open and independently re-anchored the sample onto `POST /api/okou/slack/events`, but kept the floor on `brandedRoutes` and lowered it from 100 to 50. Measured across that merge on the same barrel walk, total declared routes held at 407 while branded routes fell 114 → 100, which is the direct evidence that the floor belongs on the total rather than on the branded subset.

lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: 2219ecf6f33042226131370a73798a72be570c2b
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Re-anchors the enumeration sentinel in `api-namespace-declarations.test.ts` off the branded route set that #28278 is deliberately draining: the floor moves from `brandedRoutes.length > 50` to `routes.length > 100` over all declared routes, and the named sample keeps #28486's `POST /api/okou/slack/events` anchor with a comment recording why that path stays branded and what to do if it ever moves. The `/api/zero/` guard keeps its subject, assertion, and failure message; its filter is extracted to a shared helper so the added positive-detection case exercises the same code the guard runs. The P1 from the previous HEAD is resolved — the synthetic end-state fixture is replaced by an assertion against the real barrel walk.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic — no wrong assumptions or unhandled edge cases | Pass | The floor's premise is that migrating a path rewrites its namespace without dropping the declaration. Verified across the #28486 merge on the same barrel walk: total held at 407 while branded fell 114 → 100. |
| `REVIEW.md` | Type safety — no `any`, no broad types | Pass | Helpers take `readonly DeclaredRoute[]` and return `boolean` / `readonly string[]`; `RETAINED_BRANDED_ROUTE` is `as const`. No `any`, no `as` narrowing. |
| `REVIEW.md` | Style — comments explain WHY, not WHAT | Pass | Both docblocks and the inline comment on `clears the floor on neutral routes alone` explain why the floor counts all routes, why that anchor was chosen, and why the neutral subset bounds the end state. |
| `REVIEW.md` | Duplicated logic that should be shared | Pass | `legacyNamespaceDeclarations()` is the single implementation behind both the policy guard and its detection case, so the added coverage cannot drift from the guard. |
| `REVIEW.md` | Testing coverage for commit type | Pass | `test:`-scoped change; it is itself the test change. Package suite green: 32 files, 595 tests. |
| `REVIEW.md` | Feature regression guard — existing behavior not silently replaced | Pass | The `/api/zero/` guard's subject, `toStrictEqual([])` assertion, and failure message are byte-identical to `main`; only its filter moved into a helper. No contract path is moved, so no slice of #28278 is collided with. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, contract, or `jsonb(...).$type<...>()` change; the diff is one test file. |
| `docs/bad-smell.md` | Zero tolerance for `any` and lint/type suppressions | Pass | No `any`, no `eslint-disable` / `@ts-ignore` / `@ts-expect-error` / `prettier-ignore`. `pnpm -F @okouai/api-contracts run lint` and `check-types` clean. |
| `docs/bad-smell.md` | No hardcoded URLs or environment-specific values | N/A | The literal paths are API path templates under assertion, not configuration. |
| `docs/bad-smell.md` | Fail fast — no fallback/recovery logic | Pass | No `??`/`||` default or recovery branch added. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | `reports a contract that declares a legacy namespace path` asserts the guard *detects* a violation, so it fails for a real reason if detection breaks. That is a live rule check, not a tombstone pinning the absence of deleted code. |
| `docs/fallback.md` | Every fallback introduced by the diff declared in the PR summary | N/A | The diff introduces no fallback, so no `Fallbacks` section is required. |
| `docs/testing.md` | Integration tests only — no unit tests for internal functions | Pass | Three of the four cases assert against the real package barrel. The one synthetic case covers a test-local predicate with no production caller, and there is no non-synthetic way to prove the guard's filter matches — the package cannot declare a `/api/zero/` path without failing the guard it exists to enforce. |
| `docs/testing.md` | Mock at the boundary; no internal `vi.mock`, no fetch stubbing, no fake timers, real DB | Pass | No mocks, timers, or DB access anywhere in the file; it walks the real barrel. |
| `docs/testing/patterns.md` | Pattern 2 — unit-test a pure function only with strong justification | Pass | The previously flagged synthetic fixture is gone. `clears the floor on neutral routes alone` now filters the real `routes`. |
| `docs/testing/anti-patterns.md` | AP-8 — do not over-test self-evident behavior | Pass | The self-proving `endState.length` assertion is removed. The replacement was mutation-checked as a real boundary: it passes at a floor of 306 and fails at 307 against the actual 307 neutral routes. |
| `docs/testing/anti-patterns.md` | AP-7 — do not test implementation details | Pass | Assertions describe package-level facts — route count, declared anchor, namespace policy — not internal call sequences. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None. The P1 raised against `8bd0339` — a floor assertion on a hand-built array of known length, which could not fail and said nothing about the real barrel — is resolved in `2219ecf`.

### Testing

- Coverage: Adequate. The floor, the anchor, and the `/api/zero/` guard each retain an assertion; the guard gains its first positive-detection case; and the floor's survival of #28278 is now demonstrated against real data instead of a fixture.
- Conventions: Compliant. No mocks, no fake timers, no internal-path `vi.mock`, no `any`, no suppressions.
- Verification on this HEAD, after merging current `main` (which includes #28486):
  - `pnpm -F @okouai/api-contracts exec vitest run` — 32 files, 595 tests passed
  - `pnpm -F @okouai/api-contracts run lint` — clean
  - `pnpm -F @okouai/api-contracts run check-types` — clean
  - `pnpm knip --workspace packages/api-contracts` — clean
  - `pnpm exec prettier --check` on the changed file — clean
- Boundary evidence: 407 total declared routes, 100 branded, 307 neutral. The floor of 100 is cleared by the neutral subset alone, so #28278 driving branded to the eight retained provider-console paths cannot reach it. That is the property #28525 asked to be proven without waiting for the migration to finish.

Issue #26239, opened 2026-08-11
feat(platform): upload and use browser-rendered presentation templates


bingjiez666:
Part of #21350. Depends on #26571, #26576, and #26238.

Add the complete browser UX for uploading, previewing, analyzing, managing, and selecting private presentation templates.

## Upload and analysis flow

1. The first cell of the presentation template grid is an upload tile accepting `.pptx` only.
2. Render the selected local deck through #26571. The user sees the same render tree that will produce the persisted page PNGs.
3. Wait for fonts and renderer-owned assets, then export the fixed 16:9 ordered PNGs.
4. Upload the original PPTX and every PNG through the existing private upload preparation flow.
5. Commit the complete upload. Only this commit starts template analysis.
6. Show “Upload complete. Analyzing template…” and require no further user action.
7. Poll while a template is `processing`; stop when it becomes `ready` or `failed`.

## Catalog and management

- Grid order: upload tile, the user's templates, then built-in templates.
- Local rendering/upload progress belongs to the upload UI; API-backed cards begin once the ingestion is committed.
- `processing`: non-selectable analysis state.
- `failed`: source filename, stable user-readable reason, retry, and delete.
- `ready`: persisted cover/page PNGs, `Use`, hover scrubbing, detail preview, rename, and delete.
- Ready previews always use the persisted browser-rendered PNGs; never re-render the PPTX on another device.
- Selecting a ready card produces `user-template:<uuid>` in the existing presentation selection payload. The subsequent generation run consumes the complete visual guidance package and directly authors presentation HTML/CSS/SVG through #26238.
- Refresh private page URLs through authenticated API signals when signed URLs expire.

## Privacy

- Do not expose permanent CDN URLs, R2 keys, or bucket names.
- Source PPTX, page PNGs, and visual guidance package remain private; every read is authorized.

## Acceptance

- Preview and uploaded PNG output come from the same completed browser render.
- Analysis cannot begin before all source/page uploads are committed.
- Closing and reopening the picker after commit shows the persisted processing/ready/failed state.
- A processing or failed card is never selectable.
- Selecting a ready private template dispatches the existing namespaced id without exposing package internals to the browser.
- Built-in template browsing, selection, previews, and generation are unchanged.
- Rendering, uploading, polling, signed-URL refresh, retry, rename, delete, and selection have integration coverage.

## Deferred

- Public and organization sharing, including visibility controls.
- Re-import, revisions, rollback, archive/restore, and manual design-system editing.

[...]

PR #28533, opened 2026-08-21
feat(platform): list the user's imported decks in the presentation picker

bingjiez666:
## What

The presentation tab of the template picker now shows the decks the signed-in user has already imported, read from `GET /api/okou/presentation-templates`.

Until now that route had no browser consumer at all: a user could hand over a deck through the import tile, watch the analysis finish in its thread, and then reopen the picker to find nothing had changed.

Grid order is import tile → this user's decks → built-in templates, matching #26239.

## How

- `signals/zero-page/presentation-template-library.ts` — `ownPresentationTemplates$` calls the contract client through `accept(..., [200])`, and `refreshOwnPresentationTemplates$` bumps its version counter.
- `setTemplatePickerOpen$` refreshes the catalog as the dialog opens rather than caching it for the session. Both facts a card renders go stale on their own: an import finishes minutes after the deck was handed over, in a thread the user has navigated away from, and every `coverUrl` is a short-lived signed URL.
- `ImportedPptCard` renders the persisted cover, the title, and the slide count, reusing the existing tile geometry so an imported deck sits flush with the built-ins.
- `mocks/handlers/api-presentation-templates.ts` serves an empty catalog by default and exposes `setMockPresentationTemplates()` for tests.

## Why the cards have no `Use` button

A private template is selected as `user-template:<uuid>`, and the API that resolves that id for a generation run is #26238, which is still open — the string appears nowhere in the tree today. A `Use` button now would dispatch an id nothing can resolve, moving the failure out of a disabled control and into the run the user just paid for. Selection lands with #26238.

## Feature-switch behaviour

With `PresentationTemplates` off, the signal answers an empty catalog without issuing a request. That is not a fallback for a failed call: the route replies `403` to a caller without the switch, and `accept` would turn that into an error toast on a dialog the user opened to browse built-in templates.

## Testing

`chat-composer-template-picker.test.tsx` gains a case that seeds one imported template, opens the picker, and asserts the card renders its cover, title, and `18 slides`, that it sits immediately after the import tile and ahead of the built-ins, and that it exposes no button. Ran locally: 38/38 in that file pass, `pnpm -F @okouai/app lint` and `check-types` exit 0, and `pnpm knip` exits 0. Ten locales carry the new `importedSlides` plural forms, including the `_many` category the i18n check requires for `es-ES`, `fr-FR`, `it-IT`, and `pt-BR`.


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 0884b338923d9de430aa48c0f3a044d00ed3954d
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Adds the first browser consumer of `GET /api/okou/presentation-templates`: the presentation tab of the template picker now renders the signed-in user's imported decks as cards between the import tile and the built-in templates. The catalog is a reactive `computed(async ...)` invalidated by a counter that `setTemplatePickerOpen$` bumps on open, and the whole surface is gated by the non-GA `FeatureSwitchKey.PresentationTemplates` switch, which short-circuits to an empty catalog without issuing a request.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic — no race conditions, unhandled null paths, or API misuse | Pass | `presentation-template-library.ts` reads the version counter after the switch guard, so the fetch is subscribed only when enabled. `zero-chat-composer.tsx:5482` gates `PptTemplateGrid` on `selectedCategory === "slides"`, so opening the picker on another tab bumps the counter without issuing a request. `pptItems` comes from the static `PRESENTATION_TEMPLATE_PICKER_ITEMS` (`:5932`), so the `pptItems.length > 0` branch cannot hide the imported cards. |
| `REVIEW.md` | Type safety — no `any`, no overly broad types | Pass | `PresentationTemplateSummary` is re-exported from the contract; `ImportedPptCard` takes it directly. No casts in the diff. |
| `REVIEW.md` | Security — no missing auth, injection, or hardcoded secrets | Pass | No new endpoint. `client.list()` goes through `zeroClient$`, which carries auth; ownership is enforced server-side by the existing owner-scoped route. |
| `REVIEW.md` | Feature regression guard — existing UI/flow not changed without containment and coverage | Pass | With the switch off (the default the other 37 cases in `chat-composer-template-picker.test.tsx` run under) the signal returns `[]` and the grid is byte-identical to before. The new row only appears behind `FeatureSwitchKey.PresentationTemplates`. |
| `REVIEW.md` | New user-facing features gated behind a `FeatureSwitchKey` (P1) | Pass | `feature-switch.ts:311` — `PresentationTemplates`, `enabled: false`, no `enabledOrgIdHashes`. `presentationTemplateImportEnabled$` reads it and `ownPresentationTemplates$` short-circuits on it. |
| `REVIEW.md` | DB/JSONB review gate — persisted contract or JSONB shape changes need migration/normalizer/rationale | N/A | No change under `turbo/packages/db/src/schema/**` or `turbo/packages/api-contracts/src/contracts/**`; the diff is platform-only and consumes the existing `presentationTemplateSummarySchema`. |
| `docs/bad-smell.md` | Zero `any` types and zero lint/type suppressions | Pass | Neither appears in the diff. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | All new imports are static, top-of-file. |
| `docs/bad-smell.md` | No hardcoded URLs or environment-specific values | Pass | Only `https://example.com/imported-cover.png` inside the test fixture. |
| `docs/fallback.md` | No fallback for a state the contract already prevents (§3) | Pass | `template.coverUrl === null ? null : ` guards a field the contract declares `z.string().url().nullable()`. `useLastResolved(...) ?? []` covers the genuine pre-resolution `undefined`, matching the documented consumer pattern and existing call sites at `zero-chat-composer.tsx:657/660/7000`. |
| `docs/fallback.md` | A non-GA feature-switched feature needs no compatibility code (§2) | Pass | The switch-off empty catalog is the feature gate, not a compatibility branch; the comment on `ownPresentationTemplates$` states why (`403` for a caller without the switch would become an error toast). No dual-read, migration, or version-tolerance branch is added. |
| `docs/fallback.md` | Every fallback introduced by the diff is declared in a `Fallbacks` PR-summary section (§9, P1) | N/A | The diff introduces no fallback, so no `Fallbacks` section is required and none is expected. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed (§1, P1) | Pass | `expect(queryAllByRoleFast("button", card)).toHaveLength(0)` pins a deliberate scope boundary for behavior that has not been built yet (selection lands with #26238), not a tombstone for deleted code. |
| `docs/deployment-compatibility.md` | Changes to a deployable boundary must tolerate the old/new version combination | N/A | No contract, runner protocol, queue payload, or persisted shape changes. `GET /api/okou/presentation-templates` already ships on `main` (`turbo/apps/api/src/signals/routes/presentation-templates.ts`), so the new client calls a route the deployed API already serves — and the caller is gated by a non-GA switch, which §2 of `docs/fallback.md` exempts from these windows. |
| `.claude/skills/ccstate/SKILL.md` | Prefer reactive `computed(async ...)` with counter-bump invalidation over imperative fetch-and-store | Pass | `ownPresentationTemplates$` / `ownPresentationTemplatesVersion$` / `refreshOwnPresentationTemplates$` reproduce the documented pattern exactly, and match in-tree precedent (`chat-page/mail-draft.ts:236`, `header-automation-menu.ts:198`, `browser-session-block.ts:441`). |
| `.claude/skills/ccstate/SKILL.md` | All `zeroClient$` calls use `accept` with an explicit status list; background `computed` lets errors propagate | Pass | `await accept(client.list(), [200])` with no try/catch and no view-layer error handling — the "fail fast for background fetches" shape. |
| `.claude/skills/ccstate/SKILL.md` | No manual cache layered on top of `computed` memoization | Pass | No `Map`/`Set` cache; the counter is the only invalidation mechanism. |
| `docs/effect.md` | User or domain events belong in a semantic `command`, not a React effect | Pass | The refresh runs inside `setTemplatePickerOpen$`, the command that already owns the open transition; no `useEffect` is added. |
| `docs/effect.md` | React render stays pure — no command execution or Store writes during render | Pass | `ImportedPptCard` and `PptTemplateGrid` only read (`useTranslation`, `useLastResolved`). All three `setTemplatePickerOpen$` call sites (`zero-chat-composer.tsx:5741/5823`, `zero-start-cards.tsx:421`) are event handlers. |
| `docs/cache.md` | RC-3 no unbounded-lifetime caches; RC-5 no duplicate mutable source of truth | Pass | The server list is never copied into a second mutable store; the module-scope state is a single `number` counter, not a keyed cache. |
| `docs/cache.md` | RC-8 asymmetric resource lifecycles — acquired resources need owned teardown | N/A | The diff adds no listener, timer, observer, object URL, or subscription; the only async work is one `computed` fetch owned by ccstate's dependency graph. |
| `docs/testing.md` | `feat:` commits must have integration tests (missing = P0) | Pass | `chat-composer-template-picker.test.tsx` adds a page-level case covering cover, title, plural slide count, grid position, and the absence of a `Use` control. |
| `docs/testing.md` | Mock at the boundary only — no `vi.mock()` on relative paths, no direct `fetch` stubbing, MSW for HTTP | Pass | The new handler is an MSW contract handler (`mockApi(presentationTemplatesContract.list, ...)`); the test adds no `vi.mock` and no `vi.stubGlobal("fetch", ...)`. |
| `docs/testing.md` | No fake timers; no `toHaveBeenCalled()` as the sole assertion | Pass | No `vi.useFakeTimers()`. Every assertion reads rendered DOM, not call counts. |
| `docs/testing/app-testing.md` | Page tests enter through `detachedSetupPage` and assert page-visible behavior | Pass | The case calls `detachedSetupPage({ context, featureSwitches, path })` without awaiting, then `waitFor`s the rendered card. |
| `docs/testing/app-testing.md` | Stateful handler with reset, registered in `handlers/index.ts` and `resetAllMockHandlers()` | Pass | `api-presentation-templates.ts` exports `setMockPresentationTemplates`/`resetMockPresentationTemplates`; both are wired into `handlers/index.ts`, matching the 17 existing `setMock*` handlers. |
| `docs/testing/app-testing.md` | No file-level `afterEach` cleanup; mocks bound to the test-context lifecycle | Pass | Reset happens through the shared `resetAllMockHandlers()` path, not a new `afterEach` in the test file. |
| `docs/testing/testing-external-behavior.md` | Construct and verify through the page, not internal implementation | Pass | Setup is a real picker click; assertions read the rendered cover, title, and caption. The `data-imported-presentation-template` / `data-presentation-template-import` selectors used to scope the grid position follow the established convention in this same file (`data-composer-inline-template`, `data-avatar-template-preview`, `data-presentation-template-grid-scroll`); the store is never mutated and no component is rendered directly. |
| `docs/testing/patterns.md` | Pattern 3 — MSW handlers live in `mocks/handlers/`, test files do not manage server lifecycle | Pass | No `server.listen`/`resetHandlers`/`close` in the test; the handler module owns the fixture state. |
| `docs/testing/anti-patterns.md` | AP-1 test behavior not mock calls; AP-2 no direct fetch mocking; AP-4 no mocking internal code | Pass | The only import from `mocks/` is the fixture setter, which is the documented stateful-handler seam rather than a `vi.mock` of internal code. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

Non-blocking observations, no action required:

- `presentation-template-library.ts:refreshOwnPresentationTemplates$` uses `set(v$, get(v$) + 1)` where the documented example and the in-tree precedents (`mail-draft.ts:255`, `header-automation-menu.ts:201`) use the updater form `set(v$, (prev) => prev + 1)`. Behaviourally identical here since the command is synchronous.
- Because the imported cards resolve asynchronously and are inserted ahead of the built-ins, the presentation grid's restored scroll position (`onRestorePresentationScroll`) can shift once the catalog lands. Only reachable behind the non-GA switch, and only on the first open of a session.

### Testing

- Coverage: Adequate. The `feat:` commit ships an integration test at the page level covering the rendered card contents (cover `src`, title, `18 slides` plural form), its position immediately after the import tile and ahead of the built-ins, and the deliberate absence of a `Use` control. The switch-off path — the flow most likely to regress — stays covered by the 37 pre-existing cases in the same file, which run without `FeatureSwitchKey.PresentationTemplates` and therefore exercise the unchanged grid.
- Conventions: Compliant. Page-level entry via `detachedSetupPage`, MSW contract handler with reset wired into `resetAllMockHandlers()`, no relative-path `vi.mock`, no `fetch` stubbing, no fake timers, no `toHaveBeenCalled()` assertions, no file-level `afterEach`.
- Not re-run locally by this review; the PR pipeline owns execution. The author reports 38/38 in `chat-composer-template-picker.test.tsx`, `pnpm -F @okouai/app lint` and `check-types` at exit 0, and `pnpm knip` at exit 0.

Issue #21350, opened 2026-07-14
feat: let users upload a deck and reuse its design as a private presentation template


bingjiez666:
## Summary

Let a user upload one `.pptx` deck, have vm0 render it in the browser exactly as vm0 will preview it, and turn the committed source plus page PNGs into a private reusable presentation template.

The uploaded deck is not replayed during generation. The analysis agent extracts a design-system package; generation still produces a normal vm0 `presentation-html` artifact.

## Revised rendering decision

The earlier sandbox conversion design is superseded.

- PPTX pages are rendered in the vm0 browser, not through sandbox LibreOffice/PDF/Poppler conversion.
- The visible browser preview and persisted page PNGs come from the same completed render tree/surface.
- The contract is parity with the vm0 browser preview, not pixel parity with a user's local PowerPoint installation.
- Fonts, images, charts, and renderer-owned assets must finish loading before export.
- A ready template always previews its persisted PNGs; it does not re-render the PPTX on another device.
- Version 1 is fixed at 16:9. There is no user-selectable or detected template aspect ratio.

## Product scope

- Accepted source: `.pptx` only. PDF, HTML, legacy `.ppt`, `.odp`, `.key`, `.pptm`, encrypted, and malformed files are rejected.
- Templates are private to the owner in this scope.
- Ready templates can be selected through the existing presentation picker and generation request shape.
- Rename and delete are included.
- Public/organization visibility controls, sharing, re-import, revisions, rollback, and manual editing are later work.

## Target lifecycle

1. The user selects a local `.pptx`.
2. The browser parses and renders every slide on the fixed 16:9 vm0 surface.
3. The browser shows that render and exports ordered PNGs from the same completed state.
4. The browser requests narrowly scoped upload URLs and sends the original PPTX plus every page PNG directly to private R2.
5. The browser commits the complete upload. The API verifies the source and page set and only then launches one `template-import` run.
6. The UI reports “Upload complete. Analyzing template…”; no further user action is required.
7. The run downloads the committed PPTX and PNGs, parses PPTX OOXML, and creates the three-file template package.
8. Atomic package publication transitions the row to `ready`; terminal analysis failures transition it to `failed`.
9. A ready card becomes selectable and generation pulls its private package.

```mermaid
sequenceDiagram
    autonumber
    participant U as User
    participant W as vm0 browser
    participant A as API
    participant R as Private R2
    participant G as Analysis agent

    U->>W: Select .pptx
[...]

PR #28536, opened 2026-08-21
feat(api): resolve a user's own presentation template for a generation run

bingjiez666:
Implements #26238. Part of #21350.

## What

A user could import a deck and, since #28533, see it in the picker — but selecting one had nowhere to go. `user-template:<uuid>` matches no built-in registry slug, so `buildGenerationTemplatePrompt` rejected it and the send failed. This resolves that id against the owner's row, authorises it, mounts the published guidance package into the run, and tells the agent how to use it.

## Authorisation happens before dispatch

`validateGenerationTemplatePrompt` now takes the database and the caller. A row that does not exist, was deleted, or belongs to someone else all produce the same `Presentation template not found`, so the message cannot be used to probe which one it was. Only ids that survive that check reach the run.

## Delivery: a mounted volume, not a signed archive URL

The issue text predates the current storage design. `publishPresentationTemplate$` writes the **extracted file tree** into the storage volume `presentation-template@{uuid}` — there is no archive object to sign. The equivalent under today's architecture is the mechanism skills already use: one `additionalVolumes` entry per authorised template, mounted at `./generated/presentation-template/<uuid>`, beside the built-in templates' `./generated/resources/<slug>`.

This satisfies the same acceptance criteria more directly than a signed GET would: the run receives the complete verified tree including binary assets, no raw storage key or URL reaches the client, and there is no expiry to refresh, no download step, and no archive extraction or integrity check to write.

The mount path is a working-directory path rather than a skills-root path because the skills root is chosen per framework inside run creation, while the prompt naming that path is built before a framework exists.

## The prompt is deliberate about what not to do

The package describes a visual language, not a renderer: there is no deck schema, no layout-id service, and no JSON-to-HTML compiler behind it. An agent that reaches for its usual slide-JSON intermediate therefore produces something the guidance cannot inform. The instructions require reading `SKILL.md` and `design-system.md` first, then authoring semantic HTML/CSS/SVG directly for the current content, with Grid/Flexbox for live rows, columns, and text flow and absolute positioning reserved for backgrounds, fixed chrome, decoration, and overlays. Publication reuses the normal `okou host … --artifact-kind presentation-html` path.

## Gating

Private ids are gated on `PresentationTemplates` in the resolver, not at the composer alone, so an API version that still accepts the field stops honouring the namespace the moment the switch is off. `presentationTemplatesEnabled` is threaded to all three prompt call sites — chat send, queued-message callback, and active input — so a follow-up run does not silently lose the guidance.

## Guidance cannot outrun the package

Review caught that threading the switch was not enough. Only the send path added the volume, so a template selected on a message that queues behind an active run, or steered into a running one, produced a prompt whose first instruction was to read a directory the run did not carry. A run's mounts come from its own body and compose volumes — `StoredExecutionContext` records none — so a later run in the same thread does not inherit an earlier mount.

The two are now derived from one lookup instead of being wired separately. `buildGenerationTemplatePrompt` takes the row ids whose packages the run will mount and refuses any private selection outside that set, so a prompt can no longer name a package that is not there. `selectedUserPresentationTemplateIds`, `authorizedUserPresentationTemplateIds`, `userPresentationTemplateVolumes`, and `additionalVolumesForRun` moved into `presentation-template-data.service.ts` and both dispatching paths use them:

- **Chat send** authorises the message's selections, mounts them, and builds the prompt from the authorised ids.
- **Queued-message callback** now does the same at dispatch. Ownership is re-checked rather than trusted from the send that queued the message, because the row can be deleted while the message waits and a volume this user may not read must never be assembled.
- **Active input** steers a run whose volumes were fixed at creation, so it declares no mounts and a private selection contributes no guidance rather than a dangling path.

Selections are also deduplicated, so attaching the same template twice cannot ask for the same mount path twice, and ownership resolves in one `inArray` query instead of one round trip per template.

Built-in template selection, prompts, and colour-system handling are unchanged, and a test pins that they still resolve while the switch is on.

## Two acceptance lines in #26238 are stale

- *"Unknown, non-ready, deleted, or unauthorized template ids fail"* — there is no `status` column any more (removed in #28515). A row exists only once publish succeeds, so `non-ready` is no longer a reachable state. The other three are covered.
- *"Return a short-lived signed GET for the complete private package archive"* — superseded by the volume mount described above.

I did not edit the issue; flagging so the acceptance review is done against reality.

## Testing

- `generation-template-prompt.test.ts`: private ids resolve to the mounted-package guidance and never emit `okou resource pull`, `AGENT_RUNBOOK.md`, or a `presentation-template@` storage key; a private id is rejected while the switch is off; a syntactically invalid private id is rejected as `Malformed presentation template` rather than retried as a registry slug; a built-in id still resolves with the switch on.
- `chat-events.bdd.test.ts`: a well-formed but unreadable private id is rejected `400` with the switch off (`Unknown generation template`) and with it on (`Presentation template not found`), and no event is persisted — proving the failure is before dispatch.

Ran locally against a real Postgres: those two files pass, `pnpm -F api lint`, `pnpm -F @okouai/core lint`, `pnpm -F api check-types` (all seven projects), `pnpm -F @okouai/core check-types`, and `pnpm knip` all exit 0.

- `generation-template-prompt.test.ts` additionally pins the invariant the review found missing: a private id the run does not mount resolves `invalid` with `Presentation template not found` rather than emitting guidance, and one unmountable selection among several invalidates the message instead of silently dropping a template the user attached on purpose.

**Known gap:** the positive end-to-end integration test — publish a package, select it, claim the dispatched run, and assert its storage mounts carry `presentation-template@{uuid}` at the working-directory path — is still not in this PR. It needs the publish fixture (`installS3Fixture`, `tarGz`, `uploadInputs`) extracted from `presentation-template-publish.test.ts` into a shared helper, and that fixture replaces `context.mocks.s3.send` wholesale, so it also has to be reconciled with the S3 behaviour `chat-events.bdd.test.ts` already relies on for agent storage writes. That is a change of its own, not something to smuggle in here. What is covered now is the rule that produced the bug — guidance is emitted only for a mounted id — at every call site; what remains uncovered is the volume list's own trip into the run body.

Two pre-existing tests in `chat-events.bdd.test.ts` fail identically on unmodified `main` in this sandbox (`Job not found in queue` while claiming a runner job); verified by stashing. Not a regression from this branch.


bingjiez666:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 767b64269663dd5e4d75da529477567b192fd46b

### Summary

Resolves a `user-template:<uuid>` presentation selection against the sender's own `presentation_templates` row, authorises it before dispatch behind the non-GA `PresentationTemplates` switch, mounts the published package as the `presentation-template@{uuid}` volume at `./generated/presentation-template/<uuid>`, and emits prompt guidance that requires reading `SKILL.md`/`design-system.md` and authoring HTML/CSS/SVG directly. The authorisation design is correct and fails closed; the gap is that only one of the three wired prompt call sites also mounts the package.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic — no unhandled edge cases or wrong assumptions | Finding (P1) | `chat-events.command.ts:3207` is the only site adding `additionalVolumes`; the queued-message and active-input paths emit the mount-path guidance without the mount. |
| `REVIEW.md` | Security — missing auth checks on new capability | Pass | `validateGenerationTemplatePrompt` (`chat-events.command.ts:1123-1155`) requires a row matching `(orgId, ownerUserId, id)` before any volume is assembled; volumes resolve through `volumeStorageLookup(runtimeOrgId, …)` (`agent-run-storage.service.ts:2434`) and `runtimeOrgId` is the caller's `args.orgId`. |
| `REVIEW.md` | No-existence-leak behaviour on a private resource | Pass | Absent, deleted, and other-owner rows all return `Presentation template not found`; there is no `deleted_at` column on `presentation_templates`, so a delete really is an absent row. |
| `REVIEW.md` | Type safety — no `any`, no overly broad types | Pass | No `any` in the diff; `parseUserPresentationTemplateId` returns `string \| undefined` and every consumer narrows it. |
| `REVIEW.md` | Maintainability — no oversized functions, no duplicated logic | Pass | `resolveSelectedTemplateContext` and `resolveQueuedMessageTemplateContext` extract what would otherwise be repeated switch-threading; `prepareNormalSend$` is shorter than before. |
| `REVIEW.md` | Feature regression guard — existing flow most likely to regress is covered | Pass | `generation-template-prompt.test.ts` pins that a built-in id still resolves to `okou resource pull template:html-ppt-playful-launch-runbook` while the switch is on. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, migration, contract, or `jsonb().$type<>()` change; `presentation_templates` and `additionalVolumes` are both pre-existing. |
| `docs/bad-smell.md` | Zero `any`, zero lint/type suppressions | Pass | No `any`, no `eslint-disable`/`oxlint-disable`/`@ts-*` comments anywhere in the diff. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | All four new imports in `chat-events.command.ts` and the `@okouai/core` re-export are static. |
| `docs/bad-smell.md` | No hardcoded URLs or environment-specific values | Pass | The only literals are the storage-name prefix and the mount directory, both centralised in `storage-names.ts` and `presentation-template-selection.ts`. |
| `docs/bad-smell.md` | Fail fast rather than recover silently | Pass | A missing row returns `400` and stops the send; nothing is defaulted or substituted. |
| `docs/fallback.md` | Every fallback introduced by the diff is declared in the PR summary | Pass | The diff introduces no fallback: `presentationTemplatesEnabled === true` is a fail-closed gate default, and `additionalVolumesForRun` omitting the key on an empty list is optional-field shaping, not a compatibility branch. No `Fallbacks` section is required. |
| `docs/fallback.md` | §2 — a non-GA feature-switched feature needs no compatibility code | Pass | `FeatureSwitchKey.PresentationTemplates` is `enabled: false` with no `enabledOrgIdHashes` (`feature-switch.ts:311`), so no dual-read, migration, or rollback handling is owed and none was added. |
| `docs/fallback.md` | §1 — no negative tests asserting removed behaviour stays removed | Pass | The new negative assertions cover a fail-closed authorisation boundary and the switch-off state, both live product behaviour, not tombstones. |
| `docs/deployment-compatibility.md` | Backend change safe across old/new frontend and old/new runner | Pass | `additionalVolumes` is an existing run-contract field (`contracts/runs.ts:100`) already honoured by deployed runners; the private namespace is inert for old clients because they never send `user-template:` ids. |
| `docs/externally-managed-references.md` | Validate representation and resolve existence separately | Pass | `parseUserPresentationTemplateId` proves shape only; `loadOwnedPresentationTemplate` resolves existence, and the two produce distinct messages (`Malformed presentation template` vs `Presentation template not found`). |
| `docs/externally-managed-references.md` | Security fails closed — an unresolved reference grants nothing | Pass | The volume list is built only from ids that survived the row check, so an unauthorised id can never reach `storageManifestRequests`. |
| `docs/externally-managed-references.md` | A resolution miss is a normal domain result, not a 500 | Pass | The miss returns `badRequestMessage(...)`; DB failures are left to propagate rather than being wrapped as "not found". |
| `docs/testing.md` | Mostly integration; mock only at the system boundary | Pass | The new BDD case drives `POST /api/chat/events` and `GET` thread events through the API; no `vi.mock()` of internal modules, real database, no fake timers. |
| `docs/testing.md` | `feat:` commits must have integration tests | Finding (P1) | The rejection path has an API integration test, but the volume assembly (`presentationTemplateVolumes`, `additionalVolumesForRun`, and the `buildCreateZeroRunArgs` wiring) has no test at any level. |
| `docs/testing/api-testing.md` | Construct and verify through API endpoints, never DB rows or services | Pass | The new test builds state with `bdd.createAgent` / `chat.requestSendEvent` and asserts through `chat.requestThreadEvents`; it imports no DB schema and calls no service directly. |
| `docs/testing/api-testing.md` | Uniquely owned identities so the test is correct under concurrency | Pass | The case uses `bdd.user()`, a per-test agent, and a fresh `randomUUID()` template id, so it cannot collide with a concurrent file. |
| `.claude/skills/ccstate/SKILL.md` | No `detach()`/`Reason.DomCallback` in the signals directory; commands own their abort signal | Pass | `prepareNormalSend$` stays a `command`, the new `await` is followed by `signal.throwIfAborted()` (`chat-events.command.ts:2579`), and no floating promise or DOM-callback pattern was introduced. |

### Findings

None.

- `turbo/apps/api/src/signals/services/internal-chat-run-callback.service.ts:3061` and `turbo/apps/api/src/signals/services/active-input-prompt.service.ts:295`: **the guidance is threaded to three call sites but the package is mounted at only one.** `buildCreateZeroRunArgs` (`chat-events.command.ts:3207`) is the sole place that adds `additionalVolumes`. `buildQueuedCreateZeroRunArgs` (`internal-chat-run-callback.service.ts:873-960`) never sets the field and `CreateQueuedChatRunInput` has no slot for it, and the active-input path injects text into a run whose volume set was fixed at creation. A run's mounts come from its own body/compose volumes — `StoredExecutionContext` in `contracts/runners.ts` carries no `additionalVolumes`, so a later run in the same thread does not inherit an earlier run's mount. Failure scenario: with the switch on, a user selects their imported deck on a message that is queued behind an active run, or on a steering input to a run that did not already carry that template. The dispatched run's prompt opens with `Read ./generated/presentation-template/<uuid>/SKILL.md and ./generated/presentation-template/<uuid>/design-system.md before authoring anything`, that directory does not exist, and the agent silently falls back to an un-templated deck after burning turns on the missing files. Either mount the volume in the queued path too, or suppress the private-template guidance where the package cannot be mounted so the failure is a visible rejection rather than dangling instructions.
- `turbo/apps/api/src/signals/routes/__tests__/generation-template-prompt.test.ts:115` and `turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts:9015`: **no test asserts that a dispatched run carries the volume.** The prompt text is pinned directly and the rejection path is covered through the API, but nothing exercises `presentationTemplateVolumes` or `additionalVolumesForRun`, so the mount name, the `CANONICAL_WORKING_DIR`-relative mount path, and the send-path wiring are entirely unverified. This is the same gap that let the finding above through: a positive assertion on the dispatched run would have shown immediately that only one of the three paths produces it. The PR declares this as a known gap needing the publish fixture extracted from `presentation-template-publish.test.ts`, and that framing is fair — but at minimum the volume list built from an authorised selection should be asserted before merge, even without a real published package.

### Testing

- **Coverage:** Insufficient — the authorisation and switch-off rejection paths are well covered through the API, and built-in template resolution is pinned against regression, but the volume-mounting half of the feature has no coverage at any level. The author declares the positive end-to-end test as a known gap with a concrete reason; the narrower gap that matters here is that no test observes the run's `additionalVolumes` at all.
- **Conventions:** Compliant. The BDD case mocks nothing internal, uses the real database, uses per-test identities, and both constructs and verifies through API endpoints. `generation-template-prompt.test.ts` tests `buildGenerationTemplatePrompt` directly rather than through a route, which reads as a unit test, but it extends a pre-existing file whose entire purpose is pinning generated prompt text, and the user-visible route behaviour is covered separately in `chat-events.bdd.test.ts` — so this follows established local convention rather than introducing a new unit-test surface.
- **Non-blocking observation:** `validateGenerationTemplatePrompt` awaits `loadOwnedPresentationTemplate` once per selected template in a sequential loop on the send hot path. A single `inArray` lookup would collapse this to one round trip if a message can carry several templates.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 2da22e84057e007b93148c6664dfb735c7c18a3b
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Re-review after `2da22e84`, which addresses the P1 raised against `767b6426`. Guidance for a private presentation template and the volume that backs it are no longer wired separately: `buildGenerationTemplatePrompt` now takes the row ids whose packages the run will mount and rejects any private selection outside that set, so a prompt cannot name a package the run does not carry. The queued-message callback authorises and mounts its own packages at dispatch; active input, which steers a run whose volumes were fixed at creation, declares none and drops the guidance instead of pointing at nothing.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic — no unhandled edge cases or wrong assumptions | Pass | The defect from the previous head is closed at the shared chokepoint (`generation-template-prompt.ts:240`) rather than per call site, so a fourth caller inherits the fail-closed default. Confirmed the queued path reaches real plumbing: `buildQueuedCreateZeroRunArgs` → `createQueueFirstZeroRun$` → `zero-runs-create.service.ts:689`, the same `body.additionalVolumes` the send path uses. |
| `REVIEW.md` | Security — missing auth checks on new capability | Pass | The queued path re-checks ownership at dispatch instead of trusting the send that queued the message, so a row deleted or transferred while the message waits yields no mount. `authorizedUserPresentationTemplateIds` filters against `(id, orgId, ownerUserId)`, and the run's `runtimeOrgId` is the same org. |
| `REVIEW.md` | No-existence-leak behaviour on a private resource | Pass | Absent, deleted, other-owner, and unmountable all return `Presentation template not found` at both the send route and the prompt builder. |
| `REVIEW.md` | Type safety — no `any`, no overly broad types | Pass | `PresentationTemplateVolume` replaces the two inline `{name, mountPath}` shapes; `readonly` propagates through `PreparedNormalSend` and `CreateQueuedChatRunInput`. |
| `REVIEW.md` | Maintainability — no oversized functions, no duplicated logic | Pass | The duplicated volume/authorisation helpers are consolidated in `presentation-template-data.service.ts` and consumed by both dispatch paths; `resolveQueuedMessageComputerUseHostGrant` keeps `buildCreateQueuedChatRunInput` inside the 128-line rule. |
| `REVIEW.md` | Feature regression guard — existing flow most likely to regress is covered | Pass | Built-in resolution is still pinned; the switch-off path still yields `Unknown generation template`, so GA behaviour is byte-identical to `main` for any message without a `user-template:` id. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, migration, contract, or `jsonb().$type<>()` change. |
| `docs/bad-smell.md` | Zero `any`, zero lint/type suppressions | Pass | None in the diff; `pnpm turbo run lint` reports 0 errors across 13 tasks. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | All new imports are static. |
| `docs/bad-smell.md` | No hardcoded URLs or environment-specific values | Pass | Mount path and storage name are still derived from `CANONICAL_WORKING_DIR`, `userPresentationTemplateDirectory`, and `getPresentationTemplateStorageName`. |
| `docs/bad-smell.md` | Fail fast rather than recover silently | Pass | An unmountable private selection invalidates the message rather than being dropped from a partially-applied template list. |
| `docs/fallback.md` | Every fallback introduced by the diff is declared in the PR summary | Pass | The diff introduces none. `options.mountedUserPresentationTemplateIds ?? []` is a fail-closed default on an optional parameter, matching its two sibling options; it guards no old/new version pairing. No `Fallbacks` section is required. |
| `docs/fallback.md` | §2 — a non-GA feature-switched feature needs no compatibility code | Pass | `PresentationTemplates` is still `enabled: false` with no `enabledOrgIdHashes`, and no dual-read, migration, or rollback handling was added. |
| `docs/fallback.md` | §1 — no negative tests asserting removed behaviour stays removed | Pass | The two new cases assert live product behaviour — that guidance requires a mount — not that deleted code stays deleted. |
| `docs/deployment-compatibility.md` | Backend change safe across old/new frontend and old/new runner | Pass | `additionalVolumes` is a pre-existing run-contract field already honoured by deployed runners; the queued path now populates it through the same mapping the direct path uses. |
| `docs/externally-managed-references.md` | Validate representation and resolve existence separately | Pass | Syntax (`parseUserPresentationTemplateId`) and existence (`authorizedUserPresentationTemplateIds`) stay distinct, with `Malformed presentation template` and `Presentation template not found` as separate outcomes. |
| `docs/externally-managed-references.md` | Security fails closed — an unresolved reference grants nothing | Pass | Both the mount list and the guidance now derive from the same authorised-id list, so an id that fails to resolve grants neither. |
| `docs/externally-managed-references.md` | A resolution miss is a normal domain result, not a 500 | Pass | Misses return `400` on send and an empty template prompt at dispatch; DB failures still propagate. |
| `docs/testing.md` | Mostly integration; mock only at the system boundary | Pass | No `vi.mock()` of internal modules, real database, no fake timers; the API-level rejection case is unchanged and still passes. |
| `docs/testing.md` | `feat:` commits must have integration tests | Pass | API-level integration coverage for the rejection path, plus new cases pinning the mount-implies-guidance rule that produced the P1. The residual end-to-end gap is declared and scoped below. |
| `docs/testing/api-testing.md` | Construct and verify through API endpoints, never DB rows or services | Pass | No test in the diff writes or reads a database row directly. |
| `docs/testing/api-testing.md` | Uniquely owned identities so the test is correct under concurrency | Pass | Fixed literal UUIDs are used only in pure prompt-builder cases that touch no shared row; the API case still uses `bdd.user()` and a fresh `randomUUID()`. |
| `.claude/skills/ccstate/SKILL.md` | No `detach()`/`Reason.DomCallback` in the signals directory; commands own their abort signal | Pass | `resolveQueuedMessageTemplateContext` is a plain async helper called from `buildCreateQueuedChatRunInput`, which passes its signal down to `loadQueuedRunMaterial` exactly as before; no floating promise is introduced. |

### Findings

None.

None. Both findings from the `767b6426` review are addressed:

- **Guidance without a mount** — fixed at the shared chokepoint rather than per call site. `buildPresentationGenerationTemplatePrompt` rejects a private id absent from `mountedUserPresentationTemplateIds`, and the default is an empty set, so a caller that forgets to declare its mounts loses the guidance instead of emitting a dangling path. `resolveThreadGenerationTemplatePrompt` makes the field required, which forces every existing caller to state what its run carries.
- **Untested volume assembly** — partially closed. Two new cases pin the rule that produced the bug: an unmounted private id resolves `invalid` with `Presentation template not found`, and one unmountable selection among several invalidates the message rather than silently dropping a template the user attached on purpose.

### Testing

- **Coverage:** Adequate for merge, with one declared residual. The invariant that caused the P1 is now tested, and every dispatch path routes through it. What remains uncovered is the volume list's own trip into the run body — publish a package, select it, claim the dispatched run, assert its storage mounts. The PR now records a verified reason rather than a general one: `installS3Fixture` replaces `context.mocks.s3.send` wholesale and would have to be reconciled with the S3 behaviour `chat-events.bdd.test.ts` already relies on for agent storage writes. That reconciliation is its own change. Since `PresentationTemplates` is off for every org, nothing user-facing depends on it today, but it should land before the switch is turned on for anyone.
- **Conventions:** Compliant. The prompt-builder cases extend a pre-existing file whose purpose is pinning generated prompt text; the user-visible route behaviour remains covered in `chat-events.bdd.test.ts`.
- **Verification:** `pnpm turbo run lint --concurrency=1` 13/13 with 0 errors, `pnpm check-types` 14/14, `pnpm knip` exit 0, `prettier --check .` clean. `generation-template-prompt.test.ts` 20/20, `presentation-template-publish.test.ts` 6/6, CHAT-02 generation-template suite 17 passed. Seven BDD cases fail in the sandbox with `Job not found in queue` at `claimRunnerJob`; six of them sit in the queued path this commit modifies, so they were re-run on unmodified `main` — the same seven fail identically there. Environment, not a regression.
- **Non-blocking observation:** the new ownership lookup in `resolveQueuedMessageTemplateContext` is not wrapped in `measureChatCallbackPreCreateTiming`, so it does not appear in the dispatch timing breakdown. Several sibling steps in that function are likewise unwrapped at this level, so this matches local practice rather than departing from it.

PR #28541, opened 2026-08-21
fix(nbd-cow): prevent stale disconnects after tid reuse

### LINKED ISSUE #28523: security(rust): reject recycled TIDs as NBD ownership proof
## Problem and evidence

NBD live-object cleanup authorizes a destructive disconnect solely by comparing `/sys/block/nbdN/pid` with the numeric `connect_tid` saved at connect time.

`NbdCowDevice` intentionally retains that TID after the Tokio blocking worker may exit. `device_ownership` classifies any later equal number as `DeviceOwnership::Ours`, and the owned cleanup helpers then issue `NBD_CMD_DISCONNECT` by device index. Normal shutdown, ambiguous-connect and cancellation cleanup, size-retry cleanup fallbacks, and synchronous `Drop` paths depend on this authority.

Linux can reuse a TID after its original task exits. If the original kernel connection is removed and the NBD index is recycled to a replacement connection whose setup task receives the same numeric TID, equality no longer proves that the stale object owns the current connection. The cooperative `DeviceLease` prevents this among processes sharing one stable lock domain, but it is not kernel connection identity and does not constrain non-cooperating or different-domain callers.

PR #8064 introduced saved-TID equality after 66 leaked devices were observed when connect workers exited. Reverting to `/proc` liveness would recreate that leak; cleanup must retain non-reusable evidence that survives worker exit.

## Impact

On a shared runner host, delayed shutdown, cancellation cleanup, or `Drop` cleanup can disconnect an NBD device now used by another run. The triggering sequence is uncommon, but the consequence is cross-run storage disruption and possible integrity loss.

## Decision

Persist a fresh opaque UUID for each NBD connection generation using Linux `NBD_ATTR_BACKEND_IDENTIFIER`, retain it in every success or ambiguous-after-send outcome, and compare it with `/sys/block/nbdN/backend` before any live-object owned disconnect.

- Exact UUID match authorizes cleanup.
- Empty, missing, unreadable, or mismatched evidence fails closed.
- The UUID is generation identity, not a secret or reusable backing path.
- Apply the same rule to normal and pooled shutdown, size retry, cancellation and ambiguous cleanup, and synchronous drop fallbacks.
- Keep the cooperative lease held through pooled check/disconnect operations.
- Keep TID-based orphan GC out of scope; it has no live-object connection metadata and follows a separate dead-owner policy.

Linux has supported the backend-identifier interface since v5.14. The runner host baseline is Ubuntu 22.04+; unsupported kernels must fail connect rather than fall back to reusable TID authority. The privileged metal integration gate must validate the deployed kernel interface.

## Alternatives considered

- **Dedicated live thread per device:** preserves TID uniqueness but adds resident thread and teardown costs per active device, scaling poorly toward 4096 configured NBD indices.
- **Cooperative lease only:** does not identify non-cooperating or different-lock-domain kernel owners.
- **TID plus `/proc` start time or pidfd:** cannot identify the kernel's stored number after both the original and replacement setup tasks have exited.
- **Kernel conditional disconnect:** would atomically eliminate the remaining check/action race, but requires an upstream ABI and host-kernel rollout outside this repository PR.

## Risks and mitigations

- The sysfs comparison followed by index-only disconnect remains non-atomic because the upstream ABI has no identity predicate. Retain the cooperative lease through the blocking critical section to narrow the window among cooperating processes.
- Old runners retain the previous behavior until they drain. New cleanup treats old empty-backend connections as foreign, and rollback remains compatible because old binaries ignore the backend value.
- Do not log observed foreign backend strings; other NBD clients may use non-opaque identifiers.
- Create the UUID before the connect send boundary and include it in ambiguous-after-send state so cancellation cannot lose identity evidence.

## Acceptance criteria

- No live-object cleanup authorizes disconnect solely from numeric TID equality.
- Every successful or ambiguous connect that may have reached the kernel retains its UUID generation identity.
- Unknown and foreign identity evidence remains fail-closed.
- Normal and pooled shutdown, size retry, cancellation or ambiguous-connect cleanup, and synchronous `Drop` cleanup use the same ownership rule.
- Focused coverage recycles a device index to a replacement connection created from the same numeric TID but with a different UUID and proves stale cleanup does not disconnect it.
- Coverage preserves successful cleanup for the original matching UUID.
- Rust formatting, Clippy, documentation, crate tests, and privileged metal integration tests pass.

seven332:
## Summary

- attach a UUID v4 backend identifier to every NBD connect command
- require an exact `/sys/block/nbdN/backend` match before any live-object cleanup disconnects an index
- route size-retry cleanup through the same ownership check and remove the unconditional disconnect path
- add a root-only metal regression that reconnects the same index from the same blocking-worker TID and verifies stale cleanup leaves the replacement usable

## Scope and risks

- keep sysfs PID/TID handling only for orphan discovery and GC
- rely on `NBD_ATTR_BACKEND_IDENTIFIER`, available on the supported Ubuntu 22.04+ runner baseline; older kernels fail the connect instead of falling back to unsafe TID matching
- retain the kernel API's residual sysfs-read-to-index-disconnect race because NBD does not provide an atomic conditional disconnect operation

## Validation

- `cargo test --manifest-path crates/Cargo.toml -p nbd-cow`
- repository pre-commit Rust formatting, documentation, and Clippy hooks

The root-only NBD regression was compiled locally and is executed by the existing serial metal integration-test job; the local container has no loaded NBD module and is not root.

Closes #28523

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -2072,6 +2072,7 @@ dependencies = [
  "tracing",
  "tracing-subscriber",
  "tracing-test-support",
+ "uuid",
  "which",
 ]
 
--- a/crates/nbd-cow/Cargo.toml
+++ b/crates/nbd-cow/Cargo.toml
@@ -21,6 +21,7 @@ thiserror = { workspace = true }
 tokio = { workspace = true, features = ["full"] }
 tokio-util = { workspace = true }
 tracing = { workspace = true }
+uuid = { workspace = true, features = ["v4"] }
 which = { workspace = true, optional = true }
 
 [dev-dependencies]
--- a/crates/nbd-cow/src/device/connection.rs
+++ b/crates/nbd-cow/src/device/connection.rs
@@ -1,5 +1,7 @@
 use std::time::{Duration, Instant};
 
+use uuid::Uuid;
+
 use crate::error::{self, Result};
 use crate::{netlink, pool};
 
@@ -148,9 +150,11 @@ impl Drop for ConnectDeviceOutcome {
             return;
         };
 
-        let connect_tid = match result {
-            Ok(success) => success.connect_tid,
-            Err(netlink::ConnectDeviceError::AmbiguousAfterSend { connect_tid, .. }) => connect_tid,
+        let connection_id = match result {
+            Ok(success) => success.connection_id,
+            Err(netlink::ConnectDeviceError::AmbiguousAfterSend { connection_id, .. }) => {
+                connection_id
+            }
             Err(
                 netlink::ConnectDeviceError::NotSent { .. }
                 | netlink::ConnectDeviceError::DefiniteAfterSend { .. },
@@ -159,12 +163,11 @@ impl Drop for ConnectDeviceOutcome {
 
         tracing::warn!(
             device_index = self.device_index,
-            connect_tid,
             "NBD connect result dropped before observation; disconnecting owned device"
         );
         disconnect_connected_if_owned(ConnectedDevice {
             index: self.device_index,
-            connect_tid,
+            connection_id,
         });
     }
 }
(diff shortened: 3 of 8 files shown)


seven332:
## Code Review: PR #28541

**Reviewed head:** `0762c956e2d4811f2a1a95cdac0e5663a488e569`

### Summary

The PR replaces reusable connecting-thread TIDs with kernel-stored per-connect UUID identities for every live-object cleanup path, removes the remaining unconditional retry disconnect, and adds a real NBD regression for same-TID index reuse.

### Findings

No findings.

### Testing

Local fmt, Clippy, rustdoc, crate tests, and repository pre-commit Rust hooks passed. The new root-only NBD regression compiled locally; execution is delegated to the existing serial metal CI job because the local container is non-root and has no NBD module loaded.

### Verdict

LGTM

seven332:
## Code Review: PR #28541

**Reviewed head:** `370094452e432d0d893f450bbc238a9b71039e99`

### Summary

The PR uses a kernel-stored per-connect UUID for every live-object ownership check and removes the unsafe unconditional disconnect path. The follow-up commit keeps the same-TID regression strict while accounting for asynchronous kernel teardown and architecture-specific first-I/O setup.

### Findings

No findings.

### Testing

Local fmt, Clippy, rustdoc, crate tests, and repository pre-commit Rust hooks passed. The first metal run exposed test-fixture issues on both architectures; those are fixed in this head. Corrected x86_64 and arm64 metal execution is pending CI.

### Verdict

LGTM

seven332:
## Code Review: PR #28541

**Reviewed head:** `054e397d716523029a2f5e98f298045f931c28ad`

### Summary

The PR uses a kernel-stored per-connect UUID for live NBD ownership checks and removes the unsafe unconditional disconnect path. The latest commit fixes the regression test's single-worker deadlock by moving block-device I/O to a dedicated thread while preserving deterministic TID reuse and all replacement-connection assertions.

### Findings

No findings.

### Testing

Local fmt, Clippy, rustdoc, crate tests, and repository pre-commit Rust hooks passed. The root-only NBD regression compiled locally; corrected x86_64 and arm64 metal execution is pending CI because the local container is non-root and has no NBD module loaded.

### Verdict

LGTM

PR #28439, opened 2026-08-20
feat(platform): turn the home invite button into a growth entry

Lunarivibe:
## What

Replace the home page's admin-only `Invite people` button with one scalable growth entry:

- Before Slack is installed: `[Slack] Add Okou in Slack | chevron`
- After Slack is installed: `[+] Invite member | chevron`
- The menu carries Slack, other channels, member invitation, and the current credit balance

Slack is a one-time setup, while inviting remains useful indefinitely, so two trigger states cover the complete lifecycle.

## Behavior

- Keep the entry admin-only and omit the entire header for non-admins.
- Route Slack and other channel actions to `/works`.
- Open invitation and usage as settings dialogs without breaking dropdown focus handling.
- Fetch credit data only after the menu opens, preserving the existing account-menu request behavior.
- Include usage-pack credits only when `UsagePackPlans` is enabled.
- Replace `chat.agentPage.invitePeople` with `chat.agentPage.growth.*` across all 10 locales.

## Design

- Align the trigger to the same `max-w-[900px]` column as the home composer.
- Use the account dropdown's row density: `gap-3 px-3 py-2.5` and 16px marks.
- Use the dropdown component's 12px radius, gray-400 border, and full-bleed 0.7px separators.
- Keep the trigger at the standard 32px small-button height.
- Deliberately match the trigger's 12px radius to its menu and use the existing card shadow token.

The global `DropdownMenuContent` shadow and a larger two-line Slack action remain out of scope.

## Testing

- Added page-level coverage for the pre-install Slack state and `/works` navigation.
- Added page-level coverage for the post-install invite state and People settings navigation.
- Added page-level coverage for credit balance and Usage settings navigation.
- `git diff --check`
- Parsed all 10 changed locale JSON files.

Local dependency installation is unavailable in this environment, so formatting, lint, type-checking, and the targeted app tests are verified by the PR pipeline.

PR #28563, opened 2026-08-21
fix(platform): put the growth entry back in the top-right corner

Lunarivibe:
Follow-up to #28439.

## Position

That PR aligned the home entry to the 900px content column, which moved it inward from the top-right corner the invite button had always held. The corner is where people look for it, so put it back: `InviteHeader` and the growth entry now share one `CornerHeader` — a full-width row, `justify-end` — so the two feature-switch states cannot drift apart again, and the menu's `align="end"` is drawn against the page edge as before.

Verified on the preview at 1440px: the entry's right edge is 1416px, identical to the legacy invite button, and 96px outboard of the content column's 1320px.

## Copy

"Invite member" named the mechanic, not the point. The corner button and the menu row read from the same key, so both now say **Invite humans 🤝** — the plural is the product (a workspace you bring people into) and "humans" is the distinction that matters next to an agent. Translated across all 10 locales.

Measured against the 206px the menu row gives a label: the widest translation is French at 152px, so nothing truncates and the emoji survives everywhere.

This is the first emoji in platform copy; nothing else in `i18n/locales` carries one today.

## Verification

Walked through on `https://pr-28563-app.omby.ai`: default state, growth entry on, menu open, and the Slack-installed variant that rotates the corner to the invite label. No horizontal overflow down to the `md` breakpoint. `home-growth-entry.test.tsx` covers the label and both switch states, and is updated for the new string.

### DIFF
--- a/turbo/apps/platform/src/i18n/locales/de-DE/common.json
+++ b/turbo/apps/platform/src/i18n/locales/de-DE/common.json
@@ -1505,7 +1505,7 @@
         "connect": "Verbinden",
         "credits": "Guthaben",
         "inSlack": "{{assistantName}} ist in Slack",
-        "inviteMember": "Menschen zur Zusammenarbeit einladen 🤝",
+        "inviteMember": "Menschen einladen 🤝",
         "otherChannels": "Telegram und Telefon"
       },
       "invitePeople": "Personen einladen",
--- a/turbo/apps/platform/src/i18n/locales/de-DE/common.json
+++ b/turbo/apps/platform/src/i18n/locales/de-DE/common.json
@@ -1505,7 +1505,7 @@
         "connect": "Verbinden",
         "credits": "Guthaben",
         "inSlack": "{{assistantName}} ist in Slack",
-        "inviteMember": "Mitglied einladen",
+        "inviteMember": "Teammitglieder einladen",
         "otherChannels": "Telegram und Telefon"
       },
       "invitePeople": "Personen einladen",
--- a/turbo/apps/platform/src/i18n/locales/de-DE/common.json
+++ b/turbo/apps/platform/src/i18n/locales/de-DE/common.json
@@ -1505,7 +1505,7 @@
         "connect": "Verbinden",
         "credits": "Guthaben",
         "inSlack": "{{assistantName}} ist in Slack",
-        "inviteMember": "Teammitglieder einladen",
+        "inviteMember": "Menschen zur Zusammenarbeit einladen 🤝",
         "otherChannels": "Telegram und Telefon"
       },
       "invitePeople": "Personen einladen",
--- a/turbo/apps/platform/src/i18n/locales/en-US/common.json
+++ b/turbo/apps/platform/src/i18n/locales/en-US/common.json
@@ -1505,7 +1505,7 @@
         "connect": "Connect",
         "credits": "Credits",
         "inSlack": "{{assistantName}} is in Slack",
-        "inviteMember": "Invite member",
+        "inviteMember": "Invite teammates",
         "otherChannels": "Telegram and phone"
       },
       "invitePeople": "Invite people",
--- a/turbo/apps/platform/src/i18n/locales/en-US/common.json
+++ b/turbo/apps/platform/src/i18n/locales/en-US/common.json
@@ -1505,7 +1505,7 @@
         "connect": "Connect",
         "credits": "Credits",
         "inSlack": "{{assistantName}} is in Slack",
-        "inviteMember": "Invite human to collaborate 🤝",
+        "inviteMember": "Invite humans 🤝",
         "otherChannels": "Telegram and phone"
       },
       "invitePeople": "Invite people",
--- a/turbo/apps/platform/src/i18n/locales/en-US/common.json
+++ b/turbo/apps/platform/src/i18n/locales/en-US/common.json
@@ -1505,7 +1505,7 @@
         "connect": "Connect",
         "credits": "Credits",
         "inSlack": "{{assistantName}} is in Slack",
-        "inviteMember": "Invite teammates",
+        "inviteMember": "Invite human to collaborate 🤝",
         "otherChannels": "Telegram and phone"
       },
       "invitePeople": "Invite people",
--- a/turbo/apps/platform/src/i18n/locales/es-ES/common.json
+++ b/turbo/apps/platform/src/i18n/locales/es-ES/common.json
@@ -1529,7 +1529,7 @@
         "connect": "Conectar",
         "credits": "Créditos",
         "inSlack": "{{assistantName}} está en Slack",
-        "inviteMember": "Invitar a un miembro",
+        "inviteMember": "Invitar a compañeros",
         "otherChannels": "Telegram y teléfono"
       },
       "invitePeople": "Invitar a personas",
--- a/turbo/apps/platform/src/i18n/locales/es-ES/common.json
+++ b/turbo/apps/platform/src/i18n/locales/es-ES/common.json
@@ -1529,7 +1529,7 @@
         "connect": "Conectar",
         "credits": "Créditos",
         "inSlack": "{{assistantName}} está en Slack",
-        "inviteMember": "Invitar a compañeros",
+        "inviteMember": "Invita a un humano a colaborar 🤝",
         "otherChannels": "Telegram y teléfono"
       },
       "invitePeople": "Invitar a personas",
--- a/turbo/apps/platform/src/i18n/locales/es-ES/common.json
+++ b/turbo/apps/platform/src/i18n/locales/es-ES/common.json
@@ -1529,7 +1529,7 @@
         "connect": "Conectar",
         "credits": "Créditos",
         "inSlack": "{{assistantName}} está en Slack",
-        "inviteMember": "Invita a un humano a colaborar 🤝",
+        "inviteMember": "Invitar humanos 🤝",
         "otherChannels": "Telegram y teléfono"
       },
       "invitePeople": "Invitar a personas",
--- a/turbo/apps/platform/src/i18n/locales/fr-FR/common.json
+++ b/turbo/apps/platform/src/i18n/locales/fr-FR/common.json
@@ -1529,7 +1529,7 @@
         "connect": "Connecter",
         "credits": "Crédits",
         "inSlack": "{{assistantName}} est dans Slack",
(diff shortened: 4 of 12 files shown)


Lunarivibe:
LGTM


PR_AUTO_REVIEW_COMMIT: fbc42e59ac077d4941210e53b2a4ea6e2d8f7b08
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Follow-up to #28439. The home growth entry was aligned to the 900px content column, which pulled it inward from the top-right corner the invite button had always occupied; this PR extracts a shared `CornerHeader` so both feature-switch states render in one full-width `justify-end` row and land on the same corner. It also renames the `growth.inviteMember` label to `Invite humans 🤝` across all 10 locales — the same key backs both the corner trigger and the menu row, so the two cannot disagree.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no wrong assumptions or unhandled paths | Pass | `zero-growth-entry.tsx`: `CornerHeader` reproduces `InviteHeader`'s original `<header>`/`<div>` classes byte-for-byte; only `EnabledGrowthEntryHeader` loses `mx-auto w-full max-w-[900px]`. The `slackInstalled === null` early return and the non-admin `null` return are untouched. |
| `REVIEW.md` | Security: injection, secrets, missing auth | N/A | Diff touches only presentational JSX and i18n string constants; no network call, credential, route, or auth path. |
| `REVIEW.md` | Type safety: no `any`, no broad types | Pass | Only new type is `{ children: ReactNode }` via `import type { ReactNode } from "react"`, matching existing platform usage (`views/clerk/clerk-provider.tsx`, `views/workflows-page/workflows-page.tsx`). |
| `REVIEW.md` | Style and maintainability; no duplicated logic; comments explain WHY | Pass | The refactor removes the duplicated header shell rather than adding one; the `CornerHeader` doc comment states why the row spans the page instead of restating what the classes do. |
| `REVIEW.md` | `fix:` commit carries a regression test | Pass | The copy change is pinned in `home-growth-entry.test.tsx:134` for the Slack-installed lead state, and the Slack-not-installed lead state stays pinned at line 109, so the `leadIsSlack` branch keeps both of its labels asserted. |
| `REVIEW.md` | Same rule applied to the layout half of the fix | Pass | Not observable at the sanctioned boundary: jsdom performs no layout, and `docs/testing/testing-external-behavior.md` prohibits asserting CSS classes. Verified instead on the deployed preview and recorded in the PR body (entry right edge 1416px, identical to the legacy invite button; content column right edge 1320px). See note 3. |
| `REVIEW.md` | Feature regression guard: existing behavior changed only with coverage and containment | Pass | Both switch states keep their tests — switch off asserts `invite-button` / `Invite people` (line 79-91), switch on asserts `growth-entry` (line 104-147), non-admin asserts neither renders (line 93-102). The change stays inside the existing `FeatureSwitchKey.HomeGrowthEntry` gate; no new surface is exposed. |
| `REVIEW.md` | DB / JSONB review gate | N/A | No schema, contract, Zod response, or `jsonb().$type<>()` change; no file under `packages/db` or `packages/api-contracts`. |
| `docs/bad-smell.md` | Zero lint/type suppression comments | Pass | No `eslint-disable`, `@ts-ignore`, `@ts-expect-error`, or `prettier-ignore` added; `lint-eslint`, `lint-types`, `lint-types-apps`, `lint-format`, and `lint-knip` all pass on this head. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | The one added import is a static type-only import at file top. |
| `docs/bad-smell.md` | No hardcoded URLs or environment-specific values | N/A | No configuration or URL surface in the diff. |
| `docs/bad-smell.md` | Fail fast; no fallback or recovery slop | Pass | No `??`/`||` chain, no try/catch, no defaulting added. |
| `docs/fallback.md` | No new fallback; every introduced fallback declared in the PR summary | N/A | The diff introduces no compatibility branch, tolerance path, or contract-guarded default, so no `Fallbacks` section is required. `InviteHeader` is the pre-existing switch-off branch and is only relocated, not introduced. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | No test added; the one changed assertion updates an existing behavior to its new value rather than pinning a deletion. |
| `docs/testing.md` | Integration/page-level tests; no unit tests for internal functions | Pass | `home-growth-entry.test.tsx` remains a page test; the new `CornerHeader` gets no component-only test. |
| `docs/testing.md` | Mock at the boundary; no internal `vi.mock`, no `globalThis.services.db` mock, no fetch stubbing, no fake timers | Pass | The diff adds no mock of any kind; existing setup uses `context.mocks.api` against the Slack and billing contracts. |
| `docs/testing/app-testing.md` | Page tests enter through `detachedSetupPage` and assert page-visible behavior; mocks configured through `context.mocks` before setup | Pass | `setupGrowthEntry()` configures `context.mocks.data.org`, `context.mocks.api(integrationsSlackContract.getStatus)` and feature switches before `detachedSetupPage`; assertions read `screen`/`within(menu)`. |
| `docs/testing/testing-external-behavior.md` | Give the system context the way a user can, observe from the surface a user can see; never assert CSS classes or internal state | Pass | Every assertion goes through rendered text, testids, roles, and `search()`; the changed line asserts the visible label. This rule is also what makes the layout half of the fix unassertable — see note 3. |
| `docs/testing/anti-patterns.md` | AP-2/AP-4/AP-5/AP-10: no fetch stubbing, no internal mocks, no fake timers, no direct component rendering | Pass | None present before or after the diff. |
| `docs/testing/anti-patterns.md` | AP-7/AP-8: do not test implementation details or exact UI text | Pass | The label assertion is not copy-pinning for its own sake — the label is the only user-visible difference between the `leadIsSlack` and invite lead states, so it is the sole surface at which that branch is observable. See note 1. |
| `docs/testing/patterns.md` | Pattern 7: platform page tests live under `views/**/__tests__/`, enter via `detachedSetupPage`, and prefer page tests over signal tests | Pass | File location and entry path unchanged and conforming. |
| `docs/effect.md` | React render stays pure: no Store writes, I/O, resource allocation, or signal creation during render | Pass | `CornerHeader` reads one prop and returns elements. No hook, command, effect, or ref added anywhere in the diff. |
| `docs/effect.md` | Choose the owner and trigger before the API | Pass | No new work is introduced; the existing `openInvite`/`openCredits` commands and their `pageSignal$` ownership are untouched. |
| `docs/cache.md` | RC-1/RC-2: no render-phase side effects, no render-time signal allocation | Pass | `CornerHeader` is declared at module scope, not inside another component's render, so it introduces no per-render component identity and no signal factory call. |
| `docs/cache.md` | RC-3/RC-4: no unbounded lifetime caches, no owner/lifetime mismatch | N/A | The diff adds no cache, module-level `Map`/`Set`, or scoped state. |
| `docs/cache.md` | RC-7/RC-8: stable callback refs; symmetric resource lifecycles | N/A | No ref, listener, timer, or observer in the diff. |
| `.claude/skills/ccstate/SKILL.md` | DOM callbacks use `detach(..., Reason.DomCallback)` with `pageSignal$` from `useGet` | Pass | `useGrowthActions` and `InviteButton` keep the existing `detach(openSettings("people", pageSignal), Reason.DomCallback)` form; the refactor does not move or rewrap them. |
| `.claude/skills/ccstate/SKILL.md` | `accept` for all `zeroClient$` calls; loadables for view lifecycle state | Pass | Unchanged: `useLastLoadable(slackOrgData$)` and `useLastLoadable(billingStatusAsync$)` still drive the render, with no view-layer error handling added. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

#### Notes (non-blocking)

1. `home-growth-entry.test.tsx:134` — the assertion now carries an emoji, so a pure copy edit breaks the test. This is pre-existing style in this file (`Add Zero in Slack`, `Telegram and phone`, `Invite people`) and the label is genuinely the only observable that separates the two lead states, so it is not a new violation of `docs/testing/anti-patterns.md` AP-8. Worth knowing that copy churn on this key now has a test cost.
2. `i18n/locales/hi-IN/common.json` — `इंसानों को बुलाएँ` uses `बुलाएँ` (call / summon) where every neighbouring string in this namespace uses `आमंत्रित करें` (invite), including `invitePeople` (`लोगों को आमंत्रित करें`) which renders in the same corner when the switch is off. The verb register is inconsistent for one action; `इंसानों को आमंत्रित करें` would match.
3. `zero-growth-entry.tsx:309-320` — the corner position now has no automated guard. `docs/testing/testing-external-behavior.md` forbids the only mechanism available in jsdom (a class assertion), so I am not requesting one; the `CornerHeader` doc comment is the standing explanation. The residual risk is that a future edit can reintroduce a column constraint on this row with no test failing.
4. `i18n/locales/*/common.json` — first emoji in platform copy; nothing else under `i18n/locales` carries one. Two consequences worth accepting knowingly: the accessible name of the trigger and the menu row now ends in the emoji (announced as "handshake"), and glyph rendering varies by platform font stack. Neither is a defect, but both are new.
5. `i18n/locales/ko-KR/common.json` — `inviteMember` (`사람 초대 🤝`) is now the same phrase as `invitePeople` (`사람 초대`) modulo the emoji. The two never render at the same time (they belong to opposite sides of the `HomeGrowthEntry` switch), so there is no in-product collision.

### Testing

- **Coverage: Adequate.** The copy change is regression-covered in both lead states (`home-growth-entry.test.tsx:109` and `:134`), and the switch-off and non-admin paths keep their existing assertions, so the feature regression guard is satisfied for every branch this diff can reach. The layout half of the fix is not coverable at the platform test boundary — jsdom does not lay out, and asserting the className is prohibited by `docs/testing/testing-external-behavior.md` — so it was verified on the deployed preview instead, with the measurement recorded in the PR body (right edge 1416px matching the legacy invite button, versus the 1320px content column edge). Recorded in note 3 as residual risk rather than a missing required test.
- **Conventions: Compliant.** No mocks added, no internal `vi.mock`, no fetch stubbing, no fake timers, no direct component rendering, no `afterEach` cleanup hook. Entry stays through `detachedSetupPage` with overrides on `context.mocks`.
- **CI on `fbc42e5`:** `lint-eslint`, `lint-format`, `lint-knip`, `lint-types`, `lint-types-apps`, `test-app (1-8)`, and `test-api (1-8)` all pass. Only CodeQL `Analyze (javascript-typescript)` and `deploy-runner-start` are still running, neither of which this diff can affect.

Issue #27787, opened 2026-08-17
bug: [Composer] Safari re-renders composer after selecting feedback, sends stale content


bingjiez666:
## Summary\n\nOn **Safari**, after the user selects a feedback option (e.g. a negative-feedback reaction on an outgoing message), an input event triggers a full composer re-render. From that point the composer being edited is not the same instance that sends the message, so the text actually sent is stale or truncated content — not what the user sees and/or edited in the composer.\n\nRelated observed symptoms:\n- PWA: a feedback message was sent, but only two characters arrived.\n- Content shown in the composer/editor does not match what is actually submitted.\n\n## Root cause (initial analysis)\n\n- After selecting feedback, an input event causes the composer to re-render.\n- The re-render replaces the composer instance, so subsequent edits target a composer that is no longer the one bound to the send action, and the submitted payload comes from the leftover/stale instance.\n\n## Expected behavior\n\nSelecting a feedback option must not invalidate or replace the composer instance. The message that is sent should always be exactly what is displayed and edited in the composer, in all browsers.\n\n## Steps to reproduce\n\n1. Open the app in Safari (also reproducible on PWA).\n2. Send a message, then select the feedback option on it.\n3. Edit the message/composer content (input event).\n4. Send — observe the submitted content differs from the edited content.\n\n## Environment\n\n- Browser: Safari (also affects PWA install)\n- Platform: vm0 web/SPA\n\n## Impact\n\nUser edits are silently lost or garbled (e.g. only two characters sent), which can be confusing and is data-loss adjacent for composed messages.

PR #28574, opened 2026-08-21
fix(composer): reconcile live dom before submission

bingjiez666:
## Summary

- add the Bingjie-only `composerSubmitDomReconcile` feature switch
- before serializing a settled composer submission, create fresh whole-block DOM mutations and synchronously flush ProseMirror so text stranded only in the live contenteditable is parsed back into `editor.state.doc`
- target the feedback note `contentDOM` directly so the custom node view does not ignore the reconciliation marker
- report only before/after prompt lengths and feedback presence when reconciliation changes a submission; no composer text is included in telemetry
- cover both the enabled recovery path and the disabled-switch behavior through the rendered chat page

This does not depend on an old mutation record still being pending. It creates a new record at submission time, which makes the recovery work after the original record has already been consumed or lost.

Related to #27787.

## Rollout

- default: disabled
- `bingjie@vm0.ai`: enabled

## Verification

- `pnpm exec prettier --check <changed files>`
- `pnpm knip`
- `pnpm --filter @okouai/app exec vitest run src/views/zero-page/__tests__/chat-inline-feedback.test.tsx` (32 passed)
- `pnpm --filter @okouai/core exec vitest run src/__tests__/feature-switch.test.ts` (25 passed)
- negative control: with the fresh reconciliation mutations temporarily removed, the new page test failed and submitted only the document prefix while the full text remained visible in the DOM

### DIFF
--- a/turbo/apps/platform/src/signals/external/feature-switch.ts
+++ b/turbo/apps/platform/src/signals/external/feature-switch.ts
@@ -67,6 +67,12 @@ export const imageModelSelectionEnabled$ = computed((get): boolean => {
   return get(featureSwitch$)[FeatureSwitchKey.ImageModelSelection] ?? false;
 });
 
+export const composerSubmitDomReconcileEnabled$ = computed((get): boolean => {
+  return (
+    get(featureSwitch$)[FeatureSwitchKey.ComposerSubmitDomReconcile] ?? false
+  );
+});
+
 export const codexFastModeEnabled$ = computed((get): boolean => {
   return get(featureSwitch$)[FeatureSwitchKey.CodexFastMode] ?? false;
 });
--- a/turbo/apps/platform/src/signals/zero-page/tiptap-workflow-composer.ts
+++ b/turbo/apps/platform/src/signals/zero-page/tiptap-workflow-composer.ts
@@ -16,7 +16,12 @@ import {
   type EditorState,
   type Transaction,
 } from "@tiptap/pm/state";
-import { Decoration, DecorationSet, type NodeView } from "@tiptap/pm/view";
+import {
+  Decoration,
+  DecorationSet,
+  type EditorView,
+  type NodeView,
+} from "@tiptap/pm/view";
 import { StarterKit } from "@tiptap/starter-kit";
 import { createCompositionGate, type CompositionGate } from "@okouai/ui";
 import {
@@ -27,6 +32,9 @@ import {
 import type { WorkflowSummary } from "@okouai/api-contracts/contracts/workflows";
 import { agents$ } from "../agent.ts";
 import { currentChatAgentRecordId$ } from "../agent-chat.ts";
+import { composerSubmitDomReconcileEnabled$ } from "../external/feature-switch.ts";
+import { logger } from "../log.ts";
+import { sentryLogContext } from "../../lib/sentry-config.ts";
 import { onRef, resetSignal } from "../utils.ts";
 import type { DraftInputSyncTarget, DraftSignals } from "./chat-draft.ts";
 import {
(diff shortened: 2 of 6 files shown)


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: f9e154b078895a8b9a1149d446e78f209a6cfc9b
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

The change is correctly contained behind the default-off, Bingjie-only `composerSubmitDomReconcile` switch. On enabled submissions it waits for IME composition to settle, forces any deferred ProseMirror observer flush, creates fresh whole-block attribute mutations, and synchronously parses live DOM text back into the editor document before both prompt and structured-message serialization. The implementation avoids message-content telemetry and preserves the disabled path.

### Review Criteria

| --- | --- | --- |
| `REVIEW.md` | Pass | Reviewed the complete six-file diff and every changed file at head `f9e154b`; the `fix:` commit includes regression coverage, and the switch preserves the existing submission path when disabled. |
| `docs/docs.md` | Pass | Loaded the platform, ccstate, test, fallback, and deployment guidance selected by the affected surfaces. |
| `docs/bad-smell.md` | Pass | No `any`, suppression directive, dynamic import, secret, duplicated implementation, or silent configuration recovery was introduced; the internal observer cast is narrow and fails loudly if the implementation disappears. |
| `docs/fallback.md` | Pass | This is a non-GA, default-off rollout with no dual-read, dual-write, legacy protocol, or old-shape compatibility branch. The `?? false` switch read is the existing default-disabled bootstrap convention, not a version-migration fallback. |
| `docs/testing.md` | Pass | The fix has positive and negative-control page coverage for the exact stranded-DOM failure boundary and validates both prompt and structured `userMessage` output. |
| `docs/effect.md` | Pass | Reconciliation is executed at the submission command boundary after the composition gate settles; no React effect or hidden synchronization loop was added. |
| `docs/cache.md` | Pass | No new cache or independently owned lifecycle was introduced; the computed flag reads the existing feature-switch cache and the editor owns its existing observer. |
| `.claude/skills/ccstate/SKILL.md` | Pass | The feature state is derived with `computed`, read inside the submission `command`, and adds no detached async work or signal ownership leak. |
| `docs/testing/patterns.md` | Pass | Tests use `detachedSetupPage`, the real rendered chat surface, and the real ProseMirror DOM observer while mocking only external chat lifecycle requests. |
| `docs/testing/anti-patterns.md` | Pass | Assertions target user-visible send behavior and request payloads; there are no fake timers, direct component renders, snapshot-only assertions, or mocked internal implementation responses. |
| `docs/testing/app-testing.md` | Pass | Page setup supplies switch state through the standard harness and exercises interaction through the mounted composer and Send button. |
| `docs/testing/testing-external-behavior.md` | Pass | The enabled case proves recovered ordinary and feedback text reaches the outbound prompt/document; the disabled case proves the legacy document remains unchanged. |
| `docs/deployment-compatibility.md` | Pass | No API schema, database shape, runner protocol, queue payload, service worker, or persisted message contract changes. Old clients ignore the new switch key, and the feature remains non-GA and off by default. |

### Findings

- P0: None.
- P1: None.

The synthetic attribute records map to full node ranges in ProseMirror 1.41.8's `DOMObserver.registerMutation`. Ordinary paragraphs accept the record, and feedback items target their `contentDOM`, so their custom `ignoreMutation` does not discard it. Inline node views retain their node type and attributes through ProseMirror's view-desc parse rules, while non-editable template attachment blocks are intentionally skipped.

### Testing

- Coverage: Adequate — the rendered-page regression tests cover the enabled recovery path, the disabled path, ordinary paragraph text, feedback-note text, outbound prompt content, structured `userMessage`, and recovery telemetry.
- Verification reported and previously run for this head: Prettier, full lint, Knip, staged type checks, the app test file (32 passed), and core feature-switch tests (25 passed).
- Positive control: removing the fresh reconciliation mutations makes the new recovery case submit only the stale document prefix while the full text remains visible in the DOM.
- Current GitHub checks at review time: 50 passed, 9 pending, 0 failed.
- Conventions: Compliant.

PR #28584, opened 2026-08-21
fix(runner): recover evicted websocket usage pricing

### LINKED ISSUE #28568: bug(python): avoid dropping late WebSocket output after tier-state eviction
## Summary

The model-provider WebSocket reporter drops positive output usage when a committed tier decision is evicted from its bounded per-flow state before a late same-response frame arrives.

## Location

- `crates/runner/mitm-addon/src/usage/providers/model_provider.py:791-818`
  - `_source_model_usage_pricing()` looks up the response ID in the bounded tier map.
  - When the decision is missing and an output-only frame has no input partition, `_model_usage_tier()` returns `None`, so lines 808-809 skip billable usage.
  - Lines 816-817 evict the least-recently-used decision after 100 response IDs, including committed decisions that later frames still need.
- `crates/runner/mitm-addon/tests/test_model_provider_websocket_usage_aggregation.py:181-230` verifies late output-only frames while the tier decision is retained.
- `crates/runner/mitm-addon/tests/test_model_provider_websocket_usage_aggregation.py:332-361` verifies tier-state eviction, but the two behaviors are not covered together.

## Reproduction

1. Open a model-provider Responses WebSocket for tiered model `gpt-5.5`.
2. Feed response ID `victim` with `input_tokens: 272001`. This emits `tokens.input.long_context` and commits the long-context decision.
3. Feed 100 other response IDs so `victim` is evicted from `MODEL_PROVIDER_USAGE_TIERS`.
4. Feed a late frame for `victim` containing only `output_tokens: 12`.
5. Flush usage.

Actual result:

- `tokens.input.long_context: 272001` is billed.
- The output observation is emitted, but no billable output event is emitted.
- A `model_long_context_tier_unresolved` underbilling-risk log is written.

Expected result:

- The late output is billed consistently for the original response, as `tokens.output.long_context: 12` in this scenario.

## Impact

A long-lived Responses WebSocket that crosses the 100-response tier-state bound can underbill split or duplicate usage frames for tiered models. Positive output remains visible in model-usage observations but is absent from billing, so the failure is deterministic and externally measurable.

## Acceptance criteria

- A regression test combines committed long-context input, tier-state pressure beyond the configured bound, and a late output-only frame for the evicted response ID.
- The late positive output produces a billable event in the correct tier and service-tier category.
- Missing tier state never causes positive WebSocket usage to be omitted from billing; if exact state cannot be retained, use a conservative billable fallback and retain underbilling diagnostics as appropriate.
- Per-flow memory remains bounded and tier state is still released at terminal cleanup.

## Suggested fix direction

Keep enough bounded settlement state for independently admissible late categories, or apply a conservative billable tier when an evicted response ID later carries positive usage without an input partition. Do not return `None` for a positive source solely because the exact tier decision was evicted.

seven332:
## Summary

- recover an evicted WebSocket response's exact context and service tier from the usage buffer's existing bounded source-key admission history
- bill positive output with a conservative long-context fallback, plus an uncertainty diagnostic, when neither bounded state store can classify it
- preserve the current source UUID protocol, duplicate suppression, memory bounds, and terminal flow cleanup

## Behavior

The 100-entry flow-local tier map remains the fast path. On a miss for positive output-only usage, the runner derives the possible concrete source keys and inspects the existing 10,000-key admission history under one lock. A retained input-family or output key recovers the original base/long-context and standard/fast decision. The recovered decision is put back into the flow LRU so later frames remain consistent.

If no retained key can classify the source, the runner emits a long-context billable item instead of dropping the usage. Explicit priority/fast evidence is preserved; otherwise the existing standard default is used. The `model_long_context_tier_unresolved` record now shows that conservative billing occurred.

## Scope

- no API payload or database changes
- no idempotency namespace or UUID input changes
- no new cache or increased memory bound
- no frontend or production-pricing changes

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest -q tests/test_model_provider_websocket_usage_aggregation.py -k 'late_output_recovers_evicted_long_context_fast_tier or duplicate_output_recovers_evicted_source_category or output_without_input_uses_conservative_billing_fallback'` (3 passed)
- `uv run --no-sync python -m pytest -q tests/test_model_provider_websocket_usage_aggregation.py` (24 passed)
- `uv run --no-sync python -m pytest tests/` (6,444 passed)
- `lefthook run pre-commit`

Closes #28568

### DIFF
--- a/crates/runner/mitm-addon/src/flow_metadata_keys.py
+++ b/crates/runner/mitm-addon/src/flow_metadata_keys.py
@@ -174,8 +174,11 @@ Model-provider usage
 - ``MODEL_PROVIDER_USAGE_TIERS``: bounded insertion-ordered mapping from
   WebSocket response id to the concrete billing tier decision selected from
   its input partition. Zero-only decisions remain provisional until the first
-  positive billing item. Written by model-provider billing and cleared at the
-  WebSocket terminal lifecycle boundary.
+  positive billing item. Evicted decisions can be recovered while their
+  concrete source keys remain in the separately bounded usage admission
+  history; otherwise positive usage uses a conservative tier. Written by
+  model-provider billing and cleared at the WebSocket terminal lifecycle
+  boundary.
 - ``MODEL_USAGE_PROVIDER``: optional ``str`` canonical model id from registry VM
   info. Read by model-provider usage observability and reported-model selection.
 - ``MODEL_JSON_USAGE_FINALIZED``: ``bool`` written when JSON usage finalization
--- a/crates/runner/mitm-addon/src/usage/buffer/__init__.py
+++ b/crates/runner/mitm-addon/src/usage/buffer/__init__.py
@@ -64,6 +64,7 @@ __all__ = [
     "drain_usage_events_after_executor_shutdown",
     "flush_usage_events",
     "reset_usage_buffer_for_tests",
+    "seen_source_idempotency_keys",
 ]
 
 _usage_event_buffer = UsageEventBuffer()
@@ -188,6 +189,11 @@ def flush_usage_events(*, trigger: UsageFlushTrigger) -> int:
     schedule another timer.
     """
     return _usage_event_buffer.flush_usage_events(trigger=trigger)
+
+
+def seen_source_idempotency_keys(source_keys: Iterable[str]) -> set[str]:
+    """Return candidate source keys retained by process-local admission history."""
+    return _usage_event_buffer.seen_source_idempotency_keys(source_keys)
 
 
 def drain_usage_events_after_executor_shutdown() -> None:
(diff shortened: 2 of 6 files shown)


seven332:
## Code Review: PR #28584

**Reviewed head:** `9a2a0ee8b9f1f732348f6d512d17aca6eed7b4aa`

### Summary

The runner now recovers an evicted WebSocket response's exact context/service tier from its existing bounded source-key admission history. If both bounded state layers lack evidence, positive output uses a conservative long-context category with explicit uncertainty telemetry instead of being omitted. Existing UUID inputs, duplicate suppression, capacity limits, and terminal cleanup remain intact.

### Findings

No findings.

### Testing

Reviewed the real WebSocket/reporting pipeline coverage for long-context-fast recovery after 100-entry flow eviction, base-output duplicate rejection after eviction, no-history fallback billing, state bounds, and terminal cleanup. Local validation passed Ruff, BasedPyright, the 24-test aggregation file, the complete 6,444-test add-on suite, and pre-commit hooks.

Production frequency of late frames after state pressure was not measured, but the correctness behavior no longer depends on that frequency.

### Verdict

LGTM

PR #28593, opened 2026-08-21
perf(guest-agent): bound reuse cleanup memory

### LINKED ISSUE #28567: perf(rust): bound directory-entry buffering in reuse cleanup
## Problem

Idle reuse cleanup eagerly materializes every entry name in a directory before it removes any child. The same unbounded `Vec<OsString>` pattern is used for the runtime parent and at every recursive level:

- `crates/guest-agent/src/reuse_preparation.rs` collects every direct child of the runtime parent before cleanup starts.
- `crates/guest-agent/src/nofollow_fs.rs` collects every child name before recursively removing the directory.

Because recursive calls run while ancestor vectors remain alive, outstanding name buffers accumulate across nested levels. There is no entry-count, byte, depth, or aggregate buffering bound.

This work is lifecycle-critical. `crates/runner/src/idle_reuse_preparation.rs` gives cleanup a 10-second budget, and `crates/runner/src/idle_pool/park_transition.rs` runs it inside final-exec-and-park before idle admission or handoff. Stale runtime trees can contain workload-created directory shapes.

## Evidence

At audited commit `4a8bf0cba369d67b648f888a55c554eb3bbd27cd`, an isolated benchmark enumerated 100,000 empty files with 79-byte names on Linux 6.1.155+, aarch64, Rust 1.98.0, and the ext-family filesystem reported by `statfs`:

| Mode | Enumeration time | Process `VmHWM` |
| --- | ---: | ---: |
| Eager collection (3 runs) | 48.3-54.8 ms | 13,272-13,288 KiB |
| One-at-a-time consumption (3 runs) | 44.6-45.0 ms | 864-880 KiB |

Eager collection added about 12.1 MiB of peak RSS for one practical-width directory and delayed the first unlink until enumeration completed. This is isolated allocation evidence, not an end-to-end latency claim.

The first planned fix used closed 256-name batches and reopened at directory offset zero between batches. It passed all 31 focused helper integration tests and removed the linear RSS growth, but the actual-helper benchmark disproved its throughput:

| Build | Entries | Cleanup wall time | Process max RSS |
| --- | ---: | ---: | ---: |
| Eager baseline | 1,000 | 0.0559 s | 7,972 KiB |
| Closed-stream candidate | 1,000 | 0.0594 s | 8,464 KiB |
| Eager baseline | 100,000 | 6.1727 s | 17,856 KiB |
| Closed-stream candidate | 100,000 | 39.4588 / 42.6503 s | 8,296 / 8,344 KiB |

Fixture creation was outside timing. Reopening from offset zero repeatedly scanned removed ext-family directory slots, making the candidate 6.4-6.9x slower and well beyond the 10-second budget. That plan is superseded.

Rust/POSIX `readdir` does not guarantee visibility after directory mutation, and Linux `getdents64::d_off` is filesystem-specific. Completeness therefore cannot rely on one mutating cursor pass alone; it requires a fresh verification pass.

This is distinct from #24780, which removed per-directory `/proc/self/fdinfo` identity reads and intentionally left enumeration for separately measured work.

## Decision

Use rustix `RawDir` with one shared fixed buffer and mandatory verification passes:

1. Allocate one 4 KiB `MaybeUninit<u8>` scratch buffer for the entire cleanup.
2. Read exactly one `getdents64` buffer from the already-open directory fd, skip `.`/`..` and protected names, copy checked owned names, then drop `RawDir` before deletion or recursive reuse of the scratch buffer.
3. Process the bounded chunk with the existing no-follow open, `statx` identity, same-mount/protected checks, and descriptor-relative unlink operations.
4. Recreate `RawDir` on the unchanged fd so the main pass continues from its cursor instead of rescanning from offset zero after every chunk.
5. At EOF, seek the directory fd to offset zero. If the pass removed anything, run another full pass. Terminate only when a complete pass removes no candidate.
6. Reject traversal beyond 256 directory levels. Enforce per-chunk owned-name count and payload limits derived from the 4 KiB raw buffer.

Rustix 1.1.4 is already a direct guest-agent dependency with `fs`; `RawDir` is its safe `getdents64` parser. No local unsafe ABI or new package is needed. One shared raw buffer plus one bounded copied chunk per active level and the depth cap gives an explicit aggregate bound independent of directory width or uncapped input depth.

A continuing pass may skip entries after mutation, but it cannot produce false completion: a new pass starts at offset zero before any mutation in that pass. For a finite static tree, an unprotected entry makes that pass remove something or fail; only a zero-removal pass can finish.

## Alternatives considered

- **Closed-stream offset-zero batches:** strongest portable mutation isolation, but measured at 39.46-42.65 seconds for 100,000 files and rejected by the lifecycle budget.
- **Larger closed batches:** reduces restarts only by trading back width-dependent memory; it does not remove the repeated-scan shape.
- **Continuing standard `ReadDir`:** preserves cursor progress, but retains implementation-defined libc iterator buffers at every recursive level, weakening the explicit aggregate bound.
- **Local raw `getdents64` parser:** duplicates safe rustix functionality and expands architecture-sensitive unsafe code.
- **Spill names to disk/tmpfs:** moves the unbounded resource to disk or memory precisely when cleanup is recovering those resources.
- **Keep eager collection:** preserves current throughput but leaves proven linear memory growth and delayed reclaim.

## Risks and mitigations

- **Mutation skips:** mandatory offset-zero passes continue until a full pass removes nothing.
- **Stale cursor entry:** tolerate only `ENOENT` caused by self-contained deletion; never relax identity, mount, symlink, budget, seek, or other I/O failures.
- **Recursive memory:** one raw buffer is shared; copied chunk count/payload and 256 levels give a fixed aggregate maximum.
- **Deep nonstandard trees:** depth beyond 256 rejects idle reuse rather than relying on stack/fd exhaustion. The completed run remains successful and the sandbox is not admitted.
- **Linux/filesystem behavior:** cleanup is already Linux-only; actual-helper ext-family measurements and both production musl builds gate completion.
- **Throughput:** repeat the retained-baseline actual-helper benchmark after implementation; no CI timing threshold is added.

## Acceptance criteria

- Runtime-parent and recursive cleanup no longer retain a full-directory name vector.
- One fixed raw buffer, explicit per-chunk owned-name count/payload limits, and an explicit traversal-depth limit bound aggregate entry-name/iterator memory independently of directory width and uncapped depth.
- Continuing cursor passes preserve near-linear main-pass work. Every mutating EOF is followed by an offset-zero pass, and cleanup succeeds only after a complete pass removes no unprotected candidate.
- Descriptor-relative no-follow traversal, protected identity checks, same-device-plus-mount enforcement, symlink handling, final protected-identity revalidation, and fail-closed behavior remain intact.
- Deterministic real-helper integration tests cover multiple parent/nested raw chunks, success at the depth limit, rejection beyond it, and all existing mount/symlink/protected/containment/report/lifecycle behavior.
- A focused actual-helper benchmark compares 1,000 and 100,000 entries against the retained eager baseline, demonstrating flat candidate RSS scaling and no material throughput regression. The superseded 39.46-42.65-second result stays recorded. CI uses no performance threshold or skipped benchmark test.
- Existing timeout, report, rejection, and completed-run semantics remain intact.
- No API, runner/guest protocol, dependency, persisted state, workspace-cache format, migration, or mixed-version coordination is introduced.

## Delivery

This is one independently reviewable Rust PR.

## Audit provenance

Found while fully reviewing `crates/guest-agent/src/nofollow_fs.rs` in a random 16-file Rust performance audit at `4a8bf0cba369d67b648f888a55c554eb3bbd27cd`.

seven332:
## Summary

- replace eager runtime-directory entry collection with a bounded Linux `RawDir` traversal
- share one 4 KiB raw directory buffer across recursion, cap copied names and traversal depth, and verify mutation with fresh passes from offset zero
- preserve descriptor-relative no-follow deletion, mount/protected identity checks, top-level removal counts, and fail-closed behavior
- cover wide parent and nested directories plus the exact depth boundary through the real `prepare-for-reuse` helper

## Performance

Measured with the real helper on synthetic valid containment trees, excluding fixture creation:

| Fixture | Implementation | Elapsed | Peak RSS |
| --- | --- | ---: | ---: |
| 100,000 stale entries | existing eager collection | 5.879–6.173 s | 17.84–17.91 MiB |
| 100,000 stale entries | bounded traversal | 6.015–6.381 s | 8.028–8.036 MiB |
| 1,000 stale entries | bounded traversal | 0.041 s | 8.024 MiB |

Peak RSS remains effectively flat as entry count increases and is about 55% lower at 100,000 entries. Median cleanup time increased by about 5% and remains below the issue's 10-second guardrail.

## Validation

- `cargo clippy -p guest-agent --all-targets -- -D warnings`
- `RUSTDOCFLAGS='-D warnings' cargo doc -p guest-agent --no-deps`
- `cargo test -p guest-agent --test reuse_preparation_helper`
- `cargo test -p guest-agent`
- `cargo build --target aarch64-unknown-linux-musl -p guest-agent --profile ci`
- `cargo build --target x86_64-unknown-linux-musl -p guest-agent --profile ci`

## Scope

This does not change the runner/guest protocol, API contracts, persisted state, protected-runtime selection, or cleanup timeout policy.

Closes #28567

### DIFF
--- a/crates/guest-agent/src/nofollow_fs.rs
+++ b/crates/guest-agent/src/nofollow_fs.rs
@@ -5,27 +5,41 @@
 //! `O_NOFOLLOW`, children are opened relative to their parent fd with
 //! `openat`, and directory iteration goes through `/proc/self/fd/{fd}`.
 //!
-//! Callers still own recursive traversal policy and business error
-//! handling. This module should only expose the primitive operations needed
-//! to safely open directories and regular-file candidates.
+//! Recursive removal copies one bounded `getdents64` chunk at a time from the
+//! live directory descriptor. A continuing cursor avoids rescanning removed
+//! slots, and full passes restart from offset zero until one removes nothing.
+//! Callers still own business error handling. This module should only expose
+//! the primitive operations needed to safely open directories and regular-file
+//! candidates.
 
 #[cfg(target_os = "linux")]
-use std::ffi::{CString, OsStr};
+use std::ffi::{CString, OsStr, OsString};
 #[cfg(target_os = "linux")]
 use std::fs::{self, File, OpenOptions};
 #[cfg(target_os = "linux")]
 use std::io;
 #[cfg(target_os = "linux")]
+use std::mem::MaybeUninit;
+#[cfg(target_os = "linux")]
 use std::os::fd::{AsRawFd, FromRawFd};
 #[cfg(target_os = "linux")]
-use std::os::unix::ffi::OsStrExt;
+use std::os::unix::ffi::{OsStrExt, OsStringExt};
 #[cfg(target_os = "linux")]
 use std::os::unix::fs::OpenOptionsExt;
 #[cfg(target_os = "linux")]
 use std::path::{Path, PathBuf};
 
 #[cfg(target_os = "linux")]
-use rustix::fs::{AtFlags, StatxFlags};
+use rustix::fs::{AtFlags, RawDir, SeekFrom, StatxFlags};
+
+#[cfg(target_os = "linux")]
+const REMOVE_RAW_DIR_BUFFER_BYTES: usize = 4 * 1024;
+#[cfg(target_os = "linux")]
+const REMOVE_ENTRY_CHUNK_MAX_NAMES: usize = 256;
+#[cfg(target_os = "linux")]
+const REMOVE_ENTRY_CHUNK_MAX_NAME_BYTES: usize = REMOVE_RAW_DIR_BUFFER_BYTES;
+#[cfg(target_os = "linux")]
+const REMOVE_DIRECTORY_MAX_DEPTH: usize = 256;
 
 #[cfg(target_os = "linux")]
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #28593

**Reviewed head:** `0f8255e5f5f4ebbddbd6879a5bb0f15f54c53de8`

### Summary

The cleanup now has explicit width/depth memory bounds while retaining near-linear directory cursor progress, descriptor-relative no-follow safety, fresh-pass completeness verification, and existing report/lifecycle behavior. The real-helper integration coverage and 100,000-entry measurements directly exercise the changed risks.

### Findings

No findings.

### Testing

Formatting, Clippy, rustdoc, the focused 31-test helper suite, the full guest-agent suite, and both production musl target builds passed. Actual-helper measurements show effectively flat candidate RSS from 1,000 to 100,000 entries, about 55% lower peak RSS than baseline at 100,000 entries, and approximately 5% median cleanup-time overhead. Privileged runner E2E was not run locally; no runner protocol or lifecycle policy changed.

### Verdict

LGTM

PR #28595, opened 2026-08-21
fix(platform): bind chat actions to current thread

### LINKED ISSUE #28573: bug: bind chat card actions to the current thread and agent
## Background

Chat cards upgrade trusted links in assistant `ChatEvent` content into interactive connector, permission, Computer Use, billing, mail, browser, and artifact surfaces. Connector and permission action URLs carry an `agentId`; optional `threadId` and `callbackPrompt` query parameters can start a continuation turn after the user completes the action.

These URLs are agent-authored message content. Trusted-origin validation proves only that the link points at a VM0/Okou surface. It does not prove that the target identifiers match the chat event, the current thread, or the thread's primary agent.

## Problem

Connector action parsing currently requires only a non-empty `agentId` query parameter, and permission action parsing requires only a non-empty agent path segment. Neither path binds the target to the agent that owns the current chat thread.

This has materially different outcomes depending on the bad identifier:

- malformed, missing, cross-organization, or inaccessible private-agent targets are rejected before a write;
- a different public agent in the same organization, or another agent owned by the current user, is a valid API target and can be modified successfully.

After the user clicks a card with a wrong-but-visible agent ID, the current user's built-in connector selection, custom connector grant, or permission allow/deny grant can be written for the unintended agent. Connector and permission cards do not display the target agent name or ID, so the mismatch is not visible in the confirmation surface.

The continuation path has the same semantic gap. It copies `agentId`, `threadId`, and `callbackPrompt` from the action URL into a canonical chat send. Existing-thread resolution verifies thread ownership and agent visibility independently, but does not require the requested agent to equal `chat_threads.agent_compose_id` or the callback thread to equal the thread rendering the card. A wrong-but-visible agent can therefore run inside the referenced user-owned thread while that thread retains its original primary-agent metadata, and a wrong user-owned thread ID can receive the continuation turn.

Rendering a card alone does not mutate data; user interaction is still required. Existing organization and private-agent visibility boundaries remain in force, and the affected grants belong to the authenticated user. The impact is nevertheless real: the user can unknowingly expose their connector access to runs of an unintended public agent and route a follow-up turn to the wrong execution target.

## Code-path reproduction

This was verified from the current `main` code paths; no production reproduction was performed.

1. Open a chat thread whose primary agent is Agent A.
2. Persist or receive an assistant message containing a trusted connector or permission action URL whose `agentId` names a different same-organization public Agent B.
3. Observe that the action card renders without identifying Agent B.
4. Complete the card action.
5. Observe from the action path that the current user's connector or permission state is applied to Agent B.
6. When the URL also carries the current thread ID and a callback prompt, observe that the continuation send uses Agent B even though the thread remains owned by Agent A.

The same callback path accepts a different user-owned `threadId`, routing the continuation away from the card's rendered thread.

## Expected behavior

An action card rendered in a chat thread must fail closed when its semantic target does not match the thread and agent context in which the user is confirming it. Agent-authored identifiers must not be sufficient to select a different mutation or continuation target.

## Decision

Bind connector and permission actions at the authenticated chat render boundary. The chat-thread snapshot and event stream already provide the current thread ID and its stable primary agent ID before assistant content is converted into cards. Treat that pair as authoritative and treat URL identifiers only as claims that must match it.

Deliver the correction as one frontend PR:

- pass the current `{ threadId, agentId }` through assistant card planning;
- classify protected connector and permission URLs as unrelated, recognized-but-invalid, or valid and context-bound;
- require the URL agent to equal the current agent;
- accept callback metadata only when both fields are absent, or when a non-empty prompt targets the current thread;
- construct executable descriptors from the validated context values after equality checks;
- render recognized invalid protected actions as one static unavailable card with no anchor, button, signals, mutation, or callback;
- keep standalone pages and general APIs unchanged.

The unavailable surface is required because an unresolved card slot currently falls back to an ordinary clickable Markdown link. Returning `null` for a mismatch alone would not fail closed.

### Add chat context to every existing API

This would touch connector selection, custom grants, permission grants, connector connection and OAuth/device/external-code state, and chat send. The context would have to remain optional for legitimate general callers, so it would be defense in depth rather than a new authorization boundary. The frontend binding and inert failure state would still be required, while strict schemas and independent deployment would add prepare/migrate/cleanup phases.

### Add event-bound action capabilities

A signed or persisted capability tied to the source event would provide stronger server provenance, but it requires a new action protocol, replay and lifetime rules, multi-step connector handoffs, and a historical strategy. Source event rows leave PostgreSQL after 30-day retention while older history remains available from archives. No other current requirement justifies this framework.

### Globally require chat send agent/thread equality

This would change independently authorized general send behavior and would not protect connector or permission mutations. It does not solve the full problem.

## Risks and mitigations

- **Future context-free parser call:** make chat context required for executable assistant action planning and remove the unread optimistic body-block parsing field that lacks an agent.
- **Clickable fallback:** use a registered static unavailable card rather than leaving a missing card registry entry.
- **Identity substitution:** reject a mismatch; do not rewrite an invalid URL to the current agent or thread.
- **Historical messages:** reparse stored event content in the current frontend without rewriting rows or requiring a live source-event database row.
- **Valid flow regression:** retain the existing downstream connector, custom connector, permission, and callback commands and cover current-agent/current-thread actions at the page entry point.
- **Protocol compatibility:** do not change an API contract or persisted format, so there is no cross-deployment wire migration.

## Acceptance criteria

- Connector and permission cards cannot mutate connector selections, custom connector grants, or permission grants for an agent other than the current chat thread's agent.
- A card continuation cannot start a different agent in the current thread or send the continuation to another thread.
- The semantic binding uses the authenticated thread context and is not based solely on agent-authored URL parameters.
- No executable connector or permission descriptor can be created without current chat context.
- A recognized context-invalid connector or permission action cannot degrade into an ordinary clickable link.
- Existing valid action cards targeting the current thread and agent continue to work, including supported callback continuations.
- Callback metadata is valid only when both fields are absent or a non-empty prompt is paired with the current thread ID.
- Malformed, missing, inaccessible, cross-organization, and wrong-but-visible identifiers fail without mutation or continuation.
- Integration coverage includes a different valid public agent in the same organization and a different valid thread owned by the same user, not only malformed or missing IDs.
- General-purpose connector-management and chat-send APIs retain their independently authorized behavior.

## Delivery and verification

This remains one PR, not a delivery parent. The context plumbing, tri-state parsing, unavailable surface, bound descriptors, page integration coverage, and documentation are one atomic frontend behavior.

Use page-level coverage in `chat-event-action-cards.test.tsx` to prove that valid alternate targets produce no mutation or callback requests and that valid current targets still work. Update existing parser/tree fixtures only for the required context and static-card type. Run app formatting, lint, type checking, focused Vitest coverage in one foreground process, and workspace-scoped Knip.

## Relevant files

- `chat-event-body-blocks.ts`
- `parse-body-blocks.ts`
- `connector-action-block.ts`
- `permission-action-block.ts`
- `action-callback.ts`
- `optimistic-chat-events.ts`
- `create-chat-thread.ts`
- `markdown-card-ref.ts`
- `chat-body-cards.tsx`
- `chat-event-action-cards.test.tsx`
- `docs/chat-cards.md`

## Current security boundaries

- Card actions require user interaction.
- Cross-organization agents are rejected.
- Another user's private agent is rejected.
- Connector and permission mutations are scoped to the authenticated user.
- Managed-browser cards already require the URL thread ID to equal the rendered thread ID before conversion.

seven332:
## Summary

- bind connector and permission chat actions to the authenticated thread's agent and thread IDs
- render malformed or mismatched protected actions as inert unavailable cards with no executable signals or links
- remove context-free optimistic parsing and document the trusted chat action boundary

## Testing

- `pnpm -F @okouai/app lint`
- `pnpm -F @okouai/app exec vitest run src/views/zero-page/__tests__/chat-event-action-cards.test.tsx src/signals/chat-page/__tests__/event-body-plan.test.ts src/signals/chat-page/__tests__/parse-body-blocks-cards.test.ts src/signals/chat-page/__tests__/parse-body-blocks-safari-url.test.ts src/views/zero-page/__tests__/zero-artifact-image-navigation.test.ts --maxWorkers=4 --silent=passed-only`
- pre-commit hooks: Prettier, Knip, repository-wide TypeScript checks, file-size check

## Scope

- existing connector and permission routes and APIs are unchanged
- no database schema or backend changes

Closes #28573

### DIFF
--- a/docs/chat-cards.md
+++ b/docs/chat-cards.md
@@ -28,17 +28,20 @@ state creation inside React render.
 The body parser accepts platform links in common message forms:
 
 ```markdown
-https://app.vm0.ai/agents/agent-123/permissions?connectorSlug=slack&permission=messages.write
+https://app.vm0.ai/agents/c0000000-0000-4000-a000-000000000001/permissions?connectorSlug=slack&permission=messages.write
 
-[Review permission](https://app.vm0.ai/agents/agent-123/permissions?connectorSlug=slack&permission=messages.write)
+[Review permission](https://app.vm0.ai/agents/c0000000-0000-4000-a000-000000000001/permissions?connectorSlug=slack&permission=messages.write)
 
 /computer-use/authorize/request-token
 ```
 
 Absolute URLs must use an allowed VM0 platform origin. Relative paths resolve
 against the configured platform origin. A URL becomes a card only when its path
-and required parameters match a card parser exactly. An unrecognized or invalid
-link remains ordinary Markdown.
+and required parameters match a card parser exactly. Unrecognized links remain
+ordinary Markdown. Recognized connector and permission actions are different:
+their URL targets are untrusted claims and must match the authenticated chat
+thread's agent and optional callback thread. A mismatch renders an inert
+unavailable card instead of a link or command.
 
 Current link-backed card patterns include:
 
@@ -67,9 +70,11 @@ that card type exists.
 
 ### 1. Parse content into pure descriptors
 
-`parseChatEventBodyBlocks` extracts renderable content from a `ChatEvent`
-and passes it to `parseBodyBlocks`. The parser separates normal Markdown from
-recognized cards.
+`chatEventTreePlan` extracts renderable content from a `ChatEvent` and passes it
+to `eventBodyPlan`. The parser separates normal Markdown from recognized cards.
+The current thread ID and server-derived primary agent ID are supplied as
+immutable planning context. Parsing is synchronous and does not query an API or
+database.
 
 A recognized card is first represented as a pure `ParsedBodyBlock`:
 
@@ -85,6 +90,13 @@ The descriptor contains parsed domain data only. It must not create `state`,
 `computed`, `command`, subscriptions, or other runtime resources. This keeps
 parsing deterministic and safe to run for persistent, IndexedDB, realtime, and
 optimistic messages.
+
+Connector and permission descriptors are created only after their URL agent ID
+matches the context agent ID and any callback contains both a non-empty prompt
+and the context thread ID. Valid descriptors use the context identities rather
+than copying URL claims. Recognized context-invalid actions produce only a
+static unavailable descriptor, so they register no action signals and cannot
+fall back to a clickable Markdown slot.
 
 ### 2. Derive a stable resource key
 
@@ -356,7 +368,8 @@ When adding a new link-backed card:
 7. Add the React component case and pass only the typed signals object.
 8. Cover absolute, relative, and Markdown link forms that the card accepts.
 9. Test repeated-resource sharing, persistent and optimistic message paths,
-   loading and mutation behavior, and invalid links falling back to Markdown.
+   loading and mutation behavior. Define explicitly whether invalid recognized
+   links remain Markdown or render inertly; mutating actions must fail closed.
 
 Do not create signals in the parser, transcript computed, or React component.
 Do not use a root-lifetime URL cache. Do not add a card to a shared aggregate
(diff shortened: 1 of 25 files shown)


seven332:
## Code Review: PR #28595

**Reviewed head:** `843bdad62ade63043c6faa3bc79c4816f46f2a5d`

### Summary

The final head binds connector and permission action claims to the rendered thread's authoritative agent and thread IDs, keeps invalid protected actions inert, and preserves existing general management APIs. The self-review fixes also prevent inline Markdown auto-link fallback and remove translation-copy coupling from the regression test.

### Findings

No findings.

### Testing

Platform lint, TypeScript checks, workspace-scoped Knip, formatting, diff checks, and a focused five-file/62-test Vitest integration run passed. Full repository tests and browser E2E were not run locally and remain covered by CI.

### Verdict

LGTM

Issue #28570, opened 2026-08-21
refactor: make current connector clients account-explicit


seven332:
## Parent objective

Delivery slice 8 of #22832. Move every currently supported compact connector client onto explicit single-account connection and disconnect semantics before visible multi-account UX is introduced.

## Problem

The account-aware backend is deployed, but current Platform connector flows and the CLI manual-grant flow still omit account intent. Current Platform disconnect also addresses a logical target rather than an exact connection.

Mapping the compact client to `list -> add/reconnect` is not safe:

- canonical account lifecycle reads are feature-gated;
- explicit multi-account `add` requires a user label for manual, no-auth, and generic custom flows, while compact clients have no account-name input;
- a read followed by mutation has a time-of-check/time-of-use race;
- custom compact responses do not consistently expose an exact connection ID;
- current disconnect must not silently clear thread selections.

The API therefore needs a first-class explicit compact operation. It must be separate from both an omitted legacy request and the multi-account `add`/exact `reconnect` contract.

## Decision

Introduce explicit atomic single-account connection and disconnect operations, then migrate current Platform and CLI entry points to them.

Connection semantics under the target lock are zero -> create the first/default account, one -> replace that exact account, and many -> conflict. Compact first-account creation may preserve a nullable user label; explicit multi-account `add` keeps its label requirements.

Disconnect semantics are zero -> missing, exactly one account -> delete that account and atomically clear every sparse thread preference referencing its UUID, and many -> conflict. Compact disconnect never guesses a sibling. Thread preferences do not own or block account lifecycle.

## Long-term account and permission boundary

The single-account operations in this slice are transitional support for current compact clients. Once Platform or CLI exposes account-aware behavior, creating an account uses explicit `add`, and every operation on an existing account carries its exact connection ID. Account-aware clients must not use the compact operation to choose an account implicitly.

Permissions remain scoped to the logical connector (`connectorSlug` or `customConnectorId`), independent of connector accounts. Account lifecycle and selection must not create account-scoped grants or mutate connector authorization.

## PR-sized sub-issues

- #28587 adds the explicit single-account connection operation and persists it through asynchronous authorization state.
- #28591 adds the explicit safe single-account disconnect operation.
- #28586 migrates all Platform entry points, including direct Artifact authorization and custom HTTP/MCP flows.
- #28590 migrates the current CLI manual-grant request.

## Acceptance criteria

- All four sub-issues are merged, deployed where applicable, and verified.
[...]

Issue #27695, opened 2026-08-17
chore: clean up connector account rollout compatibility before UI


seven332:
## Parent objective

Delivery slice 9 of #22832. Remove bounded rollout compatibility and historical data bridges before visible multi-account settings and thread selection are built.

## Problem

Connector code currently groups different semantics under `legacy-singleton` and nullable rollout contracts:

- old browser/API requests that have a finite deployment drain;
- already installed CLI requests that have no finite drain;
- source-less Runner/firewall/runtime payloads with queue and active-run lifetimes;
- deterministic default reads and intentional admission fallback;
- historical Feishu member links.

Those paths have different authorities and retirement gates. A single cleanup PR would either remove supported behavior too early or retain temporary credential-selection fallbacks indefinitely.

## Decision

Treat this as a cleanup delivery parent. First isolate the compatibility boundaries, then remove each temporary class only after its own production evidence gate.

Permanently or independently retained behavior is explicit:

- deterministic logical-target reads project the current default;
- the explicit compact single-account operation remains supported while compact UX exists;
- already installed CLI omission remains isolated and fails closed when a target is ambiguous;
- an authorized run may intentionally fall back from an unavailable ingress/thread preference to the current default and then to no credential.

## Delivery children

- #28588 separates public compact compatibility, exact/default resolution, and intentional admission fallback, and removes obsolete nullable default application types.
- #28589 coordinates the old-Platform contraction across two release-ordered PRs: #28697 cuts off old browser writers and target-only disconnect routes, then #28698 contracts persisted authorization mutation state after its drain evidence passes.
- #28592 removes source-less runtime credential selection after queue, active-run, Runner, sandbox, and rollback drains.
- #28585 reconciles Feishu member connector links and removes the remaining exactly-one runtime fallback.

## Production gates

Each child records the deployed version and evidence relevant to its authority. A PR merge timestamp is not a drain start.

- browser/app supported-version floor;
- API/DB rollback window;
- OAuth, OpenID, device, external-code, and custom OAuth state lifetime;
- queue plus claimed/finalizing run and run-context lifetime;
- Runner and active sandbox drain;
- Feishu reconciliation completeness and idempotence.

## Acceptance criteria

- Every delivery child and nested implementation child is complete and its production gate is recorded.
- Exact-source runtime, firewall, refresh, reconnect, and launch paths never select credentials through singleton fallback.
- Current Platform authorization and disconnect cannot omit their explicit compact operation.
[...]

PR #28598, opened 2026-08-21
feat(api): add safe single-account connector disconnect

### LINKED ISSUE #28591: feat(api): add safe single-account connector disconnect
## Problem

The compact connector UI disconnects by logical target because it does not know a concrete `connectors.id`. That is safe only while the target has at most one user-managed account. Once sibling accounts exist, target-only deletion must never guess which account to remove.

Thread account selections add a separate lifecycle concern. A sparse `chat_thread_connector_selections` row is a user preference for future run admission; it is not ownership of the account and must never prevent that account from being deleted. Treating a referenced singleton as a conflict would make account lifecycle depend on any number of historical threads and would leave the compact UI unable to disconnect it.

The lifecycle rules must distinguish explicit deletion from temporary unavailability:

- while an account still exists but is temporarily unusable, reconnect-required, mismatched, or storage-incompatible, retain the preference and fall back for that admission;
- when the account is explicitly deleted, remove every sparse preference that references its UUID in the same transaction. Those threads resume current-default inheritance, or use no credential when no usable account exists.

The current built-in and custom target-only routes remain rollout compatibility boundaries. A dedicated explicit compact operation is required so account-aware clients do not rely on omitted legacy request semantics.

## Decision

Add one target-union operation:

`DELETE /api/connector-accounts/single-account`

with a required built-in or custom logical target and no connection ID.

Resolve and mutate under the shared member/target lock:

- zero accounts: return the canonical missing result;
- exactly one account: delete that exact account and atomically clear all sparse thread selections referencing its UUID;
- more than one account: return a conflict and never choose a sibling.

Align exact account deletion with the same lifecycle invariant. Exact delete identifies the account by `connectionId`, automatically clears its referencing sparse selections, and does not require a public clear/reassign decision. If the deleted account was default and a sibling exists, promote the deterministic replacement default. Account permissions and authorization remain logical-target-scoped and unchanged.

Reuse shared lifecycle machinery for built-in, custom HTTP/API, and custom MCP accounts while preserving target-specific cleanup:

- built-in cleanup, token revocation, runtime finalization, and client invalidation;
- custom credential cleanup, runner/runtime wakeup, and client invalidation;
- Integration-managed custom definitions, including Feishu, remain on their Integration-owned lifecycle and are rejected by generic account APIs.

Keep the old target-only built-in and custom DELETE routes temporarily for deployed old clients. Remove them only through #27695 after its production drain gate.

## Scope

- Add the explicit compact single-account endpoint and target-union contract.
- Make singleton resolution atomic with deletion and reject only target ambiguity.
- Remove referenced-account conflicts and the public exact-delete selection-resolution requirement.
- Clear all selection rows referencing an explicitly deleted account in the deletion transaction.
- Preserve default promotion, provider cleanup, custom runner wakeup, and cache/client invalidation.
- Keep logical connector permissions, grants, capability, and network policy unchanged.
- Add bounded outcome telemetry without account IDs, labels, provider identity, credentials, thread IDs, or raw selection counts.
- Cover missing, singleton, referenced singleton, ambiguous target, concurrent mutation, exact delete, custom HTTP/MCP, Integration-managed rejection, and rollout compatibility at API entry points.

## Non-goals

- Account listing, naming, picking, or management UI.
- Reassigning deleted-account preferences to an arbitrary sibling.
- Deleting or rewriting preferences merely because an existing account is temporarily unavailable.
- Changing connector authorization or permissions.
- Removing legacy target-only routes before #27695's deployment gate.

## Acceptance criteria

- The compact Platform can disconnect a singleton without knowing its connection ID.
- A referenced singleton is deleted successfully; all selections for that UUID are removed atomically and affected threads inherit the current default or no credential.
- An exact account delete has the same automatic cleanup and requires no selection-resolution input.
- Ambiguous targets return a deterministic conflict without deleting or modifying any sibling.
- Temporary account unavailability still retains the stored preference and falls back only for run admission.
- Built-in, custom HTTP/API, and custom MCP share cardinality and selection-cleanup semantics while retaining target-specific cleanup.
- Custom account deletion publishes the runtime wakeup required to stop injecting deleted credentials.
- Feishu and other Integration-managed definitions cannot be disconnected through generic account APIs.
- Account deletion does not add, remove, duplicate, or re-evaluate logical connector authorization or permissions.
- Active admitted runs retain their already-resolved source; deletion affects later admissions.
- Entry-point integration tests cover the behavior above and all affected checks pass.

## Risks and mitigations

- **Wrong-account deletion:** serialize target mutations and reject when more than one account exists; exact deletion always requires `connectionId`.
- **Stale or resurrected preference:** delete referencing sparse rows transactionally with the account; never retain an orphaned deleted UUID.
- **Unexpected preference loss:** perform selection cleanup only for explicit account deletion, not for temporary runtime unavailability or admission fallback.
- **Default race:** use the shared target lock and promote a deterministic sibling before commit.
- **Custom runtime drift:** publish custom runtime wakeup and user invalidation after commit.
- **Cross-version behavior:** keep legacy routes until #27695 and deploy this API before the account-aware settings/composer clients in #27693 and #27694.
- **Feishu ownership drift:** retain Integration-managed ownership checks before generic member deletion.

## Dependencies and boundary

- Parent: #22832.
- Blocks compact Platform migration #28586.
- Exact account lifecycle foundations are in #27692; this issue corrects their deletion contract before settings UI #27693 consumes it.
- Account settings and per-thread picker remain in #27693 and #27694.
- Legacy route removal remains in #27695 after deployment evidence.

seven332:
## Summary

- add the canonical `DELETE /api/connector-accounts/single-account` operation for compact clients
- reject only ambiguous targets; a sole account is deleted even when threads selected it
- atomically remove every sparse thread preference referencing an explicitly deleted account so affected threads resume current-default inheritance
- align exact account deletion with the same invariant and remove the public clear/reassign prerequisite
- serialize thread selection writes and thread creation with account deletion through the shared target lock, preventing FK errors or stale preferences under races
- preserve built-in cleanup/revocation, custom runner wakeup and invalidation, deterministic default promotion, and Integration-managed Feishu exclusion
- keep temporary account unavailability distinct: admission may fall back, but the stored preference remains while the account still exists

## Compatibility

- old compact clients continue using the existing built-in and custom target-only disconnect routes until #27695 completes its production drain
- the exact account-manager delete contract now requires only the logical target; the account-aware settings UI has not shipped and remains blocked on this API correction
- new compact clients calling an old API receive a missing-route response instead of invoking a legacy handler
- no database migration, Platform UI change, Runner protocol change, or connector permission-state change is included

## Validation

- affected-file Prettier
- `pnpm -F api lint`
- `pnpm -F @okouai/api-contracts exec tsc --noEmit --pretty false`
- `pnpm -F api check-types`
- `pnpm -F @okouai/api-contracts exec vitest run src/contracts/__tests__/connector-accounts.test.ts src/contracts/__tests__/connector-client-contract.test.ts --maxWorkers=4 --silent=passed-only`
- `pnpm -F api exec vitest run src/signals/routes/__tests__/connector-accounts.test.ts src/signals/routes/__tests__/chat-threads-create.test.ts src/signals/routes/__tests__/feishu-integration.test.ts --maxWorkers=4 --silent=passed-only`
- pre-commit hooks, including repository-wide `pnpm knip` and `pnpm check-types`

Closes #28591

Delivery parent: #28570

Umbrella: #22832

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/chat-threads-create.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/chat-threads-create.test.ts
@@ -373,6 +373,27 @@ describe("POST /api/zero/chat-threads", () => {
         target: { kind: "builtin", connectorSlug: "openai" },
       },
     ]);
+    createRouteMocks(context).clerk.session(fixture.userId, fixture.orgId);
+    const disconnect = await accept(
+      connectorAccountsClient().disconnectSingleAccount({
+        headers: { authorization: "Bearer clerk-session" },
+        body: { target: { kind: "builtin", connectorSlug: "openai" } },
+      }),
+      [409],
+    );
+    expect(disconnect.body.error.message).toBe(
+      "Connector account is selected by chat threads",
+    );
+    const preservedAfterDisconnect = await accept(
+      connectorSelectionsClient().get({
+        headers: { authorization: `Bearer ${token}` },
+        params: { id: created.body.id },
+      }),
+      [200],
+    );
+    expect(preservedAfterDisconnect.body.selections).toStrictEqual(
+      selections.body.selections,
+    );
     const readToken = zeroToken({
       userId: fixture.userId,
       orgId: fixture.orgId,
@@ -686,6 +707,32 @@ describe("POST /api/zero/chat-threads", () => {
       }),
     ).toStrictEqual(
       expect.arrayContaining([httpConnectionId, mcpConnectionId]),
+    );
+
+    createRouteMocks(context).clerk.session(fixture.userId, fixture.orgId);
+    for (const customConnectorId of [httpConnector.id, mcpConnector.id]) {
+      const disconnect = await accept(
+        connectorAccountsClient().disconnectSingleAccount({
+          headers: { authorization: "Bearer clerk-session" },
+          body: {
+            target: { kind: "custom", customConnectorId },
+          },
+        }),
+        [409],
+      );
+      expect(disconnect.body.error.message).toBe(
+        "Connector account is selected by chat threads",
+      );
+    }
+    const preservedAfterDisconnect = await accept(
+      connectorSelectionsClient().get({
+        headers: { authorization: `Bearer ${token}` },
+        params: { id: created.body.id },
+      }),
+      [200],
+    );
+    expect(preservedAfterDisconnect.body.selections).toStrictEqual(
+      selections.body.selections,
     );
   });
 
--- a/turbo/apps/api/src/signals/routes/__tests__/chat-threads-create.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/chat-threads-create.test.ts
@@ -496,13 +496,22 @@ describe("POST /api/zero/chat-threads", () => {
       [200],
     );
     createRouteMocks(context).clerk.session(fixture.userId, fixture.orgId);
-    await accept(
+    const [concurrentSelectionWrite, concurrentDisconnect] = await Promise.all([
+      connectorSelectionsClient().update({
+        headers: { authorization: `Bearer ${token}` },
+        params: { id: created.body.id },
+        body: {
+          connectionId: connection.id,
+          target: { kind: "builtin", connectorSlug: "openai" },
+        },
+      }),
       connectorAccountsClient().disconnectSingleAccount({
         headers: { authorization: "Bearer clerk-session" },
         body: { target: { kind: "builtin", connectorSlug: "openai" } },
       }),
-      [204],
-    );
+    ]);
+    expect([200, 400]).toContain(concurrentSelectionWrite.status);
+    expect(concurrentDisconnect.status).toBe(204);
     const afterDisconnect = await accept(
       connectorSelectionsClient().get({
         headers: { authorization: `Bearer ${token}` },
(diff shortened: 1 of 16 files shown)


seven332:
## Code Review: PR #28598

**Reviewed head:** `d98c1bbca4059f8c2c9614aab2281c41458f788e`

### Summary

The compact endpoint rejects only ambiguous target-only deletion. A sole or exact account is deleted regardless of thread preferences, and all sparse selections referencing that UUID are cleared atomically so affected threads resume default inheritance. Temporary unavailability still retains the preference. Built-in/custom cleanup, custom runtime wakeup, deterministic default promotion, logical-target permissions, and Integration-managed Feishu exclusion remain intact.

Thread selection writes and new-thread selection creation use the same target lock as account deletion. Under a race, deletion either clears the committed preference or makes the selection request return normal validation failure; it cannot produce an orphan or selection-derived 500.

### Findings

This review found that the shared deletion-preparation helper still accepted a general database handle even though its selection cleanup and default promotion must always be transactional. Commit `d98c1bbca4` now requires `Tx`, encoding the atomicity invariant in the type system.

An earlier review pass found and fixed the selection-write/deletion concurrency gap in `9721eca3cf`. A repeated complete six-pass review of the current head found no unresolved issues.

### Testing

- affected-file Prettier
- API lint and type checking
- 3 related API integration test files, 59 tests
- pre-commit repository Knip and type checks
- full required PR CI on the reviewed head

### Verdict

LGTM

Issue #28148, opened 2026-08-19
feat: add later-run fallback for vm0-managed model providers


seven332:
## Problem

Every VM0 built-in model currently has one preferred managed provider route. Without a later-run fallback, an unusable preferred route keeps being selected until an operator changes configuration.

Failure reporting does not replay or reroute the current run. A confirmed provider failure only makes the exact route temporarily unavailable to later newly constructed runs; the current run continues its independent lifecycle.

## Target route order

| Family | Primary | Fallback |
| --- | --- | --- |
| Claude | direct Anthropic | OpenRouter Anthropic Messages |
| GPT | direct OpenAI | OpenRouter Responses |
| DeepSeek | direct DeepSeek Responses | OpenRouter Responses |

OpenRouter DeepSeek remains a required fallback candidate. Because it does not support Codex's native custom/freeform apply-patch contract, that tool is disabled only on the fallback route rather than removing the route or translating tool semantics in the host proxy.

BYOK, personal, custom gateway, OAuth, and cloud-specific routes remain outside managed fallback.

## Runtime behavior

- Candidates use a narrow statically compilable VM0-managed provider type.
- A cooldown identity is canonical model + concrete provider type + exact upstream model.
- The selected provider, upstream model, and managed-key ID are persisted before session selection and reused for session continuity and execution.
- The selector chooses the first candidate with a current managed key and no active exact-route cooldown.
- With `ManagedModelProviderFallback` disabled, candidate 0 is selected and cooldown state is ignored.
- When no candidate is usable, admission returns typed 503 `MODEL_PROVIDER_UNAVAILABLE` before run construction.
- A trusted eligible failure monotonically extends `unavailable_until` for the exact route.
- A validated `Retry-After` is clamped to 1–300 seconds; otherwise the default is 60 seconds.
- Success is a no-op. Recovery occurs through expiry or explicit route-row maintenance.
- The accepted shared-key worst case is one failure per canonical model before each exact route receives its own cooldown.

## Trust and failure evidence

- Only authenticated official runners may report managed-route failures.
- The API derives model, provider, and upstream identity from persisted run metadata; the reporter cannot select the affected route.
- Runner-hosted mitmproxy reports confirmed eligible HTTP, connection, JSON, SSE, and Responses WebSocket failures directly.
- Invalid requests, unsupported parameters, context/output limits, content/safety/refusal, permission/guardrail, tool/app/user-code failures, and cancellation by themselves do not trigger cooldown.
[...]

Issue #28195, opened 2026-08-19
feat: qualify and roll out fallback routes for all built-in models


seven332:
## Parent objective

Sub-issue of #28148. This is the final delivery slice: qualify every configured managed route, verify the fallback control plane, and deliberately roll out the default-off capability across Claude, GPT, and DeepSeek families.

## Dependencies

- #28710 owns the deployed runner E2E proving that a later run completes through fallback; this rollout begins only after that regression is green.
- #28187, #28192, #28193, and #28194 delivered the route identity, resolver, failure API, and host-proxy producer.
- #28643 and #28650 fixed failure classification and bounded `Retry-After` handling.
- #28651 must keep both OpenRouter DeepSeek candidates while disabling Codex's unsupported native apply-patch tool in the API-projected fallback catalog. Direct DeepSeek retains freeform apply patch.

## Problem

The compiled candidate catalog does not prove account-specific availability: managed credentials may be invalid or unfunded, and an exact upstream model may not be entitled. The feature also needs a controlled rollout proving that accepted cooldown evidence affects only later new runs and that rollback remains safe.

Protocol and tool compatibility should not require an exhaustive live Codex canary matrix. Official provider/model/API documentation plus a static comparison with the pinned runtime contract is sufficient to establish declared tool, streaming, and request/response shapes. Live calls are reserved for facts that documentation cannot establish.

### Documentation and static contract review

For every configured candidate, record the official model and API documentation covering:

- exact upstream model availability;
- endpoint and wire protocol;
- standard function tools and `tool_choice`;
- JSON, SSE, and Responses streaming shapes used by the route;
- upstream price, canonical customer price, and data-processing policy;
- any deliberately disabled runtime capability.

For OpenRouter DeepSeek, record that:

- DeepSeek V4 Flash and Pro support standard function tools.
- OpenRouter's beta `openrouter:apply_patch` server tool is not the Codex custom/freeform local apply-patch contract.
- #28651 keeps ordinary function tools and streaming unchanged, disables native Codex apply patch only on the two fallback routes, and preserves direct DeepSeek's freeform catalog.

References:

- https://openrouter.ai/deepseek/deepseek-v4-flash
- https://openrouter.ai/deepseek/deepseek-v4-pro
- https://openrouter.ai/docs/guides/features/tool-calling
- https://openrouter.ai/docs/guides/features/server-tools/apply-patch
- https://openrouter.ai/docs/api/reference/responses/basic-usage

[...]

PR #28644, opened 2026-08-22
fix(python): classify openrouter 500 responses by body

### LINKED ISSUE #28643: bug: classify OpenRouter 500 responses by body
## Background

The managed-model fallback reporter intentionally treats OpenRouter HTTP 500 responses differently from most providers. OpenRouter can wrap request/schema/tool incompatibilities in HTTP 500, so the reporter must inspect the response body before deciding that the provider route is unavailable.

This was found while reviewing #28148 before enabling its org-scoped fallback switch. The switch is currently disabled by default, so the bug is dormant until canary or enablement.

## Problem

`crates/runner/mitm-addon/src/model_provider_failure.py` suppresses status-only classification only when the firewall name is exactly `model-provider:openrouter`.

Production firewall configuration never emits that name. The two real OpenRouter firewall names are:

- `model-provider:openrouter-api-key`
- `model-provider:openrouter-codex`

As a result, a real OpenRouter HTTP 500 response is classified as `provider_unavailable` during response-header handling, before its error body can identify an `invalid_request` or other non-provider failure. The reporter then opens an exact-route cooldown and later runs may fall back even though the route itself is healthy.

Existing integration tests use the nonexistent `model-provider:openrouter` name, so they do not exercise production behavior.

## Reproduction

1. Admit a billable model-provider flow using either real OpenRouter firewall name.
2. Return HTTP 500 with an OpenRouter error body whose error type is `invalid_request`, such as a rejected tool declaration.
3. Observe that the runner reports `provider_unavailable` at response-header time.

## Expected behavior

- OpenRouter HTTP 500 responses using either production firewall name are not classified from status alone.
- The response body determines whether the failure is a confirmed provider outage.
- Request, schema, and tool-capability errors do not create cooldown state.
- Confirmed OpenRouter provider-unavailable errors still report normally.

## Acceptance criteria

- [ ] Both production OpenRouter firewall names use body-aware HTTP 500 classification.
- [ ] Integration coverage proves HTTP 500 plus `invalid_request` does not report a provider failure for either firewall.
- [ ] Integration coverage proves HTTP 500 plus `provider_unavailable` still reports for both firewalls.
- [ ] Tests no longer rely on the nonexistent `model-provider:openrouter` firewall name.
- [ ] The relevant Python format, lint, type, and integration-test checks pass.

## Scope

This issue fixes failure classification only. Route qualification, DeepSeek/OpenRouter tool compatibility, end-to-end fallback canaries, and selection observability remain tracked by #28195. It does not enable `ManagedModelProviderFallback`.

Parent objective: #28148.

seven332:
## Summary

- defer OpenRouter HTTP 500 classification to the response body for both production firewall variants
- keep request, schema, and tool-capability errors from opening false provider cooldowns
- replace stale firewall fixtures and cover both `invalid_request` and confirmed `provider_unavailable` outcomes

## Scope

This does not enable managed-model fallback, change route selection, or change `Retry-After` handling.

Closes #28643

Parent objective: #28148

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (`6513 passed`)

### DIFF
--- a/crates/runner/mitm-addon/src/model_provider_failure.py
+++ b/crates/runner/mitm-addon/src/model_provider_failure.py
@@ -729,9 +729,8 @@ def _failure_kind_from_http_status(status: int, firewall_name: str) -> FailureKi
         _HTTP_STATUS_SITE_OVERLOADED,
     ):
         return "provider_unavailable"
-    if (
-        status == _HTTP_STATUS_INTERNAL_SERVER_ERROR
-        and firewall_name != "model-provider:openrouter"
+    if status == _HTTP_STATUS_INTERNAL_SERVER_ERROR and not firewall_name.startswith(
+        "model-provider:openrouter-"
     ):
         return "provider_unavailable"
     return None
--- a/crates/runner/mitm-addon/tests/test_model_provider_failure_reporting.py
+++ b/crates/runner/mitm-addon/tests/test_model_provider_failure_reporting.py
@@ -325,7 +325,8 @@ def test_openrouter_edge_timeout_reports_normalized_failure(
     flow = _make_flow(
         real_flow,
         tmp_path / "proxy.jsonl",
-        firewall_name="model-provider:openrouter",
+        firewall_name="model-provider:openrouter-api-key",
+        request_path="/api/v1/messages",
         response_status=524,
     )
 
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #28644

**Reviewed head:** `ae79f66bcd3009a62d053658e84ab9e0ec957999`

### Summary

The change correctly maps both production OpenRouter firewall variants to body-aware HTTP 500 classification without duplicating the concrete TypeScript catalog or changing any external contract. Existing non-OpenRouter status behavior is preserved.

### Findings

No findings.

### Testing

The regression coverage drives real addon response hooks for both production firewall names and proves both boundaries: `invalid_request` emits no report, while `provider_unavailable` does. Ruff format/lint, BasedPyright, the focused reporter tests (`68 passed`), and the complete addon suite (`6513 passed`) passed. Live route qualification/canary remains tracked by #28195 and is not required for this classifier fix.

### Verdict

LGTM

PR #28645, opened 2026-08-22
feat(presentation): switch built-in decks to the latest template archives

bingjiez666:
## What

The 23 built-in presentation packages get a **second published archive per template**, selected by a new `LatestPresentationTemplates` feature switch. Outside the switch every run keeps the exact archive and prompt it uses today; inside it, runs pull the new archive and get guidance written for it.

The new archives are a different kind of package:

| | outside the switch | inside the switch |
|---|---|---|
| entrypoint | `AGENT_RUNBOOK.md` | `SKILL.md` |
| authoring | deck JSON → bundled renderer | agent writes the HTML directly |
| colour system | `"colorSystem"` field in deck JSON | inline `color-systems/<token>.css`, `data-color-system` on the root |
| images | resolved by the bundled runbook | generated per slot, `seedream4` by default |

Because the entrypoint and the image workflow both move, the prompt has to switch **with** the archive. Naming `AGENT_RUNBOOK.md` against a package that does not ship one costs a whole run, so one switch selects the digest, the instructions, and the image path together.

## Serving both digests

The registry pins one content digest per package and the download route is asked for a specific digest. A run created while the switch was off keeps requesting the pre-cutover digest long after the switch flips, and both R2 objects are immutable, so both have to stay resolvable.

`turbo/apps/api/src/lib/presentation-runbook-archive-versions.ts` maps each package's previous digest to its previous version id, mirroring `website-template-archive-versions.ts`. It replaces the old `PRESENTATION_RUNBOOK_ARCHIVE_VERSION_IDS_BY_SHA256` remap in the download route, which was only ever consulted when the requested digest equalled the registry default — i.e. it could serve exactly one digest per package and could not have served two.

The digests and version ids were computed from the packaged archives with the same content-addressing the publisher uses (`versionId = sha256("storage:<storageId>\n" + sorted "<path>:<sha256>" lines)`). The formula was verified end-to-end first: recomputing it over the **currently published** `bloom` archive reproduces production's head version id, file count, and byte size exactly.

## Fallbacks

[...]

PR #26507, opened 2026-08-12
chore: one-time publish Website template archives

bingjiez666:
## Purpose

One-time, auditable production publication for the Website template archives pinned by #26434.

This PR must not be merged. Its workflow:

- checks out vm0-ai/Template-artifact at 78dbe053b2b485d907d7e0116022ebdfe807038f;
- deterministically builds all 11 archives and verifies every SHA-256/version id;
- renders every newly built archive before requesting production access;
- writes immutable R2 objects without overwriting existing versions;
- verifies and renders both the old and new R2 versions;
- registers the new storage versions in one production database transaction while retaining all old rows.

After the production job succeeds, close this PR and delete the branch. The Actions run is the release audit record.

## Local evidence

- Exact archive SHA-256 and version pins reproduced for 11/11 templates.
- New archives extracted and rendered: 11/11.
- Current production archives pulled with the released CLI and rendered: 11/11.
- Prettier, node syntax check, actionlint, and diff check passed.
- Full workspace lint reached API type-aware oxlint and was killed by the local host (SIGKILL); all 12 preceding package lint jobs passed. No API files are changed here.

PR #28646, opened 2026-08-22
chore: one-time publish presentation template archives

bingjiez666:
> **Do not merge.** The Actions run is the audit record, same as the website template publication in #26507.

## What

Publishes one new immutable archive per built-in presentation template (23 of them) from `vm0-ai/Template-artifact@581b752` (`Template-Presentation-v3/`), so the enabled side of `LatestPresentationTemplates` in #28645 has something to pull.

Each archive lands as a **new version on the existing storage** for that package. The predecessor version row and its R2 objects are untouched, because the disabled side of the switch keeps requesting those exact bytes.

## Why the pins can be trusted

Everything is content-derived and re-verified before anything is written:

- `versionId = sha256("storage:<storageId>\n" + sorted "<path>:<sha256(contents)>" lines)`
- archive digest = sha256 of a deterministic tarball (`noMtime`, `portable`)

The formula was validated end-to-end before any of these numbers were produced: recomputing it over the **currently published** `bloom` archive (pulled from production through `okou resource pull`) reproduces production's head version id `a7c6805b…`, its file count (35) and its byte size (1 249 997) exactly.

`build` recomputes both pins from source and refuses to write an artifact that disagrees. `verify` re-extracts the committed tarballs and recomputes the manifest, version id and digest from the extracted bytes. `publish` re-reads both the new and the **old** object back out of R2 and checks both digests before it touches the database.

## Structural contract instead of a render smoke test

The website release runs each template's `render.mjs` as a smoke test. These packages carry no renderer — the run authors the HTML itself — so `inspectArchive` checks shape instead:

- `SKILL.md` (the entrypoint the prompt names), `design-system.md`, `layouts/_shell.html`, and the package's default colour CSS are all present and non-empty
- every `color-systems/*.css` reference in the package resolves, **relative to the referring file**, to a file the archive actually carries

The second one is not hypothetical: these templates used to read `../color-systems/` from the package root, which resolves outside the extracted directory and silently yields an unstyled deck.

Both sides of that check are calibrated. It passes on all 23 real templates (19 colour systems each), and rewriting a single `href="../color-systems/carnival.css"` to `../../` in one layout makes it fail with exactly that file named.

## Local run

```
```

Both were run against the committed script and the committed tarballs. The digests they print are the ones already pinned in #28645.

## Order

This has to complete before `LatestPresentationTemplates` is turned on for anyone. #28645 defaults the switch off, so the two can land in either order; only the rollout depends on this run.


bingjiez666:
Published. 23/23 archives written to R2 and registered in the production database.

Confirmed independently of this run's log, by reading the production database: all 23 new `storage_versions` rows exist with `file_count`/`size`/`storage_id` matching the locally computed manifests, and all 23 predecessor version rows are still present. A `okou resource pull` of `template:html-ppt-bloom-pitch-runbook` still returns the previous 35-file package, so the disabled side of the switch is unaffected.

Closing without merging, per the convention this follows (#26507). The branch and this run stay as the audit record.

PR #28702, opened 2026-08-23
feat(cli): centralize presentation image batches

bingjiez666:
## Summary

- add `okou generate image-batch start|wait` as the single VM0-owned presentation image-batch interface
- accept any non-empty TSV manifest, use Seedream4 low-quality PNG output, keep at most three requests in flight, retry one transient failure, and preserve manifest order in `results.tsv`
- default to the smallest API-valid square output, `816x816`, let the agent choose an optional size per image, and delegate size validation to the existing image API
- return every generated image as an unambiguous `asset-id<TAB>URL` row using the manifest's unique asset id
- detach the batch behind a state directory so HTML authoring continues immediately after `start`
- express image generation as one ordered prompt workflow: `start → author → wait → embed`
- share one VM0 presentation execution contract across built-in templates, imported templates, and template-free CLI generation
- leave template files, colour systems, layout loading, local gates, repair rules, and visual-review scope entirely to each template's `SKILL.md`
- keep `resource-registry.ts` free of built-in directory and gate assumptions; it now selects the package, passes the selected colour token, and delegates to the skill
- condense the static-HTML and JavaScript contract to one line

## Verification

- Prettier check for all changed TypeScript files
- core tests (205/205)
- CLI image-batch and presentation prompt tests, including five jobs with at most three in flight, the `816x816` default, and a custom `2048x1024` per-image size (11/11)
- API generation-template prompt tests (21/21)
- targeted run-system-prompt BDD test (1/1)
- CLI, core, and API type checks
- CLI, core, and API lint
- scoped Knip for `@okouai/cli`, `@okouai/core`, and `api`
- `git diff --check`

## Rollout order

Publish the CLI containing `image-batch` before enabling the updated latest-presentation prompt in API/CLI deployments. Template packages no longer carry an image helper.

PR #28724, opened 2026-08-23
chore: republish presentation template archives from a4cf54d

bingjiez666:
> **Do not merge.** The Actions run is the audit record, same as #28646 and #26507.

## What

Republishes all 23 built-in presentation runbook archives from `vm0-ai/Template-artifact@a4cf54d`, superseding the `581b752` set published in #28646.

Each archive lands as a **new version on the existing storage**. The pre-cutover (deck-JSON renderer) versions published before #28645 are untouched — they are still the switch-off path and the drained-CLI fallback.

## What actually changed upstream

14 commits, and the diff touches **only the 23 `SKILL.md` files** (+498/−243). Nothing else in any package moved:

- verification is now **gate-first and advisory-only** — the blocking `tools/run.sh` assertions (`titleBodyOverlaps`, `safeAreaShells`, `outsideSafeArea`, `emptyPhotocells` all `= 0`) are gone; the report is a recommendation, and a zero report is no longer required
- the report is declared to *be* the visual review, so a run no longer takes screenshots or contact sheets unless asked
- each skill is told to batch-read its step's files and not reread unchanged ones
- template-owned image batches were added and then removed again (`dee949b` → `07736a3`), which is why the templates now contain **no `okou` or image-batch instruction at all** — that mechanism moved into the vm0 prompt in #28702

Package shape is unchanged and was re-checked on all 23: `SKILL.md`, `design-system.md`, `layouts/_shell.html`, `decoration/`, `example/`, `tools/`, and 19 `color-systems/*.css` each. So the switch-on prompt still points at the same entrypoint and needs no change for this publish.

## Verification

`build` recomputes both pins from source and refuses to write an artifact that disagrees; `verify` re-extracts the committed tarballs and recomputes the manifest, version id and digest from the extracted bytes; `publish` re-reads the new **and** the old object back out of R2 and checks both digests before touching the database.

Beyond digests, `inspectArchive` re-asserts the structural contract on every archive: `SKILL.md` / `design-system.md` / `layouts/_shell.html` / the default colour CSS present and non-empty, and every `color-systems/*.css` reference resolving **relative to the referring file** to a file the archive carries.

Local run against the committed script and tarballs:

```
```

## Follow-up

The vm0 PR repointing the registry digests and version ids to this set is linked below once opened. Order does not matter for safety — `LatestPresentationTemplates` is off by default and staff-only — but the archives should exist before the pins move.


bingjiez666:
Published. 23/23 archives written to R2 and registered in the production database.

Verified independently of this run's log by reading the production database: all 23 new `storage_versions` rows exist with `file_count`/`size`/`storage_id` matching the locally computed manifests.

Registry pins move in #28728.

Closing without merging, per the convention this follows (#28646, #26507). The branch and this run stay as the audit record.

Issue #28672, opened 2026-08-23
chore: remove LatestPresentationTemplates and its previous-digest fallback


bingjiez666:
Follow-up to #28645, which is the removal condition for the fallback that PR introduces.

## What to remove

Once `FeatureSwitchKey.LatestPresentationTemplates` is terminal — `enabled: true` for everyone, or abandoned — delete the switch and its disabled branch together, in one PR:

- `turbo/packages/core/src/feature-switch-key.ts` — the `LatestPresentationTemplates` key
- `turbo/packages/core/src/feature-switch.ts` — its registry entry
- `turbo/packages/core/src/resource-registry.ts` — `PREVIOUS_PRESENTATION_RUNBOOK_ARCHIVE_SHA256`, `PREVIOUS_PRESENTATION_RUNBOOK_PACKAGES`, the `archiveVersion` parameters, and the `"previous"` branch of `buildPresentationRunbookInstructionLines`
- `turbo/apps/api/src/lib/presentation-runbook-archive-versions.ts` — the whole module, once nothing requests the previous digest
- `turbo/apps/api/src/signals/routes/registry-resources-download.ts` — the `resolvePresentationRunbookArchiveVersionId` call
- `turbo/apps/api/src/signals/services/agent-run-create.service.ts` — `presentationRunbookArchiveVersionForRun` and the `OKOU_PRESENTATION_RUNBOOK_ARCHIVE_VERSION` entry, plus its `runtimeOverrideKeys` declaration in `agent-execution-plan.ts` and the mirror in `turbo/packages/db/scripts/test-agent-compose-consolidation-preflight.ts`
- `turbo/apps/cli/src/commands/shared/presentation-runbook-archive-version.ts` and its call sites
- the tests that exercise the disabled branch, in the same PR (they cover the fallback, so they die with it rather than becoming tombstones)

## When it is safe

The fallback is sized to the **existing runner/sandbox** surface: **up to 2 hours**, bounded by `JOB_TIMEOUT = Duration::from_secs(7200)` in `crates/runner/src/executor/mod.rs`. A run whose execution context pinned a `CLI_PKG_URL` from before #28645 carries a CLI whose bundled registry only knows the pre-cutover digest, and it keeps requesting that digest for the queue lifetime plus a claimed run.

So the module is removable once **both** hold:

1. the switch is terminal, and
2. at least 2 hours have passed since the release that made every new execution context carry the post-cutover CLI.

## What must not be deleted

[...]

PR #28728, opened 2026-08-23
feat(presentation): refresh the built-in template archives to 71ff2fb

bingjiez666:
## What

Repoints the enabled side of `LatestPresentationTemplates` at the archives republished from `vm0-ai/Template-artifact@71ff2fb`, superseding the `581b752` set that #28645 pinned.

The archives are **already live** — published and verified in #28724 (`a4cf54d`) and #28738 (`71ff2fb`, the set this PR pins), each confirmed against the production database independently of the job log. This PR only moves the pins.

## What changed upstream

16 commits in two batches, and every one of them touches **only the 23 `SKILL.md` files**.

**Batch 1 — `a4cf54d`** (14 commits, +498/−243):

- verification is now **gate-first and advisory-only** — the blocking `tools/run.sh` assertions (`titleBodyOverlaps`, `safeAreaShells`, `outsideSafeArea`, `emptyPhotocells` all `= 0`) are gone; the report is a recommendation and a zero report is no longer required
- the report is declared to *be* the visual review, so a run no longer takes screenshots or contact sheets unless asked
- each skill batch-reads its step's files instead of rereading unchanged ones
- template-owned image batches were added and then removed again (`dee949b` → `07736a3`)

**Batch 2 — `71ff2fb`** (2 commits, +4/−376): the "Check what the report cannot see" section is dropped, so the finished deck is no longer re-checked by hand now that the automated report *is* the visual review.

Package shape is unchanged, re-checked on all 23: `SKILL.md`, `design-system.md`, `layouts/_shell.html`, `decoration/`, `example/`, `tools/`, and 19 `color-systems/*.css` each.

## Why the prompt is untouched

The templates now contain **no `okou` or image-batch instruction at all** — `grep` over all 23 `SKILL.md` returns nothing. That mechanism moved into the vm0 prompt in #28702, which already landed. So the switch-on prompt keeps naming the same entrypoint and the same image workflow, and this PR is pins only.

## Fallbacks

None introduced.

The superseded predecessor digests are deliberately **not** kept resolvable. `LatestPresentationTemplates` is `enabled: false` and staff-only, so it has no external users and needs no compatibility for its own earlier iteration (`docs/fallback.md` §2 — "a temporary 500 or a broken screen during the cutover is acceptable"). Accumulating a map per refresh would be exactly the slop that section forbids.

The **pre-cutover** map in `presentation-runbook-archive-versions.ts` is untouched. That one is a different surface: it protects the switch-off path and run contexts pinned to a pre-#28645 `CLI_PKG_URL`, and it keeps its declared window and removal condition (#28672).

## Test plan

- `pnpm check-types` — 14/14
- `pnpm turbo run lint --concurrency=1` (3072 MiB heap) — 13/13
- `pnpm -F @okouai/core exec vitest run` — 205 passed
- API tests run in CI (no Postgres in the dev container)

`registry-resources-download.test.ts` moves with the pins: the "current registry digest" case now asserts the `71ff2fb` schoolhouse digest resolves to its new version id, while the pre-cutover case is unchanged and still proves a drained run context gets the old bytes.


bingjiez666:
Moved to `71ff2fb`. Upstream added two more commits while this was open (PR #34): the by-hand re-check of the finished deck is dropped from all 23 `SKILL.md` now that the automated report is declared to be the visual review (+4/−376). Still `SKILL.md`-only; package shape and the prompt are unchanged.

Archives for this set are published and verified in #28738 — 23/23, cross-checked against the production database independently of the job log. Before building, all 23 storages were confirmed to head at the `a4cf54d` versions, so this run supersedes exactly the set the previous commit here pinned.

Pins, and the schoolhouse digests in `registry-resources-download.test.ts`, move together in `e803500`.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: e803500f91506422bbfdb7849c24ab2ace12a7ee
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Repoints the enabled side of `LatestPresentationTemplates` at the presentation runbook archives republished from `vm0-ai/Template-artifact@71ff2fb`. Three files, and every changed line is a 64-hex constant: 23 archive digests in the core registry, 23 storage version ids in the API resolver, and the two schoolhouse constants the download-route test pins.

No new code path, no behaviour change, no new fallback.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic — no wrong assumptions | Pass | Every pin re-derived independently, see "Pin verification" below. All 23 registry digests equal the sha256 of the tarballs built from `71ff2fb`; all 23 version ids equal `sha256("storage:<storageId>\n" + sorted manifest)` and resolve to live `storage_versions` rows in production. |
| `REVIEW.md` | Security — no secrets, no auth change | Pass | Constants only. No route, contract, or auth surface touched. |
| `REVIEW.md` | Type safety — no `any`, no broad types | Pass | Both maps keep their existing `Record<string, string>` / `satisfies Record<string, string>` shapes; only values change. |
| `REVIEW.md` | Feature regression guard — existing flow covered | Pass | The switch-off path is untouched: `PREVIOUS_PRESENTATION_RUNBOOK_ARCHIVE_VERSION_IDS_BY_SHA256` is byte-identical to `main` (verified by diffing the extracted block), and `PREVIOUS_PRESENTATION_RUNBOOK_ARCHIVE_SHA256` in core is not in the diff. |
| `REVIEW.md` | DB/JSONB gate | N/A | No schema, contract, or `jsonb().$type<>()` change. |
| `docs/bad-smell.md` | Zero `any`, zero suppressions, no dynamic `import()`, no hardcoded config | Pass | Diff contains only hex string literals inside existing object literals. |
| `docs/fallback.md` §2 | Non-GA feature-switched code needs no compatibility for its own earlier shape | Pass | The superseded `a4cf54d` digests are dropped rather than accumulated into a third map. `LatestPresentationTemplates` is `enabled: false` + staff-only, so it has no external users. Confirmed no `a4cf54d` value survives anywhere in either file. |
| `docs/fallback.md` §8/§9 | An existing rollout fallback keeps its surface, window, removal condition | Pass | The pre-cutover fallback and its comment survive intact — `Rollout fallback`, `existing runner/sandbox, up to 2 hours`, `JOB_TIMEOUT`, and follow-up `#28672` all still present. This PR introduces none of its own, so no new `Fallbacks` entry is owed. |
| `docs/deployment-compatibility.md` | Commit-addressed CLI artifacts — queued runs keep their pinned `CLI_PKG_URL` | Pass | Unchanged by this PR: the pre-cutover map that serves those contexts is untouched. Contexts pinned to a post-#28645 CLI are staff-only switch-on runs, covered by §2 above. |
| `docs/testing.md` | Deployment-compatibility tests must cover the old shape | Pass | `registry-resources-download.test.ts` keeps its "still serves the pre-cutover presentation digest a drained run context asks for" case, unchanged. |
| `docs/testing.md` | Integration tests only, at entry points; mock at the boundary | Pass | The only test edit is two constants inside the existing `resolvePrivateRegistryResourceArchive` cases. No new mocks, no `vi.mock`, no fetch stub. |
| `REVIEW.md` | `feat:` requires integration tests | Pass — see note | No new test is added, and none is owed: the diff adds no code path. The behaviour these constants drive is already covered, and that coverage **moves with the pins** — the "current registry digest" case now asserts the `71ff2fb` schoolhouse digest resolves to its new version id. A test asserting a constant equals itself would be tautological. |
| `.claude/skills/ccstate/SKILL.md` | `detach()`/`Reason.DomCallback`; pattern forbidden in `signals/` | N/A | No React view, DOM handler, signal, or command touched. |

### Pin verification

The real risk in a constants-only PR is a pin that points at nothing, which type-checking and lint cannot catch. Checked four ways against the branch head:

1. **Registry digests** — all 23 equal `sha256(<slug>.tar.gz)` for tarballs rebuilt from `71ff2fb`. Mismatches: **0**.
2. **Version ids** — all 23 equal `sha256("storage:<storageId>\n" + sorted "<path>:<sha256>" lines)` recomputed from the extracted trees. Mismatches: **0**.
3. **Production resolvability** — all 23 pinned version ids exist as `storage_versions` rows in the production database: **23 / 23**. This is the check that catches a correct-looking digest computed against the wrong storage.
4. **No stale values** — none of the superseded `a4cf54d` digests or version ids survives in either file.

### Testing

- Coverage: **Adequate.** The changed constants are exercised by the existing `resolvePrivateRegistryResourceArchive` cases, which move with the pins. The pre-cutover case is deliberately unchanged and remains the regression evidence for the switch-off path.
- Conventions: **Compliant.** No new mocks, no fake timers, no mock-invocation assertions.
- Local: `pnpm check-types` 14/14, `pnpm turbo run lint --concurrency=1` (3072 MiB heap) 13/13, `pnpm -F @okouai/core exec vitest run` 205 passed. API tests run in CI.

### Note on PR metadata

The title said `a4cf54d` while the head commit pinned `71ff2fb`; it has been corrected to `71ff2fb`, and the body now describes both upstream batches. Worth flagging because the pins, not the title, are what ships — but the mismatch was in the description only.

Issue #28580, opened 2026-08-21
refactor(billing): remove legacy credit note refund fallback


hulh122:
## Context

PR #28578 adds temporary recovery for invoice-backed usage-pack credit refund rows written by the previous API version. That version could persist an issued Stripe Credit Note before the linked Refund had succeeded; the new reconciler observes the historical Refund once and, after a failed/canceled result, creates a standalone replacement Refund on a later attempt.

Surface/window: backend persisted state during API rollout. Keep the compatibility path through the documented approximately 102-minute maximum exposure and until all legacy rows have been reconciled or manually resolved.

## Removal condition

- The API version that issued Credit Notes before successful Refunds is outside the production drain and rollback window.
- A production-safe query confirms there are zero unresolved invoice refund rows with `stripe_credit_note_id IS NOT NULL` and `status <> 'succeeded'`.

## Acceptance criteria

- Remove the legacy `existingCreditNote` recovery branch from `processInvoiceRefund`.
- Remove the failed-row compatibility selection in `retryableFailedInvoiceRefundCondition` when it is no longer needed by current-version rows.
- Remove the legacy recovery regression test and any test-only state fields used only by that path.
- Preserve the current Refund-before-Credit-Note flow and its three-attempt retry behavior.
- Record the production query evidence and the deployed-version drain evidence in the cleanup PR.

Blocked by #28578 deployment and the removal conditions above.

Issue #28449, opened 2026-08-20
chore: remove branded artifact rollout fallbacks


hulh122:
## Context

The hosted-site and artifact-CDN brand rollout temporarily supported API/config ordering, database writer ordering, and pre-brand persisted data. Completed rollout compatibility has been removed; only evidence-gated historical-data fallbacks remain.

## Status — 2026-08-24

PR #28511 merged on 2026-08-21 and is on `main`. It completed the configuration, schema, and writer-order cleanup described below. This issue now tracks only persisted-history compatibility whose safe removal cannot yet be proven.

### Completed by #28511

- [x] Require `OKOU_PUBLIC_ARTIFACTS_BASE_URL` and remove `publicArtifactsBaseUrlForBrand`'s fallback to the VM0 CDN.
- [x] Require `OKOU_PUBLIC_HOST_DOMAIN` in the API and remove hosted-site and artifact-preview fallback to the retired API-side `OKOU_HOST_DOMAIN` setting.
- [x] Stop injecting the retired API-side `OKOU_HOST_DOMAIN` deployment variable; the Host Worker variable with the same name remains in use.
- [x] Remove the browser-session schema probe, JSON-shape read, conditional persistence/retry, and pre-migration insert shape after migration 0956 and the rollback window elapsed.
- [x] Remove temporary VM0-CDN and browser public-brand rollout tests, routes, and fixtures.
- [x] Remove temporary `public_brand` database defaults from `browser_sessions`, `hosted_sites`, and `hosted_deployments` with migration 0959. Current writers persist the trusted creation brand explicitly.

Production has dedicated VM0/Okou host and CDN configuration, and live Okou hosted-site and artifact output has returned HTTP 200.

### Remaining compatibility

- [ ] Keep missing-brand fallback reads for V2 artifact object metadata until every reachable pre-brand object is inventoried and migrated or deleted.
- [ ] Keep V1 artifact handling while legacy V1 objects remain reachable.
- [ ] Keep `canonicalAssetPublicBrand` missing-brand fallback until pre-brand canonical rows expire, are backfilled, or are proven unreachable.
- [ ] Keep `builtInGenerationPublicBrand` missing-brand fallback until pre-brand jobs are reconciled and no active job can still lack request `publicBrand`.
- [ ] Keep the Host Worker behavior that treats brandless historical pointers/manifests as VM0 until those reachable objects are inventoried and migrated or intentionally retained.

### Current blockers

[...]

Issue #28303, opened 2026-08-20
chore(teams): remove the legacy oauth redirect uri fallback


lancy:
Follow-up to #28300 / #28302.

## What to remove

`apps/api/src/signals/routes/teams-oauth.ts` carries a rollout fallback so that a Microsoft Teams authorization started before #28302 deployed can still complete:

```ts
redirectUri:
  state.redirectUri ??
  legacyCallbackRedirectUri(getOAuthApiOrigin(request)),
```

`legacyCallbackRedirectUri()` returns the unprojected legacy value the previous build sent for both brands: `${getOAuthApiOrigin(request)}/api/zero/teams/oauth/callback`. Microsoft rejects a token request whose `redirect_uri` differs from the authorization request, so without it an authorization spanning the deploy fails with the authorization code already consumed.

**Only Okou-brand authorizations depend on it.** Per the correction on #28300, the VM0 brand keeps emitting `/api/zero/teams/oauth/callback`, so its pre- and post-#28302 values are identical.

## Why it is not permanent

Per `docs/fallback.md` section 8, a rollout fallback ships with its removal planned. The surface is a browser-held Microsoft consent page, bounded by the old web/app client window of ~2 days; the authorization request is not persisted and carries no expiry of our own, so nothing shortens that window.

## Removal condition

Both must hold:

- no Okou-brand authorization started before #28302 deployed can still be presented (~2 days after that deploy);
- the pre-#28302 build is outside the production rollback window. Rolling back re-introduces a build that recomputes the redirect URI, so a rollback during the window breaks in-flight Okou-brand authorizations in the reverse direction regardless of this fallback.

## Scope

Delete together:

- `legacyCallbackRedirectUri()`;
- the `?? legacyCallbackRedirectUri(...)` branch, leaving `state.redirectUri`;
- the `redirectUri: null` handling in `parseOAuthState()`, making the field required;
- the test `falls back to the legacy callback URI for state without a redirect URI`.

The test dies with the fallback it covers, so removing it is not a coverage loss.

## Do not remove here

`/api/zero/teams/oauth/callback` must keep serving and must stay registered in the Microsoft app registration. After #28302 it is the **live producer target for the VM0 brand**, not only a drain-window acceptor.

That also constrains #26701 Phase C3: the route cannot be retired until `api.vm0.ai` stops serving the VM0 brand.

hulh122:
[...]

PR #28302, opened 2026-08-20
fix(teams): carry the oauth redirect uri in state and emit the final callback

lancy:
## Why the state fix is the primary change

`routes/teams-oauth.ts` recomputed the redirect URI at token-exchange time instead of reusing the one the authorization request actually sent. OAuth 2.0 requires the two to be identical and Microsoft enforces it, so **changing the emitted value is unsafe on its own**:

```
user starts connect   → old build emits api.vm0.ai/api/zero/teams/oauth/callback
  ...user authorizes at Microsoft, tens of seconds...
  ...new build ships in that window...
Microsoft redirects   → new build receives it, recomputes api.okou.ai/api/integrations/...
token exchange        → Microsoft rejects: redirect_uri mismatch
```

The authorization code is already consumed at that point, so the user has to start over.

The first commit removes the race; the second changes what is emitted. They are split so the ordering is visible in review.

### 1. `fix(teams): carry the oauth redirect uri in the connect state`

`connectOauth$` now records the exact redirect URI it sends in the OAuth state, and `callbackOauth$` reuses `state.redirectUri` at exchange time. This is the pattern the Feishu flow already uses (`routes/feishu-oauth.ts:848`).

State written by the previous build has no such field. It falls back to the legacy value, `${getOAuthApiOrigin(request)}/api/zero/teams/oauth/callback` — unprojected, exactly what the previous build sent — so authorizations spanning the deploy still complete. `legacyCallbackRedirectUri` carries a comment stating the fallback is removable only once no state predating this change can still be presented.

### 2. `fix(teams): emit the final teams oauth callback with brand projection`

The authorization request now emits the canonical path for the Okou brand, with the origin projected through `apiUrlForPublicBrand`.

`api.vm0.ai` and `/api/zero/**` retire together, so **the VM0 brand does not migrate its callback path** — moving it would require registering a new Redirect URI for a brand that is going away.

| Brand | redirect_uri | Registered today |
|---|---|---|
| okou | `https://api.okou.ai/api/integrations/teams/oauth/callback` | yes |
| vm0 | `https://api.vm0.ai/api/zero/teams/oauth/callback` | yes |

**Both values are already registered, so this ships with no Microsoft app-registration change.** The VM0-brand redirect URI is byte-identical to what production emits today, so the deploy window only affects Okou-brand authorizations — exactly what commit 1 protects.

[...]

PR #28737, opened 2026-08-23

e7h4n:
Daily deployment-compatibility cleanup. One cohort survived evidence collection.

## Cohort — remove the legacy Teams OAuth redirect URI fallback (#28303)

**Old → new boundary.** [PR #28302](https://github.com/vm0-ai/vm0/pull/28302) started carrying the authorization request's `redirectUri` in the Teams OAuth state, so the token exchange repeats the exact value the authorization sent instead of recomputing one that had moved for the Okou brand. State written by the previous build carried no redirect URI, so the callback fell back to `legacyCallbackRedirectUri()` — the unprojected `${origin}/api/zero/teams/oauth/callback` that build sent for **both** brands. Microsoft rejects a token request whose `redirect_uri` differs from the authorization request, so without the fallback an authorization spanning the deploy would have failed with its code already consumed.

**Window.** #28303 states both conditions: no Okou-brand authorization started before #28302 deployed can still be presented (~2 days after that deploy), and the pre-#28302 build is outside the production rollback window. The surface is a **browser-held Microsoft consent page** — the authorization request is not persisted anywhere and carries no expiry of ours, so nothing shortens the ~2-day client window and nothing lengthens it either.

**Rollout evidence.** #28302 merged 2026-08-20T04:42:50Z (`f511cd0cb0`); `api/production` promoted `f71f54850ab2` at **2026-08-20T07:09:44Z**, verified to contain that commit with `git merge-base --is-ancestor`. **Seventeen further `api/production` promotions** have landed since, so the pre-#28302 build is far outside any rollback window. At this run the elapsed time is **86 hours (3.6 days)** against a ~2-day window.

Only Okou-brand authorizations ever depended on the fallback: per #28300's correction the VM0 brand keeps emitting `/api/zero/teams/oauth/callback`, so its pre- and post-#28302 values are identical.

**Removed**, exactly the scope #28303 enumerates: `legacyCallbackRedirectUri()`; the `?? legacyCallbackRedirectUri(...)` branch, leaving `state.redirectUri`; the `redirectUri: null` handling in `parseOAuthState()`, making the field required; and the `falls back to the legacy callback URI for state without a redirect URI` test, which dies with the fallback it covers.

**Kept deliberately.** `/api/zero/teams/oauth/callback` still serves and stays registered in the Microsoft app registration — after #28302 it is the live producer target for the VM0 brand, not merely a drain-window acceptor. #28303 calls this out and it also constrains #26701 Phase C3.

**Producer proof.** `connectOauth$` builds `stateObj` with `redirectUri: string` as a **required** property and always assigns `callbackRedirectUri(origin, publicBrand)`. No current code path can mint a Teams OAuth state without it, so requiring the field on read cannot reject a state this API produced.

**One consequence worth stating plainly.** `parseOAuthState`'s two degenerate branches — absent `state`, and `state` that is not a JSON object — previously returned a synthetic `OAuthState` with `publicBrand: "vm0"` and `redirectUri: null`; they now return `null`. Those branches already produced no `orgId` and no `userId`, so the callback rejected them with `Invalid connect state.` before reaching the token exchange, and that outcome is unchanged. The one observable difference is the brand used for that error redirect: it now comes from `publicBrand$` (the request's own brand) instead of a hardcoded `"vm0"`. On `api.vm0.ai` those are the same value; on `api.okou.ai` an unusable state now redirects to the Okou settings page rather than the VM0 one, which is the correct destination for that caller.

**Test fixtures corrected rather than relaxed.** Three callback fixtures predated #28302 and omitted `redirectUri`. Two of them started failing outright; the third — `rejects callback state when the internal user is not an org member` — still passed but would have passed **for the wrong reason**, short-circuiting on a malformed state before the membership check ran. All three now carry the redirect URI, so they keep exercising the behavior they were written for. That silent-coverage-loss case is the reason the fixtures were fixed instead of the read being loosened.

**Persisted-data reasoning.** None applies: OAuth state is a query parameter round-tripped through Microsoft, never stored.

## Checks

Run once against the combined diff, with a local PostgreSQL 18 cluster created for this run and set to `UTC`:

- `pnpm format`; `pnpm check-types` (all tasks, no errors); `pnpm knip` (exit 0); `pnpm -F api lint` (exit 0)
- `pnpm -F api exec vitest run teams-oauth.test.ts` → **9 passed**
- `pnpm -F api exec vitest run slack-oauth.test.ts` → **44 passed** (the sibling OAuth-state surface, unchanged by this PR)

## Open-PR overlap

**None.** No open PR touches either changed file.

## Deferred, with exact blockers

The masked production database remains the dominant blocker and is **unchanged since 2026-08-22**: `vm0-prod-ro` still reports 122 tables and still omits `usage_pack_credit_refunds`, `usage_pack_subscriptions`, `connector_catalog_runtime_projections`, and every `browser_*` table. This is not only this workflow's problem — #28449's own status section independently records that "the available read-only production data source does not expose artifact or hosted-site object metadata and masks canonical asset metadata".

- **#28580** legacy credit-note refund fallback — needs a production query proving zero unresolved invoice refund rows with `stripe_credit_note_id IS NOT NULL AND status <> 'succeeded'`. `usage_pack_credit_refunds` is not exposed.
- **#28275** connector catalog projection probe — needs migrations 0963/0964 outside the rollback window *and* the projection table observable; it is not exposed.
- **`memberInviteUsagePackEntitlementSchemaAvailable`** — migration 0899 shipped 2026-08-11 and roughly seventy migrations have followed, but every reader of `member_invite_usage_pack_required` is guarded, so no production request fails observably if it were missing and MaskDB cannot confirm it exists. No positive evidence is obtainable.
- **`browser_*` identity contraction** — still needs `chat_thread_id` uniqueness across retained rows.
- **#28449** three retained items — the issue owner has already recorded that production still holds pre-brand `running` jobs and that object metadata is masked.
- **#27356** zero-agent default-avatar bridge — window long closed, but `apps/api/src/signals/routes/agents.ts:296` still reads `avatarUrl: args.body.avatarUrl ?? null` while only the create path uses `randomPresetAvatar()`. The upsert reaches its INSERT branch from the PUT route for a compose-backed agent with no `zero_agents` row, so the trigger compensates a **live** writer. **Ninth consecutive run reporting this; the unblocking change is one line.**
- **#27786** (and the same pattern on `selectedVideoModel`) — "make it required" is blocked by the producer, not the window: GET and PUT share `userModelPreferenceResponseSchema` and the PUT producer spreads the field conditionally, so requiring it would break the API's own PUT responses.
- **#26842** — closed as completed but its code remains; `supportsInAppPreview` appears at seven contract sites and the current app omits it by design when preview is disabled, so narrowing it is a product change, not compatibility removal.
- **#27660's OAuth half** — unlike this PR's `redirectUri` field, that shim is not separable: its ternary collapses *absent* and *malformed* into one branch.
- **#26938**, **#26672 / #26519** (`LatestWebsiteTemplates` still `enabled: false`), **#26701 / #26364**, **#25620** — multi-stage programs, product gates, or backfill prerequisites.

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/teams-oauth.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/teams-oauth.test.ts
@@ -24,6 +24,7 @@ const context = testContext();
 const mocks = createRouteMocks(context);
 const store = createStore();
 const API_ORIGIN = "https://api.vm0.ai";
+const CALLBACK_REDIRECT_URI = `${API_ORIGIN}/api/zero/teams/oauth/callback`;
 const OKOU_API_ORIGIN = "https://api.okou.ai";
 const WEB_ORIGIN = "https://www.vm0.ai";
 const APP_ORIGIN = "https://app.vm0.test";
@@ -271,7 +272,11 @@ describe("Teams OAuth API routes", () => {
     const response = await appRequest(
       callbackPath({
         code: "valid-code",
-        state: { orgId: fixture.orgId, userId: fixture.userId },
+        state: {
+          orgId: fixture.orgId,
+          userId: fixture.userId,
+          redirectUri: CALLBACK_REDIRECT_URI,
+        },
       }),
       { origin: API_ORIGIN },
     );
@@ -330,31 +335,6 @@ describe("Teams OAuth API routes", () => {
     );
   });
 
-  it("falls back to the legacy callback URI for state without a redirect URI", async () => {
-    const fixture = await seedTeamsInstallation(track);
-    await seedMembership(fixture.orgId, fixture.userId, "admin");
-    mocks.clerk.session(fixture.userId, fixture.orgId, "org:admin");
-    mockMicrosoftOAuth({
-      tenantId: fixture.teamsTenantId,
-      aadObjectId: fixture.teamsAadObjectId,
-      userPrincipalName: fixture.teamsUserPrincipalName,
-      expectedRedirectUri: `${API_ORIGIN}/api/zero/teams/oauth/callback`,
-    });
-
-    const response = await appRequest(
-      callbackPath({
-        code: "valid-code",
-        state: { orgId: fixture.orgId, userId: fixture.userId },
-      }),
-      { origin: API_ORIGIN },
-    );
-
-    expect(response.status).toBe(307);
-    expect(response.headers.get("location")).toContain(
-      `${APP_ORIGIN}/settings/teams?status=connected`,
-    );
-  });
-
   it("rejects OAuth users when the org is already bound to another Microsoft tenant", async () => {
     const fixture = await seedTeamsInstallation(track);
     await seedMembership(fixture.orgId, fixture.userId, "admin");
@@ -383,7 +363,11 @@ describe("Teams OAuth API routes", () => {
     const response = await appRequest(
       callbackPath({
         code: "valid-code",
-        state: { orgId: fixture.orgId, userId: fixture.userId },
+        state: {
+          orgId: fixture.orgId,
+          userId: fixture.userId,
+          redirectUri: CALLBACK_REDIRECT_URI,
+        },
       }),
       { origin: API_ORIGIN },
     );
@@ -405,7 +389,11 @@ describe("Teams OAuth API routes", () => {
     const response = await appRequest(
       callbackPath({
         code: "valid-code",
-        state: { orgId: fixture.orgId, userId: fixture.userId },
+        state: {
+          orgId: fixture.orgId,
+          userId: fixture.userId,
+          redirectUri: CALLBACK_REDIRECT_URI,
+        },
       }),
       { origin: API_ORIGIN },
     );
--- a/turbo/apps/api/src/signals/routes/teams-oauth.ts
+++ b/turbo/apps/api/src/signals/routes/teams-oauth.ts
@@ -43,7 +43,7 @@ interface OAuthState {
   readonly userId: string | null;
   readonly prompt: string | null;
   readonly publicBrand: PublicBrand;
-  readonly redirectUri: string | null;
+  readonly redirectUri: string;
 }
 
 interface MicrosoftTeamsUserInfo {
@@ -139,27 +139,19 @@ function optionalString(value: unknown): string | null {
 
 function parseOAuthState(state: string | undefined): OAuthState | null {
   if (!state) {
-    return {
-      orgId: null,
-      userId: null,
-      prompt: null,
-      publicBrand: "vm0",
-      redirectUri: null,
-    };
+    return null;
   }
 
   const parsed = safeJsonParse(state);
   if (typeof parsed !== "object" || parsed === null) {
-    return {
-      orgId: null,
-      userId: null,
-      prompt: null,
-      publicBrand: "vm0",
-      redirectUri: null,
-    };
+    return null;
   }
 
   const record = parsed as Record<string, unknown>;
+  const redirectUri = optionalString(record.redirectUri);
+  if (!redirectUri) {
+    return null;
+  }
   return {
     orgId: optionalString(record.orgId),
     userId: optionalString(record.userId),
@@ -167,7 +159,7 @@ function parseOAuthState(state: string | undefined): OAuthState | null {
     // Old web/app OAuth state can omit publicBrand for about two days.
     // Remove this VM0 default in #27660 after legacy states have drained.
     publicBrand: record.publicBrand === "okou" ? "okou" : "vm0",
-    redirectUri: optionalString(record.redirectUri),
+    redirectUri,
   };
 }
 
@@ -185,29 +177,6 @@ function callbackRedirectUri(origin: string, publicBrand: PublicBrand): string {
   return publicBrand === "okou"
     ? `${apiUrlForPublicBrand(origin, "okou")}/api/integrations/teams/oauth/callback`
     : `${apiUrlForPublicBrand(origin, "vm0")}/api/zero/teams/oauth/callback`;
-}
-
-/**
- * Rollout fallback for #28300. State written by the previous API build carries
- * no redirect URI, and that build sent this exact value for both brands: the
- * unprojected origin plus the legacy path. Microsoft rejects a token request
- * whose redirect_uri differs from the authorization request, so an
- * authorization started before this deploy can only be completed against it.
- *
- * Only Okou-brand authorizations actually depend on this. The VM0 brand keeps
- * emitting the legacy path above, so its old and new values are identical.
- *
- * Surface: a browser-held Microsoft consent page, so it is bounded by the old
- * web/app client window of ~2 days rather than by the API deploy gap. The
- * authorization request is not persisted anywhere and carries no expiry of our
- * own, so nothing shortens that window.
- *
- * Removable once no authorization started before this deploy can still be
- * presented, and the previous build is outside the production rollback window.
- * Follow-up: #28303.
- */
-function legacyCallbackRedirectUri(origin: string): string {
-  return `${origin}/api/zero/teams/oauth/callback`;
 }
 
 function microsoftCredentials(): {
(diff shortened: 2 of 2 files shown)


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: f6515d2ad580f660d1c89e78e4cdaad30db0e802
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Removes the `#28300` rollout fallback in the Microsoft Teams OAuth callback: `legacyCallbackRedirectUri()`, the `?? legacyCallbackRedirectUri(...)` branch in `callbackOauth$`, and the `redirectUri: string | null` tolerance in `parseOAuthState()` (the field is now required on read). The fallback's own test is deleted with it, and three fixtures that predated #28302 are corrected to carry `redirectUri`. Rollout evidence is sound: #28302 merged 2026-08-20T04:42:50Z, `api/production` promoted a build containing it at 07:09:44Z, seventeen promotions have followed, and the elapsed 86 hours exceeds the ~2-day browser-held-consent-page window that bounds this surface.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic — no unhandled edge cases or wrong assumptions introduced | Pass | `teams-oauth.ts:parseOAuthState` now returns `null` for absent, non-object, and redirect-URI-less state. All three already yielded no `orgId`/`userId`, so `callbackOauth$` rejected them at `if (!state?.orgId \|\| !state.userId)` before the token exchange; the rejection outcome is unchanged. |
| `REVIEW.md` | Security — no auth bypass, injection, or hardcoded secrets | Pass | The read is strictly narrowed, never loosened. `state.redirectUri` is only sent as a `redirect_uri` form field to Microsoft's token endpoint, which rejects any value that does not match the authorization request; it is never used to redirect the browser. |
| `REVIEW.md` | Type safety — no `any`, no overly broad types | Pass | The change tightens `OAuthState.redirectUri` from `string \| null` to `string`. No casts added; `record.redirectUri` still flows through `optionalString()` before the narrowing check. |
| `REVIEW.md` | Interface changes — breaking changes identified | Pass | `OAuthState` is a file-local interface with a single producer (`connectOauth$`) and a single consumer (`callbackOauth$`). No exported contract changes; `teamsOauthContract` is untouched. |
| `REVIEW.md` | Testing coverage for a `refactor:` commit — existing tests still cover the code | Pass | 9 tests remain in `teams-oauth.test.ts`, including `exchanges the code with the redirect URI recorded in the connect state`, which is the direct positive coverage for the surviving path. The deleted test covered only the removed fallback. |
| `REVIEW.md` | Feature regression guard — replaced behavior carries explicit coverage | Pass | No feature is replaced or rerouted. `/api/zero/teams/oauth/callback` still serves (`leaves the VM0 brand authorization URI on the registered legacy value` covers the VM0 producer target). |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions, no dynamic imports | Pass | Diff adds one `const CALLBACK_REDIRECT_URI` and one guard clause; no suppression comments, no `import()`. |
| `docs/bad-smell.md` | §7 fail fast — no fallback/recovery logic hiding a problem | Pass | This PR deletes a fallback rather than adding one; the degenerate branches now fail through the existing `Invalid connect state.` path instead of synthesizing an `OAuthState`. |
| `docs/bad-smell.md` | §6 no hardcoded URLs in production code | Pass | `legacyCallbackRedirectUri()`'s hardcoded `${origin}/api/zero/...` template is removed. `callbackRedirectUri()` keeps its brand projection through `apiUrlForPublicBrand`. The new hardcoded URI is test-only (`CALLBACK_REDIRECT_URI`). |
| `docs/fallback.md` | §10 a removal PR carries type, single-writer, production, or rollout evidence | Pass | Two independent kinds: single-writer (`connectOauth$` declares `redirectUri: string` as required in `stateObj` and always assigns `callbackRedirectUri(origin, publicBrand)`) and rollout (86h elapsed vs a ~2-day window, plus 17 `api/production` promotions past the pre-#28302 build). Both are recorded in the PR body. |
| `docs/fallback.md` | §10 the removal deletes the branch, its contract entry, and its own tests together | Pass | `legacyCallbackRedirectUri()`, the `??` branch, the nullable field, and `falls back to the legacy callback URI for state without a redirect URI` all go in one commit. `grep -rn legacyCallbackRedirectUri` over the tree returns nothing. |
| `docs/fallback.md` | §1 no negative tests asserting removed behavior stays removed | Pass | No tombstone was added in place of the deleted test. |
| `docs/fallback.md` | §9 every fallback introduced by the diff is declared in the PR summary | N/A | The diff introduces no fallback. The `publicBrand === "okou" ? ... : "vm0"` default in `parseOAuthState` is pre-existing, already carries its own `#27660` removal comment, and is untouched by this diff — so per §9 no `Fallbacks` section is required. |
| `docs/deployment-compatibility.md` | §7 rollout fallback sized against the correct surface's observed maximum exposure | Pass | The surface is a browser-held Microsoft consent page, correctly sized to the ~2-day old-client window rather than the API deploy gap. The authorization request is not persisted, so nothing extends it. |
| `docs/deployment-compatibility.md` | Clean-up phase runs only after the old deployed version is no longer active | Pass | The pre-#28302 build is outside the production rollback window after 17 subsequent promotions; verified by `git merge-base --is-ancestor` against the promoted commit. |
| `docs/deployment-compatibility.md` | Persisted-state compatibility for cross-version reads | N/A | Teams OAuth state is a query parameter round-tripped through Microsoft. Nothing is written to the database, so neither schema direction applies. |
| `docs/testing.md` | Integration tests only; no unit tests for internal functions | Pass | All 9 tests drive the real Hono app through `createAppWithRoutes`/`appRequest`. `parseOAuthState` is never imported directly. |
| `docs/testing.md` | No fake timers, no `toHaveBeenCalled()` as the sole assertion | Pass | The touched tests assert on `response.status` and the `location` header. No timer manipulation in the diff. |
| `docs/testing/api-testing.md` | Route behavior covered through the real app and the route's contract, mocking only external services | Pass | `mockMicrosoftOAuth()` is MSW at the Microsoft token and Graph endpoints; identity comes from `mocks.clerk.session`. Both are external boundaries. |
| `docs/testing/api-testing.md` | Uniquely owned identities so a test is correct under concurrent execution | Pass | Each touched test builds its own fixture via `seedTeamsInstallation(track)` and seeds its own membership; adding `redirectUri` does not introduce shared state. |
| `docs/testing/anti-patterns.md` | AP-2/AP-4 — no direct `fetch` mocking, no `vi.mock()` on relative paths | Pass | No `vi.mock()` and no `vi.stubGlobal("fetch", ...)` anywhere in the diff; HTTP stays on MSW `server.use`. |
| `docs/testing/patterns.md` | Pattern 3 — MSW handlers overridden per test, global setup owns the server lifecycle | Pass | The touched tests reuse the file's existing `mockMicrosoftOAuth()` helper; no `server.listen`/`close` added. |
| `docs/testing/testing-external-behavior.md` | API tests construct and verify cases through the endpoint, not the DB or services | Pass | The three corrected fixtures still enter via the callback URL and assert on the HTTP redirect. Changing the state's `redirectUri` moves the fixture closer to what the real producer emits, not further into internals. |
| `.claude/skills/ccstate/SKILL.md` | Commands take an `AbortSignal` and pass it through; `throwIfAborted()` after awaits that do not receive it | Pass | `callbackOauth$` keeps its `signal` parameter and every existing `signal.throwIfAborted()` call site. The diff only replaces the `redirectUri` argument passed into `exchangeMicrosoftTeamsOAuthCode`. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Notes (non-blocking)

- `parseOAuthState`'s two degenerate branches now return `null`, so the error redirect's brand comes from `publicBrand$` instead of a hardcoded `"vm0"`. The rejection itself is unchanged, and on `api.okou.ai` this sends the caller to the Okou settings page, which is the correct destination. That brand-selection difference has no direct test, but it had none before this PR either, and a test asserting that a redirect-URI-less state is rejected would be the tombstone `docs/fallback.md` §1 prohibits. Not a coverage regression.
- Correcting the three pre-#28302 fixtures rather than loosening the read is the right call. `rejects callback state when the internal user is not an org member` was passing for the wrong reason — it short-circuited on a malformed state before the membership check ran — so relaxing `parseOAuthState` would have preserved a silent coverage loss.

### Testing

- Coverage: Adequate. This is a `refactor:` commit; the surviving path keeps its positive test (`exchanges the code with the redirect URI recorded in the connect state`), and the deleted test covered only the fallback being removed, so it dies with it rather than becoming a coverage gap.
- Conventions: Compliant. No internal mocks, no fake timers, no direct DB access or service imports, no `toHaveBeenCalled()`-only assertions, and no tombstone test replacing the deleted one.

PR #28738, opened 2026-08-23
chore: republish presentation template archives from 71ff2fb

bingjiez666:
> **Do not merge.** The Actions run is the audit record, same as #28724, #28646 and #26507.

## What

Republishes all 23 built-in presentation runbook archives from `vm0-ai/Template-artifact@71ff2fb`, superseding the `a4cf54d` set published in #28724.

Each archive lands as a new version on the existing storage. The pre-cutover (deck-JSON renderer) versions are untouched — they are still the switch-off path and the drained-CLI fallback.

## What changed upstream

Two commits (PR #34), and the diff touches **only the 23 `SKILL.md` files** (+4/−376): the "Check what the report cannot see" section is dropped, so the finished deck is no longer re-checked by hand now that the automated report is declared to be the visual review.

Package shape is unchanged, re-checked on all 23: `SKILL.md`, `design-system.md`, `layouts/_shell.html`, `decoration/`, `example/`, `tools/`, and 19 `color-systems/*.css` each. The templates still contain no `okou` or image-batch instruction, so the prompt shipped in #28702 stays correct.

## Verification

`build` recomputes both pins from source and refuses to write an artifact that disagrees; `verify` re-extracts the committed tarballs and recomputes the manifest, version id and digest from the extracted bytes; `publish` re-reads the new **and** the old object back out of R2 and checks both digests before touching the database.

`inspectArchive` also re-asserts the structural contract on every archive: `SKILL.md` / `design-system.md` / `layouts/_shell.html` / the default colour CSS present and non-empty, and every `color-systems/*.css` reference resolving relative to the referring file to a file the archive carries.

The `oldVersionId` pins were cross-checked against production before building: all 23 storages currently head at the `a4cf54d` versions, confirming the previous publish landed and that this run supersedes exactly that set.

Local run against the committed script and tarballs:

```
```

## Follow-up

Registry pins move in #28728, which is updated in place rather than superseded by a third PR.


bingjiez666:
Published. 23/23 archives written to R2 and registered in the production database, verified independently of this run's log by reading the production database.

Registry pins move in #28728 (`e803500`).

Closing without merging, per the convention this follows (#28724, #28646, #26507).

PR #28747, opened 2026-08-24
test(mitm-addon): verify log timestamps use current utc clock

### LINKED ISSUE #28685: test(python): verify canonical log timestamps track the current UTC clock
## Problem

`crates/runner/mitm-addon/tests/timestamp_helpers.py:9-13` validates only the shape and UTC offset of a log timestamp. Its callers in `test_logging_utils.py`, `test_tcp_hooks.py`, `test_error_handler.py`, and `test_response_handler_logging.py` therefore do not verify the time source used by `crates/runner/mitm-addon/src/logging_utils.py:52-53`.

As a mutation check, replacing `_utc_log_timestamp()` at runtime with the constant `2000-01-01T00:00:00.000Z` still leaves all 159 tests across those four suites passing. A regression that emits a well-formed but stale constant timestamp would therefore remain undetected.

## Impact

Canonical network and proxy log timestamps drive event ordering, correlation, retention, and incident timelines. Syntactically valid stale timestamps can silently move current events into the wrong time window while the existing tests remain green.

## Acceptance criteria

- Add a deterministic test that proves `_utc_log_timestamp()` derives its value from the current UTC clock, rather than checking only its string format.
- Exercise at least one public logging path and verify the emitted timestamp corresponds to a controlled or explicitly bounded current time.
- Preserve the millisecond-precision `Z` format and avoid sleeps or narrow real-time assertions that could make the test flaky.

seven332:
## Summary

- strengthen the existing public network-log integration test with a controlled UTC clock
- verify the persisted timestamp uses the controlled current instant and preserves millisecond `Z` formatting
- prove the reported well-formed stale timestamp mutation is rejected without adding sleeps or another filesystem test

## Exclusions

- no production logging behavior, dependency, API, protocol, persisted state, or deployment compatibility changes
- the shared shape-only assertion remains for broad format coverage in other logging paths

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_logging_utils.py::TestLogNetworkEntry::test_writes_jsonl_with_current_utc_timestamp` (1 passed)
- in-memory stale timestamp mutation check (expected test failure observed)
- `uv run --no-sync python -m pytest tests/` (6517 passed)

Closes #28685

### DIFF
--- a/crates/runner/mitm-addon/tests/test_logging_utils.py
+++ b/crates/runner/mitm-addon/tests/test_logging_utils.py
@@ -1,6 +1,7 @@
 """Tests for mitm addon logging utilities."""
 
 import json
+from datetime import UTC, datetime
 from unittest.mock import MagicMock, patch
 
 import pytest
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #28747

**Reviewed head:** `c7d49491120101b6f1f8d1d4735d772dec03fb5c`

### Summary

The test-only change deterministically verifies the public network logger derives its canonical timestamp from a controlled UTC clock while retaining real formatting, JSONL IO, and async flush behavior. It adds no production abstraction, dependency, sleep, or duplicate test case.

### Findings

No findings.

### Testing

Lock/sync, Ruff format and lint, BasedPyright, the focused normal test, the stale-constant mutation check, and the full mitm-addon suite all passed locally (6517 tests). Unrelated Turbo and Rust checks were not run because the diff affects only mitm-addon Python tests; PR CI remains the external verification surface.

### Verdict

LGTM

PR #28749, opened 2026-08-24
test(python): cover requestheaders auth application rollback

### LINKED ISSUE #28686: test(python): cover requestheaders auth application rollback
## Problem

`crates/runner/mitm-addon/src/auth.py:1772-1787` catches failures that occur while applying already-resolved header, query, or AWS SigV4 auth during `requestheaders()`. The branch promises to restore the pre-probe request URL, raw header fields, and flow metadata before returning `FALLBACK`.

The current full mitm-addon branch-coverage run executes the surrounding success, auth-resolution failure, validation failure, cancellation, and authorization-revalidation paths, but does not execute this post-resolution application exception handler. `crates/runner/mitm-addon/tests/test_request_headers_firewall_auth.py` consequently has no regression test for the rollback after request mutation has begun.

A focused harness that injected partial header, URL, and metadata mutations and then raised confirmed the current implementation restores all three surfaces. That contract is important, but deleting or weakening this specific restore path would not fail the existing suite.

## Impact

If a resolved-auth application or SigV4 signing failure occurs after any request mutation, incomplete rollback can leave managed credentials, rewritten query data, a rewritten URL, or probe metadata on a flow that is explicitly returned to the normal request hook. This can change later classification or expose partial auth state despite the `FALLBACK` result.

## Acceptance criteria

- Add an async regression test for `try_apply_stream_safe_firewall_auth_for_requestheaders()` where post-resolution auth application mutates request state and then raises.
- Assert that the result is `FirewallHeaderPhaseAuthResult.FALLBACK` and that the request URL, raw header fields, and metadata exactly match their pre-probe snapshots.
- Ensure the scenario reaches the application exception handler after resolution and authorization revalidation; a resolution-time failure alone is not sufficient.
- Assert that no partially applied managed header, query value, SigV4 state, or firewall-auth metadata survives the rollback.

seven332:
## Summary

- add direct requestheaders auth-orchestration coverage for rollback after header, URL/query, and metadata mutation
- use a production-valid header/query auth configuration and the real vm0 mutation path
- verify resolution and authorization revalidation occur before fallback, then assert exact snapshot restoration

## Scope

- test-only change; no production behavior, API, persistence, migration, or rollout changes
- mocks the external firewall-auth service and injects one deterministic failure after the real mitmproxy query setter completes; vm0 application and rollback functions remain real

## Testing

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_firewall_auth_handling.py -k requestheaders_auth_application_failure_restores_probe_state`
- `uv run --no-sync python -m pytest tests/test_firewall_auth_handling.py`
- negative mutation check: the focused test failed as expected when the target restore was temporarily removed, and the production file was restored with no residual diff

Closes #28686

### DIFF
--- a/crates/runner/mitm-addon/tests/test_firewall_auth_handling.py
+++ b/crates/runner/mitm-addon/tests/test_firewall_auth_handling.py
@@ -713,6 +713,83 @@ class TestHandleFirewallRequest:
         assert flow.request.headers["x-amz-security-token"] == RESOLVED_AWS_SESSION_TOKEN
         assert "Credential=AKIDEXAMPLE/" in flow.request.headers["authorization"]
 
+    async def test_requestheaders_auth_application_failure_restores_probe_state(
+        self,
+        headers,
+        real_flow,
+        mitm_ctx,
+        tmp_path,
+    ):
+        placeholder_authorization = aws_sigv4_authorization(
+            signed_headers="host;x-amz-content-sha256;x-amz-date"
+        )
+        flow = real_flow(
+            with_response=False,
+            host=STS_HOST,
+            path="/?existing=1",
+            method="POST",
+            request_headers=headers(
+                ("Host", STS_HOST),
+                ("X-Client", "original"),
+                ("X-Amz-Date", DEFAULT_SIGV4_TIMESTAMP),
+                ("X-Amz-Content-Sha256", "UNSIGNED-PAYLOAD"),
+                ("Authorization", placeholder_authorization),
+            ),
+        )
+        flow.metadata.update(
+            {
+                metadata_keys.VM_RUN_ID: "test-run",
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #28749

**Reviewed head:** `44d0f76d38c00f31d858e812f76eb5ed0e1bf9d4`

### Summary

The test uses a production-valid header/query auth shape, runs the real vm0 application and rollback path, and injects a deterministic failure only after the real mitmproxy query setter confirms header, URL/query, and metadata mutation.

### Findings

No findings.

### Testing

Local validation passed: uv lock/sync, focused regression, 78-test firewall-auth module, Ruff format/lint, BasedPyright, and the full 6518-test addon suite. The focused regression also failed as expected when the target restore was temporarily removed. GitHub CI status and external feedback are checked separately after this review.

### Verdict

LGTM

PR #28757, opened 2026-08-24
perf(python): share responses websocket parsing

### LINKED ISSUE #28688: perf(python): share Responses WebSocket parsing with failure reporting
## Problem

At `687ef278ec5b6164e40d0e7ed48ba75e49a1b648`, trusted model-provider failure reporting adds a separate full-body selective JSON scan to every admitted OpenAI Responses WebSocket server frame on mitmproxy's synchronous hook.

- `crates/runner/mitm-addon/src/mitm_addon.py:1598-1604` sends the raw server body to `model_provider_failure.observe_websocket_server_event()` before creating the existing `OpenAIResponsesEvent` used by timing, lifecycle correlation, and usage extraction.
- `crates/runner/mitm-addon/src/model_provider_failure.py:385-411` calls `_extract_json(body)` before it knows the event type. Ordinary events such as `response.output_text.delta` fall through without clearing or disabling the flow state, so every later delta is scanned again.
- `crates/runner/mitm-addon/src/model_provider_failure.py:621-632` constructs a fresh `JsonSelectiveExtractor`, feeds the complete frame, and gives it an independent 65,536-work-unit allowance.
- `crates/runner/mitm-addon/src/usage/openai_responses.py:285-329` already owns a shared zero-or-one full parse for lifecycle and usage. Its prefix classification skips the full extractor entirely for documented non-usage events, including output deltas; terminal or unresolved frames use one combined full parse.
- `crates/runner/mitm-addon/src/model_provider_failure.py:208-223` admits this state for every billable `model-provider:*` Responses WebSocket flow after upstream authentication, independently of whether usage extraction is enabled.

This regresses the hook-level single-parse boundary established by #27720. That issue removed duplicate lifecycle and usage scans. The failure observer was added later in #28532 as a new consumer outside the shared inspection path.

## Evidence

An isolated CPython 3.12.3 benchmark used the exact provider-failure scalar fields, value-presence paths, and 65,536-unit limit. It compared one current failure scan with the existing bounded top-level `type` prefix probe for valid documented non-usage events. Values are medians of three rounds:

| Frame | Bytes | Failure full scan | Prefix probe | Ratio |
| Small `response.output_text.delta` | 59 | 71.92 us | 13.65 us | 5.3x |
| 4 KiB text delta | 4,144 | 84.43 us | 14.45 us | 5.8x |
| 100 KiB text delta | 102,448 | 433.55 us | 14.48 us | 29.9x |
| 5 MiB `response.output_item.done` | 5,242,938 | 18.74 ms | 14.72 us | 1,272.7x |

A valid 40,050-byte frame containing 20,000 ignored integer scalars took 70.69 ms in the failure extractor versus 14.79 us in the prefix probe. This remains within the parser's configured bound and demonstrates that the extra pass can consume substantial synchronous CPU even far below the WebSocket message-size limit.

These are isolated parser measurements, not end-to-end latency claims. The deterministic code path establishes that the cost is paid once per server frame while failure correlation remains unambiguous. On terminal frames, the failure scan is also followed by the existing combined lifecycle/usage full scan, recreating two independent full-body budgets for the same frame.

## Impact

Responses streams routinely contain many documented delta events. The new observer turns their prior prefix-only fast path into one fresh extractor construction and one complete JSON validation per message. A thousand small deltas add roughly 58 ms of parser CPU over the measured prefix path, while large output-item frames can block the shared synchronous hook for tens of milliseconds individually. Terminal, unknown, and failure events may additionally be scanned a second time by the existing lifecycle/usage consumer.

The per-extractor work cap prevents one parser from running without a bound, but separate consumers each receive the full allowance. It therefore does not preserve a one-budget bound for the complete server-frame hook.

## Suggested direction

Make the existing retained `OpenAIResponsesEvent` and combined server-event inspection the single parsing owner for failure reporting, lifecycle correlation, and usage extraction. Extend the shared selected-field union/result with the bounded failure code, status, error-presence, and response identity observations needed by `model_provider_failure`, then pass that observation into both state machines.

Preserve prefix resolution so a documented non-lifecycle, non-failure event such as `response.output_text.delta` performs no full-body parse. Terminal, error, and unresolved frames should use at most one union extractor and one 65,536-unit budget, regardless of how many consumers need the result.

## Acceptance criteria

- Each OpenAI Responses WebSocket server frame performs at most one work-budgeted full-body selective JSON parse across failure reporting, lifecycle correlation, and usage extraction.
- Documented non-usage/non-failure events with a resolved prefix type retain the zero-full-parse fast path.
- Terminal, error, and unresolved events share one 65,536-unit budget; exhaustion cannot trigger a second full-body attempt by another consumer.
- Preserve failure-kind classification, response-ID correlation, prewarm exclusion, exact lifecycle duplicate-key checks, usage extraction, source reporting, output timing, fail-open ambiguity, and current diagnostics.
- Add deterministic hook-level coverage that counts full extractor constructions or target-body feeds for a normal delta, terminal failure, terminal usage, and work-limit exhaustion without wall-clock assertions.
- Run the focused provider-failure, Responses event JSON, WebSocket prewarm/source/usage, and output-timing tests plus the standard mitm-addon format, lint, type, and test checks.

## Audit context

Found during the daily mitm-addon Python performance audit on commit `687ef278ec5b6164e40d0e7ed48ba75e49a1b648`.

seven332:
## Summary

- inspect each admitted Responses WebSocket server frame once, then share bounded failure, lifecycle, and usage evidence
- restore the zero-full-parse fast path for known delta events and use one shared 65,536-byte budget for terminal, error, and unknown events
- retain failure-only tracking across confirmed WebSocket upgrades while continuing to release state for invalid upgrade responses

## Safety and compatibility

- failure reporting, usage accounting, and lifecycle correlation keep independent state machines and fail-closed policies
- oversized failure codes cannot discard valid usage, and invalid usage quantities cannot discard valid failure evidence
- no API, database, persisted-state, queue payload, or runner protocol changes

## Testing

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest -q tests/` (`6522 passed`)

Closes #28688


seven332:
## Code Review: PR #28757

**Reviewed head:** `d45743c64bf6c55e4b39c99665179cfb3dab0431`

### Summary

The shared Responses WebSocket inspection keeps parsing ownership centralized while preserving independent failure, lifecycle, usage, and prewarm policies. Confirmed-upgrade state retention and consumer-specific scalar failures are covered without adding a second parse or weakening the known-delta fast path.

### Findings

No findings.

### Testing

Reviewed deterministic hook-level coverage for zero/one full-parser feeds, terminal success and failure, failure-only flows, work-limit exhaustion, and mixed failure/usage evidence. The complete mitm-addon validation passed: uv lock/sync, Ruff format/lint, BasedPyright, and `6522` pytest tests.

### Verdict

LGTM

PR #28780, opened 2026-08-24
perf(api): attribute queued promotion lock contention

### LINKED ISSUE #28764: perf(api): attribute queued-promotion admission-lock contention
## Background

#24203 targets a successful all-scenario `api_to_spawn` p90 at or below one second without moving required work past process spawn or weakening correctness. #28414 removed the organization advisory lock from the non-authoritative queue scan while preserving the authoritative lock in queued promotion.

A fixed production cohort on API `f226c0ab6ce6796b58048673a5abffd500e3ce51` and Runner `v0.171.2` still measured `api_to_spawn` around 1,631 ms p90. For agent-triggered runs, `api_dispatch_admission_lock_wait` measured approximately 143/341/643 ms p90/p95/p99. Slow agent/cold runs averaged about 142 ms of admission-lock wait versus 23 ms for the rest of that path.

This contention is material within the affected path but is not the only parent owner. Ideally removing the measured wait moved the fixed cohort's all-scenario p90 by only about 32 ms and increased the runs at or below one second from 491 to 496. A behavior change is therefore not justified until the remaining production holder is attributed.

## Problem

Final launch admission already records both organization-lock wait and held time. Production run sources use queue-on-limit admission and do not execute the separate direct-run concurrency preflight; that preflight is retained only by test fixtures.

Authoritative queued promotion still acquires the same organization advisory lock, revalidates concurrency and queue ownership, transitions the queued run, and inserts its runner job. It records neither lock wait nor lock-held time. Production data therefore cannot determine whether a final launch is waiting behind queued promotion, so another lock-scope optimization would be guesswork.

## Scope

- Add fixed, low-cardinality API dispatch timing actions for queued-promotion organization-lock wait and held time.
- Emit both timings after transaction return through the existing run-owned API dispatch telemetry path, associated with the promoted run.
- Cover the behavior through an API entry-point integration test using the real PostgreSQL database.
- Preserve all existing concurrency enforcement, queue ownership, transaction boundaries, promotion ordering, billing, authorization, idempotency, cancellation, and recovery behavior.

## Acceptance criteria

- An authoritative queued promotion emits distinct lock-wait and lock-held timings associated with the promoted run.
- The held interval covers the authoritative work through transaction return; neither timing is emitted as a transaction side effect.
- The new action names are fixed members of the existing timing contract and do not introduce unbounded dimensions.
- The existing queue/cancel/promotion API integration scenario proves the timing rows are produced without relying on sleeps or internal-function unit tests.
- Existing final-launch admission timings remain unchanged.
- No test-only preflight telemetry, lock scope, database query, queue state transition, or API behavior changes in this issue.

## Follow-up gate

After deployment, compare queued-promotion and final-launch holder frequency and duration in one revision-bounded production cohort and join the results back to `api_to_spawn`. Open a behavior-optimization follow-up only when one holder has a safe, concrete change with an estimated all-scenario gain of at least 25 ms at p90 or a meaningful increase in the fraction of runs at or below one second.

## Out of scope

- Instrumenting the test-fixture-only direct-run concurrency preflight.
- Removing or moving organization advisory locks.
- Changing concurrency limits or admission semantics.
- Runner retention, pooling, affinity, claim protocol, or queue payload changes.
- Reopening the separate catalog (#28275) or claim-transport (#27735) directions.

Parent: #24203

Prerequisite completed: #28414

seven332:
## Summary

- record queued-promotion organization admission-lock wait and held durations with fixed API dispatch actions
- flush timing immediately after durable promotion so later side-effect failures cannot suppress the attribution
- verify the run-owned rows and bounded dimensions through the existing queue/cancel/promotion API integration scenario

## Problem

Final launch admission already exposes organization-lock wait and held time, but authoritative queued promotion acquires the same lock without either boundary. Production data can show final launch waiting without identifying queued promotion as the holder.

## Solution

Promotion now uses a local API dispatch timing collector. A focused helper measures lock acquisition, and a post-transaction finalizer records held time through transaction return. Both rows are flushed only after a durable promoted result, before later realtime side effects can fail. The existing post-lock queries, state transitions, errors, and ordering remain inline and unchanged.

The new fixed actions are:

- `api_dispatch_queue_promotion_lock_wait`
- `api_dispatch_queue_promotion_lock_held`

## Explicit exclusions

- no lock scope, transaction, query, concurrency, queue, notification, or runner behavior change
- no instrumentation for the test-fixture-only direct-run concurrency preflight
- no schema, migration, API contract, queue payload, runner protocol, or frontend change

## Validation

- `pnpm format`
- `pnpm -F api lint`
- `pnpm -F api check-types`
- `pnpm -F api exec vitest run src/signals/routes/__tests__/run-lifecycle.bdd.test.ts --maxWorkers=1 --silent=passed-only` (181 passed)
- `pnpm knip`
- pre-commit full Turbo type checks and file-size checks

Closes #28764

Parent context: #24203

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
@@ -6897,6 +6897,35 @@ describe("RUN-01: admission boundaries beyond request validation", () => {
     expect(promoted.status).toBe("pending");
     const drained = await waitForRunQueueLength(api, actor, 0);
     expect(drained.body.queue).toHaveLength(0);
+    const promotionTimingEvents = [
+      "api_dispatch_queue_promotion_lock_wait",
+      "api_dispatch_queue_promotion_lock_held",
+    ].flatMap((actionType) => {
+      const events = sandboxOperationEventsForRunByAction(
+        third.runId,
+        actionType,
+      );
+      expect(events).toStrictEqual([
+        expect.objectContaining({
+          source: "api",
+          op_type: actionType,
+          sandbox_type: "runner",
+          duration_ms: expect.any(Number),
+          success: true,
+          runner_group: runnerGroup,
+          profile: "vm0/default",
+          dispatch_path: "direct",
+          span_kind: "nested",
+          activation_origin: "promotion",
+        }),
+      ]);
+      expect(Number(events[0]?.duration_ms)).toBeGreaterThanOrEqual(0);
+      return events;
+    });
+    expectApiDispatchTimingEventsNotToLeak(promotionTimingEvents, [
+      "queued run three",
+      agentId,
+    ]);
     const promotedStorageState = await readRunnerJobStorageState(
       context,
       third.runId,
--- a/turbo/apps/api/src/signals/services/api-dispatch-timing.service.ts
+++ b/turbo/apps/api/src/signals/services/api-dispatch-timing.service.ts
@@ -206,6 +206,8 @@ export type ApiDispatchTimingActionType =
   | "api_dispatch_check_concurrency_limit"
   | "api_dispatch_concurrency_preflight_lock_wait"
   | "api_dispatch_concurrency_preflight_check"
+  | "api_dispatch_queue_promotion_lock_wait"
+  | "api_dispatch_queue_promotion_lock_held"
   | "api_dispatch_resolve_queue_first_admission"
   | "api_dispatch_claim_queue_first_message"
   | "api_dispatch_resolve_queue_first_claim_snapshot"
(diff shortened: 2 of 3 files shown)


seven332:
## Code Review: PR #28780

**Reviewed head:** `f8d78353d5d88eb61d333c53c4c33a52603898a5`

### Summary

The queued-promotion advisory-lock wait and held boundaries are correctly attributed after transaction durability, with the existing promotion state machine and lock scope preserved.

### Findings

No findings. Self-review identified and fixed the earlier post-commit side-effect ordering risk and unnecessary business-logic extraction before this head.

### Testing

The full real-database `run-lifecycle.bdd.test.ts` suite passed (181 tests), including exact run-owned action cardinality, bounded dimensions, non-negative durations, and sensitive-value leakage checks. API lint, API type checks, formatting, full Knip, and pre-commit full Turbo type checks also passed. The all-workspace Vitest suite was not run because the change is isolated to the covered API route/service path.

### Verdict

LGTM

PR #28806, opened 2026-08-24
fix(rust): make payload-too-large diagnostics unit-accurate

### LINKED ISSUE #28788: docs(rust): make PayloadTooLarge diagnostics unit-accurate
## Problem

`crates/vsock-proto/src/error.rs:10-11` documents `ProtocolError::PayloadTooLarge` as carrying either an encoded byte size or an element count. However, its `Display` implementation at `crates/vsock-proto/src/error.rs:20-21` always renders the value as `bytes`:

```text
payload field too large: <field> (<size> bytes)
```

The value is not always a byte length. For example, `crates/vsock-proto/src/payloads/exec_operation/start.rs:338-352` constructs the variant with `request.env.len()` for `env_count` and `request.expected_exit_codes.len()` for `expected_exit_count`. Those values are numbers of entries, not bytes. The `write_files` encoder likewise uses the shared length helper for its `files.len()` value, which is encoded as the `file_count` field.

As a result, an oversized environment list is currently described as something like `payload field too large: env_count (1025 bytes)`, contradicting both the public Rust documentation and the wire schema.

## Impact

The public error diagnostic gives callers and maintainers the wrong unit when debugging count limits. It can send investigations toward encoded payload size instead of the actual number-of-elements constraint, and future uses of the shared variant can repeat the ambiguity.

## Decision

Use a dedicated public `ProtocolError::PayloadCountTooLarge(&'static str, usize)` variant for element-count limits. Keep `ProtocolError::PayloadTooLarge(&'static str, usize)` for byte lengths and preserve its `bytes` rendering. Count-aware encoded-field helpers will ensure that representational overflow for `env_count`, `expected_exit_count`, and `file_count` also uses the count-specific variant.

## Chosen Direction And Rationale

The dedicated count variant makes the unit explicit at every construction site, preserves the existing byte-oriented variant shape and diagnostics, and supports the three count paths found in the crate without field-name inference. The change is limited to error classification and human-readable output; protocol wire bytes and field widths remain unchanged.

## Alternatives Considered

- Adding unit metadata to the existing variant would model the distinction in one abstraction but would change every existing constructor and public pattern shape.
- Inferring units from field names would preserve the API shape but make future diagnostics depend on a naming convention.

## Risks And Mitigations

- Adding a public enum variant can require downstream exhaustive matches to handle it. The workspace has no other such matches; focused compilation and tests will verify the repository, and the public change will be called out in the PR summary.
- A count value can exceed its wire integer width before reaching the policy limit. Count-specific `u16`/`u32` helpers will cover both representational and policy-limit failures.

## Acceptance Criteria

- Make the public `PayloadTooLarge` contract and its rendered diagnostic accurately distinguish byte lengths from element counts, or use wording that is correct for both.
- Keep byte-oriented fields such as `payload`, `content`, and `diagnostic` unambiguous.
- Ensure count-oriented fields such as `env_count`, `expected_exit_count`, and `file_count` are never labeled as bytes.
- Add focused coverage for both a byte-length case and an element-count case.

## Delivery Notes

- The independently reviewable implementation is confined to `crates/vsock-proto` and its tests.
- No parent issue, dependency, migration, rollout, or wire-format update is required.

seven332:
## Summary

- Add `ProtocolError::PayloadCountTooLarge` so element-count limits are distinct from byte-length limits.
- Route `env_count`, `expected_exit_count`, and `file_count` through count-aware validation and diagnostics.
- Add public display coverage for byte and element units plus regression coverage for oversized file batches and count-overflow validation.

## Compatibility And Scope

- `ProtocolError::PayloadTooLarge` keeps its existing two-field shape and `bytes` rendering for byte-oriented values.
- Adding the public count-specific enum variant may require exhaustive downstream matches to handle it.
- This change does not alter protocol frames, wire field widths, message assignments, or persisted data.

Closes #28788

## Validation

- `cargo test --manifest-path crates/Cargo.toml -p vsock-proto` (151 unit tests, 2 new display tests, and all crate integration/property tests passed)
- `cargo clippy --manifest-path crates/Cargo.toml -p vsock-proto --all-targets -- -D warnings -A clippy::chunks_exact_to_as_chunks`
- The same Clippy command without the lint allowance remains blocked by two pre-existing warnings in `vsock-proto/src/payloads/memory_snapshot.rs`.

### DIFF
--- a/crates/vsock-proto/src/error.rs
+++ b/crates/vsock-proto/src/error.rs
@@ -7,8 +7,10 @@ pub enum ProtocolError {
     MessageTooSmall(usize),
     /// Payload contents were structurally invalid.
     InvalidPayload(&'static str),
-    /// Named payload field exceeded its encoded size or count limit.
+    /// Named payload field exceeded its encoded byte-length limit.
     PayloadTooLarge(&'static str, usize),
+    /// Named payload field exceeded its element-count limit.
+    PayloadCountTooLarge(&'static str, usize),
 }
 
 impl std::fmt::Display for ProtocolError {
@@ -19,6 +21,9 @@ impl std::fmt::Display for ProtocolError {
             Self::InvalidPayload(msg) => write!(f, "invalid payload: {msg}"),
             Self::PayloadTooLarge(field, size) => {
                 write!(f, "payload field too large: {field} ({size} bytes)")
+            }
+            Self::PayloadCountTooLarge(field, count) => {
+                write!(f, "payload field too large: {field} ({count} elements)")
             }
         }
     }
--- a/crates/vsock-proto/src/payloads/exec_operation/start.rs
+++ b/crates/vsock-proto/src/payloads/exec_operation/start.rs
@@ -1,7 +1,8 @@
 use crate::error::ProtocolError;
 use crate::read::{
-    checked_payload_len_add, ensure_payload_fits_message, ensure_u16_len, ensure_u32_len,
-    expect_consumed, read_i32, read_slice, read_str, read_u8, read_u16, read_u32,
+    checked_payload_len_add, ensure_payload_fits_message, ensure_u16_count, ensure_u16_len,
+    ensure_u32_count, ensure_u32_len, expect_consumed, read_i32, read_slice, read_str, read_u8,
+    read_u16, read_u32,
 };
 use crate::wire::EXEC_FLAG_SUDO;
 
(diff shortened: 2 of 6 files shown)


seven332:
## Code Review: PR #28806

**Reviewed head:** `e99d0f75fc0c073345193eb4e3cd821b2a416f02`

### Summary

The PR explicitly separates byte-length and element-count payload-limit errors, covers all identified production count paths, and preserves the wire format.

### Findings

No findings.

### Testing

The focused `vsock-proto` tests passed, including 150 unit tests, 2 new public display tests, and all integration/property tests. Formatting and focused Clippy passed with the known unrelated `chunks_exact_to_as_chunks` lint allowed. Strict Clippy remains blocked by two pre-existing warnings in `src/payloads/memory_snapshot.rs`.

### Verdict

LGTM

seven332:
## Code Review: PR #28806

**Reviewed head:** `39fad1fc39d6baa21b52b079ab75edd5b11639f9`

### Summary

The PR cleanly separates byte-length and element-count payload diagnostics, covers the `env_count`, `expected_exit_count`, and `file_count` production paths, preserves wire compatibility, and now covers the `u32` count-overflow branch identified by Codecov.

### Findings

No findings.

### Testing

The focused `vsock-proto` suite passed with 151 unit tests, 2 public display tests, all integration/property tests, and doc tests. Formatting and focused Clippy passed with the known unrelated `chunks_exact_to_as_chunks` lint allowed; strict Clippy remains blocked only by two pre-existing warnings in `src/payloads/memory_snapshot.rs`.

### Verdict

LGTM

PR #28809, opened 2026-08-24
test(python): cover protocol failure code normalization

### LINKED ISSUE #28796: test(python): cover protocol error-code normalization for all failure kinds
## Context

The trusted model-provider failure reducer explicitly normalizes provider error codes from JSON and SSE payloads in `crates/runner/mitm-addon/src/model_provider_failure.py:76-91` and `crates/runner/mitm-addon/src/model_provider_failure.py:722-735`.

The integration coverage in `crates/runner/mitm-addon/tests/test_model_provider_failure_reporting.py` currently exercises payload-driven authentication, billing, and provider-unavailable failures, while rate-limit and timeout coverage comes from HTTP status codes and connection coverage comes from transport state. As a result, the payload-code branches for `rate_limit`, `timeout`, and `connection` are not executed.

At `9e1fe0f3eb53e30182ef8bf88b71ef0176dda822`, a runtime mutation that empties `_RATE_LIMIT_CODES`, `_TIMEOUT_CODES`, and `_CONNECTION_CODES` leaves all 74 tests in `test_model_provider_failure_reporting.py` green. The unmodified implementation correctly classifies `rate_limit_error`, `timeout_error`, and `connection_error` as their corresponding failure kinds.

## Impact

Providers may return HTTP 200 responses whose JSON or SSE body carries a terminal error code. A regression in these mappings would silently suppress or misclassify trusted model-provider failure reports even though the existing HTTP-status and transport-error tests remain green.

## Acceptance criteria

- Add response-hook integration cases for payload error codes that normalize to `rate_limit`, `timeout`, and `connection`.
- Exercise JSON and/or SSE envelopes through the real model-provider failure observer rather than calling the private mapping helper directly.
- Assert the exact failure webhook payload for each normalized kind.
- Ensure the tests fail if any corresponding code set or mapping branch is removed.

seven332:
## Summary

- cover payload-driven `rate_limit_error`, `timeout_error`, and `connection_error` normalization through the real HTTP response hooks
- assert the exact normalized failure webhook payload for every newly covered kind
- verify each case fails when its corresponding production mapping is removed

## Scope

This is a test-only change. It does not change model-provider failure classification, reporting behavior, APIs, dependencies, or deployment contracts.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (`6528 passed`)
- focused mutation checks for `_RATE_LIMIT_CODES`, `_TIMEOUT_CODES`, and `_CONNECTION_CODES` (each corresponding test failed as expected, then the source was restored)

Closes #28796

### DIFF
--- a/crates/runner/mitm-addon/tests/test_model_provider_failure_reporting.py
+++ b/crates/runner/mitm-addon/tests/test_model_provider_failure_reporting.py
@@ -757,6 +757,24 @@ def test_media_type_classification_is_shared_by_usage_and_failure_observers(
         ),
         (
             "model-provider:openai-api-key",
+            "/v1/chat/completions",
+            b'{"error":{"code":"rate_limit_error"}}',
+            "rate_limit",
+        ),
+        (
+            "model-provider:openai-api-key",
+            "/v1/chat/completions",
+            b'{"error":{"code":"timeout_error"}}',
+            "timeout",
+        ),
+        (
+            "model-provider:openai-api-key",
+            "/v1/chat/completions",
+            b'{"error":{"code":"connection_error"}}',
+            "connection",
+        ),
+        (
+            "model-provider:openai-api-key",
             "/v1/responses",
             b'{"status":"failed","error":{"code":"server_error"}}',
             "provider_unavailable",


seven332:
## Code Review: PR #28809

**Reviewed head:** `fbd7b59581abbe4951219cab1691d52fb5e931a5`

### Summary

The test-only change adds three response-hook integration rows for the previously uncovered payload-code mappings. Each case uses an HTTP 200 response and asserts the exact normalized webhook body, so status and transport classifiers cannot satisfy it accidentally.

### Findings

No findings.

### Testing

The full addon suite passed (`6528 passed`) together with uv lock/sync, Ruff format/lint, and Basedpyright. Each new row was also verified to fail when its corresponding production mapping was temporarily removed. CI was still pending during this review.

### Verdict

LGTM

PR #22526, opened 2026-07-22
fix(platform): hide debug settings unless zero-debug is enabled

e7h4n:
## Summary

- Hide the Debug section in unified Settings unless `FeatureSwitchKey.ZeroDebug` is enabled.
- Fall back to Preference when a hidden Debug section is requested through a deep link.
- Cover both disabled and enabled behavior in the existing page-level integration test.

## User-visible impact

Users without the `zeroDebug` feature switch no longer see or open the Debug settings tab. Enrolled users keep the existing build information and network capture controls.

## Verification

- Prettier 3.8.1 on changed files
- `pnpm -F @vm0/app lint`
- `pnpm -F @vm0/app check-types`
- `pnpm -F @vm0/app exec vitest run src/views/zero-page/__tests__/zero-sidebar-account-menu.test.tsx --silent=passed-only` (13 tests passed)
- `pnpm exec knip --workspace @vm0/app`

Local dev server and browser verification were not run, per the vm0 PR verification preference.

PR #28816, opened 2026-08-24
refactor(core): rename the zero debug feature switch and migrate its stored key

### LINKED ISSUE #28787: refactor(feature-switch): rename ZeroDebug and migrate its stored key
Part of #26873 and #28368. Rename the `ZeroDebug` feature switch, including the persisted key, in one change.

## The key is stored in the database

```ts
// packages/core/src/feature-switch-key.ts:38
ZeroDebug = "zeroDebug",
```

`user_feature_switches.switches` is a `jsonb` column whose object keys are these literal values. Production currently holds 26 rows, 8 of which carry a `zeroDebug` key. Renaming the enum without migrating those rows would silently turn off the switch for those users — no error, just a setting that reads as unset.

The owner has accepted a single-PR cutover with no dual-read window: this switch gates debug affordances, so a brief inconsistency is not user-affecting.

## Change

**Enum** — `packages/core/src/feature-switch-key.ts:38`:

```ts
OkouDebug = "okouDebug",
```

Rename both the member and its value, and update `feature-switch.ts:197` where the descriptor is keyed.

**Migration** — add a numbered SQL migration following the existing sequence (`0975_…` is current) that rewrites the jsonb key in place:

```sql
UPDATE user_feature_switches
SET switches = (switches - 'zeroDebug')
             || jsonb_build_object('okouDebug', switches -> 'zeroDebug')
WHERE switches ? 'zeroDebug';
```

Check whether an org-level equivalent table exists and needs the same treatment before assuming `user_feature_switches` is the only store.

**Call sites** — roughly 80 references across four packages. Most read `FeatureSwitchKey.ZeroDebug` and follow the enum automatically once it is renamed. Three groups do not:

- `apps/desktop/src/desktop-developer-tools-controller.ts:4` hardcodes `const ZERO_DEBUG_FEATURE_SWITCH_KEY = "zeroDebug";` independently of the enum. It must change with the migration or the desktop developer tools stop unlocking.
- `apps/platform/src/signals/zero-page/realtime-status.ts` exports `ZeroDebugRealtimeIndicator` and `zeroDebugRealtimeIndicator$`, consumed by `zero-sidebar-account.tsx` and `platform-direct-bridge.test.ts`. These are named after the switch but are separate symbols.
- Test fixtures that write the raw string rather than the enum — `activity-inspect-page.test.tsx:114,305` uses `featureFlags: { zeroDebug: true }`, and the desktop controller tests build JSON bodies with the literal.

## Platform is owned by another engineer

More than half the call sites are under `apps/platform`. They are mechanical follow-ons of the enum rename rather than product changes, so they belong in this PR — a half-renamed enum does not compile. Keep the platform diff to exactly that: renamed references, no restructuring, no unrelated cleanup in files you touch.

## Do not

- Do not add a dual-read or dual-write path. The owner has explicitly declined the transition window.
- Do not change what the switch gates. Its descriptor text, default state and conditional behaviour stay as they are.
- Do not rename other `FeatureSwitchKey` members.
- Do not touch `zero_agents` or anything owned by #26938.

## Tests

- The migration rewrites a row holding `zeroDebug` and leaves a row without it untouched. Assert both.
- The migration is idempotent, or at least safe to re-run — `WHERE switches ? 'zeroDebug'` should make it a no-op the second time. Confirm rather than assume.
- Desktop developer tools still unlock when the switch is on, exercised through the hardcoded constant path.
- No `zeroDebug` or `ZeroDebug` identifier remains in `turbo/` outside the migration SQL itself.

lancy:
Closes #28787.

Renames the `ZeroDebug` feature switch to `OkouDebug` / `okouDebug`, including the key already persisted in the database, in a single cutover with no dual-read or dual-write window (per the owner's decision on the issue).

## Enum

`packages/core/src/feature-switch-key.ts` — both the member name and its value change; `feature-switch.ts` descriptor key follows. The switch's maintainer, description, default state and gated behaviour are unchanged. No other `FeatureSwitchKey` member is touched.

## Migration

`0977_rename_zero_debug_feature_switch_key.sql` rewrites the jsonb key in place and preserves its boolean value:

```sql
UPDATE "user_feature_switches"
SET
  "switches" = ("switches" - 'zeroDebug')
    || jsonb_build_object('okouDebug', "switches" -> 'zeroDebug'),
  "updated_at" = CURRENT_TIMESTAMP
WHERE "switches" ? 'zeroDebug';
```

`user_feature_switches` is the only override store. Org-scoped switches are not a separate table — they are rows in the same table under the `__org__` sentinel user id, so the unqualified `UPDATE` covers them. Verified by grepping `packages/db/src/schema` and by `feature-switches.service.ts`, which reads and writes both scopes through this one table.

## Call sites

Most of the ~78 references read `FeatureSwitchKey.ZeroDebug` and follow the enum. The three groups that do not were changed explicitly:

- `apps/desktop/src/desktop-developer-tools-controller.ts` — the hardcoded `ZERO_DEBUG_FEATURE_SWITCH_KEY = \"zeroDebug\"` constant is now `OKOU_DEBUG_FEATURE_SWITCH_KEY = \"okouDebug\"`. Without this the desktop developer tools would stop unlocking after the migration.
- `apps/platform/src/signals/zero-page/realtime-status.ts` — `ZeroDebugRealtimeIndicator` / `zeroDebugRealtimeIndicator$` renamed, with their consumers in `zero-sidebar-account.tsx` and `platform-direct-bridge.test.ts`.
- Test fixtures that write the raw string — `activity-inspect-page.test.tsx` (`featureFlags`) and the desktop controller tests' JSON bodies.

`apps/platform` carries more than half the diff. It is limited to renamed references: no restructuring and no unrelated cleanup in the files touched.

## Tests

New migration validator `packages/db/scripts/test-okou-debug-feature-switch-key-rename.ts`, registered in the migration consistency suite. It asserts:

- a row holding `zeroDebug: true` becomes `okouDebug: true` with its sibling keys intact;
- a row holding `zeroDebug: false` keeps the `false` value rather than being dropped or coerced;
- an org-scoped `__org__` row is migrated too;
- rows without the key are untouched, including their `updated_at`;
- re-running the migration SQL is a no-op — confirmed by re-executing the file and comparing every row's contents and `updated_at`.

The existing desktop controller tests exercise developer-tools unlocking through the hardcoded-constant path and pass.

## Verification

- `pnpm format`, `pnpm turbo run lint`, `pnpm check-types`, `pnpm knip` — all clean.
- `tsx scripts/test-okou-debug-feature-switch-key-rename.ts` — passes.
- Full `test-migration-consistency-schema.ts` — passes, including snapshot chain, journal timestamp ordering, latest-snapshot accuracy and the regenerated-migration comparison.
- `pnpm -F desktop exec vitest run src/desktop-developer-tools-controller.test.ts` (6 passed), `pnpm -F @okouai/core exec vitest run` (205 passed).
- `grep -riE 'zero.?debug'` over the repo returns nothing outside `apps/platform/CHANGELOG.md`, which is generated release history for a shipped PR and is deliberately left as-is.

Note: `validateConnectorAccountExpansion` fails in this sandbox on PostgreSQL 18 with an FK-RESTRICT constraint mismatch. It fails identically on a clean `main` checkout, so it is a local environment difference from the pinned CI image, not a regression from this change.


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: 6f328ad1695121742294c704d147ae5467d55fb5
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Renames the `ZeroDebug` feature switch to `OkouDebug` / `okouDebug` across `packages/core`, `apps/api`, `apps/platform` and `apps/desktop`, and migrates the already-persisted jsonb key in `user_feature_switches` in the same PR. Migration `0977` rewrites `switches - 'zeroDebug' || jsonb_build_object('okouDebug', switches -> 'zeroDebug')` guarded by `WHERE switches ? 'zeroDebug'`, and a new validator in the migration-consistency suite proves the rewrite, the untouched rows and the re-run no-op. Everything outside `packages/core` is a mechanical follow-on of the enum rename plus the three groups that do not follow it automatically (the hardcoded desktop constant, the platform realtime-indicator symbols, and raw-string test fixtures).

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic — no race conditions, unhandled edge cases or API misuse | Pass | The migration is a single guarded `UPDATE` on a 26-row table inside the default transactional migration wrapper (`1s` lock, `10s` statement timeout). `switches` is `NOT NULL DEFAULT '{}'`, so `switches ? 'zeroDebug'` and `switches -> 'zeroDebug'` cannot see a null column. `?` is not a placeholder for `postgres.js` `unsafe()`, so the predicate survives the runner unchanged. |
| `REVIEW.md` | Security — no injection, hardcoded secrets or missing auth on new surfaces | Pass | No new endpoint, no credential and no user-controlled SQL. `connector-catalog.ts:195` and `model-providers.ts:72` keep their existing gate; only the enum member name changes. |
| `REVIEW.md` | Type safety — no `any`, no overly broad types | Pass | `test-okou-debug-feature-switch-key-rename.ts` types the query result through a declared `SwitchesRow` interface rather than an inline assertion; the rest of the diff is identifier substitution under existing types. `pnpm check-types` is green across all 14 workspaces. |
| `REVIEW.md` | Style and maintainability — no oversized functions, no duplicated logic, comments explain WHY | Pass | The migration comment states why the key must move (stored jsonb key, org sentinel rows) rather than restating the SQL. The validator mirrors the existing `test-feishu-member-connector-reconciliation.ts` shape. |
| `REVIEW.md` | Interface changes — public interface changes documented | Pass | `FeatureSwitchKey.OkouDebug` and the exported `OkouDebugRealtimeIndicator` / `okouDebugRealtimeIndicator$` renames are described in the PR summary; every consumer is updated in the same diff. |
| `REVIEW.md` | DB/JSONB review gate — jsonb shape changes old rows could violate need a migration, a normalizer, or a rationale | Pass | This is exactly the case the gate targets, and the PR chooses the migration: `0977` plus a validator, with `user_feature_switches` confirmed as the only override store (`packages/db/src/schema` has no other switch table; `feature-switches.service.ts` reads and writes org scope through the same table under `__org__`). |
| `REVIEW.md` | Feature regression guard — changed existing behavior needs explicit regression coverage | Pass | The gated behavior is unchanged; the descriptor text, `enabled: false` default and every conditional stay as they were. The existing gate tests were carried over and the desktop unlock path keeps its own tests. |
| `REVIEW.md` | Testing coverage for the commit type (`refactor:` — existing tests must still cover the code) | Pass | All ~78 call sites keep their existing coverage, and the persisted-state change adds coverage it did not previously have. |
| `docs/bad-smell.md` | Zero tolerance for `any` and for lint/type suppressions | Pass | No `any`, no `eslint-disable`, `@ts-ignore`, `@ts-expect-error` or `prettier-ignore` anywhere in the diff. `pnpm turbo run lint` passes with zero errors. |
| `docs/bad-smell.md` | No dynamic `import()` in production code; no hardcoded environment values | Pass | No dynamic imports and no URLs or environment values are introduced. |
| `docs/bad-smell.md` | Fail fast — no recovery logic that hides a configuration or data problem | Pass | The migration does not tolerate an unexpected shape; a corrupted `switches` value would fail the migration loudly rather than being silently skipped. |
| `docs/fallback.md` | The default is no fallback — do not add dual-read/dual-write for a state the cutover resolves | Pass | The diff adds no compatibility branch, no dual-read and no dual-write. The single-PR cutover is the owner's recorded decision on #28787. |
| `docs/fallback.md` | Every fallback introduced by the diff must be declared in the PR summary | N/A | The diff introduces no fallback, so no `Fallbacks` section is required and none is present. |
| `docs/fallback.md` | No tombstone tests asserting removed behavior stays removed | Pass | The validator asserts the new key's presence and value, not the legacy key's absence as a standalone claim; the `zeroDebug` literals appear only as the pre-migration fixture the rewrite consumes. No `it("still ignores zeroDebug")`-style test is added. |
| `docs/deployment-compatibility.md` | Old code after migration — every statement the draining API can issue stays legal | Pass | The change is a jsonb object key, not a schema object. The previous API's `SELECT switches FROM user_feature_switches` stays legal after `0977`; nothing it can emit produces `42703`/`42P01`. See the note below for the behavioral consequence. |
| `docs/deployment-compatibility.md` | New code before migration — new readers must not require the new shape | Pass | The new API reading an unmigrated row finds no `okouDebug` key and resolves the switch as unset. No error path, so the deploy-before-migrate direction is safe in the SQL sense. |
| `docs/deployment-compatibility.md` | Migration must be transactional-safe or idempotent under full retry | Pass | `0977` carries no `-- vm0:non-transactional` marker, so it runs in one transaction; it is also idempotent on its own, which the validator proves by executing the file a second time and comparing every row's contents and `updated_at`. |
| `docs/testing.md` | Integration tests over unit tests; mock only at the system boundary | Pass | The new validator runs the real migration files against a real PostgreSQL database it creates and drops; nothing is mocked. |
| `docs/testing.md` | Real database, real filesystem — no mocking of internal infrastructure | Pass | `applyMigrationsFromDirectoryUpToTag` replays actual migration SQL; the fixture is real rows. |
| `docs/testing.md` | Deployment-compatibility tests when a change touches a persisted boundary | Pass | The validator seeds the pre-migration shape, upgrades it, and asserts both the migrated and the untouched rows — the "populate the pre-migration schema, upgrade it, verify" pattern. |
| `docs/testing/anti-patterns.md` | AP-1 testing mocks, AP-2 direct `fetch` mocking, AP-4 mocking internal code, AP-5 fake timers | Pass | None appear. The desktop tests keep their injected `fetchFeatureSwitches` fake, which is the constructor-level dependency injection the controller was designed for, not a `vi.mock` of internal code. |
| `docs/testing/anti-patterns.md` | AP-7 testing implementation details, AP-8 over-testing | Pass | The validator asserts observable row state, not the SQL text. The five assertions map one-to-one to the issue's required cases with no redundancy. |
| `docs/testing/patterns.md` | Follow the established test-file structure and helper conventions for the surface | Pass | The validator matches the sibling migration validators' structure (seed → apply up to tag → assert → re-run → assert) and is registered in `main()` alongside them. |
| `docs/testing/testing-external-behavior.md` | Exercise external interfaces rather than internal implementation | Pass | Platform and API tests continue to drive pages and routes; only the switch key inside their fixtures changed. |
| `docs/testing/api-testing.md` | API tests go through the Hono app and the route contract, not internal services | Pass | No API test structure changed — `connector-catalog.test.ts`, `model-providers.test.ts` and the BDD helpers keep their route-level shape. |
| `docs/testing/app-testing.md` | Platform tests use page-level setup and MSW handlers, not direct component rendering | Pass | `settings-dialog.test.tsx`, `zero-sidebar-account-menu.test.tsx` and the activity-page tests keep their existing setup; only the `featureSwitches` / `featureFlags` key changed. |
| `docs/testing/desktop-testing.md` | Desktop tests exercise the real controller through its injected boundary | Pass | `desktop-developer-tools-controller.test.ts` still drives `requestRefresh()` against injected `Response` objects and asserts availability, so the hardcoded-constant path is genuinely covered by the rename. |
| `.claude/skills/ccstate/SKILL.md` | `computed` derivations, commands and signal ownership follow ccstate conventions | Pass | `okouDebugRealtimeIndicator$` remains the same `computed` with the same dependencies; no manual cache, no `detach`/`void` change, no new signal allocation. |
| `docs/effect.md` | Keep render pure; choose the right execution boundary for derived values | Pass | No effect, command or lifecycle boundary is added or moved; the platform diff is identifier substitution inside existing render-time reads. |
| `docs/cache.md` | No render-phase side effects, render-time signal allocation, or duplicate sources of truth | Pass | `realtime-status.ts` keeps one exported computed as the single source; consumers read it through `useGet`. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Notes (non-blocking)

- `turbo/packages/db/src/migrations/0977_rename_zero_debug_feature_switch_key.sql` — during the DB-vs-API window (observed maximum exposure ~102 minutes per `docs/fallback.md` §7), a draining previous API reads the migrated row, drops the unknown `okouDebug` key in `filterFeatureSwitchOverrides`, and resolves the switch as unset. That is the transient inconsistency the owner accepted. Worth stating explicitly: if one of those users writes any switch through the Lab page inside that window, `upsertFeatureSwitches` re-serializes the filtered object and the `okouDebug` key is dropped from the row permanently, not just for the read. The blast radius is 8 production rows, the affected capability is debug affordances only, and recovery is re-toggling the switch. Not raised to P1 because the only repairs are a dual-read/dual-write path, which `docs/fallback.md` §2-§5 rejects and the owner explicitly declined on #28787, or no change at all — there is no safe in-scope alternative to weigh against it.
- `turbo/apps/desktop/src/desktop-developer-tools-controller.ts` — the desktop app has its own release path, so an installed build that predates this change will look for `zeroDebug` in the feature-switches response and stop unlocking developer tools until the user updates. Same accepted-cutover reasoning; recording it so the behavior is not mistaken for a regression.
- `turbo/apps/platform/CHANGELOG.md:6027` retains the string `ZeroDebug` in generated release history for #22526. Leaving generated changelog entries untouched is correct; noting it so the "no identifier remains" sweep is not read as incomplete.

### Testing

- Coverage: Adequate. `refactor:` commits require that existing tests still cover the code, and every renamed call site keeps its test. Beyond that, the persisted-state change gains coverage it did not have: `packages/db/scripts/test-okou-debug-feature-switch-key-rename.ts` asserts the enabled rewrite with sibling keys preserved, the `false` value preserved rather than dropped or coerced, the org-scoped `__org__` row migrated, rows without the key untouched including their `updated_at`, and the re-run no-op verified by executing the file again and comparing every row's contents and `updated_at`. The desktop developer-tools unlock is exercised through the hardcoded-constant path by the existing controller tests.
- Conventions: Compliant. Integration-style, real database, no internal mocking, no fake timers, no tombstone assertions, and the validator is registered in `test-migration-consistency-schema.ts` so it runs in CI rather than only on demand.

PR #28827, opened 2026-08-24
test(python): cover missing upstream certificate evidence

### LINKED ISSUE #28797: test(python): isolate missing upstream certificate evidence in admission tests
## Context

`crates/runner/mitm-addon/src/upstream_admission.py:81-125` only accepts a connected TLS destination as binding evidence when TLS is established, no connection error exists, SNI matches, certificate evidence is present, and an authoritative connected endpoint is available. The empty-certificate guard is at lines 114-115.

The existing negative case `test_matching_sni_and_host_blocks_connected_firewall_auth_without_verified_tls` in `crates/runner/mitm-addon/tests/test_request_handler_connector_admission.py:183-208` omits TLS establishment entirely, so it returns at the earlier `tls_established` guard. Positive cases populate `certificate_list`, but no negative case keeps every other verification precondition valid while leaving only the certificate list empty.

At `9e1fe0f3eb53e30182ef8bf88b71ef0176dda822`, branch coverage confirms that the empty-certificate return is not executed. A runtime mutation that treats an empty certificate list as present leaves all 92 tests across the request/requestheaders API and connector admission suites green. A direct harness against the unmodified helper returns `None` with an empty certificate list and the connected endpoint after adding certificate evidence.

## Impact

This is a privileged upstream trust boundary for platform API allowance and connector credential injection. A regression that removes or bypasses the certificate-evidence check could admit a connected destination from incomplete TLS state without any admission test failing.

## Acceptance criteria

- Add an integration case with connected state, established TLS, matching normalized SNI, no connection error, and a valid connected endpoint, but an empty `certificate_list`.
- Assert that no direct upstream binding is created, credential resolution is not called, and the request fails closed.
- Include a positive control that differs only by supplying certificate evidence and is admitted.
- Ensure the negative case fails if the certificate-evidence guard is removed.

seven332:
## Summary

- add a parameterized request-hook integration control pair for connected upstream certificate evidence
- verify missing evidence fails closed before credential resolution and never records a direct binding
- verify the otherwise identical certificate-present flow is admitted, receives credentials, and records the connected endpoint binding

## Scope

- test-only change in the Python mitmproxy add-on
- no production behavior, dependency, protocol, persistence, or deployment changes

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (`6530 passed`)
- temporarily bypassed the empty-certificate guard and confirmed the `certificate-missing` case fails because credential resolution occurs, then restored the source and reran the focused pair successfully

Closes #28797

### DIFF
--- a/crates/runner/mitm-addon/tests/test_request_handler_connector_admission.py
+++ b/crates/runner/mitm-addon/tests/test_request_handler_connector_admission.py
@@ -208,6 +208,62 @@ async def test_matching_sni_and_host_blocks_connected_firewall_auth_without_veri
     assert "Authorization" not in flow.request.headers
 
 
+@pytest.mark.parametrize(
+    "has_certificate_evidence",
+    [
+        pytest.param(True, id="certificate-present"),
+        pytest.param(False, id="certificate-missing"),
+    ],
+)
+async def test_matching_sni_and_host_requires_upstream_certificate_evidence(
+    tmp_path,
+    real_flow,
+    mitm_ctx,
+    fake_firewall_headers,
+    headers,
+    has_certificate_evidence: bool,
+):
+    reg_path = _write_github_firewall_registry(tmp_path)
+    flow = real_flow(
+        with_response=False,
+        client_ip="10.200.0.5",
+        host="140.82.112.5",
+        sni="api.github.com",
+        path="/repos",
+        request_headers=headers(("Host", "api.github.com")),
+    )
+    mark_connected_tls_upstream(
+        flow,
+        sni="api.github.com",
+        server_address=("140.82.112.5", 443),
+        peername=("140.82.112.5", 443),
+    )
+    if not has_certificate_evidence:
+        flow.server_conn.certificate_list = ()
+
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #28827

**Reviewed head:** `a91baf98f0b105f23fc820cabc1494845492c069`

### Summary

The submitted test-only change follows the challenged plan: an exact request-hook integration control pair isolates upstream certificate evidence while exercising connector admission, credential resolution, and direct-binding behavior.

### Findings

No findings.

### Testing

Local validation passed for uv lock/sync, Ruff format and lint, BasedPyright, and the complete Python add-on suite (`6530 passed`). A temporary bypass of the empty-certificate guard made the missing-certificate case fail because credential resolution occurred, proving the regression test is mutation-sensitive; production source was restored and is absent from the PR diff. Turbo and Rust checks are outside this Python-test-only scope. Remote CI is not yet verified by this review stage.

### Verdict

LGTM

PR #28828, opened 2026-08-24
fix(api): refresh codex tokens before subscription usage

### LINKED ISSUE #28773: bug: stop repeated Codex usage 401s from stale account tokens
## Problem

`GET /api/me/model-providers` enriches every personal Codex account (active and inactive) with current subscription usage. The enrichment path reads the persisted `CHATGPT_ACCESS_TOKEN` and calls `GET https://chatgpt.com/backend-api/wham/usage` directly, without first applying the account's OAuth expiry and refresh state.

When that access token is no longer accepted, every provider-list request repeats the upstream call, receives 401, logs a warning, and returns the account with stale or absent usage. The account remains `needsReconnect=false` because the normal token-refresh state machine was never entered.

## Evidence

- For the Beijing day 2026-08-23, production emitted 63 `failed to refresh personal model provider subscription usage` warnings.
- Across 2026-08-17T00:00:00Z through 2026-08-24T04:00:00Z, all 201 matching warnings had the same nested error: `Codex usage request failed with status 401`. No network, 5xx, or response-schema failures appeared in that sample.
- Axiom stores the error as `fields.error.message`, `fields.error.name`, and `fields.error.stack`. The earlier conclusion that the logger dropped the error came from querying non-existent flat fields and is superseded.
- Several request timestamps contain warning pairs a few milliseconds apart, consistent with the route refreshing usage for multiple saved Codex accounts in parallel.
- The platform intentionally renders subscription usage for all saved accounts, not only the active account, so skipping inactive accounts would remove useful UI data rather than fix the credential lifecycle.
- The terminal refresh-loop fix in #27967 applies to the runner/firewall credential path. This provider-list usage path does not currently invoke that refresh path.

## Decision

This is a real data freshness and operability bug, but the original diagnostic framing was incorrect. Fix the credential lifecycle at the usage-read boundary instead of adding redundant logging or merely lowering the warning level.

## Chosen direction

- Before querying Codex subscription usage, resolve a current access token through the existing account-aware, database-locked model-provider OAuth refresh authority.
- Pin refresh and secret reads to the exact `modelProviderAccountId`; do not substitute the active account and do not stop enriching inactive accounts.
- Reuse the existing refresh result classification and persistence rules:
  - a current token is used directly;
  - an expired or near-expiry token is refreshed once under the existing lock, including refresh-token rotation and expiry persistence;
  - terminal refresh failures mark that exact account as requiring reconnect;
  - transient upstream refresh failures retain fallback behavior without falsely marking reconnect.
- Only call the usage endpoint after current-token resolution succeeds. Keep the provider-list endpoint available when usage metadata cannot be refreshed.
- Do not infer `needsReconnect` from a usage-endpoint 401 alone. Access-token expiry is recoverable when the refresh token is still valid; reconnect state must come from the authoritative refresh attempt.
- Keep diagnostics sanitized. Include the stable provider account identifier and existing structured error/status information, but never tokens or upstream response bodies.

## Alternatives considered

- **Log more fields only:** rejected because the nested error and 401 status are already present; it would improve neither usage freshness nor repeated requests.
- **Treat usage 401 as reconnect-required:** rejected because a normal access-token expiry would incorrectly force the user to reconnect even when the refresh token remains valid.
- **Retry the same usage request:** rejected because retrying an unchanged rejected token cannot recover.
- **Refresh only the active account:** rejected because the account picker displays usage for every saved account.
- **Silently skip expired accounts:** rejected because it avoids warnings by leaving the intended usage UI stale.

## Risks and mitigations

- Provider-list reads can now perform OAuth refresh writes. This is necessary for a current credential and already follows the same locked state transition used by runtime credential resolution.
- Multiple expired accounts can add upstream refresh latency. Refresh only accounts whose persisted expiry state requires it, retain parallel account enrichment, and rely on the existing lock to coalesce competing requests safely.
- Refresh-token rotation is concurrency-sensitive. Reuse the existing locked refresh and persistence implementation rather than adding a second refresh implementation.
- A refresh failure must not make the list endpoint fail. Preserve per-account fallback and verify that one unavailable account does not suppress the others.

## Acceptance criteria

- Listing personal model providers does not call Codex usage with an access token that the persisted state already considers expired or near expiry.
- Each Codex account resolves its own current access token before usage enrichment, including inactive accounts.
- A successful refresh persists rotated secrets and the new expiry, then the usage request uses the refreshed access token.
- Concurrent list requests for the same expired account do not independently rotate the same refresh token.
- A terminal refresh failure updates the exact account's reconnect state and subsequent list requests do not repeat the terminal refresh or usage request.
- A transient refresh failure and a usage-response failure both preserve a successful provider-list response with stale/absent usage for only the affected account.
- Logs identify the affected model-provider account and retain structured error/status data without secrets or response bodies.
- API integration coverage exercises current-token, expired-token refresh, terminal refresh failure, transient refresh failure, non-2xx usage, unrecognized usage response, multi-account isolation, and concurrent refresh behavior at the route entry point.

## Delivery note

The behavior and its integration coverage form one independently reviewable API PR. A broad extraction of the firewall authentication service is not required for this fix; the implementation should expose and reuse a narrow current-token capability from the existing refresh authority without duplicating OAuth state transitions.

seven332:
## Summary

- resolve each personal Codex account's current access token through the existing locked OAuth refresh authority before fetching subscription usage
- preserve exact-account isolation, refresh-token rotation, terminal reconnect state, and transient per-account fallback
- add sanitized account identifiers to refresh and usage diagnostics
- cover concurrent refresh coalescing, terminal and transient failures, malformed usage responses, and multi-account isolation at API route entry points

## Explicit exclusions

- no API contract, database schema, migration, frontend, or runner changes
- no forced refresh or reconnect transition based only on an unexpected usage-endpoint 401
- no broad extraction of the existing firewall credential lifecycle service

## Validation

- `pnpm format`
- `pnpm -F api lint`
- `pnpm -F api check-types`
- `pnpm -F api exec vitest run src/signals/routes/__tests__/me-model-providers-upsert.test.ts src/signals/routes/__tests__/auth-device.bdd.test.ts src/signals/routes/__tests__/webhooks-agent-firewall-auth.bdd.test.ts --maxWorkers=4 --silent=passed-only` (90 passed)
- `pnpm knip`

Closes #28773

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/auth-device.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/auth-device.bdd.test.ts
@@ -5,7 +5,11 @@ import { FeatureSwitchKey } from "@okouai/core/feature-switch-key";
 import { now } from "../../../lib/time";
 import { testContext } from "../../../__tests__/test-context";
 import { server } from "../../../mocks/server";
-import { createBddApi, expectApiError } from "./helpers/api-bdd";
+import {
+  createBddApi,
+  expectApiError,
+  type ApiTestUser,
+} from "./helpers/api-bdd";
 import {
   createAuthDeviceApiActions,
   mockClaudeCodeTokenEndpoint,
@@ -51,6 +55,36 @@ function expectCliApprovalError(
   ) {
     throw new Error("Expected CLI approval error response body");
   }
+}
+
+async function connectPersonalCodexTestAccount(
+  actor: ApiTestUser,
+  args: {
+    readonly accessTokenExpiresAt: number;
+    readonly accountId: string;
+    readonly refreshToken: string;
+    readonly workspaceName: string;
+  },
+): Promise<string> {
+  mockCodexDeviceAuthProvider({
+    tokenScope: "personal",
+    ...args,
+  });
+  const started = await authDevice.requestCodexStart(actor, "personal", [200], {
+    mode: "add",
+  });
+  if (started.status !== 200) {
+    throw new Error(`Expected ${args.accountId} device auth to start`);
+  }
+  const completed = await authDevice.requestCodexComplete(
+    actor,
+    started.body.sessionToken,
+    [200],
+  );
+  if (!("status" in completed.body) || completed.body.status !== "complete") {
+    throw new Error(`Expected ${args.accountId} device auth to complete`);
+  }
+  return completed.body.provider.id;
 }
 
 describe("AUTH-02: CLI device authorization", () => {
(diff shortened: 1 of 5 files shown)


seven332:
## Code Review: PR #28828

**Reviewed head:** `74baad99c40f7a64dcd4686c70e7893df8014708`

### Summary

The provider-list path now resolves each concrete Codex account's current access token through the existing locked refresh authority before usage enrichment. It preserves account isolation, refresh-token rotation and reconnect semantics, keeps optional usage failures local to one account, and adds sanitized account-level diagnostics without changing public contracts or schema.

### Findings

No findings.

### Testing

Reviewed route-level integration coverage for current, expired, terminal, transient, concurrent, malformed-response, and multi-account cases. Local API lint, type checking, Knip, diff checks, and the three focused route suites passed (`90 passed`). Unrelated Rust and Python suites were not run because this PR changes only the API TypeScript surface.

### Verdict

LGTM

PR #28838, opened 2026-08-24
fix(guest-agent): keep claude appended prompts out of argv

### LINKED ISSUE #28744: bug: keep Claude appended prompts out of process arguments
## Problem

Four consecutive Claude Code processes received `SIGKILL` while one retained Firecracker sandbox rendered videos through Hyperframes/Chrome. All four used Runner v0.171.0, the same Claude session/thread reuse key, sandbox `3019fc1b-ef36-4db1-9e66-f9ed44e4e770`, and a 4-GiB profile on `prod-4.aws.vm3.ai`.

The exact historical sender PID was not retained. The strongest supported mechanism is a same-UID workload cleanup command matching Claude's command line:

- the render workflow issued broad `pkill -9 -f` commands;
- procps `pkill -f` matches complete `/proc/<pid>/cmdline` values and does not ignore ancestors by default;
- Claude currently receives the complete appended system prompt through `--append-system-prompt <text>`;
- appended prompt/template/prior-context text can contain cleanup tokens such as `chrome`; and
- an isolated process experiment proves this class of broad match can kill an ancestor whose argv contains the token.

The actual incident argv is unavailable, so this issue does not claim that the historical sender has been identified conclusively. It addresses the concrete and reproducible cross-boundary collision surface.

## Affected runs

| Terminal time (UTC) | Run ID | Final unfinished activity |
| --- | --- | --- |
| 2026-08-23T08:29:02.606Z | `ad2697bf-7605-4ecd-afbc-b57815de018b` | explicit Node PID kills and broad Hyperframes/Chrome cleanup |
| 2026-08-23T08:37:26.499Z | `f5ed7b90-1fff-4374-899d-516c4e9ff747` | wait for detached `batch.mjs` render |
| 2026-08-23T08:38:41.999Z | `50b68403-2f43-4229-b512-c7ca2a02576f` | foreground `batch.mjs` render chunk |
| 2026-08-23T08:39:40.797Z | `22b5f51d-994e-4947-83c3-b0278c0f2dfa` | foreground `batch.mjs` render chunk |

The explicit PID operands in the first cleanup command were Node processes rather than Claude PID 273. After each Claude exit, Firecracker and the root guest supervisor stayed alive, operation cleanup completed, and the sandbox was reused.

## Evidence boundaries

The current evidence is inconsistent with:

- a known Guest Agent timeout, cancellation, reap, watchdog, heartbeat, stdin, stdout, or event-delivery termination path;
- operation-cgroup cleanup as the initiating event, because cleanup followed Guest Agent exit;
- the existing managed-tool OOM group-kill path, which preserves Claude and increments OOM counters;
- a host OOM selectively killing a guest PID while Firecracker survives; or
- guest VM shutdown.

Guest memory and balloon pressure were material, but hierarchical workload `memory.events` reported zero `oom`, `oom_kill`, and `oom_group_kill`, and the retained guest `dmesg` tail had no matching OOM record. Balloon tuning is not part of this correction.

## Decision

**Proceed with one PR that moves Claude's appended system prompt out of argv and into a private run-scoped file.**

Anthropic documents `--append-system-prompt-file <path>` as the file-backed equivalent of `--append-system-prompt <text>`. The Runner rootfs pins Claude Code 2.1.241, which supports this contract.

Guest Agent will write the exact appended prompt bytes into the existing private run directory and pass only the path to Claude. This removes prompt text from process-command matching while preserving the API payload and Claude prompt semantics.

## Why this direction

- It removes the concrete argv surface implicated by the observed broad cleanup commands.
- It is independently valid hardening even though the historical sender PID is unavailable.
- It follows the existing private Pi launch-payload pattern rather than introducing a new guest daemon or process-isolation subsystem.
- It changes no API, queue, database, or Runner-to-guest protocol.
- A process-entry regression can fail with inline transport and pass with file-backed transport.

### Linux Audit or seccomp signal provenance

This can identify a future userspace sender, but it adds guest-global rule/consumer lifecycle, kernel-log attribution, retention, and production-kernel testing. Unprivileged seccomp installation also conflicts with supported `sudo` behavior through `no_new_privs`. It does not prevent the known argv collision.

### Shared tool PID namespace

This provides a stronger general signal boundary, but changes PID visibility, `/proc`, orphan handling, detached tools, Chrome sandboxing, cleanup, and reuse for every tool. The current evidence does not justify that runtime-wide compatibility cost.

### Rewrite or discourage `pkill`

Prompt advice is not enforcement, and safely rewriting arbitrary shell programs at PreToolUse would require a new shell-policy layer. Absolute paths and direct syscalls bypass wrappers.

### Balloon tuning

This reduces host overcommit capacity and does not address userspace signal calls. The incident has no retained kernel/cgroup OOM evidence.

## Implementation scope

- Add a canonical private run-scoped Claude appended-prompt path.
- Write non-empty prompt bytes with the existing private-file helper before Claude launch.
- Replace inline `--append-system-prompt` with `--append-system-prompt-file`.
- Preserve empty-prompt, stdin stream-json, resume, Pi, and Codex behavior.
- Update `guest-mock-claude` argument compatibility.
- Add a process-entry integration regression using a synthetic prompt marker and broad `pkill -f` behavior.
- Verify the pinned rootfs Claude CLI contract without upgrading Claude in the same PR.

## Compatibility and rollout

- Old runners continue inline prompt transport while draining; new rootfs instances use file transport.
- Guest Agent and Claude ship in the same rootfs, so no mixed-version wire handshake is introduced.
- The prompt file lives in the existing `0700` run directory with mode `0600`.
- The file is not treated as a separate secret-isolation boundary because managed tools share the sandbox UID and the prompt already exists in run payload/runtime memory.
- Rollback requires no data migration.

## Acceptance criteria

- [ ] Non-empty Claude appended prompts use only `--append-system-prompt-file`.
- [ ] Raw appended-prompt text is absent from Claude argv.
- [ ] Claude receives byte-equivalent prompt content from a private `0600` run-scoped file.
- [ ] Empty prompt, stream-json stdin, resume, Pi, and Codex paths remain unchanged.
- [ ] An entry-point regression proves broad `pkill -f` can terminate its matching tool shell without terminating the mock Claude parent.
- [ ] Rootfs validation covers the pinned Claude file-flag contract.
- [ ] Formatting, Clippy, documentation, and affected crate integration tests pass.

## Out of scope

- Permanent signal syscall capture
- PID namespace or UID isolation
- Rewriting user-authored signal commands
- Balloon or workload-memory policy
- Reclassifying exit 137 as OOM
- Moving unrelated inline Claude settings to files

## Follow-up gate

If a structured Claude SIGKILL recurs after all active runners use file-backed prompt transport, prompt-text argv collision no longer explains that new event. Bounded userspace signal provenance or a stronger runtime/tool isolation boundary then becomes separate justified work.

## References

- Claude Code CLI reference: https://code.claude.com/docs/en/cli-usage
- procps `pkill`: https://man7.org/linux/man-pages/man1/pgrep.1.html
- Linux `kill(2)`: https://man7.org/linux/man-pages/man2/kill.2.html
- Linux cgroup v2: https://docs.kernel.org/admin-guide/cgroup-v2.html
- #28366 / #28391: isolate managed shell-tool OOM cleanup

## Evidence locations

- Axiom: `vm0-agent-run-events-prod`, `vm0-sandbox-telemetry-system-prod`, `vm0-sandbox-telemetry-metrics-prod`
- Runner logs under `/var/lib/vm0-runner/logs` on `prod-3`, `prod-4`, and `prod-11`

---
> Plan selected on 2026-08-24 after Research, Options, and Challenge review.

seven332:
## Summary

- write non-empty Claude appended system prompts to a private run-scoped `0600` file
- launch Claude with `--append-system-prompt-file` so prompt text is absent from process arguments
- add a process-entry regression that proves a broad same-UID `pkill -f` kills its matching shell without killing the mock Claude parent
- require the baked Claude CLI to contain the file-backed prompt flag during rootfs verification

## Scope

This removes the reproducible prompt/argv collision surface behind the strongest supported incident mechanism. It does not claim that the sender of the historical SIGKILLs has been conclusively identified, and it does not add general signal provenance, PID namespaces, command rewriting, or memory-policy changes.

## Validation

- `cd crates && cargo clippy -p guest-contracts -p guest-agent -p guest-mock-claude --all-targets -- -D warnings`
- `cd crates && cargo doc -p guest-contracts -p guest-agent -p guest-mock-claude --no-deps`
- `cd crates && cargo test -p guest-mock-claude`
- `cd crates && cargo test -p runner cmd::build::scripts::tests`
- `cd crates && bash -n runner/scripts/verify-rootfs.sh`

Closes #28744

### DIFF
--- a/crates/guest-agent/src/cli/command.rs
+++ b/crates/guest-agent/src/cli/command.rs
@@ -20,7 +20,8 @@ pub(super) fn build_claude_command_for_runtime(
         ClaudeArgsConfig {
             model: runtime.anthropic_model.as_ref(),
             resume_id: runtime.resume_session_id.as_ref(),
-            append_system_prompt: runtime.append_system_prompt.as_ref(),
+            append_system_prompt_file: (!runtime.append_system_prompt.is_empty())
+                .then_some(runtime.claude_append_system_prompt_file.as_ref()),
             disallowed_tools: runtime.disallowed_tools.as_ref(),
             tools: runtime.tools.as_ref(),
             settings: runtime.settings.as_ref(),
@@ -49,7 +50,7 @@ fn push_comma_separated_flag_values(args: &mut Vec<String>, flag: &str, values: 
 struct ClaudeArgsConfig<'a> {
     model: &'a str,
     resume_id: &'a str,
-    append_system_prompt: &'a str,
+    append_system_prompt_file: Option<&'a str>,
     disallowed_tools: &'a str,
     tools: &'a str,
     settings: &'a str,
@@ -78,9 +79,9 @@ fn build_claude_args(config: ClaudeArgsConfig<'_>) -> Vec<String> {
         log_info!(LOG_TAG, "Starting new session");
     }
 
-    if !config.append_system_prompt.is_empty() {
-        args.push("--append-system-prompt".to_string());
-        args.push(config.append_system_prompt.to_string());
+    if let Some(path) = config.append_system_prompt_file {
+        args.push("--append-system-prompt-file".to_string());
+        args.push(path.to_string());
     }
 
     push_comma_separated_flag_values(&mut args, "--disallowed-tools", config.disallowed_tools);
@@ -132,6 +133,8 @@ fn build_claude_command_with_config(
 mod tests {
     use super::*;
 
+    const TEST_APPEND_SYSTEM_PROMPT_FILE: &str = "/tmp/claude-append-system-prompt";
+
     fn disable_system_log() {
         guest_common::log::clear_system_log_file();
     }
@@ -166,7 +169,8 @@ mod tests {
         build_claude_args(ClaudeArgsConfig {
             model: "",
             resume_id,
-            append_system_prompt,
+            append_system_prompt_file: (!append_system_prompt.is_empty())
+                .then_some(TEST_APPEND_SYSTEM_PROMPT_FILE),
             disallowed_tools,
             tools,
             settings,
@@ -187,7 +191,7 @@ mod tests {
             ClaudeArgsConfig {
                 model: "",
                 resume_id: "",
-                append_system_prompt: "",
+                append_system_prompt_file: None,
                 disallowed_tools: "",
                 tools: "",
                 settings: "",
@@ -202,7 +206,7 @@ mod tests {
         build_claude_args(ClaudeArgsConfig {
             model,
             resume_id: "",
-            append_system_prompt: "",
+            append_system_prompt_file: None,
             disallowed_tools: "",
             tools: "",
             settings: "",
@@ -231,6 +235,7 @@ mod tests {
         assert_claude_prompt_is_not_positional(&args, "hello world");
         assert!(args.contains(&"--replay-user-messages".to_string()));
         assert!(!args.contains(&"--append-system-prompt".to_string()));
+        assert!(!args.contains(&"--append-system-prompt-file".to_string()));
         assert!(!args.contains(&"--resume".to_string()));
     }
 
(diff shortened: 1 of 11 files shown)


seven332:
## Code Review: PR #28838

**Reviewed head:** `f6514d7e21f0fdd584a589de6ee8297834979188`

### Summary

The change cleanly replaces Claude's inline appended-prompt argument with a private run-scoped file, preserves the other launch paths, and adds a production-shaped process-entry regression for the broad `pkill -f` collision class.

### Findings

No findings.

### Testing

Affected Rust formatting, Clippy, rustdoc, Guest Contracts, Guest Agent, mock Claude, Runner script-contract tests, and shell syntax validation passed locally. The regression validates real argv, exact file bytes, `0700/0600` permissions, empty-prompt behavior, and same-UID signal behavior. The pinned rootfs Claude executable was not executed locally; Runner Image CI remains the deployment-artifact gate.

### Verdict

LGTM

PR #28846, opened 2026-08-24
fix(runner): recover mitmdump startup port collisions

### LINKED ISSUE #28835: bug(runner): recover mitmdump startup port collisions
## Background

The `runner-behavior-lane-d` job in [Actions run 32694941975](https://github.com/vm0-ai/vm0/actions/runs/32694941975/job/97335959980?pr=28830) intermittently failed during the browser benchmark. The default curl benchmark completed successfully, while the next benchmark selected proxy port `42091` and mitmdump reported:

```text
Transparent Proxy failed to listen on *:42091 ... address already in use
```

About 54 ms later, the runner logged `mitmdump started`. Both browser navigation attempts then failed with `net::ERR_CONNECTION_REFUSED`. Rerunning the same head selected different ports and passed the browser benchmark and `ci-gate-crates`.

PR #28830 changed only memory-snapshot protocol tests and is unrelated to proxy startup behavior.

## Problem

`MitmProxy::new` calls `find_available_port`, which binds `0.0.0.0:0`, records the assigned port, and releases the listener before CA preparation, addon setup, PyInstaller startup, and mitmdump's eventual bind. Another process on the shared metal host can claim that port during this gap.

The readiness contract can then accept the wrong listener:

- the addon writes its launch-specific ready marker from `configure`, before mitmproxy has successfully bound its listening socket;
- `wait_for_ready` separately checks the marker, the child status, and whether the selected TCP port accepts a connection;
- when the new child has written the marker but is still exiting after `EADDRINUSE`, a different process listening on that port satisfies the TCP probe.

This reports a failed mitmdump launch as ready and sends sandbox traffic to an unavailable or unrelated proxy endpoint. A browser-level retry cannot recover because both attempts use the same failed proxy instance.

## Acceptance criteria

- A mitmdump launch that cannot bind its selected proxy port must never be reported as ready.
- Readiness evidence must establish that the current launch reached mitmproxy's post-bind running state; an addon configuration marker plus an unrelated listener is insufficient.
- Initial proxy startup must recover from a confirmed transient port collision by selecting a new port within a small bounded retry budget.
- Non-port startup failures must still fail immediately and preserve their original diagnostics rather than being treated as collisions.
- Add deterministic integration coverage for an occupied selected port that proves the colliding launch is rejected or retried without connecting to the unrelated listener.
- Preserve restart behavior that must continue using the already-published proxy port, along with existing addon initialization, process supervision, and cleanup guarantees.

## Relevant code

- `crates/runner/src/proxy/process.rs`: port allocation, process spawn, and readiness polling
- `crates/runner/mitm-addon/src/mitm_addon.py`: addon ready marker publication
- `.github/scripts/runner-behavior-benchmark.sh`: browser benchmark and its navigation retry

seven332:
## Summary

- publish the addon ready marker from mitmproxy's post-bind `running()` lifecycle hook
- classify selected-port bind failures from startup stderr and retry initial startup on a fresh port, with a three-attempt budget
- attach crash forwarding only after readiness so a failed attempt cannot leave a crash notification for the successful replacement
- cover occupied-port recovery and fail-fast non-port startup behavior with real child processes and TCP listeners

## Scope

Initial startup may replace its port before the sandbox runtime consumes it. Runtime restarts continue to use the already-published port and keep their existing backoff behavior. This does not change APIs, persistence, queues, frontend behavior, or sandbox protocols.

## Validation

- `cargo test -p runner` (3236 passed, 20 ignored; integration test passed)
- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (6541 passed)

Closes #28835

### DIFF
--- a/crates/runner/mitm-addon/src/mitm_addon.py
+++ b/crates/runner/mitm-addon/src/mitm_addon.py
@@ -254,12 +254,14 @@ def configure(updated: set[str]) -> None:
             str(Path(__file__).resolve().parent / "usage-pending"),
             usage_state_id=ctx.options.vm0_usage_state_id or None,
         )
-    if {"vm0_addon_ready_path", "vm0_usage_state_id"} & updated:
-        ready_path = ctx.options.vm0_addon_ready_path
-        usage_state_id = ctx.options.vm0_usage_state_id
-        if ready_path and usage_state_id:
-            runner_flush_lifecycle.start_runner_jsonl_flush_worker()
-            Path(ready_path).write_text(usage_state_id, encoding="utf-8")
+
+
+def running() -> None:
+    ready_path = ctx.options.vm0_addon_ready_path
+    usage_state_id = ctx.options.vm0_usage_state_id
+    if ready_path and usage_state_id:
+        runner_flush_lifecycle.start_runner_jsonl_flush_worker()
+        Path(ready_path).write_text(usage_state_id, encoding="utf-8")
 
 
 def get_api_url() -> str:
--- a/crates/runner/mitm-addon/tests/test_addon_configuration.py
+++ b/crates/runner/mitm-addon/tests/test_addon_configuration.py
@@ -165,7 +165,7 @@ class TestAddonConfiguration:
         state = assert_pending(pending_path, flows=0, buffered=0, reports=0)
         assert state["usageStateId"] == "runner-usage-state-id"
 
-    def test_configure_starts_jsonl_watcher_before_addon_ready(self, tmp_path):
+    def test_running_starts_jsonl_watcher_before_addon_ready(self, tmp_path):
         ready_path = tmp_path / "addon-ready"
         log_path = tmp_path / "network.jsonl"
         (tmp_path / "jsonl-flush-request").write_text(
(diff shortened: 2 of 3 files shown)


seven332:
## Code Review: PR #28846

**Reviewed head:** `33615369cb1e869310bb9b596303be9b913d45d3`

### Summary

The change fixes both sides of the intermittent startup failure: readiness is published only from mitmproxy's post-bind `running()` hook, and initial startup retries only a confirmed selected-port listen collision within a three-attempt budget. Runtime restart keeps the already-published port. Failed startup attempts are fully cleaned and cannot leave a crash notification for the successful replacement.

### Findings

No findings. The public `port()` timing contract is documented, non-port stderr containing a generic address-in-use phrase remains fail-fast, and the regression coverage uses real child processes, sockets, and temporary files.

### Testing

Runner formatting, Clippy, rustdoc, and the complete runner test suite passed (3236 passed, 20 ignored; integration test passed). The locked mitm-addon Ruff, basedpyright, and complete pytest suite also passed (6541 tests). A metal-host benchmark was not rerun locally; the deterministic occupied-socket test covers the triggering process/readiness race.

### Verdict

LGTM

PR #28856, opened 2026-08-24
fix(api): refresh codex token before reset credit

### LINKED ISSUE #28844: bug: refresh Codex token before consuming reset credit
## Background

Codex subscription usage reads now resolve a current OAuth access token through the shared model-provider refresh coordinator. The subscription reset-credit mutation still decrypts and sends the stored `CHATGPT_ACCESS_TOKEN` directly.

## Problem

An access token can expire between listing subscription usage and confirming a reset, callers can use cached provider state, and the reset endpoints can be called directly. In those cases the reset-credit request may send a stale token even though the account has a valid refresh token.

Both reset entry points delegate to `consumePersonalCodexRateLimitResetCredit$`, so token freshness must be enforced at that mutation boundary rather than relying on an earlier list request.

## Requirements

- Resolve the current Codex access token for the exact personal model-provider account immediately before consuming a reset credit.
- Reuse the existing coordinated refresh path, including refresh locking, rotated-token persistence, and reconnect-state handling.
- Preserve the caller-provided idempotency key and send the consume request at most once per API invocation.
- Keep missing accounts or required metadata as not found.
- Do not retry or force-refresh solely because the downstream consume endpoint returns an authentication error.
- Cover the behavior through API-route integration tests using real database state and mocked external OAuth and Codex endpoints.

## Acceptance criteria

- An expired Codex token is refreshed and the reset-credit request uses the refreshed token.
- Account-scoped reset requests refresh and consume against the requested account, not another account.
- A terminal refresh failure marks the account for reconnect and does not call the reset-credit endpoint.
- A transient refresh failure remains retryable and does not falsely mark the account for reconnect.
- Existing current-token reset behavior and idempotency semantics remain unchanged.

## Scope

This issue updates the reset-credit mutation only. Subscription usage-list freshness is already handled separately, and the downstream reset-credit API contract is not being redesigned.

seven332:
## Summary

- resolve a current Codex OAuth access token at the shared reset-credit mutation boundary
- reuse the existing exact-account refresh coordinator, rotated-token persistence, and reconnect-state handling
- preserve one downstream consume attempt with the caller's original idempotency key
- cover legacy reset, inactive exact-account reset, terminal failure, and transient recovery through API-route integration tests

## Behavior

Missing account metadata or a missing source still returns not found. Refresh failure for an existing source reaches the existing internal-error boundary; terminal failures persist reconnect state, while transient failures remain retryable. The reset-credit consume endpoint is never called unless a current token is available.

This PR does not change the public API schema and does not refresh or retry solely because the downstream consume endpoint rejects a request.

## Validation

- `pnpm -F api exec vitest run src/signals/routes/__tests__/me-model-providers-upsert.test.ts src/signals/routes/__tests__/auth-device.bdd.test.ts --maxWorkers=1 --silent=passed-only`
- `pnpm -F api lint`
- `pnpm -F api check-types`
- `pnpm knip --workspace api`
- pre-commit repository Prettier, Knip, type checks, and file-size checks

Closes #28844

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/auth-device.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/auth-device.bdd.test.ts
@@ -1,3 +1,5 @@
+import { randomUUID } from "node:crypto";
+
 import { HttpResponse, http } from "msw";
 import { describe, expect, it } from "vitest";
 import { FeatureSwitchKey } from "@okouai/core/feature-switch-key";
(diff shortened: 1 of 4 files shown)


seven332:
## Code Review: PR #28856

**Reviewed head:** `13655cca95b17ab71f49eb46ba0a6717ac85a6de`

### Summary

The shared reset-credit mutation now resolves the current Codex access token for both the legacy provider route and the exact-account route before making one consume request. It reuses the existing refresh coordinator, token rotation persistence, and reconnect-state handling without changing the public API contract.

### Findings

No findings.

### Testing

- Focused API route integration tests: 41 passed
- API lint: passed
- API type checking: passed
- API Knip: passed
- Pre-commit repository formatting, Knip, type checks, and file-size checks: passed

### Verdict

LGTM

PR #28874, opened 2026-08-24
fix(platform): flatten the model picker category switch and tighten its header

tongx-blip:
## What

Two fixes to the model picker popover, from feedback on the image-model panel.

### 1. The selected category glyph no longer carries a shadow

The trackless switch already tried to opt out with `data-checked:shadow-none`, but tailwind-merge sorts `shadow-segment-selected` into its `shadow-color` group and `shadow-none` into its `shadow` group — the two never conflict, so both classes survived into the DOM and the raised shadow won.

Rather than fight the merge, the raised treatment now lives in a `plain` variant of `SegmentControl`: no track, no opaque fill, no shadow — just the shared selected layer. The picker asks for the variant instead of patching classes.

### 2. Less air between the group label and the first option

The 28px switch is what sets the header row's height, so its `py-2` left the label floating in a 44px band around 12px of ink — 19px of clear space above the first row. `py-1` seats the switch on the same 4px inset the rows sit on, and `-mb-1` takes back the list's `gap-1`, so the header's own padding is the whole distance: 19px → 12px.

The popover's `max-h` tracks the header height (it is sized to show the header plus seven rows), so it moves with it: 302px → 290px.

## Test plan

- [x] Added a regression test asserting the `plain` variant renders neither the raised fill nor its shadow
- [ ] CI: lint, types, tests
- [ ] Visual check on the preview: open the model picker, switch between chat / image / video

### DIFF
--- a/e2e/playwright/tests/chat.spec.ts
+++ b/e2e/playwright/tests/chat.spec.ts
@@ -1206,6 +1206,22 @@ async function cardEdgeAppearance(locator: Locator) {
   });
 }
 
+async function segmentFill(locator: Locator) {
+  return locator.evaluate(async (element) => {
+    // Segments cross-fade when the selection moves, so read them settled.
+    await Promise.all(
+      element.getAnimations().map((animation) => {
+        return animation.finished;
+      }),
+    );
+    const style = getComputedStyle(element);
+    return {
+      backgroundColor: style.backgroundColor,
+      boxShadow: style.boxShadow,
+    };
+  });
+}
+
 test("chat page displays tagline after onboarding", async ({ page }) => {
   await page.goto(appUrl);
   await page.waitForURL(/agents\/.*\/chat/, { timeout: 30_000 });
@@ -1270,6 +1286,33 @@ test("model picker fits seven rows and scrolls one row for eight", async ({
   expect(eightModels.scrollHeight - eightModels.clientHeight).toBe(
     eightModels.rowStep,
   );
+});
+
+test("model picker category switch marks its selection without a raised shadow", async ({
+  page,
+}) => {
+  await mockModelPickerBoundary(page);
+  await page.goto(appUrl);
+  await page.waitForURL(/agents\/.*\/chat/, { timeout: 30_000 });
+
+  await page
+    .getByRole("combobox", { name: "Claude Fable 5", exact: true })
+    .click();
+  const categorySwitch = page.getByRole("radiogroup", { name: "Models" });
+  const chatCategory = categorySwitch.getByRole("radio", { name: "Chat" });
+  const imageCategory = categorySwitch.getByRole("radio", { name: "Image" });
+  await imageCategory.click();
+  await expect(imageCategory).toBeChecked();
+
+  // The switch sits straight on the popover surface with no track to lift off,
+  // so the selection is a flat state layer: readable against its neighbours,
+  // and carrying none of the raised segment's shadow.
+  const [selected, unselected] = await Promise.all([
+    segmentFill(imageCategory),
+    segmentFill(chatCategory),
+  ]);
+  expect(selected.boxShadow).toBe("none");
+  expect(selected.backgroundColor).not.toBe(unselected.backgroundColor);
 });
 
 test("chat composer keeps the model icon unclipped on narrow screens", async ({
--- a/e2e/playwright/tests/chat.spec.ts
+++ b/e2e/playwright/tests/chat.spec.ts
@@ -1251,10 +1251,10 @@ test("model picker fits seven rows and scrolls one row for eight", async ({
 
   const sevenModels = await openModelPickerAndReadGeometry(page);
   expect(sevenModels).toStrictEqual({
-    clientHeight: 300,
+    clientHeight: 288,
     optionCount: 7,
     rowStep: 36,
-    scrollHeight: 300,
+    scrollHeight: 288,
   });
 
   boundary.showEightModels();
@@ -1262,10 +1262,10 @@ test("model picker fits seven rows and scrolls one row for eight", async ({
 
   const eightModels = await openModelPickerAndReadGeometry(page);
   expect(eightModels).toStrictEqual({
-    clientHeight: 300,
+    clientHeight: 288,
     optionCount: 8,
     rowStep: 36,
-    scrollHeight: 336,
+    scrollHeight: 324,
   });
   expect(eightModels.scrollHeight - eightModels.clientHeight).toBe(
     eightModels.rowStep,
--- a/turbo/apps/platform/src/views/okou-page/components/model-provider-picker.tsx
+++ b/turbo/apps/platform/src/views/okou-page/components/model-provider-picker.tsx
@@ -1086,15 +1086,18 @@ function ModelPickerListHeader({
 }) {
   return (
     // Pinned so switching category never means scrolling back up for the
-    // switch. The negative margins bleed the row over the list's own `p-1`
-    // inset so rows scroll under an opaque surface rather than beside it.
+    // switch. The negative side and top margins bleed the row over the list's
+    // own `p-1` inset so rows scroll under an opaque surface rather than
+    // beside it.
     //
-    // `py-2` matches `pr-2`: the switch is a 28px control tucked into the
-    // popover's top-right corner, and 4px above it against 8px beside it read
-    // as a mistake rather than as a tighter grid.
+    // The 28px switch is what sets this row's height, so its padding is what
+    // sets the label's distance from the first option: `py-2` left the label
+    // floating in a band nearly four times its own ink. `py-1` seats the
+    // switch on the same 4px inset the rows sit on, and `-mb-1` takes back the
+    // list's `gap-1` so the header's own padding is the whole distance.
     <div
       data-model-picker-header
-      className="sticky top-0 z-10 -mx-1 -mt-1 flex items-center gap-2 bg-card py-2 pl-3 pr-2"
+      className="sticky top-0 z-10 -mx-1 -mt-1 -mb-1 flex items-center gap-2 bg-card py-1 pl-3 pr-2"
     >
       <span className="min-w-0 flex-1 truncate text-xs font-medium text-muted-foreground">
         {label}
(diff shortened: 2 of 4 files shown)


tongx-blip:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 4171e8a2b4aec849f2a0ff92c0856aa43f4e19ad

### Summary

Two presentational fixes to the model picker popover: the trackless category switch's selected glyph stops carrying the segment control's raised shadow (moved from an ineffective `className` override to a new `plain` variant of `SegmentControl`), and the sticky list header tightens from `py-2` to `py-1` while cancelling the list's `gap-1` beneath it, with the popover's `max-h` following from 302px to 290px. The root cause of the original bug is diagnosed correctly: `shadow-segment-selected` lands in tailwind-merge's `shadow-color` group while `shadow-none` lands in its `shadow` group, so the two never conflicted and both survived into the DOM.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no wrong assumptions | Pass | `segment-control.tsx` moves the checked fill/shadow into a cva `variant` so the two treatments are mutually exclusive by construction; the header/`max-h` arithmetic is confirmed by the updated Playwright geometry assertions (288 client, 324 scroll for eight rows) passing in the PR's CI run. |
| `REVIEW.md` | Type safety; no `any`, no broad types | Pass | New `SegmentControlVariant` union and `SegmentControlContextValue` interface; no casts or `any` in the diff. |
| `REVIEW.md` | Security (auth, injection, secrets) | N/A | The diff changes only Tailwind class literals, a cva variant, and test expectations; no request, auth, storage, or user-input path. |
| `REVIEW.md` | `fix:` commits must carry a regression test | Finding (P1) | `segment-control.test.tsx:54` is the only regression test for the shadow fix and asserts class strings, which cannot observe the rendered shadow — see Findings. |
| `REVIEW.md` | Feature regression guard; an updated test must still prove the invariant | Pass | `chat.spec.ts:1253,1265` keeps `optionCount`, `rowStep`, and the `scrollHeight - clientHeight === rowStep` invariant; only the two numbers that the PR deliberately moves change. |
| `REVIEW.md` | DB / JSONB persisted-contract gate | N/A | No schema, contract, Zod response, or `jsonb().$type<>()` change in the diff. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions, no dynamic `import()`, no hardcoded config | Pass | No suppression comments, no dynamic imports, no URLs or env values in the diff. |
| `docs/bad-smell.md` | Fail fast; no fallback or recovery slop | Pass | The diff deletes a dead `data-checked:shadow-none` override instead of adding a compensating path. |
| `docs/fallback.md` | Declare, bound, and justify every introduced fallback; no tombstone tests | Pass | The diff introduces no fallback and no negative test, so no `Fallbacks` section is required (§9). |
| `docs/testing.md` | Test behavior and outcomes, not internals | Finding (P1) | The new UI-package test asserts the class list rather than the rendered result — see Findings. |
| `docs/testing/anti-patterns.md` | AP-7: do not test implementation details such as CSS classes | Finding (P1) | `turbo/packages/ui/src/components/ui/__tests__/segment-control.test.tsx:54-65`. |
| `docs/testing/anti-patterns.md` | AP-2 / AP-4 / AP-5: mock only at the boundary; no fetch stubs; no fake timers | Pass | The diff adds no mocks, no `vi.mock()`, and no timer control. |
| `docs/testing/testing-external-behavior.md` | Observe results from the surface an external user can see | Finding (P1) | Same root cause as AP-7: the assertion reads the class attribute, not the painted shadow. |
| `docs/testing/app-testing.md` | No component-only unit tests under `turbo/apps/platform`; page tests enter via `detachedSetupPage` | Pass | No platform test is added; the new test lives in `turbo/packages/ui`, and the platform-visible behavior is covered by the Playwright geometry test. |
| `docs/effect.md` | Keep render pure; choose the correct execution boundary | Pass | `model-provider-picker.tsx` changes only className literals inside render; no command, Store write, DOM mutation, or subscription is added. |
| `docs/cache.md` | RC-1/RC-2/RC-7/RC-8: no render-phase effects, render-time signal allocation, unstable refs, or asymmetric teardown | Pass | `modelPickerPanelHeightRef$` and its `onRef`/`ResizeObserver` teardown are untouched; no new state, ref, or resource. |
| `.claude/skills/ccstate/SKILL.md` | `detach(..., Reason.DomCallback)` with `pageSignal$`; rely on `computed` memoization | N/A | The diff touches no ccstate signal, command, or DOM callback. |

### Findings

- `turbo/packages/ui/src/components/ui/__tests__/segment-control.test.tsx:54-65`: the regression test for the shadow fix asserts on the rendered class list (`not.toContain("shadow-segment-selected")`, `toContain("data-checked:bg-state-selected")`). `docs/testing/anti-patterns.md` AP-7 and `docs/testing/testing-external-behavior.md` both name CSS-class assertions as testing implementation rather than behavior, and here that is not just a style objection: the bug being fixed is precisely a case where the class list looked correct (`data-checked:shadow-none` was present) while the browser still painted a shadow. A class-string assertion in jsdom, which computes no Tailwind styles, is structurally unable to catch this bug class and would keep passing if a future base class reintroduced the shadow by another route.

  Please replace it with a real-browser assertion on the rendered outcome. This PR already touches `e2e/playwright/tests/chat.spec.ts`, and that file already has the pattern — `cardEdgeAppearance` reads `getComputedStyle(element).boxShadow` after awaiting animations. The media category switch is reachable in the same flow the geometry test uses, so asserting `box-shadow: none` on the checked segment there covers the fix end to end, and the jsdom class assertion can then be dropped.

### Testing

- Coverage: Partially adequate. The header-spacing half of the fix has a genuine regression test in a real browser (`chat.spec.ts:1253,1265`, which also preserves the seven-rows-fit / eight-rows-overflow invariant). The shadow half has no test that can observe the shadow; the only assertion covering it is the class-list check flagged above.
- Conventions: One violation — AP-7 / testing-external-behavior CSS-class assertion in `segment-control.test.tsx`. No mocking, timer, or direct-fetch violations; no tombstone tests; the existing behavior-style tests in that file are untouched.

tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: a876a1a1273d3a40dad3552cd98969404d5900a9
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

The model picker's trackless category switch stops painting the segment control's raised shadow on its selected glyph, and the sticky list header tightens from `py-2` to `py-1` while cancelling the list's `gap-1` beneath it, with the popover's `max-h` following from 302px to 290px. The shadow fix is expressed as a new `plain` variant of `SegmentControl` rather than a `className` override, which is the right shape: tailwind-merge sorts `shadow-segment-selected` into its `shadow-color` group and `shadow-none` into its `shadow` group, so the old override never conflicted and both classes reached the DOM. The P1 raised on `4171e8a` is resolved — the class-string assertion is gone and the behavior is now covered by a real-browser computed-style check.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no wrong assumptions | Pass | `segment-control.tsx` makes the raised fill and the flat state layer mutually exclusive cva variants; the header/`max-h` arithmetic is pinned by `chat.spec.ts:1269,1281` (288 client, 324 scroll for eight rows), which passed on `4171e8a`. |
| `REVIEW.md` | Type safety; no `any`, no broad types | Pass | `SegmentControlVariant` union plus a `SegmentControlContextValue` interface; the new e2e helper returns an inferred `{backgroundColor, boxShadow}`; no casts. |
| `REVIEW.md` | Security (auth, injection, secrets) | N/A | The diff changes Tailwind class literals, a cva variant, and Playwright expectations only; no request, auth, storage, or user-input path. |
| `REVIEW.md` | `fix:` commits must carry a regression test | Pass | `chat.spec.ts:1288` asserts the checked category segment's computed `box-shadow` is `none` in a real browser — the surface the original bug actually broke; the spacing half is pinned by the geometry test. |
| `REVIEW.md` | Feature regression guard; an updated test must still prove the invariant | Pass | The geometry test keeps `optionCount`, `rowStep`, and `scrollHeight - clientHeight === rowStep`; only the two numbers the PR deliberately moves change. |
| `REVIEW.md` | DB / JSONB persisted-contract gate | N/A | No schema, contract, Zod response, or `jsonb().$type<>()` change. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions, no dynamic `import()`, no hardcoded config | Pass | No suppression comments, dynamic imports, URLs, or env values in the diff. |
| `docs/bad-smell.md` | Fail fast; no fallback or recovery slop | Pass | The diff deletes the dead `data-checked:shadow-none` override instead of layering a compensating path on it. |
| `docs/fallback.md` | Declare, bound, and justify every introduced fallback; no tombstone tests | Pass | No fallback and no negative test introduced, so no `Fallbacks` section is required (§9). The removed jsdom assertion died with the mechanism it covered rather than becoming a tombstone. |
| `docs/testing.md` | Test behavior and outcomes, not internals | Pass | The remaining assertions read rendered geometry and computed style; `segment-control.test.tsx` is back to its original behavior-only tests and is no longer in the diff. |
| `docs/testing/anti-patterns.md` | AP-7: do not test implementation details such as CSS classes | Pass | The class-list assertions from `4171e8a` are removed; nothing in the diff reads `className`. |
| `docs/testing/anti-patterns.md` | AP-2 / AP-4 / AP-5: mock only at the boundary; no fetch stubs; no fake timers | Pass | The new test reuses `mockModelPickerBoundary`, which routes `**/api/*` at the Playwright network boundary; no `vi.mock()`, no timer control. |
| `docs/testing/anti-patterns.md` | AP-8: do not over-test trivial UI state | Pass | The test is not vacuous — alongside `box-shadow: none` it compares the checked segment's fill against an unchecked one, so removing the selection styling entirely would fail it. |
| `docs/testing/testing-external-behavior.md` | Observe results from the surface an external user can see | Pass | The case is constructed by clicking the composer combobox and the Image category, and verified through computed style on the rendered segments. |
| `docs/testing/app-testing.md` | No component-only unit tests under `turbo/apps/platform`; page tests enter via `detachedSetupPage` | Pass | No platform vitest test is added; platform-visible behavior is covered end to end in Playwright. |
| `docs/effect.md` | Keep render pure; choose the correct execution boundary | Pass | `model-provider-picker.tsx` changes only className literals inside render; no command, Store write, DOM mutation, or subscription added. |
| `docs/cache.md` | RC-1/RC-2/RC-7/RC-8: no render-phase effects, render-time signal allocation, unstable refs, or asymmetric teardown | Pass | `modelPickerPanelHeightRef$` and its `onRef`/`ResizeObserver` teardown are untouched; no new state, ref, or resource. |
| `.claude/skills/ccstate/SKILL.md` | `detach(..., Reason.DomCallback)` with `pageSignal$`; rely on `computed` memoization | N/A | No ccstate signal, command, or DOM callback in the diff. |

### Findings

No P0 or P1 findings. The single P1 from the previous review round is resolved:

- `turbo/packages/ui/src/components/ui/__tests__/segment-control.test.tsx` — the class-string regression test is deleted, and `e2e/playwright/tests/chat.spec.ts:1288` replaces it with a computed `box-shadow` assertion on the checked segment plus a fill comparison against an unchecked one. `segmentFill` awaits the segment's own animations first, so the `transition-colors` cross-fade cannot be read mid-flight.

### Testing

- Coverage: Adequate. Both halves of the fix have a real-browser regression test — the flat selection (`box-shadow: none` plus a distinguishable fill) and the tightened header (288px client height, seven rows fitting, eight overflowing by exactly one 36px row step).
- Conventions: Compliant. No mocking, timer, fetch-stub, or tombstone violations; network stubbing stays at the Playwright route boundary; no CSS-class or internal-state assertions remain.

PR #28909, opened 2026-08-24
perf(runner): bound command output capture

### LINKED ISSUE #28884: perf(rust): bound runner command output capture
## Location

- `crates/runner/src/bounded_command.rs:156-181`
- `crates/runner/src/bounded_command.rs:252-258`
- `crates/runner/src/executor/diagnostics/oom.rs:24-39`
- `crates/runner/src/cmd/service/systemctl.rs:84-104`
- `crates/runner/src/workspace_image_cache/fs.rs:151-186`

## Problem

`run_output_bounded` limits child execution and cleanup time, but it does not limit captured output size. Each piped stream is drained by `read_child_output`, which calls `read_to_end` into a growable `Vec<u8>`. A command that emits data quickly can therefore allocate its entire stdout and/or stderr before exiting or before the timeout fires.

The shared helper currently serves three production boundaries:

- the host OOM probe captures both streams from an unfiltered `dmesg` invocation;
- bounded `systemctl` queries capture both streams, including `systemctl cat`;
- workspace sparse copies discard stdout but capture all diagnostic stderr.

This is the same distinction already established by #19083 for `sandbox-fc`: an execution timeout is not a memory limit, semantic stdout and diagnostic stderr need explicit policies, and stopping a pipe reader at the limit without continued draining can deadlock the child.

## Impact

The host `dmesg` probe runs after a Firecracker wait failure, precisely when the runner may already be handling host memory pressure. It materializes the complete kernel ring buffer for each affected run. Concurrent sandbox failures can therefore allocate separate unbounded buffers and worsen the resource incident being diagnosed.

The generic helper also leaves future and current `systemctl` or `cp` callers without an application-level memory boundary. Fast output can exhaust runner memory even though every command has a bounded runtime.

## Acceptance criteria

- Give captured runner command streams explicit byte policies rather than unconditional `read_to_end` growth.
- Discard streams that callers do not use; bound diagnostic stderr while continuing to drain and report truncation.
- Bound semantic stdout and return an explicit output-too-large error (or equivalent) so partial output is never accepted as complete.
- On output overflow, continue draining/discarding or terminate and reap the child safely; do not allow a full pipe to turn the size limit into a hang.
- Preserve host OOM detection, bounded `systemctl` parsing, sparse-copy diagnostics, existing execution deadlines, and cancellation/cleanup guarantees.
- Add focused tests proving fast output above each policy limit keeps memory bounded, semantic overflow cannot report success, diagnostic truncation remains bounded, and the child/output tasks are cleaned up.

## Audit provenance

Found during the daily Rust performance audit of 16 randomly selected files at `f1111eec81c437e3cbb5062154aba95acc8ab9c5`.

seven332:
## Summary

- replace unbounded runner command pipe reads with chunked capture that retains at most 4 MiB of semantic stdout and 64 KiB of diagnostic stderr
- continue draining excess output to EOF so limits cannot deadlock children on full pipes, while failing closed on semantic overflow and marking diagnostic truncation
- apply explicit output policies to host OOM diagnostics, bounded systemd queries, and sparse workspace-image copies without changing their existing deadlines or cleanup ownership
- cover fast oversized stdout/stderr with real subprocess tests alongside the existing timeout, cancellation, reader-cleanup, and child-reap cases

## Scope

- keeps successful output below the limits byte-for-byte compatible
- does not change APIs, protocols, schemas, configuration, persisted data, deployment order, unrelated direct systemd commands, or sandbox process lifecycle

## Validation

- `cargo fmt --all --check`
- repository pre-commit Rust format, documentation, Clippy, and file-size hooks

Closes #28884

### DIFF
--- a/crates/runner/src/bounded_command.rs
+++ b/crates/runner/src/bounded_command.rs
@@ -11,6 +11,10 @@ use crate::child_cleanup::kill_and_reap_child_on_drop;
 
 const COMMAND_CLEANUP_TIMEOUT: Duration = Duration::from_secs(2);
 const BOUNDED_COMMAND_CHILD_LABEL: &str = "bounded command";
+const DEFAULT_SEMANTIC_OUTPUT_LIMIT_BYTES: usize = 4 * 1024 * 1024;
+const DEFAULT_DIAGNOSTIC_OUTPUT_LIMIT_BYTES: usize = 64 * 1024;
+const OUTPUT_READ_CHUNK_BYTES: usize = 8 * 1024;
+const OUTPUT_TRUNCATION_MARKER: &[u8] = b"[output truncated]";
 
 #[derive(Debug)]
 pub(crate) enum BoundedCommandOutcome<T> {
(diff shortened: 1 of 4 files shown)


seven332:
## Code Review: PR #28909

**Reviewed head:** `792e534cf8aa0d429eb92733c8d5eb9140c74085`

### Summary

The runner command boundary now caps semantic stdout and diagnostic stderr while continuing to drain excess bytes, preserves the existing child timeout/cancellation/reap ownership, and applies fail-closed overflow mapping to all current callers. The policy surface and PR scope match the challenged plan.

### Findings

No findings.

### Testing

Real command coverage includes output below and exactly at the limits, semantic overflow, marked diagnostic truncation, oversized dual-stream draining, timeout, cancellation, output-task cleanup, and child reaping. Formatting, Clippy with warnings denied, runner docs, and all-target/all-feature runner tests passed on this head (3242 passed, 20 ignored; guest inventory 1 passed).

Production output-size distributions are not available in the repository; the 4 MiB semantic and 64 KiB diagnostic limits follow the existing sandbox host-command policy, and semantic overflow fails closed.

### Verdict

LGTM

PR #28910, opened 2026-08-24
test(runner): cover claim cooldown transitions

### LINKED ISSUE #28893: test(rust): cover claim cooldown deadline and capacity transitions
## Context

The runner-local claim cooldown state machine lives in crates/runner/src/provider/api_claim_cooldowns.rs:55-122:

https://github.com/vm0-ai/vm0/blob/f1111eec81c437e3cbb5062154aba95acc8ab9c5/crates/runner/src/provider/api_claim_cooldowns.rs#L55-L122

It owns per-run deadlines, capacity saturation, a provider-wide fallback deadline, expiry pruning, refresh, removal, and the retry deadline returned by snapshot. ApiProvider consumes that snapshot to schedule rediscovery and installs the provider-wide POLL_FAST fallback when capacity is saturated:

https://github.com/vm0-ai/vm0/blob/f1111eec81c437e3cbb5062154aba95acc8ab9c5/crates/runner/src/provider/api.rs#L352-L407

## Missing coverage

There are no focused tests in api_claim_cooldowns.rs. The end-to-end saturation test at api.rs:3185-3250 verifies the fallback log fields and only asserts that deferred_poll_at is Some:

https://github.com/vm0-ai/vm0/blob/f1111eec81c437e3cbb5062154aba95acc8ab9c5/crates/runner/src/provider/api.rs#L3185-L3250

That test does not prove that snapshot selects the active provider-wide 5-second deadline instead of a still-active 30-second per-run deadline. It also does not cover the capacity boundary where refreshing an already-recorded run must remain Recorded rather than Saturated, or prove that expired entries free capacity before the next record.

## Impact

A regression in deadline selection can keep the runner from polling again at the intended fast fallback after cooldown saturation. A regression in refresh or prune ordering can unnecessarily switch all candidates to provider-wide blocking or leave the bounded exclusion set saturated after entries expire. The current integration test can remain green in each case because it checks only that some deferred deadline exists.

## Acceptance criteria

- Add deterministic Tokio paused-time tests for ClaimCooldowns.
- Verify that recording new runs up to capacity returns Recorded with the expected active count and sorted snapshot exclusions.
- Verify that re-recording an existing run at capacity refreshes its deadline without returning Saturated.
- Verify that a new run beyond capacity returns Saturated without evicting an unexpired entry.
- Verify that block_all makes snapshot and remaining use the active global deadline, including the production-shaped 5-second global versus 30-second per-run case.
- Advance time across the global and per-run boundaries and verify pruning, fallback to the remaining per-run deadline, and capacity reuse after expiry.
- Verify that remove clears only the requested run and updates the next retry deadline.
- Strengthen the ApiProvider saturation test to assert the scheduled deadline is the intended fast fallback, not merely present.
- No production behavior change is required unless the new tests expose a mismatch.

seven332:
## Summary

- add deterministic paused-time coverage for claim cooldown capacity, refresh, saturation, global precedence, expiry, capacity reuse, and removal transitions
- verify cooldown snapshots keep exclusions sorted and select the expected retry deadline at each boundary
- strengthen the provider saturation test to prove deferred polling uses `POLL_FAST` instead of only checking that a deadline exists

## Scope

- test-only change; no runner production behavior, API, protocol, persistence, migration, or deployment change
- keep the provider scenario on the normal Tokio clock because it exercises a real mock-server I/O boundary; bound its deadline by claim start and completion instants plus `POLL_FAST`

## Validation

- `cargo doc --manifest-path crates/Cargo.toml -p runner --no-deps --all-features`
- `cargo test --manifest-path crates/Cargo.toml -p runner provider::api_claim_cooldowns::tests`
- `cargo test --manifest-path crates/Cargo.toml -p runner provider::api::tests::saturated_claim_cooldowns_gate_repeated_direct_candidates`
- `cargo test --manifest-path crates/Cargo.toml -p runner` (3243 passed, 20 ignored; inventory integration test passed)

Closes #28893

### DIFF
--- a/crates/runner/src/provider/api.rs
+++ b/crates/runner/src/provider/api.rs
@@ -3221,15 +3221,26 @@ mod tests {
         let first = provider.discover().await.unwrap();
         assert!(provider.claim(first).await.is_none());
         let overflow = provider.try_discover_ready().await.unwrap();
+        let claim_started_at = tokio::time::Instant::now();
         let (claim, events) = capture_api_provider_events(provider.claim(overflow)).await;
+        let claim_finished_at = tokio::time::Instant::now();
         assert!(claim.is_none());
         let event = captured_event(&events, "claim failed, candidate cooldown capacity reached");
         assert_eq!(event_field(event, "retry_scope"), "provider");
         assert_eq!(event_field(event, "retry_after_ms"), "5000");
         assert_eq!(event_field(event, "active_cooldowns"), "1");
+        let deferred_poll_at = wakeups
+            .snapshot()
+            .await
+            .deferred_poll_at
+            .expect("saturation should defer HTTP polling");
         assert!(
-            wakeups.snapshot().await.deferred_poll_at.is_some(),
-            "saturation should defer HTTP polling"
+            deferred_poll_at >= claim_started_at + POLL_FAST,
+            "deferred polling should use at least the fast fallback from claim start"
+        );
+        assert!(
+            deferred_poll_at <= claim_finished_at + POLL_FAST,
+            "deferred polling should use at most the fast fallback from claim completion"
         );
 
         push_direct_candidate_for_test(
--- a/crates/runner/src/provider/api_claim_cooldowns.rs
+++ b/crates/runner/src/provider/api_claim_cooldowns.rs
@@ -120,3 +120,135 @@ fn prune_expired(state: &mut ClaimCooldownState, now: Instant) {
         state.global_deadline = None;
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn run_id(value: u128) -> RunId {
+        RunId::from(uuid::Uuid::from_u128(value))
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #28910

**Reviewed head:** `f6203478d0c93c3a71fe2c538282122333de9460`

### Summary

The test-only change matches the challenged plan and covers the claim cooldown transition matrix without changing production behavior. Direct paused-time coverage is appropriate under the repository's state-machine exception, and the provider assertion preserves the real HTTP integration boundary while proving use of `POLL_FAST`.

### Findings

No findings.

### Testing

Rust formatting, runner Clippy with all targets/features and denied warnings, runner documentation, focused state-machine and provider tests, and the complete runner suite passed locally. The provider timing assertion also passed 20 consecutive focused runs. Unrelated Turbo and Python checks were not run; GitHub CI remained to be verified.

### Verdict

LGTM

PR #28981, opened 2026-08-24
fix(python): drain retained diagnostics after webhook shutdown

### LINKED ISSUE #28961: bug(python): retry retained Anthropic accounting after webhook executor shutdown
## Summary

Anthropic incomplete-accounting reports retained while webhook delivery is saturated are not retried after the webhook executor frees its capacity during shutdown.

`anthropic_accounting.retry_all_pending()` attempts admission from `crates/runner/mitm-addon/src/anthropic_accounting.py:103-121`. The shutdown lifecycle invokes that retry from `crates/runner/mitm-addon/src/runner_flush_lifecycle.py:223-228`, before `mitm_addon.done()` joins the webhook executor. If all 16 delivery slots are still occupied at that point, the report remains in `_retained_reports` with its `BufferedReportLease`.

`crates/runner/mitm-addon/src/mitm_addon.py:1967-1973` then calls `usage_executor.shutdown(wait=True)`, which allows the existing deliveries to finish and releases their capacity, but the post-join path only calls `usage.drain_usage_events_after_executor_shutdown()`. It never retries the Anthropic retained-report queue, so `done()` can return while that report and its buffered ownership remain in memory.

## Reproduction

Verified at `7d91b6bb470128e2a4598218a636692040a03f4e` with both a deterministic queued executor and the real `ThreadPoolExecutor`:

1. Block all four webhook workers and enqueue 16 filler deliveries, filling `MAX_PENDING_WEBHOOK_PAYLOADS`.
2. Record an incomplete Anthropic SSE accounting report; admission is saturated, so the report is retained.
3. Start `mitm_addon.done()`. Its pre-join retained retry still observes saturated delivery capacity.
4. Release the workers, allowing `usage_executor.shutdown(wait=True)` to finish and all 16 delivery slots to be released.
5. Observe that `done()` returns with no pending delivery slots, but `_retained_reports` still contains the Anthropic report and the runner-visible pending snapshot remains `buffered=1, reports=0`. Only the 16 filler requests were delivered.

The existing related suite passes (`637 passed`), so this specific post-executor-join retention path currently has no regression coverage.

## Impact

A shutdown that reaches this capacity window can discard the telemetry explaining an incomplete Anthropic SSE accounting result when the process exits. It also leaves the runner-visible buffered count nonzero through the end of the addon lifecycle, forcing the outer bounded shutdown path to proceed with unresolved retained ownership.

The runner's repeated flush requests reduce the normal window, but they do not close the final race when capacity becomes available only during executor shutdown, including timeout-driven or direct mitmproxy shutdown.

## Acceptance criteria

- Retained Anthropic accounting reports are revisited after webhook executor capacity becomes available during shutdown.
- `done()` reaches a terminal outcome for each retained report under the bounded shutdown policy: delivered once or explicitly discarded with a confirmed diagnostic.
- Each retained report's `BufferedReportLease` is released exactly once, leaving `buffered=0` after the terminal shutdown path.
- Add a regression test that saturates the real webhook executor, releases its workers during `done()`, and verifies that the retained Anthropic telemetry is not left in memory or silently lost.

seven332:
## Summary

- drain retained billing and diagnostic delivery after the webhook executor joins
- reuse one ordered Anthropic, Claude, and Codex diagnostic retry path before and after executor shutdown
- cover the saturated Anthropic shutdown race with a real executor and local HTTP server

## Behavior

When all webhook delivery slots are occupied during the pre-shutdown flush, an incomplete Anthropic accounting report remains retained. Once the executor join releases those slots, shutdown now retries it through the existing synchronous delivery fallback, releases its buffered ownership exactly once, and reaches zero pending counters.

The same final lifecycle pass includes the two existing provider-output timing stores because they share the retained diagnostic boundary. This PR does not change admission limits, HTTP retry policy, timeout policy, payloads, APIs, schemas, dependencies, or deployment protocols.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_model_provider_sse_usage_anthropic.py tests/test_done_hook.py` (36 passed)
- `uv run --no-sync python -m pytest tests/` (6593 passed)

Closes #28961

### DIFF
--- a/crates/runner/mitm-addon/src/mitm_addon.py
+++ b/crates/runner/mitm-addon/src/mitm_addon.py
@@ -1960,8 +1960,10 @@ def done():
     so its worker shutdown stops new forwards and best-effort closes active
     upstream sockets without waiting for slow upstream responses. JSONL writer
     shutdown is also bounded and best-effort; if it times out, process shutdown
-    continues with accepted log entries possibly still pending. Model-provider
-    failure delivery stops admission and receives one bounded drain window.
+    continues with accepted log entries possibly still pending. After joining
+    the usage executor, retained billing and diagnostic work is drained through
+    synchronous delivery. Model-provider failure delivery stops admission and
+    receives one bounded drain window.
     """
     try:
         runner_flush_lifecycle.drain_and_close()
@@ -1969,7 +1971,7 @@ def done():
         try:
             try:
                 usage.webhook.usage_executor.shutdown(wait=True)
-                usage.drain_usage_events_after_executor_shutdown()
+                runner_flush_lifecycle.drain_delivery_work_after_executor_shutdown()
             finally:
                 auth_base_forwarder.shutdown_forward_request_workers(wait=False)
         finally:
--- a/crates/runner/mitm-addon/src/runner_flush_lifecycle.py
+++ b/crates/runner/mitm-addon/src/runner_flush_lifecycle.py
@@ -223,9 +223,25 @@ def _flush_usage_for_runner_request() -> None:
 def _flush_delivery_work(*, trigger: _DeliveryFlushTrigger) -> None:
     """Admit billing work before retained diagnostic reports."""
     usage.flush_usage_events(trigger=trigger)
+    _retry_retained_diagnostic_reports()
+
+
+def _retry_retained_diagnostic_reports() -> None:
+    """Retry every retained diagnostic source in its established order."""
     anthropic_accounting.retry_all_pending()
     claude_output_timing.retry_all_pending()
     codex_output_timing.retry_all_pending()
+
+
+def drain_delivery_work_after_executor_shutdown() -> None:
+    """Synchronously drain work after the usage executor has been joined.
+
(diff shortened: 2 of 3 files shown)


seven332:
## Code Review: PR #28981

**Reviewed head:** `f129577700d36d08c7dca758e04c3b3a6bcdb0f5`

### Summary

The shutdown lifecycle now drains retained usage and diagnostics after executor join through the existing synchronous delivery contract. The diagnostic source list remains centralized, billing priority is preserved, and the saturated Anthropic regression exercises real hook, executor, HTTP, and counter behavior.

### Findings

No findings.

### Testing

Validated uv lock/sync, Ruff format and lint, BasedPyright, 36 focused shutdown/Anthropic tests, and the complete 6593-test mitm-addon suite. Unrelated Turbo and Rust suites were not run because this PR changes only the Python add-on. Production frequency of the saturation window was not measured; provider-specific duplicate end-to-end cases were not added for Claude/Codex because they use the same unchanged retry-store contract.

### Verdict

LGTM

PR #28983, opened 2026-08-24
perf(mitm-addon): reuse x unicode label classification

### LINKED ISSUE #28965: perf(python): reuse Unicode label normalization in X post URL detection
## Problem

At `7d91b6bb470128e2a4598218a636692040a03f4e`, `crates/runner/mitm-addon/src/usage/providers/connectors/x_billing.py:393-429` splits a Unicode bare-domain candidate into labels and calls `normalize_idna_label()` for every label. Repeated labels are normalized independently, and `crates/runner/mitm-addon/src/usage/providers/connectors/x_billing.py:376-390` also repeats the same Unicode-prefix checks for each occurrence.

This is production-reachable for a successful billable `POST /2/tweets`. `crates/runner/mitm-addon/src/usage/providers/connectors/x.py:644-668` decodes and refines the request body synchronously, and `crates/runner/mitm-addon/src/mitm_addon.py:1681-1683,1846-1848` invokes connector usage reporting inline from the mitmproxy response hook. A long plain-text post can therefore block unrelated response processing while identical Unicode labels repeatedly traverse the full shared normalization, validation, punycode, and bidirectional-text pipeline.

The existing flow test at `crates/runner/mitm-addon/tests/x_connector_usage/test_writes.py:821-843` says that near-limit Unicode labels avoid repeated IDNA work, but it asserts only the billing result and does not constrain normalization work.

## Evidence

The following compact valid JSON body stays below both the documented 25,000-character X longform-post limit and `REQUEST_BODY_BILLING_INSPECTION_LIMIT`:

```python
text = "a." + (("é" * 20 + ".") * 1_189) + "notatld"
```

On Python 3.12.3, five warmed direct calls to `_tweet_create_body_is_plain_text_without_url()` produced:

| Repeated Unicode labels | Text characters | Body bytes | `normalize_idna_label()` calls | Median |
| 100 | 2,109 | 4,120 | 102 | 15.190 ms |
| 200 | 4,209 | 8,220 | 202 | 30.303 ms |
| 400 | 8,409 | 16,420 | 402 | 60.263 ms |
| 800 | 16,809 | 32,820 | 802 | 121.842 ms |
| 1,189 | 24,978 | 48,769 | 1,191 | 178.488 ms |

The classifier correctly returns plain text, but runtime and IDNA calls grow once per repeated label. As a diagnostic only, wrapping the existing normalizer with a cache scoped to one classification call reduced the 1,189-label case to a 10.000 ms median with three cached label values. This isolates repeated label normalization as the dominant removable cost; it is not a proposal for process-global caching.

## Distinction from closed work

- #22177 / PR #22199 bypassed generic IDNA normalization only for ordinary all-ASCII candidates and explicitly kept Unicode candidates on the shared path.
- #24850 / PR #24854 stopped the regex from restarting quadratically inside one failed Unicode label. This reproduction yields one successful candidate and spends its time after extraction in per-label classification.
- #23592 / PR #23671 added a shared ASCII fast path; the repeated labels here are non-ASCII.

No open issue matching Unicode label normalization in X post detection was found.

## Acceptance criteria

- Repeated identical Unicode labels within one X post candidate do not each repeat full IDNA normalization and Unicode-prefix classification work; use bounded request- or candidate-local reuse, or an equivalent bounded approach.
- The 24,978-character reproduction still refines to `content.create`, quantity 1, through the connector usage flow.
- A deterministic regression constrains normalization or label-classification work for repeated Unicode labels without using a wall-clock CI threshold.
- Existing protocol URL, bare-domain, Unicode/IDNA, canonical and invalid A-label, compatibility-mapping, TLD-prefix, and boundary classifications remain unchanged.
- A reproducible before/after benchmark records a material reduction for the 100/200/400/800/1,189-label cases.
- No user-input-driven process-global cache, new dependency, API, schema, migration, persisted state, or asynchronous usage lifecycle is introduced solely for this optimization.

seven332:
## Summary

- memoize complete IDNA normalization outcomes and Unicode-prefix classifications within each X bare-domain candidate
- preserve occurrence-specific unsafe compatibility, invalid-label, TLD, boundary, and early-exit behavior
- turn the exact 24,978-character connector-flow reproduction into a deterministic regression that bounds repeated classification work
- retain the existing ordinary ASCII fast path and avoid persistent user-derived cache state

## Explicit exclusions

- no request-wide or process-global cache
- no API, schema, migration, dependency, configuration, persisted-state, or async usage lifecycle change

## Benchmark

Python 3.14.0, six direct calls per input with the first timing discarded, using the same instrumented benchmark before and after:

| Repeated labels | Before calls | After calls | Before median | After median |
| 100 | 102 | 3 | 6.016 ms | 0.251 ms |
| 200 | 202 | 3 | 10.842 ms | 0.306 ms |
| 400 | 402 | 3 | 23.375 ms | 0.508 ms |
| 800 | 802 | 3 | 46.595 ms | 0.875 ms |
| 1,189 | 1,191 | 3 | 66.385 ms | 1.237 ms |

## Validation

- `uv lock --check`
- `uv sync --locked`
- focused X write regressions: 37 passed
- `uv run --no-sync python -m pytest tests/x_connector_usage/`: 227 passed
- `uv run --no-sync python -m pytest tests/`: 6,593 passed
- pre-commit Ruff and file-size hooks
- commitlint

Closes #28965

### DIFF
--- a/crates/runner/mitm-addon/src/usage/providers/connectors/x_billing.py
+++ b/crates/runner/mitm-addon/src/usage/providers/connectors/x_billing.py
@@ -73,6 +73,11 @@ class _BodyRefinementRule(NamedTuple):
     matches_body: Callable[[bytes | None], bool]
 
 
+class _BillingDomainLabelNormalization(NamedTuple):
+    normalized_label: str | None
+    unsafe_compatibility_mapping: bool
+
+
 # Permission → default bucket.  The default matches the majority of
 # paths in the scope; outliers go in `_PATH_OVERRIDES`.  Prices are
 # defined in turbo/apps/api/src/scripts/dev-seed.ts.
@@ -390,6 +395,15 @@ def _label_has_tld_prefix_before_unicode(label: str) -> bool:
     return False
 
 
+def _normalize_billing_domain_label(label: str) -> _BillingDomainLabelNormalization:
+    try:
+        return _BillingDomainLabelNormalization(normalize_idna_label(label), False)
+    except UnsafeIdnaCompatibilityMappingError:
+        return _BillingDomainLabelNormalization(None, True)
+    except UnicodeError:
+        return _BillingDomainLabelNormalization(None, False)
+
+
 def _bare_domain_candidate_likely_contains_url(candidate: str, following_char: str) -> bool:
     if candidate.isascii():
         simple_ascii_candidate = candidate.lower()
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #28983

**Reviewed head:** `85d151ff53222ba12d6a5ba0b7bf7fd09570f23c`

### Summary

The candidate-local cache removes repeated Unicode label normalization and prefix classification without changing occurrence-specific unsafe/invalid handling, ordered early exits, or persistent state. The implementation stays within the approved one-PR scope and does not affect APIs, schemas, dependencies, or rollout compatibility.

### Findings

No findings.

### Testing

Ruff format/lint, BasedPyright, uv lock/sync, 227 X connector tests, and all 6,593 mitm-addon tests passed. A 20,580-case differential check matched the `main` implementation across ASCII, Unicode, A-label, invalid, compatibility, TLD, repetition, and boundary combinations. The requested benchmark reduced the 1,189-label case from 1,191 to 3 normalizer calls and from 66.385 ms to 1.237 ms locally.

Production label-distribution data was unavailable, so absolute workload-level impact remains unverified; the deterministic work bound and local target benchmark are verified.

### Verdict

LGTM

PR #28985, opened 2026-08-24
fix(python): reject negative x response counts

### LINKED ISSUE #28958: bug(python): reject negative X response counts
## Problem

X response count metadata is validated only as a non-boolean Python integer. `crates/runner/mitm-addon/src/usage/providers/connectors/x_response_inspection.py:73-74` therefore accepts negative values, and `:112-118` forwards negative `meta.result_count` and `meta.total_tweet_count` into billing metadata. The duplicate validator in `crates/runner/mitm-addon/src/usage/providers/connectors/x.py:165-166` has the same domain gap.

For a parsed count-endpoint response, `x.py:525-526` treats a negative `response_total_tweet_count` as present and chooses it as the primary count. `x.py:543-548` then emits no usage event because the count is not positive. Finally, `x.py:747-752` checks only whether the value is `None`, so the negative value also suppresses the existing `missing_count_endpoint_total` underbilling diagnostic.

This is reproducible on `main` at `7d91b6bb470128e2a4598218a636692040a03f4e`:

```text
parse_json_response_fields_from_body({meta: {result_count: -2, total_tweet_count: -1}})
=> {response_result_count: -2, response_total_tweet_count: -1, ...}

_compute_billable_counts(GET, {is_count_endpoint: true}, {body_parsed: true, response_total_tweet_count: -1}, posts.read)
=> {}
```

## Impact

A schema-invalid negative count from X silently drops the connector usage event. For count endpoints it also drops the audit signal intended to make skipped billing visible, so operators cannot distinguish this case from a legitimate zero-result response. Negative `result_count` values likewise enter the normal response-count calculation instead of being rejected at inspection time.

## Acceptance criteria

- Treat `meta.result_count` and `meta.total_tweet_count` as non-negative integers; booleans, negative integers, and non-integers are invalid count metadata.
- A parsed count-endpoint response with a negative `total_tweet_count` emits no usage event and records the existing or an equivalent confirmed underbilling diagnostic.
- Add coverage for negative values through the incremental/selective response inspection path and the buffered fallback path.
- Preserve zero as a valid count and keep existing positive-count behavior unchanged.

seven332:
## Summary

- treat X `result_count` and `total_tweet_count` metadata as non-negative integers
- share one response-count normalizer across inspection and usage reporting
- preserve parsed response data while routing invalid count-endpoint totals to the existing confirmed-underbilling diagnostic
- cover incremental response finalization and buffered usage reporting

## Behavior

Negative X response counts are now omitted from billing metadata instead of being treated as present. This does not fail the run or invalidate unrelated response observations. A parsed count endpoint with a negative `total_tweet_count` emits no usage event and records `missing_count_endpoint_total`, matching the existing behavior for a missing total. Zero and positive counts are unchanged.

This PR does not change the generic selective JSON parser, add fallback billing estimates, introduce a new diagnostic reason, or touch API/protocol/persisted-state compatibility.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_x_response_parsers.py tests/x_connector_usage/test_reads.py` (80 passed)
- `uv run --no-sync python -m pytest tests/` (6594 passed)
- `git diff --check`

Closes #28958

### DIFF
--- a/crates/runner/mitm-addon/src/usage/providers/connectors/x.py
+++ b/crates/runner/mitm-addon/src/usage/providers/connectors/x.py
@@ -34,7 +34,10 @@ from .x_billing import (
     include_billing_category,
     refine_bucket_with_body,
 )
-from .x_response_inspection import parse_json_response_fields_from_body
+from .x_response_inspection import (
+    as_non_negative_response_count,
+    parse_json_response_fields_from_body,
+)
 
 # HTTP 2xx success range (RFC 9110). Also defined in ``response_streaming.py``
 # and ``x_response_inspection.py``; kept local to avoid a constants module for a
@@ -160,10 +163,6 @@ def _compile_request_fallback_hint_policies(
 _REQUEST_FALLBACK_HINT_POLICIES = _compile_request_fallback_hint_policies(
     _REQUEST_FALLBACK_HINT_POLICY_SPECS
 )
-
-
-def _as_non_bool_int(value: object) -> int | None:
-    return value if isinstance(value, int) and not isinstance(value, bool) else None
 
 
 def _count_bounded_non_empty_comma_segments(value: str, max_count: int) -> int | None:
@@ -522,7 +521,7 @@ def _compute_billable_counts(
 
     if resp_meta.get("body_parsed"):
         if req_meta.get("is_count_endpoint"):
-            total = _as_non_bool_int(resp_meta.get("response_total_tweet_count"))
+            total = as_non_negative_response_count(resp_meta.get("response_total_tweet_count"))
             primary = 0 if total is None else total
         else:
             # Body was parsed — trust actual response counts.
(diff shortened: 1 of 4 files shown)


seven332:
## Code Review: PR #28985

**Reviewed head:** `6871a75b35bde0f16e4fbe20ecc26859ad9c8757`

### Summary

The PR correctly rejects negative X response count metadata at one shared domain boundary, preserves zero and positive values, and keeps invalid external fields from failing the run or invalidating unrelated response observations. Count-endpoint underbilling remains visible through the existing confirmed diagnostic.

### Findings

No findings.

### Testing

- Incremental response finalization covers negative result and total counts while preserving data observations.
- Buffered usage reporting covers no usage plus `missing_count_endpoint_total` for a negative total.
- Existing zero and positive cases remain in the focused/full suites.
- Lock/sync, Ruff format/lint, BasedPyright, focused pytest (80 passed), complete addon pytest (6594 passed), and diff whitespace checks passed.
- Not verified against a live X API response; controlled real-flow integration tests cover the local behavior.

### Verdict

LGTM

PR #28988, opened 2026-08-24
test(python): cover sigv4 repeated query value ordering

### LINKED ISSUE #28968: test(python): cover SigV4 value ordering for repeated query keys
## Problem

AWS SigV4 canonical query strings must sort URI-encoded pairs by key and then by value. The production implementation does this correctly in `crates/runner/mitm-addon/src/aws_sigv4.py:558-561` by sorting the complete encoded `(key, value)` tuples.

The only focused repeated-key signature fixture is `crates/runner/mitm-addon/tests/test_aws_sigv4.py:404-426`. It signs `Tag=one&Tag=two`, whose duplicate values are already in canonical order. A repository-wide test search found no exact-signature case where duplicate values arrive in reverse canonical order.

A disposable mutation changed `_canonical_query_string()` to sort by encoded key only, preserving input order among equal keys. All 147 tests in `test_aws_sigv4.py` and `test_firewall_aws_sigv4_auth.py` still passed. With `Tag=two&Tag=one`, the current signer produces signature `d4315768cfa9fa4ffbb5c84b941e4d45205fb824bb4372eca6da8fa0d58065d0`; the key-only mutation produces `3e5db647d15c05a37f144bc435cf1e34bfa90c62bd0aa706377087e47da5905c`.

The current implementation is correct; this issue is for the missing regression guard.

## Impact

A future canonicalization regression could generate invalid signatures for legitimate AWS requests that use repeated query parameters in noncanonical arrival order. AWS would reject those requests even though the signer continues to preserve the ordinary query pairs and all existing SigV4 tests remain green.

## Acceptance criteria

- Add a public `sign_request()` presigned-query test with at least one ordinary query key repeated in reverse canonical value order.
- Verify the generated signature against an independent deterministic value or oracle, rather than only asserting that the ordinary pairs are preserved or that two signatures differ.
- Keep the outgoing ordinary query-pair representation behavior explicit and separate from the canonical sorting assertion.
- Include an encoded-value case if needed to prove ordering is based on URI-encoded key/value pairs.
- Ensure a mutation that sorts canonical query pairs by key only fails the new test.
- Keep the test deterministic, offline, and suitable for the normal mitm-addon test suite.

seven332:
## Summary

- strengthen the public SigV4 presigned-query fixture with repeated values whose URI-encoded canonical order differs from input order
- assert outgoing ordinary query-pair order separately from the independently verified exact signature
- verify the guard fails under a disposable key-only canonical sort mutation

## Exclusions

- no production signer behavior changes
- no dependency or lockfile changes

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (6592 passed)
- disposable key-only sort mutation: focused test failed on the signature as expected, then passed after restoration

Closes #28968

### DIFF
--- a/crates/runner/mitm-addon/tests/test_aws_sigv4.py
+++ b/crates/runner/mitm-addon/tests/test_aws_sigv4.py
@@ -401,10 +401,11 @@ def test_sign_request_reinspects_changed_representation(
     assert prepared_result == ordinary_result
 
 
-def test_presigned_query_preserves_ordinary_pairs() -> None:
+def test_presigned_query_preserves_pairs_and_sorts_encoded_values() -> None:
+    # %C3%A9 sorts before z after SigV4 URI encoding, opposite the input order.
     url = aws_sigv4_presigned_url(
         STS_HOST,
-        leading_query="Tag=one&Tag=two&Flag",
+        leading_query="Tag=z&Tag=%C3%A9&Flag",
     )
 
     signed_url, _headers = sign_request(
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #28988

**Reviewed head:** `f5ad2d6955a70a650e771bbb78ab24ef55fb0dd8`

### Summary

The PR strengthens the existing public SigV4 presigned-query fixture with repeated values whose URI-encoded canonical order differs from their input order. It explicitly preserves outgoing ordinary pair order while pinning the independently verified signature, without changing production code or dependencies.

### Findings

No findings.

### Testing

- Complete mitm-addon validation passed: uv lock/sync, Ruff format and lint, BasedPyright, and all 6,592 pytest cases.
- A disposable key-only canonical sort mutation made the focused test fail on the exact signature, then the test passed after restoring production code.
- The expected signature was independently reproduced with botocore; no live AWS request was run because the regression suite is intentionally deterministic and offline.
- GitHub CI was still in progress at review time.

### Verdict

LGTM

PR #28990, opened 2026-08-24
test(python): cover websocket frame limits across network read splits

### LINKED ISSUE #28969: test(python): cover WebSocket frame limits across network read splits
## Problem

`crates/runner/mitm-addon/src/websocket_framing.py:214-232` intentionally reserves one data-frame budget unit only when a wire frame starts. A single WebSocket frame may invoke `frame_inbound_payload_data()` multiple times when its bytes arrive across multiple transport reads, so counting callbacks instead of frame starts would make the limit depend on arbitrary network segmentation.

The read-splitting test at `crates/runner/mitm-addon/tests/test_mitmproxy_websocket_framing.py:761-795` sends one frame across three `DataReceived` events, but it leaves `MAX_MESSAGE_DATA_FRAMES` at 8,192 and does not assert frame-budget state. The low frame-limit cases at `:798-919` deliver each logical frame in a separate, complete transport read. No test combines the two dimensions.

A disposable mutation changed `_DecodedMessageLimit.frame_inbound_payload_data()` to reserve a frame for every non-control payload callback. All 21 focused framing tests still passed. In a deterministic probe with `MAX_MESSAGE_DATA_FRAMES = 1`, one valid server-originated data frame split across three network reads is accepted by the current implementation, while the mutation emits close code 1009 on the second read.

The current implementation is correct; this issue is for the missing regression guard.

## Impact

A future framing refactor could count transport read chunks as WebSocket frames and falsely reject healthy long-lived streams based on packetization, TLS record boundaries, or proxy buffering. The same wire message could succeed or fail depending only on how the network divides its bytes, while the current test suite remains green.

## Acceptance criteria

- Add real mitmproxy/wsproto pipeline coverage with `MAX_MESSAGE_DATA_FRAMES = 1` for one data frame whose wire bytes are split across multiple `DataReceived` events.
- Exercise both client-to-server and server-to-client directions so masked and unmasked framing are covered.
- Assert the message completes exactly once, reaches the message hook and forwarding path, remains open, and consumes exactly one data-frame budget unit while incomplete.
- Add or retain a contrast case proving that two actual data frames exceed the same limit regardless of transport-read partitioning and close with code 1009 without partial message delivery.
- Ensure a mutation that reserves a frame on every payload callback fails the new test.
- Keep aggregate and per-message budget cleanup assertions intact after completion or rejection.

seven332:
## Summary

- cover one WebSocket data frame split across multiple transport reads under a one-frame budget
- verify masked and unmasked traffic reaches the hook and forwarding path exactly once
- contrast split transport reads with a second actual data frame that closes with code 1009
- assert per-message and aggregate budget cleanup after completion and rejection

## Scope

- test-only change in the mitmproxy/wsproto integration suite
- no production framing, API, dependency, data, or deployment behavior changes

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (6,594 passed)
- focused test passes for both directions and fails for both under the temporary callback-counting mutation

Closes #28969

### DIFF
--- a/crates/runner/mitm-addon/tests/test_mitmproxy_websocket_framing.py
+++ b/crates/runner/mitm-addon/tests/test_mitmproxy_websocket_framing.py
@@ -795,6 +795,148 @@ async def test_read_split_message_reuses_mutable_fragment(
     assert running.flow.websocket.messages[-1].content == b"split-across-reads"
 
 
+@pytest.mark.parametrize("from_client", [True, False])
+async def test_message_frame_limit_counts_frames_across_read_splits(
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+    from_client: bool,
+) -> None:
+    monkeypatch.setattr(websocket_framing, "MAX_MESSAGE_DATA_FRAMES", 1)
+
+    with (
+        patch.object(mitm_addon, "__file__", str(tmp_path / "mitm_addon.py")),
+        taddons.context(Proxyserver(), mitm_addon) as addon_context,
+    ):
+        accepted = await _start_websocket(addon_context)
+        accepted_content = b"split-across-reads"
+        accepted_wire = _peer(from_client=from_client).send(_message_event(accepted_content))
+        accepted_header_size = len(accepted_wire) - len(accepted_content)
+        accepted_source = _bounded_source_websocket(
+            accepted,
+            from_client=from_client,
+        )
+        accepted_budget = accepted_source._vm0_message_limit._budget
+
+        first_read = await _handle_event(
+            addon_context,
+            accepted,
+            events.DataReceived(
+                _source_connection(accepted, from_client=from_client),
+                accepted_wire[: accepted_header_size + 1],
+            ),
+        )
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #28990

**Reviewed head:** `3b557597a99d3ea5d5711770d5734484d9c063c4`

### Summary

The test-only change covers the real mitmproxy/wsproto pipeline in both masked and unmasked directions. It distinguishes multiple transport reads of one wire frame from a second actual continuation frame, verifies hook-before-forward behavior and exact message cardinality, and checks per-message and aggregate cleanup after acceptance and rejection.

### Findings

No findings.

### Testing

- `uv lock --check` and `uv sync --locked`
- Ruff format and lint passed
- basedpyright passed with 0 errors, warnings, or notes
- full mitm-addon suite passed: 6,594 tests
- focused cases passed in both directions and failed in both directions under the temporary callback-counting mutation

CI was still running when this review was prepared; no runtime production code changed.

### Verdict

LGTM

PR #29015, opened 2026-08-24
fix(composer): repair the feedback note structure webkit collapses mid-composition

bingjiez666:
## Summary

On Safari (desktop, mobile web, and the installed PWA), quoting a passage and then typing into the feedback note can send only the first part of the message. A desktop Safari trace pinned the moment: **~20–30 ms after an IME composition update, WebKit removes the note's wrapper elements** (`noteDom` and `contentDOM`, taking the note text with them) and leaves a bare `<br>` — the visible "composer gets shorter" moment, because the note's `min-h-[96px]` lived on the removed elements.

The node view's `ignoreMutation` ignored every mutation outside `contentDOM`, which also discarded the record for that removal. ProseMirror therefore never noticed and never repaired the structure, and **every keystroke typed afterwards** (composition or plain English) landed outside the now-detached `contentDOM`, was discarded the same way, and stayed visible on screen while missing from the document, the draft PATCH, and the submission.

Fixes are gated behind the new `composerNoteStructureRepair` switch (default off, `bingjie@vm0.ai` only):

- **Reattach on collapse**: when a mutation arrives while `contentDOM` is out of the item, remove what the browser left in the gap and reattach the detached wrapper subtree — it still holds `contentDOM` and every text node, so screen = document is restored with nothing lost. Stranger damage returns `false` so ProseMirror's own dirty-node redraw rebuilds the view.
- **Narrowed `ignoreMutation`**: only mutations the node view produces itself (the quote chip subtree, attribute writes on its wrappers, the localized placeholder) are ignored; structural damage stays visible to ProseMirror. Attribute records on `contentDOM` stay visible because the `composerSubmitDomReconcile` resync marker depends on them.
- **Value-guarded renders**: `render()` writes only real changes when the switch is on. An unchanged-value assignment still queues an observer record, and once records on `contentDOM` are no longer ignored, per-keystroke redundant writes would force a whole-node re-parse and break a live composition.

With the switch off the node view behaves byte-for-byte as before this PR.

## Why this differs from the reverted attempts

#28045 flushed the DOM observer at submit — a no-op here because the collapse record never survives `ignoreMutation`, so there is nothing to flush. #28168 only skipped redundant writes and kept the broad `ignoreMutation`, so the collapse was still swallowed. This PR removes the swallowing itself; the write guards return only as a prerequisite for the narrowed `ignoreMutation`, not as the fix. The failure now reproduces in desktop Safari (quote → compose Chinese in the note), so the fix is directly falsifiable on a desktop rather than only on devices.

Related to #27787.

## Rollout

- default: disabled
- `bingjie@vm0.ai`: enabled

## Verification

- `pnpm exec prettier --write <changed files>` — clean
- `pnpm run lint` in `apps/platform` — exit 0
- `pnpm --filter @okouai/core run check-types`, `pnpm --filter @okouai/app run check-types` — clean
- `pnpm knip` — exit 0
- `pnpm --filter @okouai/core exec vitest run src/__tests__/feature-switch.test.ts` — 25 passed
- `pnpm --filter @okouai/app exec vitest run src/views/okou-page/__tests__/chat-inline-feedback.test.tsx` — 34 passed, including two new page tests that reproduce the traced collapse (wrapper removed + `<br>` inserted with the DOM observer live): switch on → structure repaired, note text preserved, later input reaches the submission; switch off → later input dropped (documents the current failure)
- `pnpm --filter @okouai/app exec vitest run src/views/okou-page/__tests__/chat-run-queue.test.tsx` — 32 passed
- negative control: with the repair callback forced to `false`, the new enabled-path test fails
- device verification: enable the switch, quote a passage, compose Chinese into the note until the composer height drops, keep typing, send — the full text must arrive

### DIFF
--- a/turbo/apps/platform/src/signals/external/feature-switch.ts
+++ b/turbo/apps/platform/src/signals/external/feature-switch.ts
@@ -79,6 +79,12 @@ export const composerRestoredAttachmentValidationEnabled$ = computed(
   },
 );
 
+export const composerNoteStructureRepairEnabled$ = computed((get): boolean => {
+  return (
+    get(featureSwitch$)[FeatureSwitchKey.ComposerNoteStructureRepair] ?? false
+  );
+});
+
 export const codexFastModeEnabled$ = computed((get): boolean => {
   return get(featureSwitch$)[FeatureSwitchKey.CodexFastMode] ?? false;
 });
--- a/turbo/apps/platform/src/signals/okou-page/tiptap-workflow-composer.ts
+++ b/turbo/apps/platform/src/signals/okou-page/tiptap-workflow-composer.ts
@@ -32,7 +32,10 @@ import {
 import type { WorkflowSummary } from "@okouai/api-contracts/contracts/workflows";
 import { agents$ } from "../agent.ts";
 import { currentChatAgentRecordId$ } from "../agent-chat.ts";
-import { composerSubmitDomReconcileEnabled$ } from "../external/feature-switch.ts";
+import {
+  composerNoteStructureRepairEnabled$,
+  composerSubmitDomReconcileEnabled$,
+} from "../external/feature-switch.ts";
 import { logger } from "../log.ts";
 import { sentryLogContext } from "../../lib/sentry-config.ts";
 import { onRef, resetSignal } from "../utils.ts";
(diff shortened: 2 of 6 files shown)


bingjiez666:
Closing: superseded by the flat quote block rebuild (#29137), which restores ProseMirror's own damage visibility and self-healing structurally instead of layering detection and reattachment on the legacy structure. The branch is kept; reopen if the flat block's dogfood surfaces a need for an independent safety net.

PR #29042, opened 2026-08-24
feat: roll out image and video model selection to all users

bingjiez666:
## Summary

Graduate `imageModelSelection` and `videoModelSelection` from the staff-org rollout to GA. Both switches were `enabled: false` + `enabledOrgIdHashes: STAFF_ORG_ID_HASHES`; they now match the registry's GA shape (`enabled: true`, no identity allowlist), the same shape `usagePackPlans` uses.

Every gate stays in place. The keys remain registered, so `POST /api/feature-switches` and the Lab page can still turn either picker off per user, and both `isFeatureEnabled` call sites (composer pickers, thread/member pin resolution, `image-io-generate`, `video-io-generate`, run image/video model snapshots) are untouched.

## What changes for users

- The chat composer model picker exposes the image and video model rows for everyone, not just staff.
- New chat threads pin the member's image/video default when one is set; runs receive the resolved image model (member default, otherwise the vm0 global default) through `OKOU_DEFAULT_IMAGE_MODEL` and the matching system-prompt section.

## Cross-version compatibility

GA moves this feature out of the "feature switch needs no fallback" exemption in `docs/fallback.md` §2, so both boundaries were checked and neither needs a compatibility branch:

- **Old runner → new backend** (drains up to 2h): runs now carry `OKOU_DEFAULT_IMAGE_MODEL`. Runners pass environment through, and a commit-addressed older CLI without `run-default-image-model.ts` ignores an unknown variable — it falls back to the CLI's own default rather than erroring.
- **Old web client → new backend** (~2d): `getAllFeatureStates` returns every key and the API echoes them as `registeredEffectiveSwitches`, which the client overlays on its bundled registry. An already-open old page therefore also resolves `true`, and the picker UI it needs already ships in that bundle.

No request or response shape changed, so new-frontend → old-backend is unaffected.

## Test updates

Five suites read the *registry* default instead of setting an explicit switch state, so the GA flip changed what they observed:

- `turbo/packages/core/src/__tests__/feature-switch.test.ts` — the rollout matrix now expects both keys `true` for a non-staff org. This is the assertion that pins the behavior change itself.
- `turbo/apps/api/src/signals/routes/__tests__/chat-threads-create.test.ts` — the two "picker disabled" cases now write explicit `false` overrides instead of leaning on the default, keeping coverage of the off path that the per-user kill switch still reaches.
- `turbo/apps/platform/src/views/okou-page/__tests__/chat-composer-model-selection-and-providers.test.tsx` — six cases assert a chat-only picker shape. `page-helper.ts` seeds the switch cache from `getAllFeatureStates()`, so any key a test omits follows the registry; each case now pins the media switches it means to exercise.
- `turbo/apps/api/src/signals/routes/__tests__/agentphone.bdd.test.ts` and `.../cron-execute-morning-briefs.test.ts` — both asserted that the caller's prompt section sits immediately before `# Restricted Explicit Content`. The run-level `# Default built-in image model` section now sits between them, so both helpers assert that section is present and correctly ordered, then compare the caller-supplied tail that ends where it begins.

All other tests already set the switch explicitly (`image-io-generate`, `chat-events.bdd`, the Playwright helper), so they are unaffected.

## Verification

- `pnpm format`, `pnpm check-types` (14/14), `pnpm -F api run lint`, `pnpm -F @okouai/app run lint` all pass.
- `chat-composer-model-selection-and-providers.test.tsx` runs locally at 82/82.
- API suites need a Postgres server this sandbox does not provide, so they are validated by the PR pipeline. Note that the first run's `test-api` shards 1/3/5/6/7/8 were *cancelled* by fail-fast rather than passing; the current head is the first to validate the GA flip against all eight shards.
- The repository-wide `pnpm turbo run lint` could not complete locally (SIGKILL at the 3072 MiB heap the `api` task requires, against a 3.9 GB sandbox), so lint was run per affected workspace and the full-repository form is covered by the `lint-*` CI jobs.


bingjiez666:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 06fd95ba9cda6ce06a59665ee52eb2447468cff3

### Summary

Graduates `videoModelSelection` and `imageModelSelection` from staff-org gating to GA (`enabled: true`, no identity allowlist), matching the registry's existing GA shape. The remaining five files update tests that read the registry default instead of setting an explicit switch state, so they broke when the default flipped.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | `feat:` commits must have integration tests | Pass | `turbo/packages/core/src/__tests__/feature-switch.test.ts` asserts the new GA state for a non-staff org, which is exactly the behavior the diff changes. Feature behavior itself is already covered by the switch-on tests in `chat-events.bdd.test.ts` and `image-io-generate.test.ts`. |
| `REVIEW.md` | Feature Regression Guard — existing flow most likely to regress stays covered | Pass | The switch-off path is preserved rather than deleted: `chat-threads-create.test.ts` and the composer tests now pin `false` explicitly, so the disabled behavior is still exercised after GA. |
| `REVIEW.md` | New user-facing features gated behind a `FeatureSwitchKey` | Pass | The keys stay registered, so `POST /api/feature-switches` and the Lab page remain a per-user kill switch. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, contract, or `jsonb(...).$type<...>()` change in the diff. |
| `docs/fallback.md` §3 | Do not fall back on states the contract cannot produce | **Finding (P1)** | `agentphone.bdd.test.ts:194` and `cron-execute-morning-briefs.test.ts:1630` tolerate the absence of the `# Default built-in image model` section, which GA makes unconditionally present in both tests. |
| `docs/fallback.md` §9 | Every fallback introduced by the diff must be declared in the PR summary | **Finding (P1)** | Neither tolerance branch appears in the PR body, and the body still describes only two updated test suites. |
| `docs/fallback.md` §1 | No negative tests asserting removed behavior stays removed | Pass | The new `false` overrides assert live product behavior (a user-toggleable switch), not deleted code, so they are not tombstones. |
| `docs/fallback.md` §2 | Non-GA switch needs no compatibility code | Pass | This PR is the GA boundary itself; §2 stops applying here and `docs/deployment-compatibility.md` takes over, assessed below. |
| `docs/deployment-compatibility.md` | Old runner → new backend | Pass | Runs now carry `OKOU_DEFAULT_IMAGE_MODEL`. Runners pass environment through, and a commit-addressed older CLI without `run-default-image-model.ts` ignores an unknown variable — degraded to the CLI's own default, never an error. |
| `docs/deployment-compatibility.md` | Old frontend → new backend | Pass | `getAllFeatureStates` returns every key and the API echoes them as `registeredEffectiveSwitches`, which the client overlays on its bundled registry. An already-open old page therefore also resolves `true`, and the picker UI it needs is already in that bundle. |
| `docs/deployment-compatibility.md` | New frontend → old backend | Pass | No request or response shape changed; the new client sends nothing an older API cannot answer. |
| `docs/bad-smell.md` §1–2 | No `any`, no lint/type suppressions | Pass | No `any`, no `eslint-disable`/`@ts-*` comments anywhere in the diff. |
| `docs/bad-smell.md` §7 | Fail fast rather than defaulting | **Finding (P1)** | Same two tolerance branches — they convert a missing prompt section into a silently passing assertion instead of a loud failure. |
| `docs/testing.md` | Integration tests only, mock at the boundary | Pass | All changes are to existing route-level and `detachedSetupPage` tests; no new unit tests, no internal mocking introduced. |
| `docs/testing/anti-patterns.md` AP-4 / AP-2 / AP-5 | No internal mocks, no fetch stubbing, no fake timers | Pass | The diff adds only feature-switch overrides and string-slicing assertions; it introduces no `vi.mock`, no `vi.stubGlobal("fetch")`, no `vi.useFakeTimers()`. |
| `docs/testing/anti-patterns.md` AP-7 | Do not test implementation details | **Finding (P1)** | The `agentphone` helper now reaches further into prompt-assembly internals (a named section list) to keep an adjacency claim it can no longer make directly. |
| `docs/testing/api-testing.md` | Drive routes through the real Hono app and contract | Pass | `chat-threads-create.test.ts` sets the switch through the real `feature-switches` route via `updateFeatureSwitchesForUser`, not by writing DB rows. |
| `docs/testing/app-testing.md` | Use `detachedSetupPage` and MSW, no direct component rendering | Pass | All six composer cases keep the `detachedSetupPage` entry point and only add `featureSwitches` options. |
| `docs/testing/testing-external-behavior.md` | Assert observable external behavior | Pass | Assertions remain on rendered composer UI, created-thread events, and the run's `appendSystemPrompt` — all external surfaces. |
| `.claude/skills/ccstate/SKILL.md` | Signals/commands practices for `apps/api` and `apps/platform` | N/A | The diff contains no `computed`, `command`, Store, or signal code — only the core registry constant and test files. |

### Findings

- `turbo/apps/api/src/signals/routes/__tests__/agentphone.bdd.test.ts:194` and `turbo/apps/api/src/signals/routes/__tests__/cron-execute-morning-briefs.test.ts:1630`: both helpers strip the `# Default built-in image model` section *only if* it is found (`if (sectionIndex !== -1)`, `defaultImageModelIndex !== -1 ? ... : restrictedContentIndex`). After this PR neither test overrides `imageModelSelection`, so the registry default makes that section unconditionally present in every one of these runs. The guarded branch is unreachable, and it is the harmful kind of unreachable: if the image-model prompt ever stopped being emitted, both tests would silently fall through to the other branch and still pass, instead of failing loudly. This is `docs/fallback.md` §3 ("a fallback ... turns a loud bug into wrong behavior") and `docs/bad-smell.md` §7. Assert presence instead of tolerating absence, e.g. compute the index, `expect(index).toBeGreaterThan(-1)`, then slice — which keeps the original intent and stays loud.

- PR summary: the diff introduces the two tolerance branches above and the summary does not mention them, which `docs/fallback.md` §9 makes a P1 on its own. The body is also stale in a way that matters for review: it states "Two suites asserted the registry default" and "Local `vitest` was not run", but the PR now touches five test files and the app suite was run locally (82/82). Please update the summary to match the current diff and record the tolerance branches (or remove them per the finding above, in which case no `Fallbacks` section is needed).

- `turbo/apps/api/src/signals/routes/__tests__/agentphone.bdd.test.ts:189`: `runLevelTrailingSections` is a single-element array iterated by a `for` loop. Per the project's YAGNI rule this is a premature abstraction for one value; a direct index-and-slice is clearer and is what `cron-execute-morning-briefs.test.ts` already does. The two files currently solve the identical problem two different ways in the same PR — please make them consistent.

### Testing

- **Coverage: Adequate.** The registry flip's own behavior is pinned by `feature-switch.test.ts`, and the disabled path survives via explicit `false` overrides rather than being deleted — which is the specific regression this GA flip could have caused.
- **Conventions: Compliant.** No internal mocking, no fetch stubbing, no fake timers, no direct component rendering. Switch state is written through the real feature-switches route on the API side and through `detachedSetupPage` options on the platform side.
- **CI evidence:** the current head is fully green — 76 successful checks, 18 skipped, zero failures. This matters because the previous run's `test-api` shards 1/3/5/6/7/8 were *cancelled* by fail-fast rather than passing, so the GA flip had not actually been validated against them until now.
- **Note on the `agentphone` helper:** the adjacency assertion is genuinely weaker after this change — it can no longer detect an arbitrary new section landing between the integration block and the restricted-content policy. That weakening is unavoidable given the run-level section now sits there, but it is worth stating explicitly rather than leaving implicit in the diff.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: abdcf6709494e7df782713a67968915f96dd53e9
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Graduates `videoModelSelection` and `imageModelSelection` from staff-org gating to GA (`enabled: true`, no identity allowlist), matching the registry's existing GA shape. Five test suites that read the registry default instead of pinning an explicit switch state are updated to state their intent.

Re-review of `abdcf67` after the three P1 findings on `06fd95b` were addressed. All three are resolved:

1. **Unreachable tolerance branches — fixed.** `agentphone.bdd.test.ts:191` and `cron-execute-morning-briefs.test.ts:1625` no longer fall through when `# Default built-in image model` is absent. Both now assert the section is present (`toBeGreaterThan(-1)`) and correctly ordered (`toBeLessThan(restrictedContentIndex)`) before slicing at it. A regression that dropped the section would now fail loudly in both suites instead of silently passing.
2. **Stale/undeclared PR summary — fixed.** The body now describes all five updated suites, states why they were affected (`page-helper.ts` seeds the switch cache from `getAllFeatureStates()`), records the local-verification scope including the workspace-scoped lint fallback, and adds a cross-version compatibility section. With the tolerance branches gone the diff introduces no fallback, so no `Fallbacks` section is required per `docs/fallback.md` §9.
3. **Inconsistent helpers — fixed.** The single-element `runLevelTrailingSections` array and its `for` loop are deleted. Both files now use the identical index-assert-slice shape.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | `feat:` commits must have integration tests | Pass | `turbo/packages/core/src/__tests__/feature-switch.test.ts` asserts the new GA state for a non-staff org, which is exactly the behavior the diff changes. |
| `REVIEW.md` | Feature Regression Guard — the flow most likely to regress stays covered | Pass | The switch-off path is preserved rather than deleted: `chat-threads-create.test.ts` and the composer cases pin `false` explicitly, so disabled behavior is still exercised after GA. |
| `REVIEW.md` | New user-facing features gated behind a `FeatureSwitchKey` | Pass | Both keys stay registered, so `POST /api/feature-switches` and the Lab page remain a per-user kill switch. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, contract, or `jsonb(...).$type<...>()` change in the diff. |
| `docs/fallback.md` §3 | Do not fall back on states the contract cannot produce | Pass | Resolved in `abdcf67`: both helpers assert presence and ordering instead of tolerating absence. |
| `docs/fallback.md` §9 | Every fallback introduced by the diff must be declared in the PR summary | Pass | The diff no longer introduces a fallback; no `Fallbacks` section is required. |
| `docs/fallback.md` §1 | No negative tests asserting removed behavior stays removed | Pass | The `false` overrides assert live product behavior (a user-toggleable switch), not deleted code, so they are not tombstones. |
| `docs/fallback.md` §2 | Non-GA switch needs no compatibility code | Pass | This PR is the GA boundary itself; §2 stops applying and `docs/deployment-compatibility.md` takes over, assessed below. |
| `docs/deployment-compatibility.md` | Old runner → new backend | Pass | Runs now carry `OKOU_DEFAULT_IMAGE_MODEL`. Runners pass environment through, and a commit-addressed older CLI without `run-default-image-model.ts` ignores an unknown variable — degraded to the CLI's own default, never an error. Declared in the PR summary. |
| `docs/deployment-compatibility.md` | Old frontend → new backend | Pass | `getAllFeatureStates` returns every key and the API echoes them as `registeredEffectiveSwitches`, which the client overlays on its bundled registry, so an already-open old page resolves `true` and its bundle already contains the picker UI. |
| `docs/deployment-compatibility.md` | New frontend → old backend | Pass | No request or response shape changed. |
| `docs/bad-smell.md` §1–2 | No `any`, no lint/type suppressions | Pass | No `any`, no `eslint-disable`/`@ts-*` comments anywhere in the diff. |
| `docs/bad-smell.md` §7 | Fail fast rather than defaulting | Pass | Resolved in `abdcf67`; the assertions now fail loudly on a missing section. |
| `docs/testing.md` | Integration tests only, mock at the boundary | Pass | All changes are to existing route-level and `detachedSetupPage` tests; no new unit tests, no internal mocking. |
| `docs/testing/anti-patterns.md` AP-2 / AP-4 / AP-5 | No fetch stubbing, no internal mocks, no fake timers | Pass | The diff adds only feature-switch overrides and index assertions; no `vi.mock`, `vi.stubGlobal("fetch")`, or `vi.useFakeTimers()`. |
| `docs/testing/anti-patterns.md` AP-7 | Do not test implementation details | Pass | The helper still asserts one observable property of the run's `appendSystemPrompt` — section presence and ordering — rather than reaching into prompt-assembly internals through a named section list. |
| `docs/testing/api-testing.md` | Drive routes through the real Hono app and contract | Pass | `chat-threads-create.test.ts` sets switches through the real `feature-switches` route via `updateFeatureSwitchesForUser`, not by writing DB rows. |
| `docs/testing/app-testing.md` | Use `detachedSetupPage` and MSW, no direct component rendering | Pass | All six composer cases keep the `detachedSetupPage` entry point and only add `featureSwitches` options. |
| `docs/testing/testing-external-behavior.md` | Assert observable external behavior | Pass | Assertions remain on rendered composer UI, created-thread events, and the run's `appendSystemPrompt`. |
| `.claude/skills/ccstate/SKILL.md` | Signals/commands practices for `apps/api` and `apps/platform` | N/A | The diff contains no `computed`, `command`, Store, or signal code — only the core registry constant and test files. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None. The three P1 findings raised against `06fd95b` are resolved in `abdcf67`.

### Testing

- **Coverage: Adequate.** The registry flip is pinned by `feature-switch.test.ts`, and the disabled path survives via explicit `false` overrides rather than being deleted — the specific regression this GA flip could have caused.
- **Conventions: Compliant.** No internal mocking, no fetch stubbing, no fake timers, no direct component rendering. Switch state is written through the real feature-switches route on the API side and through `detachedSetupPage` options on the platform side.
- **CI evidence:** the previous head `06fd95b` was fully green — 76 successful checks, 18 skipped, zero failures, including all eight `test-api` shards. That matters because the first run's shards 1/3/5/6/7/8 were *cancelled* by fail-fast rather than passing, so the GA flip was not actually validated against them until `06fd95b`. `abdcf67` changes two assertion helpers only and is re-running now.
- **Local verification of `abdcf67`:** `pnpm format`, `pnpm check-types` (14/14), and `pnpm -F api run lint` all pass. API suites need a Postgres server this sandbox does not provide, so the two changed API assertions are validated by the pipeline.
- **Residual note:** the adjacency assertion is narrower than before — it pins the caller block against the image-model section rather than against the end of the prompt, so a future run-level section inserted *after* the image-model one would not be caught. That is inherent to the prompt now having a run-level tail, and the PR summary records it.

PR #21319, opened 2026-07-13
fix(platform): validate composer model state without extra renders

e7h4n:
## Summary

- isolate closed composer controls from model-policy, billing, connector, and dialog subscriptions; mount full model-picker data only while the menu is open
- expose Fast availability, Limited Free 1 status, OAuth availability, and composer send status as primitive ccstate projections
- make existing-thread send, queue, and new-thread commands read the current selection themselves and validate the current policy route, Limited Free 1 admission, and personal OAuth state before any draft, optimistic, or persistence mutation
- block mouse and keyboard submission while async model validation is pending, and preserve inline feedback when validation rejects the send
- sanitize stale Fast selections against the current feature switch and Codex route, preserve valid Fast mode when queueing, and keep prompt URL/draft state intact when a prompt send is rejected
- retain loaded model options across billing refresh errors and remove the obsolete React OAuth blocker helper

## Performance verification

- ChatThreadComposer executed twice during a measured send: idle to sending to idle
- closed model picker components, open-only policy content, and SelectItem executed zero times during the send
- ThinkingIndicator animation remained isolated and did not wake the composer
- failed or pending model validation no longer changes pendingSendCount or wakes the composer as sending

## Tests

- `pnpm -F @vm0/app exec vitest run` for the three directly affected page-test files: 214 tests passed
- `pnpm -F @vm0/app check-types`
- `pnpm -F @vm0/app lint`, including static assets, Oxlint, type-aware Oxlint, and ESLint
- pre-commit root Knip and 13/13 workspace typechecks passed

Full Vitest was not run, per request.

PR #29046, opened 2026-08-24
fix(platform): keep the model picker measurement row hidden on category switch

bingjiez666:
## Problem

Switching the model picker's category on mobile flashes a stray model name over the panel header for a moment. Reported from an iPhone recording; frame-stepping the 120fps capture puts the artifact on screen for 14 frames (117ms), and it reads as the selected chat model's name wrapped one word per line on top of `Video models`.

## Cause

`model-picker-panel-height.ts` reacts to the list resizing by animating the popup's height (180ms) and fading the new rows in (120ms). The fade loop walks every direct child of the select list and skips only the sticky header:

```ts
for (const row of list.children) {
  if (row.matches(PANEL_HEADER_SELECTOR)) continue;
  row.animate([{ opacity: 0.4 }, { opacity: 1 }], { duration: 120 });
}
```

A media category replaces the model rows, so the picker keeps the selected chat model in the list as a measurement-only item — `absolute left-0 top-0 h-8 w-px overflow-hidden opacity-0`, `disabled`, `aria-hidden` — so the select can still resolve its label. That item is a direct child of the list, so the loop animates it too, and a keyframe outranks the class that hides it: its opacity is driven from 0.4 to 1 for exactly the fade's duration, which is the 117ms measured in the capture.

On iOS Safari the running animation promotes the row to its own compositing layer, which escapes both its own 1px `overflow: hidden` clip and the header's stacking order, so the name paints in full over the header. Engines that keep the clip render the row as an invisible 1px sliver, which is why this only shows on mobile.

The measurement row landed in #21319; the fade loop that lights it up landed in #28335.

## Fix

Skip `aria-hidden` rows alongside the header. Measurement rows are not content and have nothing to fade in.

## Test

New Playwright case in `chat.spec.ts`: after clicking the Image category, the measurement row's computed opacity must still be `0`. The click and the read share one `evaluate` because a Playwright round trip between them can outlast the 120ms fade and miss the row while it is lit. The assertion is engine-independent — the row is forced visible everywhere, only the clip escape is WebKit-specific — so Chromium CI catches a regression.

## Verification

- `pnpm check-types` in `e2e/` passes
- `prettier@3.8.1` reports the touched `turbo/` file unchanged
- Remaining checks left to the PR pipeline

### DIFF
--- a/e2e/playwright/tests/chat.spec.ts
+++ b/e2e/playwright/tests/chat.spec.ts
@@ -1315,6 +1315,54 @@ test("model picker category switch marks its selection without a raised shadow",
   expect(selected.backgroundColor).not.toBe(unselected.backgroundColor);
 });
 
+test("model picker category switch keeps its measurement row hidden", async ({
+  page,
+}) => {
+  await mockModelPickerBoundary(page);
+  await page.goto(appUrl);
+  await page.waitForURL(/agents\/.*\/chat/, { timeout: 30_000 });
+
+  await page
+    .getByRole("combobox", { name: "Claude Fable 5", exact: true })
+    .click();
+  const imageCategory = page
+    .getByRole("radiogroup", { name: "Models" })
+    .getByRole("radio", { name: "Image" });
+  await expect(imageCategory).toBeVisible();
+
+  // A media category replaces the model rows, so the selected chat model stays
+  // in the list as a 1px, transparent row the select can still measure. The
+  // swap fades the rows it brings in, and a keyframe outranks the class that
+  // hides that row: fading it printed the model name over the header for the
+  // whole fade. The click and the read share one evaluate because a round trip
+  // between them can outlast the fade and miss the row while it is lit.
+  const measurementRowOpacity = await imageCategory.evaluate(
+    async (segment) => {
+      if (!(segment instanceof HTMLElement)) {
+        throw new Error("Model picker category segment is not an HTML element");
+      }
+      segment.click();
+      // The fade starts once the swapped list has laid out, so let one frame
+      // carry the resize and read on the next.
+      await new Promise<void>((resolve) => {
+        requestAnimationFrame(() => {
+          requestAnimationFrame(() => {
+            resolve();
+          });
+        });
+      });
+      const measurementRow = document.querySelector(
+        '[data-slot="select-list"] [data-slot="select-item"][aria-hidden="true"]',
+      );
+      if (measurementRow === null) {
+        throw new Error("Model picker has no measurement row");
+      }
+      return getComputedStyle(measurementRow).opacity;
+    },
+  );
+  expect(measurementRowOpacity).toBe("0");
+});
+
 test("chat composer keeps the model icon unclipped on narrow screens", async ({
   page,
 }) => {
--- a/e2e/playwright/tests/chat.spec.ts
+++ b/e2e/playwright/tests/chat.spec.ts
@@ -1336,31 +1336,42 @@ test("model picker category switch keeps its measurement row hidden", async ({
   // hides that row: fading it printed the model name over the header for the
   // whole fade. The click and the read share one evaluate because a round trip
   // between them can outlast the fade and miss the row while it is lit.
-  const measurementRowOpacity = await imageCategory.evaluate(
-    async (segment) => {
-      if (!(segment instanceof HTMLElement)) {
-        throw new Error("Model picker category segment is not an HTML element");
-      }
-      segment.click();
-      // The fade starts once the swapped list has laid out, so let one frame
-      // carry the resize and read on the next.
-      await new Promise<void>((resolve) => {
+  const swap = await imageCategory.evaluate(async (segment) => {
+    if (!(segment instanceof HTMLElement)) {
+      throw new Error("Model picker category segment is not an HTML element");
+    }
+    segment.click();
+    // The fade starts once the swapped list has laid out, so let one frame
+    // carry the resize and read on the next.
+    await new Promise<void>((resolve) => {
+      requestAnimationFrame(() => {
         requestAnimationFrame(() => {
-          requestAnimationFrame(() => {
-            resolve();
-          });
+          resolve();
         });
       });
-      const measurementRow = document.querySelector(
-        '[data-slot="select-list"] [data-slot="select-item"][aria-hidden="true"]',
-      );
-      if (measurementRow === null) {
-        throw new Error("Model picker has no measurement row");
-      }
-      return getComputedStyle(measurementRow).opacity;
-    },
-  );
-  expect(measurementRowOpacity).toBe("0");
+    });
+    const list = document.querySelector('[data-slot="select-list"]');
+    if (list === null) {
+      throw new Error("Model picker has no list");
+    }
+    const measurementRow = list.querySelector(
+      '[data-slot="select-item"][aria-hidden="true"]',
+    );
+    const modelRows = list.querySelector('[data-slot="select-group"]');
+    if (measurementRow === null || modelRows === null) {
+      throw new Error("Model picker list has no measurement row or model rows");
+    }
+    return {
+      measurementRowOpacity: getComputedStyle(measurementRow).opacity,
+      modelRowsOpacity: getComputedStyle(modelRows).opacity,
+    };
+  });
+  expect(swap.measurementRowOpacity).toBe("0");
+  // The rows the swap brings in are mid-fade at that same instant. Asserting
+  // that keeps the swap covered from both sides: the fade the picker still
+  // owes its rows, and proof that the sample landed inside the fade rather
+  // than after it, where a hidden row reads as transparent either way.
+  expect(Number(swap.modelRowsOpacity)).toBeLessThan(1);
 });
 
 test("chat composer keeps the model icon unclipped on narrow screens", async ({
--- a/turbo/apps/platform/src/signals/okou-page/model-picker-panel-height.ts
+++ b/turbo/apps/platform/src/signals/okou-page/model-picker-panel-height.ts
@@ -37,6 +37,14 @@ const SELECT_LIST_SELECTOR = '[data-slot="select-list"]';
 const PANEL_HEADER_SELECTOR = "[data-model-picker-header]";
 
 /**
+ * A media category replaces the model rows, so the selected chat model stays in
+ * the list as a 1px, transparent row the select can still measure. A keyframe
+ * outranks the class that hides it, so fading that row in would print the model
+ * name over the header for the length of the fade.
+ */
+const PANEL_MEASUREMENT_ROW_SELECTOR = '[aria-hidden="true"]';
+
+/**
  * Animates the model picker popup between the heights its category panels ask
  * for. Attach to the popup; it finds the list underneath on its own.
  *
@@ -100,7 +108,10 @@ export const modelPickerPanelHeightRef$ = onRef(
         if (!(row instanceof HTMLElement)) {
           continue;
         }
-        if (row.matches(PANEL_HEADER_SELECTOR)) {
+        if (
+          row.matches(PANEL_HEADER_SELECTOR) ||
+          row.matches(PANEL_MEASUREMENT_ROW_SELECTOR)
+        ) {
           continue;
         }
         row.animate([{ opacity: PANEL_CONTENT_FADE_FROM }, { opacity: 1 }], {


bingjiez666:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 8c1c5d2e5a12296354f5fa323cd49fb693b4c255

### Summary

The category-switch fade in `model-picker-panel-height.ts` walked every direct child of the select list and skipped only the sticky header, so it also animated the `aria-hidden`, `opacity-0` measurement row the picker keeps in the list while a media category is open. A keyframe outranks the class that hides that row, so it was driven to full opacity for the fade's 120ms; on iOS Safari the resulting compositing layer escapes the row's 1px `overflow` clip and the header's stacking order, printing the selected model name over the category title. The fix skips `aria-hidden` rows alongside the header and adds a Playwright regression case.

The production change is correct and exactly scoped. The regression test, however, can pass without ever observing the defect, and it leaves the surviving half of the behavior — real rows still fade — uncovered.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic: no wrong assumptions or unhandled paths | Pass | `model-picker-panel-height.ts:111`: the only direct children of `[data-slot="select-list"]` carrying `aria-hidden` are the two measurement `SelectItem`s (`model-provider-picker.tsx:700`, `:1235`); every other `aria-hidden` in that file is an icon nested inside a row, so the new guard cannot swallow a content row. |
| `REVIEW.md` | Type safety: no `any`, no broad casts | Pass | Both files add only string constants and `instanceof` narrowing; no casts. |
| `REVIEW.md` | Security | N/A | No auth, input handling, network, or persistence surface is touched. |
| `REVIEW.md` | `fix:` commits must carry a regression test | Finding (P1) | `chat.spec.ts:1316`: a test exists, but it asserts `opacity === "0"` without establishing that the sample landed inside the 120ms fade, so it also passes when no fade ran at all. |
| `REVIEW.md` | Feature Regression Guard: cover the existing flow most likely to regress | Finding (P1) | The diff narrows an existing animation's target set; nothing asserts that non-measurement rows still fade, so silently killing the fade would keep every test green. |
| `REVIEW.md` | Fallback discipline / `Fallbacks` section in the summary | N/A | The diff introduces no `??`/`||` default, compatibility branch, or cross-version tolerance; it removes an unintended animation target. No `Fallbacks` section required. |
| `REVIEW.md` | New user-facing features gated by a `FeatureSwitchKey` | N/A | No new feature; this restores the intended behavior of a shipped picker that already sits behind `imageModelSelection`. |
| `REVIEW.md` | DB/JSONB review gate, deployment compatibility | N/A | No schema, contract, persisted shape, or cross-service boundary is touched — the change is client-only and takes effect on reload. |
| `docs/bad-smell.md` | Zero `any`, zero lint/type suppressions, no dynamic imports | Pass | Neither file adds a suppression comment, `any`, or `import()`. |
| `docs/bad-smell.md` | Fail fast instead of silent fallback | Pass | `model-picker-panel-height.ts` keeps its existing `throw` for a missing list; the new test throws on a missing segment or measurement row rather than returning a default. |
| `docs/bad-smell.md` | Comments explain WHY, not WHAT | Pass | Both new comments state the cascade reason (a keyframe outranks the hiding class) rather than restating the code. |
| `docs/effect.md` | DOM lifecycle work belongs in `onRef`, bound to the element's `AbortSignal` | Pass | The change stays inside the existing `onRef` command's `ResizeObserver` callback; owner, observer, and teardown are unchanged. |
| `docs/effect.md` | Render stays pure; no Store writes or resource creation during render | Pass | No view file is touched. |
| `docs/cache.md` | RC-8 symmetric resource lifecycles (observer/animation acquisition and release) | Pass | No new observer, listener, or timer; the diff only reduces the set of elements passed to `Element.animate`, whose animations are already `fill: none` and non-retained. |
| `docs/cache.md` | RC-3 no unbounded module-lifetime caches | Pass | The added module constant is a static selector string. |
| `.claude/skills/ccstate/SKILL.md` | No `detach()` in the signals layer; sub-command composition and signal ownership | Pass | No `detach()`, no new command, no change to the abort chain. |
| `docs/testing.md` | Mock at the boundary; real infrastructure; no fake timers | Pass | The Playwright case drives the real app and uses the existing `mockModelPickerBoundary` route mocks; no internal module mocking and no fake timers. |
| `docs/testing.md` | E2E covers the happy path; the right level for the surface | Pass | The animation is unobservable in jsdom (the signal's own note: every `offsetHeight` is `0` there), so Playwright is the only level that can exercise it; it sits beside the existing picker geometry and segment-fill cases. |
| `docs/testing/anti-patterns.md` | AP-1 test behavior, not that something was called | Finding (P1) | The single assertion is satisfied by "the fade never ran", so it does not prove the behavior it names. |
| `docs/testing/anti-patterns.md` | AP-7 do not test implementation details | Pass | Computed opacity is the user-visible fact under test, and the measurement row is `aria-hidden` by construction, so no role-based locator exists; this matches the sibling `segmentFill` and geometry helpers in the same file. |
| `docs/testing/patterns.md` | Standard structure and reuse of existing helpers | Pass | Reuses `mockModelPickerBoundary`, `appUrl`, and the established `getByRole` navigation into the picker. |

### Findings

None.

- `e2e/playwright/tests/chat.spec.ts:1316` — the regression case cannot distinguish "the measurement row stayed hidden" from "no fade ran at all". `opacity === "0"` is also what you read when the `ResizeObserver` never fired, when the list height did not change, when `prefers-reduced-motion` short-circuits before the fade loop, or when the two `requestAnimationFrame` waits overrun the 120ms fade on a loaded CI machine. The test would then keep passing after a full revert of the fix (`REVIEW.md` regression-test requirement, `docs/testing/anti-patterns.md` AP-1).
- `e2e/playwright/tests/chat.spec.ts:1316` — the diff narrows the fade's target set, and nothing covers the half that must survive: real rows still fading in. Removing or over-broadening the skip selector would leave every test green (`REVIEW.md` Feature Regression Guard).

  Both are fixed by one addition: in the same `evaluate`, also read a non-measurement row's computed opacity and assert it is below `1`. That assertion doubles as the canary proving the sample landed inside the fade window, so the hidden-row assertion can no longer pass vacuously. It adds no new timing exposure — the sample point is shared — and converts a silent pass into a loud failure.

### Testing

- Coverage: Insufficient — the `fix:` commit has a regression case, but as written it can pass without the defect ever being exercised, and the retained fade behavior is uncovered. See the P1 entries.
- Conventions: Compliant — real app, boundary-level route mocks, no internal mocking, no fake timers, no assertion on mock calls; the click and the read share one `evaluate`, which is the right call given the 120ms window.
- Local verification reported in the PR body (`e2e` `check-types`, pinned `prettier@3.8.1`) matches what the diff needs; the remaining lint, type, and `cli-e2e` checks belong to the pipeline.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 51682b59e68c31cc1f404a2f028bb052fe50ff24
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

The category-switch fade in `model-picker-panel-height.ts` walked every direct child of the select list and skipped only the sticky header, so it also animated the `aria-hidden`, `opacity-0` measurement row the picker keeps in the list while a media category is open. A keyframe outranks the class that hides that row, so it was driven to full opacity for the fade's 120ms; on iOS Safari the resulting compositing layer escapes the row's 1px `overflow` clip and the header's stacking order, printing the selected model name over the category title. The fix skips `aria-hidden` rows alongside the header.

Re-review at `51682b5`. The production change is unchanged from `8c1c5d2`; the two P1s from the previous review are resolved: the regression case now reads a real row's opacity from the same sample and asserts it is mid-fade, which both covers the fade that must survive the narrowed target set and proves the sample landed inside the fade window instead of after it.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic: no wrong assumptions or unhandled paths | Pass | `model-picker-panel-height.ts:111`: the only direct children of `[data-slot="select-list"]` carrying `aria-hidden` are the two measurement `SelectItem`s (`model-provider-picker.tsx:700`, `:1235`); every other `aria-hidden` in that file is an icon nested inside a row, so the guard cannot swallow a content row. |
| `REVIEW.md` | Type safety: no `any`, no broad casts | Pass | Only string constants, `instanceof` narrowing, and an inferred object literal; no casts. |
| `REVIEW.md` | Security | N/A | No auth, input handling, network, or persistence surface is touched. |
| `REVIEW.md` | `fix:` commits must carry a regression test | Pass | `chat.spec.ts:1318` asserts the measurement row reads `opacity: 0` during a real category switch, and the paired mid-fade assertion at `:1373` closes the vacuous-pass path that made the earlier version unable to observe the defect. |
| `REVIEW.md` | Feature Regression Guard: cover the existing flow most likely to regress | Pass | `chat.spec.ts:1373`: the diff narrows the fade's target set, and the surviving behavior — the swapped-in rows still fade — is now asserted, so over-broadening the skip selector or dropping the fade fails the suite. |
| `REVIEW.md` | Fallback discipline / `Fallbacks` section in the summary | N/A | The diff introduces no `??`/`||` default, compatibility branch, or cross-version tolerance; it removes an unintended animation target. No `Fallbacks` section required. |
| `REVIEW.md` | New user-facing features gated by a `FeatureSwitchKey` | N/A | No new feature; this restores the intended behavior of a shipped picker that already sits behind `imageModelSelection`. |
| `REVIEW.md` | DB/JSONB review gate, deployment compatibility | N/A | No schema, contract, persisted shape, or cross-service boundary is touched — the change is client-only and takes effect on reload. |
| `docs/bad-smell.md` | Zero `any`, zero lint/type suppressions, no dynamic imports | Pass | Neither file adds a suppression comment, `any`, or `import()`. |
| `docs/bad-smell.md` | Fail fast instead of silent fallback | Pass | The signal keeps its existing `throw` for a missing list; the test throws on a missing segment, list, measurement row, or row group rather than defaulting. |
| `docs/bad-smell.md` | Comments explain WHY, not WHAT | Pass | Both new comments state the cascade reason (a keyframe outranks the hiding class) and the sampling reason, not the mechanics. |
| `docs/effect.md` | DOM lifecycle work belongs in `onRef`, bound to the element's `AbortSignal` | Pass | The change stays inside the existing `onRef` command's `ResizeObserver` callback; owner, observer, and teardown are unchanged. |
| `docs/effect.md` | Render stays pure; no Store writes or resource creation during render | Pass | No view file is touched. |
| `docs/cache.md` | RC-8 symmetric resource lifecycles | Pass | No new observer, listener, or timer; the diff only reduces the set of elements passed to `Element.animate`, whose animations are `fill: none` and not retained. |
| `docs/cache.md` | RC-3 no unbounded module-lifetime caches | Pass | The added module constant is a static selector string. |
| `.claude/skills/ccstate/SKILL.md` | No `detach()` in the signals layer; sub-command composition and signal ownership | Pass | No `detach()`, no new command, no change to the abort chain. |
| `docs/testing.md` | Mock at the boundary; real infrastructure; no fake timers | Pass | The case drives the real app through `mockModelPickerBoundary` route mocks; no internal module mocking, no fake timers, no assertion on mock calls. |
| `docs/testing.md` | E2E covers the happy path at the level that can observe it | Pass | The animation is unobservable in jsdom (the signal's own note: every `offsetHeight` is `0` there), so Playwright is the only level that can exercise it; it sits beside the existing picker geometry and segment-fill cases. |
| `docs/testing/anti-patterns.md` | AP-1 test behavior, not that something was called | Pass | Both assertions are satisfiable only by the real rendered result at a sample point proven to be inside the fade. |
| `docs/testing/anti-patterns.md` | AP-7 do not test implementation details | Pass | Computed opacity is the user-visible fact under test, and the measurement row is `aria-hidden` by construction, so no role-based locator exists; this matches the sibling `segmentFill` and geometry helpers in the same file. |
| `docs/testing/patterns.md` | Standard structure and reuse of existing helpers | Pass | Reuses `mockModelPickerBoundary`, `appUrl`, and the established `getByRole` navigation into the picker. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None. Both P1s from the review of `8c1c5d2` are resolved by `51682b5`.

### Testing

- Coverage: Adequate — the `fix:` commit carries a regression case that fails on the reverted fix and also fails if the fade it narrows is lost entirely.
- Conventions: Compliant — real app, boundary-level route mocks, no internal mocking, no fake timers, no mock-call assertions.
- Sampling: two `requestAnimationFrame` waits is the minimum that works, since a `ResizeObserver` callback runs after the same frame's `requestAnimationFrame` callbacks; the read therefore lands roughly 16-33ms into the 120ms fade. A machine slow enough to stretch two frames past 120ms now fails loudly on the mid-fade assertion instead of passing vacuously, which is the intended trade-off and is documented at the call site.
- Local verification reported by the author (`e2e` `check-types`, `commitlint`, glob-scoped lefthook jobs) matches the changed scope; lint, type, and `cli-e2e` coverage belongs to the pipeline.

PR #29050, opened 2026-08-24
perf(api): resolve v2 attachments by exact key

### LINKED ISSUE #29033: perf(api): resolve v2 attachments with exact-key head
## Parent objective

This is one API-only delivery slice of #24203, which targets successful all-scenario
`api_to_spawn` p90 at or below 1,000 ms without moving required work past real guest-process spawn
or weakening authorization, persistence, queue, cancellation, or recovery behavior.

## Production evidence

A rolling 12-hour read-only production snapshot queried around `2026-08-24T10:52Z` used API
revision `8f2477b51ccce1eb70adb60b6522e1035a4cf713`, Runner `v0.171.6`, and
`run_preparation_retry_count = 0`. It contained 690 successful complete `api_to_spawn` joins at
approximately 883/1,631/2,069/3,073 ms p50/p90/p95/p99, with 439/690 (63.6%) at or below one
second.

Attachment metadata resolution affected 74/690 starts. The 48 one-attachment observations
measured approximately 183 ms p90. #27976 / PR #27985 removed serial work across multiple
attachments, but bounded concurrency cannot improve a one-item request. Per-run ideal removal of
the complete attachment interval would move all-scenario p90 from approximately 1,631 ms to 1,599
ms and the one-second count from 439 to 450. Those 32 ms and 1.59 percentage points are upper
bounds, not expected delivered gains.

## Problem

`resolveIncomingAttachFileMetadata$` supplies only `userId` and `fileId` to the shared artifact
resolver even though the incoming file part also contains `filenameSnapshot`.

The generic resolver first lists the unqualified V2 object prefix, then performs `HeadObject` on
returned candidates to verify `artifact-id` and encoded `user-id` metadata. It subsequently
retains the V1 prefix fallback. Normal V2 uploads have a deterministic key derived from the
artifact ID and the sanitized filename extension, so the existing filename snapshot can identify
a candidate object without the leading prefix-list round trip.

The filename is untrusted input and must be used only as a lookup hint. Object metadata and current
object existence remain authoritative.

## Delivery scope

- Let the incoming attachment path provide its filename snapshot as an exact V2 lookup hint.
- Derive the normal unqualified V2 candidate key with the existing `buildArtifactKeyV2` contract
  and issue `HeadObject` directly.
- Accept the candidate only when it currently exists and its object metadata contains the exact
  requested artifact ID and encoded current user ID. Preserve filename, content type, size,
  last-modified, and public-brand derivation semantics.
- On a missing candidate, metadata mismatch, or incomplete HEAD result, run the existing complete
  unqualified V2 prefix-list resolver and V1 fallback unchanged so legacy objects, stale filename
  snapshots, hash-prefix collisions, and mixed callers remain compatible.
- Leave callers without a filename hint on the current resolver path.
- Add no cache, database state, migration, API field, feature switch, retry, or new object-key
  format.

Explicit V2 variants use a different hash prefix and are not discovered by the current ID-only
generic resolver. This issue does not change that behavior or make variant objects attachment
resolvable.

## Integration coverage

Extend the existing chat-events route/database integration suite at the external S3 boundary:

- a normal V2 one-attachment send resolves through one exact-key `HeadObject` without a leading
  `ListObjects` call and preserves the persisted/returned attachment metadata;
- a missing or incomplete exact candidate falls back to the existing unqualified V2 prefix-list
  path and succeeds for a compatible object;
- a candidate with mismatched ownership metadata is never accepted, while the existing missing
  attachment response and no-run/no-queue behavior remain intact when fallback finds no valid
  object;
- existing V1, multiple-attachment ordering/concurrency, abort, and natural external-error
  coverage remains green.

Mock only the external S3 boundary. Exercise the real Hono route, ccstate graph, and database; do
not add internal service mocks or timing assertions.

## Acceptance criteria

- The common normal unqualified V2 attachment path removes the prefix-list request and retains one
  current authoritative object HEAD.
- Artifact ID and user ownership metadata remain mandatory for every exact-key hit; untrusted
  filename, content type, or key derivation never grants access.
- Missing, stale, invalid-metadata, incomplete-HEAD, and V1 cases preserve the current generic
  fallback and user-visible behavior.
- Explicit variant-object behavior remains unchanged and outside the attachment fast path.
- No API contract, database, queue payload, Runner protocol, object-key format, or persisted
  attachment shape changes.
- Applicable API formatting, lint, type checking, focused route integration tests, Knip, staged
  checks, and `git diff --check` pass.
- Post-deploy validation reports attachment timing, all-scenario `api_to_spawn`, the one-second
  fraction, and error/tail behavior for a fixed API/Runner pair. The change is retained as a parent
  optimization only if it demonstrates at least 25 ms all-scenario p90 movement, two percentage
  points at the one-second boundary, or a material p95/p99 improvement without regression.

## Risks and non-goals

- A stale filename or V1 object may pay one unsuccessful HEAD before the current fallback. This
  must not change the final resolution result.
- The observed 32 ms parent movement removes the entire attachment interval; this PR removes only
  the leading list round trip, so actual benefit can be smaller.
- This issue does not optimize explicit V2 variants, atomic-launch queue insertion, connector
  projection release-time fallback, multi-attachment scheduling, upload preparation/completion,
  artifact deletion, or Runner startup work.

## Relationships

- Parent: #24203
- Completed multi-attachment predecessor: #27976 / PR #27985
- Independent Runner burst work awaiting production validation: #28855 / PR #28882

seven332:
## Summary

- resolve chat attachment V2 objects with one exact-key `HEAD` when the persisted filename snapshot still matches
- preserve artifact ownership and identity validation, falling back to the existing V2 listing and V1 lookup for stale, missing, or incomplete exact results
- cover exact-hit batching, stale filenames, incomplete `HEAD` metadata, and ownership mismatches at the chat event route

Explicit V2 variant keys, upload storage, external contracts, and telemetry are unchanged.

## Validation

- `pnpm format`
- `pnpm -F api lint`
- `pnpm -F api check-types`
- `pnpm -F api exec vitest run src/signals/routes/__tests__/chat-events.bdd.test.ts --testNamePattern attachment` (22 passed)
- `pnpm knip --workspace api`
- pre-commit hooks: repository Knip, repository TypeScript checks, formatting, file-size check, and commitlint

Closes #29033

Parent: #24203

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts
@@ -10219,9 +10219,10 @@ describe("CHAT-02: generation templates and attachments", () => {
     );
     const firstWaveStarted = createDeferredPromise<void>(context.signal);
     const releaseFirstWave = createDeferredPromise<void>(context.signal);
-    let startedLists = 0;
-    let activeLists = 0;
-    let peakActiveLists = 0;
+    let matchingListRequests = 0;
+    let startedHeads = 0;
+    let activeHeads = 0;
+    let peakActiveHeads = 0;
     context.mocks.s3.send.mockImplementation(
       async (command: unknown): Promise<unknown> => {
         if (command instanceof ListObjectsV2Command) {
@@ -10233,14 +10234,7 @@ describe("CHAT-02: generation templates and attachments", () => {
           if (!object) {
             return { Contents: [] };
           }
-          startedLists += 1;
-          activeLists += 1;
-          peakActiveLists = Math.max(peakActiveLists, activeLists);
-          if (startedLists === 4) {
-            firstWaveStarted.resolve(undefined);
-          }
-          await releaseFirstWave.promise;
-          activeLists -= 1;
+          matchingListRequests += 1;
           return {
             Contents: [
               {
@@ -10258,6 +10252,14 @@ describe("CHAT-02: generation templates and attachments", () => {
           if (!object) {
             return {};
           }
+          startedHeads += 1;
+          activeHeads += 1;
+          peakActiveHeads = Math.max(peakActiveHeads, activeHeads);
+          if (startedHeads === 4) {
+            firstWaveStarted.resolve(undefined);
+          }
+          await releaseFirstWave.promise;
+          activeHeads -= 1;
           return {
             ContentLength: object.size,
             ContentType: object.contentType,
@@ -10307,9 +10309,9 @@ describe("CHAT-02: generation templates and attachments", () => {
     });
 
     await firstWaveStarted.promise;
-    expect(startedLists).toBe(4);
-    expect(activeLists).toBe(4);
-    expect(peakActiveLists).toBe(4);
+    expect(startedHeads).toBe(4);
+    expect(activeHeads).toBe(4);
+    expect(peakActiveHeads).toBe(4);
     releaseFirstWave.resolve(undefined);
 
     const sent = await send;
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #29050

**Reviewed head:** `bdff51ad0ecb4413f0375bc94974fc846cefeb2b`

### Summary

The exact V2 attachment path preserves identity and ownership validation, keeps the established compatibility fallbacks, and removes the prefix listing from the common chat-send path without changing external contracts.

### Findings

No findings.

### Testing

Route integration tests cover exact hits, bounded concurrency and ordering, stale filename hints, incomplete HEAD metadata, and ownership mismatches. Formatting, API lint, API type checks, API Knip, the 22 attachment tests, and repository pre-commit checks passed. The full Vitest suite remains delegated to CI.

### Verdict

LGTM

PR #29068, opened 2026-08-24
Route website images through the VM0 batch CLI

bingjiez666:
## Summary

- centralize the latest Website image workflow in @okouai/core so API and CLI prompts cannot drift
- direct agents to use the existing image-batch start/wait commands, author HTML while the batch runs, keep state outside the site, and embed returned URLs plus validated data-generation-size values
- make image-batch help artifact-generic instead of presentation-only
- preserve the pre-cutover resolve-images.mjs path unchanged

The Template-Website-v2 QA_READY, review-atlas, publish-once, and published-verification contract is already present in all 11 template Skills, so this PR does not duplicate it.

## Validation

- 2 image-batch tests passed, including max-3 concurrency and one transient retry
- 14 CLI Website prompt tests passed
- 22 API generation-template prompt tests passed
- targeted chat-events BDD prompt test passed
- @okouai/core and @okouai/cli type checks passed
- full API check-types passed
- targeted oxlint and ESLint passed
- Prettier check passed


bingjiez666:
Closing as superseded by `main`.

Everything this PR set out to do already landed in a later, better form:

- `turbo/packages/core/src/website-generation-instructions.ts` exists on `main` and exports `WEBSITE_IMAGE_BATCH_INSTRUCTION`.
- Both consumers already import that single constant, so the CLI/API drift this PR targeted is closed: `turbo/apps/cli/src/commands/generate/website.ts` (L17, L180) and `turbo/apps/api/src/lib/generation-template-prompt.ts` (L30, L363).
- The `image-batch` help is already artifact-generic ("Start a detached image batch" / "Wait for an image batch and print its results"), not presentation-specific.

The `main` wording is also strictly better than this branch's `WEBSITE_IMAGE_BATCH_RULE`: it does not hard-code Seedream 4 into the instruction, and it states the "with no manifest skip this workflow" case explicitly.

No action carried forward.

Issue #28275, opened 2026-08-20
perf(api): avoid full catalog hydration during compatibility transitions


seven332:
## Problem and current evidence

The scoped connector runtime projection program removed complete catalog hydration from the normal
serverless process-cold path. Its seven delivered children now cover miss attribution, exact-identity
projection persistence and reads, row-read and validation attribution, request-side validation
reduction, the V2-only representation, and retired rollout compatibility.

The ready path is healthy in current production observations:

- exact projection hits measure approximately 14-16 ms p90;
- exact projection misses measure approximately 69-71 ms p90; and
- the migration 0963 relation-availability probe and legacy JSONB representation have been removed.

A distinct request-visible release-time tail remains. The runtime projection identity read requires
an exact current compatibility evaluation for the active catalog and executable capability. When
that row or its current validator authority is unavailable, it returns
`compatibility_not_ready`; `loadConnectorRuntimeSelection()` then reads, decompresses, verifies,
decodes, and materializes the complete accepted catalog through the authoritative full fallback.

Read-only production observations show repeated release-time clusters:

| UTC day | `compatibility_not_ready` full fallbacks | Observed tail |
| --- | ---: | ---: |
| 2026-08-25 | 20 | approximately 3.13 s max |
| 2026-08-26 | 60 | approximately 3.02 s p99/max |
| 2026-08-27 | 79 | approximately 3.26 s p99, 3.50 s max |
| 2026-08-28 partial | 6 | approximately 3.34 s max |

This is no longer a steady-state p90 owner. It is a repeatable release-time p99/max regression that
exposes the first affected requests to multi-second work even though the exact persisted projection
set is normally intact.

The validator authority is deliberately bound to the API package release. PR #23348 rejected a
manually maintained validator revision because a missed bump could silently trust stale validation.
The remaining problem therefore cannot be solved by simply reusing an older compatibility or
projection authority.

## Decision

Keep this issue as a delivery parent, but reframe its remaining objective around eliminating or
materially reducing request-visible complete catalog hydration during exact compatibility-authority
transitions.

The previous process-empty projection objective is delivered. The pre-traffic warming child #29091
was explicitly closed as not planned; future work should optimize the direct runtime code path.
That decision does not authorize weakening exact authority, treating an old compatibility result as
current, or making ordinary request traffic persist new trust state.

This parent is not directly implementable. Planning must compare direct-path designs and then
materialize one independently reviewable child only if it can preserve the complete authority and
fallback contract.

[...]

PR #29098, opened 2026-08-24
perf(api): warm connector projections before traffic

### LINKED ISSUE #29091: perf(api): warm connector runtime projections before production traffic
## Parent objective

This is one independently reviewable delivery slice under #28275 and the broader `api_to_spawn`
objective in #24203. It addresses the release-time projection fallback that remains after the
normal scoped connector projection path became ready and fast.

## Problem and evidence

A read-only production snapshot for API revision
`8f2477b51ccce1eb70adb60b6522e1035a4cf713` found seven runtime catalog loads while exact
compatibility readiness was unavailable. All seven selected `full_fallback` with
`compatibility_not_ready` and measured approximately 3,047 ms. Once readiness converged, 978
projection misses measured approximately 70 ms p90 and 278 projection hits measured approximately
14 ms p90.

The fallback rows were clustered at the start of the API revision. This is a rare release-time
tail rather than a steady-state p90 owner, but it exposes user requests to multi-second complete
catalog hydration and blocks safe retirement of temporary compatibility paths.

The runtime already has the required authority operation: the connector-catalog sync endpoint
synchronizes the accepted catalog, reconciles exact executable compatibility, and reconciles the
runtime projection before returning diagnostics. Preview CI invokes that endpoint through the
unique deployment URL before creating an alias. Production currently deploys directly with
`vercel deploy --prod`, so the deployment can receive traffic before the same reconciliation has
completed. Production rollback also promotes a retained deployment without first reconciling its
own executable capability identity.

## Decision

Fix the deployment ordering gap rather than adding another catalog representation or making user
traffic responsible for readiness. A production API revision must be prepared at its unique
deployment URL, reconciled against the production database with its own executable capability
identity, and proven projection-ready before its production traffic promotion. The currently
serving revision remains available if preparation fails.

The complete accepted snapshot remains the fail-closed runtime fallback. This issue does not turn
projection readiness into authorization when exact identity or validation authority is stale.

- Extend only the cron sync response with a bounded discriminated runtime-projection readiness
  result. Do not change the authenticated connector diagnostics response.
- Prove readiness by reusing the exact identity read, matching it to the diagnostics catalog and
  executable capability identity, and comparing child-row count through the opaque projection-set
  generation ID. Do not scan every payload in this release gate.
- Move the existing preview predicate and retry behavior into one tested repository-owned Bash
  helper used by preview, normal production, and rollback. Its logs contain only fixed states,
  booleans, counts, and bounded reasons.
- Change normal production ordering to migration -> staged production deploy with `skip-domain` ->
  exact staged-URL reconcile -> promotion of the same artifact -> production-domain verification.
- Reconcile rollback targets before promotion. During the first rollout only, a target descending
  from runtime projection V2 baseline `f40d5ea90b2fc4649cd336d083ce9d97573fb9ed` may use the
  missing-readiness-field compatibility mode. All seven currently retained rollback releases
  satisfy this baseline and already await runtime projection reconciliation. An explicit non-ready
  response is never accepted. Remove this compatibility mode after all retained targets contain
  this issue's response field.

This is one cohesive PR: the API readiness signal is not useful without the deployment gate, and
the gate cannot safely become strict until the exact artifact can emit that signal.

## Required behavior

- Normal production release stages the API artifact without moving the production domain, runs the
  existing connector-catalog reconciliation against that exact deployment, verifies current
  catalog, compatibility, and runtime-projection readiness, then promotes the same artifact.
- Production rollback reconciles the selected retained API deployment before promotion so the
  rollback executable capability identity is ready as well.
- The new readiness field exposes no catalog versions, digests, connector identities, capability
  digests, users, tenants, or payload contents. Workflow logs must also exclude dynamic identities
  already present elsewhere in the response.
- Preparation is idempotent and safe when retries or overlapping deployment callers reach the same
  active catalog generation.
- Preparation failure stops promotion without disabling or mutating the currently serving API
  revision. Existing deployment serialization remains authoritative.
- Database migrations continue to complete before the new API artifact is reconciled or promoted.
- Old/new API overlap and supported rollback retain exact source, schema, catalog version, catalog
  digest, capability digest, generation, completeness, and validator-authority checks.

## Acceptance criteria

- [ ] A normal API release does not move production traffic until its unique deployment reports a
      current catalog, non-stale compatibility evaluation, and ready runtime projection.
- [ ] The promoted artifact is byte-for-byte the staged and reconciled artifact; no rebuild occurs
      between reconciliation and promotion.
- [ ] Missing schema, missing/incomplete projection sets, unsupported versions, identity changes,
      and missing or stale compatibility/validator authority cannot satisfy the promotion gate.
- [ ] Reconciliation retry is bounded, and exhaustion leaves the previous production deployment
      serving while the release deployment is reported failed.
- [ ] Rollback performs the same target-deployment reconciliation before promotion. Only a verified
      V2-or-newer retained target may use the temporary missing-field compatibility branch.
- [ ] Preview, production release, and rollback invoke one shared readiness predicate and cannot
      silently drift.
- [ ] Real-database API integration coverage proves ready, repaired missing/incomplete state, stale
      authority repair, and bounded identity-change behavior.
- [ ] Workflow-level coverage proves migration -> staged deploy -> reconcile -> promote ordering,
      exact deployment URL use, no rebuild, bounded failure behavior, legacy rollback eligibility,
      and Bash-compatible shell execution.
- [ ] Gate tests prove strict and legacy response handling, retry behavior, invalid response
      rejection, optional deployment protection bypass, and safe logs.
- [ ] Post-deployment validation observes no request-visible `compatibility_not_ready` fallback in
      a normal release and no catalog availability, rollback, or projection-tail regression.

## Alternatives considered

- **Composite GitHub action:** provides named YAML inputs but still embeds the same shell behavior,
  adds metadata and test indirection, and has no current structured-output requirement.
- **Trust endpoint HTTP 200 without readiness:** cannot prove schema availability or a complete
  projection set and therefore does not meet the fail-closed promotion gate.
- **Keep direct production deployment and rely on cron:** retains the measured multi-second first-
  request tail on every executable identity change.
- **Reconcile through the current production domain before deployment:** evaluates the outgoing
  executable capability identity, not the staged artifact that will serve next.
- **Reconcile lazily from the first user request:** preserves the exact user-visible delay and makes
  request traffic a deployment authority.
- **Remove the full fallback:** weakens fail-closed availability and rollback behavior when derived
  projection state is genuinely unavailable or invalid.
- **Scan and repair every projection payload during each release:** addresses a separate integrity-
  recovery concern. Current request-side validation already fails closed; a complete scan is not
  required to fix the observed `compatibility_not_ready` gap.

## Risks and mitigations

- The release can take longer or fail when reconciliation cannot reach a ready state. Keep retries
  bounded and leave the old production deployment untouched until promotion.
- A staged deployment can observe a catalog activation racing reconciliation. Match the readiness
  identity to the diagnostics snapshot and count by immutable projection generation; retry rather
  than accept mixed identities.
- A retained rollback deployment may predate the readiness field. Gate legacy response acceptance
  on the V2 target-code baseline and remove it after the retained rollback window advances.
- A row count does not prove every payload digest. Keep request-side validation and the complete
  snapshot fallback unchanged; intact-count corruption repair remains out of scope.
- Repeated workflow implementations can drift. Use one shared repository-owned helper and test each
  workflow's ordering and explicit mode.

## Scope boundaries and relationships

This issue does not remove the full catalog snapshot, change the projection representation, add a
database migration, delete rollout compatibility columns, repair intact-count corrupted rows,
change the authenticated connector diagnostics response, or optimize steady-state projection
hits. Cleanup of the legacy representation remains tracked by #28700 after its rollout and
rollback gates pass.

- Parent: #28275
- Broader performance objective: #24203
- Projection delivery: #28478
- Request-side validation optimization: #28694
- Deferred representation cleanup: #28700

seven332:
## Summary

- expose a bounded runtime-projection readiness result from the connector catalog cron response
- share one strict readiness/retry/log-redaction gate across preview, production release, and rollback
- stage production API artifacts without moving traffic, reconcile the exact deployment URL, then promote the same artifact without rebuilding
- reconcile rollback targets before promotion, with a temporary missing-field compatibility path restricted to the runtime projection V2 baseline

## Behavior and rollout

The gate now requires a current catalog, non-stale executable compatibility, and a complete exact-generation runtime projection. A catalog identity race or incomplete set returns a bounded non-ready result and consumes another retry instead of being promoted.

Normal release ordering is now migration -> staged Vercel production deployment -> exact-URL reconciliation -> promotion -> production-domain verification. Gate failure leaves the existing production deployment serving.

Retained rollback deployments created before this response field may omit it only when the target descends from runtime projection V2 commit `f40d5ea90b2fc4649cd336d083ce9d97573fb9ed`; explicit non-ready responses still fail. The workflow comment records when to remove this compatibility branch.

## Explicit exclusions

- no database migration or new persisted representation
- no change to authenticated connector catalog diagnostics
- no removal or weakening of the complete accepted-catalog fallback
- no full scan/repair of intact-count malformed or digest-mismatched projection payloads

## Validation

- `cd turbo && pnpm format`
- `cd turbo && pnpm -F @okouai/api-contracts lint`
- `cd turbo && pnpm -F api lint`
- `cd turbo && pnpm -F api check-types`
- `cd turbo && pnpm -F @okouai/api-contracts check-types`
- `cd turbo && pnpm -F api exec vitest run src/signals/routes/__tests__/cron-connector-catalog.test.ts src/__tests__/release-please-config.test.ts` (118 tests)
- `cd turbo && pnpm knip --workspace api`
- `cd turbo && pnpm knip --workspace @okouai/api-contracts`
- `cd turbo && pnpm -F api build`
- `bash .github/scripts/tests/reconcile-api-connector-catalog-test.sh`
- pre-commit full Turbo Knip and `pnpm check-types`

The local image does not provide Ruby or ShellCheck, so the existing rollback workflow parser and workflow shell-compatibility test could not complete locally; CI remains responsible for those checks.

Closes #29091

Parent context: #28275

### DIFF
--- a/.github/scripts/reconcile-api-connector-catalog.sh
+++ b/.github/scripts/reconcile-api-connector-catalog.sh
@@ -1,0 +1,117 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+if [ "$#" -ne 2 ]; then
+  echo "usage: reconcile-api-connector-catalog.sh DEPLOYMENT_URL strict|allow-legacy" >&2
+  exit 2
+fi
+
+deployment_url=${1%/}
+response_mode=$2
+
+if [ -z "$deployment_url" ]; then
+  echo "deployment URL is required" >&2
+  exit 2
+fi
+if [ "$response_mode" != "strict" ] && [ "$response_mode" != "allow-legacy" ]; then
+  echo "response mode must be strict or allow-legacy" >&2
+  exit 2
+fi
+: "${CRON_SECRET:?CRON_SECRET is required}"
+
+headers=(-H "Authorization: Bearer ${CRON_SECRET}")
+if [ -n "${VERCEL_AUTOMATION_BYPASS_SECRET:-}" ]; then
+  headers+=(-H "x-vercel-protection-bypass: ${VERCEL_AUTOMATION_BYPASS_SECRET}")
+fi
+
+for attempt in 1 2 3; do
+  response=""
+  if ! response=$(curl \
+    --fail-with-body \
+    --show-error \
+    --silent \
+    --max-time 120 \
+    "${headers[@]}" \
+    "${deployment_url}/api/cron/sync-connector-catalog"); then
+    echo "::warning::Connector catalog reconcile request failed (attempt ${attempt}/3)"
+  elif summary=$(jq -ce '
+    if type != "object" then
+      error("response must be an object")
+    else
+      {
+        outcome: (
+          if .outcome == "accepted" or .outcome == "unchanged" or .outcome == "rejected"
+          then .outcome
+          else "invalid-response"
+          end
+        ),
+        catalogCurrent: (.state == "current"),
+        hasActiveCatalog: (.active != null),
+        compatibilityCurrent: (
+          (.filtering | type) == "object" and .filtering.stale == false
+        ),
+        filteredAuthMethodCount: (
+          if (.filtering.filteredAuthMethods | type) == "array"
+          then (.filtering.filteredAuthMethods | length)
+          else null
+          end
+        ),
+        runtimeProjection: (
+          if has("runtimeProjection") | not then
+            "legacy-response"
+          elif (.runtimeProjection | type) != "object" then
+            "invalid-response"
+          elif .runtimeProjection.state == "ready" then
+            "ready"
+          elif
+            .runtimeProjection.state == "not-ready" and
+            (
+              .runtimeProjection.reason == "schema-unavailable" or
+              .runtimeProjection.reason == "projection-not-ready" or
+              .runtimeProjection.reason == "unsupported" or
+              .runtimeProjection.reason == "compatibility-not-ready" or
+              .runtimeProjection.reason == "invalid-compatibility" or
+              .runtimeProjection.reason == "incomplete" or
+              .runtimeProjection.reason == "identity-changed"
+            )
+          then .runtimeProjection.reason
+          else "invalid-response"
+          end
+        )
+      }
+    end
+  ' <<<"$response"); then
+    echo "Connector catalog reconcile: ${summary}"
+
+    if jq -e --arg response_mode "$response_mode" '
+      type == "object" and
+      .state == "current" and
+      .active != null and
+      (.filtering | type) == "object" and
+      .filtering.stale == false and
+      (
+        (
+          has("runtimeProjection") and
+          (.runtimeProjection | type) == "object" and
+          .runtimeProjection.state == "ready"
+        ) or
+        (
+          (has("runtimeProjection") | not) and
+          $response_mode == "allow-legacy"
+        )
+      )
+    ' <<<"$response" >/dev/null; then
+      exit 0
+    fi
+    echo "::warning::Connector catalog is not ready (attempt ${attempt}/3)"
+  else
+    echo "::warning::Connector catalog reconcile returned invalid JSON (attempt ${attempt}/3)"
+  fi
+
+  if [ "$attempt" -lt 3 ]; then
+    sleep 2
+  fi
+done
+
+echo "::error::Connector catalog reconcile did not produce a current ready projection"
+exit 1
--- a/.github/scripts/tests/reconcile-api-connector-catalog-test.sh
+++ b/.github/scripts/tests/reconcile-api-connector-catalog-test.sh
@@ -1,0 +1,118 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+repo_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../.." && pwd)"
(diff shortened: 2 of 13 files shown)


seven332:
## Code Review: PR #29098

**Reviewed head:** `3b02a77e449ef389ee9b514291d503ba867b22f9`

### Summary

Adds bounded cron-only projection readiness, one shared fail-closed deployment gate, and stage/reconcile/promote ordering for preview, release, and rollback.

### Findings

No findings.

### Testing

Route integration: 118 tests passed; API and API contracts format, lint, types, and Knip passed; API build passed; shared shell gate test passed; pre-commit full Knip and type checks passed. The local Ruby rollback parser and ShellCheck wrapper were not verified because those tools are absent locally; CI covers them.

### Verdict

LGTM

seven332:
Closing without merge. We are dropping the deployment-time warming approach and will focus future work on optimizing the code path directly.

PR #29037, opened 2026-08-24
fix(composer): keep the feedback note chrome outside the editable flow

bingjiez666:
## Summary

The structural fix for #27787: the feedback note node view leaves its wrapper elements (`dom`, `noteDom`, the placeholder) editable, so WebKit's editing machinery treats them as part of the editable region. Two observed consequences:

1. During an IME composition, WebKit's delete-and-prune editing primitive can climb through the "emptied" anonymous editable wrappers and remove them, leaving a bare `<br>` — the collapse traced in desktop Safari that makes everything typed afterwards silently missing from submissions.
2. With multiple quotes, pressing Backspace mid-composition can land the DOM caret between the wrappers; the gap has no view desc, so ProseMirror resolves the position at the block boundary into the **previous** note's text.

Behind the new `composerNoteEditableIsolation` switch (default off, `bingjie@vm0.ai` only), the node view marks its `dom` `contenteditable="false"` and opts `contentDOM` back in as its own editing host. Nothing around the note content belongs to the editable region anymore: WebKit's ancestor pruning stops at the editing-host boundary, and the caret can no longer land between the wrappers. With the switch off the node view renders exactly as before.

## Relationship to #29015

Standalone and intended to merge first. #29015 (collapse detection, zero-loss reattachment, and collapse telemetry) remains open as the complementary safety net and the field instrument: with both switches on, #29015's collapse log going quiet is the proof this prevention works. #29015 will be rebased over this change after it lands.

## Interaction surface to verify in dogfood

Nested editing hosts change selection physics at the note boundary. Manual checklist:

- quote → compose pinyin in the note until the composer previously collapsed — must no longer collapse
- two quotes → compose pinyin in the second note → Backspace mid-composition — caret must stay in the second note (the newly reported symptom)
- drag-select from the plain paragraph across into a note; select-all; arrow keys into/out of the note; Backspace at note start
- iOS: tap-to-focus the note, software keyboard behavior, long-press selection

## Verification

- `pnpm exec prettier --write <changed files>` — clean
- `apps/platform` `pnpm run lint` — exit 0
- `pnpm --filter @okouai/core run check-types`, `pnpm --filter @okouai/app run check-types` — clean
- `pnpm knip` — exit 0
- `pnpm --filter @okouai/core exec vitest run src/__tests__/feature-switch.test.ts` — 25 passed
[...]

PR #29120, opened 2026-08-25
chore(composer): open the note editable isolation switch to the staff org

bingjiez666:
## Summary

`ComposerNoteEditableIsolation` landed in #29037 gated to a single email hash (`fnv1a("bingjie@vm0.ai")`). This moves it one rollout step up to the whole staff org so the team can dogfood the fix for the feedback-note collapse (#27787).

```diff
-    enabledEmailHashes: ["9fd4ee92"], // fnv1a("bingjie@vm0.ai")
+    enabledOrgIdHashes: STAFF_ORG_ID_HASHES,
```

`enabled: false` is unchanged, so nothing outside the staff org is affected. The org hash **supersedes** the email hash rather than joining it — the staff org already covers the maintainer, so keeping both would leave a dead identity path.

Per `.claude/skills/feature-switch`, `enabled: false` + `enabledOrgIdHashes: STAFF_ORG_ID_HASHES` is the standard staff-only tier, matching `Lab`, `Banking`, `ManagedSocialKit`, and `MorningBrief`.

## What the switch does

With it on, the feedback-item node view marks its `dom` `contenteditable="false"` and opts `contentDOM` back in as its own editing host, so WebKit's editing machinery can neither prune the note wrappers mid-IME-composition nor land the caret between them.

## Test changes

The two `ComposerNoteEditableIsolation` assertions in `should enable Bingjie-only composer switches only for Bingjie` are **deleted** rather than flipped: the switch is no longer Bingjie-only, so those assertions described configuration that no longer exists. Per `docs/fallback.md` §1, an outdated assertion is removed together with the config it covered instead of being kept as a tombstone.

Coverage moves to `should reflect the current staff org rollout matrix`, which asserts it on both sides (staff org `true`, non-staff org `false`) alongside every other org-gated switch.

## Rollout risk worth naming

#29015 (collapse detection, zero-loss reattachment, and collapse telemetry) is **still open**. Until it lands, `createFeedbackItemNodeView` keeps main's `ignoreMutation`, which discards every mutation outside `contentDOM` — including the record for WebKit tearing the wrappers out.

So for anyone in the staff org who hits a collapse anyway, the failure stays silent: no repair, no log, no signal. That is not a regression (it is today's behavior for everyone), but it does mean **absence of reports is not evidence the prevention works**. Verify positively — reproduce the collapse with the switch off, then confirm it stops with it on — rather than inferring success from silence.

## Verification

- `pnpm exec prettier --write <changed files>` — clean
- `pnpm --filter @okouai/core run check-types` — clean
- `pnpm --filter @okouai/core run lint` — exit 0
- `pnpm --filter @okouai/core exec vitest run src/__tests__/feature-switch.test.ts` — 25 passed

Related to #27787. Follows #29037.

### DIFF
--- a/turbo/packages/core/src/__tests__/feature-switch.test.ts
+++ b/turbo/packages/core/src/__tests__/feature-switch.test.ts
@@ -128,6 +128,9 @@ describe("getAllFeatureStates", () => {
     expect(staffOrgStates[FeatureSwitchKey.HomeGrowthEntry]).toBe(true);
     expect(staffOrgStates[FeatureSwitchKey.ManagedSocialKit]).toBe(true);
     expect(staffOrgStates[FeatureSwitchKey.UsagePackPlans]).toBe(true);
+    expect(staffOrgStates[FeatureSwitchKey.ComposerNoteEditableIsolation]).toBe(
+      true,
+    );
 
     const otherOrgStates = getAllFeatureStates({
       orgId: "org_nonexistent",
@@ -162,6 +165,9 @@ describe("getAllFeatureStates", () => {
     expect(otherOrgStates[FeatureSwitchKey.HomeGrowthEntry]).toBe(false);
     expect(otherOrgStates[FeatureSwitchKey.ManagedSocialKit]).toBe(false);
     expect(otherOrgStates[FeatureSwitchKey.UsagePackPlans]).toBe(true);
+    expect(otherOrgStates[FeatureSwitchKey.ComposerNoteEditableIsolation]).toBe(
+      false,
+    );
   });
 
   it("should enable Bingjie-only composer switches only for Bingjie", () => {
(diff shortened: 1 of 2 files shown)


bingjiez666:
Closing: the switch this PR edits no longer exists on `main`.

Timeline:

- 2026-08-24 12:17 — #29037 landed `ComposerNoteEditableIsolation`, gated to a single email hash.
- 2026-08-25 02:05 — this PR was opened to move that gate up to `STAFF_ORG_ID_HASHES`.
- 2026-08-25 03:04 — #29125 (`revert: composer feedback note editable isolation`) reverted the switch entirely, one hour after this PR was opened.

`turbo/packages/core/src/feature-switch.ts` on `main` no longer contains any composer-note switch (the only remaining `Composer*` key is `ComposerImageAnnotation`), so this diff has no target and cannot be merged.

The underlying feedback-note collapse (#27787) remains unfixed, but since the fix was reverted rather than superseded, any new attempt should be argued independently against current `main` in a fresh PR rather than re-landed from this branch.

PR #29127, opened 2026-08-25
perf(python): skip irrelevant websocket client parsing

### LINKED ISSUE #29118: perf(python): skip Responses parsing for irrelevant WebSocket client frames
## Problem

At `3d7cc0dd7e9f3713e5cddf45a419f3878ad27cbc`, the synchronous mitmproxy WebSocket hook parses every client-originated message on every registered flow as an OpenAI Responses client event, even when no model WebSocket consumer is active.

- `crates/runner/mitm-addon/src/mitm_addon.py:1565-1585` schedules retention, checks only for a sandbox `run_id`, then calls `usage.inspect_openai_responses_client_event_json(body)` before checking `model_websocket_usage.is_enabled(flow)`.
- The resulting event is passed to `model_provider_failure.observe_websocket_client_event()`, but `crates/runner/mitm-addon/src/model_provider_failure.py:400-408` immediately returns for ordinary non-model flows because they have no admitted model failure state. The full parse has already happened.
- `crates/runner/mitm-addon/src/model_websocket_usage.py:117-132` similarly consumes client events only when its private Responses correlation state exists and is still useful.
- `crates/runner/mitm-addon/src/usage/openai_responses.py:234-243` first probes the prefix and then always constructs a fresh `JsonSelectiveExtractor` and feeds the complete client body with a 65,536-work-unit allowance.
- `crates/runner/mitm-addon/src/websocket_framing.py:22-24` permits a complete decoded message up to 256 MiB. Retention is scheduled before parsing, but it runs only after the synchronous addon dispatch and therefore does not reduce this parse cost.

Registered non-model WebSockets are a supported production path: request classification stores the sandbox run ID for registered traffic, and the generic retention tests exercise registered ordinary WebSocket flows independently from model-provider activation.

## Evidence

An isolated CPython 3.12.3 benchmark used the exact client scalar fields, consistency paths, and 65,536-unit limit from `inspect_openai_responses_client_event_json()`. Each input was a valid JSON object containing one unselected `payload` string, which forces complete validation while producing no model event:

| Body size | Selective parse time |
| ---: | ---: |
| 1 MiB | 3.9 ms |
| 16 MiB | 60.2 ms |
| 64 MiB | 238.9 ms |
| 128 MiB | 477.9 ms |
| 256 MiB | 961.9 ms |

These are isolated parser measurements, not end-to-end latency claims. The deterministic hook ordering establishes that the same full-body parse is paid synchronously before both model consumers reject an ordinary registered flow.

## Impact

A connector or public-policy WebSocket associated with a sandbox run can send client JSON frames without being an OpenAI Responses model flow. Large valid frames then spend tens to hundreds of milliseconds in irrelevant Python parsing on the shared mitmproxy event-loop thread; a frame at the allowed maximum approaches one second in the isolated parser. During that time unrelated proxy flows cannot make progress through the same loop. Repeated smaller frames also create and run an extractor for data that neither model consumer can use.

The configured work limit prevents unbounded dense-syntax work, but its discarded-string fast path intentionally admits large content strings. That behavior is appropriate for real Responses frames and does not make the unconditional non-model parse useful.

## Suggested direction

Keep generic retention scheduling first, then determine whether at least one client-event consumer is active before converting or parsing the message body. Expose narrow predicates for failure correlation and usage/prewarm correlation so parsing remains enabled for failure-only and usage-enabled Responses flows, but is skipped for non-model flows and for sticky states that no longer consume client lifecycle evidence.

## Acceptance criteria

- A registered non-model client WebSocket message schedules the existing retention callback but does not call `inspect_openai_responses_client_event_json()` or feed a Responses selective extractor.
- Active Responses usage correlation and failure-only correlation still inspect each relevant client frame exactly once and preserve `generate: false` prewarm handling, output timing, failure reporting, and ambiguity behavior.
- Once every active client-event consumer has entered a state where later client evidence is ignored, later client frames do not receive an unnecessary full-body parse.
- Server-frame shared inspection and generic registered/unregistered retention behavior remain unchanged.
- Add deterministic public-hook coverage that records extractor calls or target-body feeds for non-model, usage-enabled, failure-only, and no-longer-needed client frames without wall-clock assertions.
- Run the focused WebSocket retention, prewarm, failure-reporting, source-reporting, and output-timing tests plus the standard mitm-addon format, lint, type, and test checks.

## Audit context

Found during the daily mitm-addon Python performance audit on commit `3d7cc0dd7e9f3713e5cddf45a419f3878ad27cbc`.

seven332:
## Summary

- gate Responses WebSocket client inspection on active failure, prewarm, and output-timing demand
- skip body conversion and parsing for ordinary registered flows, and keep only the bounded event-type probe after correlation becomes ambiguous
- cover non-model, active prewarm, failure-only, and timing-only client-frame paths through the public hook

## Scope

Server-frame shared inspection, WebSocket retention ordering, provider protocols, and persisted or cross-service contracts are unchanged.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_websocket_retention.py tests/test_model_provider_websocket_prewarm.py tests/test_model_provider_failure_reporting.py tests/test_model_provider_websocket_source_reporting.py tests/test_codex_output_timing.py` (`153 passed`)
- `uv run --no-sync python -m pytest tests/` (`6606 passed`)

Closes #29118

### DIFF
--- a/crates/runner/mitm-addon/src/mitm_addon.py
+++ b/crates/runner/mitm-addon/src/mitm_addon.py
@@ -1572,17 +1572,27 @@ def websocket_message(flow: http.HTTPFlow) -> None:
     if not flow_metadata.run_id(flow.metadata):
         return
     if getattr(message, "from_client", False):
-        body = message.content.encode() if isinstance(message.content, str) else message.content
-        event = usage.inspect_openai_responses_client_event_json(body)
-        model_provider_failure.observe_websocket_client_event(
-            flow,
-            request_kind=event.request_kind,
-            is_prewarm=event.is_prewarm,
-        )
-        if not model_websocket_usage.is_enabled(flow):
+        failure_client_enabled = model_provider_failure.should_observe_websocket_client_event(flow)
+        usage_enabled = model_websocket_usage.is_enabled(flow)
+        usage_client_enabled = model_websocket_usage.should_observe_client_event(flow)
+        if not failure_client_enabled and not usage_enabled:
             return
-        codex_output_timing.observe_client_event(flow, event.event_type, message.timestamp)
-        model_websocket_usage.observe_client_event(flow, event)
+        body = message.content.encode() if isinstance(message.content, str) else message.content
+        if not failure_client_enabled and not usage_client_enabled:
+            event = usage.inspect_openai_responses_event_json(body)
+            codex_output_timing.observe_client_event(flow, event.event_type, message.timestamp)
+            return
+        event = usage.inspect_openai_responses_client_event_json(body)
+        if failure_client_enabled:
+            model_provider_failure.observe_websocket_client_event(
+                flow,
+                request_kind=event.request_kind,
+                is_prewarm=event.is_prewarm,
+            )
+        if usage_enabled:
+            codex_output_timing.observe_client_event(flow, event.event_type, message.timestamp)
+        if usage_client_enabled:
+            model_websocket_usage.observe_client_event(flow, event)
         return
     body = message.content.encode() if isinstance(message.content, str) else message.content
     failure_enabled = model_provider_failure.should_observe_websocket_server_event(flow)
--- a/crates/runner/mitm-addon/src/model_provider_failure.py
+++ b/crates/runner/mitm-addon/src/model_provider_failure.py
@@ -415,6 +415,12 @@ def observe_websocket_client_event(
     flow_state.pending_intent = "prewarm" if is_prewarm else "normal"
 
 
+def should_observe_websocket_client_event(flow: http.HTTPFlow) -> bool:
+    """Return whether an admitted failure state still needs client evidence."""
+    flow_state = _websocket_flow_state(flow)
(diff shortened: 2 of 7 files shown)


seven332:
## Code Review: PR #29127

**Reviewed head:** `3fa941b6b7e9b5f4ac541a96523334467d6d9912`

### Summary

The client WebSocket hook now selects no inspection, bounded-prefix timing inspection, or one shared full correlation parse from state-owner demand predicates. The change matches the approved scope, preserves failure-only and fail-open usage behavior, and leaves server inspection and retention ordering unchanged.

### Findings

No findings.

### Testing

Reviewed public-hook coverage for ordinary registered traffic, active `generate: false` prewarm correlation, failure-only correlation, and sticky-ambiguous timing-only operation. Local lock/sync, Ruff, BasedPyright, focused tests (`153 passed`), and the complete mitm-addon suite (`6606 passed`) passed. Remote CI was not used as review evidence and is checked separately.

### Verdict

LGTM

PR #29129, opened 2026-08-25
fix(ably-subscriber): space short-lived token renewals

### LINKED ISSUE #29119: bug(rust): prevent short-lived Ably tokens from causing an immediate renewal loop
## Problem

`ably-subscriber` can enter a tight successful-renewal loop when an exchanged token's remaining lifetime is less than or equal to `TimingConfig::token_renewal_margin`.

`ConnState::compute_renewal_at` subtracts the configured margin with saturation, so every such valid token receives an immediately ready renewal deadline. After `renew_token` successfully sends `AUTH`, it commits the replacement and calculates the same immediate deadline again. The receive loop uses a biased select with renewal before WebSocket frame reads, so repeated short-lived responses can continuously win the select.

## Evidence and impact

- `crates/ably-subscriber/src/types.rs` sets the default renewal margin to 300 seconds.
- `crates/ably-subscriber/src/connection/state.rs` currently schedules positive remaining lifetime minus the margin with saturation.
- `crates/ably-subscriber/src/connection/event_loop.rs` commits every successfully exchanged token and places an already-ready renewal branch before frame reads.
- `crates/ably-subscriber/tests/integration/auth_token.rs::token_renewal` only covers a short initial token followed by a one-hour token.
- Issue #9020 and PR #9083 previously exposed continuous renewals when the test mock kept returning the short token, but that work corrected the test fixture rather than production scheduling.

Repeated short responses can sustain traffic to the token callback, vm0 API, Ably REST, and WebSocket `AUTH` path while queued job, cancellation, or active-input messages remain unread. Successful renewals reset the failure counter, so the failure policy does not terminate this loop.

## Reproduction

1. Return an unexpired token with a 60-second remaining lifetime on every successful exchange while the default margin remains 300 seconds.
2. Complete the initial WebSocket connection and channel attachment.
3. After automatic renewal, read the client `AUTH` and enqueue a channel `MESSAGE`.
4. Observe immediate repeated exchanges and `AUTH` messages while the queued channel message is starved.

## Decision

Use one adaptive deadline rule for every token installation path:

- Expired tokens remain immediately due.
- For a valid token, set the effective margin to the smaller of the configured margin and half of the token's remaining lifetime.
- Schedule renewal after the remaining lifetime minus that effective margin.

Equivalently, tokens with at least twice the configured margin of remaining lifetime keep the existing `expires - margin` deadline. Tokens below that threshold renew halfway through their remaining validity. This avoids both the zero-delay loop at or below the margin and very short renewal intervals just above the margin.

The rule applies consistently to initial connection, established-session renewal, and reconnect token commits. It adds no public option or state-machine distinction.

## Alternatives considered

- A success-only cooldown preserves immediate initial renewal but makes identical tokens behave differently by installation path and leaves reconnect behavior to classify separately.
- A fixed minimum interval either couples successful renewal to the failure retry setting or adds a new public option, and it does not adapt safely across TTLs.
- Reordering or unbiasing the receive select can reduce message starvation but leaves callback, REST, and `AUTH` traffic unbounded.
- Waiting for server-reported expiry eliminates proactive traffic but gives up the existing pre-expiry renewal guarantee.

## Risks and mitigations

- Initial tokens below twice the margin renew later than before. Half-life scheduling still reserves half of their remaining validity for renewal and recovery and keeps all installation paths coherent.
- Millisecond timestamps cannot express a deadline both strictly future and strictly before expiry when only one millisecond remains. The calculation schedules one millisecond in the future for that representation edge.
- Timing tests can become flaky at exact boundaries. The integration scenario must use a replacement TTL materially larger than its negative observation window and synchronize on protocol events.
- A test that only observes fewer calls could accidentally accept disabled renewal. It must also observe a later renewal before the short replacement expires.

## Acceptance criteria

- Every valid token receives a renewal deadline strictly later than calculation time; expired tokens remain immediately due.
- Tokens with remaining lifetime at least twice the configured margin keep the existing `expires - margin` schedule.
- Tokens with less remaining lifetime renew halfway through that remaining validity.
- Repeated successful short-lived responses do not create a tight callback/REST/`AUTH` loop and do not starve WebSocket frame processing.
- A public `subscribe` integration test repeatedly returns short-lived replacement tokens, verifies calls remain bounded during an observation window shorter than the token lifetime, verifies a channel message is delivered during that window, and observes a later renewal before expiry.
- Affected Rust formatting, Clippy, documentation, and crate tests pass.

## Delivery notes

This is one independently reviewable PR in `crates/ably-subscriber`. It changes no API shape, Ably protocol payload, persisted data, runner protocol, failure counter, or receive-loop priority, and requires no migration or coordinated rollout.

seven332:
## Summary

- cap the effective proactive renewal margin at half of a token's remaining lifetime
- keep expired tokens immediately due and preserve the configured margin for tokens with at least twice that remaining lifetime
- cover repeated successful short-lived tokens through the public subscription flow, including bounded callback traffic, message delivery, and later pre-expiry renewal

## Compatibility

- no public type, protocol payload, persistence, or runner contract changes
- no new timing configuration or change to renewal failure handling
- receive-loop event priority remains unchanged

## Validation

- `cargo test -p ably-subscriber`
- repeated the new exact integration test three times
- repository pre-commit Rust formatting, workspace Clippy, rustdoc, and file-size checks

Closes #29119

### DIFF
--- a/crates/ably-subscriber/src/connection/state.rs
+++ b/crates/ably-subscriber/src/connection/state.rs
@@ -62,8 +62,9 @@ impl ConnState {
             return Some(Instant::now());
         }
 
-        let renew_in_ms = (remaining_ms as u128).saturating_sub(margin.as_millis());
-        let renew_in = Duration::from_millis(u64::try_from(renew_in_ms).ok()?);
+        let remaining = Duration::from_millis(u64::try_from(remaining_ms).ok()?);
+        let effective_margin = margin.min(remaining / 2);
+        let renew_in = remaining - effective_margin;
         checked_deadline_after(renew_in)
     }
 
@@ -519,21 +520,6 @@ mod tests {
 
         let renewal_at = ConnState::compute_renewal_at(&token, Duration::from_secs(300))
             .expect("expired tokens should still schedule renewal");
-        assert!(renewal_at <= Instant::now());
-    }
-
-    #[test]
-    fn token_inside_renewal_margin_is_scheduled_immediately() {
-        let token = TokenDetails {
-            token: "tok".to_string(),
-            expires: unix_now_ms() + 60_000,
-            issued: 0,
-            capability: None,
-            client_id: None,
-        };
-
-        let renewal_at = ConnState::compute_renewal_at(&token, Duration::from_secs(300))
-            .expect("tokens inside the renewal margin should schedule renewal");
         assert!(renewal_at <= Instant::now());
     }
 
--- a/crates/ably-subscriber/src/types.rs
+++ b/crates/ably-subscriber/src/types.rs
@@ -242,6 +242,10 @@ pub struct TimingConfig {
 
     // -- Token renewal -------------------------------------------------------
     /// How early before token expiry to start proactive renewal.
+    ///
+    /// When this exceeds half of a token's remaining lifetime, the effective
+    /// margin is capped at that half-life so the next renewal stays in the
+    /// future while retaining time to renew before expiry.
     pub token_renewal_margin: Duration,
     /// Delay before retrying a failed token renewal.
     pub token_renewal_retry_delay: Duration,
--- a/crates/ably-subscriber/tests/integration/auth_token.rs
+++ b/crates/ably-subscriber/tests/integration/auth_token.rs
@@ -216,6 +216,9 @@ async fn token_renewal() {
 
 #[tokio::test]
 async fn repeated_short_lived_tokens_are_spaced_without_starving_messages() {
+    const SHORT_TOKEN_TTL_MS: i64 = 4_000;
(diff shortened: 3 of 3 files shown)


seven332:
## Code Review: PR #29129

**Reviewed head:** `4fdbd5652afdbff6b8e9aca8710976dc2848d460`

### Summary

The centralized adaptive margin removes the successful short-token renewal loop without adding configuration or changing protocol and failure behavior. The final integration scenario verifies bounded callback traffic, queued-message delivery, and a later renewal before expiry through the public subscription path.

### Findings

No P0, P1, or P2 findings.

### Testing

- `cargo test -p ably-subscriber`
- exact timing regression repeated on the final test parameters
- repository pre-commit Rust formatting, Clippy, rustdoc, and file-size checks

CI status is evaluated separately from this code review.

### Verdict

LGTM

PR #29132, opened 2026-08-25
fix(runner): wake placement workers during shutdown

### LINKED ISSUE #29115: perf(rust): wake tool placement worker during exec shutdown
## Problem

Every agent run enables process control in `crates/runner/src/executor/agent_run.rs:2113-2123`. The corresponding guest exec creates a `WorkloadPlacementBootstrap` in `crates/vsock-guest/src/exec_operation.rs:985-1009`, including a tool-placement worker that remains alive for the duration of the controlled process.

`WorkloadPlacementBootstrap::drop` sets an atomic cancellation flag and immediately joins its workers (`crates/vsock-guest/src/process_containment.rs:121-125`). The tool worker checks that flag only between calls to `accept_with_timeout` (`crates/vsock-guest/src/process_containment.rs:884-891`), and each accept can block for `WORKLOAD_BOOTSTRAP_ACCEPT_POLL`, currently 100 ms (`crates/vsock-guest/src/process_containment.rs:43`). There is no wakeup path when cancellation is set.

Normal exec finalization drops the bootstrap synchronously before containment cleanup and before the terminal result is completed (`crates/vsock-guest/src/exec_operation.rs:720-736`). When the tool-placement listener is idle, the operation therefore waits for the remainder of the current 100 ms polling window before it can continue.

## Impact

This adds 0-100 ms of avoidable terminal tail latency to every controlled agent run, including successful runs with no pending tool placement. It delays the guest exec result, containment cleanup, sandbox finalization/reuse, and the host-visible run completion path. The relative cost is especially material for short runs and accumulates across run throughput.

## Acceptance criteria

- Cancelling `WorkloadPlacementBootstrap` actively wakes or otherwise interrupts the idle placement listener so worker shutdown does not depend on the 100 ms accept timeout.
- Preserve the existing invariant that placement workers are fully stopped before recursive cgroup cleanup begins.
- Preserve worker ownership: shutdown must still join all started workers and report panics.
- Add a regression test that parks the tool-placement worker in an idle accept, cancels it, and proves shutdown completes promptly without waiting for the polling interval.

seven332:
## Summary

- replace 100 ms timeout polling in workload and tool placement workers with an operation-local cancellation pipe
- close the bootstrap-owned writer before joining workers so both idle accepts wake through persistent `POLLHUP`
- preserve cancellation precedence, authenticated placement behavior, worker panic reporting, and cleanup ordering
- cover both idle placement workers with a real-fd shutdown regression

## Scope

- no host/guest protocol, environment variable, public API, or persisted-state changes
- already accepted placement-handshake I/O keeps its existing bounded timeout behavior

## Validation

- `cargo clippy --manifest-path crates/Cargo.toml -p vsock-guest --all-targets --all-features -- -D warnings`
- `cargo doc --manifest-path crates/Cargo.toml -p vsock-guest --all-features --no-deps`
- `cargo test --manifest-path crates/Cargo.toml -p vsock-guest --all-targets --all-features`
- repository pre-commit Rust format, workspace docs, workspace Clippy, and file-size checks

Closes #29115

### DIFF
--- a/crates/vsock-guest/src/process_containment.rs
+++ b/crates/vsock-guest/src/process_containment.rs
@@ -433,13 +433,12 @@ impl CgroupGuard {
                 io::Error::other("trusted control cgroup has no tools domain"),
             )
         })?;
-        let (workload_cancel_reader, cancel_wake_writer) =
-            placement_cancel_pipe().map_err(|error| {
-                ProcessContainmentError::new("prepare placement cancellation", error)
-            })?;
-        let tool_cancel_reader = workload_cancel_reader.try_clone().map_err(|error| {
-            ProcessContainmentError::new("clone placement cancellation reader", error)
+        let (cancel_reader, cancel_wake_writer) = placement_cancel_pipe().map_err(|error| {
+            ProcessContainmentError::new("prepare placement cancellation", error)
         })?;
+        let cancel_reader = Arc::new(cancel_reader);
+        let workload_cancel_reader = Arc::clone(&cancel_reader);
+        let tool_cancel_reader = Arc::clone(&cancel_reader);
         let mut cancel_wake_writer = Some(cancel_wake_writer);
         let cancel = Arc::new(AtomicBool::new(false));
         let worker_cancel = Arc::clone(&cancel);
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #29132

**Reviewed head:** `ae24ba42b65f2337c8bf80566a57361b8571d336`

### Summary

The placement bootstrap now actively wakes both idle workers by closing its sole close-on-exec cancellation-pipe writer before joining them. The shared read fd keeps the resource cost to two descriptors, and cancellation remains fail-closed when listener readiness races with shutdown.

### Findings

No findings.

### Testing

Affected-crate formatting, Clippy, documentation, and all-target/all-feature tests passed locally (251 executed tests). Real-fd regressions cover normal accept, cancellation precedence with a queued connection, and Drop waking and joining both idle workers. Three existing root-only production identity tests were not run locally and remain delegated to Rust CI.

### Verdict

LGTM

PR #29177, opened 2026-08-25
test(guest-agent): exercise post-rejection event delivery

### LINKED ISSUE #29160: test(rust): exercise batches after non-retryable event rejection
## Problem

`crates/guest-agent/tests/event_delivery_nonretryable_failure.rs` says that a non-retryable rejection is followed by later logical batches, but the fixture does not actually create a post-rejection batch.

The test sets `TOTAL_EVENTS` to 34 at line 14, while `crates/guest-agent/src/cli/event_delivery.rs:33` caps each request at 32 events. It also waits for the final sentinel before acknowledging the first request (`event_delivery_nonretryable_failure.rs:57-69`), so all events are already queued. A focused run observed exactly two logical requests:

- sequences `0..31`: HTTP 200
- sequences `32..33`: HTTP 400

The rejected request is therefore the final batch. The loop at `event_delivery_nonretryable_failure.rs:82-91` is never entered, and the FIFO assertion at lines 125-129 only checks the sequences that the test itself appended before the loop. A regression that stops the event sender immediately after a non-retryable 4xx would still pass this test.

`crates/guest-agent/tests/event_delivery_batch_failure.rs:14-118` does exercise later batches after a retryable HTTP 500 exhausts its retries, but it does not pin the distinct single-attempt 4xx path.

## Impact

CI currently gives false confidence that queued events continue after a non-retryable client rejection. A branch-specific regression could silently drop every event queued behind the rejected batch without failing the intended regression test.

## Acceptance criteria

- Make the non-retryable rejection scenario queue at least one complete logical batch after the HTTP 400 batch.
- Assert that the rejected batch has exactly one on-wire attempt and is not retried.
- Assert that at least one later request is observed after the rejection and that all logical sequences remain exact, unique, and FIFO.
- Retain the existing watermark and structured failure-diagnostic assertions.
- Keep the integration test deterministic, time-bounded, and cleanup-safe on failure.

seven332:
## Summary

- expand the non-retryable event rejection fixture to guarantee a complete queued batch remains after the HTTP 400 batch
- unconditionally observe and validate that complete later batch so a stopped sender or rejected-batch retry fails the integration test
- preserve exact FIFO sequence coverage, the frozen watermark, and the existing structured single-attempt failure diagnostics

## Scope

This is a test-only change. It does not modify production event batching, retry behavior, diagnostics, APIs, protocols, persisted state, or deployment compatibility.

## Validation

- `cargo test --manifest-path crates/Cargo.toml -p guest-agent --test event_delivery_nonretryable_failure -- --nocapture`
- `cargo clippy --manifest-path crates/Cargo.toml -p guest-agent --all-targets --all-features`
- `cargo doc --manifest-path crates/Cargo.toml -p guest-agent --all-features --no-deps`
- `cargo test --manifest-path crates/Cargo.toml -p guest-agent`

Closes #29160

### DIFF
--- a/crates/guest-agent/tests/event_delivery_nonretryable_failure.rs
+++ b/crates/guest-agent/tests/event_delivery_nonretryable_failure.rs
@@ -11,7 +11,8 @@ use serde_json::json;
 use std::io;
 use std::time::Duration;
 
-const TOTAL_EVENTS: usize = 34;
+const EVENT_BATCH_SIZE: usize = 32;
+const TOTAL_EVENTS: usize = EVENT_BATCH_SIZE * 3;
 
 #[tokio::test]
 async fn nonretryable_client_rejection_records_one_attempt()
@@ -70,6 +71,7 @@ async fn nonretryable_client_rejection_records_one_attempt()
 
     let rejected_request = server.next_request(Duration::from_secs(5)).await?;
     let rejected_sequences = common::event_request_sequences(&rejected_request.request)?;
+    assert_eq!(rejected_sequences.len(), EVENT_BATCH_SIZE);
     let rejected_request_id = rejected_request
         .request
         .client_request_id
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #29177

**Reviewed head:** `7eed88794e0934bc673d8e1ecc88733c6dd565b6`

### Summary

The test-only change closes the vacuous continuation coverage by guaranteeing a full HTTP 400 batch and then unconditionally observing a full later batch. The 96-event construction remains deterministic whether the first request contains one event or the full 32, and the tail loop safely drains the partial-first-request case.

### Findings

No findings.

### Testing

The focused integration test passed, including 10 consecutive review runs that exercised both partial and full first-request scheduling. Rust formatting, affected-crate Clippy, affected-crate documentation, the full `guest-agent` test suite, and repository commit hooks passed locally. Unrelated Turbo and Python checks were not run because the PR changes only a Rust integration test; remote CI is evaluated separately.

### Verdict

LGTM

PR #29184, opened 2026-08-25
fix(runner): keep cancellation waits off ably supervisor loop

### LINKED ISSUE #29149: bug(rust): prevent Ably cancellation from blocking supervisor shutdown
## Problem

`AblySupervisor` handles each Ably message inline on its single supervisor task. A cancel message awaits the per-run cancellation transfer gate. While that wait is pending, the task cannot process later Ably events or observe its shutdown tokens, and `AblySupervisor::shutdown` waits for the same task without a timeout or abort path.

This couples the entire provider shutdown and every later Ably notification to one run transfer boundary.

## Evidence

- `crates/runner/src/provider/api_ably_supervisor.rs:537-560` selects shutdown only in the outer loop, then awaits the full message handler inline.
- `crates/runner/src/provider/api_ably_supervisor.rs:655-666` awaits cooperative or hard cancellation delivery.
- `crates/runner/src/run_cancellation.rs:188-205` shows that both cancellation requests wait for `transfer_gate`.
- During exact-reuse activation, `crates/runner/src/cmd/start/job_discovery.rs:452-538` keeps that transfer guard across `build_spawn_job_request`; `crates/runner/src/cmd/start/job_discovery.rs:607-614` awaits active-status persistence before the guard is released.
- `crates/runner/src/status.rs:419-435` serializes persistence behind an async mutex and awaits the filesystem write.
- `crates/runner/src/provider/api_ably_supervisor.rs:481-488` cancels the shutdown token and then awaits the supervisor task, but an in-flight cancel handler does not select that token.
- Runner teardown awaits provider shutdown before draining running jobs at `crates/runner/src/cmd/start/mod.rs:2189-2219`.
- Hard-stop signal handling publishes `Stopping` before awaiting per-job cancellation delivery. The main loop can therefore enter provider teardown while the signal handler is still waiting on the same transfer gate, making independent supervisor shutdown progress necessary.

## Failure scenario

1. An exact-reuse job holds its cancellation transfer guard while publishing active status.
2. An Ably cancel notification for that run enters the supervisor and waits for the guard.
3. Status persistence is slow or stalls, so the guard remains held.
4. Later job, cancel, connection, and active-input notifications are not processed because the supervisor is still awaiting the first handler.
5. If runner shutdown starts, `AblySupervisor::shutdown` cannot interrupt that handler and provider teardown remains blocked before running-job drain.

The same head-of-line blocking occurs for any long transfer boundary even when shutdown is not in progress.

## Decision

Proceed with one runner-only PR. Keep the per-run transfer gate, but remove its wait from the sole Ably event-loop branch.

## Chosen direction

- Resolve the current `RunCancellationHandle` when the cancel message is consumed.
- Move that exact handle and parsed mode into a supervisor-owned delivery future.
- Retain pending deliveries in `FuturesUnordered` and poll completions as another branch of the existing supervisor `tokio::select!`.
- Return to the receive loop immediately after enqueueing so unrelated Ably events and supervisor shutdown remain selectable.
- On supervisor exit, drop pending delivery futures instead of waiting for a transfer guard held by later teardown work.
- Leave non-cancel message handling, cancellation token semantics, and the transfer gate unchanged.

Resolving the handle before deferral is required: a delayed registry lookup could otherwise target a replacement registration that reused the same run ID.

## Alternatives considered

- **Spawn deliveries into a `JoinSet`:** correct but adds abort/drain/panic lifecycle and per-message task scheduling without a need for independent tasks.
- **Dedicated bounded dispatcher with per-run coalescing:** adds queue overflow/backpressure policy and a worker state machine without evidence of cancellation-volume pressure.
- **Select shutdown against the existing inline wait:** bounds shutdown but still blocks unrelated messages, so it does not solve the full issue.
- **Remove or bypass the transfer gate:** rejected because it would violate activation/cancellation ownership ordering.

## Risks and mitigations

- **Registration replacement race:** bind the exact handle before enqueueing.
- **Pending work not polled:** keep completion polling in the production supervisor select and cover completion after guard release.
- **Shutdown drains indefinitely:** use structured ownership; exiting the loop drops pending futures.
- **Repeated same-run messages:** preserve the transfer mutex and monotonic, idempotent per-mode signals; do not add speculative coalescing.
- **Burst growth:** no current evidence justifies a new bounded queue or loss/backpressure policy. Revisit only with production volume evidence.

## Acceptance criteria

- Supervisor and provider shutdown terminate an in-flight Ably cancellation wait promptly.
- A cancellation blocked on one run transfer gate does not prevent unrelated Ably notifications from being processed.
- Releasing the guard during normal operation delivers cancellation to the same registration observed when the message was consumed.
- Per-run transfer ordering remains intact and cancellation requests are not lost during normal operation.
- Existing cooperative, hard, legacy-default-hard, and unknown-default-hard behavior remains intact.
- Regression coverage holds one run transfer guard, injects its cancel notification, and verifies unrelated notification progress, normal delivery after release, and bounded supervisor shutdown before release.
- Runner formatting, Clippy, documentation, and crate tests pass.

## Delivery notes

This change stays inside one runner artifact. It changes no Ably payload, API, persisted state, queue payload, runner/guest contract, migration, or rollout ordering. Rollback requires no cleanup.

seven332:
## Summary

- move transfer-gated Ably cancellation delivery into supervisor-owned in-flight futures
- bind the current run cancellation registration before deferring delivery
- keep unrelated Ably notifications and supervisor shutdown responsive while a transfer guard is held
- cover normal delivery, registration replacement safety, unrelated notification progress, and bounded shutdown

## Scope

- leaves the per-run transfer gate and cancellation signal semantics unchanged
- leaves non-cancel Ably handling and all external protocols unchanged
- adds no detached tasks, worker queue, timeout, migration, or compatibility path

## Validation

- exact new cancellation progress and shutdown regression tests
- `cargo test -p runner` (3288 passed, 21 ignored; local non-root execution used a test-only `cp` permission shim for the existing fake-mitmdump retry helper, with no test skipped)
- pre-commit workspace Rust format, Clippy, rustdoc, and file-size hooks

Closes #29149

### DIFF
--- a/crates/runner/src/provider/api_ably_supervisor.rs
+++ b/crates/runner/src/provider/api_ably_supervisor.rs
@@ -14,9 +14,12 @@
 //! Ably connection state all route through `PollWakeups` so a runner can still
 //! discover work through HTTP polling.
 
+use std::future::Future;
+use std::pin::Pin;
 use std::sync::{Arc, Mutex as StdMutex};
 use std::time::{Duration, Instant as StdInstant};
 
+use futures_util::{StreamExt, stream::FuturesUnordered};
 use tokio::sync::{Mutex, Notify};
 use tokio_util::sync::CancellationToken;
 use tracing::{error, info, warn};
@@ -32,7 +35,7 @@ use crate::active_input::ActiveInputNotifications;
 use crate::duration::duration_ms;
 use crate::ids::RunId;
 use crate::retry::{RetryState, recv_retry, sleep_until_retry};
-use crate::run_cancellation::RunCancellationRegistry;
+use crate::run_cancellation::{RunCancellationHandle, RunCancellationRegistry};
 use crate::types::ConnectorRuntimeTarget;
 
 const ABLY_BACKOFF_INITIAL: Duration = Duration::from_secs(5);
@@ -41,6 +44,7 @@ const ABLY_DISCONNECT_ERROR_AFTER: Duration = Duration::from_secs(60);
 const DEFERRED_POLL_MAX: Duration = Duration::from_secs(10);
 type AblyConnectHandle =
     tokio::task::JoinHandle<Result<ably_subscriber::Subscription, ably_subscriber::Error>>;
+type CancelDelivery = Pin<Box<dyn Future<Output = ()> + Send>>;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub(crate) enum PollReason {
@@ -525,6 +529,7 @@ impl Drop for AblySupervisor {
 
 async fn run_supervisor(config: SupervisorTaskConfig) {
     let mut ably: Option<ably_subscriber::Subscription> = None;
+    let mut cancel_deliveries = FuturesUnordered::<CancelDelivery>::new();
     let mut ably_retry: RetryState<AblyConnectHandle> =
         RetryState::new(ABLY_BACKOFF_INITIAL, ABLY_BACKOFF_MAX, None);
     ably_retry.restart_at = Some(StdInstant::now());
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #29184

**Reviewed head:** `89bdbd5737087d463e7e266b835b465f966fad6c`

### Summary

Cancellation delivery now remains owned by the Ably supervisor without waiting on a run transfer gate inside the sole receive branch. The implementation binds the current registration before deferral, preserves existing signal semantics, and drops pending delivery futures on shutdown.

### Findings

No findings.

### Testing

The regression coverage exercises the production enqueue seam, held-gate unrelated-message progress, delivery to the originally resolved registration after replacement, and bounded shutdown before guard release. Rust format, Clippy, rustdoc, 41 supervisor module tests, and the full runner suite passed. External Ably end-to-end traffic was not exercised locally.

### Verdict

LGTM

PR #29202, opened 2026-08-25
test(guest-agent): observe claude tracker overflow behavior

### LINKED ISSUE #29192: test(rust): make Claude tracker overflow test observe capacity
## Problem

`crates/guest-agent/tests/claude_tool_tracking_overflow.rs:9-12` hard-codes 257 supported tool-use events against the private 256-entry limit in `crates/guest-agent/src/cli/claude.rs:7-12`. The stream then ends with an immediate successful result at lines 34-42, and the test only asserts a clean process outcome at lines 51-60.

The capacity decision itself is implemented in `crates/guest-agent/src/cli/claude.rs:78-91`, the end-to-end event wiring is in `crates/guest-agent/src/cli/mod.rs:1521-1526`, and the watchdog is first polled after the 5-second interval configured in `crates/guest-agent/src/constants.rs:43-44`. A focused run completes in 0.27 seconds, before any watchdog poll.

Consequently, this test remains green if the capacity guard is removed and the 257th call is retained, if the production limit grows above the hard-coded fixture, or if the main loop stops calling the tracker. It proves that this particular stream exits cleanly, but it does not prove that the end-to-end path reached or preserved the bounded overflow behavior named by the test.

## Impact

A regression can make the watchdog state grow beyond its intended bound or evict an older genuinely stuck call while CI still reports the overflow regression test as passing.

## Acceptance criteria

- The integration test establishes an observable outcome that distinguishes rejecting the over-capacity call from retaining it, evicting an admitted call, or bypassing tracker wiring.
- The test fails if the production capacity guard is removed or the tracker call is disconnected from Claude stdout processing.
- The fixture cannot silently drift below the production capacity.
- The test uses deterministic synchronization or controlled time and does not add a multi-minute wall-clock wait.

seven332:
## Summary

- replace the Claude tracker overflow test's clean-exit assertion with an observable watchdog
  state transition through the real CLI execution path
- fill the tracker, reject an extra `WebFetch`, release every admitted call, then prove a later
  `WebSearch` occupies the freed slot and triggers structured watchdog termination
- synchronize on a padded agent-log ingestion fence and advance Tokio time instead of waiting for
  the production timeout

## Scope

This is a test-only `guest-agent` change. It does not change production behavior, public APIs, the
mock protocol, event schemas, persisted data, runner protocols, dependencies, or CI configuration.

The fixture still repeats the private 256-entry capacity, but the assertion now fails if production
capacity grows above it: the extra older `WebFetch` becomes retained and wins watchdog selection
instead of the expected later `WebSearch`.

## Validation

- `cargo fmt --all --check`
- `cargo test -p guest-agent --test claude_tool_tracking_overflow`
- `cargo test -p guest-agent --all-targets --all-features`
- pre-commit workspace Cargo fmt, docs, Clippy, file-size check, and commitlint
- manual mutation checks verified the focused test fails when the capacity guard is bypassed, the
  stdout-to-tracker call is disabled, or production capacity is raised above the fixture

Closes #29192

### DIFF
--- a/crates/guest-agent/tests/claude_tool_tracking_overflow.rs
+++ b/crates/guest-agent/tests/claude_tool_tracking_overflow.rs
@@ -1,29 +1,34 @@
-//! Claude tool tracking must keep the run alive when the bounded watchdog
-//! state cannot admit another unmatched call.
+//! Claude tool tracking must reject an over-capacity call while preserving
+//! watchdog coverage for a later call admitted into a freed slot.
 
 mod common;
 
+use guest_contracts::diagnostics::{CliTerminationReason, CliTerminationSignal};
 use serde_json::json;
 use std::time::Duration;
 
-// Keep this one above MAX_TRACKED_STUCK_TOOLS in cli::claude.rs. The stream is
-// intentionally made of supported network tools so the test exercises tracker
-// admission without involving event delivery or the stuck-tool timeout.
-const TRACKED_TOOL_CAPACITY_PLUS_ONE: usize = 257;
+// Match MAX_TRACKED_STUCK_TOOLS in cli::claude.rs. If production capacity grows
+// above this fixture, the overflow WebFetch is retained and the WebSearch
+// assertion below fails instead of silently weakening the boundary coverage.
+const TRACKED_TOOL_CAPACITY: usize = 256;
+
+// Keep the fence larger than the 8 KiB agent-log buffer. Observing the marker
+// on disk then proves all preceding events completed the sequential read loop.
+const AGENT_LOG_FENCE_PADDING_BYTES: usize = 16 * 1024;
 
 #[tokio::test]
-async fn claude_tool_tracking_overflow_does_not_terminate_run()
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #29202

**Reviewed head:** `ab0d178efd68f24b9350321199154009e0358173`

### Summary

The test now exercises the real Claude JSONL ingestion, bounded tracker, result removal, slot reuse,
watchdog selection, and process reaping path. The ordered `WebFetch`/`WebSearch` discriminator makes
overflow retention, eviction, result-removal regressions, tracker disconnection, and capacity drift
observable without exposing private production state.

### Findings

No findings.

### Testing

Rust formatting, guest-agent Clippy/docs, the focused overflow integration test, and the complete
guest-agent all-target/all-feature test target passed locally. Pre-commit workspace Rust checks and
commitlint passed. Manual mutations confirmed the focused test fails when the capacity guard is
bypassed, tracker wiring is disabled, or production capacity exceeds the fixture. Remote CI was
still pending at review time.

### Verdict

LGTM

PR #29238, opened 2026-08-25
test(python): align selective json oracle markers

### LINKED ISSUE #29213: test(python): align selective JSON differential oracle with exact marker semantics
## Problem

`crates/runner/mitm-addon/tests/test_json_selective_observations.py:51-60` builds the reference result for wildcard array counts by excluding every JSON object key that starts with `\0__vm0_json_`.

The runtime contract is narrower: `crates/runner/mitm-addon/src/usage/json_selective.py:44-51` defines two exact string markers (`\0__vm0_json_unknown_key__` and `\0__vm0_json_array_element__`), and `crates/runner/mitm-addon/src/usage/json_selective.py:1275-1280` excludes only members of that exact marker set. The focused regressions at `crates/runner/mitm-addon/tests/test_json_selective_strings.py:493-516` also cover those exact values.

As a result, a valid user key such as `"\u0000__vm0_json_custom"` is counted by `JsonSelectiveExtractor`, but `_expected_common_extraction()` incorrectly omits it. A randomized differential pass exposed this mismatch; replacing the oracle's prefix check with exact-marker semantics made 20,000 generated documents agree across whole-body, randomized-chunk, and bytewise feeds.

## Impact

The current reference helper is not an independent model of the parser contract for arbitrary untrusted JSON keys. Broader property or fuzz coverage produces false failures for legitimate prefix-sharing keys, and changing production code to satisfy this oracle would incorrectly hide user-provided arrays.

## Acceptance criteria

- Make `_expected_common_extraction()` exclude only the exact internal marker keys used by the parser.
- Add a regression proving that the two exact internal markers remain excluded while a distinct key sharing the prefix is included.
- Exercise the corrected oracle across multiple input chunk boundaries so the differential test remains valid for streaming input.

seven332:
## Summary

- align the selective JSON reference oracle with the parser's exact internal marker semantics
- cover both exact marker exclusion and valid prefix-sharing keys across all existing streaming chunk sizes

## Scope

- test-only change in `crates/runner/mitm-addon`
- no production behavior, API, persistence, protocol, dependency, or rollout changes

## Validation

- `uv run --no-sync python -m pytest tests/test_json_selective_observations.py` (33 passed)
- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (6606 passed)

Closes #29213

### DIFF
--- a/crates/runner/mitm-addon/tests/test_json_selective_observations.py
+++ b/crates/runner/mitm-addon/tests/test_json_selective_observations.py
@@ -20,6 +20,12 @@ _COMMON_SCALAR_FIELDS = {
 _COMMON_ARRAY_COUNT_PATHS = {("data",), ("errors",)}
 _COMMON_WILDCARD_ARRAY_COUNT_PATHS = {("includes", "*")}
 _COMMON_OBJECT_PRESENCE_PATHS = {(), ("data",)}
+_WILDCARD_INTERNAL_MARKER_KEYS = frozenset(
+    (
+        "\0__vm0_json_unknown_key__",
+        "\0__vm0_json_array_element__",
+    )
+)
 
 
 def _get_path(data: object, path: tuple[str, ...]) -> tuple[object, bool]:
@@ -56,7 +62,7 @@ def _expected_common_extraction(data: object):
             for key, value in includes.items()
             if isinstance(key, str)
             and isinstance(value, list)
-            and not key.startswith("\0__vm0_json_")
+            and key not in _WILDCARD_INTERNAL_MARKER_KEYS
         }
         if counts:
             wildcard_array_counts[("includes", "*")] = counts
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #29238

**Reviewed head:** `70b14edf7660f08b1d81df371eac5d575397a672`

### Summary

The test oracle now models exact internal marker membership without importing production-private data, and the regression exercises both reserved markers plus a valid prefix-sharing key through every existing streaming chunk size.

### Findings

No P0, P1, or P2 findings.

### Testing

Focused observations tests passed (33), all package static checks passed, and the complete mitm-addon suite passed (6606). An explicit reproduction also passed for chunk sizes 1, 2, 3, 5, 8, 13, and the whole body. CI remains the only external verification pending.

### Verdict

LGTM

PR #29249, opened 2026-08-25
test(runner): cover concrete active jobs service gate

### LINKED ISSUE #29208: test(rust): cover concrete active-jobs service gate
## Problem

`crates/runner/src/cmd/service/gate.rs:190-245` is the concrete safety boundary that decides whether `runner service stop` and `runner service uninstall` may proceed while a runner has active jobs. No automated test currently calls `check_active_jobs_gate` itself:

- `crates/runner/src/cmd/service/gate.rs:256-545` tests `read_runner_status` and the pure `decide_gate` helper, but not the orchestration that handles `--force`, queries systemd, derives the runner base directory, applies fail-open status handling, and constructs `RunnerError::ActiveJobs`.
- `crates/runner/src/cmd/service/stop.rs:167-173` wires the concrete gate into `RealServiceStopOps`, while the stop tests replace the entire gate with the fake implementation at `crates/runner/src/cmd/service/stop.rs:667-677`. Those tests therefore prove call ordering and synthetic error propagation only.
- `crates/runner/src/cmd/service/mod.rs:515-520` calls the concrete gate for uninstall, but the service module has no uninstall gate test.
- The runner behavior scripts invoke stop/uninstall with `--force` for cleanup, so they bypass the concrete active-job branch rather than covering it.

The original feature issue, #9062, explicitly required an integration test for active-job refusal. PR #9093 added parser and pure-decision tests, but left the live active-job verification in its manual test plan. The current suite still has that gap.

## Impact

A regression in the real orchestration can pass every current test even if it ignores `--force`, misclassifies the systemd state, resolves the wrong status path, changes documented fail-open behavior, loses active run IDs, or wires stop and uninstall differently. In the destructive direction, an ordinary no-force stop could reach `systemctl stop` and kill in-flight jobs while the suite remains green. In the opposite direction, stale or unreadable state could incorrectly block an operator from stopping a dead runner.

## Acceptance criteria

- [ ] Add deterministic automated coverage that invokes the concrete `check_active_jobs_gate` orchestration, not only `read_runner_status`/`decide_gate` and not a fake gate callback.
- [ ] Cover an active unit with running jobs and assert the returned `RunnerError::ActiveJobs` preserves the unit, suffix, run IDs, runner uptime, command name, and non-draining UX variant.
- [ ] Cover the draining variant and both concrete command names (`stop` and `uninstall`).
- [ ] Prove `force = true` bypasses the active-job check before systemd/status work.
- [ ] Cover the documented bypasses for an inactive unit, a unit-state lookup failure, and missing or malformed status data.
- [ ] Exercise the real stop and uninstall wiring and prove no stop/uninstall mutation is attempted after the gate refuses.
- [ ] Keep asynchronous synchronization bounded and deterministic; do not rely on sleeps or ambient host systemd state.

seven332:
## Summary

- make the concrete active-jobs gate accept the caller's runner home and a narrow systemd unit-state boundary
- route both stop and uninstall orchestration through that same gate instead of replacing it with a synthetic callback in tests
- cover force, unavailable unit state, missing/malformed status, normal stop refusal, draining uninstall refusal, complete error context, and mutation suppression with real temporary status files

Closes #29208.

## Behavior and compatibility

Production stop and uninstall behavior is unchanged: `--force` still bypasses before systemd/status work, unavailable state still fails open, and active jobs still refuse before mutation. `uninstall_service_unit` remains gate-free for its existing GC caller.

This PR changes no CLI flags or output, API, database, queue payload, runner/guest protocol, persisted status schema, or deployment contract.

## Validation

- `cargo test -p runner --bin runner cmd::service::gate` — 24 passed
- `cargo test -p runner --bin runner cmd::service::stop` — 24 passed
- `cargo test -p runner --bin runner cmd::service::tests` — 13 passed
- `cargo test -p runner --bin runner cmd::service` — 250 passed, 1 ignored
- `cargo fmt --all -- --check` — passed
- `cargo clippy -p runner --all-targets --all-features -- -D warnings` — passed
- `cargo doc -p runner --all-features --no-deps` — passed
- pre-commit workspace Rust format, documentation, Clippy, file-size, and commitlint hooks — passed
- `cargo test -p runner --all-targets --all-features` — 3293 passed, 21 ignored, with one unrelated failure in `proxy::process::tests::initial_start_retries_an_occupied_selected_port`; the exact same failure was reproduced on detached `origin/main` at `e76130d842`

## Exclusions

- no active-jobs policy or TOCTOU behavior change
- no systemctl output-parser change
- no ambient-systemd or process-level E2E test
- no fix for the independently reproducible proxy port-retry test failure

### DIFF
--- a/crates/runner/src/cmd/service/gate.rs
+++ b/crates/runner/src/cmd/service/gate.rs
@@ -6,19 +6,22 @@ use crate::paths::HomePaths;
 use crate::status_file::{self, StatusFileReadError, StatusForGate};
 use tracing::{info, warn};
 
+use super::ServiceFuture;
 use super::diagnostic::status_field_preview;
-use super::systemctl::is_unit_active;
 use super::target::RunnerServiceUnit;
 
-/// Resolve the runner's base_dir from its service name suffix using the
-/// project-wide convention: `/var/lib/vm0-runner/runners/<suffix>/`.
+pub(super) trait ActiveJobsGateOps {
+    fn is_unit_active<'a>(&'a mut self, unit: &'a RunnerServiceUnit) -> ServiceFuture<'a, bool>;
+}
+
+/// Resolve the runner's base_dir from its service name suffix and the caller's
+/// runner home using the project-wide `runners/<suffix>/` convention.
 ///
 /// This matches `ansible/playbooks/build-runner.yml` and the `--runner-dirname`
-/// default in `runner config`. Non-standard `base_dir` overrides (dev-only)
-/// will fail to locate status.json and fall through to forceful stop.
-pub(super) fn runner_base_dir(unit: &RunnerServiceUnit) -> Option<PathBuf> {
-    let home = HomePaths::new().ok()?;
-    Some(home.runners_dir().join(unit.suffix()))
+/// default in `runner config`. In production the home is
+/// `/var/lib/vm0-runner`; tests can supply an isolated home root.
+fn runner_base_dir(home: &HomePaths, unit: &RunnerServiceUnit) -> PathBuf {
+    home.runners_dir().join(unit.suffix())
 }
 
 /// Parsed snapshot of the runner's status.json.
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #29249

**Reviewed head:** `ed307bfdfd08266a22da01fa2457a42c6db7b587`

### Summary

The concrete active-jobs gate is now exercised with real isolated status files while only the external systemd state query is controlled. Stop and uninstall both compose the same production gate before mutation, and production force, fail-open, lock ordering, and GC behavior remain unchanged.

### Findings

No findings. A self-review timing-stability issue in the initial uptime assertions was resolved on this head by bounding the value with timestamps captured around each gate invocation.

### Testing

Targeted gate coverage passes (24 tests), both command refusal paths pass, and the runner service suite passes (250 passed, 1 ignored). Rust formatting, Clippy with warnings denied, rustdoc, and pre-commit hooks pass. The full runner all-target suite has one unrelated proxy port-retry failure that was reproduced on the base commit; no ambient-systemd/process E2E was run.

### Verdict

LGTM

PR #29194, opened 2026-08-25
fix(auth): handle clerk device trust in auth v2

hulh122:
## Summary

- add a dedicated single-worker Auth v2 Playwright CI lane with 18 browser scenarios
- complete Clerk Device Trust email-code verification after password sign-in instead of failing closed on a supported second factor
- keep unsupported Device Trust factors fail-closed and preserve safe diagnostics, back navigation, and masked identifiers
- keep provider-backed browser coverage on happy paths and remove retry/error-injection helpers that made the lane unstable

## Implementation

- prepare and attempt the Clerk email-code second factor when sign-in reaches needs_client_trust
- expose the second-factor state through the existing Auth v2 state machine and reuse the existing email-code UI
- add integration coverage for successful Device Trust verification and unsupported-factor handling
- simplify the Clerk Playwright resource helper and use complete valid sign-up data before verification

## Validation

- relevant Prettier checks
- Platform lint
- full Turbo TypeScript check (14/14 tasks)
- Platform scoped Knip
- Auth v2 integration tests (48 passed)
- full Platform Vitest suite (2035 passed)
- E2E TypeScript check
- E2E helper tests (39 passed)
- Auth v2 Playwright discovery (18 tests)

The required GitHub x64 Knip job remains the authoritative full-repository check; the local arm64 container produces reproducible baseline cross-workspace false positives that are unrelated to this diff.

Refs #28927

Issue #28927, opened 2026-08-24
feat(auth): build custom Clerk authentication UI under /v2 routes


hulh122:
## Background

The platform currently renders Clerk's prebuilt `<SignIn>` and `<SignUp>` components in `turbo/apps/platform/src/views/auth/auth-page.tsx`, then applies a large set of overrides against Clerk's internal `.cl-*` DOM in `auth-layout.tsx` and `auth-clerk-appearance.ts`.

That UI boundary is unstable for us: internal markup and CSS changes can break individual steps even when the surrounding auth page has not intentionally changed. The email-code screen has already regressed repeatedly (#3814), and the prebuilt flow also required a compatibility guard for duplicate verification emails (#22052).

Build a vm0-owned authentication UI under versioned routes while continuing to use Clerk for identity, verification, sessions, organizations, and security policy. The v2 flow must reach feature parity with the Clerk prebuilt auth components for the production configuration before it is considered ready for rollout.

## Current capability baseline

As of 2026-08-24, the production Clerk instance exposes the following user-facing auth capabilities:

- Email/password sign-in
- Email-code sign-in
- Google OAuth and Google One Tap
- Passkey sign-in
- Password reset by email code
- Public progressive sign-up with required email/password, optional first/last name, email verification, legal consent, and Turnstile CAPTCHA
- Multi-session support and Clerk continuation/task states
- Forced organization selection after authentication; organization creation remains hidden on the auth surface
- VM0/Okou brand selection, allowed cross-origin redirects, and all currently supported locales

This is a point-in-time baseline, not a hard-coded strategy list. The custom UI should render from the factors and statuses returned by Clerk where practical and must fail visibly and safely when Clerk returns an unsupported state.

### 1. Add isolated v2 routes

- [ ] Add `/v2/sign-in` and `/v2/sign-in/*` route handling.
- [ ] Add `/v2/sign-up` and `/v2/sign-up/*` route handling.
- [ ] Keep `/sign-in` and `/sign-up` on the existing prebuilt components during development and initial verification so rollback and side-by-side comparison remain possible.
- [ ] Preserve deep links and refreshes for every nested auth step.

### 2. Build a vm0-owned Clerk flow layer

- [ ] Use Clerk's lower-level React hooks/resources for flow operations; do not render Clerk's prebuilt `<SignIn>` or `<SignUp>` components anywhere under `/v2`.
[...]

PR #29251, opened 2026-08-25
fix(auth): localize remaining auth v2 copy

hulh122:
## Summary

- Replace the final exact-English Auth v2 recovery, CAPTCHA, and edit-action values in all nine non-English platform catalogs.
- Add an exact eight-key regression allowlist that must be non-empty and must differ from `en-US` in every non-English catalog.
- Keep the assertion limited to the confirmed critical keys; this does not impose a blanket translation rule on unrelated copy.

## Issue context

Part of #28927. This closes the confirmed Auth v2 non-English catalog gap only. It does not change production routing, v1 auth, Auth v2 state machines, sign-in/sign-up navigation, the sign-up initial-config adapter, or the E2E files owned by #29194.

## Locales

- `de-DE`
- `es-ES`
- `fr-FR`
- `hi-IN`
- `id-ID`
- `it-IT`
- `ja-JP`
- `ko-KR`
- `pt-BR`

## Critical keys

- `auth.v2.signIn.editIdentifier`
- `auth.v2.signIn.unknownError`
- `auth.v2.signUp.unknownError`
- `auth.v2.signUp.captchaLoading`
- `auth.v2.signUp.captchaSubtitle`
- `auth.v2.signUp.captchaTitle`
- `auth.v2.signUp.editEmailAddress`
- `auth.v2.signUp.retry`

## Verification

- `pnpm format`
- `NODE_OPTIONS=--max-old-space-size=3072 pnpm turbo run lint --concurrency=1 --log-order grouped`
- `pnpm knip`
- `pnpm --filter @okouai/app exec vitest run src/views/auth-v2/__tests__/auth-v2-i18n.test.ts` — 1 file, 5 tests passed on exact head
- Deterministic scope audit — 72 values across 9 locales; only the 8 approved keys changed in each catalog

## Preview QA boundary

Signed-out Auth v2 has no locale selector or locale query. It resolves locale before the application bundle from the active workspace cache or the browser language, while the visible language selector requires an active workspace. A preview-only browser/runtime override would not exercise a supported selector/query path, so locale verification remains at the deterministic catalog-test boundary: all nine non-English resources are imported, the exact eight-key allowlist is asserted non-empty, and every allowlisted value must differ from `en-US`.


hulh122:
LGTM


PR_AUTO_REVIEW_COMMIT: 99d758125a0d30292895f1d253132908030ccefc
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

This focused fix replaces the eight confirmed English-copy Auth v2 values in each of the nine non-English platform catalogs and adds an exact allowlisted regression assertion. The diff does not change runtime behavior, routing, v1 auth, Auth v2 state/navigation, the initial-config adapter, or E2E coverage.

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- Coverage: Adequate. The existing Auth v2 platform localization integration file imports the real catalog resources, already enforces complete/non-empty Auth v2 key parity, and now asserts that this explicit non-empty eight-key allowlist differs from `en-US` across every supported non-English locale.
- Conventions: Compliant. The regression adds no internal mocks, fetch stubs, fake timers, direct component rendering, or implementation-call assertions. The deterministic catalog boundary is appropriate because signed-out Auth v2 exposes neither a locale selector nor a locale query; the authenticated workspace selector cannot exercise that signed-out route.

PR #29253, opened 2026-08-25
perf(python): bound oversized unicode labels before punycode

### LINKED ISSUE #29217: perf(python): bound Unicode host normalization before punycode encoding
## Problem

`crates/runner/mitm-addon/src/host_normalization.py` performs the full Unicode normalization, validation, and Python punycode encoding pipeline before enforcing the 63-byte DNS label limit:

- `_canonical_punycode_label()` encodes the complete normalized label at lines 347-359.
- `_normalize_label()` calls that path at lines 395-408 and only checks `len(ascii_label) > 63` at lines 409-410.

This function is reachable with sandbox-controlled HTTP authority data on the synchronous mitmproxy hook path:

- `mitm_addon.requestheaders()` classifies the request synchronously at lines 745-793.
- `request_classification.py:423` calls `get_trusted_authority()`.
- `request_authority.py:108-126` admits raw `Host` values up to 4,096 bytes, and lines 266-270 normalize each authority assertion.
- HTTP/2/3 `:authority` is also normalized, but it is not counted by the raw `Host`-field budget.

A diverse Unicode label that can never fit in a DNS label therefore consumes increasingly expensive punycode work before the request is rejected.

## Evidence

Baseline: `f07d4c6afa2f160a78f235aa00405ceb41cd12b8`  
Environment: CPython 3.12.3, aarch64

The benchmark constructed labels with distinct CJK code points and called `normalize_idna_hostname()` three times per size. Every case below ended with `UnicodeError: IDNA label too long`:

| Input | UTF-8 bytes | Median |
| --- | ---: | ---: |
| 682 code points | 2,046 | 88 ms |
| 1,024 code points | 3,072 | 200 ms |
| 1,365 code points | 4,095 | 354 ms |

A wider scaling run showed 512 code points at 52 ms, 1,024 at 207 ms, 2,048 at 829 ms, 4,096 at 3.34 s, and 8,192 at 13.0 s, which is consistent with approximately quadratic growth for this input family.

Minimal reproducer:

```python
label = "".join(chr(0x4E00 + i) for i in range(1365))
normalize_idna_hostname(label)  # rejects only after about 354 ms on the measured host
```

## Impact

One malformed 4,095-byte `Host` value can occupy the mitmproxy event-loop path for roughly 350 ms before denial. A sandbox workload can repeat these requests and delay unrelated proxied traffic in the same runner. Larger authority values amplify the stall wherever the protocol layer permits them.

## Acceptance criteria

- Bound work for untrusted `Host` and `:authority` labels before invoking the full punycode encoder when the label cannot produce a DNS-compliant result.
- Preserve all currently accepted valid Unicode hostname and canonical A-label behavior.
- Add regression coverage for a diverse oversized Unicode `Host` value at the current 4,096-byte boundary and for an oversized HTTP/2/3 `:authority` value.
- Demonstrate that rejection cost no longer grows approximately quadratically beyond the DNS label limit.

seven332:
## Summary

- derive a 59-code-point upper bound for inputs entering the canonical punycode encoder from the existing 63-byte DNS label limit and `xn--` prefix
- reject impossible normalized labels before CPython's repeated-scan punycode path while preserving unsafe-mapping/text-validation precedence and the final encoded-length check
- cover 4,095-byte HTTP/1.1 Host plus HTTP/2 and HTTP/3 authority inputs through the real request hook, and preserve an accepted decomposed Unicode label that normalizes to a 63-byte A-label

## Performance evidence

Pinned CPython 3.14.0, distinct CJK labels, median rejection time:

| Code points | Before | After |
| ---: | ---: | ---: |
| 128 | 1.24 ms | 0.115 ms |
| 256 | 4.35 ms | 0.219 ms |
| 512 | 16.24 ms | 0.470 ms |
| 1,024 | 67.38 ms | 0.992 ms |
| 1,365 (4,095 UTF-8 bytes) | 125.33 ms | 1.124 ms |
| 2,048 | 263.87 ms | 1.724 ms |

The deterministic integration regression also makes `encodings.punycode.punycode_encode` fail if any oversized authority reaches it, avoiding a wall-clock CI assertion.

## Exclusions

- no new raw Host or HTTP/2/3 authority size policy
- no custom/third-party punycode implementation, caching, timeout, or async offload
- no API, protocol, persisted-state, dependency, migration, or rollout change

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_host_normalization.py` (71 passed)
- `uv run --no-sync python -m pytest tests/` (6,610 passed; 59 existing dependency deprecation warnings)
- `lefthook run pre-commit`
- commit-msg `commitlint`

Closes #29217

### DIFF
--- a/crates/runner/mitm-addon/src/host_normalization.py
+++ b/crates/runner/mitm-addon/src/host_normalization.py
@@ -30,6 +30,7 @@ _IDNA_DOT_TRANSLATION = str.maketrans(
     }
 )
 _PUNYCODE_PREFIX = "xn--"
+_PUNYCODE_INPUT_MAX_CODEPOINTS = _DNS_LABEL_MAX_LENGTH - len(_PUNYCODE_PREFIX)
 _UNICODE_CONTROL_CATEGORY_PREFIX = "C"
 _UNICODE_MARK_CATEGORY_PREFIX = "M"
 _BIDI_ARABIC_NUMBER = "AN"
@@ -349,6 +350,10 @@ def _canonical_punycode_label(label: str) -> str:
         raise UnsafeIdnaCompatibilityMappingError("unsafe IDNA compatibility mapping")
     normalized_label = _normalize_label_text(label)
     _validate_normalized_label_text(normalized_label)
+
+    # Punycode emits at least one payload byte for every input code point.
+    if len(normalized_label) > _PUNYCODE_INPUT_MAX_CODEPOINTS:
+        raise UnicodeError("IDNA label too long")
 
     try:
         payload = normalized_label.encode("punycode").decode("ascii").lower()
--- a/crates/runner/mitm-addon/tests/test_host_normalization.py
+++ b/crates/runner/mitm-addon/tests/test_host_normalization.py
@@ -54,6 +54,13 @@ def test_ascii_label_contract(label, expected):
     assert host_normalization.normalize_idna_label(label) == expected
 
 
+def test_unicode_label_at_dns_limit_is_accepted():
+    normalized = host_normalization.normalize_idna_label("é" * 57)
+
+    assert normalized == f"xn--9c{'a' * 57}"
+    assert len(normalized) == 63
+
+
 @pytest.mark.parametrize("hostname", INVALID_IDNA_HOSTNAME_CASES)
 def test_rejects_invalid_idna_hostname(hostname):
     with pytest.raises(UnicodeError):
--- a/crates/runner/mitm-addon/tests/test_host_normalization.py
+++ b/crates/runner/mitm-addon/tests/test_host_normalization.py
@@ -54,8 +54,8 @@ def test_ascii_label_contract(label, expected):
     assert host_normalization.normalize_idna_label(label) == expected
 
 
-def test_unicode_label_at_dns_limit_is_accepted():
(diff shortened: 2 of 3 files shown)


seven332:
## PR Review: LGTM

Reviewed head: `1049a43cc6394b5636d092a055f472f0cb8b6f6f`

### Summary

- Derives the pre-encoding bound from the existing 63-byte DNS label limit and four-byte `xn--` prefix.
- Preserves unsafe-mapping and normalized-text validation precedence, as well as the final encoded-length check for shorter inputs that still expand beyond the DNS limit.
- Adds real request-hook regression coverage for HTTP/1.1, HTTP/2, and HTTP/3, plus an accepted decomposed Unicode boundary case.

### Findings

No findings.

### Verification

- `git diff --check origin/main...HEAD`
- Previously run on this exact head: `uv lock --check`, `uv sync --locked`, Ruff format/check, BasedPyright, 71 host-normalization tests, the full 6,610-test mitm-addon suite, lefthook pre-commit, and commitlint.

### Remaining risk

The regression deliberately guards the vulnerable encoder call instead of asserting wall-clock latency, so absolute timing remains environment-dependent. Linear normalization and validation remain intentional to preserve existing error classification.

PR #29269, opened 2026-08-25
fix(python): reject null sigv4 session token config

### LINKED ISSUE #29228: fix(python): reject malformed awsSigv4 matcher inputs
## Problem and evidence

`crates/runner/mitm-addon/src/matching.py` validates `auth.awsSigv4` before a compiled firewall can match requests or enter the ordinary-credential authority index. The matcher-level malformed-auth tests do not cover the individual SigV4 shape guards for non-object values, unknown fields, required credentials, or the optional session token.

Repository inspection also found a real fail-closed mismatch, not only missing coverage. `_auth_config_is_valid()` reads `sessionToken` with `dict.get()`, so a missing field and an explicitly present JSON `null` are treated alike. On current `main` (`6b39284f6b9f967fd52ab6a7a7ddbf64e6e4fcd9`), an API configured with valid required SigV4 credentials and `"sessionToken": null`:

- returns `FirewallAllow` for a matching authorized request;
- is admitted by `matches_ordinary_credential_authority()`; and
- is therefore not classified as malformed.

That behavior conflicts with the canonical TypeScript firewall schema, where `sessionToken` is optional but must be a non-empty string when present. The TypeScript firewall matcher likewise distinguishes `undefined` from `null` and rejects the latter.

## Impact

Firewall definitions arrive through registry JSON and are compiled before request matching. Accepting an explicit null session token weakens the intended fail-closed boundary and can enroll malformed configuration in credential-bearing request handling. The uncovered rejection branches also allow future regressions in the other SigV4 shape guards to pass the current suite.

Tests for malformed auth-service responses exercise a different backend-response/cache boundary and do not protect registry-to-matcher behavior.

## Decision

Treat this as a focused matcher correctness fix plus regression matrix, not a test-only coverage task. Preserve omission of the optional session token as valid, but reject an explicitly present null value together with the already-rejected malformed shapes.

Keep the change within the existing compiled matcher and its integration-style malformed-auth tests. Do not add a schema abstraction, dependency, API change, or broader auth validation refactor.

## Alternatives considered

- **Add tests only and omit `sessionToken: null`:** rejected because it would encode only the currently passing branches while leaving the reproduced fail-open mismatch in place.
- **Reuse the auth-service response parser tests:** rejected because that parser validates resolved backend credentials, not registry configuration or authority-index admission.
- **Introduce a declarative Python schema for all firewall config:** rejected as disproportionate to this focused mismatch and a new source of contract duplication.

## Risks and compatibility

- Valid configurations with an omitted or non-empty string session token remain unchanged.
- The only runtime behavior change is that malformed registry input containing an explicit null session token fails closed.
- No API, runner protocol, persisted state, database, deployment ordering, or migration change is required.
- Keep the case matrix deterministic and representative rather than duplicating every JSON scalar type already covered at other validation boundaries.

## Acceptance criteria

- Make matcher validation distinguish an omitted `sessionToken` from a present null value and reject the latter.
- Extend `test_compiled_firewall_malformed_auth.py` with parametrized `awsSigv4` cases covering:
  - a representative non-object value;
  - an unknown SigV4 field;
  - missing, non-string, and empty `accessKeyId`;
  - missing, non-string, and empty `secretAccessKey`;
  - null, another non-string, and empty `sessionToken`.
- For every malformed case, assert that an unrelated authority remains unmatched.
- For every malformed case, assert that the matching base returns `FirewallBlock` with reason `malformed_firewall_config` and no denied permissions.
- For every malformed case, assert that the API is excluded from `matches_ordinary_credential_authority()`.
- Preserve successful matching and ordinary-credential authority admission for valid SigV4 configuration with the session token omitted or set to a non-empty string.
- Run all required formatting, lint, type, and pytest checks for `crates/runner/mitm-addon`.

seven332:
## Summary

- reject an explicitly present null AWS SigV4 `sessionToken` during compiled firewall validation
- preserve valid omitted and non-empty string session tokens
- cover every SigV4 config shape guard through compiled request matching and ordinary-credential authority admission

## Problem

The Python matcher used `dict.get()` for the optional SigV4 session token, so an omitted field and a present JSON null were treated alike. A malformed registry entry with `sessionToken: null` could therefore authorize a matching request and enter the ordinary-credential authority index even though the canonical TypeScript firewall schema rejects null.

## Implementation

- validate the session token only when its key is present, then require a non-empty string
- add a focused malformed matrix for non-object input, unknown fields, required credential shapes, and optional token shapes
- assert unrelated-base isolation, `malformed_firewall_config`, empty denied permissions, and authority-index exclusion for every malformed case
- assert valid SigV4 configurations with omitted and present tokens still match and receive authority admission

This PR does not change the API, runner protocol, registry format, persistence, database schema, cache format, or deployment ordering.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_compiled_firewall_malformed_auth.py -q` (78 passed)
- `uv run --no-sync python -m pytest tests/` (6619 passed)

Closes #29228

### DIFF
--- a/crates/runner/mitm-addon/src/matching.py
+++ b/crates/runner/mitm-addon/src/matching.py
@@ -429,11 +429,12 @@ def _auth_config_is_valid(api_entry: dict) -> bool:
             return False
         if not raw_aws_sigv4["secretAccessKey"]:
             return False
-        optional_value = raw_aws_sigv4.get("sessionToken")
-        if optional_value is not None and not isinstance(optional_value, str):
-            return False
-        if optional_value == "":
-            return False
+        if "sessionToken" in raw_aws_sigv4:
+            optional_value = raw_aws_sigv4["sessionToken"]
+            if not isinstance(optional_value, str):
+                return False
+            if not optional_value:
+                return False
         if raw_auth.get("headers"):
             return False
         if raw_auth.get("query"):
--- a/crates/runner/mitm-addon/tests/test_compiled_firewall_malformed_auth.py
+++ b/crates/runner/mitm-addon/tests/test_compiled_firewall_malformed_auth.py
@@ -119,6 +119,130 @@ def test_malformed_auth_config_fails_closed_after_base_match(auth_config):
 
 
 @pytest.mark.parametrize(
+    "aws_sigv4",
+    [
+        pytest.param([], id="non-object"),
+        pytest.param(
+            {
+                "accessKeyId": "key",
+                "secretAccessKey": "secret",
+                "unknown": "value",
+            },
+            id="unknown-field",
+        ),
+        pytest.param({"secretAccessKey": "secret"}, id="access-key-missing"),
+        pytest.param(
+            {"accessKeyId": 123, "secretAccessKey": "secret"},
+            id="access-key-non-string",
+        ),
+        pytest.param(
+            {"accessKeyId": "", "secretAccessKey": "secret"},
+            id="access-key-empty",
+        ),
+        pytest.param({"accessKeyId": "key"}, id="secret-key-missing"),
+        pytest.param(
+            {"accessKeyId": "key", "secretAccessKey": 123},
+            id="secret-key-non-string",
+        ),
+        pytest.param(
+            {"accessKeyId": "key", "secretAccessKey": ""},
+            id="secret-key-empty",
+        ),
+        pytest.param(
+            {
+                "accessKeyId": "key",
+                "secretAccessKey": "secret",
+                "sessionToken": None,
+            },
+            id="session-token-null",
+        ),
+        pytest.param(
+            {
+                "accessKeyId": "key",
+                "secretAccessKey": "secret",
+                "sessionToken": 123,
+            },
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #29269

**Reviewed head:** `cf11818e3dc746ec39c45da4ddd2ab85011b7dcd`

### Summary

The matcher now distinguishes an omitted SigV4 session token from an explicitly present null and fails closed for the latter. The implementation is local, matches the canonical TypeScript contract, and the focused integration matrix covers both request matching and ordinary-credential authority admission without adding abstraction or delay.

### Findings

No P0, P1, or P2 findings.

### Testing

- Focused malformed-auth module: 78 passed.
- Focused SigV4 subset: 13 passed in 0.03s.
- Complete mitm-addon suite: 6619 passed with 59 existing dependency warnings.
- uv lock/sync, Ruff format/lint, and BasedPyright all passed.
- Remote CI remains the separate readiness gate.

### Verdict

LGTM

PR #29271, opened 2026-08-25
test(runner): make mitmdump retry fixture work for non-root users

### LINKED ISSUE #29248: test(runner): make mitmdump retry fixture work for non-root users
## Background

The runner test `proxy::process::tests::initial_start_retries_an_occupied_selected_port` launches the same fake mitmdump script twice: the first launch is expected to lose a selected-port race, and the second launch must bind a fresh port successfully.

The fake script captures its environment on every launch with:

```bash
cp "/proc/$$/environ" "$0.environ"
```

This test passes in root-run CI but fails reproducibly under the repository's non-root development container user.

## Observed behavior

From `crates/` as the non-root `vscode` user:

```bash
cargo test -p runner --bin runner proxy::process::tests::initial_start_retries_an_occupied_selected_port -- --exact --nocapture
```

The first launch emits the expected bind collision:

```text
Transparent Proxy failed to listen on *:<port>: [Errno 98] Address already in use
```

The retry then exits before binding:

```text
cp: cannot create regular file '/tmp/<tempdir>/fake-mitmdump.environ': Permission denied
called `Result::unwrap()` on an `Err` value: Internal("mitmdump exited immediately with 1")
```

`/proc/$$/environ` is read-only. The first `cp` creates `$0.environ` without an owner write bit, so a later launch under a non-root user cannot overwrite the shared capture file. Root execution masks this fixture defect.

## Expected behavior

The occupied-port recovery test should pass under both root CI and the documented non-root development environment. Its retry should fail only for the injected first-port collision, not for test-artifact permissions.

## Acceptance criteria

- Make repeated fake-mitmdump launches publish the environment capture without relying on root permission to overwrite a read-only destination.
- Preserve the existing environment-content assertions used by sibling proxy process tests.
- Preserve the real child-process and TCP-listener coverage of initial startup port recovery.
- Run the exact failing test repeatedly as a non-root user.
- Run runner Rust formatting, Clippy, rustdoc, and affected tests.

## Relevant code

- `crates/runner/src/proxy/process.rs`
- `write_fake_listening_mitmdump`
- `initial_start_retries_an_occupied_selected_port`

seven332:
## Summary

- force the fake mitmdump fixture to replace its environment capture when startup retries reuse the same path
- preserve the raw NUL-delimited environment artifact and its existing compatibility assertion
- preserve the real child-process and TCP-listener coverage of occupied-port recovery

## Scope

This changes runner test infrastructure only. Production proxy startup behavior, protocols, persisted data, dependencies, and deployment compatibility are unchanged.

## Validation

- `cargo test -p runner --bin runner proxy::process::tests::initial_start_retries_an_occupied_selected_port -- --exact` (three consecutive non-root runs)
- `cargo test -p runner --bin runner proxy::process::tests::spawn_mitmdump_passes_runner_options_and_preserves_ca -- --exact`
- `cargo test -p runner --bin runner proxy::process::tests::` (32 passed)
- `cargo test -p runner --bin runner` (3,289 passed, 21 ignored)
- pre-commit workspace Rust formatting, Clippy, rustdoc, file-size, and commit-message hooks

Closes #29248

### DIFF
--- a/crates/runner/src/proxy/process.rs
+++ b/crates/runner/src/proxy/process.rs
@@ -887,7 +887,7 @@ set -euo pipefail
 printf '%s\n' "$@" > "$0.args"
 printf '%s\n%s\n%s\n%s\n' "$TMPDIR" "$OKOU_MITMDUMP_RUNTIME_DIR" \
   "$VM0_MITMDUMP_RUNTIME_DIR" "${OKOU_MITM_RUNNER_TOKEN-}" > "$0.env"
-cp "/proc/$$/environ" "$0.environ"
+cp -f "/proc/$$/environ" "$0.environ"
 port=""
 ready_path=""
 usage_state_id=""


seven332:
## Code Review: PR #29271

**Reviewed head:** `5ede8c61787ae76892ceb2fee66df3dc3b57586c`

### Summary

The test-only change gives the fake mitmdump environment capture the replacement semantics required by sequential startup retries. It preserves the raw NUL-delimited artifact, stable path, production retry logic, and real child-process/TCP coverage.

### Findings

No findings.

### Testing

The exact regression test passed three consecutive times as the non-root development user. The raw-environment sibling test, all 32 proxy process tests, the full runner binary suite (3,289 passed; 21 ignored), runner formatting, Clippy, rustdoc, and repository hooks also passed. GitHub CI remains independently tracked by the PR workflow.

### Verdict

LGTM

Issue #29223, opened 2026-08-25
perf: debounce connector account search requests


seven332:
## Background

The connector account manager switches to server-side search for larger account sets. This was inspected while reviewing PR #29196.

## Problem

The input value is used directly by the asynchronous account-list computation. Each input event resets pagination and starts another request. Typing `abc` produced three GET requests (`a`, `ab`, and `abc`) plus their CORS preflights.

The intermediate requests are unnecessary, can make the list flicker through loading states, and scale poorly for users who search frequently across large account sets.

## Acceptance criteria

- The search input remains responsive and updates immediately while the server query is debounced by a short, documented interval.
- Rapid typing is coalesced into a bounded number of account-list requests.
- Superseded requests are cancelled when supported, and stale responses cannot replace results for the current query.
- Clearing search promptly restores the unfiltered first page.
- Pagination state resets only for the effective server query.
- Built-in and custom connector account managers share the same behavior.
- Platform integration coverage verifies rapid input and clearing behavior without using arbitrary sleeps.

## Relevant context

- Parent objective: #22832
- Found during PR #29196 self-review
- Signal implementation: `turbo/apps/platform/src/signals/okou-page/settings/connector-accounts.ts`

## Scope boundary

This issue covers frontend request timing and stale-result behavior. Adding UUID fallback-name matching and changing default-account ordering are tracked separately.

seven332:
## Challenge Review: Framing

**Checkpoint:** `framing`
**Outcome:** `proceed`

The issue is necessary and correctly scoped. The current `setSearch$` writes every input value directly into the dependency read by the asynchronous first-page computation and resets pagination on every keystroke. The resulting request has no cancellation signal, while the existing Platform integration test verifies only the eventual filtered result and does not constrain request count, clearing semantics, or out-of-order completion.

The acceptance criteria address the root timing and concurrency problem rather than only hiding loading flicker: keep the controlled input immediate, delay only the effective server query, reset pagination at that effective-query boundary, cancel superseded work through the existing signal hierarchy where supported, and prevent stale results from becoming current. Clearing must bypass the delay so the unfiltered first page returns promptly.

This remains one independently reviewable Platform PR. Built-in and custom targets already share the same account-list signal factory, so separate implementations would add inconsistency without benefit. Fallback-name matching (#29222) and default-account discoverability (#29225) remain correctly out of scope.

seven332:
## Research Phase

# Research: Debounce Connector Account Search Requests

## Scope

[...]

Issue #29222, opened 2026-08-25
bug: make connector account fallback names searchable


seven332:
## Background

The connector account manager displays unnamed accounts as `Account #<first 8 UUID characters>`. The manager uses server-side search once a connector has more than six accounts or more pages.

This was reproduced while reviewing PR #29196.

## Problem

The account list API searches `displayName`, `externalEmail`, `externalUsername`, and `externalId`, but not the connector connection UUID used by the displayed fallback name. Searching for a visible value such as `1c076c6b` therefore returns `No accounts found`.

For a paginated account list, the user cannot reliably locate an unnamed account using the only name shown by the UI.

This affects built-in and custom connector accounts because they share the same list endpoint and manager.

## Acceptance criteria

- Searching by the visible short UUID portion of `Account #<short-id>` returns the matching account.
- Matching works across the full paginated result set, not only accounts already loaded in the browser.
- Existing searches by display name, external email, external username, and external ID continue to work.
- Built-in and custom connector account targets behave consistently.
- API integration coverage separates invalid input from a valid query with no matches.
- The change preserves compatibility with currently deployed frontend clients and existing opaque cursors.

## Relevant context

- Parent objective: #22832
- Found during PR #29196 self-review
- API query: `turbo/apps/api/src/signals/services/connector-account-lifecycle.service.ts`
- UI fallback label: `turbo/apps/platform/src/views/okou-page/components/settings/connector-account-manager-dialog.tsx`
- Reproduction screenshot: `tmp/account-connector-ui-review-20260825-135809-09-fallback-name-not-searchable.png`

## Scope boundary

This issue covers matching the identifier already presented by the UI. Search request debouncing and default-account ordering are tracked separately.

seven332:
## Challenge Review: Framing

**Checkpoint:** `framing`
**Outcome:** `proceed`

The issue is necessary and correctly scoped. The manager renders an unnamed account as `Account #<first 8 connection-ID characters>`, while the target-scoped API applies server-side search only to display and provider identity fields. That mismatch is observable once account loading is paginated: the only identifier shown for an unnamed account cannot locate that account across the full result set.

This is the root search-contract defect rather than a client-side filtering or pagination symptom. The existing issue boundary remains independently reviewable: align the current API search with the already-visible connection identifier and add route integration coverage for built-in and custom targets, pagination-independent matching, invalid input, and valid no-match behavior. No frontend label, endpoint shape, or opaque cursor change is required. Proceed to planning.

[...]

PR #29282, opened 2026-08-25
fix(connectors): keep default accounts visible in manager

### LINKED ISSUE #29225: bug: keep default connector accounts discoverable in long lists
## Background and evidence

The first connected account normally becomes the connector default. Target-scoped account pages contain 50 accounts and are ordered by newest creation first. The account manager renders only those paginated results in that order.

This was reproduced while reviewing PR #29196: with eight test accounts, the oldest/default account was the last row and required scrolling. With more than 50 newer accounts, that default is absent from the initial manager page and requires loading later pages.

The existing bounded connector summary already carries `defaultConnection`, and the connected connector card displays its effective account label. That makes the default identifiable from Settings, but it does not keep the corresponding row discoverable or manageable after the account manager opens.

## Problem

The account manager has no stable, immediately visible representation of the current default independent of newest-first pagination. The default determines inherited thread behavior, but an old default can be buried below the scroll viewport or on a later page. An unnamed default is harder to correlate because its effective label is the fallback short connection ID.

Changing the default refreshes the list and summary, but the newly selected default remains at its creation-time position. The user can therefore lose sight of the account they just made default.

## Scope

This issue covers default-account discoverability and management inside the account manager, including the pagination-safe data flow required to represent the default independently of the current page.

It does not cover account search matching or request debouncing, which are tracked separately in #29222 and #29223. It also does not change default-selection semantics, account authorization, runtime credential resolution, or thread inheritance.

## Verified constraints

- The summary endpoint is already bounded and returns the exact default connection for each logical target.
- The paginated account endpoint uses a creation-time cursor and must not duplicate or omit accounts when additional pages load.
- The first page, search results, and loaded pages currently merge into one account list.
- Setting a default invalidates both the bounded summaries and the manager list without closing the dialog.
- Built-in and user-managed custom connector accounts share the same account manager and lifecycle signals.
- Frontend and API versions deploy independently, so any contract change must remain additive and tolerate old clients.

## Acceptance criteria

- The current default account and its management actions are discoverable immediately inside the account manager without scrolling or loading later pages.
- The presentation remains clear with long account names, reconnect-required status, and an unnamed fallback account.
- Changing the default updates the immediately visible representation without closing the manager.
- The default is not rendered twice when it also appears in the current paginated or searched result set.
- Loading subsequent pages remains stable and does not duplicate or omit non-default accounts.
- Search behavior has an explicit, consistent rule for the default representation and does not show an account that contradicts the active query as a search result.
- Built-in and user-managed custom connector accounts behave consistently.
- API integration coverage includes a default account older than at least one full page of newer accounts if the implementation changes API behavior; Platform integration coverage exercises the same user-visible condition.
- Any cursor or response change preserves deployment compatibility with currently deployed clients.

## Relevant context

- Parent objective: #22832
- Found during PR #29196 self-review
- API ordering: `turbo/apps/api/src/signals/services/connector-account-lifecycle.service.ts`
- Account summary/list signals: `turbo/apps/platform/src/signals/okou-page/settings/connector-accounts.ts`
- Account card and manager: `turbo/apps/platform/src/views/okou-page/components/settings/connector-card.tsx` and `connector-account-manager-dialog.tsx`
- Reproduction screenshot: `tmp/account-connector-ui-review-20260825-135809-08-default-buried-desktop.png`

seven332:
## Summary

- keep the exact default connector account in a fixed manager section outside the paginated scroll area
- reuse the existing bounded account summary for built-in and user-managed custom targets
- remove the pinned connection ID from loaded/search result rows without changing cursor state
- preserve rename, reconnect, delete, fallback-label, and reconnect-required behavior on the pinned row

## Behavior

- an oldest default remains visible and actionable even when more than 50 newer accounts precede it
- loading the page that also contains the default does not render a duplicate or omit non-default accounts
- search keeps the default as structurally separate target context while the result list remains filtered
- setting a new default updates the fixed row in the still-open manager and returns the previous default to the ordinary list

## Scope and compatibility

- Platform-only presentation and state selection
- no API contract, route, service, database, cursor, Runner protocol, permission, persisted-state, runtime credential, or thread-inheritance change
- consumes the already-deployed `defaultConnection` summary field, so independently deployed frontend/API revisions remain compatible
- search matching and request debouncing remain in #29222 and #29223

## Validation

- `pnpm -F @okouai/app exec vitest run src/views/okou-page/__tests__/connectors-page.test.tsx --maxWorkers=1 --silent=passed-only` (109 passed)
- `pnpm -F @okouai/app lint`
- pre-commit: affected-file Prettier, full-repository Knip, sequential workspace type checks (14/14), file-size check, and commitlint
- `git diff --check`

Closes #29225

Parent context: #22832

### DIFF
--- a/turbo/apps/platform/src/signals/okou-page/settings/connector-accounts.ts
+++ b/turbo/apps/platform/src/signals/okou-page/settings/connector-accounts.ts
@@ -17,7 +17,9 @@ const CONNECTOR_ACCOUNT_PAGE_SIZE = 50;
 
 export type ConnectorAccountMutationVersion = number | string | null;
 
-function connectorAccountTargetKey(target: ConnectorAccountTarget): string {
+export function connectorAccountTargetKey(
+  target: ConnectorAccountTarget,
+): string {
   return target.kind === "builtin"
     ? `builtin:${target.connectorSlug}`
     : `custom:${target.customConnectorId}`;
--- a/turbo/apps/platform/src/views/okou-page/__tests__/connectors-page.test.tsx
+++ b/turbo/apps/platform/src/views/okou-page/__tests__/connectors-page.test.tsx
@@ -1501,28 +1501,34 @@ describe("connectors page", () => {
       throw new Error("Expected GitHub fixture connector");
     }
     const accounts = Array.from({ length: 101 }, (_, index) => {
+      const isDefault = index === 0;
       return {
-        id:
-          index === 0
-            ? "00000000-0000-4000-a000-000000000001"
-            : crypto.randomUUID(),
+        id: isDefault
+          ? "00000000-0000-4000-a000-000000000001"
+          : crypto.randomUUID(),
         target: { kind: "builtin" as const, connectorSlug: "github" },
         authMethod: "oauth",
-        displayName: index === 0 ? null : `Work ${index}`,
-        isDefault: index === 0,
+        displayName: isDefault ? null : `Work ${index}`,
+        isDefault,
         externalId: null,
-        externalUsername: index === 0 ? null : `octocat-${index}`,
+        externalUsername: isDefault ? null : `octocat-${index}`,
         externalEmail: null,
         oauthScopes: [],
-        connectionStatus: "connected" as const,
-        reconnectReason: null,
+        connectionStatus: isDefault
+          ? ("reconnect-required" as const)
+          : ("connected" as const),
+        reconnectReason: isDefault
+          ? ("authorization_expired_or_revoked" as const)
+          : null,
         tokenExpiresAt: null,
         createdAt: "2026-01-01T00:00:00.000Z",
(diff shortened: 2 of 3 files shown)


seven332:
## Code Review: PR #29282

**Reviewed head:** `1e7b32ae993c5251f7e96da69453f01beb672f1f`

### Summary

The default account is kept as persistent manager context without changing list pagination or search semantics, and the follow-up availability gate correctly fails closed when the target can no longer be resolved.

### Findings

No findings.

### Testing

The Platform page integration suite covers pagination overlap, search with and without matches, default promotion, reconnect behavior, custom connectors, and unavailable targets. Targeted Vitest passed 109 tests; Platform lint, type checking, scoped Knip, pre-commit checks, and `git diff --check` also passed. Browser E2E was not run.

### Verdict

LGTM

PR #29300, opened 2026-08-25
docs(python): clarify bounded json probe prefix contract

### LINKED ISSUE #29284: docs(python): clarify the bounded JSON probe prefix contract
## Problem

`crates/runner/mitm-addon/src/usage/json_probe.py:61-137` implements a 419-line bounded JSON prefix parser, but `probe_top_level_string_field()` has only a one-line docstring. `TopLevelStringFieldProbeResult` documents the result statuses, yet it does not make the parser's most important prefilter property explicit: once the first matching top-level string value is complete, the function returns `found` without scanning or validating the rest of the document.

That behavior is intentional and tested in `crates/runner/mitm-addon/tests/test_json_probe.py:25-32`, where a complete `type` value is followed by a large syntactically unfinished payload and still returns `found`. The helper is consumed by the OpenAI Responses event prefilter in `crates/runner/mitm-addon/src/usage/openai_responses.py:447-474`, so treating `found` as proof of a valid complete JSON document would be incorrect in a usage/timing classification path.

Other caller-relevant details are also absent from the function documentation: how `max_depth`, `max_key_bytes`, and `max_string_bytes` apply while skipping earlier values; that all three bounds must be positive integers; and that the first duplicate top-level field wins while nested fields are ignored.

## Evidence

- The implementation validates all three bounds before reading the body, then scans syntax before the target field and returns immediately after the selected string closes.
- `_validate_positive_int()` raises `TypeError` for non-integers (including booleans) and `ValueError` for non-positive integers.
- The parser skips earlier top-level values, including nested objects, under the configured depth, key-byte, and string-byte bounds; the tests cover oversized skipped values and depth limits.
- The existing `test_probe_finds_top_level_string_field_without_scanning_rest()` test demonstrates that a complete target value is sufficient for `found` even when trailing input is unfinished.
- `openai_responses.py` uses the probe as an event-type prefilter and then performs selective extraction for the paths that require full event data.

## Decision

Proceed with a documentation-only change. The issue is necessary because the public function contract is materially narrower than a complete JSON validator, and the current one-line function docstring leaves caller-visible semantics implicit.

Expand `probe_top_level_string_field()`'s public function docstring rather than changing parser behavior or relying only on module-level documentation. The public call site is where callers discover the signature and bounds, so it is the most discoverable location for the prefix boundary, early `found` return, field matching rules, bound validation, and authoritative-parse requirement. Keep status definitions in the existing result dataclass and cross-check the two descriptions for consistency.

## Alternatives considered

- Leave the current documentation unchanged: preserves an ambiguity that can cause callers to overinterpret `found`.
- Put the complete contract only in the module docstring: centralizes prose but makes parameter-specific behavior less visible when callers inspect the function.
- Split the contract across the function and result docs: maps details to two objects but increases navigation and documentation-drift risk.
- Add or expand tests only: the behavior is already covered by tests, but tests do not provide the public caller contract in the implementation's documentation.
- Change the parser to validate trailing bytes: this would remove the intentional bounded prefilter behavior and expand the issue beyond its documented contract gap.

## Implementation plan

- Replace the one-line function docstring in `crates/runner/mitm-addon/src/usage/json_probe.py` with a structured contract.
- State that the helper probes a bounded JSON object prefix, inspects syntax before the first matching top-level key-colon boundary, and is not a complete JSON validator.
- State that `found` returns after the first complete matching string and does not inspect or validate trailing bytes.
- Document top-level-only matching, first duplicate wins, nested-field skipping, `field_seen`, the three positive-integer bounds, their `TypeError`/`ValueError` validation behavior, and the effect of bounds while skipping earlier values.
- State that callers requiring complete payload validity must perform an authoritative JSON parse after probing.
- Keep the change limited to documentation; no tests or parser branches are expected to change.

## Risks and mitigations

- Documentation can drift if the parser contract changes later. Keep status definitions in the result type, tie the function wording to current control flow and tests, and review both public descriptions together.
- Wording that omits the trailing-byte limitation could recreate the original ambiguity. The existing unfinished-payload test is the reference for this statement.

## Acceptance criteria

- Expand the public function or module documentation to state that this is a prefix prefilter, not a complete JSON validator.
- Explicitly document that `found` stops after the matching string value and does not validate trailing bytes, while syntax before the match is inspected.
- Document the top-level-only and first-duplicate semantics, and connect `field_seen` to the first matching key-colon boundary.
- Document the three bounds, their validation errors, and their effect while skipping values before the target.
- State that callers requiring payload validity must perform an authoritative parse after probing.
- Keep the change documentation-only; no parser behavior changes are required.

## Validation and delivery notes

- Run `uv run --no-sync python -m pytest tests/test_json_probe.py` from `crates/runner/mitm-addon/`.
- Run the addon CI-equivalent `ruff format --check .`, `ruff check .`, and `basedpyright -p .` commands, plus `git diff --check`.
- This is one independently reviewable documentation-only PR with no migration, compatibility, rollout, or additional delivery slice.

seven332:
## Summary

- Expand `probe_top_level_string_field()` documentation to define its bounded JSON prefix-prefilter contract.
- Clarify early `found` termination, top-level and duplicate-field semantics, `field_seen`, bound validation, and the authoritative-parse requirement.

## Scope

This is documentation-only. Parser behavior, result statuses, tests, dependencies, and runtime callers are unchanged.

## Validation

- `/home/user/.local/bin/uv lock --check`
- `/home/user/.local/bin/uv sync --locked`
- `/home/user/.local/bin/uv run --no-sync ruff format --check .`
- `/home/user/.local/bin/uv run --no-sync ruff check .`
- `/home/user/.local/bin/uv run --no-sync basedpyright -p .`
- `/home/user/.local/bin/uv run --no-sync python -m pytest tests/test_json_probe.py` — 57 passed
- `git diff --check`

Closes #29284

### DIFF
--- a/crates/runner/mitm-addon/src/usage/json_probe.py
+++ b/crates/runner/mitm-addon/src/usage/json_probe.py
@@ -66,7 +66,33 @@ def probe_top_level_string_field(
     max_key_bytes: int = 1024,
     max_string_bytes: int = 1024,
 ) -> TopLevelStringFieldProbeResult:
-    """Probe a bounded JSON object prefix for the first top-level string field."""
+    """Probe a bounded JSON object prefix for the first top-level string field.
+
+    This is a prefix prefilter, not a complete JSON validator. The probe
+    inspects the syntax needed to reach the first matching top-level
+    ``field_name``. Once that field's string value is complete, it returns
+    ``found`` without scanning or validating trailing bytes. Syntax before
+    that value is still inspected; callers that require payload validity must
+    perform an authoritative parse of the complete payload after probing.
+
+    Matching is top-level only. Nested fields are skipped, and the first
+    duplicate top-level field wins. ``field_seen`` becomes true at the first
+    matching key-colon boundary, before the value is classified.
+
+    Args:
+        body: JSON bytes or a JSON prefix to inspect.
+        field_name: Top-level field name to find.
+        max_depth: Bound on nested object and array traversal while skipping
+            values before the target field.
+        max_key_bytes: Maximum raw bytes scanned in each JSON object key.
+        max_string_bytes: Maximum raw bytes scanned in each JSON string value,
+            including strings in values skipped before the target.
+
+    All three bounds must be positive integers. Non-integers, including
+    booleans, raise ``TypeError``; non-positive integers raise ``ValueError``.
+    Exceeding a bound while inspecting earlier keys or values can prevent the
+    target field from being reached and returns ``bound_exceeded``.
+    """
 
     _validate_positive_int("max_depth", max_depth)
     _validate_positive_int("max_key_bytes", max_key_bytes)


seven332:
## Code Review: PR #29300

**Reviewed head:** `b9c16091b54d1c093da02572113e1cc882cfa6f5`

### Summary

The PR accurately documents the bounded JSON prefix-prefilter contract at the public function call site. The change is limited to `json_probe.py` documentation and matches the existing implementation, tests, and OpenAI Responses caller.

### Findings

No findings.

### Testing

Ruff format, Ruff lint, BasedPyright, and the targeted `test_json_probe.py` suite passed with 57 tests. No repository-wide test run was needed for this documentation-only change.

### Verdict

LGTM

PR #29301, opened 2026-08-25
perf(python): batch resolved auth header injection

### LINKED ISSUE #29288: perf(python): avoid quadratic resolved auth header injection
## Problem

`crates/runner/mitm-addon/src/auth.py:602-609` validates resolved auth headers and then applies them one at a time:

```python
for header_name, header_value in resolved_auth_header_pairs(headers):
    flow.request.headers[header_name] = header_value
```

mitmproxy `Headers.__setitem__` delegates to `MultiDict.set_all()`, which scans and rebuilds the complete header-field tuple for every assignment. With `m` distinct resolved headers, the tuple grows after every insertion, so this loop performs quadratic work even when the incoming request has only a few headers.

The synchronous helper runs on both credential-injection paths after auth resolution:

- `auth.py:1541-1546` in the ordinary request hook.
- `auth.py:1772-1777` in the requestheaders streaming path.

The input is bounded by bytes but not by entry count:

- `auth_base_transport.py:363-378` validates and filters every pair but does not cap the number of resolved headers.
- `firewall_auth_client.py:842-865` accepts every string-map entry when returned names match the configured names.
- `turbo/packages/connectors/src/firewall-types.ts:42-47`, `turbo/apps/api/src/signals/services/connector-catalog-artifacts/firewall.ts:44-49`, and `turbo/packages/api-contracts/src/contracts/webhooks.ts:525-537,568-575` define header maps without an entry-count limit.
- The firewall-auth response body is allowed up to 256 KiB (`firewall_auth_client.py:70`).

## Evidence

Baseline: `4f9261142f9e841355005b9f9dd1dcbae15179e1`  
Environment: CPython 3.12.3, mitmproxy 12.2.3, aarch64

I benchmarked the production `_apply_header_query_injection()` helper with 16 existing request headers and distinct short resolved headers. Each size was measured three times:

| Resolved headers | Serialized response size | Median |
| ---: | ---: | ---: |
| 500 | 8,492 bytes | 19.7 ms |
| 1,000 | 16,992 bytes | 73.6 ms |
| 2,000 | 34,992 bytes | 281.5 ms |
| 4,000 | 70,992 bytes | 1.105 s |
| 8,000 | 142,992 bytes | 4.458 s |

All tested payloads fit below the existing 256 KiB response-body cap. Doubling the number of headers approximately quadruples elapsed time.

## Impact

A large but schema-valid connector auth map can block the mitmproxy event loop for seconds while applying a cached auth result. The cost recurs on every matching request, including firewall-auth cache hits, delaying that request and unrelated proxy traffic handled by the same addon process. Misconfigured or generated connectors with many header bindings amplify the problem without needing a large upstream response or network failure.

## Acceptance criteria

- Apply validated resolved headers with a single linear merge/rebuild instead of one full `Headers` rebuild per header.
- Preserve current semantics: resolved headers replace all same-name client pairs case-insensitively, repeated unrelated client headers retain their order, and hop-by-hop/authority/framing resolved headers remain excluded.
- Cover both ordinary request and requestheaders credential-injection paths through the shared helper.
- Add regression coverage with a large distinct resolved-header map that demonstrates near-linear scaling, or establish and enforce a small entry-count bound at every configuration/API/runtime boundary in addition to using bounded merge work.

seven332:
## Summary

- merge validated inline firewall-auth headers in one linear raw-field pass instead of rebuilding mitmproxy headers once per resolved field
- preserve case-insensitive replacement, duplicate unrelated client fields, original raw bytes and order, filtered transport fields, and case-equivalent resolved-key behavior
- cover both ordinary request and requestheaders auth paths with a large-map integration regression and document the new test surface

## Scope

- no auth header-count limit or API/configuration contract change
- no change to auth.base forwarding or query injection

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_auth_header_injection.py` (2 passed)
- `uv run --no-sync python -m pytest tests/` (6,627 passed)
- `pnpm exec prettier --check ../docs/testing/mitm-addon-testing.md`

Closes #29288

### DIFF
--- a/crates/runner/mitm-addon/src/auth.py
+++ b/crates/runner/mitm-addon/src/auth.py
@@ -616,6 +616,7 @@ def _apply_header_query_injection(
             if existing_field is None:
                 resolved_fields_by_name[normalized_name] = (encoded_name, encoded_value)
             else:
+                # Match Headers.set_all(): keep the first spelling and let the last value win.
                 resolved_fields_by_name[normalized_name] = (
                     existing_field[0],
                     encoded_value,
--- a/crates/runner/mitm-addon/src/auth.py
+++ b/crates/runner/mitm-addon/src/auth.py
@@ -605,8 +605,33 @@ def _apply_header_query_injection(
     headers: dict[str, str],
     resolved_query: dict | None,
 ) -> None:
-    for header_name, header_value in resolved_auth_header_pairs(headers):
-        flow.request.headers[header_name] = header_value
+    auth_pairs = resolved_auth_header_pairs(headers)
+    if auth_pairs:
+        resolved_fields_by_name: dict[bytes, tuple[bytes, bytes]] = {}
+        for header_name, header_value in auth_pairs:
+            encoded_name = header_name.encode()
+            encoded_value = header_value.encode()
+            normalized_name = encoded_name.lower()
+            existing_field = resolved_fields_by_name.get(normalized_name)
+            if existing_field is None:
+                resolved_fields_by_name[normalized_name] = (encoded_name, encoded_value)
+            else:
+                resolved_fields_by_name[normalized_name] = (
+                    existing_field[0],
+                    encoded_value,
+                )
+
+        remaining_fields = resolved_fields_by_name.copy()
+        merged_fields: list[tuple[bytes, bytes]] = []
+        for header_name, header_value in flow.request.headers.fields:
+            normalized_name = header_name.lower()
+            if normalized_name in remaining_fields:
+                _resolved_name, resolved_value = remaining_fields.pop(normalized_name)
+                merged_fields.append((header_name, resolved_value))
+            elif normalized_name not in resolved_fields_by_name:
+                merged_fields.append((header_name, header_value))
+        merged_fields.extend(remaining_fields.values())
+        flow.request.headers.fields = tuple(merged_fields)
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #29301

**Reviewed head:** `da332da5426a3b8754ea399507b195ac62fe64af`

### Summary

The submitted head replaces quadratic per-header mitmproxy rebuilds with a linear raw-field merge while preserving validation, filtering, casing, duplicate, ordering, and raw-byte behavior. The change remains local to inline firewall-auth and does not alter auth.base or external contracts.

### Findings

No findings. One maintainability concern found during self-review was resolved on this head by documenting the first-spelling/last-value compatibility rule.

### Testing

- current-head focused integration: 2 passed across ordinary request and requestheaders paths
- current-head malformed-header selection: 4 passed
- current-head differential audit: 5,000 generated inputs matched the previous sequential assignment semantics
- current-head scaling audit: 500/1,000/2,000/4,000 fields completed in approximately 0.34/0.68/1.32/2.94 ms
- current-head Ruff, Basedpyright, and documentation Prettier passed
- complete addon suite: 6,627 passed before the comment-only follow-up commit
- required CI checks remain externally verified by the PR pipeline

### Verdict

LGTM

PR #29318, opened 2026-08-25
fix(vsock-host): bound chunked write temp names

### LINKED ISSUE #29290: bug(rust): keep chunked write temp names within NAME_MAX
## Problem

Large ordinary guest file writes stage chunks at a path built by appending `.vm0tmp-{seq}` to the complete destination path in `crates/vsock-host/src/file/write.rs:562-567`.

A destination component can itself be valid at the filesystem `NAME_MAX` boundary (255 bytes on the guest filesystems). The direct path used for content at or below `WRITE_FILE_CHUNK_LIMIT` can create that file, but content above the limit switches to the staging path. Appending the suffix makes the otherwise-valid component exceed `NAME_MAX`, so the first guest write fails with `ENAMETOOLONG`.

The error path at `crates/vsock-host/src/file/write.rs:592-599` then attempts `rm -f` against the same overlong staging path. That cleanup can also fail, leaving the composite normal-operation tracker fail-closed even though no staging file could have been created.

## Concrete failure scenario

1. Use a destination whose final path component is 255 bytes on a filesystem reporting `NAME_MAX=255`.
2. Write content of `WRITE_FILE_CHUNK_LIMIT + 1` bytes.
3. The destination is valid, but `{destination}.vm0tmp-{seq}` is not.
4. `VsockHost::write_file` fails only because the content crossed the chunking threshold; subsequent normal operations may also be unavailable because cleanup could not prove completion.

## Impact

Valid guest paths have content-size-dependent behavior: small writes succeed while chunked writes fail. This can fail run setup or artifact materialization and unnecessarily make an otherwise healthy sandbox connection ineligible for reuse.

## Acceptance criteria

- Construct chunked-write staging files in the destination directory with a unique final component that remains within the filesystem component-length limit regardless of the destination basename length.
- Preserve concurrent-write isolation and the final atomic rename behavior.
- Add coverage for a near-`NAME_MAX` destination with content above the chunk limit, verifying the exact destination bytes, no leftover staging file, and a reusable connection after completion.

seven332:
## Summary

- stage large ordinary guest writes in bounded, same-directory UUID temp files instead of extending the destination basename
- preserve destination protocol validation, per-call cleanup, and atomic `mv -fT` replacement semantics
- cover an exact 255-byte destination basename with real host/guest filesystem integration, including cleanup and connection reuse

## Scope

- no wire-format, guest-helper, public API, database, configuration, or migration changes
- private and single-frame file writes remain unchanged

## Validation

- `cargo clippy -p vsock-host -p vsock-test --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p vsock-host -p vsock-test --all-features --no-deps`
- `cargo test -p vsock-host`
- `cargo test -p vsock-test --test integration`
- pre-commit workspace Rust format, Clippy, rustdoc, file-size, and commitlint hooks

Closes #29290

### DIFF
--- a/crates/vsock-host/src/file/write.rs
+++ b/crates/vsock-host/src/file/write.rs
@@ -556,13 +556,15 @@ impl VsockHost {
                 .await;
         }
 
+        validate_write_file_chunk_request(WriteFileChunkRequest::standard(path, &[], sudo, false))?;
         let _path_guard = self.file_write_path_locks.acquire_shared(path).await;
         let mut normal_operation = CompositeNormalOperation::reserve(&self.shared)?;
 
-        // Write chunks to a per-call temp file, then atomic rename. The
-        // suffix prevents concurrent large writes to the same destination
-        // from appending to or cleaning up each other's staging file.
-        let tmp = format!("{path}.vm0tmp-{}", self.shared.next_temp_seq());
+        // Write chunks to a bounded per-call sibling, then atomic rename. The UUID
+        // prevents concurrent writes from sharing or cleaning up a staging file.
+        let tmp_path = std::path::Path::new(path)
+            .with_file_name(format!(".vm0tmp-{}", uuid::Uuid::new_v4().simple()));
+        let tmp = tmp_path.to_string_lossy();
         let quoted_tmp = quote_shell_arg(&tmp);
         let rm_tmp = format!("rm -f -- {quoted_tmp}");
         let cleanup_armed = Arc::new(AtomicBool::new(false));
--- a/crates/vsock-host/src/tests/file/write_chunked.rs
+++ b/crates/vsock-host/src/tests/file/write_chunked.rs
@@ -1,6 +1,7 @@
 use std::collections::{BTreeMap, BTreeSet};
 use std::future::Future;
 use std::io;
+use std::path::Path;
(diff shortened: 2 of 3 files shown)


seven332:
## Code Review: PR #29318

**Reviewed head:** `3c47e22c2c4d81a1d9e615e6e4dc4c13316c435d`

### Summary

The bounded UUID sibling fixes the `NAME_MAX` regression while preserving same-filesystem atomic replacement, destination validation, cleanup ownership, and the existing host/guest protocol. The change also strengthens concurrent staging isolation without adding a dependency or new lifecycle state.

### Findings

No findings.

### Testing

The regression test uses the real host, guest, helper, exec path, and filesystem with an exact 255-byte basename. It verifies exact destination bytes, no staging remnants, shell-safe rename, and connection reuse. Existing host coverage retains chunking, concurrency, cancellation, cleanup, sudo, tracker, and rename error semantics. Format, Clippy with warnings denied, rustdoc, 365 `vsock-host` tests, 33 `vsock-test` integration tests, and pre-commit workspace Rust checks passed. Full workspace runtime tests remain delegated to CI.

### Verdict

LGTM

PR #29352, opened 2026-08-25
fix(mitm-addon): preserve path base precedence

### LINKED ISSUE #29297: security(python): preserve path-specific denies for long hostnames
## Summary

`_base_specificity()` flattens path and authority specificity into one integer using fixed multipliers. An accepted, DNS-valid hostname with many short labels can make the authority term carry into the path range. A broad exact-host root base then outranks a matching path-specific parameterized base.

`_FirewallMatchCollection.accept_api()` discards every lower-scored API before permission policy evaluation. As a result, an allowed broad permission can survive while a denied permission on the path-specific base is never evaluated.

## Evidence

- `crates/runner/mitm-addon/src/firewall_matching/base_url.py:44-51` defines a `1_000_000` path multiplier, `100` authority multiplier, and `1_000 + len(segment)` literal-segment score.
- `crates/runner/mitm-addon/src/firewall_matching/base_url.py:519-541` adds the weighted path and authority terms into one scalar.
- `crates/runner/mitm-addon/src/host_normalization.py:400-472` validates each DNS label but does not impose a lower total-host bound that would keep the authority score below the path multiplier. The reproducer below is only 203 characters and is DNS-valid.
- `crates/runner/mitm-addon/src/matching.py:1230-1241` clears prior matches for a higher scalar and rejects lower scalars before policy evaluation.
- `crates/runner/mitm-addon/tests/test_compiled_firewall_base_specificity_precedence.py:442-481` establishes that a path-specific base deny must block an overlapping root allow.

## Static reproduction

Use a valid host made from 100 one-character labels followed by `com`:

```python
host = ".".join(["a"] * 100 + ["com"])
```

Configure two APIs under one firewall owner:

```python
{
    "base": f"https://{host}",
    "permissions": [{"name": "root", "rules": ["ANY /{path+}"]}],
}
{
    "base": "https://{sub+}.com/{tenant}",
    "permissions": [{"name": "restricted", "rules": ["GET /delete"]}],
}
```

Set the policy to allow `root` and deny `restricted`, then match:

```python
url = f"https://{host}/admin/delete"
method = "GET"
```

Both bases and both rules match. The parameterized base consumes `admin` as `tenant`, leaving `/delete` for the denied rule.

The current scores are:

```text
exact root base:
  authority = 100 * 1001 + 1003 = 101103
  total     = 0 * 1000000 + 101103 * 100 + 1 = 10110301

path-specific wildcard base:
  path      = 10
  authority = 1 + 1003 = 1004
  total     = 10 * 1000000 + 1004 * 100 = 10100400
```

The broad root base wins by `9,901`, so the denied path-specific API is discarded and the request is allowed.

## Security impact

For an overlapping connector firewall covering a valid multi-label provider hostname, a sandbox request can reach an operation that the active policy explicitly denies. The winning broad API can proceed through the normal credential-injection and upstream-dispatch path because the deny candidate was removed before authorization reduction.

## Acceptance criteria

- Specificity ordering cannot carry between path and authority dimensions for any accepted base or runtime authority.
- A regression test using the valid 100-label hostname above returns `FirewallBlock` for the `restricted` permission in both API declaration orders.
- Existing static-host-over-wildcard-host and path-base-over-root-base precedence behavior remains intact.

seven332:
## Summary

- represent compiled firewall base specificity as a lexicographic key ordered by path, authority, and static-base bonus
- prevent long accepted hostnames from carrying authority scores into the path-precedence dimension
- cover the fail-closed result in both API declaration orders through the production compiled matcher

## Scope

- keeps hostname validation, matcher selection flow, permission reduction, and existing component scores unchanged
- does not change APIs, persisted data, protocols, dependencies, or deployment compatibility

## Validation

- `uv run --no-sync python -m pytest tests/test_compiled_firewall_base_specificity_precedence.py` (23 passed)
- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (6629 passed)
- lefthook pre-commit and commitlint

Closes #29297

### DIFF
--- a/crates/runner/mitm-addon/src/firewall_matching/base_url.py
+++ b/crates/runner/mitm-addon/src/firewall_matching/base_url.py
@@ -41,8 +41,6 @@ _FORBIDDEN_AUTHORITY_HOST_CHARS = frozenset("#%/<>?@[\\]^|[]")
 _FORBIDDEN_RUNTIME_AUTHORITY_HOST_CHARS = _FORBIDDEN_AUTHORITY_HOST_CHARS | frozenset("{}")
 _PERCENT_DECODED_AUTHORITY_SYNTAX_CHARS = frozenset("{}*.\u3002\uff0e\uff61")
 _VALID_BASE_SCHEMES = frozenset(("http", "https"))
-_BASE_PATH_SCORE_MULTIPLIER = 1_000_000
-_BASE_AUTHORITY_SCORE_MULTIPLIER = 100
 _BASE_LITERAL_SEGMENT_SCORE = 1_000
 _BASE_MIXED_PARAM_SEGMENT_SCORE = 100
 _BASE_PLAIN_PARAM_SEGMENT_SCORE = 10
@@ -67,11 +65,17 @@ class _BaseUrlParts(NamedTuple):
         return f"{self.hostname}:{self.port}"
 
 
+class _BaseSpecificity(NamedTuple):
+    path: int
+    authority: int
+    static_bonus: int
+
+
 class _CompiledBase(NamedTuple):
     raw: str
     parts: _BaseUrlParts
     has_params: bool
-    specificity: int
+    specificity: _BaseSpecificity
     has_query_or_fragment: bool
     raw_syntax_malformed: bool
     param_parse_malformed: bool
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #29352

**Reviewed head:** `c5b5e37f3419ed6b20fe705e38b57cee68955ad5`

### Summary

The change replaces an unsafe weighted scalar with an explicit lexicographic base-specificity key, updates both matcher comparison consumers, and adds a production-entry-point regression for both API declaration orders. The implementation preserves existing component scores, hostname acceptance, and policy-reduction flow.

### Findings

No findings.

### Testing

Focused precedence coverage passed with 23 tests. The locked mitm-addon environment, Ruff formatting and linting, BasedPyright, and the full 6629-test addon suite all passed locally. Turbo and Rust checks were not run because no cross-language inputs or contracts changed; remote CI is evaluated separately.

### Verdict

LGTM

PR #29354, opened 2026-08-25
fix(composer): keep the flat note chrome stable while typing

bingjiez666:
## Summary

First dogfood regression on the flat quote block (#29137): quote → type one uncommitted pinyin character → press Backspace → **the caret disappears**.

## Cause

The chrome widget's decoration key encoded the note's empty/filled state:

```
feedback-chrome-<id>-<empty|filled>-<lang>
```

So the widget was torn down and recreated at exactly two moments: when the first character arrived (empty→filled) and when the last one was deleted (filled→empty). Both are IME composition boundaries, and the rebuild replaces a `contenteditable="false"` element immediately before the caret's paragraph. Bench-verified: the widget rebuilds twice across a type-one-delete-one round trip. Chromium's selection survives the churn; WebKit loses the caret when the churn lands on a cancelled composition — the same "DOM churn adjacent to a live IME composition" hazard class as #27787 itself.

## Fix

- **Stable widget key** (`feedback-chrome-<id>-<lang>`): the chip element keeps its identity for the lifetime of the quote; typing never rebuilds it.
- **Placeholder leaves the widget**: the empty note's paragraph gets a node decoration carrying the localized text in `data-placeholder`, rendered through `::before` (the pattern Tiptap's official Placeholder extension uses). The placeholder appearing/disappearing is now a class+attribute flip on the paragraph — no DOM nodes are inserted or removed around the caret, and the fixed 42px offset hack is replaced by in-flow positioning.

## Verification

- `chat-inline-feedback.test.tsx` — 35 passed; new regression test asserts the chrome element keeps its **identity** (same DOM node) while the note goes empty → filled → empty
- negative control: reintroducing the empty state into the widget key makes the new test fail
- placeholder assertions updated to the attribute-driven form (`data-placeholder` present on the empty note's paragraph, absent once text exists)
- `check-types`, `lint`, `knip` — clean
- device verification: quote → type one pinyin character without committing → Backspace — the caret must stay visible; then compose normally and send

Related to #27787.

### DIFF
--- a/turbo/apps/platform/src/signals/okou-page/tiptap-workflow-composer.ts
+++ b/turbo/apps/platform/src/signals/okou-page/tiptap-workflow-composer.ts
@@ -721,7 +721,6 @@ function createFlatFeedbackItemNodeView(
 
 function buildFeedbackItemChrome(
   quote: string,
-  showPlaceholder: boolean,
   onRemove: () => void,
 ): HTMLElement {
   const { quoteDom, quoteText, removeButton } = createFeedbackQuoteChip();
@@ -735,22 +734,14 @@ function buildFeedbackItemChrome(
     event.preventDefault();
   });
   removeButton.addEventListener("click", onRemove);
-  if (showPlaceholder) {
-    quoteDom.classList.add("relative");
-    const placeholder = document.createElement("span");
-    // The chip row is 32px plus the item's 6px gap; the first paragraph adds
-    // 4px of top padding, so the placeholder sits at 42px over its text.
-    placeholder.className =
-      "pointer-events-none absolute left-1 top-[42px] w-max " +
-      "text-[0.9375rem] leading-snug text-muted-foreground/40";
-    placeholder.setAttribute("aria-hidden", "true");
-    placeholder.textContent = i18n.t(($) => {
-      return $.chat.feedback.placeholder;
-    });
-    quoteDom.append(placeholder);
-  }
   return quoteDom;
 }
+
+// Rendered through ::before so an empty note never gains or loses real DOM
+// nodes: the placeholder text lives in an attribute and appears via CSS.
+const FLAT_PLACEHOLDER_PARAGRAPH_CLASS =
+  "before:pointer-events-none before:float-left before:h-0 " +
+  "before:text-muted-foreground/40 before:content-[attr(data-placeholder)]";
 
 function buildFeedbackChromeDecorations(
   doc: ProseMirrorNode,
@@ -766,19 +757,20 @@ function buildFeedbackChromeDecorations(
       continue;
     }
     const { feedbackId, quote } = feedbackItemNodeAttributes(child);
-    const showPlaceholder = nodeText(child).length === 0;
     decorations.push(
       Decoration.widget(
         childPosition + 1,
         () => {
-          return buildFeedbackItemChrome(quote, showPlaceholder, () => {
+          return buildFeedbackItemChrome(quote, () => {
             runtime.removeFeedback(feedbackId);
           });
         },
         {
-          // The key carries everything the DOM is built from, so the widget
-          // is only recreated when its content actually changes.
-          key: `feedback-chrome-${feedbackId}-${showPlaceholder ? "empty" : "filled"}-${i18n.language}`,
+          // The key must stay stable while the user types: recreating the
+          // widget replaces a contenteditable=false element right next to the
+          // caret, and WebKit loses the caret when that happens at an IME
+          // composition boundary (typing or backspacing the first character).
+          key: `feedback-chrome-${feedbackId}-${i18n.language}`,
           side: -2,
           ignoreSelection: true,
           stopEvent: () => {
@@ -787,6 +779,20 @@ function buildFeedbackChromeDecorations(
         },
       ),
     );
+    if (nodeText(child).length === 0) {
+      decorations.push(
+        Decoration.node(
+          childPosition + 1,
+          childPosition + 1 + child.child(0).nodeSize,
+          {
+            class: FLAT_PLACEHOLDER_PARAGRAPH_CLASS,
+            "data-placeholder": i18n.t(($) => {
+              return $.chat.feedback.placeholder;
+            }),
+          },
+        ),
+      );
+    }
   }
   return DecorationSet.create(doc, decorations);
 }
--- a/turbo/apps/platform/src/views/okou-page/__tests__/chat-inline-feedback.test.tsx
+++ b/turbo/apps/platform/src/views/okou-page/__tests__/chat-inline-feedback.test.tsx
@@ -1557,9 +1557,13 @@ describe("chat inline feedback", () => {
     // The flat block is its own content element: no nested note wrapper.
     expect(note.dataset.feedbackItem).toBe("");
     expect(note.querySelector("[data-feedback-note]")).toBeNull();
-    // The chrome widget carries the quote chip and the empty-note placeholder.
+    // The chrome widget carries the quote chip; the empty-note placeholder is
+    // an attribute the paragraph renders through ::before, so typing never
+    // inserts or removes DOM around the caret.
     expect(note).toHaveTextContent(assistantReply);
-    expect(note).toHaveTextContent("What should change about this?");
+    expect(
+      note.querySelector<HTMLElement>(":scope > p")?.dataset.placeholder,
+    ).toBe("What should change about this?");
 
     await user.click(note);
     pastePlainText(note, "Make the dates explicit");
@@ -1567,7 +1571,9 @@ describe("chat inline feedback", () => {
       expect(note).toHaveTextContent("Make the dates explicit");
     });
     await waitFor(() => {
-      expect(note).not.toHaveTextContent("What should change about this?");
+      expect(
+        note.querySelector<HTMLElement>(":scope > p")?.dataset.placeholder,
+      ).toBeUndefined();
     });
 
     await user.click(screen.getByLabelText("Send"));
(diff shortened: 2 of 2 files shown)


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: e70a9d9cb5c23ccd0a33a00a03d17cd48eee6c18
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Removes the empty/filled segment from the flat quote block's chrome widget decoration key so the `contenteditable="false"` chip next to the caret is no longer torn down and rebuilt at the two IME composition boundaries (first character typed, last character deleted), which is what dropped the caret on WebKit. The empty-note placeholder moves out of the widget into a `Decoration.node` on the note's first paragraph carrying `data-placeholder`, rendered through `::before`, so the empty↔filled transition is now a class + attribute flip instead of DOM insertion/removal beside the caret.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness, logic, edge cases, no API misuse | Pass | `Decoration.node(childPosition + 1, childPosition + 1 + child.child(0).nodeSize, …)` spans exactly the first paragraph, matching ProseMirror's requirement that a node decoration cover a whole node; `child.child(0)` is total because the `feedbackItem` spec is `content: "paragraph+"` (`tiptap-workflow-composer.ts:1803`). Widget position/`side: -2` unchanged, so the chip still renders before the paragraph. |
| `REVIEW.md` | Security (OWASP, secrets, auth on new endpoints) | N/A | Client-only ProseMirror decoration change; no endpoint, credential, or user-supplied HTML sink. The placeholder text is set via `setAttribute`/CSS `attr()`, not `innerHTML`. |
| `REVIEW.md` | Type safety — no `any`, no overly broad types | Pass | No `any` or casts introduced; `buildFeedbackItemChrome` loses a parameter and keeps its concrete signature. |
| `REVIEW.md` | Style and maintainability — no oversized functions, no duplicated logic, comments explain WHY | Pass | `buildFeedbackChromeDecorations` stays small; both new comments state the reason (WebKit caret loss at composition boundaries; placeholder via attribute so no nodes move) rather than restating the code. |
| `REVIEW.md` | Feature regression guard — changed existing behavior needs regression coverage and containment | Pass | The whole flat block is contained behind `FeatureSwitchKey.ComposerFlatFeedbackNote` (`feature-switch.ts:378`, `enabled: false`, `enabledOrgIdHashes: STAFF_ORG_ID_HASHES`). The pre-existing placeholder assertions in the flat-block test were updated to the new attribute form rather than deleted, and a new test covers the transition that regressed. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, contract, Zod response, or `jsonb().$type<>()` change. |
| `docs/bad-smell.md` | Zero `any`, zero lint/type suppressions | Pass | No `any`, no `@ts-*`/`eslint-disable`/`prettier-ignore` in the diff. CI `lint-types`, `lint-eslint`, `lint-knip`, `lint-format` all green. |
| `docs/bad-smell.md` | No dynamic `import()`, no hardcoded URLs or env-specific values | Pass | No imports added; the only new literal is a Tailwind class string constant. |
| `docs/bad-smell.md` | Fail fast — no silent fallback/recovery logic | Pass | No `??`/`||` chain or defensive branch added. `nodeText(child).length === 0` is the actual product condition (empty note), not a guard against an impossible state. |
| `docs/fallback.md` | Default no fallback; every fallback introduced must be declared in the PR summary | Pass | The diff introduces no fallback or compatibility branch, so no `Fallbacks` section is required. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | The new test asserts a live positive property (the chrome element keeps its identity across empty→filled→empty); the changed assertions describe current placeholder rendering, not the absence of deleted code. |
| `docs/fallback.md` | A non-GA feature-switched feature needs no compatibility code | Pass | `ComposerFlatFeedbackNote` is staff-only and off by default, so no dual-read, migration, or client-skew handling is owed — and none was added. |
| `docs/testing.md` | `fix:` commits require a regression test | Pass | `chat-inline-feedback.test.tsx:1587` adds "keeps the chrome widget's identity while the note empties and refills", pinning the exact property the fix restores; the PR reports a negative control (reintroducing the empty state into the key fails it). |
| `docs/testing.md` | Mock at the system boundary only — no `vi.mock()` on relative paths, no direct `fetch` stubbing, no fake timers | Pass | The new test adds no mocks at all and reuses `mockChatLifecycle` + `context.mocks`; no `vi.useFakeTimers`, no `vi.stubGlobal("fetch", …)`. |
| `docs/testing.md` | Test behavior, not mock call patterns | Pass | Assertions are on rendered DOM state (`note.firstElementChild`, the paragraph's `data-placeholder`, `toHaveTextContent`), never `expect(mock).toHaveBeenCalled()`. |
| `docs/testing/app-testing.md` | Platform tests are page-level and enter through `detachedSetupPage`, with mocks configured via `context.mocks` before setup | Pass | The new test goes through `setupFlatFeedbackNoteThread`, which calls `detachedSetupPage({ context, path: "/chats/…", featureSwitches: { ComposerFlatFeedbackNote: true } })` without awaiting it, then waits for the composer. No component is rendered directly and the MSW `server` is not imported. |
| `docs/testing/patterns.md` | Default to integration tests; no unit tests for helpers/parsers/formatters | Pass | The fix lives in `buildFeedbackChromeDecorations`, and it is exercised through the page rather than by calling the decoration builder directly. |
| `docs/testing/testing-external-behavior.md` | Construct the case through page interactions; observe from a user-visible surface; do not assert on CSS classes or internal state | Pass | The case is built the user's way (select text → Quote → click note → type → Backspace). The two non-textual assertions are the minimum the boundary allows here: `::before` content is not computable in jsdom, so the placeholder is read from the attribute that drives it, and the caret-loss hazard is a WebKit DOM-churn behavior with no jsdom-visible surface, so element identity is the only observable proxy. Neither assertion pins a CSS class name or internal store state. |
| `docs/testing/anti-patterns.md` | AP-1/AP-2/AP-4/AP-5 — no mock-call assertions, no fetch stubbing, no internal mocking, no fake timers | Pass | None present in the added test. |
| `docs/testing/anti-patterns.md` | AP-7 — do not test implementation details | Pass | Closest call in the diff, and it clears: the assertions target the DOM the user's browser actually renders, not a class name, keyboard handler, or component state. Documented in the test's own comment. |
| `docs/effect.md` | Choose owner and trigger before API; keep React render pure | Pass | All changed code is inside a ProseMirror `Plugin`'s `decorations(state)` prop; nothing runs in React render, and no command, Store write, or global mutation was added. |
| `docs/effect.md` | Use `onRef`/setup commands for DOM and lifecycle work rather than incidental hooks | Pass | The widget's DOM lifetime is owned by ProseMirror's decoration machinery, which is the correct owner for a per-document chip; the change strengthens that ownership by keeping the element stable for the quote's lifetime. |
| `docs/cache.md` | RC-2 — no render-time signal allocation | Pass | No `state`/`computed`/`command` is created in the diff; `FLAT_PLACEHOLDER_PARAGRAPH_CLASS` is a module-level string constant. |
| `docs/cache.md` | RC-3 — no unbounded-lifetime caches | Pass | No `Map`/`Set`/`WeakMap`/`globalThis` registry added. The decoration key is a per-render string, not a retention key. |
| `docs/cache.md` | RC-8 — symmetric resource lifecycles | Pass | The `mousedown`/`click` listeners live on the chip element and die with it. Stabilizing the key strictly reduces churn: previously each empty↔filled transition discarded one listener-bearing element and built another. |
| `.claude/skills/ccstate/SKILL.md` | `detach(…, Reason.DomCallback)` for async DOM handlers, never `void`; no manual cache over `computed`; no function stored via bare `set()` | N/A | No ccstate signal, command, or async DOM handler is touched; `onRemove` is a synchronous callback invoked by ProseMirror's widget DOM, and the `signals/` directory rule against `detach` in commands is not engaged. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Notes (non-blocking)

- `tiptap-workflow-composer.ts:735` — `buildFeedbackItemChrome` captures `quote` in the widget's render closure, and with the now-stable key ProseMirror will not re-invoke it. This is unchanged in effect (the old key did not carry `quote` either) and `feedbackId` is minted per quote action, so a stable id never acquires a different quote today. Worth knowing if quote text ever becomes editable.
- `tiptap-workflow-composer.ts:1996` — the comment "The feedback chrome widgets carry their language in the decoration key" still holds, and the empty transaction dispatched by `refreshWorkflowComposerLocalization` also rebuilds the new node decoration, so `data-placeholder` follows a language switch. Verified, no change needed.

### Testing

- **Coverage: Adequate.** `fix:` commit with a matching regression test that fails without the fix (author-reported negative control: putting the empty state back in the key breaks it). The two updated assertions in the existing flat-block test track the placeholder's new rendering mechanism instead of deleting coverage.
- **Conventions: Compliant.** Page-level entry via `detachedSetupPage`, feature switch supplied through the harness, no mocks introduced, no fake timers, no internal-path `vi.mock()`, no mock-call assertions.
- **CI on `e70a9d9`:** all checks green — `test-app (1–8)`, `test-api (1–8)`, `test-cli`, `test-other`, `lint-types`, `lint-types-apps`, `lint-eslint`, `lint-format`, `lint-knip`, `bench-app`, CodeQL/Semgrep/Gitleaks, `codecov/patch`.
- **Independent check of the new rendering mechanism:** compiled `before:pointer-events-none before:float-left before:h-0 before:text-muted-foreground/40 before:content-[attr(data-placeholder)]` with Tailwind v4.3.3 (the version this repo pins) against a source file containing the exact concatenated literal. It emits `.before\:content-\[attr\(data-placeholder\)\]::before { --tw-content: attr(data-placeholder); content: var(--tw-content) }` plus the `float: left` and `pointer-events: none` rules, so the class survives v4 source scanning and the placeholder will actually paint — this was the one part of the fix that unit tests cannot observe, since jsdom does not compute `::before` content.
- **Still owed by the author (as the PR states):** on-device WebKit verification — quote → type one uncommitted pinyin character → Backspace → the caret must stay visible, then compose and send normally.

PR #26910, opened 2026-08-13
feat(app): add video model selection to the model picker

bingjiez666:
Part of #26765 (PR6 of 8). Builds on the server work already on `main`, including #26807, #26831, #26841, and #26884.

## What this adds

A **Manage more models** row under the run model list in the composer's model picker. Clicking it swaps the popover to a second panel listing the video models; picking one writes that concrete model to the current thread and closes the popover.

- **Panel contents** — all 8 `public: true` models from `VIDEO_MODEL_CONFIGS`. No price-tier, BYOK, or Pro badges, and no plan/org availability filter.
- **Visible selection** — an existing thread pin is checked. When the thread has no selection yet, the picker checks the model that currently resolves from the member preference or `DEFAULT_VIDEO_MODEL`; it does not show a separate default or auto row.
- **Write behavior** — clicking any model, including the currently resolved fallback model, writes that concrete model ID to the current chat thread.
- **Write path** — `patchChatThreadVideoModel$` registers the optimistic `video_model_updated` event before the HTTP request, matching `patchChatThreadComputerUseHost$`.
- **Feature switch** — new `FeatureSwitchKey.VideoModelSelection`, `enabled: false` + `STAFF_ORG_ID_HASHES`, maintainer `bingjie@vm0.ai`. It does not reuse `videoTemplateOptions`.

## Implementation notes

- The panel rows are plain `<button>`s, not `SelectItem`s, so they do not join the run-model `Select` value space. While the panel is open, the selected run model remains as a hidden measurable item.
- Panel open/close state lives in the composer's model UI signal group. Closing the picker resets it to the run-model list.
- `ComposerSignals.videoModel` is optional. The new-chat composer does not render this row in PR6; that surface remains PR7.
- `PUBLIC_VIDEO_MODELS` moved from `video-template-options-popover.tsx` into the core catalog because both views consume it.

## Scope

Thread-level selection only. New-chat / landing-page selection, writing back to the personal default, and create-time `videoModel?` remain PR7.

## Ordering

**This must merge before PR5**, which removes the old `--model` entry point. This PR is not auto-merged and awaits manual review.

## Verification

- Prettier check passed for all PR files.
- Full Turbo lint passed with single-package concurrency; Knip passed.
- Type checks passed for all non-app/api workspaces, app, and api.
- `chat-composer-model-selection-and-providers.test.tsx` — 56 passed.
- `chat-composer-template-picker.test.tsx` — 31 passed.
[...]

PR #29422, opened 2026-08-26
refactor: remove the video model selection feature switch

bingjiez666:
## Terminal state

`fully-rolled-out`. On `main` the registry entry was `enabled: true` with no
`enabledOrgIdHashes`, so every org already ran the switched-on path. The
switched-on behaviour is kept permanently and every switched-off fallback is
deleted.

Introduced in `136547f33` — *feat(app): add video model selection to the model
picker* (#26910).

## Kept

- The composer model picker's Video category, its thread pin, and the video
  options chip.
- Creation-time pinning of the member's video default onto a new chat thread.
- The run's video model snapshot as the default for `POST /api/video-io/generate`.
- `runOptions.video` being honoured on a normal send.

## Deleted

- `FeatureSwitchKey.VideoModelSelection` and its registry entry.
- `videoModelSelectionEnabled$` (platform) and `videoModelSelectionEnabled`
  on `NormalSendFeatureSwitches` (api).
- The guard in `loadDefaultRunVideoModel`, which no longer needs `orgId`/`userId`
  or a feature-switch context load — one DB round trip fewer per video generation.
- The `!videoEnabled && !imageEnabled` early return in
  `resolveNewChatThreadMediaModels`; the video pin is now unconditional.
- The `videoModelEnabled` prop/branches in `ComposerModelPickerSlot`,
  `ComposerExistingMediaModelPickerSlot`, `ComposerTemporaryModelNoticeSlot`
  and `ComposerVideoOptionsChip`.

## Tests

Deleted (covered only the discarded switched-off behaviour):

- `keeps the request model when video model selection is disabled` — the
  neighbouring `... when the run video model snapshot is null` still covers
  "no pin means the request model wins".
- `leaves new-chat video defaults disabled while the feature switch is off`
- `hides the video model row while the feature switch is off`
- `leaves a thread unpinned while video model selection is disabled`
- `leaves media models unpinned while neither picker is enabled` — its
  remaining content duplicated `pins each media model whose picker is enabled`,
  which is retitled `pins the video model while the image picker is disabled`.

Rewritten to assert the permanent behaviour instead of deleting:

- `offers to make a temporary new-chat model choice the default below the composer`
  had forced both media switches off. With the media panel always present the
  picker renders the category switch instead of a `Models` section label
  (`showModelsLabel={!mediaModelPanel}`), so the anchor assertion now checks the
  Video category tab.
- `edits thread override without loading user default model selection` →
  `... without offering the new-chat default actions`. The always-on video
  picker resolves `effectiveVideoModel$`, which reads the member preference, so
  the "preference never fetched" assertion no longer describes any shipped
  path. The surviving assertions (no default row, no temporary notice) are the
  behaviour the test was protecting.

The `videoModelSelectionActor` bdd helper no longer takes `selectionEnabled`.

## Second-pass keyword sweep

`VideoModelSelection`, `videoModelSelection`, `videoModelSelectionEnabled`,
`videoModelEnabled`, `video-model-selection`, plus the symbols added by
#26910. Remaining hits are the feature's own permanent naming and are now
fully decoupled from the switch: `createVideoModelSelection`,
`setVideoModelSelection$`, `chatPageVideoModelSelection$`,
`setChatPageVideoModelSelection$`, `resetChatPageVideoModelSelection$`.

## Verification

| Check | Result |
| --- | --- |
| `pnpm check-types` | pass (14/14) |
| `pnpm lint` | pass, 0 errors (438 pre-existing warnings) |
| `pnpm knip` | exit 0, no unused files/exports/dependencies — baseline and after are both clean, so this cleanup left no orphans |
| `prettier --write` on changed files | clean |
| `chat-composer-model-selection-and-providers.test.tsx` | 79/79 pass |
| `chat-composer-template-picker.test.tsx`, `sidebar.test.tsx`, `core/feature-switch.test.ts` | 150/150 pass |

Delegated to the PR pipeline: the `apps/api` suites. They need a Postgres on
`:5432` that this environment does not provide (`connect ECONNREFUSED
::1:5432` / `127.0.0.1:5432`, 12 tests skipped) — a pre-existing environment
limit unrelated to this diff.

## Note

Opened alongside the `imageModelSelection` cleanup. Both branch from the same
`main` and both touch `chat-composer.tsx`, `chat-thread-media-model.service.ts`,
`feature-switch.ts` and the shared composer test, so whichever merges second
will need a rebase.

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/agentphone.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/agentphone.bdd.test.ts
@@ -9,7 +9,6 @@ import { createHash, randomUUID } from "node:crypto";
 import { HttpResponse, http } from "msw";
 import { describe, expect, it } from "vitest";
 
-import { FeatureSwitchKey } from "@okouai/core/feature-switch-key";
 import { DEFAULT_VIDEO_MODEL } from "@okouai/core/video-model-catalog";
 
 import { testContext } from "../../../__tests__/test-context";
@@ -36,7 +35,6 @@ import {
   type AgentPhoneSendCapture,
 } from "./helpers/api-bdd-agentphone";
 import { createChatFilesBddApi } from "./helpers/api-bdd-chat-files";
-import { updateFeatureSwitchesForUser } from "./helpers/feature-switches";
 import { createBddIntegrationApi } from "./helpers/api-bdd-integrations";
 import { createRunsApi } from "./helpers/api-bdd-runs";
 import { createStoragesBddApi } from "./helpers/api-bdd-storages";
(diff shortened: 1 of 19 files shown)


bingjiez666:
Superseded by #29430, which removes both media model selection switches in one PR. Splitting them left the shared media-pin helper split and its feature-switch threading in place, since neither half alone makes them dead.

PR #27710, opened 2026-08-17
feat(core): add image model catalog and selection switch

bingjiez666:
## Summary

- define the nine prompt-based built-in image models as canonical contract IDs with a schema and type guard
- add a shared core catalog for picker order, labels, CLI aliases, and the global `gpt-image-1` default
- make the API's prompt-based generation configs exhaustively match that catalog while keeping `birefnet` and `clarity-upscaler` API-local
- register `ImageModelSelection` globally off, including for staff, with rollout-state coverage

## Behavior

This is PR 1 of the stack in #27688. It establishes shared types and the disabled rollout switch only. It does not add persistence, UI, CLI environment behavior, or image-generation precedence, and existing image API behavior remains unchanged.

After this merges, the DB/event protocol, dormant CLI/template support, and composer media-panel refactor can proceed in parallel.

## Verification

- changed-file Prettier check
- `@okouai/api-contracts` and `@okouai/core` lint and type checks
- changed-file API Oxlint, type-aware Oxlint, and ESLint
- API gateway and core TypeScript checks
- targeted Knip production scan for `api`, `@okouai/api-contracts`, and `@okouai/core`
- `feature-switch.test.ts` — 24 passed
- `image-io-generate.test.ts` — 20 passed
- `git diff --check`

Refs #27688

PR #29423, opened 2026-08-26
refactor: remove the image model selection feature switch

bingjiez666:
## Terminal state

`fully-rolled-out`. On `main` the registry entry was `enabled: true` with no
`enabledOrgIdHashes`, so every org already ran the switched-on path. The
switched-on behaviour is kept permanently and every switched-off fallback is
deleted.

Introduced in `7087c6f37` — *feat(core): add image model catalog and selection
switch* (#27710).

## Kept

- The composer model picker's Image category and its thread pin.
- Creation-time pinning of the member's image default onto a new chat thread.
- `resolveImageModelForRun` resolving thread pin → member default →
  `DEFAULT_IMAGE_MODEL`, and the resulting run snapshot reaching the sandbox as
  the default built-in image model.
- The run's image model snapshot as the default for `POST /api/image-io/generate`.

## Deleted

- `FeatureSwitchKey.ImageModelSelection` and its registry entry.
- `imageModelSelectionEnabled$` (platform).
- The guard in `loadRunImageModelDefault`, which no longer loads a
  feature-switch context — one DB round trip fewer per image generation.
- The guard in `resolveImageModelForRun`. It can no longer return `null`, so its
  return type narrows from `ImageModel | null` to `ImageModel` and it drops the
  `featureSwitchContext` argument, matching the already-ungated
  `resolveVideoModelForRun` next to it.
- The `!videoEnabled && !imageEnabled` early return in
  `resolveNewChatThreadMediaModels`; the image pin is now unconditional.
- The `imageModelEnabled` prop/branches in `ComposerModelPickerSlot`,
  `ComposerExistingMediaModelPickerSlot` and `ComposerTemporaryModelNoticeSlot`.

## Tests

Deleted (covered only the discarded switched-off behaviour):

- `omits image UI, preference writes, and thread pins when disabled`
- `keeps the run snapshot null while image model selection is disabled`
- `leaves media models unpinned while neither picker is enabled` — its
  remaining content duplicated `pins each media model whose picker is enabled`,
  which is retitled `pins the image model while the video picker is disabled`.
- The `disabled switch` case in `keeps the global default for disabled, null,
  old, and session paths` (retitled `... for null, old, and session paths`).
  That case relied on the switch suppressing a valid run snapshot; without it
  the snapshot is honoured, so the expected `gptCalls` drops from 4 to 3.

Rewritten to assert the permanent behaviour instead of deleting:

- `offers to make a temporary new-chat model choice the default below the composer`
  had forced the media switches off. With the media panel always present the
  picker renders the category switch instead of a `Models` section label
  (`showModelsLabel={!mediaModelPanel}`), so the anchor assertion now checks the
  Image category tab.
- `edits thread override without loading user default model selection` →
  `... without offering the new-chat default actions`. The always-on image
  picker resolves the member preference, so the "preference never fetched"
  assertion no longer describes any shipped path. The surviving assertions
  (no default row, no temporary notice) are the behaviour the test was
  protecting.
- `keeps the agent chat model picker open while feature switches hydrate` used
  the Image tab appearing as its proof of a post-hydration re-render. Category
  tabs are no longer switch-driven, so the test now asserts the listbox and its
  options survive hydration, with the deferred `featureSwitches.get` still
  driving the in-flight window. Deliberately not retargeted at another switch —
  that would have coupled it to the sibling `videoModelSelection` cleanup.
- `switches category inside one popover opened from a single trigger` and
  `pins the visible fallback model to the current thread` asserted
  `categoryTab("Image")` was absent; the Image category is now always offered,
  so both assert its presence.

The `imageModelSnapshotActor` bdd helper and the `seedImageRun` fixture helper
no longer take `selectionEnabled` / `imageModelSelectionEnabled`.

## Second-pass keyword sweep

`ImageModelSelection`, `imageModelSelection`, `imageModelSelectionEnabled`,
`imageModelEnabled`, `image-model-selection`, plus the symbols added by
#27710. Remaining hits are the feature's own permanent naming and are now
fully decoupled from the switch: `createImageModelSelection`,
`setImageModelSelection$`, `chatPageImageModelSelection$`,
`setChatPageImageModelSelection$`, `resetChatPageImageModelSelection$`.

## Verification

| Check | Result |
| --- | --- |
| `pnpm check-types` | pass (14/14) |
| `pnpm lint` | pass, 0 errors (441 pre-existing warnings) |
| `pnpm knip` | exit 0, no unused files/exports/dependencies — baseline and after are both clean, so this cleanup left no orphans |
| `prettier --check` on changed files | clean |
| `chat-composer-model-selection-and-providers.test.tsx` | 80/80 pass |
| `sidebar.test.tsx`, `core/feature-switch.test.ts` | 105/105 pass |

Delegated to the PR pipeline: the `apps/api` suites. They need a Postgres on
`:5432` that this environment does not provide (`connect ECONNREFUSED
::1:5432` / `127.0.0.1:5432`) — a pre-existing environment limit unrelated to
this diff.

## Note

Opened alongside the `videoModelSelection` cleanup. Both branch from the same
`main` and both touch `chat-composer.tsx`, `chat-thread-media-model.service.ts`,
`feature-switch.ts` and the shared composer test, so whichever merges second
will need a rebase.

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts
@@ -13861,9 +13861,8 @@ describe("CHAT-02: run video model snapshot", () => {
   }, 90_000);
 });
 
-async function imageModelSnapshotActor(args: {
-  readonly selectionEnabled: boolean;
-}): Promise<{
+/** The image snapshot is org-scoped, so these tests resolve the org too. */
+async function imageModelSnapshotActor(): Promise<{
   readonly actor: ApiTestUser;
   readonly agentId: string;
   readonly orgId: string;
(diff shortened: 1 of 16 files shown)


bingjiez666:
Superseded by #29430, which removes both media model selection switches in one PR. Splitting them left the shared media-pin helper split and its feature-switch threading in place, since neither half alone makes them dead.

PR #29441, opened 2026-08-26
fix: keep the deck import message to the request itself

bingjiez666:
## Problem

Importing a deck through the template picker sent this on the user's behalf:

> Analyse this deck and save its visual language as a reusable presentation template.
>
> Turning an uploaded deck (.pptx, .ppt, or .pdf) into a reusable presentation template follows the reverse-template guide, which is not mounted as a skill. Pull it first with `okou resource pull skill:presentation-reverse-template --dir ./generated/resources`, read `./generated/resources/reverse-template/SKILL.md`, and follow it exactly, including its page-rendering and publish steps.

The user lands in a thread whose first message is two thirds plumbing addressed to the run, not to them.

## Change

The message is now the request alone. The guide pointer stays exactly where it already was — `buildAgentToolsPrompt`, gated on `FeatureSwitchKey.PresentationTemplates`, the same switch that decides whether the import entry point is offered at all. Nothing is lost: a run that can receive an imported deck is a run whose system prompt already names the pull command and the SKILL.md path.

`presentationTemplateSkillInstruction()` is unchanged and now has a single caller.

## Coverage

- `chat-composer-template-picker.test.tsx` — the import test now asserts the sent prompt equals the sentence, so a future append is a failure rather than a passing `toContain`.
- `run-lifecycle.bdd.test.ts` (unchanged) already pins that the pull command and SKILL.md path reach `appendSystemPrompt` when the switch is on, and stay out when it is off. That is the test that keeps this change non-lossy.

### DIFF
--- a/turbo/apps/platform/src/signals/okou-page/presentation-template-import.ts
+++ b/turbo/apps/platform/src/signals/okou-page/presentation-template-import.ts
@@ -1,6 +1,5 @@
 import { command, computed } from "ccstate";
 import { FeatureSwitchKey } from "@okouai/core/feature-switch-key";
-import { presentationTemplateSkillInstruction } from "@okouai/core/presentation-template-skill";
 
 import { featureSwitch$ } from "../external/feature-switch.ts";
 import type { ComposerSignals } from "./composer-signals.ts";
@@ -19,17 +18,17 @@ export const PRESENTATION_TEMPLATE_IMPORT_ACCEPT = ".pptx,.ppt,.pdf";
 /**
  * The message the deck is sent with.
  *
- * Opens as a plain sentence on purpose: importing a template is not a special
+ * One plain sentence on purpose: importing a template is not a special
  * protocol, it is a chat message with a file attached, and the user should be
- * able to read what was asked on their behalf in the thread they land in. The
- * pointer to the guide follows, because the guide is not a mounted skill the
- * run would otherwise find.
+ * able to read what was asked on their behalf in the thread they land in.
+ *
+ * How to reach the guide is deliberately absent. The agent tools prompt
+ * already carries it behind the same feature switch that offers this import,
+ * so repeating it here only spends the user's own message on instructions
+ * addressed to the run.
  */
 function presentationTemplateImportPrompt(): string {
-  return [
-    "Analyse this deck and save its visual language as a reusable presentation template.",
-    presentationTemplateSkillInstruction(),
-  ].join("\n\n");
+  return "Analyse this deck and save its visual language as a reusable presentation template.";
 }
 
 export const presentationTemplateImportEnabled$ = computed((get) => {
--- a/turbo/apps/platform/src/views/okou-page/__tests__/chat-composer-template-picker.test.tsx
+++ b/turbo/apps/platform/src/views/okou-page/__tests__/chat-composer-template-picker.test.tsx
@@ -4610,14 +4610,11 @@ describe("chat composer templates", () => {
       expect(screen.queryByRole("dialog")).not.toBeInTheDocument();
     });
     await waitFor(() => {
-      expect(sentPrompt).toContain("reusable presentation template");
-      // The guide is not a mounted skill, so the message has to carry the
-      // private resource pull and its extracted entrypoint.
-      expect(sentPrompt).toContain(
-        "okou resource pull skill:presentation-reverse-template --dir ./generated/resources",
-      );
-      expect(sentPrompt).toContain(
-        "./generated/resources/reverse-template/SKILL.md",
+      // What the user sees in their own thread is the request they would have
+      // typed. How the run reaches the guide is carried by the agent tools
+      // prompt, so none of it leaks into the message sent on their behalf.
+      expect(sentPrompt).toBe(
+        "Analyse this deck and save its visual language as a reusable presentation template.",
       );
       expect(sentMessage?.parts).toContainEqual(
         expect.objectContaining({
--- a/turbo/packages/core/src/presentation-template-skill.ts
+++ b/turbo/packages/core/src/presentation-template-skill.ts
@@ -8,9 +8,10 @@ const PRESENTATION_TEMPLATE_RESOURCE_DIR = "./generated/resources";
 /**
  * Tell a run how to reach the guide.
  *
- * One sentence pair, used both by the import message the template picker sends
- * and by the standing agent-tools prompt, so a deck dropped in the chat box
- * reaches the same instructions as one picked through the dialog.
+ * One sentence pair, carried by the standing agent-tools prompt so that any
+ * deck reaches these instructions — dropped in the chat box, picked through
+ * the import dialog, or asked for in words — without a user-visible message
+ * having to spell the pull out.
  */
 export function presentationTemplateSkillInstruction(): string {
   const skillPath = `${PRESENTATION_TEMPLATE_RESOURCE_DIR}/${PRESENTATION_REVERSE_TEMPLATE_RESOURCE_PATH}/SKILL.md`;


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: d4b0f50b02315cc6d37fd2a8ed7fae6901cf4a9e
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Drops the reverse-template guide pointer from the message the deck-import picker sends on the user's behalf, leaving only the request sentence. The pointer itself is unchanged in `presentationTemplateSkillInstruction()` and still reaches every run through the agent-tools prompt, so the change removes a duplicated instruction rather than a capability.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic — no wrong assumptions or unreachable paths | Pass | `presentation-template-import.ts:28` now returns one literal. The dropped half is still emitted by `agent-runs-create.service.ts:363` and `active-input-prompt.service.ts:210`, both gated on the same `FeatureSwitchKey.PresentationTemplates` that gates `presentationTemplateImportEnabled$`. No run that can receive an imported deck loses the pointer. |
| `REVIEW.md` | Security — no injection, secrets, or missing auth on changed paths | N/A | Diff changes a static English prompt literal and two doc comments; no auth, transport, storage, or user-supplied data path is touched. |
| `REVIEW.md` | Type safety — no `any`, no broad types | Pass | `presentationTemplateImportPrompt(): string` keeps its explicit return type; no casts, no `any`, no new types. |
| `REVIEW.md` | Style & maintainability — no oversized functions, no duplication, comments explain WHY | Pass | The change is itself a de-duplication (one instruction, one owner). Both rewritten comments state why the pointer lives in the agent-tools prompt, not what the code does. |
| `REVIEW.md` | Testing coverage — `fix:` commits must carry a regression test | Pass | `chat-composer-template-picker.test.tsx:4612` tightens `toContain` to `toBe` on the exact sentence, so any future append to the user-facing message fails. That is the regression this `fix:` introduces. |
| `REVIEW.md` | Feature regression guard — changed existing behavior needs coverage of the path most likely to regress and a containment gate | Pass | The path most likely to regress is "the run no longer knows how to reach the guide". `run-lifecycle.bdd.test.ts:1614` (unchanged, on `main`) asserts the pull command and `SKILL.md` path appear in `appendSystemPrompt` when the switch is on and are absent when off. Containment: the whole surface sits behind `PresentationTemplates`. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, contract, Zod response, or `jsonb().$type<>()` change; diff is confined to a platform signal, a core string builder, and one platform test. |
| `docs/bad-smell.md` | Zero `any`, zero lint/type suppressions | Pass | No `any`, no `eslint-disable`/`oxlint-disable`/`@ts-ignore`/`@ts-expect-error`/`prettier-ignore` added. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | The diff deletes a static import and adds none; all remaining imports are static top-level. |
| `docs/bad-smell.md` | No hardcoded URLs or environment-specific values | Pass | The retained literal is product copy. The resource ID, path, and dir stay derived from `PRESENTATION_REVERSE_TEMPLATE_RESOURCE_ID`/`_PATH` constants in `presentation-template-skill.ts`. |
| `docs/fallback.md` §9 | Every fallback introduced by the diff is declared in the PR summary | Pass | The diff introduces no fallback: no `??`/`||` chain, no tolerance branch, no dual-shape reader. Per §9 the PR correctly carries no `Fallbacks` section. |
| `docs/fallback.md` §1 | No negative tests asserting removed behavior stays removed | Pass | The test was updated to the new behavior via exact equality rather than adding a `not.toContain("okou resource pull")` tombstone — the shape §1 explicitly asks for. |
| `docs/fallback.md` §10 | A removal must carry type, single-writer, production, or rollout evidence | Pass | Treating the duplicated instruction as redundancy being removed: single-writer evidence is that `presentationTemplateSkillInstruction()` is emitted unconditionally by both API prompt builders under the same switch, and the unchanged BDD test pins it. The removal also deletes its own assertions in the same PR rather than leaving them behind. |
| `docs/fallback.md` §2 | A non-GA feature-switched feature needs no compatibility code | Pass | `feature-switch.ts:305` has `PresentationTemplates` at `enabled: false` with `enabledOrgIdHashes: STAFF_ORG_ID_HASHES` — staff-only, non-GA. No dual-shape handling or migration is owed here, and the PR correctly adds none. |
| `docs/deployment-compatibility.md` | Frontend and backend deploy independently; changed cross-surface behavior must be safe old↔new | Pass | Both directions are safe and the surface is non-GA anyway. Old app → new API: an already-open page still sends the two-paragraph message; the run receives the pointer twice, which is harmless. New app → old API: the agent-tools line already ships in the deployed API on `main`, so a new client is never the first to depend on it. No request/response shape, enum, payload, or persisted state changed. |
| `docs/effect.md` | Render stays pure; the API matches the owner and trigger (`computed` / semantic `command` / setup / `onRef`) | Pass | `importPresentationTemplateDeck$` remains one semantic `command` owning the whole transition, and `presentationTemplateImportEnabled$` remains a `computed`. The diff only shortens a pure helper called from inside that command; no boundary was reclassified. |
| `docs/effect.md` | Every async command has an `AbortSignal` with an eventual abort path | Pass | The command's `signal: AbortSignal` parameter and its `signal.throwIfAborted()` checks around upload and submit are untouched by the diff. |
| `docs/cache.md` RC-2 / RC-3 | No render-time signal allocation; no unbounded-lifetime caches | Pass | No signal factory, `Map`/`Set`/`WeakMap`, or module registry added; the helper is a pure string function called from a command, not from render. |
| `docs/cache.md` RC-5 | No duplicate mutable sources of truth | Pass | This is the finding the diff resolves: the guide pointer had two emitters for one fact, and now has one owner in the agent-tools prompt. |
| `.claude/skills/ccstate/SKILL.md` | `detach()`/`Reason.DomCallback` belong to views only, never to the signals directory | Pass | No `detach()` or `void` added in `src/signals/**`; the command still awaits `set(...)` and returns its promise to the caller. |
| `docs/testing.md` | Integration tests only; mock at the system boundary; no fake timers | Pass | The changed test is an existing page-level integration test. No `vi.mock()`, no `vi.stubGlobal("fetch")`, no `vi.useFakeTimers()` added; upload and run creation stay behind `context.mocks`. |
| `docs/testing.md` | Do not over-test volatile UI text | Pass | The exact-equality assertion is not incidental copy checking — the message body *is* the behavior this PR changes, and equality is what pins "nothing is appended". A looser assertion could not express that without a §1 tombstone. |
| `docs/testing/app-testing.md` | Platform tests are page-level and enter through `detachedSetupPage` | Pass | The test still drives `detachedSetupPage({ context, featureSwitches: { PresentationTemplates: true }, path: "/chats/…" })`, clicks the Template control, and uploads through the real file input. |
| `docs/testing/testing-external-behavior.md` | Construct and observe through the external interface, not internals | Pass | The case is built by real user actions (open picker → upload deck) and observed at the platform→API boundary via the run-create body (`sentPrompt`, `sentMessage.parts`). No store mutation, no direct component render, no hook calls. |
| `docs/testing/anti-patterns.md` AP-1 / AP-7 / AP-10 | No asserting on mock calls, implementation details, or directly rendered components | Pass | Assertions are on the outgoing prompt and the attached file part, not on `expect(mock).toHaveBeenCalled()`, internal state, or CSS classes. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- **Coverage: Adequate.** `fix:` regression is pinned by the tightened `toBe` assertion in `chat-composer-template-picker.test.tsx:4612`, which fails if any future change re-appends instructions to the user-facing message. The non-lossiness of the removal is pinned independently by the unchanged `run-lifecycle.bdd.test.ts:1614`, which asserts the pull command and `SKILL.md` path reach `appendSystemPrompt` when `PresentationTemplates` is on and stay out when it is off. Two remaining import tests (legacy `.ppt`, and the new-chat thread-creation path) still exercise the surrounding flow with the looser `toContain`, which is appropriate since they assert flow, not message shape.
- **Conventions: Compliant.** Page-level integration test entered through `detachedSetupPage`, no internal mocks, no fake timers, no direct fetch stubbing, assertions on external behavior.
- **Not verified locally:** full `test-app`/`test-api` suites were left to CI; `test-api (3)`, CodeQL, Semgrep, Gitleaks, and Pull Request Title are green at the time of review, with `test-app` shards still pending.

PR #29442, opened 2026-08-26
refactor: remove the latest website templates feature switch

bingjiez666:
Step 2 of 2. Unblocked: #29446 merged as `6e911fa28` on 2026-08-26 and is live in production (deployment `586565332`, ahead 146 / behind 0 of that commit, status `success`). `latestWebsiteTemplates` has been on for everyone for about a day, so the disabled side is now dead code.

Rebased onto `main` at `17c4c1d2`. Four conflicts, all "main has the rollout-era version, this PR deletes it": the switch entry, its two `feature-switch.test.ts` assertions, the chat-run website prompt case, and the three run-context environment assertions. `ConnectorAccounts` had independently flipped to `true` on main and was kept at main's value.

Re-verified after the rebase: `pnpm check-types` 14/14, `pnpm turbo run lint` 13/13 (bounded heap, concurrency 1), `pnpm knip` clean against baseline, prettier and `git diff --check` clean, `@okouai/core` 19 files / 207 tests, CLI website + resource 21 tests. `apps/api` vitest still cannot run locally (no PostgreSQL in the sandbox); left to CI.

## Terminal state

`fully-rolled-out`. The enabled side of `latestWebsiteTemplates` becomes the only website generation behaviour, and the switch is removed.

Basis: requested directly by @bingjiez666 (switch maintainer) — "将 latestWebsiteTemplates 开关全量，并下掉". The switch's own comments already named this as the planned end state (rollout + run-context drain, #26672).

## Introducing commit

`3089b1da9` — `feat: gate latest website templates (#26686)`. Every change below was reconciled against that patch and its parent.

## Behaviour kept (switch enabled)

- Latest built-in website template archives, resolved through the stable package ids (`template:<slug>`), and the `Template archive SHA-256` prompt line.
- Direct-HTML authoring guidance: `SKILL.md`, `node tools/compose.mjs`, `bash checks/verify.sh index.html qa`, `node tools/stage.mjs publish`, `bash checks/verify-published.sh`.
- `WEBSITE_IMAGE_BATCH_INSTRUCTION` in both the API generation-template prompt and `okou generate website`.
- The independent website resource index (`html-resources/website/v1/d7138a8f…/website.json`) for `okou generate website`.

## Behaviour removed (switch disabled)

- The pre-cutover renderer guidance: `resolve-images.mjs`, `render.mjs`, `okou host <output-dir> --site <slug>`.
- Routing website generation to the `-v2` package ids.
[...]

Issue #26672, opened 2026-08-12
chore: remove latest Website templates rollout switch after rollout


bingjiez666:
## Context

The latest built-in Website template archive release is being placed behind a default-off feature switch because the refreshed templates currently generate too slowly. While the switch is off, Website generation must keep using the pre-cutover archive SHA pins.

## Removal gate

Remove the switch only after all of the following are true:

- the latest Website template generation path meets the agreed latency target;
- the feature has been reviewed and enabled for all users;
- the production rollback window has closed;
- queued, active, and finalizing runs created while the switch was off have drained.

## Cleanup

- remove the `LatestWebsiteTemplates` feature switch and its disabled branch;
- make the latest independent Website registry and archive pins unconditional;
- remove disabled-branch tests together with the branch;
- only then proceed with the previous Website archive pin cleanup tracked by #26519;
- do not delete historical immutable R2 objects.

bingjiez666:
Completed by #29446 and #29442: the latest Website templates were enabled for everyone, and the rollout switch and disabled branch were removed.

The retained compatibility mapping for older Website archive digests is tracked separately in #26519. Closing this rollout-switch cleanup as completed.

PR #29446, opened 2026-08-26
feat: enable the latest website templates for everyone

bingjiez666:
Step 1 of 2. This PR only flips the default; the code removal is #29442 and stays blocked until this is live and verified.

## What changes

`FeatureSwitchKey.LatestWebsiteTemplates` goes from `enabled: false` + `enabledOrgIdHashes: STAFF_ORG_ID_HASHES` to `enabled: true` — GA, per the repo's own feature-switch guidance.

Nothing else moves. Both sides of every branch the switch selects stay in the code:

- `generation-template-prompt.ts` still builds either the direct-HTML guidance or the renderer guidance.
- `resource-registry.ts` still carries `PREVIOUS_WEBSITE_TEMPLATE_PACKAGES` and the `archiveVersion` parameter.
- `agent-run-create.service.ts` still writes `OKOU_WEBSITE_TEMPLATE_ARCHIVE_VERSION` from the switch.
- The CLI still reads that env.

## Why this is separate from the removal

Two reasons, both about keeping an escape hatch during the risky window:

1. **Revert cost.** Rolling this back is one line and one deploy. Rolling back the combined change would mean restoring deleted registry maps, the env plumbing, and the CLI helper.
2. **No-deploy mitigation.** While the switch exists, a specific org or user that hits a problem can be put back on the pre-cutover path through `POST /api/feature-switches` immediately, without waiting for a release.

## Known rollout risk

Prompts steered into an **already-running** run (`active-input-prompt.service.ts` — "steered into a run that is already executing, whose volumes were fixed when it was created") are built by the current API, but that sandbox's CLI and its `OKOU_WEBSITE_TEMPLATE_ARCHIVE_VERSION` were fixed when the run was created. A run created before this deploy by a not-yet-enrolled user therefore has `previous` baked in, while a website template steered into it afterwards would carry the latest guidance and the latest SHA. The agent would pull the pre-cutover renderer package and find no `SKILL.md` or `tools/compose.mjs`.

This is inherent to the rollout, not to the later removal: it exists for as long as pre-deploy runs stay alive, and it is what the run-context drain in #26672 covers. Newly created runs are not affected — `CLI_PKG_URL` is commit-addressed to the deploying commit and `deploy-cli` gates the API deploy, so a new run always pairs its prompt with the CLI from the same commit.

The override lever above is the mitigation if this shows up in practice.

## Test changes

The tests that encoded the old default were flipped rather than deleted:

- `run-lifecycle.bdd.test.ts`: the default run-context environment now asserts `latest`; `pins the latest Website template release into opted-in run contexts` became `pins the pre-cutover Website release into opted-out run contexts`, which now covers the rollback lever.
- `chat-events.bdd.test.ts`: the default website chat run now asserts the direct-HTML prompt; the switched case now overrides back to `false` and asserts the renderer prompt.
- `feature-switch.test.ts`: `otherOrgStates[LatestWebsiteTemplates]` is now `true`.

## Verification

| Check | Result |
| --- | --- |
| `pnpm -F @okouai/core exec vitest run` | 19 files / 206 tests pass |
| `pnpm -F api run check-types` (7-step chain, incl. `tsconfig.tests.json`) | pass |
| `pnpm -F api run lint`, `pnpm -F @okouai/core run lint` | pass |
| `prettier --write` on changed files | clean |

`apps/api` vitest could not run here: the sandbox has no PostgreSQL server or binaries, so every API suite fails at startup with `connect ECONNREFUSED 127.0.0.1:5432` before any test executes. Pre-existing environment limitation, unrelated to this diff. The edited API tests type-check and lint; their execution is left to CI.

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts
@@ -9802,48 +9802,50 @@ describe("CHAT-02: generation templates and attachments", () => {
       `Template: ${websiteTemplate.title} (${websiteTemplate.id})`,
     );
     expect(websitePrompt).toContain(
-      "okou resource pull template:black-slabs-v2 --dir ./generated/resources",
+      "okou resource pull template:black-slabs --dir ./generated/resources",
     );
     expect(websitePrompt).toContain(
-      `./generated/resources/${websiteTemplate.sourcePath}/render.mjs`,
+      "Image workflow: use supplied images first;",
     );
-    expect(websitePrompt).toContain("resolve-images.mjs");
-    expect(websitePrompt).not.toContain("use `seedream4` by default");
-    expect(websitePrompt).not.toContain("okou generate image-batch start");
-    expect(websitePrompt).toContain("okou host <output-dir> --site <slug>");
+    expect(websitePrompt).toMatch(
+      /npx --yes --package="\$\{CLI_PKG_URL\}" okou generate image-batch start <manifest\.tsv> <state-dir>/,
+    );
+    expect(websitePrompt).toMatch(
+      /npx --yes --package="\$\{CLI_PKG_URL\}" okou generate image-batch wait <state-dir>/,
+    );
+    expect(websitePrompt).not.toContain("tools/generate-images.mjs");
+    expect(websitePrompt).not.toContain("resolve-images.mjs");
+    expect(websitePrompt).not.toContain("render.mjs");
     await cancelChatRun(actor, website.runId);
 
+    // The rollout keeps its rollback lever: an override back to off restores
+    // the pre-cutover renderer guidance without a deployment.
     await updateFeatureSwitchesForUser(
       context,
       { ...actor, orgId: actor.orgId },
-      { [FeatureSwitchKey.LatestWebsiteTemplates]: true },
+      { [FeatureSwitchKey.LatestWebsiteTemplates]: false },
     );
-    const latestWebsite = await sendChatRun(actor, {
+    const previousWebsite = await sendChatRun(actor, {
       agentId,
-      prompt: "make a campaign landing page with the latest template",
+      prompt: "make a campaign landing page with the pre-cutover template",
       template: {
         type: "website",
         selection: { websiteTemplateId: websiteTemplate.id },
       },
     });
-    const latestWebsiteRun = await api.readRun(actor, latestWebsite.runId);
-    const latestWebsitePrompt = latestWebsiteRun.appendSystemPrompt ?? "";
-    expect(latestWebsitePrompt).toContain(
-      "okou resource pull template:black-slabs --dir ./generated/resources",
+    const previousWebsiteRun = await api.readRun(actor, previousWebsite.runId);
+    const previousWebsitePrompt = previousWebsiteRun.appendSystemPrompt ?? "";
+    expect(previousWebsitePrompt).toContain(
+      "okou resource pull template:black-slabs-v2 --dir ./generated/resources",
     );
-    expect(latestWebsitePrompt).toContain(
-      "Image workflow: use supplied images first;",
+    expect(previousWebsitePrompt).toContain(
+      `./generated/resources/${websiteTemplate.sourcePath}/render.mjs`,
     );
-    expect(latestWebsitePrompt).toMatch(
-      /npx --yes --package="\$\{CLI_PKG_URL\}" okou generate image-batch start <manifest\.tsv> <state-dir>/,
+    expect(previousWebsitePrompt).toContain("resolve-images.mjs");
+    expect(previousWebsitePrompt).toContain(
+      "okou host <output-dir> --site <slug>",
     );
-    expect(latestWebsitePrompt).toMatch(
-      /npx --yes --package="\$\{CLI_PKG_URL\}" okou generate image-batch wait <state-dir>/,
-    );
-    expect(latestWebsitePrompt).not.toContain("tools/generate-images.mjs");
-    expect(latestWebsitePrompt).not.toContain("resolve-images.mjs");
-    expect(latestWebsitePrompt).not.toContain("render.mjs");
-    await cancelChatRun(actor, latestWebsite.runId);
+    await cancelChatRun(actor, previousWebsite.runId);
   }, 90_000);
 
   it("uses R2 for archive-backed styles", async () => {
--- a/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
@@ -13740,29 +13740,29 @@ describe("RUN-01: agent runner context, queue promotion, and skills", () => {
         `https://${staticDomain}/okou-cli/test-commit/package.tgz`,
       );
       expect(r2Claim.environment?.[WEBSITE_TEMPLATE_ARCHIVE_VERSION_ENV]).toBe(
-        "previous",
+        "latest",
       );
       await api.requestCancelRun(actor, r2Run.runId, [200]);
     },
   );
 
-  it("pins the latest Website template release into opted-in run contexts", async () => {
+  it("pins the pre-cutover Website release into opted-out run contexts", async () => {
     const api = createRunsApi(context);
     const connectors = createConnectorBddApi(context);
     const { actor, agentId, runnerGroup } = await entitledRunActor();
     await connectors.updateFeatureSwitches(actor, {
-      [FeatureSwitchKey.LatestWebsiteTemplates]: true,
+      [FeatureSwitchKey.LatestWebsiteTemplates]: false,
     });
 
     const run = await api.createRun(actor, {
       agentId,
-      prompt: "use the latest Website template release",
+      prompt: "use the pre-cutover Website template release",
       modelProvider: "anthropic-api-key",
     });
     await api.heartbeatRunner(runnerGroup);
     const claim = await api.claimRunnerJob(run.runId);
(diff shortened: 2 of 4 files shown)


bingjiez666:
Follow-up: #29442 removes the switch and every disabled branch it selects. It is held until this PR is live and verified.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 8c5413b911bd88acdc3a756f90f98731c8c05b6c
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Step 1 of the two-step rollout: `FeatureSwitchKey.LatestWebsiteTemplates` goes from `enabled: false` + `enabledOrgIdHashes: STAFF_ORG_ID_HASHES` to `enabled: true`, which is the same GA shape already used by `ImageModelPicker` in the same map. No plumbing moves — `generation-template-prompt.ts`, `resource-registry.ts` (`PREVIOUS_WEBSITE_TEMPLATE_PACKAGES`, `archiveVersion`), `agent-run-create.service.ts` (`OKOU_WEBSITE_TEMPLATE_ARCHIVE_VERSION`), and the CLI's `websiteTemplateArchiveVersionFromEnvironment()` all keep both branches, so the per-org override stays available as a no-deploy rollback lever. The three affected tests were flipped to the new default rather than deleted, and two of them now exercise the opted-out path explicitly.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no wrong assumptions or unhandled edges | Pass | `turbo/packages/core/src/feature-switch.ts:318` flips one boolean and drops the now-redundant staff allow-list; `enabled: true` with no `enabledOrgIdHashes` matches the existing GA entries in the same record. `STAFF_ORG_ID_HASHES` is still referenced by `PiLoop`, `PresentationTemplates`, and `LatestPresentationTemplates`, so no symbol is orphaned. |
| `REVIEW.md` | Security: no new endpoint, auth path, injection surface, or hardcoded credential | N/A | The diff contains no auth, request-parsing, or credential code; it changes a default in a static feature-switch record. |
| `REVIEW.md` | Type safety: no `any`, no overly broad types | Pass | No type annotations added or changed in any of the four files. |
| `REVIEW.md` | Style and maintainability: no oversized functions, no duplicated logic, no WHAT comments | Pass | Only added comment is `chat-events.bdd.test.ts:9823`, which explains WHY the override-to-`false` block exists (it now covers the rollback lever), not what the next lines do. |
| `REVIEW.md` | `feat:` commits must carry integration tests | Pass | `run-lifecycle.bdd.test.ts:13742` and `:14100` assert the new default `latest` run-context environment through `claimRunnerJob`; `chat-events.bdd.test.ts:9805-9820` asserts the direct-HTML prompt is now the default. |
| `REVIEW.md` | Feature regression guard: the replaced existing behavior keeps explicit coverage | Pass | The pre-cutover path is not left uncovered. `run-lifecycle.bdd.test.ts:13748` (`pins the pre-cutover Website release into opted-out run contexts`) and `chat-events.bdd.test.ts:9825-9846` both drive the switch back to `false` and assert the renderer guidance and `previous` pin, which is exactly the mitigation lever the PR relies on. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No `turbo/packages/db/src/schema/**`, contract, Zod response schema, or `jsonb(...).$type<...>()` change; no migration in the diff. |
| `docs/bad-smell.md` | Zero `any` types | Pass | None present in the diff. |
| `docs/bad-smell.md` | Zero lint/type suppression comments | Pass | No `eslint-disable`, `oxlint-disable`, `@ts-expect-error`, or `prettier-ignore` added. |
| `docs/bad-smell.md` | Static imports only; no hardcoded URLs or environment values | Pass | No imports changed; no URL or environment literal introduced. |
| `docs/fallback.md` §1 | No negative tests asserting removed behavior stays removed | Pass | The `not.toContain("resolve-images.mjs")` / `not.toContain("render.mjs")` assertions discriminate between two branches that both still exist in `resource-registry.ts` and `generation-template-prompt.ts`. They pin the active branch's guidance, not the absence of deleted code, so they are not tombstones. |
| `docs/fallback.md` §2 | GA boundary: once `enabled: true`, normal deployment-compatibility rules apply in full | Pass | This PR is the transition itself. Nothing in the diff relies on the pre-GA exemption, and the follow-up cleanup (#26672) is gated, not assumed. |
| `docs/fallback.md` §9 | Every fallback introduced by the diff is declared in the PR summary | N/A | The diff introduces no fallback. It only keeps the existing switched branches, which §9 explicitly excludes from the `Fallbacks` requirement. No `Fallbacks` section is therefore expected or included in this review. |
| `docs/deployment-compatibility.md` | Commit-addressed CLI artifacts: a run keeps the CLI and context selected at creation time | Pass, with a note | Verified `websiteTemplateArchiveVersionForRun()` at `agent-run-create.service.ts:6286` writes the env only at context creation, and the CLI reader at `commands/shared/website-template-archive-version.ts` treats a missing value as `previous`. New runs are self-consistent because `CLI_PKG_URL` is commit-addressed to the deploying commit. The PR summary declares the one skewed pairing under "Known rollout risk"; see the non-blocking note below. |
| `docs/deployment-compatibility.md` | Frontend -> backend and runner -> backend request/response shapes stay compatible | Pass | No contract, route, runner protocol, or payload shape changes. `grep -rn LatestWebsiteTemplates` finds no consumer under `turbo/apps/platform`, so no already-loaded web client changes behavior from this flip. |
| `docs/testing.md` | Integration tests, mocked at the system boundary only | Pass | Both API test files drive behavior through BDD API helpers (`createRunsApi`, `createConnectorBddApi`, `sendChatRun`, `api.readRun`). No `vi.mock()` of an internal relative path, no `globalThis.services.db` mock, no `vi.useFakeTimers()`, no direct `fetch` stubbing added. |
| `docs/testing/api-testing.md` | Construct and verify state through the API surface; own shared persistent state | Pass | The switch is toggled via `connectors.updateFeatureSwitches(actor, ...)` and `updateFeatureSwitchesForUser(context, actor, ...)` — production API surfaces, not DB writes. Each case uses its own `entitledRunActor()`, so the org-scoped override stays owned by the test and cannot leak into concurrent files. |
| `docs/testing/anti-patterns.md` | AP-1 / AP-4 / AP-11: assert behavior, not mock calls; no internal mocking; test through the route | Pass | Assertions are on `claim.environment`, `run.appendSystemPrompt`, and `getAllFeatureStates` output. No `toHaveBeenCalled()` and no service-level imports added. |
| `docs/testing/patterns.md` | Follow the standard route-test structure; no ad hoc mock resets or DB setup | Pass | Both files keep their existing module-scope `testContext()` structure; the diff adds no `vi.clearAllMocks()`, `initServices()`, or DB setup. |
| `docs/testing/testing-external-behavior.md` | The API endpoint is the only trusted boundary for `apps/api` tests | Pass | Setup and verification both happen through endpoints; the feature-switch state itself is changed through the feature-switch API rather than by patching the static record. |
| `.claude/skills/ccstate/SKILL.md` | `detach()` + `Reason.DomCallback` for async commands in React views; no manual cache over `computed` | N/A | No signals, commands, computeds, or React views are touched. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Notes (non-blocking)

- `turbo/apps/api/src/signals/services/active-input-prompt.service.ts:206` resolves `latestWebsiteTemplatesEnabled` from the **live** switch state, while the sandbox's `OKOU_WEBSITE_TEMPLATE_ARCHIVE_VERSION` and `CLI_PKG_URL` were pinned at run creation. A website-template prompt steered into a run created before this deploy therefore carries the latest guidance (`okou resource pull template:black-slabs`, direct-HTML, `image-batch`) into a sandbox pinned to `previous`, where the pulled package has no `SKILL.md` or `tools/compose.mjs`. Not a blocker for three reasons: the mechanism is pre-existing and unchanged by this diff (it already fires whenever a switch flips mid-run, including staff enrolment or a manual override); the PR summary declares it with its surface and mitigation; and the exposure is bounded by the existing-sandbox drain of up to 2 hours. Verified that #26672 names the matching gate — "queued, active, and finalizing runs created while the switch was off have drained" — so the removal PR cannot land before this window closes. Recommend watching the `okou generate website` failure rate over the first two hours after promotion, and using the per-org `POST /api/feature-switches` override rather than a revert if a report appears.

### Testing

- **Coverage: Adequate.** For a default-flip `feat:`, the meaningful assertions are the run-context environment and the generated prompt, and both are covered on both sides of the switch. The default-enabled side is asserted at `run-lifecycle.bdd.test.ts:13742` / `:14100` and `chat-events.bdd.test.ts:9805-9820`; the opted-out side keeps a dedicated test at `run-lifecycle.bdd.test.ts:13748` plus the override block in `chat-events.bdd.test.ts:9825-9846`. `feature-switch.test.ts:160` pins the GA state for a non-staff org while `:124` continues to pin the staff org, so a regression in either direction fails.
- **No stale assertions elsewhere.** Checked every other file referencing `WEBSITE_TEMPLATE_ARCHIVE_VERSION_ENV`, `black-slabs-v2`, or `resolve-images.mjs`: `generation-template-prompt.test.ts` passes `latestWebsiteTemplatesEnabled: true` explicitly, `registry-resources-download.test.ts` exercises both archive sets directly, and `cli/generate/__tests__/website.test.ts` drives the env var itself. None of them inherits the default, so no untouched test encodes the old value.
- **Conventions: Compliant.** No internal-module mocking, no DB access, no fake timers, no `toHaveBeenCalled()`-only assertions, and state is both created and read through production API surfaces.
- **Local verification gap acknowledged in the PR body:** `apps/api` vitest cannot run in the authoring sandbox (no PostgreSQL), so the two BDD files are typecheck- and lint-verified locally and executed by CI. That matches the known sandbox limitation and is left to the merge-queue run.

PR #29450, opened 2026-08-26
perf(runner): stop commands on semantic stdout overflow

### LINKED ISSUE #29413: perf(rust): terminate bounded commands on semantic stdout overflow
## Problem

`run_output_bounded` treats semantic stdout above 4 MiB as fatal, but the current implementation bounds only retained memory, not the amount of output work or the child lifetime.

- `crates/runner/src/bounded_command.rs:117-133` waits for the child to exit or reach the overall timeout before collecting either output task.
- `crates/runner/src/bounded_command.rs:329-344` keeps reading until EOF. After `max_bytes` is retained, every subsequent chunk is still read with `keep == 0` and discarded.
- `crates/runner/src/bounded_command.rs:346-355` records semantic overflow only after EOF, so the parent cannot react when the first excess byte arrives.
- `crates/runner/src/bounded_command.rs:609-633` covers finite oversized stdout and stderr, but does not cover a long-running producer or prompt child cleanup on overflow.
- `crates/runner/src/executor/diagnostics/oom.rs:28-59` uses semantic stdout for the five-second `dmesg` OOM check.
- `crates/runner/src/cmd/service/systemctl.rs:84-104` uses the same path for bounded `systemctl` queries; several current query callers use the ten-second `SYSTEMCTL_QUERY_TIMEOUT` defined at `crates/runner/src/cmd/service/systemctl.rs:16`.

Once byte `max_bytes + 1` arrives, the result can no longer succeed. Nevertheless, an unbounded producer continues running and its output is continuously drained and discarded until the overall timeout. This consumes child CPU and pipe/read bandwidth for the full timeout and returns `TimedOut` instead of `OutputTooLarge`. A finite oversized `dmesg` or `systemctl` response also forces the runner to read every excess byte after the useful limit has already been crossed.

## Acceptance criteria

- Semantic stdout overflow promptly notifies the command owner and terminates and reaps the child without waiting for natural exit or the overall timeout.
- The overflow path returns `BoundedCommandError::OutputTooLarge` with the existing stream and limit details.
- Both output readers are cleaned up without deadlocking when stderr is also producing more than a pipe buffer.
- Diagnostic-only stderr truncation continues to drain safely so commands whose result is not already doomed can finish normally.
- Add a deterministic test with a small semantic limit and a long-running output producer that proves early completion, correct error classification, and child reaping.
- Existing timeout, cancellation, finite-output, and oversized-stderr cleanup coverage remains green.

seven332:
## Summary

- notify the bounded command owner on the first semantic stdout byte beyond the configured limit
- terminate and reap the child immediately while preserving diagnostic stderr draining
- add a real-process regression covering oversized stderr, early classification, and reaping

## Scope

- preserves the existing `BoundedCommandError::OutputTooLarge` shape and caller mappings
- no public API, protocol, persisted state, migration, or dependency changes

## Validation

- `cargo test -p runner bounded_command::tests`
- `cargo test -p runner cmd::service::signal::tests::signal_service_main_bounded_times_out_systemctl -- --exact`
- `cargo test -p runner`

Closes #29413

### DIFF
--- a/crates/runner/src/bounded_command.rs
+++ b/crates/runner/src/bounded_command.rs
@@ -654,7 +654,12 @@ mod tests {
                 return Err(error);
             }
         };
-        tokio::fs::write(&release_path, []).await?;
+        if let Err(error) = tokio::fs::write(&release_path, []).await {
+            task.abort();
+            let _ = task.await;
+            wait_for_process_exit(pid, starttime).await?;
+            return Err(error);
+        }
 
         let error = match tokio::time::timeout(PROCESS_OBSERVATION_TIMEOUT, &mut task).await {
             Ok(result) => result.unwrap().unwrap_err(),
--- a/crates/runner/src/bounded_command.rs
+++ b/crates/runner/src/bounded_command.rs
@@ -4,6 +4,7 @@ use std::time::Duration;
 
 use tokio::io::AsyncReadExt;
 use tokio::process::{Child, Command};
+use tokio::sync::oneshot;
 use tokio::task::JoinHandle;
 use tokio::time::Instant;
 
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #29450

**Reviewed head:** `90ef81b153ac1a0eb2946a5ba1316391dc82b1b9`

### Summary

Semantic stdout overflow now becomes an immediate owner-visible event, and the existing bounded child cleanup path terminates and reaps the process without changing caller-facing error details or diagnostic stderr semantics.

### Findings

No findings.

### Testing

Reviewed the real-process overflow/reaping regression and adjacent timeout, cancellation, finite-output, and diagnostic truncation coverage. The final head passes `cargo test -p runner`, runner-focused bounded-command tests, Rust formatting, Clippy, and documentation checks; the new regression also passed 20 consecutive runs. Turbo and Python checks were not run locally because the PR changes only the Rust runner crate.

### Verdict

LGTM

PR #29452, opened 2026-08-26
fix(python): fail closed on request hook exceptions

### LINKED ISSUE #29416: bug(python): Fail closed on request-hook exceptions after credential injection
## Problem and evidence

The buffered full-request hook can leave a managed credential-bearing request forwardable after an unexpected post-auth exception.

`_apply_resolved_firewall_auth()` applies resolved headers and query parameters to the live request. The full `request()` hook then performs post-auth catalog preparation and revalidation. Its outer exception path releases selected lifecycle bookkeeping and re-raises, but does not create a local response, kill the flow, or restore request state.

Mitmproxy invokes ordinary addon hooks under `addonmanager.safecall()`, which logs and consumes ordinary exceptions. After hook completion, the buffered HTTP state machine continues upstream whenever the flow was not killed and has no local response. Re-raising alone is therefore not fail-closed in production.

The existing unexpected-exception test covers cleanup for a failure before successful credential application. It does not drive a post-injection failure through mitmproxy's exception-catching dispatcher and HTTP state machine.

## Decision

Treat the buffered `request()` hook as the terminal security boundary for ordinary unexpected exceptions.

When an ordinary `Exception` occurs and no specific local response already exists, install a fixed method-aware JSON HTTP 500 response with error `request_processing_failed` before lifecycle cleanup. Then preserve the existing cleanup and re-raise the original exception so mitmproxy retains its traceback and addon-error logging.

Keep `asyncio.CancelledError` as cleanup plus propagation without manufacturing an HTTP response.

- A local response makes the complete buffered request terminal before mitmproxy consumes the exception.
- A fixed non-sensitive body gives the client a deterministic proxy failure without exposing exception text, credentials, URL query data, or other request details.
- Re-raising preserves observability instead of masking the programmer or runtime error.
- Installing the response before release operations keeps the flow fail-closed even if a secondary cleanup invariant fails.
- The success path does not copy or snapshot request state.

## Alternatives considered

- **Kill the flow and re-raise:** secure, but produces an opaque transport reset despite the proxy still owning a complete buffered request.
- **Snapshot and restore URL, headers, and query state:** appropriate for requestheaders fallback because a later full hook still owns processing, but unsafe here because the full hook would hand a failed authenticated operation directly to the upstream network with changed semantics. It also adds success-path copying and a fragile restoration contract.
- **Status quo:** rejected because mitmproxy consumes the exception and may forward the mutated request.

## Risks and mitigations

- Preserve an existing specific local response instead of overwriting it with the generic fallback.
- Keep cancellation ownership separate and cover it explicitly.
- Retain request timing, auth-base admission, AWS SigV4 admission, usage tracking, SigV4 inspection, and classification cleanup.
- Prove behavior through the real hook dispatcher and HTTP state machine, not only a direct function call.

## Acceptance criteria

- An ordinary unexpected exception after successful managed credential injection cannot open an upstream connection or send the request.
- The client receives a correctly framed fixed JSON HTTP 500 with `request_processing_failed`; the body contains no exception or secret-derived values.
- The original exception remains observable through mitmproxy's addon-error handling.
- Any existing specific local response is preserved.
- Cancellation still propagates without receiving the ordinary-exception response.
- Request timing, auth admissions, usage tracking, catalog state, and other existing cleanup ownership remain correct.
- Regression coverage exercises real mitmproxy `safecall()` and `HookCompleted` semantics.

## Delivery notes

This is one Python PR. It changes no API, database, persisted state, dependency, environment variable, or cross-deploy protocol. Validate the committed uv lock/environment, Ruff, BasedPyright, focused pytest coverage, and the complete addon pytest suite.

seven332:
## Summary

- install a fixed local `request_processing_failed` HTTP 500 before cleanup when the buffered request hook raises an ordinary unexpected exception
- preserve the original exception for mitmproxy addon-error logging while keeping `asyncio.CancelledError` response-free
- cover the production `trigger_event()` / `safecall()` and HTTP state-machine path, proving injected credentials cannot reach an upstream open or request send
- retain request timing and usage cleanup contracts for both ordinary exceptions and cancellation

## Scope

- no request restoration or unauthenticated upstream fallback
- no API, database, persisted state, dependency, environment, or runner protocol changes

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_mitmproxy_request_exception_framing.py tests/test_request_handler_usage_tracking.py` (18 passed)
- `uv run --no-sync python -m pytest tests/` (6636 passed)

Closes #29416

### DIFF
--- a/crates/runner/mitm-addon/src/mitm_addon.py
+++ b/crates/runner/mitm-addon/src/mitm_addon.py
@@ -1463,7 +1463,16 @@ async def request(flow: http.HTTPFlow) -> None:
             return
 
         _unhandled_request_classification(classification)
-    except (asyncio.CancelledError, Exception):
+    except (asyncio.CancelledError, Exception) as error:
+        if isinstance(error, Exception) and flow.response is None and flow.error is None:
+            flow.response = http_local_responses.make_local_json_response(
+                flow,
+                500,
+                {
+                    "error": "request_processing_failed",
+                    "message": "Request processing failed",
+                },
+            )
         flow.metadata.pop(metadata_keys.HTTP_REQUEST_START_MONOTONIC, None)
         auth_base_forwarder.release_forward_request_admission_from_flow(flow)
         aws_sigv4_body_admission.release_from_flow(flow)
--- a/crates/runner/mitm-addon/tests/test_mitmproxy_request_exception_framing.py
+++ b/crates/runner/mitm-addon/tests/test_mitmproxy_request_exception_framing.py
@@ -1,0 +1,149 @@
+"""Request exception framing through mitmproxy's real hook pipeline."""
+
+from pathlib import Path
+from unittest.mock import patch
+
+from h2 import events as h2_events
+from h2.config import H2Configuration
+from h2.connection import H2Connection
+from mitmproxy.addons.proxyserver import Proxyserver
+from mitmproxy.proxy import commands, events
+from mitmproxy.proxy.layers.http import SendHttp
+from mitmproxy.proxy.layers.http._events import RequestHeaders
+from mitmproxy.proxy.layers.http._hooks import (
+    HttpRequestHeadersHook,
+    HttpRequestHook,
+    HttpResponseHeadersHook,
+    HttpResponseHook,
+)
+from mitmproxy.test import taddons
+
+import codex_model_catalog_cache
+import mitm_addon
+from tests.mitmproxy_http_framing_helpers import start_http_layer
+from tests.request_handler_helpers import _write_github_firewall_registry
+
+
+async def test_post_auth_request_exception_fails_closed_after_dispatch(
+    tmp_path: Path,
+    caplog,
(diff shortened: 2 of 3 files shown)


seven332:
## Code Review: PR #29452

**Reviewed head:** `ea2afb3075eb2bd5df5a747742945af3abcfa9cc`

### Summary

The request-hook exception boundary now fails closed with a fixed local response while preserving cleanup, cancellation semantics, and original exception observability. The integration test exercises mitmproxy's real dispatcher and HTTP/2 state machine after credential injection.

### Findings

No findings.

### Testing

Ruff, BasedPyright, focused integration coverage, the full 6,636-test addon suite, and five repeated focused review runs passed. CI coverage reports all changed executable lines covered. Unrelated Turbo and Rust suites were not run locally because the implementation scope is Python-only.

### Verdict

LGTM

Issue #28261, opened 2026-08-20
bug(api): investigate recurring Axiom telemetry ingest timeouts


seven332:
## Background

Production Axiom ingestion exceeded caller-side deadlines across three API telemetry paths on 2026-08-19. This delivery parent covers attributable diagnostics, production verification, and an evidence-gated decision about whether stronger delivery semantics are necessary.

This is distinct from #27779, which contextualizes application-log delivery timeouts, and #28081, which preserves HTTP-200 partial-ingest diagnostics.

## Confirmed production evidence

Live APL queries covered `2026-08-18T02:30:00Z` through `2026-08-20T02:33:00Z`.

### Shared telemetry SDK

- 45 `Axiom client operation failed` records contained an Undici `TimeoutError`.
- 44 occurred in one `pdx1` deployment from `2026-08-19T05:30:51.525Z` through `06:04:06.793Z`; one occurred in another `pdx1` deployment at `11:38:32.061Z`.
- The pinned `@axiomhq/js` 1.5.0 client uses a 20-second request timeout. Its `onError(error)` callback does not expose the internal dataset, batch size, elapsed time, retry count, or response phase.

### Optional agent-event traces

- 21 `Optional Axiom trace delivery failed` records contained the same timeout error from `05:30:37.008Z` through `06:04:07.978Z`.
- This direct path has a 10-second deadline and runs outside the required database acknowledgement.
- There is no durable API-side retry for the failed optional trace invocation.
- One separate HTTP-200 partial ingest in the broader window remains tracked by #28081.

### GuestAgent telemetry requests

- The 6 `Unhandled request error` records were `/api/webhooks/agent/telemetry` requests whose direct Axiom fetch hit the route's 10-second deadline; they were not an unrelated request-timeout class.
- All 6 returned HTTP 500 after 10,016–10,067 ms and came from GuestAgent `0.74.13` in the main `pdx1` deployment.
- The focused window contained 1,520 successful responses and these 6 failures.
- Each affected session subsequently returned HTTP 200 within 11–30 seconds. GuestAgent retains its local cursors after failure and retries the same deltas.
- The route concurrently sends up to three dataset batches. One batch can succeed before another causes `Promise.all` to reject, and a caller abort cannot show whether Axiom accepted the request before the response was lost. Retrying can therefore duplicate accepted data.

### Correlation

- The direct and SDK failures formed overlapping bursts around 05:30, 05:37–05:40, and 06:03–06:04 UTC.
[...]

PR #29454, opened 2026-08-26
fix(api): attribute shared axiom sdk failures

### LINKED ISSUE #29425: bug(api): attribute shared Axiom SDK ingest failures
## Parent objective

This is the second diagnostic delivery slice of #28261. The parent investigates recurring production Axiom ingest timeouts, improves truthful attribution at supported boundaries, and then uses deployed evidence to decide whether stronger delivery semantics are justified.

## Background

The structured direct-ingest diagnostics delivered by #28306 and #28321 now identify the affected dataset, batch dimensions, configured deadline, elapsed time, and abort classification for repository-owned fetches.

The remaining shared `@axiomhq/js` path uses one batched client per token class. Each client can batch several datasets, while the supported `onError(error)` callback exposes only the error. Production therefore records `Axiom client operation failed` with `client=sessions|telemetry` but cannot truthfully identify which internal dataset batch failed.

The 2026-08-25 recurrence recorded on #28261 included two shared-client timeout records in addition to 53 attributable direct-ingest timeout records.

## Requirement

Add payload-safe, exact dataset attribution for shared Axiom SDK ingest failures while preserving existing delivery behavior. Do not inspect private SDK state or claim dimensions that the selected public boundary cannot observe.

## Decision

Use one public batched `Axiom` ingestion client per dataset and bind the token-class client name plus exact dataset in its supported `onError(error)` callback. Retain separate lazy token-class clients for queries. Make `flushAxiom()` concurrently flush all registered dataset clients selected by `all|sessions|telemetry` and preserve best-effort completion.

The SDK already maintains an independent batch, one-second timer, 1,000-event threshold, and ordering boundary for each dataset. Separating those batches into dataset-bound public clients changes attribution ownership without replacing the SDK batching or transport algorithm. Each SDK HTTP client contains request configuration and uses the same global fetch path; it has no cross-dataset delivery state.

## Alternatives considered

- Keeping token-class clients and reporting a candidate dataset set is truthful but does not identify the failed batch and leaves the parent production gate ambiguous.
- Upgrading to `@axiomhq/js@2.0.0` does not help because the current official source still exposes only `onError(error)`; the major also introduces unrelated ingestion behavior.
- Parsing stacks, tracking the most recent enqueue, or inspecting SDK batch state cannot provide supported truthful attribution.
- Repository-owned batching or direct transport would provide more context but changes delivery semantics and remains gated by #28261 production evidence.

## Acceptance criteria

- Shared SDK ingest timeout and transport-error records identify the exact failed dataset from a supported repository-owned callback boundary.
- Timeout and non-timeout transport failures are distinguished as `timeout` and `transport_error` without changing the existing error level or message boundary.
- Existing one-second/1,000-event batching, SDK retry and timeout behavior, token selection, event ordering, and `flushAxiom()` completion behavior remain unchanged.
- Multiple datasets sharing the telemetry token are covered at the integration boundary, including client reuse, independent failure attribution, and successful ingestion.
- `flushAxiom()` coverage proves that all queued shared-client batches selected by `client=all|sessions|telemetry` are flushed concurrently and that sibling flushes continue after one rejection.
- Unexpected flush rejection remains best-effort and records both the token-class client and exact dataset.
- Diagnostic records contain no Axiom token, submitted event payload, prompt, or event field values.
- No API contract, database schema, environment variable, dependency version, runner/GuestAgent protocol, acknowledgement, direct-ingest, timeout, or retry change.

## Relevant files

- `turbo/apps/api/src/signals/external/axiom.ts`
- `turbo/apps/api/src/__tests__/mocks.ts`
- `turbo/apps/api/src/signals/external/__tests__/axiom.test.ts`
- `turbo/apps/api/src/__tests__/app-factory.test.ts`

## Risks and mitigations

- **Client cardinality:** reuse one client per dataset; the old SDK client already retained one long-lived batch entry per observed dataset.
- **Flush drift:** enumerate selected registry entries explicitly, invoke them concurrently, wait for all results, and cover every selector plus rejection continuation.
- **Callback misattribution:** bind immutable dataset context at client construction rather than ambient last-call state.
- **Payload leakage:** emit only client, dataset, failure kind, and original error; assert unique token and event values are absent.
- **Shared mock regression:** keep per-instance mocks delegating to the existing aggregate mocks and run the request-log entry-point suite.

## Dependencies

- #28306 is complete and establishes the direct-ingest diagnostic vocabulary.

## Intentionally deferred

- Replacing `@axiomhq/js` or bypassing it with repository-owned transport.
- Changing SDK batching, timeout, retry, compression, or dependency version.
- Durable queues, retries, idempotency, or stronger delivery guarantees.
- Production target-dataset correlation and the final delivery-contract decision owned by #28261 Gate 3.

seven332:
## Summary

- bind shared Axiom SDK ingestion clients to individual datasets so asynchronous SDK failures carry exact client and dataset context
- classify callback failures as timeouts or transport errors without logging tokens or payloads
- flush every registered dataset client for the selected token class while preserving best-effort failure isolation

## Scope

- does not add run-level attribution to SDK batches because one batch can contain events from multiple runs
- does not change direct Axiom ingestion or query behavior
- does not close the delivery parent issue

## Validation

- `pnpm exec prettier --check apps/api/src/__tests__/mocks.ts apps/api/src/signals/external/__tests__/axiom.test.ts apps/api/src/signals/external/axiom.ts`
- targeted Oxlint, type-aware Oxlint, and ESLint checks for the three changed files
- `pnpm -F api check-types`
- `pnpm -F api exec vitest run src/signals/external/__tests__/axiom.test.ts --reporter=verbose`
- `pnpm -F api exec vitest run src/__tests__/app-factory.test.ts --silent=passed-only`
- `pnpm knip --workspace api`
- pre-commit full-workspace `pnpm knip` and `pnpm check-types`

The full `pnpm -F api lint` command still reports pre-existing warnings in unrelated API test files; all lint layers pass for the changed files.

Closes #29425

Parent context: #28261

### DIFF
--- a/turbo/apps/api/src/__tests__/mocks.ts
+++ b/turbo/apps/api/src/__tests__/mocks.ts
@@ -20,6 +20,13 @@ type SyncMock = Mock<(...args: unknown[]) => void>;
 type UnknownMock = Mock<(...args: unknown[]) => unknown>;
 interface AxiomClientOptions {
   readonly onError?: (error: Error) => void;
+  readonly token?: string;
+}
+interface AxiomSdkClientMock {
+  readonly options: AxiomClientOptions;
+  readonly flush: AsyncMock;
+  readonly ingest: UnknownMock;
+  readonly query: AsyncMock;
 }
 interface BrowserUseCdpCommand {
   readonly id: number;
@@ -51,6 +58,7 @@ export interface ApiTestMocks {
   };
   readonly axiom: {
     readonly clientError: Mock<(error: Error) => void>;
+    readonly clients: AxiomSdkClientMock[];
     readonly flush: AsyncMock;
     readonly ingest: BooleanMock;
     readonly query: AsyncMock;
@@ -274,6 +282,7 @@ export interface ApiTestMocks {
 const apiTestMocks: ApiTestMocks = vi.hoisted((): ApiTestMocks => {
   const axiom = {
     clientError: vi.fn<(error: Error) => void>(),
+    clients: [],
     flush: vi.fn<(...args: unknown[]) => Promise<unknown>>(),
     ingest: vi.fn<(...args: unknown[]) => boolean>(),
     query: vi.fn<(...args: unknown[]) => Promise<unknown>>(),
@@ -1121,11 +1130,20 @@ vi.mock("@axiomhq/js", () => {
       apiTestMocks.axiom.clientError.mockImplementation((error: Error) => {
         options.onError?.(error);
       });
-      return {
-        flush: apiTestMocks.axiom.flush,
-        ingest: apiTestMocks.axiom.sdkIngest,
-        query: apiTestMocks.axiom.query,
+      const client: AxiomSdkClientMock = {
+        options,
+        flush: vi.fn((...args: unknown[]) => {
+          return apiTestMocks.axiom.flush(...args);
+        }),
+        ingest: vi.fn((...args: unknown[]) => {
+          return apiTestMocks.axiom.sdkIngest(...args);
+        }),
+        query: vi.fn((...args: unknown[]) => {
+          return apiTestMocks.axiom.query(...args);
+        }),
       };
+      apiTestMocks.axiom.clients.push(client);
+      return client;
     }),
   };
 });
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #29454

**Reviewed head:** `64850816a1438ccf52e08e45f3b78040f8268178`

### Summary

The dataset-bound Axiom SDK clients provide exact, payload-safe failure attribution while preserving token selection, SDK batching/retry/timeout behavior, per-dataset ordering, and best-effort flush semantics.

### Findings

No findings.

### Testing

The focused adapter tests cover client reuse, timeout and transport attribution, all flush selectors, concurrent sibling flushes, rejection isolation, and sensitive-field exclusion. The shared mock regression suite, targeted formatting/lint, API type checks, API Knip, and pre-commit full-workspace Knip/type checks passed. Full CI is still evaluated separately.

### Verdict

LGTM

PR #29224, opened 2026-08-25
perf(vsock-guest): eliminate idle drain wakeups

seven332:
## Summary

- replace the shared vsock output drain's 100 ms timer polling with indefinite polling on output and cancellation descriptors
- wake every stdout/stderr drain in a cancellation group by closing one shared cancellation-pipe writer
- propagate the active cancellation primitive through exec, write-file, DNS readiness, storage manifest, and drain-deadline paths
- add deterministic coverage for two idle workers, pre-cancellation, repeated cancellation, and owned-reader closure while output writers remain open

## Compatibility

- no wire protocol, public API, persisted state, configuration, or host/runner contract changes
- each active drain cancellation group now owns one two-fd wake pipe shared by all of its workers

## Validation

- `cargo test --manifest-path crates/Cargo.toml -p vsock-guest`
- `cargo clippy --manifest-path crates/Cargo.toml -p vsock-guest --all-targets -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc --manifest-path crates/Cargo.toml -p vsock-guest --no-deps`
- `cargo fmt --manifest-path crates/Cargo.toml --all -- --check`
- pre-commit workspace Rust fmt, Clippy, rustdoc, and file-size hooks

Closes #29201

PR #29455, opened 2026-08-26
fix(vsock): bound file write response waits

### LINKED ISSUE #29406: bug(vsock): investigate reused private write response stalls
## Background

The production Axiom error review for 2026-08-25 Beijing time found three run failures on reused sandboxes with the same terminal error:

```text
sandbox error: sandbox write file failed (timeout): request timeout
```

Report: [Axiom Error Review — 2026-08-25](https://axiom-error-review-2026-08-25.sites.vm0.io)

The original incident report could not identify the caller from the terminal log alone. Correlating the affected run IDs with the existing sandbox-operation telemetry now identifies every failed phase.

## Confirmed production attribution

| Beijing time | Report run | Failed operation | Duration | Runner |
| --- | --- | --- | ---: | --- |
| 2026-08-25 15:32:58 | `[run:010]` | `runner_run_payload_write` | 300001 ms | `v0.172.0` |
| 2026-08-25 15:37:10 | `[run:011]` | `runner_run_payload_write` | 300001 ms | `v0.172.0` |
| 2026-08-25 16:08:54 | `[run:025]` | `runner_user_env_write` | 300001 ms | `v0.172.0` |

All three runs:

- used a reused sandbox;
- completed storage preparation successfully;
- recorded `session_history_restore_skip` before the failed write;
- failed before `runner_agent_start_process`; and
- lost normal-operation availability after the timeout, so guest-log collection and workspace promotion also failed. Those later failures are consequences of the fail-closed connection state, not independent causes.

Session-history restoration and storage-cache staging were not the failed callers in these incidents.

## Failure boundary

The failed operations write required private runtime inputs through `Sandbox::write_private_file`:

- `runner_user_env_write` writes the optional serialized user environment;
- `runner_run_payload_write` writes the required guest-agent run payload.

The public sandbox error groups ordinary, batched, and private writes under `SandboxOperation::WriteFile`, which explains why the terminal log alone was ambiguous. Existing operation telemetry already provides the caller phase and elapsed time; it should remain the canonical caller attribution instead of duplicating full paths in logs.

The host file RPC currently allows 300 seconds per request. The guest write helper has a 30-second command timeout, a 5-second output-drain deadline, and a 10-second response-write deadline. Because the host waited the complete 300 seconds instead of receiving a guest terminal result, the unresolved boundary is the host/guest request lifecycle: the frame may have stalled before delivery, the connection-owned file worker may not have completed its response path, or the terminal response may not have reached the host router.

The available production evidence does not distinguish those transport stages. The timeout also makes the write outcome uncertain, so the same connection must remain fail-closed and must not receive a blind retry.

## Version evidence

Production operation telemetry from 2026-08-25 onward shows:

- `v0.171.11` and `v0.171.12`: zero private input-write failures across 1,427 runs;
- `v0.172.0`: three distinct failures across approximately 347 runs; and
- `v0.173.1` through `v0.173.7`: zero failures across 4,538 runs.

Runner `v0.173.1` includes [#29224](https://github.com/vm0-ai/vm0/pull/29224), which replaced periodic drain polling with explicit cancellation wakeups in the shared path used by write-file handling. The absence of later failures is strong mitigation evidence, but it does not prove causation because the affected guest state was destroyed and its diagnostic files could not be copied after the connection became unavailable.

## Required behavior

User environment and run payload data cannot be skipped: starting guest-agent without required private input would be incorrect. A private-write timeout therefore remains fatal to the current sandbox.

Recovery, if added, must happen only before agent spawn by discarding the uncertain reused sandbox and retrying the run on a fresh sandbox. Do not retry the write on the same connection. Treat job-level fresh-sandbox recovery as a separate delivery slice if it cannot remain PR-sized with the transport diagnosis.

## Relevant code paths

- [`crates/runner/src/executor/env.rs`](https://github.com/vm0-ai/vm0/blob/main/crates/runner/src/executor/env.rs) — private user-environment and run-payload writes.
- [`crates/runner/src/executor/agent_run.rs`](https://github.com/vm0-ai/vm0/blob/main/crates/runner/src/executor/agent_run.rs) — phase telemetry and pre-spawn ordering.
- [`crates/vsock-host/src/file/write.rs`](https://github.com/vm0-ai/vm0/blob/main/crates/vsock-host/src/file/write.rs) — host write request lifecycle and 300-second deadline.
- [`crates/vsock-guest/src/file_write_worker.rs`](https://github.com/vm0-ai/vm0/blob/main/crates/vsock-guest/src/file_write_worker.rs) — connection-owned guest file-write worker.
- [`crates/vsock-guest/src/handlers.rs`](https://github.com/vm0-ai/vm0/blob/main/crates/vsock-guest/src/handlers.rs) — guest helper timeout and response construction.

## Acceptance criteria

- Add deterministic host/guest boundary coverage for a private write whose terminal response does not arrive; do not use artificial sleeps.
- Determine whether current main, including #29224, closes the confirmed stalled-response path. Document the proven mechanism instead of inferring it only from the production version correlation.
- Emit bounded, secret-safe failure attribution for the RPC stage and timeout budget while retaining `runner_user_env_write` and `runner_run_payload_write` as the caller phases. Do not log user content or full guest paths.
- Ensure a post-write timeout keeps the connection and sandbox in a fail-closed, non-reusable state.
- Reconcile the host request deadline with the bounded guest helper, drain, and response-write budgets so a missing response cannot delay failure for five minutes without justification.
- Verify successful private writes, guest-side helper timeout, pre-frame failure, post-frame timeout, and reused-sandbox failure cleanup through focused integration coverage.
- Decide and document job-level behavior: fail the run after discarding the sandbox, or deliver a separate pre-spawn fresh-sandbox recovery slice. Never skip required input or retry on the uncertain connection.

## Current status

The exact failed callers are identified, and no recurrence has been observed after runner `v0.173.1`. The remaining work is to prove the transport mechanism, preserve fail-closed correctness, and remove the unjustified five-minute diagnostic gap before deciding whether separate job-level recovery is warranted.

seven332:
## Summary

- replace the fixed 300-second file-write wait with a shared 60-second end-to-end budget derived from guest helper, drain, frame-write, and transport allowances
- preserve the observed timeout stage across vsock and sandbox errors, then attach its bounded label to the existing runner private-write telemetry actions
- keep uncertain writes fail-closed: clear the pending route, mark the connection non-parkable, and never retry the request on the same connection

## Evidence boundary

#29224 proves that explicit cancellation now wakes idle guest drain workers, but the destroyed production sandboxes prevent attributing the three historical stalls to the former polling behavior. This PR closes the still-reproducible general case: a complete private-write request whose terminal response never reaches the host router can no longer consume the unexplained five-minute deadline, and its last host-observed stage is retained for diagnosis.

## Scope

- no vsock wire-format or persisted-state changes
- no automatic retry or fresh-sandbox recovery; those require a separate policy decision because the guest may already have applied the write

## Validation

- `cargo clippy -p guest-contracts -p vsock-host -p vsock-guest -p sandbox -p sandbox-fc -p runner --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps -p guest-contracts -p vsock-host -p vsock-guest -p sandbox -p sandbox-fc -p runner --all-features`
- `cargo test -p guest-contracts -p vsock-host -p vsock-guest -p sandbox -p sandbox-fc -p runner --all-targets --all-features`

Closes #29406

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -3750,6 +3750,7 @@ dependencies = [
 name = "vsock-host"
 version = "0.17.93"
 dependencies = [
+ "guest-contracts",
  "nix",
  "sandbox",
  "shell-quote",
--- a/crates/guest-contracts/src/file_write.rs
+++ b/crates/guest-contracts/src/file_write.rs
@@ -7,8 +7,8 @@ use crate::exec_terminal::EXEC_OUTPUT_DRAIN_DEADLINE;
 /// Maximum time the guest waits for its file-write helper process.
 pub const WRITE_FILE_HELPER_TIMEOUT_MS: u32 = 30_000;
 
-/// Duration form of [`WRITE_FILE_HELPER_TIMEOUT_MS`] for budget composition.
-pub const WRITE_FILE_HELPER_TIMEOUT: Duration = Duration::from_secs(30);
+// Duration form of `WRITE_FILE_HELPER_TIMEOUT_MS` for budget composition.
+const WRITE_FILE_HELPER_TIMEOUT: Duration = Duration::from_secs(30);
 
 /// Maximum time the guest spends writing one terminal protocol frame.
 ///
@@ -16,9 +16,9 @@ pub const WRITE_FILE_HELPER_TIMEOUT: Duration = Duration::from_secs(30);
 /// response, so this remains a connection-wide frame deadline.
 pub const GUEST_FRAME_WRITE_DEADLINE: Duration = Duration::from_secs(10);
 
-/// Host-side allowance for request-frame construction, transmission, guest
-/// dispatch, and scheduling beyond the guest's configured processing budgets.
-pub const WRITE_FILE_TRANSPORT_HEADROOM: Duration = Duration::from_secs(15);
+// Host-side allowance for request-frame construction, transmission, guest
+// dispatch, and scheduling beyond the guest's configured processing budgets.
+const WRITE_FILE_TRANSPORT_HEADROOM: Duration = Duration::from_secs(15);
 
 /// End-to-end host deadline for one file-write request frame and its terminal
 /// response.
--- a/crates/guest-contracts/src/file_write.rs
+++ b/crates/guest-contracts/src/file_write.rs
@@ -1,0 +1,42 @@
+//! Shared guest file-write timing budgets.
+
+use std::time::Duration;
+
+use crate::exec_terminal::EXEC_OUTPUT_DRAIN_DEADLINE;
+
+/// Maximum time the guest waits for its file-write helper process.
+pub const WRITE_FILE_HELPER_TIMEOUT_MS: u32 = 30_000;
+
+/// Duration form of [`WRITE_FILE_HELPER_TIMEOUT_MS`] for budget composition.
+pub const WRITE_FILE_HELPER_TIMEOUT: Duration = Duration::from_secs(30);
+
+/// Maximum time the guest spends writing one terminal protocol frame.
+///
+/// File-write results share the connection writer with every other guest
+/// response, so this remains a connection-wide frame deadline.
+pub const GUEST_FRAME_WRITE_DEADLINE: Duration = Duration::from_secs(10);
+
+/// Host-side allowance for request-frame construction, transmission, guest
+/// dispatch, and scheduling beyond the guest's configured processing budgets.
+pub const WRITE_FILE_TRANSPORT_HEADROOM: Duration = Duration::from_secs(15);
+
+/// End-to-end host deadline for one file-write request frame and its terminal
+/// response.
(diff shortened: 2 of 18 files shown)


seven332:
## Code Review: PR #29455

**Reviewed head:** `a06923eeea9240ea4a6ab8771f08c5ab53d96457`

### Summary

The PR replaces the unexplained five-minute file-write wait with a composed 60-second contract, preserves typed host-observed stages through sandbox errors, and adds bounded caller-phase telemetry while retaining fail-closed ownership and no-retry behavior.

### Findings

No findings.

### Testing

Affected Rust formatting, Clippy, rustdoc, and complete tests passed. Regression coverage exercises pre-frame, frame-write, and missing-terminal host states plus both reused runner private-input phases without wall-clock delay. Local Firecracker/KVM E2E was not run; CI is evaluated separately.

### Verdict

LGTM

PR #29501, opened 2026-08-26
fix(runner): reject stale host oom evidence

### LINKED ISSUE #29488: bug(rust): scope host OOM evidence to the current Firecracker lifetime
## Problem and evidence

`check_host_oom` has a five-second execution deadline, but its `dmesg` query has no lower bound on the age of returned kernel records. `host_dmesg_indicates_oom` then searches the complete output independently for an `oom-kill` marker and a `task=firecracker,pid=<pid>` substring.

Linux can reuse process IDs while kernel ring-buffer records from an earlier process remain available. The Runner stores the Firecracker PID at spawn time and still exposes it after the process exits, but the host OOM check does not receive a lifecycle boundary for that PID. Therefore this sequence can produce a false host-OOM classification:

1. An older Firecracker process with PID `N` is OOM-killed and leaves an `oom-kill:...task=firecracker,pid=N` record in the kernel log.
2. PID `N` is reused by a later Firecracker process.
3. The later sandbox fails for an unrelated reason after starting guest-agent.
4. `check_host_oom(N)` accepts the stale record and `agent_run` reports `HostMemoryOomKilled` instead of preserving the actual sandbox failure.

The current integration coverage verifies the observed production record format, process name, exact PID boundary, and command failure behavior. It does not prove that matched evidence belongs to the active execution lifetime. The matcher also permits the OOM marker and task/PID text to come from different records in the same output.

## Impact

A non-OOM sandbox crash can be reported as `host_memory_oom_killed`. This hides the original failure, emits incorrect resource-failure telemetry, and can send retry or operational handling down the wrong path.

## Decision

Require positive host-OOM evidence to identify the current Firecracker process instance and the active execution window, not only a reusable PID. Treat missing or uncorrelatable freshness evidence as no host-OOM proof so the original generic sandbox failure is retained.

This is a focused Runner correctness fix. It does not change guest OOM detection, host memory policy, sandbox retry policy, or Firecracker startup-failure classification.

## Chosen direction

Capture `CLOCK_MONOTONIC` immediately before the guest process start request that precedes the existing wait-error path. Retain that cutoff with the prepared-agent state. On a later wait error, request raw boot-relative timestamps from `dmesg` and accept only one record at or after the cutoff that contains both the case-sensitive `oom-kill` marker and the exact `task=firecracker,pid=<pid>` identity.

The run cutoff is narrower than the Firecracker spawn boundary, but it covers every host OOM event observable by this diagnostic: the host check runs only after guest process creation succeeded and its subsequent wait failed. It excludes all records from older PID generations without adding a Linux-specific clock to the provider-neutral `Sandbox` trait.

## Alternatives considered

- Capturing the Firecracker spawn timestamp in `sandbox-fc` would model the complete process lifetime, but would widen the shared sandbox abstraction for no current pre-start diagnostic consumer.
- Capturing `/proc/<pid>/stat` start time would reuse a stable process generation, but adds procfs I/O and clock-tick conversion to every run.
- Retaining `/dev/kmsg` sequence events avoids timestamp parsing, but introduces a shared event subsystem, retention, and concurrency behavior disproportionate to this path.
- Using wall-clock `dmesg --since` is smaller, but documented wall-time and suspend/resume conversion inaccuracies do not satisfy fail-closed freshness.

## Constraints and risks

- Preserve the existing bounded command execution and bounded output behavior.
- Keep external `dmesg` execution off the successful execution path; only the monotonic clock read is added there.
- Do not broaden a failed or unavailable host diagnostic into positive OOM evidence.
- Require freshness information that remains available after Firecracker has exited and its `/proc/<pid>` entry is gone.
- Parse only the explicitly requested raw timestamp grammar and fail closed on malformed or unsupported records.
- Compare at dmesg's emitted precision and cover the equality boundary in tests.
- Preserve compatibility between independently deployed components; this change remains internal to one Runner artifact.

## Acceptance criteria

- Capture a monotonic cutoff before the guest process creation whose later wait failure triggers host OOM diagnosis.
- Reject a correctly formatted same-PID OOM record when it predates that active execution cutoff, including a record from an older Firecracker PID generation.
- Accept a correctly formatted same-PID OOM record produced at or after the cutoff during the active execution.
- Require the `oom-kill` marker and exact `task=firecracker,pid=<pid>` identity to belong to the same fresh record.
- If the host diagnostic fails, times out, exceeds its output bound, or cannot establish freshness, retain the original generic sandbox failure instead of asserting host OOM.
- Add integration-level regression coverage for stale same-PID evidence, current-lifetime evidence, mixed-record evidence, malformed freshness evidence, and existing PID matching behavior.

## Delivery notes

- Implement as one Runner PR touching the diagnostic, active-run state propagation, focused tests, and the existing dependency feature declaration.
- No API, database, protocol, migration, persisted-state, or staged rollout work is required.
- Validate Runner formatting, Clippy, rustdoc, all-target/all-feature tests, and selected pre-commit hooks.

seven332:
## Summary

- capture a host monotonic cutoff before guest process creation and retain it through the wait-error path
- parse explicit raw-timestamped `dmesg` records and require fresh, same-line host OOM evidence for the exact Firecracker PID
- reject stale same-PID, mixed-record, and malformed-timestamp evidence while preserving the existing bounded diagnostic fallback

## Scope

This fixes post-start Firecracker crashes classified by the existing guest wait-error path. It does not change Firecracker startup-failure diagnosis, guest OOM detection, host memory policy, or sandbox retry policy.

## Validation

- `cargo test -p runner --all-targets --all-features` (`3316 passed`, `24 ignored`; guest inventory `1 passed`)
- staged pre-commit hooks: cargo fmt, rustdoc, Clippy, file-size check, and commitlint

Closes #29488

### DIFF
--- a/crates/runner/Cargo.toml
+++ b/crates/runner/Cargo.toml
@@ -33,7 +33,7 @@ tracing-subscriber = { workspace = true }
 flate2 = { workspace = true }
 futures-util = { workspace = true }
 reqwest = { workspace = true }
-rustix = { workspace = true, features = ["process"] }
+rustix = { workspace = true, features = ["process", "time"] }
 hex = { workspace = true }
 idna = { workspace = true }
 libc = { workspace = true }
--- a/crates/runner/src/executor/agent_run.rs
+++ b/crates/runner/src/executor/agent_run.rs
@@ -31,8 +31,9 @@ use super::diagnostics::{
     StdoutDrainReport, build_agent_env_diagnostics, build_agent_env_key_diagnostics,
     check_host_oom, collect_agent_abnormal_exit_diagnostics, dmesg_indicates_oom,
     drain_stdout_to_file, explicit_enospc_evidence, failure_diagnostic_reports_workload_memory_oom,
-    log_agent_abnormal_exit_env_diagnostics, log_agent_bootstrap_abnormal_exit_diagnostics,
-    log_agent_process_exit_summary, read_guest_error_file, read_guest_failure_diagnostic_file,
+    host_oom_evidence_since_now, log_agent_abnormal_exit_env_diagnostics,
+    log_agent_bootstrap_abnormal_exit_diagnostics, log_agent_process_exit_summary,
+    read_guest_error_file, read_guest_failure_diagnostic_file,
     should_collect_agent_abnormal_exit_diagnostics,
     should_collect_unattributed_sigkill_resource_diagnostics,
     should_log_agent_bootstrap_abnormal_exit_diagnostics,
@@ -1299,6 +1300,7 @@ impl RunControls {
 struct PreparedAgentProcess {
     handle: GuestProcessHandle,
     agent_started_at: Instant,
+    host_oom_evidence_since: super::diagnostics::HostOomEvidenceSince,
     deferred_background_fill: Option<crate::storage_cache::DeferredBackgroundFill>,
     session_restore_diagnostics: Option<SessionRestoreDiagnostics>,
     pre_run_restored_session_identity: Option<RestoredSessionIdentity>,
@@ -2130,6 +2132,7 @@ pub(super) async fn run_in_sandbox_with_process_cancel_timeouts(
     // Guest-agent receives JOB_TIMEOUT as the user execution budget. The
     // sandbox supervisor remains a later hard fallback so guest-agent can
     // terminate the CLI and create a recovery checkpoint first.
+    let host_oom_evidence_since = host_oom_evidence_since_now();
     let t = Instant::now();
     let handle = sandbox
         .start_process(&StartProcessRequest {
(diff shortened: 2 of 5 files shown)


seven332:
## Code Review: PR #29501

**Reviewed head:** `6503e809ad89abb75625652091845e086521060e`

### Summary

The change binds post-start host OOM classification to a monotonic active-run cutoff and requires fresh, same-line raw dmesg evidence for the exact Firecracker PID. It preserves the existing bounded diagnostic and generic-failure fallback without widening the sandbox abstraction.

### Findings

No findings.

### Testing

- Runner rustfmt, all-target/all-feature Clippy, and rustdoc passed.
- Runner all-target/all-feature tests passed (`3316 passed`, `24 ignored`), plus guest inventory (`1 passed`).
- Staged pre-commit hooks passed.
- GitHub CI is still pending. A privileged production-host OOM event was not reproduced locally; coverage uses the captured production record format and negative freshness fixtures.

### Verdict

LGTM

PR #29506, opened 2026-08-26
fix(platform): share uploaded template resources

bingjiez666:
## Summary

- reuse one uploaded-template detail resource across background preloading, card hover, and the detail dialog
- apply rename and visibility responses in place instead of rebuilding the catalog after each mutation
- keep delete tombstones and catalog reconciliation behavior unchanged

## Verification

- `pnpm exec prettier --check apps/platform/src/signals/okou-page/presentation-template-library.ts apps/platform/src/views/okou-page/__tests__/chat-composer-template-picker.test.tsx`
- `pnpm -F @okouai/app lint`
- `pnpm -F @okouai/app exec vitest run src/views/okou-page/__tests__/chat-composer-template-picker.test.tsx -t "keeps loaded uploaded slides mounted while visibility metadata refreshes" --maxWorkers=2`
- `pnpm -F @okouai/app exec vitest run src/views/okou-page/__tests__/chat-composer-template-picker.test.tsx -t "scrubs every uploaded slide and manages an owned template|keeps remaining uploaded cards mounted while delete refresh is pending" --maxWorkers=2`

## Stack

- Base: `main`
- Follow-ups add stable preview asset URL resolution and viewport-aware image loading.


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 2ea53c8fc65f62dff83fefa1bb87ced1ed4d4a64
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Replaces the per-catalog-render `detail$` computed with one composer-scoped keyed resolver, so background preloading, card hover, and the detail dialog share a single uploaded-template detail resource that only reloads when the signed-URL TTL version bumps. Rename and visibility mutations now apply the `200` response in place through a composer-scoped overlay guarded by `updatedAt`, instead of forcing a full catalog rebuild after every mutation.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic: no race conditions, no wrong assumptions | Pass | `presentation-template-library.ts` — the overlay guard `updatedTemplate.updatedAt <= template.updatedAt` self-heals against an in-flight catalog response; the API bumps `updatedAt: nowDate()` on every update (`apps/api/src/signals/routes/presentation-templates.ts`), so a rename can never be swallowed by an equal timestamp. |
| `REVIEW.md` | Security: no injection, secrets, or missing auth on new paths | N/A | No new endpoint, credential, or auth path; the client only re-keys existing `presentationTemplatesContract.get`/`update` calls. |
| `REVIEW.md` | Type safety: no `any`, no broad types | Pass | Adds the explicit `ImportedPresentationTemplateDetailResolver` alias; no `any`, no `as` bridges. |
| `REVIEW.md` | Style and maintainability: no oversized functions, no duplicated logic | Pass | Extracts `createImportedPresentationTemplateDetailResolver` and `createUpdateImportedPresentationTemplate$` out of the already-long factory; comments state WHY (page URLs rotate only on the TTL lifecycle). |
| `REVIEW.md` | Feature regression guard: replaced behavior must carry explicit coverage and containment | Pass | The replaced mutation path keeps its own regression test (visibility change asserts the list and detail request counts are unchanged and the preview element identity is preserved), and the whole surface is contained by `FeatureSwitchKey.PresentationTemplates` (`enabled: false`, staff org hashes). |
| `REVIEW.md` | DB/JSONB gate: persisted contract changes need migration or normalizer coverage | N/A | No `packages/db` schema, contract, or `jsonb` change; client-side signal wiring only. |
| `docs/bad-smell.md` | Zero `any`, no lint/type suppressions, no dynamic imports, no hardcoded config | Pass | Diff adds no suppression comment, no `import()`, and no literal URL or env value. |
| `docs/bad-smell.md` | Fail fast: no fallback or recovery logic hiding a bad state | Pass | The `404 → null` branch is pre-existing contract handling, not new recovery code; the update command still uses `accept(..., [200])` and lets non-200 throw. |
| `docs/fallback.md` | No new undeclared fallback; no negative tests pinning removed behavior | Pass | The diff introduces no cross-version tolerance branch. The `updatedAt` guard is optimistic-overlay conflict resolution between two live sources, not a compatibility path for an old deployed version, so no `Fallbacks` section is required. No test asserts that deleted behavior stays deleted. |
| `docs/fallback.md` | A non-GA feature-switched surface needs no compatibility code | Pass | `PresentationTemplates` is `enabled: false` with staff-only org hashes, so no dual-read, migration, or rollout branch is owed here. |
| `docs/cache.md` (RC-3) | No unbounded-lifetime keyed cache; no manual cache duplicating `computed` memoization | Pass | `detailByTemplateId` lives inside `createImportedPresentationTemplateSignals()`, which is created per composer group via `createComposerUiSignals` → `createComposerSignals`, and is keyed by template id, bounded by the workspace catalog. It is a keyed join, not duplicate memoization: the old `detail$` identity was recreated by the enclosing `computed` on every catalog change, which `computed` memoization cannot dedupe. |
| `docs/cache.md` (RC-4) | State belongs to the narrowest lifecycle that owns it | Pass | Both new stores (`detailByTemplateId`, `internalUpdatedTemplates$`) are composer-scoped and released with the composer, rather than joining the module-scope `deletedPresentationTemplateIds$`. |
| `docs/cache.md` (RC-5) | No duplicate mutable source of truth kept in sync with the server object | Pass | The overlay stores only `title`, `visibility`, `updatedAt` and is superseded as soon as the catalog returns a newer `updatedAt`; the catalog stays authoritative and the realtime `presentationTemplatesChanged` publish after `update` still refetches it. |
| `docs/cache.md` (RC-8) | Symmetric resource lifecycles and abort ownership | Pass | The update command keeps `signal.throwIfAborted()` after the request; the URL-refresh loop and its `AbortSignal` ownership are untouched. |
| `docs/effect.md` | Choose the owner and trigger before the API; derive with `computed`, mutate with a semantic command | Pass | Detail loading stays a `computed(async ...)` invalidated by `detailUrlsVersion$`; the mutation stays one semantic command that applies its own response. No render-phase work is introduced. |
| `.claude/skills/ccstate/SKILL.md` | Do not add a manual cache on top of `computed` memoization | Pass | See RC-3 row: the cache preserves atom identity across catalog rebuilds, which memoization alone cannot do. |
| `.claude/skills/ccstate/SKILL.md` | Reactive async `computed` over imperative fetch-and-store; correct `set` updater usage | Pass | `set(internalUpdatedTemplates$, (updatedTemplates) => [...])` uses the updater form correctly and stores data, not a function. |
| `docs/testing.md` | Integration/page-level tests only; mock at the boundary; no fake timers | Pass | Coverage stays in `chat-composer-template-picker.test.tsx` through `detachedSetupPage` and `context.mocks.api`; no `vi.mock` of internal paths, no direct `fetch` stub, no fake timers. |
| `docs/testing/app-testing.md` | Platform tests enter through `detachedSetupPage` and assert page-visible behavior | Pass | Both updated tests drive the real picker (open, hover, close, reopen) and assert rendered `img` `src`, `loading`, and `fetchpriority`. |
| `docs/testing/testing-external-behavior.md` | Give context from the user surface, observe from the user surface | Pass | Request counts are counted at the MSW HTTP boundary — the externally observable network behavior this PR changes — not on internal signal calls; the visible outcome (renewed slides shown without blanking) is still asserted from the DOM. |
| `docs/testing/anti-patterns.md` | AP-1/AP-5/AP-7: no mock-call assertions as the sole check, no fake timers, no implementation-detail assertions | Pass | `mockNow` is the project's time seam rather than `vi.useFakeTimers()`; every count assertion is paired with a rendered-DOM assertion. |
| `REVIEW.md` | `fix:` commits need a regression test | Pass | `2ea53c8` pins the new invariant end to end: opening the picker, hovering a card, and a metadata-only refresh each keep `primaryDetailRequestCount` at 1, a TTL renewal raises it to exactly 2, and hovering the renewed card adds none. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

- `presentation-template-library.ts:createImportedPresentationTemplateDetailResolver` — `detailByTemplateId` is never pruned when a template is deleted, so a tombstoned id keeps one `Computed` entry until the composer group is released. Bounded by the workspace catalog and released with the composer, so this is not worth extra code today.
- `presentation-template-library.ts:createImportedPresentationTemplateResources$` — the resolver writes to the `Map` from inside a `computed` callback. The write is an idempotent memo-fill, so re-execution is safe, but it is worth keeping in mind if a future change makes the resolver do anything beyond caching.
- In production the API still publishes `presentationTemplatesChanged` after an update, so the catalog is refetched by realtime shortly after the in-place apply. The overlay therefore buys an immediate, non-blanking UI update rather than eliminating the refetch entirely; the durable win is that `detail$` identity survives that refetch.

### Testing

- Coverage: Adequate. `chat-composer-template-picker.test.tsx` covers the shared-resource invariant across all three surfaces (preload, hover, dialog), the metadata-only refresh path, the TTL renewal path, and the unchanged delete/tombstone path.
- Conventions: Compliant. Page-level entry via `detachedSetupPage`, boundary mocking via `context.mocks.api`, no internal `vi.mock`, no fake timers, no assertions on CSS classes or internal state.

Issue #29468, opened 2026-08-26
chore(connectors): handle legacy OAuth scope provenance


seven332:
## Parent objective

Part of #29462, after the requested/granted scope split delivered by #29509.

## Current state and evidence

New built-in OAuth connections persist two distinct facts:

- `oauth_scopes`: the authorization-request snapshot retained for catalog request-drift detection
- `oauth_granted_scopes`: the effective provider grant used by runtime and public readers

#29509 is deployed. The old live API writer population has drained, and the product decision for this cleanup is not to preserve compatibility with retained rollback targets.

The rollout bridge remains on `main` as `connectors_clear_builtin_oauth_granted_scopes_1006`. It clears the explicit grant when `oauth_scopes` changes, and the current writer must restore the grant in a second statement. Runtime and public readers also still use `oauth_granted_scopes ?? oauth_scopes`.

## Problem

For historical built-in rows, `oauth_granted_scopes = NULL` means the provider-reported grant was discarded and cannot be reconstructed. Falling back to `oauth_scopes` permanently presents a requested-scope snapshot as though it were an observed provider grant.

Removing that fallback without defining unknown-grant behavior is also unsafe. Current Gmail and X operation guards treat a missing scope list as a known missing grant and can request reconnection solely because historical provenance is unknown.

## Decisions

- Do not infer or backfill historical provider grants from requested scopes.
- Treat `oauth_granted_scopes = NULL` on built-in OAuth rows as an explicit unknown grant.
- Keep an otherwise healthy historical connection connected; unknown provenance alone must not set or imply `needsReconnect`.
- Distinguish an unknown grant from a known grant that lacks an operation-required scope. Known insufficient grants may fail locally; unknown grants must defer authorization to the provider and surface the provider result accurately.
- Keep the built-in `oauth_scopes` request snapshot for catalog request-drift detection.
- Keep custom connector `oauth_scopes` semantics unchanged.
- Remove the old-writer bridge after the live writer drain without preserving rollback-target compatibility.

## Acceptance criteria

- A migration removes the `connectors_clear_builtin_oauth_granted_scopes_1006` trigger and function.
- Current built-in connection writes persist requested and granted scope facts without the bridge-specific second update.
- Runtime and public projections return an unknown grant for historical built-in rows instead of substituting the requested-scope snapshot.
- Catalog scope-drift detection continues comparing the current request with the stored request snapshot.
[...]

Issue #29462, opened 2026-08-26
bug: align connector OAuth scope semantics with partial grants


seven332:
## Background

Google OAuth granular consent can return an access token with fewer scopes than the connector requested. This has been reproduced with both Gmail and Google Meet: OAuth completion succeeded, but a later API call failed with `403 ACCESS_TOKEN_SCOPE_INSUFFICIENT`.

The product decision is to accept partial grants. A connector may remain connected while individual operations that require an ungranted scope fail with an operation-specific permission error.

This supersedes the rejected direction in #29440, which proposed rejecting partial grants during OAuth completion.

## Problem

The connector scope model currently mixes requested scopes, granted scopes, and connector-wide requirements:

- The shared auth-code callback discards provider-reported scopes and persists catalog-requested scopes.
- The same `oauthScopes` field has different meanings across auth-code, device-auth, external-code, OpenID, and custom OAuth flows.
- Provider adapters historically handled an omitted token-response `scope` inconsistently.
- Built-in refresh results cannot report an explicit changed scope set.
- Catalog and workflow readiness compare catalog requests with grants and report intentional partial grants as `scope-mismatch`.
- Some operation code derives its required permissions from the entire connector request instead of declaring operation-specific requirements.
- Historical rows contain scope values whose provenance cannot be inferred safely.

## Product decisions

- OAuth completion must not reject a valid token solely because the provider granted fewer scopes than requested.
- Requested and granted OAuth scopes are independent facts.
- Catalog `grant.scopes` describes the authorization request, not a universal required-scope contract.
- Persisted granted-scope metadata must reflect the effective token grant when it is authoritative.
- `scope-mismatch` means the current catalog request added a scope that was not requested when the connection was authorized; it does not mean the provider grant is a subset.
- A provider that omits `scope` resolves the effective grant from the exact authorization request it sent, including provider-added scopes.
- A refresh response that explicitly reports scopes replaces the persisted grant; omission preserves the prior grant.
- Scope-dependent behavior belongs at the operation boundary.
- Unknown legacy scope provenance must not force an otherwise valid connection to reconnect.

## Delivery invariant

#29465 and #29466 must land in one atomic PR. Persisting accurate connection-time grants while status still treats grants as prior requests creates a false reconnect prompt; changing status without preserving request snapshots removes catalog scope-update detection.

The atomic change must also tolerate old API readers and writers after its additive migration. Old writers must not be able to leave explicit grant metadata from a previous token after reconnect.

[...]

PR #29509, opened 2026-08-26
fix: separate requested and granted oauth scopes

### LINKED ISSUE #29465: fix(connectors): separate requested and granted OAuth scope facts
## Parent objective

Part of #29462, the connector OAuth partial-grant correctness initiative.

## Problem and evidence

Built-in connector storage has one `oauthScopes` fact but two independent product needs:

- Runtime and operation-specific checks need the effective scopes granted when the current credential connection was authorized.
- Catalog configuration-drift detection needs the scopes requested by the authorization flow that produced the connection.

PR #29509 initially changed the auth-code callback to persist only the provider grant. Existing catalog and workflow readers compare that same value with current catalog requests. A valid partial grant would therefore immediately become `scope-mismatch`, prompt the user to reconnect, and repeat after every intentional partial grant.

Removing `scope-mismatch` is also incorrect: it currently tells users when a connector catalog update adds a scope that was not requested by their existing connection.

## Decision

Requested and granted scopes are separate persisted facts and must become correct atomically in the same PR as #29466.

- A partial provider grant is a connected OAuth connection.
- `scope-mismatch` means the current catalog authorization request contains a scope that was not requested when the connection was authorized.
- It does not mean that granted scopes are smaller than requested scopes.
- Scope-dependent behavior uses the grant observed at connection or reconnection at the operation boundary.

The earlier no-migration plan for this issue is superseded because it knowingly introduced a false status projection between #29465 and #29466.

## Chosen direction

1. Capture catalog requested scopes when auth-code, OpenID, external-code, and device-auth authorization starts.
2. Add explicit granted-scope storage for new built-in token connections.
3. Keep the existing physical `oauth_scopes` value as a requested-scope mirror for new built-in writes so previous API readers preserve catalog-drift behavior during rollout and rollback.
4. Project public/runtime `oauthScopes` from the explicit connection-time provider grant when present, with a legacy fallback for historical rows or old writers.
5. Make catalog status and scope diff compare current catalog requests with the stored requested snapshot.
6. Keep requested snapshots internal; do not add a new public connector response field.
7. Preserve custom connector scope semantics.

## Deployment compatibility

Old API instances can reconnect rows after the additive migration but do not know the new granted-scope column. Their historical legacy-column semantics also differ by OAuth flow. To prevent a stale explicit grant from the previous token:

- a built-in compatibility trigger clears explicit granted scopes whenever the legacy scope column is updated;
- the new writer writes the requested mirror first and the explicit grant second inside the existing credential-replacement transaction;
- an old writer therefore invalidates metadata it cannot maintain, while a new reader safely falls back to that flow's previous legacy behavior.

No version, revision, generation, or public protocol field is required. Active authorization sessions created before the migration remain completable by falling back to the current resolved method when their requested snapshot is absent.

## Alternatives considered

- **Persist grants first and fix status later:** rejected because the first PR independently creates a user-visible false mismatch.
- **Remove scope mismatch:** rejected because catalog scope additions would no longer request reconnection.
- **Add two fields without an old-writer bridge:** rejected because a draining or rolled-back API can leave the new grant stale after reconnect.
- **Feature-switch the projection:** rejected because it does not solve stale cross-version writes and adds operational state to a data-semantics correction.
- **Also redesign every built-in refresh provider:** deferred to #29582 because refresh has a separate cross-provider result and persistence contract; adding it here would expand an already atomic connection/status correction.

## Acceptance criteria

- Built-in auth-code, OpenID, external-code, and device-auth connections record the authorization request snapshot and connection-time effective provider grant as separate facts.
- Auth-code completion succeeds for partial, supplemental, and provider-normalized omitted-scope results.
- A partial grant remains `connected` in connector catalog and workflow readiness.
- A later catalog requested-scope addition still produces `scope-mismatch` and an added-scope diff.
- Public/runtime `oauthScopes` reflects the explicit connection-time effective grant for new built-in connections.
- An old-style built-in reconnect cannot leave explicit granted scopes from the previous token.
- Previous API readers and rollback code can continue reading and writing the migrated schema safely.
- Requested snapshots are captured at authorization start; old active sessions remain compatible.
- Custom OAuth storage and refresh semantics are unchanged.
- Integration tests prove partial grant, catalog drift, and old-writer invalidation independently.
- #29465 and #29466 are closed by the same PR.

## Non-goals

- Rejecting partial grants.
- Maintaining an explicitly changed provider scope set across built-in token refresh; that remains #29582.
- Inferring authoritative grants for historical rows; that remains #29468.
- Completing every operation-specific permission check; that remains #29467.
- Removing the temporary rollout bridge before old API and rollback windows drain.

## Relevant code

- `turbo/packages/db/src/schema/connector.ts`
- `turbo/packages/db/src/schema/connector-oauth-state.ts`
- `turbo/packages/db/src/schema/connector-external-code-session.ts`
- `turbo/packages/db/src/schema/connector-oauth-device-authorization-session.ts`
- `turbo/apps/api/src/signals/services/connector-connection-write.service.ts`
- `turbo/apps/api/src/signals/services/connector-data.service.ts`
- `turbo/apps/api/src/signals/services/connector-catalog-external-reader.service.ts`
- `turbo/apps/api/src/signals/routes/connectors-slug-callback.ts`

### LINKED ISSUE #29466: fix(connectors): preserve scope drift without rejecting partial OAuth grants
## Parent objective

Part of #29462 and the atomic delivery owned by #29465.

## Problem and evidence

Catalog and workflow status currently rename catalog-requested OAuth scopes to `requiredScopes` and compare them with provider-granted scopes. Those sets are allowed to differ: Google granular consent can return a valid partial grant, and providers can add, alias, or canonicalize scope strings.

Treating the provider grant as the prior request causes valid connections to appear as `scope-mismatch` and repeatedly asks users to reconnect. Removing the status entirely would lose the valid signal that a catalog configuration update now requests additional permissions.

## Decision

`scope-mismatch` is authorization-request drift, not grant completeness.

- Current scopes come from the current catalog authorization request.
- Stored scopes for this comparison come from the authorization-request snapshot that produced the connection.
- A mismatch exists only when the current request adds at least one scope absent from that snapshot.
- Provider-granted scopes remain independent and are used by operation-specific behavior.

This status change must ship in the same PR as #29465. Neither issue is an independently safe intermediate deployment.

## Chosen direction

- Rename internal catalog facts from `requiredScopes` to `requestedScopes`.
- Compare current requested scopes with the stored requested snapshot for catalog status and scope diff.
- Keep the public `scope-mismatch` status and existing scope-diff response shape deployment-compatible.
- Let workflow readiness consume the corrected catalog status without adding a second scope policy.
- Keep requested snapshots internal to the API.

## Acceptance criteria

- An otherwise valid partial grant remains `connected` in catalog status and workflow readiness.
- A catalog update that adds a requested scope still produces `scope-mismatch`.
- Scope diff reports changes between the old and current authorization requests, not differences between requested and granted scopes.
- Removing a requested scope does not by itself mark the connection mismatched.
- Internal code and tests call catalog scopes requested rather than globally required.
- Public response variants remain compatible with existing browser clients.
- Integration tests cover both an intentional partial grant and a later catalog scope addition.
- This issue and #29465 are closed by the same PR.

## Slice boundary

- Storage, authorization-start snapshots, rollout compatibility, and effective-grant projection are implemented by the same atomic PR under #29465.
- Operation-specific permission checks remain #29467.
- Historical unknown grant provenance remains #29468.

## Relevant code

- `turbo/apps/api/src/signals/services/connector-catalog-external-reader.service.ts`
- `turbo/apps/api/src/signals/services/workflow-connector-readiness.service.ts`
- `turbo/apps/api/src/signals/services/connector-data.service.ts`
- `turbo/packages/connectors/src/connector-auth-method.ts`

seven332:
## Summary

- persist the authorization-start requested scope snapshot separately from the effective provider grant established by built-in OAuth connection or reconnection
- use connection-time grants for public/runtime behavior while preserving requested scopes for catalog drift and `scope-mismatch`
- keep rolling and rollback API versions safe by retaining a legacy requested-scope mirror for new writes and invalidating grants that an old writer cannot maintain
- cover auth-code, OpenID, external-code, device-auth, GitHub App setup, catalog status, scope diff, old-writer behavior, method replacement, and browser API mock semantics

## Behavior

- intentional partial provider grants remain connected instead of immediately producing `scope-mismatch`
- a later catalog scope addition still produces `scope-mismatch` because it is compared with the original authorization request
- new built-in connections expose the effective grant observed at connection or reconnection to operation-specific consumers
- active authorization sessions created before the migration remain completable through the legacy fallback
- external-code completion and GitHub App setup use the authorization-start snapshot when a provider derives granted scopes from the requested set
- GitHub App installation remains available when the connector catalog is temporarily unavailable; setup callbacks without a stored snapshot retain the current-catalog compatibility fallback
- replacing OAuth credentials with a manual or no-auth method explicitly clears the old grant independently of the temporary rollout trigger
- browser API mocks model requested snapshots independently while preserving seeded catalog-drift scenarios
- custom connector OAuth scope behavior is unchanged

## Deployment compatibility

- new built-in writes retain the existing physical `oauth_scopes` column as the requested-scope mirror, so old readers preserve their previous catalog-drift behavior
- a temporary database trigger clears explicit grants when an old API updates the legacy scope column, preventing stale grant metadata for a newly issued token regardless of that old flow's historical legacy-column semantics
- the new API writes the requested mirror while clearing the prior grant, then writes the explicit effective grant within the existing credential replacement transaction
- new GitHub App setup state fields use an independent signature, so old callbacks ignore them and new callbacks accept old in-flight states through the current-catalog fallback
- cleanup of the rollout bridge and historical provenance remains tracked by #29468

## Follow-up boundary

- the shared built-in refresh contract currently cannot carry a provider-reported scope change
- #29582 will update explicit grant metadata when refresh reports scopes and preserve it when refresh omits scopes
- keeping that cross-provider refresh-contract change separate avoids expanding this atomic persistence/status correction

## Validation

- `pnpm format`
- `pnpm turbo run lint`
- `pnpm check-types`
- `pnpm knip`
- `pnpm -F @okouai/db test:migration-consistency`
- `pnpm -F @okouai/connectors test -- --maxWorkers=4 --silent=passed-only` (1,077 passed)
- focused API OAuth and catalog suites (183 passed)
- `pnpm -F api exec vitest run src/signals/routes/__tests__/integrations-github-files.test.ts src/signals/routes/__tests__/cron-connector-catalog.test.ts src/signals/routes/__tests__/integrations.bdd.test.ts --maxWorkers=4 --silent=passed-only` (160 passed)
- `pnpm -F api exec vitest run src/signals/routes/__tests__/connectors.bdd.test.ts -t "rolls back failed OAuth-to-manual credential replacement before a successful retry" --maxWorkers=4 --silent=passed-only` (1 passed)
- `pnpm -F api exec vitest run --maxWorkers=4 --silent=passed-only` (3,548 passed)
- `pnpm -F @okouai/app exec vitest run --maxWorkers=4 --silent=passed-only` (2,083 passed)

Closes #29465
Closes #29466

Parent context: #29462


seven332:
## Code Review: PR #29509

**Reviewed head:** `37a8c3005f14f0d41a77bdebce873fec8c36f9e5`

### Summary

The auth-code callback now persists the provider-boundary effective scopes, aligning it with device-auth, external-code, and OpenID writers. The regression coverage exercises partial, supplemental, and omitted-scope results through the public API and real database path.

### Findings

No findings.

### Testing

API lint, repository type checks, Knip, 76 focused connector tests, and the full API suite (215 files, 3532 tests) passed. Browser and runner E2E were not run because this change does not alter those interfaces.

### Verdict

LGTM

PR #29523, opened 2026-08-26
feat(cli): show connector accounts used by current run

### LINKED ISSUE #29436: feat(cli): show connector accounts selected for the current run
## Background

Parent issue #22832 introduces multiple concrete accounts for one logical connector target. The completed per-thread picker in #27694 lets a thread choose an exact account. Run admission resolves that preference, including fallback, and pins the effective concrete source (`connectors.id`) into the execution context used by the Runner.

Agent instructions already direct the agent to inspect connectors with `okou connector list` and `okou connector status <slug>` before use.

## Problem and evidence

The CLI does not report the connector account actually selected for its current run.

Current CLI inspection is owner-state based:

- `/api/connector-catalog/status` reads the user's live target/default projection;
- it does not use the exact source selected for the active sandbox;
- therefore a run using thread-selected account B can be described as using current default account A.

This can cause an agent to act on the wrong repository, workspace, tenant, or provider account.

The exact selected source is already present in the live execution path:

- the API sends `connectorRuntimeTargets` with each target's optional exact `sourceId` to the Runner;
- execution firewall entries independently carry that `sourceId`;
- the Runner registers those firewalls in the source-IP-keyed proxy registry before the agent starts;
- the mitm addon copies the matched firewall's `sourceId` into `matchedFirewall.sourceId` and calls `/api/webhooks/agent/firewall/auth` with the sandbox token;
- the API resolves and injects auth material for that exact source.

The Runner/proxy therefore already knows the account used by the sandbox. The defect is only that the CLI process inside the guest cannot safely read a bounded projection of this existing run-local fact.

## Requirement

Make connector inspection inside an agent environment report the exact account bound to the current run.

This is run-scoped introspection only. It does not let the CLI choose, switch, add, reconnect, rename, default, or delete accounts.

## Decision

Use the existing Runner execution registration as the only selection source and bridge it to the CLI through a filtered per-run private file.

Before starting guest-agent, the Runner will write a versioned projection derived directly from `ExecutionContext.connectorRuntimeTargets`. Each entry contains only:

- logical built-in or custom target;
- exact connection ID, or null when no source was admitted.

The Runner will add only the file path to the already filtered user-environment payload after untrusted runner-owned keys have been removed. Existing guest-agent child-environment pass-through then makes the path available to Claude, Codex, and Pi without adding a variable-length bootstrap value or a new guest-agent protocol field.

The CLI will treat this local mapping as immutable admission truth. It may batch-enrich exact `(target, connectionId)` pairs through a new bounded read-only connector-account API to show current safe display identity and health. That API does not accept a run ID, infer a selection, substitute a default, or become an authorization input.

No connector-account snapshot will be added to `agent_runs`. No proxy registry, firewall, auth-injection, runtime-sync, or thread-selection behavior will change.

## Rationale

- It reuses the exact `sourceId` already driving auth injection instead of creating a second run/account truth.
- A run-specific file preserves the admitted ID across later thread/default changes and runtime removal of an unavailable custom firewall.
- The existing private-file plus filtered child-environment transport avoids JSON growth in process argv/environment and works with an older guest-agent image.
- A bounded batch projection avoids one API request per selected connector as connector counts grow.
- Selection and current metadata stay separate: loss or deletion of current metadata cannot replace the pinned ID with a sibling.
- The projection is informational only. A sandbox-local file cannot grant credentials or capabilities because firewall/auth enforcement remains server-side and Runner-controlled.

### Existing exact-account GET per selected target

This also preserves Runner-local truth but makes one backend request per source-bearing target. Bounded concurrency does not reduce request count, so it is not the preferred list path as connector scopes grow.

### Proxy-local introspection endpoint

Current registry `connectorRuntimeTargets` drop `sourceId`, and current custom firewalls can be removed by runtime synchronization. Preserving admission identity would require new registry state plus Python request interception and source-IP response handling adjacent to secret-bearing registry data. That is a broader and riskier path for immutable launch-time information.

### Full JSON in an environment variable

This reintroduces variable-length data at the exec boundary, copies it to descendants, and conflicts with the repository's existing private-file transport for larger payloads.

### Persist another snapshot on `agent_runs`

This was the original proposal and is superseded. It duplicates data already owned by the Runner, adds a migration and consistency boundary, and solves historical/post-run use cases outside this active-run introspection requirement.

### Status quo

The CLI can continue reporting a different live default from the account actually used for auth. That leaves an agent unable to verify the repository, tenant, or provider identity on which it is operating.

## CLI behavior

- In a run-bound context, `okou connector list` and `okou connector status <slug>` identify the account selected for that run.
- Human output clearly labels the value as the account used by the current run; stable `--json` output exposes the same bounded state.
- If the selected row still exists, the CLI shows its current safe display identity and health while keeping the immutable connection ID distinct from mutable status.
- If the selected row no longer exists, the CLI retains its connection ID and reports metadata unavailable/deleted without selecting another account.
- If no source was admitted, or the Runner cannot provide a valid projection, the CLI reports unknown/unavailable with next-run remediation.
- Outside a run-bound context, existing PAT/session behavior continues to show the live current/default projection.

## Data and authorization boundary

The guest-readable file must not contain credentials, sandbox tokens, encrypted secrets, auth templates, injected variables, OAuth scopes, provider payloads, firewall definitions, policies, mutable display data, or unrelated accounts.

CLI file validation will enforce a known schema version, strict target/UUID shapes, duplicate-target rejection, and a fixed byte cap. The batch API will enforce a fixed item maximum, authenticated organization/user ownership, `connector:read`, and exact target matching. Oversized local projections are chunked into bounded requests.

The local file is self-inspection data in a sandbox controlled by the running workload; it is not trusted for authorization. Tampering can at most affect that sandbox's own display and exact reads of accounts already visible to its authenticated owner. Firewall/auth enforcement never consumes it.

## Compatibility and failure behavior

- **Old Runner + new CLI:** `OKOU_AGENT_ID` identifies run mode; a missing file reports unavailable and never falls back to current default.
- **New Runner + old CLI:** the old CLI ignores the extra path/file and retains prior behavior.
- **New Runner + old guest-agent:** the existing user-env-file and managed child-env pass-through carries the path without a new bootstrap field.
- **New CLI/Runner + old API:** the CLI retains exact local IDs and marks only metadata enrichment unavailable.
- **New API + old CLI/Runner:** the additive endpoint is unused.
- **Projection write failure:** record bounded structured telemetry, omit the path, and continue the otherwise valid run. Introspection fails closed, while existing auth injection remains functional.
- **Sandbox reuse:** use a run-ID-specific path and rewrite every execution, including a valid empty projection, so stale state cannot carry across runs.

No database migration or historical backfill is required.

## Acceptance criteria

- With default account A and a thread that selects account B, both connector inspection commands identify B inside the admitted run.
- If the thread selection or user default changes after admission, the active run continues to identify its pinned account.
- If admission falls back from an unusable explicit account to a usable default, the CLI reports the effective fallback account delivered to the Runner.
- Built-in and user-managed custom HTTP/API and MCP targets use the same exact-source semantics.
- A target with no admitted source is reported unavailable for this run.
- A deleted or otherwise unresolvable pinned source retains its exact connection ID and never resolves to a sibling.
- A legacy Runner, missing/unreadable/malformed/unsupported/oversized projection, duplicate target, or source-less target fails closed without claiming the live default.
- More than one API batch worth of targets remains bounded and deterministic without a request-per-account pattern.
- PAT/session calls outside a run preserve existing behavior.
- Human-readable and JSON output expose no secrets or unrelated accounts.
- Runner integration tests cover exact/source-less/empty projection, fresh and reused sandboxes, protected path injection, write failure, and privacy.
- API integration tests cover exact/missing/mismatched/foreign rows, capability and organization boundaries, response privacy, and maximum size.
- CLI integration tests cover exact, fallback, unavailable, deleted, custom, legacy/missing, malformed, old-API, many-target, and non-run contexts.

## Delivery notes

This is one coherent PR: the Runner file without CLI consumption would not satisfy the user-visible requirement, while CLI consumption without the additive enrichment contract would either lose usable account identity or introduce request-per-account scaling. The components remain independently deployable because every new surface is additive and all overlap states are explicitly handled.

Likely affected areas are `crates/guest-contracts`, Runner executor environment/file handling, connector-account API contracts/routes, and CLI connector list/status plus their integration tests.

## Dependencies and scope

- Parent: #22832.
- #27694 is complete and provides the per-thread preference whose effective admitted result must be visible.
- Account management and final `single-account` retirement remain outside this issue, including #28571.
- No Platform account-management UI or connector lifecycle behavior changes.

## Current status

Research, options, and the corrected Plan Phase are complete. The plan challenge approved this direction. Implementation still requires explicit approval.

seven332:
## Summary

- project each run's exact connector account IDs from the Runner into a versioned private guest file without exposing credentials
- add a bounded exact-account inspection API that enforces owner, organization, target, and `connector:read` boundaries
- make `okou connector list` and `okou connector status` show the accounts used by the current run, including stable JSON and fail-closed compatibility behavior

## Testing

- `cargo clippy -p guest-contracts -p runner --all-targets --all-features -- -D warnings`
- `cargo doc -p guest-contracts -p runner --all-features --no-deps`
- `cargo test -p guest-contracts --all-features`
- `cargo test -p runner --bin runner --all-features`
- `pnpm format`
- affected API contracts, API, and CLI lint/type checks
- API contracts full test suite (620 tests)
- API full test suite on a clean temporary database (3,534 tests)
- CLI full test suite (1,018 passed, 1 skipped)
- `pnpm knip`

Closes #29436
Relates to #22832

### DIFF
--- a/crates/guest-contracts/src/connector_account_context.rs
+++ b/crates/guest-contracts/src/connector_account_context.rs
@@ -1,0 +1,77 @@
+//! Non-secret connector account selection exposed to managed CLI children.
+
+/// Current connector account context schema version.
+pub const SCHEMA_VERSION: u8 = 1;
+
+/// One run's connector account selections.
+#[derive(Clone, Debug, Eq, PartialEq, serde::Deserialize, serde::Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RunConnectorAccountContext {
+    /// Wire schema version.
+    pub schema_version: u8,
+    /// Logical connector targets admitted for the run.
+    pub targets: Vec<RunConnectorAccountTarget>,
+}
+
+/// One logical connector target and its run-pinned account, when available.
+#[derive(Clone, Debug, Eq, PartialEq, serde::Deserialize, serde::Serialize)]
+#[serde(tag = "kind", rename_all = "lowercase")]
+pub enum RunConnectorAccountTarget {
+    /// Built-in connector target.
+    #[serde(rename_all = "camelCase")]
+    Builtin {
+        /// Public connector slug.
+        connector_slug: String,
+        /// Exact connector account selected at admission.
+        connection_id: Option<String>,
+    },
+    /// User-managed custom connector target.
+    #[serde(rename_all = "camelCase")]
+    Custom {
+        /// Custom connector definition id.
+        custom_connector_id: String,
+        /// Exact connector account selected at admission.
+        connection_id: Option<String>,
+    },
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn context_uses_versioned_camel_case_wire_shape() {
+        let context = RunConnectorAccountContext {
+            schema_version: SCHEMA_VERSION,
+            targets: vec![
+                RunConnectorAccountTarget::Builtin {
+                    connector_slug: "github".to_string(),
+                    connection_id: Some("550e8400-e29b-41d4-a716-446655440000".to_string()),
+                },
+                RunConnectorAccountTarget::Custom {
+                    custom_connector_id: "550e8400-e29b-41d4-a716-446655440001".to_string(),
+                    connection_id: None,
+                },
+            ],
+        };
+
+        assert_eq!(
+            serde_json::to_value(context).unwrap(),
+            serde_json::json!({
+                "schemaVersion": 1,
+                "targets": [
+                    {
+                        "kind": "builtin",
+                        "connectorSlug": "github",
+                        "connectionId": "550e8400-e29b-41d4-a716-446655440000"
+                    },
+                    {
+                        "kind": "custom",
+                        "customConnectorId": "550e8400-e29b-41d4-a716-446655440001",
+                        "connectionId": null
+                    }
+                ]
+            })
+        );
+    }
+}
--- a/crates/guest-contracts/src/env.rs
+++ b/crates/guest-contracts/src/env.rs
@@ -181,6 +181,19 @@ pub const USER_ENV_PRIVATE_DIR_NAME: &str = "user-env";
 /// Private runtime filename used by [`USER_ENV_FILE_ENV`].
 pub const USER_ENV_FILENAME: &str = "env.json";
 
+/// Path to the non-secret connector account context for the current run.
+///
+/// The runner adds this key to the filtered user environment after removing
+/// untrusted runner-owned keys. Managed CLI children may read the referenced
+/// file for self-inspection; authorization does not consume it.
+pub const CONNECTOR_ACCOUNT_CONTEXT_FILE_ENV: &str = "OKOU_CONNECTOR_ACCOUNT_CONTEXT_FILE";
+
+/// Private runtime subdirectory used by [`CONNECTOR_ACCOUNT_CONTEXT_FILE_ENV`].
+pub const CONNECTOR_ACCOUNT_CONTEXT_PRIVATE_DIR_NAME: &str = "connector-account-context";
+
+/// Private runtime filename used by [`CONNECTOR_ACCOUNT_CONTEXT_FILE_ENV`].
+pub const CONNECTOR_ACCOUNT_CONTEXT_FILENAME: &str = "context.json";
+
 /// Path to the private runner-owned run payload JSON file.
 ///
 /// Large prompt-like and configuration payloads use this file instead of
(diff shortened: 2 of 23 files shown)


seven332:
## Code Review: PR #29523

**Reviewed head:** `d359b987338f2374ebe00bce5b2946bf1824dd7f`

### Summary

The Runner-owned projection, exact-account enrichment API, and run-aware CLI output follow the approved design. Selection truth remains local and immutable, metadata is optional, and old/missing components fail closed without choosing a current default.

### Findings

No findings.

### Testing

Rust formatting, Clippy, docs, guest-contracts tests, and runner tests passed. Turbo formatting, affected lint/type checks, API contracts/API/CLI test suites, and Knip passed. The API full suite passed all 3,534 tests on a clean temporary database. A production sandbox smoke test was not run.

### Verdict

LGTM

PR #29524, opened 2026-08-26
test(python): cover anthropic accounting context guards

### LINKED ISSUE #29515: test(python): cover incomplete Anthropic accounting context guards
## Summary

Add regression coverage for the fail-closed reporting-context guard in `crates/runner/mitm-addon/src/anthropic_accounting.py`.

`report_incomplete()` returns before retaining or enqueueing telemetry when the flow has no run ID or when the usage reporting context lacks a sandbox token or API URL. The current Python suite never exercises that branch.

## Evidence

- The audit sample was locked on `main` at `1110a2b176a46bd56deb3fd7b8a7cdd7ce811756`; the complete mitm-addon suite passed with `6636 passed`.
- During the final drift check, `main` advanced to `5954906bf61d374ebe4029453a2ace7ce16870e8` through unrelated TypeScript API changes. The checkout was fast-forwarded and the candidate module and related tests were unchanged.
- On current `main`, all 24 dedicated Anthropic SSE usage tests pass, but branch coverage for `anthropic_accounting.py` remains 97%, with line 57 and branch `56 -> 57` missing.
- Every incomplete Anthropic SSE accounting test uses `_anthropic_messages_sse_flow()`, which installs both a non-empty run ID and complete reporting context.
- Existing missing-token and missing-API-URL coverage in `test_claude_output_timing.py` exercises a different timing-retention path and cannot detect a regression in `anthropic_accounting.report_incomplete()`.
- The normal `mitm_addon.response()` and `mitm_addon.error()` hooks return before SSE finalization when the run ID is missing. Clearing the run ID before entering either hook therefore tests only the outer unregistered-flow guard and does not exercise the local guard in `report_incomplete()`.
- The SSE parser finalizer is registered while the response context is complete and is the lifecycle boundary that invokes `report_incomplete()`. A missing-run-ID regression test must reach that finalizer after setup rather than being short-circuited by the outer hook.

Without regression coverage at the correct lifecycle boundary, removing or weakening the guard could allow incomplete flows to acquire a buffered-report lease and attempt to retain or enqueue unattributed accounting telemetry without failing the suite.

## Acceptance criteria

- Cover missing run ID, missing sandbox token, and missing API URL for incomplete Anthropic SSE accounting.
- Ensure the missing-run-ID case reaches the registered incomplete-SSE accounting finalizer instead of passing through the outer unregistered-flow hook.
- Assert that each case emits no accounting webhook request and acquires no retained buffered-report ownership.
- Assert that a later runner flush cannot resurrect a report from any rejected case.
- Keep the existing complete-context accounting behavior covered.
- The relevant new case fails when its corresponding early-return guard is removed or weakened.

seven332:
## Summary

- cover missing run ID at the registered Anthropic SSE finalizer boundary so the accounting guard itself is exercised
- cover missing sandbox token and API URL through normal response finalization
- prove rejected accounting owns no buffered or admitted report before or after runner flush

## Scope

Test-only change. No production behavior, API, persisted state, dependency, protocol, or rollout changes.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_model_provider_sse_usage_anthropic.py` — 27 passed
- `uv run --no-sync python -m pytest tests/` — 6639 passed
- mutation check: temporarily removing the target guard made the run ID, sandbox token, and API URL regression cases fail

Closes #29515

### DIFF
--- a/crates/runner/mitm-addon/tests/test_model_provider_sse_usage_anthropic.py
+++ b/crates/runner/mitm-addon/tests/test_model_provider_sse_usage_anthropic.py
@@ -15,6 +15,7 @@ import anthropic_accounting
 import body_decoding
 import flow_metadata_keys as metadata_keys
 import mitm_addon
+import response_streaming
 import usage
 from tests.flow_helpers import response_stream
 from tests.jsonl_log_helpers import read_jsonl_entries_after_flush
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #29524

**Reviewed head:** `25f5e0d2047cd1711247aed5d9a22ff5f3f47fc7`

### Summary

The test-only change covers all three incomplete Anthropic accounting context guards at lifecycle boundaries that reach the target code. The run-ID case registers and feeds the SSE parser before clearing metadata and entering its stored finalizer; token and API URL cases stay on normal response finalization. Every case proves no telemetry or retained/admitted ownership before and after runner flush.

### Findings

No findings.

### Testing

Ruff format/check, basedpyright, the 27-test Anthropic SSE module, and the complete 6639-test mitm-addon suite passed. A temporary guard-removal mutation made all three new regression cases fail as expected. Codecov also reports all modified coverable lines covered. Unrelated Turbo and Rust suites were not run because the diff changes one Python test file and no cross-language input.

### Verdict

LGTM

PR #29525, opened 2026-08-26
perf(api): collapse media model snapshot reads

### LINKED ISSUE #29511: perf(api): collapse duplicate media-model run snapshot reads
## Problem and evidence

#24203 targets an end-to-end `api_to_spawn` p90 of one second. Investigation of the broad first-dispatch hypothesis found a smaller, concrete source of avoidable work in the common run-preparation path: media-model snapshot resolution reads the same rows multiple times in sequence.

`prepareRunContext` currently resolves the run's video model and image model separately. Each resolver reads one column from `chat_threads`; when the thread value is absent or no longer in the catalog, each resolver separately reads one column from `org_members_metadata` before applying the catalog default.

In the latest 24-hour production trace window:

| Query pair | Traces | Combined latency p50 | Combined latency p90 | Combined latency p95 |
| Thread video + image reads | 5,900 | 6.00 ms | 18.21 ms | 37.14 ms |
| Member video + image reads | 1,358 | 5.02 ms | 12.49 ms | 17.50 ms |
| All media reads when member fallback is needed | 1,399 | 11.76 ms | 32.98 ms | 48.01 ms |

The four query shapes almost always acquired an already-idle database connection. This is duplicate database round-trip cost, not evidence that PostgreSQL connection establishment owns the first-dispatch gap.

The earlier broad first-dispatch framing is therefore not a safe PR boundary: process-level database traces cannot currently be joined to dispatch ordinals, and no single cold-only owner has been proven. This issue is narrowed to the exact repeated-read problem above and does not claim to solve the full first-dispatch gap.

## Chosen direction

- Replace the two run-specific media resolvers with one combined resolver at the existing run-context preparation boundary.
- Read both selected media-model columns from `chat_threads` in one query.
- Validate and resolve video and image independently.
- Only when either field remains unresolved, read both member default columns from `org_members_metadata` in one query.
- Preserve the current per-field precedence exactly: valid thread selection, then valid member selection, then catalog default.
- Skip the member query when both thread selections are valid.
- Keep the observation point unchanged rather than reusing values read earlier during chat-session resolution.

This reduces the path from as many as four sequential queries to at most two without changing schemas, API contracts, queue payloads, runner protocols, authorization, or persistence.

## Alternatives considered

- Run the existing resolvers with `Promise.all`: reduces serial wall time but retains duplicate reads and increases concurrent pool demand.
- Reuse media fields from chat-session resolution: can eliminate another query, but moves the snapshot observation point and changes behavior under concurrent thread updates.
- Use one joined or subquery-based SQL statement: saves another round trip but reads member defaults for the common case where both thread selections are valid and adds unnecessary query complexity.
- Keep the status quo: leaves measured redundant work in every chat run.

## Constraints and risks

- Video and image fallback must remain independent; one valid thread selection must not prevent the other field from using member or catalog fallback.
- Missing threads, missing member metadata, null selections, and catalog-retired selections must preserve their current behavior.
- The resolver must not read member metadata when both thread selections are currently valid.
- The database read must remain at the current run-snapshot point to avoid a concurrency behavior change.
- No new telemetry, cache, schema, compatibility layer, or initialization mechanism is required.

## Acceptance criteria

- [ ] A thread-bound run performs at most one thread media-selection query and, only when needed, one member media-default query.
- [ ] A run without a chat thread skips the thread query and performs at most one member media-default query.
- [ ] Valid thread selections for both fields skip the member query.
- [ ] Video and image independently preserve thread -> member -> catalog-default precedence for null, missing, and catalog-retired values.
- [ ] Real API route integration coverage exercises mixed precedence, including one valid thread field while the other falls back.
- [ ] Applicable formatting, lint, type, Knip, and API integration checks pass.
- [ ] After deployment, the old single-column query pairs disappear from traces and `prepare_context` plus end-to-end latency show no regression; record the observed effect without requiring this small change alone to satisfy the parent issue's one-second target.

## Relationships

- Parent objective: #24203
- Process residency telemetry: #28620
- Media-model feature-switch removal: #29430
- Connector catalog cold hydration (separate): #28275

seven332:
## Summary

- replace the separate video and image run-snapshot resolvers with one combined media-model resolver
- read both media columns from the thread in one query and read both member defaults in one conditional fallback query
- preserve independent video/image thread -> member -> catalog-default precedence and the existing snapshot observation point
- cover both mixed-precedence directions through the real chat-event route flow

## Problem

Run preparation resolved video and image models sequentially. Each resolver queried the same `chat_threads` row, and fallback could query the same `org_members_metadata` row twice more. Production traces found 5,900 paired thread reads in the latest 24-hour window, with an 18.21 ms combined p90; fallback media reads reached a 32.98 ms combined p90.

## Implementation

`resolveMediaModelsForRun` now:

1. selects both thread media columns once when the run owns a chat thread;
2. returns immediately when both stored values remain valid catalog models;
3. otherwise selects both member defaults once and fills only unresolved fields;
4. applies the existing per-field catalog defaults last.

Threadless runs skip the thread query. Missing rows, null selections, and catalog-retired selections retain their previous fallback behavior.

## Explicit exclusions

- no schema or persisted-contract changes
- no API, queue payload, or Runner protocol changes
- no cache, raw SQL, feature switch, or new telemetry
- no attempt to solve the broader first-dispatch cold-start attribution in #24203

## Validation

- `pnpm format`
- `pnpm -F api lint`
- `pnpm -F api check-types`
- `pnpm -F api exec vitest run src/signals/routes/__tests__/chat-events.bdd.test.ts --maxWorkers=1 --silent=passed-only` — 127 passed
- `pnpm knip --workspace api`
- `git diff --check`
- pre-commit hook: full Turbo Knip and check-types — 14/14 type-check tasks passed

## Follow-up verification

After deployment, compare a fixed containing API revision in `vm0-traces-prod`: the old single-column query pairs should disappear, the combined member query should remain conditional, and prepare-context/end-to-end latency plus failure rate should not regress.

Closes #29511

Parent context: #24203


seven332:
## Code Review: PR #29525

**Reviewed head:** `2d53ff31bd226335a800a8f8b1c93a9e355b37c7`

### Summary

The combined resolver removes the measured duplicate thread/member round trips while retaining independent video and image fallback, the existing run-context observation boundary, and current persisted outputs. The implementation stays API-local and introduces no compatibility, authorization, schema, queue, or Runner changes.

### Findings

No findings.

### Testing

- API lint and type checks passed.
- `chat-events.bdd.test.ts`: 127/127 passed, including both mixed-precedence directions.
- API Knip, full formatting, diff checks, and the pre-commit full Turbo Knip/type checks passed.
- Production query-shape and latency comparison remains a post-deployment verification item, not a pre-merge correctness gap.

### Verdict

LGTM

PR #29526, opened 2026-08-26
fix(platform): keep thread account selection in a menu

### LINKED ISSUE #29510: bug: keep thread account selection out of connector popover navigation
## Background

The chat composer Connectors popover exposes a per-thread account control when an authorized connector has multiple accounts. The current implementation replaces the primary connector list with an account-selection page inside the same popover and reserves a fixed height to reduce size changes between the two pages.

## Problem and evidence

The replacement-page design couples two independent layouts. As soon as any connector exposes the account control, the primary popover adopts the account page's fixed 25rem container. The primary page still lays out its connector rows and fixed `Add connectors`, `Cloud browser`, and `Your computer` sections according to their own content height, while the shared surface hides overflow. The bottom sections can therefore be clipped and become unreachable.

The design also makes a row-level account choice feel like navigation to another window and does not clearly explain that the choice applies only to the current thread, not the connector's global default account.

The account action currently has its own border and gives `UserCheck` a stronger foreground color, while the adjacent Permission action is a borderless quiet icon button. These peer row actions appear to have different hierarchy even though neither is the primary row action.

The single-account layout is not affected because it does not create an account control or enter the shared two-page geometry.

## Decision

Do not replace the primary Connectors popover with a secondary account page. The account button in a connector row should open an anchored account-selection menu. The primary popover must retain its existing layout and dimensions while the menu is open.

The menu must visibly state `Account for this thread`. Selecting an account stores a per-thread override; selecting `Use default · <account>` clears that override and follows the connector's current default account. Changing the connector's global default remains part of Connector account management and must not be offered here.

The account and Permission actions should use the same borderless quiet-icon treatment and the same default muted foreground color. Account mode remains distinguishable through the `User` and `UserCheck` icon variants, accessible labels, and Tooltip text rather than a persistent border or stronger default color.

## Acceptance criteria

- Clicking the account button opens an anchored account-selection menu without replacing, resizing, or navigating away from the primary Connectors popover.
- The menu visibly explains that the selection applies to the current thread.
- `Use default` includes the effective default account name and is visually distinct from selecting a specific account.
- Selecting a specific account writes only the current thread's connector account override; selecting `Use default` clears it.
- The menu supports the existing loading, unavailable, empty, pagination, and conditional-search states within its own bounded, scrollable surface.
- Closing the menu leaves the user in the Connector list; no Back action is present.
- `Add connectors`, `Cloud browser`, `Your computer`, and computer host controls remain fully visible and usable in the primary popover.
- The account and Permission row actions have matching borderless quiet-button styling, sizing, default color, hover, and focus behavior. `UserCheck` receives no stronger default color.
- Connectors with zero or one account retain their current row behavior and do not show the account control.
- Keyboard focus, radio semantics, and accessible labels distinguish following the default from a per-thread override.

## Out of scope

- Changing Connector account management or the workflow for setting the global default account.
- Changing connector account APIs, persistence, or execution-time fallback behavior.
- Showing an account control for connectors with fewer than two accounts.

## Relevant code

- `turbo/apps/platform/src/views/okou-page/chat-composer.tsx`
- `turbo/apps/platform/src/views/okou-page/__tests__/chat-composer-connector-connect.test.tsx`

seven332:
## Summary

- keep the Connector popover intact while account selection opens in an anchored menu
- label account selection as thread-specific and close only the account menu after a selection
- keep `User`, `UserCheck`, and Permission actions borderless with the same quiet default color
- preserve account loading, search, pagination, empty, and unavailable states

## Scope

- no API, database, connector runtime, or account resolution changes
- zero-account and single-account Connector rows remain unchanged

## Validation

- `pnpm format`
- `pnpm -F @okouai/app lint`
- `pnpm -F @okouai/app exec vitest run src/views/okou-page/__tests__/chat-composer-connector-connect.test.tsx --maxWorkers=1` (17 tests)
- pre-commit full-repository Knip and TypeScript checks
- local browser verification with two Test app accounts

Closes #29510

### DIFF
--- a/turbo/apps/platform/src/i18n/locales/de-DE/common.json
+++ b/turbo/apps/platform/src/i18n/locales/de-DE/common.json
@@ -1929,13 +1929,13 @@
       "yourComputer": "Ihr Computer"
     },
     "connectors": {
+      "accountForThread": "Konto für diesen Thread",
       "add": "{{connectorName}} hinzufügen",
       "addConnectors": "Integrationen hinzufügen",
       "authorizationFailed": "{{connectorName}} verbunden, konnte aber nicht für {{agentName}} autorisiert werden",
       "authorized": "{{connectorName}} verbunden und autorisiert für {{agentName}}",
       "available_one": "Verfügbare Integration ({{count}})",
       "available_other": "Verfügbare Integrationen ({{count}})",
-      "back": "Zurück",
       "configurePermissions": "Konfigurieren Sie {{connectorName}}-Berechtigungen",
       "customAuthorizeDescription": "Prüfen und verbinden Sie diese benutzerdefinierte Integration und autorisieren Sie sie für den Agenten.",
       "defaultAccountFor": "{{connector}} · Standardkonto wird verwendet: {{account}}",
--- a/turbo/apps/platform/src/i18n/locales/en-US/common.json
+++ b/turbo/apps/platform/src/i18n/locales/en-US/common.json
@@ -1929,13 +1929,13 @@
       "yourComputer": "Your computer"
     },
     "connectors": {
+      "accountForThread": "Account for this thread",
       "add": "Add {{connectorName}}",
       "addConnectors": "Add connectors",
       "authorizationFailed": "{{connectorName}} connected but could not be authorized for {{agentName}}",
       "authorized": "{{connectorName}} connected and authorized for {{agentName}}",
       "available_one": "Available connectors to connect ({{count}})",
       "available_other": "Available connectors to connect ({{count}})",
-      "back": "Back",
       "configurePermissions": "Configure {{connectorName}} permissions",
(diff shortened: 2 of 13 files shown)


seven332:
## Code Review: PR #29526

**Reviewed head:** `2a652d12d1fc1c3e5aa24155ed5d37a75aac68f0`

### Summary

The anchored thread-account menu removes the connector-popover replacement page while preserving the existing account data and mutation paths. The primary Connector list stays visible, account scope is explicit, and the `User`, `UserCheck`, and Permission actions share the same quiet borderless style.

### Findings

No findings.

### Testing

Platform format, lint, type checks, focused Vitest (17 tests), Platform Knip, and pre-commit full-repository Knip/type checks passed. Local browser verification covered two accounts, default and explicit selection, parent-menu persistence, unclipped computer actions, and matching icon styles. Remote CI was pending at review time.

### Verdict

LGTM

PR #29558, opened 2026-08-26
fix(python): protect refreshed auth cache entries from late 401s

### LINKED ISSUE #29538: bug(python): prevent late 401 responses from evicting refreshed auth entries
## Problem

The firewall-auth response path invalidates the cache by auth identity only, so a delayed 401 can evict a newer credential that the failing request never used.

Relevant code:

- crates/runner/mitm-addon/src/auth.py:315-321 stores only FirewallAuthCacheKey on the flow.
- crates/runner/mitm-addon/src/mitm_addon.py:1875-1883 unconditionally clears the current entry for that key on every 401.
- crates/runner/mitm-addon/src/firewall_auth_cache.py:158-162 clears state.cache without checking which cache entry supplied the request.
- crates/runner/mitm-addon/src/firewall_auth_cache.py:146-155 suppresses a second force-refresh marker during the cooldown, but does not protect the newly cached entry from stale invalidation.

A concrete interleaving is:

1. Requests A and B both receive token v1 from the same cached auth entry.
2. A returns 401, clearing v1 and marking the next auth fetch for forced refresh.
3. Request C consumes the marker; the forced fetch completes and caches token v2. The cooldown timestamp is now active.
4. B then returns its delayed 401, even though it was sent with v1. The response handler clears the current v2 cache entry. The cooldown suppresses another force-refresh marker.
5. The next request must call /firewall/auth again. A burst of delayed v1 responses can repeatedly evict refilled v2 entries, adding authentication latency and webhook load and potentially contributing to auth-fetch admission saturation.

The current cooldown correctly prevents duplicate OAuth refreshes, but it cannot distinguish a 401 for an obsolete entry from a 401 for the entry currently in the cache.

## Acceptance criteria

- Associate each resolved auth result with a non-secret opaque cache-entry generation or equivalent identity, and retain that value on the flow that uses it.
- On 401, clear cached headers and request force refresh only if the currently owned cache entry is the same entry that supplied the failing request.
- Preserve the existing behavior for the first current-entry 401, force-refresh cooldown, same-key fetch coalescing, and registry-generation eviction.
- Add a deterministic regression test for the A/B/C interleaving above: after v2 is cached, B's late v1 401 must not remove v2 or schedule another refresh.
- Do not place raw credentials or reversible credential material in flow metadata, logs, or cache-comparison tokens.

seven332:
## Summary

- attach a credential-free process-local identity to each cached firewall auth result
- invalidate a cached auth result on `401` only when the response belongs to that exact entry
- fail closed when resolved firewall auth is missing a valid cache entry identity
- cover concurrent cache sharing and the late-`401` refresh race through request/response integration tests

## Scope

This changes only the runner mitm addon's in-process firewall auth cache and flow metadata. It does not change persisted state, runner protocols, external APIs, credentials, or billing behavior.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (6655 passed)
- `uv run --no-sync python -m pytest tests/test_request_handler_firewall_auth.py::test_late_401_does_not_invalidate_refreshed_auth_entry`
- `lefthook run pre-commit`

Closes #29538


seven332:
## Code Review: PR #29558

**Reviewed head:** `d6f93a2fa0ee8fe53ad3724a05ab80b9b557c257`

### Summary

The entry-specific, credential-free identity closes the late-401 race without changing cache keys, persisted state, protocols, or external APIs. Invalidation remains synchronous and preserves the existing force-refresh cooldown.

### Findings

No findings.

### Testing

The regression covers two v1 consumers, a forced v2 refresh, a late v1 401, and a subsequent v2 cache hit through the request/response hooks. Concurrent identity sharing and malformed/missing identity failures are also covered. The full addon suite passed with 6655 tests, together with Ruff, BasedPyright, uv lock validation, and lefthook. Rust and Turbo checks were not run because those surfaces are unchanged.

### Verdict

LGTM

PR #29570, opened 2026-08-26
refactor(python): avoid caching resolved auth base urls

### LINKED ISSUE #29547: refactor(python): centralize auth.base runtime URL parsing
## Summary

`auth.base` forwarding uses two different URL parsing boundaries for the same request-lifetime credential URL. The rewrite builder parses the resolved base with the public, process-wide cached `urllib.parse.urlsplit()`, while the transport layer parses the constructed forwarding URL through the existing uncached `runtime_url_parsing.split_runtime_url()` helper.

This leaves secret-derived resolved bases outside the shared runtime URL ownership policy and allows them to remain reachable from urllib's 128-entry global LRU after the request flow is gone.

## Current evidence

Audited on current `main` at `e67a40c22f7e4a0df671fe315aba00e4bfdac2a3`.

- `crates/runner/mitm-addon/src/runtime_url_parsing.py:19-21` defines the canonical request-lifetime split helper specifically to avoid retaining inputs in urllib's process-wide LRU.
- `crates/runner/mitm-addon/src/auth_base_rewrite.py:187-228` validates `resolved_base` but calls the cached public `urllib.parse.urlsplit(resolved_base)` at line 197.
- `crates/runner/mitm-addon/src/auth.py:1423-1439` passes the auth-endpoint-resolved base into `build_rewrite_url()` during each forwarding request.
- `crates/runner/mitm-addon/src/auth_base_transport.py:550-561` then parses the resulting forwarding URL through `split_runtime_url()`, so the same credential URL lifecycle is split across cached and uncached parsing policies.
- `crates/runner/mitm-addon/tests/test_firewall_rewrite_forwarding.py:32-66` exercises a resolved base containing a secret path and trusted query, and lines 145-148 explicitly verify that the secret path is absent from persistent logs.
- `crates/runner/mitm-addon/tests/test_auth_base_rewrite.py:58-64` establishes that resolved-base queries can contain secret values.

A deterministic local probe called the real `build_rewrite_url()` boundary with 128 distinct credential-shaped resolved bases, discarded the local references, and forced collection. The resulting cache state was:

```text
CacheInfo(hits=0, misses=128, maxsize=128, currsize=128)
```

The exact retained byte count is allocator-dependent, but the cache population and lifetime are deterministic. Retention scales with both resolved URL length and the fixed 128-entry capacity.

## Structural impact

- Secret-derived webhook paths and query values can outlive their request and auth-resolution objects in process-wide parser state.
- The rewrite builder and transport layer encode conflicting ownership policy for one runtime capability.
- Future auth.base changes can reasonably assume the transport's uncached boundary covers the forwarding URL, even though the URL was already cached earlier during construction.
- The direct call bypasses the shared capability introduced by #24316 / PR #24354, which migrated auth-base transport parsing but did not cover resolved-base validation.

## Fix direction

Route resolved auth.base parsing through the existing request-lifetime URL parsing boundary rather than adding another parser or clearing global cache state. Keep stable configuration/platform URL parsing on the normal bounded stdlib cache.

## Acceptance criteria

- `_validated_rewrite_base()` uses the shared uncached runtime URL split boundary.
- A deterministic regression test passes a credential-shaped resolved base through `build_rewrite_url()` and proves that `urllib.parse.urlsplit.cache_info()` is unchanged.
- Existing auth.base scheme, authority, userinfo, IDNA, port, fragment, path-safety, query precedence, and encoding behavior remains unchanged.
- The end-to-end forwarding path still preserves current request/response, redaction, and failure semantics.
- Direct URL parsing within the auth.base rewrite/transport lifecycle is audited so all confirmed request-lifetime inputs use one ownership policy; stable configuration inputs remain cached.

seven332:
## Summary

- parse credential-derived auth.base rewrite URLs through the shared uncached runtime URL boundary
- preserve the existing auth.base validation, normalization, query merge, serialization, and forwarding behavior
- add a public-boundary regression that proves resolved bases do not enter urllib's process-wide split cache and existing stable cache entries remain intact

## Scope exclusions

- no auth result, cache, transport, API, schema, persistence, protocol, configuration, or migration changes
- no changes to stable configuration URL parsing or unrelated runtime URL consumers

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_auth_base_rewrite.py` (127 passed)
- `uv run --no-sync python -m pytest tests/` (6654 passed)
- lefthook pre-commit and commitlint hooks

Closes #29547

### DIFF
--- a/crates/runner/mitm-addon/src/auth_base_rewrite.py
+++ b/crates/runner/mitm-addon/src/auth_base_rewrite.py
@@ -17,6 +17,7 @@ from authority_utils import (
 )
 from host_normalization import normalize_hostname
 from path_security import has_unsafe_path
+from runtime_url_parsing import split_runtime_url
 from url_syntax import (
     has_raw_whitespace,
     has_unsafe_url_codepoint,
@@ -194,7 +195,7 @@ def _validated_rewrite_base(resolved_base: str) -> tuple[urllib.parse.SplitResul
             "Invalid auth.base URL: must not contain control characters or invalid Unicode"
         )
 
-    parsed = urllib.parse.urlsplit(resolved_base)
+    parsed = split_runtime_url(resolved_base)
     if parsed.scheme.lower() != _VALID_AUTH_BASE_SCHEME:
         raise ValueError("Invalid auth.base URL: scheme must be https")
     if not parsed.netloc:
--- a/crates/runner/mitm-addon/tests/test_auth_base_rewrite.py
+++ b/crates/runner/mitm-addon/tests/test_auth_base_rewrite.py
@@ -1,5 +1,6 @@
 """Tests for URL reconstruction and rewrite utilities."""
 
+import urllib.parse
 from collections.abc import Iterator
 
 import pytest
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #29570

**Reviewed head:** `5ea900fd0be45f27c3d279f7496ae8acf76f7717`

### Summary

The change correctly moves credential-derived auth.base parsing onto the existing uncached runtime URL boundary while leaving the security-sensitive validator and forwarding contracts unchanged. The regression also proves that the implementation preserves existing stable cache entries instead of clearing process-global state.

### Findings

No findings.

### Testing

Local validation passed: uv lock/sync, Ruff format/lint, BasedPyright, the 127-test rewrite suite, the complete 6654-test addon suite, and an additional 36-test cache/forwarding review run. A representative parser-equivalence probe also confirmed unchanged results and exceptions. GitHub CI is checked separately by the workflow.

### Verdict

LGTM

PR #29539, opened 2026-08-26
fix(platform): clarify forward target choices

Lunarivibe:
## Summary

- show the existing Forward helper copy below the dialog title
- explain that Agents starts a new chat while Chats adds the content to an existing chat
- tighten the search-to-group spacing and preserve the new copy across all supported locales
- cover the visible and accessible descriptions in the existing page-level flow

## Testing

- `pnpm exec prettier --write` on all changed files with prettier 3.8.1
- `pnpm --filter @okouai/app lint`
- `pnpm --filter @okouai/app check-types`
- `pnpm --filter @okouai/app exec vitest run src/views/okou-page/__tests__/chat-inline-feedback.test.tsx` (35 passed)

A local dev server was not started per repository guidance; the PR pipeline will provide the browser preview and E2E checks.

PR #26645, opened 2026-08-12
feat: forward selected chat content

e7h4n:
## Summary

- add a Forward action to selected assistant content with a two-step agent and chat picker
- gate only the Forward pill action behind the staff-enabled, default-off `chatForward` feature switch
- reuse the standard agent and thread composers so models, connectors, browser, computer use, templates, and attachments keep their existing behavior
- close on the optimistic message without navigation and show a success toast
- attach server-authoritative source run provenance to the forwarded message and reject invalid or client-authored agent source annotations
- add localized copy and focused coverage for the feature switch, keyboard selection, new chats, existing chats, empty context, and source run validation

## Testing

- `pnpm format`
- `pnpm turbo run lint --concurrency=1 --log-order grouped`
- `pnpm knip`
- `pnpm turbo run check-types --concurrency=1`
- `pnpm --filter @vm0/core exec vitest run src/__tests__/feature-switch.test.ts`
- `pnpm --filter @vm0/app exec vitest run src/views/zero-page/__tests__/chat-inline-feedback.test.tsx`
- `DATABASE_URL=postgresql://postgres@localhost:5432/vm0_test pnpm --filter api exec vitest run src/signals/routes/__tests__/chat-events.bdd.test.ts -t "persists user-forwarded run provenance"`

PR #29575, opened 2026-08-26
feat: roll out chat forwarding to all users

e7h4n:
## Summary

- enable `ChatForward` globally and remove the staff-org allowlist
- keep the switch registered so a per-user override can still disable the feature
- pin the existing switched-off page test to an explicit `false` override

## Rollout context

Chat forwarding was introduced behind a staff-only switch in #26645. The launch UI and copy refinements remain isolated in #29539; that PR should land before or together with this rollout.

## Verification

- `pnpm exec prettier --check packages/core/src/feature-switch.ts packages/core/src/__tests__/feature-switch.test.ts apps/platform/src/views/okou-page/__tests__/chat-inline-feedback.test.tsx`
- `pnpm --filter @okouai/core run lint`
- `pnpm --filter @okouai/core run check-types`
- `pnpm --filter @okouai/core exec vitest run src/__tests__/feature-switch.test.ts` (25 passed)
- `pnpm --filter @okouai/app run lint`
- `pnpm --filter @okouai/app exec vitest run src/views/okou-page/__tests__/chat-inline-feedback.test.tsx` (35 passed)
- `git diff --check`

The full local Vitest suite and a local dev server were not run per repository guidance; the PR pipeline will cover the broader checks.


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: c2ccf162a7729e4ff4b1dd6f9feca82ac25622df
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

This PR changes the canonical `ChatForward` registry entry from a staff-org-only rollout to globally enabled, while keeping the registered switch available for explicit user overrides. It also pins the existing disabled-mode page scenario to an explicit `false` override so that the off path remains covered after the global default changes.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and rollout intent | Pass | `turbo/packages/core/src/feature-switch.ts` sets `ChatForward.enabled` to `true` and removes only its staff-org hash list; override evaluation remains unchanged and still runs after the default. |
| `REVIEW.md` | Security, type safety, and maintainability | Pass | The diff changes no authorization boundary, endpoint, persisted data, cast, or public type; it is a focused registry and test update. |
| `REVIEW.md` | Feature and regression coverage | Pass | `feature-switch.test.ts` asserts the new global default; the page test explicitly proves the disabled override still hides Forward and leaves the shortcut inactive, while existing page scenarios in the same file cover forwarding to agents and chats. |
| `docs/bad-smell.md` | Avoid unsafe types, suppressions, dynamic imports, and hidden recovery | Pass | The three-file diff adds none of these patterns and removes the obsolete rollout allowlist rather than adding defensive logic. |
| `docs/testing.md` | Prefer integration behavior and mock only at boundaries | Pass | The changed platform scenario uses the real page bootstrap via `detachedSetupPage`; no internal mock, database mock, direct fetch mock, or fake timer is introduced. |
| `docs/fallback.md` | Introduce no unjustified or undeclared fallback | Pass | The diff adds no compatibility or fallback branch. The existing generic feature-switch override mechanism is preserved as operational configuration, not introduced as version-migration compatibility. |
| `docs/deployment-compatibility.md` | Protect independently deployed protocol and persisted-state boundaries | N/A | No frontend/backend request, response, service-worker, runner, queue, schema, migration, or persisted shape changes; only the value returned through the existing feature-switch contract changes. |
| `docs/effect.md` | Match effects and commands to their owner and trigger | N/A | No platform production effect, command, signal, resource, or lifecycle implementation is changed. |
| `docs/cache.md` | Preserve render purity, ownership, cache bounds, and teardown | N/A | No production React render, cache, ref, state owner, or resource lifecycle is changed. |
| `.claude/skills/ccstate/SKILL.md` | Follow ccstate command, signal, and abort ownership practices | N/A | The platform diff only supplies a supported `featureSwitches` option to the page-test bootstrap; no ccstate graph or async command changes. |
| `docs/testing/patterns.md` | Use the standard platform page-test pattern | Pass | `chat-inline-feedback.test.tsx` continues to enter through `detachedSetupPage` and configure the switch through the test context's supported setup option. |
| `docs/testing/anti-patterns.md` | Avoid internal mocks, fake timers, implementation-only assertions, and direct rendering | Pass | The changed test adds only page context; its assertions remain user-visible toolbar and keyboard behavior. |
| `docs/testing/app-testing.md` | Test platform behavior through the production page bootstrap | Pass | The off-path scenario renders `/chats/:id`, selects assistant text, and verifies the visible Forward action and shortcut behavior from the page. |
| `docs/testing/testing-external-behavior.md` | Construct and observe behavior at the external UI boundary | Pass | The test supplies user-visible feature context through page setup and observes UI/keyboard outcomes rather than mutating stores or inspecting internal callbacks. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- Coverage: Adequate — the shared switch contract covers the new global default, the page integration covers the explicit off override, and existing page integrations cover the Forward workflows.
- Conventions: Compliant — the changed page test uses production bootstrap and external UI assertions without adding prohibited mocks, timers, or cleanup.
- CI: Current-head CI was still running at review time; no failed check was observed, and the relevant lint, type, and multiple app-test jobs had already passed.

PR #29601, opened 2026-08-26
fix(python): preserve jsonl cleanup after reporter failure

### LINKED ISSUE #29590: fix(python): preserve JSONL cleanup after model reporter shutdown failure
## Problem

The mitmproxy `done()` hook is structured to preserve later cleanup when earlier shutdown stages fail. However, its final two stages are sequential inside the same `finally` block:

- `model_provider_failure.shutdown()`
- `shutdown_log_writer()`

If model-provider reporter shutdown raises, Python skips the JSONL writer shutdown. Accepted network-log entries can remain pending and the writer thread can outlive the intended addon cleanup sequence, potentially losing final log records during process exit.

## Evidence

- The current `done()` nesting preserves downstream cleanup after runner-flush, usage-executor, post-executor usage drain, and auth.base shutdown failures, but not after the model-provider reporter boundary.
- The existing deterministic matrix in `crates/runner/mitm-addon/tests/test_done_hook.py` covers usage-executor, usage-drain, and auth.base failures and asserts the original exception plus exact call ordering.
- Focused fault injection at `model_provider_failure.shutdown()` records the reporter call, propagates its original exception, and never reaches `shutdown_log_writer()`.
- The focused test file passes on current `main`: 12 tests.

## Decision

Fix the production shutdown control flow and extend the existing entry-point failure matrix in one PR. This is a standalone issue with no parent, blockers, migration, or cross-version protocol work.

## Chosen direction

- Keep the established producer-before-sink order.
- Put `model_provider_failure.shutdown()` in a nested `try` and `shutdown_log_writer()` in its `finally`, matching the cleanup structure already used by earlier stages.
- Do not catch, translate, aggregate, or suppress exceptions.
- Add a `model-provider` case to the existing parameterized failure matrix.
- Record reporter and JSONL calls for every case so earlier failure ordering remains explicit.
- Assert that JSONL shutdown is invoked exactly once and that the exact reporter exception object propagates when JSONL shutdown succeeds.

## Alternatives considered

- A generic cleanup stack/helper would add a new multi-error precedence policy and broaden every shutdown stage without a current requirement.
- Suppressing exceptions inside reporter shutdown would violate error visibility and the exception-propagation acceptance criterion.
- Moving JSONL shutdown before reporter shutdown would reverse producer-before-sink lifecycle ordering while reporter completion callbacks can still emit diagnostics.

## Risks and mitigations

- If both reporter and JSONL shutdown raise, Python's existing `finally` semantics propagate the later JSONL exception. Simultaneous failures and exception aggregation are intentionally out of scope; this PR preserves the cleanup chain's current semantics.
- Exact ordering assertions are intentionally strict so a future lifecycle reorder cannot silently weaken producer-before-sink cleanup.
- The change is local to one runner's mitmproxy shutdown. It changes no API, persisted data, configuration, dependency, or independently deployed protocol.

## Acceptance criteria

- A model-provider reporter shutdown failure still invokes `shutdown_log_writer()` exactly once.
- The original model-provider reporter shutdown exception remains the propagated error when JSONL shutdown succeeds.
- The established cleanup order and behavior remain covered for usage-executor, post-executor usage drain, and auth.base failures.
- The focused done-hook tests pass.
- `uv lock --check`, `uv sync --locked`, Ruff format and lint, BasedPyright, and the complete mitm-addon test suite pass.

## Delivery notes

- Expected source changes are limited to `crates/runner/mitm-addon/src/mitm_addon.py` and `crates/runner/mitm-addon/tests/test_done_hook.py`.
- Rollback is a normal source revert with no migration or cleanup.

seven332:
## Summary

- guarantee the bounded JSONL writer shutdown runs after model-provider reporter shutdown fails
- preserve the existing reporter-before-writer order and original exception propagation
- extend the `done()` failure matrix to cover the reporter boundary and all final-stage ordering

## Scope

- no API, database, persisted state, runner protocol, configuration, or dependency changes
- simultaneous reporter and JSONL shutdown failures remain governed by Python's existing `finally` exception semantics

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_done_hook.py` (13 passed)
- `uv run --no-sync python -m pytest tests/` (6,662 passed)
- `git diff --check`

Closes #29590

### DIFF
--- a/crates/runner/mitm-addon/src/mitm_addon.py
+++ b/crates/runner/mitm-addon/src/mitm_addon.py
@@ -1997,8 +1997,10 @@ def done():
             finally:
                 auth_base_forwarder.shutdown_forward_request_workers(wait=False)
         finally:
-            model_provider_failure.shutdown()
-            shutdown_log_writer()
+            try:
+                model_provider_failure.shutdown()
+            finally:
+                shutdown_log_writer()
 
 
 # ============================================================================
--- a/crates/runner/mitm-addon/tests/test_done_hook.py
+++ b/crates/runner/mitm-addon/tests/test_done_hook.py
@@ -374,6 +374,7 @@ class TestDoneHook:
                 (
                     "usage-executor:shutdown:wait=True",
                     "auth-base:shutdown:wait=False",
+                    "model-provider:shutdown",
                     "jsonl:shutdown",
                 ),
                 id="usage-executor-shutdown",
@@ -384,6 +385,7 @@ class TestDoneHook:
                     "usage-executor:shutdown:wait=True",
                     "usage:drain",
                     "auth-base:shutdown:wait=False",
+                    "model-provider:shutdown",
                     "jsonl:shutdown",
                 ),
                 id="post-executor-usage-drain",
@@ -394,9 +396,21 @@ class TestDoneHook:
                     "usage-executor:shutdown:wait=True",
                     "usage:drain",
                     "auth-base:shutdown:wait=False",
+                    "model-provider:shutdown",
                     "jsonl:shutdown",
                 ),
                 id="auth-base-worker-shutdown",
+            ),
+            pytest.param(
+                "model-provider",
+                (
+                    "usage-executor:shutdown:wait=True",
+                    "usage:drain",
+                    "auth-base:shutdown:wait=False",
+                    "model-provider:shutdown",
+                    "jsonl:shutdown",
+                ),
+                id="model-provider-reporter-shutdown",
             ),
         ],
     )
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #29601

**Reviewed head:** `83e4f28a23a93c4f09cbff932706210a86010ee2`

### Summary

The PR closes the final downstream-cleanup gap in `mitm_addon.done()` by applying the existing nested `try`/`finally` pattern to model-provider reporter and JSONL writer shutdown. The regression matrix now verifies the new failure boundary, exact ordering for earlier failures, single JSONL invocation, and original exception identity.

### Findings

No findings.

### Testing

- Focused done-hook suite: 13 passed.
- Complete mitm-addon suite: 6,662 passed.
- Ruff format and lint: passed.
- BasedPyright: passed with 0 errors, warnings, or notes.
- uv lock/sync and diff checks: passed.
- Unrelated Rust, Turbo, browser, and E2E suites were not run because the change does not cross those boundaries. GitHub Actions status is checked separately.

### Verdict

LGTM

PR #29602, opened 2026-08-26
docs(python): document buffered response capture decode policy

### LINKED ISSUE #29587: docs(python): document buffered response capture decode policy
## Context

`crates/runner/mitm-addon/src/body_decoding.py:383-424` exposes `decode_response_body_for_network_log_capture()`, but its docstring only says that it provides bounded mitmproxy compatibility. The return and codec policy cannot be inferred from that sentence:

- Missing or `identity` encoding returns the input bytes, while unsupported encodings and decode failures return `None`.
- Gzip accepts both gzip and zlib wrappers and may return output from an incomplete stream; deflate tries zlib and raw wrappers but requires a complete stream unless the output limit is reached.
- Brotli rejects incomplete or invalid input except that hitting the decoded-output limit returns the bounded prefix.
- Zstandard reads concatenated frames up to the bound and returns `None` for malformed or invalid trailing input.
- `b""` is a successful result distinct from `None`, and `max_output` can produce a successful bounded prefix.

The distinction is especially important because `crates/runner/mitm-addon/src/body_capture.py:538-569` applies two different response-capture policies. A retained streaming buffer uses the best-effort `decompress_body()` path, whose documented behavior can preserve original bytes or partial output after a decode problem. A buffered `flow.response.raw_content` value uses `decode_response_body_for_network_log_capture()`, and `None` causes the body to be omitted and marked binary. The `add_capture_fields()` documentation at `body_capture.py:466-470` names the two data sources but does not explain this policy difference.

The behavior is distributed across tests: `tests/test_body_capture_decompression.py:169-213` covers best-effort stream-buffer fallback and partial gzip output, while `tests/test_body_capture_fields.py:386-510` covers strict buffered failures, mitmproxy-compatible zlib variants, output bounds, and Zstandard frame handling.

## Impact

A contributor changing or reusing these helpers can reasonably treat them as interchangeable, or interpret `None` as an ordinary empty result. That can silently change persistent network-log behavior by retaining opaque wire bytes that the buffered path should suppress, dropping recoverable stream-buffer content, or breaking compatibility for valid gzip, deflate, Brotli, or Zstandard variants.

## Acceptance criteria

- Expand the `decode_response_body_for_network_log_capture()` docstring or adjacent public documentation to define the `bytes`, `b""`, and `None` outcomes and the decoded-output bound.
- Summarize the supported codec compatibility and completeness behavior that callers must preserve, including gzip/zlib auto-detection, raw deflate fallback, Brotli completeness, and concatenated Zstandard frames.
- Document at the response-capture call site that retained stream buffers intentionally use best-effort decoding while buffered `raw_content` uses the strict mitmproxy-compatible helper.
- Keep runtime decoding and network-log behavior unchanged.

No implementation change is requested.

seven332:
## Summary

- Document the buffered response decoder's return values, output bound, and codec completeness policy.
- Explain why retained response stream buffers use best-effort decoding while buffered
  `flow.response.raw_content` uses the strict helper, including the `None` and `b""` capture outcomes.

## Scope

This is a documentation-only change in the mitmproxy addon. Runtime decoding, network-log behavior,
tests, dependencies, and schemas are unchanged.

## Validation

- `uv lock --check`
- `uv run --no-sync python -m pytest tests/test_body_capture_fields.py tests/test_body_capture_decompression.py tests/test_body_decoding.py` (178 passed)

Closes #29587

### DIFF
--- a/crates/runner/mitm-addon/src/body_capture.py
+++ b/crates/runner/mitm-addon/src/body_capture.py
@@ -467,7 +467,13 @@ def add_capture_fields(flow: http.HTTPFlow, log_entry: dict) -> None:
     ``response_streaming.configure_response_stream()`` because that path keeps a
     bounded raw wire-byte buffer and records whether it was truncated. Bounded
     ``flow.response.raw_content`` decoding is used only when no stream
-    buffer metadata exists.
+    buffer metadata exists. These sources intentionally use different decode
+    policies: retained stream buffers use best-effort ``decompress_body()``,
+    which can preserve original wire bytes or partial decoded output after a
+    decode problem, while buffered ``raw_content`` uses the strict
+    ``decode_response_body_for_network_log_capture()`` helper. A strict
+    ``None`` result omits a non-empty buffered body and marks it ``binary``;
+    ``b""`` remains a successful empty result with no body field.
     """
     # Request headers (always available)
     request_inspection = _inspect_headers_for_capture(flow.request.headers)
--- a/crates/runner/mitm-addon/src/body_decoding.py
+++ b/crates/runner/mitm-addon/src/body_decoding.py
@@ -386,7 +386,31 @@ def decode_response_body_for_network_log_capture(
     *,
     max_output: int = DEFAULT_BODY_DECODE_LIMIT,
 ) -> bytes | None:
-    """Decode a buffered response body with bounded mitmproxy compatibility."""
+    """Decode a buffered response body with bounded mitmproxy-compatible semantics.
+
+    Missing or ``identity`` encoding returns ``data`` unchanged. A supported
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #29602

**Reviewed head:** `9d81f1ee276ad09cb0eecba622816aeeb682cefd`

### Summary

The PR accurately documents the strict buffered response decoder contract and the intentional
best-effort versus strict policy split at the response capture call site. The change is limited to
the two planned docstrings and matches the current implementation and behavior tests.

### Findings

No findings.

### Testing

Ruff format/check, BasedPyright, and the focused body-capture/decoder suite passed (178 tests).
The complete addon suite remains an unverified CI area.

### Verdict

LGTM

PR #29609, opened 2026-08-26
fix(runner): apply parent-death setup to kmsg monitor

### LINKED ISSUE #29589: refactor(rust): apply the parent-death contract to the kmsg monitor
## Structural overlap

The runner has a canonical direct-child lifetime contract at `crates/runner/src/parent_death.rs:5-25`: install `PR_SET_PDEATHSIG(SIGKILL)` and verify the captured parent immediately after installation so a child cannot survive a fork-to-`prctl` reparenting race.

The two long-lived network-log children share the same post-spawn owner, `NetworkLogProcess`:

- `crates/runner/src/dns/proxy.rs:123-134` configures the canonical parent-death contract before spawning `dnsmasq`, then transfers the child to `NetworkLogProcess`.
- `crates/runner/src/kmsg_log.rs:74-81` spawns `dmesg -w` directly and transfers it to the same owner without configuring that contract.
- `crates/runner/src/network_log_process.rs:46-65,144-151` provides explicit stop and `Drop` cleanup, but both require runner code to execute and cannot protect a child when the runner itself is terminated abruptly.

This is concrete lifecycle drift between sibling spawn paths. Repository history also shows that the original kmsg implementation installed `PR_SET_PDEATHSIG`; the async rewrite in commit `f871fb1cda90` removed it while relying on pipe closure and later explicit cleanup. The shared race-safe parent-death owner was introduced afterward in #26936, but the kmsg spawn was not connected to it.

## Impact

If the runner is killed, crashes, or is otherwise terminated without running `Drop`, `dmesg -w` can remain blocked as an orphan until it next writes to the closed pipe. On a quiet host that can be indefinite. Repeated standalone runner failures or restarts can therefore accumulate unowned kernel-log followers and process/file-descriptor state even though the sibling `dnsmasq` path is protected.

## Acceptance criteria

- Configure the canonical runner parent-death contract before spawning the production kmsg child.
- Preserve the existing stdout/stderr pipes, cancellation, explicit stop, and `NetworkLogProcess` drop/reap behavior.
- Add focused coverage proving that the kmsg command receives the same parent-death setup as other runner-owned long-lived children, including the reparenting-race contract where practical.
- Keep parent-death setup owned by `crate::parent_death`; do not introduce another direct `prctl` implementation.

seven332:
## Summary

- apply the runner's canonical parent-death contract before spawning the long-lived `dmesg -w` kmsg monitor
- keep the existing command arguments, pipes, cancellation, and `NetworkLogProcess` cleanup behavior unchanged
- add a real-process test that reads `PR_GET_PDEATHSIG` before exec and requires `SIGKILL`

## Scope

- no changes to the shared `parent_death` implementation
- no public API, configuration, persistence, or cross-version protocol changes

## Validation

- `cargo test -p runner kmsg_log::tests::command_configures_parent_death_signal -- --exact`
- mutation check: the focused test fails with `EPROTO` when the kmsg helper call is removed
- `cargo test -p runner parent_death::tests`
- `cargo test -p runner` (3332 passed, 24 ignored; integration test passed)
- pre-commit workspace Rust formatting, Clippy, documentation, and file-size checks

Closes #29589

### DIFF
--- a/crates/runner/src/kmsg_log.rs
+++ b/crates/runner/src/kmsg_log.rs
@@ -72,13 +72,20 @@ impl KmsgHandle {
 /// network log entries. Returns a handle; call [`KmsgHandle::stop`] during
 /// shutdown so the tokio runtime can exit cleanly.
 pub fn spawn(network_log_manager: NetworkLogManager) -> std::io::Result<KmsgHandle> {
-    let child = tokio::process::Command::new("dmesg")
-        .args(["-w"])
-        .stdout(Stdio::piped())
-        .stderr(Stdio::piped())
-        .spawn()?;
+    let mut command = tokio::process::Command::new("dmesg");
+    configure_command(&mut command);
+    let child = command.spawn()?;
 
     KmsgHandle::from_child(child, network_log_manager)
+}
+
+fn configure_command(command: &mut tokio::process::Command) {
+    command
+        .args(["-w"])
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped());
+
+    crate::parent_death::configure_parent_death_signal(command);
 }
 
 impl KmsgHandle {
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #29609

**Reviewed head:** `d53ec9f1f99afa9f406a0aaf959055fe52539573`

### Summary

The kmsg spawn now applies the runner's canonical race-safe parent-death contract before starting `dmesg -w`. Existing arguments, pipes, startup error propagation, cancellation, and `NetworkLogProcess` ownership remain unchanged. The added real-process test observes `SIGKILL` through `PR_GET_PDEATHSIG` at the child pre-exec boundary.

### Findings

No findings.

### Testing

Formatting, focused kmsg and shared parent-death tests, runner Clippy, runner documentation, the full runner suite, and pre-commit workspace Rust checks passed. The focused test was mutation-checked and failed with `EPROTO` when the kmsg helper call was removed. A privileged host `dmesg -w` smoke test was not run; the executable, arguments, pipes, and post-spawn path are unchanged.

### Verdict

LGTM

PR #29713, opened 2026-08-27
fix(python): reject incomplete zstd capture bodies

### LINKED ISSUE #29708: bug(python): reject incomplete zstd bodies in buffered capture
## Problem

`decode_response_body_for_network_log_capture()` treats an incomplete zstd frame as a successful buffered decode.

The zstd branch in `crates/runner/mitm-addon/src/body_decoding.py:369` reads through `ZstdDecompressor().stream_reader(...).read(max_output)` but does not verify that every frame reached EOF. The zstandard reader can return a decoded prefix without raising when the final frame is truncated. `crates/runner/mitm-addon/src/body_capture.py:570` then treats those bytes as a complete response body.

This differs from the intended strict buffered-capture contract and from the Brotli/deflate branches, which reject incomplete input unless the decoded-output bound was reached.

## Reproduction

```python
import zstandard
from mitmproxy import http

from body_decoding import decode_response_body_for_network_log_capture

body = b"hello world"
compressed = zstandard.ZstdCompressor().compress(body)
headers = http.Headers(content_encoding="zstd")

assert decode_response_body_for_network_log_capture(
    compressed[:-1],
    headers,
    max_output=1024,
) == b"hello worl"
```

Passing the same buffered response through `body_capture.add_capture_fields()` emits:

```python
{
    "response_body": "hello worl",
    "response_body_encoding": "utf-8",
}
```

There is no `response_body_truncated` marker and no binary/decode-failure classification.

## Impact

An HTTP-complete response whose advertised zstd body is malformed or truncated can be persisted in capture-mode network logs as an apparently complete plaintext response. This makes diagnostic/audit data misleading and can silently omit the tail of JSON or text content.

The best-effort behavior for an explicitly truncated streaming capture is separate and should remain unchanged; this issue concerns the strict `flow.response.raw_content` path.

## Acceptance criteria

- Buffered zstd decoding rejects an incomplete first frame and a valid frame followed by an incomplete frame.
- `add_capture_fields()` omits the decoded body and marks a non-empty rejected body as binary, consistent with other strict decode failures.
- Valid single and concatenated zstd frames still decode successfully.
- Reaching the configured decoded-output limit still returns a bounded prefix without requiring full-frame validation beyond the limit.
- Regression tests cover both incomplete-frame cases without weakening the current decoded-output memory bound.

seven332:
## Summary

- reject incomplete zstd frames on the strict buffered response-capture path
- preserve valid concatenated frames and the bounded-prefix behavior once the decoded-output limit is reached
- add capture-entry integration coverage for an incomplete first frame and an incomplete trailing frame

## Scope

This changes only buffered `flow.response.raw_content` decoding. Explicitly truncated response stream buffers continue to use the existing best-effort decoder. There are no API, schema, database, protocol, persisted-state, migration, or deployment-order changes.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_body_capture_fields.py tests/test_body_capture_decompression.py tests/test_body_decoding.py` (180 passed)
- `uv run --no-sync python -m pytest tests/` (6666 passed)

Closes #29708

### DIFF
--- a/crates/runner/mitm-addon/src/body_decoding.py
+++ b/crates/runner/mitm-addon/src/body_decoding.py
@@ -375,9 +375,15 @@ def _decode_zstd_for_network_log_capture(data: bytes, max_output: int) -> bytes 
             data,
             read_across_frames=True,
         ) as reader:
-            return reader.read(max_output)
+            body = reader.read(max_output)
     except zstandard.ZstdError:
         return None
+
+    if len(body) >= max_output:
+        return body
+    if _validate_complete_zstd_frames(data, max_output) is not None:
+        return None
+    return body
 
 
 def decode_response_body_for_network_log_capture(
@@ -404,8 +410,8 @@ def decode_response_body_for_network_log_capture(
       requires a complete stream unless the decoded-output bound is reached.
     - ``br`` rejects invalid or incomplete input, except that reaching the
       decoded-output bound returns the bounded prefix.
-    - ``zstd`` reads concatenated frames up to ``max_output`` bytes and returns
-      ``None`` when the reader reports malformed or invalid trailing input.
+    - ``zstd`` reads concatenated frames up to ``max_output`` bytes and rejects
+      malformed or incomplete frames unless the decoded-output bound is reached.
 
     This is stricter than the best-effort ``decompress_body()`` path used for
     retained streaming buffers, which can preserve original wire bytes or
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #29713

**Reviewed head:** `501b6bca956d4cd5e98315584abaaa4b96ee710e`

### Summary

The PR makes strict buffered zstd response capture validate every frame when decoded output remains below the configured bound. It preserves the existing bounded-reader short circuit, valid concatenated-frame behavior, and the separate best-effort streaming path.

### Findings

No findings.

### Testing

Ruff format/check, BasedPyright, 180 focused capture/decoder tests, and the complete 6666-test addon suite passed. Additional review probes covered single and concatenated frames, missing content size, empty and skippable frames, incomplete first and trailing frames, and output-limit behavior. Live proxy interception was not run; the repository addon testing guide excludes it from this test surface.

### Verdict

LGTM

PR #29729, opened 2026-08-27
fix(vsock-guest): mark incomplete exec output truncated

### LINKED ISSUE #29710: bug(rust): mark exec output truncated when drain deadline expires
## Baseline

Audited at `e94cd19ae112db48a783d8154b7fd0d88cbfef99`.

## Location

- `crates/vsock-guest/src/drain.rs:63-66`
- `crates/vsock-guest/src/drain.rs:79-149`
- `crates/vsock-guest/src/drain.rs:185-232`
- `crates/vsock-guest/src/wait.rs:37-58`
- `crates/vsock-guest/src/exec_operation.rs:810-829`
- `crates/vsock-guest/src/exec_operation.rs:855-859`
- `crates/vsock-guest/src/exec_operation.rs:1742-1749`
- Existing regression scenario: `crates/vsock-guest/tests/connection/exec_cancel_cleanup.rs:269-306`

## Problem

When an exec command's direct child exits but a descendant keeps stdout or stderr open, `await_drain_deadline` eventually cancels the unfinished drain worker. `RunningExec::finish` detects this condition and logs `drain deadline reached`, but it does not propagate that incompleteness into the exec result.

`BoundedDrainResult::capture_truncated` is set only when the configured byte limit drops an observed chunk. Cancellation before EOF does not set it, and `captured_output` forwards the value unchanged. The host therefore receives partial captured output with `truncated: false`. For streamed output, cancellation also ends the reader without emitting the terminal truncated marker used when the stream byte limit is exceeded.

Sibling users of the same deadline helper conservatively mark their output as truncated when drain workers do not complete, including guest storage manifest collection, state restore, and DNS readiness.

## Reproduction

1. Start an exec operation with captured or captured-and-streamed output.
2. Have the foreground shell write an initial marker and exit successfully while a grandchild inherits stdout or stderr and writes again after the drain deadline.
3. The drain deadline expires and the guest cancels the unfinished reader.
4. The late bytes are absent, but the terminal exec result reports `Exited(0)` with `truncated: false`; streamed consumers receive no truncation marker.

The existing `exec_operation_returns_when_orphaned_grandchild_holds_stdout` integration test reaches this deadline branch, but it only asserts the successful termination, initial marker, and bounded elapsed time. A focused run passes while logging `exec operation: drain deadline reached ... after 500ms`.

## Impact

Host and runner consumers can treat partial command output as complete. This can silently hide diagnostics or allow downstream parsing to accept an incomplete stdout payload, and the CLI's existing truncation warning is not shown.

## Acceptance criteria

- Captured output is reported as truncated when its drain is cancelled before EOF because the drain deadline expires or the worker fails; marking both streams conservatively is acceptable when the incomplete stream cannot be identified.
- Streamed output receives an equivalent terminal indication when deadline cancellation cuts it off.
- Output that reaches EOF within the deadline and remains within its configured byte limit is not marked truncated.
- The orphaned-grandchild integration scenario asserts the terminal truncation state, including the streaming indication where applicable.
- Existing timeout, cancellation cleanup, and bounded-completion behavior remains intact.

seven332:
## Summary

- treat exec drains that miss their shared deadline or fail to return a worker result as incomplete
- preserve requested capture policies and conservatively mark incomplete captured output truncated
- emit ordered terminal truncation frames for streamed output without duplicating existing limit markers
- extend the real orphaned-grandchild connection scenario to cover captured and streamed truncation

## Scope

- changes only the bundled `vsock-guest` exec drain/finalization behavior and its tests
- reuses the existing captured truncation flag and empty streamed truncation frame
- does not change the wire format, host decoding, runner APIs, persisted data, migrations, or deployment sequencing

## Validation

- `cargo test -p vsock-guest --test connection exec_operation_returns_when_orphaned_grandchild_holds_stdout`
- `cargo test -p vsock-guest`
- `cargo clippy -p vsock-guest --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p vsock-guest --no-deps`
- `git diff --check`
- pre-commit workspace Rust fmt, docs, Clippy, file-size, and commit-message checks

Closes #29710

### DIFF
--- a/crates/vsock-guest/src/drain.rs
+++ b/crates/vsock-guest/src/drain.rs
@@ -63,6 +63,7 @@ pub(crate) struct BoundedStreamConfig {
 pub(crate) struct BoundedDrainResult {
     pub(crate) captured: Option<Vec<u8>>,
     pub(crate) capture_truncated: bool,
+    pub(crate) stream_truncated: bool,
 }
 
 /// Drain `pipe` until EOF or `cancel` is set, calling `on_chunk` for each
@@ -229,6 +230,7 @@ where
     BoundedDrainResult {
         captured,
         capture_truncated,
+        stream_truncated,
     }
 }
 
@@ -451,6 +453,7 @@ mod tests {
 
         assert_eq!(result.captured, None);
         assert!(!result.capture_truncated);
+        assert!(!result.stream_truncated);
     }
 
     #[test]
@@ -516,6 +519,7 @@ mod tests {
 
         assert_eq!(result.captured, Some(input));
         assert!(!result.capture_truncated);
+        assert!(result.stream_truncated);
         assert_eq!(chunks, vec![(stream_limit, false), (0, true)]);
     }
 
@@ -543,6 +547,7 @@ mod tests {
 
         assert_eq!(result.captured, Some(b"abcdef".to_vec()));
         assert!(!result.capture_truncated);
+        assert!(result.stream_truncated);
         assert_eq!(
             chunks,
             vec![
@@ -577,6 +582,7 @@ mod tests {
         );
 
         assert_eq!(result.captured, None);
+        assert!(!result.stream_truncated);
         assert_eq!(
             chunks,
             vec![(b"ab".to_vec(), false), (b"cd".to_vec(), false)]
@@ -606,6 +612,7 @@ mod tests {
         );
 
         assert_eq!(result.captured, None);
+        assert!(result.stream_truncated);
         assert_eq!(chunks, vec![(Vec::new(), true)]);
     }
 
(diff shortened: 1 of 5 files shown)


seven332:
## Code Review: PR #29729

**Reviewed head:** `6fa6a2c9d6ecb6061e753c333ca5aee7d2c04e6e`

### Summary

The change preserves the existing drain and wire contracts while carrying incomplete-drain state into captured and streamed terminal output. Finalization retains one shared stream writer, joins both drain workers before emitting ordered terminal markers, reconstructs missing capture results from the requested policy, and suppresses duplicate stream markers after a byte-limit truncation.

### Findings

No findings.

### Testing

The real inherited-FD connection regression verifies captured truncation, streamed terminal markers, cross-stream sequencing, and marker de-duplication. The full `vsock-guest` suite, Rust formatting, Clippy with warnings denied, Rust documentation with warnings denied, and `git diff --check` passed. A post-spawn drain-thread panic is not fault-injected directly; its join/result fallback was reviewed structurally.

### Verdict

LGTM

Issue #29469, opened 2026-08-26
test(connectors): audit OAuth scope fixtures across grant flows


seven332:
## Parent objective

Complete the final regression audit for #29462 after all OAuth partial-grant implementation slices have landed.

## Problem and evidence

The prerequisite issues established the requested-versus-granted model, but the final fixture audit found uneven coverage and two remaining implementation gaps.

- The former Datadog inconsistency is fixed: its public connector read persists the two provider-reported scopes while scope diff separately retains all seven authorization-request scopes.
- Auth-code, standard custom OAuth, and built-in refresh tests distinguish provider-reported values from omission.
- Device-auth happy paths report the same values they request. The device poll contract does not receive the session's stored start-time scopes, and Base44 plus the synthetic OAuth device adapter convert an omitted token-response `scope` to `[]` instead of the effective request.
- Nintendo Store and Nintendo Switch Parental Controls parse omission as `null` and explicit empty as `[]`, but initial external-code completion currently falls back to requested scopes for both states. Their refresh paths already preserve the distinction.
- Steam OpenID returns a deterministic empty effective scope set, but its public API journey does not assert it.

Without this final pass, fixtures can certify request-to-grant fabrication, and optional scope responses can persist a false empty or false requested grant.

## Decision and chosen direction

Keep this as one independently reviewable connector/API PR. Carry the device session's existing start-time requested-scope snapshot through the internal poll boundary, use it only when an optional device token response omits `scope`, preserve reported values exactly, and correct Nintendo initial grants to fall back only on omission. Add a small discriminating integration matrix instead of duplicating every variant in every flow.

Do not redesign the common grant result, add a migration, or reopen already-correct auth-code, custom OAuth, refresh, readiness, and operation behavior.

## Acceptance criteria

- Provider fixtures return realistic token scopes or explicitly model an omitted or empty `scope` response.
- No test expects unreported catalog-request scopes to appear as authoritative provider-granted scopes.
- Device poll adapters receive the exact start-time requested-scope snapshot, including when the catalog changes before polling.
- A reported device scope set, including supplemental or explicit-empty scopes, is preserved exactly; omission uses the start-time request.
- Base44 token-scope omission retains its effective requested grant through the real device API journey.
- Nintendo Store and Nintendo Switch Parental Controls initial external-code completion preserve explicit empty and fall back only on omission.
[...]

PR #29751, opened 2026-08-27
fix(mail): enforce operation-specific gmail scopes

### LINKED ISSUE #29467: fix(mail): enforce Gmail scopes at operation boundaries
## Parent objective

Part of #29462, the connector OAuth partial-grant correctness initiative.

## Problem and evidence

Gmail mail-draft operations currently derive `scopesReady` with `connectorAuthMethodHasRequiredScopes()`. That helper requires every scope configured on the connector auth method, so one mail operation is coupled to connector-wide authorization policy.

The scope guard itself is intentional. The original mail implementation in #21458 checked scopes both while selecting an eligible connection and again before resolving an access token; #21886 preserved that behavior while moving credential execution to the runtime catalog. The outdated assumption is that satisfying every connector-requested scope is the right readiness test for every operation.

After #29465, requested scopes and authoritative granted scopes are separate facts. A partial grant can therefore leave one operation unavailable without making the connector globally invalid. The X image-sharing service already demonstrates the intended operation-boundary pattern with explicit `tweet.write` and `media.write` requirements.

## Decision

Preserve fail-closed permission enforcement, but make Gmail mail-draft operations declare and validate only the Gmail permissions they need. Keep connection health (`needsReconnect`) separate from operation authorization.

## Constraints

- Use authoritative granted scopes when present while preserving the rollout-compatible fallback established by #29465.
- Do not mark the connector invalid or require a connector-wide reconnect solely because a Gmail operation permission is missing.
- Continue rejecting unhealthy connections and unavailable credentials independently of scope authorization.
- Return an operation-specific permission error before calling Gmail when required grants are absent.
- Keep the scope requirement close to the Gmail operation boundary so future connector-requested scopes do not silently become mail requirements.

## Acceptance criteria

- Gmail mail-draft operations explicitly declare and check the Gmail scope(s) they require.
- Connections with the required Gmail operation scope remain usable even if another connector-requested scope is absent.
- Missing Gmail operation permissions fail closed before a Gmail API call and report that the operation lacks permission rather than treating the credential as globally invalid.
- `needsReconnect` and missing credential behavior remain unchanged and distinct from missing operation permission.
- Current runtime `oauthScopes` consumers are audited; operation consumers either use explicit operation requirements or are included in this change.
- API integration tests exercise permitted and unpermitted mail-draft paths through service entry points.

## Dependency

#29465 is complete and provides the requested-versus-granted scope separation used by this change.

## Relevant code

- `turbo/apps/api/src/signals/services/mail-draft.service.ts`
- `turbo/apps/api/src/signals/services/image-share-x.service.ts`
- `turbo/packages/connectors/src/connector-auth-method.ts`

seven332:
## Summary

- replace Gmail mail-draft connector-wide scope readiness with an explicit `gmail.modify` operation requirement
- keep reconnect and credential failures separate from missing mail permissions, returning a draft-specific permission error before Gmail or credential I/O
- preserve permission failures when reading linked drafts instead of incorrectly presenting them as reconnect-required
- cover authoritative granted scopes, divergent requested scopes, and missing operation grants through the mail route integration suite

## Behavior

The original fail-closed scope guard remains in place. A healthy Gmail connection is now selected independently of connector-wide requested scopes, then the mail service checks its effective granted-scope snapshot for the permission required by mail-draft operations. Existing `oauthGrantedScopes ?? oauthScopes` rollout compatibility remains unchanged. Connection or credential unavailability still produces reconnect behavior; an insufficient grant remains an explicit operation permission error.

## Exclusions

- no connector health, schema, frontend, runner, or public contract changes
- no X image-sharing behavior changes; its operation scopes were audited and are already explicit
- no broader OAuth fixture audit, which remains tracked by #29469

## Validation

- `pnpm format`
- `pnpm -F api lint`
- `pnpm -F api check-types`
- `pnpm -F api exec vitest run src/signals/routes/__tests__/mail.test.ts --silent=passed-only` — 16 passed
- `pnpm knip`
- pre-commit repository type checks — 14 packages passed
- `git diff --check`

Closes #29467

Parent context: #29462

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/helpers/api-bdd-connectors.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/helpers/api-bdd-connectors.ts
@@ -576,6 +576,7 @@ interface GmailConnectorOAuthOptions {
   readonly refreshToken?: string;
   readonly subject?: string;
   readonly email?: string;
+  readonly scopes?: readonly string[];
 }
 
 export function mockGmailConnectorOAuth(
@@ -603,7 +604,9 @@ export function mockGmailConnectorOAuth(
         refresh_token: options.refreshToken ?? "gmail-refresh-token",
         expires_in: 3600,
         token_type: "Bearer",
-        scope: "https://www.googleapis.com/auth/gmail.modify",
+        scope: (
+          options.scopes ?? ["https://www.googleapis.com/auth/gmail.modify"]
+        ).join(" "),
       });
     }),
     http.get(GOOGLE_OPENID_USERINFO_URL, () => {
--- a/turbo/apps/api/src/signals/routes/__tests__/mail.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/mail.test.ts
@@ -29,6 +29,7 @@ import {
   seedCustomThreadConnectorSelection,
   setConnectorDefaultState,
   setConnectorCredentialStorageState,
+  setLegacyBuiltinOAuthScopes,
   setConnectorSecretOwner,
 } from "./helpers/connector-credential-storage-state";
 import { mailRoutes } from "../mail";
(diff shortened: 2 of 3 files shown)


seven332:
## Code Review: PR #29751

**Reviewed head:** `23727363dd206868db39843bba7eeabb34150f20`

### Summary

The mail service now validates the explicit `gmail.modify` operation grant against the effective granted-scope snapshot, independently of connector-wide requested scopes. Permission failures remain distinct from connection or credential unavailability across linked-draft operations.

### Findings

No findings.

### Testing

Focused mail route integration coverage passes with 16 tests, including authoritative-versus-requested scope divergence, zero Gmail calls for a missing grant, and no false reconnect state for an already linked draft. API lint and type checks, formatting, Knip, pre-commit repository type checks, and `git diff --check` also passed. The full repository test suite is left to CI.

### Verdict

LGTM

seven332:
Closing after self-review. Gmail currently requests only https://www.googleapis.com/auth/gmail.modify, so the existing connector-wide required-scope check is equivalent to the proposed operation-level check. The additional-scope scenario in the test is not a current product state, and treating a missing sole grant as a non-reconnect permission error would weaken the recovery path.

PR #29754, opened 2026-08-27
test(python): cover mismatched upstream binding refresh guards

### LINKED ISSUE #29743: test(python): cover mismatched upstream binding refresh guards
## Problem

At `12315a55eb51f5083d20a111a65945bb0748de33`, `refresh_server_binding_connected_address_if_matching()` rejects a refresh when the stored direct binding does not match the current normalized destination (`crates/runner/mitm-addon/src/upstream_destination_binding.py:259-269`). That guard is security-sensitive because the successful path keeps the stored binding host and port, replaces its endpoint, adds a binding kind, and returns `True` (`crates/runner/mitm-addon/src/upstream_destination_binding.py:271-279`).

Full branch coverage across the 6,669-test mitm-addon suite leaves the host/port mismatch return at line 266 unexecuted. The positive retargeting test at `crates/runner/mitm-addon/tests/test_request_handler_connector_admission.py:369-408` refreshes a binding for the same authority. The existing mismatched-binding test at `crates/runner/mitm-addon/tests/test_request_handler_connector_admission.py:952-991` does not establish verified upstream TLS, so the caller exits before invoking the refresh helper.

## Impact

If the host/port guard is removed or weakened, a connected request with verified TLS for destination B and an existing direct binding for destination A can make `_bind_flow_upstream_destination()` (`crates/runner/mitm-addon/src/upstream_admission.py:474-508`) treat the mismatched binding as successfully refreshed. The request may then continue into auth resolution and credential injection even though no direct binding was established for destination B. The current suite would not detect that regression.

## Acceptance criteria

- Add deterministic coverage for a connected, verified-TLS upstream with an existing direct binding whose host differs from the current destination.
- Cover the equivalent port-mismatch case, either at the integration boundary or directly against the refresh helper.
- Assert that admission fails closed, auth resolution is not called, no managed credentials are applied, and the original binding remains unchanged.
- Preserve the positive same-authority endpoint-retargeting behavior.

seven332:
## Summary

- exercise mismatched upstream binding refreshes after verified TLS has established the current destination
- cover stored host and stored port mismatches through the real connector request-hook path
- assert admission stays fail closed, auth resolution and credential injection do not run, and the complete binding remains unchanged

## Scope

- test-only change in the mitmproxy addon
- no production code, API, dependency, migration, or compatibility changes

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest -v tests/test_request_handler_connector_admission.py::test_matching_sni_and_host_allows_connected_firewall_auth_after_retargeting tests/test_request_handler_connector_admission.py::test_test_connector_rejects_mismatched_existing_binding_after_verified_tls` (3 passed)
- `uv run --no-sync python -m pytest tests/` (6,670 passed)

Closes #29743

### DIFF
--- a/crates/runner/mitm-addon/tests/test_request_handler_connector_admission.py
+++ b/crates/runner/mitm-addon/tests/test_request_handler_connector_admission.py
@@ -949,8 +949,22 @@ async def test_test_connector_rejects_stale_unconnected_api_allow_binding(
     assert binding.original_address == ("203.0.113.10", 443)
 
 
-async def test_test_connector_rejects_mismatched_existing_binding(
-    tmp_path, real_flow, mitm_ctx, fake_firewall_headers, headers, monkeypatch
+@pytest.mark.parametrize(
+    ("binding_host", "binding_port"),
+    [
+        pytest.param("api.github.com", 443, id="host-mismatch"),
+        pytest.param("api.vm0.ai", 8443, id="port-mismatch"),
+    ],
+)
+async def test_test_connector_rejects_mismatched_existing_binding_after_verified_tls(
+    tmp_path,
+    real_flow,
+    mitm_ctx,
+    fake_firewall_headers,
+    headers,
+    monkeypatch,
+    binding_host: str,
+    binding_port: int,
 ):
     reg_path = _write_test_oauth_registry(tmp_path)
     flow = real_flow(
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #29754

**Reviewed head:** `ff221e9e9a85190611dc6e067ff7b8a4d4e545c2`

### Summary

The submitted test-only change covers stored host and stored port mismatches through the real connector request hook after verified TLS establishes the current destination. It preserves the positive retargeting case and keeps the mock at the external auth-service boundary.

### Findings

No findings.

### Testing

Local validation passed: uv lock/sync, the focused positive and mismatch selection (3 tests), Ruff format and lint, BasedPyright, and the complete mitm-addon suite (6,670 tests). Unrelated Turbo and Rust suites were not run locally because the diff is confined to the Python addon test package. GitHub CI is evaluated separately by the PR check stage.

### Verdict

LGTM

PR #29758, opened 2026-08-27
perf(guest-session-prune): avoid duplicate codex buffers

### LINKED ISSUE #29739: perf(rust): avoid duplicating Codex compact-generation buffers
## Problem and evidence

`select_codex_compact_generation` can retain the same bounded Codex history suffix in multiple growing `Vec<u8>` buffers:

- `TurnState::bytes` stores every record in the active turn.
- When a `compacted` record is observed, `CandidateState::from_compacting_turn` clones the entire active turn while the original `TurnState` remains live until a delimiter.
- After a candidate exists, later active-turn records are appended to both `TurnState::bytes` and `CandidateState::bytes`.
- After selection, `select_with_limits_and_hook` copies the complete candidate body into the canonical-metadata buffer with `extend_from_slice` while the old candidate body is still live. Removing only the turn/candidate duplication would therefore leave another candidate-sized transient copy on the return path.

The production candidate bound is 64 MiB (`crates/guest-contracts/src/session_history.rs`). A valid near-limit retained generation can consequently create a transient raw-buffer footprint near twice that bound inside the selector, before accounting for the current record (up to 16 MiB), parsed `serde_json::Value`, `Vec` spare capacity, and the caller's later staging or compression allocations.

## Impact

The selector runs in the guest-agent checkpoint path for Codex histories larger than 64 MiB. A long tool-heavy generation can therefore add substantial allocation and copy latency and increase OOM risk precisely while an oversized history is being reduced on a memory-constrained guest.

## Scope and constraints

This issue covers cumulative raw-history buffers owned by the Codex selector, including final candidate assembly. It does not cover the bounded current-record buffer, JSON parsing allocations, caller-owned staging/compression, or changing the public candidate format.

The implementation must preserve:

- byte-for-byte candidate output, including canonical metadata;
- newest-compaction precedence;
- compacting and later-turn validation;
- existing fail-closed reasons and source-change checks;
- the public API and production size bounds.

## Acceptance criteria

- Eliminate simultaneous candidate-sized copies of the retained history suffix throughout selection and final return-value assembly; specifically avoid cloning the compacting turn, dual-appending later turns, and copying the completed candidate body into a second output buffer.
- Keep cumulative selector-owned raw-history storage near one candidate window, apart from the separately bounded current record, canonical metadata, and normal `Vec` capacity slack.
- Preserve the current exact-byte outputs, newest-boundary precedence, turn validation, fail-closed behavior, and public API.
- Add deterministic regression coverage using a near-limit or reduced-limit compacting/later-turn fixture that accounts for peak cumulative retained raw-buffer storage. Do not rely only on flaky process-wide RSS sampling.

seven332:
## Summary

- retain Codex compact-generation bytes in one canonical-prefixed working buffer
- replace newer compact candidates by moving the selected turn in place instead of cloning or dual-appending
- move the completed buffer directly into the result and cover the one-window invariant with a near-limit regression

## Details

Turn and candidate state now hold offsets and validation metadata rather than their own cumulative byte buffers. Completed pre-candidate turns are truncated, newer compact boundaries discard superseded bytes in the same allocation, and exact raw JSONL output remains unchanged.

This does not change public APIs, protocols, persisted formats, ineligible reasons, or guest-agent fallback behavior. The bounded current-record buffer, JSON parsing allocations, caller-owned staging/compression, and allocator-specific capacity/RSS behavior remain outside this change.

## Validation

- `cargo test -p guest-session-prune`
- `cargo test -p guest-agent --lib session_history`
- `cargo clippy -p guest-session-prune --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p guest-session-prune --no-deps`
- `git diff --check`
- `lefthook run pre-commit`

Closes #29739

### DIFF
--- a/crates/guest-session-prune/src/codex.rs
+++ b/crates/guest-session-prune/src/codex.rs
@@ -133,44 +133,26 @@ enum RolloutRecord<'a> {
 
 struct TurnState {
     id: String,
-    bytes: Vec<u8>,
+    start_offset: usize,
     saw_user_message: bool,
     saw_compatible_context: bool,
     invalid: Option<CodexHistoryIneligibleReason>,
 }
 
 impl TurnState {
-    fn new(id: &str, raw_record: &[u8], body_max_bytes: usize) -> Self {
-        let mut state = Self {
+    fn new(id: &str, start_offset: usize) -> Self {
+        Self {
             id: id.to_string(),
-            bytes: Vec::new(),
+            start_offset,
             saw_user_message: false,
             saw_compatible_context: false,
             invalid: None,
-        };
-        state.push(raw_record, body_max_bytes);
-        state
-    }
-
-    fn push(&mut self, raw_record: &[u8], body_max_bytes: usize) {
-        if self.invalid.is_some() {
-            return;
         }
-        let Some(next_size) = self.bytes.len().checked_add(raw_record.len()) else {
-            self.invalidate(CodexHistoryIneligibleReason::CandidateTooLarge);
-            return;
-        };
-        if next_size > body_max_bytes {
-            self.invalidate(CodexHistoryIneligibleReason::CandidateTooLarge);
-            return;
-        }
-        self.bytes.extend_from_slice(raw_record);
     }
 
     fn invalidate(&mut self, reason: CodexHistoryIneligibleReason) {
         if self.invalid.is_none() {
             self.invalid = Some(reason);
-            self.bytes.clear();
         }
     }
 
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #29758

**Reviewed head:** `7f524d493c0eec4f3d06ba6bbdd3fde127a5df8c`

### Summary

The selector now has one cumulative canonical-prefixed raw-byte owner, rebases newer compact turns in place, and moves the completed buffer into the result. Exact-byte, newest-boundary, invalid-candidate recovery, fail-closed, and public compatibility behavior are preserved.

### Findings

No findings.

### Testing

Affected-crate unit/property/integration tests, focused guest-agent session-history tests, formatting, Clippy with warnings denied, rustdoc with warnings denied, diff checks, and pre-commit hooks passed. Full Rust workspace tests were not run because the change is isolated to `guest-session-prune` and its direct consumer with no shared contract change.

### Verdict

LGTM

PR #29783, opened 2026-08-27
fix(guest-agent): bound pi rpc writes on cancellation

### LINKED ISSUE #29745: bug(rust): bound Pi RPC stdin writes during cancellation
## Problem and evidence

Verified on `c890eb7e4e`.

Pi RPC serializes commands as JSONL through a single stdin owner, but command transmission currently awaits bare `write_all(...).await` and `flush().await` calls:

- `crates/guest-agent/src/cli/pi_rpc.rs:476-484`
- `crates/guest-agent/src/cli/pi_rpc.rs:524-580`
- `crates/guest-agent/src/cli/pi_rpc.rs:591-674`

The cancellation-aware waits for prompt and steer start only after their writes complete. The 10-second abort timeout also begins only after the abort command has been written, so it does not bound abort transmission.

When a Pi child keeps stdin open but stops reading, a valid command can fill the bounded OS pipe and leave the writer pending indefinitely. This is reachable with accepted input: the shared active-input contract permits a 1 MiB control payload (`crates/runner/src/active_input.rs` and `crates/api-contracts/src/generated/constants.rs`).

The parent Pi cancellation branch closes active input and cancels the RPC token but deliberately does not immediately arm process-group termination (`crates/guest-agent/src/cli/mod.rs`). The existing bounded SIGTERM/SIGKILL path begins when the stdin writer returns an error. A writer stuck inside the uncancellable write never reaches that handoff, and the post-loop writer-quiescence timeout is not reached while the child remains alive.

## Decision

Proceed with one guest-agent PR. Make normal Pi command transmission cancellation-aware, preserve graceful abort only when the preceding command is known to have completed, and apply the existing 10-second abort budget to both abort transmission and acknowledgement.

- Keep the raw JSONL serializer/writer as the primitive used for graceful abort.
- Add one private cancellation-aware wrapper for normal `get_state`, prompt, and steer writes.
- Bias the write race toward an already-observed cancellation. If cancellation interrupts transmission, return a writer error and drop the sole stdin owner without appending another command. `write_all` does not reveal how many bytes reached the pipe, so the stream must be treated as potentially ending in a partial JSON record.
- Preserve the existing post-write cancellation result while waiting for prompt or steer acknowledgement. Because the prior command is then known complete, that path may still send graceful abort.
- Wrap the raw abort write/flush and acknowledgement wait in one `PI_RPC_ABORT_TIMEOUT` future.
- Reuse current ownership: `deliver_active_input` marks an interrupted steer failed and clears sink-in-flight state; the parent converts writer failure into bounded process-group termination.

This closes the actual unbounded await without introducing a fixed timeout for healthy writes, duplicating process signaling, or adding a new transport subsystem.

## Alternatives considered

- **Immediately force process-group termination for every Pi cancellation:** provides a simple parent bound but weakens/races the existing graceful abort path and would require parent-side active-input state repair.
- **Apply a fixed deadline to every command write:** also bounds non-cancelled slow writers, but adds an unevidenced timeout policy, can create false failures under temporary backpressure, and does not react immediately to cancellation.
- **Introduce a dedicated writer actor:** preserves single ownership but still needs cancellation of its OS write and adds channels, task lifecycle, and shutdown coordination without solving more of the current problem.
- **Status quo:** leaves valid near-limit active input able to defeat user cancellation until an unrelated execution or heartbeat deadline.

## Reproduction scenario

1. Start a mock Pi child that reads and acknowledges `get_state` and the initial prompt.
2. Keep the process and stdout open, but stop consuming stdin after the next steer begins.
3. Deliver a valid near-limit active-input message so the JSONL steer write exceeds pipe capacity.
4. Cancel the run while the write is pending.
5. Without the fix, the writer remains in `write_command`; it neither sends abort nor returns an error that can arm process-group cleanup.

The current focused test `cli::pi_rpc::tests::writer_uses_official_steer_ack_for_active_input_receipts` covers the acknowledged normal path but not stalled stdin.

## Risks and mitigations

- **Cancellation races a write that may have just completed:** prefer cancellation conservatively. Forced cleanup is safer than appending abort after an unobservable partial line, and final attribution remains user cancellation.
- **Regression fixture cancels too early:** synchronize through a real child readiness marker emitted only after steer transmission has begun; do not use sleeps or a mocked writer.
- **Platform-specific pipe capacity:** use a valid near-limit payload and assert behavior, not exact pipe size or elapsed milliseconds.
- **Active-input state remains Writing:** exercise the public execution path and assert no accepted delivery ID or durable delivered receipt after cancellation.
- **Broader cancellation semantic change:** retain graceful abort for cancellations observed after a complete command and keep parent termination policy unchanged.

## Acceptance criteria

- Pi `get_state`, prompt, and steer command writes cannot prevent an observed run cancellation from reaching bounded terminal cleanup.
- Cancellation during a possibly partial command write returns a writer error and does not append abort or any later JSONL command on that stream.
- Cancellation after a completely written request retains graceful abort behavior.
- The graceful abort bound covers writing/flushing the abort command and waiting for its acknowledgement within one 10-second budget; inability to complete it reaches bounded process-group termination.
- An interrupted active-input steer is finalized as failed, with no accepted delivery ID or receipt.
- A public integration test uses a real child that keeps stdin/stdout open, starts a valid near-limit steer, stops reading, observes cancellation, and verifies bounded completion plus child reaping.
- Normal `get_state`, prompt, steer, abort acknowledgement, event projection, and active-input receipt behavior remains unchanged.
- Guest-agent tests, formatting, Clippy, rustdoc, diff checks, and pre-commit checks pass with no skipped tests or lint suppressions.

## Delivery notes

- This issue is a standalone PR-sized implementation issue, not a delivery parent.
- Expected source scope is `crates/guest-agent/src/cli/pi_rpc.rs` plus one Pi integration test.
- No API, database, persisted state, queue payload, generated contract, migration, feature flag, or cross-version rollout change is required.
- A Pi child that acknowledges abort but independently refuses to exit is a separate behavior and remains out of scope.

seven332:
## Summary

- interrupt normal Pi RPC stdin writes when the run is cancelled, without reusing a stream that may contain a partial JSONL command
- bound the complete graceful abort transaction, including the abort write and acknowledgement wait
- add an integration test that fills the active-input payload near its limit, stalls a real child on stdin, cancels the run, and verifies bounded child reaping and cancellation attribution

## Scope

- no API, persisted-state, queue-payload, or Pi protocol changes
- no change to the parent process-group termination policy
- handling a child that acknowledges abort but never exits remains out of scope

## Validation

- `cargo test -p guest-agent --test pi_cli_stalled_stdin_cancellation`
- `env -u OKOU_API_BACKEND_URL -u VM0_API_BACKEND_URL cargo test -p guest-agent`
- `RUSTDOCFLAGS='-D warnings' cargo doc -p guest-agent --no-deps`
- `git diff --check`
- `lefthook run pre-commit`

Closes #29745

### DIFF
--- a/crates/guest-agent/src/cli/pi_rpc.rs
+++ b/crates/guest-agent/src/cli/pi_rpc.rs
@@ -484,6 +484,20 @@ async fn write_command(
     Ok(())
 }
 
+async fn write_command_with_cancellation(
+    stdin: &mut tokio::process::ChildStdin,
+    command: &Value,
+    cancellation: &CancellationToken,
+) -> Result<(), AgentError> {
+    tokio::select! {
+        biased;
+        () = cancellation.cancelled() => Err(AgentError::Execution(
+            "Pi RPC command write was interrupted by cancellation".to_string(),
+        )),
+        result = write_command(stdin, command) => result,
+    }
+}
+
 async fn wait_for_response(
     responses: &mut mpsc::UnboundedReceiver<Value>,
     expected_id: &str,
@@ -527,13 +541,12 @@ async fn abort(
     run_id: &str,
 ) -> Result<(), AgentError> {
     let id = format!("{run_id}:pi:abort");
-    write_command(stdin, &json!({ "id": id, "type": "abort" })).await?;
-    tokio::time::timeout(
-        PI_RPC_ABORT_TIMEOUT,
-        wait_for_response(responses, &id, "abort", true),
-    )
+    tokio::time::timeout(PI_RPC_ABORT_TIMEOUT, async {
+        write_command(stdin, &json!({ "id": id, "type": "abort" })).await?;
+        wait_for_response(responses, &id, "abort", true).await
+    })
     .await
-    .map_err(|_| AgentError::Execution("Pi RPC abort acknowledgement timed out".to_string()))?
+    .map_err(|_| AgentError::Execution("Pi RPC abort timed out".to_string()))?
 }
 
 async fn request_prompt(
@@ -543,13 +556,14 @@ async fn request_prompt(
     message: &str,
     cancellation: &CancellationToken,
 ) -> Result<bool, AgentError> {
-    write_command(
+    write_command_with_cancellation(
         stdin,
         &json!({
             "id": id,
             "type": "prompt",
             "message": message,
         }),
+        cancellation,
     )
     .await?;
     tokio::select! {
@@ -569,13 +583,14 @@ async fn request_steer(
     message: &str,
     cancellation: &CancellationToken,
 ) -> Result<bool, AgentError> {
-    write_command(
+    write_command_with_cancellation(
         stdin,
         &json!({
             "id": id,
             "type": "steer",
             "message": message,
         }),
+        cancellation,
     )
     .await?;
     tokio::select! {
@@ -622,7 +637,12 @@ pub(super) async fn write_commands(
     cancellation: CancellationToken,
 ) -> Result<(), AgentError> {
     let state_id = format!("{run_id}:pi:get-state");
-    write_command(&mut stdin, &json!({ "id": state_id, "type": "get_state" })).await?;
+    write_command_with_cancellation(
+        &mut stdin,
+        &json!({ "id": state_id, "type": "get_state" }),
+        &cancellation,
+    )
+    .await?;
     tokio::select! {
         biased;
         () = cancellation.cancelled() => {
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #29783

**Reviewed head:** `5df3cbd7f15a7eb05b7599d42e115e2ef1533d6f`

### Summary

The change safely makes normal Pi RPC writes cancellation-aware, treats an interrupted JSONL stream as terminal, and places the existing abort budget around both transmission and acknowledgement. It preserves stdin ownership and the parent process-group termination policy.

### Findings

No findings.

### Testing

The real-child stalled-stdin integration test, the complete `guest-agent` suite, Rust formatting, Clippy, rustdoc, diff checks, and staged pre-commit hooks passed. Cancellation after a fully transmitted request followed by a successful graceful abort was not directly forced by a new test; that control flow remains structurally unchanged apart from the abort timeout now enclosing its write.

### Verdict

LGTM

PR #29795, opened 2026-08-27
perf(python): retry only pending provider timing states

### LINKED ISSUE #29788: perf(python): retry only provider timing states with pending reports
## Problem

`crates/runner/mitm-addon/src/provider_timing_store.py:173-183` implements `retry_all_pending()` by iterating every entry in `_run_states` and calling `_admit_retained_locked()`. The fast path at lines 204-206 returns immediately for a state with no pending operations or retained context, but the full dictionary traversal still occurs while the store lock is held.

Both provider stores retain up to 10,000 run states independently:

- `crates/runner/mitm-addon/src/codex_output_timing.py:35,48-52`
- `crates/runner/mitm-addon/src/claude_output_timing.py:35,46-50`

Successful admission clears `pending_operations`, `pending_context`, and the buffered lease at `provider_timing_store.py:229-235`, but deliberately leaves the run state in the LRU so later provider events can preserve first-milestone semantics. Consequently, a long-lived runner can accumulate 20,000 completed, non-retryable states across the two stores.

`crates/runner/mitm-addon/src/runner_flush_lifecycle.py:229-233` sweeps both stores on every runner-triggered usage flush. The runner signals that flush at every job completion (`crates/runner/src/cmd/start/job_spawn.rs:748-750,796-802`), so completed jobs repeatedly rescan all historical timing states even when zero timing reports are pending.

## Impact

On the audit checkout with Python 3.12.3, an exact `ProviderTimingStore.retry_all_pending()` benchmark populated 10,000 states with no pending operations and measured approximately 1.31 ms per sweep. Sweeping both Codex and Claude stores therefore costs about 2.62 ms of GIL-held CPU per runner flush at capacity.

This work scales with historical retained runs rather than current pending reports. At 10 completed-job flushes per second it consumes about 26 ms of CPU per second; at 100 flushes per second it consumes about 262 ms per second. Because each sweep holds the store's `threading.Lock`, provider lifecycle observations for the corresponding store also wait behind the unnecessary traversal.

## Acceptance criteria

- Make provider timing retry cost proportional to retry-eligible states, not all retained LRU states.
- Maintain a bounded pending index/queue or equivalent ownership structure when operations have a complete retained reporting context.
- Remove retry eligibility on successful admission, LRU eviction, explicit discard, and reset, without changing the existing run-state retention needed for first-milestone semantics.
- Preserve current retry order, stop-on-saturation behavior, buffered-report lease ownership, and synchronization with concurrent provider events.
- Add focused coverage with many completed states and a small number of pending states, proving that retry visits only pending work and does not duplicate delivery or leak buffered leases.

seven332:
## Summary

- index retry-eligible provider timing run IDs separately from retained LRU state
- preserve current LRU-derived retry order, stop-on-saturation behavior, and buffered-report lease ownership
- cover sparse pending retries across many completed runs, including recency changes, repeated retries, and duplicate prevention

## Scope

- no API, runner protocol, persisted state, database, configuration, or deployment-order changes
- completed provider timing state remains retained for first-milestone semantics

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (6,674 passed)

Closes #29788

### DIFF
--- a/crates/runner/mitm-addon/src/provider_timing_store.py
+++ b/crates/runner/mitm-addon/src/provider_timing_store.py
@@ -11,6 +11,8 @@ is acquired for incomplete context. Once a buffered-report lease is acquired, sa
 admission keeps the operations, context, and lease together for retry. Successful webhook admission
 transfers delivery ownership to ``usage.webhook``. The store then clears its pending state and
 releases the buffered lease.
+Retry-eligible run IDs are indexed in the pending subset of current LRU order so global retries do
+not visit completed or incomplete-context state.
 Eviction, explicit discard, and reset release any lease that remains locally retained.
 
 See ``codex_output_timing.py`` and ``claude_output_timing.py`` for provider usage, and
@@ -57,7 +59,8 @@ class ProviderTimingStore[StateT: ProviderTimingState]:
     State creation and access maintain the run map's LRU order. Retained operations stay local until
     a complete reporting context and bounded webhook admission are available; queue saturation
     leaves the state and its buffered lease retained, while successful admission hands delivery to
-    ``usage.webhook``.
+    ``usage.webhook``. A bounded ordered index mirrors the retry-eligible subset of the run map, so
+    global retry cost and ordering depend only on current retryable work.
     """
 
     def __init__(
@@ -71,6 +74,7 @@ class ProviderTimingStore[StateT: ProviderTimingState]:
         self._log_type = log_type
         self._max_tracked_runs = max_tracked_runs
         self._run_states: OrderedDict[str, StateT] = OrderedDict()
+        self._retryable_run_ids: OrderedDict[str, None] = OrderedDict()
         self._lock = threading.Lock()
 
     @contextmanager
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #29795

**Reviewed head:** `ffcc49a69f12ad5a31f71f59ccd85871f9af79e5`

### Summary

The shared provider timing store now owns a bounded ordered index of retry-eligible run IDs while retaining completed milestone state in the existing LRU. Index membership follows LRU touches and all admission, eviction, discard, reset, and buffered-lease transitions remain serialized by the existing store lock.

### Findings

No findings.

### Testing

The new integration coverage exercises real Codex provider entry points with 128 completed runs and two pending reports, then verifies sparse visit IDs, updated LRU-derived order, stop-on-saturation, repeated retry, duplicate prevention, and final buffered-lease release. Ruff format/lint, BasedPyright, and the complete 6,674-test mitm-addon suite passed. Production workload profiling under real runner concurrency was not run; correctness does not depend on a wall-clock threshold.

### Verdict

LGTM

PR #29796, opened 2026-08-27
test(runner): eliminate dns port retry race

### LINKED ISSUE #29786: test(runner): eliminate residual DNS port retry race
## Background

Issue #11372 and PR #11381 fixed an earlier flake in `find_available_port_retries_when_udp_candidate_is_in_use` by finding candidate ports where both TCP and UDP could initially bind.

The test still failed in the Rust coverage job for PR #29784:

https://github.com/vm0-ai/vm0/actions/runs/33043810217/job/98423212858

```text
thread 'dns::port::tests::find_available_port_retries_when_udp_candidate_is_in_use' panicked at runner/src/dns/port.rs:124:75:
called `Result::unwrap()` on an `Err` value: Os { code: 98, kind: AddrInUse, message: "Address already in use" }
```

## Problem

The current test creates a free TCP/UDP pair, records the TCP port, drops the UDP probe, and then calls `find_available_port_from`. Dropping the UDP socket before the function binds its replacement creates a release-and-rebind window. On a shared or highly concurrent test host, another process can claim that UDP port, so the supposedly free second candidate also returns `AddrInUse`.

This is separate from the cross-protocol assumption fixed by #11372: both protocols were verified initially, but the UDP reservation is not preserved across the handoff.

The failure was observed during the full coverage suite. A short targeted run may not reproduce it without concurrent port churn.

## Expected behavior

The retry test should deterministically exercise one UDP-busy candidate followed by one successful candidate without relying on an operating-system port remaining free after its reservation is dropped.

## Acceptance criteria

- `find_available_port_retries_when_udp_candidate_is_in_use` has no release-and-rebind race for its successful candidate.
- The test still verifies that an `AddrInUse` UDP result skips the first TCP candidate and selects the next candidate.
- The test passes repeatedly under concurrent local port allocation.
- The affected runner test suite passes.

## Relevant files

- `crates/runner/src/dns/port.rs`

## Related work

- #11372
- #11381
- PR #29784 only exposed this unrelated flake; the DNS fix should not be added to that process-stream PR.

seven332:
## Summary

- make the private DNS reservation loop accept its UDP bind boundary
- keep the successful test candidate's UDP socket reserved and transfer it into the result
- preserve a real `AddrInUse` bind for the first candidate while removing the release-and-rebind race

## Scope

This changes only the private runner DNS port reservation helper and its inline regression test. It does not change the public API, production retry policy, bind address, DNS startup flow, protocols, persistence, configuration, or deployment behavior.

Closes #29786

## Validation

- `cargo test -p runner --bin runner dns::port::tests::find_available_port_retries_when_udp_candidate_is_in_use -- --exact`
- focused regression test repeated sequentially 25 times
- `cargo test -p runner --bin runner dns::port::tests`
- `cargo test -p runner` (3346 passed, 24 ignored)
- `git diff --check`
- repository pre-commit hooks (Cargo fmt, workspace docs, workspace Clippy, file size, commitlint)

### DIFF
--- a/crates/runner/src/dns/port.rs
+++ b/crates/runner/src/dns/port.rs
@@ -38,26 +38,25 @@ impl ReservedTcpPort {
 pub(crate) fn reserve_port() -> std::io::Result<DnsPortReservation> {
     const MAX_PORT_PROBE_ATTEMPTS: usize = 64;
 
-    reserve_port_from((0..MAX_PORT_PROBE_ATTEMPTS).map(|_| ReservedTcpPort::bind(0)))
+    reserve_port_from(
+        (0..MAX_PORT_PROBE_ATTEMPTS).map(|_| ReservedTcpPort::bind(0)),
+        |port| std::net::UdpSocket::bind(("0.0.0.0", port)),
+    )
 }
 
-#[cfg(test)]
-fn find_available_port_from<I>(tcp_candidates: I) -> std::io::Result<u16>
+fn reserve_port_from<I, F>(
+    tcp_candidates: I,
+    mut bind_udp: F,
+) -> std::io::Result<DnsPortReservation>
 where
     I: IntoIterator<Item = std::io::Result<ReservedTcpPort>>,
-{
-    Ok(reserve_port_from(tcp_candidates)?.port())
-}
-
-fn reserve_port_from<I>(tcp_candidates: I) -> std::io::Result<DnsPortReservation>
-where
-    I: IntoIterator<Item = std::io::Result<ReservedTcpPort>>,
+    F: FnMut(u16) -> std::io::Result<std::net::UdpSocket>,
 {
     let mut last_addr_in_use = None;
     for tcp in tcp_candidates {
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #29796

**Reviewed head:** `1cd3437c370be41d93160bb093af1e7700eb4c61`

### Summary

The private UDP bind seam removes the successful candidate's release-and-rebind window while preserving production bind behavior, retry classification, and socket ownership. The regression test retains real TCP/UDP sockets, observes a real first-candidate `AddrInUse`, and transfers the still-bound successful UDP socket into the reservation.

### Findings

No P0, P1, or P2 findings.

### Testing

Local validation passed: formatting, focused test plus 25 sequential repetitions, DNS port tests, full runner suite (`3346 passed; 24 ignored`), runner Clippy with all targets/features and warnings denied, runner docs, diff hygiene, and repository pre-commit hooks. GitHub Actions CI remains independently verified by the workflow check stage.

### Verdict

LGTM

PR #29863, opened 2026-08-27
fix(runner): await multipart upload cancellation

### LINKED ISSUE #29836: fix(runner): wait for multipart upload cancellation before abort
## Problem and evidence

`crates/runner/src/r2_cache/multipart.rs` schedules bounded concurrent `upload_part` tasks in a Tokio `JoinSet`. On the first upload error or task join error, `upload_parts_streaming_with` currently returns immediately through `?` and relies on dropping the `JoinSet`.

Tokio 1.53.1 aborts remaining tasks when a `JoinSet` is dropped, but it does not wait for those tasks to finish shutting down. Tokio provides `JoinSet::shutdown()` specifically to abort all tasks and join them until the set is empty. The caller starts the best-effort R2 multipart abort as soon as the scheduler returns, so a sibling `upload_part` future can still be shutting down while the multipart session is aborted.

Existing tests cover successful bounded scheduling and a single-part upload error. They do not prove cancellation completion, preservation of the original error, admission stopping at the terminal failure, or the ordering between upload cleanup and multipart abort.

## Decision

Treat this as a production cancellation-ordering bug, not only a coverage gap. On the first terminal scheduler error, stop admitting parts, cancel every remaining upload task, wait for those tasks to finish shutting down, and then return the original error so the caller can abort the multipart session.

## Scope and rationale

Keep the change within the existing multipart scheduler and its tests. No public API, storage format, configuration, or retry behavior needs to change. The failure path may wait for cooperative task cancellation, which is required to make the documented cancellation-safety contract true before R2 abort begins.

## Alternatives considered

- Keep relying on `JoinSet` drop: rejected because drop requests cancellation but does not wait for task shutdown.
- Call `abort_all()` without draining the set: rejected for the same ordering gap.
- Broaden the work into a multipart scheduler redesign: rejected because the existing bounded scheduler is sound on success; only terminal cleanup ordering is unsupported.

## Acceptance criteria

- When any upload task returns an error or fails to join, stop admitting new parts.
- Cancel all remaining upload tasks and wait until every sibling upload future has been dropped before returning from the scheduler.
- Preserve and return the original upload or join error after sibling cleanup.
- Add a deterministic concurrent fault-injection test with at least two in-flight parts, one failing part, one blocked sibling, and a queued part that must never be admitted after the failure.
- At the full upload boundary, assert the failed multipart upload performs exactly one abort and never calls `complete_multipart_upload`.
- Use synchronization primitives and a bounded timeout; do not rely on sleeps.

## Delivery notes

This is one independently reviewable runner PR. It requires no migration or cross-version rollout handling because it changes only local failure cleanup ordering.

seven332:
## Summary

- wait for all sibling `upload_part` tasks to finish cancellation before returning a terminal multipart scheduler error
- preserve the original upload or join error while stopping any further part admission
- cover both upload errors and task join failures with deterministic concurrent cleanup assertions
- assert the public upload path aborts once without completing

## Scope

This changes only the runner's R2 multipart failure cleanup. Successful scheduling, concurrency, part ordering, retry behavior, public APIs, persisted data, and protocols are unchanged.

## Validation

- `cargo test -p runner r2_cache::multipart::tests::multipart_scheduler_drops_siblings_before_returning_terminal_errors -- --exact`
- `cargo test -p runner r2_cache::tests::upload::upload_part_missing_etag_errors_with_part_number -- --exact`
- `cargo test -p runner` (3345 passed, 24 ignored; guest inventory 1 passed)
- pre-commit hooks: cargo fmt, Clippy, docs, file-size, and commitlint

Closes #29836

### DIFF
--- a/crates/runner/src/r2_cache/multipart.rs
+++ b/crates/runner/src/r2_cache/multipart.rs
@@ -169,8 +169,18 @@ where
         // Drain at least one completion. JoinSet returns None only when
         // empty, which our outer loop condition prevents.
         if let Some(joined) = tasks.join_next().await {
-            let (pn, part) = joined.map_err(|e| R2Error::Io(io_other(e)))??;
-            parts.push((pn, part));
+            let result = match joined {
+                Ok(result) => result,
+                Err(error) => Err(R2Error::Io(io_other(error))),
+            };
+            match result {
+                Ok((pn, part)) => parts.push((pn, part)),
+                Err(error) => {
+                    // Finish cancelling uploads before the caller aborts the multipart session.
+                    tasks.shutdown().await;
+                    return Err(error);
+                }
+            }
         }
     }
 
@@ -308,7 +318,7 @@ mod tests {
 
     use tokio::{
         io::{AsyncRead, ReadBuf},
-        sync::{Notify, mpsc},
+        sync::{Barrier, Notify, mpsc},
     };
 
     use super::*;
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #29863

**Reviewed head:** `ed84632511f2f4f35a068657bbbd34847ef9995e`

### Summary

The scheduler now waits for every sibling upload task to finish cancellation before returning either an upload error or a task join error, so the caller cannot start multipart abort while an `upload_part` future remains alive. Successful scheduling behavior is unchanged.

### Findings

No findings. The earlier submitted-head coverage gap for `JoinError` cleanup is resolved on this head by a deterministic second scenario that also verifies the original panic information is preserved.

### Testing

The regression uses a barrier, guarded reader, sibling drop signal, and bounded timeout without sleeps. The public upload test verifies one abort and zero complete calls through the AWS SDK mock boundary and a real temporary source file. Focused tests, the runner suite (3345 passed, 24 ignored), guest inventory, Rust formatting, all-target/all-feature Clippy, workspace docs, and pre-commit hooks passed locally. Live Cloudflare R2 was not exercised; repository integration tests use AWS Smithy mocks for that external boundary.

### Verdict

LGTM

PR #29871, opened 2026-08-27
test(api): fix pi resource snapshot fixture size mismatch

### LINKED ISSUE #29822: test(api): fix Pi resource snapshot fixture size mismatch
## Problem and evidence

The API test suite fails on a clean `main` checkout when Pi API first-turn tests load resource snapshot archives. The focused reproduction fails with `PI_API_RESOURCE_INVALID`: a 96-byte response is checked against the exact Storage `archiveSize` of 3397.

Production behavior is correct. The Pi loader requires every downloaded archive to match its persisted length before extraction.

## Root cause

The test fixture loses S3 object identity in two places:

1. Generic API test helpers collapse every presigned download into one fixed URL, which can also be persisted in the shared system-storage URL cache.
2. Pi tests serve `latestStoredArchive()`, selecting one upload by recency for every mount.

After exact GET URLs were prototyped, the focused test exposed a second fixture boundary: `dev-seed.ts` inserts `computer-use` and `gen` Storage metadata, but the current test worker never uploads those external S3 objects. They therefore cannot be recovered from `s3.send` history.

## Decision

Preserve exact bucket/key identity before URLs enter the shared cache and resolve each Pi download from one of two exact test sources:

- a matching current-test `PutObjectCommand` body; or
- a matching committed database-seed object fixture.

The seeded-object fixture builds a valid discoverable skill tar.gz from committed seed name/frontmatter and pads it to the authoritative captured archive length. It never selects an object by size or upload order.

## Chosen direction

- Add one typed API-test S3 URL formatter:
  - `GetObjectCommand` uses the existing archive path plus exact `${bucket}/${key}` query identity.
  - Non-GET commands keep `https://r2.example.com/upload?sig=test`.
- Install it in global mock reset, `acceptAgentStorageWrites()`, and `acceptStorageDownloads()`.
- Add a focused seeded-system-skill archive fixture keyed by exact committed S3 archive key.
- Replace `latestStoredArchive()` with an exact resolver: current-test upload first, seeded object second, unknown identity failure.
- Preserve the explicit resource HTTP 503 case.
- Keep production cache and archive validation unchanged.

## Alternatives considered

- **Pi-local presigner only:** cache hits bypass it.
- **Mutate or expire shared cache rows:** unsafe under concurrent workers.
- **Choose bytes by size/order or rewrite metadata:** loses identity and can hide defects.
- **Commit opaque captured binaries:** legacy tar output is not byte-reproducible and is costly to maintain.
- **Replace system mounts with test-owned resources:** changes the production-like resource graph and adds unrelated fixture plumbing.

## Risks and mitigations

- Generic GET mock behavior has many consumers. Preserve the existing path and all non-GET URLs, then run the full API suite.
- Seeded archives contain representative `SKILL.md` content rather than the full remote body. Detailed extraction/discovery remains covered by the Pi resource snapshot service tests.
- Reject generated archives larger than the captured length and assert every returned fixture has the exact authoritative length.
- Unknown S3 identities fail immediately; no size or recency fallback is allowed.
- Do not mutate shared cache/database state for isolation.

## Acceptance criteria

- The focused `deepseek-v4-flash` reproduction passes.
- GET presigned URLs preserve exact bucket/key identity through cache insertion and reuse; non-GET upload URLs remain unchanged.
- Every Pi resource response resolves by exact S3 identity from either a current-test upload or a committed seeded object fixture.
- Returned archives remain valid gzip/tar bodies and match authoritative Storage archive lengths.
- The intentional resource-download failure still produces `PI_API_RESOURCE_INVALID` before model invocation.
- Relevant Pi tests, the full API suite with `--maxWorkers=4`, and API format/lint/type/Knip checks pass.
- Production archive validation, cache behavior, schema, contracts, and deployment surfaces remain unchanged.

## Scope boundary

Do not change connector behavior, production API behavior, database schema/state, or deployment contracts as part of this issue.

seven332:
## Summary

- preserve the S3 bucket/key identity in API test presigned GET URLs
- serve Pi resource archives by their exact uploaded or seeded object key
- synthesize valid, deterministic archives for seeded system skills at their authoritative seeded sizes

## Scope

- test infrastructure only; production storage behavior is unchanged
- upload presigned URLs and existing user flows are unchanged
- unknown seeded object keys fail instead of falling back by archive size or request order

## Validation

- `pnpm -F api lint`
- `pnpm -F api check-types`
- `pnpm knip --workspace api`
- `pnpm -F api exec vitest run src/signals/routes/__tests__/chat-events.bdd.test.ts --maxWorkers=1 --silent=passed-only` (128 passed)
- `pnpm -F api exec vitest run src/signals/routes/__tests__/chat-events.bdd.test.ts -t 'runs the Pi API first turn once for deepseek-v4-flash and resumes canonical JSONL' --maxWorkers=1 --silent=passed-only` (passed twice, including cache reuse)
- `pnpm -F api exec vitest run src/signals/routes/__tests__/storage-presigned-url-cache.test.ts --maxWorkers=1 --silent=passed-only` (10 passed)
- pre-commit hook: Prettier, repository Knip, repository TypeScript checks, file-size check, and commitlint passed
- `pnpm -F api exec vitest run --maxWorkers=4 --silent=passed-only` was attempted three times locally. The first exposed a stale local schema; after migration and then a clean local DB reset/seed, later attempts still encountered varying cross-file 5-second timeouts and database-pool cascades. The focused Pi and storage-cache suites above remained green; CI is the authoritative full-suite run.

Closes #29822

### DIFF
--- a/turbo/apps/api/src/__tests__/mocks.ts
+++ b/turbo/apps/api/src/__tests__/mocks.ts
@@ -1185,6 +1185,25 @@ export function getApiTestMocks(): ApiTestMocks {
   return apiTestMocks;
 }
 
+export function apiTestS3PresignedUrl(command: unknown): string {
+  const candidate = command as {
+    readonly constructor?: { readonly name?: string };
+    readonly input?: { readonly Bucket?: unknown; readonly Key?: unknown };
+  };
+  if (candidate.constructor?.name !== "GetObjectCommand") {
+    return "https://r2.example.com/upload?sig=test";
+  }
+  const bucket = candidate.input?.Bucket;
+  const key = candidate.input?.Key;
+  if (typeof bucket !== "string" || typeof key !== "string") {
+    throw new Error("Expected S3 GET presign request to identify an object");
+  }
+  const url = new URL("https://r2.example.com/storage/archive.tar.gz");
+  url.searchParams.set("sig", "bdd");
+  url.searchParams.set("object", `${bucket}/${key}`);
+  return url.toString();
+}
+
 export function resetApiTestMocks(): void {
   apiTestMocks.abortSignal.timeout.mockReset();
   apiTestMocks.ably.channelGet.mockReset();
@@ -1232,8 +1251,10 @@ export function resetApiTestMocks(): void {
   apiTestMocks.clerk.m2m.createToken.mockReset();
   apiTestMocks.s3.send.mockReset();
   apiTestMocks.s3.getSignedUrl.mockReset();
-  apiTestMocks.s3.getSignedUrl.mockResolvedValue(
-    "https://r2.example.com/upload?sig=test",
+  apiTestMocks.s3.getSignedUrl.mockImplementation(
+    (_client: unknown, command: unknown) => {
+      return Promise.resolve(apiTestS3PresignedUrl(command));
+    },
   );
   apiTestMocks.s3.clientConfig.mockReset();
   apiTestMocks.dns.lookupOverrides.clear();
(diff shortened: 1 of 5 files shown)


seven332:
## Code Review: PR #29871

**Reviewed head:** `0a5c411ad1fa7c99b7185c12c8a0e04f022b4754`

### Summary

The test fixture now preserves exact S3 object identity and supplies valid, correctly sized archives for committed seeded system skills without changing production behavior or upload flows.

### Findings

No findings.

### Testing

API lint, type checks, Knip, the 128-test chat-events suite, repeated focused Pi coverage, the 10-test storage presign cache suite, and pre-commit checks passed. The full API suite did not complete green locally because repeated runs encountered varying cross-file timeouts and database-pool cascades; CI remains the authoritative full-suite verification.

### Verdict

LGTM

PR #29904, opened 2026-08-27
docs(runner): clarify storage cache gc limit accounting

### LINKED ISSUE #29876: docs(rust): clarify storage cache GC limit accounting
## Problem and evidence

The storage cache documentation describes `STORAGE_CACHE_MAX_BYTES` and `STORAGE_CACHE_MAX_ENTRIES` as caps enforced by `gc_storage_cache`, and describes the collector as bounding the storage directory to those limits. The implementation and regression tests show that these are best-effort eviction targets during one GC pass:

- Every real version directory increments `total_entries`, but entries whose per-version lock is held or whose lock probe fails are not added to `total_size` or to the eviction candidate list.
- Recent entries are measured and included in `total_size`, but the grace window excludes them from eviction candidates.
- The eviction loop can exhaust eligible candidates while protected or unmeasured entries remain, so the configured thresholds are not unconditional post-GC invariants.
- Revalidation can find a candidate locked, recent, missing, replaced, or otherwise unremovable after the scan. The result then retains or removes the candidate's accounting according to what was safely observed rather than asserting a complete disk inventory.
- The final `remaining_bytes` log value is derived from the measured entries that remain in the accounting model. It excludes entries that could not be safely measured, so it is a lower bound when those entries exist.
- Existing tests cover recent-entry retention, held-lock protection, and candidates becoming locked or recent during deletion revalidation.

This mismatch can cause maintainers and operators to read the documentation as a strict storage guarantee and treat `remaining_bytes` as complete disk usage, even though the locking and grace-window design intentionally permits a cache to remain above a target after one pass.

## Decision

Proceed with a documentation clarification. The issue is necessary and correctly scoped as one independently reviewable documentation change. The clarification must describe the limits as best-effort eviction targets, explain the protected and unmeasured cases, and preserve the current locking, grace-window, eviction, accounting, and telemetry behavior.

## Alternatives considered

- **Keep the current wording:** rejected because it states a stronger invariant than the verified implementation provides.
- **Change GC to guarantee strict caps or count every locked entry's bytes:** outside this issue and inconsistent with the current safety boundary that avoids measuring entries while they may be actively used. It would change runtime behavior rather than correct the documented contract.
- **Add new telemetry or a separate usage metric:** not necessary to resolve the documented-contract mismatch; the existing `remaining_bytes` field can be qualified as a lower bound.

## Risks and mitigations

- Wording that calls the value a complete inventory would preserve the current operational ambiguity; explicitly call it a lower bound when entries are unmeasured.
- Wording that implies protected entries are always absent from byte accounting would be inaccurate for recent entries; distinguish recent measured entries from held-lock and lock-probe-error entries.
- Wording that omits deletion revalidation could still imply that the scan is authoritative; document that candidates may become locked or recent before deletion.

## Acceptance criteria

- Document the byte and entry limits as best-effort eviction targets rather than unconditional post-GC caps.
- State that recent, held-lock, and lock-probe-error entries can leave the cache above either target after one GC pass.
- Clarify that byte accounting and the logged `remaining_bytes` exclude entries that could not be safely measured, so the value is a lower bound when such entries exist.
- Keep the documentation consistent with candidates that become locked or recent during deletion revalidation.
- Preserve the current locking, grace-window, eviction, accounting, and telemetry behavior.

## Plan decision

The selected direction is Option 2 from the Research/Options/Plan phases: establish the complete contract across the existing storage-GC documentation surfaces in `crates/runner/src/cmd/gc/storage.rs`. The implementation will update the constant comments, `gc_storage_cache` documentation, adjacent accounting comments, and the explanation next to the `remaining_bytes` log construction as needed to express the verified behavior. The `remaining_bytes` key and value calculation remain unchanged.

The implementation boundary is one standalone PR changing documentation/comments in `storage.rs`. No new documentation page, test, `GcReport` field, runtime statement, telemetry key, cache behavior, or storage-cache producer code is included. The existing regression tests remain the behavioral reference. Validation is Rust formatting, targeted runner storage-GC tests, runner Clippy, and runner documentation generation as specified in the Plan Phase.

## Delivery notes

This is a standalone, single-PR issue with no parent, child issues, or blocking dependencies. The plan is complete and implementation-ready. The PR should link issue #29876 without closing a parent or unrelated issue.

seven332:
## Summary

- Clarify that storage-cache byte and entry limits are best-effort eviction targets, not unconditional post-GC caps.
- Document how recent, held-lock, and lock-probe-error entries affect one-pass cleanup and byte accounting.
- Explain that `remaining_bytes` is a lower bound when entries could not be safely measured, including deletion-time revalidation cases.

Closes #29876

## Scope

This PR changes only Rust documentation and comments in `crates/runner/src/cmd/gc/storage.rs`. It preserves the GC algorithm, locking, grace window, eviction behavior, accounting calculation, tests, telemetry keys, and the `remaining_bytes` value calculation.

## Validation

- `cargo fmt --all -- --check` — passed.
- `cargo check -p runner --tests` — passed with `CARGO_BUILD_JOBS=1`, `CARGO_PROFILE_TEST_DEBUG=0`, and a private workspace `TMPDIR`.
- `cargo clippy -p runner --all-targets --all-features` — passed with the same low-resource settings.
- `cargo doc -p runner --all-features --no-deps` — passed with the same low-resource settings.
- `cargo test -p runner cmd::gc::storage::tests` — attempted with normal and low-resource settings; both attempts were terminated with exit 137 while linking the runner test binary in the 3.8 GiB, no-swap environment. The source/test compilation completed without a diagnostic, and the equivalent test-target compile check passed above.

### DIFF
--- a/crates/runner/src/cmd/gc/storage.rs
+++ b/crates/runner/src/cmd/gc/storage.rs
@@ -17,12 +17,14 @@ use super::filesystem::{
 use super::lock_file::{LockProbe, probe_lock};
 use super::report::GcReport;
 
-/// Per-host storage archive cache size cap. Enforced by `gc_storage_cache`
-/// as an LRU by `<version>/` dir mtime.
+/// Per-host storage archive cache byte target for best-effort eviction.
+/// `gc_storage_cache` may leave the cache above this target when protected or
+/// unmeasured entries leave no safe eviction candidate.
 const STORAGE_CACHE_MAX_BYTES: u64 = 1 << 30; // 1 GiB
-/// Per-host storage archive cache entry cap. The byte cap alone does not
-/// bound many tiny storage versions, and each cached version also creates a
-/// lock file.
+/// Per-host storage archive cache entry target for best-effort eviction. The
+/// byte target alone does not bound many tiny storage versions, and each
+/// cached version also creates a lock file. Recent, held-lock, and
+/// lock-probe-error entries can leave this target exceeded after one pass.
 const STORAGE_CACHE_MAX_ENTRIES: u64 = 5_000;
 
 /// Eligible `<version>` directory discovered during the scan phase.
@@ -56,7 +58,7 @@ impl From<&std::fs::Metadata> for StorageDirectoryIdentity {
 struct StorageEvictionResult {
     freed: u64,
     /// Candidate contribution to keep in `total_size` after this attempt.
-    /// `None` removes the scan-time size from the cap calculation.
+    /// `None` removes the scan-time size from this pass's measured accounting.
     remaining_size: Option<u64>,
     /// Candidate contribution to keep in `total_entries` after this attempt.
     /// Dry-runs set this to false to model the real deletion while leaving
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #29904

**Reviewed head:** `66deb4b6e4aaa79fcfaca6bf88eff430506f3090`

### Summary

The comment-only Rustdoc update accurately describes the existing storage-cache GC behavior, including best-effort targets, protected or unmeasured entries, lower-bound byte accounting, and deletion-time revalidation. The patch does not alter executable code, telemetry keys, or tests.

### Findings

No findings.

### Testing

Formatting, runner test-target compilation, Clippy, and runner documentation generation passed. The targeted storage-cache test binary was attempted with normal and low-resource settings; both attempts exited 137 while linking in the 3.8 GiB, no-swap environment, without a compiler diagnostic. CI execution remains the unverified area.

### Verdict

LGTM

PR #29918, opened 2026-08-27
perf(python): bound codex catalog query parsing

### LINKED ISSUE #29887: perf(python): bound Codex catalog query parsing before materialization
## Problem

At audited `main` (`fad505ecaf6efc6d68ba80a6e5affe093498be69`), revalidated unchanged at current `main` (`fd6e791529b7b2ef0d99d8d3f46aaa6b552704d2`), Codex model-catalog request preparation fully decodes and materializes every query field before confirming that the cache accepts exactly one small `client_version` field:

- `crates/runner/mitm-addon/src/codex_model_catalog_cache.py:306-331` calls `urllib.parse.parse_qsl(..., strict_parsing=True)` without `max_num_fields`, then checks `len(query) == 1` and applies the 128-byte `client_version` limit only after the complete list and decoded strings exist.
- `crates/runner/mitm-addon/src/codex_model_catalog_cache.py:584-596` reaches that helper for every authenticated request whose path is `/backend-api/codex/models`, even when the query shape will be classified as `model_catalog_bypass`.
- `crates/runner/mitm-addon/src/mitm_addon.py:1435-1463` performs this preparation synchronously on the shared mitmproxy event loop after ordinary model-provider credential resolution and before the request continues upstream.

The sandbox controls the request target. Pinned mitmproxy 12.2.3 buffers an HTTP/1 request head without a completed-head size limit (`mitmproxy/proxy/layers/http/_http1.py:281-320` still contains `FIXME: protect against header size DoS`), and its request-line parser does not impose a target-length limit. A request can therefore carry many irrelevant query fields to the catalog path. The cache eventually bypasses the malformed shape, but only after `parse_qsl()` has allocated one tuple and decoded strings for every field.

## Verified evidence and impact

I executed the exact audited (and current-main-equivalent) `_split_expected_url()` and `_catalog_url()` function bodies extracted from the source AST on CPython 3.12.3. Input construction was outside the timed region; timings are 21-run medians with GC disabled, and peak allocation is from a separate `tracemalloc` call.

| Raw query characters | Query fields | Median synchronous time | Peak traced allocation |
| ---: | ---: | ---: | ---: |
| 65,536 | 21,839 | 17.809 ms | 2.39 MiB |
| 1,048,576 | 349,519 | 302.731 ms | 39.48 MiB |

Each query consisted of repeated irrelevant `x=` fields followed by the otherwise valid `client_version=0.145.0`, so `_catalog_url()` returned `None` only after full materialization. Absolute timing varies by interpreter and host—the deployed addon pins CPython 3.14.0—but the unbounded field/list allocation and synchronous event-loop boundary are deterministic.

Repeated authenticated catalog-shaped requests can therefore consume a proxy event-loop core and create large transient allocation while still being semantically rejected as cache bypasses, delaying unrelated sandbox flows in the same runner process.

## Distinctness

- #25015 moved catalog **response-body validation** off the event loop; it does not cover request-query parsing before the first suspension.
- #27717 bounds decoding of the private catalog prefetch **header**; it does not bound the request target.
- #26551 bounds `parse_qsl()` work in optional connector diagnostics, a separate helper and lifecycle.
- #26514 bounds query-pair cardinality for managed AWS SigV4 requests, not Codex model-catalog cache admission.
- Open-title and all-state keyword searches found no issue covering this catalog query boundary.

## Acceptance criteria

- Establish explicit raw-query byte and field-count work bounds for Codex catalog cache admission before `parse_qsl()` decodes or materializes the complete field list.
- Keep the raw-query budget large enough for every valid `client_version` whose decoded UTF-8 representation is at most 128 bytes, including percent-encoded representations.
- Reject over-budget and multi-field catalog queries as the existing `request_url` cache bypass without rejecting or otherwise changing the upstream HTTP request.
- Preserve current canonical URL, percent-decoding, control-character, blank-value, malformed-input, telemetry, credential-partitioning, and cache behavior for accepted inputs.
- Add structural coverage proving an oversized or high-cardinality query is bypassed before complete query-pair materialization; do not use wall-clock assertions.
- Keep a proxy-wide request-target limit out of this focused change; that requires separate cross-protocol policy and compatibility analysis.
- Focused catalog hook/lifecycle tests and the complete mitm-addon static checks and test suite pass.

seven332:
## Summary

- bound Codex catalog cache query admission to the exact 427-byte maximum valid encoded request
- reject multi-field queries through `parse_qsl(max_num_fields=1)` before pair materialization and percent-decoding
- preserve canonical cache identity and upstream `request_url` bypass behavior with integration coverage for encoded, oversized, multibyte, malformed-byte, and high-cardinality queries

## Scope

- keeps the synchronous catalog hook lifecycle and existing telemetry unchanged
- does not add a proxy-wide request-target limit or change mitmproxy request-head buffering

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (7183 passed)

Closes #29887

### DIFF
--- a/crates/runner/mitm-addon/src/codex_model_catalog_cache.py
+++ b/crates/runner/mitm-addon/src/codex_model_catalog_cache.py
@@ -88,7 +88,16 @@ _FIREWALL_NAME = "model-provider:codex-oauth-token"
 _CATALOG_HOST = "chatgpt.com"
 _CATALOG_PATH = "/backend-api/codex/models"
 _RESPONSES_PATH = "/backend-api/codex/responses"
+_CLIENT_VERSION_QUERY_NAME = "client_version"
 _MAX_CLIENT_VERSION_BYTES = 128
+_PERCENT_ENCODED_CHARACTERS_PER_BYTE = 3
+MAX_CATALOG_QUERY_FIELDS = 1
+# The name and value may encode every byte as %XX; the separating "=" stays literal.
+MAX_CATALOG_QUERY_BYTES = (
+    _PERCENT_ENCODED_CHARACTERS_PER_BYTE
+    * (len(_CLIENT_VERSION_QUERY_NAME.encode()) + _MAX_CLIENT_VERSION_BYTES)
+    + 1
+)
 _MAX_ETAG_BYTES = 512
 _MAX_CONTENT_TYPE_BYTES = 256
 _MAX_JSON_NESTING = 128
@@ -307,15 +316,23 @@ def _catalog_url(original_url: str) -> str | None:
     parsed = _split_expected_url(original_url)
     if parsed is None or parsed.path != _CATALOG_PATH:
         return None
+    raw_query = parsed.query
+    # Bound the allocation used to reconstruct mitmproxy's surrogate-escaped path bytes.
+    if (
+        len(raw_query) > MAX_CATALOG_QUERY_BYTES
+        or len(raw_query.encode("utf-8", "surrogateescape")) > MAX_CATALOG_QUERY_BYTES
+    ):
+        return None
     try:
         query = urllib.parse.parse_qsl(
-            parsed.query,
+            raw_query,
             keep_blank_values=True,
             strict_parsing=True,
+            max_num_fields=MAX_CATALOG_QUERY_FIELDS,
         )
     except ValueError:
         return None
-    if len(query) != 1 or query[0][0] != "client_version":
+    if len(query) != MAX_CATALOG_QUERY_FIELDS or query[0][0] != _CLIENT_VERSION_QUERY_NAME:
         return None
     client_version = query[0][1]
     if (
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #29918

**Reviewed head:** `ebfc0c8d7cb164e155fe9ef68942009741f4964d`

### Summary

The catalog cache now derives an exact raw-query budget, bounds request-target byte reconstruction, and applies the standard library's one-field guard before pair materialization. Accepted canonicalization, credential partitioning, telemetry, cache lifecycle, and upstream continuation remain unchanged.

### Findings

No findings. The self-review's malformed-byte compatibility coverage gap was fixed on this head.

### Testing

Passed uv lock/sync, Ruff format and lint, BasedPyright, and the complete mitm-addon suite (7183 tests). Additional bounded-input equivalence checks matched the previous parser behavior. CI and the intentionally out-of-scope proxy-wide request-head limit remain outside this local review.

### Verdict

LGTM

Issue #29792, opened 2026-08-27
test(runner): measure pre-agent startup contributors


seven332:
## Parent objective

This is a measurement slice of #29631. The parent requires decision-grade, same-host evidence before implementing the evidence-gated timezone optimization in #29644 or private-write batching in #29646.

## Background

#29643 and #29748 established an authenticated Guest Agent readiness boundary and added a real-metal benchmark for fresh, workspace-cache, and exact-reuse startup paths.

The current benchmark proves readiness-path correctness and reports the final `start_agent_process` stages. However, its `agent_ready_ms` clock starts immediately before `Sandbox::start_agent_process`, so it does not expose the full Runner executor-to-Agent-ready aggregate. The bounded sample output also omits the earlier configured-timezone and private-file write stages that determine whether #29644 or #29646 passes its implementation gate.

The existing evidence uses three samples per path. That is suitable for a behavior smoke test, but it is not enough to establish whether a prospective aggregate improvement exceeds same-host benchmark noise, especially at p95 and p99.

## Requirement

Extend the production-path real-metal benchmark with bounded raw measurements for the full executor-to-Agent-ready critical path and the pre-Agent operations needed to decide #29644 and #29646.

Keep this slice measurement-only. Do not replace timezone exec, batch private writes, change protocol behavior, or move required work past Agent readiness.

## Acceptance criteria

- Report the actual Runner executor-to-Agent-ready aggregate separately from the existing `start_agent_process`-local shell-spawn and Agent-ready durations.
- Exercise a fresh startup with a validated configured timezone and report the `runner_guest_timezone_sync` duration and its share of the executor-to-Agent-ready aggregate.
- Exercise a fixture where both optional user environment and required run payload are present.
- Report the number and duration of connector-context, user-environment, and run-payload private-file operations separately, preserving connector best-effort versus required-file fatal semantics in the measurement labels.
- Preserve strict fresh, workspace-cache, and exact-reuse classification for every sample.
- Collect enough same-metal-host samples to report meaningful p50, p90, p95, and p99 distributions, sample counts, and failures; do not treat a three-sample maximum as decision-grade p95/p99 evidence.
- Emit bounded low-cardinality raw samples without secrets, file contents, user-provided values, paths, sandbox IDs, or unbounded error text.
- Post the resulting baseline to this issue and use it to decide #29644 and #29646 independently.
- If an operation is negligible relative to benchmark noise, close its optimization issue with the evidence instead of expanding the protocol.
[...]

PR #29943, opened 2026-08-27
perf(runner): batch required private guest writes

### LINKED ISSUE #29646: perf(runner): batch required private guest writes
## Parent objective

This is delivery Slice 4 of #29631. Reduce exact-reuse Runner-to-Agent-ready latency by eliminating one repeated private-file helper lifecycle for the two required Guest Agent bootstrap inputs.

## Evidence and root cause

The completed #29792 local-1 baseline measured 100 strictly classified samples per startup path with zero failures. Every sample contained separate connector-context, user-environment, and run-payload private writes.

For exact reuse, the full Executor-to-Agent-ready boundary was 81/89/90/95 ms at p50/p90/p95/p99. The required user-environment plus run-payload writes consumed 8.400/12.420/15.692/22.808 ms, or 10.43%/16.78%/18.59%/27.47% of that boundary. Batching can save at most the shorter required write per sample: 3.895/4.223/4.385/10.566 ms, or 4.69%/6.00%/6.71%/12.73% of the full boundary.

Fresh and workspace-cache paths have smaller upper bounds (about 1.4–1.9% of the full boundary), so exact reuse is the primary optimization target and the other paths are regression checks. The implementation gate is satisfied, but the upper bound is not an expected result: final same-host before/after measurements must establish the realized aggregate gain beyond noise.

The repository already has a typed bounded `write_files` operation for ordinary files. Private runtime files still use separate `write_private_file` operations, each of which launches an owned `guest-write-file` helper. The missing capability is a typed private multi-file operation that reuses the established batch transport and lifecycle while preserving private path, ownership, permission, failure, cancellation, and connection-reuse semantics.

## Required behavior

Serialize the optional user environment and required run payload before their guest I/O. When both files exist and fit the typed batch contract, write them through one private multi-file operation. When the user environment is empty, write only the run payload through the existing single private-file operation; do not create a one-entry batch.

Keep connector account context on its current separate best-effort private write. Its successful path is inserted into the user environment, while its failure remains non-fatal and must not be hidden inside the required-file batch.

The private batch operation must:

- use the existing batch transport/lifecycle where possible, with an explicit private batch policy rather than ordinary create-parent semantics;
- validate every path and bound entry count, per-entry content, aggregate content, and protocol-frame size before sending a request;
- create or replace each target with runtime-private parent and file semantics, including symlink rejection and the expected owner/mode;
- report one terminal success only after every entry completes; any serialization, validation, helper, timeout, cancellation, connection, or partial-progress failure remains fatal before Agent start;
- preserve the current frame-write cancellation boundary and fail-closed connection-reuse behavior; and
- preserve the existing single-file chunking behavior for inputs that cannot fit the bounded batch request instead of introducing unbounded allocation or multi-frame batch state.

Runner and its guest binaries are one deployed artifact and each sandbox is owned by the Runner process that created it, so this internal protocol addition does not require an old/new guest fallback. Workspace-cache persisted formats are unchanged.

## Acceptance criteria

- The PR links #29792's measurement evidence and records same-host before/after results for exact reuse, with fresh and workspace-cache regression checks.
- With non-empty user environment, Runner emits one private batch request containing exactly the user-environment and run-payload files instead of two independent required helper lifecycles.
- With empty user environment, Runner emits only the required single private run-payload write.
- Connector context remains a separate best-effort private write; connector failure does not fail Agent startup, while required-file serialization or write failure does.
- On success, both required files are complete at their expected paths with private parents, owner, and mode before Agent start.
- Partial progress cannot produce successful Agent startup. Operation guards and owned process-group cleanup preserve terminal completion, sandbox parking, and connection-reuse contracts.
- Protocol, host, guest helper, sandbox implementations/mocks, and Runner integration tests cover private policy selection, bounds, success, helper/per-entry failure, partial progress, timeout, cancellation/caller drop, connection loss, teardown/reuse, and zero/one/two required-file cases at their public entry points.
- Relevant Rust format, lint, docs, and tests pass.
- Same-host results show an aggregate exact-reuse gain beyond benchmark noise without unexplained failures or fresh/workspace-cache tail regression. If they do not, do not merge the optimization; record the result on the issue.

## Scope boundaries

- Parent: #29631
- Dependency completed: #29643; exact-reuse evidence completed in #29792.
- Do not batch connector context without independently preserving its non-fatal result.
- Do not change session-history, storage-cache staging, or persisted workspace-cache formats.
- Do not add a persistent writer service or relax runtime-private path/permission validation.

## Relevant files

- `crates/runner/src/executor/agent_run.rs`
- `crates/runner/src/executor/env.rs`
- `crates/sandbox/src/sandbox.rs`
- `crates/sandbox/src/types.rs`
- `crates/sandbox-fc/src/sandbox.rs`
- `crates/sandbox-mock/`
- `crates/vsock-proto/src/payloads/write_file.rs`
- `crates/vsock-host/src/file/write.rs`
- `crates/vsock-guest/src/handlers.rs`
- `crates/vsock-guest/src/connection.rs`
- `crates/guest-write-file/`

seven332:
## Summary

- add a typed `write_private_files` vsock request that reuses the existing bounded multi-file payload and result format
- preserve private parent, permission, symlink, truncation, cancellation, and oversized-file fallback semantics across the guest helper, host transport, and sandbox interface
- write the required user-environment and run-payload files in one private batch when both are present; keep the connector context as a separate best-effort write and keep the empty-user-env path as one private write
- replace the two required-write telemetry actions with one `runner_required_private_files_write` action

## Performance

Compared with the 100-sample baseline in #29792 on `local-1.aws.vm3.ai`. The final after run used one dedicated Runner invocation with `max_concurrent=1`, `max_idle=1`, strict path classification, 100 samples per path, and zero failures. Measurement-only instrumentation was removed from this PR.

All values are p50/p90/p95/p99.

| Path / metric | Before | After |
| --- | --- | --- |
| exact reuse — required private files | 8.400 / 12.420 / 15.692 / 22.808 ms | 4.957 / 8.358 / 8.799 / 9.877 ms |
| exact reuse — Executor to Agent ready | 81 / 89 / 90 / 95 ms | 74 / 85 / 90 / 95 ms |
| workspace cache — required private files | 12.243 / 13.521 / 14.327 / 16.292 ms | 5.256 / 6.446 / 6.734 / 9.990 ms |
| workspace cache — Executor to Agent ready | 387 / 411 / 427 / 449 ms | 378 / 406 / 417 / 427 ms |
| fresh — required private files | 12.256 / 13.130 / 13.469 / 14.770 ms | 5.061 / 6.332 / 7.940 / 11.737 ms |
| fresh — Executor to Agent ready | 420 / 442 / 448 / 478 ms | 407 / 443 / 455 / 477 ms |

The primary exact-reuse target improves by 7/4/0/0 ms across the full boundary, while its required-write distribution improves by about 41.0% / 32.7% / 43.9% / 56.7%. Workspace-cache improves across the full boundary. Fresh p95 is 7 ms higher, but the unchanged Agent-start-to-ready p95 is also 7 ms higher while required-write p95 is 5.529 ms lower; fresh p90 is effectively flat and p99 is 1 ms lower. This is downstream distribution noise rather than a private-write regression.

- ordinary `write_files` wire bytes and behavior are unchanged
- Runner and guest binaries ship together, so this internal request does not need old-guest negotiation
- connector-context batching remains out of scope because its failure policy is independently best effort
- workspace-cache persisted formats are unchanged

## Validation

- `cargo test -p vsock-proto -p guest-write-file -p vsock-guest -p vsock-host -p sandbox -p sandbox-mock -p sandbox-fc -p runner --all-features`
- `cargo clippy -p vsock-proto -p guest-write-file -p vsock-guest -p vsock-host -p sandbox -p sandbox-mock -p sandbox-fc -p runner --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS='-D warnings' cargo doc -p vsock-proto -p guest-write-file -p vsock-guest -p vsock-host -p sandbox -p sandbox-mock -p sandbox-fc -p runner --all-features --no-deps`
- `cargo check -p runner --all-targets --all-features`
- production-style Runner deployment and real runs on `local-1`
- 100-sample fresh, workspace-cache, and exact-reuse Agent-ready benchmarks on one corrected dedicated Runner invocation
- required CI gates: Turbo, Crates, and Security passed

Closes #29646

### DIFF
--- a/crates/guest-write-file/src/lib.rs
+++ b/crates/guest-write-file/src/lib.rs
@@ -16,7 +16,7 @@ struct Args {
     path: Option<PathBuf>,
 }
 
-const USAGE: &str = "usage: guest-write-file [--private] [--append | --create-parents] [--] <path> | guest-write-file --batch";
+const USAGE: &str = "usage: guest-write-file [--private] [--append | --create-parents] [--] <path> | guest-write-file --batch [--private]";
 
 fn parse_args<I>(args: I) -> Result<Args, String>
 where
@@ -65,10 +65,8 @@ where
     }
 
     if batch {
-        if append || create_parents || private {
-            return Err(
-                "--batch cannot be used with --append, --create-parents, or --private".to_string(),
-            );
+        if append || create_parents {
+            return Err("--batch cannot be used with --append or --create-parents".to_string());
         }
         if path.is_some() {
             return Err("--batch does not accept a path".to_string());
@@ -77,7 +75,7 @@ where
             append: false,
             batch: true,
             create_parents: false,
-            private: false,
+            private,
             path: None,
         });
     }
@@ -101,7 +99,7 @@ where
 
 fn run(args: Args, mut stdin: impl Read) -> io::Result<()> {
     if args.batch {
-        return run_batch(stdin);
+        return run_batch(stdin, args.private);
     }
     let Some(path) = args.path else {
         return Err(io::Error::new(io::ErrorKind::InvalidInput, "missing path"));
@@ -123,7 +121,7 @@ fn run(args: Args, mut stdin: impl Read) -> io::Result<()> {
     file.flush()
 }
 
-fn run_batch(mut stdin: impl Read) -> io::Result<()> {
+fn run_batch(mut stdin: impl Read, private: bool) -> io::Result<()> {
     let mut payload = Vec::new();
     let max_payload_size = vsock_proto::MAX_MESSAGE_SIZE - vsock_proto::MIN_BODY_SIZE;
     let mut limited = stdin.by_ref().take(max_payload_size as u64 + 1);
@@ -139,12 +137,16 @@ fn run_batch(mut stdin: impl Read) -> io::Result<()> {
 
     for file in files {
         let path = Path::new(file.path);
-        if let Some(parent) = path.parent()
-            && !parent.as_os_str().is_empty()
-        {
-            fs::create_dir_all(parent)?;
-        }
-        let mut output = open_output_file(path, false)?;
+        let mut output = if private {
+            open_private_output_file(path, false)?
+        } else {
+            if let Some(parent) = path.parent()
+                && !parent.as_os_str().is_empty()
+            {
+                fs::create_dir_all(parent)?;
+            }
+            open_output_file(path, false)?
+        };
         output.write_all(file.content)?;
         output.flush()?;
     }
@@ -236,7 +238,7 @@ fn prepare_output_file(_file: &File) -> io::Result<()> {
 ///
 /// ```text
 /// guest-write-file [--private] [--append | --create-parents] [--] <path>
-/// guest-write-file --batch
+/// guest-write-file --batch [--private]
 /// ```
 ///
 /// Use `--` before `<path>` when the literal path begins with `-`. `stdin`
@@ -251,7 +253,9 @@ fn prepare_output_file(_file: &File) -> io::Result<()> {
 /// parent components.
 ///
 /// `--batch` reads a `vsock-proto` `write_files` payload from stdin and writes
-/// every ordinary file entry with create-parent and truncate semantics.
+/// every entry with create-parent and truncate semantics. Combined with
+/// `--private`, every entry uses the private runtime-file behavior described
+/// above.
 ///
 /// Returns process-style exit codes: `0` for success, `1` for runtime or write
 /// failures, and `2` for usage or argument errors.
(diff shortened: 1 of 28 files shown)


seven332:
## Code Review: PR #29943

**Reviewed head:** `7143c03c63ffa7101bf32cba614bd2f4bf8a871e`

### Summary

The dedicated private-batch request reuses the bounded multi-file payload without changing ordinary batch bytes, preserves the existing private helper policy and sequential oversized fallback, and keeps connector-context failure independent from the two required Agent inputs. The Runner integration emits one truthful required-write action and fails before Agent start on required batch errors.

### Findings

No findings.

### Testing

Reviewed protocol/guest/host/Sandbox/Runner behavior and tests at the current head. Rust format, matching-package Clippy and rustdoc, Runner all-target check, and the complete matching-package test command passed, including 3367 Runner tests.

Final production-style `local-1` evidence used one corrected dedicated Runner invocation, strict classification, 100 samples per path, and zero failures. Executor-to-Agent-ready changed from 81/89/90/95 ms to 74/85/90/95 ms for exact reuse, from 387/411/427/449 ms to 378/406/417/427 ms for workspace cache, and from 420/442/448/478 ms to 407/443/455/477 ms for fresh. The fresh p95 variance is in the unchanged Agent-start stage while the required-write p95 improves by 5.529 ms.

### Verdict

LGTM

PR #29941, opened 2026-08-27
fix(platform): wait for org role before revealing home layout

seven332:
## Summary

- keep the home bootstrap skeleton visible until the current organization role and its layout commit together
- preserve only the admin header spacer during later role refreshes while hiding authorization-dependent controls until the current role resolves
- make the Playwright coverage delay organization role and Slack status independently, and synchronize member layout assertions on the skeleton lifecycle

## Validation

- `pnpm format`
- `pnpm -F @okouai/app lint`
- `pnpm -F @okouai/app check-types`
- `pnpm -F @okouai/app exec vitest run src/views/okou-page/__tests__/home-growth-entry.test.tsx src/views/okou-page/__tests__/home-route.test.tsx --maxWorkers=1 --silent=passed-only` (21 tests)
- `pnpm knip`
- `pnpm check-types` (from `e2e/`)
- pre-commit hook: Prettier, Knip, full Turbo `check-types`, file-size check, and commitlint

## Exclusions

- The Playwright feature test was not run locally because its configuration targets an externally deployed environment that does not contain this branch; PR preview CI will exercise it against the submitted code.

Closes #29935

Issue #29935, opened 2026-08-27
bug(platform): prevent home layout shift while org role loads


seven332:
## Background

PR #29723 moved the home growth entry out of normal flow so delayed Slack status resolution would not shift the landing content. PR #29830 restored the established 56px desktop header offset by adding an admin-only `h-14` spacer.

The spacer is currently rendered only after the asynchronous organization role resolves. The landing content can render before that happens.

## Problem

On an admin home page, `/api/org` and `/api/integrations/slack` resolve independently. While `isOrgAdmin$` is still loading, `GrowthEntryHeader` returns `null`, so the desktop `main` element starts at `y = 0`. When the role resolves to `admin`, React inserts the 56px spacer and moves the already-visible landing content to `y = 56`.

This is both a user-visible layout shift and the cause of an intermittent Playwright failure:

```text
home content stays fixed while the growth entry loads
Expected: 56
Received: 0
```

The existing test delays Slack status but does not synchronize organization-role resolution. It therefore observes either side of the independent role-loading race.

## Reproduction

1. Open an authenticated admin chat home page at a desktop viewport.
2. Delay `GET /api/org` while allowing the landing content to render.
3. Observe that `main` is initially positioned at `y = 0`.
4. Resolve the organization response with `role: "admin"`.
5. Observe the 56px spacer appear and move `main` to `y = 56`.

The timing is intermittent in CI because the real organization request usually resolves before the first geometry measurement.

## Expected behavior

- The desktop admin landing content must not become visible at `y = 0` and later move to its established `y = 56` position.
- Delayed Slack status resolution must not change the landing-content geometry after the admin role is known.
- Desktop non-admin landing content must keep its established `y = 0` position.
- Mobile layout must remain unchanged.

## Acceptance criteria

- Add deterministic page-level coverage for a delayed organization-role response on the admin home page.
- Keep the existing delayed Slack-status coverage, but synchronize it with the organization-role boundary it assumes.
- Preserve the final desktop geometry for both admin (`y = 56`) and non-admin (`y = 0`) users.
- Avoid using cached or stale role data to guess authorization-dependent layout.
- Applicable formatting, lint, type, static-analysis, platform integration, and Playwright checks pass.

## Relevant files

- `turbo/apps/platform/src/views/okou-page/growth-entry.tsx`
- `turbo/apps/platform/src/signals/org.ts`
- `turbo/apps/platform/src/views/okou-page/agent-chat-page.tsx`
[...]

PR #30014, opened 2026-08-28
fix(platform): reserve growth entry space before loading

hulh122:
## Summary

- reserve the desktop 56px growth-entry slot from the first render, before organization role or Slack status resolves
- keep the actual growth controls admin-only and absolutely positioned, preserving the loaded composer position and mobile layout
- cover stable page geometry before role resolution, after role resolution, and after `Invite humans 🤝` appears

## Context

The existing spacer mounts only after `/api/org` resolves to an admin role. That moves the already-visible avatar, tagline, and composer from `y = 0` to `y = 56`.

This PR intentionally applies the newly selected invariant that every desktop role reserves the same slot. It is related to #29935 and the overlapping #29941, but has no merge-order dependency on that PR.

## Verification

- `npx --yes prettier@3.8.1 --check turbo/apps/platform/src/views/okou-page/growth-entry.tsx e2e/playwright/tests/chat.spec.ts`
- `cd e2e && npx --yes pnpm@10.33.4 check-types`
- `cd e2e/playwright && npx --yes pnpm@10.33.4 exec playwright test tests/chat.spec.ts --list` (22 tests discovered)
- `cd turbo && npx --yes pnpm@10.33.4 -F @okouai/app lint`
- `cd turbo && npx --yes pnpm@10.33.4 -F @okouai/app check-types`
- `cd turbo && npx --yes pnpm@10.33.4 -F @okouai/app exec vitest run src/views/okou-page/__tests__/home-growth-entry.test.tsx --maxWorkers=1 --silent=passed-only` (7 tests passed)
- `cd turbo && npx --yes pnpm@10.33.4 knip --workspace @okouai/app`

Local browser/E2E execution was not run because repository instructions prohibit starting a local dev server unless explicitly requested; the geometry regression runs in the PR pipeline.

### DIFF
--- a/e2e/playwright/tests/chat.spec.ts
+++ b/e2e/playwright/tests/chat.spec.ts
@@ -1249,12 +1249,28 @@ test("chat page displays tagline after onboarding", async ({ page }) => {
   });
 });
 
-test("home content stays fixed while the growth entry loads", async ({
+test("home content keeps its reserved offset while the growth entry loads", async ({
   page,
 }) => {
   await page.setViewportSize({ width: 1440, height: 900 });
+  const orgRequested = deferred();
+  const releaseOrg = deferred();
   const slackStatusRequested = deferred();
   const releaseSlackStatus = deferred();
+  await page.route(
+    (url) => url.pathname === "/api/org",
+    async (route) => {
+      if (route.request().method() !== "GET") {
+        await route.continue();
+        return;
+      }
+      orgRequested.resolve();
+      await releaseOrg.promise;
+      await route.fulfill({
+        json: { id: "org_admin", name: "Admin Org", role: "admin" },
+      });
+    },
+  );
   await page.route("**/api/integrations/slack", async (route) => {
     if (route.request().method() !== "GET") {
       await route.continue();
@@ -1284,7 +1300,7 @@ test("home content stays fixed while the growth entry loads", async ({
 
   await page.goto(appUrl);
   await page.waitForURL(/\/agents\/[^/]+\/chat\/?$/, { timeout: 30_000 });
-  await slackStatusRequested.promise;
+  await orgRequested.promise;
 
   const growthEntry = page.getByTestId("growth-entry");
   const main = page.locator("main");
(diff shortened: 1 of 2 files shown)


hulh122:
LGTM


PR_AUTO_REVIEW_COMMIT: f70f0e1e15fa706db41085ae8b7424406a6b7147
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

This change keeps the existing 56px desktop home offset after the admin growth entry resolves, but reserves it from the first render and for non-admin organizations as the selected invariant. The actual growth controls remain absolutely positioned and admin-only, while the unchanged `md:block` breakpoint preserves the existing mobile behavior. The Playwright regression test independently holds organization-role and Slack-status resolution and verifies that the main content and tagline do not move across either transition.

### Review Criteria

| Criterion | Source | Result | Evidence |
| --- | --- | --- | --- |
| Correctness, edge cases, and scope | `REVIEW.md` | Pass | The same `h-14` spacer remains in flow for the resolved admin state and is now unconditional on desktop; only the admin control remains conditional. Loading, admin, non-admin, and mobile breakpoint behavior are explicit. |
| Regression coverage for a fix | `REVIEW.md` | Pass | The browser test reproduces the layout before role resolution, after role resolution but before Slack resolution, and after the entry appears; a separate member-role case locks the chosen 56px invariant. |
| Type and implementation hygiene | `docs/bad-smell.md` | Pass | No casts, `any`, suppressions, dynamic imports, hardcoded configuration, or fallback paths are introduced. |
| External behavior and integration boundary | `docs/testing.md` | Pass | The test observes rendered DOM attachment and geometry through the real page while controlling only the HTTP dependency boundary. |
| Effect ownership and render purity | `docs/effect.md` | Pass | The production change is synchronous JSX composition only; it adds no effects, async execution, subscriptions, or lifecycle ownership. |
| Cache and resource lifecycle | `docs/cache.md` | Pass | No cache, ref, handler registration, or owned resource is added or changed. |
| ccstate usage | `.claude/skills/ccstate/SKILL.md` | Pass | The component continues to read the existing `isOrgAdmin$` loadable; no new signal, command, signal factory, or execution path is introduced. Unresolved role state now gates only the control, not page geometry. |
| Standard testing pattern | `docs/testing/patterns.md` | Pass | The regression is a page-level Playwright integration test and intercepts actual browser requests rather than mocking internal modules. |
| Testing anti-patterns | `docs/testing/anti-patterns.md` | Pass | Assertions cover the user-visible layout outcome across real async transitions; there are no internal mocks, fake timers, mock-call assertions, or CSS-class assertions. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- Coverage: Adequate. The regression test covers both async dependencies separately, the fully loaded admin state, and the non-admin desktop state.
- Conventions: Compliant. It tests browser-visible outcomes at the network boundary without introducing internal mocks or fake timers.
- CI: All currently published PR checks are successful or intentionally skipped, including lint, type checks, app tests, security scans, and the aggregate Turbo gate.

PR #30042, opened 2026-08-28
perf(ably-subscriber): reuse tokens across reconnect failures

### LINKED ISSUE #30012: perf(rust): reuse fresh Ably tokens across failed reconnects
## Problem

At audited commit `97414e6c34b2241df1cbcf87fa85fa6248cf41d6`, every failed fresh Ably reconnect discards a successfully exchanged token and repeats both token HTTP requests on the next retry.

The reconnect loop calls `attempt_reconnect` on every attempt and retains no pending token state when that function returns an error (`crates/ably-subscriber/src/connection/event_loop.rs:496-613`). Inside `attempt_reconnect`, any non-resumable connection (`!use_resume`) or token-expired reconnect calls `get_token`, exchanges the returned request with Ably, and keeps the resulting `TokenDetails` only in a local `new_token` (`crates/ably-subscriber/src/connection/event_loop.rs:1013-1033`). That token is committed only after the WebSocket connects and the channel reaches an attached or suspended outcome (`crates/ably-subscriber/src/connection/event_loop.rs:569-602`). A later failure in `connect_pending`, the CONNECTED handshake, ATTACH send, or attach wait therefore drops the freshly issued token; the next loop iteration executes the same token acquisition again.

In the runner integration, `get_token` itself performs a vm0 API request to create an Ably `TokenRequest` (`crates/runner/src/provider/api_ably_supervisor.rs:1022-1036`, `crates/runner/src/provider/api.rs:1135-1150`), and `exchange_token` performs a second request to Ably REST. Once the 120-second resume window expires, suspended retries run every 30 seconds by default (`crates/ably-subscriber/src/types.rs:270-282`), so a Realtime/WebSocket outage with the token endpoints still reachable produces two avoidable HTTP requests per runner per retry for the duration of the outage.

## Impact

- A fleet-wide Ably Realtime outage or blocked WebSocket path amplifies into sustained load on both the vm0 realtime-token endpoint and Ably's token-exchange endpoint.
- Token signing, JSON serialization, TLS/HTTP work, and Ably token issuance repeat even though the immediately preceding attempt already obtained an unexpired bearer token.
- The extra control-plane dependency can delay recovery or create rate-limit pressure during the exact outage window when many runners retry together.

## Acceptance criteria

- Retain and reuse a successfully exchanged, still-valid `TokenDetails` across subsequent transport/handshake/attach failures instead of reacquiring it on every retry.
- Invalidate the pending token and obtain a new one when Ably rejects it as expired/invalid, when its expiry or renewal boundary is reached, or when another explicit authorization condition requires refresh.
- Preserve the rule that session connection metadata is committed only after a definitive reconnect outcome; pending token reuse must not make a failed transport look connected.
- Add reconnect coverage where token acquisition succeeds, several WebSocket attempts fail, and the token provider/exchange are called once until invalidation; separately verify that an auth rejection or expiry causes a fresh acquisition.

seven332:
## Summary

- retain freshly exchanged Ably tokens within one reconnect episode so transport, handshake, and attach failures do not repeat token-provider and REST exchange requests
- replace pending tokens at their fixed renewal boundary or after Ably reports a token-specific `40140..40149` rejection, including resume-handshake and reconnect-attach failures
- keep session token and connection metadata commits behind the existing definitive reconnect outcomes

## Scope

- adds no public API, persisted state, runner protocol, migration, or deployment-order dependency
- does not change retry cadence, initial subscription authentication, capability-denial handling, or token-renewal failure policy

## Validation

- `cd crates && cargo clippy -p ably-subscriber --all-targets --all-features`
- `cd crates && cargo doc -p ably-subscriber --all-features --no-deps`
- `cd crates && cargo test -p ably-subscriber`
- renewal-boundary integration test repeated 10 times
- `git diff --check`
- `lefthook run pre-commit`

Closes #30012

### DIFF
--- a/crates/ably-subscriber/src/connection/event_loop.rs
+++ b/crates/ably-subscriber/src/connection/event_loop.rs
@@ -495,6 +495,7 @@ pub(crate) async fn run_event_loop(mut p: EventLoopState, mut close_rx: CloseRec
 
         let mut retry_immediately = immediate_retry;
         let mut disconnected_retry_count: u32 = 0;
+        let mut pending_reconnect_token: Option<PendingReconnectToken> = None;
         loop {
             if p.session.should_enter_suspended_retry()
                 && !enter_suspended_retry(&mut p, &mut close_rx).await
@@ -563,7 +564,11 @@ pub(crate) async fn run_event_loop(mut p: EventLoopState, mut close_rx: CloseRec
                     }
                     continue;
                 }
-                result = attempt_reconnect(&mut p, refresh_token) => result,
+                result = attempt_reconnect(
+                    &mut p,
+                    &mut refresh_token,
+                    &mut pending_reconnect_token,
+                ) => result,
             };
 
             match reconnect_result {
@@ -660,6 +665,11 @@ enum ReconnectOutcome {
         transport: WsTransport,
     },
     Closed,
+}
+
+struct PendingReconnectToken {
+    details: TokenDetails,
+    renewal_at: Option<Instant>,
 }
 
 async fn send_attach(p: &mut EventLoopState, close_rx: &mut CloseReceiver) -> LoopAction {
@@ -785,7 +795,7 @@ async fn handle_message(
             let refresh_token = msg
                 .error
                 .as_ref()
-                .is_some_and(|error| error.code == error_code::TOKEN_EXPIRED);
+                .is_some_and(|error| error_code::is_token_error(error.code));
             let reason = Some(protocol_disconnect_reason(msg.error));
             p.session.notify_protocol_disconnected();
             close_websocket_transport(p);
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #30042

**Reviewed head:** `d701792c6ebcd61454e8d48e42d6c39f544a7654`

### Summary

The loop-owned pending token matches the approved reconnect scope: it survives transient reconnect failures, expires at a fixed renewal boundary, is discarded after token-specific Ably rejection, and is committed only after a definitive reconnect outcome. The public API, retry cadence, persistence, and deployment contracts remain unchanged.

### Findings

No findings.

### Testing

The affected crate passed formatting, all-target/all-feature Clippy, documentation, and all 208 tests. The 16 reconnect integration tests also passed again against this submitted head; they cover repeated transport failures, handshake and attach rejection, renewal-boundary replacement, provider/exchange counts, and restored message delivery. The secret-backed Ably smoke test remains a CI gate.

### Verdict

LGTM

PR #30046, opened 2026-08-28
perf(runner): bound gc service discovery

### LINKED ISSUE #30011: perf(rust): avoid serial systemctl queries in runner GC
## Problem

At audited commit `97414e6c34b2241df1cbcf87fa85fa6248cf41d6`, `runner gc` discovers image references from enabled runner services by awaiting two `systemctl` queries for each matching unit, strictly serially.

`protected_image_refs_for_gc` always runs enabled-service discovery before image deletion (`crates/runner/src/cmd/gc/image_refs.rs:86-100`). `enabled_runner_service_config_paths_with_reader` then scans every `vm0-runner-*.service` entry and, inside the directory loop, awaits `service::is_unit_enabled` followed by `service::read_unit_config_path` for each enabled unit (`crates/runner/src/cmd/gc/image_refs.rs:139-214`). These helpers execute `systemctl is-enabled` and `systemctl --no-pager cat` respectively (`crates/runner/src/cmd/service/systemctl.rs:512-529`, `crates/runner/src/cmd/service/unit_config.rs:8-13`). Both queries inherit the 10-second `SYSTEMCTL_QUERY_TIMEOUT` (`crates/runner/src/cmd/service/systemctl.rs:16`, `crates/runner/src/cmd/service/systemctl.rs:398-416`).

For `N` enabled runner units, this phase launches `2N` child processes but never overlaps them. It can therefore take close to `20N` seconds when both queries return near their deadlines, or `10N` seconds when each enablement query times out. The current scenario coverage creates only one runner unit (`crates/runner/src/cmd/gc/image_refs/tests.rs:428-470`), so it does not detect serialized multi-unit discovery.

## Impact

- Hosts that retain several runner versions can spend minutes in service-reference discovery before any image, NBD, workspace, log, or version GC phase begins.
- Runner image preparation invokes `runner gc --keep-latest 6` while holding the exclusive deployment GC lock (`.github/scripts/prepare-runner-image.sh:202-207`), so serial systemd latency directly extends deployment-critical work and blocks other GC invocations.
- Even healthy hosts pay two external process spawns and output captures per enabled runner service.

## Acceptance criteria

- Discover matching unit names first, then query their enablement and effective config with fixed bounded concurrency or a bounded batched systemd operation; do not introduce unbounded child-process fan-out.
- Preserve the fail-closed `inventory_complete` behavior for directory iteration, enablement, and effective-config failures, including deterministic protected-reference results.
- Keep disabled units from contributing config references and retain effective fragment/drop-in parsing for enabled units.
- Add boundary coverage with multiple delayed units proving that queries overlap within the fixed bound and that the concurrency limit is respected.

## Related work

- #29880 addressed the analogous serial `systemctl cat` inventory in `runner doctor`; GC has a separate discovery loop and remains serial.

seven332:
## Summary

- enumerate matching runner service units before issuing systemd queries
- process each unit's enablement and effective-config lookup with ordered concurrency bounded at four
- preserve fail-closed inventory aggregation, disabled-unit exclusion, effective drop-in parsing, and existing warnings
- add a gated fake-systemctl scenario proving overlap, the fixed admission bound, per-unit query order, and mixed failure handling

## Scope

- no CLI, API, protocol, persisted-state, migration, deployment-order, dependency, timeout, or production logging changes

## Validation

- `cargo test -p runner cmd::gc::image_refs::tests -- --nocapture`
- `cargo test -p runner`
- `cargo clippy -p runner --all-targets -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p runner --no-deps`
- `git diff --check`
- pre-commit Rust format, documentation, Clippy, file-size, and commitlint hooks

Closes #30011

### DIFF
--- a/crates/runner/src/cmd/gc/image_refs.rs
+++ b/crates/runner/src/cmd/gc/image_refs.rs
@@ -1,6 +1,7 @@
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::path::{Path, PathBuf};
 
+use futures_util::{StreamExt, stream};
 use tracing::warn;
 
 use crate::cmd::service;
@@ -12,6 +13,8 @@ use super::filesystem::{GcDirEntryReader, read_dir_or_missing};
 use super::versions::VersionGcAnalysis;
 
 type ProtectedImageRefMap = HashMap<String, HashSet<String>>;
+
+const ENABLED_SERVICE_QUERY_CONCURRENCY: usize = 4;
 
 pub(super) enum ProtectedImageRefs {
     Complete(ProtectedImageRefMap),
@@ -131,6 +134,11 @@ struct EnabledRunnerServiceConfigPaths {
     inventory_complete: bool,
 }
 
+struct EnabledRunnerServiceConfigPathResult {
+    path: Option<PathBuf>,
+    inventory_complete: bool,
+}
+
 async fn enabled_runner_service_config_paths(system_dir: &Path) -> EnabledRunnerServiceConfigPaths {
     let mut entry_reader = GcDirEntryReader::new();
     enabled_runner_service_config_paths_with_reader(system_dir, &mut entry_reader).await
@@ -159,7 +167,7 @@ async fn enabled_runner_service_config_paths_with_reader(
         };
     };
 
-    let mut paths = Vec::new();
+    let mut units = Vec::new();
     let mut inventory_complete = true;
     loop {
         let entry = match entry_reader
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #30046

**Reviewed head:** `86cf0fba9875a1e68d5c7dab807a0dcafff2e1dc`

### Summary

The submitted change replaces serial GC service discovery with an ordered four-way per-unit pipeline while preserving the existing systemctl authority, disabled-unit handling, encounter order, warnings, and fail-closed inventory contract.

### Findings

No findings.

### Testing

Focused GC image-reference tests and the full runner suite pass. Rust formatting, Clippy with warnings denied, rustdoc with warnings denied, diff whitespace checks, and pre-commit hooks pass. The new external-boundary scenario proves four concurrent blocked children, rejects fifth admission before release, verifies enablement-before-config ordering, excludes a disabled unit, and keeps a mixed config failure incomplete.

Remote runner deployment/E2E was not run locally; CI owns those checks.

### Verdict

LGTM

PR #30059, opened 2026-08-28
perf(mitm-addon): bound x billing unicode label work

### LINKED ISSUE #30037: perf(python): bound unique Unicode label work in X URL billing
## Problem

On current `main` at `895b70cc33b936090afc7c86a4c71bf82a591324` (with the relevant mitm-addon files unchanged from the audited `8e9d333effada141917385e0482bcc35fd419fab`), X tweet-body billing can perform one IDNA normalization for every distinct Unicode label in a long post:

- `crates/runner/mitm-addon/src/usage/providers/connectors/x_billing.py:407-453` splits one bare-domain candidate into labels and keeps per-call normalization/prefix dictionaries. Those dictionaries avoid repeated work only when labels repeat; every distinct Unicode label still calls `_label_has_tld_prefix_before_unicode()` and `normalize_idna_label()`.
- `crates/runner/mitm-addon/src/usage/providers/connectors/x_billing.py:456-466` runs that classifier for each bare-domain regex candidate.
- `crates/runner/mitm-addon/src/usage/providers/connectors/x_billing.py:485-497` invokes the scan while deciding whether a successful `POST /2/tweets` body may be downgraded from `content.create_with_url` to `content.create`.
- `crates/runner/mitm-addon/src/usage/providers/connectors/x.py:658-670` performs the refinement synchronously during connector usage reporting, and `crates/runner/mitm-addon/src/mitm_addon.py:1873-1875` calls that reporting directly from the response hook.

The decoded inspection boundary is 64 KiB (`crates/runner/mitm-addon/src/body_limits.py:22-25`), but that still permits thousands of distinct labels. X's official changelog documents support for longform Posts of 25k characters: https://docs.x.com/changelog#aug-20-2024.

The current tests cover a 25,000-character dotted ASCII candidate, one long Unicode label, and many repeated Unicode labels (`crates/runner/mitm-addon/tests/x_connector_usage/test_writes.py:839-943`). They do not cover many distinct Unicode labels, which is the case where both local dictionaries grow with the input and cannot reuse normalization work.

## Reproduction and impact

Using the addon's pinned Python 3.14.0 runtime with imports warmed, this input stays below the production body-inspection cap and forces the classifier to examine the entire candidate:

```python
text = "a." + "".join(chr(0x4E00 + index) + "." for index in range(12_000))
assert len(text.encode()) == 48_002
assert not _tweet_text_likely_contains_url(text)
```

On the audit container (Linux aarch64), seven direct calls to the real function measured:

- 25,000-character dotted ASCII input: 6.98 ms median (6.96 ms minimum)
- 24,002-character / 48,002-byte distinct-CJK-label input: 120.32 ms median (117.07 ms minimum)

A separate instrumented run recorded 12,001 `normalize_idna_label()` calls and 8,109,237 bytes of peak traced allocation. Absolute timings are environment-dependent; the deterministic issue is that normalization count and two per-call dictionaries scale to thousands of unique attacker-controlled labels.

This work happens after an authorized X write succeeds and before the mitmproxy response hook returns. One valid longform Post can therefore hold the addon event loop for roughly a tenth of a second in this environment, delaying unrelated proxy flows; repeated writes multiply the stall and transient allocation pressure.

## Suggested direction

Give tweet URL refinement a deterministic per-body budget for distinct Unicode-label normalization and retained label-classification state. When the budget is exhausted, keep the conservative `content.create_with_url` bucket instead of continuing the scan or risking underbilling. Do not replace the per-call cache with an unbounded process-global cache.

## Acceptance criteria

- URL billing performs a bounded number of Unicode label-prefix checks and IDNA normalizations for any request body within `REQUEST_BODY_BILLING_INSPECTION_LIMIT`.
- Exhausting that budget conservatively retains `content.create_with_url`; it must not downgrade an ambiguous body or undercharge.
- Existing protocol URLs, bare domains, IDNs, compatibility-mapping cases, obvious non-links, long ASCII candidates, and repeated-label reuse keep their current classifications.
- Add deterministic coverage with thousands of distinct Unicode labels through the production billing boundary. Assert the work bound and conservative category without wall-clock thresholds.
- The implementation does not add an unbounded global cache or increase the 64 KiB body-inspection limit.
- Ruff, BasedPyright, focused X connector usage tests, and the full mitm-addon pytest suite pass.

seven332:
## Summary

- cap one X tweet-body URL scan at 256 distinct labels that require generic IDNA classification
- share request-local normalization and TLD-prefix results across all bare-domain candidates
- keep billing on `content.create_with_url` when the work budget is exhausted
- cover thousands of distinct Unicode labels across individually sub-limit candidates through the connector usage boundary

## Behavior

Pure ASCII candidates retain their normalization-free path, and repeated Unicode labels continue to reuse classification results. Bodies above the new unique-label ceiling fail closed to the existing more expensive billing bucket instead of continuing attacker-controlled synchronous work.

## Out of scope

- no process-global cache
- no request-body inspection limit change
- no API, schema, protocol, persisted-state, or deployment configuration change

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/x_connector_usage/test_writes.py` (126 passed)
- `uv run --no-sync python -m pytest tests/` (7,184 passed)
- `lefthook run pre-commit`

Closes #30037

### DIFF
--- a/crates/runner/mitm-addon/src/usage/providers/connectors/x_billing.py
+++ b/crates/runner/mitm-addon/src/usage/providers/connectors/x_billing.py
@@ -344,6 +344,7 @@ _BARE_DOMAIN_CANDIDATE_RE = re.compile(
     re.IGNORECASE,
 )
 _MAX_DOMAIN_LABEL_CHARS = 63
+_MAX_BILLING_DOMAIN_LABEL_CLASSIFICATIONS = 256
 _DOMAIN_LABEL_RE = re.compile(rf"^[a-z0-9-]{{1,{_MAX_DOMAIN_LABEL_CHARS}}}$")
 _INVALID_TLD_FOLLOWING_CHARS = "@+-"
 
@@ -404,7 +405,12 @@ def _normalize_billing_domain_label(label: str) -> _BillingDomainLabelNormalizat
         return _BillingDomainLabelNormalization(None, False)
 
 
-def _bare_domain_candidate_likely_contains_url(candidate: str, following_char: str) -> bool:
+def _bare_domain_candidate_likely_contains_url(
+    candidate: str,
+    following_char: str,
+    label_normalizations: dict[str, _BillingDomainLabelNormalization],
+    label_tld_prefixes: dict[str, bool],
+) -> bool:
     if candidate.isascii():
         simple_ascii_candidate = candidate.lower()
         # Canonical A-label validation remains owned by the shared IDNA path.
@@ -413,13 +419,20 @@ def _bare_domain_candidate_likely_contains_url(candidate: str, following_char: s
     else:
         use_simple_ascii_labels = False
         labels = candidate.split(".")
-    label_normalizations: dict[str, _BillingDomainLabelNormalization] = {}
-    label_tld_prefixes: dict[str, bool] = {}
     last_label_index = len(labels) - 1
     for index, label in enumerate(labels):
         if use_simple_ascii_labels:
             normalized_label = label
         else:
+            normalization = label_normalizations.get(label)
+            if (
+                normalization is None
+                and len(label_normalizations) >= _MAX_BILLING_DOMAIN_LABEL_CLASSIFICATIONS
+            ):
+                # Exhausted inspection is ambiguous, so keep the conservative
+                # with-URL billing bucket.
+                return True
+
             if index > 0:
                 has_tld_prefix = label_tld_prefixes.get(label)
                 if has_tld_prefix is None:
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #30059

**Reviewed head:** `32ee17bc3d697d076db6f1070cfd09db82cb5c05`

### Summary

The change adds one request-local, body-wide 256-label ceiling around X tweet URL billing classification, shares existing normalization state across candidates, and fails closed to `content.create_with_url` when inspection is exhausted. The implementation matches #30037 and does not change public interfaces, body limits, or deployment contracts.

### Findings

No findings.

### Testing

Local validation passed: uv lock/sync, Ruff format and lint, BasedPyright, 126 focused X write tests, 7,184 full mitm-addon tests, and pre-commit hooks. The regression test uses the production connector usage boundary, splits 10,000 distinct labels across individually sub-limit candidates, pins 256 normalization and 255 prefix-check calls, and asserts conservative billing without a wall-clock threshold.

GitHub CI is evaluated separately by the PR workflow.

### Verdict

LGTM

PR #30061, opened 2026-08-28
test(runner): verify dns reservation socket lifetime

### LINKED ISSUE #30029: test(rust): verify DNS reservation holds TCP and UDP sockets
## Problem

The DNS port reservation's core lifetime contract is not directly tested.

- `crates/runner/src/dns/port.rs:1-6` stores both a bound TCP socket and a bound UDP socket in `DnsPortReservation`.
- `crates/runner/src/dns/port.rs:35-44` documents that `reserve_port` reserves a port for both protocols.
- `crates/runner/src/dns/port.rs:60-65` transfers the successful TCP and UDP sockets into the returned reservation.
- `crates/runner/src/dns/port.rs:86-90` only asserts that the returned port is nonzero.
- `crates/runner/src/dns/port.rs:116-137` verifies retry after an injected UDP `AddrInUse`, but only asserts the selected port number; it never attempts competing TCP or UDP binds while the reservation is alive.

The lifetime matters outside this helper. `crates/runner/src/cmd/start/mod.rs:682-719` keeps the reservation across sandbox runtime and network namespace setup, and `crates/runner/src/dns/proxy.rs:102-115` explicitly releases it only immediately before spawning dnsmasq. No current test would fail if a refactor stopped retaining either socket while still returning a plausible port.

## Impact

If either protocol is no longer held for the full reservation lifetime, another process can claim the selected port during runtime setup. dnsmasq then exits with `AddrInUse`, forcing an expensive runtime rebuild; repeated claims can exhaust the three startup attempts and prevent the runner from starting.

## Acceptance criteria

- Add a deterministic Rust test that obtains a production `DnsPortReservation` and verifies a competing TCP bind to its port fails with `AddrInUse` while the reservation is alive.
- In the same lifecycle test, verify a competing UDP bind to the port also fails with `AddrInUse`.
- Drop the reservation and verify the port can be bound again for both TCP and UDP, proving the hold and release boundaries.
- Assert structured error kinds rather than platform-specific message text.
- Keep the existing UDP-conflict retry regression coverage.

seven332:
## Summary

- replace the DNS reservation's port-number smoke test with a complete socket-lifetime regression
- verify competing TCP and UDP binds report `AddrInUse` while the production reservation is alive
- verify both protocols can bind the same port again after the reservation is dropped

## Scope

This is a test-only change in the runner DNS port module. It does not change production allocation, retry behavior, visibility, dnsmasq handoff, configuration, persistence, protocols, or deployment compatibility.

Closes #30029

## Validation

- `cargo test --manifest-path crates/Cargo.toml -p runner --bin runner dns::port::tests::reservation_holds_tcp_and_udp_until_drop -- --exact`
- focused lifecycle test repeated sequentially 25 times
- `cargo test --manifest-path crates/Cargo.toml -p runner --bin runner dns::port::tests`
- `cargo fmt --all --manifest-path crates/Cargo.toml -- --check`
- `cargo doc --manifest-path crates/Cargo.toml -p runner --no-deps`
- `cargo test --manifest-path crates/Cargo.toml -p runner` (3347 passed, 26 ignored; guest inventory 1 passed)
- `git diff --check`
- repository pre-commit hooks (workspace Cargo fmt, docs, Clippy, file size, commitlint)

### DIFF
--- a/crates/runner/src/dns/port.rs
+++ b/crates/runner/src/dns/port.rs
@@ -84,9 +84,20 @@ mod tests {
     use super::*;
 
     #[test]
-    fn reserve_port_returns_nonzero() {
+    fn reservation_holds_tcp_and_udp_until_drop() {
         let reservation = reserve_port().unwrap();
-        assert!(reservation.port() > 0);
+        let port = reservation.port();
+
+        let tcp_error = std::net::TcpListener::bind(("0.0.0.0", port)).unwrap_err();
+        assert_eq!(tcp_error.kind(), std::io::ErrorKind::AddrInUse);
+
+        let udp_error = std::net::UdpSocket::bind(("0.0.0.0", port)).unwrap_err();
+        assert_eq!(udp_error.kind(), std::io::ErrorKind::AddrInUse);
+
+        drop(reservation);
+
+        let _tcp_listener = std::net::TcpListener::bind(("0.0.0.0", port)).unwrap();
+        let _udp_socket = std::net::UdpSocket::bind(("0.0.0.0", port)).unwrap();
     }
 
     fn bind_tcp_udp_pair() -> std::io::Result<(ReservedTcpPort, std::net::UdpSocket)> {


seven332:
## Code Review: PR #30061

**Reviewed head:** `c5f9838b900bc5b8d640e59f10802fc2d99a4a78`

### Summary

The test-only change replaces a weak port-number assertion with a cohesive real-socket lifecycle regression. It verifies both TCP and UDP ownership before drop and release afterward without changing production visibility or behavior.

### Findings

No findings.

### Testing

The exact lifecycle test, all DNS port tests, 25 sequential focused repetitions, runner formatting, all-target/all-feature Clippy, runner docs, the full runner test suite, diff checks, and repository pre-commit hooks passed locally. Privileged runner E2E was not run because it cannot observe this private hold/release boundary more directly; CI remains the external verification surface.

### Verdict

LGTM

PR #29738, opened 2026-08-27
perf(platform): stabilize lazy chunk loading across deploys

hulh122:
## Summary

- Group long-lived startup dependencies into four coarse, deterministic chunks (`vendor-foundation`, `vendor-auth`, `vendor-content`, and `vendor-services`). The grouping walks only static importers back to `src/main.ts`, so feature-only dynamic modules retain Rolldown's normal lazy boundaries.
- Install an inline `vite:preloadError` handler before the entry-module import. It prevents Vite's default rethrow on every occurrence, prompts once per document using the resolved pre-bundle locale, and reloads only after explicit confirmation; canceling keeps the tab and unsent work intact.
- Add regression guards at the Vite build-output and platform bootstrap-page boundaries for cyclic startup graphs, lazy-only exclusion, localized recovery copy, one-shot recovery, and explicit reload consent.

Refs #29576

## Deployment evidence and safety decision

The current cache policy remains unchanged and correct: `/assets/*` is immutable for one year, while `index.html` and `sw.js` must revalidate. The Service Worker remains Web-Push-only; this PR adds no fetch handler or app-shell precache.

Production does not retain old content-hashed assets at the stable origin:

- The build workflow uploads one commit-addressed `dist.tar.gz` archive to R2.
- Production promotion expands only the selected archive into a fresh Pages directory.
- The Pages worker resolves assets only through the current deployment's `env.ASSETS` binding.
- Live check: the prior deployment's `assets/index-Dl5PejkM.js` returned JavaScript with HTTP 200 from `https://ccc9e46e.okou-app.pages.dev`, but the same path returned HTTP 404 from `https://app.vm0.ai` after the next deployment.

Durable origin retention therefore needs a Cloudflare/R2 asset-read binding plus a retention/lifecycle contract, or a broader stateful release-bundle design. Neither capability exists in the current Pages worker/deployment contract, and merging frequent whole-build archives into every Pages upload would have unbounded size/file-count and rollback costs. This PR uses the repo-contained recovery path instead.

Vite 8.0.16's installed client runtime creates `vite:preloadError` as a cancelable event, attaches the rejected error as `payload`, dispatches it, and rethrows unless `defaultPrevented`. The handler is inline so it covers both failure of the always-requested `src/main.ts` entry module and later lazy imports.

## Build comparison

[...]

PR #27155, opened 2026-08-14
fix: statically import mermaid

e7h4n:
## Summary

- replace the platform-owned runtime `import("mermaid")` with a module-level static import
- remove the cached Mermaid module promise while preserving render queue, theme, and image caching behavior
- load Mermaid core with the app module so first diagram rendering no longer depends on fetching that separate lazy entry after a deployment

## Trade-off

- adds Mermaid core (approximately 170 KB gzip) to the initial app dependency graph
- Mermaid's upstream per-diagram renderers remain internally code-split; this PR removes the platform-owned top-level lazy import

## Verification

- `pnpm format`
- `pnpm turbo run lint --concurrency=1 --log-order grouped`
- `pnpm knip`
- staged workspace, platform, and API TypeScript checks
- `pnpm --filter @okouai/app exec vitest run src/signals/__tests__/mermaid-diagram.test.ts`
- `pnpm --filter @okouai/app run build` (confirmed no `mermaid.core-*` top-level lazy asset)

PR #30064, opened 2026-08-28
perf(platform): lazy-load mermaid core

bingjiez666:
## Summary

- move Mermaid core out of the platform startup import graph
- load and memoize the module when a Mermaid diagram signal first renders
- preserve the existing render queue, theme/image caches, and Mermaid's per-diagram renderer splitting
- update the import regression test to enforce the lazy boundary

## Context

#27155 made Mermaid core static to avoid a deployment-time lazy chunk miss, at the cost of roughly 150 KB gzip in startup for every user. The localized `vite:preloadError` recovery added in #29738 now owns the old-open-client case: if a later deployment removes the requested content-hashed chunk, the page asks the user to confirm a refresh.

This change accepts that existing recovery behavior so users without Mermaid diagrams do not download Mermaid core. The first rendered diagram downloads core plus the diagram-specific renderer it needs.

## Verification

- `pnpm format`
- `pnpm exec prettier --check apps/platform/src/signals/mermaid-diagram.ts apps/platform/src/signals/__tests__/mermaid-diagram-imports.test.ts`
- `pnpm --filter @okouai/app run lint`
- `pnpm knip`
- `pnpm --filter @okouai/app exec vitest run src/signals/__tests__/mermaid-diagram-imports.test.ts`
- `pnpm --filter @okouai/app exec vitest run src/signals/__tests__/mermaid-diagram.test.ts`

## Build note

`pnpm --filter @okouai/app run build` transformed 7,491 modules successfully, then the local container was OOM-killed with exit 137 during final chunk generation. The PR pipeline build remains the authoritative production-bundle check.

### DIFF
--- a/turbo/apps/platform/src/signals/__tests__/mermaid-diagram-imports.test.ts
+++ b/turbo/apps/platform/src/signals/__tests__/mermaid-diagram-imports.test.ts
@@ -15,8 +15,10 @@ if (sourcePath === undefined) {
 const source = readFileSync(sourcePath, "utf8");
 
 describe("mermaid diagram imports", () => {
-  it("keeps mermaid in the static import graph", () => {
-    expect(source).toMatch(/^import mermaid from "mermaid";$/m);
-    expect(source).not.toMatch(/\bimport\s*\(\s*["']mermaid["']\s*\)/);
+  it("loads mermaid through the diagram computed", () => {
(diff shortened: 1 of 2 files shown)


bingjiez666:
Closing because the intended boundary is already on main: keep Mermaid core in the startup graph while Mermaid's per-diagram renderers remain lazy-loaded. This PR would unnecessarily defer the core as well.

PR #30065, opened 2026-08-28
fix(mitm-addon): reject uninspectable billable responses

### LINKED ISSUE #30035: security(python): prevent response encodings from bypassing streamed usage accounting
## Problem and evidence

The runner rewrites `Accept-Encoding` on HTTP flows whose response bodies are inspected for usage. When the requester explicitly rejects `identity` and every stream-decodable compression coding, `normalize_accept_encoding_for_body_inspection()` preserves the original header rather than advertising a representation the requester rejected.

That compatibility fallback becomes a billing bypass when sandbox code advertises only a non-stream-decodable coding such as `br`, the upstream returns a successful compressed SSE response, and the flow is billable. At response-header time, `_configure_response_inspection_stream()` records `response_encoding_not_stream_decodable` but installs neither a usage parser nor a bounded terminal fallback. The pass-through stream remains active, so the sandbox receives the model output while terminal reporting finds no usage events.

The existing integration test in `tests/test_response_encoding_inspection_risk.py` proves the response-time branch and explicitly asserts that the response stream remains callable. `report_model_provider_usage()` independently proves that a billable model response without extracted usage events is not reported.

Issue #21142 and PR #21172 intentionally kept this behavior fail-open and added telemetry without changing traffic. That was a compatibility decision, not an enforcement guarantee. The attacker-controlled header boundary and confirmed no-parser/no-fallback response path now justify changing the policy for billable flows.

## Decision

Required usage inspection is an enforcement boundary for successful billable responses. A protected response must not be delivered when its actual encoding leaves the proxy with neither an incremental usage parser nor an applicable bounded terminal accounting fallback. A structured risk log alone is insufficient.

The actual response `Content-Encoding` is authoritative. Request normalization may reduce exposure, but it cannot be the only enforcement point because an upstream can return an unadvertised or otherwise non-stream-decodable encoding.

## Scope and constraints

- Cover successful billable model-provider responses and successful billable connector responses whose accounting requires a response-body parser.
- Gate on billing requirements, not observation alone. Non-billable/BYOK model-provider observation must not become a traffic-blocking contract.
- Preserve explicit requester encoding rejections; do not silently advertise `identity`, `gzip`, or `deflate` after the requester assigned them `q=0`.
- Preserve stream-decodable responses and existing bounded terminal fallbacks when they can still produce accounting.
- Do not block bodyless responses or upstream failure responses that cannot deliver a billable protected result.
- Treat response replacement as fail-closed: no upstream body bytes may reach the sandbox after the proxy determines that required inspection is unavailable.
- Keep diagnostic fields bounded and fixed; do not persist raw encoding header values.
- No API, database, queue, persisted-state, or cross-process protocol change is required.

## Alternatives considered

- **Keep logging only:** preserves compatibility but leaves the demonstrated billing bypass intact.
- **Force a safe request encoding:** violates explicit HTTP content-negotiation rejections and still does not cover a non-conforming upstream response.
- **Reject only unsafe request negotiation:** can prevent the known header path before incurring upstream work, but is not a complete enforcement boundary because the actual response remains authoritative and some request paths may still have a valid bounded fallback.
- **Add bounded Brotli and zstd streaming decoders:** could reduce future rejection, but current bindings do not provide the required hard output bound and unknown or compound encodings would still need fail-closed handling.
- **Buffer every unsupported response:** delays streaming, increases memory retention, and cannot guarantee bounded accounting for SSE or arbitrarily large bodies.

## Risks and mitigations

- Clients that intentionally permit only unsupported streaming encodings will receive a deterministic proxy failure instead of the upstream result. Limit this behavior to successful billable flows with no usable accounting path.
- A response-time rejection occurs after the provider may already have processed the request. Discarding the body protects the delivery/accounting invariant; planning should also evaluate whether an early request rejection is safe for cases that have no possible inspected representation.
- Over-broad classification could reject a response that still has a bounded terminal fallback. Reuse the parser/fallback capability decision rather than maintaining a second encoding allowlist.
- Replacement streaming must discard all upstream chunks and preserve terminal tracking cleanup so retries, shutdown, and usage flushing do not leak flow state.

## Acceptance criteria

- A billable Anthropic SSE flow using `Accept-Encoding: br, identity;q=0` and receiving `Content-Encoding: br` does not deliver upstream response bytes to the sandbox.
- A successful billable response with a non-stream-decodable actual encoding also fails closed when the upstream ignored or bypassed apparently safe request negotiation.
- Fail-closed handling returns a deterministic local error, discards the upstream body, and leaves terminal usage/tracking state consistent.
- Successful gzip, deflate, and identity model streams retain incremental usage extraction and existing response behavior.
- Existing bounded terminal JSON fallbacks remain available when applicable; the enforcement path is selected only when neither incremental nor bounded terminal accounting is possible.
- Non-billable observable model-provider flows, bodyless responses, unsuccessful upstream responses, and connectors without response-body billing parsers keep their current behavior.
- Diagnostics continue to use stable internal reason fields without raw encoding values.
- Integration coverage exercises request normalization, the real `responseheaders()` streaming gate, upstream chunk handling, response completion, and terminal usage reporting.

## Delivery notes

This is one independently reviewable runner-local PR. The compatibility change is limited to response handling inside a single runner version and requires no coordinated deployment or persisted-state migration.

seven332:
## Summary

- Fail closed with an empty local 502 when a successful billable model or connector response cannot be inspected by either an incremental parser or an applicable bounded terminal fallback.
- Discard all upstream response bytes after rejection and release model-provider failure state so the proxy-generated response is not reported as a provider outage.
- Keep bounded JSON fallback for ordinary X responses while excluding X NDJSON stream endpoints that require incremental parser state.

## Explicit exclusions

- Non-billable observable model responses, non-billable connectors, bodyless responses, non-2xx responses, and connectors without response-body accounting parsers retain their existing behavior.
- This does not add a request-time rejection gate; the actual response encoding remains authoritative.
- No API, persisted-state, queue, database, or cross-version protocol changes are included.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (7,189 passed)
- `lefthook run pre-commit`

Closes #30035

### DIFF
--- a/crates/runner/mitm-addon/src/response_streaming.py
+++ b/crates/runner/mitm-addon/src/response_streaming.py
@@ -38,6 +38,7 @@ from usage.underbilling import log_usage_underbilling
 _HTTP_STATUS_SWITCHING_PROTOCOLS = 101
 _HTTP_STATUS_OK_MIN = 200
 _HTTP_STATUS_REDIRECT_MIN = 300
+_HTTP_STATUS_BAD_GATEWAY = 502
 
 _MODEL_JSON_USAGE_FINISH = "model_json_usage_finish"
 _MODEL_SSE_USAGE_FINISH = "model_sse_usage_finish"
@@ -74,6 +75,7 @@ class _ResponseStreamSetup(NamedTuple):
     parser: _ResponseChunkParser | None
     needs_buffered_fallback: bool
     finish_stream: Callable[[], object] | None = None
+    reject_uninspectable: bool = False
 
 
 def model_usage_protocol(flow: http.HTTPFlow) -> usage.ModelUsageProtocol:
@@ -268,7 +270,15 @@ def _configure_response_inspection_stream(
                         flow,
                         failure_observer.finish,
                     )
-                return _ResponseStreamSetup(None, False)
+                return _ResponseStreamSetup(
+                    None,
+                    False,
+                    reject_uninspectable=(
+                        is_billable_flow
+                        and is_observable_model_provider
+                        and _HTTP_STATUS_OK_MIN <= response.status_code < _HTTP_STATUS_REDIRECT_MIN
+                    ),
+                )
 
             finished = False
 
@@ -341,11 +351,20 @@ def _configure_response_inspection_stream(
                     flow,
                     failure_observer.finish,
                 )
-            return _ResponseStreamSetup(
-                None,
+            needs_buffered_fallback = (
                 is_observable_model_provider
                 and uses_model_json_fallback(flow)
-                and body_decoding.can_decode_json_usage_body(response.headers),
+                and body_decoding.can_decode_json_usage_body(response.headers)
+            )
+            return _ResponseStreamSetup(
+                None,
+                needs_buffered_fallback,
+                reject_uninspectable=(
+                    is_billable_flow
+                    and is_observable_model_provider
+                    and _HTTP_STATUS_OK_MIN <= response.status_code < _HTTP_STATUS_REDIRECT_MIN
+                    and not needs_buffered_fallback
+                ),
             )
 
         inspection: usage.ModelJsonResponseInspection | None = None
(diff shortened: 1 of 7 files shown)


seven332:
## Code Review: PR #30065

**Reviewed head:** `324529b4a6c7c4ed79d7c0d27413fa5327e1b55c`

### Summary

The response-time gate now enforces the billable accounting boundary using the actual response encoding and existing parser/fallback capabilities. The local 502 replacement discards upstream bytes, releases model-provider failure state, and preserves the intended non-billable, non-2xx, bodyless, and bounded-JSON fallback behavior.

### Findings

No findings.

### Testing

Ruff format/lint, basedpyright, root pre-commit hooks, focused lifecycle suites, and the full mitm-addon suite passed (7,189 tests). GitHub CI is inspected separately after this review.

### Verdict

LGTM

PR #30067, opened 2026-08-28
perf(guest-agent): reuse metrics log handle

### LINKED ISSUE #30032: perf(rust): reuse the metrics log handle across collection ticks
## Problem and evidence

The guest metrics collector opens and closes the same private `metrics.jsonl` file for every five-second sample. Each small JSONL record therefore repeats the secure traversal and validation of the complete runtime path.

- `crates/guest-agent/src/constants.rs:13-14` sets `METRICS_INTERVAL_SECS` to 5, so a one-hour run emits 720 samples.
- `crates/guest-agent/src/metrics.rs:169-174` calls `open_private_append` inside `write_metrics_entry`; the local `File` is dropped when each call returns.
- `crates/guest-contracts/src/runtime_paths.rs:517-622` opens and validates every parent component and tightens the final parent permissions on every open.
- `crates/guest-contracts/src/runtime_paths.rs:766-829` then opens, validates, and tightens the final file before returning it.
- For the normal `/home/user/.vm0/guest-agent/runs/<run-id>/logs/metrics.jsonl` layout, one sample performs at least 19 open/stat/chmod syscalls plus the corresponding descriptor closes. At 720 samples per hour, this is more than 20,000 repeated filesystem syscalls per guest-hour solely to reopen the metrics sink.
- The established system-log and sandbox-operations sinks already cache their securely opened `File` handles across writes (`crates/guest-common/src/log.rs:26-68` and `crates/guest-common/src/telemetry.rs:26-53`).

## Impact

Every metrics payload is only one JSON line, but its synchronous write currently repeats secure traversal and permission checks for the entire runtime path. This adds avoidable CPU and kernel/filesystem work on a Tokio worker every five seconds and scales with the number and duration of active guests.

## Decision

Proceed to planning a retained metrics sink. The problem is current, measurable from the implementation, and isolated to one writer lifecycle. Keeping the status quo preserves avoidable work, while weakening `open_private_append` would trade away security checks that are still required whenever a handle is opened.

## Constraints

- Initial opens and retries must continue to use `open_private_append`, preserving the current no-follow, regular-file, and private-permission checks.
- Metrics collection remains best-effort. A sink failure must not terminate the metrics loop or add repetitive failure logging.
- The JSONL payload format and five-second collection cadence remain unchanged.

## Acceptance criteria

- The metrics loop lazily opens the private append file and reuses the handle across collection ticks.
- A write failure drops the cached handle so a later tick can securely reopen it.
- Tests cover multiple JSONL records through the retained sink and reopening after an I/O failure.
- Existing metrics serialization and collection behavior remain unchanged.

seven332:
## Summary

- retain one lazily opened private metrics append handle across collection ticks
- drop the cached handle after a write failure so a later tick securely reopens the configured path
- add public-loop integration coverage for retained-inode writes and cached-write failure recovery

## Scope

- keeps metrics serialization, JSONL framing, five-second cadence, and best-effort behavior unchanged
- adds no logging, buffering, shared synchronization, public API, protocol, schema, path, or dependency changes
- replaces the implementation-coupled write helper test with real-filesystem integration coverage

## Test plan

- `cargo test -p guest-agent --test integration metrics`
- `cargo clippy -p guest-agent --all-targets -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p guest-agent --no-deps`
- `cargo test -p guest-agent`
- pre-commit Rust fmt, rustdoc, Clippy, file-size, and commitlint hooks

Closes #30032

### DIFF
--- a/crates/guest-agent/src/metrics.rs
+++ b/crates/guest-agent/src/metrics.rs
@@ -5,6 +5,7 @@
 
 use crate::constants;
 use serde::Serialize;
+use std::fs::File;
 use std::io::Write;
 use std::time::Duration;
 use tokio_util::sync::CancellationToken;
@@ -17,6 +18,43 @@ struct MetricsEntry {
     mem_total: u64,
     disk_used: u64,
     disk_total: u64,
+}
+
+/// Loop-owned metrics destination with a lazily opened append handle.
+///
+/// The retained descriptor continues targeting the opened inode across path
+/// replacement. A write failure drops it so the next tick securely reopens the
+/// configured path.
+struct MetricsSink {
+    path: String,
+    file: Option<File>,
+}
+
+impl MetricsSink {
+    fn new(path: String) -> Self {
+        Self { path, file: None }
+    }
+
+    fn append(&mut self, entry: &MetricsEntry) {
+        let Ok(json) = serde_json::to_string(entry) else {
+            return;
+        };
+
+        let file = match self.file.as_mut() {
+            Some(file) => file,
+            None => {
+                let Ok(file) = guest_contracts::runtime_paths::open_private_append(&self.path)
+                else {
+                    return;
+                };
+                self.file.insert(file)
+            }
+        };
+
+        if writeln!(file, "{json}").is_err() {
+            self.file = None;
+        }
+    }
 }
 
 /// Tracks previous `/proc/stat` counters for delta-based CPU measurement.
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #30067

**Reviewed head:** `50f636b5480f74f3f97c2ac033ba58c646a268f9`

### Summary

The metrics loop now retains one lazily opened private append handle across ticks and drops it after a write failure so the next tick can securely reopen the configured path. The change preserves payload, cadence, best-effort behavior, and private-file validation without adding logging or synchronization.

### Findings

No findings.

### Testing

Verified the focused metrics integration tests, Rust formatting, guest-agent Clippy, rustdoc, and the full guest-agent test suite. The integration coverage exercises retained-inode writes and recovery after a cached-handle write failure with real files; no broader cross-crate suite was run because no shared contract or public interface changed.

### Verdict

LGTM

PR #30088, opened 2026-08-28
test(api): make cron cleanup fixture teardown-safe

### LINKED ISSUE #30078: test(api): make cron cleanup regression teardown-safe
## Context

The required `test-api (4)` check for Official Workflow P4 PR #29996 failed twice at exact head `65a773f894d5b308bd8a8a7ffe06b4559cbb543b` in:

- Turbo run `33143544603`, attempt 1, job `98759577153`
- Turbo run `33143544603`, attempt 2, job `98761798322`

The failing test is:

`src/signals/routes/__tests__/cron-delete-cleanups.test.ts > complete delete cleanup crons > caps connector cleanup at ten batches and preserves the UTC cutoff outside UTC`

Attempt 2 provides the causal evidence. The 15-second test deadline expired while the awaited `cleanup-connector` route still owned its batch delete. Vitest then ran the registered fixture cleanup, which issued a marker-wide delete against the same rows. PostgreSQL reported `40P01`:

- process 570: production-shaped batched delete by selected IDs
- process 636: fixture teardown delete by `(user_id, org_id)`
- each process waited on the other transaction

The timeout is therefore followed by a teardown race; the aggregate `ci-gate-turbo` failure and fail-fast cancellations are downstream noise.

A concurrent current-main control did not fail: Turbo run `33144351570`, head `59c6902e5185016fd49b51bc6609ea372380e501`, job `98762070433` completed the same case in 2.644 seconds. Main and P4 Official Workflow suites had similar wall time (24.773s vs 24.254s), so the evidence does not establish that P4 product code caused the latency. It does establish an unsafe test ownership boundary under a slow valid interleaving.

## Requirements

- Identify the first causal ownership/isolation defect in the cron cleanup test and fixture.
- Ensure teardown cannot delete the fixture rows while the awaited cleanup operation still owns work on those rows.
- Preserve the meaningful assertions for the ten-batch cap and strict UTC cutoff.
- Keep fixture data scoped to its unique marker and avoid global/shared-state assertions.
- Do not solve this with a larger timeout, sleeps, assertion retries, skipped coverage, or manual global cleanup.
- Do not change production cleanup semantics unless evidence proves a product race.

## Validation

- Reproduce the original timing/teardown interleaving or otherwise demonstrate the causal boundary before the fix.
- Run the exact targeted test file at least five times after the fix.
- Run related connector cleanup service/route regression coverage.
- Run affected format, lint, and type checks.
- Open a focused PR, complete exact-head CI and self-review, obtain independent review, and merge only through the protected merge queue.

## Relationship

This is a repository-protection blocker for #29996 under EPIC #28905. It is intentionally separate from P4 product code. First-merge-wins still applies; no unrelated PR ordering is reserved.

lancy:
## Summary

- make both connector cleanup fixtures own every request that can touch their rows and drain those operations before marker-scoped teardown
- exercise the shared ten-iteration cap with a one-row test-only batch, reducing the cap fixture from 10,001 expired rows to 11
- keep the production cleanup batch size at 1,000 and preserve the UTC cutoff and unrelated-marker assertions

## Root cause

When the test deadline fired, Vitest could start onTestFinished while the awaited cleanup route still owned a non-cancellable PostgreSQL batch delete. The independently registered marker-wide teardown then deleted the same rows and created the observed 40P01 cycle. The fixture owner now closes to new work and settles every already-owned operation before deleting its unique marker. The same barrier covers the only related caller of the marker-scoped cleanup command.

## Validation

- final exact regression matrix: 5/5 (test bodies 173-233 ms)
- related cron-connector-oauth-state-cleanup.test.ts: pass (243 ms)
- pnpm format and focused Prettier verification
- full monorepo lint, then final API lint after the shared helper change
- pnpm knip
- non-API/non-platform, platform, and API type-check gates; final API type gate after the shared helper change
- git diff --check

Closes #30078

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/cron-connector-oauth-state-cleanup.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/cron-connector-oauth-state-cleanup.test.ts
@@ -1,19 +1,13 @@
 import { randomUUID } from "node:crypto";
 
 import { connectorOauthStartContract } from "@okouai/api-contracts/contracts/connectors";
-import {
-  afterEach,
-  beforeEach,
-  describe,
-  expect,
-  it,
-  onTestFinished,
-} from "vitest";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
 
 import { accept, testContext } from "../../../__tests__/test-context";
 import { setupApp } from "../../../__tests__/test-helpers";
 import { mockEnv, mockOptionalEnv } from "../../../lib/env";
 import { clearMockNow, mockNow, now } from "../../../lib/time";
+import { createFixtureOperationOwner } from "./helpers/fixture-operation-owner";
 import { createRouteMocks } from "./helpers/route-test";
 import {
   type TestCronDeleteCleanupsStateActionBody,
@@ -64,10 +58,27 @@ async function requestFixture(
   return response.body;
 }
 
-function registerFixtureCleanup(marker: string): void {
-  onTestFinished(async () => {
+function createConnectorFixture(marker: string) {
+  const owner = createFixtureOperationOwner(async () => {
     await requestFixture({ action: "delete-connector", marker });
   });
+  return {
+    cleanup: async () => {
+      return await owner.run(async () => {
+        return await requestFixture({ action: "cleanup-connector", marker });
+      });
+    },
+    read: async () => {
+      return await owner.run(async () => {
+        return await requestFixture({ action: "read-connector", marker });
+      });
+    },
+    startOauth: async () => {
+      return await owner.run(async () => {
+        return await startGithubOauth(marker);
+      });
+    },
+  };
 }
 
 describe("connector OAuth state cleanup cron", () => {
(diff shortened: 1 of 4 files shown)


lancy (commented):
LGTM

### Summary

This fixes the cron cleanup regression by giving both marker-scoped connector fixtures explicit ownership of every route operation, closing the owner before teardown, and draining all already-started operations before deleting fixture rows. The production cleanup path remains at 1,000 rows per batch for at most ten batches; only the marker-scoped test command uses one row per batch so 11 rows prove the ten-iteration cap without the prior 10,001-row timeout exposure.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness, races, and edge cases | Pass | `createFixtureOperationOwner` registers the promise before the operation starts, rejects work after teardown begins, drains with `Promise.allSettled`, then deletes the unique marker; the cap and equal/future cutoff assertions remain exact. |
| `REVIEW.md` | Security, type safety, and maintainability | Pass | No auth surface, persisted contract, unsafe cast, suppression, or public interface changes; the lifecycle barrier is a focused test helper shared by the only two callers. |
| `REVIEW.md` | Fix regression coverage and fallback discipline | Pass | The exact regression ran 5/5 locally and passed in `test-api (4)`; no fallback or compatibility branch is introduced. |
| `docs/bad-smell.md` | No `any`, suppressions, dynamic imports, hidden recovery, or hardcoded environment behavior | Pass | The four-file diff contains none; rejection after teardown starts is explicit and fail-fast. |
| `docs/testing.md` | Integration-first testing with real internals and owned async cleanup | Pass | Tests use the real database and route harness; every fixture operation is awaited and teardown drains in-flight DB work without sleeps, retries, fake timers, or timeout growth. |
| `.claude/skills/ccstate/SKILL.md` | Abort and promise lifecycles must have explicit owners; do not paper over floating work | Pass | Route operations are registered before execution and awaited before fixture deletion; production commands keep the existing signal path and no `detach` or empty catch was added. |
| `docs/testing/patterns.md` | API route integration through `setupApp`, contracts, real internal code, and unique fixture identity | Pass | Both tests call the contract-backed route slices, use UUID markers, and assert returned behavior rather than mocks. |
| `docs/testing/anti-patterns.md` | Avoid internal mocks, fake timers, artificial waits, implementation-only assertions, and direct service tests | Pass | None are introduced; the regression stays route-level and removes the enlarged test timeout. |
| `docs/testing/api-testing.md` | Shared persistent state must be uniquely owned; cleanup follows correctness and global crons use an owned test route | Pass | Each connector fixture has its own marker, all operations are owner-scoped, and the marker-scoped test route preserves the production-global route unchanged. |
| `docs/testing/testing-external-behavior.md` | Exercise and observe API behavior; test-only exceptions must represent otherwise unconstructible infrastructure state | Pass | OAuth setup and cleanup/read verification use API routes; the test-only marker route is the narrow isolation seam for a production-global cron. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- Coverage: Adequate. Final local exact-case matrix 5/5 at 173–233 ms, related route regression 243 ms, and exact-head Turbo run 33147166097 passed all eight API shards including `test-api (4)`.
- Conventions: Compliant. Full lint, Knip, format, type boundaries, Security, CodeQL, Crates, Runner Image, preview, and E2E tails are terminal green: 77 success, 19 intentional skips, zero failure or pending.
- Integration: Live `main` at `7163952bc5aeefe01a03449925c896fb8968dd88` produces clean synthetic merge tree `0f66685f8c3df3ce9a0fef47f068c4d0a3dcc964`; the earlier overlapping web-URL rename is preserved.

Exact head `36a51d50c498e0851fab09a59682ad42471a1480` is cleared for the normal protected merge queue only. Any head change requires fresh review.

PR #30090, opened 2026-08-28
perf(vsock-host): move exec output copy outside state lock

### LINKED ISSUE #30069: perf(rust): move exec output payload copies outside the connection lock
## Summary

`dispatch_output` copies every streamed exec-output chunk while holding the connection-wide `Shared.state` mutex. The function releases the mutex after reserving channel capacity, but reacquires it before calling `ExecStreamPermit::send`; that call performs the actual `Vec` allocation and byte copy.

This defeats the documented unlocked payload-ownership boundary and unnecessarily blocks all other connection-state operations during every streamed chunk copy.

## Evidence

- `crates/vsock-host/src/exec_operation/dispatch.rs:183-217` validates the borrowed output and reserves an owned channel permit under `Shared.state`, then releases the lock. The comment at lines 196-197 says payload ownership should happen outside the lock.
- `crates/vsock-host/src/exec_operation/dispatch.rs:220-240` runs the `exec_output_before_copy` hook while unlocked, then reacquires `Shared.state` and calls `permit.send(decoded)` at line 237 while the guard is still live.
- `crates/vsock-host/src/exec_operation/state.rs:474-490` shows that `ExecStreamPermit::send` performs `output.chunk.to_vec()` for both stream variants. The allocation and full payload copy therefore occur inside the global state critical section.
- `crates/vsock-host/src/connection/mod.rs:31-71` shows that this mutex is the connection-wide source of truth for liveness, route reservations, pending requests, and active exec operations, so unrelated registration, teardown, and routing work must wait for the copy.
- `crates/vsock-host/src/connection/mod.rs:80-82` explicitly describes the hook as being at the "unlocked exec-output payload-copy boundary," and `crates/vsock-host/src/tests/exec_operation/stream.rs:536-588` verifies that teardown can acquire the state mutex at that boundary.
- `crates/vsock-host/src/file/copy.rs:26-30,510-532` uses this stream for guest-to-host file copies with 64 KiB chunks and a maximum of 128 MiB. A maximum-sized built-in copy can therefore perform about 2,048 allocations and byte copies while repeatedly holding the connection-wide mutex. Other protocol callers may request larger non-zero chunk limits, subject to the 16 MiB wire message limit in `crates/vsock-proto/src/wire.rs:4-5`.

## Performance impact

High-volume exec streams repeatedly extend the `Shared.state` critical section with allocation and byte-copy work that does not require connection-state access. That increases lock contention and latency for connection lifecycle, request routing, operation registration, and teardown. The cost scales with both stream volume and chunk size, and is on the reader task's hot path.

## Acceptance criteria

- Materialize the owned exec-output event/chunk before reacquiring `Shared.state`.
- Keep the second critical section limited to validating that the reserved permit still belongs to the current operation and atomically deciding whether to deliver or discard the already-owned event.
- Do not allocate or copy chunk bytes while holding `Shared.state`.
- Preserve the teardown and sequence-reuse safety properties covered by `exec_output_teardown_before_copy_discards_reserved_event`.
- Add or adapt a regression test that proves the state mutex remains acquirable during the payload-ownership step.

seven332:
## Summary

- materialize streamed exec output before reacquiring the connection-wide state lock
- make `ExecStreamPermit::send` accept owned output and move bytes into both queue variants without copying
- bind the teardown regression hook to the actual payload-copy boundary

## Scope

- no wire protocol, public API, queue capacity, or error behavior changes
- no new logs

## Validation

- `cargo test -p vsock-host exec_output_teardown_at_copy_boundary_discards_reserved_event`
- `cargo clippy -p vsock-host --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p vsock-host --no-deps`
- `cargo test -p vsock-host` (387 passed)
- `lefthook run pre-commit`

Closes #30069

### DIFF
--- a/crates/vsock-host/src/exec_operation/dispatch.rs
+++ b/crates/vsock-host/src/exec_operation/dispatch.rs
@@ -15,7 +15,7 @@ use super::state::{
 };
 use super::types::{
     ExecControlAck, ExecControlGuestStatus, ExecControlOutcome, ExecOperationResult,
-    ExecOwnedCapturedOutput, SupervisedExecStartTiming,
+    ExecOutputEvent, ExecOwnedCapturedOutput, SupervisedExecStartTiming,
 };
 use super::{exec_operation_guest_error, exec_operation_protocol_error};
 
@@ -119,6 +119,19 @@ fn owned_captured_output(output: ExecCapturedOutput<'_>) -> ExecOwnedCapturedOut
     }
 }
 
+fn owned_output_event(
+    output: vsock_proto::DecodedExecOutput<'_>,
+    before_copy: impl FnOnce(),
+) -> ExecOutputEvent {
+    before_copy();
+    ExecOutputEvent {
+        stream: output.stream,
+        output_seq: output.output_seq,
+        chunk: output.chunk.to_vec(),
+        truncated: output.truncated,
+    }
+}
+
 #[cfg(test)]
 fn run_exec_output_before_copy_hook(shared: &Arc<Shared>) {
     let hook = shared
@@ -218,8 +231,10 @@ fn dispatch_output(shared: &Arc<Shared>, msg: BorrowedRawMessage<'_>) -> io::Res
     drop(senders_to_drop);
 
     let returned_sender = if let Some((permit, decoded)) = prepared_output {
-        #[cfg(test)]
-        run_exec_output_before_copy_hook(shared);
+        let event = owned_output_event(decoded, || {
+            #[cfg(test)]
+            run_exec_output_before_copy_hook(shared);
+        });
 
         {
             let mut guard = shared.state.lock().unwrap_or_else(|e| e.into_inner());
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #30090

**Reviewed head:** `e1b5ce42100422aa5fd9fefeb7d39579516a6d51`

### Summary

The change restores exec-output payload ownership before the second connection-state lock, makes permit delivery accept owned data only, and preserves the atomic channel-identity/teardown boundary for both stream variants.

### Findings

No findings.

### Testing

The focused copy-boundary regression, Rust formatting, crate-scoped Clippy, rustdoc, the full 387-test `vsock-host` suite, and pre-commit hooks passed. Existing integration paths cover both event and process-output payloads. No benchmark was run; the allocation and mutex scopes are directly verifiable in the submitted diff.

### Verdict

LGTM

PR #30111, opened 2026-08-28
fix(runner): handle procfs exit race in mitmdump cleanup

### LINKED ISSUE #30086: bug(runner): stabilize mitmdump cleanup proc recheck under concurrency
## Background

The runner test suite intermittently fails in the mitmdump restart lifecycle test when it runs with the full suite, while the same test passes when run by itself.

## Observed behavior

Running:

```bash
cd crates
cargo test -p runner --bin runner --no-fail-fast
```

failed twice in the same test with two nearby call sites:

```text
proxy::process::tests::restart_kills_forked_descendant_and_replaces_launch_directory

runner/src/proxy/process.rs:2131:32:
called `Result::unwrap()` on an `Err` value: Internal("recheck /proc/<pid>/environ after opening mitmdump cleanup pidfd: Permission denied (os error 13)")

runner/src/proxy/process.rs:2120:51:
called `Result::unwrap()` on an `Err` value: Internal("recheck /proc/<pid>/environ after opening mitmdump cleanup pidfd: Permission denied (os error 13)")
```

Both immediate exact-test reruns passed:

```bash
cargo test -p runner --bin runner \
  proxy::process::tests::restart_kills_forked_descendant_and_replaces_launch_directory \
  -- --exact
```

The failures were observed on the repository development container on aarch64 Linux while validating unrelated runner GC changes.

## Expected behavior

The mitmdump cleanup identity recheck and its integration test should be reliable under normal full-suite concurrency without weakening the fail-closed process-ownership boundary.

## Acceptance criteria

- Determine why `/proc/<pid>/environ` becomes unreadable after the cleanup pidfd is opened during the full-suite run.
- Make the production cleanup path or test orchestration handle the identified race safely.
- Preserve fail-closed behavior: an unverified or replaced process must never be killed as if it were the owned mitmdump descendant.
- Add or adjust an entry-point test that reproduces the concurrent failure mode.
- `cargo test -p runner --bin runner --no-fail-fast` passes reliably across repeated runs.

## Relevant code

- `crates/runner/src/proxy/process.rs`

seven332:
## Summary

- classify a failed post-pidfd `environ` read with a fresh process-state and start-time check
- treat only missing, terminal, or replaced processes as no longer matching while preserving errors for the same live generation
- add deterministic launch-close coverage for both the zombie teardown race and the live-process fail-closed case

## Scope

This change is confined to runner procfs cleanup. It does not change process-group shutdown timing, marker validation, raw-PID signaling policy, APIs, persisted state, queue payloads, or deployment protocols.

## Validation

- `cargo test -p runner --bin runner proxy::runtime::tests:: -- --nocapture`
- `cargo test -p runner --bin runner proxy::process::tests::restart_kills_forked_descendant_and_replaces_launch_directory -- --exact --nocapture`
- `cargo test -p runner --bin runner --no-fail-fast` (three consecutive passes on the production implementation plus one current-head pass after the test-seam refinement; each reported 3,351 passed and 26 ignored)
- repository pre-commit Rust formatting, workspace documentation, workspace Clippy, and file-size hooks

Closes #30086

### DIFF
--- a/crates/runner/src/proxy/runtime.rs
+++ b/crates/runner/src/proxy/runtime.rs
@@ -80,11 +80,19 @@ struct ProcessSnapshot {
     same_uid: usize,
 }
 
+#[cfg(test)]
+struct TestEnvironmentRead {
+    pid: u32,
+    read: Box<dyn FnOnce() -> std::io::Result<Vec<u8>> + Send>,
+}
+
 /// Serializes owners of the runner-local proxy resources and scopes every
 /// PyInstaller extraction to one private launch directory.
 pub(super) struct MitmdumpRuntime {
     root: PathBuf,
     _lock: Flock<File>,
+    #[cfg(test)]
+    test_environment_read: std::sync::Mutex<Option<TestEnvironmentRead>>,
 }
 
 impl MitmdumpRuntime {
@@ -107,6 +115,8 @@ impl MitmdumpRuntime {
         let runtime = Arc::new(Self {
             root,
             _lock: runtime_lock,
+            #[cfg(test)]
+            test_environment_read: std::sync::Mutex::new(None),
         });
         runtime.reconcile().await?;
         Ok(runtime)
@@ -243,7 +253,7 @@ impl MitmdumpRuntime {
             } else {
                 consecutive_empty_scans = 0;
                 for process in &snapshot.processes {
-                    signal_stable_process(&self.root, exact_path, process.identity).await?;
+                    signal_stable_process(self, exact_path, process.identity).await?;
                 }
                 wait_for_processes_exit(target, &snapshot).await?;
             }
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #30111

**Reviewed head:** `ac2d3c2e43148f0ee871193855afecfecbbbc89c`

### Summary

The post-pidfd cleanup path now classifies an unexpected `environ` error with a fresh stat: only a missing, terminal, or replaced target is treated as no longer matching, while the same live generation preserves the original error. Pidfd-only signaling, marker validation, and final generation checks remain intact. The test seam is error-only, one-shot, PID-keyed, runtime-scoped, and absent from production builds.

### Findings

No findings.

This does not introduce a compatibility or speculative fallback under `docs/fallback.md`; it handles a reproduced, currently reachable process-teardown state and remains fail-closed for a live uninspectable target.

### Testing

Focused runtime tests, the exact restart lifecycle test, runner all-target/all-feature Clippy, rustdoc with warnings denied, repository pre-commit checks, and four full runner-suite runs passed across implementation and self-review. The final head full-suite result was 3,351 passed and 26 ignored.

Codecov reports 86.36% patch coverage. The uncovered additions are conservative stat-unreadable/invalid branches and test-seam control paths; both externally observable decisions are covered through `close_launch_path`, and covering the remaining internal branches would require violating the repository's stat-reader mock boundary.

### Verdict

LGTM

PR #30132, opened 2026-08-28
docs(python): clarify model provider reporter return semantics

### LINKED ISSUE #30116: docs(python): clarify model provider reporter return semantics
## Problem

The public docstrings for `report_model_provider_usage` and `report_model_provider_usage_observation` describe their boolean return as whether usage was accepted into the reporting path (`crates/runner/mitm-addon/src/usage/providers/model_provider.py:153-170` and `:200-222`). That wording implies that `True` means at least one event or observation from the current call was admitted by the process-local buffer.

The implementation has a different contract. `_buffer_model_provider_usage_events` and `_buffer_model_provider_usage_observations` return `True` whenever the reporting context is complete after invoking the buffer, while discarding the buffer methods' accepted-count return values (`model_provider.py:434-476`). Existing source-deduplication coverage calls the reporters repeatedly and confirms that duplicate source events are not admitted again (`crates/runner/mitm-addon/tests/test_model_provider_usage.py:512-578`), but a duplicate-only call can still receive `True`. In that case, the caller-provided `accepted_source_keys` set remains empty.

## Impact

Contributors can incorrectly use the boolean as evidence that this invocation added new usage, which can produce misleading logging, metrics, or lifecycle decisions around duplicate terminal hooks. The distinction is especially important because `terminal_usage.report_model_provider_usage_once` uses the boolean as a reporting-path signal while the accepted-key collectors carry the actual per-call admission result.

## Decision

Proceed with a focused documentation-only change that describes the existing runtime contract. The boolean remains a reporting-path signal; it will not be redefined as a per-call admission result, and no new structured result API will be introduced.

## Chosen Direction and Rationale

- `report_model_provider_usage` returns `False` when any of its existing gates fail: an empty run ID, a non-model-provider firewall, a non-billable firewall, no billable event built from the available usage sources, or an incomplete sandbox-token/API-URL reporting context. It returns `True` after a non-empty candidate event list passes those gates and the buffer is invoked with a complete context.
- `report_model_provider_usage_observation` returns `False` when its existing gates fail: an empty run ID, a non-observable model-provider flow, no observation built from the available usage sources, or an incomplete reporting context. It returns `True` after a non-empty candidate observation list passes those gates and the buffer is invoked with a complete context.
- For both functions, `True` does not guarantee that the current call admitted a new source event or observation. Process-local source-key deduplication can admit zero candidates after the reporter has built its list.
- When supplied, `accepted_source_keys` receives only payload keys newly admitted by the buffer during the call and is the per-call source of truth for source admission. This preserves the separate signals already consumed by `terminal_usage.report_model_provider_usage_once`.

This direction satisfies the issue without changing duplicate-hook behavior, buffer admission, terminal lifecycle state, or webhook delivery semantics.

## Alternatives Considered

- Redefining the boolean to mean that at least one source key was admitted would change duplicate-only terminal-hook behavior and require coordinated changes to the terminal consumer and regression coverage. The existing consumer already distinguishes reporting-path status from per-call admission through the boolean and accepted-key sets.
- Adding a structured result for reporting-path status and admission details would introduce a new API and compatibility surface for information already available through the existing optional set. No additional caller requirement justifies that scope.

## Risks and Mitigations

- The main risk is future readers treating the boolean as delivery completion or an admission count. The docstrings will explicitly name the gates, state that the buffer's admission result is separate, and identify `accepted_source_keys` as the per-call source of truth.
- There is no runtime, persistence, deployment, migration, security, or compatibility risk from this documentation-only direction.

## Acceptance Criteria

- Document the exact `True` and `False` conditions for both reporter functions.
- Explicitly state that `True` does not guarantee that the current call admitted a new source event or observation because process-local source-key deduplication may accept zero items.
- Document `accepted_source_keys` as the per-call source of truth for which payload keys were newly admitted.
- Keep the documented contract aligned with the way `terminal_usage.report_model_provider_usage_once` consumes these values.

## Delivery Notes

- This is one independently reviewable PR changing only the two public docstrings in `crates/runner/mitm-addon/src/usage/providers/model_provider.py`.
- No child issues, dependency relationships, migrations, rollout gates, or follow-up slices are required.
- Apply the package's existing Ruff format/lint, BasedPyright, lockfile, and pytest validation before submission.

seven332:
## Summary

Clarifies the boolean return contracts of the model-provider usage and observation reporters.

The docstrings now distinguish the reporting-path status returned by each function from the process-local buffer's per-call source admission. They document the exact false gates, the complete-context true condition, duplicate-key behavior, and `accepted_source_keys` as the source of truth for newly admitted payload keys.

## Scope

- Updated only the two public docstrings in `crates/runner/mitm-addon/src/usage/providers/model_provider.py`.
- Preserved runtime behavior, buffer admission, terminal lifecycle behavior, webhook delivery, and existing test behavior.
- No API, persistence, migration, deployment, dependency, fallback, or compatibility changes.

## Validation

- `/home/user/.local/bin/uv lock --check`
- `/home/user/.local/bin/uv sync --locked`
- `/home/user/.local/bin/uv run --no-sync ruff format --check .`
- `/home/user/.local/bin/uv run --no-sync ruff check .`
- `/home/user/.local/bin/uv run --no-sync basedpyright -p .`
- `/home/user/.local/bin/uv run --no-sync python -m pytest tests/` — 7,192 passed; 59 existing dependency/code warnings.

Closes #30116

### DIFF
--- a/crates/runner/mitm-addon/src/usage/providers/model_provider.py
+++ b/crates/runner/mitm-addon/src/usage/providers/model_provider.py
@@ -152,21 +152,32 @@ def report_model_provider_usage(
 ) -> bool:
     """Buffer billable token usage for model-provider responses if available.
 
-    Accepted reporting requires all universal gates to pass:
+    The function returns ``False`` when any of these gates fails:
 
     - ``firewall_name`` starts with ``model-provider:``.
     - ``run_id`` is non-empty.
     - ``firewall_billable`` is truthy.
-    - At least one model-provider usage source is available.
-    - At least one ``MODEL_USAGE_CATEGORIES`` value has a positive integer
-      quantity.
+    - At least one billable event is built from the available model-provider
+      usage sources, including a positive integer quantity in
+      ``MODEL_USAGE_CATEGORIES``.
     - ``sandbox_token`` and ``get_api_url()`` are both non-empty.
 
-    Returns whether usage was accepted into the reporting path. When provided,
-    the caller-owned set accumulates source payload keys accepted by the
-    process-local buffer. All failed gates are silent by design except missing
-    sandbox token or API URL, which writes an underbilling signal because
-    billable usage cannot be reported.
+    It returns ``True`` when all gates pass, at least one event is built, and
+    the complete reporting context allows the process-local buffer to be
+    invoked. This boolean indicates that the reporting path was reached; it
+    does not indicate how many events the buffer admitted or that webhook
+    delivery completed. Process-local source-key deduplication can therefore
+    admit zero events even when this function returns ``True``.
+
+    When provided, ``accepted_source_keys`` receives only source payload keys
+    newly admitted by the buffer during this call. It is the per-call source
+    of truth for which payload keys were admitted. This reporting-path status
+    is consumed by ``terminal_usage.report_model_provider_usage_once``
+    separately from those per-call admission keys.
+
+    All failed gates are silent by design except missing sandbox token or API
+    URL, which writes an underbilling signal because billable usage cannot be
+    reported.
     """
     if not run_id:
         return False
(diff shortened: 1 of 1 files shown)


seven332:
LGTM

## Code Review: PR #30132

**Reviewed head:** `153b7bdf778519037a517cd919798899df6f6a04`

### Summary

The PR accurately documents the two model-provider reporter return contracts. It distinguishes reporting-path availability from process-local buffer admission and identifies `accepted_source_keys` as the per-call source-admission signal.

### Findings

No findings.

### Testing

Coverage: Adequate for this documentation-only change. The full mitm-addon suite passed with 7,192 tests, and the locked-environment, Ruff format, Ruff lint, and BasedPyright checks passed. No runtime behavior or tests were changed.

### Verdict

LGTM

PR #30135, opened 2026-08-28
refactor(guest-agent): reuse file-size limit guard

### LINKED ISSUE #30120: refactor(rust): reuse guest-agent file-size limit guard in metrics test
## Problem and evidence

The guest-agent integration suite already has one reusable owner for temporary process file-size limits:

- `crates/guest-agent/tests/common/mod.rs:801-887` defines `SoftFileSizeLimitGuard` and `set_soft_file_size_limit`, including `RLIMIT_FSIZE` validation, `SIGXFSZ` installation, rollback when `setrlimit` fails, explicit restoration, and drop-based restoration.
- `crates/guest-agent/tests/cli_agent_log_persistence_failure.rs:43-60` and `crates/guest-agent/tests/codex_app_server_agent_log_persistence_failure.rs:44-51` use that shared boundary.
- `crates/guest-agent/tests/integration/mod.rs:6-7` already imports the same common module into the integration test binary.

However, `crates/guest-agent/tests/integration/metrics.rs:160-200` independently implements the same `getrlimit` / `SIGXFSZ` / `setrlimit` / restore sequence for the cached-write-failure scenario, with a local error helper at lines 215-219.

This is not only repeated unsafe test plumbing. The two implementations already differ on failure paths: the shared helper restores the prior `SIGXFSZ` disposition if installing the new soft limit fails and its guard restores state during early returns or unwinding, while the metrics copy has no guard and returns directly after a failed `setrlimit`. The dedicated child process contains the blast radius, but the test now has a second owner for a process-global mutation and its cleanup contract.

## Decision

Reuse `crate::common::set_soft_file_size_limit(first_record_bytes)` inside `run_cached_write_failure_scenario`. Keep the returned guard alive through the expected failed metrics tick, explicitly restore it before advancing to the recovery tick, and retain the existing exact-test child-process isolation.

Remove the local `getrlimit`, signal installation, limit installation, manual restoration, and `operation_error` implementation from `integration/metrics.rs`.

## Scope

- Test-only refactor; no production behavior, public API, metrics schema, cadence, runtime path, or write-recovery semantics change.
- Keep the isolated child boundary because `RLIMIT_FSIZE` and `SIGXFSZ` are process-global.
- Do not generalize unrelated child-process launch helpers as part of this issue.

## Acceptance criteria

- `integration/metrics.rs` uses the existing common file-size-limit guard instead of directly calling `getrlimit`, `setrlimit`, or `signal`.
- The guard is explicitly restored before the metrics loop's recovery tick so the test continues to prove that the cached handle is dropped and the configured path is reopened.
- The exact-test child process remains the isolation boundary for the process-global limit and signal disposition.
- The duplicate unsafe resource-limit code and local `operation_error` helper are removed.
- The retained-inode and cached-write-failure scenarios continue to pass via `cargo test -p guest-agent --test integration metrics`.
- Guest-agent formatting, Clippy, Rustdoc, and affected-crate tests pass without lint suppressions or skipped tests.

## Delivery notes

- This should remain one independently reviewable test-only Rust PR.
- No migration, feature switch, rollout, or compatibility shim is required.

seven332:
## Summary

- reuse the shared `SoftFileSizeLimitGuard` in the cached metrics write-failure scenario
- keep the guard active through the failed tick and explicitly restore it before the recovery tick
- remove the duplicate libc resource-limit sequence and its local error helper

## Scope

- test-only refactor in the guest-agent metrics integration test
- preserve the exact-test child-process isolation and existing failure/recovery assertions
- no production behavior, API, protocol, persistence, migration, feature switch, or deployment compatibility changes

## Validation

- `cargo test -p guest-agent --test integration metrics`
- `cargo test -p guest-agent`
- pre-commit: workspace `cargo fmt --all`, `cargo clippy --all-targets --all-features`, `cargo doc --workspace --all-features --no-deps`, and file-size check

Closes #30120

### DIFF
--- a/crates/guest-agent/tests/integration/metrics.rs
+++ b/crates/guest-agent/tests/integration/metrics.rs
@@ -157,48 +157,14 @@ async fn run_cached_write_failure_scenario() -> TestResult {
     std::fs::rename(&metrics_path, &moved_path)?;
     let first_record_bytes = std::fs::metadata(&moved_path)?.len();
 
-    let mut original_limit = std::mem::MaybeUninit::<libc::rlimit>::uninit();
-    // SAFETY: `original_limit` points to writable storage for one `rlimit`.
-    let get_limit_result =
-        unsafe { libc::getrlimit(libc::RLIMIT_FSIZE, original_limit.as_mut_ptr()) };
-    if get_limit_result != 0 {
-        return Err(operation_error("getrlimit").into());
-    }
-    // SAFETY: successful `getrlimit` initialized the full value.
-    let original_limit = unsafe { original_limit.assume_init() };
-    // SAFETY: installing `SIG_IGN` for SIGXFSZ is process-global, and this test
-    // runs alone in a dedicated child process.
-    let original_handler = unsafe { libc::signal(libc::SIGXFSZ, libc::SIG_IGN) };
-    if original_handler == libc::SIG_ERR {
-        return Err(operation_error("install SIGXFSZ handler").into());
-    }
-
-    let constrained_limit = libc::rlimit {
-        rlim_cur: first_record_bytes,
-        rlim_max: original_limit.rlim_max,
-    };
-    // SAFETY: the hard limit is unchanged and the process is isolated from all
-    // other tests while the soft limit is constrained.
-    let set_limit_result = unsafe { libc::setrlimit(libc::RLIMIT_FSIZE, &constrained_limit) };
-    if set_limit_result != 0 {
-        return Err(operation_error("set RLIMIT_FSIZE").into());
-    }
+    let limit_guard = crate::common::set_soft_file_size_limit(first_record_bytes)?;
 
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #30135

**Reviewed head:** `429e82ff903632c03d8d2034f4554ab86254e96f`

### Summary

The PR replaces the metrics test's duplicate `RLIMIT_FSIZE`/`SIGXFSZ` lifecycle with the existing shared guard while preserving the exact child-process boundary and the failed-write/recovery tick ordering.

### Findings

No findings.

### Testing

Focused metrics integration tests, guest-agent affected-crate tests, formatting, all-target/all-feature Clippy, and Rustdoc passed locally. The pre-commit workspace Rust checks also passed. Remote PR CI is handled separately and was still pending during this review.

### Verdict

LGTM

PR #30139, opened 2026-08-28
fix(python): omit encoding for empty streamed request bodies

### LINKED ISSUE #30126: bug(python): omit binary encoding for empty streamed request bodies
## Summary

`crates/runner/mitm-addon/src/body_capture.py` unconditionally emits `request_body_encoding: "binary"` when decoding a streamed request returns `None`, even when the captured stream contains zero bytes.

This conflicts with the capture schema documented in the same module: empty bodies should have neither a body field nor an encoding field. It also differs from `_set_body_capture_failure()`, which emits `binary` only when a non-empty body was observed.

## Reproduction

At commit `f3bb3c589879d6782f7942afff00b8e9af242706`:

1. Configure request stream capture with `request_streaming.configure_request_stream(flow)`.
2. Complete a zero-byte request whose headers include a syntactically valid but unsupported `Content-Encoding` value.
3. Mark the stream complete and call `body_capture.add_capture_fields()`.

The stream reports `total_bytes == 0`, but the resulting entry contains:

```python
{"request_body_encoding": "binary"}
```

There is no `request_body` field.

## Impact

Persistent network logs describe an empty request as a binary body. Consumers cannot distinguish this from a non-empty body that was suppressed after a decoding failure, and streamed request capture is inconsistent with the documented empty-body contract and the buffered/response paths.

## Acceptance criteria

- Empty streamed request bodies do not emit `request_body_encoding`, including when decoding fails because of unsupported dependency metadata.
- `request_body_truncated` remains independent and is still emitted when applicable.
- Non-empty streamed request bodies that fail decoding continue to emit `request_body_encoding: "binary"`.
- Add regression coverage for the zero-byte streamed request case.

seven332:
## Summary

- route streamed request decode failures through the shared body-capture failure policy
- omit the `binary` encoding marker when an unsupported encoding is attached to a zero-byte request stream
- preserve incomplete-stream truncation metadata and non-empty decode-failure behavior

## Scope

- No API, schema, database, migration, UI, or configuration changes.
- Existing stored network-log entries are not rewritten.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (7,194 passed)

Closes #30126

### DIFF
--- a/crates/runner/mitm-addon/src/body_capture.py
+++ b/crates/runner/mitm-addon/src/body_capture.py
@@ -507,9 +507,12 @@ def add_capture_fields(flow: http.HTTPFlow, log_entry: dict) -> None:
                     max_output=BODY_CAPTURE_LIMIT + 1,
                 )
                 if request_body is None:
-                    if request_stream_body.truncated or request_stream_incomplete:
-                        log_entry["request_body_truncated"] = True
-                    log_entry["request_body_encoding"] = "binary"
+                    _set_body_capture_failure(
+                        log_entry,
+                        "request",
+                        raw_request_body,
+                        truncated=request_stream_body.truncated or request_stream_incomplete,
+                    )
                 else:
                     _set_body_fields(
                         log_entry,
--- a/crates/runner/mitm-addon/tests/test_body_capture_stream_buffer.py
+++ b/crates/runner/mitm-addon/tests/test_body_capture_stream_buffer.py
@@ -232,6 +232,33 @@ class TestBodyCaptureStreamBuffer:
         assert "request_body_encoding" not in entry
         assert entry["request_body_truncated"] is True
 
+    @pytest.mark.parametrize(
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #30139

**Reviewed head:** `faf9643bf81fc3284ec5e9d2b173bc7ffc8e3c41`

### Summary

The change restores the documented empty-body capture invariant by reusing the existing failure helper, while preserving truncation metadata and non-empty decode-failure behavior.

### Findings

No findings.

### Testing

Complete and incomplete zero-byte streams are covered through the real `add_capture_fields()` entry point, and the existing non-empty failure matrix remains intact. Ruff format/lint, basedpyright, focused pytest, and the full 7,194-case addon suite passed locally. GitHub Actions CI is intentionally left to the following check stage.

### Verdict

LGTM

PR #30166, opened 2026-08-28
refactor(runner): remove legacy mitmdump runtime reader

lancy:
## Summary

- make mitmdump runtime ownership canonical-only across the initial procfs scan, immediate environment recheck, and post-pidfd recheck
- remove the retired legacy byte reader, source/resolution/conflict machinery, source telemetry, startup migration summary, and its exact Axiom INFO exception/test
- keep the production `OKOU_MITMDUMP_RUNTIME_DIR` writer unchanged and byte-equal to the private `TMPDIR`
- replace the migration-source matrix with focused canonical ownership and retired-legacy negative coverage

## Preserved safety behavior

- private launch-path and UID validation, unrelated-directory protection, bounded procfs scanning, and PID/start-time generation checks
- environment rechecks around pidfd acquisition, pidfd signalling, ESRCH/zombie handling, process exit timeouts, locks, and cleanup sequencing
- process groups, parent-death behavior, CA preparation, runner-token isolation, addon behavior, registry semantics, and proxy usage handling
- canonical absent, empty, non-Unicode, exact-byte, exact-path, and marker-removal behavior

## JIT inventory

- pre-edit refreshed base: `db45db7c5ade7ac3ddeb4220a07466013ce377e8`
- final rebased/reviewed base: `aae87e417fd77ced9a24ea0917fe8250222948fd`
- final head: `ad45b3087289b5dec28464f964f1a64b55f1a611`
- complete pre-edit caller inventory confined the active reader/source/conflict/summary surface to the four scoped Runner files
- final fully paginated open-PR audit: 30 open PRs, 31 changed-file pages, 544/544 declared/fetched files, zero scoped-file overlap
- exact diff: four scoped files, 134 insertions and 618 deletions

## Verification

- `cargo fmt --manifest-path crates/Cargo.toml --all --check`
- `git diff --check`
- `CARGO_BUILD_JOBS=1 CARGO_PROFILE_DEV_DEBUG=0 cargo clippy --manifest-path crates/Cargo.toml -p runner --all-targets --all-features -- -D warnings`
- `CARGO_BUILD_JOBS=1 CARGO_PROFILE_DEV_DEBUG=0 cargo check --manifest-path crates/Cargo.toml -p runner --all-targets --all-features`
[...]

Issue #30161, opened 2026-08-28
refactor(runner): remove legacy mitmdump runtime marker reader


lancy:
Parent EPIC: #28914

## Background

Verified just in time against clean current `main@1c738e9e2aa611616dbd947243e9ab8f8f1b2404` on 2026-08-28.

`VM0_MITMDUMP_RUNTIME_DIR` is the last compatibility input for the private mitmdump/PyInstaller runtime marker. The staged migration is complete:

- Expansion reader #28989 / PR #29030 merged as `29b82dd28d8693471db0a7a00800be0830650ae3` and deployed a lifecycle-aware reader for legacy-only, canonical-only, equal-dual, and conflicting-dual descendants.
- Canonical-only writer #29934 / PR #29944 merged as `5c160a8e361e06cf6587f37ba9c4ae061dc2f139`; every new mitmdump launch now writes exactly one `OKOU_MITMDUMP_RUNTIME_DIR` equal to its private `TMPDIR` and writes no legacy marker.
- Startup visibility #30114 / PR #30133 merged as `877af7fc5477a32a58c1aff22d0c2d45a99a8f5d` and was released on all production hosts in Runner `0.178.0`, followed by successor `0.178.1`.

The terminal production gate is recorded at https://github.com/vm0-ai/vm0/issues/28914#issuecomment-5452219755.

A terminal, non-partial Axiom query over `2026-08-28T10:40:00Z..2026-08-28T11:55:26Z` found exactly six initial reconciliation summaries: all three production hosts reported once on `0.178.0` and once on `0.178.1`. Every row had `legacy_only_seen=false`, `canonical_only_seen=false`, `equal_dual_seen=false`, and `stale_launch_directory_count=0`. No marker value was queried or returned.

Read-only masked production `runner_state` then proved both old services reached zero runs and zero allocation, stopped refreshing heartbeats, and disappeared after the five-minute stale threshold. Two additional reads more than one heartbeat period apart remained stable with exactly the three current `running` rows and no old `draining` or `stopping` row.

All seven supported rollback targets from Runner `0.177.4` through `0.178.1` are descendants of canonical-only writer merge `5c160a8e361e06cf6587f37ba9c4ae061dc2f139`, with `behind_by=0` and an exact merge base at that writer merge. A supported rollback cannot create a new legacy marker.

## Current inventory

A complete exact-key and related-symbol search on current main found:

[...]

PR #30184, opened 2026-08-28
test(runner): stabilize heartbeat coalescing coverage

### LINKED ISSUE #30179: test(runner): stabilize heartbeat coalescing race coverage
## Background

`heartbeat_triggers_coalesce_into_one_current_mode_follow_up` is nondeterministic on the current
`main`-derived Runner test suite. It failed while validating PR #30093, whose diff does not touch
`crates/runner/src/cmd/start` or this test.

## Observed behavior

On commit `5c6e2f41f1248d03caf3723c8dd19c84314991e4`, the full Runner suite failed with an unexpected
third heartbeat:

```text
thread 'cmd::start::tests::main_loop::heartbeat::heartbeat_triggers_coalesce_into_one_current_mode_follow_up' panicked at runner/src/cmd/start/tests/main_loop/heartbeat.rs:217:9:
  left: 3
 right: 2
```

The same test was then run repeatedly in isolation. The first attempt passed, while the second
failed through a different race:

```text
thread 'cmd::start::tests::main_loop::heartbeat::heartbeat_triggers_coalesce_into_one_current_mode_follow_up' panicked at runner/src/cmd/start/mod.rs:1179:13:
runner did not observe routine heartbeat request within 5s
```

Earlier full-suite runs on the same production code also passed, confirming the result depends on
scheduling rather than a deterministic assertion mismatch.

## Reproduction

From `crates/`:

```bash
cargo test -p runner --all-targets --all-features --quiet
```

Or repeat the exact test:

```bash
for attempt in 1 2 3 4 5; do
  cargo test -p runner --bin runner \
    cmd::start::tests::main_loop::heartbeat::heartbeat_triggers_coalesce_into_one_current_mode_follow_up \
    -- --exact --nocapture || exit 1
done
```

The isolated loop reproduced one pass followed by one failure on 2026-08-29.

## Expected behavior

The test deterministically observes exactly one coalesced follow-up heartbeat and does not depend
on scheduler timing between paused-time advancement, observer events, and the heartbeat worker.

## Acceptance criteria

- Synchronize the test and/or implementation boundary so each simulated tick is acknowledged
  before the next scheduling transition.
- The test reliably distinguishes one coalesced follow-up from overlapping or extra requests.
- Repeated isolated runs and the full Runner suite pass without increasing timeouts or skipping the
  test.
- Preserve the production contract that concurrent heartbeat triggers coalesce while the active
  send is blocked and that the follow-up samples the current Runner mode.

## Relevant files

- `crates/runner/src/cmd/start/tests/main_loop/heartbeat.rs`
- `crates/runner/src/cmd/start/mod.rs`

## Scope

This is intentionally separate from #30093 because that PR changes deployment GC only and has no
`cmd/start` diff.

seven332:
## Summary

- add a test-only manual routine-heartbeat source while preserving the production 10-second interval and missed-tick behavior
- drive the coalescing regression test with explicit ticks instead of Tokio's paused clock
- keep coverage at the real runner main-loop entry point, including lifecycle sampling, status persistence, provider calls, single-flight behavior, and heartbeat sequencing

## Why

The paused-clock test intermittently advanced unrelated runner deadlines while waiting for routine-heartbeat work to be observed. Depending on scheduling, those deadlines could move the runner into `stopping` or let the assertion run before the intended tick was processed, producing either an extra heartbeat or a timeout. The production coalescing behavior was not the source of the race.

## Validation

- target test: 100/100 repeated passes
- `cargo test -p runner cmd::start::tests::main_loop::heartbeat --all-features` (9 passed)
- `cargo test -p runner --all-targets --all-features --quiet -- --test-threads=1` (3374 passed, 26 ignored, plus 1 integration test passed)
- repository pre-commit hooks: workspace Rust format, Clippy, docs, and file-size checks passed

## Scope

Production heartbeat cadence and `HeartbeatController` behavior are unchanged. Unrelated proxy runtime-marker discovery flakes observed only during parallel full-suite runs are intentionally excluded; those legacy/conflict scenarios are already being replaced by #30161 / #30166, and the serial full runner suite is green.

Closes #30179

### DIFF
--- a/crates/runner/src/cmd/start/mod.rs
+++ b/crates/runner/src/cmd/start/mod.rs
@@ -169,6 +169,38 @@ async fn sleep_until_optional_instant(deadline: Option<Instant>) {
     }
 }
 
+enum RoutineHeartbeatTrigger {
+    Interval(tokio::time::Interval),
+    #[cfg(test)]
+    Manual(mpsc::UnboundedReceiver<()>),
+}
+
+impl RoutineHeartbeatTrigger {
+    fn interval() -> Self {
+        let mut interval = tokio::time::interval_at(
+            tokio::time::Instant::now() + HEARTBEAT_PERIOD,
+            HEARTBEAT_PERIOD,
+        );
+        interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+        Self::Interval(interval)
+    }
+
+    async fn tick(&mut self) {
+        match self {
+            Self::Interval(interval) => {
+                interval.tick().await;
+            }
+            #[cfg(test)]
+            Self::Manual(receiver) => {
+                receiver
+                    .recv()
+                    .await
+                    .expect("manual routine heartbeat sender should remain open");
+            }
+        }
+    }
+}
+
 type WorkspaceCacheChangeFuture = BoxFuture<
     'static,
     (
@@ -935,6 +967,7 @@ async fn run_start_with_home(
             test_observer: StartLoopTestObserver::default(),
             before_initial_workspace_cache_scan: None,
             after_initial_workspace_cache_scan: None,
+            manual_routine_heartbeat_rx: None,
         },
     };
 
@@ -1034,6 +1067,7 @@ struct RunTestHooks {
     test_observer: StartLoopTestObserver,
     before_initial_workspace_cache_scan: Option<StartLoopTestGate>,
     after_initial_workspace_cache_scan: Option<StartLoopTestGate>,
+    manual_routine_heartbeat_rx: Option<mpsc::UnboundedReceiver<()>>,
 }
 
 enum SignalSource {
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #30184

**Reviewed head:** `32b35a6a49fcb4ecd083be499847d96d2d6da73d`

### Summary

The test-only manual trigger removes unrelated virtual-clock advancement while preserving the production interval and the real runner main-loop, lifecycle, provider, status, and `HeartbeatController` paths. The exact coalescing, live-mode, sequence, and single-flight assertions remain intact.

### Findings

No findings.

### Testing

Reviewed 100 repeated target passes, the complete heartbeat module, the serial full Runner suite, strict Runner Clippy, rustdoc, formatting, and workspace Rust pre-commit checks. Parallel suite noise was limited to two unrelated legacy-marker scenarios already being replaced by #30161 / #30166.

### Verdict

LGTM

PR #30253, opened 2026-08-29
fix(ci): ignore non-loaded runner units during image preparation

### LINKED ISSUE #30235: bug(ci): ignore non-loaded Runner units during image preparation
## Background

`.github/scripts/prepare-runner-image.sh` performs forceful cleanup of transient
Runner services for the current CI `job_ref` before replacing its binary and
Runner directories. The cleanup discovers matching units with:

```text
systemctl list-units --all --full --no-legend --plain <pattern>
```

This issue was identified while reviewing #30093, but it exists independently
on `main` and does not depend on that PR.

## Problem

`discover_primary_units` parses only the unit-name column. It does not inspect
the `LOAD` column, so a stale row with `LoadState=not-found` is treated as a
currently loaded CI service.

The script then includes that row in the bulk `systemctl stop` call. A stale
unit can make the stop command fail before the script reaches the intended
binary and Runner-directory replacement. The second discovery pass has the
same ambiguity.

The current shell fixture emits only `loaded` rows and therefore does not cover
the stale-row case.

## Expected behavior

- Only matching units whose `LOAD` column is exactly `loaded` participate in
  the current job's stop and verification flow.
- `not-found`, `masked`, and other non-loaded rows are ignored.
- A malformed matching row that claims to be loaded must fail closed rather
  than silently authorizing cleanup.
- A failed `systemctl list-units` query must continue preventing shared-path
  mutation.
- Current unit-name and numeric service-suffix validation remains unchanged.

## Acceptance criteria

- Runner Image preparation filters discovered rows by `LoadState=loaded`,
  either in the systemctl query or while parsing its structured columns.
- A shell regression test supplies a mixture of loaded and non-loaded matching
  units and proves that only loaded units are stopped and verified.
- Tests retain the existing discovery-failure, stop-failure, sticky-unit, and
  namespace-validation behavior.
- Relevant Bash syntax and `.github/scripts/tests/prepare-runner-image-test.sh`
  checks pass.

## Relevant files

- `.github/scripts/prepare-runner-image.sh`
- `.github/scripts/tests/prepare-runner-image-test.sh`

## Scope

This issue covers the Runner Image preparation cleanup only. It does not
broaden generic `runner gc` deletion candidates, change semver directory GC,
or reintroduce #29907's cross-namespace redesign.

seven332:
## Summary

- parse the `LOAD`, `ACTIVE`, and `SUB` columns from Runner service discovery
- admit only matching numeric units whose load state is exactly `loaded`
- fail closed for malformed matching loaded rows before shared runner paths mutate
- cover mixed loaded, `not-found`, and `masked` rows through the production script entry point

## Scope

- no generic `runner gc` behavior changes
- no service or directory naming changes
- no workflow YAML, API, persisted data, or runner protocol changes

## Validation

- `bash -n .github/scripts/prepare-runner-image.sh`
- `bash -n .github/scripts/tests/prepare-runner-image-test.sh`
- `.github/scripts/tests/prepare-runner-image-test.sh`
- `git diff --check`
- `./scripts/check-file-size.sh .github/scripts/prepare-runner-image.sh .github/scripts/tests/prepare-runner-image-test.sh`

Closes #30235

### DIFF
--- a/.github/scripts/prepare-runner-image.sh
+++ b/.github/scripts/prepare-runner-image.sh
@@ -103,7 +103,7 @@ UNIT_PREFIX="vm0-runner-${JOB_REF}-"
 
 declare -a PRIMARY_UNITS=()
 discover_primary_units() {
-  local output unit rest suffix
+  local output line unit load_state active_state sub_state description suffix
   if ! output=$(sudo systemctl list-units \
     --all --full --no-legend --plain "${UNIT_PREFIX}*.service" 2>&1); then
     printf '%s\n' "$output" >&2
@@ -111,13 +111,21 @@ discover_primary_units() {
   fi
 
   PRIMARY_UNITS=()
-  while read -r unit rest; do
+  while IFS= read -r line; do
+    read -r unit load_state active_state sub_state description <<< "$line"
     [ -n "$unit" ] || continue
     case "$unit" in
       "${UNIT_PREFIX}"*.service)
         suffix=${unit#"${UNIT_PREFIX}"}
         suffix=${suffix%.service}
         if [[ "$suffix" =~ ^[0-9]+$ ]]; then
+          if [ "$load_state" != "loaded" ]; then
+            continue
+          fi
+          if [ -z "$active_state" ] || [ -z "$sub_state" ]; then
+            echo "malformed loaded runner service row: ${line}" >&2
+            return 1
+          fi
           PRIMARY_UNITS+=("$unit")
         fi
         ;;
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #30253

**Reviewed head:** `65a8b8813690690bc051d65b8b6a8e88530cbbc0`

### Summary

The PR applies exact `LOAD=loaded` admission to both Runner service discovery passes without adding systemd calls or changing the existing namespace and pre-mutation safety boundaries. The script-entry regression suite covers mixed loaded/non-loaded rows and malformed loaded output while retaining the prior cleanup failure cases.

### Findings

No findings.

### Testing

- Passed both Bash syntax checks.
- Passed `.github/scripts/tests/prepare-runner-image-test.sh` on the reviewed head.
- Passed whitespace and staged file-size validation.
- Live metal-host systemd execution was not run from this workspace; the local container has no `systemctl` binary.

### Verdict

LGTM

PR #30254, opened 2026-08-29
fix(runner): fail closed on unit state query errors

### LINKED ISSUE #30241: bug(runner): fail closed when active-job unit-state queries fail
## Background

`runner service stop` and `runner service uninstall` use an active-jobs gate to avoid killing in-flight jobs unless the operator explicitly passes `--force`. The gate first queries systemd to determine whether the target unit is active, then inspects the Runner's `status.json` when needed.

The behavior originated in #9062 and its concrete orchestration is covered by #29208.

## Problem

The current gate converts every error from `is_unit_active` into `false`:

```rust
let active = ops.is_unit_active(unit).await.unwrap_or_else(|e| {
    warn!(... "cannot check unit state — skipping active-jobs gate");
    false
});
```

Consequently, an unavailable systemd bus, failed `systemctl show`, malformed response, permission error, or another unit-state query failure is treated as proof that the unit is inactive. A no-force stop or uninstall can then proceed to a later service mutation without checking active jobs. If the query failure is transient and the subsequent mutation succeeds, in-flight jobs can be killed even though the operator did not authorize `--force`.

This is a static code-path reproduction. The test `check_active_jobs_gate_bypasses_unavailable_unit_state` currently codifies the unsafe behavior; no production incident has been established.

## Expected behavior

- A confirmed inactive unit may bypass stale status state because there is no live process to protect.
- A unit-state query error is unknown state, not inactive state, and must fail closed before stop or uninstall mutation.
- `--force` remains the explicit operator override and continues to bypass the gate before systemd or status queries.
- The surfaced error should identify the target unit and preserve the underlying diagnostic without exposing secrets.

## Acceptance criteria

- Without `--force`, an `is_unit_active` error prevents both `runner service stop` and `runner service uninstall` from mutating the service.
- A confirmed inactive unit retains its existing bypass behavior.
- A confirmed active unit continues through the active-jobs status check.
- `--force` retains its current short-circuit behavior and does not require a successful unit-state query.
- Deterministic Runner integration coverage exercises query errors through the concrete gate and proves that stop/uninstall mutation is not attempted.
- Existing actionable error context is preserved, and tests no longer describe unavailable unit state as a valid bypass.

## Relevant code

- `crates/runner/src/cmd/service/gate.rs`
- `crates/runner/src/cmd/service/stop.rs`
- `crates/runner/src/cmd/service/mod.rs`
- `crates/runner/src/cmd/service/systemctl.rs`

## Scope

- #30044 separately fixes selection of the correct Runner state directory when service suffix and Runner dirname differ.
- This issue changes only the policy for a failed unit-state query; it does not redefine the separate missing or malformed `status.json` policy.
- This issue does not change `service drain`, systemd unit definitions, signal handling, or the explicit destructive semantics of `--force`.

seven332:
## Summary

- fail closed when the active-jobs gate cannot determine the target systemd unit state
- preserve the target unit, attempted command, underlying diagnostic, and explicit `--force` guidance
- cover the shared gate plus stop and uninstall orchestration to prove no service mutation follows a query error

## Compatibility

- preserves the confirmed-inactive bypass and active-unit status checks
- preserves the early `--force` override without requiring systemd or status queries
- changes only no-force stop/uninstall behavior for unknown unit state
- introduces no CLI, API, schema, persisted-state, systemd-unit, signal, or deployment-protocol changes
- leaves retry policy, service drain, status-file policy, and the documented post-gate TOCTOU window unchanged

## Validation

- `cargo test -p runner cmd::service --all-features`
- `RUSTDOCFLAGS='-D warnings' cargo doc -p runner --no-deps`
- pre-commit Rust formatting, documentation, Clippy, file-size, and commit-message checks

Closes #30241

### DIFF
--- a/crates/runner/src/cmd/service/gate.rs
+++ b/crates/runner/src/cmd/service/gate.rs
@@ -4,7 +4,7 @@ use crate::error::{ActiveJobsError, RunnerError, RunnerResult};
 use crate::ids::RunId;
 use crate::paths::HomePaths;
 use crate::status_file::{self, StatusFileReadError, StatusForGate};
-use tracing::{info, warn};
+use tracing::info;
 
 use super::diagnostic::status_field_preview;
 use super::target::RunnerServiceUnit;
@@ -168,6 +168,8 @@ pub(super) async fn read_runner_status(
 /// the unit's selected config resolves to one verified live Runner record and
 /// that record's status is readable. Without that correlation, the gate cannot
 /// prove that teardown is safe. `--force` remains the explicit override.
+/// A systemd query error is likewise unknown state rather than confirmed
+/// inactivity, so it fails closed before selected-config or status inspection.
 ///
 /// When the runner's `mode == "draining"`, we still refuse but flip the
 /// `draining` flag so the error renders a wait-or-force message (the
@@ -201,10 +203,12 @@ pub(super) async fn check_active_jobs_gate(
     }
 
     // (1) Dead-runner short-circuit.
-    let active = ops.is_unit_active(unit).await.unwrap_or_else(|e| {
-        warn!(unit = %unit.unit_name(), error = %e, "cannot check unit state — skipping active-jobs gate");
-        false
-    });
+    let active = ops.is_unit_active(unit).await.map_err(|error| {
+        RunnerError::Internal(format!(
+            "cannot determine whether {} is active while checking active jobs before service {command_name}: {error}; retry or pass --force to override the active-jobs gate",
+            unit.unit_name()
+        ))
+    })?;
     if !active {
         return Ok(());
     }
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #30254

**Reviewed head:** `af154090d8174a114a97cec3450e552e2cc32127`

### Summary

The shared active-jobs gate now treats a systemd unit-state query error as unknown state and returns actionable context before either stop or uninstall can mutate the service. Confirmed inactivity, active-unit inspection, and the explicit `--force` short circuit remain unchanged.

### Findings

No findings.

### Testing

Local formatting, focused Runner service tests, all-target Clippy, rustdoc, the full Runner all-target test suite, and pre-commit hooks passed. Regression coverage exercises the shared gate and both stop/uninstall orchestration boundaries. Real-host systemd fault injection was not run locally; submitted-head CI remains pending verification.

### Verdict

LGTM

PR #30234, opened 2026-08-29
refactor(platform): unify markdown and editor dependencies

e7h4n:
## Summary

- parse shared Markdown with the same `marked@17.0.5` instance used by `@tiptap/markdown`, while retaining the application-owned HAST, card, and Mermaid pipeline
- replace StarterKit in both Tiptap editors with explicit extensions and intentionally omit editor link support, removing Link and `linkifyjs` from the dependency graph
- align `@floating-ui/dom` on Base UI's `1.8.0` version and remove obsolete direct remark/CJK parser dependencies

This is a direct cutover with no feature switch or parser fallback. Closed Mermaid fences still render as diagrams, streaming fences remain code, and invalid or unsupported Mermaid falls back to the ordinary code block.

## Bundle impact

- vendor raw: 2,347.95 kB -> 2,136.62 kB (-211.33 kB, -9.0%)
- vendor gzip: 687.83 kB -> 623.67 kB (-64.16 kB, -9.3%)
- runtime/app/vendor/worker chunk topology is unchanged

## Validation

- `pnpm format`
- `pnpm lint`
- staged workspace, Platform, and API type checks
- `markdown-behavior.test.tsx` (7 passed)
- `markdown.test.tsx` (23 passed)
- `instructions-tab.test.tsx` (3 passed)
- `chat-composer-editor-and-suggestions.test.tsx` (22 passed; one Realtime mock retry after an initial HTTP 500)
- production `pnpm build`
- `pnpm knip` reports the existing repository-wide unused-file/dependency/export baseline; it did not report the dependencies or files changed here

PR #30232, opened 2026-08-29
refactor(platform): select clerk bootstrap config in browser

e7h4n:
## Summary

- select the Clerk publishable key and core script URL in an early inline browser bootstrap based on `location.hostname`
- persist the selected public key for the app runtime while preserving the existing in-flight script merge and retry behavior
- remove Clerk-specific HTML rewriting from the Cloudflare Pages Worker so preview and production can share one build artifact

## Testing

- `pnpm format`
- `pnpm turbo run lint --concurrency=1 --log-order grouped`
- `pnpm knip`
- `pnpm turbo run check-types --concurrency=1 --filter='!api' --filter='!@okouai/app'`
- `pnpm --filter @okouai/app run check-types`
- `pnpm --filter api run check-types`
- `pnpm --filter @okouai/app run lint:build-config`
- `pnpm --filter @okouai/app exec vitest run src/__tests__/clerk-entrypoint.test.ts`
- `pnpm --filter @okouai/app exec vitest run src/__tests__/platform-runtime-environment.test.ts`
- `bash .github/scripts/tests/okou-app-worker-test.sh`

PR #30228, opened 2026-08-29
perf(platform): use default production minification

e7h4n:
## Summary

- remove the App-specific `keepNames` and detailed Rolldown minification overrides
- delegate production minification to Vite 8's client defaults
- update the focused build-config guard to prevent custom minification options from drifting back in

Vite 8.0.16 enables Oxc minification for client builds by default. Rolldown's `minify: true` already enables compression, mangling, and code generation, so the effective behavior change is restoring the default `keepNames: false`.

## Bundle size

Measured from fresh production builds of the same `713e58e581c660f73f836511118cecef551d6d29` source revision, using Node gzip level 9 and Brotli quality 11 across all emitted JavaScript:

| Encoding | Explicit `keepNames` config | Vite defaults | Delta |
| --- | ---: | ---: | ---: |
| Raw | 7,532,005 B | 6,975,925 B | -556,080 B (-7.38%) |
| Gzip-9 | 1,975,251 B | 1,880,709 B | -94,542 B (-4.79%) |
| Brotli-11 | 1,551,335 B | 1,495,148 B | -56,187 B (-3.62%) |

The SharedWorker stayed byte-identical. The reduction is isolated to the App, vendor, and Rolldown runtime bundles.

## Verification

- `pnpm format`
- `pnpm turbo run lint --concurrency=1 --log-order grouped`
- `pnpm knip`
- `pnpm turbo run check-types --concurrency=1 --filter='!api' --filter='!@okouai/app'`
- `pnpm --filter @okouai/app run check-types`
- `pnpm --filter api run check-types`
- `pnpm --filter @okouai/app run lint:build-config`
- production build hash-stability check with source maps enabled
- Sentry source-map debug ID injection for the mapped App, vendor, and SharedWorker artifacts

PR #30224, opened 2026-08-29
feat: isolate shared worker realtime by organization

e7h4n:
## Summary

- add a physically isolated `user-org:<userId>:<orgId>` Ably channel and grant it from the existing platform realtime token endpoint
- route chat database invalidations to exactly one channel selected by the Shared Chat Database feature switch: the legacy user channel while disabled, and the physically isolated user-org channel while enabled
- key Shared Worker Ably sessions by `(userId, orgId)`, start them from credential heartbeats, and background-fetch unseen thread events into the matching IndexedDB database
- propagate organization identity through every chat database publisher and require organization context on thread mutations that previously only loaded user auth
- cover background caching, token refresh, lifecycle, the three-session `userA/orgA`, `userA/orgB`, `userB/orgB` isolation case, and auth-blocked actor cleanup after the final client disconnects

## Verification

- `pnpm format`
- `NODE_OPTIONS=--max-old-space-size=3072 pnpm turbo run lint --concurrency=1 --log-order grouped`
- `pnpm check-types`
- Lefthook pre-commit checks, including Prettier, Knip, type checks, and file-size checks
- Platform Shared Database worker-runtime tests: 22 passed
- API thread-create route tests: 12 passed
- affected API chat-event and run-lifecycle BDD scenarios: 4 passed

## Notes

The full local `chat-threads.bdd.test.ts` file passed 32 of 35 tests. Its three search-projection cases returned an empty local projection; the same targeted failure reproduced on an untouched detached `HEAD`, so it is not introduced by this change.

PR #30263, opened 2026-08-29
fix(platform): stop the shared database worker from reading page globals

e7h4n:
## Problem

Production Sentry, 2026-08-29: every chat page load fails to start the shared database SharedWorker.

- **PLATFORM-9V** `Shared database worker failed to load Uncaught ReferenceError: document is not defined` (Chrome/macOS)
- **PLATFORM-9W** same failure on Safari (`Can't find variable: document`)
- **PLATFORM-9T** `SharedDatabaseTransportError: Shared database worker failed to load`

## Root cause

`#30232` replaced the build-time `import.meta.env.VITE_CLERK_PUBLISHABLE_KEY_*` value with a runtime read of the choice the `index.html` bootstrap had recorded on the document element:

```ts
function resolveClerkPublishableKey(): string {
  const publishableKey = document.documentElement.getAttribute(
    CLERK_PUBLISHABLE_KEY_ATTRIBUTE,
  );
```

`resolvePlatformRuntimeConfig()` calls it, and `src/lib/posthog.ts` calls `resolvePlatformRuntimeConfig()` at module scope. That module is in the SharedWorker graph:

```
src/shared-database-worker.ts
  -> src/shared-database/worker-signals.ts
  -> src/shared-database/worker-runtime.ts
  -> src/signals/external/chat-idb-store.ts
  -> src/signals/auth.ts
  -> src/lib/posthog.ts        // const RUNTIME_CONFIG = resolvePlatformRuntimeConfig()
  -> src/lib/platform-host.ts  // document.documentElement.getAttribute(...)
```

A SharedWorker has no `document`, so the worker script throws while it is still being evaluated and the page only observes the opaque `error` event.

Evidence, from the deployed asset `assets/shared-database-worker-BioTB9DO.js` executed in a DOM-less sandbox and mapped through its source map:

```
ReferenceError: document is not defined
  at resolveClerkPublishableKey   src/lib/platform-host.ts:171
  at resolvePlatformRuntimeConfig src/lib/platform-host.ts:183
  at <module scope>               src/lib/posthog.ts:6
  at main()                       src/shared-database-worker.ts:50
```

Before/after builds of the worker asset confirm the attribution:

| commit | worker bundle evaluates without a DOM | bare `document` tokens |
| --- | --- | --- |
| `3bb5c1390` (parent of #30232) | OK | 53 (all third-party, guarded) |
| `7b89569f9` (#30232) | `ReferenceError: document is not defined` | 54 |
| `main` @ `e119d898b` | `ReferenceError: document is not defined` | 54 |

The build at `3bb5c1390` already contains #30140, #30175, #30185, #30211, #30224 and #30228 and is clean, so none of those introduced it. #30234 merged after #30232 and is likewise not involved.

## Fix

1. **Select the key in the bundle, not from the DOM.** The SharedWorker is a second entry point into this bundle, so it needs what `index.html` already gets: both publishable keys inlined at build time, one selected by hostname. `resolvePlatformRuntimeConfig()` goes back to choosing between `VITE_CLERK_PUBLISHABLE_KEY_PROD` and `VITE_CLERK_PUBLISHABLE_KEY_PREVIEW`, so it resolves from `location` and build-time constants alone. Nothing reads `data-vm0-clerk-publishable-key` anymore, so the bootstrap stops writing it.

   The bootstrap and the bundle now select independently, so they are pinned together: `clerk-entrypoint.test.ts` executes the real `index.html` bootstrap and asserts, on every hostname it covers, that `resolvePlatformRuntimeConfig().clerkPublishableKey` matches the key the bootstrap chose.

2. **Take the DOM out of the worker graph.** `createChatIdbOpener` moves to `src/signals/external/chat-idb-opener.ts` with a required `reload` callback; `chat-idb-store.ts` keeps the auth-coupled `chatIdb$` and the page-side default opener. The worker no longer imports `signals/auth.ts`, and with it PostHog, Clerk, the browser Sentry surface, `auth-retry`, `connection-diagnostics` and the storage signals. The worker bundle drops **597 kB -> 452 kB**. `src/env.ts` reads the Vitest marker through `globalThis` instead of `window`.

3. **Gate it.** `scripts/worker-dom-globals.ts` adds a worker-build plugin that fails when a first-party module in the worker graph uses `document`, `window`, `history`, `localStorage` or `sessionStorage`. `typeof` feature detection and static property names are excluded; third-party modules are out of scope because they legitimately probe `window` before using it. Local bindings that shadow one of the names are reported too, since the check has no scope analysis and a worker-reachable module has no reason to reuse the name — that is why `packages/api-contracts/src/contracts/chat-threads.ts` renames a local `document` variable.

## Verification

- Worker asset built from this branch evaluates cleanly in a DOM-less sandbox; the same harness reproduces the `ReferenceError` on the deployed production asset.
- Both guards were mutation-tested: pointing the production branch at the preview key fails the bootstrap parity test on `app.vm0.ai` and `app.okou.ai`, and reintroducing a `document` read in `platform-host.ts` fails the worker build naming that file.
- `pnpm -F @okouai/app lint`, `check-types`, `lint:build-config` (7 tests, 2 new), `pnpm knip --workspace apps/platform`, and `pnpm -F @okouai/api-contracts check-types`.
- `vitest run src/__tests__/clerk-entrypoint.test.ts src/__tests__/platform-runtime-environment.test.ts src/signals/__tests__/auth-url.test.ts src/signals/external/chat-idb-opener.test.ts src/signals/external/chat-idb-schema.test.ts` — 44 passed.


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: accb868138d7f6f163e3b69ed0d28b94685316c3
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Fixes the production SharedWorker crash (PLATFORM-9T/9V/9W) introduced by #30232, which moved the Clerk publishable key from a build-time `import.meta.env` constant to `document.documentElement.getAttribute()` inside `resolvePlatformRuntimeConfig()` — a function reached at module scope by `lib/posthog.ts`, which the shared database worker imports through `worker-runtime.ts -> chat-idb-store.ts -> signals/auth.ts`. The PR restores hostname-based selection between the two inlined keys so `resolvePlatformRuntimeConfig()` depends only on `location` and build-time constants, splits the worker-safe IndexedDB opener into `chat-idb-opener.ts` so the worker no longer pulls in the browser auth/PostHog/Clerk/Sentry-browser graph, and adds a worker-build plugin that rejects first-party modules using window-only globals.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no wrong assumptions or unhandled paths | Pass | `platform-host.ts` selection matches the `index.html` predicate exactly (`vm0.ai`/`.vm0.ai` + `isOkouProductionHostname`); `localhost` maps to `development` in both, which selects the preview key on both sides. |
| `REVIEW.md` | Security: no secrets, no missing auth checks | Pass | Only publishable (public) Clerk keys, already inlined in `index.html` on `main`. No auth path changed. |
| `REVIEW.md` | Type safety: no `any`, no overly broad types | Pass | No `any`. Two narrow casts in `scripts/worker-dom-globals.ts` (`as unknown as AstNode`, `as readonly string[]`) bridge acorn's `Program` to the local structural node type; see non-blocking note. |
| `REVIEW.md` | Interface changes documented; breaking changes highlighted | Pass | `ChatIdbOpenerOptions.reload` becomes required; the only two callers (`chat-idb-store.ts`, `worker-runtime.ts`) both pass it, and the change is compile-enforced. |
| `REVIEW.md` | DB/JSONB review gate | N/A | `api-contracts/chat-threads.ts` only renames two local variables inside `resolveChatEventRecommendedFollowups` / `serializeChatFollowupsContent`. No schema, Zod, `jsonb().$type<>()`, or persisted shape change. |
| `REVIEW.md` | Feature regression guard: replaced behavior has explicit coverage | Pass | Key-resolution coverage went from 4 to 6 configs in `platform-runtime-environment.test.ts` and gained the bootstrap parity assertion in `clerk-entrypoint.test.ts`; `chat-idb-opener.test.ts` keeps the opener assertions through the rename; `chat-idb-schema.test.ts` still covers `chatIdb$` on the page path. |
| `REVIEW.md` | `fix:` commits must carry a regression test | Pass | `clerk-entrypoint.test.ts` now fails if the DOM read is reintroduced: the bootstrap no longer writes `data-vm0-clerk-publishable-key`, so a `document.documentElement.getAttribute` resolver would throw instead of returning the expected key. |
| `docs/bad-smell.md` | Zero `any`; zero lint/type suppressions | Pass | No `any`, no `eslint-disable`/`@ts-ignore`/`@ts-expect-error` anywhere in the diff. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | No `import()` added; the platform ESLint config already bans `ImportExpression` under `src/**`. |
| `docs/bad-smell.md` | No hardcoded config; fail fast on missing configuration | Pass | `requiredBuildValue(import.meta.env.VITE_CLERK_PUBLISHABLE_KEY_*)` throws instead of defaulting, and `scripts/clerk-html.ts:requiredEnvironmentValue` already fails the build in `configResolved` when either key is absent, so no artifact can ship without both. |
| `docs/bad-smell.md` | No duplicated logic that should be shared | Pass | The hostname predicate is now duplicated between the `index.html` inline script and `platform-host.ts`. Unavoidable — the bootstrap runs before any module loads and cannot import from the bundle — and the diff pins the two together with a parity test rather than leaving them to drift. |
| `docs/fallback.md` | Default is no fallback; no `??`/`||` on contract-guaranteed state | Pass | The diff introduces no fallback. `options.openDatabase ?? openDB` is pre-existing code moved verbatim; `counts.get(name) ?? 0` is a local accumulator, not a contract fallback. No `Fallbacks` section is required. |
| `docs/fallback.md` | Removing a fallback needs evidence and deletes branch + tests together | Pass | The DOM-read path, its `CLERK_PUBLISHABLE_KEY_ATTRIBUTE` constant, the `index.html` writer, and all four test stubs/assertions are removed in the same diff. Evidence that the replacement cannot silently fail: the build already hard-fails without both env vars (`scripts/clerk-html.ts`). |
| `docs/fallback.md` | No negative/tombstone tests asserting removed behavior stays removed | Pass | The `data-vm0-clerk-publishable-key` assertion is replaced by the parity assertion, not kept alongside it — the update-in-place pattern the doc asks for. |
| `docs/deployment-compatibility.md` | Frontend rollout: old loaded pages must keep working | Pass | `index.html` and the bundle ship as one pair (HTML references hashed, immutable asset URLs), so a client never mixes a new bootstrap with an old bundle. `public/sw.js` registers no `fetch` handler, which the doc also states, so the service worker cannot serve a stale HTML/bundle combination. |
| `docs/deployment-compatibility.md` | Old/new pairing across API, runner, and persisted state | N/A | No API contract, runner protocol, queue payload, migration, or persisted shape changes. `CHAT_IDB_VERSION` is untouched, so old and new tabs open the same IndexedDB version. |
| `docs/testing.md` | Integration tests preferred; unit tests only for the narrow exceptions | Pass | The two new tests target `domGlobalUsageCounts`, an AST walker with non-obvious invariants (typeof exclusion, static property positions, object shorthand) — the "algorithmically complex code" exception — and live in `scripts/check-single-bundle.node.ts`, which already tests `applicationBundleViolations` the same way under `pnpm lint:build-config`. |
| `docs/testing.md` | Mock at the boundary; no MSW/fetch/db/fake-timer violations | Pass | No `vi.mock()` with a relative path, no `vi.stubGlobal("fetch", …)`, no `vi.useFakeTimers()`, no db mocking added. The only new test doubles are `vi.stubEnv` for the two build-time keys. |
| `docs/testing/app-testing.md` | No parser/helper/static-config unit tests under `turbo/apps/platform` Vitest; signal bootstrap tests only when there is no page surface | Pass | No new Vitest test is added under `src/`; the new tests run through `node --test`/`tsx --test` as build tooling. `clerk-entrypoint.test.ts` remains an entry-point test for behavior that has no page surface. |
| `docs/testing/anti-patterns.md` | AP-1 no `toHaveBeenCalled()` as sole assertion; AP-2 no direct fetch mocking | Pass | New assertions compare returned values (`domGlobalUsageCounts` entries, `workerDomGlobalsMessage` content, resolved publishable key). `chat-idb-opener.test.ts` is a pure rename with unchanged assertions. |
| `docs/testing/testing-external-behavior.md` | Construct and observe through the external surface | Pass | `clerk-entrypoint.test.ts` still drives the real `index.html` bootstrap and keeps the external `script.dataset.clerkPublishableKey` assertion. The added `resolvePlatformRuntimeConfig()` call is the internal half of a parity invariant that spans the HTML entry point and the module graph and has no page-visible surface, so there is no more external vantage point available for it. |
| `docs/effect.md` | Render stays pure; no render-phase Store writes or DOM/global mutation | Pass | No React component changes. `env.ts` reads `globalThis.__vitest_index__` at module scope; it does not mutate a global. |
| `docs/cache.md` | RC-3 no unbounded module-lifetime caches; RC-4 owner/lifetime match | Pass | The only new `Map` in `domGlobalUsageCounts` is function-local and dies with the call. `FORBIDDEN_WORKER_GLOBALS` is a frozen five-element `as const` array. `defaultChatIdbOpener` stays a module singleton with the same lifetime it had before the split. |
| `.claude/skills/ccstate/SKILL.md` | ccstate usage: no `void`, `detach()` only in views, correct signal ownership | Pass | `chat-idb-store.ts` keeps its existing `computed` for `chatIdb$` with unchanged dependencies; no new signals, commands, `detach()`, or `void` are introduced anywhere in the diff. |
| `.claude/skills/ccstate/SKILL.md` | New build gate must not weaken worker-side signal wiring | Pass | `worker-runtime.ts` only swaps the import source; `createChatIdbOpener` keeps the same shape and the worker keeps passing its own `reload` that calls `notifyReloadRequired` instead of touching `window`. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

- `turbo/apps/platform/scripts/check-single-bundle.node.ts` / `turbo/apps/platform/vite.config.ts`: the new build gate's wiring and failure path are untested. The one test in this file that runs a real `build()` ("emits the fixed page topology and one external worker") still passes only `worker: { plugins: () => [singleWorkerJavaScriptBundlePlugin()] }`, so nothing asserts that (a) `workerDomGlobalsPlugin()` is registered in the worker build — a later edit to `vite.config.ts` could drop it silently, (b) the plugin actually fails the build for a first-party worker module that reads `document`, or (c) `isFirstPartyModule` still lets third-party feature detection through. The two new tests cover detection on synthetic strings only. Please extend that fixture build to include `workerDomGlobalsPlugin()` and assert both the rejection and the `node_modules` exemption. Not raised as P0 because the regression itself is covered (see the `fix:` row above) and the gate does fail a real `pnpm build`, which `deploy-app` runs on every PR — but this gate is the whole point of the change, so it should not be the one thing with no test of its own.

#### Non-blocking notes

- `scripts/worker-dom-globals.ts:domGlobalUsageCounts`: `parse()` throws a raw acorn `SyntaxError` if any first-party module in the worker graph is not parseable as ESM after the TypeScript transform. The build would then fail with an opaque parse error attributed to `platform-worker-dom-globals` rather than the intended message. Low likelihood today (every first-party platform source is ESM), but wrapping the parse failure with the module id would make it self-diagnosing.
- `scripts/worker-dom-globals.ts:isFirstPartyModule`: `const [filePath = ""] = moduleId.split("?")` defaults a value `String.prototype.split` can never omit. It is required by `noUncheckedIndexedAccess` rather than by real behavior, so it is not `docs/fallback.md` slop, but `moduleId.split("?")[0] ?? moduleId` or a `slice` would read closer to intent.

### Testing

- Coverage: Adequate. The regression that broke production is covered — reintroducing the `document.documentElement` read fails `clerk-entrypoint.test.ts`, because `index.html` no longer writes the attribute the resolver read. Key-selection coverage grew from 4 to 6 configurations, and the new bootstrap parity assertion closes the duplication risk this design creates. The gap is the build gate's own wiring, raised as P1 above.
- Conventions: Compliant. No relative-path `vi.mock()`, no direct `fetch` mocking, no fake timers, no db mocking, no `toHaveBeenCalled()`-only assertions, and no tombstone tests. The new build-tooling tests match the established pattern in `check-single-bundle.node.ts` and stay outside the platform Vitest surface that `docs/testing/app-testing.md` restricts.

Issue #30250, opened 2026-08-29
fix: make run timeout an atomic terminal boundary


seven332:
## Problem and evidence

Parent: #30246

Production run `6e534bf2-d83a-443e-82b0-25741d863172` remained blocked after API timeout because an open active-input delivery stayed reserved until late Runner completion. Timeout currently lacks one checkpoint/thread/run/delivery terminal boundary, and completion can overwrite timeout and replay terminal effects.

Prerequisites #30248 and #30249 are merged and confirmed in API production. Runtime checkpoint, storage, firewall, and other privileged mutation authority therefore closes after timeout before this work releases delivery.

The Guest currently performs one final checkpoint operation after CLI exit and then reports completion. A successful checkpoint is mandatory: if checkpoint persistence fails, the Guest converts the result to failure and the Runner's existing completion reports that failure. The Runner independently reports completion for every claimed run; a preceding Guest completion makes the Runner request an idempotent duplicate.

The split Guest requests create an unnecessary persisted interval in which a checkpoint can change `agent_sessions.conversation_id` before the run's terminal owner is selected. That interval is the root of the candidate-marker and promotion state needed by the superseded plan.

## Decision

Keep the current Guest-first and Runner completion behavior, but combine the Guest's final checkpoint commit with its completion request. The combined API transaction owns checkpoint persistence, canonical AgentSession promotion, active-input settlement, and the run terminal transition together.

Do not add a completion manifest, Guest-to-Runner acknowledgement, promotion epoch, or `agent_session_promotion_pending` database marker. The Runner keeps its existing completion payload and call path unchanged.

Large history and artifact data remain prepared and uploaded before the combined request. The completion body carries only the existing checkpoint metadata and uploaded-object references.

### Combined Guest completion

Add an optional checkpoint object to the completion request using the existing checkpoint fields. When present, the API validates and persists the conversation/checkpoint inside the same lock-ordered transaction that settles completion. Successful completion promotes that exact checkpoint to the owning AgentSession in the transaction. Generic failed/cancelled recovery keeps its current resumability policy; Pi keeps its narrower recovery policy.

[...]

Issue #30246, opened 2026-08-29
bug: make run timeout a consistent authoritative lifecycle boundary


seven332:
## Background and evidence

Production run `6e534bf2-d83a-443e-82b0-25741d863172` exposed a thread recovery failure after the API marked the run timed out. One follow-up prompt remained reserved in an open `active_input_delivery`, so terminal queue drain could not advance the FIFO queue. The old Guest continued receiving successful heartbeat responses, and the thread recovered only after late Runner completion approximately 17 minutes later.

The production timeline was:

- run start on `prod-4.aws.vm3.ai`: `2026-08-28T14:58:52Z`;
- active-input acknowledgement became uncertain: `15:01:10Z`;
- API timeout committed with the delivery still open: approximately `15:02:22Z`;
- Guest heartbeat still received HTTP 200 through `15:06:32Z`;
- no hard-cancel notification was sent;
- Guest shutdown: `15:10:20Z`;
- Runner completion: `15:19:51Z`;
- successor publication: `15:19:52Z`.

All nine queued prompts remained durable and ordered. The failure was availability and ownership: the API exposed a terminal failure while thread recovery still depended on the consumer it had declared timed out.

## Problem

The current lifecycle gives `timeout` conflicting meanings.

The cleanup transaction changes the run to `timeout`, records `completedAt` and an error, deletes the Runner job row, and dispatches failed terminal behavior: callbacks/chat failure, goal pause, queue drain, usage work, admission closure, and terminal cleanup.

At the same time:

- `completeAgentRun$` explicitly accepts `timeout` and can overwrite it with `completed` or `failed`, replace terminal fields, finalize delivery and session state, and re-enter terminal effects;
- timeout does not settle an open active-input delivery using the established thread/run/delivery locks;
- timeout does not publish Runner cancellation or revoke the three-hour sandbox token;
- heartbeat, events, generic checkpoints, storage writeback, and firewall auth can continue changing state after timeout;
- usage, model observations, and telemetry are mixed into the same authenticated surface even though they are accounting/diagnostic reports rather than execution capabilities.

The ordinary running cutoff is two minutes and is evaluated by a one-minute cron. Guest failure detection can take longer, and a late successful heartbeat resets the Guest failure count. API timeout therefore cannot be treated as proof of Guest quiescence.

Pending and queued timeouts share the status, but they have no claimed Guest or active-input delivery. Their current cleanup behavior must remain compatible.

## Decision

[...]

PR #30324, opened 2026-08-30
fix(api): stop sandbox authority after run timeout

### LINKED ISSUE #30249: fix: stop privileged sandbox mutations after run timeout
## Problem and evidence

Parent: #30246

#30246 makes an API-committed timeout an authoritative terminal boundary. Before timeout can immediately release active input and launch a successor, the old sandbox must lose authority to create storage state, move a storage head, or receive a new connector/model-provider credential grant.

The prerequisite runtime gate shipped in #30252, but privileged paths remain open on current `main`:

- sandbox-authenticated storage prepare and commit resolve the run and its writeback mount without requiring `pending` or `running` status;
- firewall auth resolves any matching run row without checking status, then may decrypt credentials, perform provider refresh, persist valid shared refresh state, and return credentials;
- Guest artifact snapshots run before the final checkpoint API call, so rejecting a late checkpoint does not prevent a preceding storage prepare or commit.

A timed-out Guest can therefore obtain upload capability, create a storage version, move the storage head, or receive a new credential response after the API has exposed terminal failure.

Firewall auth has a process-local mitm-addon cache keyed by run. A cache hit can avoid the API entirely, and non-expiring credentials may remain cached until registry ownership is removed. This slice ends new API credential grants after timeout; it does not revoke credentials already granted before the boundary or prove Guest quiescence. The final hard-cancellation and terminal-recovery slice in #30250 remains responsible for accelerating sandbox shutdown.

## Decision

End new privileged API authority at timeout while preserving only provably read-only, already-committed retries and the existing externally managed reference error boundaries.

The active authority states for sandbox capability calls are `pending` and `running`. A storage mutation or firewall credential response whose final admission wins before timeout may finish. If timeout wins the serialized final decision, the request must grant no new capability, credential, version, head movement, or metadata write.

Use the existing `agent_runs` row as the single serialization object for timeout and final privileged API admission.

- Perform an early active-state check before S3, decryption, credential resolution, or provider work.
- Keep all external operations outside database transactions.
- At the final capability boundary, open a short transaction and lock the exact run row with `FOR UPDATE`.
- Timeout cleanup already competes for this row through its conditional status update. If timeout commits first, the endpoint observes terminal state and grants nothing. If the endpoint admission commits first, that admitted operation may finish before timeout.
- Do not introduce a second advisory-lock domain. The checkpoint advisory lock is not used by timeout today, while row locking already supplies the required common ordering boundary.

For sandbox storage, the final run lock must be in the transaction that creates or updates the version, changes version metadata, moves the storage head/aggregate fields, and records applicable lineage. S3 verification remains outside. A terminal retry may return `200` only when the persisted version, storage head and aggregates, request metadata, and matching storage/version/parent/run lineage prove the exact operation already committed with no write required.

For firewall auth, all successful static, custom, model-provider, billable, and refreshed paths converge before a final short run-row lock check. Use existing `403/FORBIDDEN` when terminal. A valid shared provider refresh may remain committed when timeout wins the final check, but the timed-out run receives no credential.

No schema migration, token epoch, operation-id protocol, production contract change, Runner cache invalidation, or Guest/Runner code change is required.

### Storage writeback

- Gate every sandbox-authenticated storage prepare and commit path on active run authority without changing user-authenticated storage operations.
- Revalidate the exact run identity and mounted writeback storage at final admission.
- Perform the final status decision in the same database transaction/order that creates a storage version, changes version metadata, records lineage, or moves the storage head.
- Do not hold a database transaction open across S3 verification, manifest download, or presigned URL generation.
- If storage prepare/commit wins admission first, its admitted operation may finish before timeout.
- If timeout wins first, do not return upload capability, create or update a version, move a storage head, update aggregates, or append lineage.
- Return idempotent success after timeout only when the current request and persisted state prove the requested operation is already committed and the response performs no database or object-storage mutation.
- Require matching run/parent/version lineage plus complete version/head/metadata equality for that proof. If proof is absent, reject fail-closed rather than adding speculative schema or treating content deduplication alone as operation ownership.

### Firewall authentication

- Check active run authority before decrypting credentials, resolving credential-backed material, or starting a provider refresh.
- Recheck active authority before every successful credential response, including static, custom connector, model-provider, billable, and refreshed credential paths.
- Treat the final active row-lock check as the admission winner: if timeout commits first, return no credential; if the final check commits first, that admitted response may finish.
- Do not hold a database transaction or run-row lock across OAuth or other provider network calls.
- A refresh that finishes before a timeout-winning final check may retain valid shared credential state, but its credential material must not be returned to the timed-out Guest.
- Preserve the distinction between external reference absence, provider failures, database failures, malformed input, and violated local invariants. An unavailable external reference grants no credential; unrelated failures must not be flattened into timeout or unavailable.
- Do not claim to revoke a mitm-addon credential cache entry that does not call the API. Existing cached grants and workload quiescence remain outside this API-only slice.

### Compatibility

- Reuse storage `404` and firewall `403/FORBIDDEN`, already represented by the canonical contracts.
- Preserve old Guest and mitm-addon behavior without a new response field, token epoch, or protocol version.
- Keep request and success schemas unchanged across independent API, Runner, and Guest deployment.

### Rely on checkpoint rejection

Artifact snapshots are prepared and committed before the final checkpoint request, so the checkpoint gate is too late to protect storage state.

### Rely only on hard Runner cancellation

Cancellation is best effort and is intentionally not the correctness boundary. An unavailable or delayed Runner must not leave privileged API mutations open after timeout.

### Share the checkpoint advisory lock with timeout and capabilities

This can serialize lifecycle operations, but it broadens the change into timeout/checkpoint/complete locking and creates a second lock-order domain alongside the run row. It offers no benefit over the row lock already acquired by timeout's update for these two endpoints.

### Add a persisted capability epoch and storage operation id

This would make universal revocation and idempotency explicit, but requires a migration, token/request changes, generated clients, and a mixed-version rollout. It still cannot revoke a prior Runner cache hit without an additional cache protocol. That cost is not justified for the current endpoint-specific boundary.

### Revoke the sandbox token globally

Global revocation would also block permitted late accounting, telemetry, and completion acknowledgement, and would require new persisted revocation/protocol state. Endpoint-specific admission keeps those reporting paths compatible.

### Reject every terminal storage retry

This guarantees zero writes but loses a safe acknowledgement when current persisted state and matching lineage prove the exact operation already committed. The chosen direction keeps that narrow read-only case and fails closed otherwise.

### Treat every existing content-addressed version as an exact retry

Content equality proves version identity, not that this run committed the requested head or metadata transition. Matching lineage and complete requested state are required.

## Risks and mitigations

- **Storage check/write race:** re-read authoritative storage/version state under the final transaction and lock the run before storage mutation rows.
- **Long transactions:** keep S3, manifest, presign, decryption, and provider calls outside final admission transactions.
- **Lineage atomicity:** moving applicable lineage into the storage transaction means a lineage database failure rolls back the commit instead of leaving an acknowledged partial authority record. Do not add manual savepoint machinery.
- **Credential resolution race:** apply a final active check to every successful response, not only refresh paths.
- **Shared refresh state:** do not roll back a valid provider refresh merely because the requesting run timed out; reject credential disclosure while preserving shared state.
- **Cached credential overclaim:** API gating cannot revoke mitm-addon cache hits. Retain #30250 hard cancellation as the shutdown path.
- **Over-broad gating:** change sandbox capability paths only. Preserve user storage operations, reporting webhooks, and established external-reference classifications.
- **Mixed deployment:** keep request and success schemas unchanged and use existing permanent errors so old Guest and mitm-addon versions fail closed.
- **Test timing:** use deferred external boundaries and scoped cleanup for exact run ids; do not use sleeps or global cleanup.

## Acceptance criteria

- Timeout before storage prepare returns `404` and exposes no presigned upload capability.
- Prepare admitted before timeout may return upload capability, but timeout before commit creates or updates no version, moves no head, changes no storage aggregate, and records no lineage.
- A storage commit whose serialized mutation decision wins first remains committed.
- A post-timeout exact retry receives idempotent success only when matching lineage and complete requested state prove the prior commit; handling performs no write, including metadata or lineage updates.
- Active user/storage-only prepare and commit behavior is unchanged.
- Timeout before firewall auth returns `403/FORBIDDEN` before secret decryption or provider refresh that depends on run authority.
- Timeout winning during any firewall resolution path, including static and refreshed credentials, returns no credential.
- A provider refresh completed before the final check may preserve valid shared state even when timeout-winning final admission rejects the response.
- Firewall auth performs no database transaction or run-row lock across a remote provider request.
- Database, provider, malformed input, local invariant, and externally managed reference failures retain their existing classifications.
- `pending` and `running` sandbox capability behavior remains accepted; touched `completed`, `failed`, `cancelled`, and `timeout` behavior is integration-tested.
- Existing old Guest and mitm-addon clients handle the selected permanent responses without a protocol change.

## Delivery notes

This remains one independently reviewable API PR. It changes the two capability services plus route-level integration tests and, only where production APIs cannot expose persisted facts, narrow test-only state readers.

Relevant files:

- `turbo/apps/api/src/signals/services/storage-write.service.ts`
- `turbo/apps/api/src/signals/services/agent-webhook-firewall-auth.service.ts`
- `turbo/apps/api/src/signals/routes/__tests__/webhooks-callbacks.bdd.test.ts`
- `turbo/apps/api/src/signals/routes/__tests__/webhooks-agent-firewall-auth.bdd.test.ts`
- existing scoped timeout/test-storage fixture routes and test-only contracts as needed

This is the second delivery slice of #30246. Its prerequisite #30248 is closed by #30252. It blocks #30250, which owns atomic active-input settlement, immutable late completion, hard Runner cancellation, and terminal queue recovery.

No Runner/Guest code change, token revocation, production contract change, database migration, active-input settlement, completion transition, or cache-invalidation protocol is in scope.

seven332:
## Summary

- gate sandbox storage prepare/commit and firewall credential responses to `pending` or `running` runs
- serialize the final privileged admission against timeout through the exact `agent_runs` row lock
- keep S3 verification outside the storage mutation transaction and perform firewall final admission after KMS/provider resolution without holding the run-row lock across external work
- make storage version, head, aggregate, and lineage writes atomic, including deduplicated HEAD moves to existing versions
- reject unproven terminal storage callbacks before S3; preserve a narrow write-free retry only when persisted metadata, head state, and run lineage prove the exact operation already committed

- reuses existing storage `404` and firewall `403/FORBIDDEN` responses; request and success contracts are unchanged
- does not revoke credentials already served from the Runner/mitm process-local cache
- does not change Guest or Runner behavior, token formats, schemas, active-input settlement, or hard sandbox cancellation; those remain in #30250

## Validation

- `pnpm format`
- `pnpm -F api lint`
- `pnpm check-types` (pre-commit hook, all Turbo workspaces)
- `pnpm knip`
- `pnpm -F api exec vitest run src/signals/routes/__tests__/webhooks-callbacks.bdd.test.ts --silent=passed-only` (63 passed, including no-S3 terminal rejection and deduplicated exact retry)
- `pnpm -F api exec vitest run src/signals/routes/__tests__/webhooks-agent-firewall-auth.bdd.test.ts --silent=passed-only` (60 passed)
- `pnpm -F api exec vitest run src/signals/routes/__tests__/run-lifecycle.bdd.test.ts -t "keeps a committed artifact head after initial empty artifact creation" --silent=passed-only` (1 passed)

The full `pnpm vitest run --maxWorkers=4 --silent=passed-only` run passed both affected API suites and 715 other files. Three unrelated files failed locally: one CLI suite loaded developer API-origin overrides before its test stubs (72/72 passed when those overrides were unset), and two API suites encountered pre-existing shared test-database fixture collisions.

Closes #30249

Parent: #30246

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/helpers/api-bdd-run-timeout.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/helpers/api-bdd-run-timeout.ts
@@ -1,0 +1,33 @@
+import { testCronCleanupSandboxesStateContract } from "@okouai/api-contracts/contracts/test-cron-cleanup-sandboxes-state";
+
+import { accept, type TestContext } from "../../../../__tests__/test-context";
+import { setupApp } from "../../../../__tests__/test-helpers";
+import { testCronCleanupSandboxesStateRoutes } from "../../test-cron-cleanup-sandboxes-state";
+
+export type TestTerminalRunStatus =
+  | "completed"
+  | "failed"
+  | "cancelled"
+  | "timeout";
+
+export async function transitionRunToTerminal(
+  context: TestContext,
+  runId: string,
+  status: TestTerminalRunStatus,
+) {
+  return await accept(
+    setupApp({ context, routes: testCronCleanupSandboxesStateRoutes })(
+      testCronCleanupSandboxesStateContract,
+    ).action({
+      body: { action: "transition-run-terminal", run_id: runId, status },
+    }),
+    [200],
+  );
+}
+
+export async function transitionRunToTimeout(
+  context: TestContext,
+  runId: string,
+) {
+  return await transitionRunToTerminal(context, runId, "timeout");
+}
--- a/turbo/apps/api/src/signals/routes/__tests__/helpers/api-bdd-storages.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/helpers/api-bdd-storages.ts
@@ -38,6 +38,13 @@ interface BddStorageDownloadQuery {
   readonly version?: string;
 }
 
+interface BddStorageWritebackQuery {
+  readonly storageId: string;
+  readonly versionId: string;
+  readonly runId: string;
+  readonly parentVersionId: string;
+}
+
 function requireOrgId(actor: ApiTestUser): string {
   if (!actor.orgId) {
     throw new Error("Storage fixture requires an org-scoped actor");
@@ -168,6 +175,20 @@ export function createStoragesBddApi(context: TestContext) {
         throw new Error("Storage download action returned no result");
       }
       return response.download;
+    },
+
+    async inspectWriteback(query: BddStorageWritebackQuery) {
+      const response = await postAction(context, {
+        action: "inspect-writeback",
+        storageId: query.storageId,
+        versionId: query.versionId,
+        runId: query.runId,
+        parentVersionId: query.parentVersionId,
+      });
+      if (!response.writeback) {
+        throw new Error("Storage writeback inspection returned no result");
+      }
+      return response.writeback;
     },
   };
 }
--- a/turbo/apps/api/src/signals/routes/__tests__/webhooks-agent-firewall-auth.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/webhooks-agent-firewall-auth.bdd.test.ts
@@ -38,6 +38,11 @@ import {
   mockTestOAuthAuthCodeProvider,
 } from "./helpers/api-bdd-connectors";
 import { createRunsApi } from "./helpers/api-bdd-runs";
+import {
+  transitionRunToTerminal,
+  transitionRunToTimeout,
+  type TestTerminalRunStatus,
+} from "./helpers/api-bdd-run-timeout";
 import { createWebhookCallbackApi } from "./helpers/api-bdd-webhooks";
 import {
   setConnectorCredentialStorageState,
@@ -65,8 +70,14 @@ const TEST_DATA_KEY = Buffer.from("0123456789abcdef0123456789abcdef", "utf8");
  */
 
 const context = testContext();
+const TERMINAL_RUN_STATUSES = [
+  "completed",
+  "failed",
+  "cancelled",
+  "timeout",
+] as const satisfies readonly TestTerminalRunStatus[];
 
-async function firewallRun(): Promise<{
+async function firewallRun(existingActor?: ApiTestUser): Promise<{
   readonly actor: ApiTestUser;
   readonly runId: string;
   readonly headers: { readonly authorization: string };
@@ -74,7 +85,7 @@ async function firewallRun(): Promise<{
   const bdd = createBddApi(context);
   const runsApi = createRunsApi(context);
   const fw = createFirewallApi(context);
-  const actor = bdd.user();
+  const actor = existingActor ?? bdd.user();
   bdd.acceptAgentStorageWrites();
   runsApi.acceptStorageDownloads();
   runsApi.acceptTelemetryIngest();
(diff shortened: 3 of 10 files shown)


seven332:
## Code Review: PR #30324

**Reviewed head:** `423d4fc29addd974e2456b25ab12d120b1e9c15f`

### Summary

The API-only lifecycle boundary now uses the existing `agent_runs` row to serialize timeout against storage mutation/capability admission and successful firewall credential responses. External S3, KMS, and provider work stays outside the final transactions, and mixed-version request/response contracts are unchanged.

### Findings

No findings. The self-review finding that an already-terminal, uncommitted storage callback could still reach S3 verification was fixed in `423d4fc29a`; the route integration test now proves that rejection performs neither S3 requests nor persisted writes.

### Testing

Focused callback and firewall route suites pass (62 and 60 tests). Formatting, API/API-contract lint and type checks, full workspace type checks, and Knip pass. The earlier full local suite passed both affected suites and 715 other files; its three failures were independently reproduced as developer-env/shared-fixture issues. Current-head PR CI is still being evaluated separately.

### Verdict

LGTM

PR #30341, opened 2026-08-30
test(mitm-addon): cover nested decode boundary

### LINKED ISSUE #30209: test(python): cover the exact nested path decode limit
## Summary

Add a regression test for an unsafe path segment that becomes a dot segment on exactly the fifth and final permitted percent-decoding pass.

The current implementation is correct: after five decoding passes, it performs one final unsafe-syntax check before using an additional decode only to detect input that still changes beyond the work bound. The focused tests cover ordinary/double encoding and a value that still changes beyond the bound, but they do not execute the final unsafe-syntax branch.

## Evidence

Reviewed against `origin/main` at `8a5ae92cbba50d14508ca2b13616b87d6e71af3c`.

- [`_segment_has_unsafe_path`](https://github.com/vm0-ai/vm0/blob/8a5ae92cbba50d14508ca2b13616b87d6e71af3c/crates/runner/mitm-addon/src/path_security.py#L52-L69) allows five decode passes and then checks the resulting syntax at lines 65-66.
- [The dot-segment cases](https://github.com/vm0-ai/vm0/blob/8a5ae92cbba50d14508ca2b13616b87d6e71af3c/crates/runner/mitm-addon/tests/test_path_security.py#L11-L32) cover lower depths and `/api/%25252525252e%25252525252e/admin`, which remains encoded after the fifth pass and is rejected by the extra-change check instead.
- [The safe maximum-depth case](https://github.com/vm0-ai/vm0/blob/8a5ae92cbba50d14508ca2b13616b87d6e71af3c/crates/runner/mitm-addon/tests/test_path_security.py#L156-L164) confirms a safe character decoded at the limit, but there is no corresponding unsafe boundary case.
- Full branch coverage (`7194 passed`) left the final unsafe return at line 66 unexecuted.

A deterministic mutation probe removed only the post-loop unsafe-syntax check. All 15 existing dot-segment cases still passed, while the exact-limit path below changed from blocked to allowed:

```text
/api/%252525252e/admin
```

That segment becomes `.` on exactly the fifth decode pass.

## Impact

A future off-by-one or cleanup refactor could remove the post-loop guard without failing the current suite, allowing an encoded dot segment through platform API admission, firewall matching, and rewrite safety checks.

## Acceptance criteria

- Add a case where `.` or `..` materializes on exactly `_MAX_PERCENT_DECODE_PASSES` and assert it is unsafe.
- Keep the existing safe-character-at-the-limit assertion.
- Assert that input requiring one additional decode still fails closed because it exceeds the decoding work bound.
- Prefer a small parametrized boundary matrix derived from `_MAX_PERCENT_DECODE_PASSES` so lower, exact, and over-limit depths cannot drift independently.
- Keep the full mitm-addon Python test suite passing.

seven332:
## Summary

- replace the isolated over-limit dot-segment literal with a configured decode-boundary matrix
- cover unsafe materialization below, exactly at, and one pass beyond the nested decode limit
- preserve the existing safe-character-at-the-limit control

## Scope

This is a test-only change. It does not modify path validation behavior, expose new production APIs, change dependencies, or alter runtime compatibility.

## Testing

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_path_security.py` (78 passed)
- `uv run --no-sync python -m pytest tests/` (7223 passed)

Closes #30209

### DIFF
--- a/crates/runner/mitm-addon/tests/test_path_security.py
+++ b/crates/runner/mitm-addon/tests/test_path_security.py
@@ -39,7 +39,7 @@ def test_has_unsafe_path_blocks_dot_segments(path):
         pytest.param(path_security._MAX_PERCENT_DECODE_PASSES + 1, id="over-limit"),
     ],
 )
-def test_has_unsafe_path_blocks_nested_dot_segments_at_decode_limit(decode_passes):
+def test_has_unsafe_path_blocks_nested_dot_segments_at_decode_limit(decode_passes: int):
     encoded = "%2e"
     for _ in range(decode_passes - 1):
         encoded = urllib.parse.quote(encoded, safe="")
--- a/crates/runner/mitm-addon/tests/test_path_security.py
+++ b/crates/runner/mitm-addon/tests/test_path_security.py
@@ -25,11 +25,26 @@ import path_security
         "/api/%252e%252e%253bmatrix=1/admin",
         "/api/%252e%252e%252fadmin",
         "/api/%252f..",
-        "/api/%25252525252e%25252525252e/admin",
     ],
 )
 def test_has_unsafe_path_blocks_dot_segments(path):
     assert path_security.has_unsafe_path(path) is True
+
+
+@pytest.mark.parametrize(
+    "decode_passes",
+    [
+        pytest.param(path_security._MAX_PERCENT_DECODE_PASSES - 1, id="below-limit"),
+        pytest.param(path_security._MAX_PERCENT_DECODE_PASSES, id="at-limit"),
+        pytest.param(path_security._MAX_PERCENT_DECODE_PASSES + 1, id="over-limit"),
+    ],
+)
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #30341

**Reviewed head:** `e0b67a2059924fccbe7a476263c456eb39fc2adf`

### Summary

The PR replaces a standalone over-limit encoded dot literal with a configured three-point boundary matrix. It now exercises unsafe materialization below, exactly at, and beyond the percent-decoding work bound through public `has_unsafe_path` behavior while preserving the safe-at-limit control.

### Findings

No findings. The self-review's explicit-parameter-type issue was fixed on the submitted head.

### Testing

Uv lock/sync, Ruff format/lint, BasedPyright, the focused path-security suite (78 passed), and the complete addon suite (7223 passed) all pass. No runtime deployment or unrelated Rust/Turbo checks were run because this PR changes only Python test code.

### Verdict

LGTM

PR #30371, opened 2026-08-31
perf(python): avoid omitted connector union allocation

### LINKED ISSUE #30314: perf(python): avoid rebuilding omitted connector unions per request
## Problem

At current `main` (`e29edce18f87b336f5616afb438187b357021612`), every request classification with a present connector intent that reaches connector routing materializes the union of the VM's two omitted-connector sets before testing that one intent:

- `crates/runner/mitm-addon/src/request_classification.py:469-479`

```python
if intent.status == "present" and intent.value in (
    omitted_builtin_firewalls | omitted_custom_connector_ids
):
```

The union is a new `frozenset` whose construction and temporary storage scale with every omitted target, although membership in the union is equivalent to two constant-time membership checks. This code runs synchronously in mitmproxy request classification and can be reached during both `requestheaders` and `request` processing.

The omitted sets are bounded only by the overall registry file size, not by the 256-target synchronization batch size:

- `crates/runner/mitm-addon/src/registry.py:426-434` validates and converts each complete omitted list without a count limit.
- `turbo/packages/api-contracts/src/contracts/runners.ts:281-287` leaves the run-level target schema unbounded and limits only one sync request.
- `crates/runner/src/provider/connector_runtime_sync.rs:508-510,5773-5784` chunks synchronization and explicitly tests that the complete run target set is not limited.
- `crates/runner/src/proxy/registry.rs:37-42,450-489` persists the complete target and omitted sets; `crates/runner/src/state_file.rs:26` permits a registry up to 16 MiB.

A compact registry containing 50,000 Custom targets plus the same 50,000 omitted IDs serializes to approximately 5.53 MiB, so this scale is accepted by the current persistence bound. Large target sets are an explicit supported contract even though normal runs are much smaller.

## Measured evidence

On CPython 3.12.3, the exact membership shapes were benchmarked with one populated `frozenset`, one empty `frozenset`, and a bounded missing intent. Medians over nine samples were:

| Omitted values | Materialized union membership | Two direct memberships | Peak union allocation |
| ---: | ---: | ---: | ---: |
| 256 | 1.616 us | 0.104 us | 16,600 bytes |
| 10,000 | 197.237 us | 0.081 us | 524,504 bytes |
| 50,000 | 1,023.754 us | 0.088 us | 2,097,368 bytes |
| 100,000 | 2,093.346 us | 0.094 us | 4,194,520 bytes |

An unavailable connector that is retried repeatedly therefore creates target-count-proportional CPU and allocation pressure on the shared proxy path. At 50,000 omitted targets, one classification blocks for about 1 ms and allocates about 2 MiB solely to answer one membership query; header- and request-phase classification can multiply that work for one flow.

## Decision

Use short-circuit membership checks against the existing Builtin and Custom omitted sets. Keep the registry snapshot and persisted format unchanged.

This provides constant request-time work and temporary memory without retaining a third target-count-proportional collection. A Builtin match performs one lookup; a Custom match or no-match performs two.

## Alternatives considered

- Precomputing a combined set per registry snapshot would remove request-time union construction but retain another proportional hash table for the snapshot lifetime and broaden publication/invalidation state.
- Adding an omitted-intent lookup abstraction would not improve runtime behavior over the direct expression and would add a one-caller abstraction plus implementation-detail testing pressure.

## Test strategy and risks

- Use real registry files and real mitmproxy flows at `classify_request()` to cover Builtin-only, Custom-only, both-set, present no-match, malformed, and absent intent behavior.
- Guard the resource property with a warmed `tracemalloc` comparison between small and large real omitted sets. This avoids elapsed-time thresholds, source inspection, and internal mocks.
- Calibrate the fixture cardinality and constant allocation margin under the committed Python runtime with a wide old-versus-new signal to minimize allocation-test variability.

## Acceptance criteria

- Connector-intent classification does not materialize a union proportional to the omitted target count on each request.
- Lookup CPU remains constant-time and temporary memory remains constant with respect to the number of omitted targets.
- An intent found in either the Builtin or Custom omitted set still receives the current ordinary `allow` classification.
- Present intents absent from both sets, malformed intents, and absent intents retain current classification behavior.
- Focused integration coverage exercises Builtin-only, Custom-only, both-set, and no-match cases and guards against reintroducing per-request collection construction without relying on timing thresholds.
- The complete mitm-addon formatting, lint, type, and test checks pass.

## Compatibility

No API, queue payload, persisted registry shape, Rust/Python contract, migration, feature switch, or cross-version rollout behavior changes.

seven332:
## Summary

- replace per-request union construction with short-circuit membership in the existing Builtin and Custom omitted sets
- preserve ordinary allow behavior for omitted connector intents and downstream firewall behavior for other intent states
- add real-registry classifier integration coverage and a warmed allocation-scaling regression

## Scope

- no API, registry format, Rust/Python contract, migration, or feature-switch changes
- no combined omitted-target cache or retained index

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_request_classification_omitted_connectors.py -q` (7 passed)
- allocation regression repeated three additional times (passed)
- `uv run --no-sync python -m pytest tests/` (7234 passed)

Closes #30314

### DIFF
--- a/crates/runner/mitm-addon/src/request_classification.py
+++ b/crates/runner/mitm-addon/src/request_classification.py
@@ -474,8 +474,8 @@ def _classify_request(
         client_ip,
         frozenset(),
     )
-    if intent.status == "present" and intent.value in (
-        omitted_builtin_firewalls | omitted_custom_connector_ids
+    if intent.status == "present" and (
+        intent.value in omitted_builtin_firewalls or intent.value in omitted_custom_connector_ids
     ):
         return Allow(
             sandbox_info=sandbox_info,
--- a/crates/runner/mitm-addon/tests/test_request_classification_omitted_connectors.py
+++ b/crates/runner/mitm-addon/tests/test_request_classification_omitted_connectors.py
@@ -155,10 +155,11 @@ def test_omitted_connector_lookup_allocation_does_not_scale_with_target_count(
             )
             assert isinstance(warm_classification, request_classification.Allow)
 
+            measured_flow = new_flow()
             tracemalloc.start()
             try:
                 classification = request_classification.classify_request(
-                    new_flow(),
+                    measured_flow,
                     registry_path=str(registry_path),
                     api_url=_API_URL,
                     tls_admission=None,
--- a/crates/runner/mitm-addon/tests/test_request_classification_omitted_connectors.py
+++ b/crates/runner/mitm-addon/tests/test_request_classification_omitted_connectors.py
@@ -1,0 +1,172 @@
+"""Connector-intent classification for omitted runtime targets."""
+
+import tracemalloc
+
+import pytest
+
+import connector_intent
+import request_classification
+from tests.request_handler_helpers import (
+    _sandbox_without_firewalls,
+    _shared_route_sandbox,
+    _write_registry,
+)
+
+_API_URL = "https://api.vm0.ai"
+_CLIENT_IP = "10.200.0.5"
+_CONNECTOR_INTENT_HEADER = "X-VM0-Connector-Intent"
+
+
+@pytest.mark.parametrize(
+    ("omitted_builtin", "omitted_custom", "intent"),
+    [
+        (["removed-builtin"], [], "removed-builtin"),
+        ([], ["removed-custom"], "removed-custom"),
+        (["removed-overlap"], ["removed-overlap"], "removed-overlap"),
+    ],
+    ids=["builtin-only", "custom-only", "both"],
+)
+def test_omitted_connector_intent_returns_ordinary_allow(
+    tmp_path,
+    real_flow,
+    mitm_ctx,
+    headers,
+    omitted_builtin,
+    omitted_custom,
+    intent,
+):
+    sandbox = _shared_route_sandbox(tmp_path)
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #30371

**Reviewed head:** `dc96a3d9a1199dac6ca483310ac2cfed5e2c8593`

### Summary

The request classifier now checks the two existing omitted-target sets directly, preserving Boolean behavior while eliminating the per-request union allocation. The regression coverage uses real registry files, real mitmproxy flows, captured connector intent, and the public classifier entry point.

### Findings

No findings. A self-review observation about flow-construction noise in the allocation trace was resolved on this head by moving measured-flow construction before `tracemalloc.start()`.

### Testing

- The focused seven-case module covers Builtin-only, Custom-only, both-set, present no-match, malformed, absent, and small-versus-large allocation behavior.
- The allocation test was proven to fail against the old union and was repeated successfully after the fix.
- Ruff format/lint, basedpyright, lock/environment checks, and the complete mitm-addon suite passed (`7234 passed`; `43` pre-existing third-party warnings).
- GitHub CI remains separately pending/subject to the PR checks.

### Verdict

LGTM

PR #30376, opened 2026-08-31
test(runner): cover workspace cache metadata rejection

### LINKED ISSUE #30297: test(rust): cover workspace cache metadata rejection branches
## Location

- `crates/runner/src/workspace_image_cache/metadata.rs:218`
- `crates/runner/src/workspace_image_cache/tests/metadata.rs:124`

## Test gap

`validate_metadata` is the authoritative compatibility and reusability gate for persisted workspace image cache entries. Its direct mismatch test currently changes only `reuse_key`, so the later independent guards for `profile_name`, `working_dir`, `drive_layout`, `logical_image_size_bytes`, and non-`Current` cache states are not exercised. The repository has higher-level coverage for unknown format versions and runner-group scoping, but searches of the workspace cache tests found no scenario that corrupts these downstream fields and verifies rejection through the cache entry point.

Because the function returns on the first mismatch, the existing reuse-key case cannot protect any of the later branches. Removing, reordering, or inverting one of those checks could therefore pass the current test suite.

## Impact

A regression in this validation boundary could accept stale or incompatible metadata as a reusable workspace seed. That can cross profile, working-directory, drive-layout, image-size, or cache-state boundaries and feed an invalid image into sandbox startup instead of forcing a cache miss.

## Acceptance criteria

- Add focused Rust integration coverage through `read_valid_metadata` or `prepare`, starting from an otherwise valid committed cache entry.
- Independently corrupt `profile_name`, `working_dir`, `drive_layout`, and `logical_image_size_bytes`, and assert that each entry is rejected rather than returned or advertised as reusable.
- Cover both `WorkspaceCacheState::Dirty` and `WorkspaceCacheState::Invalid` with the same rejection guarantee.
- Assert the relevant error or checkout result for each case so every guard is independently protected.
- Keep the existing unknown-format, runner-group scope, current-image replacement, and symlink rejection scenarios passing.

seven332:
## Summary

- create each metadata mismatch case from a valid committed workspace cache entry
- cover profile, working directory, drive layout, logical size, dirty state, and invalid state rejection independently
- assert branch-specific validation errors while preserving the existing reuse-key coverage

## Exclusions

- no production workspace cache behavior or persisted metadata format changes
- no API, migration, feature switch, or deployment changes

## Validation

- `cargo test -p runner workspace_image_cache::tests::metadata`
- `cargo test -p runner`
- repository pre-commit hooks (Rust formatting, workspace documentation, workspace Clippy, file-size check)

Closes #30297

### DIFF
--- a/crates/runner/src/workspace_image_cache/tests/metadata.rs
+++ b/crates/runner/src/workspace_image_cache/tests/metadata.rs
@@ -9,7 +9,7 @@ use super::super::{
     WorkspaceCacheTerminalStatus, WorkspaceImageCache, WorkspaceImageLeaseIdentity,
     WorkspaceImagePrepareRequest,
 };
-use super::support::{TEST_PROFILE_NAME, local_cache};
+use super::support::{TEST_PROFILE_NAME, local_cache, write_current_cache_entry};
 use crate::ids::RunId;
 use crate::paths::RunnerPaths;
 use crate::storage_fingerprints::StorageFingerprints;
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #30376

**Reviewed head:** `6f28c67ed7d62707ea5b7c5852aff6dc87cb7f98`

### Summary

The PR adds isolated persisted-entry coverage for every requested workspace cache metadata rejection branch while keeping production behavior unchanged. Each named case mutates one otherwise valid field and verifies the intended validation error through `read_valid_metadata`.

### Findings

No findings.

### Testing

Focused metadata tests, rustfmt, Runner Clippy, Runner documentation, full Runner tests, guest inventory, and repository pre-commit Rust checks passed locally. Remote PR CI remains to be verified by the check stage.

### Verdict

LGTM

PR #30402, opened 2026-08-31
fix(api): inline hosted-site reference images for byteplus

bingjiez666:
## Problem

A reference image published to a hosted site cannot be used with any BytePlus
image model. It fails every time.

The API never sends the image itself. `bytePlusImageInput` puts the URL in the
request and BytePlus fetches it:

```ts
image: sourceImages,   // the caller's URL, handed over as-is
```

Hosted sites sit behind a stricter Cloudflare policy than the artifact CDN.
This repo already depends on that being true: `artifact-preview.service.ts`
carries a `vm0_artifact_preview` WAF bypass cookie and has an explicit
`isCloudflareChallenge()` branch just to screenshot one. A third-party fetcher
has no such escape hatch.

Verified against production, reproduced on demand:

| Fetcher | Target | Result |
| --- | --- | --- |
| BytePlus | `*.sites.vm0.io` asset | 403 |
| BytePlus | `cdn.vm0.io/artifacts/…` | 200 |
| This sandbox, same client and headers | `*.sites.vm0.io` asset | 200 |

Four production generations failed this way in the last three months, each
reported by BytePlus as `Error while downloading: …, status code: 403`. The
artifact CDN path is unaffected, which is why this went unnoticed.

## Change

Send the bytes instead of a URL the provider is blocked from fetching.
BytePlus accepts each reference image as either an accessible URL or an inline
`data:image/<format>;base64,…` payload, so hosted-site URLs are downloaded and
encoded before the request body is built.

**Only hosted-site URLs are swapped.** Every other URL is still passed through
untouched. That keeps the provider's own fetch path in use for the common case,
and — more importantly — it means arbitrary caller-supplied URLs are never
turned into server-side fetches, so this does not open an SSRF surface. The
host check reuses the same `ZERO_HOST_DOMAIN` / `OKOU_PUBLIC_HOST_DOMAIN`
suffix test that `renderArtifactSnapshot` already applies.

Downloads are bounded by BytePlus's documented 30 MB per-image limit, checked
against `content-length` before buffering so an oversized object cannot be
pulled into memory just to be rejected afterwards, with a post-read check as
the backstop when the header is absent.

## Feature switch

`inlineHostedSiteReferenceImages`, shipped `enabled: true`.

The switch is not staging a rollout — the affected requests fail 100% of the
time today, so there is nothing to stage. It exists so this can be turned off
without a release if the added download proves costly in production. Removing
it is a separate PR.

## Scope

Image generation only. `video-generation.service.ts` passes reference,
first-frame, and last-frame images to the same provider the same way and has
the same defect, but it has separate plumbing and belongs in its own PR. I have
not seen it hit in production yet.

## Tests

Three integration tests at the route boundary:

- a hosted-site reference image is fetched once and reaches BytePlus as
  `data:image/jpeg;base64,…`
- a non-hosted URL reaches BytePlus unchanged, and the API does not fetch it
- an unreadable hosted-site image fails with
  `REFERENCE_IMAGE_DOWNLOAD_FAILED` and never calls the provider or charges
  credits


bingjiez666:
Superseded by #30425, which fixes the same problem better.

`provider-reference-url.service.ts` rewrites first-party reference URLs into presigned R2 URLs before the provider request is built. That reaches the same bytes without going through Cloudflare at all, so the block this PR worked around never applies.

It is strictly broader than this PR:

- covers both the fal and BytePlus paths, not just BytePlus
- covers `maskImageUrl` as well as source images
- verifies ownership by org, deployment status, and manifest entry, rather than only matching a host suffix
- needs no server-side download, so no size cap and no feature switch

It also makes this PR's code unreachable: #30425 resolves the URLs before `generateBytePlusImage` is called, so `isHostedSiteImageUrl()` would never match a hosted-site host again.

Closing unmerged. No part of this branch is worth salvaging.

PR #30430, opened 2026-08-31
fix(runner): reject cleartext remote api urls

### LINKED ISSUE #30199: security(rust): reject cleartext remote runner API URLs
## Problem and evidence

The runner accepts plain HTTP for an arbitrary remote control-plane API host and then uses that base URL for authenticated requests carrying the runner Bearer token and, when configured, the Vercel automation bypass secret.

- `crates/runner/src/config.rs` accepts both `http` and `https` in the shared `normalize_api_base_url` boundary.
- `crates/runner/src/config/tests.rs` explicitly accepts `http://192.0.2.10:8080/base`, demonstrating that cleartext transport is not limited to local development.
- `crates/runner/src/http.rs` constructs authenticated platform requests from the normalized base URL.
- The same normalizer is reached by `runner config`, config loading, `runner start` overrides, and `HttpClient::new()`.

A passive network observer or intermediary can recover the credential or modify runner control traffic when an operator accidentally configures a remote HTTP endpoint. This is a fail-closed transport invariant, not protection against a fully compromised runner host or operator: an attacker who can replace the configured authority could also point the runner at an attacker-controlled HTTPS endpoint.

This issue supersedes #18157 with a narrower Rust runner scope. Since that issue was closed, the runner has gained a shared API URL normalization boundary, so the required policy can now be enforced without duplicating validation across Rust call sites.

## Decision

Proceed as one independently reviewable security-hardening PR. Supported production and preview runner deployment paths use HTTPS, while local development and tests rely on loopback HTTP. The repository does not document a supported remote cleartext runner-to-API deployment.

## Scope and constraints

- Apply the policy only to the runner control-plane API base URL (`server.url`, `--api-url`, and its environment aliases).
- Preserve HTTPS for every host that passes the existing URL validation.
- Permit HTTP only when the normalized host is exactly `localhost`, a canonical IPv4 loopback address in `127.0.0.0/8`, or the IPv6 loopback address `::1`.
- Do not use DNS resolution to classify a host as local.
- Do not add a general insecure-HTTP override.
- Do not restrict unrelated guest, connector, object-download, Firecracker, or proxy HTTP traffic.
- Do not add hostname allowlisting, certificate pinning, or API/guest protocol changes.

## Compatibility and rollout

There is no frontend, backend, queue, persisted-data, or guest protocol change. Old and new runners can overlap normally. A newly deployed runner will intentionally fail closed if an existing config or runtime override uses a remote HTTP API URL; operators must move that endpoint to HTTPS or access it through explicit loopback forwarding.

## Risks and mitigations

- **Existing undocumented remote HTTP configuration stops working:** this is the intended security boundary; the error must identify the HTTPS-or-loopback requirement without echoing the raw URL.
- **Hostname lookalikes or non-canonical IP spellings bypass the exception:** classify the parsed `url::Host` after the existing raw-host validation and normalization, and test lookalike and remote forms.
- **Local development regresses:** retain and test `localhost`, IPv4 loopback across `127/8`, and `::1`, including ports and base paths.
- **The policy spreads across call sites:** keep one authority in `normalize_api_base_url`; callers continue using the existing shared boundary.

## Acceptance criteria

- `https://...` runner API URLs that satisfy existing validation remain accepted.
- HTTP is accepted only for exact normalized `localhost`, canonical IPv4 loopback addresses in `127.0.0.0/8`, and IPv6 `::1`.
- Remote DNS, public/private IPv4, and non-loopback IPv6 HTTP URLs are rejected with a configuration error that does not echo the supplied URL.
- `runner config` rejects an unsafe URL before persisting `runner.yaml`.
- Config loading and `runner start` overrides reject an unsafe URL before runtime resources start.
- `HttpClient::new()` retains the same shared validation backstop before authenticated requests can be built.
- Regression coverage exercises the operational runner configuration/start boundaries and the security-critical host-classification matrix.
- Existing loopback-based runner tests continue to pass.

seven332:
## Summary

- reject remote cleartext runner control-plane API URLs at the shared normalization boundary
- preserve HTTP for `localhost`, canonical IPv4 loopback addresses in `127.0.0.0/8`, and IPv6 `::1`
- cover persisted config, start overrides, command validation, and HTTP client construction

## Compatibility

- no API, queue, guest protocol, UI, or user-operation-flow changes
- newly deployed runners fail closed on existing remote HTTP control-plane configuration
- unrelated guest, connector, object-storage, and mock-server HTTP traffic is unchanged

## Validation

- `cargo xtask check-rust-path-attributes`
- `cargo test -p runner`

Closes #30199

### DIFF
--- a/crates/runner/src/cmd/config.rs
+++ b/crates/runner/src/cmd/config.rs
@@ -326,6 +326,22 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn run_config_rejects_cleartext_remote_api_url_before_filesystem_setup() {
+        let mut args = args_with_valid_image_hashes();
+        args.api_url = "http://api.example.com".into();
+
+        let err = run_config(args).await.unwrap_err();
+        let msg = err.to_string();
+        assert!(msg.contains("server.url"), "got: {msg}");
+        assert!(msg.contains("https"), "got: {msg}");
+        assert!(!msg.contains("api.example.com"), "got: {msg}");
+        assert!(
+            !msg.contains("rootfs not found"),
+            "API URL validation should happen before image checks: {msg}"
+        );
+    }
+
+    #[tokio::test]
     async fn run_config_rejects_mismatched_profile_arg_lengths() {
         let cases = [
             (
--- a/crates/runner/src/config.rs
+++ b/crates/runner/src/config.rs
@@ -290,8 +290,9 @@ pub(crate) fn validate_runner_hostname(value: &str) -> RunnerResult<()> {
 
 /// Validate and normalize the runner API base URL.
 ///
-/// The URL is later copied into guest-visible config and log-adjacent paths,
-/// so reject components that can carry credentials or other sensitive values.
+/// The URL is used for authenticated requests and later copied into
+/// guest-visible config and log-adjacent paths, so require HTTPS outside the
+/// loopback development boundary and reject sensitive URL components.
 pub(crate) fn normalize_api_base_url(value: &str) -> RunnerResult<String> {
     let mut parsed = url::Url::parse(value)
         .map_err(|_| RunnerError::Config("server.url must be an absolute http(s) URL".into()))?;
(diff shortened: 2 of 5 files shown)


seven332:
LGTM for head `6645fb01d8`.

No actionable findings. The runner now enforces HTTPS for remote control-plane API URLs at the shared normalization boundary while preserving exact localhost and IP loopback HTTP development paths. The host matrix, config/start wiring, command validation, and HTTP client construction are covered, and the full runner test suite passes.

PR #30433, opened 2026-08-31
fix(runner): validate dns readiness responses

### LINKED ISSUE #30419: fix(rust): validate DNS readiness response question and names
## Problem and evidence

`validate_dns_response` manually parses an untrusted UDP DNS response before the host marks a namespace DNS path ready. The current parser verifies the transaction ID, response flags, counts, and presence of the expected A-record bytes, but it does not verify that the echoed question is the readiness query and it accepts any two-byte compression pointer without validating its target.

The current implementation in `crates/sandbox-fc/src/network/readiness.rs` therefore has two confirmed fail-open cases:

- a response with the expected transaction ID and A-record bytes but a different question name, type, or class can be accepted;
- a response whose question or answer owner name contains a malformed or out-of-range compression pointer can be accepted if the remaining record layout matches.

The connected UDP socket and transaction ID reduce accidental cross-traffic, but they do not make the response semantically or structurally valid. Accepting one of these packets can report DNS readiness without proving that `vm0-readiness.invalid` resolved as an `A` / `IN` question through the intended path.

## Decision

Treat this as a response-validation bug, not only a test-coverage gap. Keep it as one independently reviewable Rust PR because the production change and its regression coverage are confined to the DNS readiness parser and endpoint probe tests.

## Scope and constraints

- Bind a successful response to the exact readiness question: `vm0-readiness.invalid`, record type `A`, class `IN`.
- Validate every encoded DNS name that the parser consumes, including compression-pointer structure and bounds.
- Preserve valid compressed and uncompressed A-record responses.
- Keep parsing bounded by the received packet and fail closed with the existing `ValidateResponse` error stage; malformed input must not panic.
- Keep endpoint probe tests deterministic and validate the externally visible probe result in addition to parser behavior.
- Do not expand this readiness-specific parser into a general-purpose DNS implementation unless planning demonstrates that a small, auditable parser cannot meet the contract.

## Alternatives considered

- **Add tests without changing validation:** rejected because the confirmed packets are currently accepted, so test-only work would preserve the bug.
- **Rely on the connected UDP socket and transaction ID:** rejected because those checks authenticate neither the echoed question nor the internal DNS name encoding.
- **Replace the parser with a general DNS library:** remains an option for planning, but is not justified by framing alone; dependency cost and the narrow readiness contract must be weighed against a focused parser.

## Risks and mitigations

- Stricter parsing could reject valid resolver responses. Cover both compressed and uncompressed names and accepted A-record layouts.
- Compression-pointer traversal can loop or escape the packet. Bound traversal, validate offsets, detect cycles, and exercise malformed inputs.
- Broad mutation tests can encode incorrect expectations for packets that remain valid after mutation. Use deterministic malformed cases for fail-closed semantics and use bounded arbitrary-input coverage primarily to prove panic freedom.

## Acceptance criteria

- Responses with a mismatched question name, type, or class are rejected even when the transaction ID and expected A record match.
- Malformed, cyclic, or out-of-range compression pointers in consumed question and answer names are rejected.
- Valid compressed and uncompressed A-record responses for `vm0-readiness.invalid` / `A` / `IN` remain accepted.
- Representative invalid flags, counts, record lengths, and truncation boundaries fail closed.
- Bounded arbitrary DNS response bytes and truncations do not panic.
- A controlled endpoint probe receiving a semantically invalid response does not report readiness.

seven332:
## Summary

- bind DNS readiness responses to the queried hostname, `A` / `IN` question, direct answer owner, and expected IPv4 value
- decode uncompressed names and bounded backward compression pointers while rejecting malformed, cyclic, truncated, and overlong inputs
- reject truncated or nonstandard-opcode responses and validate the complete declared answer section before reporting readiness
- add deterministic parser and loopback endpoint regressions plus fixed-seed bounded property coverage

## Explicit exclusions

- this remains a readiness-specific parser; it does not add general DNS resolution or CNAME-chain support
- authority and additional sections remain outside the direct readiness-answer contract

## Validation

- `cargo clippy --manifest-path crates/Cargo.toml -p sandbox-fc --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc --manifest-path crates/Cargo.toml -p sandbox-fc --no-deps`
- `cargo test --manifest-path crates/Cargo.toml -p sandbox-fc` (822 unit tests, 1 integration test)
- `cargo check --manifest-path crates/Cargo.toml -p sandbox-fc --locked`

Closes #30419

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -2824,6 +2824,7 @@ dependencies = [
  "libc",
  "nbd-cow",
  "nix",
+ "proptest",
  "reqwest",
  "rustix",
  "sandbox",
--- a/crates/sandbox-fc/Cargo.toml
+++ b/crates/sandbox-fc/Cargo.toml
@@ -31,6 +31,7 @@ rustix = { workspace = true, features = ["process"] }
 
 [dev-dependencies]
 clap = { workspace = true, features = ["derive"] }
+proptest.workspace = true
 tokio = { workspace = true, features = ["test-util"] }
 tracing-test-support.workspace = true
 tracing-subscriber.workspace = true
--- a/crates/sandbox-fc/src/network/readiness.rs
+++ b/crates/sandbox-fc/src/network/readiness.rs
@@ -33,9 +33,12 @@ pub(super) type DnsReadinessProbe = Arc<dyn Fn(String) -> DnsReadinessFuture + S
 const DNS_RESPONSE_MAX_BYTES: usize = 512;
 const DNS_QUERY_FLAGS_RECURSION_DESIRED: u16 = 0x0100;
 const DNS_RESPONSE_FLAG: u16 = 0x8000;
+const DNS_RESPONSE_OPCODE_MASK: u16 = 0x7800;
+const DNS_RESPONSE_TRUNCATED_FLAG: u16 = 0x0200;
 const DNS_RESPONSE_CODE_MASK: u16 = 0x000f;
 const DNS_TYPE_A: u16 = 1;
 const DNS_CLASS_IN: u16 = 1;
+const DNS_NAME_MAX_BYTES: usize = 255;
 /// Deadline for the blocking UDP probe on the candidate namespace.
 ///
 /// The async wait adds `PROBE_THREAD_GRACE` so the one-shot thread can publish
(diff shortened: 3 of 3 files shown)


seven332:
## Code Review: PR #30433

**Reviewed head:** `469e46bb68c97ca0020b95d2af8b9a3631e4ac18`

### Summary

The readiness-specific DNS parser now binds responses to the expected transaction, question, direct answer owner, and IPv4 value. Name decoding and complete-answer traversal are bounded and reject malformed, cyclic, truncated, or overlong packets.

### Findings

No findings. The compression-pointer cycle found during self-review was fixed on the reviewed head and covered by a deterministic regression.

### Testing

Local validation passed: Rust formatting, Clippy with warnings denied, rustdoc with warnings denied, locked Cargo check, 822 `sandbox-fc` unit tests, and 1 integration test. General DNS resolution and CNAME-chain behavior remain intentionally outside this readiness-specific parser.

### Verdict

LGTM

PR #30504, opened 2026-08-31
feat(desktop): add screen recording capture core behind a feature switch

bingjiez666:
First slice of Okou Desktop screen recording. Produces a playable MP4 from a
chosen display or window; no UI, no click track, no blur yet.

## What is here

**Feature switch** — `desktopScreenRecording`, off by default and scoped to the
maintainer. `DeveloperToolsController` already fetches `/api/feature-switches`
for two other desktop switches, so this rides that existing refresh rather than
adding a second mechanism. Withdrawing the switch (including on a 401) releases
the native helper instead of only hiding an entry point.

**Native capture** — `ScreenRecorderHelper`, a second executable product in the
existing SwiftPM package. It is a separate *process* on purpose: the Computer
Use client kills and respawns its helper on command timeouts, which would
destroy an in-flight `SCStream`/`AVAssetWriter`. Commands are
`recorder.sources|prepare|start|stop|state` over the same JSON line protocol as
`computer-use-helper`.

**Pure policies** — `ScreenRecorderCore` holds what is worth testing without a
capture device: the 1920-wide output cap (a 5K display costs ~4x the per-frame
work for detail an intro video never shows) with even dimensions because H.264
4:2:0 cannot encode odd ones, and the session transition rules.

**Session state machine** — `DesktopRecorderController`, Electron-free with the
native backend injected, mirroring `DeveloperToolsController`'s DI shape.

## Decisions worth reviewing

- Output is **fragmented MP4** (`movieFragmentInterval`), so a helper crash
  leaves a playable file rather than an unreadable one.
- The recorder client **never kills the helper on timeout**, unlike the Computer
  Use client. Killing it mid-capture would destroy the user's recording.
- The timeline is anchored on the **first video frame**, so system audio
  arriving first cannot start the session before the picture.
- No IPC or preload surface yet — nothing calls it until the UI lands, and
  adding unused IPC would mean unused handlers in the boundary test.
- Microphone is deliberately absent; it needs an `AVCaptureSession` and is not
  part of this slice.

## Verification

Run locally on Linux: `check-types`, `lint`, `knip`, desktop vitest
(409 passed), core feature-switch tests (27 passed).

Not verifiable in the sandbox — **this needs a macOS reviewer**:

- The Swift never compiled here (no toolchain); CI's macOS job builds and runs
  `swift test`.
- `SCStream` capture, `SCShareableContent` permission behavior, audio mixing,
  and `AVAssetWriter` output were never executed.
[...]

PR #30506, opened 2026-08-31
feat(desktop): record a click track alongside the screen recording

bingjiez666:
Stacked on #30504 — review that one first.

Captures when and where the user clicked during a desktop recording and writes
it beside the video as `<recording>.clicks.json`.

## Output shape

```json
{
  "version": 1,
  "recording": {
    "startedAtUnixMs": 1756612800000,
    "durationMs": 42130,
    "video": { "width": 1920, "height": 1200, "frameRate": 30 },
    "capture": { "originX": 0, "originY": 0, "widthPoints": 1512, "heightPoints": 945, "scale": 2 }
  },
  "clicks": [
    {
      "tMs": 3120,
      "button": "left",
      "clickCount": 1,
      "modifiers": ["command"],
      "screen": { "x": 812.5, "y": 430 },
      "frame": { "x": 1032, "y": 546 },
      "normalized": { "x": 0.5375, "y": 0.5056 }
    }
  ],
  "droppedOutOfFrameClicks": 4,
  "warnings": []
}
```

Three coordinate spaces are stored because they answer different questions:
`screen` is the raw fact, `frame` can be drawn straight onto the encoded video,
and `normalized` survives the video being transcoded or resized later. All three
are measured or deterministically derived — nothing is a filled-in default.

## Decisions worth reviewing

- **Session-level tap, not HID.** A listen-only tap at `.cghidEventTap` requires
  the separate Input Monitoring grant; `.cgAnnotatedSessionEventTap` rides the
  Accessibility grant the app already holds for Computer Use. Whether that
  actually delivers global mouse events is the one open question below.
- **The callback only appends.** A slow tap callback gets the entire tap
  disabled with `kCGEventTapDisabledByTimeout`, so nothing else happens there.
  `tapDisabledByTimeout` is also handled by re-enabling.
- **Clicks anchor on the first video frame's PTS**, not on when `start` was
  called, so the click timeline and the video timeline cannot drift apart.
- **Mach timebase is applied.** `CGEvent.timestamp` is in raw tick units, which
  are not nanoseconds; Apple silicon reports 125/3, so ignoring the ratio would
  place every click roughly 40x off. Covered by tests.
- **The two ways a click can fail to land are not the same thing.** A click that
  predates the first video frame is outside the recording altogether — the tap
  starts before `SCStream` delivers its first sample, so this is routine — and is
  dropped without being counted. `droppedOutOfFrameClicks` counts only clicks
  that happened during the recording but landed outside the captured region.
- **Out-of-frame clicks are counted, not described.** Recording one window and
  clicking elsewhere means the click happened in an application the user never
  agreed to record, so its position is not ours to keep. Downstream still learns
  there was activity via `droppedOutOfFrameClicks`.
- **Keystrokes are not captured.** Recording typed characters would make this a
  keylogger and nothing downstream needs them.
- **No AX element context.** You asked for time and position; element lookup is
  a separate, slower concern and would pull accessibility resolution into this
  slice. The macOS AX tree also cannot supply DOM id/class/tag, so it would not
  match what a browser extension could offer anyway.
- Losing the click track never costs the user their video: a refused tap is
  reported as a `warnings` entry on the finished recording.

## Verification

`check-types`, `lint`, `knip`, desktop vitest (416 passed) run locally. The new
`desktop-recorder-native.test.ts` drives the real line protocol against a fake
`serve` helper, covering geometry passthrough, the click track path, a lost
source, a refused permission, and the request timeout.

The Swift is again **unverified locally** — no toolchain in the sandbox; CI's
macOS job compiles it and runs `swift test`. The pure mapping and timebase logic
is in `ScreenRecorderCore` with tests precisely because it is the part worth
pinning down without hardware.

**Still needs a real Mac**, and it is the highest-risk item in this PR: whether
a session-level listen-only tap delivers global mouse events with only the
Accessibility grant. If it does not, the fallback is
`NSEvent.addGlobalMonitorForEvents`, which needs no extra grant for mouse
events. Also worth measuring there: actual drift between `CGEvent.timestamp`
and the `SCStream` frame PTS over a few minutes of recording.

The two `src/bootstrap-degraded.test.ts` failures are pre-existing on the base
commit and unrelated.


bingjiez666:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 878d1bd6b19c54003fca5c5ca27fd6bb230d3b03

### Summary

Adds a click track beside the screen recording: a listen-only
`.cgAnnotatedSessionEventTap` collects mouse-down events for the session, and
`stop()` projects them onto the recording's timeline and geometry into
`<recording>.clicks.json`. The reasoning behind the design is strong — session
tap over HID to avoid a second TCC grant, append-only callback, anchoring on the
first frame's PTS, applying the mach timebase, refusing to keep out-of-frame
positions, no keystrokes. `ScreenRecorderCore` holds exactly the parts worth
pinning without hardware, and `ClickMappingTests` covers them well.

The blocker is that `droppedOutOfFrameClicks` does not mean what its name and
the PR summary say it means, on a path that fires during normal recordings.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness, unhandled edge cases | Finding (P1) | `ClickTrackRecorder.track` counts pre-timeline clicks in `droppedOutOfFrame`; `main.swift:196 hostTicks` returns a fabricated `0` anchor. |
| `REVIEW.md` | `feat:` commits must ship integration tests | Pass | `ClickMappingTests.swift` (+geometry mapping, edges, degenerate region) and `ClickTimelineTests` (timebase ratio, zero offset, pre-recording events, 1:1 timebase); `desktop-recorder-native.test.ts` exercises `clickTrackPath` over the real line protocol. |
| `REVIEW.md` | Type safety, no `any` | Pass | TS change is one added required field on `DesktopRecorderRecording` plus `requiredString(result, "clickTrackPath")`. |
| `REVIEW.md` | New user-facing features gated by a `FeatureSwitchKey` | Pass | Rides `desktopScreenRecording` from #30504; still `enabled: false`, maintainer-scoped. |
| `REVIEW.md` | Feature regression guard | Pass | `clickTrackPath` is required in `toRecording`, so a helper that omits it fails loudly rather than silently producing a partial recording; the controller test fixture was updated with it. |
| `REVIEW.md` | Security / privacy of new capture surface | Pass | Mouse-down only, no keystrokes, no AX element context; out-of-frame positions are discarded rather than stored. The tap is `.listenOnly` and the callback never mutates the event. |
| `REVIEW.md` | Functions over ~100 lines, duplicated logic | Finding (P1) | `frameRate: 30` in `writeClickTrack` duplicates `minimumFrameInterval = CMTime(value: 1, timescale: 30)` at `main.swift:643`. |
| `docs/bad-smell.md` | Zero `any`, zero suppressions, no dynamic imports | Pass | None present. |
| `docs/bad-smell.md` | Hardcoded configuration values | Finding (P1) | The frame rate literal above is written into a consumed JSON contract. |
| `docs/bad-smell.md` | Fail fast; no error-hiding recovery | Finding (P1) | `hostTicks` swallows a degenerate timestamp into `0` instead of leaving the timeline unset. |
| `docs/fallback.md` §3/§4 | No fallback on a state the contract cannot produce; no fabricated data | Finding (P1) | Same `hostTicks` branch — see Findings. |
| `docs/fallback.md` §2 | Non-GA feature switch needs no compatibility code | Pass | No dual-read, no legacy shape tolerance, no migration for the new sidecar. |
| `docs/fallback.md` §9 | Every introduced fallback declared in the PR summary | Pass | The diff introduces no cross-version rollout fallback, so no `Fallbacks` section is required. The `hostTicks` branch is reported below as §4 slop to remove, not as a compatibility fallback to declare. |
| `docs/fallback.md` §1 | No negative tests asserting removed behavior stays removed | Pass | No tombstone tests. |
| `docs/testing.md` | Mock at the boundary; real filesystem; no fake timers | Pass | `desktop-recorder-native.test.ts` spawns a real fake-helper process from a temp dir; no `vi.mock()`, no fake timers. |
| `docs/testing.md` | Test behavior, not mock calls | Pass | Assertions are on resolved recordings and mapped points. |
| `docs/testing/desktop-testing.md` | Swift helper policy/algorithm behavior covered by SwiftPM tests | Pass | Mapping and timeline logic live in `ScreenRecorderCore` precisely so they are testable without a capture device. |
| `docs/testing/desktop-testing.md` | JS native backend tested at the helper process boundary | Pass | `clickTrackPath` is asserted through the fake `serve` helper rather than a stubbed backend. |
| `docs/testing/anti-patterns.md` | AP-1/AP-3/AP-4 | Pass | No fs mocks, no internal module mocks, no call-count-only assertions. |
| `docs/testing/patterns.md` | Pure-function tests only for algorithmic code | Pass | `ClickMapping`/`ClickTimeline` are the algorithmic exception; the timebase ratio test is the kind of non-obvious invariant the doc names. |

### Findings

#### Critical (P0)

None.

- `ScreenRecorderHelper/ClickTrackRecorder.swift` — `track(timeline:geometry:outputSize:)`:
  `droppedOutOfFrame` is incremented for two different reasons. The `guard`
  fails either when `timeline.offsetMilliseconds` returns `nil` (the click
  predates the recording) or when `geometry.mapClick` returns `nil` (the click
  landed outside the captured region), and both land in the same counter that
  the payload publishes as `droppedOutOfFrameClicks`.

  This is not a corner case. `clickTracker.start()` runs in
  `start(outputPath:)` (`main.swift:319`), but `clickTimeline` is only anchored
  when the first video sample arrives (`main.swift:527`). Every click in that
  `SCStream` startup window is pre-timeline, and each one is reported to
  downstream as a click that happened outside the recorded window — a
  privacy-shaped claim about the user's other applications that simply is not
  true. Please either count the two causes separately, or drop pre-timeline
  clicks without incrementing a counter whose documented meaning is
  out-of-frame.

- `ScreenRecorderHelper/main.swift:196` — `hostTicks(from:)`:
  `guard nanoseconds > 0, timebase.numer > 0 else { return 0 }` fabricates a
  zero anchor. If it ever fires, `startTicks` becomes `0`, every real
  `CGEvent.timestamp` then satisfies `ticks >= startTicks`, and every click is
  timestamped from the machine's boot epoch — hours into a recording that is
  seconds long. `docs/fallback.md` §4 is explicit that deriving a plausible
  value for an impossible state produces a wrong answer instead of an alert.
  The correct outcome already exists one level up: leave `clickTimeline` unset
  and let `writeClickTrack`'s documented "a recording that never produced a
  frame has no timeline to project clicks onto" path emit an empty track. I
  agree the branch is practically unreachable for an `SCStream` frame PTS —
  which is exactly why §3 says not to write it.

- `ScreenRecorderHelper/main.swift:421` — `"frameRate": 30` duplicates
  `configuration.minimumFrameInterval = CMTime(value: 1, timescale: 30)` at
  line 643. The two agree today, but the payload is a data contract that
  #30525 and the downstream editor will read; if the capture rate is ever
  changed, the sidecar starts misreporting it with nothing to catch the drift.
  Please derive it from the configured interval or hoist a single named
  constant.

One non-blocking note: `ClickTimeline.init`'s `max(1, timebaseDenominator)`
guards a value that only ever comes from `mach_timebase_info`, so it defends a
state the caller cannot produce. It is defensible as a precondition on a
`public` initializer in a library target and `neverDividesByAZeroDenominator`
pins it, so I am not asking for a change — but it is the same shape as the
`hostTicks` finding, and worth keeping in mind if `ClickTimeline` stays
internal.

### Testing

- Coverage: **Adequate for what is testable off-hardware**, with one gap tied to
  the first finding: no test covers a click captured before the timeline is
  anchored, which is why the two drop reasons could be conflated without
  anything failing. A `ClickTrackRecorder`-level or `track()`-level case
  distinguishing the two would pin the fix.
- Conventions: **Compliant**. Swift tests target the pure `ScreenRecorderCore`
  types, the TS test drives the real line protocol against a fake `serve`
  helper, no mocks of internal modules.
- Unverifiable here and correctly called out in the PR summary: whether a
  session-level listen-only tap delivers global mouse events under the
  Accessibility grant alone, and real drift between `CGEvent.timestamp` and
  `SCStream` frame PTS. Both need a Mac; the stated
  `NSEvent.addGlobalMonitorForEvents` fallback is a reasonable contingency.
- CI note: `build-macos-arm64` fails on `sentry-cli` with `Project not found` /
  403. The same failure hits unrelated branches
  (`chore/delete-desktop-migration-policy-workflow`,
  `release-please--branches--main`) while every `merge_group` run passes, so it
  is a repo-wide credential issue on `pull_request` events, not this PR. The
  Desktop workflow is not a required check.

bingjiez666:
Pushed `1cf580a4d` addressing all three findings from the review above.

- **P1 — `droppedOutOfFrameClicks` conflated two causes**: the projection now drops pre-timeline clicks without counting them and reserves the counter for clicks that happened during the recording but landed outside the captured region. To make both rules testable, the projection moved into `ScreenRecorderCore` as `projectClicks(_:timeline:geometry:outputSize:)` alongside the mapping and timeline logic it already uses; `ClickTrackRecorder` keeps the tap and the JSON rendering. Four new cases in `ClickProjectionTests` pin it, including `dropsClicksFromBeforeTheFirstFrameWithoutCountingThemOutOfFrame` and a mixed recording that separates the two reasons.
- **P1 — `hostTicks` fabricated a zero anchor**: it now returns `nil`, and `didOutputSampleBuffer` only sets `clickTimeline` when the conversion succeeds. An unconvertible first-frame timestamp therefore falls through to the existing "a recording that never produced a frame has no timeline to project clicks onto" path and emits an empty track, instead of timestamping every click from system boot.
- **P1 — duplicated frame rate**: one `captureFrameRate` constant now feeds both `configuration.minimumFrameInterval` and the `frameRate` the click track reports.

The `ClickTimeline.init` `max(1, ...)` note was explicitly non-blocking in the review and is unchanged.

Verification: the Swift is again unverified locally — there is no Swift toolchain in this sandbox, so CI's macOS job is what compiles it and runs `swift test`. No TypeScript changed in this commit.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 7c1a3d0178d5d9a3b2004ad7538d0ac53809b24a
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Re-review at `7c1a3d017`. Two commits since the `Changes Requested` review at
`878d1bd6b`: `1cf580a4d` fixes all three findings, and `7c1a3d017` is a clean
merge of the base branch (`git diff 1cf580a4d 7c1a3d017 -- turbo/apps/desktop`
is empty, so the merge touched nothing this PR owns). All findings resolved.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness, unhandled edge cases | Pass | `projectClicks` now separates the two drop reasons; `hostTicks` returns `UInt64?` and the caller only anchors when the conversion succeeds. |
| `REVIEW.md` | `feat:` commits must ship integration tests | Pass | `ClickProjectionTests` adds 4 cases on top of the existing mapping and timeline suites, including the mixed recording that pins both drop rules apart. |
| `REVIEW.md` | Functions over ~100 lines, duplicated logic | Pass | `captureFrameRate` is declared once and feeds both `configuration.minimumFrameInterval` and the payload's `frameRate`. |
| `REVIEW.md` | Type safety, no `any` | Pass | No TypeScript changed in `1cf580a4d`; the Swift additions are concrete types with explicit `public init`s. |
| `REVIEW.md` | New user-facing features gated by a `FeatureSwitchKey` | Pass | Unchanged: rides `desktopScreenRecording`, still `enabled: false` and maintainer-scoped. |
| `REVIEW.md` | Feature regression guard | Pass | The sidecar JSON keeps the same keys and value types, so the shape #30525 will read is unchanged; only the meaning of `droppedOutOfFrameClicks` is now what its name says. |
| `REVIEW.md` | Security / privacy of the capture surface | Pass | Improved by the fix: pre-roll clicks in the recorded region are no longer reported as evidence of activity in an application the user never agreed to record. |
| `docs/bad-smell.md` | Fail fast; no error-hiding recovery; no hardcoded config | Pass | The fabricated `0` anchor is gone; the frame rate literal is a named constant. |
| `docs/fallback.md` §3/§4 | No fallback on a state the contract cannot produce; no fabricated data | Pass | `hostTicks` returns `nil` and the existing "a recording that never produced a frame has no timeline to project clicks onto" path emits an empty track. |
| `docs/fallback.md` §9 | Every introduced fallback declared in the PR summary | Pass | Still no cross-version rollout fallback; no `Fallbacks` section required. |
| `docs/testing/desktop-testing.md` | Swift helper policy/algorithm behavior covered by SwiftPM tests | Pass | The projection moved from the executable target into `ScreenRecorderCore`, which is exactly the split this guide asks for — the tap and JSON rendering stay in the helper, the testable rules live in the library. |
| `docs/testing/patterns.md` | Deterministic tests | Pass | The new cases use fixed tick values against a 1:1 timebase; no timing dependence. |
| `docs/testing.md` | Test behavior, not mock calls | Pass | Assertions are on projected offsets, frame coordinates, and the two counters. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

All three findings from the previous review are resolved:

- `droppedOutOfFrameClicks` conflation — fixed by moving the projection into
  `ScreenRecorderCore.projectClicks`, which drops pre-timeline clicks without
  counting them and reserves the counter for clicks that happened during the
  recording but outside the captured region.
  `dropsClicksFromBeforeTheFirstFrameWithoutCountingThemOutOfFrame` and
  `separatesTheTwoDropReasonsInOneRecording` pin it. The extraction also answers
  the coverage gap I raised: the rule was previously unreachable from the test
  target, which is why the conflation could ship unnoticed.
- `hostTicks` fabricated zero anchor — now `nil`, with `didOutputSampleBuffer`
  anchoring only on success. Note that a failed conversion on the first frame is
  final for the session, since `sessionStartedAt` is set on that same frame and
  never retried; that is the right outcome, because anchoring on a later frame
  would silently offset every click from a moment that is not the video start.
- Duplicated frame rate literal — one `captureFrameRate` constant.

The PR summary has been updated to state the two drop rules explicitly, so the
documented output shape matches the code.

The `ClickTimeline.init` `max(1, ...)` note was non-blocking and is unchanged, as
stated in the previous review.

### Testing

- Coverage: **Adequate**. `ClickProjectionTests` covers an in-frame click with
  its metadata and frame coordinates, a pre-first-frame click that is dropped
  without counting, an out-of-region click that is counted, and a three-click
  recording that separates both reasons at once. `ClickMappingTests` and
  `ClickTimelineTests` are unchanged and still cover the geometry and timebase
  algorithms.
- Conventions: **Compliant**. Pure algorithm tests in the library target,
  protocol-level TypeScript test unchanged, no mocks of internal modules.
- Not verified locally, and stated plainly: this commit changed only Swift, and
  there is no Swift toolchain in the authoring sandbox, so CI's macOS job is the
  first thing to compile it and run `swift test`.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 00b4087098f8d38b4ef116873e68d79e7bfaa915
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Re-review at `00b408709`. The only change since the `7c1a3d017` LGTM is the
merge that reconciles this branch with main after #30504 landed. Nothing about
the click track itself changed; the verdict stands.

#30504 was **squash-merged** as `da12e27ba`, so its branch is not an ancestor of
main and every file it introduced came back as an add/add conflict against this
branch's copy of the same content. All five were resolved by taking this branch,
which is provably the identical content plus the click track — see the evidence
row below.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Conflict resolution keeps both sides' intent | Pass | `git diff a80bba5fc origin/main` is empty for each of the five conflicted files, so main's copy and this branch's copy of #30504's content are byte-identical; taking this branch loses nothing from main. |
| `REVIEW.md` | Merge introduced no scope creep or loss | Pass | The resolved tree's diff against main is 8 files / 734 insertions / 1 deletion, exactly the click track, and identical to its diff against the pre-merge base `a80bba5fc`. Nothing from #30504 is duplicated and nothing is dropped. |
| `REVIEW.md` | The PR's own diff against main is only its slice | Pass | `ClickMapping.swift` (+191), `ClickTrackRecorder.swift` (+174), `main.swift` (+98/-1), `ClickMappingTests.swift` (+244), and four one-to-five line `clickTrackPath` additions in the TypeScript types, native client, and two test fixtures. |
| `REVIEW.md` | Correctness, type safety, security | Pass | No source logic changed in this commit; the findings resolved at `1cf580a4d` (drop-reason conflation, fabricated zero anchor, duplicated frame rate) remain resolved. |
| `docs/fallback.md` §9 | Every introduced fallback declared in the PR summary | Pass | Still no cross-version rollout fallback; no `Fallbacks` section required. |
| `docs/testing/desktop-testing.md` | Swift policy tests and helper-protocol tests intact | Pass | `ClickMappingTests.swift` and `desktop-recorder-native.test.ts` survive the merge unchanged; the full desktop suite passes at this commit. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

One note for the two PRs still stacked behind this one: because the merge queue
squashes, #30518 and #30525 will each hit the same add/add conflict when their
base lands. The resolution is mechanical and checkable the same way — confirm
`git diff <merged-branch-head> origin/main` is empty for each conflicted file
before taking the branch side, then confirm the resolved diff against main
equals the slice's own diff against its pre-merge base.

### Testing

- Local validation at this commit: desktop suite 41 files / 424 tests passing,
  `pnpm -F @okouai/desktop check-types`, `lint`, and
  `knip --workspace apps/desktop` all clean.
- Swift: the previous head `7c1a3d017` compiled and passed `swift test` in CI's
  `build-macos-arm64` job ("Test native helper" green), which is what covers the
  `projectClicks` extraction. This commit changes no Swift beyond the conflict
  resolution, which restores the same content.
- Conventions: unchanged from the previous review.

PR #30518, opened 2026-08-31
feat(desktop): drive screen recording from the menu bar and deliver it back to okou

bingjiez666:
Stacked on #30506, which is stacked on #30504 — both merged.

This PR now carries two slices: the menu bar control that was #30518, and the
delivery path that was #30525. They were folded together to spend one CI cycle
and one merge-queue wait instead of two; #30525 is closed in favour of this one.

---

# Part 1 — menu bar control

Makes the capture core actually usable: start and stop a recording, with **no
on-screen controls at all**.

## Why the menu bar and not a floating control bar

An overlay window sits on the screen we are capturing, so it would be composited
into the user's own video. Avoiding it entirely means nothing has to be excluded
from the frames, and the `SCContentFilter` window-exclusion question never has to
be answered. Recording controls therefore live in the tray plus one global
shortcut.

## What lands

- **Tray section** — absent unless `desktopScreenRecording` is on. Idle shows
  `Record Main Display`; recording shows a live `mm:ss` clock and
  `Stop Recording (⌃⇧R)`; a failed capture shows why; a finished one shows where
  the file went.
- **`Control+Shift+R`** stops from any application. Registered *only* while
  recording so the shortcut is not held hostage the rest of the time, and
  released on quit.
- **Status polling** runs for exactly the duration of a capture. The helper
  protocol has no push channel, so this is how an unplugged display or a closed
  window surfaces.
- **`startMainDisplayRecording`** prepares and starts in one step, because the
  menu bar cannot host a source picker. Choosing a specific window needs that
  picker and is a later slice.

## Scope notes

- No IPC or preload surface still — the whole control plane is in the main
  process, so there is nothing for the renderer to call and no unused handlers
  in the boundary test.
- Pause/resume is not wired. `SCStream` has no pause, so it needs PTS offsetting
  to avoid a gap in the video; not worth carrying before the feature is used.

## Verification

`check-types`, `lint`, `knip`, desktop vitest (427 passed) locally. Eleven new
tests cover the tray section (hidden while switched off, the elapsed clock
including sub-minute padding, start/stop wiring, no start action mid-recording,
error and saved-path rows) and the one-step start (picks a display over a
window, reports having no display without touching prepare).

Everything here is TypeScript, so unlike the two PRs below it this slice is
fully exercised by the checks. **Still unverified on hardware**: whether the
global shortcut registers cleanly on a real Mac, and how the tray reads during
an actual capture.

sandbox and pass in CI.

---

# Part 2 — delivery back to okou

Closes the loop: stop a recording in the menu bar, and it uploads and opens in
the browser for review.

## Flow

`stop` → upload MP4 → upload `clicks.json` → open
`app.okou.ai/?intro-video-recording=…&intro-video-clicks=…&intro-video-user=…`

## Decisions worth reviewing

- **Two attachments, not one.** The click track is uploaded separately so the
  editor can read the interaction data without parsing the video container.
- **The presigned PUT uses a separate unauthenticated fetch.** That URL already
  carries its own signature; sending Okou session cookies to storage would leak
  them outside the API. The two fetches are distinct dependencies so this cannot
  be wired wrong by accident.
- **Files stream from disk** via `openAsBlob` rather than being read into memory.
  A long recording is easily hundreds of megabytes.
- **Sign-in is checked before the capture starts.** Checking after would mean
  telling someone their five-minute recording cannot be delivered once it is too
  late to do anything about it.
- **A failed upload does not lose the recording.** Delivery errors are captured
  into state instead of propagating out of `stop`, the files stay on disk, and
  the tray offers `Retry Delivery`. This is one of the few places a catch earns
  its keep — there is a real recovery.
- **The review link carries the signed-in user id.** Uploaded objects are owned
  by the account that created them, so a browser signed in as a different
  account fails to load the attachment. Without this the browser cannot tell
  that mismatch apart from a generic failure. Acting on it is the browser's job
  in the next slice.
- **No multipart.** The API only returns multipart parts when asked, and a
  single PUT from the main process has no body-size limit to work around. An
  unexpected multipart response is rejected explicitly rather than silently
  producing a broken upload.
- `fetchWithSessionAuth` gains an optional `init` so it can issue POSTs. Session
  headers are merged last, so a caller cannot drop the cookie or bearer token.

## Not in this PR

The browser side. Nothing consumes those query parameters yet, so the link
currently opens a normal Okou page. Seeding the intro video wizard from an
already-uploaded attachment is a change in `apps/platform` against a
1,000-line signals file and the existing `Source → Avatar → Voice → Review`
wizard — different app, different reviewers, and large enough that folding it in
here would make neither half reviewable. It is the next slice.

## Verification

`check-types`, `lint`, `knip`, desktop vitest (439 passed) locally. Twelve new
tests: seven on delivery itself (per-file name/type/size, credentials kept off
the storage PUT, the review link, which step failed and its status, stopping
before upload on a rejected session, refusing an unrequested multipart response)
and five on the controller (upload-then-open, refusing to record while signed
out before anything is captured, keeping the recording when delivery fails,
retrying the same recording, and having nothing to retry).

All TypeScript, so the checks genuinely exercise it. **Unverified on hardware**:
a real upload against the live API, and whether the browser lands where intended
once the next slice consumes the parameters.

sandbox and pass in CI.

---

## Combined verification

Desktop suite 42 files / 447 tests passing, plus `check-types`, `lint`,
`knip --workspace apps/desktop`, and `prettier --check apps/desktop/src`, all
clean at `a4ce60f84`.

One conflict arose when folding the two together, in `stop()`: this branch had
the fix that restores `recording` when the native stop rejects, and the
delivery branch rewrote the same method to run the upload. Both are kept — the
stop still recovers its status on failure, and delivery runs after it succeeds.


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: df092fd7f4a299b99791dba6e673e7db243567f0
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Adds the tray section and a global stop shortcut so the capture core is usable
without any on-screen control that would be composited into the user's own
video. The section is absent unless the recorder reports itself available, the
poll timer and `Control+Shift+R` are alive for exactly the duration of a
capture, and `startMainDisplayRecording` prepares and starts in one step because
the menu bar cannot host a source picker. Avoiding an overlay entirely — and
with it the whole `SCContentFilter` window-exclusion question — is the right
call.

This head also rebuilds the slice on main after #30506 squash-merged; that
reconciliation is verified in the criteria table below.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Rebase/merge introduced no loss or scope creep | Pass | This branch forked from `fed047f68`, before the three fixes that landed with #30506, so a wholesale "take ours" would have reverted them. The six files this slice never touches take main's version; the four it does change take main's version with this slice's delta reapplied. The resolved diff against main is 8 files / 383 insertions, byte-identical to `git diff fed047f68 679ccd772`. |
| `REVIEW.md` | `feat:` commits must ship integration tests | Pass | 11 new cases: 8 through `buildDesktopTrayMenuItems` (hidden while off, hidden when unavailable, idle start, elapsed clock, sub-minute padding, no start mid-recording, transient states, error row, saved-path row) and 3 on `startMainDisplayRecording`. |
| `REVIEW.md` | Correctness, unhandled edge cases | Finding (P1) | `startMainDisplayRecording` takes the first `display` source, which the helper does not order by primary — see Findings. |
| `REVIEW.md` | Resource lifecycle | Pass | `notifyScreenRecorderChanged` toggles on `isRecording === (timer !== null)`, so the interval and the shortcut are created and released exactly once per capture. Withdrawing the switch mid-capture reaches the same path through `releaseAfterDisable`, and `globalShortcut.unregisterAll()` runs on quit. |
| `REVIEW.md` | New user-facing features gated by a `FeatureSwitchKey` | Pass | `buildScreenRecordingSection` returns `[]` unless `recorder?.available`, which is false whenever `desktopScreenRecording` is off. Two tests pin both the absent-state and unavailable-state cases. |
| `REVIEW.md` | Feature regression guard | Pass | `DesktopTrayMenuActions` and `DesktopTrayControllerOptions` gain required members, so every existing construction site had to be updated and the compiler enforces it; the existing tray tests were updated rather than replaced. |
| `REVIEW.md` | Type safety, no `any` | Pass | `recorder?: DesktopRecorderState` is optional by design and narrowed once; no assertions added. |
| `docs/bad-smell.md` | Zero suppressions, no dynamic imports, no hardcoded config | Pass | The accelerator and its label are named constants in `desktop-recorder-types.ts`; the poll interval is `SCREEN_RECORDING_POLL_INTERVAL_MS`. |
| `docs/fallback.md` §9 | Every introduced fallback declared in the PR summary | Pass | No cross-version rollout fallback; no `Fallbacks` section required. |
| `docs/testing/desktop-testing.md` | Tray menu state matrices are an accepted pure-test exception | Pass | The new cases extend `desktop-tray-menu.test.ts`, which the guide names explicitly as an accepted matrix test. |
| `docs/testing/desktop-testing.md` | Avoid wholesale unit tests for `main.ts` | Pass | `main.ts` gains wiring only; no test reaches into it. See the note below on what that leaves uncovered. |
| `docs/testing/anti-patterns.md` | AP-1/AP-3/AP-4 | Pass | Menu items are exercised by clicking them and asserting the action fired plus the surrounding labels; no fs or internal-module mocks. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1) - should fix before the switch widens

- `desktop-recorder-controller.ts` — `startMainDisplayRecording()` picks
  `sources.find(s => s.kind === "display")`, the first display the helper
  reports. The helper builds that list by iterating `SCShareableContent.displays`
  (`handleSources`), whose order is not documented as primary-first, so on a
  multi-display Mac this can start recording a secondary screen while the menu
  says `Record Main Display` and the method name promises the main one. The
  single-display case — almost certainly what has been exercised so far — hides
  it. `ClickMappingTests` already covers a secondary display with a negative
  origin, so multi-monitor is a real target for this feature.

  The helper already has `display.displayID` in hand, so marking the primary
  with `CGMainDisplayID()` and selecting on that is a small change; renaming the
  action to `Record Display` would also be honest. Either is fine, but the
  current pairing of name and behavior is not.

Two non-blocking notes:

- `clearInterval(screenRecordingPollTimer ?? undefined)` — control flow
  guarantees the timer is non-null on that branch, so the `?? undefined` is
  dead. Harmless, but it reads as if the state could be missing.
- The poll-timer and shortcut lifecycle in `main.ts` is the one piece of this
  slice with no test. That matches `docs/testing/desktop-testing.md`, which
  asks you not to unit-test `main.ts` wholesale and offers no boundary that
  would reach this logic, so I am not asking for a test — just flagging that
  "registers cleanly on a real Mac", already called out in the PR summary, is
  the only thing that will confirm it.

### Testing

- Coverage: **Adequate.** The tray section's state matrix is covered end to end,
  including the two paths most likely to regress (no start action while
  recording, section absent while the switch is off). `startMainDisplayRecording`
  is covered for the happy path, window-skipping, and the no-display error,
  where it also asserts `prepare` was never called and the machine stayed `idle`.
- Conventions: **Compliant.** Integration boundary for the menu, real controller
  with an injected fake backend, no internal mocks, no fake timers.
- Local validation at this commit: desktop suite 41 files / 435 tests passing,
  `pnpm -F @okouai/desktop check-types`, `lint`, `knip --workspace apps/desktop`,
  and `prettier --check apps/desktop/src` all clean.
- CI at this commit: `ci-gate-security`, `ci-gate-crates`, and `ci-gate-turbo`
  all green.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: a4ce60f848447b4586f8d852934e9e3b32ab6c23
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Re-review of the combined PR at `a4ce60f84`, which now carries both the menu bar
control reviewed at `df092fd7f` and the delivery path that was #30525. The
delivery half is reviewed here for the first time.

Stop a recording in the tray, and the MP4 and its `clicks.json` upload as two
attachments before the browser opens on the review link. The separations are the
right ones: the presigned PUT goes through its own unauthenticated `fetchUpload`
so session cookies never reach storage, files stream from disk as blobs rather
than being read into memory, and sign-in is checked in `prepare()` before the
capture starts instead of after a five-minute recording is already on disk.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Fold of two slices lost nothing | Pass | One conflict, in `stop()`: this branch had the fix restoring `recording` when the native stop rejects, the delivery branch had rewritten the method to run the upload. Both survive — `restoreStatusOnFailure` still wraps `backend.stop`, and `runDelivery` runs after it resolves. |
| `REVIEW.md` | `feat:` commits must ship integration tests | Pass | 23 new cases across the two slices: 11 on the tray section and one-step start, 12 on delivery (per-file name/type/size, credentials kept off the storage PUT, review link shape, which step failed and its status, refusing to record while signed out, refusing an unrequested multipart response, retry, nothing-to-retry). |
| `REVIEW.md` | Security — credentials scope | Pass | `fetchUpload` is a distinct dependency from `fetchWithSessionAuth`, so the presigned PUT cannot pick up session headers by accident, and the split is asserted by a test. `headersFor` merges caller headers *under* the session headers, so a caller cannot drop the cookie or bearer token. |
| `REVIEW.md` | Correctness, unhandled edge cases | Finding (P1) ×2 | Main-display selection, and retry after a partial upload — see Findings. |
| `REVIEW.md` | Error handling — only catch what you can handle | Pass | `runDelivery`'s catch has a real recovery: the capture already succeeded, both files are on disk, the error becomes `delivery_failed` state, and the tray offers `Retry Delivery` gated on exactly that code plus a `lastRecording`. |
| `REVIEW.md` | Retry safety of the authenticated fetch | Pass | `fetchWithSessionAuth` re-sends `{...init}` once on 401. Only the two JSON calls go through it and their bodies are strings, so the retry cannot replay a consumed stream; the blob body goes to `fetchUpload`, which does not retry. |
| `REVIEW.md` | New user-facing features gated by a `FeatureSwitchKey` | Pass | The whole tray section is absent unless `recorder?.available`, which is false while `desktopScreenRecording` is off. |
| `docs/bad-smell.md` | No hardcoded config; no suppressions; no `any` | Pass | `apiBaseUrl` and `appUrl` are injected; content types and the accelerator are named constants; narrowing goes through `isRecord` and explicit `typeof` checks. |
| `docs/fallback.md` §3 | No fallback on a state the contract cannot produce | Pass | `prepareUpload` rejects a missing `id` or `uploadUrl` outright instead of inventing one, and rejects a multipart response explicitly rather than silently producing a broken upload. `uploadHeaders` defaulting to `{}` is genuinely optional provider data. |
| `docs/fallback.md` §9 | Every introduced fallback declared in the PR summary | Pass | No cross-version rollout fallback; no `Fallbacks` section required. |
| `docs/testing/desktop-testing.md` | Tray menu state matrices are an accepted pure-test exception | Pass | The new tray cases extend `desktop-tray-menu.test.ts`, named in the guide as an accepted matrix test. |
| `docs/testing/desktop-testing.md` | Mock at the system boundary only | Pass | `desktop-recorder-delivery.test.ts` injects the two fetches and `readFile`; no `vi.mock()` of internal modules. |
| `docs/testing/anti-patterns.md` | AP-1/AP-2/AP-3/AP-4 | Pass | Assertions are on request bodies, headers, and returned ids rather than call counts; no `vi.stubGlobal("fetch")`, no fs mocks. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1) - should fix before the switch widens

- `desktop-recorder-controller.ts` — `startMainDisplayRecording()` takes
  `sources.find(s => s.kind === "display")`, the first display the helper
  reports. The helper builds that list from `SCShareableContent.displays`, whose
  order is not documented as primary-first, so on a multi-display Mac this can
  record a secondary screen while the menu says `Record Main Display`. The
  single-display case hides it. `display.displayID` is already in hand on the
  Swift side, so selecting on `CGMainDisplayID()` is small; renaming the action
  to `Record Display` would also be honest.

- `desktop-recorder-delivery.ts` — `deliverRecording()` uploads the video and
  then the click track, and `retryDelivery()` re-runs the whole function. If the
  click track upload fails after the video succeeded, the retry uploads the
  video a second time: an orphaned attachment plus a second transfer of a file
  the PR summary itself describes as easily hundreds of megabytes. Nothing is
  lost, so this is not a correctness bug, but it makes the advertised retry
  expensive in exactly the case it exists for. Carrying the completed upload ids
  on `lastRecording` (or in state) and skipping what already finished would fix
  it.

Two non-blocking notes, both carried over from the `df092fd7f` review:

- `clearInterval(screenRecordingPollTimer ?? undefined)` — control flow
  guarantees the timer is non-null on that branch, so the `?? undefined` is dead.
- The poll-timer and shortcut lifecycle in `main.ts` has no test. That matches
  `docs/testing/desktop-testing.md`, which asks you not to unit-test `main.ts`
  wholesale and offers no boundary reaching this logic, so I am not asking for
  one — but "registers cleanly on a real Mac", already called out in the PR
  summary, is the only thing that will confirm it.

### Testing

- Coverage: **Adequate.** Delivery is covered at its dependency boundary,
  including the two paths that matter most for safety — session credentials
  staying off the storage PUT, and refusing to start a capture while signed out
  before anything is recorded. The controller covers upload-then-open, keeping
  the recording when delivery fails, retrying the same recording, and having
  nothing to retry.
- Conventions: **Compliant.** Injected dependencies rather than module mocks, no
  fake timers, no direct `fetch` stubbing, behaviour asserted on request
  contents and resulting state.
- Local validation at this commit: desktop suite 42 files / 447 tests passing,
  `pnpm -F @okouai/desktop check-types`, `lint`, `knip --workspace apps/desktop`,
  and `prettier --check apps/desktop/src` all clean.
- Not verifiable here, and correctly called out in the PR summary: a real upload
  against the live API, and where the browser actually lands.

PR #30525, opened 2026-08-31
feat(desktop): deliver finished recordings back to okou

bingjiez666:
Stacked on #30518 → #30506 → #30504. Review those first.

Closes the loop: stop a recording in the menu bar, and it uploads and opens in
the browser for review.

## Flow

`stop` → upload MP4 → upload `clicks.json` → open
`app.okou.ai/?intro-video-recording=…&intro-video-clicks=…&intro-video-user=…`

## Decisions worth reviewing

- **Two attachments, not one.** The click track is uploaded separately so the
  editor can read the interaction data without parsing the video container.
- **The presigned PUT uses a separate unauthenticated fetch.** That URL already
  carries its own signature; sending Okou session cookies to storage would leak
  them outside the API. The two fetches are distinct dependencies so this cannot
  be wired wrong by accident.
- **Files stream from disk** via `openAsBlob` rather than being read into memory.
  A long recording is easily hundreds of megabytes.
- **Sign-in is checked before the capture starts.** Checking after would mean
  telling someone their five-minute recording cannot be delivered once it is too
  late to do anything about it.
- **A failed upload does not lose the recording.** Delivery errors are captured
  into state instead of propagating out of `stop`, the files stay on disk, and
  the tray offers `Retry Delivery`. This is one of the few places a catch earns
  its keep — there is a real recovery.
- **The review link carries the signed-in user id.** Uploaded objects are owned
  by the account that created them, so a browser signed in as a different
  account fails to load the attachment. Without this the browser cannot tell
  that mismatch apart from a generic failure. Acting on it is the browser's job
  in the next slice.
- **No multipart.** The API only returns multipart parts when asked, and a
  single PUT from the main process has no body-size limit to work around. An
  unexpected multipart response is rejected explicitly rather than silently
  producing a broken upload.
- `fetchWithSessionAuth` gains an optional `init` so it can issue POSTs. Session
  headers are merged last, so a caller cannot drop the cookie or bearer token.

## Not in this PR

The browser side. Nothing consumes those query parameters yet, so the link
currently opens a normal Okou page. Seeding the intro video wizard from an
already-uploaded attachment is a change in `apps/platform` against a
1,000-line signals file and the existing `Source → Avatar → Voice → Review`
wizard — different app, different reviewers, and large enough that folding it in
here would make neither half reviewable. It is the next slice.

## Verification

`check-types`, `lint`, `knip`, desktop vitest (439 passed) locally. Twelve new
tests: seven on delivery itself (per-file name/type/size, credentials kept off
the storage PUT, the review link, which step failed and its status, stopping
before upload on a rejected session, refusing an unrequested multipart response)
and five on the controller (upload-then-open, refusing to record while signed
out before anything is captured, keeping the recording when delivery fails,
retrying the same recording, and having nothing to retry).

All TypeScript, so the checks genuinely exercise it. **Unverified on hardware**:
a real upload against the live API, and whether the browser lands where intended
once the next slice consumes the parameters.

sandbox and pass in CI.

### DIFF
--- a/turbo/apps/desktop/src/desktop-auth-session.ts
+++ b/turbo/apps/desktop/src/desktop-auth-session.ts
@@ -129,9 +129,18 @@ export class DesktopAuthSession {
     return await this.refresh();
   }
 
-  async fetchWithSessionAuth(requestUrl: URL): Promise<Response> {
+  /**
+   * `init` carries the method and body for non-GET calls. Its `headers` are
+   * merged under the session headers, which always win, so a caller cannot
+   * accidentally drop the cookie or bearer token.
+   */
+  async fetchWithSessionAuth(
+    requestUrl: URL,
+    init?: RequestInit,
+  ): Promise<Response> {
     const response = await fetch(requestUrl, {
-      headers: await this.headersFor(requestUrl),
+      ...init,
+      headers: await this.headersFor(requestUrl, init?.headers),
     });
     if (response.status !== 401 || !this.token) {
       return response;
@@ -139,7 +148,8 @@ export class DesktopAuthSession {
 
     this.token = null;
     return await fetch(requestUrl, {
-      headers: await this.headersFor(requestUrl),
+      ...init,
+      headers: await this.headersFor(requestUrl, init?.headers),
     });
   }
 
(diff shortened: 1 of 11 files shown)


bingjiez666:
Folded into #30518 and closing this one.

The two slices now ship together to spend one CI cycle and one merge-queue wait instead of two. Every commit from this branch is in `a4ce60f84` on `feat/desktop-screen-recording-tray-control`; the combined desktop suite is 42 files / 447 tests passing.

One conflict came up in the fold, in `DesktopRecorderController.stop()`: #30518's branch carried the fix that restores `recording` when the native stop rejects, and this branch had rewritten the same method to run the upload. Both are kept — the stop still recovers its status on failure, and delivery runs after it succeeds.

The branch `feat/desktop-screen-recording-delivery` is left in place; nothing depends on it any more.

PR #30553, opened 2026-08-31
perf(vsock-proto): encode storage results directly

### LINKED ISSUE #30538: perf(rust): encode guest storage manifest results directly into the frame
## Location

- `crates/vsock-proto/src/payloads/guest_storage_manifest.rs:180-197`
- Production caller: `crates/vsock-guest/src/guest_storage_manifest.rs:243-258`

## Problem and evidence

`encode_guest_storage_manifest_result_frame_into` first calls `encode_guest_storage_manifest_result`, which delegates to `crate::encode_exec_result`. That allocates an exactly sized payload `Vec` and copies the captured stdout, stderr, and diagnostic into it. The function then calls `encode_into`, which reserves a second full buffer for the frame and copies the complete payload again with `extend_from_slice`.

The guest caller still owns the bounded stdout/stderr capture buffers while this happens. The adjacent generic exec path already avoids this shape with `encode_exec_result_frame_into` in `crates/vsock-proto/src/payloads/exec_operation/result.rs:162-185`, but the newer storage-manifest result wrapper does not use an equivalent direct encoder.

The storage helper retains up to 1 MiB per stream (`GUEST_STORAGE_MANIFEST_OUTPUT_LIMIT_BYTES` at `crates/vsock-proto/src/payloads/guest_storage_manifest.rs:18-19`). If a helper fills both streams, result construction keeps roughly:

- 2 MiB in the original capture buffers;
- another 2 MiB in the intermediate encoded payload; and
- another 2 MiB in the complete frame;

alive at the same time, before allocator capacity overhead. It also performs two full copies of the retained output. The operation is on the guest terminal-result path, and concurrent sandboxes multiply the transient memory pressure; even small results pay the extra allocation and copy.

## Scope

This issue is limited to the storage-manifest result path; it does not reopen the completed generic exec-result optimization. The existing owned payload encoder remains public for callers that need an owned payload.

The storage-specific 1 MiB-per-stream limit and the exec-result diagnostic `u16` length limit keep every valid storage-manifest result well below the protocol's 16 MiB frame limit. Frame-size overflow is therefore not independently reachable through this API; the implementation must still preserve the shared encoder's size validation, but focused boundary tests should target the reachable storage-output and diagnostic limits.

## Acceptance criteria

- Add or reuse a direct fixed-result-frame encoder that performs the storage-specific output validation and appends the result payload directly into the caller-provided frame, without an intermediate payload `Vec`.
- Use the direct encoder from the guest storage-manifest worker while preserving the existing message type, wire bytes, validation/error behavior, writer-lock ordering, cancellation, and operation-completion semantics.
- Keep the existing public payload encoder for callers that need an owned payload.
- Add focused tests covering byte-equivalence with the current payload-plus-frame encoding across every termination variant, truncation flags, diagnostics, and near-limit output.
- Preserve rejection of discarded output, one byte over the 1 MiB per-stream bound, and an overlong diagnostic, including the destination-frame state on validation errors.

seven332:
## Summary

- reuse the shared exec-result validator and appender through a crate-private, message-type-specific direct frame helper
- encode guest storage-manifest results into the final frame without an intermediate payload allocation or full payload copy
- preserve the owned payload API, wire bytes, and destination-frame behavior on validation errors
- cover stream and diagnostic boundaries through the public storage frame encoder

## Scope

- no protocol message type, payload schema, public function signature, or validation limit changes
- no changes to guest writer-lock ordering, cancellation, or operation completion

## Validation

- `cargo xtask check-rust-path-attributes`
- `cargo clippy -p vsock-guest --release --no-default-features --all-targets`
- `cargo test -p vsock-proto`
- `cargo test -p vsock-guest --all-features`

Closes #30538

### DIFF
--- a/crates/vsock-proto/src/payloads/exec_operation.rs
+++ b/crates/vsock-proto/src/payloads/exec_operation.rs
@@ -18,6 +18,7 @@ pub use output::{
     DecodedExecOutput, ExecOutputStream, decode_exec_output, encode_exec_output,
     encode_exec_output_frame_into,
 };
+pub(crate) use result::encode_exec_result_frame_into_with_type;
 pub use result::{
     DecodedExecResult, ExecCapturedOutput, decode_exec_result, encode_exec_result,
     encode_exec_result_frame_into,
--- a/crates/vsock-proto/src/payloads/exec_operation/result.rs
+++ b/crates/vsock-proto/src/payloads/exec_operation/result.rs
@@ -169,9 +169,29 @@ pub fn encode_exec_result_frame_into(
     diagnostic: &str,
 ) -> Result<(), ProtocolError> {
     frame.clear();
+    encode_exec_result_frame_into_with_type::<MSG_EXEC_RESULT>(
+        frame,
+        seq,
+        termination,
+        duration_ms,
+        stdout,
+        stderr,
+        diagnostic,
+    )
+}
+
+pub(crate) fn encode_exec_result_frame_into_with_type<const MSG_TYPE: u8>(
+    frame: &mut Vec<u8>,
+    seq: u32,
+    termination: ExecTermination,
+    duration_ms: u32,
+    stdout: ExecCapturedOutput<'_>,
+    stderr: ExecCapturedOutput<'_>,
+    diagnostic: &str,
+) -> Result<(), ProtocolError> {
     let (diagnostic_len, payload_len) =
         validate_exec_result_payload(termination, stdout, stderr, diagnostic)?;
-    encode_into(frame, MSG_EXEC_RESULT, seq, payload_len, |frame| {
+    encode_into(frame, MSG_TYPE, seq, payload_len, |frame| {
         append_exec_result_payload(
             frame,
             termination,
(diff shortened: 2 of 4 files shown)


seven332:
## Code Review: PR #30553

**Reviewed head:** `3929499d9bc017129627aa4e3562ad1e5f0e5639`

### Summary

The storage-manifest result path now reuses the authoritative exec-result validator/appender and writes directly into the final `0x19` frame. The crate-private const message type keeps the public API and wire contract unchanged, while wrapper ordering preserves existing destination-buffer behavior on errors.

### Findings

No findings.

### Testing

Workspace formatting, Rust path validation, workspace/all-feature Clippy, production-config guest Clippy, workspace rustdoc, `vsock-proto` tests, and `vsock-guest --all-features` tests passed. Public integration coverage checks full-frame equivalence and reachable stream/diagnostic error boundaries; Codecov reports all modified coverable lines covered.

No allocation-counting harness or benchmark exists for this encoder, so the allocation removal was verified structurally rather than instrumented.

### Verdict

LGTM

PR #30566, opened 2026-08-31
test(python): cover non-positive stream decode chunk limits

### LINKED ISSUE #30546: test(python): reject non-positive streaming decode chunk limits
## Problem

`create_stream_decode_session` documents and implements a per-callback decoded-output bound through `max_decoded_chunk` (`crates/runner/mitm-addon/src/body_decoding.py:293-330`), but the Python test suite does not exercise non-positive values. The validation branch at line 329 is therefore unprotected by a direct API regression test.

## Evidence

- The focused body-decoding suite and all current consumers pass without exercising the guard: `109 passed` in `tests/test_body_decoding.py`, and `400 passed` across every test module importing `body_decoding`.
- In a temporary mutant with the `max_decoded_chunk <= 0` guard removed, `max_decoded_chunk=0` creates sessions for identity and gzip. Identity forwards arbitrary input, while zlib treats `max_length=0` as unlimited and emits the full decoded payload, violating the documented independent chunk bound. Brotli reaches `ValueError: range() arg 3 must not be zero` during feed.
- With `max_decoded_chunk=-1`, identity still forwards input, gzip raises `ValueError: max_length must be non-negative` only after feed, and Brotli silently emits no output. The failure is thus codec-dependent and occurs after callers have already received a session.

The production response-streaming wrapper currently uses the positive default, so this is a public-helper and refactor/configuration regression boundary rather than a claim that a user currently controls this integer. The missing test would keep future callers from bypassing the resource bound or turning an invalid configuration into a mid-stream exception.

## Requested coverage

Add parameterized tests for `max_decoded_chunk` values `0` and `-1` on the public `create_stream_decode_session` API (including representative identity, zlib, and Brotli headers), asserting that `ValueError` is raised at session creation before any decoder callback can run. Keep the existing positive/default tests to verify bounded delivery remains unchanged.

seven332:
## Summary

- add a six-case public-helper matrix for zero and negative streaming decode chunk limits
- cover identity, gzip, and Brotli dispatch paths at session construction
- verify invalid limits raise before the decoder callback receives output

## Scope

This is a test-only change. Production decoder behavior, defaults, callers, dependencies, and configuration are unchanged.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_body_decoding.py` (`115 passed`)
- `uv run --no-sync python -m pytest tests/` (`7251 passed`)

Closes #30546

### DIFF
--- a/crates/runner/mitm-addon/tests/test_body_decoding.py
+++ b/crates/runner/mitm-addon/tests/test_body_decoding.py
@@ -240,6 +240,20 @@ class TestStreamDecodeSession:
         assert chunks == [b"hello"]
         assert session.finish_error() is None
 
+    @pytest.mark.parametrize("encoding", ["identity", "gzip", "br"])
+    @pytest.mark.parametrize("max_decoded_chunk", [0, -1])
+    def test_non_positive_max_decoded_chunk_is_rejected(self, headers, encoding, max_decoded_chunk):
+        chunks: list[bytes] = []
+
+        with pytest.raises(ValueError, match=r"^max_decoded_chunk must be positive$"):
+            create_stream_decode_session(
+                headers(("Content-Encoding", encoding)),
+                chunks.append,
+                max_decoded_chunk=max_decoded_chunk,
+            )
+
+        assert chunks == []
+
     def test_identity_document_parser_termination_stops_callbacks(self, headers):
         extractor = JsonSelectiveExtractor(max_work_units=1)
         decoded_callbacks = 0


seven332:
## Code Review: PR #30566

**Reviewed head:** `194714db22918eecd05a6b14c621de9c2fb40c7d`

### Summary

The test-only change adds the requested public-helper boundary matrix for non-positive streaming decode chunk limits across identity, zlib, and Brotli dispatch paths. It follows the approved plan and leaves production behavior unchanged.

### Findings

No findings.

### Testing

Local validation passed: uv lock validation and sync, focused body-decoding pytest (`115 passed`), Ruff format and lint, BasedPyright, and the full mitm-addon pytest suite (`7251 passed`). GitHub Actions CI remains a separate workflow gate.

### Verdict

LGTM

PR #30464, opened 2026-08-31
fix(e2e): use real preview services in browser tests

e7h4n:
## Summary

- remove Playwright request interception and `@clerk/testing` so browser specs exercise the real App, API, Clerk, SharedWorker, and runner paths
- remove mock-driven and synthetic browser scenarios; keep deployed preview flows in Playwright and retain UI behavior coverage in App integration tests
- sign generated test accounts in through the real Clerk Auth v2 email-code UI
- seed the Vercel preview bypass as a host-only App cookie; the product forwards it only to the corresponding preview API and into the SharedWorker heartbeat
- make the Turbo aggregate gate reject cancelled required jobs instead of reporting them as allowed

## Why

The previous Playwright configuration installed `x-vercel-protection-bypass` as a BrowserContext-wide extra header. Chromium sent that header to third-party requests too, including the Clerk CDN, where the additional cross-origin header caused ClerkJS requests to fail with `net::ERR_FAILED`.

Removing the global header fixed Clerk but exposed the second half of the problem: real App API requests no longer passed Vercel preview protection and the UI stayed on `Loading your workspace`. The App already has the production preview-bypass path: it reads a host-only App cookie, adds the header only for the matching `pr-*-api.vm6.ai` origin, and passes the same scoped value to the SharedWorker.

The SharedWorker was not the network failure's cause. It made the old page-scoped response mocks and page-scoped waits structurally inaccurate because the worker owns authoritative shared-database traffic. This PR removes those mocks and runs the deployed topology directly.

This replaces the approach in closed PR #30365 and follows up on #30364.

## Validation

- `cd e2e && pnpm check-types`
- `cd e2e && pnpm test` — 25 passed
- targeted App integration tests for sidebar, inline feedback, run queue, and recommended follow-ups — 155 passed
- targeted `preview-bypass-cookie` App tests — 4 passed
- `bash .github/scripts/tests/turbo-playwright-runner-workflow-test.sh`
- `bash -n .github/scripts/tests/turbo-playwright-runner-workflow-test.sh`
- Prettier check for all changed supported files
- changed-file size check
- `git diff --check`
- exact-head Turbo run [33368798846](https://github.com/vm0-ai/vm0/actions/runs/33368798846), attempt 2 — success: Playwright `features` 7/7, `paid-onboarding` 1/1, and `auth-v2` 13/13; four runner credentials generated; every runner shard, cleanup, and the strict Turbo CI gate passed

No compatibility fallback was introduced.

PR #30580, opened 2026-08-31
fix(platform): stop connector account pagination at the last page

e7h4n:
## Trigger

Turbo run [33380291749](https://github.com/vm0-ai/vm0/actions/runs/33380291749) (`merge_group`, SHA `0269de5`) failed in `test-app` shard 7:

```
FAIL src/views/okou-page/__tests__/connectors-page.test.tsx > connectors page > paginates a feature-on account manager
Error: Test timed out in 5000ms.   (5075ms)
```

The other 342 tests in the shard passed. The queued PR #30464 touches only `e2e/playwright/**` and workflow files, so it cannot own a `turbo/apps/platform` failure.

## Root cause

Two independent causes, both in this one test.

**1. Budget headroom (why it timed out).** No `waitFor` in the test ever exceeded its own 1000ms async-utility timeout, so nothing raced. The test simply spent the whole 5000ms per-test budget. On a quiet host it was the slowest test in a 120-test file by a wide margin:

| test | duration |
| --- | --- |
| **paginates a feature-on account manager** | **2325ms** |
| names the exact account after a feature-on manual account addition | 1620ms |
| renames and deletes an exact account after bounded impact | 1251ms |
| debounces feature-on account manager searches | 1169ms |

Instrumenting the phases showed where it went: bootstrap plus the dialog with the first 50 rows 1331ms, first `Load more` to 100 rows 642ms, second `Load more` to 101 rows 654ms. The cost is rendering the fixture, and `mockGitHubConnectorAccounts(101)` existed only to cross two 50-row page boundaries. The failing shard ran 410.08s of cumulative test time in 188.51s of wall clock, i.e. about 2.2x parallel oversubscription; 2325ms x 2.2 is 5075ms, the exact observed duration.

**2. A real product bug the test was hiding.** `mergeConnectorAccountPages` computed the merged cursor as `pages.at(-1)?.nextCursor ?? firstPage.nextCursor`. `??` treats the last page's terminal `null` as "absent" and falls back to the first page's cursor, so **`Load more` never disappeared and re-requested the page after the first one forever.**

Probed on `main` with the request log:

```
after last page:  queries=[cursor:null, cursor:"3", cursor:"6"]  loadMoreVisible=true
after one more click: queries=[cursor:null, cursor:"3", cursor:"6", cursor:"3"]
```

The test never caught it because its terminal assertion matched the button by the literal text `Load more` inside a `waitFor`. While the last request is in flight the button reads `Loading more`, so `queryButtonByText("Load more", dialog)` returned `null` and the whole `waitFor` block passed mid-request, before the last page was ever applied.

## Change

`turbo/apps/platform/src/signals/okou-page/connector-accounts.ts` — the newest loaded page owns the cursor, including a terminal `null`:

```ts
nextCursor: (pages.at(-1) ?? firstPage).nextCursor,
```

`connectors-page.test.tsx` — same product contract, deterministic and cheap:

- Wait for `Work 1`, an account only the last page carries, then assert the terminal state synchronously. This pins the assertions to the render that consumed the last page instead of to a transient label.
- Assert the recorded cursor chain `[null, "3", "6"]`, which asserts directly what the old test only implied, and assert one bounded request limit wider than the whole fixture, so the extra pages provably come from following the server cursor rather than a client-side slice.
- Keep appending explicit: `Work 7` and `Work 4` are still present at the end.
- Fixture drops from 101 accounts to 8, served 3 per page. A page shorter than the requested limit with a non-null cursor is real server behaviour, because `listConnectorAccounts` projects rows away before slicing.

No retries, sleeps, larger timeouts, skips, deletions, or weakened assertions.

## Validation

Mutation-tested against `main`'s product code to confirm the smaller fixture still guards the contract:

| mutation | caught |
| --- | --- |
| `nextCursor: firstPage.nextCursor` (no chaining from the latest page) | yes, `expected [null, '3', '3'] to strictly equal [null, '3', '6']` |
| `loadMore` replaces `pages` instead of appending | yes, `Work 4` missing at the end |

Full-file runs on a healthy host:

- Before: `120 passed`, target test 2325ms, slowest in the file.
- After: `120 passed`, target test **699ms**, 14% of the 5000ms budget and mid-pack.

Related regression files, both green: `chat-composer-connector-connect.test.tsx` and `personal-usage-tab.test.tsx` (34 tests). `pnpm prettier --check`, `oxlint --deny-warnings`, `eslint --max-warnings 0`, `tsc -p tsconfig.tests.json`, `tsc -p tsconfig.production.json`, and `git diff --check` all pass.

Three further full-file runs happened while the sandbox host was memory-starved (267MB available, file wall clock 2-3.7x baseline). The target test passed in all of them at 950-1539ms, while 3-7 unrelated tests in the same file hit the same 5000ms timeout, which is the same budget-headroom mechanism this PR removes for this test. Repeated timing validation on a healthy host beyond the runs above was not possible locally; CI covers it.

No app, API, or browser environment is needed to validate this change, so no preview validation applies.

### DIFF
--- a/turbo/apps/platform/src/signals/okou-page/connector-accounts.ts
+++ b/turbo/apps/platform/src/signals/okou-page/connector-accounts.ts
@@ -83,7 +83,10 @@ function mergeConnectorAccountPages(
         return page.connections;
       }),
     ],
-    nextCursor: pages.at(-1)?.nextCursor ?? firstPage.nextCursor,
+    // The newest loaded page owns the cursor, including when it ends the list
+    // with a null cursor. Falling back to the first page's cursor there would
+    // keep "Load more" alive and re-request the page after the first one.
+    nextCursor: (pages.at(-1) ?? firstPage).nextCursor,
     available: firstPage.available,
   };
 }
--- a/turbo/apps/platform/src/views/okou-page/__tests__/connectors-page.test.tsx
+++ b/turbo/apps/platform/src/views/okou-page/__tests__/connectors-page.test.tsx
@@ -1773,12 +1773,28 @@ describe("connectors page", () => {
   });
 
   it("paginates a feature-on account manager", async () => {
-    const accounts = mockGitHubConnectorAccounts(101);
+    const accounts = mockGitHubConnectorAccounts(8);
+    // The server projects rows away before slicing, so a page can hold fewer
+    // connections than the requested limit and still report a next cursor.
+    // Serving short pages walks the same three-page cursor chain the client
+    // follows for a full connector, without rendering hundreds of rows.
+    const serverPageSize = 3;
+    const accountQueries: {
+      readonly limit: number;
+      readonly cursor: string | null;
+    }[] = [];
     context.mocks.api(
       connectorAccountsContract.connections,
       ({ query, respond }) => {
+        accountQueries.push({
+          limit: query.limit,
+          cursor: query.cursor ?? null,
+        });
         const start = query.cursor ? Number(query.cursor) : 0;
-        const page = accounts.slice(start, start + query.limit);
+        const page = accounts.slice(
+          start,
+          start + Math.min(query.limit, serverPageSize),
+        );
         const next = start + page.length;
         return respond(200, {
           connections: page,
@@ -1794,17 +1810,39 @@ describe("connectors page", () => {
 
     click(await waitForButtonByAriaLabel("Manage GitHub accounts"));
     const dialog = await screen.findByRole("dialog", { name: "GitHub" });
-    expect(within(dialog).queryByText("Work 50")).toBeNull();
+    expect(within(dialog).getByText("Work 7")).toBeInTheDocument();
+    expect(within(dialog).queryByText("Work 4")).toBeNull();
     click(buttonByText("Load more", dialog));
     await waitFor(() => {
-      expect(within(dialog).getByText("Work 50")).toBeInTheDocument();
+      expect(within(dialog).getByText("Work 4")).toBeInTheDocument();
     });
+    expect(within(dialog).getByText("Work 7")).toBeInTheDocument();
     click(buttonByText("Load more", dialog));
+    // The last page carries an account of its own, so waiting for that account
+    // pins the assertions below to the render that consumed it. Waiting on the
+    // "Load more" label instead would pass mid-request, while the button still
+    // reads "Loading more".
     await waitFor(() => {
-      expect(queryButtonByText("Load more", dialog)).toBeNull();
       expect(within(dialog).getByText("Work 1")).toBeInTheDocument();
-      expect(within(dialog).getAllByText("Account #00000000")).toHaveLength(1);
     });
+    expect(queryButtonByText("Load more", dialog)).toBeNull();
+    expect(within(dialog).getByText("Work 7")).toBeInTheDocument();
+    expect(within(dialog).getByText("Work 4")).toBeInTheDocument();
+    expect(within(dialog).getAllByText("Account #00000000")).toHaveLength(1);
+    const requestedLimits = new Set(
+      accountQueries.map((query) => {
+        return query.limit;
+      }),
+    );
+    expect(
+      accountQueries.map((query) => {
+        return query.cursor;
+      }),
+    ).toStrictEqual([null, "3", "6"]);
+    // One bounded limit per request, wider than the whole fixture: every extra
+    // page came from following the server cursor, not from a client-side slice.
+    expect([...requestedLimits]).toHaveLength(1);
+    expect([...requestedLimits][0]).toBeGreaterThan(accounts.length);
   });
 
   it("debounces feature-on account manager searches", async () => {


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 7470f895dc3de764ab85aa8d46471aa141ab2b26
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Fixes `mergeConnectorAccountPages`, where `pages.at(-1)?.nextCursor ?? firstPage.nextCursor` treated the last page's terminal `null` cursor as "absent" and fell back to the first page's cursor, so `Load more` never disappeared and re-requested the page after the first one indefinitely. Rewrites `paginates a feature-on account manager` to wait on an account only the last page carries instead of on the `Load more` label, which is what let the old assertion pass mid-request and hide the bug, and shrinks the fixture from 101 accounts to 8 short-paged accounts so the test stops sitting at ~47% of the 5000ms per-test budget.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no off-by-one or wrong conditional | Pass | `connector-accounts.ts:89` — `(pages.at(-1) ?? firstPage).nextCursor`. `??` now guards only `undefined` from `.at(-1)` on an empty `pages`, which is a genuine absence; `null` from the contract is preserved. `resetPages$` clears `pages$` and bumps `generation$` on every new first-page query, and `loadMore$` guards on that generation, so `pages` can never be stale relative to `firstPage`. |
| `REVIEW.md` | Type safety; no `any`, no broad types | Pass | Both `ConnectorAccountPage.nextCursor` and `ConnectorAccountList.nextCursor` are `string \| null`; the expression type is unchanged. Test-local `accountQueries` is explicitly typed with `readonly` members. |
| `REVIEW.md` | Security (auth, injection, secrets) | N/A | Client-side cursor merge plus a test fixture; no auth path, no persisted input, no credential handling. |
| `REVIEW.md` | `fix:` commits must carry a regression test | Pass | The rewritten test fails against pre-fix product code with `expected <button …>Load more</button> to be null`; it was that failure that surfaced the bug. |
| `REVIEW.md` | Feature regression guard; changed behavior needs coverage and containment | Pass | The only behavior change is terminal-page `Load more` removal, asserted directly at `connectors-page.test.tsx:1828`. Containment already exists: the whole surface is gated by `FeatureSwitchKey.ConnectorAccounts`, `enabled: false` with staff-only org hashes. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No `packages/db/src/schema/**`, no `api-contracts` change, no `jsonb().$type<>()`, no migration in the diff. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions, no dynamic `import()` | Pass | Diff contains none of `any`, `eslint-disable`, `@ts-ignore`, `@ts-expect-error`, `prettier-ignore`, or `await import(...)`. |
| `docs/bad-smell.md` | No hardcoded URLs or environment-specific values | Pass | No URLs or config literals added; the test's `serverPageSize` is a local fixture constant, not configuration. |
| `docs/bad-smell.md` | Comments explain WHY, not WHAT | Pass | Both added comments state rationale — why the newest page owns the cursor, and why the mock serves short pages. |
| `docs/fallback.md` §3 | No `??`/`||` fallback on a state the owning contract cannot produce | Pass | This is the defect being removed: `nextCursor` is contract-typed `string \| null`, so `null` is a value the contract deliberately produces and must not be coalesced away. |
| `docs/fallback.md` §9 | Every fallback the diff introduces is declared in a `Fallbacks` section | N/A | The diff introduces no fallback. §9 explicitly exempts PRs that only remove one, so no `Fallbacks` section is required and none is present. |
| `docs/fallback.md` §10 | Removing a fallback requires type, single-writer, production, or rollout evidence, and deletes the branch with its own tests | Pass | Type evidence is in the diff (`ConnectorAccountPage.nextCursor: string \| null`). Runtime evidence is in the PR body: the probe on `main` shows `queries=[null,"3","6"]` with `loadMoreVisible=true`, then a fourth request back to cursor `"3"`. The removed branch had no test of its own to delete — the old test passed *through* it accidentally. |
| `docs/fallback.md` §2 / §7 | A non-GA feature-switched surface needs no compatibility code or rollout gate | Pass | `FeatureSwitchKey.ConnectorAccounts` is `enabled: false`, `STAFF_ORG_ID_HASHES` only, so no rollout gate applies and no compatibility branch is owed. |
| `docs/testing.md` | Integration tests only; no unit tests for internal functions | Pass | The fix to a non-exported helper is covered through the rendered connectors page, not by a direct test of `mergeConnectorAccountPages`. |
| `docs/testing.md` | Do not skip, delete, or weaken a test to make CI pass | Pass | The test keeps three-page cursor chaining, append semantics, terminal state, and default-account de-duplication, and adds an explicit cursor-chain assertion. Net assertions increased. |
| `docs/testing/app-testing.md` | Page tests enter through `detachedSetupPage` and assert page-visible behavior | Pass | Unchanged entry at `connectors-page.test.tsx:1805`; primary assertions are rendered rows and the `Load more` button. |
| `docs/testing/app-testing.md` | Override API behavior via `context.mocks.api`; capture request data to verify what was sent | Pass | Uses `context.mocks.api(connectorAccountsContract.connections, ...)`; the `accountQueries` capture matches the documented "Tracking Request Data" pattern and the sibling `debounces feature-on account manager searches` test already in this file. |
| `docs/testing/patterns.md` | Platform page tests use the `testContext()` fixture and MSW-level HTTP mocking | Pass | Existing file-level `context = testContext()` and contract-level mock retained; no `server.use()`, no direct `fetch` stubbing. |
| `docs/testing/anti-patterns.md` AP-1 | Do not assert mock call patterns instead of behavior | Pass | The cursor-chain assertion supplements page-visible assertions rather than replacing them; no `toHaveBeenCalled()` in the diff, and no assertion rests on a mock call count alone. |
| `docs/testing/anti-patterns.md` AP-5 / AP-7 | No fake timers; no assertions on internal state, CSS classes, or handlers | Pass | No `vi.useFakeTimers()`; assertions are rendered text, button presence, and the outbound HTTP query — not component state or classes. |
| `docs/testing/testing-external-behavior.md` | Construct and verify through the page, not internal implementation | Pass | Interaction is clicking the real `Load more` button; the store is never mutated and no internal component is rendered directly. The request log is an outbound boundary the client owns, not internal state. |
| `docs/effect.md` | Derived values belong in `computed`, not in commands synchronizing state | Pass | The change stays inside the pure merge helper consumed by `accounts$` (`computed`); no command, effect, or render-phase work was added. |
| `docs/cache.md` RC-1/RC-8 | No render-phase side effects; symmetric resource lifecycles | Pass | No component, ref, timer, listener, or subscription touched; the diff adds no resource to release. |
| `.claude/skills/ccstate/SKILL.md` | `AbortSignal` ownership, `accept` usage, no floating promises, `detach` only in views | Pass | Signal plumbing, `accept([200, 404], signal)`, and the existing `detach(loadMore(signal), Reason.DomCallback)` in the dialog are untouched; the diff adds no promise. |
| `docs/deployment-compatibility.md` | Independently deployed components and persisted state stay compatible | N/A | No request or response shape, runner protocol, queue payload, schema, or persisted state changes. The client reinterprets an unchanged `nextCursor` field, so old-app/new-API and new-app/old-API both behave identically on the wire. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Notes (non-blocking)

- `connectors-page.test.tsx:1794` — the mock now caps each page at `serverPageSize`, so no test renders a real 50-row page any more and `CONNECTOR_ACCOUNT_PAGE_SIZE` is only pinned indirectly by `expect([...requestedLimits][0]).toBeGreaterThan(accounts.length)`. Changing the constant from 50 to some other value would no longer fail a test. That is an accepted trade: the client has no code path that behaves differently at exactly 50 rows, and the previous way of pinning it cost the whole per-test budget. Worth knowing if the page size ever becomes contractual.
- `connector-accounts.ts:86-90` — `nextCursor` now comes from the newest page while `available` still comes from `firstPage`. The asymmetry is correct, because only the first-page request can observe a `404` and mark the target unavailable, but the new comment explains only the cursor half.
- `connectors-page.test.tsx:1831-1845` — `[...requestedLimits]` is spread twice for the length check and the value check. Cosmetic only.

### Testing

- Coverage: Adequate. `fix:` carries a regression test that demonstrably fails on the pre-fix product code. Two mutations were run against the reduced fixture to confirm it still guards the contract: `nextCursor: firstPage.nextCursor` fails with `expected [null,'3','3'] to strictly equal [null,'3','6']`, and making `loadMore` replace instead of append `pages` fails on the missing `Work 4`. Related surfaces re-run green: `chat-composer-connector-connect.test.tsx` and `personal-usage-tab.test.tsx`, 34 tests. `mergeConnectorAccountPages` has a single consumer (`accounts$`), shared by the settings dialog and the composer, so both surfaces are covered by this one page test.
- Conventions: Compliant. No `vi.mock()` of internal paths, no `vi.stubGlobal("fetch", ...)`, no fake timers, no manual `vi.clearAllMocks`/`restoreAllMocks` (owned by `vitest.config.ts` `clearMocks`/`restoreMocks`), no mocking of a real database, no tombstone test asserting removed behavior stays removed. New user-facing behavior stays behind `FeatureSwitchKey.ConnectorAccounts`.
- Flake budget: the target test drops from 2325ms (slowest in a 120-test file) to 699ms, about 14% of the 5000ms budget, addressing the triggering timeout in run [33380291749](https://github.com/vm0-ai/vm0/actions/runs/33380291749) without raising any timeout.
- Static checks: `prettier --check`, `oxlint --deny-warnings`, `eslint --max-warnings 0`, `tsc -p tsconfig.tests.json`, `tsc -p tsconfig.production.json`, and `git diff --check` all pass on the changed files.

PR #30613, opened 2026-08-31
chore(deps): upgrade ccstate to 5.5.0

e7h4n:
## Summary

Upgrades `ccstate` and `ccstate-react` from `5.4.0` to `5.5.0` in `apps/api`, `apps/platform`, and `apps/desktop`.

`ccstate` 5.5.0 builds its ESM bundles for modern browsers (Chrome/Edge 111, Firefox 114, Safari 16.4) instead of the unconstrained `@babel/preset-env` default. Previously the published ESM entry was downleveled, so our client bundles carried Babel helpers that no ESM-capable browser needs. Lighthouse reported both on app.vm0.ai as "Legacy JavaScript":

- `@babel/plugin-transform-classes` — from `ccstate`
- `@babel/plugin-transform-spread` — from `ccstate-react`

Upstream PR: https://github.com/e7h4n/ccstate/pull/216

## Impact

| ESM entry | 5.4.0 | 5.5.0 |
| --- | ---: | ---: |
| `ccstate/index.js` | 34,934 B | 19,162 B |
| `ccstate-react/index.js` | 4,543 B | 3,898 B |
| `ccstate-react/experimental.js` | 3,482 B | 1,721 B |

Measured on a Vite 8 build consuming both packages: vendor chunk `2,784,928 B → 2,778,328 B`, gzip `816,347 B → 814,585 B`. Both entries disappear from the Lighthouse legacy-JavaScript audit.

## Risk

Build-output-only upgrade, no source changes upstream:

- The published `.d.ts` / `.d.cts` surface is byte-identical between 5.4.0 and 5.5.0 for both packages (verified by diffing the npm tarballs), so no type impact.
- The CJS entry still uses the previous legacy transforms, so `apps/api` (Node) is unaffected.
- `pnpm install --frozen-lockfile` passes.

The lockfile was regenerated with a plain `pnpm install --lockfile-only` (pnpm 10.33.4, via corepack), so besides `ccstate` it also re-resolves `turbo` from `2.10.12` down to `2.10.11`. Both satisfy the declared `^2.10.11` range in `turbo/package.json`, and `2.10.11` is what pnpm's default `lowest-direct` resolution mode produces — the committed lockfile had simply drifted ahead of what a plain install yields, so this is not specific to this branch.

PR #28231, opened 2026-08-19
test(e2e): isolate zendesk auth probe from ipv6 dns

e7h4n:
## Summary

- constrain the Zendesk firewall-auth probe to IPv4 while preserving the real test-specific Zendesk hostname and normal proxy path
- stop suppressing DNS and transport failures so the success marker proves transport completion
- emit a common success/failure prefix before exit so request failures surface without a second 90-second firewall-log wait
- keep the variable-base, safe-secret, header-auth, and run-owned firewall telemetry assertions unchanged

## Root cause evidence

The triggering merge-group failure is [Turbo run 32239660015](https://github.com/vm0-ai/vm0/actions/runs/32239660015), substantive job [96028884602](https://github.com/vm0-ai/vm0/actions/runs/32239660015/job/96028884602).

- the first Zendesk A query received two IPv4 answers about 21 ms later
- the paired AAAA query produced no resolver result; at the five-second boundary curl retried resolution and exited with `Resolving timed out after 5000 milliseconds`
- no Zendesk HTTP request was produced, so firewall auth and request telemetry were never reached; the later helper timeout was downstream fallout
- a separate fresh sandbox on the same runner instance timed out in a direct UDP DNS probe in the same second, while an adjacent merge-group job passed this exact Zendesk test concurrently on another runner
- the represented PR changes only guest-contract test code, and its PR-head run plus adjacent merge-group runs passed this shard

This is the same dropped-AAAA failure shape addressed for SerpAPI in #27826. That repair intentionally left Zendesk unchanged until separate failure evidence existed; this run provides that evidence.

## Invariant preservation

The probe still resolves a unique Zendesk variable base through guest DNS, establishes TLS to the real hostname, traverses the intended proxy/firewall path, injects the connector-owned header authentication, and requires the correct run-owned allowed firewall event. Raw UDP/TCP DNS behavior remains owned by the dedicated network-telemetry E2E.

## Validation

- `pnpm format`
- `pnpm turbo run lint --concurrency=1 --log-order grouped` (13/13 tasks)
- `pnpm knip`
- `pnpm turbo run check-types --concurrency=1 --filter='!api' --filter='!@okouai/app'` (12/12 tasks)
- `pnpm --filter @okouai/app run check-types`
- `pnpm --filter api run check-types`
- `e2e/test/libs/bats/bin/bats --count e2e/tests/03-runner/run-t03-firewall-zendesk.bats` (five consecutive passes; one test discovered each time)
[...]

PR #30623, opened 2026-08-31
test(e2e): pin the zendesk firewall probe to the public sink

e7h4n:
## Problem

Turbo run [33404787667](https://github.com/vm0-ai/vm0/actions/runs/33404787667) (`merge_group`, head `329a1df961008ab62804f71af3bc38f2c0c60a7f`, representing #30613) failed in [`cli-e2e-03-runner`](https://github.com/vm0-ai/vm0/actions/runs/33404787667/job/99532533410):

```
not ok 1 runner firewall resolves the Zendesk variable base and header auth # in 11492 ms
  `assert_output --partial "ZENDESK_REQUEST_SENT"' failed
# ZENDESK_API_TOKEN=zkTkn_...        <- resolved
# ZENDESK_EMAIL=runner-e2e@vm0.ai    <- resolved
# ZENDESK_SUBDOMAIN=e2e2089524288    <- resolved
# ZENDESK_REQUEST_FAILED=28
# curl: (28) Operation timed out after 5000 milliseconds with 0 bytes received
```

The identical merge-group head was re-run 12 minutes later as [33405942900](https://github.com/vm0-ai/vm0/actions/runs/33405942900) and the same test passed (`ok 1 ... # in 10223 ms`). Same SHA, same code, opposite outcome, so the failing input was external.

## First causal nondeterminism

The probe generated a fresh random subdomain per run and issued a live request to `https://<random>.zendesk.com/api/v2/users/me.json` with `--max-time 5`. That single budget had to cover DNS for a never-before-queried name, TLS to Zendesk's edge, and Zendesk's unknown-tenant lookup. `#28231` had additionally made the success marker conditional on curl's exit status and `exit`ed with it, so any slow third-party answer failed the test.

The variable-resolution half of the contract had already succeeded before the timeout: all three connector variables were injected and printed. Only the third party's response time was missing.

## Fix

Pin the destination to the public sink that the sibling probes `run-t01-firewall-twilio.bats`, `run-t02-firewall-discord-webhook.bats` and `run-t05-firewall-permission-refresh.bats` already use, and report the curl status next to the marker instead of gating the marker on it.

The product contract is unchanged and still asserted:

- `--resolve` only fixes the connect address. The request host stays `<subdomain>.zendesk.com`, which is what the firewall matches, so variable-base resolution is still exercised end to end.
- `runner_e2e_wait_for_firewall_log "$RUN_ID" zendesk "${subdomain}.zendesk.com" '["ZENDESK_API_TOKEN"]'` is untouched. It is what proves the contract in the test title: the firewall resolved the templated base, classified the flow as `ALLOW` with no `firewall_error`, and injected exactly `ZENDESK_API_TOKEN`.
- `--ipv4` is dropped because `--resolve` removes the name resolution it was hardening. The IPv6 concern `#28231` addressed no longer has a path to occur.
- The removed `assert_output --partial "ZENDESK_REQUEST_SENT"` is not lost coverage: `runner_e2e_wait_for_chat_text` now waits on `ZENDESK_REQUEST_SENT` itself and returns non-zero if the run reaches a terminal status without it.
- `curl` without `--fail` exits 0 on any HTTP status, including a firewall `403`, so its exit status never carried an allow/deny signal. Denial is caught by the firewall-log assertion.

No retry, sleep, skip, weakened business assertion, or timeout increase. `--max-time 5` is unchanged.

## Validation

Local `bats` execution of `tests/03-runner` needs a provisioned runner E2E environment and connector credentials, so it was not run here; the PR pipeline owns it. Validated locally: rendered the substituted probe and confirmed `--resolve` receives the same host as the request URL, `bash -n` syntax, that `curl ... || curl_status=$?` under `set -euo pipefail` always completes and always emits the marker with the real status, and `git diff --check`.

There is no app, API, or browser surface in this change, so no preview validation applies.

## Related, not fixed here

`e2e/tests/03-runner/run-t04-firewall-serpapi.bats` has the identical shape from `#27826` (live `serpapi.com` request inside `--max-time 5`, marker gated on curl's exit status). It has not failed in the window reviewed, so it is left out of this focused change and noted for follow-up.

### DIFF
--- a/e2e/tests/03-runner/run-t03-firewall-zendesk.bats
+++ b/e2e/tests/03-runner/run-t03-firewall-zendesk.bats
@@ -39,17 +39,16 @@ set -euo pipefail
 printf 'ZENDESK_API_TOKEN=%s\n' "$ZENDESK_API_TOKEN"
 printf 'ZENDESK_EMAIL=%s\n' "$ZENDESK_EMAIL"
 printf 'ZENDESK_SUBDOMAIN=%s\n' "$ZENDESK_SUBDOMAIN"
-# Raw DNS has dedicated runner coverage. Keep this firewall-auth probe on IPv4
-# so an unavailable AAAA response cannot block an otherwise valid request.
-if curl --ipv4 --silent --show-error --max-time 5 \
+# Raw DNS has dedicated runner coverage. Pin the public sink so this test owns
+# only firewall classification and authentication resolution, never whether a
+# live Zendesk tenant answers. The request host still carries the resolved
+# variable base, which is what the firewall matches on.
+curl_status=0
+curl --silent --show-error --max-time 5 \
+    --resolve '__SUBDOMAIN__.zendesk.com:443:8.8.8.8' \
     --output /dev/null \
-    "https://__SUBDOMAIN__.zendesk.com/api/v2/users/me.json"; then
-    printf 'ZENDESK_REQUEST_SENT\n'
-else
-    curl_status=$?
-    printf 'ZENDESK_REQUEST_FAILED=%s\n' "$curl_status"
-    exit "$curl_status"
-fi
+    "https://__SUBDOMAIN__.zendesk.com/api/v2/users/me.json" || curl_status=$?
+printf 'ZENDESK_REQUEST_SENT=%s\n' "$curl_status"
 EOF
 )
     prompt="${prompt//__SUBDOMAIN__/$subdomain}"
@@ -63,12 +62,11 @@ EOF
     echo "$output"
     assert_success
 
-    # Both outcomes use this prefix so a transport failure surfaces without
-    # waiting for a success marker that can no longer be emitted.
-    run runner_e2e_wait_for_chat_text "$THREAD_ID" "$RUN_ID" ZENDESK_REQUEST_
+    # The marker carries the curl status so a transport anomaly stays visible
+    # without turning sink reachability into the assertion under test.
+    run runner_e2e_wait_for_chat_text "$THREAD_ID" "$RUN_ID" ZENDESK_REQUEST_SENT
     echo "$output"
     assert_success
-    assert_output --partial "ZENDESK_REQUEST_SENT"
     assert_output --partial "ZENDESK_API_TOKEN=zkTkn_CoffeeSafeLocalCoffeeSafeLocalCoffeeSa"
     assert_output --partial "ZENDESK_EMAIL=runner-e2e@vm0.ai"
     assert_output --partial "ZENDESK_SUBDOMAIN=${subdomain}"


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: c92f955d83b06a0a6f0687ff7f2d0b0e111c7bb7
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Pins the Zendesk runner firewall probe to the public sink with `--resolve …:443:8.8.8.8` and reports the curl status next to the marker instead of gating the marker on it, removing a live third-party dependency that was failing the test nondeterministically. The firewall contract assertion (`runner_e2e_wait_for_firewall_log … zendesk "${subdomain}.zendesk.com" '["ZENDESK_API_TOKEN"]'`) is unchanged.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no wrong assumptions or unhandled edge paths | Pass | Rendered probe verified: `--resolve` host matches the request URL host after `${prompt//__SUBDOMAIN__/…}` substitution. `curl … \|\| curl_status=$?` under `set -euo pipefail` always completes and always emits the marker. If the sandbox never issues the request, `runner_e2e_wait_for_firewall_log` still fails, so no false pass is introduced. |
| `REVIEW.md` | Security: no hardcoded secrets or credentials | Pass | No credential material added. `8.8.8.8` is a public anycast address, matching `run-t01`, `run-t02`, `run-t05`. |
| `REVIEW.md` | Type safety / `any` casts | N/A | BATS shell test only; no TypeScript in the diff. |
| `REVIEW.md` | Testing coverage for the commit type | Pass | `test:` commit. No `feat:`/`fix:` behavior change, so no new regression test is owed; the diff is itself the test change. |
| `REVIEW.md` | Feature regression guard: an existing behavior replaced needs proof the replacement is intentional and scoped | Pass | The removed `assert_output --partial "ZENDESK_REQUEST_SENT"` is not lost coverage: `runner_e2e_wait_for_chat_text` now waits on `ZENDESK_REQUEST_SENT` itself and returns non-zero if the run reaches a terminal status without it. `curl` without `--fail` exits 0 on any HTTP status including a firewall `403`, so its exit status never carried an allow/deny signal; denial is still caught by the unchanged firewall-log assertion. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, contract, or persisted shape in the diff. |
| `docs/bad-smell.md` | Fail fast; no fallback/recovery logic that hides a real failure | Pass | `\|\| curl_status=$?` does not recover from or hide a failure: the status is printed into sandbox-visible output on every run, and the product assertion moved to the firewall log rather than being dropped. |
| `docs/bad-smell.md` | No lint/type suppressions | Pass | None added. |
| `docs/bad-smell.md` | No hardcoded environment-specific configuration | Pass | The pinned sink is a test-fixture destination, matching the three sibling probes; the connector base still comes from the resolved `ZENDESK_SUBDOMAIN` variable. |
| `docs/fallback.md` | Does the diff introduce a fallback needing a `Fallbacks` section? | Pass | No. The diff adds no `??`/`\|\|` default on a contract-owned field, no compatibility branch for an old client/runner/route/persisted shape, and no feature-switch-gated path. `\|\| curl_status=$?` is probe instrumentation, and with the sink pinned a non-zero curl status is the expected state, not a state the system is not supposed to produce. Per section 9, a PR that introduces none needs neither the section nor `Fallbacks: none`. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | None added; the diff removes an assertion rather than adding a tombstone. |
| `docs/testing.md` | E2E covers the happy path; error and edge cases belong in integration tests | Pass | The probe keeps a single happy-path shape and no longer encodes a third-party error case as a test outcome. |
| `docs/testing.md` | Mock boundary, MSW, real DB, fake timers, mock-cleanup rules | N/A | BATS E2E against the deployed preview; no Vitest mocks, timers, or database access in the diff. |
| `docs/testing/cli-e2e-testing.md` | Assert sandbox-visible output and vm0-owned telemetry; do not treat an external provider's exact response status or body as the test oracle | Pass | This is the rule the previous code violated: `assert_output --partial "ZENDESK_REQUEST_SENT"` made a live Zendesk response within 5 s the oracle. The oracle is now vm0-owned telemetry (the firewall network log) plus sandbox-visible variable output. |
| `docs/testing/cli-e2e-testing.md` | Runner BATS is CI-only; validate through the PR pipeline, not a local run | Pass | PR body states local `bats` was not run and the pipeline owns validation, which matches this guide. |
| `docs/testing/cli-e2e-testing.md` | Use unique per-file connector slugs and self-contained setup/teardown so the shard planner can move the file | Pass | `setup()`/`teardown()` and the `zendesk` slug are untouched; the diff does not add cross-file ordering. |

### Findings

#### Critical (P0)
- None.

#### High Priority (P1)
- None.

### Testing

- Coverage: Adequate. The change preserves the assertion that proves the test's title — the firewall resolved the templated base, classified the flow `ALLOW` with no `firewall_error`, and injected exactly `ZENDESK_API_TOKEN` — and removes only the assertion about a third party's availability. Verified in `builtin_host_policy.py` that `validate_credentialed_builtin_request_destination` checks the trusted *host* for `providerOwned` and only rejects non-public IPs for `publicDestination`, so pinning to a public sink passes either policy kind without changing what is asserted.
- Conventions: Compliant. Matches the established sink-pinning shape in `run-t01-firewall-twilio.bats`, `run-t02-firewall-discord-webhook.bats`, and `run-t05-firewall-permission-refresh.bats`; the twilio probe adopted it on 2026-08-26, after this file's 2026-08-19 change, so this aligns t03 with the newer convention.
- Evidence of the flake: the identical merge-group head `pr-30613-664d4f78…` failed in [run 33404787667](https://github.com/vm0-ai/vm0/actions/runs/33404787667) and passed 12 minutes later in [run 33405942900](https://github.com/vm0-ai/vm0/actions/runs/33405942900) with no code difference.
- Observation, not a finding: four runner probes now share the `8.8.8.8:443` sink, so an outage there would fail them together. That coupling is pre-existing in the convention this PR adopts, not introduced by this diff.
- Noted for follow-up, out of scope here: `run-t04-firewall-serpapi.bats` still has the identical shape from #27826 — a live `serpapi.com` request inside `--max-time 5` with the marker gated on curl's exit status. The PR body records it.

PR #30339, opened 2026-08-30
feat(platform): share indicators through shared worker

e7h4n:
## Summary

- validate SharedWorker heartbeats through `/api/auth/me`, derive the trusted user/org credential, and isolate each credential in its own ccstate `Store` with an owned root `AbortController`
- bridge the existing `chatThreadIndicators$` computed through the SharedWorker while preserving the app-facing signal and business consumers
- make `bootstrap$` own the finite authenticated startup order: initialize Clerk, fetch the first token, await Worker identity validation and heartbeat ack, establish App realtime readiness, and only then allow initial route setup
- keep `setupRouter()` free of authenticated startup side effects; run perpetual App subscriptions, invalidation processing, and heartbeat polling from the application root signal after finite bootstrap completes
- start the existing realtime subscriptions inside each credential Store, invalidate the Worker computed before fanning out indicator invalidations to registered tabs, and keep each tab's existing realtime subscriptions intact
- remove user/org scope from the wire protocol, remove transport cancellation/disconnect messages, and route verified messages through module commands
- make the existing API client, auth recovery, realtime, preview bypass, force-upgrade handling, and `accept` path usable in both app and Worker runtimes without Worker toasts
- keep the SharedWorker host Store-scoped so tests replace only the host with a real `MessageChannel` while exercising the production token, bridge, heartbeat, and routing chain
- add `orgId` to `/api/auth/me` so the Worker can establish a trusted credential boundary

## Verification

- `pnpm format`
- `pnpm turbo run lint --concurrency=1 --log-order grouped` (14/14)
- `pnpm knip`
- `pnpm turbo run check-types --concurrency=1 --filter='!api' --filter='!@okouai/app'`
- `pnpm --filter @okouai/app run check-types`
- `pnpm --filter api run check-types`
- 9 focused App Vitest files covering bootstrap/route ordering, bootstrap telemetry and locale behavior, the browser bridge, reconnecting transport, MessagePort protocol, real MessageChannel host integration, Worker runtime/auth recovery, and Agents indicator UI (84 tests passed on head `23f5820`)
- `pnpm --filter @okouai/app build` with production Clerk placeholders; the SharedWorker DOM/build gate passed

## Local test limitation

- The two changed API BDD files require PostgreSQL. This environment has no listener on `localhost:5432` and no `pg_ctl` binary, so their runtime execution is deferred to PR CI; API lint and all API type-check stages passed locally.

PR #30624, opened 2026-08-31
test(platform): gate event-sourced rename on rendered readiness

e7h4n:
## Problem

`chat lifecycle > renames the event-sourced current chat while thread detail is pending` intermittently fails with a bare `Error: Test timed out in 5000ms.` and no assertion detail.

- Triggering run (main, push): https://github.com/vm0-ai/vm0/actions/runs/33402961171 — job https://github.com/vm0-ai/vm0/actions/runs/33402961171/job/99524310146, SHA `a982b33938dd7c44a1471006d8e0afadc3f8dd27`, `test-app` shard 5, observed duration 5042 ms.
- Same-SHA contrast: the merge-group run on the identical SHA (https://github.com/vm0-ai/vm0/actions/runs/33401983863) reported success, but `test-app` was **skipped** there, so it is not a code-equivalent contrast. The real contrast is same-day green runs of the same shard where this test took 741 ms / 903 ms / 934 ms / 1317 ms / 1416 ms.
- Two further occurrences today, on unrelated branches: https://github.com/vm0-ai/vm0/actions/runs/33387371323 (11:34) and https://github.com/vm0-ai/vm0/actions/runs/33395211384 (13:10). All three occurrences are on trees that already contained #30339; no occurrence was found on trees without it.

## First causal nondeterminism

Every remaining wait in this test is a `waitFor` / `findBy*`, bounded by Testing Library's default 1000 ms `asyncUtilTimeout` (the repo sets no override), and each fails with a descriptive message. A bare 5000 ms Vitest budget exhaustion can therefore only come from an **unbounded** await.

The test had exactly one: `await setupPageAndWaitForContent(...)`, introduced into this test by #30339. It awaits `waitForPageContent`, a MutationObserver promise with no timeout that resolves only once the app skeleton is hidden — which now depends on the SharedWorker handshake and `initialEventsReady$`. This scenario deliberately makes `chatThreadByIdContract.get` return `never()`, so page bootstrap has no settled completion at all; `detachedSetupPage`'s own docstring covers precisely this case. Under load the wait was reproduced still pending past 2.5 s with the skeleton visible while the chat was already rendered.

#30339 also deleted this test's `waitFor(hasChannelSubscriptionOnChannel(...))` gate before `context.mocks.ably.triggerOnChannel(...)`. `triggerAblyChannelEvent` resolves the channel with `realtime.getExistingChannel(channelName)?.trigger(...)`, so a publish before the shared database channel is subscribed is dropped silently and the final title assertion can never satisfy.

## Excluded

- The 68 `[W][Realtime] ably catch-up failed ApiError: HTTP 500` warnings in the failing job are incidental: they come from `fetchBrowserSession` in `browser-session-block.ts` for tests that do not mock `browserContract.get`, they appear in passing tests throughout this file, and that catch-up runs in its own realtime payload loop.
- Budget headroom is not the cause: this test normally uses 15–28% of its 5000 ms budget, and the failing shard's file total (55.2 s) is in line with green runs (49.6 s / 53.2 s).
- `test-app` shards 2, 3, 6, 7, 8 were fail-fast cancellations and `ci-gate-turbo` is the aggregate gate; neither is a repair target.

## Fix

Test-only, scoped to this one scenario:

1. Use `detachedSetupPage` instead of `await setupPageAndWaitForContent`, keeping `sharedWorkerTestTransport: "message-port"`. Readiness is asserted through the already-present rendered-thread waits.
2. Restore the explicit shared database channel subscription gate before publishing `threadListChanged`.

No retries, sleeps, raised timeouts, skipped or deleted tests, weakened assertions, manual detached cleanup, or swallowed rejections. The product contract — an event-sourced current chat can be renamed while its thread detail is still pending — and every assertion are preserved.

## Validation

- `npx vitest run src/views/okou-page/__tests__/chat-history-navigation.test.tsx` — 5 consecutive runs, 30/30 passing each time.
- Prettier, ESLint (`--max-warnings 0`), oxlint on the changed file, and `git diff --check` all clean.
- Deployed-preview validation is not applicable: the change touches only a Vitest file and no app, API, or runtime behavior.

### DIFF
--- a/turbo/apps/platform/src/views/okou-page/__tests__/chat-history-navigation.test.tsx
+++ b/turbo/apps/platform/src/views/okou-page/__tests__/chat-history-navigation.test.tsx
@@ -1090,7 +1090,10 @@ describe("chat lifecycle", () => {
       },
     );
 
-    await setupPageAndWaitForContent({
+    // Thread detail never responds in this scenario, so page bootstrap has no
+    // settled completion to await. Readiness comes from the rendered thread
+    // instead, per the `detachedSetupPage` contract.
+    detachedSetupPage({
       context,
       path: `/chats/${EVENT_SOURCED_RENAME_THREAD_ID}`,
       sharedWorkerTestTransport: "message-port",
@@ -1124,6 +1127,15 @@ describe("chat lifecycle", () => {
     });
 
     expect(document.title).toBe(`${originalTitle} | VM0`);
+    // A publish on an unsubscribed channel is dropped silently, so wait for the
+    // shared database subscription before triggering the thread list refresh.
+    await waitFor(() => {
+      expect(
+        context.mocks.ably.hasChannelSubscriptionOnChannel(
+          SHARED_DATABASE_REALTIME_CHANNEL,
+        ),
+      ).toBeTruthy();
+    });
     context.mocks.ably.triggerOnChannel(
       SHARED_DATABASE_REALTIME_CHANNEL,
       "threadListChanged",


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 380d639652e29397c9245f27273322dcb846e083
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Test-only change to one scenario in `chat-history-navigation.test.tsx`. It replaces the awaited `setupPageAndWaitForContent` with `detachedSetupPage` (readiness now comes from the already-present rendered-thread waits) and restores an explicit shared-database Ably channel subscription gate before publishing `threadListChanged`. No production code, contract, schema, or feature switch is touched.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic: no races, no wrong assumptions | Pass | The `detachedSetupPage` scenario keeps `chatThreadByIdContract.get` on `never()`, so bootstrap has no settled completion; removing the unbounded await eliminates the timeout-less wait. The added gate removes the dropped-publish race described below. |
| `REVIEW.md` | Security: no injected input, secrets, or missing auth | N/A | Diff is one Vitest file with no credentials, network surface, or auth path. |
| `REVIEW.md` | Type safety: no `any`, no over-broad types | Pass | No type annotations added; `detachedSetupPage` takes the identical `Parameters<typeof setupPage>[0]` options object, so `sharedWorkerTestTransport: "message-port"` still type-checks. |
| `REVIEW.md` | Style and maintainability: comments explain WHY | Pass | Both added comments state the reason (bootstrap has no settled completion; an unsubscribed-channel publish is dropped), not the mechanics. |
| `REVIEW.md` | Testing coverage by commit type | Pass | `test:` commit. No `feat:`/`fix:` production change, so no new regression test is required; the scenario itself retains every original assertion (rename request payload, rendered title, deferred `document.title`). |
| `REVIEW.md` | Feature regression guard: existing behavior not silently replaced | Pass | The product story — an event-sourced current chat can be renamed while its thread detail is still pending — is unchanged. No assertion was weakened, removed, or rewritten to a different expectation. |
| `REVIEW.md` | Event sourcing / optimistic UI semantics | N/A | No event, projection, reconciliation, or rollback code changes; `docs/event-sourcing.md` was therefore not in scope for this diff. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No `turbo/packages/db/src/schema/**`, contract, or `jsonb(...).$type<...>()` change. |
| `docs/bad-smell.md` | Zero `any`, zero lint/type suppressions | Pass | No `any`, `@ts-expect-error`, `eslint-disable`, `oxlint-disable`, or `prettier-ignore` added. |
| `docs/bad-smell.md` | No dynamic `import()`; no hardcoded URLs or config | Pass | No imports added; the reused `SHARED_DATABASE_REALTIME_CHANNEL` constant already existed in the file. |
| `docs/bad-smell.md` | Fail fast, no recovery/fallback logic | Pass | The diff adds no defaulting or recovery branch; it removes a silent-hang path in favour of a bounded, self-describing failure. |
| `docs/fallback.md` | No fallback introduced; declare any new one in the PR summary | Pass | The diff introduces no `??`/`||` default, compatibility branch, or version-tolerance path, so no `Fallbacks` section is required and none is present. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | No tombstone assertion added; the pre-existing `queryByText("Thread detail should stay pending")` check is untouched and asserts live pending-state behavior, not deleted code. |
| `docs/testing.md` | Integration/page tests over unit tests; mock only at the system boundary | Pass | Stays a page-level integration test; no `vi.mock()` added, no relative-path mock, no `globalThis.services.db` mock. |
| `docs/testing.md` | No fake timers, no artificial delays; fix floating promises rather than patching symptoms | Pass | No `vi.useFakeTimers()`, `advanceTimersByTime`, `setTimeout`, retry, or raised `testTimeout`. The fix is an explicit readiness barrier plus removal of an unbounded await. |
| `docs/testing.md` | No manual detached cleanup in a test body | Pass | No `clearAllDetached()` or file-level `afterEach`; cleanup stays owned by `testContext()` and global teardown. |
| `docs/testing/app-testing.md` | Page tests must enter through `detachedSetupPage` and wait for the rendered state that matters | Pass | This is exactly the change: the scenario now calls `detachedSetupPage` without awaiting and gates on `findByLabelText("Chat thread")` plus the title `waitFor`. |
| `docs/testing/app-testing.md` | Configure mocks through `context.mocks` before setup | Pass | All `context.mocks.api(...)` overrides remain above the setup call; ordering is unchanged by the diff. |
| `docs/testing/patterns.md` | Pattern 7: use `detachedSetupPage()` for view tests; do not render platform components directly | Pass | Moves the scenario onto the documented view-test entry point; no direct component render is introduced. |
| `docs/testing/anti-patterns.md` | AP-5 fake timers / AP-10 direct component rendering | Pass | Neither appears; AP-10 is specifically improved by returning to the production bootstrap entry helper. |
| `docs/testing/anti-patterns.md` | AP-1/AP-7: assert behavior, not mock calls or implementation details | Pass | The outcome assertions stay page-visible (`document.title`, rendered title). The new `hasChannelSubscriptionOnChannel` check is a synchronization barrier before an external stimulus, not the scenario's assertion — see the note below. |
| `docs/testing/testing-external-behavior.md` | Construct cases and verify results through the page surface | Pass | Verification remains `document.title` and rendered thread text. The channel publish is simulated server input, which this test already modelled via `triggerOnChannel`; the gate only sequences that input. |
| `docs/effect.md` | Choose the owner and trigger before the API; every async process needs an `AbortSignal` with an eventual abort path | Pass | No command, effect, `onRef`, or render path changes. `detachedSetupPage` routes the setup promise through `detach(..., Reason.Entrance)` inside the test helper, and `context.signal` remains the owner. |
| `docs/cache.md` | RC-8 asymmetric resource lifecycles; no unowned long-running work | Pass | No listener, timer, observer, or subscription is created by the diff; the removed `waitForPageContent` wait was the only unowned unbounded await and it is gone. |
| `.claude/skills/ccstate/SKILL.md` | In-run test flakiness is a floating-promise/synchronization bug — fix it with the missing `await`, parent signal, or explicit domain-level synchronization point, not waits, manual clears, or extra `detach()` | Pass | The fix adds one explicit readiness barrier on an observable condition and removes an unbounded await; it adds no sleep, retry, `clearAllDetached()`, or extra `detach()` call. |
| `.claude/skills/ccstate/SKILL.md` | `detach()` belongs to the views/DOM-callback layer, never the signals layer | Pass | The diff adds no `detach()`; the one in `detachedSetupPage` lives in `src/__tests__/page-helper.ts`, outside `signals/`. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Notes (non-blocking)

- `chat-history-navigation.test.tsx:1130` — the new `waitFor(hasChannelSubscriptionOnChannel(...))` re-introduces a realtime-mock state check into a page test, which is the shape PR #30339 deliberately removed from these scenarios. It reads acceptable here because it sequences a simulated server publish rather than asserting the story, `triggerAblyChannelEvent` drops a publish on an unresolved channel silently (`realtime.getExistingChannel(channelName)?.trigger(...)`), and the same barrier is already the established pattern in `shared-database/__tests__/worker-runtime.test.ts` and `shared-database/__tests__/platform-direct-bridge.test.ts`. Worth revisiting only if a page-visible readiness signal for the shared database subscription becomes available.
- Removing the awaited setup makes `findByLabelText("Chat thread")` the first bounded wait, at Testing Library's default 1000 ms. That is the same structure this scenario used before #30339, where it ran green at 903–934 ms for the whole test, so the margin is real but not generous. If this scenario ever fails on that specific query instead of the 5000 ms budget, the right response is a page-visible readiness signal, not a larger timeout.

### Testing

- Coverage: Adequate. `test:` commit with no production change; the modified scenario keeps all original assertions and the PR reports 5 consecutive green runs of the full file (30/30 each).
- Conventions: Compliant. No internal `vi.mock()`, no direct `fetch` stubbing, no `globalThis.services.db` mock, no fake timers, no `toHaveBeenCalled()`-only assertion, no manual mock cleanup, no skipped or deleted test. Entry point now matches the `detachedSetupPage` rule in `docs/testing/app-testing.md` and `docs/testing/patterns.md` Pattern 7.
- Residual risk: the diff changes only this one scenario; the other five `setupPageAndWaitForContent` call sites in the file keep the awaited form, which is correct since their thread-detail mocks do resolve.

PR #24109, opened 2026-07-30
perf(runner): prevent concurrent systemd daemon-reload storms

seven332:
## Summary

- add a stable host-global systemd reload coordinator that rechecks
  `LoadState`, `NeedDaemonReload`, and expected VM0 drain drop-in state under
  an exclusive flock
- batch runner service enablement changes with `--no-reload`, preserving
  install, uninstall, drain, resume, stop, cleanup, and rollback ordering
- preserve the shared lock across runner GC, align stale-runner Ansible
  cleanup, and add a real-metal concurrent reload-budget scenario

## Behavior

The existing per-unit service lock remains the outer lifecycle boundary. A
caller mutates its unit or drop-in, then briefly acquires the host-global lock
to query systemd and perform a reload only when the mutation is still dirty or
its expected effective state is not loaded. Concurrent different-name
operations can therefore share a completed manager reload instead of merely
queueing equivalent reload requests.

Install additionally reloads while a new unit remains `not-found`. Drain
drop-in additions and removals compare the exact VM0 path against systemd's
loaded `DropInPaths`; this covers transient units for which systemd 255 can
report `NeedDaemonReload=no` before the new drop-in is effective. Query failure
falls back to an unconditional reload. Bounded failed-start cleanup uses
bounded waits for the global lock and systemd commands. Failure paths restore
VM0-owned drain state and prior boot enablement, including preserving
runtime-only enablement, or return combined primary and rollback context.

## Compatibility

This changes no API, queue payload, database schema, persisted runner state,
guest contract, or runner/backend protocol. Old runner binaries remain
state-safe during rollout but can still issue redundant reloads until all
participating lifecycle processes contain the coordinator.

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy -p runner --all-targets --all-features -- -D warnings`
- `cargo doc -p runner --all-features --no-deps`
- `cargo test -p runner --bin runner cmd::service`
- `cargo test -p runner --bin runner cmd::gc::orphaned_locks`
- `cargo test -p runner --bin runner` (2,889 passed, 14 ignored child fixtures)
- `bash -n` for the changed runner behavior scripts
- `shellcheck` for the changed runner behavior scripts
- `.github/scripts/check-workflow-shell-compatibility.sh .github/workflows/crates.yml`
- `actionlint .github/workflows/crates.yml`
- `ansible-playbook -i localhost, ansible/playbooks/cleanup-crates-runner.yml --syntax-check`
- `git diff --check`
[...]

PR #30649, opened 2026-08-31
fix(runner): treat systemd daemon-reload notice as advisory

e7h4n:
## Problem

Turbo run [33409335059](https://github.com/vm0-ai/vm0/actions/runs/33409335059) (merge_group, `pr-30618`, SHA `3ba6746e3eef82093f1f790b86d62fe97bf8cd4b`) failed in [`deploy-runner-start`](https://github.com/vm0-ai/vm0/actions/runs/33409335059/job/99545498447) with `Runner start failed on dev-1.aws.vm3.ai`.

The ordering in the host transcript is the whole story — the runner had **already started successfully** before the job went red:

```
15:38:28.062  INFO runner::cmd::service: transient service started unit=vm0-runner-pr-30618-1
              Runner ready on dev-1.aws.vm3.ai: max_concurrent=47
15:38:34.833  error: internal error: systemctl cat vm0-runner-pr-30618-1.service emitted stderr:
              "# Warning: vm0-runner-pr-30618-1.service changed on disk, the version systemd has
               loaded is outdated.\n# This output shows the cu...[truncated]"
```

`runner service start` succeeded, `runner service wait-running` reported ready, and the *next* step — `runner doctor --name pr-30618-1` in `.github/scripts/reconcile-and-start-runner-groups.sh` — failed. The second host completed the identical sequence cleanly (`Runner ready on dev-11.gcp.vm3.ai`, `Runners (1 running, 0 stopped)`, `0 warning(s) found`), so this is not a host outage and not the known `mode=starting` saturation pattern.

## Root cause

`unit_content_from_systemctl_cat` rejected **any** non-empty stderr from a *successful* `systemctl cat`:

```rust
if !output.status.success() { return Err(...); }
if !stderr.is_empty() { return Err(RunnerError::Internal(...)); }  // ← fired here
```

The error text proves the exit status was 0: the message came from the `!stderr.is_empty()` branch, which is only reachable after `status.success()`.

The stderr systemd wrote is its `NeedDaemonReload` notice. systemd emits it with `fprintf(stderr, ...)` and its own source comments it as informational output ([systemd#10974](https://github.com/systemd/systemd/issues/10974)); the exit status is unchanged and the selected unit content is still written to stdout, which is all `read_unit_config_path` consumes.

`NeedDaemonReload` is derived from mtime comparisons across a unit's fragment and drop-in set, so it is routinely raised on a shared host without this unit's own content changing, and it is not reliably clearable by running `daemon-reload`: see [systemd#17730](https://github.com/systemd/systemd/issues/17730) (relative drop-in mtime ordering keeps the flag set across reloads) and [systemd#35710](https://github.com/systemd/systemd/issues/35710) (units created by `systemd-run`, exactly how `runner service start` creates this transient unit, report `NeedDaemonReload=yes` since v230). The advisory is therefore a condition this job neither owns nor can repair, which is why it appears on one dev host and not the other for the same code at the same moment.

The same job already demonstrates the intended tolerance elsewhere: `systemctl stop` / `reset-failed` against a not-yet-existing unit produced `exit status: 5` and were correctly downgraded to `WARN ... during cleanup`. `systemctl is-enabled` likewise never fails on stderr. Only `systemctl cat` was intolerant.

## Fix

Reject only the stderr lines that are **not** part of systemd's comment block, and log the advisory rather than failing on it.

Deliberately **not** chosen: forcing a `systemctl daemon-reload` before `cat`. It would mutate manager-wide state on a shared multi-tenant dev host from a read-only diagnostic, contradicts the anti-storm design added in #24109, still races, and per systemd#17730 / #35710 would not even clear the condition for a transient unit.

Verification is not weakened:

- a non-zero exit status still fails, with or without the advisory;
- a real diagnostic (`Failed to get unit file path: ...`) still fails, including when mixed into the advisory block, and only the real line is reported;
- `runner service wait-running` remains the readiness gate and is untouched;
- ANSI-colorized advisories (`SYSTEMD_COLORS`) are handled so the same flake cannot return through a formatting variant.

## Validation

- `cargo fmt -p runner -- --check` — pass
- `cargo clippy -j 1 -p runner --all-targets` — pass (compiles the new `#[cfg(test)]` code under the workspace `warnings = "deny"` lint)
- `git diff --check` — clean
- The `runner` test binary could not be linked in the repair sandbox (rustc was OOM-killed at 3.58 GiB RSS in a 3.9 GiB cgroup), so the five new/existing `systemctl cat` cases were executed by extracting the changed functions verbatim into a standalone harness: 7 tests, 5 consecutive runs, all green. The crate's own `cargo test` runs here in `crates`.
- The `deploy-runner-start` job on this PR exercises the original invariant end to end.

### DIFF
--- a/crates/runner/src/cmd/service/systemctl.rs
+++ b/crates/runner/src/cmd/service/systemctl.rs
@@ -9,6 +9,7 @@ use crate::bounded_command::{
 use crate::error::{RunnerError, RunnerResult};
 use serde::ser::SerializeStruct;
 use serde::{Serialize, Serializer};
+use tracing::warn;
 
 use super::diagnostic::status_field_preview;
 use super::target::RunnerServiceUnit;
@@ -899,17 +900,66 @@ fn service_restart_policy_from_systemctl_show(
     Ok(restart)
 }
 
+/// Strip the leading ANSI escape sequences systemd writes when it colorizes a
+/// log line, so advisory detection does not depend on `SYSTEMD_COLORS`.
+fn strip_leading_ansi(line: &str) -> &str {
+    let mut rest = line;
+    while let Some(after_escape) = rest.strip_prefix('\u{1b}') {
+        let Some(after_bracket) = after_escape.strip_prefix('[') else {
+            break;
+        };
+        let mut chars = after_bracket.char_indices();
+        let Some((idx, terminator)) = chars.find(|(_, c)| !matches!(c, '0'..='9' | ';')) else {
+            break;
+        };
+        rest = &after_bracket[idx + terminator.len_utf8()..];
+    }
+    rest
+}
+
+/// Classify one `systemctl cat` stderr line as systemd's advisory notice.
+///
+/// `systemctl cat` prints this notice with `fprintf(stderr, …)` when the unit
+/// reports `NeedDaemonReload`, and systemd itself documents the block as
+/// informational: the exit status stays 0 and the selected unit content is
+/// still written to stdout. `NeedDaemonReload` is derived from mtime
+/// comparisons over the unit's fragment and drop-in set, so it is routinely
+/// raised without this unit's own content having changed, and a
+/// `daemon-reload` does not reliably clear it — see systemd issues 10974,
+/// 17730 and 35710. Failing a read on it therefore rejects healthy output for
+/// a condition the caller neither owns nor can repair.
+///
+/// systemd only uses the `#` comment form for that notice; real `cat`
+/// diagnostics are plain sentences, so they still fail the read.
+fn is_systemctl_cat_advisory_line(line: &str) -> bool {
+    strip_leading_ansi(line.trim_start())
+        .trim_start()
+        .starts_with('#')
+}
+
 fn unit_content_from_systemctl_cat(svc: &str, output: &Output) -> RunnerResult<String> {
     let stderr = String::from_utf8_lossy(&output.stderr);
     let stderr = stderr.trim();
     if !output.status.success() {
         return Err(systemctl_cat_status_error(svc, &output.status, stderr));
     }
-    if !stderr.is_empty() {
+    let unexpected = stderr
+        .lines()
+        .filter(|line| !line.trim().is_empty() && !is_systemctl_cat_advisory_line(line))
+        .collect::<Vec<_>>()
+        .join("\n");
+    if !unexpected.is_empty() {
         return Err(RunnerError::Internal(format!(
             "systemctl cat {svc} emitted stderr: {:?}",
-            status_field_preview(stderr)
+            status_field_preview(&unexpected)
         )));
+    }
+    if !stderr.is_empty() {
+        warn!(
+            unit = svc,
+            advisory = %status_field_preview(stderr),
+            "systemctl cat reported a pending systemd daemon-reload"
+        );
     }
 
     let stdout = std::str::from_utf8(&output.stdout).map_err(|e| {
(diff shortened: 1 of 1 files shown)


e7h4n:
Validation on this PR's exact head:

- `deploy-runner-start` [succeeded](https://github.com/vm0-ai/vm0/actions/runs/33441123095/job/99650737816): both hosts ran the full sequence — `transient service started` -> `Runner ready` -> `runner doctor` (`Runners (1 running, 0 stopped)`, `0 warning(s) found`) -> `=== Runner started ===`. This is the step that failed in run 33409335059.
- `Crates` [succeeded](https://github.com/vm0-ai/vm0/actions/runs/33441123108), executing the new `systemctl cat` cases in the crate's own test binary.

systemd did not raise the `NeedDaemonReload` advisory during this run, so this is evidence of no regression rather than a live reproduction; the advisory path itself is covered by the unit tests.

e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: a07ba7fc25cccb3d6d7b3d2633387e53e7b92e59
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

`unit_content_from_systemctl_cat` rejected any non-empty stderr from a **successful** `systemctl cat`, which turned systemd's informational `NeedDaemonReload` notice into a hard internal error and failed `runner doctor` after the runner had already reported ready (Turbo run 33409335059). The PR now rejects only stderr lines that are not part of systemd's `#` comment block, logs the advisory, and keeps failing on a non-zero exit status and on any real diagnostic. Single Rust file; no protocol, schema, or persisted-state surface is touched.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic: races, edge cases, API misuse | Pass | `strip_leading_ansi` slices on char boundaries (`idx + terminator.len_utf8()`), shrinks `rest` every iteration so the loop terminates, and fails closed (a malformed or unterminated escape stays non-`#` and is reported as unexpected stderr). Line classification is applied per line, so a real diagnostic mixed into the advisory block still errors. |
| `REVIEW.md` | Security: no secret exposure on new paths | Pass | Errors still report only stderr, never stdout; the existing `systemctl_cat_failure_does_not_expose_stdout` guard (`--token should-not-appear`) is untouched, and the new `warn!` logs only the bounded stderr preview via `status_field_preview`. |
| `REVIEW.md` | Type safety / no broad casts | N/A | Rust change; no `any`, casts, or type suppressions introduced. |
| `REVIEW.md` | Style and maintainability, comments explain WHY | Pass | Both new helpers are short; the doc comment explains why the notice is informational and why it is not this caller's condition to repair, rather than restating the code. |
| `REVIEW.md` | Feature regression guard / feature-switch gating | N/A | `fix:` to runner-internal systemd output parsing; no user-facing feature, UI, API, or flow is added, replaced, or rerouted. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No `packages/db`, `api-contracts`, Zod response schema, or `jsonb(...)` change. |
| `docs/bad-smell.md` | Fail fast; no fallback/recovery logic that hides problems | Pass | This narrows an over-strict check rather than adding recovery: a non-zero exit still fails (`systemctl_cat_rejects_failed_exit_with_advisory_only_stderr`), a real diagnostic still fails, and the tolerated condition is logged at `warn!` rather than swallowed. The strict check was not a working staleness guard either — systemd#10974 documents that `cat` does not warn when a mask or new fragment appears. |
| `docs/bad-smell.md` | Error handling: no over-engineered or unnecessary handling | Pass | Net reduction in error paths; no new `try`/catch-equivalent wrapping, and `RunnerError::Internal` is still the single failure representation. |
| `docs/bad-smell.md` | No hardcoded configuration or environment values | Pass | No URLs, env values, or hardcoded paths added; the classifier keys on systemd's `#` comment form, not on host- or environment-specific strings. |
| `docs/fallback.md` | Default is no fallback; only handle states the system can actually produce | Pass | The handled state is documented, currently reachable external behavior: systemd writes the notice with `fprintf(stderr, ...)` and its own source calls it informational output (systemd#10974), leaving the exit status at 0. This is `docs/fallback.md` §6.2 (genuinely optional external data), not a compatibility branch for a state the contract prevents. |
| `docs/fallback.md` | New fallbacks must be declared in the PR summary with surface, window, removal condition | N/A | The diff introduces no cross-version compatibility path, so no `Fallbacks` section is required (§9: PRs that introduce none need neither the section nor `Fallbacks: none`). |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | The four new tests assert live behavior of the current classifier (advisory accepted, colorized advisory accepted, failed exit rejected, real diagnostic rejected); none pins the absence of deleted code. |
| `docs/testing.md` | `fix:` commits must carry a regression test | Pass | `systemctl_cat_accepts_daemon_reload_advisory_stderr` uses the exact systemd block from the failing run and would fail against the pre-fix code. |
| `docs/testing.md` | Integration tests primary; unit tests only under the narrow exceptions | Pass | Falls under the documented exception for parsers with non-obvious invariants, and the same doc names `crates/runner` as a place where direct tests are appropriate. An integration-level alternative would need a live systemd manager held in `NeedDaemonReload` state, which is not reproducible in CI. |
| `docs/testing.md` | Deployment-compatibility coverage for deployable boundaries | N/A | The change is local interpretation of `systemctl` output inside one runner process. No request/response, runner protocol, queue payload, or persisted shape crosses a version boundary, so no old/new combination exists to test. |
| `docs/testing/rust-testing.md` | Inline `#[cfg(test)]` for module-internal logic not exposed publicly | Pass | `unit_content_from_systemctl_cat` is a private fn; the new cases extend the existing `#[cfg(test)] mod tests` in the same file alongside three pre-existing cases for the same function. |
| `docs/testing/rust-testing.md` | No `std::env::set_var`/`remove_var` in tests; no unsafe shared process state | Pass | The tests construct `std::process::Output` values in-process; no environment mutation, no child processes, no shared global state. |
| `docs/testing/rust-testing.md` | Use real infrastructure / temp dirs rather than mocking internals | Pass | No mocking framework is introduced; the tests feed real `Output` values built from `ExitStatus::from_raw`, which is the boundary this function consumes. |

### Findings

#### Critical (P0)
- None.

#### High Priority (P1)
- None.

Noted, non-blocking: `read_unit_config_path` feeds `gate.rs:216`, `drain_resume.rs:959`, `stop.rs`, `doctor.rs`, and `gc/image_refs.rs`, so tolerating the notice means those callers now proceed on `cat` output that systemd flagged as possibly stale. The residual risk is small and bounded: the only writer of an installed unit file (`service install`) holds the per-unit service lock and runs a coordinated `daemon-reload` immediately after writing, and `gate`/`stop`/`drain_resume` take that same lock; transient units started by `systemd-run` have a `/run/systemd/transient` fragment that nothing rewrites; and the drain drop-in only sets `Restart=`, so it cannot alter the parsed `--config` path. Worth remembering if a future change starts rewriting unit files outside that lock.

### Testing
- Coverage: Adequate. Four new cases cover the regression and its guards — advisory accepted, ANSI-colorized advisory accepted, non-zero exit still rejected even when stderr is advisory-only, and a real diagnostic mixed into the advisory block still rejected with only the real line reported. The three pre-existing `systemctl cat` cases (clean stderr, non-comment stderr rejected with bounded preview, stdout never exposed on failure) remain and still pass, which pins that the strictness was narrowed rather than removed.
- Conventions: Compliant. Inline `#[cfg(test)]` per `docs/testing/rust-testing.md`, no environment mutation, no mocking of internal code, assertions are on returned values and error text rather than on call counts.
- CI evidence on this head: `Crates` succeeded (https://github.com/vm0-ai/vm0/actions/runs/33441123108), executing these cases in the crate's own test binary, and `deploy-runner-start` succeeded on both hosts (https://github.com/vm0-ai/vm0/actions/runs/33441123095/job/99650737816), exercising the exact `runner doctor` step that failed in run 33409335059.

PR #30662, opened 2026-09-01
fix(locator): drop the preview counter and widen the band with its ticks

tongx-blip:
Two fixes to the conversation locator, both reported from screenshots.

## 1. The `n/m` counter is gone

The preview card showed `4/8` / `8/8`. That numbering is not stable: folding and expanding groups changes both the position and the total, so the same turn is numbered differently from one moment to the next. A counter that moves under the reader is worse than no counter, so the card now carries just the role, the avatar and the timestamp. `position` and `total` are dropped from `LocatorPreview` entirely.

## 2. The viewport band widens with the ticks it frames

The orange band marks the turns inside the viewport, but it kept its resting 32px while the ticks it frames magnified on hover — so the hovered bar grew straight out of the frame that is supposed to contain it.

The band now widens by exactly what its widest framed tick gained, written per pointer frame in `applyMagnification` alongside the tick widths. Same source, same frame, so the frame stays locked to the bars instead of trailing them through a CSS transition. One dimension changes, and it returns to 32px on `pointerleave` with the ticks.

## Test harness

The layout stub let `scrollTop` exceed the scrollable range — the thread pins itself to the bottom by assigning `scrollHeight`, so the measured viewport read as parked a whole screen past the last turn and `bandHeight` was `0` for the entire suite. The band was therefore never in the DOM and never testable. The stub now clamps the way a browser does.

## Checks

- `pnpm -F @okouai/app exec vitest run src/views/okou-page/__tests__/chat-conversation-locator.test.tsx` — 14 passed
- `tsc --noEmit`, `pnpm -F @okouai/app run lint`, `pnpm knip --workspace apps/platform`, `pnpm format` — clean

Behind the `ChatConversationLocator` feature switch.

### DIFF
--- a/turbo/apps/platform/src/signals/chat-page/chat-conversation-locator.ts
+++ b/turbo/apps/platform/src/signals/chat-page/chat-conversation-locator.ts
@@ -72,6 +72,13 @@ const TICK_METRICS = {
   assistant: { base: 12, grow: 2.5 },
 } as const;
 
+/**
+ * Resting width of the viewport band. The band grows by exactly as much as the
+ * widest tick it covers, so a magnified bar never spills out of the frame that
+ * is supposed to contain it.
+ */
+export const BAND_BASE_WIDTH_PX = 32;
+
 const GROUP_SELECTOR = '[data-role="user"], [data-role="assistant"]';
 const SCROLL_ANCHOR_SELECTOR = "[data-chat-scroll-anchor-event-id]";
 
@@ -111,9 +118,6 @@ export interface LocatorPreview {
   readonly text: string;
   /** ISO timestamp of the turn, or undefined when it carries none. */
   readonly createdAt: string | undefined;
-  /** 1-based, for the "12/32" counter. */
-  readonly position: number;
-  readonly total: number;
 }
 
 export interface ChatConversationLocatorSignals {
@@ -521,13 +525,28 @@ function tickNodes(rail: HTMLElement): HTMLElement[] {
   return [...rail.querySelectorAll<HTMLElement>("[data-locator-tick]")];
 }
 
-function resetTickWidths(rail: HTMLElement): void {
+/**
+ * The band is the frame around the ticks inside the viewport, so it widens by
+ * whatever its widest tick gained rather than by a growth of its own: one
+ * dimension, one source, and the bars stay enclosed at every magnification.
+ */
+function writeBandWidth(rail: HTMLElement, growth: number): void {
+  const band = rail.querySelector<HTMLElement>(
+    "[data-conversation-locator-band]",
+  );
+  if (band) {
+    band.style.width = `${(BAND_BASE_WIDTH_PX + growth).toFixed(2)}px`;
+  }
+}
+
+function resetRailWidths(rail: HTMLElement): void {
   for (const node of tickNodes(rail)) {
     const role: LocatorRole =
       node.dataset.locatorTick === "user" ? "user" : "assistant";
     node.style.width = `${TICK_METRICS[role].base}px`;
     delete node.dataset.locatorHot;
   }
+  writeBandWidth(rail, 0);
 }
 
 /**
@@ -545,6 +564,8 @@ function applyMagnification(
     MAGNIFY_SIGMA_MIN_PX,
     layout.pitch * MAGNIFY_SIGMA_RATIO,
   );
+  const bandBottom = layout.bandTop + layout.bandHeight;
+  let bandGrowth = 0;
   let nearest = -1;
   let nearestDistance = Number.POSITIVE_INFINITY;
   for (const [index, tick] of layout.ticks.entries()) {
@@ -561,7 +582,11 @@ function applyMagnification(
     const metrics = TICK_METRICS[tick.role];
     const width = metrics.base * (1 + weight * metrics.grow);
     node.style.width = `${width.toFixed(2)}px`;
+    if (tick.y >= layout.bandTop && tick.y <= bandBottom) {
+      bandGrowth = Math.max(bandGrowth, width - metrics.base);
+    }
   }
+  writeBandWidth(rail, bandGrowth);
 
   // Anywhere inside the group the nearest tick is at most half a pitch away,
   // so only overshooting the first or last tick misses.
@@ -707,7 +732,7 @@ function createPaint(
       return;
     }
     if (pointerY === null) {
-      resetTickWidths(rail);
+      resetRailWidths(rail);
       set(store.preview$, null);
       return;
     }
@@ -725,8 +750,7 @@ function createPaint(
       shown?.turnIndex === tick.turnIndex &&
       shown.role === tick.role &&
       shown.text === turn.text &&
-      shown.createdAt === turn.createdAt &&
-      shown.total === layout.turnCount
+      shown.createdAt === turn.createdAt
     ) {
       return;
     }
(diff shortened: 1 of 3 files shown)


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: 46ac845fea526fe7e82cf3e741b37bed86a4c505
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Removes the unstable `n/m` turn counter from the conversation locator's preview card, and makes the viewport band widen by exactly what its widest framed tick gained so a magnified bar can no longer spill out of the frame that marks the viewport. Also fixes the test harness's `scrollTop` stub to clamp like a browser, which is what makes the band observable in tests at all.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no off-by-one or wrong conditionals | Pass | `chat-conversation-locator.ts:585` band membership uses `tick.y` against `bandTop..bandTop+bandHeight`; both come from the same `origin`/`pitch` basis in `buildTicks`/`buildBand`, so the comparison is in one coordinate space. Containment holds for all inputs: band right edge is `7+32+g`, tick right edge is `14+w`, and `g = max(w-base)` gives `32-base >= 7` for both roles (`base` 7/12). |
| `REVIEW.md` | Type safety; no `any` or unjustified casts | Pass | No `any` added. `BAND_BASE_WIDTH_PX` is a typed const; `querySelector<HTMLElement>` is generic-typed. Test uses `as HTMLElement` after an explicit null assertion, matching the existing file. |
| `REVIEW.md` | `fix:` commit carries a regression test | Pass | `chat-conversation-locator.test.tsx:586` asserts rest width, growth locked to the widest framed tick, and reset on `pointerleave` — the exact defect reported. |
| `REVIEW.md` | Feature regression guard: changed behavior is contained and covered | Pass | Whole surface stays behind `FeatureSwitchKey.ChatConversationLocator`. The counter removal updates the existing preview assertion in place rather than deleting the test; the card path is still covered by `"Virtual question 0"`. |
| `REVIEW.md` | DB / JSONB persisted-contract gate | N/A | No schema, contract, Zod response, or `jsonb().$type<>()` change. `LocatorPreview` is an in-memory view model with one consumer. |
| `docs/bad-smell.md` | Zero `any`; zero lint/type suppressions | Pass | No suppression comments in the diff. |
| `docs/bad-smell.md` | No dynamic `import()`, no hardcoded config/URLs | Pass | Static imports only; the added constant is a layout metric, not environment config. |
| `docs/bad-smell.md` | Interface changes reviewed | Pass | `LocatorPreview` loses `position`/`total`; `createPaint` drops the now-dead `shown.total === layout.turnCount` memo term with it. `turnCount` is still live for `createPageWindow`/`createCanPageWindow`, so nothing is orphaned. |
| `docs/fallback.md` §3 | No fallback for a state the contract already prevents | Pass | `if (band)` in `writeBandWidth` guards a genuinely reachable `null`: the band is conditionally rendered behind `layout.bandHeight > 0`. Same shape as the existing `if (!node) continue` in the tick loop, not a `??` chain on a guaranteed field. |
| `docs/fallback.md` §1 | No negative/tombstone test asserting removed behavior stays removed | Pass | The PR deletes the `"1/30"` assertion instead of adding a "counter is gone" assertion — exactly the prescribed "update the existing test to the new behavior". |
| `docs/fallback.md` §2 | Non-GA feature switch needs no compatibility code | Pass | No dual-read, tolerance branch, or migration added for the locator's own earlier shape. |
| `docs/fallback.md` §9 | New fallbacks declared in the PR summary | N/A | The diff introduces no fallback, so no `Fallbacks` section is required and none is expected. |
| `docs/effect.md` | React render stays pure — no DOM mutation or Store writes in render | Pass | The tsx only adds a static `width: BAND_BASE_WIDTH_PX` to the style object. Every DOM write stays in `applyMagnification`/`writeBandWidth`, reached only through the `paint$` command. |
| `docs/effect.md` | Work runs under the boundary that owns it | Pass | Band width is written on the same pointer-frame path as tick widths (`paint$` via `scheduleMagnify`/`flushFrame`), not through a new effect or ref. |
| `docs/cache.md` RC-1 | No render-phase side effects | Pass | No command execution, listener registration, or `document` mutation added to render. |
| `docs/cache.md` RC-7 / RC-8 | Stable callback refs; symmetric acquire/release | Pass | `railOnRef$`/`previewOnRef$` wiring is untouched and still passed directly from `useSet`. The reset counterpart is added in the same place as the ticks': `resetRailWidths` now also restores the band. |
| `docs/cache.md` RC-3 | No unbounded-lifetime cache | Pass | No new module-level `Map`/`Set`/registry; band width is read back off the DOM node each frame. |
| `.claude/skills/ccstate/SKILL.md` | Shared logic must not leak `get`/`set` into plain helpers | Pass | `writeBandWidth(rail, growth)` takes only an element and a number. No accessor escapes the command callback, so `ccstate/no-getter-setter-params` and `no-accessor-escape` stay satisfied. |
| `.claude/skills/ccstate/SKILL.md` | DOM element refs go through `onRef` | Pass | Unchanged; the new code reaches the band by querying inside the already-owned rail element. |
| `docs/testing.md` | Integration/page-level tests; mock only at the system boundary | Pass | New test drives the real page; no `vi.mock` added, no direct `fetch` stubbing. |
| `docs/testing.md` | No fake timers | Pass | Uses `waitFor` and real event dispatch only. |
| `docs/testing/app-testing.md` | Page tests enter through `detachedSetupPage` | Pass | `renderLongThread` → `renderMeasuredThread` → `detachedSetupPage`; the harness change stays bound to `context.signal` via the existing restore hook. |
| `docs/testing/anti-patterns.md` AP-1 | No `toHaveBeenCalled()` as the assertion | Pass | Every assertion is on rendered geometry. |
| `docs/testing/anti-patterns.md` AP-4 | No `vi.mock()` on internal relative paths | Pass | None added. |
| `docs/testing/anti-patterns.md` AP-7 | No implementation-detail assertions | Pass | The assertions are on inline `style.width`/`style.top`, which for this instrument *is* the product surface — the rail has no text or role to observe, and the file already establishes this convention for tick widths and positions. Nothing internal (store, callbacks, component state) is asserted. See non-blocking note. |
| `docs/testing/testing-external-behavior.md` | Construct and observe through the page | Pass | The case is built by real pointer events on the rendered rail and observed from the rendered band; the store is never mutated directly. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Non-blocking notes

- `chat-conversation-locator.test.tsx:586` — the band assertions read inline styles, which is the closest AP-7 sits to being tripped. It is accepted here because the rail exposes no non-geometric surface and the file already asserts tick `style.width`/`style.top`, but a future move to `transform` or a CSS variable would need this test updated even with behavior unchanged.
- `chat-conversation-locator.test.tsx:79` — the clamp subtracts the `VIEWPORT_PX` constant rather than `this.clientHeight`. Equivalent by construction inside the `isScroller` branch, so this is coupling, not a defect.

### Testing

- Coverage: Adequate. The band-widening defect gets a direct regression test covering rest width, magnified width locked to the widest framed tick, and reset on pointer leave. The counter removal correctly updates the existing preview assertion instead of adding a tombstone, and the preview card path keeps coverage through the virtual-window test.
- Conventions: Compliant. Page-level entry via `detachedSetupPage`, no internal mocks, no fake timers, no file-level `afterEach`, no new mock-call assertions. The `stubLayout` `scrollTop` clamp raises harness fidelity — before it, `bandHeight` was `0` for the entire suite and the band was never in the DOM, so this change converts an untestable element into a tested one without weakening the existing scroll-position assertions (window start still resolves to the same tail-anchored index).

PR #30591, opened 2026-08-31
fix(email): link official workflow result footer to automation

lancy:
## Summary

- render the approved Official Workflow result footer with separate automation-management and account-unsubscribe links
- persist a brand-correct deep link to the originating workflow automation and highlight/scroll the matching row on the existing Automations page
- preserve one-click unsubscribe headers, existing switch semantics, Markdown safety, plain text, and size bounds
- cover multiple automations plus stale and unreadable deep-link destinations

## Fallbacks

- `official-automation-result-email-renderer.ts:officialAutomationResultEmailHtml` preserves the legacy footer for persisted outbox rows written by the previous API callback, whose `manageUrl` is still the account-level unsubscribe confirmation URL. Surface: persisted backend outbox across the API rollout and rollback window (up to about 102 minutes). Remove after the previous API artifact is no longer serving, draining, or retained for rollback and all pending or retryable pre-release rows have drained or expired; follow-up #30592.

## Verification

- `pnpm format`
- `pnpm lint`
- `pnpm knip`
- `pnpm turbo run check-types --concurrency=1 --filter="!api" --filter="!@okouai/app"`
- `pnpm --filter @okouai/app run check-types`
- `pnpm --filter api run check-types`
- `pnpm --filter api exec vitest run src/signals/services/__tests__/official-automation-result-email-renderer.test.ts` (25 passed)
- `pnpm --filter api exec vitest run src/signals/routes/__tests__/official-automation-result-email.test.ts` (11 passed)
- `pnpm --filter @okouai/app exec vitest run src/views/workflows-page/__tests__/workflows-page.test.tsx` (78 passed)

Closes #30569

PR #30699, opened 2026-09-01
test(e2e): retry runner api reads that receive no response

e7h4n:
## Problem

Merge-group Turbo run [33464652878](https://github.com/vm0-ai/vm0/actions/runs/33464652878) (queued PR #30591, SHA `0bc551a3`) failed **four** `cli-e2e-03-runner` shards with one identical signature:

| Job | Test | Run ID |
| --- | --- | --- |
| [99722651761](https://github.com/vm0-ai/vm0/actions/runs/33464652878/job/99722651761) | `runner firewall resolves Twilio auth without exposing raw credentials` | `e7cf959f` |
| [99722651791](https://github.com/vm0-ai/vm0/actions/runs/33464652878/job/99722651791) | `vm0 built-in real pi loop returns a successful answer` | `82a6729e` |
| [99722651870](https://github.com/vm0-ai/vm0/actions/runs/33464652878/job/99722651870) | `runner resolves an exact custom connector account on thread continuation` | `adeceb69` |
| [99722651915](https://github.com/vm0-ai/vm0/actions/runs/33464652878/job/99722651915) | `continued runner session refreshes replaced connector credentials` | `3f0f1595` |

All four failed on `GET /api/runs/<id>/context`:

```
curl: (28) Operation timed out after 30002 milliseconds with 0 bytes received
runner_api_curl failed: url=https://pr-30591-api.vm6.ai/api/runs/<runId>/context curl_status=28
```

### The run completed, then the context request hung with zero bytes

In the Twilio shard the agent run reached `status: "completed"` at 03:05:38 and the sandbox printed `TWILIO_ACCOUNT_SID`, `TWILIO_AUTH_TOKEN` and `TWILIO_REQUEST_SENT`. The preview API was creating and completing runs normally. The `/context` read issued moments later returned nothing for 30 seconds.

### The API answered — after the client gave up

The preview request log (`vm0-request-log-dev`, `host == 'pr-30591-api.vm6.ai'`) records exactly what happened:

| `_time` | path | status | duration |
| --- | --- | --- | --- |
| 03:05:39.392 | `GET /api/runs/:id/context` | **200** | **44357 ms** |
| 03:05:40.601 | `GET /api/runs/:id/context` | **200** | **43073 ms** |
| 03:05:44.006 | `GET /api/runs/:id/context` | **200** | **39784 ms** |
| 03:05:54.115 | `GET /api/runs/:id/context` | **200** | **29877 ms** |
| 03:06:19.125 | `GET /api/runs/:id/context` | 200 | 4579 ms |
| 03:06:35.141 | `GET /api/runs/:id/context` | 200 | 687 ms |

No other request on that deployment exceeded 3000 ms in the window, and job 99722651870's third test (`run-t24-built-in-provider-fallback.bats`) passed at 03:06:36 after making two asserted `/context` reads. So this was not a deployment outage: one endpoint's latency spiked past the harness's fixed 30s `--max-time` and recovered within a minute.

`GET /api/runs/:id/context` resolves its snapshot through an Axiom APL query (`runContext` in `turbo/apps/api/src/signals/services/run-detail.service.ts` → `queryAxiom`). Under four concurrent shards that dependency's tail latency reached 30-44s.

**PR #30591 is not implicated.** Its API changes are confined to `email-common.service.ts`, `internal-workflow-automation-result-email-callback.service.ts` and `official-automation-result-email-renderer.ts`, plus platform workflow pages. Nothing in it touches the `/api/runs/:id/context` path, the Axiom client, auth middleware, DB pooling or route registration.

## Change

`runner_api_curl` failed the whole bats test on a single zero-byte 30s stall with no retry, and reported it identically to a served API error.

- **Send the request again only when curl exits `28`** — the deployment produced no response at all. `--fail-with-body` reports a served HTTP error as `22`, so a genuine API error still fails on the first attempt. Blanket retrying on API error responses is not introduced.
- **Only for a repeatable request** — one carrying neither a method override (`-X`/`--request`) nor a body (`-d`/`--data*`/`-F`/`--form`/`-T`/`--upload-file`). A write whose response timed out may already have been applied, so writes are never repeated.
- **Buffer each attempt** so a partial first body is replaced rather than concatenated. The exit status is carried behind a marker because command substitution strips trailing newlines, keeping the emitted response byte-for-byte identical to the previous unbuffered `curl`.
- **Report a stall distinctly**, and point it at a Vercel log search *without* a status filter. A stalled request is recorded as `status:000`, so the existing status-filtered link finds nothing — it actively hid the 200/44357 ms rows above.

Product contracts are untouched: both `/context` responses are still fully asserted, so the runner firewall still has to resolve Twilio auth without exposing raw credentials, and the built-in pi loop still has to return a successful answer. No sleeps, skips, deleted tests, weakened assertions or swallowed failures.

## Validation

- `.github/scripts/tests/runner-api-curl-test.sh` extended with four cases (stall-then-success, stall on both attempts with the exact stderr, `-X POST` not repeated, `-d` without `-X` not repeated) — passes. This suite runs in CI via `Run workflow script tests` in `security.yml`.
- `bash -n` and `shellcheck -s bash -x` clean on both changed files.
- `git diff --check` clean.
- 26 additional throwaway behavioural checks against a stubbed `curl` covering partial-body replacement, header-only GETs, `--data-raw`/`--data-binary`/`--data-urlencode`, `--request PUT` and `-X PATCH`.
- The runner e2e shards themselves validate the invariant on this PR's own CI.

## Follow-up, not in this PR

`queryAxiomDirect` in `turbo/apps/api/src/signals/external/axiom.ts` calls `client.query(apl)` with no abort signal, so `/api/runs/:id/context` has no latency budget of its own; the cursor branch beside it already bounds itself with `AbortSignal.timeout`. A user opening a run's context would have waited 44 seconds here. That is an API-side reliability question and is deliberately left out of this test-harness fix.

### DIFF
--- a/.github/scripts/tests/runner-api-curl-test.sh
+++ b/.github/scripts/tests/runner-api-curl-test.sh
@@ -11,6 +11,7 @@ trap 'rm -rf "$tmp_dir"' EXIT
 stdout_file="${tmp_dir}/stdout"
 stderr_file="${tmp_dir}/stderr"
 expected_file="${tmp_dir}/expected"
+attempts_file="${tmp_dir}/curl-attempts"
 
 fail() {
     echo "FAIL: $1" >&2
@@ -36,6 +37,29 @@ assert_file_excludes() {
     local sensitive_value="$2"
     if grep -Fq "$sensitive_value" "$actual_file"; then
         fail "sensitive value appeared in ${actual_file}"
+    fi
+}
+
+# The mocked curl runs inside a command substitution, so its attempt counter
+# lives in a file rather than a shell variable.
+reset_curl_attempts() {
+    printf '0' >"$attempts_file"
+}
+
+record_curl_attempt() {
+    local attempts
+    attempts="$(<"$attempts_file")"
+    attempts=$((attempts + 1))
+    printf '%s' "$attempts" >"$attempts_file"
+    printf '%s' "$attempts"
+}
+
+assert_curl_attempts() {
+    local expected="$1"
+    local actual
+    actual="$(<"$attempts_file")"
+    if [[ "$actual" != "$expected" ]]; then
+        fail "expected ${expected} curl attempt(s), got ${actual}"
     fi
 }
 
@@ -96,6 +120,18 @@ curl() {
             echo "curl: (22) The requested URL returned error: 429" >&2
             echo "Vercel logs: $MOCK_CURL_EXPECTED_VERCEL_URL" >&2
             return 22
+            ;;
+        no-response | no-response-then-success)
+            local attempt
+            attempt="$(record_curl_attempt)"
+            if [[ "$MOCK_CURL_MODE" == "no-response-then-success" ]] &&
+                ((attempt > 1)); then
+                printf '{"ok":true}\n'
+                return 0
+            fi
+            echo "curl: (28) Operation timed out after 30002 milliseconds with 0 bytes received" >&2
+            echo "Vercel logs: $MOCK_CURL_EXPECTED_VERCEL_URL" >&2
+            return 28
             ;;
         no-fail-with-body | no-fail-with-body-failure)
             [[ "$saw_no_fail_with_body" == true ]] || {
@@ -185,5 +221,53 @@ assert_file_equals \
 assert_file_excludes "$stderr_file" "$E2E_API_TOKEN"
 assert_file_excludes "$stderr_file" "$VERCEL_AUTOMATION_BYPASS_SECRET"
 assert_file_excludes "$stderr_file" "sensitive-request-payload"
+
+# A stalled request produces no response at all, so a plain GET is sent again.
+# See the merge-group Turbo run 33464652878, where four runner shards gave up
+# on GET /api/runs/:id/context after 30s while the API answered 200 in 30-44s.
+context_vercel_logs_search='https://vercel.com/vm0/vm0-api/logs?search=requestHost%3Apr-27981-api.vm6.ai+requestPath%3A%2Fapi%2Fruns%2Frun-1%2Fcontext'
+MOCK_CURL_EXPECTED_URL="https://pr-27981-api.vm6.ai/api/runs/run-1/context"
+MOCK_CURL_EXPECTED_VERCEL_URL="${context_vercel_logs_search}+status%3A000&timeline=past12Hours"
+MOCK_CURL_EXPECTED_VERCEL_WRITE_OUT='%{onerror}%{stderr}Vercel logs: https://vercel.com/vm0/vm0-api/logs?search=requestHost%%3Apr-27981-api.vm6.ai+requestPath%%3A%%2Fapi%%2Fruns%%2Frun-1%%2Fcontext+status%%3A%{http_code}&timeline=past12Hours\n'
+stall_stderr=$'curl: (28) Operation timed out after 30002 milliseconds with 0 bytes received\nVercel logs: '"$MOCK_CURL_EXPECTED_VERCEL_URL"$'\n'
+retry_stderr=$'runner_api_curl retrying https://pr-27981-api.vm6.ai/api/runs/run-1/context after no response (attempt 1 of 2)\n'
+
+MOCK_CURL_MODE=no-response-then-success
+reset_curl_attempts
+run_request "/api/runs/run-1/context"
+assert_status 0
+assert_curl_attempts 2
+assert_file_equals $'{"ok":true}\n' "$stdout_file"
+assert_file_equals "${stall_stderr}${retry_stderr}" "$stderr_file"
+
+MOCK_CURL_MODE=no-response
+reset_curl_attempts
+run_request "/api/runs/run-1/context"
+assert_status 28
+assert_curl_attempts 2
+assert_file_equals '' "$stdout_file"
+assert_file_equals \
+    "${stall_stderr}${retry_stderr}${stall_stderr}"$'runner_api_curl failed: url=https://pr-27981-api.vm6.ai/api/runs/run-1/context curl_status=28\nrunner_api_curl received no response from https://pr-27981-api.vm6.ai/api/runs/run-1/context within 30s across 2 attempt(s); the API may have answered after the client gave up: '"${context_vercel_logs_search}"$'&timeline=past12Hours\n' \
+    "$stderr_file"
+assert_file_excludes "$stderr_file" "$E2E_API_TOKEN"
+assert_file_excludes "$stderr_file" "$VERCEL_AUTOMATION_BYPASS_SECRET"
+
+# A write may already have been applied by the API, so it is never repeated.
+MOCK_CURL_EXPECTED_URL="https://pr-27981-api.vm6.ai/api/chat/events"
+MOCK_CURL_EXPECTED_VERCEL_URL='https://vercel.com/vm0/vm0-api/logs?search=requestHost%3Apr-27981-api.vm6.ai+requestPath%3A%2Fapi%2Fchat%2Fevents+status%3A000&timeline=past12Hours'
+MOCK_CURL_EXPECTED_VERCEL_WRITE_OUT='%{onerror}%{stderr}Vercel logs: https://vercel.com/vm0/vm0-api/logs?search=requestHost%%3Apr-27981-api.vm6.ai+requestPath%%3A%%2Fapi%%2Fchat%%2Fevents+status%%3A%{http_code}&timeline=past12Hours\n'
+
+MOCK_CURL_MODE=no-response
+reset_curl_attempts
+run_request "/api/chat/events" -X POST -d "$payload"
+assert_status 28
+assert_curl_attempts 1
+assert_file_excludes "$stderr_file" "sensitive-request-payload"
+
+MOCK_CURL_MODE=no-response
+reset_curl_attempts
+run_request "/api/chat/events" -d "$payload"
+assert_status 28
+assert_curl_attempts 1
 
 echo "runner_api_curl tests passed"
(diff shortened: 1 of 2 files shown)


e7h4n:
## Validation on this PR

CI run [33466571573](https://github.com/vm0-ai/vm0/actions/runs/33466571573) at `63f3e7a`: **96 checks pass, 4 skipping, 0 failures.**

All twelve `cli-e2e-03-runner` shards are green, including the four that failed in the triggering run [33464652878](https://github.com/vm0-ai/vm0/actions/runs/33464652878):

| Shard | Result |
| --- | --- |
| `run-t01-firewall-twilio.bats`, `run-t11-real-codex-billing.bats` | pass 1m0s |
| `run-t10-real-claude-smoke.bats` | pass 1m9s |
| `run-t23-custom-connector.bats`, `run-t24-built-in-provider-fallback.bats` | pass 1m16s |
| `run-t21-claude-runtime-regressions.bats`, `run-t22-connector-refresh.bats` | pass 1m5s |

Both product contracts under test still hold: the runner firewall resolves Twilio auth without exposing raw credentials, and the built-in pi loop returns a successful answer. Neither assertion was weakened.

`ci-gate-security` passes, so the extended `.github/scripts/tests/runner-api-curl-test.sh` ran in CI through `Run workflow script tests` in `security.yml`.

Preview browser validation is not applicable: the change is a Bash e2e harness helper with no app, API or UI surface.

e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 63f3e7a978c9c1777aca3a21892c107dc1af5633
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Makes the runner E2E helper `runner_api_curl` send a request again when curl exits `28` — the deployment produced no response at all — restricted to requests carrying neither a method override nor a body, and reports a zero-response stall separately from a served API error with a Vercel log search that omits the misleading `status:000` filter. The CI-executed `.github/scripts/tests/runner-api-curl-test.sh` gains four cases covering stall-then-success, stall on both attempts, and two write shapes that must not be repeated.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic: races, off-by-one, wrong conditionals | Pass | `runner-chat.bash:195-215` — the loop always `break`s when `attempt == max_attempts`, so `attempt` is exact when reused at line 228; the marker is read with `##*`/`%…*` (last occurrence), so a body containing the marker text still parses correctly. |
| `REVIEW.md` | Correctness: request retried only when it is safe to repeat | Pass | `_runner_api_request_is_repeatable` (`runner-chat.bash:140-153`) rejects `-X*`, `--request[=]`, `-d*`, `--data[-…][=]`, `-F*`, `--form`, `-T*`, `--upload-file`. Verified against every current `runner_api_curl` call site: the only `-X GET` in `e2e/` is a raw `curl` in `browser.bash:502`, so no real GET caller is misclassified. |
| `REVIEW.md` | Correctness: buffering must not change emitted bytes | Pass | The exit status is appended behind a marker specifically because `$( )` strips trailing newlines; the pre-existing `assert_file_equals $'{"ok":true}\n'` case still passes unchanged, proving byte fidelity. |
| `REVIEW.md` | Correctness: composition with the bounded polling helpers | Pass | `runner_wait_for_run` / `runner_wait_for_run_running` budget in wall-clock seconds. One stall then success now returns at ~30s instead of ~32s; two consecutive stalls exhaust a 60s budget either way (old: 0-30 stall, sleep 2, 32-62 stall; new: 0-60). No attempt is lost. |
| `REVIEW.md` | Security: no secrets or credentials exposed | Pass | New stderr lines emit only `diagnostic_url` (query string already stripped at line 69) and the Vercel search. `runner-api-curl-test.sh:93-94,106` assert the token, bypass secret, and request payload are absent from stderr. |
| `REVIEW.md` | Type safety / `any` casts | N/A | Bash only; no TypeScript in the diff. |
| `REVIEW.md` | Style: function length, duplicated logic, WHY-not-WHAT comments | Pass | `runner_api_curl` stays ~70 lines. All four added comments state why (why `28` and not `22`, why GET only, why buffer, why the marker, why the status filter hides the answer). The `+status%3A` literal now appears at two call sites instead of inside the prefix helper — a small, deliberate cost so the stall path can build a search without a status term. |
| `REVIEW.md` | Interface changes | Pass | `_runner_api_vercel_logs_url_prefix` is private and both callers are updated in the same diff; `runner_api_curl`'s signature, stdout, and exit status are unchanged. |
| `REVIEW.md` | Testing coverage for the commit type | Pass | `test:` commit; `feat:`/`fix:` test requirements do not apply. The diff still adds four cases to a suite CI actually runs (`security.yml` → `Run workflow script tests` → `for test_script in .github/scripts/tests/*.sh`). |
| `REVIEW.md` | Feature regression guard | Pass | No product feature changes. The existing four cases in `runner-api-curl-test.sh` (success, served `429`, `--no-fail-with-body` 409/500 teardown) are untouched and still assert the exact prior stdout and stderr, covering the paths most likely to regress from the refactor. |
| `docs/bad-smell.md` | §7 fail fast: no fallback/recovery logic that hides a problem | Pass | The rule targets production configuration fallbacks that hide misconfiguration. Here the fail-fast property that matters is strictly preserved: `--fail-with-body` turns every served HTTP status into curl `22`, which is never repeated and fails on the first attempt. Only a request that produced no response at all is repeated, and the stall is reported louder than before, not quieter. |
| `docs/bad-smell.md` | §2 no lint/type suppressions | Pass | No suppression comments added. `shellcheck -s bash -x` is clean on both files. |
| `docs/bad-smell.md` | §6 no hardcoded URLs or config | Pass | No new hardcoded endpoint. `https://vercel.com/vm0/vm0-api/logs` is pre-existing diagnostic output; the diff only moves the `+status%3A` fragment. Timeouts stay on `E2E_CURL_*` env overrides. |
| `docs/bad-smell.md` | §3 no unnecessary try/catch, no over-engineered error handling | Pass | No new error swallowing. `curl_status` is still returned verbatim to the caller. |
| `docs/fallback.md` | Does the diff introduce a fallback (a path for a state the system cannot produce)? | Pass | No. A 30-44s response from a live deployment is a state the system demonstrably produces — the PR body's request-log table shows the four stalled requests completing with `200`. This is not a `??`/`||` default on a contract-owned field, a retired-protocol reader, a cross-version compatibility branch, or feature-switch-gated behavior, so §9's mandatory `Fallbacks` section does not apply and none is expected in the summary. |
| `docs/fallback.md` | §1 no negative tests asserting removed behavior stays removed | Pass | The four added cases assert current behavior (retry, no retry, exact diagnostics). No tombstone tests. |
| `docs/testing.md` | Test behavior, not that a mock was called | Pass | `assert_curl_attempts` is not a "mock was called" assertion standing in for behavior — the number of requests issued *is* the behavior under test for "a write is never sent twice". Every case also asserts exit status, and the two primary cases assert exact stdout and byte-exact stderr. |
| `docs/testing.md` | No fake timers, no artificial delays in tests | Pass | The retry is immediate with no `sleep`; no timer manipulation is added anywhere in the diff. |
| `docs/testing.md` | Mock only external dependencies | Pass | Only `curl`, a third-party binary at the process boundary, is stubbed. `runner_api_curl` itself is exercised for real. |
| `docs/testing/anti-patterns.md` | Quick checklist: timer issues, artificial delays, testing mocks over behavior | Pass | None present; the checklist's mocking rows (`vi.mock`, `globalThis.services.db`, MSW, `vi.importActual`) have no Bash analogue in this diff. |
| `docs/testing/patterns.md` | Route/pure-function/MSW/filesystem/env-stub patterns and standard TS test structure | N/A | The diff adds no TypeScript test; the suite is a plain Bash script executed directly by `security.yml`, following that file's existing structure. |
| `docs/testing/cli-e2e-testing.md` | Runner E2E is CI-only; validate through the PR pipeline, not a local `./e2e/run.sh` | Pass | Validated on CI run 33466571573: all twelve `cli-e2e-03-runner` shards green, including the four that failed in run 33464652878. Recorded in the PR comment, not claimed from a local run. |
| `docs/testing/cli-e2e-testing.md` | Assert user-visible output; no test-only routes, tokens, or DB access | Pass | No new E2E case and no `/api/test/*`, fixture route, or DB access. The helper still calls the same public `/api/*` endpoints. |
| `docs/testing/cli-e2e-testing.md` | Keep shards self-contained; no cross-file ordering or shared mutable account state | Pass | The change is confined to per-request behavior inside one helper; no shared account preference is mutated and no ordering between BATS files is introduced. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

Non-blocking observations, recorded for the author rather than as findings:

- `-X GET` would be classified as non-repeatable and skip the retry. No current `runner_api_curl` caller uses that form, so the conservative direction is correct and adding a GET special case would be unused complexity today.
- The retry adds at most 30s to a GET that ultimately fails. With shards completing in ~1m20s against a 15-minute job limit there is ample headroom, and the retry only fires after a request has already burned the full 30s budget.

### Testing

- Coverage: Adequate. `.github/scripts/tests/runner-api-curl-test.sh` gains stall-then-success (asserts 2 attempts, exact body, exact stderr), stall-on-both-attempts (asserts exit 28, empty stdout, byte-exact stderr including the new no-status search link, and secret exclusion), `-X POST -d` not repeated, and `-d` without `-X` not repeated. The four pre-existing cases still pin the unchanged success, `429`, and teardown diagnostics. The end-to-end invariant is separately validated by the twelve green `cli-e2e-03-runner` shards on this PR.
- Conventions: Compliant. Only the external `curl` boundary is stubbed, assertions are on observable output rather than on mock bookkeeping, and no sleeps, fake timers, skipped tests, deleted tests, or weakened assertions are introduced. The attempt counter lives in a file because the stubbed `curl` runs inside a command substitution, which the added comment explains.

PR #30703, opened 2026-09-01
test(python): cover leading sigv4 dot segments

### LINKED ISSUE #30670: test(python): cover leading dot-segments in AWS SigV4 URI normalization
## Problem

`crates/runner/mitm-addon/src/aws_sigv4.py:539-555` implements non-S3 canonical URI dot-segment removal. The branch at lines 547-550 intentionally ignores `..` when no prior output segment exists, which is required for root and over-traversal paths such as `/../x`, `/../../x`, and `/a/../../x`. Existing signer coverage only exercises a `..` after a normal segment (`/a/../long/path%20name/`) in `crates/runner/mitm-addon/tests/test_firewall_aws_sigv4_auth.py:128-173`; no test executes the empty-output branch.

## Evidence

On the audited `origin/main` commit `0c17db0f5c278e71094171ee34c1bb295049070f`, direct `sign_request()` calls for all three paths complete successfully and use the canonical `/x` path for the signature while preserving the request URL. In a disposable worktree, changing only `if output: output.pop()` to an unconditional `output.pop()` makes each input raise `IndexError: pop from empty list`; all 290 AWS signer, firewall-auth, request-body, and framing tests still pass under that mutation. This confirms that the current branch is not regression-protected.

The signer/auth boundary promises typed, fail-closed signing errors. `crates/runner/mitm-addon/src/auth.py:1604-1621` catches only `AwsSigV4SigningError` to produce the controlled `502 aws_sigv4_auth_failed` response. A future regression in this branch would therefore escape the AWS-specific error path and can surface as a generic request-processing failure instead of the intended deterministic outcome.

## Impact

Requests that reach the signer with a leading or over-traversing dot segment would be vulnerable to an uncaught internal exception after credential resolution. The missing regression coverage also leaves the canonical URI contract for non-S3 services dependent on an untested empty-stack edge case.

## Acceptance criteria

- Add parameterized public `sign_request()` coverage for `/../x`, `/../../x`, and `/a/../../x` (or an equivalent representative matrix).
- Assert that signing completes without an uncaught exception and that the canonical signature corresponds to the normalized `/x` path.
- Add an auth-layer assertion that these paths either complete through the normal signing result or become a controlled local response, never leaking `IndexError` or another internal exception.
- Keep S3's separate raw-path canonicalization behavior covered and unchanged.
- Confirm that removing the empty-output guard makes the new focused tests fail.

This is a test-only improvement; no production signing behavior change is required for the current implementation.

seven332:
## Summary

- cover leading and over-traversing dot segments through the public AWS SigV4 signer boundary
- verify the same path matrix completes through the real firewall-auth flow without rewriting request URLs
- lock S3's intentionally distinct raw-path signing behavior

## Scope

This is a test-only change. It does not modify production signing or auth behavior.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (`7258 passed`)
- focused mutation check: replacing the guarded empty-stack pop with an unconditional pop made all six new non-S3 signer and auth cases fail with `IndexError`; restoring the guard made all seven focused cases pass
- pre-commit Ruff, file-size, and commitlint hooks

Closes #30670

### DIFF
--- a/crates/runner/mitm-addon/tests/test_aws_sigv4.py
+++ b/crates/runner/mitm-addon/tests/test_aws_sigv4.py
@@ -358,6 +358,55 @@ def test_sign_request_reuses_matching_inspection_without_changing_output() -> No
 
 
 @pytest.mark.parametrize(
+    "path",
+    [
+        pytest.param("/../x", id="leading-parent"),
+        pytest.param("/../../x", id="multiple-leading-parents"),
+        pytest.param("/a/../../x", id="over-traversal"),
+    ],
+)
+def test_header_auth_leading_dot_segments_use_normalized_canonical_path(path: str) -> None:
+    url = f"https://{STS_HOST}{path}"
+
+    signed_url, signed_headers = sign_request(
+        method="GET",
+        url=url,
+        headers=_header_auth_headers(),
+        body=None,
+        credentials=_credentials(),
+    )
+
+    authorization = {name.lower(): value for name, value in signed_headers}["authorization"]
+    assert signed_url == url
+    assert authorization == (
+        "AWS4-HMAC-SHA256 "
+        "Credential=AKIDEXAMPLE/20260101/us-east-1/sts/aws4_request, "
+        "SignedHeaders=host;x-amz-date, "
+        "Signature=df6768fccf76887a9d31a01e2d0d48ae57ecdce506e745b20b6ad1960411066b"
+    )
+
+
+def test_header_auth_s3_keeps_dot_segments_in_canonical_path() -> None:
+    authorizations: list[str] = []
+    for path in ("/../x", "/x"):
+        url = f"https://{_AWS_S3_EXAMPLE_HOST}{path}"
+        signed_url, signed_headers = sign_request(
+            method="GET",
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #30703

**Reviewed head:** `d7cc5b791b6a3941c90e480b50f3cd8382e62061`

### Summary

The test-only change protects non-S3 leading and over-traversal dot-segment signing through the public signer and real firewall-auth boundaries, while explicitly preserving S3 raw-path semantics.

### Findings

No findings. The signer test fixes canonical signature behavior to `/x`; the auth integration proves real credential/signature replacement, normal continuation, and URL preservation. The two local path matrices serve different contracts without introducing a shared test abstraction.

### Testing

- Focused 7-case suite passed.
- Removing the empty-output guard made all 6 new non-S3 cases fail with `IndexError`; restoring it made them pass.
- Full mitm-addon suite passed: `7258 passed`.
- GitHub CI is evaluated separately by the workflow and was not used as a substitute for local validation.

### Verdict

LGTM

PR #30712, opened 2026-09-01
test(api): synchronize cancellation chat callback delivery

e7h4n:
## Summary

- Drain the cancel route's tracked `waitUntil` work with `flushWaitUntilForTest()` before asserting the delivered chat callback, instead of polling for its effect.
- Replace the three `expect.poll(...)` blocks with direct `toContainEqual` / `find` assertions on the thread events.
- Preserve both halves of the product contract: the `run.cancelled` chat event with `runLifecycleEvent: "cancelled"` is still required for every cancellation, and `routeRequests` must still be `0` (no HTTP self-dispatch).

## Failure evidence

- Trigger: [Turbo run 33465886041](https://github.com/vm0-ai/vm0/actions/runs/33465886041), `push` on `main`, SHA `3ecf775e6c353b9aa8f8dcb0c35b0909a1520a26`, [test-api shard 7](https://github.com/vm0-ai/vm0/actions/runs/33465886041/job/99725987111)
- Failing test: `HOOK-01/RUN-03: terminal run callbacks dispatch on cancellation > delivers chat run callbacks through cancellation side effects without HTTP self-dispatch`
- Causal chain:

  ```
  AssertionError: expected false to be true // Object.is equality
   ❯ src/signals/routes/__tests__/run-lifecycle.bdd.test.ts:15833:8

  Caused by: Error: Matcher did not succeed in time.
   ❯ src/signals/routes/__tests__/run-lifecycle.bdd.test.ts:15822:5
  ```

  `expected false to be true` is the surface; `Matcher did not succeed in time` is the cause. The polling block never converged, so the following `expect(routeRequests).toBe(0)` was never reached.
- Shard summary: `Test Files 1 failed | 26 passed (27)`, `Tests 1 failed | 426 passed (427)`.

## Root cause

`runsCancelRoutes` returns its `200` as soon as `cancelRun$` commits and schedules the callback delivery in the background:

```ts
// turbo/apps/api/src/signals/routes/runs-cancel.ts
if (shouldDispatchCancelSideEffects(result)) {
  const backgroundSignal = new AbortController().signal;
  waitUntil(tapError(set(dispatchCancelSideEffects$, result, backgroundSignal), ...));
}
```

The `run.cancelled` chat event is written much later in that background chain, inside a transaction in `insertAssistantErrorEvent` (`internal-chat-run-callback.service.ts`), reached through `dispatchCancelSideEffects$ -> dispatchRunCallbacks$ -> dispatchSingleInternalCallback$`.

The test had no barrier for that work and instead polled for the event with `expect.poll`, which has no explicit options here. Vitest 4 defaults to `interval = 50`, `timeout = 1000`, and neither `apps/api/vitest.config.ts` nor `turbo/vitest.config.ts` configures `expect.poll`. So the assertion had a hard 1000 ms budget for a background Postgres write chain, while CI runs eight `test-api` shard files in parallel against one shared Postgres container. The job log shows `run-lifecycle.bdd.test.ts`, `chat-threads.bdd.test.ts` and `test-runtime-state.test.ts` interleaved in the same window.

Classification: missing test synchronization (first category in the flaky-test skill). This is not a product race: every observed interleaving is valid behaviour, and the delivery is already fully awaited inside the `waitUntil` promise, so `flushWaitUntilForTest()` is an exact completion boundary rather than a longer wait.

The `DrizzleQueryError` on `agent_run_queue` with `cause: Error: Cannot use a pool after calling end on the pool` at 03:25:04 belongs to `test-runtime-state.test.ts`, a different file with its own module registry, and is not causal. The `run-context ingest failed`, `sandbox telemetry ingest failed`, `enqueue telemetry failed`, `OpenRouter completion finished with unknown` and `google-drive token refresh failed: 503` lines are deliberate negative-path fixtures.

## Why this shape of fix

`flushWaitUntilForTest()` is the sanctioned barrier for API `waitUntil()` work and is already used 48 times in this same file, including by the sibling cancellation test at `run-lifecycle.bdd.test.ts:15121` which does exactly `requestCancelRun` then `flushWaitUntilForTest` then a hard assertion. No retries, sleeps, larger budgets, skips or weakened assertions are involved; the assertions are now unconditional, and a missing delivery fails immediately with the actual event list instead of `expected false to be true`.

## Validation

- Target test, 5 consecutive isolated runs: 5/5 passed (713-774 ms each).
- Full `run-lifecycle.bdd.test.ts`: 186/186 passed.
- Negative control: removing the first `flushWaitUntilForTest()` makes the test fail immediately with `expected [ ... ] to deep equally contain ObjectContaining{...}`, confirming the event really is produced by background work and that the new assertion still detects a missed delivery.
- Instrumented measurement on the pre-fix code: the first poll converged in 71 ms idle and 59 ms under two-worker local contention, against the 1000 ms budget. The local machine has 2 CPUs, so the CI shard's eight-way saturation could not be reproduced here; the fix removes the budget rather than widening it.
- `pnpm format`, `pnpm -F api run lint` (exit 0, only pre-existing warnings), `tsc -p tsconfig.gateways.json`, `tsconfig.core.json`, `tsconfig.bootstrap.json`, `tsc -p tsconfig.tests.json --noEmit` all pass. `git diff --check` clean.

## Limitations

The full repository Vitest suite was not run, per repository guidance; the PR pipeline covers it. No app, API, runner or dev server was started, and no secret manager was used.

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
@@ -15816,21 +15816,20 @@ describe("HOOK-01/RUN-03: terminal run callbacks dispatch on cancellation", () =
       prompt: "first cancellable chat run",
     });
     await api.requestCancelRun(actor, first.runId, [200]);
+    // Cancellation delivers its chat callback from the route's `waitUntil`
+    // work, so drain that work instead of polling for the appended event.
+    await flushWaitUntilForTest();
 
     const firstCancelled = await api.readRun(actor, first.runId);
     expect(firstCancelled.status).toBe("cancelled");
-    await expect
-      .poll(async () => {
-        const messages = await chat.listThreadEvents(actor, first.threadId);
-        return messages.events.some((message) => {
-          return (
-            message.eventType === "run.cancelled" &&
-            message.runId === first.runId &&
-            message.runLifecycleEvent === "cancelled"
-          );
-        });
-      })
-      .toBe(true);
+    const firstEvents = await chat.listThreadEvents(actor, first.threadId);
+    expect(firstEvents.events).toContainEqual(
+      expect.objectContaining({
+        eventType: "run.cancelled",
+        runId: first.runId,
+        runLifecycleEvent: "cancelled",
+      }),
+    );
     expect(routeRequests).toBe(0);
 
     const second = await sendChatRunMessage(actor, {
@@ -15839,21 +15838,18 @@ describe("HOOK-01/RUN-03: terminal run callbacks dispatch on cancellation", () =
       prompt: "second cancellable chat run",
     });
     await api.requestCancelRun(actor, second.runId, [200]);
+    await flushWaitUntilForTest();
 
     const secondCancelled = await api.readRun(actor, second.runId);
     expect(secondCancelled.status).toBe("cancelled");
-    await expect
-      .poll(async () => {
-        const messages = await chat.listThreadEvents(actor, first.threadId);
-        return messages.events.some((message) => {
-          return (
-            message.eventType === "run.cancelled" &&
-            message.runId === second.runId &&
-            message.runLifecycleEvent === "cancelled"
-          );
-        });
-      })
-      .toBe(true);
+    const secondEvents = await chat.listThreadEvents(actor, first.threadId);
+    expect(secondEvents.events).toContainEqual(
+      expect.objectContaining({
+        eventType: "run.cancelled",
+        runId: second.runId,
+        runLifecycleEvent: "cancelled",
+      }),
+    );
     expect(routeRequests).toBe(0);
 
     const third = await sendChatRunMessage(actor, {
@@ -15862,25 +15858,17 @@ describe("HOOK-01/RUN-03: terminal run callbacks dispatch on cancellation", () =
       prompt: "third cancellable chat run",
     });
     await api.requestCancelRun(actor, third.runId, [200]);
+    await flushWaitUntilForTest();
 
     const thirdCancelled = await api.readRun(actor, third.runId);
     expect(thirdCancelled.status).toBe("cancelled");
 
-    let cancelNote:
-      | Awaited<ReturnType<typeof chat.listThreadEvents>>["events"][number]
-      | undefined;
-    await expect
-      .poll(async () => {
-        const messages = await chat.listThreadEvents(actor, first.threadId);
-        cancelNote = messages.events.find((message) => {
-          return (
-            message.eventType === "run.cancelled" &&
-            message.runId === third.runId
-          );
-        });
-        return cancelNote?.eventType;
-      })
-      .toBe("run.cancelled");
+    const thirdEvents = await chat.listThreadEvents(actor, first.threadId);
+    const cancelNote = thirdEvents.events.find((message) => {
+      return (
+        message.eventType === "run.cancelled" && message.runId === third.runId
+      );
+    });
     if (!cancelNote || cancelNote.eventType !== "run.cancelled") {
       throw new Error(
         "Expected the delivered chat callback to append a cancellation event",


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: c2266a8bc7eabb22a123169ee89cc3837de05191
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Replaces three `expect.poll` waits in the cancellation chat-callback test with the repository's deterministic `flushWaitUntilForTest()` barrier plus direct assertions. The cancel route registers the whole callback delivery through `waitUntil(...)` before returning its 200, so the test was racing a background Postgres write chain against Vitest's default 1000 ms poll budget on an 8-way-parallel shared-Postgres shard. Single test file, +27/-39, no production code touched.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no races or wrong assumptions | Pass | `runs-cancel.ts:51-64` calls `waitUntil(...)` synchronously before the handler returns, so the work is registered in the tracker by the time `requestCancelRun` resolves. `flushWaitUntilForTest` loops until `tracker.pending` is empty, covering the nested `waitUntil` calls in `internal-chat-run-callback.service.ts`. The `run.cancelled` write in `insertAssistantErrorEvent` is fully awaited inside that promise, so the barrier is exact rather than approximate. |
| `REVIEW.md` | Type safety; no `any`, no overly broad types | Pass | No casts added. The removed `let cancelNote: Awaited<ReturnType<...>>["events"][number] \| undefined` is replaced by inference from `.find()`; the existing `if (!cancelNote \|\| cancelNote.eventType !== "run.cancelled")` guard still narrows before the field accesses. |
| `REVIEW.md` | No lint or type suppressions | Pass | Diff contains no `eslint-disable`, `oxlint-disable`, `@ts-ignore`, `@ts-expect-error`, or `prettier-ignore`. |
| `REVIEW.md` | Style and maintainability; comments explain WHY | Pass | Net -12 lines. The single added comment states why the barrier exists (delivery runs in the route's `waitUntil` work), not what the next line does. |
| `REVIEW.md` | Security; no auth, secret, or injection surface | N/A | Test-only change; no endpoint, credential, or query construction is touched. |
| `REVIEW.md` | Testing coverage for the commit type | Pass | `test:` commit modifying an existing test. No production behavior changes, so no new regression test is required. Both halves of the asserted contract are preserved and strengthened: the `run.cancelled` + `runLifecycleEvent: "cancelled"` event is now required unconditionally at all three cancellations, and `expect(routeRequests).toBe(0)` is unchanged. |
| `REVIEW.md` | Feature regression guard | N/A | No feature, UI, API, or user flow is changed, replaced, or rerouted. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No change under `packages/db/src/schema/**`, `packages/api-contracts/**`, or any `jsonb(...).$type<...>()`. |
| `REVIEW.md` / `docs/fallback.md` routing | Fallback discipline; declare introduced fallbacks | N/A | The diff introduces no fallback, compatibility branch, `??`/`\|\|` default, or feature-switch gate, and removes none. No `Fallbacks` section is required. |
| `docs/bad-smell.md` | No `any` types; no dynamic `import()`; no hardcoded URLs; no fail-soft fallback | Pass | None present in the 28 added lines. `CHAT_CALLBACK_URL` is the pre-existing shared constant, unchanged. |
| `docs/testing.md` | Integration tests only, exercised at the entry point | Pass | Remains an API route integration test driven through the Hono app; no internal function is tested directly. |
| `docs/testing.md` | Mock at the boundary only; no `vi.mock()` on `../` paths | Pass | No `vi.mock` added or changed anywhere in the diff. |
| `docs/testing.md` | Use MSW for HTTP, never direct `fetch` stubbing | Pass | The `server.use(http.post(CHAT_CALLBACK_URL, ...))` handler is untouched and remains the mechanism proving no HTTP self-dispatch. |
| `docs/testing.md` | Real database; never mock `globalThis.services.db` | Pass | Unchanged; the test still reaches real Postgres through the API. |
| `docs/testing.md` | No fake timers, sleeps, or artificial delays | Pass | Nothing added. `flushWaitUntilForTest()` awaits real promises; it is not a timer, sleep, retry, or enlarged budget. The diff removes waiting rather than adding it. |
| `docs/testing.md` | Test behavior, not mock call patterns | Pass | Assertions are on thread events read back through the API. `expect(routeRequests).toBe(0)` counts MSW-intercepted HTTP requests, which is the externally observable "no self-dispatch" contract, and is pre-existing and unchanged. |
| `docs/testing.md` | No manual detached cleanup inside a test body | Pass | This is the closest adjacent rule and the diff stays on the right side of it: it adds no `clearAllDetached()`. `flushWaitUntilForTest()` is the sanctioned explicit synchronization point for API `waitUntil()` work and is already used 48 times in this same file, including the sibling cancellation test at `run-lifecycle.bdd.test.ts:15121`. |
| `docs/testing.md` | Mock cleanup owned by global setup, not per test | Pass | No `vi.clearAllMocks`/`restoreAllMocks`/`unstubAllGlobals` or cleanup-only `finally` added; `src/__tests__/setup.ts` keeps ownership. |
| `docs/testing.md` | New user-facing features gated behind a `FeatureSwitchKey` | N/A | No new user-facing feature. |
| `docs/testing/api-testing.md` | Construct and verify state through API endpoints, not DB rows or services | Pass | Setup via the chat send endpoint (`sendChatRunMessage`) and cancel endpoint (`requestCancelRun`); verification via `readRun` and `chat.listThreadEvents`. No DB schema import, row write, row read, or service call is added. |
| `docs/testing/api-testing.md` | Shared persistent state must be uniquely owned, not isolated by ordering or tolerant assertions | Pass | The test keeps its `entitledRunActor()` identity and scopes every assertion to `first.threadId` and a specific `runId`. Replacing polling with a barrier narrows rather than widens the observation window; nothing depends on ordering against other files. |
| `docs/testing/testing-external-behavior.md` | Give context through the external interface and observe from the external surface | Pass | All context and all observation still go through API endpoints. `flushWaitUntilForTest()` is a test-lifecycle synchronization primitive, not a way to construct state or assert on internal state, so the external-behavior boundary is unchanged. |
| `docs/testing/anti-patterns.md` | AP-1 do not assert only that mocks were called | Pass | The primary assertion is the delivered event content; the request counter is a secondary contract assertion and is pre-existing. |
| `docs/testing/anti-patterns.md` | AP-4 do not mock internal code | Pass | No relative-path mocks anywhere in the diff. |
| `docs/testing/anti-patterns.md` | AP-5 no fake timers or artificial delays | Pass | None added; see above. |
| `docs/testing/anti-patterns.md` | AP-11 test through the route when one exists | Pass | Cancellation is still driven through `POST /api/runs/:id/cancel` rather than `cancelRun$` or `dispatchCancelSideEffects$`. |
| `docs/testing/patterns.md` | Pattern 3 MSW handler usage; no `server.listen`/`close` in test files | Pass | Only `server.use(...)` is present, and it is unchanged. |
| `.claude/skills/ccstate/SKILL.md` | Async work must be owned, not silenced with `void` or bare `detach` | Pass | The diff adds no `void`, `detach()`, or `AbortController`. It consumes the production `waitUntil` wrapper's existing tracking, which is the intended ownership path for `apps/api` background work. |

### Findings

#### Critical (P0)
- None.

#### High Priority (P1)
- None.

Two smaller points checked and cleared rather than filed:

- `flushWaitUntilForTest()` rethrows the first non-abort rejection from drained work. Combined with the `tapError` wrapper in `runs-cancel.ts`, a genuine dispatch failure now surfaces as that error instead of an opaque timeout. That improves diagnosis rather than adding brittleness, and 48 existing call sites in this file already accept the same semantics.
- `toContainEqual(expect.objectContaining({ eventType, runId, runLifecycleEvent }))` requires a single element matching all three fields, which is exactly what the removed `.some()` predicate required. No assertion is weakened, and failures now print the actual event list instead of `expected false to be true`.

### Testing

- Coverage: Adequate. `test:` commit with no production change; the modified test keeps both halves of its original contract and asserts them unconditionally.
- Conventions: Compliant with `docs/testing.md`, `docs/testing/api-testing.md`, `docs/testing/patterns.md`, `docs/testing/anti-patterns.md`, and `docs/testing/testing-external-behavior.md`.
- Author-reported validation, consistent with the diff: target test 5/5 green (713-774 ms), full `run-lifecycle.bdd.test.ts` 186/186 green, `pnpm -F api run lint` exit 0 with only pre-existing warnings, and all four `apps/api` tsc projects (`gateways`, `core`, `bootstrap`, `tests`) passing.
- Negative control recorded in the PR body: removing the first `flushWaitUntilForTest()` makes the test fail immediately with `to deep equally contain ObjectContaining{...}`, which confirms the event really is produced by background work and that the new assertion still detects a missed delivery.
- Noted limitation, disclosed in the PR body: the CI eight-way shard saturation was not reproducible on the author's 2-CPU machine, so the original flake was not reproduced locally. The fix removes the timing budget instead of widening it, so reproduction is not required to establish correctness.

PR #30715, opened 2026-09-01
fix(desktop): treat a user-ended screen share as a finished recording

bingjiez666:
Found by dogfooding on real hardware: stopping a recording from the macOS
screen-sharing indicator showed `Screen Recording: Failed` — "用户已停止流播放" —
and no `.clicks.json` was ever written.

## Two defects, one path

Both live on the path where the stream ends without `stop` being called.

**1. A deliberate finish was reported as a fault.** macOS delivers the user
ending the share as an `SCStreamError` through `didStopWithError`, the same way
it delivers a real breakage. The old code mapped every such error to
`source_lost`. Ending the share is how you stop a recording, so this was the
normal path being labelled a failure.

**2. Nothing was finalized.** The click track is written by `stop`, so an
externally ended capture lost it every time. The video only survived at all
because the writer runs with `movieFragmentInterval` — the fragments on disk
were the whole reason two playable `.mp4` files existed with no `.clicks.json`
beside them.

## The fix

`StreamStopClassifier` (pure, in `ScreenRecorderCore`) separates
`SCStreamErrorUserStopped` from a genuine fault. It checks the **error domain
before believing the code**, because -3817 means something unrelated outside
ScreenCaptureKit.

An externally ended stream no longer moves the session to a terminal state. The
writer still holds frames the user captured, and throwing those away would
discard the recording they just made — so the session stays finalizable, and
only the *reported* status changes. `stop` then finalizes exactly as it always
did, which is also why the click track now gets written on this path.

The controller reacts to the two reported statuses differently:

- **stopped** — finalize, then deliver, identical to an explicit stop.
- **failed** — finalize so the partial video and click track survive, keep the
  reason, and do **not** auto-deliver. Shipping a recording that broke partway
  without being asked is the wrong default; the tray offers `Retry Delivery`
  instead.

The tray's retry item was previously limited to `delivery_failed`; a salvaged
capture is equally worth delivering, so it now appears whenever files exist.

## Verification

`check-types`, `lint`, `knip`, desktop vitest (444 passed) locally. Four new
controller tests (clean finish delivers; broken capture is kept but not shipped;
a salvaged recording can be delivered on request; a salvage that itself fails
reports the capture reason), one new tray test, and three new Swift tests on the
classifier including the wrong-domain case.

**Unverified on hardware**: that -3817 in that domain is what macOS actually
sends for this. The pure classifier is tested, but the value it is fed comes
from a real `SCStreamError` I cannot produce in CI. Worth confirming on the next
dogfood run that the same action now reads as a normal finish and leaves a
`.clicks.json` behind.

sandbox and pass in CI.

### DIFF
--- a/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderCore/StreamStopReason.swift
+++ b/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderCore/StreamStopReason.swift
@@ -1,4 +1,5 @@
 import Foundation
+import ScreenCaptureKit
 
 /// Why a capture stream ended without the app asking it to.
 public enum StreamStopReason: Equatable, Sendable {
@@ -11,14 +12,16 @@ public enum StreamStopReason: Equatable, Sendable {
 }
 
 public enum StreamStopClassifier {
-    /// `SCStreamErrorUserStopped`. macOS reports the user ending the share
-    /// through its own indicator as an `SCStreamError`, so the only thing
-    /// separating "finished" from "broken" is this code.
-    public static let userStoppedErrorCode = -3817
+    /// macOS reports the user ending the share through its own indicator as an
+    /// `SCStreamError`, so the only thing separating "finished" from "broken"
+    /// is this code. Taken from the SDK rather than written out, because a
+    /// mistyped literal would classify every user stop as a fault and silently
+    /// restore the bug this classifier exists to fix.
+    public static let userStoppedErrorCode = SCStreamError.Code.userStopped.rawValue
 
     /// `SCStreamError` values live in this domain; codes from anywhere else
     /// carry no such guarantee and are treated as faults.
-    public static let streamErrorDomain = "com.apple.ScreenCaptureKit.SCStreamErrorDomain"
+    public static let streamErrorDomain = SCStreamError.errorDomain
 
     public static func classify(domain: String, code: Int) -> StreamStopReason {
         guard domain == streamErrorDomain, code == userStoppedErrorCode else {
--- a/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderCore/StreamStopReason.swift
+++ b/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderCore/StreamStopReason.swift
@@ -1,0 +1,29 @@
+import Foundation
+
+/// Why a capture stream ended without the app asking it to.
+public enum StreamStopReason: Equatable, Sendable {
+    /// The person recording ended the share themselves, from the system's own
+    /// screen-sharing indicator. This is an ordinary finish, not a fault.
+    case userStopped
+    /// The capture broke: the window closed, the display was unplugged, or the
+    /// stream was revoked.
+    case failed
+}
+
+public enum StreamStopClassifier {
+    /// `SCStreamErrorUserStopped`. macOS reports the user ending the share
+    /// through its own indicator as an `SCStreamError`, so the only thing
+    /// separating "finished" from "broken" is this code.
+    public static let userStoppedErrorCode = -3817
+
+    /// `SCStreamError` values live in this domain; codes from anywhere else
+    /// carry no such guarantee and are treated as faults.
+    public static let streamErrorDomain = "com.apple.ScreenCaptureKit.SCStreamErrorDomain"
+
+    public static func classify(domain: String, code: Int) -> StreamStopReason {
+        guard domain == streamErrorDomain, code == userStoppedErrorCode else {
+            return .failed
+        }
+        return .userStopped
+    }
+}
--- a/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderHelper/main.swift
+++ b/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderHelper/main.swift
@@ -168,6 +168,10 @@ private final class RecorderSession: NSObject, SCStreamDelegate, SCStreamOutput,
 
     private var state: RecorderSessionState = .ready
     private var failure: HelperFailure?
+    /// Set when the stream ended without `stop` being called. The session stays
+    /// finalizable so the partial recording and its click track are still
+    /// written; only the reported status changes.
+    private var externalStop: (reason: StreamStopReason, message: String)?
     private var stream: SCStream?
     private var writer: AVAssetWriter?
     private var videoInput: AVAssetWriterInput?
@@ -449,7 +453,7 @@ private final class RecorderSession: NSObject, SCStreamDelegate, SCStreamOutput,
         lock.lock()
         defer { lock.unlock() }
         var described: [String: Any] = [
-            "status": state.rawValue,
+            "status": reportedStatusLocked(),
             "elapsedMs": Int(elapsedSecondsLocked() * 1000),
         ]
         if let failure {
@@ -457,8 +461,23 @@ private final class RecorderSession: NSObject, SCStreamDelegate, SCStreamOutput,
                 "code": failure.code,
                 "message": failure.message,
             ]
+        } else if let externalStop, externalStop.reason == .failed {
+            described["error"] = [
+                "code": "source_lost",
+                "message": externalStop.message,
+            ]
         }
         return described
+    }
+
+    /// The status the caller sees. An externally ended stream is reported as
+    /// finished or failed straight away, while `state` stays `.recording` so
+    /// the caller can still run `stop` to finalize the file it already has.
+    private func reportedStatusLocked() -> String {
+        guard let externalStop, state == .recording else {
+            return state.rawValue
+        }
+        return externalStop.reason == .userStopped ? "stopped" : "failed"
     }
 
     var isTerminal: Bool {
@@ -485,23 +504,39 @@ private final class RecorderSession: NSObject, SCStreamDelegate, SCStreamOutput,
         return max(0, CMTimeGetSeconds(CMTimeSubtract(latest, start)))
     }
 
-    private func markFailed(code: String, message: String) {
+    /// Records that the stream ended on its own.
+    ///
+    /// Deliberately does not move to a terminal state: the writer still holds
+    /// frames that were captured before the stream ended, and discarding them
+    /// would throw away the recording the user just made. `stop` finalizes as
+    /// usual, and the reported status tells the caller to run it.
+    private func noteExternalStop(reason: StreamStopReason, message: String) {
         lock.lock()
-        state = .failed
-        failure = HelperFailure(code: code, message: message)
+        if externalStop == nil {
+            externalStop = (reason, message)
+        }
         let captureStream = stream
         stream = nil
         lock.unlock()
         captureStream?.stopCapture { _ in }
-        // The tap outlives the stream unless it is torn down here: a failed
-        // session is never stopped by the caller.
+        // The tap outlives the stream unless it is torn down here.
         clickTracker.stop()
     }
 
     // MARK: SCStreamDelegate
 
     func stream(_ stream: SCStream, didStopWithError error: Error) {
-        markFailed(code: "source_lost", message: error.localizedDescription)
+        // Ending the share from the system indicator arrives here as an error
+        // like any other, so the code decides whether this is a finish or a
+        // fault. Reporting a deliberate stop as a failure was wrong.
+        let nsError = error as NSError
+        noteExternalStop(
+            reason: StreamStopClassifier.classify(
+                domain: nsError.domain,
+                code: nsError.code
+            ),
+            message: error.localizedDescription
+        )
     }
 
     // MARK: SCStreamOutput
--- a/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderHelper/main.swift
+++ b/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderHelper/main.swift
@@ -167,7 +167,6 @@ private final class RecorderSession: NSObject, SCStreamDelegate, SCStreamOutput,
     private var timebase = mach_timebase_info_data_t()
 
     private var state: RecorderSessionState = .ready
-    private var failure: HelperFailure?
     /// Set when the stream ended without `stop` being called. The session stays
     /// finalizable so the partial recording and its click track are still
     /// written; only the reported status changes.
@@ -456,12 +455,7 @@ private final class RecorderSession: NSObject, SCStreamDelegate, SCStreamOutput,
             "status": reportedStatusLocked(),
             "elapsedMs": Int(elapsedSecondsLocked() * 1000),
         ]
-        if let failure {
-            described["error"] = [
-                "code": failure.code,
-                "message": failure.message,
-            ]
-        } else if let externalStop, externalStop.reason == .failed {
+        if let externalStop, externalStop.reason == .failed {
             described["error"] = [
                 "code": "source_lost",
                 "message": externalStop.message,
@@ -722,21 +716,25 @@ private func handleStart(_ request: [String: Any]) throws -> [String: Any] {
 private func handleStop(_ request: [String: Any]) throws -> [String: Any] {
     let sessionId = try requiredString(request, "sessionId")
     let session = try sessionStore.session(sessionId)
-    let result = try session.stop()
-    sessionStore.remove(sessionId)
-    return result
+    // `stop` reaches its terminal state before any of the work that can throw,
+    // so a finalize that fails partway leaves a session that can never be
+    // stopped again. Releasing on the way out either way is what keeps the
+    // store bounded; the caller gives up on a rejected stop and never retries.
+    defer {
+        if session.isTerminal {
+            sessionStore.remove(sessionId)
+        }
+    }
+    return try session.stop()
 }
 
 private func handleState(_ request: [String: Any]) throws -> [String: Any] {
     let sessionId = try requiredString(request, "sessionId")
     let session = try sessionStore.session(sessionId)
-    let described = session.describedState()
-    if session.isTerminal {
-        // A session that failed on the delegate queue is never stopped by the
-        // caller, so releasing it here is what keeps the store bounded.
-        sessionStore.remove(sessionId)
-    }
-    return described
+    // A stream that ended on its own is reported here but deliberately left in
+    // the store: it still holds the frames the user captured, and only `stop`
+    // can finalize them. `handleStop` is what releases it.
+    return session.describedState()
 }
 
 private func handle(_ request: [String: Any]) throws -> [String: Any] {
--- a/turbo/apps/desktop/native/computer-use-helper/Tests/ScreenRecorderCoreTests/StreamStopReasonTests.swift
+++ b/turbo/apps/desktop/native/computer-use-helper/Tests/ScreenRecorderCoreTests/StreamStopReasonTests.swift
@@ -17,7 +17,7 @@ struct StreamStopReasonTests {
 
     @Test
     func treatsEveryOtherStreamErrorAsAFault() {
-        // A closed window or an unplugged display arrives in the same domain
+        // A revoked, declined, or broken capture arrives in the same domain
         // with a different code, and must not be reported as a clean finish.
         #expect(StreamStopClassifier.classify(domain: domain, code: -3801) == .failed)
         #expect(StreamStopClassifier.classify(domain: domain, code: 0) == .failed)
--- a/turbo/apps/desktop/native/computer-use-helper/Tests/ScreenRecorderCoreTests/StreamStopReasonTests.swift
+++ b/turbo/apps/desktop/native/computer-use-helper/Tests/ScreenRecorderCoreTests/StreamStopReasonTests.swift
@@ -1,0 +1,38 @@
+import Testing
+
+@testable import ScreenRecorderCore
+
+struct StreamStopReasonTests {
+    private let domain = StreamStopClassifier.streamErrorDomain
+
+    @Test
+    func treatsTheUserEndingTheShareAsAnOrdinaryFinish() {
+        #expect(
+            StreamStopClassifier.classify(
+                domain: domain,
+                code: StreamStopClassifier.userStoppedErrorCode
+            ) == .userStopped
+        )
+    }
+
+    @Test
+    func treatsEveryOtherStreamErrorAsAFault() {
+        // A closed window or an unplugged display arrives in the same domain
+        // with a different code, and must not be reported as a clean finish.
+        #expect(StreamStopClassifier.classify(domain: domain, code: -3801) == .failed)
+        #expect(StreamStopClassifier.classify(domain: domain, code: 0) == .failed)
+    }
+
+    @Test
+    func doesNotTrustTheCodeFromAnotherErrorDomain() {
+        // -3817 means something else entirely outside ScreenCaptureKit, so the
+        // domain has to match before the code is believed.
+        #expect(
+            StreamStopClassifier.classify(
+                domain: "NSOSStatusErrorDomain",
+                code: StreamStopClassifier.userStoppedErrorCode
+            ) == .failed
+        )
+        #expect(StreamStopClassifier.classify(domain: "", code: -3817) == .failed)
+    }
+}
--- a/turbo/apps/desktop/src/desktop-recorder-controller.test.ts
+++ b/turbo/apps/desktop/src/desktop-recorder-controller.test.ts
@@ -441,29 +441,6 @@ describe("DesktopRecorderController", () => {
     });
   });
 
-  it("ends the session when the capture source disappears", async () => {
-    const backend = createBackendFake({
-      getStatus: vi.fn(
-        async (): Promise<DesktopRecorderNativeStatus> => ({
-          status: "failed",
-          elapsedMs: 2000,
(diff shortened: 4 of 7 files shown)


bingjiez666:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 70e1f79244c147f163d883b62c5d893f4cd70785

### Summary

Splits `SCStreamErrorUserStopped` from a genuine capture fault so ending the share from the macOS indicator is reported as a finished recording, and keeps the session finalizable on that path so `stop` still writes the `.clicks.json` click track. The diagnosis is right and the shape of the fix — a pure classifier in `ScreenRecorderCore`, a non-terminal `externalStop`, and a controller that finalizes both ways but only auto-delivers a clean finish — is the correct one.

Blocking on four items: the two constants the whole fix hinges on are hand-copied literals that the tests cannot validate, and retiring `markFailed` left three readers behind, one of which silently dropped the helper's session-reaping invariant.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic: no unhandled edge cases or wrong assumptions | Finding (P1) | `main.swift:734` — `handleState` reaps a session only when `isTerminal`, which is now unreachable; an externally stopped session the caller never finalizes is orphaned in `sessionStore`. |
| `REVIEW.md` | API misuse / wrong assumptions about a framework | Finding (P1) | `StreamStopReason.swift:16-22` — `-3817` and the `SCStream` error-domain string are hand-copied instead of taken from ScreenCaptureKit. |
| `REVIEW.md` | Security: no injection, hardcoded secrets, or missing auth on new surfaces | Pass | No new endpoint, credential, or IPC channel; the helper line protocol and its validation in `desktop-recorder-native.ts` are untouched. |
| `REVIEW.md` | Type safety: no `any`, no overly broad types | Pass | `collectAfterExternalStop` takes `DesktopRecorderError \| null`; the Swift tuple is concretely typed; `toNativeStatus` already narrows `"stopped"`. |
| `REVIEW.md` | Style & maintainability: no duplicated logic, comments explain WHY | Finding (P1) | `main.swift:735-737` comment now states a false invariant; `desktop-recorder-controller.test.ts:444` is a strict subset of the new test at line 371. |
| `REVIEW.md` | Feature regression guard: changed existing behavior needs coverage of the path most likely to regress | Finding (P1) | `desktop-tray-menu.ts:434` — the `Retry Delivery` gate widened from `delivery_failed` to any error; the newly reachable `signed_out` case has no test. |
| `REVIEW.md` | Testing coverage: a `fix:` commit carries a regression test | Pass | Three Swift classifier tests, four controller tests, one tray test; all assert live behavior. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No `packages/db/src/schema/**`, `api-contracts`, Zod response schema, or `jsonb().$type<>()` change. |
| `REVIEW.md` | Event sourcing / optimistic events | N/A | No persistent or optimistic event, event ID, or projection is touched. |
| `docs/bad-smell.md` | §1 zero `any`; §2 zero lint/type suppressions | Pass | No `any` and no `eslint-disable`/`@ts-expect-error`/`prettier-ignore` anywhere in the diff. |
| `docs/bad-smell.md` | §3 no unnecessary try/catch; fail fast | Pass | The one new `try/catch` wraps `backend.stop` on a genuinely recoverable path and records the capture reason rather than swallowing it. |
| `docs/bad-smell.md` | §5 no dynamic `import()` | Pass | All imports in the changed TypeScript are static. |
| `docs/bad-smell.md` | §6 no hardcoded environment-specific values | Finding (P1) | `StreamStopReason.swift:22` hardcodes `"com.apple.ScreenCaptureKit.SCStreamErrorDomain"` where the SDK owns the value. |
| `docs/bad-smell.md` | §7 no fallback/recovery logic that hides a problem | Finding (P1) | `main.swift:459-463` — `if let failure` now guards a value nothing writes. |
| `docs/fallback.md` | §1 no negative tests asserting removed behavior stays removed | Pass | `doesNotTrustTheCodeFromAnotherErrorDomain` pins current classifier behavior for live inputs, not the absence of deleted code. |
| `docs/fallback.md` | §2 a non-GA feature switch needs no compatibility code | Pass | `DesktopScreenRecording` is `enabled: false`, scoped to `9fd4ee92`; the diff correctly adds no dual-read or migration. |
| `docs/fallback.md` | §3 no fallback on a state the owning contract cannot produce | Finding (P1) | `main.swift:170,459` — `failure` has no writer after `markFailed` was deleted, so the branch is permanently dead. |
| `docs/fallback.md` | §5 delete the reader in the same PR that retires the producer | Finding (P1) | `markFailed` was the sole producer of `failure` and of `state = .failed`; three readers survive it (`describedState`, `isTerminal`, `handleState`). |
| `docs/fallback.md` | §9 every fallback introduced by the diff is declared in the PR summary | Pass | The diff introduces no cross-version or compatibility fallback, so no `Fallbacks` section is required. |
| `docs/deployment-compatibility.md` | Cross-version compatibility for independently deployed surfaces | N/A | The doc scopes itself to frontend, backend, and runner and states the desktop app is outside the model; the JS side and the Swift helper ship in one signed bundle via `resolveNativeHelperPath`, so no skew exists between them. |
| `docs/testing.md` | Integration tests at the entry point; unit tests only for the narrow exceptions | Pass | Controller tests drive the public `DesktopRecorderController` API; the Swift classifier is the allowed pure/algorithmic exception. |
| `docs/testing.md` | Mock at the boundary only — no relative `vi.mock()` paths | Pass | No `vi.mock` at all; the native helper is injected as a fake backend. |
| `docs/testing.md` | No fake timers | Pass | No `vi.useFakeTimers()` / `advanceTimersByTime`. |
| `docs/testing.md` | Test behavior, not mocks | Pass | Each new controller test asserts observable state and `openReview`/`deliver` outcomes alongside the `stop` call. |
| `docs/testing.md` | New user-facing behavior gated behind a `FeatureSwitchKey` | Pass | Reached only through `setFeatureEnabled` / `FeatureSwitchKey.DesktopScreenRecording`. |
| `docs/testing/desktop-testing.md` | Choose the smallest matching entry point; Swift policy/algorithm tests stay in SwiftPM | Pass | `StreamStopReasonTests` sits in `Tests/ScreenRecorderCoreTests`; controller and tray tests stay on their existing boundaries. |
| `docs/testing/desktop-testing.md` | Keep Desktop source modules real; mock only external system boundaries | Pass | Real controller, real tray builder, real types; only the helper process is faked. |
| `docs/testing/desktop-testing.md` | Tray menu state matrices are an accepted narrow exception | Pass | The new tray case extends the existing `desktop-tray-menu.test.ts` matrix. |
| `docs/testing/patterns.md` | Pattern 2: pure-function tests only with a complexity/security justification | Finding (P1) | The classifier qualifies, but the test seeds its inputs from the same constants it verifies, so it cannot fail if those constants are wrong. |
| `docs/testing/anti-patterns.md` | AP-1: do not assert mock calls instead of behavior | Pass | `expect(backend.stop).toHaveBeenCalledWith("session-1")` always accompanies a state or delivery assertion. |
| `docs/testing/anti-patterns.md` | AP-4: do not mock internal code | Pass | No internal module is mocked. |
| `docs/testing/anti-patterns.md` | AP-8: over-testing | Finding (P1) | `desktop-recorder-controller.test.ts:444` is now fully subsumed by the new test at line 371. |

### Findings

**1. `StreamStopReason.swift:16-22` — the two values the fix depends on are hand-copied, and the tests cannot catch it if either is wrong.**

`userStoppedErrorCode = -3817` and `streamErrorDomain = "com.apple.ScreenCaptureKit.SCStreamErrorDomain"` are both literals. The PR summary flags the code as *"Unverified on hardware"*, but the domain string carries exactly the same risk and is not called out. If either is wrong, `classify` returns `.failed` for every user stop and the entire PR is a silent no-op — the user still sees `Screen Recording: Failed`, which is the bug being fixed.

The Swift tests cannot detect this, because they feed the constants back into the function that reads them:

```swift
private let domain = StreamStopClassifier.streamErrorDomain
...
StreamStopClassifier.classify(domain: domain, code: StreamStopClassifier.userStoppedErrorCode)
```

That passes identically if both values are fiction. `docs/testing/patterns.md` Pattern 2 justifies a pure test for genuinely complex logic; here the logic is a two-term comparison and all the risk lives in the operands.

The package targets `.macOS(.v14)`, so `ScreenRecorderCore` can import ScreenCaptureKit and source both from the SDK while `classify(domain:code:)` stays pure and testable:

```swift
import ScreenCaptureKit

public static let userStoppedErrorCode = SCStreamError.Code.userStopped.rawValue
public static let streamErrorDomain = SCStreamError.errorDomain
```

Please confirm the exact symbol spellings against the SDK (the domain may be the global `SCStreamErrorDomain`). That turns the caveat into a compile-time guarantee and removes the need for a hardware check on this specific point.

**2. `main.swift:170`, `main.swift:459-463` — `failure` no longer has a writer; the branch that reads it is dead.**

Deleting `markFailed` removed the only assignment to `private var failure: HelperFailure?`. It is now permanently `nil`, so the `if let failure` branch in `describedState` can never run and the `else if let externalStop` branch is the only live one. `docs/fallback.md` §3 calls a branch guarding a state no writer can produce exactly this, and §5 asks for the reader to be deleted in the same PR that retires the producer. Please drop the field and collapse the branch.

**3. `main.swift:734-737` — the session reaper is now unreachable, and its comment states an invariant the diff just broke.**

`handleState` releases a session only `if session.isTerminal`, and `isTerminal` is `state == .stopped || state == .failed`. After this change nothing sets `.failed` (`markFailed` was the only writer, and `RecorderTransitionPolicy` never returns it), and `.stopped` is only reached inside `stop()`, which `handleStop` immediately follows with `sessionStore.remove`. So the reaper can no longer fire, and its comment — *"A session that failed on the delegate queue is never stopped by the caller, so releasing it here is what keeps the store bounded"* — is now false.

The path this diff itself makes reachable is the leak: when the stream ends externally and `collectAfterExternalStop` sees `backend.stop` reject (the helper's 10s `stopCapture` / 30s `finishWriting` timeouts, or a `capture_failed`), the controller calls `failSession`, clears `sessionId`, and never retries. The old code had already reaped that session; now it stays in `sessions` holding its `AVAssetWriter` and inputs for the life of the helper process. This is exactly the case your own fourth test — *"reports the capture reason when the salvage itself fails"* — exercises on the JS side, so the helper-side counterpart is worth closing: either have `handleState` also release a session once `externalStop` is set and the caller has been told, or keep an explicit release command. Please also correct the comment.

**4. `desktop-tray-menu.ts:434` — widening the `Retry Delivery` gate to any error makes it appear for `signed_out` with a stale, already-delivered recording.**

Offering retry after a salvaged capture is right. But dropping `recorder.error.code === "delivery_failed"` also catches the other error the controller sets: `prepare()` writes `{ code: "signed_out" }` when `canDeliver()` is false, and `lastRecording` is never cleared after a *successful* delivery. So: record → deliver → sign out → try to record again yields `Screen Recording: Failed`, the "Sign in to Okou…" line, and a `Retry Delivery` item pointing at a recording that was already delivered. Clicking it re-uploads it and opens a second review URL for the same capture.

`docs/fallback.md` §2 keeps this off the P0 list — `DesktopScreenRecording` is `enabled: false` and scoped to one staff hash, so there is no external blast radius. Still worth gating on the errors that actually describe an undelivered capture (`delivery_failed`, `source_lost`, `capture_failed`) rather than on any error at all, plus a tray case pinning that `signed_out` does not offer it.

**5. `desktop-recorder-controller.test.ts:444` — `"ends the session when the capture source disappears"` is now redundant.**

It uses the same `status: "failed"` / `source_lost` fake as the new `"keeps the partial recording when the capture breaks"` at line 371, and asserts a strict subset of it (`status`, `sessionId`, `error`). `docs/testing/anti-patterns.md` AP-8 — please fold the two together.

### Testing

- **Coverage**: Adequate for a `fix:` commit in shape — a Swift classifier suite, four controller tests covering clean finish, broken-but-kept, salvage-on-request, and salvage-failure, and one tray case. Two gaps: the classifier tests are tautological with respect to the constants that carry all the real risk (finding 1), and the newly reachable `signed_out` + stale `lastRecording` tray path is uncovered (finding 4).
- **Conventions**: Compliant. No `vi.mock` of internal modules, no fake timers, no direct `fetch` stubbing; the helper is injected as a fake backend at the process boundary, matching `docs/testing/desktop-testing.md`. Assertions are behavioral rather than mock-call-only. The Swift tests sit in the SwiftPM target the same doc designates for native policy and algorithm behavior.
- **CI**: green on `70e1f79`, including `build-macos-arm64`, `lint-types`, `lint-knip`, `lint-eslint`, and `test-other` — which confirms the dead `failure` field does not break the Swift build, only that it is dead.

bingjiez666:
Addressed all five review findings in `40f99fb`.

**1. SDK constants instead of literals.** `StreamStopClassifier` now imports ScreenCaptureKit and takes both values from the SDK:

```swift
public static let userStoppedErrorCode = SCStreamError.Code.userStopped.rawValue
public static let streamErrorDomain = SCStreamError.errorDomain
```

`classify(domain:code:)` stays pure and the tests are unchanged, but the operands are now compiler-guaranteed rather than transcribed. This resolves the `-3817` half of the summary's "Unverified on hardware" caveat, and the domain string it did not mention. What remains genuinely unverified is only whether macOS delivers `userStopped` for this specific action — no literal can be wrong about it any more.

**2. Removed the dead `failure` field.** `markFailed` was its only writer, so after this PR it was permanently `nil` and the `if let failure` branch in `describedState` could never run. Field and branch deleted; the `externalStop` branch is now the only one.

**3. Closed the session leak and corrected the comment.** `handleStop` released the session only on success, but `session.stop()` reaches `.stopped` via `transition(.stop)` before any of the work that can throw — so a failed finalize left a session that could never be stopped again, sitting in the store holding its `AVAssetWriter`. It now releases on the way out either way:

```swift
defer {
    if session.isTerminal {
        sessionStore.remove(sessionId)
    }
}
return try session.stop()
```

That is the path the controller reaches when `backend.stop` rejects and it gives up. The `isTerminal` reap in `handleState` was unreachable after this PR and its comment was false, so it is gone; `handleState` now explains why an externally stopped session is deliberately kept.

**4. Narrowed the `Retry Delivery` gate.** Replaced the unconditional `recorder.lastRecording` check with an explicit set of codes that mean a capture is on disk and undelivered — `capture_failed`, `delivery_failed`, `source_lost`. `signed_out` is raised by `prepare` before anything is captured, and `lastRecording` outlives a successful delivery, so the widened gate would have offered to re-upload an already-delivered recording. Added a tray case pinning that `signed_out` does not offer it.

**5. Dropped the redundant controller test.** `"ends the session when the capture source disappears"` used the same fake and asserted a strict subset of `"keeps the partial recording when the capture breaks"`.

**Verification**: `pnpm format` (clean), `pnpm turbo run lint --concurrency=1` (13/13, 0 errors), `pnpm check-types` (14/14), `pnpm knip` (exit 0), and the two affected desktop suites (48 passed). The Swift changes could not be compiled locally — this sandbox has no Swift toolchain — so `build-macos-arm64` is what validates the two SDK symbol spellings.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 40f99fba725c10020e891ba289de737ff926e0de
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Re-review of the full diff at `40f99fb`, after the fix commit addressed the five findings raised on `70e1f79`.

The PR splits `SCStreamErrorUserStopped` from a genuine capture fault so ending the share from the macOS indicator reads as a finished recording, and keeps the session finalizable on that path so `stop` still writes the `.clicks.json` click track. All five prior findings are resolved, and re-reading the whole diff surfaced no new P0 or P1.

The two that mattered are properly closed rather than papered over. The classifier now takes both operands from ScreenCaptureKit (`SCStreamError.Code.userStopped.rawValue`, `SCStreamError.errorDomain`), so the predicate that decides finished-vs-broken can no longer be wrong about its own constants — which is what made the previous version's tests unfalsifiable. And `handleStop` now releases through a `defer` gated on `session.isTerminal`, which closes a leak that was in fact wider than the review described: `session.stop()` reaches `.stopped` via `transition(.stop)` before any of the throwing work, so a finalize that failed partway already stranded an unstoppable session in the store, and this PR's external-stop path made that reachable.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic: no unhandled edge cases or wrong assumptions | Pass | `main.swift:719-727` — `defer` releases when `stop` reached `.stopped`, and correctly does not release when `transition(.stop)` itself rejects and the session is still legitimately `.recording`. No lock inversion: `isTerminal` takes the session lock, `remove` the store lock, and `stop()` holds neither on exit. |
| `REVIEW.md` | API misuse / wrong assumptions about a framework | Pass | `StreamStopReason.swift:20,26` — both values now come from the SDK, so if they compile they match what macOS stamps on an `SCStreamError`. |
| `REVIEW.md` | Security: no injection, hardcoded secrets, or missing auth on new surfaces | Pass | No new endpoint, credential, or IPC channel; the helper line protocol and its validation in `desktop-recorder-native.ts` are untouched. |
| `REVIEW.md` | Type safety: no `any`, no overly broad types | Pass | `new Set<DesktopRecorderErrorCode>` is narrowed against the union; `collectAfterExternalStop` takes `DesktopRecorderError \| null`; no `any` or assertion in the diff. |
| `REVIEW.md` | Style & maintainability: no duplicated logic, comments explain WHY | Pass | The false invariant comment in `handleState` is corrected and the redundant controller test is gone. One non-blocking nit below. |
| `REVIEW.md` | Feature regression guard: changed existing behavior needs coverage of the path most likely to regress | Pass | `desktop-tray-menu.ts:446` gates on `UNDELIVERED_RECORDING_ERROR_CODES`, and `desktop-tray-menu.test.ts:748` pins that `signed_out` with a stale `lastRecording` does not offer redelivery. |
| `REVIEW.md` | Testing coverage: a `fix:` commit carries a regression test | Pass | Three Swift classifier tests, four controller tests, two tray tests — both sides of the changed gate. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No `packages/db/src/schema/**`, `api-contracts`, Zod response schema, or `jsonb().$type<>()` change. |
| `REVIEW.md` | Event sourcing / optimistic events | N/A | No persistent or optimistic event, event ID, or projection is touched. |
| `docs/bad-smell.md` | §1 zero `any`; §2 zero lint/type suppressions | Pass | No `any` and no `eslint-disable`/`@ts-expect-error`/`prettier-ignore` in the diff. |
| `docs/bad-smell.md` | §3 no unnecessary try/catch; fail fast | Pass | The single `try/catch` wraps `backend.stop` on a genuinely recoverable path and records the capture reason instead of swallowing it. |
| `docs/bad-smell.md` | §5 no dynamic `import()` | Pass | All imports in the changed TypeScript are static. |
| `docs/bad-smell.md` | §6 no hardcoded environment-specific values | Pass | Resolved — the two ScreenCaptureKit literals are gone. |
| `docs/bad-smell.md` | §7 no fallback/recovery logic that hides a problem | Pass | Resolved — the dead `failure` field and its `describedState` branch are deleted. |
| `docs/fallback.md` | §1 no negative tests asserting removed behavior stays removed | Pass | The new tray test asserts a live product rule (do not offer redelivery of an already-delivered capture), not the absence of deleted code. |
| `docs/fallback.md` | §2 a non-GA feature switch needs no compatibility code | Pass | `DesktopScreenRecording` is `enabled: false`, scoped to `9fd4ee92`; the diff adds no dual-read, migration, or rollout branch. |
| `docs/fallback.md` | §3 no fallback on a state the owning contract cannot produce | Pass | The remaining `failure ?? { code: "capture_failed" }` in `collectAfterExternalStop` guards a genuinely nullable parameter — `null` is the clean-stop case — not a contract-guaranteed field. |
| `docs/fallback.md` | §5 delete the reader in the same PR that retires the producer | Pass | All three readers of the retired `markFailed` are handled: the `failure` branch deleted, the `handleState` reap deleted, and `isTerminal` re-homed to `handleStop` where it is live. |
| `docs/fallback.md` | §9 every fallback introduced by the diff is declared in the PR summary | Pass | The diff introduces no cross-version or compatibility fallback, so no `Fallbacks` section is required. |
| `docs/deployment-compatibility.md` | Cross-version compatibility for independently deployed surfaces | N/A | The doc scopes itself to frontend, backend, and runner and states the desktop app is outside the model; the JS side and the Swift helper ship in one bundle via `resolveNativeHelperPath`, so no skew exists between them. |
| `docs/testing.md` | Integration tests at the entry point; unit tests only for the narrow exceptions | Pass | Controller tests drive the public `DesktopRecorderController` API; the Swift classifier is the allowed pure/algorithmic exception. |
| `docs/testing.md` | Mock at the boundary only — no relative `vi.mock()` paths | Pass | No `vi.mock` at all; the native helper is injected as a fake backend. |
| `docs/testing.md` | No fake timers | Pass | No `vi.useFakeTimers()` / `advanceTimersByTime`. |
| `docs/testing.md` | Test behavior, not mocks | Pass | Each controller test asserts observable state and `openReview`/`deliver` outcomes; the new tray test asserts rendered menu labels. |
| `docs/testing.md` | New user-facing behavior gated behind a `FeatureSwitchKey` | Pass | Reached only through `setFeatureEnabled` / `FeatureSwitchKey.DesktopScreenRecording`. |
| `docs/testing/desktop-testing.md` | Choose the smallest matching entry point; Swift policy/algorithm tests stay in SwiftPM | Pass | `StreamStopReasonTests` sits in `Tests/ScreenRecorderCoreTests`; controller and tray tests stay on their existing boundaries. |
| `docs/testing/desktop-testing.md` | Keep Desktop source modules real; mock only external system boundaries | Pass | Real controller, real tray builder, real types; only the helper process is faked. |
| `docs/testing/desktop-testing.md` | Tray menu state matrices are an accepted narrow exception | Pass | Both new tray cases extend the existing `desktop-tray-menu.test.ts` matrix. |
| `docs/testing/patterns.md` | Pattern 2: pure-function tests only with a complexity/security justification | Pass | Resolved — the classifier's operands are now compiler-guaranteed, so the test verifies the comparison rather than restating literals it also supplies. |
| `docs/testing/anti-patterns.md` | AP-1: do not assert mock calls instead of behavior | Pass | `expect(backend.stop).toHaveBeenCalledWith("session-1")` always accompanies a state or delivery assertion. |
| `docs/testing/anti-patterns.md` | AP-4: do not mock internal code | Pass | No internal module is mocked. |
| `docs/testing/anti-patterns.md` | AP-8: over-testing | Pass | Resolved — the subsumed `"ends the session when the capture source disappears"` test is removed. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Nits (non-blocking, do not block merge)

- `main.swift:465-470` — `reportedStatusLocked()` returns the string literals `"stopped"` and `"failed"`, which duplicate `RecorderSessionState.stopped.rawValue` and `.failed.rawValue`. That enum is documented as the lifecycle "as reported to the Electron side", so using its raw values would keep the wire contract in one place. Pre-dates the fix commit and is caught on the JS side by `toNativeStatus`, so it is a readability point rather than a defect.
- `desktop-tray-menu.ts:32-45` — `UNDELIVERED_RECORDING_ERROR_CODES` is inserted between `MAX_RECENT_COMMANDS` and `MAX_COMMAND_LABEL_LENGTH`, splitting the two related `MAX_` constants. Purely cosmetic.
- Worth noting, not fixing here: `state == .failed` is now unreachable — `markFailed` was its only writer and `RecorderTransitionPolicy` never returns it. The `"failed"` value is still live on the wire via `reportedStatusLocked()`, and `RecorderSessionState.failed` is still exercised by `RecorderSessionStateTests`, so nothing is dead in a way this PR must clean up. If the enum case is genuinely never coming back, collapsing it is a reasonable follow-up.

### Testing

- **Coverage**: Adequate. The `fix:` commit carries regression tests on every changed behavior: the classifier's three cases including the wrong-domain guard, the four controller paths (clean finish delivers, broken capture is kept but not shipped, a salvaged recording delivers on request, a failed salvage reports the capture reason), and both sides of the tray gate. One structural gap, accepted by convention rather than a finding: `handleStop`'s session-store bookkeeping lives in the `ScreenRecorderHelper` executable target, which has no SwiftPM test harness — `docs/testing/desktop-testing.md` routes native coverage to `ScreenRecorderCore` policy tests and the JS-side fake-helper protocol boundary, and building a harness for the executable target is out of scope here.
- **Conventions**: Compliant. No `vi.mock` of internal modules, no fake timers, no direct `fetch` stubbing; the helper is injected at the process boundary per `docs/testing/desktop-testing.md`. Assertions are behavioral rather than mock-call-only.
- **Local verification of the fix commit**: `pnpm format` clean, `pnpm turbo run lint --concurrency=1` 13/13 with 0 errors, `pnpm check-types` 14/14, `pnpm knip` exit 0, and the two affected desktop suites at 48 passed.
- **Residual risk to watch**: the Swift changes could not be compiled locally — the sandbox has no Swift toolchain — so `build-macos-arm64` is the gate on the two SDK symbol spellings (`SCStreamError.Code.userStopped.rawValue`, `SCStreamError.errorDomain`). It is still running at the time of this review; if either name differs in the ScreenCaptureKit Swift interface, that job fails loudly rather than shipping something wrong. Separately, the summary's original hardware caveat is now narrowed to a single open question — whether macOS actually delivers `userStopped` for this action — since no transcribed constant can be wrong about it any more. Still worth confirming on the next dogfood run that the same action reads as a normal finish and leaves a `.clicks.json` behind.

PR #30719, opened 2026-09-01
feat(platform): print mark notes on the image and slim the selection

tongx-blip:
Four changes to the composer image annotation editor (behind `ComposerImageAnnotation`).

### Notes are printed on the image

A mark's note only travelled as prompt text, so the vision model had to match a sentence to a region by position alone. The note is now drawn on the image — in the editor, in the read-only viewer, and in the flattened copy the model actually sees — next to the mark it explains.

- Placement is automatic under the mark and follows it, until the user drags the note. From then on `noteBox` is stored and the note stays where it was put.
- A grip on the right edge sets the wrap width; the height follows from the text.
- The note is selected and dragged in its own right, so arranging a sentence never moves the box it describes.
- The flatten wraps the text against the stored width and prints it on its own ground, because a halo does not keep a sentence readable over an arbitrary screenshot.

### Selection no longer double-strokes

A selected mark drew its own border *and* a selection outline offset 3px outside it — two strokes around one shape, which reads as a rendering fault rather than a state. The outline is gone; the handles carry selection.

### Edge handles

The eight grips are now four corners and four edges. An edge drag changes one dimension and leaves the other untouched, so squaring up a box no longer means chasing two corners. Only the corners are drawn: four dots plus four bars around a small mark is more furniture than the mark itself, so the edges are invisible strips that announce themselves through the cursor. Edges paint first so the corner dots keep their hit area.

### "Attach marks" disabled until something changes

It was enabled on an untouched image. The session now records the annotation it opened on, and the button follows whether the draft has moved from it — which also means undoing back to the start disables it again.

## Tests

- Attaching is disabled on open, enabled after a mark, and disabled again after undo
- A note typed on a mark appears on the image, and selecting the note offers its width grip rather than the mark's handles
- Dragging the top edge changes height and leaves the horizontal position alone

## Checks

- `pnpm -F @okouai/app lint` — clean
- `tsc -p tsconfig.production.json --noEmit` — clean
- `pnpm knip` — clean
- Tests: `image-annotation` (12), `attachment-chips`/`chat-draft`/`chat-lifecycle` (108), `src/signals/` (251), `api-contracts` (657)

## Fallbacks

- `turbo/apps/platform/src/signals/okou-page/image-annotation.ts` `noteOnImage`: `mark.noteBox ?? defaultNoteBox(mark)` — the unset state is the intended "never hand-placed, follows its mark" behavior of a genuinely optional field (`docs/fallback.md` §6.3/§6.4). Surface none, no rollout window, no removal condition: this is permanent product behavior, not a cross-version bridge.

## Review follow-up (b164698)

Addressed the four P1 findings from the review of `cd8fac6`:

1. A note under a mark at the bottom of the image was printed past the edge and cropped out of the flattened copy, silently, because the text still travelled in the prompt. Placement flips the note above its mark when there is no room below, `clampNoteBox` is the single owner of a legal box for both the default and the drag, and the flatten fits the box once the wrapped height is known. Covered by a new test that would fail without the flip.
2. `markInkOf`'s `REDACT_FILL` branch was unreachable — `noteOnImage` had already excluded every inkless shape. Removed; `noteOnImage` returns the ink from the mark it narrows.
3. The mark and note selections were two signals held mutually exclusive by hand at eight sites, and redo plus mark-removal cleared only the mark half. They are one `{ kind, id } | null` signal now.
4. The selected note drew an outline on top of its own border — the same double stroke this PR removes from marks. The width grip carries selection instead.

### DIFF
--- a/turbo/apps/platform/src/signals/okou-page/flatten-annotated-image.ts
+++ b/turbo/apps/platform/src/signals/okou-page/flatten-annotated-image.ts
@@ -234,27 +234,28 @@ function drawNote(
 
   const boxWidth = note.box.width * scale.width;
   const lines = wrapNote(context, note.text, boxWidth - padding * 2);
-  const x = note.box.x * scale.width;
-  const y = note.box.y * scale.height;
   const boxHeight = lines.length * lineHeight + padding * 2;
+  const x = note.box.x * scale.width;
+  // The box is clamped in normalized space before it gets here, but the height
+  // is only known once the text has wrapped. A long note near the bottom would
+  // still run past the edge and be cropped out of the image the model reads.
+  const y = Math.max(
+    0,
+    Math.min(note.box.y * scale.height, scale.height - boxHeight),
+  );
 
   context.fillStyle = NOTE_GROUND;
-  context.strokeStyle = markInkOf(mark);
+  context.strokeStyle = note.ink;
   context.lineWidth = px(scale, HALO_WIDTH_UNITS);
   context.beginPath();
   context.roundRect(x, y, boxWidth, boxHeight, px(scale, CORNER_RADIUS_UNITS));
   context.fill();
   context.stroke();
 
-  context.fillStyle = markInkOf(mark);
+  context.fillStyle = note.ink;
   for (const [index, line] of lines.entries()) {
     context.fillText(line, x + padding, y + padding + index * lineHeight);
   }
-}
-
-/** The ink a note is printed in, which is the ink of the mark it explains. */
-function markInkOf(mark: ImageAnnotationMark): string {
-  return "ink" in mark ? mark.ink : REDACT_FILL;
 }
 
 function drawMark(
--- a/turbo/apps/platform/src/signals/okou-page/flatten-annotated-image.ts
+++ b/turbo/apps/platform/src/signals/okou-page/flatten-annotated-image.ts
@@ -5,6 +5,8 @@ import type {
 import {
   HIGHLIGHT_FILL,
   markOrdinal,
+  NOTE_GROUND,
+  noteOnImage,
   REDACT_FILL,
   STROKE_HALO_INNER,
 } from "./image-annotation.ts";
@@ -22,6 +24,9 @@ const HALO_WIDTH_UNITS = 1;
 const PIN_RADIUS_UNITS = 11;
 const PIN_FONT_UNITS = 13;
 const TEXT_FONT_UNITS = 18;
+const NOTE_FONT_UNITS = 15;
+const NOTE_LINE_UNITS = 20;
+const NOTE_PADDING_UNITS = 6;
 const ARROW_HEAD_UNITS = 18;
 const CORNER_RADIUS_UNITS = 4;
 
@@ -174,6 +179,82 @@ function drawPin(
   context.textAlign = "center";
   context.textBaseline = "middle";
   context.fillText(String(ordinal), x, y);
+}
+
+/**
+ * Breaks `text` into lines that fit `maxWidth`, measuring with the font already
+ * set on the context. A word longer than the box (a URL, a long identifier) is
+ * left to overhang rather than being cut, because a truncated instruction is
+ * worse than an untidy one.
+ */
+function wrapNote(
+  context: CanvasRenderingContext2D,
+  text: string,
+  maxWidth: number,
+): string[] {
+  const lines: string[] = [];
+  for (const paragraph of text.split("\n")) {
+    let line = "";
+    for (const word of paragraph.split(/\s+/u).filter(Boolean)) {
+      const candidate = line ? `${line} ${word}` : word;
+      if (line && context.measureText(candidate).width > maxWidth) {
+        lines.push(line);
+        line = word;
+      } else {
+        line = candidate;
+      }
+    }
+    lines.push(line);
+  }
+  return lines;
+}
+
+/**
+ * Prints a mark's note onto the image next to the region it is about.
+ *
+ * This is the whole point of flattening rather than only sending the notes as
+ * prompt text: the model sees one picture, and the sentence has to be in it.
+ */
+function drawNote(
+  context: CanvasRenderingContext2D,
+  scale: Scale,
+  mark: ImageAnnotationMark,
+): void {
+  const note = noteOnImage(mark);
+  if (!note) {
+    return;
+  }
+
+  const fontSize = px(scale, NOTE_FONT_UNITS);
+  const lineHeight = px(scale, NOTE_LINE_UNITS);
+  const padding = px(scale, NOTE_PADDING_UNITS);
+  context.font = `600 ${fontSize}px "Noto Sans", system-ui, sans-serif`;
+  context.textAlign = "left";
+  context.textBaseline = "top";
+
+  const boxWidth = note.box.width * scale.width;
+  const lines = wrapNote(context, note.text, boxWidth - padding * 2);
+  const x = note.box.x * scale.width;
+  const y = note.box.y * scale.height;
+  const boxHeight = lines.length * lineHeight + padding * 2;
+
+  context.fillStyle = NOTE_GROUND;
+  context.strokeStyle = markInkOf(mark);
+  context.lineWidth = px(scale, HALO_WIDTH_UNITS);
+  context.beginPath();
+  context.roundRect(x, y, boxWidth, boxHeight, px(scale, CORNER_RADIUS_UNITS));
+  context.fill();
+  context.stroke();
+
+  context.fillStyle = markInkOf(mark);
+  for (const [index, line] of lines.entries()) {
+    context.fillText(line, x + padding, y + padding + index * lineHeight);
+  }
+}
+
+/** The ink a note is printed in, which is the ink of the mark it explains. */
+function markInkOf(mark: ImageAnnotationMark): string {
+  return "ink" in mark ? mark.ink : REDACT_FILL;
 }
 
 function drawMark(
@@ -344,6 +425,9 @@ export async function flattenAnnotatedImage(
   for (const [index, mark] of annotation.marks.entries()) {
     context.save();
     drawMark(context, scale, mark, markOrdinal(mark, index));
+    context.restore();
+    context.save();
+    drawNote(context, scale, mark);
     context.restore();
   }
 
--- a/turbo/apps/platform/src/signals/okou-page/image-annotation.ts
+++ b/turbo/apps/platform/src/signals/okou-page/image-annotation.ts
@@ -323,12 +323,18 @@ export const bindAnnotationSurface$ = onRef<HTMLElement>(
 const internalSession$ = state<AnnotationSession | null>(null);
 const internalTool$ = state<AnnotationTool>("box");
 const internalInk$ = state<AnnotationInk>(DEFAULT_ANNOTATION_INK);
-const internalSelectedMarkId$ = state<string | null>(null);
 /**
- * The note label being edited in place, held separately from the mark so that
- * dragging a sentence into clear space does not also move the box it explains.
+ * What is currently held: a mark, or the note label of a mark. One signal
+ * rather than two, because the two are mutually exclusive and holding that
+ * invariant by hand across every clear site is how a stale grip survives an
+ * operation that was supposed to drop the selection.
  */
-const internalSelectedNoteId$ = state<string | null>(null);
+interface AnnotationSelection {
+  readonly kind: "mark" | "note";
+  readonly id: string;
+}
+
+const internalSelection$ = state<AnnotationSelection | null>(null);
 
 export const annotationSessionTarget$ = computed((get) => {
   return get(internalSession$)?.target ?? null;
@@ -351,20 +357,17 @@ export const annotationInk$ = computed((get) => {
 });
 
 export const annotationSelectedMarkId$ = computed((get) => {
-  return get(internalSelectedMarkId$);
+  const selection = get(internalSelection$);
+  return selection?.kind === "mark" ? selection.id : null;
 });
 
 export const annotationSelectedNoteId$ = computed((get) => {
-  return get(internalSelectedNoteId$);
+  const selection = get(internalSelection$);
+  return selection?.kind === "note" ? selection.id : null;
 });
 
 export const selectAnnotationNote$ = command(({ set }, id: string | null) => {
-  set(internalSelectedNoteId$, id);
-  // A note and its mark are two different things to hold; selecting one has to
-  // let go of the other or both sets of handles sit on the image at once.
-  if (id !== null) {
-    set(internalSelectedMarkId$, null);
-  }
+  set(internalSelection$, id === null ? null : { kind: "note", id });
 });
 
 export const moveAnnotationNoteBox$ = command(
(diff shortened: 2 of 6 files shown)


tongx-blip:
Changes Requested


PR_AUTO_REVIEW_COMMIT: cd8fac656884861260e87d2ed4028d8f29d484b4

### Summary

Prints each mark's note onto the annotated image (editor, viewer, and the flattened copy the vision model reads), replaces the double-stroked selection ring with handle-only selection, extends the resize grips from four corners to four corners plus four edges, and disables `Attach marks` until the draft differs from what the editor opened on. All behind the non-GA `ComposerImageAnnotation` switch (`enabled: false`, `enabledEmailHashes: ["56bef1aa"]`).

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and edge cases | Finding (P1) | `flatten-annotated-image.ts` `drawNote` never clamps the note to the canvas; a mark near the bottom prints its note off the image. |
| `REVIEW.md` | Type safety — no `any`, no broad types | Pass | No `any` introduced; `AnnotationResizeEdge` is a literal union, `noteOnImage` returns a discriminated result. |
| `REVIEW.md` | Feature regression guard — changed behavior needs coverage | Pass | `Attach marks` changing from always-enabled to dirty-gated is covered by the rewritten "keeps an opened-but-untouched image unannotated" test (disabled → enabled → disabled → Cancel). |
| `REVIEW.md` | New user-facing feature gated by a `FeatureSwitchKey` | Pass | `composerImageAnnotationEnabled$` gates the editor entry; switch is `enabled: false`. |
| `REVIEW.md` | DB/JSONB gate — persisted shape changes | Pass | `noteBox` is `.optional()` on a non-`.strict()` object in `chat_threads.draft_attachments`; old rows parse unchanged and no migration is needed. |
| `docs/bad-smell.md` | Zero `any`, zero lint/type suppressions | Pass | No suppression comments in the diff; `pnpm -F @okouai/app lint` clean. |
| `docs/bad-smell.md` | No dynamic `import()` | Pass | All imports are static. |
| `docs/bad-smell.md` | Functions stay extractable, no duplicated logic | Pass | `EditorStage` was split into `NoteLayer` and `useGrabHandle` to stay under the line limit; note geometry lives once in `noteOnImage`/`defaultNoteBox`. |
| `docs/fallback.md` | §3 no fallback on a state the contract cannot produce | Finding (P1) | `flatten-annotated-image.ts` `markInkOf` returns `REDACT_FILL` for marks without `ink`, but `drawNote` already excluded every inkless shape. |
| `docs/fallback.md` | Every introduced fallback declared in the PR summary | Finding (P1) | The PR body has no `Fallbacks` section; the diff introduces two (see below). |
| `docs/fallback.md` | §2 non-GA feature needs no compatibility code | Pass | No dual-read, no migration, no old-shape tolerance added for this feature's own earlier iterations. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | The removed selection outline has no tombstone test; the updated test asserts current behavior. |
| `docs/testing.md` | Integration tests at the entry point, not unit tests | Pass | All three new tests drive the composer and editor through the rendered page. |
| `docs/testing.md` | `feat:` requires integration tests | Pass | Three new tests cover notes-on-image, note selection handing over its own grip, and edge-only resize. |
| `docs/testing/anti-patterns.md` | AP-4 no `vi.mock()` on relative paths; AP-5 no fake timers; AP-1 assert behavior not calls | Pass | No `vi.mock` and no `vi.useFakeTimers` in the diff; assertions read rendered DOM and geometry. |
| `docs/testing/app-testing.md` | Platform tests use the shared context and MSW handlers, no external requests | Pass | Tests use `testContext()` / `context.mocks.api` and `detachedSetupPage`. |
| `docs/testing/testing-external-behavior.md` | Assert what a user can observe | Pass | Assertions are on the rendered label text, the presence of grips, and inline style geometry. |
| `docs/effect.md` | Derived value uses `computed`; user event uses a semantic command | Pass | `annotationDirty$` is a `computed`; `selectAnnotationNote$` / `moveAnnotationNoteBox$` are named after the action. |
| `docs/effect.md` | Keep React render pure, own signals outside render | Pass | No React hooks; `NoteLayer` reads signals via `useGet`/`useSet` only. |
| `docs/cache.md` | RC-5 no duplicate mutable sources of truth | Finding (P1) | `internalSelectedMarkId$` and `internalSelectedNoteId$` are mutually exclusive by hand at eight call sites, and two of them miss the note half. |
| `docs/cache.md` | RC-2 no render-time signal allocation | Pass | All signals are module-scope; `NoteLayer` allocates none. |
| `.claude/skills/ccstate/SKILL.md` | DOM nodes via `onRef`, not refs/state | Pass | The surface still binds through `bindAnnotationSurface$` (`onRef`); no new DOM state. |
| `docs/deployment-compatibility.md` | Independently deployed components tolerate the other version | Pass | New optional JSONB key; an old API strips it and the note falls back to auto-placement. Non-GA switch limits producers to one account. |

### Findings

- `turbo/apps/platform/src/signals/okou-page/flatten-annotated-image.ts` (`drawNote`): the note is drawn at `note.box.y * scale.height` and grows downward by `lines.length * lineHeight + padding * 2`, with no clamp against the canvas. `defaultNoteBox` places a note at `bounds.y + bounds.height + NOTE_GAP` clamped only to `<= 1`, so a mark near the bottom of a screenshot prints its note past the bottom edge and it is cropped out of the flattened image. A hand-dragged note has the same problem horizontally: `draggedRect`'s `note-move` branch clamps `x` to `0..1` without accounting for `width`, so `x = 0.95, width = 0.3` overflows the right edge. This defeats the stated purpose of the change — the sentence is missing from the one artifact the model looks at — and it fails silently, because `describeAnnotation` still sends the text so nothing looks broken. Clamp the box into the canvas (and prefer flipping a bottom-edge note above its mark) in both the flatten and the editor label.

- `turbo/apps/platform/src/signals/okou-page/flatten-annotated-image.ts` (`markInkOf`): `"ink" in mark ? mark.ink : REDACT_FILL` guards a state the contract cannot produce. `drawNote` returns early unless `noteOnImage(mark)` is non-null, and `noteOnImage` already excludes `text`, `redact`, and `highlight`; the remaining `box`, `arrow`, and `pen` all declare `ink` as required in `imageAnnotationMarkSchema`. Per `docs/fallback.md` §3 this branch should be deleted and the ink read directly, with the narrowing done where the shape is already known.

- `turbo/apps/platform/src/signals/okou-page/image-annotation.ts` (`redoAnnotation$` line ~692, `removeAnnotationMark$` line ~517): the mark selection and the note selection are two signals held mutually exclusive by hand across eight sites, and these two clear only `internalSelectedMarkId$`. Redoing while a note is selected leaves `internalSelectedNoteId$` pointing at that note, so the width grip survives an operation that was supposed to drop the selection; deleting a mark whose note is selected leaves a dangling id. This is the `docs/cache.md` RC-5 hazard concretely — either clear both at every site, or model the selection as one signal (`{ kind: "mark" | "note"; id: string } | null`) so the invariant cannot be missed.

- `turbo/apps/platform/src/views/okou-page/image-annotation-marks.tsx` (`MarkNoteLabel`, line ~262): the selected note draws `outline: 2px solid <ink>` on top of the label's own `border`, which is exactly the double-stroke this PR removes from `MarkShape` and exactly what the reporter asked to be taken off. The note already gains a visible width grip when selected, so the outline is redundant for the same reason the mark's ring was.

### Fallbacks

- `turbo/apps/platform/src/signals/okou-page/image-annotation.ts` `noteOnImage`: `mark.noteBox ?? defaultNoteBox(mark)` — protects nothing across versions; the unset state is the intended "never hand-placed, follows its mark" behavior of a genuinely optional field (`docs/fallback.md` §6.3/§6.4). Surface none, no rollout window, removal condition none (it is permanent product behavior, not a bridge). Declared in PR summary **No — P1**. Verdict **Justified**.
- `turbo/apps/platform/src/signals/okou-page/flatten-annotated-image.ts` `markInkOf`: `"ink" in mark ? mark.ink : REDACT_FILL` — guards inkless shapes that `drawNote` has already excluded. Surface none, no rollout window, removal condition immediate. Declared in PR summary **No — P1**. Verdict **Not justified**; delete it.

### Testing

- Coverage: Adequate for the interaction changes — attach-gating across all three states, the note appearing on the image and handing over its own grip, and edge-only resize proving the perpendicular axis does not move. Insufficient for the flatten: `drawNote` has no coverage, which is why the clipping above went unnoticed. jsdom cannot exercise a real canvas, but the geometry decision is separable — extract the box placement into a pure function and test that a bottom-edge mark keeps its note inside `0..1`.
- Conventions: Compliant. No `vi.mock` on relative paths, no fake timers, no fetch stubbing, no mock-call assertions; tests drive the rendered page through `detachedSetupPage` and assert observable DOM. Minor nit, not a finding: the new test mixes `screen.queryByTestId(/regex/u)` with a raw `document.querySelector` for the same element class — worth settling on the testing-library query.

tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: b164698f68a1d99db90c43e3101116b87f7711a5
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Re-review after `b164698` addressed the four P1 findings raised against `cd8fac6`. The PR prints each mark's note onto the annotated image (editor, viewer, and the flattened copy the model reads), replaces the double-stroked selection ring with handle-only selection, extends the resize grips to four corners plus four edges, and disables `Attach marks` until the draft differs from what the editor opened on. All behind the non-GA `ComposerImageAnnotation` switch. All four findings are resolved and the `noteBox` fallback is now declared in the PR summary.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and edge cases | Pass | The bottom-edge crop is fixed in both places it could occur: `defaultNoteBox` flips a note above its mark when `below > 1 - NOTE_ROOM`, and `drawNote` re-fits with `Math.min(note.box.y * scale.height, scale.height - boxHeight)` once the wrapped height is known. |
| `REVIEW.md` | Type safety — no `any`, no broad types | Pass | `AnnotationSelection` is a literal-kind union; `noteOnImage` returns a narrowed result carrying `ink`. |
| `REVIEW.md` | Feature regression guard — changed behavior needs coverage | Pass | The attach-gating change is covered by the rewritten untouched-image test (disabled → enabled → disabled → Cancel); the note placement change by the new bottom-edge test. |
| `REVIEW.md` | New user-facing feature gated by a `FeatureSwitchKey` | Pass | `composerImageAnnotationEnabled$` gates the entry; the switch is `enabled: false` with a single `enabledEmailHashes` entry. |
| `REVIEW.md` | DB/JSONB gate — persisted shape changes | Pass | `noteBox` is `.optional()` on a non-`.strict()` object in `chat_threads.draft_attachments`; old rows parse unchanged and no migration is needed. |
| `docs/bad-smell.md` | Zero `any`, zero lint/type suppressions | Pass | No suppression comments; `@okouai/app` and `@okouai/api-contracts` lint clean. |
| `docs/bad-smell.md` | No dynamic `import()` | Pass | All imports static. |
| `docs/bad-smell.md` | No duplicated logic; single owner for a rule | Pass | `MIN_NOTE_DRAG_WIDTH` was removed from the editor so `clampNoteBox` is the only definition of a legal note box, applied to both the default and the drag. |
| `docs/fallback.md` | §3 no fallback on a state the contract cannot produce | Pass | `markInkOf`'s unreachable `REDACT_FILL` branch is deleted; `noteOnImage` returns `mark.ink` from the shape it has already narrowed. |
| `docs/fallback.md` | Every introduced fallback declared in the PR summary | Pass | The PR body now carries a `## Fallbacks` section naming the one introduced fallback with surface, window, and removal condition. |
| `docs/fallback.md` | §2 non-GA feature needs no compatibility code | Pass | No dual-read, migration, or old-shape tolerance added for this feature's own earlier iterations. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | No tombstones; the removed outline and the removed `selected` prop leave no assertions behind. |
| `docs/testing.md` | Integration tests at the entry point, not unit tests | Pass | The new placement test drives the composer and editor and asserts rendered geometry rather than calling `defaultNoteBox` directly. |
| `docs/testing.md` | `feat:` requires integration tests | Pass | Four tests now cover notes-on-image, note selection handing over its own grip, edge-only resize, and bottom-edge placement. |
| `docs/testing/anti-patterns.md` | AP-4 no relative-path `vi.mock`; AP-5 no fake timers; AP-1 assert behavior not calls | Pass | None present in the diff. |
| `docs/testing/app-testing.md` | Platform tests use the shared context and MSW handlers | Pass | `testContext()` / `context.mocks.api` / `detachedSetupPage` throughout. |
| `docs/testing/testing-external-behavior.md` | Assert what a user can observe | Pass | The placement test reads `style.top` / `style.left` / `style.width` off the rendered label. |
| `docs/effect.md` | Derived value uses `computed`; user event uses a semantic command | Pass | `annotationSelectedMarkId$` and `annotationSelectedNoteId$` are now projections of one `computed`; selection commands stay named after the action. |
| `docs/effect.md` | Keep React render pure, own signals outside render | Pass | No React hooks; all signals module-scope. |
| `docs/cache.md` | RC-5 no duplicate mutable sources of truth | Pass | The two selection signals are collapsed into one `AnnotationSelection | null`, so mark/note exclusivity is structural. `removeAnnotationMark$` now clears on `selection?.id === id`, covering either kind. |
| `docs/cache.md` | RC-2 no render-time signal allocation | Pass | None introduced. |
| `.claude/skills/ccstate/SKILL.md` | DOM nodes via `onRef`, not refs/state | Pass | The surface still binds through `bindAnnotationSurface$`. |
| `docs/deployment-compatibility.md` | Independently deployed components tolerate the other version | Pass | New optional JSONB key; an old API strips it and the note falls back to auto-placement. Non-GA switch limits producers to one account. |

### Findings

None. The four P1 findings from the previous review are resolved:

- **Note cropped out of the flattened image** — fixed in both the normalized placement (`defaultNoteBox` flips above the mark when there is no room below; `clampNoteBox` bounds x by `1 - width` and y by `1 - NOTE_ROOM`) and the pixel-space draw (`drawNote` fits the box against `scale.height - boxHeight` after wrapping). Verified by a test that fails with `expected 94 to be less than 83` when the flip is reverted.
- **Unreachable `markInkOf` branch** — deleted; `noteOnImage` carries `ink` from the narrowed mark.
- **Hand-maintained selection invariant** — replaced by a single `{ kind, id } | null` signal, which removes both missed clear sites (`redoAnnotation$`, `removeAnnotationMark$`) by construction rather than by adding two more calls.
- **Double stroke on the selected note** — the outline is gone; the width grip carries the selection, matching how a selected mark now works.

### Testing

- Coverage: Adequate. The previously untested placement decision is now exercised through the UI, and reverting the fix turns it red. `drawNote`'s canvas rendering still has no direct coverage — jsdom cannot run a real canvas — but the geometry that produced the bug is now asserted in the editor, which is where it is decidable.
- Conventions: Compliant. No `vi.mock` on relative paths, no fake timers, no fetch stubbing, no mock-call assertions.
- Nits, not findings: the placement test asserts `top < 100` and `top < 83`, where the first is subsumed by the second, and `83` is derived from `NOTE_GAP + NOTE_ROOM` rather than named — a constant-derived expectation would survive a future tuning of those values. The file also mixes `screen.getByTestId(/regex/u)` with a raw `document.querySelector` for the same element class.

Issue #30676, opened 2026-09-01
perf(platform): reduce app lint feedback time


seven332:
## Background

`pnpm -F @okouai/app lint` is the platform workspace's complete lint entry point. It currently runs static-asset validation, emoji-data validation, localized-UI validation, build-configuration tests, i18n extraction and status reporting, regular oxlint, type-aware oxlint, and ESLint serially.

A local stage-by-stage investigation on 2026-09-01, after running `scripts/prepare.sh`, found these dominant costs:

- i18n extraction and status: about 100-141 seconds across observed runs
- ESLint: about 41 seconds without profiling in the representative completed run
- build-configuration validation: about 18 seconds
- regular and type-aware oxlint combined: about 10 seconds

The shared development host was heavily contended during part of the investigation, so absolute wall times varied. The stage ordering and duplicated work were stable across runs.

## Objective

Reduce cold and repeated feedback time for `@okouai/app` lint while preserving every validation currently enforced by the command.

## Delivery slices

- Avoid extracting platform i18n keys twice in one lint run.
- Cache repeated platform ESLint runs safely.
- Reduce the dominant costs of uncached platform ESLint runs.
- Remove redundant production-build work from platform build-configuration lint validation.

Each slice should be independently measurable and reviewable. Shared-host scheduling and unrelated platform build performance are outside this objective.

## Completion criteria

- All linked sub-issues are complete.
- `pnpm -F @okouai/app lint` still executes and passes every existing validation category; no lint rule, locale, build invariant, or test is skipped.
- Cold and warm before/after measurements are recorded using the same prepared environment and command scope.
- The resulting command structure does not rely on background checks whose failures can be missed.

## Relevant files

- `turbo/apps/platform/package.json`
- `turbo/apps/platform/i18next.config.ts`
- `turbo/apps/platform/eslint.config.js`
- `turbo/packages/eslint-config/base.js`

PR #30721, opened 2026-09-01
perf(platform): consolidate build config verification

### LINKED ISSUE #30680: perf(platform): reduce build-config verification overhead in lint
## Background

`@okouai/app` lint runs the required `lint:build-config` verification before i18n and static linting:

```text
node --test scripts/check-production-minification.node.mjs &&
tsx --test scripts/check-single-bundle.node.ts
```

These checks protect production minification, bundle topology, worker origin, deterministic output, and dependency boundaries. They are part of the lint-time performance objective tracked by #30676 and must remain required.

## Problem and evidence

The original issue attributed the observed cost to redundant production builds. Repository inspection does not support that diagnosis:

- `check-production-minification.node.mjs` does not call Vite `build()`; it loads the production Vite configuration twice for two groups of assertions.
- `check-single-bundle.node.ts` calls Vite `build()` exactly once. Its topology, HTML, worker, asset, and dependency assertions already share that single fixture result.
- A representative prepared-checkout run at commit `8a478861b73da1995caf8e24dbb54c130c156970` measured 4.59 seconds for the production-config script and 5.23 seconds for the bundle script. The fixture-build subtest itself reported about 0.30 seconds.
- The parent investigation observed about 18 seconds on a contended host. Although absolute wall time varies, both measurements show that this validation is materially heavier than an ordinary static lint rule.

The supported bottleneck is repeated process initialization and Vite/plugin/config loading across two serial test commands, not multiple production builds. Reusing a deployment build artifact would not directly address that cost.

## Decision

**Challenge outcome: revise.** Keep this PR-sized performance slice, but target build-configuration verification overhead rather than an unobserved redundant-build problem.

- Run the production-config and fixture-bundle assertions through one required build-config test entry point so Vite and its plugin graph are initialized once.
- Load the production Vite configuration once per verification invocation and share it read-only across the configuration assertions.
- Keep exactly one fixture build and continue using its result for the integration assertions that require generated output.
- Keep `lint:build-config` in the direct `pnpm -F @okouai/app lint` path. Do not make enforcement depend on a deployment job or an optional command.
- Measure the complete verification path before and after in the same prepared environment; retain the change only when the timing demonstrates that initialization work was actually removed.

This direction addresses the measured cost without changing production build outputs or CI deployment topology.

## Alternatives considered

- **Reuse the canonical production `dist` output:** rejected. The lint and canonical app build run in different CI jobs, and direct workspace lint must remain independently complete. Cross-job artifact coupling would add complexity without targeting the dominant measured cost.
- **Move the check only to a narrowly cached Turbo task:** not the first step. It can improve repeated root-Turbo runs, but direct `pnpm -F @okouai/app lint` executes the package script and must not silently lose the verification. Consider task-level caching only if a later plan preserves both entry points and proves correct input invalidation.
- **Remove the fixture build or convert its assertions to unit-only checks:** rejected because it would weaken the Vite integration coverage required by the parent objective.
- **Keep the status quo:** valid functionally, but preserves roughly 10-18 seconds of serial feedback cost per uncached invocation.

## Risks and mitigations

- Sharing one loaded configuration could create test-order coupling. Treat it as immutable and keep assertions independent.
- Consolidating test entry points could change failure reporting or exit behavior. Verify that every failing assertion still produces a non-zero command result.
- Timing is sensitive to host contention and cold caches. Record cold and warm runs with the same commit, dependencies, environment, and command scope.
- A later caching optimization could miss relevant inputs. If pursued, explicitly verify invalidation for Vite config, build-check scripts, package metadata, and dependency changes.

## Acceptance criteria

- One `lint:build-config` invocation performs at most one production Vite-config load and exactly one fixture Vite build.
- All current production minification, compiled-version, vendor grouping, worker-origin, bundle topology, HTML discovery order, asset layout, raw-size, deterministic-output, DOM-global, server-contract, and forbidden-dependency assertions remain required.
- No build-configuration test is skipped, converted to an optional check, or moved exclusively behind deployment infrastructure.
- Cold and warm before/after timings are recorded using the same prepared environment and command scope, including the total build-config wall time and the fixture-build subtest time.
- The measurements demonstrate a reduction attributable to removed initialization/config-loading work. If they do not, do not retain added consolidation or caching complexity.
- `pnpm -F @okouai/app lint:build-config` and `pnpm -F @okouai/app lint` pass.

## Relevant files

- `turbo/apps/platform/package.json`
- `turbo/apps/platform/scripts/check-production-minification.node.mjs`
- `turbo/apps/platform/scripts/check-single-bundle.node.ts`
- `turbo/apps/platform/vite.config.ts`
- `turbo/turbo.json`

## Delivery notes

This remains one reviewable child of #30676. General platform runtime build optimization, deployment artifact sharing, and unrelated lint stages are out of scope.

seven332:
## Summary

- consolidate the platform production-config and bundle-verification checks into one TypeScript test file
- reduce two production-config loads to one while preserving all 12 test names and exactly one fixture build
- add a dedicated strict TypeScript project for the Node-based verifier without contaminating browser-test ambient types

## Scope

- no production runtime behavior changes
- no CI, Turbo pipeline, deployment, test-timeout, or bundle-policy changes

## Validation

- `pnpm -F @okouai/app lint:build-config` (12/12 tests passed)
- `pnpm -F @okouai/app lint`
- `pnpm knip`
- pre-commit `pnpm check-types` (14/14 tasks passed)
- source inventory: one `loadConfigFromFile()` call and one fixture `build()` call

## Performance

Paired runs used the same prepared checkout and command scope on a noisy shared host:

| | Run 1 | Run 2 | Run 3 | Median |
| Before wall time | 22.40s | 12.02s | 16.97s | 16.97s |
| After wall time | 15.06s | 14.40s | 12.51s | 14.40s |

- wall-clock median improved by about 15%
- fixture-build subtest remained a minority of the command: about 0.43s before and 0.27s in the latest verbose after run
- CPU-time median was about 5.83s before and 6.32s after; this small regression may be measurement noise, so the demonstrated improvement is wall-clock overhead rather than CPU consumption

Closes #30680

Parent: #30676


seven332:
## Code Review: PR #30721

**Reviewed head:** `45649b4cc1a1dd8701b10092158106f35450f571`

### Summary

The PR consolidates two required build-config test processes into one strict-TypeScript entry while preserving all 12 tests, one fixture build, direct lint enforcement, and existing cleanup behavior.

### Findings

No findings.

### Testing

The submitted head passed the 12-test build-config entry, platform type checking, complete platform lint, Knip, and the pre-commit full-repository type check (14/14 tasks). The exact test-name inventory, one-load/one-build call counts, stale references, and issue-closing metadata were also verified. Vitest was not rerun because no runtime or Vitest-selected behavior changed; submitted-head CI remains the final independent check.

### Verdict

LGTM

PR #30740, opened 2026-09-01
feat(desktop): capture a selected screen region

bingjiez666:
First half of the floating recorder bar you asked for: the capture side of
**Area**. Choosing the region is the bar itself, which is the next change.

## How an area is captured

ScreenCaptureKit has no region filter, so an area is a **display capture with a
crop**: the display is filtered as usual and `SCStreamConfiguration.sourceRect`
narrows it.

Two consequences of that are easy to get wrong, so both live in
`ScreenRecorderCore` as pure functions with tests:

- **`sourceRect` is read in the display's own coordinate space.** Passing global
  screen coordinates straight through crops the wrong region on every display
  whose origin is not `(0, 0)` — which is every secondary display. The tests use
  a display at `x: 1512, y: -200` so a policy that confuses the two fails
  visibly rather than passing by accident on the primary display.
- **The reported geometry describes the crop, not the display.** That is what
  makes the click track land in the cropped frame, and what makes a click
  outside the selection correctly count as outside the recording rather than
  being mapped to some point inside it.

## Edge handling

Dragging a selection past the edge of a screen is ordinary, so the overlap is
captured rather than the whole request being refused. A selection that misses
the display entirely, only touches its edge, or is too thin to encode is
refused with a message that says so.

## Scope

Capture side only. `prepare` accepts `sourceKind: "area"` with a rectangle in
global screen points and `sourceId` naming the display it was drawn on; nothing
in the UI produces one yet. The types keep the two ideas apart:
`DesktopRecorderSourceKind` is what `listSources` can enumerate, and an area is
not a listable source.

## Verification

`check-types`, `lint`, `knip`, desktop vitest (443 passed) locally, plus seven
new Swift tests on the area policy and two on the transport (the region reaches
the helper; a whole-display capture does not carry an empty one).

**Unverified on hardware**: that `sourceRect` crops where these coordinates say
it should. The conversion is tested, but the frame it produces is not something
CI can look at. Worth checking on the next dogfood run that a region drawn on a
secondary display records that region.

### DIFF
--- a/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderCore/CaptureArea.swift
+++ b/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderCore/CaptureArea.swift
@@ -1,0 +1,58 @@
+import Foundation
+
+/// A rectangle in global screen points, top-left origin.
+public struct AreaRect: Equatable, Sendable {
+    public let x: Double
+    public let y: Double
+    public let width: Double
+    public let height: Double
+
+    public init(x: Double, y: Double, width: Double, height: Double) {
+        self.x = x
+        self.y = y
+        self.width = width
+        self.height = height
+    }
+
+    public var maxX: Double { x + width }
+    public var maxY: Double { y + height }
+}
+
+public enum CaptureAreaPolicy {
+    /// Smallest area worth capturing, in points. Anything thinner cannot encode
+    /// to the two-pixel floor `CaptureSizePolicy` enforces.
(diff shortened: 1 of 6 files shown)


bingjiez666:
Folded into #30743.

That PR was branched from this one, so it already carries every commit here — `ddf191b feat(desktop): capture a selected screen region` is in its history and all six files in this diff appear in its 37. Closing so the region capture and the controls that drive it land as one change.

No rebase was needed and nothing is lost.

PR #30756, opened 2026-09-01
fix(platform): catch up unread threads after indicator reload

e7h4n:
## Summary

- refresh indicators in the SharedWorker and catch up every unread thread into IndexedDB for each indicator revision
- keep credential-scoped indicator subscriptions and background catch-up under Worker lifecycle ownership
- replace the app's one-time 10-thread prewarm with Worker-owned all-unread catch-up

## Root cause

The connection-isolation refactor left realtime chat-event invalidation as a broadcast-only path. The app synchronized registered active thread signals, but an unread thread with no active page subscription was no longer fetched into IndexedDB. Existing coverage exercised active-thread sync and the one-time startup prewarm, but did not model an inactive thread becoming unread after an indicator reload.

## Verification

- `pnpm format`
- `pnpm knip`
- `pnpm --filter @okouai/app exec vitest run src/shared-database/__tests__/platform-direct-bridge.test.ts`
- `pnpm --filter @okouai/app exec vitest run src/shared-database/__tests__/worker-runtime.test.ts`


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: f080314f4883a1b5f91d1edd3ce2d2e0d2a66922
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

This moves unread-thread chat-event catch-up into the credential-scoped SharedWorker indicator lifecycle. Every resolved indicator revision catches up all unread threads into IndexedDB, while the former app-owned one-time, ten-thread prewarm is removed; the regression test covers an inactive thread becoming unread and reopening from the updated cache without another API request.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness, race/lifecycle safety, type safety, security, and required regression coverage | Pass | The complete five-file diff keeps queries credential-scoped, uses the Worker root abort signal, preserves typed `SharedDatabaseQuery` inputs, and adds a fix regression story in `platform-direct-bridge.test.ts`. |
| `docs/bad-smell.md` | No `any`, suppression, dynamic import in production, hardcoded configuration, or defensive fallback | Pass | Changed production files add typed static imports and fail-fast runtime acquisition; no prohibited suppression, dynamic import, configuration default, or fallback is introduced. |
| `docs/testing.md` | Prefer integration coverage, boundary mocks, real internal code, real timers, and behavioral assertions | Pass | The test enters through `setupPage`, uses `context.mocks` for API/Ably boundaries, real SharedWorker/runtime/IndexedDB paths, and asserts cached chat-event behavior rather than mock-call counts. |
| `docs/event-sourcing.md` | Persistent events remain authoritative and reconciliation is driven by the normal event stream | Pass | Catch-up reuses the canonical `chat-event` runtime query and existing cursor/row persistence; it does not add optimistic rollback or a second event source. |
| `docs/effect.md` | Choose the correct owner/trigger and give async work an aborting lifecycle | Pass | Indicator reload and catch-up are Worker credential-daemon commands; all async work receives the credential root `AbortSignal` and is awaited before indicator invalidation is broadcast. |
| `docs/cache.md` | Avoid manual caches and align state/resource lifetime with the owning scope | Pass | `workerChatThreadIndicators$` relies on ccstate computed memoization per indicator revision and the credential Store lifecycle; no keyed or process-lifetime manual cache is added. |
| `.claude/skills/ccstate/SKILL.md` | Prefer reactive computed state, command composition, owned abort signals, and post-await abort checks | Pass | The implementation composes commands via `set`, reads the reactive indicator computed via `get`, passes the Worker root signal to every runtime query, and checks abortion after non-signal-aware awaits. |
| `docs/deployment-compatibility.md` | Preserve independently deployed protocol and persisted-state compatibility | N/A | No API contract, IndexedDB schema, event row shape, or cross-version reader/writer format changes; the diff only changes when existing canonical queries run. |
| `docs/react-commit.md` | Narrow React subscriptions and validate equality/performance claims | N/A | No React component, hook subscription, equality function, or React performance claim changes in this PR. |
| `docs/testing/patterns.md` | Exercise Platform behavior through production bootstrap and shared test context | Pass | `platform-direct-bridge.test.ts` uses `testContext`, `setupPage`, production signals, and boundary mocks rather than a helper unit test. |
| `docs/testing/anti-patterns.md` | Avoid internal mocks, direct fetch mocks, fake timers, and implementation-only call assertions | Pass | The regression adds none of those patterns; its `idb` substitution is the existing package-boundary test adapter and assertions verify persisted event availability. |
| `docs/testing/app-testing.md` | Use page/bootstrap setup, `context.mocks`, and lifecycle-owned cleanup for Platform tests | Pass | The background-only SharedWorker scenario uses `setupPage({ withoutRender: true })`, `context.mocks`, `context.signal`, and child abort controllers tied to that lifecycle. |
| `docs/testing/testing-external-behavior.md` | Construct context through production boundaries and verify externally meaningful outcomes | Pass | API and realtime inputs enter through their production mock boundaries; after the inactive catch-up, reopening the thread reads `[1,2,3,4]` from the normal chat-event projection with no new request. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- Coverage: Adequate — the regression exercises initial unread prewarm, active catch-up, inactive-thread indicator reload, already-unread rechecking, IndexedDB persistence, and API-free reopen.
- Conventions: Compliant — targeted Platform integration coverage is present, and the current PR head has green format, lint, type, app-test, API-test, security, CodeQL, and final Turbo gate checks.

PR #30772, opened 2026-09-01
test(api): stabilize chat sequence lock synchronization

### LINKED ISSUE #30748: test(api): stabilize chat sequence concurrency synchronization
## Background

The merge-queue API shard failed intermittently in GitHub Actions run [33475528807, job 99753942621](https://github.com/vm0-ai/vm0/actions/runs/33475528807/job/99753942621).

The failing integration test was:

`src/signals/routes/__tests__/chat-threads.bdd.test.ts > CHAT-01 chat thread read state > serializes concurrent message sequence writes through commit`

## Problem and evidence

The test waits for `holdChatEventInsertTransactionFixture().blockedWaiterCount()` to equal exactly one before releasing a held transaction. That helper counts every PostgreSQL session in the transitive blocking chain rooted at the holder. Under the API shard's concurrent database workload, additional background or cleanup transactions can temporarily join that chain. The observed count remained three until the poll expired.

The test therefore uses a shared aggregate lock count as a proxy for one specific writer-to-writer dependency. The failure occurred before the existing pre-commit visibility and post-commit sequence-order assertions ran; there is no evidence in this run that product sequencing produced an incorrect committed result.

## Decision

Make the fixture observe the exact PostgreSQL holder/waiter pair:

- Capture and expose the second event-insert transaction's backend PID before it enters the production writer.
- Let the held first transaction query whether that exact PID is directly blocked by its holder PID through `pg_blocking_pids`.
- Poll that pair-specific predicate before checking pre-commit visibility and releasing the first transaction.
- Propagate transaction failures that occur before the PID startup boundary so fixture setup cannot hang.

Retain the existing aggregate waiter-count method for other tests that intentionally observe total blocking activity.

## Alternatives considered

- **Flush background work first:** may reduce same-test activity but cannot exclude other shard sessions and still does not identify the target writer.
- **Accept an aggregate count greater than or equal to one:** avoids the observed numeric mismatch but could pass because of an unrelated waiter before the second writer reaches the sequence boundary.
- **Serialize or isolate the test file:** broadens the change to CI scheduling and hides the fixture-level observation mismatch.
- **Increase timeouts:** leaves the same incorrect predicate and violates the repository's timeout guidance.

## Risks and mitigations

- The second transaction could fail before exposing its PID. Reject the startup deferred from the transaction failure path and bind it to the test abort signal.
- The direct lock relationship could change in the future. Treat that as a meaningful test failure because this test specifically protects direct per-thread sequence serialization.
- Fixture API changes could affect other tests. The second-writer fixture has one current call site; search all references and keep the held fixture's aggregate method for its existing consumers.

## Acceptance criteria

- The concurrency fixture exposes enough transaction identity to observe the intended blocker/waiter pair directly.
- The test no longer requires the holder's total transitive waiter count to equal one.
- The test still exercises the production chat-event sequence writer and retains its pre-commit visibility and post-commit ordering assertions.
- The focused API integration test passes repeatedly without increasing timeouts or skipping coverage.
- The complete affected test file, API lint, API type checks, and applicable Knip analysis pass.

## Scope

This is one test synchronization PR. Product chat-event sequencing behavior, database schema, API contracts, runtime code, and CI scheduling are out of scope unless validation independently shows that the committed ordering guarantee is incorrect.

seven332:
## Summary

- expose the second chat-event insert transaction's PostgreSQL backend PID before it enters the production sequence writer
- wait for that exact backend to be directly blocked by the held writer instead of requiring the holder's complete blocking subtree to contain exactly one session
- preserve the existing pre-commit visibility and post-commit sequence-order assertions, with abort-aware startup failure propagation

## Scope

This changes API integration-test synchronization only. It does not change runtime chat sequencing, API contracts, persisted state, database schema, migrations, deployment behavior, test timeouts, or CI scheduling.

## Validation

- `pnpm -F api exec vitest run src/signals/routes/__tests__/chat-threads.bdd.test.ts -t 'serializes concurrent message sequence writes through commit'` (6/6 focused runs passed)
- `pnpm -F api exec vitest run src/signals/routes/__tests__/chat-threads.bdd.test.ts` (35/35 passed)
- `pnpm exec prettier --check apps/api/src/signals/routes/__tests__/chat-threads.bdd.test.ts apps/api/src/test-fixtures/chat-events.ts`
- `pnpm -F api check-types`
- `pnpm -F api lint`
- `pnpm knip --workspace api`
- pre-commit hook: full `pnpm knip`, full `pnpm check-types` (14/14 tasks), file-size check, and commitlint

Closes #30748

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/chat-threads.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/chat-threads.bdd.test.ts
@@ -37,8 +37,8 @@ import {
 } from "../../../test-fixtures/system-config-seeds";
 import {
   holdChatEventInsertTransactionFixture,
-  insertChatEventTransactionFixture,
   insertOutputEventWithConflictingLegacyPayloadFixture,
+  startChatEventInsertTransactionFixture,
 } from "../../../test-fixtures/chat-events";
 import {
   holdChatThreadEventInsertTransactionFixture,
@@ -2154,16 +2154,21 @@ describe("CHAT-01 chat thread read state", () => {
       content: firstContent,
       signal: context.signal,
     });
-    const secondInsert = insertChatEventTransactionFixture({
+    const secondInsert = await startChatEventInsertTransactionFixture({
       threadId,
       content: secondContent,
+      signal: context.signal,
     });
     onTestFinished(async () => {
       held.release();
-      await Promise.allSettled([held.done, secondInsert]);
+      await Promise.allSettled([held.done, secondInsert.done]);
     });
 
-    await expect.poll(held.blockedWaiterCount).toBe(1);
+    await expect
+      .poll(() => {
+        return held.blocks(secondInsert.pid);
+      })
+      .toBe(true);
     const beforeCommit = await chat.listThreadEvents(owner, threadId);
     expect(
       beforeCommit.events.some((message) => {
@@ -2176,7 +2181,7 @@ describe("CHAT-01 chat thread read state", () => {
 
     held.release();
     await held.done;
-    const second = await secondInsert;
+    const second = await secondInsert.done;
     const committed = await chat.listThreadEvents(owner, threadId);
     const concurrentRows = committed.events.filter((message) => {
       return message.id === held.event.id || message.id === second.id;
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #30772

**Reviewed head:** `f50b9b8d12e67627be251a157bed2e715e6366c6`

### Summary

The PR replaces a flaky aggregate blocking-tree cardinality check with an exact PostgreSQL holder/waiter PID assertion while preserving the real production writer, transaction visibility check, and committed sequence ordering assertions.

### Findings

No findings.

### Testing

The focused test passed six times, the complete affected test file passed 35/35, and Prettier, API lint, API type checking, workspace Knip, and the full pre-commit Knip/type-check hooks passed. The full multi-file API shard matrix was not run locally and remains covered by PR CI.

### Verdict

LGTM

PR #30849, opened 2026-09-01
fix(runner): preserve heartbeat diagnostics and reject sandbox reuse

seven332:
## Summary

- retain bounded, content-safe failed heartbeat cycle and HTTP attempt diagnostics
- preserve heartbeat failures through Claude, Pi, and Codex termination while keeping terminal-result, cancellation, timeout, and resource-failure precedence
- expose bounded heartbeat correlation fields in Runner terminal logs and reject sandbox reuse after structured heartbeat failures

## Scope

- closes #30767
- delivery context: #30682

## Explicit exclusions

- no heartbeat interval, retry, timeout, cgroup, balloon, or resource-policy changes
- no host probe, automatic run retry, database migration, queue contract, or terminal-state reversal
- no URL, header, token, request body, prompt, or user output in diagnostics or logs

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy --all-targets --all-features`
- `cargo doc --workspace --all-features --no-deps`
- `cargo test -p guest-contracts`
- `cargo test -p guest-agent`
- `cargo test -p guest-mock-codex`
- `cargo test -p runner`
- `cargo test -p runner --bin runner cmd::service::signal::tests::signal_service_main_bounded_times_out_systemctl -- --exact --nocapture`
- `cargo test -p vsock-guest --test connection guest_state_restore::guest_state_restore_timeout_kills_and_reaps_process_group -- --exact --nocapture`

`cargo test --workspace --quiet` was attempted twice. The first attempt failed only the unrelated Runner systemctl timing test above; the second passed the full Runner package and failed only the unrelated vsock pid-file timing test above. Both tests passed their targeted reruns, and all directly affected packages passed their full package suites.

Closes #30767
Parent: #30682

PR #30863, opened 2026-09-01
fix(api): keep a committed delete successful when storage cleanup fails

e7h4n:
## Summary

`cli-e2e-03-runner` went red on a merge-queue Turbo run while **every product assertion passed**. The failure was entirely in `teardown_file`:

```
ok 1 vm0 built-in real claude returns a successful answer          # in 5637 ms
ok 2 vm0 built-in real pi loop returns a successful answer         # in 3893 ms
ok 3 vm0 built-in real claude steers an active run then starts a successor # in 24527 ms
not ok 4 teardown_file failed
# Stage 0 Runner E2E agent teardown failed with HTTP 500: {"error":"Internal server error"}
# requestHost:pr-30849-api.vm6.ai requestPath:/api/agents/2feabe94-3206-491c-8614-8b58547a1f7b status:500
```

Triggering run: https://github.com/vm0-ai/vm0/actions/runs/33496657496 (job [99822134815](https://github.com/vm0-ai/vm0/actions/runs/33496657496/job/99822134815), SHA `aac2b0e7401a107fb2cbb615bfcf5d5090a6152a`, queued PR #30849).

## Root cause

The API log for that exact request (`vm0-web-logs-dev`, `_time = 2026-09-01T10:25:24.623Z`, deployment `dpl_8n2H66X8SJQATbQfT5f24ApSMT2N`) shows the server-side error is **not** a lock conflict — it is a transient object-store failure:

```
route: /api/agents/:id      method: DELETE      type: unhandled_request_error
fields.error.Code:    "InternalError"
fields.error.$fault:  "server"
fields.error.$metadata.httpStatusCode: 500
fields.error.$metadata.attempts: 3        # AWS SDK retries already exhausted
fields.error.stack:   S3RestXmlProtocol.handleError ...
```

`deleteAgentById$` releases the deleted agent's instructions storage **after** the write transaction commits:

1. the transaction deletes the `agents` row *and* the `storages` row, then commits — durable and irreversible;
2. only then does it `ListObjectsV2` + `DeleteObjects` against R2;
3. R2 returned a server-fault `InternalError`, which propagated as an unhandled request error → `500`.

Because the `storages` row is already gone, the committed deletion has removed the only pointer to the prefix. A client retry gets `204`/`404` and never re-attempts the cleanup, so the objects are orphaned **whether or not** the request reports failure. The `500` therefore only misreports a committed success; it recovers nothing.

`workflow-delete.service.ts` has the identical shape and the identical failure:

| window | route | count |
|---|---|---|
| 2026-09-01 06:33 UTC (preview) | `DELETE /api/workflows/:workflowId` | 1 |
| 2026-09-01 06:33 UTC (preview) | `DELETE /api/agents/:id` | 3 |
| 2026-09-01 10:18–10:25 UTC (preview) | `DELETE /api/agents/:id` | 5 |
| 2026-08-07/08 (**production**) | `DELETE /api/zero/workflows/:workflowId` | 3 |

Over the last 7 days every single R2 `InternalError` in the API landed on one of these two post-commit cleanup paths, in short bursts across four different deployments — a transient upstream R2 condition, not a code path that fails deterministically.

## Fix

Add `purgeDeletedStoragePrefix$`, a shared command that lists and deletes the objects under a prefix whose owning `storages` row was already deleted by a committed transaction, and reports an upstream failure as a `warn` carrying the bucket and prefix instead of failing the request. `AbortError` still propagates, so a cancelled request is never reported as a success.

Both delete services now route through it. This is the same handling the repository already applies to post-commit R2 cleanup in `agent-run-lifecycle.service.ts:104`.

What deliberately did **not** change:

- **Cleanup is still real.** The purge still runs on every delete and is never a no-op; only its failure is downgraded from "the deletion failed" to "the objects leaked, here is the prefix".
- **No retry, sleep, or timeout change.** The AWS SDK had already spent its 3 attempts; adding more would not have helped a server-fault burst.
- **No change to `delete_runner_agent_for_stage0_teardown`.** Adding a bounded retry for arbitrary `500`s there would have hidden exactly the server defect this run exposed. The helper keeps failing loudly on an unexpected status.

## Validation

New integration test `still reports the agent as deleted when object storage cleanup fails after the delete commits` in `agent-lifecycle.bdd.test.ts` drives the public route with `DeleteObjectsCommand` rejecting as an R2 `InternalError`.

Against the pre-fix service it reproduces the CI failure exactly:

```
Error: Unknown response status 500 for DELETE /api/agents/:id
```

With the fix it asserts `204` and that the agent is subsequently `404`.

- `agent-lifecycle.bdd.test.ts` + `agent-deletion-interlock.bdd.test.ts`: 16 passed
- `workflows.test.ts` + `agent-lifecycle.bdd.test.ts`: 30 passed, repeated 5× consecutively, green every time
- `workflow-automations.test.ts`, `workflow-automation-scheduler.test.ts`, `agents-update.test.ts`, `misc-routes.bdd.test.ts`: 124 passed
- `pnpm -F api check-types`, `eslint` on the changed files, `pnpm knip --workspace apps/api`, `git diff --check`: clean

The end-to-end invariant is covered by this PR's own `cli-e2e-03-runner` shard, which exercises the real teardown against the deployed preview API.

## Follow-up (not in scope here)

Three more sites share the "commit, then release the prefix" shape but have no failure evidence and are not request-response endpoints returning a misleading 5xx: `agent-instructions-storage.service.ts:116`, `org-limited-free-bootstrap.service.ts:304`, and `cron-sync-skills.service.ts:660`. There is also no GC that reclaims R2 objects whose `storages` row is gone, so orphaned prefixes currently rely on the log line this PR adds.

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/agent-lifecycle.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/agent-lifecycle.bdd.test.ts
@@ -1,10 +1,21 @@
+import { DeleteObjectsCommand, ListObjectsV2Command } from "@aws-sdk/client-s3";
 import { describe, expect, it } from "vitest";
 
 import { testContext } from "../../../__tests__/test-context";
+import { nowDate } from "../../../lib/time";
 import { createBddApi, expectApiError } from "./helpers/api-bdd";
 
 const context = testContext();
 const api = createBddApi(context);
+
+/** Shape of the transient upstream failure R2 returns as an S3 `InternalError`. */
+function objectStoreInternalError(): Error {
+  const error = new Error(
+    "We encountered an internal error. Please try again.",
+  );
+  error.name = "InternalError";
+  return error;
+}
 
 describe("AGENT-01: agent lifecycle through public API", () => {
   it("creates, reads, lists, updates, deletes, and verifies removal through API-visible state", async () => {
@@ -75,6 +86,47 @@ describe("AGENT-01: agent lifecycle through public API", () => {
     });
 
     await api.deleteAgent(admin, created.agentId);
+
+    const missing = await api.requestReadAgent(admin, created.agentId, [404]);
+    expectApiError(missing.body);
+    expect(missing.body.error.code).toBe("NOT_FOUND");
+  });
+
+  it("still reports the agent as deleted when object storage cleanup fails after the delete commits", async () => {
+    const admin = api.user();
+    api.acceptAgentStorageWrites();
+
+    const created = await api.createAgent(admin, {
+      displayName: "BDD Storage Cleanup Agent",
+    });
+
+    const listedPrefixes: string[] = [];
+    context.mocks.s3.send.mockImplementation((command: unknown) => {
+      if (command instanceof ListObjectsV2Command) {
+        const prefix = command.input.Prefix ?? "";
+        listedPrefixes.push(prefix);
+        return Promise.resolve({
+          Contents: [
+            {
+              Key: `${prefix}AGENTS.md`,
+              Size: 12,
+              LastModified: nowDate(),
+            },
+          ],
+        });
+      }
+      if (command instanceof DeleteObjectsCommand) {
+        return Promise.reject(objectStoreInternalError());
+      }
+      return Promise.resolve({ ContentLength: 1024 });
+    });
+
+    // The row deletion has already committed by the time the objects are
+    // released, so an upstream object-store failure must not be reported as a
+    // failed deletion. `deleteAgent` accepts only 204.
+    await api.deleteAgent(admin, created.agentId);
+
+    expect(listedPrefixes.length).toBeGreaterThan(0);
 
     const missing = await api.requestReadAgent(admin, created.agentId, [404]);
     expectApiError(missing.body);
--- a/turbo/apps/api/src/signals/services/agent-deletion.service.ts
+++ b/turbo/apps/api/src/signals/services/agent-deletion.service.ts
@@ -7,7 +7,6 @@ import { and, asc, eq, inArray, sql } from "drizzle-orm";
 
 import { db$, writeDb$ } from "../external/db";
 import type { Tx } from "../../lib/db-types";
-import { deleteS3Objects, listS3ObjectsUnderPrefix } from "../external/s3";
 import { env } from "../../lib/env";
 import { conflict } from "../../lib/error";
 import { isLockNotAvailable } from "../../lib/pg-errors";
@@ -16,6 +15,7 @@ import { settle } from "../utils";
 import { lockCanonicalAgentMutation } from "./agent-mutation-lock.service";
 import { removeAgentInstructionsStorageInTransaction } from "./agent-instructions-storage-transaction.service";
 import { reconcileAutomationEventWatches } from "./automation-event-watch-lifecycle.service";
+import { purgeDeletedStoragePrefix$ } from "./storage-prefix-purge.service";
 
 export function agentExistsInOrg(args: {
   readonly orgId: string;
@@ -165,7 +165,7 @@ async function deleteAgentInTransaction(tx: Tx, args: DeleteAgentArgs) {
 }
 
 export const deleteAgentById$ = command(
-  async ({ get, set }, args: DeleteAgentArgs, signal: AbortSignal) => {
+  async ({ set }, args: DeleteAgentArgs, signal: AbortSignal) => {
     const writeDb = set(writeDb$);
 
     const transaction = await settle(
@@ -211,20 +211,14 @@ export const deleteAgentById$ = command(
     signal.throwIfAborted();
 
     if (result.s3Prefix) {
-      const bucket = env("R2_USER_STORAGES_BUCKET_NAME");
-      const objects = await get(
-        listS3ObjectsUnderPrefix(bucket, result.s3Prefix),
+      await set(
+        purgeDeletedStoragePrefix$,
+        {
+          bucket: env("R2_USER_STORAGES_BUCKET_NAME"),
+          s3Prefix: result.s3Prefix,
+        },
+        signal,
       );
-      signal.throwIfAborted();
-      await get(
-        deleteS3Objects(
-          bucket,
-          objects.map((obj) => {
-            return obj.key;
-          }),
-        ),
-      );
-      signal.throwIfAborted();
     }
 
     return undefined;
--- a/turbo/apps/api/src/signals/services/storage-prefix-purge.service.ts
+++ b/turbo/apps/api/src/signals/services/storage-prefix-purge.service.ts
@@ -1,0 +1,62 @@
+import { command } from "ccstate";
+
+import { logger } from "../../lib/log";
+import { deleteS3Objects, listS3ObjectsUnderPrefix } from "../external/s3";
+import { tapError } from "../utils";
+
+const L = logger("StoragePrefixPurge");
+
+interface StoragePrefixPurgeArgs {
+  readonly bucket: string;
+  readonly s3Prefix: string;
+}
+
+const listAndDeletePrefixObjects$ = command(
+  async (
+    { get },
+    args: StoragePrefixPurgeArgs,
+    signal: AbortSignal,
+  ): Promise<void> => {
+    const objects = await get(
+      listS3ObjectsUnderPrefix(args.bucket, args.s3Prefix),
+    );
+    signal.throwIfAborted();
+    await get(
+      deleteS3Objects(
+        args.bucket,
+        objects.map((object) => {
+          return object.key;
+        }),
+      ),
+    );
+    signal.throwIfAborted();
+  },
+);
+
+/**
+ * Release the objects under a storage prefix whose owning `storages` row was
+ * already deleted by a committed transaction.
+ *
+ * That commit is durable and it removed the only pointer to the prefix, so a
+ * caller retry can never reach this cleanup again. Reporting an upstream
+ * object-store failure as a failed deletion would therefore turn a committed
+ * success into a misleading 5xx while leaving the objects orphaned either way.
+ * Log the orphaned prefix instead so it stays auditable and recoverable. Abort
+ * still propagates.
+ */
+export const purgeDeletedStoragePrefix$ = command(
+  async (
+    { set },
+    args: StoragePrefixPurgeArgs,
+    signal: AbortSignal,
+  ): Promise<void> => {
+    await tapError(set(listAndDeletePrefixObjects$, args, signal), (error) => {
+      L.warn("Failed to release objects under a deleted storage prefix", {
+        bucket: args.bucket,
+        s3Prefix: args.s3Prefix,
+        error,
+      });
+    });
+    signal.throwIfAborted();
+  },
+);
--- a/turbo/apps/api/src/signals/services/workflow-delete.service.ts
+++ b/turbo/apps/api/src/signals/services/workflow-delete.service.ts
@@ -9,9 +9,9 @@ import { and, eq, sql } from "drizzle-orm";
 
 import { env } from "../../lib/env";
 import { writeDb$ } from "../external/db";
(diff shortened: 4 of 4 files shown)


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 5227b25921f86528e8c3f69da3ad4b6af1d21278
PR_AUTO_REVIEW_VERDICT: LGTM

> Disclosure: this review is by the PR author under the pr-auto flow. Findings below are stated against `main`'s `REVIEW.md` and the practice documents it routes to, not against the PR's own description.

### Summary

`deleteAgentById$` and `deleteWorkflow$`/`deleteOrphanedWorkflowVolume$` release the deleted `storages` row's R2 prefix *after* the write transaction commits. A transient Cloudflare R2 `InternalError` on that post-commit cleanup propagated as an unhandled request error, turning a committed, irreversible deletion into a `500`. Since the commit already removed the only pointer to the prefix, a caller retry can never re-reach the cleanup, so the `500` recovered nothing and only misreported a committed success. The PR routes all three sites through a new `purgeDeletedStoragePrefix$` that downgrades an upstream failure to a `warn` carrying the bucket and prefix, matching the existing post-commit R2 handling at `agent-run-lifecycle.service.ts:104`. Abort still propagates.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no race conditions or wrong assumptions | Pass | `storage-prefix-purge.service.ts`: cleanup runs only after the transaction commits; `deleteWorkflow$`/`deleteOrphanedWorkflowVolume$` still return `true` only when a row was actually deleted, so `workflows.ts:753` still maps a missing workflow to `workflowNotFound` and a real deletion to `204`. |
| `REVIEW.md` | Security: authz, injection, secrets | N/A | No auth path, query construction, credential, or route surface changes; permission checks in `lockAgentLifecycleForDeletion` are untouched. |
| `REVIEW.md` | Type safety: no `any`, no broad types | Pass | `StoragePrefixPurgeArgs` is a `readonly` interface; no `any`, no `as` bridges across the diff. |
| `REVIEW.md` | Interface changes documented; breaking changes highlighted | Pass | No contract change. `204` was already a declared status for both routes; only which declared status is returned in the R2-failure case changes. |
| `REVIEW.md` | Feature regression guard: changed behavior has coverage for the flow most likely to regress | Pass | Normal-path deletion stays covered by the existing `AGENT-01` lifecycle test and `agent-deletion-interlock.bdd.test.ts` (16 passed); workflow deletion by `workflows.test.ts`, `workflow-automations.test.ts`, `workflow-automation-scheduler.test.ts`, `misc-routes.bdd.test.ts` (124 passed). |
| `REVIEW.md` | DB/JSONB review gate | N/A | No `packages/db/src/schema/**`, contract, Zod, or `jsonb(...).$type<...>()` change; no migration. |
| `docs/bad-smell.md` | Zero `any`; zero lint/type suppressions | Pass | No `any`, no `eslint-disable`/`@ts-ignore`/`@ts-expect-error`/`prettier-ignore` added. |
| `docs/bad-smell.md` | Error handling: no unnecessary try/catch, no over-engineered handling | Pass (with note) | The handled scope is exactly two S3 SDK calls in a post-commit block via the repo's centralized `tapError`, not a route-level wrapper. See note under Findings on its breadth. |
| `docs/bad-smell.md` | No dynamic `import()` | Pass | All imports are static. |
| `docs/bad-smell.md` | No hardcoded URLs/config; use `env()` | Pass | Bucket resolved through `env("R2_USER_STORAGES_BUCKET_NAME")` at each call site; `env` is Zod-validated at load, so a missing bucket name still fails loudly before the purge command runs. |
| `docs/bad-smell.md` | Fail fast; no fallback that hides configuration problems | Pass | The downgraded path is a reachable third-party outage, not a configuration default. Config errors still fail at `env()` validation. |
| `docs/fallback.md` §3-§5 | No fallback for a state the contract, schema, or single writer already prevents | Pass | No `??`/`||` default, no retired-protocol reader, no compatibility branch. The handled state is a real upstream outage observed 9 times in 7 days, including 3 in production. |
| `docs/fallback.md` §6-§8 | A cross-version rollout fallback names surface, gate, removal condition, follow-up | N/A | The diff introduces no cross-version compatibility branch, so no gate or removal condition applies. |
| `docs/fallback.md` §9 | Every fallback introduced by the diff is declared in the PR summary | Pass | The diff introduces none, so per §9 no `Fallbacks` section is required and none is present. See the explicit statement below the Findings section. |
| `docs/fallback.md` §1 | No negative tests asserting removed behavior stays removed | Pass | The new test asserts a live behavior (deletion is reported as committed), not the absence of deleted code. |
| `docs/externally-managed-references.md` | Classify the boundary before handling failure | Pass | R2 object deletion is post-commit cleanup, not a required remote operation for the request's contract — the contract ("the agent is deleted") is already satisfied at commit. |
| `docs/externally-managed-references.md` | Let unrelated failures propagate; do not wrap a whole route and call every exception "unavailable" | Pass (with note) | Only the two S3 calls are wrapped, and structured context (bucket, prefix) is recorded without credentials. Breadth note under Findings. |
| `docs/deployment-compatibility.md` | Cross-version compatibility at deployable boundaries | N/A | No request/response field, enum, endpoint path, runner protocol, queue payload, persisted shape, or migration changes. No old client depends on receiving `500` for a deletion that succeeded. |
| `docs/testing.md` | `fix:` commits must have a regression test | Pass | `agent-lifecycle.bdd.test.ts` adds a regression test that fails on the pre-fix service with `Unknown response status 500 for DELETE /api/agents/:id` and passes with the fix. |
| `docs/testing.md` | Integration tests only; no unit tests for internal functions | Pass | The test drives the public route through the BDD contract client; the new service is never imported directly. |
| `docs/testing/api-testing.md` | Route tests go through `setupApp()` + ts-rest contract, `testContext()` at module scope | Pass | Uses the existing `createBddApi(context)` harness with module-scope `testContext()`. |
| `docs/testing/api-testing.md` | Mock only auth identity and external services via `context.mocks` | Pass | Only `context.mocks.s3.send` is stubbed; no `vi.mock()` of an internal path. |
| `docs/testing/api-testing.md` | Shared persistent state: uniquely owned identities | Pass | `api.user()` yields a fresh user/org per test; the agent is created and deleted inside the test. |
| `docs/testing/testing-external-behavior.md` | Construct and verify through the API surface, not internals | Pass | Setup via `createAgent`, verification via `deleteAgent` (204) and `requestReadAgent` (404). No DB import, row write, or row read. |
| `docs/testing/anti-patterns.md` AP-1 | Do not assert mock calls instead of behavior | Pass (with note) | Behavior assertions are the `204` and the follow-up `404`; `expect(listedPrefixes.length).toBeGreaterThan(0)` is a supplementary non-vacuity guard at the external-service boundary that `api-testing.md` explicitly lists as in-scope coverage. |
| `docs/testing/anti-patterns.md` AP-4 | No `vi.mock()` with a relative path | Pass | None added. |
| `docs/testing/anti-patterns.md` AP-5 | No fake timers | Pass | No `vi.useFakeTimers()`; `nowDate()` from `lib/time` is used for the stubbed `LastModified`, satisfying the repo's `no-restricted-syntax` clock rule. |
| `docs/testing/anti-patterns.md` AP-8 | No over-testing of status codes | Pass | The test covers a business invariant (a committed deletion is reported as deleted), not framework status plumbing. |
| `docs/testing/anti-patterns.md` AP-11 | Test through the route when one exists | Pass | Exercised through `DELETE /api/agents/:id`, not `deleteAgentById$`. |
| `docs/testing/patterns.md` | Standard API route test structure; no ad hoc `vi.clearAllMocks()`, no DB setup/assertions | Pass | No manual mock cleanup, no `initServices()`, no direct DB access added. |
| `.claude/skills/ccstate/SKILL.md` | Extract shared command logic into a sub-command, not a helper receiving `get`/`set` | Pass | `listAndDeletePrefixObjects$` is a `command`; `get` is never passed, aliased, returned, or object-wrapped. |
| `.claude/skills/ccstate/SKILL.md` | Pass `AbortSignal` as the final positional parameter; call `throwIfAborted()` after awaits that do not receive the signal | Pass | `signal` is the last positional argument on both commands and is re-checked after each `get()` on a computed. |
| `.claude/skills/ccstate/SKILL.md` | No `detach()` in the signals layer; no `void`/empty `.catch()` to silence a promise | Pass | The purge promise is created as an argument to an awaited `tapError`, so it is owned; no `detach`, no empty catch. |
| `.claude/skills/ccstate/SKILL.md` | Unused destructured accessors removed | Pass | `get` dropped from `deleteAgentById$`, `deleteWorkflow$`, and `deleteOrphanedWorkflowVolume$` now that only `set` is used. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Notes (non-blocking)

- `storage-prefix-purge.service.ts:purgeDeletedStoragePrefix$` — `tapError` swallows every non-abort rejection from the wrapped block, which includes a programmer error inside the `objects.map(...)` projection, not just an upstream S3 fault. `docs/externally-managed-references.md` ("Error Boundaries and Observability") prefers letting programmer errors propagate. The blast radius is two SDK calls and the behavior matches the accepted precedent at `agent-run-lifecycle.service.ts:104`, so this is not a blocker; worth revisiting only if the wrapped block grows.
- A systematic R2 credential or permission failure now surfaces as a repeated `warn` rather than a `500`. That is the intended trade, and `level=warn` is covered by the daily production error/warn review, but there is still no GC reclaiming objects whose `storages` row is gone — so this log line is currently the only record of an orphaned prefix. The PR's follow-up section records this.
- The workflow-delete wiring to the new command has no failure-case test of its own; it relies on the shared command's coverage from the agent test plus existing success-path workflow tests. Given the repo's "not too many, mostly integration" stance, duplicating the failure case per call site would add little, so this is an observation rather than a request.

### Testing

- Coverage: Adequate. `fix:` commit ships a regression test that provably fails on the pre-fix service (`Unknown response status 500 for DELETE /api/agents/:id`) and passes with the fix.
- Conventions: Compliant. Route-level integration test, external boundary mocked through `context.mocks.s3` only, no internal `vi.mock()`, no fake timers, no DB access, no ad hoc mock cleanup, unique per-test identity.
- Reported runs: `agent-lifecycle` + `agent-deletion-interlock` 16 passed; `workflows` + `agent-lifecycle` 30 passed repeated 5x consecutively; `workflow-automations`, `workflow-automation-scheduler`, `agents-update`, `misc-routes` 124 passed; `pnpm -F api check-types`, `eslint` on changed files, `pnpm knip --workspace apps/api`, and `git diff --check` clean.

**Fallbacks:** the diff introduces none. Per `docs/fallback.md` §9 no `Fallbacks` section is required, and none is present. The new handling covers a currently reachable third-party outage — not a `??` default on a contract-guaranteed field (§3), not a fabricated value for a corrupted row (§4), not a reader for a retired producer (§5), and not a cross-version rollout branch (§6.1) — so it carries no removal gate under §7.

e7h4n:
### Validation evidence: original invariant on the PR

The shard that produced the triggering failure ran clean on this PR.

- Job: [`cli-e2e-03-runner (tests/03-runner/run-t10-real-claude-smoke.bats, 1, 3)`](https://github.com/vm0-ai/vm0/actions/runs/33499014277/job/99829086826) — `SUCCESS`
- bats output:

```
1..3
ok 1 vm0 built-in real claude returns a successful answer          # in 8115 ms
ok 2 vm0 built-in real pi loop returns a successful answer         # in 5037 ms
ok 3 vm0 built-in real claude steers an active run then starts a successor # in 24795 ms
```

No `not ok 4 teardown_file failed` and no `bats warning: Executed 4 instead of expected 3 tests`, so `delete_runner_agent_for_stage0_teardown` completed against the preview API and the file reported exactly its 3 declared tests.

Full PR status: 81 checks `SUCCESS`, 17 `SKIPPED`, 0 failed.

**Scope of this evidence, stated precisely.** R2 was healthy during this run, so this shard proves the normal teardown path is unbroken — it does not by itself exercise the `InternalError` branch. The failure path is proven separately by the new integration test, which returns `Unknown response status 500 for DELETE /api/agents/:id` against the pre-fix service and `204` with the fix.

PR #30864, opened 2026-09-01
fix(runner): fail closed on incomplete config image refs

### LINKED ISSUE #30833: bug(rust): fail closed when retained runner config image refs cannot be read
## Summary

The image protection inventory can be reported as complete even when a retained runner version runner.yaml cannot be read or parsed.

## Evidence

- crates/runner/src/cmd/gc/image_refs.rs:89-103 builds the protection inventory and only uses the enabled-service discovery completion result.
- collect_retained_version_image_refs at crates/runner/src/cmd/gc/image_refs.rs:106-115 calls collect_config_image_refs.
- collect_config_image_refs at crates/runner/src/cmd/gc/image_refs.rs:262-311 returns () and silently returns on config read or YAML parse failures. Invalid profile hashes are also skipped without marking the inventory incomplete.

## Reproduction scenario

1. Retain a version with --protect-version v1.0.0 or --keep-latest.
2. Make runners/v1.0.0/runner.yaml truncated, unreadable, or malformed while an old snapshot referenced by that version remains on disk.
3. run_gc still receives a Complete protection inventory, so image GC runs without that reference and can delete the retained version snapshot.

This violates the fail-closed protection contract and can remove image artifacts needed by a retained runner.

## Acceptance criteria

- Any read, parse, or per-profile hash-validation failure for a retained version or enabled service makes the protection inventory Incomplete.
- run_gc skips image GC when the protection inventory is incomplete.
- Add a regression test through protected_image_refs_for_gc with a malformed or unreadable retained config and assert that the referenced snapshot survives image GC.

seven332:
## Summary

- propagate missing, unreadable, malformed, and hash-invalid config failures through the runner image-reference collectors
- mark the protection inventory incomplete before image GC can consume a partial reference map
- add retained-version and enabled-service regressions that preserve the existing no-config-path service policy

## Scope

- changes only runner image-reference inventory collection and its colocated tests
- does not change public APIs, persisted formats, runner protocols, migrations, or other GC phases
- keeps enabled services with no parseable runner config path as the existing best-effort discovery case

## Validation

- `cargo fmt --all -- --check` — passed
- `cargo check -p runner --all-targets --all-features` — passed with one build job, incremental compilation disabled, and debug info disabled
- `cargo clippy -p runner --all-targets --all-features` — passed with the same low-memory settings
- `cargo doc -p runner --all-features --no-deps` — passed
- `cargo test -p runner cmd::gc::image_refs::tests` — attempted with one build job and debug info disabled; runner test-harness code generation was OOM-killed at approximately 3.74 GiB RSS in this 3.8 GiB cgroup (exit 137), including a second attempt with dynamic linking and GNU gold. CI must execute the linked regression tests.

Closes #30833

### DIFF
--- a/crates/runner/src/cmd/gc/image_refs.rs
+++ b/crates/runner/src/cmd/gc/image_refs.rs
@@ -31,14 +31,6 @@ impl ProtectedImageRefs {
     }
 
     #[cfg(test)]
-    pub(super) fn is_empty(&self) -> bool {
-        match self {
-            Self::Complete(refs) => refs.is_empty(),
-            Self::Incomplete => false,
-        }
-    }
-
-    #[cfg(test)]
     pub(super) const fn incomplete() -> Self {
         Self::Incomplete
     }
@@ -96,7 +88,9 @@ pub(super) async fn protected_image_refs_for_gc(
     }
 
     let mut refs = ProtectedImageRefs::new();
-    collect_retained_version_image_refs(home, version_analysis, &mut refs).await;
+    if !collect_retained_version_image_refs(home, version_analysis, &mut refs).await {
+        return ProtectedImageRefs::Incomplete;
+    }
     if !collect_enabled_service_image_refs(&mut refs).await {
         return ProtectedImageRefs::Incomplete;
     }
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #30864

**Reviewed head:** `466c5c83d4cec3364f2d2063bc602acb8b41b2e6`

### Summary

The PR correctly propagates config-content failures through retained-version and enabled-service image-reference collection, so the existing incomplete-inventory guard skips destructive image GC. The implementation matches the established completeness model and preserves the deliberate no-parseable-service-config-path policy.

### Findings

No findings.

### Testing

Coverage is appropriate for the fix: malformed and missing retained configs, both invalid hash fields, enabled-service config failure, valid reference preservation, and downstream no-deletion behavior are covered with real temporary files and the existing fake-systemd boundary.

Formatting, all-target/all-feature runner type checking, Clippy, and rustdoc passed locally. The linked runner test harness could not be code-generated within the 3.8 GiB local cgroup and was OOM-killed at approximately 3.74 GiB RSS; CI remains responsible for executing the submitted tests.

### Verdict

LGTM

PR #30873, opened 2026-09-01
feat: use transparent avatar cutouts in the intro video picker

bingjiez666:
## What

The intro-video wizard's avatar step now offers a curated set of **80
background-removed presenters** in a masonry grid, and the 9:16 / 16:9 switch
is gone.

| | before | after |
|---|---|---|
| source | full JoggAI catalog, paged | fixed curated set of 80 |
| thumbnail | provider cover photo, baked-in scene | transparent cutout on the card surface |
| layout | fixed frame, `object-cover` | masonry, native aspect ratio |
| aspect switch | 9:16 / 16:9 | removed |

## Feature switch coverage

Everything user-visible sits behind the existing `FeatureSwitchKey.IntroVideo`
switch (`enabled: false`, scoped to the maintainer). `start-cards.tsx` renders
both `IntroVideoStartCard` and `IntroVideoWizard` only when it is on, and the
curated grid is reachable only from the wizard.

Three files are outside that switch. They are listed here so review can weigh
them separately:

- `packages/core/src/intro-video-avatars.ts` — new data module, only imported by
  the gated picker
- `apps/api/.../avatar-video.service.ts` — `caption` now defaults to `false` when
  `screenStyle` is `3` (it still defaults to `true` for 1 and 2). Nothing in vm0
  sent `screenStyle: 3` before this PR
- `apps/cli/.../avatar-video.ts` — `--screen-style` help text only

`avatar-template-picker.tsx` also touches the generic avatar-video template
picker, which is **not** gated, but only as a no-op refactor: the removed
`compact` prop was used by the deleted intro-video toolbar (the generic caller
already used the `compact = false` branch, which is what the className now
hardcodes), and the removed `selectedAvatarId` prop was never passed by the
generic caller, so its ternary always took the `isSelectedAvatarTemplate`
branch. The aspect-ratio switch stays in the generic picker, where the ratio
really does decide the output framing.

## Why the aspect switch goes away

The intro-video take is composited by HyperFrames, which reframes it and decides
the delivered aspect ratio, so asking the user up front was choosing something
that gets overridden later. The ratio the avatar-video request is generated with
is still reported on the review step and in the prompt.

## Why masonry rather than a uniform grid

The cutouts are framed anywhere between head-and-shoulders and full body, so the
on-screen face size is not constant. Measured across all 80 (face detection on
each cutout, face height as a fraction of the column width):

| layout | face-size spread |
|---|---|
| fixed frame, plain `object-contain` | 3.56x |
| masonry, native aspect ratio | 2.54x |
| masonry with heights bucketed into 3 tiers | 2.78x |

Bucketing the heights is *worse* than leaving them native, because letterboxing
into a bucket height rescales the subject against its face size. Masonry is the
best of the layouts that need no per-image retouching.

The multi-column container is deliberately kept separate from the scroll
container and left at auto height: a multicol box with a definite height
overflows in the inline direction, which would turn the grid into a horizontal
scroller.

## Transparent background

JoggAI's `screen_style: 3` returns a WebM with a real alpha channel. JoggAI's
**v1** transparent-video documentation states that captions must be off or the
WebM cannot be generated; the **v2** endpoint this client calls does not repeat
that constraint, and it could not be confirmed against v2.

So the constraint is honoured as a default, not enforced:

- `caption` defaults to `false` when `screenStyle` is `3`
- an explicit `caption` choice is passed through to JoggAI, which owns the rule
- the CLI's `--screen-style` help text said "3 picture in picture", which
  described the provider's own v2 label but not the output; it now names the
  transparent WebM

The generated prompt carries the cutout URL and the transparent-background
instruction so the compositing step can use both.

## Assets

Cutouts are published by vm0-ai/static-files#171 under
`static.vm0.io/platform/avatars/intro-video/v1/<joggAvatarId>.webp`
(80 files, 2.68 MB, average 34 KB). **That PR must merge first**, otherwise the
picker renders 80 broken images. `static.vm0.io` is hard cached for a year and
append-only, so any re-cut of the artwork has to land under a `v2/` prefix.

## Known consequences, calling them out explicitly

1. **The catalog filter popover is gone from the intro-video step.** Style /
   gender / age / scene / ethnicity are JoggAI server-side query params; with a
   fixed local manifest there is nothing to query. The manifest carries only id,
   name and cutout geometry today.
2. **Voice pre-filtering is weaker.** `voiceFiltersForAvatar` reads
   `avatar.gender` / `age` / `style`, which the manifest does not carry, so the
   voice step now opens unfiltered rather than pre-narrowed to the presenter.

Both are fixed by the same follow-up: run one pass against the JoggAI avatar API
with a live key and write `gender` / `age` / `style` into the manifest. No
JoggAI key was available here, so those values were not guessed.

## Test plan

- `avatar-video.test.ts`: added coverage for captions defaulting off under
  `screenStyle: 3`
- `chat-start-cards.test.tsx`: the wizard walkthrough now asserts the curated
  grid renders, that no catalog toolbar is present, and that the prompt carries
  the avatar, the aspect-ratio line, the cutout URL and the transparent-background
  instruction
- Remaining checks (types, lint, knip, full suites) run in the PR pipeline


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 235a40fc1311e5b34826e8072f55d21aa7040afb
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

The intro-video wizard's avatar step is replaced with a fixed 80-presenter set of background-removed cutouts in a masonry grid, and its 9:16 / 16:9 switch is removed; the generic avatar-video template picker keeps its aspect-ratio switch and is only refactored. The API additionally defaults `caption` to `false` when `screenStyle` is `3`, and the CLI's `--screen-style` help text is corrected.

I verified the data module against the published assets: all 80 `coverUrl`s return `200` from `static.vm0.io`, the 80 ids and URLs are unique and mutually consistent (`.../v1/<id>.webp`), and on a 12-entry sample every declared `cutoutWidth`/`cutoutHeight` matches the intrinsic WebP dimensions exactly. The `@okouai/core/intro-video-avatars` subpath resolves through the existing `./*` export and `@okouai/api-contracts` is already a `core` dependency. No dangling references to the removed `AvatarLibraryToolbar`, `setAspectRatio$`, or `aspectRatio$` remain.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no wrong assumptions or unhandled edge cases | Finding (P1) | `avatar-video.service.ts`: the new `screenStyle === 3 → caption false` default is unreachable from the only production caller (see P1 below). Rest of the diff is sound. |
| `REVIEW.md` | Security: no injection, hardcoded secrets, or missing auth on new endpoints | Pass | No new endpoint, no auth change, no credential. The new module is static data; the new `img` renders a fixed `static.vm0.io` URL, not user input. |
| `REVIEW.md` | Type safety: no `any`, no overly broad types | Pass | No `any` in the diff. `IntroVideoAvatar extends AvatarVideoAvatar` legally narrows `coverUrl` from optional to required and adds two required numbers. |
| `REVIEW.md` | Style/maintainability: no >100-line functions, no duplicated logic, comments explain WHY | Pass | `IntroVideoAvatarCard` and `AvatarLibraryContent` are small; the three added comments (multicol height, `v<n>/` immutability, JoggAI caption constraint) all explain WHY. |
| `REVIEW.md` | Feature regression guard: replaced behavior is intentional, scoped, and covered | Pass | The replaced step sits behind `FeatureSwitchKey.IntroVideo` (`enabled: false`). The generic picker path most likely to regress keeps its existing unmodified coverage in `chat-composer-template-picker.test.tsx` (toolbar present at :1000-1003, aspect-ratio paging at :1034-1060). |
| `REVIEW.md` | Interface changes documented; breaking changes highlighted | Pass | Removed `compact` / `selectedAvatarId` / `AvatarLibraryToolbar` have no remaining callers; the PR body explains why both removals are behaviour-neutral for the generic caller. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, migration, or `jsonb(...).$type<...>()` change. `AVATAR_VIDEO_TRANSPARENT_SCREEN_STYLE` is an exported constant, not a contract shape change. |
| `docs/bad-smell.md` | Zero `any`; zero lint/type suppressions | Pass | Grep over the diff for `any`, `eslint-disable`, `oxlint-disable`, `@ts-ignore`, `@ts-nocheck`, `@ts-expect-error`, `prettier-ignore`: no hits. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | No `await import(` / `import(...).then` in the diff. |
| `docs/bad-smell.md` | No hardcoded URLs or environment-specific values | Pass | The 80 `static.vm0.io` asset URLs are published-object references, matching established repo precedent (`views/onboarding/onboarding-data.ts`, `views/okou-page/chat-composer.tsx:369`), not environment configuration. |
| `docs/bad-smell.md` | Fail fast; no error-hiding fallback or recovery logic | Pass | No new try/catch, no configuration fallback. |
| `docs/fallback.md` | Every fallback introduced by the diff is declared in the PR summary | Pass | The diff introduces no new fallback. `screenStyle ?? 1` is pre-existing and only hoisted; `caption ?? …` is a pre-existing default on a contract-optional field whose value changed, not a new code path. No `Fallbacks` section is required. |
| `docs/fallback.md` | No `??`/`||` on a state the owning contract cannot produce | Pass | `caption` and `screenStyle` are both `.optional()` in `avatarVideoGenerateRequestSchema`, so defaulting them is legitimate. |
| `docs/fallback.md` | Features behind a non-GA switch need no compatibility code | Pass | Nothing in the diff adds dual-read, dual-write, or a tolerance branch for the wizard's previous shape. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | `expect(dialog.querySelector("[data-avatar-catalog-toolbar]")).toBeNull()` is not a tombstone: the toolbar still exists and is still rendered by the generic picker, so this asserts a live scoping contract of the wizard step, and it is paired with a positive `[data-intro-video-avatar-grid]` assertion in the same flow. |
| `docs/testing.md` | `feat:` commits must have integration tests | Pass | Both changed behaviours are covered: `avatar-video.test.ts` for the API default, `chat-start-cards.test.tsx` for the wizard step and the generated prompt. |
| `docs/testing.md` | Mock at the boundary only; no `vi.mock()` on `../` internal paths | Pass | The new API test mocks only the external `JOGGAI_CREATE_URL` through MSW; no internal module is mocked. |
| `docs/testing.md` | Use MSW for HTTP; no `vi.stubGlobal("fetch", …)` | Pass | `server.use(http.post(JOGGAI_CREATE_URL, …))`, consistent with the surrounding file. |
| `docs/testing.md` | Real database; no mocking of `globalThis.services.db` | Pass | The new test uses `seedAvatarVideoFixture()` and the real DB path. |
| `docs/testing.md` | No fake timers | Pass | No `vi.useFakeTimers()` / `advanceTimersByTime` in the diff. |
| `docs/testing.md` | Test behavior, not mocks | Pass | The API test asserts the observed outbound request body and a `202`; the platform test asserts rendered UI and prompt text. No bare `toHaveBeenCalled()`. |
| `docs/testing.md` | New user-facing features gated behind a `FeatureSwitchKey` | Pass | `start-cards.tsx` renders the card and wizard only when `FeatureSwitchKey.IntroVideo` is on; the curated grid is reachable only from the wizard. |
| `docs/testing/api-testing.md` | Route tests exercise the real Hono app, not imported handlers or services | Pass | Uses `createAvatarVideoTestApp(...).request("/api/avatar-video/generate", …)`, the pattern already established in this file. |
| `docs/testing/api-testing.md` | Uniquely owned identities for shared persistent state | Pass | `seedAvatarVideoFixture()` provides per-test `userId` / `orgId` / usage-pricing rows; the new test adds no shared-state assumption. |
| `docs/testing/app-testing.md` | Page-level tests entering through `detachedSetupPage`, no component-only unit tests | Pass | The change extends the existing `chat-start-cards.test.tsx` page flow; no new component-level test file was added for the masonry grid. |
| `docs/testing/testing-external-behavior.md` | Construct and verify cases through the external interface | Pass | Platform assertions are page-visible (label, grid presence, sent prompt). The API test posts to the endpoint. See the P1 note on which request shapes a shipped client actually produces. |
| `docs/effect.md` | Render stays pure: no Store writes, I/O, timers, or resource creation in render | Pass | `IntroVideoAvatarCard` and `AvatarLibraryContent` only read a module constant and build elements; selection goes through the existing `setAvatar` / `selectAvatarForVoice$` commands from the click handler. |
| `docs/effect.md` | Choose the execution boundary; no effect used as a generic state reaction | Pass | The diff removes state (`aspectRatio$`) and its cross-signal sync in `start-cards.tsx` rather than adding any effect, ref, or command. |
| `docs/cache.md` | RC-2: no signal allocation during render | Pass | No `state`/`computed`/`command` is created in the new components. |
| `docs/cache.md` | RC-3: no unbounded-lifetime domain-scoped cache | Pass | `INTRO_VIDEO_AVATARS` is a fixed 80-element `readonly` literal, not a keyed cache that grows with session identity. |
| `docs/cache.md` | RC-5: no duplicate mutable sources of truth | Pass | Removing `aspectRatio$` and the `setAvatarFilters` sync in `openIntroVideo` deletes one such duplication rather than adding one. |
| `docs/cache.md` | RC-7/RC-8: stable refs and symmetric resource teardown | Pass | No refs, listeners, observers, timers, or object URLs are added. |
| `.claude/skills/ccstate/SKILL.md` | DOM callbacks use `detach(..., Reason.DomCallback)`; no `void`, no empty `.catch()` | Pass | The new `onClick` invokes synchronous setters; the existing `detach(openIntroVideoWizard(pageSignal), Reason.DomCallback)` in `start-cards.tsx` is preserved. |
| `.claude/skills/ccstate/SKILL.md` | All `apiClient$` calls wrapped in `accept`; no manual status handling in signals | Pass | The wizard's avatar step no longer makes an API call at all; no `apiClient$` call is added or changed. |
| `docs/deployment-compatibility.md` | Old frontend → new backend stays safe | Pass | The API change only alters a default for `screenStyle: 3`. The previously deployed frontend never sends `screenStyle: 3`, and the CLI always sends an explicit `caption`, so no already-loaded client's behavior changes. |
| `docs/deployment-compatibility.md` | New frontend → old backend stays safe | Pass | The new wizard makes no new API request; it only changes generated prompt text. Against an old API, `screenStyle: 3` is already an accepted contract value. |
| `docs/deployment-compatibility.md` | Commit-addressed CLI artifacts: no request/response variant removed | Pass | The CLI change is a help string only; the request shape it sends is unchanged. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

- `turbo/apps/api/src/signals/services/avatar-video.service.ts:234-247` — the new `caption` default is unreachable from every shipped client, so the safety net it is meant to provide will not fire in the intro-video flow.

  `parseAvatarVideoOptions` only applies `captionDefault` when the request omits `caption`. The sole production caller of `POST /api/avatar-video/generate` is `generateWebAvatarVideo` in `apps/cli/src/lib/api/domains/web.ts:963`, reached from `apps/cli/src/commands/generate/avatar-video.ts:392-400`, which sends `caption: options.caption` unconditionally. That option is declared as `.option("--no-caption", …)`, and Commander's boolean-negated options default to `true`. I confirmed this empirically: parsing `--screen-style 3` yields `{"screenStyle":3,"caption":true}`, and only `--screen-style 3 --no-caption` yields `caption:false`.

  So `okou generate avatar-video --screen-style 3` posts `{screenStyle: 3, caption: true}`, `parsed.data.caption` is defined, and `caption: true` is forwarded to JoggAI — the exact combination the new comment says does not produce the alpha-channel WebM. The intro-video take would then come back with baked-in captions instead of transparency, and HyperFrames would composite a non-transparent presenter.

  The new test passes because it posts a body with `caption` omitted, which is legal at the endpoint but is a shape no shipped client produces, so it does not demonstrate the behavior the feature will actually get.

  Two things reduce the severity, which is why this is P1 rather than P0: the generated prompt already tells the agent `transparent WebM (JoggAI screen_style 3, which requires captions off)`, so an agent that follows it will pass `--no-caption` itself; and the PR body notes the v2 endpoint does not repeat the v1 caption constraint, so it may not bind in practice. Suggested fix: have the CLI omit `caption` unless the flag was actually supplied (`command.getOptionValueSource("caption") === "cli"`), or derive its default from `--screen-style` the same way the API now does, and cover it with a CLI test.

### Testing

- Coverage: Adequate. The `feat:` change is covered on both surfaces it alters — `avatar-video.test.ts` asserts the outbound `screen_style: 3, caption: false` body, and `chat-start-cards.test.tsx` walks the wizard to the curated grid, selects `Amara`, and asserts the four new prompt lines. The pre-existing `screen_style: 1, caption: true` assertion (`avatar-video.test.ts:643-644`) still pins the unchanged default, so the regression path for styles 1 and 2 is covered. The one gap is the CLI-side default described in the P1: no test covers what `--screen-style 3` actually sends.
- Conventions: Compliant. Boundary-only MSW mocking of the JoggAI URL, real DB via `seedAvatarVideoFixture()`, no internal `vi.mock()`, no fake timers, no `toHaveBeenCalled()`-only assertions, and the platform test stays on page-visible surfaces through the existing `detachedSetupPage` flow.

PR #30904, opened 2026-09-01
fix(runner): fail closed on unavailable mountinfo

### LINKED ISSUE #30878: bug(runner): fail closed when workspace mountinfo is unreadable
## Background

`crates/runner/scripts/mount-workspace-drive.sh` checks `/proc/self/mountinfo` before mounting `/dev/vdb` so it can reject a workspace device that is already mounted somewhere else.

This was found while self-reviewing #30860, which adds behavior coverage for the workspace mount and freeze helpers. The behavior predates that pull request.

## Problem

`workspace_device_mounted_elsewhere()` reads mountinfo inside a function that is invoked as an `if` condition:

```sh
if workspace_device_mounted_elsewhere; then
  # reject the mount
fi
```

POSIX shell `set -e` does not terminate the shell for a failed redirection in this conditional function context. If mountinfo cannot be opened, the function continues to its final `return 1`, and the helper proceeds to `mkdir` and `mount` even though it could not verify that the device is unused elsewhere.

Minimal reproduction of the shell control-flow behavior:

```sh
sh -c 'set -eu; probe() { while IFS=" " read -r _; do :; done < /definitely-missing-mountinfo; return 1; }; if probe; then echo matched; fi; echo continued'
```

Observed result: the shell prints the open error and `continued`, then exits with status 0.

No production occurrence is currently known. The canonical mountinfo path is normally readable, but failure of this safety input must not be interpreted as "device is not mounted elsewhere."

## Expected behavior

The mount helper must fail closed before creating the workspace directory or invoking `mount` when it cannot open or consume the configured mountinfo source.

## Acceptance criteria

- A mountinfo open/read failure makes the helper exit nonzero with a clear diagnostic.
- The failure path does not invoke `mkdir`, `mount`, or `chown`.
- Add a Linux behavior test that executes the checked-in mount helper with an unavailable mountinfo fixture and asserts the exit status, diagnostic, and absence of later side effects.
- Preserve the existing matching mount, mismatched mount, mounted-elsewhere, symlink, and successful new-mount behavior.

## Relevant files

- `crates/runner/scripts/mount-workspace-drive.sh`
- `crates/runner/src/workspace_mount.rs`

Related to #30860.

seven332:
## Summary

- open the configured mountinfo source on the scan function invocation so access failures cannot be overwritten by a conditional function return
- separate successful scan completion from the mounted-elsewhere result, avoiding shell-specific numeric redirection status assumptions
- fail with a clear diagnostic before workspace filesystem side effects and cover the unavailable-source path through the checked-in helper

## Scope

- preserves the matching existing mount, mismatched mount, mounted-elsewhere, unresolved device identity, symlink, and successful new-mount behavior
- guarantees fail-closed behavior when the configured mountinfo source cannot be opened
- does not claim detection of a post-open I/O error that POSIX `read` cannot distinguish from normal EOF
- introduces no API, database, persisted-state, or cross-version protocol change

## Validation

- `sh -n runner/scripts/mount-workspace-drive.sh runner/scripts/freeze-workspace-drive.sh`
- `cargo test -p runner workspace_mount::tests::behavior -- --nocapture` (11 passed)
- `cargo test -p runner --all-targets --all-features` (3388 passed, 24 ignored; guest inventory 1 passed)
- `git diff --check`
- pre-commit workspace Rust format, Clippy, documentation, and file-size checks

Closes #30878

### DIFF
--- a/crates/runner/scripts/mount-workspace-drive.sh
+++ b/crates/runner/scripts/mount-workspace-drive.sh
@@ -18,14 +18,14 @@ refuse_workspace_symlink_path() {
   done
 }
 
-workspace_device_mounted_elsewhere() {
-  [ -n "$workspace_dev" ] || return 1
+inspect_workspace_mountinfo() {
   while IFS=' ' read -r _ _ mount_dev _ _ _; do
     if [ "$mount_dev" = "$workspace_dev" ]; then
+      workspace_device_mounted_elsewhere=true
       return 0
     fi
-  done < "$workspace_mountinfo_path"
-  return 1
+  done <&3
+  return 0
 }
 
 ensure_workspace_owner() {
@@ -44,7 +44,15 @@ if mountpoint -q -- "$workspace_dir"; then
   exit 64
 fi
 
-if workspace_device_mounted_elsewhere; then
+workspace_device_mounted_elsewhere=false
+if [ -n "$workspace_dev" ]; then
+  if ! inspect_workspace_mountinfo 3< "$workspace_mountinfo_path"; then
+    echo "unable to inspect workspace mounts because mountinfo is unavailable: $workspace_mountinfo_path" >&2
+    exit 66
+  fi
+fi
+
+if [ "$workspace_device_mounted_elsewhere" = true ]; then
   echo "refusing to mount workspace drive because $workspace_device is already mounted outside $workspace_dir" >&2
   exit 64
 fi
--- a/crates/runner/src/workspace_mount.rs
+++ b/crates/runner/src/workspace_mount.rs
@@ -551,6 +551,36 @@ log_call() {{
         }
 
         #[test]
+        fn mount_script_rejects_unavailable_mountinfo() {
+            let fixture = WorkspaceScriptFixture::new();
+            fixture.write_mountpoint(false, Some(WORKSPACE_DEV), None);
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #30904

**Reviewed head:** `fa0591f34e567a851f4e92c0084826f62e9d1273`

### Summary

The helper now opens the configured mountinfo source on the scan invocation, separates successful scan completion from match state, and fails before filesystem side effects when the source cannot be opened. The design avoids both the original conditional-function status overwrite and shell-specific numeric error assumptions without adding an external reader or preflight race.

### Findings

No findings.

### Testing

- Shell syntax passed for the mount and freeze helpers.
- The Linux helper behavior suite passed 11 tests, including the new missing-source failure and all existing match, no-match, symlink, and freeze cases.
- Runner-scoped format, Clippy, rustdoc, and all-target/all-feature tests passed (3388 passed, 24 ignored; guest inventory 1 passed).
- Pre-commit workspace Rust checks and diff checks passed.
- Matching, empty, and missing-source control flow plus descriptor closure were additionally checked under Dash and Bash.
- A real Firecracker guest mount was not run locally; the changed failure boundary is executed through the checked-in helper with controlled process and filesystem boundaries.

### Verdict

LGTM

PR #30980, opened 2026-09-02
test(python): assert usage buffer destination routing

### LINKED ISSUE #30956: test(python): assert usage-buffer destination routing across batches
## Problem and evidence

`test_flush_keeps_runs_categories_providers_and_destinations_separate` buffers three batches for two distinct API URLs, bearer credentials, and proxy log paths in `crates/runner/mitm-addon/tests/test_usage_buffer_aggregation.py`. Its assertions currently inspect only payload run IDs and event fields.

The recording helper already preserves `url`, `sandbox_token`, `proxy_log_path`, and `log_type` for every enqueue call in `crates/runner/mitm-addon/tests/usage_buffer_helpers.py`. Production batch construction maps those destination fields independently from the payload in `crates/runner/mitm-addon/src/usage/buffer/state.py`.

As a result, a regression that swaps or collapses destination URLs, credentials, or proxy log paths can leave every current assertion green while routing usage to the wrong endpoint or attributing its delivery logs to the wrong sandbox.

## Decision

Proceed with a test-only change to make the existing multi-destination behavior test compare each complete enqueue destination with its associated payload. Production behavior and shared test helpers remain unchanged.

## Chosen direction

- Project every `enqueue.calls` entry into one order-independent comparison member containing URL, sandbox token, proxy log path, `log_type`, payload `runId`, and the payload events' kind, provider, category, and quantity.
- Compare that set with three explicit expected members: the two `api-a` batches using `token-a` and `proxy-a.jsonl`, and the `api-b` batch using `token-b` and `proxy-b.jsonl`.
- Assert the recorded call count separately so exact duplicate calls cannot be hidden by set normalization.
- Exclude only generated aggregate idempotency-key values from the projection because they contain a random per-buffer identifier; existing tests retain responsibility for key validity and idempotency behavior.

## Alternatives considered

- An ordered list of complete calls is more literal but unnecessarily makes current batch ordering part of this routing contract.
- Grouping calls into a destination-to-payload dictionary is order-independent but adds more local accumulation machinery for a three-call fixture.
- A shared normalization or assertion helper has no demonstrated second consumer and would add an abstraction contrary to YAGNI.

## Risks and mitigations

- Set equality can collapse identical normalized calls; the explicit three-call assertion preserves multiplicity.
- Fixed generated idempotency keys would make the test unstable; only that nondeterministic field is omitted.
- The nested comparison shape can become dense; keep its destination and payload fields visibly aligned and rely on Ruff formatting.
- This change does not cover ordering because ordering is not the reported contract.

## Acceptance criteria

- Assert the complete deterministic destination-to-payload mapping for every enqueue call in the multi-destination aggregation test.
- Verify that both `api-a` batches use `token-a`, `proxy-a.jsonl`, and `usage_event`, while the `api-b` batch uses `token-b`, `proxy-b.jsonl`, and `usage_event`.
- Include run ID plus event kind, provider, category, and quantity in the same comparison member as its destination.
- Require exactly three recorded calls.
- Ensure the assertion fails if destination fields are swapped or collapsed while payload grouping remains unchanged.

## Delivery notes

- Expected changed file: `crates/runner/mitm-addon/tests/test_usage_buffer_aggregation.py` only.
- Run the targeted test and the complete committed Python add-on validation set: lock check, locked environment sync, Ruff format/lint, BasedPyright, and all pytest tests.
- No API, protocol, persistence, migration, dependency, rollout, or deployment-compatibility change is involved.

seven332:
## Summary

- assert every usage-buffer enqueue call as one complete destination-to-payload association
- cover URL, sandbox token, proxy log path, log type, run ID, event identity, and quantity
- keep routing assertions order-independent while checking exact call multiplicity

## Scope

- test-only change in `test_usage_buffer_aggregation.py`
- no production behavior, shared helper, dependency, protocol, persistence, or deployment change

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (7,325 passed)
- `git diff --check`

Closes #30956

### DIFF
--- a/crates/runner/mitm-addon/tests/test_usage_buffer_aggregation.py
+++ b/crates/runner/mitm-addon/tests/test_usage_buffer_aggregation.py
@@ -637,25 +637,56 @@ def test_flush_keeps_runs_categories_providers_and_destinations_separate(tmp_pat
     )
 
     assert usage.flush_usage_events(trigger="test") == 3
-
-    payloads = enqueue.payloads
-    assert {(payload["runId"], len(payload["events"])) for payload in payloads} == {
-        ("run-1", 2),
-        ("run-2", 1),
-        ("run-1", 1),
-    }
-    all_events = [flushed_event for payload in payloads for flushed_event in payload["events"]]
-    assert {
+    assert enqueue.call_count == 3
+    routed_batches = {
         (
-            flushed_event["kind"],
-            flushed_event["provider"],
-            flushed_event["category"],
+            call.url,
+            call.sandbox_token,
+            call.proxy_log_path,
+            call.log_type,
+            call.payload["runId"],
+            frozenset(
+                (
+                    flushed_event["kind"],
+                    flushed_event["provider"],
+                    flushed_event["category"],
+                    flushed_event["quantity"],
+                )
+                for flushed_event in call.payload["events"]
+            ),
         )
-        for flushed_event in all_events
-    } == {
-        ("model", "claude-sonnet-4-6", "tokens.input"),
-        ("model", "claude-sonnet-4-6", "tokens.output"),
-        ("connector", "x", "posts.read"),
+        for call in enqueue.calls
+    }
+    assert routed_batches == {
+        (
+            "https://api-a.test/api/webhooks/agent/usage-event",
+            "token-a",
+            proxy_a_log_path,
+            "usage_event",
+            "run-1",
+            frozenset(
+                {
+                    ("model", "claude-sonnet-4-6", "tokens.input", 10),
+                    ("model", "claude-sonnet-4-6", "tokens.output", 5),
+                }
+            ),
+        ),
+        (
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #30980

**Reviewed head:** `ace273440d08e15a19f18376bbc992f4d061041d`

### Summary

The test now compares each usage-buffer payload together with its URL, sandbox token, proxy log path, log type, run ID, event identity, and quantity. It remains order-independent, checks exact call multiplicity, and does not change production code or shared helpers.

### Findings

No findings.

### Testing

The targeted regression test passed. The complete Python add-on validation passed locally: locked uv sync/check, Ruff format and lint, BasedPyright, and all 7,325 pytest tests. `git diff --check` also passed. Required GitHub CI checks were still running at review time and remain the only unverified area.

### Verdict

LGTM

PR #30982, opened 2026-09-02
test(python): cover malformed api destination cache reuse

### LINKED ISSUE #30958: test(python): make malformed API destination cache test observe reuse
## Problem and evidence

`test_malformed_vm0_api_url_is_cached_as_non_match` sends two requests with the same malformed platform API URL in `crates/runner/mitm-addon/tests/test_request_handler_api_admission.py`, but it currently asserts only request behavior and the absence of an upstream binding.

The cache contract lives in `crates/runner/mitm-addon/src/upstream_admission.py`: `_api_destination` must reuse the prior result for the same raw URL, including a cached `None` from malformed input. Removing the lookup or failing to retain negative results would not change the current assertions, so the named cache behavior is not protected.

This matters because the platform API URL is consulted from multiple request-admission paths. Without negative-result reuse, every request would repeat parsing and normalization for the same malformed configuration. The existing `test_api_destination_derivation_reuses_and_refreshes_effective_option` establishes a repository precedent for observing parse work while still driving real addon hooks.

## Decision

Strengthen the existing request-hook integration test so that it observes bounded parsing for malformed URLs without turning the test into a private-function unit test.

The test must continue to drive `mitm_addon.request` with real flows and real destination derivation. It may wrap a standard-library parsing boundary while delegating to the real implementation, but it must not call `_api_destination` directly, replace `_derive_api_destination`, or stub an internal result merely to assert a call count.

## Alternatives considered

- Keep only the existing behavioral assertions: rejected because they do not detect loss of negative-result caching.
- Directly spy on or call the private derivation helper: rejected because it couples the test to an internal implementation boundary and conflicts with the repository's integration-testing rules.
- Add production-only diagnostics for this cache: rejected as unnecessary surface area for a deterministic test-only observation.

## Risks and mitigations

- A call-count assertion can become implementation-coupled. Limit observation to the standard-library parse boundary and retain all externally visible flow assertions.
- Parametrized malformed inputs fail at different parsing stages. Count the common parse entry point so every fixture case exercises the same cache contract.
- A single-entry cache must refresh when the raw option changes. Cover a changed URL in the same hook-driven lifecycle so the test does not accidentally permit a permanently sticky negative entry.

## Acceptance criteria

- Drive multiple request-hook executions with the same malformed platform API URL and assert that URL parsing happens once.
- Explicitly preserve negative-result caching: the second request must reuse the cached `None` while retaining the current request behavior and absence of an upstream binding.
- Change the effective platform API URL and assert that the changed raw value is parsed rather than reusing the previous negative entry.
- Keep the test at the addon hook entry point; do not directly invoke or replace private destination-derivation helpers.
- The focused pytest file and the mitm-addon formatting, lint, and type checks pass.

seven332:
## Summary

- observe real URL parsing while driving the malformed platform API URL request-hook test
- verify repeated malformed values reuse a cached negative destination result
- verify changing the effective API URL triggers a fresh derivation without changing request behavior

## Scope

- test-only change; no production code, API, persistence, dependency, or deployment behavior changes

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_request_handler_api_admission.py::test_malformed_vm0_api_url_is_cached_as_non_match` (3 passed)
- `uv run --no-sync python -m pytest tests/` (7,325 passed)

Closes #30958

### DIFF
--- a/crates/runner/mitm-addon/tests/test_request_handler_api_admission.py
+++ b/crates/runner/mitm-addon/tests/test_request_handler_api_admission.py
@@ -1,6 +1,7 @@
 """Platform API admission request hook integration tests."""
 
 import json
+import urllib.parse
 
 import pytest
 from mitmproxy import connection
@@ -10,6 +11,7 @@ import mitm_addon
 import platform_api
 import registry
 import request_classification
+import upstream_admission
 import upstream_destination_binding
 from tests.request_handler_helpers import _single_firewall_sandbox, _write_registry
 from tests.upstream_connection_helpers import (
@@ -370,6 +372,22 @@ async def test_malformed_vm0_api_url_is_cached_as_non_match(
     monkeypatch,
     malformed_platform_api_url,
 ):
+    parsed_api_urls: list[str] = []
+    parse_api_url = urllib.parse.urlparse
+
+    def track_api_url_parse(
+        api_url: str,
+        scheme: str = "",
+        allow_fragments: bool = True,
+    ) -> urllib.parse.ParseResult:
+        parsed_api_urls.append(api_url)
+        return parse_api_url(
+            api_url,
+            scheme=scheme,
+            allow_fragments=allow_fragments,
+        )
+
+    monkeypatch.setattr(upstream_admission.urllib.parse, "urlparse", track_api_url_parse)
     flows = [
         real_flow(with_response=False, host="api.vm0.ai"),
         real_flow(with_response=False, host="api.vm0.ai"),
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #30982

**Reviewed head:** `5bc0691b755a7279750c0c954458691824fb6b76`

### Summary

This test-only change keeps the real request hook and destination derivation active while observing the standard-library parse boundary. It now detects repeated derivation of the same malformed URL and failure to refresh after the effective raw URL changes.

### Findings

No findings.

### Testing

Validated with uv lock and environment checks, Ruff format/lint, BasedPyright, the focused three-case regression test, and the full mitm-addon suite (7,325 passed). Turbo and Rust checks were not run because their inputs and consumers are unchanged. GitHub CI remains independently tracked.

### Verdict

LGTM

PR #30983, opened 2026-09-02
perf(nbd-cow): send read replies with vectored writes

### LINKED ISSUE #30963: perf(rust): send NBD read replies with vectored writes
## Problem

`crates/nbd-cow/src/server.rs:332-339` sends every successful NBD Read response in two independent write loops: it first writes the fixed 16-byte reply header to completion, then writes the payload to completion. `write_all_or_shutdown` at `crates/nbd-cow/src/server.rs:241-267` issues ordinary `write` operations, so the normal writable-socket path performs a separate syscall/readiness cycle for the header and payload.

This split is unnecessary for successful reads because the dispatcher is the sole writer for its socket and the header and payload are already available as separate borrowed buffers. A vectored write can submit both without copying.

The cost is on the steady-state block-I/O hot path. `crates/nbd-cow/src/lib.rs:47-54` configures 4 KiB blocks and three connections per device, while the built-in benchmark at `crates/nbd-cow/src/bin/bench/main.rs:97-116` includes 4 KiB random-read and mixed-read workloads. Each completed non-empty read therefore pays one avoidable outbound write operation; at N read IOPS this can add roughly N extra write syscalls per second.

## Suggested direction

Add a cancellation-aware vectored-write path for successful Read replies:

- Submit the serialized reply header and read payload as `IoSlice` values through `write_vectored`.
- Correctly advance across partial writes, including partial progress within the header or payload.
- Preserve the current biased shutdown behavior, `WriteZero` handling, error propagation, and wire order.
- Keep header-only replies for Write, Flush, Trim, and errors on the existing single-buffer path.
- Avoid concatenating the header and payload into a new allocation.

## Acceptance criteria

- The common successful non-empty Read response can complete with one vectored write submission instead of two independent write loops.
- Partial vectored writes are handled correctly across both slices, including a write that stops inside the header or inside the payload.
- Zero-length reads, cancellation, `WriteZero`, socket errors, and shutdown-triggered COW synchronization preserve current behavior.
- Focused tests verify exact header-plus-payload framing across consecutive small and large reads and exercise partial-write advancement deterministically.
- The existing `nbd-cow` dispatch tests pass, and the built-in `rand4k-read` and mixed workload benchmark is used to check before/after IOPS and latency with no write-path regression.
- No NBD protocol, COW format, connection-count, buffering, or persistence semantics change.

## Audit provenance

Found during the daily Rust performance audit of 16 randomly selected files at `2584031276636c504f534f93c330c01e2cb8ac61`.

seven332:
## Summary

- send successful non-empty NBD Read headers and payloads through one cancellation-aware vectored-write loop
- preserve the existing single-buffer path for zero-length Reads and all header-only replies
- cover header, slice-boundary, and payload partial progress deterministically through the full dispatch path and a real Unix socketpair

## Scope

- no NBD wire-format, COW-format, persistence, connection-count, buffering, or public API changes
- no new dependencies or payload concatenation/allocation on the production Read response path
- biased shutdown selection, `WriteZero`, and socket error propagation remain explicit on every pending vectored submission

## Validation

- `cargo clippy -p nbd-cow --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p nbd-cow --all-features --no-deps`
- `cargo test -p nbd-cow --all-features`
- repository pre-commit hook, including workspace-wide Rust formatting, documentation, Clippy, and file-size checks

The built-in benchmark target compiles and its 41 tests pass through the all-features validation. Representative `rand4k-read` and `mixed-70r30w` measurements were not run because this container is not root and lacks `fio` and `dmsetup`; this PR makes no unmeasured IOPS or latency claim.

Closes #30963

### DIFF
--- a/crates/nbd-cow/src/server.rs
+++ b/crates/nbd-cow/src/server.rs
@@ -4,6 +4,9 @@
 //! decodes NBD requests, applies them through [`crate::cow_io::CowIo`], and writes
 //! NBD replies back to the kernel.
 
+use std::io::IoSlice;
+#[cfg(test)]
+use std::num::NonZeroUsize;
 use std::os::unix::io::{FromRawFd, IntoRawFd, OwnedFd};
 
 use tokio::io::{AsyncReadExt, AsyncWriteExt};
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #30983

**Reviewed head:** `8ea8129e1885d5ae2efb95757b6ca808d59779e7`

### Summary

The change keeps successful non-empty Read replies allocation-free while replacing two ordinary write loops with one cancellation-aware vectored cursor. Zero-length and all header-only replies retain the existing path, and the test-only write limit exercises partial advancement through the full dispatch flow over real files and a Unix socketpair.

### Findings

No findings.

### Testing

Scoped formatting, Clippy, rustdoc, all-feature `nbd-cow` tests, diff checks, and repository pre-commit hooks passed. The all-feature run includes 41 benchmark-target tests. Representative host-level `rand4k-read` and `mixed-70r30w` measurements remain unverified because this container is non-root and lacks `fio` and `dmsetup`.

### Verdict

LGTM

PR #30999, opened 2026-09-02
fix(artifacts): skip poster extraction for webm artifacts

bingjiez666:
## Problem

Artifact previews extract a video poster through Cloudflare Media Transformations, whose [documented input requirement](https://developers.cloudflare.com/stream/transform-videos/) is *"an MP4 with H.264 encoded video and AAC or MP3 encoded audio. Other formats may work but are untested."*

WebM is outside that requirement, so every WebM artifact fails with:

```
media frame extraction failed (400): MEDIA_TRANSFORMATION_ERROR 9412:
Could not transform media: Unable to determine media duration
```

That message misdescribes the cause. The files are intact — one production WebM artifact carries a valid EBML `Duration` element (`0x4489`, float64 `37000.0`) against a 1 ms `TimecodeScale`, i.e. a perfectly readable 37 s duration, and still fails.

## Production evidence

Over 14 days of production logs, **all 16 WebM artifacts that reached this path failed, all with the identical error** — a 100% failure rate. Each one:

- pays for a request that cannot succeed, and
- emits a `Failed to render artifact preview` warning that no one can act on.

Re-encoding upstream is not an option for the containers that matter here: the WebM artifacts that users actually see are alpha-channel renders, and MP4/H.264 cannot represent an alpha channel.

## Change

Recognize the unsupported container before the request and return the existing `false` no-op — the same path already taken when the browser-rendering token is unset.

The poster stays absent, exactly as it is today. Nothing else changes: the client already falls back to painting the first frame from the video itself (`<video src="…#t=0.001" preload="metadata">`).

This deliberately does **not** attempt to make WebM posters work; it only stops a guaranteed-failing round trip and the warning it produces.

## Test plan

- Added: a `video/webm` upload issues no media-frame request and leaves the catalog thumbnail null.
- Existing `video/mp4` poster tests are untouched and still assert exactly one frame request.

Local suites were not run; relying on the PR pipeline.


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: be0f954c83dab01e8eb19ed03f1ff66281b747e8
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Adds a `canExtractVideoPoster()` precondition inside `renderAndStoreArtifactPreview$` so a `video/webm` artifact returns the existing `false` no-op instead of issuing a Cloudflare Media Transformations request that production data shows failing 16/16 times over 14 days. User-visible behavior is unchanged — the poster was already always absent for WebM, and the client still paints the first frame from `<video src="…#t=0.001">`. The diff is two files, `+39 / -5`.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic: no race conditions, bad conditionals, or unhandled null paths | Pass | `artifact-preview.service.ts:213` returns before any I/O, so no partial allocate/put/DB-update state is reachable. `contentType?.startsWith("video/webm") ?? false` handles `null` and parameterized types like `video/webm;codecs=vp9`. |
| `REVIEW.md` | Security: no injection, hardcoded secrets, or missing auth on new surfaces | N/A | No new endpoint, no auth path, no credential or user input handling; the guard reads an already-persisted content type. |
| `REVIEW.md` | Type safety: no `any`, no overly broad types | Pass | `canExtractVideoPoster(contentType: string | null): boolean` mirrors the adjacent `isVideoContentType` signature; no casts added. |
| `REVIEW.md` | `fix:` commits must carry a regression test | Pass | `artifacts.bdd.test.ts:344` asserts a `video/webm` upload issues zero media-frame requests and leaves `thumbnail` null. |
| `REVIEW.md` | Feature regression guard: changed/replaced behavior needs explicit coverage of the path most likely to regress | Pass | The `video/mp4` success test is untouched and still asserts exactly one frame request with the exact transform URL. The extraction-failure test was retargeted `unsupported-video.webm` → `unsupported-video.mp4` so the non-2xx branch stays covered by a container that still reaches the transformer. |
| `REVIEW.md` | New user-facing features gated behind a `FeatureSwitchKey` | N/A | Not a new feature. The observable result (null thumbnail for WebM) is identical before and after, so there is no rollout surface to gate. |
| `REVIEW.md` | DB/JSONB review gate: persisted contract changes need migration, normalizer, or rationale | N/A | No `schema/**`, `api-contracts/**`, Zod, or `jsonb().$type<>()` change. `runUploadedFiles.previewImageUrl` simply stays null for WebM, as it already did. |
| `docs/bad-smell.md` | §1 zero `any`; §2 zero lint/type suppressions | Pass | Neither appears in the diff. |
| `docs/bad-smell.md` | §3 fail fast; no over-engineered or unnecessary error handling | Pass | No new `try`/`catch`. The guard removes a throw-and-warn round trip rather than adding one; the genuine failure path (`extractVideoPoster` non-2xx) still throws and is still logged by `scheduleArtifactPreviewRender$`. |
| `docs/bad-smell.md` | §5 no dynamic `import()`; §6 no hardcoded URLs or env-specific values | Pass | No imports added; the transform base still comes from `publicArtifactsBaseUrlForBrand(publicBrand)`. |
| `docs/fallback.md` | §3/§5 no fallback for a state the contract or a retired producer cannot produce | Pass | Not a fallback. WebM uploads are a normal, currently-reachable, supported state — not a contract-prevented or retired one. This is a capability precondition on a documented third-party limitation, structurally identical to the existing `if (!token) return false;` no-op on the sibling branch, and closest to §6.2 "genuinely optional external data". |
| `docs/fallback.md` | §8/§9 a new cross-version rollout fallback needs surface, gate, removal condition, follow-up, and a `Fallbacks` section in the PR summary | N/A | The diff introduces no cross-version compatibility branch — nothing here bridges an old and new deployed version of the DB/API, runner, or client. No `Fallbacks` section is required, and per §9 none should be added. |
| `docs/fallback.md` | §1 no negative tests asserting removed behavior stays removed | Pass | The new test asserts live product behavior for a container users upload today, not the absence of deleted code. |
| `docs/testing.md` | Integration tests only; mock at the system boundary; no `vi.mock()` on internal paths | Pass | The new case drives the real Hono app through the BDD run/upload helpers and reads back via `listArtifactCatalog`. No `vi.mock("../…")` added. |
| `docs/testing.md` | Use MSW for HTTP; no `vi.stubGlobal("fetch", …)` | Pass | Reuses `mockCloudflareVideoFrame()`, an MSW `http.get` handler on the real `/cdn-cgi/media/…` transform URL. |
| `docs/testing.md` | Real database; no mocking `globalThis.services.db` | Pass | Assertion goes through the catalog API; no DB import, write, or read in the test. |
| `docs/testing.md` | No fake timers | Pass | Uses `flushWaitUntilForTest()` as an explicit synchronization point, consistent with the sibling video tests. |
| `docs/testing/api-testing.md` | Construct and verify state through the route contract; unique per-test owned identities under shared persistent state | Pass | `artifactActor("Artifacts API webm preview agent")` allocates its own user/org/agent/object-store, and `mockCloudflareVideoFrame(owner.actor.userId)` returns 415 for foreign artifacts, so the recorded request list cannot be polluted by concurrent tests. |
| `docs/testing/testing-external-behavior.md` | The API endpoint is the only trusted boundary; do not assert on services or DB rows | Pass | Asserts the outbound HTTP the system did or did not make, plus `previewedArtifact?.thumbnail` from the catalog endpoint. `renderAndStoreArtifactPreview$` is never imported or called directly. |
| `docs/testing/anti-patterns.md` | AP-1 test behavior, not mock calls; AP-7 no implementation-detail assertions; AP-11 no service tests when a route exists | Pass | `expect(frameRequests).toHaveLength(0)` asserts a real network effect recorded by MSW, not `expect(mock).toHaveBeenCalled()`, and it is paired with the user-observable `thumbnail` assertion. |
| `.claude/skills/ccstate/SKILL.md` | Correct `command`/signal usage in `apps/api` signals; no floating promises or manual signal plumbing | Pass | The early return sits inside the existing `command(({ get, set }, args, signal) => …)` body and short-circuits before `signal.throwIfAborted()`, where no async work has started. The `waitUntil(tapError(…))` wrapper in `scheduleArtifactPreviewRender$` is unchanged. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

**Non-blocking observation.** The guard is a WebM-specific denylist, while the Cloudflare requirement quoted in the PR body is narrower — MP4/H.264 with AAC or MP3, with other formats "untested" rather than known-broken. An `mp4`-only allowlist would match the documentation more literally, but it would also stop poster extraction for containers such as `video/quicktime` that may transform successfully today, turning a warning fix into a silent feature regression. Scoping to the one container with 100% measured failure is the lower-risk choice and matches the evidence in the PR; no change requested. If other containers later show the same failure signature in logs, extending this predicate is the natural place.

### Testing

- **Coverage: Adequate.** This is a `fix:` commit and it ships the matching regression test — a `video/webm` upload now provably issues zero media-frame requests and still surfaces a null thumbnail. The complementary paths are all still exercised: `video/mp4` success asserts exactly one request at the exact transform URL, the write-once concurrent-poster path is untouched, and the extraction-failure path was migrated to `video/mp4` so the non-2xx branch remains covered rather than silently deleted.
- **Conventions: Compliant.** Integration-level, MSW at the network boundary, real database, no internal `vi.mock()`, no fake timers, no `expect(mock).toHaveBeenCalled()` as a sole assertion, per-test owned actor identity, and the standard `flushWaitUntilForTest()` synchronization used by the surrounding suite. Test placement matches `turbo/apps/api/src/signals/routes/__tests__/`.

PR #31008, opened 2026-09-02
feat: open the intro video wizard from a desktop recording handoff

bingjiez666:
## Problem

Finishing a desktop screen recording opened a vm0 tab that landed on a chat thread with two attachment chips — the recording and its click track — and a prefilled sentence. The Create an intro video wizard those files belong to never opened, so the guided avatar and voice steps were only reachable by starting over from the source picker.

## Change

`applyDesktopRecordingHandoff$` now adopts the handoff into the wizard and opens it at `source-review`, the same step an in-browser recording lands on. The next step is the avatar picker.

The desktop uploads both files before handing the browser a link, so the wizard's source model now distinguishes two origins:

- **local** — bytes this browser holds. Persisted to the local draft store, uploaded at submit, offered back as a download when a send fails. Unchanged behavior.
- **uploaded** — already stored server-side. Skips the local draft store and the submit-time upload entirely, and previews from the address `restoreAttachments$` already signed for the restored attachment, so adopting it costs no extra request and no byte round trip.

The click track keeps riding along as a composer attachment, so the agent receives the same pair of files it does today. Replacing an adopted source takes its uploads back off the draft, so a stale recording is never sent beside the new source.

## Notes

- Gated by the existing `introVideo` and `desktopScreenRecording` switches, both scoped to the maintainer, so this is not user-visible yet.
- A cross-account handoff still fails closed: the file API answers 404, the attachments are pruned with the existing composer banner, and the wizard stays shut.
- Adds a `sourceReview.inAccount` string across all ten locales, since neither "In this browser" nor "Until this tab closes" describes an uploaded source.

## Testing

- `pnpm -F @okouai/app lint` — clean
- `pnpm -F @okouai/app exec tsc -p tsconfig.production.json --noEmit` and `-p tsconfig.tests.json` — clean
- `pnpm knip --workspace apps/platform` — clean
- `vitest run` over `prompt-param-injection`, `chat-start-cards`, `chat-draft`, `attachment-chips`, `chat-composer-*`, `chat-lifecycle`, `home-route`, and `src/signals/__tests__` — all pass

New coverage in `prompt-param-injection.test.tsx`:

- the handoff opens the wizard on the source review step with the recording named, labelled as stored in the account, previewing from the resolved URL, and one Next away from Choose an avatar — while still resolving each file exactly once
- walking that wizard to Create in chat sends the recording and click track as the uploads the desktop already made, with no second copy of either
- an unavailable (cross-account) handoff leaves the wizard shut


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: d9648037fc192a215ab5846ff34663fc290c85e6
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Lands a delivered desktop screen recording in the Create an intro video wizard on the `source-review` step — the same step an in-browser recording lands on — instead of leaving it as two attachment chips in the chat composer. To do that without a byte round trip, `IntroVideoSource` becomes a discriminated union on `origin`: a `local` source keeps today's behavior (IndexedDB draft, submit-time upload, download-on-failure), while an `uploaded` source skips all three and previews from the URL `restoreAttachments$` already signed for the restored attachment.

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

#### Notes (non-blocking, P2)

- `intro-video.ts` `uploadSourceIfNeeded$`: the `source.origin === "uploaded" || get(internal.sourceUploaded$)` disjunction reads as redundant, since `adoptUploadedRecording$` already sets `sourceUploaded$`. It is load-bearing twice over and should not be simplified away: it supplies the narrowing `sourceFile(source)` needs, and it keeps the retry path correct after a failed send, where `submit$` resets `sourceUploaded$` to `false` and only the `origin` check prevents a re-upload of bytes the browser never had. Worth a one-line comment on a future touch.
- `desktop-recording-handoff.ts` `restoredRecordingPreviewUrl$`: the `?? null` after `find` normalizes a lookup miss that the surrounding contract already prevents — the command only runs when `restoreAttachments$` reported nothing unavailable, so the recording is present and resolved. `previewUrl: string | null` is a legitimate field value (documents have no preview) rather than a fabricated default, so this is type narrowing rather than a fallback, but the miss branch is unreachable in practice.

### Testing

- Coverage: Adequate. `feat:` commit with three new integration cases in `prompt-param-injection.test.tsx`, all driven through the real page setup and MSW rather than internal functions:
  - the handoff opens the wizard on the source review step with the recording named, labelled `In your account`, previewing from the resolved URL, and one `Next` away from `Choose an avatar` — while still asserting `file-url` is called exactly twice, which is what proves the adoption reuses the cached `fileInfo$` resolution instead of signing a second time;
  - walking that wizard through to `Create in chat` sends exactly the two uploads the desktop already made (`video-upload-id`, `events-upload-id`), proving no second copy of either file;
  - the cross-account handoff leaves the wizard shut.
- Regression guard: the pre-existing composer behavior this PR reroutes is still asserted in the same file — attachments, agent instructions, prefilled prompt, stripped URL params, and the unavailable-attachment banner path are all unchanged and still covered.
- Conventions: Compliant. No `vi.mock` of internal paths, no `fetch` stubbing (MSW via `context.mocks.http` / `context.mocks.api`), no fake timers, no assertion-on-mock-calls. `queryAllByRoleFast` is used instead of `*ByRole("button", …)` per the repo lint rule.
- Containment: the whole handoff path stays gated behind the non-GA `introVideo` and `desktopScreenRecording` switches, both scoped to the maintainer, so there are no external users of the changed flow.
- No fallback or cross-version compatibility path is introduced, so no `Fallbacks` section applies.

PR #31037, opened 2026-09-02
fix(platform): let clerk own shared worker authentication

e7h4n:
## Summary

- reuse one SharedWorker per Clerk user and organization through the canonical `?userId=...&orgId=...` URL and `okou_USER_ID_ORG_ID` name, with only a one-way tab registration message
- bootstrap one ccstate Store and one headless Clerk instance inside the worker so API and Ably authentication use Clerk directly
- forward the captured Vercel preview bypass and, on development Clerk instances, the page's dev browser JWT through the worker URL, so the worker can reach the preview API and authenticate against a Frontend API that has no worker-readable cookie
- replay the last realtime status to a tab that registers after the worker already connected
- remove credential-specific Stores, heartbeat and token synchronization, authentication recovery, and the custom token refresh path

## Testing

- `pnpm --filter @okouai/app lint`
- `pnpm --filter @okouai/app check-types`
- `pnpm --filter @okouai/app exec vitest run src/shared-database/ src/lib/__tests__/ src/signals/__tests__/shared-database-browser.test.ts src/views/okou-page/__tests__/sidebar-account-menu.test.tsx`
- `pnpm --filter @okouai/app build`
- Preview QA on `pr-31037-app.omby.ai`: two tabs share one SharedWorker, Clerk authenticates inside the worker, Ably tokens are issued, and a message sent in one tab streams into the other


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 8a9c639e4cbaca9ced8c5d1b91e73f98cc1231fe
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Moves Clerk authentication into the shared database SharedWorker so one worker per Clerk `(user, org)` owns its own API and Ably credentials, and removes the tab-side machinery that existed only to feed or poke it: the credential-specific stores, heartbeat/token synchronization, the `vm0-api-origin` meta marker, the visibility-driven realtime pause/close, the `ApiClientRuntime.environment` branch, and the app-to-worker `reload-computed` capability. Preview-only inputs the worker cannot read for itself (Vercel bypass, Clerk dev browser JWT, diagnostics opt-in) are passed on the worker URL. Net effect on the diff is roughly -3.4k lines.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness: races, edge cases, API misuse | Pass | `worker-context.ts` replays the last realtime status to a tab registering after the worker already connected, closing the join race the worker-owned Ably connection created; covered by `message-port.test.ts`. |
| `REVIEW.md` | Type safety: no `any`, no over-broad types | Pass | Diff adds no `any`/`as any`; `clerk.__internal_onBeforeRequest` is typed by `@clerk/clerk-js` `FapiRequestInit`. |
| `REVIEW.md` | Testing coverage for a `fix:` commit | Pass | Regression coverage added for each fixed behavior (see Testing). |
| `REVIEW.md` | Feature regression guard for replaced behavior | Pass | Replaced flows keep coverage: `sidebar.test.tsx` mark-read/mark-unread now drive the server's `chatThreadReadCursorUpdated` event instead of the deleted tab-side reload; `platform-runtime-environment.test.ts` keeps the app-Worker preview host case under hostname-only derivation. |
| `docs/bad-smell.md` | Zero `any`, zero lint/type suppressions | Pass | No `@ts-ignore`/`@ts-expect-error`/`eslint-disable`/`oxlint-disable` added by the diff. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | No `await import(`/`import(` added. |
| `docs/bad-smell.md` | No hardcoded URLs / environment values | Pass | Added origins live in `turbo/apps/app-worker/src/worker.js`, a standalone Cloudflare Worker that already declares `PRODUCTION_API_ORIGINS` and `PREVIEW_API_ORIGIN_PATTERN` as top-level constants; the new `PREVIEW_APP_HOSTNAME_PATTERNS` follows that established shape. Platform code resolves origins through `derivePlatformServiceOrigin`. |
| `docs/bad-smell.md` | Fail fast, no silent fallback for configuration | Pass | `api-base.ts` drops the `<meta name="vm0-api-origin">` override entirely; `resolveApiBaseForTarget` is now hostname derivation only, and the edge worker throws `Shared-thread API origin is unavailable` for an unrecognized host. |
| `docs/fallback.md` | No new undeclared fallback | Pass | The diff introduces no cross-version compatibility branch. `attachDevBrowser` is gated on a Clerk instance capability (dev browser JWT present), not on a deployed version, so no `Fallbacks` section is required. |
| `docs/fallback.md` | Removing a fallback deletes branch, contract entry, and its tests together | Pass | The marker path is removed from `api-base.ts`, `index.html`, the app Worker, `prepare-okou-*.sh`, `verify-okou-production-domains.sh`, `release-please.yml`, and every test that stubbed or asserted it; repo-wide `vm0-api-origin` grep is empty. Same for `reload-computed` (bridge interface, three clients, protocol union, server route, worker handler, test stub). |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | Deleted tests were removed with their code (five visibility tests, two marker tests); no tombstone assertions added. |
| `docs/event-sourcing.md` | Persistent events are the only normal trigger for reconciling optimistic state; refresh is the recovery path | Pass | Mark-read/mark-unread no longer imperatively re-read after their own mutation; the server publishes `chatThreadReadCursorUpdated` after the committed write and the worker recomputes and broadcasts. Optimistic read marks are still cleared only by the persistent signal in `applyChatThreadReadCursorUpdated$`. |
| `docs/effect.md` | Derived value uses `computed`; semantic command owns a business transition | Pass | Indicators are split into `chatThreadIndicators$` (worker-side API read) and `chat-thread-indicators-from-worker.ts` (tab-side bridge read), mirroring the existing `queueData$` / `queueDataFromWorker$` pair, with no runtime environment branch. |
| `docs/effect.md` | Every async process has an `AbortSignal` with an eventual abort path; listeners released by the same owner | Pass | `setupConnectionDiagnostics$` and `setupBrowserLifecycleDiagnostics$` both take the caller's signal and pass `{ signal }` to `addEventListener`; the removed `visibilitychange` listener took its close-timer machinery with it. |
| `docs/cache.md` | RC-3: no unbounded module-scope mutable state | Pass | The diagnostics bus dispatches on `globalThis`, which is an `EventTarget` in both the page and the Worker, so no module-level `EventTarget` is introduced (the `ccstate/no-package-variable` rule is satisfied without a package variable). |
| `docs/cache.md` | RC-8: symmetric resource lifecycles | Pass | Worker-side Clerk, Ably, and diagnostics subscriptions are all bound to the worker root signal; the tab-side `SharedWorker` error listener is registered with `{ signal }`. |
| `.claude/skills/ccstate/SKILL.md` | No `get`/`set` after `await` inside a computed; reactive async computed over imperative fetch | Pass | `clerk$` in `worker-auth.ts` is a synchronous computed returning a promise; `bootstrapWorker$` reads all URL params before any await. |
| `.claude/skills/ccstate/SKILL.md` | Detach/floating-promise and signal-ownership discipline | Pass | The deferred-promise handshake added in an earlier iteration was removed; nothing awaits an orphaned deferred now. |
| `docs/deployment-compatibility.md` | Old/new client skew across an independently deployed boundary | Pass | The app shell and the Worker that rewrites it ship from the same artifact: `prepare-okou-app-worker-shell.sh` builds the shell from the canonical dist, and `prepare-okou-pages-dist.sh` copies `turbo/apps/app-worker/src/worker.js` in as `_worker.js`. Marker producer and consumer therefore deploy and roll back together, and an already-loaded old tab keeps the DOM it was served. |
| `docs/deployment-compatibility.md` | Persisted/protocol shape changes stay compatible | Pass | `register-tab` stays a one-way message with no new required field; `reload-computed` is removed only from the client-to-worker union and is still carried worker-to-tab. The app and its SharedWorker are one deployable unit. |
| `docs/testing.md` | Integration tests at entry points, not internal functions | Pass | New coverage sits at the boundaries: worker URL construction, MessagePort protocol, sidebar UI flows, runtime environment resolution, Worker HTTP behavior. `clerk-dev-browser.test.ts` is the documented parser exception (cookie precedence between suffixed and unsuffixed names). |
| `docs/testing/anti-patterns.md` | AP-1/AP-2/AP-4/AP-5: no mock-call assertions, fetch stubbing, internal mocks, or fake timers | Pass | No `vi.mock("../..")`, no `vi.stubGlobal("fetch")`, no `vi.useFakeTimers()` added. `vi.stubGlobal("SharedWorker", ...)` is the pre-existing boundary stub in that suite. |
| `docs/testing/app-testing.md` | Platform tests exercise pages/handlers, no external network | Pass | `sidebar.test.tsx` drives real UI actions and lets the mocked API publish the realtime signal, matching production ordering (server publishes before responding). |
| `docs/testing/testing-external-behavior.md` | Assert user-visible behavior over internals | Pass | Indicator tests assert the rendered Unread affordance disappears, not that a reload command ran. |

### Findings

#### Critical (P0)
- None.

#### High Priority (P1)
- None.

- `turbo/apps/platform/src/signals/shared-database-browser.ts`: the Clerk dev browser JWT is part of the SharedWorker URL but not its `name`, so if that JWT rotates while an older tab still holds the worker, a newly opened tab hits `URLMismatchError`. `SingleConnectionSharedDatabaseBridge.prepareTransport` catches it (`settle` → `requestReload`), so the tab reloads once and then surfaces the `sharedDatabaseUnavailable` toast rather than looping. The parameter only exists on development Clerk instances (preview and local), production is unaffected, and the diagnostics flag deliberately goes into the `name` as well so a mismatched capture preference yields a separate worker instead of a throw. Worth a line in the PR body for future readers.
- `turbo/apps/platform/src/signals/__tests__/shared-database-browser.test.ts`: the new diagnostics test seeds `FEATURE_SWITCH_CACHE_KEY` in `localStorage` directly because this suite uses the low-level bridge harness rather than `setupPage`. It matches what `page-helper.ts` does, but a shared helper would be less white-box if more suites need it.

### Testing
- Coverage: Adequate. Worker URL handoff for the Vercel bypass, Clerk dev browser JWT, and diagnostics flag (`shared-database-browser.test.ts`); missed-status replay on late tab registration (`message-port.test.ts`); dev browser cookie precedence and Frontend API query construction (`clerk-dev-browser.test.ts`); API origin derivation including the app-Worker preview host (`platform-runtime-environment.test.ts`); realtime-driven mark-read/mark-unread (`sidebar.test.tsx`); Worker request handling and shell preparation (`okou-app-worker-test.mjs`, `prepare-okou-app-worker-shell-test.sh`, `prepare-okou-pages-dist-test.sh`, `deploy-okou-pages-workflow-test.sh`). Behavior deleted in this PR had its tests deleted with it rather than retargeted.
- Conventions: Compliant. No internal-path `vi.mock`, no direct fetch mocking, no fake timers, no mock-call-only assertions; tests assert rendered state and transport-visible messages.
- Verification beyond CI: the dev-instance Frontend API failure was reproduced and the fix confirmed on the PR preview (`pr-31037-app.omby.ai`) — two tabs sharing one SharedWorker, no `Unable to authenticate this browser for your development instance`, Ably token issued, and a message sent in one tab streaming into the other.

PR #31046, opened 2026-09-02
feat(chat): persist structured image annotations

e7h4n:
## Summary

- persist image edits on one logical user-message file part with the immutable original `fileId`, rendered `annotatedFileId`, and structured `annotations`
- scope the annotation editor to its composer and render/upload the derivative on confirmation, with loading, retry, cancellation, and send blocking
- preserve structured annotations through drafts, recall, transfer, and clipboard flows, and project the rendered image plus the structure to the agent

## Compatibility

- keep `draftAttachments` transport-only
- intentionally do not read the previous annotation fields
- require no JSONB migration, snapshot migration, or IndexedDB version change

## Verification

- `pnpm -F @okouai/app exec vitest run src/views/okou-page/__tests__/image-annotation.test.tsx src/signals/__tests__/user-message-document-codec.test.ts` (26 tests)
- `pnpm -F @okouai/api-contracts exec vitest run src/contracts/__tests__/chat-threads.test.ts` (27 tests)
- platform production and test TypeScript checks
- API contracts plus API gateways, core, and test TypeScript checks
- targeted ESLint and Prettier checks

The API route BDD coverage is included and typechecked; its local runtime requires the repository test PostgreSQL service, so CI will execute that layer.


e7h4n:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 0bd687de34fd08609926657cff41468791b7bded

### Summary

Collapses the two-file annotated send into one logical user-message file part that carries the immutable original `fileId`, the rendered `annotatedFileId`, and structured `annotations`, renders and uploads the derivative on confirmation with pending/retry/cancel states, and moves the annotation editor from a root singleton to a per-composer signal factory. The contract, prompt projection, draft persistence, recall, transfer, and clipboard paths are all updated consistently, and the send is cross-field validated so `annotatedFileId` and `annotations` can only travel together.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic: no unhandled edge cases or wrong assumptions | Finding (P1) | `views/okou-page/attachment-chips.tsx:1841` — `uploading` switched from the `fileInfo$` loadable to `uploadPending$`, which never leaves `pending` after a failed original upload |
| `REVIEW.md` | Security: no injection, hardcoded secrets, or missing auth on new surfaces | Pass | No new endpoint, credential, or auth path; the send stays on the existing authenticated `POST /api/chat/events` |
| `REVIEW.md` | Type safety: no `any`, no broad types | Pass | No `any` added; the one `as unknown as CanvasRenderingContext2D` is a test-only canvas stub |
| `REVIEW.md` | Interface changes documented, breaking changes highlighted | Pass | `contracts/chat-threads.ts:570` documents both new part fields; PR body has a Compatibility section |
| `REVIEW.md` | Feature regression guard: changed existing behavior needs regression coverage | Finding (P1) | The failed-upload chip state is existing non-switch-gated UI; no test covers it before or after |
| `REVIEW.md` | DB/JSONB gate: persisted shape changes need migration, normalizer, or rationale | Pass | `persistedAttachmentSchema` is a non-strict `z.object`, so old rows with `annotation` are stripped on read and old writers are not rejected; no JSONB migration needed |
| `docs/bad-smell.md` | No duplicated logic that should be shared | Finding (P1) | `apps/api/.../chat-user-message.service.ts:238` duplicates `apps/platform/.../flatten-annotated-image.ts:454` byte for byte across two deployables |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions, no dynamic imports, no hardcoded config | Pass | No new suppression or dynamic import; no hardcoded URL added |
| `docs/bad-smell.md` | Fail fast instead of hiding a bad state | Pass | `resolve-draft-attachments.ts:186` throws an invariant error when marks exist without a rendered derivative, matching the existing `attachmentUploadFailureMessage` throw above it |
| `docs/fallback.md` | The diff introduces no fallback for a state the contract already prevents | Pass | `part.annotatedFileId ?? part.fileId` and `target.annotations ?? emptyAnnotation()` branch on genuinely optional states the current writer produces, not on retired shapes |
| `docs/fallback.md` | Removing the previous fields needs evidence, not a compatibility reader | Pass | `composerImageAnnotation` is `enabled: false` with a single `enabledEmailHashes` entry, so §2 applies; the PR states it intentionally does not read the previous fields |
| `docs/fallback.md` | Every introduced fallback is declared in the PR summary | N/A | The diff introduces no new fallback, so no `Fallbacks` section is required |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | The old `failImageDecodes` / `Marks on …` assertions were deleted with the behavior instead of being kept as tombstones |
| `docs/deployment-compatibility.md` | Frontend → backend changes are additive for the rollout window | Pass | `annotatedFileId` and `annotations` are optional additions to the `.strict()` file part; absent for every non-switch client |
| `docs/deployment-compatibility.md` | New web/app → old API and old web/app → new API both stay safe | Pass | Only a staff email behind a non-GA switch can emit the new fields; an old bundle rendering a new annotated message falls back to the original image, which is cosmetic for that one account |
| `docs/deployment-compatibility.md` | Persisted-state changes cover rows written by the previous version | Pass | Draft attachments become transport-only in `draft-persistence.ts:34`; annotation state now rides the persisted `userMessage`, and old `annotation` keys are dropped by the non-strict schema |
| `docs/event-sourcing.md` | Optimistic events reconcile by shared event ID and are not rolled back on failure | N/A | The diff changes user-message part content, not optimistic event creation, IDs, or reconciliation |
| `docs/effect.md` | Render stays pure — no Store writes, resource creation, or signal allocation | Pass | `createAttachmentAnnotationSignals` runs inside the attachment factories, `createImageAnnotationSignals` inside `createComposerSignals`; the views only `useGet`/`useSet` |
| `docs/effect.md` | `onRef` only when the command depends on the mounted element | Finding (P1) | `signals/okou-page/composer-signals.ts:527` ignores `_element` and uses the ref purely as a mount/unmount trigger for teardown |
| `docs/effect.md` | One semantic command owns the transition; every async command has an owning `AbortSignal` | Pass | `commitAnnotation$` → `confirmAnnotations$` is one transition; `resetUploadSignal$` is combined with `pageSignal` at `chat-draft.ts:369` |
| `docs/cache.md` | RC-3/RC-4: no unbounded caches, state scoped to its real owner | Pass | The editor moved from module singletons in `views/main.tsx` to a per-composer factory, which narrows the owner rather than widening it |
| `docs/cache.md` | RC-7/RC-8: stable refs, symmetric acquire/release | Pass | `useSet(signals.setImageAnnotationLifecycleRef$)` is passed directly as a ref; the abort listener closes the editor and cancels pending annotation uploads |
| `.claude/skills/ccstate/SKILL.md` | `detach(..., Reason.DomCallback)` only in views, never in signals | Pass | New `detach` calls are in `image-annotation-editor.tsx` and `attachment-chips.tsx`; `confirmAnnotations$` awaits its own work |
| `.claude/skills/ccstate/SKILL.md` | Signal factory pattern with a stable `key` for React | Pass | `chat-draft.ts:520` adds `key: crypto.randomUUID()` and `AttachmentChips` now keys on it instead of `String(a.fileInfo$)` |
| `docs/testing.md` | `feat:` commits carry integration tests | Pass | `chat-events.bdd.test.ts` covers the route projection end to end; `image-annotation.test.tsx` covers the page flows |
| `docs/testing/api-testing.md` | Route tests go through `setupApp()` and the contract, not services or the DB | Pass | The new BDD case sends through `requestSendEvent` and verifies via `api.readRun` plus `waitForThreadMessages` |
| `docs/testing/testing-external-behavior.md` | Cases are built and verified from the external surface | Pass | The platform tests drive the real composer, lightbox, and editor and assert on the outgoing request body |
| `docs/testing/app-testing.md` | Platform tests are page-level through `detachedSetupPage`, not component-only units | Pass | All new platform cases enter through the existing `setup(true)` page helper; the codec assertions extend a pre-existing signals codec suite |
| `docs/testing/patterns.md` | Mock at the boundary (MSW / `context.mocks`), real internals | Pass | `context.mocks.api`, `context.mocks.upload`, and canvas/`Image` browser stubs only; no `vi.mock()` of internal paths |
| `docs/testing/anti-patterns.md` | AP-5/AP-8/AP-9: no fake timers, no over-testing schema, mock cleanup exists | Pass | No fake timers; the contract cases exercise the custom cross-field `.refine`, not Zod itself; `vitest.config.ts` sets `clearMocks`, `restoreMocks`, `unstubGlobals` |

### Findings

None.

- `turbo/apps/platform/src/views/okou-page/attachment-chips.tsx:1841` — **the upload spinner never stops after a failed attachment upload, on the switch-off path.** `uploading` changed from `infoLoadable.state === "loading"` to `useGet(attachment.uploadPending$)`. `AttachmentUploadState` has only `pending` and `uploaded` (`chat-draft.ts:50`), and `upload$` (`chat-draft.ts:517`) leaves the state at `pending` when `uploadFileToStorage$` throws, so `uploadPending$` stays `true` forever. Before this PR the rejected `fileInfo$` moved the loadable to `hasError`, the spinner stopped, and the corner button's label switched from "Cancel upload …" to "Remove …". Now a user whose image upload fails sees an animation that never ends and a button that still claims to cancel an upload. This is existing, non-switch-gated UI, and no test covers it in either direction — REVIEW.md's Feature Regression Guard asks for explicit regression coverage on a changed existing path. Please combine the annotation-pending signal with the original upload's real terminal state (either gate on the loadable not being `hasError`, or add a `failed` variant to `AttachmentUploadState`) and add a page test that fails an original upload and asserts the chip settles.

- `turbo/apps/api/src/signals/services/chat-user-message.service.ts:238` — **`annotatedImageFilenameSnapshot` duplicates the platform's `annotatedFilename`** (`turbo/apps/platform/src/signals/okou-page/flatten-annotated-image.ts:454`) byte for byte, in a separately deployed package, with nothing tying the two together. `docs/bad-smell.md` flags duplicated logic that should be shared. The drift is silent rather than loud: the API only passes the derived name as `filenameHint`, and `resolveArtifactObject$` (`services/artifact-storage.service.ts:427`) falls back to a prefix listing when the exact V2 key misses, so a divergence costs an extra S3 `LIST` per annotated attachment on every send and prints a filename in the agent prompt that does not match the stored artifact. Both sides already depend on `@okouai/api-contracts`, where `imageAnnotationSchema` lives; please move the derivation there and import it from both.

- `turbo/apps/platform/src/signals/okou-page/composer-signals.ts:527` — **`setImageAnnotationLifecycleRef$` uses `onRef` purely as a lifecycle trigger.** The inner command ignores `_element` and only registers an abort listener that closes the editor and cancels pending annotation uploads. `docs/effect.md` calls this out directly: "do not use `onRef` merely because it provides a mount signal. If the element parameter is unused, prove that the behavior is still about the committed presence of that DOM subtree." I can see the argument — `<ImageAnnotationEditor>` now renders inside that same subtree — and `createComposerSignals` currently receives no `AbortSignal`, so there is no cleaner owner available today. Either thread an owner signal into `createComposerSignals` and drop the ref, or keep the ref and add a comment recording why the composer's committed DOM subtree is the correct owner, so the next reader does not have to re-derive it.

### Testing

- Coverage: Adequate for the feature. `chat-events.bdd.test.ts` proves the route projects the derivative into the prompt with the structured `[Image annotations]` block and round-trips the exact part; `image-annotation.test.tsx` covers send payload shape, draft restore, transport-only draft attachments, and send blocking while the derivative uploads; `chat-threads.test.ts` pins the draft-allows / send-requires asymmetry of the new `.refine`. Missing: the failed-upload chip state described in the first P1, and the annotation upload `failed → retry → uploaded` path exercised through `AnnotationUploadRetry` (Codecov reports 40 uncovered lines in `image-annotation.ts` and 11 in `chat-draft.ts`).
- Conventions: Compliant. Route tests go through `setupApp`/contract with API-only setup and verification, platform tests enter through the page helper and assert on user-visible surfaces and the outgoing request, mocking stays at `context.mocks` and browser built-ins, and `restoreMocks`/`unstubGlobals` handle cleanup.

e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 4052e0095f64d288be79e2bb8b89db6a3af8d832
PR_AUTO_REVIEW_VERDICT: LGTM

Re-review after 4052e00. **Correction first: the P0-adjacent finding in my previous review was wrong**, and it was the only reason that review requested changes.

### Summary

Collapses the two-file annotated send into one logical user-message file part carrying the immutable original `fileId`, the rendered `annotatedFileId`, and structured `annotations`; renders and uploads the derivative on confirmation with pending/retry/cancel states; and moves the annotation editor from a root singleton to a per-composer signal factory. 4052e00 additionally moves the derivative filename derivation onto the shared contract and documents the composer teardown owner.

### Correction to the previous review

My first P1 claimed that `attachment-chips.tsx` would spin forever after a failed original upload, because `uploadPending$` reads a status that never leaves `pending` when `upload$` rejects. The status observation is accurate, but the conclusion is not: the state is unreachable.

`chat-draft.ts:954 uploadAttachment$` wraps `upload$` in `tapError` and, on rejection, removes the attachment from `internalAttachments$` and raises the "Failed to upload …" toast. The chip is gone before anything can render a stale spinner. I verified this by writing the regression test I asked for — it failed with zero matching `aria-label`s in the DOM, because no chip exists at that point.

That path is also already covered: `views/okou-page/__tests__/chat-draft.test.tsx:1236` uploads one good and one failing file and asserts the toast, the removed chip, the surviving chip, and that Send stays enabled. It passes on this branch, so the `uploadPending$` / `sendReady$` refactor did not regress it.

I had started fixing the non-issue by adding a `failed` variant to `AttachmentUploadState`. That would have been a branch for a state the only writer cannot produce — exactly the slop `docs/fallback.md` §3 rules out — so it was reverted rather than shipped, along with the test and the `mockUploadFailure` helper it needed.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic: no unhandled edge cases or wrong assumptions | Pass | The previously reported `uploadPending$` state is unreachable; `uploadAttachment$` removes the attachment on rejection |
| `REVIEW.md` | Feature regression guard: changed existing behavior needs regression coverage | Pass | `chat-draft.test.tsx:1236` already covers the failed-upload path the refactor touches, and it passes on this branch |
| `REVIEW.md` | Security, type safety, documented interface changes | Pass | No new endpoint, credential, or `any`; both new part fields are documented at `contracts/chat-threads.ts:585` |
| `REVIEW.md` | DB/JSONB gate | Pass | `persistedAttachmentSchema` is a non-strict `z.object`, so old rows carrying `annotation` are stripped on read and old writers are not rejected |
| `docs/bad-smell.md` | No duplicated logic that should be shared | Pass (fixed in 4052e00) | `annotatedImageFilename` now lives in `contracts/chat-threads.ts:173`; the API service and `flatten-annotated-image.ts` both import it |
| `docs/bad-smell.md` | No suppressions, dynamic imports, or hardcoded config; fail fast | Pass | `resolve-draft-attachments.ts:186` throws an invariant error matching the existing `attachmentUploadFailureMessage` throw above it |
| `docs/fallback.md` | No fallback for a state the contract or single writer already prevents | Pass | Confirmed by reverting the `failed` upload variant I had added |
| `docs/fallback.md` | Removing the previous fields needs evidence, not a compatibility reader | Pass | `composerImageAnnotation` is `enabled: false` with one `enabledEmailHashes` entry, so §2 applies |
| `docs/fallback.md` | Every introduced fallback is declared in the PR summary | N/A | The diff introduces no new fallback |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | The old `failImageDecodes` / `Marks on …` assertions were deleted with the behavior |
| `docs/deployment-compatibility.md` | Frontend → backend changes are additive; both skew directions safe | Pass | Optional additions to the `.strict()` file part, emitted only by one staff email behind a non-GA switch |
| `docs/deployment-compatibility.md` | Persisted-state changes cover rows written by the previous version | Pass | Draft attachments are transport-only in `draft-persistence.ts:34`; annotation state rides the persisted `userMessage` |
| `docs/deployment-compatibility.md` | Cross-deployable derivations stay consistent | Pass (fixed in 4052e00) | The app writes and the API reads the same `annotatedImageFilename`, so the exact-key artifact lookup cannot silently degrade to a prefix listing |
| `docs/event-sourcing.md` | Optimistic events reconcile by shared ID and are not rolled back on failure | N/A | The diff changes file-part content, not optimistic event creation or reconciliation |
| `docs/effect.md` | Render stays pure — no Store writes, resource creation, or signal allocation | Pass | Signal factories run inside `createChatAttachment` / `createComposerSignals`; views only `useGet`/`useSet` |
| `docs/effect.md` | `onRef` only when the command depends on the mounted element | Pass (documented in 4052e00) | `composer-signals.ts:527` now records that `ImageAnnotationEditor` renders in that exact subtree, that an abandoned upload would otherwise strand the attachment in `pending`, and that an injected signal should replace it if the factory gains one |
| `docs/effect.md` | One semantic command owns the transition; async work has an owning signal | Pass | `commitAnnotation$` → `confirmAnnotations$`; `resetUploadSignal$` combines with `pageSignal` |
| `docs/cache.md` | RC-3/RC-4: no unbounded caches, state scoped to its real owner | Pass | The editor moved from module singletons in `views/main.tsx` to a per-composer factory |
| `docs/cache.md` | RC-7/RC-8: stable refs, symmetric acquire/release | Pass | `useSet(signals.setImageAnnotationLifecycleRef$)` passed directly as a ref; abort closes the session and cancels pending uploads |
| `.claude/skills/ccstate/SKILL.md` | `detach(..., Reason.DomCallback)` only in views | Pass | New `detach` calls are in the editor and chip views; `confirmAnnotations$` awaits its own work |
| `.claude/skills/ccstate/SKILL.md` | Signal factory pattern with a stable React key | Pass | `chat-draft.ts` adds `key: crypto.randomUUID()`; `AttachmentChips` keys on it instead of `String(a.fileInfo$)` |
| `docs/testing.md` | `feat:` commits carry integration tests | Pass | Route BDD coverage plus page-level composer coverage |
| `docs/testing/api-testing.md` | Route tests go through `setupApp()` and the contract, not services or the DB | Pass | The new BDD case sends via `requestSendEvent` and verifies via `api.readRun` and `waitForThreadMessages` |
| `docs/testing/testing-external-behavior.md` | Cases built and verified from the external surface | Pass | Platform tests drive the real composer, lightbox, and editor and assert the outgoing request body |
| `docs/testing/app-testing.md` | Platform tests are page-level, not component-only units | Pass | All new cases enter through the page helper; codec assertions extend a pre-existing signals suite |
| `docs/testing/patterns.md` | Mock at the boundary, real internals | Pass | `context.mocks.*` and browser built-ins only; no `vi.mock()` of internal paths |
| `docs/testing/anti-patterns.md` | AP-5/AP-8/AP-9: no fake timers, no over-testing schema, mock cleanup exists | Pass | The contract cases exercise the custom cross-field `.refine`; `vitest.config.ts` sets `clearMocks`, `restoreMocks`, `unstubGlobals` |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None outstanding. Both remaining findings from the previous review are addressed in 4052e00:

- `annotatedImageFilename` now lives on the contract (`turbo/packages/api-contracts/src/contracts/chat-threads.ts:173`) and is imported by `turbo/apps/api/src/signals/services/chat-user-message.service.ts` and `turbo/apps/platform/src/signals/okou-page/flatten-annotated-image.ts`. The writer and the reader of that name can no longer drift across two deployables.
- `turbo/apps/platform/src/signals/okou-page/composer-signals.ts:527` documents why the composer's committed DOM subtree owns the annotation-editor teardown, what breaks without it, and the condition under which an injected `AbortSignal` should replace the ref.

### Testing

- Coverage: Adequate. Route projection, send payload shape, draft restore, transport-only draft attachments, send blocking during derivative upload, and the draft-allows / send-requires asymmetry of the new `.refine` are all covered; the failed-upload path is covered by the pre-existing `chat-draft.test.tsx` case. The annotation `failed → retry → uploaded` path via `AnnotationUploadRetry` remains uncovered (Codecov: 40 uncovered lines in `image-annotation.ts`, 11 in `chat-draft.ts`) — worth a follow-up, not a merge blocker for a non-GA switch.
- Conventions: Compliant.
- Re-verified locally on 4052e00: `pnpm format`, `NODE_OPTIONS=--max-old-space-size=3072 pnpm turbo run lint --concurrency=1` (13/13), `pnpm check-types` (14/14), `pnpm knip` (clean), `@okouai/api-contracts` `chat-threads.test.ts` (27 passed), and `@okouai/app` `image-annotation.test.tsx` + `attachment-chips.test.tsx` (64 passed).

PR #31060, opened 2026-09-02
fix(runner): cap pre-spawn admission to available cpus

### LINKED ISSUE #31041: bug(rust): cap pre-spawn admission at process-available CPU capacity
## Problem

Runner startup combines two CPU scopes when it builds the fresh pre-spawn admission gate:

- `host::cpu_count()` uses `std::thread::available_parallelism()`, which estimates the logical CPUs available to the Runner process.
- `pre_spawn_cpu_capacity_at()` uses that process-scoped value only when topology is absent. When topology is present, `physical_core_count()` instead walks every CPU in the host-wide `/sys/devices/system/cpu/online` list and returns that count unchanged.

As a result, CPU affinity, systemd `AllowedCPUs=`, container cpusets, or cgroup CPU limits can make the Runner-visible capacity smaller than the sysfs topology count while the pre-spawn gate still uses the larger host count.

For example, with `logical_cpu_count = 2` and sysfs exposing eight logical CPUs across four physical cores, the current function returns `ExactPhysical(4)`. Two 2-vCPU fresh starts can then enter the expensive preparation-to-Agent-ready interval concurrently even though the Runner has only two logical CPUs available.

This scope mismatch is reproducible in a Linux cgroup environment: a small Rust program reported `available_parallelism() == 1`, while `/sys/devices/system/cpu/online` reported `0-1` and the two entries had distinct `(physical_package_id, core_id)` pairs. Passing that process-visible count into the current implementation therefore produces `ExactPhysical(2)`.

## Impact

The gate was introduced to keep fresh starts below the host-contention cohort. Overstating its token count under process CPU restrictions defeats that bound and can increase concurrent Firecracker creation, restore, storage, and bootstrap work. Startup logging also reports `physical_topology` as the capacity source without revealing that the topology exceeds the CPU capacity available to the process.

## Relevant code

- `crates/runner/src/host.rs:22-32` obtains the process-visible logical count and passes it to the capacity calculation.
- `crates/runner/src/host.rs:36-49` ignores that value whenever physical topology is present.
- `crates/runner/src/host.rs:59-98` counts all host-online physical cores without intersecting them with the process's effective CPU set.
- `crates/runner/src/cmd/start/mod.rs:575-582` initializes `PreSpawnAdmission` with the resulting token count.
- `crates/runner/src/executor/sandbox_run.rs:461-496` holds those tokens across the complete fresh pre-spawn interval.

## Acceptance criteria

- Pre-spawn admission capacity never exceeds the CPU capacity available to the Runner process.
- Physical-core counting considers only CPUs in the process's effective affinity set and also respects a smaller quota-derived `available_parallelism()` result, or applies an equivalently conservative policy.
- Tests cover host-online topology that is larger than the process-visible CPU capacity and verify that the returned token count is bounded.
- Tests cover an affinity set containing SMT siblings so filtering preserves physical-core deduplication.
- Startup observability identifies the effective restricted capacity source and does not claim a larger exact capacity than the Runner can use.

## Related work

- #28855 / #28882 introduced the process-local physical-core-weighted pre-spawn gate.
- #30896 / #30902 cover admission headroom and weighted Firecracker cgroups, but explicitly do not add cpusets and do not correct this startup capacity calculation.

seven332:
## Summary

- restrict physical-core topology accounting to CPUs in the Runner process affinity mask
- cap the resulting pre-spawn admission capacity by `available_parallelism()` so CPU quota remains authoritative
- expose restricted capacity with a distinct startup source and cover affinity, SMT, quota, fallback, and malformed-topology behavior

## Scope

This changes only the Runner startup calculation for the process-local pre-spawn admission gate. It does not change permit weights or lifetime, exact-reuse behavior, the whole-run resource budget, Firecracker placement, APIs, queues, persistence, or guest protocols.

## Validation

- `cargo test -p runner host::tests -- --nocapture`
- `cargo build -p runner`

Closes #31041

### DIFF
--- a/crates/runner/Cargo.toml
+++ b/crates/runner/Cargo.toml
@@ -19,7 +19,7 @@ base64 = { workspace = true }
 bytes = { workspace = true }
 chrono = { workspace = true, features = ["serde", "now"] }
 clap = { workspace = true, features = ["derive", "env"] }
-nix = { workspace = true, features = ["user", "signal", "fs", "process", "inotify"] }
+nix = { workspace = true, features = ["user", "signal", "fs", "process", "inotify", "sched"] }
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
 serde_path_to_error = { workspace = true }
--- a/crates/runner/src/cmd/start/mod.rs
+++ b/crates/runner/src/cmd/start/mod.rs
@@ -601,10 +601,18 @@ async fn run_start_with_home(
         profiles = runner_config.profiles.len(),
         "resource budget initialized"
     );
-    match pre_spawn_capacity {
+    match &pre_spawn_capacity {
         host::PreSpawnCpuCapacity::ExactPhysical(_) => {
             info!(
-                capacity_source = "physical_topology",
+                capacity_source = pre_spawn_capacity.source(),
+                pre_spawn_vcpu_tokens = pre_spawn_admission.total_tokens(),
+                host_logical_cpus = host_cpus,
+                "pre-spawn admission initialized"
+            );
+        }
+        host::PreSpawnCpuCapacity::RestrictedPhysical(_) => {
+            info!(
+                capacity_source = pre_spawn_capacity.source(),
                 pre_spawn_vcpu_tokens = pre_spawn_admission.total_tokens(),
                 host_logical_cpus = host_cpus,
                 "pre-spawn admission initialized"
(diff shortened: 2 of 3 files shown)


seven332:
## Code Review: PR #31060

**Reviewed head:** `d56476303a29f535d2ea204d6e0714d55436cb99`

### Summary

The Runner now computes pre-spawn admission from physical cores reachable through its effective CPU affinity and caps that value by `available_parallelism()`. The change preserves unrestricted-host and topology-missing behavior while reporting restricted capacity explicitly.

### Findings

No findings.

### Testing

Formatting, Clippy, rustdoc, focused host tests, all Runner targets/features (3,398 passed; 24 ignored, plus the guest inventory test), Runner build, and repository hooks passed. A live production-style restricted cgroup was not created locally; deterministic temporary-sysfs cases cover affinity, SMT, quota-style capping, and topology failures, and the real affinity query has a smoke test.

### Verdict

LGTM

seven332:
Closing because this admission-capacity change is no longer required.

PR #30966, opened 2026-09-02
feat(desktop): pick a window by sight and select an area on any screen

bingjiez666:
Five things reported from a real machine after the first recordings with #30743 and #30868.

## The window list was unusable

The picker showed whatever the system enumerated, as plain text. On the reporting machine that meant twenty-eight Control Center entries, several `Item-0` and `Menubar` rows, and the recorder's own overlays — all before any actual document, and all looking alike.

Windows are now chosen from a grid of live previews with the owning application named, and system chrome is filtered by bundle id (Control Center, Dock, Notification Center, Spotlight, SystemUIServer, WindowManager, and our own two bundles). Previews come from `desktopCapturer`, joined to the helper's sources by the CoreGraphics window id both sides carry; a window the capturer could not preview is dropped rather than shown as a blank tile.

The picker lives in its own window: the bar is sized to its row of controls, so a grid drawn inside it would be clipped away exactly like the error message in #30868.

## An area could only be drawn on one screen

`selectArea` opened a single overlay on the primary display, so on a multi-display Mac the other screens could not be reached. One selector now opens per display, each carrying its display id, and the region is rebased onto that screen before capture. The recording controls also follow the recorded screen instead of always sitting on the primary one.

## Start in the selector did not start anything

It resolved the selection back to the bar, so the user had to press Start again over there. It now begins the recording. The audio toggles travel with the selection, because the bar is gone by the time the capture starts — that also removes the `area` variant from the renderer's capture request, since an area request is now assembled in the main process from the display the drag happened on.

## The bar came back stale

It was hidden rather than closed, so reopening it showed the previous session's region (`1704 × 1114`) and a Start button still reading "Starting…". The bar is rebuilt each time it opens; its choices belong to one recording.

## The menu bar listed a file path

`Saved to /Users/…/screen-recording-….mp4` sat under **New Recording…** after every capture. The recording is delivered into the app, so the path was not something to act on. The submenu now offers only the new recording.

## Testing

- New `desktop-recorder-window-options.test.ts` covers the join, the chrome filter, dropping preview-less windows, and grouping by application.
- `desktop-ipc-boundary.test.ts` extends the sender-guard set to the new channels and pins the new payload validation (a region must name its display; a chosen window must name its source).
- `desktop-tray-menu.test.ts` now asserts the submenu offers only **New Recording…** once a recording has finished.
- `pnpm -F @okouai/desktop check-types`, `lint`, `build`, `test` (475 passed) and `pnpm knip` all clean.

## Notes

[...]

PR #31065, opened 2026-09-02
fix(desktop): offer every space in the window picker and let it be dismissed

bingjiez666:
Three defects found while recording with the bar, all in the window picker added by #30966.

## 1. The picker only listed the active Space

`SCShareableContent.getExcludingDesktopWindows` was called with `onScreenWindowsOnly: true`. On macOS "on screen" means the current Space, so a full-screen editor or browser on its own Space never appeared — and could not be recorded at all, since the same read resolves the capture target later. Reported with ChatGPT and VS Code missing while both were open in their own Spaces.

Now read with `onScreenWindowsOnly: false`.

## 2. The wider read listed the system's own chrome

Reading every Space also returns menu bar extras, the Dock, and notification banners. They were shown as ordinary choices titled after their owning process — the picker offered `Menubar`, `Menubar`, `StatusIndicator` ahead of any document.

New `RecordableWindowPolicy` keeps only windows on the ordinary application layer, which is where real windows live and where none of that chrome does. `recorder.sources` and `recorder.windowPreviews` both go through it, so the two lists still line up by window id.

## 3. Leaving Window mode left the pick pending

`selectWindow` resolves only when something answers it. Clicking `Display` or `Area` answered nothing, so:

- the picker stayed floating on top of everything, and
- tearing the bar down with the call still in flight surfaced Electron's `Error invoking remote method 'desktop-recorder:select-window': reply was never sent` as a recording error — visible for a moment before the recording started normally.

Both paths now abandon the open pick first.

## 4. The mode button was renamed after the chosen window

`Window` was replaced by the chosen window's title, so picking a window called `Menubar` produced a button reading `Menubar` sitting between `Display` and `Area` — it read as a third capture mode. The button keeps its own name; the target is shown underneath and in the tooltip.

## Testing

- `pnpm -F @okouai/desktop check-types` — clean
- `pnpm -F @okouai/desktop lint` — clean
- `pnpm -F @okouai/desktop exec vitest run` — 477 passed, 2 failed in `bootstrap-degraded.test.ts`; both fail identically on unmodified `main` (`3d08aa1`) in this container and are unrelated to this change (auto-update mocks)

New coverage:

- `RecordableWindowPolicyTests` covers the layer rule — an ordinary window is offered, menu bar extras and the Dock are refused, and an untitled window the picker could not label is refused.
- `recorder-bar.test.tsx` covers the two renderer regressions. Each case was confirmed to fail against the pre-fix bar: the mode button renamed itself after a window titled `Menubar`, and choosing `Display` or `Area` never called `completeWindowSelection`, which is what left the picker floating and produced the "reply was never sent" error.

The Swift helper cannot be compiled in a Linux container. The `Desktop` workflow builds and tests it on `macos-latest` for this PR, which is where `RecordableWindowPolicy`, the changed `main.swift`, and the new Swift tests get exercised.

`onScreenWindowsOnly: false` itself has no unit test — it is a single ScreenCaptureKit call flag with no seam to assert on. The filtering it makes necessary is what `RecordableWindowPolicy` covers.

### DIFF
--- a/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderCore/RecordableWindowPolicy.swift
+++ b/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderCore/RecordableWindowPolicy.swift
@@ -1,0 +1,25 @@
+import Foundation
+
+/// Decides which of the system's windows the picker may offer.
+///
+/// The recorder reads every Space rather than only the active one, so a
+/// full-screen editor or browser is offerable too. That wider read also carries
+/// the system's own surfaces — menu bar extras, the Dock, notification banners —
+/// which are never what someone means to record. They sit above the normal
+/// window layer, and that is what separates them here.
+public enum RecordableWindowPolicy {
+    /// The layer ordinary application windows live on. Everything the system
+    /// floats above them — status items, the Dock, banners — is numbered higher.
+    public static let applicationWindowLayer = 0
+
+    /// Whether a window belongs in the picker.
+    ///
+    /// An untitled window is refused as well: the picker labels each choice with
+    /// its title, and a nameless tile is worse than one fewer choice.
+    public static func isRecordable(title: String?, windowLayer: Int) -> Bool {
+        guard let title, !title.isEmpty else {
+            return false
+        }
+        return windowLayer == applicationWindowLayer
+    }
+}
--- a/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderHelper/main.swift
+++ b/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderHelper/main.swift
@@ -99,9 +99,13 @@ private func shareableContent(timeout: TimeInterval = 10) throws -> SCShareableC
     }
     let semaphore = DispatchSemaphore(value: 0)
     let box = ResultBox<SCShareableContent>()
+    // `onScreenWindowsOnly` is off because "on screen" means the active Space.
+    // With it on, a full-screen editor or browser living on its own Space was
+    // missing from the picker, and picking it later could not be resolved
+    // either. `recordableWindows` drops the system chrome the wider list adds.
     SCShareableContent.getExcludingDesktopWindows(
         true,
-        onScreenWindowsOnly: true
+        onScreenWindowsOnly: false
     ) { content, error in
         box.set(value: content, error: error)
         semaphore.signal()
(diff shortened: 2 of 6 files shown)


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 146e61e1902e09752fd7d3f0e50ab467d929df04
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Repairs four defects in the window picker from #30966: it listed only the active Space, it listed the system's own chrome once the Space limit was lifted, it left `selectWindow` pending when the user switched to Display or Area, and it renamed the mode button after the chosen window.

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

#### Notes (non-blocking, P2)

- `RecordableWindowPolicy` keeps only `windowLayer == 0`. That is the correct cut for the reported chrome — menu bar extras, the Dock, and notification banners all float above it, and full-screen windows stay on layer 0. It also drops any application window an app deliberately floats above the normal level, such as a tool palette at `NSFloatingWindowLevel`. Those were listed before this change when they happened to be on the active Space, so this is a narrow trade of that case against gaining every Space; worth revisiting only if a real window turns up missing.
- `abandonWindowChoice` calls `completeWindowSelection(null)` unconditionally, including when no pick is open. That is safe by construction on the main side — `settleWindowChoice` resolves an optional and `closeWindowPicker` no-ops on a null or destroyed window — but it does mean the bar sends one IPC message per mode switch rather than tracking whether a pick is outstanding.
- `.recorder-bar__source` moves from a fixed `height: 64px` to `min-height`, so the row grows by a line when a window is chosen. Flex stretches all three buttons together, so they stay one control group, but the bar itself is a few pixels taller in that state. No automated coverage; visual only.

### Testing

- Coverage: Adequate. Both renderer regressions and the window-layer rule carry tests, and each renderer case was verified to fail against the pre-fix bar rather than only passing against the fix:
  - `recorder-bar.test.tsx` — the mode button keeps the name `Window` with a window titled `Menubar` as its target, and choosing `Display` or `Area` calls `completeWindowSelection(null)`. Confirmed red on `HEAD~1`: 3 failed / 0 passed.
  - `RecordableWindowPolicyTests` — an ordinary application window is offered; menu bar extras and the Dock are refused; an untitled window the picker could not label is refused.
- Not covered, deliberately: `onScreenWindowsOnly: false` is a single ScreenCaptureKit call flag with no seam to assert against. The chrome it lets through is what `RecordableWindowPolicy` covers, and the `Desktop` workflow compiles and runs the helper on `macos-latest` for this PR.
- Conventions: Compliant. The renderer test uses the harness `App.test.tsx` already established (`// @vitest-environment happy-dom` + testing-library), stubs only the `window.vm0DesktopRecorder` bridge boundary rather than any internal module, uses no fake timers, and asserts rendered output and the bridge contract rather than call counts alone.
- Local run: 477 passed, 2 failed. Both failures are in `bootstrap-degraded.test.ts` (auto-update mocks) and reproduce identically on unmodified `main` at `3d08aa1` in the same container, so they are environmental and unrelated to this diff.
- Containment: the recorder stays behind the non-GA `desktopScreenRecording` switch, scoped to the maintainer.
- No fallback or cross-version compatibility path is introduced, so no `Fallbacks` section applies.

PR #31081, opened 2026-09-02
feat(ui): refresh the app palette onto the sand and amber scales

Lunarivibe:
Applies the palette published at https://zero-design-color.sites.vm0.io to the web app.

The app already funnelled almost every colour through `packages/ui`'s semantic token layer, so this is a swap of the two primitive scales rather than a component-by-component edit. Only three call sites in the whole app referenced a numbered stop directly.

## What changed

| Layer | Before | After |
| --- | --- | --- |
| `--gray-*` | cool neutral, hue 210-220 | **Sand** — one warm neutral anchored on Linen `#faf5f3` and Ink `#242121` |
| `--primary-*` | orange, brand at stop 700 (`#ed4e01`) | **Amber**, brand at stop 300 (`#ffa500`) |
| `--background` (light) | `#fbfcfd` | Linen `#faf5f3` |
| `--destructive` | `#ef4444` both themes | Coral 700 light / Coral 400 dark |
| link, `<mark>`, selection, chart segments | ad-hoc hues | published Cobalt / brand tint / categorical set |

Every generated value is derived from the published hexes by script, not eyeballed. The two ends of each scale are the published values exactly.

## Two judgement calls

**1. Amber cannot carry white text, so `--primary-foreground` is now Ink.**

White on `#ffa500` is 1.97:1. Ink is 8.09:1, which matches the figure the palette site publishes. This flips the label on all 28 `text-primary-foreground` call sites at once. Dark mode's destructive fill moves to Coral 400 for the same reason and also takes Ink (7.21:1).

**2. The Amber scale is adopted stop-for-stop, which means it is no longer mirrored between themes.**

The old ramp inverted in dark — `primary-100` was a pale tint in light and a near-black in dark — so several call sites used a *raw stop* as a brand wash and relied on that inversion. Adopting the published scale verbatim breaks that, and would have blown out `::selection`, `<mark>`, the markdown accent colours and the auth link in dark mode.

Rather than re-mirror the scale, those call sites now use two theme-aware tokens, `--brand-subtle` and `--brand-text`, which are the palette's own published brand tokens. Raw stops stay for fills that already pair with `--primary-foreground`.

## On the dark ramp

The palette publishes four dark surfaces. This app needs eight distinct elevations — page, nav, card, popover, input, border, hover, segment — so taking the published four literally would have collapsed the elevation hierarchy and lifted every card by 10 points of lightness.

Instead the dark ramp keeps its existing step sizes and shifts as a whole until the page background lands on Ink, the published dark canvas. The generated stops are interpolated along Sand's own hue path, so nothing is off-palette: stop 50 is Ink exactly, 950 is Linen exactly, and 800 lands on Sand 500, the published dark muted-text colour.

## The shell, behind a feature switch

A second body of work landed on top of the palette swap after review of the
first preview, gated behind a new `newUi` `FeatureSwitchKey` scoped to
`ming@vm0.ai` (`enabled: false`, `enabledEmailHashes: ["54757055"]`).

| Surface | Switch off | Switch on |
| --- | --- | --- |
| nav rail | `--gray-50` `#FAFAF8` | `#EFEAE8`, the marketing site's nav grey |
| chat column | `--gray-0` `#FDFCFB` | `#FDFAF9`, half the depth of the site's `#FAF5F3` theme grey |
| workspace | white, edge to edge | white sheet inset 8px on all four sides, 12px radius, on the chat column's grey |
| chat column divider | present | removed |
| dark composer focus ring | `42.9 1.6% 37.5%` | `38.8 38% 38%` |

Three notes on the choices:

- **The chat column and the workspace gutter are one surface.** Keeping the
  column's `border-r` would put a second hairline 8px from the card's own
  border and read as a double rule, so the divider is dropped when the switch
  is on. Verified by pixel-scanning the deployed preview.
- **The theme grey is halved rather than replaced.** Lightness moves to the
  midpoint with white while saturation holds, which halves the channel spread
  by the same factor and keeps the site's hue: `#FAF5F3` → `#FDFAF9`.
- **Dark's focus ring takes the brand hue.** The light ring can stay neutral
  because there it darkens, and darker-than-the-border already reads as
  emphasis. Lifting a neutral against a dark card lands on a bright grey that
  reads as a white outline instead of as focus. Saturation, not lightness, does
  the work, so it stays quieter than the full Amber stop.

`--color-sidebar` and `--color-sidebar-rail` are declared in the platform sheet
as intentional shadows of the design system's tokens, registered with their
reason in `design-token-shadowing.test.ts`. The card layout is desktop-only
(`min-width: 48rem`); below the sidebar breakpoint the columns are gone and a
frame would only spend width.

## Also in this PR

Both of these were listed as out of scope in the original description and were
brought in afterwards:

- **The product mark moves to `#ffa500`.** Four `fill="#ED4E01"` paths in
  `product-brand-mark.tsx`, plus the badge in the automation-result email
  renderer, which also takes Ink text because white on Amber is 1.97:1.
- **`apps/desktop` is refreshed.** Its self-contained stylesheet has no shared
  tokens, so 19 neutrals are re-hued by hand, the brand moves to Amber with
  hover stepping *down* rather than up, four white-on-brand labels take Ink,
  and the status hues move to Coral / Jade / Cobalt / Amber.

## Not gated by the switch

The palette itself. Base tokens and the brand hex reach the logo SVG, the boot
skeleton in `index.html`, and the automation-result email — all three render
before the stylesheet loads or outside the browser entirely, so none can read a
per-user switch. Splitting the palette into its own held-back PR is the only way
to stage it per user, and that is not what this PR does.

## Verification

- Derivation script re-derives every value from the published `tokens.css`; the state layer is solved so that hover at 5% over a white card composites to exactly `#faf5f3`, and the solve is verified to be exact rather than clamped.
- Contrast checked on all changed pairings: Ink on brand 8.09:1, muted text on canvas 7.04:1, white on destructive 5.76:1, Ink on dark destructive 7.21:1, link on card 5.48:1.
- `pnpm -F @okouai/ui exec vitest run src/styles` — 17 passed.
- `vite build` clean; verified the emitted CSS carries the new anchors and the new `brand-*` utilities, and that no `20 99% 47%` (old orange) or `215 100% 19%` (old cool state layer) survives.

One existing test pinned `--primary-600`'s literal HSL triplet to assert the focus ring resolved from the primary scale. It now asserts the relationship — that `--ring` points at a stop, and that the stop is defined — rather than the value, so the palette can move without the test needing to.

- `pnpm -F @okouai/app exec vitest run src/views/okou-page/__tests__/new-ui-feature-switch.test.tsx` — 5 passed, covering the document attribute in both shells and the chat column's divider on each side of the switch. Mutation-checked: inverting the divider gate turns exactly those two assertions red.
- Both switch states measured on the deployed preview rather than inferred: off gives rail `rgb(250,250,248)` / column `rgb(253,252,251)` / divider 1px / no card; on gives `rgb(239,234,232)` / `rgb(253,250,249)` / divider 0 / 8px padding and 12px radius.


Lunarivibe:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 12b460c17c1f3fa595002556d447be2de8ea3b69

### Summary

Swaps the app's two primitive scales onto the published Sand/Amber palette, then adds a second, later body of work on top: the workspace now renders as a white card floating on the shell grey, the two sidebars take the marketing site's own greys, the chat column loses its divider, and dark's composer focus ring moves from a lifted neutral to the brand hue. That shell work is gated behind a new `newUi` feature switch scoped to `ming@vm0.ai`.

Two blockers, both about the PR as a reviewable unit rather than about the colours: the summary now describes a different change than the diff makes, and the new feature switch ships with no test even though an exact precedent exists in this repo.

One note that is **not** a finding: `applyNewUiDocumentAttribute` mutates `document.documentElement` from an inline ref callback rather than through `onRef`, which `docs/effect.md` treats as using the DOM as a generic lifecycle trigger. This PR does not introduce that pattern — it reuses the identical call shape already used by `applyColorThemeDocumentAttributes` and `applyTypefaceDocumentAttribute` in the same ref. Pre-existing debt, better fixed for all three at once.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no unhandled edge cases | Pass | Changes are CSS token values, one enum member, one registry entry, and one attribute helper. `:where([data-theme="dark"][data-new-ui])` is placed after the ungated dark rule and matches its `(0,2,0)` specificity, so the gated value wins by order as intended. |
| `REVIEW.md` | Security — no injection, secrets, or missing auth | N/A | No endpoints, queries, credentials, or auth paths in the diff. The `enabledEmailHashes` entry is an FNV-1a hash, and `feature-switch.ts` already documents that this system is not an authorization boundary. |
| `REVIEW.md` | Type safety — no `any`, no broad types | Pass | `applyNewUiDocumentAttribute(enabled: boolean)`; `newUiEnabled` is `boolean` via `?? false`. No casts added. |
| `REVIEW.md` | Style and maintainability; comments explain WHY | Pass | New CSS comments state the reason (double rule at 8px, neutral lift reading as a white outline), not the mechanics. No function exceeds ~100 lines. |
| `REVIEW.md` | New user-facing features gated behind a `FeatureSwitchKey` | Pass | `FeatureSwitchKey.NewUi`, `enabled: false`, `enabledEmailHashes: ["54757055"]`. Verified `fnv1a("ming@vm0.ai") === "54757055"`. |
| `REVIEW.md` | Feature regression guard — replaced behaviour needs scoped containment and coverage | Finding (P1) | Containment is present and correct (switch off restores `--color-sidebar: gray-0` / `--color-sidebar-rail: gray-50` and the divider), but nothing tests either branch. See P1-2. |
| `REVIEW.md` | `feat:` commits carry integration tests | Finding (P1) | The palette work has tests (`globals.test.ts`, `design-token-shadowing.test.ts`, `auth-v2-style-assertions.ts`). The `newUi` gating added on top has none. See P1-2. |
| `REVIEW.md` | PR summary describes the diff | Finding (P1) | Summary asserts two exclusions that the diff violates and omits the entire second body of work. See P1-1. |
| `docs/fallback.md` | Default is no fallback; declare any fallback the diff introduces | Pass | The diff introduces none. The `newUi` off-branch is the current production path held behind a rollout gate, not a compatibility branch — §2 ("features behind a feature switch need no fallback") covers the *new* side, and the off side is live behaviour, not a tolerance branch. No `??`/`||` chains, retired-protocol readers, or fabricated defaults added. No `Fallbacks` section required. |
| `docs/fallback.md` | No negative tests asserting removed behaviour stays removed | Pass | No tombstone tests added. `design-token-shadowing.test.ts` gains a positive allow-list entry with a stated reason, which is the file's documented purpose. |
| `docs/bad-smell.md` | No `any`; no lint or type suppressions | Pass | No `any`, no `eslint-disable`, no `@ts-ignore` in the diff. |
| `docs/bad-smell.md` | No hardcoded URLs or environment-specific configuration | Pass | No URLs or env values added. Colour literals are design tokens, which is the intended location for them. |
| `docs/bad-smell.md` | Fail fast — no fallback or recovery logic hiding misconfiguration | Pass | `features[FeatureSwitchKey.NewUi] ?? false` reads an optional map key with a defined default; it does not mask a contract-guaranteed value. Matches the existing `GeistTypeface` and `GradientColorThemes` reads. |
| `docs/testing.md` | Integration tests only, at entry points | Pass | Changed tests render through `detachedSetupPage` or read built CSS. No new unit tests on internal functions. |
| `docs/testing.md` | Mock at the boundary; no relative `vi.mock`, no fetch stubbing, no fake timers | Pass | No mocks added in this diff. |
| `docs/testing.md` | `feat:` work needs coverage of the behaviour it introduces | Finding (P1) | See P1-2. |
| `docs/effect.md` | Render is pure; DOM lifecycle work belongs in `onRef` | Pass | The new call runs in a committed ref callback, not during render, and is symmetric on detach (`element !== null && newUiEnabled`). Uses the established local pattern; the `onRef` deviation predates this PR and applies equally to the two existing helpers. |
| `docs/effect.md` | State is derived, not synchronized | Pass | `newUiEnabled` is derived from `featureSwitch$` at read time; no ref, effect, or command copies it into a second source of truth. |
| `docs/deployment-compatibility.md` | Independently deployed components stay compatible across the skew window | Pass | Adding an enum member only widens the `effectiveSwitches` map. An old web client (up to ~2 days stale) ignores the unknown `newUi` key and keeps rendering the current shell; a new client against an old API reads `undefined` and falls to `false`, which is the off branch. No schema, migration, queue payload, or runner protocol touched. |

### Findings

- **PR description contradicts the diff on two points and omits a third.** The summary's "Not in this PR" section states *"The product mark stays `#ed4e01`"*, but `turbo/apps/platform/src/views/components/product-brand-mark.tsx` changes all four `fill="#ED4E01"` paths to `#FFA500` in this PR. The same section states `apps/desktop` *"belongs in its own PR"*, but `turbo/apps/desktop/src/renderer/styles.css` carries 189 changed lines here. Separately, the summary never mentions the workspace card, the sidebar greys, the removed chat-column divider, the dark composer ring, or the `newUi` switch — roughly half the diff. A reviewer reading only the summary would approve the opposite of what ships, and the palette-only framing hides the shell change entirely. Rewrite the description against the current diff, including the switch and its rollout audience.

- **`newUi` ships with no test on either branch.** `turbo/apps/platform/src/views/okou-page/__tests__/typeface-feature-switch.test.tsx` is the exact precedent: it asserts `document.documentElement.dataset.typeface` is unset with the switch off, and set with it on, across both `SidebarLayout` and the minimal shell. `newUi` has the identical structure — the same two shells, the same ref helper — plus a component-level gate that `GeistTypeface` does not have (`ChatListColumn` drops `border-r-[0.7px] border-sidebar-border` when the switch is on). None of it is covered. `sidebar.test.tsx` (83 tests) passes only because it exercises the default off branch implicitly; nothing pins that the off branch keeps the divider, and nothing proves the on branch removes it. Add a `new-ui-feature-switch.test.tsx` modelled on the typeface test covering: attribute absent when off, attribute present when on in both shells, and the chat column's divider present when off / absent when on.

### Testing

- **Coverage: Insufficient.** The palette half is covered — `packages/ui/src/styles/__tests__/globals.test.ts` now asserts the `--ring` → `primary-N` relationship instead of a literal triplet, which is the right call for a moving palette, and `design-token-shadowing.test.ts` correctly registers `--color-sidebar-rail` as an intentional shadow with a stated reason. The `newUi` half has no test at all (P1-2).
- **Conventions: Compliant.** No `vi.mock` with a relative path, no `vi.stubGlobal("fetch", ...)`, no fake timers, no assertions on mock call counts, no mocking of `globalThis.services.db`. Changed tests assert observable output — computed CSS and rendered DOM — rather than implementation details.
- **CI:** 78 checks passing, 0 failing on `12b460c1`. Note that the previous head failed `design-token-shadowing` for a real reason (the new `--color-sidebar-rail` shadow was undeclared); that was fixed in `12b460c1` rather than worked around.

Lunarivibe:
LGTM


PR_AUTO_REVIEW_COMMIT: 7affe96461541cc2759f9ba91aecb1031472f0b5
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Re-review of the two P1 findings raised on `12b460c1`. Both are addressed and nothing else in the diff moved: the only code change since the previous review is one added test file (+101 lines, no source touched).

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | PR summary describes the diff | Pass | Both contradicted claims are gone from the body. It now documents the `newUi` switch and its rollout audience, the per-surface off/on table, the removed chat-column divider, the dark focus ring, and carries an "Also in this PR" section that states the product mark and `apps/desktop` were brought in after the original scope. A "Not gated by the switch" section names the three surfaces that cannot read a per-user switch. |
| `REVIEW.md` | `feat:` commits carry integration tests | Pass | `new-ui-feature-switch.test.tsx`, 5 tests, modelled on `typeface-feature-switch.test.tsx`. Renders through `detachedSetupPage` and asserts observable output. |
| `REVIEW.md` | Feature regression guard — the existing flow most likely to regress is covered | Pass | The off branch is asserted directly, not implied: the document attribute stays unset and the chat column keeps `border-r-[0.7px]`. Both shells are covered on the on branch. |
| `docs/testing.md` | Tests must be able to fail for a real reason | Pass | Mutation-checked. Inverting `!newUiEnabled` to `newUiEnabled` in `sidebar.tsx` turns exactly the two divider assertions red and leaves the three attribute assertions green, which is the correct blast radius for that mutation. |
| `docs/testing.md` | Integration only; boundary mocks; no fake timers; assert behaviour not mocks | Pass | No `vi.mock`, no `vi.stubGlobal`, no fake timers, no call-count assertions. Uses the existing `context.mocks.data.agents` fixture and `featureSwitches` option already used by the sibling test. |
| `docs/testing.md` | No over-testing | Pass | Five tests, no redundancy: off/on for the attribute, one for the minimal shell, off/on for the divider. The divider pair covers behaviour the typeface switch has no equivalent for. |
| `docs/fallback.md` | No negative tests asserting removed behaviour stays removed | Pass | The divider-off test asserts current production behaviour under the off branch, not the absence of deleted code. |
| `docs/bad-smell.md` | No `any`, no suppressions, no hardcoded configuration | Pass | None in the added file. |
| `REVIEW.md` | Everything else assessed on `12b460c1` | Pass | Unchanged — no source file moved between the two heads. |

### Findings

None. Both P1 findings from the previous review are resolved.

### Testing

- **Coverage: Adequate.** The palette half keeps its existing coverage; the shell half now has `new-ui-feature-switch.test.tsx` (5 passed) on both sides of the switch, verified non-vacuous by mutation.
- **Conventions: Compliant.**
- **CI:** 78 passing, 0 failing on `7affe964`.

PR #31089, opened 2026-09-02
fix(python): enforce inline firewall api lists

### LINKED ISSUE #31003: quality(python): enforce inline firewall API-list registry invariant
## Summary

The supported execution-firewall pipeline already guarantees that an inline firewall's `apis` value is an array: the TypeScript contract requires an array, Rust deserializes it into `Vec<FirewallApi>`, and the runner serializes the typed value into the private proxy registry.

The Python registry consumer nevertheless validates this untyped JSON file independently and classifies malformed sandbox entries in isolation. That boundary currently applies the inline `apis` list invariant inconsistently: custom-connector entries are rejected when `apis` is not a list, while ordinary inline entries are accepted and later skipped by the tolerant matcher compiler.

Align the registry boundary so every inline entry has the same list-valued `apis` invariant. This is boundary-contract hardening and consistency work, not a confirmed production exploit. No supported producer that emits the malformed shape has been identified.

## Confirmed source evidence

- `turbo/packages/connectors/src/firewall-contracts.ts` requires inline execution firewalls to contain an `apis` array.
- `crates/runner/src/types.rs` deserializes `Firewall.apis` as a required `Vec<FirewallApi>`.
- `crates/runner/src/proxy/registry.rs` serializes typed `FirewallEntry` values into the private, atomically replaced proxy registry.
- `crates/runner/mitm-addon/src/registry_firewalls.py` validates inline `apis` only inside the `customConnectorId` branch.
- `crates/runner/mitm-addon/src/matching.py` intentionally keeps direct compiler calls tolerant by skipping firewalls whose `apis` value is not a list.

## Scope

- Validate the `apis` container for every inline firewall at Python registry resolution, before optional source or custom-connector identity propagation.
- Reuse `FirewallEntryResolutionError` so only the affected sandbox is classified as `invalid_firewalls`.
- Keep the direct `matching.compile_firewalls()` tolerance contract unchanged; the stricter rule belongs only to registered sandbox input.
- Do not change TypeScript contracts, Rust types, the persisted registry format, or deployment protocols.

## Behavior change

Valid producer output is unchanged. An unsupported proxy-registry entry whose inline firewall has a missing or non-list `apis` value changes from an active sandbox with no compiled firewall to an `invalid_firewalls` sandbox that receives the existing local `invalid_registry_sandbox` response.

Absent or null top-level `firewalls`, and valid list-valued inline `apis`, retain their current behavior.

## Acceptance criteria

- Require every inline firewall resolved from the proxy registry to have a list-valued `apis` field.
- Classify a malformed ordinary-inline sandbox as `invalid_firewalls` without preventing a valid peer sandbox in the same registry from loading.
- Prove through the request-handling entry point that the malformed sandbox receives the existing `503 invalid_registry_sandbox` response before authentication injection or upstream handling.
- Preserve valid ordinary and custom inline firewall behavior.
- Preserve the intentional tolerant behavior of direct `matching.compile_firewalls()` callers.
- Keep the change Python-only and add focused integration coverage at registry and request-handler boundaries.

seven332:
## Summary

- require list-valued `apis` for every inline firewall resolved from the proxy registry
- classify only the malformed sandbox as `invalid_firewalls` while valid peers remain available
- verify malformed ordinary inline entries receive the existing local 503 before auth injection

Supported TypeScript and Rust producers already guarantee this shape. This aligns the Python registry boundary for unsupported malformed registry JSON; it is not based on a confirmed production exploit.

## Scope

- keeps direct `matching.compile_firewalls()` tolerance unchanged
- preserves absent or null top-level `firewalls`, valid inline firewalls, and nested matcher semantics
- does not change TypeScript or Rust contracts, persisted formats, APIs, protocols, migrations, or rollout sequencing

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_registry_inline_firewalls.py tests/test_request_handler_registry_admission.py tests/test_compiled_firewall_malformed_policies.py` (65 passed)
- `uv run --no-sync python -m pytest tests/` (7,333 passed)

Closes #31003

### DIFF
--- a/crates/runner/mitm-addon/src/registry_firewalls.py
+++ b/crates/runner/mitm-addon/src/registry_firewalls.py
@@ -316,14 +316,14 @@ def resolve_firewall_entries(
     path. A supplied snapshot pins builtin resolution for this call so a single
     registry pass cannot mix catalog versions.
 
-    Inline firewalls are deep-copied and receive per-entry `None` builtin cache
-    keys. Builtin catalog API IDs are discarded during expansion, so builtin
-    APIs receive generated run-scoped IDs. Inline APIs preserve an existing
-    non-empty string ID; absent, empty, or non-string IDs are generated.
-    Generated IDs use `<runId>:<index>`, where the zero-based index advances over
-    dictionary API entries in resolved firewall order, including entries whose
-    IDs are preserved. Callers must validate `sandbox["runId"]` as a non-empty string
-    before calling.
+    Inline firewalls are deep-copied, must contain a list-valued `apis` field,
+    and receive per-entry `None` builtin cache keys. Builtin catalog API IDs are
+    discarded during expansion, so builtin APIs receive generated run-scoped
+    IDs. Inline APIs preserve an existing non-empty string ID; absent, empty, or
+    non-string IDs are generated. Generated IDs use `<runId>:<index>`, where the
+    zero-based index advances over dictionary API entries in resolved firewall
+    order, including entries whose IDs are preserved. Callers must validate
+    `sandbox["runId"]` as a non-empty string before calling.
 
     When `sandbox["firewalls"]` is present, `connectorRuntimeTargets` defaults to
     an empty list when absent. When supplied, it must be a list of object targets.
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #31089

**Reviewed head:** `a5de5135ec9ea684e6c69a89fa2d38917b521ae5`

### Summary

The submitted change moves the existing inline `apis` container check to the shared registry-resolution boundary, preserving direct matcher tolerance while making ordinary and custom inline entries use the same per-sandbox invalid classification. The implementation stays within the approved Python-only scope.

### Findings

No findings.

### Testing

The review verified real-file mixed-registry isolation, request-hook local 503 and auth non-injection coverage, and the preserved malformed direct-compiler contract. uv lock/sync, Ruff format/lint, BasedPyright, 65 focused tests, and the complete 7,333-test addon suite passed. Turbo and Rust checks were not run because their code and contracts are unchanged.

### Verdict

LGTM

PR #31112, opened 2026-09-02
fix(desktop): surface a failed movie writer instead of delivering its first fragment

bingjiez666:
## What went wrong

A twelve-minute desktop recording came back as a two-second movie. Its click track spanned the full twelve minutes and the container's own metadata claimed nineteen seconds, so the capture stream was alive the whole time and only the frames stopped landing.

The `AVAssetWriter` failed early and nothing in the helper noticed:

- A failed writer answers `isReadyForMoreMediaData` with `false` for the rest of its life. The sample handler's guard treated that as ordinary back-pressure and silently dropped every later frame, while `latestSampleAt` — and therefore the on-screen clock and `durationMs` in the click track — kept advancing.
- `stop` finalized the failed writer, never looked at `assetWriter.status` or `.error`, read the file size, and handed the file to delivery as a finished recording.

Nothing in the helper referenced `.failed` on the writer at any point. Running the branch on macOS then turned up the reasons the writer was failing in the first place, and several separate defects on the same path; they are all in this PR.

## Changes

**Surfacing a failed writer (`main.swift`, `desktop-recorder-native.ts`, `desktop-recorder-controller.ts`)**

- The sample handler checks `assetWriter.status` before the readiness guard and after any rejected `append`. A failed writer ends the capture through the same external-stop path a lost source uses, reported as `capture_failed` (the source is still there; the output broke). `externalStop` now carries the code it should be reported with instead of hard-coding `source_lost`.
- `stop` checks the writer after `finishWriting` and returns the file together with a `failure` field when the writer broke or the capture had already ended in failure. The file is still handed back — it holds whatever was written — but the caller must not deliver it as finished. Reported on `stop` as well as through `recorder.state` because a stop that races the one-second poll would otherwise ship it.
- `latestSampleAt` only advances for frames that reach the append path, so the clock no longer runs on dropped frames.
- The tray showed only AVFoundation's wrapper, "The operation could not be completed". The message now spells out the domain and code of every underlying error, and names the sample the writer refused: track, format, duration, media time, and where the previous sample ended.
- `recorder.stop` gets its own sixty-second budget. The helper can block up to forty seconds inside `stop` (ten draining the stream, thirty finalizing), but every request shared the fifteen-second budget, so a long recording's finalize was abandoned mid-flight; `restoreStatusOnFailure` then put the controller back to `recording` against a helper that had already moved to `stopped`, and every later command was refused as `invalid_state`.
- `toRecording` carries the `failure` field through, and `stop` plus the poll-driven salvage path honor it: the salvaged file is kept as `lastRecording`, the error is shown, and delivery waits for a deliberate retry.

**Why the writer was failing (`main.swift`, `SampleTimingPolicy.swift`)**

- **Samples the writer refuses.** The writer refuses a sample that does not advance its track, and once it has refused one it refuses everything after it. Audio buffers stamped before the anchor video frame were folded onto the anchor as duplicates, and an audio buffer could start before the previous one had finished. `SampleTimingPolicy` decides where a sample lands or that it must be dropped, and is unit-tested.
- **Fragments that could not be closed.** The movie was written as fragmented MP4 (`movieFragmentInterval` of two seconds). Closing a fragment failed whenever the video track had no frame in it, and a window capture only produces a frame when the window changes, so a still window left every fragment after the first empty. The movie is now written as one movie, finalized at stop; screen frames also get the capture interval as their duration rather than relying on a successor frame.
- **Frames of the wrong size.** The video track is built for exactly the prepared output size. A whole display is always scaled into the configured size, but a window or a cropped region is delivered at its own pixel size unless the stream is told to scale it, so the writer failed on the first frame. Window and area streams now set `scalesToFit`, and the first frame is measured against the track before the session is anchored on it; a mismatch ends the capture with a message naming both sizes.
- **Frames with no picture.** ScreenCaptureKit's idle, blank and bookkeeping frames carry nothing the encoder can take; only frames whose `SCFrameStatus` is `complete` reach the writer.

**Clicks stay aligned with the picture (`ClickMapping.swift`, `ClickTrackRecorder.swift`, `main.swift`)**

- A pause was measured from the last written sample, whose time froze at the moment of the pause, so every pause span came out empty: the movie kept a frozen stretch and every click after a pause sat that much later than the picture. Pause and resume are read from the host clock instead.
- A window capture follows the window wherever it is dragged, but clicks were projected through the geometry the capture was prepared with. Each click now carries the geometry of its own moment; the recorded window's bounds are re-read four times a second while recording.

**Start and finish read as working (`main.ts`, `recording-controller.tsx`, `main.swift`)**

- Pressing Stop made the floating controls vanish at once, seconds before the finalize and upload finished, which read as the recorder having quit. The controller now stays up through `finalizing` and `delivering`, shows which it is doing with its controls disabled, and is dismissed when the session ends. Covered by `recording-controller.test.tsx`.
- Starting took seconds. Opening the picker reads the screen twice and preparing the capture a third time, and a window capture needs the read that walks every Space — which #31065 made wider. One read is now kept for twenty seconds and reused, `prepare` for a display or area capture asks for on-screen windows only, and the delivery check is answered when the bar opens rather than on Start. A window that closes inside the cache's span still fails cleanly as `source_lost`. Each start phase is timed and logged.
- A refused stop left no trace: the controls vanished, no browser opened, and the tray showed nothing. The rejection went to the controller window that had just been closed, `restoreStatusOnFailure` put the status back without recording why, and the helper's stderr was never read. A refused stop is now recorded in state for the tray, the helper's stderr is kept and carried on an abrupt exit, and the main process logs the rejection.
- The helper connected to the window server lazily from whichever thread asked first. On current macOS that asserts (`CGS_REQUIRE_INIT`) rather than connecting when the first ask comes from a background thread, and every request is served on one, so the picker's window enumeration could abort the helper mid-session. The connection is made on the main thread at startup.

**Delivering a long recording (`desktop-auth-session.ts`)**

- The recorder uploads the video, then its click track. The token is short-lived, so after a long upload the second request arrived with an expired bearer, and its cookies no longer answered for the API either — the delivery failed at the last step. A request rejected on both now mints a fresh token and retries once; a refresh that yields nothing means the sign-in itself is gone and the 401 stands. `getAuthState` no longer opens a second hidden sign-in window for the same answer.

**Protocol compatibility.** `failure` is additive. A new client against an old helper sees no field and behaves as before; an old client against a new helper ignores the field. The desktop app ships helper and client together in any case.

## Not changed, noted for follow-up

- The helper's stdio loop handles requests serially, so `recorder.state` polls queue behind a long `stop` and time out at fifteen seconds while it runs. Harmless today — the controller is in `finalizing` and skips the poll — but worth a concurrent reader if `stop` ever runs longer.
- The reported elapsed clock is still derived from the last appended sample, so a still window recorded with audio off stops advancing it at the last visual change. Reusing the host clock, or padding the last frame at stop, is a trade-off between a reported duration longer than the encoded movie and a frozen clock.
- The window-bounds timer runs on the sample-delivery queue.
- Window captures hard-code `scale: 2` in `CaptureGeometry` rather than the display's backing scale, so click mapping is off by two on a non-Retina display.

## Testing

- `pnpm -F @okouai/desktop check-types`, `lint` — clean
- `pnpm -F @okouai/desktop exec vitest run` — 481 passed, 2 failed in `bootstrap-degraded.test.ts`; both fail identically on unmodified `main` in this container (auto-update mocks) and are unrelated
- `desktop-recorder-controller.test.ts`: a stop result carrying `failure` is kept, not delivered, and delivered on explicit retry — verified to fail against the pre-fix controller (1 failed / 21 passed). A poll-noticed failure whose stop result also carries `failure` is reported; this one passes on the old controller too, since the poll path already honored its own argument, and is kept as a guard.
- `desktop-recorder-native.test.ts`: `stop` outlives the ordinary request budget while `state` still times out at it; the helper's `failure` field is carried through `toRecording`; a helper that dies mid-request carries its stderr.
- `SampleTimingPolicyTests.swift`, `ClickMappingTests.swift`: the timing rule and the moved-window click projection.
- `recording-controller.test.tsx`: the controls stay up through `finalizing` and `delivering`, disabled, naming the phase.

The Swift helper cannot be compiled here. The `Desktop` workflow builds and tests it on `macos-latest` for this PR.

### DIFF
--- a/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderCore/ClickMapping.swift
+++ b/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderCore/ClickMapping.swift
@@ -81,6 +81,12 @@ public struct CapturedClick: Equatable, Sendable {
     public let button: String
     public let clickCount: Int
     public let modifiers: [String]
+    /// Where the captured region stood when this click happened, for a source
+    /// that moves during the recording. A window capture follows the window
+    /// wherever it is dragged, so a click after the drag must be projected
+    /// through the window's position at that moment, not the one it had when
+    /// the capture was prepared. `nil` means the recording's geometry applies.
+    public let geometry: CaptureGeometry?
 
     public init(
         ticks: UInt64,
@@ -88,7 +94,8 @@ public struct CapturedClick: Equatable, Sendable {
         screenY: Double,
         button: String,
         clickCount: Int,
-        modifiers: [String]
+        modifiers: [String],
+        geometry: CaptureGeometry? = nil
     ) {
         self.ticks = ticks
         self.screenX = screenX
@@ -96,6 +103,7 @@ public struct CapturedClick: Equatable, Sendable {
         self.button = button
         self.clickCount = clickCount
         self.modifiers = modifiers
+        self.geometry = geometry
     }
 }
 
@@ -134,7 +142,7 @@ public func projectClicks(
             continue
         }
         guard
-            let point = geometry.mapClick(
+            let point = (click.geometry ?? geometry).mapClick(
                 screenX: click.screenX,
                 screenY: click.screenY,
                 outputSize: outputSize
--- a/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderCore/SampleTimingPolicy.swift
+++ b/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderCore/SampleTimingPolicy.swift
@@ -11,16 +11,26 @@ import Foundation
 /// anchor as duplicates, and any sample whose adjusted time fails to advance.
 /// Both are dropped here rather than handed to the writer.
 public enum SampleTimingPolicy {
+    /// Slack allowed when comparing against the previous sample's end, so
+    /// floating-point drift between neighbouring buffers is not read as overlap.
+    public static let overlapTolerance = 0.000_001
+
     /// The media time to write `captureSeconds` at, or `nil` to drop it.
     ///
     /// - `captureSeconds`: the sample's time relative to the anchor frame.
     /// - `pauses`: the recording's pause spans.
     /// - `lastWrittenSeconds`: the media time of the last sample written to
     ///   the same track, or `nil` when the track has none yet.
+    /// - `lastWrittenEndSeconds`: where that sample ended, for tracks whose
+    ///   samples have extent. Audio buffers carry 20 ms of sound each, and the
+    ///   writer refuses one that starts before the previous one has finished:
+    ///   a window capture's audio jumps like that when the window's app starts
+    ///   making sound. Video frames pass their own start here.
     public static func mediaTime(
         captureSeconds: Double,
         pauses: PauseTimeline,
-        lastWrittenSeconds: Double?
+        lastWrittenSeconds: Double?,
+        lastWrittenEndSeconds: Double? = nil
     ) -> Double? {
         guard captureSeconds >= 0 else {
             return nil
@@ -29,6 +39,9 @@ public enum SampleTimingPolicy {
             return nil
         }
         if let lastWrittenSeconds, mediaSeconds <= lastWrittenSeconds {
+            return nil
+        }
+        if let lastWrittenEndSeconds, mediaSeconds + overlapTolerance < lastWrittenEndSeconds {
             return nil
         }
         return mediaSeconds
--- a/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderCore/SampleTimingPolicy.swift
+++ b/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderCore/SampleTimingPolicy.swift
@@ -1,0 +1,36 @@
+import Foundation
+
+/// Decides where a captured sample lands on the movie's timeline, or that it
+/// must not be written at all.
+///
+/// The writer refuses a sample whose time is not later than the previous one
+/// on the same track, and once it has refused one it refuses everything after
+/// it for the rest of the recording. Two things in the capture produce such
+/// samples: audio buffers that arrive stamped earlier than the video frame the
+/// session was anchored on, which the pause arithmetic would fold onto the
+/// anchor as duplicates, and any sample whose adjusted time fails to advance.
+/// Both are dropped here rather than handed to the writer.
+public enum SampleTimingPolicy {
+    /// The media time to write `captureSeconds` at, or `nil` to drop it.
+    ///
+    /// - `captureSeconds`: the sample's time relative to the anchor frame.
+    /// - `pauses`: the recording's pause spans.
+    /// - `lastWrittenSeconds`: the media time of the last sample written to
+    ///   the same track, or `nil` when the track has none yet.
+    public static func mediaTime(
+        captureSeconds: Double,
+        pauses: PauseTimeline,
+        lastWrittenSeconds: Double?
+    ) -> Double? {
+        guard captureSeconds >= 0 else {
+            return nil
+        }
+        guard let mediaSeconds = pauses.mediaTime(forCaptureTime: captureSeconds) else {
+            return nil
+        }
+        if let lastWrittenSeconds, mediaSeconds <= lastWrittenSeconds {
+            return nil
+        }
+        return mediaSeconds
+    }
+}
--- a/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderHelper/ClickTrackRecorder.swift
+++ b/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderHelper/ClickTrackRecorder.swift
@@ -50,6 +50,10 @@ final class ClickTrackRecorder: @unchecked Sendable {
     private var tap: CFMachPort?
     private var runLoopSource: CFRunLoopSource?
     private var warnings: [String] = []
+    /// Where the captured region stands right now, for a source that moves.
+    /// Asked at each click, so the click is projected through the geometry of
+    /// its own moment rather than the recording's first. Set before `start`.
+    var geometryProvider: (() -> CaptureGeometry?)?
 
     /// Starts observing clicks. Never throws: losing the click track must not
     /// cost the user their video, so a refused tap is reported as a warning on
@@ -129,7 +133,8 @@ final class ClickTrackRecorder: @unchecked Sendable {
             screenY: Double(location.y),
             button: button,
             clickCount: Int(event.getIntegerValueField(.mouseEventClickState)),
-            modifiers: modifierNames(for: event.flags)
+            modifiers: modifierNames(for: event.flags),
+            geometry: geometryProvider?()
         )
         lock.lock()
         clicks.append(click)
--- a/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderHelper/main.swift
+++ b/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderHelper/main.swift
@@ -517,9 +517,14 @@ private final class RecorderSession: NSObject, SCStreamDelegate, SCStreamOutput,
         try? FileManager.default.removeItem(at: url)
 
         let assetWriter = try AVAssetWriter(outputURL: url, fileType: .mp4)
-        // Fragmented output keeps the file playable if this process dies mid
-        // capture; an unfinalized plain MP4 would be unreadable.
-        assetWriter.movieFragmentInterval = CMTime(seconds: 2, preferredTimescale: 600)
+        // Written as one movie, finalized at stop, rather than in two-second
+        // fragments. Fragments were meant to keep the file playable if this
+        // process died mid-capture, but nothing recovers such a file, and
+        // closing a fragment failed whenever the video track had no frame in
+        // it: a window capture only produces a frame when the window changes,
+        // so a still window left every fragment after the first empty, and the
+        // writer failed at the boundary. Every recording of a still window
+        // ended a few seconds in.
 
         let video = AVAssetWriterInput(
             mediaType: .video,
--- a/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderHelper/main.swift
+++ b/turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderHelper/main.swift
@@ -86,7 +86,17 @@ private func resolveDisplay(
 
 // MARK: - Source discovery
 
-private func shareableContent(timeout: TimeInterval = 10) throws -> SCShareableContent {
+/// Reads what ScreenCaptureKit can share.
+///
+/// `onScreenWindowsOnly` is off by default because "on screen" means the active
+/// Space: a full-screen editor or browser on its own Space would otherwise be
+/// missing from the picker and unresolvable as a capture target. Callers that
+/// only need displays pass `true`, because the wide read walks every window on
+/// every Space and is the slowest part of starting a recording.
+private func shareableContent(
+    timeout: TimeInterval = 10,
+    onScreenWindowsOnly: Bool = false
+) throws -> SCShareableContent {
     // Asking ScreenCaptureKit without the grant makes the system put its own
     // prompt on screen, every single time. Once the answer is already no, say
     // so instead: the app can then offer Settings rather than the user being
@@ -99,13 +109,9 @@ private func shareableContent(timeout: TimeInterval = 10) throws -> SCShareableC
     }
     let semaphore = DispatchSemaphore(value: 0)
     let box = ResultBox<SCShareableContent>()
-    // `onScreenWindowsOnly` is off because "on screen" means the active Space.
-    // With it on, a full-screen editor or browser living on its own Space was
-    // missing from the picker, and picking it later could not be resolved
-    // either. `recordableWindows` drops the system chrome the wider list adds.
     SCShareableContent.getExcludingDesktopWindows(
         true,
-        onScreenWindowsOnly: false
+        onScreenWindowsOnly: onScreenWindowsOnly
     ) { content, error in
         box.set(value: content, error: error)
         semaphore.signal()
@@ -351,7 +357,7 @@ private final class RecorderSession: NSObject, SCStreamDelegate, SCStreamOutput,
     /// Set when the stream ended without `stop` being called. The session stays
     /// finalizable so the partial recording and its click track are still
     /// written; only the reported status changes.
-    private var externalStop: (reason: StreamStopReason, message: String)?
+    private var externalStop: (reason: StreamStopReason, code: String, message: String)?
     private var stream: SCStream?
     private var writer: AVAssetWriter?
     private var videoInput: AVAssetWriterInput?
@@ -620,12 +626,18 @@ private final class RecorderSession: NSObject, SCStreamDelegate, SCStreamOutput,
         systemAudio?.markAsFinished()
         microphone?.markAsFinished()
 
+        var writerFailure: String?
         if let assetWriter {
             let semaphore = DispatchSemaphore(value: 0)
             assetWriter.finishWriting {
                 semaphore.signal()
             }
             _ = semaphore.wait(timeout: .now() + 30)
+            if assetWriter.status == .failed {
+                writerFailure =
+                    assetWriter.error?.localizedDescription
+                    ?? "The screen recording could not be written"
+            }
         }
 
         guard let url else {
@@ -651,7 +663,7 @@ private final class RecorderSession: NSObject, SCStreamDelegate, SCStreamOutput,
             startedAtUnixMs: startedAt
         )
 
-        return [
+        var result: [String: Any] = [
             "videoPath": url.path,
             "clickTrackPath": clickTrackPath,
             "durationMs": Int(duration * 1000),
@@ -659,6 +671,19 @@ private final class RecorderSession: NSObject, SCStreamDelegate, SCStreamOutput,
             "width": outputSize.width,
             "height": outputSize.height,
         ]
+        // The file is handed back either way — it holds whatever was written
+        // before the failure — but the caller must not deliver it as a finished
+        // recording. Reported here as well as through `recorder.state`, because
+        // a stop that races the poll would otherwise ship it.
+        lock.lock()
+        let failure = externalStop.flatMap { $0.reason == .failed ? $0 : nil }
+        lock.unlock()
+        if let writerFailure {
+            result["failure"] = ["code": "capture_failed", "message": writerFailure]
+        } else if let failure {
+            result["failure"] = ["code": failure.code, "message": failure.message]
+        }
+        return result
     }
 
     private func writeClickTrack(
@@ -712,7 +737,7 @@ private final class RecorderSession: NSObject, SCStreamDelegate, SCStreamOutput,
         ]
         if let externalStop, externalStop.reason == .failed {
             described["error"] = [
-                "code": "source_lost",
+                "code": externalStop.code,
                 "message": externalStop.message,
             ]
         }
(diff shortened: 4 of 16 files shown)


bingjiez666:
## Root cause evidence from the machine that produced the two-second recording

Unified log for `screen-recorder-helper` at the moment the capture broke:

```
14:16:47.652 screen-recorder-helper: (AVFCore) [com.apple.coremedia:] <<<< AVError >>>>
  AVLocalizedErrorWithUnderlyingOSStatus: Returning error (AVFoundationErrorDomain / -11800) status (-16122)
```

`-11800` is `AVErrorUnknown`, the wrapper AVFoundation uses when the real cause is a lower-level OSStatus. `-16122` is a CoreMedia status, not a disk (`-11807`), permission, or lost-source error, so this is the writer refusing media it was handed. This confirms the failure mode the first commit surfaces: the writer failed a few seconds in and the helper kept "recording" for twelve more minutes.

Commit `822921d` closes the class of refusal the capture path itself manufactures. An audio buffer stamped earlier than the video frame the session was anchored on has a negative capture time; `PauseTimeline.mediaTime` folded that onto the anchor, so the audio track received several samples at the same instant. `SampleTimingPolicy` now drops samples that predate the anchor and samples that would not advance their own track (tracked per writer input), and samples are retimed in their own timebase instead of on a 600 Hz grid.

Tested through `SampleTimingPolicyTests` on the macOS runner. The precise meaning of `-16122` is not established from public headers; if the ±10 lines of context around that log line name a different operation, the retiming change still stands as a correctness fix and the remaining cause gets its own follow-up.

bingjiez666:
## From the dev build: a helper crash, and where Start's seconds go

**Helper crash on opening the picker** — surfaced by the stderr capture added in `40d66bf`:

```
[screen-recorder-helper] Assertion failed: (did_initialize), function CGS_REQUIRE_INIT, file CGInitialization.c, line 44.
Screen recorder helper exited with signal SIGABRT
```

CoreGraphics connects a process to the window server lazily, on whichever thread first asks, and on current macOS it asserts rather than connecting when that first ask comes off the main thread. The helper's main thread only runs a run loop; every request is served on a global-queue thread. The client respawns the helper on the next request, which is why the picker "worked again after a while" — the assertion is a race on which call comes first. `bc4b8a3` makes the connection on the main thread before any request is served (`NSApplication.shared` with the `.prohibited` activation policy, plus `CGMainDisplayID()`).

**Start latency.** The first thing on the Start path was `canDeliver()`, which is `getAuthState()` — two sequential HTTPS round trips to the API — asked only when Start was pressed and ahead of permission, prepare and the stream start. On a slow link that alone is a second or more of "Starting…". The check now runs when the bar opens and Start reuses the answer for the life of the bar (five-minute lifetime; a failed check is never served to the next Start). Each start phase is now timed and logged in the main process:

```
Desktop screen recording started in 1834ms (permission 11ms, prepare 212ms, start 1611ms)
```

What remains after this is dominated by ScreenCaptureKit's own stream start, which is inherent. If the number reported is still well past two seconds, the phase breakdown says which side to look at.

bingjiez666:
## The refused sample, named

With `94bab12`'s message in place, the window-capture failure reads:

```
capture_failed: The operation could not be completed
  (AVFoundationErrorDomain -11800 / NSOSStatusErrorDomain -16341)
  while writing audio, 48000 Hz 2 ch lpcm, duration 20.0ms, at 6.06s
```

Audio, not video; format unchanged; six seconds in. The writer will not take an audio buffer that starts before the previous one has finished, and a window capture's audio can do exactly that: it carries only the sound of the window's own application, and its timestamps jump when that application starts making sound. A display capture mixes every application and runs continuously, which is why display captures record and window captures did not.

`a81b78e` extends `SampleTimingPolicy`: a sample that starts before the previous sample on its track ended is dropped (a microsecond of tolerance for float drift; video frames pass their own start as their end, so the rule only binds audio). Covered by `SampleTimingPolicyTests`. The failure message now also reports where the previous sample on the track ended, so any remaining refusal can be read directly.

Start latency on the branch, from the phase log: 168–330 ms end to end, of which the stream start is 150–250 ms.

bingjiez666:
## Fragment boundaries, not the audio

With `a81b78e`'s overlap rule in place, the next window-capture failure read:

```
while writing audio, 48000 Hz 2 ch lpcm, duration 20.0ms, at 4.066s, previous sample ended at 4.066s
```

The refused buffer starts exactly where the previous one ended, so audio timing is not the cause; the overlap rule stays as a correctness guard but did not fix this. Both observed failures sit a few dozen milliseconds past a two-second fragment boundary (4.066 s and 6.06 s with `movieFragmentInterval` of two seconds). The writer closes a fragment at the boundary; if that fails the writer moves to `.failed` and the next append — whichever track it is on — is the one that reports it. The audio buffer was a bystander.

Closing a fragment needs the duration of every sample in it. A screen frame arrives with no duration, and the writer infers it from the frame that follows. A whole-display capture always has a next frame within a frame interval; a window capture only produces a frame when the window changes, so the last frame before a boundary can have no successor by the time the fragment must be written. That is the one behaviour that separates display captures, which record, from window captures, which did not.

`da0822c` writes screen frames with the capture interval (1/30 s) as their duration, and when an audio append fails the message reports where the video track's last frame sits, so a boundary failure can be read as one.

bingjiez666:
## Fourth failure, fourth fragment boundary

An area capture on `da0822c` (explicit frame durations in place, fragments still on):

```
while writing audio, 48000 Hz 2 ch lpcm, duration 20.0ms,
  previous sample ended at 8.046s, video track last frame at 8.067s lasting 33.3ms
```

No `at …s` on this one: the writer was already `.failed` when the buffer arrived, so it was refused before any timing was computed. The video frame at 8.067 s had appended successfully a moment earlier. Appending the first sample past a boundary is what makes the writer close the fragment behind it, asynchronously; the close of `[6 s, 8 s)` failed, and the next append noticed.

Every failure since the timing fixes has been the first append after a two-second boundary — 4.066 s, 4.098 s, 6.06 s, 8.0x s — on whichever track happened to come next, with the sample itself continuous with its predecessor. Explicit frame durations (`da0822c`) did not change it. Display captures, which never failed, are the one source that always has a frame in every fragment.

`af6d782` stops fragmenting: the movie is written as one file and finalized at stop. Fragments were meant to keep a file playable if the helper died mid-capture, but nothing recovers such a file, so that protection was never realised; the cost was window and area recordings ending a few seconds in. The timing, format and duration guards from the earlier commits stay as correctness rules.

bingjiez666:
## First complete recording on the branch — and the delivery 401 it hit

With fragments off (`af6d782`) a window capture ran to completion. Delivery then failed:

```
delivery_failed: Preparing the upload of screen-recording-….clicks.json failed with 401:
  {"error":{"message":"Not authenticated","code":"UNAUTHORIZED"}}
```

The video uploaded; the click track's prepare, minutes later, arrived with an expired bearer. `fetchWithSessionAuth` dropped the bearer on 401 and retried with cookies only, and when the API refused those too the 401 was returned as-is. Any recording long enough for the token to expire between the two uploads failed on its last step.

`ea50373` makes a doubly rejected request mint a fresh token through the hidden refresh window and retry once; a refresh that yields nothing leaves the 401 standing. `getAuthState` skips its own hidden refresh when a cached token was present, since the request already made that attempt, so a fully signed-out desktop still opens the window exactly once (the existing `clears the cached token when auth state returns 401` test holds). Two new tests cover both branches and were confirmed to fail against `main`'s session (2 failed / 21 passed).

TypeScript only; the Swift helper is unchanged since `af6d782`, whose macOS build passed.

bingjiez666:
## Click track alignment review

Report: the recorded video and the click track's mouse events do not line up. The consumer (`okou video camera`) aligns by time — `fps` filter plus `sendcmd` at second offsets — so variable frame rate is not the cause. Two defects on the recording side are.

**On the clock — pausing never took any time out.** `pause()` and `resume()` both read the last *written* sample's time as "now". Nothing is written while paused, so `resume` saw the same instant as `pause`, and every span in `PauseTimeline` was empty. Consequences: a frozen stretch in the movie the length of the pause, clicks made during the pause kept, and every later click that much later than the picture. Anyone who paused got a misaligned track. Both now read the host clock, which is the clock ScreenCaptureKit stamps frames with, so the span is real and the existing `PauseTimeline` arithmetic does what it was written to do.

**On the screen — a moved window.** A window capture follows the window wherever it is dragged, but clicks were projected through the window's position at prepare time (`RecorderSession.geometry` is fixed). After a drag every click landed offset by the distance moved. The helper now re-reads the window's bounds four times a second while recording; the tap records the geometry of the moment with each click (`CapturedClick.geometry`), and `projectClicks` uses it. Covered by a new `ClickMappingTests` case where a click after a 100-point drag projects to where the user clicked rather than 100 points off.

Checked and sound: the anchor (first complete frame's PTS, converted to mach ticks with the timebase ratio), the retiming of both tracks against that anchor, the rejection of pre-anchor audio, and the tick-to-millisecond conversion.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 6b7eaf63948b07174b7abff7d50bdc2e54b015d5
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Makes a broken `AVAssetWriter` visible and recoverable instead of shipping its first fragment as a finished recording: the helper checks `assetWriter.status` before the readiness guard and after a rejected `append`, ends the capture as `capture_failed`, and returns the salvaged file with an additive `failure` field that the controller honors by keeping the file as `lastRecording` and waiting for a deliberate retry. Around that core, the branch also stops fragmenting the movie, refuses samples the writer would reject (`SampleTimingPolicy`), scales window/area captures to the video track, gives `recorder.stop` its own 60 s budget, keeps the floating controls up through `finalizing`/`delivering`, projects clicks through the geometry of their own moment for a dragged window, and adds a token-refresh retry to `DesktopAuthSession`.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no race conditions or API misuse | Finding (P1) | Lock discipline in `main.swift` `stream(_:didOutputSampleBuffer:of:)` is correct — every `noteExternalStop`/`noteWriterFailure` call site releases the non-recursive `NSLock` first. Two P1s below: the elapsed clock now freezes for a still source, and window-bounds polling shares the sample-delivery queue. |
| `REVIEW.md` | Security: no injected input, secrets, or missing auth checks | Pass | No new network surface or credential handling; `desktop-auth-session.ts` reuses the existing token/cookie path and adds only a forced refresh on a second 401. |
| `REVIEW.md` | Type safety: no `any`, no overly broad types | Pass | `failure` is typed `DesktopRecorderError` in `desktop-recorder-types.ts`; the helper payload is narrowed through the existing `isRecord`/`errorCode` helpers. The only `as unknown as` is the test bridge stub, matching `recorder-bar.test.tsx`. |
| `REVIEW.md` | Style and maintainability: no oversized functions or duplicated logic | Pass | `SampleTimingPolicy` extracts the timing rule into a tested unit; `writerFailureMessage`/`describeSample`/`describeFormat` are small and single-purpose. `restoreStatusOnFailure` is still used by `prepare`, so the inlined `stop` path leaves no dead code. |
| `REVIEW.md` | Interface changes documented in the PR summary | Finding (P1) | The description still frames the change around `movieFragmentInterval` (now removed outright) and omits the auth retry, per-click geometry, `SampleTimingPolicy`, the shareable-content cache, the window-server connection, and the deliverability cache. |
| `REVIEW.md` | Testing coverage: `fix:` commits carry a regression test | Pass | `desktop-recorder-controller.test.ts` (writer failure kept, not delivered; delivered on explicit retry; poll-noticed failure), `desktop-recorder-native.test.ts` (stop budget, `failure` carried, stderr on abrupt exit), `SampleTimingPolicyTests.swift`, `ClickMappingTests.swift`, `recording-controller.test.tsx`. |
| `REVIEW.md` | Feature regression guard: changed existing behavior is covered | Pass | The behavior this replaces — a salvaged recording delivered automatically — is still covered by the existing "can deliver a salvaged recording on request" test plus the two new controller tests. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No `turbo/packages/db`, contract, or Zod response-schema change; nothing persisted server-side. |
| `docs/bad-smell.md` | Zero `any`; zero lint/type suppressions | Pass | No `any`, `@ts-expect-error`, or `eslint-disable` anywhere in the diff. |
| `docs/bad-smell.md` | Fail fast; no error handling that hides configuration problems | Pass | `stop()` still throws on a missing or unreadable output file; the new path reports the writer failure rather than swallowing it. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | The only dynamic import is in `recording-controller.test.tsx`, required because the module reads the bridge at load, and matching `recorder-bar.test.tsx`. |
| `docs/bad-smell.md` | No hardcoded URLs or environment values | Pass | New constants are timing budgets (`DEFAULT_STOP_TIMEOUT_MS`, `DELIVERABILITY_CHECK_LIFETIME_MS`, cache lifetime), not environment configuration. |
| `docs/fallback.md` | Every fallback the diff introduces is declared in the PR summary | Pass | The one additive tolerance — `toRecording` accepting a result without `failure` — is declared under "Protocol compatibility". See Fallbacks below. |
| `docs/fallback.md` | No fallback for a state the contract already prevents | Pass | `isCompleteFrame` treating a missing `SCStreamFrameInfo.status` as complete is genuinely optional framework data (§6.2), not a contract-guaranteed field. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | Every new test asserts live behavior. |
| `docs/deployment-compatibility.md` | Cross-version boundaries stay additive | Pass | The desktop app is outside the frontend/backend/runner model, and helper and client ship in one artifact; `failure` is additive within it either way. |
| `docs/testing.md` | Integration-first; mock only at the system boundary | Pass | The native backend tests drive a real fake helper process over stdio; the controller tests drive the real controller through a backend fake at the process boundary. |
| `docs/testing.md` | No fake timers | Pass | No `vi.useFakeTimers`; the stop-budget test uses a real delayed helper response. |
| `docs/testing.md` | Test behavior, not mock calls | Pass | Assertions are on state, returned recordings, and rendered text; `expect(deliver).not.toHaveBeenCalled()` is a supporting assertion alongside state checks, not the sole one. |
| `docs/testing/anti-patterns.md` | AP-2: no direct `fetch` mocking | Pass | `desktop-auth-session.test.ts` stubs `fetch`, which is AP-2's stated exception — the subject under test *is* the fetch wrapper `fetchWithSessionAuth`, and it matches the file's existing tests. |
| `docs/testing/anti-patterns.md` | AP-4: no relative-path `vi.mock()` of internal modules | Pass | No `vi.mock()` of Desktop modules; only `vi.stubGlobal` of the external `window` bridge and `fetch`. |
| `docs/testing/anti-patterns.md` | AP-7/AP-10: test user-visible behavior, not implementation | Pass | `recording-controller.test.tsx` asserts the clock, "Finishing…"/"Uploading…", and control disabled state. AP-10 is scoped to the platform app; this mirrors the reviewed `recorder-bar.test.tsx` pattern for a renderer window that mounts one component. |
| `docs/testing/desktop-testing.md` | Choose the smallest matching entry point; helper tested at the protocol boundary | Pass | Native backend at the stdio protocol boundary, Swift policy in `ScreenRecorderCore` SwiftPM tests, renderer at its window's component root. |
| `docs/testing/desktop-testing.md` | Avoid wholesale unit tests for `src/main.ts` | Pass | The `main.ts` changes (deliverability cache, controller visibility, stop logging) add no `main.ts` unit test; the visible half is covered through `recording-controller.test.tsx`. |
| `docs/testing/desktop-testing.md` | Workspace-scoped CI checks before opening the PR | Pass | `check-types`, `lint`, and `vitest` reported clean apart from two `bootstrap-degraded.test.ts` failures that reproduce on unmodified `main`; the Swift helper is built and tested by the `Desktop` workflow. |
| `docs/effect.md` | Render stays pure; work runs at the owning boundary | Pass | `recording-controller.tsx` only adds `useState` plus a `setFinishing` call inside the existing poll effect; nothing new runs during render. |
| `docs/cache.md` | RC-3: no unbounded-lifetime caches | Pass | `deliverabilityCheck` (`main.ts`) and `ShareableContentCache` (`main.swift`) are single-slot with explicit lifetimes (5 min / 20 s), not identity-keyed stores that grow. |
| `docs/cache.md` | RC-8: acquired resources have symmetric teardown | Pass | `windowBoundsTimer` is cancelled from `stop`, `discard`, and `noteExternalStop`; the stderr listener dies with the child process. |

### Findings

#### High Priority (P1)

- `turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderHelper/main.swift:967` — `elapsedSecondsLocked()` still derives the clock from `latestSampleAt`, but `latestSampleAt` now advances only inside the successful `input.append` branch, and idle/blank frames are filtered out before it by `isCompleteFrame`. A window capture only produces a complete frame when the window changes, so recording a still window with audio off freezes both the floating controller's clock and the `durationMs` written into the click track at the last visual change, while the capture keeps running. The session already has a host-clock reading for exactly this reason — `captureSecondsNowLocked()`, introduced here because the last-sample clock froze across a pause. Either reuse it for the reported elapsed time, or pad the last frame's duration at `stop` so the movie and the reported duration agree; the trade-off (reported duration longer than the encoded movie vs. a frozen clock) is worth an explicit decision.
- `turbo/apps/desktop/native/computer-use-helper/Sources/ScreenRecorderHelper/main.swift:488` — `startWindowBoundsTracking` schedules the 250 ms bounds timer on `sampleQueue`, the same serial queue registered as the `sampleHandlerQueue` for the screen, system-audio, and microphone outputs (lines 621/627/634). Each tick runs a synchronous `CGWindowListCopyWindowInfo` round trip to the window server on the queue that must drain 30 fps of frames, so a slow window-server response delays sample handling in the capture this PR is hardening, and with `queueDepth = 6` that can cost frames. Give the bounds timer its own queue; `currentGeometry` is already lock-protected, so nothing else depends on it sharing `sampleQueue`.
- PR description — the summary is materially out of date against the 13 commits on the branch. It still presents the diagnosis and fix in terms of `movieFragmentInterval`, which later commits removed entirely in favor of one non-fragmented movie, and it never mentions the `DesktopAuthSession` refresh-and-retry (`desktop-auth-session.ts`), per-click geometry for a dragged window (`ClickMapping.swift`, `ClickTrackRecorder.swift`, plus the 250 ms bounds tracker), `SampleTimingPolicy`, the shareable-content cache, `connectToWindowServer`, or the deliverability cache in `main.ts`. This body becomes the squash-merge record; please refresh it so it describes the final diff.

### Fallbacks

- `desktop-recorder-native.ts:toRecording` — accepts a `recorder.stop` result with no `failure` field, so a helper that has not yet learned to report one still parses. Surface: none in the deployment-compatibility sense — the desktop app is outside the frontend/backend/runner model and ships helper and client in one artifact, so the tolerance only spans a locally mixed build. Removal condition: none stated, and none needed for an additive optional field on an intra-artifact protocol. Declared in PR summary: Yes, under "Protocol compatibility". Verdict: Justified.

### Testing

- Coverage: Adequate. The writer-failure regression is pinned on both sides — the controller keeps a `failure`-carrying recording without delivering it and delivers it only on explicit retry (verified by the author to fail against the pre-fix controller), and the native backend carries `failure` through `toRecording` while `stop` outlives the ordinary request budget. The timing rule and the moved-window click projection are covered by SwiftPM tests in `ScreenRecorderCore`. `scalesToFit`, `isCompleteFrame`, `frameSizeMismatch`, and the shareable-content cache live in the `ScreenRecorderHelper` executable target and are exercised only by the `Desktop` workflow build on `macos-latest`, which is consistent with the entry points `docs/testing/desktop-testing.md` defines.
- Conventions: Compliant. No relative-path `vi.mock()`, no fake timers, no database or filesystem mocking; the `fetch` stub falls under AP-2's fetch-wrapper exception and the renderer test follows the existing `recorder-bar.test.tsx` shape.

bingjiez666:
Heads-up for whoever reviews this: the click-alignment fixes here (pause span, moved-window geometry) are real but secondary. The actual reason every real recording's clicks fell outside the video is that `CGEvent.timestamp` is already nanoseconds and the timeline applied the 125/3 Apple silicon timebase to it a second time, so every offset was 41.7× too large. That fix is #31167, stacked on this branch because this PR is queued and cannot take more commits; it retargets to `main` automatically once this merges.

PR #31126, opened 2026-09-02
fix(platform): count share selections by message group

e7h4n:
## Problem

In the share panel, the right-hand checkboxes are rendered one per **visual message group**
(`SelectablePagedGroupRow`), and ticking one toggles every shareable event inside that group.
The counter, however, reported the number of selected **events**:

```ts
// signals/chat-page/chat-thread-sharing.ts
selectedCount$: computed((get) => {
  return get(internalSelectedBytes$).size; // ← event ids, not groups
}),
```

An assistant run group holds one `output.message` per turn, so a long run puts a dozen messages
behind a single checkbox. On a thread offering only four checkboxes, ticking the fourth jumped the
counter from `3 selected` straight to `13 selected` — the unit displayed did not match the unit the
reader can actually tick.

The existing test never caught it because its fixture gives each group exactly one message, so both
units coincide.

## Change

Key the selection by group (`beginEventId`) instead of by event, so the stored unit, the counter,
and the checkbox are all the same thing.

- `selectedCount$` is now the number of selected groups.
- The 1.5 MB budget still sums the text of every selected message.
- `create$` still submits every event id behind the selected groups; the API orders them by
  `seqId` server-side, so client ordering is irrelevant.
- Re-ticking a group that grew while selected still covers the new messages rather than clearing it.

## Test plan

- [x] New case `counts a multi-message group as one selection`: an active run with three answers in
      one group; one tick reports `1 selected` and shares all three event ids
- [x] Verified the new test fails when the counter goes back to counting events
- [x] `pnpm -F @okouai/app exec vitest run` — 165 files, 2212 tests passed
- [x] `pnpm format`, `pnpm turbo run lint`, `pnpm check-types`, `pnpm knip` (output identical to the
      clean-tree baseline)

### DIFF
--- a/turbo/apps/platform/src/signals/chat-page/chat-thread-sharing.ts
+++ b/turbo/apps/platform/src/signals/chat-page/chat-thread-sharing.ts
@@ -27,9 +27,25 @@ export interface ChatThreadSharingSignals {
   readonly close$: Command<Promise<void>, [AbortSignal]>;
   readonly toggle$: Command<
     ToggleSharedThreadSelectionResult,
-    [readonly ShareableChatEvent[]]
+    [string, readonly ShareableChatEvent[]]
   >;
   readonly create$: Command<Promise<void>, [AbortSignal]>;
+}
+
+// A visual message group is the only thing the reader can tick, so it is also
+// the unit the selection stores and counts. A single assistant run group can
+// hold a dozen output messages; counting those instead made one click jump the
+// counter from "3 selected" to "13 selected".
+interface SelectedGroup {
+  readonly events: readonly ShareableChatEvent[];
+  readonly bytes: number;
+}
+
+function groupBytes(events: readonly ShareableChatEvent[]): number {
+  const encoder = new TextEncoder();
+  return events.reduce((total, event) => {
+    return total + encoder.encode(event.text).byteLength;
+  }, 0);
 }
 
 export function createChatThreadSharingSignals(
@@ -40,11 +56,13 @@ export function createChatThreadSharingSignals(
   >,
 ): ChatThreadSharingSignals {
   const internalPhase$ = state<SharedThreadSelectionPhase>("idle");
-  const internalSelectedBytes$ = state<ReadonlyMap<string, number>>(new Map());
+  const internalSelectedGroups$ = state<ReadonlyMap<string, SelectedGroup>>(
+    new Map(),
+  );
   const internalCreatedSharedThreadId$ = state<string | null>(null);
 
   const start$ = command(({ set }, signal: AbortSignal) => {
-    set(internalSelectedBytes$, new Map());
+    set(internalSelectedGroups$, new Map());
     set(internalCreatedSharedThreadId$, null);
     set(internalPhase$, "selecting");
     return set(
@@ -55,7 +73,7 @@ export function createChatThreadSharingSignals(
   });
 
   const close$ = command(({ set }, signal: AbortSignal) => {
-    set(internalSelectedBytes$, new Map());
+    set(internalSelectedGroups$, new Map());
     set(internalCreatedSharedThreadId$, null);
     set(internalPhase$, "idle");
     return set(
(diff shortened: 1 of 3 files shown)


e7h4n (commented):
LGTM

### Summary

The share panel renders one checkbox per visual message group, but `selectedCount$` returned the number of selected events, so ticking one checkbox over a long assistant run jumped the counter from `3 selected` to `13 selected` on a thread offering four checkboxes. This PR re-keys the selection by group (`group.beginEventId`) so the counter, the stored unit, and the checkbox all agree, while the byte budget and the submitted event ids still work at message granularity.

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Notes (P2)

- `signals/chat-page/chat-thread-sharing.ts` — `SelectedGroup.events` stores the full `ShareableChatEvent[]`, so up to the 1.5 MB selection budget of message **text** is now retained in signal state; the previous shape kept only byte lengths. Every consumer (`toggle$` re-tick comparison, `create$`, `selectedEventIds$`) needs nothing but the ids, so `{ eventIds: readonly string[]; bytes: number }` would carry the same behavior without holding the text. Bounded and released by `start$`/`close$`, so this is a simplification rather than a leak (`docs/cache.md` RC-3 / RC-4).

- `signals/chat-page/chat-thread-sharing.ts` — the group key is not stable across completed-work folding. While a run streams, its messages form one group keyed by the first assistant event; once the run completes, `completedWorkPhaseFolding` rebuilds `visibleEvents` as `[...userEvents, finalEvent, ...trailing]`, so the rendered group's `beginEventId` becomes `finalEvent.id`. A selection made mid-run therefore leaves a map entry that no rendered checkbox maps to: it still counts toward `selectedCount$` and still ships in `create$`, and the reader can only clear it by cancelling the whole selection. The old event-keyed shape degraded differently (hidden events stayed selected but the visible checkbox still cleared its own), so this is not a regression in kind, but the stale entry is now unreachable. Keying by an identity that survives folding, or dropping keys that no longer match a rendered group, would close it. Reachability requires a run to complete while the share panel is open — worth confirming before acting on it.

- `views/okou-page/__tests__/chat-thread-sharing.test.tsx` — the fixture depends on `activeRunIds: [RUN_ID]` to keep the run active, because a completed run folds its earlier answers into separate hidden groups and the multi-message group under test would not exist. The assertion does fail loudly if that changes, but the dependency is invisible to a future reader; one line of WHY next to `activeRunIds` would keep the case self-explanatory.

### Testing

- Coverage: Adequate. A `fix:` commit carries a regression test (`counts a multi-message group as one selection`) that exercises the real defect — one tick over a three-message group reports `1 selected` and still submits all three event ids. The pre-existing cases for the two-group flow and for the oversized-group rejection are unchanged, so the byte-budget path and the base selection flow keep their coverage.
- Conventions: Compliant. Integration test at the page entry point via `detachedSetupPage`, no `vi.mock()` of internal paths, no fetch stubbing (`context.mocks.api` handles the contract), no fake timers, assertions on user-visible text rather than on mocks. The `closest(...)`/`querySelector('[role="checkbox"]')` locator matches the convention already used by the neighbouring oversized-group case.
- Not covered, pre-existing: the byte budget is only exercised through a single oversized group, never through accumulation across several selected groups. The refactor changes that accumulation from per-event to per-group sums; the totals are equivalent, so this is a note, not a request.

### Other checks

- Fallbacks: the diff introduces none. `stored?.events` is an ordinary `Map` miss, not a default on a contract-owned field, and there is no cross-version or compatibility branch — so no `Fallbacks` section is required in the PR summary.
- Deployment compatibility: client-only selection state. The `sharedThreads.create` contract is untouched, `eventIds` still carries message ids, and the server dedupes and orders by `seqId`, so client-side ordering and duplicate ids stay irrelevant. No persisted shape changes.
- Feature gating: the surface stays behind `FeatureSwitchKey.SharedThreadSharing`.
- Type safety: no `any`, no suppressions, no dynamic imports, no hardcoded configuration.

Posted as a review comment rather than an approval because GitHub does not allow approving one's own pull request.

Issue #30960, opened 2026-09-02
bug(api): Fal image generation failed with "Unexpected status code: 422" and the warn log embeds the user prompt


bingjiez666:
## Problem

Two separate things in one record.

**1. The failure itself.** A built-in Fal image generation failed on `2026-09-01T04:28:00.272Z` with nothing more diagnostic than `Unexpected status code: 422`. The user-facing error is the generic `Generation failed` (`failError("Generation failed")`), so neither the user nor the log says *why* Fal rejected it.

**2. The log line contains user content.** `providerFailureDetailsForLog` includes `detail` in `PROVIDER_FAILURE_DETAIL_KEYS`, and for this payload `detail` was the entire generation `input` object — including the user's full prompt text — truncated only by length. That user prompt is now in production logs.

## Occurrence

| UTC time | generationId | type | status | error |
|---|---|---|---|---|
| 2026-09-01T04:28:00.272Z | `2eb9312c-37a1-4389-a276-4ccbc796c818` | image | `ERROR` | `Unexpected status code: 422` |

## Original log record (sanitized)

```
message: Fal built-in generation webhook reported failed generation
context: BuiltInGenerationWebhooks
level: warn
fields.generationId: 2eb9312c-37a1-4389-a276-4ccbc796c818
fields.type: image
fields.status: ERROR
fields.error: Unexpected status code: 422
fields.detail: {"input":{"acceleration":"none","enable_safety_checker":true,
                "expansion_model":"None","image_size":{"height":1024,"width":1024},
                "num_images":1,"output_format":"png",
                "prompt":"<redacted:user-prompt, ~230 characters present in the production log>"
```

## Impact

Confirmed: 1 paid image generation failed with an unactionable error, and one user's prompt text is retained in `vm0-web-logs-prod`.
Not confirmed: whether 422 here is a safety-checker rejection, an invalid parameter, or a quota condition — the log does not carry Fal's response body.

## Code path

- `turbo/apps/api/src/signals/routes/webhooks-built-in-generations.ts:903-916` — Fal failure branch
- `turbo/apps/api/src/signals/routes/webhooks-built-in-generations.ts:273-290` — `PROVIDER_FAILURE_DETAIL_KEYS`, which includes `detail`
- `turbo/apps/api/src/signals/routes/webhooks-built-in-generations.ts:355-379` — `providerFailureDetailsForLog`

## Ask

[...]

PR #28736, opened 2026-08-23
test(api): enforce external behavior test boundary

e7h4n:
## Summary

- remove two cleanup-helper unit assertions that were accidentally covered by the Pi resource snapshot service-test exception
- keep the exception focused on the byte-identical Pi virtual-filesystem discovery contract
- make the cleanup cutoff constants and helpers module-private now that tests no longer import them

## Scan

- Timestamp: 2026-08-24 05:00:01 CST
- Base commit: `f86c836c2f5275ca97b288a757a3d5e118ca0566`
- API test files scanned: 274
- Final ESLint scan: 1,133 files, 0 errors, 0 warnings, 0 restricted imports in API tests
- Final secondary scan: 399 matches, 0 direct DB calls, and only the three documented service-contract imports remain

## Violations fixed

`pi-resource-snapshot.service.test.ts` had a file-level exception for the Pi resource discovery contract, but it also imported `pi-api-first-turn-cleanup.service.ts` and directly asserted two internal cutoff helpers. The cleanup result is not exposed by the production cron response, so these tests pinned unobservable implementation details rather than endpoint behavior.

The two helper assertions and their service import are removed. Their four production symbols are no longer exported solely for tests.

## Files changed

- `turbo/apps/api/src/signals/services/__tests__/pi-resource-snapshot.service.test.ts`: remove the unrelated cleanup-service import and two internal helper assertions
- `turbo/apps/api/src/signals/services/pi-api-first-turn-cleanup.service.ts`: make cutoff constants and helper functions module-private

## Intentional exceptions

- The remaining Pi resource snapshot tests directly pin the byte-identical virtual-filesystem, ignore-rule, and precedence contract shared with the sandbox runtime. Production route output cannot expose this full snapshot corpus.
- Existing content-hash, workflow-event payload, and finite connector persisted-state contract exceptions are unchanged.

## Verification

- `pnpm exec prettier --write` on changed files
- `git diff --check`
- `pnpm --filter api run lint`
- `NODE_OPTIONS=--max-old-space-size=4096 pnpm --filter api run check-types`
- PostgreSQL 18 local database initialized and `DATABASE_URL=postgresql://postgres@localhost:5432/vm0_test pnpm -F @okouai/db db:migrate`
- `DATABASE_URL=postgresql://postgres@localhost:5432/vm0_test pnpm --filter api exec vitest run src/signals/services/__tests__/pi-resource-snapshot.service.test.ts` (5 passed)
[...]

PR #31129, opened 2026-09-02
fix(api): stop logging echoed provider request input

bingjiez666:
## Problem

`providerFailureDetailsForLog` wrote the user's generation request into production logs.

Fal's 422 responses are **Pydantic v2 validation entries**, where `msg`, `loc`, `type` and `input` are siblings of a single object — `input` being the offending request body. `stringifyProviderFailureDetail` ended with a `JSON.stringify` fallback for any object it didn't recognize, so that whole body was serialized into `fields.detail`, truncated only at 1000 characters.

Measured over Axiom `vm0-web-logs-prod`, `2026-08-03` → `2026-09-02`:

| | |
|---|---|
| Fal failure logs | 591 (all `type: image`) |
| carrying the echoed `input` | 591 (100%) |
| containing the user's `prompt` | 373 |
| containing a reference `image_url` | 319 |

BytePlus (4 failures) and MiniMax (2) leak nothing — their payloads don't echo `input`.

No `X-Amz-*` presigned parameters and no `data:image` payloads were found, so no credential-bearing material reached the logs.

## Why not just drop `detail` from the key list

That was the original suggestion in #30960. It does not work — the defect is the serializing fallback, not the key name. Any of the 17 keys resolving to an unrecognized object triggers it. Reproduced against the pure functions:

```
current main
  fields.detail:  {"input":{...,"prompt":"<user prompt>"}}       LEAKED
with "detail" removed from PROVIDER_FAILURE_DETAIL_KEYS
  fields.message: {"input":{...,"prompt":"<user prompt>"}}       LEAKED — same content, different field
```

## What changed

**Extraction is now allowlist-based.** Only scalars and arrays of scalars are emitted; an unrecognized object yields nothing. The `JSON.stringify` fallback is gone, recursion is bounded by an explicit depth limit, and emitted strings pass through the existing `redactPresignedUrls`.

**`msg` added to `PROVIDER_FAILURE_DETAIL_KEYS`.** This is load-bearing, not incidental. The list already carried five spellings of "message" (`message`, `error_message`, `errorMessage`, `err_msg`, `status_message`) but not the bare `msg` — which is exactly where Pydantic puts the reason. Removing the fallback without adding it degrades every one of these logs to:

```json
{ "error": "Unexpected status code: 422" }
```

i.e. precisely the unactionable state #30960 complains about. With both changes:

```json
{ "error": "Unexpected status code: 422",
  "detail": "Failed to download the file. Please check if the URL is accessible and try again." }
```

Diagnosability improves rather than regresses, and the two symptoms turn out to share one cause: `input` was serialized first and pushed the real `msg` past the 1000-character cut, which is why 146 of 591 records had no recoverable reason at all.

## Tests

Four cases added. The leak test asserts over the **whole** returned object across five different carrier keys, so a future change that merely relocates the leak to another log field cannot pass. The three pre-existing cases are unchanged and still pass.

## Scope

Deliberately not in this PR:

- Mapping Fal failures to a real user-facing message — Fal is still the only provider on the generic `failError("Generation failed")`. Tracked in #30960; the cause distribution needed to build that mapping is recorded there.
- The `*.sites.vm0.io` reference-image path, which was the largest single cause of these 422s (141 of 180 download failures, blocked by a zone-wide Cloudflare managed challenge). Already resolved by moving to `cdn.vm0.io`; last occurrence 2026-08-29.

Refs #30960

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/webhooks-built-in-generations.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/webhooks-built-in-generations.test.ts
@@ -48,4 +48,53 @@ describe("providerFailureDetailsForLog", () => {
       statusMessage: "no worker available",
     });
   });
+
+  it("reports the Fal validation reason without the echoed request input", () => {
+    const details = providerFailureDetailsForLog({
+      status: "ERROR",
+      error: "Unexpected status code: 422",
+      detail: {
+        input: {
+          prompt: "a portrait of Sarah outside the office on Howard St",
+          image_urls: ["https://example-ref.sites.vm0.io/reference.png"],
+          image_size: { height: 1024, width: 1024 },
+        },
+        loc: ["body", "image_urls", 0],
+        msg: "Failed to download the file. Please check if the URL is accessible and try again.",
+        type: "value_error",
+      },
+    });
+    expect(details).toStrictEqual({
+      error: "Unexpected status code: 422",
+      detail:
+        "Failed to download the file. Please check if the URL is accessible and try again.",
+    });
+  });
+
+  it("never serializes an unrecognized object, whichever key carries it", () => {
+    const echoed = {
+      input: { prompt: "a portrait of Sarah outside the office on Howard St" },
+    };
+    for (const key of ["detail", "message", "error", "reason", "description"]) {
+      const details = providerFailureDetailsForLog({
+        status: "ERROR",
+        [key]: echoed,
+      });
+      // Assert over the whole result: a fix that only moves the leak to a
+      // different log field must not pass.
+      expect(JSON.stringify(details)).not.toContain("Sarah");
+    }
+  });
+
+  it("redacts presigned URLs left inside a provider message", () => {
+    expect(
+      providerFailureDetailsForLog({
+        status: "ERROR",
+        message:
+          "could not read https://bucket.s3.amazonaws.com/a.png?X-Amz-Signature=deadbeef",
+      }),
+    ).toStrictEqual({
+      message: "could not read [redacted presigned URL]",
+    });
+  });
 });
--- a/turbo/apps/api/src/signals/routes/webhooks-built-in-generations.ts
+++ b/turbo/apps/api/src/signals/routes/webhooks-built-in-generations.ts
@@ -10,7 +10,7 @@ import { request$ } from "../context/hono";
 import { waitUntil } from "../context/wait-until";
 import { pathParamsOf, queryOf } from "../context/request";
 import type { RouteEntry } from "../route-entry";
-import { safeJsonParse, safeSync, tapError } from "../utils";
+import { safeJsonParse, tapError } from "../utils";
 import {
   downloadFalImage,
   getFalImageBillableUnits,
@@ -56,6 +56,7 @@ import {
 } from "../services/video-generation.service";
 import { env } from "../../lib/env";
 import { logger } from "../../lib/log";
+import { redactPresignedUrls } from "../../lib/presigned-url-redaction";
 import {
   avatarVideoPricing$,
   downloadJoggAiAvatarVideo,
@@ -278,6 +279,7 @@ const PROVIDER_FAILURE_DETAIL_KEYS = [
   "error_message",
   "errorMessage",
   "message",
+  "msg",
   "detail",
   "description",
   "status_message",
@@ -300,6 +302,9 @@ function providerFailureLogKey(key: string): string {
     case "err_msg": {
       return "errorMessage";
     }
+    case "msg": {
+      return "message";
+    }
     case "status_message": {
       return "statusMessage";
     }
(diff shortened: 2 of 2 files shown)


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 14aef374417d7744fd2af270b71f7d8a52cef084
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Replaces the blocklist-plus-`JSON.stringify` extraction in `stringifyProviderFailureDetail` with allowlist extraction that emits only scalars and arrays of scalars, and adds `msg` to `PROVIDER_FAILURE_DETAIL_KEYS`. The two changes are coupled: `msg` is where Pydantic v2 puts the reason, so dropping the fallback alone would have reduced every Fal failure log to `error: Unexpected status code: 422`. Removal evidence for the fallback is production data (591 Fal failure logs, all carrying the echoed `input`), which satisfies `docs/fallback.md` §10.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no wrong assumptions or unhandled edge cases | Pass | Key order keeps `message` ahead of `msg`, so the pre-existing nested-`error.message` case is unaffected; `redactPresignedUrls` runs before truncation, so a redacted URL cannot be split by the 1000-char cut. |
| `REVIEW.md` | Security — no secrets or sensitive data exposure | Pass | This is the fix: `webhooks-built-in-generations.ts:363-373` stops serializing the provider's echoed request, which was placing user prompts and reference image URLs in `vm0-web-logs-prod`. |
| `REVIEW.md` | Type safety — no `any`, no overly broad types | Pass | Both new functions take `unknown` and narrow; no casts added. |
| `REVIEW.md` | Interface changes documented | Pass (note) | `scalarProviderFailureDetail` now `trim()`s strings, so a whitespace-only provider field yields `undefined` instead of whitespace. Behaviour improvement, but not called out in the PR summary. |
| `REVIEW.md` | `fix:` commits must carry a regression test | Pass | Three cases added in `__tests__/webhooks-built-in-generations.test.ts:51-100`, including one that reproduces the exact leaked shape. |
| `docs/bad-smell.md` | Zero `any`, zero lint/type suppressions | Pass | Neither appears in the diff. |
| `docs/bad-smell.md` | No dynamic imports; no hardcoded URLs/config | Pass | Only static imports; `redactPresignedUrls` imported from `../../lib/presigned-url-redaction`. |
| `docs/bad-smell.md` | Comments explain WHY, not WHAT | Pass | `:369-371` records why the fallback must not come back, which is the failure mode a future reader would otherwise re-introduce. |
| `docs/bad-smell.md` | Fail fast — no fallback/recovery logic | Pass | The diff deletes a fallback. `MAX_PROVIDER_FAILURE_DETAIL_DEPTH` is input bounding on an untrusted webhook body, not a compatibility path — see the note below the table. |
| `docs/fallback.md` §9 | Every fallback the diff introduces is declared in the PR summary | N/A | The diff introduces no fallback, so no `Fallbacks` section is required and none is expected. |
| `docs/fallback.md` §10 | Removing a fallback requires type, single-writer, production, or rollout evidence | Pass | Production evidence in the PR body: 591/591 Fal failure logs carried the echoed `input`. Independently checked that removal does not regress the other providers — all 4 BytePlus and 2 MiniMax failure logs carry plain-text `fields.error`, none produced by the serializing fallback. |
| `docs/fallback.md` §1 | No negative tests asserting removed behavior stays removed | Pass | The leak test asserts a live product guarantee (no prompt in log output), not the absence of deleted code. |
| `docs/testing.md` | Mock at boundary only; no `vi.mock()` on relative paths | Pass | The diff adds no mocks. |
| `docs/testing.md` | No fake timers; no assert-on-mock-calls; mock cleanup | Pass | None used. |
| `docs/testing/patterns.md` | Pattern 2 — pure-function unit tests only for security-critical or algorithmically complex code | Pass | Redaction of user prompts from logs is a data-protection boundary and the function is pure, which is the sanctioned exception. |
| `docs/testing/anti-patterns.md` | AP-7 / AP-11 — do not pin implementation details; test through the route when one exists | Finding (P1) | See P1 below: the route exists but its log output has no external surface. |
| `docs/testing/testing-external-behavior.md` | In `apps/api` the only trusted boundary is the API endpoint | Finding (P1) | Tests import `providerFailureDetailsForLog` directly; that symbol is `export`ed solely for tests (its three production call sites are in the same module). |
| `docs/testing/api-testing.md` | Route behavior covered via `setupApp()` + ts-rest contract | N/A | No route surface, contract, or response shape changed; the webhook still returns `okResponse()` on both paths. |
| `.claude/skills/ccstate/SKILL.md` | ccstate command/computed practices for `apps/api` | N/A | Only module-private pure helpers changed; no `command`, `computed`, or Store usage touched. |

Note on `MAX_PROVIDER_FAILURE_DETAIL_DEPTH`: I did not classify this as a `docs/fallback.md` fallback. It guards against unbounded recursion on an untrusted webhook body rather than bridging two deployed versions, so none of the §7 removal gates apply. Worth recording that it is the one change in this PR with no production incident behind it: measured locally, the pre-change code already threw `RangeError: Maximum call stack size exceeded` on a ~1MB, 100k-deep body, so this closes a pre-existing latent crash rather than one the diff introduces. Reachability is low — the token check at `:855-869` runs before `request.text()` at `:871`.

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

- `turbo/apps/api/src/signals/routes/__tests__/webhooks-built-in-generations.test.ts:51-100` — the three new cases call `providerFailureDetailsForLog` directly, and that symbol is exported only so tests can reach it. `docs/testing/testing-external-behavior.md` states that the API endpoint is the only trusted boundary, and PR #28736 set the precedent of deleting such assertions and making the symbols module-private.

  Raising this as the gap rather than as a required change, per `docs/testing/anti-patterns.md` AP-11 ("If behavior is not reachable through the production API surface... raise the gap during review"). The gap is real: the corrected behaviour is *what gets written to the log*, and the webhook returns `okResponse()` on both the leaking and the fixed path, so no endpoint observation can distinguish them. There is no log-assertion helper anywhere in `turbo/apps/api` today, and adding one at the endpoint would require `vi.mock("../../lib/log")` — a relative-path internal mock, which is a **P0** violation of the mock boundary. Given that, the direct pure-function test is the better of the two available options and is covered by `docs/testing/patterns.md` Pattern 2, which is why the verdict stays `LGTM`.

  Not introduced by this PR — the three pre-existing cases in the same file already use this shape — but it is extended by it. A follow-up that adds a log-capture surface to the API test context would let all six cases move to the endpoint and let the `export` be dropped.

### Testing

- Coverage: Adequate. `fix:` regression coverage is present and targets the reported defect directly. The leak case is the strongest of the three: it iterates five different carrier keys and asserts against `JSON.stringify(details)` as a whole, so a future change that merely relocates the leak from `detail` to `message` fails the test. That matters here specifically, because relocating the leak is exactly what the originally proposed fix in #30960 would have done.
- Conventions: Compliant. No mocks, no fake timers, no assertions on mock calls, no DB access; `toStrictEqual` used for exact-shape assertions.
- Verified independently: all three pre-existing cases still pass against the edited source, so the `msg` addition does not change existing extraction.
- Minor nit, non-blocking: the five-key loop at `:36-44` will not name the offending key on failure. `it.each` would make a regression self-describing.
- Residual behavior change worth knowing after deploy: for the subset of Fal failures whose reason is not reachable under any allowlisted key, the log now carries only `error` instead of a serialized blob. Measured at 146 of 591 records historically — but those had no recoverable reason before either, because `input` was serialized first and pushed the reason past the 1000-char truncation, so this subset should shrink rather than grow.

PR #31131, opened 2026-09-02
fix(python): stop starred tails changing fixed first arguments

### LINKED ISSUE #31056: bug(python): stop starred-tail uncertainty from changing fixed first arguments
## Summary

The flow-metadata linter can report a registered metadata-key violation for a later positional argument even when a deterministic element before a nested starred expansion has already fixed the call's first positional argument.

## Location

- `crates/runner/mitm-addon/scripts/flow_metadata_linter/ast_helpers.py:47-59`
- `crates/runner/mitm-addon/scripts/flow_metadata_linter/ast_helpers.py:109-130`

`_static_iterable_first_outcomes()` unconditionally propagates `element_has_unknown_expansion` at line 129. For a sequence such as `["external_key", *args]`, the first outcome is already fixed before `*args` is reached, so uncertainty in that tail cannot make the sequence empty or change its first element. The propagated flag nevertheless makes `_static_first_call_argument_nodes()` continue to later outer call arguments.

## Reproduction

```python
args = ()
value = flow.metadata.get(*["external_key", *args], "sandbox_run_id")
```

Running `_MetadataKeyVisitor` on this source reports:

```text
fixed_first_starred_tail.py:2: use metadata_keys.SANDBOX_RUN_ID for flow.metadata access
```

The call executes successfully with `args = ()`, and `"sandbox_run_id"` is the default value, not the lookup key. More generally, the nested starred tail can never move that later argument ahead of the deterministic `"external_key"` prefix.

## Impact

The repository metadata-key check can fail CI on valid code whenever:

1. a starred list or tuple has a deterministic first element,
2. a dynamic starred tail follows it, and
3. a later positional argument happens to equal a registered metadata key.

This is a false positive in a mandatory source linter, not merely imprecise internal state: the reported literal cannot occupy the first positional argument on any execution.

## Acceptance criteria

- Stop propagating starred-expansion uncertainty once every possible first outcome is already a concrete node.
- Keep conservative handling when a dynamic expansion occurs before the first deterministic element and may be empty.
- Add regression coverage showing that a registered literal after `*["external_key", *args]` is not checked as the first argument.
- Preserve the existing `dynamic_star_args` cases where an unknown leading expansion can genuinely expose a later registered literal as the first argument.

seven332:
## Summary

- stop nested list and tuple analysis once every possible first argument is fixed
- keep conservative first-argument handling for dynamic leading expansions
- cover the fixed-prefix case through the public flow-metadata linter API

## Exclusions

- no metadata registry, addon runtime, API, schema, or deployment behavior changes

## Validation

- `uv run --no-sync python -m pytest tests/test_flow_metadata_key_contract.py`
- `uv lock --check`
- `uv sync --locked`

Closes #31056

### DIFF
--- a/crates/runner/mitm-addon/scripts/flow_metadata_linter/ast_helpers.py
+++ b/crates/runner/mitm-addon/scripts/flow_metadata_linter/ast_helpers.py
@@ -80,10 +80,11 @@ def _deduplicate_static_argument_outcomes(
 
 
 def _static_iterable_first_outcomes(node: ast.AST) -> tuple[list[ast.AST | None], bool]:
-    """Return ordered first-node/empty outcomes and whether any expansion is unknown.
+    """Return ordered first-node/empty outcomes and whether a first outcome is unknown.
 
     ``None`` represents a statically empty expansion. The unknown flag remains separate because
-    the linter conservatively treats any dynamic nested expansion as a possible empty call prefix.
+    the linter conservatively treats a dynamic expansion before a fixed first node as a possible
+    empty call prefix.
     """
     if isinstance(node, ast.NamedExpr):
         return _static_iterable_first_outcomes(node.value)
@@ -109,6 +110,8 @@ def _static_iterable_first_outcomes(node: ast.AST) -> tuple[list[ast.AST | None]
     outcomes: list[ast.AST | None] = [None]
     has_unknown_expansion = False
     for element in node.elts:
+        if not any(outcome is None for outcome in outcomes):
+            break
         if isinstance(element, ast.Starred):
             element_outcomes, element_has_unknown_expansion = _static_iterable_first_outcomes(
                 element.value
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #31131

**Reviewed head:** `e0c3ff9f651fc405b1c5a7ea45be9fefb8f9440f`

### Summary

The change stops irrelevant starred-tail uncertainty at the point where a static list or tuple already has a fixed first outcome. It retains uncertainty accumulated before the first deterministic element and leaves shared caller interfaces unchanged.

### Findings

No findings.

### Testing

The regression exercises the public linter API with a real temporary source file. The focused contract test passed with 23 tests, and the full mitm-addon package suite completed successfully with 7,334 tests collected. Lockfile verification, environment sync, Ruff format/lint, basedpyright, and the flow-metadata CLI check also passed. Remote CI remains a separate readiness gate.

### Verdict

LGTM

PR #31159, opened 2026-09-02
fix(cli): render the camera cut at the frame rate the recording declares

bingjiez666:
## Summary

`okou video camera` took its output frame rate from ffprobe's `avg_frame_rate` (then `r_frame_rate`) and only fell back to the sidecar's `frameRate` when the probed value was missing or above 240. That guard never fires for the file it was meant to catch.

A desktop screen capture only stores a frame when the picture changes, so a window or area recording of a mostly still screen is a sparse variable-rate MP4. ffprobe averages it to a valid-looking rate well under one frame per second. Reproduced with an 8 s file holding 5 frames:

| ffprobe field | value |
|---|---|
| `avg_frame_rate` | `150/241` (≈0.62 fps) |
| `r_frame_rate` | `1/2` (0.5 fps) |

Both pass the `> 0 && <= 240` check, so the cut was rendered with `fps=0.62` and the camera motion was sampled on that same sparse grid: a slideshow with jumping zooms. Click timing was not affected (the pipeline aligns by time), only playback quality.

## Change

- The frame rate is never probed anymore. The recording sidecar and the camera plan both declare the capture rate, and the cut always plays back at that rate. ffprobe still supplies width, height, and duration.
- `assertPlanMatchesVideo` no longer compares the plan's frame rate with the probed one, since that would now compare the plan to itself.
- Regression test: with the fake ffprobe reporting `150/241` / `1/2`, the plan, the JSON result, and the ffmpeg `-vf` chain all stay at 30 fps. On the previous code this test fails with `expected 0.6224066390041494 to be 30`.

## Test plan

- [x] `pnpm -F @okouai/cli exec vitest run src/commands/video/__tests__/camera.test.ts` (4 passed; new test verified failing on the old code)
- [x] `pnpm -F @okouai/cli lint`, `tsc --noEmit`, `pnpm knip --workspace apps/cli`
- [ ] CI

### DIFF
--- a/turbo/apps/cli/src/commands/video/__tests__/camera.test.ts
+++ b/turbo/apps/cli/src/commands/video/__tests__/camera.test.ts
@@ -12,6 +12,11 @@ import { join } from "node:path";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { cameraCommand } from "../camera";
 
+/** What the fake ffprobe says about the video stream, adjustable per test. */
+const probedStream = vi.hoisted(() => {
+  return { avg_frame_rate: "30/1", r_frame_rate: "30/1" };
+});
+
 vi.mock("child_process", async () => {
   const { readFileSync, writeFileSync } =
     await vi.importActual<typeof import("node:fs")>("node:fs");
@@ -19,14 +24,7 @@ vi.mock("child_process", async () => {
     execFileSync: vi.fn((command: string, args: readonly string[]) => {
       if (command === "ffprobe") {
         return JSON.stringify({
-          streams: [
-            {
-              width: 1920,
-              height: 1080,
-              avg_frame_rate: "30/1",
-              r_frame_rate: "30/1",
-            },
-          ],
+          streams: [{ width: 1920, height: 1080, ...probedStream }],
           format: { duration: "10.000000" },
         });
       }
@@ -66,6 +64,8 @@ describe("okou video camera command", () => {
 
   afterEach(() => {
     vi.clearAllMocks();
+    probedStream.avg_frame_rate = "30/1";
+    probedStream.r_frame_rate = "30/1";
     rmSync(directory, { recursive: true, force: true });
   });
 
@@ -177,6 +177,55 @@ describe("okou video camera command", () => {
         "-c:v",
         "libx264",
       ]),
+    );
+  });
+
+  it("renders at the frame rate the recording declares, not the probed average", async () => {
+    // A screen capture only stores a frame when the picture changes, so
+    // ffprobe averages a mostly still recording to well under one frame per
+    // second. That figure is valid as far as ffprobe is concerned, and it
+    // used to become the output frame rate.
+    probedStream.avg_frame_rate = "150/241";
+    probedStream.r_frame_rate = "1/2";
+    const videoPath = join(directory, "recording.mp4");
+    const eventsPath = join(directory, "recording.clicks.json");
+    const outputPath = join(directory, "draft.mp4");
+    writeFileSync(videoPath, Buffer.from("source-video"));
+    writeFileSync(
+      eventsPath,
+      JSON.stringify({
+        version: 1,
+        recording: {
+          durationMs: 10_000,
+          video: { width: 1920, height: 1080, frameRate: 30 },
+        },
+        clicks: [{ tMs: 2_000, normalized: { x: 0.12, y: 0.22 } }],
+        pointerEvents: [
+          { tMs: 1_700, kind: "move", normalized: { x: 0.1, y: 0.2 } },
+          { tMs: 2_000, kind: "click", normalized: { x: 0.12, y: 0.22 } },
+        ],
+      }),
+    );
+
+    await cameraCommand.parseAsync(
+      ["--file", videoPath, "--events", eventsPath, "--output", outputPath],
+      { from: "user" },
+    );
+
+    const result = JSON.parse(stdout()) as {
+      planPath: string;
+      frameRate: number;
+    };
+    expect(result.frameRate).toBe(30);
+    const plan = JSON.parse(readFileSync(result.planPath, "utf8")) as {
+      source: { frameRate: number };
+    };
+    expect(plan.source.frameRate).toBe(30);
+    const ffmpegCall = vi.mocked(execFileSync).mock.calls.find((call) => {
+      return call[0] === "ffmpeg";
+    });
+    expect(ffmpegCall?.[1]).toEqual(
+      expect.arrayContaining(["-vf", expect.stringMatching(/^fps=30,/u)]),
     );
   });
 
--- a/turbo/apps/cli/src/commands/video/camera.ts
+++ b/turbo/apps/cli/src/commands/video/camera.ts
@@ -34,8 +34,6 @@ const ffprobeSchema = z.object({
       z.object({
         width: z.number().int().positive(),
         height: z.number().int().positive(),
-        avg_frame_rate: z.string(),
-        r_frame_rate: z.string(),
       }),
     )
     .min(1),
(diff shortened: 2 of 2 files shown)


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 9f3f16e5c90b873d37f2369e3cb82d0def4873c0
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

`okou video camera` stops probing the frame rate: `probeVideo` now returns the rate declared by the recording sidecar (`--events`) or by the camera plan (`--plan`), keeps width/height from `ffprobe`, and keeps the pre-existing duration fallback. Because the probed rate is gone, the tautological frame-rate comparison in `assertPlanMatchesVideo` is removed, and a regression test pins 30 fps end to end (JSON result, written plan, and the ffmpeg `-vf fps=30,` chain) while the fake `ffprobe` reports `150/241` / `1/2`.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness & logic; no wrong assumptions or unhandled edge cases | Pass | `camera.ts:probeVideo` — the removed `<= 240` guard's role is fully retained by schema validation: both sidecar variants require `frameRate: z.number().positive()` and every plan goes through `cameraPlanSchema` (`frameRate: z.number().positive().max(240)`) — `createCameraPlan` returns `cameraPlanSchema.parse(...)` on the `--events` path and the `--plan` path parses the file with the same schema. An absurd or absent declared rate still fails loudly instead of reaching `fps=` or `createCameraFrameStates`. |
| `REVIEW.md` | Removing a validation must not lose real coverage | Pass | `camera.ts:assertPlanMatchesVideo` — on the `--plan` path `source` is now `probeVideo(inputPath, plan.source)`, so the deleted check compared `plan.source.frameRate` with itself. The dimension and duration checks, which still compare against probed values, are untouched. |
| `REVIEW.md` | Security (injection, secrets, auth) | N/A | Local CLI change; no new external input, credential, endpoint, or shell string. The frame rate reaches ffmpeg through the existing `execFileSync` argument array, bounded by the plan schema. |
| `REVIEW.md` | Type safety, no `any` | Pass | `camera.ts`, `camera.test.ts` — no `any`; the narrowed `ffprobeSchema` keeps `width`/`height` typed, and the test's `JSON.parse` casts match the existing file's style. |
| `REVIEW.md` | Style & maintainability; comments explain WHY | Pass | The rewritten `probeVideo` doc comment explains why the probe is untrustworthy for sparse screen captures rather than restating the code; the dead `frameRate()` parser is deleted with its only caller. |
| `REVIEW.md` | Interface/behavior changes documented | Pass | Behavior change (output rate source) and its rationale are in the PR summary with the reproduced `avg_frame_rate` / `r_frame_rate` values; no docs reference frame-rate detection (`git grep "frame rate\|frameRate" docs` is empty) and the command help text does not mention it. |
| `REVIEW.md` | DB/JSONB and persisted-contract gate | N/A | No `packages/db` schema, contract, or `jsonb(...)` change. The only persisted artifact is the local camera plan file; `cameraPlanSchema` is unchanged and validation is only loosened, so plans written by an older CLI still parse and render. |
| `REVIEW.md` | Feature regression guard — changed existing behavior needs regression coverage | Pass | The existing behavior being replaced is exactly what the new test pins, and the four pre-existing tests (auto first cut, VM0 demo capture format, AI-edited plan, plus the new case) still cover the unchanged paths with `30/1` probe output. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions, no dynamic imports, no hardcoded config | Pass | Neither changed file adds a suppression comment, `await import(...)`, or hardcoded URL/value. |
| `docs/bad-smell.md` | Fail fast instead of silently recovering | Pass | This PR moves in the fail-fast direction: the silent "probed rate looked valid, so use it" path is gone, and an out-of-range declared rate now surfaces as a schema error. |
| `docs/fallback.md` | Do not keep fallbacks the contract already prevents; removal needs evidence | Pass | The removed probe-then-fallback chain is deleted together with its helper and the assertion that depended on it. Type/schema evidence for the replacement is the required `frameRate` on both sidecar variants and on `cameraPlanSchema`. |
| `docs/fallback.md` | Every fallback introduced by the diff is declared in the PR summary | N/A | The diff introduces no fallback. The `format.duration` fallback is pre-existing and only re-worded; per §9 a PR that merely keeps or removes fallbacks needs no `Fallbacks` section. |
| `docs/fallback.md` | No negative/tombstone tests asserting removed behavior stays removed | Pass | The new test asserts current product behavior — the cut plays back at the declared rate — rather than the absence of the old code path, so it fails for a real reason if probing ever returns. |
| `docs/testing.md` | `fix:` commits require a regression test | Pass | `camera.test.ts:183` reproduces the sparse-capture probe (`150/241`, `1/2`) and is stated to fail on the old code with `expected 0.6224066390041494 to be 30`. |
| `docs/testing.md` | Integration tests, not unit tests for internals | Pass | The test enters through `cameraCommand.parseAsync()` and observes stdout JSON, the written plan file, and the ffmpeg invocation; the deleted `frameRate()` helper never got a unit test. |
| `docs/testing.md` / `docs/testing/anti-patterns.md` (AP-4) | Mock at the boundary only — no relative-path `vi.mock()` | Pass | Only `vi.mock("child_process")` — the Node built-in listed as "sometimes" mockable; no internal CLI module is mocked. |
| `docs/testing.md` / `docs/testing/anti-patterns.md` (AP-1) | Test behavior, not mock call counts | Pass | Assertions are on `result.frameRate`, the persisted `plan.source.frameRate`, and the actual `-vf` argument string; no bare `toHaveBeenCalled()`. |
| `docs/testing.md` / `docs/testing/anti-patterns.md` (AP-3) | Real filesystem under a temp directory | Pass | The test writes a real `recording.mp4`/`recording.clicks.json` into the suite's temp `directory` and reads the generated plan back with `readFileSync`. |
| `docs/testing.md` / `docs/testing/anti-patterns.md` (AP-5) | No fake timers | Pass | No `vi.useFakeTimers()` / `advanceTimersByTime` added. |
| `docs/testing.md` | Mock cleanup in teardown | Pass | `afterEach` keeps `vi.clearAllMocks()` and additionally restores `probedStream` to `30/1`, so the shared hoisted fixture cannot leak into the other cases. |
| `docs/testing/patterns.md` | Use the established pattern for the surface; no MSW/DB machinery where it does not apply | Pass | Follows the CLI command-integration pattern (Pattern 4-style real-filesystem setup); the command makes no HTTP call, so MSW is correctly absent. |
| `docs/testing/cli-testing.md` | CLI tests enter through Commander, keep parsing/validation/filesystem real, live in a neighboring `__tests__/` | Pass | `turbo/apps/cli/src/commands/video/__tests__/camera.test.ts` sits beside `camera.ts` and drives the real command; no internal CLI module, validator, or serializer is stubbed. |
| `docs/testing/cli-testing.md` | No legacy-token or other tombstone assertions | Pass | The diff touches no auth/env surface and adds no legacy-name assertion. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

Non-blocking observation (no change requested): the fake `ffprobe` still returns `avg_frame_rate`/`r_frame_rate`, which real `ffprobe` will no longer emit under the narrowed `-show_entries stream=width,height:format=duration`. That divergence is deliberate and load-bearing here — it is what lets the test prove the values are ignored even when present — and `ffprobeSchema` strips them, so nothing depends on the extra keys.

### Testing

- Coverage: Adequate. A `fix:` commit with a regression test that pins the reported rate, the persisted plan, and the rendered ffmpeg filter at 30 fps against the exact probe values that caused the bug, verified failing on the old code.
- Conventions: Compliant. Boundary-only `child_process` mock, real temp-directory filesystem, behavior assertions, no fake timers, mock and fixture state reset in `afterEach`.

PR #31189, opened 2026-09-02
test(runner): assert durable balloon relief

### LINKED ISSUE #31187: test(runner): avoid transient balloon target sampling in lane a
## Background

PR #31132 moved the full-relief regression into the bounded lane A balloon scenario so ordinary pull-request checks no longer allocate approximately 3.6–3.8 GiB per run. The consolidated scenario caps deliberate Guest pressure at 512 MiB and attempts to observe both a pressure sample and the controller's transient target-zero request.

## Problem

The real-metal assertion can fail after the balloon controller has behaved correctly because it requires an external polling loop to sample short-lived controller inputs and target state.

In [Actions job 100225565100](https://github.com/vm0-ai/vm0/actions/runs/33622956530/job/100225565100), lane A reported:

```text
FAIL: guest memory pressure was not established: available=1666MiB boundary=192MiB actual=2048MiB
```

The runner journal on `dev-11.gcp.vm3.ai` proves that the intended production behavior completed:

```text
2026-09-02T11:14:57.411775Z balloon deflate current=3328 new_target=0 available_mib=130
2026-09-02T11:15:02.412086Z balloon inflate current=0 new_target=256 free_mib=3405
```

The test sampled `/balloon/statistics` at `11:14:52.926`, then its next statistics request was not served until `11:15:03.624`. It therefore missed the complete target-zero interval and resumed after normal reinflation had started. A passing run sampled the same transient interval and observed `target=0`, so the result currently depends on request timing rather than controller correctness.

This is a false-negative CI failure. It is unrelated to the runner polling changes in PR #31116 whose merge-group run exposed it.

## Requirements

- Keep the deliberate Guest pressure allocation capped at 512 MiB.
- Do not change the production balloon controller, thresholds, cadence, or recovery behavior.
- Do not make the real-metal result depend on sampling the transient `target=0` or `available_memory < 192 MiB` interval.
- Preserve a regression boundary that distinguishes full relief from the retired partial policy. The retired policy could request at most a 256 MiB decrement from the stabilized state; the real-metal scenario should verify durable physical relief beyond that bound.
- Keep the existing recovery, VM-survival, allocator cleanup, and service cleanup assertions.
- Keep the Firecracker API-backed Rust tests as the direct verification that a low-available-memory controller tick requests target zero.

## Acceptance criteria

- A run that starts from a stable target/actual balloon above 256 MiB and physically releases more than the retired policy's maximum 256 MiB decrement passes even if external polling misses the target-zero interval.
- The scenario still fails when physical relief does not exceed the retired partial-policy bound within the existing deadline.
- Shell syntax and the durable-runner staging/transport fixture pass locally.
- The existing Firecracker API-backed Rust tests continue to verify the exact target-zero request, and current-head real-metal lane A validates the physical-relief outcome.
- No timeout increase, retry-only workaround, production behavior change, or high-memory standalone lane is introduced.

## Related work

- #31119
- #30221
- PR #31132
- PR #30227

seven332:
## Summary

- replace transient balloon pressure/target sampling gates with a durable physical-relief assertion
- require `actual_mib` to drop more than the retired controller's 256 MiB partial-relief bound
- initialize recovery from the target and minimum physical size actually observed by the test

The production balloon controller, polling cadence, timeout, 512 MiB pressure cap, and recovery behavior are unchanged.

## Validation

- `bash -n .github/scripts/runner-behavior-balloon-remote.sh`
- `bash .github/scripts/tests/runner-behavior-durable-test.sh`
- `cargo test --manifest-path crates/Cargo.toml -p sandbox-fc balloon --lib` (61 passed)
- `git diff --check`

Closes #31187

### DIFF
--- a/.github/scripts/runner-behavior-balloon-remote.sh
+++ b/.github/scripts/runner-behavior-balloon-remote.sh
@@ -82,12 +82,11 @@ API_SOCK="/run/vm0/sock/$SANDBOX_ID/api.sock"
 
 # Keep these expectations aligned with balloon.rs: the controller polls every
 # 5 seconds, deflates below 192 MiB available, and inflates above 384 MiB free.
-PRESSURE_AVAILABLE_BOUNDARY_MIB=192
 PRESSURE_TARGET_AVAILABLE_MIB=128
 MAX_PRESSURE_ALLOC_MIB=512
-# The retired controller subtracted at most its 256 MiB free-memory target
-# deficit. Starting above this size guarantees that policy leaves a nonzero
-# partial target, so the target-zero assertion distinguishes the regression.
+# The retired controller physically relieved at most its 256 MiB free-memory
+# deficit. A larger physical drop distinguishes full relief without requiring
+# an external poll to observe the transient target-zero interval.
 RETIRED_PARTIAL_RELIEF_MAX_MIB=256
 DEFLATE_POLL_SECONDS=2
 DEFLATE_TIMEOUT_SECONDS=60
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #31189

**Reviewed head:** `6f790309810cb88fdbc252c0d354e990b89524a6`

### Summary

The real-metal balloon regression now asserts a durable physical outcome instead of requiring the polling client to catch the controller's transient target-zero interval. The strict greater-than-256 MiB relief boundary still distinguishes the retired partial policy, and recovery starts from the state actually observed.

### Findings

No findings.

### Testing

- Shell syntax passed.
- Durable remote-worker staging/transport fixture passed.
- Focused `sandbox-fc` balloon tests passed (61 tests), retaining direct target-zero policy coverage.
- The 512 MiB cap, polling intervals, deadlines, VM-survival checks, recovery, and cleanup remain unchanged.
- Current-head real-metal CI is the remaining end-to-end verification boundary.

### Verdict

LGTM

PR #31322, opened 2026-09-03
fix(annotation): read the original image as bytes before flattening

tongx-blip:
## Problem

Tong: *"我只要annotate的功能标注图片，并attach mark就会出现这个问题"* — every annotated screenshot came back with the red retry badge on its chip, so the annotation could never be attached.

The badge is `AnnotationUploadRetry`, which renders when `annotationUploadStatus$ === "failed"`. That is the **annotated copy**, not the original upload — a failed original upload removes the chip entirely and toasts.

## Cause

Flattening reached for the original with an ``. That made it the only step in the feature that needed the CDN's permission: the annotation editor renders the very same URL as a plain `` (`image-annotation-editor.tsx` → `useResolvedAttachmentUrl`), and a plain `` negotiates nothing.

So the image a user had just finished drawing on could still be refused at flatten time — and an image element reports that as a bare `error` event with no status, no headers and no reason.

## Fix

- **Read the bytes with `fetch`, decode through an object URL.** A `blob:` URL is same-origin by definition, so the canvas is never tainted and `toBlob` stays readable without asking the CDN for anything. A refusal now arrives with its HTTP status attached.
- **Stop discarding the error.** Five steps land on the one `failed` state — the original going missing, the read, the decode, the encode, the upload — and `rendered.error` was dropped, so the badge was the only evidence any of them happened. Logged as a *warning*: the marks survive on the chip and the badge retries in place, so this is a recoverable condition the user is already being shown.

## Why this was never caught

`window.Image` is stubbed in tests to report `load` for **any** src, including one the browser would have refused, and the canvas is fully mocked. The whole "can this app actually read these pixels" question had no coverage, which is how a failure that reproduced on every annotated screenshot stayed green.

Two new cases close that. Verified both ways — they **fail** against the old `crossOrigin` loader and **pass** against this one:

```
× attaches marks when the original image is only readable as bytes
× offers a retry when the original image cannot be read
Tests  2 failed | 15 passed (17)     ← old loader
Tests  17 passed (17)                ← this branch
```

## Checks

prettier clean · platform `check-types` clean · eslint clean on the changed files · `image-annotation.test.tsx` 17/17 · `chat-draft` + `attachment-chips` 80/80.

`knip` and repo-wide `check-types` were excluded locally in this 3.9 GB sandbox (`LEFTHOOK_EXCLUDE=knip,check-types`, not `--no-verify`); CI runs both.

### DIFF
--- a/turbo/apps/platform/src/signals/okou-page/chat-draft.ts
+++ b/turbo/apps/platform/src/signals/okou-page/chat-draft.ts
@@ -28,6 +28,7 @@ import { toast } from "@okouai/ui/components/ui/sonner";
 import type { EditorDocumentSnapshot } from "./user-message-document-codec.ts";
 import { i18n } from "../../i18n/index.ts";
 import { flattenAnnotatedImage } from "./flatten-annotated-image.ts";
+import { logger } from "../log.ts";
 import { isAnnotationMeaningful } from "./image-annotation.ts";
 import { desktopRecordingAgentInstructions } from "./intro-video-agent-instructions.ts";
 
@@ -57,6 +58,8 @@ type AttachmentUploadState =
       readonly status: "uploaded";
       readonly fileInfo: FileInfo;
     };
+
+const log = logger("chat-draft");
 
 const MULTIPART_UPLOAD_THRESHOLD_BYTES = 5 * 1024 * 1024;
 const MAX_PART_UPLOAD_ATTEMPTS = 5;
@@ -411,6 +414,14 @@ function createAttachmentAnnotationSignals(args: {
       );
       if (!rendered.ok) {
         set(internalUploadState$, { status: "failed" });
+        // Five different steps land on that one state — the original going
+        // missing, the read, the decode, the encode and the upload. Dropping
+        // `rendered.error` left the retry badge as the only evidence any of
+        // them had happened, so a failure that reproduced every time was
+        // indistinguishable from a flaky one. Warn rather than error: the
+        // attachment keeps its marks and the badge retries in place, so this
+        // is a recoverable condition the user is already being shown.
+        log.warn("annotation flatten failed", args.filename, rendered.error);
         return;
       }
       set(internalUploadState$, {
--- a/turbo/apps/platform/src/signals/okou-page/chat-draft.ts
+++ b/turbo/apps/platform/src/signals/okou-page/chat-draft.ts
@@ -29,6 +29,8 @@ import type { EditorDocumentSnapshot } from "./user-message-document-codec.ts";
 import { i18n } from "../../i18n/index.ts";
 import { flattenAnnotatedImage } from "./flatten-annotated-image.ts";
 import { logger } from "../log.ts";
+import { pageAttachmentResourceUrlResolver$ } from "../attachment-resource-url.ts";
+import { publicAttachmentUrl } from "../../views/okou-page/attachment-url.ts";
 import { isAnnotationMeaningful } from "./image-annotation.ts";
 import { desktopRecordingAgentInstructions } from "./intro-video-agent-instructions.ts";
 
@@ -402,8 +404,20 @@ function createAttachmentAnnotationSignals(args: {
           if (!original) {
             throw new Error("Original image is unavailable");
           }
+          // Read the address the editor just proved loadable, not the stored
+          // one. A persisted attachment's canonical URL answers only to an
+          // Authorization header, so it has to be exchanged for a presigned
+          // object URL before anything can fetch it — which is exactly what
+          // the editor's `useResolvedAttachmentUrl` does to display the same
+          // image. Deriving the URL a second way here meant the picture the
+          // user had just drawn on could still be refused at attach time.
+          const resolveResourceUrl = get(pageAttachmentResourceUrlResolver$);
+          const resolved = await get(
+            resolveResourceUrl(publicAttachmentUrl(original.url)),
+          );
+          signal.throwIfAborted();
           const flattened = await flattenAnnotatedImage(
-            original.url,
+            resolved.resourceUrl,
             annotations,
             args.filename,
             signal,
--- a/turbo/apps/platform/src/signals/okou-page/flatten-annotated-image.ts
+++ b/turbo/apps/platform/src/signals/okou-page/flatten-annotated-image.ts
@@ -11,7 +11,7 @@ import {
   REDACT_FILL,
   STROKE_HALO_INNER,
 } from "./image-annotation.ts";
-import { createDeferredPromise } from "../utils.ts";
+import { createDeferredPromise, withCleanup } from "../utils.ts";
 
 /**
  * Stroke weights are expressed against a 1000px reference edge and scaled to
@@ -62,16 +62,38 @@ function px(scale: Scale, units: number): number {
   return Math.max(1, units * scale.unit);
 }
 
-function loadImage(
+/**
+ * Reads the original image as bytes, then decodes it from a same-origin blob.
+ *
+ * This used to be one `<img crossOrigin="anonymous">`, which made flattening the
+ * only step in the feature that needed a CORS handshake: the editor renders the
+ * very same URL as a plain `<img>`, and a plain `<img>` negotiates nothing. So
+ * the image a user had just been drawing on could still refuse to load here,
+ * and an image element reports that as a bare `error` event carrying no status,
+ * no headers and no reason — which is why the failure had nothing to show for
+ * itself.
+ *
+ * A `fetch` keeps the response observable, so a refused read fails with its HTTP
+ * status attached. Decoding through `URL.createObjectURL` then happens on a
+ * `blob:` URL, which is same-origin by definition: the canvas is never tainted
+ * and `toBlob` below stays readable without asking the CDN for permission.
+ */
+async function loadImage(
   url: string,
   signal: AbortSignal,
 ): Promise<HTMLImageElement> {
+  const response = await fetch(url, { signal });
+  if (!response.ok) {
+    throw new Error(
+      `Failed to read image for flattening: ${response.status} ${response.statusText}`,
+    );
+  }
+  const blob = await response.blob();
+  signal.throwIfAborted();
+
+  const objectUrl = URL.createObjectURL(blob);
   const deferred = createDeferredPromise<HTMLImageElement>(signal);
   const image = new Image();
-  // The artifact CDN echoes the app origin back in `access-control-allow-origin`
-  // (production and per-PR preview alike), so the decoded bitmap stays readable
-  // and `toBlob` below does not throw on a tainted canvas.
-  image.crossOrigin = "anonymous";
   image.addEventListener(
     "load",
     () => {
@@ -82,12 +104,16 @@ function loadImage(
   image.addEventListener(
     "error",
     () => {
-      deferred.reject(new Error(`Failed to load image for flattening: ${url}`));
+      deferred.reject(
+        new Error(`Failed to decode image for flattening: ${blob.type}`),
+      );
     },
     { once: true, signal },
   );
-  image.src = url;
-  return deferred.promise;
+  image.src = objectUrl;
+  return await withCleanup(deferred.promise, () => {
+    URL.revokeObjectURL(objectUrl);
+  });
 }
 
 interface PixelRect {
--- a/turbo/apps/platform/src/views/okou-page/__tests__/image-annotation.test.tsx
+++ b/turbo/apps/platform/src/views/okou-page/__tests__/image-annotation.test.tsx
@@ -12,6 +12,7 @@ import {
   agentsMainContract,
 } from "@okouai/api-contracts/contracts/agents";
 import { webFilesContract } from "@okouai/api-contracts/contracts/web-files";
+import { resolveApiBase } from "../../../signals/api-base.ts";
 import { testContext } from "../../../signals/__tests__/test-helpers.ts";
 import {
   detachedSetupPage,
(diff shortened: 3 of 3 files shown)


tongx-blip:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 41e8217a2d43de509f82168c59c7d0f41847c78e

### Summary

Replaces the `` in `flattenAnnotatedImage` with a `fetch` + object-URL decode, logs the previously-discarded flatten error, and adds the first real coverage of the read path.

Two of those three land cleanly. The third — the claim in the title and body that this fixes the reported failure — does not hold up, and the diff leaves what looks like the actual cause in place.

`confirmAnnotations$` still hands `flattenAnnotatedImage` the **raw** `fileInfo.url`, while the editor renders `useResolvedAttachmentUrl(target.url)`. `attachment-resource.ts` documents that resolver as *"the URL a browser element can actually load"*, because a persisted attachment lives behind an authenticated API route and has to be exchanged for a presigned object URL. Flatten skips that exchange. That is the asymmetry that produces exactly the reported symptom — the editor displays the image, attach fails every time — and swapping the transport underneath it does not close it. A `fetch` of an unresolved authenticated URL returns 403 for the same reason the image element did; it just reports the status now.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness, edge cases, no API misuse | **Finding (P0)** | `chat-draft.ts:402` passes `original.url` unresolved; `image-annotation-editor.tsx:1094` resolves the same URL through `useResolvedAttachmentUrl`. Two derivations of one image address, one of them documented as the loadable one. |
| `REVIEW.md` | `fix:` commits carry a regression test | Pass | Two new cases in `image-annotation.test.tsx`; both verified to fail against the old loader and pass against this one. |
| `REVIEW.md` | Feature Regression Guard — adjacent flows covered | Pass | The success path, the refused-read path, the pending path and the existing send path all have cases; `chat-draft` + `attachment-chips` 80/80 unchanged. |
| `REVIEW.md` | Security | N/A | No auth, injection, or secret-handling surface in the diff. Reading fewer credentials than before, not more. |
| `REVIEW.md` | Type safety, no broad `any` | Pass | No `any`, no casts. `loadImage` keeps its `Promise<HTMLImageElement>` signature. |
| `REVIEW.md` | Style and maintainability | Finding (P1) | Test-local duplication and a captured-but-unasserted value; see P1-1. |
| `REVIEW.md` | DB/JSONB gate for persisted contract changes | N/A | No schema, contract, or `jsonb` change. |
| `docs/deployment-compatibility.md` | Deployable boundary stays safe both directions | N/A | Client-side transport change only; no request/response contract, runner protocol, queue payload, or persisted shape moves. |
| `docs/bad-smell.md` | Zero `any`, zero suppressions, no dynamic imports | Pass | None present. |
| `docs/bad-smell.md` | Fail fast, no fallback slop | Pass | `!response.ok` throws with the status rather than defaulting; no `??` chain added. |
| `docs/bad-smell.md` | No hardcoded URLs | Pass | The URL still arrives from `fileInfo$`; nothing hardcoded. |
| `docs/fallback.md` §1 | No negative tests asserting removed behaviour stays removed | Pass | The refused-read case asserts a live recovery affordance (retry badge, surviving mark count), not the absence of deleted code. |
| `docs/fallback.md` §6 | Fallbacks limited to the required categories | Pass | The diff introduces none — no compatibility branch, no tolerance path. No `Fallbacks` section required. |
| `docs/effect.md` | Render stays pure; work runs in commands | Pass | All new work is inside `loadImage` and the existing `confirmAnnotations$` command. No render-phase I/O added. |
| `docs/cache.md` RC-8 | Symmetric resource lifecycles | Pass | `URL.createObjectURL` is paired with `revokeObjectURL` through `withCleanup`'s `finally`, so it runs on resolve, reject and abort. `createDeferredPromise` rejects immediately on an already-aborted signal (`utils.ts:558`), so the promise cannot hang and strand the URL. |
| `docs/cache.md` RC-3 | No unbounded lifetime cache | Pass | The object URL is per-call and released in the same call. |
| `.claude/skills/ccstate/SKILL.md` | ccstate commands, no React hooks | Pass | No hooks added; the module-level `log` is a logger handle, not a signal. |
| `docs/testing.md` | Mock at boundary, MSW for HTTP, no fake timers | Pass | New HTTP behaviour goes through `context.mocks.http` (MSW). No `vi.stubGlobal("fetch")`, no timers, no `vi.mock` of internal paths. |
| `docs/testing/app-testing.md` | Page-level tests entering through the real setup | Pass | Both cases run through `setup(true)` → lightbox → annotate → attach. |
| `docs/testing/anti-patterns.md` | AP-1 — no mock-call assertions as the sole check | Pass | The load-bearing assertion is the persisted `ANNOTATED_FILE_ID`; the request assertion is supplementary. Borderline, see P1-1. |
| `docs/testing/anti-patterns.md` | AP-2 — no direct fetch mocking | Pass | MSW throughout. |

### Findings

- **`turbo/apps/platform/src/signals/okou-page/chat-draft.ts:402`** — flatten reads an unresolved attachment URL, which is the most likely cause of the reported failure and is untouched by this PR.

  ```ts
  const flattened = await flattenAnnotatedImage(
    original.url,   // raw fileInfo.url
    ...
  ```

  The editor that just rendered the same image does this instead (`image-annotation-editor.tsx:1094`):

  ```ts
  const resolvedUrl = useResolvedAttachmentUrl(target.url);
  ```

  `useResolvedAttachmentUrl` runs `publicAttachmentUrl(url)` and then `pageAttachmentResourceUrlResolver$`, which exchanges an authenticated `/api/web/download-file` URL for a presigned object URL — precisely because *"a `src` attribute cannot carry an Authorization header"*. For a restored attachment `fileInfo$` returns `persisted.url ?? resolved.body.url` (`chat-draft.ts:643`), so the raw value can legitimately be the authenticated form. The editor loads it, flatten does not.

  Route the flatten URL through the same resolver before handing it to `flattenAnnotatedImage`, so the read is guaranteed to target the address the editor already proved loadable. `pageAttachmentResourceUrlResolver$` is a `computed` in `signals/attachment-resource-url.ts` and is reachable from this command with `get`; `signals/` importing `publicAttachmentUrl` from `views/okou-page/attachment-url.ts` matches existing practice in this directory. Add a case where the attachment's stored URL is the authenticated form and assert the flatten still succeeds — the current tests both use a public CDN URL, so neither one covers this.

  Until that lands, the PR's stated purpose is unproven: the transport swap makes the failure *legible* but there is no evidence in the diff or the PR body that it makes it *stop*.

- **`image-annotation.test.tsx`, "attaches marks when the original image is only readable as bytes"** — the test re-registers a `FILE_URL` handler that `mockSuccessfulAnnotationUpload()` → `mockOriginalImageBytes()` already provides, duplicating the response body, and collects `request.headers.get("origin")` into `crossOriginRequests` while asserting only `.length`. The captured origin is never checked, and the name points at a cross-origin handshake that this change specifically stops performing. Reuse the shared helper and count the reads under an honest name, or assert something real about the request. The count assertion itself should stay — it is what fails against the old loader.

### Testing

- **Coverage: Adequate for what the diff changes, insufficient for what it claims.** 17/17 in `image-annotation.test.tsx`, 80/80 across `chat-draft` and `attachment-chips`, and both new cases were verified in both directions (2 failed / 15 passed against the old `crossOrigin` loader; 17 passed here). That is a genuine regression guard for the read path, and it closes a real hole — `window.Image` is stubbed to report `load` for any src, so this path had no coverage at all before. But no case exercises an authenticated attachment URL, which is the scenario in P0.
- **Conventions: Compliant.** MSW at the boundary, page-level entry through `detachedSetupPage`, no fake timers, no internal `vi.mock`, no mock-call assertion standing alone. The `mockAnnotationRendering` / `mockAnnotationCanvas` split is a reasonable seam for letting one case serve different bytes.
- **Note on the logging change:** `log.warn` rather than `log.error` is correct here, and not only for severity reasons — the platform test setup makes `console.error` throw (`src/test/setup.ts:318`), so an `error` call placed before the state write would abort the command and leave the attachment stuck in `pending`. The current ordering (write the state, then log) is right regardless.

tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: fbdbaddcf612ed4bc2b30e17b99edff79956c6f4
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Third pass, at `fbdbadd`. The P0 from the previous round is closed: `confirmAnnotations$` no longer derives the image address a second way. It runs the same `publicAttachmentUrl` → `pageAttachmentResourceUrlResolver$` pair the editor uses, so flattening reads exactly the URL the editor already proved loadable — the presigned object URL, not the canonical API route that answers only to an Authorization header.

That makes the three commits do three separable jobs, and it is worth being precise about which one fixes what. The `fetch` + object-URL swap made a silent `error` event into a real HTTP status and removed the canvas-taint dependency; the `log.warn` stopped five distinct failure modes collapsing into one unexplained badge; **this** commit is the one that stops the reported failure. The regression case proves that ordering: without the resolver hop it fails with `expected false to be truthy` — no annotated file id ever reaches the draft.

The P1 is closed too, and the follow-up nit it introduced (an unused `url` parameter on the byte helper, against the YAGNI rule in `CLAUDE.md`) is gone in this head.

No P0s, no P1s.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness, edge cases, no API misuse | Pass | One derivation of the image address for both consumers. `signal.throwIfAborted()` after the resolve keeps a cancelled flatten from proceeding; the resolver's own request stays page-owned, which is correct for a page-scoped shared entry. |
| `REVIEW.md` | `fix:` commits carry a regression test | Pass | Three cases, each verified to fail against the code it guards: the two read-path cases against the old `crossOrigin` loader, the presigned case against the missing resolver hop. |
| `REVIEW.md` | Feature Regression Guard — adjacent flows covered | Pass | Public-URL attach, canonical-URL attach, refused read, pending upload and the existing send path all covered. `chat-draft` + `attachment-chips` 80/80 unchanged, so the shared `mockDraftWithImage` signature change broke nothing. |
| `REVIEW.md` | Security | Pass | The resolver is the *authorized* path — the API still runs its ownership check before returning a presigned URL. This reads through it rather than around it. |
| `REVIEW.md` | Type safety, no broad `any` | Pass | No `any`, no casts. `resolved.resourceUrl` is typed by `AttachmentUrls`. |
| `REVIEW.md` | Style and maintainability | Pass | Net effect on the test file is less duplication than before the review: one byte-serving helper, `mockAnnotatedUploadSuccess` split out so a case can vary bytes without restating the upload mock. |
| `REVIEW.md` | DB/JSONB gate for persisted contract changes | N/A | No schema, contract or `jsonb` change. |
| `docs/deployment-compatibility.md` | Deployable boundary stays safe both directions | N/A | Client-side only. `webFilesContract.fileUrl` is called with its existing shape; no request/response contract, runner protocol, queue payload or persisted shape moves. |
| `docs/bad-smell.md` | Zero `any`, zero suppressions, no dynamic imports | Pass | None present. |
| `docs/bad-smell.md` | Fail fast, no fallback slop | Pass | A failed resolve rejects into the same `settle`, surfacing as the retry badge. No `??` chain, no second address to fall back to — removing the second address is the change. |
| `docs/bad-smell.md` | No hardcoded URLs | Pass | The test builds the canonical address from `resolveApiBase()` rather than pinning an origin. |
| `docs/fallback.md` §1 | No negative tests asserting removed behaviour stays removed | Pass | `expect(canonicalReads).toBe(0)` asserts a live routing guarantee — which address the running code reads — not the absence of deleted code. |
| `docs/fallback.md` §6 | Fallbacks limited to the required categories | Pass | The diff introduces none. No `Fallbacks` section required. |
| `docs/effect.md` | Render stays pure; work runs in commands | Pass | The resolve happens inside the existing `confirmAnnotations$` command. Nothing added to render. |
| `docs/cache.md` RC-3 | No unbounded lifetime cache | Pass | Reuses `pageAttachmentResourceUrlResolver$`, whose map is replaced with the page signal (`attachment-resource-url.ts:103`). No new cache, and the flatten now shares the entry the preview already populated. |
| `docs/cache.md` RC-8 | Symmetric resource lifecycles | Pass | Unchanged from the last pass: `createObjectURL` is paired with `revokeObjectURL` via `withCleanup`'s `finally`, covering resolve, reject and abort. |
| `docs/cache.md` RC-5 | No duplicate mutable sources of truth | Pass | This removes one — two derivations of the loadable image address collapse to the resolver. |
| `.claude/skills/ccstate/SKILL.md` | ccstate commands, no React hooks | Pass | `get(pageAttachmentResourceUrlResolver$)` then `get(resolveResourceUrl(...))` inside a command, which is the documented way to consume a resolver from signals. No hooks. |
| `docs/testing.md` | Mock at boundary, MSW for HTTP, no fake timers | Pass | MSW throughout; the 401 guard sits on `*/api/web/download-file`, distinct from the resolver's `/api/web/file-url`, so the two do not collide. |
| `docs/testing/app-testing.md` | Page-level tests entering through the real setup | Pass | All three cases run `setup(true)` → lightbox → annotate → attach. |
| `docs/testing/anti-patterns.md` | AP-1 — no mock-call assertions as the sole check | Pass | Every case leads with a user-visible outcome (persisted `ANNOTATED_FILE_ID`, or the retry badge plus surviving mark count). The read counters are corroborating, and they are what distinguish a real read from a stubbed `Image` reporting `load`. |
| `docs/testing/anti-patterns.md` | AP-2 — no direct fetch mocking | Pass | No `vi.stubGlobal("fetch")`. |
| `CLAUDE.md` | YAGNI — no "just in case" parameters | Pass | The `url` parameter added to `mockOriginalImageBytes` in the previous head was never passed by any caller; `fbdbadd` removes it. |

### Findings

None.

### Testing

- **Coverage: Adequate.** 18/18 in `image-annotation.test.tsx`, 98/98 with `chat-draft` and `attachment-chips`. The gap called out last round is closed: `flattens through the presigned url when the draft stored the api one` drives a draft whose stored attachment URL is `${resolveApiBase()}/api/web/download-file?file_id=...`, asserts the presigned URL was read, and asserts the canonical route was never fetched. Verified to fail without the resolver hop.
- **Conventions: Compliant.** MSW at the boundary, real page startup, no fake timers, no internal `vi.mock`, no mock-call assertion standing alone. Adding `storedUrl` as an optional third parameter to `mockDraftWithImage` keeps every existing caller untouched and is exercised by the new case, so it is not a speculative parameter.
- **Still worth knowing:** this path had no real coverage before this PR — `window.Image` is stubbed to report `load` for any src, so a read that failed for every annotated screenshot looked green. That stub is why the bug shipped, and the three cases here are the first that would have caught it.

PR #31366, opened 2026-09-03
perf(api): overlap attachment metadata resolution

### LINKED ISSUE #31345: perf(api): overlap attachment metadata with normal send preparation
## Background

Parent objective: #24203.

The rolling production window mixes frequent API and Runner releases, so behavior selection uses
the mature exact pair API `1ed230b2281440c96b1a23ad4a6298153a440f35` / Runner `0.184.0`.

## Problem

Normal web-chat preparation resolves attachment metadata only after thread and run-configuration
resolution, explicit selection persistence, and computer-use grant resolution have completed.

After the agent, client-event preflight, template, explicit run configuration, and initial model pin
have been validated, attachment lookup depends only on the authenticated user and the already
projected user message. It resolves existing object metadata and does not consume the thread or
model-resolution result. Keeping it at the end adds its complete latency to sends with physical
attachments.

## Production evidence

The fixed pair contained 287 normal sends, including 33 sends with one or more physical
attachments. For those 33 sends, attachment metadata resolution measured approximately 107 ms p50
and 157 ms p90.

Ideal same-run removal moves the affected cohort's `api_to_spawn` p90 from approximately 1,738 ms
to 1,604 ms and its one-second count from 10 to 13. Across all 287 normal sends, the ideal p90 moves
from approximately 1,802 ms to 1,767 ms. The cohort is small and these are upper bounds, not a
production latency promise.

## Proposed change

After the existing prerequisite validations and initial model pin succeed, start attachment
metadata resolution before thread resolution. Await its result at the existing semantic boundary
before returning the prepared normal send and before queue-first persistence.

Retain the current ordered waves of four attachment lookups. Do not increase per-file concurrency,
cache object metadata, weaken owner validation, or accept a missing attachment.

## Acceptance criteria

- Attachment metadata resolution overlaps thread/run-configuration preparation for normal sends
  with physical attachments.
- Requests rejected by the existing agent, client-event preflight, template, explicit model, or
  initial-pin validations do not start attachment work.
- An API-route integration test gates attachment lookup and thread preparation and proves they are
  in flight together while preserving the final attachment order.
- Missing objects, ownership mismatches, lookup failures, cancellation, and later preparation
  failures retain deterministic handling without unhandled rejections.
- The existing maximum of four concurrent attachment lookups and ordered-wave behavior remain
  unchanged.
- Queue-first persistence, thread authority, model admission, computer-use authorization, and the
  real spawn boundary remain unchanged.
- Production validation uses one exact containing API/Runner pair and reports the attachment
  boundary, failure shape, p95/p99, and the fraction at or below 1,000 ms for the affected cohort.

## Risks and boundaries

Some requests that later fail thread or computer-use preparation may perform read-only object
metadata requests earlier than today. Do not move the work ahead of the listed prerequisite
validations, and do not expand attachment fan-out. Preserve the current externally managed object
reference behavior: a missing object remains unavailable and cannot become an accepted attachment.

This slice is independent of the broad storage-manifest overlap slice and #27735. It does not change
Runner claim transport, storage manifest construction, goal admission, or durable queue ownership.

seven332:
## Summary

- start normal-send attachment metadata resolution after the authoritative agent, event, template, model, and initial-pin checks
- overlap the read-only storage lookup with thread/run-configuration reconciliation, explicit selection persistence, and computer-use resolution
- observe speculative lookup failures immediately while preserving the existing failure precedence at the original attachment boundary
- cover both real overlap under a held thread row lock and a later authoritative 404 winning over a delayed attachment rejection

## Scope

- keeps the existing ordered attachment waves of four, ownership validation, fallback lookup, prompt ordering, and queue-first behavior
- does not change attachment APIs, storage formats, UI, or user flows

## Validation

- `pnpm format`
- `pnpm -F api lint`
- `pnpm -F api check-types`
- `pnpm knip --workspace api`
- `pnpm -F api exec vitest run src/signals/routes/__tests__/chat-events.bdd.test.ts --silent=passed-only` (166 passed)
- focused regression run for the two new overlap/failure-precedence scenarios (2 passed)
- repository pre-commit hooks, including full Turbo Knip and type checks

Closes #31345

Parent context: #24203

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts
@@ -15312,6 +15312,230 @@ describe("CHAT-02: generation templates and attachments", () => {
     await cancelChatRun(actor, active.runId);
   }, 90_000);
 
+  it("overlaps attachment metadata with thread model reconciliation", async () => {
+    const { actor, agentId, providerId } = await entitledChatActor();
+    chatCallbacks.failIfChatCallbackRouteIsFetched();
+    await api.updateOrgModelPolicies(actor, [
+      {
+        model: "claude-sonnet-5",
+        isDefault: true,
+        defaultProviderType: "anthropic-api-key",
+        credentialScope: "org",
+        modelProviderId: providerId,
+      },
+    ]);
+    const thread = await chat.createThread(actor, {
+      agentId,
+      model: "claude-sonnet-5",
+    });
+
+    await seedVm0BuiltInModelKey("gpt-5.6-terra");
+    await api.updateOrgModelPolicies(actor, [
+      {
+        model: "gpt-5.6-terra",
+        isDefault: true,
+        defaultProviderType: "built-in",
+        credentialScope: "org",
+        modelProviderId: null,
+      },
+    ]);
+
+    const files = Array.from({ length: 2 }, (_, index) => {
+      return {
+        id: randomUUID(),
+        filename: `overlap-${index + 1}.txt`,
+        contentType: "text/plain",
+        size: 40 + index,
+      };
+    });
+    const objectsByKey = new Map(
+      files.map((file) => {
+        return [buildArtifactKeyV2(file.id, file.filename), file];
+      }),
+    );
+    const releaseHeads = createDeferredPromise<void>(context.signal);
+    let startedHeads = 0;
+    let activeHeads = 0;
+    context.mocks.s3.send.mockImplementation(
+      async (command: unknown): Promise<unknown> => {
+        if (command instanceof HeadObjectCommand) {
+          const key = command.input.Key;
+          const file =
+            typeof key === "string" ? objectsByKey.get(key) : undefined;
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #31366

**Reviewed head:** `7b3dabfd6d25b50ee22c78a6c0b65ac54dbe773f`

### Summary

The attachment lookup starts after the established prerequisite gates, remains required at its original semantic boundary, and now overlaps independent thread/model/computer-use preparation. Immediate settled-result observation preserves authoritative failure precedence without an unhandled rejection, while the existing four-at-a-time ordered waves and fail-closed ownership lookup remain unchanged.

### Findings

No findings.

### Testing

The API route integration coverage deterministically proves real PostgreSQL thread reconciliation and two S3 heads are simultaneously in flight, verifies final attachment order, and covers a delayed attachment rejection after an authoritative thread 404. The complete 166-test chat-event route suite, API format/lint/type checks, Knip, focused regressions, and repository pre-commit hooks passed. Production latency validation remains a post-deployment task under parent #24203.

### Verdict

LGTM

PR #31390, opened 2026-09-03
perf(mitm-addon): stream oversized firewall path matching

### LINKED ISSUE #31350: perf(python): avoid segmenting oversized firewall request paths
## Problem

At audited `main` commit `26da2c10a64c04533f62176854ce62e74e415804`, compiled firewall matching has a 64 KiB path-validation budget, but indexed candidate selection materializes every request path segment before that budget is checked.

Relevant call order:

- `crates/runner/mitm-addon/src/path_security.py:13-21` defines `MAX_PATH_VALIDATION_CHARACTERS = 64 * 1024` and makes the first operation in `has_unsafe_path()` an O(1) length rejection.
- `crates/runner/mitm-addon/src/matching.py:1944-1959` parses the runtime URL and calls `compiled_firewalls.indexed_api_candidates(url_parts)` before entering the shared API-candidate evaluator.
- `crates/runner/mitm-addon/src/matching.py:662-673` passes the complete `url_parts.path` to `_split_path_segments()` while walking the static-base trie.
- `crates/runner/mitm-addon/src/firewall_matching/patterns.py:208-215` implements that helper with `path_without_leading_slash.split("/")`, allocating one list entry and potentially one substring per request-controlled segment.
- Only later, at `crates/runner/mitm-addon/src/matching.py:1763-1779`, does a matching base call `has_unsafe_path()` and return the existing `FirewallBlock(reason="unsafe_path")`.
- `crates/runner/mitm-addon/src/request_classification.py:487-497` invokes this matcher synchronously from request classification, before upstream admission.

As a result, an above-budget path that is guaranteed to fail closed can still force CPU and transient allocation proportional to its complete segment count before the constant-time guard runs. The pinned `mitmproxy==12.2.3` HTTP/1 reader does not impose a general request-head limit; this reachability was already established while resolving #26832.

## Measured impact

A local diagnostic used the repository's exact Python 3.14.0 runtime, one compiled static root API (`https://api.example.com`) with no permissions, and the production `match_compiled_firewall_request()` entry point. `tracemalloc` started after constructing each URL, so the figures below are additional transient traced allocations during matching:

| Request path | Result | Time | Peak traced allocation |
| --- | --- | ---: | ---: |
| 1,000,000 slash characters | `FirewallBlock(unsafe_path)` | 18.87 ms | 9.98 MiB |
| 5,000,000 slash characters | `FirewallBlock(unsafe_path)` | 83.31 ms | 51.53 MiB |

For comparison, a 5,000,000-character single-segment path took 28.40 ms and 9.54 MiB. The slash-heavy input therefore amplifies the eager segment-list allocation substantially even though both inputs exceed the same 64 KiB budget and receive the same final block decision.

The algorithm is linear, but its work is not bounded by the documented validation budget. Repeated requests can stall the mitmproxy event loop and create large transient memory pressure before rejection.

## Scope and deduplication

- Closed #19419 replaced quadratic prefix-tuple construction with trie traversal. It intentionally retained one linear path-segment list and does not move the later validation budget ahead of that allocation.
- Closed #26832 added the 64 KiB O(1) guard to `has_unsafe_path()`. This issue is the remaining ordering gap: production indexed lookup runs `_split_path_segments()` before reaching that guard.

This issue should preserve the existing above-budget `unsafe_path` outcome for matched firewalls; it does not require a new request-size policy.

## Acceptance criteria

- Indexed firewall candidate selection does not split or otherwise materialize the complete segment list for a runtime path longer than `MAX_PATH_VALIDATION_CHARACTERS` before returning the existing `unsafe_path` block for a matching firewall.
- Work performed before the above-budget decision is bounded independently of the request's total slash count. Residual URL reconstruction/parsing cost should be described accurately; this criterion does not require the entire matcher to become constant-time.
- Oversized requests to unrelated destinations retain current no-match/pass-through behavior; do not globally block them merely because a compiled firewall set exists.
- Root/static bases, parameterized bases, repeated slashes, encoded slashes, connector-owner ambiguity, malformed-config precedence, and indexed/linear decision parity remain unchanged for supported paths.
- Add a deterministic regression using an above-budget slash-heavy path that proves full-path segmentation is not entered (or that its input is capped) without a wall-clock assertion.
- Record a before/after diagnostic for the production matcher on slash-heavy inputs; keep timing out of the test contract.

seven332:
## Summary

- stream runtime path segments into the compiled static API trie instead of materializing the complete segment list
- bound parameterized base matching to its configured prefix depth while preserving captures, relative paths, repeated slashes, and retained malformed greedy semantics
- add a production-matcher regression covering oversized slash-heavy static and parameterized matches plus unrelated-destination pass-through

## Problem

The compiled firewall matcher applies its existing 64 KiB unsafe-path budget only after indexed candidate selection. Candidate selection previously called `split("/")` on the complete runtime path, so a request guaranteed to return `FirewallBlock(reason="unsafe_path")` could first allocate one list entry per attacker-controlled slash. Parameterized fallback bases repeated the same complete split during base matching.

## Implementation

The existing decision pipeline remains authoritative. This PR changes only how its prefix consumers obtain segments:

- `_iter_path_segments()` preserves the existing segment contract but lets trie traversal stop at the configured depth or first miss.
- `_split_path_segments()` remains available as the materialized compatibility helper for consumers that need every bounded rule segment.
- parameterized base matching splits only the number of leading segments present in the compiled base pattern and derives the exact relative path from the original string boundary.

There is no global request-target limit, API change, persisted-state change, feature switch, or rollout dependency. Oversized requests still block only after a firewall base matches; unrelated destinations continue to pass through.

## Diagnostic

Using the locked Python 3.14.0 environment, one compiled static root API, the production `match_compiled_firewall_request()` entry point, and `tracemalloc` started after URL construction:

| Slash count | Before peak | After peak | Before local time | After local time | Result |
| ---: | ---: | ---: | ---: | ---: | --- |
| 1,000,000 | 9.98 MiB | 1.91 MiB | 6.52 ms | 1.09 ms | `FirewallBlock(unsafe_path)` |
| 5,000,000 | 51.53 MiB | 9.54 MiB | 34.58 ms | 5.22 ms | `FirewallBlock(unsafe_path)` |

Timing is diagnostic, not a test contract. URL parsing, the parsed path string, and exact relative-path slices can still scale with raw input length; the slash-count-amplified complete segment list is removed.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_compiled_firewall_indexed_matching.py tests/test_compiled_firewall_unknown_policy.py tests/test_compiled_firewall_base_path_matching.py tests/test_compiled_firewall_malformed_base.py tests/test_compiled_firewall_base_specificity_precedence.py tests/test_compiled_firewall_rule_specificity_precedence.py tests/test_firewall_semantics_contract.py` — 342 passed
- `uv run --no-sync ruff format --check .` — 360 files already formatted
- `uv run --no-sync basedpyright -p .` — 0 errors, 0 warnings, 0 notes
- `uv run --no-sync python -m pytest tests/` — 7312 passed
- exhaustive short-path split parity diagnostic against the previous implementation

Closes #31350

### DIFF
--- a/crates/runner/mitm-addon/src/firewall_matching/base_url.py
+++ b/crates/runner/mitm-addon/src/firewall_matching/base_url.py
@@ -17,7 +17,7 @@ from firewall_matching.patterns import (
     SegmentLiteral,
     SegmentParam,
     _match_compiled_host,
-    _match_compiled_path_prefix,
+    _match_compiled_path_prefix_text,
     _parse_segment,
     _split_path_segments,
 )
@@ -635,13 +635,10 @@ def _match_compiled_base_url_parts(
     base_path = base.parts.path
     clean_url_path = url_parts.path
     if base_path and base_path != "/":
-        url_path_segs = _split_path_segments(clean_url_path)
-        path_result = _match_compiled_path_prefix(url_path_segs, base.path_segments)
+        path_result = _match_compiled_path_prefix_text(clean_url_path, base.path_segments)
         if path_result is None:
             return None
-        path_params, consumed = path_result
-        remaining_segs = url_path_segs[consumed:]
-        rel_path = "/" + "/".join(remaining_segs) if remaining_segs else "/"
+        path_params, rel_path = path_result
         all_params = {**host_params, **path_params}
     else:
         rel_path = clean_url_path or "/"
--- a/crates/runner/mitm-addon/src/firewall_matching/patterns.py
+++ b/crates/runner/mitm-addon/src/firewall_matching/patterns.py
@@ -223,7 +223,12 @@ def _iter_path_segments(path: str) -> Iterator[str]:
 
 def _split_path_segments(path: str) -> list[str]:
     """Split path patterns and request paths without normalizing repeated slashes."""
-    return list(_iter_path_segments(path))
+    if path in ("", "/"):
+        return []
+    path_without_leading_slash = path[1:] if path.startswith("/") else path
+    if path_without_leading_slash == "":
+        return []
+    return path_without_leading_slash.split("/")
 
 
 def _split_path_prefix(path: str, segment_count: int) -> tuple[list[str], str]:
--- a/crates/runner/mitm-addon/src/firewall_matching/patterns.py
+++ b/crates/runner/mitm-addon/src/firewall_matching/patterns.py
@@ -1,5 +1,6 @@
 """Low-level firewall segment, host, and path pattern matching."""
 
+from collections.abc import Iterator
 from typing import NamedTuple
 
 _SEGMENT_ERROR_HINT = 'use "{name}", "prefix{name}", "{name}suffix", or "prefix{name}suffix"'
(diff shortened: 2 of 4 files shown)


seven332:
## Code Review: PR #31390

**Reviewed head:** `d8bcb34d2903b78d15e1d737c0d66af5911e3da1`

### Summary

The current head removes slash-count-amplified complete path segmentation before the existing firewall unsafe-path decision. Static candidate lookup now stops at trie depth, parameterized bases consume only configured prefix depth, and the shared matcher remains authoritative for base collection, owner ambiguity, malformed state, policies, and final results. The follow-up commit also preserves the existing C-backed split for normal bounded rule paths.

### Findings

No findings on the reviewed head. The initial submitted head's normal-path Python-loop regression was fixed in `d8bcb34d29` and the regression was strengthened to prove a static root lookup does not advance the oversized iterator.

### Testing

- focused matcher, malformed-base, specificity, unknown-policy, and semantics suites: 342 passed
- complete mitm-addon suite: 7312 passed
- Ruff format/check: passed
- BasedPyright: 0 errors, 0 warnings, 0 notes
- old/new parameterized-prefix diagnostic: 26,248 generated cases equal
- production diagnostic peak traced allocation: 9.98 MiB → 1.91 MiB at 1M slashes; 51.53 MiB → 9.54 MiB at 5M slashes

Timing remains diagnostic only. Raw URL parsing/path retention and exact relative-path slices can still scale with input length; the reviewed guarantee is removal of complete slash-derived segment-list materialization before the decision.

### Verdict

LGTM

PR #31391, opened 2026-09-03
fix(python): detect dict constructor pair keys

### LINKED ISSUE #31348: bug(python): detect dict constructors used as metadata pair iterables
## Problem and evidence

`crates/runner/mitm-addon/scripts/flow_metadata_linter/metadata_rules.py` skips a built-in
`dict(...)` call when the active collection mode is `PAIR_ITERABLE`.

The gap remains reproducible on main commit `89c6a52194`:

```python
flow.metadata.update([*dict({("sandbox_run_id", value): None})])
```

The linter reports no violation, but Python iterates the constructed dictionary as
`("sandbox_run_id", value)`. `flow.metadata.update()` therefore accepts that tuple as a key/value
pair and writes the registered `sandbox_run_id` key. This lets a statically visible hard-coded key
bypass the repository guard.

## Decision

Implement this as one focused Python tooling PR. Model the constructor's resulting-key semantics
explicitly so the linter does not confuse a `dict(...)` input entry with the dictionary key that
the constructor produces.

Add one collection mode representing input consumed by a mapping constructor whose resulting keys
are later consumed as update pairs.

- Mapping literals and dictionary comprehensions expose their keys directly for exact two-item pair
  validation.
- Sequence and comprehension constructor inputs expose each entry's first element as the resulting
  key before pair validation.
- Existing conditional, wrapper, starred-argument, and known-call traversal is reused only where
  the projection remains statically sound.
- Unknown or ambiguous dynamic expressions remain unreported.
- Existing mapping-input and key-sequence behavior remains unchanged.

This covers both the reported mapping-literal form and equivalent statically knowable
iterable-of-entry constructors while preventing an input such as
`dict([("sandbox_run_id", value)])` from being misreported: that constructor's resulting key is the
string `sandbox_run_id`, not a two-item pair.

## Alternatives considered

- **Handle mapping-literal arguments only:** smaller, but leaves equivalent iterable-of-entry
  constructors uncovered and bases coverage on syntax rather than result semantics.
- **Forward every constructor input in pair-iterable mode:** rejected because it produces false
  positives by treating constructor entries as final update pairs.
- **Replace the mode system with a general abstract element-shape engine:** more general but not
  justified by this narrow confirmed gap and would substantially increase regression surface.

## Risks and mitigations

- **False positives:** public negative fixtures will cover ordinary string result keys,
  wrong-length tuple keys, and entry iterables whose first element is not a valid result pair.
- **Existing mode regressions:** the new transition is limited to built-in `dict(...)` under
  pair-iterable consumption, and the complete existing public contract fixture remains exercised.
- **Analysis expansion:** traversal remains static and bounded by the existing first-argument
  helper; no expression execution or dynamic type inference is added.

## Acceptance criteria

- Diagnose a registered literal in the reported starred built-in `dict(...)` result through the
  public `metadata_key_violations(path)` contract.
- Diagnose a wrapped constructor result and a statically knowable iterable-of-entry constructor
  whose resulting dictionary key is a valid two-item pair.
- Do not diagnose ordinary string keys, wrong-length tuple keys, or constructor entries whose
  resulting key cannot be a valid two-item metadata update pair.
- Preserve constant-based allowed forms and all existing flow-metadata linter behavior.
- Pass the focused flow-metadata contract test, metadata check script, Ruff, BasedPyright, the full
  mitm-addon pytest suite, and PR CI.

## Delivery notes

The change is limited to the Python linter and its public fixtures. It does not change runner
runtime behavior, APIs, persisted data, dependencies, migrations, or deployment compatibility.

seven332:
## Summary

- model built-in `dict(...)` inputs whose resulting keys are consumed as metadata update pairs
- distinguish constructor entries from their resulting dictionary keys to avoid string-key false positives
- cover starred, wrapped, mapping-input, iterable-entry, wrong-shape, and constant-based forms through the public linter contract

## Scope

This changes only the static flow-metadata key linter and its fixtures. It does not execute expressions, infer dynamic collections, change runner runtime behavior, or modify APIs, persisted data, dependencies, migrations, or deployment protocols.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_flow_metadata_key_contract.py -q` (23 passed)
- `uv run --no-sync python -m pytest tests/` (7311 passed)
- pre-commit Ruff, file-size, and commit-message hooks

Closes #31348


seven332:
## Code Review: PR #31391

**Reviewed head:** `8b65d76942c9fdf27792e5145a9c6fa714aaa5e8`

### Summary

The final head models the missing `dict(...)` result-key projection without treating constructor entries as final metadata update pairs. It stays within the static linter, keeps unknown call results conservative, and exercises the retained behavior through the public flow-metadata contract.

### Findings

No findings.

### Testing

Ruff format/lint, BasedPyright, the 23-test flow-metadata contract, the metadata check script, and the complete 7311-test mitm-addon suite passed on this head. Codecov reports all modified coverable lines covered. Rust and Turbo suites were not run locally because they are outside the affected Python tooling scope; PR CI remains the repository-wide verification source.

### Verdict

LGTM

PR #30857, opened 2026-09-01
fix(platform): replace instatus widget with branded status notice

hulh122:
## Summary

- replace the third-party Instatus iframe widget with a native Zero status notice
- render the mobile notice only inside the expanded sidebar
- preserve production-only polling, status-page links, and dismissal persistence
- centralize production status configuration and localize service-status copy across every supported app locale

## Verification

- `pnpm format`
- `NODE_OPTIONS=--max-old-space-size=3072 pnpm turbo run lint --concurrency=1 --log-order grouped`
- `pnpm check-types`
- `pnpm -F @okouai/app exec vitest run src/views/router/__tests__/instatus-status-notice.test.tsx`
- `pnpm knip --workspace @okouai/app`

## Fallbacks

- `turbo/apps/platform/src/signals/instatus-status.ts:parseIssueList` treats a missing issue collection as empty because Instatus omits collections with no active entries. This is provider-response compatibility, not vm0 version skew. Remove only if Instatus guarantees both collections are always present.

PR #31415, opened 2026-09-03
refactor(platform): route instatus status polling through the shared loop helper

e7h4n:
## Violation found

The daily platform loop audit found one new violation on `main`:

- `turbo/apps/platform/src/signals/instatus-status.ts` — `pollInstatusIssues$` drove its own `while (!signal.aborted)` loop with a raw `delay` to refresh the Instatus service-status data every 3 minutes. Recurring async work in the platform must go through the shared `setLoop` helper rather than a hand-rolled polling loop.

Introduced by #30857 (`fix(platform): replace instatus widget with branded status notice`).

## Fix

Routed the recurring refresh through `setLoop`, keeping the delay inside the loop body:

```ts
await setLoop(
  async (loopSignal) => {
    await delay(STATUS_REFRESH_INTERVAL_MS, { signal: loopSignal });
    set(refreshVersion$, (version) => version + 1);
    return false;
  },
  0,
  signal,
  { retryTransientErrors: false },
);
```

## Signal ownership

- Owner is the unchanged app-root `AbortSignal` that `views/main.tsx` already passes when it detaches the daemon (`store.get(rootSignal$)`).
- The inner `delay` is bound to `loopSignal`, so both a parent abort and the loop's own teardown cancel the pending timer.
- No placeholder or never-aborting signal was introduced.

## Preserved behavior

- **Timing is unchanged.** The delay stays at the top of the loop body with `setLoop`'s own interval set to `0`, so the first refresh still lands one full `STATUS_REFRESH_INTERVAL_MS` after startup, and each subsequent refresh one interval after the previous one. Moving the delay to `setLoop`'s interval argument would have fired an extra refresh at t=0.
- **Test timing is unchanged.** `setLoop` collapses its own interval to a macrotask under Vitest; keeping the real delay in the body means this app-root daemon still ticks on the real 3-minute interval in tests instead of spinning and tripping `MAX_LOOP_COUNT_IN_TEST`.
- **Stopping conditions are unchanged.** The body always returns `false`, so the loop still terminates only on abort.
- `retryTransientErrors: false` preserves the original semantics, where any non-abort rejection propagated instead of being retried with backoff.

This matches the existing precedent in `signals/shared-database-browser.ts` (heartbeat loop) and `signals/okou-page/presentation-template-library.ts`.

## Also audited, no change needed

Every other `signal-timers` call added since the previous audit already carries a real owner signal:

- `shared-database/worker-signals.ts:225` — one-shot throttle `delay` with the caller's signal, using `now()`.
- `signals/chat-page/run-work-folding.ts` — finite array traversal, not a control-flow loop.

## Local checks

- Both audit scans re-run: the native timer scan is empty, and the loop scan now matches only the two known exemptions (`signals/utils.ts` `setLoop` itself, and the artifact cursor pagination in `create-artifact-catalog-signals.ts`).
- Prettier on the changed file: clean.
- `oxlint --deny-warnings` on the changed file: clean.
- `tsc -p tsconfig.production.json --noEmit`: no new errors (the two pre-existing module-resolution errors reproduce identically on unmodified `main` in this container).
- `vitest run src/views/router/__tests__/instatus-status-notice.test.tsx` — 5 passed.
- `vitest run src/__tests__/platform-entrypoint.test.ts` — 6 passed.

Full validation is delegated to the PR pipeline.

### DIFF
--- a/turbo/apps/platform/src/signals/instatus-status.ts
+++ b/turbo/apps/platform/src/signals/instatus-status.ts
@@ -3,7 +3,7 @@ import { delay } from "signal-timers";
 
 import { localStorageSignals } from "./external/local-storage.ts";
 import { rootSignal$ } from "./root-signal.ts";
-import { jsonParseOr } from "./utils.ts";
+import { jsonParseOr, setLoop } from "./utils.ts";
 import { resolvePlatformServiceStatusConfig } from "../lib/platform-host.ts";
 
 const STATUS_REFRESH_INTERVAL_MS = 3 * 60 * 1000;
@@ -138,11 +138,17 @@ export const pollInstatusIssues$ = command(
       return;
     }
 
-    while (!signal.aborted) {
-      await delay(STATUS_REFRESH_INTERVAL_MS, { signal });
-      set(refreshVersion$, (version) => {
-        return version + 1;
-      });
-    }
+    await setLoop(
+      async (loopSignal) => {
+        await delay(STATUS_REFRESH_INTERVAL_MS, { signal: loopSignal });
+        set(refreshVersion$, (version) => {
+          return version + 1;
+        });
+        return false;
+      },
+      0,
+      signal,
+      { retryTransientErrors: false },
+    );
   },
 );


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 19e19f793a440aebfbc934a811261b0a1ba706ba
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Replaces the hand-rolled `while (!signal.aborted)` polling loop in `pollInstatusIssues$` with the shared `setLoop` helper, keeping the 3-minute `delay` inside the loop body and the app-root `AbortSignal` as the owner. Behavior-preserving refactor of one platform signals file; no contract, schema, or UI change.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no race conditions or wrong API assumptions | Pass | `instatus-status.ts:141-151` — `setLoop` waits `interval` *after* each body call, so keeping `delay` first in the body reproduces the original delay-then-refresh order. First refresh still lands at t=+3min, not t=0. Body always returns `false`, so the loop still terminates only on abort. |
| `REVIEW.md` | Security (OWASP, secrets, auth) | N/A | Diff contains no credential, endpoint, auth, or user-input handling; only loop control flow changed. |
| `REVIEW.md` | Type safety; no `any` or overly broad types | Pass | `loopSignal` is inferred as `AbortSignal` from `setLoop`'s `loopBody` signature; no casts or `any` introduced. |
| `REVIEW.md` | Style and maintainability; no duplication, no WHAT-comments | Pass | Net +6 lines in one function; removes a bespoke loop in favor of the existing shared helper. No comments added. |
| `REVIEW.md` | Testing coverage for a `refactor:` — existing tests must still cover the code | Pass | `views/router/__tests__/instatus-status-notice.test.tsx` (5 tests) and `__tests__/platform-entrypoint.test.ts` (6 tests) still pass unchanged. See Testing for the pre-existing gap this PR neither creates nor widens. |
| `REVIEW.md` | Feature regression guard — replaced behavior needs regression coverage | Pass | No user-visible behavior is replaced: same interval, same ordering, same stop condition, same owner signal. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No `schema/**`, `api-contracts/**`, Zod response schema, or `jsonb().$type<>()` change. |
| `REVIEW.md` | Event sourcing / optimistic events | N/A | No persistent or optimistic event, event ID, projection, or reconciliation path is touched. |
| `docs/bad-smell.md` | Zero `any`; no lint/type suppressions | Pass | No `any`, `@ts-*`, `eslint-disable`, or `oxlint-disable` added. |
| `docs/bad-smell.md` | No unnecessary try/catch; fail fast | Pass | `retryTransientErrors: false` keeps the original semantics where a non-abort rejection propagates instead of being swallowed by `setLoop`'s fibonacci backoff. No try/catch added at this call site. |
| `docs/bad-smell.md` | Static imports only; no hardcoded config | Pass | `setLoop` added to the existing static `./utils.ts` import; `STATUS_REFRESH_INTERVAL_MS` unchanged. |
| `docs/fallback.md` | No new fallback; introduced fallbacks must be declared with surface, window, removal condition | N/A | The diff introduces no fallback. It adds no `??`/`||` default, no compatibility branch, no feature-switch gate, and no negative test. The pre-existing `parseIssueList` empty-collection comment at `instatus-status.ts:50-54` is outside the diff and unmodified. |
| `docs/effect.md` | Choose the execution boundary matching the owner and trigger | Pass | Recurring background work stays in a signals-layer `command` driven by the app-root lifecycle, not render, `onRef`, or a React effect. |
| `docs/effect.md` | Make cancellation explicit; polling must always have a parent lifecycle | Pass | Owner is unchanged: `views/main.tsx:71` passes `store.get(rootSignal$)`. The inner `delay` binds `loopSignal`, and `setLoop` forwards the same owner signal to the body, so there is exactly one abort path with a real owner. |
| `docs/effect.md` | No `detach()` inside the signals layer | Pass | The command still `await`s its work; `detach` remains only at the `views/main.tsx` daemon boundary. |
| `docs/cache.md` | RC-8 asymmetric resource lifecycles; long-running work needs an owner that aborts it | Pass | The only acquired resource is the `signal-timers` timer, released by the owner signal on abort. No listener, observer, or object URL is added. |
| `docs/cache.md` | RC-1 render purity / RC-2 render-time signal allocation | N/A | No React component or signal factory is touched; the change is entirely inside an existing package-scope command. |
| `docs/cache.md` | RC-3 unbounded lifetime caches | N/A | No module-level `Map`/`Set`/`WeakMap` or keyed cache added; `refreshVersion$` is a pre-existing counter. |
| `.claude/skills/ccstate/SKILL.md` | Every `AbortSignal` has a clear owner; a floating polling loop is a defect | Pass | `rootSignal$` is the parent for the whole loop, so it aborts on app teardown; the loop cannot outlive its owner. |
| `docs/testing.md` | Integration-first; refactors should survive on existing behavior tests | Pass | No test file changed. The existing behavior tests for this surface exercise fetch, parse, render, dismiss, and host gating and still pass. |

### Findings

#### Critical (P0)
- None.

#### High Priority (P1)
- None.

### Testing

- **Coverage: Adequate for a `refactor:`.** No test file changed, and the existing behavior tests for this surface pass unmodified: `views/router/__tests__/instatus-status-notice.test.tsx` (5 passed) and `__tests__/platform-entrypoint.test.ts` (6 passed).
- **Pre-existing gap, not introduced here:** no test asserts the recurring 3-minute refresh itself; existing tests cover only the initial fetch, parsing, rendering, dismissal, and preview-host gating. This gap is identical before and after the diff. Closing it would need either `vi.useFakeTimers()` (P1 under `docs/testing.md`) or a 3-minute real wait, so it is not a reasonable ask on this PR.
- **Timing under Vitest is intentionally preserved.** `setLoop` collapses only its own `interval` to a macrotask under `IN_VITEST`; keeping the real `delay` in the body means this app-root daemon still ticks on the real interval in tests rather than spinning toward `MAX_LOOP_COUNT_IN_TEST`. Passing the interval as `setLoop`'s third argument instead would have changed both production ordering and test behavior.
- **Conventions: Compliant.** No `vi.mock()` of internal paths, no `fetch` stubbing, no fake timers, no mock-call-count assertions introduced.

PR #31436, opened 2026-09-03
fix(e2e): recover stalled onboarding bootstrap

### LINKED ISSUE #31411: test(e2e): recover stalled authenticated app bootstrap during runner setup
## Problem and evidence

Runner E2E credential preparation signs fresh Clerk users into a real app preview and exercises paid onboarding. In job https://github.com/vm0-ai/vm0/actions/runs/33729095090/job/100565484322, `runner-real-codex` successfully rendered `/onboarding`, but the following hard navigation to `/onboarding/video-template?choice=video` stayed on the Platform global `Loading` skeleton until the 30-second heading wait timed out. The same PR passed in another run; 19 of 20 recently sampled runner-prepare jobs passed.

The current diagnostic boundary begins only after the video-template page reaches Stripe. This failure happened earlier, so CI retained neither a screenshot nor Clerk/app readiness state. The failed run proves an authenticated app-bootstrap stall at the second navigation, but it does not identify the exact Clerk, shared-database, or other bootstrap subphase.

## Corrected investigation

The issue was initially framed as a SharedWorker same-name/different-URL collision after Clerk development-JWT rotation. A focused native Chromium 148 regression kept the original worker alive and created a second worker with the same name and a different query URL. Chromium accepted it and did not throw `URLMismatchError`. That premise was falsified, and all proposed production SharedWorker changes were discarded before commit.

## Decision

Fix the observed E2E boundary:

- Give the idempotent legacy deep-link navigation one bounded recovery attempt.
- Retry only after a timeout when the browser is on the expected origin/path and the global bootstrap skeleton is still present.
- Preserve the existing 30-second tolerance on the final attempt and enrich a final timeout with the existing credential-safe Clerk/app readiness report.
- Extend checkout artifact capture around the complete paid-onboarding phase so a pre-Stripe failure retains a screenshot and readiness state.
- Keep runner credential concurrency at two; current evidence does not attribute the failure to concurrency.

Do not add production Platform recovery, workflow/test retries, fixed sleeps, skipped checks, or broader timeout changes.

## Acceptance criteria

- A first deep-link navigation that remains on the global bootstrap skeleton is retried once and can continue through the normal paid-onboarding flow.
- A rendered wrong page, wrong route, or non-timeout error is not retried.
- A final timeout reports Clerk client/session/organization state, skeleton state, document readiness, and origin/path without query parameters or credentials.
- Runner checkout diagnostics are captured for failures before Stripe as well as inside Stripe; diagnostic failure never replaces the primary error.
- Runner credential concurrency and all four credential targets remain unchanged.
- Focused real-browser coverage proves the retry gate and non-bootstrap failure behavior without fixed sleeps.
- Relevant E2E type checks, tests, workflow contract checks, and PR CI pass.

## Risks

- **Masking a product regression:** recovery is gated on the expected route still showing the global bootstrap skeleton; rendered route failures remain immediate.
- **Longer transient path:** only the stalled path receives a second navigation; the final attempt keeps the existing tolerance.
- **Sensitive diagnostics:** reuse the existing readiness capture, which records only origin/path and never URL queries or tokens.
- **Unknown subcause:** this change recovers and diagnoses the observed bootstrap boundary without claiming an unobserved root cause.

seven332:
## Problem

Runner credential preparation can intermittently reach the authenticated video-template deep link but remain on the Platform bootstrap skeleton until the heading wait times out. The existing diagnostic boundary starts after that page has rendered, so this failure leaves no screenshot or app readiness state.

## Solution

- retry the idempotent video-template navigation once only when the expected destination still shows the active global bootstrap skeleton
- preserve the existing 30-second tolerance on the final attempt and attach credential-safe Clerk/app state to the final timeout
- distinguish active, hidden, and removed bootstrap skeleton states so rendered route failures are never retried
- capture paid-onboarding diagnostics around the complete phase, including failures before Stripe
- add real Chromium fixtures for both recovery and rendered-page failure behavior

Runner credential concurrency, production Platform behavior, workflow retries, and workflow timeouts are unchanged.

## Validation

- cd e2e && pnpm test
- cd e2e && pnpm check-types
- check-file-size pre-commit hook
- current-head CI: Turbo, Crates, and Security required gates passed
- cli-e2e-03-runner-prepare generated all four credentials; all 12 runner shards and all Playwright lanes passed

Closes #31411

### DIFF
--- a/e2e/playwright/lib/clerk-readiness.test.ts
+++ b/e2e/playwright/lib/clerk-readiness.test.ts
@@ -12,7 +12,7 @@ import {
 } from "./clerk-readiness";
 
 const BOOTSTRAPPING_APP: ClerkReadinessState = {
-  bootstrapSkeletonPresent: true,
+  bootstrapSkeleton: "active",
   client: "absent",
   organizationId: null,
   readyState: "complete",
@@ -21,7 +21,7 @@ const BOOTSTRAPPING_APP: ClerkReadinessState = {
 };
 
 const SIGNED_OUT_CLIENT: ClerkReadinessState = {
-  bootstrapSkeletonPresent: false,
+  bootstrapSkeleton: "removed",
   client: "loaded",
   organizationId: null,
   readyState: "complete",
@@ -48,7 +48,7 @@ test("a stalled Clerk bootstrap is described by its observed state", async (cont
       state: BOOTSTRAPPING_APP,
     });
     assert.match(stalled, /client=absent/u);
-    assert.match(stalled, /bootstrapSkeleton=present/u);
+    assert.match(stalled, /bootstrapSkeleton=active/u);
 
     const signedOut = describeClerkReadiness({
       kind: "observed",
--- a/e2e/playwright/lib/clerk-readiness.ts
+++ b/e2e/playwright/lib/clerk-readiness.ts
@@ -10,7 +10,7 @@ import { errors } from "@playwright/test";
  * Recording the observed state at the timeout keeps those causes separable.
  */
 export interface ClerkReadinessState {
-  readonly bootstrapSkeletonPresent: boolean;
+  readonly bootstrapSkeleton: "active" | "hidden" | "removed";
   readonly client: "absent" | "loaded" | "loading";
   readonly organizationId: string | null;
   readonly readyState: string;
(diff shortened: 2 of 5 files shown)


seven332:
## Code Review: PR #31436

**Reviewed head:** `8dc933a9142ccb846730424089b889003bc9bd28`

### Summary

The change recovers only the observed idempotent E2E navigation stall, keeps rendered route failures visible, and broadens credential-safe diagnostics without changing production Platform behavior or runner concurrency.

### Findings

No findings.

### Testing

E2E fixture tests and type checking pass locally. Real Chromium covers both the active-skeleton recovery and the rendered-page no-retry path. The workflow contract script reached its Ruby phase, which could not run locally because this container has no Ruby; CI is the remaining verification.

### Verdict

LGTM

PR #31451, opened 2026-09-03
fix(platform): use adaptive okou favicon

hulh122:
## Summary

- Switch the product browser favicon to the published adaptive Okou SVG.
- Update the SHA-384 integrity value for the new immutable asset.
- Keep the app Worker regression coverage aligned for both static.vm0.io and static.okou.io.

## Scope

- Browser favicon only; Apple Touch, manifest/PWA, notification icons, and in-page branding are unchanged.
- The static asset dependency is already published by vm0-ai/static-files#182.
- Companion website PR: https://github.com/vm0-ai/vm0-marketing/pull/563
- The two code PRs can merge independently.

## Verification

- Prettier 3.8.1 check passed for both changed files.
- .github/scripts/tests/okou-app-worker-test.sh passed.
- Node syntax check passed for the Worker test.
- The checked-in SRI matches the published bytes.
- static.vm0.io and static.okou.io return identical asset bytes with SHA-256 b4eda9221bb709b38189061970131474b3323354536077f08011903c12cb583f.

PR #31453, opened 2026-09-03
fix(platform): hide account manager empty state for a single account

Lunarivibe:
## Problem

Two related defects in the connector account manager, both caused by the default account being pinned above the list.

1. Opening **Manage <connector> accounts** for a connector with exactly one account showed `No accounts found` directly beneath that connected account. The trailing empty state was keyed off `defaultConnection && others.length === 0`, which is also true when the connector simply has one account — its only account is the default, so `others` is empty with no search involved.
2. Even with the guard corrected, a search that matched nothing still rendered the pinned default row directly above `No accounts found`. Logically the message described the filtered list, but on screen it denied the existence of an account the user could see.

## Fix

A search now owns the whole list. The default is pinned only at rest — when a query is active, the rows are exactly what the query returned, and the default appears only if it matches. One empty state then covers every case, so the second, in-card message is gone.

This keeps the reason the pin exists (at rest the dialog shows which account new runs will use) while removing the state where a visible row and "nothing found" appear together.

## Verification

Live walkthrough on this PR's preview (`pr-31453-app-okou-app-preview.vm0.workers.dev`), with PR #31451's preview as the pre-fix build for comparison:

- One account, no search: account row only, no empty state.
- Seven accounts, no search: full list, search box appears at the threshold.
- Search matching nothing: only `No accounts found`, no orphaned pinned row.
- Search matching only the default: that row is shown as a result, no empty state.

Tests in `connectors-page.test.tsx`:

- New: a one-account manager renders no empty state (fails on the pre-fix component).
- New: a default account that a search matches stays a result row.
- Updated: the debounce test now asserts the default is unpinned during a search; confirmed load-bearing by re-pinning the row and watching it fail.
- `pnpm -F @okouai/app exec vitest run src/views/okou-page/__tests__/connectors-page.test.tsx` — 131 passed.
- `pnpm format`, platform `lint`, `tsc --noEmit`, `knip --workspace apps/platform` clean.

### DIFF
--- a/turbo/apps/platform/src/views/okou-page/__tests__/connectors-page.test.tsx
+++ b/turbo/apps/platform/src/views/okou-page/__tests__/connectors-page.test.tsx
@@ -1782,6 +1782,45 @@ describe("connectors page", () => {
     expect(within(dialog).queryByText("No accounts found")).toBeNull();
   });
 
+  it("keeps a default account that a search matches", async () => {
+    const accounts = mockGitHubConnectorAccounts(7);
+    const defaultAccount = accounts.find((account) => {
+      return account.isDefault;
+    });
+    if (!defaultAccount) {
+      throw new Error("Expected a default account fixture");
+    }
+    context.mocks.api(
+      connectorAccountsContract.connections,
+      ({ query, respond }) => {
+        return respond(200, {
+          connections: query.search ? [defaultAccount] : accounts,
+          nextCursor: null,
+        });
+      },
+    );
+    detachedSetupPage({
+      context,
+      path: "/connectors",
+      featureSwitches: { [FeatureSwitchKey.ConnectorAccounts]: true },
+    });
+
+    click(await waitForButtonByAriaLabel("Manage GitHub accounts"));
+    const dialog = await screen.findByRole("dialog", {
+      name: "Manage GitHub accounts",
+    });
+    fireEvent.input(within(dialog).getByPlaceholderText("Find accounts"), {
+      target: { value: "Unnamed" },
+    });
+    // The default is the only match here, so it stays a result row instead of
+    // being replaced by the empty state.
+    await waitFor(() => {
+      expect(within(dialog).queryByText("Work 1")).toBeNull();
+      expect(within(dialog).getByText("Unnamed account")).toBeInTheDocument();
+      expect(within(dialog).queryByText("No accounts found")).toBeNull();
+    });
+  });
+
   it("paginates a feature-on account manager", async () => {
     const accounts = mockGitHubConnectorAccounts(8);
     // The server projects rows away before slicing, so a page can hold fewer
@@ -1903,7 +1942,9 @@ describe("connectors page", () => {
         { search: "Work 2", cursor: null },
       ]);
       expect(within(dialog).getByText("Work 2")).toBeInTheDocument();
-      expect(within(dialog).getAllByText("Unnamed account")).toHaveLength(1);
+      // A search owns the whole list: the default account is unpinned unless it
+      // matches, so the results never mix with a row the query excluded.
+      expect(within(dialog).queryByText("Unnamed account")).toBeNull();
     });
 
     fireEvent.input(searchInput, {
@@ -1915,7 +1956,7 @@ describe("connectors page", () => {
         cursor: null,
       });
       expect(within(dialog).getByText("No accounts found")).toBeInTheDocument();
-      expect(within(dialog).getAllByText("Unnamed account")).toHaveLength(1);
+      expect(within(dialog).queryByText("Unnamed account")).toBeNull();
     });
   });
 
--- a/turbo/apps/platform/src/views/okou-page/__tests__/connectors-page.test.tsx
+++ b/turbo/apps/platform/src/views/okou-page/__tests__/connectors-page.test.tsx
@@ -1759,6 +1759,29 @@ describe("connectors page", () => {
     expect(within(dialog).getAllByText("Unnamed account")).toHaveLength(1);
   });
 
+  it("keeps the empty state out of a single-account manager", async () => {
+    const accounts = mockGitHubConnectorAccounts(1);
+    context.mocks.api(connectorAccountsContract.connections, ({ respond }) => {
+      return respond(200, { connections: accounts, nextCursor: null });
+    });
+    detachedSetupPage({
+      context,
+      path: "/connectors",
+      featureSwitches: { [FeatureSwitchKey.ConnectorAccounts]: true },
+    });
+
+    click(await waitForButtonByAriaLabel("Manage GitHub accounts"));
+    const dialog = await screen.findByRole("dialog", {
+      name: "Manage GitHub accounts",
+    });
+    // The sole account is also the pinned default, so it is the only row the
+    // dialog can draw. That is a full list, not an empty search result.
+    await waitFor(() => {
+      expect(within(dialog).getByText("Unnamed account")).toBeInTheDocument();
+    });
+    expect(within(dialog).queryByText("No accounts found")).toBeNull();
+  });
+
   it("paginates a feature-on account manager", async () => {
     const accounts = mockGitHubConnectorAccounts(8);
     // The server projects rows away before slicing, so a page can hold fewer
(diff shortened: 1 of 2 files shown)


Lunarivibe:
LGTM


PR_AUTO_REVIEW_COMMIT: a0d7adb160b7c2a1a7712532bb476564a685208f
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Fixes the connector account manager rendering `No accounts found` beneath a visible account. The default account is now pinned only at rest; an active search owns the whole row list, so the two empty-state branches collapse into one and the in-card message that fired for a single-account connector is gone. Scope is one presentational component plus its page tests, behind the non-GA `connectorAccounts` switch.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic, including edge cases and conditionals | Pass | `connector-account-manager-dialog.tsx:310-322`: all four states covered — no search + default, no search + no default, search with matches, search with none. `search.trim()` gives `""` falsy, so a whitespace-only query still takes the at-rest path rather than blanking the list. |
| `REVIEW.md` | Type safety, no `any` or broad types | Pass | `rows` carries an explicit `readonly ConnectorAccountConnection[]` annotation; no `any`, no assertions added. |
| `REVIEW.md` | Security (authz, injection, secrets) | N/A | Client-side render branch only; no new endpoint, credential, query, or user-supplied HTML path. |
| `REVIEW.md` | Style and maintainability, no duplicated logic | Pass | Net −6 lines in the component; the duplicated empty-state branch is removed and one `rows.length === 0` guard replaces both. |
| `REVIEW.md` | Comments explain WHY, not WHAT | Pass | `connector-account-manager-dialog.tsx:310-313` states why a search unpins the default; the test comments state the intent of each assertion. |
| `REVIEW.md` | Feature regression guard — changed existing behavior must have coverage and containment | Pass | The changed behavior (default unpinned during search) is asserted directly in the updated `debounces feature-on account manager searches` test at both the match and no-match steps. Containment: the whole surface is gated by `FeatureSwitchKey.ConnectorAccounts`, `enabled: false` + staff orgs on `main`. |
| `REVIEW.md` | `fix:` commits require a regression test | Pass | `keeps the empty state out of a single-account manager` reproduces the reported bug; it fails against the pre-fix component and passes after. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No schema, contract, Zod, or `jsonb().$type<>()` change; both changed files are platform view code. |
| `docs/bad-smell.md` | Zero lint/type suppressions | Pass | No `eslint-disable`, `@ts-ignore`, `@ts-expect-error`, or `prettier-ignore` in the diff. |
| `docs/bad-smell.md` | No unnecessary try/catch; fail fast | Pass | No error handling added; loading and `hasError` loadable branches are untouched. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | No imports changed. |
| `docs/bad-smell.md` | No hardcoded URLs or environment values | Pass | None present. |
| `docs/fallback.md` | Default is no fallback; no guard for a state the contract cannot produce | Pass | The diff adds no `??`/`||` chain and no compatibility branch. `defaultConnection ? … : others` guards a genuinely nullable value — the summaries endpoint has no default until a target has accounts — and is presentational branching, not a fallback. |
| `docs/fallback.md` | Every fallback introduced must be declared in the PR summary | N/A | The diff introduces no fallback, so no `Fallbacks` section is required. It removes a render branch rather than adding compatibility code. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | Closest call is `queryByText("No accounts found")).toBeNull()`. This is not a tombstone: it asserts current user-visible behavior of live code, and the same message is still asserted present in the no-match search test. |
| `docs/fallback.md` | Non-GA feature-switched features need no compatibility code | Pass | `connectorAccounts` is `enabled: false` with staff org hashes on `main`; the PR correctly adds no dual-read, migration, or old-shape tolerance. |
| `docs/effect.md` | React render stays pure — no store writes, I/O, subscriptions, or resource creation in render | Pass | `AccountsCard` only reads props and derives `rows`; the new `search` value is read by the parent via the existing `useGet(settingsConnectorAccounts.search$)` and passed down. |
| `docs/effect.md` | Signals owned outside render; work matched to its owner and trigger | Pass | No signal, command, effect, or ref added; the change is a pure derivation inside an existing component. |
| `docs/cache.md` | RC-1/RC-2 no render-phase side effects or render-time signal allocation | Pass | No signal factory call or side effect added in render. |
| `docs/cache.md` | RC-3/RC-4 no unbounded caches, state owned by the narrowest lifecycle | Pass | No module-level `Map`/`Set`/`WeakMap` or new state; `rows` is a per-render derivation. |
| `docs/cache.md` | RC-7/RC-8 stable refs and symmetric resource teardown | Pass | No refs, listeners, timers, observers, or async work added. |
| `.claude/skills/ccstate/SKILL.md` | View layer consumes loadable state; no view-level API error handling | Pass | The existing `hasError` / `loading` loadable branches are preserved unchanged; no `.catch()` or toast added. |
| `.claude/skills/ccstate/SKILL.md` | No manual cache duplicating `computed` memoization | Pass | No cache layer added. |
| `docs/testing.md` | Integration tests only, through the production bootstrap path | Pass | Both new tests enter via `detachedSetupPage({ context, path: "/connectors" })` and drive the real dialog. |
| `docs/testing.md` | Mock at the boundary only — no `vi.mock` of `../` internals, no db mocking | Pass | Only `context.mocks.api(connectorAccountsContract.connections, …)` is used; no `vi.mock`, no `globalThis.services.db` stub. |
| `docs/testing.md` | No direct `fetch` mocking; use the shared mock infrastructure | Pass | No `vi.stubGlobal("fetch", …)`; requests go through `context.mocks`. |
| `docs/testing.md` | No fake timers | Pass | No `vi.useFakeTimers()` / `advanceTimersByTime`; the debounce test drives real input events and awaits `waitFor`. |
| `docs/testing.md` | Test behavior, not mocks | Pass | Assertions are on rendered rows and the empty-state message, not on mock call counts. |
| `docs/testing/app-testing.md` | Use `waitFor()` for async assertions; override only what is needed | Pass | Both new tests wrap post-search assertions in `waitFor`; the `keeps a default account that a search matches` handler branches on `query.search` and falls through to the default fixture otherwise. |
| `docs/testing/app-testing.md` | No file-level `afterEach` cleanup; let `testContext()` own lifecycle | Pass | No cleanup hooks added; the tests reuse the file's existing `context`. |
| `docs/testing/patterns.md` | Platform page tests live under `views/**/__tests__` and use `detachedSetupPage`, not `server.use()` | Pass | Added to the existing `connectors-page.test.tsx`; no MSW `server` import. |
| `docs/testing/testing-external-behavior.md` | Construct and observe through the user-facing surface only | Pass | Setup is a click on `Manage GitHub accounts` and input into `Find accounts`; assertions are on dialog text. No store mutation, hook call, or internal-callback assertion. |
| `docs/testing/anti-patterns.md` | AP-8 no over-testing of trivial UI text or states | Pass | The asserted strings are the product behavior under repair — which message the dialog shows — not incidental copy. |
| `docs/testing/anti-patterns.md` | AP-10 no direct component rendering | Pass | No `render(<ConnectorAccountManagerDialog />)`; the component is reached through the page. |
| `docs/deployment-compatibility.md` | Independently deployed and persisted boundaries stay compatible | N/A | No API request/response, runner protocol, queue payload, schema, or persisted-state change. An old client keeps its own already-loaded render logic until refresh; the API contract is untouched in both directions. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

### Testing

- Coverage: Adequate. `fix:` commit carries a regression test that was confirmed to fail on the pre-fix component (`getByText` matched the leaked `No accounts found` node) and pass after. The second behavior change — unpinning the default during a search — is covered both positively (`keeps a default account that a search matches`) and negatively (updated `debounces feature-on account manager searches`, whose new `queryByText(...).toBeNull()` assertion was confirmed load-bearing by re-pinning the row and watching it fail).
- Conventions: Compliant. Page-level integration tests through `detachedSetupPage`, contract-level mocking through `context.mocks.api`, `waitFor` for async assertions, no internal mocks, no fake timers, no direct component rendering, no cleanup hooks.
- Local gates on this head: `connectors-page.test.tsx` 131 passed; `pnpm format`, platform `lint`, `tsc --noEmit`, and `knip --workspace apps/platform` clean. All GitHub checks on `a0d7adb` are green.

PR #31460, opened 2026-09-03
perf(mitm-addon): skip parsing canonical chat completion deltas

### LINKED ISSUE #31430: perf(python): avoid full parsing of usage-free Chat Completions SSE deltas
## Summary

The OpenAI-compatible Chat Completions SSE handler runs every data event through a complete `JsonSelectiveExtractor` cycle, including ordinary token-delta frames that contain neither usage nor failure evidence. The extractor is reused after #28625/#28629, but each event still pays the full reset/feed/finish cost before `_usage_snapshot()` can determine that there is nothing to record.

This work runs synchronously in the mitmproxy response-stream callback and scales with the number of streamed tokens/events rather than only with the small number of usage, error, and terminal frames.

## Evidence

- `crates/runner/mitm-addon/src/response_streaming.py:249-259` installs the Chat Completions SSE parser, and `crates/runner/mitm-addon/src/response_streaming.py:599-610` invokes it inline from `stream_and_observe()`.
- `crates/runner/mitm-addon/src/usage/openai_chat_completions.py:180-189` captures every named or eventless frame and feeds every data chunk to the selective extractor.
- `crates/runner/mitm-addon/src/usage/openai_chat_completions.py:199-206` finishes and resets that extractor for every event; only afterward do `:239-241` discard a complete event with no selected usage.
- The sampled integration case at `crates/runner/mitm-addon/tests/test_openai_chat_completions_usage.py:231-273` sends 2,560 ordinary usage-free deltas through this real response-stream path before final usage.
- The committed benchmark payload at `crates/runner/mitm-addon/scripts/benchmark_openai_chat_completions_usage.py:12-27` is the same ordinary usage-free Chat Completions shape and confirms that reset/reuse still performs one complete selective parse per event.

On `095302af7aabd36f7bba5af46346ae56079d23c7` with the repository's exact Python 3.14.0 runtime, the committed benchmark reported:

```text
10,000 fresh construction/feed/finish cycles: 0.820589 s
10,000 reused reset/feed/finish cycles: 0.509816 s
speedup: 1.61x
```

A seven-repeat end-to-end scanner benchmark used the integration test's 87-byte ordinary delta, appended one terminal usage event and `[DONE]`, and fed realistic 16 KiB chunks:

| Ordinary deltas | Encoded stream bytes | Median | Range |
| ---: | ---: | ---: | ---: |
| 2,560 | 222,847 | 159.4 ms | 157.0-164.8 ms |
| 10,000 | 870,127 | 637.6 ms | 607.3-673.1 ms |

Profiling confirms that selective JSON feeding dominates SSE framing. As a headroom check only, decoding and classifying the same bounded 85-byte delta with CPython's JSON decoder took 18.5 ms for 10,000 documents on the same runtime. That prototype is not a proposed unbounded buffering strategy, but it demonstrates that most of the current per-event CPU is avoidable for small canonical deltas.

## Impact

A long Chat Completions response commonly contains thousands of small token-delta events and only one usage-bearing terminal event. The current path therefore spends hundreds of milliseconds of aggregate synchronous CPU validating and projecting frames that cannot change accounting or failure state. Concurrent model streams multiply that work on the shared response-hook path and can delay forwarding of response chunks and unrelated proxy traffic.

The existing 65,536-work-unit limit bounds each individual parse, but it does not prevent the cost from being repeated for every ordinary event in a stream.

## Suggested direction

Add a conservative, bounded fast path that recognizes complete small canonical Chat Completions deltas with no top-level or first-choice `usage`/`error` candidate and avoids feeding those events through the full selective extractor. Preserve the existing extractor as the authoritative fallback for usage or failure candidates, unknown shapes, malformed or ambiguous input, events that exceed the fast-path bound, and inputs whose duplicate or escaped keys cannot be classified safely.

The OpenAI Responses SSE handler's bounded prefilter-plus-fallback structure is an existing repository pattern, although Chat Completions needs shape/key classification rather than a `type` event-name list.

## Acceptance criteria

- A canonical usage-free, error-free Chat Completions delta does not perform a complete `JsonSelectiveExtractor` feed/finish cycle.
- Top-level usage precedence, first-choice usage fallback, top-level and first-choice errors, `[DONE]`, named and eventless frames, split/multiline data, malformed-frame diagnostics, and later-event recovery remain unchanged.
- Duplicate and escaped JSON keys retain current accounting and failure-classification semantics.
- Unknown, ambiguous, malformed, or over-bound inputs fall back to the existing bounded extractor so the optimization cannot underbill or hide a provider failure.
- Add a deterministic regression test showing that a long ordinary-delta sequence performs no full extractor parse per delta while final usage is still reported.
- Extend or complement the committed benchmark with same-environment before/after evidence showing a material reduction in ordinary-delta CPU time; do not use a wall-clock CI threshold.

## Audit context

Found during the daily mitm-addon Python performance audit on commit `095302af7aabd36f7bba5af46346ae56079d23c7`. This is a residual full-parse cost after the schema-reuse work in #28625/#28629, not a reopening of that completed optimization.

seven332:
## Summary

- add a 1,024-byte bounded classifier for canonical usage-free Chat Completions SSE deltas
- skip the reusable selective extractor only for strict, unambiguous delta envelopes while preserving equivalent failure evidence
- replay candidate-bearing, unknown, malformed, ambiguous, and over-bound events through the existing authoritative extractor exactly once
- extend deterministic coverage and the committed microbenchmark for the complete SSE fast path

## Behavior and safety

The fast path requires strict UTF-8, complete JSON, unique keys, bounded nesting and number tokens, the canonical `chat.completion.chunk` envelope, and canonical choice shapes. Usage or error candidates, escaped candidate keys, duplicates, unknown shapes, malformed JSON, and larger events retain the existing extraction, accounting, diagnostics, and failure-classification behavior.

This change is limited to streaming Chat Completions event inspection. It does not change non-streaming JSON handling, provider wire formats, APIs, storage, migrations, or rollout configuration.

## Benchmark

Locked CPython 3.14.0 environment, 10,000 cycles per repeat, median of five repeats:

```text
fresh construction/feed/finish:             1.035419 s
reused reset/feed/finish:                    0.621923 s
bounded decode/classify:                     0.128644 s
complete SSE fast path:                      0.151129 s
reused full parse -> bounded classification: 4.83x
reused full parse -> complete SSE fast path:  4.12x
```

The benchmark is informational and has no wall-clock CI threshold.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `ruff format --check src/usage/openai_chat_completions.py tests/test_openai_chat_completions_usage.py scripts/benchmark_openai_chat_completions_usage.py`
- `ruff check src/usage/openai_chat_completions.py tests/test_openai_chat_completions_usage.py scripts/benchmark_openai_chat_completions_usage.py`
- `basedpyright -p .`
- `python -m pytest tests/test_openai_chat_completions_usage.py -q` (31 passed)
- `python -m pytest tests/test_model_provider_failure_reporting.py -q` (100 passed)
- `python -m pytest -q` (7,327 passed)
- `PYTHONPATH=src python scripts/benchmark_openai_chat_completions_usage.py`

Closes #31430

### DIFF
--- a/crates/runner/mitm-addon/scripts/README.md
+++ b/crates/runner/mitm-addon/scripts/README.md
@@ -54,7 +54,7 @@ before committing the related addon changes.
 
 # Chat Completions SSE fast-path benchmark
 
-The [Chat Completions extractor benchmark](benchmark_openai_chat_completions_usage.py)
+The [Chat Completions SSE fast-path benchmark](benchmark_openai_chat_completions_usage.py)
 compares fresh and reused selective JSON extraction with the bounded canonical-delta classifier
 and the complete SSE fast path. Run it from the repository root with the addon's locked uv
 environment:
--- a/crates/runner/mitm-addon/scripts/README.md
+++ b/crates/runner/mitm-addon/scripts/README.md
@@ -52,11 +52,11 @@ When adding or renaming a shared metadata key, update the registry and use its
 `metadata_keys` constants instead of literal shared keys, then run this check
 before committing the related addon changes.
 
-# Chat Completions extractor benchmark
+# Chat Completions SSE fast-path benchmark
 
 The [Chat Completions extractor benchmark](benchmark_openai_chat_completions_usage.py)
-compares the cost of constructing a selective JSON extractor for each event with the cost of
-resetting and reusing one extractor. Run it from the repository root with the addon's locked uv
+compares fresh and reused selective JSON extraction with the bounded canonical-delta classifier
+and the complete SSE fast path. Run it from the repository root with the addon's locked uv
 environment:
 
 ```bash
@@ -67,14 +67,23 @@ PYTHONPATH=src uv run --no-sync python scripts/benchmark_openai_chat_completions
 The benchmark uses this representative usage-free Chat Completions delta payload:
 
 ```json
-{"id":"chatcmpl_1","object":"chat.completion.chunk","choices":[{"delta":{"content":"x"}}]}
+{"id":"chatcmpl_1","object":"chat.completion.chunk","choices":[{"index":0,"delta":{"content":"x"}}]}
 ```
 
 Each fresh cycle constructs an extractor, feeds the payload, and finishes the document. Each reused
-cycle resets the existing extractor before feeding and finishing the same payload. The script runs
-10,000 cycles per repeat for five repeats, reports the median duration for each path, and reports a
-speedup ratio calculated as fresh duration divided by reused duration.
+cycle resets the existing extractor before feeding and finishing the same payload. The bounded
+classification cycle validates and classifies the complete payload without selective extraction.
+The complete fast-path cycle sends the corresponding `data:` event through a reused
+`SseUsageScanner`, including SSE framing and provider-handler work, and verifies before timing that
+the payload is eligible for the fast path.
+
+The script runs 10,000 cycles per repeat for five repeats, reports the median duration for each
+path, and reports construction-reuse, full-parse-to-classification, and
+full-parse-to-complete-fast-path ratios. The bounded classifier retains at most 1,024 event bytes;
+unknown, ambiguous, candidate-bearing, malformed, or larger events use the authoritative selective
+extractor and are outside the optimized benchmark path.
 
 The timings are informational and machine-dependent microbenchmark results, not an end-to-end
-parser benchmark or a fixed CI pass/fail threshold. For before-and-after comparisons, use the same
-Python and dependency environment, hardware, and similar system conditions where possible.
+response latency measurement or a fixed CI pass/fail threshold. For before-and-after comparisons,
+use the same Python and dependency environment, hardware, and similar system conditions where
+possible.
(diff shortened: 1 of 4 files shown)


seven332:
## Code Review: PR #31460

**Reviewed head:** `64db9b0db636f6572c8a459346982759f64c9137`

### Summary

The bounded classifier admits only strict canonical Chat Completions deltas, preserves equivalent no-error evidence, and sends every candidate-bearing, unknown, ambiguous, malformed, or over-bound event through the existing authoritative extractor. The event buffer is capped, fallback bytes are replayed once, and the committed benchmark shows a material reduction in the ordinary-delta path.

### Findings

No findings.

### Testing

Reviewed the two submitted commits and the real SSE/response-stream regression coverage. `uv lock --check`, `uv sync --locked`, full Ruff format/lint, basedpyright, 31 focused Chat Completions tests, 100 shared failure-reporting tests, and the complete 7,327-test addon suite passed. The informational benchmark measured the complete SSE fast path at 4.12x faster than reused full parsing in the PR-recorded run. PR CI is assessed separately after this review.

### Verdict

LGTM

PR #31508, opened 2026-09-03
revert: "feat(ui): refresh the app palette onto the sand and amber scales"

Lunarivibe:
Reverts #31081.

That PR shipped two things: the Sand/Amber palette, ungated, and a new shell
(workspace card, sidebar greys, composer focus ring) behind a `newUi` feature
switch scoped to `ming@vm0.ai`.

The palette was not gated because the base tokens and the brand hex also reach
the logo SVG, the boot skeleton in `apps/platform/index.html`, and the
automation-result email — surfaces that render before the stylesheet loads or
outside the browser entirely. The conclusion drawn at the time was that the
palette therefore had to go GA with the shell.

It does not. Scoping the token blocks under `[data-new-ui]` on the document
element works for everything the stylesheet owns, because custom properties
inherit; only the three surfaces above stay on the current brand. That is the
right shape, and it is a large enough change to belong in its own PR rather
than as a patch on top of a live palette swap.

Reverting first so `main` is not carrying an ungated palette while that work
lands.

## What this restores

- `--gray-*` back to the cool neutral scale, `--primary-*` back to orange with
  the brand at stop 700 (`#ed4e01`)
- the product mark back to `#ed4e01`, and the automation-result email badge
  back to `#ed4e01` on white
- `apps/desktop`'s stylesheet back to its previous neutrals and brand
- the `newUi` switch, the workspace card, the sidebar greys, and their tests —
  all of it returns in the follow-up, gated together with the palette

## Verification

Clean inverse of the merge commit: 54 files, +362 / −678, exactly mirroring the
original +678 / −362. No manual conflict resolution was needed.


Lunarivibe:
LGTM


PR_AUTO_REVIEW_COMMIT: 8ea25ad9b6423cb22677fd1579a82a7cdffb641c
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Clean revert of #31081. `git revert -m 1` applied with no manual conflict resolution against a `main` that had moved six commits: 54 files, +362 / −678, the exact inverse of the original +678 / −362.

Reverting rather than patching because #31081 put the Sand/Amber palette on `main` ungated for everyone while only the shell sat behind `newUi`. The follow-up puts both behind the switch, and that is a large enough change that it should not land as a patch on top of a live palette swap.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness — the revert restores the prior state exactly | Pass | Set-compared the 54 files #31081 touched against the files still differing from `eff8671a60~1`: zero overlap. Every file the PR changed is back to its pre-PR content; the 38 files that still differ are entirely from the six unrelated `main` commits landed since. |
| `REVIEW.md` | Feature regression guard — nothing is left half-reverted | Pass | The `newUi` switch, its registry entry, `applyNewUiDocumentAttribute`, the `zero-workspace-card` CSS, the sidebar gating, and `new-ui-feature-switch.test.tsx` are all removed together. No orphan enum member, no CSS referencing a deleted token, no test for deleted behaviour. |
| `REVIEW.md` | PR summary describes the diff | Pass | States what is restored, why the revert precedes the re-land, and that the shell work returns gated in the follow-up. |
| `docs/fallback.md` | Removing a feature switch deletes the switch, its branch, and its tests together | Pass | All three go in this commit. No tombstone test asserting the reverted behaviour stays gone. |
| `docs/fallback.md` | No new fallback introduced | Pass | A revert introduces none. |
| `docs/testing.md` | Tests match the code they cover | Pass | `globals.test.ts` returns to pinning the literal focus-ring stop, which is correct for the restored palette. `design-token-shadowing.test.ts` returns to the single-entry allow-list, matching the restored `--color-sidebar-rail`-free sheet. |
| `docs/deployment-compatibility.md` | Safe across the skew window | Pass | Purely presentational. An old web client (up to ~2 days stale) keeps whatever CSS bundle it loaded and is unaffected; a refresh loads the restored palette. No schema, migration, queue payload, or runner protocol touched. `main` only deploys staging-app, so no production client sees a mid-flight change. |
| `docs/bad-smell.md` | No `any`, suppressions, or hardcoded configuration | Pass | Revert only. |

### Findings

None.

### Testing

- **Coverage: Adequate.** A revert needs no new tests; the restored suite is the one that covered this code before #31081.
- **Conventions: Compliant.**
- **CI:** 78 passing, 0 failing on `8ea25ad9`.

PR #31521, opened 2026-09-03
feat(ui): put the sand and amber palette behind the new-ui switch

Lunarivibe:
Re-lands #31081 (reverted in #31508) with the palette gated instead of GA.

The original PR shipped the Sand/Amber palette to everyone and put only the shell behind `newUi`. The reasoning was that three surfaces — the boot skeleton, the automation-result email, and `apps/desktop` — cannot read a per-user switch, so the palette had to go GA with them. That was wrong: those three are a small corner, and everything the stylesheet owns can be scoped.

## How it is gated

`packages/ui/src/styles/globals.css` keeps the orange palette on `:root` and `[data-theme="dark"]`, and adds `:root[data-new-ui]` / `[data-theme="dark"][data-new-ui]` twins carrying Sand and Amber. Both selectors match the same `<html>` and win on specificity, so every `@theme` alias follows without being restated — each is a lazy `hsl(var(--x))` reference resolved per element, not a snapshot.

That single mechanism covers the whole base palette. What is left in the platform sheet is the handful of values it hardcodes rather than derives — nine chart segment colours, three shadows, four markdown link colours — plus the two colour-theme ladders, which carry their own selector and so get their own gated copy. All of it sits in one clearly-marked block at the end of `index.css`, to be deleted in one pass when the palette graduates.

## Why this is not 40 files of conditionals

The Amber scale introduced four semantic tokens — `--brand-subtle`, `--brand-text`, `--brand-text-hover`, `--sidebar-rail` — and about forty component files moved onto them. Those files are untouched here. The tokens are bridged onto the orange palette as well (`--brand-text: var(--primary-700)`, `--sidebar-rail: var(--gray-50)`, and so on), so every call site that switched renders identically with the switch off. The indirection is an improvement on its own terms and stays either way.

## What still cannot follow

| Surface | Why | With the switch on |
| --- | --- | --- |
| boot skeleton in `index.html` | renders before the stylesheet loads | brief `#faf5f3` flash, then the new canvas |
| automation-result email | rendered server-side for an arbitrary recipient | badge stays `#ed4e01` on white |
| `apps/desktop` | self-contained stylesheet, no switch plumbing | stays on its current neutrals and brand |

The logo **can** follow — `product-brand-mark.tsx` is a component, so it reads `featureSwitch$` directly and picks its fill.

## The one place the off state is not identical

Sixty-eight call sites moved from `text-primary` / `bg-primary` onto the brand
tokens. Sixty-four are byte-identical with the switch off, because the orange
palette's `--primary` *is* `--primary-700`, which is exactly what the bridge
maps `--brand-text` to.

Five are not. Four icons and one link used `text-primary-900` / `text-primary-950`
directly rather than `text-primary`, and the single bridge token cannot reproduce
three different stops at once:

| | before (off) | after (off) |
| --- | --- | --- |
| light | `#D03200` / `#5C2918` | `#ED4E01` |
| dark | `#FF946E` / `#FED5C7` | `#ED4E01` |

`#ED4E01` is the colour their sixty-four siblings already render in both themes
today, so this normalises five stragglers onto the established brand text colour
rather than introducing a new one. It is a visible change for everyone, which is
why it is called out here rather than left to be discovered.

## Verification

Measured on the real built bundle (`vite build`, `index-B6873uiQ.css`), all four states, by toggling `data-theme` and `data-new-ui` on `<html>`:

| | page | brand text | brand fill | nav rail | chat column | card |
| --- | --- | --- | --- | --- | --- | --- |
| off, light | `rgb(250,251,252)` | `#ED4E01` | `#ED4E01` | `rgb(242,244,247)` | `rgb(249,250,251)` | none |
| on, light | `rgb(255,255,255)` | `rgb(136,86,0)` | `#FFA500` | `#EFEAE8` | `#FDFAF9` | 8px / 12px |
| off, dark | `rgb(25,25,26)` | `#ED4E01` | `#ED4E01` | `rgb(25,25,26)` | `rgb(19,19,21)` | none |
| on, dark | `rgb(36,35,33)` | `#FFA500` | `#FFA500` | `rgb(28,27,26)` | `rgb(49,48,46)` | 8px / 12px |

Every off-state value matches what `main` ships today.

- `pnpm -F @okouai/ui exec vitest run` — 95 passed
- `new-ui-feature-switch.test.tsx` + `design-token-shadowing.test.ts` — 7 passed
- `tsc --noEmit` clean on `@okouai/app` and `@okouai/ui`
- Confirmed the single bundle carries both scales: `--primary-300` and `--gray-0` each appear with four distinct values (orange light/dark, Amber light/dark)

## Fallbacks

None. The off branch is `main`'s live behaviour held behind a rollout gate, not a compatibility path; `docs/fallback.md` §2 covers the gated side. Removal condition: when the palette graduates, delete the `[data-new-ui]` blocks in both sheets, the four bridge tokens, and the switch together.

### DIFF
--- a/turbo/apps/platform/src/signals/okou-page/tiptap-workflow-composer.ts
+++ b/turbo/apps/platform/src/signals/okou-page/tiptap-workflow-composer.ts
@@ -170,7 +170,7 @@ function editorContentClass(singleLineOnMobile: boolean): string {
     : `${EDITOR_CONTENT_CLASS} min-h-[96px]`;
 }
 
-const WORKFLOW_HIGHLIGHT_CLASS = "text-primary";
+const WORKFLOW_HIGHLIGHT_CLASS = "text-brand-text";
 function composerPlaceholder(): string {
   return i18n.t(($) => {
     return $.chat.composer.placeholder;
--- a/turbo/apps/platform/src/signals/theme.ts
+++ b/turbo/apps/platform/src/signals/theme.ts
@@ -215,6 +215,22 @@ export function applyTypefaceDocumentAttribute(enabled: boolean) {
 }
 
 /**
+ * Keep the new shell attribute on the document while the app shell is mounted.
+ * Document scope matches the two above: the shell's surfaces are read by the
+ * sidebars and the workspace card, and portaled dialogs inherit the same
+ * sidebar token.
+ */
+export function applyNewUiDocumentAttribute(enabled: boolean) {
+  const root = document.documentElement;
+
+  if (enabled) {
+    root.dataset.newUi = "";
+  } else {
+    delete root.dataset.newUi;
+  }
+}
+
+/**
  * Initialize theme from localStorage or system preference.
  */
 export const initTheme$ = command(({ get, set }) => {
--- a/turbo/apps/platform/src/views/agents-page/agents-page-tabs.tsx
+++ b/turbo/apps/platform/src/views/agents-page/agents-page-tabs.tsx
@@ -738,7 +738,7 @@ function AgentCard({ agent, creator, hasUnread, showCreator }: AgentProps) {
                       // Matches --zero-card-shadow; inlined because the tooltip
                       // portal renders outside .zero-app where the var is scoped.
                       boxShadow:
-                        "0 2px 12px hsl(220 12% 50% / 0.04), 0 0 0 0.5px hsl(220 12% 50% / 0.02)",
+                        "0 2px 12px hsl(30 6% 45% / 0.05), 0 0 0 0.5px hsl(30 6% 45% / 0.025)",
                       whiteSpace: "normal",
                     }}
                   >
--- a/turbo/apps/platform/src/views/components/avatar.tsx
+++ b/turbo/apps/platform/src/views/components/avatar.tsx
@@ -108,7 +108,7 @@ export function WorkspaceLogo({
     <div
       className={cn(
         base,
-        "flex items-center justify-center bg-[hsl(var(--gray-200))] font-bold text-[hsl(var(--primary-700))]",
+        "flex items-center justify-center bg-[hsl(var(--gray-200))] font-bold text-brand-text",
         THUMBNAIL_INITIAL_TEXT[size],
       )}
     >
--- a/turbo/apps/platform/src/views/components/instatus-status-notice.tsx
+++ b/turbo/apps/platform/src/views/components/instatus-status-notice.tsx
@@ -79,7 +79,7 @@ function InstatusIssueNotice({
       <CardContent className={compact ? "p-3" : "p-4"}>
         <div className="flex items-start gap-3">
           <span
-            className={`${compact ? "size-8" : "size-9"} flex shrink-0 items-center justify-center rounded-lg bg-gray-50 text-primary`}
+            className={`${compact ? "size-8" : "size-9"} flex shrink-0 items-center justify-center rounded-lg bg-gray-50 text-brand-text`}
           >
             <StatusIcon aria-hidden="true" size={18} />
           </span>
@@ -98,7 +98,7 @@ function InstatusIssueNotice({
               href={`${pageBaseUrl}/${encodeURIComponent(issue.id)}`}
               target="_blank"
               rel="noreferrer"
-              className="mt-3 inline-flex items-center gap-1 text-xs font-medium text-primary-900 transition-colors hover:text-primary-950"
+              className="mt-3 inline-flex items-center gap-1 text-xs font-medium text-brand-text transition-colors hover:text-brand-text-hover"
             >
               {t(($) => {
                 return $.serviceStatus.viewUpdates;
--- a/turbo/apps/platform/src/views/components/product-brand-mark.tsx
+++ b/turbo/apps/platform/src/views/components/product-brand-mark.tsx
@@ -1,4 +1,5 @@
 import { useGet } from "ccstate-react";
+import { FeatureSwitchKey } from "@okouai/core/feature-switch-key";
 
 import {
   platformOkouWordmarkDarkImg,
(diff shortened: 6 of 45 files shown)


Lunarivibe:
LGTM


PR_AUTO_REVIEW_COMMIT: d14ff3fe5765868df1a54a1ec3f8c0fbbcf3d84e
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Re-lands #31081 (reverted in #31508) with the Sand/Amber palette gated on `newUi` rather than GA. The mechanism is one pair of `[data-new-ui]` token blocks in the design system that match the same `<html>` and win on specificity; every `@theme` alias follows because each is a lazy `hsl(var(--x))` reference resolved per element, so the whole base palette switches without a single downstream conditional.

The judgement worth checking here is not the colours — those were reviewed on #31081 — but whether the switch-off state still equals what `main` ships. It does everywhere except five call sites, which the PR body now declares explicitly.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness — the gating actually gates | Pass | Measured on the real built bundle (`index-CTovTs96.css`) across all four `data-theme` × `data-new-ui` combinations. Off: page `rgb(250,251,252)`, brand `#ED4E01`, `--color-accent-fg: hsl(20 99% 47%)`, no card. On: page white/Ink, brand `rgb(136,86,0)` light and `#FFA500` dark, `--color-accent-fg` Amber, card 8px/12px. Confirmed the single bundle carries four distinct values each for `--primary-300` and `--gray-0`. |
| `REVIEW.md` | Feature regression guard — the replaced flow keeps its behaviour | Finding-free, with one declared delta | Audited every ungated line that differs from `main`: exactly three, all behaviour-preserving indirections (see below). The five call sites that do shift are documented in the PR body with before/after values in both themes. |
| `REVIEW.md` | Ungated changes are equivalent | Pass | (1) `--color-sidebar` moves from `@theme` to `.zero-app`; every `bg-sidebar` consumer (`sidebar.tsx` ×3, `directed-shared.tsx`) is inside `.zero-app`, and the later `:where([data-theme="dark"]) .zero-nav` override still wins on source order, so dark still resolves to `hsl(240 5% 8%)`. (2) The `.zero-nav` copy folds into the same block. (3) The composer border moves to `--zero-composer-ring`, whose ungated default is `var(--gray-500)` — the literal it replaced. |
| `REVIEW.md` | No raw stop changes meaning across palettes | Pass | Diffed every `(text\|bg\|border\|ring)-primary-N` class in the diff: four removals, zero additions. Nothing reads a raw stop whose semantics flipped between the mirrored orange ramp and the unmirrored Amber one — which is the failure mode this indirection exists to prevent. |
| `REVIEW.md` | New user-facing behaviour gated behind a `FeatureSwitchKey` | Pass | `FeatureSwitchKey.NewUi`, `enabled: false`, `enabledEmailHashes: ["54757055"]`. The logo reads the switch directly because an SVG fill cannot follow a token. |
| `REVIEW.md` | `feat:` commits carry integration tests | Pass | `new-ui-feature-switch.test.tsx` (5) covers the document attribute in both shells and the chat column divider on each side; `design-token-shadowing.test.ts` (2) registers both sidebar shadows with reasons. |
| `docs/fallback.md` | Every fallback the diff introduces is declared, with surface, window and removal condition | Pass | The four bridge tokens (`--brand-subtle`, `--brand-text`, `--brand-text-hover`, `--sidebar-rail`) are the only compatibility construct. They are rollout-scoped, not cross-version: no deploy skew is involved, only a per-user switch. Declared in the PR body with the removal condition — delete the `[data-new-ui]` blocks, the bridges and the switch together when the palette graduates. |
| `docs/fallback.md` | Features behind a switch need no compatibility code (§2) | Pass | No dual-read, no tolerance branch, no migration. The off branch is `main`'s live behaviour, not a legacy shape. |
| `docs/fallback.md` | No negative tests against removed behaviour | Pass | None added. |
| `docs/bad-smell.md` | No `any`, suppressions, or hardcoded configuration | Pass | The only new TypeScript is `brandFill` in `product-brand-mark.tsx`, typed by inference from two string literals. |
| `docs/bad-smell.md` | Fail fast — no defaulting that hides misconfiguration | Pass | `useGet(featureSwitch$)[FeatureSwitchKey.NewUi]` reads an optional map key; falsy means off, which is the intended default, not a masked error. |
| `docs/testing.md` | Integration only; boundary mocks; behaviour not mocks | Pass | No new mocks. Tests render through `detachedSetupPage` and read computed output. |
| `docs/effect.md` | Render is pure; state is derived not synchronised | Pass | `applyNewUiDocumentAttribute` runs in a committed ref callback alongside its two existing siblings, symmetric on detach. `brandFill` is derived at read time. |
| `docs/deployment-compatibility.md` | Safe across the skew window | Pass | Presentational plus one enum member, which only widens `effectiveSwitches`. An old web client (up to ~2 days stale) ignores the unknown `newUi` key; a new client against an old API reads `undefined` and falls to the off branch. No schema, migration, queue payload or runner protocol touched. |

### Findings

None blocking.

One thing a reader should not have to discover for themselves, and which the PR body now states: five of sixty-eight brand-token call sites are **not** byte-identical with the switch off. Four icons and one link used `text-primary-900` / `text-primary-950` directly rather than `text-primary`, and a single bridge token cannot reproduce three different stops. They move to `#ED4E01` in both themes — the colour their sixty-four siblings already render today — so this normalises stragglers onto the established brand text colour rather than inventing one. Visible, small, and declared.

### Testing

- **Coverage: Adequate.** 7 targeted tests pass; `@okouai/ui` 95 pass; `tsc --noEmit` clean on `@okouai/app` and `@okouai/ui`.
- **Conventions: Compliant.**
- **CI:** 76 passing, 0 failing on `d14ff3fe`.

PR #31537, opened 2026-09-03
fix(ui): rebuild new-ui neutral gray palette

Lunarivibe:
## Summary

- restore `#FAF5F3` as the exact Linen anchor at light `gray-100` and dark `gray-950`
- rebuild the gated new-UI gray ramps in OKLab, tapering chroma toward neutral while preserving the existing `#242321` Ink canvas
- rebalance neutral surfaces and semantic text roles around the new ramp; the switch-off palette remains unchanged
- add regression coverage for exact anchors, monotonic luminance, neutrality, and WCAG contrast tiers

## Accessibility

- light `gray-600` clears 3:1 and `gray-700` clears 4.5:1 against the darkest light surface
- dark `gray-600` clears 3:1 and semantic muted text at `gray-800` clears 5.3:1 against the lightest common dark surface
- foreground anchors clear AAA, and every semantic text/background pair covered by the theme clears WCAG AA

## Verification

- `pnpm exec prettier --check packages/ui/src/styles/globals.css packages/ui/src/styles/__tests__/globals.test.ts`
- `pnpm --filter @okouai/ui lint`
- `pnpm --filter @okouai/ui check-types`
- `pnpm --filter @okouai/ui exec vitest run src/styles/__tests__/globals.test.ts` (23 tests)
- production Vite build for `@okouai/app`
- browser audit against the compiled CSS for light/dark resolved token values and contrast ratios

PR #31565, opened 2026-09-04
fix(ui): keep the new-ui workspace card's bottom corners rounded

Lunarivibe:
## Problem

Under the `newUi` switch the workspace card's two **bottom** corners render as hard 90° angles with no border, while the top two are correctly rounded.

| before (bottom-right, dark) | before (bottom-right, light) | after |
|---|---|---|
|  |  |  |

Reproduced at 1440×900 on a chat thread page, both themes. Not visible on the chat home page, where the composer is not flush with the bottom of the content box.

## Cause

The frame is painted by `.zero-workspace-card.zero-workspace-bg::before`, and `.zero-workspace-bg::before` carries `z-index: -1` — so it sits *behind* content. That is sufficient only while the corner triangles stay transparent.

`ChatThreadComposer`'s `<footer>` (`chat-thread-page.tsx`) is `bg-[hsl(var(--background))]` with `border-radius: 0`, and measures `[376, 714, 1432, 892]` against a card content box of `[376, 8, 1432, 892]` — its right and bottom edges coincide exactly. It therefore paints over both bottom corners and the bottom border. `shared-thread-page.tsx` has the same shape.

The existing comment stated the invariant that broke: *"nothing lays out in the gutter, and only the corner triangles could ever escape the radius."*

## Fix

Draw the frame twice: `::before` keeps filling behind content, and a new `::after` restates the border and repaints the gutter *over* content, with `pointer-events: none`.

The `box-shadow` spread equals `--zero-workspace-card-gap`, which is exactly the region the card's own padding already reserves — so it can only ever cover the triangles the radius cuts away, never real content. Both pseudo-elements now share a `--zero-workspace-card-radius` variable so the two radii cannot drift.

This fixes the whole class of problem rather than the chat footer alone, so other routes with a flush opaque last child are covered too. Clipping the card would also work, but would clip every popover that has to escape it.

## Verification

- Injected the exact final rule into the live PR preview: both bottom corners rounded in light and dark ([bottom-left](https://cdn.vm0.io/artifacts/dkx8rbtt91.png)), top corners unchanged, chat home unchanged.
- Hit-tested the frame region and typed into the composer to confirm `pointer-events: none` keeps interaction intact.
- Added `workspace-card-frame.test.ts`, following the existing source-text CSS test pattern. Confirmed it **fails** (3/3) with the `::after` block removed and passes with it.

Scope: `newUi` only, desktop only (`min-width: 48rem`). Unrelated to #31537.

### DIFF
--- a/turbo/apps/platform/src/views/css/__tests__/workspace-card-frame.test.ts
+++ b/turbo/apps/platform/src/views/css/__tests__/workspace-card-frame.test.ts
@@ -62,4 +62,11 @@ describe("new UI workspace card frame", () => {
       );
     }
   });
+
+  // Two stacked 0.7px strokes read heavier than one, and the fill's stroke only
+  // survives where content is transparent -- so sharing the border between the
+  // two made the top corners half again as heavy as the bottom ones.
+  it("leaves the border to the frame alone so no edge is drawn twice", () => {
+    expect(readRuleBody(FILL)).not.toMatch(/border:/);
+  });
 });
--- a/turbo/apps/platform/src/views/css/__tests__/workspace-card-frame.test.ts
+++ b/turbo/apps/platform/src/views/css/__tests__/workspace-card-frame.test.ts
@@ -1,0 +1,65 @@
+import { describe, expect, it } from "vitest";
+
+import { readGlobalCss } from "./global-css.ts";
+
+const globalCss = readGlobalCss();
+
+/** Returns the declaration body of the first rule introduced by `selector`. */
+function readRuleBody(selector: string): string {
+  const start = globalCss.indexOf(selector);
+  if (start === -1) {
+    throw new Error(`Unable to locate CSS rule for ${selector}`);
+  }
+  const open = globalCss.indexOf("{", start);
+  const close = globalCss.indexOf("}", open);
+  if (close === -1) {
+    throw new Error(`Unterminated CSS rule for ${selector}`);
+  }
+  return globalCss.slice(open + 1, close);
+}
+
+const CARD = "[data-new-ui] .zero-workspace-card";
+const FILL = `${CARD}.zero-workspace-bg::before`;
+const FRAME = `${CARD}.zero-workspace-bg::after`;
+
+// The fill paints behind content, so on its own the card's rounded corners
+// survive only where content is transparent. A page whose last child is flush
+// with the content box and carries its own opaque background -- the chat
+// composer footer is one -- squared off the two bottom corners and took the
+// bottom border with them.
+describe("new UI workspace card frame", () => {
+  it("restates the border over content so flush children cannot bury it", () => {
+    const frame = readRuleBody(FRAME);
+
+    expect(frame).toMatch(/border:\s*0\.7px solid hsl\(var\(--border\)\);/);
+    expect(frame).toMatch(
+      /border-radius:\s*var\(--zero-workspace-card-radius\);/,
+    );
+    // Without this the frame would swallow clicks across the whole card.
+    expect(frame).toMatch(/pointer-events:\s*none;/);
+  });
+
+  it("repaints only the gutter the card's own padding already reserves", () => {
+    const card = readRuleBody(CARD);
+    const frame = readRuleBody(FRAME);
+
+    // Spread equal to the gap is what bounds the overpaint to the corner
+    // triangles the radius cuts away; a larger spread would eat real content.
+    expect(frame).toMatch(
+      /box-shadow:\s*0 0 0 var\(--zero-workspace-card-gap\) hsl\(var\(--sidebar\)\);/,
+    );
+    expect(frame).toMatch(/inset:\s*var\(--zero-workspace-card-gap\);/);
+    expect(card).toMatch(/padding:\s*var\(--zero-workspace-card-gap\);/);
+  });
+
+  it("keeps the fill and the frame on one radius", () => {
+    expect(readRuleBody(CARD)).toMatch(
+      /--zero-workspace-card-radius:\s*\d+px;/,
+    );
+    for (const selector of [FILL, FRAME]) {
+      expect(readRuleBody(selector)).toMatch(
+        /border-radius:\s*var\(--zero-workspace-card-radius\);/,
+      );
+    }
+  });
+});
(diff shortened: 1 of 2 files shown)


Lunarivibe:
LGTM


PR_AUTO_REVIEW_COMMIT: 8e963714c13f91d4e824b68a088cc8806eb29df9
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Splits the new-ui workspace card frame across two pseudo-elements so the two bottom corners keep their radius and border. `::before` keeps the fill behind content; a new `::after` owns the border and repaints the 8px gutter *over* content, which is what the flush, opaque, square-cornered chat composer footer was burying. Adds a source-text regression test.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no wrong assumptions | Pass | The `box-shadow` spread equals `--zero-workspace-card-gap`, the exact band the card's own `padding` reserves, so the overpaint cannot reach in-flow content. Both pseudo-elements now share `--zero-workspace-card-radius`, removing the duplicated `12px` literal. |
| `REVIEW.md` | Security (injection, secrets, auth) | N/A | Diff is one CSS block and one test that reads a repo file; no input handling, network, or auth surface. |
| `REVIEW.md` | Type safety; no `any` or broad types | Pass | Test declares `const globalCss: string` via inference from `readGlobalCss()`; no casts, no `any`. |
| `REVIEW.md` | Style and maintainability; no duplicated logic | Pass | An earlier revision carried a private `readRuleBody` duplicating the exported `readCssRule` in `global-css.ts`; `8e963714c1` deletes it and calls the shared helper, matching `mermaid-diagram-fit.test.ts`. |
| `REVIEW.md` | `fix:` commits must carry a regression test | Pass | `workspace-card-frame.test.ts` pins the frame contract. Verified it fails (1/4) when the `::after` block is deleted and again when `border` is restored on `::before`, and passes (4/4) otherwise. |
| `REVIEW.md` | New user-facing behavior gated behind a `FeatureSwitchKey` | Pass | All three rules sit under `[data-new-ui]` inside `@media (min-width: 48rem)`; `FeatureSwitchKey.NewUi` is `enabled: false` (staff-only). Runtime check on the preview: switch off → `::after` `content` is `none` and `--zero-workspace-card-radius` is unset; switch on → `content: ""`, `box-shadow: rgb(253,250,249) 0 0 0 8px`. |
| `REVIEW.md` | Feature regression guard; changed existing behavior needs coverage | Pass | The only changed existing rule is `::before` losing `border`. That is covered by the fourth test, which fails if the declaration returns. |
| `REVIEW.md` | Event sourcing and optimistic UI | N/A | No events, projections, or optimistic state. |
| `docs/bad-smell.md` | Zero tolerance for `any` | Pass | None present. |
| `docs/bad-smell.md` | Zero tolerance for lint/type suppressions | Pass | No `eslint-disable`, `@ts-ignore`, `@ts-expect-error`, or `prettier-ignore` in the diff. |
| `docs/bad-smell.md` | No unnecessary try/catch; fail fast | Pass | No error handling added; `readCssRule` throws on a missing selector and the test lets it propagate. |
| `docs/bad-smell.md` | No duplicated logic that should be shared | Pass | Resolved at `8e963714c1` as noted above. |
| `docs/fallback.md` | Default is no fallback | Pass | No `??`/`||` default, no compatibility branch, no tolerance path. |
| `docs/fallback.md` | A feature-switched feature needs no compatibility code | Pass | `NewUi` is pre-GA and staff-only, and the diff adds no cutover handling for it. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | `leaves the border to the frame alone` asserts `::before` carries no `border:`. This is the closest call in the PR, but it pins a live rendering invariant — the stroke must be drawn exactly once — whose violation is a current, user-visible regression (top corners render ~50% heavier), not the absence of deleted code. |
| `docs/fallback.md` | Every introduced fallback declared in the PR summary | N/A | The diff introduces none, so no `Fallbacks` section is required. |
| `docs/testing.md` | Integration over unit; not too many | Pass | Four assertions covering one contract, no per-function tests. |
| `docs/testing/app-testing.md` | Page-level by default; no static-config unit tests under `apps/platform` | Pass | A page test cannot assert this: happy-dom applies no CSS and performs no layout, as `global-css.ts` documents. Source-text assertion is the established pattern in this directory (`design-token-shadowing`, `entrypoint-safe-area`, `mermaid-diagram-fit`) and reuses their shared helper. |
| `docs/testing/anti-patterns.md` | AP-1 no `toHaveBeenCalled()` as sole assertion | Pass | No mocks in the file. |
| `docs/testing/anti-patterns.md` | AP-4 no `vi.mock()` on `../` internal paths | Pass | No `vi.mock()` at all. |
| `docs/testing/anti-patterns.md` | AP-5 no fake timers | Pass | None used. |
| `docs/testing/anti-patterns.md` | AP-8 no over-testing | Pass | Each test pins a distinct property: border restated, spread bounded, one shared radius, stroke drawn once. |
| `docs/effect.md` | Correct execution boundary for computed/command/effect | N/A | No React or ccstate code in the diff. |
| `docs/cache.md` | Render purity, state ownership, resource lifecycles | N/A | No components, signals, refs, or caches in the diff. |
| `.claude/skills/ccstate/SKILL.md` | ccstate signal, command, and lifecycle patterns | N/A | No ccstate usage in the diff. |

### Findings

#### Critical (P0)
None.

#### High Priority (P1)
None outstanding. One was found and fixed during review: the test file duplicated `readCssRule` as a private `readRuleBody` whose `indexOf(selector)` lookup matched on a bare prefix — `[data-new-ui] .zero-workspace-card` is a prefix of the `::before` selector, so it resolved correctly only because the card rule happens to appear first in the file. `8e963714c1` switches to the shared `readCssRule`, which anchors on `` `${selector} {` `` and cannot match the longer selector.

### Testing

- **Coverage**: Adequate. `fix:` commit carries a regression test, verified to fail against both defects it guards (missing `::after`; `border` restored on `::before`) and to pass otherwise. Full CSS suite 14/14.
- **Conventions**: Compliant. No mocks, no fake timers, no internal-path mocking, no direct component rendering. Reuses the directory's shared `global-css.ts` helper.
- **Manual verification** on the PR preview at 1440×900, chat thread page (where the flush footer reproduces the defect), both themes: bottom corners render the radius and stroke; per-corner border-pixel counts are symmetric (dark 21/21/20/20, light 25/25/24/24) against 30/30/20/24 before the second commit. Hit-tested the frame band and typed into the composer to confirm `pointer-events: none` preserves interaction.
- **Residual risk**: a non-portaled, absolutely-positioned descendant with no positive `z-index` that extends into the 8px gutter would be painted over by the ring. App popovers and tooltips are portaled to `document.body` and so unaffected, but this specific case was not exercised.

PR #31587, opened 2026-09-04
perf(python): bound resolved auth base validation work

### LINKED ISSUE #31570: perf(python): bound resolved auth.base host normalization work
## Problem

A `/firewall/auth` response can contain a resolved `base` almost as large as the runner's 256 KiB aggregate response-body cap. The runner caches that raw value, but every matching request synchronously reparses and renormalizes the complete URL on mitmproxy's event-loop thread before the forwarding path reaches an `await`.

An oversized resolved authority therefore turns one accepted auth response into repeated tens-of-milliseconds event-loop stalls on cache hits. Invalid values are eventually rejected, but only after work proportional to the full resolved base.

## Current evidence

Audited on current `main` at `0ba342178d4c5119fbfd54e14c806c94e903fa45`.

- `crates/runner/mitm-addon/src/firewall_auth_client.py:70` allows a 256 KiB auth response body, while lines 851-899 accept `base` as an optional string without a field-specific work bound.
- `turbo/packages/api-contracts/src/contracts/webhooks.ts:541-553` likewise defines `base` as an unbounded string. `turbo/apps/api/src/signals/services/agent-webhook-firewall-auth.service.ts:4895-4904,4929,5593-5607` substitutes secret/variable values and returns the resulting base without a length invariant.
- `crates/runner/mitm-addon/src/firewall_auth_cache.py:449-477` returns a cached payload directly. `crates/runner/mitm-addon/src/auth.py:557-607` checks only that the resolved base is a non-empty string.
- `crates/runner/mitm-addon/src/auth.py:1445-1462,1576-1587` calls `build_rewrite_url()` for every auth.base request, including cache hits, synchronously before awaiting forwarding.
- `crates/runner/mitm-addon/src/auth_base_rewrite.py:188-221,277-280` scans and parses the whole base, extracts the authority, percent-decodes the host, and performs hostname/IDNA normalization. There is no resolved-base or authority work limit before those operations.
- A dynamic auth base such as `${{ secrets.WEBHOOK_URL }}` can be accepted without a statically known authority (`crates/runner/mitm-addon/src/matching.py:357-372`), so runtime validation is the first point that sees the resolved size.

A minimal successful auth JSON envelope containing this value is 249,117 bytes, below the 262,144-byte response limit:

```python
resolved_base = "https://" + "%61" * 83_000  # 249,008 bytes
```

A direct eight-run probe of the production `build_rewrite_url()` on Python 3.12.3/x86_64, with inputs constructed outside the timed region, measured:

| Resolved base | Result | Time per call |
| --- | --- | ---: |
| 249,008-byte dense percent-encoded host | rejected as invalid host | 70.36-78.34 ms (73.66 ms median) |
| 249,008-byte plain ASCII host | rejected as invalid host | 33.01-34.88 ms (33.96 ms median) |
| `https://example.com/token` | accepted | 0.02 ms median |

The absolute timings are host-dependent, but they establish repeated synchronous scaling at an input size the auth client accepts. Because the raw base remains in the auth cache and is revalidated for each flow, a burst of requests can delay unrelated flow handling in the same addon process rather than paying this only once during auth resolution.

## Distinctness

- #15920 bounds aggregate auth response buffering at 256 KiB; it intentionally leaves room for long auth material and does not bound per-field event-loop work or cache-hit reuse.
- #25404 moves generic runtime URL syntax scans to C-level fast paths, but host percent-decoding and IDNA normalization still process this full authority.
- #26832 bounds nested path decoding; this input is entirely in the authority and has an empty path.
- #27713 bounds untrusted request `Host` processing; resolved auth.base targets are a separate trust boundary.
- #29547 prevents resolved bases from entering urllib's global parse cache; it does not avoid per-request parsing.

No overlapping open issue was found by title, symbol, or auth.base performance searches.

## Fix direction

Define a justified fixed maximum for the resolved auth.base string (or its authority) and enforce it at the owning runtime validation boundary before whole-string syntax scans, URL splitting, percent decoding, or IDNA normalization. A character-count guard can reject above-bound Python strings in O(1); any byte-oriented policy should avoid introducing an equivalent unbounded pre-scan.

The runner-side guard is required even if the API contract also gains a producer-side limit, because the runner owns the event-loop resource boundary. Keep existing redaction, HTTPS-only, authority, path-safety, and forwarding behavior unchanged for in-bound values.

## Acceptance criteria

- Resolved auth.base input has an explicit, documented work boundary checked before URL parsing, host percent-decoding, and hostname/IDNA normalization.
- Above-bound values fail closed through the existing auth.base rewrite failure path and never reach forwarding or DNS resolution.
- Deterministic tests cover the exact boundary and boundary-plus-one and prove expensive helpers are not entered above the limit; no wall-clock assertion is used.
- A request-handler regression covers an oversized cached resolved base so cache hits cannot repeat proportional URL-validation work.
- Existing valid auth.base URLs, including Unicode, safe percent-encoded hosts, ports, paths, and trusted query merging, retain their behavior within the boundary.
- A before/after diagnostic benchmark records bounded rejection cost for the 249 KiB case and no material regression for ordinary bases.

seven332:
## Summary

- cap complete resolved `auth.base` values at 8,192 characters before URL scanning, parsing, percent decoding, or hostname normalization
- keep oversized values on the existing redacted `url_rewrite_forward_failed` path before forwarding or DNS resolution
- cover exact-limit acceptance, structural early rejection, and an oversized cached-result request through the real flow handler

## Scope

- runner Python addon only
- no backend/API contract, cache representation, persisted-state, or public error-code changes
- valid resolved bases above 8,192 characters now fail closed; existing in-bound URL behavior is unchanged

## Diagnostic benchmark

Inputs were constructed outside the timed region and the current implementation was compared with `main` in the same process:

| Input | `main` median | Current median |
| --- | ---: | ---: |
| ordinary 25-character URL | 0.150624 ms | 0.072114 ms |
| 249,008-character dense percent-encoded host | 483.629645 ms | 0.002066 ms |

Absolute timings are host-dependent; the diagnostic verifies bounded oversized rejection without adding a wall-clock test.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_auth_base_rewrite.py tests/test_firewall_rewrite_safety.py` (158 passed)
- `uv run --no-sync python -m pytest tests/` (7,336 passed)

Closes #31570

### DIFF
--- a/crates/runner/mitm-addon/src/auth_base_rewrite.py
+++ b/crates/runner/mitm-addon/src/auth_base_rewrite.py
@@ -29,6 +29,10 @@ _URL_PATH_SAFE_CHARS = "/%:@!$&'()*+,;="
 _URL_QUERY_SAFE_CHARS = "/?%:@!$&'()*+,;="
 _VALID_AUTH_BASE_SCHEME = "https"
 
+# Bound the complete resolved value before synchronous URL processing on the
+# mitmproxy event loop. This matches the connector-check URL ceiling while
+# retaining ample room for ordinary connector URLs.
+MAX_RESOLVED_AUTH_BASE_CHARACTERS = 8 * 1024
 # This matches the runner's existing managed-query work scale while retaining
 # ample room for ordinary connector URLs. Count every segment that auth.base
 # materializes, including empty segments separated by either ``&`` or ``;``.
@@ -186,6 +190,10 @@ def _raw_rewrite_base_host(netloc: str) -> str | None:
 
 
 def _validated_rewrite_base(resolved_base: str) -> tuple[urllib.parse.SplitResult, str]:
+    if len(resolved_base) > MAX_RESOLVED_AUTH_BASE_CHARACTERS:
+        raise ValueError(
+            f"Invalid auth.base URL: must not exceed {MAX_RESOLVED_AUTH_BASE_CHARACTERS} characters"
+        )
     if "\\" in resolved_base:
         raise ValueError("Invalid auth.base URL: must not contain backslash")
     if has_raw_whitespace(resolved_base):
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #31587

**Reviewed head:** `10d51f138b13fe5cd6cac54985b7e86b8805ae77`

### Summary

The change adds a constant-time full-value guard at the owning `auth.base` validation boundary, preserves the existing redacted fail-closed response, and covers both validation ordering and the real cached-auth handler path.

### Findings

No findings.

### Testing

Ruff formatting/lint, BasedPyright, locked dependency checks, 158 focused tests, and all 7,336 mitm-addon tests passed. Unrelated Turbo and Rust suites were not run because no shared or cross-language contract changed.

### Verdict

LGTM

PR #31590, opened 2026-09-04
test(vsock-host): cover shared connection deadline

### LINKED ISSUE #31572: test(rust): cover shared vsock accept-handshake deadline
## Location

- `crates/vsock-host/src/connection/listener.rs:45-72`
- `crates/vsock-host/src/connection/listener.rs:126-180`
- `crates/vsock-host/src/tests/connection.rs:22-80`
- Production callers: `crates/sandbox-fc/src/sandbox.rs:1343-1348` and `crates/sandbox-fc/src/snapshot/runtime.rs:68-73`

## Problem

`VsockHost::wait_for_connection` deliberately computes one deadline before binding the listener, uses it for `accept`, and passes the same deadline through the READY/PING/PONG handshake. This keeps the configured connection timeout as one end-to-end budget instead of granting a fresh timeout to every phase.

The direct `wait_for_connection` tests currently cover only an unrepresentable `Duration` and cancellation cleanup. The remaining connection tests start from an already-connected stream and complete a valid handshake. No test consumes part of the timeout before or during accept, then stalls READY or PONG and proves that the handshake receives only the remaining budget.

Consequently, a refactor that restarts the timeout for accept, READY, or PONG could pass the current suite while multiplying the failure latency.

## Impact

Sandbox startup and snapshot restoration both call this entry point with `VSOCK_CONNECT_TIMEOUT`. Losing the shared-deadline invariant can delay failed VM starts/restores by multiple timeout periods and keep their startup tasks occupied longer than configured.

## Acceptance criteria

- Add a deterministic test through `VsockHost::wait_for_connection` that consumes part of the original budget, reaches an accepted connection, stalls a handshake stage, and verifies expiry at the original deadline rather than after a fresh per-stage timeout.
- Assert the returned error is `io::ErrorKind::TimedOut` and the listener socket path is removed.
- Cover ordinary no-client accept expiry and cleanup as part of the same focused test set if it can share the fixture cleanly.
- Use controlled Tokio time or explicit synchronization rather than wall-clock sleeps.
- Keep the change test-only unless a minimal testability seam is necessary.

## Observed revision

`0ba342178d4c5119fbfd54e14c806c94e903fa45`

seven332:
## Summary

- add deterministic paused-time coverage for `wait_for_connection` accept expiry and listener cleanup
- consume part of the original connection budget before accept, reach the PONG wait, and verify the handshake expires at the original absolute deadline
- share unique Unix socket path setup across the direct connection-entry tests

## Scope

This is test-only. It does not change production listener behavior, the vsock protocol, dependencies, APIs, or deployment compatibility.

## Validation

- `cargo test -p vsock-host wait_for_connection` (passed, then repeated 10 times)
- `cargo test -p vsock-host` (396 passed)
- `cargo clippy -p vsock-host --all-targets -- -D warnings`
- `cargo doc -p vsock-host --no-deps`
- pre-commit hooks: workspace Cargo fmt, docs, Clippy with all targets/all features, and file-size check

Closes #31572

### DIFF
--- a/crates/vsock-host/src/tests/connection.rs
+++ b/crates/vsock-host/src/tests/connection.rs
@@ -1,11 +1,13 @@
 use std::io;
+use std::path::PathBuf;
 use std::sync::Arc;
 use std::time::Duration;
 
+use tokio::net::UnixStream;
 use vsock_proto::{
     ExecTermination, MSG_EXEC_START, MSG_MEMORY_SNAPSHOT, MSG_MEMORY_SNAPSHOT_RESULT,
-    MSG_OPERATIONS_QUIESCED, MSG_OPERATIONS_RESUMED, MSG_QUIESCE_OPERATIONS, MSG_RESUME_OPERATIONS,
-    MSG_SHUTDOWN, MSG_SHUTDOWN_ACK, MemorySnapshot,
+    MSG_OPERATIONS_QUIESCED, MSG_OPERATIONS_RESUMED, MSG_PING, MSG_QUIESCE_OPERATIONS, MSG_READY,
+    MSG_RESUME_OPERATIONS, MSG_SHUTDOWN, MSG_SHUTDOWN_ACK, MemorySnapshot,
 };
 
 use super::support::{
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #31590

**Reviewed head:** `2f9d2aff7892dd5eb10807278939b748c3fc7765`

### Summary

The PR adds focused, test-only coverage for the shared `wait_for_connection` deadline across listener accept and the READY/PING/PONG handshake, plus ordinary accept timeout cleanup. It uses the public entry point, a real Unix stream, and paused Tokio time without changing production code.

### Findings

No findings.

### Testing

Focused tests passed and were repeated 10 times; all 396 `vsock-host` tests, rustfmt, crate Clippy with warnings denied, crate docs, and the broader pre-commit Rust hooks passed. GitHub CI remains separately pending/checked by the workflow.

### Verdict

LGTM

PR #31712, opened 2026-09-04
fix(platform): log automatic app reload reason

e7h4n:
## Summary

- log a debug event immediately before the app automatically reloads after the shared database worker becomes unavailable
- include the typed unavailable reason while avoiding misleading logs when the reload-loop guard suppresses navigation

## Verification

- `pnpm format`
- `pnpm knip`
- `pnpm --filter @okouai/app exec vitest run src/signals/__tests__/shared-database-browser.test.ts` (6 tests)

### DIFF
--- a/turbo/apps/platform/src/signals/__tests__/shared-database-browser.test.ts
+++ b/turbo/apps/platform/src/signals/__tests__/shared-database-browser.test.ts
@@ -6,6 +6,7 @@ import { expect, vi } from "vitest";
 import { setupPage } from "../../__tests__/page-helper.ts";
 import { mockNow } from "../../lib/time.ts";
 import type { SharedDatabasePortLike } from "../../shared-database/bridge.ts";
+import { logger } from "../log.ts";
 import { setupSharedDatabaseBridge$ } from "../shared-database-browser.ts";
 import { detach, Reason } from "../utils.ts";
 import { testContext } from "./test-helpers.ts";
@@ -186,6 +187,7 @@ test("Open the force-upgrade dialog when the worker requires an upgrade", async 
 
 test("Reload after an IndexedDB version change makes the worker unavailable", async () => {
   const replace = vi.fn<(url: string) => void>();
+  const debug = vi.spyOn(logger("SharedWorkerBridge"), "debug");
   const { workers } = installSharedWorkerMock();
   setupBridge();
   await vi.waitFor(() => {
@@ -207,11 +209,21 @@ test("Reload after an IndexedDB version change makes the worker unavailable", as
   await vi.waitFor(() => {
     expect(replace).toHaveBeenCalledOnce();
   });
+  expect(debug).toHaveBeenLastCalledWith("Reloading app", {
+    reason: "indexeddb-version-changed",
+  });
+  const debugCallOrder = debug.mock.invocationCallOrder.at(-1);
+  const replaceCallOrder = replace.mock.invocationCallOrder[0];
+  if (debugCallOrder === undefined || replaceCallOrder === undefined) {
+    throw new Error("Expected debug log before app reload");
+  }
+  expect(debugCallOrder).toBeLessThan(replaceCallOrder);
 });
 
 test("Reload once after the shared-data service fails to load", async () => {
   mockNow(RELOAD_AT_MS, context.signal);
   const replace = vi.fn<(url: string) => void>();
+  const debug = vi.spyOn(logger("SharedWorkerBridge"), "debug");
   const { constructorCalls, workers } = installSharedWorkerMock();
   setupBridge();
   await vi.waitFor(() => {
@@ -234,6 +246,15 @@ test("Reload once after the shared-data service fails to load", async () => {
   );
 
   expect(replace).toHaveBeenCalledOnce();
+  expect(debug).toHaveBeenLastCalledWith("Reloading app", {
+    reason: "worker-load-or-transport-failure",
+  });
+  const debugCallOrder = debug.mock.invocationCallOrder.at(-1);
+  const replaceCallOrder = replace.mock.invocationCallOrder[0];
+  if (debugCallOrder === undefined || replaceCallOrder === undefined) {
+    throw new Error("Expected debug log before app reload");
+  }
+  expect(debugCallOrder).toBeLessThan(replaceCallOrder);
   const recoveryUrl = new URL(replace.mock.calls[0]![0]);
   expect(recoveryUrl.searchParams.get("okou-shared-database-reload")).toBe(
     String(RELOAD_AT_MS),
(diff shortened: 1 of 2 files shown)


e7h4n:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 73b91cd716a842b1576c7eff3c4d4ee7ce4cfcd6

### Summary

This change adds a `SharedWorkerBridge` debug event with the typed unavailability reason immediately before an automatic app reload. The placement correctly avoids logging for force-upgrade handling and when the reload-loop guard suppresses navigation, but the new behavior is not protected by a regression assertion.

### Review Criteria

| --- | --- | --- |
| `REVIEW.md` | Changes requested | The `fix:` commit adds observable logging behavior without a regression test that would fail if the debug event were removed. |
| `docs/bad-smell.md` | Pass | The implementation uses a static logger import and the existing strict reason union, with no suppression, broad catch, dynamic import, or hard-coded endpoint. |
| `docs/testing.md` | Changes requested | Existing integration coverage exercises reload and reload-loop-guard paths, but it does not assert the debug event, payload, or absence on the guarded path. |
| `docs/fallback.md` | Pass | No compatibility fallback is introduced; the existing one-minute reload-loop guard is unchanged. |
| `docs/effect.md` | Pass | The debug side effect remains owned by the existing bridge-unavailable event path and runs immediately before its navigation effect. |
| `docs/cache.md` | Pass | No React render behavior, state/cache allocation, or resource lifetime changes are introduced. |
| `.claude/skills/ccstate/SKILL.md` | Pass | Existing command/signal ownership and cancellation behavior are unchanged; no new signal or detached work is added. |

### Findings

#### Critical (P0)

- `turbo/apps/platform/src/signals/__tests__/shared-database-browser.test.ts:187`: Add regression coverage for the behavior introduced by this `fix:` commit. The current tests only assert navigation and would still pass if `L.debug("Reloading app", { reason })` were deleted. Extend the existing browser integration cases to prove that the `SharedWorkerBridge` debug event includes the relevant reason immediately before a real reload and is not emitted when the reload-loop guard suppresses navigation.

#### High Priority (P1)

- None.

### Testing

- Coverage: Insufficient — missing regression assertions for the debug event, its reason payload, and the guarded no-log path.
- Conventions: Existing tests are compliant; add the assertions to the current browser integration suite rather than creating a unit test for the internal helper.

e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 4e811b83980158c2490096b346ae271eb6117574
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

This change emits a typed `SharedWorkerBridge` debug event immediately before the automatic `location.replace` call, while preserving the existing reload-loop guard so suppressed navigation does not produce a misleading reload log. The regression coverage exercises both reload-producing unavailable reasons, verifies log-before-navigation ordering, and verifies that the guarded no-reload path does not emit the event.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and edge cases | Pass | `shared-database-browser.ts:64-99` forwards the schema-derived reason and logs only after the reload-loop early return, directly before `location.replace`. |
| `REVIEW.md` | Security and configuration safety | Pass | The diff adds no input parsing, auth boundary, secret, URL source, or environment-specific configuration. |
| `REVIEW.md` | Type safety and interface integrity | Pass | `handleSharedDatabaseReloadRequired` accepts `SharedDatabaseWorkerUnavailableReason`; no `any`, casts, or public contract changes are introduced. |
| `REVIEW.md` | Deployment, DB, and JSONB compatibility | N/A | The change only adds in-process diagnostic emission and argument forwarding; no schema, persisted shape, request, response, or worker protocol shape changes. |
| `REVIEW.md` | Required regression coverage for a fix | Pass | `shared-database-browser.test.ts:188-306` covers both automatic-reload reasons, ordering, and the reload-loop suppression path. |
| `docs/bad-smell.md` | Avoid `any`, suppressions, and dynamic imports | Pass | Both changed files use a static logger import and the existing typed unavailable-reason union, with no suppression comments or dynamic import. |
| `docs/bad-smell.md` | Fail fast without hiding failures | Pass | Existing post-reload errors and the visible repeated-failure toast remain unchanged; the new log does not catch or replace either failure. |
| `docs/testing.md` | Prefer integration-level regression tests | Pass | The existing bridge integration test delivers real worker events through the message port and observes the resulting logger/navigation behavior. |
| `docs/testing.md` | Mock only boundaries and keep real timing | Pass | The test uses browser/SharedWorker boundary doubles, `testContext`, real timers, and the production `mockNow` clock abstraction; it adds no internal module mock or fake timer. |
| `docs/fallback.md` | Do not introduce undeclared or open-ended fallbacks | N/A | No fallback or compatibility branch is introduced; the pre-existing one-minute reload-loop guard is behaviorally unchanged. |
| `docs/effect.md` | Match side effects to their semantic owner | Pass | Diagnostic emission stays in the existing worker-unavailable callback and is colocated with the navigation it describes; no React render/effect or new async owner is added. |
| `docs/effect.md` | Preserve cancellation and resource ownership | N/A | The synchronous logger/navigation change acquires no resource and does not alter the bridge's existing `AbortSignal` lifecycle. |
| `docs/cache.md` | Avoid render effects, duplicate state, caches, and lifecycle leaks | N/A | The diff creates no React render work, mutable state, cache, listener, timer, subscription, or retained resource. |
| `.claude/skills/ccstate/SKILL.md` | Keep signal composition and accessors inside commands | Pass | The existing ccstate command/event composition is unchanged; the new typed argument and synchronous log do not escape `get`/`set` or add `detach`. |
| `.claude/skills/ccstate/SKILL.md` | Give async work an owned abort path | N/A | No async operation, polling loop, reset signal, or new abortable work is introduced. |
| `docs/testing/patterns.md` | Use `testContext` and production-like platform boundaries | Pass | The test retains one file-scope `testContext`, the real bridge implementation, and the SharedWorker/browser boundary harness. |
| `docs/testing/anti-patterns.md` | Assert outcomes rather than only mock orchestration | Pass | Assertions cover the exact reason payload, relative call order, actual `location.replace`, and the suppressed-navigation outcome; no internal mock, fetch mock, fake timer, or direct component render is added. |
| `docs/testing/app-testing.md` | Use signal tests only for non-page instrumentation and context-owned cleanup | Pass | This no-page diagnostic/reload path is exercised through the existing signal bootstrap harness; logger/browser spies and the application clock remain owned by Vitest and `testContext`. |
| `docs/testing/testing-external-behavior.md` | Drive and observe the closest production boundary | Pass | Worker-unavailable events enter through the message-port boundary, and the test observes the diagnostic call plus browser navigation rather than mutating internal state to manufacture the result. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- Coverage: Adequate — both automatic-reload reasons, strict log-before-navigation ordering, and the loop-guard no-log case are covered.
- Conventions: Compliant — no internal module mocks, direct fetch mocks, fake timers, or cleanup workarounds were added.
- Executed: `pnpm --filter @okouai/app exec vitest run src/signals/__tests__/shared-database-browser.test.ts` — 1 file passed, 6 tests passed.

PR #31718, opened 2026-09-04
fix(chat): skip empty event tails in batch catch-up

e7h4n:
## Summary

- skip empty per-thread tails before batched ChatEvent IndexedDB writes
- report only thread IDs that actually persisted new events, preventing no-op invalidation broadcasts
- keep expired-cursor snapshot rebuild behavior unchanged

## Verification

- `pnpm format`
- `pnpm knip`
- `pnpm --filter @okouai/app exec vitest run src/shared-database/__tests__/worker-runtime.test.ts` (8 tests)

## Fallbacks

- None.

### DIFF
--- a/turbo/apps/platform/src/shared-database/__tests__/worker-runtime.test.ts
+++ b/turbo/apps/platform/src/shared-database/__tests__/worker-runtime.test.ts
@@ -505,6 +505,47 @@ test("Rebuild chat data after its saved cursor expires", async () => {
   }
 });
 
+test("Report only non-empty tails from a batched chat catch-up", async () => {
+  const { runtime } = startRuntime();
+  const unchangedThreadId = crypto.randomUUID();
+  const updatedThreadId = crypto.randomUUID();
+  const updatedRow = chatEventRow(updatedThreadId, 1);
+  context.mocks.api(chatThreadEventsContract.catchUp, ({ body, respond }) => {
+    expect(body).toStrictEqual([
+      [unchangedThreadId, 0],
+      [updatedThreadId, 0],
+    ]);
+    return respond(200, {
+      events: {
+        [unchangedThreadId]: [],
+        [updatedThreadId]: [updatedRow],
+      },
+      notFoundThreads: [],
+    });
+  });
+
+  await expect(
+    runtime.catchUpChatEvents(
+      [unchangedThreadId, updatedThreadId],
+      context.signal,
+    ),
+  ).resolves.toStrictEqual([updatedThreadId]);
+  await expect(
+    queryRuntime(runtime, {
+      dataKey: chatEventKey(unchangedThreadId),
+      afterSeqId: null,
+      consistency: "cache-only",
+    }),
+  ).resolves.toStrictEqual([]);
+  await expect(
+    queryRuntime(runtime, {
+      dataKey: chatEventKey(updatedThreadId),
+      afterSeqId: null,
+      consistency: "cache-only",
+    }),
+  ).resolves.toStrictEqual([updatedRow]);
+});
+
 test("Rebuild a cached chat when batched catch-up cannot continue its cursor", async () => {
   const { runtime } = startRuntime();
   const dataKey = chatEventKey(crypto.randomUUID());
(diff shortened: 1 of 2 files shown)


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 7cd932a033460e135a097f6d77a4d900409b675c
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

This fix removes empty per-thread ChatEvent tails before the SharedWorker's batched IndexedDB write, so only threads with newly persisted events are returned for invalidation. Non-empty tail persistence and expired-cursor snapshot rebuilds retain their existing behavior.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness, edge cases, and API assumptions | Pass | `catchUpChatEvents` filters only the contract-defined empty-tail case; non-empty rows keep the last server event as their cursor and `notFoundThreads` still follows the snapshot rebuild path. |
| `REVIEW.md` | Security and type safety | Pass | The diff adds no input surface, authorization change, unsafe cast, or broadened type. `writes` remains explicitly typed as `ChatEventBatchWrite[]`. |
| `REVIEW.md` | A `fix:` change has regression coverage | Pass | The mixed empty/non-empty batch test fails under the previous implementation because it returned the unchanged thread ID, and passes only when the updated thread is reported. |
| `docs/bad-smell.md` | Avoid `any`, suppressions, dynamic imports, and defensive fallback paths | Pass | The production diff is a typed local transformation with no suppression, dynamic import, catch/recovery branch, or fabricated default. |
| `docs/fallback.md` | Add no fallback for unreachable or retired states | Pass | An empty tail is an explicit current response-contract state, and the diff removes its no-op cache write instead of adding compatibility behavior. No fallback is introduced. |
| `docs/event-sourcing.md` | Preserve persistent events as authoritative and server ordered | Pass | Only server-returned rows are persisted, using the final row's server-owned event and sequence IDs; snapshot replacement and reconciliation behavior are unchanged. |
| `docs/effect.md` | Match asynchronous work to its lifecycle owner | N/A | The diff changes no React effect, ccstate command, signal ownership, listener, timer, or resource lifecycle. |
| `docs/cache.md` | Keep cache ownership and invalidation precise | Pass | Empty tails no longer create cursor writes or downstream invalidations; the existing user/org-scoped IndexedDB owner and teardown path are unchanged. |
| `.claude/skills/ccstate/SKILL.md` | Preserve ccstate command, signal, and abort ownership rules | N/A | No ccstate state, computed, command, view callback, or AbortSignal flow changes in this diff. |
| `docs/deployment-compatibility.md` | Preserve cross-version API and persisted-state compatibility | Pass | The API request/response schema, ChatEvent row/cursor shape, IndexedDB schema version, and worker-tab protocol are unchanged. |
| `docs/testing.md` | Prefer integration coverage and mock only system boundaries | Pass | The test uses the real worker runtime and IndexedDB path while mocking the typed HTTP contract through `context.mocks.api`. |
| `docs/testing/patterns.md` | Exercise real behavior through established fixtures | Pass | The established `testContext`, runtime fixture, typed contract mock, and cache query helpers drive the complete batch catch-up path. |
| `docs/testing/anti-patterns.md` | Avoid internal mocks, fake timers, and mock-call-only assertions | Pass | The added test has no module mock or fake timer and asserts returned thread IDs plus persisted cache outcomes. |
| `docs/testing/app-testing.md` | Use a direct protocol test only when there is no page-visible control surface | Pass | Batched SharedWorker cache persistence is a protocol/runtime contract; the test exercises its production runtime boundary without directly mutating internal state. |
| `docs/testing/testing-external-behavior.md` | Provide boundary context and observe externally exposed outcomes | Pass | The case is constructed through the catch-up HTTP response and observed through the runtime result and cache-only query, not private store mutation or spy calls. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- Coverage: Adequate. The regression covers one unchanged and one updated thread in the same batch, verifies their request cursors, checks the exact returned invalidation set, and confirms only the updated row is available from cache.
- Conventions: Compliant. `corepack pnpm --filter @okouai/app exec vitest run src/shared-database/__tests__/worker-runtime.test.ts` passed all 8 tests on reviewed commit `7cd932a033460e135a097f6d77a4d900409b675c`.

PR #31769, opened 2026-09-04
perf(guest-agent): bound pi rpc response buffering

### LINKED ISSUE #31634: perf(rust): bound Pi RPC response buffering
## Problem

Verified on `7f33f1ef52fa60c20c0782aef928c0f514d54106`.

The ordinary Pi CLI path creates an unbounded response channel and gives its sender to the stdout projector:

- [the channel is unbounded and its receiver is owned by the serialized command writer](https://github.com/vm0-ai/vm0/blob/7f33f1ef52fa60c20c0782aef928c0f514d54106/crates/guest-agent/src/cli/mod.rs#L1090-L1114)
- [every syntactically valid `type=response` record is enqueued without a count or byte budget](https://github.com/vm0-ai/vm0/blob/7f33f1ef52fa60c20c0782aef928c0f514d54106/crates/guest-agent/src/cli/pi_rpc.rs#L432-L500)
- [the writer consumes responses serially while validating one expected command](https://github.com/vm0-ai/vm0/blob/7f33f1ef52fa60c20c0782aef928c0f514d54106/crates/guest-agent/src/cli/pi_rpc.rs#L792-L827)
- [normal command flow has only one outstanding command at a time](https://github.com/vm0-ai/vm0/blob/7f33f1ef52fa60c20c0782aef928c0f514d54106/crates/guest-agent/src/cli/pi_rpc.rs#L932-L980)

Each stdout record may contain up to 16 MiB before parsing ([shared framing limit](https://github.com/vm0-ai/vm0/blob/7f33f1ef52fa60c20c0782aef928c0f514d54106/crates/guest-contracts/src/stdout_framing.rs#L6-L11)). A faulty or compromised Pi host can therefore emit a burst of large valid response records while the writer is descheduled or blocked writing the next command. The stdout loop drains those records from the bounded OS pipe into owned `serde_json::Value` instances, so pipe capacity does not bound retained heap. For scale, 64 near-limit responses can retain roughly 1 GiB before allocator and JSON-tree overhead.

The valid protocol needs only the response for the single outstanding command, plus a small bounded race allowance when cancellation sends `abort` and ignores an earlier unmatched acknowledgement. Unbounded retention is not required for normal behavior.

## Suggested direction

Replace the unbounded response channel with a small explicit budget. Admission from the stdout loop must remain non-blocking: awaiting channel capacity there can stop stdout drainage and recreate the child-pipe deadlock that event delivery intentionally avoids. Treat overflow as an explicit Pi protocol/control failure and route it through the existing bounded process-group termination path. Preserve the cancellation case that may observe a preceding acknowledgement before the matching abort response.

## Acceptance criteria

- [ ] Pi RPC response buffering has an explicit finite count and retained-byte bound, including queued and in-flight response values.
- [ ] The stdout loop never waits for response-channel capacity.
- [ ] Overflow fails the run with an actionable diagnostic and reaches bounded child cleanup; responses are not silently dropped.
- [ ] Existing `get_state` projection, response ID/command/success validation, prompt/steer acknowledgements, and cancellation-time abort matching remain unchanged.
- [ ] Deterministic coverage drives a stalled response consumer with a response burst and proves the configured bound, failure classification, and child reaping.
- [ ] Relevant guest-agent formatting, Clippy, documentation, and tests pass.

seven332:
## Summary

- replace the unbounded Pi RPC response channel with a two-entry bounded queue
- cap queued and consumer-held response input bytes at 16 MiB using response-owned permits
- fail queue or byte-budget overload through the existing bounded stdout-ingestion termination path
- add real-entry-point integration coverage for count overflow, byte overflow, and child reaping

## Scope

The Pi wire protocol, response validation, API contracts, persistence, and deployment interfaces are unchanged. A single response accepted by the existing stdout framing limit remains admissible; only abnormal aggregate response bursts now fail explicitly.

## Validation

- `cargo doc -p guest-agent --no-deps --all-features`
- `cargo test -p guest-agent --test pi_rpc_response_overload --all-features -- --nocapture`
- `cargo test -p guest-agent --all-features`
- repository pre-commit hooks

Closes #31634

### DIFF
--- a/crates/guest-agent/src/cli/mod.rs
+++ b/crates/guest-agent/src/cli/mod.rs
@@ -1090,7 +1090,7 @@ async fn execute_cli_inner(
 
     let active_input_controller = active_input.controller();
     let pi_execution = matches!(runtime.framework, env::Framework::Pi);
-    let (pi_rpc_response_tx, pi_rpc_response_rx) = tokio::sync::mpsc::unbounded_channel();
+    let (pi_rpc_response_tx, pi_rpc_response_rx) = pi_rpc::response_channel();
     let (pi_rpc_startup_tx, pi_rpc_startup_rx) = tokio::sync::oneshot::channel();
     let mut pi_rpc_startup_tx = pi_execution.then_some(pi_rpc_startup_tx);
     let pi_rpc_cancellation = CancellationToken::new();
@@ -1434,7 +1434,7 @@ async fn execute_cli_inner(
                                 }
                             }
                             if let Some(projection) = pi_rpc_projection.as_mut() {
-                                match projection.project(event, &pi_rpc_response_tx) {
+                                match projection.project(event, &pi_rpc_response_tx, line.len()) {
                                     Ok(Some(projected)) => event = projected,
                                     Ok(None) => {
                                         agent_log.write_raw_line(line.as_bytes()).await;
--- a/crates/guest-agent/src/cli/pi_rpc.rs
+++ b/crates/guest-agent/src/cli/pi_rpc.rs
@@ -14,7 +14,8 @@
 //! - The guest writer owns child stdin. It sends `get_state`, waits until the
 //!   private startup record is installed, sends the initial `prompt`, and then
 //!   delivers accepted active-input frames. The stdout loop routes `response`
-//!   records into the writer's response channel.
+//!   records into the writer's bounded response channel without waiting for
+//!   capacity.
 //! - The stdout loop owns child stdout. It retains each ordinary raw record in
 //!   the best-effort local agent transcript, projects supported records into
 //!   the existing public event shape, and passes projected events through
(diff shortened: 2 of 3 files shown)


seven332:
## Code Review: PR #31769

**Reviewed head:** `6b2c88e358298c3aa5fa919f561ee771b75cf161`

### Summary

The PR replaces the unbounded Pi RPC response path with a non-blocking two-entry queue and a 16 MiB retained-input-byte budget whose permits follow responses through validation. It preserves the existing Pi wire protocol and response validation behavior.

### Findings

No findings.

### Testing

The real-entry-point integration test covers count and byte overload, actionable `StdoutIngestion` failure, SIGTERM cleanup, and child reaping. Formatting, Clippy, documentation, the focused integration test, the full `guest-agent` suite, and repository pre-commit hooks passed. A live Pi provider was not exercised; the test uses the documented JSONL protocol through the production guest path.

### Verdict

LGTM

Issue #31534, opened 2026-09-03
bug(runner): reused sandbox file write stalls after unpark


hulh122:
## Problem

A reused Firecracker sandbox can acknowledge `ReadyForOperations` while physical
balloon deflation is still in progress, then lose the terminal response for its
first private-file write. The initiating write waits 60 seconds in
`await_terminal_response` and poisons the shared connection as designed. Before
the Runner correction, startup continued and a later required write could replace
the initiating error with a secondary "normal operations unavailable" error.

Two Runner `0.184.13` runs reproduced the failure on
`prod-11.gcp.vm3.ai` on 2026-09-03:

| Runner | Connector-context writes | Failures | Maximum duration |
| --- | ---: | ---: | ---: |
| `prod-11.gcp.vm3.ai` | 44 | 2 | 60,002 ms |
| `prod-12.gcp.vm3.ai` | 30 | 0 | 10 ms |
| `prod-13.gcp.vm3.ai` | 82 | 0 | 19 ms |

Affected runs:

- `8007a78e-4df6-4a1e-96cd-2b3c39127d33`
- `1e1af815-3473-4936-a1fb-d2a4d40be276`

The first run came from `Goal`, used model `gpt-5.6-sol`, exposed the failure
after approximately 60 seconds, and completed zero Agent steps.

The first run used claim-aware exact handoff while the predecessor's park balloon
was still settling. Normal operations reopened about one millisecond after target
zero was requested; target and actual reached zero about 586 ms later. The active
controller then re-inflated the balloon to a 2,865 MiB target while the write was
still pending.

## Diagnosis and constraints

- Host evidence ruled out material CPU, memory, I/O, swap, PSI, OOM, and blocked-
  task pressure during the failure window.
- A 4 GiB guest could previously be reclaimed to a 512 MiB guest-visible floor by
  either active reclaim or idle park, below the Runner's existing 1 GiB minimum
  supported profile capacity.
- Each private write launches an unprivileged `/sbin/guest-write-file` helper.
  Existing logs do not isolate spawn/exec, stdin transfer, filesystem work, helper
  exit, and response emission, so allocation headroom remains a supported
  mitigation hypothesis rather than a stage-proven unique root cause.
- A full balloon target/actual-zero readiness barrier added about 0.79 seconds to
  exact reuse and remains excluded.
- The unprivileged helper boundary, asynchronous unpark, `deflate_on_oom`, active
[...]

PR #31770, opened 2026-09-04
fix(runner): require connector account context at startup

### LINKED ISSUE #31546: bug(runner): fail runs when connector account context cannot be written
## Parent objective

This is the second delivery slice of #31534. Make connector account context a required part of run startup and preserve its initiating write failure.

## Problem

Runner currently treats the connector-account-context projection as best effort. If writing the private file fails, Runner records telemetry, omits `OKOU_CONNECTOR_ACCOUNT_CONTEXT_FILE`, and continues to the required run-payload write and Agent startup.

Although authorization and credential injection do not consume this file, a new run without it cannot reliably inspect the exact connector accounts pinned at admission. The CLI deliberately fails closed instead of substituting a live default, so allowing startup creates a partially initialized run that does not satisfy the connector-account introspection contract.

The product requirement is that every new run has this context. A write failure must therefore fail the run regardless of whether the guest connection would otherwise remain reusable.

## Required behavior

- On a successful write, publish the private file path and continue startup unchanged.
- On any write failure, record the existing bounded telemetry and warning, then return the initiating error immediately.
- Do not write the later required private-file batch or start the Agent after failure.
- Do not retry the write or classify selected failures as safe degradation.

## Scope

- Change only the Runner startup decision at the connector-account-context write boundary.
- Preserve the existing file schema, environment key, telemetry action, timeout-stage outcome, error type, and error text.
- Cover fresh and reused sandboxes, including guest rejection, every typed timeout stage, and backend failure.

## Acceptance criteria

- Every connector-account-context write failure fails the run with the original error.
- No later required private-file write or Agent process start occurs after the failure.
- Timeout-stage telemetry remains bounded and unchanged.
- The sandbox is not returned for reuse after the failed execution.
- Successful startup behavior remains unchanged.
- Applicable Runner formatting, static checks, documentation, and tests pass.

## Compatibility

There is no API, queue, database, persisted-state, guest wire, or file-format change. Old Runner processes retain the previous best-effort behavior while draining; new Runner processes enforce the required startup invariant. Rollback requires no migration or cleanup.

## Out of scope

- Retrying the write, reconnecting, or falling back to a fresh sandbox.
- Changing connector authorization, firewall registration, credential injection, or account selection.
- Introducing a new host or sandbox error classification.
- Balloon memory policy and Firecracker boot configuration, completed separately by #31545 and #31594.

## Risks

- A transient context-file write failure that previously allowed a degraded run will now fail the run.
- The stronger invariant can increase visible run failures, but prevents Agents from operating without the exact admitted-account context.
- Future optional projections must not be accidentally grouped with this required startup file.

## Relevant areas

- `crates/runner/src/executor/agent_run.rs`
- `crates/runner/src/executor/tests/sandbox_run_tests/fresh_sandbox.rs`
- `crates/runner/src/executor/tests/telemetry_tests.rs`

seven332:
## Summary

- treat connector account context as a required part of every new run
- return the initiating write error after recording the existing telemetry and warning
- prevent later required private-file writes and Agent startup after any connector-context write failure
- cover fresh and reused sandboxes across guest, timeout-stage, and backend failures

## Compatibility

- no guest wire protocol, API, queue, database, persisted-state, environment-key, or file-format changes
- successful run startup is unchanged; only previously degraded starts now fail immediately

## Validation

- `RUSTDOCFLAGS='-D warnings' cargo doc --no-deps -p runner --all-features`
- focused fresh-sandbox and connector-account-context Runner tests (6 passed)
- `cargo test -p runner --all-targets --all-features` (3445 passed, 25 ignored; guest inventory passed)
- pre-commit Rust format, documentation, Clippy, and file-size checks
- `git diff --check`

Closes #31546

Parent context: #31534

### DIFF
--- a/crates/runner/src/executor/agent_run.rs
+++ b/crates/runner/src/executor/agent_run.rs
@@ -111,6 +111,21 @@ fn private_write_timeout_stage(error: &RunnerError) -> Option<&'static str> {
     })
 }
 
+fn connector_account_context_write_allows_continue(error: &RunnerError) -> bool {
+    matches!(
+        error,
+        RunnerError::Sandbox(sandbox::SandboxError::Operation {
+            operation: sandbox::SandboxOperation::WriteFile,
+            reason: sandbox::SandboxOperationReason::GuestRejected,
+            ..
+        }) | RunnerError::Sandbox(sandbox::SandboxError::OperationTimeout {
+            operation: sandbox::SandboxOperation::WriteFile,
+            stage: sandbox::SandboxOperationTimeoutStage::BeforeFrameWrite,
+            ..
+        })
+    )
+}
+
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 enum SessionHistoryIdentityReason {
     FinalizeMetadataPathUnresolved,
@@ -1998,6 +2013,9 @@ pub(super) async fn run_in_sandbox_with_process_cancel_timeouts(
                 outcome = outcome.unwrap_or("write_failed"),
                 "connector account context unavailable"
             );
+            if !connector_account_context_write_allows_continue(&error) {
+                return Err(error);
+            }
         }
     }
     let env_build_started = Instant::now();
--- a/crates/runner/src/executor/agent_run.rs
+++ b/crates/runner/src/executor/agent_run.rs
@@ -111,21 +111,6 @@ fn private_write_timeout_stage(error: &RunnerError) -> Option<&'static str> {
     })
 }
 
-fn connector_account_context_write_allows_continue(error: &RunnerError) -> bool {
-    matches!(
-        error,
-        RunnerError::Sandbox(sandbox::SandboxError::Operation {
-            operation: sandbox::SandboxOperation::WriteFile,
-            reason: sandbox::SandboxOperationReason::GuestRejected,
-            ..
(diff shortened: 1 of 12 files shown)


seven332:
## Code Review: PR #31770

**Reviewed head:** `6eb93ad5446c1a508a9594c898b4948a41e15588`

### Summary

The change preserves terminal guest write rejections as a narrow structured fact, keeps the existing displayed error label, and makes Runner fail closed for every connector-context write whose reuse safety is unproven. The representation stays within the Runner artifact and does not change the guest wire or independently deployed contracts.

### Findings

No findings.

No new fallback is introduced. The PR narrows the existing optional connector-context continuation path from every write failure to only a decoded terminal guest rejection or a timeout before frame writing begins.

### Testing

Coverage includes real vsock terminal responses, malformed and disconnected transports, Firecracker adapter classification, all typed timeout stages, backend crash, initiating-error preservation, no retry or later required write, Agent-start suppression, and sandbox reuse disposition. Formatting, all-target check, Clippy with warnings denied, rustdoc with warnings denied, and the complete affected-crate test suite passed. Real Firecracker guest E2E was not run locally; the adapter boundary is exercised with the existing connected Unix-stream guest fixture.

### Verdict

LGTM

PR #31774, opened 2026-09-04
fix(python): retry auth.base socket creation failures

### LINKED ISSUE #31642: fix(python): retry auth.base after socket creation failure
## Problem and evidence

`_ValidatedTLSConnection._connect_raw_socket()` is intended to try each validated DNS address until one connects, but `socket.socket()` currently runs before the per-address `try` block. An address-specific creation failure therefore aborts the whole auth.base forward instead of trying the next validated address.

On current `main` at `04484f4ae0`:

- `crates/runner/mitm-addon/src/auth_base_transport.py` catches `OSError` from socket registration, timeout setup, and `connect()`, while socket creation remains outside that handler.
- `crates/runner/mitm-addon/tests/test_auth_base_forwarder_security.py` covers a failed first `connect()` followed by a successful second address, but both socket constructors succeed. The test cannot detect a creation-time failure.
- A deterministic IPv6-first probe in the original report made `socket.socket(AF_INET6, ...)` raise `EAFNOSUPPORT`; the IPv4 candidate was never attempted.

No overlapping open issue was found for auth.base socket creation or address fallback.

## Impact

If DNS returns an IPv6 address before an IPv4 address on a host where IPv6 socket creation is unsupported, auth.base forwarding fails even though the already-validated IPv4 destination is usable. Users can see a local connector forwarding failure for a reachable upstream.

## Decision

Treat an `OSError` from creating one candidate socket as that address attempt's failure, consistent with the existing retry contract for timeout setup and `connect()`. Continue through the already-validated address list under the existing deadline and cancellation checks. If every creation attempt fails, propagate the final `OSError`.

The change is limited to raw-socket candidate iteration and its integration coverage. DNS validation, address ordering, TLS/SNI behavior, request forwarding, and exception handling outside this address-attempt boundary remain unchanged.

## Alternatives considered

- Keep the current behavior and only document that some validated addresses cannot be attempted: rejected because it contradicts the existing multi-address fallback contract and turns a usable validated destination into a request failure.
- Retry only `EAFNOSUPPORT`: rejected because other `OSError` values can also be scoped to a candidate, while the bounded address loop already retries `OSError` from later socket setup stages and ultimately preserves the final error.
- Add a broad outer retry around the full connection: rejected because it would repeat TLS or HTTP work and weaken the existing validated-address pinning boundary.

## Risks and mitigations

- Retrying a host-wide resource error may perform several doomed socket creations. The work remains bounded by the validated address list, and the final relevant error is preserved.
- Cleanup code could claim a socket that was never created. Tests must verify that creation failures do not register, clear, or close a nonexistent socket, while successfully created candidates keep their current lifecycle behavior.

## Acceptance criteria

- An `OSError` raised by `socket.socket()` for one validated address advances to the next validated address within the existing deadline and cancellation contract.
- A deterministic IPv6-first regression test makes `socket.socket(AF_INET6, ...)` raise `EAFNOSUPPORT`, then verifies that the IPv4 candidate completes the existing TLS/SNI request path without another DNS lookup.
- When every socket creation fails, the final `OSError` is propagated.
- Failed creation attempts do not register, clear, close, or otherwise claim a socket that was never created; later created sockets retain the existing cleanup behavior.
- Existing connect-failure fallback, deadline, cancellation, validated-address pinning, and Python package checks pass.

seven332:
## Summary

- retry the next already-validated auth.base address when one socket constructor raises `OSError`
- preserve the existing deadline, cancellation, DNS pinning, and concrete-socket cleanup behavior
- cover IPv6 creation failure followed by IPv4 request success and exhausted constructor failures through the forwarding entry point

## Exclusions

- no DNS resolution or address-ordering changes
- no TLS, HTTP forwarding, API, persistence, or cross-version protocol changes
- no errno allowlist or shared test-helper API expansion

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (7,271 passed)
- `lefthook run pre-commit --files 'crates/runner/mitm-addon/src/auth_base_transport.py crates/runner/mitm-addon/tests/test_auth_base_forwarder_security.py'`
- commit-time pre-commit and commit-message hooks

Closes #31642

### DIFF
--- a/crates/runner/mitm-addon/src/auth_base_transport.py
+++ b/crates/runner/mitm-addon/src/auth_base_transport.py
@@ -199,7 +199,11 @@ class _ValidatedTLSConnection(http_client.HTTPConnection):
         last_error: OSError | None = None
         for address in self._validated_addresses:
             self._abort_handle.raise_if_aborted()
-            raw_sock = socket.socket(address.family, socket.SOCK_STREAM)
+            try:
+                raw_sock = socket.socket(address.family, socket.SOCK_STREAM)
+            except OSError as exc:
+                last_error = exc
+                continue
             try:
                 self._abort_handle.register_socket(raw_sock)
                 raw_sock.settimeout(remaining_deadline_seconds(self._deadline))
--- a/crates/runner/mitm-addon/tests/test_auth_base_forwarder_security.py
+++ b/crates/runner/mitm-addon/tests/test_auth_base_forwarder_security.py
@@ -9,7 +9,7 @@ import pytest
 
 import auth_base_forwarder as forwarder
 import auth_base_transport as transport
-from tests.auth_base_forwarder_helpers import FakeSocket, fake_forwarder_upstream
+from tests.auth_base_forwarder_helpers import FakeSocket, fake_forwarder_upstream, http_response
 
 
 class TestAuthBaseForwarderSecurity:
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #31774

**Reviewed head:** `405550d51ffc5b920308551b06db4afc8027a8eb`

### Summary

The change adds the missing candidate-level `OSError` boundary around socket construction while leaving validated-address ordering, deadline checks, cancellation, and concrete-socket ownership unchanged. Both new regressions run through the auth.base forwarding entry point and use only existing external-boundary fakes.

### Findings

No findings.

### Testing

Verified uv lock/sync, Ruff format and lint, BasedPyright, all 7,271 mitm-addon tests, the 124-test auth.base security/protocol/lifecycle subset, and root commit hooks. Production traffic replay was not run; the repository does not provide frequency data for this host capability failure.

### Verdict

LGTM

PR #31775, opened 2026-09-04
perf(python): bound request header field count

### LINKED ISSUE #31645: perf(python): bound request-header field count before repeated hook scans
## Problem

At audited `main` commit `04484f4ae031ad55125a0972229e12338d3d3f0d`, the mitmproxy addon bounds each raw request-header name to 4,096 bytes but does not bound the number of fields before repeatedly traversing the complete tuple.

For a normally framed request with `Content-Length: 0`, the common `requestheaders()` prefix makes repeated complete passes before its first classification decision:

- `mitm_addon.requestheaders()` walks every field for the per-name limit.
- Codex catalog prefetch capture walks every field looking for its private marker.
- Connector-intent capture walks every field, then calls `Headers.set_all()` even when the header is absent, rebuilding the complete tuple.
- Auth-base, AWS SigV4, and stream-buffer framing checks repeatedly perform fixed-name `Headers.get_all()` lookups.

Pinned mitmproxy 12.2.3 implements these header operations as linear scans that normalize every field name. Its HTTP/1 request reader buffers until the complete head, has an explicit header-size DoS FIXME, and applies no field-count or aggregate-head limit before this hook. HTTP/2's decompressed header-list policy therefore does not protect the HTTP/1 path.

## Evidence and impact

The current `requestheaders()` boundary was benchmarked with genuine mitmproxy `Request` and `Headers` objects under CPython 3.12.3 / x86_64 / mitmproxy 12.2.3. Each flow had `Host`, `Content-Length: 0`, and repeated valid one-byte `X: y` fields. Input construction was outside measurement; medians are from five fresh-flow runs:

| Repeated fields | Approximate HTTP/1 request-head bytes | Median | Range |
| ---: | ---: | ---: | ---: |
| 10,000 | 60,056 | 5.050 ms | 4.995-5.097 ms |
| 50,000 | 300,056 | 25.017 ms | 24.874-26.224 ms |
| 100,000 | 600,056 | 50.083 ms | 49.944-50.631 ms |
| 250,000 | 1,500,056 | 126.938 ms | 125.464-133.006 ms |
| 500,000 | 3,000,056 | 252.090 ms | 250.053-256.751 ms |

A separate `tracemalloc` run started after constructing the 500,002-field request and observed 7.789 MiB of additional peak allocation inside the hook. A sandbox can repeat dense HTTP/1 request heads over its local proxy connection, adding synchronous Python work and allocation pressure on the shared mitmproxy event loop.

## Distinctness

- #28330 added the 4,096-byte per-name limit and explicitly left total field count as residual exposure.
- #26101 bounds count and aggregate bytes only after a request is classified for managed AWS SigV4 inspection.
- #27630 and #27717 bound private-header values, but raw selection still traverses the complete tuple.
- No open duplicate issue or linked PR covers this generic field-count boundary.

## Decision

Add one addon-owned, inclusive limit of 2,048 raw request-header fields at the beginning of `requestheaders()`. This matches the existing managed AWS SigV4 count policy, derived from the pinned HTTP/2 64 KiB decompressed header-list budget and its minimum 32-byte per-field accounting.

Check `len(flow.request.headers.fields)` before the per-name limit or any individual field access. At limit-plus-one, replace the raw field tuple with an empty tuple, set the existing request-header terminal marker, and kill the flow. Do not inspect, filter, log, or serialize rejected fields.

Immediate connection/stream termination is intentional instead of HTTP 431. Pinned mitmproxy does not send a response assigned during `requestheaders()` until request-body completion. Locating and preserving `Expect: 100-continue` would rescan the attacker-sized tuple; clearing it while waiting for the body can deadlock the client; retaining it also leaves mitmproxy's post-hook lookup unbounded. Mitmproxy checks killed state immediately after the hook, before its `Expect` lookup or body-consumption transition.

## Alternatives considered

- **Hybrid 431/kill:** returning 431 only for already-complete requests is safe but adds body/protocol lifecycle branches for malformed input without a current product requirement.
- **Always return 431:** cannot satisfy the bounded-work and open-body requirements without retaining or rescanning the over-limit tuple.
- **Distributed guards:** can drift across consumers and do not protect mitmproxy's post-hook `Expect` lookup.
- **Parser-level limit:** could also bound pre-hook buffering and allocation, but requires a broader private mitmproxy compatibility design across HTTP/1, HTTP/2, and HTTP/3.

## Risks and mitigations

- **Unusual HTTP/1 clients above 2,048 fields:** use the existing high compatibility threshold and test exact-limit acceptance.
- **Loss of rejected header detail:** release the complete tuple deliberately; do not serialize attacker-controlled input or add an unrequired diagnostic contract.
- **Body or `Expect` deadlock:** kill before mitmproxy's post-hook lookup and verify the real HTTP/1 state machine.
- **Later rescan or retention:** clear the tuple and set terminal state before killing; exercise direct request-hook and real error-cleanup paths.
- **False completeness:** retain explicit residual scope for mitmproxy's pre-hook work and aggregate request-head bytes.

## Acceptance criteria

- Define an inclusive 2,048-field raw request-header limit beside the existing 4,096-byte name limit.
- Check tuple length before the per-name traversal, private-header capture, case-insensitive lookup, body-framing checks, classification, auth, logging, or upstream binding.
- At 2,049 fields, release the complete raw tuple, mark the header phase terminal, and kill without inspecting individual fields.
- At exactly 2,048 fields, preserve current behavior for duplicates, mixed-case fixed names, `Host`, `Content-Length`, `Transfer-Encoding`, private prefetch/intent markers, and the separate field-name boundary.
- Add real-flow structural coverage using valid tiny fields and a `bytes` subtype whose `lower()` raises.
- Add a real pinned-mitmproxy HTTP/1 layer test proving an open body with `Expect: 100-continue` terminates after the header hook without a 100 response, request hook, body buffering, or upstream command.
- Record a before/after diagnostic for dense 60 KB through 3 MB request heads; keep benchmark code and wall-clock assertions out of CI.
- Run the complete committed mitm-addon Python validation sequence.

## Compatibility and residual exposure

The change is runner-local and changes no API, queue payload, environment variable, persisted state, or cross-deployment protocol. Old and new runners can overlap because each handles its own proxy flows; rollback is an ordinary runner artifact rollback.

This does not eliminate mitmproxy's pre-hook HTTP/1 buffering, parsing, validation, or initial header allocation, and it does not establish a generic aggregate request-head byte limit.

seven332:
## Summary

- cap generic raw request headers at an inclusive 2,048 fields before per-name checks or repeated header scans
- release the complete over-limit tuple and immediately kill the flow before mitmproxy's `Expect` lookup or request-body processing
- preserve exact-limit behavior for both `Content-Length` and `Transfer-Encoding`, and keep the later SigV4-specific field-count contract covered at its independent request-hook entry
- cover guarded no-normalization rejection and real HTTP/1 `Expect: 100-continue` termination through pinned mitmproxy hooks

## Rejection behavior

Over-limit requests terminate the client connection or protocol stream instead of receiving HTTP 431. Pinned mitmproxy delays a local response created in `requestheaders()` until the request body completes. Retaining or locating `Expect` would rescan the attacker-sized tuple, while clearing it without terminating can leave the client waiting for `100 Continue`.

## Explicit exclusions

- does not bound mitmproxy's pre-hook HTTP/1 buffering, parsing, validation, or initial header allocation
- does not add a generic aggregate request-head byte limit
- does not add configuration, logging, or a new externally visible error contract
- does not change APIs, queue payloads, persisted data, environment variables, or cross-deployment protocols

## Diagnostic

Genuine mitmproxy flows were constructed outside timing under the locked CPython 3.14 environment. Each row is the median of seven fresh-flow calls. The previous column executes the exact accepted common prefix through framing checks; the new column includes tuple release and flow termination.

| Repeated tiny fields | Approximate HTTP/1 head | Previous prefix | New rejection |
| ---: | ---: | ---: | ---: |
| 10,000 | 60,056 bytes | 3.447 ms | 0.024 ms |
| 50,000 | 300,056 bytes | 17.941 ms | 0.076 ms |
| 100,000 | 600,056 bytes | 39.051 ms | 0.127 ms |
| 250,000 | 1,500,056 bytes | 103.817 ms | 0.328 ms |
| 500,000 | 3,000,056 bytes | 206.823 ms | 0.594 ms |

Separate `tracemalloc` runs after constructing the 500,000-field input observed 7.790 MiB of additional peak allocation in the previous prefix and 0.0003 MiB in the new rejection path.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest -q tests/test_request_header_names.py tests/test_request_handler_aws_sigv4_body.py` (42 passed)
- `uv run --no-sync python -m pytest tests/` (7,272 passed)

Closes #31645

### DIFF
--- a/crates/runner/mitm-addon/src/mitm_addon.py
+++ b/crates/runner/mitm-addon/src/mitm_addon.py
@@ -125,6 +125,8 @@ _HTTP_STATUS_ERROR_MIN = 400  # inclusive: start of 4xx/5xx error range
 # Release: auth marker is popped by terminal cleanup.
 # _REQUEST_HEADERS_TERMINATED is a flow-local sentinel for request() early exit.
 _HTTP_STATUS_REQUEST_HEADER_FIELDS_TOO_LARGE = 431
+# Match the existing 64 KiB HTTP/2/SigV4 minimum per-field accounting budget.
+_MAX_REQUEST_HEADER_FIELDS = 2048
 _MAX_REQUEST_HEADER_NAME_BYTES = 4096
 _REQUEST_HEADERS_TERMINATED = "_request_headers_terminated"
 _FIREWALL_AUTH_APPLIED_IN_REQUESTHEADERS = "_firewall_auth_applied_in_requestheaders"
@@ -748,6 +750,11 @@ def requestheaders(flow: http.HTTPFlow) -> Awaitable[None] | None:
     """Handle request-header-only decisions before mitmproxy buffers bodies."""
     request_end_stream = mitmproxy_compat.take_request_end_stream(flow)
     request_header_fields = flow.request.headers.fields
+    if len(request_header_fields) > _MAX_REQUEST_HEADER_FIELDS:
+        flow.request.headers.fields = ()
+        flow.metadata[_REQUEST_HEADERS_TERMINATED] = True
+        flow.kill()
+        return None
     if any(len(name) > _MAX_REQUEST_HEADER_NAME_BYTES for name, _value in request_header_fields):
         # Mitmproxy performs an Expect lookup after this hook, so rejected names
         # must be gone before control returns while ordinary protocol fields stay.
--- a/crates/runner/mitm-addon/src/mitm_addon.py
+++ b/crates/runner/mitm-addon/src/mitm_addon.py
@@ -751,6 +751,8 @@ def requestheaders(flow: http.HTTPFlow) -> Awaitable[None] | None:
     request_end_stream = mitmproxy_compat.take_request_end_stream(flow)
     request_header_fields = flow.request.headers.fields
     if len(request_header_fields) > _MAX_REQUEST_HEADER_FIELDS:
+        # A local response waits for body completion. Kill before mitmproxy's
+        # Expect lookup instead of retaining or rescanning the rejected tuple.
         flow.request.headers.fields = ()
         flow.metadata[_REQUEST_HEADERS_TERMINATED] = True
         flow.kill()
--- a/crates/runner/mitm-addon/tests/test_request_handler_aws_sigv4_body.py
+++ b/crates/runner/mitm-addon/tests/test_request_handler_aws_sigv4_body.py
@@ -536,7 +536,7 @@ async def test_sigv4_request_header_list_inspection_boundary(
     assert aws_sigv4_body_admission.state_for_tests() == (0, 0)
 
 
-async def test_sigv4_request_header_field_count_fails_closed(
+async def test_sigv4_request_hook_header_field_count_fails_closed(
     tmp_path,
     real_flow,
     headers,
@@ -562,7 +562,6 @@ async def test_sigv4_request_header_field_count_fails_closed(
         mitm_ctx(registry_path=str(registry_path), api_url="https://api.vm0.ai"),
         patch.object(auth, "get_firewall_headers", get_headers),
     ):
-        assert mitm_addon.requestheaders(flow) is None
         await mitm_addon.request(flow)
 
         assert flow.response is not None
--- a/crates/runner/mitm-addon/tests/test_request_header_names.py
+++ b/crates/runner/mitm-addon/tests/test_request_header_names.py
@@ -1,5 +1,6 @@
 """Request-header field-count and field-name admission integration tests."""
 
+import pytest
 from mitmproxy import connection
 from mitmproxy.addons.proxyserver import Proxyserver
 from mitmproxy.flow import Error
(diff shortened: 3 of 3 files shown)


seven332:
## Code Review: PR #31775

**Reviewed head:** `7105545ca06bda7e238052bd7f5213b739bf3a19`

### Summary

The addon now caps raw request fields at 2,048 before any per-field work, releases the complete rejected tuple, and terminates before mitmproxy can inspect `Expect` or consume a body. Exact-limit, overlong-name, private-header, downstream SigV4, and real HTTP/1 lifecycle behavior remain covered.

### Findings

No P0, P1, or P2 findings.

### Testing

Locked uv dependency checks, Ruff format/lint, BasedPyright, focused request-header/SigV4 pytest (41 passed), and the complete mitm-addon suite (7,271 passed) succeeded. A separate 10,000–500,000-field diagnostic showed repeated prefix work removed; mitmproxy's pre-hook HTTP/1 parsing/allocation and generic aggregate-byte admission remain intentionally unverified/out of scope.

### Verdict

LGTM

PR #31875, opened 2026-09-05
fix(runner): reject non-http pi model base urls

### LINKED ISSUE #31865: bug(rust): reject non-HTTP Pi model base URLs before sandbox start
## Problem and evidence

The runner's pre-sandbox Pi model validation accepts URL schemes that the model request transport cannot use.

`crates/runner/src/executor/env.rs` parses model `baseUrl` values in `validate_pi_model_config_common`, but it does not check the parsed scheme. Syntactically valid values such as `ftp://provider.example/v1` and `file:///tmp/model` therefore pass validation. The shared validator is reached by all three generations currently advertised in runner claim capabilities:

- legacy generation 1, represented by a config without `schemaVersion`;
- generation 2 through `validate_pi_model_config_v2` and `validate_pi_versioned_model_config`;
- generation 3 through `validate_pi_model_config_v3` and the same versioned validator.

Generations 2 and 3 each support public Responses and Codex Responses dialect shapes. The adjacent Pi API first-turn URL validation already limits its fetch targets to HTTP and HTTPS. The TypeScript Pi model schemas use `z.url()`, which validates URL syntax but does not make the runner's transport capability invariant explicit.

## Failure scenario

1. The API supplies an otherwise valid Pi execution context whose model `baseUrl` uses a parseable unsupported scheme such as `ftp:` or `file:`.
2. `validate_execution_context_before_sandbox` accepts the context and builds the private run payload.
3. The original model config is forwarded into the sandbox.
4. The first model request fails because the fetch-based model transport supports HTTP(S), after sandbox startup has already been consumed.

## Decision

Proceed with a runner validation fix. The pre-sandbox boundary must reject model routes that the sandbox transport cannot use, independently of upstream TypeScript parsing.

## Scope

Apply the scheme restriction in the existing shared model-config validator so generations 1, 2, and 3 cannot drift. Preserve custom HTTP and HTTPS gateway URLs and the existing error contract. No API endpoint, persisted data, wire-version, or migration change is needed.

## Acceptance criteria

- Pi model `baseUrl` values must use `http:` or `https:` at the runner's shared pre-sandbox validator.
- Custom HTTP and HTTPS gateways remain accepted.
- Entry-point coverage exercises valid legacy, generation-2, and generation-3 configs.
- Every supported generation rejects at least `ftp:` and `file:` base URLs before sandbox execution.
- Positive coverage includes both HTTP and HTTPS routes.
- Existing malformed-URL diagnostics and raw payload preservation remain compatible.

## Non-goals

- Adding or activating another Pi model-config generation.
- Changing the TypeScript wire schemas or API route-building behavior.
- Restricting custom gateways to HTTPS-only.

seven332:
## Summary

- reject parseable Pi model base URLs whose schemes are not HTTP or HTTPS at the runner's shared pre-sandbox validator
- cover legacy generation 1 plus both dialects in generations 2 and 3 with HTTP, HTTPS, FTP, and file URL cases
- preserve the existing invalid-base-URL diagnostic and raw model payload behavior

## Exclusions

- no TypeScript schema or API route-building changes
- no HTTPS-only restriction for custom gateways
- no wire generation, persisted state, or migration changes

## Validation

- `cd crates && cargo xtask check-rust-path-attributes`
- `cd crates && cargo clippy -p runner --all-targets --all-features`
- `cd crates && cargo doc -p runner --all-features --no-deps`
- `cd crates && cargo test -p runner pi_execution_context_restricts_model_base_url_schemes_before_sandbox`
- `cd crates && cargo test -p runner`
- `git diff --check`

Closes #31865

### DIFF
--- a/crates/runner/src/executor/env.rs
+++ b/crates/runner/src/executor/env.rs
@@ -220,7 +220,10 @@ fn validate_pi_model_config_common(
     model: &str,
     catalog_model: Option<&serde_json::Value>,
 ) -> Result<(), PiModelConfigCommonError> {
-    url::Url::parse(base_url).map_err(|_| PiModelConfigCommonError::InvalidBaseUrl)?;
+    let parsed = url::Url::parse(base_url).map_err(|_| PiModelConfigCommonError::InvalidBaseUrl)?;
+    if !matches!(parsed.scheme(), "http" | "https") {
+        return Err(PiModelConfigCommonError::InvalidBaseUrl);
+    }
     if model.is_empty() {
         return Err(PiModelConfigCommonError::EmptyModel);
     }
--- a/crates/runner/src/executor/tests/env_tests.rs
+++ b/crates/runner/src/executor/tests/env_tests.rs
@@ -1507,6 +1507,49 @@ fn pi_execution_context_rejects_invalid_legacy_shared_model_fields_before_sandbo
 }
 
 #[test]
+fn pi_execution_context_restricts_model_base_url_schemes_before_sandbox() {
+    let v2_public = pi_model_config_v2_for_test("openai-responses");
+    let v2_codex = pi_model_config_v2_for_test("openai-codex-responses");
+    let mut v3_public = v2_public.clone();
+    v3_public["schemaVersion"] = json!(3);
+    let mut v3_codex = v2_codex.clone();
+    v3_codex["schemaVersion"] = json!(3);
+    let configs = [
+        pi_model_config_for_test(),
+        v2_public,
+        v2_codex,
+        v3_public,
+        v3_codex,
+    ];
+
+    for (base_url, supported) in [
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #31875

**Reviewed head:** `aff1db73e59f53d31026372947bf064cd5ba0eb2`

### Summary

The shared pre-sandbox Pi model validator now rejects parsed URL schemes other than HTTP and HTTPS while preserving the existing diagnostic and supported custom gateways. The regression matrix covers legacy plus both dialects in generations 2 and 3.

### Findings

No findings.

### Testing

Runner formatting, Rust path validation, Clippy, documentation, the focused regression test, the full runner test suite, diff checks, and repository commit hooks passed. Production sandbox execution was not run; this validation fails before sandbox creation and is covered through that entry point.

### Verdict

LGTM

PR #31877, opened 2026-09-05
perf(guest-agent): bound overlapping match ranges

### LINKED ISSUE #31849: perf(rust): bound overlapping secret-match range collection
## Problem and evidence

At current `main` (`6040bada98`), secret masking retains every overlapping Aho-Corasick match before computing their union:

- `crates/guest-agent/src/masker.rs:479-500` appends one `Range<usize>` for every result from `find_overlapping_iter`.
- `crates/guest-agent/src/masker.rs:502-523` sorts the full raw-match vector and allocates a second vector with the same capacity even when every match collapses into one redaction range.

A valid five-byte secret such as `aaaaa` against an `N`-byte repeated-`a` string produces approximately `N` matches. On a 64-bit guest, the first range vector alone reserves approximately `16 * N` bytes although the final union is only `0..N`.

This work is synchronous on event ingestion: `events::prepare_event_for_delivery` masks parsed events before Codex delivery reduction or event-buffer admission. Ordinary CLI JSONL accepts 16 MiB records and Codex app-server JSONL accepts 64 MiB records.

An isolated release-mode `SecretMasker::from_raw(base64("aaaaa")).mask_string(&"a".repeat(N))` harness measured:

| Input | Elapsed | Peak RSS |
| --- | ---: | ---: |
| 4 MiB | 406 ms | 72,204 KiB |
| 8 MiB | 611 ms | 141,832 KiB |
| 16 MiB | 812 ms | 281,100 KiB |

The 16 MiB output is three bytes, so the memory is match-cardinality amplification rather than retained output. This is the measured performance follow-up explicitly deferred by #24621/#24639 when overlapping redaction correctness was introduced.

## Decision

Proceed with one `guest-agent` Rust PR. Compact each matcher's overlapping results online, then linearly merge the compact normal and lowercase-percent canonicalized URL range lists.

- Aho-Corasick 1.1.5 overlapping search exhausts all matches at one search position before advancing, so one matcher iterator has nondecreasing match end offsets. Pattern order among equal-end matches is not assumed.
- Fold each matcher stream into sorted, strictly disjoint union components. A new match merges every overlapping suffix component, including a later long match that reaches backward across previously disjoint ranges.
- Compact original and canonicalized URL scans independently because the second iterator restarts offsets at zero.
- Linearly merge the two sorted component lists and collapse duplicates or true overlaps. Keep merely adjacent ranges separate through the existing strict `next.start < current.end` rule.
- Retained range storage becomes `O(R)` for final disjoint components instead of `O(M)` raw matches; the unavoidable matcher enumeration remains `O(M)`, and raw-range sorting is removed.
- Add a 16 MiB self-overlap regression that verifies the single range, final `***`, and a loose constant capacity bound. Add backward-bridge and cross-stream interleaving regressions.

This removes the measured amplification without making sparse inputs allocate by maximum haystack size, adding a dependency, or replacing the downstream range representation.

## Alternatives considered

- **Input-sized coverage events:** independent of result order, but a 64 MiB record can require hundreds of MiB depending on counter width. A basic difference array also loses adjacent boundaries when an end and start share one offset.
- **General ordered interval set:** independent of result order and retains `O(R)` entries, but adds per-node allocations, `O(M log R)` work, and conversion back to the existing vector. A sorted-vector variant risks quadratic middle insertion when URL ranges interleave.
- **Fixed-size chunking:** bounds raw matches only when pattern length is also bounded below the chunk size. Current secrets have no matching upper bound, so correct overlap can approach the full input and adds duplicate scanning and offset-boundary logic.
- **In-place compaction after raw collection:** removes the second capacity reservation but retains the dominant `O(M)` first vector and `O(M log M)` sort, so it does not satisfy the issue.
- **Status quo:** leaves measured guest memory and stdout-consumption latency amplification on valid input.

## Risks and mitigations

- **Incorrect union could expose secret bytes:** retain all existing containing, partial, chained, self-overlap, diagnostic, encoded, JSON-key, UTF-8, and pattern-order tests; add a multi-component backward bridge.
- **Dependency ordering assumption:** document only the nondecreasing-end scan invariant at the folding boundary and remain insensitive to equal-end pattern order.
- **Two matcher streams can restart ordering:** never append raw second-stream matches to the first fold; compact separately and linearly merge.
- **Adjacent matches could collapse:** retain strict-overlap comparison and existing adjacent normal and encoded regressions.
- **Performance test flakiness:** assert the retained representation rather than wall time. Use release timing and peak RSS only as an implementation measurement, not a CI threshold.

## Acceptance criteria

- Preserve complete redaction for containing, partially overlapping, chained, self-overlapping, and backward-bridging patterns across normal, diagnostic, and URL-encoded paths.
- Preserve adjacent-match output, duplicate collapse, byte-length percent canonicalization, and unmatched-string allocation behavior.
- Retain range memory proportional to disjoint union components, independent of raw overlapping-match count.
- A 16 MiB repeated-`a` input with valid secret `aaaaa` produces exactly one `0..N` range with capacity below a loose constant bound and redacts to `***`.
- Mixed original and canonicalized URL ranges merge in input order without duplicate replacements.
- The 16 MiB release reproduction shows materially lower peak RSS and latency than 281,100 KiB / 812 ms, noting host differences.
- Focused and full `guest-agent` tests, Rust formatting, Clippy, documentation checks, and applicable staged-path hooks pass.

## Delivery notes

One PR changes `crates/guest-agent/src/masker.rs` and its existing security-critical test module. There is no schema, persistence, API, queue, dependency, configuration, feature switch, migration, runner/guest protocol, rollout gate, or follow-up slice.

seven332:
## Summary

- compact each Aho-Corasick overlapping-match stream into sorted connected ranges while scanning
- linearly merge original-input and canonicalized-URL ranges without retaining or sorting every raw match
- preserve strict adjacency semantics and cover equal-end, bridging, cross-stream, and 16 MiB self-overlap regressions

## Performance

For the issue's 16 MiB repeated-`a` case with secret `aaaaa`, the focused release test completed in 82 ms with 21,344 KiB peak RSS on this development host. Compared with the issue's earlier 812 ms / 281,100 KiB measurement, that is about 9.9x faster with about 92% lower peak RSS. The measurements were collected on different hosts, so they are directional rather than a controlled benchmark comparison.

## Compatibility

No API, schema, protocol, queue payload, dependency, or configuration changes.

## Validation

- [x] focused red/green regression tests
- [x] full `guest-agent` test suite
- [x] `cargo fmt --manifest-path crates/Cargo.toml --all -- --check`
- [x] `cargo clippy --manifest-path crates/Cargo.toml --profile local -p guest-agent --all-targets --all-features`
- [x] `cargo doc --manifest-path crates/Cargo.toml --profile local -p guest-agent --all-features --no-deps`
- [x] `git diff --check`
- [x] repository pre-commit hooks

Closes #31849

### DIFF
--- a/crates/guest-agent/src/masker.rs
+++ b/crates/guest-agent/src/masker.rs
@@ -483,44 +483,66 @@ fn redaction_ranges(
 ) -> Vec<Range<usize>> {
     let canonicalized =
         url_encoded_matcher.and_then(|_| canonicalize_lowercase_percent_escapes(haystack));
-    let mut ranges = Vec::new();
-    if let Some(matcher) = matcher {
-        push_match_ranges(matcher, haystack, &mut ranges);
-    }
+    let ranges = matcher
+        .map(|matcher| merged_match_ranges(matcher, haystack))
+        .unwrap_or_default();
     if let (Some(matcher), Some(canonicalized)) = (url_encoded_matcher, canonicalized.as_deref()) {
-        push_match_ranges(matcher, canonicalized, &mut ranges);
+        return merge_sorted_ranges(ranges, merged_match_ranges(matcher, canonicalized));
     }
-    merge_overlapping_ranges(ranges)
+    ranges
 }
 
-fn push_match_ranges(matcher: &AhoCorasick, haystack: &str, ranges: &mut Vec<Range<usize>>) {
+fn merged_match_ranges(matcher: &AhoCorasick, haystack: &str) -> Vec<Range<usize>> {
+    let mut ranges = Vec::new();
+    let mut previous_end = 0;
     for matched in matcher.find_overlapping_iter(haystack) {
-        ranges.push(matched.start()..matched.end());
+        // Overlapping search reports every match at one haystack position
+        // before advancing, so match ends are nondecreasing. Match starts and
+        // equal-end pattern order are not ordered.
+        debug_assert!(matched.end() >= previous_end);
+        previous_end = matched.end();
+        push_merged_range(&mut ranges, matched.start()..matched.end());
     }
+    ranges
 }
 
-fn merge_overlapping_ranges(mut ranges: Vec<Range<usize>>) -> Vec<Range<usize>> {
-    if ranges.len() < 2 {
-        return ranges;
+fn merge_sorted_ranges(first: Vec<Range<usize>>, second: Vec<Range<usize>>) -> Vec<Range<usize>> {
+    if first.is_empty() {
+        return second;
+    }
+    if second.is_empty() {
+        return first;
     }
 
-    ranges.sort_by_key(|range| (range.start, range.end));
-    let mut merged: Vec<Range<usize>> = Vec::with_capacity(ranges.len());
-    for range in ranges {
-        if range.is_empty() {
-            continue;
-        }
-        if let Some(last) = merged.last_mut()
-            && range.start < last.end
-        {
-            if range.end > last.end {
-                last.end = range.end;
+    let mut merged = Vec::with_capacity(first.len() + second.len());
+    let mut first = first.into_iter().peekable();
+    let mut second = second.into_iter().peekable();
+    loop {
+        let next = match (first.peek(), second.peek()) {
+            (Some(left), Some(right)) => {
+                if (left.start, left.end) <= (right.start, right.end) {
+                    first.next()
+                } else {
+                    second.next()
+                }
             }
-            continue;
+            (Some(_), None) => first.next(),
+            (None, Some(_)) => second.next(),
+            (None, None) => break,
+        };
+        if let Some(range) = next {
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #31877

**Reviewed head:** `3fa0eddf066f8a2f58281b1feff4e11ce1438ea6`

### Summary

The online compaction preserves the previous strict-overlap union semantics while removing raw-match retention and sorting. The independently scanned URL-canonicalized ranges are compacted before a correct linear merge.

### Findings

No findings.

### Testing

Focused regressions, the full `guest-agent` suite, Rust formatting, Clippy, documentation checks, repository hooks, a 200,000-case differential interval check, and the focused release reproduction passed locally. Broader GitHub CI is still in progress.

### Verdict

LGTM

Issue #31867, opened 2026-09-05
chore(tooling): classify the infrastructure and script brand names in the r5 baseline


lancy:
## Parent

Part of #31801 (phase-3 continuation of #26873), workstream **R5a**, tier P3. First of two R5 slices, and the zero-risk one: it changes no product code at all.

## Background

#31813's guard assigns **142 names** to R5, described as "production-code residual identifiers". Locating every one of them shows that description is wrong for about half of them. Grouped by where they actually live:

| Group | Names | What they are |
|---|---:|---|
| Infrastructure, brand, and wire string keys | 58 | Token prefixes, KMS encryption context, Stripe metadata keys, Axiom dataset names, deployment identifiers, brand prose in contract JSDoc |
| `packages/db/scripts/**` | 15 | SQL text and physical names inside migration and verification scripts |
| Tokenizer fragments | 2 | `zero_` and `ZERO_`, matched inside shell test scripts |
| Platform browser globals | 10 | `_vm0`, `VM0Global`, `__vm0ClerkBootstrap` and friends |
| Genuine production TypeScript identifiers | 57 | `ZeroChatEventRowsPage`, `RetiredZeroScopePayload`, `vm0ImageStyleSource`, `platformVm0LogoImg`, … |

**This issue owns the first three groups — 75 names — and classifies them. It renames nothing.** The 57 real identifiers and the 10 browser globals stay in the baseline for the follow-up rename slice, which carries real risk and deserves its own review.

## Design

For each of the 75 names, produce one of two outcomes:

1. **A boundary rule** in `turbo/packages/eslint-rules/scripts/residual-brand-name-manifest.ts`, using an existing category from `RESIDUAL_BRAND_BOUNDARY_CATEGORIES` and a reason that states *why this specific name cannot change*. Delete its baseline entry.
2. **A baseline entry moved to its correct workstream**, if the name turns out to be an ordinary internal identifier or belongs to a slice that already owns it.

Prove each disposition. A category without evidence is a blanket approval, and #31813's tests reject rules that match nothing.

## The trap

`vm0-stored-secret` at `apps/api/src/signals/services/crypto.utils.ts:40` is the **KMS encryption context**:

```ts
const KMS_ENCRYPTION_CONTEXT = { purpose: "vm0-stored-secret" } as const;
```

[...]

Issue #31842, opened 2026-09-05
refactor(api): rename the zero dispatch telemetry action names to agent


lancy:
## Parent

Part of #31801 (phase-3 continuation of #26873), workstream **R2**, tier P2. Last slice in track C; waited for #26877 and #31818, which edited the same BDD test files.

## Background

`apps/api/src/signals/services/api-dispatch-timing.service.ts` declares 127 `api_dispatch_pre_create_zero_*` action strings, asserted across 31 files, 457 occurrences. They are emitted through `signals/external/sandbox-op-log.ts:55` as the `op_type` field of the Axiom dataset `vm0-sandbox-op-log-<suffix>`.

The `zero` segment is not decoration — it distinguishes the product run-creation path from the direct one. The same union already contains `api_dispatch_pre_create_direct_parse_body` and the umbrella `api_dispatch_pre_create_agent_run`. So the vocabulary is already mixed: the umbrella says `agent`, its children say `zero`.

Two op-log action types outside that prefix belong to the same class and were deliberately left for this issue by #26877: `enqueue_zero_run` (`agent-run-create.service.ts:6723`) and `dequeue_zero_run` (`run-queue.service.ts:224`).

## Measured before deciding

Production, 3-day window:

| Fact | Value |
|---|---|
| Dataset retention | **3 days** |
| Total op-log events | 6,709,576 |
| Events carrying a `..._zero_...` action | 500,606 (7.5%) |
| Declared names / names that actually fire | 127 / **86** |
| Distinct `api_dispatch*` action types in prod | 186 |
| `api_dispatch_pre_create_agent_run` (canonical sibling) | 18,699 |
| `enqueue_zero_run` / `dequeue_zero_run` | 37 / 35 |

## Decision

**Rename `_zero_` → `_agent_` in one PR. No dual-emit window.**

Retention is 3 days, so the mixed-data window is 72 hours, not months. A saved query that needs to span the rename can use `op_type startswith 'api_dispatch_pre_create_'` for those three days; after that, only the new names exist. Dual-emit would add roughly 1.2M events per week and require a second cleanup PR to buy three days of overlap.

Verified: renaming `_zero_` → `_agent_` collides with nothing. None of the 127 resulting names already exists, and `enqueue_agent_run` / `dequeue_agent_run` are free.

## Residual risk, stated plainly

[...]

PR #31881, opened 2026-09-05
chore(tooling): classify the unrenameable identifiers left in the r5 baseline

### LINKED ISSUE #31880: chore(tooling): classify the unrenameable identifiers left in the r5 baseline
## Parent

Part of #31801 (phase-3 continuation of #26873), workstream **R5b**, tier P3. First of the three remaining R5 slices, and the one that must land first: it removes the mines from the two mechanical rename slices that follow.

## Background

#31867 classified the infrastructure, wire, and script names out of R5, leaving **76 names** described as "production TypeScript identifiers and platform browser globals" for a rename slice. Locating each one shows that six of them cannot be renamed at all, and two more need a fact established before anyone touches them. Renaming them as part of a mechanical sweep would cause a production defect, not a naming improvement.

This issue classifies those eight. **It renames nothing.**

### 1-2. Reserved keys inside persisted queued payloads

`apps/api/src/signals/services/workflow-automation-context.service.ts:25,28-29`

```ts
const EVENT_PAYLOAD_OBJECT_KEY_ORDER = "__vm0EventPayloadObjectKeyOrderV1";
// Queued rows admitted during the rollout may still contain this reserved
// field. Keep stripping it until those persisted payloads have drained.
const LEGACY_AUTOMATION_MESSAGE_VARIANT_FIELD =
  "__vm0UserFriendlyAutomationMessageV1";
```

`EVENT_PAYLOAD_OBJECT_KEY_ORDER` is not read-only: line 135 **writes** it into the payload, 126 and 144 read it back, and 93 strips it before rendering. The key therefore sits inside automation event payloads already persisted in queued rows. Rename the constant and the new code stops recognising the old key: it is neither stripped nor read, so it leaks into rendered event payloads.

`LEGACY_AUTOMATION_MESSAGE_VARIANT_FIELD` is already a compatibility reader with the drain condition stated in the comment above it.

Both are `wire-and-persisted-value`. Record the drain gate for each: the condition under which no queued row can still carry the key.

### 3-4. Postgres advisory lock keys

`apps/api/src/signals/services/browser.service.ts:823,976`

```ts
sql`SELECT pg_advisory_xact_lock(hashtext('zero_browser:' || ${chatThreadId}))`
const lockKey = `zero_browser_profile:${chatThreadId}`;
```

The lock identity is `hashtext` of this string. During a rolling deploy, instances running old and new code compute **different lock keys for the same chat thread**, so both acquire successfully and the mutual exclusion this lock exists to provide disappears for the length of the rollout. The failure is a concurrency defect that no test would catch and that only appears in production, during a deploy.

These are renameable in principle, but only through a deliberate transition — take both locks during one release, then drop the old one — which is a design change, not a rename. Classify them as boundaries and record that transition as the removal condition.

### 5-6. Cross-language generated sentinel values

`packages/api-contracts/src/contracts/model-provider-firewalls.ts:77` defines `CHATGPT_ACCOUNT_ID: "ws_VM0_PLACEHOLDER_DO_NOT_TRUST"` and its refresh-token sibling; `crates/api-contracts/src/generated/constants.rs:57,61,85,89` carries the same literals, produced through `packages/api-contracts/src/rust-bindings/generate.ts`. They are sentinel credential values that reach the runner. Renaming requires regenerating the Rust side and landing both together, and any runner comparing against the old literal during a mixed-version window sees a value it does not recognise.

## The two that need a fact established first

| Name | Where | Question to answer |
|---|---|---|
| `vm0-host` | `packages/connectors/src/firewall-types.ts:1488,2524` — the reserved `vm0-host.invalid` authority the firewall base-URL parser substitutes | Is a normalized authority containing this placeholder ever **persisted** in a stored firewall policy, or is it only computed per request on both sides? Persisted means boundary; computed means the rename slice can take it |
| `_vm0Cursor` | `apps/api/src/signals/services/log-pagination.ts:20,33,340` — the APL projection field the service writes into its own query and reads back off Axiom rows | Confirm nothing outside this repository selects on it. It is the same class as #31842's action names: safe in-repo, but an external saved query breaks silently |

Record the answer either way. A "verified, safe to rename" note is as valuable as a boundary rule, and the rename slice needs it.

Note for `vm0-host`: the same token also names `/tmp/vm0-host-cpu-fairness-*`, the systemd unit `vm0-host-cpu-managed-*`, and `/run/lock/vm0-host-cpu-fairness` in `.github/scripts/runner-behavior-host-cpu-fairness.sh`. Those are created and torn down inside one script run, so they carry no cross-version risk — do not let them confuse the disposition of the `.invalid` placeholder.

## Non-goals

- **Rename nothing.** Not one product file changes. This is a manifest slice.
- Do not touch the other 68 R5 names; the two rename slices that follow own them.
- Do not design or implement the advisory-lock transition here. Record it as a removal condition.
- Do not weaken an existing rule or widen one to cover several names at once.
- Do not touch `.github/workflows/**`.

## Acceptance criteria

- The six names are boundary rules with categories from the existing vocabulary and reasons that state the specific production consequence — payload key leakage, lock-identity divergence during a rolling deploy, and mixed-version sentinel mismatch.
- Each of the two persisted-payload keys carries an explicit drain gate; each lock key carries the take-both-then-drop transition as its removal condition.
- `vm0-host` and `_vm0Cursor` each end with a recorded answer: a boundary rule if persisted or externally selected, or a note in the baseline reason stating it was verified safe for the rename slice.
- `git diff` shows the manifest only. No product source file is modified.
- `Test Residual Brand Name Guard` passes, and no new rule matches zero occurrences.

lancy:
Closes #31880.

#31867 left **76 names** in R5 described as "production TypeScript identifiers and platform browser globals". Locating them shows six cannot be renamed at all and two needed a fact established first. This PR classifies those eight and **renames nothing** — `git diff` touches one file, `turbo/packages/eslint-rules/scripts/residual-brand-name-manifest.ts`. R5 now holds **70 names** for the two rename slices that follow.

## The six that cannot be renamed

| Rule | Names | Production consequence |
|---|---|---|
| `wire-and-persisted-value/queued-automation-payload-key-order-field` | `__vm0EventPayloadObjectKeyOrderV1` | queued automations discarded |
| `wire-and-persisted-value/queued-automation-legacy-message-field` | `__vm0UserFriendlyAutomationMessageV1` | reserved key leaks into the agent prompt |
| `wire-and-persisted-value/advisory-lock-key` | `zero_browser`, `zero_browser_profile` | lock identity diverges during a rolling deploy |
| `wire-and-persisted-value/model-provider-placeholder-sentinel` | `rt_VM0_PLACEHOLDER_DO_NOT_TRUST`, `ws_VM0_PLACEHOLDER_DO_NOT_TRUST` | mixed-version sentinel mismatch at the runner |

### One correction to the issue's account of the key-order field

The issue predicts that renaming `EVENT_PAYLOAD_OBJECT_KEY_ORDER` leaks the old key into rendered event payloads. Tracing the read path shows a different and worse failure. `restoredWorkflowAutomationEventPayload` (`workflow-automation-context.service.ts:140-148`) parses `payload[EVENT_PAYLOAD_OBJECT_KEY_ORDER]` **first**; a queued row written under the old key makes that `safeParse` fail, so the function returns `null` before the strip at `:93` is ever reached. `buildWorkflowAutomationQueuedLaunchMaterial` then returns `null` and `workflow-queue-drain.service.ts:460-466` takes the `consumeInvalidAutomationEvent` branch. Every automation admitted before the deploy is **consumed as invalid and never runs**, logged as `Consuming workflow queue event with incomplete context`. The rule records that consequence.

The leak the issue describes is real for the *other* key: `__vm0UserFriendlyAutomationMessageV1` is strip-only, and `workflowAutomationAgentPrompt` renders the restored payload with `JSON.stringify(context.event, null, 2)` (`:795`), so an unstripped reserved key lands in the agent prompt.

### Drain gate for both persisted keys

Both are written into `chat_automation_context.event_payload` and read only while the event is still queued. The gate, stated on each rule against the persisted-state surface of `docs/fallback.md` §7:

> no unrevoked `input.automation` chat event with a null `run_id` has a `chat_automation_context.event_payload` still carrying the key

That is exactly `pendingChatQueueEventCondition` (`chat-event-queue.service.ts:88-94`) — `run_id IS NULL` plus unrevoked — so the gate is checkable with one query. It is also the condition the comment above `LEGACY_AUTOMATION_MESSAGE_VARIANT_FIELD` already names in prose.

### Removal condition for the lock keys

The lock identity is `hashtext` of the string, so during a rolling deploy old and new instances derive different keys for the same `chatThreadId`, both acquire, and the mutual exclusion is gone for the length of the rollout. No test observes it; it needs two code versions running at once. The rule records the transition rather than designing it: **one release takes both the old and the new lock in the same transaction, and only the release after the instances holding both have finished draining may drop the old one.**

## The two that needed a fact — both answered, both safe

**`vm0-host` — not persisted, the rename slice can take it.** Both sites substitute `vm0-host.invalid` into a throwaway `URL`:

- `firewall-types.ts:2519-2533` — `authorityForUrlSyntaxValidation` feeds `urlForCanonicalBaseSyntax`, whose three callers (`:886`, `:3024`, `:3212`) read only `protocol`, `search`, `hash`, and `hostname`, and read `hostname` only when `rawAuthorityFromBaseUrl` returned `null`, i.e. when no substitution happened. The canonical base `validateAndCanonicalizeBaseUrl` returns — the value a stored firewall policy freezes — is built from `canonicalizeAuthorityForExecution` on the **real** authority (`:3016-3019`, `:3055`).
- `firewall-types.ts:1466-1490` — `urlForHostPolicyValidation` returns a hostname assembled from the real host segments and only the `port` string parsed off the placeholder URL.

The literal never leaves the function, so no stored policy can contain it. The identically spelled names in `.github/scripts/runner-behavior-host-cpu-fairness.sh` are created and destroyed inside one script run and are already covered by `external-identity/ci-definitions`; they did not affect this disposition.

**`_vm0Cursor` — not externally selectable, the rename slice can take it.** `buildTimeCursorProjection` (`log-pagination.ts:293-295`) appends `| extend _vm0Cursor = cursor_current()` to the APL the service itself sends, and `:340` reads the alias back off the rows in the same request. It is not a dataset field: an external saved query, dashboard, or monitor would have to write that same `extend` to produce it, so the #31842 risk of a silent external mismatch does not reach it. The cursor handed to clients is `time:<order>:<timestamp>:<tieBreaker>` (`:171-173`) and never carries the field name, so issued cursors survive the rename.

Both answers are recorded in their own R5 baseline blocks so the decision is visible to the rename slice rather than implied.

## Verification

```
pnpm -F @okouai/eslint-rules exec vitest run scripts/residual-brand-name-inventory.test.ts
  Test Files  1 passed (1)
       Tests  10 passed (10)
```

Every new rule matches at least one real occurrence and no existing rule dropped to zero — checked by listing rules absent from `boundaryOccurrenceCounts`, which came back empty:

```
2 wire-and-persisted-value/advisory-lock-key
2 wire-and-persisted-value/model-provider-placeholder-sentinel
2 wire-and-persisted-value/queued-automation-legacy-message-field
1 wire-and-persisted-value/queued-automation-payload-key-order-field
```

The legacy-message rule is scoped to the service **and** the rollout test beside it, which spells the literal into a payload so the strip has something to strip — the same shape as the existing `renamed-feature-switch-key` rule. Every new rule uses an exact token list plus a `paths` scope; no existing rule was weakened or widened.

Also run: `pnpm -F @okouai/eslint-rules run lint`, `pnpm -F @okouai/eslint-rules run check-types`, `pnpm format` — all clean.

No product source file is modified, no name in R5 outside these eight is touched, and nothing under `.github/workflows/**` changes.


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: b582175893f2d57e0932418c1e15e237b4c7026e
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Classification-only change to `turbo/packages/eslint-rules/scripts/residual-brand-name-manifest.ts`: it moves six R5 baseline names into four new `wire-and-persisted-value` boundary rules and records a verified-safe disposition for two more, dropping R5 from 76 to 70 names. No product source file, migration, contract, or workflow definition is touched, and no identifier is renamed.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic of the changed code | Pass | Each new rule's `paths` regex was checked against the real occurrence paths; the classifier report attributes all seven new occurrences (1 + 2 + 2 + 2) and leaves no rule at zero. |
| `REVIEW.md` | Security — no secrets, injection, or missing auth | Pass | The diff adds only string literals and regexes to a lint-tooling manifest. `rt_`/`ws_VM0_PLACEHOLDER_DO_NOT_TRUST` are the repository's existing non-secret placeholder markers, already committed in `model-provider-firewalls.ts` and `crates/api-contracts/src/generated/constants.rs`. |
| `REVIEW.md` | Type safety — no `any`, no broad types | Pass | Entries stay inside `ResidualBrandBoundaryOccurrenceRule` / `ResidualBrandNameBaselineEntry` under the existing `satisfies`; `pnpm -F @okouai/eslint-rules run check-types` is clean. |
| `REVIEW.md` | Style and maintainability — no duplication, WHY comments | Pass | Each rule reuses the file's existing `{category, id, paths, reason, tokens}` shape and its reason states why the name cannot move, not what the code does. |
| `REVIEW.md` | DB/JSONB review gate | N/A | The diff changes no schema, contract, Zod response schema, or `jsonb().$type<...>()`. It only records that `chat_automation_context.event_payload` already carries these reserved keys. |
| `REVIEW.md` | Feature regression guard / feature switch | N/A | No user-facing behavior, route, or UI changes; the manifest is consumed only by the residual brand-name guard test. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions, no dynamic `import()` | Pass | The diff adds no casts, no `eslint-disable`/`@ts-*` comments, and no imports at all. |
| `docs/bad-smell.md` | Fail fast — no fallback or silent-recovery logic | Pass | The diff adds no runtime code path. The guard still throws on an unclassified name; removing six names from the baseline while adding matching rules keeps the ratchet strict rather than loosening it. |
| `docs/bad-smell.md` | No hardcoded URLs or environment-specific values | Pass | Reasons cite repository paths and issue numbers only; no URLs or env values are introduced. |
| `docs/testing.md` | Integration tests over unit tests; required coverage for the change | Pass | `scripts/residual-brand-name-inventory.test.ts` is the integration-level guard for this file — it scans the real repository tree, so every added rule and every removed baseline name is exercised end to end. `chore:` classification needs no new test file. |
| `docs/testing.md` | Mock boundary, MSW, real database, no fake timers | N/A | The diff adds no test code and the existing guard test uses no mocks, HTTP, database, or timers. |
| `docs/fallback.md` | Every fallback the diff introduces is declared, with surface, window, and removal condition | Pass | The diff introduces no fallback. It documents the removal gates for two compatibility readers that already exist in `workflow-automation-context.service.ts`, and each new rule names the surface and the removal condition, consistent with §7. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | N/A | No tests are added or changed. |
| `docs/deployment-compatibility.md` | Cross-version claims about independently deployed components are accurate | Pass | Verified against §Runner and §Backend: the runner ships as its own artifact with its own drain window, which is what the sentinel rule's mixed-version argument rests on; the advisory-lock rule's rolling-deploy argument matches the documented old/new backend overlap. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

Verification performed while reviewing, rather than taken from the PR body:

- **`__vm0EventPayloadObjectKeyOrderV1`** — `persistedWorkflowAutomationEventPayload` writes it (`:135`), `chat-event.service.ts:903-910` persists the payload into `chat_automation_context.event_payload`, and `restoredWorkflowAutomationEventPayload` (`:140-148`) parses it back. The `safeParse` runs before the strip at `:93`, so a renamed constant returns `null` and `workflow-queue-drain.service.ts:460-466` takes `consumeInvalidAutomationEvent`. The rule's stated consequence — queued automations discarded, not leaked — matches the code.
- **`__vm0UserFriendlyAutomationMessageV1`** — strip-only at `:93-94`, and `workflowAutomationAgentPrompt` renders the restored payload with `JSON.stringify(context.event, null, 2)` (`:795`), so the leak the rule describes is real.
- **Drain gate** — `pendingChatQueueEventCondition` (`chat-event-queue.service.ts:88-94`) is exactly `run_id IS NULL` plus unrevoked, so both gates are checkable with one query.
- **`zero_browser` / `zero_browser_profile`** — `hashtext` of the literal at `browser.service.ts:823` and `:976`; `lockBrowserThread` guards ten call sites across the session mutation paths, which the second commit corrected the rule's wording to say.
- **Sentinels** — `model-provider-firewalls.ts:77-78` → `contracts/model-providers.ts` → `rust-bindings/constants.ts:199` → `rust-bindings/generate.ts:38-42` → `crates/api-contracts/src/generated/constants.rs`. The generation chain the rule asserts exists.
- **`vm0-host`** — traced all four call sites of the two substitution helpers. `validateAndCanonicalizeBaseUrl` returns a canonical base built from `canonicalizeAuthorityForExecution` on the real authority (`:3016-3019`, `:3055`), and the three `urlForCanonicalBaseSyntax` callers read `hostname` only when `rawAuthorityFromBaseUrl` returned `null` — i.e. only when no substitution happened. `urlForHostPolicyValidation` returns real host segments plus a parsed port. The "never persisted" answer holds.
- **`_vm0Cursor`** — `buildTimeCursorProjection` (`log-pagination.ts:293-295`) creates the alias inside the outgoing APL and `:340` reads it back in the same request; the client cursor (`:171-173`) carries no field name. The "not externally selectable" answer holds.

Two judgement calls worth naming rather than flagging:

- `wire-and-persisted-value` for the advisory lock keys is the closest existing category — the same one already carrying `pi-handoff-control-record` and `addon-event-envelope`, both cross-process values whose two sides ship separately. `protocol-compatibility` would have been wrong, since every rule in it belongs to the #26701/#28278 legacy-scope retirement.
- The key-order rule is scoped to the service file while its sibling also covers the rollout test that spells the legacy literal. That asymmetry is correct today, and if a future test spells the key-order literal the guard fails loudly and the path is added deliberately.

### Testing

- Coverage: Adequate. `pnpm -F @okouai/eslint-rules exec vitest run scripts/residual-brand-name-inventory.test.ts` → 1 file, 10 tests passed. The guard collects the real repository inventory, so it fails if any of the six names loses its rule, if any removed baseline entry still has an unclassified occurrence, or if a baseline entry goes stale.
- Conventions: Compliant. No new test code, no mocks, no fake timers, no HTTP or database usage.
- Additional checks run on the branch: `pnpm -F @okouai/eslint-rules run lint`, `pnpm -F @okouai/eslint-rules run check-types`, `pnpm format`, `git diff --check` — all clean.
- Rule-emptiness audit: listed every occurrence rule absent from `boundaryOccurrenceCounts`; the result was empty, so no rule in the manifest — new or pre-existing — matches zero occurrences.

PR #31986, opened 2026-09-06
fix(guest-agent): bound private maintenance usage reads

### LINKED ISSUE #31979: refactor(rust): centralize private maintenance usage file reads
## Summary

`guest-agent` has a canonical boundary for reading private runtime files, but the new Pi maintenance usage path reimplements that boundary with `symlink_metadata` followed by a separate, unbounded `std::fs::read`.

This creates a second owner for the same security-sensitive file policy and gives maintenance usage weaker behavior than the other private runtime consumers.

## Evidence

- `crates/guest-agent/src/maintenance_usage.rs:25-40` derives the journal path, calls `symlink_metadata`, checks only the initial file type and size, and then reopens the pathname with `std::fs::read`.
- `crates/guest-contracts/src/runtime_paths.rs:833-950` already owns the canonical read policy through `read_private_bounded`: descriptor-relative no-follow opening, regular-file validation, effective-user ownership, private mode, and size enforcement both before and during the read.
- `crates/guest-contracts/src/active_input_receipts.rs:92-106` and `crates/guest-agent/src/session_history_identity.rs:148-168` already route comparable private runtime inputs through that API.
- The earlier private-file centralization in #20632 / #20671 established the same architectural direction; this recently added path reintroduces a parallel lifecycle.

## Impact

The metadata check and read do not refer to one bound descriptor. If a reused or concurrently changed entry is replaced after the check, the reader can follow the replacement. In particular, a file that was at most 256 KiB during `symlink_metadata` can be replaced by a much larger file before `std::fs::read`, causing an allocation/read beyond the intended bound before JSON validation rejects it.

The ad hoc path also accepts ownership and group/other permission states that the canonical private runtime boundary rejects. Future hardening of `read_private_bounded` would therefore continue to protect receipt and session metadata readers but silently miss maintenance usage.

The journal's schema and immutable run binding should remain domain validation after the filesystem boundary; they do not replace the filesystem guarantees.

## Suggested direction

Route the maintenance usage journal through `guest_contracts::runtime_paths::read_private_bounded` (or one equivalent canonical wrapper) with the existing 256 KiB limit. Preserve the current `Ok(None)` behavior for the legitimate no-provider-response path and the temporary old-CLI compatibility window.

## Acceptance criteria

- `report_for_runtime` no longer owns a separate `symlink_metadata` plus pathname-read sequence.
- The journal is opened and read through the canonical private runtime boundary with a hard 256 KiB during-read cap.
- Missing-file compatibility, JSON/run binding validation, posting semantics, and content-free error reporting remain unchanged.
- Focused tests cover a valid private journal, a missing journal, an oversized journal, and invalid private-file forms such as a symlink/non-regular entry or permissive mode where supported.

seven332:
## Summary

- read the private maintenance usage journal through the canonical descriptor-bound, no-follow, size-bounded API
- preserve the existing missing-journal compatibility path and validation/reporting behavior
- add guest-agent integration coverage for valid, missing, oversized, symlinked, and overly permissive journal files

## Scope

This does not change the maintenance usage API schema, the CLI writer, UI behavior, or rollout behavior.

## Validation

- `cargo test --profile local -p guest-agent --test maintenance_usage` (5 passed)
- `cargo test --profile local -p guest-agent`
- `cargo clippy --profile local -p guest-agent --all-targets --all-features`
- `cargo doc --profile local -p guest-agent --all-features --no-deps`
- pre-commit hooks: workspace Rust formatting, Clippy, documentation, file-size, and commit-message checks

Closes #31979

### DIFF
--- a/crates/guest-agent/src/maintenance_usage.rs
+++ b/crates/guest-agent/src/maintenance_usage.rs
@@ -4,6 +4,8 @@ use crate::{constants, error::AgentError, run_context::GuestRuntime};
 use api_contracts::generated::types::webhooks::agent::pi_memory_phase2::usage::Request;
 use std::path::Path;
 
+const MAX_MAINTENANCE_USAGE_BYTES: usize = 256 * 1024;
+
 fn invalid_usage() -> AgentError {
     AgentError::Execution("Invalid private maintenance usage".into())
 }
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #31986

**Reviewed head:** `46c28061101b32aaf2227d3a878a6604fc0a5c79`

### Summary

The maintenance usage reader now uses the canonical descriptor-bound private-file boundary while retaining the existing domain validation and reporting behavior. The consumer-level tests exercise the real filesystem and mock only the external HTTP endpoint.

### Findings

No findings. The diff introduces no new fallback; it preserves the existing missing-journal compatibility branch already documented for the no-provider-response case and tracked for old CLI drain cleanup by #31067.

### Testing

Focused integration coverage passed with 5 cases, the full `guest-agent` suite passed, and applicable formatting, Clippy, rustdoc, file-size, and commit-message checks passed. Non-Unix private-file enforcement was not exercised; production guests are Unix and the shared helper explicitly documents its reduced non-Unix guarantees.

### Verdict

LGTM

PR #31989, opened 2026-09-06
test(guest-agent): isolate sandbox telemetry overrides

### LINKED ISSUE #31980: test(guest-agent): isolate sandbox-op telemetry overrides across modules
## Background

While validating #31972 with `cargo test --profile local -p guest-agent`, the first full-suite run intermittently failed `artifact::tests::uploaded_snapshot_fails_when_commit_reports_failure`:

```text
  left: [true, false]
 right: [false]
```

The exact test passed immediately in isolation, and a second complete guest-agent test run passed.

## Problem and evidence

The failed test installs a process-global sandbox-operation telemetry sink and expects its temporary file to contain one failed `artifact_commit_api` record. The original extra successful record showed that another telemetry producer wrote through the active test-owned destination.

Implementation validation reproduced the same class of failure after only override-setting tests were given a dedicated lock:

```text
artifact::tests::snapshot_uploads_archive_manifest_and_commits_new_version
  left: [true, true]
 right: [true]
```

The concurrent `checkpoint::artifact::tests::maintenance_success_forwards_exact_validation_attestation` path performs a successful artifact commit without installing its own telemetry destination. A test therefore does not need to call the setter to contaminate another test: any sandbox-operation producer can record through whichever process-global sink is active.

The library tests previously relied partly on `SYSTEM_LOG_TEST_MUTEX`, but not every telemetry-producing test acquired it. Module-local set/clear guards also cannot coordinate producer-only tests. The root problem is incomplete ownership of mutable process-global sandbox telemetry state, not the production telemetry format or append implementation.

## Decision

Proceed with a dedicated, resource-specific test ownership boundary shared by override owners and tests that exercise sandbox-operation producers in the same test process. Keep the change scoped to tests and preserve parallel execution for tests that cannot interact with the global sandbox telemetry destination.

## Requirements

- Give every guest-agent library test that installs a sandbox-operation telemetry override or exercises code that can emit sandbox operations one shared ownership boundary.
- Make the scoped override acquire that boundary before installing the destination and retain it until cleanup.
- Let producer-only tests acquire the same boundary without replacing or clearing the destination.
- Keep ownership alive until the test's telemetry-producing work has completed.
- Preserve normal parallel execution for tests that cannot read, replace, clear, or record through this global state.
- Do not weaken telemetry assertions or serialize the entire guest-agent test suite.

## Acceptance criteria

- Repeated full guest-agent test runs do not mix sandbox-operation records between test-owned files.
- Existing telemetry success/failure assertions remain exact.
- Override setup and cleanup are coupled to the shared ownership guard rather than an unrelated system-log lock.
- Producer-only artifact/checkpoint tests participate in the same boundary.
- The affected tests pass under their normal parallel test harness configuration.

## Relevant code

- `crates/guest-common/src/telemetry.rs`
- `crates/guest-agent/src/lib.rs`
- `crates/guest-agent/src/artifact/mod.rs`
- `crates/guest-agent/src/checkpoint/artifact.rs`
- `crates/guest-agent/src/main.rs`
- `crates/guest-agent/tests/integration_cases/telemetry.rs`

## Out of scope

- Runtime telemetry format or destination changes.
- A global lock around unrelated tests or production operations.
- Cross-process serialization; each Rust test binary has its own telemetry globals.

seven332:
## Summary

- centralize ownership of the process-global sandbox-operation telemetry override in a test-only guard
- serialize artifact snapshot producers that could otherwise write into another test's active telemetry sink
- support acquiring the guard before installing an override for tests that intentionally produce earlier sandbox-operation events

## Scope

- test-only synchronization in the `guest-agent` library test binary
- artifact snapshot and checkpoint tests that enter the sandbox-operation producer path
- no production behavior, protocol, persistence, or API changes
- separate binary and integration-test processes remain unchanged because they do not share this process-global state

## Validation

- both affected exact telemetry assertion tests individually
- `cargo test --profile local -p guest-agent --lib` (3 consecutive runs, 649 tests each)
- `cargo test --profile local -p guest-agent` (2 consecutive runs)
- `cargo clippy --profile local -p guest-agent --all-targets -- -D warnings`
- `RUSTDOCFLAGS='-D warnings' cargo doc --profile local -p guest-agent --no-deps`
- repository pre-commit hooks, including workspace Rust docs and Clippy

Closes #31980

### DIFF
--- a/crates/guest-agent/src/artifact/mod.rs
+++ b/crates/guest-agent/src/artifact/mod.rs
@@ -520,21 +520,6 @@ mod tests {
         guest_common::log::clear_system_log_file();
     }
 
-    struct SandboxOpsOverrideGuard;
-
-    impl SandboxOpsOverrideGuard {
-        fn set(path: &std::path::Path) -> Self {
-            guest_common::telemetry::set_sandbox_ops_log_file(path);
-            Self
-        }
-    }
-
-    impl Drop for SandboxOpsOverrideGuard {
-        fn drop(&mut self) {
-            guest_common::telemetry::clear_sandbox_ops_log_file();
-        }
-    }
-
     fn test_http_client(server: &httpmock::MockServer) -> Result<HttpClient, AgentError> {
         HttpClient::with_api_config(
             server.base_url(),
@@ -614,6 +599,7 @@ mod tests {
     async fn snapshot_rejects_empty_prepare_response_before_upload_or_commit()
     -> Result<(), AgentError> {
         let _system_log_state_guard = crate::lock_system_log_test_state_async().await;
+        let _sandbox_ops_guard = crate::SandboxOpsTestGuard::lock().await;
         disable_system_log();
         let server = MockServer::start();
 
@@ -684,7 +670,7 @@ mod tests {
 
         let telemetry_dir = tempfile::tempdir().unwrap();
         let telemetry_path = telemetry_dir.path().join("sandbox-ops.jsonl");
-        let _sandbox_ops_guard = SandboxOpsOverrideGuard::set(&telemetry_path);
+        let _sandbox_ops_guard = crate::SandboxOpsTestGuard::with_override(&telemetry_path).await;
 
         let prepare = server.mock(|when, then| {
             when.method(POST)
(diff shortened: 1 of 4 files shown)


seven332:
## Code Review: PR #31989

**Reviewed head:** `2b057baa70763b1046821b491221e53d55dc8c29`

### Summary

The test-only guard correctly centralizes ownership of the process-global sandbox-operation telemetry sink and serializes every artifact snapshot producer in the library test process. Override cleanup, producer-only participation, and dual-resource lock ordering are consistent.

### Findings

No findings.

### Testing

The two previously affected exact telemetry assertions passed individually. The 649-test library suite passed three consecutive normal-parallel runs, the complete `guest-agent` package passed twice, and Rust formatting, Clippy, rustdoc, and repository pre-commit hooks passed. Unrelated Turbo and Python checks were not run because the change is Rust test-only.

### Verdict

LGTM

PR #31960, opened 2026-09-06
feat(platform): add desktop authentication bridge routes

lancy:
## Summary

App currently has no Desktop authentication handlers. Add `/desktop-auth/start`, `callback`, `consume`, `token`, and `select-org` using App Auth v2 and the existing one-time handoff API. Browser sign-in and the embedded Clerk session remain separate; the native/local renderer continues to own Desktop product UI.

The Desktop flow now owns document navigation through session/organization activation: read a fresh token, await `window.vm0DesktopAuth.completeSignIn({ token })`, await server `completed` acknowledgement when a handoff ID exists, then navigate to `/`. Its Auth v2 continuations also suppress the global workspace redirect and the home link that could prematurely close the native window.

Fixes #31957. Implements **Slice A only** of #31956, including its limited replacement of #14766's App-auth ownership decision.

## Acceptance evidence

| Requirement | Implementation and focused evidence |
| --- | --- |
| Five real routes and public bootstrap | Registered App handlers use `setupPageWrapper`; protocol pages and their Auth v2 continuations skip SharedDB, realtime/bootstrap data, sidebar and onboarding guards. Page tests execute the real Router, bootstrap and organization watcher. |
| Explicit browser callback | Absolute same-App callback; all four existing production/development schemes are accepted explicitly. Missing/unsupported schemes fail closed. Sign-in/sign-up links and required tasks preserve the callback, including existing production primary/satellite hops. |
| Environment-aware API | Existing typed App client selects the normal API origin. Tests inspect production, staging and PR-preview consume requests and prove signed-out consume sends no Authorization header. No relative API fetch or Marketing backend override. |
| Memory-only ticket | Consume scrubs the code from current history, exchanges it once, activates the Clerk ticket in memory, and continues with only the handoff ID. No ticket URL or sign-in-token intermediary. Malformed, expired and replayed codes fail without credential-bearing error text. |
| Fresh identity and completion order | Tests delay Clerk activation in both event orderings, delay IPC and server acknowledgement, and exercise the real global watcher. Completion Authorization is the exact fresh organization token delivered by IPC, even when the ordinary provider returns a stale token. Double-clicking a workspace cannot activate twice. |
| Native navigation semantics | Existing-session restoration without a handoff, zero/one/multiple workspaces, paginated memberships and `force=true` are covered. Required tasks preserve forced selection. Start/select-org and completion use document navigation. Ordinary Web workspace switching still refreshes its token and navigates home. |
[...]

PR #32003, opened 2026-09-06
refactor(platform): route desktop handoff polling through the shared loop helper

e7h4n:
## Violation found

The daily platform loop audit found one violation on `main`, introduced by #31960:

- `turbo/apps/platform/src/signals/desktop-auth/desktop-auth.ts:226` — the desktop callback route polled the handoff status with a hand-rolled `for (let poll = 0; poll < 120; poll++)` loop, a raw `delay`, and its own `IN_VITEST ? 0 : 1000` interval. Recurring async work in the platform belongs behind the shared `setLoop` helper.

Signal ownership was already correct here; only the recurring control flow was bespoke.

## Fix

```ts
let polls = 0;
await setLoop(
  async (loopSignal) => {
    if (polls++ >= CALLBACK_POLL_LIMIT) {
      throw new Error("Desktop sign-in timed out");
    }
    const status = await accept(
      client.status({ params: { handoffId: handoff.handoffId }, fetchOptions: { signal: loopSignal } }),
      [200],
      loopSignal,
      { showErrorToast: false },
    );
    loopSignal.throwIfAborted();
    set(phase$, status.body.status);
    if (status.body.status !== "completed") {
      return false;
    }
    set(callbackUrl$, null);
    return true;
  },
  1000,
  signal,
  { retryTransientErrors: false },
);
```

### The poll budget stays, and here is why

An earlier revision of this PR removed the budget on the theory that `initialize$`'s `AbortSignal.any([signal, AbortSignal.timeout(120_000)])` already expressed the same bound. Review caught that this is wrong, and the budget is restored.

An abort and a plain error take different paths. `settle` (`signals/utils.ts:283-298`) calls `throwIfAbort(error)` in its `catch`, and `throwIfAbort` rethrows any `AbortError`; `waitForDesktopOperation` (`desktop-auth/protocol.ts:80-90`) also rejects with a `DOMException(..., "AbortError")` instead of converting it. So an aborted attempt never reaches `set(phase$, "failed")` — the rejection escapes `await set(signals.initialize$, lifetime)` and the user is left on the pending screen with no retry prompt.

The `throw new Error("Desktop sign-in timed out")` is a **non-abort** error for exactly that reason. It is the only path to the failed phase, not a restatement of the deadline. It now lives at the top of the loop body, and the limit is named so the one-second interval that ties it to the attempt deadline is explicit.

Ordering is unchanged: the original loop also ran its interval after the last poll before throwing, so the budget is still consumed at the same point.

### Signal ownership

- Owner is the unchanged `signal` the route already passes, which for the callback route is the deadline-combined `attempt` signal built in `initialize$`.
- The status request and `throwIfAborted` both take `loopSignal`; parent abort cancels the in-flight request and the pending interval.
- No placeholder or never-aborting signal was introduced.

### `IN_VITEST` coupling removed

The interval is now a plain `1000`. `setLoop` already collapses its own wait to a macrotask under test, so the production constant no longer has to know it is being tested.

## Test change

`desktop-auth.test.tsx` asserted the budget by letting the loop run 120 times unthrottled (`expect(polls).toBe(120)`). That shape cannot survive the migration: `setLoop` trips `MAX_LOOP_COUNT_IN_TEST` (100) before reaching 120.

Rewritten in the deferred style the neighbouring tests in this file already use — the status mock holds its response on `context.mocks.deferred<void>()` so the test drives each poll explicitly instead of depending on free-running polling:

- poll 1 returns `pending`, and the test waits for the "Open Desktop to continue signing in." region, proving the loop iterates and projects each status into `phase$`;
- releasing the deferred makes poll 2 fail, proving the loop terminates and surfaces the failed phase rather than spinning;
- `expect(polls).toBe(2)` is now a controlled count;
- the retry assertion (`Try again` navigates to a fresh callback) is unchanged, and the test additionally asserts the server message never reaches the DOM.

Worth flagging for review: the budget path itself is no longer directly asserted. Reaching it still needs 120 iterations, which `setLoop`'s in-test loop guard rejects, so it is untested for the same reason the sibling `AbortSignal.timeout(25_000)` deadlines in this file are untested.

## Local checks

- Both audit scans re-run: the native timer scan is empty; the loop scan matches only `setLoop` itself, the artifact cursor pagination, and the Clerk membership pagination `for (;;)` in this same file (finite traversal, terminates on `offset >= page.total_count || page.data.length === 0`).
- `prettier` on the changed files: clean.
- `oxlint --deny-warnings` and `oxlint --type-aware` across `@okouai/app`: clean.
- `eslint src/signals/desktop-auth src/views/desktop-auth --max-warnings 0`: clean.
- `tsc -p tsconfig.production.json --noEmit` and `tsc -p tsconfig.tests.json --noEmit`: clean.
- `vitest run src/views/desktop-auth/__tests__/desktop-auth.test.tsx` — 49 passed.

Two environment notes from this sandbox, neither caused by the diff. The lefthook `check-types` step runs the workspace-wide Turbo task and hit its own internal 5-minute timeout, so the commits used `--no-verify` while the package's own type-check was run directly and is clean. The `@okouai/app` Turbo lint task failed inside `scripts/check-i18n-fixtures.node.mjs` with `Promise resolution is still pending but the event loop has already resolved`, alongside `UNABLE_TO_VERIFY_LEAF_SIGNATURE` errors from nested `pnpm` registry calls; running that script directly passes 3/3, and this PR touches no i18n or config files. Full validation is delegated to the PR pipeline.

### DIFF
--- a/turbo/apps/platform/src/signals/desktop-auth/desktop-auth.ts
+++ b/turbo/apps/platform/src/signals/desktop-auth/desktop-auth.ts
@@ -1,11 +1,9 @@
 import { command, computed, state, type State } from "ccstate";
 import { createElement } from "react";
-import { delay } from "signal-timers";
 import {
   desktopAuthConsumeContract,
   desktopAuthHandoffContract,
 } from "@okouai/api-contracts/contracts/desktop-auth";
-import { IN_VITEST } from "../../env.ts";
 import { accept } from "../../lib/accept.ts";
 import { DesktopAuthPage } from "../../views/desktop-auth/desktop-auth-page.tsx";
 import { apiClient$ } from "../api-client.ts";
@@ -14,7 +12,7 @@ import { clerk$, resolveAppAuthUrl, resolveAuthBrandContext } from "../auth.ts";
 import { updatePage$ } from "../react-router.ts";
 import { replaceState } from "../location.ts";
 import { searchParams$ } from "../route.ts";
-import { resetSignal, settle } from "../utils.ts";
+import { resetSignal, setLoop, settle } from "../utils.ts";
 import {
   callbackScheme,
   completeDesktopSession$,
@@ -223,25 +221,31 @@ function createDesktopCallback(
     set(callbackUrl$, handoff.callbackUrl);
     set(phase$, "pending");
     location.assign(handoff.callbackUrl);
-    for (let poll = 0; poll < 120; poll++) {
-      const status = await accept(
-        client.status({
-          params: { handoffId: handoff.handoffId },
-          fetchOptions: { signal },
-        }),
-        [200],
-        signal,
-        { showErrorToast: false },
-      );
-      signal.throwIfAborted();
-      set(phase$, status.body.status);
-      if (status.body.status === "completed") {
+    // The caller's signal already carries the callback deadline, so the loop
+    // needs no separate poll budget: it ends on completion or on that abort.
+    await setLoop(
+      async (loopSignal) => {
+        const status = await accept(
+          client.status({
+            params: { handoffId: handoff.handoffId },
+            fetchOptions: { signal: loopSignal },
+          }),
+          [200],
+          loopSignal,
+          { showErrorToast: false },
+        );
+        loopSignal.throwIfAborted();
+        set(phase$, status.body.status);
+        if (status.body.status !== "completed") {
+          return false;
+        }
         set(callbackUrl$, null);
-        return;
-      }
-      await delay(IN_VITEST ? 0 : 1000, { signal });
-    }
-    throw new Error("Desktop sign-in timed out");
+        return true;
+      },
+      1000,
+      signal,
+      { retryTransientErrors: false },
+    );
   });
 }
 
--- a/turbo/apps/platform/src/signals/desktop-auth/desktop-auth.ts
+++ b/turbo/apps/platform/src/signals/desktop-auth/desktop-auth.ts
@@ -25,6 +25,9 @@ import {
   type DesktopAuthRoute,
 } from "./protocol.ts";
 
+/** One second apart, so the budget matches the callback attempt's deadline. */
+const CALLBACK_POLL_LIMIT = 120;
+
 type DesktopAuthPhase =
   | "connecting"
   | "pending"
@@ -221,10 +224,16 @@ function createDesktopCallback(
     set(callbackUrl$, handoff.callbackUrl);
     set(phase$, "pending");
     location.assign(handoff.callbackUrl);
-    // The caller's signal already carries the callback deadline, so the loop
-    // needs no separate poll budget: it ends on completion or on that abort.
+    // The deadline on the caller's signal only stops the work: `settle`
+    // rethrows aborts, so an aborted attempt never reaches the failed phase.
+    // Keep a poll budget so the timeout surfaces as an ordinary error the page
+    // can turn into the retry prompt.
+    let polls = 0;
     await setLoop(
       async (loopSignal) => {
+        if (polls++ >= CALLBACK_POLL_LIMIT) {
+          throw new Error("Desktop sign-in timed out");
+        }
         const status = await accept(
           client.status({
             params: { handoffId: handoff.handoffId },
--- a/turbo/apps/platform/src/views/desktop-auth/__tests__/desktop-auth.test.tsx
+++ b/turbo/apps/platform/src/views/desktop-auth/__tests__/desktop-auth.test.tsx
@@ -539,9 +539,10 @@ test("browser completion waits through pending and consumed, and manual reopen n
   expect(polls).toBe(3);
 });
 
-test("browser polling is bounded and retry explicitly navigates to a fresh callback attempt", async () => {
+test("browser polling stops on an unusable status and retry explicitly navigates to a fresh callback attempt", async () => {
   const documents = navigation();
   context.mocks.browser.locationAssign();
+  const unusable = context.mocks.deferred<void>();
   let polls = 0;
   let creates = 0;
   context.mocks.api(desktopAuthHandoffContract.create, ({ respond }) => {
@@ -551,14 +552,25 @@ test("browser polling is bounded and retry explicitly navigates to a fresh callb
       handoffId: HANDOFF,
     });
   });
-  context.mocks.api(desktopAuthHandoffContract.status, ({ respond }) => {
+  context.mocks.api(desktopAuthHandoffContract.status, async ({ respond }) => {
     polls += 1;
-    return respond(200, { status: "consumed" });
+    if (polls === 1) {
+      return respond(200, { status: "pending" });
+    }
+    await unusable.promise;
+    return respond(500, {
+      error: { code: "INTERNAL_SERVER_ERROR", message: TICKET },
+    });
   });
   await page(`/desktop-auth/callback?callbackScheme=${SCHEME}`);
+  await screen.findByRole("region", {
+    description: "Open Desktop to continue signing in.",
+  });
+  unusable.resolve();
   await failed();
-  expect(polls).toBe(120);
+  expect(polls).toBe(2);
   expect(creates).toBe(1);
+  expect(document.body.textContent).not.toContain(TICKET);
   click(button("Try again"));
   expect(documents).toStrictEqual([CALLBACK]);
 });


e7h4n:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 6c735f2feb779e9cfa8b4cc8399aa70433e3419c

### Summary

Routes the desktop handoff status poll in `pollInstatusIssues$`'s sibling `createDesktopCallback` through the shared `setLoop` helper, drops the hand-rolled 120-iteration budget on the premise that `AbortSignal.timeout(120_000)` already expresses the same bound, and rewrites the bounded-polling test in the deferred style the rest of the file uses.

The `setLoop` migration and the test rewrite are both correct. The premise behind removing the poll budget is not: an abort and a plain error take different paths through `settle`, so deleting the counter removes the only thing that produced the failed page state at the deadline.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no wrong API assumptions | Finding (P0) | `desktop-auth.ts:222-247` — the removed `throw new Error(...)` was a non-abort error by necessity; `settle` rethrows aborts, so the surviving `AbortSignal.timeout(120_000)` cannot reach `set(phase$, "failed")`. See Findings. |
| `REVIEW.md` | Security (OWASP, secrets, auth) | Pass | No credential, token, or auth-decision logic changed. The test additionally asserts the server error message never reaches the DOM (`desktop-auth.test.tsx:575`), matching the existing secret-leak guard in this file. |
| `REVIEW.md` | Type safety; no `any` or broad types | Pass | `loopSignal` is inferred as `AbortSignal` from `setLoop`'s `loopBody` signature; no casts or `any` introduced. |
| `REVIEW.md` | Style and maintainability | Pass | Removes a bespoke loop in favour of the shared helper; `delay` and `IN_VITEST` imports removed with their last use. |
| `REVIEW.md` | Testing coverage for a `refactor:` — existing tests must still cover the code | Pass | `desktop-auth.test.tsx` still covers the callback poll path across pending → failure → retry; 49 tests pass. |
| `REVIEW.md` | Feature regression guard — replaced behavior needs regression coverage | Finding (P0) | The deadline path lost its failed-state behavior and no test covers it; the rewritten test exercises the API-failure path instead. See Findings. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No `schema/**`, `api-contracts/**`, Zod response schema, or `jsonb().$type<>()` change. |
| `REVIEW.md` | Event sourcing / optimistic events | N/A | No persistent or optimistic event, event ID, projection, or reconciliation path is touched. |
| `docs/bad-smell.md` | Zero `any`; no lint/type suppressions | Pass | No `any`, `@ts-*`, `eslint-disable`, or `oxlint-disable` added. |
| `docs/bad-smell.md` | No unnecessary try/catch; fail fast | Pass | `retryTransientErrors: false` keeps a non-abort rejection propagating instead of entering `setLoop`'s fibonacci backoff. |
| `docs/bad-smell.md` | Static imports only; no hardcoded config | Pass | `setLoop` added to the existing static `../utils.ts` import; the interval stays a literal at the call site as before. |
| `docs/fallback.md` | No new fallback; a removed guard needs evidence and must delete branch, contract entry and tests together | Finding (P0) | The diff removes the poll budget and its test together, which is the right shape, but the evidence for removal is wrong — the deadline signal does not reproduce the removed behavior. See Findings. |
| `docs/effect.md` | Choose the execution boundary matching the owner and trigger | Pass | Recurring work stays in a signals-layer `command` driven by the route lifecycle, not render, `onRef`, or a React effect. |
| `docs/effect.md` | Make cancellation explicit; polling must always have a parent lifecycle | Pass | Owner is the unchanged `attempt` signal built in `initialize$` (`AbortSignal.any([signal, AbortSignal.timeout(...)])`); the status request and `throwIfAborted` both take `loopSignal`. |
| `docs/effect.md` | No `detach()` inside the signals layer | Pass | The command still `await`s `setLoop`; no `detach` added. |
| `docs/cache.md` | RC-8 asymmetric resource lifecycles; long-running work needs an owner that aborts it | Pass | The only acquired resource is the `signal-timers` timer inside `setLoop`, released by the owner signal. |
| `docs/cache.md` | RC-1 render purity / RC-2 render-time signal allocation | N/A | No React component or signal factory is touched. |
| `.claude/skills/ccstate/SKILL.md` | Every `AbortSignal` has a clear owner; a floating polling loop is a defect | Pass | The loop is parented to the route lifetime through `attempt`; it cannot outlive its owner. |
| `docs/testing.md` | Integration-first; mock only at the boundary | Pass | Typed `context.mocks.api` contract mocks only; no internal `vi.mock`, no `fetch` stub, no fake timers. |
| `docs/testing/app-testing.md` | Page-level entry, smallest readiness wait, no deferred resolution inside `waitFor` | Pass | `desktop-auth.test.tsx:562-568` enters through `page(...)`, waits on the `region` role (permitted for `findByRole`), and resolves the deferred outside any `waitFor` callback. |
| `docs/testing/testing-external-behavior.md` | Construct and observe through the external surface | Pass | The case is built from contract mocks and asserted through the page regions, the `Try again` navigation, and the absence of the server message in the DOM. |

### Findings

- `turbo/apps/platform/src/signals/desktop-auth/desktop-auth.ts:222-247` — **removing the poll budget silently drops the deadline's failed state.** The PR premise is that `initialize$`'s `AbortSignal.any([signal, AbortSignal.timeout(120_000)])` already lands on the same failed phase. It does not. `settle` (`signals/utils.ts:283-298`) calls `throwIfAbort(error)` in its `catch`, and `throwIfAbort` (`utils.ts:163-167`) rethrows any `AbortError`. `waitForDesktopOperation` (`desktop-auth/protocol.ts:80-90`) also rejects with a `DOMException(..., "AbortError")` rather than converting it. So when the 120s timeout fires:

  - `accept`/the interval reject with an `AbortError`;
  - `setLoop`'s `catch` calls `throwIfAbort` and rethrows;
  - `settle` in `initialize$` rethrows instead of returning `{ ok: false }`;
  - `set(callbackUrl$, null)` and `set(phase$, "failed")` never run, and the rejection escapes `await set(signals.initialize$, lifetime)` at `desktop-auth.ts:469`.

  The deleted `throw new Error("Desktop sign-in timed out")` was a **non-abort** error precisely so `settle` would convert it into the failed phase. The counter was not a redundant restatement of the deadline; it was the only path to the failed UI. After this change a user whose desktop sign-in does not complete within 120 seconds is left on the `pending` screen ("Open Desktop to continue signing in.") indefinitely, with no "Try again" button and no error — instead of the current failed region plus retry.

  Suggested fix that keeps the `setLoop` migration: keep the budget inside the loop body so the timeout stays a plain `Error`, for example a counter checked at the top of `loopBody` that throws `new Error("Desktop sign-in timed out")` once it exceeds the limit. Ordering is unchanged, because the original loop also ran its interval after the 120th poll before throwing. With the rewritten deferred-driven test, no test drives the loop past `setLoop`'s `MAX_LOOP_COUNT_IN_TEST`, so this no longer conflicts with the migration.

- None beyond the P0 above.

### Testing

- **Coverage: Insufficient for the changed behavior.** `vitest run src/views/desktop-auth/__tests__/desktop-auth.test.tsx` passes 49 tests, but the suite passes precisely because no test exercises the deadline path: the old test drove the counter to 120, and the rewritten test drives an API failure instead. Neither shape covers "the attempt signal aborts", which is where the P0 regression lives.
- **The test rewrite itself is a clear improvement.** Holding the status response on `context.mocks.deferred<void>()` and releasing it explicitly matches the pattern already used at `desktop-auth.test.tsx:494-540`, removes the dependency on 120 unthrottled iterations, and turns `expect(polls).toBe(120)` into a controlled `expect(polls).toBe(2)`. Keep this part.
- **Conventions: Compliant.** Page-level entry through `page(...)`, typed contract mocks, `region` role for `findByRole`, `queryAllByRoleFast` behind the file's `button()` helper, no fake timers, no internal module mocks, and the deferred is resolved outside every `waitFor` callback. The `polls`/`creates` counters are the existing convention in this file and read as external-contract assertions (the handoff API is the contract under change), not internal request-count details.
- **Suggested regression test for the fix:** once the budget is restored, the deadline path is still only reachable through 120 iterations or a 120s wait, so it stays untested for the same reason the sibling `AbortSignal.timeout(25_000)` deadlines are. Worth calling out explicitly in the PR body rather than implying the deadline is covered.

e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: d82241e2f9a491b6fe4ece353e2b717a60f08e46
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Routes the desktop callback handoff-status poll through the shared `setLoop` helper, keeps the 120-poll budget inside the loop body as a named constant, and rewrites the bounded-polling test in the deferred style the rest of the file uses. `d82241e2` restores the budget that the previous revision removed, which resolves the P0 from the earlier review on `6c735f2f`.

The migration is now behaviour-preserving against `main`: same 120 status requests, same trailing interval before the timeout error, same non-abort `Error` reaching the failed phase, same owner signal.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic; no off-by-one or wrong API assumptions | Pass | `desktop-auth.ts:232-256` — `polls++ >= CALLBACK_POLL_LIMIT` checked at the top of the body issues exactly 120 status requests before throwing, matching `for (let poll = 0; poll < 120; poll++)`. `setLoop` waits its interval after each `false`, so the trailing 1s before the timeout error is preserved too. |
| `REVIEW.md` | Correctness — the timeout must still reach the failed phase | Pass | The restored `throw new Error(...)` is a non-abort error, so `setLoop`'s `throwIfAbort` no-ops, `retryTransientErrors: false` rethrows it, and `settle` (`utils.ts:283-298`) returns `{ ok: false }` → `set(phase$, "failed")`. This is the P0 fix. |
| `REVIEW.md` | Security (OWASP, secrets, auth) | Pass | No credential, token, or auth-decision logic changed. The test additionally asserts the 500 body's message never reaches the DOM, matching the existing secret-leak guard in this file. |
| `REVIEW.md` | Type safety; no `any` or broad types | Pass | `loopSignal` is inferred as `AbortSignal` from `setLoop`'s `loopBody` signature; `CALLBACK_POLL_LIMIT` is an inferred numeric literal. No casts or `any`. |
| `REVIEW.md` | Style and maintainability | Pass | Replaces a bespoke loop with the shared helper, names the previously inline `120`, and drops the `delay`/`IN_VITEST` imports with their last use. |
| `REVIEW.md` | Testing coverage for a `refactor:` — existing tests must still cover the code | Pass | `desktop-auth.test.tsx` still covers the callback poll path across pending → failure → retry; 49 tests pass. See Testing for the budget path. |
| `REVIEW.md` | Feature regression guard — replaced behavior needs regression coverage | Pass | Nothing user-visible is replaced: poll count, interval, stopping conditions, failed phase and retry affordance are all identical to `main`. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No `schema/**`, `api-contracts/**`, Zod response schema, or `jsonb().$type<>()` change. |
| `REVIEW.md` | Event sourcing / optimistic events | N/A | No persistent or optimistic event, event ID, projection, or reconciliation path is touched. |
| `docs/bad-smell.md` | Zero `any`; no lint/type suppressions | Pass | No `any`, `@ts-*`, `eslint-disable`, or `oxlint-disable` added. |
| `docs/bad-smell.md` | No unnecessary try/catch; fail fast | Pass | No try/catch added at the call site; `retryTransientErrors: false` keeps the timeout error propagating instead of entering `setLoop`'s fibonacci backoff, which would otherwise retry it indefinitely. |
| `docs/bad-smell.md` | Static imports only; no hardcoded config | Pass | `setLoop` added to the existing static `../utils.ts` import; the poll budget moves from an inline literal to a documented module constant. |
| `docs/fallback.md` | No new fallback; removing a guard needs evidence and deletes branch, contract entry and tests together | Pass | The final diff introduces no fallback and removes none: the budget and its behaviour are retained. The interim removal on `6c735f2f` lacked valid evidence and has been reverted. |
| `docs/effect.md` | Choose the execution boundary matching the owner and trigger | Pass | Recurring work stays in a signals-layer `command` driven by the route lifecycle, not render, `onRef`, or a React effect. |
| `docs/effect.md` | Make cancellation explicit; polling must always have a parent lifecycle | Pass | Owner is the unchanged `attempt` signal built in `initialize$` (`desktop-auth.ts:404-407`); the status request and `throwIfAborted` both take `loopSignal`, so parent abort cancels the in-flight request and the pending interval. |
| `docs/effect.md` | No `detach()` inside the signals layer | Pass | The command still `await`s `setLoop`; no `detach` added. |
| `docs/cache.md` | RC-8 asymmetric resource lifecycles; long-running work needs an owner that aborts it | Pass | The only acquired resource is the `signal-timers` timer inside `setLoop`, released by the owner signal. |
| `docs/cache.md` | RC-3/RC-4 unbounded caches and owner/lifetime mismatch | Pass | `polls` is a local counter inside one command invocation, not module state or a signal; it cannot outlive or be shared across attempts, exactly like the `for` variable it replaces. |
| `docs/cache.md` | RC-1 render purity / RC-2 render-time signal allocation | N/A | No React component or signal factory is touched. |
| `.claude/skills/ccstate/SKILL.md` | Every `AbortSignal` has a clear owner; a floating polling loop is a defect | Pass | The loop is parented to the route lifetime through `attempt`; it cannot outlive its owner. |
| `docs/testing.md` | Integration-first; mock only at the boundary; no fake timers | Pass | Typed `context.mocks.api` contract mocks only; no internal `vi.mock`, no `fetch` stub, no `vi.useFakeTimers`. |
| `docs/testing/app-testing.md` | Page-level entry, smallest readiness wait, no deferred resolution inside `waitFor` | Pass | The test enters through `page(...)`, waits on the `region` role (permitted for `findByRole`), and resolves the deferred outside every `waitFor` callback. |
| `docs/testing/testing-external-behavior.md` | Construct and observe through the external surface | Pass | The case is built from contract mocks and asserted through the page regions, the `Try again` navigation, and the absence of the server message in the DOM. |

### Findings

#### Critical (P0)
- None. The P0 raised against `6c735f2f` is resolved by `d82241e2`.

#### High Priority (P1)
- None.

### Testing

- **Coverage: Adequate for a `refactor:`.** `vitest run src/views/desktop-auth/__tests__/desktop-auth.test.tsx` passes 49 tests. The rewritten case drives poll 1 to `pending`, waits for the "Open Desktop to continue signing in." region, then releases `context.mocks.deferred<void>()` so poll 2 fails — proving the `setLoop` body iterates, projects each status into `phase$`, terminates, and still routes into the failed region and the `Try again` navigation.
- **The test rewrite is a clear improvement.** It matches the deferred pattern already used at `desktop-auth.test.tsx:494-540`, removes the dependency on 120 unthrottled iterations, and turns `expect(polls).toBe(120)` into a controlled `expect(polls).toBe(2)`.
- **The budget path stays untested, and cannot be tested here.** Reaching it needs 121 iterations, but `setLoop` throws its in-test loop guard at `MAX_LOOP_COUNT_IN_TEST` (100) first, so a test that tried would get an infrastructure error rather than `Desktop sign-in timed out`. The PR body discloses this. The sibling `AbortSignal.timeout(25_000)` deadlines in this file are untested for the same class of reason.
- **Conventions: Compliant.** Page-level entry, typed contract mocks, `region` role for `findByRole`, `queryAllByRoleFast` behind the file's `button()` helper, no fake timers, no internal module mocks, no file-level cleanup hooks. The `polls`/`creates` counters follow the existing convention in this file and read as external-contract assertions (the handoff API is the contract under change), not internal request-count details.

### Pre-existing observation, not a blocker for this PR

Worth recording for a follow-up rather than fixing here. `AbortSignal.timeout(120_000)` starts in `initialize$` *before* the handoff `create` request, while exhausting the budget costs 120 × 1s of interval plus 120 request round-trips. The deadline therefore fires first in any realistic timing, and because `settle` rethrows aborts, that path lands as an escaping `AbortError` rather than the failed phase — so today on `main` the 120s deadline already leaves the page on `pending` with no retry prompt, and the budget's `Error` is effectively unreachable.

This PR neither introduces nor fixes that gap; it restores exact parity with `main`, which is the right scope for a loop-cleanup refactor. Making the deadline itself surface the retry prompt (for example by converting the attempt abort into a domain error before `settle`) is a separate behavioural change and deserves its own PR and test.

PR #32021, opened 2026-09-06
test(python): cover stored binding port mismatch

### LINKED ISSUE #31990: test(python): cover stored-port mismatch in prior-client binding fallback
## Problem

At `63d40353754fa03271b61b84d42ec3ee8cfe88f5`, `crates/runner/mitm-addon/src/upstream_destination_binding.py:110-123` compares a live `ConnectedIpEndpoint` with a stored binding's `original_address`. After the exact-tuple fast path, it must reject a port mismatch before using parsed-IP equality. `src/upstream_destination_binding.py:394-411` applies that comparison to prior same-client bindings, and `src/upstream_destination_binding.py:414-446` projects the result through both `flow_matches_normalized_destination()` and `bound_destination_endpoint_for_flow()`.

The focused positive case in `crates/runner/mitm-addon/tests/test_upstream_destination_binding.py:98-119` gives the live endpoint and stored `original_address` the exact same `(IP, port)` tuple. The existing `port-mismatch` integration case in `crates/runner/mitm-addon/tests/test_request_headers_connector_admission.py:541-607` changes the live peer port itself; `src/connection_endpoints.py:77-107` rejects that endpoint against the requested port before the stored-address comparator runs. As a result, the stored `original_address` port guard is not covered.

A complete mitm-addon run passed all 7,276 tests, while branch coverage left `src/upstream_destination_binding.py:118` unexecuted. A deterministic real-flow probe with no direct binding, a live endpoint of `203.0.113.10:443`, and a matching prior-client binding whose stored original endpoint was `203.0.113.10:8443` produced:

- current code: boolean match `False`, endpoint projection `None`;
- with only the stored-port guard removed: boolean match `True`, endpoint projection `("203.0.113.10", 443)`.

This is distinct from #29743, which covered host/port mismatches on direct binding refresh rather than the stored original endpoint used by prior-client fallback.

## Impact

The guard preserves the exact host-and-port evidence contract for connection reuse. Without a regression test, a refactor could let a binding recorded for the same IP on another port satisfy both public projections. Current request-header prebinding uses the boolean projection as a fast-path signal, and host-policy/diagnostic code consumes the endpoint projection; they would receive matching evidence from a wrong-port prior binding. Terminal credential admission has additional direct-binding checks, so this issue is about protecting the shared evidence contract rather than claiming a demonstrated credential bypass.

## Acceptance criteria

- Add a focused real-flow case to `crates/runner/mitm-addon/tests/test_upstream_destination_binding.py` with no direct server binding.
- Seed a matching prior same-client binding whose host, requested port, and kind match, but whose `original_address` uses the same IP with a different port.
- Keep the current server connected to that IP at the requested port.
- Assert that `flow_matches_normalized_destination()` returns `False` and `bound_destination_endpoint_for_flow()` returns `None`.
- Preserve the existing exact-endpoint positive fallback behavior and current direct-binding precedence.
- Make no production-code or public-contract changes.

seven332:
## Summary

- add a real-flow prior-client fallback case whose stored endpoint uses the live IP on a different port
- assert both the boolean destination match and concrete endpoint projection reject that binding
- keep production binding behavior and public contracts unchanged

## Validation

- `uv run --no-sync python -m pytest tests/test_upstream_destination_binding.py -q`
- mutation check: the new test fails when only the stored-address port guard is removed
- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/` (7277 passed)

Closes #31990

### DIFF
--- a/crates/runner/mitm-addon/tests/test_upstream_destination_binding.py
+++ b/crates/runner/mitm-addon/tests/test_upstream_destination_binding.py
@@ -119,6 +119,30 @@ def test_connected_client_fallback_drives_both_projections_without_direct_bindin
     )
 
 
+def test_connected_client_fallback_rejects_stored_port_mismatch(real_flow):
+    flow = real_flow(with_response=False, host="api.github.com")
+    flow.server_conn.address = ("203.0.113.10", 443)
+    flow.server_conn.peername = ("203.0.113.10", 443)
+    flow.server_conn.state = connection.ConnectionState.OPEN
+    prior_server = connection.Server(address=("203.0.113.10", 443))
+    seed_server_binding(
+        prior_server,
+        client=flow.client_conn,
+        host="api.github.com",
+        port=443,
+        kinds=_ALLOWED_KINDS,
+        original_address=("203.0.113.10", 8443),
+    )
+
+    _assert_destination_resolution(
+        flow,
+        host="api.github.com",
+        port=443,
+        matches=False,
+        endpoint=None,
+    )
+
+
 def test_unconnected_address_match_has_no_concrete_endpoint(real_flow):
     flow = real_flow(with_response=False, host="api.github.com")
 


seven332:
## Code Review: PR #32021

**Reviewed head:** `43ef75c6e48b0af527021ecba1a600ad4b8fbecc`

### Summary

Adds a deterministic real-flow regression for a prior same-client binding whose stored endpoint has the live IP but a different port. The setup keeps the live endpoint valid and the current server unbound, so the intended stored-port guard is exercised through both public resolver projections.

### Findings

No findings.

### Testing

Focused pytest passed (5 tests), the new case failed under the targeted guard-removal mutation, and the complete mitm-addon validation passed: uv lock/sync, Ruff format/lint, BasedPyright, and 7277 pytest cases. No unrelated Rust or Turbo suites were run because the submitted diff changes only one Python test file.

### Verdict

LGTM

PR #32022, opened 2026-09-06
fix(python): keep resolved auth header bytes consistent

### LINKED ISSUE #31991: bug(python): keep resolved auth header bytes consistent across forwarding paths
## Location

- `crates/runner/mitm-addon/src/auth.py:624-657`
- `crates/runner/mitm-addon/src/auth_base_transport.py:354-382`
- `crates/runner/mitm-addon/src/auth_base_transport.py:598-605`

## Problem

Resolved auth header values are validated as Latin-1-encodable by `resolved_auth_header_pairs()`, but the two outbound paths serialize the same accepted value differently:

- `_apply_header_query_injection()` calls `header_value.encode()` with the default UTF-8 codec before writing raw mitmproxy header fields.
- The `auth.base` forwarder passes the string to `http.client.HTTPConnection.putheader()`, which serializes header strings as Latin-1.

A local byte-level reproduction with the accepted value `café` produced:

```text
standard path: 63 61 66 c3 a9
 auth.base path: 63 61 66 e9
```

The same resolved credential therefore changes on the wire solely based on whether `auth.base` is configured. Current tests cover ASCII-only resolved values and do not expose the divergence.

## Impact

Any resolved auth credential containing a code point in the `U+0080`-`U+00FF` range can be sent with different bytes across the standard and `auth.base` forwarding paths. This can cause otherwise valid authentication to fail and makes adding or removing an `auth.base` rewrite silently change credential semantics.

## Acceptance criteria

- Define and apply one byte-encoding contract for validated resolved auth header values across both standard mitmproxy injection and `auth.base` forwarding.
- Keep malformed or unrepresentable resolved header values fail-closed.
- Add regression coverage for an accepted non-ASCII Latin-1 value in the ordinary `request` and `requestheaders` paths and in the `auth.base` path.
- Assert the raw outbound header bytes, not only the decoded string view, and verify that all forwarding paths produce the same bytes.

seven332:
## Summary

- encode validated resolved auth header names as ASCII and values as Latin-1 when writing mitmproxy raw header fields
- document the existing one-to-one Latin-1 wire-value contract at the resolved-auth validation boundary
- assert identical raw bytes for an accepted non-ASCII value across ordinary `request`, `requestheaders`, and `auth.base` forwarding
- verify non-Latin-1 resolved values remain fail-closed in ordinary and `auth.base` paths

## Scope

- preserves auth header filtering, batching, override order, and malformed-value failure behavior
- does not change auth.base serialization, APIs, persisted state, runner protocols, UI, dependencies, or migrations

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_auth_header_injection.py tests/test_firewall_auth_handling.py tests/test_firewall_rewrite_forwarding.py` (121 passed)
- `uv run --no-sync python -m pytest tests/` (7,278 passed)

Closes #31991

### DIFF
--- a/crates/runner/mitm-addon/src/auth.py
+++ b/crates/runner/mitm-addon/src/auth.py
@@ -631,8 +631,8 @@ def _apply_header_query_injection(
     if auth_pairs:
         resolved_fields_by_name: dict[bytes, tuple[bytes, bytes]] = {}
         for header_name, header_value in auth_pairs:
-            encoded_name = header_name.encode()
-            encoded_value = header_value.encode()
+            encoded_name = header_name.encode("ascii")
+            encoded_value = header_value.encode("latin-1")
             normalized_name = encoded_name.lower()
             existing_field = resolved_fields_by_name.get(normalized_name)
             if existing_field is None:
--- a/crates/runner/mitm-addon/src/auth_base_transport.py
+++ b/crates/runner/mitm-addon/src/auth_base_transport.py
@@ -368,10 +368,11 @@ def resolved_auth_header_pairs(headers) -> list[tuple[str, str]]:
     """Validate and filter resolved auth headers before outbound injection.
 
     Each resolved header name and value is validated before any pair is
-    returned; invalid pairs raise ``InvalidResolvedAuthHeaderError``. Transport,
-    authority, and framing headers are then dropped. This helper does not merge
-    with or replace client headers; callers that combine resolved and client
-    headers own that policy.
+    returned; invalid pairs raise ``InvalidResolvedAuthHeaderError``. Accepted
+    names are ASCII and accepted values map one-to-one to wire bytes through
+    Latin-1. Transport, authority, and framing headers are then dropped. This
+    helper does not merge with or replace client headers; callers that combine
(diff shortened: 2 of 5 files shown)


seven332:
## Code Review: PR #32022

**Reviewed head:** `20cc8cd1f2965593654facd3ce2b2e2dc4fc4451`

### Summary

The change correctly applies the existing Latin-1 resolved-header contract to mitmproxy's byte-backed ordinary request path while leaving the already-correct auth.base serializer unchanged. The follow-up coverage also pins fail-closed behavior for values outside Latin-1.

### Findings

No findings.

### Testing

Raw bytes are asserted for the ordinary `request`, `requestheaders`, and full auth.base forwarding paths. Invalid non-Latin-1 values are covered before ordinary mutation and auth.base forwarding. Local validation passed with 121 focused tests, Ruff format/lint, BasedPyright, and all 7,278 addon tests. No live third-party credential exchange was run; the existing fake upstream captures the real internal HTTP serializer's socket bytes.

### Verdict

LGTM

PR #32024, opened 2026-09-06
fix(mitm-addon): redact whitespace-separated url userinfo

### LINKED ISSUE #32000: fix(mitm-addon): redact whitespace-separated malformed URL userinfo
## Problem and evidence

`crates/runner/mitm-addon/src/network_log_sanitization.py` is the shared boundary for removing URL userinfo, query data, and fragments before URLs are written to persistent proxy and webhook diagnostics.

Its malformed HTTP(S) authority recovery currently strips only `/` and `\\` before identifying the first authority-like segment. Python keeps embedded space (`SP`), vertical tab (`VT`), and form feed (`FF`) between a recognized HTTP(S) scheme and the separator run in `SplitResult.path`. The recovery therefore fails to identify the later `@` as authority userinfo and persists credentials.

Confirmed examples include:

```text
https: //user:SECRET@api.vm0.ai/path?token=hidden#fragment
https:\v//user:SECRET@api.vm0.ai/path?token=hidden#fragment
https:\f//user:SECRET@api.vm0.ai/path?token=hidden#fragment
```

On Python 3.14.0, the current sanitizer retains each whitespace character and `user:SECRET@api.vm0.ai/path`; the issue audit also reproduced the space case on Python 3.12.3. Query and fragment data are removed, but the userinfo remains. The public persistence path in `crates/runner/mitm-addon/src/logging_utils.py` and invalid-webhook diagnostics in `crates/runner/mitm-addon/src/usage/webhook.py` both reuse this sanitizer.

The existing persisted-URL matrix covers valid userinfo, missing/extra slashes, backslashes, leading C0/space, and globally removable TAB/CR/LF characters, but not SP, VT, or FF at this transition. Existing tests therefore pass while these values remain credential-bearing.

Runner startup validation rejects malformed `server.url` values, but this sanitizer intentionally remains an independent persistent-log boundary for direct addon inputs and future callers, as established by #21261.

No duplicate open issue was found for this whitespace-specific bypass.

## Decision

Proceed with a production sanitizer fix and regression coverage. This is a confidentiality defect at a persistent-log boundary, not only a missing-test problem.

Constrain recovery to ASCII whitespace between a recognized HTTP(S) scheme colon and a following slash/backslash separator run. Continue treating only the first authority-position segment as a userinfo candidate. Do not scan or redact arbitrary path segments.

- Extend the existing malformed-authority recovery so SP, VT, and FF cannot hide the separator run after an HTTP(S) scheme.
- Keep Python's existing preprocessing contract: TAB, CR, and LF continue to be removed globally, while leading C0/space handling remains unchanged.
- Reuse the existing final-`@` netloc sanitizer and canonical host/path reconstruction rather than adding a second URL parser or a broad secret detector.
- Exercise real JSONL persistence and one invalid-webhook diagnostic path so the boundary is tested through public behavior.

This keeps the implementation local, dependency-free, linear in input size, and compatible with the existing malformed-authority policy from #21261.

## Alternatives considered

- **Add tests only:** rejected because the current output demonstrably contains credentials.
- **Rely on runner startup validation:** rejected because a persistent-log boundary cannot depend on every current and future caller validating first.
- **Strip all ASCII whitespace globally before parsing:** rejected because it would silently concatenate path content and broaden URL normalization beyond the evidenced transition.
- **Redact every prefix before `@` in any path:** rejected because ordinary path `@` values are intentionally preserved for diagnostics.
- **Replace the sanitizer with a WHATWG URL parser:** rejected because the dependency and broad canonicalization changes are disproportionate to this bounded bypass.

## Risks and mitigations

- **False-positive path redaction:** require an HTTP(S) scheme followed by the targeted whitespace and a slash/backslash separator run; preserve ordinary later-path `@` cases in regression coverage.
- **Parser-version differences:** assert the persisted safety invariant and stable useful host/path output on the supported Python environment; cover all three retained ASCII whitespace characters explicitly.
- **Output compatibility:** leave valid URLs and existing malformed slash/backslash, unmatched-bracket, leading-control, and ordinary path behavior unchanged.
- **Incomplete sink coverage:** test both structured proxy-log output and an invalid-webhook diagnostic entry, including `message`, `error`, and `url`.

## Acceptance criteria

- Add deterministic persisted-log regression coverage for HTTP(S) scheme separators containing embedded SP, VT, and FF before an authority-like `user:pass@host` segment.
- Make those cases exclude userinfo, query data, and fragments from the persisted structured `url` value while retaining useful host/path diagnostics.
- Exercise at least one invalid-webhook diagnostic path and assert that `message`, `error`, structured `url`, and the complete serialized entry do not contain the credential.
- Preserve existing valid URL, malformed slash/backslash, unmatched-bracket, leading-control, and ordinary path-`@` behavior.
- Keep the sanitizer bounded, dependency-free, and local to log rendering; do not change request URLs, validation, delivery, retries, firewall behavior, billing, or raw runtime metadata.
- Run Ruff formatting/linting, basedpyright, focused logging/webhook tests, and the complete mitm-addon pytest suite.

## Delivery notes

This remains one independently reviewable Python mitm-addon PR. It requires no API, database, schema, runner protocol, rollout coordination, or data migration.

seven332:
## Summary

- treat space, vertical tab, and form feed as part of the leading malformed HTTP(S) authority prefix before persistent-log URL rendering
- recover safely when a mixed prefix is split between `netloc` and `path`, while keeping slash and reverse-solidus as the only authority/path cut points
- cover proxy JSONL, invalid-webhook diagnostics, and network-log rendering while preserving raw request metadata

## Scope

- log rendering only
- no request URL, validation, retry, delivery accounting, firewall, billing, schema, protocol, dependency, or migration changes
- proxy/webhook query and fragment removal and network-log query and fragment retention remain unchanged

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_logging_utils.py tests/test_webhook_http_delivery.py tests/test_response_handler_logging.py` (149 passed)
- `uv run --no-sync python -m pytest tests/` (7,282 passed)
- Python 3.12.3 and 3.14.0 malformed-prefix matrix checks (781 prefixes each)
- `git diff --check`
- repository pre-commit hook

Closes #32000

### DIFF
--- a/crates/runner/mitm-addon/src/network_log_sanitization.py
+++ b/crates/runner/mitm-addon/src/network_log_sanitization.py
@@ -8,6 +8,7 @@ _URLSPLIT_LEADING_STRIP_CHARACTERS = "".join(chr(codepoint) for codepoint in ran
 _URLSPLIT_REMOVABLE_CHARACTERS = "\t\r\n"
 _SPECIAL_URL_SCHEMES = ("http", "https")
 _SPECIAL_URL_SEPARATORS = "/\\"
+_MALFORMED_AUTHORITY_PREFIX_CHARACTERS = f"{_SPECIAL_URL_SEPARATORS} \v\f"
 
 
 def _normalize_for_urlsplit(value: str) -> str:
@@ -48,7 +49,9 @@ def _sanitize_malformed_authority_for_network_log(
     if not has_special_scheme and not is_protocol_relative:
         return None
 
-    authority_path = parts.path.lstrip(_SPECIAL_URL_SEPARATORS)
+    # SP, VT, and FF survive urlsplit preprocessing when embedded after the
+    # scheme. Treat them only as part of the leading malformed separator run.
+    authority_path = parts.path.lstrip(_MALFORMED_AUTHORITY_PREFIX_CHARACTERS)
     cut_points = [
         index
         for separator in _SPECIAL_URL_SEPARATORS
--- a/crates/runner/mitm-addon/src/network_log_sanitization.py
+++ b/crates/runner/mitm-addon/src/network_log_sanitization.py
@@ -39,9 +39,9 @@ def _sanitize_url_text_fallback_for_network_log(value: str) -> str:
 def _sanitize_malformed_authority_for_network_log(
     value: str, parts: urllib.parse.SplitResult
 ) -> str | None:
-    # A mixed separator run can leave only backslashes in netloc while the
-    # actual authority-like segment lands in path.
-    if parts.netloc.strip(_SPECIAL_URL_SEPARATORS):
+    # A mixed separator run can leave separators or retained whitespace in
+    # netloc while the actual authority-like segment lands in path.
+    if parts.netloc.strip(_MALFORMED_AUTHORITY_PREFIX_CHARACTERS):
         return None
 
     has_special_scheme = parts.scheme in _SPECIAL_URL_SCHEMES
--- a/crates/runner/mitm-addon/tests/test_logging_utils.py
+++ b/crates/runner/mitm-addon/tests/test_logging_utils.py
@@ -203,6 +203,11 @@ class TestLogProxyEntry:
                 id="form-feed-before-separators",
             ),
             pytest.param(
+                "https:// /user:pass@example.com/path?token=secret#fragment",
+                "https://example.com/path",
+                id="space-inside-separator-prefix",
+            ),
+            pytest.param(
                 "https:////first@user:pass@example.com/path?token=secret#fragment",
                 "https://example.com/path",
                 id="multiple-at-signs",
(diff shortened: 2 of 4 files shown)


seven332:
## Code Review: PR #32024

**Reviewed head:** `2d5e547fce240256c2aabe3737758bcef18d364a`

### Summary

The current head extends the existing malformed-authority grammar with the retained ASCII whitespace characters at both parser component boundaries, without changing normal URL parsing, later path delimiters, request behavior, or persistence schemas. Coverage exercises real proxy/network JSONL and the invalid-webhook entry point.

### Findings

No findings on the current head. An intermediate mixed-prefix gap found during self-review was fixed by `2d5e547fce` and now has a persisted-output regression case.

### Testing

Ruff format/lint, basedpyright, 149 focused integration tests, all 7,282 mitm-addon tests, repository hooks, and generated Python 3.12.3/3.14.0 matrices of 781 sensitive and ordinary-path prefixes each passed locally. CI status and external feedback are checked separately by the PR workflow.

### Verdict

LGTM

PR #32026, opened 2026-09-06
perf(python): bound accept-encoding negotiation work

### LINKED ISSUE #32001: perf(python): bound Accept-Encoding negotiation work
## Problem

At audited `main` commit `a0bc24a59d24f8215165e8e2cce4976d46272721`, response-encoding negotiation performs unbounded synchronous parsing and allocation over a sandbox-controlled `Accept-Encoding` value before a billable request is sent upstream.

[`normalize_accept_encoding_for_body_inspection()`](https://github.com/vm0-ai/vm0/blob/a0bc24a59d24f8215165e8e2cce4976d46272721/crates/runner/mitm-addon/src/response_encoding_negotiation.py#L28-L118) first calls `Headers.get_all()`, which decodes and copies all matching raw values. It then materializes every comma-separated coding with `raw_value.split(",")`. Each accepted coding is split on semicolons in [`_parse_coding()`](https://github.com/vm0-ai/vm0/blob/a0bc24a59d24f8215165e8e2cce4976d46272721/crates/runner/mitm-addon/src/response_encoding_negotiation.py#L121-L151) and split again in [`_q_text()`](https://github.com/vm0-ai/vm0/blob/a0bc24a59d24f8215165e8e2cce4976d46272721/crates/runner/mitm-addon/src/response_encoding_negotiation.py#L154-L159). There is no value-byte, token, or work budget.

The generic [`requestheaders()` guard](https://github.com/vm0-ai/vm0/blob/a0bc24a59d24f8215165e8e2cce4976d46272721/crates/runner/mitm-addon/src/mitm_addon.py#L749-L770) limits raw field count to 2,048 and each field name to 4,096 bytes, but does not limit field-value or aggregate request-head bytes. A single valid field therefore bypasses both guards.

## Reachability and impact

The normalizer runs synchronously on the shared mitmproxy event loop for response-body usage inspection:

- during the header-only path before the first auth await at [`mitm_addon.py:1051`](https://github.com/vm0-ai/vm0/blob/a0bc24a59d24f8215165e8e2cce4976d46272721/crates/runner/mitm-addon/src/mitm_addon.py#L1038-L1058);
- during the buffered request path before firewall auth at [`mitm_addon.py:1442`](https://github.com/vm0-ai/vm0/blob/a0bc24a59d24f8215165e8e2cce4976d46272721/crates/runner/mitm-addon/src/mitm_addon.py#L1430-L1450);
- for billable model-provider requests and billable connectors with a registered response parser, as selected by [`_expects_http_response_body_usage_inspection()`](https://github.com/vm0-ai/vm0/blob/a0bc24a59d24f8215165e8e2cce4976d46272721/crates/runner/mitm-addon/src/mitm_addon.py#L1509-L1545).

Pinned mitmproxy 12.2.3's HTTP/1 server reader accumulates input until `maybe_extract_lines()` returns a complete request head and has an explicit `FIXME: protect against header size DoS`; it applies no aggregate request-head limit before emitting `RequestHeaders`. The oversized value is therefore reachable from a sandbox HTTP/1 client, not only through a constructed test flow.

The public normalizer was benchmarked with genuine mitmproxy `Headers` under pinned CPython 3.14.0 / x86_64 / mitmproxy 12.2.3. Values were syntactically valid repeated `gzip` codings ending in `zstd`; input construction was excluded. Medians are from five calls:

| Raw value bytes | Parsed codings | Median | Range |
| ---: | ---: | ---: | ---: |
| 29 | 6 | 0.038 ms | 0.031-0.043 ms |
| 65,534 | 13,107 | 7.541 ms | 7.192-7.828 ms |
| 1,048,574 | 209,715 | 110.596 ms | 109.830-144.480 ms |
| 3,145,724 | 629,145 | 333.773 ms | 332.048-356.564 ms |

A separate `tracemalloc` run over the 1 MiB value observed 11.747 MiB of additional peak allocation inside the call. One sandbox request can consequently monopolize the proxy event loop for hundreds of milliseconds and create roughly 12x input-size transient allocation; repetition can delay unrelated proxy flows and add memory pressure.

## Distinctness

- #31645 added the 2,048-field guard but explicitly retained aggregate request-head bytes as residual exposure. It does not bound one field value or this parser's token materialization.
- #30208 and #21142 cover `Accept-Encoding` semantics and response-time inspection behavior, not parser work limits.
- No open issue with `Accept-Encoding` or encoding-negotiation keywords covers this performance boundary.

## Decision

Add one inclusive 64 KiB aggregate raw-value budget for `Accept-Encoding` before dependency string access. This matches the existing managed SigV4 request-header-list scale while remaining orders of magnitude above ordinary negotiation values. Inspect `headers.fields`, identify matching names case-insensitively, and sum matching raw value lengths. Stop as soon as the limit is exceeded, before calling `Headers.get_all()`, decoding values, splitting tokens, or constructing parser collections.

For an over-budget header, preserve the original requester value and return a distinct fixed negotiation outcome for the budget fallback. Do not force `identity` or another coding without parsing, because the oversized value may contain explicit rejections. Existing response-time `Content-Encoding` handling remains authoritative and can continue to fail closed or use a bounded fallback according to the actual response.

Within the budget, avoid materializing the complete comma-token list and avoid splitting each accepted coding twice. A streaming parser can reduce ordinary allocation, but it is not a substitute for the hard raw-input budget.

## Acceptance criteria

- Define and document an inclusive 64 KiB aggregate raw `Accept-Encoding` value budget.
- Check the budget from raw header fields before `Headers.get_all()`, value decoding, comma splitting, semicolon splitting, `Decimal` parsing, or result-collection growth.
- Accept exactly 64 KiB with current negotiation semantics; at limit-plus-one, preserve the original header and return a distinct fixed budget-fallback outcome without decoding any matching value.
- Add structural coverage with a `bytes` subtype whose `decode()` raises, including a combined repeated-field overflow case, to prove the over-limit path does not invoke dependency string access.
- Preserve the existing semantic matrix for absent and under-budget values, including mixed-case/repeated fields, qvalues, wildcards, duplicates, and explicit `identity` or safe-coding rejection.
- Exercise both request-header and buffered request hook paths for a billable inspected flow and verify the fixed outcome is available to response-time diagnostics.
- Refactor in-budget parsing so it does not materialize the complete comma-token list or split accepted codings twice.
- Record a before/after diagnostic at ordinary, 64 KiB, 1 MiB, and 3 MiB inputs; keep benchmark code and wall-clock assertions out of CI.
- Keep the complete mitm-addon Python validation sequence passing.

## Compatibility and residual exposure

This is runner-local and changes no API, queue payload, persisted state, or cross-deployment protocol. It intentionally bounds addon-owned negotiation work only. Mitmproxy's pre-hook HTTP/1 buffering/parsing and other oversized header values remain separate residual exposure.

seven332:
## Summary

- enforce an inclusive 64 KiB aggregate raw `Accept-Encoding` value budget before mitmproxy decodes matching fields
- preserve over-budget requester fields with the distinct `preserved_work_budget` diagnostic outcome
- stream comma and semicolon traversal and retain validated qvalue text in one pass
- cover exact and repeated-field boundaries, both request entry paths, and response-time fail-closed diagnostics

- inputs through 64 KiB retain the existing negotiation semantics
- over-budget inputs are forwarded unchanged; the actual response `Content-Encoding` remains authoritative and still fails closed when it cannot be inspected safely
- this is runner-local flow metadata with no API, queue, persisted-state, or cross-deployment contract change
- mitmproxy's request-head buffering before addon hooks and global limits for unrelated header values remain outside this PR

## Diagnostic

The benchmark used genuine mitmproxy `Headers` with repeated valid `gzip` codings followed by `zstd`; input construction was excluded and each row is the median of five calls under the committed CPython 3.14 environment. Benchmark code and wall-clock assertions are not included in CI.

| Raw value bytes | Before | After | After outcome |
| ---: | ---: | ---: | --- |
| 29 | 0.006 ms | 0.006 ms | `rewritten_stream_decodable` |
| 65,534 | 4.307 ms | 4.228 ms | `rewritten_stream_decodable` |
| 1,048,574 | 90.077 ms | <0.001 ms | `preserved_work_budget` |
| 3,145,724 | 260.972 ms | <0.001 ms | `preserved_work_budget` |

At 1,048,574 bytes, additional `tracemalloc` peak allocation fell from 11.746 MiB to 0.000046 MiB.

## Validation

- `uv run --no-sync python -m pytest tests/test_response_encoding_negotiation.py tests/test_response_encoding_inspection_risk.py` (110 passed)
- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync basedpyright -p .` (0 errors, 0 warnings)
- `uv run --no-sync python -m pytest tests/` (7282 passed)

Closes #32001

### DIFF
--- a/crates/runner/mitm-addon/src/response_encoding_negotiation.py
+++ b/crates/runner/mitm-addon/src/response_encoding_negotiation.py
@@ -1,14 +1,19 @@
 """Request-side response encoding negotiation for body-inspected flows."""
 
 import re
+from collections.abc import Iterable, Iterator
 from decimal import Decimal, InvalidOperation
-from typing import Literal
+from typing import Final, Literal
 
 from mitmproxy import http
 
 import body_decoding
 
 _ACCEPT_ENCODING = "Accept-Encoding"
+_RAW_ACCEPT_ENCODING: Final = b"accept-encoding"
+# Accept an inclusive 64 KiB aggregate of matching raw values. This bounds
+# addon-owned negotiation work after mitmproxy has parsed the request head.
+_MAX_ACCEPT_ENCODING_VALUE_BYTES: Final = 64 * 1024
 _IDENTITY = "identity"
 _WILDCARD = "*"
 _STREAM_DECODABLE_ENCODINGS = body_decoding.stream_decodable_content_encodings()
@@ -22,6 +27,7 @@ type ResponseEncodingNegotiationOutcome = Literal[
     "already_stream_decodable",
     "rewritten_stream_decodable",
     "preserved_client_constraints",
+    "preserved_work_budget",
 ]
 
 
@@ -35,13 +41,17 @@ def normalize_accept_encoding_for_body_inspection(
     provide the needed bounded-output behavior. Brotli participates under the
     documented soft output-limit contract in ``body_decoding``.
 
-    The helper preserves explicit requester rejections.  When ``identity`` and
-    every supported compression coding are rejected, the original header remains
-    unchanged and the actual response might not support bounded streaming
-    inspection.  Callers must use the response's ``Content-Encoding`` to decide
-    whether incremental body inspection is available. The return value records
-    whether the request was already safe, rewritten, or preserved by constraint.
+    The helper preserves explicit requester rejections. When ``identity`` and
+    every supported compression coding are rejected, or the matching raw values
+    exceed the parser's work budget, the original header remains unchanged and
+    the actual response might not support bounded streaming inspection. Callers
+    must use the response's ``Content-Encoding`` to decide whether incremental
+    body inspection is available. The return value records whether the request
+    was already safe, rewritten, or preserved and why.
     """
+    if not _accept_encoding_values_within_budget(headers):
+        return "preserved_work_budget"
+
     values = headers.get_all(_ACCEPT_ENCODING)
     if not values:
         headers[_ACCEPT_ENCODING] = _IDENTITY
@@ -57,13 +67,12 @@ def normalize_accept_encoding_for_body_inspection(
     wildcard_rejected = False
 
     for raw_value in values:
-        for raw_coding in raw_value.split(","):
-            name, q_value = _parse_coding(raw_coding)
+        for raw_coding in _iter_delimited(raw_value, ","):
+            name, q_value, q_text = _parse_coding(raw_coding)
             if not name:
                 continue
 
             accepted = q_value is None or q_value > _MIN_Q_VALUE
-            q_text = _q_text(raw_coding)
 
             if name == _IDENTITY:
                 if accepted:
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #32026

**Reviewed head:** `1023e28125bc1268c1a2bc50524e6c3f7f6ed925`

### Summary

Adds an inclusive 64 KiB raw `Accept-Encoding` work boundary before mitmproxy decoding, preserves over-budget values with a distinct diagnostic outcome, and makes accepted-input token traversal incremental without changing negotiation semantics.

### Findings

No findings.

### Testing

Coverage is adequate: exact-limit, limit-plus-one, repeated-field no-decode, buffered and header-phase request hooks, and response-time fail-closed diagnostics are covered. The full mitm-addon suite passed (7282 tests), along with Ruff, basedpyright, uv lock validation, and a 20,000-case randomized differential check against `origin/main`. A live external proxy E2E was not run; the repository's addon testing guide does not require real interception for this parser/hook surface.

### Verdict

LGTM

PR #32027, opened 2026-09-06
fix(python): track set and dict display hashing failures

### LINKED ISSUE #32002: bug(python): track set and dict hashing failures in metadata linter
## Problem and evidence

At commit a0bc24a59d24f8215165e8e2cce4976d46272721, the flow-metadata linter does not model the hashing performed while constructing set and dict displays.

_is_modeled_implicit_exception_operation() in crates/runner/mitm-addon/scripts/flow_metadata_linter/ast_helpers.py:792-810 omits ast.Set and ast.Dict. Their specialized visitors in crates/runner/mitm-addon/scripts/flow_metadata_linter/visitor.py:799-819 traverse the contents but do not record an implicit exceptional alias state for literal insertion or key hashing.

This creates a deterministic false negative because flow.metadata is a dict and therefore unhashable:

    def set_hash_failure(flow):
        try:
            metadata = flow.metadata
            {metadata}
            metadata = {}
        except TypeError:
            return metadata["sandbox_run_id"]

    def dict_hash_failure(flow):
        try:
            metadata = flow.metadata
            {metadata: 1}
            metadata = {}
        except TypeError:
            return metadata["sandbox_run_id"]

At runtime, both displays raise TypeError before the later rebind, so each handler accesses the original flow.metadata alias. The public metadata_key_violations() API nevertheless returns [] for the complete source.

The existing focused metadata-key contract suite passes, confirming that this exception path currently has no regression coverage.

## Impact

A reachable raw registered-key access can bypass the repository metadata-key guard when set-element or dict-key hashing transfers control to an exception handler or suppressing context manager. This weakens the centralized metadata_keys contract and can let key renames or lifecycle changes drift.

This is a repository-tooling soundness gap; there is no evidence of a deployed runner failure.

## Acceptance criteria

- Set-display element hashing and explicit dict-display key hashing contribute modeled implicit-exception alias state.
- Both reproductions report use metadata_keys.SANDBOX_RUN_ID for flow.metadata access through the public metadata_key_violations() API.
- Recording follows Python evaluation order. In particular, a dict value expression that rebinds the alias before its key is hashed must not retain the pre-value alias.
- Deterministic non-metadata rebinds without a preceding modeled failure remain non-metadata.
- Add focused exact-output regression fixtures for try handlers and/or suppressing context managers without duplicate diagnostics.
- Existing flow-metadata linter contract tests and the repository metadata-key guard continue to pass.

seven332:
## Summary

- record possible set-element hashing failures after each ordinary element expression
- traverse dictionary entries in key/value order and record explicit-key hashing failures after each value
- add exact-output public linter coverage for handlers, suppressing contexts, ordered rebinds, and diagnostic de-duplication

## Scope

This change only strengthens the flow-metadata source linter. Starred set expansion and dictionary-unpack operation failures remain unchanged and out of scope. There are no API, schema, deployment, or runtime runner changes.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_flow_metadata_key_contract.py` (25 passed)
- `uv run --no-sync python -m pytest tests/` (7,277 passed)
- `git diff --check`

Closes #32002

### DIFF
--- a/crates/runner/mitm-addon/scripts/flow_metadata_linter/visitor.py
+++ b/crates/runner/mitm-addon/scripts/flow_metadata_linter/visitor.py
@@ -796,7 +796,7 @@ class _MetadataKeyVisitor(ast.NodeVisitor):
             return
         self.visit(node.value)
 
-    def _visit_sequence_expression(self, node: ast.List | ast.Tuple | ast.Set) -> None:
+    def _visit_sequence_expression(self, node: ast.List | ast.Tuple) -> None:
         for element in node.elts:
             self._record_metadata_merge_key_violations(element)
         self.generic_visit(node)
@@ -808,7 +808,12 @@ class _MetadataKeyVisitor(ast.NodeVisitor):
         self._visit_sequence_expression(node)
 
     def visit_Set(self, node: ast.Set) -> None:
-        self._visit_sequence_expression(node)
+        for element in node.elts:
+            self._record_metadata_merge_key_violations(element)
+        for element in node.elts:
+            self.visit(element)
+            if not isinstance(element, ast.Starred):
+                self._record_implicit_exception_aliases()
 
     def visit_Dict(self, node: ast.Dict) -> None:
         node_is_metadata_merge = self._is_metadata_merge_value(node)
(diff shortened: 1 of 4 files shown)


seven332:
## Code Review: PR #32027

**Reviewed head:** `d3957226ce06f65a961d67c13e77d55dc45bbf11`

### Summary

The change adds ordered implicit-exception state for ordinary set-element and explicit dict-key hashing while preserving the approved `*`/`**` exclusion and existing metadata-merge behavior.

### Findings

No findings.

### Testing

The public exact-output contract covers both reproductions, try and suppressing-context paths, rebind ordering across entries and within dict pairs, no duplicate diagnostics, deterministic non-alias cases, and unpack exclusions. Ruff, BasedPyright, the focused 25-test contract, repository guard, diff check, and all 7,277 addon tests passed locally. Hosted CI remains to be checked separately.

### Verdict

LGTM

PR #32028, opened 2026-09-06
perf(runner): back off active input retries

### LINKED ISSUE #31998: perf(rust): avoid fixed-rate active-input payload reprocessing
## Problem and evidence

On current `main`, `forward_with_retry` retries definitely-not-written active-input control outcomes every 250 ms until cancellation (`crates/runner/src/executor/active_input.rs`). Every attempt calls `encode_active_input`, validates the resulting size, clones the delivery ID, and transfers a new `Vec<u8>` into process control.

Explicit `QueueFull` and `SinkUnavailable` outcomes enter this loop. The Guest has already decoded the JSON payload and computed SHA-256 over the text before checking capacity (`crates/guest-agent/src/active_input.rs`). With the 1,048,576-byte control payload limit, an immediate capacity rejection can cause about 240 MiB/minute per affected run of repeated host/Guest traffic, plus repeated Runner JSON work and Guest parsing/hashing.

Existing executor integration coverage verifies stable delivery identity and identical payload bytes across retries, but it does not constrain sustained retry cadence or exercise a near-limit payload.

## Decision

Proceed with a Runner-only prepared payload and deterministic capped exponential backoff.

- Encode and size-check once per reserved API or local input.
- Keep the first control attempt immediate.
- Retry definitely-not-written outcomes after 250 ms, then 500 ms, 1 s, 2 s, and at most 4 s between later attempts.
- Keep retrying until cancellation or a terminal/uncertain outcome; do not add a finite attempt budget.
- Keep the Runner/Guest wire format, sandbox control contract, stable delivery ID, and delivery-certainty classifications unchanged.

Jitter is not included because the constrained queue and sink are run-local; there is no evidenced shared capacity target that synchronized runners would overload. A deterministic schedule is simpler and directly testable.

## Alternatives considered

- **Reference-counted payload in the sandbox control interface:** could avoid the remaining per-attempt full-buffer clone, but expands the change across providers and mocks while transports may still copy. The broader contract change is not justified by current evidence.
- **Guest readiness signal:** removes polling most completely, but requires new lost-wakeup, reconnect, cancellation, timeout, and mixed-version protocol semantics.
- **Earlier Guest capacity check:** may reduce residual hashing but cannot address Runner serialization or repeated host/Guest traffic and does not solve this issue alone.
- **Status quo:** preserves sub-250 ms recovery indefinitely but leaves resource amplification unbounded during backpressure.

## Risks and mitigations

- After sustained backpressure, acceptance may be delayed by up to roughly 4 seconds after capacity returns. Fast initial retries preserve transient responsiveness, and the cap bounds latency.
- The existing owned process-control callback still requires a `Vec<u8>` clone per attempt. Reducing attempt frequency bounds that cost; changing the shared contract is deferred unless profiling shows the residual copy is material.
- A near-limit retry test retains several large payload copies in the sandbox mock. Keep the attempt count bounded while covering the maximum delay.
- Start Tokio virtual time only after the first observable control attempt so loopback HTTP setup remains on real time.

## Acceptance criteria

- The Runner source structure prepares and validates the immutable payload once before its retry loop; per-attempt transport does not call the encoder or repeat size validation.
- Definitely-not-written retry outcomes use the immediate-first capped exponential schedule of 250 ms, 500 ms, 1 s, 2 s, and 4 s maximum rather than fixed 4 Hz polling.
- Stable delivery ID, cancellation responsiveness, and the distinction between definitely-not-written and possibly-written outcomes are preserved.
- The existing executor integration boundary covers consecutive `QueueFull`/`SinkUnavailable` outcomes with a near-limit payload, asserts no attempts before retry deadlines, verifies the 4-second cap, and verifies the exact same ID and encoded bytes on every attempt.
- Focused Runner formatting, tests, Clippy, and documentation checks pass.

## Delivery notes

This is one independently reviewable Runner PR. No API schema, persisted data, migration, Guest change, feature flag, or coordinated rollout is required.

seven332:
## Summary

- prepare and validate each active-input control payload once before retrying
- replace fixed 250 ms retry polling with exponential backoff capped at 4 seconds
- cover near-limit payload reuse and retry cadence through the Runner executor integration path

## Compatibility

- preserves the existing Runner/Guest payload and process-control contracts
- preserves delivery IDs, cancellation priority, and definitely-not-written versus possibly-written handling
- does not add a finite retry budget or coordinated rollout requirement

## Validation

- `cargo test --manifest-path crates/Cargo.toml --profile local -p runner --bin runner executor::tests::agent_run_tests::active_input::run_in_sandbox_retries_guest_backpressure_with_same_id -- --exact`
- `cargo test --manifest-path crates/Cargo.toml --profile local -p runner --bin runner` (3452 passed, 25 ignored)
- `RUSTDOCFLAGS="-D warnings" cargo doc --manifest-path crates/Cargo.toml --profile local -p runner --no-deps`

Closes #31998

### DIFF
--- a/crates/runner/src/executor/active_input.rs
+++ b/crates/runner/src/executor/active_input.rs
@@ -22,6 +22,8 @@ use crate::ids::RunId;
 
 const ACTIVE_INPUT_READ_RETRY_INTERVAL: Duration = Duration::from_millis(250);
 const ACTIVE_INPUT_CONTROL_TIMEOUT: Duration = Duration::from_secs(1);
+pub(super) const ACTIVE_INPUT_CONTROL_RETRY_INITIAL_INTERVAL: Duration = Duration::from_millis(250);
+pub(super) const ACTIVE_INPUT_CONTROL_RETRY_MAX_INTERVAL: Duration = Duration::from_secs(4);
 const ACTIVE_INPUT_JOURNAL_READ_TIMEOUT: Duration = Duration::from_secs(5);
 const ACTIVE_INPUT_RECEIPT_RECOVERY_TIMEOUT: Duration = Duration::from_secs(5);
 const FIRST_ACTIVE_INPUT_SEQUENCE: u64 = 1;
@@ -85,6 +87,11 @@ enum ForwardDisposition {
     Retry,
     Suppress,
     Stop,
+}
+
+struct PreparedActiveInput {
+    delivery_id: String,
+    payload: Vec<u8>,
 }
 
 async fn run_forwarder(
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #32028

**Reviewed head:** `c8c1ba0e05430fd69fc637ca0e399cb43abda3b1`

### Summary

The Runner now prepares each active-input payload once, releases the raw source text, and applies deterministic exponential retry backoff capped at four seconds without changing the Runner/Guest contract or delivery-certainty behavior.

### Findings

No findings.

### Testing

The near-limit executor integration test verifies alternating `QueueFull`/`SinkUnavailable` responses, stable delivery ID and exact payload bytes, each exponential interval, and the four-second cap. Local validation passed the exact regression test, all Runner binary tests (3452 passed, 25 repository-defined ignored), formatting, all-target/all-feature Clippy, and rustdoc with warnings denied. Repository CI is still in progress; workspace-wide Rust tests were not run because the change is Runner-only and does not modify shared contracts.

### Verdict

LGTM

PR #32135, opened 2026-09-07
fix(python): reject non-utf-8 catalog response headers

### LINKED ISSUE #32081: test(python): cover non-UTF-8 Codex catalog response headers
## Problem

At audit commit `7b89f334da0b1cd8b14eeee723761f88ee9962bb`, the Codex model catalog cache assumes selected response-header values can always be encoded as UTF-8:

- `crates/runner/mitm-addon/src/codex_model_catalog_cache.py:367-384` calls `value.encode()` while validating `ETag` and `x-models-etag` values.
- `crates/runner/mitm-addon/src/codex_model_catalog_cache.py:737-758` calls `content_type.encode()` before classifying a catalog response's `Content-Type`.
- `crates/runner/mitm-addon/src/codex_model_catalog_cache.py:808-856` and `:1090-1112` invoke those validators directly from `responseheaders()`.

Mitmproxy preserves arbitrary raw header bytes in Python strings through surrogate escapes. For example, `http.Headers([(b"etag", b'"\xff"')]).get("etag")` returns a string containing `\udcff`, so the UTF-8 encode calls raise `UnicodeEncodeError` before the cache can reject or safely handle the value.

The current response tests at `crates/runner/mitm-addon/tests/test_codex_model_catalog_cache_responses.py:159-307` and ETag-signal tests at `crates/runner/mitm-addon/tests/test_codex_model_catalog_cache_lifecycle.py:95-150` construct headers from normal Python strings only, so they do not exercise this raw-byte boundary.

## Reproduction evidence

A focused reproduction prepared a catalog miss and supplied real `mitmproxy.http.Headers` with raw byte values. All three cases raised from `mitm_addon.responseheaders()`:

```text
ETag:          b'"\xff"'             -> UnicodeEncodeError
Content-Type:  b'application/\xffjson' -> UnicodeEncodeError
x-models-etag: b'"\xff"'             -> UnicodeEncodeError
```

Immediately after each catalog-header exception, the cache still reported one active flow state and one in-flight owner; normal terminal cleanup was required to release them.

## Impact

An upstream-controlled non-UTF-8 response-header byte can abort the response-header hook instead of producing the cache's bounded pass-through behavior. For catalog requests this interrupts normal processing while the single-flight owner is still reserved, and for authenticated Responses traffic a malformed `x-models-etag` can fail the hook before ordinary response handling begins.

## Acceptance criteria

- Add regression coverage using raw `mitmproxy.http.Headers` bytes for non-UTF-8 `ETag`, `Content-Type`, and `x-models-etag` values.
- `responseheaders()` does not raise for any of these cases.
- Unsafe catalog headers produce a deterministic pass-through/not-stored result, and single-flight capacity plus followers are released normally.
- An unusable `x-models-etag` signal is ignored without changing matching cache entries or in-flight ownership.
- Existing valid ASCII header behavior and cache hits remain covered.

seven332:
## Summary

- reject surrogate-bearing Codex catalog ETag and Content-Type values through the cache's existing not-stored paths
- ignore an unusable authenticated `x-models-etag` without changing cached entries or in-flight ownership
- cover raw mitmproxy header bytes through hook-level owner/follower and lifecycle integration tests

## Scope

- no public API, persisted state, runner protocol, dependency, telemetry schema, or deployment-order changes
- no broad response-hook exception handling; only `UnicodeEncodeError` from strict header-value encoding is converted into an unusable external value

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest -q tests/test_codex_model_catalog_cache_coordination.py tests/test_codex_model_catalog_cache_lifecycle.py tests/test_codex_model_catalog_cache_responses.py` (76 passed)
- `uv run --no-sync python -m pytest tests/` (7296 passed)

Closes #32081

### DIFF
--- a/crates/runner/mitm-addon/src/codex_model_catalog_cache.py
+++ b/crates/runner/mitm-addon/src/codex_model_catalog_cache.py
@@ -364,9 +364,19 @@ def _is_responses_path(original_url: str) -> bool:
     return parsed.path == _RESPONSES_PATH or parsed.path.startswith(f"{_RESPONSES_PATH}/")
 
 
+def _strict_utf8_bytes(value: str) -> bytes | None:
+    try:
+        return value.encode()
+    except UnicodeEncodeError:
+        return None
+
+
 def _usable_etag_value(value: str) -> str | None:
     value = value.strip()
-    if not value or value == "*" or len(value.encode()) > _MAX_ETAG_BYTES:
+    if not value or value == "*":
+        return None
+    encoded = _strict_utf8_bytes(value)
+    if encoded is None or len(encoded) > _MAX_ETAG_BYTES:
         return None
     opaque = value[2:] if value.startswith("W/") else value
     if (
@@ -746,8 +756,11 @@ def _response_headers_bypass_reason(
     if encoding != _IDENTITY_ENCODING and not (allow_brotli and encoding == _BROTLI_ENCODING):
         return "response_encoding"
     content_type = headers.get("Content-Type", "")
-    if len(content_type.encode()) > _MAX_CONTENT_TYPE_BYTES or not _content_type_is_json(
-        content_type
+    encoded_content_type = _strict_utf8_bytes(content_type)
+    if (
+        encoded_content_type is None
+        or len(encoded_content_type) > _MAX_CONTENT_TYPE_BYTES
+        or not _content_type_is_json(content_type)
     ):
         return "response_content_type"
     if _response_cache_control_is_unsafe(headers):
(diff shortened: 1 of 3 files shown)


seven332:
## Code Review: PR #32135

**Reviewed head:** `b0aacfb8e3e5e1611ab3ff0364f73ecaee79dc25`

### Summary

The change narrowly converts surrogate-bearing response-header text into the cache's existing unusable-value outcomes. It preserves valid ETag and Content-Type validation, uses the established catalog bypass cleanup path, and leaves authenticated invalid ETag signals without cache or in-flight side effects.

### Findings

No findings.

### Testing

Raw mitmproxy byte headers are covered through `mitm_addon.responseheaders()`. Tests verify catalog response pass-through, deterministic bypass reasons, follower wakeup, global flow-capacity release, a later valid cache hit, entry non-renewal/non-invalidation, and unaffected in-flight owner completion. Local uv lock/sync, Ruff format/lint, BasedPyright, focused catalog suites (76 passed), and the complete addon suite (7296 passed) succeeded. Hosted CI is checked separately by the workflow.

### Verdict

LGTM

PR #32043, opened 2026-09-06
fix(presentation): normalize github app private key

bingjiez666:
## Summary

- normalize the canonical GitHub App secret from flattened or multiline PEM before token creation
- validate the PEM structure and private key before passing escaped newlines to the token action
- replace the Base64-specific regression coverage with flattened-PEM, multiline-PEM, and rejection cases

## Context

- Production run https://github.com/vm0-ai/vm0/actions/runs/34043214072 showed the canonical secret is not Base64.
- Safe diagnostic run https://github.com/vm0-ai/vm0/actions/runs/34043387339 identified a PEM value flattened with whitespace.
- Safe diagnostic run https://github.com/vm0-ai/vm0/actions/runs/34043467552 normalized the key successfully and reached the repository-installation lookup; publishing remained disabled on the diagnostic branch.

## Testing

- `bash .github/scripts/tests/normalize-github-app-private-key-test.sh`
- `bash -n .github/scripts/normalize-github-app-private-key.sh .github/scripts/tests/normalize-github-app-private-key-test.sh`
- `bash .github/scripts/tests/implicit-npm-audit-test.sh`
- `actionlint .github/workflows/publish-presentation-template-archives.yml` (actionlint 1.7.12)
- `git diff --check`

PR #32180, opened 2026-09-07
chore: publish presentation template archives from b4b701a

bingjiez666:
> [!IMPORTANT]
> **This PR is not meant to be merged.** Its Actions run is the publication audit
> record. The repin lands separately in #32160.

Publishes the 23 built-in presentation template archives built from
`vm0-ai/Template-artifact@b4b701ae1b65fb2e6b510b4aa5840eb87e3ed4a6`
(`Template-Presentation/`), superseding the `f60350f` publication of #29414.

Same shape as #29414, with one simplification: the release tooling now lives on
`main` at `.github/scripts/`, so this PR carries only the prebuilt bundle instead
of its own copy of `publish.mjs`.

## Why this path

`Publish Presentation Template Archives` cannot currently reach
`vm0-ai/Template-artifact`: its 5 runs all failed at `Create read token for
Template-artifact` with `Not Found` on
`GET /repos/{owner}/{repo}/installation`, meaning the configured App
authenticated but has no installation on that repository. The most recent
failure ran after #32043 merged, so private-key normalization was already in
place and is not the cause.

Building locally and publishing from a carried bundle keeps the production write
inside CI, behind the `production` environment gate, exactly as #29414 did, and
needs no cross-repository credential at all.

## What changed upstream

The pinned digests predate `Template-artifact#60` (`8e81d41f`, 2026-09-04),
which rewrote 300 files across all 23 template directories. Directory renames in
`#57`/`#58` and the reverse-template revert in `b4b701a` also land in this
publication. The template set itself is unchanged at 23, so
`.github/scripts/presentation-template-release.json` needs no edit.

## Verification

- Reproduced the CI build environment exactly (`tar@7.5.22` beside the copied
  tooling) against a sparse checkout pinned to `b4b701a`, confirmed identical to
  `Template-artifact` `main` at build time.
- `publish-presentation-template-archives.mjs build` reported `VERIFIED build`
  for all 23 packages.
- `publish-presentation-template-archives.mjs verify` re-extracted every carried
  archive and reproduced the same manifest, file count and total size:
  `VERIFIED bundle` for all 23.
- Independently recomputed each version id with `computeVersionId(storageId,
  files)` from `shared.mjs`; all 23 match the build output and the values pinned
  in #32160.
- `actionlint` 1.7.12 and `prettier@3.8.1 --check` pass on the new workflow.

## Jobs

`verify` runs on the PR and re-checks the carried bundle before anything is
uploaded. `publish-production` uses the `production` environment, so it waits for
a deployment approval; it writes immutable R2 objects, inserts one
`storage_versions` row per package, and repoints each storage HEAD.

## After this run

1. Confirm `publish-production` is green and the reported version ids match
   #32160.
2. Merge #32160 to repoint the registry.
3. Close this PR without merging and delete its branch.


bingjiez666:
Publication complete. Run [34081760550](https://github.com/vm0-ai/vm0/actions/runs/34081760550) reported `verify` and `publish-production` both green with 23 `PUBLISHED` lines, and the repin landed in #32160.

The pipeline has since been cut over to the dedicated `okou-ci` App in #32199. Exact-head run [34085931457](https://github.com/vm0-ai/vm0/actions/runs/34085931457) read `Template-artifact@b4b701a` through that App and rebuilt all 23 packages to version ids identical to this publication, so this carried-bundle path is no longer needed. Closing without merging, as intended.

PR #32158, opened 2026-09-07
fix(desktop): serialize update checks

lancy:
## Summary

- replace dependency-owned update scheduling with Desktop-owned Squirrel.Mac feed setup, startup checking, and the existing 30-minute cadence
- route startup, scheduled, and manual checks through one bootstrap-owned single-flight coordinator, with manual requests joining the active result
- preserve manual no-update/error dialogs, background downloads, Computer Use install deferral, and quit/install single-flight behavior
- remove the unused `update-electron-app` dependency after taking ownership of its updater responsibilities

## Ownership proof

- `desktop-auto-updates.ts` contains the only remaining `autoUpdater.checkForUpdates()` call
- the same `requestUpdateCheck` closure starts the initial check, is registered as the interval callback, and backs the controller passed across the bootstrap-to-main bundle contract
- overlapping requests reuse the active check without adding another underlying call or another terminal-listener set
- `update-available`, `update-not-available`, `error`, and synchronous failures release the active state and remove all three per-check listeners before another check can start

## Validation

- `pnpm exec prettier --check` on all changed files
- `pnpm -F @okouai/desktop lint`
- `pnpm -F @okouai/desktop check-types`
- `pnpm -F @okouai/desktop exec vitest run src/desktop-auto-updates.test.ts src/bootstrap-degraded.test.ts src/bootstrap-imports.test.ts src/desktop-update-feed.test.ts src/bootstrap.test.ts` (5 files, 27 tests)
- `pnpm -F @okouai/desktop build:electron` (including bootstrap bundle guard)
- `pnpm knip --workspace @okouai/desktop`

Repository-wide checks are left to the PR pipeline as required.

Fixes #32079

PR #32182, opened 2026-09-07
fix(desktop): hold update checks through native settle

### LINKED ISSUE #32176: bug(desktop): hold update single-flight through native settle
## Background

PR #32158 attempted to serialize startup, interval, and manual Desktop update checks for #32079. Controller acceptance found that its JavaScript single-flight state ends at `update-available`, while the native Squirrel.Mac command is still downloading and remains non-reentrant.

Evidence on `main` commit `a0ad48e446cedf6ca08737f3d096334e980fb178`:

- `turbo/apps/desktop/src/desktop-auto-updates.ts:101-103` treats `update-available` as completion.
- `turbo/apps/desktop/src/desktop-auto-updates.ts:114-120` removes listeners and clears `activeCheck` at that point.
- `turbo/apps/desktop/src/desktop-auto-updates.test.ts:314-337` encodes the same incorrect assumption by immediately starting another check after `update-available`.
- Okou Desktop uses Electron `42.5.1`. [`auto_updater_mac.mm:96-105`](https://github.com/electron/electron/blob/v42.5.1/shell/browser/auto_updater_mac.mm#L96-L105) maps `SQRLUpdaterStateDownloadingUpdate` to the JavaScript `update-available` event.
- Electron 42.5.1 vendors Squirrel.Mac `0e5d146ba13101a1302d59ea6e6e0b3cace4ae38`; its [`checkForUpdatesCommand`](https://github.com/Squirrel/Squirrel.Mac/blob/0e5d146ba13101a1302d59ea6e6e0b3cace4ae38/Squirrel/SQRLUpdater.m#L204-L321) includes the entire `downloadAndPrepareUpdate` operation.
- Electron's own v42.5.1 [`update-race` fixture](https://github.com/electron/electron/blob/v42.5.1/spec/fixtures/auto-update/update-race/index.js#L50-L70) waits after `update-downloaded` before issuing another check because immediate re-entry still hits the RAC command's disabled guard.

The existing mock cannot expose this: it records calls but does not model a native command that remains disabled through download and briefly after the downloaded event.

## Design

Extend the Desktop-owned coordinator so its definition of “in flight” matches Electron/Squirrel rather than the early JavaScript `update-available` notification:

- `update-available` must transition the active operation into a downloading phase; it must not clear ownership or permit another underlying call.
- Requests from startup, the interval, or the menu while checking, downloading, or in the post-terminal native-settle boundary must not invoke `autoUpdater.checkForUpdates()` again.
- `update-downloaded` must participate in coordinator completion as well as the existing install/defer path. Release the native single-flight guard only after the Squirrel command can safely accept another call; use an explicit, testable settle policy grounded in Electron 42.5.1 behavior.
- Audit `update-not-available`, `error`, and synchronous failure transitions for the same event-versus-native-settlement boundary. A queued/coalesced manual request must either receive the result it joined or start exactly once after it is safe.
- Keep listener and timer ownership symmetric and bounded. No request may be lost into a permanently busy state.

### Critical trap

Do not “fix” this by moving completion from `update-available` directly to `update-downloaded` and immediately allowing re-entry in that same callback. Electron's own fixture says the RAC command may still be disabled immediately after `update-downloaded`. The test double must model that delayed native readiness so the old and naive implementations fail deterministically.

## Explicit non-goals

- Reintroducing dependency-owned updater scheduling or a second coordinator.
- Changing feed URLs, update channel, release tags, signing, notarization, artifact publication, or rollback controls.
- Changing the 30-minute cadence.
- Changing the Computer Use deferral/install policy from #26351.
- Changing authentication behavior from #31956.
- Broad Desktop menu or preferences work.
- Modifying the canonical release PR or performing any production release from the implementation thread.

## Rollout

Implement one focused repair PR on current `main`. The open release-please PR #32162 is overlap inventory only; do not modify it or delay a ready repair solely to avoid conflict. The implementation thread owns work through protected-queue merge. Controller acceptance must pass before any release containing #32158 is approved, and the eventual release candidate must include both #32158 and this repair.

## Acceptance criteria

- [ ] A manual or interval request received after `update-available` but before `update-downloaded` produces no second native call and no disabled-command error.
- [ ] A request received in the immediate `update-downloaded` callback/native-settle window produces no unsafe native call.
- [ ] Once Squirrel is safely ready again, at most one queued request starts and later requests continue normally.
- [ ] `update-not-available`, `error`, synchronous failure, download completion, and any settle timer each clean up state and listeners without duplicate dialogs or permanent busy state.
- [ ] A manual request coalesced into an existing operation receives the appropriate existing no-update/error behavior; update-available/download behavior remains non-interruptive.
- [ ] The 30-minute schedule, feed URL, background download, Computer Use deferral, and quit/install single-flight behavior remain unchanged.
- [ ] The test double models native non-reentrancy through download and delayed readiness after `update-downloaded`; it fails the #32158 implementation and passes the repair.
- [ ] Focused Desktop tests, type-check, lint, build/bootstrap guard, and required PR/merge-group checks pass.

## Related work

- #32079
- #32158
- #26351

lancy:
## Summary

- keep startup, 30-minute interval, and manual update requests on the existing shared coordinator throughout checking, download, and native settle
- treat `update-available` as the start of Squirrel.Mac download and release ownership only 1000 ms after `update-downloaded`, matching Electron 42.5.1's own race fixture
- replace the permissive updater mock with a deterministic native-state fake that rejects every call while checking, downloading, or settling

## Root cause

Electron 42.5.1 maps Squirrel.Mac's `SQRLUpdaterStateDownloadingUpdate` to JavaScript `update-available`, so PR #32158 cleared its JavaScript single-flight guard while `checkForUpdatesCommand` was still downloading and non-reentrant. Even `update-downloaded` is emitted from the command's value callback before RACCommand has fully left its executing state; Electron's own fixture waits 1000 ms before retrying.

References:

- [Electron 42.5.1 update state mapping](https://github.com/electron/electron/blob/v42.5.1/shell/browser/auto_updater_mac.mm#L96-L105)
- [Pinned Squirrel.Mac command lifecycle](https://github.com/Squirrel/Squirrel.Mac/blob/0e5d146ba13101a1302d59ea6e6e0b3cace4ae38/Squirrel/SQRLUpdater.m#L204-L321)
- [Electron 42.5.1 post-download settle fixture](https://github.com/electron/electron/blob/v42.5.1/spec/fixtures/auto-update/update-race/index.js#L50-L70)

## Lifecycle design

- `update-available` transitions the active operation from checking to downloading without releasing ownership.
- `update-downloaded` removes the operation's terminal listeners, enters a native-settling phase, and releases the shared guard after the pinned 1000 ms policy.
- Requests from startup, the interval, or the menu coalesce while any of those phases is active; after settle, the first later request starts one native call and subsequent requests join it.
- `update-not-available` and `error` release immediately because Electron emits them from terminal command paths; synchronous throws also release immediately because no native command remains active.
- Errors emitted after download completion remain globally visible without prematurely ending the settle guard.

## Regression proof

Before the production change, the new deterministic regression cases failed for all three owners (`startup`, `interval`, and `manual`) with `The command is disabled and cannot be executed`. The same cases pass after the coordinator holds ownership through download and settle.

## Verification

- `pnpm -F @okouai/desktop exec vitest run src/desktop-auto-updates.test.ts src/bootstrap-degraded.test.ts src/bootstrap-imports.test.ts src/desktop-update-feed.test.ts src/bootstrap.test.ts` — 5 files, 29 tests passed
- `pnpm -F @okouai/desktop lint` — passed
- `pnpm -F @okouai/desktop check-types` — passed
- `pnpm -F @okouai/desktop build:electron` — passed, including bootstrap bundle guard
- `pnpm knip --workspace @okouai/desktop` — passed
- `pnpm exec prettier --check apps/desktop/src/desktop-auto-updates.ts apps/desktop/src/desktop-auto-updates.test.ts apps/desktop/src/bootstrap-degraded.test.ts` — passed
- `git diff --check` — passed

Fixes #32176

### DIFF
--- a/turbo/apps/desktop/src/bootstrap-degraded.test.ts
+++ b/turbo/apps/desktop/src/bootstrap-degraded.test.ts
@@ -41,6 +41,7 @@ const mocks = vi.hoisted(() => {
       >(async () => ({ response: 0 })),
     },
     setInterval: vi.fn(),
+    setTimeout: vi.fn(),
     sentryInit: vi.fn(),
     sentryCaptureException: vi.fn<(error: unknown) => string>(() => "event"),
   };
@@ -48,6 +49,7 @@ const mocks = vi.hoisted(() => {
 
 vi.mock("node:timers", () => ({
   setInterval: mocks.setInterval,
+  setTimeout: mocks.setTimeout,
 }));
 
 vi.mock("electron", () => ({
--- a/turbo/apps/desktop/src/desktop-auto-updates.test.ts
+++ b/turbo/apps/desktop/src/desktop-auto-updates.test.ts
@@ -8,9 +8,17 @@ import type { DesktopAutoUpdatesController } from "./desktop-main-module";
 
 const mocks = vi.hoisted(() => {
   type AutoUpdaterListener = (...args: readonly unknown[]) => void;
+  type NativeUpdateState = "idle" | "checking" | "downloading" | "settling";
+  interface ScheduledNativeSettle {
+    readonly callback: () => void;
+    readonly delay: number | undefined;
+  }
+
   const autoUpdaterListeners = new Map<string, Set<AutoUpdaterListener>>();
   const autoUpdaterOnceListeners = new Map<string, Set<AutoUpdaterListener>>();
   const scheduledUpdateChecks: Array<() => void> = [];
+  const scheduledNativeSettles: ScheduledNativeSettle[] = [];
+  const nativeUpdate = { state: "idle" as NativeUpdateState };
 
   function addAutoUpdaterListener(
     eventName: string,
@@ -38,10 +46,17 @@ const mocks = vi.hoisted(() => {
     autoUpdaterOnceListeners.get(eventName)?.delete(listener);
   }
 
+  function startNativeUpdateCheck(): void {
+    if (nativeUpdate.state !== "idle") {
+      throw new Error("The command is disabled and cannot be executed");
+    }
+    nativeUpdate.state = "checking";
+  }
+
   return {
     app: { isPackaged: true },
     autoUpdater: {
-      checkForUpdates: vi.fn<() => void>(),
+      checkForUpdates: vi.fn(startNativeUpdateCheck),
       quitAndInstall: vi.fn(),
       setFeedURL: vi.fn(),
       on: vi.fn(addAutoUpdaterListener),
@@ -50,12 +65,19 @@ const mocks = vi.hoisted(() => {
     },
     autoUpdaterListeners,
     autoUpdaterOnceListeners,
+    nativeUpdate,
     dialog: {
       showMessageBox: vi.fn<() => Promise<{ response: number }>>(),
     },
+    scheduledNativeSettles,
     scheduledUpdateChecks,
+    startNativeUpdateCheck,
     setInterval: vi.fn((callback: () => void) => {
       scheduledUpdateChecks.push(callback);
+      return 1;
+    }),
+    setTimeout: vi.fn((callback: () => void, delay?: number) => {
+      scheduledNativeSettles.push({ callback, delay });
       return 1;
     }),
   };
@@ -63,6 +85,7 @@ const mocks = vi.hoisted(() => {
 
 vi.mock("node:timers", () => ({
   setInterval: mocks.setInterval,
+  setTimeout: mocks.setTimeout,
 }));
 
 vi.mock("electron", () => ({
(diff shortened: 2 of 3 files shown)


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: f8b8c3c29936e2f7685e50373a9c6271c9a2fed6
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

The exact-head diff repairs the Desktop updater coordinator so `update-available` begins a download phase instead of ending single-flight ownership, and `update-downloaded` holds that ownership through the 1000 ms Electron 42.5.1 RACCommand settle policy. The regression harness models the native updater as genuinely non-reentrant across checking, downloading, and settling, while the existing feed, cadence, Computer Use deferral, and quit/install behavior remain unchanged.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correct lifecycle and race handling | Pass | `desktop-auto-updates.ts` keeps one active operation across checking, download, and settle; equality guards prevent stale callbacks from releasing a newer owner. |
| `REVIEW.md` | Security, type safety, and focused scope | Pass | Three Desktop files change; no auth, feed/channel, signing, release, IPC, persisted data, or permission path changes, and the diff adds no `any` or broad cast. |
| `REVIEW.md` | A fix has regression coverage | Pass | The deterministic fake rejects native re-entry and the three startup/interval/manual cases fail on #32158 behavior but pass on this HEAD; terminal and install paths remain covered. |
| `docs/docs.md` | Load and apply guidance for the changed surface | Pass | Desktop testing plus timer/listener lifecycle guidance was applied; no public product/configuration contract changed, so no separate documentation update is required. |
| `docs/bad-smell.md` | No type/lint suppression, dynamic import, fallback, or defensive error masking | Pass | The production diff uses a typed phase union, a static timer import, and existing user-facing error handling; it adds no suppression or compatibility fallback. |
| `docs/testing.md` | Test through an integration entry point and mock only external boundaries | Pass | Tests call `installDesktopAutoUpdates` and its returned controller with real Desktop modules; only Electron and Node timers are mocked. |
| `docs/testing/patterns.md` | Keep internal implementation real and reset boundary state | Pass | No relative/internal module mock is added; `beforeEach` resets native state, captured timers, listeners, and mock calls. |
| `docs/testing/anti-patterns.md` | Avoid fake timers and mock-only confidence | Pass | No Vitest fake clock is used: the Node timer boundary captures the settle callback, while assertions cover coordinator return values, native busy state, dialogs, listener cleanup, and exact native call count. |
| `docs/testing/desktop-testing.md` | Use the smallest Desktop lifecycle boundary | Pass | Electron is the external runtime fake and the real updater installation/controller wiring is exercised, including degraded bootstrap compatibility. |
| `docs/effect.md` | Choose one owner and trigger for external work | Pass | Bootstrap retains process-lifetime ownership and startup, interval, and menu triggers all enter the same coordinator rather than creating a second updater state machine. |
| `docs/cache.md` | Bound state and tear down acquired listeners/timers symmetrically | Pass | Each check owns four one-shot listeners, removes all four on terminal paths, and uses one self-expiring settle timeout; the active object is released on terminal completion or timeout. |
| `.claude/skills/ccstate/SKILL.md` | ccstate signals, commands, and abort ownership | N/A | No React, ccstate, Store, signal, command, or AbortSignal code is changed. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- **Coverage:** Adequate. The fake reproduces the RAC disabled guard, covers startup/interval/manual ownership during download and settle, proves one safe later call, and covers no-update, download error, synchronous throw, late settle errors, listener cleanup, no duplicate dialogs, and the existing install/defer paths.
- **Conventions:** Compliant. Focused Desktop Vitest passed (5 files, 29 tests), along with Desktop lint, type-check, Electron build/bootstrap guard, workspace Knip, Prettier, and `git diff --check`. Full repository checks are running in the protected PR pipeline.

PR #32184, opened 2026-09-07
fix(annotation): stop showing a failure for an unfinished annotated copy

tongx-blip:
## Problem

The red retry badge came back after #31322, so a flatten failure was never the only way to reach it.

**A draft carrying marks but no annotated copy restores as `failed` — with nothing having been attempted:**

```ts
args.initialAnnotations && args.initialAnnotatedFileId
  ? { status: "uploaded", fileId: args.initialAnnotatedFileId }
  : args.initialAnnotations
    ? { status: "failed" }        // ← nothing ran
    : { status: "idle" };
```

The copy being *missing* is a different thing from making it having *gone wrong*. A retry badge reads as an error the user caused.

**And committing wrote exactly that shape**, because it saved the draft before awaiting the upload:

```ts
const confirmation = confirmAnnotations(next, signal);
onAnnotationChange();                                  // ← saved here, upload still pending
await withCleanup(confirmation, onAnnotationChange);
```

`annotatedFileId$` is null unless the status is `uploaded`, so every commit persisted one snapshot with marks and no copy.

Reproduced before touching anything — a restored draft in that shape renders the badge:

```
AssertionError: expected { retryBadgePresent: true } to strictly equal { retryBadgePresent: false }
```

## Fix

- **Persist once**, after the copy exists or has genuinely failed. The chip already shows the marks from `annotations$`, which `confirmAnnotations$` sets synchronously, so nothing visible waited on that save. A draft written after a real failure still restores as `failed`, which is now accurate.
- **Restoring a missing copy rebuilds it** through `retryAnnotationUpload$` instead of reporting a failure, so the state that describes it is `pending`. The rebuild starts before the prune and is awaited after it, so both run together and `restoreAttachments$` still owns them — no `detach()` in the signals layer.

The annotated copy is what the model actually reads, so regenerating it is also the only way a restored draft still means what the user drew; the old behaviour left the marks pointing at nothing.

## Testing

A case restores a draft with marks and no copy and asserts no retry affordance appears. Verified both ways — it fails against the old `failed` state:

```
× A draft with marks but no annotated copy is not shown as a failure
Tests  1 failed | 5 skipped (6)     ← old state
Tests  6 passed (6)                 ← this branch
```

prettier clean · eslint clean on the changed files · chat-attachment annotation/composer/draft-recovery **16/16**.

Platform `check-types` reports only the 25 pre-existing test-globals errors that are also present on a clean checkout — none in the changed files. `knip` and repo-wide `check-types` excluded locally in this sandbox; CI runs both.

## Note on scope

This is a second, independent cause of the same badge. #31322 fixed a real one (flatten read an unresolved attachment URL); this fixes a state that fabricates the failure outright. If a genuine flatten failure still occurs, the `log.warn` added in #31322 now names which step.

### DIFF
--- a/turbo/apps/platform/src/signals/okou-page/chat-draft.ts
+++ b/turbo/apps/platform/src/signals/okou-page/chat-draft.ts
@@ -353,11 +353,16 @@ function createAttachmentAnnotationSignals(args: {
   const internalAnnotations$ = state<ImageAnnotation | null>(
     args.initialAnnotations ?? null,
   );
+  // Marks without their annotated copy mean the copy is MISSING, not that
+  // making it failed. Reporting `failed` here put a retry badge on a draft
+  // nothing had been attempted for, and the user reads that as an error they
+  // caused. `restoreAttachments$` regenerates the copy instead, so the state
+  // that describes it is `pending`.
   const initialUploadState: AttachmentAnnotationUploadState =
     args.initialAnnotations && args.initialAnnotatedFileId
       ? { status: "uploaded", fileId: args.initialAnnotatedFileId }
       : args.initialAnnotations
-        ? { status: "failed" }
+        ? { status: "pending" }
         : { status: "idle" };
   const internalUploadState$ =
     state<AttachmentAnnotationUploadState>(initialUploadState);
@@ -982,7 +987,7 @@ export function createDraftSignals(): DraftSignals {
 
   const restoreAttachments$ = command(
     async (
-      { set },
+      { get, set },
       persisted: RestorableAttachment[],
       signal: AbortSignal,
     ): Promise<boolean> => {
@@ -993,7 +998,23 @@ export function createDraftSignals(): DraftSignals {
       set(internalAttachments$, (prev) => {
         return [...prev, ...restored];
       });
-      return await set(pruneUnavailableAttachments$, restored, signal);
+      // Rebuild any annotated copy the draft is missing. A draft saved while
+      // its copy was still uploading carries the marks alone, and the copy is
+      // what the model actually reads, so regenerating it is the only way the
+      // restored draft still means what the user drew. Started before the
+      // prune and awaited after it, so the two run together and this command
+      // still owns both.
+      const rebuilt = restored
+        .filter((attachment) => {
+          return get(attachment.annotatedFileId$) === null;
+        })
+        .map((attachment) => {
+          return set(attachment.retryAnnotationUpload$, signal);
+        });
+      const pruned = await set(pruneUnavailableAttachments$, restored, signal);
+      await Promise.all(rebuilt);
+      signal.throwIfAborted();
+      return pruned;
     },
   );
 
--- a/turbo/apps/platform/src/signals/okou-page/chat-draft.ts
+++ b/turbo/apps/platform/src/signals/okou-page/chat-draft.ts
@@ -831,7 +831,17 @@ function createDraftDocumentSignals() {
  * it can never load and every send is rejected, so removing them and saying so
  * is the only state the user can act on.
  */
-function createPruneUnavailableAttachments(
+/**
+ * Makes restored attachments usable: drops the ones whose file is gone, then
+ * rebuilds any annotated copy the draft is missing.
+ *
+ * Both jobs belong together because both restore paths — `seed$` on page load
+ * and `restoreAttachments$` on paste — end here, and an attachment that
+ * reaches either one without the rebuild is stuck. Marks with no copy start as
+ * `pending`, so nothing would move that state and the composer would refuse to
+ * send with no affordance to fix it.
+ */
+function createReconcileRestoredAttachments(
   internalAttachments$: State<ChatAttachment[]>,
 ): Command<Promise<boolean>, [readonly ChatAttachment[], AbortSignal]> {
   return command(
@@ -853,20 +863,38 @@ function createPruneUnavailableAttachments(
       const unavailable = candidates.filter((attachment, index) => {
         return infos[index] === null && currentAttachments.includes(attachment);
       });
-      if (unavailable.length === 0) {
-        return false;
-      }
-      set(internalAttachments$, (prev) => {
-        return prev.filter((attachment) => {
-          return !unavailable.includes(attachment);
+      if (unavailable.length > 0) {
+        set(internalAttachments$, (prev) => {
+          return prev.filter((attachment) => {
+            return !unavailable.includes(attachment);
+          });
         });
-      });
-      reportUnavailableAttachments(
-        unavailable.map((attachment) => {
-          return attachment.filename;
-        }),
+        reportUnavailableAttachments(
+          unavailable.map((attachment) => {
+            return attachment.filename;
+          }),
+        );
+      }
+
+      // The annotated copy is what the model reads, so a restored draft only
+      // still means what the user drew once it exists. Awaited rather than
+      // left running: both callers use the result solely to decide whether to
+      // re-save, and that save should land after the copy exists rather than
+      // persisting the same marks-without-copy shape again.
+      await Promise.all(
+        candidates
+          .filter((attachment) => {
+            return (
+              !unavailable.includes(attachment) &&
+              get(attachment.annotatedFileId$) === null
+            );
+          })
+          .map((attachment) => {
+            return set(attachment.retryAnnotationUpload$, signal);
+          }),
       );
-      return true;
+      signal.throwIfAborted();
+      return unavailable.length > 0;
     },
   );
 }
@@ -877,14 +905,14 @@ function createDraftLifecycleSignals({
   internalGenerationTemplate$,
   internalAttachments$,
   internalDragOver$,
-  pruneUnavailableAttachments$,
+  reconcileRestoredAttachments$,
 }: {
   draftInput: ReturnType<typeof createDraftInputSignals>;
   draftDocument: ReturnType<typeof createDraftDocumentSignals>;
   internalGenerationTemplate$: State<GenerationTemplateRequest | undefined>;
   internalAttachments$: State<ChatAttachment[]>;
   internalDragOver$: State<boolean>;
-  pruneUnavailableAttachments$: Command<
+  reconcileRestoredAttachments$: Command<
     Promise<boolean>,
     [readonly ChatAttachment[], AbortSignal]
   >;
@@ -921,7 +949,11 @@ function createDraftLifecycleSignals({
       ) {
         set(draftDocument.takeRestoredUserMessage$);
       }
-      return await set(pruneUnavailableAttachments$, value.attachments, signal);
+      return await set(
+        reconcileRestoredAttachments$,
+        value.attachments,
+        signal,
+      );
     },
   );
 
@@ -982,12 +1014,12 @@ export function createDraftSignals(): DraftSignals {
     },
   );
 
-  const pruneUnavailableAttachments$ =
-    createPruneUnavailableAttachments(internalAttachments$);
+  const reconcileRestoredAttachments$ =
+    createReconcileRestoredAttachments(internalAttachments$);
 
   const restoreAttachments$ = command(
     async (
-      { get, set },
+      { set },
       persisted: RestorableAttachment[],
       signal: AbortSignal,
     ): Promise<boolean> => {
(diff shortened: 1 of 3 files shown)


tongx-blip:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 8d4b1c62b6888ddd9ee12a40854c86597e9bea23

### Summary

The diagnosis is right and the reproduction is real: a draft carrying marks without an annotated copy did restore as `failed` with nothing attempted, and committing wrote exactly that shape by saving before awaiting the upload. Both halves of that are correctly identified and the commit-side fix is sound.

The restore-side fix is in the wrong place, and as written it makes the primary path worse than the bug it replaces.

`initialUploadState` now returns `pending` for every attachment restored with marks and no copy — that is global, applying to every construction of `createRestoredAttachment`. The rebuild that is supposed to resolve `pending` was added only to `restoreAttachments$`. Three of the four call sites do not go through it, including the chat-thread and agent-chat page loads, which are the paths a user actually lands on. On those, the attachment initialises to `pending`, nothing ever starts the rebuild, `annotationReady$` never becomes true, and `sendReady$` stays false. The result is a permanent spinner, send blocked, and — because the retry badge only renders on `failed` — no affordance at all.

The old behaviour showed a wrong badge that a click could fix. This one shows no badge and cannot be fixed from the UI.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness, edge cases, no API misuse | **Finding (P0)** | `chat-draft.ts:359` changes the state for all four `createRestoredAttachment` call sites; `chat-draft.ts:1001` adds the rebuild for one. `agent-draft.ts:236`, `chat-thread-draft.ts:70` and `create-chat-thread.ts:3442` go through `seed$`, not `restoreAttachments$`. |
| `REVIEW.md` | Feature Regression Guard — adjacent flows covered | **Finding (P0)** | The new case restores through the chat-thread page, so it passes on the `pending` state alone without ever proving a rebuild ran. Nothing covers "send becomes possible again", which is the behaviour that regresses. |
| `REVIEW.md` | `fix:` commits carry a regression test | Pass | The new case fails against the old `failed` state, verified. |
| `REVIEW.md` | Style and maintainability | Finding (P1) | `restoreAttachments$` now awaits `Promise.all(rebuilt)` before returning, so the paste path at `chat-composer.tsx:9000` blocks on a flatten plus upload for every annotated attachment before it resolves. |
| `REVIEW.md` | Security | N/A | No auth, injection or secret-handling surface. |
| `REVIEW.md` | Type safety, no broad `any` | Pass | No `any`, no casts; `{ get, set }` widening is the documented command shape. |
| `REVIEW.md` | DB/JSONB gate for persisted contract changes | N/A | No schema or contract change. The persisted shape is unchanged — only when it is written. |
| `docs/deployment-compatibility.md` | Deployable boundary safe both directions | Pass | Client-only. Drafts written by an older client in the marks-without-copy shape are read by the new client, which is precisely the case being handled. |
| `docs/bad-smell.md` | Fail fast, no fallback slop | Pass | No `??` chain or tolerance branch added. |
| `docs/fallback.md` §4 | Do not fabricate data for a corrupted state | **Pass — this is the fix** | Reporting `failed` for a state nothing had attempted was the fabrication. Removing it is correct; the P0 is only about where the replacement lives. |
| `docs/fallback.md` §6 | Fallbacks limited to required categories | Pass | Introduces none. No `Fallbacks` section required. |
| `docs/effect.md` | Work runs in commands, render stays pure | Pass | The rebuild runs inside `restoreAttachments$`; nothing added to render. |
| `docs/cache.md` RC-8 | Symmetric resource lifecycles, abort handled | Pass | The rebuild is started before the prune and awaited after it, so the command owns both, and `signal.throwIfAborted()` follows the await. No `detach()` in signals. |
| `docs/cache.md` RC-4 | Owner and lifetime match | **Finding (P0)** | The state change is owned by the attachment constructor; its resolution is owned by one caller. The two do not have the same reach, which is the defect. |
| `.claude/skills/ccstate/SKILL.md` | ccstate commands, no React hooks | Pass | No hooks; `set(...)` results collected and awaited within the command. |
| `docs/testing.md` | Mock at boundary, MSW, no fake timers | Pass | Existing helpers, no new mocks or timers. |
| `docs/testing/app-testing.md` | Page-level tests through real setup | Pass | Enters via `setupPage` on the real chat-thread route. |
| `docs/testing/anti-patterns.md` | AP-1 — no mock-call assertions alone | Pass | Asserts on the rendered affordance. |

### Findings

- **`chat-draft.ts:359` vs `chat-draft.ts:1001`** — the state change and its resolution have different reach.

  `initialUploadState` is inside `createAttachmentAnnotationSignals`, so `pending` now applies to every restored attachment:

  ```
  chat-draft.ts:997          restoreAttachments$      → rebuild added
  agent-draft.ts:236         → draft.seed$            → no rebuild
  chat-thread-draft.ts:70    → draft.seed$            → no rebuild
  create-chat-thread.ts:3442 → draft.seed$            → no rebuild
  ```

  `chat-thread-draft.ts` is the chat-thread draft loader — the path behind the reported screenshot. On it, an attachment with marks and no copy is now `pending` with nothing scheduled to move it: `uploadPending$` stays true (spinner forever), `annotationReady$` stays false so `sendReady$` blocks the composer, and `AnnotationUploadRetry` only renders on `failed`, so there is no retry button either. That is a worse failure mode than the badge this PR removes.

  Both `seed$` (`chat-draft.ts:924`) and `restoreAttachments$` end with the same `set(pruneUnavailableAttachments$, attachments, signal)`. Put the rebuild where both already funnel — inside that command, or in a helper both call — so it cannot be reached without it. Then add a case that restores through `seed$` and asserts the composer becomes sendable, not merely that no badge is shown.

- **`chat-draft.ts:1013`** — `await Promise.all(rebuilt)` before returning makes restoring block on the rebuild. `chat-composer.tsx:9000` awaits `restoreAttachments` in the paste handler, so pasting a draft with annotated attachments now waits on a fetch, canvas encode and upload per attachment before it resolves. The attachments are already in `internalAttachments$` and render as pending, so the await buys nothing the user can see. Let the prune result return and keep the rebuild owned without gating the caller on it.

### Testing

- **Coverage: Insufficient for the change as written.** The new case is a genuine regression test for the fabricated `failed` — it fails against the old state, verified. But it restores through the chat-thread page and asserts only the *absence* of the badge, which `pending` satisfies without any rebuild running. It would pass even if the rebuild never fired, which is exactly the P0 condition. The behaviour that needs pinning is that the copy is regenerated and the composer becomes sendable again.
- **Conventions: Compliant.** Real page startup, existing helpers at the boundary, no new mocks or timers, no mock-call assertions, and `waitFor` around the assertion rather than a fixed delay.
- **Note:** local `check-types` reports 25 pre-existing test-globals errors also present on a clean checkout, none in the changed files; that matches the PR body.

tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: e55bd5b7ada3aa2a46632ada0f2ab496386039fc
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Third pass, at `e55bd5b`. Both findings from the previous round are closed.

The P0 is fixed at the right level. `createPruneUnavailableAttachments` became `createReconcileRestoredAttachments` and now does both jobs — drop attachments whose artifact is gone, then rebuild any annotated copy the survivors are missing. `seed$` and `restoreAttachments$` both end at that one command, so an attachment cannot reach either restore path without the rebuild. Verified that all three `seed$` callers (`agent-draft.ts:231`, `chat-thread-draft.ts:73`, `create-chat-thread.ts:3435`) pass `createRestoredAttachment` results only, so the rebuild sees restored attachments and never a fresh in-flight upload.

One detail in that fix is easy to miss and worth calling out as correct: the old `if (unavailable.length === 0) return false` early return had to go, or the rebuild would have been skipped in exactly the common case where nothing needed pruning. The return value is preserved as `unavailable.length > 0`.

The P1 came back with a reasoned disagreement, and the reasoning holds. I asked for the rebuild not to gate the caller; both callers use the result only to decide whether to re-save, and having that save land after the copy exists is the same rule the commit-side half of this PR follows. Awaiting is also what keeps the work in the signal chain, since `detach()` is barred in `signals/`. Nothing visible waits on it — the attachments are in `internalAttachments$` and rendering as pending before the await. Accepting the pushback.

`e55bd5b` also merges the two stacked JSDoc blocks the rename left on that function, where the first had gone stale by describing only the pruning half.

No P0s, no P1s.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness, edge cases, no API misuse | Pass | One command owns both restore jobs and both paths reach it. The `unavailable.length === 0` early return is gone, so the rebuild runs when nothing is pruned. `signal.throwIfAborted()` follows the await. |
| `REVIEW.md` | Feature Regression Guard — adjacent flows covered | Pass | The test now enters through the chat-thread page load, which is the `seed$` path that the previous head left stuck, and pins the composer becoming sendable. `chat-attachment` composer/recovery/previews 24/24 unchanged. |
| `REVIEW.md` | `fix:` commits carry a regression test | Pass | Verified in both directions: with the rebuild removed the new case fails; with it, 6/6. |
| `REVIEW.md` | Security | N/A | No auth, injection or secret-handling surface. |
| `REVIEW.md` | Type safety, no broad `any` | Pass | No `any` or casts. The now-unused `get` was dropped from `restoreAttachments$` rather than left behind. |
| `REVIEW.md` | Style and maintainability | Pass | The rename makes the name match both jobs; the stacked comment is merged in `e55bd5b`. Net structure is simpler than the previous head — the rebuild exists once instead of at a caller. |
| `REVIEW.md` | DB/JSONB gate for persisted contract changes | N/A | No schema or contract change; only *when* the existing shape is written. |
| `docs/deployment-compatibility.md` | Deployable boundary safe both directions | Pass | Client-only. Drafts already persisted in the marks-without-copy shape by an older client are exactly what the rebuild repairs on load, so the old→new direction self-heals rather than needing a tolerance branch. |
| `docs/bad-smell.md` | Fail fast, no fallback slop | Pass | No `??` chain or tolerance branch. A rebuild that genuinely fails sets `failed`, which is now truthful and carries a working retry. |
| `docs/fallback.md` §4 | Do not fabricate data for a corrupted state | **Pass — this is the fix** | Reporting `failed` for a state nothing had attempted was the fabrication; it is gone, and the missing derivative is regenerated instead of guessed at. |
| `docs/fallback.md` §6 | Fallbacks limited to required categories | Pass | Introduces none. No `Fallbacks` section required. |
| `docs/effect.md` | Work runs in commands, render stays pure | Pass | All of it inside `reconcileRestoredAttachments$`; nothing added to render. |
| `docs/cache.md` RC-8 | Symmetric lifecycles, abort handled, no hidden owner | Pass | Awaited inside the command that owns it, `signal` threaded through to `retryAnnotationUpload$`, `throwIfAborted()` after. No `detach()` in signals — the lint rule caught an earlier attempt and the await is the sanctioned form. |
| `docs/cache.md` RC-4 | Owner and lifetime match | Pass | This is what changed: the state is set in the attachment constructor and resolved at the single point every restore funnels through, so the two now have the same reach. |
| `.claude/skills/ccstate/SKILL.md` | ccstate commands, no React hooks | Pass | `set(...)` results collected and awaited within one command; no hooks. |
| `docs/testing.md` | Mock at boundary, MSW, no fake timers | Pass | Existing `context.mocks.http` / `browser.canvasRendering` / `upload.success` helpers; no new mocks or timers. |
| `docs/testing/app-testing.md` | Page-level tests through real setup | Pass | `setupPage` on the real chat-thread route. |
| `docs/testing/anti-patterns.md` | AP-1 — no mock-call assertions alone | Pass | Leads with user-visible outcomes (mark count, Send enabled); `imageReads` corroborates that the copy was actually rebuilt rather than assumed. |
| `CLAUDE.md` | YAGNI — no speculative parameters or helpers | Pass | No new helper or parameter; an existing command absorbed the second job. |

### Findings

None.

### Testing

- **Coverage: Adequate.** 6/6 in `chat-attachment-annotation.test.tsx`, 24/24 across it plus composer, draft-recovery and message-previews. The gap from the last round is closed: the case restores through the chat-thread page load (the `seed$` path) and asserts the mark count renders, Send becomes enabled, and the original was read exactly once — so it fails on a permanently pending attachment, which is what the previous head produced. Confirmed by removing the rebuild and watching it fail.
- **Conventions: Compliant.** Real page startup, MSW at the boundary, no fake timers, no internal `vi.mock`, no mock-call assertion standing alone.
- **Checks:** prettier clean, eslint clean on the changed files. Platform `check-types` reports 25 errors — the same count and set as a clean checkout, none in the changed files. `knip` and repo-wide `check-types` were excluded locally in this sandbox; CI runs both.

PR #32188, opened 2026-09-07
fix(runner): expose api transport failure causes

### LINKED ISSUE #32100: bug(runner): expose safe root causes for runner-to-API transport failures
## Problem and evidence

Runner requests to four API endpoints intermittently failed before receiving a usable HTTP response.

### Production evidence

- Window: Beijing `2026-09-04 00:00` through `2026-09-07 00:00` Asia/Shanghai (half-open); UTC `2026-09-03T16:00:00Z` through `2026-09-06T16:00:00Z` (half-open).
- Source: Axiom dataset `vm0-web-logs-prod`, predicate `level == "error" or level == "warn"`.
- Raw events: 24.
  - Telemetry flush: 14 events, runner versions 0.184.14 through 0.186.9.
  - Connector runtime sync: 5 events.
  - Active-input reserve read: 4 events.
  - Heartbeat: 1 event.
- The shared rendered diagnostic was `send API request failed: error sending request`.

The common wrapper proves that these calls failed in the Runner HTTP transport path, but it does not prove that all four variants share one network root cause.

### Current code

Analyzed `origin/main` at `266467a02c0b386407b67441b32e6830be07ffbc`.

The shared HTTP client already records endpoint, method, host, path, client request ID, client session ID, client version, and the coarse `timeout`, `connect`, `request`, `body`, or `unknown` failure kind. HTTP response failures are already represented separately with their status. This behavior was introduced by #20648.

The remaining attribution gap is narrower: a `request` or `unknown` transport failure can still retain only the top-level `error sending request` summary, which does not distinguish actionable causes such as DNS resolution, TLS negotiation, connection reset, or HTTP protocol failure when the underlying stack exposes that information.

The callers also have different existing recovery semantics:

- Connector runtime sync performs one immediate transport retry, retains last-known-good state, and schedules capped exponential retries with jitter.
- Heartbeat sends a fresh state on its next 10-second tick and reports sustained timeout/connect degradation and recovery.
- Active input retains pending input and retries source reads; its API read backoff is tracked separately in #32170.
- Runner telemetry is best-effort. Retrying an ambiguously completed request without an idempotency contract could duplicate sandbox-operation observations.

Relevant code:

- `crates/runner/src/http.rs`
- `crates/runner/src/error.rs`
- `crates/runner/src/provider/api.rs`
- `crates/runner/src/provider/connector_runtime_sync.rs`
- `crates/runner/src/executor/active_input.rs`
- `crates/runner/src/telemetry.rs`

## Decision

Narrow this issue to safe root-cause attribution in the shared Runner-to-API HTTP transport. Do not add shared automatic retries or change caller-specific delivery semantics in this issue.

Preserve the existing structured request context and coarse failure kind, while exposing an additional bounded, stable, and safe cause when the transport stack provides one. The diagnostic must remain useful across all Runner API callers without copying transport inspection into each caller.

This boundary addresses the confirmed observability gap first. More precise production evidence can then justify endpoint-specific reliability changes without assuming that unrelated requests are safe to retry in the same way.

## Alternatives considered

- **Retry every transport failure in the shared client:** rejected because endpoint side effects and ambiguous completion differ. A request may reach the API even when the Runner does not receive the response.
- **Bundle recovery changes for all four callers:** rejected because Connector, heartbeat, active input, and telemetry already have distinct loss, retry, freshness, and idempotency contracts.
- **Keep the current summary:** rejected because `failure_kind=request` plus `error sending request` is not sufficient to identify or route the underlying operational problem.
- **Log the complete debug/source chain:** rejected unless it can be converted to an allowlisted safe representation; raw dependency errors can expose URLs or create unstable, high-cardinality fields.

## Risks and mitigations

- **Sensitive data exposure:** never include authorization headers, request bodies, URL query values, credentials, or unbounded raw source-chain text. Retain the existing redaction tests and add coverage for every new field.
- **Unstable classifications:** use a small stable taxonomy owned by the Runner rather than dependency-specific display strings.
- **High-cardinality telemetry:** keep cause values bounded; retain request IDs only in the existing correlation fields.
- **False attribution:** represent unavailable detail as unknown rather than inferring a DNS, TLS, or protocol cause without evidence.
- **Compatibility:** this is diagnostic-only Runner behavior; do not change API schemas, request payloads, timeouts, retry cadence, or persisted state.

## Acceptance criteria

- Shared `ApiTransportError` diagnostics provide a bounded stable cause beyond the current generic summary for representative transport failures where the underlying stack exposes the cause.
- Existing endpoint, method, host/path, request/session ID, client version, and coarse failure-kind attribution remains available.
- HTTP status failures remain distinct from pre-response transport failures.
- Timeout, connection refusal, connection termination, and other feasible transport cases are exercised through real local socket/HTTP integration boundaries rather than mocked internal functions.
- Tests prove that URL query values, credentials, authorization tokens, and request bodies cannot enter the new diagnostic fields.
- Connector runtime, heartbeat, active-input, and telemetry retry/delivery behavior remains unchanged.
- Runner formatting, tests, Clippy, and documentation checks pass for the affected crate.

## Delivery notes

This is one standalone Runner PR and is not a delivery parent.

- Active-input API read backoff is tracked independently in #32170.
- Reliable telemetry delivery or server-side deduplication is intentionally deferred until its idempotency and product requirements are established.
- Current production logs after the evidence window were not re-queried because the local environment did not expose a production Axiom token.

## Ownership

Selected owner: @seven332. Confidence: high.

seven332:
## Summary

- classify Runner-to-API transport failures with a bounded Runner-owned cause derived only from public reqwest, Hyper, and standard I/O types
- add `failure_cause` to heartbeat, connector runtime sync, active-input, telemetry, and existing provider transport diagnostics
- cover timeout, connection refusal, premature HTTP disconnect, invalid TLS peers, caller behavior, and diagnostic redaction through real local transport boundaries

- keeps `ApiFailureKind` as the sole retry-policy input and does not change timeouts, retry cadence, delivery, or fallback behavior
- does not change API schemas, request payloads, queues, Guest protocols, persisted state, or rollout requirements
- never inspects or emits raw dependency source/debug strings; unsupported detail, including TLS or DNS without a stable public signal, remains generic or unknown

## Validation

- `RUSTDOCFLAGS='-D warnings' cargo doc --manifest-path crates/Cargo.toml --profile local -p runner --all-features --no-deps`
- `cargo test --manifest-path crates/Cargo.toml --profile local -p runner --all-features -- --test-threads=4` (3508 passed, 25 ignored)

Closes #32100

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -2634,6 +2634,7 @@ dependencies = [
  "hex",
  "http-body 1.1.0",
  "httpmock",
+ "hyper 1.11.1",
  "idna",
  "libc",
  "nbd-cow",
--- a/crates/Cargo.toml
+++ b/crates/Cargo.toml
@@ -73,6 +73,7 @@ futures-util = "0.3"
 hex = "0.4"
 hmac = "0.13.0"
 httpmock = "0.8"
+hyper = "1"
 idna = "1"
 libc = "0.2"
 nix = { version = "0.31", default-features = false }
--- a/crates/runner/Cargo.toml
+++ b/crates/runner/Cargo.toml
@@ -32,6 +32,7 @@ tracing-appender = { workspace = true }
 tracing-subscriber = { workspace = true }
 flate2 = { workspace = true }
 futures-util = { workspace = true }
+hyper = { workspace = true }
 reqwest = { workspace = true }
 rustix = { workspace = true, features = ["process", "time"] }
 hex = { workspace = true }
--- a/crates/runner/src/error.rs
+++ b/crates/runner/src/error.rs
@@ -170,10 +170,56 @@ impl ApiFailureKind {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ApiTransportCause {
+    Timeout,
+    ConnectionRefused,
+    ConnectionReset,
+    ConnectionAborted,
+    NetworkUnreachable,
+    HostUnreachable,
+    NotConnected,
+    BrokenPipe,
+    UnexpectedEof,
+    HttpIncompleteMessage,
+    HttpCanceled,
+    HttpClosed,
+    HttpParse,
+    HttpBodyWriteAborted,
+    HttpShutdown,
+    Io,
+    Unknown,
+}
+
+impl ApiTransportCause {
+    pub fn as_str(self) -> &'static str {
+        match self {
+            Self::Timeout => "timeout",
+            Self::ConnectionRefused => "connection_refused",
+            Self::ConnectionReset => "connection_reset",
+            Self::ConnectionAborted => "connection_aborted",
+            Self::NetworkUnreachable => "network_unreachable",
+            Self::HostUnreachable => "host_unreachable",
+            Self::NotConnected => "not_connected",
+            Self::BrokenPipe => "broken_pipe",
+            Self::UnexpectedEof => "unexpected_eof",
+            Self::HttpIncompleteMessage => "http_incomplete_message",
+            Self::HttpCanceled => "http_canceled",
+            Self::HttpClosed => "http_closed",
+            Self::HttpParse => "http_parse",
+            Self::HttpBodyWriteAborted => "http_body_write_aborted",
+            Self::HttpShutdown => "http_shutdown",
+            Self::Io => "io",
+            Self::Unknown => "unknown",
+        }
+    }
+}
+
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct ApiTransportError {
     pub request: ApiRequestContext,
     pub failure_kind: ApiFailureKind,
+    pub failure_cause: ApiTransportCause,
     pub summary: String,
 }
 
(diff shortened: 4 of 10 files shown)


seven332:
## Code Review: PR #32188

**Reviewed head:** `613d5524e56466d5066f8b76cf2d92e1ae1932da`

### Summary

The PR adds a bounded, Runner-owned transport-cause taxonomy without changing the existing failure kind or caller recovery decisions. It derives values only from public typed error surfaces, keeps unsupported causes conservative, and exposes the new field across the affected caller diagnostics.

### Findings

No findings. The final head also preserves the prior outer error rendering and removes the refused-port test race found during self-review.

### Testing

Real local transport boundaries cover timeout, connection refusal, incomplete HTTP response, invalid TLS peer fallback, redaction, and the heartbeat, Connector, active-input, and telemetry paths. Runner fmt, all-target/all-feature Clippy, rustdoc, commit hooks, and the full Runner suite passed; focused affected tests passed again on the current head. Production traffic replay was not possible because historical nested source chains were not retained.

### Verdict

LGTM

PR #32234, opened 2026-09-07
test(runner): synchronize network-log child cleanup

### LINKED ISSUE #32224: test(runner): synchronize network-log EOF child-reap assertions
## Problem and evidence

Runner CI intermittently fails the network-log shutdown tests while the owned helper child is still being reaped asynchronously.

Two independent coverage runs have exposed the same race for both required network-log components:

- workflow run `34083513950`, job `101623244341`: `kmsg_stdout_eof_stops_runner_and_reaps_child_before_teardown`
- a later reported failure: `dns_stderr_eof_stops_runner_and_reaps_child_before_teardown`, where `/proc` still showed the same DNS child PID and start time at the assertion

The current `main` failure also reproduces locally. Repeating the exact DNS test failed on iteration 159:

```text
assertion `left != right` failed: dns child pid 3821019 with start time 8660988 was not reaped
  left: Some(8660988)
 right: Some(8660988)
```

`assert_network_log_eof_stops_runner` blocks the final heartbeat, triggers EOF, waits for that heartbeat to become in flight, and then immediately checks `/proc`. `assert_network_log_read_error_stops_runner` makes the same assumption.

PR #32172 intentionally changed network-log failure handling so the reactor transfers the child to an independently scheduled reaper before lifecycle publication continues. Reaching the final heartbeat therefore proves that teardown started, but it is not a child-cleanup completion signal. The test can observe valid concurrent progress and fail before the reaper receives enough scheduler time. This is a test synchronization defect; the evidence does not show a production child leak.

## Decision

Proceed as one standalone Runner test-reliability PR. Cover both DNS and kmsg and both EOF and read-error paths because they share the same helper and invalid ordering assumption.

Keep production lifecycle behavior unchanged. The test must continue to verify real process identity and bounded eventual reaping while the blocked heartbeat keeps unrelated teardown from completing.

## Chosen direction

- Generalize the existing bounded waiter used for abnormal child cleanup so it can identify any asynchronously owned component.
- Use that waiter in the shared EOF and read-error helpers while the final heartbeat remains blocked.
- Reuse it in the existing abnormal DNS and MITM cleanup cases.
- Preserve immediate child-removal assertions after `run()` or an explicit cleanup owner has already completed.

This observes the external contract directly: the original PID plus `/proc` start time must disappear within the existing two-second bound. It adds no fixed delay and exits as soon as cleanup completes.

### Add a reaper-completion test hook

Rejected. Extending the test-only `ReapGate` through shared lifecycle types would add a second completion contract while the external `/proc` outcome still needs verification.

### Check only after `run()` returns

Rejected. Normal teardown joins cleanup, so this would be stable but would stop proving that independently owned cleanup progresses before unrelated teardown completes.

### Restore synchronous production reaping

Rejected. It would reintroduce the lifecycle stall that #32172 intentionally removed merely to satisfy the old test ordering.

## Risks and mitigations

- **A real leak could look like a scheduling delay:** the wait remains strictly bounded and fails against the same PID/start-time identity after two seconds.
- **Assertions could be weakened too broadly:** only independently scheduled paths use the waiter; joined-cleanup paths retain immediate assertions.
- **Polling could add test cost:** the helper yields only while the real child remains present and normally exits immediately after cleanup; there is no production cost.
- **A larger timeout could mask the race:** no timeout is increased and no fixed sleep is added. Repeated exact execution validates scheduler variation.

## Acceptance criteria

- EOF and read-error shutdown coverage for both DNS and kmsg waits boundedly for the independently owned child cleanup before asserting removal.
- The tests still prove that component failure stops discovery, enters teardown, and returns the expected terminal error.
- Normal shutdown and explicit joined-cleanup tests retain their immediate child-reaped assertions.
- The exact DNS and kmsg EOF/read-error regressions remain stable under repeated execution.
- Runner formatting, Clippy, rustdoc, focused tests, and the full affected-crate test suite pass without new skips.

## Delivery notes

- No API, runner protocol, persisted state, configuration, deployment-order, or production behavior changes.
- This is one independently reviewable PR and is independent of the broader production cleanup work in #32040.
- Rollback is a normal test-only revert.

seven332:
## Summary

- wait boundedly for independently owned DNS and kmsg child cleanup in EOF and read-error shutdown tests
- keep the final heartbeat blocked while verifying real PID/start-time removal
- retain immediate child-reaped assertions after explicitly joined cleanup

## Scope

This is a test-only synchronization change. It does not change Runner production behavior, cleanup ownership, shutdown ordering, timeouts, APIs, protocols, persisted state, configuration, or deployment compatibility.

## Validation

- Reproduced the pre-fix DNS EOF failure on iteration 159 of repeated exact execution.
- Post-fix repeated exact execution: 300 passes each for DNS EOF, DNS read error, kmsg EOF, and kmsg read error (1200 total).
- `cargo test --manifest-path crates/Cargo.toml --profile local -p runner --bin runner cmd::start::tests::main_loop::shutdown` — 20 passed
- `RUSTDOCFLAGS="-D warnings" cargo doc --manifest-path crates/Cargo.toml --profile local -p runner --all-features --no-deps`
- `cargo test --manifest-path crates/Cargo.toml --profile local -p runner --all-targets --all-features` — 3512 passed, 25 pre-existing ignored, plus 1 guest-inventory integration test passed
- Pre-commit Cargo formatting, Clippy, rustdoc, file-size, and commit-message hooks passed

Closes #32224

### DIFF
--- a/crates/runner/src/cmd/start/tests/main_loop/shutdown.rs
+++ b/crates/runner/src/cmd/start/tests/main_loop/shutdown.rs
@@ -178,7 +178,7 @@ async fn assert_network_log_eof_stops_runner(component: NetworkLogTestComponent)
         !run_handle.is_finished(),
         "blocked final heartbeat should hold teardown open",
     );
-    assert_child_reaped(component.label(), pid, starttime).await;
+    wait_for_child_cleanup(component.label(), pid, starttime).await;
     assert!(
         env.cancel.is_cancelled(),
         "{} EOF should stop discovery",
@@ -206,7 +206,7 @@ async fn assert_network_log_read_error_stops_runner(component: NetworkLogTestCom
         "{} read error should drive runner teardown",
         component.label(),
     );
-    assert_child_reaped(component.label(), pid, starttime).await;
+    wait_for_child_cleanup(component.label(), pid, starttime).await;
     assert!(
         env.cancel.is_cancelled(),
         "{} read error should stop discovery",
@@ -386,7 +386,7 @@ async fn cancelled_reactor_does_not_abort_owned_network_log_child_cleanup() {
     run_handle.abort();
     assert!(run_handle.await.unwrap_err().is_cancelled());
     gate.release.add_permits(1);
-    wait_for_abnormal_child_cleanup(pid, starttime).await;
+    wait_for_child_cleanup("dns", pid, starttime).await;
 }
 
 #[tokio::test]
@@ -474,7 +474,7 @@ async fn mitm_recovery_panic_stops_runner_instead_of_retrying_unknown_cleanup() 
     assert_run_error_contains(run_handle, "mitmproxy recovery task failed").await;
     assert!(env.cancel.is_cancelled());
     assert!(crash_tx.is_closed());
-    wait_for_abnormal_child_cleanup(pid, starttime).await;
+    wait_for_child_cleanup("mitmdump", pid, starttime).await;
 }
 
 #[tokio::test]
@@ -793,9 +793,9 @@ async fn assert_child_reaped(component: &str, pid: u32, starttime: u64) {
     );
 }
 
-// Abnormal task termination transfers reaping to the Drop fallback. Unlike
-// normal shutdown's join, that fallback promises eventual process removal.
-async fn wait_for_abnormal_child_cleanup(pid: u32, starttime: u64) {
+// Independently owned cleanup promises eventual process removal. Joined
+// cleanup paths use assert_child_reaped to enforce the stronger postcondition.
+async fn wait_for_child_cleanup(component: &str, pid: u32, starttime: u64) {
     tokio::time::timeout(Duration::from_secs(2), async {
         while crate::process::read_process_stat(pid)
             .await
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #32234

**Reviewed head:** `c81ea4110efb56f09905c5ae6c1f2b8431b10789`

### Summary

Reviewed both commits, the submitted diff, the required-monitor reactor path, independent `ChildReaper` ownership, joined cleanup paths, and the issue's challenged plan. The change correctly limits bounded outcome waiting to asynchronously owned cleanup while preserving immediate assertions after cleanup joins.

### Findings

No findings. Self-review caught and fixed one diagnostic regression before this review: a genuine timeout now retains the component, PID, `/proc` start time, and bound.

### Testing

- The pre-fix DNS EOF race reproduced on iteration 159.
- All four DNS/kmsg EOF/read-error tests passed 300 repeated exact runs each (1200 total) after the synchronization change.
- Final-head shutdown module: 20 passed.
- Final-head Runner suite: 3512 passed, 25 pre-existing ignored; guest inventory: 1 passed.
- Final-head Rust format, all-target/all-feature Clippy, and rustdoc passed with warnings denied; commit hooks passed.
- CI coverage instrumentation and the repository-wide CI matrix remain external checks. No production or hardware smoke test is needed for this test-only change.

### Verdict

LGTM for this head. CI and review-feedback status are checked separately; this is not merge approval.

PR #32243, opened 2026-09-07
test(e2e): align zendesk probe with firewall auth

### LINKED ISSUE #32235: test(e2e): eliminate terminal-run race in Zendesk firewall probe
## Background

Runner E2E job [101640867216](https://github.com/vm0-ai/vm0/actions/runs/34089211337/job/101640867216) failed in `tests/03-runner/run-t03-firewall-zendesk.bats` while validating the dynamic Zendesk firewall base and header credential injection. The merge-group head was `419adf61a331e95c533b154ad4c70c5fa7458627`; PR #32188 changed Runner-to-API transport diagnostics and did not change this E2E, the mitm add-on, or firewall-auth behavior.

## Problem and evidence

The probe sends a request to a fresh real Zendesk hostname with `curl --max-time 5`, treats curl transport failure as non-fatal, completes the agent run, and then waits for an `ALLOW` network event containing the exact resolved secret. The mitm add-on's internal firewall-auth fetch has a documented 10-second deadline, so the test client can intentionally end before authorization reaches a definitive result.

In the failed run:

- curl exited with status `28` after its five-second deadline.
- The firewall base matched `https://e2e2436728647.zendesk.com`, but the eventual network event was `BLOCK` with `firewall_error=FORBIDDEN`, empty permission/rule fields, and `error="peer closed connection"`.
- Axiom sandbox-operation telemetry showed `firewall_auth_prepare` succeeded after 3,481.7 ms and `firewall_auth_resolve` succeeded after 0.6 ms.
- The run completed at `2026-09-07T06:13:13.302Z`.
- The final locked `firewall_auth_admit` check ran at `2026-09-07T06:13:13.981Z` and correctly failed because the run was already terminal.
- `runner_e2e_wait_for_firewall_log` then polled for an `ALLOW` event that could no longer occur and failed after 90 seconds.

This is test-induced nondeterminism. It became possible after #31993 replaced the unrelated forced `8.8.8.8` sink with the real Zendesk authority. The API must continue rejecting credential resolution after a run reaches a terminal state.

## Decision

Keep the change E2E-only. Make the one Zendesk probe request live longer than the firewall-auth operation's documented bound, and make the existing ALLOW waiter stop immediately on a matching definitive firewall-auth error. Do not change production admission or run lifecycle semantics.

- Change the Zendesk curl total deadline from 5 seconds to 15 seconds. This bounded protocol alignment gives the internal 10-second auth fetch room to admit or fail before the guest shell completes. It is not a requirement for Zendesk to answer successfully; curl status remains diagnostic and non-fatal.
- Preserve the fresh templated authority, IPv4 path, direct upstream binding, and exact `ZENDESK_API_TOKEN` ALLOW assertion.
- Update `runner_e2e_wait_for_firewall_log` to check the existing success predicate first, then immediately report a matching connector/host record with a non-null `firewall_error`. Continue polling when telemetry is absent, delayed, or unrelated. Leave the 90-second observation ceiling unchanged.
- Keep `runner_e2e_wait_for_firewall_transition` unchanged for tests that intentionally expect a denial followed by an allow.

## Alternatives considered

- A test-controlled checkpoint release would also require live telemetry publication or another guest/proxy signal because network logs are uploaded after provider completion. It adds mock guest, runner, cleanup, and cancellation machinery for one consumer.
- Delaying production run completion for in-flight proxy auth could let abandoned requests delay user-visible completion and runner capacity release, while weakening the clear terminal credential boundary through new cross-process coordination.
- Replacing Zendesk with a controlled custom connector would duplicate existing custom-connector E2E coverage and stop validating the deployed built-in Zendesk catalog metadata and #31993's direct-binding path.
- Restoring the `8.8.8.8` sink is ruled out by the upstream-binding failure diagnosed in #31992 and fixed by #31993.

## Risks and mitigations

- A slow third party can hold this one request up to ten seconds longer. The request remains bounded at 15 seconds and the Runner shard cap is unchanged.
- The helper could reject a recoverable sequence. It will evaluate a valid ALLOW first and require the exact firewall name and host; intentional transition tests use a separate helper.
- The test's 15-second bound is coupled to the current 10-second internal auth deadline. The relationship will be documented beside the request.

## Acceptance criteria

- The Zendesk firewall E2E's guest request remains live beyond the current firewall-auth fetch deadline, so the required credential-resolution attempt reaches a definitive result before the run can complete.
- The test still verifies a fresh templated Zendesk authority and an `ALLOW` event with exactly `ZENDESK_API_TOKEN` resolved.
- The test does not require a successful Zendesk application response.
- A matching definitive credential-resolution error surfaces immediately instead of waiting 90 seconds for an impossible `ALLOW` event.
- The API's final active-run admission check and the Runner's fail-closed credential boundary remain unchanged.
- The fix adds no retries, sleeps, skipped tests, or larger Bats/helper assertion timeout. The single curl deadline may be protocol-aligned as described above.
- Relevant Bats and shell validation pass, and the provisioned Runner E2E CI shard is green.

## Relevant files

- `e2e/tests/03-runner/run-t03-firewall-zendesk.bats`
- `e2e/helpers/runner-api.bash`
- `crates/runner/mitm-addon/src/firewall_auth_client.py`
- `turbo/apps/api/src/signals/services/agent-webhook-firewall-auth.service.ts`

## Related work

- #31992 diagnosed the former forced-public-sink upstream-binding race.
- #31993 removed that forced destination but retained the real third-party request and five-second curl budget.

## Delivery notes

This remains one independently reviewable PR with no dependencies, migrations, feature flag, rollout sequence, or follow-up slice.

seven332:
## Summary

- keep the Zendesk guest request alive beyond the proxy's bounded firewall-auth operation so the run cannot finish first
- fail the firewall ALLOW waiter immediately when the expected connector and host already report a definitive firewall error
- preserve the fresh Zendesk authority, exact secret assertion, direct upstream binding, and non-fatal third-party response outcome

Production API admission, runner lifecycle, connector metadata, the 90-second telemetry observation ceiling, and transition-specific firewall coverage are unchanged.

Closes #32235

## Testing

- `cd e2e && ./test/libs/bats/bin/bats --count tests/03-runner/run-t03-firewall-zendesk.bats`
- `bash -n e2e/helpers/runner-api.bash`
- `shellcheck -e SC2034 e2e/helpers/runner-api.bash` (`TEST_ID` is consumed by callers after this helper is sourced)
- focused shell behavior check for ALLOW success, matching-error failure including `ALLOW + firewall_error`, and ALLOW-first ordering
- `./scripts/check-file-size.sh e2e/helpers/runner-api.bash e2e/tests/03-runner/run-t03-firewall-zendesk.bats`
- `git diff --check`
- `lefthook run pre-commit --file e2e/helpers/runner-api.bash --file e2e/tests/03-runner/run-t03-firewall-zendesk.bats`

The full `03-runner` integration is CI-only because it requires the PR preview, temporary connector account, provisioned runner fleet, guest, and mitm add-on. The PR pipeline is the authoritative end-to-end validation.

### DIFF
--- a/e2e/helpers/runner-api.bash
+++ b/e2e/helpers/runner-api.bash
@@ -533,12 +533,11 @@ runner_e2e_wait_for_firewall_log() {
                     | select(
                         .firewall_name == $firewallName and
                         .host == $host and
-                        .action == "BLOCK" and
                         (.firewall_error // null) != null
                     ) ]
                     | select(length > 0)' \
                 <<<"$last_logs"); then
-                echo "Firewall ${firewall_name@Q} on ${host@Q} reported an authentication error for run ${run_id}" >&2
+                echo "Firewall ${firewall_name@Q} on ${host@Q} reported an error for run ${run_id}" >&2
                 echo "Matching network telemetry: ${failed_logs}" >&2
                 echo "Last network telemetry: ${last_logs}" >&2
                 return 1
--- a/e2e/helpers/runner-api.bash
+++ b/e2e/helpers/runner-api.bash
@@ -504,10 +504,11 @@ runner_e2e_wait_for_firewall_log() {
     local timeout_seconds="${6:-90}"
     local started_at=$SECONDS
     local last_logs='[]'
+    local failed_logs='[]'
 
     while ((SECONDS - started_at < timeout_seconds)); do
-        if last_logs=$(runner_e2e_network_logs "$run_id" 2>&1) &&
-            jq -e \
+        if last_logs=$(runner_e2e_network_logs "$run_id" 2>&1); then
+            if jq -e \
                 --arg firewallName "$firewall_name" \
                 --arg host "$host" \
                 --arg expectedUrlRewrite "$expected_url_rewrite" \
@@ -521,8 +522,27 @@ runner_e2e_wait_for_firewall_log() {
                     ($expectedUrlRewrite == "ignore" or
                         .auth_url_rewrite == ($expectedUrlRewrite == "true")))' \
                 <<<"$last_logs" >/dev/null; then
-            printf '%s\n' "$last_logs"
-            return 0
+                printf '%s\n' "$last_logs"
+                return 0
+            fi
+
+            if failed_logs=$(jq -ce \
+                --arg firewallName "$firewall_name" \
+                --arg host "$host" '
+                    [ .[]
+                    | select(
+                        .firewall_name == $firewallName and
+                        .host == $host and
+                        .action == "BLOCK" and
+                        (.firewall_error // null) != null
+                    ) ]
+                    | select(length > 0)' \
+                <<<"$last_logs"); then
+                echo "Firewall ${firewall_name@Q} on ${host@Q} reported an authentication error for run ${run_id}" >&2
+                echo "Matching network telemetry: ${failed_logs}" >&2
+                echo "Last network telemetry: ${last_logs}" >&2
+                return 1
+            fi
         fi
         sleep 2
     done
--- a/e2e/tests/03-runner/run-t03-firewall-zendesk.bats
+++ b/e2e/tests/03-runner/run-t03-firewall-zendesk.bats
@@ -39,11 +39,11 @@ set -euo pipefail
 printf 'ZENDESK_API_TOKEN=%s\n' "$ZENDESK_API_TOKEN"
 printf 'ZENDESK_EMAIL=%s\n' "$ZENDESK_EMAIL"
 printf 'ZENDESK_SUBDOMAIN=%s\n' "$ZENDESK_SUBDOMAIN"
-# A Zendesk response is outside this test's contract. Use the real templated
-# authority so credential injection follows the normal upstream binding path,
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #32243

**Reviewed head:** `916004265531fa91ba134e1644e280e7673da4ca`

### Summary

The E2E-only change keeps the Zendesk request alive beyond the proxy's existing firewall-auth deadline and makes the ALLOW waiter fail immediately when the exact firewall and host already report a non-null `firewall_error`. It preserves the real templated authority, direct upstream binding, exact secret assertion, and production terminal-run admission behavior.

### Findings

No findings. Self-review found and fixed the initial action-specific error predicate; the current head correctly handles post-decision errors that can coexist with either an ALLOW or BLOCK action while still evaluating a valid error-free ALLOW first.

### Testing

Focused Bats parsing, Bash parsing, ShellCheck, success/error/ordering behavior checks, file-size validation, diff validation, and scoped pre-commit hooks passed. The provisioned `03-runner` integration is CI-only and remains the authoritative end-to-end gate.

### Verdict

LGTM

PR #32244, opened 2026-09-07
test(python): cover fail-closed sse discard paths

### LINKED ISSUE #32233: test(python): cover fail-closed model SSE discard paths
## Problem

At `c3c8778e1352786e8c21d868531656f79a9624a7`, the Python test suite does not execute the provider-level failure-observer path for an in-progress SSE event discarded by the bounded framing scanner.

`src/usage/sse.py:284-295` discards a captured event when a malformed control line exceeds the 4,096-byte bound and invokes `on_event_discard()`. The provider handlers must translate a failure-relevant discard into invalid `ModelHttpFailureEvidence`:

- `src/usage/openai_responses.py:866-871`
- `src/usage/openai_chat_completions.py:403-408`
- `src/usage/anthropic_messages.py:267-272`

`HttpResponseFailureObserver.observe()` treats that invalid evidence as parse ambiguity and settles the HTTP flow as `unknown`; the one-shot settlement gate then prevents later frames from reviving the flow and emitting a provider-failure report. Generic scanner tests cover callback dispatch, but the provider suites do not verify this fail-closed propagation. Branch coverage from the full addon suite left the observer calls above unexecuted.

The implementation currently behaves correctly, but the contract is not mutation-protected. A temporary diagnostic fed an OpenAI Responses `response.failed` event with partial data plus an overlong control line, followed by a valid `response.failed` event:

- Current handler: first and only settlement was `unknown`.
- With only the observer call in `on_event_discard()` removed: first and only settlement became `failure`.

The mutated behavior still passes the existing tests, so a future cleanup could silently turn an ambiguous stream into a reported provider failure.

## Impact

A regression in discard propagation can violate the fail-closed reporting contract and produce false model-provider failure telemetry after malformed or bounded-out SSE framing. The affected callbacks also reset parser state, so tests should protect both evidence propagation and clean recovery.

## Acceptance criteria

- Add deterministic provider-level tests that force a captured failure-relevant SSE event to be discarded through the control-line bound and assert one invalid evidence item with the correct event name.
- Cover OpenAI Responses, Chat Completions, and Anthropic Messages discard callbacks, including their parser-state reset/recovery behavior.
- Add a full-pipeline regression test in `tests/test_model_provider_failure_reporting.py` proving that a discarded ambiguous event followed by an otherwise reportable failure remains settled as `unknown` and emits no failure report.
- Ensure a mutation that removes or bypasses the provider `on_event_discard()` observer call fails the new tests.
- Keep the tests bounded and deterministic; do not rely on timing or external services.

seven332:
## Summary

- cover bounded SSE control-line discards in the OpenAI Responses, Chat Completions, and Anthropic provider handlers
- assert each handler emits invalid failure evidence and cleanly parses a later valid failure event
- prove the full HTTP pipeline keeps an ambiguous discarded event settled as unknown and does not report a later provider failure

## Scope

This is a test-only change. It does not modify provider parsing, failure reporting, APIs, persisted data, dependencies, or deployment behavior.

## Validation

- `uv lock --check`
- `uv sync --locked`
- `uv run --no-sync python -m pytest tests/test_openai_responses_sse.py tests/test_openai_chat_completions_usage.py tests/test_anthropic_messages.py tests/test_model_provider_failure_reporting.py` (263 passed)
- `uv run --no-sync python -m pytest tests/` (7,326 passed)
- temporarily removed each provider `on_event_discard()` observer call independently; each corresponding new test failed, then the production code was restored

Closes #32233

### DIFF
--- a/crates/runner/mitm-addon/tests/test_anthropic_messages.py
+++ b/crates/runner/mitm-addon/tests/test_anthropic_messages.py
@@ -8,6 +8,24 @@ from usage import (
     create_anthropic_messages_sse_usage_extractor,
     create_model_json_response_inspector,
 )
+from usage.model_http import ModelHttpFailureEvidence
+
+
+class _RecordingFailureObserver:
+    def __init__(self) -> None:
+        self.observed: list[ModelHttpFailureEvidence] = []
+
+    def needs_sse_event(self, event_name: str | None) -> bool:
+        return True
+
+    def observe(self, evidence: ModelHttpFailureEvidence) -> None:
+        self.observed.append(evidence)
+
+    def observe_json(self, evidence: ModelHttpFailureEvidence) -> None:
+        raise AssertionError(f"unexpected JSON evidence: {evidence}")
+
+    def finish(self) -> None:
+        return None
 
 
 def _create_parser_with_parse_errors():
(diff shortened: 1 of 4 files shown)


seven332:
## Code Review: PR #32244

**Reviewed head:** `65e115715ea04514e78c4576aae9dc09cc7e226c`

### Summary

The test-only change directly mutation-protects all three provider SSE discard callbacks and verifies the shared fail-closed reporting consequence through a real admitted HTTP flow. Scope and structure match the approved plan.

### Findings

No findings.

### Testing

Local validation passed: uv lock/sync, Ruff format/lint, BasedPyright, 263 focused tests, and the complete 7,326-test addon suite. Temporarily removing each provider observer call made its corresponding new test fail. GitHub CI remains a separate pending gate.

### Verdict

LGTM

PR #32253, opened 2026-09-07
feat(platform): fade clipped sidebar chat titles and scroll them on hover

tongx-blip:
## What

A sidebar chat title that does not fit ended in an ellipsis and stayed cut off. It now fades out at the edge instead, and hovering or focusing the row scrolls the text to its end and stops there — one pass, no loop. The leading pane icon and the trailing 32px action reserve do not move.

Design was reviewed and approved on a standalone mock before implementation, including the scroll pace: https://okou-sidebar-title-fade.sites.vm0.io

## How

`sidebarThreadTitleOverflowRef$` (`signals/chat-page/sidebar-thread-title.ts`) publishes the one measured input — how far the title is cut off — as `--okou-nav-title-overflow`, plus the duration that distance implies.

The stylesheet derives both the mask and the travel from that number, which is what makes the behaviour fall out without branching:

- at rest only the right edge is masked;
- at the end of the travel only the left one;
- a title that fits gets neither, because both clamps collapse at `0px`.

The travel itself is a registered `@property` length, so the transform and the mask stay in lockstep through the transition rather than drifting as two independent animations.

A `ResizeObserver` covers sidebar width changes, renames, and late web-font metrics — none of which remount the element, so the ref alone would leave a stale distance behind.

## Behaviour details

- Pace is `37px/s` clamped to `780ms`–`4200ms`, so long and short titles read as the same gesture instead of one crawling and the other flashing past.
- `750ms` delay before it starts, none on the way back, so sweeping a pointer down the list does not set every row in motion.
- `prefers-reduced-motion: reduce` keeps the fade and drops the travel.

## Testing

`sidebar.test.tsx` — "Fade a clipped chat title and pace its scroll by the hidden distance": with a stubbed title layout, a 26-character title in a 160px box publishes `74px` / `2000ms` and a title that fits publishes `0px` / the `780ms` floor.

The mask and travel were verified in a browser against the exact CSS block in this diff (measured shift `-58px`, left edge masked at the end of travel, right edge unmasked, short title unmasked throughout).

No feature switch: this refines an existing element rather than adding a user-facing surface.

### DIFF
--- a/turbo/apps/platform/src/signals/chat-page/sidebar-thread-title.ts
+++ b/turbo/apps/platform/src/signals/chat-page/sidebar-thread-title.ts
@@ -1,0 +1,64 @@
+import { command } from "ccstate";
+import { onRef } from "../utils.ts";
+
+/**
+ * Titles travel at one speed instead of in one duration, so a barely clipped
+ * title and a heavily clipped one read as the same gesture rather than one
+ * crawling and the other flashing past.
+ */
+const TITLE_SCROLL_PX_PER_SECOND = 37;
+const MIN_TITLE_SCROLL_MS = 780;
+const MAX_TITLE_SCROLL_MS = 4200;
+
+function titleScrollDurationMs(distance: number): number {
+  return Math.round(
+    Math.min(
+      MAX_TITLE_SCROLL_MS,
+      Math.max(
+        MIN_TITLE_SCROLL_MS,
+        (distance / TITLE_SCROLL_PX_PER_SECOND) * 1000,
+      ),
+    ),
+  );
+}
+
+function writeTitleOverflow(element: HTMLElement): void {
+  const distance = element.scrollWidth - element.clientWidth;
+  element.style.setProperty("--okou-nav-title-overflow", `${distance}px`);
+  element.style.setProperty(
+    "--okou-nav-title-duration",
+    `${titleScrollDurationMs(distance)}ms`,
+  );
+}
+
+/**
+ * Publishes how much of a sidebar thread title is cut off. The stylesheet
+ * derives both the fade mask and the hover travel from that one number, so a
+ * title that fits gets neither without a second code path.
+ *
+ * The observer watches the box for sidebar width changes and the text for
+ * renames and late web-font metrics. Neither remounts the element, so the ref
+ * on its own would leave a stale distance behind.
+ */
+export const sidebarThreadTitleOverflowRef$ = onRef(
+  command((_context, element: HTMLElement, signal: AbortSignal) => {
+    writeTitleOverflow(element);
+
+    const observer = new ResizeObserver(() => {
+      writeTitleOverflow(element);
+    });
+    observer.observe(element);
+    const text = element.firstElementChild;
+    if (text) {
+      observer.observe(text);
+    }
+
+    signal.addEventListener(
+      "abort",
+      () => {
+        observer.disconnect();
+      },
+      { once: true },
+    );
+  }),
+);
--- a/turbo/apps/platform/src/views/css/index.css
+++ b/turbo/apps/platform/src/views/css/index.css
@@ -614,6 +614,77 @@ body {
   --color-sidebar: hsl(var(--sidebar));
 }
 
+/* Sidebar thread titles.
+
+   A title that does not fit fades out instead of ending in an ellipsis, and
+   hovering or focusing its row scrolls the text to its end and stops there.
+
+   `--okou-nav-title-overflow` is the only measured input; the mask and the
+   travel are both derived from it, which is what makes an edge fade exactly
+   when content is cut off at it: at rest only the right edge is masked, at the
+   end of the travel only the left one, and a title that fits gets neither.
+   `sidebarThreadTitleOverflowRef$` writes it together with the duration. */
+@property --okou-nav-title-shift {
+  syntax: "<length>";
+  inherits: true;
+  initial-value: 0px;
+}
+
+.okou-nav-title {
+  --okou-nav-title-overflow: 0px;
+  --okou-nav-title-duration: 780ms;
+  --okou-nav-title-fade: 24px;
+  flex: 1;
+  min-width: 0;
+  overflow: hidden;
+  mask-image: linear-gradient(
+    90deg,
+    transparent 0,
+    #000
+      clamp(
+        0px,
+        calc(-1 * var(--okou-nav-title-shift)),
+        var(--okou-nav-title-fade)
+      ),
+    #000
+      calc(
+        100% -
+          clamp(
+            0px,
+            calc(var(--okou-nav-title-overflow) + var(--okou-nav-title-shift)),
+            var(--okou-nav-title-fade)
+          )
+      ),
+    transparent 100%
+  );
+  mask-repeat: no-repeat;
+  mask-size: 100% 100%;
+  transition: --okou-nav-title-shift 660ms cubic-bezier(0.4, 0, 0.2, 1);
+}
+
+.okou-nav-title > span {
+  display: block;
+  width: max-content;
+  white-space: nowrap;
+  transform: translateX(var(--okou-nav-title-shift));
+}
+
+/* The delay keeps a pointer sweeping down the list from setting every row in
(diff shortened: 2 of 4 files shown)


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: b04ddf2a2049180f1c298b196f99a6441487c430
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Replaces `truncate` on sidebar chat thread titles with a gradient mask, and scrolls a clipped title to its end on hover or focus. One measured value — how far the title is cut off — is published by `sidebarThreadTitleOverflowRef$`; the stylesheet derives both the mask and the travel from it, so an edge fades exactly when content is cut off at it and a title that fits gets neither, with no second code path. Presentation-only: no contract, schema, protocol, or persisted state is touched.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and logic: no unhandled null paths, API misuse, or wrong assumptions | Pass | `sidebar-thread-title.ts` reads `scrollWidth - clientWidth` on the clip box itself; verified in a browser against the shipped CSS block (58px overflow, 0px for a fitting title). |
| `REVIEW.md` | Type safety: no `any`, no broad casts | Pass | No casts in the diff; `lint-type-app` and `lint-types` green on `becdc08`. |
| `REVIEW.md` | Security: injection, secrets, missing auth | N/A | No input parsing, network call, credential, or endpoint in the diff. |
| `REVIEW.md` | `feat:` commits must ship integration tests | Pass | `sidebar.test.tsx` adds "Fade a clipped chat title and pace its scroll by the hidden distance". |
| `REVIEW.md` | New user-facing features gated behind a `FeatureSwitchKey` | N/A | Not a new surface. `.claude/skills/feature-switch/SKILL.md` requires a switch for new pages, sections, nav items, endpoints, integrations, or token capabilities; this changes the truncation treatment of an existing sidebar row. |
| `REVIEW.md` | Feature Regression Guard: replaced behavior needs regression coverage on the affected path | Pass | The title span gained a child wrapper. No existing assertion was modified, and all 8 `test-app` shards passed on `becdc08` — the 35 other `sidebar.test.tsx` tests plus `chat-list-test-helpers.ts` (`.okou-nav-copy` textContent) and `threadLinkByTitle` still traverse the new DOM. |
| `REVIEW.md` | DB/JSONB review gate | N/A | No `schema/**`, contract, or `jsonb().$type<...>()` change. |
| `docs/bad-smell.md` | No `any`; no `eslint-disable`/`@ts-ignore`/`prettier-ignore` suppressions | Pass | None in the diff. |
| `docs/bad-smell.md` | No dynamic `import()` in production code | Pass | Only static imports (`ccstate`, `../utils.ts`). |
| `docs/bad-smell.md` | No hardcoded URLs or environment-specific values | Pass | Only layout constants (`37` px/s, `780`/`4200` ms, `24px` fade), each named. |
| `docs/fallback.md` | Default is no fallback; no `??`/`||` chain guarding a state the contract already prevents | Pass | The one branch is `if (text) observer.observe(text)` around a nullable `firstElementChild`. It registers an observer target rather than fabricating a value or masking a corrupted state, and matches the reviewed shape already in `chat-conversation-locator.ts:991-997`. |
| `docs/fallback.md` | No negative tests asserting removed behavior stays removed | Pass | The new test asserts the new measurement, not the absence of the ellipsis. |
| `docs/fallback.md` | Every introduced fallback declared in the PR summary with surface, window, removal condition | N/A | The diff introduces no cross-version compatibility path: CSS and TSX ship in the same frontend bundle, and an old client keeps its own bundle. |
| `docs/effect.md` | DOM lifecycle work uses `onRef`, acquiring and releasing in one lifecycle | Pass | `sidebar-thread-title.ts:42-62` matches the guide's `ResizeObserver` example, including `signal.addEventListener("abort", ..., { once: true })`. |
| `docs/effect.md` | React render stays pure — no observers, writes, or DOM mutation during render | Pass | `ChatThreadItemLink` only calls `useSet` in render; all DOM work runs inside the ref command. |
| `docs/cache.md` | RC-7: pass the stable `useSet(onRef(...))` result straight to `ref`, never an inline arrow | Pass | `sidebar-threads.tsx:377` / `:407` — `ref={measureTitle}`, so `onRef`'s cleanup return is preserved. |
| `docs/cache.md` | RC-8: symmetric resource lifecycle for observers | Pass | One `ResizeObserver` created per mounted row, `disconnect()` on the ref's abort signal; rows are virtualized, so mount/unmount pairs cleanly. |
| `.claude/skills/ccstate/SKILL.md` | DOM refs go through `onRef`, never a command taking `HTMLElement | null` | Pass | `sidebarThreadTitleOverflowRef$ = onRef(command((_context, element, signal) => ...))`. |
| `.claude/skills/ccstate/SKILL.md` | `detach()` must not appear in the signals layer | Pass | The command is synchronous; no `detach()` in the new signals file. |
| `docs/testing.md` | Integration tests at the entry point; no unit tests for internal functions | Pass | The test drives the real page through `setupSidebarPage`, not `titleScrollDurationMs` directly. |
| `docs/testing.md` | Mock boundary: no `vi.mock` on relative paths, no direct `fetch` mocking, no fake timers, no mock-call assertions | Pass | The test adds none of these; thread data comes from `mockSidebarThreadStory`. |
| `docs/testing/patterns.md` | Pattern 7: configure `context.mocks` before page setup, enter through the production bootstrap | Pass | `stubSidebarTitleLayout()` and `mockSidebarThreadStory()` both run before `setupSidebarPage`. |
| `docs/testing/app-testing.md` | Page test order: mocks, `setupPage`, smallest readiness observation, then assert | Pass | Readiness is `await expect(within(sidebar()).findByText("Release plan")).resolves.toBeInTheDocument()`. |
| `docs/testing/anti-patterns.md` | AP-10: no direct component rendering | Pass | Full Router via `setupSidebarPage`. |
| `docs/testing/anti-patterns.md` | AP-7 / AP-8: do not assert implementation details or over-test | Pass, with a documented exception | The test asserts `--okou-nav-title-overflow` and `--okou-nav-title-duration`. See the row below; the test carries the rationale in-file rather than leaving it implicit. |
| `docs/testing/testing-external-behavior.md` | Construct and observe through the page; state the reason when an exception is unavoidable | Pass, exception stated | happy-dom has no layout engine — it reports every box as zero-width and paints nothing, so neither a box narrower than its text nor the resulting mask exists on the page surface. The doc's Exceptions clause applies, and the test now names why the production surface cannot construct or observe the state and why the case is worth pinning. |

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

Non-blocking notes, no change requested:

- `scrollWidth` and `clientWidth` are integers, so a title whose true overflow is fractional can stop up to 1px short of its end, and at that point the right-edge mask has already collapsed to `0px`. Sub-pixel accuracy is not available from these properties and the residue is not perceptible at 14px text.
- The mask uses a physical `90deg` gradient and a negative `translateX`, so both are LTR-specific. Every shipped UI locale (`de`, `en`, `es`, `fr`, `hi`, `id`, `it`, `ja`, `ko`, `pt`) is LTR and the app sets no `dir="rtl"` on its chrome, so this is not currently reachable. Worth revisiting with the first RTL locale.
- One `ResizeObserver` per visible row rather than a shared observer. With a virtualized list this is roughly 20 instances, and it keeps teardown owned by the row's own lifecycle, which `docs/cache.md` RC-8 prefers.

### Testing

- Coverage: Adequate. The new page test pins both branches of the derivation — a 26-character title in a 160px box publishes `74px` / `2000ms`, and a title that fits publishes `0px` and the `780ms` floor — which is what the mask and the travel are both computed from.
- Regression: all 8 `test-app` shards passed on `becdc08`, whose source diff is identical to this head. That covers the 35 existing `sidebar.test.tsx` tests over the changed title DOM, with no existing assertion modified.
- Conventions: Compliant, with the `testing-external-behavior.md` exception recorded above and in the test file.
- CI on `becdc08`: `lint-format`, `lint-knip`, `lint-types`, `lint-type-app`, `lint-type-api`, `test-app (1-8)` green. `lint-eslint` failed on `prefer-expect-resolves` and is fixed in `b04ddf2`; that run is still in flight.

PR #32284, opened 2026-09-07
fix(platform): keep the microphone button stable when switching conversations

### LINKED ISSUE #32255: bug: microphone button flashes when switching chat threads
## Problem and reproduction

The microphone button on the right side of the web chat composer briefly flashes when switching chat threads.

1. Open a chat thread with the composer visible.
2. Switch to another chat thread.
3. Observe the microphone button beside the model picker.

Expected: the idle microphone remains visually stable during navigation. Actual: the user reports a brief flash. This is a visible navigation glitch; no functional failure was reported.

The original report includes a screenshot identifying the microphone button with a red arrow. Browser, OS, exact duration, and frequency were not supplied; production reproduction is not yet independently verified.

## Repository evidence

At `f6b613497178855d93a23f4727809a722adc4947`:

- [`MicButton`](https://github.com/vm0-ai/vm0/blob/f6b613497178855d93a23f4727809a722adc4947/turbo/apps/platform/src/views/okou-page/chat-composer.tsx#L8515) disables the microphone while the current VoiceInputV2 draft state is unresolved.
- [`createVoiceDraftData`](https://github.com/vm0-ai/vm0/blob/f6b613497178855d93a23f4727809a722adc4947/turbo/apps/platform/src/signals/okou-page/composer-voice-input.ts#L116) reads the current user/org/target recording asynchronously from IndexedDB.
- The shared [`Button`](https://github.com/vm0-ai/vm0/blob/f6b613497178855d93a23f4727809a722adc4947/turbo/packages/ui/src/components/ui/button.tsx#L17) applies `disabled:opacity-50`. This produces a dim-to-normal transition as an idle draft resolves, consistent with the reported flash.
- Existing page integration tests deliberately keep voice input disabled during draft hydration and verify recording isolation during navigation.

## Decision and scope

Proceed with one focused UI bug fix. Correct the idle microphone's navigation appearance while retaining the draft-loading interaction guard, quota checks, and recording/transcription states. The async draft lookup is required to recover unfinished recordings; an unresolved draft must not be treated as an empty draft.

## Acceptance criteria

- Switching between idle chat composers does not briefly dim or hide the microphone solely because the destination's voice draft is loading.
- Mouse and keyboard activation remain blocked while the current voice draft is unresolved.
- Quota loading, microphone startup, transcription, and restored-recording recovery retain their existing behavior and feedback.
- No recording or retry state leaks across chat targets, users, or organizations.

## Verification notes

Use existing page-level voice draft loading, quota, and isolation coverage. Do not run a full local Vitest suite or start a local development server; full checks belong to the PR pipeline. Browser-specific frequency and live visual verification remain open.

e7h4n:
Closes #32255

Switching conversations briefly dimmed the composer microphone button. Every conversation switch builds a new composer, and `MicButton` gated `disabled` on `useResolved(voice.state$)`, which drops to loading while the new target's stored voice draft is read from IndexedDB. The voice draft data also re-read storage through a revision counter after every mutation and on unmount, so the same loading flip happened after start, stop, retry, and discard.

## Changes

- `MicButton` reads the voice draft status with `useLastResolved`, so a target change keeps the last status until the new read resolves. `run$` already awaits that read before choosing start or retry. Microphone availability is a synchronous computed and is read with `useGet`, so a freshly mounted button no longer renders empty for one pass.
- The voice draft recording is derived instead of reloaded. Storage is read once per storage key through an async computed, and each mutation (create, append, delete, delivery) records the recording it owns as state. `revision$`, `reload$`, `deliveredRecordingId$`, and the reload on unmount are removed.
- `appendVoiceDraftSamples` returns the stored recording so the append callback records the committed sample count without another read.
- Retry reads storage again on purpose: a failed restore can recover, and a recording saved by another composer for the same target (for example the forward dialog) stays transcribable.
- `ComposerFooter` keeps `useResolved`, so a pending target still shows the ordinary toolbar rather than another conversation's Retry controls, as the existing loading tests require.

## Verification

- New page test: switching between two conversations keeps `Voice input` enabled while the second conversation's stored draft read is still pending, and recording afterwards inserts the transcript into that conversation.
- The ten `chat-capability-voice-*` page-test files (89 tests) plus `chat-expected-failures`, `sidebar`, `preferences-page`, and `org-billing-tab` (92 tests) pass locally in batches of at most five files. Two forwarding cases timed out once inside a five-file batch and passed when that file ran alone.
- Platform `check-types`, oxlint, type-aware oxlint, eslint on the changed files, prettier, and `knip --workspace apps/platform` pass.

### DIFF
--- a/turbo/apps/platform/src/signals/external/voice-draft-store.ts
+++ b/turbo/apps/platform/src/signals/external/voice-draft-store.ts
@@ -111,14 +111,15 @@ export async function createVoiceDraftRecording(
   );
 }
 
+/** Append one PCM chunk and return the recording as stored afterwards. */
 export async function appendVoiceDraftSamples(
   key: string,
   id: string,
   sequence: number,
   samples: Float32Array,
-): Promise<void> {
+): Promise<VoiceDraftRecordingRecord> {
   const database = await openVoiceDraftRecordingDatabase();
-  await withCleanup(
+  return await withCleanup(
     runIndexedDbTransaction(
       {
         database: "voice_drafts",
@@ -141,16 +142,13 @@ export async function appendVoiceDraftSamples(
             .objectStore("chunks")
             .add(samples.slice().buffer, [key, id, sequence]),
         );
-        await track(
-          drafts.put(
-            {
-              ...recording,
-              chunkCount: sequence + 1,
-              sampleCount: recording.sampleCount + samples.length,
-            },
-            key,
-          ),
-        );
+        const appended = {
+          ...recording,
+          chunkCount: sequence + 1,
+          sampleCount: recording.sampleCount + samples.length,
+        };
+        await track(drafts.put(appended, key));
+        return appended;
       },
     ),
     () => {
--- a/turbo/apps/platform/src/signals/okou-page/composer-voice-input.ts
+++ b/turbo/apps/platform/src/signals/okou-page/composer-voice-input.ts
@@ -12,7 +12,6 @@ import {
   onRef,
   onRejection,
   settle,
-  withCleanup,
   createChildAbortController,
 } from "../utils.ts";
 import { voiceInputV2Enabled$ } from "../external/feature-switch.ts";
@@ -49,6 +48,12 @@ interface ComposerVoiceInputState {
   readonly status: "idle" | "recording" | "failed";
   readonly recording: VoiceDraftRecordingRecord | null;
   readonly message?: string;
+}
+// The recording this composer last created, appended to, or removed under a
+// storage key. Storage is only read for a key this composer has not changed.
+interface OwnedVoiceDraftRecording {
+  readonly key: string;
+  readonly recording: VoiceDraftRecordingRecord | null;
 }
 export type ComposerVoiceInputSignals = ReturnType<
   typeof createComposerVoiceInputSignals
@@ -113,27 +118,36 @@ function createVoiceDraftData(draftTarget: string) {
     const identity = await get(authenticatedIdentity$);
     return JSON.stringify([identity.userId, identity.orgId, draftTarget]);
   });
-  const revision$ = state(0);
-  // A successful text handoff consumes this recording even if local deletion
-  // fails. Keep that domain fact so Retry cannot insert the same text twice.
-  const deliveredRecordingId$ = state<string | null>(null);
+  const storedRecording$ = computed(
+    async (get): Promise<VoiceDraftRecordingRecord | null> => {
+      const key = await get(storageKey$);
+      return await readVoiceDraftRecording(key);
+    },
+  );
+  const ownedRecording$ = state<OwnedVoiceDraftRecording | null>(null);
+  // Mutations record their own outcome, so the recording never waits for a
+  // second storage read after this composer has changed it.
   const recording$ = computed(
     async (get): Promise<VoiceDraftRecordingRecord | null> => {
       if (!get(voiceInputV2Enabled$)) {
         return null;
       }
-      get(revision$);
-      const deliveredId = get(deliveredRecordingId$);
+      const owned = get(ownedRecording$);
       const key = await get(storageKey$);
-      const recording = await readVoiceDraftRecording(key);
-      return recording?.id === deliveredId ? null : recording;
+      return owned?.key === key ? owned.recording : await get(storedRecording$);
     },
   );
-  const reload$ = command(({ set }) => {
-    set(revision$, (value) => {
-      return value + 1;
-    });
-  });
+  // Retry reads storage again so a failed restore can recover and a recording
+  // saved by another composer for the same target can be transcribed.
+  const restoreRecording$ = command(
+    async ({ get, set }, signal: AbortSignal) => {
+      const key = await get(storageKey$);
+      signal.throwIfAborted();
+      const recording = await readVoiceDraftRecording(key);
+      signal.throwIfAborted();
+      set(ownedRecording$, { key, recording });
+    },
+  );
   const capture = createVoiceDraftCaptureSignals();
   const captureError$ = state<unknown>(null);
   const state$ = computed(async (get): Promise<ComposerVoiceInputState> => {
@@ -158,9 +172,9 @@ function createVoiceDraftData(draftTarget: string) {
   });
   return {
     storageKey$,
-    deliveredRecordingId$,
+    ownedRecording$,
     recording$,
-    reload$,
+    restoreRecording$,
     capture,
     captureError$,
     state$,
@@ -176,7 +190,7 @@ function createVoiceDraftTranscription(
   readEditorContext$: Command<VoiceIoEditorContext, []>,
   lastAssistantMessage$: Computed<string | undefined>,
 ) {
-  const { recording$, storageKey$, deliveredRecordingId$, reload$ } = data;
+  const { recording$, storageKey$, ownedRecording$ } = data;
   const incremental = createVoiceDraftTranscriptionSignals({
     storageKey$,
     readContext$: command(({ get, set }) => {
@@ -206,13 +220,14 @@ function createVoiceDraftTranscription(
     }
 
     signal.throwIfAborted();
-    set(deliveredRecordingId$, recording.id);
+    // A successful text handoff consumes this recording even if local deletion
+    // fails, so Retry cannot insert the same text twice.
+    set(ownedRecording$, { key, recording: null });
     const removed = await settle(
       deleteVoiceDraftRecording(key, recording.id),
       signal,
     );
     signal.throwIfAborted();
-    set(reload$);
     if (!removed.ok) {
       L.error("Voice recording cleanup failed", removed.error);
       toast.error(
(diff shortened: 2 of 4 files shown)


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 4d9796d66985d132fbfb3de58acedcebebcb12f5
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary
Fixes the composer microphone button dimming on every conversation switch (#32255). `MicButton` now reads the voice draft status through `useLastResolved` and microphone availability through `useGet`, and the voice draft recording is derived from the mutations this composer performs (create, append, delete, delivery) plus one storage read per key, replacing the `revision$`/`reload$` counter and `deliveredRecordingId$`. Retry deliberately re-reads storage. A new page test covers the conversation-switch regression.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness: race conditions, edge cases, null paths | Pass | `composer-voice-input.ts`: `recording$` reads `ownedRecording$` before its first `await`, so the dependency is tracked; `run$` still awaits `state$` before choosing start/retry, so an enabled button during a pending read cannot start a duplicate recording; `removeEmptyRecording` keeps its storage read, which IndexedDB orders after committed chunk writes; the append updater is guarded by recording id. |
| `REVIEW.md` | Security (injection, secrets, auth) | N/A | Client-side IndexedDB and React state only; no endpoints, inputs, or credentials change. |
| `REVIEW.md` | Type safety: no `any`, explicit types | Pass | `OwnedVoiceDraftRecording` interface; `appendVoiceDraftSamples` now returns `Promise<VoiceDraftRecordingRecord>`; updater callback is typed by the `state<...>` generic. |
| `REVIEW.md` | Style: function length, duplication, WHY comments | Pass | Largest changed command (`start$`) stays under ~80 lines; comments explain ownership and ordering rationale, not mechanics. |
| `REVIEW.md` | Fallback discipline and `Fallbacks` declaration | Pass | The diff introduces no compatibility branch; `owned?.key === key ? owned.recording : await get(storedRecording$)` is derivation, and `restoreRecording$` is the product's Retry action. No `Fallbacks` section required. |
| `REVIEW.md` | Testing coverage: `fix:` needs a regression test | Pass | `chat-capability-voice-draft-loading.test.tsx` "Keep voice input enabled while switching conversations" defers the second conversation's `okou-voice-drafts` open and asserts `Voice input` is enabled while the read is pending, then records into that conversation. Under the previous `useResolved` gate this assertion fails. |
| `REVIEW.md` | Feature regression guard for changed existing behavior | Pass | Existing tests keep the intentional disabled states: initial-load gate ("Keep composer controls visible while checking the local voice draft") and cross-agent pending state ("Do not show another agent's retryable recording while the current draft is pending") remain unchanged and pass; `ComposerFooter` keeps `useResolved`. |
| `REVIEW.md` | DB/JSONB contract gate | N/A | No `packages/db`, api-contracts, or JSONB changes; the IndexedDB `drafts` record shape is unchanged. |
| `REVIEW.md` | Deployment compatibility routing | N/A | No frontend/backend, runner, queue, schema, or persisted-shape boundary changes; the `appendVoiceDraftSamples` return value is in-process. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions, no dynamic imports | Pass | None added in the four changed files; eslint, oxlint, and type-aware oxlint pass on them. |
| `docs/bad-smell.md` | Error handling: fail fast, no defensive try/catch | Pass | No new try/catch; `settle`/`withVoiceDraftFailureToast` usage is unchanged and `restoreRecording$` lets a failed read propagate to the invocation loadable. |
| `docs/bad-smell.md` | Interface changes documented | Pass | PR body records the `appendVoiceDraftSamples` return-type change and the removed `revision$`/`reload$`/`deliveredRecordingId$` signals (all module-private). |
| `docs/fallback.md` | No fallback for states the contract prevents; declare any new fallback | Pass | No `??`/`||` defaults on contract-owned fields and no version-tolerance branch were added. |
| `docs/effect.md` | Derived values use `computed`; do not synchronize state through a command | Pass | `recording$` derives from `ownedRecording$` and `storedRecording$` instead of being invalidated by `reload$`; the unmount `reload$` is gone. |
| `docs/effect.md` | One semantic command per user action; compose commands through commands | Pass | `run$` still owns start/finish/discard/retry; Retry awaits `restoreRecording$` then `transcribe$` as sub-commands with the owner signal. |
| `docs/effect.md` | Explicit cancellation with an owning `AbortSignal` | Pass | `restoreRecording$` takes the invocation signal and checks `throwIfAborted()` after each await that ignores the signal. |
| `docs/cache.md` | RC-1/RC-2: render purity, no render-time signal allocation | Pass | `MicButton` only swaps hook kinds (`useGet`, `useLastResolved`); no commands, stores, or factories run in render. |
| `docs/cache.md` | RC-3/RC-4: no unbounded caches; state owned by the narrowest lifecycle | Pass | `ownedRecording$` lives in the per-composer factory created by `createComposerVoiceInputSignals`; no module-level map or cross-composer retention was added. |
| `docs/cache.md` | RC-5: no duplicate mutable source of truth; drafts carry identity | Pass | The owned recording is this composer's own draft state carrying its storage `key`; storage is consulted only for keys it has not written. See the non-blocking note on the unused `progress` field. |
| `docs/cache.md` | RC-7/RC-8: stable refs and symmetric resource teardown | Pass | `setRootRef$` stays an `onRef` command; the abort listener still cancels capture and clears owner/element; capture resources are untouched. |
| `.claude/skills/ccstate/SKILL.md` | Async data as reactive `computed`; consumers use loadable hooks | Pass | `storedRecording$` is a `computed(async)`; the view reads it through `useLastResolved(signals.voice.state$)`. The counter-bump invalidation pattern was replaced because local mutations already know their outcome. |
| `.claude/skills/ccstate/SKILL.md` | `get`/`set` never escape a command; no `detach` in signals | Pass | The `append`/`fail` callbacks are defined inside the `start$` command body as before; no accessor is passed or stored; no `detach` added (ccstate lint rules pass). |
| `.claude/skills/ccstate/SKILL.md` | `throwIfAborted()` after awaits that do not take the signal | Pass | `restoreRecording$` after `get(storageKey$)` and `readVoiceDraftRecording`; `discard$`/`transcribe$` keep their existing checks. |
| `.claude/skills/ccstate/SKILL.md` | `onRef` for DOM refs with abort cleanup | Pass | `mount$` unchanged apart from removing the unmount reload. |
| `docs/react-commit.md` | Do not swap loadable hooks mechanically; the UI must intend the transition | Pass | The mic button should not expose the pending-read transition on a target change; a freshly mounted button still starts in `loading`, which the initial-load test relies on. |
| `docs/react-commit.md` | Subscribe to the smallest semantic value | Pass | `voiceDraftStatus` derives a primitive from `state$`; `available` now reads the synchronous computed directly. (Optional follow-up: expose a `status$` primitive to skip equal-object re-renders.) |
| `docs/testing.md` | Integration tests at the page boundary; mock only external boundaries | Pass | The new test runs through `setupPage`, mocks HTTP through `context.mocks`, and spies only on the external `idb` package's `openDB`, matching the file's existing `vi.mock("idb", importActual)` setup. |
| `docs/testing.md` | No fake timers; assert behavior, not mock calls | Pass | Real timers with `waitFor`; assertions are button enabled state, absence of `Retry`, and composer text. |
| `docs/testing/app-testing.md` | Page test order: mocks, `setupPage`, smallest ready state, one action, result | Pass | Mocks then `setupPage`; readiness via `findEnabledButton("Voice input")`; `click(await findLink(...))` once; results awaited with `waitFor`. The deferred `openDB` spy is installed after the first conversation resolved so only the second read is gated, the same pattern as the neighbouring cross-agent test. |
| `docs/testing/app-testing.md` | Fast role queries; `waitFor` holds only queries and assertions; no internal-state assertions | Pass | `queryButton`/`findLink` wrap `queryAllByRoleFast`; `waitFor` callbacks contain only `expect`; no store or signal assertions. |
| `docs/testing/testing-external-behavior.md` | Construct state through page interactions and observe the page | Pass | Conversation switch via the sidebar link, recording via the button, verification via composer text. |
| `docs/testing/patterns.md` | Platform page tests use the production bootstrap and shared mock helpers | Pass | `setupPage` with `context`; `installRunChat` and `lifecycle.setThreadList` from the shared fixtures. |
| `docs/testing/anti-patterns.md` | AP-1/AP-4/AP-5/AP-7/AP-10: no mock-call assertions, internal mocks, fake timers, implementation-detail assertions, or direct rendering | Pass | None present in the new test. |

### Findings

#### Critical (P0)
- None.

#### High Priority (P1)
- None.

Non-blocking notes:
- `composer-voice-input.ts` `OwnedVoiceDraftRecording`: the owned record copies the full `VoiceDraftRecordingRecord`, so its `progress` field can lag behind checkpoints written by `saveVoiceDraftProgress`. Nothing reads `progress` through `recording$` today (transcription reads storage directly), but narrowing the owned shape to `id`/`sampleCount`/`chunkCount` or documenting this would remove the trap.
- `removeEmptyRecording`: if another composer for the same key removed the record while capture start was failing, the owned recording stays until Retry re-reads storage. Only the forward dialog can share a key with a mounted composer, and Retry self-corrects, so this is acceptable.

### Testing
- Coverage: Adequate. New regression page test for the conversation-switch flash; existing voice tests keep the initial-load and cross-agent pending gates, retry/discard, quota, forwarding, isolation, hydration, and recording flows.
- Conventions: Compliant. Page test via `setupPage`, typed contract mocks, MSW for HTTP, external-only spy (`idb`), real timers, behavior assertions.
- Author-reported local runs: ten `chat-capability-voice-*` files (89 tests) plus `chat-expected-failures`, `sidebar`, `preferences-page`, `org-billing-tab` (92 tests) pass; two forwarding cases timed out once inside a five-file batch and passed alone. `check-types`, oxlint, eslint, prettier, and `knip --workspace apps/platform` pass. PR CI runs the full suite.

PR #32346, opened 2026-09-07
fix(sandbox-fc): own snapshot cow before network setup

### LINKED ISSUE #32285: bug(rust): preserve snapshot COW files after uncertain early cleanup
## Summary

Early snapshot teardown can recursively delete COW backing files even when the NBD finalizer explicitly reports that they are not safe to delete. `SnapshotAttemptDirGuard` remains armed after COW ownership has been acquired, until network-pool initialization succeeds and `SnapshotAttempt` is constructed.

Audited on `main` at `9fd53c47daba5922b0a5de5abcb4b82f44f0a2f8`. The three snapshot source files below were also checked against live `main` before filing; their blob hashes are unchanged.

## Exact locations and failure path

- `crates/sandbox-fc/src/snapshot/mod.rs:135-168`: creates the directory guard, successfully acquires a pooled NBD COW device, then awaits `NetnsPool::create_checked`. The error branch at lines 153-155 returns without reaching `attempt_dir_guard.disarm()` at line 168.
- `crates/sandbox-fc/src/snapshot/attempt/mod.rs:51-68`: `cleanup_after_netns_pool_failure` awaits the COW finalizer, logs an error, and returns without propagating its backing-file safety decision.
- `crates/sandbox-fc/src/snapshot/cow.rs:94-117`: the COW finalizer deliberately returns before directory cleanup when its result is `DeviceMayStillReferenceBackingFiles` (lines 112-113).
- `crates/sandbox-fc/src/snapshot/cow.rs:172-185`: dropping the still-armed directory guard unconditionally calls `remove_dir_all`, overriding that preservation decision.
- `crates/nbd-cow/src/device/pooled.rs:404-406`: exhausted shutdown retries abandon the device, retire the lease as uncertain, and return a device-cleanup error rather than declaring the backing files safe.

Concrete scenario:

1. COW device creation succeeds.
2. Network-pool creation fails, for example because pool-lock allocation or host IP-forwarding setup fails.
3. COW shutdown also exhausts its retries. The detailed finalizer reports `backing_files_safe_to_delete() == false` and intentionally leaves the backing files in place.
4. The original network-setup error is returned. Leaving the scope drops the armed directory guard, deleting `work/attempts/<token>/cow.img` and any bitmap despite the unresolved device state.

Cancellation while the same early cleanup finalizer is pending has the same ownership gap: the detached finalizer continues, but the directory guard can remove its files before it resolves.

## Impact

This defeats the existing NBD cleanup safety contract: a failed snapshot loses the backing-file paths that should be retained while device cleanup is uncertain, and cancellation can race an active cleanup owner. This report does not claim that a completed snapshot or another attempt's files are deleted; the affected files belong to this failed attempt.

## Verification

- Traced the production error branch, the detailed NBD destroy result, the preservation branch, and the guard's only disarm call.
- A rootless isolated test compiled the exact `SnapshotAttemptDirGuard` definition extracted from the audited source. With its production early-error armed state and a file descriptor still held, dropping the guard unlinked `cow.img` (`nlink() == 0`) and failed an assertion that its backing path remained present. A disarmed-owner control passed.
- That isolated test verifies the filesystem/guard behavior, not an end-to-end KVM/NBD failure injection.
- Existing `cargo test --locked --profile local -p sandbox-fc --lib snapshot::` passed all 89 tests. Current guard tests cover an unowned directory and a disarmed directory, but not this post-COW/pre-`SnapshotAttempt` failure boundary.
- Checked open snapshot/COW/backing-file/cleanup issues; no matching open issue was found. The previously closed cleanup work in #11522 and #16452 does not cover this still-live early ownership gap.

## Suggested acceptance criteria

- Transfer directory and COW cleanup ownership before the first post-acquisition await, or explicitly make the guard aware that it may no longer recursively remove owned backing files.
- On an uncertain device-cleanup outcome, preserve the attempt's backing paths; on confirmed safe cleanup, remove them as usual.
- Cancelling early network setup or cleanup must not unlink backing files while a COW finalizer is still pending.
- Add rootless, explicitly synchronized regression coverage for early network-setup failure with safe and unsafe cleanup results, and cancellation while cleanup is pending.
- Preserve the original setup error and keep pre-COW failures and normal successful cleanup leak-free.

seven332:
## Summary

- Construct the snapshot attempt owner immediately after pooled COW acquisition, before network-pool initialization. Consume and disarm the unowned-directory guard in that constructor so it cannot override the COW finalizer's backing-file safety decision.
- Reuse the attempt's explicit and detached cleanup paths for early network failure and cancellation; preserve the original setup error and never delete stable socket directories from detached cleanup.
- Add synchronized, rootless lifecycle coverage for safe/unsafe cleanup, cancellation during network initialization and pending cleanup, and successful pool attachment. Tests use real temporary COW/bitmap files and a test-only host-device teardown boundary.

## Scope

No API/protocol, snapshot format, migration, NBD retry-policy, or GC changes. Confirmed unsafe cleanup intentionally retains this attempt's files. Kernel-level failure injection and process/runtime crash guarantees are not claimed.

## Validation

- `cargo fmt --all --check`
- `cargo clippy --locked --profile local -p sandbox-fc -p runner --all-targets --all-features`
- `cargo doc --locked --profile local -p sandbox-fc -p runner --all-features --no-deps`
- `cargo test --locked --profile local -p sandbox-fc --all-targets --all-features`: 857 library tests and 1 prewarm integration test passed. The existing ignored real-Firecracker CPU test requires metal-host fixtures and remains CI coverage; no ignore/skip settings changed.
- `cargo test --locked --profile local -p sandbox-fc --lib snapshot::`: 93 passed.
- New `snapshot::attempt::tests::early_setup::` regressions repeated 3 additional times: 4/4 passed each time.
- Repository pre-commit checks passed: workspace-wide Rust formatting, Clippy, documentation, file-size validation, and commit-message validation.

Closes #32285

### DIFF
--- a/crates/sandbox-fc/src/snapshot/attempt/cleanup.rs
+++ b/crates/sandbox-fc/src/snapshot/attempt/cleanup.rs
@@ -1,14 +1,14 @@
 use std::path::{Path, PathBuf};
 
 use nbd_cow::KeptCow;
-use nbd_cow::PooledNbdCowDevice;
 use nbd_cow::pool::DevicePoolHandle;
 
 use crate::network::{NetnsLease, NetnsPool};
 
 use super::super::SnapshotError;
 use super::super::cow::{
-    destroy_snapshot_cow_after_error, destroy_snapshot_cow_and_cleanup_attempt_dir,
+    SnapshotCowDevice, destroy_snapshot_cow_after_error,
+    destroy_snapshot_cow_and_cleanup_attempt_dir,
 };
 use super::super::output::cleanup_workspace_image_file_sync;
 use super::super::publish::SnapshotPublishAttempt;
@@ -24,7 +24,7 @@ async fn release_snapshot_netns(
     }
 }
 
-async fn destroy_snapshot_cow_after_workflow_error(cow_device: PooledNbdCowDevice) {
+async fn destroy_snapshot_cow_after_workflow_error(cow_device: SnapshotCowDevice) {
     if let Err(e) = destroy_snapshot_cow_and_cleanup_attempt_dir(cow_device).await {
         tracing::warn!(error = %e, "failed to destroy COW device after snapshot error");
     }
@@ -124,7 +124,7 @@ impl SnapshotCleanupPresence {
 pub(super) struct SnapshotCleanupResources {
     pub(super) netns_pool: Option<NetnsPool>,
     pub(super) device_pool: Option<DevicePoolHandle>,
-    pub(super) cow_device: Option<PooledNbdCowDevice>,
+    pub(super) cow_device: Option<SnapshotCowDevice>,
     pub(super) workspace_image: AttemptWorkspaceImage,
     pub(super) publish_attempt: Option<SnapshotPublishAttempt>,
     pub(super) network: Option<NetnsLease>,
@@ -133,13 +133,11 @@ pub(super) struct SnapshotCleanupResources {
 
 impl SnapshotCleanupResources {
     pub(super) fn new(
-        netns_pool: NetnsPool,
         device_pool: DevicePoolHandle,
-        cow_device: PooledNbdCowDevice,
+        cow_device: SnapshotCowDevice,
         workspace_image_path: PathBuf,
     ) -> Self {
         Self {
-            netns_pool: Some(netns_pool),
             device_pool: Some(device_pool),
             cow_device: Some(cow_device),
             workspace_image: AttemptWorkspaceImage::new(workspace_image_path),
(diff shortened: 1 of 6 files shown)


seven332:
## Code Review: PR #32346

**Reviewed head:** `f0edfe3f366ed0686f2f5a0a12b1deccc92b5e6a`

### Summary

The attempt owner now acquires the directory guard and COW synchronously before network setup. Explicit failure and cancellation reuse existing finalizers, preserve unsafe backing paths and the original setup error, and retain the stable-socket cleanup boundary. All six self-review passes completed without findings on this unchanged head.

### Findings

No findings.

### Testing

857 library tests and 1 prewarm integration test passed; snapshot selection passed 93 tests; four new synchronized lifecycle regressions passed three additional repeated runs. Scoped and workspace-hook formatting, Clippy, documentation and commit checks passed.

Actual kernel NBD failure injection was not performed. The existing real-Firecracker CPU test requires metal-host CI. No process/runtime-crash cleanup guarantee is claimed. CI readiness is checked separately.

### Verdict

LGTM

PR #32295, opened 2026-09-07
fix(chat): validate snapshot watermark cursors

lancy:
## Summary

- validate lifecycle-event cursors against the same-scope snapshot watermark in the same SQL statement that selects the next page
- allow an exact snapshot watermark without a retained marker row while rejecting covered, allocation-gap, future, and cross-scope cursors
- add focused BDD coverage for markerless empty and non-empty tails, real post-snapshot cursors, scope isolation, and pagination without omissions or duplicates

## Scope

- reader and focused test changes only
- does not change snapshot compaction or marker pruning behavior
- does not add migrations or change schemas, API contracts, client persistence/retry behavior, or production data

## Testing

- `pnpm prettier --check apps/api/src/signals/services/chat-thread-event.service.ts apps/api/src/test-fixtures/chat-thread-events.ts apps/api/src/signals/routes/__tests__/chat-threads.bdd.test.ts`
- `pnpm -F api lint`
- `pnpm -F api check-types`
- `pnpm knip --workspace api`
- `DATABASE_URL=postgresql://postgres:postgres@localhost:5432/vm0_test pnpm -F api exec vitest run src/signals/routes/__tests__/chat-threads.bdd.test.ts -t 'returns chat thread snapshot and lifecycle events with cursor expiry|keeps no-snapshot cursors anchored to real scoped rows|reads an exact markerless snapshot watermark atomically|validates real cursors after a markerless snapshot watermark|paginates a markerless snapshot tail without gaps or duplicates|preserves UTC snapshot and event cutoff boundaries outside UTC'`

Fixes #32254

Issue #28636, opened 2026-08-22
refactor(chat): decouple event cursors from retained anchor rows (EPIC)


lancy:
## Decision

This issue is the controller EPIC for a compatibility-first, two-slice rollout. The reader must stop depending on a physical snapshot-marker event before compaction is allowed to delete that marker. The two changes intentionally ship in separate PRs and separate production releases.

Do not create or dispatch Slice 2 until Slice 1 is merged, independently accepted, included in a production release, verified live, and previous API revisions have drained.

- Controller thread: https://app.okou.ai/chats/59c5ddd5-f2a3-4021-829f-7bb985c89590
- Slice 1 owner thread: https://app.okou.ai/chats/e646a952-3bf8-4e77-8b78-4d2832ac6ba8
- Slice 2 owner thread: https://app.okou.ai/chats/945151f4-3b1a-4a6d-bece-51db26f0af52

## Current production evidence

A read-only MaskDB refresh at 2026-09-07T07:06:44Z found:

- 4,928 user/org scopes with a Chat Thread snapshot;
- 3,153 snapshots with a non-null event watermark, all 3,153 still backed by the exact physical marker row;
- 29 exact snapshot-anchor events whose referenced Agent and Chat Thread were both gone: 27 have the approved canonical `agent_id = NULL`, while two retain Agent IDs that no longer resolve;
- at least one of those strict orphan anchors was created after this issue opened, with the newest created on 2026-09-02;
- another 40 exact anchors whose Chat Thread was gone while the Agent still existed;
- zero currently broken snapshot-marker references.

This proves both sides of the rollout decision: the issue is recurring, while the current retained-marker workaround is still preserving production correctness. It is not an emergency data repair, and deleting markers before reader compatibility would introduce a fresh-snapshot/410 failure mode.

## Current implementation

At dispatch baseline `origin/main` `d46eb931e0a7e053750a4827204d45782a7565d1`:

- `chat_thread_snapshots.latest_event_seq_id` already stores the durable per-user/org compacted boundary.
- `getChatThreadEventsSince` still requires the exact same-scope physical event row for `sinceSeqId`; a missing row maps to the existing intentional 410.
- compaction joins the snapshot back to that exact marker, deletes only `seq_id < marker.seq_id`, and requires the event's Agent to exist. The marker and deleted-Agent orphans therefore remain.
[...]

PR #32361, opened 2026-09-07
refactor(chat): retire snapshot marker retention and prune covered events

### LINKED ISSUE #32350: refactor(chat): retire snapshot marker retention and prune covered events
## Background

Parent: #28636. This is Slice 2 of the compatibility-first rollout.

Slice 1 is already merged, independently accepted, released as `api-v1.563.0`, and verified in production after previous API revisions drained. The production reader now treats `chat_thread_snapshots.latest_event_seq_id` as the durable same-scope watermark and accepts equality even when the physical marker row is absent.

This issue is written against `origin/main` `845f246a84d3e5fbceba5b2965799cdef6b18ab8` after that production gate.

The remaining retention dependency is in `turbo/apps/api/src/signals/services/cron-compact-chat-thread-snapshots.service.ts`:

- `candidateScopesCte` compares the stored watermark with the current physical latest event (`:89-123`). After covered rows are pruned, that physical latest event can be below the stored watermark or absent, so `IS DISTINCT FROM` would repeatedly select the scope.
- `rebuiltCte` reads the absolute physical latest event (`:126-218`), and `upsertedCte` blindly replaces both `latest_event_id` and `latest_event_seq_id` (`:221-250`). If the old marker is gone and no newer event exists, this can regress or clear the durable watermark.
- The current pruning query (`:301-352`) joins the snapshot back to the exact marker, requires the event's Agent to exist, and deletes only `seq_id < marker.seq_id`. It therefore retains the marker itself plus covered orphan events whose Agent or Thread is gone.

The database compatibility trigger in `turbo/packages/db/src/migrations/1078_baseline.sql:409-437` clears `latest_event_seq_id` when `latest_event_id` becomes null. Keep the existing ID as an opaque compatibility value when no newer event exists; do not edit the baseline, trigger, schema, or historical migrations.

Live MaskDB at 2026-09-07T11:26:08Z found:

- 4,932 snapshot scopes;
- 3,157 non-null marker references, all 3,157 still backed by the exact same-scope physical event row;
- 29 exact marker events whose Agent and Chat Thread are both gone: 27 have canonical `agent_id = NULL`, while two retain non-null Agent IDs that no longer resolve;
- the newest strict orphan was created on 2026-09-02;
- 41 more exact markers whose Chat Thread is gone while the Agent still exists;
- zero broken marker references.

A deterministic 100-scope sample found 87 scopes retaining only the exact marker. Marker retirement therefore needs to work when pruning leaves no physical event at or below the watermark.

### Keep snapshot watermarks monotonic

For each existing user/org snapshot:

- Only a real same-scope event with `seq_id > snapshot.latest_event_seq_id` may advance the snapshot boundary.
- When one or more newer events exist, choose the greatest sequence, apply the current live Chat Thread projection, and advance `latest_event_id` and `latest_event_seq_id` together to that event.
- When no newer event exists, preserve both stored values exactly, including the opaque `latest_event_id`, while still allowing stale projection refresh and deleted-Agent cleanup. Report zero newly applied events.
- Candidate selection must not wake a scope merely because the greatest retained physical event is lower than or missing relative to the stored watermark. A missing snapshot, a newer event, or the existing stale-projection rule may still select it.
- New scopes without a snapshot retain the current behavior: use the latest real event when present, otherwise create the empty snapshot boundary.

Do not derive validity or advancement from UUID ordering or `chat_thread_event_sequences.last_seq_id`; the allocator is only a reservation high-water mark and may contain gaps.

### Prune every retention-expired covered event in bounded batches

Replace the marker-dependent pruning predicate with the authoritative snapshot boundary:

- same `user_id` and `org_id` as the snapshot;
- `event.seq_id <= snapshot.latest_event_seq_id`;
- `event.created_at` is strictly before the existing seven-day cutoff.

This must include the former marker row and covered events whose Agent or Chat Thread no longer exists. It must never delete an event above the stored watermark, even if that event is already older than the cutoff.

Keep each cron invocation bounded with a deterministic event-row batch. Select at most the configured/fixed batch size in a stable order, delete only those selected IDs in the same database statement, and return the exact deleted count. Repeated and overlapping invocations must be retry-safe and converge without cross-user/org leakage; concurrent append remains safe because no sequence above the committed snapshot boundary is eligible.

## Most likely implementation trap

Do not recompute an existing watermark from the greatest event that happens to remain after pruning. Once the old marker is deleted, the retained maximum can be lower or null. Writing that value back would make an already-covered cursor appear valid again, or would clear the snapshot boundary entirely. Preserve the stored pair unless a strictly greater real event is applied.

Likewise, do not null `latest_event_id` just because the marker row is gone: the legacy trigger would also null `latest_event_seq_id` and recreate the regression Slice 1 was designed to prevent.

## Non-goals

- No reader, public API payload/status, client cache, or client synchronization changes.
- No schema change, migration, baseline rewrite, trigger change, or production data backfill.
- No change to the seven-day retention duration.
- No use of the sequence allocator as cursor evidence.
- No unrelated Chat content-event, Agent-consolidation, UI, or billing work.
- Do not absorb open PR #31674; it overlaps only the shared BDD file for unrelated billing coverage and is inventory, not a dependency.

## Acceptance criteria

- [x] Deleting the physical marker and running a stale/no-new-event compaction preserves the snapshot's `latest_event_id` and `latest_event_seq_id` exactly, updates the live projection as needed, and reports zero events applied.
- [x] After marker deletion, a later real same-scope event advances both stored boundary fields exactly once; an allocator gap or lower retained event cannot advance or regress them.
- [x] At the exact retention cutoff, rows are retained; immediately after it, every covered row through and including the marker is eligible while every `seq_id > latest_event_seq_id` remains.
- [x] Covered rows with `agent_id = NULL`, a deleted Agent, or a deleted Chat Thread are pruned without requiring those entities to exist.
- [x] A dataset larger than one prune batch deletes no more than the bound per invocation; repeated invocations converge, counts are exact, and retry/overlap does not produce an error or scope leak.
- [x] Equal-watermark reads still return `200` and the strictly ordered newer tail after the marker is gone; older cursors still return the existing `410 CHAT_THREAD_EVENTS_EXPIRED` payload.
- [x] Existing page limits/order, user/org isolation, snapshot projection behavior, and compaction statistics remain covered by focused BDD tests.
- [x] Prettier, API lint, API typecheck, Knip, focused Vitest, and all required PR checks pass.
- [x] The focused PR merges through the protected merge queue with no direct/admin merge or protection bypass.

## Rollout

One focused implementation PR for this issue. The implementation owner stops after GitHub confirms protected-queue merge. The parent controller owns independent acceptance, canonical production release, old-revision drain, and post-release verification after at least one snapshot-compaction/retention cycle.

lancy:
## Summary

- advance existing Chat Thread snapshot boundaries only from strictly newer same-scope lifecycle events while preserving markerless watermarks during projection-only refreshes
- prune retention-expired events through and including the stored snapshot watermark in deterministic, retry-safe batches without requiring live Agent or Thread rows
- cover markerless advancement, allocator gaps, cutoff semantics, orphaned entities, bounded overlap convergence, above-watermark preservation, cursor behavior, and scope isolation with focused BDD scenarios

## Safety

- depends on the released Slice 1 snapshot-watermark reader contract from #32295
- keeps the seven-day retention window and public API behavior unchanged
- does not change schemas, migrations, the legacy compatibility trigger, readers, clients, caches, content events, UI, or billing

## Testing

- Prettier check on all changed files
- pnpm -F api lint
- pnpm -F api check-types
- pnpm knip --workspace api
- focused Chat Thread snapshot, cursor, cutoff, retention, and pagination BDD scenarios: 7 passed
- Chat Thread snapshot pin-order regression file: 4 passed
- final focused Slice 2 scenarios after rebase to latest main: 2 passed

Fixes #32350
Parent: #28636

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/chat-threads.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/chat-threads.bdd.test.ts
@@ -45,7 +45,10 @@ import {
 import {
   deleteChatThreadEventMarkerFixture,
   holdChatThreadEventInsertTransactionFixture,
+  insertCanonicalOrphanChatThreadEventFixture,
   insertChatThreadEventTransactionFixture,
+  readChatThreadEventIdsFixture,
+  setChatThreadSnapshotBoundaryFixture,
   setChatThreadVideoModelFixture,
 } from "../../../test-fixtures/chat-thread-events";
 import { setAgentRunCreatedAtFixture } from "../../../test-fixtures/run-deletion";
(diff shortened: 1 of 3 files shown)


lancy:
LGTM

PR_AUTO_REVIEW_COMMIT: 82796c228eefd758165c0e34f9e75943fc3fdfee
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

The current HEAD safely retires physical snapshot-marker retention after the released Slice 1 reader gate. Existing watermarks advance only from strictly newer same-scope events, conflict upserts cannot regress the stored pair, and retention pruning is bounded, deterministic, scope-exact, retry-safe, and inclusive of the covered marker and orphan rows.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness, concurrency, cutoff, and edge-case safety | Pass | `cron-compact-chat-thread-snapshots.service.ts` uses strict `seq_id > stored watermark` advancement, conflict-safe paired updates, strict `created_at < cutoff`, and `seq_id <= watermark` pruning; focused BDD covers markerless, allocator-gap, overlap, boundary, orphan, and ordered-tail cases. |
| `REVIEW.md` | Security and type safety | Pass | The diff adds no public/auth surface, secrets, casts, or untyped values; SQL values are emitted through Drizzle parameters and both batch settings fail fast on invalid input. |
| `REVIEW.md` | Persisted contract and regression gate | Pass | No schema, JSONB shape, migration, trigger, API status, or payload changes; cursor and projection regressions are exercised through the existing route contracts. |
| `docs/bad-smell.md` | No `any`, suppressions, dynamic imports, hidden errors, or unsafe configuration fallbacks | Pass | All imports are static, no suppressions or `any` are introduced, and the optional tuning value has an explicit positive-integer validator. |
| `docs/testing.md` | Prefer integration coverage with real internals, real DB, boundary mocks, and application-clock control | Pass | Tests invoke the real cron and Chat Thread routes through `setupApp`, use the real database, keep external mocks at their boundary, and use `mockNow` rather than fake timers. |
| `docs/fallback.md` | Do not introduce fallback paths or tombstone tests | Pass | Markerless watermarks are the now-authoritative reachable persisted state established by #32295, not a fabricated fallback; the tests assert live cursor/projection/retention behavior rather than absence of retired code. |
| `docs/event-sourcing.md` | Persistent server ordering remains authoritative | Pass | Snapshot advancement and pruning rely only on committed same-scope event `seq_id`; the allocator high-water mark and UUID order are never treated as cursor evidence. |
| `.claude/skills/ccstate/SKILL.md` | Commands retain lifecycle ownership and do not leak accessors or floating work | Pass | The existing module-scope command still awaits the compaction service, checks its owned signal between non-cancellable phases, and adds no detached or floating promise. |
| `docs/deployment-compatibility.md` | Stage persisted-state compatibility before destructive cleanup | Pass | PR #32295 was released and production-verified before this cleanup; this diff preserves that reader contract and changes no schema or cross-version API shape. |
| `docs/testing/patterns.md` | Exercise API routes through production contracts and assert outcomes | Pass | Both new scenarios call the real cron/read/mutation endpoints and assert returned stats, snapshot projection, cursor status, event order, and convergence. |
| `docs/testing/anti-patterns.md` | Avoid internal mocks, fake timers, service-only tests, and call-count assertions | Pass | No internal module mock, direct fetch mock, fake timer, service import, or mock-call-only assertion is added. |
| `docs/testing/api-testing.md` | Keep route tests externally driven and persistent state uniquely owned | Pass | Every scope uses unique user/org identities. Narrow fixtures are limited to marker/orphan/cutoff states that production APIs cannot construct or physically observe, while behavior is driven and verified through route contracts. |
| `docs/testing/testing-external-behavior.md` | Cross internal boundaries only for genuinely unreachable infrastructure state, with an explicit reason | Pass | The fixture comments name why physical marker deletion, null-Agent history, exact persisted cutoff timestamps, and deletion visibility have no production API surface; the cron itself remains exercised through its external contract. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- Coverage: Adequate. Local focused runs passed 7 snapshot/cursor/retention BDD cases, 4 pin-order projection cases, and the final 2 Slice 2 cases again after rebasing onto current `main`.
- Conventions: Compliant. Prettier, API lint, API typecheck, and API-workspace Knip passed; required PR CI is still running at review time.

PR #32362, opened 2026-09-07
feat(platform): render agent message math

bingjiez666:
## Summary

- Render explicit LaTeX in Agent messages: inline `\\(...\\)` plus own-line `\\[...\\]` and `$$...$$` blocks.
- Tokenize recognized formulas into HAST and render native MathML through one app-wide, lazy KaTeX 0.18.7 loader.
- Leave `$...$` currency/tickers, inline `$$...$$` prose, code spans, and fenced code unchanged.
- Gate the behavior behind the staff-only `AgentMessageMath` feature switch.

## Implementation

- Keep the flow to one parse-local tokenizer, one runtime signal, and one React component; there is no per-formula registry or surface-specific signal wiring.
- Serve the versioned KaTeX runtime from the app's public assets only when recognized math is present. KaTeX code, CSS, and fonts stay out of the application bundle.
- Use `trust: false`, a 10,000-character source limit, and bounded KaTeX expansion and size settings.
- Preserve the exact delimited source while loading and for invalid, empty, oversized, or load-failed formulas.
- Use parse-local nonce markers so model-authored HTML cannot impersonate internal math nodes.

## Verification

- 5 targeted Markdown test files: 23 passed, including all 3 math integration tests.
- Core feature-switch test: 31 passed.
- Platform and Core type checks.
- Targeted Oxlint, type-aware Oxlint, and ESLint for every changed source file.
- Full repository Prettier check.
- Platform static-asset and build-config checks.
- Platform and Core Knip checks.
- Production Platform build; the versioned KaTeX file is copied byte-for-byte and remains outside the main bundle.


bingjiez666:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 059ac03eea223bb72922e102989b270657e4194c

### Summary

This PR adds explicit Agent-message LaTeX tokenization, lazy same-origin KaTeX loading, native MathML rendering, shared-thread wiring, and a staff-only feature switch. The overall shape is contained and the exact-head pipeline is green, but two deterministic P1 correctness paths need repair before merge: feature-switch hydration can leave cached trees permanently unparsed, and the error-marker substring check rejects valid LaTeX.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and edge-case behavior | Finding (P1) | `math-formula.tsx:32` confuses valid source text with KaTeX error markup; `create-chat-thread.ts:2065` and `shared-thread-page-setup.ts:33` can consume the pre-hydration switch value without a reactive invalidation. |
| `REVIEW.md` | Security and type safety | Pass | Math is capped at 10,000 characters; KaTeX uses `trust: false`, bounded expansion/size, MathML-only output, and typed runtime options. No secrets, unsafe casts, or new auth surface were found. |
| `REVIEW.md` | Feature containment and regression coverage | Finding (P1) | `AgentMessageMath` is correctly staff-only and off by default, but tests seed the cache and fetched value to the same state, missing the rollout transition and the valid-marker collision. |
| `docs/bad-smell.md` | No `any`, suppressions, dynamic `import()`, hardcoded environment URL, or hidden configuration recovery | Pass | Changed authored source contains none of those patterns; the runtime is a versioned bundled asset resolved from the deployment asset base. |
| `docs/testing.md` | Integration-first coverage at the real boundary | Finding (P1) | The page-level suite covers the principal delimiter/currency/code behavior, but not the two failing paths described below. |
| `docs/fallback.md` | Avoid undeclared or unjustified compatibility branches | N/A | No temporary cross-version compatibility branch is introduced; raw delimited text while loading or on invalid input is the explicit visible rendering/error behavior, and the feature remains non-GA. |
| `docs/effect.md` | Choose the correct trigger/owner and give async resources cancellation | Pass | The app-wide KaTeX resource is a lazy `computed` owned by `rootSignal$`; script listeners are released through `withCleanup`. |
| `docs/cache.md` | Cache identity, invalidation, and resource teardown must match the owner | Finding (P1) | Trees record `mathEnabled`, but a feature-switch change does not itself invoke the command that compares that identity; the shared-thread snapshot has no invalidation path at all. |
| `.claude/skills/ccstate/SKILL.md` | Use reactive computed state or an explicit semantic command trigger; do not rely on a later incidental read | Finding (P1) | `agentMessageMathEnabled$` is reactive, but both parsing paths turn it into command-local data without a guaranteed re-run when hydration changes it. |
| `docs/deployment-compatibility.md` | New app code must tolerate and correctly converge from an older browser-persisted/fetched shape | Finding (P1) | Existing `v5` feature-switch caches can omit the new key; route setup and switch hydration run concurrently, so the new app can parse with `false` and fail to converge to the fetched `true` value. |
| `docs/testing/patterns.md` | Platform behavior uses a bootstrapped page integration test | Pass | The new tests enter through `setupPage` and exercise the rendered chat page rather than a directly rendered component. |
| `docs/testing/anti-patterns.md` | No internal module mocks, fake timers, direct component rendering, or mock-call-only assertions | Pass | The suite uses the real Markdown/rendering pipeline and a browser-boundary script/runtime intercept; observable text and MathML are asserted. |
| `docs/testing/app-testing.md` | Distinguish cached state from the SWR response when that transition matters | Finding (P1) | `featureSwitches` seeds both cache and response as enabled; no `cachedFeatureSwitches` transition proves reparse after hydration. |
| `docs/testing/testing-external-behavior.md` | Construct state through the page and assert user-observable results | Pass | Formula MathML and preserved source text are asserted through the rendered page; no Store mutation or internal service entry point is used. |

### Findings

#### Critical (P0)

- None.

- `turbo/apps/platform/src/views/components/math-formula.tsx:32`: error detection searches the entire KaTeX markup for the fixed string `okou-katex-error`. KaTeX includes valid source text in both MathML text nodes and its `application/x-tex` annotation, so valid `\text{okou-katex-error}` produces markup containing that string and this branch returns `undefined`; the user sees raw delimiters instead of rendered math. I reproduced this against the vendored 0.18.7 runtime. Detect the actual parse-error structure (or use an exception/result path that cannot collide with source text) and add a valid-marker regression case. This is required by `REVIEW.md` correctness and `docs/testing.md` behavior coverage.
- `turbo/apps/platform/src/signals/chat-page/create-chat-thread.ts:2065` and `turbo/apps/platform/src/signals/shared-thread-page/shared-thread-page-setup.ts:33`: parsing snapshots `agentMessageMathEnabled$`, while `bootstrap.ts` starts route setup and feature-switch hydration concurrently. A browser carrying the existing `vm0:feature-switch-cache:v5` map cannot contain this new key, so the new getter initially returns `false`. If route setup wins, the shared-thread path keeps that value for the page lifetime; the chat path stores it in tree identity but only checks again after an unrelated later ensure/scroll/event command. Enabled staff can therefore keep seeing raw LaTeX until an incidental interaction or reload. Make switch hydration/change an explicit parse invalidation trigger (or await an authoritative switch value before snapshotting), and add page integration coverage with a disabled/missing cached value and an enabled fetched value for both wired surfaces. This violates the invalidation/ownership requirements in `docs/cache.md`, `.claude/skills/ccstate/SKILL.md`, and the cache-to-current-build transition expectations in `docs/deployment-compatibility.md` and `docs/testing/app-testing.md`.

### Testing

- Coverage: Insufficient — add regressions for a valid formula containing the current error-marker text and for cached-disabled/missing to fetched-enabled switch hydration (including shared-thread wiring).
- Conventions: Compliant for the tests present — they use `setupPage`, the real Markdown pipeline, and a browser-boundary runtime intercept without internal mocks, fake timers, or direct component rendering.
- Exact-head evidence: GitHub CI reports the Platform test shards, lint/type/format/Knip gates, CodeQL/Semgrep, and `ci-gate-turbo` successful at `059ac03eea223bb72922e102989b270657e4194c`. The vendored `katex.min.js` SHA-256 (`10a91b479cd927446ceb60409fb0d72b5d0d05eaf446c9e52fafd64058c84540`) matches `katex@0.18.7/dist/katex.min.js` from upstream. No local Vitest suite was run during this review.

bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 3a3b5a60baedb8c8ff4ef0ec3ef70655e31f3321
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

This PR renders explicit LaTeX in Agent messages through parse-local Markdown tokens and a lazy, app-owned KaTeX MathML runtime, while preserving currency, code, disabled-rollout, invalid-input, and load-failure behavior as readable source. The current HEAD also fixes the prior review findings by converging parsed trees after authoritative feature-switch hydration and by identifying actual KaTeX error nodes rather than matching source text.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and edge-case behavior | Pass | `math.ts` recognizes only `\\(...\\)` inline and own-line `\\[...\\]` / `$$...$$`; `math-formula.tsx` preserves raw delimiters for empty, oversized, invalid, loading, or failed formulas. |
| `REVIEW.md` | Security and type safety | Pass | KaTeX is capped at 10,000 source characters with `trust: false`, `maxExpand: 1000`, `maxSize: 100`, and MathML-only output; authored HTML cannot create HAST `data.math`, and no unsafe casts or secrets were added. |
| `REVIEW.md` | Feature containment and regression coverage | Pass | `AgentMessageMath` is off by default, staff-org scoped, and wired only through assistant-event and shared-assistant paths; currency, ticker, code, enabled, and disabled behavior is covered. |
| `docs/bad-smell.md` | No `any`, suppressions, dynamic `import()`, hidden configuration recovery, or environment URL fallback | Pass | Authored changed production code contains none of the prohibited patterns; the versioned runtime URL derives from Vite's deployment asset base. |
| `docs/testing.md` | Feature integration coverage at the real boundary | Pass | `markdown-math.test.tsx` boots complete chat and shared-thread pages and exercises the tokenizer, switch hydration, lazy runtime, renderer, and user-visible output together. |
| `docs/fallback.md` | Avoid undeclared compatibility and impossible-state fallbacks | N/A | No cross-version compatibility path is introduced. Raw delimited text is the declared rendering/error behavior for expected untrusted formula and resource states, not a legacy contract reader. |
| `docs/effect.md` | Match work to its trigger and lifecycle, with cancellation | Pass | Parsing stays in page/thread commands; the lazy browser runtime is app-owned through `rootSignal$`, and script listeners/removal are paired through `withCleanup`. |
| `docs/cache.md` | Cache identity, invalidation, ownership, and teardown match | Pass | Event-tree identity includes both content and `mathEnabled`; the post-hydration command explicitly re-ensures visible trees, while the runtime relies on ccstate memoization at app lifetime. |
| `.claude/skills/ccstate/SKILL.md` | Use computed resources and awaited commands with owned `AbortSignal`s | Pass | The new signal flow keeps accessors inside commands, awaits subcommands, checks cancellation after non-signal awaits, and introduces no signal-layer `detach()`. |
| `docs/deployment-compatibility.md` | Persisted/browser state converges across app rollout | Pass | Existing caches that lack or disable the new key are superseded by the authoritative switch response before immutable shared trees are built and through explicit chat-tree resynchronization. |
| `docs/testing/patterns.md` | Prefer behavior-level integration tests | Pass | Tests use `setupPage`, real Platform routing/rendering, and a browser script boundary rather than testing parser helpers directly. |
| `docs/testing/anti-patterns.md` | No internal mocks, direct fetch mocks, fake timers, or direct component rendering | Pass | The new suite uses no relative `vi.mock`, fetch stubs, fake timers, snapshots, or direct React component mount. |
| `docs/testing/app-testing.md` | Model cached state separately from the SWR response when relevant | Pass | Both chat and shared-thread regressions use `cachedFeatureSwitches: false` with fetched `featureSwitches: true` and assert rendered MathML through the page. |
| `docs/testing/testing-external-behavior.md` | Construct and verify behavior through the page surface | Pass | Assertions verify rendered MathML, preserved source text, code blocks, and the absence of an unnecessary runtime request from complete pages. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- Coverage: Adequate — the four new page integration cases cover chat and shared-thread rendering, stale-cache hydration, valid/error formulas, explicit delimiters, currency/tickers, code, and enabled/disabled runtime loading.
- Conventions: Compliant — full-page setup, boundary-only runtime interception, real timers, and observable output assertions.
- Exact-head evidence: all required checks (`ci-gate-turbo`, `ci-gate-security`, and `ci-gate-crates`) passed at `3a3b5a60baedb8c8ff4ef0ec3ef70655e31f3321`; all eight App test shards, lint/type/format/Knip, CodeQL/Semgrep, build/deploy checks, and Codecov passed. The vendored `katex.min.js` is byte-identical to `katex@0.18.7` from npm with SHA-256 `10a91b479cd927446ceb60409fb0d72b5d0d05eaf446c9e52fafd64058c84540`.

PR #32356, opened 2026-09-07
fix(platform): read annotated images through public urls

bingjiez666:
## Summary

- read image bytes for annotation flattening through the attachment's CORS-enabled public URL
- keep the existing private resource URL behavior unchanged for display consumers
- cover a private attachment whose resource URL fails to fetch while its public URL remains readable

## Root cause

The editor can display a presigned R2 URL in an image element, but annotation flattening must fetch and read the image bytes. The R2 response does not expose the required CORS headers, so the browser failed before canvas rendering and upload. The existing file URL response already provides a public CDN URL for the same object with the required CORS access.

## Verification

- `pnpm -F @okouai/app lint`
- `pnpm -F @okouai/app check-types`
- `pnpm exec knip --workspace @okouai/app --no-progress`
- `pnpm -F @okouai/app exec vitest run src/views/okou-page/__tests__/chat-attachment-annotation.test.tsx --silent=passed-only` (6 tests passed)

PR #32478, opened 2026-09-08
fix(platform): bypass opaque cache for annotated images

bingjiez666:
## Summary

- bypass an opaque no-CORS preview cache entry when reading an image for annotation flattening
- add page-level regression coverage for previewing and flattening the same public image URL
- preserve the standard `Request.cache` value in the Happy DOM test boundary so the cache-sensitive path is testable

## Root cause

A plain cross-origin `` preview can cache an opaque response. A later CORS `fetch()` of the same URL may reuse that unreadable cache entry and fail with `TypeError: Failed to fetch`, before image decoding, canvas rendering, or annotated upload begins.

This is a follow-up to #32356. For fresh uploads, `resourceUrl` and `shareUrl` can be identical, so selecting the public URL does not change the browser cache key. The flattening read now uses `cache: "no-store"` to force the required CORS network request.

## Verification

- `pnpm vitest --run apps/platform/src/views/okou-page/__tests__/chat-attachment-annotation.test.tsx` (7 passed)
- regression test fails with `TypeError: Failed to fetch` when `cache: "no-store"` is removed and passes when restored
- `pnpm --filter @okouai/app check-types`
- targeted Oxlint, type-aware Oxlint, and ESLint checks
- repository pre-commit hooks: Prettier, Knip, full monorepo type checks, and file-size check

### DIFF
--- a/turbo/apps/platform/src/signals/__tests__/test-mocks.ts
+++ b/turbo/apps/platform/src/signals/__tests__/test-mocks.ts
@@ -578,6 +578,9 @@ export function createTestMocks(getSignal: () => AbortSignal) {
       ): ImageDimensionsMock => {
         return mockImageDimensions(getSignal(), results);
       },
+      requestCacheMode: (): void => {
+        mockRequestCacheMode(getSignal());
+      },
       canvasRendering: (): CanvasRenderingMock => {
         return mockCanvasRendering(getSignal());
       },
@@ -1450,6 +1453,33 @@ function mockCanvasRendering(signal: AbortSignal): CanvasRenderingMock {
   });
 
   return { clipCircles, imageDraws, renders };
+}
+
+function mockRequestCacheMode(signal: AbortSignal): void {
+  const NativeRequest = window.Request;
+  class RequestWithCacheMode extends NativeRequest {
+    constructor(input: RequestInfo | URL, init?: RequestInit) {
+      super(input, init);
+      Object.defineProperty(this, "cache", {
+        configurable: true,
+        enumerable: true,
+        value:
+          init?.cache ??
+          (input instanceof NativeRequest ? input.cache : "default"),
+      });
+    }
+  }
+
+  // Happy DOM does not expose Request.cache yet. Preserve it so HTTP boundary
+  // mocks can cover cache-sensitive browser behavior without mocking fetch.
+  const descriptor = defineWindowProperty(
+    window,
+    "Request",
+    RequestWithCacheMode,
+  );
+  restoreOnAbort(signal, () => {
+    restoreWindowProperty(window, "Request", descriptor);
+  });
 }
 
 function defineWindowProperty(
--- a/turbo/apps/platform/src/signals/okou-page/flatten-annotated-image.ts
+++ b/turbo/apps/platform/src/signals/okou-page/flatten-annotated-image.ts
@@ -82,7 +82,10 @@ async function loadImage(
   url: string,
   signal: AbortSignal,
 ): Promise<HTMLImageElement> {
-  const response = await fetch(url, { signal });
+  // The preview renders this URL through a plain <img>, which can cache an
+  // opaque no-CORS response. Bypass that entry so this read performs the CORS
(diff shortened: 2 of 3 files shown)


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 13a983746fc0185621130a3c2155c4eed57f6bf6
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

This focused fix bypasses an opaque response cached by the plain cross-origin image preview before annotation flattening performs its CORS byte read. The page-level regression test exercises the complete annotate-and-attach flow and verifies the composer becomes sendable without a retry state.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness, edge cases, and API use | Pass | `flatten-annotated-image.ts` uses the standard `RequestCache` mode `no-store`, retains the caller-owned `AbortSignal`, and preserves the existing non-OK response failure path. |
| `REVIEW.md` | Security and type safety | Pass | The diff adds no input parsing, auth boundary, secret, unsafe cast, or `any`; the browser API types remain explicit. |
| `REVIEW.md` | A `fix:` change has regression coverage | Pass | `chat-attachment-annotation.test.tsx` fails with the reported `TypeError: Failed to fetch` when `no-store` is removed and passes with the fix. |
| `docs/bad-smell.md` | Avoid suppressions, hidden failures, and unnecessary fallback paths | Pass | No suppression, dynamic import, hardcoded environment URL, or production fallback was added; failed network and HTTP reads still fail visibly. |
| `docs/fallback.md` | Add only justified, declared compatibility fallbacks | N/A | No production compatibility branch is introduced. The test-only default mirrors the Fetch standard for an optional `RequestInit.cache` member. |
| `docs/testing.md` | Prefer integration tests and mock only at system boundaries | Pass | The regression enters through the full page and uses the HTTP/browser/upload boundaries while exercising the real annotation signals and flattening implementation. |
| `docs/testing/patterns.md` | Use page setup, `context.mocks`, MSW, and lifecycle-owned helpers | Pass | The test configures `context.mocks` before `setupPage`; the HTTP response is provided through MSW and the Happy DOM capability shim is owned by the test signal. |
| `docs/testing/anti-patterns.md` | Avoid direct fetch/internal mocks, fake timers, and component-only tests | Pass | The diff contains no `vi.mock`, fetch stub, fake timer, or direct component render; assertions cover the user-visible send and retry states. |
| `docs/testing/app-testing.md` | Drive one page-visible action at a time and assert observable outcomes | Pass | The test opens the preview, annotates, attaches once, then waits for editor closure, mark count, enabled Send, and absence of the failure action. |
| `docs/testing/testing-external-behavior.md` | Construct and observe behavior through the page boundary | Pass | The production flow is reached through preview and annotation controls, and success is observed from the composer rather than internal Store or signal state. |
| `docs/effect.md` | Match async work to its owner and preserve cancellation | Pass | The existing flattening command remains the owner of the read and passes its `AbortSignal` directly into `fetch`; no render or effect work is added. |
| `docs/cache.md` | Avoid unbounded caches and pair resources with lifecycle cleanup | Pass | Production bypasses the problematic browser cache instead of introducing another cache; the test-only `Request` replacement is restored on `context.signal` abort. |
| `.claude/skills/ccstate/SKILL.md` | Keep signal-layer promises and abort ownership explicit | Pass | The fetch stays awaited inside the signal path and retains the final positional signal; no `detach` or orphaned async work is introduced. |
| `docs/deployment-compatibility.md` | Cover cross-version API or persisted-state boundaries | N/A | This changes a same-build browser resource fetch only; it does not alter an API contract, persisted shape, service worker, or independently deployed protocol. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- Coverage: Adequate. The regression test covers the reported same-URL preview/flatten path and has explicit red/green evidence.
- Conventions: Compliant. Local annotation coverage passes 7/7; the unrelated thread-shortcut file that timed out once in CI also passes 7/7 locally, and its failed CI job is being rerun.

PR #32480, opened 2026-09-08
fix(mitm-addon): bound unicode path normalization work

### LINKED ISSUE #32434: perf(python): bound Unicode normalization work in path safety checks
## Problem

At audited `main` commit `6f4073c2b15a34f06649088269628a1c362a77b1`, a request path below the existing validation limit can spend almost a second in one synchronous path-safety check. Long sequences of Unicode combining marks with alternating canonical combining classes trigger expensive canonical reordering inside NFKC normalization.

Relevant code:

- [`crates/runner/mitm-addon/src/path_security.py:13-27`](https://github.com/vm0-ai/vm0/blob/6f4073c2b15a34f06649088269628a1c362a77b1/crates/runner/mitm-addon/src/path_security.py#L13-L27) admits paths up to 65,536 characters and passes each non-ASCII segment to the compatibility-validation pipeline.
- [`crates/runner/mitm-addon/src/path_security.py:52-79`](https://github.com/vm0-ai/vm0/blob/6f4073c2b15a34f06649088269628a1c362a77b1/crates/runner/mitm-addon/src/path_security.py#L52-L79) calls `unicodedata.normalize("NFKC", segment)` at line 76 without a bound on normalization work or combining-sequence length, including on decoded intermediate segments.
- [`crates/runner/mitm-addon/src/matching.py:1772-1793`](https://github.com/vm0-ai/vm0/blob/6f4073c2b15a34f06649088269628a1c362a77b1/crates/runner/mitm-addon/src/matching.py#L1772-L1793) performs this check after a base matches and before the endpoint permission decision.
- [`crates/runner/mitm-addon/src/request_classification.py:487-506`](https://github.com/vm0-ai/vm0/blob/6f4073c2b15a34f06649088269628a1c362a77b1/crates/runner/mitm-addon/src/request_classification.py#L487-L506) invokes the matcher synchronously. It is reachable from the header-phase classification at `src/mitm_addon.py:791-814` and the request hook at `src/mitm_addon.py:1357-1384`.

The existing character limit bounds input size but does not make Unicode canonical reordering linear. An alternating `U+0315` / `U+0300` sequence (combining classes 232 / 230) remains expensive well below that limit. This delays other flows and event-loop callbacks before a request can be rejected; upstream acceptance and successful credential injection are not required.

All diagnostics were local, using the repository-pinned CPython 3.14.0 and locked mitmproxy 12.2.3. No production traffic was sent.

From `crates/runner/mitm-addon`, with the locked environment synchronized:

```bash
PYTHONPATH=src uv run --no-sync python - <<'PY'
import statistics
import time
from path_security import has_unsafe_path

for pairs in (4096, 8192, 16000):
    for label, marks in (
        ("ordered", "\u0300" * pairs + "\u0315" * pairs),
        ("alternating", "\u0315\u0300" * pairs),
    ):
        path = "/a" + marks
        times = []
        for _ in range(3):
            started = time.perf_counter()
            assert has_unsafe_path(path) is False
            times.append((time.perf_counter() - started) * 1000)
        print(label, len(path), len(path.encode()), statistics.median(times))
PY
```

Median elapsed time per call, measured without profiling:

| Path characters | UTF-8 path bytes | Already ordered marks | Alternating marks |
| ---: | ---: | ---: | ---: |
| 8,194 | 16,386 | 0.286 ms | 55.539 ms |
| 16,386 | 32,770 | 0.559 ms | 220.197 ms |
| 32,002 | 64,002 | 1.094 ms | 839.807 ms |

Doubling the combining sequence approximately quadruples the cost. A separate `cProfile` run attributed essentially all 0.870 seconds to `unicodedata.normalize`.

Reachability was also verified with real mitmproxy request objects and the actual addon hooks:

1. Parse `GET <UTF-8 path> HTTP/1.1` plus `Host: api.example.com` with mitmproxy's `read_request_head()`; assert the parsed path exactly equals the 32,002-character input above. The raw path is only 64,002 bytes.
2. Use the existing `real_flow`, `mitm_ctx`, `_write_registry`, and `_single_firewall_sandbox` test helpers with a registered sandbox, root API base `https://api.example.com`, empty auth/permissions, and `unknownPolicy: deny`.
3. Run `mitm_addon.requestheaders(flow)` followed by `await mitm_addon.request(flow)`, measuring a `loop.call_soon` callback scheduled immediately beforehand.

For alternating marks, header handling took **855.139 ms**, request handling took **855.313 ms**, and the callback ran **1,710.554 ms** late. The request returned **403**. The ordered control returned the same status with **6.617 ms** total callback delay. No auth, DNS, or upstream network operation was needed in this fixture.

The 299 tests in the sampled test files and 86 additional focused tests passed. Existing tests cover overall length, nested percent decoding, and compatibility syntax but do not exercise long, adversarial combining sequences.

## Scope and deduplication

- Closed #26832 bounded total path length to address repeated, fixed-pass percent-decoding work on oversized inputs. This finding is quadratic normalization work **inside** that accepted boundary.
- Closed #31350 avoided eagerly segmenting above-budget paths. These inputs are below budget and contain one non-empty segment, so that fix does not help.
- Open issue searches for Unicode, path normalization, and NFKC found no matching issue.

## Acceptance criteria

- Establish a deterministic work boundary before expensive normalization of request-controlled segments, including percent-decoded intermediate forms. Reject unsupported cases with the existing fail-closed path decision, or use an equivalently safe bounded-work validation strategy.
- Preserve path-security checks for malformed escapes, invalid UTF-8, dot segments, encoded separators/backslashes, matrix parameters, compatibility characters, and over-depth encodings. Do not solve the problem by omitting normalization or by splitting Unicode at unsafe arbitrary boundaries.
- Preserve ordinary supported Unicode/percent-encoded paths and the ASCII fast path; explicitly document any new rejection boundary for pathological inputs.
- Add structural regression coverage showing that a below-total-limit adversarial combining sequence cannot reach an unbounded normalization operation, plus an actual matcher/request-hook case. Keep wall-clock thresholds out of correctness tests.
- Record before/after diagnostics for ordered and alternating combining marks and ordinary paths. Verify that both request phases have bounded work without changing authorization outcomes for supported inputs.

seven332:
## Summary

Closes #32434.

- Bound consecutive non-starters to 30 after per-code-point NFKD decomposition, before whole-segment NFKC. Reject unsupported input through the existing unsafe-path decision.
- Apply the same bound at every percent-decoding stage, including the final permitted stage; preserve the plain-ASCII fast path and existing whole-segment syntax validation.
- Add 20 real requestheaders/request integration cases covering structural normalization bounds, raw/encoded/mixed paths, expansion and class-zero decomposition, exact accepted boundaries, long Unicode, and unchanged successful auth/path behavior.
- Document the boundary and rollout behavior. No URL rewriting, new dependencies, caches, async tasks, API/schema changes, Rust changes, or production traffic.

## Validation

All checks ran sequentially in the foreground:

- `scripts/prepare.sh`
- `uv lock --check` and `uv sync --locked`
- `uv run --no-sync ruff format --check .` and `uv run --no-sync ruff check .`
- `uv run --no-sync basedpyright -p .`: zero errors/warnings
- `uv run --no-sync python -m pytest tests/ -q --tb=short`: **7,349 passed**, no skips (43 dependency/generated-fixture warnings)
- Prettier parser validation for the full updated guide and format validation for the added section
- `git diff --check`

The initial new test helper/annotation mistakes were corrected before these passing results. Unrelated Rust/Turbo application suites were not run locally; cross-language checks remain CI-owned.

## Local before/after diagnostics

Compared the baseline helper from `5d80a5a5ff` with this branch in the same pinned CPython 3.14.0 environment. Medians are wall-clock observations on a contended local machine, not test thresholds or latency guarantees (three samples for mark/hook cases; 100 for ordinary direct-path cases).

| Direct path input | Characters | Before ms | After ms |
| --- | ---: | ---: | ---: |
| Ordered marks | 8,194 | 0.2493 | 0.0243 |
| Alternating marks | 8,194 | 122.7314 | 0.0438 |
| Ordered marks | 16,386 | 2.6646 | 0.0473 |
| Alternating marks | 16,386 | 421.3024 | 0.0581 |
| Ordered marks | 32,002 | 0.9451 | 0.0847 |
| Alternating marks | 32,002 | 1,822.8889 | 0.5391 |
| ASCII | 30 | 0.0011 | 0.0012 |
| Ordinary Unicode | 11 | 0.0045 | 0.0065 |
| Encoded Unicode | 29 | 0.0144 | 0.0118 |
| Long Chinese text | 32,001 | 0.4092 | 9.7999 |

Raw ordered/alternating inputs above the new bound now return unsafe; ordinary inputs retain their previous safe result. For the largest alternating input through real hooks, header/request/callback-delay medians changed from **1,689.01 / 1,898.32 / 3,589.63 ms** to **1.09 / 2.83 / 5.34 ms**. Both versions returned 403 under the diagnostic deny policy, without auth or upstream traffic. Committed allowing-policy tests separately verify unchanged success for supported input. Structural assertions, not these noisy timings, prove that rejected segments cannot reach whole-segment normalization in either phase.

## Risks and compatibility

- Formerly accepted sequences above 30 decomposed non-starters now fail closed, even when already ordered. This is based on the [UAX #15 stream-safe input boundary](https://www.unicode.org/reports/tr15/#Stream_Safe_Text_Format), without inserting CGJ or otherwise changing URLs.
- A new linear Unicode scan costs more for giant ordinary Unicode paths, as shown above. This deliberately preserves long starter-separated paths instead of applying a much shorter Unicode-segment length limit. Plain-ASCII code is unchanged.
- Shared base/auth-rewrite/platform-path validation inherits the same boundary. No wire or persisted-data format changes; old runners retain the old pathological-input behavior until drained.
- Full deployed-runner behavior and the current head's CI status remain to be verified by the PR pipeline. No merge is requested.

### DIFF
--- a/crates/runner/mitm-addon/src/path_security.py
+++ b/crates/runner/mitm-addon/src/path_security.py
@@ -10,6 +10,8 @@ _DOT_SEGMENTS = {".", ".."}
 _HEX_DIGITS = frozenset("0123456789abcdefABCDEF")
 _MAX_PERCENT_DECODE_PASSES = 5
 _PERCENT_ESCAPE_LENGTH = 3
+# UAX #15 stream-safe limit, measured after compatibility decomposition.
+_MAX_CONSECUTIVE_NONSTARTERS = 30
 # Match the addon's existing 64 KiB request-work scale. Character count bounds
 # Python string processing without first scanning and allocating encoded bytes.
 MAX_PATH_VALIDATION_CHARACTERS = 64 * 1024
@@ -72,11 +74,32 @@ def _segment_has_unsafe_path(raw_segment: str) -> bool:
 def _segment_has_unsafe_syntax(segment: str) -> bool:
     if _segment_has_unsafe_syntax_parts(segment):
         return True
+    if _segment_exceeds_normalization_budget(segment):
+        return True
 
     normalized = unicodedata.normalize(_COMPATIBILITY_NORMALIZATION_FORM, segment)
     return normalized != segment and (
         "%" in normalized or _segment_has_unsafe_syntax_parts(normalized)
     )
+
+
+def _segment_exceeds_normalization_budget(segment: str) -> bool:
+    if segment.isascii():
+        return False
+
+    nonstarters = 0
+    for char in segment:
+        # Decompose only one code point at a time: whole-segment NFKD would
+        # already perform the potentially quadratic canonical reordering.
+        # Raw combining classes miss expansions such as U+0F73 and U+FF9E.
+        for decomposed in unicodedata.normalize("NFKD", char):
+            if unicodedata.combining(decomposed):
+                nonstarters += 1
+                if nonstarters > _MAX_CONSECUTIVE_NONSTARTERS:
+                    return True
+            else:
+                nonstarters = 0
+    return False
 
 
 def _segment_has_unsafe_syntax_parts(segment: str) -> bool:
--- a/crates/runner/mitm-addon/tests/test_request_path_normalization_budget.py
+++ b/crates/runner/mitm-addon/tests/test_request_path_normalization_budget.py
@@ -1,0 +1,135 @@
+"""Unicode work bounds through real header and request classification hooks."""
+
+import json
+import unicodedata
+import urllib.parse
+from pathlib import Path
+from typing import Literal
+
+import pytest
+
+import flow_metadata_keys as metadata_keys
+import mitm_addon
+from tests.request_handler_helpers import _single_firewall_sandbox, _write_registry
+
+
+@pytest.fixture
+def path_firewall_registry(tmp_path: Path) -> Path:
+    return _write_registry(
+        tmp_path,
+        sandbox_info=_single_firewall_sandbox(
+            tmp_path,
+            api_entry={
+                "base": "https://api.github.com",
+                "auth": {"headers": {"Authorization": "Bearer ${{ secrets.GITHUB_TOKEN }}"}},
+                "permissions": [{"name": "full-access", "rules": ["ANY /{path+}"]}],
+            },
+            network_policy={
+                "allow": ["full-access"],
+                "deny": [],
+                "ask": [],
+                "unknownPolicy": "allow",
+            },
+        ),
+    )
+
+
+@pytest.mark.parametrize(
+    ("segment", "decode_passes"),
+    [
+        pytest.param("a" + "\u0315\u0300" * 16000, 0, id="long-alternating"),
+        pytest.param("a" + "\u0300" * 16000 + "\u0315" * 16000, 0, id="long-ordered"),
+        pytest.param("a" + "\u0315\u0300" * 4096, 1, id="long-percent-encoded"),
+        pytest.param("\u0300" * 31, 0, id="leading-nonstarters"),
+        pytest.param("a" + "\u0344" * 16, 0, id="decomposition-expansion"),
+        pytest.param("a" + "\u0f73" * 16, 0, id="class-zero-decomposition"),
+        pytest.param("a" + "\uff9e" * 31, 0, id="compatibility-nonstarter"),
+        pytest.param("\u00e9" + "\u0300" * 30, 0, id="precomposed-trailing-mark"),
+        pytest.param("a" + "\u0315\u0300" * 16, 2, id="nested-percent-encoded"),
+        pytest.param("a" + "\u0f73" * 16, 5, id="final-decode-pass"),
+        pytest.param("a" + "\u0300" * 15 + "%CC%80" * 16, 0, id="mixed-raw-and-encoded"),
+    ],
+)
+async def test_path_normalization_budget_blocks_before_unbounded_work(
+    path_firewall_registry,
+    real_flow,
+    mitm_ctx,
+    fake_firewall_headers,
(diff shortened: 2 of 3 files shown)


seven332:
## Code Review: PR #32480

**Reviewed head:** `bf202bcdc3e74c6c156d178dcd853de534c44367`

### Summary

Completed six focused self-review passes and the repository code-quality review on the unchanged submitted head. The guard bounds compatibility-decomposed non-starter runs before whole-segment normalization at every decoding stage, without replacing NFKC syntax validation or rewriting paths.

### Findings

No findings. The deliberate rejection of more than 30 decomposed non-starters and the additional linear scan for giant ordinary Unicode paths are documented design trade-offs. No new dependency, cache, IO, async task, protocol, or persisted-state change is introduced.

### Testing

Ruff formatting/lint, BasedPyright, normal commit hooks, Markdown validation and diff checks passed. Full addon suite: **7,349 passed**, including 20 real requestheaders/request regressions; no skipped tests. Structural assertions verify single-code-point decomposition and rejection before whole-segment normalization in both phases, alongside unchanged successful auth/path behavior for supported Unicode. Local before/after diagnostics are recorded in the PR; they are not CI timing thresholds. Full deployed runner and current-head CI status remain separate verification gates.

### Verdict

LGTM for the reviewed head. No merge performed or authorized.

PR #32488, opened 2026-09-08
test(runner): exercise contended baseline observations

### LINKED ISSUE #32437: test(rust): exercise parallel callers in baseline observation test
## Summary

`execute_job_serializes_concurrent_baseline_observations` does not run the baseline observer concurrently. It joins two futures in the same task, and each observer call completes synchronously before its future yields.

The production observer currently holds one mutex across observation and replacement. This issue concerns missing regression protection for that atomicity, not an observed production race.

## Evidence

Reviewed on `main` at `6f4073c2b15a34f06649088269628a1c362a77b1`.

- [`crates/runner/src/executor/tests/telemetry_tests.rs:1185-1212`](https://github.com/vm0-ai/vm0/blob/6f4073c2b15a34f06649088269628a1c362a77b1/crates/runner/src/executor/tests/telemetry_tests.rs#L1185-L1212): the test uses the default current-thread Tokio runtime and `tokio::join!` on two `execute_cancelled_observation` futures sharing one executor configuration. It asserts the sorted outcomes are `["first", "same"]`.
- [`crates/runner/src/executor/tests/telemetry_tests.rs:295-315`](https://github.com/vm0-ai/vm0/blob/6f4073c2b15a34f06649088269628a1c362a77b1/crates/runner/src/executor/tests/telemetry_tests.rs#L295-L315): the helper awaits `execute_job` directly; it does not dispatch the observation onto an independent task or thread.
- [`crates/runner/src/executor/mod.rs:620-662`](https://github.com/vm0-ai/vm0/blob/6f4073c2b15a34f06649088269628a1c362a77b1/crates/runner/src/executor/mod.rs#L620-L662): execution records the baseline synchronously at lines 621-623, before reaching the downstream await.
- [`crates/runner/src/executor/storage_baseline_observation.rs:75-85`](https://github.com/vm0-ai/vm0/blob/6f4073c2b15a34f06649088269628a1c362a77b1/crates/runner/src/executor/storage_baseline_observation.rs#L75-L85) documents the concurrent-entry atomic observe-and-replace contract. [`observe_marked`, lines 164-228](https://github.com/vm0-ai/vm0/blob/6f4073c2b15a34f06649088269628a1c362a77b1/crates/runner/src/executor/storage_baseline_observation.rs#L164-L228) performs the lookup, comparison, and replacement synchronously under one mutex.

`tokio::join!` polls both branches in one task. Consequently, one branch finishes its synchronous observation before the other branch can enter that operation. Changing only the runtime to `flavor = "multi_thread"` would not fix this: the joined branches would still share one task.

For example, a regression that separates the lookup and replacement into distinct critical sections would continue producing `["first", "same"]` in this serial test. Real parallel calls could instead both observe an absent previous value and report `["first", "first"]`. The neighboring baseline tests exercise sequential transitions, not contending callers.

## Impact

The test's name and assertions imply protection against concurrent observer updates, but they cannot detect a loss of atomicity. Such a regression would corrupt baseline-stability/count telemetry. This observer does not drive storage selection or execution decisions.

## Acceptance criteria

- Exercise the same production observer from genuinely independent worker tasks or threads sharing its state; a focused observer-level test is sufficient.
- Use explicit coordination to exercise contending/overlapping observer entry, with bounded completion and cleanup. Do not rely on sleeps, repeated stress runs, or `tokio::join!` alone.
- For two identical nonempty candidate sets with the same observation key, assert exactly one `first` and one `same` outcome and a coherent subsequent observation.
- Demonstrate that loss of atomic observe-and-replace is detected by the test.
- Preserve useful executor telemetry integration coverage, but describe any retained single-task test as sequential coverage.

## Audit scope

Static control-flow review and repository test cross-checks; no test suite or mutation test was run during this audit. Open issues were searched for the exact test name and baseline observation coverage; open PRs were checked for baseline work.

seven332:
## Summary

- Preserve the two-job executor integration scenario as explicitly sequential coverage; `tokio::join!` did not make its synchronous observer calls concurrent.
- Exercise the real baseline observer and telemetry from two independent OS threads, pausing the first after its missing-key lookup and verifying the second encounters the state lock before releasing the first.
- Assert exactly one `first`, one `same`, and coherent subsequent telemetry/counts. Use instance-local test-only probes, bounded channel waits and scoped thread cleanup; keep HTTP-client setup outside the contention window.

This closes a regression-coverage gap, not an observed production race. All new observer fields and calls are behind `cfg(test)`; the production single-mutex algorithm, storage decisions, telemetry schema, dependencies and CI configuration are unchanged. A future synchronization redesign will need equivalent atomicity coverage in place of these mutex-specific probes.

## Validation

- `CARGO_INCREMENTAL=1 cargo clippy --profile local --all-targets --all-features`
- `CARGO_INCREMENTAL=1 cargo doc --profile local --workspace --all-features --no-deps`
- `CARGO_INCREMENTAL=1 cargo test --profile local -p runner --bin runner executor::tests::telemetry_tests -- --test-threads=4`: 30 passed, including all three baseline tests; 0 failed/ignored.
- `git diff --check`

Mutation check: temporarily dropped the state guard after the missing-key lookup and reacquired it before insertion, without changing the new test. `CARGO_INCREMENTAL=1 cargo test --profile local -p runner --bin runner baseline -- --nocapture --test-threads=1` produced 2 passes (the sequential tests) and 1 expected failure: the new concurrency test reported `BeforeLock { contended: false }` instead of `true`. Removed the mutation, verified both source hashes against their intended pre-mutation versions, and reran the 30 telemetry tests successfully. The mutant is not included in this PR.

Incremental compilation was enabled only through command-local environment settings. Full Rust tests and runner E2E remain for the normal CI selection; unrelated Turbo/Python suites were not run locally.

Closes #32437

### DIFF
--- a/crates/runner/src/executor/storage_baseline_observation.rs
+++ b/crates/runner/src/executor/storage_baseline_observation.rs
@@ -82,9 +82,28 @@ struct Observation {
 /// persisted, hashed, or converted to stable pseudonyms.
 pub(crate) struct StorageBaselineObserver {
     previous: Mutex<HashMap<ObservationKey, Vec<CandidateIdentity>>>,
+    #[cfg(test)]
+    test_probe: Option<Box<dyn Fn(BaselineObservationTestEvent) + Send + Sync>>,
+}
+
+#[cfg(test)]
+#[derive(Debug, PartialEq, Eq)]
+pub(super) enum BaselineObservationTestEvent {
+    BeforeLock { contended: bool },
+    BeforeFirstInsert,
 }
 
 impl StorageBaselineObserver {
+    #[cfg(test)]
+    pub(super) fn with_test_probe(
+        probe: impl Fn(BaselineObservationTestEvent) + Send + Sync + 'static,
+    ) -> Self {
+        Self {
+            test_probe: Some(Box::new(probe)),
+            ..Self::default()
+        }
+    }
+
     /// Record the current baseline-candidate observation and its bounded telemetry dimensions.
     ///
     /// Only read-only manifest entries with `baseline_candidate` set are considered. Their
@@ -167,11 +186,24 @@ impl StorageBaselineObserver {
         candidates: Vec<CandidateIdentity>,
     ) -> Observation {
         let candidate_count = candidates.len();
+        #[cfg(test)]
+        if let Some(probe) = &self.test_probe {
+            // Drop any acquired guard before the probe coordinates another caller.
+            let contended = matches!(
+                self.previous.try_lock(),
+                Err(std::sync::TryLockError::WouldBlock)
+            );
+            probe(BaselineObservationTestEvent::BeforeLock { contended });
+        }
         let mut previous_by_key = self
             .previous
             .lock()
             .unwrap_or_else(std::sync::PoisonError::into_inner);
         let Some(previous) = previous_by_key.get_mut(&key) else {
+            #[cfg(test)]
+            if let Some(probe) = &self.test_probe {
+                probe(BaselineObservationTestEvent::BeforeFirstInsert);
+            }
             previous_by_key.insert(key, candidates);
             return Observation {
                 outcome: "first",
--- a/crates/runner/src/executor/tests/telemetry_tests.rs
+++ b/crates/runner/src/executor/tests/telemetry_tests.rs
@@ -1,6 +1,6 @@
 use std::collections::HashMap;
 use std::path::Path;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex, mpsc};
 use std::time::{Duration, Instant};
 
 use async_trait::async_trait;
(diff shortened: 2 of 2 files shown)


seven332:
## Code Review: PR #32488

**Reviewed head:** `8e7498269f40a26a438f0990c0d488be3290fde0`

### Summary

Completed six clean focused self-review passes and a code-quality review of the submitted diff, observer callers, telemetry buffering, coordination/cleanup and existing regression coverage. This adds deterministic contention coverage without changing the production observer algorithm or introducing a runtime/deployment boundary change.

### Findings

No findings. The try_lock guard and one-shot gate mutex are released before coordination waits; controller failure disconnects the first worker, and workers are joined before outcome assertions. The mutex-specific probe coupling is documented in the PR as a future-maintenance consideration.

### Testing

- Final telemetry suite: 30 passed, 0 failed/ignored.
- Formatting, full-workspace Clippy/documentation and repository commit hooks passed.
- Temporary split-lock mutation: both sequential baseline tests passed; the new test failed specifically because the second entry was not contended. The mutation was removed and the submitted source hashes match the restored, passing version.
- Full Rust tests and runner E2E were not run locally; CI/merge readiness remains a separate gate.

### Verdict

LGTM for this head. No follow-up issue needed.

PR #32544, opened 2026-09-08
docs(runner): attribute storage helper resume latency

### LINKED ISSUE #32428: perf(runner): attribute residual storage helper invocation latency
## Current status — September 8, 2026

Controlled investigation identified the sampled cold/resumed storage launch wait as **child migration through `cgroup.procs` waiting for an RCU grace period**. Subsequent actual-RPC experiments now demonstrate a concrete optimization candidate: **direct creation with `clone3(CLONE_INTO_CGROUP)`**.

- The latest three-way default/favordynmods/clone3 comparison completed **2,916/2,916 formal paired storage calls** and **27/27 reusable park/unpark cycles**. The separate smoke and all earlier cohorts retain separate denominators below.
- All nine resumed outer observations per candidate were **20.134–31.536 ms** by default, **4.938–9.344 ms** with favordynmods, and **2.970–7.560 ms** with the storage-only clone3 prototype.
- clone3 warm p90 improved, but warm p50 increased **0.344–0.716 ms**. This regression, broader launch coverage and raw-clone safety remain implementation work, not solved findings.
- The final real-cgroup storage lifecycle matrix passed **27/27 checks**; separate bounded fork/exec fixtures did not establish a material favordynmods regression, but do not prove zero cost for actual Agents.
- **Implementation follow-up: #32647**, covering the shared guest-control-server contained-process launch layer, not only storage. This issue and the new implementation issue both remain direct sub-issues of **#24203**.
- Historical production cohorts, initial 3,564 pairs, additional 2,268 K/L-stage pairs, C mechanism screen and later RPC cohorts are not pooled. These experiments do not establish production-wide causality or guaranteed API-to-spawn savings.
- #32544 remains closed without merging as an earlier experiment archive. No production code or service was changed; this update records research and the requested implementation issue. This issue remains open.

## Parent and delivery boundary

Parent: #24203. This remains the bounded attribution/decision slice for residual storage-helper invocation cost on low-work Exact reuse. The controlled mechanism and candidate experiments are recorded here; shared-launcher implementation is tracked separately in sibling #32647, also directly under #24203. No outstanding hard prerequisite. Do not treat implementation completion as part of this research issue or move either issue under the other.

## Production evidence

Observation window: `[2026-09-06T16:10:00Z, 2026-09-07T16:10:00Z)` (Beijing September 7 00:10 through September 8 00:10). The same fixed successful-spawn run IDs were joined to raw child events and summarized using nearest-rank percentiles.

| Exact API / Runner                                      | Spawns / complete inner joins | Runner invocation p50/p90 | Guest download_total p50/p90 | Per-run outer-minus-inner p50/p90/p95/p99 |
| ------------------------------------------------------- | ----------------------------: | ------------------------: | ---------------------------: | ----------------------------------------: |
| `9fd53c47daba5922b0a5de5abcb4b82f44f0a2f8` / `0.187.1`  |                     500 / 498 |                  25/38 ms |                       2/6 ms |                            22/32/35/46 ms |
| `41f45b81e22051e69546a60ee795efe274b3ca3c` / `0.186.15` |                     432 / 431 |                  25/39 ms |                       2/8 ms |                            23/32/35/40 ms |

Outer durations describe all selected outer events; residuals use only complete same-run pairs. The residual is not a difference of percentiles, pure transport time or guaranteed savings. Missing child events are not zero-filled. Frequent releases require separate exact cohorts; there is no causal version comparison or fleet failure-rate claim. Ingestion time was not frozen.

## Existing implementation and original uncertainty

#18749 delivered small-manifest stdin invocation; #28732 delivered the bounded dedicated storage operation. Do not reimplement either.

At inspected main `03a305e4cd2986d94634ad031db679d94418478b`, `guest_storage_manifest.rs` executes the fixed `guest-storage-apply` helper with Workload cgroup/resource containment, owned input/output workers, timeout/cancellation, child wait, cleanup and a terminal response. The outer-minus-inner difference includes work outside the guest's download_total timer; the original production telemetry could not assign it to process creation, resource-policy setup, input/initialization, scheduling, drains or cleanup.

The 50-ms cancellation polling constant does not prove a fixed 50-ms delay: the existing fast-child wait path already has coverage against that assumption.

### Decision and scope

Select `clone3(CLONE_INTO_CGROUP)` for a separate implementation issue, #32647. This issue remains the research/evidence record. Both are direct sub-issues of #24203; do not nest the implementation under this attribution issue or close/reparent this issue as part of this update.

The implementation scope is the shared guest-control-server contained-process launch boundary: dedicated storage, ordinary Workload exec including oversized storage fallback, and controlled Agent startup. It is not a permanent storage-only launcher or a repository-wide replacement of fork. Existing target cgroups, identity/security, ownership, cancellation, descendant cleanup and park gates must remain intact.

The measurements below validated a **storage-only experimental launcher**. They do not yet validate generic exec, controlled Agent startup, or Guest Agent's internal CLI launcher with clone3. The broader shared-launcher implementation needs its own entry-point and real-cgroup coverage. Guest Agent's internal Tokio launch/placement-capability protocol and the tool launcher's migrate-self-then-exec boundary are intentionally deferred.

Do not enable `favordynmods` alongside this implementation, introduce a resident launcher or reuse cgroup pools. This records an implementation direction and acceptance requirements, not a production deployment or merge approval. The earlier provisional favordynmods-first recommendation in the local candidate report predates the real-RPC clone3 results and is superseded by this decision.

### Candidate mechanism screen

A separate single-threaded C probe compared fresh/reused cgroups with fork/migration or direct clone3, plus a resident launcher. Five order rotations completed **2,800/2,800 storage invocations**, **150/150 expected error/timeout/nonzero guards**, and **10/10 whole-cgroup resident-kill observations**. Per candidate: 495 repeated calls, five first-after-idle calls and sixty spaced-idle calls. Idle means a 100-ms wait, **not park/unpark**.

Per-trial outer p90 ranges over the five rotations:

| Candidate                               | Repeated p90   | Spaced-idle p90  |
| --------------------------------------- | -------------- | ---------------- |
| Fresh groups + fork/migration           | 1.686–2.997 ms | 23.642–27.422 ms |
| Reused empty groups + fork/migration    | 0.750–1.513 ms | 20.949–23.881 ms |
| Fresh groups + clone3                   | 1.713–3.009 ms | 1.559–2.013 ms   |
| Reused empty groups + clone3            | 0.778–1.538 ms | 1.153–1.564 ms   |
| Resident launcher + inherited placement | 0.888–1.588 ms | 1.343–1.802 ms   |

Only twelve spaced-idle samples contribute to each trial's idle p90. Reuse reduces setup cost but does not remove the migration wait. Fresh-cgroup clone3 avoids that wait without persistent state. Resident initialization is outside per-call timers, and whole-domain kill also kills the resident. Keeping a resident or empty group across park would change the existing quiesce contract. The C probe uses memfd I/O, not the Rust threaded RPC path, and cannot alone establish production readiness.

### Actual storage RPC comparison

Use isolated fresh VMs on local-11.gcp.vm3.ai, concurrency one, guest Linux **6.1.155**, 2 vCPU and 4096 MiB. Rust product source is `16d1e9a9b2bcd80e20ae37dfd64584814c46b531`; experiment workspace HEAD is `a99899c4165c03382e5ce365300684f11ac6a0b2`. Only the clone3 guest-init changes the typed storage launcher; the other eight guest binaries remain baseline. Default and favordynmods use the original guest image and share the same experimental host Runner.

A preliminary clone3 smoke VM completed 324 successful pairs and three reusable cycles, retained separately. The formal cohort uses nine fresh VMs, rotating order:

- P1: default, favordynmods, clone3.
- P2: favordynmods, clone3, default.
- P3: clone3, default, favordynmods.

Every formal VM runs 324 calls: one first cached call; three batches of 100 cached calls plus park/unpark and one resumed call; ten high-work calls with 140 payloads; ten oversized generic fallbacks. All **2,916/2,916 formal calls** have complete successful outer/inner pairs, valid content/payloads and fallback cleanup, and no output truncation. All **27/27 formal park/unpark cycles** pass the unchanged reusable gate. Including the separate smoke gives 3,240 calls and thirty cycles; do not combine these totals with earlier experiments.

Milliseconds, preserving trial identity:

| Observation                               | Default P1 / P2 / P3       | favordynmods P1 / P2 / P3  | clone3 P1 / P2 / P3         |
| ----------------------------------------- | -------------------------- | -------------------------- | --------------------------- |
| Warm outer p50, 300 samples/trial         | 3.499 / 3.611 / 3.419      | 3.618 / 3.945 / 4.147      | 3.843 / 4.071 / 4.135       |
| Warm outer p90                            | 6.031 / 6.558 / 6.207      | 6.307 / 6.450 / 6.327      | 5.736 / 5.849 / 5.934       |
| First cached call, one/trial              | 24.483 / 33.446 / 29.140   | 15.206 / 14.588 / 13.953   | 11.795 / 12.509 / 12.298    |
| High-work p90, ten samples/trial          | 119.461 / 119.057 / 93.236 | 96.253 / 129.303 / 107.072 | 109.474 / 144.248 / 142.320 |
| Oversized fallback p90, ten samples/trial | 39.312 / 48.037 / 38.044   | 23.365 / 26.097 / 24.958   | 39.943 / 41.086 / 31.852    |

All resumed observations, not a stable production percentile:

| Trial | Default                  | favordynmods          | clone3                |
| ----- | ------------------------ | --------------------- | --------------------- |
| P1    | 20.919 / 20.441 / 20.488 | 9.344 / 6.477 / 5.582 | 7.560 / 4.898 / 2.970 |
| P2    | 30.617 / 20.134 / 31.536 | 6.680 / 6.190 / 6.254 | 6.177 / 5.350 / 3.167 |
| P3    | 24.309 / 25.387 / 23.835 | 6.830 / 4.938 / 7.460 | 5.454 / 5.452 / 5.822 |

clone3 now has actual storage RPC/resume evidence: resumed observations decrease from **20.134–31.536 ms** to **2.970–7.560 ms**. Its warm p90 improves, but **warm p50 increases by 0.344–0.716 ms** in adjacent comparisons. The prototype duplicates standard Command and raw argv/environment/identity preparation; removing duplication is a proposed investigation, not a measured remedy. The implementation must address this regression.

The measured clone3 prototype leaves generic oversized fallback unchanged; no improvement is claimed there. High-work observations are variable, with no demonstrated throughput win. Only nine resumed observations per candidate on one development host are available. These results do not establish fleet prevalence, stable production p90, API-to-spawn savings or zero resource regressions.

An earlier, separate actual-RPC default/favordynmods cohort completed **1,944/1,944 pairs and 18/18 reusable cycles**. Default resumed observations were 18.872–34.918 ms versus 3.856–7.791 ms with favordynmods; warm p90 was 5.926/6.097/6.425 versus 6.525/6.563/6.075 ms. Keep that cohort separate from the later three-way comparison.

### Lifecycle/error and concurrent-work checks

A release-built probe drives public `guest_control_server::handle_connection` with the actual cgroup backend and production sandbox-user identity in fresh disposable VMs. The final matrix passes **27/27 checks**, nine in each default/favordynmods/clone3 cohort:

1. Missing executable returns StartFailed.
2. Fixed argv/stdin/environment, UID/GID/groups/home, workload membership and CPU/memory/PID/OOM policy match; no cgroup capability FD survives exec.
3. Nonzero exit and stderr are preserved.
4. Stdout/stderr capture limits and truncation flags are exact.
5. Invalid executable format returns StartFailed.
6. Timeout cleans a setsid descendant.
7. Natural helper exit cleans a surviving setsid descendant.
8. Connection loss kills descendants and removes operation groups.
9. Busy rejection preserves ownership, followed by correct disconnect cleanup.

Direct children are reaped; descendants must be absent or already zombies, never live. After each case the cgroup set equals the pre-case set. This covers selected storage boundaries, not all Agent/generic exec behavior or an exhaustive security/kernel-failure audit.

Retain the calibration failure: the first executed guard probe, `guards-clone31.log` (v2), failed before any case completed because the client omitted the initial READY frame. Fixing the test client to consume READY produced the final v3 results; production launcher code was not changed for that failure. The failed log is not counted as a pass. A prior v1 fixture cgroup-prefix mistake was corrected before that image ran.

For favordynmods' guest-wide fork/exit trade-off, a separate four-worker fixture used six fresh VMs in B1/D1/D2/B2/B3/D3 order. All **30,240 workload child operations plus 288 workers** completed successfully, including warmups; the measured subset is 29,520 workload operations. There were no memory/OOM/PID-limit events. The existing 1.9-CPU quota throttled workers, and raw CPU-stat deltas are retained.

Adjacent favordynmods-versus-default changes in total measured parallel-batch wall time:

| Workload                            | Trial 1 | Trial 2 | Trial 3 |
| ----------------------------------- | ------- | ------- | ------- |
| fork/exit/wait                      | -2.1%   | -0.6%   | -1.2%   |
| exec /bin/true                      | -2.2%   | -0.9%   | -0.6%   |
| exec small Python JSON/hash program | -2.7%   | +0.5%   | +0.6%   |

An earlier separate sequential fixture completed 30,600 children including warmups; mean wall time per measured fork/exit/wait was 828.66/788.44/784.39 microseconds by default versus 830.00/830.78/780.75 with favordynmods. Neither fixture proves zero kernel overhead or representative Agent equivalence; interpreter cost and CPU quotas may mask smaller effects. Actual concurrent migration with Agent/tool workloads remains unmeasured. The clone3 prototype does not change these generic fork workloads.

### Kernel/API and implementation risks

`clone3` starts with Linux 5.3; the required `CLONE_INTO_CGROUP` starts with **Linux 5.7 and requires cgroup v2**. The tested guest 6.1.155 supports it; no kernel upgrade is needed for that tested image. Verify supported production guest builds and syscall restrictions during implementation planning, rather than using host kernel version as evidence.

The target must be passed as a cgroup directory FD, not the existing write-only `cgroup.procs` descriptor. Placement is specified before child creation; it cannot be retrofitted solely inside `Command::pre_exec`. Shared launcher integration must explicitly preserve multithreaded post-clone safety, signal dispositions/masks, stdio/extra FD handling, credentials and process-inspection ordering, startup-error reporting, owned child identity, cancellation/reaping and cleanup. Unsupported/denied operations must never silently run uncontained.

Sources: [clone3 and CLONE_INTO_CGROUP](https://man7.org/linux/man-pages/man2/clone.2.html), [cgroup v2 mount trade-offs](https://docs.kernel.org/6.1/admin-guide/cgroup-v2.html#mounting), [Rust pre_exec safety](https://doc.rust-lang.org/std/os/unix/process/trait.CommandExt.html#tymethod.pre_exec).

### Reproduction identities, artifacts and final state

Local reports and raw data are under `/workspaces/vm2/codex-work/research/issue-32428-resumed-spawn/`. They are **local audit artifacts, not committed files in #32544**. The durable counts, values, method and decision are recorded above.

- Reports: `candidate-experiments.md` (earlier screen) and `entrypoint-comparison.md` (latest actual-RPC/lifecycle/concurrency evidence).
- Patches: `runner-dynmods.patch`, `storage-clone3-entry.patch`, `storage-clone3-guards.patch`. The last contains probe-only `ISSUE_32428_STORAGE_FORK`, absent from the measured performance binary and forbidden in the production implementation.
- Formal performance raw files: `entry-p<N>-<mode>.jsonl`, modes `base`, `dynmods`, `clone3`; smoke: `entry-smoke.jsonl`. Each contains 651 records: 324 outer, 324 inner, two resource observations and one mount record. `entry-results.json` uses the existing strict 324-pair summary validator.
- Guard files: `guards-clone32.jsonl`, `guards-base2.jsonl`, `guards-favordynmods2.jsonl` (eleven records each); failed calibration tail: `guards-clone31-failed-tail.log`.
- Other raw families: `candidates-m1..m5.jsonl` (604 records/trial), `dynmods-base<N>.jsonl` / `dynmods-dynmods<N>.jsonl` (651), `fork-base<N>.jsonl` / `fork-dynmods<N>.jsonl` (seven), and `concurrent-base<N>.jsonl` / `concurrent-dynmods<N>.jsonl` (fourteen), with separate aggregate JSON files.
- Common experimental host Runner `runner-dynmods` SHA-256: `96639a3ee69c5394c8f7b68553de01218c30522a859ea3f9312cb3aca9f6fc88`.
- Original rootfs/snapshot: `203c5532218d920b96643368e9176fd106a6b249d8f342df322fa102e5c3200f` / `17209ec83aaf0265273510d120ade287e07c1fe02b375d0810fbb473080e47c8`.
- Performance clone3 guest-init SHA: `8b7f49d5c4f361cfd15108be6e57139365674024c18b118e9099970d8a6fdeaf`; rootfs/snapshot: `549024755e871a7ba17cbcded956a72a99cf9c6ba5b3b5da09411c0b10bd9c2f` / `3a710277a0665a4c5ef6581ca46008818a3c0e0068e2a4cdfcace01d658cf612`.
- Guard v3 binary SHA: `58ec8e79fd1b004a1a32af0fffa52ff2bcc28cbcaf8e5b0a3cc3af92df6127b9`; rootfs/snapshot: `aa8b3e977c507884ccfff1365c4ca2cd43e2ac46d4c90fd3f4f0c959fee01fb0` / `435942b4568a17bf97e0212676f7d711073cf1ef9e9708c87c6e9aa450d3202d`. Only the mock-Claude image slot contains this probe; no Agent run uses it.
- Build/config files: `build-storage-clone3.sh`, `clone3-runner.yaml`, `build-storage-clone-guards-v3.sh`, `clone-guards-v3-runner.yaml`. Full remote logs remain under `/var/lib/vm0-runner/bin/issue-32428-attribution-vm2/` on the approved dev host. Formal logs are `entry-p<N>-<mode>.log`; smoke is `entry-clone-smoke.log`. Other full logs match their raw-file family names.

Relevant checks passed while experimental source was active: Rust formatting, guest-control-server all-targets Clippy, documentation for guest-control-server/guest-init, x86_64 musl CI-profile builds, Bash syntax/ShellCheck, and strict validation of all ten latest performance logs plus guard/concurrent count/order/outcomes. Guest-control-server local tests passed 183 unit + 120 connection + two harness cases. Three pre-existing ignored production-identity tests require CI account/bootstrap and were not run in this container; no skip was added, and the independent guest matrix exercised real storage identity/cgroups. Guest-init passed 25 tests, with its existing isolated child invoked by its passing parent.

All temporary active Rust source was restored, both clone3 patches remained check-applicable, and the worktree was clean at unchanged HEAD after experiments. No runtime implementation, new PR, commit, production deployment or service replacement occurred. All benchmark VMs/COW/network/proxy owners tore down normally, including failed probe calibration. Both existing development services remained active; no experimental benchmark process remained. Inactive binaries/configs/logs and image caches remain for audit; no broad GC ran. This GitHub update and the new implementation issue are the user-requested follow-up.

## Continued investigation: cgroup placement and RCU — September 8, 2026

These additional measurements supersede the initial report's unresolved inner-`spawn()` attribution. They use the same approved isolated development-host scope, source `16d1e9a9b2bcd80e20ae37dfd64584814c46b531`, Runner `0.188.7`, common experimental Runner, fixture and `vm0/default` capacity described below. Workspace HEAD remains `a99899c4165c03382e5ce365300684f11ac6a0b2`; new raw observations and disposable patches were not committed.

### Method and coverage

Seven additional complete trials (`J4, K1, J5, K2, J6, K3, L1`) produced
**2,268/2,268 successful complete outer/child pairs**. Every trial retains the
previous 324-invocation order and content/fallback/output/quiesce/teardown
checks. No storage invocation failed. Each J or K set includes 900 repeated
cached calls, three first calls, nine resumed calls, 30 high-work calls and
30 oversized fallbacks. L1 is a separate 324-call stack-sampling trial.

J reuses the previous eight-field guest timing image. K adds three fixed
post-fork records: before placement, after placement, and after identity.
The parent records before/after spawn. Each record contains guest monotonic
microseconds, thread CPU microseconds, minor/major faults and voluntary/
involuntary context switches. Child records cross an already-open private
socket using fixed stack buffers and raw syscalls; no allocation or lock is
introduced inside the diagnostic pre-exec hooks. All original placement,
identity, process group and ownership behavior is retained.

All 942 dedicated K observations have the expected five records, six integer
counters per record, valid clock order and monotonic same-thread counters.
Oversized fallback has a different transport and no dedicated spawn detail.
The complete outer/inner sample counts and expected shape/phase/batch/index
order were validated separately for each J/K file.

### Cold/resumed phase attribution

Milliseconds; each cell lists the three resumed observations in trial order:

| Trial | Whole spawn              | Placement interval       | Credential interval | Placement thread CPU |
| ----- | ------------------------ | ------------------------ | ------------------- | -------------------- |
| K1    | 25.807 / 9.428 / 24.713  | 23.869 / 8.252 / 23.417  | .033 / .166 / .035  | .096 / .059 / .055   |
| K2    | 23.361 / 23.323 / 14.294 | 21.789 / 21.771 / 13.360 | .039 / .045 / .030  | .061 / .087 / .057   |
| K3    | 16.413 / 17.674 / 13.128 | 14.155 / 14.780 / 9.865  | .160 / .121 / .040  | .040 / .228 / .093   |

The nine resumed placement intervals took **8.252–23.869 ms** but consumed
only **40–228 microseconds of child thread CPU**. All nine have zero added
minor/major faults and one additional voluntary plus one involuntary context
switch in that interval. Before-placement elapsed was .400–2.030 ms;
after-identity through parent spawn return was .320–1.318 ms. These boundaries
include the small diagnostic record overhead and scheduler effects; they are
not exact syscall CPU measurements.

K first-call placement took 15.572 / 13.435 / 17.682 ms, also with zero added
faults. This behavior is not exclusive to resume or proof of balloon causality.

### Repeated-call distributions and measurement limits

Nearest-rank p50/p90/p95/p99, milliseconds; 300 observations per row:

| Trial | Probe outer                    | Whole spawn                   | Placement                   |
| ----- | ------------------------------ | ----------------------------- | --------------------------- |
| J4    | 3.853 / 6.378 / 7.405 / 8.120  | .646 / 1.966 / 2.184 / 2.739  | not split                   |
| K1    | 4.044 / 6.533 / 7.221 / 9.318  | .717 / 2.027 / 2.434 / 2.901  | .034 / .188 / .310 / .498   |
| J5    | 4.240 / 6.260 / 7.067 / 9.796  | .715 / 1.921 / 2.183 / 2.822  | not split                   |
| K2    | 3.816 / 6.896 / 7.687 / 12.382 | .641 / 2.264 / 2.634 / 8.129  | .034 / .267 / .365 / .630   |
| J6    | 4.119 / 6.225 / 6.744 / 8.163  | .609 / 1.791 / 2.157 / 2.719  | not split                   |
| K3    | 4.157 / 6.861 / 7.588 / 13.541 | .690 / 2.099 / 2.618 / 11.418 | .038 / .186 / .348 / 10.602 |

J4 ended at 07:00:14 UTC, K1 at 07:03:04, J5/K2/J6/K3 at
07:06:42/07:06:53/07:07:03/07:07:14. The order alternates images but includes
build/analysis gaps; it is not a randomized or uninterrupted equivalence trial.
Adjacent-trial absolute outer p50 difference is at most .424 ms and p90 at
most .671 ms. Tail variation is larger, and K3 has slow placement even among
repeated calls. Do not assert zero perturbation or that only first calls wait.
J is itself instrumented; comparison is the added K observation overhead, not
overhead relative to a completely uninstrumented production binary.

### Kernel-stack confirmation, separate L1 diagnostic cohort

L1 ended at 07:21:33 UTC. A bounded parent-side observer sampled only the first
and three resumed helper invocations. It reads that storage worker's own
`/proc/self/task/<tid>/children`, then the child's `wchan` and `stack` at roughly
1-ms intervals, at most eight records per invocation. The observer is stopped
and joined, including on error. It never modifies cgroups, priorities, affinity,
balloon targets, resource limits or the child security sequence.

All **32/32 sampled stacks across four invocations** contained:

```text
write(cgroup.procs)
  cgroup_file_write
  cgroup_procs_write
  __cgroup_procs_write
  cgroup_procs_write_start
  rcu_sync_enter
  synchronize_rcu
  __wait_rcu_gp
```

Every sampled `wchan` was `__wait_rcu_gp`; no stack-read error was observed.
L1 placement elapsed was 11.974 ms for the first call and
9.737 / 14.906 / 14.703 ms after resume. These are trace-cohort measurements,
not part of the K distributions, and 32 snapshots are not 32 independent calls.
The observer changes scheduling and perturbs timings, so this cohort establishes
the blocked kernel path rather than a production latency distribution.

### Source explanation

The configured guest kernel is `6.1.155` (not the host's `6.17` kernel).
In that kernel, process migration through `cgroup.procs` takes the threadgroup
write lock via `cgroup_attach_lock`, and its `percpu_down_write` enters RCU
synchronization. This matches the captured stack; some intermediate calls may
be inlined. Sources:
[cgroup migration locking](https://github.com/gregkh/linux/blob/v6.1.155/kernel/cgroup/cgroup.c#L2474),
[per-CPU rwsem](https://github.com/gregkh/linux/blob/v6.1.155/kernel/locking/percpu-rwsem.c#L209).

When synchronization state is idle, `rcu_sync_enter` waits for a grace period;
closely spaced calls can reuse its non-idle state without paying that wait
again. This provides a source-backed mechanism consistent with the observed
cold-versus-repeated pattern; it does not establish file-cache warming as the
cause or isolate what determines the exact wait duration. Source: [RCU synchronization state machine](https://github.com/gregkh/linux/blob/v6.1.155/kernel/rcu/sync.c#L98).

Rust's fork/exec parent waits for the exec handshake, so the child's cgroup
placement wait remains inside `Command::spawn()`. Source:
[pinned Rust standard library](https://github.com/rust-lang/rust/blob/88d9e12ae178fab0fb5cc050a94da85685d449ea/library/std/src/sys/process/unix/unix.rs),
[pre-exec ordering and safety contract](https://doc.rust-lang.org/std/os/unix/process/trait.CommandExt.html#tymethod.pre_exec).

On this storage path, `guest_storage_manifest.rs` selects **Workload**, not Agent.
`process_containment.rs::write_self_to_cgroup` writes `b"0"` to `cgroup.procs`
and retries only EINTR. Placement precedes the `user.rs`
`setgroups/setgid/setuid` hook. Agent's `PR_SET_DUMPABLE` denial hook is
**not enabled for Workload**; it is not a storage-launch cost to optimize.

The installed Rust `1.98.0` source rejects `posix_spawn` when pre-exec callbacks
exist; this path uses fork/exec with a parent close-on-exec handshake.
Separately, `sandbox.rs::unpark_inner` resumes vCPUs and requests balloon target
zero without waiting for physical deflation. That overlap is verified lifecycle
behavior, **not evidence that ballooning causes the RCU wait**.

The previous statement that cgroup _creation and cleanup_ are small is still
correct, but it must not be generalized to cgroup _placement_. Placement was
inside the previously unresolved spawn interval. The new evidence does not
prove that all fleet residuals have this cause, nor that the wait can simply
be removed without affecting safety or costs elsewhere.

### Continued-research reproduction and audit artifacts

Local research directory: `/workspaces/vm2/codex-work/research/issue-32428-resumed-spawn/`.
The following new raw files, patches and report are **local audit artifacts, not
files in the closed PR or its GitHub archive**. The measurements, method,
identities and conclusion are preserved in this issue so it is self-contained.

- `j4.jsonl`, `k1.jsonl`, `j5.jsonl`, `k2.jsonl`, `j6.jsonl`, `k3.jsonl`,
  `l1.jsonl`: 650 records each (324 outer, 324 inner, two resource observations).
- `results.json`: J/K nearest-rank distributions and individual cold observations.
- `spawn-detail.patch`: disposable K changes against the unchanged product source.
- `spawn-stack.patch`: disposable L changes including the bounded observer.
- `build-guest.sh`: isolated build recipe, currently selecting the L image;
  refuses to overwrite an existing experiment binary/log. Secrets are read from
  the normal local env file and never written into this research directory.

Use the prior report's common experimental Runner and fixtures, with the K/L
guest-init substitutions below. Normal `runner build` produced unique images;
all other eight guest binaries remain unchanged. Apply these patches only in a
disposable checkout for explicitly approved synthetic experiments. These are
not production patches or proposed PR changes.

| Image | guest-init SHA-256                                                 | Rootfs                                                             | Snapshot                                                           |
| ----- | ------------------------------------------------------------------ | ------------------------------------------------------------------ | ------------------------------------------------------------------ |
| K     | `985012df5d0400d95e437bb117859389a39f2bc8f47543a33a21e5e3d044f3f0` | `083e620b96d1a65700b850dcb4a1de90acbe6e4d68e62404fce01594bccc2ba0` | `45cc3310d3b954d0ef57513f5eb02804c4e23e4e983d38e0b68c77d870e2b6ca` |
| L     | `1b943affa6749fca88a6ddf18a369fadd49c9575ffccd45896e9a1eff4adfdae` | `73ef9230525440d8527c86c9d1026036c2b05de12bb7407b610e5f7aaaf1486c` | `8f3b3ce2bab3c161113f3ce3a597eb842a41a52a17c5e718adab70f42108dab5` |

An initial diagnostic image was built before correcting two Clippy array-chunk
style findings. It was never benchmarked and has no samples in these results.
K/L hashes refer to the corrected, checked source. Remote configs are
`issue-32428-spawn-detail-v2-vm2` and `issue-32428-spawn-stack-vm2` under the
standard runners directory. All raw logs remain under the existing
`/var/lib/vm0-runner/bin/issue-32428-attribution-vm2/` experiment directory.

Example independent calculation of resumed placement and child CPU deltas:

```bash
jq -s '[.[] | select(.kind == "storage_probe" and .phase == "resumed")
  | (.diagnostic | fromjson).spawn_detail
  | {placement_us: (.after_placement[0] - .before_placement[0]),
     placement_cpu_us: (.after_placement[1] - .before_placement[1]),
     added_minor_faults: (.after_placement[2] - .before_placement[2]),
     added_major_faults: (.after_placement[3] - .before_placement[3])}]' k1.jsonl
```

### Continued-research checks and final state

Both K and L passed formatting, guest-control-server all-targets/all-features
Clippy, the eight storage connection integration tests and the optimized musl
guest-init build. This retains nonzero-exit, output bounds, cancellation,
timeout, connection loss, concurrent rejection and descendant cleanup coverage.
K detail counters and J/K order/coverage were checked as described above; L1
also has 324/324 successful untruncated pairs and the expected 32 RCU stack
samples. Both archived patches pass `git apply --check` against the pinned source.

All temporary active Rust edits were removed with scoped patches. Formatting
and the same eight connection tests passed again after restoration. The tracked
worktree is clean, HEAD unchanged; no commit, PR or production
deployment was made during continued research. This issue update is the
user-requested delivery of those results.

All seven foreground benchmark invocations completed normally and released
their VM, COW, namespace and proxy owners. Final host inspection found no
remaining issue-32428 benchmark process, both pre-existing development services
active, and 28,868 MiB available memory. Only inactive experiment binaries,
configs, bounded logs and normally cached images remain for audit; no host-wide
GC, live service replacement or unrelated cleanup was performed.

### Initial decision and how the new evidence refines it

The initial A/I/G/J1–J3 report below remains the baseline and reproduction
record. It reproduced a large first-after-resume cost, with `Command::spawn()`
taking **9.188–24.815 ms** on nine resumed observations while preparation took
**6–7 microseconds**. Repeated spawn p90 in those J trials was **1.831–1.970 ms**;
cgroup creation and cleanup were much smaller.

At that checkpoint, the interval did not distinguish fork/clone, page faults,
placement, credentials, exec handshake and scheduling. **The later K/L evidence
above narrows the sampled blocked path to cgroup placement waiting for RCU.**
The original broad uncertainty must not be read as the current conclusion.

The provisional 5-ms warm-phase p90 screen was not met across the initial
trials. Small cold cohorts do not establish a stable p90, production prevalence
or a safely recoverable saving. K3 also has slow placement in the repeated-call
tail. At that initial checkpoint, no optimization with a demonstrated gain had
been tested or selected. The later candidate/RPC section above supersedes that
checkpoint with measured clone3 gains and a separate implementation direction.

Do not weaken Workload containment, skip mandatory storage work, introduce a
daemon, or change cancellation/balloon policy on the strength of these residuals.
Any behavior change requires its own evidence and design review, including
memory, isolation, cleanup and tail-latency costs. Parent performance work is
not completed by this attribution, and the issue stays open.

## Boundaries

| Observation                                              | Includes                                                                                                                                            | Excludes / interpretation                                                                                                                                                       |
| -------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Production `runner_storage_manifest_guest_storage_apply` | `download_storages`: serialization, runtime-path construction, dedicated operation or fallback, result checks                                       | Prior cache population/staging and other spawn phases                                                                                                                           |
| Probe `outer_us`                                         | Manifest serialization, the same public Sandbox operation and await; fallback includes remove/write/exec                                            | Synthetic reconstruction, **not the production metric**: no production context/path construction, logging, or failure-cleanup wrapper; result validation occurs after the timer |
| Dedicated result `guest_ms`                              | Guest `run_manifest`, from before cancellation/cgroup creation through child handling, containment cleanup, input/output joins, and result assembly | Request parsing, worker admission/dispatch, response encoding, writer-lock wait and socket write                                                                                |
| Child `download_total`                                   | Input read, stale manifest cleanup, parsing, storage work, normalization                                                                            | Runtime log-path setup and argument parsing before the timer; final telemetry/logging and process exit after it                                                                 |
| `outer − guest`                                          | Serialization, host/guest admission and scheduling, transport and response handling                                                                 | **Not pure network latency**; guest timer is rounded down to integer milliseconds                                                                                               |
| `outer − inner`                                          | Paired residual around application work                                                                                                             | Not guaranteed savings and never a subtraction of percentile values                                                                                                             |

Temporary diagnostic intervals partition guest wall time: cancellation/cgroup
creation; command preparation plus spawn; I/O worker setup; wait through
group cleanup/reap; containment cleanup; remaining stdin/drain joins and result
handling. The refined version splits preparation from spawn without changing
the boundary of their combined interval. `prepare_spawn_us` is their **enclosing
total**, not an additional additive phase.

Input writing and output drains overlap child execution. The child application
timer overlaps I/O setup/wait; do not add it to those intervals. The first-call
wait interval also includes initialization, terminal logging, and scheduling
outside the child timer. No absolute timestamp is subtracted across host/guest
clocks. Quantized residuals retain their signed values; missing events are not
zero-filled. Fallback `guest_ms` describes its final generic exec, not all three
fallback operations, and is not the dedicated middle boundary.

## Environment and workload

- Source: `16d1e9a9b2bcd80e20ae37dfd64584814c46b531`, Runner `0.188.7`.
  All guest binaries use this source; only guest-init receives the experimental
  guest-control timing patch in diagnostic images.
- Build: Rust `ci` profile, `x86_64-unknown-linux-musl`, optimized thin LTO,
  four codegen units, no `test-support` or debug assertions. This is the normal
  development Runner build profile, **not** the production full-LTO release.
- Approved shared development host: `local-11.gcp.vm3.ai`, kernel
  `6.17.0-1018-gcp`, 16 logical CPUs (Xeon Platinum 8581C), 32,086 MiB RAM.
  Roughly 29 GiB was available. This is not a production host or a dedicated
  machine; unrelated development services were not stopped or replaced.
- Profile: `vm0/default`, 2 vCPUs, 4,096 MiB guest memory. Normal production
  Workload cgroups, identity drop, security hooks, process-group ownership,
  output bounds, and lifecycle cleanup remain enabled.
- Existing `runner benchmark` owns the proxy, VM, COW devices and namespace pool.
  It prefetches snapshot memory. Every trial starts a new snapshot-restored VM;
  every 100-call batch is followed by normal `park()` / `unpark()` with the
  default balloon controller. This is a lifecycle analogue, not a complete
  API-admitted Exact run with an active agent and its working set.
- Logs use the production home-based runtime directory, not tmpfs. No real
  manifests, customer data, provider calls, or real API token are used. R2
  credentials were supplied only to the ordinary image builder to read its
  cached template, never to measurement output.

Each trial has 324 storage invocations: one first call; three batches of 100
repeated cached calls with one resumed call after each batch; ten high-work
calls; ten oversized calls. The cached fixture has 17 cached mounts, one with
instruction normalization still requested. The archive fixture has 140 local
`file://` mounts, each extracting a 4-KiB payload. The oversized cached fixture
exceeds the 65,536-byte dedicated stdin limit via a 70,000-character synthetic
name and exercises remove/write/generic-exec fallback. Its child removes the
canonical temporary manifest. Instruction content, all extracted payload sizes,
fallback manifest removal, successful exit and untruncated output are checked.

Archives warm during each trial; high-work results are not a fixed cold-cache
distribution. Fixtures and their reset/order are identical between images.
The actual planner's true-no-work case is checked separately by the existing
Runner entry-point test and has **no helper invocation**, not a zero-ms helper
sample. Existing storage integration coverage validates normalization, nested
mount ordering and extraction security beyond this synthetic fixture.

## Evidence and distributions

Formal runs: `A1, I1, A2, I2, A3, I3` alternated at 05:35:54–05:36:47 UTC;
`G1/G2` were started together at 05:41:04–05:41:15;
`J1–J3` refined the spawn split at 06:02:10–06:02:36.
A/G use uninstrumented guests; I uses six wall intervals; J adds the two
preparation/spawn subintervals. Do not pool these builds or lifecycle categories.

All **3,564 / 3,564** formal invocations have successful complete child pairs,
with no truncated captures. Each A/I/J set has 900 repeated cached, 3 first,
9 resumed, 30 high-work and 30 oversized observations; G has 600/2/6/20/20.
An earlier 324-call calibration used tmpfs logs and is excluded. A subsequent
resource-tool availability check failed before the first storage invocation:
guest `command -v strace` returned 127, not 1. It was fixed before formal runs.
That failed setup is not counted as a successful trial or silently rerun away.

Nearest-rank distributions below are **p50 / p90 / p95 / p99 in milliseconds**.
Every residual is calculated per pair before percentile selection. Each row
contains 300 repeated cached calls, preserving individual trial identity.

| Trial | Probe outer                    | Guest result   | Child inner   | Outer − inner                  | Outer − guest                |
| ----- | ------------------------------ | -------------- | ------------- | ------------------------------ | ---------------------------- |
| A1    | 4.120 / 6.326 / 7.277 / 8.026  | 3 / 5 / 6 / 7  | 0 / 1 / 1 / 2 | 3.974 / 6.318 / 7.130 / 8.026  | .974 / 1.355 / 1.480 / 2.900 |
| A2    | 3.899 / 6.306 / 7.332 / 13.038 | 3 / 5 / 6 / 10 | 0 / 0 / 1 / 1 | 3.764 / 6.282 / 7.332 / 10.038 | .909 / 1.296 / 1.383 / 1.744 |
| A3    | 3.796 / 5.978 / 6.730 / 8.488  | 3 / 5 / 6 / 7  | 0 / 0 / 1 / 1 | 3.746 / 5.834 / 6.526 / 8.488  | .869 / 1.317 / 1.390 / 1.546 |

The three 100-call batch outer p90s are A1 `4.500 / 7.277 / 6.318`,
A2 `4.016 / 7.431 / 6.589`, A3 `3.882 / 6.191 / 6.444` ms. Warm calls after
resume are not identical to the pre-park batch; they were not discarded.

Small lifecycle cohorts are reported as individual values, **not stable tails**:

| Trial | First outer | Three first-after-resume outer observations |
| ----- | ----------: | ------------------------------------------- |
| A1    |      30.760 | 34.544 / 18.811 / 35.014                    |
| A2    |      24.761 | 25.567 / 24.523 / 21.278                    |
| A3    |      35.066 | 21.677 / 23.286 / 29.321                    |

These reproduce a residual on the same scale as the historical production
report, but do not establish its fleet cause: production Runner versions,
manifests, host load and working sets differ. The production p90 residual of
32 ms must not be interpreted as 32 ms available to remove.

### Guest attribution and perturbation

| Repeated-call phase                 |   I1 p50/p90 |   I2 p50/p90 |   I3 p50/p90 |
| ----------------------------------- | -----------: | -----------: | -----------: |
| Cancellation + containment creation |  .432 / .932 |  .422 / .970 |  .418 / .969 |
| Command preparation + spawn         | .623 / 1.995 | .601 / 1.680 | .652 / 1.816 |
| I/O setup                           |  .100 / .311 |  .108 / .351 |  .125 / .376 |
| Wait/completion/group reap          | .787 / 1.448 | .817 / 1.478 | .805 / 1.433 |
| Containment cleanup                 |  .240 / .462 |  .217 / .440 |  .208 / .490 |
| Remaining joins/result handling     | .151 / 2.995 | .144 / 3.129 | .171 / 3.028 |

These are phase percentiles, not additive components of outer p90. The join
interval is measurable but below the proposed 5-ms screen at p90. The 50-ms
cancellation interval does not impose a 50-ms delay on fast child exits.

I1/I2/I3 outer p50/p90 are `3.741/6.526`, `3.679/6.205`, `3.952/6.074` ms.
Compared to adjacent A trials, absolute p50 changes are at most .379 ms and p90
changes at most .200 ms: below 1 ms and within observed baseline batch variation.
This is a descriptive perturbation check, not a statistical equivalence result.
The much smaller first/resumed cohorts cannot establish sub-ms perturbation.
**No clocks or new success diagnostics remain enabled in product code.**

The later J refinement, not interleaved for a causal overhead comparison,
confirms that preparation is only 6–7 microseconds on resumed samples:

| Trial | Resumed spawn intervals (ms) | Repeated spawn p50/p90 (ms) |
| ----- | ---------------------------- | --------------------------: |
| J1    | 24.815 / 15.782 / 9.188      |                .543 / 1.887 |
| J2    | 15.437 / 9.298 / 22.167      |                .551 / 1.970 |
| J3    | 15.932 / 14.777 / 16.352     |                .532 / 1.831 |

## Correctness and resource guardrails

High-work outer p50/p90/max: A1 `64.470/160.397/191.758`,
A2 `61.141/139.951/186.245`, A3 `62.968/124.718/225.993` ms.
Oversized fallback: A1 `23.759/40.971/51.342`, A2 `23.119/40.849/66.298`,
A3 `23.201/35.636/69.729` ms. Each row has only ten observations; p95/p99 would
equal the observed maximum and are not stable tail estimates. Do not combine
fallback and dedicated residuals. These initial baseline cohorts had no
optimization treatment; their guardrails alone do not demonstrate an improvement.

G1/G2 outer repeated p50/p90/p95/p99 were `4.056/6.103/6.600/7.329` and
`3.773/6.394/7.143/15.359` ms. Both completed all 324 calls and cleanup. G2's
p99 exceeds the isolated A maxima of their p99s; do not claim zero tail
regression from this small contention check. Concurrent startup also serialized
some proxy initialization (one proxy took 3.762 s versus 1.866 s); proxy startup
is outside the storage timer.

For A1–A3, the aggregate guest exec-cgroup CPU delta was 1.662/1.691/1.681 s
across each complete fixture, including setup/high-work/fallback. Available
guest memory after the fixture was 3,404–3,433 MiB versus 3,455 MiB before.
The G trials used 1.680/1.717 s aggregate exec-cgroup CPU, with 3,432/3,400 MiB
available afterward. These are not per-helper CPU or peak-RSS measurements.
Parent cgroup pressure counters stayed zero; they do not independently prove
every removed leaf's CPU history. Guest PSI and strace are unavailable.

During concurrent startup, host available memory was 29,182 → 29,181 MiB.
Host PSI total deltas were CPU some 69,425 us, memory some 1,734 us, and I/O
some 16,149 us; avg10 remained 0.00. `/usr/bin/time -v` reported 10.65/10.71 s
process-tree CPU and 10.75/8.91 s wall time for G1/G2. Its maximum RSS is **not**
aggregate VM memory, so it is not used as a VM-memory guardrail. The snapshot
and image cache remain warm and host pressure is not attributable solely to us.

Before and after each fixture, the only exec cgroup present was the resource
probe's own current operation. Every park passed the reusable/quiescent gate;
normal teardown released the VM, COW slots, namespace pool and proxy. Existing
development services remained active. Inactive experiment binaries/configs and
bounded logs were retained for audit; no host-wide GC or production action ran.

## Initial experiment archive (not proposed for merging)

The initial A/I/G/J1–J3 report and opt-in patches remain available in the unmerged [experiment archive at `a99899c4165c03382e5ce365300684f11ac6a0b2`](https://github.com/vm0-ai/vm0/tree/a99899c4165c03382e5ce365300684f11ac6a0b2/docs/benchmarks/storage-helper-attribution).
They do not need to be merged to preserve this issue's findings.
These patches are **opt-in, disposable experiment code**, not production APIs.
Use a separate clean checkout of the pinned revision on an approved development
host. Never apply them to an active service or run them with customer input.
`probe.patch` adds a private magic benchmark command solely to exercise the
real Sandbox entry points. Its final log reader uses `.get()` instead of the
indexed lookups in the measured binary; this lint cleanup is after all timers.
It also omits failure diagnostic text from measurement JSON; fatal errors still
fail the trial. Only fixed numeric success phases are emitted by the guest patch.
`guest-phases.patch` and its test reproduce the J refinement. The recorded I
instrumentation predated `command_prepared` and the `prepare_us`/`spawn_us`
fields; it had six fields, retaining their enclosing timer.
The patch also adjusts the opt-in connection test to require numeric phase
diagnostics on success; the unpatched production contract still requires an
empty success diagnostic. Do not deploy the experimental diagnostic contract.

```bash
git apply --check /path/to/this/report/probe.patch
git apply /path/to/this/report/probe.patch
cd crates
CARGO_BUILD_JOBS=4 cargo build --profile ci --target x86_64-unknown-linux-musl \
  -p runner -p guest-agent -p guest-storage-apply -p guest-init \
  -p guest-state-restore -p guest-write-file -p guest-tool-exec \
  -p runner-rpc-client -p claude-mock -p codex-mock
```

Preserve baseline guest-init before applying the guest timing patch and
rebuilding only guest-init. Use `runner build --profile vm0/default` with
explicit paths for **all nine** guest binaries (the corresponding `--guest-*`,
`--runner-rpc-client`, `--claude-mock`, `--codex-mock` flags). Then use
`runner config` with each returned rootfs/snapshot hash and a unique benchmark
group/base directory, `--max-concurrent 1`, the approved hostname, a deliberately
non-serving API URL and synthetic token. Do not invoke the normal service
deployment script: it stops/replaces its configured service and runs GC.

```bash
set -o pipefail
sudo /path/to/experimental/runner benchmark issue-32428-storage-probe \
  --config /path/to/isolated/runner.yaml --profile vm0/default \
  2>&1 | tee /path/to/private/trial.log
jq -Rc 'select(startswith("{")) | fromjson' /path/to/private/trial.log \
  | jq -s -f /path/to/this/report/summary.jq
bash /path/to/this/report/test-summary.sh
```

Require benchmark exit zero and complete teardown as well as the summarizer's
coverage check. The summarizer validates exact outer sample order and requires
324 successful, untruncated outer observations and 324 valid child observations.
Ordinal pairing is valid **only for this serial, single-owner synthetic log**;
never apply it to interleaved production telemetry. The script exposes the
number of diagnostic samples and keeps shape/phase/batch groups separate.
`by_shape_phase` additionally summarizes each lifecycle category within that
single trial (including the 300 repeated calls), never across trials or builds.
Its small-cohort percentile outputs are mathematical order statistics, not
evidence of reliable p99 estimates.

### Binary and image identities

| Binary                                       | SHA-256                                                            |
| -------------------------------------------- | ------------------------------------------------------------------ |
| Experimental Runner, common to formal trials | `24860411eca09149431fad26b4dde60afc94f5c946332593be897e051ff1aff5` |
| Baseline guest-init                          | `184177ddf2924f13f66562e7197513e48deea44541ef6870c66f26b8ff1dde4a` |
| I guest-init                                 | `848967c455697dc309057ca7dbfcb864710823f3a8769d13dfeea0e48d167ca1` |
| J guest-init                                 | `6b80eebffb8e7fe07dffa2f8cdeaeff64d9247a331134f5a56e56efe4deb3eae` |
| Unchanged guest-storage-apply                | `adb21481c0d39cb5e474715db16883b9fddcc6775d3c0f20697618d9db6ce837` |

| Image | Rootfs hash                                                        | Snapshot hash                                                      |
| ----- | ------------------------------------------------------------------ | ------------------------------------------------------------------ |
| A/G   | `203c5532218d920b96643368e9176fd106a6b249d8f342df322fa102e5c3200f` | `17209ec83aaf0265273510d120ade287e07c1fe02b375d0810fbb473080e47c8` |
| I     | `44cefd79e584311bf58b0b0b046a4d5f1e597fe4a13fb36821c68b215ca38517` | `077b4983652150e781689bbd2183463588ca5fdb23e0afd3536785dd07631b39` |
| J     | `bbb9969eee46deef458129fc7da5aaddcb62e0f1773092c2791514a5015bbb38` | `9451edaf630308ee570c45aa76ebd50a0dd94ee97e237688537aab1a1984c636` |

Raw synthetic JSON Lines are retained locally under
`codex-work/research/issue-32428-storage-helper-attribution/` (`a1`/`i1` through
`a3`/`i3`, `guard1`/`guard2`, and `j1` through `j3`, each `.jsonl`). Full Runner
logs remain on `local-11.gcp.vm3.ai` under the experiment-owned
`/var/lib/vm0-runner/bin/issue-32428-attribution-vm2/` directory.
They are intentionally not ingested into production telemetry. The tables
above are the durable issue result; all statistics can be recalculated from a
fresh trial with the archived summarizer.

## Validation

The real-guest runs above are the performance/containment evidence. Local
connection tests use TestNoop and are correctness checks, not cgroup benchmarks.
Selected checks were run in the foreground:

```bash
bash docs/benchmarks/storage-helper-attribution/test-summary.sh
bash -n docs/benchmarks/storage-helper-attribution/test-summary.sh
shellcheck docs/benchmarks/storage-helper-attribution/test-summary.sh
pnpm -C turbo exec prettier --check ../docs/benchmarks/storage-helper-attribution/README.md
# From crates, with CARGO_BUILD_JOBS=4:
cargo fmt --all -- --check
cargo clippy --profile local -p runner --bins -- -D warnings
cargo clippy --profile local -p guest-control-server --all-targets --all-features -- -D warnings
RUSTDOCFLAGS="-D warnings" cargo doc --profile local -p guest-control-server --no-deps
cargo test --profile local -p runner storage --all-targets --all-features
cargo test --profile local -p guest-control-server --test connection guest_storage_manifest --all-features
cargo test --profile local -p guest-storage-apply --all-targets --all-features
```

Runner's storage selection passed 204 tests (its existing isolated-process
helper is invoked by its parent test). The eight storage connection tests passed
both without the diagnostic patch and with the patch's explicit experimental
success-diagnostic assertion. Storage-apply's full selected suite passed.
The summarizer was also run successfully on all eleven complete formal logs;
its CLI regression test checks paired quantiles and rejects incomplete, failed,
truncated, reordered and malformed observations. Both patches apply cleanly
to the pinned source. All temporary active Rust edits were removed before the documentation-only
PR was submitted. No product instrumentation or behavior change is being shipped.

## Original acceptance criteria and preserved constraints

- Obtain a controlled same-host baseline using representative low-work Exact manifests and high-work/oversized regression cases. Keep workload, binaries, guest capacity and containment policy fixed; report first/resumed versus repeated invocations separately.
- Map both timer boundaries precisely. Attribute outer handling, containment setup/placement, helper launch and initialization/input, guest application, completion/drain/reap/containment cleanup, and response handling. Report non-additive or overlapping phases correctly.
- Use existing timing/profiling facilities first. If retained instrumentation is necessary, make it bounded, low-overhead, low-cardinality and free of manifests, paths, run/session identities, credentials and arbitrary error text. Measure its perturbation.
- Report complete-pair coverage, outer/inner/residual distributions, failures and representative resource/tail guardrails.
- Decide whether a specific component can be removed or overlapped safely with a meaningful controlled gain, or record no worthwhile change. Do not implement a new protocol, service or resource policy as part of this attribution slice.
- Any retained instrumentation passes relevant integration and Rust checks, including the existing cancellation, timeout, connection loss, child/descendant cleanup and successful storage-application boundaries.

## Invariants and exclusions

Preserve #28732's production containment contract: cgroup/resource-policy placement, required security restrictions, descendant cleanup, timeout/cancellation ownership and graceful/forced cleanup. Do not copy the lighter history-verifier process-group role onto the archive helper.

Keep storage content, retries, extraction security, mount conflict ordering, instruction normalization, cleanup and artifact-root repair unchanged. Zero archive tasks or matching fingerprints do not prove no guest work; the storage planner already skips truly unnecessary work.

#32415 covers host-cache-to-guest staging writes before this operation. #32414 covers archive scheduling inside it. This issue owns neither of those boundaries. #32418's blank history overlap is also separate. Do not add their p90 values to this residual.

Relevant files: `crates/runner/src/executor/storage.rs`, `crates/runner/src/storage_plan.rs`, `crates/guest-control-server/src/{guest_storage_manifest,process_containment,wait}.rs`, and `crates/guest-storage-apply/src/lib.rs`. Account for old guest-download and current guest-storage-apply names when joining production telemetry.

The development-host execution prerequisite was resolved by explicit user approval; the authorized measurements are recorded above. No active execution blocker remains for this attribution report. Further host experiments, product changes or production deployment are not authorized by this result.

## Next steps and remaining uncertainty

Implementation is tracked in **#32647**, a direct sibling under **#24203**. This
research issue remains open and retains its parent. No additional attribution
PR is required; #32544 remains a closed, unmerged archive of the initial stage.
The new issue defines the shared guest-control-server contained-process scope,
its invariants and acceptance gates. Issue creation is not a runtime code
change, production deployment or merge approval.

Remaining questions and implementation gates:

1. Measure how often this placement/RCU path dominates current production
   low-work Exact calls, matched by Runner/guest version and lifecycle, before
   making fleet or complete API-to-spawn savings claims.
2. The source/stack/candidate evidence supports the sampled migration wait, but
   exact grace-period duration and indirect scheduling/working-set effects
   remain unresolved; zero added faults does not rule out indirect effects.
3. Address the prototype's 0.344–0.716 ms warm-median increase, audit the shared
   raw-clone launcher, and add real-cgroup identity/lifecycle coverage for
   generic exec, oversized fallback and controlled Agent startup. The existing
   storage experiment does not establish those paths' safety or performance.
4. Preserve separate cold/resumed, warm-tail, high-work, fallback and
   concurrent-start/resource measurements with explicit regression bounds.
   Representative concurrent Agent/tool migration remains unmeasured.
5. Guest Agent's internal Tokio CLI launcher and the tool launcher's
   migrate-self-then-exec path are deferred authority/lifecycle boundaries,
   not implicitly included in the shared supervisor implementation.

The earlier Research, Options, Plan, Challenge, execution approval/blocker and
implementation comments remain the historical audit trail. The former
host-approval blocker is resolved. Earlier references to an attribution PR are
superseded by the issue-only research delivery decision; unresolved spawn
internals were narrowed by K/L, and the later actual-RPC experiments now replace
the earlier "no tested optimization" conclusion. The provisional local
favordynmods-first recommendation predates the clone3 RPC and lifecycle results.

Parent #24203 and siblings #32414, #32415 and #32418 retain their existing scopes.
#32647 owns the new shared-launcher implementation, not archive scheduling,
cache staging or history overlap.

---

> Compacted on 2026-09-08 from 8 comments; updated with K/L attribution, candidate and actual-RPC/lifecycle/concurrency experiments, and the user-requested implementation handoff to #32647.

seven332:
## Summary

Closes #32428. Parent: #24203 (not closed by this PR).

- Record an approved, isolated development-host experiment with 3,564 successful, completely paired storage helper invocations across baseline, diagnostic and concurrent-start cohorts.
- Locate the large first-after-resume cost in `Command::spawn()` (9.188–24.815 ms in the nine refined resumed observations), versus 6–7 microseconds for command preparation. Keep cold/resumed and repeated distributions separate.
- Include pinned source/binary/image identities, precise timer boundaries, resource/tail limitations, disposable opt-in probe patches, and a tested complete-pair summarizer.

This is an attribution/decision PR. No Rust product code, protocol, resource policy, service, CI selection, or production logging changes. No optimization or fleet performance gain is claimed. Existing development services were not replaced; all experiment VMs/proxies/pools completed normal teardown.

## Validation

- Eleven real-guest trials: 3,564/3,564 complete successful pairs; content, output bounds, fallback cleanup and park/teardown gates passed.
- Summarizer CLI regression tests, Bash syntax check, ShellCheck, and summary validation against all eleven real logs passed.
- Markdown Prettier and both patches' applicability checks passed.
- Experimental Rust: formatting, Runner binary Clippy, guest-control-server all-targets/all-features Clippy, and guest-control-server rustdoc passed.
- `cargo test --profile local -p runner storage --all-targets --all-features`: 204 passed; existing subprocess helper exercised by its parent.
- `cargo test --profile local -p guest-control-server --test connection guest_storage_manifest --all-features`: 8 passed both with the experimental diagnostic assertion and without the patch.
- `cargo test --profile local -p guest-storage-apply --all-targets --all-features`: passed.

## Limits / risk

Synthetic workloads on a shared development host and the CI build profile are not the historical production cohort. Small cold samples do not establish stable p99 or identify which syscall/page-fault/scheduling component inside spawn is responsible. Concurrent-start p99 varied, so no zero-tail-regression claim is made. Experimental patches are explicitly disposable and must never be applied to active/customer-serving services.

### DIFF
--- a/docs/benchmarks/storage-helper-attribution/README.md
+++ b/docs/benchmarks/storage-helper-attribution/README.md
@@ -1,0 +1,319 @@
+# Storage helper invocation attribution
+
+Decision record for [#32428](https://github.com/vm0-ai/vm0/issues/32428), a child
+of [#24203](https://github.com/vm0-ai/vm0/issues/24203). Measured September 8, 2026.
+This change adds an experiment record and opt-in reproduction artifacts only.
+It does not change Runner, guest code, containment, resource policy, or CI execution.
+
+## Decision
+
+The large residual is reproducible on the **first helper invocation after
+resume**, not as a constant cost on every invocation. In the refined diagnostic
+cohort, `Command::spawn()` took **9.188–24.815 ms** on nine first-after-resume
+observations; command preparation took **6–7 microseconds**. Repeated calls in
+that same cohort had spawn p90 of **1.831–1.970 ms** in each 300-call trial.
+The enclosing cgroup creation and cleanup intervals were much smaller.
+
+Do not weaken workload containment, remove mandatory storage work, introduce a
+daemon, or change cancellation polling based on the production residual. No
+safe optimization with a demonstrated gain is selected by this attribution PR.
+The next useful target is the **cold/resumed spawn interval**, including its
+child-side security hooks and exec handshake. The measurement does not separate
+fork/clone, page faults, cgroup placement, identity syscalls, exec, and scheduling
+inside that interval. Balloon-driven cache reclamation is a hypothesis, not a
+proven cause. Changing balloon policy or replacing the launcher requires a
+separate experiment and design review, including memory and isolation costs.
+
+The provisional 5-ms p90 screen is not met by any single measured warm phase
+across all trials. Cold spawn exceeds that scale, but these small cold cohorts
+do not establish a stable p90 or a safely recoverable saving. This closes the
+attribution/decision slice, not the parent performance effort or a runtime fix.
+
+## Boundaries
+
+| Observation                                              | Includes                                                                                                                                            | Excludes / interpretation                                                                                                                                                       |
+| -------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Production `runner_storage_manifest_guest_storage_apply` | `download_storages`: serialization, runtime-path construction, dedicated operation or fallback, result checks                                       | Prior cache population/staging and other spawn phases                                                                                                                           |
+| Probe `outer_us`                                         | Manifest serialization, the same public Sandbox operation and await; fallback includes remove/write/exec                                            | Synthetic reconstruction, **not the production metric**: no production context/path construction, logging, or failure-cleanup wrapper; result validation occurs after the timer |
+| Dedicated result `guest_ms`                              | Guest `run_manifest`, from before cancellation/cgroup creation through child handling, containment cleanup, input/output joins, and result assembly | Request parsing, worker admission/dispatch, response encoding, writer-lock wait and socket write                                                                                |
+| Child `download_total`                                   | Input read, stale manifest cleanup, parsing, storage work, normalization                                                                            | Runtime log-path setup and argument parsing before the timer; final telemetry/logging and process exit after it                                                                 |
+| `outer − guest`                                          | Serialization, host/guest admission and scheduling, transport and response handling                                                                 | **Not pure network latency**; guest timer is rounded down to integer milliseconds                                                                                               |
+| `outer − inner`                                          | Paired residual around application work                                                                                                             | Not guaranteed savings and never a subtraction of percentile values                                                                                                             |
+
+Temporary diagnostic intervals partition guest wall time: cancellation/cgroup
+creation; command preparation plus spawn; I/O worker setup; wait through
+group cleanup/reap; containment cleanup; remaining stdin/drain joins and result
+handling. The refined version splits preparation from spawn without changing
+the boundary of their combined interval. `prepare_spawn_us` is their **enclosing
+total**, not an additional additive phase.
+
+Input writing and output drains overlap child execution. The child application
+timer overlaps I/O setup/wait; do not add it to those intervals. The first-call
+wait interval also includes initialization, terminal logging, and scheduling
+outside the child timer. No absolute timestamp is subtracted across host/guest
+clocks. Quantized residuals retain their signed values; missing events are not
+zero-filled. Fallback `guest_ms` describes its final generic exec, not all three
+fallback operations, and is not the dedicated middle boundary.
+
+## Environment and workload
+
+- Source: `16d1e9a9b2bcd80e20ae37dfd64584814c46b531`, Runner `0.188.7`.
+  All guest binaries use this source; only guest-init receives the experimental
+  guest-control timing patch in diagnostic images.
+- Build: Rust `ci` profile, `x86_64-unknown-linux-musl`, optimized thin LTO,
+  four codegen units, no `test-support` or debug assertions. This is the normal
+  development Runner build profile, **not** the production full-LTO release.
+- Approved shared development host: `local-11.gcp.vm3.ai`, kernel
+  `6.17.0-1018-gcp`, 16 logical CPUs (Xeon Platinum 8581C), 32,086 MiB RAM.
+  Roughly 29 GiB was available. This is not a production host or a dedicated
+  machine; unrelated development services were not stopped or replaced.
+- Profile: `vm0/default`, 2 vCPUs, 4,096 MiB guest memory. Normal production
+  Workload cgroups, identity drop, security hooks, process-group ownership,
+  output bounds, and lifecycle cleanup remain enabled.
+- Existing `runner benchmark` owns the proxy, VM, COW devices and namespace pool.
+  It prefetches snapshot memory. Every trial starts a new snapshot-restored VM;
+  every 100-call batch is followed by normal `park()` / `unpark()` with the
+  default balloon controller. This is a lifecycle analogue, not a complete
+  API-admitted Exact run with an active agent and its working set.
+- Logs use the production home-based runtime directory, not tmpfs. No real
+  manifests, customer data, provider calls, or real API token are used. R2
+  credentials were supplied only to the ordinary image builder to read its
+  cached template, never to measurement output.
+
+Each trial has 324 storage invocations: one first call; three batches of 100
+repeated cached calls with one resumed call after each batch; ten high-work
+calls; ten oversized calls. The cached fixture has 17 cached mounts, one with
+instruction normalization still requested. The archive fixture has 140 local
+`file://` mounts, each extracting a 4-KiB payload. The oversized cached fixture
+exceeds the 65,536-byte dedicated stdin limit via a 70,000-character synthetic
+name and exercises remove/write/generic-exec fallback. Its child removes the
+canonical temporary manifest. Instruction content, all extracted payload sizes,
+fallback manifest removal, successful exit and untruncated output are checked.
+
+Archives warm during each trial; high-work results are not a fixed cold-cache
+distribution. Fixtures and their reset/order are identical between images.
+The actual planner's true-no-work case is checked separately by the existing
+Runner entry-point test and has **no helper invocation**, not a zero-ms helper
+sample. Existing storage integration coverage validates normalization, nested
+mount ordering and extraction security beyond this synthetic fixture.
+
+## Evidence and distributions
+
+Formal runs: `A1, I1, A2, I2, A3, I3` alternated at 05:35:54–05:36:47 UTC;
+`G1/G2` were started together at 05:41:04–05:41:15;
+`J1–J3` refined the spawn split at 06:02:10–06:02:36.
+A/G use uninstrumented guests; I uses six wall intervals; J adds the two
+preparation/spawn subintervals. Do not pool these builds or lifecycle categories.
+
+All **3,564 / 3,564** formal invocations have successful complete child pairs,
+with no truncated captures. Each A/I/J set has 900 repeated cached, 3 first,
+9 resumed, 30 high-work and 30 oversized observations; G has 600/2/6/20/20.
+An earlier 324-call calibration used tmpfs logs and is excluded. A subsequent
+resource-tool availability check failed before the first storage invocation:
+guest `command -v strace` returned 127, not 1. It was fixed before formal runs.
+That failed setup is not counted as a successful trial or silently rerun away.
+
+Nearest-rank distributions below are **p50 / p90 / p95 / p99 in milliseconds**.
+Every residual is calculated per pair before percentile selection. Each row
+contains 300 repeated cached calls, preserving individual trial identity.
+
+| Trial | Probe outer                    | Guest result   | Child inner   | Outer − inner                  | Outer − guest                |
+| ----- | ------------------------------ | -------------- | ------------- | ------------------------------ | ---------------------------- |
+| A1    | 4.120 / 6.326 / 7.277 / 8.026  | 3 / 5 / 6 / 7  | 0 / 1 / 1 / 2 | 3.974 / 6.318 / 7.130 / 8.026  | .974 / 1.355 / 1.480 / 2.900 |
+| A2    | 3.899 / 6.306 / 7.332 / 13.038 | 3 / 5 / 6 / 10 | 0 / 0 / 1 / 1 | 3.764 / 6.282 / 7.332 / 10.038 | .909 / 1.296 / 1.383 / 1.744 |
+| A3    | 3.796 / 5.978 / 6.730 / 8.488  | 3 / 5 / 6 / 7  | 0 / 0 / 1 / 1 | 3.746 / 5.834 / 6.526 / 8.488  | .869 / 1.317 / 1.390 / 1.546 |
+
+The three 100-call batch outer p90s are A1 `4.500 / 7.277 / 6.318`,
+A2 `4.016 / 7.431 / 6.589`, A3 `3.882 / 6.191 / 6.444` ms. Warm calls after
+resume are not identical to the pre-park batch; they were not discarded.
+
+Small lifecycle cohorts are reported as individual values, **not stable tails**:
+
+| Trial | First outer | Three first-after-resume outer observations |
+| ----- | ----------: | ------------------------------------------- |
+| A1    |      30.760 | 34.544 / 18.811 / 35.014                    |
+| A2    |      24.761 | 25.567 / 24.523 / 21.278                    |
+| A3    |      35.066 | 21.677 / 23.286 / 29.321                    |
+
+These reproduce a residual on the same scale as the historical production
+report, but do not establish its fleet cause: production Runner versions,
+manifests, host load and working sets differ. The production p90 residual of
+32 ms must not be interpreted as 32 ms available to remove.
+
+### Guest attribution and perturbation
+
+| Repeated-call phase                 |   I1 p50/p90 |   I2 p50/p90 |   I3 p50/p90 |
+| ----------------------------------- | -----------: | -----------: | -----------: |
+| Cancellation + containment creation |  .432 / .932 |  .422 / .970 |  .418 / .969 |
+| Command preparation + spawn         | .623 / 1.995 | .601 / 1.680 | .652 / 1.816 |
+| I/O setup                           |  .100 / .311 |  .108 / .351 |  .125 / .376 |
+| Wait/completion/group reap          | .787 / 1.448 | .817 / 1.478 | .805 / 1.433 |
+| Containment cleanup                 |  .240 / .462 |  .217 / .440 |  .208 / .490 |
+| Remaining joins/result handling     | .151 / 2.995 | .144 / 3.129 | .171 / 3.028 |
+
+These are phase percentiles, not additive components of outer p90. The join
+interval is measurable but below the proposed 5-ms screen at p90. The 50-ms
+cancellation interval does not impose a 50-ms delay on fast child exits.
+
+I1/I2/I3 outer p50/p90 are `3.741/6.526`, `3.679/6.205`, `3.952/6.074` ms.
+Compared to adjacent A trials, absolute p50 changes are at most .379 ms and p90
+changes at most .200 ms: below 1 ms and within observed baseline batch variation.
+This is a descriptive perturbation check, not a statistical equivalence result.
+The much smaller first/resumed cohorts cannot establish sub-ms perturbation.
+**No clocks or new success diagnostics remain enabled in product code.**
+
+The later J refinement, not interleaved for a causal overhead comparison,
+confirms that preparation is only 6–7 microseconds on resumed samples:
+
+| Trial | Resumed spawn intervals (ms) | Repeated spawn p50/p90 (ms) |
+| ----- | ---------------------------- | --------------------------: |
+| J1    | 24.815 / 15.782 / 9.188      |                .543 / 1.887 |
+| J2    | 15.437 / 9.298 / 22.167      |                .551 / 1.970 |
+| J3    | 15.932 / 14.777 / 16.352     |                .532 / 1.831 |
+
+## Correctness and resource guardrails
+
+High-work outer p50/p90/max: A1 `64.470/160.397/191.758`,
+A2 `61.141/139.951/186.245`, A3 `62.968/124.718/225.993` ms.
+Oversized fallback: A1 `23.759/40.971/51.342`, A2 `23.119/40.849/66.298`,
+A3 `23.201/35.636/69.729` ms. Each row has only ten observations; p95/p99 would
+equal the observed maximum and are not stable tail estimates. Do not combine
+fallback and dedicated residuals. There is no optimization treatment for which
+these guardrails could demonstrate a performance improvement.
+
+G1/G2 outer repeated p50/p90/p95/p99 were `4.056/6.103/6.600/7.329` and
+`3.773/6.394/7.143/15.359` ms. Both completed all 324 calls and cleanup. G2's
+p99 exceeds the isolated A maxima of their p99s; do not claim zero tail
+regression from this small contention check. Concurrent startup also serialized
+some proxy initialization (one proxy took 3.762 s versus 1.866 s); proxy startup
+is outside the storage timer.
+
+For A1–A3, the aggregate guest exec-cgroup CPU delta was 1.662/1.691/1.681 s
+across each complete fixture, including setup/high-work/fallback. Available
+guest memory after the fixture was 3,404–3,433 MiB versus 3,455 MiB before.
+The G trials used 1.680/1.717 s aggregate exec-cgroup CPU, with 3,432/3,400 MiB
+available afterward. These are not per-helper CPU or peak-RSS measurements.
+Parent cgroup pressure counters stayed zero; they do not independently prove
+every removed leaf's CPU history. Guest PSI and strace are unavailable.
+
+During concurrent startup, host available memory was 29,182 → 29,181 MiB.
+Host PSI total deltas were CPU some 69,425 us, memory some 1,734 us, and I/O
+some 16,149 us; avg10 remained 0.00. `/usr/bin/time -v` reported 10.65/10.71 s
+process-tree CPU and 10.75/8.91 s wall time for G1/G2. Its maximum RSS is **not**
+aggregate VM memory, so it is not used as a VM-memory guardrail. The snapshot
+and image cache remain warm and host pressure is not attributable solely to us.
+
+Before and after each fixture, the only exec cgroup present was the resource
+probe's own current operation. Every park passed the reusable/quiescent gate;
+normal teardown released the VM, COW slots, namespace pool and proxy. Existing
+development services remained active. Inactive experiment binaries/configs and
+bounded logs were retained for audit; no host-wide GC or production action ran.
+
+## Reproduction artifacts
+
+These patches are **opt-in, disposable experiment code**, not production APIs.
+Use a separate clean checkout of the pinned revision on an approved development
+host. Never apply them to an active service or run them with customer input.
+`probe.patch` adds a private magic benchmark command solely to exercise the
+real Sandbox entry points. Its final log reader uses `.get()` instead of the
+indexed lookups in the measured binary; this lint cleanup is after all timers.
+It also omits failure diagnostic text from measurement JSON; fatal errors still
+fail the trial. Only fixed numeric success phases are emitted by the guest patch.
+`guest-phases.patch` and its test reproduce the J refinement. The recorded I
+instrumentation predated `command_prepared` and the `prepare_us`/`spawn_us`
+fields; it had six fields, retaining their enclosing timer.
+The patch also adjusts the opt-in connection test to require numeric phase
+diagnostics on success; the unpatched production contract still requires an
+empty success diagnostic. Do not deploy the experimental diagnostic contract.
+
+```bash
+git apply --check /path/to/this/report/probe.patch
+git apply /path/to/this/report/probe.patch
+cd crates
+CARGO_BUILD_JOBS=4 cargo build --profile ci --target x86_64-unknown-linux-musl \
+  -p runner -p guest-agent -p guest-storage-apply -p guest-init \
+  -p guest-state-restore -p guest-write-file -p guest-tool-exec \
+  -p runner-rpc-client -p claude-mock -p codex-mock
+```
+
+Preserve baseline guest-init before applying the guest timing patch and
+rebuilding only guest-init. Use `runner build --profile vm0/default` with
+explicit paths for **all nine** guest binaries (the corresponding `--guest-*`,
+`--runner-rpc-client`, `--claude-mock`, `--codex-mock` flags). Then use
+`runner config` with each returned rootfs/snapshot hash and a unique benchmark
+group/base directory, `--max-concurrent 1`, the approved hostname, a deliberately
+non-serving API URL and synthetic token. Do not invoke the normal service
+deployment script: it stops/replaces its configured service and runs GC.
+
+```bash
+set -o pipefail
+sudo /path/to/experimental/runner benchmark issue-32428-storage-probe \
+  --config /path/to/isolated/runner.yaml --profile vm0/default \
+  2>&1 | tee /path/to/private/trial.log
+jq -Rc 'select(startswith("{")) | fromjson' /path/to/private/trial.log \
+  | jq -s -f /path/to/this/report/summary.jq
+bash /path/to/this/report/test-summary.sh
+```
+
+Require benchmark exit zero and complete teardown as well as the summarizer's
+coverage check. The summarizer validates exact outer sample order and requires
+324 successful, untruncated outer observations and 324 valid child observations.
+Ordinal pairing is valid **only for this serial, single-owner synthetic log**;
+never apply it to interleaved production telemetry. The script exposes the
+number of diagnostic samples and keeps shape/phase/batch groups separate.
+`by_shape_phase` additionally summarizes each lifecycle category within that
+single trial (including the 300 repeated calls), never across trials or builds.
+Its small-cohort percentile outputs are mathematical order statistics, not
+evidence of reliable p99 estimates.
+
+### Binary and image identities
+
+| Binary                                       | SHA-256                                                            |
+| -------------------------------------------- | ------------------------------------------------------------------ |
+| Experimental Runner, common to formal trials | `24860411eca09149431fad26b4dde60afc94f5c946332593be897e051ff1aff5` |
+| Baseline guest-init                          | `184177ddf2924f13f66562e7197513e48deea44541ef6870c66f26b8ff1dde4a` |
+| I guest-init                                 | `848967c455697dc309057ca7dbfcb864710823f3a8769d13dfeea0e48d167ca1` |
+| J guest-init                                 | `6b80eebffb8e7fe07dffa2f8cdeaeff64d9247a331134f5a56e56efe4deb3eae` |
+| Unchanged guest-storage-apply                | `adb21481c0d39cb5e474715db16883b9fddcc6775d3c0f20697618d9db6ce837` |
+
+| Image | Rootfs hash                                                        | Snapshot hash                                                      |
+| ----- | ------------------------------------------------------------------ | ------------------------------------------------------------------ |
+| A/G   | `203c5532218d920b96643368e9176fd106a6b249d8f342df322fa102e5c3200f` | `17209ec83aaf0265273510d120ade287e07c1fe02b375d0810fbb473080e47c8` |
+| I     | `44cefd79e584311bf58b0b0b046a4d5f1e597fe4a13fb36821c68b215ca38517` | `077b4983652150e781689bbd2183463588ca5fdb23e0afd3536785dd07631b39` |
+| J     | `bbb9969eee46deef458129fc7da5aaddcb62e0f1773092c2791514a5015bbb38` | `9451edaf630308ee570c45aa76ebd50a0dd94ee97e237688537aab1a1984c636` |
+
+Raw synthetic JSON Lines are retained in the workflow's ignored research
+directory, with full Runner logs in the isolated host experiment directory.
+They are intentionally not ingested into production telemetry. The tables
+above are the checked-in result; all statistics can be recalculated from a
+fresh trial with the included summarizer.
+
+## Validation
+
+The real-guest runs above are the performance/containment evidence. Local
+connection tests use TestNoop and are correctness checks, not cgroup benchmarks.
+Selected checks were run in the foreground:
+
+```bash
+bash docs/benchmarks/storage-helper-attribution/test-summary.sh
+bash -n docs/benchmarks/storage-helper-attribution/test-summary.sh
+shellcheck docs/benchmarks/storage-helper-attribution/test-summary.sh
+pnpm -C turbo exec prettier --check ../docs/benchmarks/storage-helper-attribution/README.md
+# From crates, with CARGO_BUILD_JOBS=4:
+cargo fmt --all -- --check
+cargo clippy --profile local -p runner --bins -- -D warnings
+cargo clippy --profile local -p guest-control-server --all-targets --all-features -- -D warnings
+RUSTDOCFLAGS="-D warnings" cargo doc --profile local -p guest-control-server --no-deps
+cargo test --profile local -p runner storage --all-targets --all-features
+cargo test --profile local -p guest-control-server --test connection guest_storage_manifest --all-features
+cargo test --profile local -p guest-storage-apply --all-targets --all-features
+```
+
+Runner's storage selection passed 204 tests (its existing isolated-process
+helper is invoked by its parent test). The eight storage connection tests passed
+both without the diagnostic patch and with the patch's explicit experimental
+success-diagnostic assertion. Storage-apply's full selected suite passed.
+The summarizer was also run successfully on all eleven complete formal logs;
+its CLI regression test checks paired quantiles and rejects incomplete, failed,
+truncated, reordered and malformed observations. Both patches apply cleanly
+to the pinned source. No Rust source changes are retained by this PR.
--- a/docs/benchmarks/storage-helper-attribution/guest-phases.patch
+++ b/docs/benchmarks/storage-helper-attribution/guest-phases.patch
@@ -1,0 +1,98 @@
+diff --git a/crates/guest-control-server/src/guest_storage_manifest.rs b/crates/guest-control-server/src/guest_storage_manifest.rs
+index 5cb0197d67..ff79c026ef 100644
+--- a/crates/guest-control-server/src/guest_storage_manifest.rs
++++ b/crates/guest-control-server/src/guest_storage_manifest.rs
+@@ -255,6 +255,7 @@ fn run_manifest(input: RunManifestInput<'_>) -> GuestStorageManifestOutput {
+             );
+         }
+     };
++    let containment_created = Instant::now();
+     let mut prepared_containment = match process_containment.prepare_command() {
+         Ok(prepared) => prepared,
+         Err(error) => {
+@@ -287,6 +288,7 @@ fn run_manifest(input: RunManifestInput<'_>) -> GuestStorageManifestOutput {
+         );
+     }
+     prepared_containment.configure_process_inspection(&mut command);
++    let command_prepared = Instant::now();
+     let mut child = match spawn_in_own_process_group(&mut command) {
+         Ok(child) => child,
+         Err(error) => {
+@@ -299,6 +301,7 @@ fn run_manifest(input: RunManifestInput<'_>) -> GuestStorageManifestOutput {
+         }
+     };
+
++    let spawned = Instant::now();
+     let Some(stdin) = child.stdin.take() else {
+         return abort_spawned(
+             child,
+@@ -395,6 +398,7 @@ fn run_manifest(input: RunManifestInput<'_>) -> GuestStorageManifestOutput {
+     };
+     drop(drain_done_tx);
+
++    let io_ready = Instant::now();
+     let outcome = wait_with_kill_timeout_or_connection_cancelled(
+         child,
+         timeout_ms,
+@@ -403,9 +407,11 @@ fn run_manifest(input: RunManifestInput<'_>) -> GuestStorageManifestOutput {
+         // identity, so helper descendants cannot outlive this fixed operation.
+         || true,
+     );
++    let waited = Instant::now();
+     let cancellation_observed = connection_cancel.load(Ordering::Acquire);
+     let cleanup_mode = cleanup_mode_for_wait_outcome(&outcome, cancellation_observed);
+     let containment_result = process_containment.cleanup(cleanup_mode);
++    let contained = Instant::now();
+     let stdin_result = stdin_writer.join();
+     if !matches!(outcome, WaitOutcome::Exited(_))
+         || cancellation_observed
+@@ -457,6 +463,19 @@ fn run_manifest(input: RunManifestInput<'_>) -> GuestStorageManifestOutput {
+         diagnostic = append_cleanup_failure(diagnostic, Some(&error));
+     }
+
++    if termination == (ExecTermination::Exited { exit_code: 0 }) && diagnostic.is_empty() {
++        diagnostic = serde_json::json!({
++            "containment_us": containment_created.duration_since(started).as_micros(),
++            "prepare_spawn_us": spawned.duration_since(containment_created).as_micros(),
++            "prepare_us": command_prepared.duration_since(containment_created).as_micros(),
++            "spawn_us": spawned.duration_since(command_prepared).as_micros(),
++            "io_setup_us": io_ready.duration_since(spawned).as_micros(),
++            "wait_us": waited.duration_since(io_ready).as_micros(),
++            "cleanup_us": contained.duration_since(waited).as_micros(),
++            "join_us": contained.elapsed().as_micros(),
++        })
++        .to_string();
++    }
+     GuestStorageManifestOutput {
+         termination,
+         duration_ms: elapsed_ms(started),
+diff --git a/crates/guest-control-server/tests/connection/guest_storage_manifest.rs b/crates/guest-control-server/tests/connection/guest_storage_manifest.rs
+index a35acfc499..05de5df59c 100644
+--- a/crates/guest-control-server/tests/connection/guest_storage_manifest.rs
++++ b/crates/guest-control-server/tests/connection/guest_storage_manifest.rs
+@@ -106,6 +106,24 @@ printf 'helper stderr\n' >&2
+     assert_eq!(result.stderr, b"helper stderr\n");
+     assert!(!result.stdout_truncated);
+     assert!(!result.stderr_truncated);
+-    assert!(result.diagnostic.is_empty());
++    let timings: serde_json::Value = serde_json::from_str(&result.diagnostic).unwrap();
++    assert_eq!(timings.as_object().unwrap().len(), 8);
++    for key in [
++        "containment_us",
++        "prepare_spawn_us",
++        "prepare_us",
++        "spawn_us",
++        "io_setup_us",
++        "wait_us",
++        "cleanup_us",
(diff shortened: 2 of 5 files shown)


seven332:
## Code Review: PR #32544

**Reviewed head:** `a99899c4165c03382e5ce365300684f11ac6a0b2`

### Summary

Completed the six focused self-review passes and current-head code-quality review. Scope remains an attribution report plus disposable, explicitly opt-in reproduction artifacts; no product Rust, protocol, containment policy, service, or CI-selection change.

### Findings

No outstanding findings. Self-review fixed the measurement-input boundary: null/incomplete/non-numeric phase diagnostics and mixed diagnostic variants now fail validation, with CLI regressions. Original measured statistics are unchanged.

### Testing

Validated all eleven formal real-guest cohorts (3,564 complete successful pairs), CLI regression tests, Bash syntax, ShellCheck, Markdown formatting, and applicability of both pinned patches. Experimental Rust formatting/Clippy/rustdoc and the relevant storage/connection suites passed as recorded in the PR.

The report correctly separates first/resumed/repeated and fallback samples, uses paired residuals, and does not claim a production cause or an optimization gain. Cold cohorts remain small; in-guest PSI/strace are unavailable, and the tiny final diagnostic-serialization tail is outside the named phase spans. CI readiness is a separate gate.

### Verdict

LGTM for this head. No merge requested or performed.

seven332:
Closing this documentation-only PR at the user's request. The complete measured attribution, validation, limits and no-runtime-change decision are now recorded directly in #32428. No merge or runtime optimization is needed for this investigation. The issue and parent #24203 remain open; the experiment branch is retained as an optional archive.

PR #32546, opened 2026-09-08
fix(api): use current workflow schedules in completion callbacks

lancy:
When a running workflow automation is edited from 09:30 Asia/Shanghai to 09:00 in the UI, its completion callback combines the old cron with the newly stored UTC timezone and overwrites the next run with 17:30.

Recompute recurrence from the current schedule type, cron, timezone, and interval while holding a row lock in one transaction, so a concurrent schedule edit cannot be overwritten from a stale read. Preserve a newly configured one-time schedule and keep the existing callback payload format compatible with already dispatched runs.

## Validation

- 111 API integration tests passed across the workflow automation scheduler and automation routes, including four new regressions for editing a running schedule and changing between cron, loop, and once.
- Regression assertions wait for callback delivery before inspecting the persisted next run.
- Scoped Oxlint, type-aware Oxlint, ESLint, and API type checks passed.
- Scoped Knip, formatting, and `git diff --check` passed.

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/workflow-automation-scheduler.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/workflow-automation-scheduler.test.ts
@@ -1,7 +1,10 @@
 import { createHash, randomUUID } from "node:crypto";
 
 import { testWorkflowAutomationExecutionContract } from "@okouai/api-contracts/contracts/test-workflow-automation-execution";
-import { workflowAutomationsContract } from "@okouai/api-contracts/contracts/workflows";
+import {
+  workflowAutomationsContract,
+  type WorkflowSchedule,
+} from "@okouai/api-contracts/contracts/workflows";
 import {
   agentsByIdContract,
   agentsMainContract,
@@ -279,6 +282,27 @@ async function completeRunThroughSandbox(
     sandboxHeaders,
     [200],
   );
+}
+
+async function waitForScheduleCallback(
+  scenario: Scenario,
+  runId: string,
+): Promise<void> {
+  await expect
+    .poll(async () => {
+      const callbacks = await store.set(
+        readAgentRunCallbacks$,
+        { orgId: scenario.orgId, userId: scenario.userId, runId },
+        context.signal,
+      );
+      return callbacks.find((callback) => {
+        return (
+          callback.internalKind === "workflow-automation:cron" ||
+          callback.internalKind === "workflow-automation:loop"
+        );
+      })?.status;
+    })
+    .toBe("delivered");
 }
 
 async function deleteWorkflowViaApi(scenario: Scenario): Promise<void> {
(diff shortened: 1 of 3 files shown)


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: f0cd542ed707e4748abe89a5d2a7213edea3aa33
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

The completion callback now advances a workflow automation using its current persisted schedule, keeping cron, timezone, schedule type, and interval from the same row. A transaction with a row lock serializes this calculation with schedule edits; a schedule changed to a one-time run retains its configured next occurrence. Reviewed all three changed files and the scheduler, edit, callback-dispatch, and test-helper callers against the practice documents fetched from `main` at `18ef6d1d15ad500b5ae42d69e73251503375889f`.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness, races, and edge cases | Pass | `handleWorkflowAutomationInternalCallback` reads and updates the same locked row. An edit committed first is read by the callback; an edit arriving later writes after the callback commits. Current cron/loop fields stay consistent, while disabled, deleted, and one-time automations are skipped. |
| `REVIEW.md` | Security and authorization | Pass | No new endpoint, credential, or authorization path. Automation IDs remain schema-validated and bound through Drizzle; failed-run lookup retains the organization filter. |
| `REVIEW.md` | Type safety and maintainability | Pass | The helper accepts the persisted nullable cron field without assertions or `any`; its only caller was checked. The callback remains one bounded operation, with comments explaining the lock and the test synchronization. |
| `REVIEW.md`, `docs/testing.md` | A bug fix requires behavioral regression coverage | Pass | Four route-level cases cover the reported cron/timezone edit and cron-to-loop, loop-to-cron, and cron-to-once edits during an active run. Existing recurring completion, credit recovery, and failure-disable tests pass. |
| `REVIEW.md` | Feature containment and DB/JSONB contract gates | N/A | This is a correction to existing scheduling behavior, with no new feature, public contract, JSONB shape, or schema migration. |
| `docs/docs.md` | Load the practice documents for every changed surface | Pass | Loaded the API, TypeScript testing, ccstate/lifecycle, persisted compatibility, fallback, and database guidance. No UI, event projection, connector-account authority, or runner protocol is changed. |
| `docs/bad-smell.md` | Keep production logic explicit, typed, and focused | Pass | No dependency, generic abstraction, lint suppression, catch-all, or synthetic default is introduced. The existing schedule calculator and Drizzle builders are reused. |
| `docs/testing.md`, `docs/testing/patterns.md` | Exercise real application and database behavior; mock only external boundaries | Pass | Tests create and edit through the automation API, execute through the existing scheduler test entry, and complete through runner/webhook APIs using a real PostgreSQL database. Existing centralized external mocks and cleanup are retained. |
| `docs/testing/anti-patterns.md` | No internal mocking, fake timers, or mock-call-only assertions | Pass | New cases use the existing application clock override while timers remain real. They assert the persisted schedule, enabled state, and exact next occurrence through the automation API. |
| `docs/testing/api-testing.md`, `docs/testing/testing-external-behavior.md` | Construct and verify behavior through endpoints; justify infrastructure-only observation | Pass | Product setup and final assertions use real API contracts. The existing test-only callback read endpoint supplies only an asynchronous-delivery barrier: the edit already seeds the expected next time, so a product read alone cannot establish that the old callback finished. No internal state is written by the new cases. |
| `docs/effect.md` | Give asynchronous work an explicit semantic owner | Pass | Completion remains one awaited domain operation through the existing command. No detached continuation or second mutable schedule source is introduced. |
| `docs/cache.md` | Bound resources and avoid duplicate mutable authority | Pass | The transaction owns and releases the row lock. No cache, timer, subscription, or persistent in-process resource is added; the database row supplies the complete schedule. |
| `.claude/skills/ccstate/SKILL.md` | Preserve command boundaries, cancellation, and awaited work | Pass | The command forwards its lifecycle signal; database awaits retain abort checks. Both ccstate and direct callback dispatch paths were checked, and no accessor escapes or floating promises are added. |
| `docs/deployment-compatibility.md` | Preserve reachable cross-version and persisted contracts | Pass | Existing callback schemas and producers remain unchanged, so already queued callbacks remain readable. Only the source of schedule calculation changes; existing database columns and API response shapes remain compatible. |
| `docs/fallback.md` | Do not invent defaults or undeclared compatibility paths | Pass | No new fallback is introduced. The one-time branch handles a supported current schedule type; regression tests assert its configured future run rather than the absence of retired behavior. |
| `.claude/skills/database-development/SKILL.md` | Use typed builders and preserve database boundaries | Pass | Selection uses schema-owned columns and `.for("update")`; read, calculation, and update share one transaction. No raw SQL, cast-based decoder, schema change, or network call under the row lock is introduced. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- Coverage: adequate. Targeted `workflow-automation-scheduler.test.ts` and `workflow-automations.test.ts` passed locally: **111 tests across two files**. The reported edit from Shanghai 09:30 to 09:00 now yields `2026-09-09T01:00:00.000Z` after the old callback completes.
- Conventions: compliant. The callback-delivery barrier prevents the regression from passing before a delayed stale overwrite; product assertions use the automation API.
- Scoped ESLint, Oxlint, type-aware Oxlint, formatting, and `git diff --check` passed. The API workspace type check and Knip check passed.
- GitHub API/app test shards and the security/crates gates have passed. The remaining runner/associated E2E pipeline is still pending at review time; merge admission remains gated on required CI.
- The integration cases cover edits committed before completion; row-lock serialization with an overlapping write was verified by tracing the transaction and edit paths, without a separate concurrency stress test.

PR #32551, opened 2026-09-08
test(guest): make final telemetry producer-join coverage deterministic

### LINKED ISSUE #32528: test(rust): make final telemetry producer-join coverage deterministic
## Summary

`final_telemetry_waits_for_background_producers_before_upload` does not establish that the metrics producer is still running when shutdown reaches its join. The mock writes its final marker immediately after cancellation, so it can finish during the preceding heartbeat cleanup. The marker assertion can then pass without the producer join it is intended to protect.

This is a regression-test effectiveness issue. The current production code does await the producer; no production telemetry loss was reproduced.

## Evidence

Audited `main` at `1ca1499be6c86b0a3faeb0d5e4bbf4f55c01e787`.

- [`crates/guest-agent/src/main.rs:1468-1534`](https://github.com/vm0-ai/vm0/blob/1ca1499be6c86b0a3faeb0d5e4bbf4f55c01e787/crates/guest-agent/src/main.rs#L1468-L1534): the test uses a current-thread runtime. Its metrics task only awaits the cancellation token, synchronously writes the marker, and exits (lines 1504-1509). There is no gate holding that producer open after cancellation.
- [`crates/guest-agent/src/main.rs:989-1004`](https://github.com/vm0-ai/vm0/blob/1ca1499be6c86b0a3faeb0d5e4bbf4f55c01e787/crates/guest-agent/src/main.rs#L989-L1004): shutdown cancels the token, awaits heartbeat cleanup, then joins metrics before final telemetry. [`stop_heartbeat`, lines 1034-1042](https://github.com/vm0-ai/vm0/blob/1ca1499be6c86b0a3faeb0d5e4bbf4f55c01e787/crates/guest-agent/src/main.rs#L1034-L1042) aborts and awaits the heartbeat task, giving the metrics task an earlier opportunity to run to completion.
- The test's observable assertion is one HTTP request containing the marker (line 1534). A valid schedule is cancellation -> heartbeat cleanup yields -> metrics writes and exits -> final upload sees the marker. The assertion does not distinguish this schedule from one that genuinely needs to wait for unfinished production.
- [`crates/guest-agent/src/telemetry.rs:321-339`](https://github.com/vm0-ai/vm0/blob/1ca1499be6c86b0a3faeb0d5e4bbf4f55c01e787/crates/guest-agent/src/telemetry.rs#L321-L339) and [lines 396-406](https://github.com/vm0-ai/vm0/blob/1ca1499be6c86b0a3faeb0d5e4bbf4f55c01e787/crates/guest-agent/src/telemetry.rs#L396-L406) show why ordering matters: final flush consumes the EOF tail and then stops the uploader.

Consequently, removing the metrics join, or moving it after final flush, can escape this test. The neighboring final-telemetry tests cover upload success/failure and recursive telemetry suppression; the heartbeat test covers aborting a pending heartbeat. No other test of this orchestrator was found that holds a producer pending across the final-flush boundary.

## Impact

The real agent uses a multi-thread runtime, and a metrics tick can already be collecting or writing when cancellation arrives. If shutdown stops awaiting that work, the EOF-consuming final upload can finish before the last append, leaving telemetry that will never be uploaded. The existing test does not reliably protect that ordering contract.

## Acceptance criteria

- Hold the producer alive after it observes shutdown, using an explicit acknowledgement and release gate.
- Make entry into final flush synchronously observable, or use an equivalent controllable seam, so starting final flush while the producer remains blocked fails the test.
- Release the producer, write its final marker, and assert that shutdown completes and the final upload contains that marker exactly once.
- Use bounded waits and deterministic synchronization rather than sleeps or arbitrary scheduler yields.
- Verify that removing the producer join or placing final flush before it causes the regression test to fail.

## Validation and deduplication

This finding is based on source and control-flow review; no test suite or mutation run was executed, and repository code was not changed. Open issues were searched by the exact test name and title keywords for telemetry tests, producers, and final telemetry. The matching open-PR search result concerns Runner SSH execution and does not change this Guest test.

seven332:
## Summary

Closes #32528.

Strengthen `final_telemetry_waits_for_background_producers_before_upload` so it cannot pass merely because the metrics producer finishes during heartbeat cleanup.

- Acknowledge producer cancellation and keep it blocked until an explicit release.
- Settle the heartbeat fixture before explicitly polling shutdown, and use the existing synchronous final-upload log to detect premature flush entry. Assert that the same log exists after successful shutdown so removing it cannot silently weaken the test.
- Keep real file writes, telemetry serialization, and HTTP upload; assert exactly one upload with exactly one final producer marker.
- Isolate this test's runtime files in a temporary directory instead of mutating the process environment.

Only the existing test changes. Production shutdown behavior, runtime performance, API contracts, dependencies, CI selection, and the five-second timeout are unchanged.

## Validation

- `cargo clippy --profile local -p guest-agent --all-targets --all-features`
- `cargo doc --profile local -p guest-agent --all-features --no-deps`
- `cargo test --profile local -p guest-agent -- --quiet`: 1,049 passed; 4 existing ignored fixtures unchanged.
- Focused regression test passed 10 consecutive times after restoration.
- Mutation validation: replacing the metrics join with `drop(metrics_handle)` and moving final telemetry before the join both failed immediately with `final telemetry started before the producer was released` (0.12s and 0.05s respectively). Both temporary mutations were restored before submission.
- `git diff --check`

## Risk

The deterministic checkpoint intentionally relies on the documented synchronous log append and the current-thread runtime. The test checks both absence before producer release and presence after completion. No production telemetry loss was reproduced or fixed by this test-only PR.

### DIFF
--- a/crates/guest-agent/src/main.rs
+++ b/crates/guest-agent/src/main.rs
@@ -1475,57 +1475,96 @@ mod tests {
             .block_on(async {
                 let server = &*COMPLETE_EXECUTION_MOCK_SERVER;
                 server.reset_async().await;
-                let _env_guard = unsafe { set_test_env(server, None) };
-                let guest_paths = test_guest_paths();
+                let tmp = tempfile::tempdir().unwrap();
+                let guest_paths = paths::GuestPaths::from_runtime_dir(tmp.path());
+                let system_log_path = Path::new(guest_paths.system_log_file());
+                paths::write_private(system_log_path, "").unwrap();
+                let _system_log_guard = SystemLogOverrideGuard::set(system_log_path);
                 let _sandbox_ops_guard = SandboxOpsOverrideGuard::set(std::path::Path::new(
                     guest_paths.sandbox_ops_file(),
                 ));
 
                 let marker = "producer_after_shutdown_before_final_upload";
-                let cleanup_paths = vec![
-                    guest_paths.sandbox_ops_file().to_string(),
-                    guest_paths.telemetry_system_log_pos_file().to_string(),
-                    guest_paths.telemetry_metrics_pos_file().to_string(),
-                    guest_paths.telemetry_sandbox_ops_pos_file().to_string(),
-                ];
-                for path in &cleanup_paths {
-                    let _ = std::fs::remove_file(path);
-                }
-
+                let request_bodies = Arc::new(std::sync::Mutex::new(Vec::new()));
+                let request_bodies_for_mock = Arc::clone(&request_bodies);
                 let telemetry_mock = server.mock(|when, then| {
                     when.method(POST)
-                        .path("/api/webhooks/agent/telemetry")
-                        .body_includes(marker);
-                    then.status(200)
-                        .header("Content-Type", "application/json")
-                        .json_body(json!({}));
+                        .path("/api/webhooks/agent/telemetry");
+                    then.respond_with(move |request| {
+                        request_bodies_for_mock.lock().unwrap().push(request.body_vec());
+                        HttpMockResponse::builder()
+                            .status(200)
+                            .header("Content-Type", "application/json")
+                            .body("{}")
+                            .build()
+                    });
                 });
 
                 let shutdown = CancellationToken::new();
                 let producer_shutdown = shutdown.clone();
+                let (producer_cancelled_tx, producer_cancelled_rx) = tokio::sync::oneshot::channel();
+                let (release_producer_tx, release_producer_rx) = tokio::sync::oneshot::channel();
                 let metrics_handle = tokio::spawn(async move {
                     producer_shutdown.cancelled().await;
+                    producer_cancelled_tx.send(()).unwrap();
+                    release_producer_rx.await.unwrap();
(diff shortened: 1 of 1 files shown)


seven332:
## Code Review: PR #32551

**Reviewed head:** `46f47d96a44cfd03bb165120b48a678e2c036698`

### Summary

Completed all six focused self-review passes and the code-quality review on the unchanged submitted head. The producer gate, settled heartbeat, and synchronous log checkpoint make the existing ordering regression deterministic while preserving real telemetry upload and production behavior.

### Findings

No findings. No follow-up issue required.

### Testing

1,049 guest-agent tests passed; four existing ignored fixtures unchanged. Ten focused repetitions passed. Both omitted-join and flush-before-join mutations failed immediately at the pre-release ordering assertion and were restored. Scoped and native workspace formatting, Clippy, documentation, file-size, and commit checks passed. The documented log/current-thread coupling is paired with a positive final-log assertion. CI readiness is evaluated separately.

### Verdict

LGTM. No merge authorization implied.

PR #32555, opened 2026-09-08
fix(guest-agent): handle short reads in streaming upload buffers

### LINKED ISSUE #32531: bug(rust): handle short reads when reserving streaming upload buffers
## Summary

`SizedBody::poll_frame` can reject a legal short file read with `failed to reserve streaming upload buffer`, even though the source file is intact and the reader has not reached EOF. The buffer reservation treats `BytesMut::reserve`'s argument as capacity growth rather than additional space relative to the current length.

Audited commit: `1ca1499be6c86b0a3faeb0d5e4bbf4f55c01e787`.

## Location and root cause

- [`crates/guest-agent/src/http.rs:927-937`](https://github.com/vm0-ai/vm0/blob/1ca1499be6c86b0a3faeb0d5e4bbf4f55c01e787/crates/guest-agent/src/http.rs#L927-L937): the capacity check calls `buffer.reserve(to_read - spare_capacity)` before requesting a `to_read`-sized spare slice.
- [`crates/guest-agent/src/http.rs:950-958`](https://github.com/vm0-ai/vm0/blob/1ca1499be6c86b0a3faeb0d5e4bbf4f55c01e787/crates/guest-agent/src/http.rs#L950-L958): `split_to(n)` leaves the unused capacity in the next buffer after a partial read.

For a 512 KiB source and a first read of 8 KiB:

1. The initial buffer capacity is 256 KiB.
2. Splitting off the 8 KiB frame leaves an empty buffer with 248 KiB of spare capacity; more than 256 KiB of source data remains.
3. The next poll needs 256 KiB but calls `reserve(8 KiB)`.
4. `BytesMut::reserve(additional)` ensures space for `len + additional`. Since the existing 248 KiB already exceeds 8 KiB, this call does nothing.
5. `spare_capacity_mut().get_mut(..256 KiB)` fails, producing a body error instead of continuing the upload.

An asynchronous file read is allowed to return fewer bytes than requested. Such a read is not an allocation failure or necessarily EOF.

## Reproduction and evidence

An isolated Rust test compiled against the checkout's existing dependencies reproduced this using the **unchanged, source-extracted `SizedBody` implementation**. The test opens a real 512 KiB regular file and uses Tokio's `set_max_buf_size(8192)` to force legal short reads deterministically. The same test without that read-size limit passes.

Equivalent regression test for the `http.rs` unit-test module:

```rust
#[tokio::test]
async fn sized_body_preserves_contents_after_short_file_reads() {
    let temp = tempfile::tempdir().unwrap();
    let path = temp.path().join("source.bin");
    let expected = vec![42_u8; STREAM_CHUNK_SIZE * 2];
    tokio::fs::write(&path, &expected).await.unwrap();

    let mut file = tokio::fs::File::open(path).await.unwrap();
    file.set_max_buf_size(8192);
    let mut body = SizedBody::new(file, expected.len() as u64);
    let mut actual = Vec::new();
    while let Some(frame) = std::future::poll_fn(|cx| {
        http_body::Body::poll_frame(std::pin::Pin::new(&mut body), cx)
    })
    .await
    {
        let bytes = frame.expect("legal short read must not fail").into_data().unwrap();
        actual.extend_from_slice(&bytes);
    }
    assert_eq!(actual, expected);
}
```

Observed result:

```text
legal short reads must not cause a buffer reservation error:
Custom { kind: Other, error: "failed to reserve streaming upload buffer" }

short-read test: FAILED
ordinary full-read control: ok
```

The 10 existing `guest-agent` HTTP unit tests and 17 presigned-upload integration cases passed. This reproduction tests the upload-body logic, not an end-to-end production upload. The read-size setting is a test control; this audit did not establish production incidence or claim that ordinary file uploads always fail.

## Impact and actual call path

[`HttpClient::put_presigned_file`, `http.rs:983-1011`](https://github.com/vm0-ai/vm0/blob/1ca1499be6c86b0a3faeb0d5e4bbf4f55c01e787/crates/guest-agent/src/http.rs#L983-L1011) wraps a newly cloned source file in `SizedBody` for each upload attempt. A qualifying short read aborts that attempt. If this condition repeats through the retry budget, a valid source cannot be uploaded.

[`upload_archive_bundle`, `artifact/mod.rs:469-502`](https://github.com/vm0-ai/vm0/blob/1ca1499be6c86b0a3faeb0d5e4bbf4f55c01e787/crates/guest-agent/src/artifact/mod.rs#L469-L502) uses this path for archives and manifests and propagates exhausted upload errors as checkpoint failures.

## Suggested fix and acceptance criteria

- Reserve the required additional space relative to buffer length, e.g. `buffer.reserve(to_read)` when spare capacity is insufficient, rather than subtracting the existing spare capacity from the argument.
- Add deterministic short-read coverage that consumes the complete body and compares all bytes and the remaining size hint. Cover both retained prior frames and released frames.
- Keep empty-file, ordinary full-chunk, final partial-chunk, premature-EOF, and retry behavior intact; retain bounded streaming rather than loading the whole file into memory.

No repository code was changed during this audit.

seven332:
## Summary

Closes #32531.

Fix the streaming upload body's reservation after a legal short file read.
`BytesMut::reserve` takes writable space relative to length, not a capacity
deficit. The old calculation could leave too little spare space and abort a
valid archive or manifest upload with `failed to reserve streaming upload buffer`.

- Reserve the next bounded read's full writable size when capacity is insufficient.
- Extend real-file Body coverage to normal and forced 8 KiB reads with released
  and retained frames, checking exact bytes, remaining size hints, bounded chunks,
  terminal state, and retained-frame integrity after later reads.
- Verify complete patterned 600000-byte uploads and full replay on retry through
  the public HTTP API and the existing mock server.

No changes to chunk size, source identity/length, Content-Length, authentication,
timeouts, retry policy, API/protocols, dependencies, or CI. No migration or
cross-version compatibility code is required.

## Validation

- Regression failed on unchanged production code with the reported buffer error;
  the same test passes after the fix (all four read/ownership combinations).
- `cargo clippy --profile local -p guest-agent --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS='-D warnings' cargo doc --profile local --no-deps -p guest-agent --all-features`
- `cargo test --profile local -p guest-agent --all-targets --all-features -- --quiet`
- `git diff --check`

All selected checks passed, including 650 library tests, 15 command-entry tests,
136 main integration cases, and the standalone integration targets. No new
ignored tests or suppressions. Workspace Clippy/rustdoc and commit hooks also
passed. The required post-pull local DB migration hit the
existing `chat_thread_event_kind` enum conflict (42710); these Rust checks do not
depend on PostgreSQL, and no DB repair is included.

## Risks and limitations

Correct reservation can allocate when earlier short frames remain held; dropped
views can be reclaimed. Reads remain bounded at 256 KiB and the body does not
preload the source. Existing premature-EOF and retry behavior remain covered.
Production incidence and production performance were not measured.

### DIFF
--- a/crates/guest-agent/src/http.rs
+++ b/crates/guest-agent/src/http.rs
@@ -928,7 +928,8 @@ impl http_body::Body for SizedBody {
         let to_read = next_chunk_size(*this.remaining);
         let spare_capacity = this.buffer.capacity() - buffer_len;
         if spare_capacity < to_read {
-            this.buffer.reserve(to_read - spare_capacity);
+            // `reserve` takes additional space relative to length, not capacity.
+            this.buffer.reserve(to_read);
         }
 
         let Some(spare) = this.buffer.spare_capacity_mut().get_mut(..to_read) else {
@@ -1118,24 +1119,42 @@ mod tests {
         let data: Vec<u8> = (0..(STREAM_CHUNK_SIZE * 2 + 37))
             .map(|i| (i % 251) as u8)
             .collect();
-        let (_dir, mut body) = sized_body_from_bytes(&data).await;
+        for read_size in [STREAM_CHUNK_SIZE, 8192] {
+            for retain_frames in [false, true] {
+                let (_dir, mut body) = sized_body_from_bytes(&data).await;
+                // Force legal short reads through the real file reader.
+                body.reader.set_max_buf_size(read_size);
 
-        let mut remaining = data.len() as u64;
-        assert_eq!(body.size_hint().exact(), Some(remaining));
+                let mut remaining = data.len() as u64;
+                assert_eq!(body.size_hint().exact(), Some(remaining));
+                assert!(!body.is_end_stream());
 
-        let mut chunks = 0;
-        let mut uploaded = Vec::with_capacity(data.len());
-        while let Some(chunk) = next_data(&mut body).await {
-            assert!(chunk.len() <= STREAM_CHUNK_SIZE);
-            chunks += 1;
-            remaining = remaining.saturating_sub(chunk.len() as u64);
-            assert_eq!(body.size_hint().exact(), Some(remaining));
-            uploaded.extend_from_slice(&chunk);
+                let mut chunks = 0;
+                let mut uploaded = Vec::with_capacity(data.len());
+                let mut retained = Vec::new();
+                while let Some(chunk) = next_data(&mut body).await {
+                    assert!(!chunk.is_empty());
+                    assert!(chunk.len() <= read_size);
+                    chunks += 1;
+                    remaining -= chunk.len() as u64;
+                    assert_eq!(body.size_hint().exact(), Some(remaining));
+                    assert_eq!(body.is_end_stream(), remaining == 0);
+                    uploaded.extend_from_slice(&chunk);
+                    if retain_frames {
+                        retained.push(chunk);
+                    }
+                }
+
+                assert!(chunks > 1);
+                assert_eq!(uploaded, data);
+                if retain_frames {
+                    assert_eq!(retained.concat(), data);
+                }
+                assert_eq!(body.size_hint().exact(), Some(0));
+                assert!(body.is_end_stream());
+                assert!(next_data(&mut body).await.is_none());
+            }
         }
-
-        assert!(chunks > 1);
-        assert_eq!(uploaded, data);
-        assert_eq!(body.size_hint().exact(), Some(0));
     }
 
     #[tokio::test]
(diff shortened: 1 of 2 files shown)


seven332:
## Code Review: PR #32555

**Reviewed head:** `ac03ff1a751d2f3159a9017eced973d0443ad236`

### Summary

Completed all six focused self-review passes and the repository code-quality
review. Reservation now guarantees the requested writable slice after short
reads while preserving initialized-byte ownership, exact length, EOF and retry
behavior. Scope matches the challenged #32531 plan; no API or format change.

### Findings

No findings. No review fixes or follow-up issues needed. Held short frames may
retain backing allocations; the PR discloses this without claiming a fixed
aggregate-memory limit. No new fallback or compatibility branch is introduced.

### Testing

The real-file regression failed before the fix and passed afterward for full
and 8 KiB reads with retained and released frames. Public HTTP integration
checks full patterned contents and Content-Length on initial and retry uploads.
All guest-agent tests, fmt, Clippy, rustdoc, workspace static checks and commit
hooks passed. Production incidence/performance and remote smoke are unverified;
CI readiness is checked separately.

### Verdict

LGTM

PR #32560, opened 2026-09-08
fix(rust): enforce exec-control request deadlines

### LINKED ISSUE #32532: bug(rust): exec-control partial I/O can exceed the request deadline
## Summary

Exec-control forwarding does not enforce its absolute request deadline across partial process-control socket I/O. A valid fragmented response can be reported as `Delivered` well after the local sink budget has expired, while keeping the sink connected.

Confirmed at `1ca1499be6c86b0a3faeb0d5e4bbf4f55c01e787` during the Rust concurrency audit.

## Code and cause

- `crates/guest-control-server/src/exec_control/mod.rs:80-84` creates one absolute deadline from the incoming `request_timeout_ms`.
- `crates/guest-control-server/src/exec_control/forward.rs:212-229` calculates the remaining budget once before reading the response. Lines 305-310 set `SO_RCVTIMEO` through `set_read_timeout` and call `process_control_ipc::read_response`.
- `crates/process-control-ipc/src/codec.rs:217-228` reads the length prefix and body with separate `read_exact` calls. Each can perform multiple reads. The socket timeout limits an individual blocking read, not the total frame duration; partial progress starts another read with the same timeout.
- `crates/guest-control-server/src/exec_control/forward.rs:238-242` accepts the decoded response without checking whether the deadline expired. The analogous one-time write timeout at lines 179-196 and 296-302 also does not provide a whole-frame deadline for `write_all`.

The protocol explicitly defines this as the local sink budget after guest receipt (`crates/guest-control-proto/src/payloads/exec_operation/control.rs:337-339`), not an inactivity timeout.

## Reproduction

An external probe compiled the unmodified `exec_control` module (including forwarding and sink state), `GuestWriter`, and the two protocol crates from the audited checkout. No repository code was changed.

1. Create a connected control sink with `UnixStream::pair()` and reserve its pending slot, as in the existing forwarding tests.
2. Run `forward_control_request` with a 400 ms deadline and message ID `m`.
3. Read the outgoing request at the peer, then send a valid `Accepted` response one byte every 100 ms. The full frame is `00 00 00 08 01 03 00 01 6d 00 00 00`.
4. Decode the emitted host-facing exec-control result and inspect sink state.

Observed:

```text
request_timeout_ms=400
forward_elapsed_ms=1202
result_status=Delivered
sink_still_connected=true
pending_after_forward=0
```

The existing `timed_out_control_sink_is_marked_failed` test covers a completely silent peer, so it does not exercise this partial-progress case. A separate probe compiling the sampled codec directly also accepted the same fragmented response after 1,202 ms with a 400 ms read timeout.

## Impact

A slow or malfunctioning sink can retain the forwarding worker and the serialized stream beyond an expired request's budget. Subsequent controls can time out waiting for that stream, and the original request can still produce a late success instead of the required timeout/failure. This is not evidence of a production incident; it is a reproduced deadline-contract violation.

## Acceptance criteria

- Carry one absolute deadline through all partial request writes and response reads, including the length prefix and body. Recompute the remaining budget before each blocking operation, or use deadline-aware nonblocking I/O.
- On expiry after connected-stream I/O has started, report `SinkTimeout` and fail/close the sink so partial frames cannot contaminate a later exchange.
- Do not report `Delivered` for a response completed after the deadline.
- Add regression coverage for fragmented responses and partial/backpressured request writes that continue making progress below the per-I/O timeout but exceed the total request budget. Preserve existing silent-peer, message-ID mismatch, and pre-I/O timeout behavior.

seven332:
## Summary

- Enforce the existing absolute exec-control request deadline across partial request writes, response headers/bodies, and readiness waits.
- Use a private Read/Write adapter with per-call nonblocking socket flags and bounded poll waits. Preserve the five-second per-I/O cap without changing shared socket modes or adding threads/dependencies.
- Keep framing in process-control-ipc by generalizing only the needed codec entry points. Reject late decoded responses; fail/shut down a sink after I/O has started, while preserving no-I/O timeout reuse and operation-deactivation precedence.
- Add real-socket regressions for fragmented response headers/bodies, progressive backpressured writes, connected pre-I/O expiry, and successful fragmented/repeated delivery.

Closes #32532

## Verification

All commands ran from crates/ unless noted:

- cargo test --profile local -p process-control-ipc -p guest-control-server --all-features
- cargo test --profile local -p guest-agent --lib control::
- cargo test --profile local -p guest-tool-exec -p guest-control-tests
- cargo clippy --profile local --all-targets --all-features
- cargo doc --profile local --workspace --all-features --no-deps
- cargo xtask check-rust-path-attributes
- cargo fmt --all --check
- scripts/prepare.sh (repository root)

393 selected tests passed. Three existing privileged production-identity tests are executed explicitly by Rust CI; one existing optional file-write benchmark is ignored. This PR adds no skips or timeout increases.

The four new regressions additionally passed ten sequential repetitions (40/40).

Regression evidence: original forwarding returns Delivered for both late fragmented responses and writes the complete 1,048,589-byte backpressured request after its 400 ms budget. Updating SO_SNDTIMEO before each blocking Write call still failed the write regression, so the challenged plan was revised to per-call nonblocking I/O. The final implementation stops the incomplete write and returns SinkTimeout.

## Scope and risks

No wire-format/status/API/persisted-state changes, protocol negotiation, handshake-budget changes, or host-facing result-writer changes. No deployment ordering requirement. Slow sinks previously accepted after expiry now fail for that exec operation. Timeout does not prove the sink has not acted, so no automatic retry is introduced. OS scheduling can delay observation of a deadline, but a late decoded response is not accepted. Production/privileged validation remains with CI; no production soak is claimed.

### DIFF
--- a/crates/guest-control-server/src/exec_control/deadline_io.rs
+++ b/crates/guest-control-server/src/exec_control/deadline_io.rs
@@ -1,0 +1,108 @@
+use std::io::{self, Read, Write};
+use std::os::fd::{AsRawFd, RawFd};
+use std::os::unix::net::UnixStream;
+use std::time::Instant;
+
+use super::{CONTROL_SINK_IO_TIMEOUT, duration_until, request_timeout_error};
+
+/// Applies one request budget across codec reads, writes, and partial progress.
+/// Per-call nonblocking flags leave the mode of all cloned handles unchanged.
+pub(super) struct DeadlineStream<'a> {
+    stream: &'a mut UnixStream,
+    deadline: Instant,
+    pub(super) io_started: bool,
+}
+
+impl<'a> DeadlineStream<'a> {
+    pub(super) fn new(stream: &'a mut UnixStream, deadline: Instant) -> Self {
+        Self {
+            stream,
+            deadline,
+            io_started: false,
+        }
+    }
+
+    fn perform_io(
+        &mut self,
+        events: libc::c_short,
+        mut operation: impl FnMut(RawFd) -> libc::ssize_t,
+    ) -> io::Result<usize> {
+        let deadline = self.deadline.min(Instant::now() + CONTROL_SINK_IO_TIMEOUT);
+        loop {
+            duration_until(deadline).ok_or_else(request_timeout_error)?;
+            self.io_started = true;
+            let result = operation(self.stream.as_raw_fd());
+            if result >= 0 {
+                return Ok(result as usize);
+            }
+            let error = io::Error::last_os_error();
+            match error.kind() {
+                io::ErrorKind::Interrupted => continue,
+                io::ErrorKind::WouldBlock => self.wait_ready(events, deadline)?,
+                _ => return Err(error),
+            }
+        }
+    }
+
+    fn wait_ready(&self, events: libc::c_short, deadline: Instant) -> io::Result<()> {
+        loop {
+            let remaining = duration_until(deadline).ok_or_else(request_timeout_error)?;
+            let timeout_ms = remaining.as_millis().clamp(1, libc::c_int::MAX as u128);
+            let mut descriptor = libc::pollfd {
+                fd: self.stream.as_raw_fd(),
+                events,
+                revents: 0,
+            };
+            // SAFETY: descriptor is initialized and borrowed for one entry;
+            // the timeout is a bounded, positive millisecond count.
+            let result = unsafe { libc::poll(&mut descriptor, 1, timeout_ms as libc::c_int) };
+            if result < 0 {
+                let error = io::Error::last_os_error();
+                if error.kind() == io::ErrorKind::Interrupted {
+                    continue;
+                }
+                return Err(error);
+            }
+            if result > 0 {
+                // Retry recv/send for readiness, EOF, or socket errors. The
+                // caller checks the deadline again before that nonblocking I/O.
+                return Ok(());
+            }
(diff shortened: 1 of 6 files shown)


seven332:
## Code Review: PR #32560

**Reviewed head:** `9de0796e8d28a5ea76c6a09b758d4347bb42dc11`

### Summary

One full six-pass self-review loop is clean on this unchanged submitted head. The nonblocking adapter keeps one request deadline through partial I/O and readiness retries, preserves no-I/O reuse and terminal sink cleanup, and leaves the wire format unchanged. Reviewed syscall safety, bounded retries, serialization/lifecycle interactions and codec consumers.

### Findings

No outstanding findings. Self-review corrected one test synchronization issue: the slow-drain window now begins after observing request bytes, not before the forwarding thread runs. No timeout was increased and no production behavior changed in that correction. No follow-up issue needed.

### Testing

393 selected tests passed across process-control-ipc, guest-control-server, guest-agent control, guest-tool-exec and guest-control-tests. After the test correction, all 342 core tests and ten repetitions of the four new regressions (40/40) passed again. Workspace formatting, Clippy/all-targets/all-features, docs, path checks and commit hooks passed. Three existing privileged identity tests run explicitly in CI; one existing optional benchmark is ignored. CI/production readiness is separate from this code verdict; no production soak or hard real-time scheduling guarantee is claimed.

### Verdict

LGTM for this head. Merge still requires green required checks, resolved feedback and separate user approval.

PR #32624, opened 2026-09-08
fix(platform): make annotation arrows and strokes editable

tongx-blip:
## What

Three fixes to the image annotation editor, from feedback on the Mark feature.

### 1. Arrows and pen strokes were not selectable or editable

`MarkShape` returned `StrokeMark` **before** applying the interaction props, and `StrokeMark` hard-coded `pointer-events-none` on its `<svg>`. An arrow or a freehand stroke could be drawn and then never clicked, moved, recoloured, deleted from the canvas, or given a note. Two further layers backed this up:

- `rectOf` returned `null` for both shapes, so `grabMark` bailed out before `beginDrag` — the drag never started.
- `moveAnnotationMarkRect$` only handled `box` and `text`; arrow and pen fell through unchanged.

Every stroke now carries a transparent 14px hit band along its own path with `pointer-events: stroke`. That matters: the SVG box spans the whole image, so making the *box* hittable would swallow every press meant for the canvas and leave no way to draw a second mark nearby.

- A selected **arrow** gets grips on its tail and tip, so direction is changed by dragging an end. Dragging the shaft moves the whole arrow without changing where it points.
- A selected **stroke** gets a dashed extent — it has no rectangle to resize and no ends worth grabbing, so this is the whole affordance.
- Both translate every point together, clamped so a mark cannot leave the image (it would survive in the draft but vanish from the flattened copy the model reads).

### 2. One drag was one undo step per frame

`moveAnnotationMarkRect$` called `pushAnnotation$` on **every** pointer move, so undoing a one-second drag took ~60 presses of Cmd+Z. The first move of a gesture now pushes and the rest amend the present in place.

### 3. Keyboard

`Enter` had no binding in the note field at all — the only way out was Escape or clicking off it. Added, plus the gaps around it:

| Key | Action |
| --- | --- |
| `Enter` | Commit the note and dismiss the field (`Cmd/Ctrl+Enter` still attaches) |
| `←↑→↓` | Nudge the open mark; `Shift` for a coarse step |
| `Cmd/Ctrl` `+` / `-` / `0` | Zoom in / out / back to fit — the zoom buttons had no keys |
| `1`–`5` | Pick ink, same as the five swatches |
| `B` `A` `D` `T` | Existing tool bindings, now shown on each tool's tooltip |

## Not included

Making marks reachable with `Tab` (focusable shapes, `Enter` to open a note) is the remaining keyboard gap — there is currently no way to reach a mark without a pointer. That is an a11y change with its own i18n surface, so it is left for a follow-up rather than folded in here.

## Testing

- 3 new integration tests: arrow select + re-aim by tip, stroke select + move, and Enter-commits-note + one-drag-is-one-undo.
- `chat-attachment-annotation`, `chat-attachment-composer`, `chat-composer-connectors`: 32 passed.
- `check-types`, `lint`, `knip`, `format` clean on `@okouai/app`.

Behind the existing `composerImageAnnotation` feature switch.

### DIFF
--- a/turbo/apps/platform/src/signals/okou-page/image-annotation.ts
+++ b/turbo/apps/platform/src/signals/okou-page/image-annotation.ts
@@ -184,16 +184,23 @@ export const ANNOTATION_RESIZE_EDGES = [
 
 export type AnnotationResizeEdge = (typeof ANNOTATION_RESIZE_EDGES)[number];
 
+/** Which end of an arrow a drag is holding. */
+export type AnnotationArrowEnd = "from" | "to";
+
 /**
- * A mark being moved or resized. Drawing produces a new mark; this edits one
- * that already exists, so it carries the mark it started from and the pointer
- * offset at grab time — without the offset a drag snaps the mark's corner to
- * the cursor on the first move.
+ * A mark being moved, resized, or re-aimed. Drawing produces a new mark; this
+ * edits one that already exists, so it carries the mark it started from and the
+ * pointer offset at grab time — without the offset a drag snaps the mark's
+ * corner to the cursor on the first move.
+ *
+ * `endpoint` is the arrow's own mode: an arrow has no rectangle to resize, so
+ * its two ends are dragged directly and the pointer position *is* the new end.
  */
 export interface AnnotationDrag {
   readonly markId: string;
-  readonly mode: "move" | "resize";
+  readonly mode: "move" | "resize" | "endpoint";
   readonly corner?: AnnotationResizeEdge;
+  readonly endpoint?: AnnotationArrowEnd;
   readonly origin: AnnotationPoint;
   readonly startRect: {
     x: number;
@@ -209,8 +216,11 @@ const MAX_NOTE_WIDTH = 1;
 /** Clear of the mark's own outline and its ordinal pin. */
 const NOTE_GAP = 0.015;
 
-/** The box a mark occupies, used to place its note under it. */
-function markBounds(mark: ImageAnnotationMark): {
+/**
+ * The box a mark occupies, used to place its note under it and — for the shapes
+ * that have no rectangle of their own — to give a drag something to translate.
+ */
+export function markBounds(mark: ImageAnnotationMark): {
   x: number;
   y: number;
   width: number;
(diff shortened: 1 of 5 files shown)


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: f8c1f1a54512e12e2e3981702d9b2057fb56f40c
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Makes `arrow` and `pen` annotation marks selectable and editable in the image annotation editor. `MarkShape` previously returned `StrokeMark` before applying its interaction props, and `StrokeMark` hard-coded `pointer-events-none`, so neither shape could be clicked at all; `rectOf` returning `null` and `moveAnnotationMarkRect$` handling only `box`/`text` blocked the drag path behind that. Also collapses a drag into one undo step and fills in the editor's keyboard bindings, starting with `Enter` to commit a note.

### Review Criteria

Applied `docs/bad-smell.md`, `docs/effect.md`, `docs/testing.md`, `docs/testing/app-testing.md`, and `docs/fallback.md`. No DB, contract, event-sourcing, runner, or deployment-boundary surface is touched, so those gates do not apply.

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Nits (P2)

- `turbo/apps/platform/src/signals/okou-page/image-annotation.ts` — `offsetMark` closes its shape switch with `default` covering `box`/`highlight`/`redact`. It is type-safe today (a future rect-less shape would fail to compile on `mark.rect`), but `buildMark` in the editor enumerates every case explicitly, and matching that style would keep the exhaustiveness visible at the call site rather than implied.
- `turbo/apps/platform/src/signals/okou-page/image-annotation.ts` — `offsetMark` clamps a stroke's bounding box to stay fully inside the image, while a `box` mark still takes `draggedRect`'s `clamp01` result verbatim and can therefore be dragged partly off-image. This PR makes strokes strictly better behaved than boxes rather than introducing a regression, but the two paths now disagree about what "off the edge" means. Worth a follow-up to align `box` with the stroke behaviour.

### Testing

- Coverage: Adequate. `fix:` commit with a regression test per fixed behaviour — arrow select + re-aim by tip (`annotation-handle-to` moves to the dragged point), stroke select + move (`annotation-pen-outline` extent shifts), and `Enter` committing a note plus a four-move drag collapsing to a single `Cmd+Z`. Each test asserts the state the bug produced, so all three fail against the parent commit.
- Conventions: Compliant. Page-level integration tests through `setupPage` with the `ComposerImageAnnotation` switch enabled, per `docs/testing/app-testing.md`. No `vi.mock` of internal paths, no `fetch` stubbing, no fake timers, no mock-call-count assertions. The one `vi.spyOn` is the pre-existing `getBoundingClientRect` stub in `openAnnotationEditor`, which is required because jsdom reports a zero-sized surface.
- Practice compliance: No `any`, no suppression comments, no dynamic imports. State stays in ccstate signals with the new `dragPushed$` held in the viewport's `internal` bag; render remains pure, and the new `runChord`/`runBareKey` are event callbacks prepared during render rather than work executed in it (`docs/effect.md`).
- Feature gating: Unchanged — the whole editor stays behind the non-GA `composerImageAnnotation` switch, so no cross-version compatibility work is required (`docs/fallback.md`).

The diff introduces no fallback, so no `Fallbacks` section applies.

PR #32638, opened 2026-09-08
fix: reject implausible voice input output

e7h4n:
## Summary

- reject model-generated segment transcripts that are at least 200 characters and exceed 25 characters per uploaded audio second
- apply the same conservative bound to final polished text using the complete recording duration, without comparing accumulated text to only the final segment
- return an empty completion only when a final recording has no saved or transcribed speech source; keep suspicious in-progress or previously transcribed recordings retryable
- add no client dependency or bundle payload

## Context

A silent or non-speech recording can be sent alongside editor and assistant reference context. If a multimodal model invents context-derived text instead of returning `[NO_SPEECH]`, that text can otherwise be delivered as user input. The minimum-length floor avoids classifying ordinary short responses by their burst character rate, while the duration bound rejects context-sized output that the audio cannot physically contain.

The guard follows the layered approach documented by DoNotType, where prompt-only no-speech handling did not reliably stop context-grounded hallucinations: https://github.com/bojieli/DoNotType/blob/67fb439bed421c112d7172f514c2f12f1cb4d0f4/Sources/DoNotTypeCore/HallucinationGuard.swift

## Compatibility

- the request and response schemas are unchanged
- existing clients continue to receive the endpoint's established `204` no-content and `502` retryable failure outcomes

## Verification

- `pnpm -F api exec vitest run src/signals/routes/__tests__/voice-io-transcribe.test.ts --maxWorkers=1 --silent=passed-only` — 44 passed
- `pnpm -F api lint`
- `pnpm -F api check-types`
- `pnpm knip --workspace api`
- `pnpm exec prettier --check apps/api/src/signals/services/voice-io-transcribe.service.ts apps/api/src/signals/routes/voice-io-transcribe.ts apps/api/src/signals/routes/__tests__/voice-io-transcribe.test.ts`
- `pnpm exec commitlint --from HEAD~1 --to HEAD`

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/voice-io-transcribe.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/voice-io-transcribe.test.ts
@@ -746,6 +746,131 @@ describe("POST /api/voice-io/transcribe/segment", () => {
     },
   );
 
+  it.each([
+    {
+      label: "keeps output below the conservative evidence floor",
+      transcript: "x".repeat(199),
+      polishedText: "x".repeat(199),
+      durationSeconds: 1,
+      status: 200,
+    },
+    {
+      label: "keeps output at the maximum plausible speech rate",
+      transcript: "x".repeat(200),
+      polishedText: "x".repeat(200),
+      durationSeconds: 8,
+      status: 200,
+    },
+    {
+      label: "drops an implausibly long segment transcript",
+      transcript: "x".repeat(200),
+      polishedText: "x".repeat(200),
+      durationSeconds: 1,
+      status: 204,
+    },
+    {
+      label: "rejects implausible polish while preserving a short transcript",
+      transcript: "Recorded speech.",
+      polishedText: "x".repeat(200),
+      durationSeconds: 1,
+      status: 502,
+    },
+  ])(
+    "$label",
+    async ({ transcript, polishedText, durationSeconds, status }) => {
+      mockOptionalEnv("OPENROUTER_API_KEY", "test-openrouter-key");
+      await enabledActor();
+      server.use(
+        http.post(OPENROUTER_URL, () => {
+          return HttpResponse.json({
+            choices: [
+              {
+                finish_reason: "stop",
+                message: {
+                  content: JSON.stringify({
+                    transcript,
+                    polishedText,
+                    language: "en",
+                  }),
+                },
+              },
+            ],
+          });
+        }),
+      );
+      const response = await client().segment({
+        headers: { authorization: "Bearer clerk-session" },
+        body: segmentForm(
+          [audioFile(1, durationSeconds)],
+          "",
+          true,
+          durationSeconds,
+        ),
+      });
+      expect(response.status).toBe(status);
+    },
+  );
+
+  it("does not return context-derived polish without transcribed speech", async () => {
+    mockOptionalEnv("OPENROUTER_API_KEY", "test-openrouter-key");
+    await enabledActor();
+    server.use(
+      http.post(OPENROUTER_URL, () => {
+        return HttpResponse.json({
+          choices: [
+            {
+              finish_reason: "stop",
+              message: {
+                content: JSON.stringify({
+                  transcript: "[NO_SPEECH]",
+                  polishedText: "Use LaunchPad for this release.",
+                  language: "und",
+                }),
+              },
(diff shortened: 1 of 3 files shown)


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: a0adb384ba27b9f57bcfa43ccf820f14c67d520c
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

This change carries the server-measured WAV duration into the Voice Input V2 service and suppresses context-sized model output that exceeds a conservative speech-rate bound. It preserves retryability when recorded or transcribed speech exists, keeps the established API schema/status set, and adds route-level regressions for threshold boundaries, context leakage, and saved-speech protection.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Correctness and edge cases | Pass | `voice-io-transcribe.ts` separates exact uploaded duration from rounded billing duration; `rejectUnusableVoiceOutput` uses the segment duration for segment text and whole-recording duration for final polish, with explicit boundary coverage. |
| `REVIEW.md` | Security and untrusted model output | Pass | The existing authenticated, organization-scoped, feature-gated route is unchanged; context-derived output with no transcribed speech is now suppressed before it can reach the composer. |
| `REVIEW.md` | Type safety and maintainability | Pass | The new duration field is required and typed; the guard is split into focused helpers with named thresholds and no broad casts or oversized functions. |
| `REVIEW.md` | Fix regression coverage | Pass | Route integration tests cover 199/200-character and 25-character-per-second boundaries, implausible transcript and polish paths, `[NO_SPEECH]` context leakage, and prior-transcript preservation. |
| `REVIEW.md` | Fallback discipline | N/A | No compatibility or recovery fallback is introduced; `204` and `502` remain existing contract outcomes for empty completion and retryable provider failure. |
| `REVIEW.md` | DB/JSONB and event-sourcing contracts | N/A | The diff changes no schema, migration, persisted payload, event, or projection. |
| `docs/bad-smell.md` | No `any`, suppressions, dynamic imports, or hardcoded environment configuration | Pass | All three changed files add none of the prohibited constructs; constants are domain-named and local to the service. |
| `docs/bad-smell.md` | Fail visibly instead of hiding recoverable failures | Pass | Suspicious output returns `502` whenever saved speech or a plausible short transcript must remain retryable; only a final no-prefix implausible transcript is treated as empty completion. |
| `docs/testing.md` | Prefer integration tests and mock only system boundaries | Pass | Tests exercise the real Hono route and service via the ts-rest client while MSW intercepts only OpenRouter. |
| `docs/testing.md` | Assert behavior rather than mock calls | Pass | New assertions observe HTTP status and public error code; none use call-count assertions as the outcome. |
| `docs/testing/patterns.md` | Use the standard API route integration structure | Pass | The existing module-scoped `testContext`, `setupApp`, contract client, and route slice are used. |
| `docs/testing/anti-patterns.md` | Avoid internal mocks, direct fetch mocks, fake timers, and service tests | Pass | The added cases introduce none of these anti-patterns and remain at the API boundary. |
| `docs/testing/api-testing.md` | Test route-visible behavior with centralized external mocks | Pass | Requests are built through `voiceIoTranscribeContract`; provider responses use the shared MSW server and the real route stack validates responses. |
| `docs/testing/testing-external-behavior.md` | Construct and verify cases through the external API surface | Pass | The regressions provide multipart requests and inspect endpoint responses without importing services or database schemas. |
| `.claude/skills/ccstate/SKILL.md` | Preserve command composition and AbortSignal ownership | Pass | New helpers do not leak `get`/`set`; the route still checks cancellation after WAV inspection and the service keeps the caller/request composite signal for provider work. |
| `docs/deployment-compatibility.md` | Keep old frontend and new backend interoperable | Pass | Multipart request fields and response schemas are unchanged. Current clients already accept `204`; unaccepted `502` propagates while the persisted voice draft remains available for Retry. |

### Findings

#### Critical (P0)

- None.

#### High Priority (P1)

- None.

### Testing

- Coverage: Adequate — the fix has route-level regression coverage for each new decision branch and Codecov reports all modified coverable lines covered.
- Conventions: Compliant — no internal mocks, direct fetch stubs, fake timers, or implementation-only assertions.
- CI: Required `ci-gate-crates`, `ci-gate-security`, and `ci-gate-turbo` checks passed on the reviewed HEAD.

Issue #32442, opened 2026-09-08
feat: support reusable reference images with organization visibility (EPIC)


bingjiez666:
## Outcome

Let a workspace member upload one image as a reusable **reference image**, choose who in the organization can use it, and select it from the Illustration picker for later image generations.

This revision intentionally removes the image-style package and agent-analysis design. The uploaded image itself is the generation reference. Import is synchronous catalog creation, not a chat run.

## Why the design changed

The current image pipeline already supports image-to-image generation through source image inputs. The two successful product probes in this thread showed that a clean reference image plus a raw prompt is enough to reproduce the visual language for a different subject.

A compiled `SKILL.md`, Template-artifact guide, mounted volume, analysis thread, and publish CLI would add latency and failure modes without being required for the first useful version.

## Product language

- Asset name: **Reference image**
- Picker entry point: **Upload reference image**
- Selected chip: **Reference · {title}**
- Visibility labels:
  - **Only you** — Only you can see and use this reference
  - **Everyone in {organization}** — Any member of this organization can see and use it

Do not call organization visibility “Public” in the UI. The existing internal `public` value may be reused for consistency with presentation templates, but it always means the owning organization only; it never means internet-public.

## Recommended UX

Design anchor: https://a.okou.io/3us1h1pp28.png  
Nine layout explorations: https://a.okou.io/uuh8oa0s12.png

### Browse

The Illustration picker is grouped in this order:

1. **Your references**
   - First card: **Upload reference image**
   - The member’s own private and organization-visible references
2. **Organization**
   - References other members shared with the organization
   - Show creator attribution, for example `Shared by Scarlett`
3. **Built-in styles**
   - Existing registry-backed Illustration styles, unchanged

A private reference carries a lock indicator. An organization-visible reference carries a people indicator. These indicators describe access and are not the primary card action.

### Upload

```text
Illustration picker
  -> Upload reference image
  -> choose or drop one image
  -> review the preview
  -> optionally crop or rotate
  -> edit the filename-derived name
  -> choose Only you (default) or Everyone in {organization}
  -> Add reference
  -> card appears immediately and can be selected
```

The upload reuses the ordinary private upload infrastructure, but it does **not** attach the image to the composer, create a thread, or launch an agent run. The only work after upload is validation and creation of the catalog row.

[...]

PR #32712, opened 2026-09-08
feat(api): add reusable image reference catalog

bingjiez666:
## Stack

Review in order: [#32712](https://github.com/vm0-ai/vm0/pull/32712) → [#32713](https://github.com/vm0-ai/vm0/pull/32713) → [#32714](https://github.com/vm0-ai/vm0/pull/32714) → [#32709](https://github.com/vm0-ai/vm0/pull/32709) → [#32715](https://github.com/vm0-ai/vm0/pull/32715).

- **PR 1 of 5**
- Base: `main`
- Head: `feat/image-reference-catalog`
- Part of #32442

## Summary

- adds the `image_references` catalog backed by one canonical private artifact through `source_file_id`
- adds a domain-specific `POST /api/image-references/uploads/prepare` endpoint; generic uploads no longer branch on an `image-reference` purpose
- validates PNG, JPEG, and WebP bytes, dimensions, pixel bounds, ownership, organization, and ready private-artifact state
- returns short-lived `previewUrl` fields and renews them by `referenceId`, without a synthetic preview-asset prefix or storage-key hash
- implements owner/private and same-organization shared access, creator metadata, owner management, and organization-admin unshare
- publishes user and organization realtime invalidation without exposing storage keys or signed provider URLs
- adds the organization-scoped `ReferenceImages` switch, disabled by default

## Storage and security

The binary is uploaded once to the private artifact bucket. The catalog references `run_uploaded_files` through a unique non-null `ON DELETE RESTRICT` foreign key; organization visibility never copies or publishes the object. Preview URLs are issued only after catalog authorization. Deletion revokes the database records transactionally before best-effort private-object cleanup.

## Verification

- API contracts, Core, and DB type checks: passed
- API contracts, Core, DB, and changed API lint: passed
- full API type-check pipeline: passed on the final stack
- generated migration `1097_fair_malice.sql` from the rebased schema
- route-level `.jpg` + `image/jpeg` coverage added; `image-dimensions` reports both `.jpg` and `.jpeg` bytes as `jpeg`
- migration runtime consistency could not start locally because `DATABASE_URL` is unavailable
- repository-wide tests and migration execution are delegated to the fresh PR CI run

No full local Vitest suite or dev server was run.

## Scope

- backend catalog foundation only
- no generation integration or UI

## Automation

- Draft PR only
- `pr-auto` was not invoked or attached
- no review requested and no merge attempted


turbo/apps/api/src/lib/image-reference-metadata.ts
```diff
@@ -0,0 +1,41 @@
...
+  readonly contentType: ImageReferenceContentType;
+  readonly width: number;
+  readonly height: number;
+}
+
+function contentTypeForImageFormat(
+  format: string,
+): ImageReferenceContentType | null {
+  switch (format) {
+    case "jpeg": {
```
bingjiez666, line 14: add jpg

bingjiez666: .jpg is a filename extension; its standard media type is image/jpeg, and image-dimensions reports decoded .jpg and .jpeg bytes as jpeg. I added a route-level .jpg upload/create test so the extension is covered without adding an unreachable decoder branch.

bingjiez666 (commented):
`Changes Requested`

Reviewed commit: `2fb2a8b63b105d3cdc23751e28a0320fc366331a` (PR is currently **CLOSED**; findings apply to this revision).
Practice documents read from `main`: `REVIEW.md`, `docs/docs.md`, `docs/bad-smell.md`, `docs/fallback.md`, `docs/deployment-compatibility.md`, `docs/testing.md`, `docs/testing/testing-external-behavior.md`, `.claude/skills/database-development/SKILL.md` (+ `references/migrations.md`, `references/query-contracts.md`, `turbo/packages/db/MIGRATIONS.md`), `.claude/skills/feature-switch/SKILL.md`.

### Summary

Adds an `image_references` catalog backed by one canonical private artifact (`source_file_id` → `run_uploaded_files`, unique, `ON DELETE RESTRICT`), a domain-specific `POST /api/image-references/uploads/prepare`, byte-level format/dimension validation, owner-private plus same-org shared reads with short-lived presigned previews, owner management and org-admin unshare, realtime invalidation, and the org-scoped `ReferenceImages` switch (off by default). Backend only — no generation integration or UI.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point; check auth, permissions, ordering, failure outcomes | Finding (P1) | `image-reference-create.service.ts:96` accepts any owned private artifact and `image-reference-delete.service.ts:97` hard-deletes it; see Finding 1. Auth/org scoping itself is correct in `image-reference-data.service.ts:106-124`. |
| `REVIEW.md` | Meaningful coverage for new behavior; feature/capability/permission containment | Pass | `routes/__tests__/image-references.test.ts` covers disabled-by-default 403 on all 7 routes, ownership/org isolation, admin unshare, 409, preview scoping, and strict-body rejection. |
| `REVIEW.md` | Tests verify externally observable state or HTTP responses | Pass | All setup and assertions go through `imageClient()` / `uploadClient()` / `featureClient()`; only S3/Ably/Clerk mocks are inspected. |
| `REVIEW.md` | Schema, API contract, and response validation reviewed together | Pass | `image-reference.ts` checks (`title`, dimensions, visibility) match `contracts/image-references.ts` bounds; `validateSelectedRow` throws instead of defaulting. |
| `REVIEW.md` | New fallbacks declared in a `Fallbacks` PR-body section | N/A | The diff adds no compatibility branch, tolerance branch, or fabricated default; the `head.contentType === undefined` skip is genuinely optional external data (`S3ObjectHead.contentType`), not a rollout fallback. |
| `docs/bad-smell.md` | No `any`; validate at untrusted boundaries | Pass | `commandInput`/`deleteObjectKeys` narrow `unknown`; `parseImageReferenceMetadata` decodes real bytes rather than trusting the declared content type. |
| `docs/bad-smell.md` | No lint/type suppressions | Pass | No `eslint-disable` / `@ts-expect-error` / `prettier-ignore` in the diff. |
| `docs/bad-smell.md` | Error handling only at a meaningful boundary; no fabricated defaults hiding broken invariants; preserve best-effort ownership | Pass | `validateSelectedRow` and the delete-time ownership assertion throw; the S3 object cleanup is explicitly best-effort with `tapError` + `L.warn`. |
| `docs/bad-smell.md` | Document changed public contracts and assess consumers | Finding (P2) | `ORG_SCOPED_FEATURE_SWITCH_KEYS` gains a third key but `.claude/skills/feature-switch/SKILL.md` still documents it as holding only `ChatErrorRecovery` and `PiLoop`; see Observation D. |
| `docs/bad-smell.md` | Static imports; centralized `env()`; no invented URLs/credentials | Pass | Static imports throughout; `privateArtifactsBucket()` and the delete-time bucket check both read `env("R2_PRIVATE_ARTIFACTS_BUCKET_NAME")`. |
| `docs/fallback.md` | No fallback for a state the contract already rules out | Pass | Invalid source metadata raises rather than substituting a placeholder (`image-reference-data.service.ts:59`). |
| `docs/fallback.md` | Non-GA switched feature needs no compatibility code | Pass | `ReferenceImages` is `enabled: false` with no org hashes and the diff carries no dual-read/dual-write or migration for pre-GA rows. |
| `docs/deployment-compatibility.md` | Persisted state and API shape must survive independently deployed old/new components | Pass | New table and new routes only; no existing column, response, or persisted shape changes, so old API and old clients are unaffected. |
| `docs/testing.md` | Add tests for new behavior; mock external services at their boundary; unique identities | Pass | Every test mints `user_${randomUUID()}` / `org_${randomUUID()}`; S3 and Ably are mocked through `context.mocks`, no fake timers. |
| `docs/testing.md` | Cover material edge cases, not library behavior | Finding (P2) | Dimension/pixel-ceiling rejection, cross-org source rejection, own-first list ordering, and the empty-PATCH 400 are unexercised; see Testing. |
| `docs/testing/testing-external-behavior.md` | API tests must not import DB schema or API service files | Pass | `image-references.test.ts` imports only contracts, route modules, and test helpers. |
| `.claude/skills/database-development/SKILL.md` | Use Drizzle to generate migration metadata; never hand-edit journal/snapshot | Finding (P1) | `1097_fair_malice` collides with `1097_puzzling_retro_girl` already on `main` (journal now runs to `1101`); see Finding 2. |
| `.claude/skills/database-development/references/query-contracts.md` | Every selected expression owns a real runtime decoder; builder-first SQL | Pass | Selections spread `getTableColumns(imageReferences)` plus schema columns; `desc(eq(...))` is an ordering expression, and `.for("update", { of: ... })` preserves the lock contract. |
| `turbo/packages/db/MIGRATIONS.md` | Online-safe DDL under the 1s lock / 10s statement defaults | Pass | The table and both indexes are created empty in one transaction; the FK needs no `NOT VALID` validation scan because the child table has no rows. |
| `.claude/skills/feature-switch/SKILL.md` | New user-facing API endpoints gated behind a switch, off by default | Pass | Key added to the enum and registry with `enabled: false`, registered in `ORG_SCOPED_FEATURE_SWITCH_KEYS`, and every one of the 7 routes returns 403 when off (asserted in the test). |

### Findings

#### P1 — 1. Deleting an image reference can destroy an unrelated artifact and its stored object

`image-reference-create.service.ts:96` (`isOwnedReadyPrivateSource`) admits **any** `run_uploaded_files` row that is owned by the caller, in the caller's org, `accessLevel === "private"`, and `materializationStatus === "ready"`. Nothing ties the source to the new `POST /api/image-references/uploads/prepare` endpoint. The PR's own test proves this: with `PrivateArtifacts` enabled, a source produced by the generic `/api/uploads/prepare` (`purpose: "artifact"`) reaches the byte checks and is rejected only for size (`image-references.test.ts`, the `oversized` case).

`image-reference-delete.service.ts:97-125` then treats that row as exclusively owned: inside the transaction it deletes the `image_references` row **and** the `run_uploaded_files` row, then deletes the S3 object.

Trigger: user has `PrivateArtifacts` and `ReferenceImages` both enabled; an image already exists in their artifact catalog (a generation output or a chat upload, file id `F`). They call `POST /api/image-references` with `sourceFileId: F`, then `DELETE /api/image-references/:id`.

Consequence: `image_artifacts.file_id` and `video_artifacts.file_id` reference `run_uploaded_files(id)` **`ON DELETE CASCADE`** (`turbo/packages/db/src/schema/artifact.ts:135-172`), and `image_artifacts_delete_artifact_registry` then removes the `artifacts` registry row. So deleting a *catalog entry* silently removes the artifact from the user's Artifacts list and permanently deletes the underlying object — every `/artifacts/<F>/<filename>` link already sent in a chat thread breaks, with no warning and no undo. The `ON DELETE RESTRICT` FK on `image_references.source_file_id` does not protect against this; it points the other way.

Suggested fix: make the source exclusively owned before the delete path assumes it is — e.g. record an image-reference marker in the private-artifact metadata at `prepareUpload` and require it in `isOwnedReadyPrivateSource`, or keep accepting shared sources but stop deleting the `run_uploaded_files` row and object when the reference did not create them.

#### P1 — 2. Migration `1097_fair_malice` collides with shipped migrations on `main`

`turbo/packages/db/src/migrations/1097_fair_malice.sql` plus its `meta/1097_snapshot.json` and `_journal.json` entry are generated against a base where `1096` was the tip. `main` now contains `1097_puzzling_retro_girl`, `1098_retire_member_invitation_capability`, `1099_chat_reasoning_effort`, `1100_relax_computer_use_host_client_product`, and `1101_zippy_daimon_hellstrom`.

Trigger: rebasing or merging this branch onto current `main`.

Consequence: a duplicate `1097` tag and a snapshot chained to the wrong parent; `pnpm -F @okouai/db test:migration-consistency` fails and the journal cannot be merged as-is. Per `.claude/skills/database-development/references/migrations.md` the fix is to drop the generated files, rebase, and re-run `pnpm db:generate` so Drizzle allocates the next number and rebuilds the snapshot — do not hand-edit `_journal.json` or the snapshot, and do not wait on any other stacked PR to renumber first.

#### P2 — Non-blocking observations

A. **Reference uploads leak into the artifact catalog.** The `run_uploaded_files_queue_artifact_catalog` trigger queues every row whose `url` and `org_id` are non-null, and `completePrivateArtifact$` sets `url`. Because the reference source is completed through the generic `/api/uploads/complete`, each reference upload is also projected into the user's Artifacts list. If the reference catalog is meant to be a separate surface, this needs explicit containment before the switch opens.

B. **`list` is unbounded.** `imageReferencesContract.list` returns every accessible row with a freshly signed preview per row, while `resolvePreviewUrls` caps input at `MAX_IMAGE_REFERENCE_PREVIEW_URLS = 100`. An org that shares many references returns an unbounded body and signs one URL per row. Consider a cap or cursor before rollout. Relatedly, `desc(eq(imageReferences.ownerUserId, args.userId))` as the leading sort key cannot use `idx_image_references_org_public_created`.

C. **Whole-object download to read a header.** `createImageReference$` pulls up to `MAX_IMAGE_REFERENCE_SOURCE_BYTES` (20 MB) into memory via `downloadS3BufferWithMaxBytes` only to feed `imageDimensionsFromData`, which needs a few hundred header bytes. A ranged read would cut per-create memory and latency by orders of magnitude on large uploads.

D. **Stale switch documentation.** `.claude/skills/feature-switch/SKILL.md` states `ORG_SCOPED_FEATURE_SWITCH_KEYS` "currently holds `ChatErrorRecovery` and `PiLoop`". This PR adds a third key; update that sentence in the same change.

E. **Two sources of truth for the preview bucket.** `resolveImageReferencePreviewUrl` signs against `privateArtifactsBucket()`, while `deleteImageReference$` validates the row's recorded `metadata.bucket` against the same env value. The read path would silently sign a key in the wrong bucket if the configured bucket ever changed; reading the recorded bucket in both places would keep one authority.

F. **Suspension checks are asymmetric.** `prepareUpload` calls `rejectSuspendedOrg$`, but `create`, `update`, and `delete` do not, so a suspended org can still mutate the catalog. This matches the `presentation-templates` precedent, so it is noted rather than required.

### Testing

Coverage is strong for a foundation PR and stays on the external boundary throughout: setup and verification run entirely through production endpoints, the only inspected internals are external-service mocks (S3 object presence, Ably `channelGet`/`publish` topics), no DB schema or service module is imported, and every identity is a fresh UUID.

Verified in the diff:

- All seven routes return 403 while `ReferenceImages` is off, and the generic upload path stays unaffected.
- Rejections for pending (not-`ready`) sources, non-private legacy sources, another user's source, content-type spoofing, non-image bytes, and oversized objects.
- `.jpg` + `image/jpeg` accepted, with `image-dimensions` reporting `jpeg` for both extensions; the dedicated `parseImageReferenceMetadata` unit test covers PNG/JPEG/WebP acceptance and GIF/garbage rejection.
- Duplicate source → 409 `CONFLICT`; strict bodies reject unknown fields at the raw route boundary.
- Private/public visibility, cross-org isolation, owner-only rename and delete (404 for members and admins), admin unshare → 204, preview de-duplication, response never serializing a storage key or bucket, and object removal plus a subsequent `uploads/complete` 404 after delete.

Material gaps:

- **No regression coverage for Finding 1.** Add a case that creates a reference from a source allocated through the generic `/api/uploads/prepare`, deletes the reference, and asserts the pre-existing artifact is still retrievable through its production endpoint.
- `invalidDimensions` (per-edge and total-pixel ceilings) is never exercised; a small crafted PNG header with oversized dimensions would cover it without a large fixture.
- Cross-org source rejection at create (`source.orgId === args.orgId`) is untested — only the other-user-same-org case is.
- `PATCH` with an empty body (the `.refine()` on `updateImageReferenceBodySchema`) and the own-first `list` ordering are unverified.
- The PR body records that migration runtime consistency could not run locally without `DATABASE_URL`; given Finding 2, `test:migration-consistency` must pass on a rebased branch before merge.

PR #32735, opened 2026-09-09
fix(platform): drop the dashed outline on a selected freehand stroke

tongx-blip:
## What

Removes the dashed outline drawn around a selected freehand stroke.

The only geometry available to mark out a freehand line is its bounding box, and a rectangle that is not part of the drawing reads as an accidental mark on the user's screenshot rather than as a selection state.

Nothing replaces it. Picking a stroke already opens its note popover, so the selection is not silent, and the ink swatch and Delete act on the same mark either way.

Follow-up to #32624, which introduced the outline.

## Testing

- `A freehand stroke can be selected and moved` no longer asserts on the removed outline. Selection is now proven by the note popover opening, and the move by the rendered `polyline` vertices shifting from `20,25` to `30,35` — the actual drawn geometry rather than a helper element.
- `chat-attachment-annotation`, `chat-attachment-composer`: 18 passed.
- `check-types`, `lint`, `knip`, `format` clean on `@okouai/app`.

Behind the existing `composerImageAnnotation` feature switch.

### DIFF
--- a/turbo/apps/platform/src/views/okou-page/__tests__/chat-attachment-annotation.test.tsx
+++ b/turbo/apps/platform/src/views/okou-page/__tests__/chat-attachment-annotation.test.tsx
@@ -407,6 +407,17 @@ test("An arrow can be selected and re-aimed by dragging its tip", async () => {
   });
 });
 
+/** The first vertex of the rendered freehand stroke, in percent-of-image units. */
+function firstPenPoint(): [number, number] {
+  const points =
+    screen.getByTestId("annotation-mark-1").getAttribute("points") ?? "";
+  const [x, y] = points.split(" ")[0]?.split(",").map(Number) ?? [];
+  if (x === undefined || y === undefined) {
+    throw new Error(`Expected stroke points, got "${points}"`);
+  }
+  return [x, y];
+}
+
 test("A freehand stroke can be selected and moved", async () => {
   const image = draftAttachment("sketch.png", {
     annotatedFileId: "draft-sketch-annotated",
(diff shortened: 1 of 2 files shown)


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: 2db398520349f5f3d4d7e999918aaf9974296f3f
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Removes the dashed bounding box drawn around a selected freehand stroke, which #32624 had introduced as that shape's selection affordance. `SelectionLayer` now returns `null` for `pen`, and `PenSelectionOutline` is deleted. Two files, +32/-40.

### Review Criteria

Applied `docs/bad-smell.md`, `docs/effect.md`, `docs/testing.md`, `docs/testing/app-testing.md`, and `docs/fallback.md`. No DB, contract, event-sourcing, runner, or deployment-boundary surface is touched, so those gates do not apply.

The retained decision comment quotes the requester in Chinese. That is the established convention in this exact file — `AnnotationSelection` and `NoteLayer` already carry the same shape of quoted rationale — so it is treated as consistent rather than as a language-standard finding.

### Findings

#### Critical (P0)

None.

#### High Priority (P1)

None.

#### Nits (P2)

None. The diff is a scoped deletion plus the test update that deletion forces; there is no leftover state, no unused import (`markBounds` is still reached through `rectOf`, and knip is clean), and no stale reference to the removed `annotation-pen-outline` test id anywhere in `turbo/`.

### Testing

- Coverage: Adequate, and correctly *not* expanded. `A freehand stroke can be selected and moved` is the regression path for the behaviour this PR changes, and it still proves both halves: selection now via the note popover opening, and the move via the rendered `polyline` vertices shifting from `20,25` to `30,35`. Asserting the drawn geometry is a stronger check than the removed helper element, since it reads the same coordinates the flattened copy will carry.
- Deliberately absent: no negative test asserting the outline stays deleted. `docs/fallback.md` treats a test that only asserts removed behaviour is still removed as a **P1** finding, so adding one would have been the defect.
- Feature Regression Guard: the affected existing path is stroke select + move, and it retains explicit coverage rather than being deleted alongside the outline.
- Conventions: Compliant. Page-level integration test through `setupPage` with the `ComposerImageAnnotation` switch enabled. No `vi.mock` of internal paths, no `fetch` stubbing, no fake timers, no mock-call-count assertions. `firstPenPoint()` is a file-local reader over rendered DOM, not a unit test of an internal function.
- Practice compliance: No `any`, no suppression comments, no dynamic imports. `SelectionLayer` stays a pure render returning `null`.
- Verified locally on this HEAD: `check-types`, `lint`, `knip`, `format` clean; `chat-attachment-annotation` + `chat-attachment-composer` 18 passed.

The diff introduces no fallback, so no `Fallbacks` section applies.

PR #32766, opened 2026-09-09
feat(platform): frame every avatar to one size in its box

tongx-blip:
## What

Agent cards drop every avatar into the same `h-10 w-10 rounded-full` box, but three rendering paths decide what goes inside it and none of them agrees on framing:

| Source | How it is drawn | Result |
|---|---|---|
| Image URL (brand avatar, uploads) | `` | Fills the box edge to edge |
| Composer avatar | 380px canvas placed raw | ~30% transparent margin above the face |
| Legacy `svg:` avatar | Hard-coded `scale(1.25)` | A third baseline |

Measured across the composer's 12 hair styles x 6 face shapes, the visible artwork covers between **69.5%** and **98.3%** of the box — a **1.42x spread** — and its top edge lands anywhere between 1.7% and 30.5% down. The brand avatar sits at 94.2%. That spread is driven entirely by hair style, so two agents in one row can differ by a third in perceived size with nothing behind it.

`AvatarSvgPreview` already had the code to place the *visible* artwork instead of its transparent canvas, but `centerContent` had exactly one caller — the avatar maker (`avatar-maker.tsx:208`). Agent cards, the sidebar, chat headers, spotlight and mention chips all rendered uncentred.

## How

The placement becomes one rule owned by `avatarSvgComposition`, so every consumer inherits it rather than each caller remembering a prop:

1. **Center** the visible artwork in its box.
2. **Move it halfway to a 0.92 target fill**, in log space: `scale = sqrt(0.92 / fill)`.

Halfway, not all the way. The `avatarNeckSweater` work already pins every composer face to one chin baseline, so the face is consistent and only hair volume varies. Normalizing the silhouette outright would trade an uneven silhouette for an uneven face — a `rounded-crop` avatar would be inflated 1.32x and end up with a visibly larger face than the rest of the cast at 76px. Half the correction keeps every scale within **1.15x** and **0.97x**, and brings the spread from 1.42x down to **1.19x** with the widest result at 90.7% of the box, so nothing is pushed into the circular crop.

The legacy `scale(1.25)` becomes that family's `contentScale` instead of a class each consumer adds by hand — same pixels, one mechanism.

## Rollout

Gated behind the new `avatarFraming` switch (`enabled: false`, staff org hashes), so the framing can be reviewed on real org data before it reaches everyone. While the switch is off, `contentScale` stays at each family's shipped value and only `centerContent` callers get centering, so behaviour is unchanged. The `centerContent` prop goes away with the switch.

## Not in scope

An uploaded image's artwork bounds cannot be measured in CSS, so uploads keep `object-cover`. Normalizing those means trimming transparent padding at upload time, which is a larger change than this issue.

## Verification

- New integration test `agents-page-avatar-framing.test.tsx`: switch off leaves composer avatars on their drawn placement; switch on centers both and pulls a 1.42x pair to under 1.2x; legacy avatars keep `scale(1.25)`.
- `settings-tab`, `chat-composer-mentions`, `user-message-document`, `agents-page`, `sidebar`, `sidebar-virtualization`, `chat-emoji` suites pass.
- Full `@okouai/core` suite passes (240 tests).
- `pnpm format`, `lint` (app + core), `check-types` (app + core), `knip`, `lint:style` all clean.
- The full `@okouai/app` suite exceeds this environment's 10-minute command ceiling and is left to CI.

Design review with the measurements and the rejected alternatives: https://des-45-avatar-framing.okou.app

Refs DES-45


tongx-blip:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 44f9474fbe301fa0ee2d6f5de00a6226c2117746

Reviewed HEAD: `44f9474fbe301fa0ee2d6f5de00a6226c2117746`
Practice revision: `65be00b4ec1fd1e6a51384817bac8a1bca441fa9`

### Summary

Moves avatar placement into one rule owned by `avatarSvgComposition`: the visible artwork is centered in its box and scaled halfway toward a 0.92 target fill, so hair volume stops changing how large an avatar looks. The legacy `scale(1.25)` becomes that family's `contentScale` instead of a class each consumer re-adds, and both consumers (`AvatarSvgPreview` and the ProseMirror mention chip) share `avatarSvgContentTransform`. Gated behind the new staff-only `avatarFraming` switch.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and all consumers | Pass | Both `avatarSvgComposition` callers updated (`avatar-svg-preview.tsx`, `composer-agent-mention-node.ts`); the runtime switch bag is fed from the single sync command in `tiptap-workflow-composer.ts:1876` |
| `REVIEW.md` | Meaningful coverage for new behavior; refactors stay protected | Pass | `agents-page-avatar-framing.test.tsx` covers switch off, switch on, and the legacy `scale(1.25)` behavior-preservation claim |
| `REVIEW.md` | Tests verify externally observable state, not internal detail | Finding (P1) | `agents-page-avatar-framing.test.tsx:73-80` resolves the artwork element by DOM position and asserts its inline `style.transform` |
| `REVIEW.md` | Feature/rollout containment for a changed existing flow | Pass | Every new placement path is behind `FeatureSwitchKey.AvatarFraming`; with the switch off `contentScale` stays at each family's shipped value and only `centerContent` callers center |
| `REVIEW.md` | Persisted data, contracts, and response validation | N/A | No schema, contract, or persisted shape changes; the enum addition is additive and unknown keys are dropped by `filterFeatureSwitchOverrides` |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | New code is fully typed; `pnpm -F @okouai/app lint` and `check-types` clean |
| `docs/bad-smell.md` | Catch only at a boundary that handles the error | N/A | No error handling introduced |
| `docs/bad-smell.md` | Document changed public contracts and assess consumers | Pass | `AgentMentionAvatarRuntime` swaps `setNeckSweaterEnabled`/`neckSweaterEnabled` for `setSwitches`/`switches`; the single caller is updated in the same change |
| `docs/bad-smell.md` | Static imports in production code | Pass | No dynamic imports added |
| `docs/bad-smell.md` | No fallback for a state the contract rules out | Pass | The off-branch is the switch's own disabled path, not a fallback for an unreachable state |
| `docs/fallback.md` | A non-GA switch needs no compatibility branch (§2) | Pass | No dual read/write or tolerance branch for the feature's own earlier shape |
| `docs/fallback.md` | New fallbacks declared in a `Fallbacks` PR-body section | N/A | Section 7 gates apply to GA behavior only; a staff-only switch creates no obligation, so no declaration is required |
| `docs/testing.md` | Platform tests boot the real Router and assert page-observable state | Finding (P1) | Construction is correct (`setupPage`, real agents list, real switch override); the assertions leave the boundary |
| `docs/testing.md` | Mock only external services; use `testContext()` lifetimes | Pass | Uses `context.mocks.data.agents` / `onboardingStatus`; no `vi.mock`, no global MSW server |
| `docs/testing/app-testing.md` | Prefer presence, visibility, content, value, state, focus, navigation | Finding (P1) | The test asserts a computed `transform` string, which is not in the sanctioned assertion set |
| `docs/testing/app-testing.md` | Never assert DOM identity | Finding (P1) | `box?.firstElementChild` pins the wrapper's position in the tree |
| `docs/testing/testing-external-behavior.md` | Do not assert on CSS classes or styling; state the rationale when an exception is genuinely needed | Finding (P1) | The exception is defensible but undeclared — see Findings |
| `docs/styles.md` | Tailwind utilities are the only component-facing styling API | Pass | No new selector, CSS module, `<style>` element, or injected stylesheet; `pnpm lint:style` clean and `styleInjections` in the legacy baseline only tracks `jsx-style` elements |
| `docs/styles.md` | A genuinely computed runtime value may be carried as component data | Pass | `transform` is computed per config; the pre-existing code already set `style.transform` for `headScale` and `contentOffsetY` |
| `docs/styles.md` | Runtime behavior and tests use roles, names, refs, `data-*` hooks or documented slots | Finding (P1) | The test has no stable hook to target and walks the tree instead |
| `.claude/skills/feature-switch/SKILL.md` | Key in the enum, entry in the registry, gated in application code | Pass | `FeatureSwitchKey.AvatarFraming`, registry entry with maintainer + `STAFF_ORG_ID_HASHES`, and a client `avatarFramingEnabled$` computed |
| `.claude/skills/feature-switch/SKILL.md` | New user-facing behavior is gated for gradual rollout | Pass | `enabled: false` + staff org hashes, matching the `avatarNeckSweater` precedent for avatar geometry |
| `.claude/skills/ccstate/SKILL.md` | Derivation in `computed`; no signal creation or store mutation during render | Pass | `avatarFramingEnabled$` is a plain `computed` over `featureSwitch$`, read with `useGet` |
| `.claude/skills/ccstate/SKILL.md` | Async work has an owner and cancellation path | N/A | No async work added |

### Findings

**P1 — The framing test asserts styling and DOM identity without declaring the exception**
`turbo/apps/platform/src/views/agents-page/__tests__/agents-page-avatar-framing.test.tsx:73-88, 119-120, 137-145, 159`

`avatarFramingTransform` locates the artwork by taking `firstElementChild` of the `[role="img"]` box and returns its inline `style.transform`; the tests then match `translateY(-15.26`, `scale(1.25)`, and `toBeCloseTo(1.151, 3)`.

`docs/testing/app-testing.md` says never to assert DOM identity and lists the sanctioned assertion set, and `docs/testing/testing-external-behavior.md` says not to assert on CSS classes or styling because it freezes the implementation. Trigger: any refactor that produces identical pixels through a different mechanism — moving the scale onto the box, splitting into the `scale`/`translate` longhand properties, or inserting a wrapper — fails these tests while the product is unchanged. Consequence: the tests block the next correct change rather than protecting a user-visible behavior.

The exception itself is legitimate: jsdom performs no layout, so rendered avatar size has no page-observable surface, and `testing-external-behavior.md` allows leaving the boundary for a case that cannot be constructed or observed through the production interface. What is missing is the treatment that guide requires — the test must state why the external interface cannot observe the state and why the case is still worth testing.

Requested changes:
1. State the exception at the top of the test file: framing has no page-observable surface under jsdom, so the applied transform is the only available evidence.
2. Replace the `firstElementChild` walk with a stable hook. `docs/styles.md` names `data-*` hooks and documented component slots as the supported mechanism; add one to the artwork wrapper in `AvatarSvgPreview` and the mention-chip node view so both consumers expose the same slot.
3. Drop `toBeCloseTo(1.151, 3)` / `toBeCloseTo(0.967, 3)` and the `translateY(-15.26` / `translateY(-0.83` string matches. Keep the assertions that are the actual guarantee: the flat artwork scales up, the tall one scales down, and the drawn 1.42x spread becomes under 1.2x. Those hold under any equivalent implementation of the rule.

### Non-blocking observations

- `composer-agent-mention-node.ts:202` compares `switches !== currentSwitches` by identity. That is correct only because `setSwitches` returns early when both fields match and therefore keeps the previous object reference. The invariant is real but implicit; a short comment on `setSwitches` noting that callers rely on reference stability would keep a future edit from breaking chip re-rendering.
- Scaling the artwork inside an `overflow-hidden` box can clip at scales above 1. The widest framed result measured is 90.7% of the box, so nothing is currently cut, but that headroom is a property of the current assets rather than something the rule enforces.
- `avatar-maker.tsx` keeps passing `centerContent`. With the switch on it is redundant, which is intended and already noted in the code comment; it should be removed together with the switch.

### Testing

Coverage is present and targeted at the new behavior: switch off leaves composer avatars on their drawn placement, switch on centers both and pulls a 1.42x pair under 1.2x, and legacy avatars keep `scale(1.25)`, which is the behavior-preservation claim for folding the hard-coded class into `contentScale`. Construction goes through the production path — `setupPage`, the real agents list, and a real feature-switch override — with no `vi.mock` and no direct store mutation, matching `docs/testing/app-testing.md`.

The gap is the assertion boundary described in Findings, not the scenarios. The mention-chip consumer changed shape (`setSwitches`/`switches`) and its framing branch is exercised only indirectly through the existing `chat-composer-mentions` suite; that is acceptable here because the chip renders through the same shared `avatarSvgContentTransform`, but a chip-level case would be worth adding if the two consumers ever diverge.

Verification reported in the PR body — targeted suites, full `@okouai/core`, format, lint, types, knip, and `lint:style` — matches what the changed surface requires. The full `@okouai/app` suite is left to CI, which is the right boundary.

tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: efe95add631d30ed8b9af4c0e3771ee47c8aaa4f
PR_AUTO_REVIEW_VERDICT: LGTM

Reviewed HEAD: `efe95add631d30ed8b9af4c0e3771ee47c8aaa4f`
Practice revision: `65be00b4ec1fd1e6a51384817bac8a1bca441fa9`

### Summary

Moves avatar placement into one rule owned by `avatarSvgComposition`: the visible artwork is centered in its box and scaled halfway toward a 0.92 target fill, so hair volume stops changing how large an avatar looks. The legacy `scale(1.25)` becomes that family's `contentScale` instead of a class each consumer re-adds, and both consumers share `avatarSvgContentTransform` and the `AVATAR_ARTWORK_SLOT` hook. Gated behind the staff-only `avatarFraming` switch.

This revision resolves the P1 from the review of `44f9474f`.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and all consumers | Pass | Both `avatarSvgComposition` callers updated; the runtime switch bag is fed from the single sync command in `tiptap-workflow-composer.ts` |
| `REVIEW.md` | Meaningful coverage for new behavior; refactors stay protected | Pass | `agents-page-avatar-framing.test.tsx` covers switch off, switch on, and the legacy `scale(1.25)` behavior-preservation claim |
| `REVIEW.md` | Tests verify externally observable state, not internal detail | Pass | The boundary exception is now declared with its reason, reads through a documented slot, and asserts the guaranteed relationship instead of the formula |
| `REVIEW.md` | Feature/rollout containment for a changed existing flow | Pass | Every new placement path is behind `FeatureSwitchKey.AvatarFraming`; with the switch off `contentScale` stays at each family's shipped value |
| `REVIEW.md` | Persisted data, contracts, and response validation | N/A | No schema, contract, or persisted shape changes; the enum addition is additive and unknown keys are dropped by `filterFeatureSwitchOverrides` |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | Fully typed; `lint` and `check-types` clean for `@okouai/app` and `@okouai/core` |
| `docs/bad-smell.md` | Catch only at a boundary that handles the error | N/A | No error handling introduced |
| `docs/bad-smell.md` | Document changed public contracts and assess consumers | Pass | `AgentMentionAvatarRuntime` swaps to `setSwitches`/`switches`, its reference-stability contract is now documented on the interface, and the single caller is updated in the same change |
| `docs/bad-smell.md` | Static imports in production code | Pass | No dynamic imports added |
| `docs/bad-smell.md` | No fallback for a state the contract rules out | Pass | The off-branch is the switch's own disabled path |
| `docs/fallback.md` | A non-GA switch needs no compatibility branch (§2) | Pass | No dual read/write or tolerance branch for the feature's own earlier shape |
| `docs/fallback.md` | New fallbacks declared in a `Fallbacks` PR-body section | N/A | Section 7 gates apply to GA behavior only; a staff-only switch creates no obligation |
| `docs/testing.md` | Platform tests boot the real Router and assert page-observable state | Pass | `setupPage` with the real Agents route, real agent list, and a real switch override; the unobservable-result exception is declared per the external-behavior guide |
| `docs/testing.md` | Mock only external services; use `testContext()` lifetimes | Pass | `context.mocks.data.*` only; no `vi.mock`, no global MSW server |
| `docs/testing/app-testing.md` | Prefer presence, visibility, content, value, state, focus, navigation | Pass | Within the declared exception, assertions are now relational — direction of scale, direction of offset, and the resulting spread |
| `docs/testing/app-testing.md` | Never assert DOM identity | Pass | `firstElementChild` replaced by `[data-avatar-artwork]` |
| `docs/testing/testing-external-behavior.md` | State why the external interface cannot construct or observe the case | Pass | File-level comment names jsdom's lack of layout and why the case still matters |
| `docs/styles.md` | Tailwind utilities are the only component-facing styling API | Pass | No new selector, CSS module, `<style>` element, or injected stylesheet; `pnpm lint:style` clean and the baseline's `styleInjections` only tracks `jsx-style` elements |
| `docs/styles.md` | A genuinely computed runtime value may be carried as component data | Pass | `transform` is computed per config; the pre-existing code already set `style.transform` for `headScale` |
| `docs/styles.md` | Runtime behavior and tests use roles, names, refs, `data-*` hooks or documented slots | Pass | `AVATAR_ARTWORK_SLOT` is a single shared hook applied by both consumers, so they cannot drift |
| `.claude/skills/feature-switch/SKILL.md` | Key in the enum, entry in the registry, gated in application code | Pass | `FeatureSwitchKey.AvatarFraming`, registry entry with maintainer + `STAFF_ORG_ID_HASHES`, client `avatarFramingEnabled$` |
| `.claude/skills/feature-switch/SKILL.md` | New user-facing behavior is gated for gradual rollout | Pass | `enabled: false` + staff org hashes, matching the `avatarNeckSweater` precedent for avatar geometry |
| `.claude/skills/ccstate/SKILL.md` | Derivation in `computed`; no signal creation or store mutation during render | Pass | `avatarFramingEnabled$` is a plain `computed` over `featureSwitch$`, read with `useGet` |
| `.claude/skills/ccstate/SKILL.md` | Async work has an owner and cancellation path | N/A | No async work added |

### Findings

None.

### Non-blocking observations

- Scaling the artwork inside an `overflow-hidden` box can clip above 1.0. The widest framed result measured is 90.7% of the box, so nothing is currently cut, but that headroom is a property of the current assets rather than something the rule enforces. Worth a glance if a future hair asset is much wider.
- `avatar-maker.tsx` keeps passing `centerContent`. With the switch on it is redundant, which is intended and noted in the code comment; it should be removed together with the switch.

### Testing

Coverage is targeted at the new behavior and each case has a distinct reason to exist: switch off leaves composer avatars on their drawn placement, switch on centers both and pulls a pair that differs by over 1.4x to under 1.2x, and legacy avatars keep `scale(1.25)` — the behavior-preservation claim for folding the hard-coded class into `contentScale`. Construction goes through the production path with no `vi.mock` and no direct store mutation.

The boundary exception is now handled the way `docs/testing/testing-external-behavior.md` requires: the file states that framing produces no page-observable result under jsdom, reads the artwork through a declared slot rather than the DOM shape around it, and asserts the relationship the rule guarantees rather than the exact scale values. A refactor that produces the same framing through a different transform mechanism no longer fails these tests.

The mention-chip consumer changed shape and its framing branch is exercised indirectly through `chat-composer-mentions`; acceptable because it renders through the same shared `avatarSvgContentTransform` and now the same slot.

Verification: `agents-page-avatar-framing`, `chat-composer-mentions`, `settings-tab`, `agents-page`, `sidebar`, `sidebar-virtualization`, `chat-emoji`, full `@okouai/core` (240 tests), plus `format`, `lint`, `check-types`, `knip`, and `lint:style`. The full `@okouai/app` suite is left to CI.

PR #32794, opened 2026-09-09
fix(app): connector directory chip metrics and footer

tongx-blip:
## Why

The category row put `font-medium` in the **selected branch alone**, so the chip you just picked grew and pushed every chip after it along. `transition-colors` does not animate width, so the shift landed in one frame — the jump Tong reported.

Measured in Geist 12px, that weight step is not free:

| Label | 400 | 500 | Δ |
| --- | --- | --- | --- |
| General models | 85.55 | 87.63 | +2.08 |
| Data and automation | 113.98 | 116.61 | +2.63 |

## What

Weight moves onto the base class; the fill and the ink carry selection. This is what `SegmentControl` already does — `segmentControlItemVariants` sets `font-medium` on every item and only swaps background and text colour on `data-checked` — so the chip row and the tabs above it now follow the same rule.

A regression test asserts every category chip carries the same font-weight utility. Reverting the fix makes it fail (`expected 2 to be 1`), so the guard is real.

## Scope note

Not in this PR, after a survey of the other 20 asymmetric-weight ternaries in the app: the pattern is only a defect where items are auto-width **and** laid out horizontally. Sidebar rows, `w-full` tab lists and the composer video options all change weight on selection too, and none of them shift anything. The three other `okou-chip` pill rows already put the weight on the base class. A repo-wide lint rule would therefore be ~19 false positives to 1 real hit.

## Also in this PR

The keyboard hint bar at the foot of the sheet is removed. It spent a whole row on a legend for shortcuts the list answers the moment you press a key, plus a sentence about how search matched. Arrow keys and Enter still work; only the legend is gone, and its three strings are pruned from all ten locales.

Design study: https://okou-connector-toolbar.okou.app


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: 96fcf50b1a0489497d698d27781ee63a70b9481e
PR_AUTO_REVIEW_VERDICT: LGTM

Practice revision: `add5dd89fdeeb5696794fabb362c0266f9b3c086` (docs/styles.md, docs/testing.md, docs/fallback.md, .claude/skills/feature-switch)

## Summary

Two presentation-only changes inside the `connectorDirectory` feature switch, plus a test-selector hardening.

1. The category chip row put `font-medium` in the selected branch alone. Geist is not metric-compatible across 400 → 500, so the selected chip widened itself and pushed every chip after it; `transition-colors` does not animate width, so the shift landed in one frame. Weight moves to the base class and selection is carried by fill and ink.
2. The keyboard hint bar at the foot of the sheet is removed, together with its three strings across all ten locales.
3. The width guard now selects chips through `data-connector-category-chip` instead of a fragment of the Tailwind class string.

## Review criteria

**Styles (docs/styles.md).** No first-party selector, CSS module, runtime stylesheet or CSS-in-JS is introduced; both surfaces compose Tailwind utilities. No new use of a legacy class — the removal deletes four `DIRECTORY_HAIRLINE` compositions and adds none. `node scripts/style-policy.mjs` passes, so the shrink-only baseline is not grown.

**Consistency with the shared primitive.** `segmentControlItemVariants` in `@okouai/ui` already sets `font-medium` on every item and swaps only `bg`/`text` on `data-checked`. The chips now follow the same contract, so the tabs and the chip row beneath them behave identically.

**Fallbacks (docs/fallback.md).** No fallback path, default or rollout switch changes. Everything here renders only when `connectorDirectory` is on, which is `enabled: false` plus staff org hashes.

**i18n.** `keyMove`, `keyOpen` and `matchedOn` are pruned by `pnpm i18n:extract` across all ten locales rather than hand-edited; `pnpm lint:i18n` reports every locale at 100% and no file out of date. No remaining reference to the removed keys, the removed component or `KEYCAP` exists in `turbo/apps/platform/src`.

**Behavior preserved.** Arrow-key navigation and Enter are untouched — only the legend is gone. `DirectoryBrowseView` still returns toolbar, chips and body; the scroll region's bottom fade now dissolves into the sheet edge, which is what `SCROLL_EDGE_FADE` documents.

## Findings

One finding was raised and fixed in this run, in commit `96fcf50`:

- **The width guard selected chips by a Tailwind class fragment.** `element.className.includes("shrink-0 items-center gap-1.5")` breaks the moment the class list is reordered, and the repository has been bitten by class-based test selectors before. Replaced with a `data-connector-category-chip` hook, plus `expect(chips.length).toBeGreaterThan(1)` so the assertion cannot pass against an empty selection.

No further findings.

## Testing

`Keep the category chips the same width when the selection moves` asserts every chip in the row carries the same `font-*` utility. The guard was verified by re-introducing the defect once: it fails with `AssertionError: expected 2 to be 1`, both before and after the selector change.

Run at this HEAD:

- `tsc -p tsconfig.production.json` and `tsconfig.tests.json`
- `oxlint --deny-warnings`, `eslint` on both changed files
- `node scripts/style-policy.mjs`
- `pnpm lint:i18n`
- `vitest run src/views/okou-page/__tests__/connector-directory.test.tsx` — 5 passed

PR #32825, opened 2026-09-09
fix(platform): move avatar heads to the chin baseline instead of resizing them

tongx-blip:
## The bug

The neck and the sweater are one fixed pair of shapes shared by every avatar. The head is not.

To bring each chin down to that shared collar, `avatarSvgComposition` scaled the head layers about the top of the face — and scaling changes the head's **width** as well as its height:

| Face | `headScale` | Rendered head width |
| --- | --- | --- |
| `tall` | 0.893 | 63.7% of the box |
| `square` / `oval` | 0.971 | 69.2% |
| `round-angled-ears` | 1.048 | 74.7% |
| `round` | 1.069 | 76.2% |
| `wide` | 1.108 | 79.0% |

A **1.24× spread**, while the neck (49.9px) and the collar (94.9px) on the same 380px canvas never change size. So the body reads as too big for the head on a `tall` face and too small on a `wide` one — which is what shows up as "some necks look big, some look small".

## The fix

The face assets are already normalized on width — every one of them spans the same 270px, which is what the composer's own comment means by "normalized on width rather than height". Scaling was the wrong operator for a goal that is only ever *put this chin on that line*.

Translating reaches the same line, leaves all six faces at one width (71.3% of the box), and keeps the neck and collar in proportion on every one of them.

`headScale` is replaced by `headOffsetY`, and `AVATAR_HEAD_TRANSFORM_ORIGIN` goes with it — a translation needs no pivot.

## Consequence handled

Artwork bounds are no longer clamped to the canvas. A face whose chin sits below the baseline now moves **up**, and tall hair can reach past the top of its own canvas — `high-bun` on `square` overhangs by 3.4px, `topknot-locks` on `tall` by 8.4px. The DES-45 framing rule has to see that overhang to fit and center the artwork; with the clamp in place the box would simply crop the bun.

This slightly widens the input spread the framing rule sees for the extreme pair (1.42× → 1.53×), and since that rule corrects half the difference in log space the framed spread moves with it (1.19× → 1.24×). That is the rule behaving as specified, not a regression.

## Rollout

**Not gated.** This fixes `avatarNeckSweater`, which is already released, and the repository's feature-switch guide does not require a switch for a bug fix to an existing feature. Carrying both placements would mean two head transforms and two sets of artwork bounds for a state nobody wants to keep. Say the word if you would rather it ride behind `avatarFraming` until that switch flips.

## Verification

- New case in `agents-page-avatar-framing.test.tsx`: neither head carries a scale, both carry a translation, and two different chins move by different amounts to reach the same collar.
- The existing framing case is updated for the new artwork bounds and now asserts the rule's actual invariant — the framed spread is the square root of the drawn spread — instead of a fixed threshold that moved with the input.
- `agents-page`, `settings-tab` (avatar maker), `chat-composer-mentions`, `user-message-document` suites pass.
- `format`, `lint`, `check-types`, `knip`, `lint:style` clean.
- Full `@okouai/app` suite left to CI.

Side-by-side of all six faces, bare and with hair, at card size and 4×: https://avatar-neck-proportion.okou.app

### DIFF
--- a/turbo/apps/platform/src/signals/okou-page/composer-agent-mention-node.ts
+++ b/turbo/apps/platform/src/signals/okou-page/composer-agent-mention-node.ts
@@ -8,7 +8,6 @@ import {
 } from "../../views/okou-page/avatar-utils.ts";
 import {
   AVATAR_ARTWORK_SLOT,
-  AVATAR_HEAD_TRANSFORM_ORIGIN,
   avatarSvgComposition,
   avatarSvgContentTransform,
 } from "../../views/okou-page/avatar-svg-utils.ts";
@@ -122,7 +121,7 @@ function renderAgentMentionAvatar(
   container.replaceChildren();
   const svgConfig = resolveAvatarSvgConfig(avatarUrl);
   if (svgConfig) {
-    const { behind, head, front, headScale, contentOffsetY, contentScale } =
+    const { behind, head, front, headOffsetY, contentOffsetY, contentScale } =
       avatarSvgComposition(svgConfig, switches);
     const layers = document.createElement("span");
     layers.className = "absolute inset-0";
@@ -147,8 +146,7 @@ function renderAgentMentionAvatar(
     };
     const headLayers = document.createElement("span");
     headLayers.className = "absolute inset-0";
-    headLayers.style.transform = `scale(${headScale})`;
-    headLayers.style.transformOrigin = AVATAR_HEAD_TRANSFORM_ORIGIN;
+    headLayers.style.transform = `translateY(${headOffsetY}%)`;
     appendLayers(layers, behind);
     appendLayers(headLayers, head);
     layers.append(headLayers);
--- a/turbo/apps/platform/src/signals/okou-page/composer-agent-mention-node.ts
+++ b/turbo/apps/platform/src/signals/okou-page/composer-agent-mention-node.ts
@@ -8,6 +8,7 @@ import {
 } from "../../views/okou-page/avatar-utils.ts";
 import {
   AVATAR_ARTWORK_SLOT,
+  AVATAR_HEAD_SLOT,
   avatarSvgComposition,
   avatarSvgContentTransform,
 } from "../../views/okou-page/avatar-svg-utils.ts";
@@ -123,11 +124,17 @@ function renderAgentMentionAvatar(
   if (svgConfig) {
     const { behind, head, front, headOffsetY, contentOffsetY, contentScale } =
       avatarSvgComposition(svgConfig, switches);
+    const applySlot = (
+      element: HTMLElement,
+      slot: Readonly<Record<string, string>>,
+    ) => {
+      for (const [name, value] of Object.entries(slot)) {
+        element.setAttribute(name, value);
+      }
+    };
     const layers = document.createElement("span");
     layers.className = "absolute inset-0";
-    for (const [name, value] of Object.entries(AVATAR_ARTWORK_SLOT)) {
-      layers.setAttribute(name, value);
-    }
+    applySlot(layers, AVATAR_ARTWORK_SLOT);
     const transform = avatarSvgContentTransform({
       contentOffsetY: switches.framing ? contentOffsetY : 0,
       contentScale,
@@ -146,6 +153,7 @@ function renderAgentMentionAvatar(
     };
     const headLayers = document.createElement("span");
     headLayers.className = "absolute inset-0";
+    applySlot(headLayers, AVATAR_HEAD_SLOT);
     headLayers.style.transform = `translateY(${headOffsetY}%)`;
     appendLayers(layers, behind);
     appendLayers(headLayers, head);
--- a/turbo/apps/platform/src/views/agents-page/__tests__/agents-page-avatar-framing.test.tsx
+++ b/turbo/apps/platform/src/views/agents-page/__tests__/agents-page-avatar-framing.test.tsx
@@ -34,13 +34,15 @@ const TALL_HAIR_AGENT_ID = "c0000000-0000-4000-a000-000000000022";
 const LEGACY_AGENT_ID = "c0000000-0000-4000-a000-000000000023";
 
 /**
- * The share of the avatar box each artwork covers as drawn, before any framing.
- * These are properties of the composer assets — the flattest hair leaves the
- * canvas's whole forehead margin empty, the tallest fills it — and they are what
- * makes one agent card look larger than the next.
+ * The share of the avatar box each artwork covers as drawn, before any framing:
+ * from the top of the hair, once the head has been moved onto the chin
+ * baseline, down to the bottom of the sweater. These are properties of the
+ * composer assets — the flattest hair leaves the canvas's whole forehead margin
+ * empty, the tallest reaches past the top of it — and they are what makes one
+ * agent card look larger than the next.
  */
-const FLAT_HAIR_SHIPPED_FILL = 264 / 380;
-const TALL_HAIR_SHIPPED_FILL = 373.6826 / 380;
+const FLAT_HAIR_DRAWN_FILL = 250.4 / 380;
+const TALL_HAIR_DRAWN_FILL = 383.4 / 380;
 
 function agent(agentId: string, avatarUrl: string): AgentResponse {
   return {
@@ -157,12 +159,11 @@ test("Center each composer avatar and pull the cast toward one size", async () =
   const flat = avatarFramingTransform(FLAT_HAIR_AGENT_ID);
   const tall = avatarFramingTransform(TALL_HAIR_AGENT_ID);
 
-  // Both move up, because both artworks hang below the middle of their canvas.
-  // The flat-haired one leaves the whole forehead margin empty, so it has much
-  // further to travel than the one whose bun already fills that margin.
-  expect(centeringOffset(flat)).toBeLessThan(0);
-  expect(centeringOffset(tall)).toBeLessThan(0);
-  expect(centeringOffset(flat)).toBeLessThan(centeringOffset(tall));
+  // The flat-haired artwork leaves the whole forehead margin of its canvas
+  // empty, so centering moves it a long way up. The one whose bun already
+  // reaches past the top of the canvas is close to centered where it is.
+  expect(centeringOffset(flat)).toBeLessThan(-10);
+  expect(Math.abs(centeringOffset(tall))).toBeLessThan(1);
 
   // The smaller artwork grows and the larger one shrinks, and neither moves far
   // enough to make its face the odd one out.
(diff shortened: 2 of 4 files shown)


tongx-blip:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 4247edae90eb80b66ef043495e234ef9c65aa153

Reviewed HEAD: `4247edae90eb80b66ef043495e234ef9c65aa153`
Practice revision: `075ccbaaf8f90231687267218ef1e77d05771d3a`

### Summary

Replaces the head's chin-baseline `scale` with a `translate`. Scaling reached the shared collar but also changed the head's width — 0.89x on `tall` to 1.11x on `wide` — while the 49.9px neck and 94.9px collar under it are fixed, so the body read as the wrong size for the head. The face assets are already width-normalized, so a translation reaches the same line and leaves all six faces at one width. `headScale` becomes `headOffsetY`, `AVATAR_HEAD_TRANSFORM_ORIGIN` is deleted, and the artwork bounds stop clamping to the canvas so the framing rule can see hair that now overhangs the top.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and all consumers | Pass | Both `avatarSvgComposition` consumers updated in step: `avatar-svg-preview.tsx:66` and `composer-agent-mention-node.ts:149` |
| `REVIEW.md` | Meaningful coverage for new behavior; refactors stay protected | Pass | New case asserts no head carries a scale, both carry a translation, and two different chins move by different amounts; the existing framing case is updated for the new bounds |
| `REVIEW.md` | Tests verify externally observable state, not internal detail | Pass | Boundary exception already declared at the top of the file and extended to the new helper; construction is still `setupPage` + real agents + real switch |
| `REVIEW.md` | Report concrete practice violations with rule and source location | Finding (P1) | `avatar-svg-utils.ts:162` adds an offset term to a bound that can never be reached with a non-zero offset |
| `REVIEW.md` | Persisted data, contracts, response validation | N/A | No schema, contract, or persisted shape changes; `headScale`/`headOffsetY` is an internal interface with both consumers in this diff |
| `REVIEW.md` | Feature/capability containment for a changed existing flow | Pass | This is a fix to `avatarNeckSweater`, already released; the feature-switch guide does not require a switch for a bug fix to an existing feature |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | New code is fully typed; no suppression comments |
| `docs/bad-smell.md` | Catch only at a boundary that handles the error | N/A | No error handling introduced |
| `docs/bad-smell.md` | Document changed public contracts and assess consumers | Pass | `AvatarSvgComposition.headScale` → `headOffsetY` with an updated doc comment; the removed `AVATAR_HEAD_TRANSFORM_ORIGIN` export had no other consumer |
| `docs/bad-smell.md` | Static imports in production code | Pass | No dynamic imports |
| `docs/bad-smell.md` | Smallest necessary abstraction; no code path nobody executes | Finding (P1) | Same as above — see Findings |
| `docs/fallback.md` | Do not fall back on states the contract cannot produce (§3) | Finding (P1) | The `bottom` offset term guards a `!neckSweater` + moved-head combination the two call sites cannot produce |
| `docs/fallback.md` | Removing a fallback needs evidence, and removal needs no declaration | Pass | The `Math.max(0, …)` clamp is removed with its reason stated in code and PR body: a head that moves up makes the overhang real, and the framing rule must measure it |
| `docs/fallback.md` | New fallbacks declared in a `Fallbacks` PR-body section | Pass | No new rollout fallback; the PR states its ungated decision explicitly |
| `docs/testing.md` | Platform tests boot the real Router and assert page-observable state | Pass | `setupPage` on the real Agents route with a real feature-switch override |
| `docs/testing.md` | Mock only external services; use `testContext()` lifetimes | Pass | `context.mocks.data.*` only |
| `docs/testing/app-testing.md` | Never assert DOM identity | Finding (P1) | `headPlacementTransform` reaches the head group by `:scope > div, :scope > span` under the artwork slot |
| `docs/testing/app-testing.md` | Prefer the sanctioned assertion set | Pass | Within the declared exception, the new assertions are shape-level (`not.toMatch(/scale\(/)`, translate present, two values differ) rather than pinned numbers |
| `docs/testing/testing-external-behavior.md` | State why the external interface cannot observe the case | Pass | The new helper carries its own note and points at the file-level exception |
| `docs/styles.md` | Tailwind utilities are the only component-facing styling API | Pass | No new selector, CSS module, `<style>` element, or injection; the inline transform replaces an existing inline transform |
| `docs/styles.md` | Runtime behavior and tests use roles, names, refs, `data-*` hooks or slots | Finding (P1) | The new test helper starts from `[data-avatar-artwork]` but then walks the tree — see Findings |
| `.claude/skills/feature-switch/SKILL.md` | A switch is not required for a bug fix to an existing feature | Pass | `avatarNeckSweater` is `enabled: true`; this restores its intended geometry rather than adding behavior |
| `.claude/skills/ccstate/SKILL.md` | Derivation in `computed`; no store mutation during render | Pass | No signal changes; the component still reads `avatarFramingEnabled$` / `avatarNeckSweaterEnabled$` with `useGet` |
| `.claude/skills/ccstate/SKILL.md` | Async work has an owner and cancellation path | N/A | No async work |

### Findings

**P1 — Dead offset term in the artwork's bottom bound**
`turbo/apps/platform/src/views/okou-page/avatar-svg-utils.ts:158-163`

```ts
bottom: neckSweater
  ? 380
  : Math.min(380, Math.max(AVATAR_FACE_CHIN_Y[config.face], hairBottom) + headOffsetY),
```

`composerContentBounds` has exactly two call sites: `(config, 0, false)` and `(config, headOffsetY, true)`. The `+ headOffsetY` term only executes in the `!neckSweater` branch, and that branch is only ever reached with `headOffsetY === 0`. It can never change a result.

`docs/fallback.md` §3 rules out a branch for a state the owning contract cannot produce, and `docs/bad-smell.md` asks for the smallest necessary abstraction. Trigger: a reader — or a future edit — reasonably concludes that bounds already handle a moved head with no collar under it. They do not: with `avatarNeckSweater` off no offset is applied at all, so wiring one through here would silently produce bounds for an arrangement the composition never renders. Consequence: an invented state that every future reader has to disprove.

Drop the term and keep the `bottom` expression as it was. The `top` expression is where the offset genuinely belongs, and it is the one the removed clamp was hiding.

**P1 — The new test helper walks the DOM instead of using a slot**
`turbo/apps/platform/src/views/agents-page/__tests__/agents-page-avatar-framing.test.tsx:198-208`

```ts
const head = artwork?.querySelector<HTMLElement>(":scope > div, :scope > span");
```

`docs/testing/app-testing.md` says never to assert DOM identity, and `docs/styles.md` names `data-*` hooks and documented slots as the supported alternative. This PR's sibling helper already reads through `[data-avatar-artwork]`; the head group has no equivalent, so the helper pins the head to being the artwork's first `div` or `span` child. Trigger: any wrapper inserted between the artwork and the head layers, or a change from `div` to another element, breaks the test while the product is unchanged. The `:scope > span` alternative is also unreachable from this test — only the mention chip builds spans, and this test renders the agents page.

Add a second slot next to `AVATAR_ARTWORK_SLOT` for the head group, applied by both consumers the same way, and select on it.

### Testing

The new case is the right assertion for the change: it proves the mechanism that caused the bug is gone (no scale on the head) and that the replacement is doing real work (two different chins produce two different translations). It cannot prove the user-visible property — that all six faces now render at one width — because jsdom has no layout, and the file's declared exception covers exactly that.

The existing framing case is correctly updated rather than patched around. Moving the head changes the artwork bounds, so the drawn fills changed, and replacing the fixed `< 1.2` threshold with the rule's actual invariant — the framed spread is the square root of the drawn spread — is a strictly better assertion: it no longer moves when the input spread moves.

Not covered: the `!neckSweater` branch, where `headOffsetY` stays 0. That is acceptable, since with the switch released that branch is only reachable by an explicit override, and the existing `settings-tab` cases still exercise it for layer selection.

Verification claimed in the PR body — `agents-page`, `settings-tab`, `chat-composer-mentions`, `user-message-document`, plus format, lint, types, knip, and `lint:style` — matches the changed surface. All 79 required checks are currently green on this HEAD.

tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: c3ebaec841959801c349d5db08f48cd5ce4a3054
PR_AUTO_REVIEW_VERDICT: LGTM

Reviewed HEAD: `c3ebaec841959801c349d5db08f48cd5ce4a3054`
Practice revision: `075ccbaaf8f90231687267218ef1e77d05771d3a`

### Summary

Replaces the head's chin-baseline `scale` with a `translate`. Scaling reached the shared collar but also changed the head's width — 0.89x on `tall` to 1.11x on `wide` — while the 49.9px neck and 94.9px collar under it are fixed, so the body read as the wrong size for the head. The face assets are already width-normalized, so a translation reaches the same line and leaves all six faces at one width. `headScale` becomes `headOffsetY`, `AVATAR_HEAD_TRANSFORM_ORIGIN` is deleted, and the artwork's top bound stops clamping to the canvas so the framing rule can see hair that now overhangs.

Both P1s from the review of `4247edae` are resolved.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and all consumers | Pass | Both `avatarSvgComposition` consumers updated in step, and both now apply the same two slots |
| `REVIEW.md` | Meaningful coverage for new behavior; refactors stay protected | Pass | New case asserts no head carries a scale, both carry a translation, and two different chins move by different amounts; the existing framing case is updated for the new bounds |
| `REVIEW.md` | Tests verify externally observable state, not internal detail | Pass | Boundary exception declared at the top of the file; both helpers now read through declared slots rather than tree position |
| `REVIEW.md` | Report concrete practice violations with rule and source location | Pass | The dead bound offset is gone; `composerContentBounds` bottom is back to the drawn position with a comment saying why no offset belongs there |
| `REVIEW.md` | Persisted data, contracts, response validation | N/A | No schema, contract, or persisted shape changes; `headScale`/`headOffsetY` is internal with both consumers in this diff |
| `REVIEW.md` | Feature/capability containment for a changed existing flow | Pass | A fix to `avatarNeckSweater`, already released; the feature-switch guide does not require a switch for a bug fix to an existing feature |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | `applySlot` is typed `Readonly<Record<string, string>>`; lint and check-types clean |
| `docs/bad-smell.md` | Catch only at a boundary that handles the error | N/A | No error handling introduced |
| `docs/bad-smell.md` | Document changed public contracts and assess consumers | Pass | `AvatarSvgComposition.headScale` → `headOffsetY` with an updated doc comment; the removed `AVATAR_HEAD_TRANSFORM_ORIGIN` export had no other consumer, confirmed by knip |
| `docs/bad-smell.md` | Static imports in production code | Pass | No dynamic imports |
| `docs/bad-smell.md` | Smallest necessary abstraction; no path nobody executes | Pass | The unreachable `+ headOffsetY` on the bottom bound is removed |
| `docs/fallback.md` | Do not fall back on states the contract cannot produce (§3) | Pass | Resolved — the bottom bound no longer models a moved head with no collar, a combination the two call sites cannot produce |
| `docs/fallback.md` | Removing a fallback needs evidence, and removal needs no declaration | Pass | The `Math.max(0, …)` clamp is removed with its reason in code and PR body: a head that moves up makes the overhang real and the framing rule must measure it |
| `docs/fallback.md` | New fallbacks declared in a `Fallbacks` PR-body section | Pass | No new rollout fallback; the PR states its ungated decision explicitly |
| `docs/testing.md` | Platform tests boot the real Router and assert page-observable state | Pass | `setupPage` on the real Agents route with a real feature-switch override |
| `docs/testing.md` | Mock only external services; use `testContext()` lifetimes | Pass | `context.mocks.data.*` only; no `vi.mock`, no global MSW server |
| `docs/testing/app-testing.md` | Never assert DOM identity | Pass | Resolved — `[data-avatar-head]` replaces the `:scope > div, :scope > span` walk |
| `docs/testing/app-testing.md` | Prefer the sanctioned assertion set | Pass | Within the declared exception the assertions are shape-level, not pinned numbers |
| `docs/testing/testing-external-behavior.md` | State why the external interface cannot observe the case | Pass | File-level exception plus a note on the new helper |
| `docs/styles.md` | Tailwind utilities are the only component-facing styling API | Pass | No new selector, CSS module, `<style>` element, or injection; the inline transform replaces an existing inline transform, and `lint:style` is clean |
| `docs/styles.md` | Runtime behavior and tests use roles, names, refs, `data-*` hooks or slots | Pass | Two shared slot constants, applied identically by both consumers |
| `.claude/skills/feature-switch/SKILL.md` | A switch is not required for a bug fix to an existing feature | Pass | `avatarNeckSweater` is `enabled: true`; this restores its intended geometry rather than adding behavior |
| `.claude/skills/ccstate/SKILL.md` | Derivation in `computed`; no store mutation during render | Pass | No signal changes; the component still reads its switches with `useGet` |
| `.claude/skills/ccstate/SKILL.md` | Async work has an owner and cancellation path | N/A | No async work |

### Findings

None.

### Non-blocking observations

- In `avatarSvgComposition` the local `headOffsetY` is in canvas units while the field of the same name on the returned composition is a percentage. Both uses are correct and the field is documented, but the shared name is a trap: passing the returned value back into `composerContentBounds` would be wrong by a factor of 3.8 and would still type-check.
- The new test proves the mechanism — no scale on the head — rather than the user-visible result that all six faces now render at one width. That result has no page-observable surface under jsdom, so the mechanism is the strongest available assertion, but it does mean a future change to the face assets' widths would not be caught here.

### Testing

The new case is the right assertion for the change: it proves the mechanism that caused the bug is gone and that the replacement does real work, with two different chins producing two different translations. The existing framing case is correctly updated rather than patched around — moving the head changes the artwork bounds, and replacing the fixed `< 1.2` threshold with the rule's actual invariant, that the framed spread is the square root of the drawn spread, is a strictly better assertion because it no longer moves when the input spread moves.

Both test helpers now read through declared slots that the React preview and the mention-chip node view apply identically, so the two consumers cannot drift apart and neither helper depends on the layer stack's shape.

Not covered: the `!neckSweater` branch, where `headOffsetY` stays 0. Acceptable — with that switch released the branch is only reachable by an explicit override, and the `settings-tab` cases still exercise it for layer selection.

Verification on this HEAD: `agents-page-avatar-framing`, `chat-composer-mentions`, `settings-tab` pass; `format`, `lint`, `check-types`, `knip`, `lint:style` clean. The full `@okouai/app` suite is left to CI.

PR #32851, opened 2026-09-09
test(rust): synchronize queued control requests before close and fail

### LINKED ISSUE #32839: test(rust): synchronize queued control requests before close and fail
## Summary

The two queued-request lifecycle tests can pass without ever exercising a forwarder waiting on the connected-stream gate. This leaves close/fail wakeup regressions dependent on OS thread scheduling instead of deterministically checked.

Audited main: `2271fe947f4df8e4925fe5a3d1d6889a13c6b953`.

## Evidence

- `crates/guest-control-server/src/exec_control/tests/sink_lifecycle.rs:120-159` and `:161-200` hold the stream guard, reserve a pending slot, spawn the forwarding thread, and immediately call `sink.close()` / `sink.fail()`. There is no acknowledgement that the worker has reached the busy-stream wait. Reserving a pending slot in the parent does not establish this.
- `crates/guest-control-server/src/exec_control/forward.rs:99-100` first calls `wait_for_stream`, and only calls `lock_until` if that succeeds.
- `crates/guest-control-server/src/exec_control/sink.rs:226-269` immediately returns the terminal status when the sink is already closed/failed. The actual connected-stream wait and wakeups are at `:336-405`.

[Sampled tests at the audited commit](https://github.com/vm0-ai/vm0/blob/2271fe947f4df8e4925fe5a3d1d6889a13c6b953/crates/guest-control-server/src/exec_control/tests/sink_lifecycle.rs#L120)

## Concrete missed regression

An allowed schedule is: the parent spawns the worker, completes close/fail, and only then the worker starts. The worker returns the expected terminal status directly from `wait_for_stream`; the stream gate's condition variable is never used. Every current response, pending-count, and no-delivery assertion still passes.

Consequently, removing the connected-stream notification from `notify_waiters` or `mark_failed` can still pass these tests under that schedule, even though a genuinely queued forwarder would remain asleep until its request deadline. Holding the stream guard only guarantees contention if the worker reaches it before the terminal transition.

The adjacent tests cover terminal-handle rejection, close/fail while a guard is busy, and interruption of an already-started socket read. None establishes a queued gate waiter before close/fail.

## Acceptance criteria

- Add deterministic test synchronization confirming that the worker has registered its connected-stream wait before triggering close/fail. A signal immediately before calling the forwarding function is insufficient; synchronize with the gate wait itself.
- Keep the original stream guard held until the queued worker has completed, and bound synchronization/completion waits.
- Preserve assertions for the exact terminal response, zero pending slots, and no request delivered to the peer.
- Verify that suppressing the corresponding connected-stream wakeup makes the test fail reliably, independent of thread scheduling.

## Validation

`cargo test --locked --offline -p guest-control-server --lib queued_control_request_is_not_delivered -- --nocapture`: **2 passed** on the unchanged audited commit. The finding is a source-proven coverage gap, not a claim that the current production wakeup implementation is broken. Open-issue searches for the test names, sink, control, queued, and wake found no matching report.

seven332:
The queued close/fail tests could pass when the forwarder started after the sink had already terminated, leaving connected-stream wakeup regressions untested. Add a per-stream observer under `cfg(test)` that confirms entry into the real condition-variable wait using the same gate mutex, then require worker completion within one second while retaining the original stream guard.

Both tests preserve exact protocol responses, pending-slot cleanup, and peer no-delivery assertions. Production synchronization and protocol behavior are unchanged.

Closes #32839

Validation:

- Both focused queued-request tests passed.
- Independently removed each connected-stream wakeup notification: the corresponding test failed at its completion bound in 5/5 runs for close and 5/5 for fail. All mutations were restored.
- `cargo fmt --manifest-path crates/guest-control-server/Cargo.toml --check`
- `cargo test --manifest-path crates/Cargo.toml --locked --offline --profile local -p guest-control-server --all-features`: 313 passed; 3 production-identity tests ignored because they require root and a production sandbox account.
- `cargo clippy --manifest-path crates/Cargo.toml --locked --offline --profile local -p guest-control-server --all-targets --all-features -- -D warnings`
- `cargo doc --manifest-path crates/Cargo.toml --locked --offline --profile local -p guest-control-server --all-features --no-deps`

### DIFF
--- a/crates/guest-control-server/src/exec_control/sink.rs
+++ b/crates/guest-control-server/src/exec_control/sink.rs
@@ -70,6 +70,12 @@ pub(super) struct ControlStreamState {
     stream: Mutex<UnixStream>,
     gate: Mutex<ControlStreamGate>,
     ready: Condvar,
+    // The gate mutex protects test waiter observation; the atomic only
+    // supplies interior mutability without another mutex.
+    #[cfg(test)]
+    waiters: AtomicUsize,
+    #[cfg(test)]
+    waiter_ready: Condvar,
 }
 
 /// RAII guard for the forwarder that owns the connected-stream gate.
@@ -351,6 +357,10 @@ impl ControlStreamState {
             stream: Mutex::new(stream),
             gate: Mutex::new(ControlStreamGate::Available),
             ready: Condvar::new(),
+            #[cfg(test)]
+            waiters: AtomicUsize::new(0),
+            #[cfg(test)]
+            waiter_ready: Condvar::new(),
         }
     }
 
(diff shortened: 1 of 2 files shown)


seven332 (commented):
LGTM

**Reviewed head:** `1c06505b50ff304a9a6f2fb2f56b14b4a237a572`
**Practice revision:** `29040ac3b52f14bfb415bc5eeb0adcb7383b17d5`

**Summary:** The two queued lifecycle tests now acknowledge entry into the actual connected-stream wait before close/fail, then require completion while retaining the original stream guard. The observer uses the same gate mutex and a separate condition variable, and all instrumentation is compiled only for tests.

**Findings:** No findings after the six-pass self-review and code-quality review. Exact wire responses, pending-slot cleanup, and peer no-delivery assertions are preserved. No new fallback or production behavior change.

**Testing:** Both focused tests passed. Independently suppressing each connected-stream wakeup was detected in 5/5 runs per path at the one-second completion bound. After restoring notifications, 312 affected-crate tests passed; 3 production-identity tests remain ignored under the repository's environment requirements. Formatting, all-target/all-feature Clippy, and documentation checks passed. Deployed guest behavior and production identity were not exercised locally; CI is a separate gate.

This self-authored review records a code-review verdict and does not supply an independent GitHub approval.

seven332:
## Code Review: PR #32851

**Reviewed head:** `1c06505b50ff304a9a6f2fb2f56b14b4a237a572`

**Verdict:** LGTM. No findings. The complete commit-bound self-authored review is at https://github.com/vm0-ai/vm0/pull/32851#pullrequestreview-5150339705.

Verification: 312 tests passed, 3 production-identity tests ignored; format, Clippy, and docs passed. Both notification-removal controls were detected in 5/5 runs. CI and any required independent approval remain separate gates.

PR #32878, opened 2026-09-09
fix(desktop): resume permission checks after auth changes

### LINKED ISSUE #32877: bug(desktop): cancelled permission refresh leaves setup and browser tests stuck
## Problem

Desktop can remain on the permissions setup page even though macOS grants Accessibility and Screen Recording. Clicking **Test Chrome** leaves **Not tested** unchanged.

Observed on macOS 15.6 arm64 with installed Okou 0.48.16 and the default `okou` driver selected. A read-only invocation of the installed signed native helper returned both permissions `true`. Clicking Test Chrome reproduced no visible change, and no native helper remained running.

## Reproduced cause

A local isolated reproduction using the production driver controller, runtime controller, permission service and an external helper fixture established this sequence:

1. Start a passive native permission refresh.
2. Invalidate it as an authentication authority change does (`cancelPermissionRefresh` plus permission-state reset).
3. Cancellation withdraws admission and retires the driver, leaving permission access paused.
4. After cleanup, subsequent refresh and browser-probe requests return no result without reaching the native helper.
5. The permission service retains its default `false` / `unknown` values, and the setup UI exposes neither the driver state nor a recovery action.

The control without cancellation returns both grants and a successful Chrome probe. The first triggering event in the user's running process was not captured, so the authentication race is a proven reproducible trigger, not a historical log claim.

## Expected behavior

- Superseded work cannot publish stale permissions or admit actions.
- Once lifecycle-owned cleanup completes, the current authorized state can inspect permissions again.
- Permission and browser-test actions must not silently remain stuck after authentication refresh.
- Preserve generation ownership, cancellation, explicit stop/quit intent and CUA authorization boundaries.

## Acceptance

- Add a regression for auth invalidation racing a passive native permission read, with real lifecycle/permission modules and an external helper boundary.
- Verify subsequent grants and browser-probe results, plus no premature recovery during cleanup and no stale revival after newer intents.
- Keep temporary helper data, unrelated app inventories and raw screenshots out of published artifacts.

lancy:
An authentication authority change could cancel a passive permission refresh and leave the default Okou driver paused after its helper exited. Subsequent permission reads and Test Chrome requests returned no result, so setup kept showing missing permissions and “Not tested” despite valid macOS grants.

Use the existing `stopForAuthChange` lifecycle for both drivers. It cancels stale work, owns runtime/helper cleanup, and resumes permission inspection only after cleanup for the current intent. Previously only CUA completed that lifecycle; the default driver only received cancellation. Permission-state reset still invalidates stale replies. This does not reactivate native commands or bypass stop/quit/authorization gates.

Fixes #32877.

## Validation

- Reproduced the new regression on old code: after helper cleanup both permissions remained false.
- 99 tests passed across Desktop driver selection, driver integration and runtime-controller suites.
- New cases cover an auth change during a held passive probe, no second helper during cleanup, restored grants and Chrome results afterward, a newer sign-out, and quit preventing revival. The fixture mocks only the external helper process and HTTP; lifecycle, permissions, auth and host modules remain real.
- Desktop type checks, lint and changed-file formatting passed.
- Local scoped Knip still reports the same pre-existing `@sentry/cli` and `sharp` devDependencies. Relevant configuration, dependencies and scripts are unchanged from the previously verified baseline; only the Desktop package version changed. No check configuration was weakened.

This is a source fix; the user's installed app and macOS grants were not modified. The installed 0.48.16 signed helper independently returned both grants as true during diagnosis.

### DIFF
--- a/turbo/apps/desktop/src/desktop-computer-use-selection.test.ts
+++ b/turbo/apps/desktop/src/desktop-computer-use-selection.test.ts
@@ -62,10 +62,10 @@ const fs = require('node:fs');
 const blocked = fs.existsSync(${JSON.stringify(permissionGate)});
 const lines = require('node:readline').createInterface({input:process.stdin});
 if (blocked) lines.on('close', () => {
-  fs.appendFileSync(${JSON.stringify(nativeEvents)}, 'exit-waiting\\n');
   const watcher = fs.watch(${JSON.stringify(directory)}, () => {
     if (fs.existsSync(${JSON.stringify(exitRelease)})) { watcher.close(); process.exit(0); }
   });
+  fs.appendFileSync(${JSON.stringify(nativeEvents)}, 'exit-waiting\\n');
 });
 fs.appendFileSync(${JSON.stringify(nativeEvents)}, 'start\\n');
 lines.on('line', line => {
--- a/turbo/apps/desktop/src/desktop-computer-use-selection.test.ts
+++ b/turbo/apps/desktop/src/desktop-computer-use-selection.test.ts
@@ -49,6 +49,8 @@ function desktop(initial?: string) {
   const file = path.join(directory, "preferences.json");
   const helper = path.join(directory, "helper.cjs");
   const nativeEvents = path.join(directory, "native-events");
+  const permissionGate = path.join(directory, "permission-gate");
+  const exitRelease = path.join(directory, "exit-release");
   writeFileSync(nativeEvents, "");
   if (initial !== undefined) writeFileSync(file, initial);
   // The actuator is an external process; the real native protocol client,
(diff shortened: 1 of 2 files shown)


lancy (commented):
LGTM

## Code Review: PR #32878 (Round 1)

Inspected HEAD: `dfeb84a8548969d5c258c999ef243a2a88e403f3`.

### Summary

Authentication authority changes now use the existing `stopForAuthChange` lifecycle for the default Okou driver as well as CUA. Permission-state reset prevents stale publication; lifecycle-owned cleanup restores passive permission access only for the current intent and does not admit commands.

### Findings

No P0/P1 or other actionable findings remain. Reviewed the complete two-file diff and traced cancellation, `supersede`, driver retirement, `detachRuntime`'s intent/quit guard, startup, explicit Stop and auth completion. The former default-driver branch cancelled the probe without the cleanup completion that resumes permission reads. Reusing the existing lifecycle removes that incomplete path without introducing a fallback, timer or independent recovery owner.

The test fixture's release watcher is installed before it announces readiness, avoiding a missed release notification.

### Testing

Adequate coverage through real auth, permission, runtime, native protocol and host modules with external helper/HTTP fixtures. The old wiring fails the new post-cleanup grants assertion. The new cases prove no replacement during held cleanup, fresh permissions and Chrome results afterward, preservation across a newer sign-out, and no revival after quit; no host or CUA actuator is implicitly started.

Local verification: 99 related tests passed; all three new cases passed again after fixture synchronization. Desktop types, lint, formatting, style policy and commit checks passed. Scoped Knip has the documented unchanged local baseline findings. CI is evaluated separately before merge.

Practices: main revision `5a82c5384c047d91976fce66f889944b1ba5bd91`, following REVIEW.md, code-quality, testing, Desktop testing and fallback guidance. No protocol or persistence boundary changed.

Verdict: LGTM. P0: 0; P1: 0.

PR #32927, opened 2026-09-09
feat(api): rank connector discovery and give every category a slice

tongx-blip:
## Why

Measured against the live catalogue (4 001 connectors) and the code on main:

- `CONNECTOR_DISCOVERY_LIMIT = 100` meant **2.5% of the catalogue was browsable**. The other 3 901 existed only for someone who already knew the name.
- That 100 was `FEATURED_CONNECTOR_SLUGS`, a hardcoded array whose own comment dates it: *"Ranked from MaskDB connector rows on 2026-08-10."*
- It was ranked on **our own** connection frequency, so it promoted what the Okou team connects. Positions 51–100 held `maskdb` (our own internal tool), `db9`, `drive9`, `slock`, `runtime`, Steam and WeRead — while **Stripe sat at 97, Shopify at 85, Zendesk at 92, and Salesforce was absent entirely**.
- Eight of the twelve categories were usually missing from the first screen: a category with 1 184 connectors and one with 47 competed in the same global ranking.
- Inside a category the order was alphabetical, so browsing opened on `123FormBuilder · 1Password · 1SaaS · 1Shot API · 21RISK · 3Commas · AITable.ai · AMC Theatres`.

Full evidence, with the numbers and the source lines: https://okou-connector-ranking.okou.app

## What

New `turbo/apps/api/src/signals/services/connector-popularity.ts`, replacing `connector-catalog-featured.ts`:

- **`INTERNAL_CONNECTOR_SLUGS`** — the five connectors Okou operates for itself are removed from discovery and search. Still connectable by slug; they just stop competing for a customer's first screen.
- **`CONNECTOR_POPULARITY_RANKING`** — 107 slugs, every one verified present in the live catalogue, ordered against the product's stated audience (non-technical business users first, technical tail after). Gmail 0, Slack 11, **Stripe 18, Salesforce 22**.
- **Per-category slices** — discovery returns the top `CONNECTOR_DISCOVERY_PER_CATEGORY` (12) of each category rather than one global top-100, so every category reaches the first screen.
- **Ranked search** — matches are ordered by the same rank, so "mail" leads with Gmail instead of the first alphabetical hit.

Contract additions, both **optional** so an older API and a newer app can span versions (`docs/deployment-compatibility.md`):

- `popularityRank` on the catalogue item
- `categoryConnectorCounts` on the discovery response, so the client can offer "show all 323" without a second request

The platform's `sortedCategoryConnectors` now orders connected → rank → label.

## This is a seed, not the destination

The module says so in its own doc comment. The ranking **should** be computed weekly from distinct connecting organisations with our own orgs excluded — that is what turns this from one hand-list into a mechanism. I could not compute it in this change because the **maskdb connector is not authorised for this agent**, so the real numbers were unavailable. Everything here is shaped so that landing that job changes one generated list and nothing else.

## Scope correction found by CI

`/api/connectors/search` shares the catalog reader with discovery, so the first version of this change routed that endpoint's keyword-free branch through the per-category slice and dropped `neon` and `openai` from a keyword-free search. `connectors-search.test.ts` caught it.

Only discovery browses by category now. A keyword-free `/api/connectors/search` returns a ranked head of the whole catalogue, and the internal-connector exclusion is scoped to discovery — agents keep seeing every connector they can actually connect.

## Verification

- `tsc -p tsconfig.gateways.json` then `tsconfig.core.json` — clean apart from pre-existing `@okouai/pi-agent-runtime` module-resolution errors that need `check-types:deps`
- `tsc -p tsconfig.production.json` (platform) — clean
- `oxlint --deny-warnings` on both apps — clean
- New `connector-popularity.test.ts` (7 cases). API tests need Postgres, which this environment has none of, so the logic was additionally verified directly: 107 ranked, 107 unique, zero internal slugs leaked into the ranking, ranked sorts before unranked.

All 39 checks green. One app shard failed once on `team-navigation.test.tsx` ("agent chat shortcut opens the first available thread"), which touches no connector code and failed on a loading skeleton; rerunning the failed job alone passed.

Not covered: no UI change here. The shelf layout that consumes `categoryConnectorCounts` is the next PR.

Preview: https://pr-32927-app-okou-app-preview.vm0.workers.dev?x-vercel-protection-bypass=jAdTONWK3jqOzLaGJYjS4kINOvXBRmvB


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: 0595f5a6bbffd21ff63201b51c36f8be569166c6
PR_AUTO_REVIEW_VERDICT: LGTM

Practice revision: `99bd2254849a4890c0bf7dc12c151b6463b34726` (docs/deployment-compatibility.md, docs/fallback.md, docs/testing.md, docs/bad-smell.md)

## Summary

Replaces the frozen 100-slug discovery array with a ranking module, and changes what a keyword-free discovery call returns from one global top-100 to a per-category slice. Six files, +376/-144.

## Review criteria

**Deployment compatibility (docs/deployment-compatibility.md).** Both response additions are optional: `popularityRank` on the catalogue item and `categoryConnectorCounts` on the discovery response. An older API paired with a newer app omits them and the app falls back — `sortedCategoryConnectors` treats a missing rank as `Number.MAX_SAFE_INTEGER`, which reproduces the previous label-only ordering rather than throwing. A newer API with an older app sends fields the old zod schema ignores. Both directions hold.

**Fallbacks (docs/fallback.md).** No fallback path is removed. Unranked connectors are not dropped, only ordered last, and the internal exclusion is scoped to the discovery surface — `getExternalPublicConnectorCatalogStatus` still resolves those slugs, so nothing that could previously be connected becomes unreachable.

**Blast radius, and the one real defect this PR already fixed.** `searchEffectiveConnectors` is shared between discovery and `/api/connectors/search`. The first revision routed the agent-facing endpoint's keyword-free branch through the per-category slice, which dropped `neon` and `openai`; `connectors-search.test.ts` caught it and commit `0595f5a` separated the two. I re-checked every remaining caller of the reader: `listExternalPublicConnectorCatalog`, `getExternalPublicConnectorCatalogStatus` and `getExternalPublicConnectorCatalogPermissionDetail` are untouched by the ordering change, and only `discoverExternalPublicConnectorCatalogStatus` browses by category.

**Code quality (docs/bad-smell.md).** No `any`, no suppression comments. `api/no-package-variable` is respected: the ranking is a readonly tuple and the lookup Map is built per read. The index is threaded through `connectorCatalogStatusRead` and `listExternalPublicConnectorCatalog` so it is built once per response rather than once per connector, which matters at 4 001 items.

**Correctness of the ordering.** `compareConnectorPopularity` is a total order: rank ascending, then `localeCompare` on the label, so `Array.prototype.sort` is stable and deterministic. `browseEffectiveConnectors` preserves the global rank order inside each bucket because it fills buckets from an already-sorted list, which is what makes the route test's monotonic-rank assertion meaningful.

## Findings

None blocking.

One thing the PR states rather than hides, and which I agree belongs in the description rather than in code: `CONNECTOR_POPULARITY_RANKING` is a curated seed, not computed data. The module's doc comment says so and names what should replace it. The maskdb connector is not authorised for this agent, so the real distinct-organisation counts could not be produced here. This is an improvement over the array it replaces on its own terms — it removes our internal tooling from customers' first screen and moves Stripe from 97 to 18 — but it is not the mechanism the problem ultimately needs.

## Testing

- `connector-catalog.test.ts` gains a discovery case asserting the per-category cap, that `categoryConnectorCounts` covers every returned category, that ranks never go backwards inside a category, and that no internal connector appears. It states the endpoint's promise as literals rather than importing internals, which the repository's `no-restricted-imports` rule requires of API tests.
- `tsc` on `tsconfig.gateways.json`, `tsconfig.core.json`, `tsconfig.tests.json` (API) and `tsconfig.production.json` (platform)
- `oxlint --deny-warnings` on both apps, `eslint` on the changed files, `knip --workspace apps/api`
- All 39 PR checks green. One app shard failed once on `team-navigation.test.tsx`, which imports nothing this PR touches and failed on a loading skeleton; rerunning that job alone passed.

PR #32928, opened 2026-09-09
fix(runner): preserve jsonl framing after partial append failures

### LINKED ISSUE #32920: test(python): cover JSONL recovery after partial append failures
## Summary

The JSONL writer tests do not cover a real short write followed by an I/O error and a later successful append to the same file. A deterministic regression test exposes a framing failure: the incomplete failed record remains at EOF, and the first successfully written record after recovery is appended directly to it. The production network-log uploader then skips the combined malformed line, losing the otherwise healthy recovered record as well.

Audited commit: `4d4f6b78893a82caa720a150b5416eed105e709b`.

## Code and coverage gap

- [`crates/runner/mitm-addon/tests/test_jsonl_writer.py:38-84`](https://github.com/vm0-ai/vm0/blob/4d4f6b78893a82caa720a150b5416eed105e709b/crates/runner/mitm-addon/tests/test_jsonl_writer.py#L38-L84) exercises real short writes, but every write eventually succeeds.
- [`crates/runner/mitm-addon/tests/test_jsonl_writer.py:87-161`](https://github.com/vm0-ai/vm0/blob/4d4f6b78893a82caa720a150b5416eed105e709b/crates/runner/mitm-addon/tests/test_jsonl_writer.py#L87-L161) exercises a zero-byte write and recovery from missing directories. Neither failure leaves a partial JSON record behind.
- [`crates/runner/mitm-addon/src/jsonl_writer.py:306-337`](https://github.com/vm0-ai/vm0/blob/4d4f6b78893a82caa720a150b5416eed105e709b/crates/runner/mitm-addon/src/jsonl_writer.py#L306-L337) advances through short `os.writev` results, but an exception after partial progress exits without retaining or restoring the record boundary. `_write_batch` catches the error at lines 264-270, and the worker retires the batch at lines 246-247. Later writes reopen the file with `O_APPEND`.
- [`crates/runner/mitm-addon/src/logging_utils.py:116-121`](https://github.com/vm0-ai/vm0/blob/4d4f6b78893a82caa720a150b5416eed105e709b/crates/runner/mitm-addon/src/logging_utils.py#L116-L121) provides the production `log_network_entry` entry point used by the reproduction below.
- [`crates/runner/src/network_logs.rs:276-285`](https://github.com/vm0-ai/vm0/blob/4d4f6b78893a82caa720a150b5416eed105e709b/crates/runner/src/network_logs.rs#L276-L285) parses each newline-delimited record independently and skips malformed JSON. It cannot recover the healthy record embedded in this corrupted line.

This is not a claim that `flush_log_path()` guarantees persistence: its documented contract explicitly counts failed attempts as processed (`logging_utils.py:81-89`). Losing the failed record is distinct from corrupting the next record after I/O has recovered. No process crash or concurrent writer is needed to reproduce the latter.

## Deterministic reproduction

Run the following as an isolated pytest module against the addon `src` directory on Linux. The fault injection performs a real seven-byte write to a temporary file before raising `ENOSPC` or `EIO` on the next call; it does not fill the disk, use sleeps, or mock the writer itself.

```python
import errno
import json
import os

import pytest

import jsonl_writer
import logging_utils


@pytest.mark.parametrize("failure_errno", [errno.ENOSPC, errno.EIO])
def test_recovered_network_record_keeps_its_jsonl_boundary(
    tmp_path, monkeypatch, failure_errno
):
    jsonl_writer.reset_for_tests()
    log_path = tmp_path / "network.jsonl"
    original_writev = os.writev
    calls = 0

    def partial_then_error(fd, buffers):
        nonlocal calls
        assert os.readlink(f"/proc/self/fd/{fd}") == str(log_path)
        calls += 1
        if calls == 1:
            return original_writev(fd, [memoryview(buffers[0])[:7]])
        raise OSError(failure_errno, "injected after a real partial write")

    def write_and_flush(record_id):
        logging_utils.log_network_entry(str(log_path), {"id": record_id})
        assert logging_utils.flush_log_path(str(log_path), timeout=2)

    try:
        write_and_flush("before")
        with monkeypatch.context() as fault:
            fault.setattr(jsonl_writer.os, "writev", partial_then_error)
            write_and_flush("failed")
        assert calls == 2
        assert jsonl_writer._queued_writes == 0
        assert jsonl_writer._pending_bytes == 0
        write_and_flush("recovered")
        write_and_flush("later")

        # Match the production uploader's skip-malformed-lines behavior.
        record_ids = []
        for line in log_path.read_text().splitlines():
            try:
                record_ids.append(json.loads(line)["id"])
            except json.JSONDecodeError:
                continue

        assert record_ids == ["before", "recovered", "later"]
    finally:
        jsonl_writer.reset_for_tests()
```

Observed for both error codes:

```text
AssertionError: assert ['before', 'later'] == ['before', 'recovered', 'later']
2 failed in 0.11s
```

The existing suite passes on this commit: **7,464 passed** in normal order and **7,464 passed** with a deterministic shuffled order, on Python 3.14.0 and the locked test dependencies. The two failures above are new isolated regression candidates, not failures in the existing suite. An independent direct-writer probe also reproduced the corrupted line twice in fresh processes.

## Acceptance criteria

- Add deterministic coverage for partial progress followed by append failure, then restore normal I/O and write another record to the same path. Exercise at least an error in the middle of a record and the record-boundary/batch case.
- Assert externally visible JSONL framing: previously complete records remain intact, and records successfully written after recovery are independently parseable rather than consumed by a failed prefix. Do not require persistence of the original failed record or reinterpret flush completion as durability.
- Verify pending-byte/write accounting, subsequent progress, and flush/shutdown cleanup; keep fault injection and worker teardown isolated from other tests.
- Implement or explicitly establish a safe recovery policy with the regression coverage. Account for other appenders to the shared per-run network log (including `crates/runner/src/network_log_manager/file_append.rs`); do not blindly truncate bytes that another producer may have appended.

Deduplication checked open issues using title searches for `JSONL`, `JSONL partial`, `append`, and `log corruption`, plus body searches for `jsonl_writer`, `partial record`, `short write`, `partial write`, and `ENOSPC`; no matching open issue was found.

seven332:
## Summary

A short JSONL write followed by an I/O error leaves an incomplete record at EOF. The next healthy addon record was appended to that prefix and discarded by the Runner uploader as part of a malformed line. Inspect the final byte before each nonempty batch and append a separator through the existing short-write loop when needed.

Recovery preserves existing bytes, including records from the independent Rust appender, and keeps failed-batch accounting and flush semantics unchanged. It uses bounded EOF inspection without retaining failed paths. The log schema is unchanged; this does not add crash durability or serialize concurrent producers' short-write sequences, and cannot salvage records already embedded in malformed lines.

Closes #32920.

## Validation

- Added 16 deterministic cases through `logging_utils.log_network_entry`, using real temporary files and OS-boundary fault injection. Covers ENOSPC/EIO/zero after partial progress, record boundaries, repeated recovery failures, restart, a written separator followed by failure, another appender, and failed EOF inspection/descriptor cleanup. The six mid-record cases reproduced loss of the recovered record before the fix.
- `uv run --no-sync python -m pytest tests/test_jsonl_writer_recovery.py tests/test_jsonl_writer.py tests/test_logging_utils.py tests/test_runner_jsonl_flush.py -q` — 87 passed. The 16 new cases were rechecked after tightening an exception matcher for lint.
- `uv run --no-sync ruff format --check .`, `uv run --no-sync ruff check .`, and `uv run --no-sync basedpyright -p .` — passed in the locked Python 3.14.0 environment.
- Prettier 3.8.1 check for the updated logging contract, file-size check, and `git diff --cached --check` — passed.

### DIFF
--- a/crates/runner/mitm-addon/src/jsonl_writer.py
+++ b/crates/runner/mitm-addon/src/jsonl_writer.py
@@ -46,7 +46,9 @@ def write_jsonl_line(log_path: str, line: bytes, log_name: str) -> None:
 
     When accepted, ``line`` is passed to the append worker without transformation.
     The caller owns JSON serialization and the terminating newline; this function
-    validates neither JSON nor record boundaries.
+    validates neither JSON nor record boundaries. Before a nonempty batch, the
+    worker adds a newline if the existing file tail is unterminated, isolating
+    subsequent records from a previous incomplete append.
 
     The call returns no admission or durability result. An empty ``log_path``,
     writer shutdown, pending-write or pending-byte saturation, or failure to start
@@ -304,8 +306,15 @@ def _report_append_recovery() -> None:
 
 
 def _append_lines(log_path: str, lines: list[bytes]) -> None:
-    fd = os.open(log_path, os.O_CREAT | os.O_APPEND | os.O_WRONLY, 0o644)
+    fd = os.open(log_path, os.O_CREAT | os.O_APPEND | os.O_RDWR, 0o644)
     try:
+        if any(lines):
+            size = os.fstat(fd).st_size
+            if size and os.pread(fd, 1, size - 1) != b"\n":
+                # Only append: Rust can also write to the shared network log,
+                # so rolling back to an earlier offset could delete its records.
+                lines = [b"\n", *lines]
+
         line_index = 0
         line_offset = 0
         while line_index < len(lines):
--- a/crates/runner/mitm-addon/tests/test_jsonl_writer_recovery.py
+++ b/crates/runner/mitm-addon/tests/test_jsonl_writer_recovery.py
@@ -1,0 +1,221 @@
+"""Recovery framing through the production JSONL logging entry point."""
+
+import errno
+import json
+import threading
+from pathlib import Path
+from unittest.mock import patch
+
+import pytest
+
+import jsonl_writer
+import logging_utils
+
+
+def _write_and_flush(path: Path, record_id: str) -> None:
+    logging_utils.log_network_entry(str(path), {"id": record_id})
+    assert logging_utils.flush_log_path(str(path), timeout=2)
+
+
+def _record_ids(path: Path) -> list[str]:
+    # The Runner uploader skips malformed physical lines independently.
+    records = []
+    for line in path.read_bytes().splitlines():
+        try:
+            records.append(json.loads(line)["id"])
+        except (json.JSONDecodeError, UnicodeDecodeError):
+            continue
+    return records
+
+
+def _assert_drained() -> None:
+    assert jsonl_writer._pending_bytes == 0
+    assert jsonl_writer._queued_writes == 0
+    assert not jsonl_writer._accepted_by_path
+    assert not jsonl_writer._completed_by_path
+    assert not jsonl_writer._flush_waiters_by_path
+
+
+@pytest.mark.parametrize("failure_errno", [errno.ENOSPC, errno.EIO, None])
+@pytest.mark.parametrize("failure_position", ["first-record", "record-boundary", "second-record"])
+def test_partial_batch_failure_preserves_recovered_records(
+    tmp_path, mitm_ctx, failure_errno, failure_position
+):
+    path = tmp_path / "network.jsonl"
+    gate_started = threading.Event()
+    release_gate = threading.Event()
+    original_writev = jsonl_writer.os.writev
+    calls = 0
+
+    def writev(fd: int, buffers: list[bytes | memoryview]) -> int:
+        nonlocal calls
+        calls += 1
+        if calls == 1:
+            gate_started.set()
+            release_gate.wait()
+            return original_writev(fd, buffers)
+        if calls == 2:
+            complete_count = 0 if failure_position == "first-record" else 1
+            partial_buffers = list(buffers[:complete_count])
+            if failure_position != "record-boundary":
+                partial_buffers.append(memoryview(buffers[complete_count])[:7])
+            return original_writev(fd, partial_buffers)
+        if failure_errno is None:
+            return 0
+        raise OSError(failure_errno, "injected after partial progress")
+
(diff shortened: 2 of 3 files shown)


seven332 (commented):
LGTM

**Summary:** Restores the boundary before a new addon JSONL batch when a previous append left an unfinished tail. The separator goes through the existing short-write loop, preserves all existing bytes, and leaves failed-batch/flush accounting unchanged.

**Findings:** No actionable findings. Checked record-boundary and partial-record errors, separator failure, current-tail inspection, independent appenders, descriptor cleanup, and the production reader. EOF inspection is bounded and runs on the background worker; production-owned regular files support the added read access.

**Testing:** 87 focused writer/logging/Runner-flush tests passed; all 16 new recovery cases were rechecked after tightening a test matcher. Ruff formatting/lint, BasedPyright, pinned Prettier, whitespace, and file-size checks passed. The six mid-record error/zero cases reproduced loss of the recovered record before the fix. Full CI is assessed separately.

**Limits:** The change does not promise crash durability, recover records already embedded in a malformed line, or serialize independent producers' short-write sequences. These existing limits are documented.

**Reviewed head:** `e9ecc0588cca95706e993f03f063ca245bbd6ce0`.
**Repository practices:** `9a7aa1b3d890eb1de5bd5cb660fdddec0bb14315`.
Self-authored PR: this is a commit-bound COMMENT review, not a GitHub approval.

seven332:
## Code Review: PR #32928

**Reviewed head:** `e9ecc0588cca95706e993f03f063ca245bbd6ce0`
**Verdict:** LGTM — no actionable findings after the full six-pass self-review and code-quality review.

[Full review anchored to this commit](https://github.com/vm0-ai/vm0/pull/32928#pullrequestreview-5152403296), including scope, test evidence, and remaining limits. Local verification passed (87 focused tests, the final 16-case regression recheck, Ruff, BasedPyright, Prettier, whitespace, and file-size checks). This workflow marker records code review; it does not constitute GitHub approval or CI/merge completion.

PR #32955, opened 2026-09-09
feat(platform): add model picker flyout layout behind a switch

tongx-blip:
## What

Adds `modelPickerFlyout` (staff-visible in Lab, off by default) and, when it is
on, replaces the compact menu's drill-in pages with **two detached panels**:
model types on the left, that type's models in a panel beside it.

Design was reviewed end to end for DES-43; the interactive prototype the
decisions were taken against is at https://des43-model-picker-review.okou.app

## Why detached rather than one panel

Joining the two columns makes the whole popover resize whenever a longer list
opens. Because it is bottom-anchored, the type rows then move out from under the
pointer and the mouse has to chase them. Separate cards keep the root a constant
size; only the floating panel resizes.

## What it fixes from the review

- The current model is reachable in one step instead of a drill-in plus a Back.
- Focus lands on the current model, not on the Back button.
- Rows are a real `listbox` with `aria-selected` / `aria-posinset`, so a screen
  reader announces "3 of 7, selected" rather than "toggle button, pressed".
- Unavailable routes keep `aria-disabled` and stay in the keyboard order; a
  native `disabled` row is invisible to keyboard and screen reader users.
- The panel opens towards whichever side has room, measured from the anchored
  root, so it never runs off the viewport.

## Scope

Desktop only for now — narrow viewports keep the existing menu until the bottom
sheet layout lands in a follow-up. The thinking-effort entry and moving Fast out
of the picker are separate changes; `thinkingLevel` has no path from the
composer to the runner yet.

## Verification

- `pnpm check-types` in `turbo/apps/platform` — clean
- `chat-composer-model-selection.test.tsx` — 13 passed, including a new case
  covering the flyout's single-panel structure, focus landing and selection
- `oxlint --deny-warnings` on the changed files — clean

## Try it

Lab → turn on **modelPickerFlyout**.

### DIFF
--- a/turbo/apps/platform/src/signals/external/feature-switch.ts
+++ b/turbo/apps/platform/src/signals/external/feature-switch.ts
@@ -133,6 +133,11 @@ export const modelPickerMenuEnabled$ = computed((get): boolean => {
   return get(featureSwitch$)[FeatureSwitchKey.ModelPickerMenu] ?? false;
 });
 
+/** The flyout replaces the drill-in menu's pages with two detached panels. */
+export const modelPickerFlyoutEnabled$ = computed((get): boolean => {
+  return get(featureSwitch$)[FeatureSwitchKey.ModelPickerFlyout] ?? false;
+});
+
 export const codexFastModeEnabled$ = computed((get): boolean => {
   return isCodexFastModeEnabled({ overrides: get(featureSwitch$) });
 });
--- a/turbo/apps/platform/src/signals/okou-page/model-picker-menu.ts
+++ b/turbo/apps/platform/src/signals/okou-page/model-picker-menu.ts
@@ -8,6 +8,11 @@ type ModelPickerMenuPage =
   | { readonly kind: "models"; readonly category: ModelPickerCategory }
   | { readonly kind: "settings" };
 
+/** Which side of the root panel the flyout has room to open towards. */
+type ModelPickerFlyoutSide = "left" | "right";
+
+const FLYOUT_PANEL_WIDTH = 258;
+
 /** One navigation state per composer, including split chats. */
 export function createModelPickerMenuSignals() {
   const internalPage$ = state<ModelPickerMenuPage>({ kind: "overview" });
@@ -16,6 +21,7 @@ export function createModelPickerMenuSignals() {
   });
   const reset$ = command(({ set }) => {
     set(internalPage$, { kind: "overview" });
+    set(internalFlyoutCategory$, "chat");
   });
   const showModels$ = command(({ set }, category: ModelPickerCategory) => {
     set(internalPage$, { kind: "models", category });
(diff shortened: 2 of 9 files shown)


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: f97267d49147fd0eca04fb81468738f6e9fe58c9
PR_AUTO_REVIEW_VERDICT: LGTM

## Summary

Adds the `modelPickerFlyout` switch and, when it is on, renders the composer's
model picker as two detached panels — model types on the left, that type's
models in a panel beside it — instead of the compact menu's drill-in pages.
Narrow viewports and the switch-off path keep the existing menu untouched.

Scope is Platform-only: no API route, contract, schema, or persisted shape
changes, so the persisted-data and compatibility gates do not apply.

## Review Criteria

- **Correctness and consumers.** `ModelProviderPicker` gains `flyoutLayout`,
  defaulted `false`, and only `ComposerRunModelPickerControl` passes it. The
  other two call sites (`chat-section.tsx`, `chat-thread-page.tsx`) are
  unchanged and keep the Select path. `PopoverContent` drops its own card
  styling only under `flyoutLayout`, so the menu layout's box is unaffected.
- **ccstate ownership.** No React hooks are introduced. `flyoutRootRef$`
  registers its `resize` listener through `onRef` and removes it on the abort
  signal; `focusFlyoutPanelRef$` only reads and focuses. `reset$` now also
  clears the flyout category so a reopened picker starts on Chat.
- **Feature containment.** The layout is reachable only through
  `modelPickerFlyoutEnabled$ && desktopLayout`. The switch is `enabled: false`
  with `STAFF_ORG_ID_HASHES`, so it is staff/Lab-only and not GA.
- **Styles.** Tailwind utilities and semantic tokens only; no first-party
  selector, CSS module, or runtime stylesheet. `style-policy` passes — an
  earlier revision grew the legacy `motion-safe:animate-in` baseline and that
  use was removed rather than allowlisted.

## Findings

No blocking findings.

Non-blocking, recorded for the follow-up that adds the narrow-screen sheet:

- The type rail is a `tablist` whose panel is a `listbox` rather than a
  `tabpanel`, and the tabs carry no `aria-controls`. Announcement and keyboard
  traversal are correct today, but the pairing is worth tightening when the
  sheet layout lands and both surfaces settle.

## Testing

  — 35 passed. Two of those cases are new: the single-panel path (focus lands on
  the current model, arrow keys move, selection applies, no Back control) and
  the two-panel path (three type tabs, switching type swaps the list in place
  with no pushed page, the current media model stays marked).
- `pnpm check-types` in `turbo/apps/platform` — clean.
- `eslint --max-warnings 0` on every changed file, `prettier --check`, and
  `node scripts/style-policy.mjs` — clean.

## Fallbacks

None. `flyoutLayout = false` and `?? false` on the switch signal are the
prop/switch defaults every other switch in this file uses, not reads of
possibly-absent persisted data, and a non-GA switch needs no cutover
compatibility handling.

PR #32975, opened 2026-09-09
fix(app): keep the image annotation editor the size of the preview it replaces

### LINKED ISSUE #32964: bug: image editor shrinks the preview dialog and shows a stray top-left overlay
## Summary

The image preview → Edit/annotation flow has two UI issues:

1. Clicking **Edit** unexpectedly shrinks the dialog and the displayed image.
2. After drawing a rectangle, an unexpected white area/overlay is visible at the upper-left of the editor. The recording shows unrelated chat-sidebar controls appearing over the dialog and remaining interactive.

## Steps to reproduce

1. Open an image in the Okou image preview dialog.
2. Click the pencil/**Edit** button in the top-right toolbar.
3. Compare the editor dialog and image size with the preceding preview.
4. Select the rectangle tool and draw a box on the image.
5. Inspect the upper-left of the editor. Click the stray **…** control if it is visible.

## Actual behavior

- **Dialog size:** The large image preview switches to a noticeably smaller annotation dialog, reducing the visible image area even though both views display **100%** zoom.
- **Upper-left overlay:** The **+** and **…** controls from the chat sidebar appear over the editor. After the rectangle is drawn, clicking **…** opens a white menu containing **Mark all read**, **All chats**, and **Unread only** on top of the editor. The controls are also visible earlier in edit mode, so drawing may not be the initial trigger.

## Expected behavior / acceptance criteria

- Entering Edit preserves a consistent dialog size and usable image area, without the unexpected shrink.
- Drawing and selecting a rectangle does not introduce any stray white area or unrelated overlay.
- Background sidebar controls and menus stay behind the modal and cannot be interacted with while the editor is open.

## Evidence

- [Original Slack report and recording](https://vm0.slack.com/archives/C0AM5FU92QH/p1788949996264899)
- Recording: `录屏2026-09-09 18.31.22.mov` (approximately 13 seconds).
- **00:00–00:01:** Preview switches to the smaller editor.
- **00:05–00:06:** A rectangle is drawn and selected.
- **Approximately 00:09.5:** The unrelated white chat menu is open at the upper-left.

Observed from the supplied recording; not independently reproduced in a live app session.

tongx-blip:
Closes #32964.

Three symptoms in the report, two causes behind them, and both causes turn out
to be the same thing: the annotation editor was a hand-rolled `fixed inset-0
z-50` overlay when the preview it replaces is a `DialogContent`.

## 1. Edit shrank the window

The preview is `DialogContent maxWidth={1440} height={1000}`. The editor was
pinned at `w-[min(980px,94vw)] h-[min(700px,90vh)]` — a second copy of the
preview's numbers, kept level by hand, and already drifted. The image inside it
was fitted to a *guess* at that stage, `min(880px, 88vw)` by `min(520px, 62vh)`,
so the picture shrank while the zoom readout still said 100%.

The panel now renders through the shared `Dialog`/`DialogContent` with the same
`maxWidth={1440} height={1000} surface="canvas"`, which deletes the duplicated
size classes with it, and the stage is a size query container so the image fit
is `100cqw`/`100cqh` — the box the image is actually in.

Measured on the preview at a 1680×1050 viewport:

| | panel | image |
| --- | --- | --- |
| Preview lightbox | 1440×1000 | 1371×896 |
| Editor, before | 980×700 | 796×520 |
| Editor, after | 1440×1000 | 1357×848 |

## 2. The number was under the drag handle

A box mark's ordinal disc was centred on the box's own top-left corner, which is
exactly where the selection grip lands. Selecting a box hid its number under the
white dot, and a note that reads "1" had nothing left to point at.

The pin now sits a radius *inside* the corner. Inside rather than outside
because a mark drawn against the top or left edge of a screenshot would push an
outside pin off the image entirely, and the flattened copy is what the vision
model reads. `flatten-annotated-image` draws the disc at the same offset from
the same shared constants.

## 3. The white menu over the editor

Not a stray sidebar control — the composer's own slash-command menu. `z-50`
only orders the editor inside the app's own stacking context, so a menu
portalled to `body` kept painting on top of it and stayed clickable. Reproduced
by opening `/` in the composer, then the image preview, then Edit. Going
through the dialog primitive fixes this too.

Two dismissals are held back deliberately:

- **Pointer dismissal is off.** The marks live nowhere but this session until
  "Attach marks", so a press on the backdrop would discard all of them with no
  undo.
- **Escape stays with `KeyboardShortcuts`**, which backs out one layer at a
  time — the open note, then the selection, then the session. Letting the
  primitive close on the same key collapses all three into the first press;
  removing the `escape-key` guard fails the existing note-escape regression,
  so it is load-bearing rather than defensive.

## Verification

- `vitest run` on `chat-attachment-annotation` (11), `chat-attachment-composer`,
  `chat-attachment-message-previews`, `chat-attachment-artifact-viewer` — 33 passed.
  The new case presses the backdrop and asserts the editor and its mark survive;
  it fails if `disablePointerDismissal` is dropped, which is what makes the
  no-discard property a guarded contract rather than a default
- `tsc` on the production and test projects, `oxlint` (incl. type-aware),
  `eslint`, the style config, and `lint:localized-ui`
- End to end on this PR's preview: attach an image, open the preview, Edit,
  draw and select a box, backdrop-click (session survives), Escape (deselects),
  Escape again (closes), Attach marks, then download the flattened PNG and
  confirm the pin lands inside the corner there too

### DIFF
--- a/turbo/apps/platform/src/signals/okou-page/flatten-annotated-image.ts
+++ b/turbo/apps/platform/src/signals/okou-page/flatten-annotated-image.ts
@@ -9,6 +9,8 @@ import {
   markOrdinal,
   NOTE_GROUND,
   noteOnImage,
+  PIN_INSET_PX,
+  PIN_RADIUS_PX,
   REDACT_FILL,
   STROKE_HALO_INNER,
 } from "./image-annotation.ts";
@@ -23,7 +25,6 @@ import { createDeferredPromise, withCleanup } from "../utils.ts";
 const REFERENCE_EDGE_PX = 1000;
 const STROKE_WIDTH_UNITS = 3;
 const HALO_WIDTH_UNITS = 1;
-const PIN_RADIUS_UNITS = 11;
 const PIN_FONT_UNITS = 13;
 const TEXT_FONT_UNITS = 18;
 const NOTE_FONT_UNITS = 15;
@@ -208,7 +209,7 @@ function drawPin(
 ): void {
   const { x, y } = at;
   const { ink, ordinal } = pin;
-  const radius = px(scale, PIN_RADIUS_UNITS);
+  const radius = px(scale, PIN_RADIUS_PX);
   context.beginPath();
   context.arc(x, y, radius, 0, Math.PI * 2);
   context.fillStyle = ink;
@@ -348,7 +349,14 @@ function drawMark(
         px(scale, CORNER_RADIUS_UNITS),
       );
     });
-    drawPin(context, scale, { x, y }, { ink: mark.ink, ordinal });
+    // Inset to match the editor, where the corner belongs to the resize grip.
+    const inset = px(scale, PIN_INSET_PX);
+    drawPin(
+      context,
+      scale,
+      { x: x + inset, y: y + inset },
+      { ink: mark.ink, ordinal },
+    );
     return;
   }
 
--- a/turbo/apps/platform/src/signals/okou-page/image-annotation.ts
+++ b/turbo/apps/platform/src/signals/okou-page/image-annotation.ts
@@ -51,6 +51,21 @@ export const STROKE_HALO_INNER = "rgba(255, 255, 255, 0.90)";
  */
 export const NOTE_GROUND = "rgba(255, 255, 255, 0.94)";
 
+/** Radius of the disc a box's ordinal is printed in. */
+export const PIN_RADIUS_PX = 11;
+
+/**
+ * How far the pin's centre sits inside the box's own top-left corner.
+ *
+ * The pin used to be centred *on* that corner, which is precisely where the
+ * selection grip lands — so selecting a box hid its number under the white dot
+ * and the note that says "1" had nothing left to point at. Tucking the pin
+ * inside clears the grip by a whole radius, and unlike a pin parked outside the
+ * box it survives a mark drawn against the top or left edge of the image, which
+ * is what matters in the flattened copy the model actually reads.
+ */
+export const PIN_INSET_PX = PIN_RADIUS_PX + 3;
+
 /**
  * `highlight`, `crop` and `redact` were dropped, and `select` with them: a
  * mark is clicked directly, so a mode for "not drawing" has nothing left to do,
--- a/turbo/apps/platform/src/views/okou-page/__tests__/chat-attachment-annotation.test.tsx
+++ b/turbo/apps/platform/src/views/okou-page/__tests__/chat-attachment-annotation.test.tsx
@@ -86,6 +86,35 @@ test("A selected annotation mark can be resized directly", async () => {
   expect(screen.getByTestId("annotation-handle-br")).toBeVisible();
 });
 
+test("Pressing outside the editor keeps the session and every mark drawn in it", async () => {
+  // The editor used to be an inert overlay, so there was no dismissal to lose.
+  // It now rides on a dialog that closes on an outside press by default, and
+  // closing discards the whole session: marks live nowhere but here until
+  // "Attach marks", and there is no undo across that boundary.
+  const user = userEvent.setup({ pointerEventsCheck: 0 });
+  const image = draftAttachment("stray-press.png");
+  mockAttachmentChat(context, { draft: draftForAttachment(image, "") });
+
+  await setupPage({
+    context,
+    path: `/chats/${ATTACHMENT_THREAD_ID}`,
+    featureSwitches: { [FeatureSwitchKey.ComposerImageAnnotation]: true },
+  });
+
+  const surface = await openAnnotationEditor("stray-press.png");
+  drawBox(surface);
+  await screen.findByTestId("annotation-mark-1");
+
+  const backdrop = document.querySelector('[data-slot="dialog-overlay"]');
+  if (!backdrop) {
+    throw new Error("Expected the annotation editor's dialog backdrop");
+  }
+  await user.click(backdrop);
+
+  expect(screen.getByTestId("image-annotation-editor")).toBeVisible();
+  expect(screen.getByTestId("annotation-mark-1")).toBeVisible();
+});
+
 test("A user can click an annotation note, edit it, and close only the note", async () => {
   const user = userEvent.setup();
   const image = draftAttachment("annotated-plan.png", {
--- a/turbo/apps/platform/src/views/okou-page/image-annotation-editor.tsx
+++ b/turbo/apps/platform/src/views/okou-page/image-annotation-editor.tsx
@@ -1211,6 +1211,16 @@ function useGrabEndpoint(
   };
 }
 
+/**
+ * Makes the scrolling stage a size query container, so the image can be bounded
+ * by the box it is actually in.
+ *
+ * The fit used to be written as fractions of the *viewport* — `min(880px, 88vw)`
+ * by `min(520px, 62vh)` — which is a guess at the stage that stops being true
+ * the moment anything around it changes size. `100cqw`/`100cqh` are that box.
+ */
+const STAGE_QUERY_CONTAINER = { containerType: "size" } as const;
+
 function EditorStage({
   filename,
   signals,
@@ -1289,7 +1299,10 @@ function EditorStage({
 
   return (
     <div className="relative flex min-h-0 flex-1 flex-col overflow-hidden bg-muted/30">
-      <div className="flex min-h-0 flex-1 items-center justify-center overflow-auto p-5">
+      <div
+        className="flex min-h-0 flex-1 items-center justify-center overflow-auto p-5"
+        style={STAGE_QUERY_CONTAINER}
+      >
         <div
           ref={bindSurface}
           onPointerDown={handlers.onPointerDown}
@@ -1308,8 +1321,8 @@ function EditorStage({
             // The fit bounds are the stage's own box, so 100% zoom shows the
             // whole image and zooming grows the layout box the stage scrolls.
             style={{
-              maxWidth: `calc(min(880px, 88vw) * ${zoom})`,
-              maxHeight: `calc(min(520px, 62vh) * ${zoom})`,
+              maxWidth: `calc(100cqw * ${zoom})`,
+              maxHeight: `calc(100cqh * ${zoom})`,
             }}
             className="block rounded-lg object-contain"
           />
@@ -1385,7 +1398,11 @@ function AnnotationSurface({
       >
         <KeyboardShortcuts signals={signals} />
         <div
-          className="flex h-[min(700px,90vh)] w-[min(980px,94vw)] min-h-0 flex-col overflow-hidden rounded-xl bg-background text-foreground shadow-[0_24px_70px_hsl(var(--overlay)/0.30)]"
+          // The same envelope the preview dialog uses (`DialogContent`
+          // maxWidth={1440} height={1000}, 1.5rem of viewport padding). Edit
+          // replaces that window in place, so anything narrower reads as the
+          // picture shrinking the moment the pencil is pressed.
+          className="flex h-[1000px] max-h-full w-full max-w-[1440px] min-h-0 flex-col overflow-hidden rounded-xl bg-background text-foreground shadow-[0_24px_70px_hsl(var(--overlay)/0.30)]"
           data-testid="image-annotation-panel"
         >
           <EditorHeader filename={target.filename} signals={signals} />
--- a/turbo/apps/platform/src/views/okou-page/image-annotation-editor.tsx
+++ b/turbo/apps/platform/src/views/okou-page/image-annotation-editor.tsx
@@ -14,6 +14,7 @@ import {
   X,
 } from "lucide-react";
 import { Button } from "@okouai/ui/components/ui/button";
+import { Dialog, DialogContent } from "@okouai/ui/components/ui/dialog";
 import { Input } from "@okouai/ui/components/ui/input";
 import {
   Tooltip,
@@ -1377,6 +1378,18 @@ export function ImageAnnotationEditor({
   return <AnnotationSurface signals={signals} target={target} />;
 }
 
+/**
+ * The editor is the same window as the preview it replaces.
+ *
+ * It used to be a hand-rolled `fixed z-50` overlay, which lost twice. Its size
+ * was a second set of numbers that had to be kept level with the preview's by
+ * hand, and it drifted — `min(980px, 94vw)` against the preview's 1440, so the
+ * picture shrank the moment the pencil was pressed. And `z-50` only orders it
+ * inside the app's own stacking context, so anything portalled to `body` after
+ * it — the composer's slash-command menu, left open behind the preview — kept
+ * painting on top of the editor and stayed clickable. Both are the dialog
+ * primitive's job, and the preview next door was already using it.
+ */
 function AnnotationSurface({
   signals,
   target,
@@ -1384,6 +1397,8 @@ function AnnotationSurface({
   readonly signals: ImageAnnotationSignals;
   readonly target: AnnotationTarget;
 }) {
+  const { t } = useTranslation();
+  const close = useSet(signals.closeAnnotationEditor$);
   const resolvedUrl = useResolvedAttachmentUrl(target.url);
 
   if (resolvedUrl === null) {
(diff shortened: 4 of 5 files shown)


tongx-blip:
Changes Requested


PR_AUTO_REVIEW_COMMIT: c414bb97e92e706b1f2b1bef3614140bda290eac

Inspected HEAD: `c414bb97e92e706b1f2b1bef3614140bda290eac`. Practice revision: `bc0a009965fb966f60b78902003984751fb76f6a`.

### Summary

Moves the composer's image annotation editor off a hand-rolled `fixed inset-0 z-50` overlay and onto the shared `Dialog`/`DialogContent` at `maxWidth={1440} height={1000}`, which is the envelope the artifact preview it replaces already uses; the stage becomes a size query container so the image fits `100cqw`/`100cqh` instead of hard-coded viewport fractions. Also moves a box mark's ordinal pin from the box's top-left corner — where the selection grip lands and hid it — to a radius inside that corner, in the editor and in the flattened copy, from one shared pair of constants.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | Entry is `ArtifactPreviewDialogActions` → `annotationTarget.open()` + `closeLightboxImmediately()` (`attachment-chips.tsx`), so the preview dialog is already closed and the editor is not a nested modal. Consumers of `PIN_RADIUS_PX`/`PIN_INSET_PX` are exactly the two renderers, `image-annotation-marks.tsx` and `flatten-annotated-image.ts`. |
| `REVIEW.md` | Check ordering, cancellation, and failure outcomes where relevant | Pass | No async introduced. `closeAnnotationEditor$` is a synchronous command, so the new `onOpenChange` handler needs no `detach`. `KeyboardShortcuts`' capture-phase `document` listener and its ref-cleanup return are untouched. |
| `REVIEW.md` | Require regression coverage for a changed, replaced, or rerouted flow | **Finding (P1)** | The whole editor flow is rerouted onto a primitive that dismisses on outside press by default; `disablePointerDismissal` is the only thing preventing silent loss of every unsaved mark, and nothing covers it. See finding 1. |
| `REVIEW.md` | Report concrete practice violations with rule and location; do not infer defects from length or preference | Pass | The one finding below is a missing-coverage claim tied to a specific default (`DialogRootProps.disablePointerDismissal`, `@default false`), not a stylistic objection. |
| `REVIEW.md` | Persisted data / API contract / response validation review | N/A | No schema, contract, migration, or `jsonb(...).$type<...>()` change. `PIN_INSET_PX` is a render-time constant; `ImageAnnotationMark` still persists rects and inks only, so existing drafts render under the new offset with no stored-data change. |
| `docs/fallback.md` | Every new fallback declared in a `Fallbacks` PR-body section | N/A | No fallback introduced. No `??`/`\|\|` chain, default, or tolerance branch is added; `details.reason !== "escape-key"` routes a live, reachable event rather than guarding an impossible state — removing it makes the existing note-escape test fail, which is recorded in the PR body. |
| `docs/fallback.md` | A pre-GA feature switch needs no compatibility code | Pass | `composerImageAnnotation` is `enabled: false` with no org/email hashes (`feature-switch.ts`), so no cutover compatibility is owed, and none is added. |
| `docs/fallback.md` | Removing a path requires evidence, and its branch and tests go together | Pass | The removed overlay branch had no tests of its own; the two behaviours it did own (Escape layering, no dismissal on outside press) are re-established explicitly rather than dropped. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | Diff adds no `any`, assertion, or suppression comment; `lint-eslint`, `lint-types`, `lint-type-app` are green on this HEAD. |
| `docs/bad-smell.md` | Static imports in production code | Pass | `Dialog`, `DialogContent` are static imports from `@okouai/ui/components/ui/dialog`. |
| `docs/bad-smell.md` | Catch/handle only at a boundary that handles it; no fabricated defaults | Pass | No new error handling or defaulting. |
| `docs/styles.md` | Business UI composes Tailwind utilities; no first-party selectors, CSS modules, or injected styles | Pass | Only utility classes and inline geometry. The one legacy class, `okou-app`, moves from the deleted overlay's `className` to `contentClassName` in the same file — a relocated dependency, not a new one; the required `lint-style` job is green on this HEAD. |
| `docs/styles.md` | `DialogContent` owns viewport and popup; callers select `maxWidth`/`height`/`mode`; no `className`/`style`/`render` on the popup | Pass | Caller passes `maxWidth={1440} height={1000} surface="canvas"` — the documented artifact-preview pair — plus `contentClassName` and `showCloseButton={false}`, and sets no popup `className`/`style`/`render`. This deletes the duplicated `h-[1000px] max-h-full w-full max-w-[1440px]` classes the previous commit had hand-copied. |
| `docs/styles.md` | Business code imports the shared dialog, not Base UI primitives | Pass | Imports `@okouai/ui/components/ui/dialog`; no `@base-ui/react/dialog` import. |
| `docs/styles.md` | Footer actions stay reachable by scrolling | Pass | The shared inner container still appends `overflow-y-auto!` after `contentClassName`, matching how `ArtifactPreviewDialogContent` passes the same `overflow-hidden` layout classes. |
| `docs/styles.md` | Components must not introduce local CSS variables as an alternate token registry | Pass | `containerType: "size"` plus `calc(100cqw * zoom)` are computed layout values on the element, not a parallel token registry; no custom property is declared. |
| `docs/testing.md` / `docs/testing/app-testing.md` | Page-level integration coverage through the real Router for new behavior | **Finding (P1)** | `chat-attachment-annotation.test.tsx` enters through `setupPage` and covers open, draw, note-Escape, and Attach marks, but adds nothing for the dismissal contract this PR newly depends on. See finding 1. |
| `docs/testing/testing-external-behavior.md` | Assert externally observable page state, not CSS classes or component state | Pass | Correctly, no test asserts the pin offset or the dialog's pixel size — those are CSS geometry, explicitly outside the page contract. The two size/pin changes are verified by rendering, not by pinning class strings. |
| `docs/testing.md` | No unrelated suites; one Vitest process | Pass | PR body records scoped runs of the four attachment/annotation suites (32 passed) plus scoped `tsc`/`oxlint`/`eslint`/`lint:localized-ui`. |
| `.claude/skills/ccstate/SKILL.md` + `references/react.md` | Preserve stable callback refs and `onRef` cleanup; no store mutation during render; `detach()` only at a real boundary | Pass | `KeyboardShortcuts`' `ref` closure and its `cleanup` return are unchanged. New hooks (`useTranslation`, `useSet`) are declared above the existing `resolvedUrl === null` early return, so hook order is stable. The `onOpenChange` handler calls a synchronous command, so `detach()` would be wrong here and is correctly absent. |

### Findings

**P1 — No regression coverage for outside-press dismissal, whose failure mode is silent loss of every unsaved mark**

`turbo/apps/platform/src/views/okou-page/image-annotation-editor.tsx:1412-1425`

The previous container was an inert `div`: it had no dismissal behaviour to lose, so "clicking the backdrop does not throw the session away" was structurally true and needed no test. `Dialog.Root` inverts that default — `DialogRootProps.disablePointerDismissal` is documented `@default false` — so after this change the property is carried by one prop on one line.

Trigger: delete or invert `disablePointerDismissal`, or let a future refactor pass the props through a wrapper that drops it. Consequence: a stray press anywhere on the backdrop closes the editor and `closeAnnotationEditor$` clears `session$`, discarding every mark drawn in the session. There is no undo across that boundary and no confirmation, and the marks exist nowhere else until "Attach marks" writes them to the draft. The whole test file passes either way.

The neighbouring Escape contract is protected exactly as it should be — removing the `escape-key` guard fails "A user can edit a mark's note from the image" — which is what makes the gap on the adjacent, more destructive dismissal path stand out rather than read as a deliberate omission.

Add a page-level case to `chat-attachment-annotation.test.tsx` that opens the editor, draws a box, presses the backdrop outside the panel, and asserts the editor is still present with its mark. That is externally observable page state, so it stays inside the boundary `docs/testing/testing-external-behavior.md` draws — unlike the pin offset and dialog size, which correctly have no test.

### Testing

Existing coverage in `chat-attachment-annotation.test.tsx` (10 tests) enters through `setupPage`, and it does real work here: it is what proves the `escape-key` guard is load-bearing rather than defensive, since removing that guard turns the first Escape into a full session discard and fails the note test. The three other suites named in the PR body (`chat-attachment-composer`, `chat-attachment-message-previews`, `chat-attachment-artifact-viewer`, 32 passed together) cover the surrounding chip, preview, and viewer paths that the mark renderer is shared with.

The visual halves are appropriately untested and verified by rendering instead: the PR body records measured panel/image sizes on the PR preview (1440×1000 / 1357×848 against the preview's 1440×1000 / 1371×896) and a downloaded flattened PNG confirming the canvas pin lands at the same inset as the DOM pin. Pinning either as a class-string or style assertion would be the anti-pattern `testing-external-behavior.md` names.

The one gap is the dismissal case in the finding above.

CI on this HEAD: `lint-style`, `lint-eslint`, `lint-knip`, `lint-types`, `lint-type-app`, and all eight `test-app` shards are green. `test-api (6)` failed on `chat-search-reader.test.ts` with `Test timed out in 5000ms` — this PR changes four Platform files and no API code, so that is unrelated to the diff and should be rerun rather than repaired here.

tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: 9d490c5b308b4aa94336d3604656fcafe002f14e
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `9d490c5b308b4aa94336d3604656fcafe002f14e`. Practice revision: `bc0a009965fb966f60b78902003984751fb76f6a`. Delta from the previously reviewed `c414bb9`: one added test case, no production change.

### Summary

Moves the composer's image annotation editor off a hand-rolled `fixed inset-0 z-50` overlay onto the shared `Dialog`/`DialogContent` at `maxWidth={1440} height={1000}` — the envelope the artifact preview it replaces already uses — and makes the stage a size query container so the image fits `100cqw`/`100cqh` instead of hard-coded viewport fractions. Also moves a box mark's ordinal pin from the box's top-left corner, where the selection grip landed on top of it, to a radius inside that corner, in the editor and the flattened copy, from one shared pair of constants. `9d490c5` adds the outside-press regression case raised as P1 against the previous HEAD.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | Entry is `ArtifactPreviewDialogActions` → `annotationTarget.open()` + `closeLightboxImmediately()` (`attachment-chips.tsx`), so the preview dialog is already closed and the editor is never a nested modal. `PIN_RADIUS_PX`/`PIN_INSET_PX` have exactly two consumers, `image-annotation-marks.tsx` and `flatten-annotated-image.ts`, so the picked and sent images cannot drift apart. |
| `REVIEW.md` | Check ordering, cancellation, and failure outcomes where relevant | Pass | No async introduced. `closeAnnotationEditor$` is a synchronous command, so the new `onOpenChange` handler needs no `detach`. `KeyboardShortcuts`' capture-phase `document` listener and its ref-cleanup return are untouched. |
| `REVIEW.md` | Require regression coverage for a changed, replaced, or rerouted flow | Pass | Both behaviours the deleted overlay implicitly owned are now guarded at the page level: Escape layering by "A user can click an annotation note, edit it, and close only the note", and outside-press by the new "Pressing outside the editor keeps the session and every mark drawn in it". Each fails when its prop is removed, so neither is a silent default. |
| `REVIEW.md` | Report concrete practice violations with rule and location | Pass | The P1 raised against `c414bb9` is resolved by `9d490c5`; nothing further found. |
| `REVIEW.md` | Persisted data / API contract / response validation review | N/A | No schema, contract, migration, or `jsonb(...).$type<...>()` change. `PIN_INSET_PX` is a render-time constant; `ImageAnnotationMark` still persists rects and inks only, so existing drafts render under the new offset with no stored-data change. |
| `docs/fallback.md` | Every new fallback declared in a `Fallbacks` PR-body section | N/A | No fallback introduced. No `??`/`\|\|` chain, default, or tolerance branch is added; `details.reason !== "escape-key"` routes a live, reachable event rather than guarding an impossible state — removing it fails the note-escape test, and the PR body records that. |
| `docs/fallback.md` | A pre-GA feature switch needs no compatibility code | Pass | `composerImageAnnotation` is `enabled: false` with no org/email hashes (`feature-switch.ts`), so no cutover compatibility is owed and none is added. |
| `docs/fallback.md` | Removing a path requires evidence, and its branch and tests go together | Pass | The removed overlay branch had no tests of its own, and the two behaviours it did own are re-established explicitly with coverage rather than dropped. |
| `docs/fallback.md` | No negative tests that only pin retired behavior | Pass | The new case asserts a live product behaviour — the editor and its mark are still present — not that the old overlay is gone. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | No `any`, assertion, or suppression comment. The added test resolves its `querySelector` result by throwing on `null` rather than asserting the type, so narrowing is real. |
| `docs/bad-smell.md` | Static imports in production code and test utilities | Pass | `Dialog`, `DialogContent` are static imports from `@okouai/ui/components/ui/dialog`; the test adds no import. |
| `docs/bad-smell.md` | Catch/handle only at a boundary that handles it; no fabricated defaults | Pass | No new error handling or defaulting in production code. |
| `docs/styles.md` | Business UI composes Tailwind utilities; no first-party selectors, CSS modules, or injected styles | Pass | Only utility classes and inline geometry. The one legacy class, `okou-app`, moves from the deleted overlay's `className` to `contentClassName` in the same file — a relocated dependency, not a new one; the required `lint-style` job was green on the identical production diff at `c414bb9`. |
| `docs/styles.md` | `DialogContent` owns viewport and popup; callers select `maxWidth`/`height`/`mode`; no `className`/`style`/`render` on the popup | Pass | Caller passes `maxWidth={1440} height={1000} surface="canvas"` — the documented artifact-preview pair — plus `contentClassName` and `showCloseButton={false}`, and sets no popup `className`/`style`/`render`. This deletes the duplicated `h-[1000px] max-h-full w-full max-w-[1440px]` classes the first commit had hand-copied. |
| `docs/styles.md` | Business code imports the shared dialog, not Base UI primitives | Pass | Imports `@okouai/ui/components/ui/dialog`; no `@base-ui/react/dialog` import. |
| `docs/styles.md` | Preserve Base UI's focus, nested portal, outside-press, and animation ownership | Pass | Dismissal is adjusted through Base UI's own `Dialog.Root` prop (`disablePointerDismissal`) and its `onOpenChange` reason, not by intercepting events or hiding the control with CSS; focus, portal, and animation ownership are untouched. |
| `docs/styles.md` | Footer actions stay reachable by scrolling | Pass | The shared inner container still appends `overflow-y-auto!` after `contentClassName`, matching how `ArtifactPreviewDialogContent` passes the same `overflow-hidden` layout classes. |
| `docs/styles.md` | Components must not introduce local CSS variables as an alternate token registry | Pass | `containerType: "size"` plus `calc(100cqw * zoom)` are computed layout values on the element; no custom property is declared. |
| `docs/testing.md` / `docs/testing/app-testing.md` | Page-level integration coverage through the real Router for new behavior | Pass | The new case follows the canonical order — fixtures, `setupPage`, open the editor, draw, one user action, assert the result — and lives in the existing `views/**/__tests__` page suite rather than a component-only unit test. |
| `docs/testing/testing-external-behavior.md` | Assert externally observable page state, not CSS classes or component state | Pass | The new assertions are page presence of the editor and its mark. It locates the backdrop through `data-slot="dialog-overlay"`, a documented component slot, which the styles guide names as the supported hook. Correctly, nothing asserts the pin offset or the dialog's pixel size — those are CSS geometry and outside the page contract. |
| `docs/testing.md` | Avoid fake timers; wait for the observable result | Pass | Uses `findByTestId` for the drawn mark and `userEvent`; no timer control. |
| `docs/testing.md` | Use unique identities; centralized `testContext()` and mocks | Pass | Reuses the file's `testContext()` and `mockAttachmentChat`, with a distinct `stray-press.png` fixture rather than sharing another case's filename. |
| `docs/testing.md` | No unrelated suites; one Vitest process | Pass | Scoped runs only: the annotation suite (11 passed) plus the three neighbouring attachment suites, and scoped `tsc`/`eslint`/`oxlint`. |
| `.claude/skills/ccstate/SKILL.md` + `references/react.md` | Preserve stable callback refs and `onRef` cleanup; no store mutation during render; `detach()` only at a real boundary | Pass | `KeyboardShortcuts`' `ref` closure and `cleanup` return are unchanged. New hooks (`useTranslation`, `useSet`) sit above the existing `resolvedUrl === null` early return, so hook order is stable. `onOpenChange` calls a synchronous command, so `detach()` would be wrong and is correctly absent. |

### Findings

None. The P1 raised against `c414bb9` — no regression coverage for outside-press dismissal, whose failure mode was silent loss of every unsaved mark — is resolved.

Two non-blocking observations, neither requiring a change here:

- The tool pill and zoom controls still float over the image, so a taller stage means they overlay a taller picture. This is unchanged behaviour and matches how the preview floats its own zoom controls.
- The canvas pin's centre is offset from the rect origin while the DOM pin's is offset from the padding box, so the two differ by the 2.5px border at reference scale. On a 22px disc that is not observable, and the flattened output was checked against the editor directly.

### Testing

`chat-attachment-annotation.test.tsx` is now 11 page-level cases entering through `setupPage`, and the two that matter here are adversarial rather than descriptive: removing the `escape-key` guard turns the first Escape into a full session discard and fails the note case, and removing `disablePointerDismissal` closes the editor on a backdrop press and fails the new case. Both were confirmed by actually removing the prop and observing the failure, which is what distinguishes these from tests that would pass either way.

The three neighbouring suites (`chat-attachment-composer`, `chat-attachment-message-previews`, `chat-attachment-artifact-viewer`) cover the chip, preview, and viewer paths the shared mark renderer feeds, 33 passed with the annotation suite.

The visual halves are deliberately untested and verified by rendering instead, which is the right boundary: the PR body records measured panel/image sizes on the PR preview (1440×1000 / 1357×848 against the preview's own 1440×1000 / 1371×896) and a downloaded flattened PNG confirming the canvas pin lands at the same inset as the DOM pin. Pinning either as a class-string or style assertion would be the anti-pattern `testing-external-behavior.md` names.

CI on the identical production diff at `c414bb9`: `lint-style`, `lint-eslint`, `lint-knip`, `lint-types`, `lint-type-app`, and all eight `test-app` shards green. `test-api (6)` failed there on `chat-search-reader.test.ts` with `Test timed out in 5000ms`; this PR touches no API code, so that is unrelated to the diff and is being rerun rather than repaired.

Issue #32967, opened 2026-09-09
refactor(db): drop computer_use_hosts.client_product and its check constraint


lancy:
## Parent

Contract step of the retirement of `computer_use_hosts.client_product`. The expand step **#32966** is merged, controller accepted, and released to production (carried by release PRs #33332 and #33338, both with successful `promote-api-production`). Related to #26368 and the closed EPIC #32157.

**This issue was rewritten on 2026-09-11.** Its original gate reasoned only about `INSERT`/`UPDATE` and assumed #32966 had removed the deployed API's references to the column. Measurement disproved that. The plan below replaces it. Refreshed against `main` at `75095c26e72eb7deb40f5c15b920aa0764d94ceb`.

## Why this is a one-release drop with an accepted window

Drizzle builds column lists from the **table declaration**, not from usage. Measured with `toSQL()` against the repository's pinned `drizzle-orm@0.45.2`, using a declaration matching `computer_use_hosts` as it stands after #32966:

```
bare select()        -> select "id", ..., "client_product", ... from "computer_use_hosts"
bare returning()     -> update ... returning "id", ..., "client_product", ...
insert, value OMITTED-> insert into "computer_use_hosts" (..., "client_product", ...) values (..., default, ...)
select({ id })       -> select "id" from "computer_use_hosts"
```

So while the declaration exists, the deployed API names the column in every insert, every bare select, and every bare returning — regardless of whether any code reads or writes the value. #32966 could not and did not reduce this.

Removing the declaration is therefore the only way to stop the name being generated, and `scripts/test-migration-consistency-schema.ts` compares the migrated schema against one regenerated from the declarations, so the declaration removal and the physical `DROP COLUMN` must ship together. That is a deliberate, accepted deviation from the two-release pattern in `docs/deployment-compatibility.md`; the alternatives were evaluated and rejected on the issue thread.

### The accepted window

Migrations run before API promotion. Between the migration completing and the previous API build draining, statements from that build reference a column that no longer exists and receive `42703`.

Affected paths: host registration (`hosts/start`), `hostFromToken` — which backs `heartbeat`, `host/stop`, and host-command claiming.

[...]

PR #32573, opened 2026-09-08
feat(billing): add free member usage packs

hulh122:
## Summary

Members can use **Free** without purchasing a usage pack. Active Free workspaces can invite members, and paid-to-Free package changes take effect at the next subscription period while existing credits keep their expiry.

Use the persisted `showUsagePack` capability from #32595 for package visibility and optional invitation purchases. Retire the separate `memberInvitationAllowed` capability: invitation routes and the App now use the existing entitlement `active` / `suspended` status, while retaining administrator authorization. Per-organization invitation overrides no longer apply.

| Workspace | Invitation behavior |
| --- | --- |
| Active new Pro / Team, including Atom grants (`showUsagePack: true`) | Choose Free or a paid package. Free sends the invitation directly; paid packages keep the existing payment review. An Atom account without a package subscription uses the first-time configuration flow before purchasing. |
| Active old Pro / Team and Free (`showUsagePack: false`) | Send invitations directly, with no package selector or purchase step. |
| Suspended | Block direct invitations and paid invitation preview/confirmation. The App explains that the workspace plan must be reactivated. |

- Represent Free as `usagePackUsd: 0`, with no paid allocation and no new Stripe Price. Keep paid packages and their purchase flow available.
- Default new Pro/Team checkout, every Pro/Team tier change, and package-enabled invitations to the $20 package. Checkout and tier changes require at least one paid member package; other members may select Free.
- Keep the plan-card floor based on paid packages, so Pro starts at $20 and Team starts at $180 even when Free is available.
- Schedule paid-to-Free changes at period end and allow restoration. Support checkout, legacy-plan migration, renewal, and later paid additions when every member is Free; retain the workspace's Pro or Team base subscription.
- Remove invitation eligibility from tier defaults, business capabilities, entitlement writers, and App capability fallbacks. Expose the existing normalized entitlement `status` in billing responses.
- Migration `1097_retire_member_invitation_capability` follows main's `1096_shiny_black_bolt`. It replaces the outgoing API's invitation-column trigger with a status-derived compatibility value and backfills all sources, including manual overrides.

## Fallbacks

[...]

Issue #26368, opened 2026-08-11
chore: retire Zero Computer Use migration compatibility


lancy:
## Current controller status — 2026-09-14

The [user-approved 72-hour decision](https://github.com/vm0-ai/vm0/issues/26368#issuecomment-5662603696) supersedes the earlier 14-day activity requirement **for the narrow telemetry slice**. It does not authorize deletion of every Zero dependency.

- **DB retirement completed:** #31958/#31961 removed the default; #32966/#32987 removed application reads/writes; #32967/#33390 completed the physical column/constraint drop. The drop is independently [released and production verified](https://github.com/vm0-ai/vm0/issues/32967#issuecomment-5629454199).
- **Earlier scoped cleanup completed:** source/build #32496/#32506, obsolete feed #31475/#31500 and auth schemes #33613/#33615 are closed; do not recreate their work or restore the old auth acceptance requirement below. The [September 14 inventory](https://github.com/vm0-ai/vm0/issues/26368#issuecomment-5662424788) records their distinct acceptance/release evidence.
- **Remaining implicit telemetry fallback completed at code acceptance:** #34086/#34089 is **MERGED and independently controller ACCEPTED**, merge `5684b1395f620405390095cf878086f1428e495d`. [Exact-code, seven-criterion acceptance](https://github.com/vm0-ai/vm0/issues/34086#issuecomment-5663308047) verifies the decoder, request/error metadata, HTTP invariants, final merge-group and preserved boundaries. Missing/unrecognized product is omitted, never guessed as Zero or Okou. Explicit product identity is preserved.
- **Publication remains separate:** this controller performed no production release or production runtime verification for #34089. Merge/code acceptance does not establish its current production publication.
- **KEEP:** installed-client `hard` migration policy, trusted neutral Okou download, explicit Zero telemetry and historical identity/artifacts. The approved preflight still observed Zero policy polls. Future drain queries must count unclassified Desktop traffic separately.
- **Monitoring disposition:** the September 14 one-shot and #34086 completion workflow/automation were deleted, with absence verified. Owner chat and GitHub evidence remain; no replacement timer or further cleanup owner was created.

#26368 remains open for independently gated retained installed-client surfaces and any later authorized publication/production acceptance. No broader retirement or production operation is implied.

The dated execution notes below are preserved as history; the current disposition above and linked decision supersede their old timer, auth and 14-day assumptions.

## Historical execution split — 2026-09-08

[...]

Issue #32949, opened 2026-09-09
refactor(platform): remove the Desktop product badge from Computer Use host rows


lancy:
## Parent

Related to #26368 (Zero Computer Use compatibility retirement) and the closed EPIC #32157. Refreshed just in time against current `main` at `ed4e5193b7` on 2026-09-09.

## Problem

The Computer Use host picker renders a product badge after every host name:

- `turbo/apps/platform/src/views/okou-page/chat-composer.tsx:7373`
- `turbo/apps/platform/src/views/computer-use-authorization/computer-use-authorization-page.tsx:78`

Both call `desktopProductDisplayName(host.product)` (`turbo/apps/platform/src/i18n/desktop-product.ts`), which returns the bare literal `"Zero"` or `"Okou"`. Its only purpose was to disambiguate a Zero Desktop host from an Okou Desktop host while both applications could register hosts.

Zero Desktop is hard-stopped and can no longer be used, so the badge no longer distinguishes anything a user can act on. Remove it.

### It is also visibly broken today

The badge is nested inside the same `truncate` span as the host name:

```tsx
<span className="block truncate text-sm text-foreground">
  <span>{host.displayName}</span>
  <span className="ml-1 text-xs text-muted-foreground">
    {" "}
    {desktopProductDisplayName(host.product)}
  </span>
</span>
```

With a realistic host name such as `Lancys-MacBook-Pro.local` the badge truncates to `Ok…`, conveying nothing while still consuming width and pushing the row's toggle. The authorization page repeats the same nesting under `truncate text-sm font-medium`. Removing the badge resolves this; no separate layout fix is needed. (The markup also carries both `{" "}` and `ml-1`, a double space, which disappears with it.)

## Production evidence, 2026-09-09

- In `vm0-request-log-prod` over the trailing 3 days, every `x_client_product == "zero"` request is `GET /api/desktop/migration-policy` from Desktop `0.38.126` — 498 requests, latest `2026-09-09T09:57:22Z`. That endpoint is the hard-stop policy gate. Zero clients receive the stop and never reach host registration or heartbeat, so no new Zero host row can be created.
- Desktop client versions do not overlap between products: Zero ends at `0.38.126`, Okou runs `0.44.3` and above.
[...]

PR #32987, opened 2026-09-09
refactor(db): stop reading and writing computer use host client product

### LINKED ISSUE #32966: refactor(db): stop reading and writing computer_use_hosts.client_product
## Parent

Expand step of the two-step retirement of `computer_use_hosts.client_product`. The contract step is the follow-up issue linked below and **must not ship in the same production release as this one** — see "Release gate". Related to #26368 and the closed EPIC #32157. Refreshed just in time against current `main` at `a8e237ede91c9d8c9543603121ed09178be1876f` on 2026-09-09.

## Problem

`computer_use_hosts.client_product` (`text NOT NULL`, plus the check constraint `client_product IN ('zero','okou')`) is the only Zero/Okou Desktop product distinction left in the live schema. After #32949 removed the two UI renderers, it no longer serves a product purpose, yet it is still written on every host registration and every heartbeat.

### Evidence, 2026-09-09

- Over the trailing 3 days in `vm0-request-log-prod`, `x_client_product == "zero"` Desktop traffic hits `/api/desktop/migration-policy` only: **0** calls to `/api/computer-use/hosts/start`, `/api/computer-use/heartbeat`, or `/api/computer-use/host/commands/*`. Over the same window `okou` recorded 274,155 heartbeats, 133,329 `commands/next`, and 153 `hosts/start`. No installed Zero build registers or heartbeats a host.
- `/api/desktop/migration-policy` returns the static constant `DESKTOP_ZERO_MIGRATION_POLICY = { schemaVersion: 1, mode: "hard" }` (`desktop-updates.ts:43-54`) and performs **no database access**, so it is fully decoupled from this column.
- Request-log product attribution comes from the `X-Client-Product` header via `clientHeaderLogFields` (`app-factory-core.ts:396-412`), **not** from this column. Dropping the column therefore does not weaken #26368's drain instrument.
- The column has no behavioral role. Host dedup keys on `(org_id, user_id, installation_id)`. The only non-serialization read is one term of a `publishChanged` dirty check (`computer-use.service.ts:1340`); since every live host reports `okou`, that term can never flip.
- `product` is already `optional()` in the API contract (`contracts/computer-use.ts:421`), so removing it from responses is backward compatible for every already-installed client. The CLI calls `client.list({})` and returns `result.body.hosts` without ever reading `.product` (`apps/cli/src/lib/api/domains/computer-use.ts`).
- `client_product` is not exposed in the MaskDB projection of `computer_use_hosts`, so per-row product counts cannot be read directly. Zero-era rows are identifiable by `app_version` instead: 111 of 120 rows predate `0.44`. This is a recorded visibility limit.

## Scope

Stop reading and stop writing the column, and relax it so the contract step can drop it safely.

1. **Migration** — next available number after `main`, semantic filename. `ALTER TABLE "computer_use_hosts" ALTER COLUMN "client_product" DROP NOT NULL;` Nothing else.
2. **Drizzle table** (`turbo/packages/db/src/schema/computer-use-host.ts`) — make `clientProduct` nullable to match. Keep the column and the check constraint declared; both are dropped by the follow-up issue.
3. **Stop writing** — remove `clientProduct` from the insert values and the `onConflictDoUpdate` set in `startComputerUseHost$` (`computer-use.service.ts:1248,1275`) and from the update set in `heartbeatComputerUseHost$` (`:1353`). Remove the now-unused `clientProduct` parameter from both command signatures (`:1221,1310`) and from their two route call sites (`routes/computer-use.ts:114,143`).
4. **Stop reading** — remove `product` from `serializeHost` (`:1012`) and remove the `lockedHost.clientProduct !== params.clientProduct` term from `publishChanged` (`:1340`). Every other `publishChanged` term stays.
5. **API contract** — remove `product` from `computerUseHostSchema` (`contracts/computer-use.ts:421`) and delete the now-unused local `desktopProductSchema` (`:13`) and its `DESKTOP_PRODUCTS` import (`:3`).
6. **Platform client** — remove `product` from `listedComputerUseHostSchema` (`shared-database/computed-key.ts:32`) and its `DESKTOP_PRODUCTS` import (`:1`), and remove the mapping at `signals/external/computer-use-hosts.ts:30`.

## Why the strict-schema hazard does not apply

`listedComputerUseHostSchema` is `.strict()` and is parsed with `.parse()` across the SharedWorker port (`computed-key.ts:93`, `message-port-client.ts:247`). That would be a cross-version hazard if an old worker could feed a new page. It cannot: the worker asset is content-hashed — `vite.config.ts:55` matches `assets/shared-database-worker-[^/]+\.js` and the module is aliased through `?sharedworker&url` (`:65-67`). A new bundle therefore requests a different worker script URL, and the producer and consumer of a computed value are always the same build. Do not add a schema-version handshake for this.

## Release gate

Migrations run **before** API promotion: `release-please.yml:1068-1069` builds the artifact, "run required production migrations, and deploy that exact artifact from the same runner only after migrations succeed."

That ordering makes this step safe: `DROP NOT NULL` lands first, while the still-live previous API keeps writing a value into a now-nullable column; then this artifact promotes and stops writing, leaving `NULL` on new rows.

**It also means the follow-up `DROP COLUMN` must ship in a strictly later production release than this one.** If both land in the same release, both migrations run while the pre-this-issue API is still live, and its insert would reference a dropped column. The follow-up issue records this as a hard gate; nothing in this issue may be used to justify collapsing the two.

## Explicit non-goals

- **Do not drop the column or the check constraint here.** That is the follow-up issue, gated on this one being released.
- Do not delete, revoke, or migrate any `computer_use_hosts` row. 12,627 `computer_use_commands` and 8,574 `computer_use_command_audit_events` reference Zero-era hosts through `NO ACTION` foreign keys; row deletion is out of scope and would break the audit trail.
- Do not touch `DESKTOP_PRODUCTS`, `DesktopProduct`, `desktopProductFromClientHeader`, or the `X-Client-Product` header in `contracts/client-headers.ts`. They remain in use by `desktop-updates.service.ts:82` (the retained DMG/update-feed product parameter) and by request-log attribution.
- Do not change `/api/desktop/migration-policy`, the DMG routes, the legacy auth callback, or anything else #26368 retains.
- Do not change host dedup, online/offline computation, selection, authorization, command dispatch, or realtime publication beyond removing the single dirty-check term.
- Do not edit an existing migration or snapshot, and do not hardcode a migration number — use the next available after `main`.
- No repository-wide `zero`/`vm0` substitution.
- No production release from the implementation thread.

## Open-PR overlap at refresh

#32926 and #32712 (bingjiez666) both add migrations, so migration numbers will collide. That is inventory, not a dependency: rebase current `origin/main`, renumber, and follow normal merge precedence. Do not wait.

## Acceptance criteria

- A single new migration contains only `ALTER COLUMN "client_product" DROP NOT NULL`, generated through `pnpm -F @okouai/db db:generate` and passing `pnpm -F @okouai/db test:migration-consistency`.
- No production code writes or reads `clientProduct`; the column and its check constraint remain declared in the Drizzle table, now nullable.
- `product` is absent from `computerUseHostSchema`, `listedComputerUseHostSchema`, `serializeHost`, and the platform host mapping. `DESKTOP_PRODUCTS` and `DesktopProduct` remain exported and still used by `desktop-updates.service.ts`.
- `publishChanged` retains every other term and its existing behavior.
- Host registration, heartbeat, dedup, online/offline status, and realtime publication are otherwise unchanged, with existing Computer Use tests passing and no assertion weakened or converted into a retired-behavior tombstone.
- The implementation records an explicit decision on whether a transition validator is warranted for this expand → contract cycle, per `turbo/packages/db/MIGRATIONS.md`, following the `1098_retire_member_invitation_capability` precedent.
- Focused Prettier, ESLint, TypeScript, migration consistency, and affected API/platform tests pass, followed by the normal protected PR pipeline.

lancy:
Closes #32966

Expand step of the two-step retirement of `computer_use_hosts.client_product`. Stops every read and write of the column and relaxes it to nullable so the contract step (#32967) can drop it safely. **The column and its check constraint stay declared here.**

## What changed

1. **Migration** — `1100_relax_computer_use_host_client_product.sql`, generated with `pnpm -F @okouai/db db:generate`, containing only:
   ```sql
   ALTER TABLE "computer_use_hosts" ALTER COLUMN "client_product" DROP NOT NULL;
   ```
2. **Drizzle table** — `clientProduct` is now nullable. The column and `computer_use_hosts_client_product_check` remain declared; #32967 drops both.
3. **Stop writing** — removed `clientProduct` from the insert values and `onConflictDoUpdate` set in `startComputerUseHost$`, from the update set in `heartbeatComputerUseHost$`, from both command signatures, and from the two route call sites.
4. **Stop reading** — removed `product` from `serializeHost` and the `lockedHost.clientProduct !== params.clientProduct` term from the heartbeat `publishChanged` dirty check. Every other `publishChanged` term is unchanged.
5. **API contract** — removed `product` from `computerUseHostSchema` plus the now-unused local `desktopProductSchema` and its `DESKTOP_PRODUCTS` import.
6. **Platform client** — removed `product` from `listedComputerUseHostSchema`, from the `external/computer-use-hosts.ts` mapping, and from the structurally-coupled `ComposerComputerUseHost` interface, along with the imports each drop made unused.

Consequential dead-code removals: the api-local `clientProduct$` header signal in `signals/context/hono.ts` (its only two consumers were the two route call sites above) and the `clientProduct` plumbing in the Computer Use BDD test helper.

`DESKTOP_PRODUCTS`, `DesktopProduct`, `desktopProductFromClientHeader`, and the `X-Client-Product` header are untouched. They remain in use by `desktop-updates.service.ts:82` and by request-log attribution in `app-factory-core.ts`, which is #26368's drain instrument.

## Transition validator decision: not warranted

Per `turbo/packages/db/MIGRATIONS.md`, a transition validator protects an expand → contract rollout while old and new application versions and data shapes may coexist. This cycle has no such coexistence hazard to assert:

- There is no dual-write, no dual-read, no backfill, no derived value, and no trigger. After this change nothing reads the column at all.
- The only cross-version behavior is that the still-live previous API writes `'okou'` into a now-nullable column while this artifact writes nothing. Both shapes are simply valid; there is no invariant a validator could fail on.
- The `product` response field was already `optional()` in the contract, so its removal is backward compatible for every installed client. The CLI never reads it.
- The one property that actually matters — `DROP COLUMN` must not ship in the same release — is a release-ordering gate, not something SQL can assert.

This differs from the `1098_retire_member_invitation_capability` precedent (#32573), which kept a trigger deriving a legacy column that old API readers still read, i.e. real dual-shape behavior during rollout.

The existing permanent-tier validator `scripts/test-computer-use-host-product-permanent.ts` was updated instead, since it asserted the `NOT NULL` invariant this migration removes. It now asserts the invariants that actually hold: the column is nullable with no default, omitted and explicit `NULL` writes land as `NULL`, and the retained check constraint still rejects out-of-contract values.

## Release gate

Migrations run before API promotion (`release-please.yml:1068-1069`). `DROP NOT NULL` lands first while the previous API keeps writing into a now-nullable column; then this artifact promotes and stops writing, leaving `NULL` on new rows.

**#32967 must ship in a strictly later production release than this PR.** If both landed together, the pre-this-PR API would still be live when `DROP COLUMN` ran and its insert would reference a dropped column, breaking registration and heartbeat.

## Why the strict-schema hazard does not apply

`listedComputerUseHostSchema` is `.strict()` and parsed across the SharedWorker port, but the worker asset is content-hashed (`vite.config.ts:55` matches `assets/shared-database-worker-[^/]+\.js`; the module is aliased through `?sharedworker&url` at `:65-67`). A new bundle requests a different worker URL, so the producer and consumer of a computed value are always the same build. No schema-version handshake was added.

## Tests

Removed the two BDD cases whose entire subject was the retired product identity (`tracks Desktop product identity across start and heartbeat`, `refreshes product identity when registering the same installation`) and the contract case that asserted the optional `product` field. No surviving assertion was weakened and no retired-behavior tombstone was added. The coverage they incidentally carried is retained elsewhere: installation dedup by `keeps installation hosts stable across stop and restart`, heartbeat responses by `keeps multiple active hosts and lets stale heartbeats recover`, and `desktopProductFromClientHeader` by `client-headers.test.ts`.

## Verification

- `pnpm -F @okouai/db test:migration-consistency` — pass, against a real PostgreSQL 18
- `check-types`: `@okouai/api-contracts`, `@okouai/db`, `api`, `@okouai/app`, `@okouai/cli`, `@okouai/desktop` — pass
- `lint`: `@okouai/api-contracts`, `@okouai/db`, `api`, `@okouai/app` — pass; `pnpm knip` — pass
- Prettier — pass on all changed files
- `api`: `computer-use.bdd.test.ts` (19), `app-factory.test.ts` (67) — pass
- `@okouai/api-contracts` (821), `@okouai/db` (106) — pass
- `@okouai/app`: `chat-capability-computer-use`, `computer-use-authorization-page`, `single-connection-client`, `message-port`, `realtime`, `chat-list-configuration` (51) — pass
- `@okouai/desktop`: `desktop-computer-use-runtime.test.ts` (10) — pass

Not run: the full suite. `chat-threads.bdd.test.ts` and `run-lifecycle.bdd.test.ts` fail identically on this branch and on unmodified `main` in this sandbox (`claimRunnerJob` → `Job not found in queue`); that is a missing local queue service, verified against a clean baseline, not a regression from this change.

## Not done here

No column or check-constraint drop, no row deletion/revocation/migration, no change to `/api/desktop/migration-policy`, the DMG routes, the legacy auth callback, host dedup, online/offline computation, selection, authorization, or command dispatch.


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: 7f5bc6e4f938e60bc593f9b2f02ad7408fdda64f
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `7f5bc6e4f938e60bc593f9b2f02ad7408fdda64f`. Practice revision: `1cafc4b03060b1cca2c9b0289c01b5bad6b87552`.

### Summary

Expand step of the two-step retirement of `computer_use_hosts.client_product` (#32966): migration `1100` relaxes the column to nullable, and the API stops writing it on registration/heartbeat and stops reading it in `serializeHost` and the heartbeat `publishChanged` dirty check. The `product` field is removed from `computerUseHostSchema`, `listedComputerUseHostSchema`, and the platform host mapping, while the physical column, its check constraint, and the Drizzle declaration stay for the contract step in #32967.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | `routes/computer-use.ts` start/heartbeat → `computer-use.service.ts` commands → `hosts/list` response → platform `external/computer-use-hosts.ts` → composer. No consumer reads `product`: `apps/cli/src/lib/api/domains/computer-use.ts` returns `result.body.hosts` untouched, and `apps/desktop` never reads it. |
| `REVIEW.md` | Existing coverage still protects the refactor; no assertion weakened | Pass | The two deleted BDD cases and the deleted contract case covered only the retired product identity. Installation dedup remains covered by `keeps installation hosts stable across stop and restart`, heartbeat responses by `keeps multiple active hosts and lets stale heartbeats recover`, and `desktopProductFromClientHeader` by `contracts/client-headers.test.ts`. |
| `REVIEW.md` | Review schema, API contract, and response validation together | Pass | Migration, Drizzle declaration, `computerUseHostSchema`, `listedComputerUseHostSchema`, and the worker computed value move together in this diff. |
| `REVIEW.md` | ccstate ownership preserved for API and Platform | Pass | `startComputerUseHost$`/`heartbeatComputerUseHost$` keep their `command` shape, transaction, `signal.throwIfAborted()` calls, and publish ownership; only a parameter is removed. `computerUseHosts$` remains a `computed` with the same `accept(..., [200, 403])` boundary. |
| `docs/bad-smell.md` §2 | No lint/type suppressions introduced | Pass | No suppression comment in the diff; `@okouai/api-contracts`, `@okouai/db`, `api`, `@okouai/app` lint and `check-types` pass. |
| `docs/bad-smell.md` §3 | Preserve meaningful error recovery when removing unused code | Pass | Only the product parameter/field is removed. `hostFromToken`, `invalid_token` handling, revocation filtering, and the `for("update")` row lock are unchanged. |
| `docs/bad-smell.md` §4 | Document changed public contracts and assess consumers | Pass | `product` removal from `computerUseHostSchema` is described in the PR body; consumers audited above. |
| `docs/bad-smell.md` §7 / `docs/fallback.md` §10 | Removing a fallback carries type, single-writer, production, or rollout evidence | Pass | The removed platform default `desktopProductFromClientHeader(host.product)` is unreachable once the contract field is gone (schema evidence), and #32966 records the 3-day production evidence: 0 Zero `hosts/start`, `heartbeat`, or `host/commands/*` calls against 274,155 okou heartbeats. |
| `docs/fallback.md` §1 | No negative/tombstone tests for removed behavior | Pass | No "column is absent" or "field is gone" assertion was added. The updated `test-computer-use-host-product-permanent.ts` asserts current live schema state (nullable, no default, check constraint still rejects `'unknown'`), not the absence of removed code. |
| `docs/fallback.md` §5 | Delete the reader in the same PR that retires the producer | Pass | `serializeHost`'s `product`, the `publishChanged` term, the contract field, the platform schema field, and the platform mapping are removed together with the writers. |
| `docs/fallback.md` §9 | Every introduced fallback declared in a `Fallbacks` PR-body section | N/A | The diff introduces no new fallback or tolerance branch; it removes one. Per §9, removal-only PRs need no `Fallbacks` section. |
| `docs/deployment-compatibility.md` — old code after migration | Every statement the draining API can issue stays legal | Pass | `DROP NOT NULL` only widens the accepted domain. The still-serving previous API keeps writing `'okou'` into a now-nullable column, and its `SELECT`/`RETURNING` shapes are unaffected because the column still exists. |
| `docs/deployment-compatibility.md` — new code before migration | New readers/writers must not require the new schema before the migration lands | Pass | Migrations run before API promotion (`release-please.yml:1068-1069`), and the new artifact requires nullability only for its own inserts. It reads nothing that the migration adds. |
| `docs/deployment-compatibility.md` — frontend/API version skew | Backend tolerates the previous frontend after deployment | Pass | `computerUseHostSchema` is a non-strict `z.object`, so a new app strips `product` from an old API response. An old app against the new API resolves `desktopProductFromClientHeader(undefined)` to `"zero"` and still satisfies its own `.strict()` `listedComputerUseHostSchema`; #32949 already removed both renderers, so nothing displays the value. |
| `docs/deployment-compatibility.md` — SharedWorker/computed values | Cross-build producer/consumer pairing for `.strict()` computed values | Pass | Verified the PR's claim: `apps/platform/vite.config.ts:55` matches `assets/shared-database-worker-[^/]+\.js` in `renderBuiltUrl` and `:65-67` aliases the module through `?sharedworker&url`, so a new bundle requests a different worker URL. The value is produced live in the worker (`shared-database/worker-signals.ts:21` imports the edited `signals/external/computer-use-hosts.ts`) and transported over the MessagePort; it is not persisted across builds. No version handshake is warranted. |
| `.claude/skills/database-development/SKILL.md` | Use Drizzle to generate migration metadata; do not hand-edit journals or snapshots | Pass | `1100_relax_computer_use_host_client_product.sql`, `meta/1100_snapshot.json`, and the `_journal.json` entry are `db:generate` output. The snapshot diff against `1099` is confined to `id`/`prevId` and `client_product.notNull: true → false`. |
| `.claude/skills/database-development/references/migrations.md` | Auto-generated migration flow, then verify | Pass | Generated with `pnpm -F @okouai/db db:generate`; the migration contains exactly one statement, and no existing migration or snapshot is modified. |
| `.claude/skills/database-development/references/query-contracts.md` | Query rewrites preserve the whole SQL contract | Pass | Only column entries are removed from an insert `values`, an `onConflictDoUpdate` `set`, and an update `set`. Conflict target, `targetWhere`, `returning({ id })`, transaction, and `for("update")` lock are unchanged; no raw expression or decoder is introduced. |
| `turbo/packages/db/MIGRATIONS.md` | Explicit transition-validator decision for the expand → contract cycle | Pass | The PR body records "not warranted" with reasoning (no dual-write, dual-read, backfill, derived value, or trigger; the release-ordering gate is not SQL-assertable), and contrasts it with the `1098_retire_member_invitation_capability` precedent it was routed from. |
| `turbo/packages/db/MIGRATIONS.md` | Permanent-tier invariants match the current schema | Pass | `test-computer-use-host-product-permanent.ts` was updated rather than deleted, so the retained check constraint keeps behavioral coverage. Verified green: `pnpm -F @okouai/db test:migration-consistency` against a real PostgreSQL 18. |
| `docs/testing.md` | Verification selected from changed files and consumers; no unscoped suites | Pass | Scoped Prettier, lint, `check-types` (contracts, db, api, app, cli, desktop), migration consistency, and the affected API/contract/platform/desktop test files were run. |
| `docs/testing/testing-external-behavior.md` | API tests construct and verify through production endpoints | Pass | The surviving `computer-use.bdd.test.ts` cases still register, heartbeat, list, and stop through the routes. No direct DB assertion was added; the DB-level checks stay in the `@okouai/db` migration-consistency tier, which is a schema validator rather than an API test. |

### Findings

No P0 or P1 defects.

Non-blocking observations:

1. **`SELECT` projection still enumerates the column — a planning input for #32967, not a defect here.** With the Drizzle declaration retained, `hostFromToken`'s `tx.select().from(computerUseHosts)` still generates an explicit projection containing `"client_product"`. Confirmed against the repo's Drizzle version:

   ```
   select "id", "org_id", "user_id", "installation_id", "display_name", "token_hash",
          "client_product", "app_version", ... from "computer_use_hosts" where ... limit $2
   ```

   That query backs heartbeat, stop, and host-command claiming. `docs/deployment-compatibility.md` ("Drop a Column as a Two-release Contract", and the migration `0722` post-mortem) warns that a physical drop while the previous release's Drizzle declaration still names the column yields `42703` on the draining API. #32967 currently scopes the declaration removal and the `DROP COLUMN` into one release, and its stated gate reasons only about `INSERT`/`UPDATE`. Keeping the declaration here is correct and matches the `1098` precedent this issue was routed from (`org-plan-entitlement.ts:105` keeps `legacyMemberInvitationAllowed` with a cleanup comment), and it is also forced by `test:migration-consistency`, which requires the ORM schema and the physical schema to match. So this is not a change request against this PR — it is evidence that #32967's release plan must account for the draining API's `SELECT` list, not only its writes.

2. **Old-app compatibility for the contract change is verified but not stated in the PR body.** The body covers the DB direction and the SharedWorker hazard thoroughly; it does not spell out the old-app-against-new-API direction. It is safe (see the frontend/API row above), so this is a documentation nicety, not a defect.

### Testing

Existing Computer Use coverage is preserved and no surviving assertion was weakened. Three tests were deleted, each covering only the retired behavior, which `docs/fallback.md` §8 treats as dying with the behavior rather than becoming a coverage loss:

- `tracks Desktop product identity across start and heartbeat`
- `refreshes product identity when registering the same installation`
- `accepts legacy host responses without product identity` (contracts)

The incidental coverage they carried survives elsewhere, as noted in the criteria table. The BDD helper's `clientProduct` option and `clientProductHeaders` were removed with their last callers, leaving `X-Client-Product` attribution coverage intact in `app-factory.test.ts` (`x_client_product` for okou, zero, and absent).

Verification observed on this HEAD:

- `pnpm -F @okouai/db test:migration-consistency` — pass, real PostgreSQL 18, including the updated permanent validator on both the applied-migration and regenerated-migration databases
- `check-types`: `@okouai/api-contracts`, `@okouai/db`, `api`, `@okouai/app`, `@okouai/cli`, `@okouai/desktop` — pass
- `lint`: `@okouai/api-contracts`, `@okouai/db`, `api`, `@okouai/app` — pass; `pnpm knip` — pass; Prettier — pass on all changed files
- `api`: `computer-use.bdd.test.ts` (19), `app-factory.test.ts` (67) — pass
- `@okouai/api-contracts` (821), `@okouai/db` (106), `@okouai/desktop` `desktop-computer-use-runtime.test.ts` (10) — pass
- `@okouai/app`: `chat-capability-computer-use`, `computer-use-authorization-page`, `single-connection-client`, `message-port`, `realtime`, `chat-list-configuration` (51) — pass

Material gap: none for this diff. `chat-threads.bdd.test.ts` and `run-lifecycle.bdd.test.ts` were not usable as evidence in the review sandbox — they fail identically on this branch and on unmodified `main` with `claimRunnerJob` → `Job not found in queue`, a missing local queue service rather than a regression. CI covers them.

PR #32997, opened 2026-09-09
feat(app): browse connectors as shelves in the directory and the connectors page

tongx-blip:
## What

Both connector surfaces — the composer directory dialog and the connectors page — now open on **shelves** instead of a wall of cards behind a filter.

- **Six per category**, in one-line cells, closed by a cell that names three of the rest: *"See Lark, Zendesk and 321 more"*. A count alone says nothing to someone who does not already know the product's name, which is the reader this catalogue has.
- **A head shelf** takes the best-ranked connector from each category — not the globally top six, which would be six Google products in a row. Nothing appears on two shelves.
- **A category with fewer than four ranked connectors cannot fill a shelf** (Voice and audio would open on `3Scribe · AiVOOV · Amara`), so it degrades to a counted chip under "More categories".
- **Connectors page**: status becomes its own segment control, category becomes a chip row, and the agent list leaves the filter — which agents may use a connector is already answered on the connector's own card, and that list grew without bound as an account aged. The floating category rail is a scroll-spy for the full dump and stands down with it.
- A keyword, a chip or a status is already a filter, so it returns the plain result list rather than shelves on top of it.

Design study, rendered with the platform's own tokens and the live catalogue: https://okou-connector-shelves.okou.app

## Compatibility

`popularityRank` and `categoryConnectorCounts` are optional (#32927). An API that predates them sends neither, so `buildConnectorShelves` falls back to one shelf per category and drops the head shelf rather than rendering an empty sheet — covered by a test.

## Feature switch

Both surfaces stay behind the existing `connectorDirectory` switch (Lab → Beta), so this is one toggle for the whole new connector experience.

## Verification

- `pnpm -F @okouai/app check-types`, `lint`, `pnpm lint:style`, `pnpm knip` — clean
- `connector-shelves.test.ts` (new, 5 cases incl. the unranked-API fallback), `connector-directory.test.tsx` (7), `connectors-catalog.test.tsx` (13), `chat-composer-connectors.test.tsx` — 45 passing

PR #33003, opened 2026-09-09
feat(app): put the connector shelves on cards and drop the yours tab

tongx-blip:
Follow-up to #32997, from acceptance on staging.

## Card shelves

The shelf cell goes back to the description card the product already ships — **four to a shelf** in the composer dialog, **six** on the connectors page, which is two whole rows of each surface's own card grid so a shelf never ends on a ragged row. The one-line cell fitted more products per screen but gave up the description, and the description is what tells someone who does not know a product's name whether it is the one they want. That is the reader the shelves exist for.

## No "Yours" tab

The composer's connector popover already lists every connected connector with its accounts and permissions, so the tab repeated it one click further away. The directory is now simply where a connector gets added, and its segment control is `Discover | Custom`.

Two things the tab uniquely carried move into discovery rather than disappearing:

- A connected connector that needs reconnecting or a permission update still appears at the top of Discover under **Needs attention** — the popover lists what is connected but not whether it still works.
- **Search now answers with connected connectors too.** Discovery only offers what can be added, so without this a search for something already connected would report "no match" for a connector the user is currently using. Its detail panel is reached from that result.

The strings the tab owned (`tabYours`, `searchYours`, `emptyYoursTitle`, `emptyYoursBody`, `browse`) are deleted from all ten locales with it.

## Feature switch

Unchanged: everything stays behind `connectorDirectory`, one toggle for both surfaces.

## Verification

- `@okouai/app` `check-types` and `lint`, `pnpm lint:style`, `pnpm knip` — clean
- 66 tests across `connector-directory`, `connectors-catalog`, `chat-composer-connectors`, `chat-composer-connector-accounts`, `connectors-accounts`
- `connector-directory.test.tsx` covers the new search-finds-a-connected-connector path and detail-from-search; the shelf tail count moves with the four-card preview

### DIFF
--- a/turbo/apps/platform/src/i18n/locales/de-DE/common.json
+++ b/turbo/apps/platform/src/i18n/locales/de-DE/common.json
@@ -1968,7 +1968,6 @@
       "directory": {
         "allCategories": "Alle",
         "askAdmin": "Lassen Sie diese Integration von einem Admin freischalten",
-        "browse": "Integrationen durchsuchen",
         "configure": "Konfigurieren",
         "connected": "Verbunden",
         "createCustom": "Eigene erstellen",
@@ -1980,8 +1979,6 @@
         "detailPermissions": "Berechtigungen",
         "emptyCustomBody": "Erstellen Sie eine aus einer HTTP-API oder einem MCP-Server, und Okou nutzt sie wie jede andere Integration.",
         "emptyCustomTitle": "Noch keine eigenen Integrationen",
-        "emptyYoursBody": "Verbinden Sie eine App, damit Okou darin für Sie lesen und handeln kann.",
-        "emptyYoursTitle": "Sie haben noch nichts verbunden",
         "matchCount_one": "{{count}} Treffer",
         "matchCount_other": "{{count}} Treffer",
         "moreAccounts_one": "+{{count}}",
@@ -1993,12 +1990,10 @@
         "otherCategory": "Sonstige",
         "permissionCount_one": "{{count}} Berechtigung",
         "permissionCount_other": "{{count}} Berechtigungen",
-        "searchYours": "Ihre Integrationen durchsuchen",
         "showAll_one": "Alle {{count}} anzeigen",
         "showAll_other": "Alle {{count}} anzeigen",
         "tabCustom": "Eigene",
-        "tabDiscover": "Entdecken",
-        "tabYours": "Ihre Integrationen"
+        "tabDiscover": "Entdecken"
       },
       "fallsBackTo": "Wechselt zu {{account}} zurück",
       "find": "Integrationen suchen...",
--- a/turbo/apps/platform/src/i18n/locales/en-US/common.json
+++ b/turbo/apps/platform/src/i18n/locales/en-US/common.json
@@ -1968,7 +1968,6 @@
       "directory": {
         "allCategories": "All",
         "askAdmin": "Ask an admin to enable",
-        "browse": "Browse connectors",
         "configure": "Configure",
         "connected": "Connected",
         "createCustom": "Create custom",
@@ -1980,8 +1979,6 @@
         "detailPermissions": "Permissions",
         "emptyCustomBody": "Build one from any HTTP API or MCP server, and Okou can use it like any other connector.",
         "emptyCustomTitle": "No custom connectors yet",
-        "emptyYoursBody": "Connect an app and Okou can read and act in it for you.",
-        "emptyYoursTitle": "You haven't connected anything yet",
         "matchCount_one": "{{count}} match",
         "matchCount_other": "{{count}} matches",
         "moreAccounts_one": "+{{count}}",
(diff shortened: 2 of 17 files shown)


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: 41bc3809fa901a972277d6eec9ef5657ce01edde
PR_AUTO_REVIEW_VERDICT: LGTM

Reviewed HEAD: `41bc3809fa901a972277d6eec9ef5657ce01edde` · practice revision `66bb28547ec2cbdaf7f4e10190c6931ccb4c1d95`

### Summary

Follow-up to #32997 from acceptance. The shelf cell returns to the description card the product already ships — four to a shelf in the composer dialog, six on the connectors page, two whole rows of each surface's card grid. The dialog's "Yours" tab is removed because the composer's connector popover already lists every connected connector; connection health moves to a "Needs attention" section at the top of Discover, and search now also answers with connected connectors so the detail panel stays reachable. Net −309 lines.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | `ComposerConnectorsSlot` opens the dialog on `directoryTab: "discover"`; `navigableDirectorySlugs` now returns `discoverSlugs` or nothing, and `discoverSlugs` still follows whichever branch renders (shelves, or the filtered list). |
| `REVIEW.md` | Regression coverage for a replaced or removed flow | Pass | The removed tab's two jobs are asserted in `connector-directory.test.tsx`: a search finds a connected connector under a "Connected" heading, and the detail panel opens from that result and steps back. |
| `REVIEW.md` | Removing code preserves meaningful recovery and permission behaviour | Pass | Nothing was dropped silently: `attention` keeps the reconnect/scope-mismatch surface, `renderCard(item, true)` keeps the detail entry, and account/permission management continues to live in the popover and on the connectors page. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | None added; the pre-existing `oxlint-disable max-lines-per-function` header on `connectors-page.tsx` is untouched. |
| `docs/bad-smell.md` | §4 interface changes assessed against consumers | Pass | `ConnectorDirectoryTab` narrows to `"discover" \| "custom"` and `ConnectorDirectoryModel` drops `healthy`/`yoursSlugs`/`connectedCount` for `matchedConnected`; both are app-internal, and `initialComposerConnectorUiState` is in-memory state with no storage or URL persistence, so no stale `"yours"` value can arrive. |
| `docs/bad-smell.md` | Delete unused code with the behaviour it served | Pass | `DirectoryYoursPanel`, `ConnectorShelfRow`, the `RenderShelfRow` plumbing, `connectorBusySlugs`, and the five strings the tab owned (`tabYours`, `searchYours`, `emptyYoursTitle`, `emptyYoursBody`, `browse`) are removed together across all ten locales. |
| `docs/fallback.md` | §1 no tombstone tests for removed behaviour | Pass | The "Separate connected connectors from the catalog" case is rewritten to the new behaviour rather than kept alongside it. |
| `docs/fallback.md` | §2/§9 no compatibility code for a non-GA switch; declare any fallback | N/A | The diff introduces no fallback; `previewSize` is a required layout input from each surface, not a tolerance branch. |
| `docs/deployment-compatibility.md` | Cover old/new combinations for changed contracts | N/A | No request, response, protocol, or persisted shape changes; `previewSize` and the tab type are client-local. |
| `docs/testing.md` | Prefer integration coverage through real entry points | Pass | All eight dialog cases boot the real composer via `setupPage` and drive the dialog through clicks and typing. |
| `docs/testing/testing-external-behavior.md` | Platform tests construct and verify through the page | Pass | No internal module is imported or asserted; the shelf preview change is observed through the closing cell's text ("See Telegram, Lark and 323 more"). |
| `docs/testing.md` | Mock external services at their boundary, centralized lifetimes | Pass | Unchanged fixtures (`installComposerConnectorFixture`); no new MSW handler or global server use. |
| `.claude/skills/ccstate/SKILL.md` | Derivation in `computed`, actions in commands, no store mutation in render | Pass | `buildConnectorDirectoryModel` remains a pure derivation called from the view; tab changes still go through `updateConnectorUiState$`. |
| `.claude/skills/feature-switch/SKILL.md` | New user-facing behaviour stays gated | Pass | Both surfaces remain behind `connectorDirectory`; with the switch off the composer renders `AddConnectorsDialog` and the page renders its previous toolbar and sections, which the "Keep the existing dialog when the directory switch is off" case still asserts. |
| `docs/styles.md` | Tailwind utilities and registered tokens only | Pass | `connector-shelf.tsx` now only lays out (`grid gap-3`, heading, tail button); the card surface it composes is `ConnectorCard`'s existing treatment. `pnpm lint:style` passes with no allowlist change. |
| `docs/styles.md` | Reuse the shared component instead of a parallel one | Pass | The shelf renders the surface's own `renderCard`, so the dialog and the page each keep exactly one connector card implementation instead of a second compact variant. |

### Findings

None blocking.

Non-blocking observations:

- `attention` is derived from `connected` without the search filter, so a broken connection stays pinned above the results while the reader types. That is deliberate for a health signal, but if it reads as noise during a search it is a one-line change.
- With four description cards plus a closing cell, a shelf is about 270px tall, so roughly one and a half shelves fit the 600px sheet. That is the density trade accepted when choosing the card variant over the one-line cell; the head shelf is what keeps the first screen useful.

### Testing

`connector-directory.test.tsx` covers the dialog opening on discovery without repeating a connected connector, finding that connector by name and seeing it under "Connected", opening its detail from the search result and stepping back, the closing cell naming its products and opening its category, one-shelf-only, the small-catalog fall-through to the plain list, the chip-metric guard, and the switch-off path. Verified on this HEAD: `@okouai/app` `check-types` and `lint`, `pnpm lint:style`, `pnpm knip`, and 66 tests across `connector-directory`, `connectors-catalog`, `chat-composer-connectors`, `chat-composer-connector-accounts`, and `connectors-accounts`.

No gap identified: the popover behaviour this change now relies on is covered by the existing `chat-composer-connectors` suite, which still passes unchanged.

PR #33015, opened 2026-09-09
fix(app): name and count connector category chips from discovery

tongx-blip:
Caught in acceptance: the chips read **"Communication Collaboration 11"** where the catalog says **"Communication and Collaboration"** and holds **329**.

## The name

The dialog took its category metadata from `/api/connector-catalog/status` — 4 048 connectors, **6.7 MB** — while rendering its connectors from the discovery response it already had. Until that much larger request resolved, `groupConnectorsByCategory` had no metadata and fell back to a name derived from the category id, so every chip *and* every shelf heading showed `Communication Collaboration` instead of the catalog's own label.

Discovery returns the same `categoryMetadata` (198 KB, 144 connectors), so the dialog now takes it from the response its connectors came from and **no longer subscribes to the full-catalog status endpoint at all**.

## The count

The number was `section.connectors.length` — the slice discovery returned for that category, at most twelve, and smaller once connectors are connected. A chip stands for the whole category, so it now uses `categoryConnectorCounts`, the same total the shelf's closing cell already used, and shows **no number** when the server reported none rather than a number that means something else.

## Verification

- `@okouai/app` `check-types`, `lint`, `pnpm knip` — clean
- 40 tests across `connector-directory`, `chat-composer-connectors`, `chat-composer-connector-accounts`, `chat-composer-connector-loading`
- New case "Count the whole category on a chip, not the slice discovery returned" asserts both halves through the dialog. Verified it fails without the fix: with `categoryMetadata` withheld the heading reads `Mail` and the chip reads `Communication9`.

### DIFF
--- a/turbo/apps/platform/src/signals/okou-page/connectors.ts
+++ b/turbo/apps/platform/src/signals/okou-page/connectors.ts
@@ -1,6 +1,7 @@
 import { command, computed, state, type Command, type Computed } from "ccstate";
 import type { ConnectorSlug } from "@okouai/api-contracts/contracts/connector-identity";
 import type { CustomConnectorResponse } from "@okouai/api-contracts/contracts/custom-connectors";
+import type { PublicConnectorCatalogCategoryMetadata } from "@okouai/api-contracts/contracts/connector-catalog";
 import { userConnectorsContract } from "@okouai/api-contracts/contracts/user-connectors";
 import {
   agentCustomConnectorsContract,
@@ -84,6 +85,12 @@ interface ComposerConnectorData {
   readonly categoryConnectorCounts:
     | Readonly<Record<string, number>>
     | undefined;
+  /**
+   * The catalog's own category names. They ride on the same discovery response
+   * as the connectors, so the directory never has to name a category before it
+   * knows what that category is called.
+   */
+  readonly categoryMetadata: PublicConnectorCatalogCategoryMetadata | undefined;
 }
 
 export interface ComposerConnectorSignals {
@@ -391,6 +398,7 @@ export function createComposerConnectorSignals(
       customConnectors,
       authorization,
       categoryConnectorCounts: catalog.categoryConnectorCounts,
+      categoryMetadata: catalog.categoryMetadata,
     };
   });
   const addDialogKeyword$ = computed((get) => {
--- a/turbo/apps/platform/src/views/okou-page/__tests__/chat-composer-connectors-test-helpers.ts
+++ b/turbo/apps/platform/src/views/okou-page/__tests__/chat-composer-connectors-test-helpers.ts
@@ -14,6 +14,7 @@ import {
 import {
   connectorCatalogContract,
   type PublicConnectorCatalogAuthMethodDetail,
+  type PublicConnectorCatalogCategoryMetadata,
   type PublicConnectorCatalogPermissionDetail,
   type PublicConnectorCatalogStatusItem,
 } from "@okouai/api-contracts/contracts/connector-catalog";
@@ -72,6 +73,8 @@ interface ConnectorFixtureOptions {
   readonly featuredConnectorSlugs?: readonly ConnectorSlug[];
   /** Category totals discovery reports, so a shelf can close on a real count. */
   readonly categoryConnectorCounts?: Readonly<Record<string, number>>;
+  /** The catalog's own category names, as discovery returns them. */
+  readonly categoryMetadata?: PublicConnectorCatalogCategoryMetadata;
   readonly customConnectors?: readonly CustomConnectorResponse[];
   readonly builtinAuthorizations?: Readonly<
     Record<string, readonly ConnectorSlug[]>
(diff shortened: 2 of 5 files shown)


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: 54b76eac06e8eca1db5759a91c2849b5010b490c
PR_AUTO_REVIEW_VERDICT: LGTM

Reviewed HEAD: `54b76eac06e8eca1db5759a91c2849b5010b490c` · practice revision `f42d0232ab3fdfa89cbf4c7f7262b55c9bc95074`

### Summary

The directory's category chips showed an id-derived name and the size of the slice discovery returned. Both are fixed by reading from the response the dialog already renders from: `categoryMetadata` now rides on `ComposerConnectorData` instead of being pulled from the 6.7 MB `/connector-catalog/status`, and the chip count comes from `categoryConnectorCounts`, rendering no number when the server reported none.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | `composerRelatedCatalog$` already produced the metadata; it is threaded through `ComposerConnectorData` → `ComposerConnectorsSlot` → `ConnectorDirectoryDialog` → `buildConnectorDirectoryModel`, the one place that consumed it. |
| `REVIEW.md` | Regression coverage for the fixed defect | Pass | "Count the whole category on a chip, not the slice discovery returned" asserts both halves through the dialog; the author records that withholding `categoryMetadata` makes it fail with `Mail` / `Communication9`, so the guard is load-bearing. |
| `REVIEW.md` | Distinguish introduced defects from unchanged code | Pass | The id-derived fallback inside `groupConnectorsByCategory` is untouched; this change only stops feeding it `undefined` on every first render. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | None added. |
| `docs/bad-smell.md` | §3 no fabricated defaults hiding a broken invariant | Pass | The chip renders a count only when `categoryCounts?.[category] !== undefined`, rather than substituting the slice length, which is what made the old number wrong rather than absent. |
| `docs/bad-smell.md` | §4 interface changes assessed against consumers | Pass | `ComposerConnectorData` gains one optional-valued field and `ConnectorDirectoryDialogProps` gains `categoryMetadata` and `loading`; both are app-internal with a single call site, updated in the same diff. |
| `docs/fallback.md` | §2/§9 no compatibility code for a non-GA switch; declare any fallback | N/A | No fallback is introduced. `categoryMetadata` being absent already had defined behaviour in `groupConnectorsByCategory`, and the count is now omitted rather than substituted. |
| `docs/deployment-compatibility.md` | Cover old/new combinations for changed contracts | N/A | No contract change; both fields already exist on the discovery response and stay optional. |
| `docs/testing.md` | Prefer integration coverage through real entry points | Pass | The new case boots the composer with `setupPage`, opens the dialog, and reads the rendered chips and heading. |
| `docs/testing/testing-external-behavior.md` | Platform tests construct and verify through the page | Pass | The case is constructed entirely from the mocked discovery response and verified from the DOM; no internal module is imported or asserted. |
| `docs/testing.md` | Mock external services at their boundary | Pass | `installComposerConnectorFixture` gains an optional `categoryMetadata` so the fixture can describe what discovery actually returns; the mock stays on `connectorCatalogContract.discovery`. |
| `.claude/skills/ccstate/SKILL.md` | Derivation in `computed`; view reads a loadable for lifecycle UI | Pass | `data$` is where the metadata is joined; the dialog's skeleton now keys off the composer data it renders from (`connectorData === undefined`) instead of an unrelated endpoint's loading state. |
| `.claude/skills/feature-switch/SKILL.md` | Changed behaviour stays gated | Pass | The dialog only renders behind `connectorDirectory`; the switch-off path is untouched. |
| `docs/styles.md` | Tailwind utilities and tokens; tests use roles or `data-*` hooks | Pass | No style change; the test selects the existing `data-connector-category-chip` hook and an accessible heading name. |

### Findings

None blocking.

Non-blocking observations:

- Dropping `connectorCatalogStatus$` means the composer no longer triggers the 6.7 MB full-catalog request when the dialog opens. That is a real win, and it is worth checking during acceptance that no other composer surface silently depended on that request being warm — `connectorCatalogStatusBySlug$` still has its own consumers, so nothing in this diff removes the endpoint itself.
- `expect(chips).toContain("Communication329")` compares concatenated `textContent`, so inserting a separator between the label and the count would fail the assertion for a cosmetic reason. Acceptable here because the concatenated string is exactly what the defect produced, but a two-part assertion would be less brittle if the chip gains more content.

### Testing

The new case pins both halves of the reported defect at the page boundary. `connector-directory.test.tsx` now has nine cases covering the shelf layout, the closing cell, the small-catalog fall-through, search reaching a connected connector, detail from search, and the chip metrics guard. Verified on this HEAD: `@okouai/app` `check-types`, `lint`, `pnpm knip`, and 40 tests across `connector-directory`, `chat-composer-connectors`, `chat-composer-connector-accounts`, and `chat-composer-connector-loading`.

PR #33018, opened 2026-09-09
fix(platform): use the popover surface and keep the flyout panel on screen

tongx-blip:
## What

Two fixes to the model picker flyout added in #32955, both found on the live
Lab build.

**Surface.** The two panels were hand-rolled cards — `border-border`,
`rounded-xl`, `shadow-md` — instead of the treatment the shared
`PopoverContent` uses (`rounded-[12px] border-[0.7px]
border-[hsl(var(--gray-400))] bg-card`, no shadow). The heavy drop shadow was
the most visible symptom. Both panels now take that surface from one constant.

**Placement.** The side measurement ran on mount, before the popover had
committed its collision-adjusted position, so it measured a stale box and could
open the panel past the right edge of the viewport (models' trailing badges were
cut off). It now measures after two animation frames, requires the panel to
actually fit on the chosen side, and only falls back to the roomier side when
neither side fits.

## Verification

  — 35 passed
- `pnpm check-types` in `turbo/apps/platform` — clean
- `eslint --max-warnings 0` on the changed files, `prettier`, and
  `node scripts/style-policy.mjs` — clean

Both behaviours are geometry and computed styles that jsdom does not lay out,
so they are verified against the live Lab build rather than by new assertions.

## Fallbacks

None.

### DIFF
--- a/turbo/apps/platform/src/signals/okou-page/model-picker-menu.ts
+++ b/turbo/apps/platform/src/signals/okou-page/model-picker-menu.ts
@@ -12,6 +12,7 @@ type ModelPickerMenuPage =
 type ModelPickerFlyoutSide = "left" | "right";
 
 const FLYOUT_PANEL_WIDTH = 258;
+const FLYOUT_VIEWPORT_MARGIN = 8;
 
 /** One navigation state per composer, including split chats. */
 export function createModelPickerMenuSignals() {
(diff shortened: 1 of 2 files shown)


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: 5bd4eb4c39fd36af8e2f4132af883752a1aeb9e0
PR_AUTO_REVIEW_VERDICT: LGTM

## Summary

Two follow-up fixes to the flyout from #32955: both panels adopt the shared
popover surface instead of a hand-rolled card, and the side measurement now
runs after the popover settles and requires the panel to fit.

Platform-only; no contract, schema, persisted shape, or API change.

## Review Criteria

- **Correctness.** `FLYOUT_PANEL_CLASS` is the only surface definition and is
  applied to both panels, so they cannot drift apart again. The side rule is
  now total: fits right, else fits left, else the roomier side; the previous
  `roomRight >= roomLeft` short-circuit could pick a side the panel overflowed.
  An 8px viewport margin keeps the panel off the edge.
- **ccstate ownership.** The `onRef` still owns everything it starts: the
  `requestAnimationFrame` handle is cancelled and the `resize` listener removed
  on abort. Frame scheduling replaced no other lifecycle.
- **Feature containment.** Both changes are inside the flyout layout, reachable
  only through `modelPickerFlyout` (`enabled: false`, staff/Lab only).
- **Styles.** Uses the same tokens the shared `PopoverContent` uses; no new
  first-party selector, and `style-policy` passes with no baseline growth.

## Findings

No blocking findings.

## Testing

  — 35 passed, covering both the single-panel and two-panel flyout paths.
- `pnpm check-types`, `eslint --max-warnings 0` on the changed files,
  `prettier --check`, `node scripts/style-policy.mjs` — clean.
- Both fixed behaviours are layout geometry and computed styles that jsdom does
  not produce, so no assertion can meaningfully pin them; they were diagnosed
  from the live Lab build and are verified there.

## Fallbacks

None.

PR #33102, opened 2026-09-10
fix(platform): match the flyout panel's surface and close it on selection

tongx-blip:
## What

Three fixes to the model picker flyout, all found on the Lab build.

**Mismatched shadow.** `PopoverContent` applies its drop shadow as an inline
`style`, not a class, so the `shadow-none` the flyout passed could never remove
it: the type rail carried the popover's shadow while the floating panel, which
painted its own card, had none. The popover is now the rail's card — it keeps
the component's hairline, radius and shadow and only overrides the width — and
the floating panel restates that same surface, with `shadow-lg` reproducing the
popover's inline shadow exactly.

**Selection left the panel open.** Picking a model committed the selection but
left the flyout up with nothing else to offer. The composer owns the open state,
so it passes a close callback the flyout invokes once a selection is committed;
this covers chat models and media models alike.

**Rows compressed instead of scrolling.** A flex column with a max-height
shrinks its children before it scrolls, and the rows carried no `shrink-0`.
Seven chat models fit inside the 284px panel, so those rows kept their 36px; ten
image models did not, so every row was squeezed to roughly 26px and the list
never scrolled — the same row rendered at a different height depending on how
many models the category had. Both row types now take `shrink-0`, so a long list
scrolls at its max height, and both are built from the shared ghost `Button`
instead of a raw element with hand-written hover, focus and disabled classes.

## Verification

  — 35 passed, with the flyout case asserting the panel is gone after a model is
  picked.
- `pnpm check-types` in `turbo/apps/platform` — clean.
- `eslint --max-warnings 0` on the changed files, `prettier --check`, and
  `node scripts/style-policy.mjs` — clean.

The shadow and the row compression are computed layout that jsdom does not
produce, so both are verified against the Lab build rather than by assertions.

## Fallbacks

None.

### DIFF
--- a/turbo/apps/platform/src/views/okou-page/__tests__/chat-composer-model-selection.test.tsx
+++ b/turbo/apps/platform/src/views/okou-page/__tests__/chat-composer-model-selection.test.tsx
@@ -740,4 +740,10 @@ test("Choose a model from the flyout without leaving the type list", async () =>
   expect(next).toHaveFocus();
   click(next);
   await expect(findButton("GPT 5.6 Luna")).resolves.toBeVisible();
+  // Picking a model finishes the task, so the panel leaves with it.
+  await waitFor(() => {
+    expect(
+      screen.queryByRole("listbox", { name: "Chat models" }),
+    ).not.toBeInTheDocument();
+  });
 });
--- a/turbo/apps/platform/src/views/okou-page/chat-composer.tsx
+++ b/turbo/apps/platform/src/views/okou-page/chat-composer.tsx
@@ -9689,6 +9689,9 @@ function ComposerRunModelPickerControl({
             : undefined
         }
         flyoutLayout={modelFlyoutEnabled}
+        onSelected={() => {
+          setModelPickerOpen(false);
+        }}
         compactTrigger
         mobileIconTrigger
         open={modelPickerOpen}
--- a/turbo/apps/platform/src/views/okou-page/components/model-picker-menu.tsx
+++ b/turbo/apps/platform/src/views/okou-page/components/model-picker-menu.tsx
@@ -472,16 +472,18 @@ function ModelPickerFlyoutTypeRow({
   onActivate: () => void;
 }) {
   return (
-    <button
+    <Button
       type="button"
+      variant="ghost"
       role="tab"
       aria-selected={active}
       aria-posinset={index + 1}
       aria-setsize={total}
       tabIndex={active ? 0 : -1}
       className={cn(
-        "flex h-11 w-full items-center gap-2 rounded-lg px-2 text-left transition-colors",
-        "hover:bg-state-hover focus-visible:ring-inset",
+        // shrink-0 keeps the row at its own height: a flex column with a
+        // max-height compresses its children before it will scroll.
+        "h-11 w-full shrink-0 justify-start gap-2 px-2 text-left font-normal",
         active && "bg-state-hover",
       )}
       onMouseEnter={onActivate}
@@ -504,7 +506,7 @@ function ModelPickerFlyoutTypeRow({
         aria-hidden="true"
         className="shrink-0 text-muted-foreground"
       />
-    </button>
+    </Button>
   );
 }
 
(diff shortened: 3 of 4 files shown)


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: 94bbee8b496ac3d6c4231967b8c28644831bd02f
PR_AUTO_REVIEW_VERDICT: LGTM

## Summary

Two flyout fixes: the floating panel now wears the same popover surface as the
type rail, and committing a selection closes the flyout. Platform-only; no
contract, schema, persisted shape, or API change.

## Review Criteria

- **Correctness.** The shadow mismatch was a real class-vs-inline-style
  conflict: `PopoverContent` sets `boxShadow` in `style`, which no utility class
  can clear. Making the popover the rail's card removes the conflict instead of
  fighting it, and `shadow-lg` is exactly the popover's inline value
  (`0 10px 15px -3px rgba(0,0,0,.1), 0 4px 6px -2px rgba(0,0,0,.05)`). With one
  type the list is now `w-full` inside the popover rather than a nested card,
  so the single-category case does not draw two boxes.
- **Ownership of open state.** `onSelected` is supplied by the composer, which
  already owns `modelPickerOpen$`; the picker does not reach for the popover's
  `onOpenChange`, whose signature requires a real `Event`. The prop is optional
  and spread only when present, so the settings and error-recovery call sites
  are unchanged and the menu layout still stays open after a selection.
- **Coverage.** Both selection paths call it: chat options through `onChange`,
  media options after `option.onSelect()`.
- **Styles.** Tailwind utilities and the same tokens the component uses; the
  style policy passes with no baseline growth.

## Findings

No blocking findings.

## Testing

  — 35 passed. The flyout case now asserts the `Chat models` listbox is gone
  after picking a model, which is the externally observable result of the close.
- `pnpm check-types`, `eslint --max-warnings 0` on the changed files,
  `prettier --check`, `node scripts/style-policy.mjs` — clean.
- The shadow itself is a computed style jsdom does not produce; it is verified
  on the Lab build.

## Fallbacks

None.

tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: 99f70ba58af84d80d4b7386e8aefdf39f3bc73f3
PR_AUTO_REVIEW_VERDICT: LGTM

## Summary

Re-review for the new HEAD, which adds a third fix: the flyout list compressed
its rows instead of scrolling, and both row types move onto the shared ghost
`Button`. The earlier two fixes (popover surface, close on selection) are
unchanged from the previous reviewed HEAD.

Platform-only; no contract, schema, persisted shape, or API change.

## Review Criteria

- **Correctness.** The row-height inconsistency was a flexbox default, not a
  spacing mistake: `flex flex-col max-h-[284px] overflow-y-auto` shrinks
  children before it scrolls, so `h-9` acted as a maximum rather than a floor
  once the content exceeded the panel. `shrink-0` on both the option and the
  type row restores the height and lets the panel scroll, which is what the
  `max-h` was there for. Nothing else depended on the rows being shrinkable.
- **Components.** Both rows now render the shared `Button variant="ghost"`, the
  same primitive the existing menu rows use, so hover, active and focus states
  come from the design system instead of hand-written utilities. The
  `aria-disabled` row keeps its own `hover:bg-transparent` because it stays
  focusable by design and must not read as actionable.
- **Accessibility.** `role`, `aria-selected`, `aria-posinset`, `aria-setsize`,
  `tabIndex` and the keyboard handler are unchanged; `Button` forwards them.
  Focusing the selected option on open now also scrolls it into view, which is
  the behaviour a scrolling list needs.
- **Styles.** Tailwind utilities and design-system tokens only; `style-policy`
  passes with no baseline growth.

## Findings

No blocking findings.

## Testing

  — 35 passed, covering the single-panel path, the two-panel type switch and
  close-on-selection.
- `pnpm check-types`, `eslint --max-warnings 0` on the changed files,
  `prettier --check`, `node scripts/style-policy.mjs` — clean.
- Row height and the shadow are computed layout that jsdom does not produce, so
  neither can be pinned by an assertion; both are verified on the Lab build.

## Fallbacks

None.

PR #32227, opened 2026-09-07
fix(platform): handle expected client failures and unsupported browsers

e7h4n:
## Summary

Expected connectivity, permission, and client-version failures no longer produce Sentry errors. Existing reconnect, microphone guidance, request rejection, and user-controlled update flows continue to handle them. Browsers missing a required capability now reach the existing unsupported-browser page before application bootstrap.

- Filter the exact Ably connection-unavailable and attach-timeout messages, microphone `NotAllowedError` outcomes (including stringified and chained causes), and the observed WebKit media-controls error and Safari wording in the shared page/worker Sentry policy.
- Preserve HTTP status and API error codes across MessagePort serialization, so 401 and 426 remain classified after deserialization and outer error wrapping. Genuine application, recorder/storage/transcription, protocol, and 5xx failures remain reportable.
- Handle HTTP 426 before checking Chat Event schema headers in catch-up, row, and snapshot requests. Keep the existing force-upgrade notification and manual refresh dialog.
- Check actual `SharedWorker` availability alongside the existing browser version checks. Present ordinary update/browser/device guidance using the existing compatibility page, including its pre-bootstrap English-copy policy.

## Root cause correction for #32148

`assertChatEventSchemaVersion()` reads an HTTP response header. The API compatibility middleware returns HTTP 426 without that header. Checking the header before the status turned an intentional upgrade response into `Unexpected Chat Event schema version null`. The observed PLATFORM-AQ events followed event-row HTTP 426 responses. This change corrects the response handling order; successful responses with missing or incompatible schema headers still fail and remain reportable. No database migration is needed.

## Verification

- 47 focused Platform tests passed across Sentry delivery hooks, worker runtime, MessagePort, browser entrypoint, and chat page integration tests.
- Coverage includes rejected 401 followed by a successful later query; direct and MessagePort 426 upgrade dialogs; legacy and draft voice permission guidance and retry; raw/serialized/wrapped expected errors; and retained 500, protocol, recorder, storage, transcription, and unrelated TypeError reporting.
- Platform formatting, lint (including localized UI, build configuration, i18n, Oxlint, and ESLint), type checking, and Knip passed.
- Full verification is delegated to the PR pipeline, as requested; no full local Vitest suite or local dev server was run.

Closes #32140
Closes #32141
Closes #32145
Closes #32147
Closes #32148
Closes #32151
Closes #32152

PR #33109, opened 2026-09-10
fix(platform): stop reporting webkit media-controls errors from chat pages

### LINKED ISSUE #33093: bug(platform): identify an unmapped Safari InvalidStateError on chat pages
## Summary

identify an unmapped Safari InvalidStateError on chat pages. The current-window observations and their causal limits are recorded below.

## Impact

One unhandled Mobile Safari 26.6.1 capture occurred on a chat route. User-visible damage and the initiating browser API are unknown.

Severity: **medium**. Type: **unmapped browser error investigation**. Severity is triage, not a raw-event-to-user equivalence.

### Coverage

- Sentry: projects platform; exact `environment:production level:error`, all issue statuses and complete event pagination. **1 events / 1 groups**, first `2026-09-09T14:39:54Z`; last `2026-09-09T14:39:54Z`. Group IDs: PLATFORM-BS.
- Distinct identified Sentry users: **1**, using populated native IDs locally; 0 captures lack a populated ID. No cross-source user union is claimed.
- Correlation: Native exception/caller/status identity groups repeated captures; cross-source correlation limits are stated below.

### Variant 1 — PLATFORM-BS / Error (1 events)

- Source: Sentry / platform / PLATFORM-BS; first `2026-09-09T14:39:54Z`; last `2026-09-09T14:39:54Z`.
- Native causal fields: `entries[type=exception].data.values[].type/value/mechanism/stacktrace.frames`, `culprit`, timestamp and tags.
- browser: Mobile Safari 26.6.1: 1.
- runtime: <not logged>: 1.
- release: <not logged>: 1.

Original sanitized sample 1:

```json
{
  "dateCreated": "2026-09-09T14:39:54Z",
  "message": "InvalidStateError: The object is in an invalid state.",
  "culprit": "?(chats/<redacted:chat-id>)",
  "tags": [
    {
      "key": "browser",
      "value": "Mobile Safari 26.6.1"
    },
    {
      "key": "browser.name",
      "value": "Mobile Safari"
    },
    {
      "key": "environment",
      "value": "production"
    },
    {
      "key": "level",
      "value": "error"
    },
    {
      "key": "os",
      "value": "iOS 18.7"
    }
  ],
  "entries": [
    {
      "type": "exception",
      "data": {
        "values": [
          {
            "type": "Error",
            "value": "InvalidStateError: The object is in an invalid state.",
            "mechanism": {
              "type": "auto.browser.global_handlers.onerror",
              "handled": false
            },
            "stacktrace": {
              "frames": [
                {
                  "filename": "chats/<redacted:chat-id>",
                  "function": null,
                  "lineNo": 1756,
                  "colNo": 34,
                  "inApp": true
                }
              ]
            }
          }
        ]
      }
    }
  ]
}
```

## Current code analysis

- [turbo/apps/platform/src/signals/shared-database-browser.ts](https://github.com/vm0-ai/vm0/blob/baa6e4c1f12b712e6e50dcf13e24e84cc77c8ecd/turbo/apps/platform/src/signals/shared-database-browser.ts)

The native exception type is Error, with value InvalidStateError: The object is in an invalid state. Its only app frame is the private chat URL at line 1756:34 with an empty source context. This is not the IndexedDB transaction stack in #32760/#32465, so it cannot be merged merely because the name matches. The linked current browser integration is an investigation entry point, not a claimed causal frame. No local production-browser reproduction was performed.

## Desired outcome

Recover a privacy-safe script identity/source map and operation/lifecycle classification, then reproduce or reach an explicit evidence-limited disposition. Fix a genuine browser lifecycle fault; if this is a demonstrably expected browser/extension condition, narrow only that capture boundary rather than suppress all InvalidStateError events.

## Ownership

Selected owner: @e7h4n. @e7h4n is the browser/platform lifecycle maintainer and fallback coordinator for this unlocalized first-party capture. Confidence: low until the actual source frame is recovered.

e7h4n:
## Summary

`InvalidStateError: The object is in an invalid state.` on chat pages (#33093) does not come from our code. It comes from Safari's own `<video>` controls script.

WebKit runs `modern-media-controls` inside the page and lets its exceptions escape to `window.onerror`. Its fullscreen button calls [`HTMLVideoElement.webkitEnterFullscreen()`](https://github.com/WebKit/WebKit/blob/main/Source/WebCore/Modules/modern-media-controls/media/fullscreen-support.js#L46-L53) unguarded, and [that API returns a message-less `InvalidStateError`](https://github.com/WebKit/WebKit/blob/main/Source/WebCore/html/HTMLVideoElement.cpp#L460-L472) when a fullscreen transition is already in flight, when there is no user gesture, or when fullscreen is unsupported. Rapid taps on an inline video hit the first condition. No application code can observe or prevent it, and nothing user-visible breaks — at worst that one tap does not enter fullscreen.

## Evidence

- Every capture has `DOMException.code = 11`, `mechanism = auto.browser.global_handlers.onerror`, `handled = no`, and a single frame whose `filename` is the document URL at **`1756:34`**.
- That line/column is identical across `app.vm0.ai` and `app.okou.ai`, four different chat URLs, and ten days (`PLATFORM-A2` on 2026-08-30/08-31/09-02, `PLATFORM-BS` on 2026-09-09). The served HTML document is 536 lines and the bundle changes every release, so the script is shipped by the browser, not by us.
- The same device also reports `Can't find variable: EmptyRanges` — [an identifier that exists only in WebKit's `media-controller.js`](https://github.com/WebKit/WebKit/blob/main/Source/WebCore/Modules/modern-media-controls/media/media-controller.js#L31) and nowhere in this repository — and the `this.mediaController.media.addEventListener` `TypeError` that #32227 already classified as WebKit media-controls noise.
- Breadcrumbs before each capture are repeated `ui.click` events on the same `<video>` element (3 taps in 6s, and 6 taps in 19s).
- This repository never calls `webkitEnterFullscreen`, `requestFullscreen`, `requestPictureInPicture`, or `webkitSetPresentationMode` on a media element.

Because the frame carries the chat UUID, the `culprit` differs per URL and each occurrence opens a **new** Sentry group: 6 groups / 8 events in 90 days for this message alone, plus 35 events for the `EmptyRanges` variant.

## Change

Third-party frame classification uses Sentry's own [`thirdPartyErrorFilterIntegration`](https://docs.sentry.io/platforms/javascript/configuration/filtering/#using-thirdpartyerrorfilterintegration). `@sentry/vite-plugin` embeds `applicationKey` in every bundled file and the integration resolves it per stack frame, so "is this frame ours?" is answered by the build rather than guessed from a filename. The key lives in one shared constant used by both the Vite config and the SDK options; a drift between them would mark everything as third-party.

The integration runs in **tagging** mode (`apply-tag-if-exclusively-contains-third-party-frames`), never a dropping one, and is registered for the page runtime only. Unmarked frames are an expected state — the Sentry plugin only runs when `SENTRY_AUTH_TOKEN` is set, and the shared worker is built by a separate Vite worker pipeline the plugin does not process — so an unmarked frame must not by itself discard an event.

`beforeSend` then drops a capture only when all of these hold:

- the page runtime;
- `mechanism.handled === false` and the mechanism is the global `onerror` handler;
- the exception value matches one of four exact media-controls messages;
- **and** the event carries `third_party_code: true`.

The message list stays the deciding condition, which keeps this narrow: the identical message raised from our own bundle keeps an application frame, is never tagged, and still reports. This deliberately avoids `ignoreErrors` (substring match over the whole message), `denyUrls` (the URL is our own page), and a blanket `name === "InvalidStateError"` rule, which would hide real IndexedDB, WebSocket, and recorder failures.

Also sets `release` from the existing build-time app version. Every capture reported `<not logged>` before, which is why #33093 could not attribute events to a build — and without it this filter cannot be verified after deployment either. Source maps resolve through `@sentry/vite-plugin` debug IDs and are unaffected.

## Verification

No tests in this PR, at the requester's direction. Existing `sentry-reporting.test.ts` cases construct events without a `mechanism` or a `third_party_code` tag, so the new predicate cannot match them and their behavior is unchanged.

After release, over a bounded 7-day window: expect zero new groups for `message:"The object is in an invalid state."` and `message:"EmptyRanges"` (baselines 8 and 35 events), zero new groups with a `?(chats/…)` culprit, and unchanged volume on application-side groups such as `PLATFORM-74` and `PLATFORM-AG` as a negative control. The `third_party_code` tag also becomes searchable in the issue stream, so the marking's coverage can be checked directly against real traffic.

Closes #33093

### DIFF
--- a/turbo/apps/platform/src/lib/__tests__/sentry-reporting.test.ts
+++ b/turbo/apps/platform/src/lib/__tests__/sentry-reporting.test.ts
@@ -181,3 +181,90 @@ test.each(["page", "shared-worker"] as const)(
     ).resolves.toBeNull();
   },
 );
+
+// WebKit's own <video> controls script escapes to window.onerror with no
+// application frame, so thirdPartyErrorFilterIntegration tags it before
+// delivery. Only that exact combination may be discarded.
+function userAgentEvent(
+  value: string,
+  overrides: {
+    readonly handled?: boolean;
+    readonly mechanismType?: string;
+    readonly thirdPartyCode?: boolean;
+  } = {},
+): ErrorEvent {
+  return {
+    type: undefined,
+    exception: {
+      values: [
+        {
+          type: "Error",
+          value,
+          mechanism: {
+            handled: overrides.handled ?? false,
+            type:
+              overrides.mechanismType ?? "auto.browser.global_handlers.onerror",
+          },
+        },
+      ],
+    },
+    tags: (overrides.thirdPartyCode ?? true) ? { third_party_code: true } : {},
+  };
+}
+
+const WEBKIT_FULLSCREEN_MESSAGE =
+  "InvalidStateError: The object is in an invalid state.";
+const WEBKIT_MEDIA_RANGES_MESSAGE = "Can't find variable: EmptyRanges";
+
+test("discards tagged webkit media-controls captures on the page", async () => {
+  const beforeSend = startSentry("page");
+  for (const message of [
+    WEBKIT_FULLSCREEN_MESSAGE,
+    "The object is in an invalid state.",
+    WEBKIT_MEDIA_RANGES_MESSAGE,
+    `ReferenceError: ${WEBKIT_MEDIA_RANGES_MESSAGE}`,
+  ]) {
+    await expect(
+      Promise.resolve(beforeSend(userAgentEvent(message), {})),
+    ).resolves.toBeNull();
+  }
+});
+
+test("reports the same messages when they carry an application frame", async () => {
+  const beforeSend = startSentry("page");
+  for (const message of [
+    WEBKIT_FULLSCREEN_MESSAGE,
+    WEBKIT_MEDIA_RANGES_MESSAGE,
+  ]) {
+    const event = userAgentEvent(message, { thirdPartyCode: false });
+    await expect(Promise.resolve(beforeSend(event, {}))).resolves.toBe(event);
+  }
+});
+
+test("reports webkit media-controls messages our own code captures", async () => {
+  const beforeSend = startSentry("page");
+  const handled = userAgentEvent(WEBKIT_FULLSCREEN_MESSAGE, {
+    handled: true,
+    mechanismType: "generic",
+  });
+  await expect(Promise.resolve(beforeSend(handled, {}))).resolves.toBe(handled);
+});
+
+test("reports neighbouring third-party and storage failures", async () => {
+  const beforeSend = startSentry("page");
+  for (const message of [
+    // The recovered IndexedDB failure shares the InvalidStateError name.
+    "InvalidStateError: Failed to execute 'transaction' on 'IDBDatabase': The database connection is closing.",
+    // An unrelated third-party capture must not be suppressed as a class.
+    "Can't find variable: somethingElse",
+  ]) {
+    const event = userAgentEvent(message);
+    await expect(Promise.resolve(beforeSend(event, {}))).resolves.toBe(event);
+  }
+});
+
+test("reports tagged media-controls captures in the shared worker", async () => {
+  const beforeSend = startSentry("shared-worker");
+  const event = userAgentEvent(WEBKIT_FULLSCREEN_MESSAGE);
+  await expect(Promise.resolve(beforeSend(event, {}))).resolves.toBe(event);
+});
--- a/turbo/apps/platform/src/lib/sentry-application-key.ts
+++ b/turbo/apps/platform/src/lib/sentry-application-key.ts
@@ -1,0 +1,6 @@
+// The Sentry bundler plugin embeds this key in every bundled application file,
+// and thirdPartyErrorFilterIntegration reads it back at runtime to tell our own
+// frames apart from scripts the browser or a third party injected into the page.
+// The build-time and runtime values must stay identical: a mismatch would mark
+// every frame as third-party code.
+export const SENTRY_APPLICATION_KEY = "okou-platform";
--- a/turbo/apps/platform/src/lib/sentry-config.ts
+++ b/turbo/apps/platform/src/lib/sentry-config.ts
@@ -62,6 +62,10 @@ const EXPECTED_ERROR_MESSAGES: ReadonlySet<string> = new Set([
 // throws a message-less InvalidStateError while a fullscreen transition is
 // already in flight, and the controls call it unguarded, so rapid taps on an
 // inline video report an error no application code can observe or prevent.
+// Each condition has two spellings because the SDK keeps a DOMException's bare
+// message when it carries a stack and prefixes the name when it does not.
+// Production has shown the prefixed fullscreen value and the bare ranges value;
+// the opposite spelling of each is kept for the other WebKit capture shape.
 const USER_AGENT_MEDIA_CONTROLS_MESSAGES: ReadonlySet<string> = new Set([
   "InvalidStateError: The object is in an invalid state.",
   "The object is in an invalid state.",
--- a/turbo/apps/platform/src/lib/sentry-config.ts
+++ b/turbo/apps/platform/src/lib/sentry-config.ts
@@ -1,6 +1,11 @@
 import { isDesktopAuthFlow } from "./desktop-auth-flow.ts";
 import * as Sentry from "@sentry/browser";
-import type { BrowserOptions, Contexts, User } from "@sentry/browser";
+import type {
+  BrowserOptions,
+  Contexts,
+  ErrorEvent,
+  User,
+} from "@sentry/browser";
 import { CLIENT_FORCE_UPGRADE_STATUS } from "@okouai/api-contracts/contracts/client-headers";
 
 import { setLogErrorHandler } from "../signals/log.ts";
@@ -51,6 +56,54 @@ const EXPECTED_ERROR_MESSAGES: ReadonlySet<string> = new Set([
   "this.mediaController.media.addEventListener is not a function. (In 'this.mediaController.media.addEventListener(eventType,this,true)', 'this.mediaController.media.addEventListener' is undefined)",
 ]);
 
+// WebKit runs its own <video> controls script inside the page and lets its
+// exceptions escape to window.onerror. HTMLVideoElement.webkitEnterFullscreen()
+// throws a message-less InvalidStateError while a fullscreen transition is
+// already in flight, and the controls call it unguarded, so rapid taps on an
+// inline video report an error no application code can observe or prevent.
+const USER_AGENT_MEDIA_CONTROLS_MESSAGES: ReadonlySet<string> = new Set([
+  "InvalidStateError: The object is in an invalid state.",
+  "The object is in an invalid state.",
+  "ReferenceError: Can't find variable: EmptyRanges",
+  "Can't find variable: EmptyRanges",
+]);
+
+// Application code always ships as a JavaScript asset. The user-agent script
+// carries no such frame: its only frame is the document URL of the page.
+const APPLICATION_SCRIPT_FILENAME = /\.m?js($|[?#])/u;
+
+const GLOBAL_ONERROR_MECHANISM = "auto.browser.global_handlers.onerror";
+
+function hasApplicationStackFrame(event: ErrorEvent): boolean {
+  return (event.exception?.values ?? []).some((value) => {
+    return (value.stacktrace?.frames ?? []).some((frame) => {
+      return (
+        frame.filename !== undefined &&
+        APPLICATION_SCRIPT_FILENAME.test(frame.filename)
+      );
+    });
+  });
+}
+
+// Narrow to the exact user-agent condition: an unhandled global capture that
+// matches a known media-controls message and attributes no application frame.
+// The same message raised from our own code keeps its asset frame and reports.
+function isUserAgentMediaControlsCapture(event: ErrorEvent): boolean {
+  const values = event.exception?.values;
+  if (values === undefined || hasApplicationStackFrame(event)) {
+    return false;
+  }
+  return values.some((value) => {
+    const mechanism = value.mechanism;
+    return (
+      mechanism?.handled === false &&
+      mechanism.type.startsWith(GLOBAL_ONERROR_MECHANISM) &&
+      value.value !== undefined &&
+      USER_AGENT_MEDIA_CONTROLS_MESSAGES.has(value.value)
+    );
+  });
+}
+
 function isExpectedErrorDescription(
   name: string | undefined,
   message: string | undefined,
@@ -100,6 +153,10 @@ export function createPlatformSentryOptions(
 
     environment: runtimeConfig.environment,
 
+    // Without a release every capture reports `<not logged>`, so a filter or
+    // fix cannot be verified against the build that produced the events.
+    release: __OKOU_APP_VERSION__,
+
     initialScope: {
       tags: {
         app: "platform",
@@ -137,6 +194,11 @@ export function createPlatformSentryOptions(
         statusCode >= 400 &&
         statusCode < 500
       ) {
+        return null;
+      }
+
+      // Only the page runtime renders <video>; the worker keeps every capture.
+      if (runtime === "page" && isUserAgentMediaControlsCapture(event)) {
         return null;
       }
 
--- a/turbo/apps/platform/src/lib/sentry-config.ts
+++ b/turbo/apps/platform/src/lib/sentry-config.ts
@@ -12,6 +12,7 @@ import { setLogErrorHandler } from "../signals/log.ts";
 import { SharedDatabaseHttpError } from "../shared-database/http-error.ts";
 import { ApiError } from "./api-error.ts";
 import { resolvePlatformRuntimeConfig } from "./platform-host.ts";
+import { SENTRY_APPLICATION_KEY } from "./sentry-application-key.ts";
 
 type PlatformSentryRuntime = "page" | "shared-worker";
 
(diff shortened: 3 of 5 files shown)


e7h4n:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 21f9c27ff79e221d6187eeeb556f2a0d583b16fd

Inspected HEAD: `21f9c27ff79e221d6187eeeb556f2a0d583b16fd`. Practice revision: `70c8a944dfe6c5dd61c1c5bf83fc6efb022cdcda`.

### Summary

Registers Sentry's `thirdPartyErrorFilterIntegration` in tagging mode for the page runtime, embeds a matching `applicationKey` through `@sentry/vite-plugin`, and drops an unhandled `onerror` capture in `beforeSend` when it matches a known WebKit media-controls message and carries `third_party_code: true`. Also sets `release` from the build-time app version and completes the mocked `@sentry/browser` surface.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | `sentry-config.ts:createPlatformSentryOptions` is reached only through `lib/sentry.ts:initSentry` and `shared-database/worker-sentry.ts`. Integration `processEvent` runs inside `_prepareEvent`, before `processBeforeSend` (`@sentry/core@10.46.0` `client.js`), so the `third_party_code` tag is present when `beforeSend` reads it. |
| `REVIEW.md` | Ordering, failure outcomes, and no unintended widening of a suppression boundary | Pass | The predicate is a four-way AND (page runtime, `handled === false`, global `onerror` mechanism, exact message, `third_party_code === true`). `integrations: []` on the worker preserves defaults: `getIntegrationsToSetup` spreads `[...defaultIntegrations, ...userIntegrations]`, so no shared-worker integration is dropped. |
| `REVIEW.md` | Require meaningful coverage for new behavior | **Finding (P1)** | No test exercises `isUserAgentMediaControlsCapture`. `src/lib/__tests__/sentry-reporting.test.ts` already drives the production `beforeSend` and has zero cases for the new branch. See Findings. |
| `REVIEW.md` | Persisted data, API contracts, response validation, compatibility windows | N/A | No schema, contract, persisted state, or request/response shape changes. `release` and `integrations` are client SDK options only. |
| `REVIEW.md` | Fallbacks declared in a PR-body `Fallbacks` section | Pass | The diff introduces no fallback under `docs/fallback.md` §6. `integrations: runtime === "page" ? [...] : []` is deliberate runtime scoping for a state the build genuinely produces (the worker bundle is built by a separate Vite worker pipeline the Sentry plugin does not process), not a compatibility branch. No `Fallbacks` section is required. |
| `REVIEW.md` | Optimistic/persistent event reconciliation | N/A | No events are produced or reconciled by this diff. |
| `docs/bad-smell.md` | No `any`; narrow untrusted boundaries | Pass | No `any`. `ErrorEvent`, `Parameters<...>`, and `ReturnType<...>` are used; `mechanism.type` is non-optional in `@sentry/core`'s `Mechanism`, so the access needs no guard. |
| `docs/bad-smell.md` | No lint/type suppression comments; fix violations at the source | Pass | No suppression comments. The `lint-type-app` failure on the previous HEAD was fixed by deriving the mock's return type from the real factory rather than suppressing it. |
| `docs/bad-smell.md` | Error handling only at a boundary that meaningfully handles it; no fabricated defaults hiding a broken invariant | Pass | `beforeSend` is the delivery boundary. The drop is bounded by an explicit signature list plus a build-derived third-party tag; it does not swallow an application failure class. |
| `docs/bad-smell.md` | Public contract changes documented with consumers assessed | Pass | `sentry-application-key.ts` is a new internal constant with exactly two consumers (`vite.config.ts`, `sentry-config.ts`); its comment states the build/runtime equality invariant. |
| `docs/bad-smell.md` | Static imports in production code | Pass | `sentry-config.ts` uses a static import. The `typeof import(...)` forms in the test mock are type-only and erased. |
| `docs/bad-smell.md` | Centralized configuration; no invented fallback URLs or credentials | Pass | `applicationKey` is a single shared constant, not an env-specific value; `release` reuses the existing `__OKOU_APP_VERSION__` build define already used by `lib/client-telemetry.ts`. |
| `docs/fallback.md` | No branch for a state the owning contract cannot produce (§3) | **Finding (P2, non-blocking)** | Two of the four entries in `USER_AGENT_MEDIA_CONTROLS_MESSAGES` are not observed in the cited production data. See Findings. |
| `docs/fallback.md` | No negative tests pinning removed behavior (§1) | N/A | The diff removes no behavior and adds no tombstone test. |
| `docs/testing.md` | Add tests for new behavior where they provide confidence beyond existing checks | **Finding (P1)** | The new suppression branch has no case in any suite. See Findings. |
| `docs/testing.md` | Mock external services at their boundary; use real internal code | Pass | `src/test/mocks/sentry-browser.ts` is the existing module-level mock of the external SDK, registered through `vitest.config.ts` aliases. The new export follows the file's existing typing idiom and stays inert. |
| `docs/testing/testing-external-behavior.md` | Platform tests use the page as the external interface | Pass (with note) | The Sentry delivery boundary has no page surface; `sentry-reporting.test.ts` already treats `initSentry()` as the production entry point and the delivered/dropped event as the observable outcome. Extending that established file is the correct boundary here. |

### Findings

**P1 — the new suppression branch has no test coverage**

- File: `turbo/apps/platform/src/lib/sentry-config.ts:74-90` (`isUserAgentMediaControlsCapture`) and `:200-203` (the `beforeSend` call site).
- Trigger: any future edit to the message set, the mechanism check, or the `third_party_code` condition.
- Consequence: `beforeSend` silently discards production error events. A widened predicate — dropping the tag check, loosening `handled === false`, or matching a substring instead of the exact value — would hide real application failures with no failing test and no runtime signal. This is precisely the class `docs/testing.md` asks to cover: a meaningful security/observability contract rather than static configuration.
- Practice: `REVIEW.md` → "Require meaningful coverage for new behavior and regressions"; `docs/testing.md` → "Add tests for new behavior and regressions where they provide confidence beyond existing checks."
- Required: `src/lib/__tests__/sentry-reporting.test.ts` already builds `ErrorEvent` objects and asserts `beforeSend(...)` resolves to `null` or the event. Add, in that same file and style:
  - two positive cases (tagged + `onerror`/`handled: false` + each observed message) asserting `null`;
  - a negative control with the identical message and **no** `third_party_code` tag asserting the event is preserved — this is the case that proves an application-raised `InvalidStateError` still reports;
  - a negative control that is tagged but `handled: true` asserting preservation;
  - a negative control with the real IndexedDB message `InvalidStateError: Failed to execute 'transaction' on 'IDBDatabase': The database connection is closing.` asserting preservation;
  - a negative control with a different tagged third-party `onerror` message asserting preservation, proving the change does not suppress all third-party captures;
  - a `shared-worker` runtime case asserting the tagged media-controls event is preserved, pinning the `runtime === "page"` scope.

**P2 (non-blocking) — two message entries are not backed by observed production data**

- File: `turbo/apps/platform/src/lib/sentry-config.ts:65-70`.
- The PR body cites `PLATFORM-BS`/`PLATFORM-A2` with `type: "Error"`, `value: "InvalidStateError: The object is in an invalid state."`, and `PLATFORM-50` with `type: "ReferenceError"`, `value: "Can't find variable: EmptyRanges"`. The other two entries — the unprefixed `"The object is in an invalid state."` and the prefixed `"ReferenceError: Can't find variable: EmptyRanges"` — are the opposite spellings and appear in neither cited group.
- `docs/fallback.md` §3 asks not to carry paths for states the owning contract does not produce. Both spellings are reachable in principle across SDK/WebKit variants (Sentry's `eventFromError` path keeps the bare `message`, while `_eventFromIncompleteOnError` strips only recognised JS error names), so this is a defensible hedge rather than a defect.
- Not blocking. Either keep them with a one-line comment recording that they are the alternate SDK spelling and unobserved so far, or drop them to the two observed values. Please make the choice explicit rather than leaving it implicit.

### Testing

- Existing coverage is preserved: every case in `sentry-reporting.test.ts` constructs events without a `mechanism` or a `third_party_code` tag, so the new predicate cannot match them, and the previously covered classifications are unchanged.
- The mock change is required, not cosmetic: `initSentry()` now calls `Sentry.thirdPartyErrorFilterIntegration`, which the aliased mock did not export. Without it every suite that boots the platform entrypoint throws.
- CI on this HEAD is green, including `lint-type-app`, all eight `test-app` shards, `deploy-app`, and the three `ci-gate-*` jobs.
- Reported out-of-suite verification: the author exercised the deployed preview bundle's live `beforeSend` with seven constructed events (two targets dropped; five controls preserved, including the untagged same-message case and the real IndexedDB message) on two successive preview deployments. That is useful evidence, but it is a manual one-off against a preview build and is not a substitute for the committed cases requested above.
- Material gap: none beyond the P1. The end-to-end path (real iOS Safari, marked bundle, live Sentry ingestion) cannot be exercised in preview because preview builds resolve `sentryDsn: null` and run without `SENTRY_AUTH_TOKEN`; the PR body's bounded post-release observation window is the right instrument for that and is already specified.

e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: d89de9430088d8b90f4953e558ad2a6d0a613f4c
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `d89de9430088d8b90f4953e558ad2a6d0a613f4c`. Practice revision: `70c8a944dfe6c5dd61c1c5bf83fc6efb022cdcda`. Supersedes the `21f9c27f` review; both findings raised there are resolved.

### Summary

Registers Sentry's `thirdPartyErrorFilterIntegration` in tagging mode for the page runtime, embeds a matching `applicationKey` through `@sentry/vite-plugin`, and drops an unhandled `onerror` capture in `beforeSend` when it matches a known WebKit media-controls message and carries `third_party_code: true`. Also sets `release` from the build-time app version, covers the new delivery boundary, and completes the mocked `@sentry/browser` surface.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | `createPlatformSentryOptions` is reached only through `lib/sentry.ts:initSentry` and `shared-database/worker-sentry.ts`. Integration `processEvent` runs inside `_prepareEvent`, before `processBeforeSend` (`@sentry/core@10.46.0` `client.js`), so the tag is present when `beforeSend` reads it. |
| `REVIEW.md` | Ordering, failure outcomes, and no unintended widening of a suppression boundary | Pass | Four-way AND (page runtime, `handled === false`, global `onerror` mechanism, exact message, `third_party_code === true`). `integrations: []` on the worker preserves defaults: `getIntegrationsToSetup` spreads `[...defaultIntegrations, ...userIntegrations]`. |
| `REVIEW.md` | Require meaningful coverage for new behavior | Pass | `sentry-reporting.test.ts:218-269` adds five cases through `startSentry()`/`beforeSend`: the discarded set, the untagged same-message control, the handled control, the IndexedDB and unrelated third-party controls, and the shared-worker scope. |
| `REVIEW.md` | Regression coverage for a rerouted existing flow and correct containment | Pass | The prior classification path is untouched; every pre-existing case in the same file constructs events without a `mechanism` or tag, so the new predicate cannot reach them. The new shared-worker case pins the `runtime === "page"` scope. |
| `REVIEW.md` | Persisted data, API contracts, response validation, compatibility windows | N/A | No schema, contract, persisted state, or request/response shape changes. `release` and `integrations` are client SDK options only. |
| `REVIEW.md` | Fallbacks declared in a PR-body `Fallbacks` section | Pass | The diff introduces no fallback under `docs/fallback.md` §6. `integrations: runtime === "page" ? [...] : []` is deliberate runtime scoping for a state the build genuinely produces, not a compatibility branch. No `Fallbacks` section is required. |
| `REVIEW.md` | Optimistic/persistent event reconciliation | N/A | No events are produced or reconciled by this diff. |
| `docs/bad-smell.md` | No `any`; narrow untrusted boundaries | Pass | No `any`. `ErrorEvent`, `Parameters<...>`, and `ReturnType<...>` are used; `mechanism.type` is non-optional in `Mechanism`, so the access needs no guard. |
| `docs/bad-smell.md` | No lint/type suppression comments | Pass | None added. The earlier `lint-type-app` failure was fixed by deriving the mock's return type from the real factory. |
| `docs/bad-smell.md` | Error handling only at a boundary that meaningfully handles it | Pass | `beforeSend` is the delivery boundary. The drop is bounded by an explicit signature list plus a build-derived third-party tag; it does not swallow an application failure class. |
| `docs/bad-smell.md` | Public contract changes documented with consumers assessed | Pass | `sentry-application-key.ts` has exactly two consumers (`vite.config.ts`, `sentry-config.ts`); its comment states the build/runtime equality invariant. |
| `docs/bad-smell.md` | Static imports in production code | Pass | `sentry-config.ts` uses a static import; `typeof import(...)` in the mock is type-only. |
| `docs/bad-smell.md` | Centralized configuration; no invented fallback values | Pass | `applicationKey` is a single shared constant; `release` reuses the existing `__OKOU_APP_VERSION__` define already used by `lib/client-telemetry.ts`. |
| `docs/fallback.md` | No branch for a state the owning contract cannot produce (§3) | Pass | `sentry-config.ts:60-63` now records why each condition carries two spellings: the SDK keeps a DOMException's bare message when it has a stack and prefixes the name when it does not. Both shapes are reachable, and the choice is explicit rather than implicit. |
| `docs/fallback.md` | No negative tests pinning removed behavior (§1) | Pass | The added tests assert live behavior on both sides of the boundary; none pins removed code. |
| `docs/testing.md` | Tests add confidence beyond existing checks, not restated configuration | Pass | Four of the five added cases are negative controls that fail if the predicate widens. They exercise behavior, not the message set as data. |
| `docs/testing.md` | Mock external services at their boundary; use real internal code | Pass | `src/test/mocks/sentry-browser.ts` is the existing module-level SDK mock registered through `vitest.config.ts`; the new export follows the file's typing idiom and stays inert. The tests use the real `createPlatformSentryOptions` and real `initSentry`. |
| `docs/testing.md` | Centralized `testContext()`/`context.mocks` lifetimes; no fake timers | Pass | The added cases reuse the file's existing `startSentry()` helper and module-level `testContext()`. No timers, no manual detached cleanup. |
| `docs/testing/testing-external-behavior.md` | Platform tests use the external interface, not internal state | Pass (with note) | The Sentry delivery boundary has no page surface. The added cases drive `initSentry()` as the production entry point and assert the delivered/dropped event, matching the file's established convention; they assert no internal state or call counts. |

### Findings

No P0 or P1 findings.

Both findings from the `21f9c27f` review are resolved:

- **P1 (missing coverage) — resolved.** `sentry-reporting.test.ts` now covers the discarded combination and, more importantly, five neighbouring conditions that must still report. The untagged same-message case is the one that proves an application-raised `InvalidStateError` is unaffected.
- **P2 (unobserved message spellings) — resolved.** The comment at `sentry-config.ts:60-63` states the SDK capture-shape reason for both spellings, so the entries production has not yet shown are a recorded decision.

Non-blocking observations, no action required:

- `vite.config.ts` importing `src/lib/sentry-application-key.ts` points the build config at application source. That is the correct trade here — a drift between the two values would tag every frame as third-party — and the file is a single `const` with no runtime dependencies.
- The end-to-end path (real iOS Safari, a marked bundle, live Sentry ingestion) is not reachable in CI or preview, because preview builds resolve `sentryDsn: null` and run without `SENTRY_AUTH_TOKEN`. The PR body's bounded post-release observation window is the right instrument and is already specified.

### Testing

- Added: five cases in `src/lib/__tests__/sentry-reporting.test.ts` covering the discarded set (both spellings of both conditions), the untagged control, the handled control, the recovered IndexedDB `InvalidStateError`, an unrelated tagged third-party capture, and the `shared-worker` runtime.
- Preserved: every pre-existing case in that file is untouched and unreachable by the new predicate.
- The mock change is required rather than cosmetic: `initSentry()` now calls `Sentry.thirdPartyErrorFilterIntegration`, which the aliased mock did not export.
- CI verification for this HEAD is pending at review time and is the gate for merge.

PR #33127, opened 2026-09-10
fix(api): serve desktop update feeds through transient manifest 5xx

### LINKED ISSUE #33084: chore(desktop): resolve manifest 5xx error reporting after deferred reliability work
## Summary

resolve manifest 5xx error reporting after deferred reliability work. The current-window observations and their causal limits are recorded below.

Related context: #32462. Closed issues remain closed; this review does not override prior implementation/release authority.

## Impact

One desktop update-feed request failed, captured once by Axiom and once by Sentry API-11, not two independent failures.

Severity: **medium**. Type: **upstream update-feed failure and reporting policy**. Severity is triage, not a raw-event-to-user equivalence.

### Coverage

- Axiom: ONLY `vm0-web-logs-prod`, `level == "error" or level == "warn"`; **1 records**, first `2026-09-09T00:08:36.952Z`; last `2026-09-09T00:08:36.952Z`. Levels: error: 1.
- Sentry: projects api; exact `environment:production level:error`, all issue statuses and complete event pagination. **1 events / 1 groups**, first `2026-09-09T00:08:36Z`; last `2026-09-09T00:08:36Z`. Group IDs: API-11.
- Correlation: Native exception/caller/status identity groups repeated captures; cross-source correlation limits are stated below.
- Axiom and Sentry are source counts, not an additive unique-incident total.

### Variant 1 — Unhandled request error: Desktop update manifest fetch failed with 500 (1 records)

- Source: Axiom; first `2026-09-09T00:08:36.952Z`; last `2026-09-09T00:08:36.952Z`.
- Native causal fields: `fields.error.message`, `fields.error.name`, `fields.errorSummary`, `fields.method`, `fields.route`, `fields.type`.

Original sanitized sample 1:

```json
{
  "_time": "2026-09-09T00:08:36.952Z",
  "errorSummary": "Desktop update manifest fetch failed with 500",
  "fields.context": "App",
  "fields.error.message": "Desktop update manifest fetch failed with 500",
  "fields.error.name": "Error",
  "fields.error.stack": "Error: Desktop update manifest fetch failed with 500\n    at fetchDesktopUpdateManifest (file:///var/task/index.js:8453:21397)\n    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)\n    at async loadDesktopUpdateManifest (file:///var/task/index.js:8453:21830)\n    at async loadDesktopUpdateFeed (file:///var/task/index.js:8453:22063)\n    at async Object.write (file:///var/task/index.js:8453:24548)\n    at async file:///var/task/index.js:74:53301\n    at async Dn (file:///var/task/index.js:72:42188)\n    at async webClientCompatibilityMiddleware (file:///var/task/index.js:7680:35121)\n    at async Dn (file:///var/task/index.js:72:42188)\n    at async file:///var/task/index.js:7680:37648",
  "fields.errorSummary": "Desktop update manifest fetch failed with 500",
  "fields.method": "GET",
  "fields.route": "/api/desktop/updates/:product/:channel/:platform/:arch/RELEASES.json",
  "fields.type": "unhandled_request_error",
  "level": "error",
  "message": "Unhandled request error: Desktop update manifest fetch failed with 500",
  "method": "GET",
  "route": "/api/desktop/updates/:product/:channel/:platform/:arch/RELEASES.json",
  "source": "api",
}
```

### Variant 2 — API-11 / Error (1 events)

- Source: Sentry / api / API-11; first `2026-09-09T00:08:36Z`; last `2026-09-09T00:08:36Z`.
- Native causal fields: `entries[type=exception].data.values[].type/value/mechanism/stacktrace.frames`, `culprit`, timestamp and tags.
- browser: <not logged>: 1.
- runtime: node v22.23.2: 1.
- release: bf29b32672ca38d03979754e8a022bbdab35fb6a: 1.

Original sanitized sample 1:

```json
{
  "dateCreated": "2026-09-09T00:08:36Z",
  "culprit": "fetchDesktopUpdateManifest(index)",
  "tags": [
    {
      "key": "environment",
      "value": "production"
    },
    {
      "key": "level",
      "value": "error"
    },
    {
      "key": "os",
      "value": "Linux"
    },
    {
      "key": "release",
      "value": "bf29b32672ca38d03979754e8a022bbdab35fb6a"
    },
    {
      "key": "runtime",
      "value": "node v22.23.2"
    }
  ],
  "entries": [
    {
      "type": "exception",
      "data": {
        "values": [
          {
            "type": "Error",
            "value": "Desktop update manifest fetch failed with 500",
            "mechanism": {
              "type": "generic",
              "handled": true
            },
            "stacktrace": {
              "frames": [
                {
                  "filename": "/var/task/index.js",
                  "function": "loadDesktopUpdateFeed",
                  "lineNo": 8453,
                  "colNo": 22063,
                  "inApp": true
                },
                {
                  "filename": "/var/task/index.js",
                  "function": "loadDesktopUpdateManifest",
                  "lineNo": 8453,
                  "colNo": 21830,
                  "inApp": true
                },
                {
                  "filename": "/var/task/index.js",
                  "function": "fetchDesktopUpdateManifest",
                  "lineNo": 8453,
                  "colNo": 21397,
                  "inApp": true
                }
              ]
            }
          }
        ]
      }
    }
  ]
}
```

Variant reconciliation: 1 + 1 = **2 source records** (not unique incidents).

## Current code analysis

- [turbo/apps/api/src/signals/services/desktop-updates.service.ts](https://github.com/vm0-ai/vm0/blob/baa6e4c1f12b712e6e50dcf13e24e84cc77c8ecd/turbo/apps/api/src/signals/services/desktop-updates.service.ts)

The exception, fetchDesktopUpdateManifest stack and same-second event match across sources. Current code throws on a non-OK manifest fetch and only uses unexpired cached manifests; no safe stale fallback is implemented here. #32462 was explicitly closed not_planned, not fixed, and its closure already mentions this same September 9 500. Preserve that prioritization context while tracking the still-required error-signal disposition; this is not a new outage claim.

## Desired outcome

Decide and verify a bounded retry/stale-feed policy or a controlled temporary-unavailable outcome compatible with normal client polling. If no intervention is expected, do not continue emitting an unhandled request error/Sentry error for that routine condition. Genuine invalid manifests and sustained update unavailability remain actionable.

## Ownership

Selected owner: @lancy. @lancy owns exact desktop-update delivery issue #32462 and this service. Confidence: high.

lancy:
## Summary

Closes #33084.

The desktop update routes read a release manifest from `github.com/vm0-ai/vm0/releases/download/...`, a host we do not control. Any non-OK response threw, so one upstream blip became a `500`, an `unhandled_request_error`, and a Sentry event — for a condition that needs no intervention and that the client recovers from on its next 30-minute poll.

This classifies the failure instead of flattening it, and reclassifies the signal to match.

## Evidence this is worth separating

Sentry group `API-11` holds **34 events since 2026-06-17**, and they are two different problems:

| Upstream status | Events | What it is |
| --- | --- | --- |
| `404` | 30 | The manifest asset was missing — 30 of them clustered on 2026-08-12, the Okou identity cutover. A real broken release. |
| `504` / `503` / `500` | 4 | The host was briefly unreadable. Self-correcting. |

So the loud channel has caught a real problem exactly once, and the transient class fires about **once every three weeks**. Over 2026-09-06 → 09-10 the feed route served **662 / 736 / 790** requests per day with **one** `5xx` on 09-07 and **one** on 09-09 — a 0.04% failure rate. Both were real Squirrel clients (`Okou/… CFNetwork/… Darwin/…`), and the affected build was back to `200` within the same 30-minute window.

The two failures also had very different shapes: the 09-09 one failed in **271 ms**, the 09-07 one took **11 072 ms** before the upstream gateway gave up.

## What changed

**Availability failures** — connection failure, per-attempt deadline, `408`, `429`, `5xx`:
1. retried up to 3 times, 200 ms apart;
2. then answered from an already-validated cached manifest, bounded to 30 minutes from the fetch;
3. then reported as `503` with `Cache-Control: no-store` and `Retry-After: 60`.

**Integrity failures** — upstream `404`, schema failure, product mismatch — still throw, still reach Sentry, and are never answered from the cache. That manifest is published by our own release pipeline, so its absence is a broken release, not an outage. A `404` is not retried either.

**Each attempt now carries its own 3s deadline.** This is the part that is easy to miss: given the recorded 11s gateway hang, retrying *without* a deadline would have made the worst case ~33s — worse than the single unbounded attempt it replaces.

**Outcomes are one structured event family**, `desktop_update_manifest_upstream`, lifted into the Axiom event root:

| Outcome | Level | Sentry |
| --- | --- | --- |
| `retry_recovered` | `debug` | no |
| `served_stale` | `debug` | no |
| `unavailable` (the `503`) | `warn` | no |

`debug` for the absorbed cases follows the repo's `api/no-logger-info` rule — routine diagnostics are not `info`. A single blip no longer pages anyone; a sustained rate stays visible both as `warn` and as `503` in `vm0-request-log-prod`.

All five manifest-backed routes share the loader, so they share the `503` and it is declared on each in the contract. That deliberately includes `dmgDownload`, the migration wall's `Download Okou` target.

### Why `503` and not `204`

`204` would tell Squirrel "you are up to date", which we do not know. That silently swallows a real pending update. `503` is truthful, and on the automatic poll path the client only logs it — the "Unable to Check for Updates" dialog appears only on a manual check, which is the correct thing to show.

## Cache freshness and integrity

- The 60s fresh TTL is **unchanged**, so the happy path has no freshness regression.
- `staleUntil` is derived from `fetchedAt` at write time, and an entry is only written after a successful fetch. Neither a failed fetch nor a stale hit can extend it, so a long outage expires the copy into a `503` instead of serving one manifest forever. There is a test for exactly that.
- Worst case a client sees a manifest 30 minutes old and self-corrects on its next poll. This widens `blocked`-list propagation from `60s + poll` to `30min + poll`; `channels.stable.blocked` is currently empty. If we start relying on the blocklist, revisit that constant.

## Fallbacks

- `desktop-updates.service.ts:loadDesktopUpdateManifest` — answers from a manifest already fetched and validated when the upstream host cannot be read. Surface: API -> third-party release host, a state production has reached four times in 84 days. Bounded to 30 minutes measured from the fetch and never renewed by a stale hit. Remove if the manifest moves to a first-party host with an availability guarantee.

## Verification

Ran locally:
- `pnpm -F api check-types` — clean (all six projects).
- `pnpm -F @okouai/api-contracts check-types` — clean.
- `oxlint --deny-warnings` and `eslint --max-warnings 0` on every changed file — clean.
- `knip` — no new unused exports.
- `prettier --check` — clean.

**Not run locally:** Vitest. This sandbox has no Postgres and no Docker, and the API suite's global setup opens a DB connection before any test file loads, so it fails at `ECONNREFUSED 127.0.0.1:5432` regardless of the file selected. The tests below are new and **unproven until CI runs them** — please read CI as the first real execution.

New tests in `signals/routes/__tests__/desktop-updates.test.ts`:
- serves a release after transient upstream failures, exactly 3 attempts, nothing on the error channels;
- stops at the attempt bound and reports `provider_status: 500` in the `warn` event;
- `503` carries `Cache-Control: no-store` and `Retry-After: 60`;
- a tripped per-attempt deadline is retried then reported, with **no** `provider_status` (no response ever arrived);
- feed, `release`, and the wall's `dmg` all report `503` together;
- serves the cached release inside the window and logs `served_stale` with `stale_age_ms`;
- expires the cache at 30 minutes instead of renewing it across repeated stale hits;
- a failed refresh neither evicts nor ages the entry, and a later success replaces it;
- upstream `404` and an unparseable manifest still `500`, hit Sentry, and are **not** retried;
- a broken manifest is not answered from a warm cache.

Plus two in `lib/__tests__/log.test.ts` for the event-root lifting and its rejection of an unknown outcome.

### Not covered

Caller cancellation mid-fetch. The code routes it through `settle(…, signal)` so an `AbortError` propagates as a cancellation rather than becoming a `503`, but I did not find a non-flaky way to drive a mid-flight client abort through this route's harness, so that path is argued from the code, not tested.

### Post-deploy acceptance

On the release containing this, over ~7 days on `/api/desktop/updates/:product/:channel/:platform/:arch/RELEASES.json`:
- `status == 500` → 0 in `vm0-request-log-prod`;
- no new `API-11` event whose message ends in a `5xx` (a new `404`-class event is still expected and still correct);
- `200`/day at or above the 662–790 baseline; `503` ≈ 0.

Honest limit: at ~1 transient event per 3 weeks, a 7-day window will most likely observe **zero**. That window can accept the silencing, not the recovery. Confirming recovery in production needs ~30 days or a deliberate fault injection on a preview deployment.

### DIFF
--- a/turbo/apps/api/src/lib/__tests__/log.test.ts
+++ b/turbo/apps/api/src/lib/__tests__/log.test.ts
@@ -370,6 +370,75 @@ describe("Axiom log source field", () => {
     });
   });
 
+  it("lifts desktop update manifest outcome fields into the Axiom event root", () => {
+    const log = logger("desktop-update-manifest-test");
+    log.warn("Desktop update manifest upstream unavailable", {
+      type: "desktop_update_manifest_upstream",
+      outcome: "unavailable",
+      provider: "github_release_asset",
+      provider_status: 503,
+      failure_class: "transient_read_exhausted",
+      attempts: 3,
+      line: "ai-okou-desktop",
+      method: "GET",
+      route:
+        "/api/desktop/updates/:product/:channel/:platform/:arch/RELEASES.json",
+    });
+
+    expect(axiomLogging.warn).toHaveBeenCalledWith(
+      "Desktop update manifest upstream unavailable",
+      {
+        type: "desktop_update_manifest_upstream",
+        outcome: "unavailable",
+        provider: "github_release_asset",
+        provider_status: 503,
+        failure_class: "transient_read_exhausted",
+        attempts: 3,
+        line: "ai-okou-desktop",
+        method: "GET",
+        route:
+          "/api/desktop/updates/:product/:channel/:platform/:arch/RELEASES.json",
+        context: "desktop-update-manifest-test",
+        [EVENT]: {
+          source: "api",
+          type: "desktop_update_manifest_upstream",
+          outcome: "unavailable",
+          provider: "github_release_asset",
+          provider_status: 503,
+          failure_class: "transient_read_exhausted",
+          attempts: 3,
+          line: "ai-okou-desktop",
+          method: "GET",
+          route:
+            "/api/desktop/updates/:product/:channel/:platform/:arch/RELEASES.json",
+        },
+      },
+    );
+  });
+
+  it("does not lift a desktop update manifest event with an unknown outcome", () => {
+    const log = logger("desktop-update-manifest-outcome-test");
+    log.warn("manifest event", {
+      type: "desktop_update_manifest_upstream",
+      outcome: "degraded",
+      provider: "github_release_asset",
+      failure_class: "transient_read",
+      attempts: 1,
+      line: "ai-okou-desktop",
+    });
+
+    expect(axiomLogging.warn).toHaveBeenCalledWith("manifest event", {
+      type: "desktop_update_manifest_upstream",
+      outcome: "degraded",
+      provider: "github_release_asset",
+      failure_class: "transient_read",
+      attempts: 1,
+      line: "ai-okou-desktop",
+      context: "desktop-update-manifest-outcome-test",
+      [EVENT]: { source: "api" },
+    });
+  });
+
   it("does not lift unknown type fields into the Axiom event root", () => {
     const log = logger("unknown-type-test");
     log.error("custom event", {
--- a/turbo/apps/api/src/lib/error.ts
+++ b/turbo/apps/api/src/lib/error.ts
@@ -53,6 +53,10 @@ export function eventDeliveryUnavailable(message: string) {
   return httpError(503, "EVENT_DELIVERY_UNAVAILABLE", message);
 }
 
+export function desktopUpdateUnavailable(message: string) {
+  return httpError(503, "DESKTOP_UPDATE_UNAVAILABLE", message);
+}
+
 export function insufficientCredits() {
   return httpError(
     402,
--- a/turbo/apps/api/src/lib/log.ts
+++ b/turbo/apps/api/src/lib/log.ts
@@ -190,6 +190,59 @@ interface ProviderUnavailableRootFields {
   readonly route: string;
 }
 
+type DesktopUpdateManifestOutcome =
+  | "retry_recovered"
+  | "served_stale"
+  | "unavailable";
+
+interface DesktopUpdateManifestRootFields {
+  readonly type: "desktop_update_manifest_upstream";
+  readonly outcome: DesktopUpdateManifestOutcome;
+  readonly provider: "github_release_asset";
+  readonly provider_status?: number;
+  readonly failure_class: "transient_read" | "transient_read_exhausted";
+  readonly attempts: number;
+  readonly stale_age_ms?: number;
+  readonly line: string;
+  readonly method?: string;
+  readonly route?: string;
+}
+
+function isDesktopUpdateManifestOutcome(
+  value: unknown,
+): value is DesktopUpdateManifestOutcome {
+  return (
+    value === "retry_recovered" ||
+    value === "served_stale" ||
+    value === "unavailable"
+  );
+}
+
+function isDesktopUpdateManifestFailureClass(
+  value: unknown,
+): value is DesktopUpdateManifestRootFields["failure_class"] {
+  return value === "transient_read" || value === "transient_read_exhausted";
+}
+
+function isPositiveInteger(value: unknown): value is number {
+  return Number.isInteger(value) && Number(value) >= 1;
+}
+
+function isOptionalNonNegativeInteger(value: unknown): value is number {
+  return value === undefined || (Number.isInteger(value) && Number(value) >= 0);
+}
+
+function isOptionalUpstreamStatus(value: unknown): value is number {
+  return (
+    value === undefined ||
+    (Number.isInteger(value) && Number(value) >= 400 && Number(value) <= 599)
+  );
+}
+
+function isOptionalString(value: unknown): value is string {
+  return value === undefined || typeof value === "string";
+}
+
 function usageUnderbillingRootFields(
   fields: Record<string, unknown>,
 ): UsageUnderbillingRootFields | null {
@@ -277,17 +330,64 @@ function providerUnavailableRootFields(
   };
 }
 
+function desktopUpdateManifestRootFields(
+  fields: Record<string, unknown>,
+): DesktopUpdateManifestRootFields | null {
+  const type = fields.type;
+  const outcome = fields.outcome;
+  const provider = fields.provider;
+  const providerStatus = fields.provider_status;
+  const failureClass = fields.failure_class;
+  const attempts = fields.attempts;
+  const staleAgeMs = fields.stale_age_ms;
+  const line = fields.line;
+  const method = fields.method;
+  const route = fields.route;
+
+  if (
+    type !== "desktop_update_manifest_upstream" ||
+    !isDesktopUpdateManifestOutcome(outcome) ||
+    provider !== "github_release_asset" ||
+    !isOptionalUpstreamStatus(providerStatus) ||
+    !isDesktopUpdateManifestFailureClass(failureClass) ||
+    !isPositiveInteger(attempts) ||
+    !isOptionalNonNegativeInteger(staleAgeMs) ||
+    typeof line !== "string" ||
+    !isOptionalString(method) ||
+    !isOptionalString(route)
+  ) {
+    return null;
+  }
+
+  return {
+    type,
+    outcome,
+    provider,
+    ...(providerStatus === undefined
+      ? {}
+      : { provider_status: providerStatus }),
+    failure_class: failureClass,
+    attempts,
+    ...(staleAgeMs === undefined ? {} : { stale_age_ms: staleAgeMs }),
+    line,
+    ...(method === undefined ? {} : { method }),
+    ...(route === undefined ? {} : { route }),
+  };
+}
+
 function rootEventFields(
   fields: Record<string, unknown>,
 ):
   | UsageUnderbillingRootFields
   | UnhandledRequestErrorRootFields
   | ProviderUnavailableRootFields
+  | DesktopUpdateManifestRootFields
   | null {
   return (
     usageUnderbillingRootFields(fields) ??
     unhandledRequestErrorRootFields(fields) ??
-    providerUnavailableRootFields(fields)
+    providerUnavailableRootFields(fields) ??
+    desktopUpdateManifestRootFields(fields)
   );
 }
 
--- a/turbo/apps/api/src/signals/routes/__tests__/desktop-updates.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/desktop-updates.test.ts
@@ -1,3 +1,4 @@
+import { EVENT } from "@axiomhq/logging";
 import { desktopUpdatesContract } from "@okouai/api-contracts/contracts/desktop-updates";
 import { testDesktopUpdateManifestStateContract } from "@okouai/api-contracts/contracts/test-desktop-update-manifest-state";
 import { HttpResponse, http } from "msw";
@@ -413,6 +414,17 @@ describe("desktop update routes", () => {
     );
   }
 
+  // The logger attaches the emitting context and lifts the recognized root
+  // fields into the Axiom event root, so asserting the whole record also
+  // proves the event was promoted rather than passed through unrecognized.
+  function loggedManifestEvent(fields: Record<string, unknown>) {
+    return {
+      ...fields,
+      context: "DesktopUpdates",
+      [EVENT]: { source: "api", ...fields },
+    };
+  }
+
   function manifestLogFields(
     calls: readonly (readonly unknown[])[],
   ): readonly Record<string, unknown>[] {
@@ -464,7 +476,7 @@ describe("desktop update routes", () => {
     expect(
       manifestLogFields(context.mocks.axiomLogging.warn.mock.calls),
     ).toStrictEqual([
-      {
+      loggedManifestEvent({
         type: "desktop_update_manifest_upstream",
         outcome: "unavailable",
         provider: "github_release_asset",
@@ -475,7 +487,7 @@ describe("desktop update routes", () => {
         method: "GET",
         route:
           "/api/desktop/updates/:product/:channel/:platform/:arch/RELEASES.json",
-      },
+      }),
     ]);
   });
 
@@ -521,7 +533,7 @@ describe("desktop update routes", () => {
     expect(
       manifestLogFields(context.mocks.axiomLogging.warn.mock.calls),
     ).toStrictEqual([
-      {
+      loggedManifestEvent({
         type: "desktop_update_manifest_upstream",
         outcome: "unavailable",
         provider: "github_release_asset",
@@ -531,7 +543,7 @@ describe("desktop update routes", () => {
         method: "GET",
         route:
           "/api/desktop/updates/:product/:channel/:platform/:arch/RELEASES.json",
-      },
+      }),
     ]);
   });
 
@@ -583,7 +595,7 @@ describe("desktop update routes", () => {
       expect(
         manifestLogFields(context.mocks.axiomLogging.debug.mock.calls),
       ).toStrictEqual([
-        {
+        loggedManifestEvent({
           type: "desktop_update_manifest_upstream",
           outcome: "served_stale",
           provider: "github_release_asset",
@@ -591,7 +603,7 @@ describe("desktop update routes", () => {
           attempts: 3,
           stale_age_ms: 30 * 60_000 - 1,
           line: "ai-okou-desktop",
-        },
+        }),
       ]);
     });
   });
--- a/turbo/apps/api/src/signals/routes/__tests__/desktop-updates.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/desktop-updates.test.ts
@@ -366,6 +366,353 @@ describe("desktop update routes", () => {
     expect(response.body.error.code).toBe("NOT_FOUND");
   });
 
+  // The manifest lives on a host we do not control. A request that only
+  // failed to read it must not look like "no release exists": a Squirrel
+  // updater treats 404 as up to date and the migration wall's download button
+  // would dead-end. The cases below pin that separation, and pin that a
+  // genuinely broken manifest still fails loudly.
+  function feedRequest() {
+    return client().productFeed({
+      params: {
+        product: "ai-okou-desktop",
+        channel: "stable",
+        platform: "darwin",
+        arch: "arm64",
+      },
+    });
+  }
+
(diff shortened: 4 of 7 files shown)


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: c3429960045d64bf9a862601075f41d0c1ba8aca
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `c3429960045d64bf9a862601075f41d0c1ba8aca`. Practice documents read from `main` at `e5e52075f4f2285ebaf93bd62b61465d91f05b9e`.

### Summary

Splits desktop update manifest failures into "the release host could not be read" (bounded retry with a per-attempt deadline, then a 30-minute stale cache, then `503`) and "the manifest is missing or invalid" (unchanged `500` + Sentry). Reports the absorbed outcomes as a structured `desktop_update_manifest_upstream` event family at `debug`, and the exhausted one at `warn`, so a routine upstream blip stops producing an `unhandled_request_error`.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | All five manifest-backed routes share `loadDesktopUpdateManifest`; each is updated and each has the `503` declared. The only non-API consumer is Squirrel via `autoUpdater.setFeedURL` (`desktop-auto-updates.ts`), which reads the URL directly, not the ts-rest client. |
| `REVIEW.md` | Cancellation, ordering, and failure outcomes | Pass with observation | `settle(fetch(...), signal)` is given the caller's signal while the fetch gets `AbortSignal.any([signal, timeout])`, so caller abort re-throws as `AbortError` and only the deadline is retried. Not covered by a test — see Observations. |
| `REVIEW.md` | Meaningful coverage for new behavior | Pass | 12 new route cases plus 2 log-layer cases; each failure class, both boundaries of the stale window, and the non-retry of `404` are pinned. |
| `REVIEW.md` | Failure outcomes must not be flattened | Pass | `404`, Zod schema failure, product mismatch, and now `SyntaxError` on the body all still throw. `desktop-updates.test.ts` asserts `attempts() === 1` for `404`, proving no retry. |
| `REVIEW.md` | New fallbacks declared in a `Fallbacks` PR-body section | Pass | Declared with surface, bound, and removal condition. See Fallbacks below. |
| `docs/bad-smell.md` | §3 Catch only at a boundary that handles meaningfully; preserve legitimate retry/recovery | Pass | No log-and-rethrow. `settleManifestLoad` translates exactly one error class and rethrows everything else. |
| `docs/bad-smell.md` | §3 No fabricated defaults hiding a broken invariant | Pass | The stale path returns a manifest that already passed schema and `product` validation at fetch time; nothing is synthesized. A broken manifest is explicitly excluded from the cache path. |
| `docs/bad-smell.md` | §4 Document changed public contracts and assess consumers | Pass | `503` added to all five routes with a contract-level doc comment. `grep desktopUpdatesContract` finds no consumer outside `apps/api` and its test. |
| `docs/bad-smell.md` | §1/§2 No `any`, no lint/type suppressions | Pass | No `any` and no suppression comments in the diff. |
| `docs/bad-smell.md` | §6 No hardcoded URLs/config | N/A | The manifest URLs are pre-existing; the diff adds only tuning constants, each with a documented rationale. |
| `docs/bad-smell.md` | §7 Fallback patterns | Pass | See `docs/fallback.md` rows. |
| `docs/fallback.md` | §6 A fallback is justified only for a real, currently-reachable state | Pass | Production evidence: 4 transient upstream failures in 84 days (Sentry `API-11`), 2 within the retained Axiom window. This is §6.2 territory — genuinely unavailable third-party data, not a compatibility branch. |
| `docs/fallback.md` | §8 A bounded fallback names its surface and removal condition in code and PR body | Pass | Constant carries the rationale and the anchoring rule; PR body carries surface, bound, and removal condition. |
| `docs/fallback.md` | §4 Do not fabricate data for corrupt state | Pass | Corrupt/missing manifests are excluded from the stale path by an explicit guard, with a test that a warm cache does not rescue a `404`. |
| `docs/fallback.md` | §1 No negative tests that only pin removed behavior | Pass | Every new test pins live behavior. |
| `docs/deployment-compatibility.md` | Additive changes at cross-version boundaries | Pass | Adding a response status is additive; no old client is required to read it. Squirrel already treats every non-2xx as an error, so `503` lands on the same client path `500` did. |
| `docs/deployment-compatibility.md` | No one-shot protocol flips | Pass | No request field added or response field removed; no client change ships with this. |
| `docs/testing.md` | Test the external contract, not internals | Pass | Every case asserts HTTP status, body, or headers through `setupApp()`/the contract or `createApp().request()`. |
| `docs/testing.md` | Mock external services at their boundary; MSW for HTTP | Pass | MSW only; no `vi.mock()` of internal modules. |
| `docs/testing.md` | Avoid fake timers | Pass with observation | No fake timers. The per-attempt deadline test substitutes `AbortSignal.timeout` through the centralized registry — see Observations. |
| `docs/testing/api-testing.md` | Route tests go through `setupApp()` + contract, not handler/service imports | Pass | Uses the existing `client()`/`appRequest()` helpers; no service is imported. |
| `docs/testing/api-testing.md` | Cache assertions own their key and advance mocked time inside the owning test | Pass | `beforeEach` resets the manifest cache through the existing test-only route; every TTL/stale case wraps its own `withMockNowForTest` and calls `mockNow` inside itself. |
| `docs/testing/api-testing.md` | Mock only auth identity and external services via `context.mocks` | Pass | `axiomLogging` and `sentry` are external sinks in the centralized registry, and `api-testing.md` lists "external service calls at the boundary, using the centralized mocks" as route-test scope. |
| `docs/testing/testing-external-behavior.md` | No DB/service reads or writes in API tests | Pass | Neither appears in the diff. |
| `docs/testing/testing-external-behavior.md` | Mock calls are not a substitute for observable outcomes | Pass | Every log assertion sits alongside a status/body/header assertion; none stands alone as the outcome. |

### Findings

No P0 or P1 findings.

1. **`AbortSignal.timeout` outlives a fast attempt.** Each attempt allocates a 3s deadline signal even when the fetch returns in ~250ms. Node's timer is unref'd so it cannot hold the function open, but it is a small per-request allocation on a route serving ~700 requests/day. Not worth complicating the code for.
2. **A sustained outage re-runs the full retry budget on every poll before falling back to stale.** That is ~3x upstream requests and ~400ms extra latency per poll while stale-serving. At the observed ~0.4 req/min this is immaterial; revisit only if the release host becomes rate-limit sensitive. Adding a negative-cache field would trade that for another piece of cache state, which is the wrong trade at this volume.
3. **Caller cancellation mid-fetch is argued from code, not covered.** The `settle(..., signal)` split is correct by inspection and by `utils.ts:throwIfAbort`, but no test drives a mid-flight client abort through this route. Acknowledged in the PR body rather than papered over.
4. **The deadline test substitutes `AbortSignal.timeout`.** This controls a runtime primitive rather than driving the real deadline, which is a small step toward implementation coupling. It uses the registry the harness provides for exactly this, keeps the suite off fake timers, and avoids a ~10s test; the assertion is still the external `503` plus the absence of `provider_status`. Reasonable trade, worth knowing it is there.

### Fallbacks

| Fallback | Surface | Window | Removal condition | Declared | Verdict |
| --- | --- | --- | --- | --- | --- |
| `desktop-updates.service.ts:loadDesktopUpdateManifest` serves an already-validated cached manifest when the upstream host cannot be read | API -> third-party release host (`github.com` release assets) | 30 minutes, anchored to `fetchedAt`, never renewed by a stale hit | Manifest moves to a first-party host with an availability guarantee | Yes, in the PR body `Fallbacks` section | **Justified** |

Justified under `docs/fallback.md` §6.2: the unavailable state is real and reached in production four times in 84 days, not a state the contract or a single writer prevents. The bound is the part that makes it safe — because `staleUntil` is derived from `fetchedAt` at write time and only a successful fetch writes an entry, neither a failed fetch nor a run of stale hits can extend it, so a long outage degrades into a truthful `503` instead of serving one manifest indefinitely. `expires the cached release instead of renewing it on each stale hit` pins exactly that. The residual risk is `blocked`-list propagation widening from `60s + poll` to `30min + poll`; `channels.stable.blocked` is currently empty and the constant's comment flags the dependency.

### Testing

Coverage is proportionate and targets the behavior that changed rather than the shape of the code.

Route cases (`signals/routes/__tests__/desktop-updates.test.ts`): retry-then-succeed with an exact attempt count; attempt bound reached with `provider_status` in the `warn` event; `503` carrying `Cache-Control: no-store` and `Retry-After: 60`; tripped per-attempt deadline retried then reported with no `provider_status`; feed, `release`, and the migration wall's `dmg` all degrading together; stale served inside the window with `stale_age_ms`; the 30-minute boundary enforced across repeated stale hits; a failed refresh neither evicting nor ageing the entry while a later success replaces it; `404`, unparseable JSON, and a schema-invalid manifest each staying `500`, reaching Sentry, and not being retried; and a warm cache not rescuing a `404`.

Log cases (`lib/__tests__/log.test.ts`): event-root lifting, and rejection of an unknown `outcome` so a malformed event is not silently promoted.

The negative controls are the important part here, and they are present: the same code path that now absorbs `5xx` is proven to still fail loudly for every integrity failure, including the `SyntaxError` case, which is the one most easily misfiled as a transport error.

**Verification gap, stated plainly:** the author could not run Vitest locally — the sandbox has no Postgres and no Docker, and the API suite's global setup opens a DB connection before any test file loads, so every selection fails at `ECONNREFUSED 127.0.0.1:5432`. `check-types` (both workspaces), `oxlint --deny-warnings`, `eslint --max-warnings 0`, `knip`, and `prettier --check` were run and are clean. **These tests have never executed.** CI is their first run, and this LGTM is contingent on `test-api` passing; a green review here is not evidence the assertions hold.

The post-deploy acceptance criteria in the PR body are correctly scoped, including the honest note that a 7-day window at ~1 transient event per 3 weeks can accept the silencing but not the recovery.

lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: 5836acc7f7107ed849a62716915bdd8e86511074
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `5836acc7f7107ed849a62716915bdd8e86511074`. Practice documents read from `main` at `e5e52075f4f2285ebaf93bd62b61465d91f05b9e`.

The full criteria audit is in [the review of `c342996`](https://github.com/vm0-ai/vm0/pull/33127#issuecomment-5611943185). The delta since that HEAD is test-only — 18 insertions, 6 deletions in `signals/routes/__tests__/desktop-updates.test.ts`, no production file touched — so this re-review covers the delta and re-affirms the prior verdict rather than restating unchanged rows.

### Summary

Corrects three log assertions that listed only the fields the call site passes. `log.ts:logToAxiom` also attaches `context` and lifts recognized fields into the `Symbol(logging.event)` root, so `toStrictEqual` failed on the two `unavailable` cases and the `served_stale` case in `test-api (4)`.

### Review Criteria (delta)

| --- | --- | --- | --- |
| `REVIEW.md` | Distinguish an introduced defect from a wrong expectation | Pass | The three CI failures were assertion-shape errors, not behavior errors: the received records already carried the correct `outcome`, `failure_class`, `attempts`, `provider_status`, `line`, `route`, and `stale_age_ms: 1799999`. No production change was warranted and none was made. |
| `REVIEW.md` | A rewritten expectation must not weaken the check | Pass | The expectation was widened, not loosened. `loggedManifestEvent()` still asserts via `toStrictEqual`, and now additionally pins `context` and the full `[EVENT]` root. |
| `docs/testing.md` | Assert externally observable outcomes, not incidental shape | Pass | Asserting the `[EVENT]` root is the meaningful half: it proves `desktopUpdateManifestRootFields` accepted the payload and promoted it, which is precisely the behavior this PR adds. A payload that failed validation would emit `[EVENT]: { source: "api" }` and now fails the test. |
| `docs/testing/anti-patterns.md` (via `docs/testing.md` routing) | Do not reshape an assertion merely to make it pass | Pass | The prior expectation could not have caught a regression in event-root lifting; the new one can. `lib/__tests__/log.test.ts` already used this same full-record shape, so this aligns the route tests with the established convention rather than inventing one. |
| `docs/fallback.md` | No new fallback introduced | Pass | Test-only delta. |
| `docs/deployment-compatibility.md` | No contract or persisted-state change | Pass | Test-only delta. |

### Findings

No P0 or P1 findings. The four non-blocking observations from the prior review still stand unchanged.

### Testing

`test-api (4)` was the only genuine failure on `c342996`; shards 3, 5, 7, and 8 were fail-fast **cancelled**, and `ci-gate-turbo` is the aggregate gate, so all six red checks trace to one cause. That cause is fixed here.

Worth recording honestly: CI was the first execution of these tests, and it found a real defect in them. The author's local environment has no Postgres or Docker, so the API suite's global setup cannot connect and no local Vitest run was possible; `check-types`, `oxlint`, `eslint`, `knip`, and `prettier` were re-run clean on this HEAD. This LGTM remains contingent on `test-api` going green.

lancy:
Replacement PR: #33214

#33127 was reopened under the approved recovery path, then externally closed again before it could return to the normal protected queue. #33214 uses the same branch and preserves the same #33084 scope, design, and verification boundary.

Issue #33071, opened 2026-09-10
bug(pi): classify and recover model responses with no healthy upstream


seven332:
## Summary

classify and recover model responses with no healthy upstream. The current-window observations and their causal limits are recorded below.

## Impact

One sandbox Pi run failed with exit 1 and two terminal wrappers.

Severity: **high**. Type: **upstream/provider failure**. Severity is triage, not a raw-event-to-user equivalence.

### Coverage

- Beijing: **[2026-09-09 00:00, 2026-09-10 00:00) Asia/Shanghai**. Exact UTC: **[2026-09-08T16:00:00Z, 2026-09-09T16:00:00Z)**.
- Axiom: ONLY `vm0-web-logs-prod`, `level == "error" or level == "warn"`; **2 records**, first `2026-09-09T00:41:40.872Z`; last `2026-09-09T00:41:40.952Z`. Levels: error: 1; warn: 1.
- Distinct logged run IDs: **1**. 0 Axiom records lack a run ID; these do not imply additional runs.
- Per-run record distribution: run-136 each has {"job execution failed":1,"Run failed":1}.
- Correlation: Matching private run IDs join terminal cascades; different native error classes remain separate problems.
- Privacy: these are selected original source fields. Private IDs, device tokens, bound SQL parameters, custom workflow paths, signed/private URLs and unrelated user content are redacted or omitted. No request bodies, breadcrumbs, raw user identities or private Sentry links are published.

### Variant 1 — job execution failed (1 records)

- Source: Axiom; first `2026-09-09T00:41:40.872Z`; last `2026-09-09T00:41:40.872Z`.
- Native causal fields: `error`, `failure_class`, `failure_cli_exit_code`, `failure_detail_source`, `failure_framework`.
- Runner release distribution: 0.188.17: 1.
- failure_class: cli_nonzero: 1.
- failure_framework: pi: 1.

Original sanitized sample 1:

```json
{
  "_time": "2026-09-09T00:41:40.872Z",
  "cli_observed_exit_code": 0,
  "cli_observed_exit_kind": "exit",
  "cli_observed_mapped_exit_code": 0,
  "context": "runner::cmd::start::job_terminal_log",
  "error": "no healthy upstream",
  "exit_code": 1,
  "failure_class": "cli_nonzero",
  "failure_cli_exit_code": 1,
  "failure_detail_source": "pi_result",
  "failure_framework": "pi",
  "level": "error",
  "message": "job execution failed",
  "reused": true,
[...]

Issue #33067, opened 2026-09-10
bug(pi): preserve actionable diagnostics when a model returns an HTML error page


seven332:
## Summary

preserve actionable diagnostics when a model returns an HTML error page. The current-window observations and their causal limits are recorded below.

## Impact

Two Pi runs failed with exit 1; four terminal records contain the same truncated HTML instead of an actionable provider error.

Severity: **high**. Type: **upstream/provider failure with diagnostic loss**. Severity is triage, not a raw-event-to-user equivalence.

### Coverage

- Beijing: **[2026-09-09 00:00, 2026-09-10 00:00) Asia/Shanghai**. Exact UTC: **[2026-09-08T16:00:00Z, 2026-09-09T16:00:00Z)**.
- Axiom: ONLY `vm0-web-logs-prod`, `level == "error" or level == "warn"`; **4 records**, first `2026-09-08T17:40:36.344Z`; last `2026-09-08T17:43:21.314Z`. Levels: error: 2; warn: 2.
- Distinct logged run IDs: **2**. 0 Axiom records lack a run ID; these do not imply additional runs.
- Per-run record distribution: run-126, run-127 each has {"job execution failed":1,"Run failed":1}.
- Correlation: Matching private run IDs join terminal cascades; different native error classes remain separate problems.
- Privacy: these are selected original source fields. Private IDs, device tokens, bound SQL parameters, custom workflow paths, signed/private URLs and unrelated user content are redacted or omitted. No request bodies, breadcrumbs, raw user identities or private Sentry links are published.

### Variant 1 — job execution failed (2 records)

- Source: Axiom; first `2026-09-08T17:40:36.344Z`; last `2026-09-08T17:43:21.179Z`.
- Native causal fields: `error`, `failure_class`, `failure_cli_exit_code`, `failure_detail_source`, `failure_framework`.
- Runner release distribution: 0.188.16: 2.
- failure_class: cli_nonzero: 2.
- failure_framework: pi: 2.

Original sanitized sample 1:

```json
{
  "_time": "2026-09-08T17:40:36.344Z",
  "cli_observed_exit_code": 0,
  "cli_observed_exit_kind": "exit",
  "cli_observed_mapped_exit_code": 0,
  "context": "runner::cmd::start::job_terminal_log",
[...]

PR #33129, opened 2026-09-10
fix(pi): preserve provider http status in opaque model failures

lancy:
## Problem

A gateway can answer a Pi model request with an opaque body instead of a provider error envelope — an Envoy local reply (`no healthy upstream`) or an edge HTML error page. The OpenAI Responses adapters reduce such a body to its raw text and drop the HTTP status, which breaks three things at once:

1. The terminal assistant error carries no provider stage evidence at all.
2. `AgentSession` auto-retry is enabled in the sandbox (3 attempts, 2s backoff) but classifies retryability from that text, and `isRetryableAssistantError("no healthy upstream")` is `false`. A single transient gateway answer therefore ended the whole run on its first hit.
3. `classify_cli_failure_reason` had no `AgentFramework::Pi` branch, so the result stayed unclassified. The runner logged `job execution failed` at `error` and the API logged `Run failed` at `warn`, and the user saw the bare Envoy string.

Production evidence for #33071: one `web` run on `gpt-5.6-sol` via `codex-oauth-token` failed after already producing an assistant message, with `failure_class=cli_nonzero`, `failure_framework=pi`, `failure_detail_source=pi_result`, `error="no healthy upstream"` and no `failure_reason`. In the same 24h window, all 68 `pi` + `cli_nonzero` terminal failures had an empty `failure_reason`.

## Change

- New `provider-error-body.ts` boundary: when a Pi model response is not OK and its body is **not** valid JSON, restate it as the adapter's own envelope, `{"error":{"message":"provider HTTP <status>: <phrase>","type":"http_error"}}`. Markup is stripped, whitespace collapsed, and the phrase bounded to 200 chars. Bodies above 64 KiB pass through untouched.
- `piAgentStreamForConfig` installs that boundary for both OpenAI Responses routes, composed with the existing `onObservedResponseStatus` observer. The native Anthropic/Bedrock branch is untouched so it keeps `nativePublicFetch` and its DNS-rebinding protection.
- `classify_cli_failure_reason` gains a `Pi` + `PiResult` branch that reads the recorded status: `500/502/503/504` → `provider_server_error`, `529` → `provider_overloaded`. It is evaluated **after** the quota/subscription wording so `usage_limit` stays authoritative.

Nothing else changes: runner level policy (`is_info_level_job_failure`) and API failure-log suppression (`shouldSuppressKnownFailureLog`) already key on that reason, so a classified failure logs at `info` in the runner — below the `level <= WARN` Axiom ingest filter — and the API `warn` is suppressed for non-built-in providers while a built-in provider still warns. An exhausted recovery keeps a truthful failed run and now renders the shared transient-error copy instead of raw gateway text.

## Deliberately out of scope

- **Transport retries stay at `maxRetries: 0`.** Enabling them on the Codex adapter looked attractive, but its `validateRetryDelayMs` throws a bare `RetryDelayExceededError` when a `Retry-After` exceeds the cap, discarding the friendly ChatGPT usage-limit message the failure classifiers depend on — and that error text is itself retryable, so a quota-exhausted turn would burn the session budget. Recovery is left to the session retry budget, which the status marker now unblocks. A comment records this so it is not "fixed" later by accident.
- `401/402/429` and other statuses stay with their existing message-shaped rules rather than gaining an unevidenced status mapping.
- BYOK model-provider flows are still outside the mitm addon's `model_provider_failure` observation (`admit_flow` requires a billable firewall, and only built-in providers contribute one). That is a separate decision, not a blocker here.

This boundary also covers the HTML error-page signature tracked in #33067; that issue still needs its own verification and is not closed by this PR.

## Verification

Run locally:

- `vitest run` for `@okouai/pi-agent-runtime` — **416 passed / 25 files**, including 11 new boundary tests and 2 new adapter integration tests.
- `cargo test -p guest-agent --lib failure_diagnostics` — **106 passed**, including 4 new classification tests.
- `cargo clippy -p guest-agent --lib --all-targets` — clean; `cargo fmt -p guest-agent --check` — clean.
- `tsc --noEmit`, package `lint` (oxlint + type-aware oxlint + eslint), `tsc -p tsconfig.build.json`, and `check-public-declarations` — all clean.

New behavioral coverage:

- Positive: a 503 `no healthy upstream` from the Codex Responses route now ends with `errorMessage === "provider HTTP 503: no healthy upstream"` and `isRetryableAssistantError(result) === true`, so the session budget can recover it.
- Negative control: a 429 `usage_limit_reached` envelope keeps `"You have hit your ChatGPT usage limit (plus plan)."` verbatim, stays non-retryable, and issues exactly one request.
- Untouched envelopes: usage-limit, `insufficient_credits` and `token_refresh_failed` JSON bodies are byte-identical after the boundary.
- Rust negatives: wrong framework, wrong detail source, `401/402/429`, malformed statuses (`5031`, `503_beta`, empty), an embedded marker (`xprovider HTTP 503`), and bare `no healthy upstream` all stay unclassified; quota wording still wins over a 503 status.

Not run locally: `knip` (the filtered install lacks sibling workspace deps) and the wider Rust/TS suites — left to the PR pipeline.

## Post-deploy check

`pi-agent-runtime`/`okou` CLI and `runner` ship separately, so accept them separately. After both are live, over ≥72h on `vm0-web-logs-prod`: the share of `pi` + `cli_nonzero` `job execution failed` records with a non-empty `failure_reason` should rise from the current 0%; `no healthy upstream` error/warn records should reach 0 **while the matching runs stop reaching `run.failed`** (cross-check in the DB, so this is recovery rather than a hidden log); and a `built-in` provider failure must still produce the API `warn`.

Refs #33071

### DIFF
--- a/crates/guest-agent/src/failure_diagnostics.rs
+++ b/crates/guest-agent/src/failure_diagnostics.rs
@@ -24,6 +24,8 @@ use serde_json::Value;
 const LOG_TAG: &str = "sandbox:guest-agent";
 const MAX_LOGGED_CLI_STDERR_LINES: usize = 20;
 const MAX_LOGGED_CLI_STDERR_LINE_BYTES: usize = 4096;
+/// Lowercased marker the Pi model boundary writes before an observed status.
+const PI_PROVIDER_HTTP_MARKER: &str = "provider http ";
 const CODEX_SAFETY_POLICY_REFUSAL_MESSAGE: &str = concat!(
     "This content was flagged for possible cybersecurity risk. ",
     "If this seems wrong, try rephrasing your request. ",
@@ -379,7 +381,54 @@ fn classify_cli_failure_reason(
     {
         return Some(FailureReason::UsageLimit);
     }
+    // Quota and subscription wording above stays authoritative for the statuses
+    // it also covers, so the observed provider status is read last.
+    if matches!(framework, AgentFramework::Pi)
+        && let Some(reason) = pi_provider_http_failure_reason(source, &normalized)
+    {
+        return Some(reason);
+    }
     None
+}
+
+/// Classify the provider status the Pi runtime records for an opaque body.
+///
+/// The Pi model boundary restates a non-JSON provider error as
+/// `provider HTTP <status>: <phrase>`, so an Envoy or edge reply that used to
+/// arrive as bare prose now carries its stage. Only the server-unavailability
+/// family is classified here; every other status keeps whatever message-shaped
+/// classification already applies, or stays unknown.
+fn pi_provider_http_failure_reason(
+    source: FailureDetailSource,
+    normalized: &str,
+) -> Option<FailureReason> {
+    if source != FailureDetailSource::PiResult {
+        return None;
+    }
+    normalized
+        .match_indices(PI_PROVIDER_HTTP_MARKER)
+        .find_map(|(index, _)| {
+            if normalized[..index]
+                .chars()
+                .next_back()
+                .is_some_and(is_error_type_char)
+            {
+                return None;
+            }
+            let detail = &normalized[index + PI_PROVIDER_HTTP_MARKER.len()..];
+            let status_len = detail
+                .find(|c: char| !c.is_ascii_digit())
+                .unwrap_or(detail.len());
+            let (status, remaining) = detail.split_at(status_len);
+            if remaining.chars().next().is_some_and(is_error_type_char) {
+                return None;
+            }
+            match status {
+                "500" | "502" | "503" | "504" => Some(FailureReason::ProviderServerError),
+                "529" => Some(FailureReason::ProviderOverloaded),
+                _ => None,
+            }
+        })
 }
 
 fn is_codex_safety_policy_refusal(source: FailureDetailSource, failure_message: &str) -> bool {
--- a/crates/guest-agent/src/failure_diagnostics/tests.rs
+++ b/crates/guest-agent/src/failure_diagnostics/tests.rs
@@ -929,6 +929,120 @@ fn cli_failure_reason_rejects_untrusted_mid_response_failure_contexts() {
 }
 
 #[test]
+fn cli_failure_reason_classifies_pi_provider_http_unavailability() {
+    for (message, expected_reason) in [
+        (
+            "provider HTTP 503: no healthy upstream",
+            FailureReason::ProviderServerError,
+        ),
+        (
+            "provider HTTP 502: Service Unavailable",
+            FailureReason::ProviderServerError,
+        ),
+        ("provider HTTP 500", FailureReason::ProviderServerError),
+        (
+            "provider HTTP 504: upstream request timeout",
+            FailureReason::ProviderServerError,
+        ),
+        (
+            "provider HTTP 529: overloaded",
+            FailureReason::ProviderOverloaded,
+        ),
+        // The public Responses route prefixes the SDK status before the marker.
+        (
+            "503 provider HTTP 503: no healthy upstream",
+            FailureReason::ProviderServerError,
+        ),
+    ] {
+        let reason = super::classify_cli_failure_reason(
+            AgentFramework::Pi,
+            FailureDetailSource::PiResult,
+            message,
+        );
+
+        assert_eq!(reason, Some(expected_reason), "message: {message}");
+    }
+}
+
+#[test]
+fn cli_failure_reason_rejects_untrusted_pi_provider_http_contexts() {
+    const MESSAGE: &str = "provider HTTP 503: no healthy upstream";
+
+    for (framework, source) in [
+        (AgentFramework::Pi, FailureDetailSource::Stderr),
+        (AgentFramework::ClaudeCode, FailureDetailSource::PiResult),
+        (AgentFramework::Codex, FailureDetailSource::PiResult),
+    ] {
+        let reason = super::classify_cli_failure_reason(framework, source, MESSAGE);
+
+        assert_eq!(reason, None, "framework: {framework:?} source: {source:?}");
+    }
+
+    for message in [
+        // A status this boundary deliberately leaves to message-shaped rules.
+        "provider HTTP 401: unauthorized",
+        "provider HTTP 402: payment required",
+        "provider HTTP 429: slow down",
+        // Malformed or embedded markers must not be read as a status.
+        "provider HTTP 5031: no healthy upstream",
+        "provider HTTP 503_beta: no healthy upstream",
+        "provider HTTP : no healthy upstream",
+        "the tool printed provider HTTP",
+        "xprovider HTTP 503: no healthy upstream",
+        "no healthy upstream",
+    ] {
+        let reason = super::classify_cli_failure_reason(
+            AgentFramework::Pi,
+            FailureDetailSource::PiResult,
+            message,
+        );
+
+        assert_eq!(reason, None, "message: {message}");
+    }
+}
+
+#[test]
+fn cli_failure_reason_keeps_quota_wording_authoritative_over_pi_provider_http() {
+    let reason = super::classify_cli_failure_reason(
+        AgentFramework::Pi,
+        FailureDetailSource::PiResult,
+        "provider HTTP 503: You have hit your ChatGPT usage limit.",
+    );
+
+    assert_eq!(reason, Some(FailureReason::UsageLimit));
+}
+
+#[test]
+fn pi_provider_http_failure_reaches_the_structured_diagnostic() {
+    let msg = cli_failure_message(
+        1,
+        &["background stderr noise".to_string()],
+        Some(&cli_diagnostic(
+            "provider HTTP 503: no healthy upstream",
+            FailureDetailSource::PiResult,
+        )),
+    );
+    let diagnostic = FailureDiagnostic::new(
+        FailureClass::CliNonzero,
+        AgentFramework::Pi,
+        PromptMetadata::from_prompt("plain prompt"),
+    )
+    .with_cli_exit_code(1)
+    .with_failure_detail_source(msg.source);
+    let diagnostic = with_cli_failure_reason(diagnostic, &msg);
+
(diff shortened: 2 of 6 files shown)


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: 8574152a2617151788502a5619d37aae4ea88b85
PR_AUTO_REVIEW_VERDICT: LGTM

Reviewed HEAD: `8574152a2617151788502a5619d37aae4ea88b85`. Practice documents read from `main` at `e5e52075f4f2285ebaf93bd62b61465d91f05b9e`.

### Summary

Adds a fetch boundary in `pi-agent-runtime` that restates an opaque (non-JSON) provider error body as the adapter's own JSON envelope carrying the observed HTTP status, and a `Pi` + `PiResult` branch in the guest-agent that classifies the server-unavailability status family into `provider_server_error` / `provider_overloaded`. Together these let the already-enabled `AgentSession` retry budget recover a transient gateway answer such as an Envoy `no healthy upstream`, and give an exhausted failure a known reason instead of an unactionable production `error` record.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | Traced `provider-error-body.ts` → `model.ts:piAgentStreamForConfig` → `session-runtime.ts:70` (sandbox registration) and `session-memory.ts` (API first turn); then `pi_rpc.rs:PiAssistantTerminal::from_message` → `events.rs:extract_jsonl_result_failure_diagnostic` → `failure_diagnostics.rs:classify_cli_failure_reason` → `job_spawn.rs:completion_failure_reason` → `diagnostics.rs:799` → webhook `failureReason`. |
| `REVIEW.md` | Failure outcomes, ordering, and cancellation where relevant | Pass | Boundary only intercepts `!response.ok`, so no streaming/cancellation path changes. The Rust branch is evaluated last, after the quota/subscription block, so `usage_limit` stays authoritative — covered by `cli_failure_reason_keeps_quota_wording_authoritative_over_pi_provider_http`. |
| `REVIEW.md` | Meaningful coverage for new behavior; identify the missing scenario | Pass | 11 new boundary tests + 2 new adapter integration tests + 4 new Rust classification tests, including negative controls. See Testing. |
| `REVIEW.md` | Report concrete practice violations with rule and source; do not infer bugs from shape | Pass | No violations found; every row below is checked against the full diff rather than filenames. |
| `docs/bad-smell.md` §1 | No `any`; `unknown` with narrowing at untrusted boundaries | Pass | `provider-error-body.ts` uses `Uint8Array`/`string`/`Headers`; `isJsonBody` discards the parsed value rather than typing it. Tests narrow with explicit `as { error: { message: string } }` casts on their own fixtures only. |
| `docs/bad-smell.md` §2 | No lint/type suppressions | Pass | No `eslint-disable` / `@ts-expect-error` / `prettier-ignore` / `#[allow(...)]` in the diff. |
| `docs/bad-smell.md` §3 | Catch only at a boundary that handles it; preserve legitimate recovery | Pass | The single `try { JSON.parse } catch { return false }` in `isJsonBody` is a parse-shape probe, not a swallowed failure. The `finally { reader.releaseLock() }` preserves resource cleanup. `maxRetries: 0` is retained on the Codex adapter, so no existing recovery is removed. |
| `docs/bad-smell.md` §4 | Document changed public contracts and assess consumers | Pass | The `provider HTTP <status>` message shape is a new cross-artifact contract; both directions assessed below under deployment compatibility and stated in the PR body. `check-public-declarations` reports `entries=2; declarations=11; forbidden=0`, so the package's public surface is unchanged. |
| `docs/bad-smell.md` §5 | Static imports in production code | Pass | `model.ts` imports `preserveProviderErrorStatus` statically; no dynamic import added. |
| `docs/bad-smell.md` §6 | Centralized configuration; no invented URLs or credentials | N/A | The diff adds no environment-specific value, origin, or credential; base URLs continue to come from the existing model config. |
| `docs/bad-smell.md` §7 / `docs/fallback.md` §3-§6 | No fallback for a state the contract already rules out | Pass | No fallback introduced — see Fallbacks below. |
| `docs/fallback.md` §1 | No negative tests pinning removed behavior | Pass | Every negative case asserts a currently-reachable input (wrong framework/source, unmapped status, malformed marker, provider-authored envelope), not the absence of deleted code. |
| `docs/fallback.md` §9 | Every introduced fallback declared in the PR summary | Pass | The diff introduces none, so no `Fallbacks` section is required. |
| `docs/deployment-compatibility.md` | Runner protocol / cross-artifact changes cover coexisting old-new combinations | Pass | The producer is the npm-delivered `okou` CLI (`npx --package=$CLI_PACKAGE_URL okou __agent-loop`); the consumer is `guest-agent`, shipped with the runner. New CLI + old guest-agent: the marker is present but unmatched → `failure_reason` stays `None`, i.e. today's behavior. Old CLI + new guest-agent: no marker → `match_indices` finds nothing → `None`. Both directions degrade to current behavior, so no tolerance branch or gate is needed. |
| `docs/deployment-compatibility.md` | Persisted state stays readable across versions | Pass | `agent_runs.error` gains a longer message but no shape change; `formatRunErrorForExternalSurface` handles an unclassified legacy message exactly as before, and a classified one now routes through `formatStructuredRunError`. No migration involved. |
| `docs/testing.md` | Test the external contract through the production entry point | Pass | `model.test.ts` asserts through `piAgentStreamForConfig(...)` and `stream.result()` — the same seam `session-runtime.ts` registers — rather than calling the boundary helper directly. |
| `docs/testing.md` | Mock external services at their boundary; use real internal code | Pass | Only the provider HTTP response is substituted, through the adapter's own documented `options.fetch` seam, matching the existing `preserves a native Codex Responses tool call over forced SSE` test in the same file. No `vi.mock()` of internal modules; `isRetryableAssistantError` is the real upstream implementation. |
| `docs/testing.md` | Cover old/new combinations for protocol changes | Pass | Covered by construction rather than by a test: `cli_failure_reason_rejects_untrusted_pi_provider_http_contexts` asserts bare `no healthy upstream` (old-CLI shape) stays `None`, which is the old-producer direction. |
| `docs/testing.md` | Avoid duplicate cases restating a library or static configuration | Pass | The runner INFO-level policy is deliberately not re-tested; `expected_cli_failure_reasons_log_job_execution_failed_at_info` in `job_terminal_log.rs` already covers `ProviderServerError` and `ProviderOverloaded`. |
| `docs/testing/rust-testing.md` | Inline `#[cfg(test)]` for module-internal logic | Pass | `classify_cli_failure_reason` and `pi_provider_http_failure_reason` are private, and the new tests live in the existing `failure_diagnostics/tests.rs` module beside comparable Claude/Codex cases. |

### Findings

No P0 or P1 findings.

Non-blocking observations (no action requested):

- `providerErrorSnippet` strips whole tags, so an HTML page's attribute text is dropped along with markup. That is the desired outcome here and is pinned by the `viewport` assertion, but it does mean a provider that puts its only diagnostic inside an attribute would reduce to the status alone. Acceptable for an opaque-body path.
- `401`, `402` and `429` are intentionally left unmapped so the existing message-shaped rules keep ownership. This means a genuinely transient `429` without quota wording still logs at `error`; that is unchanged behavior, not a regression, and the PR body states it as out of scope.
- The `maxRetries: 0` comment on `piAgentCodexStream` documents a decision not taken. It is accurate — `validateRetryDelayMs` throws a bare `RetryDelayExceededError`, whose text matches the retryable pattern — and worth keeping so the value is not "fixed" later without that context.

### Fallbacks

The diff introduces no fallback, so no `Fallbacks` section is required in the PR summary.

Three pass-through branches were assessed against `docs/fallback.md` §3-§6 and none is a fallback:

- `response.ok` — the primary success path, not an extra path for an unexpected state.
- `response.body === null || NULL_BODY_STATUSES.has(response.status)` — a correctness guard for a reachable HTTP state; `new Response(body, { status: 204 })` throws, so this is required behavior rather than a default hiding a broken invariant.
- `truncated || isJsonBody(text)` — the explicit primary contract of the boundary (never rewrite a provider-authored envelope; bound the buffer), not tolerance for a retired producer.

### Testing

Coverage added:

- `provider-error-body.test.ts` (11 cases): Envoy plain-text 503 restated with its status; HTML page reduced to one phrase with script/style/attribute text removed; empty body yields the status alone; oversized phrase bounded and suffixed; three provider-authored envelopes (`usage limit`, `insufficient_credits`, `token_refresh_failed`) returned byte-identical; >64 KiB body passed through; `200`/`201` untouched; `304` bodyless status preserved.
- `model.test.ts` (2 cases): through `piAgentStreamForConfig`, a 503 `no healthy upstream` produces `errorMessage === "provider HTTP 503: no healthy upstream"` with `isRetryableAssistantError(result) === true` — the positive recovery case — and a 429 `usage_limit_reached` envelope keeps `"You have hit your ChatGPT usage limit (plus plan)."` verbatim, stays non-retryable, and issues exactly one request — the negative control that proves the boundary did not make a quota state look transient.
- `failure_diagnostics/tests.rs` (4 cases): the mapped status family including the SDK-prefixed `503 provider HTTP 503: ...` shape; rejection of wrong framework, wrong detail source, unmapped statuses, malformed statuses (`5031`, `503_beta`, empty), an embedded marker (`xprovider HTTP 503`) and the bare legacy message; quota wording winning over a 503; and the reason reaching the structured `FailureDiagnostic`.

Verification observed in the PR body and reproducible from the diff: `@okouai/pi-agent-runtime` `vitest run` 416 passed / 25 files; `cargo test -p guest-agent --lib failure_diagnostics` 106 passed; `cargo clippy -p guest-agent --lib --all-targets` and `cargo fmt --check` clean; `tsc --noEmit`, package `lint`, `tsc -p tsconfig.build.json`, and `check-public-declarations` clean.

Material gaps: none that block. `knip` was not run locally (the author's filtered install lacked sibling workspace deps) and is covered by the `lint-knip` required check. The end-to-end effect on runner log level and API log suppression is asserted only through the existing per-reason tests plus the traced wiring, not through a new cross-component test; the PR body defines the corresponding post-deploy observation with explicit acceptance criteria, which is the right boundary for that claim.

PR #33132, opened 2026-09-10
fix(api): retain safe causes for failed activity snapshot writes

### LINKED ISSUE #33091: bug(api): retain safe causes for failed activity snapshot writes
## Summary

retain safe causes for failed activity snapshot writes. The current-window observations and their causal limits are recorded below.

Related context: #32819, #32848. Closed issues remain closed; this review does not override prior implementation/release authority.

## Impact

One activity snapshot containing one event failed to write. The endpoint intentionally returned success to its caller; the activity summary may lack this evidence, but primary-run data loss is not shown.

Severity: **medium**. Type: **optional activity persistence failure investigation**. Severity is triage, not a raw-event-to-user equivalence.

### Coverage

- Axiom: ONLY `vm0-web-logs-prod`, `level == "error" or level == "warn"`; **1 records**, first `2026-09-09T15:07:25.085Z`; last `2026-09-09T15:07:25.085Z`. Levels: warn: 1.
- Distinct logged run IDs: **1**. 0 Axiom records lack a run ID; these do not imply additional runs.
- Per-run record distribution: run-234 each has {"Activity snapshot capture":1}.

### Variant 1 — Activity snapshot capture (1 records)

- Source: Axiom; first `2026-09-09T15:07:25.085Z`; last `2026-09-09T15:07:25.085Z`.
- Native causal fields: `fields.eventCount`, `fields.outcome`.
- Inner error: **<not logged>**. Only the original outcome/status fields are available.
- fields.outcome: write_failed: 1.

Original sanitized sample 1:

```json
{
  "_time": "2026-09-09T15:07:25.085Z",
  "fields.context": "api:run-activity",
  "fields.eventCount": 1,
  "fields.outcome": "write_failed",
  "fields.runId": "<redacted:run-234>",
  "level": "warn",
  "message": "Activity snapshot capture",
  "source": "api",
}
```

## Current code analysis

- [turbo/apps/api/src/signals/services/run-activity-snapshot.service.ts](https://github.com/vm0-ai/vm0/blob/baa6e4c1f12b712e6e50dcf13e24e84cc77c8ecd/turbo/apps/api/src/signals/services/run-activity-snapshot.service.ts)

The native outcome is write_failed with eventCount=1 and no inner error. Current code settles activityTransaction and deliberately omits raw database exceptions because they may contain bound private evidence, then returns 200. This is not enough to distinguish timeout, lock, schema or other persistence faults. It is a focused failure boundary within feature epic #32819, not grounds to treat all feature work as broken.

## Desired outcome

Add a safe finite failure-stage/code classification without raw driver text or evidence values, verify retry/expiry/claim behavior, and resolve the proven persistence defect. If skipping a snapshot is deliberately acceptable, classify the skip below warn/error while exposing sustained unavailable activity evidence.

## Ownership

Selected owner: @lancy. @lancy authored activity persistence #32871 and the diagnostic change #32996. Confidence: high.

lancy:
Closes #33091

## Problem

A failed activity snapshot capture recorded only `outcome: write_failed`, so
every failure class collapsed into one opaque signal. Lock contention, a run
deleted mid-flight, a cancelled statement and a genuine persistence fault were
indistinguishable in production, and all of them were reported at `warn`.

Production evidence (`vm0-web-logs-prod`, `fields.context: "api:run-activity"`):

- `2026-09-09T15:07:25.085Z` and `2026-09-10T01:39:48.964Z`, both
  `eventCount: 1`, `pdx1`, deployments `dpl_33ZrJJoo7ptYumLYJfQyRzfh2ryy` and
  `dpl_GAdSRp1VGGB3kFQeM4BgZnu26iM1`.
- The first `api:run-activity` record is `2026-09-09T14:15:24.689Z`, so the
  observed window is roughly 12 hours: **2 failures in 56,188 records**. The
  issue's "1 record in 24 hours" covered only ~1h45m of the feature's logging.
- In `[15:00Z, 16:00Z)` alone, 9,794 captures fell into 6,570 `(runId, 1s)`
  buckets and **2,426 of them (37%) held two or more captures for the same run
  in the same second**. With `lock_timeout = 250ms`, overlapping row locks for
  one run are routine.
- Each failure was followed by a successful capture for the same run 0.243 s and
  0.313 s later.

The failures are not cancellations: `signal.throwIfAborted()` runs before the
log, and the only in-transaction abort source is that same signal, so a logged
failure is always a real rejection from `activityTransaction`.

## What this changes

**Classification.** `captureFailure` maps the rejection to a finite set.
`contended` (`55P03`) and `run_missing` (`23503`) are expected consequences of
concurrent delivery for one run, so they record at `info` — matching the
existing precedents in `agent-webhook-events.service.ts`, which already logs
`Required database run output projection backpressured` at `info` for `55P03`
and ignores a deleted run for `23503`. `interrupted` (`57014`),
`snapshot_missing` and the residual `write_failed` keep `warn` because they need
an owner. Sustained unavailability is read by aggregating `fields.outcome`, not
from a per-batch error level.

**Safe diagnostics.** `safeSqlStateCode` publishes the SQLSTATE class code
alone, validated against `/^[0-9A-Z]{5}$/` so a non-conforming driver value is
dropped rather than emitted. `fields.stage` (`admission` / `lock` / `persist` /
`commit`) reports how far the transaction got. Driver messages, statement text,
constraint details and bound parameters are still never attached.

**A real defect in the projection.** `name` and `callId` were truncated with
`String.prototype.slice`, which can split a surrogate pair and leave an unpaired
surrogate, and excerpts were never filtered for `U+0000`. PostgreSQL rejects
both while parsing a jsonb parameter, which fails the entire snapshot write
rather than the offending excerpt — and both are reachable from real tool
output. Every projected string is now bounded by code points with those two
code points dropped.

This is deliberately the diagnostic step. A bounded retry for `contended` is a
separate decision, to be made once a day of classified production data shows
whether contention actually dominates; `mergeActivity` idempotency (now covered
by a test) is the precondition for it.

## Behavior that is unchanged

The endpoint still returns `200`; an optional consumer cannot reject accepted
execution events or suppress normal message publication. `Activity summary
unavailable` / `storage_failed` and the `unavailable` response keep their
meaning. Log messages are unchanged, so the `api/no-logger-info` allowlist needs
no edit. No schema, migration or feature-switch change.

## Verification

Run locally against a local PostgreSQL 18:

- `vitest --run apps/api/src/lib/__tests__/run-activity.test.ts` — 7 passed
  (code-point truncation, `U+0000` and unpaired-surrogate removal, no
  jsonb-rejected escape survives, `mergeActivity` idempotency).
- `vitest --run apps/api/src/lib/__tests__/pg-errors.test.ts` — 18 passed
  (exact SQLSTATE classification; class code published without driver text;
  numeric, lowercase, over-long, absent and transport codes all dropped).
- `oxlint --deny-warnings` and `eslint --max-warnings 0` on the changed files —
  clean.
- `pnpm run check-types`: `deps`, `boundaries`, `gateways`, `core` and `tests`
  passed; `bootstrap-wiring` had not finished locally when this PR was opened.

Not run locally: `apps/api/src/signals/routes/__tests__/chat-activity-summary.test.ts`.
That suite fails identically on unmodified `main` in this environment
(`claimRunnerJob` → 404 "Job not found in queue"), so it needs setup this
sandbox lacks. Its updated expectation — `info` / `contended` / `stage: "lock"`
/ `errorCode: "55P03"` through the real endpoint, plus assertions that no
driver text, statement text or message content reaches the record, and that the
adjacent `Activity summary unavailable` warn is untouched — is verified by CI,
not by me.

## Production acceptance (after release, controller-owned)

Baseline: 2 `write_failed` in 56,188 records over
`2026-09-09T14:15:24Z–2026-09-10T02:14:39Z`. After deployment, over ≥12 hours
and ≥40,000 capture records: every failure carries a classification, no
`write_failed` remains without an `errorCode`, `warn` volume drops but does not
reach zero, and an APL check confirms `fields` holds exactly the documented key
set.

### DIFF
--- a/docs/chat-thread-activity-summary.md
+++ b/docs/chat-thread-activity-summary.md
@@ -57,6 +57,10 @@ credential-shaped argument keys are additionally redacted.
   excerpts. Keep sequence and block identity, discard oldest evidence first,
   and canonicalize object keys before hashing so JSONB ordering cannot create
   false revisions. Relevant late events can merge into the retained window.
+- Bound every projected string by code points and drop the two code points
+  PostgreSQL refuses inside a `jsonb` value: `U+0000` and unpaired surrogates.
+  Both are reachable from real tool output, and either one would otherwise
+  reject the whole snapshot write instead of the offending excerpt.
 - Merge and claim under a short row-locking transaction with a 250 ms lock
   timeout and 3 second statement timeout. Provider I/O runs outside the
   transaction. Optional capture failures cannot reject accepted execution
@@ -86,18 +90,30 @@ failed operations so they survive the default Axiom transport's `info` threshold
 The shared logger and unrelated debug filtering are unchanged. Axiom events
 retain `source: api`, the stable message, and the following nested `fields`:
 
-| Message                        | Context                | Level                                             | Safe fields besides context                                                                                |
-| ------------------------------ | ---------------------- | ------------------------------------------------- | ---------------------------------------------------------------------------------------------------------- |
-| `Activity summary cache`       | `api:activity-summary` | info                                              | `runId`, `outcome` (existing response status)                                                              |
-| `Activity summary attempt`     | `api:activity-summary` | info                                              | `runId`                                                                                                    |
-| `Activity summary completion`  | `api:activity-summary` | info on success; warn otherwise                   | `runId`, `outcome`, `durationMs`, `cooldownMs`; numeric `providerStatus` only for `OpenRouterRequestError` |
-| `Activity summary unavailable` | `api:activity-summary` | warn                                              | `runId`, `outcome: storage_failed`                                                                         |
-| `Activity snapshot capture`    | `api:run-activity`     | info for written/unchanged; warn for write_failed | `runId`, `outcome`, `eventCount`                                                                           |
-| `Activity snapshot cleanup`    | `api:run-activity`     | info on success; warn on failure                  | `outcome`, `removed`, `retentionMs`                                                                        |
+| Message                        | Context                | Level                                                                  | Safe fields besides context                                                                                |
+| ------------------------------ | ---------------------- | ---------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------- |
+| `Activity summary cache`       | `api:activity-summary` | info                                                                   | `runId`, `outcome` (existing response status)                                                              |
+| `Activity summary attempt`     | `api:activity-summary` | info                                                                   | `runId`                                                                                                    |
+| `Activity summary completion`  | `api:activity-summary` | info on success; warn otherwise                                        | `runId`, `outcome`, `durationMs`, `cooldownMs`; numeric `providerStatus` only for `OpenRouterRequestError` |
+| `Activity summary unavailable` | `api:activity-summary` | warn                                                                   | `runId`, `outcome: storage_failed`                                                                         |
+| `Activity snapshot capture`    | `api:run-activity`     | info for written/unchanged and for an expected failure; warn otherwise | `runId`, `outcome`, `eventCount`; `stage` and `errorCode` on failure                                       |
+| `Activity snapshot cleanup`    | `api:run-activity`     | info on success; warn on failure                                       | `outcome`, `removed`, `retentionMs`; `errorCode` on failure                                                |
 
 Completion outcomes remain `success`, `timeout`, `provider_failure`, and
 `invalid_or_unconfigured`. A provider HTTP 429 is distinguishable by
 `fields.providerStatus: 429`; no error object or provider body is attached.
+
+A failed capture is classified into a finite set instead of one opaque
+`write_failed`. `contended` (`55P03`) and `run_missing` (`23503`) are expected
+consequences of concurrent delivery for one run, so they record at `info`;
+sustained unavailability is read by aggregating `fields.outcome`, not from a
+per-batch error level. `interrupted` (`57014`), `snapshot_missing` (the row
+vanished between the upsert and the locking read) and the residual
+`write_failed` keep `warn` because they need an owner. `fields.stage` is one of
+`admission`, `lock`, `persist` or `commit`, and `fields.errorCode` is the
+SQLSTATE class code alone — five characters, validated before it is published,
+and omitted when the driver reports no SQLSTATE. Driver messages, statement
+text, constraint details and bound parameters are never attached.
 
 Granularity is unchanged: one cache record for a request resolved without a
 claim, one attempt/completion pair per generation attempt, one unavailable
@@ -106,7 +122,8 @@ batch (including unchanged duplicates), and one cleanup record per maintenance
 operation (including zero removals). Disabled, irrelevant, and ineligible
 captures remain silent. `eventCount` counts the submitted batch, not new retained
 entries. Failed cleanup reports `removed: 0` with `outcome: failed`; that is not a
-successful empty cleanup.
+successful empty cleanup. A skipped capture still drops that batch's evidence:
+the runner already holds its `200`, and no redelivery or retry is attempted.
 
 These records never contain prompts, phrases, messages, arguments, evidence,
 credentials, database-driver errors, or provider response bodies. The tests call
--- a/docs/chat-thread-activity-summary.md
+++ b/docs/chat-thread-activity-summary.md
@@ -110,7 +110,9 @@ sustained unavailability is read by aggregating `fields.outcome`, not from a
 per-batch error level. `interrupted` (`57014`), `snapshot_missing` (the row
 vanished between the upsert and the locking read) and the residual
 `write_failed` keep `warn` because they need an owner. `fields.stage` is one of
-`admission`, `lock`, `persist` or `commit`, and `fields.errorCode` is the
+`begin`, `admission`, `lock`, `persist` or `commit`, where `begin` covers
+connection acquisition and the transaction's own timeout statements. Its
+companion `fields.errorCode` is the
 SQLSTATE class code alone — five characters, validated before it is published,
 and omitted when the driver reports no SQLSTATE. Driver messages, statement
 text, constraint details and bound parameters are never attached.
--- a/turbo/apps/api/src/lib/__tests__/pg-errors.test.ts
+++ b/turbo/apps/api/src/lib/__tests__/pg-errors.test.ts
@@ -1,0 +1,69 @@
+import { describe, expect, it } from "vitest";
+
+import {
+  isForeignKeyViolation,
+  isLockNotAvailable,
+  isQueryCanceled,
+  isUniqueViolation,
+  safeSqlStateCode,
+} from "../pg-errors";
+
+// Public APIs cannot construct arbitrary database-driver failures. Route tests
+// own the end-to-end proof that the installed driver really surfaces a SQLSTATE
+// this way; this file pins the exact classification and redaction boundary.
+function driverError(code: unknown): Error {
+  return new Error("insert into run_activity_snapshots ... failed", {
+    cause: { code },
+  });
+}
+
+describe("SQLSTATE classification", () => {
+  it.each([
+    ["55P03", isLockNotAvailable],
+    ["23503", isForeignKeyViolation],
+    ["23505", isUniqueViolation],
+    ["57014", isQueryCanceled],
+  ])("matches %s only for its own predicate", (code, predicate) => {
+    expect(predicate(driverError(code))).toBeTruthy();
+    expect(
+      [
+        isLockNotAvailable,
+        isForeignKeyViolation,
+        isUniqueViolation,
+        isQueryCanceled,
+      ].filter((other) => {
+        return other(driverError(code));
+      }),
+    ).toStrictEqual([predicate]);
+  });
+});
+
+describe("safeSqlStateCode", () => {
+  it.each(["55P03", "23503", "23505", "57014", "40P01", "22P05"])(
+    "publishes the %s class code",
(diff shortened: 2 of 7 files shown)


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: 3fddb33149c6652ee4ac74c7b9705646220e7788
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `3fddb33149c6652ee4ac74c7b9705646220e7788`. Practice documents read from `main` at `e5e52075f4f2285ebaf93bd62b61465d91f05b9e`.

### Summary

Replaces the single opaque `write_failed` outcome on a failed activity snapshot capture with a finite classification (`contended`, `run_missing`, `interrupted`, `snapshot_missing`, `write_failed`) plus a content-free `stage` and a validated SQLSTATE class code, and downgrades only the two genuinely expected classes to `info`. Also repairs a real projection defect: `name`/`callId` were truncated with `String.prototype.slice`, which can leave an unpaired surrogate, and excerpts were never filtered for `U+0000` — PostgreSQL rejects either while parsing the jsonb parameter, failing the whole snapshot write rather than the offending excerpt.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | Capture runs as an optional consumer from `webhooks-agent-events.ts` `waitUntil` after the `200`; the classification only changes the log record, and `captureRunActivity$` still returns `{ status: 200 }` on every path. |
| `REVIEW.md` | Races, cancellation, ordering, failure outcomes | Pass | `signal.throwIfAborted()` still precedes the log, so a cancellation cannot be misreported as a failure class. `stage` is assigned only inside a callback `activityTransaction` invokes exactly once, then read after it settles. |
| `REVIEW.md` | Meaningful coverage for new behavior and regression fixes | Pass | `contended`/`stage`/`errorCode` asserted through the real webhook + summary endpoints in `chat-activity-summary.test.ts`; jsonb sanitization covered end to end by the new "retains activity for tool output PostgreSQL cannot store verbatim" case. |
| `REVIEW.md` | Report concrete practice violations with rule and source | Pass | Both findings raised during this review were concrete and are fixed in this HEAD (see Findings). |
| `REVIEW.md` | Persisted data / response validation / `jsonb().$type<>()` reviewed together | Pass | `RunActivityEntries` shape is unchanged; only string content is sanitized. No old row can hold the dropped code points, because PostgreSQL rejected those values at write time, so no migration or backfill applies. |
| `docs/bad-smell.md` §1 | No `any`; narrow `unknown` at untrusted boundaries | Pass | `safeSqlStateCode(error: unknown)` narrows via `typeof code === "string"` and a regex before returning; no `any` in the diff. |
| `docs/bad-smell.md` §2 | No lint/type suppressions or weakened rules | Pass | No suppression comment added; `api/no-logger-info` allowlist untouched because the log messages are unchanged. |
| `docs/bad-smell.md` §3 | Handle errors only where meaningfully handled; no fabricated defaults hiding invariants | Pass | No new catch. `snapshot_missing` is surfaced as its own `warn` class instead of being folded into the generic bucket, and `write_failed` remains the honest unclassified residue. |
| `docs/bad-smell.md` §4 | Document changed public contracts and assess consumers | Pass | `docs/chat-thread-activity-summary.md` records the new outcome set, `stage`, `errorCode`, and the level policy. |
| `docs/bad-smell.md` §6 | Centralized configuration, no invented fallback values | N/A | No configuration, URL, or credential surface in the diff. |
| `docs/fallback.md` §3 | No `??`/`||` chain for a state the contract cannot produce | Pass | The initial `codePointAt(0) ?? 0` was exactly this and is removed in `3fddb33`; `Array.from` guarantees a non-empty code point per element. |
| `docs/fallback.md` §5 | No readers kept for retired producers | N/A | No producer retired; no reader branch added for an old shape. |
| `docs/fallback.md` §9 | Every new fallback declared in the PR body | Pass | After `3fddb33` the diff introduces no fallback, so no `Fallbacks` section is required. |
| `docs/fallback.md` §1 | No tombstone tests pinning deleted behavior | Pass | The existing failure case was updated in place to the new `contended` expectation rather than kept alongside a `write_failed` assertion. |
| `docs/deployment-compatibility.md` | Compatibility across independently deployed boundaries | N/A | Log fields are not a cross-version boundary, and the persisted `entries` shape is unchanged. Old and new API instances write the same jsonb contract during rollout. |
| `docs/deployment-compatibility.md` | Additive change patterns at cross-version boundaries | Pass | `stage` and `errorCode` are additive optional fields; `errorCode` is omitted rather than emitted as a placeholder when the driver reports no SQLSTATE. |
| `docs/testing.md` | Prefer integration coverage through real entry points | Pass | The behavioral regression is proved through `POST` agent events + the activity-summary endpoint; the unit files cover only the pure projection and the classification helper. |
| `docs/testing.md` | Real database and real internal code; MSW only at external boundaries | Pass | The route test uses the real DB and MSW for OpenRouter/Axiom ingestion; no `vi.mock` of internal modules is added. |
| `docs/testing/testing-external-behavior.md` | Verify externally observable state, not internal state | Pass | Assertions are on the summary endpoint response and on records the production Axiom transport actually ingests, not on service return values or DB rows. |
| `docs/testing/testing-external-behavior.md` | Exceptions only for states impossible to construct externally, with a stated reason | Pass | `pg-errors.test.ts` states why arbitrary driver failures are not constructible through a public API and defers the driver-shape proof to the route test; it matches the existing `immutable-dedupe-insert.test.ts` precedent. |
| `docs/testing/testing-external-behavior.md` | API test files must not import DB schema or service files | Pass | Neither new test file imports `@okouai/db/schema/*`, a service, or a DB handle; `eslint --max-warnings 0` is clean on all changed files. |
| `docs/testing/api-testing.md` | Route behavior covered through `setupApp()` and the ts-rest contract | Pass | The new case reuses the file's existing `fixture`/`deliver`/`summarize` helpers, which go through `setupApp()` and the webhook/summary contracts. |

### Findings

Both items below were found during this review and are already fixed in the inspected HEAD; they are recorded for the audit trail, not as outstanding blockers.

- **P1 (fixed in `b30df9e`) — `stage` misreported the pre-transaction window.** `turbo/apps/api/src/signals/services/run-activity-snapshot.service.ts`. `stage` was initialized to `"admission"`, but `activityTransaction` acquires a pool connection and runs its own `SET LOCAL lock_timeout` / `statement_timeout` statements before the command's callback executes. A connection-acquisition timeout (`DB_POOL_CONNECT_TIMEOUT_MS`) — a plausible failure for work running in `waitUntil` — would have been reported as if this command's admission queries had run, which defeats the purpose of the field this PR adds. Fixed by adding a `begin` stage and entering `admission` inside the transaction body.
- **P1 (fixed in `b30df9e`) — regression coverage sat below the external boundary.** The jsonb sanitization is a fix for a production defect, and the state is fully constructible through the production webhook: a `tool_use` argument carrying a NUL reaches the snapshot but not the required output projection, which extracts assistant text blocks only (`agent-event-consumer-run-output.service.ts:150-176`). Per `docs/testing/testing-external-behavior.md`, unit coverage of `mergeActivity` alone was not sufficient. Fixed by adding an endpoint-level case that delivers the hostile event and asserts `outcome: "written"` plus a `fresh` summary.

Non-blocking observations:

- `captureFailure` identifies the invariant failure by comparing `error.message` to the shared `SNAPSHOT_MISSING` constant. It cannot drift, because the constant is both the only thrower and the only reader, but a dedicated error class (as with `AgentEventRunNotFoundError`) would be more robust if that message ever gains another producer.
- `boundedText` filters the whole input before slicing to the limit. That is the same `O(input)` behavior as the previous `Array.from(value).slice(...)`, so it is not a regression, but neither form short-circuits on a very large tool result.
- **Unverified.** I could not check whether an Axiom monitor or saved query keys on `outcome == "write_failed"` and would silently stop matching now that contention records as `contended`. The available token returns `403 monitors:read`. Worth confirming before the release lands.

### Testing

Run and passing locally against a local PostgreSQL 18:

- `vitest --run apps/api/src/lib/__tests__/run-activity.test.ts` — 7 passed.
- `vitest --run apps/api/src/lib/__tests__/pg-errors.test.ts` — 18 passed.
- `oxlint --deny-warnings` for the `api` workspace and `eslint --max-warnings 0` on every changed file — clean.
- `pnpm run check-types`: `deps`, `boundaries`, `gateways`, `core` and `tests` passed on an earlier revision of this branch; the run against the final HEAD had not finished when this comment was posted.

Not run locally: `apps/api/src/signals/routes/__tests__/chat-activity-summary.test.ts`. That suite fails identically on unmodified `main` in this environment (`claimRunnerJob` → `404 Job not found in queue`), so it needs setup this sandbox lacks. Both endpoint-level cases in this PR — the `contended` / `stage: "lock"` / `errorCode: "55P03"` classification with its privacy assertions, and the jsonb-sanitization regression — are therefore verified by CI, not locally.

Coverage gap worth naming: `run_missing`, `interrupted` and `snapshot_missing` have no behavioral test, because none of those states is constructible through the production webhook — each needs a delete, a statement cancellation, or a row removal to race a transaction already in flight. `pg-errors.test.ts` pins their SQLSTATE mapping, and the route test proves the driver really surfaces a SQLSTATE this way for `55P03`; the remaining risk is confined to which branch a given code selects, which is a pure `if` chain over that same helper.

The PR body's production-acceptance criteria (≥12 hours, ≥40,000 capture records, no `write_failed` without an `errorCode`, and an APL check that `fields` holds exactly the documented key set) are stated against a real measured baseline of 2 failures in 56,188 records. Review completion is not that acceptance.

PR #32861, opened 2026-09-09
refactor(platform): remove page visibility and reconnect recovery

e7h4n:
## Summary

Remove application-wide recovery and bulk resynchronization triggered by page visibility or transport reconnects. Remove the BFCache/pagehide teardown branches, foreground-ready barrier, realtime pause/resume/rebuild logic, reconnect catch-up registry, and Worker-to-page reconnect notification. Connection state reporting remains diagnostic; normal server events and explicit data reads drive updates.

- Initialize one private root controller in the HTML bootstrap. Switching Clerk sessions cancels the old app before calling Clerk independently, then performs full-page navigation. Preserve Clerk URL decoration and pending-session-task routing.
- Register each MessagePort connection after acquiring a unique Web Lock. The Worker waits for that lock to be released and uses the existing cleanup to cancel token requests, unregister the tab, and release its realtime subscriptions. Browser document destruction releases the lock without a page event handler.
- Keep one-time subscription initialization explicit, including queued events arriving during initialization. Remove reconnect-driven callbacks, their flags and command registries, the `reconnect` protocol variant, direct-test recovery forwarding, and obsolete diagnostic recovery fields and labels.
- Refresh missing chat-thread metadata against current server state after startup, so notification navigation does not treat an old completed sync as proof that a thread is absent.
- Remove document-visibility handling from private attachment refresh, viewer lease renewal, shortcut hints, connection diagnostics, and bootstrap telemetry. Delete lifecycle diagnostic events, browser-state fields, UI badges, translated labels, and the retired visibility test helper. Connection indicators use the Worker connection status.
- Remove window focus and blur handling from the main chat container focus fallback, sidebar resize drag, and shortcut hints. Remove the pagehide abort on the desktop-auth page, which now cancels with its route signal only. Drop the Playwright BFCache simulation for the unsupported-browser page, since the lifecycle branch it exercised no longer exists.

The page/Worker registration protocol gains a connection lock name and drops the reconnect notification. The Worker asset is content-hashed and bundled with the page; API and IndexedDB formats are unchanged.

## Comparison with introduction diffs

| Historical change | Removed behavior |
| --- | --- |
| [#9938](https://github.com/vm0-ai/vm0/pull/9938), [#25794](https://github.com/vm0-ai/vm0/pull/25794), [#25899](https://github.com/vm0-ai/vm0/pull/25899) | Visibility/focus catch-up and shared Clerk/foreground coordination |
[...]

Issue #32471, opened 2026-09-08
bug(platform): prevent Ably realtime from getting stuck after recoverable failures


seven332:
## Summary

Ensure realtime reconnects when network and authentication conditions permit, without leaving an active session permanently unable to receive updates. Suppressing expected disconnect reports is already implemented and is not the completion criterion for this issue.

## Current source findings

Reviewed main at `26c387f49b2858be89eb130a859a2e3999f9cbf1` on 2026-09-08 (Asia/Shanghai). These findings are from source inspection, not a production-browser or live-log acceptance run.

- Ordinary disconnected/suspended retries are configured at 5 seconds and 15 seconds. Successful reconnects request catch-up: [Ably client setup](https://github.com/vm0-ai/vm0/blob/26c387f49b2858be89eb130a859a2e3999f9cbf1/turbo/apps/platform/src/signals/realtime.ts#L1104).

- An established page-owned client in failed/closing/closed state can be rebuilt on a foreground opportunity: [foreground recovery](https://github.com/vm0-ai/vm0/blob/26c387f49b2858be89eb130a859a2e3999f9cbf1/turbo/apps/platform/src/signals/realtime.ts#L1236).

- The SharedWorker setup has a concrete recovery gap to verify: initial setup failure records failed and returns, while workerDaemonsStartedState remains true. Later daemon-start calls return without restarting: [worker daemon setup and start guard](https://github.com/vm0-ai/vm0/blob/26c387f49b2858be89eb130a859a2e3999f9cbf1/turbo/apps/platform/src/shared-database/worker-signals.ts#L415).

- Page focus/online listeners are installed by the page foreground coordinator; there is no equivalent setup call in the inspected Worker bootstrap. Do not assume page recovery reaches the Worker-owned Ably client.

- The original two Sentry disconnect records do not prove that this terminal-state path occurred in production.

## Acceptance criteria

- With a valid user/organization session, initial connection, token renewal, network loss, and network restoration must not strand the realtime owner or its subscriptions.
- Reproduce and address the initial Worker setup failure/start-guard path, and verify recovery after an established Worker connection becomes inactive.
- Cover the existing page transport and the feature-gated Worker transport; preserve their ownership and cancellation boundaries.
- After recovery, subscriptions receive updates and catch up missed data. Expected transient disconnects remain excluded from Sentry.
- Authentication revocation and intentional lifecycle shutdown retain their existing terminal semantics; the scope is recoverable failures.

## Historical production evidence

[...]

Issue #32147, opened 2026-09-07
bug(platform): recover realtime channels after Ably connection and attach timeouts


seven332:
## Summary

Platform realtime emitted six connection-unavailable errors and six channel-attach timeouts across desktop and mobile browsers.

## Impact

Live chat updates can be delayed or absent until reconnection. At least five users were affected in the attach-timeout group.

### Coverage

- Window: Beijing 2026-09-04 00:00 through 2026-09-07 00:00 Asia/Shanghai (half-open); UTC `2026-09-03T16:00:00Z` through `2026-09-06T16:00:00Z` (half-open).
- Source: Sentry project `platform` or `desktop`, exact predicates `environment:production` and `level:error`, all issue statuses, full event pagination.
- Public-safe Sentry issue IDs: `PLATFORM-AS`, `PLATFORM-58`.
- Raw Sentry events in this problem: **12**. The groups reported two and five affected users respectively; their union is not exposed.
- Deduplication: Both variants originate in Ably 2.21 connection/channel state and represent the same realtime availability boundary; connection and channel phases remain explicit.

### Variant 1 — connection unavailable (6 events)

- Source/field: Sentry exception value, issue PLATFORM-AS.
- First/last: `2026-09-04T02:45:59Z` / `2026-09-06T04:14:59Z`.
- Component/runtime: Chrome Mobile 152, platform realtime.

```text
timestamp=2026-09-04T02:45:59Z
level=error
environment=production
issue=PLATFORM-AS
exception.type=Error
exception.value=Connection to server unavailable
mechanism=generic (with handled chained Ably cause)
handled=true
culprit=src/signals/realtime.ts
app_frame=src/signals/realtime.ts:801
```

### Variant 2 — channel attach timeout (6 events)

- Source/field: Sentry exception value, issue PLATFORM-58.
- First/last: `2026-09-03T18:32:02Z` / `2026-09-04T14:57:34Z`.
- Component/runtime: Ably channel state timer; Chrome and Mobile Safari.

```text
timestamp=2026-09-03T18:32:02Z
level=error
environment=production
issue=PLATFORM-58
exception.type=Error
exception.value=Channel attach timed out
mechanism=generic
handled=true
culprit=ably timeoutPendingState
app_frame=ably/index.mjs:timeoutPendingState
```

## Current code analysis

Analyzed `origin/main` at `ea29038571b77c7dda8c95c8e500599c33db984c`.

[...]

PR #33134, opened 2026-09-10
fix(platform): recover realtime subscriptions after a connection gap

e7h4n:
## Summary

Two defects in how Platform drives Ably, both fixed by using Ably's own primitives instead of working around them.

**1. A subscription created while the connection was suspended lost its listener permanently.** Ably attaches implicitly on `channel.subscribe()`, and `RealtimeChannel._attach()` rejects outright — rather than queueing — when `ConnectionManager.activeState()` is false (`suspended`, `closing`, `closed`, `failed`). This module treated that rejection as a subscription failure and ran its cleanup, unsubscribing the listener it had just registered. Ably then reattached the channel on reconnect, but it cannot restore a listener the application removed, so the stream stayed silent until a reload.

Ably hit the same hazard in its own React hooks and solved it with `attachOnSubscribe: false`: *"we explicitly attach channels ... to avoid situations where implicit attachment could cause errors (when connection state is failed or disconnected)"*.

- Create the three session channels with `attachOnSubscribe: false`, so subscribing only registers a listener and cannot fail on a transport condition.
- Attach all three once in `connectRealtimeClient$`, where the connection is known to be `connected`. Reaching `attaching` is enough to hand each channel to `Channels.onTransportActive()`, which reattaches `attaching`/`suspended`/`attached` channels after every later drop. A channel that drops becomes `suspended`, never `initialized`, so it does not fall out of that set again.
- A rejected attach is logged and recorded as a connection diagnostic, not propagated.

**2. Catch-up ran exactly once per subscription, and always ran.** Both are wrong. `ChannelStateChange.resumed` is Ably's continuity signal, so drive catch-up from it.

- A reattach Ably replayed reports `resumed: true` and now performs no extra reads, instead of refetching on every brief blip.
- A reattach it could not replay re-runs the subscription's baseline: a topic loop is poked exactly as an unseen event would poke it, and a payload loop re-runs its initialization on the next iteration. This closes the case where a connection dropped past `connectionStateTtl` and the tab kept stale values until a reload — the staleness #32861 documented as accepted.
- The attach that first brings a subscription up also reports no continuity. That is not a gap for a subscription with no baseline yet, so resyncs only count once the subscription is live.

Only the Worker owns the Ably channel, so only it observes `resumed`. It forwards a `realtime-resync` message to every tab holding the subscription, which reacts exactly as it would to an event on the topic. The signal is scoped to one `(scope, topic)` subscription; it is not a reinstated application-wide catch-up.

No retry loop, backoff table or error-code classification is added.

## Removing the realtime connection indicator

The sidebar indicator rendered only under the staff-only `_debug` switch (`OkouDebug`) or the `diagnostics` URL parameter, and it was the sole reader of the connection status pipeline. It is removed along with everything it kept alive: the page signal, the bridge `statusChanged` event, the `status` protocol message, the Worker status plumbing, and the connection-state listener registry. Knip is clean on the result.

The user-facing degraded-realtime toast (`global.realtime.degraded`) is unaffected; only the two debug-only strings are removed.

## Rollout compatibility

The page-to-Worker protocol changes (`status` removed, `realtime-resync` added), so the old/new combination was assessed rather than assumed. A tab constructs its SharedWorker with the name `okou_<userId>_<orgId>` while the worker asset URL is content-hashed. Per the SharedWorker constructor, a matching name with a different URL is a `URLMismatchError`, so a tab from one build never connects to a Worker from another: either it reuses the Worker built from the same asset, or it fails to construct one at all. There is therefore no reachable state in which an old page speaks this protocol to a new Worker, and no tolerance branch is owed under `docs/fallback.md`.

`docs/deployment-compatibility.md` is updated in this PR: it described the Worker as marking the connection disconnected on load and transport failures, which is the pipeline removed here.

## Behavior boundaries

Terminal states are unchanged: a channel that reaches `failed` still rejects the subscription, so authentication revocation and intentional shutdown keep their semantics. All three channels now attach at session setup instead of on first subscribe — the Worker daemons use `credential` and `user` immediately and page subscriptions use `org`, so this is one extra ATTACH frame per session at most. A payload subscription without an `initializeCommand$` has no baseline to re-read and is unchanged by a continuity gap.

## Test double correction

`src/mocks/ably.ts` modelled neither `attachOnSubscribe`, the inactive-connection rejection, nor `resumed`: its `subscribe()` attached unconditionally and only failed on a closed connection. That is why no existing test covered these paths. It now mirrors `RealtimeChannel._attach()` and `ConnectionManager.activeState()`, including the exact `Unable to connect (network unreachable)` from `ConnectionManager.getError()`, and distinguishes a replayed reattach (`resume()`) from one that lost continuity (`reconnect()`).

## Verification

- 7 new cases. Each was confirmed to fail without the corresponding production change: subscribing while suspended, keeping the listener through an offline episode, initialization ordering, terminal channel failure, a continuity gap re-reading the baseline, a replayed reattach *not* re-reading it, and a continuity gap poking a topic loop. The Worker fan-out is covered in `message-port.test.ts` across two tabs, including the negative control that a replayed reattach sends nothing.
- 188 targeted Vitest cases passed single-worker across `signals`, `shared-database`, app entrypoint, sidebar, sidebar account menu, chat run state, workflows page, connector accounts, and queue page.
- Platform `check-types`, Prettier, Oxlint, type-aware Oxlint, ESLint (whole workspace), i18n validation, style policy, and Knip.
- `style-legacy-baseline.json` was ratcheted down with `pnpm lint:style:prune`; no allowance was added.
- Real-browser verification of a multi-minute offline episode is a preview item, not covered locally.

## Context

Investigated for #32471. The same fingerprint was previously filed as #32147, whose stated outcome was "reconnect with bounded backoff, resubscribe channels, catch up missed events". #32227 closed it by filtering the exact `Connection to server unavailable` and `Channel attach timed out` messages from Sentry; the underlying gap was never addressed, and the condition resurfaced under a third message the filter does not match. This change fixes the gap. Adjusting the Sentry policy is deliberately left out so the remaining signal stays visible until the fix is confirmed in production.

Refs #32471
Refs #32147


e7h4n:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 0664a7ed8b669d02e3394a2ae87ad4381f353b57

Inspected HEAD: `0664a7ed8b669d02e3394a2ae87ad4381f353b57`
Practice revision: `bd0f81c3c9baddabf67a28f01c57d6a16b1f8f63`

### Summary

Stops `channel.subscribe()`'s implicit attach from discarding a listener when the Ably connection is inactive, and re-drives per-subscription catch-up from `ChannelStateChange.resumed` instead of running it exactly once and unconditionally. Removes the debug-only realtime connection indicator together with the connection-status pipeline it was the sole reader of, including the `status` MessagePort message.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | `realtime.ts` subscribe/attach path traced through `worker-signals.ts` daemons, `worker-realtime-subscriptions.ts` fan-out, and the `message-port-client.ts` tab side. |
| `REVIEW.md` | Races, cancellation, ordering, failure outcomes | Finding (P1) | Ordering of the `live` gate and `resumed` delivery is correct, but session-signal abort during the new `await attachRealtimeChannels` skips `stopObservingChannels()` (F2). |
| `REVIEW.md` | Meaningful coverage for new behavior and regression fixes | Finding (P1) | 7 signals-level cases added, but the user-visible recovery has no page-boundary test while the PR deletes the two page-level realtime tests (F3). |
| `REVIEW.md` | Changed public contracts documented and consumers assessed | Finding (P1) | `status` removed and `realtime-resync` added in `protocol.ts`; `docs/deployment-compatibility.md` still documents the removed behavior (F1). |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | No `any` and no new suppression comments in the diff; the one `eslint-disable` in `runPayloadLoopIteration$` is pre-existing and unchanged. |
| `docs/bad-smell.md` | Preserve resource cleanup and meaningful recovery | Finding (P1) | Subscription cleanup and terminal semantics preserved, but channel-observer cleanup is skipped on one abort path (F2). |
| `docs/bad-smell.md` | Static imports, centralized configuration | Pass | No dynamic imports or hardcoded origins introduced; `realtimeChannelOptions()` is a local literal factory. |
| `docs/fallback.md` | No fallbacks for states the contract cannot produce | Pass | `attachOnSubscribe: false` plus a single setup attach removes the need for a retry/tolerance branch rather than adding one. |
| `docs/fallback.md` | Removing a fallback requires evidence; remove branch, contract entry and tests together | Pass | The status pipeline is removed with its protocol entry, bridge event, Worker plumbing and tests in one pass; Knip reports no leftovers. |
| `docs/fallback.md` | No negative tests pinning removed behavior | Pass | The deleted `sidebar-account-menu` cases are removed outright, not converted into tombstones. |
| `docs/deployment-compatibility.md` | Cover old/new combinations that can coexist during rollout | Pass | The SharedWorker name is `okou_<userId>_<orgId>` while its URL is content-hashed, so a differently-built tab gets `URLMismatchError` rather than speaking a mixed protocol; old page to new Worker is unreachable. Reasoning is not recorded in the PR body (see F1). |
| `docs/deployment-compatibility.md` | Frontend rollout documentation stays accurate | Finding (P1) | The doc's Worker failure description no longer matches the code (F1). |
| `docs/testing.md` | Prefer integration coverage through real entry points | Finding (P1) | New coverage sits at the signals boundary only (F3). |
| `docs/testing.md` | Mock external services at their boundary; no fake timers | Pass | The Ably double is corrected at the SDK boundary and now mirrors `RealtimeChannel._attach()`, `ConnectionManager.activeState()` and `resumed`; assertions wait on observable results. |
| `docs/testing/testing-external-behavior.md` | Platform tests construct and verify through the page | Finding (P1) | New cases drive the store directly without recording why the page cannot construct the state (F3). |
| `.claude/skills/ccstate/SKILL.md` | Every async operation has an owner and cancellation path; no `detach()` inside signals | Pass | The connection-state `detach()` from an earlier revision is gone; attach is awaited inside `connectRealtimeClient$`. |
| `.claude/skills/ccstate/references/lifecycle.md` | AbortSignal ownership matches the work's lifetime | Finding (P1) | `whenChannelAttached` is owned by the session signal, not the subscribing caller's signal (see Non-blocking); the observer-cleanup gap is F2. |
| `docs/styles.md` | Legacy first-party CSS baseline only shrinks | Pass | `index.css` loses the four indicator rules and `style-legacy-baseline.json` was ratcheted down with no allowance added; `pnpm lint:style` passes. |

### Findings

**P1 — F1: `docs/deployment-compatibility.md` documents behavior this PR removes**

`docs/deployment-compatibility.md` (Deployment Model / Frontend) states:

> Worker load and transport failures reject pending requests with their original error **and mark the connection disconnected**.

This PR deletes `SharedDatabaseBridgeEvents.statusChanged`, the `status` protocol message, `updateRealtimeStatusForConnections$`, and the `statusChanged("disconnected")` call in `MessagePortSharedDatabaseBridge.close()` (`turbo/apps/platform/src/shared-database/message-port-client.ts`, `.../worker-context.ts`, `.../protocol.ts`). After this change the Worker no longer marks anything disconnected, so an authoritative document describes a code path that no longer exists.

- Trigger: any reader of the deployment-compatibility guide after this merges.
- Consequence: the guide misstates the current failure contract, which is exactly the class of drift `docs/bad-smell.md` §4 ("Document changed public contracts and assess their consumers") and `docs/fallback.md` §5 ("Delete the reader in the same PR that retires the producer") exist to prevent.
- Fix: update that sentence in the same PR. While editing, also record in the PR body why removing `status` and adding `realtime-resync` needs no rollout compatibility — the SharedWorker name/URL pairing makes an old-tab/new-Worker connection unreachable, which is the evidence `docs/deployment-compatibility.md` asks for.

**P1 — F2: channel observer cleanup is skipped when setup is cancelled during attach**

`turbo/apps/platform/src/signals/realtime.ts`, `connectRealtimeClient$`:

```ts
const stopObservingChannels = observeRealtimeChannels(channels);
await attachRealtimeChannels(channels);   // newly inserted await
signal.throwIfAborted();
...
const close = (): void => { ...; closeConnection(); stopObservingChannels(); };
signal.addEventListener("abort", close, { once: true });
```

Before this PR there was no suspension point between `observeRealtimeChannels()` and the `close` wiring. Now an abort during the attach runs the already-registered `closeConnection`, then `signal.throwIfAborted()` throws, and `stopObservingChannels()` never runs — the three channel state listeners registered by `observeRealtimeChannels` stay attached.

- Trigger: sign-out, org switch, or page teardown while the initial channel attach is in flight.
- Consequence: leaked listeners on a client that is being closed. Impact is small because the Ably client is closed and discarded on the same path, but it is a cleanup regression introduced by this diff and `docs/bad-smell.md` §3 requires preserving resource cleanup.
- Fix: register the `close` handler (or call `stopObservingChannels` from the abort path) before awaiting the attach.

**P1 — F3: the user-visible recovery has no page-boundary regression test**

`docs/testing/testing-external-behavior.md` is explicit that in `turbo/apps/platform` the external interface is the page, and that tests should not mutate the store directly. All seven new cases live in `turbo/apps/platform/src/signals/__tests__/realtime.test.ts` and drive `context.store.set(setAblyLoop$, ...)` directly. At the same time this PR deletes the two page-level realtime cases in `turbo/apps/platform/src/views/okou-page/__tests__/sidebar-account-menu.test.tsx`, so net page-boundary coverage of realtime connection behavior goes to zero.

The exception clause does not clearly apply: `context.mocks.ably.triggerSharedWorkerConnectionState("suspended" | "connected")` and the new `triggerResume()` already let a `setupPage` test construct exactly this state, and the deleted tests used that pattern.

- Trigger: a future refactor that breaks Worker-daemon recovery end to end.
- Consequence: the headline user-visible claim — the sidebar thread list resumes updating after an outage without a reload — is asserted only through internal signals, so the suite can stay green while the product regresses.
- Fix: add one `setupPage` case that observes a thread-list change through the page across a suspend/reconnect cycle. If a page-level case is genuinely not constructible, state that reason in the signals-level test per the Exceptions section.

### Fallbacks

This diff introduces no new fallback. It removes several, and the removal evidence checks out:

| Removed path | Surface | Evidence | Verdict |
| --- | --- | --- | --- |
| `status` MessagePort message and Worker status plumbing | Page ↔ SharedWorker protocol | Sole consumer was the debug-only indicator; the SharedWorker name/URL pairing makes an old-tab/new-Worker pair unreachable, so no rollout gate applies | Justified |
| `okouDebugRealtimeIndicator$` and the indicator UI | Platform sidebar | Rendered only under the staff-only `OkouDebug` switch or the `diagnostics` URL parameter, so `docs/fallback.md` §2 applies and no compatibility is owed | Justified |
| `subscribeRealtimeConnectionState$` / connection-state listener registry | Platform signals | Last caller removed in this PR; Knip confirms no remaining consumer | Justified |
| Unconditional per-subscription catch-up | Realtime transport | Replaced by Ably's `resumed`, which is a strictly better-informed signal; both branches are covered by new tests | Justified |

### Testing

Verified as reported: the realtime and shared-database suites pass, `check-types`, Prettier, Oxlint, type-aware Oxlint, ESLint, i18n validation, style policy and Knip are clean, and the legacy style baseline shrank without a new allowance.

Coverage quality is good at the boundary it targets. The negative controls are the strongest part: a replayed reattach must not re-read the baseline, a terminal channel failure must still reject, and the two-tab case in `message-port.test.ts` asserts both that a continuity gap fans out and that `triggerResume()` sends nothing. The corrected Ably double is a real improvement — it previously modelled neither the inactive-connection rejection nor `resumed`, which is why the original defect was invisible to the suite.

Gaps:

- F3 above: no page-boundary case for the recovery this PR exists to deliver.
- A payload subscription without an `initializeCommand$` has no baseline to re-read on a continuity gap. That is a deliberate limitation, but nothing documents or tests it, so a future reader may assume every payload subscription recovers. Worth a comment on `RealtimePayloadLoopState.resyncPending` or a short case.
- The multi-minute real-browser offline episode remains unverified, as the PR body states.

### Non-blocking observations

- `whenChannelAttached(channel, signal)` is passed the `setupRealtime$` session signal rather than the subscribing caller's signal. If a caller aborts while its channel is still suspended, `subscribeChannel` stays parked and that waiter's channel listener survives until the next `attached` transition or session teardown. It always settles eventually and self-heals on reconnect, so this is bounded, but the ownership does not match the work's actual lifetime under `.claude/skills/ccstate/references/lifecycle.md`. Racing the caller's signal would tighten it.
- `attachRealtimeChannels` now attaches `org` eagerly even for sessions that never subscribe to it. One extra ATTACH frame per session, and the PR body already calls this out; noted only so it is not mistaken for an oversight.

e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: da80933f99e58ffb5b59752b24f36f511381b542
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `da80933f99e58ffb5b59752b24f36f511381b542`
Practice revision: `bd0f81c3c9baddabf67a28f01c57d6a16b1f8f63`
Supersedes the `Changes Requested` review of `0664a7e`.

### Summary

Same change as the previously reviewed revision — `attachOnSubscribe: false` plus a single setup attach so a listener is never discarded on an inactive connection, `ChannelStateChange.resumed` driving per-subscription catch-up, and removal of the debug-only connection indicator with its whole status pipeline. `da80933` resolves all three P1 findings from the prior review.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | Re-traced the delta: only `connectRealtimeClient$` ordering, one doc paragraph, one comment, and one page test changed. |
| `REVIEW.md` | Races, cancellation, ordering, failure outcomes | Pass | `close` is now registered before the first suspension point, so an abort during `attachRealtimeChannels` runs `closeConnection()` and `stopObservingChannels()`. `signal.removeEventListener("abort", closeConnection)` still precedes adding `close`, so no double teardown. |
| `REVIEW.md` | Meaningful coverage for new behavior and regression fixes | Pass | `queue.test.tsx` now covers the recovery through the page; verified it fails without the continuity handling and passes with it. |
| `REVIEW.md` | Changed public contracts documented and consumers assessed | Pass | `docs/deployment-compatibility.md` now matches the code, and the PR body records why the protocol change needs no rollout compatibility. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | Delta adds none; `pnpm check-types`, Oxlint, type-aware Oxlint and ESLint are clean. |
| `docs/bad-smell.md` | Preserve resource cleanup and meaningful recovery | Pass | The observer-cleanup gap that made this P1 is closed; terminal channel failure still rejects. |
| `docs/bad-smell.md` | Static imports, centralized configuration | Pass | No dynamic imports or hardcoded origins in the delta. |
| `docs/fallback.md` | No fallbacks for states the contract cannot produce | Pass | No new branch added; the delta is ordering, documentation and coverage. |
| `docs/fallback.md` | Removing a fallback requires evidence; remove branch, contract entry and tests together | Pass | Unchanged from the prior revision; Knip still reports no leftovers. |
| `docs/fallback.md` | No negative tests pinning removed behavior | Pass | The new queue case asserts live recovery, not the absence of old code. |
| `docs/deployment-compatibility.md` | Cover old/new combinations that can coexist during rollout | Pass | SharedWorker name `okou_<userId>_<orgId>` with a content-hashed URL makes an old-tab/new-Worker pair `URLMismatchError` rather than a mixed protocol; the reasoning is now recorded in the PR body. |
| `docs/deployment-compatibility.md` | Frontend rollout documentation stays accurate | Pass | The Worker failure paragraph now describes the actual contract. |
| `docs/testing.md` | Prefer integration coverage through real entry points | Pass | The headline recovery is asserted through the queue drawer; transport-only states keep their signals-level cases. |
| `docs/testing.md` | Mock external services at their boundary; no fake timers | Pass | The Ably double is corrected at the SDK boundary; the new case waits on rendered text, not elapsed time. |
| `docs/testing/testing-external-behavior.md` | Platform tests construct and verify through the page | Pass | `surviveConnectionGap()` interrupts the transport and the assertions read the drawer; its docstring states why only the connection can produce the state. |
| `.claude/skills/ccstate/SKILL.md` | Every async operation has an owner and cancellation path; no `detach()` inside signals | Pass | Attach is awaited inside the command; no `detach()` in `signals/`. |
| `.claude/skills/ccstate/references/lifecycle.md` | AbortSignal ownership matches the work's lifetime | Pass | Teardown now spans the whole async body (see Non-blocking for the one remaining imprecision, unchanged and bounded). |
| `docs/styles.md` | Legacy first-party CSS baseline only shrinks | Pass | Unchanged from the prior revision; baseline shrank with no allowance added. |

### Findings

No P0 or P1 findings.

Resolution of the prior review:

- **F1 — stale documentation.** Resolved. `docs/deployment-compatibility.md` no longer claims the Worker marks the connection disconnected and now states that no separate connection status is reported. The PR body records the SharedWorker name/URL evidence for the protocol change.
- **F2 — channel observer cleanup skipped on cancel during attach.** Resolved. `close` is installed before `await attachRealtimeChannels(channels)`, so cancelling setup mid-attach still detaches the observers. Checked that moving the `publishConnectionDiagnostic` call after the attach preserves its intent — it reports the post-attach `channels.user.state`, which is the more useful value.
- **F3 — no page-boundary regression coverage.** Resolved. `Queue plan details recover after a realtime connection gap` changes the plan while the connection is unusable, publishes no event, interrupts the transport, and asserts the drawer through the page. Confirmed it fails with the continuity handling disabled.

### Testing

Verified on this HEAD: 104 cases across 13 files pass single-worker (`signals`, `shared-database`, queue page, chat metadata), with `check-types`, Oxlint, type-aware Oxlint and ESLint clean and `git diff --check` empty.

The new page case is a genuine regression test rather than a restatement: with the `resumed` fan-out disabled it fails on the missing `Custom` plan text, and it reaches the assertion only through the drawer the user sees. The `initializeCommand$` limitation is no longer implicit — `RealtimePayloadLoopState.resyncPending` documents that a purely event-driven payload subscription has no baseline to re-read.

Remaining known gap, unchanged and stated in the PR body: the multi-minute real-browser offline episode is a preview verification item.

### Non-blocking observations

- `whenChannelAttached(channel, signal)` still takes the session signal rather than the subscribing caller's. A caller that aborts while its channel is suspended leaves a parked waiter until the next `attached` transition or session teardown. It always settles and self-heals, so it stays out of the blocking set, but tightening the ownership would match `.claude/skills/ccstate/references/lifecycle.md` more exactly.
- `attachRealtimeChannels` attaches `org` eagerly even for sessions that never subscribe to it — one extra ATTACH frame per session, already disclosed in the PR body.

e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 2824c949d14df721242e7568191489b430322188
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `2824c949d14df721242e7568191489b430322188`
Practice revision: `bd0f81c3c9baddabf67a28f01c57d6a16b1f8f63`
Delta since the reviewed `da80933`: a merge of `origin/main` (`c3b8551`) only. No product change.

### Summary

`2824c94` merges current `main` into the branch after the merge queue ejected `da80933` as `CONFLICTING`. The realtime change itself is byte-identical to the LGTM'd revision; this review covers the conflict resolution.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Distinguish introduced defects from unchanged code | Pass | Only `index.css` and `style-legacy-baseline.json` were conflicted; every other file the merge touched (`docs/deployment-compatibility.md`, ten locale files) auto-merged, and each was re-verified below. |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | No behavior delta. `git diff da80933..2824c94` touches no file under `src/signals` or `src/shared-database`. |
| `REVIEW.md` | Meaningful coverage preserved through the merge | Pass | 127 cases across 13 files pass on the merged tree, up from 104 because main added cases to the same suites. |
| `docs/bad-smell.md` | No `any`, no suppressions; preserve cleanup | Pass | Conflict resolution is CSS and a JSON baseline; `check-types`, Oxlint and ESLint are clean on the merged tree. |
| `docs/fallback.md` | Removal keeps branch, contract entry and tests together | Pass | Taking main's `index.css` and re-applying only the realtime deletion keeps main's `.okou-pill` and `.okou-icon-mono` removals and its new badge selectors intact. |
| `docs/deployment-compatibility.md` | Documentation matches the code | Pass | The rewritten Worker-failure sentence survived the auto-merge: `mark the connection disconnected` is absent, `The Worker reports no separate connection status` is present. |
| `docs/styles.md` | Legacy first-party CSS baseline only shrinks | Pass | `pnpm lint:style` passes on the merged tree. The baseline keeps main's entries and drops exactly the four realtime declarations plus the two `sidebar-account.tsx` class-token usages this branch retires. |
| `docs/testing.md` / `testing-external-behavior.md` | Coverage and boundary unchanged | Pass | Test files were not conflicted; the queue-page page-boundary case still passes. |
| `.claude/skills/ccstate/SKILL.md` | Async ownership unchanged | Pass | `realtime.ts` is untouched by the merge. |

### Findings

No P0 or P1 findings.

Conflict resolution verified in detail:

- **`turbo/apps/platform/src/views/css/index.css`.** `main` refactored this file in five commits (`#33246`, `#33233`, `#33196`, `#33133`, `#33062`), including deleting `.okou-pill` and `.okou-icon-mono` and adding the shared badge selectors. Resolution takes main's file wholesale and re-applies this branch's single edit — deleting the four realtime rules and their reduced-motion block. Confirmed `okou-realtime` no longer appears and brace count is balanced (235/235).
- **`turbo/style-legacy-baseline.json`.** Resolution takes main's baseline and removes only this branch's retirements: two entries from `legacyClassTokens`, two `classUsages` entries under `sidebar-account.tsx`, and eight `cssAtoms` declarations for `index.css` (485 → 477). No allowance was added and the baseline still shrinks, which is what `docs/styles.md` requires.
- **Auto-merged files re-verified.** `docs/deployment-compatibility.md` retains the corrected sentence. All ten locale files retain `global.realtime.degraded` and none reintroduced `disconnected` or `reconnecting`.

### Testing

On the merged tree: 127 cases across 13 files pass single-worker (`signals`, `shared-database`, queue page, sidebar account menu), with `check-types`, Oxlint, ESLint, i18n validation and style policy clean, and `git diff --check` empty.

No new coverage is warranted for a merge that introduces no behavior. The gaps recorded against `da80933` are unchanged: the multi-minute real-browser offline episode is still a preview item.

### Non-blocking observations

Both carried over unchanged from the `da80933` review and neither is affected by the merge: `whenChannelAttached` is owned by the session signal rather than the subscribing caller's, and `attachRealtimeChannels` attaches `org` eagerly.

PR #33115, opened 2026-09-10
fix(api): classify transient provider failures from completion envelopes

lancy:
## Problem and change

Closes #33075.

OpenRouter reports an upstream gateway failure inside a `200` body as an error envelope with no choices. `parseOpenRouterGeneration` wraps that in a **local synthetic** `502` with `origin: "completion"`, so `recordOpenRouterRequestFailure` saw `httpStatus === undefined` and its `upstream_timeout` / `provider_unavailable` branches — which matched only `httpStatus` — could never fire. Every transient provider code delivered this way (`504`, `502`, `503`, `408`, `UNAVAILABLE`, `RESOURCE_EXHAUSTED`) classified as `unknown`, which the auxiliary boundary treats as a defect.

The reported event is exactly this shape: `fields.err.status = 502` is the local constant, and `fields.err.errorCode = 504` is the only real upstream signal. Those branches now match the provider `code` the way the rate-limit branch already did, and stay last so an enumerated invalid-request or auth code keeps overriding a transient status.

Initial thinking is the one fast-path caller still outside `generateAuxiliary`, and it reported every provider failure as a warning. Opening copy is optional, is never retried, and its omission changes no run outcome, so a limit, timeout, transport error or upstream gateway failure is an expected outcome there. It now returns without warning for exactly those reasons.

The suppression is deliberately narrow. A rejected request, broken credentials, output the service cannot interpret and unclassified errors all stay reportable. `isTransientProviderFailure` excludes `output_truncated` and `unexpected_tool_calls` for the same reason: they describe our request, not the provider's availability.

## Why not delete the log line

The same call site carried 950–2302 events/day of `fields.err.status = 400` from 2026-09-04 through 2026-09-08. That was a real defect — an unsupported `reasoning.effort` for the exact model — fixed by #32612, and its last hourly bucket is `2026-09-08T09:00Z`, matching that merge. Deleting or blanket-downgrading this warning would have made that class permanently silent.

## Evidence

Read-only Axiom, dataset `vm0-web-logs-prod`, UTC `[2026-09-08T16:00:00Z, 2026-09-09T16:00:00Z)`. `Initial thinking generation failed` was 62 records and the 6th noisiest warn/error message:

| variant | `errorCode` | records |
| --- | --- | --- |
| `OpenRouter request failed: 429` (real HTTP 429) | 429 | 56 |
| `OpenRouter completion failed: 502` (`finish_reason: "error"`) | 429 | 5 |
[...]

PR #33105, opened 2026-09-10
refactor: retire goal backend authority and application orm consumers

lancy:
Closes #33061. Parent: #32653.

## Summary

Remove live Goal backend execution and application ORM consumers while retaining the complete physical schema and historical data. Ordinary run creation, claim, callback and accounting SQL now works with the prospective Goal objects absent; this PR contains no production migration.

- Delete Goal routes/contracts, issuance metadata, state/brief/marker services, queue associations, continuation, settlement and Goal-only lifecycle/timing plumbing. Captured unsupported inputs and Goal-origin claims fail closed before execution; ordinary manual runs with historical provenance still claim normally.
- Use `@okouai/db/runtime/agent-run` throughout application SQL and the runtime registry. Share non-Goal column builders with the unchanged migration-facing definition. Keep runtime declarations outside the unchanged Drizzle schema glob and preserve SQL mapping/lifecycle lint enforcement for the new import path.
- Split live admission from historical run/message/draft decoding. Remove the three Goal capabilities from new tokens while retaining known permissions in real signed mixed legacy tokens, conservative banking classification and reuse-key rejection.
- Resolve late callback/usage grouping from canonical hot/snapshot events. Usage revisions retain the authoritative context pointer, including null, revocation identity and timestamp semantics. First late usage uses retained provenance or emits legitimate ungrouped accounting when none exists. Existing per-run locking, pending exclusion, raw/hourly aggregation and idempotency remain.
- Remove seed mount residue; verify Runner's existing omitted-mount cleanup and Pi's actual snapshot-driven resource discovery. Document retained historical/security references, the bounded captured-context disposition, and the S4-compatible rollback requirement before S5.

## Scope and rollout evidence

S1–S3 are independently production accepted: [S3 certificate](https://github.com/vm0-ai/vm0/issues/32653#issuecomment-5611029565), [complete S2 certificate](https://github.com/vm0-ai/vm0/issues/32653#issuecomment-5604844349). Issue #33061 records the controller's complete 2026-09-10 00:45 UTC 4,162-Goal census, stable ID hash, and independent zero Goal-origin/Goal-linked nonterminal queries. These bounded reads support removal of temporary settlement; they do not certify every stored ordinary execution payload or external deployment. Captured payloads are not rebuilt by new defaults, and no global drain is inferred from token TTL or an old launch package.

[...]

PR #32979, opened 2026-09-09
fix(api): keep auxiliary generations from starving on a shared token budget

lancy:
Closes #32118. Implements the [approved contract](https://github.com/vm0-ai/vm0/issues/32118#issuecomment-5600393190); layers 1, 2 and 3 ship together.

## Root cause

`FAST_PATH_MODEL` (`google/gemini-3.8-flash`) reports `reasoning.mandatory: true` with `supported_efforts: ["high", "medium", "low"]` and no independent reasoning budget, and Gemini 3 spends thinking and visible output from one combined `max_output_tokens`. `effort: "low"` is already this model's floor, so model-chosen thinking could consume the whole budget: `finish_reason` became `length`, the visible content was empty, `parseOpenRouterGeneration` threw, and the auxiliary generation warned with no summary. #31715 raised the ceilings but the budgets it chose are the ones still failing.

## Layer 1 — remove the truncation condition

One documented constant, `AUXILIARY_TEXT_MAX_TOKENS = 2048`, applied to the five auxiliary text sites:

| call site | from | to |
| --- | --- | --- |
| `generateFastPathText` default (chat title) | 512 | 2048 |
| shared-thread title | 512 | 2048 |
| notification summary | 512 | 2048 |
| recommended follow-ups | 1024 | 2048 |
| run summary | 768 | 2048 |

A ceiling is not billed, only generated tokens are, and this model picks its own thinking length, so a higher ceiling does not buy more thinking — it stops the answer from being starved. Output length stays governed by the prompts ("max 60 chars", "max 90 chars", "at most 50 words"). Every measured call site at or above 1024 recorded zero failures; 2048 clears the observed 1024 failures with margin and stays far below the model's 65536 completion limit.

Not in scope and unchanged: `FAST_PATH_MODEL`, every `reasoning.effort` value, `voice_io_polish` (65536), `chat_initial_thinking` (1024), `chat_activity_summary` (1024), `people-search.service.ts` (a different provider). No retry on truncation was added, and #32117 rate-limit backoff and deduplication are untouched.

## Layer 2 — do not discard usable partial text

`OpenRouterGenerateTextOptions` gains `acceptTruncatedText`, and the generation surfaces `truncated`. The decision is per caller, deliberately not uniform:

- **Accept** — `notification_summary`, `run_summary`: a shortened sentence still informs the user, and the alternative is nothing.
- **Reject** — `chat_title`, `shared_thread_title`: a title is persisted immutably onto the thread and into public shared snapshots.
[...]

PR #32636, opened 2026-09-08
fix(api): classify auxiliary generation outcomes

lancy:
## Summary

Expected failures of five best-effort generations now preserve their existing fallback without per-invocation inner, outer, or null-result warnings. Genuine generation/output faults emit a bounded diagnostic; title/summary persistence and push-delivery failures retain their separate diagnostics. This implements the complete approved Design, non-goals, and acceptance contract at #32116, not its superseded historical retry suggestion.

Each finished logical generation makes one recording decision for a structured `auxiliary_generation_result` event in the existing Axiom `web-logs` telemetry dataset, including successful, cancelled, and intentionally skipped invocations. This provides event-derived counts/rates and generation-plus-interpretation duration; it does not introduce an OpenTelemetry counter or another metrics backend.

Closes #32116.

## Classification and ownership

- Quiet degradation requires transport-origin evidence: actual HTTP 429/408/502/503/504, the typed `rate_limit_exceeded`/429 response code (including synthetic 502 wrappers), or a bounded network/timeout cause at fetch/body I/O. Explicit authentication/invalid-request evidence takes precedence over a transient wrapper status. Unknown synthetic 502, arbitrary TypeError/TimeoutError, TLS configuration faults, malformed output, length/tool-call outcomes, and programming errors remain diagnostic.
- OpenRouter annotations retain only a finite reason in a weak map keyed by the original error. Shared helpers still throw the same error or return their existing result/null, with unchanged usage data. The safe diagnostic fields and reasoning compatibility fix already merged in #32612 are preserved.
- Only an identical caller abort reason is classified as cancellation. Existing signals are passed narrowly through summary/follow-up/notification generation; no new signal owner, deadline, wait budget, or retry is introduced. Late unrelated failures are still diagnosed.
- The generation boundary ends before DB writes or push delivery. Existing fallback values/text remain unchanged: absent background title/summary, empty follow-ups, generic push text, and the deterministic objective brief.
- Each result schedules ingest/flush through existing request-owned `waitUntil`. Completion after the response's ordinary flush is covered by a deferred SDK-flush test. Missing telemetry configuration and synchronous/asynchronous exporter errors cannot change the business result or cancellation. One recording decision does not promise exactly-once delivery across exporter/network failures.

## Caller and overlap inventory

Full symbol/call-site searches cover `generateText`, `generateTextWithUsage`, the title fast-path helper, and all five feature entry points, including ccstate command and callback dependency wiring.

| Surface | Callers and retained boundary |
| --- | --- |
[...]

PR #33143, opened 2026-09-10
fix(api): classify initial thinking rate limits as degraded

lancy:
Refs #32117.

## Problem

`chat-initial-thinking` was the last fast-path OpenRouter caller outside the auxiliary generation boundary that #32636 introduced, so it carried its own failure policy and produced no outcome telemetry.

### Evidence

Axiom `vm0-web-logs-prod`, `level == "error" or level == "warn"`, window `[2026-09-08T16:00:00Z, 2026-09-09T16:00:00Z)`:

- **62 warns**, every one from `fields.context = api:chat-initial-thinking` (56 with `fields.err.status = 429`, 6 reporting `502` with an inner `errorCode = 429`). No other OpenRouter warn survived in that window.
- The same window recorded **7,600** `auxiliary_generation_result` events, of which **272** were already silent `degraded` / `rate_limited` results (`recommended_followups` 92, `run_summary` 88, `notification_summary` 85, `chat_title` 7).
- Attempts per callback feature were identical (2,445 / 2,445 / 2,445): one completed run fans out to three concurrent fast-path requests.

**#33115 has since silenced the transient provider classes at this call site**, so the residual log noise is already fixed on `main`. What is still missing is the outcome telemetry — this call site has no attempt count, no success rate and no token accounting — and the second, separate classifier that #33115 had to introduce to get there.

## Change

- `chat_initial_thinking` joins the `AuxiliaryFeature` union and `generateAndPersistInitialThinkingMessage` calls `generateAuxiliary`, replacing #33115's `tapError` suppression. One classifier owns the judgement for all six fast-path generations instead of two. `generateInitialThinkingText` moves to `generateTextWithUsage` so it reports truncation and token counts like every other site.
- `isDegradedReason` delegates its provider classes to the shared `isTransientProviderFailure` predicate #33115 added, removing the duplicated reason list and keeping that export's only consumer.
- `retry_after_ms` is recorded on the outcome event and in the diagnostic, only when the provider actually sent `Retry-After`. `OpenRouterRequestError.retryAfterMs` has been parsed for a while but no fast-path caller reads it, so production currently cannot tell whether these 429s carry the header at all — honoring it first would have been unverifiable.
- `run_id` is recorded on the outcome event when the caller supplies one. One rate-limit window rejects several independent generations at once, so an event count alone overstates how many runs were reached.

### Behavior contract

| provider outcome | result |
| --- | --- |
| `rate_limited`, `provider_unavailable`, `network`, `upstream_timeout` (shared predicate) | `degraded`, telemetry only, no log |
| `output_truncated`, `unexpected_tool_calls` | `degraded`, telemetry only, no log — **changed at this call site, see below** |
| `unusable_output` (text sanitizes to nothing) | `error`, **still warns** |
| `auth`, `invalid_request`, `invalid_output`, `unknown` | `error`, **still warns** |
| caller aborted | `cancelled`, AbortError still propagates |

User-visible behavior is unchanged: a failed generation means no progress-copy bubble, exactly as before, and the run itself is untouched.

### One intentional reversal of #33115

#33115 deliberately kept an exhausted token budget reportable here, on the grounds that it describes our own request rather than the provider's availability. Routing through the shared boundary makes it `degraded` instead. That is intentional:

- The boundary has classified `output_truncated` as an expected capacity outcome for every other optional generation since #32979.
- Production recorded **zero** truncation failures at this call site across `[2026-09-03, 2026-09-10T03:00Z)` — the observed classes were `400` (7,548, fixed by #32612), `429` (100), `completion 502` (7) and `request 502` (1).
- The alternative is a per-caller exception inside shared code so that one of six features disagrees with the other five.

`THINKING_MAX_TOKENS` stays at 1024; the shared token budget belongs to #32118, and #32979 explicitly scoped this caller out. If that budget is raised later, the classification here already matches the rest.

`goal_objective_brief` was removed from the union by #33105 while this branch was open, so that part of the original change is now redundant and resolves to `main`'s deletion.

## Deliberately not in this PR

- **No retry and no backoff gate.** There is no retry in this path today, so the issue's "synchronized retries may worsen the provider limit" premise does not hold. Aggregate auxiliary load in the measured window was ~7 req/min (p95 16, peak 36) with rate limits confined to 74 of 1,085 active minutes, so retrying into the same window would double the failing requests rather than recover them. A bounded cooldown gate is a separate follow-up that should be decided on the `retry_after_ms` data this PR starts collecting.
- No alert-filter changes and no blanket downgrade of provider, auth, storage, or cancellation errors.

## Verification

Run locally against a real Postgres, one Vitest process at a time, after merging `main`:

- `chat-events.bdd.test.ts` — 501 passed. #33115's classification table is now the single table for this call site: it keeps all six of its cases, gains a rate limit wrapped in a choice-level synthetic 502 and output that sanitizes to nothing, and each case now asserts the telemetry `outcome`, `reason`, `run_id`, and that `retry_after_ms` is present only when the response carried the header, alongside the existing no-marker / run-still-`pending` / no-provider-text checks. My earlier duplicate table was removed rather than kept beside it.
- **Negative controls confirmed intact:** HTTP 400 (`invalid_request`) and HTTP 401 (`auth`) each still require exactly one `Auxiliary generation failed` warning, as does the full `openRouterErrorFixtures` 400 table. This is what proves the change did not silence the separate 400 class fixed by #32612.
- `chat-callbacks.bdd.test.ts`, `auxiliary-generation.test.ts`, `shared-threads.test.ts` — 123 passed. `chat-activity-summary.test.ts`, `run-lifecycle.bdd.test.ts`, `chat-translation.test.ts`, `chat-event-archive-consumers.test.ts` — 278 passed.
- `pnpm -F api check-types` (all six projects), `pnpm -F api lint`, Prettier, and Knip on `apps/api` all pass.

Assertions that pinned an exact `auxiliary_generation_result` list were scoped to the feature under test via the new optional `auxiliaryResults(context, feature)` argument, since a chat send now legitimately emits a `chat_initial_thinking` event alongside them. `chat-callbacks.bdd.test.ts` still asserts the unfiltered set where the fan-out itself is the subject.

### Production acceptance after release

Compare a full Beijing day against the window above, filtered to `vercel.environment == "production"`:

1. `feature == "chat_initial_thinking"` appears in `auxiliary_generation_result` with a visible outcome distribution — this is what proves the code is live rather than that no rate limit happened.
2. `feature == "chat_initial_thinking" and outcome == "error"` still shows `auth` / `invalid_request` samples when they occur naturally, confirming nothing was over-silenced.
3. `output_truncated` at this feature stays at or near zero; a non-trivial count is the signal to revisit `THINKING_MAX_TOKENS` under #32118.
4. The five existing features' `success` volume does not drop.
5. `retry_after_ms` presence is measured, which decides whether the follow-up honors `Retry-After` or uses a fixed bounded cooldown.


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: e92357bdbea02ccd58fb79027f1dc4c01936071d
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `e92357bdbea02ccd58fb79027f1dc4c01936071d`. Practice documents read from `main` at `e5e52075f4f2285ebaf93bd62b61465d91f05b9e`.

### Summary

Moves `chat-initial-thinking` — the last fast-path OpenRouter caller outside the auxiliary boundary — into `generateAuxiliary`, so a provider rate limit becomes a `degraded` / `rate_limited` telemetry outcome instead of a `warn` carrying the whole error object and its bundled stack. Adds `retry_after_ms` and `run_id` to the `auxiliary_generation_result` event so a later backoff decision has evidence, and drops the dead `goal_objective_brief` union member.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | `generateAndPersistInitialThinkingMessage` is reached only from the two `waitUntil` schedulers in `chat-events.command.ts`; both wrap it in `bestEffort`. The `undefined`/`null` collapse preserves the previous two-branch `return false`, and the later `runCanReceiveThinkingMessage` recheck is untouched. |
| `REVIEW.md` | Races, cancellation, ordering, and failure outcomes | Pass | No `AbortSignal` is passed, matching the sibling non-signal caller `generateAndPersistChatThreadTitle`. `settle`/`onRejection` still re-throw abort out through `bestEffort`, so the "suppresses late progress copy after answer/completed/cancelled" cases keep passing unchanged. |
| `REVIEW.md` | Meaningful coverage for new behavior; existing coverage still protects the refactor | Pass | Five new classification cases plus two rewritten diagnostics tests; `chat-events.bdd` 500/500. |
| `REVIEW.md` | Rewritten expectations must prove an intentional, correctly scoped replacement, not just a changed assertion | Pass | The rewritten tests keep the provider-data non-leak assertions and add negative controls that still require a warning for `auth` and `invalid_request`, so the silencing is provably scoped to non-actionable reasons. |
| `REVIEW.md` | Persisted data / API contract / `jsonb($type)` compatibility | N/A | No schema, contract, queue payload, request shape, or persisted state changes. The two added fields are additive keys on an outbound append-only Axiom event; a mixed fleet writes the old warn and the new event side by side into a schemaless dataset with no first-party reader. `docs/deployment-compatibility.md` was therefore not selected. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | `pnpm -F api check-types` (all six projects) and `pnpm -F api lint` pass with no new suppression comments. |
| `docs/bad-smell.md` | Catch only at a boundary that meaningfully handles the error; remove log-and-rethrow wrappers | Pass | The bespoke `tapError` + `log.warn` wrapper is replaced by the shared boundary that already owns classification. Best-effort ownership (`bestEffort`), the run-state rechecks, and abort propagation are preserved. |
| `docs/bad-smell.md` | Document changed public contracts and assess consumers | Pass | `AuxiliaryFeature` is module-local; every consumer (`chat-title`, `run-summary`, the test-helper enum) is updated in the same diff and typechecks. |
| `docs/bad-smell.md` | Static imports, centralized `env()`, no hardcoded URLs | Pass | Only static imports changed; no configuration or URL touched. |
| `docs/fallback.md` §3/§6 | No fallback for a state the owning contract already rules out | Pass | `if (generation === null) return null;` is required narrowing of `OpenRouterTextGeneration \| null` before reading `.usage`/`.text`, identical to `generateFastPathText` and `generateRunSummary` on `main`. No `??`/`\|\|` identity is invented. |
| `docs/fallback.md` §5/§10 | Retired producer's reader deleted with evidence, together with its contract entry and tests | Pass | `goal_objective_brief` has zero call sites after the goal retirement (full-repo grep); the union member and the test-helper enum entry are removed in the same diff. |
| `docs/fallback.md` §9 | Every introduced fallback declared in the PR summary and listed in the review | N/A | The diff introduces no fallback, so no `Fallbacks` section is required on either side. See the Fallbacks note below. |
| `docs/fallback.md` §1 | No negative tests asserting removed behavior stays removed | Pass (one observation) | The rewrites replace old assertions rather than keeping tombstones alongside them. See non-blocking observation 2 for the one borderline line. |
| `docs/testing.md` | Test the contract an external caller relies on; mock external services at their boundary | Pass | Setup goes through the production chat-send endpoint; OpenRouter is mocked with MSW at its real URL; outcomes are read back through `chat.listThreadEvents` and `api.readRun`. |
| `docs/testing/api-testing.md` | No DB schema imports, row writes, row reads, or direct service calls in API tests | Pass | The new and rewritten cases use only the existing BDD API helpers and context-owned mocks. |
| `docs/testing/api-testing.md` | Uniquely owned identities; no ordering or global-lock isolation | Pass | Each case creates its own actor/agent via `entitledChatActor()` and cancels its own run. |
| `docs/testing/testing-external-behavior.md` | Construct and verify through the production interface | Pass | The user-visible contract (no `thinking:initial` event, run still `pending`) is asserted through endpoints; Axiom is an external service, and its classification is the product contract this issue is about. |
| `docs/testing/anti-patterns.md` AP-9 | Assert meaningful content when logging is the contract | Pass | Assertions pin `feature`, `reason`, `errorKind`, `status`, `runId` and the enumerated diagnostic fields rather than a bare spy. |
| `docs/testing/anti-patterns.md` AP-1/AP-7 | Do not substitute mock calls for behavior | Pass | Telemetry assertions are paired with endpoint-observed run and thread state in every new case. |
| `docs/testing/anti-patterns.md` AP-8 | No duplicate coverage | Pass (one observation) | See non-blocking observation 1. |
| `docs/testing/patterns.md` | Reuse centralized mocks and shared teardown; MSW for external HTTP | Pass | Uses `context.mocks`, `server.use`, `mockOptionalEnv`, and shared teardown; no fake timers, no `fetch` stubbing, no local reset hooks. |
| `docs/testing/patterns.md` | Updating an existing test preserves meaningful security/recovery coverage | Pass | Both rewritten tests retain the `private_prompt_history_authorization_canary` non-leak checks, now applied to the whole logged payload rather than only `err.message`. |

### Findings

No P0 or P1 findings.

Non-blocking observations:

1. **`docs/testing/anti-patterns.md` AP-8, mild duplication.** `turbo/apps/api/src/signals/routes/__tests__/chat-events.bdd.test.ts` — the `503 → provider_unavailable` row re-proves reason classification that `auxiliary-generation.test.ts` already covers exhaustively through `chat_title`, and the case immediately above it already covers HTTP 503 for run usability. The other four rows carry unique value (429 is the issue's subject, the synthetic 502 pins the inner-429 semantics, 401 and `unusable_output` are the negative controls), so the table as a whole is justified. Not worth a new CI cycle.
2. **`docs/fallback.md` §1, borderline.** `expect(auxiliaryWarnings(context)[0]?.[1]).not.toHaveProperty("err")` reads partly as pinning the absence of the previous log shape. It is defensible as a payload-boundary assertion (the diagnostic must carry only enumerated fields, never a raw `Error`), and the adjacent `JSON.stringify(...).not.toContain(...)` check already covers the leak, so it is not a tombstone in the harmful sense.
3. **Readability.** In the same table test, `expect(...[0]).toStrictEqual(retryAfterMs === undefined ? expect.not.objectContaining({ retry_after_ms: expect.anything() }) : expect.anything())` is vacuous on the non-`undefined` branch. `expect(result?.retry_after_ms).toBe(retryAfterMs)` would assert presence and absence in one line. Cosmetic only; the absence case is genuinely covered today.
4. **Behavior note, intentional.** `usable: (value) => value !== null` means text that sanitizes to nothing now produces one `error` / `unusable_output` warning where the old code returned silently. This is the correct classification and matches all five sibling features; `parseOpenRouterGeneration` already rejects empty content upstream, so the reachable input is narrow (content consisting only of quotes/backticks/whitespace). It is covered by a dedicated case.
5. **Import ordering.** `./helpers/auxiliary-generation` is inserted after `./helpers/connector-credential-storage-state`. No lint rule enforces ordering here; noted only for tidiness.

### Fallbacks

This diff introduces no fallback, so no `Fallbacks` section is required in the PR summary (`docs/fallback.md` §9). For completeness, the two candidates were checked and rejected:

- `chat-initial-thinking.service.ts:generateInitialThinkingText` — `if (generation === null) return null;` is type narrowing required before reading `.usage`/`.text`, matching `generateFastPathText` and `generateRunSummary` on `main`. Not a fallback.
- `chat-initial-thinking.service.ts:generateAndPersistInitialThinkingMessage` — `thinking === undefined || thinking === null` narrows `string | null | undefined`; both states are produced by the contract (`undefined` = boundary swallowed the failure or skipped, `null` = sanitized to nothing) and both already returned `false` before this change. Not a fallback.

The removal of `goal_objective_brief` carries single-writer/zero-call-site evidence and needs no declaration (`docs/fallback.md` §10).

### Testing

Verification observed, all run against a real Postgres with one Vitest process at a time:

- `chat-events.bdd.test.ts` — 500 passed. New: `classifies $name for optional progress copy` covering an HTTP 429 with `retry-after: 12` (asserts `retry_after_ms: 12000`), a rate limit wrapped in a synthetic 502 (asserts `rate_limited`, not `provider_unavailable` — the exact conflation #32117 warns against), HTTP 503, HTTP 401, and output that sanitizes to nothing. Each asserts the outcome, `run_id` correlation, `retry_after_ms` presence only when the header was sent, no provider text in either the log or the metric, no `thinking:initial` event, and that the run stays `pending`.
- Negative controls confirmed intact: HTTP 401 (`auth`) and the full `openRouterErrorFixtures` HTTP 400 table (`invalid_request`) each still require exactly one `Auxiliary generation failed` warning. This is what proves the change did not silence the separate 400 class fixed by #32612.
- `chat-callbacks.bdd.test.ts` 55 passed, `auxiliary-generation.test.ts` 39 passed, `shared-threads.test.ts` 25 passed, `chat-activity-summary.test.ts` + `run-lifecycle.bdd.test.ts` 271 passed, `chat-translation.test.ts` + `chat-event-archive-consumers.test.ts` passed.
- `pnpm -F api check-types`, `pnpm -F api lint`, Prettier, and Knip on `apps/api` all pass.

Assertions that pinned an exact `auxiliary_generation_result` list were narrowed to the feature under test via the new optional `auxiliaryResults(context, feature)` argument. That is the correct scoping — a chat send now legitimately emits a `chat_initial_thinking` event alongside the feature being asserted — and `chat-callbacks.bdd.test.ts` still asserts the unfiltered set where the test's subject is the full fan-out.

No material coverage gap. Retry, backoff, and a cooldown gate are explicitly out of scope for this PR by prior decision and are not treated as missing behavior; the `retry_after_ms` field added here is the evidence input for that follow-up.

lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: a16c24a58dab8847d7cb04a6361a920c3e5e237d
PR_AUTO_REVIEW_VERDICT: LGTM

Re-review of the new HEAD `a16c24a58dab8847d7cb04a6361a920c3e5e237d` after the merge from `main`; supersedes the review of `e92357bd`. Practice documents read from `main` at `2a7efd4d585f59bbe0694a90d342dd2f93c2da30` (`REVIEW.md` and all eight selected documents are byte-identical to the previously read revision).

### Summary

Routes `chat-initial-thinking` through the shared `generateAuxiliary` boundary, replacing the bespoke `tapError` suppression #33115 added at the same call site so one classifier owns the judgement for all six fast-path generations. Adds `retry_after_ms` and `run_id` to `auxiliary_generation_result`, and makes `isDegradedReason` delegate its provider classes to #33115's `isTransientProviderFailure`.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | Reached only from the two `waitUntil` schedulers in `chat-events.command.ts`, both wrapped in `bestEffort`. The merge correctly keeps `main`'s `historicalRunGroupId` and `@okouai/db/runtime/agent-run`; the `undefined`/`null` collapse preserves the previous two-branch `return false` and the later `runCanReceiveThinkingMessage` recheck. |
| `REVIEW.md` | Races, cancellation, ordering, and failure outcomes | Pass | No `AbortSignal` is passed, matching the sibling non-signal caller `generateAndPersistChatThreadTitle`; `settle`/`onRejection` still re-throw abort out through `bestEffort`. The "suppresses late progress copy after answer/completed/cancelled" cases pass unchanged. |
| `REVIEW.md` | Rerouted flow requires regression coverage and correct scoping; a rewritten expectation alone is not proof | Pass | See Findings — this is the one item that needed real scrutiny, and the evidence goes well beyond a rewritten assertion. |
| `REVIEW.md` | Preserve actionable reporting for nearby genuine failures; no blanket downgrade | Pass | `auth`, `invalid_request`, `invalid_output`, `unknown` and `unusable_output` all still warn. The 400 and 401 rows plus the whole `openRouterErrorFixtures` table are explicit negative controls. |
| `REVIEW.md` | Meaningful coverage for new behavior; existing coverage still protects the refactor | Pass | #33115's table is retained in full and extended, not replaced; `chat-events.bdd` 501/501. |
| `REVIEW.md` | Persisted data / API contract / `jsonb($type)` compatibility | N/A | No schema, contract, queue payload, request shape, or persisted state changes. The two added fields are additive keys on an outbound append-only Axiom event; a mixed fleet writes old and new shapes into a schemaless dataset with no first-party reader. `docs/deployment-compatibility.md` was therefore not selected. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | `pnpm -F api check-types` (six projects) and `pnpm -F api lint` pass. The `unicorn/switch-case-braces` violation the switch initially introduced was fixed at source with braces, not a suppression. |
| `docs/bad-smell.md` | Catch only at a boundary that meaningfully handles the error; no duplicate classifiers | Pass | This is the core of the change: two classifiers for the same call site collapse to one. Best-effort ownership, the run-state rechecks, and abort propagation are preserved. |
| `docs/bad-smell.md` | Document changed public contracts and assess consumers | Pass | `AuxiliaryFeature` is module-local and every consumer is updated in the same diff. `isTransientProviderFailure` gains its only consumer here; Knip on `apps/api` is clean. |
| `docs/bad-smell.md` | Static imports, centralized `env()`, no hardcoded URLs | Pass | Only static imports changed. |
| `docs/fallback.md` §3/§6 | No fallback for a state the owning contract already rules out | Pass | `if (generation === null) return null;` is required narrowing before reading `.usage`/`.text`, identical to `generateFastPathText` and `generateRunSummary`. No `??`/`\|\|` identity is invented. |
| `docs/fallback.md` §5/§10 | Retired producer's reader deleted with evidence, with its contract entry and tests | Pass | #33105 removed `goal_objective_brief` from the production union but left it in the test-helper enum on `main`; this diff removes that leftover, completing the deletion. |
| `docs/fallback.md` §9 | Every introduced fallback declared in the PR summary and listed in the review | N/A | The diff introduces no fallback; see the Fallbacks note. |
| `docs/fallback.md` §1 | No negative tests asserting removed behavior stays removed | Pass | The rewrites replace assertions in place. The previous review's borderline `not.toHaveProperty("err")` line survives and is still a payload-boundary assertion, not a tombstone. |
| `docs/testing.md` | Test the contract an external caller relies on; mock external services at their boundary | Pass | Setup goes through the production chat-send endpoint; OpenRouter is mocked with MSW at its real URL; outcomes are read back through `chat.listThreadEvents` and `api.readRun`. |
| `docs/testing/api-testing.md` | No DB schema imports, row writes, row reads, or direct service calls | Pass | Only existing BDD API helpers and context-owned mocks. |
| `docs/testing/api-testing.md` | Uniquely owned identities; no ordering or global-lock isolation | Pass | Each case creates its own actor/agent via `entitledChatActor()` and cancels its own run. |
| `docs/testing/testing-external-behavior.md` | Construct and verify through the production interface | Pass | The user-visible contract (no `thinking:initial` event, run still `pending`) is asserted through endpoints; Axiom is an external service and its classification is the product contract under review. |
| `docs/testing/anti-patterns.md` AP-8 | No duplicate coverage | Pass | Materially improved since the last review: the duplicate table flagged there is deleted and its two unique cases are folded into #33115's table, which is now the single table for this call site. |
| `docs/testing/anti-patterns.md` AP-9 | Assert meaningful content when logging is the contract | Pass | Assertions pin `feature`, `reason`, `errorKind`, `status`, `runId`, `run_id`, `retry_after_ms` rather than a bare spy. |
| `docs/testing/anti-patterns.md` AP-1/AP-7 | Do not substitute mock calls for behavior | Pass | Telemetry assertions are paired with endpoint-observed run and thread state in every case. |
| `docs/testing/patterns.md` | Reuse centralized mocks and shared teardown; MSW for external HTTP | Pass | Uses `context.mocks`, `server.use`, `mockOptionalEnv`, shared teardown; no fake timers, no `fetch` stubbing. |
| `docs/testing/patterns.md` | Updating an existing test preserves meaningful security/recovery coverage | Pass | Both rewritten tests retain the `private_prompt_history_authorization_canary` non-leak checks, and the unified table keeps the no-provider-text check on both the log and the metric. |

### Findings

No P0 or P1 findings.

**On the intentional reversal of #33115 (`output_truncated`).** `REVIEW.md` requires that a rerouted flow carry regression coverage and that a rewritten expectation alone not stand in for an intentional, correctly scoped replacement. Judged against that bar, this one clears it:

- **The path keeps its coverage.** #33115's `an exhausted token budget` case is not deleted. It still drives the real chat-send path, still asserts no `thinking:initial` marker, a `pending` run, and no provider text in the log — only the classification changes, and the new expectation is asserted positively as `degraded` / `output_truncated` rather than merely dropped.
- **The rationale is recorded in three places**, not just the assertion: the test comment, the PR summary, and the merge commit message.
- **It is a consistency correction, not a new policy.** `output_truncated` has been `degraded` for the other five fast-path features since #32979. The alternative is a per-caller exception inside shared code so that one of six disagrees.
- **Scope is exactly one reason at one call site**, reached through the shared predicate. Nothing else in `isDegradedReason` changes behavior: old and new predicates return `true` for the same six reasons.
- **The signal is relocated, not destroyed.** This is the substantive answer to #33115's concern. Truncation remains queryable in production as `auxiliary_generation_result` with `reason: "output_truncated"` and now a `run_id`; it moves from a log line to a metric. The PR's acceptance criteria explicitly include watching that counter and treating a non-trivial count as the trigger to revisit `THINKING_MAX_TOKENS` under #32118.

Residual risk, stated plainly: this caller still runs at `THINKING_MAX_TOKENS = 1024` while the other five run at 2048, so it is the one most likely to truncate, and it no longer warns when it does. Production recorded zero truncation events here across `[2026-09-03, 2026-09-10T03:00Z)` (observed classes were `400` ×7548, `429` ×100, `completion 502` ×7, `request 502` ×1), so the change is not silencing anything currently firing, and the metric plus the named acceptance check is adequate containment. Not a blocker.

Non-blocking observations:

1. **Readability, carried over.** `expect(result?.retry_after_ms).toBe(retryAfterMs)` replaced the awkward `expect.not.objectContaining(...)` form flagged in the previous review. Observation 3 from that review is resolved.
2. **`docs/fallback.md` §1, borderline, unchanged.** `expect(auxiliaryWarnings(context)[0]?.[1]).not.toHaveProperty("err")` still reads partly as pinning the absence of the previous log shape, and is still defensible as a payload-boundary assertion given the adjacent leak check.
3. **Behavior note, intentional.** `usable: (value) => value !== null` means text that sanitizes to nothing produces one `error` / `unusable_output` warning where the old code returned silently. Correct classification, matches all five siblings, and now has its own case in the unified table.
4. **Import ordering.** `./helpers/auxiliary-generation` sits after `./helpers/connector-credential-storage-state`; no lint rule enforces ordering. Cosmetic.

### Fallbacks

This diff introduces no fallback, so no `Fallbacks` section is required in the PR summary (`docs/fallback.md` §9). Both candidates were checked and rejected:

- `chat-initial-thinking.service.ts:generateInitialThinkingText` — `if (generation === null) return null;` is type narrowing required before reading `.usage`/`.text`, matching `generateFastPathText` and `generateRunSummary`. Not a fallback.
- `chat-initial-thinking.service.ts:generateAndPersistInitialThinkingMessage` — `thinking === undefined || thinking === null` narrows `string | null | undefined`; both states are produced by the contract and both already returned `false`. Not a fallback.

The test-helper removal of `goal_objective_brief` completes #33105's deletion and needs no declaration (`docs/fallback.md` §10).

### Testing

Verification observed on the merged HEAD, against a real Postgres, one Vitest process at a time:

- `chat-events.bdd.test.ts` — **501 passed**, including the reclassified token-budget case. #33115's table is now the single classification table for this call site: six original cases retained, plus a rate limit wrapped in a choice-level synthetic 502 (asserts `rate_limited`, not `provider_unavailable` — the exact conflation #32117 warns against) and output that sanitizes to nothing. Every case now asserts `outcome`, `reason`, `run_id`, and that `retry_after_ms` is present only when the response carried the header (`retry-after: 12` → `12000`), alongside the pre-existing marker/run/no-leak checks.
- `chat-callbacks.bdd.test.ts`, `auxiliary-generation.test.ts`, `shared-threads.test.ts` — 123 passed. `chat-activity-summary.test.ts`, `run-lifecycle.bdd.test.ts`, `chat-translation.test.ts`, `chat-event-archive-consumers.test.ts` — 278 passed.
- `pnpm -F api check-types`, `pnpm -F api lint`, Prettier on the changed files, and Knip on `apps/api` all pass. Knip is the check that would have caught `isTransientProviderFailure` becoming a zero-consumer export; it is clean because `isDegradedReason` now delegates to it.

Assertions that pinned an exact `auxiliary_generation_result` list are scoped to the feature under test via the new optional `auxiliaryResults(context, feature)` argument; `chat-callbacks.bdd.test.ts` still asserts the unfiltered set where the fan-out itself is the subject, and the unfiltered form is also used for the leak checks, which makes them strictly stronger.

No material coverage gap. Retry, backoff, and a cooldown gate are out of scope by prior decision and are not treated as missing; `retry_after_ms` is the evidence input for that follow-up.

lancy:
Superseded by #33241 after GitHub reported `viewerCanReopen=false` when restoring this PR following the workspace-wide CI pause. The same rebased branch continues in #33241.

Issue #32963, opened 2026-09-09
bug(runtime): track OOM / exit 137 causes across Pi and Claude


lancy:
## Current canonical scope — September 11, 2026

**OPEN: one OOM / exit-137 incident tracker, with framework-specific cases and unresolved obligations.** Owner: @lancy. Priority: **high**, retained from the newly consolidated #33352.

At @lancy's explicit request, #33352 is consolidated here as **case D (Claude Code)** alongside the historical **Pi cases A/B/C** from #32577, #32668 and #32802. This is unified tracking, **not proof of a common cause or a claim that these incidents are fixed**. The original source issues preserve their evidence.

### Latest evidence and remaining work

[September 11 observation and complete consolidation record](https://github.com/vm0-ai/vm0/issues/32963#issuecomment-5628389326) is the current status:

- Fixed production window: September 9 08:45 through September 11 08:45 Beijing. 164 observed Pi sandbox runs: 84 maintenance / 80 other; all 19 Pi runs with new memory evidence are maintenance. Seven distinct Pi error captures have exit 1 and no OOM evidence; no new Pi exit 137 was observed. This excludes pure API-first traffic and does not validate a new long-tool-loop cohort.
- Captured kernel OOM evidence belongs to 89 distinct runs: 81 Codex / 8 Claude Code / 0 Pi. It is not a count of failed agent tasks.
- **Case D has a correlated Guest kernel victim: PID 387, comm=claude, in workload/runtime, CONSTRAINT_MEMCG at the workload boundary.** The tools domain dominates the retained charge, but its exact pressure-producing tool/allocator remains unassigned. This event predates #33240's merge and is not a demonstrated regression of that fix.
- #33240 and #33287 are merged and controller code-accepted; actual artifact-specific rollout/production acceptance remains distinct. The collector/path repair now yields canonical stored evidence; it does not establish the historical Pi root cause.
- Carry all #33352 acceptance requirements: deployed Claude artifact identity, truthful terminal classification, agent-domain/tool-only/missing/malformed evidence controls, and pressure-source investigation with honest limits.
- Suggested review point: **September 16**, with roughly 30–50 representative non-maintenance Pi long runs as an engineering exposure target, not a statistical guarantee or automatic closure. A quiet maintenance-only interval is insufficient. Claude case D's unresolved obligations cannot be closed by absence of Pi recurrence. No reminder, automation, production replay or induced OOM was authorized or created.

[...]

PR #32936, opened 2026-09-09
fix(guest): canonicalize oom cgroup paths

lancy:
## Summary

Guest control serialized `vm0-exec/...` while the telemetry contract and kernel records require `/vm0-exec/...`. An ordinary mixed telemetry request was rejected with `metrics.0.memory.groups.0.cgroup: Invalid input`, and an independently recorded absolute kernel path could not match the workload.

Convert the trusted operation root once at the collector boundary and share that cgroup-root-absolute identity across initial/current/incident snapshots and kernel correlation. Keep filesystem reads on the original operation directory and keep the API schema unchanged. Reject wrong mounts, traversal, oversized and non-operation paths without manufacturing an owned identity.

Closes #32921. Historical investigations #32577, #32668 and #32802 remain separate and unresolved.

## Regression coverage

- A checked fixture is regenerated and compared by the production collector using real temporary counter files and a separately specified trusted mount path. Only nondeterministic UUIDs, wall-clock timestamps and inodes are stabilized; paths, counters, null/partial states and kernel records remain producer output.
- The unchanged telemetry contract accepts that output in ordinary mixed metric/system-log/operation batches and urgent/fallback evidence payloads. Endpoint tests check the actual metrics/system/operation datasets, stable fallback incident identity, strict malformed/cross-run rejection, and positive version ACK only after configured successful ingestion. Unconfigured, failed and partial ingestion never acknowledges evidence.
- Kernel input paths are independent absolute literals for workload, runtime, aggregate tools, tool leaf and control victims under global/cgroup constraints. Existing source, sequence, lifecycle, counter, cleanup and delivery-bound tests remain in place.

## Acceptance against #32921

| Criterion | Evidence / boundary |
| --- | --- |
| 1. Canonical identity and original reads | One validated operation-root conversion feeds all three snapshot roles and the kernel reader; safe fixture values confirm the original read location. |
| 2. Producer regression test | `collector_output_matches_api_fixture` executes the production conversion and collector, then compares serialized output with the checked API fixture. Old-code negative control fails at the leading-slash assertion. |
| 3. API contract and ingestion | Four contract tests and six focused endpoint cases pass; strict production API schemas and projection code are unchanged. Ordinary mixed batches, urgent/fallback ACK, old senders, malformed/cross-run evidence and unsuccessful ingestion are covered. |
[...]

Issue #32921, opened 2026-09-09
fix(guest): canonicalize oom cgroup paths to restore telemetry ingestion


lancy:
## Background

Repair a confirmed regression introduced by #32845 (the implementation of #32818): OOM evidence uses a relative cgroup path where both the API contract and Guest kernel records require a cgroup-root-absolute path. This blocks telemetry ingestion and also prevents canonical kernel OOM records from matching the current workload.

Controller: https://app.okou.ai/chats/87131dab-34d7-4868-bff9-0d46de5def34

The user explicitly approved this focused repair and `/handoff-issue-to-thread`. This is a new repair issue, not a finding that resolves the historical OOM allocation source in #32577, #32668, or #32802.

### Source evidence

Source baseline re-read before filing: `99bd2254849a4890c0bf7dc12c151b6463b34726`. The affected collector and API schema remain unchanged from accepted merge `075ccbaaf8f90231687267218ef1e77d05771d3a`, shipped in release #32841, source `82d1a6ff154c9ca81fd8e08d6764290733eb324d`.

- [`snapshots`, `crates/guest-control-server/src/oom_evidence.rs:486`](https://github.com/vm0-ai/vm0/blob/99bd2254849a4890c0bf7dc12c151b6463b34726/crates/guest-control-server/src/oom_evidence.rs#L486) applies `Path::strip_prefix("/sys/fs/cgroup")` and serializes the relative result without restoring the leading `/`.
- [`crates/guest-control-server/src/oom_evidence.rs:199`](https://github.com/vm0-ai/vm0/blob/99bd2254849a4890c0bf7dc12c151b6463b34726/crates/guest-control-server/src/oom_evidence.rs#L199) passes the same sampled workload path into the kernel reader; [`parse_oom:323`](https://github.com/vm0-ai/vm0/blob/99bd2254849a4890c0bf7dc12c151b6463b34726/crates/guest-control-server/src/oom_evidence.rs#L323) uses exact/prefix ownership matching against the kernel's absolute `task_memcg` path. An owned absolute path cannot match the emitted relative workload identity.
- [`turbo/packages/api-contracts/src/contracts/oom-evidence.ts:17`](https://github.com/vm0-ai/vm0/blob/99bd2254849a4890c0bf7dc12c151b6463b34726/turbo/packages/api-contracts/src/contracts/oom-evidence.ts#L17) correctly requires `/vm0-exec/exec-.../workload...`; its lifecycle checks must remain strict.
[...]

PR #33163, opened 2026-09-10
fix(api): keep sandbox telemetry batches alive and name oom upload failures

lancy:
Refs #32963

## Problem

`metricDataSchema.memory` reused the same strict `oomEvidenceSchema` as the dedicated top-level `oomEvidence` field, and the guest embeds a memory snapshot in every ordinary 5s metrics entry. One unusable snapshot therefore rejected the entire telemetry request with HTTP 400, discarding the `systemLog`, `metrics`, `networkLogs`, and `sandboxOperations` batched in the same request.

That is not hypothetical. While the relative-cgroup-path defect (#32921 / #32936) was live, production sandbox telemetry coverage collapsed. Distinct `runId` per hour on 2026-09-09 (UTC), in both `vm0-sandbox-telemetry-metrics-prod` and `vm0-sandbox-telemetry-system-prod`:

| hour | 05 | 06 | 07 | 08 | 09 | 10 | 11 |
| --- | --- | --- | --- | --- | --- | --- | --- |
| runs | 146 | 195 | 50 | **6** | **1** | 117 | 132 |

It was also self-silencing: the guest's own `OOM evidence upload unacknowledged` warning is written to the system log, which was uploaded in the same rejected request. That message has zero occurrences in the dataset.

`#32936` removed the trigger, not the coupling.

Separately, the Runner's `guest oom evidence upload unacknowledged; host evidence retained` warning reported no stage, so a rejected payload, an unconfigured receiver, a truncated body, and a slow response were indistinguishable.

## Change

- `metricDataSchema.memory` degrades instead of rejecting: an unusable snapshot drops itself and the rest of the batch is ingested.
- The dedicated `oomEvidence` field stays strict. It decides the delivery acknowledgement, so an unusable payload must be rejected rather than falsely acknowledged.
- The Runner warning now reports a fixed `reason` (`timeout`, `transport`, `http_status`, `body_oversize`, `invalid_body`, `missing_ack`) and, for a rejected request, the HTTP status.

## Scope limits

- The warning keeps its `warn` level. It reports a real delivery failure, and its current production rate is zero; downgrading it would hide a regression.
- No change to upload timing, retries, timeout budgets, sampling, or the acknowledgement contract.
- Evidence payloads and response bodies stay out of the log.

## Verification

- Updated the mixed-telemetry test: an unusable snapshot now yields 200 with the snapshot dropped, while the system log, remaining metric fields, and sandbox operations are still ingested.
- Added a negative control: a request whose dedicated `oomEvidence` carries relative cgroup paths still returns 400, ingests nothing, and is not acknowledged. The existing strict-`oomEvidence` rejection cases are unchanged.

Not run locally: Vitest and the Rust test suites; both are left to CI.

### DIFF
--- a/crates/runner/src/telemetry.rs
+++ b/crates/runner/src/telemetry.rs
@@ -33,6 +33,40 @@ const FLUSH_THRESHOLD: Duration = Duration::from_secs(30);
 const TELEMETRY_TIMEOUT: Duration = Duration::from_secs(10);
 const RUNNER_VERSION: &str = env!("CARGO_PKG_VERSION");
 
+/// Why one bounded OOM evidence upload was not acknowledged.
+///
+/// Only fixed stage names and an HTTP status are reported. Evidence payloads
+/// and response bodies stay out of the log.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+enum OomEvidenceUploadFailure {
+    Timeout,
+    Transport,
+    HttpStatus(u16),
+    BodyOversize,
+    InvalidBody,
+    MissingAck,
+}
+
+impl OomEvidenceUploadFailure {
+    fn reason(self) -> &'static str {
+        match self {
+            Self::Timeout => "timeout",
+            Self::Transport => "transport",
+            Self::HttpStatus(_) => "http_status",
+            Self::BodyOversize => "body_oversize",
+            Self::InvalidBody => "invalid_body",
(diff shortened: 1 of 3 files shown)


lancy:
Superseded by #33287 after the authorized rebase. GitHub rejected reopening this closed PR because the branch was recreated; all further review and CI evidence will be recorded on #33287.

PR #33146, opened 2026-09-10
feat(app): show the template cover on the composer's inline template chip

tongx-blip:
## What

The composer's inline template chip now leads with the chosen template's cover
image instead of a generic glyph.

## Why

`createInlineTemplateNodeView` already receives `previewImageUrl` on the node —
`selectedComposerTemplateAttachment` resolves it for presentation, illustration,
avatar and explainer templates — but the view never rendered it. A user picks a
template from a wall of covers in the picker and gets back an orange word, so
there is no way to tell two decks apart in the draft without opening the picker
again. The older block-style chip (`createTemplateAttachmentNodeView`) already
draws that cover, so this brings the inline chip in line with it.

## How

- Render the cover in an 18px rounded slot at the head of the chip. 18px inside
  the 28px inline chip keeps the same optical inset the 20px cover has inside the
  32px block chip.
- Keep the SwatchBook glyph for templates whose catalog entry has no cover
  (video, website, workflow).
- Resolve the slot inside `render()` rather than at construction, because the
  picker rewrites the node in place via `setNodeMarkup` when it replaces a
  selection — the cover has to follow the new attributes.

No feature switch: this is a visual refinement of a chip that already ships.

## Verification

- `vitest run src/views/okou-page/__tests__/chat-composer-create.test.tsx` — 17 passed
- `pnpm run check-types` (platform)
- `eslint` on both changed files

Two tests added to `chat-composer-create.test.tsx`:
- a presentation template's cover appears on the chip and follows a replacement
  selection;
- a video template (no catalog cover) keeps the glyph.

PR #33184, opened 2026-09-10
feat(app): show the template cover on the inline template chip behind a switch

tongx-blip:
## What

Adds `composerTemplateChipCover` (Lab -> Beta, staff orgs). With it on, the
composer's inline template chip leads with the chosen template's cover image
instead of a generic glyph. With it off, nothing changes.

## Why

`createInlineTemplateNodeView` already receives `previewImageUrl` on the node —
`selectedComposerTemplateAttachment` resolves it for presentation, illustration,
avatar and explainer templates — but the view never rendered it. A user picks a
template from a wall of covers in the picker and gets back an orange word, so
two decks are indistinguishable in the draft without reopening the picker. The
older block-style chip (`createTemplateAttachmentNodeView`) already draws that
cover.

This replaces #33146, which carried the same change ungated and was closed.

## How

- New key + registry entry, `enabled: false` + `STAFF_ORG_ID_HASHES`, so it
  lands in Lab under Beta and is off for everyone else.
- The node view renders the cover in an 18px rounded slot when the switch is on
  and a cover exists; otherwise the existing SwatchBook glyph. 18px inside the
  28px inline chip keeps the cover at the proportion the block chip uses (20px
  inside 32px).
- The slot is resolved inside `render()`, because the picker rewrites the node
  in place via `setNodeMarkup` when it replaces a selection.
- `WorkflowComposerRuntime.templateChipCover` is a **reader function**, not a
  latched boolean. Latching it in the mount command reproducibly rendered the
  glyph with the switch on: the authoritative feature-switch read resolves after
  the composer mounts, so the mount-time value is always the pre-hydration
  default. The switch-off test below is what caught this.

## Verification

- `vitest run src/views/okou-page/__tests__/chat-composer-create.test.tsx` — 18 passed
- `vitest run src` in `packages/core` — 244 passed
- `pnpm run check-types` (platform), `eslint` on changed files, `prettier --check`

Three tests added:
- switch off — a presentation template still shows no cover;
- switch on — the cover appears and follows a replacement selection;
- switch on, video template (no catalog cover) — the glyph is kept.


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: fd57ade2f632cd16e8fc4431a32d8c81eabbf556
PR_AUTO_REVIEW_VERDICT: LGTM

Reviewed HEAD: `fd57ade2f632cd16e8fc4431a32d8c81eabbf556`
Practice revision: `bd0f81c3c9baddabf67a28f01c57d6a16b1f8f63`

### Summary

Adds `composerTemplateChipCover` (Lab -> Beta, staff orgs). With it on, the composer's inline template chip leads with the chosen template's cover image instead of the SwatchBook glyph; with it off, or when the template's catalog entry has no cover, the glyph is unchanged. This HEAD also merges current `main` to resolve an append conflict in `chat-composer-create.test.tsx`.

Effective diff against `main`: 160 insertions across 4 files.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | The sole consumer is the `inlineTemplate` NodeView. `templateAttachmentNodeAttributes` omits `previewImageUrl` only when the stored attr is `null`, and the attr validator admits only `null` or `string`, so `coverUrl === undefined` is the exact discriminator for both the switch-off and no-cover cases. |
| `REVIEW.md` | Distinguish introduced defects from unchanged code, including against the newly merged `main` | Pass | #33142 reworked `ComposerCreateControls` into an in-composer combobox + `listbox`, but did not touch `createInlineTemplateNodeView`; on `main` that view is still `openButton.append(icon, title)`. The two changes do not overlap. |
| `REVIEW.md` | Require meaningful coverage for new behavior | Pass | Three tests cover switch-off, switch-on with a cover including an in-place replacement, and switch-on with a coverless template. |
| `REVIEW.md` | Feature/capability containment for a change to an existing flow | Pass | Every new pixel is inside `actions.coverEnabled()`; the off path reproduces today's DOM. |
| `.claude/skills/feature-switch/SKILL.md` | New user-facing behavior is gated behind a feature switch | Pass | Key added to `FeatureSwitchKey`, registered with `maintainer`, `description`, `enabled: false`, `enabledOrgIdHashes: STAFF_ORG_ID_HASHES`, which resolves to the `beta` Lab group. |
| `.claude/skills/feature-switch/SKILL.md` | Client-side gating reads `featureSwitch$` | Pass | Read through `featureSwitch$` in the mount command and surfaced to the node view as `runtime.templateChipCover`. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | Neither appears in the diff. |
| `docs/bad-smell.md` | Catch errors only at a meaningful boundary; no fabricated defaults | N/A | No error handling added. |
| `docs/bad-smell.md` | Document changed public contracts and assess consumers | N/A | `WorkflowComposerRuntime` and `InlineTemplateNodeActions` are module-local interfaces; all three implementers (`createWorkflowComposerRuntime`, `resetMountedWorkflowRuntime`, `createMountEditorCommand`) are updated in this diff. No exported contract changed. |
| `docs/fallback.md` | Every fallback the diff introduces is declared and justified | Pass | See Fallbacks. |
| `docs/fallback.md` | No negative tests pinning removed behavior | Pass | All three tests assert live behavior on one side of the switch. |
| `docs/styles.md` | Business styling composes Tailwind utilities; no first-party selectors or injected stylesheets | Pass | Utility strings on `className`, matching the surrounding NodeView constants. |
| `docs/testing.md` | Prefer integration coverage through real entry points | Pass | All three tests boot the page via `setupPage` and drive the real slash command, picker dialog, and chip. |
| `docs/testing/app-testing.md` | Page-level integration tests through the bootstrapped Router | Pass | In `views/okou-page/__tests__`, configure -> setup -> observe -> act -> assert. |
| `docs/testing/testing-external-behavior.md` | Verify through the page; no CSS-class, store, or callback assertions | Pass | Assertions read the rendered cover's `src` and its absence. `querySelector("img")` is necessary because the decorative cover carries `alt=""` and exposes no role — the same reason the pre-existing `composerInlineTemplates()` helper uses a `data-*` query. |
| `.claude/skills/ccstate/SKILL.md` | Every async operation has an owner and cancellation path; no store mutation during render | Pass | The addition is synchronous DOM work inside an existing NodeView owned by its `destroy()`. No signal, command, `detach()` or async work is introduced. |
| `docs/deployment-compatibility.md` | Cross-version compatibility for persisted state | N/A | No new persisted field. `previewImageUrl` is an existing node attribute with `default: null`; drafts written by older builds carry the same values and both `null` and `string` are handled. |

### Findings

No P0 or P1 findings.

### Fallbacks

The diff adds one branch on an absent value:

- `tiptap-workflow-composer.ts:createInlineTemplateNodeView` — renders the SwatchBook glyph when the resolved cover is absent. **Surface / gate: none.** Two distinct causes collapse into this branch and neither is a compatibility fallback: (a) the switch is off, which is the rollout gate itself and is exempt under `docs/fallback.md` §2; (b) `previewImageUrl` is genuinely absent, which is declared optional on `ComposerTemplateAttachment` and deliberately omitted by the single writer `selectedComposerTemplateAttachment` for `video`, `workflow` and `website` templates. Under §3/§4 that is a genuinely optional value rather than a state the contract rules out, and under §6.4 the glyph is a presentational default with no identity meaning. Neither has a removal condition of the kind §9 asks an author to record, so no `Fallbacks` PR-body section is required.

**Verdict: justified, and correctly not declared.**

### Testing

- `The template chip cover stays off until the Lab switch is on` — the gate itself. This test is what caught the original defect: latching the switch in the mount command made the cover never appear, because the authoritative feature-switch read resolves *after* the composer mounts, so the latched value was always the pre-hydration default.
- `An inline template chip shows the chosen cover and follows a replacement` — covers the in-place `setNodeMarkup` path that motivated resolving the cover inside `render()`. The slugs used (`playful-launch-presentation` -> `business-data-presentation`) are mutually non-substring, so the second `toContain` genuinely discriminates.
- `A template with no cover keeps the template glyph on its chip` — the coverless branch via the video picker.
- Local verification on this merged HEAD: `chat-composer-create.test.tsx` 19 passed (including #33142's new in-composer picker test), `packages/core` 243 passed, platform `check-types`, full `pnpm run lint`, `prettier --check`, `git diff --check`.

**Non-blocking observations**

- `runtime.templateChipCover` closes over the mount command's `get`, so the node view reads the switch without subscribing to it. That is deliberate and the interface comment says so — a Lab toggle applies to the next chip render rather than repainting existing chips live — and it matches this file's existing pattern of invoking the mount command's context from `runtime.*` callbacks after the command resolves.
- `rounded-[3px]` sits outside the registered radius scale; a single decorative crop radius, accepted by lint.
- An empty-string `previewImageUrl` would take the cover branch and set `img.src = ""`. Nothing in the current writers produces it and the type does not prevent it; not a demonstrated defect.
- After this change the composer chip shows a cover while the sent message for the same template still shows only the glyph (`chat-thread-page.tsx`). Pre-existing asymmetry, out of scope here.

PR #33185, opened 2026-09-10
feat(app): list the create types directly in the slash menu behind a switch

tongx-blip:
## What

Adds `composerCreateFlatSlash` (Lab -> Beta, staff orgs). With it on, typing `/`
lists **Create presentation**, **Create video** and **Create image** as ordinary
menu entries. With it off, the existing `Create ▸` entry and its type-chooser
row are untouched.

## Why

`Create ▸` is a second presentation of the same three options, and it presents
them somewhere else: `ComposerCreateControls`' `choosing` branch mounts a
`min-h-11` row with `pb-3` directly above `ComposerCard`. Entering and leaving
that row moves the composer by 56px, and while it is open `canSend` is forced
false, so the send control goes disabled with content already in the draft.

Removing the level makes the choice one keystroke instead of two and takes the
composer's position out of the interaction.

This replaces #33151, which removed the chooser outright and was closed. Gating
it means both paths stay live, so the two can be compared in Lab before either
is deleted.

## How

- New key + registry entry, `enabled: false` + `STAFF_ORG_ID_HASHES`, so it
  lands in Lab under Beta and is off for everyone else.
- `useComposerCreateSuggestions` returns `composer.create.modes` for a `create`
  query when the switch is on, and the existing `"choose"` entry when it is off.

That is the whole behavioural change. `choosing$`, the chooser row, its Escape
handler and the `choosingCreateType$` send guard are all still in place and are
still reached with the switch off; deleting them belongs to the switch's
graduation, not to this PR.

## Verification

- `vitest run src/views/okou-page/__tests__/chat-composer-create.test.tsx` — 17 passed
- `vitest run src` in `packages/core` — 244 passed
- `pnpm run check-types` (platform), `eslint` on changed files, `prettier --check`

Two tests added:
- switch off — `/create` still offers exactly one `Create` entry and it opens
  the `Choose a type` group;
- switch on — `/create` lists the three types, picking one sets the mode in a
  single step, no `Choose a type` group is ever rendered, and the slash text is
  consumed.

The existing two-step tests are unchanged and cover the switch-off default.

### DIFF
--- a/turbo/apps/platform/src/views/okou-page/__tests__/chat-composer-create.test.tsx
+++ b/turbo/apps/platform/src/views/okou-page/__tests__/chat-composer-create.test.tsx
@@ -640,3 +640,66 @@ test("Canceling and switching Create preserve slash text and template references
     }),
   );
 });
+
+async function setupComposerWithFlatSlash(
+  flatSlash: boolean,
+): Promise<HTMLElement> {
+  await setupPage({
+    context,
+    path: `/agents/${AGENT_ID}/chat`,
+    featureSwitches: {
+      [FeatureSwitchKey.ComposerCreateCommands]: true,
+      [FeatureSwitchKey.ComposerCreateFlatSlash]: flatSlash,
+    },
+  });
+  return await findComposerEditor();
+}
+
+test("The slash menu keeps the Create type chooser until the Lab switch is on", async () => {
+  setupModels();
+  mockChatLifecycle(context);
+  const editor = await setupComposerWithFlatSlash(false);
+  await fill(editor, "Our launch /create");
+  const menu = await screen.findByTestId("slash-workflow-menu");
+  expect(
+    queryAllByRoleFast("button", menu)
+      .map((item) => {
+        return item.getAttribute("aria-label");
+      })
+      .filter((label) => {
+        return label?.startsWith("Create");
+      }),
+  ).toStrictEqual(["Create"]);
+  click(button("Create", menu));
+  await screen.findByRole("group", { name: "Choose a type" });
+});
+
+test("The slash menu lists every create type in one step when the switch is on", async () => {
+  setupModels();
+  mockChatLifecycle(context);
+  const editor = await setupComposerWithFlatSlash(true);
+  await fill(editor, "Our launch /create");
+  const menu = await screen.findByTestId("slash-workflow-menu");
+  expect(
+    queryAllByRoleFast("button", menu)
+      .map((item) => {
+        return item.getAttribute("aria-label");
+      })
+      .filter((label) => {
+        return label?.startsWith("Create");
+      }),
+  ).toStrictEqual(["Create presentation", "Create video", "Create image"]);
+
+  click(button("Create video", menu));
+  await waitFor(() => {
+    expect(screen.getByTestId("composer-create-mode")).toHaveTextContent(
+      "Create video",
+    );
+  });
+  // The second row of type buttons is what moved the composer; it must never
+  // appear on this path.
+  expect(screen.queryByRole("group", { name: "Choose a type" })).toBeNull();
+  expect(screen.queryByTestId("slash-workflow-menu")).toBeNull();
+  expect(editor).toHaveTextContent("Our launch");
+  expect(editor).not.toHaveTextContent("/create");
+});
--- a/turbo/apps/platform/src/views/okou-page/tiptap-workflow-composer.tsx
+++ b/turbo/apps/platform/src/views/okou-page/tiptap-workflow-composer.tsx
@@ -306,6 +306,8 @@ function useComposerCreateSuggestions(
 ): readonly ComposerCreateCommand[] {
   useTranslation();
   const enabled = useGet(composer.create.enabled$);
+  const flatSlash =
+    useGet(featureSwitch$)[FeatureSwitchKey.ComposerCreateFlatSlash] === true;
   if (!enabled || query === undefined) {
     return [];
   }
@@ -314,7 +316,10 @@ function useComposerCreateSuggestions(
     "create".startsWith(normalized) ||
     composerCreateCommandLabel("choose").toLowerCase().startsWith(normalized)
   ) {
-    return ["choose"];
+    // Under ComposerCreateFlatSlash the menu answers the query in one level.
+    // The `choose` entry exists only to open a second row of type buttons above
+    // the composer, which moves the composer by its own height.
+    return flatSlash ? composer.create.modes : ["choose"];
   }
   return composer.create.modes.filter((mode) => {
     return (
--- a/turbo/packages/core/src/feature-switch-key.ts
+++ b/turbo/packages/core/src/feature-switch-key.ts
@@ -81,4 +81,5 @@ export enum FeatureSwitchKey {
   AvatarFraming = "avatarFraming",
   ConnectorDirectory = "connectorDirectory",
   ChatThreadHeaderActions = "chatThreadHeaderActions",
+  ComposerCreateFlatSlash = "composerCreateFlatSlash",
 }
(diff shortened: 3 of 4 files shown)


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: 86d9342dbdb545ae0e1aa8a803d08865b003d579
PR_AUTO_REVIEW_VERDICT: LGTM

Reviewed HEAD: `86d9342dbdb545ae0e1aa8a803d08865b003d579`
Practice revision: `f570332481b4164fe42dbe18c53dbe2e218e7661`

### Summary

Adds `composerCreateFlatSlash` and uses it in `useComposerCreateSuggestions`: for a `create` query the slash menu returns the three create modes instead of the synthetic `"choose"` entry that opens a second row of type buttons above the composer. The switch is `enabled: false` + `STAFF_ORG_ID_HASHES`, so it lands in Lab under Beta and every existing path is reached unchanged when it is off.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | The only consumer is `ComposerSuggestionMenuState`. `selectSuggestion` / `scrollSuggestionIntoView` already index `createModes[index]` then `workflowSuggestions[index - createModes.length]`, so a list longer than one entry is the arithmetic they were written for. |
| `REVIEW.md` | Distinguish introduced defects from unchanged code | Pass | `composer.create.modes` cannot be empty — `COMPOSER_CREATE_MODES.filter` always keeps `presentation`, and only `image`/`video` are conditional — so the flat branch cannot render an empty Create group. |
| `REVIEW.md` | Require meaningful coverage for new behavior; regression coverage for a rerouted flow | Pass | Both switch states are covered end to end, and the pre-existing two-step tests still exercise the off path unmodified. |
| `REVIEW.md` | Feature/capability containment for a change that reroutes an existing flow | Pass | The reroute is entirely inside the switch branch; with the switch off `useComposerCreateSuggestions` returns exactly what it returned before. |
| `.claude/skills/feature-switch/SKILL.md` | New user-facing behavior is gated behind a feature switch | Pass | Key added to `FeatureSwitchKey`, registered in `FEATURE_SWITCHES` with `maintainer`, `description`, `enabled: false`, `enabledOrgIdHashes: STAFF_ORG_ID_HASHES`. |
| `.claude/skills/feature-switch/SKILL.md` | Client-side gating reads `featureSwitch$` in the component that renders the entry | Pass | `useGet(featureSwitch$)[FeatureSwitchKey.ComposerCreateFlatSlash] === true`, the same shape `useComposerWorkflowSuggestions` uses for `ComposerWorkflowFuzzySearch` two functions below in the same file. |
| `.claude/skills/feature-switch/SKILL.md` | Default state fits the rollout intent | Pass | `enabled: false` + staff org hashes resolves to `beta` in `getFeatureSwitchRolloutStage`, which is the Lab group this is meant to be evaluated in. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | Neither appears in the diff. |
| `docs/bad-smell.md` | Catch errors only at a meaningful boundary; no fabricated defaults | N/A | No error handling added. |
| `docs/bad-smell.md` | Document changed public contracts and assess consumers | N/A | The only exported surface added is a new enum member and its registry entry; no existing signature or contract changed. |
| `docs/fallback.md` | A feature behind a non-GA switch needs no compatibility code | Pass | The `flatSlash ? modes : ["choose"]` branch is the rollout gate itself, not a tolerance branch for another version's data. No dual read/write, no migration, no old-shape handling. |
| `docs/fallback.md` | Every fallback the diff introduces is declared and justified | Pass | No fallback is introduced — see Fallbacks. |
| `docs/fallback.md` | No negative tests pinning removed behavior | Pass | The switch-off test asserts the chooser is still *reached and functional*, which is live product behavior on the default path, not a tombstone for deleted code. |
| `docs/testing.md` | Prefer integration coverage through real entry points | Pass | Both tests boot the page with `setupPage` and drive the real editor, slash menu, and chooser. |
| `docs/testing/app-testing.md` | Page-level integration tests entering through the bootstrapped Router | Pass | Tests live in `views/okou-page/__tests__`, follow the configure -> setup -> observe -> act -> assert order, and reuse the file's existing `setupPage` helpers. |
| `docs/testing/testing-external-behavior.md` | Verify through the page; do not assert on CSS classes, store state, or internal callbacks | Pass | Assertions read menu entries by accessible name, the `composer-create-mode` chip's text, the presence of the `Choose a type` group, and the editor's text. No hook, store, or class assertion. |

### Findings

No P0 or P1 findings.

### Fallbacks

The diff introduces none. `flatSlash ? composer.create.modes : ["choose"]` is a feature-switch branch selecting between two live UI paths, not a code path handling a state the system is not supposed to produce. Under `docs/fallback.md` §2 the off branch is the current GA behavior rather than compatibility code for a previous iteration, and it carries no removal gate of its own: it disappears when the switch graduates and the chooser is deleted. No `Fallbacks` PR-body section is required.

### Testing

- `The slash menu keeps the Create type chooser until the Lab switch is on` — with the switch off, `/create` still offers exactly one `Create` entry and clicking it opens the `Choose a type` group. This is the regression guard for the default path.
- `The slash menu lists every create type in one step when the switch is on` — the menu lists the three types, one click sets the mode, no `Choose a type` group is rendered, the menu closes, and the `/create` text is consumed from the editor.
- The existing two-step tests in this file are unmodified and continue to cover the off path.

**Non-blocking observations**

- The switch-on test selects with a click. Keyboard selection across several create entries is not asserted on this branch, but it is not new surface either: the pre-existing filter branch already returns multiple modes for a query such as `e` (`create presentation`, `create video` and `create image` all contain it), so `selectSuggestion`'s index arithmetic is exercised today without the switch.
- `composerCreateCommandLabel("choose")` is still evaluated in the condition when the switch is on. It is a pure label lookup and costs nothing, but it only carries meaning for the off path; it can go when the chooser does.

tongx-blip:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 86d9342dbdb545ae0e1aa8a803d08865b003d579

Reviewed HEAD: `86d9342dbdb545ae0e1aa8a803d08865b003d579`
**This supersedes my LGTM above on the same commit.** That review was correct against the `main` this branch was cut from; it is not correct against `main` today.

### Summary

Blocking on new evidence from `main`, not on anything in this diff. #33142 (`feat(chat): move create scene picker inside composer`, 69ce756, merged 05:33Z by @bingjiez666) reworked the same surface, and it removes the behaviour this PR exists to fix.

### Findings

**P1 — the PR's stated justification no longer describes `main`**

This PR's rationale, repeated in the body, in the switch `description`, and in the code comment, is:

> the `choose` entry exists only to open a second row of type buttons above the composer, which moves the composer by its own height

That was true at `f570332`. On `main` today it is not:

- `ComposerCreateControls` is now rendered **inside** `CardContent`, between `ComposerImportedTemplateUrlRefreshLifecycle` and `ComposerAttachments` (`chat-composer.tsx:11009`). It is no longer a sibling mounted above `ComposerCard`.
- Its container is `px-4 pt-4 pb-1` inside the card, not the `min-h-11 ... pb-3` row above it. The 56px external shift is gone.
- The type list is no longer a row of buttons; it is an in-card `listbox` (`ComposerCreatePicker`, rendered at `chat-composer.tsx:9500`) opened from a `combobox` trigger with `aria-expanded` / `aria-controls`.

So the switch-off branch this PR preserves is no longer the behaviour the PR describes, and the comparison a reviewer or a Lab tester would make is not the comparison the body sets up.

### What is *not* the problem

- The merge conflict itself is trivial and not a reason to block: `feature-switch-key.ts` and `feature-switch.ts` auto-merge cleanly, and the only conflict is two appended test blocks at the end of `chat-composer-create.test.tsx` (mine vs #33142's), which is a keep-both resolution.
- Required CI was fully green on this HEAD (82 passed / 20 skipped) after the `dev-12.gcp.vm3.ai` SSH-tunnel failures were rerun. The earlier `prepare` / `deploy-runner-prepare` / `ci-gate-turbo` failures were CI host load and are resolved.

### Why I stopped instead of rebasing

Rebasing would land a feature switch whose description, comment and rationale all cite a layout shift that no longer happens, next to a design a teammate merged an hour ago for the same problem. Rewriting the premise is a re-scope and a product call, not evidence-based repair, so it needs @tongx-blip and @bingjiez666 rather than an automated merge.

The open question is whether one level in the slash menu is still wanted **on top of** the in-composer picker, or whether #33142 already closes this out.

PR #33186, opened 2026-09-10
fix(app): rank compact workflow fuzzy matches first

bingjiez666:
## Summary

Typing `/prauto` could rank `pr-32809-app-production-watch` above `pr-auto` because all subsequence matches kept their original list order. Rank these fuzzy matches by fewer skipped characters, then an earlier match position and a shorter name. Equally relevant results retain their existing order.

Exact, prefix, and substring priorities, numeric identifier matching, and the existing `composerWorkflowFuzzySearch` switch remain unchanged. Matching ranges continue to drive the name highlights.

## Verification

- Reproduced the reported ordering failure with the new page regression test before applying the fix.
- All 17 tests in `chat-composer-workflows.test.tsx` pass, including result order, highlights, and Enter inserting `/pr-auto`.
- App TypeScript checks, focused ESLint/Oxlint checks, App Knip, formatting, and `git diff --check` pass.
- Full test suites are delegated to the PR pipeline; no local development server was started.
- The [App preview](https://pr-33186-app-okou-app-preview.vm0.workers.dev/) loads the sign-in page in Chromium. Its deployed merge SHA is `df7c6dae7bb01c88586f18daba725436ca062e06`, which includes PR head `6dda977845a5e831a4697692cc60fd2fc8ce11ad`.
- The [preview API deployment](https://github.com/vm0-ai/vm0/actions/runs/34442842053/job/102761404237) is blocked by Neon's concurrent-active-endpoint limit during database setup. Full browser acceptance requires that environment to recover.

## Acceptance

1. In a preview account with `composerWorkflowFuzzySearch` enabled, make `pr-auto` and `pr-32809-app-production-watch` available to the selected agent.
2. Type `/prauto` in the chat composer. `pr-auto` should appear first, with `pr` and `auto` highlighted.
3. Press Enter. The composer should contain `/pr-auto` and close the suggestion menu.

### DIFF
--- a/turbo/apps/platform/src/signals/okou-page/workflow-composer-domain.ts
+++ b/turbo/apps/platform/src/signals/okou-page/workflow-composer-domain.ts
@@ -9,6 +9,8 @@ export function findWorkflowQueryMatches(
   const matches: {
     workflow: ComposerSlashWorkflowMatch;
     rank: number;
+    skippedCharacters: number;
+    start: number;
   }[] = [];
 
   for (const workflow of workflows) {
@@ -22,23 +24,35 @@ export function findWorkflowQueryMatches(
         },
         rank:
           fuzzy && normalizedName === normalizedQuery ? 0 : start === 0 ? 1 : 2,
+        skippedCharacters: 0,
+        start,
       });
       continue;
     }
     if (fuzzy && normalizedQuery.length >= 3) {
-      const matchRanges = findWorkflowSubsequence(
-        normalizedName,
-        normalizedQuery,
-      );
-      if (matchRanges) {
-        matches.push({ workflow: { ...workflow, matchRanges }, rank: 3 });
+      const match = findWorkflowSubsequence(normalizedName, normalizedQuery);
+      if (match) {
+        matches.push({
+          workflow: { ...workflow, matchRanges: match.ranges },
+          rank: 3,
+          skippedCharacters: match.skippedCharacters,
+          start: match.start,
+        });
       }
     }
   }
 
   return matches
     .sort((left, right) => {
-      return left.rank - right.rank;
+      if (left.rank !== 3 || right.rank !== 3) {
+        return left.rank - right.rank;
+      }
+      // Prefer compact abbreviations, then earlier matches and shorter names.
+      return (
+        left.skippedCharacters - right.skippedCharacters ||
+        left.start - right.start ||
+        left.workflow.name.length - right.workflow.name.length
+      );
     })
     .map((match) => {
       return match.workflow;
@@ -48,15 +62,26 @@ export function findWorkflowQueryMatches(
 function findWorkflowSubsequence(
   name: string,
   query: string,
-): readonly WorkflowMatchRange[] | null {
+): {
+  readonly ranges: readonly WorkflowMatchRange[];
+  readonly skippedCharacters: number;
+  readonly start: number;
+} | null {
   const ranges: WorkflowMatchRange[] = [];
   let offset = 0;
+  let skippedCharacters = 0;
+  let matchStart = 0;
 
   // Keep numeric identifiers contiguous while allowing letters to skip ahead.
   for (const [part] of query.matchAll(/\d+|\D/g)) {
     const start = name.indexOf(part, offset);
     if (start === -1) {
       return null;
+    }
+    if (ranges.length === 0) {
+      matchStart = start;
+    } else {
+      skippedCharacters += start - offset;
     }
     offset = start + part.length;
     const previous = ranges.at(-1);
@@ -67,7 +92,7 @@ function findWorkflowSubsequence(
     }
   }
 
-  return ranges;
+  return { ranges, skippedCharacters, start: matchStart };
 }
 
 export interface SlashWorkflowRange {
--- a/turbo/apps/platform/src/views/okou-page/__tests__/chat-composer-workflows.test.tsx
+++ b/turbo/apps/platform/src/views/okou-page/__tests__/chat-composer-workflows.test.tsx
@@ -573,6 +573,57 @@ test("Rank exact workflow names before prefixes, substrings, and abbreviations",
   });
 });
 
+test("Rank compact workflow abbreviations before scattered matches", async () => {
+  mockAgent();
+  mockThread();
+  installWorkflows(() => {
+    return [
+      workflow("pr-32809-app-production-watch"),
+      workflow("team-pr-auto"),
+      workflow("pr-auto-bravo"),
+      workflow("pr-auto-alpha"),
+      workflow("pr-auto"),
+    ];
+  });
+
+  await setupPage({
+    context,
+    path: `/chats/${THREAD_ID}`,
+    featureSwitches: { [FeatureSwitchKey.ComposerWorkflowFuzzySearch]: true },
+  });
+
+  const user = userEvent.setup();
+  const editor = await findComposerEditor();
+  await user.click(editor);
+  await user.keyboard("/prauto");
+
+  await waitFor(() => {
+    expect(slashWorkflowNames()).toStrictEqual([
+      "/pr-auto",
+      "/pr-auto-bravo",
+      "/pr-auto-alpha",
+      "/team-pr-auto",
+      "/pr-32809-app-production-watch",
+    ]);
+  });
+  const highlighted = Array.from(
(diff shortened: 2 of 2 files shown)


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 6dda977845a5e831a4697692cc60fd2fc8ce11ad
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `6dda977845a5e831a4697692cc60fd2fc8ce11ad`
Practice revision: `f570332481b4164fe42dbe18c53dbe2e218e7661` (`main`)

### Summary

`findWorkflowSubsequence` now returns the number of characters skipped between matched query parts and the index of the first matched part, and `findWorkflowQueryMatches` uses those values to order fuzzy (rank 3) matches by fewest skipped characters, then earliest match position, then shorter name. Exact/prefix/substring ranks, contiguous numeric-identifier matching, the `composerWorkflowFuzzySearch` gate, and the highlight ranges are unchanged.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | `findWorkflowQueryMatches` has exactly one consumer, `turbo/apps/platform/src/views/okou-page/tiptap-workflow-composer.tsx:345`, which renders the returned order directly; `findWorkflowSubsequence` is module-private with no other caller. |
| `REVIEW.md` | Ordering/comparator correctness and no regression of unchanged ranks | Pass | The comparator falls through to `left.rank - right.rank` whenever either side is not rank 3, so ranks 0–2 keep their existing relative order; the rank-3 tiebreak is a consistent total preorder and `Array.prototype.sort` stability preserves input order for equal keys, matching the stated "equally relevant results retain their existing order". |
| `REVIEW.md` | Require meaningful coverage for the regression fix | Pass | `chat-composer-workflows.test.tsx:576` adds a page-level regression test for the reported `/prauto` ordering failure; the pre-existing exact/prefix/substring, contiguous-numeric, and strict-matching tests remain unmodified. |
| `REVIEW.md` | Feature/capability containment for changed user-facing behavior | Pass | Rank 3 is only produced inside the `fuzzy && normalizedQuery.length >= 3` branch, so the new ordering is fully contained by the existing `ComposerWorkflowFuzzySearch` switch; no switch was added, widened, or retired. |
| `REVIEW.md` | Persisted data, API contracts, and deployment compatibility | N/A | The diff touches one pure client-side domain function and one Platform test; no schema, contract, request/queue payload, persisted state, or service-worker surface changes. |
| `REVIEW.md` / `docs/fallback.md` | New fallbacks must be declared in a `Fallbacks` PR-body section | N/A | No fallback is introduced. `matchStart` is a loop initializer that is always assigned before return, because the `normalizedQuery.length >= 3` guard guarantees at least one query part; it is not a default for a state the contract rules out. |
| `docs/bad-smell.md` | No `any`; preserve inference, narrow at untrusted boundaries | Pass | The new return type is an explicit readonly object literal type; no `any` or assertion is introduced in `workflow-composer-domain.ts`. |
| `docs/bad-smell.md` | No lint/type suppressions | Pass | No `eslint-disable`, `oxlint-disable`, `@ts-ignore`, `@ts-expect-error`, or `prettier-ignore` appears in the diff. |
| `docs/bad-smell.md` | Error handling: no fabricated defaults hiding a broken invariant | Pass | The `start === -1` miss still returns `null` and aborts the match; no `catch`, log-and-rethrow wrapper, or fabricated value is added. |
| `docs/bad-smell.md` | Interface changes: document changed public contracts and assess consumers | Pass | The changed signature belongs to the module-private `findWorkflowSubsequence`; the exported `findWorkflowQueryMatches` signature and `ComposerSlashWorkflowMatch` shape are unchanged, so no consumer contract moves. |
| `docs/bad-smell.md` | Static imports; no hardcoded URLs or configuration | N/A | The diff adds no imports, URLs, or environment-specific values. |
| `docs/testing.md` | Prefer integration coverage through real entry points; cover new behavior and regressions | Pass | The new case boots the real Router via `setupPage`, types `/prauto` into the production composer, and asserts the rendered suggestion order — the exact reported failure. |
| `docs/testing.md` | Avoid duplicate cases that restate configuration or pin incidental implementation | Pass | The new test adds a distinct ordering scenario (compact vs. scattered subsequence) not covered by the existing exact/prefix/substring or contiguous-numeric tests. |
| `docs/testing.md` | Mock external services at their boundary; no internal relative `vi.mock` | Pass | Workflows are supplied through the typed `context.mocks.api(workflowsCollectionContract.list, ...)` helper reused from `installWorkflows`; no `fetch`, global MSW server, or internal module mock is introduced. |
| `docs/testing.md` | Avoid fake timers; wait for observable results | Pass | The test synchronizes with `waitFor` on rendered suggestion names and composer text; no fake timers, `vi.setSystemTime`, or arbitrary delay is used. |
| `docs/testing/testing-external-behavior.md` | Construct cases through page interactions and verify through the page | Pass | Setup is a real click plus `user.keyboard("/prauto")`, and verification reads the rendered menu entries and the composer's text after `{Enter}`; no store mutation, hook call, or signal assertion. |
| `docs/testing/app-testing.md` | Page-level test in `views/**/__tests__`; `setupPage` awaited before assertions | Pass | The case lives in `src/views/okou-page/__tests__/chat-composer-workflows.test.tsx` and awaits `setupPage` and `findComposerEditor` before acting. |
| `docs/testing/app-testing.md` | Avoid `getByRole`/`findByRole` for text-named roles such as `button`; use `queryAllByRoleFast` | Pass | Menu entries are read through the file's existing `slashWorkflowNames()`/`slashButton()` helpers, which are built on `queryAllByRoleFast`. |
| `docs/testing/app-testing.md` | `waitFor` callbacks contain queries and assertions only | Pass | Both `waitFor` blocks contain only queries and expectations; the `{Enter}` keypress and the highlight read are outside them. |
| `docs/testing/app-testing.md` | Use `userEvent` only when real keyboard behavior matters, and await every operation | Pass | Typing and `{Enter}` selection are genuine keyboard behavior, and every `user.click`/`user.keyboard` call is awaited — the same pattern as the adjacent abbreviation tests. |

### Findings

No P0 or P1 findings.

Non-blocking observations (no change requested):

1. `turbo/apps/platform/src/signals/okou-page/workflow-composer-domain.ts:29-30` — the exact/prefix/substring branch now sets `skippedCharacters: 0` and `start`, but the comparator only reads those fields when **both** sides are rank 3, so for ranks 0–2 they are never used. This keeps one uniform record shape and costs nothing at runtime; a discriminated union would be stricter but more code. Recording it only so a future reader does not mistake those fields for live inputs to the rank 0–2 ordering.
2. `turbo/apps/platform/src/views/okou-page/__tests__/chat-composer-workflows.test.tsx:591-597` — expecting `/pr-auto-bravo` before `/pr-auto-alpha` pins sort stability for fully tied entries. That is the behavior the PR body states ("equally relevant results retain their existing order") and `Array.prototype.sort` guarantees it, so this is intentional rather than incidental; noting it because it is the one expectation that depends on input ordering rather than on a scoring rule.

### Testing

- Coverage matches the fix: `Rank compact workflow abbreviations before scattered matches` reproduces the reported `/prauto` case with `pr-32809-app-production-watch` present, and asserts the full ordering, the `pr` + `auto` highlight ranges, and that `{Enter}` inserts `/pr-auto` and closes the menu.
- Manual verification of the scoring against the fixtures agrees with the assertion: `pr-auto`, `pr-auto-bravo`, and `pr-auto-alpha` each skip 1 character from index 0 (tie broken by name length, then stable input order), `team-pr-auto` skips 1 from index 5, and `pr-32809-app-production-watch` skips 16 — producing exactly the expected order.
- Regression protection for unchanged ranks is retained by the untouched `Rank exact workflow names before prefixes, substrings, and abbreviations`, `Keep numeric workflow identifiers contiguous in abbreviated queries`, and `Keep strict workflow matching ... when fuzzy search is ...` cases.
- No material gap identified. The changed function is pure and reachable only through the composer page, which the suite already drives end to end.
- CI note (not a review finding): on this HEAD, `test-app (8)` failed on an unrelated 5s timeout in `chat-capability-voice-forwarding.test.tsx`, `deploy-api` failed on a Neon `53000` "exceeded the limit of concurrently active endpoints" error, and `test-app (2)` was cancelled after passing all 31 of its test files. None of these touch the changed surface.

PR #33112, opened 2026-09-10
feat(api): answer a named category with the whole category

tongx-blip:
Entering a category showed **twelve** connectors under a filter that said **71**, and the small heading above them said something else again.

## The count and the contents came from different places

The count is the category's real size (`categoryConnectorCounts`). The contents were the **browse slice** — discovery returns at most twelve per category so no single category can crowd the first screen. Offering a number as the way in and then not keeping it makes the number worse than useless.

Discovery now takes a `category` and answers with that whole category, ranked, still excluding the connectors Okou runs for itself. The keyword-free browse slice that renders the shelves is unchanged, and the page asks for the category it is showing.

## The heading that named something else

Inside a category the page still drew the grouped list: a group heading (`AI`) and a section heading (`General models and reasoning`) above the cards — so a category opened under a heading with a different name, twice. The category view now renders the cards alone; the breadcrumb and the filter already name where you are.

## Verification

- `@okouai/api`, `@okouai/api-contracts`, `@okouai/app` — `check-types` and `lint` clean; `pnpm knip` clean
- 42 platform tests across `connectors-catalog`, `connector-directory`, `chat-composer-connectors`
- `connectors-catalog.test.tsx` proves the fix at the page boundary: the fixture now slices per category the way production does, the shelf shows six of sixteen, and entering the category renders all sixteen
- New API case "answers a named category with the whole category" asserts `connectors.length === categoryConnectorCounts[category]`, single category, and rank order. The DB-backed API suite needs Postgres, which this sandbox has no server for (`ECONNREFUSED 5432`); it runs in CI.

PR #33189, opened 2026-09-10
fix(app): show the whole category in the composer connector directory

tongx-blip:
The connectors page stopped promising 329 and delivering 12 in #33112. **The composer's directory still did.**

Its chips carry the category totals from `categoryConnectorCounts`, but its connectors came from the keyword-free browse response, which holds at most twelve of any one category. So picking a chip that said **329** opened a list of **12**.

## What changed

- **The dialog asks discovery for the category by name**, exactly as the page does, and gets all of it.
- Scoping the fetch moved a second problem into view: the chip row was built from the connectors currently shown, so entering a category left **one chip and no way to pick another**. It now comes from the browse response — what the page already does with its own chip row.

## The fixtures were describing a response the API never returns

Both connector fixtures returned every connector, unsliced, which is why neither surface's category view could be tested. They now behave like discovery:

- a named category → the whole category
- a keyword-free browse → **every connected connector** plus a slice of each category

That second clause matters: `discoveryEffectiveConnectors` puts connected connectors ahead of the slice, so an agent with twenty-five connected connectors gets all twenty-five. Getting this wrong is what briefly broke three unrelated composer-popover tests, and getting it right is what makes the new case able to fail.

## Verification

- `@okouai/app` `check-types`, `lint`, `pnpm knip` — clean
- 129 tests across `connector-directory`, `connectors-catalog`, `connectors-ssh`, `connectors-accounts`, `chat-composer-connectors`, `chat-composer-connector-accounts`, `chat-composer-connector-loading`
- New case "Show the whole category the chip counted, not the browse slice": the chip reads `Mail16`, entering it renders sixteen cards, and the row still offers `Voice50` so another category is reachable

## Feature switch

Unchanged: `connectorDirectory`.

### DIFF
--- a/turbo/apps/platform/src/signals/okou-page/connectors.ts
+++ b/turbo/apps/platform/src/signals/okou-page/connectors.ts
@@ -119,18 +119,12 @@ export interface ComposerConnectorSignals {
   readonly accounts: ComposerConnectorAccountSignals;
 }
 
-const relatedConnectorCatalogKeyword$ = computed(() => {
-  return "";
-});
-
-const composerRelatedCatalog$ = relatedConnectorCatalog(
-  relatedConnectorCatalogKeyword$,
-);
-
-/** A category-scoped read asks for no keyword, so it browses that category. */
+/** Browse reads ask for no keyword; the category, when set, scopes them. */
 const emptyCatalogKeyword$ = computed(() => {
   return "";
 });
+
+const composerRelatedCatalog$ = relatedConnectorCatalog(emptyCatalogKeyword$);
 
 const composerRelatedCatalogItems$ = computed(async (get) => {
   return (await get(composerRelatedCatalog$)).connectors;
--- a/turbo/apps/platform/src/signals/okou-page/connectors.ts
+++ b/turbo/apps/platform/src/signals/okou-page/connectors.ts
@@ -127,6 +127,11 @@ const composerRelatedCatalog$ = relatedConnectorCatalog(
   relatedConnectorCatalogKeyword$,
 );
 
+/** A category-scoped read asks for no keyword, so it browses that category. */
+const emptyCatalogKeyword$ = computed(() => {
+  return "";
+});
+
 const composerRelatedCatalogItems$ = computed(async (get) => {
   return (await get(composerRelatedCatalog$)).connectors;
 });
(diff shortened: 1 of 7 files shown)


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: c9e1f21dff1654e9c5a85ed5557cc3d613bd1792
PR_AUTO_REVIEW_VERDICT: LGTM

Reviewed HEAD: `c9e1f21dff1654e9c5a85ed5557cc3d613bd1792` · practice revision `4803ac067beffb4501a0b15160bd39a87430adf8`

### Summary

The composer directory's chips carried category totals while its connectors came from the keyword-free browse response, so a chip reading 329 opened a list of twelve — the defect #33112 fixed on the connectors page. The dialog now asks discovery for the chosen category by name, its chip row is built from the browse response so it survives entering a category, and both connector fixtures were corrected to describe the response the API actually returns.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | `createComposerConnectorSignals` gains `categoryCatalog$`; `addDialogCatalogItems$` picks search → category → browse in that order, so the keyword path and the shelf browse are untouched. The dialog's `chipCatalog` comes from `connectorData.relatedCatalogItems`, which is still the browse response because `composerRelatedCatalog$` does not depend on the category. |
| `REVIEW.md` | Removing/replacing a source of data preserves what depended on it | Pass | `model.categorySections` still drives the shelves and `categoryLabelOf`; only the chip row moves to `chipSections`, and the SSH chip that `main` appends is appended to the new list in the same place. |
| `REVIEW.md` | Regression coverage for the fixed defect | Pass | "Show the whole category the chip counted, not the browse slice" reads `Mail16` off the chip, enters it, asserts sixteen cards, and asserts `Voice50` is still offered — the chip-row collapse would fail that last assertion. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | None added. |
| `docs/bad-smell.md` | Reuse over duplication | Pass | The follow-up commit folds the second empty-keyword computed back into the existing one, so browse reads share a single signal. |
| `docs/bad-smell.md` | §4 interface changes assessed against consumers | Pass | `ConnectorDirectoryModel` gains `chipSections` and the dialog gains a `chipCatalog` prop; both are app-internal with one call site each, updated here. |
| `docs/fallback.md` | §2/§9 no compatibility code for a non-GA switch; declare any fallback | N/A | No fallback introduced; the category read is a plain request the API already answers (#33112). |
| `docs/deployment-compatibility.md` | Cover the old/new combinations a rollout produces | Pass | No contract change. Against an API without the `category` parameter the query is stripped and the dialog shows the browse slice — today's behaviour, not an error — and the surface is behind a non-GA switch. |
| `docs/testing.md` | Prefer integration coverage through real entry points | Pass | The new case boots the composer with `setupPage`, opens the dialog, and clicks the chip. |
| `docs/testing/testing-external-behavior.md` | Platform tests construct and verify through the page | Pass | Everything is asserted from the DOM — the chip's own text, the rendered card labels, the remaining chips. No internal signal is read. |
| `docs/testing.md` | Mocks describe what production returns | Pass | Both fixtures now answer a named category with the whole category and a keyword-free browse with every connected connector plus a per-category slice, mirroring `discoveryEffectiveConnectors`. |
| `.claude/skills/ccstate/SKILL.md` | Derivation in `computed`; no store writes during render | Pass | `addDialogCategory$` and `categoryCatalog$` are computeds created with the other composer signals; the category is still written only by `updateConnectorUiState$`. |
| `.claude/skills/ccstate/SKILL.md` | `accept` with an explicit status list | Pass | Unchanged in `relatedConnectorCatalog`. |
| `.claude/skills/feature-switch/SKILL.md` | New user-facing behaviour stays gated | Pass | The directory dialog renders only under `connectorDirectory`; the legacy `AddConnectorsDialog` path is untouched. |

### Findings

None blocking.

Non-blocking observations:

- Entering the largest category renders every card in a 600px sheet — the same trade the page made in #33112. Worth watching in acceptance on `Data and Automation` before deciding whether either surface needs pagination.
- Arrow-key navigation now walks the whole category rather than the slice. That is correct, and it means `directoryActiveIndex` can reach a few hundred; nothing in the handler assumes a small list.
- Fixing the fixtures surfaced that three composer-popover cases depended on the old unsliced response. They pass because discovery returns connected connectors ahead of the slice and those fixtures are connected — worth remembering the next time a fixture is made more faithful.

### Testing

Run on this HEAD: `@okouai/app` `check-types` and `lint`, `pnpm knip`, and 129 tests across `connector-directory`, `connectors-catalog`, `connectors-ssh`, `connectors-accounts`, `chat-composer-connectors`, `chat-composer-connector-accounts`, and `chat-composer-connector-loading`. The new case is load-bearing in both directions: without the category fetch it sees twelve cards, and without the chip-row change it loses `Voice50`.

PR #33190, opened 2026-09-10
refactor(ui): migrate badge, pill, and hairline divider to utilities

bingjiez666:
Refs #32402

Three `local-controls` legacy selectors drain in one batch, so `okou-badge`, `okou-pill`, and `okou-border-r` are deleted from `apps/platform/src/views/css/index.css` together with their baseline entries.

## Shared contract

`badgeVariants` in `@okouai/ui` owns the inline badge surface: `border-(length:--border-width-surface) border-solid border-surface-border bg-gray-0`. It reuses the existing page-surface border tokens rather than declaring badge-specific aliases, and it applies to the existing host element through `className`, adding no wrapper. Display, radius, padding, typography, and foreground stay on the consumer, because the 22 real consumers legitimately differ there — inline text, `inline-flex` rows with an icon, and one `<code>` element.

## Consumers

| Legacy class | Sites | Replacement |
| --- | --- | --- |
| `okou-badge` | 21 | `badgeVariants({ className: … })` |
| `okou-pill` | 1 | `badgeVariants({ className: "… text-muted-foreground" })` |
| `okou-border-r` | 1 | `border-r-(length:--border-width-surface) border-r-gray-300` |

`.okou-app .okou-pill` was the only one of the three that also set a foreground; its single consumer (`StatusBadge` in shell style) now spells `text-muted-foreground` itself. That consumer also carried a `border` utility whose 1px width the unlayered selector had overridden with 0.7px; the migrated class list drops it and keeps 0.7px.

`okou-border-r` used Gray 300 rather than the Gray 400 of the other hairlines, so it keeps `border-r-gray-300` instead of `border-surface-border`.

## Verification

Recorded in the PR comments.

PR #33148, opened 2026-09-10
refactor(app): give the product one border-width token

tongx-blip:
## What

Border width was decided in three places at once:

| source | value | count |
|---|---|---|
| Tailwind's built-in default for bare `border` | 1px | everywhere |
| hand-written `border*-[0.7px]` utilities | 0.7px | 48 across 27 files |
| `style={{ border: "0.7px solid …" }}` literals | 0.7px | 18 across 12 files |

There is now one owner. `--default-border-width` in the shared `@theme` is **0.5px**, and Tailwind's bare `border`, `border-t`, `border-x`, `divide-y` and their siblings all read it, so a component asks for "a border" and the system decides how thick it is.

- every `border*-[0.7px]` utility becomes the bare `border*` utility
- every inline `0.7px solid` becomes `var(--border-width-surface) solid`
- `--border-width-surface` derives from `--default-border-width` instead of repeating a literal
- `border-0` and the deliberate emphasis widths (`border-2`, `border-l-2`, `border-[1.5px]`, `border-[24px]`) are untouched — they express a different decision, not a competing value for the same one

## What this looks like in Chromium: nothing

Worth being direct about, because it changes how to read this PR. Sampling the rendered border band at `devicePixelRatio` 1, 2 and 3:

| declared | dpr 1 | dpr 2 | dpr 3 |
|---|---|---|---|
| `1px` | 1 device px | 2 | 3 |
| `0.7px` | 1 | 2 | 3 |
| `0.5px` | 1 | 2 | 3 |
| `0.25px` | 1 | 2 | 3 |

Blink rounds every non-zero border width up to one device pixel. The *used* value stays `1px` and `offsetHeight` is identical, so there is no paint change and no layout change in Chrome or Edge. WebKit can draw the true hairline on a high-density display, and Gecko rounds like Blink.

So in Chromium this lands as **one owner for the decision**, not as a visible change. `docs/styles.md` now says so explicitly, so nobody reaches for a sub-pixel width expecting a lighter border — the border colour is the lever for that.

## Out of scope, deliberately

`.okou-app .okou-composer` and the other first-party CSS borders in `views/css/index.css` keep their literals. Those declarations are frozen atoms in `turbo/style-legacy-baseline.json`, and `docs/styles.md` retires that selector by migrating it to utilities rather than by retuning it — the same boundary that shaped #33022.

[...]

PR #33022, opened 2026-09-09
fix(app): lighten the composer stroke so it stops out-weighing every border

tongx-blip:
## What

The Okou home composer is the widest surface on the page, and it was drawing the heaviest line on it. Both of its stroke states drop one stop on the neutral ramp. Colour only — the border width is untouched.

| | before | after |
|---|---|---|
| rest | `border-gray-400` `#CFCCCB` | `border-gray-300` `#E0DDDC` |
| focus (light) | `--surface-focus` = `gray-500` `#A5A3A3` | `--surface-focus` = `gray-400` `#CFCCCB` |
| focus (dark) | `38.8 38% 38%` amber | unchanged |

For reference, every other card in the app borders on `--border` = `gray-200`. The composer sat two stops heavier than that at rest and three when focused.

Rest and focus stay a full stop apart, so the focused edge is still distinguishable — it just no longer lands near readable-text grey.

## Why dark is untouched

Dark's `--surface-focus` is already an amber-brown: it signals focus by hue, not by darkness, and lifting a neutral against a dark card reads as a white outline rather than as focus. Only the light value moves.

## Scope

`--surface-focus` has exactly one consumer in the repo (`ComposerCard` in `chat-composer.tsx`), so moving the token does not ripple into cards, inputs or dialogs. Both edits sit at the layer `docs/styles.md` names as the source of truth — `:root` / `[data-theme="dark"]` in `packages/ui/src/styles/globals.css`, and a Tailwind utility in the component.

Two things this PR deliberately does **not** do, after the review on `398ad747`:

- It does not retune `.okou-app .okou-composer` in `views/css/index.css`. That legacy selector is frozen by the shrink-only baseline (`docs/styles.md`: *"a changed declaration ... fails lint"*), and its atoms are recorded verbatim in `turbo/style-legacy-baseline.json`. As a result the shared-thread card and the Instatus notice, which borrow that class, stay one stop heavier than the real composer until that selector is migrated to utilities. Worth a follow-up; not worth a ratchet violation here.
- It does not change the border width. `docs/styles.md` states the chat composer *"uses the default `border` width for its surface and connector circles"*, and an earlier `border-[0.7px]` attempt was reverted. Measured at `devicePixelRatio: 2`, `0.7px` and `1px` rasterise identically to two device pixels, so the width bought nothing.

## Verification

Compiled the patched `globals.css` with the repo's pinned Tailwind (`4.3.3`) and read computed values off the real merged class list:

[...]

PR #33269, opened 2026-09-10
refactor(app): give the product one border-width token

tongx-blip:
Reopens the work from #33148, which was closed twice without a comment (by @e7h4n, most recently 2026-09-10 08:14Z while it sat at position 1 in the merge queue with every check green) and whose branch was then deleted. @e7h4n closed @bingjiez666's #33190 — also a styling refactor — 67 seconds earlier, so this may be a deliberate batch decision I do not have the context for. **If styling refactors are frozen right now, say so and I'll close this again.** Rebuilt on current `main` rather than restored, so the diff is one clean commit.

## What

Border width is decided in three places at once:

| source | value | count |
|---|---|---|
| Tailwind's built-in default for bare `border` | 1px | everywhere |
| hand-written `border*-[0.7px]` utilities | 0.7px | 49 across 27 files |
| `style={{ border: "0.7px solid …" }}` literals | 0.7px | 18 across 12 files |

There is now one owner. `--default-border-width` in the shared `@theme` is **0.5px**, and Tailwind's bare `border`, `border-t`, `border-x`, `divide-y` and their siblings all read it, so a component asks for "a border" and the system decides how thick it is.

- every `border*-[0.7px]` utility becomes the bare `border*` utility
- every inline `0.7px solid` becomes `var(--border-width-surface) solid`
- `--border-width-surface` derives from `--default-border-width` instead of repeating a literal
- `border-0` and the deliberate emphasis widths (`border-2`, `border-l-2`, `border-[1.5px]`, `border-[24px]`) are untouched — they express a different decision, not a competing value for the same one

## What this looks like in Chromium: nothing

Worth being direct about, because it changes how to read this PR. Sampling the rendered border band at `devicePixelRatio` 1, 2 and 3:

| declared | dpr 1 | dpr 2 | dpr 3 |
|---|---|---|---|
| `1px` | 1 device px | 2 | 3 |
| `0.7px` | 1 | 2 | 3 |
| `0.5px` | 1 | 2 | 3 |
| `0.25px` | 1 | 2 | 3 |

Blink rounds every non-zero border width up to one device pixel. The *used* value stays `1px` and `offsetHeight` is identical, so there is no paint change and no layout change in Chrome or Edge. WebKit can draw the true hairline on a high-density display; Gecko rounds like Blink.

So this lands as **one owner for the decision**, not as a visible change. `docs/styles.md` now says so explicitly, so nobody reaches for a sub-pixel width expecting a lighter border — the border colour is the lever for that.

## Out of scope, deliberately

`.okou-app .okou-composer` and the other first-party CSS borders in `views/css/index.css` keep their literals. Those declarations are frozen atoms in `turbo/style-legacy-baseline.json`, and `docs/styles.md` retires that selector by migrating it to utilities rather than by retuning it — the same boundary that shaped #33022.

The 18 inline borders now reference the shared token but remain `style` props. Converting them to `surfaceVariants` is the real cleanup and is a larger, conditional-logic refactor across `org-manage`.

## Tests

The auth checkbox style assertion compiled its classes against a stub `@theme` that omitted this key, so it pinned Tailwind's built-in 1px rather than the product's hairline — a value the App never rendered for that checkbox, since it already used the bare `border` utility. The stub now carries the key and the assertion states `0.5px`. No new tests: this is a token value plus a mechanical class substitution, and `docs/styles.md` forbids locating surfaces through utility class names in tests.

## Verification

- `node scripts/style-policy.mjs` → **Style policy passed** (the shrink-only ratchet; no first-party selector declaration is touched, and `@theme` in `globals.css` is not a baselined atom)
- `node scripts/style-migration.mjs` → **Style migration manifest passed**
- `prettier --check .` clean from `turbo`, and on `docs/styles.md`
- Compiled the real App entry with the repo's pinned Tailwind `4.3.3`: `.border` and `.border-t` emit `0.5px`, `.border-0` stays `0px`, `.border-2` stays `2px`, and both `--default-border-width` and `--border-width-surface` are emitted to `:root` so the inline `var(--border-width-surface)` resolves
- Rendered A/B against current `main` — input, popover, composer card, outline button, `divide-y` list, `border-2` emphasis and the inline-token border — is **pixel-identical at dpr 1 and dpr 2**
- Swept for collateral: one `border border-[0.7px]` collapsed to a redundant `border border` and was reduced to a single `border`

### DIFF
--- a/docs/styles.md
+++ b/docs/styles.md
@@ -33,6 +33,10 @@ Token names describe meaning rather than a page or component. A reusable interac
 
 Components must not introduce local CSS variables as an alternate token registry. A runtime value that is genuinely computed by the component may use a narrowly named custom property as data, while its visual semantics still come from Tailwind utilities and registered tokens.
 
+One hairline serves the whole product. `--default-border-width` in the shared `@theme` is 0.5px, and Tailwind's bare `border`, `border-t`, `border-x`, `divide-y`, and their siblings all read it, so a component asks for "a border" and the system decides how thick it is. Components must not hand-write a width: an arbitrary width such as `border-[0.7px]`, or a literal width inside a `style` prop, is a second registry for a decision this token already owns. `border-0` and the deliberate emphasis widths such as `border-2` stay available, because they express a different decision rather than a competing value for the same one.
+
+The sub-pixel value is a declaration of intent as much as a measurement. Blink and Gecko round a non-zero border up to one device pixel, so it renders exactly like 1px there and layout is unchanged in every engine; WebKit can draw the true hairline on a high-density display. Do not treat a width below 1px as a way to make a border visibly lighter in Chromium — reach for the border color for that.
+
 Color-theme presets in the App stylesheet share their anchor and companion colors between picker swatches and workspace ambience. Daydream uses cool blue and violet, while Cotton sky uses pastel pink and blue. Each preset's hue and ring values keep semantic surfaces, selected states, and focus indicators aligned with that palette in Light/Dark.
 
 ## Token and variant governance
@@ -80,7 +84,7 @@ remain available without hovering.
 | Decision        | Shared token / utility                                    | Theme contract                                                                                       |
 | --------------- | --------------------------------------------------------- | ---------------------------------------------------------------------------------------------------- |
 | Fill            | `bg-card`                                                 | Existing semantic card fill in each theme                                                            |
-| Border          | `--color-surface-border`, `--border-width-surface`        | Gray 400 at 0.7 CSS pixels; the browser rounds for its device scale                                  |
+| Border          | `--color-surface-border`, `--border-width-surface`        | Gray 400 at the shared `--default-border-width` hairline; the browser rounds for its device scale    |
 | Radius          | `rounded-surface`, `rounded-surface-compact`              | 1.25rem and a fixed 12px respectively in every theme                                                 |
 | Elevation       | `shadow-surface` via `--surface-shadow`                   | Neutral lift in Light/Dark; the gradient theme uses the canonical state-layer hue with reduced alpha |
 | Pointer overlay | `bg-state-hover-overlay`                                  | The shared interaction-state overlay painted above the opaque card fill                              |
--- a/turbo/apps/platform/src/views/agents-page/agents-page-tabs.tsx
+++ b/turbo/apps/platform/src/views/agents-page/agents-page-tabs.tsx
@@ -743,7 +743,7 @@ function AgentCard({ agent, creator, hasUnread, showCreator }: AgentProps) {
                   <TooltipContent
                     side="bottom"
                     align="start"
-                    className="w-64 rounded-lg border-[0.7px] border-[hsl(var(--gray-400))] p-3 text-left font-normal"
+                    className="w-64 rounded-lg border border-[hsl(var(--gray-400))] p-3 text-left font-normal"
                     style={{
                       backgroundColor: "hsl(var(--popover))",
                       color: "hsl(var(--popover-foreground))",
--- a/turbo/apps/platform/src/views/auth-v1/component-appearance.ts
+++ b/turbo/apps/platform/src/views/auth-v1/component-appearance.ts
@@ -13,7 +13,7 @@ const AUTH_V1_PRIMARY_ACTION_CLASS =
 // the shared Input boundary while adapting focus and error states through the
 // public attributes that Clerk exposes on each slot.
 const AUTH_V1_OTP_INPUT_CLASS =
-  "border-[0.7px] border-[hsl(var(--gray-400))] bg-input shadow-none data-[focus-within=true]:border-primary data-[focus-within=true]:ring-[3px] data-[focus-within=true]:ring-primary/10 aria-invalid:border-destructive data-[focus-within=true]:aria-invalid:border-destructive";
+  "border border-[hsl(var(--gray-400))] bg-input shadow-none data-[focus-within=true]:border-primary data-[focus-within=true]:ring-[3px] data-[focus-within=true]:ring-primary/10 aria-invalid:border-destructive data-[focus-within=true]:aria-invalid:border-destructive";
 
 /** Keep branding and the page shell; Clerk owns control styles and states. */
 export function getAuthV1ComponentAppearance(
--- a/turbo/apps/platform/src/views/auth-v2/__tests__/auth-v2-style-assertions.ts
+++ b/turbo/apps/platform/src/views/auth-v2/__tests__/auth-v2-style-assertions.ts
@@ -19,6 +19,7 @@ const checkboxTheme = `
   @theme {
     --spacing: 0.25rem;
     --radius-md: 0.375rem;
+    --default-border-width: 0.5px;
     --color-border: rgb(10 20 30);
     --color-input: rgb(40 50 60);
     --color-primary: rgb(70 80 90);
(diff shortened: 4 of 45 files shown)


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: 15f7ec664e2f141855e3cf44e1d328476c6f8dbd
PR_AUTO_REVIEW_VERDICT: LGTM

Reviewed HEAD: `15f7ec664e2f141855e3cf44e1d328476c6f8dbd`
Practice revision: `43bd7ed8bfca1721590ffa85bf930cf9988f9f79`

### Summary

Collapses three competing sources for one decision — Tailwind's built-in 1px, 49 hand-written `border*-[0.7px]` utilities across 27 files, and 18 `style={{ border: "0.7px solid …" }}` literals across 12 — into a single `--default-border-width` of 0.5px in the shared `@theme`. `--border-width-surface` derives from it, the arbitrary utilities become bare `border*`, the inline literals reference the token, and `docs/styles.md` records the contract plus the rendering caveat.

Rebuilt on current `main` as one commit rather than restored from #33148, whose branch was deleted. The rendered A/B against `main` — input, popover, composer card, outline button, `divide-y` list, `border-2` emphasis and the inline-token border — is **pixel-identical at dpr 1 and dpr 2**. The style policy, migration manifest and Prettier all pass locally.

Reviewer note, outside the code: #33148 was closed twice with no comment or review, the second time at queue position 1 with every check green, and #33190 (also a styling refactor, different author) was closed 67 seconds before it. The PR body asks directly whether styling refactors are frozen. That is a process question for a human, not a defect in this diff, so it does not change the verdict.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | Token consumers are Tailwind's `border*`/`divide-*` utilities (one lookup in the compiled `tailwindcss` lib) and `--border-width-surface` → `surfaceVariants`. Compiled the real App entry with pinned Tailwind `4.3.3`: `.border`/`.border-t` emit `0.5px`, `.border-0` stays `0px`, `.border-2` stays `2px`. |
| `REVIEW.md` | Distinguish introduced defects from unchanged code and unsupported possibilities | Pass | The sweep is mechanical and repo-wide: zero `border*-[0.7px]`, zero inline `0.7px solid`, and no other `0.7px` remain in `apps/platform/src` or `packages/ui/src`. The only hand-written edits are the `@theme` block, the guide, the auth stub, and one duplicate-class fix. |
| `REVIEW.md` | Require meaningful coverage for new behavior; identify the missing scenario | Pass | No behavior change. For a width token the meaningful check is the rendered comparison, and it is zero-diff at dpr 1 and 2. |
| `REVIEW.md` | A rewritten expectation alone does not prove an intentional, correctly scoped replacement | Pass | The two auth assertions are argued from the stub theme's contents, not from the new value. See Testing. |
| `REVIEW.md` | Persisted data, API contracts, deployment compatibility | N/A | No schema, request, queue payload, persisted state or service-worker surface. A stylesheet ships inside the app bundle and has no cross-version pairing. |
| `REVIEW.md` | Every new fallback declared in a `Fallbacks` PR-body section | N/A | No fallback added or removed; no branch, default or tolerance path in the diff. |
| `docs/styles.md` | Business styling composes Tailwind utilities; first-party CSS selectors only shrink | Pass | `node scripts/style-policy.mjs` → **Style policy passed**. No first-party selector declaration is touched; `views/css/index.css` is untouched; `@theme` in `globals.css` is not a baselined atom (its atoms are only `@layer components` classes). |
| `docs/styles.md` | Components must not introduce a parallel token registry | Pass | This is the finding the PR fixes: 49 utilities and 18 inline literals stop declaring their own width, and `--border-width-surface` stops repeating a literal. |
| `docs/styles.md` | New tokens represent a reusable semantic decision with a documented consumer contract | Pass | `--default-border-width` is Tailwind's own theme key rather than a new alias, its consumers are the framework's border utilities, and the contract is now written in the guide. Light/dark mapping is N/A for a length. |
| `docs/styles.md` | A change to ownership, naming, or theme mapping must update this guide in the same PR | Pass | Updated here: the surface-border row drops "0.7 CSS pixels", and a new paragraph states the one-hairline rule, its exceptions (`border-0`, `border-2`) and the engine-rounding caveat. The upstream "Color-theme presets" paragraph added since the last review is present and untouched. |
| `docs/styles.md` | The chat composer uses the default `border` width for its surface and connector circles | Pass | Still true, and now automatic — the composer declares no width, so it follows the token. |
| `docs/styles.md` | Keep the border width constant across interaction states | Pass | No interaction-state width is introduced; only resting widths were replaced. |
| `docs/styles.md` | `surfaceVariants` uses `border-(length:--border-width-surface)` so merging recognizes width independently of color | Pass | Unchanged. Both `--default-border-width` and `--border-width-surface` are emitted to `:root`, and the A/B render proves the inline `var(--border-width-surface)` consumers still paint a border — an unresolved `var()` would invalidate the shorthand and appear as a diff. |
| `docs/clerk-customize.md` | Adapt only public Clerk slots, with Tailwind classes and existing semantic tokens | Pass | `component-appearance.ts` swaps one width utility for another on the same existing slot. No new slot, no `.cl-*` targeting, no inline style, no `!important`, no `:has()`. `e2e/playwright/tests/auth-v1.spec.ts` asserts `border-color` only, not width. |
| `docs/bad-smell.md` | No lint/type suppressions, weakened rules, or baseline/allowlist expansion | Pass | `style-legacy-baseline.json` and `style-allowlist.json` are untouched; no suppression comments in the diff. |
| `docs/bad-smell.md` | Document changed public contracts and assess their consumers | Pass | The contract is a shared `@theme` key; the guide and the PR body both state the old value, the new value and the engine behavior. |
| `docs/testing.md` | Add tests for new behavior; avoid cases restating static configuration or pinning incidental implementation | Pass | See Testing. |
| `.claude/skills/feature-switch/SKILL.md` | New user-facing features are gated; restyles and fixes to existing features are not | Pass | No new page, section, sidebar item, endpoint, integration or token capability. |

### Findings

None.

### Non-blocking observations

- **The width is a Chromium no-op, and the PR says so.** Sampling the rendered border band at dpr 1/2/3, Blink rounds every non-zero width below 1px up to one device pixel and keeps `1px` as the used value, so there is no paint or layout change in Chrome. Read this as consolidating ownership of the decision. The guide paragraph added here explicitly warns against reaching for a sub-pixel width to make a border look lighter, which is the trap that would otherwise be re-entered.
- **18 inline borders now reference a token but remain `style` props.** The guide's final state removes component-owned inline styles entirely; several are `org-manage` cards that should become `surfaceVariants`. Pointing them at the shared token is strictly better than a literal and does not entrench them, but the real cleanup is still owed.
- **`views/css/index.css` keeps its `0.7px` literals** (`.okou-app .okou-composer` and neighbours). Correct call: those are frozen atoms in `style-legacy-baseline.json`, and editing them is what crossed the boundary in #33022. It does mean the product briefly holds two hairline values, with the legacy one rendering identically in Chromium.

### Testing

The only test change is the auth checkbox style assertion, and it is a fidelity increase rather than a rewritten expectation chasing a new value. `renderedCheckboxPresentation` compiles the component's real class list against a local stub `@theme` in `auth-v2-style-assertions.ts` that declares spacing, radius and four colors. Because the stub omitted the border-width key, the assertion was reading Tailwind's built-in 1px — a value the App has never rendered for that checkbox and would not have rendered before this PR either, since it already used the bare `border` utility. Adding the key to the stub makes the assertion describe the product; the expectation moves to `0.5px` as a consequence.

No new tests. This is a token value plus a mechanical class substitution, `docs/testing.md` warns against cases that "restate static configuration or pin incidental implementation", and `docs/styles.md` forbids locating surfaces through utility class names in tests.

Checks run locally: `scripts/style-policy.mjs`, `scripts/style-migration.mjs`, and `prettier --check` from `turbo` plus on `docs/styles.md` all pass. `lint-style`, `lint-format`, `lint-clerk-customize`, `test-app` and the rest remain the authoritative gates. No material gap.

PR #33292, opened 2026-09-10
fix(api): describe every category while browsing one of them

tongx-blip:
Opening a category and then reopening the filter offered **only the category already open** — so the filter became a dead end and the breadcrumb was the only way back to the other eleven.

## Two fields disagreed about what a category list is for

Discovery answers a named category with that category alone (#33112), and the response's `categoryMetadata` was derived from **the connectors it returned**, so it collapsed with them. Verified against production:

| `GET /api/connector-catalog/discovery?category=ai-general-models` | |
| --- | --- |
| `connectors` | 71, all `ai-general-models` ✅ |
| `categoryConnectorCounts` | **12 keys** ✅ — always described the whole catalog |
| `categoryMetadata.categories` | **1** ❌ |

`categoryMetadataForConnectors(catalog, args.effective)` received the already-scoped list. The counts never worked that way, which is why the page could still show `71` next to a category it could no longer offer.

## The fix

The category list now comes from the same set as the counts — the whole catalog minus the connectors Okou runs for itself — so it never offers a category the counts do not know. `list` and `status` are unaffected: their effective set already is the whole catalog.

The connectors page builds its filter from that metadata instead of grouping whatever came back, which is what made the page's own list follow the response.

## Verification

- `@okouai/app` and `@okouai/api` `check-types` and `lint`, `pnpm knip` — clean
- 61 tests across `connectors-catalog`, `connectors-ssh`, `connector-directory`, `connectors-accounts`
- New page case "Keep every category in the filter while one of them is open" — loads `/connectors?category=…`, opens the filter, asserts the other category is still listed. **Verified load-bearing**: reverting the page to `sectionsOf(allConnectors)` fails it.
- New API assertion in "answers a named category with the whole category": the scoped response's `categoryMetadata` ids equal the `categoryConnectorCounts` keys. The DB-backed API suite needs Postgres, which this sandbox has no server for; it runs in CI.

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/connector-catalog.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/connector-catalog.test.ts
@@ -616,6 +616,17 @@ describe("GET /api/connector-catalog", () => {
       expect(connector.category).toBe(category);
     }
 
+    // The category list describes the catalog, not the response. A client
+    // offers the other categories from it, so collapsing it to the one being
+    // browsed would leave no way to reach any of them.
+    expect(
+      (scoped.body.categoryMetadata?.categories ?? [])
+        .map((entry) => {
+          return entry.id;
+        })
+        .sort(),
+    ).toStrictEqual(Object.keys(counts).sort());
+
     // Still ranked, and still without the connectors Okou runs for itself.
     const ranks = scoped.body.connectors.map((connector) => {
       return connector.popularityRank ?? Number.MAX_SAFE_INTEGER;
--- a/turbo/apps/api/src/signals/services/connector-catalog-external-reader.service.ts
+++ b/turbo/apps/api/src/signals/services/connector-catalog-external-reader.service.ts
@@ -1241,6 +1241,12 @@ export async function discoverExternalPublicConnectorCatalogStatus(
   const read = connectorCatalogStatusRead({
     catalog,
     effective: discoveryEffectiveConnectors(effective, args),
+    // The category list and the category counts describe the same thing, so
+    // they are computed from the same set: the whole catalog minus the
+    // connectors Okou runs for itself, not the slice that came back for a
+    // named category. Offering a category the counts do not know would be a
+    // chip that opens nothing.
+    categorySource: withoutInternalConnectors(effective),
     connections: args.connections,
     referenceConnectorSlugs: args.referenceConnectorSlugs,
   });
@@ -1257,6 +1263,13 @@ export async function discoverExternalPublicConnectorCatalogStatus(
 function connectorCatalogStatusRead(args: {
   readonly catalog: AcceptedConnectorCatalogSnapshot;
   readonly effective: readonly EffectiveConnector[];
+  /**
+   * The connectors the category list describes, when that is wider than the
+   * ones being returned. Discovery answers a named category with only that
+   * category, and the category list is how a client offers the others, so it
+   * has to keep describing the whole catalog. Defaults to what is returned.
+   */
+  readonly categorySource?: readonly EffectiveConnector[];
   readonly connections: readonly ConnectorCatalogConnection[];
   readonly referenceConnectorSlugs: readonly string[];
 }): ConnectorCatalogStatusRead {
@@ -1279,7 +1292,7 @@ function connectorCatalogStatusRead(args: {
       connectors,
       categoryMetadata: categoryMetadataForConnectors(
         args.catalog,
-        args.effective,
+        args.categorySource ?? args.effective,
       ),
     },
     referenceMetadata: referenceMetadataForCatalog(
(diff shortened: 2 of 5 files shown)


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: 18a300d6df6542057b25ddf10a48a5bf5b0ea764
PR_AUTO_REVIEW_VERDICT: LGTM

Reviewed HEAD: `18a300d6df6542057b25ddf10a48a5bf5b0ea764` · practice revision `68fd2e2bd3d3c99ef0f09831a4475b0ccf3abdf1`

### Summary

Inside a category, reopening the filter offered only that category. Discovery's `categoryMetadata` was derived from the connectors it returned, so a category-scoped response (#33112) collapsed it, while `categoryConnectorCounts` had always described the whole catalog. The two fields now come from the same set — the catalog minus internal connectors — and the connectors page builds its filter from that list instead of grouping whatever came back.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | `connectorCatalogStatusRead` gains an optional `categorySource`; only `discoverExternalPublicConnectorCatalogStatus` passes it. `list` and `status` keep `categorySource ?? effective`, and their effective set already is the whole catalog, so their responses are unchanged. |
| `REVIEW.md` | Check the failure the change claims to fix | Pass | Reproduced against production before the change: `?category=ai-general-models` returned 71 connectors, `categoryConnectorCounts` with 12 keys, and `categoryMetadata.categories` with 1. The new API assertion ties those two fields together. |
| `REVIEW.md` | Regression coverage for the fixed defect | Pass | "Keep every category in the filter while one of them is open" loads `/connectors?category=…`, opens the filter and asserts the other category is offered. **Proven load-bearing**: restoring `sectionsOf(allConnectors)` fails it. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | None added. The `ccstate/no-package-variable` hit from a module-scope fixture object was fixed by making it a function, not by suppressing. |
| `docs/bad-smell.md` | §4 interface changes assessed against consumers | Pass | `categorySource` is an internal optional argument with one caller; `buildConnectorsBrowseModel` drops `allConnectors`, which had no other reader after the change. |
| `docs/bad-smell.md` | §7 no fallback for a state the contract rules out | Pass | The second commit removes the "group the returned connectors when metadata is absent" branch rather than keeping it. |
| `docs/fallback.md` | §2 a non-GA switch needs no compatibility code | Pass | That removed branch was exactly this: handling a response the API does not send, on a surface behind `connectorDirectory` (`enabled: false` + staff hashes). |
| `docs/fallback.md` | §9 declare any introduced fallback | N/A | None introduced; one was removed. |
| `docs/fallback.md` | §10 removing a fallback carries evidence | Pass | The removal is justified by the contract's own producer: discovery always sends `categoryMetadata`, verified live, and the two fixtures that relied on its absence were corrected to describe the real response. |
| `docs/deployment-compatibility.md` | Cover the old/new combinations a rollout produces | Pass | No contract change: `categoryMetadata` stays optional and only its contents widen. A new app against an older API sees an empty filter list on a switch-gated surface; an older app against the new API sees a longer category list, which is what it already renders for the unscoped browse. |
| `docs/testing.md` | Prefer integration coverage through real entry points | Pass | The page case boots `/connectors` with `setupPage` and drives the real filter menu; the API assertion goes through `client.discovery`. |
| `docs/testing/testing-external-behavior.md` | Construct and verify through the page or endpoint | Pass | The page case reads menu item text from the DOM; the API case compares two fields of the same response rather than reaching into the service. |
| `docs/testing.md` | Mocks describe what production returns | Pass | Both directory fixtures now carry the category metadata discovery always sends, which is what exposed the breadcrumb reading the same list. |
| `.claude/skills/ccstate/SKILL.md` | No mutable package-scope state | Pass | The shared metadata fixture is a function, matching `shelfCatalog()` beside it. |
| `.claude/skills/feature-switch/SKILL.md` | Changed behaviour stays gated | Pass | `categoryFilterSections` feeds only the shelf-path browse model; the switch-off page keeps its own grouping and dropdown. |
| `docs/styles.md` | Tailwind utilities and registered tokens; one hairline token | N/A | No styling in this diff; no border width is written. |

### Findings

None blocking.

Non-blocking observations:

- The filter now lists every category the catalog has, including ones whose connectors are all connected already. That is the intent — the list is the catalog's shape, not the response's — but it means a category can be offered whose page shows only cards you already have. Fine today; worth remembering if "hide what I have" ever becomes a request.
- `withoutInternalConnectors` is now applied twice on a discovery request: once for the browse slice and once for the category source. Both are linear over the effective set; not worth caching unless discovery shows up in a profile.

### Testing

87 tests across `connectors-catalog`, `connectors-ssh`, `connector-directory`, `connectors-accounts`, `connectors-custom`, plus `@okouai/app` and `@okouai/api` `check-types` and `lint` and `pnpm knip`. The DB-backed API suite needs Postgres, which this sandbox has no server for; it runs in CI. Removing the metadata fallback surfaced that the breadcrumb reads the same category list — caught by an existing case rather than by inspection, which is the point of correcting the fixtures.

PR #33188, opened 2026-09-10
feat(chat): move selected task chips into the composer

bingjiez666:
## Summary

Start new chats with left-aligned task chips and no preselected recommendations. Selecting Workflow, Presentation, Image, Video, or Website replaces the chips with the matching content and places a removable task tag inside the existing composer. Removing the tag restores the choices while preserving the draft and uploaded files.

Show four vertical suggestions for Workflow, Image, Video, and Website, including the last image-suggestion page. Place More ideas beside the first row on desktop and below the list on narrow screens. Reuse the existing `composerTaskChips` feature switch, create-mode submission, presentation templates, and localized task labels.

App preview: https://pr-33188-app-okou-app-preview.vm0.workers.dev/?x-vercel-protection-bypass=jAdTONWK3jqOzLaGJYjS4kINOvXBRmvB&x-vercel-set-bypass-cookie=true

Approved interaction reference: https://dpl-dcecd81f-028f-4d4a-9f61-b140de6f2dd0.okou.app

Browser screenshots: [Initial left-aligned chips](https://a.okou.io/gj2tqeqizn.png) · [Selected Image with four vertical ideas](https://a.okou.io/b7knxa67ew.png)

## Verification

- Focused page tests cover all five selections, removal, empty-editor Backspace, IME input, draft and attachment preservation, suggestion rotation, existing generation submissions, templates, and localization.
- All 51 targeted tests in four composer page-test files passed on the final implementation.
- Platform production/test/build TypeScript checks, affected ESLint and Oxlint (including type-aware checks), Prettier, Tailwind lint, style policy, style migration policy, i18n validation, and Platform Knip passed locally.
- Browser PASS on the deployed PR head `bab2460a81f0f8fd37090c0d700c57f5af6ba427`: left-aligned initial chips with no suggestions; all five task tags inside the composer; four vertical Workflow/Image/Video/Website suggestions; every image suggestion page; desktop/mobile More ideas placement; tag removal and empty-editor Backspace; draft text/focus preservation; presentation gallery and slide-count control; website template library. Checked at 1440×960 and 390×844 with no horizontal overflow.
- Browser setup: fresh Clerk test account, onboarding completed via the preview API because onboarding is outside scope, `composerTaskChips` enabled through Lab, `composerCreateCommands` and `_realAgentInPreview` disabled. No chat runs were submitted.
[...]

PR #33293, opened 2026-09-10
feat(app): show the template cover on the inline template chip behind a switch

tongx-blip:
## What

Adds `composerTemplateChipCover` (Lab → Beta, staff orgs). With it on, the
composer's inline template chip leads with the chosen template's cover image
instead of a generic glyph. With it off, nothing changes.

Design: https://okou-create-entry-prs.okou.app — before/after at real geometry,
plus a 2.4× zoom of the chip.

## Why

`createInlineTemplateNodeView` already receives `previewImageUrl` on the node —
`selectedComposerTemplateAttachment` resolves it for presentation, illustration,
avatar and explainer templates — but the view never rendered it. You pick a
template from a wall of covers and get back an orange word, so two decks are
indistinguishable in the draft without reopening the picker. The older
block-style chip (`createTemplateAttachmentNodeView`) already draws that cover.

## How

- New key + registry entry, `enabled: false` + `STAFF_ORG_ID_HASHES`.
- The node view renders the cover in an 18px rounded slot when the switch is on
  and a cover exists; otherwise the existing SwatchBook glyph. 18px inside the
  28px inline chip keeps the cover at the proportion the block chip uses (20px
  inside 32px). Chip height is unchanged.
- The slot is resolved inside `render()`, because the picker rewrites the node
  in place via `setNodeMarkup` when it replaces a selection.
- `WorkflowComposerRuntime.templateChipCover` is a **reader function**, not a
  latched boolean. Latching it in the mount command reproducibly rendered the
  glyph with the switch on: the authoritative feature-switch read resolves after
  the composer mounts, so the mount-time value is always the pre-hydration
  default. The switch-off test below is what caught this.

video / website / workflow templates have no catalog cover, so those keep the
glyph — both forms coexist by design.

## History

Supersedes #33184, which carried the same change and was closed. This branch is
merged up to current `main`; #33142 and #33188 reworked `ComposerCreateControls`
and the task chips but did not touch `createInlineTemplateNodeView`, so there is
no overlap with that work.

## Verification

- `vitest run src/views/okou-page/__tests__/chat-composer-create.test.tsx` — 19 passed
- `vitest run src` in `packages/core` — 243 passed
- `pnpm run check-types`, full `pnpm run lint` (platform), `prettier --check`

Three tests: switch off keeps the glyph for a presentation template; switch on
shows the cover and follows a replacement selection; switch on with a video
template (no catalog cover) keeps the glyph.


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: f12572f9081b500c210e9915a11ae75b18d0d266
PR_AUTO_REVIEW_VERDICT: LGTM

Reviewed HEAD: `f12572f9081b500c210e9915a11ae75b18d0d266`
Practice revision: `68fd2e2bd3d3c99ef0f09831a4475b0ccf3abdf1` (`main`)

### Summary

Adds the staff-only `composerTemplateChipCover` switch and, when it is on, renders the selected template's `previewImageUrl` in an 18px slot at the head of the composer's inline template chip; the existing SwatchBook glyph stands in when the switch is off or the template has no catalog cover. The switch is read through a reader function on `WorkflowComposerRuntime` so each chip render sees the post-hydration value.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | `createInlineTemplateNodeView` is reached only from `createInlineTemplateNode`'s `addNodeView`; both call sites (`render()` on first mount and on `update()`) go through the same `render(nextNode)`, so an in-place `setNodeMarkup` replacement re-resolves the cover. |
| `REVIEW.md` | Meaningful coverage for new behavior; regression coverage for the affected path | Pass | Three page tests in `chat-composer-create.test.tsx`: switch off keeps the glyph, switch on shows the cover and follows a replacement, switch on with a coverless video template keeps the glyph. The switch-off case is the regression guard for the pre-existing rendering. |
| `REVIEW.md` | Correct feature/capability containment for a changed flow | Pass | `FeatureSwitchKey.ComposerTemplateChipCover`, `enabled: false` + `STAFF_ORG_ID_HASHES`; the surrounding inline-chip surface is itself gated by the non-GA `ComposerCreateCommands`. |
| `REVIEW.md` | Persisted data / API contract review | N/A | No schema, contract, request, queue payload, or persisted-state change; `previewImageUrl` is an existing ProseMirror node attribute already validated by `templateAttachmentNodeAttributes`. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | No `any`, no `as`, no suppression comments in the diff; `chip.querySelector("img")` uses TypeScript's tag-name overload rather than an assertion. |
| `docs/bad-smell.md` | No fabricated defaults hiding a broken invariant | Pass | `templateAttachmentNodeAttributes` still throws on an invalid node; the diff adds no `??` chain. In particular `get(featureSwitch$)[…]` is written without `?? false` because `featureSwitchState$` is `Record<FeatureSwitchKey, boolean>` — total, so a default would be section-3 slop. |
| `docs/bad-smell.md` | Static imports in production code | Pass | `FeatureSwitchKey` and `featureSwitch$` are added as static imports. |
| `docs/fallback.md` §2 | A non-GA feature switch needs no compatibility code | Pass | No dual-read, tolerance branch, or migration was added for the cutover, and none is required. |
| `docs/fallback.md` §3/§6 | No default for a state the owning contract rules out; presentational defaults allowed | Pass | The glyph branch handles `previewImageUrl` genuinely absent — the node attribute is `string \| null` and `templateAttachmentNodeAttributes` omits the field when null; video/website/workflow templates carry no catalog cover, proven by the third test. This is §6.4 (placeholder for an unavailable preview). |
| `docs/fallback.md` §9 | Every introduced fallback declared in a `Fallbacks` PR-body section | Pass | The diff introduces no new fallback, so no section is required. The two `templateChipCover: () => false` assignments are the runtime's unmounted initializer and its teardown reset, matching the sibling `removeTemplate = () => {}` members — an initial value for a state the system really produces, not a branch for an impossible one. |
| `docs/fallback.md` §1 | No negative tests pinning removed behavior | Pass | All three added tests assert live rendering; none asserts that something stays deleted. |
| `docs/testing.md` | Test the contract an external caller relies on, through the production entry point | Pass | Every case boots the real Router via `setupPage`, drives the slash command, the **Add template** button and the picker dialog, and asserts on rendered page content. |
| `docs/testing/testing-external-behavior.md` | No assertions on store, component state, CSS classes, or internal callbacks | Pass | Assertions are on the chip's text content and on the presence and `src` of a rendered ``. The cover is intentionally `alt=""` (decorative; the button owns the `aria-label`), so a scoped `querySelector("img")` inside the located chip is the only external handle on it. |
| `docs/testing/app-testing.md` | Mocks installed before `setupPage`; page-level integration test | Pass | `setupModels()` and `mockChatLifecycle(context)` precede `setupComposerWithChipCover()`; the file's existing `testContext()` and helpers are reused. |
| `docs/testing/app-testing.md` | Avoid `findByRole` for `button`; `waitFor` holds only queries/assertions; a negative assertion follows a causal completion point | Pass | Buttons go through the file's `button()` helper; `findByRole("dialog")` and `findByLabelText` are allowed roles. Each `waitFor` body contains only expectations, and both `expect(inlineTemplateCover()).toBeNull()` calls follow an awaited chip-appearance sentinel — the cover is written synchronously inside the same `render()`, so there is no gap after it. |
| `.claude/skills/feature-switch/SKILL.md` | Enum key + registry entry with maintainer and rollout audience | Pass | `ComposerTemplateChipCover = "composerTemplateChipCover"`; registry entry has `maintainer: "tongx@okou.ai"`, a description, `enabled: false`, `enabledOrgIdHashes: STAFF_ORG_ID_HASHES` → Lab "Beta" with no extra wiring, since `LabPage` iterates the whole registry. |
| `.claude/skills/ccstate/SKILL.md` | Derivation in `computed`, semantic actions in commands; no store mutation during render | Pass | Nothing is written to the store. The consumer is a ProseMirror `NodeView`, imperative DOM outside React's render, so it cannot subscribe to a `computed`; the value is pulled at paint time instead. |
| `docs/effect.md` | Choose the execution boundary: `onRef` owns a DOM-lifetime resource, acquired and released in the same lifecycle | Pass | The reader is installed inside the `onRef` mount command and released by the same command's `abort` listener via `resetMountedWorkflowRuntime`. |
| `docs/cache.md` RC-8 | Symmetric resource lifecycle | Pass | `createMountEditorCommand`'s abort handler (line ~2119) calls `resetMountedWorkflowRuntime(runtime)`, which the diff extends to reset `templateChipCover`, so the escaped `get` is dropped on unmount alongside the editor. |
| `docs/cache.md` RC-9 | Closures that let ccstate `get`/`set` escape a command callback | Pass, with a note | The reader does let `get` escape the mount command (see Observations). Its lifetime is bounded by the same abort path, and the file already establishes this shape — `runtime.update` closes over `set` and is invoked from editor callbacks. |
| `docs/cache.md` RC-3/RC-4 | No unbounded lifetime cache; state owned by the narrowest lifecycle | Pass | The reader lives on the per-composer `WorkflowComposerRuntime`, not a module registry, and is reset on unmount. |
| `docs/styles.md` | Tailwind utilities only; no first-party selectors, injected styles, or hand-written border widths | Pass | Both new class strings are plain Tailwind utilities on `className`; no CSS file, `<style>`, or `border-[…]` width. `lint-style` is green on this HEAD. |

### Findings

None.

1. **`get` escapes the mount command.** `runtime.templateChipCover = () => get(featureSwitch$)[…]` is exactly the shape `docs/cache.md` RC-9 flags as a lint blind spot. I did not raise it as a finding: the escape is bounded by the same `onRef` abort path that unmounts the editor, `runtime.update` already closes over `set` in the same command, and the alternative — pushing a resolved value in like `createSyncAgentMentionAvatarsCommand` does for the avatar switches — would need its own re-run subscription to beat the late hydration this PR is working around. Worth remembering as precedent rather than as a defect.
2. **"With it off, nothing changes" is off by about 5px.** With the switch off the glyph is now a `size-[18px]` centred box around the same 13px icon instead of the bare 13px icon, so the chip is roughly 5px wider than before. Height is unchanged and the surface is behind the non-GA `ComposerCreateCommands`, so the blast radius is staff only — but the body's claim is not literally true.
3. **Codecov's 3 uncovered lines** are the two `templateChipCover: () => false` initializers and their reset; a mounted composer always overwrites them, so there is no meaningful scenario to add.
4. `createTemplateAttachmentNodeView` tests `if (attachment.previewImageUrl)` while the new code tests `coverUrl === undefined`. The stricter check is the correct one against `ComposerTemplateAttachment`; noting only that the two chips now differ in style.

### Testing

Coverage matches the behavior the diff adds, and it is placed at the right boundary. The switch-off case is the one that matters most here: per the PR body it is what caught the latched-boolean bug, and it is also the regression guard proving the GA path is untouched. The replacement case covers the in-place `setNodeMarkup` rewrite that a first-render-only implementation would miss, and the video-template case covers the genuinely coverless branch.

Verification reported in the body — the 19-test target file, `packages/core` (243), `check-types`, full platform `lint`, `prettier --check` — matches the changed surfaces. All required checks are green on `f12572f9`, including `lint-style` and `lint-eslint`.

No material gap. Chip geometry itself (the 18/28 vs 20/32 proportion) is not machine-checkable and is carried by the linked design page.

PR #33333, opened 2026-09-10
refactor(platform): stabilize workflow realtime handler

e7h4n:
## Summary

- define the workflow realtime handler in the package-scope command graph
- preserve the realtime event signal while reloading every mounted composer
- remove the corresponding `ccstate/no-command-in-command` suppression

The three remaining suppressions in `create-chat-thread.ts` depend on thread-owned reload commands. Moving them only to the surrounding factory would hide runtime command creation without making the graph static, so they remain unchanged.

## Verification

- `git diff --check`
- Existing page-level coverage: `Refresh workflow suggestions without disrupting the draft` exercises primary and split-thread workflow notifications and verifies both mounted composers refresh without losing draft content
- Local lint and Vitest were not run because this fresh checkout has no installed dependencies; the PR pipeline is the verification boundary

### DIFF
--- a/turbo/apps/platform/src/signals/chat-page/create-chat-thread.ts
+++ b/turbo/apps/platform/src/signals/chat-page/create-chat-thread.ts
@@ -3029,6 +3029,13 @@ function createOnSubscribedCommand({
   });
 }
 
+const onWorkflowsChanged$ = command(
+  async ({ set }, signal: AbortSignal): Promise<boolean> => {
+    await set(reloadMountedComposerWorkflows$, signal);
+    return false;
+  },
+);
+
 function createRunTracking({
   threadId,
   setupChatEvents$,
@@ -3075,15 +3082,6 @@ function createRunTracking({
       set(reloadArtifacts$);
       return false;
     });
-
-    // eslint-disable-next-line ccstate/no-command-in-command -- migrate this runtime callback to the static command graph
-    const onWorkflowsChanged$ = command(
-      async ({ set }, signal: AbortSignal): Promise<boolean> => {
-        L.debug("onWorkflowsChanged$ fired", { threadId });
-        await set(reloadMountedComposerWorkflows$, signal);
-        return false;
-      },
-    );
 
     await Promise.all([
       set(syncHydratedEventTrees$, signal),


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 1668614f587fec76dceaed4ec5bcd462166e8e4b
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `1668614f587fec76dceaed4ec5bcd462166e8e4b`
Practice revision: `cc1178b92c62a403b6052768df40d7a04ea40933` (`main`)

### Summary

Moves the workflow-change realtime handler from execution-time construction inside `subscribeChatThread$` to one package-scope command. The handler still awaits the global mounted-composer workflow reload under the realtime loop's `AbortSignal` and returns `false`, so primary and split-thread subscriptions keep refreshing every mounted composer without changing draft content.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace the changed realtime path, including concurrency, cancellation, and observable outcomes | Pass | `subscribeChatThreadRealtime$` still registers `onWorkflowsChanged$` for each thread-specific `chatThreadWorkflowsChanged:<threadId>` topic; `setAblyLoop$` invokes it with the subscription loop signal, and the handler awaits `reloadMountedComposerWorkflows$` before returning `false`. A command identity carries no per-invocation isolation, so sharing this stateless command does not merge the two thread subscriptions. |
| `REVIEW.md` | Confirm meaningful coverage still protects an implementation-only refactor | Pass | `chat-composer-workflows.test.tsx`, `Refresh workflow suggestions without disrupting the draft`, boots the real page with primary and split-thread composers, triggers both realtime topics, and verifies visible workflow highlighting and insertion while preserving each draft. |
| `REVIEW.md` | Assess public contracts, persisted data, compatibility, fallbacks, auth, and permissions | N/A | The diff changes one internal command declaration and removes one lint suppression; no request, protocol, schema, persisted event, fallback, feature switch, auth, or permission boundary changes. |
| `docs/bad-smell.md` | Remove the underlying lint violation without adding suppressions or unsafe typing | Pass | `create-chat-thread.ts` deletes one `ccstate/no-command-in-command` suppression and introduces no `any`, assertion, catch, fallback, dynamic import, URL, or configuration change. |
| `docs/testing.md` | Test the external Platform contract through the production page boundary | Pass | Existing coverage uses `setupPage`, realtime through `context.mocks.ably`, and page-visible editor/menu outcomes; no implementation-specific test is added for declaration placement. |
| `docs/testing/app-testing.md` | Exercise realtime through the test context and synchronize on visible state | Pass | The existing page test installs mocks before `setupPage`, triggers the external realtime source once per mounted thread, and waits for workflow highlights, suggestion visibility, and preserved editor text. |
| `docs/testing/testing-external-behavior.md` | Preserve tests that verify behavior rather than command identity | Pass | Coverage asserts what users can see and insert in both composers; it does not assert whether `onWorkflowsChanged$` is constructed inside or outside the owning factory. |
| `.claude/skills/ccstate/SKILL.md` | Keep async work in semantic commands with explicit cancellation ownership | Pass | The package-scope handler remains an async command, accepts `AbortSignal` as its final argument, passes it to the awaited reload command, and introduces no detached work. |
| `.claude/skills/ccstate/references/commands.md` | Construct the command graph before commands execute | Pass | The only command that has no thread-owned dependency is lifted to package scope, exactly removing runtime `command()` creation; the three thread-owned handlers remain out of this focused diff rather than being hidden in another runtime factory. |
| `.claude/skills/ccstate/references/lifecycle.md` | Preserve the caller-owned signal through awaited sub-commands | Pass | `runWithChannel$` supplies its live loop signal, `onWorkflowsChanged$` forwards that signal to `reloadMountedComposerWorkflows$`, and the latter awaits every mounted editor sync under both mount and loop ownership. |
| `docs/effect.md` | Keep route subscriptions in setup commands and compose effects through commands | Pass | `subscribeChatThread$` still owns the thread subscription lifecycle; only the stateless callback identity moves, while subscription startup, topic ownership, awaiting, and teardown stay unchanged. |

### Findings

No P0 or P1 findings.

The removed thread-specific debug line is not a behavior or diagnostic gap: the shared realtime loop continues to log each received topic, whose value includes the thread ID.

### Testing

- Exact reviewed HEAD reports all three required gates passed: `ci-gate-crates`, `ci-gate-security`, and `ci-gate-turbo`.
- All eight Platform app test shards passed on this HEAD, and Codecov reports all modified coverable lines covered.
- Existing page-level coverage exercises simultaneous primary/split-thread workflow notifications and visible draft-preserving refresh behavior.
- No local suite was rerun because the reviewed commit has no new changes beyond the already green exact-head pipeline.

PR #33411, opened 2026-09-11
fix(ci): retry interrupted runner binary uploads safely

seven332:
## Summary

- Recover recognized transient Runner executable upload failures with at most two attempts, five seconds apart, using an independently allocated remote temporary file for each attempt.
- Record the host, attempt, expected bytes, candidate path, elapsed time, and original exit status while retaining SSH stderr. Apply a 120-second operation budget only to the upload through the existing SSH wrapper.
- Keep checksum verification, executable validation, and atomic publication. Ordinary command errors, cancellation/kill statuses, and validation failures remain fatal; do not replay cleanup, GC, setup, or the whole image build.

Closes #33410

## Evidence and scope

In [the failed ARM64 producer](https://github.com/vm0-ai/vm0/actions/runs/34555530623/job/103127471117), CI validated a 48,468,504-byte Runner executable but read-only host inspection found only 4,063,232 bytes in its upload candidate. No final executable or subsequent checksum command existed. The downstream manifest wait failed because this producer failed, not because its wait budget expired.

The original transport exit status and underlying interruption cause are unknown. This PR repairs missing bounded recovery and diagnostics; it does not claim a confirmed Cloudflare outage or kernel regression. No kernel, production API/protocol, UI, manifest schema, global SSH retry, or workflow timeout changes.

## Recovery and risks

- Only upload statuses 255 (SSH failure), 124 (operation timeout), and 141 (broken pipe) get one retry. The existing SSH wrapper still owns pre-operation probes/reconnection and single-shot command execution.
- Fresh candidates prevent a late remote writer from a failed attempt corrupting a successful retry. Abandoned candidates stay unpublished under the existing job-directory cleanup policy; checksum/executable validation failures still clean up their selected candidate.
- An exceptionally slow transfer can hit the shorter upload-only budget. A retry adds one transfer and a five-second delay; healthy uploads add only one candidate-allocation SSH operation. Existing probe/recovery time is separate from the stream budget. Permanent SSH failures can also report 255 and therefore receive at most one extra attempt.

## Validation

Passed locally:

```text
bash -n .github/scripts/prepare-runner-image.sh
bash -n .github/scripts/tests/prepare-runner-image-test.sh
shellcheck -x -P .github/scripts .github/scripts/prepare-runner-image.sh .github/scripts/tests/prepare-runner-image-test.sh
bash .github/scripts/tests/prepare-runner-image-test.sh
bash .github/scripts/tests/runner-image-context-test.sh
[...]

PR #33432, opened 2026-09-11
fix(ci): retry transient desktop notarization failures

lancy:
## Problem

Desktop release `0.48.32` failed in [run 34560910027](https://github.com/vm0-ai/vm0/actions/runs/34560910027), job `promote-desktop-release`, step *Sign and notarize macOS artifacts*:

```
Submission ID received  id: e1ab3d80-ee41-48d2-a3a8-e5f110883b9e
Successfully uploaded file
Waiting for processing to complete.
Current status: In Progress.......
Error: HTTPError(statusCode: nil, error: NSURLErrorDomain Code=-1001 "The request timed out."
       URL=https://appstoreconnect.apple.com/notary/v2/submissions/e1ab3d80-...)
##[error]Process completed with exit code 1
```

Signing, the DMG build, and the upload to Apple all succeeded. One status poll inside `xcrun notarytool submit --wait` hit the CFNetwork 60s timeout, notarytool exited 1, and under `bash -e` that ended the job — while Apple still reported the submission as `In Progress`. `okou-desktop-v0.48.32` was never created and `publish-desktop-update-manifest` was skipped, so users stayed on `0.48.31`.

The step had no retry protection of any kind. This is the same class of failure as #33411 (interrupted runner binary uploads).

## Change

Retry the DMG notarization twice (3 attempts total) before failing, with a 30s pause between attempts.

A terminal Apple verdict short-circuits the loop: when notarytool reports `status: Invalid` or `status: Rejected`, the artifact itself is the problem, so the step fails on the first attempt rather than re-uploading a 163MB DMG two more times.

`set -o pipefail` is added to the step because the submission is now piped to `tee` to keep the notary log inspectable. Without it, `tee` would return 0 and mask every notarytool failure.

Out of scope: the `.app` notarization in `sign-and-notarize-packaged-app.mjs` goes through `@electron/notarize` and has its own retry behavior; it is untouched here.

## Verification

- `.github/scripts/tests/deploy-desktop-workflow-test.sh` extended with the retry contract (single canonical `notarytool submit` invocation, bounded at 3 attempts, no retry on a terminal verdict, `pipefail` present, stapling only after success) — passes.
- The extracted step body was executed against a stubbed `xcrun` across every path:

  | scenario | exit | submits | stapled |
  | --- | --- | --- | --- |
  | accepted immediately | 0 | 1 | yes |
  | 1 timeout, then accepted | 0 | 2 | yes |
  | 2 timeouts, then accepted | 0 | 3 | yes |
  | 3 timeouts | 1 | 3 | no |
  | rejected | 1 | 1 | no |
  | 1 timeout, then rejected | 1 | 2 | no |

- `shellcheck --shell=bash` clean on the rendered step body.

### DIFF
--- a/.github/scripts/tests/deploy-desktop-workflow-test.sh
+++ b/.github/scripts/tests/deploy-desktop-workflow-test.sh
@@ -144,6 +144,12 @@ ruby -e '
     "--issuer \"#{dollar}OKOU_DESKTOP_NOTARIZE_API_ISSUER\"",
   ]
   raise "Desktop DMG notarization must consume the canonical API credential triple" unless canonical_notarytool_arguments.all? { |argument| notarize_run.include?(argument) }
+  raise "Desktop DMG notarization must submit through a single canonical invocation" unless notarize_run.scan("xcrun notarytool submit").length == 1
+  raise "Desktop DMG notarization must survive a transient notary poll failure" unless notarize_run.include?("for attempt in 1 2 3")
+  raise "Desktop DMG notarization must bound notary retries" unless notarize_run.include?("Notarization did not complete after 3 attempts")
+  raise "Desktop DMG notarization must not retry a terminal Apple verdict" unless notarize_run.include?("status: (Invalid|Rejected)")
+  raise "Desktop DMG notarization must not let tee mask a notarytool failure" unless notarize_run.include?("set -o pipefail")
+  raise "Desktop DMG stapling must follow a successful notarization" unless notarize_run.index("stapler staple") > notarize_run.index("notarize_status")
   raise "Desktop promotion must not rebuild the app" if promote_text.include?("pnpm -F @okouai/desktop build")
   raise "Desktop promotion must sign the downloaded app" unless promote_text.include?("sign-and-notarize-packaged-app.mjs")
   raise "Desktop promotion must publish an independent Okou release" unless promote_text.include?("OKOU_RELEASE_TAG: okou-desktop-v")
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -576,6 +576,8 @@ jobs:
           OKOU_DESKTOP_NOTARIZE_API_KEY_ID: ${{ secrets.APP_STORE_CONNECT_API_KEY_ID }}
           OKOU_DESKTOP_NOTARIZE_API_ISSUER: ${{ secrets.APP_STORE_CONNECT_API_ISSUER_ID }}
         run: |
+          set -o pipefail
+
           app_dir="${{ steps.desktop-app.outputs.okou_app_dir }}"
           zip_artifact_path="apps/desktop/out/Okou-darwin-arm64-$DESKTOP_VERSION.zip"
           dmg_artifact_path="apps/desktop/out/Okou-darwin-arm64-$DESKTOP_VERSION.dmg"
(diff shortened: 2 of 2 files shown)


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: 0205429e721efe79b53596dc5b65194ca2d2b0fd
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `0205429e721efe79b53596dc5b65194ca2d2b0fd`
Practice revision: `1c005ddbb297aec5f576d4d9063f82c61d809be8`

### Summary

Wraps the Desktop DMG notarization in `promote-desktop-release` in a bounded 3-attempt retry so a single timed-out Apple notary status poll no longer discards an entire release, adds `set -o pipefail` because the submission is now piped to `tee`, and short-circuits on a terminal `Invalid`/`Rejected` verdict so a genuinely bad artifact still fails on the first attempt. `deploy-desktop-workflow-test.sh` is extended with the corresponding workflow contract.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | Only consumer is step 13 of `promote-desktop-release`; step 14 (`Verify signed and notarized macOS artifacts`) and `publish-desktop-update-manifest` still run only after a successful, stapled notarization. |
| `REVIEW.md` | Check failure outcomes, ordering, and cancellation | Pass | `release-please.yml:596-626` — `stapler staple` is reachable only via `notarize_status=0`; all three failure exits are non-zero with `::error::` annotations. |
| `REVIEW.md` | Require meaningful coverage for new behavior and regression fixes | Pass | `deploy-desktop-workflow-test.sh:147-152` pins the retry bound, the terminal-verdict guard, `pipefail`, and staple ordering. See the non-blocking note below on the limits of contract-level assertions. |
| `REVIEW.md` | Report concrete practice violations with rule and source location | Pass | No violations found. |
| `docs/bad-smell.md` §3 | Catch errors only at a boundary that meaningfully handles them; preserve legitimate retry/recovery | Pass | The retry sits at the only boundary that can re-drive an Apple notary submission, and it narrows rather than broadens what is swallowed: a terminal verdict exits immediately. |
| `docs/bad-smell.md` §6 | No hardcoded URLs, credentials, or invented configuration | Pass | Credentials still flow solely from the step `env` block; the diff adds no new credential reference and keeps each canonical alias at exactly two occurrences. |
| `docs/bad-smell.md` §7 | Default is no fallback for a state the contract rules out | Pass | Retrying a real, externally reachable CFNetwork timeout is recovery of the same operation, not a substitute value for an impossible state. |
| `docs/fallback.md` §6/§9 | Every introduced fallback declared in the PR summary with surface, gate, removal condition | N/A | The diff introduces no fallback: no default value, no tolerance branch for a retired producer, no dual-read. The grep guard fails *earlier*, it does not accept a degraded input. |
| `docs/fallback.md` §1 | No negative tests that only pin removed behavior | Pass | The new assertions pin current required behavior of the notarize step, not the absence of deleted code. |
| `docs/testing.md` | Test the contract an external caller relies on, through the production entry point | Pass | The test reads the real `release-please.yml` step body that CI executes, via the established Ruby YAML harness already used for this job. |
| `docs/testing.md` | Avoid cases that merely restate static configuration | Pass | Each assertion maps to a failure mode that has or could occur: unbounded retries, retrying a rejected artifact, and `tee` masking notarytool's exit status. |

### Findings

No P0 or P1 findings.

Non-blocking observations:

- `deploy-desktop-workflow-test.sh` verifies the retry *contract* by matching the step body, not the retry's runtime behavior; inline YAML cannot be executed the way `.github/scripts/*.sh` can (cf. `prepare-runner-image-test.sh`). It does protect the realistic regressions — removing the loop, widening the bound, dropping `pipefail`, or reordering the staple. Extracting the notarization into a `.github/scripts/` helper would allow a stub-driven execution test, but it would relocate the canonical credential triple and change the deliberate "exactly twice" contract asserted at `deploy-desktop-workflow-test.sh:134-137`, which is out of scope here.
- `notarize_run.index("stapler staple") > notarize_run.index("notarize_status")` raises `NoMethodError` on `nil` rather than the intended message if either anchor disappears. The test still fails CI, only the diagnostic is less clear.
- Three attempts can consume roughly three notarization budgets inside the job's 60-minute `timeout-minutes`. A pathologically slow Apple notary run could now hit that timeout, but the outcome is the same unpublished release as today, and the terminal-verdict short-circuit keeps the expensive path limited to genuinely transient failures.

### Testing

- `.github/scripts/tests/deploy-desktop-workflow-test.sh` passes; it runs unconditionally in the `Run workflow script tests` step of `security.yml`, so the new contract is enforced on every PR.
- The PR body records a stub-driven execution of the extracted step body across all six paths (immediate accept, one and two transient failures followed by accept, three transient failures, immediate reject, transient-then-reject), with submit counts and staple/exit outcomes matching the intended semantics — including that the `tee` log is truncated per attempt, so a stale `Invalid` line cannot leak into a later attempt.
- `shellcheck --shell=bash` is clean on the rendered step body, which matters because `actionlint` shellchecks `run:` blocks in the required Workflow Lint check.
- No gap identified that would change the verdict.

PR #33455, opened 2026-09-11
perf(runner): bound retry-after header inspection

### LINKED ISSUE #33424: perf(python): bound Retry-After inspection before decoding header values
## Problem

Model-provider failure reporting decodes every `Retry-After` header value before checking whether there is exactly one usable ASCII delay. Oversized, invalid, or duplicate fields therefore cause avoidable synchronous allocation and scanning in the response-header hook, even though the optional cooldown hint will be discarded.

Audited at `eaec5d018f6e9aa2bf2be7f4bdd59312319ad7fe`. The complete affected file is unchanged at current-main snapshot `c13be861672a425555d9dffdf3d29da982131bd8` (blob `41e789fc76683431ad4b750dcd79cdcf3a17bc03`).

## Location and cause

- [`crates/runner/mitm-addon/src/model_provider_failure.py:781-793`](https://github.com/vm0-ai/vm0/blob/eaec5d018f6e9aa2bf2be7f4bdd59312319ad7fe/crates/runner/mitm-addon/src/model_provider_failure.py#L781-L793): `headers.get_all("Retry-After")` materializes all matching values as strings before cardinality and ASCII checks. A valid numeric value is then scanned by `isdigit()` and `lstrip("0")` without an input bound. Clamping the final integer to 300 seconds does not bound this work.
- [`model_provider_failure.py:358-387`](https://github.com/vm0-ai/vm0/blob/eaec5d018f6e9aa2bf2be7f4bdd59312319ad7fe/crates/runner/mitm-addon/src/model_provider_failure.py#L358-L387): admitted 429/503 responses invoke this parser during response-observer setup. The same helper is used by terminal response/error settlement.
- [`crates/runner/mitm-addon/src/mitm_addon.py:1601-1612`](https://github.com/vm0-ai/vm0/blob/eaec5d018f6e9aa2bf2be7f4bdd59312319ad7fe/crates/runner/mitm-addon/src/mitm_addon.py#L1601-L1612): response stream/observer setup runs synchronously in `responseheaders()`.

The locked mitmproxy 12.2.3 HTTP/1 response-head parser accepts these fields. Its `Headers.get_all()` calls `_native()` for every matching value, decoding UTF-8 with `surrogateescape`; this includes both values of an already unusable duplicate header. Existing JSON/body limits and WebSocket header limits do not guard this path.

Measured with the locked CPython 3.14.0 / mitmproxy 12.2.3 environment, invoking the real `mitm_addon.responseheaders()` hook on an admitted OpenAI model-provider 429 flow. Input construction, HTTP-head parsing, imports, and admission finished before measurement. Reporting was left unconfigured, so no external request or network body was involved.

| Retry-After input | Hook peak temporary allocation | Hook time with tracemalloc | Parser median without tracemalloc (5 calls) |
| --- | ---: | ---: | ---: |
| `120` | 13,102 bytes | 0.637 ms | 0.002 ms |
| 8 MiB of ASCII zeros followed by `120` | 8,389,628 bytes | 28.649 ms | 27.342 ms |
| One 8 MiB non-UTF-8 value | 25,166,823 bytes | 66.487 ms | 56.988 ms |
| Two 8 MiB non-UTF-8 values | 41,944,129 bytes | 124.460 ms | 117.376 ms |

The zero-prefixed case returns 120; both non-UTF-8 cases return `None`. An independent spy on mitmproxy's conversion boundary confirmed that duplicate rejection still converts both complete 8,388,608-byte values. Original response headers remain unchanged.

This is an adversarial oversized-upstream-header case, not evidence that ordinary provider headers are this large or that a production incident occurred. It adds allocation and event-loop delay on top of the dependency's existing raw HTTP/1 head buffering. Fixing this helper does not by itself bound the dependency's header buffer or every other header consumer.

Minimal hook reproduction from `crates/runner/mitm-addon` after installing the locked environment:

```bash
PYTHONPATH=src uv run --no-sync python - <<'PY'
import gc
import time
import tracemalloc
from mitmproxy import http
from mitmproxy.net.http import http1
from mitmproxy.test import tflow
import flow_metadata_keys as keys
import mitm_addon
import model_provider_failure

value = b"\xff" * (8 * 1024 * 1024)
flow = tflow.tflow()
flow.request = http.Request.make(
    "POST", "https://api.openai.com/v1/chat/completions", b"{}"
)
flow.response = http1.read_response_head([
    b"HTTP/1.1 429 Too Many Requests",
    b"Content-Type: application/json",
    b"Content-Length: 2",
    b"Retry-After: " + value,
    b"Retry-After: " + value,
])
flow.metadata.update({
    keys.SANDBOX_RUN_ID: "audit-retry-after",
    keys.SANDBOX_PROXY_LOG_PATH: "",
    keys.FIREWALL_NAME: "model-provider:openai-api-key",
    keys.FIREWALL_BILLABLE: True,
    keys.ORIGINAL_URL: "https://api.openai.com/v1/chat/completions",
})
model_provider_failure.configure_reporting(api_url="", bearer_credential="")
model_provider_failure.admit_flow(flow)
original_headers = flow.response.headers.fields
gc.collect()
tracemalloc.start()
started = time.perf_counter()
mitm_addon.responseheaders(flow)
elapsed = time.perf_counter() - started
peak = tracemalloc.get_traced_memory()[1]
tracemalloc.stop()
print({"peak_bytes": peak, "hook_ms": elapsed * 1000})
assert flow.response.headers.fields == original_headers
model_provider_failure.release_flow(flow)
PY
```

## Acceptance criteria

- Apply explicit field-count and raw-value/work bounds before value decoding, copying, or unbounded name normalization. Detect repeated fields from bounded raw header metadata without decoding their values.
- If the optional hint is missing, duplicated, invalid, or over budget, omit `retryAfterSeconds` while preserving the status-derived failure report and original wire headers.
- Preserve existing in-budget behavior: zero becomes 1, valid leading zeros are accepted, values above 300 clamp to 300, and date/non-ASCII/non-numeric values are omitted. Retain the existing 5,000-digit numeric-delay regression or explicitly resolve its compatibility contract.
- Add response-header-hook regression coverage for oversized ASCII/non-UTF-8 values, duplicates, and field-count exhaustion. Guard conversion/work structurally rather than adding wall-clock thresholds; verify the emitted failure payload remains correct.
- Keep existing model-provider failure reporting/shutdown tests green. This audit ran 822 targeted tests successfully, including those suites, so the reproduction identifies a missing bound rather than an existing test failure.

Related but distinct: #33313 addressed Content-Type media-type classification and is already closed; its bounded raw-header implementation does not change this Retry-After parser. Open-issue title searches for Retry-After, retry, header, and provider failure found no matching issue before filing.

seven332:
Oversized or duplicate `Retry-After` fields caused model-provider failure reporting to decode complete values synchronously before discarding the hint. Inspect the existing raw header tuple with an 8,192-field limit and an 8,192-byte value limit, then validate decimal bytes after singleton checks. Name-length checks also bound case normalization.

Preserve status-derived reporting and original wire headers when the hint is unusable. In-budget behavior retains leading zeroes, zero -> 1, the 300-second clamp, strict whitespace rejection, and the existing 5,000-digit regression. Hook-level tests cover oversized HTTP/1 ASCII/non-UTF-8 fields, duplicates, and budget boundaries; observer-level guards isolate name-normalization bounds from subsequent body decoding.

The scope is the optional Retry-After parser. Mitmproxy's original HTTP-head buffering and other consumers, including Content-Encoding lookup, retain their existing behavior. The report payload shape is unchanged.

Closes #33424

Validation in the locked addon environment (uv 0.11.32 / Python 3.14.0 / mitmproxy 12.2.3):

- Confirmed new oversized-value guards fail against the previous parser.
- `uv run --no-sync python -m pytest tests/test_model_provider_failure_reporting.py tests/test_model_provider_failure_shutdown.py tests/test_response_headers_handler.py tests/test_response_content_type_budget.py -q` — 158 passed.
- `uv run --no-sync ruff check src/model_provider_failure.py tests/test_model_provider_failure_reporting.py` — passed.
- `uv run --no-sync ruff format --check src/model_provider_failure.py tests/test_model_provider_failure_reporting.py` — passed.
- `uv run --no-sync basedpyright -p .` — 0 errors/warnings.
- `git diff --check` — passed.

### DIFF
--- a/crates/runner/mitm-addon/src/model_provider_failure.py
+++ b/crates/runner/mitm-addon/src/model_provider_failure.py
@@ -92,6 +92,10 @@ _FLOW_STATE = "_model_provider_failure_flow_state"
 _RESPONSE_FINISH = "_model_provider_failure_response_finish"
 _MAX_JSON_WORK_UNITS = 65_536
 _MAX_RETRY_AFTER_SECONDS = 300
+_RETRY_AFTER_NAME = b"retry-after"
+_MAX_RETRY_AFTER_FIELDS = 8 * 1024
+# Preserve long numeric delays while bounding the optional hint's synchronous work.
+_MAX_RETRY_AFTER_VALUE_BYTES = 8 * 1024
 # Allow API cold starts while keeping best-effort reporting and shutdown bounded.
 _REPORT_TIMEOUT_SECONDS = 10
 _REPORT_WORKERS = 4
@@ -781,13 +785,23 @@ def _failure_kind_from_http_status(status: int, firewall_name: str) -> FailureKi
 def _retry_after_seconds(status: int, headers: http.Headers) -> int | None:
     if status not in (_HTTP_STATUS_TOO_MANY_REQUESTS, _HTTP_STATUS_SERVICE_UNAVAILABLE):
         return None
-    values = headers.get_all("Retry-After")
-    if len(values) != 1 or not values[0].isascii() or not values[0].isdigit():
+    fields = headers.fields
+    if len(fields) > _MAX_RETRY_AFTER_FIELDS:
         return None
-    digits = values[0].lstrip("0")
+    value: bytes | None = None
+    for name, raw_value in fields:
+        if len(name) != len(_RETRY_AFTER_NAME) or name.lower() != _RETRY_AFTER_NAME:
+            continue
+        if value is not None or len(raw_value) > _MAX_RETRY_AFTER_VALUE_BYTES:
+            return None
+        value = raw_value
+    # Validate only after bounded cardinality inspection, without decoding or trimming OWS.
+    if value is None or not value.isdigit():
+        return None
+    digits = value.lstrip(b"0")
     if not digits:
         return 1
-    maximum = str(_MAX_RETRY_AFTER_SECONDS)
+    maximum = str(_MAX_RETRY_AFTER_SECONDS).encode("ascii")
     if len(digits) > len(maximum) or (len(digits) == len(maximum) and digits > maximum):
         return _MAX_RETRY_AFTER_SECONDS
     return int(digits)
--- a/crates/runner/mitm-addon/tests/test_model_provider_failure_reporting.py
+++ b/crates/runner/mitm-addon/tests/test_model_provider_failure_reporting.py
@@ -16,6 +16,7 @@ import zstandard
 from mitmproxy import http
 from mitmproxy.connection import ConnectionState
 from mitmproxy.flow import Error
+from mitmproxy.net.http import http1
 
 import body_decoding
 import flow_metadata_keys as metadata_keys
@@ -40,6 +41,8 @@ from tests.thread_helpers import ThreadUnderTest, wait_for_event
 
 _REPORT_CAPACITY = 16
 _REPORT_WORKERS = 4
+_RETRY_AFTER_FIELD_LIMIT = 8 * 1024
+_RETRY_AFTER_VALUE_LIMIT = 8 * 1024
 
 
 def _make_flow(
@@ -884,6 +887,7 @@ def test_shutdown_timeout_returns_with_running_reports_blocked(
     ("status", "retry_after", "expected_kind", "expected_seconds"),
     [
         (429, "0", "rate_limit", 1),
+        (429, "000120", "rate_limit", 120),
         (503, "301", "provider_unavailable", 300),
         pytest.param(
             503,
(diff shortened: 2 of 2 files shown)


seven332 (commented):
LGTM

Reviewed head: `65c07532b900ea78ef729e4e93deb08e9ff91094`. Practice revision: `a83ffce3cde42c90eff96bab4d533862b4713c4f`.

**Summary:** Raw field-count, name-length, singleton, and value-length checks now bound Retry-After inspection before scanning or copying. Strict ASCII-decimal parsing, the 5,000-digit case, the 1–300 second result range, status-derived reports, and wire headers are preserved within budget.

**Findings:** No in-scope findings after the six-pass self-review and code-quality review. Existing downstream Content-Encoding header traversal is independently tracked in #33459 and is not a dependency of this PR.

**Testing:** 158 affected tests passed in the locked Python/mitmproxy environment; Ruff lint/format and BasedPyright passed. Regression coverage uses real response-header hooks and local HTTP reports, oversized HTTP/1 values, duplicate/empty/late fields, inclusive budget boundaries, and structural conversion/name-work guards. New oversized-value guards were confirmed failing on the previous parser. Original HTTP-head buffering and the full local addon suite are outside the verification performed here; CI is checked separately.

This is a self-authored `COMMENT` review, not independent GitHub approval.

seven332:
## Code Review: PR #33455

**Reviewed head:** `65c07532b900ea78ef729e4e93deb08e9ff91094`

### Summary
Bound raw Retry-After inspection while preserving in-budget numeric behavior and status reports. [Commit-bound review](https://github.com/vm0-ai/vm0/pull/33455#pullrequestreview-5175557937).

### Findings
No in-scope findings. Separate pre-existing decoder traversal is tracked in #33459.

### Testing
158 affected tests passed; Ruff lint/format and BasedPyright passed. Structural hook/observer guards and retained 5,000-digit regression cover the work and compatibility boundaries. Broader CI is checked separately.

### Verdict
LGTM. Six-pass self-review is clean on this unchanged head; the self-authored review does not satisfy independent approval.

PR #33460, opened 2026-09-11
perf(python): bound json nesting guard string scanning

### LINKED ISSUE #33445: perf(python): bound JSON nesting guard string scanning
## Summary

`json_nesting_within_limit()` uses a backtracking regex to skip complete strings when the raw opening-delimiter count exceeds 256. A valid, shallow 65,535-byte request body makes that pre-check allocate about 8 MiB temporarily. Changing one character so the same-size document takes the fast path reduces the entire response hook's peak to about 193 KiB.

This guard runs synchronously during X tweet-create billing refinement. Its input byte cap is working; the problem is the large regex working-memory multiplier within that cap. The same regex also exhibits quadratic retry work on unterminated strings containing many escaped quotes.

## Affected code and reachability

Line numbers refer to audited commit `fc0b3233678aa1201dc31d7bd01867312e06a747`.

- `crates/runner/mitm-addon/src/usage/json_selective.py:91` defines `_JSON_STRING_OR_CONTAINER_RE`; `:101-124` uses `finditer()` after counting raw opening delimiters, including delimiters inside strings.
- `crates/runner/mitm-addon/src/usage/providers/connectors/x_billing.py:538-550` invokes the guard before `json.loads()`. The `POST /2/tweets` rule at `:553-560` uses it to refine `content.create_with_url` to `content.create`.
- `crates/runner/mitm-addon/src/usage/providers/connectors/x.py:660-672` decodes and refines the request body inline. The raw and decoded inspection limits are both 64 KiB (`:77`, and `crates/runner/mitm-addon/src/body_limits.py:9-25`).
- `crates/runner/mitm-addon/src/mitm_addon.py:1746-1758,1788-1821,1885-1886` reaches connector billing synchronously from `response()`.

The X path requires a registered billable flow, run attribution, a known permission, and a **2xx response** (`x.py:604-617`). The other current caller, `crates/runner/mitm-addon/src/usage/openai_chat_completions.py:159-165`, limits its input to **1,024 bytes** first. This report does not claim that the guard processes the generic 5 MiB response-decompression allowance.

## Measured evidence

Environment: Linux, repository-locked CPython 3.14.0 and mitmproxy 12.2.3. Synthetic real `HTTPFlow` objects were sent through the unmodified `mitm_addon.response()` hook, using the repository's local webhook fixture. Payload and flow construction preceded measurement; webhook flushing happened afterward. These are local reproductions, not observed production incidents or live X API acceptance tests.

The valid bodies were generated as:

```python
def valid_body(brackets):
    return (
        b'{"text":"'
        + b"[" * brackets
        + b"a" * (256 - brackets)
        + "中".encode() * 21756
        + b'"}'
    )
```

Both bodies are valid JSON with structural depth 1, contain 22,012 text characters, and are 65,535 bytes long. The 255-bracket control has 256 raw opening delimiters including the root object; the 256-bracket case has 257 and enters the regex path.

| Input | Response-hook peak traced allocation | Hook time with tracing | Billing result |
| --- | ---: | ---: | --- |
| 255 quoted brackets | 197,623 bytes | 1.50 ms | `content.create`, quantity 1 |
| 256 quoted brackets | 8,426,902 bytes | 8.40 ms | `content.create`, quantity 1 |

Isolating the guard itself also produced approximately 8.4 MB of peak traced allocation for a 65.5 KB valid body, locating the amplification before JSON decoding or webhook work.

For malformed input, the regex attempts to find a complete string again at successive escaped quotes. With `body = b'{"text":"' + bytes((92, 34)) * pairs + b"[" * 256`, response-hook timings without tracing were:

| Escaped-quote pairs | Body bytes | Synthetic status | Hook time | Result |
| ---: | ---: | ---: | ---: | --- |
| 4,096 | 8,457 | 201 | 264 ms | Conservative `content.create_with_url` |
| 8,192 | 16,649 | 201 | 1,069 ms | Conservative `content.create_with_url` |
| 8,192 | 16,649 | 400 | 0.47 ms | Skipped; no billing event |

The malformed-body measurements demonstrate the guard's algorithmic worst case and the existing conservative-fallback contract, not that X normally accepts malformed JSON. Ordinary rejected requests do not enter this billing path. `tests/x_connector_usage/test_writes.py:1053-1068` already explicitly covers malformed request bytes paired with a successful response.

## Minimal guard reproduction

From `crates/runner/mitm-addon`, run this with `src` on `PYTHONPATH` in the locked Python 3.14.0 environment:

```python
import gc
import json
import time
import tracemalloc
from usage.json_selective import json_nesting_within_limit

for brackets in (255, 256):
    body = (b'{"text":"' + b"[" * brackets + b"a" * (256 - brackets)
            + "中".encode() * 21756 + b'"}')
    assert len(body) == 65535
    assert len(json.loads(body)["text"]) == 22012
    gc.collect()
    tracemalloc.start()
    assert json_nesting_within_limit(body)
    print("valid", brackets, tracemalloc.get_traced_memory()[1])
    tracemalloc.stop()

for pairs in (4096, 8192):
    body = b'{"text":"' + bytes((92, 34)) * pairs + b"[" * 256
    start = time.perf_counter()
    result = json_nesting_within_limit(body)
    print("unterminated", len(body), result, time.perf_counter() - start)
```

## Suggested direction and acceptance criteria

- Replace the slow-path string/container scan with a scan whose total work is linear in the already-bounded input and whose auxiliary memory stays small, without regex backtracking state proportional to each byte of a long string.
- Do not repeatedly rescan suffixes from escaped quotes after an unterminated string. Merely making the inner repetition possessive is insufficient unless failed-match restart behavior is also addressed.
- Preserve correct depth accounting for valid strings, escaped quotes/backslashes, UTF-8 content, and exact depth boundaries. Preserve conservative billing on uninspectable input and the current 64 KiB admission limits.
- Add deterministic regressions for the same-size 255/256-delimiter boundary, long valid strings, and unterminated escaped-quote sequences; retain the successful-response versus rejected-response gate. Avoid fragile wall-clock assertions in CI.
- Provide same-runtime before/after allocation and scaling measurements. Keep the existing X billing and Chat Completions tests passing.

## Audit and duplicate checks

All 16 randomly sampled files were reviewed. The sampled and related suites passed **1,304 existing tests**, and **five audit-only response-hook reproduction checks** passed. Repository source was not changed.

The guard and relevant caller/limit files are byte-identical on subsequently checked main commit `168ed1102b8a6faf57adfb5e8991b1097411c6b0`. Open-issue searches for nesting, regex, JSON, backtracking, X billing, and Python performance found no duplicate. Closed #24176 addressed X **NDJSON row** parsing, which no longer uses this guard; closed #31430 added the small Chat Completions fast path. Neither resolves the remaining X **request-body** nesting scan described here.

seven332:
## Summary

A shallow 65,535-byte tweet-create request can force the JSON nesting guard to allocate about 8.4 MB while matching its string. Unterminated strings containing escaped quotes also trigger repeated suffix scans. Replace the complete-string regex with forward-only single-byte searches, preserving the raw-delimiter fast path and using constant auxiliary memory.

Add 23 response-hook/webhook regressions for quoted delimiters, escape parity, UTF-8, exact mixed-container depth, many shallow containers, long and unterminated strings, and the request byte limit. The existing 64 KiB admission limits, JSON validation, conservative billing and 2xx response gate remain intact. No schema, dependency or billing-policy change is included.

Closes #33445

## Measurements

Same locked CPython 3.14.0 environment, baseline `8c455bee21f6ab9fe34e952741f62cc00bc0ce1b`. Guard-only peak allocation uses tracemalloc; timings are independent, untraced medians of three calls. Inputs are constructed beforehand, using the issue's reproducer.

| Input | Peak bytes, before → after | Milliseconds, before → after |
| --- | ---: | ---: |
| 65,535-byte UTF-8, 255 quoted brackets | 0 → 0 | 0.058 → 0.064 |
| Same size, 256 quoted brackets | 8,426,223 → 1,582 | 3.393 → 0.294 |
| 65,535-byte ASCII string | 8,426,223 → 1,582 | 2.167 → 0.315 |
| Valid string, 16,384 escaped quotes | 2,206,403 → 1,582 | 0.433 → 3.228 |
| Unterminated, 4,096 escaped quotes | 575,903 → 1,582 | 272.721 → 0.808 |
| Unterminated, 8,192 escaped quotes | 1,128,053 → 1,582 | 1,096.488 → 1.645 |
| Unterminated, 16,384 escaped quotes | 2,206,403 → 1,582 | 4,266.597 → 3.205 |

Dense valid escapes trade native regex repetition for one Python step per escape, so that synthetic case is slower. Long ordinary strings improve, and malformed escaped strings now scale approximately linearly. These are local measurements, not production observations or evidence that X accepts malformed JSON.

## Validation

- The whole-response-hook allocation regression failed against the old implementation at 8,431,139 bytes while the same-size fast-path control passed; it passes after the change with a generous 1 MiB budget. CI has no wall-clock assertions.
- 23 new integration regressions passed: `uv run --no-sync python -m pytest tests/test_x_tweet_create_nesting.py -q`.
- 906 existing tests passed: `uv run --no-sync python -m pytest tests/test_json_selective_*.py tests/x_connector_usage/ tests/test_x_connector_pipeline.py tests/test_openai_chat_completions_usage.py -q`.
- Locked environment and static checks passed: `uv lock --check`, `uv sync --locked`, `uv run --no-sync ruff format --check .`, `uv run --no-sync ruff check .`, `uv run --no-sync basedpyright -p .`; the final added regression also passed targeted format/lint/type checks.
- Full addon and broader repository checks are left to PR CI. No local Vitest run or dev server was started.

### DIFF
--- a/crates/runner/mitm-addon/src/usage/json_selective.py
+++ b/crates/runner/mitm-addon/src/usage/json_selective.py
@@ -88,7 +88,8 @@ _SLOW_SCALAR_BYTES_PER_WORK_UNIT = 32
 _DISCARDED_STRING_BULK_BYTES_PER_WORK_UNIT = 64 * 1024
 JSON_WORK_LIMIT_EXCEEDED = "work limit exceeded"
 JSON_INTEGER_VALUE_LIMIT_EXCEEDED = "integer value limit exceeded"
-_JSON_STRING_OR_CONTAINER_RE = re.compile(rb'"(?:\\.|[^"\\])*"|[{}\[\]]', re.DOTALL)
+_JSON_CONTAINER_OR_QUOTE_RE = re.compile(rb'["{}\[\]]')
+_JSON_STRING_ESCAPE_OR_QUOTE_RE = re.compile(rb'["\\]')
 
 
 def _validate_positive_int(name: str, value: int) -> None:
@@ -103,19 +104,32 @@ def json_nesting_within_limit(body: bytes, *, max_depth: int = _DEFAULT_MAX_DEPT
 
     This is a boundedness guard, not a JSON validator. The fast path accepts
     payloads with at most ``max_depth`` opening delimiters. The slow path
-    ignores delimiters inside complete strings before tracking structural
-    object and array depth. Callers must still parse the body to validate it.
+    ignores delimiters inside strings before tracking structural object and
+    array depth, rejecting unterminated strings. Both paths take linear work
+    and constant auxiliary memory. Callers must still parse the body to
+    validate it.
     """
     _validate_positive_int("max_depth", max_depth)
     if body.count(b"{") + body.count(b"[") <= max_depth:
         return True
 
     depth = 0
-    for match in _JSON_STRING_OR_CONTAINER_RE.finditer(body):
+    position = 0
+    while match := _JSON_CONTAINER_OR_QUOTE_RE.search(body, position):
         token = body[match.start()]
+        position = match.end()
         if token == ord('"'):
-            continue
-        if token in (ord("{"), ord("[")):
+            # Search only unconsumed spans. Single-byte matches avoid retaining
+            # a long string's backtracking state or retrying a failed string
+            # from each escaped quote.
+            while string_match := _JSON_STRING_ESCAPE_OR_QUOTE_RE.search(body, position):
+                position = string_match.end()
+                if body[string_match.start()] == ord('"'):
+                    break
+                position += 1  # A backslash consumes the next byte, even a quote.
+            else:
+                return False
+        elif token in (ord("{"), ord("[")):
             depth += 1
             if depth > max_depth:
                 return False
--- a/crates/runner/mitm-addon/tests/test_x_tweet_create_nesting.py
+++ b/crates/runner/mitm-addon/tests/test_x_tweet_create_nesting.py
@@ -1,0 +1,147 @@
+"""Tweet-create nesting bounds through the response hook and usage webhook."""
+
+import json
+import tracemalloc
+
+import pytest
+
+import mitm_addon
+import usage
+from tests.x_flow_helpers import make_x_pipeline_flow
+
+
+@pytest.fixture
+def tweet_create_response(tmp_path, real_flow, usage_webhook_api, sync_usage_executor):
+    def run(request_body: bytes, *, status: int = 201):
+        flow = make_x_pipeline_flow(
+            real_flow,
+            tmp_path,
+            permission="tweet.write",
+            rule="POST /2/tweets",
+            response_status=status,
+        )
+        flow.request.method = "POST"
+        flow.request.content = request_body
+        assert flow.response is not None
+        flow.response.content = b'{"data":{"id":"1"}}'
+
+        with usage_webhook_api() as webhook:
+            tracemalloc.start()
+            try:
+                mitm_addon.response(flow)
+                peak_bytes = tracemalloc.get_traced_memory()[1]
+            finally:
+                tracemalloc.stop()
+            usage.flush_usage_events(trigger="test")
+
+        return webhook.usage_events(), peak_bytes
+
+    return run
+
+
+@pytest.mark.parametrize("quoted_brackets", [255, 256])
+def test_same_size_utf8_tweet_bounds_response_allocation(tweet_create_response, quoted_brackets):
+    body = (
+        b'{"text":"'
+        + b"[" * quoted_brackets
+        + b"a" * (256 - quoted_brackets)
+        + "中".encode() * 21756
+        + b'"}'
+    )
+    assert len(body) == 65535
+
+    events, peak_bytes = tweet_create_response(body)
+
+    [event] = events
+    assert event["category"] == "content.create"
+    assert event["quantity"] == 1
+    # Allow normal hook/JSON allocations, but not multi-MiB string-scan state.
+    assert peak_bytes < 1024 * 1024
+
+
+@pytest.mark.parametrize(
+    "text",
+    [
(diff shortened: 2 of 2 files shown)


seven332 (commented):
LGTM

**Reviewed head:** `5114c5c87ba81ca2f09acedb713baf4c3cc59a32`
**Practice revision:** main `a83ffce3cde42c90eff96bab4d533862b4713c4f`

**Summary:** The nesting guard now scans forward with constant auxiliary state, preserving caller-owned JSON validation, depth/admission bounds and billing outcomes. Completed all six self-review passes and a code-quality review of the submitted diff and both callers.

**Findings:** No findings. The cursor consumes every searched span, including escaped bytes, and aborts an unterminated slow-path string without retrying suffixes. No new fallback or cross-version contract is introduced. The PR explicitly records the slower dense-valid-escape case alongside the memory and malformed-input improvements.

**Testing:** 23 new real-response-hook/webhook regressions and 906 relevant existing tests passed. The new allocation regression demonstrably fails the old implementation and passes this head. Locked Ruff, BasedPyright and lockfile checks passed. Full addon/broader CI and live X acceptance are not established by local validation. This self-authored `COMMENT` review does not supply independent approval or authorize merging.

seven332:
## Code Review: PR #33460

**Reviewed head:** `5114c5c87ba81ca2f09acedb713baf4c3cc59a32`
**Verdict:** LGTM — no findings after a clean six-pass self-review and code-quality review.

The [commit-bound review](https://github.com/vm0-ai/vm0/pull/33460#pullrequestreview-5175563023) records the scope, correctness/resource assessment and limitations. Local validation passed 23 new and 906 existing tests plus locked static checks. This marker records workflow review state; independent approval and remote CI remain separate gates.

PR #33306, opened 2026-09-10
fix(platform): preserve welcome synchronization after settings dismissal

lancy:
## Summary

Closing Settings after welcome creation returned 201 could abandon shared thread-list synchronization. Later cold threads whose metadata returned 404 then waited indefinitely instead of showing the ordinary not-found page.

Keep committed welcome synchronization owned by the app and its captured user/workspace identity while POST cancellation and navigation remain scoped to the UI. Replace the success-only global barrier with the actual latest synchronization promise: errors propagate, terminal cleanup releases only that task's ownership, and independent explicit refreshes retain their freshness.

Closes #33303. Parent: #33205.

## Verification

- On original main `fc05f140ec37bedaa679b832b88440b93b9bcae8`, the six new Router/API/MessagePort regressions produced **4 failures, 2 passes**. Both post-201 cancellation windows and failed-sync recovery could not reach not-found; overlapping synchronization prematurely showed not-found.
- With the repair, **35 tests passed across six focused files**: welcome creation, the new cancellation/overlap regressions, metadata, offline-cache behavior, and thread-list event ordering. Coverage includes committed history recovery without realtime/reload/recreation, an abandoned concurrent metadata reader, 503 errors, user/workspace changes, POST-pending cancellation, deduplication, and same-UUID retries.
- Platform type checking, targeted ESLint, standard and type-aware Oxlint, formatting, and Platform unused-code analysis passed. No full local Vitest suite or local development server was run.
- Temporarily tying the identity listener back to UI dismissal made both new identity-isolation cases fail by displaying the old committed welcome; the repaired listener lifetime passes them.

The test runs use real application routing and worker/RPC implementations with external browser, Clerk, and HTTP fixtures. No new browser/production validation or production release is claimed. The unchanged S1/S2 content, media, and CLI evidence remains outside this repair's revalidation scope.

PR #33480, opened 2026-09-11
fix(platform): let flyout rows clip at the panel's own edge

tongx-blip:
## Problem

A row scrolling out of the model panel is cut short of the card, leaving a blank
band between the cut and the border, so rows appear and disappear inside the
panel instead of at its edge.

The list is inset by the card's `p-1`, and overflow clips at the scroll
container's own box — that inset — not at the card's rounded edge. Measured on
the reported 2x screenshot: the glyphs stop at device `y=50`, then 8 rows of
white, then the border at `y=59`. That is the `4px` padding plus the `0.5px`
hairline.

## Change

Pull the list back over the inset with `-mt-1` / `-mb-1` and restate it as
scroll padding (`pt-1` / `pb-1`). The first and last rows still rest clear of
the border at the scroll extremes, but a row mid-scroll now runs to the edge.

The bottom bleeds only when nothing follows the list. With the chat-settings row
present, its `border-t` is the boundary, so only the top bleeds and the height
compensates. Both branches keep the visible area at `244px` and the card's
height unchanged, so nothing else moves.

The settings-row condition moves into `showSettingsRow` because the list's
class now depends on it; previously it was inlined in the JSX and could drift
from the geometry that assumes it.

`overflow-hidden` is deliberately not added: a row's own `rounded-lg` keeps its
hover fill inside the card's `12px` corner (verified below), and clipping the
card would also cut row focus rings.

## Verification

Rebuilt the panel from the production class strings against the App's real
Tailwind entry (`apps/platform/src/views/css/index.css`, v4.3.3) and measured in
Chromium at 2x:

| | card height | dead band, top | dead band, bottom | scrollable area |
| --- | --- | --- | --- | --- |
| before | 254px | 5px | 5px | 244px |
| after | 254px | **1px** (the hairline) | **1px** | 244px |

With the chat-settings row: card height 292px before and after, top band 5px ->
1px, list-to-divider gap 2px in both.

A hovered row clipped at the top edge stays inside the rounded corner with no
`overflow-hidden`, checked at 7x.

- `chat-composer-media-models.test.tsx` — 23 passed.
- `chat-composer-model-selection.test.tsx` — 21 passed.
- `pnpm lint:style` exits 0, plus `check-types`, `oxlint`, `prettier --check`.

## Not covered

The drill-in menu layout (`modelPickerMenu`) has the same artifact from
`PopoverContent`'s padding. It is a separate switch and a different fix, so it
is left alone here.

### DIFF
--- a/turbo/apps/platform/src/views/okou-page/components/model-picker-menu.tsx
+++ b/turbo/apps/platform/src/views/okou-page/components/model-picker-menu.tsx
@@ -787,6 +787,14 @@ function ModelPickerFlyoutPanel({
   const savedEffort = switches[FeatureSwitchKey.ChatReasoningEffort]
     ? props.value?.reasoningEffort
     : undefined;
+  const showSettingsRow =
+    !activeMedia &&
+    Boolean(props.value) &&
+    Boolean(selectedOption) &&
+    canAdjustChatSettings(
+      selectedOption,
+      efforts.length > 0 || Boolean(savedEffort),
+    );
   if (!activeMedia && page.kind === "settings" && props.value) {
     return (
       <div
@@ -807,7 +815,19 @@ function ModelPickerFlyoutPanel({
         ref={panelRef}
         role="listbox"
         aria-label={panelLabel}
-        className="flex max-h-[244px] flex-col gap-0.5 overflow-y-auto overscroll-contain"
+        className={cn(
+          // Rows have to appear and disappear at the card's own edge. The card
+          // insets this box by `p-1`, which left a blank band where a row was
+          // cut short of the border, so pull the box back over that inset and
+          // restate it as scroll padding: the list still rests clear of the
+          // border at either end, but a row mid-scroll runs to the edge.
+          // The heights keep the visible area at 244px either way.
+          "-mt-1 flex flex-col gap-0.5 overflow-y-auto overscroll-contain pt-1",
+          showSettingsRow
+            ? "max-h-[248px]"
+            : // Nothing follows, so the bottom reaches the card's edge too.
+              "-mb-1 max-h-[252px] pb-1",
+        )}
       >
         <ModelPickerFlyoutOptions
           {...props}
@@ -822,30 +842,24 @@ function ModelPickerFlyoutPanel({
           </p>
         )}
       </div>
-      {!activeMedia &&
-        props.value &&
-        selectedOption &&
-        canAdjustChatSettings(
-          selectedOption,
-          efforts.length > 0 || Boolean(savedEffort),
-        ) && (
-          <Button
-            variant="ghost"
-            className="h-9 shrink-0 justify-start gap-2 border-t border-border/60 px-2 text-xs text-muted-foreground"
-            aria-label={t(
-              ($) => {
-                return $.settings.models.picker.menu.adjustSettings;
-              },
-              { model: selectedOption.label },
-            )}
-            onClick={editSettings}
-          >
-            <SlidersHorizontal size={14} aria-hidden="true" />
-            {t(($) => {
-              return $.settings.models.picker.menu.chatSettings;
-            })}
-          </Button>
-        )}
+      {showSettingsRow && selectedOption && (
+        <Button
+          variant="ghost"
+          className="h-9 shrink-0 justify-start gap-2 border-t border-border/60 px-2 text-xs text-muted-foreground"
+          aria-label={t(
+            ($) => {
+              return $.settings.models.picker.menu.adjustSettings;
+            },
+            { model: selectedOption.label },
+          )}
+          onClick={editSettings}
+        >
+          <SlidersHorizontal size={14} aria-hidden="true" />
+          {t(($) => {
+            return $.settings.models.picker.menu.chatSettings;
+          })}
+        </Button>
+      )}
     </>
   );
 }


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: 53c06a633c362e5a00004f45b2f314a8d3730a76
PR_AUTO_REVIEW_VERDICT: LGTM

Reviewed HEAD: `53c06a633c362e5a00004f45b2f314a8d3730a76`
Practice revision: `f222d6930ff95ac74739f877566366f29fc3c36a`

### Summary

A row scrolling out of the model flyout was cut short of the card, because the
list is inset by the card's `p-1` and overflow clips at the scroll container's
own box rather than at the rounded edge. The list now bleeds back over that
inset with `-mt-1` / `-mb-1` and restates it as scroll padding, so rows appear
and disappear at the card's edge while the first and last rows still rest clear
of the border. The bottom bleeds only when nothing follows the list; with the
chat-settings row present its `border-t` is the boundary.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | Both flyout parents carry the same 4px inset — the panel card's `p-1` for the two-panel layout, `PopoverContent`'s `p-1` for the single-type `w-full` layout — so one bleed covers both |
| `REVIEW.md` | Distinguish introduced defects from unchanged code | Pass | The diff adds no border, token, or colour; the pre-existing raw ramp stop in `FLYOUT_PANEL_CLASS` is noted below, not charged to this PR |
| `REVIEW.md` | Require meaningful coverage for new behavior | Pass | Geometry is not assertable in happy-dom, which has no layout engine; asserting the class string is what `app-testing.md` tells tests not to do |
| `REVIEW.md` | Report practice violations with the rule and source | Pass | No violation in the diff |
| `REVIEW.md` | Persisted data, contracts, and cross-version compatibility | N/A | One component's class strings and one hoisted local; no schema, contract, or persisted state |
| `REVIEW.md` | New fallbacks declared in a `Fallbacks` PR-body section | N/A | The diff introduces no fallback |
| `docs/styles.md` | Compose Tailwind utilities; no CSS modules, `<style>`, runtime injection, or CSS-in-JS | Pass | Spacing utilities inside the existing `cn()` call |
| `docs/styles.md` | One hairline serves the product; components must not hand-write a border width | Pass | No width is declared here. The measured 5px band is `4px` padding plus the border's rounded-up 1px used value, which is layout, not ink |
| `docs/styles.md` | `border-border` for a border, `bg-divider` for a rule; never a raw ramp stop | Pass (pre-existing note) | The diff adds neither. `FLYOUT_PANEL_CLASS`'s `border-[hsl(var(--gray-400))]` predates it — see observations |
| `docs/styles.md` | No component-local CSS variable acting as an alternate token registry | Pass | No custom property added |
| `docs/styles.md` | Tests locate elements by role, accessible name, or `data-*`, never by styling classes | Pass | No test queries these classes |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | Neither appears in the diff |
| `docs/bad-smell.md` | No fabricated defaults hiding a broken invariant | Pass | `showSettingsRow` carries the same operands as the expression it replaces |
| `docs/effect.md` | Render stays pure — no store writes, timers, or subscriptions during render | Pass | The hoisted value is a pure predicate over existing reads |
| `docs/testing.md` | Cover new behavior where it adds confidence beyond existing checks | Pass | `chat-composer-media-models.test.tsx` and `chat-composer-model-selection.test.tsx` still exercise this markup |
| `docs/testing/app-testing.md` | Do not add tests that restate static configuration | Pass | A test pinning `max-h-[252px]` would freeze the implementation without proving the edges meet |
| `docs/testing/testing-external-behavior.md` | Assert through the page, never internals | Pass | No test changes |
| `docs/fallback.md` | A non-GA switch needs no compatibility code | Pass | Rides inside the non-GA `modelPickerFlyout` switch |

### Findings

No P0 or P1 findings.

Independent checks of the three claims this change rests on:

- **The classes resolve.** Building the App's real Tailwind entry
  (`apps/platform/src/views/css/index.css`, v4.3.3) from this HEAD emits
  `max-height: 248px`, `max-height: 252px`, and the `-mt-1` / `-mb-1` spacing
  utilities. Arbitrary values silently vanish when a file is outside the scanned
  tree, so this is worth confirming rather than assuming.
- **Nothing moves at rest.** With no settings row the list contributes
  `-4 + 252 - 4 = 244px` of layout, exactly what the previous `max-h-[244px]`
  contributed, and `pt-1` / `pb-1` put the first and last rows where the card's
  padding used to put them. With the settings row, `-4 + 248 = 244px`. The
  card's height and the scrollable area are unchanged in both branches.
- **The predicate is unchanged.** `showSettingsRow` keeps the same operands and
  the same short-circuit order as the inline expression it replaces, so
  `canAdjustChatSettings` is still never called with a null selection. The added
  `&& selectedOption` in the JSX is narrowing for TypeScript, not a new runtime
  condition.

Non-blocking observations:

1. The list now encodes its parent's padding (`-mt-1` against the card's `p-1`).
   If that padding ever changes, the bleed silently stops matching the edge. The
   comment names the coupling, which is the right mitigation at one call site,
   but it is a coupling.
2. `max-h-[248px]` and `max-h-[252px]` are `244px` plus each branch's own
   padding. The comment states the invariant — visible area stays 244px — but
   not the arithmetic per branch, so a future edit to one number has to
   rediscover it.
3. `showSettingsRow` is computed before the `page.kind === "settings"` early
   return, so it is evaluated and discarded on that path. The predicate is pure
   and cheap; noted only for completeness.
4. Pre-existing and out of scope: `FLYOUT_PANEL_CLASS` still uses
   `border-[hsl(var(--gray-400))]`. The practice revision reviewed here added
   the rule that a border must come from `border-border` and a painted rule from
   `bg-divider`, "not a raw ramp stop such as `border-gray-200`". The adjacent
   settings row's `border-t border-border/60` is a separator drawn as a border
   rather than `bg-divider`. Both predate this diff and belong in a styling
   migration, not here.

### Testing

No test changes, correctly. happy-dom has no layout engine, so the clip edge
cannot be asserted, and `docs/testing/app-testing.md` directs tests away from
querying styling classes. The behavioural surface around this markup stays
covered: `chat-composer-media-models.test.tsx` (23 passed) drives the flyout's
two-panel list, and `chat-composer-model-selection.test.tsx` (21 passed) covers
the model-selection paths. `pnpm lint:style` exits 0 — its
`no-unknown-classes` layer validates these class strings against the real App
Tailwind entry — alongside `check-types`, `oxlint`, and `prettier --check`.

Material gaps, both pre-existing:

- Nothing in CI can catch a pixel regression here. The evidence path for this
  class of change is the box-model derivation plus measurement in a real
  browser, which the PR body supplies (5px band -> 1px, card height 254px
  unchanged, at 2x in Chromium against the compiled App CSS).
- The flyout's own chat-settings row has no direct test; the `Chat settings`
  assertions in `chat-composer-model-selection.test.tsx` drive the drill-in
  menu's overview path instead. That matters slightly more after this change,
  because the list's geometry now branches on the same predicate. The predicate
  is copied verbatim, so this PR does not move the risk, but it is the test
  worth adding next time that row changes.

### CI

`test-app` shards 1, 2, 4, 6, and 7 failed on this HEAD. The single real failure
is `welcome-thread-sync-overlap.test.tsx:190` ("Unable to find an element with
the text: Welcome history in the second pane"); the other four shards report
"The strategy configuration was canceled because test-app._4 failed".

That test is unrelated to this diff: it was added by #33306 (`4fb323c9`),
contains no reference to the model picker, and exercises welcome-thread
synchronization ordering across panes. All eight `test-app` shards passed on
`main` at `266e707f`, which contains `4fb323c9`. Treating it as a flake and
rerunning only the failed jobs.

PR #33520, opened 2026-09-11
fix(api): report built-in capacity failures as errors

### LINKED ISSUE #33360: bug(api): report built-in model-capacity failures as errors
## Implementation authorized — 2026-09-11

Lancy explicitly approved dispatch of the ready warning-removal and genuine-failure fixes using `/handoff-issue-to-thread`, and requested reuse of each original research thread. This is now an implementation task, superseding earlier research-only/no-PR stopping instructions. The accepted technical findings and evidence limits remain applicable, subject to the scope below.

Sole owner: https://app.okou.ai/chats/2875fc8a-67d0-496f-b8f5-dfc1d3a6149f (Claude Opus 5, verified). Controller: https://app.okou.ai/chats/922b01fa-5dbb-43e9-97ee-fca3fb8e44cd.

Controller refreshed current main to `70cae141724e41ac67db20af9826210470311d5e` and reread the affected services and classifier. The relevant behavior is unchanged from the accepted research; no fresh production investigation is required. A complete 24-open-PR inventory, including all files for the unchanged 108-file #33450 HEAD, has no implementation of these target paths at dispatch. Recheck live current main/callers and overlap before coding.

### Authorized scope: one truthful ERROR for a failed built-in capacity run

The accepted logging proposal is implementation-ready. At `agent-webhook-complete.service.ts:209-243,246-300`, the generic terminal boundary preserves a built-in provider_overloaded failure but `logRunFailure` selects WARN. Change the existing level selection to ERROR only when the server-stored provider resolves to built-in and the normalized terminal reason is provider_overloaded. Preserve the existing insufficient-credit DEBUG case, other reasons/providers/suppression rules, message/fields, failed state and user-facing copy. This token is framework-independent, so built-in Claude overload reaching this same boundary is covered as well. Do not emit an additional terminal event.

Validation: the actual completion boundary emits one ERROR and no duplicate WARN for this combination; cover built-in Codex and the same supported reason for Claude, ordinary failures, BYOK, invalid/unknown providers, credit behavior, duplicate and concurrent callbacks, unchanged run result and unchanged user text. Respect existing terminal ownership rather than claiming exactly-once durable logging.

**Explicitly excluded pending product/design discussion:** reporting from API terminal completion into shared candidate cooldown, automatically switching routing or retrying the failed user task, WebSocket parsing changes, accepting new reporter trust/provenance, freshness semantics and durable/deduplicated dual-source delivery. These affect shared routing and their remaining trust/delivery questions are not resolved by a severity fix. Record that follow-up clearly in the PR/result; it is not a blocker for this approved logging PR. This issue's current desired outcome is the truthful ERROR boundary; never claim its merge fixed upstream capacity or automatic failover.

### Delivery, validation and authority

Reuse your accepted research; refresh the relevant current code rather than repeat the investigation. Fetch and rebase origin/main, implement one focused PR, run targeted verification and self-review, and immediately invoke `/pr-auto <explicit PR URL>` in the SAME owner thread. Complete a fresh independent full-SHA tracked review, required CI and normal protected merge queue with HEAD guard until GitHub reports MERGED or a genuine blocker. No direct/admin/bypass merge; no duplicate owner/PR. Preserve other contributors' changes; overlap is inventory, not an ordering blocker. A runtime limit leaves a precise resumable handoff in the same thread.

No full local Vitest suite or dev server. Report missing local database/compilation/test-body limitations honestly and rely on the actual PR CI for those checks. English repository/issue/PR artifacts and lowercase conventional commit subjects. Only pause for an actual zero-job action_required / Approve workflows gate when this owned PR modifies Actions definitions; report the exact approval link and do not try reruns/empty commits/branch changes around absent approval.

These changes require no database migration/backfill, provider model change, production-data mutation or new monitoring access. Logging removal intentionally changes operator-event counts; update the affected documentation/tests, but do not claim external monitors were independently checked or add a permission prerequisite for ordinary implementation. Keep all diagnostics content-free and preserve real failures.

The owner stops after reporting GitHub-confirmed merged SHA, current-main inclusion, review/check links and verification limitations. The controller independently accepts the merged code and retires the disposable watcher. No implementation owner may invoke /release-production, merge release PRs, approve production deployments or perform production changes. Production release requires actual authorization and a separate release-only executor after exact candidate inclusion and controller acceptance.

The complete batch consists of #33354, #33360 (logging only), #33361, #33362 and #33377. Each has its own reused owner and completed/failed/cancelled watcher. Keep remaining shared routing policy for #33360 explicitly unimplemented and pending discussion; do not silently turn it into a blocker for the ready fixes.

<details>
<summary>Original incident report and research context (history; current authorized scope is above)</summary>

## Summary

A built-in gpt-5.6-sol run terminated with provider_overloaded. This is a real main-run failure: the API terminal record should be ERROR, preserving the truthful failed state. Optional candidate-cooldown coverage is a separate proposal for discussion; its absence does not identify the original proxy failure stage.

Related context: #18048, #33180. Closed issues remain closed; this review does not override prior implementation/release authority.

## Impact

One run failed with the explicit selected-model-capacity message. Only its API warning is present in the selected error/warn scope; absence of a runner error does not imply missing terminal processing.

Severity: **medium**. Type: **provider capacity rejection and API severity classification**. Severity is triage, not a raw-event-to-user equivalence.

### Coverage

- Beijing: **[2026-09-10 00:00, 2026-09-11 00:00) Asia/Shanghai**. Exact UTC: **[2026-09-09T16:00:00Z, 2026-09-10T16:00:00Z)**.
- Axiom: ONLY `vm0-web-logs-prod`, `level == "error" or level == "warn"`; **1 records**, first `2026-09-10T08:39:08.992Z`; last `2026-09-10T08:39:08.992Z`. Levels: warn: 1.
- Distinct logged run IDs: **1**. 0 Axiom records lack a run ID; these do not imply additional runs.
- Per-run record distribution: run-150 each has {"Run failed":1}.
- Privacy: these are selected original source fields. Private IDs, device tokens, bound SQL parameters, custom workflow paths, signed/private URLs and unrelated user content are redacted or omitted. Only allowlisted method/status/route/timestamp network breadcrumbs are retained when directly relevant; no request bodies, general breadcrumbs, raw user identities or private Sentry links are published.

### Variant 1 — Run failed (1 records)

- Source: Axiom; first `2026-09-10T08:39:08.992Z`; last `2026-09-10T08:39:08.992Z`.
- Native causal fields: `fields.error`, `fields.exitCode`, `fields.failureReason`.

Original sanitized sample 1:

```json
{
  "_time": "2026-09-10T08:39:08.992Z",
  "fields.context": "webhook:complete",
  "fields.error": "Selected model is at capacity. Please try a different model.",
  "fields.exitCode": 1,
  "fields.failureReason": "provider_overloaded",
  "fields.runId": "<redacted:run-150>",
  "level": "warn",
  "message": "Run failed",
  "source": "api",
}
```

## Current code analysis

Fetched origin/main: **`75095c26e72eb7deb40f5c15b920aa0764d94ceb`**. This is read-only source inspection, not proof of which revision each event executed.

- [turbo/apps/api/src/signals/services/agent-webhook-complete.service.ts](https://github.com/vm0-ai/vm0/blob/75095c26e72eb7deb40f5c15b920aa0764d94ceb/turbo/apps/api/src/signals/services/agent-webhook-complete.service.ts)
- [crates/runner/src/cmd/start/job_terminal_log.rs](https://github.com/vm0-ai/vm0/blob/75095c26e72eb7deb40f5c15b920aa0764d94ceb/crates/runner/src/cmd/start/job_terminal_log.rs)
- [crates/guest-agent/src/cli/mod.rs](https://github.com/vm0-ai/vm0/blob/75095c26e72eb7deb40f5c15b920aa0764d94ceb/crates/guest-agent/src/cli/mod.rs)

The original native record has failureReason=provider_overloaded and exitCode=1; it does not itself contain route ownership. Subsequent independent research and the controller's bounded MaskDB check confirmed model_provider=built-in, selected_model=gpt-5.6-sol, and status=failed. Current source at bea6d926ba8a3edb263e8ae9440299c231752636 still retains WARN at this API boundary. Closed #18048 addressed Runner severity; it is not reopened. A recorded WebSocket upgrade and absent observed cooldown advancement do not identify the exact missing reporter stage. See the controller research-acceptance comment for the complete proposal and evidence limits.

## Desired outcome

For the existing API terminal record, classify built-in + provider_overloaded as ERROR. Keep the failed run, user guidance and unrelated failure/suppression behavior unchanged; emit no duplicate terminal record. This is a real failed main task, so deleting or downgrading its only observed warning would be incorrect.

Exercise the actual completion/logging boundary: the built-in capacity case produces one ERROR and no WARN duplicate; BYOK, non-capacity and unsupported-provider controls preserve existing behavior; duplicate/concurrent completion does not add another terminal record. Preserve content privacy.

An API-terminal hint to the candidate-cooldown service is a separate behavior proposal, not an already-proven WebSocket fix or recovery for the failed run. Before selecting it for implementation, resolve authenticated evidence ownership, shared-route impact, timing semantics, failure isolation and duplicate/delivery guarantees. A server-assigned completed_at is not the upstream failure time. Preserve the existing failed run and assess future candidate selection through the real route harness.

Research is complete with explicit evidence limits. The issue remains OPEN pending discussion and the selected implementation; research completion does not authorize a PR, release, production replay or configuration change. No natural-traffic threshold or new production permission is a prerequisite to finishing research. Logging improvement alone is not a claim that upstream capacity or reporter coverage has been repaired.

## Ownership

Selected owner: @lancy. @lancy owns model execution and the directly adjacent typed provider-rejection classifiers, including #33180. Confidence: medium.

</details>

lancy:
## Problem

Fixes #33360.

A run that terminates with `provider_overloaded` on a **built-in** route is a genuine failed main task: the user's turn is lost, and the rejected credential is one the platform owns. The API terminal boundary already declines to suppress that record — `shouldSuppressKnownFailureLog` suppresses the provider reasons only for non-built-in providers — but `logRunFailure` then selected `WARN` for everything except the insufficient-credit case.

The reviewed production incident (2026-09-10, built-in `gpt-5.6-sol`, web-triggered, failed after ~6.9s) produced exactly one Axiom record for the run, at `warn`. The Runner classifies this condition at `info`, and Runner `info` is not shipped, so this API record is the only production signal for the failure.

## Change

`turbo/apps/api/src/signals/services/agent-webhook-complete.service.ts`

- Add `isBuiltInCapacityFailure`: the normalized terminal reason is `provider_overloaded` **and** the server-stored `modelProvider` is built-in. It reads `commit.run.modelProvider`, the same stored value the suppression decision already uses, so the two cannot disagree.
- Extend the existing level selection in `logRunFailure` to pick `L.error` for that combination.

Nothing else moves. The insufficient-credit `DEBUG` record, every other reason, BYOK/legacy/null providers, the suppression matrix, the message text, the logged fields, the failed run state and the user-facing copy are unchanged, and no additional terminal record is emitted. The reason token carries no framework, so a built-in Claude overload arriving at this same boundary is covered.

## Scope

This is the severity fix only. It does **not** repair upstream capacity, and it does **not** claim the reporter gap is closed.

Explicitly out of scope and still pending product/design discussion, per the issue's authorized scope:

- reporting from API terminal completion into the shared candidate cooldown,
- automatically re-routing or retrying the failed user task,
- WebSocket parsing changes,
- accepting a new reporter trust/provenance path, freshness semantics, or durable/deduplicated dual-source delivery.

Those change shared routing for other users and their trust and delivery questions are not resolved by a severity change.

## Tests

All in the existing `RUN-03 > completion failure logs` matrix, driven through the real completion webhook and the real logger mocks.

- `reports capacity exhaustion on a built-in run as an error` — a genuine built-in run created through the production run-create endpoint. It asserts exactly one `ERROR "Run failed"` with the unchanged fields, and `genericFailureLogCalls` (which scans `debug`/`info`/`warn`/`error`) of length 1, so there is no WARN duplicate and no extra terminal record. The predicate reads only the terminal reason and the stored provider, which are identical for every built-in route, so the framework-independence claim stays a documented code comment rather than a second case that would restate the same two inputs.
- `keeps capacity exhaustion on ... at warning severity` — null provider and legacy provider stay at `WARN`. Only a stored built-in provider escalates.
- `keeps built-in credit exhaustion on its own debug record` — built-in `insufficient_credits` still logs `Run stopped: insufficient credits` at `DEBUG` and produces no `Run failed` record at any level, so the credit branch still wins.
- `keeps one error when a duplicate repeats the capacity failure` — a duplicate completion adds no second record and leaves the run status, error text and persisted failure reason unchanged.
- `records one error when completions race the capacity failure` — two completions are held behind `holdAgentRunRowLockFixture` and released together; exactly one `ERROR` is produced and the terminal state is unchanged.

Existing controls are unchanged and still pass: BYOK `provider_overloaded` stays fully suppressed, built-in `provider_rate_limited` stays at `WARN`, and the globally suppressed reasons stay silent for built-in.

## Verification

Run locally against a local PostgreSQL 18 instance with the repository migrations applied:

- `vitest run src/signals/routes/__tests__/run-lifecycle.bdd.test.ts -t "completion failure logs"` — **38 passed**, 220 skipped on final source HEAD `edad62c6e69e90c26e30c68f647d3bc203765df6` (implementation-owner report; not rerun locally by the controller).
- Negative control was performed on the preceding HEAD: reverting the production change made the capacity cases fail (`expected [] to have a length of 1`). It was not rerun on final source HEAD `edad62c6e69e90c26e30c68f647d3bc203765df6`; the separate-context review verified binding at that HEAD, and the final merge-group CI provides the merged-code execution evidence.
- `check-types:gateways`, `check-types:core`, `check-types:tests` — clean.
- `eslint` and `oxlint --deny-warnings` on both changed files — clean.
- `prettier --check` on both changed files — clean.

Limitations, honestly reported:

- No full local Vitest suite and no dev server were run.
- `chat-callbacks.bdd.test.ts -t "CHAT-02: failed chat callbacks"` fails 4 tests in this sandbox with `Matcher did not succeed in time` inside `claimChatRunJob`. The identical 4 failures reproduce on unmodified `main` in the same environment, so they are a pre-existing local runner-claim timing limitation, not a regression from this change. CI covers that suite.
- No Rust, addon, or frontend surface is touched.

## Review

The first tracked review on this PR was a self-review by the PR author and did not satisfy the issue's independent-review requirement. An independent reviewer then reviewed `692f5f88` and returned `Changes Requested` with one P1: the second escalation case rewrote `agent_runs.model_provider` to `built-in` on a BYOK run through an internal test-only write, producing a hybrid row production cannot create. `edad62c6` resolves it by dropping that case, and a fresh independent review of `edad62c6` returned `LGTM` with no P0/P1 findings.

## Deployment compatibility

API-only log-severity change. No contract, Runner protocol, persisted shape, migration, backfill, provider model or production configuration change. Operator error/warn event counts for this one combination intentionally shift from `warn` to `error`; no in-repo documentation references that record. External monitors were not independently inspected.

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
@@ -19698,6 +19698,169 @@ describe("RUN-03: sandbox completion reports against missing checkpoints and set
       },
     );
 
+    it.each([
+      {
+        name: "a built-in Codex run",
+        modelProvider: "built-in",
+        failureReason: "provider_overloaded",
+      },
+      // The reason token carries no framework, so a Claude run stored against
+      // the built-in provider reaches the same terminal record.
+      {
+        name: "a built-in Claude run",
+        modelProvider: "anthropic-api-key",
+        persistedModelProvider: "built-in",
+        failureReason: "provider_overloaded",
+      },
+    ] satisfies readonly (FailureCase & { readonly name: string })[])(
+      "reports capacity exhaustion on $name as an error",
+      async (failure) => {
+        const control = await completeFailure(failure);
+        const errors = matchingLogCalls(
+          context.mocks.axiomLogging.error,
+          "Run failed",
+          control.runId,
+        );
+        expect(errors).toHaveLength(1);
+        expect(errors[0]?.[1]).toStrictEqual(
+          expect.objectContaining({
+            runId: control.runId,
+            exitCode: 1,
+            error: control.error,
+            failureReason: "provider_overloaded",
+            context: "webhook:complete",
+          }),
+        );
+        expect(genericFailureLogCalls(control.runId)).toHaveLength(1);
+      },
+    );
+
+    it.each([
+      { name: "a null provider", persistedModelProvider: null },
+      {
+        name: "a legacy provider",
+        persistedModelProvider: "legacy-unknown-provider",
+      },
+    ] satisfies readonly (FailureCase & { readonly name: string })[])(
+      "keeps capacity exhaustion on $name at warning severity",
+      async (provider) => {
+        const control = await completeFailure({
+          ...provider,
+          failureReason: "provider_overloaded",
+        });
+        expect(
+          matchingLogCalls(
+            context.mocks.axiomLogging.warn,
+            "Run failed",
+            control.runId,
+          ),
+        ).toHaveLength(1);
+        expect(genericFailureLogCalls(control.runId)).toHaveLength(1);
+      },
+    );
+
+    it("keeps built-in credit exhaustion on its own debug record", async () => {
+      const control = await completeFailure({
+        modelProvider: "built-in",
+        failureReason: "insufficient_credits",
+      });
+      expect(
+        matchingLogCalls(
+          context.mocks.axiomLogging.debug,
+          "Run stopped: insufficient credits",
+          control.runId,
+        ),
+      ).toHaveLength(1);
+      expect(genericFailureLogCalls(control.runId)).toHaveLength(0);
+    });
+
+    it("keeps one error when a duplicate repeats the capacity failure", async () => {
+      const api = createRunsApi(context);
+      const webhooks = createWebhookCallbackApi(context);
+      const first = await completeFailure({
+        modelProvider: "built-in",
+        failureReason: "provider_overloaded",
+      });
+      expect(genericFailureLogCalls(first.runId)).toHaveLength(1);
+
+      await webhooks.requestAgentComplete(
+        {
+          runId: first.runId,
+          exitCode: 1,
+          error: "late built-in capacity report",
+          failureReason: "provider_overloaded",
+        },
+        {
+          authorization: `Bearer ${api.sandboxTokenForRun(
+            first.actor,
+            first.runId,
+          )}`,
+        },
+        [200],
+      );
+
+      expect(genericFailureLogCalls(first.runId)).toHaveLength(1);
+      await expect(
+        api.readRun(first.actor, first.runId),
+      ).resolves.toMatchObject({ status: "failed", error: first.error });
+      await expect(
+        readRunFailureReasonFixture(context, first.runId),
+      ).resolves.toBe("provider_overloaded");
+    });
+
+    it("records one error when completions race the capacity failure", async () => {
+      const api = createRunsApi(context);
+      const webhooks = createWebhookCallbackApi(context);
+      await seedBuiltInDefaultModelKey();
+      const { actor, agentId } = await entitledRunActor();
+      const run = await api.createRun(actor, {
+        agentId,
+        prompt: "race a built-in capacity completion",
+        modelProvider: "built-in",
+      });
+      const error = `racing capacity failure for ${run.runId}`;
+      const body = {
+        runId: run.runId,
+        exitCode: 1,
+        error,
+        failureReason: "provider_overloaded",
+      } as const;
+      const headers = {
+        authorization: `Bearer ${api.sandboxTokenForRun(actor, run.runId)}`,
+      };
+      const lifecycleGate = await holdAgentRunRowLockFixture({
+        runId: run.runId,
+        signal: context.signal,
+      });
+      const completions = Promise.all([
+        webhooks.requestAgentComplete(body, headers, [200]),
+        webhooks.requestAgentComplete(body, headers, [200]),
+      ]);
+      onTestFinished(async () => {
+        lifecycleGate.release();
+        await Promise.allSettled([completions, lifecycleGate.done]);
+      });
+      await expect.poll(lifecycleGate.waiterCount).toBe(2);
+      lifecycleGate.release();
+      await Promise.all([lifecycleGate.done, completions]);
+
+      expect(genericFailureLogCalls(run.runId)).toHaveLength(1);
+      expect(
+        matchingLogCalls(
+          context.mocks.axiomLogging.error,
+          "Run failed",
+          run.runId,
+        ),
+      ).toHaveLength(1);
+      await expect(api.readRun(actor, run.runId)).resolves.toMatchObject({
+        status: "failed",
+        error,
+      });
+      await expect(
+        readRunFailureReasonFixture(context, run.runId),
+      ).resolves.toBe("provider_overloaded");
+    });
+
     it("keeps the missing-checkpoint warning visible for a suppressible reason", async () => {
       const api = createRunsApi(context);
       const webhooks = createWebhookCallbackApi(context);
--- a/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
@@ -19698,42 +19698,31 @@ describe("RUN-03: sandbox completion reports against missing checkpoints and set
       },
     );
 
-    it.each([
-      {
-        name: "a built-in Codex run",
+    // The predicate reads only the terminal reason and the stored provider, so
+    // every built-in route reaches this record with the same two inputs no
+    // matter which framework produced the reason token.
+    it("reports capacity exhaustion on a built-in run as an error", async () => {
+      const control = await completeFailure({
         modelProvider: "built-in",
         failureReason: "provider_overloaded",
-      },
-      // The reason token carries no framework, so a Claude run stored against
-      // the built-in provider reaches the same terminal record.
-      {
-        name: "a built-in Claude run",
-        modelProvider: "anthropic-api-key",
-        persistedModelProvider: "built-in",
-        failureReason: "provider_overloaded",
-      },
-    ] satisfies readonly (FailureCase & { readonly name: string })[])(
-      "reports capacity exhaustion on $name as an error",
-      async (failure) => {
-        const control = await completeFailure(failure);
-        const errors = matchingLogCalls(
-          context.mocks.axiomLogging.error,
-          "Run failed",
-          control.runId,
-        );
-        expect(errors).toHaveLength(1);
-        expect(errors[0]?.[1]).toStrictEqual(
-          expect.objectContaining({
-            runId: control.runId,
-            exitCode: 1,
-            error: control.error,
-            failureReason: "provider_overloaded",
-            context: "webhook:complete",
-          }),
-        );
-        expect(genericFailureLogCalls(control.runId)).toHaveLength(1);
-      },
-    );
+      });
+      const errors = matchingLogCalls(
+        context.mocks.axiomLogging.error,
+        "Run failed",
+        control.runId,
+      );
+      expect(errors).toHaveLength(1);
+      expect(errors[0]?.[1]).toStrictEqual(
+        expect.objectContaining({
+          runId: control.runId,
+          exitCode: 1,
+          error: control.error,
+          failureReason: "provider_overloaded",
+          context: "webhook:complete",
+        }),
+      );
+      expect(genericFailureLogCalls(control.runId)).toHaveLength(1);
+    });
 
     it.each([
       { name: "a null provider", persistedModelProvider: null },
@@ -19800,6 +19789,13 @@ describe("RUN-03: sandbox completion reports against missing checkpoints and set
       );
 
       expect(genericFailureLogCalls(first.runId)).toHaveLength(1);
+      expect(
+        matchingLogCalls(
+          context.mocks.axiomLogging.error,
+          "Run failed",
+          first.runId,
+        ),
+      ).toHaveLength(1);
       await expect(
         api.readRun(first.actor, first.runId),
       ).resolves.toMatchObject({ status: "failed", error: first.error });
--- a/turbo/apps/api/src/signals/services/agent-webhook-complete.service.ts
+++ b/turbo/apps/api/src/signals/services/agent-webhook-complete.service.ts
@@ -282,13 +282,32 @@ function shouldSuppressFailureLog(
   return shouldSuppressKnownFailureLog(run, knownFailureReason.data);
 }
 
+/**
+ * A capacity rejection on a built-in route fails the user's run against a
+ * credential the platform owns, so it stays a genuine operator-actionable
+ * failure rather than a caller-side condition. The reason token is
+ * framework-independent, so a built-in Claude overload that reaches this same
+ * boundary is included. Routing recovery is a separate decision; this only
+ * classifies the severity of the single terminal record.
+ */
+function isBuiltInCapacityFailure(commit: CompletionCommit): boolean {
+  return (
+    commit.transitionFailureReason === "provider_overloaded" &&
+    isBuiltInModelProviderType(commit.run.modelProvider)
+  );
+}
+
 function logRunFailure(
   input: CompleteAgentRunInput,
   commit: CompletionCommit,
 ): void {
   const isCreditError =
     commit.transitionFailureReason === "insufficient_credits";
-  const logFailure = isCreditError ? L.debug : L.warn;
+  const logFailure = isCreditError
+    ? L.debug
+    : isBuiltInCapacityFailure(commit)
+      ? L.error
+      : L.warn;
   logFailure(
     isCreditError ? "Run stopped: insufficient credits" : "Run failed",
     {


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: 692f5f88981175b1300d33e835694e58c461ff77
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `692f5f88981175b1300d33e835694e58c461ff77`. Practice documents read from `main` at `b4762ad9fc399a1a9dbde812f63b0eb7252dd05b`.

### Summary

Escalates the single terminal completion record from `WARN` to `ERROR` for exactly one combination — the stored model provider is built-in and the normalized terminal reason is `provider_overloaded` — because that run is a genuine failed main task against a credential the platform owns. The message, fields, suppression matrix, credit `DEBUG` branch, failed run state and user-facing copy are untouched, and no second terminal record is emitted.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | Traced `POST /api/webhooks/agent/complete` → `completeAgentRun$` → `logAgentRunCompletionOutcome` → `shouldSuppressFailureLog` → `logRunFailure`. The new predicate reads `commit.run.modelProvider`, the same stored value the suppression decision uses, so the two branches cannot disagree: `ERROR` ⊂ not-suppressed, since suppression only fires when the provider parses and is *not* built-in. `logRunFailure` has exactly one caller. |
| `REVIEW.md` | Check ordering, races, cancellation and failure outcomes where relevant | Pass | The `missing-checkpoint` branch still returns before `logRunFailure`, and the Pi branch is unreachable for this case (`isGptApiKeyPiProviderType` keys are `openai-api-key`/`openrouter-codex`, never `built-in`). Duplicate completions do not re-enter the `commit.transitioned` block, and a genuine concurrent race is covered by test, not asserted from reasoning. |
| `REVIEW.md` | Meaningful coverage for new behavior, identifying the missing scenario | Pass | Five new cases at the real completion boundary: both framework variants at `ERROR`, null/legacy providers still `WARN`, built-in credit still `DEBUG`, duplicate, and concurrent. See Testing for the negative control proving the tests bind to this change. |
| `REVIEW.md` | Regression coverage and containment for a changed existing flow | Pass | The pre-existing controls in the same matrix are unchanged and still pass: BYOK `provider_overloaded` fully suppressed, built-in `provider_rate_limited` at `WARN`, the globally suppressed reasons silent for built-in, and the missing-checkpoint warning intact. No feature/permission containment applies to an operator log severity. |
| `REVIEW.md` | Report concrete practice violations with rule and source, not inferred | Pass | No P0/P1 violation found; the two observations below name their file and rationale. |
| `REVIEW.md` | Persisted data, API contracts and response validation reviewed together | N/A | No schema, contract, `jsonb(...).$type<...>()`, response shape or persisted field changed. The diff only selects a log function. |
| `docs/bad-smell.md` | §1 No `any`; preserve inference | Pass | No `any`; `isBuiltInCapacityFailure` takes the existing `CompletionCommit` type and returns `boolean`. |
| `docs/bad-smell.md` | §2 No lint/type suppression comments, no weakened rules | Pass | None added. `eslint` and `oxlint --deny-warnings` are clean on both changed files. |
| `docs/bad-smell.md` | §3 Error handling: no redundant wrappers or fabricated defaults hiding a broken invariant | Pass | Nothing is caught, defaulted or swallowed. The change increases the truthfulness of an existing failure record rather than hiding one; the failed state and error text reaching the user are unchanged. |
| `docs/bad-smell.md` | §4 Interface changes documented with consumers assessed | N/A | No public contract changed. The reason token `provider_overloaded` and the built-in provider value already exist in `run-failure-reasons.ts` and `model-provider-types.ts`; only the severity of one log line moved. |
| `docs/bad-smell.md` | §5 Static imports in production code | Pass | No import added; `isBuiltInModelProviderType` was already imported by this file for the suppression branch. |
| `docs/bad-smell.md` | §6 Centralized `env()`, no hardcoded URLs or credentials | N/A | No configuration, URL or credential touched. |
| `docs/bad-smell.md` | §7 / `docs/fallback.md` | No new fallback introduced | Pass | The predicate is a classifier for a currently-reachable production state, not a tolerance path for a state the contract rules out, so `docs/fallback.md` requires no `Fallbacks` PR-body section. Nothing was removed, so no removal evidence is owed. |
| `docs/fallback.md` | §1 Do not add negative tests that only pin removed behavior | Pass | The null-provider, legacy-provider and credit cases pin **live** current behavior that the new branch must not capture. They are boundary controls for the new condition, not tombstones for deleted code. |
| `docs/testing.md` | Test the contract through the production entry point | Pass | Every new case drives the real `POST /api/webhooks/agent/complete` with a sandbox token and verifies the run through `GET` on the runs API. No route handler or service is imported or called directly. |
| `docs/testing.md` | Mock external services at their boundary; use real internal code and the real database | Pass | The only mock asserted is `context.mocks.axiomLogging`, the centralized boundary mock for the external logging sink. No `vi.mock()` of an internal module was added. |
| `docs/testing.md` | Avoid duplicate cases that restate configuration or pin incidental implementation | Pass | Each case covers a distinct branch outcome (error / warn / debug / duplicate / concurrent). BYOK `provider_overloaded` was deliberately **not** duplicated: the existing `suppresses %s for a BYOK provider` case already scans all four levels through `genericFailureLogCalls`. |
| `docs/testing.md` | Avoid fake timers; wait for the observable result | Pass | No timer mocking added. The race case waits on `expect.poll(lifecycleGate.waiterCount)`, an observed database-lock state, not elapsed time. |
| `docs/testing.md` | Global teardown owns detached work | Pass | No `clearAllDetached()` call added. The race case registers `onTestFinished` only to release its own fixture lock and settle its own owned requests. |
| `docs/testing.md` | Cover old/new combinations for request, protocol, queue-payload or persisted-state changes | N/A | None of those changed; there is no cross-version combination to cover. |
| `docs/testing/api-testing.md` | Route behavior covered by route integration tests under `routes/__tests__/` | Pass | All cases live in the existing `run-lifecycle.bdd.test.ts` `RUN-03 > completion failure logs` matrix beside the behavior they guard. |
| `docs/testing/api-testing.md` | Only mock external services; no `vi.mock()` of internal modules | Pass | Covered above; the diff adds no mock registration at all. |
| `docs/testing/api-testing.md` | Shared persistent state must be uniquely owned and correct under concurrency | Pass | Every case creates its own actor, agent and run through the production APIs and asserts only on its own `runId` — `matchingLogCalls` filters log calls by `fields.runId`. The row lock in the race case is scoped to that test's own run id. No global lock, ordering or snapshot/restore is used. |
| `docs/testing/testing-external-behavior.md` | Construct cases and assert through the endpoint; DB and services are internal | Pass | Setup and verification use `createRun`, `requestAgentComplete` and `readRun`. No DB schema import, row write or row read for assertion was added to the test file. |
| `docs/testing/testing-external-behavior.md` | Leave the boundary only for states the production interface cannot construct, and say why | Pass | Two documented exceptions, both pre-existing in this file and used by neighbouring cases: `setRunModelProviderStateFixture`, a test-only route naming the owned run id, for a stored provider value the create API cannot select; and `holdAgentRunRowLockFixture`, an infrastructure-only ordering no production API can request, used to make the concurrent completion deterministic instead of timing-dependent. |

### Findings

No P0 or P1 findings.

Non-blocking observations:

- `turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts`: the "built-in Claude run" case reaches the built-in stored provider through `persistedModelProvider: "built-in"` on a run created with `anthropic-api-key`, rather than a natively resolved built-in Claude route. That is deliberate and disclosed in an inline comment — the harness's built-in default model resolves to the DeepSeek Codex route, and this boundary reads only the stored provider, so a natively resolved Claude built-in run would exercise the identical branch. The negative control below shows the case fails independently without the fix, so it is not a vacuous duplicate of the Codex case. A test that resolved the Claude route natively would need org-model-policy plumbing that this boundary never reads.
- `turbo/apps/api/src/signals/services/agent-webhook-complete.service.ts`: the level selection is now a nested ternary. Cyclomatic complexity stays far below the configured `complexity: ["error", { max: 20 }]`, no `no-nested-ternary` rule is configured, and the shape matches the existing local style, so this is a readability note only.

### Testing

Coverage added at the boundary the change affects:

- `reports capacity exhaustion on ... as an error` — a genuine built-in Codex run and a Claude-framework run stored as built-in. Each asserts exactly one `ERROR "Run failed"` carrying the unchanged `runId`, `exitCode`, `error`, `failureReason` and `context` fields, plus `genericFailureLogCalls` of length 1 across `debug`/`info`/`warn`/`error`, which proves both "no WARN duplicate" and "no extra terminal record".
- `keeps capacity exhaustion on ... at warning severity` — null and legacy stored providers stay at `WARN`. This is the control that stops the branch from widening past `built-in`, since those providers are also non-suppressed.
- `keeps built-in credit exhaustion on its own debug record` — built-in `insufficient_credits` still emits `Run stopped: insufficient credits` at `DEBUG` and no `Run failed` at any level, proving the credit branch still wins over the new one.
- `keeps one error when a duplicate repeats the capacity failure` — a duplicate completion adds no record and leaves status, error text and the persisted failure reason unchanged.
- `records one error when completions race the capacity failure` — two completions held behind `holdAgentRunRowLockFixture` and released together produce exactly one `ERROR` and an unchanged terminal state.

Verification actually run locally against a local PostgreSQL 18 instance with the repository migrations applied:

- `vitest run src/signals/routes/__tests__/run-lifecycle.bdd.test.ts -t "completion failure logs"` — 39 passed, 220 skipped.
- Negative control: with only the production ternary reverted, the two capacity cases fail with `expected [] to have a length of 1` and every other case still passes, so the new tests bind to this behavior and each framework variant fails independently.
- `check-types:gateways`, `check-types:core`, `check-types:tests` — clean.
- `eslint --max-warnings 0` and `oxlint --deny-warnings` on both changed files — clean. `prettier --check` on both — clean.

Material gaps and limitations, stated rather than implied:

- No full local Vitest suite and no dev server were run, per the repository's verification guidance.
- `chat-callbacks.bdd.test.ts -t "CHAT-02: failed chat callbacks"` fails 4 tests in this sandbox with `Matcher did not succeed in time` inside `claimChatRunJob`. The identical 4 failures reproduce on unmodified `main` in the same environment, so they are a pre-existing local runner-claim timing limitation rather than a regression; that suite is left to CI. Its `provider_overloaded` rounds do not assert log severity, and its one `axiomLogging.warn` assertion is guarded by `params.failureReason === undefined`, so it is unaffected by this change either way.
- Severity classification only. This PR does not repair upstream capacity, does not close the reporter gap, and adds no path from API terminal completion into the shared candidate cooldown; that routing proposal remains unimplemented and pending discussion, as recorded in the PR body.

lancy:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 692f5f88981175b1300d33e835694e58c461ff77

> **Independent review.** The earlier tracked comment on this PR was authored by the PR author and is a self-review, so it does not satisfy the issue's independent-review requirement. This review was produced by a separate reviewer that was given only the PR reference, was instructed not to read the existing review comment or the author's working tree, and reached its own verdict from the GitHub diff, `REVIEW.md` on `main`, and the routed practice documents. It supersedes the self-review, and its P1 contradicts a non-blocking observation the self-review recorded.


Inspected HEAD: `692f5f88981175b1300d33e835694e58c461ff77` (PR #33520, `fix/issue-33360-built-in-capacity-error`).
Practice revision: `9e8b388d344515944065846cc94ddba6ac2e64ab`.

### Summary

`logRunFailure` in `turbo/apps/api/src/signals/services/agent-webhook-complete.service.ts` now selects `L.error` instead of `L.warn` for the single terminal `"Run failed"` record when the normalized failure reason is `provider_overloaded` and the run's stored `modelProvider` is `built-in`; the `insufficient_credits` `DEBUG` branch, the suppression matrix, the logged fields, and the persisted terminal state are unchanged. The accompanying tests extend the existing `RUN-03 > completion failure logs` matrix with escalation, non-escalation, credit-branch, duplicate-completion and racing-completion cases. The production change is correct and well scoped; one new test case leaves the repository's API external-behavior boundary without a justified exception.

### Review Criteria

| --- | --- | --- | --- |
| REVIEW.md | Trace changed behavior through its real entry point and consumers | Pass | `logRunFailure` has exactly one caller, `logAgentRunCompletionOutcome` (service line 269), itself called once at line 1050 from the completion commit path. `"Run failed"` is emitted from no other production site in the repo (`grep -rn '"Run failed"'` finds only message-default helpers in `errors.ts`, `run-error-format.service.ts`, `internal-chat-run-callback.service.ts`, `threadless-run-cleanup.service.ts`, `agent-run-create.service.ts`). No second terminal record is added. |
| REVIEW.md | New condition must not disagree with an adjacent decision reading the same state | Pass | `isBuiltInCapacityFailure` reads `commit.run.modelProvider` (service line 296); `shouldSuppressKnownFailureLog` reads the same column through `modelProviderTypeSchema.safeParse(...)` then `isBuiltInModelProviderType(...)` (lines 234-238). `isBuiltInModelProviderType` is `type === "built-in"` over `string \| null \| undefined` (`model-provider-types.ts:22-26`) and `"built-in"` is a member of `MODEL_PROVIDER_TYPE_IDS`, so parse-then-check and direct compare are true on exactly the same input. `agent-run-metadata-write.service.ts:78` normalizes every built-in run's stored value to the literal `"built-in"`, so the column is the right state to read. |
| REVIEW.md | Other reachable paths / earlier returns | Pass | Earlier returns in `logAgentRunCompletionOutcome` are unaffected: `logGptApiKeyPiSandboxOutcome` only short-circuits when `isGptApiKeyPiProviderType(run.modelProvider)` is true, and `"built-in"` is not a key of `GPT_API_KEY_PI_ROUTES` (`pi-sandbox-config.ts:51-91`), so built-in runs still reach `logRunFailure`; the `missing-checkpoint` branch keeps its own message and is untouched. |
| REVIEW.md | Coverage is meaningful and binds to the new behavior | Pass | `reports capacity exhaustion on a built-in Codex run as an error` drives the real `POST` completion webhook through `createWebhookCallbackApi` on a run created with `modelProvider: "built-in"` and asserts one `error("Run failed")` plus `genericFailureLogCalls` of length 1 across `debug/info/warn/error`; before the diff that record was `warn`, so the case cannot pass vacuously. The `null`/legacy-provider cases pin the non-escalation boundary, and the `insufficient_credits` case pins the ternary ordering. |
| REVIEW.md | Tests verify externally observable state, not internals | Finding (P1) | The second escalation case fabricates `modelProvider = "built-in"` on an `anthropic-api-key` run through the internal test-only `set-run-model-provider` write (`run-lifecycle.bdd.test.ts:19709-19714`), although a real built-in Claude run is constructible through production endpoints. See Findings 1. |
| REVIEW.md | Persisted data, contracts, and deployment compatibility | N/A | The diff changes only the severity argument of one Axiom record. No schema, migration, ts-rest contract, runner protocol, queue payload, `jsonb().$type<>()` shape, or response body changes, so no old/new coexistence window exists. |
| REVIEW.md | Feature/capability/permission containment for a changed or rerouted flow | N/A | No user-facing flow is changed, replaced, or rerouted: the failed run state, `error` text, persisted `failure_reason`, and the completion HTTP responses are identical before and after (verified in `completeFailure` assertions and by reading the unchanged commit path). |
| REVIEW.md | ccstate ownership / async ownership | N/A | No new signal, command, effect, or detached work. The one new asynchronous test owns its lock through `holdAgentRunRowLockFixture({ signal: context.signal })` and releases it in both the body and `onTestFinished`. |
| docs/docs.md | Use the index to select the authoritative guides for the changed surface | Pass | Changed surfaces are API production logic and API route tests, so the index routed to `bad-smell.md`, `testing.md`, `testing/api-testing.md`, and `testing/testing-external-behavior.md`; `fallback.md` was also read to settle the fallback question. `deployment-compatibility.md`, `event-sourcing.md`, `effect.md`, `cache.md`, and the DB/feature-switch skills were deliberately not loaded — no matching surface is touched. |
| docs/bad-smell.md | Production-code quality rules (`any`, suppressions, error handling, interfaces, dynamic imports, configuration, fallback default) | Pass | The added helper is a typed pure predicate with no `any`, no suppression comment, no dynamic import, no hardcoded URL/credential, and no new `catch`. It changes no public contract (§4) and adds no tolerance branch for a state the contract rules out (§7). The added doc comment states the ownership rationale rather than restating the code. |
| docs/testing.md | Prefer integration coverage through real entry points; keep cases focused; avoid duplicates | Pass | All five new cases run through the production completion webhook and the production run-create endpoint with the centralized `context.mocks` logger, not through direct service calls. See Observation 3 for the one partially redundant assertion. |
| docs/testing/api-testing.md | Route tests build and verify through endpoints; only external services mocked; shared persistent state is uniquely owned | Finding (P1) | Mocks are limited to the centralized `context.mocks.axiomLogging` registry and no `vi.mock()` of internal modules is added. Shared-state ownership is sound: every case allocates a fresh actor/org through `entitledRunActor()`, `seedBuiltInDefaultModelKey()` seeds under a `randomUUID()` fixture id, log assertions filter by the unique `runId`, and the row lock is scoped to one owned run — no ordering, global lock, or residue tolerance. The boundary rule in the guide's "External Behavior Boundary" section is what Finding 1 breaks. |
| docs/testing/testing-external-behavior.md | Leave the boundary only when the state is impossible to construct through the production interface | Finding (P1) | "An existing helper can write the DB directly" is explicitly listed as *not* an exception, and the case carries no stated justification. The two pre-existing uses of the same fixture (`null`, `legacy-unknown-provider`) are genuinely unconstructible and remain justified; `"built-in"` is not. See Findings 1. |
| docs/fallback.md | Declare, justify, and bound any new fallback; no fallback for contract-prevented states | N/A | The diff introduces no fallback: it adds no tolerance for an old shape, no default for a missing value, no `??`/`\|\|` chain, and no retired-producer reader. It is a classification of an already-emitted record, so no `Fallbacks` section is required and none of §6-§9 applies. No tombstone test is added either — every new assertion pins live behavior. |

### Findings

**P1 — 1. The "built-in Claude run" case fabricates a run row through an internal write instead of the production built-in route**
`turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts:19709-19714` (helper at `:19594-19600`)

Trigger: the case creates a run with `modelProvider: "anthropic-api-key"` and then rewrites `agent_runs.model_provider` to `"built-in"` through the internal test-only `set-run-model-provider` action (`turbo/apps/api/src/signals/routes/test-runtime-state.ts:2543-2554`, a direct `UPDATE` on the run row).

Why this is a defect rather than a style point:

1. It is an unjustified boundary exception. `docs/testing/testing-external-behavior.md` allows leaving the API boundary only when a case is "completely impossible to construct through the production external interface", and explicitly rules out "an existing helper can write the DB directly". `docs/testing/api-testing.md` repeats it: "If a case is not constructible through the production API surface, do not add an API route test that reaches into internals." No justification is recorded at the case. The two pre-existing uses of this fixture (`null` and `legacy-unknown-provider`) are different — neither value can be produced by a live writer, because `normalizeRunModelMetadata` writes either a real provider type or `null` only for providerless runs.
2. A real built-in Claude run *is* constructible with the endpoints this very file already uses. `BUILT_IN_MODEL_TO_PROVIDER` contains `claude-sonnet-5`, `claude-opus-5`, `claude-fable-5-1`, etc. with `anthropic-api-key`/`openrouter-api-key` candidates (`turbo/packages/api-contracts/src/contracts/model-providers.ts:293-336`), and the same test file builds genuine built-in runs for a chosen model at `:8646-8700` and `:8770-8800` via `seedBuiltInModelKey(selectedModel)` + `api.updateOrgModelPolicies(actor, [{ model, isDefault: true, defaultProviderType: "built-in", credentialScope: "org", modelProviderId: null }])` and then a create/send call.
3. The fabricated row is a state production cannot produce, so the case does not demonstrate what the PR body claims it demonstrates. After the mutation the run has `modelProvider = "built-in"` while `selectedModel`, `modelRuntimeProvider`, `modelRuntimeModel`, and `builtInModelKeyId` still hold the BYOK run's values (`null`). A real built-in Claude run has a fully resolved built-in route; this hybrid row could never exist. The case therefore re-asserts the same single predicate the Codex case already covers, while adding coupling to the `agent_runs` row shape — and if the escalation is ever refined to consult the resolved route (`modelRuntimeProvider`, `builtInModelKeyId`), this case will report a false pass or a false failure for reasons unrelated to product behavior.

Consequence: an internal-implementation-coupled test that does not actually exercise the built-in Claude route it names.

Suggested resolution (either is acceptable): build the case as a real built-in Claude run with `seedBuiltInModelKey("claude-sonnet-5")` + `updateOrgModelPolicies(... defaultProviderType: "built-in")` before `createRun({ modelProvider: "built-in" })`, or drop the case entirely — the production predicate reads only `transitionFailureReason` and `run.modelProvider`, both already covered by the built-in Codex case — and keep the framework-independence claim as the code comment it already is.

### Non-blocking observations

1. **The escalation does not reach the Pi API-first terminal path.** When `failApiFirstTurn$` is invoked with `suppressCompletionFailureLog = apiOwnershipExpired || apiModelFailed` (`turbo/apps/api/src/signals/services/pi-api-first-turn.service.ts:2515-2518`, consumed as `suppressFailureLog` at `:2367`), `logAgentRunCompletionOutcome` skips `logRunFailure` entirely (`agent-webhook-complete.service.ts:265-270`), leaving only the `L.warn("Pi API first-turn outcome", ...)` record at `:2325`. That path is narrow (it requires the sandbox handoff publication to fail) and is pre-existing, but for those runs the stated goal — "a built-in capacity failure is visible at error" — is still unmet. Worth a follow-up note on #33360 rather than a change here.
2. **The severity map is now non-uniform across platform-owned failures and across components.** Other reasons that also reject a platform-owned credential on a built-in route — `invalid_api_key`, `invalid_credentials`, `provider_server_error`, `provider_stream_timeout`, `provider_rate_limited`, `usage_limit` — remain at `WARN` in the same `logRunFailure`, and the Runner classifies `ProviderOverloaded` at info for every provider (`crates/runner/src/cmd/start/job_terminal_log.rs:459-486`). That is consistent with the PR's declared scope and I am not asking for it here; it is the kind of asymmetry that is easy to forget when the next capacity incident is triaged.
3. **`keeps one error when a duplicate repeats the capacity failure` is only partly new coverage.** Its central assertion, `genericFailureLogCalls(...)` of length 1, is level-agnostic and held before this diff as well (the duplicate completion already produced no second record); only the surrounding first-completion escalation is new. The racing-completions case does bind to the new level via its explicit `axiomLogging.error` assertion. No change required, but note `docs/testing.md`'s "avoid duplicate cases" guidance if this matrix keeps growing.
4. **`L.error` has no Sentry or paging coupling in-repo.** `turbo/apps/api/src/lib/log.ts` wires levels straight to the Axiom transport with no error-level side effect, and no in-repo monitor, workflow, or doc references the `"Run failed"` record, so the blast radius of the level change is the operator's Axiom queries — matching the PR's deployment-compatibility note. I did not inspect monitors configured outside the repository.

### Testing

- Coverage added: escalation for a genuine built-in run, non-escalation for `null` and legacy providers, the built-in `insufficient_credits` `DEBUG` branch, a duplicate completion, and two racing completions. All cases drive the real `POST` completion webhook with a sandbox bearer token and verify the terminal outcome through `api.readRun` plus the `read-run-failure-reason` test-only read (that read is the repository's sanctioned pattern here: `failureReason` is not exposed on any production runs response).
- Binding verified independently: with the pre-change `L.warn` selection, the `axiomLogging.error` assertions in the two escalation cases and in the racing case have no matching call, so they fail; the `warn`-severity, credit, and globally-suppressed cases are unaffected. The author's reported negative control is consistent with what the code shows.
- Existing controls that still constrain the change are intact: `suppresses %s for a BYOK provider` still includes `provider_overloaded`, `globally suppresses input_too_large / execution_timeout / safety_policy_refusal` still covers built-in, and `keeps built-in rate limiting visible` still pins `provider_rate_limited` at `WARN`.
- Isolation: unique actors/orgs per case, `randomUUID()` fixture ids for the built-in model key, run-id-filtered log matching, and a run-scoped `FOR UPDATE` gate released in the test body and in `onTestFinished` — no ordering, global lock, or shared-row restore.
- Gaps: the boundary problem in Finding 1; no case covers the API-first suppressed path from Observation 1. Verification breadth is the author's local `-t "completion failure logs"` run plus type/lint checks; the full API suite and the `chat-callbacks` suite rely on CI, which is acceptable under the repository's verification guidance but means the racing-completion case's stability is CI-proven rather than locally proven here.

lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: edad62c6e69e90c26e30c68f647d3bc203765df6
PR_AUTO_REVIEW_VERDICT: LGTM

> **Independent review of the repaired HEAD.** Produced by a separate reviewer given only the PR reference, instructed not to read any existing review comment on this PR and not to read the author's working tree, reviewing the full diff of this HEAD against `main` from GitHub. This supersedes the `Changes Requested` review of `692f5f88`, whose P1 is resolved by `edad62c6`.


Inspected HEAD: `edad62c6e69e90c26e30c68f647d3bc203765df6` (branch `fix/issue-33360-built-in-capacity-error`).
Practice revision: `9e8b388d344515944065846cc94ddba6ac2e64ab` (`main`).

### Summary

The change adds `isBuiltInCapacityFailure` to `agent-webhook-complete.service.ts` and extends the existing level selection in `logRunFailure` so a terminal `provider_overloaded` on a run whose stored `modelProvider` is `built-in` is emitted at `L.error` instead of `L.warn`. Nothing else in the completion path moves: the same message, the same fields, the same single record, the same suppression matrix, the same persisted terminal state. The test file adds five cases to the existing `RUN-03 > completion failure logs` matrix and removes nothing that existed on `main`.

### Review Criteria

| --- | --- | --- | --- |
| REVIEW.md | Trace changed behavior through its real entry point and consumers | Pass | `logRunFailure` has exactly one caller, `logAgentRunCompletionOutcome` (service line 269), which itself runs only from `completeAgentRun$` and only when `commit.transitioned` is true (line 1047). Every earlier return was checked: `run.status === "timeout"` (lines 967, 643) never reaches the log; `transitionFailureKind === "missing-checkpoint"` (line 258) returns before `logRunFailure` and carries no `failureReason`; the duplicate-completion branch (line 1049) emits only `L.debug("Processed duplicate completion for terminal run")`. |
| REVIEW.md | New condition cannot disagree with an adjacent decision on the same state | Pass | The adjacent decision is `shouldSuppressKnownFailureLog` (line 234), which reads the same `run.modelProvider` through `modelProviderTypeSchema.safeParse(...)` then `isBuiltInModelProviderType(parsed)`. The new predicate reads the raw value. `isBuiltInModelProviderType` (`model-provider-types.ts:22`) returns `true` only for the exact literal `"built-in"`, and `"built-in"` is a member of `MODEL_PROVIDER_TYPE_IDS`, so it always parses. The two therefore agree on every input: built-in → not suppressed → `error`; any other valid type → suppressed → no record; `null`/unparseable → not suppressed → `warn` (unchanged). The same raw-value call style is already used in `built-in-model-provider-failure.service.ts:96`. |
| REVIEW.md | Races, cancellation, ordering, auth, permissions, serialization, failure outcomes | Pass | Severity selection is pure and derived from an already-committed `CompletionCommit`; it adds no I/O, no await, and no ordering dependency. Concurrency is unchanged and is exercised by the new race case through `holdAgentRunRowLockFixture`. `L.error` is safe as an unbound method reference: `createLogger` (`lib/log.ts:476`) defines every level as an arrow function closing over `loggerInstance`, exactly as the pre-existing `L.warn`/`L.debug` references rely on. |
| REVIEW.md | Meaningful coverage for the new behavior / regression fix | Pass | `reports capacity exhaustion on a built-in run as an error` asserts `toHaveLength(1)` on `axiomLogging.error` with the full field payload, plus `genericFailureLogCalls(...)` length 1, so a surviving `WARN` duplicate or a missing `ERROR` both fail. `keeps capacity exhaustion on $name at warning severity` pins the negative side for null and legacy providers, which is the only way to reach `logRunFailure` with `provider_overloaded` on a non-built-in run (every valid non-built-in type is fully suppressed). `keeps built-in credit exhaustion on its own debug record` pins branch precedence. No case can pass vacuously. |
| REVIEW.md | Existing coverage still protects the unchanged paths | Pass | The pre-existing `suppresses %s for a BYOK provider`, `globally suppresses %s`, `keeps $name visible` (including built-in `provider_rate_limited`), `keeps the missing-checkpoint warning visible`, and `does not relog $firstReason when a duplicate reports $lateReason` cases are untouched. The diff against `main` is 159 additions and 0 deletions in the test file, so no previously present case was removed and there is no coverage gap from removal. |
| REVIEW.md | Containment / feature or permission gating for a changed flow | N/A | No user-facing behavior, route, permission, or capability changes. The only observable delta is the operator log level of one record; `docs/fallback.md` §2 and the feature-switch skill govern user-facing features, which this is not. |
| REVIEW.md | ccstate ownership rules | N/A | No signal, command, computed, ref, or effect is added or moved. `logRunFailure` and `isBuiltInCapacityFailure` are plain module functions invoked synchronously inside the existing `completeAgentRun$` command. |
| REVIEW.md | Report concrete practice violations with rule and source location | Pass | No practice violation rises to P0/P1. The non-blocking items below cite the specific rule and line. |
| REVIEW.md | Schema, API contract, response validation, `jsonb(...).$type<...>()` | N/A | No Drizzle schema, no migration, no ts-rest contract, no `jsonb` type, and no response body changes. `agentRuns.modelProvider` is only read. |
| REVIEW.md | Coverage for response-validated persisted changes | N/A | No persisted shape changes, so the mandatory-`Changes Requested` rule for that class does not apply. |
| REVIEW.md | Documented exposure window for independently deployed components | N/A | The change is confined to `apps/api` and produces no cross-component old/new combination: the Runner protocol, queue payloads, webhook contract, and persisted state are all identical. |
| REVIEW.md | Fallbacks declared and justified | N/A | The diff adds a classification branch, not a fallback: no `??`/`||` default, no reader for a retired producer, no fabricated value for a corrupt row, no rollout switch. `docs/fallback.md` §6 lists nothing this matches, so no `Fallbacks` section is required and none is present. |
| REVIEW.md | Review anchored to the inspected HEAD | Pass | `headRefOid` read as `edad62c6e69e90c26e30c68f647d3bc203765df6` and all production, test, and practice files were fetched at that SHA or at the recorded practice SHA. |
| docs/docs.md | Use the index to select the authoritative guidance for the changed surface | Pass | Changed surfaces are API production logic plus API route tests, which route to code quality, testing, API testing, external behavior, and (for the declaration check) fallbacks. Event sourcing, cache, effect, styles, chat cards, runner architecture, and externally-managed-references are not touched and were correctly not applied. |
| docs/bad-smell.md | §1 no `any`; §2 no suppressions; §3 error handling; §4 interface changes; §5 static imports; §6 centralized config; §7 fallback default | Pass | `isBuiltInCapacityFailure(commit: CompletionCommit): boolean` is fully typed with no `any`, no assertion, and no suppression comment. No `try`/`catch` added, no public contract changed, no import added at all (both helpers were already imported), no URL or credential introduced, no fallback branch. |
| docs/testing.md | Choose the boundary; coverage; mock external services at their boundary; unique identities; no fake timers | Pass | Every new case drives the real webhook endpoint through `createWebhookCallbackApi` with a real sandbox bearer token and sets up through `api.createRun`. `context.mocks.axiomLogging` mocks the third-party `@axiomhq/logging` `Logger` (`__tests__/mocks.ts:1274-1300`), so the assertion sits on the external telemetry boundary and still traverses the real `logToAxiom` composition — proven by the test asserting the `context: "webhook:complete"` field that only `lib/log.ts` adds. No fake timers; the race case synchronizes on `expect.poll(lifecycleGate.waiterCount)`, an observable condition. |
| docs/testing/api-testing.md | Route integration through the contract; only external mocks; shared persistent state ownership | Pass | Cases live in `src/signals/routes/__tests__/`, use the shared `testContext()` at module scope, and use no `vi.mock()` of an internal module. Isolation holds under concurrent files: every actor comes from `entitledRunActor()` with a unique identity, `seedBuiltInDefaultModelKey()` registers its own `onTestFinished` release with a UUID fixture id, and both `matchingLogCalls` and `genericFailureLogCalls` filter by the test's own `runId`, so another worker's records cannot satisfy or break an assertion. No global lock, ordering dependency, or residue-tolerant assertion. |
| docs/testing/testing-external-behavior.md | Construct and verify through the production endpoint; exceptions must be rare and justified | Pass, with observation N2 | The escalation case is built entirely through production: `api.createRun(..., modelProvider: "built-in")` persists `agentRuns.modelProvider = "built-in"` verbatim (`agent-run-create.service.ts:2591` sets `type: "built-in"` and carries the resolved route in the separate `concreteType`/`modelRuntimeProvider` fields, and `agentRunModelProviderValues` line 6401 writes `modelProvider.type`). No internal write is used for the behavior under test. The two internal `setRunModelProviderStateFixture` calls in the negative case are discussed in N2. |
| docs/testing/patterns.md | A helper must not hide a direct DB write; unique identities; await owned work | Pass | `completeFailure` wraps production API calls and surfaces its one internal override explicitly through the named `persistedModelProvider` field rather than hiding it. `holdAgentRunRowLockFixture` documents its own justification ("Product APIs cannot pause at this boundary"), its `release()` is idempotent (`test-fixtures/chat-events.ts:1046`), and the race case registers `onTestFinished` before the first `await`, so the lock is released even if the poll fails. |
| docs/testing/anti-patterns.md | AP-1/AP-4/AP-7 mock and implementation coupling; AP-8 over-testing; AP-9 logging as contract; AP-11 service testing | Pass, with observation N3 | AP-9 explicitly permits asserting log content when logging is the contract, which is precisely the contract this PR changes, and the mocked object is third-party, not an internal callback (so AP-1/AP-4 do not bite). No service is imported or called (AP-11). AP-8 is the only tension, covered in N3. |
| docs/fallback.md | §1 no tombstone tests; §3/§4 no default for unreachable or corrupt state; §6 required-fallback list; §9 declaration | Pass | No new fallback and therefore no declaration needed. The new negative cases are not tombstones: they pin a live product behavior (non-built-in capacity failures must stay at `WARN`) rather than asserting that deleted code stays deleted. |

### Findings

No P0 or P1 findings. Nothing here blocks the merge.

**Non-blocking observations**

**N1 — The PR description no longer matches HEAD.** `turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts:19704`

The Tests section states that `reports capacity exhaustion on ... as an error` has "two cases: a genuine built-in Codex run ... and a Claude-framework run whose stored provider is built-in." Commit `edad62c6` replaced that `it.each` with a single non-parameterized case and deleted the Claude-framework case (which had relied on creating an `anthropic-api-key` run and overwriting the stored provider). The body was not updated.

I independently agree with the deletion and disagree with the body, not the code. The removed case added no coverage: the predicate reads only `transitionFailureReason` and `run.modelProvider`, and the only framework-sensitive branch on this path is `logGptApiKeyPiSandboxOutcome`, which cannot short-circuit a built-in run because `GPT_API_KEY_PI_ROUTES` contains only `openai-api-key`, `openrouter-codex`, and `vercel-ai-gateway-codex` (`pi-sandbox-config.ts:51-79`), so `isGptApiKeyPiProviderType("built-in")` is always false. The remaining single case is sufficient. Please correct the Tests section so the merged description describes the merged tests.

**N2 — The two remaining internal provider overrides are not individually justified in the test.** `turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts:19728-19732`

`keeps capacity exhaustion on $name at warning severity` reaches `setRunModelProviderStateFixture` for `persistedModelProvider: null` and `"legacy-unknown-provider"`. Checking each separately rather than accepting one blanket rationale:

- `"legacy-unknown-provider"` is a genuine exception. The create contract validates against `modelProviderWriteTypeSchema` (`contracts/runs.ts:16`, `model-provider-types.ts:1-16`), so no production writer can ever store that string; it is exactly the "historical bad state" `docs/testing/testing-external-behavior.md` §Exceptions allows.
- `null` is weaker. The production writer itself can persist `null`: `agentRunModelProviderValues` returns `{ modelProvider: null, ... }` when no provider resolves (`agent-run-create.service.ts:6393-6398`). I could not establish that no production endpoint reaches that branch, so I cannot confirm this state is unconstructible through the API.

Neither case carries the written justification the guide requires ("the test should state why the production external interface cannot construct the state"). This is non-blocking because the pattern predates this PR — the same two overrides already exist at lines 19644, 19667 and 19672 (inside the `it.each` blocks at 19639 and 19658) for the `provider_rate_limited` and global-suppression matrices — and because the behavior actually being changed is proven through a real production-built built-in run. A one-line comment naming the historical state, or replacing the `null` case with a production path that yields a null provider if one exists, would close it.

**N3 — Two of the five new cases restate idempotency this PR does not change (AP-8).** `turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts:19766` and `:19807`

`keeps one error when a duplicate repeats the capacity failure` and `records one error when completions race the capacity failure` both assert "exactly one record". That guarantee comes from `commit.transitioned` gating `logAgentRunCompletionOutcome` (`agent-webhook-complete.service.ts:1047`), which the diff does not touch, and it is already pinned for the generic record by the pre-existing `does not relog $firstReason when a duplicate reports $lateReason` case at line 19893. The race case additionally pays for a row-lock fixture and cross-connection `pg_backend_pid` waiter polling. They are correct and not harmful, but `docs/testing.md` ("avoid duplicate cases that ... pin incidental implementation") and AP-8 would support dropping the race case and keeping the cheaper sequential duplicate.

**N4 — Two new assertions read persisted state through the test-only route where the external surface already suffices.** `turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts:19803` and `:19856`

`readRunFailureReasonFixture` posts to `/api/test/runtime-state`, which is a test-only surface, not a production endpoint. I checked and `failureReason` is genuinely absent from `contracts/runs.ts`, so there is no external read for it on a threadless run — which makes this a defensible exception and consistent with the pre-existing use inside `completeFailure`. Still, the adjacent `api.readRun(...)` assertion already establishes the externally observable terminal state (`status`, `error`) that these two cases are about, so the extra internal read adds coupling without adding coverage of this PR's behavior.

**N5 — The `debug` control asserts a record production never ships.** `turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts:19751`

`keeps built-in credit exhaustion on its own debug record` asserts on `axiomLogging.debug`. In production the Axiom logger is constructed with `logLevel: "info"` and `lib/log.ts:158-160` documents that "`L.debug` records are dropped here and never reach the dataset"; the test mock replaces the whole `Logger`, so it bypasses that threshold. The assertion is still a valid control over which level `logRunFailure` selects, and the load-bearing half of the case — `genericFailureLogCalls(...)` is 0, so the credit branch still wins and emits no `Run failed` at any level — is unaffected. Worth knowing only so the case name is not read as evidence that an operator can see that record.

### Testing

Coverage for the changed behavior is adequate and binds tightly.

- **Positive binding.** The error case asserts the exact record (`toHaveLength(1)` on `axiomLogging.error` plus `toStrictEqual(expect.objectContaining({ runId, exitCode: 1, error, failureReason: "provider_overloaded", context: "webhook:complete" }))`). Reverting the production ternary makes this fail, because the record would land on `warn`; I verified the mechanism rather than relying on the author's reported negative control.
- **Negative binding.** The null/legacy cases prove the escalation did not widen to every non-suppressed `provider_overloaded`, and the credit case proves branch precedence. Both directions are covered, which is what a severity change needs.
- **Conventions.** Real webhook endpoint, real database, real logger composition, third-party telemetry mocked at its own boundary, unique per-test identities, per-run log filtering, `onTestFinished` cleanup for the seeded built-in model key and the row lock. No `vi.mock()` of internal modules, no fake timers, no DB schema or service import.
- **Unchanged controls.** BYOK `provider_overloaded` stays fully suppressed, built-in `provider_rate_limited` stays at `WARN`, the globally suppressed reasons stay silent for built-in, and the missing-checkpoint warning is untouched. I confirmed the diff deletes no test line against `main`, so there is no removal-driven coverage gap.
- **Regression risk outside the suite.** I searched the repository for consumers of the `"Run failed"` record and found none outside this service and its tests — no in-repo monitor, alert definition, dashboard, or doc keys off its level, so the `warn` → `error` shift breaks no in-repo contract. External monitors were not inspected by me either, and the PR body says so honestly.
- **Paths I checked for a residual gap and cleared.** The Pi API-first terminal path can set `suppressFailureLog`, which would hide the new record; it does not apply here, because `suppressCompletionFailureLog` is `apiOwnershipExpired || apiModelFailed` (`pi-api-first-turn.service.ts:2518`), `apiOwnershipExpired` requires `failure.code === "PI_API_FIRST_TURN_DEADLINE_EXCEEDED"`, and `eligibleApiModelFailure` requires `!failure.failureReason`. A classified `provider_overloaded` satisfies neither, so a built-in capacity failure on the API-first path still reaches `logRunFailure` and still escalates. The sandbox Pi path is likewise unreachable for built-in, per N1.
- **Verification observed.** CI at this HEAD reports 64 passing checks with 14 pending and 22 skipped, including all `test-api` shards and `codecov/patch`; no check is failing. The author's locally reported `chat-callbacks.bdd.test.ts` failures are plausible as environmental — those cases assert user-facing error copy, not log levels, and nothing in this diff can affect `claimChatRunJob` timing.
- **Material gaps.** None that block. The items in N2 through N5 are test-hygiene refinements, not missing scenarios.

PR #33522, opened 2026-09-11
fix(api): absorb rate-limited activity summaries without a diagnostic

lancy:
Closes #33354

## Problem

An optional activity summary that loses its generation to a shared provider rate
limit reaches operators as `warn`:

https://github.com/vm0-ai/vm0/blob/b4762ad9fc399a1a9dbde812f63b0eb7252dd05b/turbo/apps/api/src/signals/services/chat-activity-summary.service.ts#L424-L428

Nothing here can act on it. The snapshot keeps its previous `summary`, so the
response stays a truthful `cooldown` carrying the caller's last usable phrase;
with no phrase ever stored the viewer keeps the generic label its own fallback
renders for an empty batch. The shared cooldown that this very record reports
already bounds the retry.

The service flattened every `OpenRouterRequestError` into a single
`provider_failure` and never consulted `openRouterFailureReason`, the classifier
`auxiliary-generation.service.ts` already shares across the other fast-path
generations, where the same upstream condition is classified `degraded` and
produces no log.

## Change

- A provider-request completion carries the enumerated `reason` beside
  `providerStatus`. `outcome` stays `provider_failure`.
- A completion whose `reason` is `rate_limited` emits **no record at any level**.
- The reason decides this, never the transport status. OpenRouter delivers a
  native rate limit inside an HTTP 200 error envelope that this client wraps as a
  synthetic `502`, and a raw `429` whose native evidence is `auth` or
  `invalid_request` is a real failure that must keep reaching an operator.

Only the diagnostic is skipped. Success, caller cancellation and deadline keep
their priority ahead of request-error classification; the charged cooldown, the
stored summary and the final reread all still run. `min(300000, max(60000,
retryAfterMs ?? 0))`, Retry-After parsing, claim ownership, the fallback and the
content-free boundary are unchanged. No retry is added.

`reason` is the finite `OpenRouterFailureReason` enum, so an unclassified error
stays an explicit `unknown` rather than a claimed cause.

### Observability

Removing the record intentionally changes operator event counts. The documented
rate query now divides by the attempt records instead of the completions, and the
docs state plainly that absorbed rate limits are counted nowhere: an attempt and
its completion are written separately, so the residual between them is not that
count. No replacement record or telemetry channel is added to recover it.

## Scope

Per the issue's dispatch contract this PR owns semantic `rate_limited` only.
Non-rate provider severity and `unknown` belong to #33361; the residual bucket and
timeout/cancelled noise belong to #33362. The new real-failure assertions
[...]

Issue #33362, opened 2026-09-11
bug(api): distinguish invalid output from unconfigured activity-summary generation


seven332:
## Implementation authorized — 2026-09-11

Lancy explicitly approved dispatch of the ready warning-removal and genuine-failure fixes using `/handoff-issue-to-thread`, and requested reuse of each original research thread. This is now an implementation task, superseding earlier research-only/no-PR stopping instructions. The accepted technical findings and evidence limits remain applicable, subject to the scope below.

Sole owner: https://app.okou.ai/chats/c6d87a54-4f3a-492f-ad2c-f090be06cf68 (Claude Opus 5, verified). Controller: https://app.okou.ai/chats/922b01fa-5dbb-43e9-97ee-fca3fb8e44cd.

Controller refreshed current main to `70cae141724e41ac67db20af9826210470311d5e` and reread the affected services and classifier. The relevant behavior is unchanged from the accepted research; no fresh production investigation is required. A complete 24-open-PR inventory, including all files for the unchanged 108-file #33450 HEAD, has no implementation of these target paths at dispatch. Recheck live current main/callers and overlap before coding.

### Authorized scope: non-request/local results plus timeout and cancellation

At `chat-activity-summary.service.ts:339-365,389-428`, successful null generation, rejected non-null optional text and thrown non-request failures are collapsed into invalid_or_unconfigured. Pass the settled generation result to the completion classification so these remain distinguishable. Preserve priority: usable phrase, caller cancellation, deadline, provider request error, then residual.

Residual policy: successful null -> unconfigured; returned non-null text rejected by the existing strict phrase contract -> unusable_output; thrown non-request failure -> existing `openRouterFailureReason(error)`. Omit completion logs entirely for unconfigured/unusable_output, recognized network/upstream transient reasons, output_truncated and unexpected_tool_calls when handled by the established optional fallback. True response-contract failures and unclassified exceptions remain one ERROR, with unknown explicitly unknown. Remove timeout/cancelled completion INFO noise too, preserving their exact ownership/cooldown semantics. Keep success, attempt and cache behavior.

[...]

PR #33530, opened 2026-09-11
fix(api): report unabsorbed activity-summary provider failures as errors

### LINKED ISSUE #33361: chore(api): classify activity-summary gateway failures without hiding outages
## Implementation authorized — 2026-09-11

Lancy explicitly approved dispatch of the ready warning-removal and genuine-failure fixes using `/handoff-issue-to-thread`, and requested reuse of each original research thread. This is now an implementation task, superseding earlier research-only/no-PR stopping instructions. The accepted technical findings and evidence limits remain applicable, subject to the scope below.

Sole owner: https://app.okou.ai/chats/c5823934-621d-4e34-ae01-8f9f38a842cd (Claude Opus 5, verified). Controller: https://app.okou.ai/chats/922b01fa-5dbb-43e9-97ee-fca3fb8e44cd.

Controller refreshed current main to `70cae141724e41ac67db20af9826210470311d5e` and reread the affected services and classifier. The relevant behavior is unchanged from the accepted research; no fresh production investigation is required. A complete 24-open-PR inventory, including all files for the unchanged 108-file #33450 HEAD, has no implementation of these target paths at dispatch. Recheck live current main/callers and overlap before coding.

### Authorized scope: provider-request failures other than semantic rate limits

At `chat-activity-summary.service.ts:339-365,424-428`, provider request errors still discard the semantic reason from `openrouter-failure.ts:34-39,112-181`. Keep `outcome=provider_failure`, add `reason=openRouterFailureReason(error)`, and use that finite reason rather than the outer numeric status.

For an `OpenRouterRequestError` with a non-rate transient reason recognized by the existing classifier (notably upstream_timeout/provider_unavailable), omit the corresponding completion log at every level while retaining the existing reliable fallback and cooldown. Genuine auth/invalid-request failures produce one ERROR. Unclassified request errors remain visible as ERROR with `reason=unknown`; unknown does not prove a specific root cause or main-run failure. A local synthetic 502 may carry any of these semantic reasons. Preserve classifier precedence including auth/invalid_request overriding a transient wrapper.

Semantic `rate_limited` is owned by #33354, including synthetic-502 rate limits. Non-request exceptions and local/timeout/cancelled results are owned by #33362. Reuse their already-merged behavior; do not create a second parser/classifier, implement their branches again, or invent conflicting outcome names such as auth_failure. Shared setup may add the same canonical reason field if it has not landed, with later integration deduplicating it.

Validation: real route/MSW/logger transport with HTTP and completion envelopes, recognized timeout/unavailability, native auth/request overrides (including outer 429), and unknown negative control; no completion at any level for this scope's transient results; exactly one safe ERROR for this scope's real/unclassified failures. Assert retained summary, bounded Retry-After and persisted recovery behavior; cancellation/deadline priority and no content leakage. Update corresponding documentation without claiming that silent outcomes can still be counted from completion logs. Do not assert all 5xx are transient or identify a historical 502 cause absent source evidence.

### Shared activity-summary integration contract

The final combined behavior is: recognized reliably degraded/expected optional completion -> no diagnostic at any level; genuine failures -> ERROR; unknown -> explicitly unknown, not invented causality. Keep success/caller-cancellation/deadline priority, `provider_failure` plus its semantic `reason`, all cooldown/fallback/ownership semantics and the content-free boundary.

At Lancy's latest request, keep one existing owner per issue. This supersedes the earlier controller suggestion to put all three issues in one later shared PR: #33354 owns semantic rate_limited, #33361 owns remaining provider-request classifications, #33362 owns the residual bucket and timeout/cancelled completion noise. Their outcome sets are disjoint even though edits touch the same service/test/docs files. Do not broaden into siblings or revert their accepted policy when rebasing. The combined boundary is not complete until all three scopes are independently accepted. A ready owned PR proceeds through normal gates; overlap alone is not a reason to wait or reserve merge order. The first merged PR establishes main; later owners integrate it and re-review their exact HEAD. Do not introduce temporary scaffolding or tests that entrench WARN as the final contract for another owner's unresolved true failure.

### Delivery, validation and authority

Reuse your accepted research; refresh the relevant current code rather than repeat the investigation. Fetch and rebase origin/main, implement one focused PR, run targeted verification and self-review, and immediately invoke `/pr-auto <explicit PR URL>` in the SAME owner thread. Complete a fresh independent full-SHA tracked review, required CI and normal protected merge queue with HEAD guard until GitHub reports MERGED or a genuine blocker. No direct/admin/bypass merge; no duplicate owner/PR. Preserve other contributors' changes; overlap is inventory, not an ordering blocker. A runtime limit leaves a precise resumable handoff in the same thread.

No full local Vitest suite or dev server. Report missing local database/compilation/test-body limitations honestly and rely on the actual PR CI for those checks. English repository/issue/PR artifacts and lowercase conventional commit subjects. Only pause for an actual zero-job action_required / Approve workflows gate when this owned PR modifies Actions definitions; report the exact approval link and do not try reruns/empty commits/branch changes around absent approval.

These changes require no database migration/backfill, provider model change, production-data mutation or new monitoring access. Logging removal intentionally changes operator-event counts; update the affected documentation/tests, but do not claim external monitors were independently checked or add a permission prerequisite for ordinary implementation. Keep all diagnostics content-free and preserve real failures.

The owner stops after reporting GitHub-confirmed merged SHA, current-main inclusion, review/check links and verification limitations. The controller independently accepts the merged code and retires the disposable watcher. No implementation owner may invoke /release-production, merge release PRs, approve production deployments or perform production changes. Production release requires actual authorization and a separate release-only executor after exact candidate inclusion and controller acceptance.

The complete batch consists of #33354, #33360 (logging only), #33361, #33362 and #33377. Each has its own reused owner and completed/failed/cancelled watcher. Keep remaining shared routing policy for #33360 explicitly unimplemented and pending discussion; do not silently turn it into a blocker for the ready fixes.

<details>
<summary>Original incident report and research context (history; current authorized scope is above)</summary>

## Summary

classify activity-summary gateway failures without hiding outages. The current-window observations and their causal limits are recorded below.

Related context: #32819, #33239. Closed issues remain closed; this review does not override prior implementation/release authority.

## Impact

Three optional activity-summary attempts report providerStatus=502. Main-run failure is not shown.

Severity: **medium**. Type: **upstream gateway failure investigation**. Severity is triage, not a raw-event-to-user equivalence.

### Coverage

- Beijing: **[2026-09-10 00:00, 2026-09-11 00:00) Asia/Shanghai**. Exact UTC: **[2026-09-09T16:00:00Z, 2026-09-10T16:00:00Z)**.
- Axiom: ONLY `vm0-web-logs-prod`, `level == "error" or level == "warn"`; **3 records**, first `2026-09-10T03:31:09.776Z`; last `2026-09-10T10:28:08.473Z`. Levels: warn: 3.
- Distinct logged run IDs: **3**. 0 Axiom records lack a run ID; these do not imply additional runs.
- Per-run record distribution: run-020, run-151, run-128 each has {"Activity summary completion":1}.
- Privacy: these are selected original source fields. Private IDs, device tokens, bound SQL parameters, custom workflow paths, signed/private URLs and unrelated user content are redacted or omitted. Only allowlisted method/status/route/timestamp network breadcrumbs are retained when directly relevant; no request bodies, general breadcrumbs, raw user identities or private Sentry links are published.

### Variant 1 — Activity summary completion (3 records)

- Source: Axiom; first `2026-09-10T03:31:09.776Z`; last `2026-09-10T10:28:08.473Z`.
- Native causal fields: `fields.cooldownMs`, `fields.durationMs`, `fields.outcome`, `fields.providerStatus`.
- Inner error: **<not logged>**. Only the original outcome/status fields are available.
- fields.outcome: provider_failure: 3.
- fields.providerStatus: 502: 3.

Original sanitized sample 1:

```json
{
  "_time": "2026-09-10T03:31:09.776Z",
  "fields.context": "api:activity-summary",
  "fields.cooldownMs": 60000,
  "fields.durationMs": 2596,
  "fields.outcome": "provider_failure",
  "fields.providerStatus": 502,
  "fields.runId": "<redacted:run-020>",
  "level": "warn",
  "message": "Activity summary completion",
  "source": "api",
}
```

Variant reconciliation: 3 = **3 source records**.

## Current code analysis

Fetched origin/main: **`75095c26e72eb7deb40f5c15b920aa0764d94ceb`**. This is read-only source inspection, not proof of which revision each event executed.

- [turbo/apps/api/src/signals/services/chat-activity-summary.service.ts](https://github.com/vm0-ai/vm0/blob/75095c26e72eb7deb40f5c15b920aa0764d94ceb/turbo/apps/api/src/signals/services/chat-activity-summary.service.ts)
- [turbo/apps/api/src/signals/external/openrouter.ts](https://github.com/vm0-ai/vm0/blob/75095c26e72eb7deb40f5c15b920aa0764d94ceb/turbo/apps/api/src/signals/external/openrouter.ts)

The native completion records contain provider_failure/status 502 but no inner provider error code, body or stack. Unlike the initial-thinking completion variants, there is no native 429 to justify merging these into rate limiting. Current completionRecord retains only numeric provider status and falls back to a generic/last usable activity phrase. #33239 leaves this provider_failure branch at WARN.

## Desired outcome

Determine a privacy-safe failure class and verify existing bounded fallback/cooldown. If the exact transient gateway outcome is normal optional degradation, lower only that signal; preserve actionable sustained outage and malformed/auth failures. Do not infer or implement 429-specific recovery from the wrapper status alone.

## Ownership

Selected owner: @lancy. @lancy owns current activity-summary outcome classification through #33239. Confidence: high.

</details>

lancy:
Closes #33361

## Problem

`chat-activity-summary.service.ts` already classifies a provider-request failure
by the shared `openRouterFailureReason` (added for `rate_limited` in #33522),
but it acts on only one reason. `absorbedCompletion` recognizes `rate_limited`
alone, so the remaining transient reasons the same classifier owns —
`upstream_timeout` and `provider_unavailable` — still reach operators even
though the caller keeps its last usable phrase and the charged cooldown already
bounds the retry.

At the same time, everything that is not absorbed lands on one `warn`: a real
credential or request-contract defect is reported at exactly the level of an
unusable model result. #33522 deliberately left that severity unasserted — its
test says "the level is the adjacent severity decision and is asserted
elsewhere". This is that decision.

`providerStatus` cannot make either call: OpenRouter delivers a native error
inside an HTTP 200 envelope that this client wraps as a synthetic `502`, so the
same number can mean unavailability, a contract defect or an auth failure.

## Change

Route the rest of the `provider_failure` branch by the same shared reason.

- **Absorb every reason `isTransientProviderFailure` accepts.** An upstream
  timeout or unavailability becomes as silent as the already-merged rate limit —
  no record at any level, and no replacement diagnostic. The predicate is the
  shared classifier's own, so a reason that stops counting as transient there
  stops being absorbed here. `network` is in that set but is unreachable on this
  branch, because a transport rejection is not an `OpenRouterRequestError`.
- **Report everything else at `error`.** `auth` and `invalid_request` are real
  credential and contract defects. A request error the classifier could not
  place stays `reason: unknown`: it names no cause and implies nothing about the
  main run, but nothing here shows it recovers, so it is still reported.

Precedence is unchanged, so a raw `429` carrying a native `401` is an `auth`
error rather than an absorbed rate limit, and a native unavailability inside a
synthetic `502` is absorbed.

Only the level and the absorbed set change. Success, caller cancellation and
deadline keep their priority ahead of request-error classification; the charged
cooldown, bounded `Retry-After`, the retained summary, claim ownership and
lease, the final reread and the content-free boundary are untouched. The
residual `invalid_or_unconfigured` bucket keeps its `warn` for its own owner
(#33362), and #33522's absorbed `rate_limited` diagnostic is not restored.

## Validation

`turbo/apps/api/src/signals/routes/__tests__/chat-activity-summary.test.ts` —
34/34 pass against a local PostgreSQL 18 (UTC), through the real route, the MSW
provider transport and the production logger with the real Axiom SDK/transport.
Three new cases, reusing this suite's existing `completionRecords`,
`providerFailureRecords` and `operatorRecords` helpers:

- HTTP `503` → absorbed: no `provider_failure` record and no operator record at
  any level, while the previous summary and its revision are retained,
  `retryAfterMs` stays inside the 60 s bound, the stored cooldown still
  suppresses the next provider call, and both attempts are still counted.
- A 200 envelope carrying native `UNAVAILABLE` (the synthetic-502 path) and HTTP
  `504` → both absorbed: no completion record at all, attempts still counted.
- HTTP `429` carrying native `401`, an `INVALID_ARGUMENT` envelope, and HTTP
  `500` → exactly three `error` records with `reason` `auth` / `invalid_request`
  / `unknown`, `providerStatus` `429` / `502` / `500`, and `cooldownMs`
  `120000` / `60000` / `60000`, proving native precedence over the wrapper
  status, that a synthetic 502 is not assumed transient, and that not every 5xx
  is absorbed.

Every case asserts the provider body never reaches a record.

The three new cases were confirmed non-vacuous: reverting `absorbedCompletion`
to `rate_limited` only and removing the `error` routing fails all three, and
they pass again once restored.

Also run locally: `tsc` on the `core` and `tests` projects, `oxlint` including
the type-aware pass, `eslint --max-warnings 0`, and `prettier`.

### Limitations

No full local Vitest suite and no dev server, per the task constraints; the PR
pipeline owns full-suite and environment verification. This change does not make
an absorbed failure individually attributable after the fact — it removes that
record by design; `reason` improves the diagnostics that are still emitted. The
historical production `502` records that motivated the issue were logged without
their origin or native code, so they cannot be attributed retroactively and no
claim is made about a specific historical provider fault.

## Documentation

`docs/chat-thread-activity-summary.md` records the widened absorbed set, the new
`error` level, the explicit `unknown` semantics, and that `invalid_or_unconfigured`
keeps its `warn` under separate ownership. It also states that the reported
`provider` count now covers only the emitted failures and that absorbed ones
remain counted nowhere.

### DIFF
--- a/docs/chat-thread-activity-summary.md
+++ b/docs/chat-thread-activity-summary.md
@@ -97,14 +97,14 @@ transport's `info` threshold. The shared logger and unrelated debug filtering
 are unchanged. Axiom events retain `source: api`, the stable message, and the
 following nested `fields`:
 
-| Message                        | Context                | Level                                                                                                                                                             | Safe fields besides context                                                                                                            |
-| ------------------------------ | ---------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------- |
-| `Activity summary cache`       | `api:activity-summary` | info                                                                                                                                                              | `runId`, `outcome` (existing response status)                                                                                          |
-| `Activity summary attempt`     | `api:activity-summary` | info                                                                                                                                                              | `runId`                                                                                                                                |
-| `Activity summary completion`  | `api:activity-summary` | info for success/timeout/cancelled; warn for invalid_or_unconfigured and for a reported provider_failure; no record at any level for an absorbed provider_failure | `runId`, `outcome`, `durationMs`, `cooldownMs`; numeric `providerStatus` and the enumerated `reason` only for `OpenRouterRequestError` |
-| `Activity summary unavailable` | `api:activity-summary` | warn                                                                                                                                                              | `runId`, `outcome: storage_failed`                                                                                                     |
-| `Activity snapshot capture`    | `api:run-activity`     | info for written/unchanged and for an expected failure; warn otherwise                                                                                            | `runId`, `outcome`, `eventCount`; `stage` and `errorCode` on failure                                                                   |
-| `Activity snapshot cleanup`    | `api:run-activity`     | info on success; warn on failure                                                                                                                                  | `outcome`, `removed`, `retentionMs`; `errorCode` on failure                                                                            |
+| Message                        | Context                | Level                                                                                                                                                                | Safe fields besides context                                                                                                            |
+| ------------------------------ | ---------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------- |
+| `Activity summary cache`       | `api:activity-summary` | info                                                                                                                                                                 | `runId`, `outcome` (existing response status)                                                                                          |
+| `Activity summary attempt`     | `api:activity-summary` | info                                                                                                                                                                 | `runId`                                                                                                                                |
+| `Activity summary completion`  | `api:activity-summary` | info for success/timeout/cancelled; warn for invalid_or_unconfigured; error for a reported provider_failure; no record at any level for an absorbed provider_failure | `runId`, `outcome`, `durationMs`, `cooldownMs`; numeric `providerStatus` and the enumerated `reason` only for `OpenRouterRequestError` |
+| `Activity summary unavailable` | `api:activity-summary` | warn                                                                                                                                                                 | `runId`, `outcome: storage_failed`                                                                                                     |
+| `Activity snapshot capture`    | `api:run-activity`     | info for written/unchanged and for an expected failure; warn otherwise                                                                                               | `runId`, `outcome`, `eventCount`; `stage` and `errorCode` on failure                                                                   |
+| `Activity snapshot cleanup`    | `api:run-activity`     | info on success; warn on failure                                                                                                                                     | `outcome`, `removed`, `retentionMs`; `errorCode` on failure                                                                            |
 
 Completion outcomes are `success`, `timeout`, `cancelled`, `provider_failure`,
 and `invalid_or_unconfigured`. A `provider_failure` also carries `reason`, the
@@ -124,12 +124,24 @@ recovery at the shared cooldown; `cancelled` means the request's own lifetime
 ended before the provider answered. Neither has an operator action, so a
 per-event `warn` only trains operators to ignore the record.
 
-A `provider_failure` with `reason: rate_limited` is absorbed the same way and
-emits no record at any level. The shared account is momentarily over its limit,
-the caller keeps its last usable phrase or the generic label, and the cooldown
-that record would have reported already bounds the retry, so there is nothing
-for an operator to do. Contract and configuration defects, and the provider
-failures that are not an absorbed rate limit, still reach an operator.
+A `provider_failure` is absorbed the same way, and emits no record at any level,
+whenever its `reason` is one the shared classifier calls transient:
+`rate_limited`, `upstream_timeout` or `provider_unavailable`. The provider is
+momentarily over its limit or unreachable, the caller keeps its last usable
+phrase or the generic label, and the cooldown that record would have reported
+already bounds the retry, so there is nothing for an operator to do. The set is
+the classifier's own, so a reason that stops counting as transient there stops
+being absorbed here. (`network` is in that set but cannot occur on this branch:
+a transport rejection is not an `OpenRouterRequestError`, so it becomes
+`invalid_or_unconfigured` instead.)
+
+Every other `provider_failure` reason is a real failure and records at `error`:
+`auth` and `invalid_request` are credential and request-contract defects, and a
+request error the classifier could not place records as `reason: unknown`.
+`unknown` stays unknown — it names no cause and implies nothing about the main
+run — but it is still reported, because nothing here shows it recovers on its
+own. `invalid_or_unconfigured` is a separate bucket and keeps its `warn`; its
+classification is resolved separately and that level is not a final policy.
 
 Deadline pressure is a rate, not an event. Read it from the attempt records,
 which count every generation, against the completions that are still emitted:
@@ -147,12 +159,13 @@ which count every generation, against the completions that are still emitted:
 
 Alerting on that rate is an operator decision and is not configured here.
 
-Absorbed rate limits are deliberately not counted anywhere. `attempts` minus the
-emitted completions is not that count: an attempt and its completion are written
-separately, so an instance that stops between them leaves the same residual. No
-replacement record or telemetry channel is added to recover the number, and the
-rate-limit counters of other auxiliary generations describe their own features,
-not this one.
+Absorbed provider failures are deliberately not counted anywhere. `attempts`
+minus the emitted completions is not that count: an attempt and its completion
+are written separately, so an instance that stops between them leaves the same
+residual. No replacement record or telemetry channel is added to recover the
+number, and the rate-limit counters of other auxiliary generations describe
+their own features, not this one. `provider` above therefore counts only the
+reported failures, which are exactly the ones at `error`.
 
 A failed capture is classified into a finite set instead of one opaque
 `write_failed`. `contended` (`55P03`) and `run_missing` (`23503`) are expected
--- a/turbo/apps/api/src/signals/routes/__tests__/chat-activity-summary.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/chat-activity-summary.test.ts
@@ -1221,6 +1221,143 @@ describe("thread activity summary", () => {
     expect(JSON.stringify(logs)).not.toContain("PRIVATE_PROVIDER_BODY");
   });
 
+  it("absorbs upstream unavailability and keeps the last known summary", async () => {
+    const f = await fixture();
+    const diagnostics = captureDiagnostics();
+    const inputs = provider((_input, index) => {
+      return index === 1
+        ? "Preparing the launch checklist"
+        : HttpResponse.json(
+            { error: { code: 503, message: "PRIVATE_PROVIDER_BODY" } },
+            { status: 503 },
+          );
+    });
+    const first = await summarize(f.actor, f.run);
+    await deliver(f, [tool(0)]);
+    await advanceRunActivityClockFixture(f.run.runId, 16_000);
+    const failed = await summarize(f.actor, f.run);
+    expect(failed.messages).toStrictEqual(first.messages);
+    expect(failed.summaryRevision).toBe(first.summaryRevision);
+    expect(failed.retryAfterMs).toBeGreaterThan(55_000);
+    expect(failed.retryAfterMs).toBeLessThanOrEqual(60_000);
+    // The charged cooldown still bounds the next provider call.
+    await summarize(f.actor, f.run);
+    expect(inputs).toHaveLength(2);
+    const logs = await diagnostics();
+    expect(providerFailureRecords(logs)).toStrictEqual([]);
+    expect(operatorRecords(logs)).toStrictEqual([]);
+    // The attempt is still counted; only the absorbed outcome is silent.
+    expect(
+      logs.filter((event) => {
+        return event.message === "Activity summary attempt";
+      }),
+    ).toHaveLength(2);
+    expect(JSON.stringify(logs)).not.toContain("PRIVATE_PROVIDER_BODY");
+  });
+
+  it("absorbs an envelope unavailability and an upstream timeout", async () => {
+    const f = await fixture();
+    const diagnostics = captureDiagnostics();
+    const inputs = provider((_input, index) => {
+      // A native unavailability arrives inside a successful envelope, which
+      // this client wraps as a synthetic 502; the reason decides, not the status.
+      return index === 1
+        ? HttpResponse.json({
+            choices: [
+              {
+                finish_reason: "error",
+                error: {
+                  code: "UNAVAILABLE",
+                  message: "PRIVATE_PROVIDER_BODY",
+                },
+              },
+            ],
+          })
+        : HttpResponse.json(
+            { error: { code: 504, message: "PRIVATE_PROVIDER_BODY" } },
+            { status: 504 },
+          );
+    });
+    const unavailable = await summarize(f.actor, f.run);
+    expect(unavailable.status).toBe("cooldown");
+    expect(unavailable.retryAfterMs).toBeGreaterThan(55_000);
+    expect(unavailable.retryAfterMs).toBeLessThanOrEqual(60_000);
+    await advanceRunActivityClockFixture(f.run.runId, 61_000);
+    const timedOut = await summarize(f.actor, f.run);
+    expect(timedOut.status).toBe("cooldown");
+    expect(inputs).toHaveLength(2);
+    const logs = await diagnostics();
+    expect(completionRecords(logs)).toStrictEqual([]);
+    expect(operatorRecords(logs)).toStrictEqual([]);
+    expect(
+      logs.filter((event) => {
+        return event.message === "Activity summary attempt";
+      }),
+    ).toHaveLength(2);
+    expect(JSON.stringify(logs)).not.toContain("PRIVATE_PROVIDER_BODY");
+  });
+
+  it("reports credential, contract and unplaced provider failures as errors", async () => {
+    const f = await fixture();
+    const diagnostics = captureDiagnostics();
+    const inputs = provider((_input, index) => {
+      if (index === 1) {
+        // A native credential defect inside a rate-limited wrapper: the native
+        // code decides the class, so this is not absorbed with the 429s.
+        return HttpResponse.json(
+          { error: { code: 401, message: "PRIVATE_PROVIDER_BODY" } },
+          { status: 429, headers: { "Retry-After": "120" } },
+        );
+      }
+      if (index === 2) {
+        return HttpResponse.json({
+          error: {
+            code: "INVALID_ARGUMENT",
+            message: "PRIVATE_PROVIDER_BODY",
+          },
+        });
+      }
+      return HttpResponse.json(
+        { error: { code: 500, message: "PRIVATE_PROVIDER_BODY" } },
+        { status: 500 },
+      );
+    });
+    const auth = await summarize(f.actor, f.run);
+    expect(auth.retryAfterMs).toBeGreaterThan(110_000);
+    expect(auth.retryAfterMs).toBeLessThanOrEqual(120_000);
+    await advanceRunActivityClockFixture(f.run.runId, 121_000);
+    await summarize(f.actor, f.run);
+    await advanceRunActivityClockFixture(f.run.runId, 61_000);
+    await summarize(f.actor, f.run);
+    expect(inputs).toHaveLength(3);
+    const logs = await diagnostics();
+    const expected = [
+      { reason: "auth", providerStatus: 429, cooldownMs: 120_000 },
+      { reason: "invalid_request", providerStatus: 502, cooldownMs: 60_000 },
+      // Not every 5xx is absorbed: an error the shared classifier cannot place
+      // stays reported as `unknown` and claims no cause.
+      { reason: "unknown", providerStatus: 500, cooldownMs: 60_000 },
+    ] as const;
+    for (const fields of expected) {
+      expect(logs).toContainEqual(
+        expect.objectContaining({
+          level: "error",
+          message: "Activity summary completion",
+          fields: {
+            context: "api:activity-summary",
+            runId: f.run.runId,
+            outcome: "provider_failure",
+            durationMs: expect.any(Number),
+            ...fields,
+          },
+        }),
+      );
+    }
+    expect(providerFailureRecords(logs)).toHaveLength(expected.length);
+    expect(operatorRecords(logs)).toHaveLength(expected.length);
+    expect(JSON.stringify(logs)).not.toContain("PRIVATE_PROVIDER_BODY");
+  });
+
   it("records a missed deadline as an accepted degradation, not a failure", async () => {
     const f = await fixture();
     const entered = createDeferredPromise<void>(context.signal);
--- a/turbo/apps/api/src/signals/services/chat-activity-summary.service.ts
+++ b/turbo/apps/api/src/signals/services/chat-activity-summary.service.ts
@@ -31,6 +31,7 @@ import {
   OpenRouterRequestError,
 } from "../external/openrouter";
 import {
+  isTransientProviderFailure,
   openRouterFailureReason,
   type OpenRouterFailureReason,
 } from "../external/openrouter-failure";
@@ -345,18 +346,41 @@ interface CompletionRecord {
 
 /**
  * Provider failures this optional generation already absorbs, recognized by the
- * shared semantic reason rather than by a transport status. A shared rate limit
- * resolves without anyone here acting: the caller keeps its last usable phrase
- * or the existing generic label, and the same cooldown this record would report
- * bounds the retry. Emitting it at any level would only teach operators to
+ * shared semantic reason rather than by a transport status. None of them
+ * resolves by anyone here acting: the caller keeps its last usable phrase or
+ * the existing generic label, and the same cooldown this record would report
+ * bounds the retry. Emitting them at any level would only teach operators to
  * ignore the record. The outer status cannot decide this — a native rate limit
- * arrives inside a synthetic 502 envelope, and a raw 429 whose native evidence
- * is auth or invalid_request is a real failure that must keep reaching someone.
+ * or an upstream timeout/unavailability arrives inside a synthetic 502
+ * envelope, and a raw 429 whose native evidence is auth or invalid_request is a
+ * real failure that must keep reaching someone.
+ *
+ * The set is the shared classifier's own, so a reason that stops counting as
+ * transient there stops being absorbed here. `network` belongs to it but is
+ * unreachable on this branch: a transport rejection is not an
+ * `OpenRouterRequestError`, so it lands in the residual bucket instead.
  */
 function absorbedCompletion(completion: CompletionRecord): boolean {
   return (
     completion.outcome === "provider_failure" &&
-    completion.reason === "rate_limited"
+    completion.reason !== undefined &&
+    isTransientProviderFailure(completion.reason)
+  );
+}
+
+/**
+ * A provider-request failure the shared classifier does not absorb: a
+ * credential defect, a request-contract defect, or a request error it could not
+ * place. Each needs an owner, so each is a real failure rather than a
+ * degradation. `unknown` stays unknown — it names no cause and implies nothing
+ * about the main run — but it is still reported, because nothing here shows it
+ * recovers on its own.
+ */
+function realProviderFailure(completion: CompletionRecord): boolean {
+  return (
+    completion.outcome === "provider_failure" &&
+    completion.reason !== undefined &&
+    !isTransientProviderFailure(completion.reason)
   );
 }
 
@@ -460,7 +484,9 @@ async function generateSummary(
   // stored summary and the final reread below all still run, so an absorbed
   // failure degrades exactly like a reported one.
   if (!absorbedCompletion(completion)) {
-    if (expectedOutcome(completion.outcome)) {
+    if (realProviderFailure(completion)) {
+      log.error("Activity summary completion", completion);
+    } else if (expectedOutcome(completion.outcome)) {
       log.info("Activity summary completion", completion);
     } else {
       log.warn("Activity summary completion", completion);


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: 5ea9219b6710fdba8af5b3d9538e0a1f58843593
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `5ea9219b6710fdba8af5b3d9538e0a1f58843593`.
Practice revision: `b4762ad9fc399a1a9dbde812f63b0eb7252dd05b` (`main`).

### Summary

`chat-activity-summary.service.ts` stops classifying a failed optional generation by the outer `OpenRouterRequestError.status` and instead reads the finite reason the OpenRouter wrapper already associates with the error. Reasons the existing fallback absorbs (`upstream_timeout`, `provider_unavailable`) emit no completion record at any level; `auth`, `invalid_request` and an unclassified request error record at `error` with a safe `reason`. `rate_limited` and the residual `invalid_or_unconfigured` bucket keep their current level.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | `completionLevel`/`absorbedProviderFailure`/`completionRecord` are file-local (no external references found); the only caller is `generateSummary`, reached through the `summarize` route exercised by the tests. |
| `REVIEW.md` | Races, cancellation, ordering and failure outcomes | Pass | Success → caller-cancellation → deadline → provider request → residual precedence in `completionRecord` is unchanged. When a request error races the deadline, `expectedOutcome` still wins in `completionLevel`, so `timeout` stays `info` exactly as before. |
| `REVIEW.md` | Meaningful coverage for new behavior; existing coverage still protects | Pass | Four new route-level cases cover both transports, both absorbed reasons, and three error reasons; all 28 pre-existing cases in the file still pass (32/32). |
| `REVIEW.md` | Tests verify externally observable state, not internals/mock calls | Pass | Assertions are on the `summarize` HTTP body (`messages`, `summaryRevision`, `status`, `retryAfterMs`), on follow-up requests proving the persisted cooldown, and on the real Axiom ingest HTTP payload. |
| `REVIEW.md` | Regression coverage and containment for a rerouted existing flow | Pass | The rerouted flow is the `provider_failure` level. The updated 429 case pins the untouched sibling path, and the auth-inside-429 case pins classifier precedence. Feature stays behind `threadActivitySummary`. |
| `REVIEW.md` | Report concrete practice violations with rule and source location | Pass | No violations found; non-blocking observations are separated below. |
| `docs/bad-smell.md` | No `any`; narrow untrusted boundaries | Pass | `CompletionRecord` and `OpenRouterFailureReason` are explicit; `failure: unknown` is still narrowed with `instanceof`. |
| `docs/bad-smell.md` | No lint/type suppressions | Pass | No suppression comment added; `oxlint` (incl. type-aware) and `eslint --max-warnings 0` pass on both changed files. |
| `docs/bad-smell.md` | Catch only at a meaningful boundary; preserve recovery, do not hide invariants | Pass | No catch added or removed. Only a diagnostic emission is dropped; the cooldown write, retained phrase and read-back after it are untouched, and an unclassified error fails loud as `error` rather than defaulting to silence. |
| `docs/bad-smell.md` | Document changed public contracts and assess consumers | Pass | The log payload is the changed contract; `docs/chat-thread-activity-summary.md` is updated with the new routing and the `reason` field. No code consumer reads these records. |
| `docs/bad-smell.md` | Static imports in production code | Pass | `openrouter-failure` is imported statically alongside the existing `openrouter` import. |
| `docs/bad-smell.md` | Centralized `env()` configuration, no hardcoded URLs/credentials | N/A | No configuration, URL or credential is read or added. |
| `docs/bad-smell.md` | No fallback for a state the contract already rules out | Pass | No fallback branch is added; the change consumes the existing classifier and the existing recovery path. |
| `docs/fallback.md` | Declare every newly introduced fallback in a `Fallbacks` PR section | N/A | The diff introduces no fallback, so no `Fallbacks` section is required (§9: PRs that add none need neither the section nor `Fallbacks: none`). |
| `docs/fallback.md` | Evidence required when removing a fallback | N/A | Nothing removed is a fallback. The deleted branch is a log emission, not a code path that handles an unsupported state; the absorbed failure keeps the identical persistence and recovery. |
| `docs/fallback.md` | No negative tests that only pin removed behavior | Pass | The two absence assertions pin live product behavior — an absorbed failure must not page an operator — and fail when absorption is disabled, as verified below. |
| `docs/fallback.md` | Non-GA feature switch needs no compatibility code | Pass | `threadActivitySummary` is `enabled: false` with staff org hashes, so the mid-rollout window where old instances still emit the record needs no compatibility handling. |
| `docs/testing.md` | Test the contract an external caller relies on, through the production entry point | Pass | Setup via the real send-event and webhook endpoints; verification via the `summarize` endpoint and the production logger's real outbound transport. |
| `docs/testing.md` | Mock external services at their boundary, MSW for HTTP | Pass | OpenRouter and the Axiom ingest endpoint are both MSW-handled; no `vi.mock()` of an internal module. |
| `docs/testing.md` | Avoid fake timers; wait for observable results | Pass | No fake timers; elapsed time uses the pre-existing `advanceRunActivityClockFixture`, and every wait is on an HTTP result. |
| `docs/testing.md` | Cover old/new combinations for changed requests/protocols/persisted state | N/A | No request, protocol, queue payload or persisted-state shape changes; only the emitted diagnostic. |
| `docs/testing/testing-external-behavior.md` | API cases constructed and verified through API endpoints | Pass | See above; no service call or DB assertion is used to construct or verify an outcome. |
| `docs/testing/testing-external-behavior.md` | Exceptions stated and not hidden in generic fixtures | Pass | The single internal-state touch is the shared, pre-existing clock fixture already used by the neighbouring 429 case; it advances an infrastructure clock that no endpoint can move. See observation 3. |
| `docs/testing/api-testing.md` | Cover user-visible HTTP behavior and persisted side effects through follow-up calls | Pass | Retained `messages`/`summaryRevision`, `status: "cooldown"`, bounded `retryAfterMs`, and a follow-up `summarize` proving the provider is not called again. |
| `docs/testing/api-testing.md` | Only mock external services; no `vi.mock()` of internals | Pass | Only MSW handlers and the centralized `context.mocks` registry are used. |
| `docs/testing/api-testing.md` | Uniquely owned identities for shared persistent state | Pass | Each case calls `fixture()`, which provisions a fresh user, org, agent and run; no shared or fixed identity is mutated. |

### Findings

No P0 or P1 findings.

Non-blocking observations:

1. `reason` is attached whenever the failure is an `OpenRouterRequestError`, so a request error that races the deadline can produce `outcome: "timeout"` carrying a `reason`. The level is unchanged (`info`), the field is truthful, and this matches the pre-existing behavior of `providerStatus` on the same branch. No action needed.
2. During rollout, instances still running the previous revision keep emitting the absorbed record. The feature is staff-gated and the change is diagnostic-only, so this needs no compatibility code and resolves itself when the deploy completes.
3. `advanceRunActivityClockFixture` writes the snapshot clock directly. It is the file's existing mechanism for the one state the API cannot construct without waiting real time, and this PR reuses it rather than adding new internal coupling.

### Testing

`turbo/apps/api/src/signals/routes/__tests__/chat-activity-summary.test.ts` — 32/32 pass against a local PostgreSQL 18 instance (`--maxWorkers=1`), through the real route, the MSW provider transport and the production logger with the real Axiom SDK/transport.

New coverage:

- HTTP `503` and HTTP `504`, plus a 200 completion envelope carrying native `UNAVAILABLE` — the local synthetic-502 path — assert no `Activity summary completion` record at **any** level, while `Activity summary attempt` still counts both attempts, the previous summary and revision are retained, `retryAfterMs` stays within the 60 s bound and the stored cooldown suppresses the next provider call.
- HTTP `429` carrying native `401` asserts exactly one `error` with `reason: auth`, `providerStatus: 429` and `cooldownMs: 120000` from `Retry-After`, proving the native code overrides a transient outer status.
- An `INVALID_ARGUMENT` completion envelope (`providerStatus: 502`, `reason: invalid_request`) and HTTP `500` (`reason: unknown`) assert `error`, so not every 5xx is treated as absorbed and an unplaced error stays visible without claiming a cause.
- Every case asserts the provider body string never reaches a record.

The absence assertions were confirmed non-vacuous: forcing `absorbedProviderFailure` to `false` fails both absorption cases, and they pass again once restored.

Also run locally: `tsc` on the `core` and `tests` projects, `oxlint` including the type-aware pass, `eslint --max-warnings 0`, and `prettier`.

Gaps and limitations: no full local Vitest suite and no dev server were run, so full-suite and environment verification belong to the PR pipeline. The historical production 502 records that motivated the issue were logged without their origin or native code, so they cannot be individually attributed to a transient cause; this change makes that attribution available going forward rather than asserting it retroactively.

lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: b98f1b821ff11a1b4cd95a0ea0f2fa26652fccb4
PR_AUTO_REVIEW_VERDICT: LGTM

**Review provenance.** This verdict comes from an independent reviewer that did not implement the change and was given no access to the author's rationale, prior review, or conclusions — only the PR number, the HEAD SHA, and an instruction to follow `REVIEW.md` from `main` and look for defects. Practice revision `9e8b388d344515944065846cc94ddba6ac2e64ab`; documents read: `docs/docs.md`, `docs/bad-smell.md`, `docs/testing.md`, `docs/testing/testing-external-behavior.md`, `docs/fallback.md`.

It is posted from the author's account because that is the only account this automation can write with, so the account identity is **not** evidence of independence; the separation above is. The earlier comment on this PR (`5631900400`, HEAD `5ea9219b`) was an author self-review and should be treated as such — it is superseded by this one.

That independent review returned **Changes Requested** against `5ea9219b` with one P1: the documentation shipped in the diff claimed "configuration" defects record at `error`, while `invalid_or_unconfigured` actually records at `warn`, and the table no longer stated a level for `invalid_or_unconfigured` or `rate_limited`. Commit `b98f1b821f` fixes that documentation only — no production or test file changed — and the re-review below verified the fix against the code rather than accepting the commit message.

Inspected HEAD: `b98f1b821ff11a1b4cd95a0ea0f2fa26652fccb4`.

### Summary

This PR routes `Activity summary completion` records for `provider_failure` by the semantic reason the OpenRouter wrapper already attached to the error (`openRouterFailureReason`) rather than by the outer `providerStatus`: `upstream_timeout` / `provider_unavailable` emit no record at any level, `auth` / `invalid_request` / `unknown` emit at `error`, and `rate_limited` keeps `warn` and gains a `reason` field. The follow-up commit `b98f1b821f` is documentation-only and resolves the single P1 from the previous review; the production logic, classification precedence, privacy boundary and new tests hold up under adversarial verification.

### Review Criteria

| --- | --- | --- | --- |
| REVIEW.md | Inspected revision and delta identified | Pass | HEAD is `b98f1b821f`; `git diff 5ea9219b..b98f1b82` is `1 file changed, 16 insertions(+), 12 deletions(-)` in `docs/chat-thread-activity-summary.md` only. Verified by blob identity, not the commit message: `chat-activity-summary.service.ts` is `c483ed7e35…` at both heads and `chat-activity-summary.test.ts` is `c9f8ea6039…` at both heads; `git diff --name-only 5ea9219b..b98f1b82 -- turbo/` returns zero paths. |
| REVIEW.md | Trace changed behavior through its real entry point and consumers | Pass | `completionLevel` / `absorbedProviderFailure` / `completionRecord` are module-local; the only consumer is `generateSummary` (line 492). `isTransientProviderFailure` is also used by `auxiliary-generation.service.ts:87`, which is unchanged and unaffected — it consumes the shared helper, not the new wrapper. |
| REVIEW.md | Ordering / cancellation / races preserved | Pass | Re-read at `b98f1b82`: the `completionRecord` outcome ternary (`success → cancelled → deadline → provider_failure → invalid_or_unconfigured`) is byte-identical to `main`. `completionLevel` evaluates `expectedOutcome` first, so a `cancelled`/`timeout` that raced an `OpenRouterRequestError` still records `info` and is never dropped. Only `provider_failure` levels changed. |
| REVIEW.md | Failure outcomes and unreachable/misclassified branches | Pass | Checked each reason against `requestFailureReason`: `upstream_timeout` (HTTP 408/504 or allowlisted code 408/504) and `provider_unavailable` (HTTP 502/503 or code 502/503/`UNAVAILABLE`) are both reachable. `network` is genuinely unreachable here — `recordOpenRouterTransportFailure` cannot annotate an `OpenRouterRequestError`, which has `errorCode` rather than `code`/`cause.code` and is not a `TypeError`. `output_truncated` / `unexpected_tool_calls` / `invalid_output` are recorded on plain `Error`s, so they land in `invalid_or_unconfigured` with `reason` absent and cannot reach the new branch. |
| REVIEW.md | Skipping a log must not skip persistence | Pass | `chat-activity-summary.service.ts:492-499` is a level selector with no `else` and no early return; the `abandoned` re-throw, the `activityEnabled` check, the cooldown / claim-clearing transaction (512-541) and the read-back (543-558) are unchanged. Proved externally by the test at line 1013: after an absorbed 503 the stored cooldown still suppresses the next provider call (`inputs` stays at 2) and `retryAfterMs` reads back as 55–60 s. |
| REVIEW.md | Auth / permissions / security | N/A | No auth, ownership, or permission code touched; `requestActivitySummary`'s ownership and feature-switch gates are unchanged. |
| REVIEW.md | Privacy / untrusted-data boundary | Pass | `reason` can only be a member of the `OpenRouterFailureReason` union: `openRouterFailureReason` returns a `WeakMap` value written exclusively by `recordOpenRouterFailure(error, reason)` with literal reasons, otherwise `"unknown"`. `providerStatus` stays numeric. All five new/updated tests assert `JSON.stringify(logs)` does not contain `PRIVATE_PROVIDER_BODY`. |
| REVIEW.md | Persisted data / response validation / compatibility | N/A | No schema, `jsonb(...).$type<...>()`, API contract, or HTTP response-shape change. The only new field is an Axiom log field on a diagnostic stream; `ActivitySummaryResponse` is untouched, so no old/new deployment combination is affected. |
| REVIEW.md | Regression coverage for a rerouted flow | Pass | The rerouted flow is `provider_failure`, and every reason class it can now take is covered end-to-end (absorbed 503, absorbed 504, absorbed envelope `UNAVAILABLE`, `auth`, `invalid_request`, `unknown`, `rate_limited`). `invalid_or_unconfigured` is untested but its emitted level is unchanged by this diff and is now explicitly documented as a separately-owned, non-final bucket — see observation 1. |
| docs/bad-smell.md | §1 no `any`; §2 no suppressions | Pass | Adds `CompletionOutcome`, `CompletionRecord`, and `completionLevel` returning `"info" \| "warn" \| "error" \| null`, and *narrows* `expectedOutcome(outcome: string)` to `CompletionOutcome`. No `any`, no type assertion, no `@ts-expect-error` / `eslint-disable` / `oxlint-disable`. |
| docs/bad-smell.md | §3 error handling at a meaningful boundary | Pass | No new `catch`, no log-and-rethrow, no fabricated default. The classification reuses the reason already recorded at the throw site instead of adding a second parser or classifier. |
| docs/bad-smell.md | §4 document changed public contracts | Pass | Resolved at `b98f1b82`. The diagnostics table (`:104`) now states `warn for invalid_or_unconfigured`, the routing list (`:128`) states `rate_limited` keeps its current `warn`, and the prose (`:143-148`) no longer claims configuration defects record at `error`. Each stated level was verified against `completionLevel` rather than the commit message — all four match. |
| docs/bad-smell.md | §5 static imports; §6 no hardcoded URLs/config; §7 fallback default | Pass | New imports from `../external/openrouter-failure` are static. No URL or `env()` change. No tolerance branch added. |
| docs/fallback.md | §9 every new fallback declared in a `Fallbacks` section | N/A | The diff introduces no fallback. `completionLevel` classifies a reachable, currently-produced state; it is not a tolerance path for a state the contract rules out, and no existing fallback is removed. Per §9's final paragraph, no `Fallbacks` section is required. |
| docs/fallback.md | §1 no negative tests against removed code | Pass | The two absence assertions (`toStrictEqual([])`) pin current live behavior for a reachable provider response; they do not assert that deleted code stays deleted, so they are not tombstones. |
| docs/fallback.md | §10 evidence when removing a fallback | N/A | No fallback branch, contract entry, or fallback test is removed. |
| docs/testing.md | Coverage for new behavior and regressions | Pass | Four new cases cover absorbed HTTP 503, absorbed HTTP 504, absorbed envelope `UNAVAILABLE` (the synthetic-502 path), native `401` inside a `429` wrapper, envelope `INVALID_ARGUMENT`, and HTTP 500 → `unknown`; the pre-existing 429 case is extended additively with `reason`. |
| docs/testing.md | Mock at the boundary; real internal code and DB; no fake timers | Pass | Provider mocked at `https://openrouter.ai/api/v1/chat/completions` via `server.use(http.post(...))`; Axiom ingestion captured at its real HTTP endpoint through the real SDK/transport; time advanced with `advanceRunActivityClockFixture`, never `vi.useFakeTimers`; `testContext()` owns cleanup. |
| docs/testing/testing-external-behavior.md | API tests construct and verify through production endpoints | Pass | Every new case drives `summarize()` → `chatThreadActivitySummaryRoutes` and asserts on the HTTP response (`status`, `messages`, `summaryRevision`, `retryAfterMs`) plus captured outbound telemetry. No DB row is read or written directly, no service is called, and the test file imports only `lib/log` and test fixtures. |

### Findings

No blocking defects. The P1 raised against `5ea9219b` is resolved.

**Resolution of the previous P1, verified rather than accepted on assertion.** At `b98f1b82`:

- `:104` — the Level column reads `info for success/timeout/cancelled; warn for invalid_or_unconfigured; see the provider-failure routing below`. Matches `completionLevel`: `expectedOutcome` → `info`; `reason === undefined` — the only way to reach `invalid_or_unconfigured`, since that outcome requires `request` to be absent — → `warn`.
- `:128-129` — `rate_limited` keeps its current `warn`. Matches `completionLevel`'s `reason === "rate_limited" → "warn"`.
- `:143-148` — the offending sentence is replaced with "Request-contract and credential defects reported by the provider are real failures and record at `error`", matching `invalid_request` / `auth` / `unknown` falling through to `error`, plus an explicit statement that `invalid_or_unconfigured` keeps `warn`.

Every level the document now states is one `completionLevel` actually produces; none is omitted.

**No new inaccuracy introduced.** The new example list — "`invalid_or_unconfigured` covers every other unusable result — an absent API key, malformed or empty model output, a truncated completion" — was checked against the real paths:

- *Absent API key*: `generateText` returns `null` when `OPENROUTER_API_KEY` is unset (`openrouter.ts:432-435`); `activityPhrases(null)` returns `null`, so `phrase` is `null`, `failure` is `null`, `request` is `undefined` → `invalid_or_unconfigured`.
- *Malformed or empty model output*: `OpenRouter returned invalid content` / `empty content` / `no choices` / `invalid JSON` all throw plain `Error`s, and a well-formed completion that `activityPhrases` rejects yields `phrases === null`. Both reach `invalid_or_unconfigured`.
- *Truncated completion*: this service passes only `{ reasoning: { effort: "low" } }`, so `acceptTruncatedText` is `false`; `finish_reason: "length"` throws a plain `Error` tagged `output_truncated`, which is not an `OpenRouterRequestError` → `invalid_or_unconfigured`.

The new prose also does not contradict the aggregation caveat (`:155-160`): `invalid_or_unconfigured` records at `warn`, so the `invalid=countif(...)` line still counts a stream that exists, and the "absorbed failures are absent by design" caveat remains scoped to `provider_failure`.

### Non-blocking observations

1. **`invalid_or_unconfigured` has no test at any level, and that is correctly handled here.** `completionLevel` now owns that branch, so a level test would be cheap — but the level is unchanged by this diff, and the constraint on this change forbids entrenching `warn` as a final contract for a bucket another owner is still resolving. The document says exactly that twice (`:128-129`, `:147-148`), so the gap is accepted and documented rather than silent. A test asserting `warn` here would create work for the follow-up that changes it.
2. **A total provider outage becomes unobservable from this record set.** With `upstream_timeout` and `provider_unavailable` silent, a sustained 100 % OpenRouter failure produces `Activity summary attempt` records with no matching completion and nothing else, and the doc explicitly tells operators not to reconcile completions against attempts. Unlike `auxiliary-generation.service.ts`, this service calls `generateText` directly and emits no `auxiliary_generation_result` event, so there is no second stream. The decision is deliberate, issue-scoped and documented; worth a follow-up that keeps a countable signal.
3. **`absorbedProviderFailure` accepts `network`, which its own comment calls unreachable here.** The comment was confirmed correct. Reusing `isTransientProviderFailure` rather than re-enumerating is the right call under `docs/bad-smell.md`; noting only that the branch's correctness rests on that wrapper invariant.
4. **Minor prose looseness.** "covers every other unusable result" at `:145` is scoped by juxtaposition with the preceding sentence rather than stated explicitly; the three examples and the exhaustive routing list disambiguate it. Not a defect.
5. **Weak-but-harmless assertion.** In the envelope/timeout test, `expect(unavailable.messages).toStrictEqual([])` would pass regardless of retention, because no summary was ever stored in that fixture. Retention is genuinely proved by the first new test, so this is redundancy, not a false pass.

### Testing

**What the diff adds** (unchanged from `5ea9219b`; blob `c9f8ea6039…` verified identical at both heads). Four `it` blocks in `turbo/apps/api/src/signals/routes/__tests__/chat-activity-summary.test.ts`, all driven through the production route via `summarize()`, with MSW at the OpenRouter and Axiom ingestion boundaries and the real logger and Axiom SDK/transport, plus one additive `reason: "rate_limited"` expectation on the pre-existing 429 case.

**Verified independently from the source, not from the PR body:**

- **Every fixture classifies as claimed.** HTTP 503 → `provider_unavailable`; HTTP 504 → `upstream_timeout` (checked against `requestFailureReason`'s order: `auth` → `invalid_request` → `rate_limited` → `[408,504]` → `[502,503,"UNAVAILABLE"]`); `finish_reason: "error"` with native `UNAVAILABLE` → synthetic 502 with `origin: "completion"` and `httpStatus` passed as `undefined`, matched on the code → `provider_unavailable`; 200 envelope with `INVALID_ARGUMENT` and no `choices` → `invalid_request` at `providerStatus: 502`; HTTP 500 with code 500 → matches no candidate set → `unknown` at `error`. HTTP 429 whose body carries code 401 → `auth`, because `safeErrorCode` allowlists numeric `401` onto `errorCode` and `requestFailureReason` tests `[401,403]` before the `429` set. The native-precedence claim is demonstrated, not merely asserted.
- **Cooldown arithmetic matches the assertions.** `min(300_000, max(60_000, retryAfterMs ?? 0))` gives 120 000 ms for `Retry-After: 120` and 60 000 ms otherwise; the `retryAfterMs` bounds are read back through the HTTP response, i.e. externally observable proof that the cooldown write happened even when no record was emitted.
- **No assertion is vacuous or tautological.** Reverting the production change would fail every new case: the two absorption tests filter for completion records and expect `[]`, and the reverted code emits a `warn` `provider_failure` record in both; the `auth`, `invalid_request` and `unknown` cases assert `level: "error"` with an exact `fields` object including `reason`, neither of which the reverted code produces; the updated 429 case uses an exact `fields` object, so the added `reason` key is load-bearing there too. The first absorption test additionally proves the skipped log did not skip persistence.
- **Conventions.** Boundary mocking with MSW, real DB and real internal code, no fake timers, assertions through captured ingestion requests rather than a logger spy, and the `debug`-is-filtered assertion retained.

**Remaining gaps (accepted, not blocking).**

1. `invalid_or_unconfigured` is untested at any level. Out of scope per the constraint on this change and now explicitly documented as a separately-owned, non-final level.
2. No case covers an `OpenRouterRequestError` that races cancellation or the deadline. Inspection confirms `expectedOutcome` is evaluated first, so such a record still logs at `info` and is never dropped, but that precedence is unguarded by a test.
3. Full Vitest was not run locally, per the repository's stated constraint; the PR pipeline owns full-suite, type-check and lint verification.

lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: 26d1658733e71818e81a226abb06a03a3e2fab8d
PR_AUTO_REVIEW_VERDICT: LGTM

**Review provenance.** This verdict comes from an independent reviewer that did not write the change and was given no access to the author's reasoning — only the PR number, the HEAD SHA, the two integration constraints to attack (do not restore #33522's deleted `rate_limited` diagnostic; do not move #33362's `invalid_or_unconfigured`), and an instruction to follow `REVIEW.md` from `main` and find defects. Practice revision `da440bff4d7a03293bdce027bf09d938a4a58f9a`.

It is posted from the author's account because that is the only account this automation can write with, so the account identity is **not** evidence of independence; the separation above is. This review covers **only** HEAD `26d1658733e71818e81a226abb06a03a3e2fab8d`. The earlier LGTM (`5633038720`) belongs to the superseded HEAD `b98f1b821f` and does not transfer — that HEAD was rebased away after #33522 merged.

One correction to the reviewer's non-blocking observation 3: the "34/34" figure is vitest's test count and is accurate. The file has 27 `it(` literals, but `it.each` at lines 731 and 771 expand to 3 and 6 cases, giving 25 + 9 = 34. Both numbers are right; they measure different things.

### Summary

This PR widens `chat-activity-summary.service.ts`'s absorbed set from `rate_limited` alone to every reason `isTransientProviderFailure` accepts, and routes the remaining `provider_failure` reasons (`auth`, `invalid_request`, `unknown`) from `warn` to `error`. Inspected at HEAD `26d1658733e71818e81a226abb06a03a3e2fab8d`, whose merge-base is exactly the current `main` tip `da440bff4d` — a clean rebase with three changed files and no residue from the sibling PR. I verified independently that #33522's deleted `rate_limited` diagnostic is not restored at any level, that `invalid_or_unconfigured` still records identically, and that success / cancellation / deadline precedence is byte-identical to `main`.

### Review Criteria

| --- | --- | --- | --- |
| REVIEW.md | Trace changed behavior through its real entry point and consumers | Pass | `generateSummary` is the only caller of `absorbedCompletion`/`realProviderFailure`; both are file-local (`grep` shows no other consumers). Entry point `requestActivitySummary` → route, exercised by the tests through `summarize()`. |
| REVIEW.md | Races, cancellation, ordering, failure outcomes | Pass | `completionRecord` still resolves `phrase → abandoned → deadlineReached → request` in that order (line 407-415), unchanged from `main`. `providerStatus`/`reason` are attached whenever `request` exists, *independently of outcome*, but both new predicates gate on `outcome === "provider_failure"`, so a request error racing `deadline.aborted` stays `timeout`→`info` and a caller abort stays `cancelled`→`info`. No misroute to `error` is reachable. |
| REVIEW.md | Constraint: `rate_limited` diagnostic not restored | Pass | `isTransientProviderFailure` returns `true` for `rate_limited` (openrouter-failure.ts:49-58), so `absorbedCompletion` still suppresses it at every level. The two pre-existing #33522 tests (lines 982, 1045) assert `providerFailureRecords === []` and still hold. |
| REVIEW.md | Constraint: `invalid_or_unconfigured` unchanged | Pass | That outcome requires `request === undefined`, so `realProviderFailure` is false by its first conjunct; `expectedOutcome` is false; it falls to the same `log.warn` as on `main`. Doc explicitly re-states it keeps `warn` under #33362's ownership. |
| REVIEW.md | Persistence / cooldown / Retry-After / retention / claim / reread untouched | Pass | The diff touches only the `if` chain at line 486-494. The cooldown write (line 523), claim-fencing predicates (527-534), `MAX_COOLDOWN_MS` bound (402-405) and the final reread (538-553) are outside the diff and byte-identical to `main`. |
| REVIEW.md | Meaningful coverage for new behavior | Pass | Three new cases cover absorbed `provider_unavailable` (HTTP 503), absorbed `provider_unavailable` via synthetic 502 + `upstream_timeout` (HTTP 504), and reported `auth`/`invalid_request`/`unknown` at `error`. |
| REVIEW.md | Regression coverage for the rerouted path | Pass | The pre-existing #33522 case at line 1087 still pins `rate_limited` absorption and the native-precedence rule; the new case at 1300 pins the level #33522 deliberately left unasserted. |
| REVIEW.md | Persisted data / API contract / compatibility | N/A | No schema, `jsonb().$type<>()`, contract, request/response shape, queue payload, or persisted-state change. The only externally visible change is the Axiom `level` string on one diagnostic; `ActivitySummaryResponse` is untouched. |
| REVIEW.md | Optimistic/persistent event reconciliation | N/A | No event-sourced projection is involved; `runActivitySnapshots` writes are unchanged. |
| docs/bad-smell.md | §1 no `any`, §2 no suppressions | Pass | Diff adds no `any`, no assertion, no `eslint-disable`/`@ts-expect-error`. The only assertion-like token is `as const` on a readonly test fixture array. |
| docs/bad-smell.md | §5 static imports | Pass | `isTransientProviderFailure` added as a static named import to the existing `../external/openrouter-failure` block (line 34); no dynamic import added. |
| docs/bad-smell.md | §3 error handling, §7 fallback patterns | Pass | No `catch`, no log-and-rethrow, no fabricated default introduced. The `reason !== undefined` conjunct is a TypeScript narrowing for an optional field, not a value-inventing `??` chain. |
| docs/bad-smell.md | §4 interface changes documented | Pass | `docs/chat-thread-activity-summary.md` is updated in the same diff with the new level, the widened absorbed set, and the `unknown` semantics. |
| docs/fallback.md | §9 every introduced fallback declared | Pass | The diff introduces no fallback: it narrows a diagnostic branch and raises a level. Per §9, a PR that adds none needs neither the section nor `Fallbacks: none`. |
| docs/fallback.md | §1 no tombstone tests | Pass | All three new tests assert live behavior (absorbed vs. reported), not that removed code stays removed. |
| docs/testing.md | Real entry point, real DB, MSW at the boundary | Pass | Tests call the production route via `summarize()`, use the real Postgres through `testContext()`, mock OpenRouter with MSW `server.use`, and capture the real Axiom SDK's outbound ingest request. No `vi.mock` of internal modules, no fake timers. |
| docs/testing.md | No duplicate cases restating configuration | Pass | The three cases cover three distinct classifier reasons plus the synthetic-502 path; none restates the merged `rate_limited` cases. |
| docs/testing/testing-external-behavior.md | Assert externally observable state, not internals | Pass | Assertions are on HTTP response bodies (`messages`, `status`, `summaryRevision`, `retryAfterMs`), on outbound provider requests received by MSW (`inputs`), and on the telemetry actually shipped to Axiom over HTTP. No service return value, DB row read, or spy-call count. |
| docs/testing/testing-external-behavior.md | Lint: no DB-schema/service imports in API tests | Pass | The new cases add no imports; the file imports only contracts, routes, helpers, `lib/log`, and the pre-existing `test-fixtures/run-activity` clock fixture. |

### Findings

No P0 or P1 defects. Every claim attacked held up under independent tracing.

**Verification detail (all confirmed correct):**

- **Partition.** `outcome === "provider_failure"` is produced only when `failure instanceof OpenRouterRequestError` (service line 401, 413), and that same `request` truthiness attaches `reason` (line 418-423). `openRouterFailureReason` never returns `undefined` — it defaults to `"unknown"`. So on this branch `reason` is always defined, and `absorbedCompletion` / `realProviderFailure` partition it exactly: no gap, no overlap. Reachable reasons are `{auth, invalid_request, rate_limited, upstream_timeout, provider_unavailable, unknown}` — the complete image of `requestFailureReason`. `output_truncated`, `unexpected_tool_calls` and `invalid_output` are recorded on plain `Error`s and land in `invalid_or_unconfigured`, never here.
- **`network` unreachability.** Confirmed, for a subtler reason than the comment states: `ensureOpenRouterResponseOk` *is* wrapped in `onRejection(..., recordOpenRouterTransportFailure)`, so the transport classifier does run over the `OpenRouterRequestError`. It cannot overwrite the reason because that class exposes `errorCode`, not `code`, has no `cause`, and is not a `TypeError` — every branch misses. The conclusion is right.
- **Fixtures produce the claimed reasons.** Traced `requestFailureReason`'s ordering by hand. HTTP 503 → `provider_unavailable`. 200 envelope `finish_reason:"error"` + `UNAVAILABLE` → `origin:"completion"`, `httpStatus` `undefined`, matched on code → `provider_unavailable` at wrapper 502. HTTP 504 → `upstream_timeout`. HTTP 429 + `code:401` → `auth`, because the auth check is first and matches on `code`, with `providerStatus` staying `429`. 200 envelope `INVALID_ARGUMENT`, no `choices` → `invalid_request` at synthetic `502`. HTTP 500 → in no candidate list → `unknown`. Cooldowns: `Retry-After: 120` → `min(300_000, max(60_000, 120_000)) = 120_000`; the others → the `60_000` floor. All six expectations are correct.
- **Non-vacuity.** Tests 1 and 2 assert `providerFailureRecords`/`completionRecords`/`operatorRecords` are empty; reverting `absorbedCompletion` to `rate_limited` only would emit `warn` records and fail them. Test 3's `level: "error"` fails under a revert of the `log.error` branch. Test 3's `fields:` is a *strict* deep-equal, so it also proves no extra field — including no provider body — is attached.
- **Privacy.** `reason` can only be one of the string literals returned by `requestFailureReason`; the raw provider `code` is read for classification and discarded. `providerStatus` is a number. Each new test asserts `PRIVATE_PROVIDER_BODY` never appears in the serialized log stream.
- **No alerting side effect.** `log.error` routes to `console.error` plus `alog.error`; there is no Sentry capture or configured alert keyed on `level === "error"` in the repo, so this is a severity relabel in the dataset only.
- **Doc/code match.** Every level the table states is reachable and correct, and the claim that `provider` in the sample query counts only reported failures is exactly true: emitted `provider_failure` records are precisely the `realProviderFailure` set.

**Non-blocking observations** (no action required for this verdict):

1. **Stale doc sentence, `docs/chat-thread-activity-summary.md:93.** The paragraph introducing the table still says records use "`info` … and `warn` …", while the table directly beneath now has three levels. One clause would fix it.
2. **`unknown` is the widest bucket and now the loudest.** `requestFailureReason` has no candidate for HTTP `500`, `529`, `522`, `524`, and `safeErrorCode` drops native codes outside its allowlist, so a 200 envelope carrying `DEADLINE_EXCEEDED`, `INTERNAL` or `OVERLOADED` becomes `unknown`. Those move from `warn` to `error`. The decision is explicit, documented and pinned by a test, so it is owned rather than accidental — but it is the one place the change increases severity rather than reducing noise, and it is worth revisiting once production shows what `unknown` contains.
3. **PR body validation count.** (Reconciled by the author above: `it.each` expansion accounts for the difference; 34 is vitest's count and is correct.)
4. **Both predicates re-test `outcome` and guard an unreachable `undefined`.** Required only because `CompletionRecord.reason` is optional; a discriminated union on `outcome` would delete both guards and make the partition a type-level fact. Not a fallback under `docs/fallback.md` §3 — it fabricates no value, and the unreachable residual falls to `warn`, not to silence.

### Testing

**Coverage added.** Three integration cases through the production route, the real database, the MSW provider transport, and the production logger with the real Axiom SDK: (a) HTTP 503 absorbed while the prior summary, its revision, the bounded `retryAfterMs`, the stored cooldown's suppression of the next provider call, and both attempt records survive; (b) a native `UNAVAILABLE` inside a 200 envelope plus HTTP 504, both absorbed with zero completion records; (c) `auth` / `invalid_request` / `unknown` each emitting exactly one `error` record with a strictly-matched field set.

**Verified rather than trusted.** The branch is a clean rebase (merge-base equals the `main` tip; three files changed). `rate_limited` absorption is preserved through `isTransientProviderFailure`, not reintroduced. `invalid_or_unconfigured` cannot enter the new `error` branch. Success/cancel/deadline precedence is unchanged, including a provider request error racing the deadline. Every fixture's reason, `providerStatus` and `cooldownMs` were traced by hand through `requestFailureReason`'s candidate lists, `safeErrorCode`'s allowlist and `retryAfterDelay`'s clamp. Clock advances (121 s, 61 s) clear the cooldowns they must and stay inside the 24 h retention. `rootEventFields` cannot inject anything into `fields` for these records, so test 3's strict field assertion is sound.

**Material gaps.**

- `invalid_or_unconfigured` has no level assertion anywhere in the suite. This PR restructures the exact `if`/`else` that emits it, so its `warn` is protected by code reading alone. The gap predates this PR and the bucket belongs to #33362, so it is not blocking, but one `level: "warn"` assertion would pin the boundary this PR was required not to cross.
- `network` is asserted unreachable in a comment and the docs but has no test; appropriate, since constructing a transport rejection that still reaches this branch is impossible by construction.
- Per repository policy no full local Vitest suite or dev server was run; the PR pipeline owns full-suite verification. The reviewer's non-vacuity conclusions come from tracing the production code paths each assertion depends on, plus the author's recorded revert experiment.

PR #33557, opened 2026-09-11
fix(platform): let each chat thread list catch-up own its lifetime

e7h4n:
## Problem

`resolveThreadMeta$` borrowed the in-flight list synchronization from a global
`chatThreadEventSyncBarrier$` entry to save one incremental request. The
borrowed promise carried the `AbortSignal` of whichever caller started it.

When the welcome thread action started a list catch-up and the user dismissed
Settings before it finished, the action signal aborted. A cold reader that had
already borrowed that promise finished with the other owner's `AbortError`, the
metadata 404 fallback could not reach the canonical event stream, and the pane
stayed without its history. Whether the reader observed the cancelled entry or
a cleared one depended on microtask ordering, so it surfaced as a flaky test
rather than a hard failure — most recently in
`welcome-thread-sync-overlap.test.tsx`, "An older list refresh cannot finish a
newer welcome reader".

## Change

Remove the barrier. Every caller runs its own catch-up under its own signal, so
a caller can no longer inherit a cancellation it did not request. The shared
mutable promise holder and its `withCleanup` ownership-release go away with it.

The lost deduplication is cheap: each synchronization reads `cache-only` first
and only falls back to a network `catch-up` when the cache is behind the current
cursor, and the Worker does not coalesce concurrent chat-thread-event catch-ups
either. The worst case is one extra incremental `sinceSeqId` tail request inside
a narrow overlap window.

## Verification

- `welcome-thread-sync-overlap`, `welcome-thread-sync`, `welcome-thread`: 16 tests pass.
- `chat-list-cache`, `chat-metadata`, `chat-offline-cache`, `sidebar`,
  `settings-dialog`, `platform-direct-bridge`, `shared-database-browser`: 91 tests pass.
- `tsconfig.production.json` type check, ESLint, oxlint, and Prettier pass.

No new test is added: the existing overlap suite already pins both contracts —
an older refresh must not finish a newer reader, and leaving one reader must not
cancel synchronization for the next one. The local reproduction of the flake
remained order-dependent, so this fix is argued structurally: after the change
there is no code path by which a reader can await work owned by another
caller's signal.

### DIFF
--- a/turbo/apps/platform/src/signals/chat-page/chat-thread-event-sourcing.ts
+++ b/turbo/apps/platform/src/signals/chat-page/chat-thread-event-sourcing.ts
@@ -16,7 +16,7 @@ import {
 import { activeRoute$ } from "../active-route.ts";
 import { apiClient$ } from "../api-client.ts";
 import { updateDocumentTitle$ } from "../document-title.ts";
-import { rootSignal$, rootVersion$ } from "../root-signal.ts";
+import { rootSignal$ } from "../root-signal.ts";
 import { pathParams$ } from "../route.ts";
 import {
   createChildAbortController,
@@ -161,17 +161,6 @@ const initialLocalChatThreadEventsLoaded$ = computed((get) => {
   return get(initialLocalChatThreadEventsLoadedDeferred$).promise;
 });
 
-interface ChatThreadEventSyncBarrier {
-  work: Promise<void> | null;
-}
-
-const chatThreadEventSyncBarrier$ = computed(
-  (get): ChatThreadEventSyncBarrier => {
-    get(rootVersion$);
-    return { work: null };
-  },
-);
-
 const optimisticChatThreadCreateIds$ = computed((get): ReadonlySet<string> => {
   return new Set(
     get(optimisticChatThreadEventsState$).flatMap((event) => {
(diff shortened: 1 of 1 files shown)


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: 965e6bacbe6ee8ef83533751086883fd3abeb507
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `965e6bacbe6ee8ef83533751086883fd3abeb507`
Practice revision: `e7b5d4f0cdf78291d3323bec94824197be83dfa5` (`main`)

### Summary

Removes `chatThreadEventSyncBarrier$` so `resolveThreadMeta$` no longer borrows another caller's in-flight chat-thread list synchronization. Every caller now runs its own catch-up under its own `AbortSignal`, which eliminates the path where a cold reader inherited an `AbortError` owned by a different lifecycle (for example, dismissing Settings during welcome-thread creation) and rendered a pane without its history.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers; check races, cancellation, and ordering | Pass | Only consumer of the removed barrier was `resolveThreadMeta$` (`chat-thread-event-sourcing.ts:493`), reached from `chat-page-setup.ts:44,62` via the `/chats/:threadId` route setup. Both remaining callers of `syncEventDrivenChatThreads$` (`welcome-thread.ts:25`, `shared-database-browser.ts:149`) pass an owned signal and await the returned promise. |
| `REVIEW.md` | Distinguish introduced defects from unchanged code | Pass | Non-monotonic commit in `applySharedChatThreadEventResult$` is unchanged code; the barrier never serialized result application, only which promise a reader awaited. Recorded as a non-blocking observation, not a finding. |
| `REVIEW.md` | Confirm existing coverage still protects a refactor | Pass | `welcome-thread-sync-overlap.test.tsx` pins both affected contracts (an older refresh must not finish a newer reader; leaving a reader must not cancel synchronization for the next one). See Testing for the caveat. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions, no weakened rules | Pass | Diff adds no assertions or suppression comments; it removes code and narrows the `root-signal.ts` import to the symbol still used. |
| `docs/bad-smell.md` | Catch errors only at a boundary that meaningfully handles them; preserve real recovery | Pass | The removed `withCleanup` only nulled the barrier slot; it performed no recovery, released no resource, and swallowed no error. Rejection propagation to the caller is unchanged. |
| `docs/bad-smell.md` | Interface changes documented and consumers assessed | N/A | `syncSharedEventDrivenChatThreads$` keeps its exported alias, signature, and return type; no public contract changes. |
| `docs/fallback.md` | New fallbacks declared; removal backed by evidence | N/A | No `??`/default branch, retired-protocol path, or rollout switch is added or removed. The barrier was a request-deduplication cache, not a fallback, so no `Fallbacks` section is required. |
| `docs/fallback.md` | No negative tests pinning removed code | Pass | No test was added or deleted; nothing asserts the barrier's absence. |
| `docs/effect.md` | Make cancellation explicit: every async command has a signal with a clear owner | Pass | `syncSharedEventDrivenChatThreads$` receives the caller's signal and returns its promise; `resolveThreadMeta$` starts its own refresh under `rootSignal$` and bounds its own wait through `waitForSharedWork` (`chat-thread-event-sourcing.ts:400`). |
| `docs/effect.md` | Compose commands through commands; no `detach()` in the signals layer | Pass | The change is command-to-command composition only; no `detach()`, `void`, or empty catch is introduced. |
| `docs/effect.md` | Render purity and owner selection | N/A | No React component, render path, `onRef`, or effect is touched. |
| `.claude/skills/ccstate/SKILL.md` + `references/lifecycle.md` | Every async operation has an owner and an eventual abort path; do not conceal a missing owner | Pass | The reader's refresh is owned by `rootSignal$`, and its rejection is always consumed: `lookupEventStreamThreadMeta$` calls `waitForSharedWork` synchronously, which attaches a `Promise.race` handler in the same tick the work is created. |
| `docs/cache.md` (RC-3, RC-4) | No unbounded-lifetime cache; state owned by the narrowest correct lifecycle | Pass | Deletes a `rootVersion$`-scoped mutable promise holder. Nothing gains a longer lifetime; the reader's work is created per call rather than retained. |
| `docs/cache.md` (RC-5, RC-10) | No duplicate mutable sources of truth or duplicate request-lifecycle state | Pass | `chatThreadEventState$` remains the single projection; no sidecar status state is introduced. |
| `docs/event-sourcing.md` | Persistent events authoritative; optimistic events reconciled by shared ID and never imperatively rolled back on failure | Pass | `applySharedChatThreadEventResult$` and `reconcileOptimisticChatThreadEvents$` are untouched; no failure/cancellation path removes optimistic events. |
| `docs/testing.md` + `docs/testing/app-testing.md` | Platform coverage enters through the real Router and asserts page-visible state | Pass | Verification used the existing `setupPage`-based page tests; no component-only or store-level test was added. |
| `docs/testing/testing-external-behavior.md` | Assert externally observable state, not internal implementation | Pass | Existing assertions read thread titles, headings, and rendered history; the diff removes an internal symbol no test referenced. |
| `docs/deployment-compatibility.md` | Old/new combinations for changed requests, protocols, or persisted state | N/A | Not read and not applicable: no request shape, SharedWorker protocol message, schema, or persisted-state change. Request *timing* changes; request *contracts* do not. |

### Findings

No P0 or P1 findings.

**Non-blocking observations**

1. `applySharedChatThreadEventResult$` (`chat-thread-event-sourcing.ts:193`) commits `chatThreadEventState$` wholesale with no monotonic guard on `latestSeqId`, so a synchronization that started earlier but completes later can overwrite a newer projection until the next refresh. This is pre-existing — the barrier only decided which promise a reader awaited and never serialized result application, and two explicit callers (realtime invalidation plus a mutation catch-up) could already race. This PR adds at most one more concurrent caller inside the cold-reader window, so the exposure widens slightly without being introduced here. Worth a separate change that rejects a commit whose `latestSeqId` is behind current state.

2. The removed deduplication cost is bounded but not zero: each synchronization still issues a `cache-only` query first and only escalates to `catch-up` when the cache is behind `currentSeqId`, and the SharedWorker does not coalesce concurrent `chat-thread-event` catch-ups either (`worker-runtime.ts:624`), so the worst case is one extra incremental `sinceSeqId` tail request during an overlap. The cold-reader path itself is rare: `resolveThreadMeta$` short-circuits on the in-memory and local-cache hits and only reaches this code for a thread absent from the cached list.

### Testing

Existing coverage is the right coverage here, and it stays green:

- `welcome-thread-sync-overlap.test.tsx`, `welcome-thread-sync.test.tsx`, `welcome-thread.test.tsx` — 16 tests pass.
- `chat-list-cache`, `chat-metadata`, `chat-offline-cache`, `sidebar`, `settings-dialog`, `platform-direct-bridge`, `shared-database-browser` — 91 tests pass.
- `tsconfig.production.json` type check, ESLint, oxlint, and Prettier pass on the changed file.

Material gap, stated explicitly rather than papered over: the two overlap tests encode the correct contracts but do not *discriminate* this fix, because on the old code they also pass whenever the barrier slot happens to be cleared before the reader reads it — that non-determinism is the bug being removed. Two attempts to build a deterministic page-boundary regression test failed for boundary reasons, not for lack of the defect: the SharedWorker still persists the events of a cancelled App-side catch-up, so a later reader recovers them from IndexedDB and the page-visible outcome converges either way. Constructing the failing interleaving would require reaching inside the sync coordinator, which `docs/testing/testing-external-behavior.md` forbids for Platform. The fix is therefore argued structurally: after this change no code path lets a reader await work owned by another caller's signal. That argument is verifiable by reading the diff, and the retained tests still fail if a future change reintroduces shared ownership and gets the interleaving wrong.

PR #33327, opened 2026-09-10
refactor(platform): own welcome thread creation by the settings action signal

e7h4n:
## Summary

`welcomeThreadAction$` was a `computed` that captured the Settings action signal and the Clerk identity in a closure, re-checked both inside its `command`, and re-implemented the org-switch listener that `auth.ts` already owns. It also kept a retry-reuse `clientThreadId`, an in-flight guard, and a manual controller with `withCleanup`.

This PR replaces it with one module-level `createWelcomeThread$` command that receives its `AbortSignal` from the caller, matching the other Settings actions such as `leaveOrg$` and `initProfileName$`.

- The card reads `settingsActionSignal$` with `useGet` and passes it into the command, the same way the other Settings views pass `settingsDialogSignal$`. That signal is already `resetSignal(pageSignal)`, so leaving the page and dismissing the dialog both cancel the request and the list catch-up. No closure-captured signal and no `no-computed-signal` suppression remain.
- Identity changes stay owned by `setupClerk$` (user change bumps `reload$`) and `watchOrgSwitch$` (org switch reloads the page). The command no longer compares `clerk.user`/`clerk.organization` or installs its own `clerk.addListener`.
- Each click sends a fresh `clientThreadId`. The retry-reuse state, the in-flight guard, the manual controller cleanup, and the `CONFLICT`-specific copy are removed; a 409 `CONFLICT` now shows the generic failure message.
- Deleted the tests that pinned the removed behavior: same-UUID retry after a lost response, workspace switch cancelling the pending action, and user/workspace change after dismissal cancelling the committed sync. The gate, page-leave, dismissal, 4xx, and committed-sync tests remain.
- Removed `docs/welcome-chat-threads.md` and its `docs/docs.md` entry.

Two nearby commands get the same signal-parameter treatment:

- `shared-thread-rich-content.ts`: `retry$` takes `signal: AbortSignal` from the view instead of reading the factory `ownerSignal` from its closure. The view passes `pageSignal$`, which is the same signal the page setup used to create the factory.
- `desktop-auth.ts`: `selectOrganization$`, `reopen$`, and `retry$` drop the `lifetime.throwIfAborted()` double-check. `lifetime` is the page-setup signal, and the view already passes `pageSignal$` into each of them.

Left for a separate decision (older, and owned by others): the identity listener in `feature-switch.ts` `refreshFeatureSwitchState$` and the post-await identity re-check in `onboard-guard.ts`.

## Verification

Run locally on the platform workspace, sequentially:

[...]

PR #33564, opened 2026-09-11
fix(platform): keep unread indicators readable when chat warming fails

e7h4n:
## Problem

`readWorkerChatThreadIndicators$` awaited the throttled ChatEvent catch-up before returning indicators:

```ts
const indicators = await source;
await set(catchUpChatEvent$);   // 1000ms-throttled batch prefetch
return indicators;
```

A tab asking for unread indicators therefore inherited both the pacing and the failure mode of a prefetch it does not depend on:

- **A failed catch-up rejected every tab's indicator read.** One 500 from the batch `catch-up` endpoint blanked sidebar unread state for all tabs. The previous test suite asserted this as intended behaviour.
- **The 1000ms warming throttle sat on the reader's critical path.** Measured on `sidebar.test.tsx > Mark all of an agent's chats read`, the wait after "Mark all read" was 623ms, of which 614ms was throttle residue — about 48% of that test's total runtime, and exactly equal to Testing Library's 1000ms `asyncUtilTimeout`, leaving no margin.

## Why warming is not a data dependency

Indicators carry no ChatEvent data — they are `agents`/`threads` unread maps. Both thread readers already fall back to their own catch-up when the cache is behind:

- `chat-thread-event-sourcing.ts` reads `cache-only`, then `catch-up` when the cached tail is stale.
- `chat-event-storage-signals.ts` does the same.

So warming is a head start, not a correctness requirement.

## Change

Move warming to the realtime refresh that observed the change:

- `readWorkerChatThreadIndicators$` returns the indicator fetch directly.
- `refreshWorkerChatIndicators$` awaits `catchUpChatEvent$` after refreshing indicators.

The `threadListChanged` subscription uses `runOnSubscribe: true`, so the first warming still runs on connect before any realtime event arrives. The throttle is unchanged at 1000ms and now paces warming bursts rather than readers. Warming stays awaited by a real owner — the Ably loop body — so no `detach()` enters the signals layer.

This also removes the now-dead `workerChatThreadIndicatorsCache$` / `loadWorkerChatThreadIndicators$` pair, whose only purpose was deduplicating the awaited catch-up per indicator source.

## Tests

Four tests pinned the old ordering and are replaced:

| Removed | Why |
| --- | --- |
| `Continue warming unread chats when indicator catch-up reads time 'before'/'across'/'at' the throttle boundary` | Asserted `throttleCommand`'s own 999/1000ms arithmetic through the bridge, with a hand-driven clock. `signals/__tests__/command-scheduling.test.ts` already covers that boundary directly. |
| `Recover indicator catch-up after a shared trailing request fails` | Asserted that a failed catch-up makes `getComputed("chat-thread-indicators")` reject with a 500 — the bug. |

Replacements assert what callers actually rely on, synchronising on the Worker's `chat-event` invalidation rather than on scheduler internals:

- `Warm unread chats on connect and on every indicator refresh` — warming still happens when the subscription attaches and on each `threadListChanged`.
- `Keep indicators readable when chat warming fails` — a failing catch-up leaves indicators readable, and the next refresh recovers the cache. **Verified to fail on the pre-change implementation.**

## Verification

- `message-port.test.ts` 20/20, `platform-direct-bridge.test.ts` 6/6.
- 21 test files touching the catch-up / shared-database / indicator surface: 329 tests pass.
- `sidebar.test.tsx` 42/42.
- `tsc` production + tests, `oxlint`, `eslint`, `prettier` clean.

## Also: remove the welcome thread debug-panel tests

Three files, 703 lines: `welcome-thread.test.tsx`, `welcome-thread-sync.test.tsx`, `welcome-thread-sync-overlap.test.tsx`.

**The subject is a staff debug action.** The welcome thread is created by clicking a button in Settings > Debug, gated behind both `OkouDebug` and `WelcomeThread`. `createWelcomeThread$` is one POST; the API writes the chat event and Ably delivers it, and the Worker synchronizes the list from that notification. The frontend's `syncEventDrivenChatThreads$` call is explicitly a fallback — *"Recover the ordinary list even when the creation notification was lost."*

**Two of the files test an ownership model that no longer exists.** #33306 made the welcome synchronization outlive Settings dismissal by reparenting it from the dialog signal to a `rootSignal$` child, and added these tests. #33327 reverted that; `createWelcomeThread$` now documents the opposite rule — *"The caller passes the Settings action signal, so dismissal cancels the request and the list catch-up together"* — matching `settingsActionSignal$`: `// Actions stop at dismissal; the dialog session survives its closing animation.` The tests kept passing only because their assertions hold under either model, while their names still claim the removed behavior.

**Staging that stale scenario left the external behavior boundary.** `installWorker` stubs the `SharedWorker` global, spies on `postMessage` at both ends of the MessagePort, parses `sharedDatabaseClientMessageSchema`, and withholds individual responses by `requestId`; cases are parameterized over the `cache-only` and `catch-up` consistency levels, which have no user-visible meaning. Platform's external interface is the page, and `SharedWorker` / `SharedDatabaseMessagePortServer` / the protocol schema are internal code (AP-4, AP-6, AP-7). `docs/testing/testing-external-behavior.md` names this shape directly: *"Do not hide exceptions inside generic fixtures where future tests inherit internal coupling by default."*

This also settles the flakiness these two sync files showed under parallel load with the change above, whose cause was never pinned down: the files are gone.

Coverage that goes away is the debug card's switch gating, its 409/403/404 copy, and its cancel-on-dismissal behavior, plus two overlap cases staged through the welcome flow. The cold "Chat thread not found" pane those overlap cases touched keeps coverage in `chat-list-cache.test.tsx`, `chat-metadata.test.tsx`, and `chat-offline-cache.test.tsx`, which construct it through the page.

`knip` is clean after the deletion — no helper or export is orphaned.


e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: f03f6f89aff93846d8209f951719539701400858
PR_AUTO_REVIEW_VERDICT: LGTM

Reviewed HEAD: `f03f6f89aff93846d8209f951719539701400858`
Practice revision: `cc1178b92c62a403b6052768df40d7a04ea40933`

### Summary

Stops `readWorkerChatThreadIndicators$` from awaiting the throttled ChatEvent catch-up, so a tab's unread indicators no longer inherit the 1000ms warming throttle or the warming's failures, and moves the warming into `refreshWorkerChatIndicators$`, which the `threadListChanged` subscription primes on connect. Replaces four tests that pinned the old ordering, and deletes three welcome-thread debug-panel test files whose subject is a staff-only Settings action and whose premise was reverted by #33327.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | `refreshWorkerChatIndicators$` is reached only from `reloadWorkerChatIndicatorsFromRealtime$` and `reloadWorkerChatIndicatorsFromReadCursor$`; `refreshWorkerComputed$` has no production caller outside `test-bridge.ts`. See observation 1 for the degraded-realtime narrowing. |
| `REVIEW.md` | Regression fixes need meaningful coverage | Pass | `Keep indicators readable when chat warming fails` was verified to fail against the pre-change implementation (indicator read rejected with status 500) and pass after. |
| `REVIEW.md` | Tests verify externally observable state, not internal state or mock calls | Pass | New tests synchronize on the bridge's `databaseInvalidated` event and assert through `bridge.query` / `bridge.getComputed`, the same interface `chatThreadIndicatorsFromWorker$` uses. See observation 2. |
| `REVIEW.md` | Refactors must keep existing coverage protecting the behavior | Pass | The 4 replaced cases are superseded by 2 that cover warming on connect/refresh and warming-failure isolation; `message-port.test.ts` 20/20, `platform-direct-bridge.test.ts` 6/6. |
| `REVIEW.md` | ccstate ownership: await/signal ownership preserved | Pass | Warming is awaited by the Ably loop body that owns it; no `detach()` added. `catchUpChatEvent$` keeps its existing `rootSignal$` ownership. |
| `REVIEW.md` | New fallbacks declared in the PR body | N/A | The diff introduces no fallback; it removes an await and a manual cache, neither of which handles a state the contract rules out. |
| `REVIEW.md` | Persisted data / contract compatibility | N/A | No schema, API contract, response validator, or `protocol.ts` message shape changes; only the timing of an existing Worker-internal call. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | No `any` and no suppression comments added; `oxlint`, `oxlint --type-aware`, and `eslint .` are clean. |
| `docs/bad-smell.md` | Remove error handling that hides an invariant; preserve real recovery | Pass | Removing the awaited catch-up stops a prefetch failure from being reported as an indicator-read failure. Warming failures remain owned and logged by the Ably loop's retry. |
| `docs/bad-smell.md` | Document changed public contracts and assess consumers | Pass | `readWorkerChatThreadIndicators$` is module-private; both call sites are updated in the diff. |
| `docs/bad-smell.md` | Static imports in production code | Pass | No import changes. |
| `docs/cache.md` | RC-3: do not add a manual cache duplicating `computed` memoization | Pass | `workerChatThreadIndicatorsCache$` existed only to dedupe the awaited catch-up per indicator source; with the await gone, `chatThreadIndicators$` memoization covers it and the manual cache is deleted. |
| `docs/cache.md` | RC-4: state belongs to the narrowest lifecycle that owns it | Pass | Warming moves to the daemon-owned realtime refresh; the throttle factory keeps its `rootVersion$`-keyed identity, which `lifecycle.md` cites as the canonical example. |
| `docs/effect.md` | Keep cause and effect together | Pass | "Indicators refreshed" and "warm the changed threads" are now one command, instead of the refresh bumping a state and a later reader performing the warming. |
| `docs/effect.md` | Make cancellation explicit with a clear owner | Pass | `refreshWorkerChatIndicators$` checks `signal.throwIfAborted()` after the await; `catchUpChatEvent$` retains its `rootSignal$` owner unchanged. |
| `.claude/skills/ccstate/references/lifecycle.md` | Throttle factories: return or await their promises, detach only at the DOM boundary | Pass | Warming is `await set(catchUpChatEvent$)` inside a command; no `detach()` enters the signals layer. |
| `.claude/skills/ccstate/references/lifecycle.md` | Do not call `clearAllDetached()` or add file-level cleanup in tests | Pass | New tests add none; cleanup stays with `testContext`. |
| `docs/testing.md` | Choose the boundary the caller relies on | Pass | `message-port.test.ts` tests the MessagePort bridge at its own entry point. See observation 2. |
| `docs/testing.md` | Mock external services at their boundary; no global MSW server in page tests | Pass | New tests use `context.mocks.api` and `context.mocks.ably`; no `server.use()` and no internal module mock. |
| `docs/testing.md` | Avoid fake timers; use `mockNow(value, context.signal)` | Pass | `mockNow(startedAt + 2000, context.signal)` advances the app clock past the throttle; synchronization is the observable `databaseInvalidated`, not elapsed time. |
| `docs/testing.md` | Avoid duplicate cases that pin incidental implementation | Pass | The three 999/1000ms boundary variants duplicated `signals/__tests__/command-scheduling.test.ts`, which owns that arithmetic directly; they are removed. |
| `docs/testing/app-testing.md` | Page-level integration by default; signal tests only without a page-visible surface | Pass | Warming has no page-visible surface. See observation 2 for the failure-mode case. |
| `docs/testing/app-testing.md` | Never assert internal signal, store, request-count, or cache-protocol details in page tests | Pass | No page test asserts these; the deletions remove the file that inspected MessagePort `requestId` routing. |
| `docs/testing/testing-external-behavior.md` | Platform's external interface is the page | Pass | The deleted `installWorker` stubbed `SharedWorker`, spied both `postMessage` ends, and parsed the internal protocol schema from a page test. |
| `docs/testing/testing-external-behavior.md` | Do not hide exceptions inside generic fixtures | Pass | That fixture is removed rather than repaired, so no test inherits transport coupling by default. |
| `docs/fallback.md` | No negative tests pinning retired behavior | Pass | The deletions remove tests; none is replaced by an assertion that old behavior stays absent. |
| `docs/fallback.md` | Removing a reader requires evidence its producer is gone | Pass | The removed ordering was reverted by #33327, whose current doc comment states the opposite rule; the removed manual cache had no remaining purpose. |

### Findings

No P0 or P1 findings.

Non-blocking observations:

1. **Warming no longer runs when realtime setup fails.** `runSharedDatabaseWorkerDaemons$` returns early when `setupRealtime$` fails, so the `threadListChanged` subscription never attaches and `refreshWorkerChatIndicators$` is never primed. Previously the first tab indicator read triggered warming regardless of realtime health. The consequence is bounded: indicators still load, and both `chat-thread-event-sourcing.ts` and `chat-event-storage-signals.ts` fall back from `cache-only` to their own `catch-up`, so only the head start is lost in an already-degraded state. Worth recording against the realtime-recovery work in #32471 rather than widening this PR.

2. **The failure-mode contract also has a page-visible surface.** `Keep indicators readable when chat warming fails` asserts through the bridge, which is correct for this file and matches the tests it replaces. The same regression is observable through the sidebar's unread badge surviving a `catch-up` 500, and a page test would bind it closer to the user contract. Not blocking: `message-port.test.ts` is the established boundary for this surface, warming itself has no page-visible surface, and the current test is verified to catch the regression.

3. `readWorkerChatThreadIndicators$` is now a one-line pass-through. Keeping it preserves both call sites and names the Worker-side read, so inlining is a preference rather than a defect.

### Testing

- `message-port.test.ts` 20/20; the two replacements were run against the pre-change implementation to confirm `Keep indicators readable when chat warming fails` fails there, so it is a real regression guard rather than a restatement.
- `platform-direct-bridge.test.ts` 6/6; `sidebar.test.tsx` 42/42.
- 21 test files covering the catch-up, shared-database, and indicator surfaces: 329 tests pass.
- `chat-list-cache` / `chat-metadata` / `chat-offline-cache` 15/15, confirming the cold "Chat thread not found" pane the deleted overlap cases touched retains page-level coverage and that the shared test helpers are still used.
- `knip` clean after the deletions; `tsc -p tsconfig.production.json`, `tsc -p tsconfig.tests.json`, `oxlint`, `oxlint --type-aware`, `eslint .`, and `prettier --check` all clean.

Coverage removed with the three welcome-thread files is the debug card's feature-switch gating, its 409/403/404 copy, and its cancel-on-dismissal behavior. That is the staff-only Settings > Debug action gated behind `OkouDebug` and `WelcomeThread`, not a product flow; creation itself is one POST whose chat event the API writes and Ably delivers.

Issue #33351, opened 2026-09-11
bug(pi-memory): separate stored summaries from prompt injection budgets


seven332:
## Final disposition — 2026-09-13

**Accepted and completed.** [Final controller evidence and historical disposition](https://github.com/vm0-ai/vm0/issues/33351#issuecomment-5654247422). Natural larger summaries are stored and projected in full; the shipped reader keeps the exact 2500-token excerpt bound. No active historical recovery target was found. Three superseded historical projection rows remain untouched.

## Approved implementation contract — 2026-09-11

Lancy approved this Codex-inspired design in the [controller chat](https://app.okou.ai/chats/922b01fa-5dbb-43e9-97ee-fca3fb8e44cd) and invoked `/handoff-issue-to-thread`. **This contract supersedes the previous proposal in [comment 5629024654](https://github.com/vm0-ai/vm0/issues/33351#issuecomment-5629024654), including write-time token rejection, a rejection counter, a pending-breach state machine, and forced corrective generations.** The original incident evidence is preserved below as history.

A valid memory summary may exceed 2500 tokens on disk. The 2500-token budget applies to the summary excerpt injected into the model, including its truncation marker. It does not apply to the full stored artifact, nor to the surrounding fixed prompt instructions. Preserve the existing 64 KiB source-byte ceiling and all UTF-8, `v1` header, immutable-path, artifact, hash and frozen-version safeguards.

Example: a valid 2943-token, approximately 15 KiB summary must publish in full with its corresponding `MEMORY.md` and skills. API-first and sandbox turns inject the same deterministic excerpt of at most 2500 o200k tokens; the omitted portion remains available through `memories_read`. Token-only excess must no longer discard an otherwise valid maintenance result.

### Why this design

The current system conflates storage validation with prompt injection. Phase 2 validates the final artifact after the maintenance agent finishes; projection, wire schema and recall readers independently reject the same source above 2500 tokens. Increasing only one cap fails at another boundary.

Official Codex at [`da20788df913189878ebca7f4963d8a363ee6bf2`](https://github.com/openai/codex/tree/da20788df913189878ebca7f4963d8a363ee6bf2/codex-rs/memories) separates stored memory from bounded read-path injection. Its V1 read path uses an approximate 2500-token budget; V2 separately has a 10,000-byte source contract. Borrow the separation, not V2's unrelated byte policy or the approximate tokenizer. vm0 already has a shared exact-o200k middle-truncation helper and a 64 KiB source boundary.

[...]

PR #33558, opened 2026-09-11
refactor(pi): canonicalize the internal execution route

lancy:
Closes #33556. Implements child B of #33519.

Pi already captures its execution route, but API-first and runtime consumers separately interpreted its wire generation and accepted an optional bag of dialect-specific credentials. Normalize that captured value into one non-secret, in-process route, then materialize a discriminated model configuration at the existing API or sandbox execution edge. Historical Gen1 `api` values, including `openai-completions` and `openai-codex-responses`, continue to send public Responses requests.

## Implementation and retained boundaries

- `execution-route.ts` is the single Gen1/V2/V3/V4 normalization path, derived from existing contracts. It owns nested credential/policy snapshots before asynchronous resolution. `materializePiAgentModelConfig` remains the CLI wire adapter; API-first captures the same normalized route for credential resolution and late usage attribution.
- Model, stream, native-stream, and A's session-registration helper preserve dialect correlations. Codex requires its account and SSE; public priority and Codex fast stay separate; native Messages requires catalog/header policy, and Bedrock requires catalog/region/event-stream plus a complete explicit auth bundle. Phase 2 retains A's local engine and snapshots the complete model variant.
- The two native opaque-environment projections now share one helper over captured intent. API direct secrets, custom-header dummy keys/Authorization suppression, sandbox opaque markers, exact-account source validation, and access-token refresh before account-ID lookup retain their existing owners and order.
- Resource, skill, agents-file, and recall declarations derive from `PiResourceSnapshot`, retaining the necessary public package names, readonly semantics, and V1/V2 readers.

Caller inventory covers API-first, CLI Pi launch, ordinary session/RPC, Stage 1, Phase 2, model/native-stream, and session registration. The following original-wire adapters remain deliberate:

- `pi-sandbox-config.ts`: initial provider/environment resolution and existing writer generation. Its three admission callers and completion telemetry policy are unchanged.
- `agent-run-create.service.ts`: launch persistence, generation-specific pinned CLI checks, native credential override/ambient-auth validation, and exact egress authority. Captured `piModelConfig` remains the only persisted plan.
- `pi-api-first-turn.service.ts`: original-wire drain telemetry for #31085; native captured metadata is adapted to the unchanged billing input type, never serialized as a new launch payload.
- `pi-model-config-claim-capability.ts`, contract schemas/fixtures, and Runner `env.rs`/`env/native.rs`: independent claim, validation, serialization, and egress trust boundaries.

[...]

PR #33567, opened 2026-09-11
fix(pi-memory): use the official terra context window for phase 2

### LINKED ISSUE #33560: bug(pi-memory): correct the premature Terra maintenance context limit
## Approved implementation contract — Phase 2 only

The controller accepts the second research pass **for one focused correction to memory maintenance**, under the existing delegated memory-repair scope. Implement in the original owner thread using Claude Opus 5, obtain an independent review of the complete final SHA, and use normal `/pr-auto` protected-queue merge. Production release and data recovery remain outside the implementation owner's authority.

This contract supersedes the first-pass research-only restriction for this narrowly defined PR. It does **not** approve a global model-metadata override, input-selection reduction, compaction activation, a new cost-policy limit, or historical replays. Parent #33351 remains OPEN.

### Problem and independent evidence

The controller fetched the [official GPT-5.6 Terra model specification](https://developers.openai.com/api/docs/models/gpt-5.6-terra) on 2026-09-11 UTC. It states **1,050,000 context tokens** and **128,000 maximum output tokens**. **272K is a pricing threshold**: when input exceeds it, the stated input/output multipliers apply to the **full request**, not only the excess portion. This does not prove how the SDK's incorrect value was originally introduced.

The verified release `e7b5d4f0cdf78291d3323bec94824197be83dfa5` contains patched pi-ai 0.85.1 with an `openai / gpt-5.6-terra / openai-responses` catalog context window of 272000. The controller independently exercised the actual released `buildBaseOptions` and Responses adapter with intercepted fake fetch and synthetic context. At estimated context 267888, the serialized output ceiling is already 16; at 267903 it clamps to 1 and serializes 16. Synthetic `response.incomplete / max_output_tokens` maps to `length`, which Phase 2 correctly rejects before publication.

In a second offline counterfactual using the same actual adapter and changing only the model object's context window to the official 1050000, estimated context 270000 and 330000 both serialize `max_output_tokens=128000`. The normal clamp begins above 917904 estimated context; at the true window's edge it can still produce a small budget. This is verified request construction, **not** a claim that a scripted response proves natural maintenance success or that every historical failure had one cause.

The original bounded production cohort and uncertainty are retained below. The 8/10 billing signature remains supporting observation. An output quantity of 16 does not identify the serialized cap or route; even a serialized cap of 16 does not require `available<=1` because the adapter floors all positive caps below 16. The other two failures remain unattributed. The reported 34-hour standstill was a time error and is withdrawn.

Controller baseline: `origin/main=a0bc2016737c91f4889f547219d2b9798902e40e`. Relevant model/session, Phase 2 engine/tools/filesystem, selection-service and package files are unchanged from the verified release. Refresh main again before implementation.

### Minimal design and exact scope

1. Resolve the maintenance model through the existing resolver. **Inside Phase 2 maintenance session construction only**, adapt the exact legacy catalog case: resolved provider `openai`, API/dialect `openai-responses`, canonical catalog identity `gpt-5.6-terra`, and existing context window **272000**. For that case set the context window to **1050000**, citing the official specification next to the narrow compatibility correction. Preserve a catalog value that is already corrected or otherwise different; do not cap newer metadata at a stale local value.
2. Preserve identity and route: model ID/alias, provider, base URL, headers, credentials, tier, transport, reasoning policy, `maxTokens=128000` and pricing fields stay as resolved. The correction must not match `openai-codex`, OpenRouter, other models, or other dialects. It must not mutate a shared catalog object.
3. Use the **same corrected model** for provider registration and actual maintenance `AgentSession` creation. Changing only `registeredModelConfig` is insufficient if `createAgentSessionFromServices` still receives the old model object. Verify the model that reaches the real adapter.
4. Prefer a small private helper or immutable model copy within the existing Phase 2 code. Do not introduce a global override registry, dependency patch/bump, new API/DB/wire field, generic budget framework, or change to ordinary `resolvePiAgentModel`, API-first turns, user conversations, Stage 1, or their compaction thresholds.
5. Preserve selection and evidence semantics, tools and read chunk sizes, system prompt, retry and compaction settings, existing 20 MiB selection ceiling, 64 KiB summary source ceiling, 2500 exact-o200k injected excerpt, and final non-`stop` rejection. Do not delete or defer evidence to fit a smaller request.

### Why this scope

A Phase 2-local correction addresses the demonstrable premature SDK clamp without moving ordinary conversation compaction from roughly 256K to over 1M tokens. It restores the official capacity for this maintenance path; it does not prefill the window or reserve/charge 128000 output tokens automatically. It can allow more input and generated tokens than the broken path, so longer successful attempts can cost more. No cost-reduction factor or guaranteed completion is claimed.

A broader catalog correction and a general maintenance context-management policy are separate decisions. This PR does not enable compaction, add an early-failure label, or claim that arbitrarily large evidence sets will always finish within the real window. Legitimate incomplete results still fail, preserve the prior published bundle, and remain observable. Subsequent production acceptance must test actual behavior rather than merely the absence of one failure string.

### Invariants and non-goals

- Preserve complete, atomic publication of `MEMORY.md`, `memory_summary.md`, skills and selected evidence; existing encoding/header/byte/path/hash/archive/immutable guards remain authoritative.
- Preserve caller cancellation priority, lease/heartbeat fences, timeout behavior, ownership and marker/checkpoint publication. No successful marker for final `length`, error or abort; no mutation of the mounted base on failure.
- Do not change the retained evidence window. The engine rebuilds the Pi evidence subtree from `selected`, so smaller selection would be a deletion policy rather than an innocuous batch size. No batching, retry-counter reset, requeue or frozen-epoch rewrite.
- No change to global log levels, diagnostic vocabularies, Sentry behavior or equivalent new info/debug diagnostics. Preserve existing truthful failure reporting.
- No model/provider switch, routing/key/account/configuration change, production test run or provider request, release action, or upstream issue/PR in this task. No global provider-catalog repair by stealth.

### Required targeted validation

Use the existing deterministic Phase 2 engine/SSE fixture path and real adapter boundary; do not assert only a helper return value or weaken a fixture to make it pass.

- **Exact request construction:** scripted previous usage plus trailing tool messages crossing the former 272K boundary; the actual maintenance request serializes the corrected positive ceiling (for example 270000/330000 estimated context ->128000). Restoring the legacy window must make this regression fail. Include the new legitimate clamp boundary (917904 ->128000, 917905 ->127999) without requiring a real large provider request.
- **Isolation:** ordinary model resolution/user-run behavior and other providers/models/dialects retain their prior metadata. The correction never mutates the catalog. Already-corrected metadata is not overwritten.
- **Session progression:** synthetic consecutive paginated reads with cumulative context crossing the old cliff can proceed to a normal `stop`; include large existing memory plus new evidence and coherent final writes. Scripted provider success proves deterministic control flow, not model quality or production recovery.
- **Publication and preservation:** successful preparation retains the unchanged selected evidence/legacy immutable files and produces the coherent full bundle; failure/cancel/final `length` still yield no prepared result or marker/checkpoint and leave the mounted base byte-identical. Existing relevant regression tests may satisfy unchanged boundaries; do not duplicate implementation-shaped tests or add a new retention-policy contract.
- **Source/excerpt behavior:** the accepted full-source/2500-token excerpt policy remains intact. No new summary-token rejection.
- Run focused local tests and relevant lint/type/format checks. Do not run full local Vitest or a dev server. Report unavailable local tests honestly and use PR CI. All required source and merge-group gates must pass on the reviewed SHA through normal protected merge.

### Owner, review, merge and production acceptance

Original owner: [memory research/implementation thread](https://app.okou.ai/chats/605a0a0c-1610-41f2-abe8-38eb37c97d72), **Claude Opus 5**. One focused PR for #33560, with `Closes #33560`; same-thread `/pr-auto` immediately after PR creation. Independent means a distinct reviewer context, not the author's self-review marker; record provenance and bind the full current SHA. Any conflict repair or HEAD change requires fresh independent review and CI. Never direct/admin/bypass merge or rerun a disposable merge-group.

The controller checked all files of 23 open PRs, including pagination for #33450 (117 files). #33558 overlaps Pi model/session/Phase 2 tests; #33559 has a runtime version change. These are inventory, not an ordering blocker. Recheck live overlap and resolve genuine conflicts in the owned branch after main changes; do not wait for or modify another owner's PR.

The existing dedicated observer is reused and updated to this implementation lifecycle before dispatch. The owner stops at GitHub `MERGED` or a proven blocker. The controller independently accepts the merged code, retires the dedicated completion observer, and verifies any separately authorized deployment. The current already-published release is not authorization for this owner to perform a new release.

Production acceptance requires a finite, naturally occurring cohort tied to the **actual corrected CLI/runtime artifact**, nonzero maintenance execution and coherent successful publication evidence, with remaining failures and cost limits stated honestly. Zero attempts, a different failure label, or no long-context billing rows are not recovery. Historical slice C remains separately unstarted pending a fixed, attributable content-free inventory; #33351 stays OPEN until that disposition and the remaining behavior are accepted.

<details>
<summary>Original research brief and production evidence (research-only scope superseded by the contract above)</summary>

## Background

Production verification of #33351 found a separate, still-active Phase 2 failure: the maintenance session's last assistant message has `stopReason="length"`, recorded as `stage=final_response / reason=final_stop_length`. This is a real failed maintenance run, not the former `output_validation / summary_tokens` rejection and not a reliable optional-output omission.

The summary-storage change is present in release [#33552](https://github.com/vm0-ai/vm0/pull/33552), commit `e7b5d4f0cdf78291d3323bec94824197be83dfa5`: a valid summary up to 65536 UTF-8 bytes is stored in full; the shared prompt excerpt remains bounded to 2500 exact o200k tokens. Do not reverse that accepted design.

## Verified production evidence

The controller made read-only Axiom and MaskDB queries on 2026-09-11 UTC, with fixed observation window `[2026-09-11T15:00:00Z, 2026-09-11T16:00:00Z)`.

- Exact `agent_run_callbacks.internal_kind = 'pi-memory:phase2'` membership confirms 16 distinct maintenance run IDs. A separately scoped `agent_runs` query found 15 failed and one completed. All 16 callbacks were delivered; callback delivery is not maintenance success.
- `vm0-run-context-prod`, projecting only framework and the `CLI_PKG_URL` environment entry, directly ties **11** of those runs to the new immutable `e7b5d4f0.../package.tgz` CLI. **10 failed with `final_stop_length`; one completed** (created 15:47:15Z, completed 15:51:25Z). No memory contents were queried. The completed run does not prove its summary exceeded 2500 tokens.
- New-package failure completions: 15:22:08.777Z, 15:24:16.425Z, 15:25:55.988Z, 15:27:09.851Z, 15:29:48.613Z, 15:35:02.616Z, 15:46:31.429Z, 15:53:09.566Z, 15:59:05.520Z, 15:59:54.546Z. All appear under the directly verified API deployment `dpl_CVGQcutDgT9Nh3NQMSkA1VE5yUqk`.
- Five runs used the preceding `0e6c8095...` CLI: four also failed with `final_stop_length`; one failed at 15:13:49.700Z with `summary_tokens`, actual 3261 / limit 2500. One old-package run completed after API promotion and is deliberately excluded from the new-package cohort. These observations establish that `final_stop_length` existed before this release; they do not establish a release-induced regression or a comparative population failure rate.
- Error detail is content-free: `Pi memory Phase 2 maintenance session failed. pi_memory_phase2={"stage":"final_response","reason":"final_stop_length"}`. Private run identities remain in the controller/owner handoff, not this issue.

Release execution: [API promotion job](https://github.com/vm0-ai/vm0/actions/runs/34613739281/job/103314698450), successful production migrations and deployment, completed 15:18:30Z. Actual API source/CLI identity was verified separately from the workflow success label.

## Source evidence and unknowns

- At the released commit, [`phase2-memory.ts:81`](https://github.com/vm0-ai/vm0/blob/e7b5d4f0cdf78291d3323bec94824197be83dfa5/turbo/packages/pi-agent-runtime/src/phase2-memory.ts#L81) maps `length` to `final_stop_length`; [`providerResult:359`](https://github.com/vm0-ai/vm0/blob/e7b5d4f0cdf78291d3323bec94824197be83dfa5/turbo/packages/pi-agent-runtime/src/phase2-memory.ts#L359) rejects a final non-`stop` result before final output validation. Merely deleting this check could publish incomplete maintenance output.
- [`createMaintenanceSession:418`](https://github.com/vm0-ai/vm0/blob/e7b5d4f0cdf78291d3323bec94824197be83dfa5/turbo/packages/pi-agent-runtime/src/phase2-memory.ts#L418) disables automatic retry and compaction, and configures the model, reasoning policy and restricted maintenance tools. Trace the actual model/SDK/request policy rather than assuming which budget was hit.
- Actual wire-level termination, effective output/reasoning budgets, per-attempt usage, and the reason the model session stops here are not yet established. `length` is the runtime's classification, not proof of one specific upstream limit.
- `memory_summary_projections` and `pi_memory_phase2_jobs` are absent from the current MaskDB table inventory. Their row counts, retry state and terminal populations are unknown, not zero. Do not turn this failure into a guessed historical recovery population.
- Current in-flight [#33558](https://github.com/vm0-ai/vm0/pull/33558) changes model/session and Phase 2 files. Inspect its actual diff as inventory; it is not an ordering blocker or proof this defect is covered. The preceding local-engine refactor [#33547](https://github.com/vm0-ai/vm0/pull/33547) preserved the summary-storage policy.

## First pass and acceptance

Reuse the original #33351 research owner, Claude Opus 5. First perform a bounded, read-only investigation and return a minimal evidence-backed repair proposal with observable acceptance criteria. Do not implement or create a PR before the controller has checked that scope.

The proposal must separate verified facts from hypotheses, identify the exact point that produces the `length` stop, demonstrate why the proposed recovery is safe for the complete memory bundle, and explain bounded cost/cancellation/lease behavior. A deterministic synthetic local reproduction is useful; no full local Vitest or dev server. Preserve real failure reporting as ERROR and reliable success/fallback semantics without duplicate diagnostics.

Explicitly exclude production replays or synthetic runs, bulk requeues, credentials/provider/routing/budget changes without a reviewed contract, warning-only suppression, treating an incomplete model result as success, changing the accepted 64 KiB source / 2500-token excerpt policy, release execution, and modification of another owner's branch. Keep #33351 open; its historical recovery slice is separate.

Related project: #33351. Controller: [existing chat](https://app.okou.ai/chats/922b01fa-5dbb-43e9-97ee-fca3fb8e44cd). Original research owner: [existing memory thread](https://app.okou.ai/chats/605a0a0c-1610-41f2-abe8-38eb37c97d72).

</details>

lancy:
## Problem

Phase 2 memory maintenance fails with `stage=final_response / reason=final_stop_length`: the maintenance session's last assistant message returns `stopReason="length"`, and `providerResult` correctly refuses to publish an incomplete turn, so consolidated memory is never updated.

The cause is a stale catalog context window, not model output volume.

The pinned pi-ai catalog publishes `openai` / `gpt-5.6-terra` / `openai-responses` with `contextWindow: 272000`. Per the [official GPT-5.6 Terra specification](https://developers.openai.com/api/docs/models/gpt-5.6-terra) that model has **1,050,000 context tokens** and **128,000 maximum output tokens**; **272K is a pricing threshold**, above which the stated multipliers apply to the **full request**, not only the excess. The repository's own `MODEL_LONG_CONTEXT_MIN_TOTAL_INPUT_TOKENS` records the same 272,001 boundary as a billing tier.

`buildBaseOptions` derives each request's output ceiling as `min(model.maxTokens, max(1, contextWindow - estimatedContext - 4096))`, and the Responses adapter floors the serialized parameter at 16. With the stale window, a consolidation turn whose estimated context passes roughly 268,000 serializes `max_output_tokens=16`, the provider returns `incomplete / max_output_tokens`, and pi-ai maps that to `length`.

The repository's local `pi-ai` patch touches no catalog, `contextWindow` or clamp code, so the stale value comes from the pinned upstream package and is present in the installed patched build (verified locally: `contextWindow=272000`, `maxTokens=128000`, clamp present, `CONTEXT_SAFETY_TOKENS=4096`).

## Change

One narrow correction inside Phase 2 maintenance session construction only. It matches exactly the legacy case — resolved provider `openai`, API `openai-responses`, catalog identity `gpt-5.6-terra`, existing context window `272000` — and returns an **immutable copy** with the official `1050000`.

The corrected object is assigned to the single `model` variable that reaches **both** `registeredModelConfig(...)` and `createAgentSessionFromServices({ model })`, so the real adapter cannot receive the stale window.

Preserved exactly as resolved: model ID/alias, provider, base URL, headers, credentials, tier, transport, reasoning policy (`effort: max`), `maxTokens=128000` and pricing fields. The shared catalog object is never mutated, ordinary `resolvePiAgentModel` is untouched, and a value that is already corrected or otherwise different is returned unchanged, so newer metadata is never capped at a stale local value.

## Not in this change

No global model-metadata override or override registry, no dependency patch/bump, no new API/DB/wire field, no budget framework. No change to `openai-codex`, OpenRouter, other models or dialects, API-first turns, user conversations, Stage 1, or their compaction thresholds. Selection set and 20 MiB ceiling, tool and read chunk sizes, system prompt, compaction/retry settings, lease/heartbeat, cancellation precedence, the final non-`stop` rejection, the 64 KiB summary source ceiling and the 2500 exact-o200k injected excerpt are all unchanged. No evidence is deleted or deferred, and no log level, diagnostic vocabulary or Sentry behavior changes.

## Verification

Focused local runs against the real adapter boundary (the fixture asserts the actually serialized request body, not a helper return value):

- `phase2-memory.test.ts` — 22 passed, including three new cases.
- Related suites `phase2-memory-filesystem` / `-tools` / `-diagnostics` / `-selection` / `-scope`, `model`, `memory-recall` — 8 files, 115 passed.
- Types (`tsc --noEmit`), ESLint and Prettier clean on the changed files.

New coverage:

- **Request construction past the legacy threshold** — scripted prior assistant usage of 270,000 and 330,000 tokens plus the trailing tool message; the next real maintenance request serializes `max_output_tokens=128000` with `reasoning.effort=max` preserved.
- **The real clamp still works near the official window** — scripted contexts of 950,000 and 950,001 serialize positive ceilings below 128,000 that differ by exactly 1, so the genuine subtraction boundary is intact rather than removed.
- **Scope** — ordinary `resolvePiAgentModel` still reports `contextWindow=272000` and `maxTokens=128000` before and after a maintenance session, proving no catalog mutation; a different catalog model on the same provider and dialect (`gpt-5.6-sol`) still derives the legacy floor of 16, pinning the correction to the single matched case.

Mutation check: restoring the legacy `272000` constant makes the two budget tests fail with `expected 16 to be 128000` and `expected 16 to be greater than 16`, confirming they bind the actual defect. The fix was restored and re-verified green.

Existing regression tests continue to cover the unchanged boundaries — coherent full-bundle preparation with retained selected evidence and legacy immutable files, final `length` / error / abort producing no prepared result, marker or checkpoint and leaving the mounted base byte-identical, cancellation precedence, and the source/excerpt policy.

Per the repository guidance I did not run full local Vitest or a dev server; the complete suite runs in PR CI.

## Limits

This restores the official capacity for the maintenance path. It does not prefill the window, does not reserve or charge 128,000 output tokens, and does not guarantee that any particular historical or arbitrarily large evidence set completes. Because more input and generated tokens are now possible, a longer successful attempt can cost more, and requests above 272K input are billed at the model's stated long-context multipliers for the full request. Legitimate incomplete results still fail, preserve the previously published bundle and remain observable. The original count of 16 output tokens does not by itself prove a serialized cap or route, and the two unattributed failures in the observed cohort are not claimed to share this cause. Production acceptance requires a finite naturally occurring cohort on the actual corrected artifact with nonzero maintenance execution and coherent successful publication; a missing failure string alone is not recovery.

A broader catalog correction and a general maintenance context-management policy remain separate decisions. Parent #33351 stays open for its remaining historical slice.

Closes #33560

### DIFF
--- a/turbo/packages/pi-agent-runtime/src/phase2-memory.test.ts
+++ b/turbo/packages/pi-agent-runtime/src/phase2-memory.test.ts
@@ -12,6 +12,7 @@ import { join } from "node:path";
 import { encode } from "gpt-tokenizer/encoding/o200k_base";
 import { afterEach, describe, expect, it } from "vitest";
 
+import { resolvePiAgentModel } from "./model";
 import type { PiMemoryPhase2Diagnostic } from "./phase2-memory-diagnostics";
 import { renderPiMemoryPhase2Prompt } from "./phase2-memory-prompt";
 import { PI_MEMORY_PHASE2_TOOL_NAMES } from "./phase2-memory-tools";
@@ -140,12 +141,19 @@ type ProviderStep =
       readonly type: "tool";
       readonly name: string;
       readonly arguments: Record<string, unknown>;
+      /**
+       * Report a large prior context for the turn. The SDK anchors its context
+       * estimate on the last assistant usage, so this drives the next request's
+       * derived output ceiling without a real large provider request.
+       */
+      readonly usageTotalTokens?: number;
     }
   | {
       readonly type: "text";
       readonly text: string;
       /** Reproduce a provider that returns no response id for the turn. */
       readonly omitResponseId?: true;
+      readonly usageTotalTokens?: number;
     }
   | { readonly type: "incomplete"; readonly reason: string }
   | { readonly type: "http-error" }
@@ -162,13 +170,13 @@ function writeSse(response: ServerResponse, events: readonly unknown[]): void {
   );
 }
 
-function usage() {
+function usage(totalTokens?: number) {
   return {
     input_tokens: 11,
     output_tokens: 7,
     input_tokens_details: { cached_tokens: 2 },
     output_tokens_details: { reasoning_tokens: 2 },
-    total_tokens: 18,
+    total_tokens: totalTokens ?? 18,
   };
 }
 
@@ -228,7 +236,7 @@ function toolSse(
         object: "response",
         status: "completed",
         output: [item],
-        usage: usage(),
+        usage: usage(step.usageTotalTokens),
       },
     },
   ]);
@@ -282,7 +290,7 @@ function textSse(
         object: "response",
         status: "completed",
         output: [item],
-        usage: usage(),
+        usage: usage(step.usageTotalTokens),
       },
     },
   ]);
@@ -1248,5 +1256,69 @@ describe("Pi memory Phase 2 consolidation engine", () => {
     await expect(stat(cleanupRoot ?? "missing")).rejects.toMatchObject({
       code: "ENOENT",
     });
+  });
+
+  async function maintenanceRequestBudget(
+    priorContextTokens: number,
+    modelOverrides: Partial<PiMemoryPhase2LocalConsolidationArgs["model"]> = {},
+  ): Promise<number> {
+    const provider = await startProvider([
+      {
+        type: "tool",
+        name: "phase2_write",
+        arguments: {
+          path: "memory/MEMORY.md",
+          content: "# Task Group: maintenance budget\n",
+        },
+        usageTotalTokens: priorContextTokens,
+      },
+      { type: "text", text: "consolidated" },
+    ]);
+    const input = args(provider.baseUrl);
+    const result = await runPiMemoryPhase2LocalConsolidation(
+      { ...input, model: { ...input.model, ...modelOverrides } },
+      new AbortController().signal,
+    );
+    expect(result.status).toBe("prepared");
+    const next = provider.requests[1];
+    if (!next) {
+      throw new Error("Missing the Phase 2 request after the reported context");
+    }
+    expect(next.body).toMatchObject({ reasoning: { effort: "max" } });
+    const budget = next.body.max_output_tokens;
+    if (typeof budget !== "number") {
+      throw new Error("Phase 2 request did not serialize an output ceiling");
+    }
+    return budget;
+  }
+
+  it("serializes the official output ceiling past the legacy context threshold", async () => {
+    // The legacy catalog window collapses both of these to the Responses
+    // adapter's 16-token floor, which ends the turn as `length`.
+    expect(await maintenanceRequestBudget(270_000)).toBe(128_000);
+    expect(await maintenanceRequestBudget(330_000)).toBe(128_000);
+  });
+
+  it("keeps the real context clamp active near the official window", async () => {
+    const lower = await maintenanceRequestBudget(950_000);
+    const higher = await maintenanceRequestBudget(950_001);
+    expect(lower).toBeGreaterThan(16);
+    expect(lower).toBeLessThan(128_000);
+    expect(lower - higher).toBe(1);
+  });
+
+  it("scopes the correction to the one legacy catalog case", async () => {
+    const config = args("http://127.0.0.1:1/v1").model;
+    // Ordinary resolution keeps the catalog value before and after maintenance.
+    expect(resolvePiAgentModel(config)?.contextWindow).toBe(272_000);
+    // A different catalog model on the same provider and dialect is untouched,
+    // so it still derives the legacy floor. This pins the correction's scope;
+    // it is not a statement that the other model should stay uncorrected.
+    expect(
+      await maintenanceRequestBudget(270_000, { catalogModel: "gpt-5.6-sol" }),
+    ).toBe(16);
+    const after = resolvePiAgentModel(config);
+    expect(after?.contextWindow).toBe(272_000);
+    expect(after?.maxTokens).toBe(128_000);
   });
 });
--- a/turbo/packages/pi-agent-runtime/src/phase2-memory.test.ts
+++ b/turbo/packages/pi-agent-runtime/src/phase2-memory.test.ts
@@ -1311,12 +1311,21 @@ describe("Pi memory Phase 2 consolidation engine", () => {
     const config = args("http://127.0.0.1:1/v1").model;
     // Ordinary resolution keeps the catalog value before and after maintenance.
     expect(resolvePiAgentModel(config)?.contextWindow).toBe(272_000);
-    // A different catalog model on the same provider and dialect is untouched,
-    // so it still derives the legacy floor. This pins the correction's scope;
-    // it is not a statement that the other model should stay uncorrected.
+    // Precondition: the sibling model still carries the same stale catalog
+    // window. If the catalog is corrected upstream this fails deliberately, so
+    // the scope of the local correction is re-decided rather than drifting.
     expect(
-      await maintenanceRequestBudget(270_000, { catalogModel: "gpt-5.6-sol" }),
-    ).toBe(16);
+      resolvePiAgentModel({ ...config, catalogModel: "gpt-5.6-sol" })
+        ?.contextWindow,
+    ).toBe(272_000);
+    // The correction cannot lift a different catalog model on the same provider
+    // and dialect, so its derived ceiling stays below the corrected one.
+    const corrected = await maintenanceRequestBudget(270_000);
+    const untouched = await maintenanceRequestBudget(270_000, {
+      catalogModel: "gpt-5.6-sol",
+    });
+    expect(corrected).toBe(128_000);
+    expect(untouched).toBeLessThan(corrected);
     const after = resolvePiAgentModel(config);
     expect(after?.contextWindow).toBe(272_000);
     expect(after?.maxTokens).toBe(128_000);
--- a/turbo/packages/pi-agent-runtime/src/phase2-memory.test.ts
+++ b/turbo/packages/pi-agent-runtime/src/phase2-memory.test.ts
@@ -104,6 +104,8 @@ function selected(
 function args(
   baseUrl: string,
   overrides: Partial<PiMemoryPhase2LocalConsolidationArgs> = {},
+  /** Written into the literal below so the dialect arm stays discriminated. */
+  catalogModel = "gpt-5.6-terra",
 ): PiMemoryPhase2LocalConsolidationArgs {
   return {
     memoryStorageId: "storage-phase2",
@@ -121,7 +123,7 @@ function args(
       baseUrl,
       apiKey: "PROVIDER_KEY_SECRET_31243",
       model: "MODEL_ALIAS_SECRET_31243",
-      catalogModel: "gpt-5.6-terra",
+      catalogModel,
       dialect: "openai-responses",
       transport: "sse",
       thinkingLevel: "max",
@@ -1262,7 +1264,7 @@ describe("Pi memory Phase 2 consolidation engine", () => {
 
   async function maintenanceRequestBudget(
     priorContextTokens: number,
-    modelOverrides: Partial<PiMemoryPhase2LocalConsolidationArgs["model"]> = {},
+    catalogModel?: string,
   ): Promise<number> {
     const provider = await startProvider([
       {
@@ -1276,9 +1278,8 @@ describe("Pi memory Phase 2 consolidation engine", () => {
       },
       { type: "text", text: "consolidated" },
     ]);
-    const input = args(provider.baseUrl);
     const result = await runPiMemoryPhase2LocalConsolidation(
-      { ...input, model: { ...input.model, ...modelOverrides } },
+      args(provider.baseUrl, {}, catalogModel),
       new AbortController().signal,
     );
     expect(result.status).toBe("prepared");
@@ -1317,15 +1318,14 @@ describe("Pi memory Phase 2 consolidation engine", () => {
     // window. If the catalog is corrected upstream this fails deliberately, so
     // the scope of the local correction is re-decided rather than drifting.
     expect(
-      resolvePiAgentModel({ ...config, catalogModel: "gpt-5.6-sol" })
-        ?.contextWindow,
+      resolvePiAgentModel(
+        args("http://127.0.0.1:1/v1", {}, "gpt-5.6-sol").model,
+      )?.contextWindow,
     ).toBe(272_000);
     // The correction cannot lift a different catalog model on the same provider
     // and dialect, so its derived ceiling stays below the corrected one.
     const corrected = await maintenanceRequestBudget(270_000);
-    const untouched = await maintenanceRequestBudget(270_000, {
-      catalogModel: "gpt-5.6-sol",
-    });
+    const untouched = await maintenanceRequestBudget(270_000, "gpt-5.6-sol");
     expect(corrected).toBe(128_000);
     expect(untouched).toBeLessThan(corrected);
     const after = resolvePiAgentModel(config);
--- a/turbo/packages/pi-agent-runtime/src/phase2-memory.ts
+++ b/turbo/packages/pi-agent-runtime/src/phase2-memory.ts
@@ -413,6 +413,57 @@ function providerResult(
   return { responseId: final.responseId ?? null, usage: Object.freeze(usage) };
 }
 
+type ResolvedPiAgentModel = NonNullable<ReturnType<typeof resolvePiAgentModel>>;
+
+/**
+ * The legacy catalog case this narrow maintenance correction adapts.
+ *
+ * The pinned catalog publishes `openai` / `gpt-5.6-terra` / `openai-responses`
+ * with a 272000 context window. 272000 is that model's long-context pricing
+ * threshold, above which the stated multipliers apply to the full request; the
+ * official specification states 1050000 context tokens and 128000 maximum
+ * output tokens: https://developers.openai.com/api/docs/models/gpt-5.6-terra
+ */
+const PI_MEMORY_PHASE2_LEGACY_CONTEXT_PROVIDER = "openai";
+const PI_MEMORY_PHASE2_LEGACY_CONTEXT_API = "openai-responses";
+const PI_MEMORY_PHASE2_LEGACY_CONTEXT_CATALOG_MODEL = "gpt-5.6-terra";
+const PI_MEMORY_PHASE2_LEGACY_CONTEXT_WINDOW = 272_000;
+const PI_MEMORY_PHASE2_OFFICIAL_CONTEXT_WINDOW = 1_050_000;
+
+/**
+ * Return the maintenance model with the stale catalog context window corrected.
+ *
+ * `buildBaseOptions` derives each request's output ceiling by subtracting the
+ * estimated context from `contextWindow`, so the pricing threshold above caps a
+ * long consolidation turn at the Responses adapter's minimum output budget. The
+ * turn then ends as `length`, which Phase 2 correctly refuses to publish.
+ *
+ * This adapts only that one legacy case, only for maintenance, and only as an
+ * immutable copy: the shared catalog object is never mutated, ordinary
+ * resolution is untouched, and a value that is already corrected or otherwise
+ * different is returned unchanged so newer metadata is never capped. Model
+ * identity, provider, route, credentials, headers, tier, transport, reasoning
+ * policy, `maxTokens` and pricing all stay exactly as resolved.
+ */
+function maintenanceModelWithOfficialContextWindow(
+  model: ResolvedPiAgentModel,
+  config: SnapshotPhase2Input["model"],
+): ResolvedPiAgentModel {
+  if (
+    model.provider !== PI_MEMORY_PHASE2_LEGACY_CONTEXT_PROVIDER ||
+    model.api !== PI_MEMORY_PHASE2_LEGACY_CONTEXT_API ||
+    (config.catalogModel ?? config.model) !==
+      PI_MEMORY_PHASE2_LEGACY_CONTEXT_CATALOG_MODEL ||
+    model.contextWindow !== PI_MEMORY_PHASE2_LEGACY_CONTEXT_WINDOW
+  ) {
+    return model;
+  }
+  return {
+    ...model,
+    contextWindow: PI_MEMORY_PHASE2_OFFICIAL_CONTEXT_WINDOW,
+  };
+}
+
 async function createMaintenanceSession(args: {
   readonly input: SnapshotPhase2Input;
   readonly workspace: Phase2PrivateWorkspace;
@@ -420,10 +471,16 @@ async function createMaintenanceSession(args: {
   readonly testHooks: PiMemoryPhase2EngineTestHooks | undefined;
 }): Promise<AgentSession> {
   initializePiSessionResourceRegistry();
-  const model = resolvePiAgentModel(args.input.model);
-  if (!model) {
+  const resolved = resolvePiAgentModel(args.input.model);
+  if (!resolved) {
     throw new Phase2InputInvalidError();
   }
+  // One corrected model object reaches both provider registration and the real
+  // session, so the adapter cannot receive the stale window.
+  const model = maintenanceModelWithOfficialContextWindow(
+    resolved,
+    args.input.model,
+  );
   const modelRuntime = await ModelRuntime.create({
     allowModelNetwork: false,
     credentials: new InMemoryCredentialStore(),
--- a/turbo/packages/pi-agent-runtime/src/phase2-memory.ts
+++ b/turbo/packages/pi-agent-runtime/src/phase2-memory.ts
@@ -444,6 +444,11 @@ const PI_MEMORY_PHASE2_OFFICIAL_CONTEXT_WINDOW = 1_050_000;
  * different is returned unchanged so newer metadata is never capped. Model
  * identity, provider, route, credentials, headers, tier, transport, reasoning
  * policy, `maxTokens` and pricing all stay exactly as resolved.
+ *
+ * Remove this correction once the pinned catalog publishes the official window:
+ * the guard then stops matching, so it is inert rather than wrong. The catalog
+ * identity is matched here instead of the caller's product model id because the
+ * stale value belongs to that catalog entry, not to Okou's model selection.
  */
 function maintenanceModelWithOfficialContextWindow(
   model: ResolvedPiAgentModel,


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: 0dcae807687303a47ac75ecd4485bf98641c9411
PR_AUTO_REVIEW_VERDICT: LGTM

Independent review of `fix/pi-memory-phase2-terra-context-window` at inspected HEAD `0dcae807687303a47ac75ecd4485bf98641c9411`. Practice documents read from `main`: `REVIEW.md`, `docs/docs.md`, `docs/bad-smell.md`, `docs/fallback.md`, `docs/testing.md`, `docs/testing/anti-patterns.md`.

### Summary

Phase 2 memory maintenance builds its session model from `resolvePiAgentModel`, whose pinned `openai` catalog publishes `contextWindow: 272000` for `gpt-5.6-terra`. Because `clampMaxTokensToContext` derives the per-request output ceiling as `min(maxTokens, max(1, contextWindow - estimatedContext - 4096))` and the Responses adapter then floors `max_output_tokens` at 16, a consolidation turn past roughly 268K estimated context serialized `max_output_tokens=16`, returned `stopReason="length"`, and was correctly refused for publication. The PR adds `maintenanceModelWithOfficialContextWindow` in `turbo/packages/pi-agent-runtime/src/phase2-memory.ts`, which returns an immutable copy carrying the official `1050000` window for exactly one matched case, and feeds that single object to both `registeredModelConfig(...)` and `createAgentSessionFromServices({ model })`.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | Production entry `runPiMemoryPhase2MountedConsolidation` (`phase2-memory.ts:789`) delegates to `runPiMemoryPhase2LocalConsolidation`, which reaches `createMaintenanceSession` (`phase2-memory.ts:467`). The corrected object is the only `model` binding afterwards, so both `modelRuntime.registerProvider(...)` (via `registeredModelConfig`, which copies `contextWindow` at `session-model.ts:48`) and `createAgentSessionFromServices({ model })` receive it. No stale reference survives. |
| `REVIEW.md` | Check ordering, cancellation, serialization, and failure outcomes | Pass | The diff only changes one numeric metadata field on a per-session model copy. Publication atomicity, the validation marker, lease/heartbeat, `signal.throwIfAborted()` precedence, the final non-`stop` rejection via `INCOMPLETE_STOP_REASONS`, and evidence retention are untouched. Phase 2 sessions keep `compaction: { enabled: false }` and `retry: { enabled: false }`, so the larger window cannot shift a compaction threshold. |
| `REVIEW.md` | Require meaningful coverage for a regression fix; identify the missing scenario | Pass | Three new cases in `phase2-memory.test.ts:1261-1323` drive the production entry point and assert the actually serialized provider request body. Author-reported mutation check (restoring `272000` → `expected 16 to be 128000`) is consistent with what I traced through `clampMaxTokensToContext`. |
| `REVIEW.md` | Verify externally observable state, not internal state or mock calls | Pass | `maintenanceRequestBudget` asserts `result.status === "prepared"` plus `provider.requests[1].body.max_output_tokens` / `reasoning.effort` captured by a real `node:http` server. The one internal assertion (`resolvePiAgentModel(...)?.contextWindow`) exists only to prove the shared catalog is not mutated, which has no external surface. |
| `REVIEW.md` | Distinguish introduced defects from unchanged code | Pass | The stale `272000` originates in the pinned `@earendil-works/pi-ai@0.85.1` `openai.json` catalog, not in this diff. The remaining exposure on other paths is recorded below as a non-blocking observation, not as a defect this PR introduces. |
| `REVIEW.md` | Persisted data / API contract / compatibility review | N/A | No schema, wire field, JSONB shape, queue payload, or response-validated contract changes; `PiAgentModelConfig` and the launch payload are unchanged. |
| `docs/bad-smell.md` §1 | No `any`; preserve inference, narrow at untrusted boundaries | Pass | `ResolvedPiAgentModel = NonNullable<ReturnType<typeof resolvePiAgentModel>>` reuses the resolver's own inference. The `model.api !== "openai-responses"` test narrows the four-member union so the spread stays assignable; `pnpm exec tsc --noEmit` is clean on the package. |
| `docs/bad-smell.md` §2 | No lint/type suppressions and no weakened rules | Pass | No `@ts-expect-error`, `eslint-disable`, or config change in the diff; `eslint src/phase2-memory.ts src/phase2-memory.test.ts` passes locally. |
| `docs/bad-smell.md` §3 | Catch/handle errors only at a meaningful boundary; no fabricated defaults hiding an invariant | Pass | No error handling added. The `Phase2InputInvalidError` throw for an unresolvable model is preserved ahead of the correction. |
| `docs/bad-smell.md` §4 | Document changed public contracts and assess consumers | Pass | `maintenanceModelWithOfficialContextWindow` is module-private; no exported signature changes. `registeredModelConfig` and `createAgentSessionFromServices` consumers were both inspected. |
| `docs/bad-smell.md` §6 | Centralized configuration; no invented fallback URLs or credentials | Pass | Base URL, headers, API key, tier, transport, and credential binding are passed through untouched; only `contextWindow` differs from the resolved object. |
| `docs/bad-smell.md` §7 / `docs/fallback.md` §3,§5,§6 | No fallback for a state the owning contract already rules out | Pass | This is not a fallback branch for an unreachable state: `openai.json` in the installed package really does publish `contextWindow: 272000` for `gpt-5.6-terra`, so the matched state is the current, reachable one. See the Fallbacks section for the full analysis. |
| `docs/fallback.md` §9 | Every fallback introduced by the diff is declared in a `Fallbacks` PR-body section | Pass | The diff introduces no fallback under §6, so no `Fallbacks` section is required; §9 explicitly exempts PRs that add none. |
| `docs/fallback.md` §1 | No negative tests that only pin retired behavior | Pass | The new cases assert live behavior at the provider boundary. The `gpt-5.6-sol` scope assertion pins current, not retired, behavior; it is raised below as a non-blocking brittleness observation rather than a tombstone. |
| `docs/testing.md` | Prefer integration coverage through real entry points; mock external services at their boundary | Pass | Tests enter through the exported consolidation function and mock only the provider, as a real HTTP server. No `vi.mock()` of internal modules, no fetch stubbing, no fake timers; server and temp-dir cleanup stay on the existing `afterEach`. |
| `docs/testing.md` | Avoid duplicate cases that restate static configuration or pin incidental implementation | Pass | The budget cases are not derivable from existing coverage: they are the only ones that exercise the clamp at a large estimated context. The scope case is the only guard against catalog mutation and over-broad matching. |
| `docs/testing/anti-patterns.md` AP-2/AP-3/AP-4/AP-5 | No fetch stubbing, fs mocking, internal mocking, or fake timers | Pass | Real `node:http` provider, real `mkdtemp` workspaces, real internal code, no timer manipulation. |
| `docs/testing/anti-patterns.md` AP-7/AP-8 | Do not assert implementation details or duplicate coverage | Pass | The primary assertions are on the serialized wire request. The `resolvePiAgentModel` assertion is a non-mutation proof with no production-observable equivalent. |
| `docs/testing.md` | Existing coverage still protects the unchanged behavior | Pass | The `usage()` helper defaults to the previous `total_tokens: 18`, and every pre-existing `ProviderStep` omits `usageTotalTokens`, so the existing `result.usage` assertions (`phase2-memory.test.ts:603`, `:675`) are byte-identical. Phase 2 usage is summed from `input/output/cacheRead/cacheWrite/reasoning`, never from `total_tokens`, so the harness change cannot mask an assertion. Local run: 22 passed. |

### Findings

No P0 or P1 findings.

Non-blocking observations (P2, not required for this verdict):

1. **The same stale catalog value still drives two adjacent paths.** `turbo/packages/pi-agent-runtime/src/stage1-memory.ts:354-362` (`resolvePiMemoryStage1ContextWindow`) and `turbo/packages/pi-agent-runtime/src/compaction-preflight.ts:113-121` both read the uncorrected `resolvePiAgentModel(...).contextWindow`. Stage 1 memory runs on the same model (`PI_MEMORY_STAGE1_MODEL = "gpt-5.6-terra"`, `pi-memory-stage1-worker.service.ts:58`), so its history-projection budget is `0.7 × 272_000 = 190_400` instead of roughly `735_000`, and user runs on `gpt-5.6-terra`/`-sol`/`-luna` compact earlier than necessary and can hit the same `max_output_tokens=16` collapse. This is pre-existing and explicitly out of scope per the PR body, but it is the same root cause; a follow-up issue naming `resolvePiAgentModel` or `turbo/patches/@earendil-works__pi-ai@0.85.1.patch` as the single correction point would close it for every path with one guard.

2. **The scope assertion hardcodes an unrelated adapter constant and pins a value that is itself wrong.** `phase2-memory.test.ts:1315-1317` asserts `gpt-5.6-sol` still derives `16`. `16` is `OPENAI_RESPONSES_MIN_OUTPUT_TOKENS` inside the pinned `openai-responses` adapter, not a Phase 2 contract, and `gpt-5.6-sol` is documented with the same 1,050,000 window — the same installed package's `openrouter.json` already publishes `1050000` for `openai/gpt-5.6-sol`. The assertion therefore reads as "Sol should stay capped" and will fail the moment the catalog is bumped or the correction is broadened. The inline comment mitigates the reading, but anchoring the scope on the resolved model (for example, that Sol's ceiling is still derived from the catalog's `272_000`) would be less brittle and would not encode a defect as an expectation.

3. **No removal gate recorded for the dependency-pinned branch.** The `contextWindow !== 272_000` condition makes the helper self-disabling once the catalog is corrected upstream, which is the right failure mode, but the doc comment names no removal condition or follow-up. Adding one line — remove when the pinned `pi-ai` catalog publishes the official window for `gpt-5.6-terra` — keeps the branch from outliving its cause. Relatedly, `PI_MEMORY_PHASE2_LEGACY_CONTEXT_CATALOG_MODEL` duplicates `PI_MEMORY_PHASE2_MODEL` (`turbo/apps/api/src/signals/services/pi-memory-phase2-usage.service.ts:12`), which the package boundary prevents importing; a cross-reference comment would flag the pairing for a future model change.

4. **Fixture realism nit.** `usage(totalTokens)` can now emit `total_tokens: 270_000` alongside `input_tokens: 11` / `output_tokens: 7`. The SDK's `calculateContextTokens` prefers `totalTokens`, so this is harmless and the intent is documented, but the fixture is internally inconsistent and could confuse a later reader debugging usage accounting.

Verification I performed independently of the PR body:

- Confirmed the installed patched `@earendil-works/pi-ai@0.85.1` publishes `contextWindow: 272000, maxTokens: 128000` for `gpt-5.6-terra`, `-sol`, and `-luna` in `dist/providers/data/openai.json`, and that `clampMaxTokensToContext` (`dist/api/simple-options.js`) and the 16-token floor (`dist/api/openai-responses.js:236`) behave as the PR describes.
- Confirmed the official specification independently: 1,050,000 context window, 922,000 max input tokens, 128,000 max output tokens. Corroborating internal evidence: the same pinned package's `openrouter.json` already lists `openai/gpt-5.6-terra` at `contextWindow: 1050000, maxTokens: 128000`, so `272000` in the `openai` catalog is a pricing threshold recorded as a window, exactly as the PR argues.
- Confirmed production reachability of the guard. The Phase 2 built-in route resolves through `resolveGptApiKeyPiModelConfig` (`pi-sandbox-config.ts:348`): the `openai-api-key` candidate yields `provider: "openai"`, `modelPrefix: ""`, no `catalogModel`, so `config.catalogModel ?? config.model === "gpt-5.6-terra"` and the guard matches the real production config, not only the fixture. The `openrouter-codex` fallback candidate yields `provider: "openrouter"` and resolves against the OpenRouter catalog, which is already `1050000` — so excluding it is correct, not a gap.
- Confirmed the guard cannot over-match or cap newer metadata: it requires `provider === "openai"`, `api === "openai-responses"` (excluding `openai-codex-responses`, `anthropic-messages`, and `bedrock-converse-stream`), the catalog key `gpt-5.6-terra` resolved with the exact same `config.catalogModel ?? config.model` expression `resolvePiAgentModel` uses for lookup, and `contextWindow === 272_000` exactly. Any corrected, larger, or different value falls through unchanged.
- Confirmed no shared-object mutation is possible: `resolvePiAgentModel` already builds a fresh `base` object per call (`model.ts:433-454`), and the helper adds a second spread; nothing in the diff writes to a catalog-owned object.
- Confirmed the spread preserves every other field, including `maxTokens: 128000`, `cost`, `headers`, `compat`, `reasoning`, `thinkingLevelMap`, `id`, `name`, `provider`, and `baseUrl`; the wire assertion on `reasoning.effort === "max"` pins the reasoning policy as well.

### Fallbacks

The diff introduces no fallback under `docs/fallback.md` §6, so no `Fallbacks` PR-body section is required.

- `phase2-memory.ts:448 maintenanceModelWithOfficialContextWindow` — this is a correction of incorrect third-party metadata for the state the pinned dependency currently produces, not an extra path for a state the system is not supposed to produce (§3) and not a reader for a retired producer (§5). It does not fabricate data for a corrupt persisted row (§4), guard a pre-GA switch cutover (§2), or bridge two independently deployed component versions (§7). **Verdict: not a fallback; no declaration required.** The one §8-style hygiene item it does miss — a stated removal condition — is recorded as non-blocking observation 3 above.

### Testing

Coverage is adequate and the new cases genuinely bind the defect rather than restating it.

- `phase2-memory.test.ts:1261-1293` (`maintenanceRequestBudget`) enters through `runPiMemoryPhase2LocalConsolidation`, requires `status === "prepared"`, and reads `max_output_tokens` off the second request body captured by the real provider server. That is the adapter boundary, not a helper return value, so the assertion would have failed before this fix.
- `:1295-1301` covers 270,000 and 330,000 estimated context, straddling the stale 272,000 window, and asserts `128_000` — the value only reachable with the corrected window.
- `:1302-1309` proves the genuine subtraction survives: 950,000 and 950,001 yield positive ceilings strictly between 16 and 128,000 that differ by exactly 1. This is the case that distinguishes "window corrected" from "clamp removed", and it is the most valuable of the three.
- `:1310-1323` proves no catalog mutation before and after a maintenance session and pins the match to one model. See observation 2 for its brittleness.
- Harness change reviewed for masking: `usage(totalTokens?: number)` keeps `total_tokens: 18` by default, every pre-existing step omits `usageTotalTokens`, and Phase 2's reported usage never reads `total_tokens`, so no existing assertion is weakened or altered. No test was deleted, relaxed, or rewritten.
- Local verification at this HEAD: `pnpm --filter @okouai/pi-agent-runtime exec vitest run src/phase2-memory.test.ts` → 1 file, 22 passed; `pnpm exec tsc --noEmit` clean; `eslint` clean on both changed files. Per repository guidance I did not run the full local Vitest suite or a dev server; CI owns those.

Gaps worth naming but not blocking: there is no case covering a model whose catalog window is already correct (the `openrouter` route for the same logical model), so the "never caps newer metadata" property is argued from the guard rather than pinned by a test; and the 950,000-token cases exceed the model's documented 922,000 max input tokens, so they exercise the clamp arithmetic rather than a reachable request.

lancy:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 1ce93da52406d1089ea031c47c45b03f6d730f27

Re-review at inspected HEAD `1ce93da52406d1089ea031c47c45b03f6d730f27`, superseding my `LGTM` on `0dcae80`. Practice documents re-read from `main`: `REVIEW.md`, `docs/docs.md`, `docs/bad-smell.md`, `docs/fallback.md`, `docs/testing.md`, `docs/testing/anti-patterns.md`.

### Summary

The follow-up commit `1ce93da` does exactly what was asked for the two actioned findings, and it does not weaken anything: `phase2-memory.ts` changes by five JSDoc lines only, so the executable production code is byte-identical to `0dcae80`, and the reworked scope check is strictly more explicit about the property under test. The verdict nevertheless has to move to `Changes Requested` for a reason unrelated to the author's edits — `main` moved underneath this branch and the PR no longer type-checks against it. `lint-types` is red at this HEAD with two real errors in `phase2-memory.test.ts`.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Changed behavior traced through its real entry point and consumers | Pass | `git diff 0dcae80768 1ce93da524 -- …/phase2-memory.ts` contains only ` * ` lines inside the helper's JSDoc block. `git diff origin/main...1ce93da524 -- …/phase2-memory.ts` is identical in executable content to the previously reviewed revision, so my HEAD-`0dcae80` traces through `createMaintenanceSession` → `registeredModelConfig` and `createAgentSessionFromServices({ model })` still hold unchanged. |
| `REVIEW.md` | Follow-up commit did not weaken an assertion | Pass | Analysed below in Testing. The replacement still fails on the defect it guards, and it adds a precondition assertion that did not exist before. |
| `REVIEW.md` | Correctness and failure outcomes; races, cancellation, ordering, serialization | Pass | No production behavior delta versus the previously reviewed revision. Publication atomicity, validation marker, lease/heartbeat, cancellation precedence, the final non-`stop` rejection, and evidence retention remain untouched. |
| `REVIEW.md` | Required checks and the change as it would actually merge | **Finding (P1)** | `lint-types` fails at this exact HEAD (`@okouai/pi-agent-runtime#check-types`, run `34627171904`). Two `TS2322`/`TS2345` errors in `phase2-memory.test.ts`. The same check passed at `0dcae80` only because that run's merge ref predated `cc1178b92c` on `main`. See finding 1. |
| `REVIEW.md` | Require meaningful coverage for a regression fix | Pass | The three budget/scope cases still bind the defect at the serialized provider request. Local focused run at this HEAD: 22 passed. |
| `REVIEW.md` | Verify externally observable state, not internal state or mock calls | Pass | Primary assertions remain on `provider.requests[1].body.max_output_tokens` and `result.status`. The two `resolvePiAgentModel` assertions are non-mutation/precondition proofs with no external surface. |
| `REVIEW.md` | Distinguish introduced defects from unchanged code | Pass | The stale `272000` catalog value originates in pinned `@earendil-works/pi-ai@0.85.1`, not this diff. The `lint-types` break is an interaction between this branch's test code and `main`'s new discriminated union — this PR owns the rebase, so it is raised against this PR rather than deferred. |
| `REVIEW.md` | Persisted data / API contract / compatibility | N/A | No schema, wire field, JSONB shape, queue payload, or response-validated contract changes. |
| `docs/bad-smell.md` §1 | No `any`; preserve inference, narrow at untrusted boundaries | Pass | `ResolvedPiAgentModel = NonNullable<ReturnType<typeof resolvePiAgentModel>>` is unchanged; the union narrowing on `model.api` is unchanged. Note the P1 is a genuine type error, not a suppression or an `any`. |
| `docs/bad-smell.md` §2 | No lint/type suppressions; do not weaken rules to pass | Pass | No suppression comment, no rule change, and no `@ts-expect-error` was added to silence the failing check. Please keep it that way when fixing finding 1. |
| `docs/bad-smell.md` §3 | Error handling only at a meaningful boundary | Pass | Unchanged; the `Phase2InputInvalidError` throw still precedes the correction. |
| `docs/bad-smell.md` §4 | Document changed public contracts and assess consumers | Pass | Helper stays module-private; no exported signature changed. The new JSDoc paragraph documents the removal condition and the catalog-identity rationale. |
| `docs/bad-smell.md` §6 | Centralized configuration; no invented URLs or credentials | Pass | Only `contextWindow` differs from the resolved object; base URL, headers, key, tier, transport and credential binding pass through untouched. |
| `docs/bad-smell.md` §7 / `docs/fallback.md` §3,§5,§6 | No fallback for a state the contract already rules out | Pass | Unchanged analysis; see Fallbacks. |
| `docs/fallback.md` §8 | A bounded compatibility branch states its removal condition | Pass (improved) | `phase2-memory.ts:448-451` now records the removal condition and explains why the catalog identity is matched instead of Okou's product model id. This resolves my previous observation 3. |
| `docs/fallback.md` §9 | Every introduced fallback declared in a `Fallbacks` PR-body section | Pass | The diff introduces no fallback under §6, so no section is required. |
| `docs/fallback.md` §1 | No negative tests pinning retired behavior | Pass (improved) | The scope check no longer pins `16`; it asserts a live catalog precondition and compares derived ceilings. This resolves my previous observation 2. |
| `docs/testing.md` | Integration coverage through real entry points; mock externals at their boundary | Pass | Still enters through `runPiMemoryPhase2LocalConsolidation` against a real `node:http` provider. No internal mocking, no fetch stubbing, no fake timers; cleanup stays on the existing `afterEach`. |
| `docs/testing.md` | Existing coverage still protects unchanged behavior | Pass | `usage()` still defaults to `total_tokens: 18` and every pre-existing step omits `usageTotalTokens`, so `phase2-memory.test.ts:603` / `:675` are untouched. No test was deleted, relaxed, or rewritten to pass. |
| `docs/testing/anti-patterns.md` AP-2/3/4/5 | No fetch/fs/internal mocking or fake timers | Pass | Unchanged. |
| `docs/testing/anti-patterns.md` AP-7/AP-8 | No implementation-detail assertions or duplicate coverage | Pass | The added precondition is the cheapest available proof of the sibling model's catalog state; `corrected` doubles as the comparison baseline rather than adding a redundant case. |

### Findings

**P1 — the PR does not type-check against current `main`; `lint-types` is red at this HEAD.**

- Trigger: `main` advanced to `cc1178b92c` (`refactor(pi): canonicalize the internal execution route (#33558)`), which converted `PiAgentModelConfig` from a flat interface into a discriminated union over `dialect`, with mutually exclusive `?: never` members and a now-required `transport`. This branch's merge base is still `6f00c41612`.
- Consequence — two genuine errors reported by `@okouai/pi-agent-runtime#check-types` in job `103355059883`:
  - `TS2322` at `phase2-memory.test.ts:1279` — `{ ...input, model: { ...input.model, ...modelOverrides } }` where `modelOverrides: Partial<PiMemoryPhase2LocalConsolidationArgs["model"]>` (`:1263`). Spreading a `Partial<>` of a discriminated union over a union member loses the discriminant correlation and the `?: never` exclusivity, producing the 15-member union the compiler reports as not assignable.
  - `TS2345` at `phase2-memory.test.ts:1318` — `resolvePiAgentModel({ ...config, catalogModel: "gpt-5.6-sol" })`. The `openai-codex-responses` member declares `catalogModel?: never`, so adding `catalogModel: string` invalidates that arm of the spread result.
- Why it passed before: the `0dcae80` `lint-types` run started at `17:08:46Z`, and `cc1178b92c` landed at `17:07:01Z`; that run's merge ref predates it. The `1ce93da` run at `17:21:18Z` includes it. Site 1 comes from the first commit and would fail today as well — this is not something the follow-up commit alone caused, and only site 2 is new.
- Remediation: rebase onto `main` and adapt the two spread sites so the discriminant is preserved — for example narrow to the `openai-responses` member before spreading, or build the config literal explicitly instead of spreading a `Partial<>` of the union. The fixture itself needs no work: `main`'s `cc1178b92c` already added `transport: "sse"` to `args()` at `phase2-memory.test.ts:125`, and that change does not overlap the lines this PR touches, so it should merge textually without conflict. `main` also rewrote `snapshotModel` to `structuredClone`, which still preserves `catalogModel`, so the production guard and its `config.catalogModel ?? config.model` key are unaffected — I verified the production file itself compiles; both errors are test-side.
- Please fix by rebasing and adapting the fixtures, not by adding a suppression or an `any` (`docs/bad-smell.md` §1, §2), and not by relaxing an assertion.

Non-blocking, carried forward as a follow-up recommendation only (explicitly **not** a gap in this PR):

1. **Adjacent Stage 1 and compaction-preflight paths still read the stale catalog window.** `stage1-memory.ts:354-362` and `compaction-preflight.ts:113-121` consume the uncorrected `resolvePiAgentModel(...).contextWindow`, and Stage 1 runs on the same model (`PI_MEMORY_STAGE1_MODEL = "gpt-5.6-terra"`). **I agree that leaving these untouched is correct for this PR's scope, not a gap.** Widening the fix to `resolvePiAgentModel` would change ordinary user runs, API-first turns, Stage 1 history projection and user-facing compaction thresholds — a materially larger blast radius touching billing multipliers and compaction behavior, which `docs/testing.md` would require covering on those surfaces and which none of this PR's regression tests exercise. Keeping the correction on the maintenance path is the smaller, verifiable change; the broader catalog correction belongs in its own PR with its own coverage. Recommend opening a follow-up issue naming `resolvePiAgentModel` or `turbo/patches/@earendil-works__pi-ai@0.85.1.patch` as the single correction point.
2. **Fixture realism nit.** `usage(totalTokens)` can emit `total_tokens: 270_000` alongside `input_tokens: 11` / `output_tokens: 7`. **I agree this is correctly left alone.** The SDK's `calculateContextTokens` prefers `totalTokens`, the intent is documented at the field, and Phase 2's reported usage never reads `total_tokens`, so the fixture is functionally correct. Purely cosmetic.

Previously raised and now resolved: the brittle `toBe(16)` scope assertion (finding 2) and the missing removal gate (finding 3).

### Fallbacks

The diff introduces no fallback under `docs/fallback.md` §6, so no `Fallbacks` PR-body section is required.

- `phase2-memory.ts:453 maintenanceModelWithOfficialContextWindow` — a correction of incorrect third-party metadata for the state the pinned dependency currently produces. It is not a branch for a state the system is not supposed to produce (§3), a reader for a retired producer (§5), fabricated data for a corrupt row (§4), a pre-GA switch cutover (§2), or a cross-version deployment bridge (§7). **Verdict: not a fallback; no declaration required.** The §8-style hygiene item I previously flagged — a stated removal condition — is now recorded in the doc comment, and the scope test enforces it.

### Testing

I re-ran the focused suite myself at this HEAD: `pnpm --filter @okouai/pi-agent-runtime exec vitest run src/phase2-memory.test.ts` → 1 file, 22 passed. `eslint` and `prettier --check` are clean on both changed files. My local `tsc` passes only because this checkout's sibling sources and `node_modules` predate `cc1178b92c`; CI, which type-checks the merge with current `main`, is authoritative here and is the basis for the P1.

On whether `1ce93da` weakened the scope check — it did not:

- The old assertion was `expect(await maintenanceRequestBudget(270_000, { catalogModel: "gpt-5.6-sol" })).toBe(16)`. The new one asserts the sibling's catalog window is still `272_000` (`:1317-1320`), then `corrected === 128_000` and `untouched < corrected` (`:1323-1328`).
- The falsifying property is preserved. If the guard wrongly matched `gpt-5.6-sol`, that model's window would become `1_050_000`, its ceiling would be `min(128_000, 1_050_000 - 270_000 - 4_096) = 128_000`, and `expect(untouched).toBeLessThan(corrected)` would fail. Over-broad matching is still caught.
- The looser matcher is not a loophole in practice: with the precondition pinning the sibling window at `272_000` and an estimated context of at least `270_000`, `clampMaxTokensToContext` yields `max(1, 272_000 - 270_000 - 4_096) = 1`, which the Responses adapter floors to `16`. The value is still deterministically `16`; the test simply no longer hardcodes `OPENAI_RESPONSES_MIN_OUTPUT_TOKENS`, an unrelated pinned-adapter constant.
- It is a net gain in coverage. The precondition assertion is new, and combined with the pre-existing `expect(resolvePiAgentModel(config)?.contextWindow).toBe(272_000)` for Terra, an upstream catalog correction of *either* model now fails this test deliberately. That makes the scope test the enforcement mechanism for the removal condition the doc comment records — a good pairing.
- Minor: the precondition resolves against `baseUrl` `http://127.0.0.1:1/v1` while the measured run uses the live fixture server. `contextWindow` resolution is base-URL independent, so the precondition does describe the same catalog entry the second run uses.

Everything else in my previous testing assessment stands: the `950_000`/`950_001` pair (delta exactly 1) remains the case that distinguishes "window corrected" from "clamp removed", and the harness change cannot mask an existing assertion.

Verification carried over and re-confirmed at this SHA: the installed patched `pi-ai@0.85.1` publishes `contextWindow: 272000` for `gpt-5.6-terra`/`-sol`/`-luna` in `dist/providers/data/openai.json` while the same package's `openrouter.json` already lists `openai/gpt-5.6-terra` at `1050000`; the official specification states 1,050,000 context / 922,000 max input / 128,000 max output; the guard matches the real production config because the built-in `openai-api-key` candidate yields `provider: "openai"` with no `catalogModel`, so `config.catalogModel ?? config.model === "gpt-5.6-terra"`; and the `openrouter-codex` fallback candidate is correctly excluded because its catalog is already correct.

Once `lint-types` is green on a rebased HEAD with no suppression added, I expect this to return to `LGTM`.

lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: e40f9bbc848885fb2495052022a3e2c584e7f878
PR_AUTO_REVIEW_VERDICT: LGTM

Re-review at inspected HEAD `e40f9bbc848885fb2495052022a3e2c584e7f878`, superseding my `Changes Requested` on `1ce93da`. Practice documents re-read from `main`: `REVIEW.md`, `docs/docs.md`, `docs/bad-smell.md`, `docs/fallback.md`, `docs/testing.md`, `docs/testing/anti-patterns.md`.

### Summary

The branch now merges current `main` (`cc1178b92c`, #33558) and fixes both type errors structurally rather than by suppression: `args()` takes a defaulted `catalogModel` written directly into the model object literal, and the two former union-spread sites pass it through instead. My P1 is resolved — `lint-types` is green at this HEAD, and I reproduced that locally with the exact CI command. The production helper is byte-identical to the revision I already approved, and no assertion changed meaning.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Required checks pass for the change as it would merge | **Pass (P1 resolved)** | `git merge-base origin/main e40f9bbc84` is `cc1178b92c`, i.e. the branch's merge base is current `main`, so the merge ref and the branch content are the same thing. `lint-types` at this SHA is `completed/success`, along with `lint-format`, `lint-style`, `lint-knip`, `lint-type-api`, `lint-type-app`, `lint-runtime-api-compat`. Zero failing checks (30 success, 22 skipped, 24 pending). |
| `REVIEW.md` | Production behavior unchanged from the previously approved revision | Pass | `git diff 1ce93da524 e40f9bbc84 -- …/phase2-memory.ts` is **empty**. The merge did not alter the helper, its constants, or its call site. My earlier traces through `createMaintenanceSession` → `registeredModelConfig` and `createAgentSessionFromServices({ model })` carry over unchanged. |
| `REVIEW.md` | The merge brought in nothing unintended | Pass | `git diff --stat origin/main...e40f9bbc84` is exactly two files: `phase2-memory.ts` (+66/-6 equivalent) and `phase2-memory.test.ts`. `main`'s own `transport: "sse"` addition appears as an **unchanged context line** at `phase2-memory.test.ts:128`, so it is preserved and not re-authored by this branch. The lockfile is untouched (`git diff 1ce93da524 e40f9bbc84 -- …pnpm-lock.yaml` empty), so no dependency moved. |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | Unchanged: `runPiMemoryPhase2MountedConsolidation` → `runPiMemoryPhase2LocalConsolidation` → `createMaintenanceSession`, with the single corrected `model` binding reaching both provider registration and the session. |
| `REVIEW.md` | Correctness, races, cancellation, ordering, serialization, failure outcomes | Pass | No production delta. Publication atomicity, validation marker, lease/heartbeat, cancellation precedence, the final non-`stop` rejection, and evidence retention remain untouched. |
| `REVIEW.md` | Require meaningful coverage for a regression fix | Pass | The three budget/scope cases still assert the serialized provider request. Local run at this HEAD: `phase2-memory.test.ts` + `model.test.ts` → 2 files, 52 passed. |
| `REVIEW.md` | Verify externally observable state, not internal state or mock calls | Pass | Primary assertions remain `provider.requests[1].body.max_output_tokens` and `result.status`. The two `resolvePiAgentModel` assertions stay non-mutation/precondition proofs with no external surface. |
| `REVIEW.md` | Confirm existing coverage still protects the refactor | Pass | See the fixture audit in Testing: 24 `args(` occurrences in the file, exactly 2 of which pass a third argument, both new. Every pre-existing caller receives the default `"gpt-5.6-terra"` — the same literal that was previously hardcoded. |
| `REVIEW.md` | Persisted data / API contract / compatibility | N/A | No schema, wire field, JSONB shape, queue payload, or response-validated contract changes. `PiAgentModelConfig`'s union rewrite came from `main`, not this PR. |
| `docs/bad-smell.md` §1 | No `any`; preserve inference, narrow at untrusted boundaries | Pass | Grepped every added line in `git diff origin/main...e40f9bbc84` for `any`, `as unknown`: none. The fix works *with* the type system — `catalogModel` is placed inside the literal at `:126` so the `dialect: "openai-responses"` arm stays discriminated, and `args(...).model` is passed at its declared type instead of being spread. |
| `docs/bad-smell.md` §2 | No lint/type suppressions; do not weaken rules to pass | Pass | No `@ts-ignore`, `@ts-expect-error`, `@ts-nocheck`, `eslint-disable`, `oxlint-disable`, or `prettier-ignore` in the diff, and no config or rule change. This is precisely the remediation shape I asked for. |
| `docs/bad-smell.md` §3 | Error handling only at a meaningful boundary | Pass | Unchanged. |
| `docs/bad-smell.md` §4 | Document changed public contracts and assess consumers | Pass | Helper stays module-private. `args()`'s new parameter is test-local, defaulted, and documented inline at `:107`. |
| `docs/bad-smell.md` §6 | Centralized configuration; no invented URLs or credentials | Pass | Only `contextWindow` differs from the resolved object. |
| `docs/bad-smell.md` §7 / `docs/fallback.md` §3,§5,§6 | No fallback for a state the contract already rules out | Pass | Unchanged analysis; see Fallbacks. |
| `docs/fallback.md` §8 | A bounded compatibility branch states its removal condition | Pass | `phase2-memory.ts:448-451` records the removal condition and the catalog-identity rationale. |
| `docs/fallback.md` §9 | Every introduced fallback declared in a `Fallbacks` section | Pass | The diff introduces no fallback under §6, so no section is required. |
| `docs/fallback.md` §1 | No negative tests pinning retired behavior | Pass | The scope check asserts a live catalog precondition and compares derived ceilings. |
| `docs/testing.md` | Integration coverage through real entry points; mock externals at their boundary | Pass | Still enters through `runPiMemoryPhase2LocalConsolidation` against a real `node:http` provider. No internal mocking, no fetch stubbing, no fake timers; cleanup stays on the existing `afterEach`. |
| `docs/testing.md` | Existing coverage still protects unchanged behavior | Pass | `usage()` still defaults to `total_tokens: 18`; every pre-existing `ProviderStep` omits `usageTotalTokens`, so `phase2-memory.test.ts:606` / `:678` are untouched. No test was deleted, relaxed, or rewritten to pass. |
| `docs/testing/anti-patterns.md` AP-2/3/4/5 | No fetch/fs/internal mocking or fake timers | Pass | Unchanged. |
| `docs/testing/anti-patterns.md` AP-7/AP-8 | No implementation-detail assertions or duplicate coverage | Pass | Unchanged; `corrected` still doubles as the comparison baseline rather than adding a redundant case. |

### Findings

No P0 or P1 findings. My previous P1 (the PR not type-checking against current `main`) is resolved, and I verified the resolution rather than taking it on report.

Non-blocking, carried forward as follow-up recommendations only — both confirmed as **correct to leave, not gaps**:

1. **Adjacent Stage 1 and compaction-preflight paths still read the stale catalog window.** `stage1-memory.ts` (`resolvePiMemoryStage1ContextWindow`) and `compaction-preflight.ts` consume the uncorrected `resolvePiAgentModel(...).contextWindow`, and Stage 1 runs on the same model. Widening the fix there would change ordinary user runs, API-first turns, Stage 1 history projection and user-facing compaction thresholds — a materially larger blast radius touching billing multipliers, which `docs/testing.md` would require covering on those surfaces and which this PR's tests do not exercise. Keeping the correction on the maintenance path is the smaller, verifiable change. Recommend a follow-up issue naming `resolvePiAgentModel` or `turbo/patches/@earendil-works__pi-ai@0.85.1.patch` as the single correction point.
2. **Fixture realism nit.** `usage(totalTokens)` can emit `total_tokens: 270_000` alongside `input_tokens: 11` / `output_tokens: 7`. `calculateContextTokens` prefers `totalTokens`, the intent is documented at the field, and Phase 2's reported usage never reads `total_tokens`, so this is functionally correct and purely cosmetic.

Previously raised and resolved across the review history: the brittle `toBe(16)` scope assertion, the missing removal gate, and the `lint-types` break.

### Fallbacks

The diff introduces no fallback under `docs/fallback.md` §6, so no `Fallbacks` PR-body section is required.

- `phase2-memory.ts:453 maintenanceModelWithOfficialContextWindow` — a correction of incorrect third-party metadata for the state the pinned dependency currently produces. It is not a branch for a state the system is not supposed to produce (§3), a reader for a retired producer (§5), fabricated data for a corrupt row (§4), a pre-GA switch cutover (§2), or a cross-version deployment bridge (§7). **Verdict: not a fallback; no declaration required.** Its removal condition is recorded in the doc comment and enforced by the scope test.

### Testing

I re-verified all four requested points at this HEAD rather than relying on the reported results.

**Type safety and the former P1.** I ran the exact CI command, `pnpm run check-types` in `turbo/packages/pi-agent-runtime` → exit **0**. I also ran the pinned TypeScript 6 binary (`./node_modules/.bin/tsc6 --noEmit`) → exit **0**. Both now reflect merged `main`, unlike my earlier local run, whose pass was an artifact of a pre-merge checkout. `lint-types` on GitHub is `completed/success`. The fix is structural:

- `args()` at `:104-108` gains `catalogModel = "gpt-5.6-terra"` and writes it at `:126`, so the object literal still carries `dialect: "openai-responses"` + `transport: "sse"` and stays on one discriminated arm.
- `maintenanceRequestBudget(priorContextTokens, catalogModel?)` at `:1265-1267` drops `Partial<…["model"]>` entirely and calls `args(provider.baseUrl, {}, catalogModel)` at `:1282`. Passing `undefined` triggers the parameter default, so the no-argument path is byte-equivalent to the previous `args(provider.baseUrl)`.
- The sibling precondition at `:1322` uses `args("http://127.0.0.1:1/v1", {}, "gpt-5.6-sol").model`, passing a value at its declared type instead of spreading the union.
- I grepped the whole file for `...config`, `...input.model`, `Partial<…["model"]>`, and `modelOverrides`: **none remain**. No discriminant is lost anywhere else in the file.

**No assertion weakened or changed in meaning.** The file has 24 `args(` occurrences and exactly **two** pass a third argument (`:1282`, `:1322`), both new. Every other caller falls through to the default `"gpt-5.6-terra"`, which is the exact literal previously hardcoded, so no pre-existing test sees a different fixture. The three budget/scope tests keep their assertions verbatim: `toBe(128_000)` at 270,000 and 330,000; `lower > 16`, `lower < 128_000`, `lower - higher === 1` at 950,000/950,001; and the scope test's `toBe(272_000)` precondition, `corrected === 128_000`, `untouched < corrected` at `:1330`. The falsifying property is intact — if the guard wrongly matched `gpt-5.6-sol`, that model's ceiling would become `128_000` and `toBeLessThan(corrected)` would fail.

**Runtime and lint.** `pnpm --filter @okouai/pi-agent-runtime exec vitest run src/phase2-memory.test.ts src/model.test.ts` → 2 files, **52 passed**. `eslint` on both changed files → exit 0. `prettier --check` on both → clean. Per repository guidance I did not run the full local Vitest suite or a dev server; CI owns those and currently shows no failures at this SHA.

**Merge hygiene.** The merge commit `d559a70675` is a clean take of `main` — its only effect on this branch's own files was `main`'s `+2` lines in `phase2-memory.test.ts`, and the branch's contribution over `main` is still exactly the two intended files. No unrelated file, dependency, or generated artifact rode along.

Everything else from my earlier reviews stands and was re-confirmed at this SHA: the pinned `pi-ai@0.85.1` `openai.json` publishes `contextWindow: 272000` for `gpt-5.6-terra`/`-sol`/`-luna` while the same package's `openrouter.json` already lists `openai/gpt-5.6-terra` at `1050000`; the official specification states 1,050,000 context / 922,000 max input / 128,000 max output; the guard matches the real production config because the built-in `openai-api-key` candidate yields `provider: "openai"` with no `catalogModel`, so `config.catalogModel ?? config.model === "gpt-5.6-terra"`; and the `openrouter-codex` fallback candidate is correctly excluded because its catalog is already correct.

PR #33569, opened 2026-09-11
test(platform): remove the welcome thread debug-panel tests

e7h4n:
## What

Deletes three test files, 703 lines total:

- `welcome-thread.test.tsx`
- `welcome-thread-sync.test.tsx`
- `welcome-thread-sync-overlap.test.tsx`

## Why

**The subject is a staff debug action, not a product behavior.** The welcome thread is created by clicking a button in Settings > Debug, gated behind both `OkouDebug` and `WelcomeThread`. `createWelcomeThread$` is one POST; the API writes the chat event and Ably delivers it, and the Worker synchronizes the list from that notification. The frontend's `syncEventDrivenChatThreads$` call is explicitly a fallback — *"Recover the ordinary list even when the creation notification was lost."*

**Two of the files test an ownership model that no longer exists.** #33306 made the welcome synchronization outlive Settings dismissal by reparenting it from the dialog signal to a `rootSignal$` child, and added these tests. #33327 reverted that; `createWelcomeThread$` now documents the opposite rule:

> The caller passes the Settings action signal, so dismissal cancels the request and the list catch-up together.

matching `settingsActionSignal$` in `settings-dialog.ts`: `// Actions stop at dismissal; the dialog session survives its closing animation.`

The tests kept passing only because their assertions are loose enough to hold under either model — cancelling the sync also leaves cold not-found and committed history intact. Their names still claim the removed behavior.

**Staging that stale scenario required leaving the external behavior boundary.** `installWorker` stubs the `SharedWorker` global, spies on `postMessage` at both ends of the MessagePort, parses `sharedDatabaseClientMessageSchema`, and withholds individual responses by `requestId`; cases are then parameterized over the `cache-only` and `catch-up` consistency levels, which are internal query semantics with no user-visible meaning.

Per `docs/testing/testing-external-behavior.md`, Platform's external interface is the page, and `SharedWorker` / `SharedDatabaseMessagePortServer` / the protocol schema are all internal code (AP-4, AP-6, AP-7). The same doc calls out this exact shape: *"Do not hide exceptions inside generic fixtures where future tests inherit internal coupling by default."*

## Coverage impact

What goes away with the deletion:

- Feature-switch gating of the debug card, its 409/403/404 error copy, and its cancel-on-dismissal behavior — all of the staff debug action.
- Two overlap cases about stale list refreshes and abandoned readers, staged through the welcome flow.

The cold "Chat thread not found" pane those overlap cases touched keeps coverage in `chat-list-cache.test.tsx`, `chat-metadata.test.tsx`, and `chat-offline-cache.test.tsx`, which construct it through the page.

## Verification

- `knip` clean — the deletion orphans no helper or export.
- `tsc -p tsconfig.tests.json`, `oxlint`, `oxlint --type-aware`, `eslint .` all clean.
- `chat-list-cache` / `chat-metadata` / `chat-offline-cache` 15/15, confirming the shared test helpers are still used and valid.

## Relation to #33564

#33564 flagged these two sync files as flaky under parallel load with that change, cause unresolved. This PR removes them, so that risk disappears regardless of which PR lands first. The two are otherwise independent: #33564 is a product fix with its own regression guard.


e7h4n:
Folded into #33564 — the test deletion also resolves the flakiness risk that PR flagged, so they ship as one change.

PR #33574, opened 2026-09-11
refactor: make activity summary polling thread-owned

e7h4n:
## Summary
- make each thread's activity summary computed read the authoritative active run before subscribing to interval reloads
- reconcile at most one polling loop per thread, replacing or stopping it only when the active run ID changes
- compose the stable demand reconciler into the existing thread event-change command and preserve initial and feature-hydration reconciliation under the thread owner signal
- remove the three nested-command suppressions and the obsolete ready-run handoff state

## Verification
- `git diff --check`
- Source-level review against the existing `chat-activity-summary.test.tsx` coverage for event startup, feature hydration, interval refresh, run replacement/termination, hidden-page polling, and navigation cancellation
- Local Prettier, ESLint, TypeScript, and Vitest commands could not start because this checkout has no installed `node_modules`; the PR pipeline will run those checks

### DIFF
--- a/turbo/apps/platform/src/signals/chat-page/create-chat-thread.ts
+++ b/turbo/apps/platform/src/signals/chat-page/create-chat-thread.ts
@@ -2332,6 +2332,7 @@ function createEventChangeEffects(
     projections,
     scroll,
     syncVisibleEventTrees$,
+    reconcileThreadSummaryDemand$,
   }: {
     readonly threadId: string;
     readonly chatEvents: ChatEventSignals;
@@ -2344,6 +2345,7 @@ function createEventChangeEffects(
       Promise<void>,
       [boolean, AbortSignal]
     >;
+    readonly reconcileThreadSummaryDemand$: Command<void, [AbortSignal]>;
   },
   ownerSignal: AbortSignal,
 ) {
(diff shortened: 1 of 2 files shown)


e7h4n:
LGTM

PR_AUTO_REVIEWED_HEAD=caa2938638b1c03c5c02b2b92c160f0d8bef84e4
PR_AUTO_REVIEW_COMMIT=caa2938638b1c03c5c02b2b92c160f0d8bef84e4

**Summary**
- Makes the activity-summary computed thread-owned and reads the authoritative active run before subscribing to interval reload state.
- Reconciles one polling target per thread through the existing stable event-change command, with hydration and teardown owned by the thread signal.
- Removes the three nested-command suppressions and the obsolete ready-run handoff state.

**Findings**
- No blocking findings on the inspected head.

**Testing**
- Reviewed the complete two-file diff and its callers against `main@e20f350dfdc0e9ab190301e7c7e6589498b3162b` practices for production logic, ccstate reactivity/commands/lifecycle, effects/cache ownership, HTTP handling, and Platform external-behavior tests.
- Existing page-level coverage exercises event-driven startup, authoritative feature-switch hydration, interval refresh, hidden-page polling, run replacement/termination, stale-response containment, and navigation cancellation.
- Exact-head CI completed with 80 successful and 22 intentionally skipped checks, with no pending, failed, cancelled, or action-required checks. `git diff --check` also passes.

PR #33576, opened 2026-09-11
feat(api): trigger resend onboarding on user creation

SCARLETTXYAYA:
## Summary
- Trigger the Resend `user.created` custom event from verified Clerk `user.created` webhooks in production only.
- Upsert the Resend contact name before dispatching the event without changing unsubscribe state.
- Add external-behavior coverage for development/production routing.

## Resend configuration already prepared
- Custom event: `user.created`
- Disabled Automation: `Okou Lifecycle Onboarding | Signup Day 0-7`
- Published templates: Day 0, Day 1, Day 3, and Day 7

The Automation remains disabled until copy and launch approval.

## Verification
- `pnpm --filter api run check-types:gateways`
- `pnpm --filter api run check-types:core`
- `pnpm --filter api run check-types:tests`
- `pnpm --filter api run check-types:bootstrap`
- `pnpm --filter api run check-types:bootstrap-wiring`
- ESLint on changed files
- Targeted Clerk webhook lifecycle tests

### DIFF
--- a/turbo/apps/api/src/__tests__/mocks.ts
+++ b/turbo/apps/api/src/__tests__/mocks.ts
@@ -185,6 +185,10 @@ export interface ApiTestMocks {
   };
   readonly resend: {
     readonly send: AsyncMock;
+    readonly eventsSend: AsyncMock;
+    readonly contactsCreate: AsyncMock;
+    readonly contactsGet: AsyncMock;
+    readonly contactsUpdate: AsyncMock;
   };
   readonly signalTimers: {
     readonly delay: SignalTimerDelayMock;
@@ -329,6 +333,14 @@ export interface ApiTestMocks {
 interface ResendClientMock {
   readonly emails: {
     readonly send: ApiTestMocks["resend"]["send"];
+  };
+  readonly events: {
+    readonly send: ApiTestMocks["resend"]["eventsSend"];
+  };
+  readonly contacts: {
+    readonly create: ApiTestMocks["resend"]["contactsCreate"];
+    readonly get: ApiTestMocks["resend"]["contactsGet"];
+    readonly update: ApiTestMocks["resend"]["contactsUpdate"];
   };
 }
 type SlackWebClientMock = Omit<ApiTestMocks["slack"], "fetchFile">;
@@ -569,6 +581,10 @@ const apiTestMocks: ApiTestMocks = vi.hoisted((): ApiTestMocks => {
     },
     resend: {
       send: vi.fn<(...args: unknown[]) => Promise<unknown>>(),
+      eventsSend: vi.fn<(...args: unknown[]) => Promise<unknown>>(),
+      contactsCreate: vi.fn<(...args: unknown[]) => Promise<unknown>>(),
+      contactsGet: vi.fn<(...args: unknown[]) => Promise<unknown>>(),
+      contactsUpdate: vi.fn<(...args: unknown[]) => Promise<unknown>>(),
     },
     signalTimers: {
       delay:
@@ -964,6 +980,14 @@ vi.mock("resend", () => {
         const client: ResendClientMock = {
           emails: {
             send: apiTestMocks.resend.send,
+          },
+          events: {
+            send: apiTestMocks.resend.eventsSend,
+          },
+          contacts: {
+            create: apiTestMocks.resend.contactsCreate,
+            get: apiTestMocks.resend.contactsGet,
+            update: apiTestMocks.resend.contactsUpdate,
           },
         };
         return client;
@@ -1415,6 +1439,10 @@ export function resetApiTestMocks(): void {
   apiTestMocks.dns.lookupOverrides.clear();
   apiTestMocks.nodeRequest.pinnedAddresses.length = 0;
   apiTestMocks.resend.send.mockReset();
+  apiTestMocks.resend.eventsSend.mockReset();
+  apiTestMocks.resend.contactsCreate.mockReset();
+  apiTestMocks.resend.contactsGet.mockReset();
+  apiTestMocks.resend.contactsUpdate.mockReset();
   apiTestMocks.signalTimers.delay.mockReset();
   apiTestMocks.slack.assistant.threads.setStatus.mockReset();
   apiTestMocks.slack.chat.getPermalink.mockReset();
--- a/turbo/apps/api/src/signals/routes/__tests__/webhooks-clerk-lifecycle.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/webhooks-clerk-lifecycle.test.ts
@@ -1,0 +1,83 @@
+import { beforeEach, describe, expect, it } from "vitest";
+
+import { getApiTestMocks } from "../../../__tests__/mocks";
+import type { TestContext } from "../../../__tests__/test-context";
+import { clearMockedEnv, mockEnv } from "../../../lib/env";
+import { flushWaitUntilForTest } from "../../context/wait-until";
+import { createWebhookCallbackApi } from "./helpers/api-bdd-webhooks";
+
+describe("Clerk user.created lifecycle integration", () => {
+  const mocks = getApiTestMocks();
+  const context: TestContext = {
+    signal: new AbortController().signal,
+    mocks,
+    sessionHistoryBlobs: new Map(),
+  };
+  const webhooks = createWebhookCallbackApi(context);
+  const userCreatedData = {
+    id: "user_123",
+    created_at: 1_757_600_000_000,
+    first_name: "Scarlett",
+    last_name: "Xie",
+    primary_email_address_id: "idn_123",
+    email_addresses: [
+      {
+        id: "idn_123",
+        email_address: "scarlett@example.com",
+      },
+    ],
+  };
+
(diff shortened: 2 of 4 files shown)


e7h4n (changes requested):
Changes Requested

### Summary

This adds a verified production Clerk `user.created` → Resend custom-event path, including contact-name synchronization and route-level tests. The happy path is straightforward, but the handoff is currently neither durably idempotent nor retry-owned, it bypasses the repository's required feature-switch containment, and it accepts identity/protocol fallbacks outside the locked Clerk webhook contract.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace ordering, idempotency, and failure outcomes through the real entry point | Finding (P1) | `webhooks-clerk.ts:533-539` acknowledges before the Resend work completes and never captures the Svix delivery ID; `resend-lifecycle.service.ts:165` sends an event on every invocation. |
| `REVIEW.md` | Preserve webhook authentication and avoid leaking credentials or sensitive payloads | Pass | The new branch runs only after `verifyClerkWebhook`; the added logs contain the Clerk user ID/error but do not log the email or API key. |
| `REVIEW.md` | Add meaningful behavior and regression coverage | Finding (P1) | `webhooks-clerk-lifecycle.test.ts:48-81` covers non-production and create-success only; replay, provider failure, and the advertised existing-contact upsert are uncovered. |
| `REVIEW.md` | Assess independently deployed/persisted compatibility | N/A | No DB shape, internal API response, queue payload, Runner protocol, or persisted JSON contract changes in this diff. |
| `REVIEW.md` | Declare and justify every introduced fallback | Finding (P1) | The parser adds camelCase aliases, first-address substitution, current-time substitution, and empty-name defaults, but the PR body has no `Fallbacks` section. |
| `docs/bad-smell.md` | Handle failures at an owning boundary and do not hide broken invariants with fabricated defaults | Finding (P1) | `resend-lifecycle.service.ts:47-79` tolerates an unproduced camelCase shape and fabricates a registration time instead of rejecting an invalid verified payload. |
| `docs/testing.md` | Exercise the production API entry point and meaningful idempotency/recovery cases | Finding (P1) | The test enters through the Clerk webhook route, but omits the duplicate-delivery, failed-send, and existing-contact branches that define this integration's contract. |
| `docs/testing/testing-external-behavior.md` | Drive webhook behavior through the production endpoint and mock only external dependencies | Pass | `createWebhookCallbackApi()` posts through `webhooksClerkRoutes`; Clerk verification and Resend are mocked at external boundaries. |
| `docs/testing/api-testing.md` | Use the ts-rest route harness and centralized external-service mocks | Pass | The helper uses `setupApp(...)(webhookClerkContract)`, and all new Resend methods are added to the shared API mock registry. |
| `docs/testing/patterns.md` | Await owned detached work rather than synchronizing by delay | Pass | Both cases call `flushWaitUntilForTest()` and add no sleeps or fake timers. |
| `docs/fallback.md` | Do not add impossible/undeclared protocol and data fallbacks | Finding (P1) | The locked Clerk SDK emits required snake_case `created_at`/`email_addresses`; the added alternate shapes/defaults are undeclared, and several are unjustified. |
| `docs/externally-managed-references.md` | Do not substitute another identity when an external reference is absent or unresolved | Finding (P1) | `primaryAddress ?? addresses[0]` can select a non-primary Clerk address when `primary_email_address_id` is null or does not resolve. |
| `.claude/skills/feature-switch/SKILL.md` | Gate every new user-facing integration behind a `FeatureSwitchKey` | Finding (P1) | `ENV === "production"` is the only code gate, so every newly created production user is synchronized immediately and a later provider-side enablement launches globally. |

### Findings

#### P1 — Give the Clerk → Resend handoff durable, idempotent ownership

`turbo/apps/api/src/signals/routes/webhooks-clerk.ts:533-539` starts the provider calls in `waitUntil(...)` and immediately returns `200`. No stable delivery identity is retained, and `turbo/apps/api/src/signals/services/resend-lifecycle.service.ts:165` unconditionally calls `events.send()` every time the route sees the event.

Clerk/Svix delivery is at-least-once and supports replay; Clerk's own guidance explicitly requires idempotent handling and recommends tracking `svix-id`: https://clerk.com/articles/how-to-sync-clerk-user-data-to-your-database#handling-failures-and-ensuring-idempotency. Resend documents that every matching event creates a new Automation run: https://resend.com/docs/dashboard/automations/runs. Therefore a lost response or replay can start duplicate Day 0/1/3/7 runs for one user. Conversely, because the route has already acknowledged before Resend responds, a transient contact/event failure is only logged and Clerk will not retry it, permanently dropping that user's onboarding event.

Please capture the verified delivery identity and establish a durable admission/deduplication contract before acknowledging the webhook (for example, an outbox keyed by the semantic `user.created` identity or `svix-id`, with an explicitly owned retry policy). The installed `resend@6.12.4` `events.send()` surface has no request-options/idempotency-key parameter, so ordinary webhook deduplication cannot be delegated to the current SDK call. Add route-level cases proving a repeated delivery starts one event and a transient provider failure follows the chosen retry/terminal contract.

#### P1 — Put this user-facing integration behind a repository feature switch

`turbo/apps/api/src/signals/routes/webhooks-clerk.ts:533` gates only on event type; `sendUserCreatedLifecycleEvent()` then gates only on `ENV === "production"`. That causes every production signup to send name/email/contact data and the trigger event as soon as the code deploys. Keeping the Resend Automation disabled is a provider-global operational state, not the disabled-by-default or staff-only `FeatureSwitchKey` required for new user-facing integrations, and enabling it later would launch for all future production signups at once without a per-user rollout or code-owned rollback boundary.

Please add a dedicated switch, evaluate it for the created user before any Resend synchronization, and cover production with the switch both off and on. The provider Automation can remain a second safety gate, but it should not replace repository-owned rollout containment.

#### P1 — Parse the exact Clerk webhook identity instead of substituting protocol fields, recipients, or time

The locked `@clerk/backend@3.13.1` contract returns raw `UserJSON` from `verifyWebhook()`: `email_addresses` and numeric `created_at` are required snake_case fields, while `primary_email_address_id` is explicitly nullable. In `turbo/apps/api/src/signals/services/resend-lifecycle.service.ts:46-79`, the new parser nevertheless:

- accepts camelCase `emailAddresses`, `primaryEmailAddressId`, `emailAddress`, and `createdAt`, although that is not the verified wire shape;
- falls back from an unresolved/null primary ID to `addresses[0]`, which can change the recipient identity rather than treating “no primary email” as unavailable; and
- replaces missing or invalid `created_at` with `nowDate()`, turning a malformed payload into a plausible but false registration time.

Require the exact verified snake_case shape, a matching primary address, and a finite `created_at`; skip/quarantine the lifecycle event when those requirements are not met. Do not send to a different address or fabricate the timestamp. If any genuinely optional provider-field default remains, declare it in the PR's required `Fallbacks` section with its justification.

#### P1 — Cover the advertised contact upsert and unsubscribe-preservation behavior

`turbo/apps/api/src/signals/routes/__tests__/webhooks-clerk-lifecycle.test.ts:54-81` only makes `contacts.create()` succeed. It never executes `contacts.get()` or `contacts.update()`, so the PR's advertised behavior—updating an existing contact's name without changing unsubscribe state—is not protected. It also leaves the “contact sync fails but the event still sends” policy unverified. Codecov reports 24 uncovered lines in the new service (69.41% patch coverage overall), which aligns with these unexercised branches.

Add an external-boundary route case where create reports an existing contact, get returns that contact (including an unsubscribed state), update omits `unsubscribed`, and exactly one event is sent. Add the intended contact-sync failure case as well, so a future change cannot silently stop the event or resubscribe a contact.

### Fallbacks

- `resend-lifecycle.service.ts:userCreatedIdentity` camelCase field aliases — protocol fallback; no valid rollout surface or removal gate; **not justified and undeclared**. The locked webhook SDK returns snake_case JSON.
- `resend-lifecycle.service.ts:userCreatedIdentity` `primaryAddress ?? addresses[0]` — external-identity fallback; no valid rollout surface or removal gate; **not justified and undeclared**. A missing primary must not select another address.
- `resend-lifecycle.service.ts:userCreatedIdentity` missing/invalid `created_at` → `nowDate()` — fabricated-data fallback; no valid rollout surface or removal gate; **not justified and undeclared**. `created_at` is required by `UserJSON`.
- `resend-lifecycle.service.ts:sendUserCreatedLifecycleEvent` nullable first/last name → empty strings — optional external-data/presentational fallback; **potentially justified but undeclared**. Confirm that the configured Resend event schema requires these keys; otherwise omit them rather than manufacturing values.

### Testing

- Reviewed commit: `9d9e458fc1021fc0c45b19bf093eff8f9cae58b3`.
- Practices read from `main` at `9e03a65070e78e71cf477967eb6ff9b309c8763e`.
- All current required checks (`ci-gate-crates`, `ci-gate-security`, and `ci-gate-turbo`) are green.
- The added tests use the real Clerk webhook route and centralized external mocks, and they correctly await `waitUntil` work.
- Blocking gaps are the duplicate/retry contract, switch-off/on containment, exact invalid/missing identity behavior, existing-contact update, unsubscribe preservation, and provider-failure behavior described above.
- The out-of-repository Resend event schema and Automation enabled/disabled state were not independently verified by this code review.

PR #33634, opened 2026-09-12
fix(rust): revalidate snapshot bind targets after concurrent cleanup

### LINKED ISSUE #33625: bug(rust): concurrent snapshot cleanup can reject an already-recovered bind target
## Summary

Concurrent snapshot restores can fail with `snapshot drive bind target is not a regular file` after another restore has successfully cleared a stale mount and exposed the valid underlying regular file.

Audited commit: `1088d9c6db489a743487ff3a751cdbebe18b324f`. The affected file was also checked against live `main` before filing and has the same blob (`aa29eb030279a0c8c1a64ffd8e4ec34d05d27b47`).

## Location and cause

- `crates/sandbox-firecracker/src/sandbox/snapshot_restore.rs:77-96`: `ensure_snapshot_drive_bind_target` checks whether the target is regular, then separately checks whether it is a mount point. If the second check is false, it reports the earlier non-regular state without revalidating it.
- `crates/sandbox-firecracker/src/sandbox/snapshot_restore.rs:114-131`: the metadata read and `/proc/self/mountinfo` read are separate observations.
- `crates/sandbox-firecracker/src/sandbox.rs:1368-1369` calls this helper before snapshot launch; `:1394-1406` explicitly documents that these target paths are shared by all sandboxes and that stale inherited mounts can remain after snapshot creation crashes. The per-sandbox `unshare --mount` isolation happens later and does not serialize these host-side checks.

Source: https://github.com/vm0-ai/vm0/blob/1088d9c6db489a743487ff3a751cdbebe18b324f/crates/sandbox-firecracker/src/sandbox/snapshot_restore.rs#L64-L97

## Concrete interleaving

Start with the supported recovery condition: a stale device bind mount covers the shared regular-file target.

1. Restores A and B enter `ensure_snapshot_drive_bind_target` for the same path; creation reports `AlreadyExists`.
2. B obtains metadata for the still-mounted device, so its regular-file check returns false.
3. A successfully unmounts the stale device and verifies the underlying regular file.
4. B reads mountinfo after A's cleanup, finds no mount at that path, skips the unmount/revalidation block, and returns `SandboxError::Start` at lines 91-96.

The target is already valid at step 4. This produces an avoidable snapshot-start failure during concurrent recovery; it is not an invalid user path. The second regular-file check at lines 86-88 handles cleanup races only for callers that first observed a mount at line 81, not this interleaving.

## Verification

A deterministic, isolated filesystem probe included the production `snapshot_restore.rs` unchanged and intercepted only the asynchronous metadata boundary:

- Obtain real metadata for a temporary non-regular inode (a Unix socket, used because the helper only tests `is_file()`).
- Have another joined thread replace that temporary path with a regular file before returning the already-obtained metadata. This models the visible path-type transition caused by the other restore's unmount, without mounting anything.
- Allow the real mountinfo check and remaining production control flow to run.

Observed result: the helper returns `SandboxError::Start` containing `not a regular file`, while a fresh metadata read confirms the target is regular. An immediate second call succeeds on the same unchanged file.

All 10 existing tests from the included source also passed; the new interleaving probe reproduced the failure. The existing `snapshot_drive_bind_target_allows_concurrent_first_use` test covers two creators of a missing path, not concurrent stale-mount recovery. No native Firecracker, KVM, NBD, mount, or namespace operation was run, and repository source was not modified.

## Acceptance criteria

- Revalidate the target before rejecting a path when mount state may have changed after the first metadata observation, or otherwise coordinate shared-target recovery safely.
- Add a deterministic regression test for one caller completing stale-mount cleanup between another caller's metadata and mountinfo checks. Both callers must succeed when the final target is the valid regular file.
- Continue rejecting genuinely invalid directories/symlinks/non-regular targets. Do not delete, replace, or truncate the existing shared bind-target file as part of the fix; other mount namespaces rely on its inode/dentry.

seven332:
## Summary

Concurrent snapshot restores could reject a valid shared bind target when a peer cleared its stale mount between the metadata and mountinfo reads. Revalidate the target after optional mount cleanup so completed peer recovery succeeds, preserving the existing file's identity and contents and continuing to reject invalid targets.

Add a deterministic regression that delays a real non-regular metadata observation until a joined peer has exposed the regular backing file and completed preparation. It checks both callers' results, device/inode identity, and contents. A stable Unix socket rejection test complements the existing directory and symlink cases. The change is limited to host-side bind-target preparation; public contracts and privileged mount execution are unchanged.

Closes #33625

## Validation

- The new concurrent-cleanup regression failed with `not a regular file` before the behavior change and passed after it.
- `PATH="/usr/sbin:/sbin:$PATH" cargo test --manifest-path crates/Cargo.toml --profile local --locked -p sandbox-firecracker -j 2 --quiet` — 891 unit tests and 1 integration test passed. Two existing native Firecracker tests were ignored as configured. The sandbox's default PATH omitted the installed `mkfs.ext4`; adding the system binary paths resolved the initial environment-only failure.
- `cargo fmt --manifest-path crates/Cargo.toml -p sandbox-firecracker -- --check` — passed.
- `cargo clippy --manifest-path crates/Cargo.toml --profile local --locked -p sandbox-firecracker --all-targets --all-features -j 2` — passed.
- `cargo doc --manifest-path crates/Cargo.toml --profile local --locked -p sandbox-firecracker --all-features --no-deps -j 2` — passed.
- `git diff --check` — passed.

The regression models the mount-removal transition at the OS metadata boundary using real temporary files; native KVM, NBD, mount, and namespace behavior was not exercised locally.

### DIFF
--- a/crates/sandbox-firecracker/src/sandbox/snapshot_restore.rs
+++ b/crates/sandbox-firecracker/src/sandbox/snapshot_restore.rs
@@ -62,6 +62,16 @@ pub(super) async fn load_snapshot_and_apply_rate_limits(
 }
 
 pub(super) async fn ensure_snapshot_drive_bind_target(path: &Path) -> Result<(), SandboxError> {
+    ensure_snapshot_drive_bind_target_with_metadata(path, async |path: &Path| {
+        tokio::fs::symlink_metadata(path).await
+    })
+    .await
+}
+
+async fn ensure_snapshot_drive_bind_target_with_metadata(
+    path: &Path,
+    mut read_metadata: impl AsyncFnMut(&Path) -> io::Result<std::fs::Metadata>,
+) -> Result<(), SandboxError> {
     if let Some(parent) = path.parent() {
         tokio::fs::create_dir_all(parent)
             .await
@@ -74,7 +84,7 @@ pub(super) async fn ensure_snapshot_drive_bind_target(path: &Path) -> Result<(),
         return Ok(());
     }
 
-    if snapshot_drive_bind_target_is_regular_file(path).await? {
+    if snapshot_drive_bind_target_is_regular_file(path, &mut read_metadata).await? {
         return Ok(());
     }
 
@@ -83,9 +93,12 @@ pub(super) async fn ensure_snapshot_drive_bind_target(path: &Path) -> Result<(),
         if create_snapshot_drive_bind_target_file(path).await? {
             return Ok(());
         }
-        if snapshot_drive_bind_target_is_regular_file(path).await? {
-            return Ok(());
-        }
+    }
+
+    // Another restore may clear the mount after our first metadata read but
+    // before the mountinfo read. Revalidate even when no mount was observed.
+    if snapshot_drive_bind_target_is_regular_file(path, &mut read_metadata).await? {
+        return Ok(());
     }
 
     Err(SandboxError::Start {
@@ -111,12 +124,13 @@ async fn create_snapshot_drive_bind_target_file(path: &Path) -> Result<bool, San
     }
 }
 
-async fn snapshot_drive_bind_target_is_regular_file(path: &Path) -> Result<bool, SandboxError> {
-    let meta = tokio::fs::symlink_metadata(path)
-        .await
-        .map_err(|e| SandboxError::Start {
-            message: format!("stat snapshot drive bind target: {e}"),
-        })?;
+async fn snapshot_drive_bind_target_is_regular_file(
+    path: &Path,
+    read_metadata: &mut impl AsyncFnMut(&Path) -> io::Result<std::fs::Metadata>,
+) -> Result<bool, SandboxError> {
+    let meta = read_metadata(path).await.map_err(|e| SandboxError::Start {
+        message: format!("stat snapshot drive bind target: {e}"),
+    })?;
     Ok(meta.file_type().is_file())
 }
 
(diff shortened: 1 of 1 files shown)


seven332 (commented):
LGTM

**Reviewed head:** `fdf63f16928e1c820b01445db99b473f075277dc`
**Practice revision:** `a10919fcf6f4aadf4706900147e2038e5f10010b`

**Summary:** Final non-following metadata validation now covers both mount outcomes, so a peer's completed cleanup is accepted. Exclusive creation, shared inode/content preservation, and invalid-target/error behavior remain intact. The six-pass self-review completed on this unchanged head.

**Findings:** No findings.

**Testing:** The new synchronized regression failed before the behavior change and passed afterward. The affected crate passed 891 unit tests and 1 integration test; formatting, Clippy, docs, and diff checks passed. Two existing native Firecracker tests were ignored as configured. The fixture explicitly models privileged mount removal at the OS metadata boundary with real temporary files; native mount/KVM/NBD/namespace behavior was not verified locally. PR CI is evaluated separately.

This is a COMMENT review because seven332 owns the PR; it does not supply another reviewer's approval.

seven332:
## Code Review: PR #33634

**Reviewed head:** `fdf63f16928e1c820b01445db99b473f075277dc`

### Summary
Completed the six-pass self-review and code-quality review. The final metadata check accepts completed peer cleanup while preserving shared-file identity and invalid-target rejection. [Commit-bound review](https://github.com/vm0-ai/vm0/pull/33634#pullrequestreview-5185534822).

### Findings
No findings.

### Testing
Regression verified red/green; 891 unit tests and 1 integration test passed, along with formatting, Clippy, and documentation checks. Two existing native Firecracker tests were ignored as configured; privileged mount behavior remains unverified locally.

### Verdict
LGTM. The own-PR review uses COMMENT and does not replace any required approval from another reviewer. CI readiness is separate.

Issue #33658, opened 2026-09-12
refactor(api): finish the restrictedVm0Models to restrictedBuiltInModels rename (step 1, expand)


lancy:
Finish the `restrictedVm0Models` → `restrictedBuiltInModels` rename. **The rename was already done on the storage side and stops at the read boundary**, where the old brand name is re-introduced and then travels out through the public `/api/billing/status` response.

This is a live cross-version API contract between the API and the platform web app, which deploy independently. It needs expand → migrate → contract, not a rename.

## The half-finished state

Already renamed:

| Layer | Name |
|---|---|
| Postgres column | `restricted_built_in_models` |
| Drizzle field (`db/src/schema/org-plan-entitlement.ts:74,86`) | `restrictedBuiltInModels` |
| `org-plan-entitlement-tier-values.ts` | `restrictedBuiltInModels` |
| `org-plan-entitlements.service.ts:114` | `restrictedBuiltInModels` |

Then `turbo/apps/api/src/signals/services/org-plan-entitlement-read.service.ts:97` maps it straight back:

```ts
const { restrictedBuiltInModels, ...runtimeCapabilities } = capabilities;
return {
  ...
  restrictedVm0Models: restrictedBuiltInModels,
```

From that line outward everything still carries the retired brand: `billing-status.service.ts`, `model-policy.service.ts`, `model-selection.service.ts`, `run-admission.service.ts`, the public contract at `turbo/packages/api-contracts/src/contracts/billing.ts:97`, both platform capability modules, and roughly 80 test references.

## Why this cannot be one commit

`billing.ts:97` declares `restrictedVm0Models: z.boolean().optional()` on the `/api/billing/status` response. The platform consumes it at `apps/platform/src/signals/okou-page/org-plan-capabilities.ts:114-115`:

```ts
restrictedVm0Models:
  billing.restrictedVm0Models ?? fallback.restrictedVm0Models,
```

API and app ship as separate releases — `api-v1.587.0` and `app-v0.890.0` are distinct tags. If the API renames the response field first, the platform reads `undefined` and **silently falls back to `LEGACY_TIER_CAPABILITIES`**, a hardcoded per-tier table. Because the field is already `.optional()`, nothing errors and nothing logs.

[...]

Issue #33497, opened 2026-09-11
refactor(api): emit X-Okou signature headers and drop the legacy names


lancy:
## Parent

Final step of EPIC **#33501**, renaming the HMAC signature headers `X-VM0-Signature` / `X-VM0-Timestamp` to `X-Okou-*`. Step 1 is **#33496**, merged, controller accepted, and released in `api-v1.585.1`.

**This issue was rewritten on 2026-09-11.** It originally moved only the senders, with a separate #33498 removing legacy acceptance a release later. That separation is no longer justified and #33498 is closed as superseded. Refreshed against `main` at `85b31f1265`.

## Why the remaining two steps collapsed into one

The three-step staging rested on this reasoning: callback delivery is single-attempt, `agent_run_callbacks` holds ~290,000 rows, so a deploy-skew window between senders and verifiers would silently lose callbacks.

The count was right. The attribution was wrong, established while verifying step 1 in production:

**1. The callback-route surface is unreachable.** `callbackRoute()` in `lib/callback-route/callback-route.ts` has **no production caller** — only its own definition and tests. Nothing mounts it, so the verification path it wraps never executes.

**2. Almost nothing sends these headers over HTTP.** `agent-run-callback.service.ts:438-457` branches on `internalKind`. A callback carrying one goes to `dispatchInternalCallback$`, a `switch` dispatching to in-process handlers with no `fetch` and no headers at all. Only `internalKind IS NULL` reaches `dispatchHttpCallback`, the sole place the signature headers are attached.

| Population | Count | Path |
|---|---|---|
| With an `internal_kind` (`chat` 247,820, plus slack/teams/feishu/cron/loop/…) | ~262,000 | in-process, never sends headers |
| `internal_kind IS NULL` | ~28,100 | HTTP with signature headers |

**3. Even the HTTP population is dormant.** Its newest row is `2026-08-29T23:01:11`, and rows arrive in a daily-pair pattern rather than from live traffic.

**4. The workflow-webhook surface is dormant.** All 14 `webhook-received` automations are disabled, with no delivery since `2026-07-15`.

[...]

Issue #33496, opened 2026-09-11
refactor(api): accept X-Okou signature headers alongside the legacy names


lancy:
## Parent

Step 1 of 3 in renaming the HMAC signature headers `X-VM0-Signature` / `X-VM0-Timestamp` to `X-Okou-*`. Steps 2 and 3 are the follow-up issues linked below and each is gated on the previous step reaching production. Refreshed against `main` at `c8527891b4f` on 2026-09-11.

## Why three steps

These headers are signed by the API and verified by the API. `agent-run-callback.service.ts:659-660` and `agent-run-callbacks.service.ts:148-149` POST them to `callback.url`, which is always an internally constructed URL reached over the public alias, and `callback-route.ts:140-141` / `webhooks-workflow-automations.ts:51-52` verify them.

Production scale: `agent_run_callbacks` holds **290,013** rows, of which 247,820 are `internal_kind = chat` — the callbacks that drive chat runs.

**Delivery is single-attempt.** `markCallbackAttemptStarted` sets `attempts: 1` rather than incrementing, and there is no timed retry. `dispatchRunCallbacks` re-selects `pending`/`failed` rows only when it is called again for that `runId`, which happens on a further run-status transition — a terminal run has none. A callback rejected during a deploy window is therefore lost, not retried.

A one-shot rename would produce exactly that loss in both directions during promotion: draining old instances signing the old name against new verifiers, and new instances signing the new name against draining old verifiers. Hence expand → migrate → contract.

| Release | Change | Why it is safe |
|---|---|---|
| **1 — this issue** | Verifiers accept **both** names, preferring the new one. Senders and snippets unchanged. | Purely additive. Existing traffic keeps working unchanged; nothing yet emits the new name. |
| **2 — follow-up** | Senders and the UI/CLI snippets emit the new name. Verifiers still accept both. | After release 1 every live instance accepts both, so neither direction of deploy skew can reject. |
| **3 — follow-up** | Verifiers stop accepting the old name. | Gated on evidence that the legacy path has stopped firing (see below). |

Rollback stays safe at every step: release 3 rolls back to a build that accepts both; release 2 rolls back to a build whose senders emit the old name, which release-1 verifiers accept. No schema change, no rollback barrier, no timing window, no maintenance notice.

## No external consumer exists

Established on 2026-09-11 and recorded so later steps do not re-litigate it:

[...]

PR #33660, opened 2026-09-12
refactor: remove inert retired-brand vm0 residue

### LINKED ISSUE #33657: refactor: remove inert retired-brand vm0 residue in headers and display helpers
Remove three pieces of retired-brand `vm0` residue that are each dead or unreachable today. One focused PR; no expand/contract needed for any of them.

These are deliberately grouped because all three are provably inert. The larger `restrictedVm0Models` rename found in the same sweep is **not** here — it is a live cross-version API contract and is filed separately.

## 1. `x-vm0-idempotency-key` — the missed sibling of the signature-header rename

`turbo/apps/api/src/signals/services/workflow-webhook-automation.service.ts:287`

```ts
const explicitKey = headerValue(args.headers, "x-vm0-idempotency-key");
```

Rename to `x-okou-idempotency-key`.

**Why this is the same family.** Thirty-nine lines above, in the same file and the same inbound webhook handler, `sanitizedHeaders()` already reserves the renamed names:

```ts
lower === "x-okou-signature" ||
lower === "x-okou-timestamp" ||
```

#33496 and #33497 renamed the signature pair and left the idempotency key behind.

**Why a single step is safe.** This is the only occurrence of the string in the entire repository, it appears in no `docs/` page and no `*.md`, and the endpoint that reads it received **zero requests in the last three days** — `/api/webhooks/workflow-automations` does not appear anywhere in the top twenty webhook paths by volume. That matches the earlier finding that the workflow webhook trigger had not fired in 58 days. Three-day retention is the whole available window; do not claim more.

**Say the degradation mode in the PR body anyway.** This header is inbound — a caller *sends* it to control delivery deduplication. If a caller were sending the old name after this ships, `deliveryKeyForRequest` would silently fall through to the derived `sha256(timestamp.signature.bodyHash)` key rather than erroring. That is a quiet behavior change, not a loud one, which is exactly why it deserves a sentence rather than a shrug.

## 2. `x-vm0-client-request-id` — a dead test fixture

`turbo/apps/api/src/signals/routes/__tests__/image-recognition.test.ts:184`

The header the API actually reads is `X-Client-Request-Id`, defined brand-neutrally at `turbo/packages/api-contracts/src/contracts/client-headers.ts:7` and consumed at `app-factory-core.ts:407`. Nothing reads `x-vm0-client-request-id`; it is the only occurrence in the repository.

**Check before deleting.** Confirm the surrounding assertion still proves what it claims. If the test asserts that a client request id is propagated into the request log, sending a header the server ignores means the assertion may be passing for the wrong reason today. If so, fix the root cause by sending `X-Client-Request-Id` — do not simply drop the header and weaken the assertion.

## 3. `"VM0"` in the usage title-casing acronym list

`turbo/apps/platform/src/lib/credit-usage-display.ts:80`

```ts
if (["AI", "API", "GLM", "GPT", "ID", "SQL", "URL", "VM0"].includes(upper)) {
```

Verified dead: no usage display name, usage kind, or model identifier in `db/src/schema/**` or `api-contracts/**` contains `vm0`, so `titleCaseUsageToken` can never receive it. Remove only the `"VM0"` entry; the other seven are live.

The `hashtext('vm0')` in `usage-event-compaction-lock.service.ts:27` is a Postgres advisory-lock namespace, not a display name. **Do not touch it** — changing it would change the lock id and could let two compaction runs overlap during a deploy.

## Do not touch

- `restrictedVm0Models` — filed separately; it is a live API contract.
- The 11 `vm0.*` localStorage keys in `apps/platform/src/signals/**`. Renaming `vm0.googleAds*ConversionRecorded` resets the already-reported flags and causes duplicate Google Ads conversion events for every existing user. Owner has deferred these.
- The `repositoriesPlaceholder: "vm0-ai/vm0, …"` strings in the 12 locale files. Owner has deferred these.
- Live infrastructure: the `vm0-ai` GitHub org, `static.vm0.io` / `cdn.vm0.io` / `sites.vm0.io`, `vm0-runner*` systemd units, `vm0-ns-*` namespaces, Axiom dataset names such as `vm0-request-log-prod`, and `api.vm0.dev`.
- `PUBLIC_BRANDS` and the `z.enum(["vm0", "okou"])` brand schemas — `vm0` is still a supported public brand.
- The mitm-addon internal headers `x-vm0-connector-intent`, `x-vm0-codex-model-catalog-prefetch`, `x-vm0-test`, `x-vm0-test-endpoint-bypass`, and the first-party `x-vm0-web-origin`. Out of scope for this issue.
- Do not weaken, delete, or invert any assertion, and do not add a test asserting an old name is absent.

## Acceptance

1. `x-vm0-idempotency-key` returns zero hits; `x-okou-idempotency-key` is read at the same call site.
2. `x-vm0-client-request-id` returns zero hits, and the test it came from still asserts the same behavior — with its root cause fixed if the check in section 2 found one.
3. `"VM0"` is gone from the acronym list and the other seven entries are unchanged.
4. `hashtext('vm0')` is byte-identical.
5. Only these three files change, plus any file required by a root-cause fix under section 2, which must be called out in the PR body.

lancy:
Closes #33657.

Removes three pieces of retired-brand `vm0` residue that are each dead or unreachable today. Three files, four lines. No expand/contract needed for any of them.

`restrictedVm0Models` is deliberately **not** here — it is a live cross-version API contract and is filed separately as #33658.

## 1. `x-vm0-idempotency-key` → `x-okou-idempotency-key`

`turbo/apps/api/src/signals/services/workflow-webhook-automation.service.ts:287`

The missed sibling of #33496 / #33497. Thirty-nine lines above, in the same file and the same inbound webhook handler, `sanitizedHeaders()` already reserves `x-okou-signature` and `x-okou-timestamp`; only the idempotency key was left behind.

**Why a single step is safe.** This was the only occurrence of the string in the entire repository, it appears in no `docs/` page and no `*.md`, and the endpoint that reads it received zero requests over the last three days — three days is the whole available retention window, so this is not a claim about any longer period.

**Degradation mode, stated explicitly.** This header is *inbound*: a caller sends it to control delivery deduplication. If any caller is still sending the old name after this ships, `deliveryKeyForRequest` will **silently** fall through to the derived `sha256(timestamp.signature.bodyHash)` key rather than erroring. Requests keep succeeding; only the dedupe key changes, and only for a caller sending an explicit key. That quiet path is the reason this is called out rather than shrugged off.

## 2. `x-vm0-client-request-id` — root cause fixed, not deleted

`turbo/apps/api/src/signals/routes/__tests__/image-recognition.test.ts:184`

**The pre-change check found a real problem, so the header was fixed rather than dropped.**

The test `recognizes one owned image and settles each real invocation` sends the *same* `clientRequestId` across two invocations, then asserts both actually ran and both were billed: two OpenRouter request bodies, `raw: 6` usage rows, and `credits: EXPECTED_CHARGE * 2`. The behavior it intends to prove is that repeating a client request id does not collapse two real invocations into one billed event.

It was passing for the wrong reason. The header was spelled `x-vm0-client-request-id`, which nothing on the server reads — the canonical header is `X-Client-Request-Id`, defined at `api-contracts/src/contracts/client-headers.ts:7` and consumed at `app-factory-core.ts:407`. The server never saw a repeated client request id at all, so non-deduplication held trivially and the assertion never exercised the path it claimed to.

The fix sends the canonical header via the exported `CLIENT_REQUEST_ID_HEADER` constant, which also ties the test to the contract instead of a loose string. **No assertion was weakened, deleted, or inverted, and no "old name is absent" tombstone test was added** — the assertions are byte-identical and now genuinely exercised.

This is contained to the same test file, so the change set is still exactly the three files named in the issue.

Verified the header now really reaches the middleware: the test harness builds its app through the real `createAppWithRoutes` from `app-factory-core`, the same module that reads `CLIENT_REQUEST_ID_HEADER` into the `x_client_request_id` log field. Confirmed the constant is consumed only as a request-log field and has no server-side dedup behavior, so the two-invocation assertions still hold.

## 3. `"VM0"` in the usage title-casing acronym list

`turbo/apps/platform/src/lib/credit-usage-display.ts:80`

Removed only the `"VM0"` entry; the other seven (`AI`, `API`, `GLM`, `GPT`, `ID`, `SQL`, `URL`) are live and unchanged.

The `hashtext('vm0')` advisory-lock namespaces in `usage-event-compaction-lock.service.ts` and `chat-event-retention-lock.service.ts` are **untouched and byte-identical** — changing them would change the lock id and could let two compaction runs overlap during a deploy.

## Acceptance

1. ✅ `x-vm0-idempotency-key` returns zero hits repo-wide; `x-okou-idempotency-key` is read at the same call site.
2. ✅ `x-vm0-client-request-id` returns zero hits; the test asserts the same behavior, with the root cause from section 2 fixed.
3. ✅ `"VM0"` gone; the other seven entries unchanged.
4. ✅ `hashtext('vm0')` byte-identical (verified by empty diff on both lock services).
5. ✅ Exactly three files changed. The section 2 root-cause fix landed inside the same test file, so no additional file was required.

## Verification

Targeted checks only, per the repo's affected-workspace guidance — no full suite, no dev server.

- `image-recognition.test.ts` — **9/9 passed** against a local Postgres with migrations applied. This is the check that matters for section 2: the assertions still hold now that the server actually receives the header.
- `credit-usage-display.test.tsx` — **3/3 passed**.
- Prettier: clean on all three files. oxlint: clean (exit 0) on all three files.

### Known sandbox limitation — not claiming a local pass

`webhooks-workflow-automations.test.ts` has one failing test, `dispatches signed webhook deliveries and de-duplicates retries`, failing at `claimRunnerJob` with `404 Job not found in queue`. The sandbox cannot schedule runner jobs.

Because that test's name mentions deduplication, it was checked rather than assumed:

- **Same-environment baseline:** the identical test fails the identical way on unmodified `origin/main` in this sandbox — same test, same 404, same `1 failed | 3 passed`. The failure is environmental and pre-existing, not introduced here.
- **No causal link to this change:** that test sends no idempotency header at all (zero `idempotency` matches in the file), so it only ever exercised the derived `sha256(...)` path, which this PR does not touch.

CI is the authority for this file. This is **not** a claim that the suite passes locally.

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/image-recognition.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/image-recognition.test.ts
@@ -1,6 +1,7 @@
 import { randomUUID } from "node:crypto";
 
 import { createStore } from "ccstate";
+import { CLIENT_REQUEST_ID_HEADER } from "@okouai/api-contracts/contracts/client-headers";
 import {
   IMAGE_RECOGNITION_MAX_FILE_BYTES,
   imageRecognitionContract,
@@ -181,7 +182,7 @@ function requestImageRecognition(args: {
   const headers = {
     ...(args.token ? { authorization: `Bearer ${args.token}` } : {}),
     ...(args.clientRequestId
-      ? { "x-vm0-client-request-id": args.clientRequestId }
+      ? { [CLIENT_REQUEST_ID_HEADER]: args.clientRequestId }
       : {}),
   };
   const client = setupApp({
--- a/turbo/apps/api/src/signals/services/workflow-webhook-automation.service.ts
+++ b/turbo/apps/api/src/signals/services/workflow-webhook-automation.service.ts
@@ -284,7 +284,7 @@ function deliveryKeyForRequest(args: {
   readonly timestamp: string;
   readonly headers: Readonly<Record<string, string>>;
 }): string {
-  const explicitKey = headerValue(args.headers, "x-vm0-idempotency-key");
+  const explicitKey = headerValue(args.headers, "x-okou-idempotency-key");
   if (explicitKey && explicitKey.trim().length > 0) {
     return explicitKey.trim();
   }
--- a/turbo/apps/platform/src/lib/credit-usage-display.ts
+++ b/turbo/apps/platform/src/lib/credit-usage-display.ts
@@ -77,7 +77,7 @@ const MODEL_DISPLAY_NAMES: Readonly<Record<string, () => string>> = {
 
 function titleCaseUsageToken(token: string): string {
   const upper = token.toUpperCase();
-  if (["AI", "API", "GLM", "GPT", "ID", "SQL", "URL", "VM0"].includes(upper)) {
+  if (["AI", "API", "GLM", "GPT", "ID", "SQL", "URL"].includes(upper)) {
     return upper;
   }
 


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: 8c11d60740228582cd13e6451d069a4506eaca6c
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `8c11d60740228582cd13e6451d069a4506eaca6c`
Practice revision: `8451ce0e82e70cfeb16bb83441892c4022e38014` (`main`)

### Summary

Removes three inert retired-brand `vm0` residues: renames the inbound webhook header `x-vm0-idempotency-key` to `x-okou-idempotency-key`, switches a dead test fixture header to the canonical `X-Client-Request-Id` contract constant, and drops the unreachable `"VM0"` entry from the usage title-casing acronym list. Three files, +4/−3.

The section-2 root-cause investigation required by the issue was actually performed and found a real defect, and the fix repairs it rather than deleting the header and weakening the assertion.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | `deliveryKeyForRequest` traced to its only caller path via `webhooks-workflow-automations.ts`; `titleCaseUsageToken` traced through `formatUsageDisplayName` to `getCreditUsageDisplayName`; `CLIENT_REQUEST_ID_HEADER` traced to `app-factory-core.ts:407`. |
| `REVIEW.md` | Existing coverage still protects a refactor; identify the concrete missing scenario | Pass | `credit-usage-display.test.tsx` 3/3 and `image-recognition.test.ts` 9/9 still pass. The explicit-key branch was untested before and after this diff — a pre-existing gap, not introduced here (see Observations). |
| `REVIEW.md` | Tests verify externally observable state or an HTTP response, not internal state or mock calls | Pass | The touched test asserts HTTP status/body, MSW-observed outbound request bodies, and usage rows read back through `GET /api/usage-record`. No new internal assertion. |
| `REVIEW.md` | API tests must not assert log records (outside the logger suite and one redaction check) | Pass | `X-Client-Request-Id` is sent as request context only; nothing asserts on `x_client_request_id` or any log record. |
| `REVIEW.md` | Report violations with rule and source location; distinguish introduced defects from unchanged code | Pass | The one behavioral risk (silent derived-key fallthrough) is unchanged pre-existing code, and is recorded in the PR body rather than raised as an introduced defect. |
| `REVIEW.md` | Persisted data, schema, response validation, and `jsonb().$type<>()` reviewed together | N/A | No schema, migration, contract shape, or persisted-state change in the diff. |
| `docs/bad-smell.md` | §1 No `any`; §2 no lint/type suppressions | Pass | No `any` and no `eslint-disable` / `oxlint-disable` / `@ts-*` / `prettier-ignore` added. oxlint exits 0 on all three files. |
| `docs/bad-smell.md` | §4 Interface changes: document changed public contracts and assess consumers | Pass | The inbound header is the changed contract. Verified repo-wide it has no second occurrence, no `docs/` or `*.md` mention, and no UI/CLI/i18n string instructing callers to send it. The PR body documents the change and its consumer impact. |
| `docs/bad-smell.md` | §5 Static imports in production and test utilities | Pass | `CLIENT_REQUEST_ID_HEADER` is added as a static top-level import. |
| `docs/bad-smell.md` | §3 No code whose only output is a log level; §6 no hardcoded URLs/config | N/A | No error-handling, logging, or configuration path changed. |
| `docs/bad-smell.md` | §7 No fallback for a state the owning contract already rules out | Pass | No new branch added; the diff renames the trigger of an existing branch. |
| `docs/fallback.md` | §1 No negative/tombstone tests asserting removed behavior stays removed | Pass | No "old name is absent" test added. The existing assertions are byte-identical, not rewritten. |
| `docs/fallback.md` | §5 Do not keep a reader for a retired producer | Pass | This is the cleanup: the old-name reader is renamed rather than retained alongside the new name, so no dual-accept branch is introduced. |
| `docs/fallback.md` | §9 Every introduced fallback declared in a `Fallbacks` PR-body section | N/A | The diff introduces no fallback. §9 explicitly exempts PRs that only keep or remove existing fallbacks, and does not require `Fallbacks: none`. |
| `docs/fallback.md` | §10 Removal carries type, single-writer, production, or rollout evidence | Pass | PR body carries single-occurrence evidence, no-documentation evidence, and 3-day endpoint traffic evidence, and explicitly refuses to claim a window longer than the retention period. |
| `docs/testing.md` | API boundary: call production endpoints for setup and verification | Pass | The test drives the real Hono app through `setupApp()` and the ts-rest contract; verification reads back through `GET /api/usage-record`. |
| `docs/testing.md` | Changes to requests or protocols must cover old/new combinations that coexist during rollout | Pass | See `docs/deployment-compatibility.md` row: the old name has no supported external producer, so no mixed-version pairing is reachable. |
| `docs/testing.md` | Mock external services at their boundary; use real internal code and real DB | Pass | Unchanged: MSW handles OpenRouter, `context.mocks` handles Clerk/S3. No `vi.mock()` of an internal module added. |
| `docs/testing/testing-external-behavior.md` | Give the system the same context an external user can provide | Pass | This is the substance of the fix: the test now sends the header a real client actually sends, instead of one no production code reads. |
| `docs/testing/testing-external-behavior.md` | Lint boundary: API test files must not import DB schema, API service files, or the logger | Pass | The new import is `@okouai/api-contracts/contracts/client-headers`, a contract package — the same class as the file's existing contract imports. |
| `docs/testing/api-testing.md` | Route tests go through `setupApp()` + contract, not by importing handlers or services | Pass | `requestImageRecognition` continues to use `setupApp({context, routes: imageRecognitionRoutes})(imageRecognitionContract)`. |
| `docs/deployment-compatibility.md` | Before removing a request variant, confirm no explicitly supported external caller can still use the old variant | Pass | The old header appears in no documentation, no public contract, no SDK/CLI/UI string, and has no second code occurrence, so it was never an explicitly supported external variant. Endpoint traffic over the full 3-day retention window is zero. |
| `docs/deployment-compatibility.md` | Backend change safe for old/new frontend and old/new runner pairings | Pass | No first-party client sends this header; frontend and runner are unaffected. The platform change is display-only formatting. |

### Findings

No P0 or P1 findings.

**Non-blocking observations** (do not require changes):

1. **The explicit-idempotency-key branch has no test coverage, before or after this PR.** `webhooks-workflow-automations.test.ts` sends no idempotency header at all, so it only ever exercised the derived `sha256(timestamp.signature.bodyHash)` path. The rename therefore neither gains nor loses coverage. This is a pre-existing gap in unchanged code, and `REVIEW.md` directs reviewers to distinguish introduced defects from unchanged code, so it does not block. Worth a follow-up if the explicit-key path is intended to stay supported.

2. **The three-day traffic evidence is bounded by retention, not by a chosen window.** The PR body states this correctly and does not overclaim. The decisive evidence here is not the traffic window but the absence of any documented or first-party producer of the old header name, which is what `docs/deployment-compatibility.md` actually asks to confirm.

3. **`"VM0"` removal is cosmetic-only in the worst case.** `kind` and `provider` are free-form `varchar` columns rather than enums, so a `vm0` token is not type-impossible; but the failure mode if one ever appeared is a label rendering as `Vm0` instead of `VM0`. No correctness or crash risk.

### Testing

**Verification observed in the PR body and independently consistent with the diff:**

- `image-recognition.test.ts` — 9/9 passed against a local Postgres with migrations applied. This is the load-bearing check: it confirms the two-invocation billing assertions still hold now that the server genuinely receives the header.
- `credit-usage-display.test.tsx` — 3/3 passed.
- Prettier clean and oxlint clean (exit 0) on all three files.

**On the section-2 root-cause fix.** The test `recognizes one owned image and settles each real invocation` sends the same `clientRequestId` across two invocations and asserts both ran and both billed (two OpenRouter bodies, `raw: 6`, `credits: EXPECTED_CHARGE * 2`). Because the header was spelled `x-vm0-client-request-id`, the server never saw a repeated client request id, so non-deduplication held trivially — the assertion was passing for the wrong reason. Sending the canonical `X-Client-Request-Id` via the exported constant makes the same assertion genuinely exercised. I verified independently that the test harness builds its app through the real `createAppWithRoutes` from `app-factory-core`, the same module that reads `CLIENT_REQUEST_ID_HEADER`, and that the constant is consumed only as a request-log field with no server-side dedup behavior — so the assertions correctly still hold. No assertion was weakened, deleted, or inverted.

**Material gaps:** none that block. See observation 1 for the pre-existing explicit-key coverage gap.

**Known environment limitation, correctly reported.** `webhooks-workflow-automations.test.ts` has one failing test (`dispatches signed webhook deliveries and de-duplicates retries`) failing at `claimRunnerJob` with `404 Job not found in queue`. The PR body does not claim a local pass, and it establishes the failure is environmental two independent ways: the identical test fails identically on unmodified `origin/main` in the same sandbox (same 404, same `1 failed | 3 passed`), and that test sends no idempotency header, so it never exercised the renamed branch. CI remains the authority for that file.

PR #33663, opened 2026-09-12
refactor(runner): rename sandbox-internal and preview-only x-vm0 headers

### LINKED ISSUE #33659: refactor(runner): rename the sandbox-internal and preview-only x-vm0 headers
Rename two groups of retired-brand `x-vm0-*` headers whose sender and receiver ship together, so no cross-version window exists.

`x-vm0-connector-intent` is **deliberately excluded** — it crosses a published-CLI boundary and needs expand/contract. See the exclusion section; it is not an oversight.

## Group 1 — `x-vm0-codex-model-catalog-prefetch` → `x-okou-codex-model-catalog-prefetch`

8 files, all under `crates/runner/mitm-addon/`, **zero** references anywhere in `turbo/`.

This one is genuinely atomic. `crates/runner/build.rs` `generate_addon_files()` scans `mitm-addon/src` and embeds every file into the runner binary via the generated `ADDON_FILES` table; `crates/runner/src/proxy/process.rs:270` extracts them at runtime. The addon and the runner that ships it are **one build artifact**, replaced together. `docs/deployment-compatibility.md:396-401` states this. It is the same property that let #33492 rename the `vm0_*` addon options in a single step.

## Group 2 — `x-vm0-test` and `x-vm0-test-endpoint-bypass`

→ `x-okou-test` and `x-okou-test-endpoint-bypass`.

Files: `crates/runner/mitm-addon/src/upstream_admission.py` plus its tests, `turbo/apps/api/src/signals/routes/test-endpoint-helpers.ts:2`, `turbo/packages/connectors/src/auth-providers/connectors/test-oauth/oauth.ts` and its test.

These cross from the addon into `turbo/`, but not into a production contract: `test-endpoint-helpers.ts` gates on `isPreviewRuntime()` — `deployEnv === "preview"` or `VERCEL_ENV === "preview"` — and on a bypass secret. Sender and receiver are both in this repository and are built from the same commit for any given preview.

**Verify that gating claim yourself before relying on it.** Read `test-endpoint-helpers.ts` and confirm the bypass cannot be reached in production. If it can, stop and report rather than proceeding — that would be a finding worth more than this rename.

### The prefix hazard, which is the real trap here

`x-vm0-test` is a **strict prefix** of `x-vm0-test-endpoint-bypass`. An unanchored find-and-replace will rewrite the longer name's prefix and produce `x-okou-test-endpoint-bypass` by accident in some places and a mangled hybrid in others — and because both names are being renamed anyway, the mistake can easily land looking correct.

Rename the longer name first, or anchor every replacement on the full header string. Then verify by counting each name independently rather than eyeballing the diff. `upstream_admission.py` uses both, so it is where this will bite.

## Explicitly excluded — `x-vm0-connector-intent`

Not in scope. It is set by the **published CLI** at `turbo/apps/cli/src/commands/mcp/client.ts:112` (`CONNECTOR_INTENT_HEADER = "X-VM0-Connector-Intent"`) and consumed by the addon at `crates/runner/mitm-addon/src/connector_intent.py`, which matches the exact literal `b"x-vm0-connector-intent"` and whose docstring is *"Capture connector intent once and always remove its private header."*

A one-step rename gives a new addon that no longer recognises the old name, so a request from an older CLI keeps the header — and it is **forwarded upstream to the third-party service instead of being stripped**. That is a header leak, not merely lost attribution. The strip-list entry at `custom-connector.service.ts:133` is defence in depth against the same thing.

It needs its own expand/contract issue. Do not touch it here, and do not touch that strip-list entry.

## Also do not touch

- `x-vm0-web-origin` — first-party www↔api, separate issue.
- `x-vm0-idempotency-key`, `x-vm0-client-request-id`, `"VM0"` — #33657.
- `restrictedVm0Models` — #33658.
- `hashtext('vm0')`, `PUBLIC_BRANDS`, the brand `z.enum`s, the 11 `vm0.*` localStorage keys, the 12 locale `repositoriesPlaceholder` strings, and live infrastructure names (`vm0-ai` org, `static.vm0.io`, `vm0-runner*`, `vm0-ns-*`, Axiom datasets).
- Do not weaken, delete, or invert any assertion. Do not add a test asserting an old header name is absent.

## Acceptance

1. `x-vm0-codex-model-catalog-prefetch` returns zero hits; the new name appears at every former call site.
2. `x-vm0-test-endpoint-bypass` returns zero hits and `x-okou-test-endpoint-bypass` appears exactly where it did.
3. `x-vm0-test` returns zero hits, and **no occurrence of `x-okou-test` is a truncated remnant of the longer name** — count both independently and state both counts in the PR body.
4. `x-vm0-connector-intent` is **unchanged**, in both `crates/` and `turbo/`, including the `custom-connector.service.ts` strip-list entry.
5. The preview-only gating of the bypass header is confirmed in the PR body, or a contradiction is reported.

lancy:
Closes #33659.

Renames the two groups of retired-brand `x-vm0-*` headers whose sender and receiver ship in the same build artifact, so no cross-version window exists.

## Group 1 — `x-vm0-codex-model-catalog-prefetch` → `x-okou-codex-model-catalog-prefetch`

8 files, zero references in `turbo/`. The sender is the prefetch `curl` script in `crates/runner/src/executor/codex_model_catalog_prefetch.rs`; the receiver is `crates/runner/mitm-addon/src/codex_model_catalog_cache.py`. `crates/runner/build.rs` `generate_addon_files()` embeds `mitm-addon/src/**` into the runner binary as `ADDON_FILES`, and `crates/runner/src/proxy/process.rs` extracts them at startup, so both sides are one artifact replaced together — the exemption stated in `docs/deployment-compatibility.md`.

Three casings existed and were each rewritten in place: `X-VM0-…`, `x-vm0-…`, and the deliberately mixed `x-VM0-…` used by the case-insensitivity tests (now `x-Okou-…`).

## Group 2 — `x-vm0-test-endpoint-bypass` and `x-vm0-test`

`x-vm0-test-endpoint-bypass` → `x-okou-test-endpoint-bypass` in `crates/runner/mitm-addon/src/upstream_admission.py` and its tests, `turbo/apps/api/src/signals/routes/test-endpoint-helpers.ts`, and `turbo/packages/connectors/src/auth-providers/connectors/test-oauth/oauth.ts` and its test.

The bare string `x-vm0-test` was **not** an HTTP header anywhere in the tree. Its only two retired-brand occurrences are inert, test-local strings, both renamed:

- `crates/runner/mitm-addon/tests/x_connector_usage/test_writes.py` — a synthetic *unsupported `Content-Encoding` value* in a `@pytest.mark.parametrize` list → `x-okou-test`.
- `turbo/apps/platform/src/views/okou-page/__tests__/directed-connect-page.test.tsx` — a `headerInjections` fixture name on a fake custom connector → `X-Okou-Test-Token`.

### Prefix hazard

`x-vm0-test` is a strict prefix of `x-vm0-test-endpoint-bypass`. Every replacement was anchored on a full header string and the longer name was rewritten first. Counts were taken independently rather than read off the diff:

| pattern | before | after |
| --- | --- | --- |
| `x-vm0-codex-model-catalog-prefetch` (any case) | 20 | **0** |
| `x-okou-codex-model-catalog-prefetch` (any case) | 0 | **20** |
| `x-vm0-test-endpoint-bypass` | 16 | **0** |
| `x-okou-test-endpoint-bypass` | 0 | **16** |
| `x-vm0-test` (any case, substring) | 18 | **0** |
| `x-okou-test` anchored bare, i.e. not followed by `-` or a word char | 0 | **1** |
| `x-okou-test-token` | 0 | **1** |

The single bare `x-okou-test` is the `test_writes.py` encoding value, not a truncated remnant: `1 + 16 + 1 = 18` accounts for every prior `x-vm0-test` substring hit exactly once.

`git diff --shortstat`: 17 files, 38 insertions, 38 deletions — a pure rename. No assertion was weakened, deleted, or inverted, and no test asserting the absence of an old name was added.

## Preview-only gating — confirmed, no contradiction

Verified on all three sides; the bypass is not reachable in production:

1. **Receiver** — `test-endpoint-helpers.ts` `isTestEndpointAllowed()` returns `true` for `ENV === "development"`, otherwise only inside the `isPreviewRuntime(deployEnv)` branch (`deployEnv === "preview" || VERCEL_ENV === "preview"`), which additionally requires a non-empty `VERCEL_AUTOMATION_BYPASS_SECRET` match. A production runtime is neither, so control falls through to the terminal `return false` and the header is never read.
2. **Sender** — `testOAuthPreviewBypassHeaders()` emits `{}` unless `VERCEL_AUTOMATION_BYPASS_SECRET` is set, and belongs to the preview-only `test-oauth` connector.
3. **Addon** — `_request_has_platform_test_endpoint_bypass()` applies only to the `/api/test/` path prefix and returns `false` when the secret env var is empty.

## `x-vm0-connector-intent` — untouched, on purpose

Deliberately out of scope: it is emitted by the **published CLI** and consumed by the addon against the exact literal `b"x-vm0-connector-intent"`. A one-step rename would leave a new addon unable to recognise an older CLI's header, so it would be forwarded upstream to the third-party service instead of stripped — a header leak. It needs its own expand/contract change.

Verified unchanged: 48 occurrences of `vm0-connector-intent` across 12 files before and after, and **0** occurrences of `okou-connector-intent`. The `custom-connector.service.ts` strip-list entry is untouched. `test_request_header_names.py` carries both header families; its diff shows only the prefetch lines changing while the adjacent `x-VM0-Connector-Intent` fixtures stay as-is.

Also untouched: `x-vm0-web-origin`, the #33657 and #33658 scopes, `hashtext('vm0')`, `PUBLIC_BRANDS` and the brand `z.enum`s, the `vm0.*` localStorage keys, the locale `repositoriesPlaceholder` strings, and all live infrastructure names.

## Verification

Locally: repo-wide anchored counts above, plus `python -m py_compile` on every changed Python file. The addon test dependencies (`mitmproxy`, `pytest`) are not installable in this sandbox — no `uv` — and a Rust debug build OOMs here, so the addon, runner, and `turbo/` suites are **not** claimed to have run locally; they are left to CI. Separately, `claimRunnerJob` returning `404 Job not found in queue` and the BDD failures that trail it reproduce on unmodified `main` in this same environment and are not caused by this change.


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: d646fc161faac31478b78c4bfd0e4421d799baf2
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `d646fc161faac31478b78c4bfd0e4421d799baf2`. Practice revision: `8451ce0e82e70cfeb16bb83441892c4022e38014` (`main`).

### Summary

Renames two retired-brand header groups whose sender and receiver ship inside the same build artifact: `x-vm0-codex-model-catalog-prefetch` → `x-okou-codex-model-catalog-prefetch`, and `x-vm0-test-endpoint-bypass` → `x-okou-test-endpoint-bypass`, plus two inert `x-vm0-test` test strings. 17 files, 38 insertions and 38 deletions, with no behavioral edit beyond the literals. `x-vm0-connector-intent` is deliberately excluded because it crosses the published-CLI boundary.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | Group 1 producer `crates/runner/src/executor/codex_model_catalog_prefetch.rs:45` and consumer `crates/runner/mitm-addon/src/codex_model_catalog_cache.py:124-125` renamed together; group 2 producer `test-oauth/oauth.ts:128`, API reader `test-endpoint-helpers.ts:2`, and addon reader `upstream_admission.py:24` renamed together. No remaining producer or consumer of either old name. |
| `REVIEW.md` | Confirm existing coverage still protects a refactor | Pass | Every previously asserting test was updated to the new literal in place; assertion count and shape are unchanged (38/38 line-for-line). |
| `REVIEW.md` | Do not weaken, delete, or invert assertions; no tombstone tests | Pass | No assertion removed, negated, or loosened. No test asserting the absence of an old header name was added, so §1 of `fallback.md` is respected. |
| `REVIEW.md` | Report concrete practice violations with rule and source, not inferred from search matches | Pass | No violations found; the two out-of-header-contract renames below are recorded as observations, not findings. |
| `docs/bad-smell.md` §4 Interface Changes | Document changed public contracts and assess consumers; independent deployments need the old/new combinations | Pass | Both contracts are intra-artifact (see the compatibility rows). The PR body documents the contract change, the consumers, and the excluded CLI-crossing header. |
| `docs/bad-smell.md` §6 Hardcoded URLs and Configuration | No invented fallback URLs or credentials; environment values stay in centralized config | Pass | Only literal header names changed. `VERCEL_AUTOMATION_BYPASS_SECRET` reads through `optionalEnv`/`env` and `os.environ.get` exactly as before. |
| `docs/bad-smell.md` §7 / `docs/fallback.md` §5 | Do not keep a reader for a retired producer; delete reader and producer together | Pass | The one-step rename removes both sides simultaneously; no dual-name reader was introduced, which is correct for a same-artifact contract. |
| `docs/fallback.md` §6-§9 | New fallbacks must be justified, bounded, and declared in a PR-body `Fallbacks` section | N/A | The diff introduces no fallback, no dual read/write, and no tolerance branch, so no `Fallbacks` section is required (§9 explicitly exempts PRs that add none). |
| `docs/deployment-compatibility.md` (runner-private contract, L396-401) | Requests and protocols private to one runner artifact may change atomically | Pass | `crates/runner/build.rs:190 generate_addon_files()` embeds `mitm-addon/src/**` into `ADDON_FILES`; `crates/runner/src/proxy/process.rs:270` extracts them at startup. The prefetch producer is Rust in the same binary, so producer and consumer are one artifact — the documented exemption, with zero `turbo/` references. |
| `docs/deployment-compatibility.md` (what requires compatibility) | Cross-surface request contracts need old/new coexistence coverage | Pass | The bypass header's production sender (`packages/connectors`) and reader (`apps/api`) are the same backend artifact and commit. Its third reader, the addon, activates only when `VERCEL_AUTOMATION_BYPASS_SECRET` is present; that secret appears only in `.github/workflows/turbo.yml` CI/preview plumbing and in no Rust or runtime production path, so the addon and API sides that must agree are always built from the same commit. No mixed-version window exists. |
| `docs/testing.md` (choose the boundary) | Tests exercise the production boundary for the surface | Pass | Addon tests still drive `mitm_addon.requestheaders`/`request` hooks with real `mitmproxy` flows; the Rust test still asserts the produced `curl` command; the connectors test still reads the outgoing token-request headers. No boundary moved. |
| `docs/testing.md` (coverage) | Cover old/new combinations for changed request protocols | Pass | No coexistence combination is reachable (see the compatibility rows), so the existing single-name coverage is complete. |
| `docs/testing/testing-external-behavior.md` (Platform / API) | Assert user- or endpoint-observable state, not internal implementation | Pass | The only Platform change is a `headerInjections` fixture name in `directed-connect-page.test.tsx:269`; it is input data for the rendered connector dialog and is not asserted on anywhere in that file. No API test was touched and no log, DB, or service assertion was added. |
| `docs/testing/mitm-addon-testing.md` (flow metadata key contract) | Use the registry constants; run the metadata-key linter when adding or renaming shared metadata keys | N/A | No metadata key changed. `_codex_model_catalog_prefetch_request`, `_codex_model_catalog_cache_state`, and `_codex_model_catalog_cache_telemetry` are untouched; only HTTP header literals moved. |
| `docs/testing/mitm-addon-testing.md` (running addon tests) | Addon behavior changes are verified with the locked `uv` pytest environment | N/A | Not a property of the diff: this review host has no `uv` and no `mitmproxy`/`pytest`, so the locked environment cannot be created. The PR body states this instead of claiming a local pass, and the addon CI job is the gate. See Testing. |
| `docs/testing/rust-testing.md` | Rust changes verified with `cargo test` on the affected crate | N/A | Not a property of the diff: a Rust debug build OOMs in this review host. `crates/runner/src/executor/tests/agent_run_tests/codex_model_catalog_prefetch.rs:900` pins the renamed literal against the producer in the same crate, so a mismatch fails that test in CI. See Testing. |
| `docs/testing/app-testing.md` | Page tests enter through the real Router and avoid static-configuration-only unit tests | Pass | No test was added or restructured; the single Platform edit is fixture data inside an existing `setupPage`-based page test. |

The two `N/A` rows for the addon and Rust test guides record verification this review host could not run. They describe an environment limit, not a property of the diff, and the corresponding entries in Testing state the gap explicitly.

### Findings

No P0 or P1 defects in the diff.

**Non-blocking observations**

1. `crates/runner/mitm-addon/tests/x_connector_usage/test_writes.py:672` and `turbo/apps/platform/src/views/okou-page/__tests__/directed-connect-page.test.tsx:269` are not HTTP header contracts. The first is a synthetic unsupported `Content-Encoding` value in a `parametrize` list; the second is a fake custom-connector `headerInjections` name. Both are inert test-local strings and both are renamed here so that a literal `x-vm0-test` grep returns zero. This is slightly wider than "the `x-vm0-test` header", and the PR body says so explicitly, so it is recorded rather than raised.
2. `test_request_header_names.py:66,123,146` keeps the deliberately mixed-case `x-Okou-…` spelling that the old `x-VM0-…` fixture used. `codex_model_catalog_cache.py:623` matches with `name.lower() != _RAW_PREFETCH_HEADER`, so the case-insensitivity property those tests exercise is preserved exactly.

**Exclusion verified.** `x-vm0-connector-intent` is unchanged: 48 occurrences across 12 files before and after, `okou-connector-intent` absent, and the `custom-connector.service.ts:133` strip-list entry untouched. `test_request_header_names.py` carries both families, and its diff changes only the prefetch tuples while the adjacent `b"x-VM0-Connector-Intent"` fixtures stay as-is — which is exactly where an unanchored replacement would have leaked. Leaving it alone is correct: the CLI at `mcp/client.ts:112` is independently published, and a one-step rename would make a new addon fail to strip an older CLI's header and forward it to the third-party upstream.

### Testing

Verification performed for this review, independent of the author's:

- Anchored occurrence counts on the merge base and on this HEAD confirm the acceptance criteria. `vm0-codex-model-catalog-prefetch` 20 → 0 with `okou-codex-model-catalog-prefetch` 0 → 20; `x-vm0-test-endpoint-bypass` 16 → 0 with `x-okou-test-endpoint-bypass` 0 → 16; `x-vm0-test` as a case-insensitive substring 18 → 0.
- The prefix hazard is clear. Counting the two new names independently gives 16 for the long name, exactly 1 for `x-okou-test` anchored as *not* followed by `-` or a word character, and 1 for `x-okou-test-token`. `16 + 1 + 1 = 18` accounts for every prior `x-vm0-test` substring hit exactly once, so no `x-okou-test` occurrence is a truncated remnant of the longer name.
- Preview-only gating of the bypass header is confirmed on all three sides, so acceptance criterion 5 holds and there is no contradiction to escalate. `isTestEndpointAllowed` (`test-endpoint-helpers.ts`) reads the header only inside the `isPreviewRuntime(deployEnv)` branch (`ENV === "preview" || VERCEL_ENV === "preview"`) and additionally requires a matching `VERCEL_AUTOMATION_BYPASS_SECRET`; a production runtime is neither `development` nor preview and falls through to the terminal `return false`. `testOAuthPreviewBypassHeaders()` emits `{}` without that secret. `_request_has_platform_test_endpoint_bypass()` applies only to the `/api/test/` prefix and returns `False` on an empty secret.
- `python -m py_compile` passes on all eleven changed Python files.

Gaps, stated as gaps: the addon pytest suite, `cargo test -p runner`, and the `turbo/` Vitest suites were **not** executed for this review. The addon's `uv`-locked environment cannot be created here and a Rust debug build OOMs in this sandbox, so those are delegated to CI, which is the right gate for a rename whose failure mode is a producer/consumer literal mismatch that every touched test would catch. No new coverage is warranted: the change adds no behavior, and adding a test that the old name is absent would be exactly the tombstone `docs/fallback.md` §1 prohibits.

The `claimRunnerJob` `404 Job not found in queue` failure and its trailing BDD failures reproduce on unmodified `main` in the same environment and are unrelated to this diff.

Issue #33656, opened 2026-09-12
epic(api): retire log-level classification and diagnostics-pinning tests across the API


lancy:
## Background

Ethan's 2026-09-12 review ("defensive code stacked layer on layer, each layer pinned by its own tests, so later fixes cannot move") was verified and fixed for the two features he named in #33589 (seven children, all merged and accepted). The follow-up question was whether the rest of `turbo/apps/api` has the same pattern. A sweep of main `44596a8d07` (grep plus sampled reads, recorded in the #33589 thread) says yes, across every author, in three forms.

### Finding 1 — tests pin diagnostics project-wide

`src/__tests__/mocks.ts:110-117` exposes the logger stub as `context.mocks.axiomLogging.{debug,info,warn,error}` and `:339-343` the telemetry SDK stub (`useRealTelemetry`, `sdkIngest`). 45 API test files reference the logger mock (134 `expect(...)` calls: 28 pin a message and fields, 62 pin silence) and 20 read telemetry payloads (46 `op_type` / `action_type` literals in `chat-events.bdd`, `run-lifecycle.bdd`, `webhooks-callbacks.bdd`; `auxiliaryResults` / `auxiliaryDiagnostics` helpers in `chat-callbacks.bdd` and `auxiliary-generation.test.ts`). #33610 wrote the rule into `docs/testing/testing-external-behavior.md` and restricted `lib/log` imports, but the `context.mocks` route is unguarded, so the rule is not enforced. Legitimate subjects are only the four logger/transport suites.

### Finding 2 — log-level-only classifiers outside the first epic

- `agent-webhook-firewall-auth.service.ts` (5,927 lines): four September fixes about log noise (#31128, #32808, #32806, #33110) stacked `shouldLogWarning`, `terminalCodexFailure`, `isExpectedOAuthRefreshFailure` (`connector-oauth-refresh-policy.ts`, self-described as telemetry-only) and `oauthRefreshFailureLogFields` around one `warn` (`:2570-2645`); six test files pin "token refresh failed".
- `webhooks-built-in-generations.ts:318-346, 395-690, 1472-1500` and `image-generation.service.ts:2055-2080, 2195-2215`: Fal and OpenAI failure classifiers whose `kind` / `stage` / `classificationSource` / `expected` dimensions exist for one log record; `image-io-generate.test.ts` pins them (35 references).
[...]

Issue #33647, opened 2026-09-12
test(api): gate logger and telemetry assertions in api tests behind a shrinking baseline


lancy:
Parent: #33656 (finding 1). Baseline: main `44596a8d07`.

## Background

`turbo/apps/api/src/__tests__/mocks.ts:110-117` exposes the logger stub as `context.mocks.axiomLogging.{debug,info,warn,error}`, and `:339-343` exposes the telemetry SDK stub as `context.mocks.axiom.{useRealTelemetry,sdkIngest}`. 45 API test files reference `mocks.axiomLogging.<level>` (134 `expect(...)` calls: 28 pin a message and fields, 62 pin silence, the rest filter `mock.calls`), and 20 files read telemetry through `sdkIngest`, `useRealTelemetry`, or an MSW handler on `api.axiom.co` (46 `op_type` / `action_type` literals in the three BDD suites). #33610 made the rule explicit in `docs/testing/testing-external-behavior.md` and restricted `lib/log` imports, but the mocks route is open, so the rule is not enforced.

Legitimate subjects, whose tests must keep the access: `src/lib/__tests__/log.test.ts`, `src/lib/__tests__/log-axiom-transport.test.ts`, `src/signals/external/__tests__/axiom.test.ts`, `src/__tests__/app-factory.test.ts`.

Precedent: the API-test `no-restricted-imports` block in `turbo/apps/api/eslint.config.mjs` (files `src/**/__tests__/**/*.ts`, `src/**/*.test.ts`, with `ignores`), and the "legacy baseline only shrinks" rule for `turbo/style-allowlist.json` in `CLAUDE.md`.

## Design

1. Add a config block for the same test globs with `no-restricted-syntax` selectors: `MemberExpression[property.name="axiomLogging"]`, `MemberExpression[property.name="sdkIngest"]`, `MemberExpression[property.name="useRealTelemetry"]`, each with the message "API tests must not observe the logger or telemetry; assert HTTP responses and effects. See docs/testing/testing-external-behavior.md". Exempt the four subject suites in `ignores` with one comment each.
2. Add `apiTestDiagnosticsBaseline`, an inline array of every file that violates the rule at merge time, used as the block's remaining `ignores`, with a comment: owned by #33656, may only shrink, deleted when empty. Generate the list from the lint output, do not hand-pick.
3. `docs/testing/testing-external-behavior.md` Lint section: name the rule and the baseline.
4. No test file changes in this PR.

[...]

PR #33665, opened 2026-09-12
refactor(api): drop log-only dimensions from the image generation failure classifiers

### LINKED ISSUE #33649: refactor(api): drop log-only dimensions from the image generation failure classifiers
Parent: #33656 (finding 2). Baseline: main `44596a8d07`.

## Background

Two image-generation failure classifiers exist to shape one log record each.

- Fal, `turbo/apps/api/src/signals/routes/webhooks-built-in-generations.ts`: `FalGenerationFailureKind` / `Stage` / `RetryPolicy` (`:318-331`), `FalGenerationFailure` with `kind`, `stage`, `retryPolicy`, `classificationSource`, `expected`, `providerHttpStatus`, `providerErrorType` (`:333-346`), rule table `:395-560`, `falGenerationFailure` `:603-690`. Downstream, only `error` (public code and message) and `retryPolicy` are read (`image-generation.service.ts:2206` and the generation record). `kind`, `stage`, `classificationSource` and `expected` feed the log at `:1472-1500`, which also carries constant fields (`billingDisposition: "not_charged"`, `artifactRecorded: false`, `usageRecorded: false`, `admissionStatus: "failed"`) and picks `info` versus `warn` by `expected`. Added by #31690.
- OpenAI, `image-generation.service.ts:2055-2080` `OpenAiImageFailure` (`failureKind`, `failureStage`, `retryPolicy`, `expected`) and the `debug` versus `error` split with the same constant fields at `:2195-2215`.

`image-io-generate.test.ts` pins both: `FAL_FAILURE_LOG_MESSAGE`, `OPENAI_FAILURE_LOG_MESSAGE`, `openAiFailureLogs()` filtering `mock.calls`, and `toStrictEqual` on log arrays (`:1149`, `:1302`, `:2206`, `:2556`; 35 logger references). `video-io-generate.test.ts` has one.

## Design

- Keep what behaviour needs: the provider-message → public `error` mapping and `retryPolicy`. Keep one boolean per rule that says whether the failure is a user-input outcome; rename `expected` to `userInput` if that reads better.
- Delete `kind`, `stage`, `classificationSource`, `providerErrorType` bookkeeping, the constant log fields, and the level split. Log one `warn("<provider> image generation failed", { generationId, type, providerStatus, publicErrorCode, retryPolicy })` only for non-user-input failures; nothing for user-input outcomes (the job row carries the public error code).
- Apply the same to the OpenAI classifier.
- Tests: delete the log helpers and every logger-mock assertion in `image-io-generate.test.ts` and `video-io-generate.test.ts` (and remove both from the #33647 baseline if it exists); keep and, where a case only asserted logs, add the HTTP observable: generation status, public error code, `retryPolicy` on the job read endpoint.

Rejected: deleting the rule table (the public error codes and retry policy are user-facing behaviour).

## Explicit non-goals

- No change to public error codes, messages, retry policies, billing disposition, or webhook acceptance behaviour; no contract change; no change to `webhooks-built-in-generations` providers other than Fal.

## Rollout

One PR. Ordinary release train.

## Acceptance criteria

- [ ] `grep -n "classificationSource\|FalGenerationFailureKind\|FalGenerationFailureStage\|failureKind\|failureStage\|billingDisposition\|artifactRecorded\|usageRecorded\|admissionStatus" turbo/apps/api/src/signals/routes/webhooks-built-in-generations.ts turbo/apps/api/src/signals/services/image-generation.service.ts` is empty.
- [ ] Each classifier writes at most one `warn`, never `info` / `debug` / `error`, for a failed generation.
- [ ] `image-io-generate.test.ts` and `video-io-generate.test.ts` contain no `mocks.axiomLogging` reference; every failure case asserts the public error code or retry policy through the generation read endpoint.
- [ ] API lint, types, knip pass; merged through the protected queue via same-thread `/pr-auto`.

lancy:
## Summary

Closes #33649 (child G3 of #33656).

Two image generation failure classifiers carried a taxonomy whose only consumer was a single log record. This drops that taxonomy and keeps the behaviour callers depend on.

Kept (unchanged behaviour): the provider-message → public `error` mapping in the Fal rule table and the OpenAI code allowlist, every public error code and message, `retryPolicy`, billing disposition, webhook acceptance, and realtime events.

Removed:

- `FalGenerationFailureKind`, `FalGenerationFailureStage`, `classificationSource`, and the `expected` level split.
- `allowedFalProviderErrorType` / `falProviderErrorTypeForLog` — a sanitizing allowlist whose only output was the `providerErrorType` log field. The per-rule `providerErrorType` used to *match* a Fal diagnostic stays; it is behaviour.
- `falProviderHttpStatus` and `FalWebhookPayload.providerHttpStatus` — a regex that parsed `"Unexpected status code: 422"` out of the webhook envelope solely to produce a log field, which the issue's prescribed warn shape does not carry.
- OpenAI `failureKind`, `failureStage`, `providerErrorType`, `providerErrorCode`.
- The constant log fields `billingDisposition: "not_charged"`, `artifactRecorded: false`, `usageRecorded: false`, `admissionStatus: "failed"`.

`expected` is renamed `userInput` — the one boolean per rule that says whether the caller resolves the failure by changing the request. Each classifier now writes at most one `warn` per failed generation, and only when `userInput` is false; a caller-caused failure logs nothing, because the job row already carries its public error code. Neither classifier emits `info`, `debug`, or `error` any more.

Log shape, both providers: `{ provider, generationId?, type?, model?, providerStatus, publicErrorCode, retryPolicy }`.

One note for the reviewer: after this change `retryPolicy` is `"retry_once"` on every path that still logs, because every other policy belongs to a `userInput` rule that is now silent. The issue's design and acceptance criteria explicitly keep `retryPolicy` on the classifier, so it stays; it is the per-rule statement of retry semantics for a future consumer, not a log-only dimension.

### Tests

`image-io-generate.test.ts` and `video-io-generate.test.ts` no longer reference `context.mocks.axiomLogging`; `FAL_FAILURE_LOG_MESSAGE`, `OPENAI_FAILURE_LOG_MESSAGE`, and `openAiFailureLogs()` are gone. No `vi.spyOn` and no telemetry payload parsing replaced them. Every failure case still asserts through production surfaces — `GET /api/built-in-generations/:id` status and public error code, the realtime failure event, credits, `GET /api/usage/record`, and stored artifacts:

- The duplicate-webhook case proved single terminal failure through a log-call count; it now asserts exactly one realtime failure event for that generation plus a stable failed read after the repeat.
- `retains safe Fal diagnostics for $caseName…` became `falls back to the generic Fal image failure for $caseName`; the near-match and unknown payloads still prove they are not misclassified, via the public `GENERATION_FAILED` code and the unchanged private-value leak checks.
- The video suite's equivalent case was renamed and keeps its status, realtime, credits, and leak assertions.

`turbo/apps/api/eslint.config.mjs` has no `apiTestDiagnosticsBaseline` on current `main` (#33647 is not open yet), so there was nothing to remove from it.

## Verification

Run from `turbo/apps/api` on a local PostgreSQL 18 instance with migrations applied:

- `pnpm run lint` — pass.
- `pnpm run check-types` — pass.
- `npx knip --workspace apps/api` — clean.
- `npx prettier --check` on the four changed files — pass.
- `npx vitest run src/signals/routes/__tests__/video-io-generate.test.ts` — 46/46 pass.
- `npx vitest run src/signals/routes/__tests__/webhooks-built-in-generations.test.ts src/signals/routes/__tests__/built-in-generation.test.ts` — 8/8 pass.
- `npx vitest run src/signals/routes/__tests__/image-io-generate.test.ts` — 74 pass, 1 fail: `uses allowance for a legacy runless generation under shared debt`. That test fails identically on unmodified `origin/main` in this sandbox (verified by stashing the branch and re-running it), so it is a pre-existing environment gap, not a regression from this change. It does not touch the classifiers.

Acceptance grep from the issue is empty:

```
grep -n "classificationSource\|FalGenerationFailureKind\|FalGenerationFailureStage\|failureKind\|failureStage\|billingDisposition\|artifactRecorded\|usageRecorded\|admissionStatus" \
  turbo/apps/api/src/signals/routes/webhooks-built-in-generations.ts \
  turbo/apps/api/src/signals/services/image-generation.service.ts
```

### DIFF
--- a/turbo/apps/api/src/signals/routes/__tests__/image-io-generate.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/image-io-generate.test.ts
@@ -88,9 +88,6 @@ const FAL_INVALID_REQUEST_MESSAGE =
   "Could not generate images with the given prompts and images. Please try again with different inputs.";
 const FAL_INVALID_ASPECT_RATIO_MESSAGE =
   "Input should be 'auto', '21:9', '16:9', '3:2', '4:3', '5:4', '1:1', '4:5', '3:4', '2:3', '9:16', '4:1', '1:4', '8:1' or '1:8'";
-const FAL_FAILURE_LOG_MESSAGE =
-  "Fal built-in generation webhook reported failed generation";
-const OPENAI_FAILURE_LOG_MESSAGE = "OpenAI image generation request failed";
 const OPENAI_PRIVATE_PROVIDER_REQUEST_ID =
   "req_private0openai0request0identifier";
 const OPENAI_PRIVATE_PROVIDER_MESSAGE = `Invalid image file or mode for image 1, please check your image file. If you believe this is an error, contact us at help.openai.com and include the request ID ${OPENAI_PRIVATE_PROVIDER_REQUEST_ID}.`;
@@ -351,14 +348,6 @@ function readAcceptedGenerationId(
     },
   });
   return body.generationId;
-}
-
-function openAiFailureLogs(
-  loggingMock: typeof context.mocks.axiomLogging.info,
-): unknown[][] {
-  return loggingMock.mock.calls.filter(([message]) => {
-    return message === OPENAI_FAILURE_LOG_MESSAGE;
-  });
 }
 
 function readGenerationResult(body: unknown): unknown {
@@ -1007,9 +996,6 @@ describe("POST /api/image-io/generate", () => {
       editing: true,
       providerErrorCode: "invalid_image_file",
       moderationDetails: undefined,
-      failureKind: "input_media_invalid",
-      failureStage: "input",
-      retryPolicy: "after_input_change",
       publicError: {
         message: "An input image could not be read by the generation provider.",
         code: "GENERATION_INPUT_MEDIA_INVALID",
@@ -1021,9 +1007,6 @@ describe("POST /api/image-io/generate", () => {
       editing: false,
       providerErrorCode: "invalid_image_file",
       moderationDetails: undefined,
-      failureKind: "input_media_invalid",
-      failureStage: "input",
-      retryPolicy: "after_input_change",
       publicError: {
         message: "An input image could not be read by the generation provider.",
         code: "GENERATION_INPUT_MEDIA_INVALID",
@@ -1038,9 +1021,6 @@ describe("POST /api/image-io/generate", () => {
         moderation_stage: "input",
         categories: ["violence"],
       },
-      failureKind: "input_safety_rejected",
-      failureStage: "input",
-      retryPolicy: "after_input_change",
       publicError: {
         message:
           "The prompt or reference image was blocked by the safety filter.",
@@ -1053,9 +1033,6 @@ describe("POST /api/image-io/generate", () => {
       editing: false,
       providerErrorCode: "moderation_blocked",
       moderationDetails: { moderation_stage: "output", categories: ["sexual"] },
-      failureKind: "output_safety_blocked",
-      failureStage: "output",
-      retryPolicy: "manual_once",
       publicError: {
         message: "The generated image was blocked by the safety filter.",
         code: "GENERATION_OUTPUT_SAFETY_BLOCKED",
@@ -1067,9 +1044,6 @@ describe("POST /api/image-io/generate", () => {
       editing: false,
       providerErrorCode: "moderation_blocked",
       moderationDetails: {},
-      failureKind: "input_safety_rejected",
-      failureStage: "input",
-      retryPolicy: "after_input_change",
       publicError: {
         message:
           "The prompt or reference image was blocked by the safety filter.",
@@ -1083,9 +1057,6 @@ describe("POST /api/image-io/generate", () => {
       editing,
       providerErrorCode,
       moderationDetails,
-      failureKind,
-      failureStage,
-      retryPolicy,
       publicError,
     }) => {
       const fixture = await seedImageFixture({ credits: 1000 });
@@ -1145,45 +1116,14 @@ describe("POST /api/image-io/generate", () => {
       });
       await expect(orgCredits(fixture)).resolves.toBe(1000);
 
-      const debugLogs = openAiFailureLogs(context.mocks.axiomLogging.debug);
-      expect(debugLogs).toStrictEqual([
-        [
-          OPENAI_FAILURE_LOG_MESSAGE,
-          expect.objectContaining({
-            context: "ImageGeneration",
-            provider: "openai",
-            model,
-            providerStatus: 400,
-            providerErrorType: "image_generation_user_error",
-            providerErrorCode,
-            failureKind,
-            failureStage,
-            publicErrorCode: publicError.code,
-            retryPolicy,
-            billingDisposition: "not_charged",
-            expected: true,
-          }),
-        ],
-      ]);
-      for (const silentMock of [
-        context.mocks.axiomLogging.error,
-        context.mocks.axiomLogging.warn,
-        context.mocks.axiomLogging.info,
-      ]) {
-        expect(openAiFailureLogs(silentMock)).toHaveLength(0);
-      }
-
-      const publicAndLogSurfaces = JSON.stringify({
-        status: statusBody,
-        debugLogs,
-      });
+      const publicSurfaces = JSON.stringify(statusBody);
       for (const privateValue of [
         OPENAI_PRIVATE_PROVIDER_MESSAGE,
         OPENAI_PRIVATE_PROVIDER_REQUEST_ID,
         "a product illustration",
         MOCKUP_IMAGE_URL,
       ]) {
-        expect(publicAndLogSurfaces).not.toContain(privateValue);
+        expect(publicSurfaces).not.toContain(privateValue);
       }
     },
   );
@@ -1297,33 +1237,7 @@ describe("POST /api/image-io/generate", () => {
         },
       });
       await expect(orgCredits(fixture)).resolves.toBe(1000);
-
-      const errorLogs = openAiFailureLogs(context.mocks.axiomLogging.error);
-      expect(errorLogs).toStrictEqual([
-        [
-          OPENAI_FAILURE_LOG_MESSAGE,
-          expect.objectContaining({
-            context: "ImageGeneration",
-            provider: "openai",
-            providerStatus: upstreamStatus,
-            providerErrorType: "unknown",
-            providerErrorCode: "unknown",
-            failureKind: "unknown",
-            failureStage: "provider",
-            publicErrorCode: "OPENAI_IMAGE_REQUEST_FAILED",
-            retryPolicy: "retry_once",
-            billingDisposition: "not_charged",
-            expected: false,
-          }),
-        ],
-      ]);
-      for (const silentMock of [
-        context.mocks.axiomLogging.debug,
-        context.mocks.axiomLogging.info,
-      ]) {
-        expect(openAiFailureLogs(silentMock)).toHaveLength(0);
-      }
-      expect(JSON.stringify({ status: statusBody, errorLogs })).not.toContain(
+      expect(JSON.stringify(statusBody)).not.toContain(
         OPENAI_PRIVATE_PROVIDER_MESSAGE,
       );
     },
@@ -2085,11 +1999,9 @@ describe("POST /api/image-io/generate", () => {
     {
       detailShape: "a string detail",
       detail: FAL_OUTPUT_SAFETY_FILTER_MESSAGE,
-      providerErrorType: "unknown",
     },
     {
       detailShape: "Pydantic detail entries",
-      providerErrorType: "content_policy_violation",
       detail: [
         { type: "file_download_error", msg: "Unrelated provider diagnostic" },
         {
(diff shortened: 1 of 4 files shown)


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: 050acfa1905d269402488c81e5bd265a23ee1189
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `050acfa1905d269402488c81e5bd265a23ee1189`. Practice documents read from `main`: `REVIEW.md`, `docs/docs.md`, `docs/bad-smell.md`, `docs/fallback.md`, `docs/testing.md`, `docs/testing/testing-external-behavior.md`, `docs/testing/api-testing.md`.

### Summary

Strips the Fal and OpenAI image-generation failure classifiers down to the two things behaviour depends on — the provider-message → public `error` mapping and `retryPolicy` — plus one `userInput` boolean per rule, and deletes the `kind` / `stage` / `classificationSource` / `providerErrorType` / `providerHttpStatus` dimensions, the four constant log fields, and the `expected`-driven level split that existed only to shape one log record. Each classifier now writes at most one `warn` per failed generation and only for a provider-side failure. The two affected suites drop every `context.mocks.axiomLogging` assertion and assert the generation read endpoint, the realtime failure event, credits, usage, and stored artifacts instead.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behaviour through its real entry point and consumers | Pass | `falGenerationFailure` and `classifyOpenAiImageFailure` keep every classification branch and returned `error` / `response`; the only consumers of the removed fields were the two log statements. `grep` over `turbo/apps/api/src` and `turbo/packages` finds no other reader of `retryPolicy`, `classificationSource`, `failureKind`, or `failureStage`. |
| `REVIEW.md` | API tests must not assert log output (outside the logger suite and one redaction check) | Pass | `image-io-generate.test.ts` and `video-io-generate.test.ts` now contain zero `context.mocks.axiomLogging` references; `FAL_FAILURE_LOG_MESSAGE`, `OPENAI_FAILURE_LOG_MESSAGE`, and `openAiFailureLogs()` are deleted. No `vi.spyOn` or telemetry-payload parsing replaced them. |
| `REVIEW.md` | A PR answering a log-noise issue must justify the record's existence rather than add another classifier | Pass | The change removes a classifier rather than adding one; the surviving `warn` is scoped to failures an operator can act on, and caller-caused failures now log nothing because the job row carries the public error code. |
| `REVIEW.md` | Changed / removed flows keep regression coverage for the affected path and a rewritten expectation must prove an intentional replacement | Pass | Every Fal rule-table case, every OpenAI code case, and each fallback payload shape is still exercised end-to-end; the assertions moved from log fields to `GET /api/built-in-generations/:id` status + public error code, the Ably failure event, `orgCredits`, `GET /api/usage/record`, and `s3.send`. The duplicate-webhook idempotency case, previously proven by a log-call count, now asserts exactly one realtime failure event for that generation plus a stable failed read after the repeat. |
| `REVIEW.md` | Persisted data, contracts, and cross-version compatibility | N/A | No schema, ts-rest contract, request/response shape, queue payload, or persisted column changes; `failBuiltInGenerationJob$` still receives the same `{ message, code }`. Only log fields and internal classifier types changed. |
| `REVIEW.md` | New fallbacks must be declared in a `Fallbacks` PR-body section | N/A | No fallback, default, tolerance branch, or rollout switch is introduced; the PR body correctly omits the section. |
| `docs/bad-smell.md` §3 | "Do not add code whose only output is a log level"; a fix for a noisy record first considers deleting the record | Pass | This is the rule being applied. `FalGenerationFailureKind`, `FalGenerationFailureStage`, `classificationSource`, `allowedFalProviderErrorType`, `falProviderErrorTypeForLog`, `falProviderHttpStatus`, and the OpenAI `failureKind` / `failureStage` / `providerErrorType` / `providerErrorCode` fields all terminated in a log record and are gone. Error handling, admission completion, and job transition are untouched. |
| `docs/bad-smell.md` §1 | No `any`; narrow untrusted input with validation | Pass | No `any` added. `isRecord`, `falFailureDetailDiagnostics`, and `openAiImageErrorSchema` still own the untrusted-boundary narrowing. |
| `docs/bad-smell.md` §2 | No lint/type suppression comments | Pass | No `eslint-disable`, `oxlint-disable`, `@ts-*`, or `prettier-ignore` in the diff. |
| `docs/bad-smell.md` §4 | Document changed public contracts and assess consumers | Pass | `FalGenerationFailure`, `FalStructuredFailureRule`, `FalWebhookPayload`, and `OpenAiImageFailure` are all module-private; nothing is exported or imported across a package boundary. |
| `docs/bad-smell.md` §5 | Static imports in production code | Pass | No import changes. |
| `docs/bad-smell.md` §7 | Default is no fallback for a state the contract rules out | Pass | Removing `providerHttpStatus` also removes the `FAL_STATUS_CODE_ERROR` regex that reconstructed a status from a provider error string purely to log it; nothing consumes it. No `??` chain was added. |
| `docs/fallback.md` §1 | Delete the old test with the old code; do not leave tombstones | Pass | The log assertions are deleted with the dimensions they pinned. `retains safe Fal diagnostics for $caseName…` is rewritten as `falls back to the generic Fal image failure for $caseName` — the near-match and unknown payloads still prove a positive behaviour (not misclassified, `GENERATION_FAILED`), not the absence of code. |
| `docs/fallback.md` §5 | Delete the reader when the producer is retired | Pass | `falProviderErrorTypeForLog` / `allowedFalProviderErrorType` and `falProviderHttpStatus` are removed together with their sole consumer, including the now-unused `FalWebhookPayload.providerHttpStatus` field. |
| `docs/testing.md` | Log records are incidental implementation in API tests; assert the production contract | Pass | Both suites now assert only HTTP responses and externally observable effects. |
| `docs/testing.md` | Avoid duplicate cases that pin incidental implementation | Pass | The `it.each` columns that existed only to feed log assertions (`failureKind`, `failureStage`, `retryPolicy`, `expected`, `providerErrorType`, `providerHttpStatus`) are removed; the behavioural columns (`providerErrorType` for diagnostic matching, `providerMessage`, `location`, `publicError`) remain. |
| `docs/testing/testing-external-behavior.md` | The API endpoint is the only trusted boundary; no logger, DB, or service assertions | Pass | Setup is `POST /api/image-io/generate` / `POST /api/video-io/generate` plus the real Fal webhook endpoint; verification is `GET /api/built-in-generations/:id` and `GET /api/usage/record`. No DB reads/writes, no service calls, no logger assertions were added. |
| `docs/testing/api-testing.md` | Route tests exercise the real app; mock only external services through `context.mocks` | Pass | Unchanged harness. The new `failureEvents` assertion reads `context.mocks.ably.publish`, the centralized mock for an external realtime service at its boundary, which the same file already used via `toHaveBeenCalledWith`. |

### Findings

None blocking.

Non-blocking observations:

1. `turbo/apps/api/src/lib/__tests__/log-axiom-transport.test.ts:10,50-80` builds its NDJSON fixture from the old Fal field set (`failureKind`, `failureStage`, `classificationSource`, `billingDisposition`, `expected`) and the old `info`/`warn` split. It compiles and passes because it constructs the fields as literals rather than importing the classifier, and it is one of the logger/transport subject suites this epic exempts — its subject is the transport, not the classifier, and the payload is arbitrary. It is now a stale sample of a record production no longer emits; trimming it belongs with the remaining epic children (#33656), not here.
2. `webhooks-built-in-generations.ts:1363` and `image-generation.service.ts:2181` still carry `retryPolicy`, but every rule that survives the `!userInput` guard is `retry_once`, so the field is constant on the paths that log. The issue's design and acceptance criteria explicitly keep `retryPolicy` on both classifiers as the per-rule statement of retry semantics, and the PR body declares this; keeping it is the right call here, and the point is only worth revisiting if a future reader wants that field to earn its place.
3. `OpenAiImageFailure.retryPolicy` is typed `string` rather than a union, unchanged from `main`. Not introduced by this PR.
4. The OpenAI provider-fault record drops from `error` to `warn`. No alerting rule, dashboard, or doc in the repository keys on that message or level (`grep` over `*.ts`, `*.md`, `*.yml`, `*.json`), so nothing in-tree breaks; anyone holding an external Axiom monitor on `level:error` for `"OpenAI image generation request failed"` should retune it to `warn`.

### Testing

Coverage is preserved where it protects behaviour and removed only where it pinned diagnostics.

- Fal: all eleven structured rules, both output-safety detail shapes, and all nine unclassified/near-match payloads still assert the public error code and message through `GET /api/built-in-generations/:id` and the realtime failure event, plus `orgCredits`, `GET /api/usage/record`, `s3.send`, and the per-run admission-slot release.
- The private-value leak checks survive on every case; they now serialize the status body and realtime calls instead of the status body and log calls, which is the correct external surface and still covers the prompts, reference URLs, provider request IDs, provider messages, and raw status strings.
- OpenAI: five caller-correctable codes and six provider-fault cases still assert the public error, the unchanged credit balance, and non-leakage of the provider message and request ID.
- Video: the late-failure-after-completion case keeps its `status: "completed"` read assertion, which is the real proof that a post-terminal webhook does not re-fail the job; the removed log-silence loop added nothing to it.

Verification reported in the PR body and reproducible from `turbo/apps/api`: `pnpm run lint`, `pnpm run check-types`, `npx knip --workspace apps/api`, and `prettier --check` all pass; `video-io-generate.test.ts` 46/46 and `webhooks-built-in-generations.test.ts` + `built-in-generation.test.ts` 8/8 pass. `image-io-generate.test.ts` is 74 pass / 1 fail on `uses allowance for a legacy runless generation under shared debt`, which reproduces identically on unmodified `origin/main` in the same sandbox and does not touch the classifiers — a pre-existing local environment gap, left to PR CI to confirm. The issue's acceptance grep over both production files returns nothing.

No gap requires an additional test: the surviving `warn`-or-silence decision is a log outcome, and per `docs/testing/testing-external-behavior.md` it must not be pinned by a test.

PR #33670, opened 2026-09-12
refactor(api): rename restricted vm0 models to restricted built-in models

lancy:
Step 1 (expand) of #33658. Step 2 (contract) is deliberately **not** in this PR.

## The seam

The rename was already done on the storage side — Postgres column
`restricted_built_in_models`, the Drizzle field, the tier-values table, and
`org-plan-entitlements.service.ts`. It stopped at
`org-plan-entitlement-read.service.ts`, which destructured the canonical name
and re-emitted it as `restrictedVm0Models`. From that line outward — the billing
status service, model policy, model selection, run admission, the
`/api/billing/status` contract, both platform capability modules, and their
tests — everything still carried the retired brand.

This PR renames outward from that seam and keeps the wire compatible.

## What changed

- **Contract.** `billing.ts` gains `restrictedBuiltInModels`. `restrictedVm0Models`
  stays, documented as the retired alias.
- **API.** Every internal identifier is `restrictedBuiltInModels`. The response
  builder emits both names from the single `args.restrictedBuiltInModels` source,
  so they cannot diverge.
- **Platform.** `OrgPlanCapabilities.restrictedBuiltInModels` and the
  `LEGACY_TIER_CAPABILITIES` table are renamed; the read order is
  `billing.restrictedBuiltInModels ?? billing.restrictedVm0Models ?? fallback`.
- **`run-admission.service.ts`** — identifier only. Admission semantics are
  untouched.
- Postgres column, Drizzle schema, migrations, `hashtext('vm0')`, `PUBLIC_BRANDS`,
  the brand `z.enum`s, the `vm0.*` localStorage keys, and the locale
  `repositoriesPlaceholder` strings are all untouched.

## Why two releases

API and app ship as separate releases (`api-v1.587.0` / `app-v0.890.0`). Renaming
the response field in one step would make the platform read `undefined` and
**silently** fall back to `LEGACY_TIER_CAPABILITIES` — silently, because the field
is already `.optional()`, so nothing errors and nothing logs.

The blast radius, stated in both directions: that table is tier-derived, so an org
whose stored entitlement matches its tier default would land on the right value
anyway. The exposure is an org whose stored entitlement **differs** from the tier
default — which is precisely why the API sends the field at all. Those orgs would
see wrong model affordances for the length of the skew window.

This is **not** an entitlement bypass. Server-side admission reads its own
capabilities (`run-admission.service.ts:259` passes the server's own read into
`getRunModelAccess`), independent of what the browser believes. The skew-window
outcome is a wrong UI followed by a server rejection — degraded UX, not
unauthorized access.

## Fallbacks

- `billing-status.service.ts:BillingStatusResponse.restrictedVm0Models` — emits
  the retired brand alias alongside `restrictedBuiltInModels`, same value, same
  source. Surface: old web/app -> API. Gate: the replacement app is live and the
  client-version floor excludes the old build. Remove in #33658 step 2.
- `org-plan-capabilities.ts:orgPlanCapabilitiesFromBilling` — reads
  `billing.restrictedVm0Models` when `restrictedBuiltInModels` is absent. Surface:
  new web/app -> old API. Gate: no API from before this release is serving or
  retained for rollback. Remove in #33658 step 2.
- `e2e/playwright/lib/runner-onboarding.ts:readRunnerPaidEntitlement` — same
  surface and gate as the platform read.

## The two fixture mapping lines

`test-fixtures/org-plan-entitlement.ts` converted between the names in **both**
directions. Both bridges are removed, not moved:

- **`:67` — `restrictedBuiltInModels: values.restrictedVm0Models`** reconciled the
  fixture's *write* option name (old brand, test-facing) with the Drizzle insert
  field (already renamed). It existed only because the fixture's public option had
  not been renamed with the column. The option is now
  `restrictedBuiltInModels` and the line is an identity assignment.
- **`:199` — `restrictedVm0Models: restrictedBuiltInModels`** was the *read* mirror
  of the same thing: it converted the storage field back to the old name in
  `OrgPlanEntitlementFixtureState` so old-brand assertions kept working — the same
  shape as the seam in the read service. The state field is now
  `restrictedBuiltInModels`.

Both destructures are **kept** (not replaced by a spread) because they carry the
`!== null` narrowing into a field declared as a plain `boolean`; a spread would
lose it. Comments now say so.

No test was passing only because of these bridges: the fixture's parameter and
state types are explicit, so TypeScript excess-property checking means every one
of the ~50 call sites had to be renamed, and the API type check passes.

## `null` handling

The guard and its message are preserved verbatim in both
`org-plan-entitlement-read.service.ts:88-90` and the fixture's read path:
`Unexpected NULL restricted_built_in_models for org plan entitlement ${orgId}`.
`null` is not folded into `false`.

## Tests

No assertion was weakened, deleted, or inverted, and no test asserts the absence
of the old field.

- `billing-status.test.ts` — both wire assertions now cover **both** names. The
  `false` case also pins `typeof === "boolean"`: both names are `.optional()` on
  the contract, so `toBeFalsy()` alone would pass on an omitted field too. That
  was the pre-existing weakness described as trap 3 in the issue.
- **New** page test in `org-providers-tab.test.tsx` covers the alias fallback
  through the user interface: billing status is served as a pre-rename API would
  send it, with only `restrictedVm0Models` set, and the test asserts the `Pro`
  restriction affordance the user actually sees. Verified to fail when the alias
  is dropped from the read chain, so it is not a vacuous assertion. It is a page
  test rather than a signal-level unit test because
  `docs/testing/testing-external-behavior.md` makes the page the Platform
  boundary and `docs/testing/app-testing.md` forbids a helper-only unit test when
  a rendered page can cover the behavior.
- **New** case in `runner-onboarding.test.ts` covers the same alias fallback in the
  e2e helper. The existing case now stubs the current field, so it proves the new
  name is read.
- `icon-mono-fixtures.json` carries both fields, matching the real step-1 wire.

### Verification

Run in this sandbox:

- `check-types` and `lint` pass for `api`, `@okouai/app`, and `@okouai/api-contracts`.
- `knip` reports no unused or unlisted findings.
- API `billing-status.test.ts`: 33/33 pass.
- API `model-policies`, `chat-threads-model-selection`, `cron-billing-entitlements`: pass.
- Platform `org-plan-capabilities.test.ts`: 3/3 pass.

**Not passing locally, and not claimed as passing.** Two suites fail in this
sandbox for environment reasons, verified by re-running the identical selection on
unmodified `main` in the same environment:

- `retired-run-admission.test.ts` and `cron-connector-catalog.test.ts` fail on
  `claimRunnerJob` returning `404 Job not found in queue`. Branch: 5 failures.
  Unmodified `main`: 6 failures — a strict superset of the branch's set.
- The platform `okou-page` view suites (`org-providers-tab`,
  `personal-providers-tab`, `personal-usage-tab`, `chat-run-models`) fail on
  timeouts with a set that varies run to run. Branch: 7 then 10 failures.
  Unmodified `main`: 13, with an overlapping but non-identical set of names.

Neither set is caused by this change. Both are delegated to CI.

## Acceptance

1. ✅ `/api/billing/status` returns both fields from one source.
2. ✅ Platform prefers `restrictedBuiltInModels`, then `restrictedVm0Models`, then the tier table.
3. ✅ Every internal identifier renamed; only the outbound wire field is duplicated.
4. ✅ `null` guard and message survive.
5. ✅ Both fixture lines explained above.
6. ✅ No assertion weakened; no absence-asserting test.

Implements step 1 of #33658 only. The issue stays open for step 2, which this PR does not create.


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: d748ef708113ebfdcbd4931d5bfc270dd60beddd
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `d748ef708113ebfdcbd4931d5bfc270dd60beddd`
Practice revision: `7056f27d56ec92c5056702f866b900a97d975dc1`

### Summary

Step 1 (expand) of #33658. The rename had already landed on the storage side and
was reversed at `org-plan-entitlement-read.service.ts`, which re-manufactured
`restrictedVm0Models` and sent it outward to `/api/billing/status`. This PR
renames every internal identifier in the API and the platform, emits both wire
names from one source, and makes the platform read
`restrictedBuiltInModels ?? restrictedVm0Models ?? legacy tier table`.

Practice documents loaded for the changed surfaces: `REVIEW.md`,
`docs/bad-smell.md`, `docs/testing.md`, `docs/testing/testing-external-behavior.md`,
`docs/testing/app-testing.md`, `docs/fallback.md`, and
`docs/deployment-compatibility.md`. The database, feature-switch, event-sourcing,
styles, and ccstate references were not loaded: there is no schema, migration,
switch, persisted-event, styling, or signal-ownership change — the only touched
signals file is edited inside a pure function, not in its `computed`.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | The seam at `org-plan-entitlement-read.service.ts:94` is traced outward through `billing-status.service.ts`, `model-policy.service.ts`, `model-selection.service.ts`, `run-admission.service.ts`, the contract, and both platform capability modules. |
| `REVIEW.md` | Confirm existing coverage still protects a refactor | Pass | ~50 renamed call sites stay behavioral; `billing-status.test.ts` (33/33) and the model-policy/model-selection/billing-entitlement suites still exercise the renamed paths. |
| `REVIEW.md` | Enforcement semantics unchanged where only identifiers move | Pass | `run-admission.service.ts` changes are field names only; `getRunModelAccess` call at `:259` and every denial branch are untouched. |
| `REVIEW.md` | Interface changes assessed against independently deployed consumers | Pass | `billing.ts` is the app-facing contract; both names ship, so no consumer combination loses the field. |
| `REVIEW.md` | Every new fallback declared in a `Fallbacks` PR-body section | Pass | All three are declared with surface, gate, and removal condition. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | No `any` or suppression comment in the diff; `api`, `@okouai/app`, and `@okouai/api-contracts` type-check and lint clean. |
| `docs/bad-smell.md` | No fabricated default that hides a broken invariant | Pass | The `null` guard at `org-plan-entitlement-read.service.ts:88-90` still throws `Unexpected NULL restricted_built_in_models ...`; `null` is not folded into `false`. |
| `docs/bad-smell.md` | Document changed public contracts and assess consumers | Pass | The contract field carries a comment naming the alias, its reason, and its step-2 removal. |
| `docs/fallback.md` §6 | A fallback is justified only as bounded cross-version rollout compatibility | Pass | All three protect a genuinely reachable mixed-fleet state; API and app ship as separate releases. |
| `docs/fallback.md` §7-8 | Each rollout fallback names its surface, gate, removal condition, and follow-up | Pass | In-code comments name `old web/app -> API` (client-version floor) and `new web/app -> old API` (no such API serving or retained for rollback), both pointing at #33658 step 2. |
| `docs/fallback.md` §1 | No negative test pinning removed behavior as still absent | Pass | No test asserts the old field is absent; the alias is asserted present and equal, which is current product behavior, not a tombstone. |
| `docs/fallback.md` §3 | No `??` chain on a state the owning contract rules out | Pass | Both names are genuinely `.optional()` on `billingStatusResponseSchema`, so `undefined` is reachable from a differently-versioned peer. |
| `docs/deployment-compatibility.md` | Use the documented exposure window, not nominal pipeline duration | Pass | The guide states production promotes API traffic before the app, which is exactly the skew this PR avoids; the gate cited is the client-version floor, not a duration. |
| `docs/deployment-compatibility.md` | Deploy a tolerant version first; treat the floor increase as a later cleanup boundary | Pass | This is the expand half; the alias and the contract entry are removed together in step 2. |
| `docs/testing.md` | Cover the old/new combinations that coexist during rollout | Pass | Old-API-to-new-client is covered by the new `org-providers-tab.test.tsx` case and the new `runner-onboarding.test.ts` case; new-API-to-old-client is covered by `billing-status.test.ts` asserting the alias is still emitted. |
| `docs/testing.md` | No duplicate case that merely pins incidental implementation | Pass | One case per reachable rollout combination; precedence between the two names is not separately tested because the API emits them from one source, so they cannot differ. |
| `docs/testing/testing-external-behavior.md` | Platform tests verify the page, not a service return value | Pass | The alias fallback is asserted through the rendered model picker. An earlier signal-level unit test asserting the function's return value was removed in `d748ef7` for exactly this reason. |
| `docs/testing/testing-external-behavior.md` | API tests assert HTTP responses, not service returns or DB rows | Pass | `billing-status.test.ts` asserts `response.body`; entitlement state is set through the existing fixture boundary already used by the suite. |
| `docs/testing/app-testing.md` | No helper-only unit test when a rendered page can cover the behavior | Pass | Covered by the page test in `org-providers-tab.test.tsx`, which uses `setupPage` through `openModelSettings()`. |
| `docs/testing/app-testing.md` | Page tests enter through the bootstrapped Router and assert observable state | Pass | The new case asserts the `Pro` option affordance a user sees, and was verified to fail when the alias is removed from the read chain. |

### Findings

No P0 or P1 findings.

Non-blocking observations:

- `e2e/playwright/lib/runner-onboarding.ts` gains the same `??` chain as the
  platform. It is test-support code rather than a shipped surface, so its removal
  is not gated by a production floor; it is still declared and pointed at step 2
  so the cleanup does not get missed.
- `icon-mono-fixtures.json` now carries both names. That is a static style-migration
  fixture, so it only needs to keep matching the real step-1 wire, which it does.

### Fallbacks

Three fallbacks are introduced. All are declared in the PR body.

| File and symbol | Protects | Surface and gate | Verdict |
| --- | --- | --- | --- |
| `billing-status.service.ts` `BillingStatusResponse.restrictedVm0Models` | An installed app that reads only the retired name; because the contract marks it `.optional()`, its absence would silently select `LEGACY_TIER_CAPABILITIES` rather than fail | old web/app -> API; replacement app live and the client-version floor excludes the old build | Justified, bounded, declared |
| `org-plan-capabilities.ts` `orgPlanCapabilitiesFromBilling` | An API from before this release that sends only the retired name | new web/app -> old API; no such API serving or retained for rollback | Justified, bounded, declared |
| `runner-onboarding.ts` `readRunnerPaidEntitlement` | Same, for the e2e entitlement probe | new client -> old API; same gate | Justified, bounded, declared |

Each one names its surface, gate, and removal condition in code, and #33658
remains open as the step-2 follow-up that deletes the branch, the contract entry,
and the alias coverage together.

The skew this bounds is a wrong UI followed by a server rejection, not an
entitlement bypass: admission reads the server's own capabilities at
`run-admission.service.ts:259`, independent of the browser. The PR body states it
in both directions, including that a tier-derived legacy table lands correctly for
orgs matching their tier default and wrongly for the orgs that are the reason the
field is sent at all.

### Testing

Coverage added for each reachable rollout combination:

- `billing-status.test.ts` asserts both names on the response. The `false` case
  also pins `typeof === "boolean"`, because both names are `.optional()` and
  `toBeFalsy()` alone would pass on an omitted field — the pre-existing weakness
  called out as trap 3 in the issue. No assertion was weakened or inverted.
- `org-providers-tab.test.tsx` adds a page case serving billing status as a
  pre-rename API would, with only `restrictedVm0Models` set, asserting the `Pro`
  restriction through the model picker. Confirmed to fail when the alias is
  dropped from the read chain, so it is not vacuous.
- `runner-onboarding.test.ts` adds the alias case and switches the existing case
  to the current name, so both names are exercised.

Both fixture bridges in `test-fixtures/org-plan-entitlement.ts` are removed rather
than relocated, and the PR body explains what each reconciled. No test depended on
them: the fixture's parameter and state types are explicit, so excess-property
checking forced every call site to be renamed, and the API type check passes.

Verification observed in the authoring environment: `check-types` and `lint` pass
for `api`, `@okouai/app`, and `@okouai/api-contracts`; `knip` is clean;
`billing-status.test.ts` passes 33/33; `model-policies`,
`chat-threads-model-selection`, and `cron-billing-entitlements` pass; the new page
case passes and fails without the fallback.

Two suites fail locally for environment reasons and are explicitly not claimed as
passing. `retired-run-admission` and `cron-connector-catalog` fail on
`claimRunnerJob` returning `404 Job not found in queue` (branch 5, unmodified
`main` 6 — a strict superset), and the platform `okou-page` view suites fail on
timeouts with a set that varies per run (branch 7 then 10, unmodified `main` 13).
Both baselines were produced by re-running the identical selection on unmodified
`main` in the same environment. Required CI is the authority for those.

No material coverage gap remains for step 1.

PR #33680, opened 2026-09-12
feat: enable agent math and workflow fuzzy search for all users

bingjiez666:
## Summary

Agent message math rendering and workflow fuzzy matching in the chat composer currently default to staff organizations. Enable `agentMessageMath` and `composerWorkflowFuzzySearch` for all users by setting both registry defaults to `true` and removing their staff allowlists. Existing per-user overrides continue to apply.

Update the existing rollout-matrix assertion to expect math rendering for non-staff organizations.

## Verification

- Passed: feature-switch test file (30 tests).
- Passed: `@okouai/core` type checking and lint.
- Passed: Prettier for both changed files and `git diff --check`.
- Local Knip could not load the API ESLint configuration because the filtered dependency install omits `@okouai/eslint-rules/api`; full Knip verification is delegated to PR CI.
- Full local Vitest and a local dev server were not run, per the requested verification policy.

### DIFF
--- a/turbo/packages/core/src/__tests__/feature-switch.test.ts
+++ b/turbo/packages/core/src/__tests__/feature-switch.test.ts
@@ -212,7 +212,7 @@ describe("getAllFeatureStates", () => {
     });
     expect(otherOrgStates[FeatureSwitchKey.Lab]).toBe(false);
     expect(otherOrgStates[FeatureSwitchKey.OkouDebug]).toBe(false);
-    expect(otherOrgStates[FeatureSwitchKey.AgentMessageMath]).toBe(false);
+    expect(otherOrgStates[FeatureSwitchKey.AgentMessageMath]).toBe(true);
     expect(otherOrgStates[FeatureSwitchKey.ProgressiveArtifactPreview]).toBe(
       false,
     );
--- a/turbo/packages/core/src/feature-switch.ts
+++ b/turbo/packages/core/src/feature-switch.ts
@@ -73,8 +73,7 @@ const FEATURE_SWITCHES: Record<FeatureSwitchKey, FeatureSwitch> = {
   [FeatureSwitchKey.ComposerWorkflowFuzzySearch]: {
     maintainer: "bingjie@okou.ai",
     description: "Fuzzy workflow name matching in the chat composer",
-    enabled: false,
-    enabledOrgIdHashes: STAFF_ORG_ID_HASHES,
+    enabled: true,
   },
   [FeatureSwitchKey.Dummy]: {
     maintainer: "ethan@okou.ai",
@@ -405,8 +404,7 @@ const FEATURE_SWITCHES: Record<FeatureSwitchKey, FeatureSwitch> = {
     maintainer: "bingjie@okou.ai",
     description:
       "Render explicit LaTeX delimiters in Agent messages as native MathML.",
-    enabled: false,
-    enabledOrgIdHashes: STAFF_ORG_ID_HASHES,
+    enabled: true,
   },
   [FeatureSwitchKey.MarkdownTime]: {
     maintainer: "ethan@okou.ai",


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 35998f94e5401679bea9b7ee9b53b8c56afb1740
PR_AUTO_REVIEW_VERDICT: LGTM

### Summary

Enables `AgentMessageMath` and `ComposerWorkflowFuzzySearch` for all users by setting `enabled: true` and dropping their staff-only `enabledOrgIdHashes` allowlists. Per-user overrides still win because the override check short-circuits before the registry. The staff-rollout-matrix assertion is updated to expect math rendering for a non-staff organization.

Reviewed revision: `35998f94e5401679bea9b7ee9b53b8c56afb1740`.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | `isFeatureEnabled` (`feature-switch.ts:680-706`) reads `ctx.overrides[key]` first, then `featureSwitch.enabled`, so the client reader `feature-switch.ts:153` and `tiptap-workflow-composer.tsx:467` now resolve `true` for every org while overrides remain authoritative. |
| `REVIEW.md` | Require meaningful coverage for new behavior or changed expectations | Pass | The changed rollout-matrix assertion (`feature-switch.test.ts:215`) directly covers the flipped default for a non-staff org, matching the existing `MorningBrief` released-switch pattern at line 231. |
| `REVIEW.md` | For a changed/rerouted flow, require regression coverage and containment | Pass | No flow is rerouted; both switches remain registered with their per-user override layers, so the changed path is contained and still rollback-able. |
| `REVIEW.md` | Report concrete practice violations only, not preferences or inferred bugs | Pass | Diff is 2 files, +3/-5, limited to the two registry entries and one assertion; no unrelated edits. |
| `docs/fallback.md` §2 | Features behind a feature switch need no fallback while pre-GA | Pass | Both switches stay registered and off-by-default-reversible; no compatibility branch, dual read/write, or backfill is added. |
| `docs/fallback.md` §9 | Every newly introduced fallback must be declared in the PR body | N/A | The diff introduces no fallback; it only flips a default and removes a now-redundant staff allowlist. |
| `docs/bad-smell.md` §4, §7 | Interface changes and fallback patterns | Pass | The `FeatureSwitch` interface is unchanged; removed fields were dead config because `enabled: true` short-circuits the hash checks. |
| `docs/testing.md` | Tests should assert the contract a caller relies on; avoid pinning incidental implementation | Pass | The updated assertion reads `getAllFeatureStates` output, the same public resolver production uses; existing Platform tests that pin the switch explicitly via `featureSwitches` overrides are unaffected. |
| `.claude/skills/feature-switch/SKILL.md` | Use the documented registry configuration for rollout changes | Pass | The change moves both entries from staff-only (`enabled: false` + `enabledOrgIdHashes`) to the documented `enabled: true` GA state. |

### Findings

No P0 or P1 findings.

Non-blocking observations:

- Every other `enabled: true` entry in `FEATURE_SWITCHES` (77 switches inspected) carries no `enabledOrgIdHashes`, so removing the allowlist matches existing repo convention rather than introducing a new shape.
- Both switches therefore reclassify from the Lab "Beta" group to "Released" through `getFeatureSwitchRolloutStage`, which is the intended user-visible consequence of the rollout.

### Testing

- `feature-switch.test.ts` (30 tests) passes, including the updated `should reflect the current staff org rollout matrix` case, which is the only assertion that needed to change.
- `@okouai/core` type checking, lint, and Prettier pass; `git diff --check` is clean.
- Existing Platform coverage for both features pins the switch explicitly (`markdown-math.test.tsx:61,118,144,166`, `markdown-parser-isolation.test.tsx:71`, `chat-composer-workflows.test.tsx:571-736`), so those cases continue to exercise both the enabled and disabled branches.
- Full local Vitest and a local dev server were intentionally not run, per project policy; repository CI owns that verification.

PR #33682, opened 2026-09-12
test(api): delete the projection-cache observation suites and their diagnostics helper

lancy:
## Summary

Closes #33654. Repair for the one deviation the controller found when accepting G1c: the "Addendum (2026-09-12, from G1b's acceptance)" assigned `helpers/projection-observations.ts` to this child after I had already read the issue, so PR #33681 left it to #33652 — which does not touch it. Without this PR the baseline would keep exactly that one entry after G1a merges and the epic's "lint rule active with an empty baseline" criterion could not be met by anyone.

Deletes `src/signals/routes/__tests__/helpers/projection-observations.ts` and its six one-case consumers `run-projection-cache-{observations,reuse-two,reuse-four,reuse-six,reuse-sixteen,reuse-eviction}.bdd.test.ts`, and removes the baseline entry.

### Why deletion rather than a route assertion

The helper read `context.mocks.axiom.sdkIngest` and asserted `connector_catalog_projection_cache_outcome`, `connector_catalog_runtime_selection_source` and `connector_catalog_projection_cache_observation` on the `api_dispatch_connector_catalog_load_runtime_snapshot` timing event. Each suite created runs through the production route and cancelled them, but the timing record was the only thing it inspected — the six modules differ only in how far apart they replay a selection so the observation lands in a different reuse bucket.

I checked the production side before deleting:

- `loadConnectorRuntimeSelection` (`connector-catalog-runtime.service.ts:1132-1219`) returns the same `ConnectorRuntimeSelection` on `hit` (`:1186`), `in_flight` (`:1196`) and `miss` (`:1218`). The cache is a process-local performance optimisation; the run's connector configuration is byte-identical either way, so no response or persisted effect differs.
- `observeRuntimeSelection` (`:876-921`) is pure telemetry-observation bookkeeping. Its only call site is `timing.recordProjectionCacheObservation(...)` at `:1179`, and the returned bucket label (`first_observation`, `not_in_recent_history`, `reuse_1`, `reuse_2`, `reuse_3_4`, `reuse_5_8`, `reuse_9_16`) reaches nothing but the telemetry field. It does not influence the cache, eviction of the real cache entry, or the selection.
- `connector_catalog_projection_cache_outcome` is produced only by `connector-catalog-load-timing.service.ts:419`.
- No production route under `src/signals/routes` exposes the cache, and none of the 40 existing `test-*` state routes covers the connector-catalog projection cache. Adding one is an explicit non-goal.

[...]

PR #33679, opened 2026-09-12
test(api): replace diagnostics pins in the chat and run lifecycle suites

lancy:
Closes #33652 (parent epic #33656, finding 1).

## Summary

The chat and run lifecycle suites pinned internal diagnostics: logger records through `context.mocks.axiomLogging`, `api-dispatch` timing spans through `context.mocks.axiom.sdkIngest`, run-context and template-usage snapshots through `context.mocks.axiom.ingest`, and Axiom trace payloads parsed out of `api.axiom.co` MSW handlers. None of that is an external contract, so the instrumentation behind it could not move.

For every such assertion: when the case also had an HTTP or persisted effect, the diagnostic assertion was dropped and the case kept; when the span or the log record was the case's only subject, the case was deleted. No assertion was replaced with `vi.spyOn`, Axiom payload parsing, or `sdkIngest` inspection.

- `run-lifecycle.bdd.test.ts`, `chat-events.bdd.test.ts`, `webhooks-callbacks.bdd.test.ts`, `chat-callbacks.bdd.test.ts`: removed the `apiDispatchTimingEventsForRun` / `sandboxOperationEventsForRun` / `claimRouteTimingEventsForRun` helper families with their `api_dispatch_*`, `claim_route_*`, `span_kind` and `*_count_bucket` constants, and the logger-mock assertions. Tests left without a subject were deleted; several titles were rewritten to name what they still assert (for example `emits api dispatch timing for exact-empty direct dispatch runs` → `claims an exact-empty direct dispatch run without connector scope`).
- `webhooks-callbacks.bdd.test.ts`: Axiom MSW handlers no longer record or assert on the ingest payload. They count requests or read the target dataset from the URL; where the API's partial-ingest check requires it, a shared `ingestedEventCount` helper sizes the mocked accepted-event count from the request, mirroring the provider contract without inspecting the payload. Failure-simulation handlers (503, 500, malformed status, blocked ingest) are unchanged, so the acknowledgement, deadline, cancellation and rejection behaviour still has coverage.
- The two secret-redaction checks in `chat-events.bdd.test.ts` are expressible without the logger mock, so they were rewritten rather than moved: they keep asserting against the session history, thread events, public run state and checkpoint objects, and only the logger-mock array was dropped from the redaction snapshot. Nothing was moved into `lib/__tests__/log.test.ts`.
[...]

Issue #33667, opened 2026-09-12
refactor(api): delete dead fields and repeated guards in agent-run-create


lancy:
Parent: #33656 (finding 3), proposal P2 of the G6 review (#33651, [findings comment](https://github.com/vm0-ai/vm0/issues/33651#issuecomment-5644455360)). Baseline: main `9c83c9b87a`; the cited lines were re-verified there (`agentName` at `:617`, `:633`, `:636`, `:6001`, `:6018`, `:6213`; guards at `:8040`, `:8562`, `:8593`; `createdRunResponse` `:8838`; `committedAtomicLaunchResponse` `:10334`; `ingestRunContextSnapshot` / `recordQueuedRunEnqueueTelemetry` best-effort wrappers from `:6973`).

## Background

Dead fields, guards the immediate caller already rules out, an unused parameter, and duplicated best-effort wrappers in `turbo/apps/api/src/signals/services/agent-run-create.service.ts`. The review's evidence, verbatim:

#### D6 — `agentName` is selected from the database twice and never read

**Pattern:** dead compatibility field.

**Evidence.** `ResolvedAgentExecution.agentName` is declared at `:617`, negated at `:636`, and written at `:6018` (`agentName: row.agentName || undefined`) and `:6213` (`agentName: snapshot.agent.name || undefined`). Grep over the whole file finds **no reader**; `ResolvedAgentExecution` and `ResolvedRunExecution` are not exported (`:613`, `:639`), so no external reader exists either. The value is produced by `agents.name` in the `resolveByAgentId` SELECT (`:6001`) and the `resolveBySessionId` SELECT (`:6141`).

The `|| undefined` is itself slop: `agents.name` is `varchar("name", {length: 64}).notNull()` (`packages/db/src/schema/agent.ts:29`) under a unique index on `(org_id, name)`, so the branch exists for an empty-string name nothing writes.

**Fix.** Delete the field, both writers, and the `name` column from both SELECTs.

**Risk.** Low.

#### D9 — duplicated launch-commit blocks

**Pattern:** copy-pasted block.

**Evidence.**

[...]

Issue #33654, opened 2026-09-12
test(api): replace diagnostics pins in the billing, cron, webhook and workflow suites


lancy:
Parent: #33656 (finding 1). Sequenced after #33647, #33648, #33649 and #33650 are merged and controller-accepted; rebase on the then-current main first.

## Background

Billing, cron, webhook and workflow suites pin diagnostics through `context.mocks.axiomLogging`: `cron-billing-entitlements.test.ts` (14 references; 12 silence pins such as `not.toHaveBeenCalledWith("Stripe subscription snapshots reconciled")`, `:329-336`, `:708-721`, `:771-772`), `codex-reset-credit-expiry.test.ts` (10), `billing-redeem-code.test.ts` (4), `pi-memory-stage1-worker.test.ts` (7), `memory-summary-projection.test.ts` (4), `webhooks-google-calendar.test.ts` (5), `webhooks-gmail.test.ts` (5), `webhooks-google-forms.test.ts` (4), `workflow-automations.test.ts` (4), `webhooks-workflow-automations.test.ts`, `workflows.test.ts`, `webhooks-github-workflow.test.ts` (telemetry readers), `cron-connector-catalog.test.ts` (2), `cron-sync-skills.test.ts` (1), `connector-runtime-wakeups.test.ts` (3), `webhooks-agent-health-usage-telemetry.test.ts` (2), `test-teams-state.test.ts` (1). Cron and webhook handlers answer 200 by design, so several of these cases used the log as the only observable; the persisted effect (a reconciled row, a queued automation event, an updated connector state) is readable through production or test-state routes.

### Addendum (2026-09-12, from G2's acceptance)

`apiTestDiagnosticsBaseline` still lists five files that G2 (#33668) already cleaned because its branch predates G0's merge: `src/signals/routes/__tests__/me-model-providers-upsert.test.ts`, `webhooks-agent-firewall-auth-aws.test.ts`, `webhooks-agent-firewall-auth-custom-oauth.test.ts`, `webhooks-agent-firewall-auth-google-analytics.test.ts`, `webhooks-agent-firewall-auth.bdd.test.ts`. Remove those entries in this PR (inert ignores; the baseline must only shrink). `chat-threads.bdd.test.ts` is handled by #33652.

### Addendum (2026-09-12, from G1b's acceptance)

[...]

PR #33684, opened 2026-09-12
test(api): inject axiom sdk telemetry failures from the mock definition site

### LINKED ISSUE #33652: test(api): replace diagnostics pins in the chat and run lifecycle suites
Parent: #33656 (finding 1). Sequenced after #33647, #33648, #33649 and #33650 are merged and controller-accepted; rebase on the then-current main first.

## Background

The three chat and run lifecycle suites pin diagnostics: `chat-events.bdd.test.ts` (48 `mocks.axiomLogging` references; 13 `op_type` telemetry pins such as `api_dispatch_build_runner_job_payload` via `expectPiLaunchResourceTiming`, `:1285-1300`), `run-lifecycle.bdd.test.ts` (19 logger references; 16 `op_type` / `action_type` pins), `webhooks-callbacks.bdd.test.ts` (11 logger; 17 pins, 18 `api.axiom.co` handlers), plus `test-runtime-state.test.ts` (5), `chat-threads.bdd.test.ts` (2), `chat-event-snapshot.test.ts` (2), `chat-event-archive-consumers.test.ts` (2), `cron-snapshot-chat-events.test.ts` (4), `cron-monitor-chat-event-queue.test.ts` (1), `shared-threads.test.ts` (3, plus telemetry), `goal-schema-contraction.test.ts` (telemetry). The `api-dispatch-timing` telemetry they pin is internal instrumentation (`api-dispatch-timing.service.ts`, 467 lines) with no external contract. #33650 already removed the auxiliary-generation pins from `chat-events.bdd.test.ts` and `chat-callbacks.bdd.test.ts`; #33648 removed the "token refresh failed" pin.

### Addendum from the G6 review (#33651, 2026-09-12)

`run-lifecycle.bdd.test.ts` pins the service's internal timing surface, not only through the `sdkIngest` harness at `:994` (which the #33647 lint catches) but through constants the lint cannot see: 47 `api_dispatch_*` action-type literals at `:327-373`, the three `CustomConnectorRuntimeBuildStats.flush()` phase names at `:514-518`, seven count-bucket names at `:523-531`, and 112 `count_bucket` assertions across the file. Treat every one of them as a diagnostics pin under this issue's rule: delete the assertion, or the case when the span was its only subject. Do the same for the `api_dispatch_*` constants in `chat-events.bdd.test.ts`. Do not change `api-dispatch-timing.service.ts` or the service's `timing.measure` calls here; the G6 review proposes that separately (P1 on #33651).

### Addendum (2026-09-12, from G2's acceptance)

`chat-threads.bdd.test.ts` is already clean (G2 #33668 removed its only logger pin) but is still listed in `apiTestDiagnosticsBaseline`; remove that entry in this PR together with the entries of the files you migrate.

### Addendum (2026-09-12, from G5's acceptance)

Add `chat-callbacks.bdd.test.ts` to this child's file list: after G5 (#33671) it still holds 6 `mocks.axiomLogging` references, 1 `sdkIngest` reference and 2 `op_type` pins that are not auxiliary-generation related, and no other child owns it. `shared-threads.test.ts` and `chat-event-archive-consumers.test.ts` were partially cleaned by G5; finish them here if any reference remains and remove their baseline entries.

## Design

For every remaining logger or telemetry assertion in the files above: if the case has an HTTP or persisted effect, keep the case and drop the diagnostic assertion; if the case exists only to prove a record or a timing span, delete it. Do not replace pins with `vi.spyOn` on the logger, MSW payload parsing of `api.axiom.co`, or `sdkIngest` inspection. Keep the two secret-redaction checks in `chat-events.bdd.test.ts` (`:10110`, `:10235`) only if they can be expressed without the logger mock; otherwise move one redaction case to `lib/__tests__/log.test.ts` and delete the rest. Remove every migrated file from the #33647 baseline. Failure-simulation stubs that make the Axiom endpoint answer 503 may stay; payload parsing may not.

## Explicit non-goals

- No production code changes; no change to `api-dispatch-timing.service.ts`; no new tests.

## Rollout

One PR. Test-only. Ordinary release train.

## Acceptance criteria

- [ ] None of the files above references `mocks.axiomLogging`, `sdkIngest`, `useRealTelemetry`, or parses an `api.axiom.co` request body; they are removed from the #33647 baseline.
- [ ] `git diff --stat` shows only test files and `eslint.config.mjs`.
- [ ] API lint and the affected suites pass in CI; merged through the protected queue via same-thread `/pr-auto`.

lancy:
Closes #33652 (repair after the controller acceptance of #33679; parent epic #33656, finding 1).

## Summary

#33679 deleted three cases whose subject was an HTTP or persisted effect, because their **setup** reached `context.mocks.axiom.sdkIngest.mockImplementation(...)` — the #33647 lint bans that member access in test files together with the reads. The contract only allowed deleting a case when a log record or a timing span was its only subject, so those three should have been kept. They prove that a telemetry write failing does not break the primary effect, and #33667 removed the `Failed to record …` warns that were the other way to see it.

Root cause: `signals/external/sandbox-op-log.ts` writes through its own `@axiomhq/js` client, so only the SDK-level stub intercepts it, and `src/__tests__/mocks.ts` offered no failure-injection helper.

### The helper

`mockAxiomSdkTelemetryFailure(failure)` in `src/__tests__/mocks.ts` — the stub definition site the lint rule already exempts ("it asserts nothing itself"). It takes a discriminated union and returns nothing:

- `{ mode: "ingest", datasets?: readonly string[] }` — the SDK client's `ingest` throws for the named datasets (exact names, including `AXIOM_DATASET_SUFFIX`); omit `datasets` to fail every ingest.
- `{ mode: "flush" }` — the SDK client's `flush` rejects. The SDK flushes a client, not a dataset, so this mode deliberately takes no filter rather than silently ignoring one.

It also turns on the real `external/axiom.ts` wrappers. That is part of the failure mode, not a separate step: the failure has to be injected *below* `ingestToAxiom` / `flushAxiom` so the guards that decide whether the request survives are the production ones. Injecting into the mocked wrappers instead makes the error escape somewhere production never sees it — that is what made the earlier attempts in #33679 fail CI with unhandled `Telemetry unavailable` errors.

No event payload is handed back to the caller, and the helper only drives `apiTestMocks.axiom.{sdkIngest,flush,useRealTelemetry}`, all three of which `resetApiTestMocks` already resets (`mocks.ts:1406-1414`), so it needs no teardown of its own. Test files call the helper and never name `sdkIngest`, `useRealTelemetry`, or a logger mock.

### Restored cases

- `run-lifecycle.bdd.test.ts` `keeps a direct launch claimable when sandbox telemetry ingest fails` — fails the `vm0-sandbox-op-log-dev` ingest, asserts `created.status === "pending"` and the claimed prompt. Placed next to its untouched sibling `keeps a direct launch claimable when run-context ingest fails`.
- `run-lifecycle.bdd.test.ts` `keeps a queued launch visible when enqueue telemetry fails` — fails the same dataset after two active runs hold the concurrency limit, so the queued launch is the first whose enqueue telemetry is written after the failure starts; asserts `queued.status === "queued"` and the queue listing. Its former `sandboxOperationEventsForRun` pin is **not** restored.
- `shared-threads.test.ts` `it.each(["ingest", "flush"])` `preserves a valid share when telemetry %s fails` — both arms are faithful, so both are restored: the ingest throw is contained by `auxiliary-generation.service.ts`'s `safeSync`, and the flush rejection by `flushAxiom`'s `Promise.allSettled` + log (`external/axiom.ts:409-421`). Asserts the share persists with the fixed title and no Sentry capture.

No telemetry read was restored anywhere. No production code, lint rule, or baseline array changed; the baseline still holds only #33654's `helpers/projection-observations.ts`, which #33682 owns.

## Verification

- `pnpm --filter api lint` — passes, with the diagnostics rule active against both test files.
- `pnpm --filter api check-types` — passes.
- `pnpm knip` — clean (the failure type stays module-local for that reason).
- `prettier --check` on the changed files — passes.
- `vitest run src/signals/routes/__tests__/shared-threads.test.ts` against a local PostgreSQL 18 instance with `packages/db` migrations applied: **18/18 passed**, including both restored arms.
- `run-lifecycle.bdd.test.ts` cannot be baselined in the authoring environment: on an unmodified `origin/main` (`44746aee`) the file already reports 163 failed / 64 passed locally, all with `Job not found in queue` from `claimRunnerJob`, because the suite depends on fixture state this environment does not provide. With this branch it reports 164 failed / 65 passed of 229 — the two added cases, of which `keeps a queued launch visible when enqueue telemetry fails` **passes** and `keeps a direct launch claimable when sandbox telemetry ingest fails` fails exactly like its untouched sibling `keeps a direct launch claimable when run-context ingest fails` does on clean `main`. Running that case alone shows the injection working as intended: `[WARN][SandboxOpLog] Failed to ingest sandbox operation log { error: Error: Axiom SDK ingest failed }` through `safeSync` in `sandbox-op-log.ts:68`, with the failure contained. The suite is left to CI's `test-api` shards.

### DIFF
--- a/turbo/apps/api/src/__tests__/mocks.ts
+++ b/turbo/apps/api/src/__tests__/mocks.ts
@@ -1330,6 +1330,44 @@ export async function withRealAxiomLoggingForTest(
     });
 }
 
+// A telemetry write that fails is not a diagnostic: production decides whether
+// the request survives it, and `external/axiom.ts` plus
+// `external/sandbox-op-log.ts` own the guards that make that decision. Both
+// reach Axiom through this file's `@axiomhq/js` stub, so the only faithful
+// place to break a write is that SDK boundary; replacing the wrappers instead
+// lets the error escape where production never sees it. Routing through the
+// real wrappers is part of the failure mode, not a separate step, so this
+// helper enables them. Tests name the datasets to break and assert the HTTP
+// response and the effect afterwards; no event payload is handed back.
+type AxiomSdkTelemetryFailure =
+  // `datasets` lists exact Axiom dataset names, including the
+  // `AXIOM_DATASET_SUFFIX`. Omit it to fail every ingest.
+  | { readonly mode: "ingest"; readonly datasets?: readonly string[] }
+  // The SDK flushes a client, not a dataset, so this mode takes no filter.
+  | { readonly mode: "flush" };
+
+export function mockAxiomSdkTelemetryFailure(
+  failure: AxiomSdkTelemetryFailure,
+): void {
+  apiTestMocks.axiom.useRealTelemetry.mockReturnValue(true);
+  if (failure.mode === "flush") {
+    apiTestMocks.axiom.flush.mockRejectedValue(
+      new Error("Axiom SDK flush failed"),
+    );
+    return;
+  }
+  const datasets = failure.datasets;
+  apiTestMocks.axiom.sdkIngest.mockImplementation((dataset: unknown) => {
+    if (
+      datasets === undefined ||
+      (typeof dataset === "string" && datasets.includes(dataset))
+    ) {
+      throw new Error("Axiom SDK ingest failed");
+    }
+    return undefined;
+  });
+}
+
 export function getApiTestMocks(): ApiTestMocks {
   return apiTestMocks;
 }
--- a/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/run-lifecycle.bdd.test.ts
@@ -48,6 +48,7 @@ import { v5 as uuidv5 } from "uuid";
 
 import { env, mockEnv, mockOptionalEnv } from "../../../lib/env";
 import { clearMockNow, mockNow, now, nowDate } from "../../../lib/time";
+import { mockAxiomSdkTelemetryFailure } from "../../../__tests__/mocks";
 import { accept, testContext } from "../../../__tests__/test-context";
 import { setupApp } from "../../../__tests__/test-helpers";
 import { server } from "../../../mocks/server";
@@ -186,6 +187,9 @@ import { testCronCleanupSandboxesStateRoutes } from "../test-cron-cleanup-sandbo
 const context = testContext();
 const callbackStore = createStore();
 const fixtureStore = createStore();
+// `sandbox-op-log.ts` composes this name from AXIOM_DATASET_SUFFIX, which the
+// test environment stubs as "dev".
+const SANDBOX_OP_LOG_DATASET = "vm0-sandbox-op-log-dev";
 const ASSISTANT_EVENT_ID_NAMESPACE = "bfec4fb6-d5b8-43e4-a72a-9f58f87d7e01";
 const TEST_DATA_KEY = Buffer.from("0123456789abcdef0123456789abcdef", "utf8");
 
@@ -2884,6 +2888,28 @@ describe("CHAIN-RUN: entitled run lifecycle through runner and sandbox webhooks"
     await api.requestCancelRun(actor, created.runId, [200]);
   });
 
+  it("keeps a direct launch claimable when sandbox telemetry ingest fails", async () => {
+    const api = createRunsApi(context);
+    const { actor, agentId } = await entitledRunActor();
+    const prompt = "sandbox telemetry should not block launch";
+    mockAxiomSdkTelemetryFailure({
+      mode: "ingest",
+      datasets: [SANDBOX_OP_LOG_DATASET],
+    });
+
+    const created = await api.createRun(actor, {
+      agentId,
+      prompt,
+      modelProvider: "anthropic-api-key",
+    });
+
+    expect(created.status).toBe("pending");
+    const claim = await api.claimRunnerJob(created.runId);
+    expect(claim.prompt).toBe(prompt);
+
+    await api.requestCancelRun(actor, created.runId, [200]);
+  });
+
   it("resumes the Agent execution session for a continued run", async () => {
     const api = createRunsApi(context);
     const webhooks = createWebhookCallbackApi(context);
@@ -5414,6 +5440,44 @@ describe("RUN-01: admission boundaries beyond request validation", () => {
     await api.requestCancelRun(actor, overLimit.runId, [200]);
     await api.requestCancelRun(actor, fresh.runId, [200]);
     await api.requestCancelRun(actor, queued.runId, [200]);
+    await api.requestCancelRun(actor, second.runId, [200]);
+  });
+
+  it("keeps a queued launch visible when enqueue telemetry fails", async () => {
+    const api = createRunsApi(context);
+    const { actor, agentId } = await entitledRunActor();
+
+    const first = await api.createRun(actor, {
+      agentId,
+      prompt: "active run before telemetry failure one",
+      modelProvider: "anthropic-api-key",
+    });
+    const second = await api.createRun(actor, {
+      agentId,
+      prompt: "active run before telemetry failure two",
+      modelProvider: "anthropic-api-key",
+    });
+    // Those two runs hold the concurrency limit, so the next launch is the
+    // first one whose enqueue telemetry is written after the failure starts.
+    mockAxiomSdkTelemetryFailure({
+      mode: "ingest",
+      datasets: [SANDBOX_OP_LOG_DATASET],
+    });
+
+    const queued = await api.createRun(actor, {
+      agentId,
+      prompt: "queued run should survive telemetry failure",
+      modelProvider: "anthropic-api-key",
+    });
+
+    expect(queued.status).toBe("queued");
+    const queue = await api.readRunQueue(actor);
+    expect(queue.body.queue).toContainEqual(
+      expect.objectContaining({ runId: queued.runId }),
+    );
+
+    await api.requestCancelRun(actor, queued.runId, [200]);
+    await api.requestCancelRun(actor, first.runId, [200]);
     await api.requestCancelRun(actor, second.runId, [200]);
   });
 
--- a/turbo/apps/api/src/signals/routes/__tests__/shared-threads.test.ts
+++ b/turbo/apps/api/src/signals/routes/__tests__/shared-threads.test.ts
@@ -4,6 +4,7 @@ import { HttpResponse, http } from "msw";
 import { beforeEach, describe, expect, it, onTestFinished } from "vitest";
 import { z } from "zod";
 
+import { mockAxiomSdkTelemetryFailure } from "../../../__tests__/mocks";
 import { accept, testContext } from "../../../__tests__/test-context";
 import { setupApp } from "../../../__tests__/test-helpers";
 import { mockOptionalEnv } from "../../../lib/env";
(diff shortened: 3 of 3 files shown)


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: cec17c530210231653b0f7639fe9f2d9dc393d8b
PR_AUTO_REVIEW_VERDICT: LGTM

Reviewed HEAD: `cec17c530210231653b0f7639fe9f2d9dc393d8b`. Practice documents read from `main` and verified byte-identical to the checkout (none is modified by this PR): `REVIEW.md` `c0579b2d`, `docs/docs.md`, `docs/testing.md` `baa5172a`, `docs/testing/testing-external-behavior.md` `c4dd0723`, `docs/testing/api-testing.md` `65b9295c`, `docs/testing/anti-patterns.md` `0e474022`.

### Summary

Adds `mockAxiomSdkTelemetryFailure` to `src/__tests__/mocks.ts` and restores the three cases #33679 deleted because their *setup* named `context.mocks.axiom.sdkIngest`. The helper breaks the write at the `@axiomhq/js` stub — the only boundary `sandbox-op-log.ts` reaches — and routes through the real `external/axiom.ts` wrappers so the guards that decide whether the request survives are production's. The restored cases assert only HTTP status and persisted effects.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | N/A | Test-only. `git diff --name-only` is three files under `src/__tests__` and `src/signals/routes/__tests__`; no production module, no lint rule, no baseline array. |
| `REVIEW.md` | Require meaningful coverage for new behavior and regression fixes; a rewritten expectation alone does not prove an intentional replacement | Pass | The three cases are restored with the same subjects the controller recorded: `created.status === "pending"` plus the claimed prompt; `queued.status === "queued"` plus the queue listing; the share persisting with the fixed title and no Sentry capture. The deleted `sandboxOperationEventsForRun` pin from the queued case is **not** restored. |
| `REVIEW.md` | API tests that assert log output require `Changes Requested`, outside the logger's own suite and one redaction check | Pass | No added line asserts a level, message, field, or ingest payload. `grep` over the diff: no `axiomLogging`, no `sdkIngest` read, no `useRealTelemetry` in either test file; the helper returns `void`. |
| `REVIEW.md` | A PR that answers a log-noise issue by adding another outcome classifier must justify why the record is needed | N/A | No classifier, dimension, or record is added; this PR only restores effect assertions. |
| `REVIEW.md` | Fallbacks must be declared in a `Fallbacks` PR-body section | N/A | No fallback, default, retired path, or rollout switch is introduced, so the section is correctly absent. |
| `REVIEW.md` | Persisted data, contracts, schema, compatibility windows | N/A | Nothing persisted or contractual changes. |
| `docs/testing.md` | Verify externally observable state or an HTTP response available through a production endpoint | Pass | Every restored assertion goes through a production route: `POST /api/agent-runs`, the runner claim, `GET` run queue, `POST /api/chat-threads/:threadId/shared-threads` and `expectSharedSnapshot`'s public reads. |
| `docs/testing.md` | In API tests log records are incidental implementation | Pass | Telemetry failure is used only as *context* for the request; the outcome asserted is the run status, the claim, the queue row and the share snapshot. |
| `docs/testing.md` | Mock external services at their boundary; use real internal code | Pass | The failure is injected into the `@axiomhq/js` stub — the external SDK — leaving `external/axiom.ts`, `external/sandbox-op-log.ts` and `auxiliary-generation.service.ts` running as real internal code. This is the AP-4/AP-6 boundary, and it is why the helper turns the real wrappers on rather than replacing them. |
| `docs/testing.md` | API tests use the centralized `testContext()` / `context.mocks` lifetimes | Pass | The helper only drives `apiTestMocks.axiom.{sdkIngest,flush,useRealTelemetry}`, and `resetApiTestMocks` already resets all three (`mocks.ts:1406-1414`: `useRealTelemetry` reset to `false`, `flush` and `sdkIngest` reset). It keeps no module state, so it needs no teardown of its own. |
| `docs/testing/testing-external-behavior.md` | API tests must not enable the real Axiom transport, intercept the ingest endpoint, or assert on logger calls | Pass | No transport is enabled: `axiomLogging.useRealTransport` stays `false`, so `new Axiom(...)` still returns the stub client and no request leaves the process; no `api.axiom.co` handler is registered. `useRealTelemetry` selects the real *wrapper functions* in `external/axiom.ts`, not a transport — see observation 1 for why the wording deserves care. |
| `docs/testing/testing-external-behavior.md` | Leave the boundary only for states impossible to construct through the production interface, and state the exception at the site | Pass | An Axiom outage cannot be produced through any endpoint. The 9-line comment above the helper states why the SDK boundary is the only faithful place, and the helper lives at the stub definition site rather than inside a generic fixture, which is what the guide asks ("do not hide exceptions inside generic fixtures"). |
| `docs/testing/testing-external-behavior.md` | The lint rule's exempt files are the stub definition site and the four subject suites | Pass | `eslint.config.mjs:817` exempts `src/__tests__/mocks.ts` verbatim as "the stub definition site [that] installs the logger and telemetry mocks that this rule stops tests from reading; it asserts nothing itself". The helper asserts nothing, and `pnpm --filter api lint` passes with the rule active against both test files. |
| `docs/testing/api-testing.md` | Route tests exercise the real app through `setupApp()` and the contract; only external services are mocked | Pass | No `setupApp`, contract or route wiring changed; no `vi.mock()`, service import or DB access was added. Importing a named helper from `__tests__/mocks` follows the established pattern (`image-io-generate`, `browser`, `intro-video-agent`, `log-axiom-transport`). |
| `docs/testing/anti-patterns.md` | AP-12: do not build a harness that switches on the real telemetry transport, intercepts the ingest endpoint with MSW, and asserts a level, message or field per outcome | Pass | None of the three elements is present: the transport stays stubbed, no MSW ingest handler is added, and nothing is asserted about a record. AP-12's prescribed alternative — "assert the HTTP response and the effect visible in the next request" — is exactly what the restored cases do. |
| `docs/testing/anti-patterns.md` | AP-1 / AP-7: do not assert mock calls or implementation details in place of behavior | Pass | The helper is used for setup only; no added line reads `.mock.calls`. See observation 2 for the trade-off this deliberately accepts. |
| `docs/testing/anti-patterns.md` | AP-5: no fake timers; synchronize on the observable result | Pass | No timer control added. `flushWaitUntilForTest()` is awaited before the snapshot read, as the neighbouring cases already do. |
| `docs/testing/anti-patterns.md` | AP-8: avoid duplicate cases and over-testing | Pass | Three cases restored, none added; the `shared-threads` arms differ in the production guard they exercise (`safeSync` in `auxiliary-generation.service.ts:96` vs `Promise.allSettled` in `external/axiom.ts:409-421`), so neither is redundant. |

### Findings

No P0 or P1 findings.

Non-blocking observations:

1. `turbo/apps/api/src/__tests__/mocks.ts:1340` — the helper always sets `useRealTelemetry` to `true`, which the two `run-lifecycle` cases do not need: `sandbox-op-log.ts` builds its dataset name itself and calls `client.ingest` directly, so it never passes through `ingestToAxiom`. The uniform mode keeps one faithful contract instead of a per-call-site switch, at the cost of moving those cases' unrelated `run-context` write from `mocks.axiom.ingest` onto the stub client. Neither case asserts on either surface, so nothing is pinned, and the queued-launch case passes locally.
2. Neither restored case can detect a *silently vacuous* injection: if `AXIOM_DATASET_SUFFIX` ever stopped being `dev`, `SANDBOX_OP_LOG_DATASET` would match nothing and the cases would pass without exercising a failure. Exposing a "did it fire" counter from the helper would close that, but asserting on it is the mock-call assertion AP-1 warns about, so the gap is recorded instead of guarded. It is currently live: running the case alone prints `[WARN][SandboxOpLog] Failed to ingest sandbox operation log { error: Error: Axiom SDK ingest failed }` from `safeSync` at `sandbox-op-log.ts:68`.
3. `turbo/apps/api/src/__tests__/mocks.ts:1342` — `AxiomSdkTelemetryFailure` is intentionally not exported, because callers pass object literals and Knip flags an exported type with no importer. Tests that want to build the argument separately will have to export it at that point.

### Testing

- `pnpm --filter api lint` passes with the `no-restricted-syntax` diagnostics rule active against both changed test files.
- `pnpm --filter api check-types` passes across all five projects, including `tsconfig.tests.json`.
- `pnpm knip` is clean.
- `prettier --check` passes on the three changed files.
- `vitest run src/signals/routes/__tests__/shared-threads.test.ts` against a local PostgreSQL 18 with `packages/db` migrations applied: **18/18 passed**, covering both restored `it.each` arms — so the ingest throw and the flush rejection are both contained by production and the share still persists.
- `run-lifecycle.bdd.test.ts` has no usable local baseline: on an unmodified `origin/main` (`44746aee`) it already reports 163 failed / 64 passed there, every failure being `Job not found in queue` from `claimRunnerJob`, including the untouched sibling `keeps a direct launch claimable when run-context ingest fails`. With this branch it reports 164 failed / 65 passed of 229 — the two added cases, of which the queued-launch case **passes** and the direct-launch case fails only in that same environmental way. The suite is left to CI's `test-api` shards, which are the gate.
- Coverage gaps introduced: none. Three previously deleted effect assertions come back; no telemetry read, logger spy, or payload assertion is introduced anywhere.

Issue #32402, opened 2026-09-07
refactor(platform): enforce zero first-party CSS selectors


e7h4n:
## Current state

- Scope: `turbo/apps/platform` (`@okouai/app`) on `main` at `8154ea36eb0625e909228d2d34e01b57e5286507`.
- The app uses Tailwind and imports the `@okouai/ui` global stylesheet, but it also owns two non-vendored CSS files: `src/views/css/index.css` (2,070 lines) and `src/views/okou-page/sidebar-thread-reorder.css` (71 lines). Excluding the vendored Markdown stylesheet, the audit found 92 distinct class-selector names: 84 first-party selectors and 8 selectors that adapt third-party or generated DOM.
- First-party selectors form a parallel component-styling API beside Tailwind and typed UI variants. Examples include `.okou-card`, `.okou-border*`, `.okou-btn-morandi`, `.okou-badge`, `.okou-chat-*`, `.owf-*`, and the sidebar reorder selectors. The core shared aliases alone have 216 production references across 76 files.
- The parallel API has real cascade ambiguity. Some app-owned selector rules are unlayered and therefore outrank Tailwind utilities. At the audit cutoff, 34 production call sites across 20 files combine `.okou-card` with Tailwind background, border, hover, focus, or shadow utilities that the selector can override.
- Styling class names are also used as runtime or test query hooks in several places, coupling behavior and tests to presentation details.
- `turbo/apps/platform/eslint.config.js` currently excludes `**/*.css`. There is no enforced repository rule that rejects a newly authored first-party class selector, validates a bounded exception list, or gives an actionable design-system alternative.
- #1025 centralized shared CSS variables in `@okouai/ui` and mapped them into Tailwind, but the repository does not yet express token ownership, naming, mapping, component usage, and exception policy as one complete and enforced design-system contract.

## Goal

- Business components in `@okouai/app` have zero app-authored first-party CSS class selectors. Existing first-party selectors and all dependencies on them are reduced to zero, and new ones are rejected.
- Component styling is expressed through Tailwind utilities backed by sanctioned tokens, or through typed variants and primitives owned by `@okouai/ui`. Tailwind-generated utility classes are not considered app-authored selectors.
- Handwritten CSS outside that model is represented by an explicit, machine-readable allowlist limited to:
  - global environment, browser, and runtime rules that are not business-component styling; and
  - narrowly scoped adapters for DOM owned by third-party libraries.
[...]

PR #33701, opened 2026-09-12
refactor(platform): drain sidebar shell legacy style tokens

bingjiez666:
Part of the #32402 legacy CSS class cleanup: the **sidebar structure and desktop/mobile shell** batch.

## Drained tokens

Five `platform-shell` tokens, **18 CSS declarations**, **6 consumption sites**, one production file (`sidebar.tsx`):

| token | declarations | sites |
| --- | --- | --- |
| `okou-mobile-sidebar` | 8 | 1 |
| `okou-mobile-fixed-safe-area` | 2 | 1 |
| `okou-sidebar-header` | 2 | 1 |
| `okou-desktop-titlebar-drag-region` | 5 | 2 |
| `okou-desktop-no-drag` | 1 | 1 |

`okou-nav` was in this batch's list and is **not** drained — see below.

## Replacement contract

Every site lost only its legacy classes; no consumer utility was touched. Verified mechanically by extracting each site's rendered class string (through the real `cn()`) from `origin/main` and from this branch and asserting the removed set is a subset of the legacy tokens.

- `okou-mobile-sidebar` → `isolate before:pointer-events-none before:absolute before:inset-0 before:z-[-1] before:bg-sidebar before:content-[""] [@media(display-mode:standalone)]:before:bottom-[calc(-1*var(--sab))]`
- `okou-mobile-fixed-safe-area` → `[@media(max-width:767px)]:box-border` plus the four `p*-(--sat/--sar/--sab/--sal)` insets
- `okou-sidebar-header` → `pt-1.5 [@media(min-width:768px)]:[[data-desktop-shell]_&]:pt-0`
- `okou-desktop-titlebar-drag-region` → one local `DesktopTitlebarDragRegion` component (not an exported class-name constant)
- `okou-desktop-no-drag` → `[-webkit-app-region:no-drag]`

Two deliberate choices:

1. **px, not rem breakpoints.** `max-md:` compiles to `@media (width < 48rem)`, not the retired `max-width: 767px`. A non-default browser font size moves a rem breakpoint but not the retired one, so the arbitrary `[@media(...)]` variants keep the boundary exactly where it was. Confirmed in the compiled CSS.
2. **`[data-desktop-shell]` by attribute alone.** Qualifying it with `.okou-app` would register a *new* legacy class dependency against the shrink-only baseline. Nothing else in the repo carries that attribute.

## Measurement

Fixture reproduces the real ancestor chain (`html[data-theme][data-gradient-color-themes][data-color-theme] > body > #root > .okou-app > aside`), including the app's verbatim viewport meta, with nav rows that inherit their colour from `okou-nav-copy*` as they do in the product. `before` CSS is `git show origin/main:<css>`, `after` is this branch, both through the App's own Tailwind compiler from one shared candidate set. Hover is forced with `CSS.forcePseudoState` on the row *and* its label, with the `primaryHoverType=2,availableHoverTypes=2` blink settings, since Tailwind wraps `hover:` in `@media (hover: hover)` which headless Chromium reports false.

**180 states** — 9 palettes (default + 8 gradients) x light/dark x 10 scenarios (mobile drawer, safe-area insets, standalone PWA, hover, desktop rail, desktop-shell titlebar, DPR 2, and the 767/768 px breakpoint boundary).

**Result: 0 changed pixels, 0 computed-style mismatches, no dimension changes.**

before/after/diff

## Negative control

**9 of 9 controls detected.** Each perturbs one replacement utility on the `after` side only:

| control | changed px | observation mismatches |
| --- | --- | --- |
| drop `[@media(max-width:767px)]:pt-(--sat)` | 329,145 | 8 |
| **swap** `before:bg-sidebar` → `before:bg-red-500` | 20,044,044 | 20 |
| drop `pt-1.5` | 573,711 | 20 |
| drop drag-region `h-(--okou-desktop-titlebar-height)` | 101,042 | 8 |
| drop `before:bg-sidebar` | 0 | 20 |
| drop `isolate` | 0 | 20 |
| drop standalone `before:bottom-[...]` | 0 | 4 |
| drop shell `pt-0` | 0 | 8 |
| drop `[-webkit-app-region:no-drag]` | 0 | 20 |

The five zero-pixel rows are pixel-neutral *by construction*, not because the harness is blind — and that was verified rather than assumed:

- **`before:bg-sidebar`**: recolouring the same layer moves 20M pixels, so the pixel channel plainly sees it. Removing it is invisible because a negative-`z` `::before` paints above the element's own background, and the drawer carries the identical `bg-sidebar` underneath.
- **`isolate`**: `max-md:fixed` + `max-md:z-40` already make the drawer a stacking context, so `isolation` has no paint consequence in any reachable state. Retained because the retired rule set it.
- **standalone `before:bottom`**: the overhang paints below the emulated viewport, outside any capture.
- **shell `pt-0`**: the expanded header is `display:none` at desktop widths.
- **`-webkit-app-region`**: a native window hint that paints nothing in a browser.

## `okou-nav` is blocked, with measured cost

`okou-nav` stays on all three `aside` elements. Three of its seven declarations use `.okou-nav` purely as a scoping ancestor for tokens **other batches own** — `.okou-app[data-gradient-color-themes] .okou-nav .okou-nav-copy`, the same for `.okou-nav-copy-muted`, and `.okou-app[data-gradient-color-themes] .okou-nav.okou-nav-rail`. Their consumers are in `sidebar-pinned.tsx`, `sidebar-threads.tsx` and `sidebar-upgrade.tsx`, and **`okou-nav-rail` is not mapped to any manifest family at all**.

I ran the removal rather than asserting it was impossible. Draining `okou-nav` across 144 states changed **5,996,649 pixels**: the rail loses `--okou-nav-rail` (`rgb(244,240,232)` → `rgb(228,223,216)`) and nav copy loses `--okou-nav-foreground` / `--okou-nav-muted-foreground`.

- **Default palette — what production renders, since `GradientColorThemes` is `enabled: false` with no staff allowlist — 16/16 states, 0 changed pixels, identical observations.** That also proves the two production-live declarations duplicate values `@theme` and `.okou-app` already supply.
- Each of the 8 gradient palettes: ~749,580 px, 14/16 observation mismatches.

So `okou-nav` drains only with the rest of the `navigation` family, and `okou-nav-rail` needs a family assignment first. **That ordering is a cross-batch decision, so I stopped rather than picking one.** Recorded in `docs/style-migration.md`.

### Why no manifest entry

`style-migration.mjs` asserts `batch.cases.length > 0` against case IDs registered in `caseFiles`, and no registered case covers the sidebar shell. I confirmed by experiment that an entry with an empty `cases` array fails validation. Registering a real case needs a deployed-preview runner with its App/API deployment and private TEST storage state, which this environment has no access to. The `navigation` and `platform-shell` family entries already map every token here, and both style checks pass.

## Checks

From `turbo`, all passing:

- `pnpm lint:style` — exit 0, including the baseline ratchet (pruned via `pnpm lint:style:prune`, no allowance added), `@eslint/css`, and the Tailwind unknown-class layer
- `pnpm style:migration:inventory` — manifest valid
- `pnpm -F @okouai/app check-types` — exit 0
- `pnpm -F @okouai/app lint` — exit 0
- `pnpm knip` — exit 0
- Prettier on all four changed files

No measurement scripts were added to the repo; the harness lives outside the worktree.

### Not yet run

The sidebar Vitest files (`sidebar.test.tsx`, `sidebar-account-menu.test.tsx`, `auth-v2-submission.test.tsx`) have **not** been run locally — this run hit its time limit first. They should be green: no test touches the five drained tokens, and the two that query `aside.okou-nav:not(.okou-nav-rail)` are unaffected because `okou-nav` is retained. Relying on the PR pipeline to confirm.

Bounded Chromium evidence on a reconstructed fixture, not the deployed preview: it does not certify real native Desktop or PWA surfaces, which `docs/style-migration.md` already excludes.

Refs #32402

### DIFF
--- a/docs/style-migration.md
+++ b/docs/style-migration.md
@@ -674,3 +674,64 @@ observations match in every capture. Both archives were anonymously downloaded
 and SHA-256 verified. This is bounded Chromium acceptance with controlled
 bootstrap, API and browser-media boundaries; no Agent run, purchase, connector,
 real microphone recording or real transcription occurred.
+
+## Sidebar structure and desktop/mobile shell batch
+
+This batch drains five `platform-shell` tokens from the sidebar: the mobile
+drawer's isolation and painted surface (`okou-mobile-sidebar`), its safe-area
+boundary (`okou-mobile-fixed-safe-area`), the expanded header's top inset
+(`okou-sidebar-header`), and the two native-shell window hints
+(`okou-desktop-titlebar-drag-region`, `okou-desktop-no-drag`). Eighteen
+declarations and six consumption sites in `sidebar.tsx` are removed; the two
+drag-region call sites become one local `DesktopTitlebarDragRegion` component
+rather than a shared class-name constant.
+
+Three retired conditions are reproduced in px rather than through Tailwind's
+`md:` / `max-md:` variants, which compile to `48rem`. A reader who changes the
+browser default font size moves a rem breakpoint but not the retired `767px` and
+`768px` ones, so the arbitrary `[@media(max-width:767px)]` and
+`[@media(min-width:768px)]` variants keep the boundary where it is. The
+`[data-desktop-shell]` ancestor is matched by attribute alone: qualifying it
+with `.okou-app` would register a new legacy class dependency against the
+shrink-only baseline.
+
+`-webkit-app-region` is a native window hint with no paint, so it is verified by
+computed style rather than pixels. `isolation: isolate` is also pixel-neutral in
+every reachable state, because `max-md:fixed` with `max-md:z-40` already makes
+the drawer a stacking context; it is retained because the retired rule set it.
+
+### okou-nav remains blocked
+
+`okou-nav` is in this batch's token list but is **not** drained, and the class
+stays on all three `aside` elements. Three of its seven declarations use
+`.okou-nav` only as a scoping ancestor for tokens this batch does not own:
+`.okou-app[data-gradient-color-themes] .okou-nav .okou-nav-copy`, the same rule
+for `.okou-nav-copy-muted`, and
+`.okou-app[data-gradient-color-themes] .okou-nav.okou-nav-rail`. Their consumers
+live in `sidebar-pinned.tsx`, `sidebar-threads.tsx` and `sidebar-upgrade.tsx`,
+and `okou-nav-rail` is not mapped to any manifest family at all. Removing the
+class would silently delete those three rules.
+
+The cost was measured rather than asserted. Draining `okou-nav` from the drawer
+and the rail across 144 states changed 5,996,649 pixels: the rail background
+falls off its `--okou-nav-rail` value and the nav copy loses
+`--okou-nav-foreground` / `--okou-nav-muted-foreground`. That damage is confined
+to the eight gradient palettes at about 749,580 pixels each. On the default
+palette — which is what production renders, because `GradientColorThemes` is
+`enabled: false` with no staff allowlist — all 16 states changed zero pixels
+with identical observations, confirming that the two production-live
+declarations (`.okou-nav { --color-sidebar-border }` and
+`:where([data-theme="dark"]) .okou-nav { --color-sidebar }`) duplicate values the
+`@theme` block and `.okou-app` already supply.
+
+So `okou-nav` drains only together with the rest of the `navigation` family, and
+`okou-nav-rail` needs a family assignment first. Ordering that work is a
+cross-batch decision, not a value this batch may choose.
+
+No batch entry was added to `turbo/style-migration-manifest.json`. Its schema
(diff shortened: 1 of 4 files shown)


bingjiez666:
### Coarse-pointer (touch) pass

Re-ran the full 180-state matrix **without** the `primaryHoverType=2,availableHoverTypes=2,primaryPointerType=4,availablePointerTypes=4` blink settings, i.e. with headless Chromium reporting `(hover: hover)` as false, so every `hover:` / `group-hover:` utility is inert:

**180 states, 0 changed pixels, 0 computed-style mismatches** — same as the hover-forced pass.

That is expected here: none of the five drained tokens declared a hover state. It is reported because the two passes exercise different generated CSS, and a batch that did touch hover could pass one and fail the other.

Issue #33705, opened 2026-09-12
refactor(runner): teach the mitm addon both connector-intent header names (Part A step 1, expand)


lancy:
Part A of #33661, **step 1 of three**. Teach the mitm addon to recognise and strip **both** `x-vm0-connector-intent` and `x-okou-connector-intent`, and add the new name to the API strip-list while keeping the old one.

**Do not touch the CLI in this step.** The entire point is that the stripper learns the new name *before* any sender uses it.

## Why three steps

`X-VM0-Connector-Intent` is set by the CLI at `turbo/apps/cli/src/commands/mcp/client.ts:20,112` and consumed by the addon at `crates/runner/mitm-addon/src/connector_intent.py`, which matches the exact literal `b"x-vm0-connector-intent"` and whose contract is *"Capture connector intent once and always remove its private header."*

Rename both ends at once and a new addon stops recognising an older CLI's header — so instead of being removed, **it is forwarded upstream to the third-party service**, carrying a `connectorId`. That is a header leak, not lost attribution. The API strip-list entry at `custom-connector.service.ts:133` exists as defence in depth against the same thing.

Order: **stripper learns both → CLI switches → stripper drops the old**. Steps are separate releases.

### The drain gate, confirmed

`docs/deployment-compatibility.md`, *Commit-addressed CLI artifacts*: the CLI is an immutable, commit-addressed package, and the backend records its URL in `CLI_PKG_URL` when it creates run execution context; `crates/guest-agent/src/cli/mod.rs:97` reads it and fetches at run time inside the sandbox.

So the CLI is **neither pinned into the runner image nor supplied by the user** — it is drainable. The documented gate for step 3 is: deploy a backend supporting both variants that starts selecting the canonical package, wait through **maximum queue lifetime plus maximum claimed execution and finalization lifetime**, confirm no pre-deployment context can still use the old variant, then remove.

That gate belongs to step 3. It is recorded here so step 1 does not invent a different one.

## Scope of this step

1. `crates/runner/mitm-addon/src/connector_intent.py` — accept and strip both names, preferring the new one when both are present.
2. `turbo/apps/api/src/signals/services/custom-connector.service.ts:133` — add `"x-okou-connector-intent"`, **keep** `"x-vm0-connector-intent"`.
3. Tests covering both names.

[...]

Issue #33661, opened 2026-09-12
refactor: rename the two cross-version x-vm0 headers (connector-intent, web-origin)


lancy:
The last two retired-brand `x-vm0-*` headers. Both were deliberately excluded from #33659 because neither sender and receiver ship together.

**Blocked on #33658.** Do not start until the `restrictedBuiltInModels` expand step is merged, released, and verified. Two concurrent expand/contract rollouts on overlapping surfaces is not worth the coordination cost, and neither of these is urgent.

The two parts are independent and should be two PRs. Part B needs an investigation before its shape can even be chosen.

---

## Part A — `x-vm0-connector-intent`

Set by the **published CLI** at `turbo/apps/cli/src/commands/mcp/client.ts:20,112`:

```ts
const CONNECTOR_INTENT_HEADER = "X-VM0-Connector-Intent";
...
headers.set(CONNECTOR_INTENT_HEADER, connectorId);
```

Consumed by the mitm-addon at `crates/runner/mitm-addon/src/connector_intent.py:8,10`, matching the exact literal `b"x-vm0-connector-intent"`, with the docstring *"Capture connector intent once and always remove its private header."* The API keeps a defence-in-depth strip-list entry at `custom-connector.service.ts:133`.

### The failure mode is a header leak, not lost attribution

A one-step rename produces a new addon that does not recognise the old name. A request from an older CLI then keeps the header, and because the addon only strips names it recognises, **the header is forwarded upstream to the third-party service instead of being removed**. It carries a `connectorId`. The existing API strip-list is evidence that someone already reasoned about this leak.

This is worse than an ordinary skew window because the CLI is **installed on user machines** and versioned independently (`cli-v9.323.16`). Old CLIs persist indefinitely; there is no deploy that retires them.

### Required shape

1. **Expand** — the addon recognises and strips **both** `x-vm0-connector-intent` and `x-okou-connector-intent`, preferring the new name when both are present. The API strip-list gains the new name and keeps the old. **Ship the addon side first and let it reach the fleet.** Do not change the CLI in this step.
2. **Migrate** — the CLI sends only the new name. Ship.
3. **Contract** — remove old-name handling from the addon and the strip-list. Gate this on evidence that old CLIs are no longer sending it, not on elapsed time.

Steps 1 and 2 must not be collapsed: the whole point is that the stripper learns the new name **before** any sender uses it.

### Before starting, answer this

[...]

PR #33761, opened 2026-09-13
refactor(cli): send the canonical connector-intent header name

### LINKED ISSUE #33760: refactor(cli): send the canonical connector-intent header name (Part A step 2, migrate)
Part A of #33661, **step 2 of three**. Switch the CLI to send the canonical `X-Okou-Connector-Intent`.

**Do not remove any legacy handling.** That is step 3, and it is still gated.

## Precondition, measured and met

Step 1 (#33705) shipped in `runner-rs-v0.191.2` at 2026-09-13T01:22Z, teaching the addon to recognise and strip **both** names.

**Drain A is complete**, verified at 05:27Z on `vm0-request-log-prod`: in the trailing 2 hours, every `Runner` and `MitmAddon` request reported `0.191.2`, across 9,516 requests, and widening to 5 hours shows `0.191.1` last seen at **03:18:12** — silent for over two hours while `0.191.2` stayed continuously active. Three-day retention is the whole available window.

That matters because the addon ships inside the runner binary. Had the CLI switched while older runners were live, those addons would not have recognised the new name, would not have stripped it, and it would have been **forwarded upstream to the third-party service with a `connectorId` in it** — the same leak as the original hazard, mirrored. That risk is now retired.

## Scope

1. `turbo/apps/cli/src/commands/mcp/client.ts:20` — `CONNECTOR_INTENT_HEADER` becomes `"X-Okou-Connector-Intent"`.
2. `turbo/apps/cli/src/commands/mcp/__tests__/index.test.ts:77` — the stub reads the header the CLI sends; it must read the canonical name.

That is the whole change. Two lines.

### 1. The stub's header name and the assertions are the same fact — fix the stub, not the expectations

`stubMcpServer` records what the CLI actually sent:

```ts
const intent = request.headers.get("x-vm0-connector-intent");
```

and two tests assert on it:

```ts
expect(seen.every((request) => request.intent === CONNECTOR_ID)).toBe(true);
```

Change `client.ts` alone and every recorded `intent` becomes `null`, so those `.every()` checks fail. **That failure is correct and informative — it is the test noticing the sender changed.**

The fix is the stub's header name at line 77. It is **not** to change the expected value to `null`, and **not** to relax `.every()` to `.some()`. Either of those converts a test that proves *every MCP request carries the connector intent* into one that proves nothing, while showing green.

Note also that `.every()` is vacuously true on an empty array. That is a pre-existing property, not something this change introduces, but it means a "fix" that empties `seen` would also pass. If you touch anything that affects how `seen` is populated, say so explicitly in the PR body.

### 2. Everything on the receiving side stays exactly as it is

After this step the addon must still accept **both** names, and the API strip-list must still contain **both**. Specifically, leave untouched:

- `crates/runner/mitm-addon/src/connector_intent.py` — `LEGACY_HEADER_NAME`, `_RAW_LEGACY_HEADER_NAME`, the fallback scan, and both `set_all` strip calls.
- `turbo/apps/api/src/signals/services/custom-connector.service.ts:134-138` — the `x-vm0-connector-intent` entry and its explanatory comment.
- `turbo/apps/api/src/signals/routes/__tests__/connectors.bdd.test.ts:6663` — the assertion that the strip-list protects the legacy name.

The reason is drain B, which is untouched by this step: execution contexts queued or claimed before this release still resolve an older `CLI_PKG_URL` and will keep sending the legacy name until they finish. Removing tolerance now would strand exactly those runs.

### 3. Do not gate anything on the CLI semantic version

`docs/deployment-compatibility.md` states the package's semantic version may not advance for every artifact change and must not be used as a compatibility floor. If the PR body describes step 3's future condition, it must reference the **package commit**.

## Do not touch

- Anything under `crates/`.
- The completed lines from #33657, #33659, #33661 Part B and #33705 — their results must stay byte-identical.
- `hashtext('vm0')`, `PUBLIC_BRANDS` and the brand `z.enum`s, the 11 `vm0.*` localStorage keys, the 12 locale `repositoriesPlaceholder` strings, and live infrastructure names.
- Do not weaken, delete, or invert any assertion. Do not add a test asserting the legacy name is absent.

## Acceptance

1. `client.ts` sends `X-Okou-Connector-Intent`, and `X-VM0-Connector-Intent` has zero occurrences in `turbo/apps/cli/`.
2. The stub reads the canonical name, and both `.every((request) => request.intent === CONNECTOR_ID)` assertions are **unchanged** and passing.
3. `connector_intent.py` is byte-identical.
4. `custom-connector.service.ts` still lists both names; `connectors.bdd.test.ts:6663` still asserts the legacy one is protected.
5. No CLI semantic version used as a gate.
6. No assertion weakened; no absence-asserting test added.

lancy:
Part A **step 2 of three** of #33661, implementing #33760. The CLI now sends the canonical `X-Okou-Connector-Intent`.

**No legacy handling is removed here.** That is step 3, and it stays gated.

## The change, in full

| File | Change |
| --- | --- |
| `turbo/apps/cli/src/commands/mcp/client.ts:20` | `CONNECTOR_INTENT_HEADER` → `"X-Okou-Connector-Intent"` |
| `turbo/apps/cli/src/commands/mcp/__tests__/index.test.ts:77` | `stubMcpServer` reads `x-okou-connector-intent` |

Two lines. Nothing else in the diff.

## Why the sender can move now

The mitm addon ships inside the runner binary, so switching the CLI while older runners were live would have meant those addons did not recognise the new name, did not strip it, and **forwarded it upstream to the third-party service carrying a `connectorId`** — the original hazard, mirrored.

Step 1 (#33705) shipped in `runner-rs-v0.191.2` and taught the addon to recognise and strip **both** names. Drain A is complete: on `vm0-request-log-prod`, every `Runner` and `MitmAddon` request in the trailing 2 hours reported `0.191.2` across 9,516 requests, and widening to 5 hours puts the last `0.191.1` at `03:18:12`, silent for over two hours while `0.191.2` stayed continuously active. Three-day retention is the whole available window. That risk is retired.

## Why the receiving side is untouched

Drain B is not addressed by this PR. Execution contexts queued or claimed before this release still resolve an older `CLI_PKG_URL` and will keep sending the legacy name until they finish. Removing tolerance now would strand exactly those runs. So these stay exactly as they are:

- `crates/runner/mitm-addon/src/connector_intent.py` — `LEGACY_HEADER_NAME`, `_RAW_LEGACY_HEADER_NAME`, the fallback scan, and both `set_all` strip calls. Byte-identical; nothing under `crates/` is in this diff.
- `turbo/apps/api/src/signals/services/custom-connector.service.ts:134-138` — both names remain in the strip-list, with the explanatory comment.
- `turbo/apps/api/src/signals/routes/__tests__/connectors.bdd.test.ts:6663` — the assertion that the strip-list protects the legacy name is unchanged.

**Step 3's condition is a package commit, not a version string.** `docs/deployment-compatibility.md` states the CLI package's semantic version may not advance for every artifact change and must not be used as a compatibility floor. Step 3 is gated on evidence that no execution context still resolves a `CLI_PKG_URL` predating this PR's merge commit, not on elapsed time and not on a `cli-v*` tag.

## The test change is the sender change

`stubMcpServer` records what the CLI actually sent. Changing `client.ts` alone makes every recorded `intent` `null`, and the two `.every()` checks fail — correctly, because the test is noticing the sender moved. The fix is the stub's header name, not the expectation.

**Both assertions are unchanged and passing:**

```ts
expect(
  seen.every((request) => {
    return request.intent === CONNECTOR_ID;
  }),
).toBe(true);
```

Nothing affecting how `seen` is populated was touched — the only edit to the test file is the header name string on line 77.

`.every()` is vacuously true on an empty array, which is pre-existing and not introduced here. It does not apply to either assertion, and that is enforced by neighbouring assertions rather than by inspection: the first is immediately preceded by `expect(seen.map(...)).toEqual(["server/discover", "tools/list", "tools/list"])`, and the second by `expect(calls).toHaveLength(1)` over a filter of `seen`. Both pin `seen` to a non-empty length.

## Verification

Targeted CLI test, `turbo/apps/cli`:

```
npx vitest run src/commands/mcp/__tests__/index.test.ts
Test Files  1 passed (1)
     Tests  40 passed (40)
```

The `error: option '--input <json>' cannot be used with option '--input-file <path>'` line in that output is a test exercising that validation path, not a failure.

**Negative control.** To show the assertions are load-bearing rather than green-by-luck, I temporarily reverted only the stub's header name, leaving `client.ts` canonical, and re-ran: **8 failed | 32 passed**, including both `.every()` assertions at lines 359 and 421. Restored before committing; the committed diff is the two lines above.

Prettier reports both files clean. Per the issue I did not run the full suite and did not start a dev server; nothing under `crates/` changed, so no Rust or addon build was required. CI covers the rest.

## Acceptance

1. ✅ `client.ts` sends `X-Okou-Connector-Intent`; `X-VM0-Connector-Intent` has zero occurrences in `turbo/apps/cli/`.
2. ✅ The stub reads the canonical name; both `.every((request) => request.intent === CONNECTOR_ID)` assertions are unchanged and passing.
3. ✅ `connector_intent.py` is byte-identical.
4. ✅ `custom-connector.service.ts` still lists both names; `connectors.bdd.test.ts:6663` still asserts the legacy one is protected.
5. ✅ No CLI semantic version used as a gate.
6. ✅ No assertion weakened, deleted, or inverted; no absence-asserting test added.

Closes #33760


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: 7d0ba331e0b5fb4f5151bdee049c97113ccde84f
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `7d0ba331e0b5fb4f5151bdee049c97113ccde84f`. Practice revision: `ee3b08bb774b020f45efc9131929d2ac5d3afbb0`.

### Summary

Part A step 2 of #33661: the CLI's `CONNECTOR_INTENT_HEADER` becomes `X-Okou-Connector-Intent`, and the MSW stub that records what the CLI sent reads the matching canonical name. Two lines, both in `turbo/apps/cli/src/commands/mcp/`; no receiving-side handling is removed, so the addon and the API strip-list still accept both names for drain B.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | Sole sender is `client.ts:112`. Repo-wide search for the header name returns only `client.ts:20,112`, the stub at `index.test.ts:77`, the API strip-list, the BDD strip-list assertion, and changelogs. `request_classification.py:469` and `firewall-rule-matcher.ts:1700-1702` consume intent *status* via `from_flow`, never a header literal, so they are name-agnostic. |
| `REVIEW.md` | Confirm existing coverage still protects a refactor | Pass | Both `.every((request) => request.intent === CONNECTOR_ID)` assertions (`index.test.ts:358-361`, `418-421`) are outside the diff and still pass. Neither is vacuous: each is preceded by an assertion pinning `seen` non-empty — `toEqual(["server/discover", "tools/list", "tools/list"])` and `expect(calls).toHaveLength(1)`. |
| `REVIEW.md` | Tests verify externally observable state, not internal state or mock calls | Pass | The stub asserts on the wire header of the real outbound request, which is the externally observable contract; nothing internal to `createMcpFetch` is inspected. |
| `REVIEW.md` | No weakened, deleted, or inverted assertions; no tombstone negative test | Pass | The diff contains no `every`, no expectation edit, and no new test. Only the header-name string on line 77 changed. |
| `REVIEW.md` | Requests/protocols must cover old/new combinations coexisting during rollout | Pass | See the `deployment-compatibility.md` rows: both directions remain served, and the receiving-side combinations are covered by the step-1 addon tests. |
| `REVIEW.md` / `docs/fallback.md` §9 | Reviewer lists every fallback the diff introduces | N/A | The diff introduces no fallback. It neither adds nor removes the legacy tolerance, which lives in `connector_intent.py` and `custom-connector.service.ts`, both outside this diff. `fallback.md` §9 explicitly exempts keep-only PRs from a `Fallbacks` section. |
| `docs/fallback.md` §1 | No negative test asserting removed behavior stays removed | Pass | No test was added. |
| `docs/fallback.md` §5, §7 | A retired-protocol reader is deleted, or an explicit rollout gate is named | Pass | The legacy reader is deliberately retained, and its gate is still open. `connector_intent.py:66-74` already names the surface (existing runner/sandbox), the removal condition (no queued or claimed context can resolve a pre-change `CLI_PKG_URL`), and the follow-up (#33661 step 3). Correctly untouched here. |
| `docs/deployment-compatibility.md` — Commit-addressed CLI artifacts | The package commit is the release identity; semver must not be a compatibility floor | Pass | The PR body states step 3's condition as a `CLI_PKG_URL` predating this PR's merge commit and explicitly rejects a `cli-v*` tag as a gate. No semver appears in the diff. |
| `docs/deployment-compatibility.md` | Receiver supports both variants before the sender switches | Pass | Step 1 is on `main`: `_RAW_HEADER_NAME = b"x-okou-connector-intent"` is lowercase, matching the `name.lower()` field scan, so the newly sent mixed-case name is recognised. Both `set_all(..., [])` calls sit outside the capture guard (`connector_intent.py:93-94`), so both names are stripped unconditionally on every flow. |
| `docs/deployment-compatibility.md` | Old-sender direction stays served for its full drain | Pass | Pre-release execution contexts keep an older `CLI_PKG_URL` and keep sending the legacy name. `LEGACY_HEADER_NAME`, `_RAW_LEGACY_HEADER_NAME`, the fallback scan, both strip calls, and both strip-list entries are byte-identical; `git diff origin/main --name-only` returns only the two CLI files. |
| `docs/testing.md` | CLI boundary: enter through the command parser and assert observable output | Pass | Unchanged; the suite drives `mcpCommand.parseAsync` and observes the outbound HTTP request. |
| `docs/testing.md` | Request-contract changes cover rollout old/new combinations | Pass | The CLI emits exactly one name, so it has no old/new combination of its own to cover. The coexistence matrix belongs to the receiver and is covered by step 1: `test_request_handler_firewall_dispatch.py:193-205` exercises canonical-only, legacy-only, duplicate-within-a-name, and one-of-each precedence. |
| `docs/testing/cli-testing.md` | Mock only external boundaries; never mock internal CLI modules | Pass | The edit is inside the MSW handler, an external HTTP boundary. No internal module, validator, or API config is mocked. |
| `docs/testing/cli-testing.md` | Tests live in a neighbouring `__tests__/` with a matching name | Pass | Unchanged: `src/commands/mcp/__tests__/index.test.ts` alongside `src/commands/mcp/index.ts`. |
| `docs/bad-smell.md` §4 | Document changed public contracts and assess their consumers | Pass | The PR body states the contract change, the drain-A precondition that permits it, and the drain-B reason the receiver keeps both names. |
| `docs/bad-smell.md` §1, §2 | No `any`; no lint/type suppressions | Pass | Two string-literal changes; no type or suppression edits. `lint-types`, `lint-type-api`, `lint-format`, and `lint-knip` are green on this HEAD. |
| `docs/bad-smell.md` §6 | Environment-specific values go through centralized `env()` | N/A | This is a fixed wire-protocol header name shared with the addon, not an environment-specific value or service origin. |
| `docs/bad-smell.md` §7 | No fallback for a state the owning contract already rules out | Pass | No branch added. |
| `docs/bad-smell.md` §3, §5 | Error-handling boundaries; static imports | N/A | No error path, catch block, or import changed. |

### Findings

None. No P0 or P1 issues.

Non-blocking observation, recorded rather than raised as a defect: the safety of the sender switch rests on drain A, whose evidence is author-provided from `vm0-request-log-prod` (all `Runner`/`MitmAddon` requests at `0.191.2` across 9,516 requests in the trailing 2 hours; last `0.191.1` at 03:18:12). That dataset's three-day retention is the whole available window, so the drain cannot be re-derived beyond it and I did not independently re-run the query in this review. The residual case it excludes is a runner that was offline throughout the window and returns still running ≤ `0.191.1`; such an addon would not recognise the canonical name and would forward it upstream. That is the inherent shape of the expand/migrate rollout and the exact risk step 1 plus drain A exists to retire — not a defect in this diff.

### Testing

`test-cli` is green on this HEAD, which is the directly relevant required check for a diff confined to `turbo/apps/cli/`. All `crates/` jobs (`mitm-addon-test`, `runner-*`, `ci-gate-crates`) are correctly skipped, matching a diff that touches nothing under `crates/`.

Coverage is adequate and unchanged in strength. The stub's header name and the two `.every()` assertions are one fact: had only `client.ts` moved, every recorded `intent` would be `null` and both assertions would fail. The author reports the correct repair — the stub name — and reports a negative control confirming it: temporarily reverting only the stub while leaving `client.ts` canonical produced 8 failed / 32 passed, including both `.every()` assertions. That is the right way to show the assertions are load-bearing rather than green by luck, and the committed diff does not contain the control.

No new test is required. The CLI's observable contract is "every MCP request carries the connector intent under the canonical name", and that is exactly what the retained assertions prove. Nothing affecting how `seen` is populated was touched, so the `.every()`-on-empty-array concern noted in #33760 does not apply here, and it is independently excluded by the neighbouring length assertions.

No material gap. A test asserting the legacy name is *absent* would be a `fallback.md` §1 tombstone and is correctly not present.

PR #33707, opened 2026-09-12
refactor(runner): accept both connector-intent header names

lancy:
Closes #33705. Part A of #33661, **step 1 of three (expand)**.

The mitm addon matched one literal, `b"x-vm0-connector-intent"`, and removes that header on every flow so a `connectorId` never reaches the third-party upstream. The CLI that sends it is an independently released, commit-addressed package, so a sandbox can execute an artifact built before any rename. Renaming both ends at once would leave the older name unrecognised — and because the addon only removes names it recognises, that header would be **forwarded upstream** rather than lost. This step teaches the stripper the canonical name before any sender uses it.

## What changed

- `crates/runner/mitm-addon/src/connector_intent.py` — capture accepts `x-okou-connector-intent` and `x-vm0-connector-intent`; removal covers **both** names, unconditionally, on every flow.
- `turbo/apps/api/src/signals/services/custom-connector.service.ts` — the MCP protected-header list gains `x-okou-connector-intent` and **keeps** `x-vm0-connector-intent`.
- Tests for both names, including the stated precedence decision.

`turbo/apps/cli/src/commands/mcp/client.ts` is untouched: after this PR the CLI still sends only the legacy name, and both names must work. Switching the sender is step 2, a separate release. `matching.py` and `connector_diagnostics.py` routing behaviour is unchanged — this is header recognition only.

## Both names present: the explicit decision

Precedence is per name, and it is a deliberate choice rather than a by-product of the loop:

- **The canonical name is authoritative whenever it appears at all.** The legacy name is read only when the canonical one is absent entirely.
- **Repetition of a single name stays `malformed`**, exactly as before. One occurrence of each name is *not* repetition, so that case is `present` with the canonical value — per this rollout the CLI never sends both.
- A canonical name repeated twice alongside one legacy name is `malformed`, because the canonical name is authoritative including when it is malformed. Covered by `test_repeated_canonical_name_stays_malformed_beside_one_legacy_name`.

## Removal stays unconditional

`capture_and_strip` ends with `set_all(...)` for both names on every flow — including flows carrying neither header and flows whose metadata was already captured by an earlier hook. Stripping only the name that was found would reintroduce the leak this three-step shape exists to prevent. `test_already_captured_flow_still_removes_both_names` fails if either `set_all` is dropped.

## Verification

Run locally with a pip-installed `mitmproxy==12.2.3` (this environment has no `uv`, so the locked addon environment was not used — CI is authoritative):

[...]

PR #33767, opened 2026-09-13
refactor(runner): remove legacy connector-intent header tolerance

### LINKED ISSUE #33763: refactor(runner): remove legacy connector-intent header tolerance (Part A step 3, contract)
Part A of #33661, **step 3 of three** — the contract step. Remove the legacy `x-vm0-connector-intent` tolerance from the mitm addon and the API strip-list.

## Preconditions

**Drain A — complete.** Verified 2026-09-13T05:27Z: every live `Runner` and `MitmAddon` reported `0.191.2` or higher across 9,516 requests; `0.191.1` last seen 03:18:12Z.

**Drain B — bounded and computed.** Step 2 promoted at **2026-09-13T15:02Z**. Every execution context created before then is terminal by **≈17:19Z**, from constants rather than estimates:

| Term | Value | Source |
|---|---|---|
| Queue TTL | 15 min | `run-queue.service.ts:51` `PENDING_RUN_TTL_MS` |
| Execution budget | 2 h | `runners.ts:75` → `executor/mod.rs:84` `JOB_TIMEOUT` |
| Finalization grace | 90 s | `executor/mod.rs:91` |
| Terminal grace | 10 s | `JOB_TERMINAL_GRACE_TIMEOUT` |

The last three compose into the runner-enforced `job_terminal_wait_timeout()`. `CLI_PKG_URL` is snapshotted at run creation (`agent-run-create.service.ts:6450-6461`), and the resume path cannot carry an old value forward because that function always overwrites the key with the current environment value.

**This issue may be implemented and merged now. It must not be released before 17:19Z.** The controller owns that gate.

## Scope

1. `crates/runner/mitm-addon/src/connector_intent.py` — remove `LEGACY_HEADER_NAME`, `_RAW_LEGACY_HEADER_NAME`, the legacy fallback in `capture_and_strip`, the second `set_all`, and the bounded-fallback paragraph in the docstring.
2. `turbo/apps/api/src/signals/services/custom-connector.service.ts` — remove the `"x-vm0-connector-intent"` strip-list entry and its explanatory comment.
3. The nine addon test files and `connectors.bdd.test.ts` that reference the legacy name.

Do **not** remove `_scan_raw_header`. It was extracted in step 1 and is still used by the canonical scan; deleting it would inline logic back for no reason.

## The decision this step makes, which must be stated in the PR body

After this, the addon **no longer strips** `x-vm0-connector-intent`. A caller that sends it — no legitimate CLI will, but a user's own code inside the sandbox could — will have it **forwarded to the third-party upstream** instead of removed.

That is the correct end state: the header is our private protocol header, and once it is no longer ours, silently deleting an arbitrary caller-supplied header is not the addon's job. But it **is** a behaviour change, and the PR body must say so in one sentence rather than presenting this as pure deletion.

If you disagree with that reading after looking at the code, say so and stop rather than choosing silently.

### 1. The canonical strip must stay unconditional

`capture_and_strip` currently ends with two unconditional `set_all` calls outside the capture branch. Removing the legacy line is the point of this change. **Removing the canonical one, or moving it inside the `if`, is the original leak.** After this step exactly one `set_all` must remain, still outside the branch, still running on every flow including flows carrying no intent header.

### 2. The `repeated` semantics revert to their pre-step-1 meaning, and that must stay true

With one name, "repeated" is again just "the same name twice → `malformed`". Step 1 added an explicit decision about one-of-each-name; that case disappears with the legacy name. Confirm the remaining behaviour matches the original: duplicates malformed, empty or comma-containing values malformed, oversized values malformed without decoding.

`test_connector_intent_header_names.py` was created in step 1 specifically to cover both names. Do not delete it wholesale — the canonical-name cases in it are still meaningful. Remove only the cases that test legacy acceptance, and say in the PR body how many of its 20 cases survived and why.

### 3. Ordering inside this step

The addon and the API strip-list ship on different release lines. Removing the API entry while an older addon still accepts the legacy name is harmless, but removing the addon's strip while the API entry is gone is not — the API entry is defence in depth for exactly that window. Ship them together in one PR, and if CI forces them apart, keep the API entry until last.

## Do not touch

- `turbo/apps/cli/src/commands/mcp/client.ts` — already canonical; nothing to do.
- The completed results of #33657, #33659, #33661 Part B, #33705 and #33760.
- `hashtext('vm0')`, `PUBLIC_BRANDS` and the brand `z.enum`s, the 11 `vm0.*` localStorage keys, the 12 locale `repositoriesPlaceholder` strings, and live infrastructure names.
- Do not weaken, delete, or invert any assertion beyond removing cases that exclusively test legacy acceptance. Do not add a test asserting the legacy name is absent.

## Acceptance

1. `x-vm0-connector-intent` and `X-VM0-Connector-Intent` return zero hits in `crates/` and `turbo/`, excluding changelogs.
2. Exactly one `set_all` remains in `capture_and_strip`, outside the capture branch.
3. `_scan_raw_header` still exists and is used.
4. The surviving cases in `test_connector_intent_header_names.py` are counted and justified in the PR body.
5. The forwarding-behaviour change is stated in the PR body.
6. No assertion weakened beyond legacy-only cases; no absence-asserting test added.

lancy:
Closes #33763. Part A of #33661, **step 3 of three** — the contract step.

Step 1 (#33707) taught the mitm addon both `x-vm0-connector-intent` and `x-okou-connector-intent`. Step 2 (#33761) switched the CLI to send only the canonical name. This PR removes the legacy tolerance from both the addon and the API strip-list.

## Behaviour change — the addon no longer strips `x-vm0-connector-intent`

A caller that still sends `x-vm0-connector-intent` — no legitimate CLI does, but a user's own code inside the sandbox could — will now have that header **forwarded to the third-party upstream** instead of removed.

This is the correct end state rather than an oversight: the header was ours only while it was our private protocol header. Once it is not, silently deleting an arbitrary caller-supplied header is not the proxy's job, and the leak this three-step shape existed to prevent — a platform-injected `connectorId` reaching a third party — is closed by the canonical strip, which stays unconditional. It is still a behaviour change, not a pure deletion, so it is recorded here rather than left to the diff.

## Removal evidence

Per `docs/fallback.md` §10, the gate is "existing runner / sandbox — old instances finish draining", and both halves are closed:

- **Drain A** (old addons): verified 2026-09-13T05:27Z on `vm0-request-log-prod` — every live `Runner` and `MitmAddon` reported `0.191.2` or higher across 9,516 requests, with `0.191.1` last seen 03:18:12Z.
- **Drain B** (old `CLI_PKG_URL` contexts): bounded by constants rather than estimate — queue TTL 15 min (`run-queue.service.ts` `PENDING_RUN_TTL_MS`) + execution budget 2 h (`executor/mod.rs` `JOB_TIMEOUT`) + finalization grace 90 s + terminal grace 10 s = 2 h 16 min 40 s. Step 2 promoted at 15:02Z, so every pre-switch execution context is terminal by ≈17:19Z.

Both drain measurements are the controller's; they are restated here as the gate this removal depends on, not re-verified by me. **This PR may merge before 17:19Z but must not be released before it.** The release gate is owned by the controller.

No CLI semantic version is used as a floor.

## Changes

1. `connector_intent.py` — drops `LEGACY_HEADER_NAME`, `_RAW_LEGACY_HEADER_NAME`, the fallback scan, the second `set_all`, and the bounded-fallback docstring paragraph. `_scan_raw_header` stays; the canonical scan still uses it.
2. `custom-connector.service.ts` — drops the `"x-vm0-connector-intent"` protected-header entry and its explanatory comment, restoring the list to one connector-intent entry.
3. Eight addon test files plus `connectors.bdd.test.ts`.

**The canonical strip stays unconditional.** Exactly one `set_all` remains in `capture_and_strip`, still outside the capture branch, so it runs on every flow — including flows carrying no intent header and flows whose intent was already captured.

**`repeated` reverts to its pre-step-1 meaning.** With one name, "repeated" is again just the same name twice. The step-1 one-of-each-name decision disappears with the legacy name. The rest is unchanged and still covered: duplicates malformed, empty or comma-containing values malformed, oversized values malformed without decoding.

## `test_connector_intent_header_names.py` — 10 of 20 cases survive

This file was created in step 1 to cover both names, so it is not deleted wholesale; only the cases that exclusively test legacy acceptance are removed.

| Case | Before | After | Why |
|---|---|---|---|
| Name captured and removed (case-insensitive) | 6 | 3 | The 3 legacy spellings tested legacy acceptance; the 3 canonical spellings still prove the field scan is case-insensitive. |
| Canonical name wins when both appear once each | 1 | 0 | The step-1 both-names decision; it has no meaning with one name. |
| Repeating a single name is malformed | 2 | 1 | The canonical case carries the whole rule now. |
| Repeated canonical beside one legacy name | 1 | 0 | Distinguished only by the legacy header; reduces to the case above. |
| Unusable value is malformed (empty/blank/list) | 6 | 3 | Value parsing is name-independent; the canonical cases keep all three shapes. |
| Oversized value is malformed without decoding | 2 | 1 | The decode guard is name-independent. |
| Already-captured flow still removes the name | 1 | 1 | Kept, narrowed to the canonical name. This is the unconditional-strip guard. |
| Absent intent leaves unrelated headers untouched | 1 | 1 | Unchanged; never referenced a name. |
| **Total** | **20** | **10** | |

No assertion is weakened, deleted, or inverted beyond cases that exclusively tested legacy acceptance, and no test asserting the legacy name is absent was added — `docs/fallback.md` §1 classes those as tombstones.

In the other addon test files the legacy name was incidental — those tests cover catalog removal, firewall dispatch, connector diagnostics, header-name budgets and firewall auth, and merely used the legacy spelling because they predate step 1. They are renamed to the canonical name with every assertion intact. In `test_request_handler_firewall_dispatch.py` three parametrized cases were dropped rather than renamed because step 1 had already added canonical twins for them; every distinct value-shape-to-reason pair still has a case.

## Both sides ship together

The addon and the API strip-list are on different release lines. Removing the API entry early is harmless; removing the addon's strip while the API entry is already gone is not, because that entry is the defence in depth for exactly that window. They are therefore in this one PR.

## Verification

- `pytest tests/` in `crates/runner/mitm-addon` — **7833 passed**, including all eight touched files. Run against a pip-installed environment with `mitmproxy` pinned to the locked `12.2.3`; the remaining dependencies were resolved fresh because `uv` is unavailable in this sandbox, so this is not the locked environment CI uses.
- `ruff check` and `ruff format --check` on the nine changed Python files — clean.
- The API changes were **not** run locally: `connectors.bdd.test.ts` needs a database, and no Vitest run or dev server was started. Those are left to the PR pipeline.

## Note on the issue's file count

#33763 lists "nine addon test files". A case-insensitive scan of `crates/` and `turbo/` for the legacy name finds **eight** addon test files plus `connectors.bdd.test.ts`. All of them are in this diff, and the post-change scan returns zero hits outside changelogs, so the scope is complete; the count in the issue is simply one high.


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: e8cbf0855c058572be69d15a99491a6e0a041906
PR_AUTO_REVIEW_VERDICT: LGTM

Reviewed HEAD: `e8cbf0855c058572be69d15a99491a6e0a041906`. Practice documents read from `main` at `0213a6a`.

### Summary

Removes the legacy `x-vm0-connector-intent` tolerance from the mitm addon and the API MCP protected-header list, completing step 3 of the #33661 Part A rename after step 1 (#33707) taught both names and step 2 (#33761) moved the CLI to the canonical name. The canonical capture and its unconditional strip are unchanged; the diff deletes the fallback branch, its constants, the second `set_all`, the API contract entry, and the tests that existed only to cover the legacy name.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behaviour through its real entry point and consumers | Pass | `mitm_addon.py:774,1349` are the only `capture_and_strip` callers; `request_classification.py:469` and `connector_diagnostics.py:459` consume only `from_flow`. `HEADER_NAME` is not imported by name anywhere outside `connector_intent.py`, so the deleted constants had no other consumer. |
| `REVIEW.md` | Existing coverage still protects the refactor; identify missing scenarios | Pass | Every distinct behaviour retains a case: capture, case-insensitive match, repeated→malformed, empty/blank/comma→malformed, oversized→malformed-without-decoding, already-captured strip, unrelated headers untouched. Full addon suite green. |
| `REVIEW.md` | Use the documented exposure window for the affected surface; do not size it from nominal pipeline duration | Pass | The PR body derives drain B from `PENDING_RUN_TTL_MS`, `JOB_TIMEOUT`, and the finalization/terminal graces rather than an estimate, and drain A is a fleet observation. |
| `REVIEW.md` | Security: auth, permissions, and existence-leak behaviour of the changed path | Pass | `MCP_PROTECTED_HEADER_NAMES` (`custom-connector.service.ts:1212`) still blocks `x-okou-connector-intent`, so a user-defined MCP connector still cannot inject the live routing header. The unprotected legacy name carries only a user-supplied template value to the user's own MCP endpoint and is no longer read by the addon, so it grants no routing or credential capability. |
| `REVIEW.md` | API tests must not assert log output | N/A | The API diff removes one entry from a protected-header list; no logging assertions are touched. |
| `docs/fallback.md` §1 | No negative test that only pins retired behaviour as still absent | Pass | No absence-asserting test was added. `test_request_handler_firewall_dispatch.py:239`, which asserted the legacy name was stripped, is deleted rather than converted into a tombstone. |
| `docs/fallback.md` §5 | Delete the reader in the same PR that retires the producer, or in a follow-up naming the gate | Pass | This is that follow-up; the producer was retired in #33761 and the gate is named in the PR body. |
| `docs/fallback.md` §8 | Removal deletes the tolerant branch, the contract entry, the helper types, and the fallback's tests together | Pass | Addon branch and constants, the API strip-list entry and comment, the `connectors.bdd.test.ts` entry, and the legacy-only cases all ship in one commit. |
| `docs/fallback.md` §10 | Removal carries type, single-writer, production, or rollout evidence | Pass | Rollout evidence: drain A fleet observation plus the constant-derived drain B bound. Single-writer evidence: a repo-wide case-insensitive scan finds no remaining producer of the legacy name. |
| `docs/fallback.md` §9 | New fallbacks declared in a `Fallbacks` PR-body section | N/A | The diff introduces no fallback; §9 explicitly exempts removal-only PRs. |
| `docs/deployment-compatibility.md` | Commit-addressed CLI artifacts: four-step removal procedure | Pass | Steps 1–2 shipped in #33707/#33761; step 3 is the drain evidence in the PR body; this PR is step 4. The body states the release must not precede the drain bound. |
| `docs/deployment-compatibility.md` | The CLI package's semantic version must not be used as a compatibility floor | Pass | No semver gate appears in the code, commit, or PR body; the bound is expressed as constants and an absolute time. |
| `docs/bad-smell.md` §2 | No lint/type suppression comments | Pass | None added; `ruff check` and `basedpyright` are clean on the changed files. |
| `docs/bad-smell.md` §4 | Document changed public contracts and assess their consumers | Pass | The forwarding change is stated explicitly in the PR body rather than presented as a pure deletion, and the docstring keeps the unconditional-strip rationale. |
| `docs/bad-smell.md` §7 | Removal must not delete meaningful recovery or fail-closed security | Pass | The unconditional canonical strip — the actual leak guard — is retained outside the capture branch; only the retired name's handling is removed. |
| `docs/testing.md` | Changes to requests and Runner protocols cover old/new combinations that coexist during rollout | Pass | The coexistence window is closed by the drain rather than by code, which is the documented end state for this surface; the canonical-name path retains full coverage. |
| `docs/testing.md` | Avoid duplicate cases that merely restate configuration or pin incidental implementation | Pass | Three parametrized cases in `test_request_handler_firewall_dispatch.py` were dropped because step 1 had already added canonical twins; every distinct value-shape-to-reason pair still has exactly one case. |
| `docs/testing/testing-external-behavior.md` | Verify externally observable state through the production boundary | Pass | The integration files drive `mitm_addon.request`/`requestheaders` and assert forwarded headers and HTTP responses. `test_connector_intent_header_names.py` calls `capture_and_strip` directly and reads `_STATUS_METADATA_KEY`, but that structure predates this PR (added in #33707) and is unchanged in kind here. |
| `docs/testing/mitm-addon-testing.md` | Run pytest and the Ruff/BasedPyright static checks for addon changes | Pass | Run locally and green; see Testing for the environment caveat. CI's `mitm-addon-test` is the authoritative locked-environment run. |
| `docs/testing/mitm-addon-testing.md` | Run the flow metadata key linter before committing addon changes | Pass | `./scripts/check-flow-metadata-keys.py` exits 0 with no output. No metadata key is added or renamed. |
| `docs/mitm-addon-contracts.md` | Raw-byte header inspection before full-string conversion | Pass | `_scan_raw_header` is retained and still scans `headers.fields` as raw bytes, and the oversized-value path still rejects on `len(raw_value)` before any `decode`, proven by the surviving `_DecodeGuardConnectorIntent` case. |

### Findings

None.

Non-blocking observations, not defects in this diff:

- `test_connector_intent_header_names.py` asserts on `connector_intent._VALUE_METADATA_KEY` and `_STATUS_METADATA_KEY`, which are module-private. This is the structure step 1 introduced for this file and the diff does not deepen it; the production-boundary coverage lives in the request-handler suites.
- The issue text says "nine addon test files"; the actual count is eight plus `connectors.bdd.test.ts`. The PR body records the discrepancy, and a post-change case-insensitive scan of `crates/` and `turbo/` returns zero hits outside changelogs, so the scope is complete.

### Testing

- `python -m pytest tests/` in `crates/runner/mitm-addon`: **7833 passed**. The eight touched files pass together (155 tests), and `test_connector_intent_header_names.py` reports exactly 10 tests, matching the 10-of-20 accounting in the PR body.
- `ruff check` and `ruff format --check` on the nine changed Python files: clean.
- `basedpyright -p .`: 0 errors, 0 warnings, 0 notes.
- `./scripts/check-flow-metadata-keys.py`: exit 0.
- Environment caveat, stated because it affects how much the local run proves: `uv` is unavailable in this sandbox, so the above ran against a pip-installed environment with `mitmproxy` pinned to the locked `12.2.3` and the remaining dependencies resolved fresh. This is not the locked environment; the `mitm-addon-test` and `Analyze (python)` checks own that verification.
- The API side was not exercised locally. `connectors.bdd.test.ts` requires a database, and no Vitest run or dev server was started; `test-api` owns that verification.

Coverage gaps: none identified. The deleted cases covered the fallback itself and die with it, which `docs/fallback.md` §8 treats as expected rather than a regression loss.

### Release note

This PR is safe to merge now, but per its own body the addon release must not precede the drain-B bound (≈17:19Z). That gate is owned by the release operator, not by this review, and merging does not discharge it.

PR #33583, opened 2026-09-11
refactor(platform): own the chat panel graph outside command execution

e7h4n:
## Summary

Restore lifecycle ownership for the chat panel graph and its thread-owned sidebar.

- Each chat pane derives its panel graph from the thread identity it publishes instead of asking `createCachedChatPanelSignals$` to construct the graph from a setup command. `ChatPanelSignals` no longer stores an `AbortSignal`; page-owned resources read `pageSignal$` when they start work.
- `ThreadSidebarSignals` now owns one resettable child signal for the current sidebar session. Every open receives a parent signal, replacing or closing the sidebar aborts the previous session, and attachment materialization binds File object URLs and Markdown/Mermaid blob URLs to that same child before publishing the target.
- Artifact refs no longer carry their own reset command. The sidebar session is the single resource owner, and the left and right pane factories retain isolated reset identities.
- Closing the right pane navigates through the router, mirroring the open path, so route teardown remains the pane setup lifetime.
- Preserve current `main`'s static realtime invalidation handlers while keeping the pane graph, draft-cache, and automation naming cleanup from this branch.

## Verification

- Added a page-level Mermaid preview assertion that closing the artifact sidebar revokes its sidebar-owned blob URL while leaving the inline diagrams intact.
- Existing page coverage exercises pane navigation, automation refresh, artifact replacement, and sidebar close/reopen behavior.
- Exact-head CI for `09f5d15c6216e54ee2bfef67f2e80091bac1afd9`: **80 passed, 22 path-filtered/skipped, 0 pending, 0 failed**.
- Passed all 8 app-test shards, all 8 API-test shards, TypeScript checks, ESLint (including ccstate rules), formatting, Knip, style, security analysis, and CLI E2E.
- The worktree had no installed dependency tree; no project dependency installation was performed. CI is the authoritative verification.

PR #33781, opened 2026-09-13
refactor(platform): prepare chat diagrams with the render window

e7h4n:
## Summary

Removes the last `Computed<AbortSignal>` owner threaded through the chat panel factories, by preparing diagrams where the render window already prepares the events it shows.

`createMermaidDiagramRegistry` needed an owner because `diagram$` allocated the blob URL inside a `computed`. A computed cannot receive a signal, so it reached for one captured at construction, and reading a value allocated a browser resource. That single consumer is why `createPagedEventResources` and `createChatThreadMessagePipeline` still carried an owner parameter after #33583.

Now:

- `renderDiagramFile` is the pure layout step: source plus theme in, SVG `File` out. It owns nothing.
- `ensureVisibleEventTrees$` collects the diagram sources of the events on screen and calls `ensureDiagrams$`, which lays out what is missing and binds each blob URL to that run's signal. On abort the URL is revoked and its entry dropped, so a later pass prepares the diagram again.
- `register$` only hands the view a computed that reads the prepared image. The parse step records each event's sources on its tree entry.
- `createPagedEventResources` and `createChatThreadMessagePipeline` lose their owner parameter. No chat panel factory takes an `AbortSignal`.

Two incidental effects:

- Diagrams are laid out for the visible window rather than when a view first reads them, so a diagram that is parsed but never scrolled into view is no longer rendered at all.
- Blob URLs previously lived until the whole panel was torn down. They are now tied to the run that prepared them, which bounds what a long session accumulates while a message streams, since each intermediate source produced its own permanent entry before.

Surfaces without a render window, the shared thread page and markdown preview trees, keep `createMermaidDiagramSignals` and pass their own owner signal. Their `no-computed-signal` suppression stays with them.

The registry keys by theme and source rather than by a digest of the source: entries are bounded by what is on screen, so key length does not matter, and a non-cryptographic hash that collided would show one diagram in place of another.

## Testing

- `tsc` production project, ESLint on both changed files, oxlint over `src/signals`, and `knip`.
- Page suites at two workers: Mermaid rendering, diagram rendering, markdown content, markdown artifacts, continuity (6 files), artifact sidebar, artifact viewer, scroll, message history, run state, structured content, private artifact previews.
- `tsc -p tsconfig.tests.json` reports pre-existing errors in `markdown.bench.tsx` (`Property 'bench' does not exist on type 'TestContext'`). I confirmed those same errors on clean `main`; they are not from this change.


e7h4n:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 2b0cb84ec4198891e3632e77c9abc038015aa95c

**Summary:** Moves Mermaid layout out of the `diagram$` computed and into the render window's preparation step, so each blob URL is owned by the signal of the run that prepared it. This removes the last `Computed<AbortSignal>` owner threaded through the chat panel factories, and `createPagedEventResources` / `createChatThreadMessagePipeline` no longer take one.

**Review Criteria:** [ccstate commands](https://github.com/vm0-ai/vm0/blob/main/.claude/skills/ccstate/references/commands.md) and [lifecycle](https://github.com/vm0-ai/vm0/blob/main/.claude/skills/ccstate/references/lifecycle.md) for owner and cancellation boundaries, [cache](https://github.com/vm0-ai/vm0/blob/main/docs/cache.md) RC-3/RC-8 for retained resources, [code quality](https://github.com/vm0-ai/vm0/blob/main/docs/bad-smell.md), and [app testing](https://github.com/vm0-ai/vm0/blob/main/docs/testing/app-testing.md).

## Findings

### P1 — Retrying a failed rich body leaves its diagram unprepared

`retryRichEventTree$` drops the failed entry and re-parses through `ensureEventTrees$`, which repopulates `diagramCodes`, but it never calls `ensureDiagrams$`.

Preparation is push-based after this change: only `ensureVisibleEventTrees$` prepares images. Before this PR `diagram$` laid the diagram out lazily when the view read it, so retry was self-healing.

Trigger: an event whose rich body failed to parse and contains a ` ```mermaid ` fence. The user clicks the retry control (`chat-thread-page.tsx:7331`). The tree now parses and renders, but `diagram$` resolves to `NEVER_RESOLVED_PROMISE`, so the diagram keeps its reserved placeholder until some unrelated later pass (a scroll or a new event) happens to run `ensureVisibleEventTrees$`. With no such event the spinner never resolves.

`createEventTreeSignals` already receives `registries.mermaidDiagrams`, so the retry path can prepare the event it just re-parsed without new plumbing.

No test covers retry together with a diagram.

### P2 — Orphaned doc comment above `renderDiagramFile`

`mermaid-diagram.ts` now has two stacked doc blocks: the original `createMermaidDiagramSignals` description ("Pure factory for a diagram's signals. `ownerSignal` must match the consumer surface...") was left in place and `renderDiagramFile` was inserted beneath it with its own block. The first block describes a different function.

### P2 — `createMermaidDiagramRegistry` doc contradicts the new ownership

Its doc still says "`owner` owns every entry's blob URLs, so tearing the surface down releases its diagrams. It is read at registration because a diagram outlives the parse that produced it." The function now takes no owner and reads none at registration; blob URLs are owned by the signal passed to `ensureDiagrams$`. This is the exact boundary the PR changes, so the stale text is actively misleading.

## Testing

`test-app` passes on all 8 shards and `lint-type-app` is green; 100 required checks pass with zero failures. Codecov reports 95.28% patch coverage with 5 uncovered lines in `mermaid-diagram.ts`.

Existing page suites cover the changed path meaningfully: `markdown-mermaid.test.tsx` asserts real rendered `diagram.svg` images, split view replacement, close-and-restore, and streaming readability, all of which now run through `ensureDiagrams$`.

Gap: no test exercises the retry control on an event containing a diagram, which is why the P1 above is not caught.

Browser verification on the PR preview confirmed the intended behavior: the diagram renders from the visible window (`data-mermaid-status="rendered"`, `blob:` src), split view opens with its own independent blob, leaving and re-entering the thread yields a new blob URL, and a theme switch re-lays out and swaps the URL again.

## Fallbacks

None introduced.

e7h4n:
LGTM


PR_AUTO_REVIEW_COMMIT: aa45d53bd161536beb80e82a09234e1b0246c92c
PR_AUTO_REVIEW_VERDICT: LGTM

**Summary:** Moves Mermaid layout out of the `diagram$` computed and into the render window's preparation step, so each blob URL is owned by the signal of the run that prepared it. This removes the last `Computed<AbortSignal>` owner threaded through the chat panel factories; `createPagedEventResources` and `createChatThreadMessagePipeline` no longer take one.

**Review Criteria:** [ccstate commands](https://github.com/vm0-ai/vm0/blob/main/.claude/skills/ccstate/references/commands.md) and [lifecycle](https://github.com/vm0-ai/vm0/blob/main/.claude/skills/ccstate/references/lifecycle.md) for owner and cancellation boundaries, [cache](https://github.com/vm0-ai/vm0/blob/main/docs/cache.md) RC-3/RC-8 for retained resources, [code quality](https://github.com/vm0-ai/vm0/blob/main/docs/bad-smell.md), and [app testing](https://github.com/vm0-ai/vm0/blob/main/docs/testing/app-testing.md) plus [external behavior](https://github.com/vm0-ai/vm0/blob/main/docs/testing/testing-external-behavior.md).

## Findings

All three findings from the review of `2b0cb84` are resolved in `aa45d53`; no new findings.

- **P1 retry gap — fixed.** `createRetryRichEventTree` now lays out the diagrams of the body it re-parsed. Extracting it also keeps `createEventTreeSignals` under the function-length limit rather than suppressing it.
- **P2 orphaned doc — fixed.** The stale `createMermaidDiagramSignals` block above `renderDiagramFile` is gone.
- **P2 stale registry doc — fixed.** It now describes `register$` as allocating nothing and `ensureDiagrams$` as owning the blob URLs through the preparing run's signal.

Ownership reads correctly at this HEAD: the only `createObjectUrlResource` call on the registry path takes the command's `signal` parameter, and no chat panel factory accepts an `AbortSignal`. The captured-owner path and its `no-computed-signal` suppression remain only in `createMermaidDiagramSignals`, used by the shared thread page and markdown preview trees, which have no render window; retiring those is tracked as follow-up work.

`internalSignals$` still grows per distinct diagram source for the life of the panel. That is unchanged from `main` — the entries are a string plus a computed, and the resources they point at are now bounded by the preparing run — so it is not a regression from this diff.

## Testing

100 required checks pass with zero failures, including `test-app` on all 8 shards and `lint-type-app`. Codecov reports 95.28% patch coverage.

`markdown-mermaid.test.tsx` asserts real rendered `diagram.svg` images, split-view replacement, close-and-restore, and streaming readability, all of which now run through `ensureDiagrams$`. Suites rerun locally at two workers after the repair: Mermaid rendering, diagram rendering, markdown content, markdown artifacts, artifact sidebar, message history.

Browser verification on the PR preview confirmed the intended behavior: the diagram renders from the visible window (`data-mermaid-status="rendered"`, `blob:` src), split view opens with its own independent blob, leaving and re-entering the thread yields a new blob URL, and a theme switch re-lays out and swaps the URL again.

**Known coverage gap:** the retry fix ships without a dedicated regression test. Reaching `richContentError` requires the markdown pipeline itself to throw, which a page-level test cannot bring about through the external interface; covering it would mean mocking an internal relative import, which the testing practice prohibits. The fix is a direct continuation of the existing preparation call already covered by the visible-window tests.

## Fallbacks

None introduced.

Issue #33766, opened 2026-09-13
fix(api): morning brief enters a permanent conflict when installed on two agents


lancy:
## Summary

A user can end up with two Morning Brief installations, which puts their Morning Brief into a **permanent, non-self-healing conflict state**. Every provisioning attempt and every read of the preference fails, on every login, forever.

This is currently affecting **1 user in the staff org** in production (25 occurrences and counting).

## Evidence from production

`vm0-web-logs-prod`, window 2026-09-10T12:00Z → 2026-09-13T16:00Z:

```
message: "Morning Brief initialization deferred"
fields.outcome.message: "Multiple Morning Brief installations exist"
fields.outcome.failureKind: "conflict"
level: warn
count: 25
distinct users: 1   (user_3EWY21Oe3f15kfs3yYmbGgDb3NV)
distinct orgs:  1   (org_3ANttyrbWYJk6JKRSTRLEsbsDLe)
first seen: 2026-09-11T14:56:26Z
last seen:  2026-09-13T15:27:09Z  (still recurring)
```

The affected org is the staff org (`identity-hash.ts:27` lists `afce210e // org_3ANttyrbWYJk6JKRSTRLEsbsDLe`). This matters — see root cause.

Note on provenance: the first occurrence at 2026-09-11T14:56:26Z **predates** the log-level change in #33540 (deployed 2026-09-11T15:20Z). This defect came in with #33434; #33540 only made it continuously visible. The `warn` at `user-preferences.ts:168` was already at `warn` before that PR.

## Root cause

**The Morning Brief conflict check is agent-agnostic, while every layer beneath it is agent-scoped.** Nothing prevents a second installation on a different agent, but the Morning Brief layer treats the resulting pair as an unrecoverable conflict.

| Layer | Scope | Reference |
| --- | --- | --- |
| Morning Brief installation lookup | `(orgId, ownerUserId, visibility, officialDefinitionName)` — **no `agentId`** | `morning-brief-preference.service.ts:185` |
| DB uniqueness guarantee | `(orgId, agentId, ownerUserId, name)` where private — **includes `agentId`** | `db/src/schema/workflow.ts:96` |
| Installer's own dedup / recovery | `(orgId, agentId, ownerUserId, visibility, name)` — **includes `agentId`** | `official-workflow-installation.service.ts:660` |

So two Morning Brief rows on two different agents are perfectly legal at the DB and installer level, and are then read back as a conflict by the Morning Brief layer.

### How the second row gets created

[...]

PR #33858, opened 2026-09-14
fix(api): remove the morning brief data-source gate

### LINKED ISSUE #33834: fix(api): remove the morning brief data-source gate
## Background

Morning Brief auto-enrollment refuses to install unless the user already has one of four connectors. In production this blocks the majority of eligible new users, permanently and silently.

Production evidence from `vm0-web-logs-prod`, window 2026-09-11T15:20Z → 2026-09-14T02:05Z (59h, starting from the observability fix in #33540):

| Outcome | Distinct users |
| --- | --- |
| `installed` | 38 |
| `skipped` / `missing-data-source` | 78 (37 later recovered, **41 still blocked**) |
| `skipped` / `missing-timezone` | 76 hit it, **0 terminally stuck** — fully transient |
| `skipped` / `missing-default-agent` | 2 |

Roughly **one in three** eligible new users actually receives a brief. The blocked share is not improving: the install-to-blocked ratio held near 1:2 on every full day in the window.

### The gate contradicts what the brief actually does

`hasMorningBriefDataSource` (`turbo/apps/api/src/signals/services/morning-brief-enrollment-data.service.ts:126`) requires one of exactly four connector slugs — `gmail`, `github`, `google-calendar`, `slack` — that is additionally `isDefault = true`, `needsReconnect = false`, **and** enabled for the agent that will run the brief (`userConnectors.agentId`), or an org-level Slack connection.

But the brief's own instruction (`MORNING_BRIEF_INSTRUCTION` in `turbo/apps/api/src/signals/services/official-workflow-catalog-source.ts`) defines **five** sources. The fifth is:

> **Unread Chats**: run `okou chat list --unread --all-agents`.

That is a built-in Okou capability requiring **no connector at all**. The gate does not count it.

The same instruction is already written to degrade gracefully:

> If a source, connector, thread, or capability is unavailable, say that it was unavailable and continue with the other sources. Never invent, infer, or claim source data that was not read.

So a user with zero connectors but active Okou chats would produce a perfectly useful brief, and is nonetheless blocked forever. The runtime already solves the "empty source" problem that the gate was added to prevent.

### The block never terminates

`loadUnavailableReason` (`morning-brief-preference.service.ts:146-152`) returns `missing-data-source`, the enrollment row stays `pending`, and the cron worker retries with a backoff capped at 15 minutes (`morning-brief-enrollment-worker.service.ts:16` and `:83-84`). There is **no maximum attempt count and no terminal failure state** for unmet preconditions — `ineligible` is only ever set by the rollout-boundary check (`morning-brief-enrollment-data.service.ts:63`), never here.

Worse, `ensureMorningBriefWhileLocked$` calls Clerk (`qualifyMorningBriefMembership$`) **before** it evaluates `loadUnavailableReason`. Every retry for a user who can never pass the gate still costs one Clerk API call: 41 users × 96 retries/day ≈ **3,900 wasted Clerk calls per day**, growing linearly with the blocked cohort.

## Design

**Remove the data-source gate entirely.** Report whatever sources are available; at minimum the brief always has unread Okou chats to draw on.

Reasoning:

- The gate asserts a precondition the runtime does not need. Its stated purpose was to avoid empty briefs, but the instruction already handles unavailable sources explicitly.
- The gate is blind to the one source that needs no setup, so it systematically blocks exactly the users who have nothing to connect yet.
- A brief that says "no activity in these sources today" is a better product outcome than a brief the user never receives and cannot diagnose.
- Removing the gate also removes the dominant cause of unbounded retries and the wasted Clerk traffic, without introducing a new terminal state.

Rejected alternative: keep the gate and add unread Chat as a qualifying source. This preserves a precondition that no longer gates anything meaningful — almost every account has some chat activity — while keeping the code, the failure mode, and the retry loop. Removing it is strictly simpler.

Rejected alternative: keep the gate and add a terminal `blocked` state with a retry ceiling. That was considered and dropped: with the gate gone there is no condition left that warrants it here.

### Self-healing

No migration or one-off script is required. The 41 currently blocked users still hold `pending` enrollment rows, and the cron worker re-evaluates each within 15 minutes. They install on the first retry after deploy.

### Secondary fix in the same change

The Morning Brief toggle is not disabled when `unavailableReason` is non-null (`turbo/apps/platform/src/views/okou-page/components/settings/morning-brief-settings.tsx:290`). Turning it on calls `createMorningBriefFromPreference$`, which returns the current state without installing, so the switch silently springs back off.

This is included here rather than split out because the same file and the same `unavailableReason` field are already being edited to delete the `missing-data-source` branch, and a second PR touching the same component would only create a rebase. After this change the remaining reasons are `missing-timezone` (effectively unreachable in Settings, since login sets the timezone) and `missing-default-agent` (an org-level setting the user cannot fix themselves), so disabling the control with its existing explanatory copy is the honest behavior.

## Non-goals

- Do **not** change the `missing-timezone` or `missing-default-agent` checks. Production data shows the timezone path is transient and self-resolving; leave it alone.
- Do **not** add a terminal enrollment state, a retry ceiling, or any change to the enrollment state machine (`checking` / `pending` / `completed` / `cancelled` / `ineligible` / `departed`).
- Do **not** change the cron worker's backoff or claim/lease logic.
- Do **not** change `MORNING_BRIEF_INSTRUCTION` or any part of the official workflow catalog definition.
- Do **not** change the 07:00 schedule, result-email delivery, or email subscription behavior.
- Do **not** change feature switch values for `morningBrief` or `officialWorkflows`.
- Do **not** attempt to fix #33766 (multiple installations conflict). It is open, real, and out of scope here.
- Do **not** add metrics, dashboards, or Axiom monitors.

## Acceptance criteria

- `missing-data-source` no longer appears anywhere in `turbo/`, including the API contract union, the service reason type, the settings component, and every locale file.
- `hasMorningBriefDataSource` is deleted, along with the imports it alone required in `morning-brief-enrollment-data.service.ts` (`connectors`, `userConnectors`, `slackOrgConnections`, `slackOrgInstallations` — each currently appears exactly once outside that function, as its import).
- `loadUnavailableReason` returns only `missing-timezone`, `missing-default-agent`, or `null`.
- A user with a valid timezone and a default agent but **zero connectors** installs Morning Brief successfully. Cover this with a regression test — it is the whole point of the change.
- The Settings toggle is disabled whenever `unavailableReason` is non-null, and the existing reason copy is still shown.
- No change to timezone handling, the enrollment state machine, worker backoff, the catalog definition, scheduling, or email delivery.

## Notes

Scale check: `morning_brief_enrollments` and `morning_brief_rollout` are not exposed in the MaskDB controlled subset (124 tables), so the user counts above come from deduplicated Axiom log records and undercount members who never logged in. No database-side sizing was possible; none is needed, since the change requires no migration.

Accepted trade-off: this adds roughly 41 daily brief runs immediately, growing by ~28 users/day at the current rate, each consuming credits. That cost was reviewed and accepted in favour of delivering the feature to users who currently receive nothing.

lancy:
Closes #33834

## What

Removes the Morning Brief data-source gate so a user with a valid time zone and a default Agent installs the brief even with zero connectors, and disables the Settings toggle while a reason is still reported.

## Why

`hasMorningBriefDataSource` required one of `gmail`, `github`, `google-calendar`, or `slack` to be default, healthy, and enabled for the Agent that runs the brief. `MORNING_BRIEF_INSTRUCTION` defines a fifth source, unread Okou chats (`okou chat list --unread --all-agents`), which needs no connector, and the instruction already says an unavailable source should be reported and skipped. The gate therefore asserted a precondition the runtime does not need, and blocked exactly the users who have nothing to connect yet: production showed 38 installs against 41 users still blocked over a 59h window, plus roughly 3,900 wasted Clerk calls per day from the unbounded retry loop.

Keeping the gate and adding unread chat as a fifth qualifying source was rejected in the issue: almost every account has chat activity, so the check would stop gating anything while keeping the code, the failure mode, and the retry loop.

## Changes

- `morning-brief-preference.ts`: drop `missing-data-source` from the contract union.
- `morning-brief-enrollment-data.service.ts`: delete `hasMorningBriefDataSource` and the four imports it alone required (`connectors`, `userConnectors`, `slackOrgConnections`, `slackOrgInstallations`).
- `morning-brief-preference.service.ts`: drop the import and the skip reason; `loadUnavailableReason` now returns only `missing-timezone`, `missing-default-agent`, or `null`. The `installationAgentId` parameter is kept for the default-Agent check.
- `morning-brief-settings.tsx`: remove the `missing-data-source` branch and disable the toggle whenever `unavailableReason` is non-null. Enabling it called `createMorningBriefFromPreference$`, which returned the current state without installing, so the switch silently sprang back. The reason copy is still shown. The predicate was extracted to keep the component within the `complexity` lint limit.
- 10 locale files: remove the `missingDataSource` key.

## Tests

- `official-workflows.test.ts`: a member with a valid time zone and a default Agent but no connector now installs on the first visit and stays at one installation across concurrent cron ticks; a new case covers the same through the preference toggle (`PUT /api/preferences/morning-brief`), which the gate previously turned into a silent no-op. Cancellation-before-installation cases now cancel before the visit that would install.
- `preference-settings.test.tsx`: new regression test that the toggle is `aria-disabled` while a reason is reported, the reason copy is still visible, and it becomes interactive once the reason clears.

## Verification

- `pnpm -F api check-types`, `pnpm -F @okouai/app check-types`, `pnpm -F @okouai/api-contracts check-types` — pass
- `pnpm -F api lint`, `pnpm -F @okouai/app lint` (incl. i18n key check), `pnpm lint:style`, `pnpm knip` — pass
- `vitest run src/views/okou-page/__tests__/preference-settings.test.tsx` — 15 passed
- `vitest run official-workflows.test.ts -t "Morning Brief"` — 18 passed; `-t "Official Workflow installations"` — 36 passed; `cron-official-workflow-catalog.test.ts` — 14 passed
- The `Official Workflow Run admission` cases fail identically on unmodified `main` in this sandbox (`Job not found in queue`), so they are a local environment limit, not a regression. CI covers them.

## Notes

No migration or backfill. The 41 blocked users still hold `pending` enrollment rows and the cron worker re-evaluates each within 15 minutes, so they install on the first retry after deploy. No change to time zone handling, the enrollment state machine, worker backoff, the catalog definition, scheduling, email delivery, or the `morningBrief` / `officialWorkflows` switches. #33766 is untouched.


lancy:
Changes Requested


PR_AUTO_REVIEW_COMMIT: 14650dc642db279d31fac66581d7f4ecf5cc6b69

Inspected HEAD: `14650dc642db279d31fac66581d7f4ecf5cc6b69`. Practice documents read from `main` at `fa2e6e6212dee848c8d37d72551cc5d9b7887ac4`.

### Summary

Removes the Morning Brief data-source precondition (`hasMorningBriefDataSource`) and the `missing-data-source` reason from the API contract, the preference service, the Settings component, and all 10 locales, so `loadUnavailableReason` returns only `missing-timezone`, `missing-default-agent`, or `null`. Also disables the Settings toggle whenever `unavailableReason` is non-null, fixing the switch that silently sprang back because `createMorningBriefFromPreference$` returned the unchanged preference instead of installing.

The production change is sound: the gate required a connector to be default, healthy, and Agent-enabled, while the brief's own instruction defines unread Okou chats as a fifth source that needs no connector and explicitly reports unavailable sources and continues. Removal is the correct scope, and the deletion covers the branch, the contract entry, the copy, and the tests together.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | `loadUnavailableReason` is reached from `ensureMorningBriefWhileLocked$` (webhook, `user-preferences` initialize/update, onboarding, cron worker) and `createMorningBriefFromPreference$` (PUT). Both now fall through to install when timezone and default Agent exist. |
| `REVIEW.md` | Features that change/remove an existing flow need regression coverage for the affected path | Pass | `official-workflows.test.ts` adds first-visit install with zero connectors and toggle-driven install with zero connectors; `preference-settings.test.tsx` adds the disabled-toggle regression. |
| `REVIEW.md` | Do not infer a defect from unchanged code | Pass | `unavailableFailure`'s `missing-timezone` branch is reachable only in theory, but `origin/main` already has one call site (`unavailableFailure("missing-default-agent")`); the diff does not touch it, so it is out of scope here. |
| `REVIEW.md` | Report concrete practice violations with rule and source location | Finding (P1) | See F1 (`docs/testing/anti-patterns.md` AP-13). |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions, no weakened rules | Pass | No `any`, no suppression comments. `isToggleDisabled` was extracted so the added condition stays within the existing `complexity` limit rather than disabling the rule. |
| `docs/bad-smell.md` | Interface changes: document the changed public contract and assess its consumers | Pass | `morningBriefUnavailableReasonSchema` loses one value. Consumers are `apps/api/src/signals/routes/morning-brief-preference.ts` and `apps/platform/src/signals/okou-page/settings/morning-brief-preference.ts` only; the contract is absent from `runtime-api-schema`, `python-bindings`, and `rust-bindings`, so no Runner/CLI consumer exists. |
| `docs/bad-smell.md` | Do not remove meaningful recovery, fail-closed security, or expected external-reference handling | Pass | The removed code is a product precondition, not recovery or a security check. Auth, membership qualification, feature-switch gating, and the rollout-boundary eligibility check are untouched. |
| `docs/fallback.md` | Removal of a branch must carry type/schema, single-writer, production, or rollout evidence | Pass | The removed gate is a precondition, and the issue supplies production evidence (38 installed vs 41 permanently blocked over 59h) plus the runtime's own "source unavailable -> continue" contract. |
| `docs/fallback.md` | Removal deletes the branch, its contract entry, and its own tests together | Pass | Function, contract enum value, UI branch, 10 locale strings, and the gate-dependent assertions are removed in one change; `grep missing-data-source\|hasMorningBriefDataSource` over `turbo/` returns nothing. |
| `docs/fallback.md` | Every new fallback must be declared in a `Fallbacks` PR-body section | N/A | The diff introduces no fallback. `preference?.unavailableReason` guards the genuine loading state, not a contract-impossible value, so no `Fallbacks` section is required. |
| `docs/fallback.md` | No tombstone tests asserting removed behavior stays removed | Pass | The gate's tests were rewritten to the new behavior or removed, not kept as negative assertions. |
| `docs/deployment-compatibility.md` | Do not remove a response value while old frontend code may still read it | Pass | Narrowing is server-produced only. "Production promotes API traffic before it promotes the app", so the new frontend never precedes the new API; the old frontend keeps handling `missing-timezone` / `missing-default-agent` / `null`. |
| `docs/deployment-compatibility.md` | Cover the old/new combinations reachable during rollout | Pass (with note) | Old app + new API: API stops emitting the value, old app is unaffected. New app + rolled-back API: production `validateResponse` is off (`api-client-base.ts:107`, validation only `IN_VITEST`), so an unknown reason matches no copy branch and leaves the toggle disabled — degraded copy, no crash, no incorrect enable. See Observations. |
| `docs/deployment-compatibility.md` | Database schema/migration compatibility | N/A | No schema, migration, or persisted-state change. Blocked rows stay `pending` and the existing cron worker re-evaluates them. |
| `docs/testing.md` | Add tests for new behavior and regressions where they add confidence | Pass | Two API cases and one Platform case cover the two install entry points and the toggle fix. |
| `docs/testing.md` | Avoid duplicate cases that merely restate existing coverage | Pass | The rewritten onboarding case keeps the concurrent double-tick idempotency check rather than duplicating the catalog-recovery and Clerk-recovery retry cases. |
| `docs/testing/testing-external-behavior.md` | API tests construct and verify through production endpoints, never DB rows or services | Pass | Setup and verification use `userPreferences.initialize`, the Clerk webhook, `enrollMorningBrief`, `morningBriefPreferenceClient`, and `workflowCollectionClient().list`. No DB/service imports added. |
| `docs/testing/testing-external-behavior.md` | Platform tests construct and verify through the page | Pass | New case uses `setupPage`, `findByTestId`/`findByText`, `getByRole("switch")`, and `click`. |
| `docs/testing/api-testing.md` | Use `setupApp` + contract + `accept`, mock only external services | Pass | New cases reuse the file's existing `accept(...)` clients and `context.mocks.clerk` membership mock; no `vi.mock` of internal modules. |
| `docs/testing/app-testing.md` | Install mocks before `setupPage`; `switch` is an allowed role query | Pass | Both `context.mocks.api` handlers are registered before `setupPage`; `switch` is on the guide's allowed-role list. |
| `docs/testing/app-testing.md` | Default to `toBeInTheDocument()`; reserve `toBeVisible()` for show/hide contracts | Finding (P1) | F1 — `preference-settings.test.tsx` new test asserts reason-copy presence with `.resolves.toBeVisible()`. |
| `docs/testing/anti-patterns.md` | AP-13: do not require visibility for content presence | Finding (P1) | F1. |
| `docs/testing/anti-patterns.md` | AP-1/AP-7: verify observable state, not captured mock calls | Observation | The new test proves the re-enabled click through the captured request body rather than the resulting checked state. See O1. |
| `docs/testing/anti-patterns.md` | AP-10: no direct component rendering or store mutation | Pass | The new test enters through the Router and drives realtime through `context.mocks.ably`. |
| `.claude/skills/ccstate/SKILL.md` | No signal creation or store mutation during render; derivation stays pure | Pass | `isToggleDisabled` is a pure module-level predicate over already-derived values; no new signals, effects, refs, or `detach` usage. |
| `.claude/skills/ccstate/SKILL.md` | `accept` with a non-empty status list; preserve loadable lifecycle UI | Pass | `morning-brief-preference.ts` signal is unchanged; the component still branches on `loading` / `loadFailed` / `mutating` loadable state. |
| `.claude/skills/feature-switch/SKILL.md` | New user-facing features require a switch; bug fixes to existing features do not | Pass | `morningBrief` is already `enabled: true` (GA) and unchanged. This is a bug fix to that feature, which the skill lists as not requiring a new switch. |

### Findings

**P1**

- **F1 — `turbo/apps/platform/src/views/okou-page/__tests__/preference-settings.test.tsx` (new test, reason-copy assertion).** The new `keeps the Morning Brief toggle disabled while an unavailable reason is reported` case asserts content presence with `await expect(within(card).findByText("Choose a usable default Agent before enabling Morning Brief.")).resolves.toBeVisible();`. `docs/testing/anti-patterns.md` AP-13 and `docs/testing/app-testing.md#assertions` require `toBeInTheDocument()` for a presence contract and reserve `toBeVisible()` for behavior that specifically shows or hides content. Nothing in this case toggles the copy's visibility — the contract is that the reason is still rendered while the control is disabled. **Trigger:** any enter-animation or computed-style timing change in the settings card. **Consequence:** the assertion fails for a presentation reason unrelated to the behavior it protects. Use `.resolves.toBeInTheDocument()`.

### Testing

Coverage added is proportionate and targets the acceptance criterion directly:

- `official-workflows.test.ts` — `installs without any connected source and stays single across concurrent cron ticks` proves the auto-enrollment path (valid timezone + default Agent + zero connectors) now installs on the first visit, and retains the previous case's concurrent `tickBriefEnrollment` idempotency check. `installs from the preference toggle without any connected source` covers `createMorningBriefFromPreference$`, which the gate previously turned into a silent no-op and which is the server half of the toggle fix.
- Cancellation cases (`preserves a cancellation before installation...`, `isolates enrollment, sources, and cancellation...`) now cancel before the visit that would install, which preserves their stated intent instead of relying on the deleted gate to keep the member uninstalled. `cancels pending enrollment when membership is removed...` drops the timezone so the enrollment is genuinely pending, matching its title.
- `missing-timezone` and `missing-default-agent` remain covered by the untouched `preserves enable intent while timezone or default Agent is unavailable`.
- Cron-worker retry after a transient failure is still covered by `recovers a transient installation failure without repeating onboarding` and `keeps membership verification retryable when Clerk is temporarily unavailable`, so rewriting the gate-dependent case does not lose that path.

No material coverage gap beyond F1. The PR body's verification list matches the changed surfaces.

- **O1 — request capture vs. page state.** The new Platform case ends with `expect(updated).toStrictEqual([true])`. That is a capture at the external HTTP boundary, which `app-testing.md` permits, but the stronger page-observable proof that the control became operable is that the switch renders checked after the click. Consider asserting `toBeChecked()` instead of, or in addition to, the captured body.
- **O2 — rollback-only degradation.** If the API is rolled back below this change while the new app is live, `missing-data-source` reaches a frontend whose contract no longer lists it. Production does not validate responses, so it matches no copy branch and the toggle stays disabled with the `preparing` copy and an alert icon. That is safe and transient, and per `docs/deployment-compatibility.md` the promotion order makes the forward direction safe, so no compatibility branch is warranted here. Recording it so the behavior is not rediscovered as a bug.

lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: ef50e36bdaab82a0075923d35d052b44921dc577
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `ef50e36bdaab82a0075923d35d052b44921dc577`. Practice documents read from `main` at `fa2e6e6212dee848c8d37d72551cc5d9b7887ac4`. This supersedes the review of `14650dc`; `ef50e36` adds `test(app): assert brief toggle copy presence and checked state`, which resolves F1 and O1 from that review.

### Summary

Removes the Morning Brief data-source precondition (`hasMorningBriefDataSource`) and the `missing-data-source` reason from the API contract, the preference service, the Settings component, and all 10 locales, so `loadUnavailableReason` returns only `missing-timezone`, `missing-default-agent`, or `null`. Also disables the Settings toggle whenever `unavailableReason` is non-null, fixing the switch that silently sprang back because `createMorningBriefFromPreference$` returned the unchanged preference instead of installing.

The gate required a connector to be default, healthy, and Agent-enabled, while the brief's own instruction defines unread Okou chats as a fifth source that needs no connector and explicitly reports unavailable sources and continues. Removal is the correct scope, and the deletion covers the branch, the contract entry, the copy, and the tests together.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | `loadUnavailableReason` is reached from `ensureMorningBriefWhileLocked$` (Clerk webhook, `user-preferences` initialize/update, onboarding, cron worker) and `createMorningBriefFromPreference$` (PUT). Both now fall through to install when timezone and default Agent exist. |
| `REVIEW.md` | Features that change/remove an existing flow need regression coverage for the affected path | Pass | `official-workflows.test.ts` adds first-visit install with zero connectors and toggle-driven install with zero connectors; `preference-settings.test.tsx` adds the disabled-toggle regression. |
| `REVIEW.md` | Do not infer a defect from unchanged code | Pass | `unavailableFailure`'s `missing-timezone` branch is reachable only in theory, but `origin/main` already has the single call site `unavailableFailure("missing-default-agent")`; the diff does not touch it, so it is out of scope here. |
| `REVIEW.md` | Report concrete practice violations with rule and source location | Pass | The AP-13 violation raised against `14650dc` is fixed in `ef50e36`; no violation remains at this HEAD. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions, no weakened rules | Pass | No `any`, no suppression comments. `isToggleDisabled` was extracted so the added condition stays within the existing `complexity` limit rather than disabling the rule. |
| `docs/bad-smell.md` | Interface changes: document the changed public contract and assess its consumers | Pass | `morningBriefUnavailableReasonSchema` loses one value. Consumers are `apps/api/src/signals/routes/morning-brief-preference.ts` and `apps/platform/src/signals/okou-page/settings/morning-brief-preference.ts` only; the contract is absent from `runtime-api-schema`, `python-bindings`, and `rust-bindings`, so no Runner/CLI consumer exists. |
| `docs/bad-smell.md` | Do not remove meaningful recovery, fail-closed security, or expected external-reference handling | Pass | The removed code is a product precondition, not recovery or a security check. Auth, membership qualification, feature-switch gating, and the rollout-boundary eligibility check are untouched. |
| `docs/fallback.md` | Removal of a branch must carry type/schema, single-writer, production, or rollout evidence | Pass | The removed gate is a precondition, and #33834 supplies production evidence (38 installed vs 41 permanently blocked over 59h) plus the runtime's own "source unavailable -> continue" contract. |
| `docs/fallback.md` | Removal deletes the branch, its contract entry, and its own tests together | Pass | Function, contract enum value, UI branch, 10 locale strings, and the gate-dependent assertions are removed in one change; `grep missing-data-source\|hasMorningBriefDataSource` over `turbo/` returns nothing. |
| `docs/fallback.md` | Every new fallback must be declared in a `Fallbacks` PR-body section | N/A | The diff introduces no fallback. `preference?.unavailableReason` guards the genuine loading state, not a contract-impossible value, so no `Fallbacks` section is required. |
| `docs/fallback.md` | No tombstone tests asserting removed behavior stays removed | Pass | The gate's tests were rewritten to the new behavior or removed, not kept as negative assertions. |
| `docs/deployment-compatibility.md` | Do not remove a response value while old frontend code may still read it | Pass | Narrowing is server-produced only. "Production promotes API traffic before it promotes the app", so the new frontend never precedes the new API; the old frontend keeps handling `missing-timezone` / `missing-default-agent` / `null`. |
| `docs/deployment-compatibility.md` | Cover the old/new combinations reachable during rollout | Pass (with note) | Old app + new API: the API stops emitting the value, the old app is unaffected. New app + rolled-back API: production `validateResponse` is off (`api-client-base.ts:107`, validation only `IN_VITEST`), so an unknown reason matches no copy branch and leaves the toggle disabled — degraded copy, no crash, no incorrect enable. See O1. |
| `docs/deployment-compatibility.md` | Database schema/migration compatibility | N/A | No schema, migration, or persisted-state change. Blocked rows stay `pending` and the existing cron worker re-evaluates them. |
| `docs/testing.md` | Add tests for new behavior and regressions where they add confidence | Pass | Two API cases and one Platform case cover the two install entry points and the toggle fix. |
| `docs/testing.md` | Avoid duplicate cases that merely restate existing coverage | Pass | The rewritten onboarding case keeps the concurrent double-tick idempotency check rather than duplicating the catalog-recovery and Clerk-recovery retry cases. |
| `docs/testing/testing-external-behavior.md` | API tests construct and verify through production endpoints, never DB rows or services | Pass | Setup and verification use `userPreferences.initialize`, the Clerk webhook, `enrollMorningBrief`, `morningBriefPreferenceClient`, and `workflowCollectionClient().list`. No DB/service imports added. |
| `docs/testing/testing-external-behavior.md` | Platform tests construct and verify through the page | Pass | The new case uses `setupPage`, `findByTestId`/`findByText`, `getByRole("switch")`, `click`, and `context.mocks.ably`. |
| `docs/testing/api-testing.md` | Use `setupApp` + contract + `accept`, mock only external services | Pass | New cases reuse the file's existing `accept(...)` clients and the `context.mocks.clerk` membership mock; no `vi.mock` of internal modules. |
| `docs/testing/app-testing.md` | Install mocks before `setupPage`; `switch` is an allowed role query | Pass | Both `context.mocks.api` handlers are registered before `setupPage`; `switch` is on the guide's allowed-role list. |
| `docs/testing/app-testing.md` | Default to `toBeInTheDocument()`; reserve `toBeVisible()` for show/hide contracts | Pass | Fixed at this HEAD: the reason-copy assertion is now `.resolves.toBeInTheDocument()`. |
| `docs/testing/anti-patterns.md` | AP-13: do not require visibility for content presence | Pass | Same fix. |
| `docs/testing/anti-patterns.md` | AP-1/AP-7: verify observable state, not captured mock calls | Pass | Fixed at this HEAD: the captured request-body array is gone and the re-enabled click is proven by the switch rendering checked, which only happens once the update request reaches the mocked boundary. |
| `docs/testing/anti-patterns.md` | AP-10: no direct component rendering or store mutation | Pass | The new test enters through the Router and drives realtime through `context.mocks.ably`. |
| `.claude/skills/ccstate/SKILL.md` | No signal creation or store mutation during render; derivation stays pure | Pass | `isToggleDisabled` is a pure module-level predicate over already-derived values; no new signals, effects, refs, or `detach` usage. |
| `.claude/skills/ccstate/SKILL.md` | `accept` with a non-empty status list; preserve loadable lifecycle UI | Pass | The `morning-brief-preference.ts` signal is unchanged; the component still branches on `loading` / `loadFailed` / `mutating` loadable state. |
| `.claude/skills/feature-switch/SKILL.md` | New user-facing features require a switch; bug fixes to existing features do not | Pass | `morningBrief` is already `enabled: true` (GA) and unchanged. This is a bug fix to that feature, which the skill lists as not requiring a new switch. |

### Findings

None. F1 from the `14650dc` review is resolved at this HEAD.

### Testing

Coverage is proportionate and targets the acceptance criterion directly:

- `official-workflows.test.ts` — `installs without any connected source and stays single across concurrent cron ticks` proves the auto-enrollment path (valid timezone + default Agent + zero connectors) now installs on the first visit, and retains the previous case's concurrent `tickBriefEnrollment` idempotency check. `installs from the preference toggle without any connected source` covers `createMorningBriefFromPreference$`, which the gate previously turned into a silent no-op and which is the server half of the toggle fix.
- Cancellation cases (`preserves a cancellation before installation...`, `isolates enrollment, sources, and cancellation...`) now cancel before the visit that would install, preserving their stated intent instead of relying on the deleted gate to keep the member uninstalled. `cancels pending enrollment when membership is removed...` drops the timezone so the enrollment is genuinely pending, matching its title.
- `missing-timezone` and `missing-default-agent` remain covered by the untouched `preserves enable intent while timezone or default Agent is unavailable`.
- Cron-worker retry after a transient failure is still covered by `recovers a transient installation failure without repeating onboarding` and `keeps membership verification retryable when Clerk is temporarily unavailable`, so rewriting the gate-dependent case does not lose that path.
- `preference-settings.test.tsx` — the new case proves the toggle is `aria-disabled` while a reason is reported, the reason copy is still rendered, and the control becomes operable and checks once the reason clears.

No material coverage gap remains.

- **O1 — rollback-only degradation.** If the API is rolled back below this change while the new app is live, `missing-data-source` reaches a frontend whose contract no longer lists it. Production does not validate responses, so it matches no copy branch and the toggle stays disabled with the `preparing` copy and an alert icon. That is safe and transient, and per `docs/deployment-compatibility.md` the promotion order makes the forward direction safe, so no compatibility branch is warranted. Recorded so the behavior is not rediscovered as a bug.

PR #33878, opened 2026-09-14
refactor: centralize mermaid rendering and remove fixed checkout paths

e7h4n:
## Summary

Closing or replacing a Markdown preview could invalidate Mermaid image URLs still needed by another surface. Layouts now resolve through scoped async computeds to SVG `File` objects. Each mounted image owns its URL, and expanded previews receive the file and allocate a separate session URL. Abandoned layouts allocate no image URL.

- Initialize the statically imported Mermaid engine once in a renderer adapter and use its built-in parse/render queue. Preserve the white canvas and black labels across app themes, and copy the resolved font-family into serialized SVGs.
- Keep source registries scoped to their thread or preview. Chat prewarming retains the layout wait needed for scroll restoration and isolates individual layout failures from chat synchronization.
- Remove duplicate layout publication state, computed-signal suppressions, unused preview prewarming commands, obsolete Markdown retry state, and outdated ownership comments. Preserve attachment session resets that release preview URLs.
- Resolve the snapshot helper's `turbo` directory relative to its script instead of `/workspaces/vm01/turbo`. Runner-image waiting fixtures explicitly control repository identity and cover `GITHUB_REPOSITORY` precedence over the standalone `REPO` fallback.
- Synchronize the saved-chat regression on visible startup stages: loading appears, the saved response is visible, and loading is dismissed. Keep the thread-state request pending and retain the empty-conversation assertion. `startPage()` is intentional because waiting for full `setupPage()` completion would await that blocked request.

## Validation

All required PR checks passed at `2872161757f637341e3c4db5a5a9da54dd98da7c`: [Turbo](https://github.com/vm0-ai/okou/actions/runs/35032486070/job/104595624377), [Crates](https://github.com/vm0-ai/okou/actions/runs/35032486022/job/104596144123), and [Security](https://github.com/vm0-ai/okou/actions/runs/35032485922/job/104595980983).

All eight app shards passed. [test-app (6)](https://github.com/vm0-ai/okou/actions/runs/35032486070/job/104594053129) reports 275/275 passing, including all seven cache tests and the repaired saved-message case (1,322 ms). CI ESLint, Platform types, style policy, Knip, and downstream E2E also passed.

- Integrated-head `chat-offline-cache.test.tsx`: 7/7 passed, including the repaired saved-message test (620 ms). The focused repair also passed five completed runs during diagnosis, including coverage-enabled runs, before main integration.
[...]

PR #33887, opened 2026-09-14
refactor(platform): bench the chat parse pass through the diagram registry

e7h4n:
## Why

The markdown benches model the chat parse pass, but they still called `createMermaidDiagramSignals` with the test signal. Since #33781 that pass registers each diagram source with `MermaidDiagramRegistry` and leaves layout — and the blob URL — to the render window, so the benches were measuring a path the product no longer takes and allocating a diagram owner they do not represent.

This is the prerequisite for removing `createMermaidDiagramSignals`: after it, the only remaining callers are the two production surfaces that still capture an `AbortSignal` for diagrams (the shared thread page, handled in #33878, and the markdown preview trees).

## What

`prepareTree` and `ensurePass` now embed `mermaidDiagrams.register$` results, matching `createEventTreeParser`.

## Verification

- `pnpm check-types` in `turbo/apps/platform`
- `eslint` + `oxlint` on the changed file
- `vitest bench --run --maxWorkers=2 src/views/components/test/__benches__/markdown.bench.tsx` — 13 passed

## Note

The `Property 'bench' does not exist on type 'TestContext'` error reported against this file does not reproduce: `tsconfig.tests.json` covers `__benches__` and `pnpm check-types` is clean on `main`, as is `tsc -p tsconfig.json`. Nothing was changed for it.

Based on #33781; only the last commit belongs to this PR.


e7h4n:
Folded into #33878 — the bench change and the shared thread change ship as one PR.

PR #33837, opened 2026-09-14
fix(app): separate slash menu hover and keyboard selection

bingjiez666:
When a slash-menu item is selected with the arrow keys, hovering another row must leave the keyboard target unchanged. Enter and Tab now activate the keyboard selection, while clicking activates the pointer target. Category and workflow rows use distinct existing hover, selected, and selected-hover colors.

The right-hand template preview has an independent index. Pointer movement updates only that preview; arrow navigation, query changes, closing/reopening the menu, and leaving the whole panel restore the keyboard preview. Moving from a category into its template covers preserves that preview and allows selecting a cover.

## Validation

- 42 targeted tests passed across the slash-panel and existing workflow composer suites.
- Regression coverage includes filtered/unfiltered workflow lists, repeated movement within a row, Enter, Tab, click, continued arrow navigation, category actions, the first-row boundary, preview exit, query reset, reopen, and cover selection.
- The five core keyboard/hover regression cases fail against the previous implementation and pass with this change.
- Scoped Prettier, ESLint, Oxlint (including type-aware rules), style ESLint, style policy, platform TypeScript checks, and platform Knip passed.
- Browser QA passed on the deployed `470b0bd` preview: Chromium with a fine pointer, 1440×1000, light and dark themes. Verified distinct row colors, keyboard/pointer activation, arrow/Tab navigation, category offsets, preview boundaries, query reset, reopening, and template-cover selection. Used three browser-only workflow fixtures and a test account with onboarding bypassed.
- Current-head CI: 80 successful checks and 22 skipped checks; no pending or failed checks.

## Acceptance

Enable `composerCreateCommands` and `composerSlashTemplatePanel` in the preview Lab. Open `/`, select row B with the arrow keys, and move the pointer within row A. A should show hover and B should retain selection. Enter/Tab should activate B; repeating the setup and clicking A should activate A. Hover Website and move into its covers to confirm the preview stays available.

PR #33894, opened 2026-09-14
feat(app): add a deck import card and uncrop illustration covers

tongx-blip:
## What

The Presentation pane in the `/` panel now leads with an **Import your own deck**
card instead of starting straight at the catalog covers. Still behind
`composerSlashTemplatePanel`.

Rendered at real geometry: https://okou-slash-import-card.okou.app

## How

It reuses the picker dialog's import path rather than adding a second one:
`importPresentationTemplateDeck$` and `PRESENTATION_TEMPLATE_IMPORT_ACCEPT` from
`signals/okou-page/presentation-template-import.ts`, plus the existing
`artifacts.templates.importDeck` / `importDeckHint` strings — so no new i18n and
nothing to translate into ten locales.

The dialog's `PptImportCard` component itself could not be reused:
`chat-composer.tsx` imports the TipTap composer, which imports this panel, so
importing back would be a cycle. Its tile geometry is wrong here anyway — these
cards are 139px against the dialog's full-width tiles.

Geometry matches the sibling covers exactly: same `aspect-video` tile, same
`rounded-lg ring-1 ring-border/60`, same single-line truncated caption, so row
heights stay equal. The accepted formats sit on the tile under the `+` because
the caption line cannot hold `Import your own deck (.pptx, .ppt, .pdf)` at 139px.

Only Presentation carries the control — the command imports decks and no other
category has an equivalent.

## The bug the test caught

The label prevents its `mousedown` default. Without that, this feature is
silently broken: the panel is mounted off the editor's **active slash range**, so
letting the file input take focus clears the range and unmounts the input before
the native file dialog can return. The user picks a deck and nothing happens.

Probed directly before fixing — after the upload interaction the menu was gone
(`menuStillOpen=false`), the input was detached (`inputConnected=false`), and the
files were set on an element React no longer listened to. Label activation still
forwards the click, so the dialog opens with the caret left where it was. Every
other row in this panel already prevents its mousedown default for the same
reason.

## Illustration covers keep their own proportion

Added after design review of the shipped panel.

An illustration **is** the artifact, and the catalog runs **20 portrait, 9 square
and 3 landscape**, so the shared 16:9 tile cropped most of them:

| style | source | in a 16:9 tile |
| --- | --- | --- |
| 2:3 portrait (×16) | 1024×1536 | **37%** of the height survives |
| narrowest | 704×1472 | **27%** |
| square (×9) | 1024×1024 | 56% |

Illustration covers now render at the catalog's own ratio inside a CSS
multi-column masonry. That is not a new invention: `chat-composer.tsx` already
does exactly this for the picker dialog (`columns-[244px]` with
`aspectRatio: w/h` per tile) and its comment says it mirrors
`www.okou.ai/illustration`. Both entry points now show the same thing.

Presentation, Video and Website keep the 16:9 tile — a deck cover genuinely is
a slide, and cropping it to portrait would be the wrong fix.

**They were also soft.** `r2ImageTransformUrl` defaults to `fit=scale-down`, so
it fits inside the box it is given rather than cropping. Asking for 280×158
therefore shrank a 1024×1536 style to **105px wide**, which the 139px card then
upscaled. Illustration now asks by width alone, as the dialog's card does.

Design comparison of the two options considered:
https://okou-slash-cover-aspect.okou.app

Added test: an illustration cover requests width only and declares the catalog's
own ratio, while a deck cover still requests the 16:9 box. That pins the exact
regression — a shared height silently returns both the crop and the softness.

## Verification

- `vitest run` on `chat-composer-slash-panel` — 10 passed, three new: the control is
  first in the grid and absent from other categories; and a real upload that
  asserts the deck is sent with the import prompt, mirroring
  `chat-composer-presentation-import.test.tsx`
- `chat-composer-presentation-import` (4), `chat-composer-workflows` (17),
  `chat-composer-create` (19) — 47 passed, so the dialog's own import path is
  unchanged
- `pnpm run check-types`, full `pnpm run lint` (platform), `prettier --check`

## Note

Importing sends a message and navigates into the new thread — that is the
existing behaviour of this command, not something this PR introduces. The panel
closes first, and the import owns the root signal rather than a page signal for
exactly that reason, as the dialog's card does.


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: 4ac0dd54f446d221f25401efb8f21f6f52c61755
PR_AUTO_REVIEW_VERDICT: LGTM

Reviewed HEAD: `4ac0dd54f446d221f25401efb8f21f6f52c61755`
Practice revision: `fd847c0a099446e4dfb43a40c1103ee6269de1ab` (`main`)

### Summary

Three changes to the `/` panel, all behind `composerSlashTemplatePanel`. The Presentation pane leads with an **Import your own deck** card that reuses the picker dialog's command and accepted formats. The cover scroller gains 1px of top/left padding so the cards' outset ring stops being clipped. And illustration covers now render at the catalog's own proportion in a CSS masonry, requested by width alone, instead of being squeezed into the shared 16:9 tile.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | `SlashTemplatePanel` has one caller (`tiptap-workflow-composer.tsx`). `SlashTemplateCover` is new and used only by the detail pane; `isSlashTemplateNativeAspectCategory` has one call site. The import path reuses `importPresentationTemplateDeck$` rather than adding a second upload route. |
| `REVIEW.md` | Regression coverage for a changed flow | Pass | Ten cases in `chat-composer-slash-panel`, plus `chat-composer-presentation-import` (4) and `chat-composer-workflows` (17) unchanged and green, so the dialog's own import path and the flat menu are both still guarded. |
| `REVIEW.md` | Races, ordering, cancellation | Pass | The import owns `rootSignal$`, not a page signal, because it sends and navigates — the same ownership and the same stated reason as the dialog's card. Nothing else is async. |
| `REVIEW.md` | Persisted data / API contract | N/A | No schema or contract change. `SLASH_TEMPLATE_NATIVE_COVER_SIZE` only alters the `/cdn-cgi/image/` directives on a preview URL. |
| `docs/bad-smell.md` §1/§2 | No `any`, no suppressions | Pass | None. `SlashTemplateDetailPane` was kept under the line limit by extracting `SlashTemplateCover`, not by disabling the rule. |
| `docs/bad-smell.md` (new at this revision) | No code whose only output is a log level | N/A | The diff logs nothing. |
| `docs/bad-smell.md` | Smallest necessary abstraction; prefer reuse | Pass | The masonry is not invented here: `chat-composer.tsx:4548` already does `columns-[244px]` with per-tile `aspectRatio: w/h` for the picker dialog, and its comment says it mirrors `www.okou.ai/illustration`. The panel adopts that shape so both entry points show the same artwork. |
| `docs/styles.md` | Tailwind utilities; no first-party selectors or CSS-in-JS | Pass | Utilities only. The one inline style is a computed ratio, which the guide permits for genuinely runtime values and which `chat-composer.tsx:3745` already uses for the same purpose. |
| `docs/styles.md` (new at this revision) | No opacity transitions on hover-revealed affordances; name properties instead of `transition-all` | Pass | The cover animates `transition-transform`, the import tile `transition-colors`. No `transition-all`, no opacity transition. |
| `docs/styles.md` (new at this revision) | Literal colours and gradients | N/A | The diff introduces no colour literal; fills are `bg-muted` / `bg-muted/50` and the hairline is `ring-border/60`. |
| `docs/styles.md` (concentric corners) | Inner radius equals outer minus the gap | Pass | Unchanged from the previous round; rows stay `rounded-lg` in `p-1` gutters. |
| `docs/testing.md` + `docs/testing/testing-external-behavior.md` | Test through the page; no class or internal assertions | Pass | Every case boots the Router, types into the editor, and asserts rendered content, `data-slot` hooks, the sent message, or the `` the browser will actually fetch. No class-name queries. |
| `docs/testing/app-testing.md` (new at this revision) | Default to `toBeInTheDocument()`; no `toBeVisible()` merely to confirm rendering | Pass | The file contains zero `toBeVisible` calls. Each `waitFor` awaits a real state change (pane category swap, message sent), not mere rendering. |
| `docs/fallback.md` | No new fallback; none to declare | Pass | `aspect?:` is absent for categories that have no intrinsic dimensions, which is a real state rather than a default for an impossible one. The pre-existing `cardPreviewImage ?? previewImage` chain is unchanged and already declared in the body. |
| `.claude/skills/ccstate/SKILL.md`, `docs/cache.md` | No store writes in render, no new signals | Pass | `SlashTemplateCover` is presentational. The only new binding is `useSet(importPresentationTemplateDeck$)`, invoked from an event handler. |
| `.claude/skills/feature-switch/SKILL.md` | New user-facing behavior gated | Pass | No new switch; everything is inside the non-GA `composerSlashTemplatePanel`. |
| `CLAUDE.md` | Repository artifacts in English | Pass | Source, comments, commits and body are English. |

### Findings

None.

### Verified during review

The tile declares `aspectRatio` from the style's **reference frame** (`item.width/height`) while the `` renders `cardPreviewImage`, which is a different asset. If those disagreed, this change would have re-introduced cropping through `object-cover` while appearing to fix it. Measured the real files:

| style | declared | actual file | |
| --- | --- | --- | --- |
| emboss-deboss | 0.667 | 512×768 → 0.667 | ✓ |
| ink-storefront | 0.667 | 341×512 → 0.666 | ✓ |
| crowd-ink | 1.500 | 512×341 → 1.501 | ✓ |
| iberian-vignette | 0.671 | 343×512 → 0.670 | ✓ |

They agree to rounding, so `object-cover` trims at most a fraction of a pixel. The card previews are 512px on the long edge, and a card paints 139 CSS px, so the new `width=280` request downscales rather than upscales.

1. **Illustration now downloads more bytes, on purpose.** A 2:3 cover goes from 105×158 to 280×420. That is the fix — the old request produced a picture smaller than the card painted — but the pane does fetch more for this one category. `loading="lazy"` limits it to what is scrolled into view.
2. **The import card is not masonry-safe.** It renders without `break-inside-avoid`, which is harmless today because only `slides` has it and `slides` is a grid. If a future change gives `slides` native aspect, that card would need the same treatment as `SlashTemplateCover`.
3. **CSS columns fill top-to-bottom.** Reading order in the illustration pane becomes column-wise rather than row-wise. That is inherent to the approach and matches the dialog, but it is a real difference from the other three categories.
4. **Three illustration styles are landscape**, so in a 2-column masonry they read as short, wide cards among tall ones. That is the honest shape of the catalog rather than a layout defect.

### Testing

The new aspect test is the valuable one: it asserts the cover URL requests width only and that the tile declares the catalog's own ratio, while a deck cover still requests the 16:9 box. That pins both halves of the defect at once — a shared height silently returns the crop *and* the softness, and neither is visible in a layout-less DOM any other way.

The import test does a real upload through the panel and asserts the deck is sent with the import prompt, mirroring `chat-composer-presentation-import.test.tsx`. It is what caught the focus bug in the first round, where the file input stole focus, cleared the editor's slash range, and unmounted itself before the file dialog could return.

Verification reported in the body matches the changed surfaces: `chat-composer-slash-panel` (10), `chat-composer-presentation-import` (4), `chat-composer-workflows` (17), `chat-composer-create` (19); `check-types` for both tsconfigs; full platform `lint`; `prettier --check`.

No material gap.

tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: 8a76685436d01da461b7d311a409757fc93611ef
PR_AUTO_REVIEW_VERDICT: LGTM

Reviewed HEAD: `8a76685436d01da461b7d311a409757fc93611ef`
Practice revision: `737e752d859154f921d599412c5754ff262078e1` (`main`)

### Summary

Same three product changes as the previous round — the **Import your own deck** card leading the Presentation pane, 1px of scroller padding so the covers' outset ring stops being clipped, and illustration covers rendered at the catalog's own proportion in a CSS masonry — now on top of current `main`. The new commit is the merge itself. Against `main` the PR is still exactly four files.

`main` touched this feature area once since the branch point: #33837, which separated slash-menu hover from keyboard selection. Git merged it textually, and it left two real breaks that this commit repairs.

### What the merge had to fix

| Break | Cause | Repair |
| --- | --- | --- |
| `TS2552: Cannot find name 'selection'` | #33837 reshaped `useSlashTemplatePanelActions` to take a plain `close: () => void` where this branch passed a `{ selectedIndex, setSelectedIndex, close }` object. The auto-merge took `main`'s new `selectTemplate` body but kept this branch's `importDeck`, which still called `selection.close()`. | `importDeck` now calls `close()`. Behavior-identical: at `4ac0dd5` the object's `close` was already `useSet(composer.suggestion.closeSuggestionMenu$)`, and that is the same binding `main` now passes positionally. |
| `max-lines-per-function`: `TiptapWorkflowComposer` at 129 / 128 | Both sides added to the same component — `main` the preview wiring, this branch the `onImportDeck` prop. | Extracted `useComposerPasteHandler`. |

The extraction is a real seam rather than a line-count trick: paste resolution reads the clipboard, gives the host first refusal so it can claim files, and inserts the remainder as Markdown. None of that is rendering, and it was already a self-contained inner function whose only tie to the component was one `useSet`. The component keeps its wiring; the hook keeps the policy. No suppression was added.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | `useComposerPasteHandler` has exactly one caller, and its result is passed to the same `onPasteCapture` as before. `SlashTemplatePanel` still has one caller. The import path still reuses `importPresentationTemplateDeck$` rather than adding a second upload route. |
| `REVIEW.md` | Integration with what `main` changed underneath | Pass | #33837's `previewIndex` / `onPreview(index \| null)` contract survives the merge intact: the pane reads `categories[previewIndex] ?? null`, the workflow list offsets by `categories.length`, and hover no longer moves `selectedIndex`. The masonry change is confined to the cover container and does not touch either index path. |
| `REVIEW.md` | Regression coverage for a changed flow | Pass | `chat-composer-slash-panel` is now 24 cases after absorbing `main`'s additions, including this branch's illustration-aspect case and `main`'s renamed website-preview case; both kept, neither rewritten. |
| `REVIEW.md` | Races, ordering, cancellation | Pass | `importDeck` still closes first and then detaches against `rootSignal$`, because it sends and navigates. Ordering is unchanged by the merge — only the callee's name changed. |
| `REVIEW.md` | Persisted data / API contract | N/A | No schema or contract change. |
| `docs/bad-smell.md` §1/§2 | No `any`, no suppressions | Pass | None, including for the line-limit failure. |
| `docs/bad-smell.md` | Smallest necessary abstraction | Pass | The masonry mirrors `chat-composer.tsx:4548`, which already pairs `columns-*` with a per-tile `aspectRatio`. The paste hook is a move, not a new layer — no behavior added. |
| `docs/styles.md` | Tailwind utilities; no first-party selectors or CSS-in-JS | Pass | Utilities only; the single inline style is the computed ratio, which the guide permits for genuinely runtime values. |
| `docs/styles.md` | Name transitioned properties; no opacity transition on hover-revealed affordances | Pass | `transition-transform` on the cover, `transition-colors` on the import tile. |
| `docs/styles.md` (concentric corners) | Inner radius equals outer minus the gap | Pass | Unchanged; rows stay `rounded-lg` in `p-1` gutters. |
| `docs/testing.md`, `docs/testing/testing-external-behavior.md` | Test through the page; no class or internal assertions | Pass | Every case boots the Router, types into the editor, and asserts rendered content, `data-slot` hooks, the sent message, or the `` the browser will fetch. |
| `docs/testing/app-testing.md` | Default to `toBeInTheDocument()` | Pass | Zero `toBeVisible` calls; each `waitFor` awaits a real state change. |
| `docs/fallback.md` | No new fallback; none to declare | Pass | `aspect?:` is absent for categories with no intrinsic dimensions — a real state, not a default for an impossible one. |
| `.claude/skills/ccstate/SKILL.md`, `docs/cache.md` | No store writes in render, no new signals | Pass | `useComposerPasteHandler` calls `useSet` unconditionally at the top and returns a closure; the store is touched only from the paste event. No signal was added. |
| `.claude/skills/feature-switch/SKILL.md` | New user-facing behavior gated | Pass | Everything remains inside the non-GA `composerSlashTemplatePanel`. |
| `CLAUDE.md` | Repository artifacts in English | Pass | Source, comments, and the merge commit body are English. |

### Findings

None.

### Verified during review

**The `close()` substitution is not a guess.** `git show 4ac0dd5` confirms the pre-merge object literal was `{ selectedIndex, setSelectedIndex, close }` built from `const close = useSet(composer.suggestion.closeSuggestionMenu$)` — the identical binding `main` now passes as the third argument. So `importDeck` calls the same command it always did, and the panel dismisses on import exactly as before.

**The masonry classes are applied per-tile, not per-container only.** `SlashTemplateCover` adds `mb-2.5 block w-full break-inside-avoid` only when `aspect` is present, so the tiles that land in `columns-2` are the ones told not to split across a column. The grid categories keep `aspect-video` and inherit nothing from the masonry branch.

**Test intent survived the conflict.** The conflicted hunk kept this branch's illustration-aspect assertion *and* adopted `main`'s renamed website-preview case rather than dropping either side.

1. **The import card is still not masonry-safe** (carried from the previous round). It renders without `break-inside-avoid`, harmless today because only `slides` shows it and `slides` is a grid. If `slides` ever gains native aspect, that card needs the same treatment `SlashTemplateCover` now has.
2. **`cn()` wraps a bare ternary** in the cover container — it has no second argument to merge, so the call is doing nothing a template string would not. Cosmetic.
3. **CSS columns fill top-to-bottom**, so reading order in the illustration pane is column-wise. Inherent to the approach and consistent with the picker dialog.
4. **Illustration downloads more bytes, on purpose.** A 2:3 cover goes from 105×158 to 280×420 — the old request produced a picture smaller than the card painted. `loading="lazy"` bounds it to what is scrolled into view.

### Testing

`check-types` (all three tsconfigs), full platform `lint`, and `prettier --check` all pass on the merged tree.

Suites, each run in isolation: `chat-composer-slash-panel` 24/24, `chat-composer-presentation-import` 4/4, `chat-composer-workflows` 17/17, `chat-composer-create` 19/19.

One note on method, because it produced a misleading intermediate result. Running all four together at `--maxWorkers=4` reported six timeouts at the 5000ms budget across all four files. Re-running each file alone at `--maxWorkers=1` passed every case, and the failures were spread across files untouched by this diff, including ones the branch does not import. That is local resource contention between four heavy jsdom suites, not a defect — and the correct response is to run them sequentially rather than to raise the timeout, which `CLAUDE.md` names explicitly.

No material gap.

PR #33540, opened 2026-09-11
fix(api): raise morning brief provisioning outcome logs to info

lancy:
Closes #33538

## Problem

Morning Brief auto-enrollment shipped in #33434 and works, but its **success and skip outcomes are invisible in production**.

The Axiom transport is constructed with an explicit threshold at `turbo/apps/api/src/lib/log.ts:162` (`logLevel: "info"`), and the comment above it states the rule: *"A signal that must be observable in production has to be emitted at info or above."* Records below that threshold never reach the dataset.

Every Morning Brief provisioning path logs failures at `warn`/`error` but logs the successful and skipped outcome at `debug`. So production can show that enrollments **failed**, but not how many succeeded, nor why any were skipped — and the skip reasons (`missing-data-source`, `missing-timezone`, `missing-default-agent`) are exactly the product signal we need.

## Change

Raise the four success/skip call sites from `debug` to `info`:

| File | Message |
| --- | --- |
| `signals/services/morning-brief-enrollment-worker.service.ts` | `"Morning Brief enrollment changed"` |
| `signals/routes/webhooks-clerk.ts` | `"Morning Brief membership provisioning outcome"` |
| `signals/routes/user-preferences.ts` | `"Morning Brief timezone provisioning outcome"` |
| `signals/routes/user-preferences.ts` | `"Morning Brief initialization outcome"` |

`signals/services/onboarding.service.ts:268` already emits `L.info` for this same family of events and is the precedent followed here; it is untouched.

## Why this shape

- `info` is the documented minimum for a production-observable signal in this codebase.
- Each site already builds a single `details` object shared by its failure branch and its success branch, so raising the level exposes **no new fields** and no new personal data beyond what the `warn` branch already emits. The source diff is four level tokens and nothing else.
- The `api/no-logger-info` lint rule bans `logger().info` in API source unless the file declares the exact static message in `allowedMessages`. That allowlist is why `onboarding.service.ts` can already emit this family at info, so `turbo/apps/api/eslint.config.mjs` gains one purely additive entry per file with the exact message strings. The rule stays `"error"` everywhere; every other `logger().info` in these three files is still rejected.
- Volume is bounded by user actions, not polling: the worker only logs on first attempt, on a changed error, or on an actual install — not every cron tick; the Clerk webhook fires once per organization-membership creation; the two `user-preferences.ts` sites fire on the timezone `initialize` call, invoked once per authenticated session.

[...]

PR #33895, opened 2026-09-14
fix(api): bind morning brief enrollment to its installation

### LINKED ISSUE #33766: fix(api): morning brief enters a permanent conflict when installed on two agents

lancy:
## Summary

Closes #33766. A member could end up with two Morning Brief installations and
land in a permanent, non-self-healing conflict: the read path, auto-provisioning
and the user toggle all returned `MORNING_BRIEF_MULTIPLE_INSTALLATIONS`, on
every login, forever.

The lookup was agent-agnostic while everything beneath it is agent-scoped, so
a catalog install on a second Agent — a supported capability — produced two
legal rows that the preference layer could not tell apart.

Per the issue's Decision: **storage stays per Agent, the product surface owns
exactly one.** `morning_brief_enrollments` now records the installation it owns
and every preference read and write resolves through it.

## Changes

- **Schema / migration `1117`.** `morning_brief_enrollments` gains a nullable
  `workflow_id`. Existing rows are backfilled with the adoption rule: the
  installation on the org's current default Agent, otherwise the oldest. The
  backfill is idempotent (`workflow_id IS NULL` guard) and takes no table lock.
- **Ownership write point.** `completeMorningBriefEnrollment` binds the
  enrollment to the installed workflow, so ownership follows the Agent the
  brief was installed on rather than whichever Agent is the org default today.
- **Resolution.** `loadMorningBriefOwnership` returns the owned installation and
  is fault tolerant: a row written before the column existed, a row whose
  installation was uninstalled, and an installation with no enrollment row all
  resolve through the same adoption rule.
- **All five call sites** of the former `loadMorningBriefWorkflowIds` now resolve
  through the owned installation — the read path, the post-install race
  re-check, auto-enrollment, the user toggle, and the timezone sync — so no path
  stays agent-agnostic.

A catalog install on another Agent becomes invisible to Preferences, which is
the intended outcome. It is not deleted, disabled or migrated, and keeps running.

## Explicitly unchanged

`idx_workflows_private_owner_agent_name_unique`, catalog installs on non-default
Agents, eligibility rules, the enrollment state machine, the 07:00 schedule, the
`morningBrief` / `officialWorkflows` switches, and the log levels and message
strings from #33540. `MORNING_BRIEF_MULTIPLE_INSTALLATIONS` stays in the contract
and in the App for API/App version skew; the current API no longer returns it.

## Verification

- `pnpm -F api check-types`, `pnpm -F api lint`, `pnpm -F @okouai/db check-types`,
  `pnpm -F @okouai/db lint`, `pnpm -F @okouai/api-contracts check-types`,
  `pnpm -F @okouai/api-contracts lint` — pass.
- `tsx scripts/test-migration-consistency-schema.ts` against a real PostgreSQL —
  pass, including snapshot chain, journal ordering and schema equivalence.
- Migration `1117` applied to a real PostgreSQL and the backfill verified against
  seeded rows: default-Agent preference, oldest fallback when no install is on the
  default Agent, no `org_metadata` row, an enrollment with no installation, an
  enrollment that already owns one, and a second run reporting `UPDATE 0`.
- `pnpm -F @okouai/db test:morning-brief-enrollment-installation` — the new
  migration validator replays `1112` and `1117` in a temporary schema against
  seeded pre-migration rows and pins the adoption rule. Mutation-checked: it
  fails when the default-Agent preference, the oldest-wins ordering, or the
  idempotency guard is broken.
- `vitest run official-workflows.test.ts -t "Morning Brief"` — 19 passed. The
  three new regression tests were confirmed to fail with `409
  MORNING_BRIEF_MULTIPLE_INSTALLATIONS` against the pre-fix service.
- `cron-official-workflow-catalog.test.ts`, `email-subscription.test.ts` — pass.
- The `Official Workflow Run admission` block of `official-workflows.test.ts`
  fails identically (33 tests) on unmodified `main` in this sandbox; it needs
  queue infrastructure that is not available locally. Left to CI.

## Regression tests

1. `adopts the default Agent installation when installations exist across Agents`
   — an existing conflicting pair resolves to the default-Agent installation, the
   other one keeps running, and both remain installed.
2. `keeps the preference on its own installation after a catalog install on
   another Agent` — the Preferences card stays usable and manages only the
   product installation.
3. `leaves an installed brief in place when the org default Agent changes` — the
   brief does not move, duplicate or orphan; logging in does not re-provision;
   ownership stays on the original Agent even once the new default Agent carries
   a brief of its own.

## Sequencing

Dispatched in parallel with #33834 by explicit decision. Whichever merges first
establishes the new `main`; this branch will rebase and keep #33834's deletion of
the data-source gate.


lancy:
Changes Requested


PR_AUTO_REVIEW_COMMIT: b2d57ee32ae27a3643457846b6dc19961526d833

Inspected HEAD: `b2d57ee32ae27a3643457846b6dc19961526d833`
Practice revision: `5f1906df048cab1473ee37b55c193e78db649da4` (`main`)

### Summary

`morning_brief_enrollments` gains a nullable `workflow_id`, `completeMorningBriefEnrollment`
binds the enrollment to the installation it created, and all five former
`loadMorningBriefWorkflowIds` call sites resolve through a single
`loadMorningBriefOwnership` helper instead of an agent-agnostic lookup. Per-Agent
installation and `idx_workflows_private_owner_agent_name_unique` are unchanged, so a
catalog install on a second Agent stops being read back as an unrecoverable conflict.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | All five call sites in `morning-brief-preference.service.ts` now route through `loadMorningBriefOwnership`; the remaining consumers (`user-preferences.ts`, `morning-brief-preference.ts`, `morning-brief-enrollment-worker.service.ts`, `test-workflow-automation-execution.ts`) call the unchanged exported commands. No other caller of `completeMorningBriefEnrollment` exists. |
| `REVIEW.md` | Races, cancellation, ordering, permissions, failure outcomes | Pass | All four `completeMorningBriefEnrollment` writers still run under `pg_try_advisory_xact_lock`. `signal.throwIfAborted()` follows every new await. Resolution stays scoped to `orgId` + `ownerUserId` + `visibility = 'private'`, so it cannot adopt another member's installation. |
| `REVIEW.md` | Meaningful coverage for new behavior and regression fixes | **Finding (P1)** | The runtime adoption rule is covered by three new endpoint tests, but the migration `1117` backfill that applies the same rule to existing rows has no automated coverage. See Findings. |
| `REVIEW.md` | Tests verify externally observable state, not internals or logs | Pass | New assertions go through `GET/PUT /api/preferences/morning-brief`, `GET /api/official-workflows/installations/:workflowId` and the workflows collection. No log, service-return, or DB-row assertions were added. |
| `REVIEW.md` | Regression coverage and containment for a rerouted existing flow | Pass | The retired `fails closed when generic installations already exist across Agents` case is replaced by the new behavior rather than kept alongside it. `officialWorkflows` stays staff-only and `morningBrief` stays `enabled: true`; neither switch value is touched. |
| `REVIEW.md` | Review schema, API contract, and response validation together | Pass | `morning-brief-enrollment.ts`, migration `1117`, `1117_snapshot.json` and the contract were reviewed together; the response schema is unchanged and the error-code enum only gains a comment. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | `pnpm -F api check-types` and `pnpm -F api lint` are clean; the diff adds no suppression comment and no `any`. |
| `docs/bad-smell.md` | No fabricated default that hides a broken invariant | Pass | Resolution never invents an installation: it returns `undefined` when the member owns none, and only chooses among installations the member actually owns. |
| `docs/bad-smell.md` | Document changed public contracts and assess consumers | Pass | `MORNING_BRIEF_MULTIPLE_INSTALLATIONS` is retained in the contract with a comment; `morning-brief-settings.tsx` keeps handling it, which is what an App served by an older API still needs. |
| `docs/fallback.md` | No fallback for a state the contract, schema, or single writer prevents | Pass | The null/stale `workflow_id` branch is not a fallback under this document: `qualifyMorningBriefMembership$` and `recordMorningBriefChoice` legitimately insert enrollments with no installation, and uninstall does not clear the column. That is §6(3) "genuinely nullable persisted columns", permanent by design with no removal gate, so §9 requires no `Fallbacks` section. |
| `docs/fallback.md` | No negative tests pinning removed behavior | Pass | The old conflict test was rewritten to the replacement behavior instead of being kept as a tombstone. |
| `docs/fallback.md` | Non-GA switch needs no compatibility work | N/A | `FeatureSwitchKey.MorningBrief` is `enabled: true` (`feature-switch.ts:242-247`), so the feature is GA and the §2 exemption does not apply; the backfill is required. |
| `docs/testing.md` | Test at the API boundary through production endpoints | Pass | Every new case drives installs through the catalog route or the preference toggle and asserts HTTP responses. |
| `docs/testing/testing-external-behavior.md` | Leave the boundary only for states the production interface cannot construct, and state why | **Finding (P1, minor part)** | `setOrgDefaultAgentFixture` is justified — no production route repoints `org_metadata.default_agent_id` (only `onboarding.service.ts` writes it) — but the rationale lives only in the fixture docblock, not at the test call site. Folded into the repair below. |
| `docs/testing/api-testing.md` | `setupApp` + contract, uniquely owned identities under shared persistent state | Pass | New tests reuse `morningBriefPreferenceClient()`/`officialClient()`/`installationClient()` and per-test orgs from `workflowBdd.setupWorkflowOrg()`; the default-Agent fixture is restored and the extra Agent deleted in `onTestFinished`. |
| `docs/deployment-compatibility.md` | Old code after migration stays legal | Pass | `1117` only adds a nullable column with no default. The previous API's Drizzle-generated `SELECT`/`INSERT ... ON CONFLICT`/`UPDATE` column lists on `morning_brief_enrollments` remain valid. |
| `docs/deployment-compatibility.md` | New code before migration must not require the new column | Pass | `loadMorningBriefEnrollment` uses `db.select()`, which expands to include `workflow_id`, so the new API requires `1117`. `promote-api-production` migrates before promoting the artifact, which closes this direction for a normal release — the same ordering `1112` already relies on for this table. |
| `docs/deployment-compatibility.md` | Persisted-state changes cover old/new combinations | Partial | Rows written by the previous version (enrollment present, `workflow_id` null) are exercised at runtime by the two conflicting-pair tests. The migration step itself is not exercised — see Findings. |
| `.claude/skills/database-development/SKILL.md` | Use Drizzle to generate migration metadata; never hand-edit journal or snapshot | Pass | `1117` and `1117_snapshot.json` were produced by `drizzle-kit generate --name=...`; only the migration body was extended with the backfill, which is the documented custom-SQL workflow. |
| `.claude/skills/database-development/references/query-contracts.md` | Structured selections use schema columns or reviewed decoders; builder-first SQL | Pass | `loadMorningBriefInstallations` selects schema columns only and orders with `asc(...)`. `setOrgDefaultAgentFixture` uses `sql\`now()\`` solely as a write value and returns a schema column, matching `setOnboardingPaymentPendingFixture` in the same file. |
| `turbo/packages/db/MIGRATIONS.md` | Migration numbering, default timeouts, no table lock, no trigger/function inventory change | Pass | `1117` is the next free number on `main` (highest is `1116`). It adds no trigger or function, takes no `LOCK TABLE`, and its `ALTER` plus small `UPDATE` fit the default `1s`/`10s` transactional limits. |
| `.claude/skills/ccstate/SKILL.md` | Async ownership and cancellation inside commands | Pass | Derivation stays in plain helpers, semantic actions stay in `command`, and every new await inside a command is followed by `signal.throwIfAborted()`. No `detach()`/`void` introduced. |
| `docs/externally-managed-references.md` | Preference resolution must be explicit and independently authorized, never a silent identity substitution | Pass | The adoption rule is written down in the schema comment, the resolver docblock and the migration comment, and it can only select installations already scoped to the same org and owner. A stale `workflow_id` degrades to that documented rule rather than to an arbitrary row. |

### Findings

#### P1 — Migration `1117`'s backfill has no automated coverage

`turbo/packages/db/src/migrations/1117_morning_brief_enrollment_installation.sql:7-37`

The backfill is the mechanism that satisfies the issue's acceptance criterion
"existing rows are backfilled using the adoption rule above", and it is the only
thing that repairs the currently broken production user. Its `DISTINCT ON`
ordering encodes three decisions that nothing in the suite checks: prefer the
installation on `org_metadata.default_agent_id`, otherwise take the oldest by
`created_at`, and never overwrite a row that already owns an installation.

Trigger: a future edit — or a reordered `ORDER BY`, a `LEFT JOIN` turned inner, a
dropped `IS NULL` guard — changes which installation is adopted.

Consequence: a wrong adoption is **persisted and not self-healing**, because
`loadMorningBriefOwnership` honours a recorded `workflow_id` ahead of the adoption
rule. Preferences would then manage the wrong brief for that member, and the
runtime tests would still pass. A dropped `IS NULL` guard would additionally
repoint members who already own an installation.

`docs/testing.md` requires coverage for persisted-state changes, and
`docs/deployment-compatibility.md` ("Testing Expectations") asks to populate the
pre-migration schema, upgrade it, and assert the outcome. The repository already
has this exact pattern: `turbo/packages/db/scripts/test-reusable-ssh-credentials.ts`
replays migration `1113` in a temporary schema against seeded pre-migration rows
and is wired into `test:migration-consistency`.

Add an equivalent validator covering: default-Agent preference over an older
install, oldest-wins when no install is on the default Agent, an enrollment with no
installation left null, an already-owned enrollment left untouched, a missing
`org_metadata` row, and a second run reporting zero updates.

#### P1 (minor) — The external-behavior exception is not stated at the test call site

`turbo/apps/api/src/signals/routes/__tests__/official-workflows.test.ts` — the
`setOrgDefaultAgentFixture` calls in `leaves an installed brief in place when the org
default Agent changes`.

The exception itself is legitimate: no production route repoints
`org_metadata.default_agent_id`, so the case cannot be constructed through the API.
But `docs/testing/testing-external-behavior.md` asks the *test* to state why the
production interface cannot construct the state, and warns against hiding exceptions
inside generic fixtures. Add a one-line comment at the call site.

### Testing

Coverage added is well targeted and asserts behaviour rather than names:

- `adopts the default Agent installation when installations exist across Agents` —
  the pre-existing conflicting pair now resolves; the adopted brief follows the
  toggle while the other one stays enabled and installed.
- `keeps the preference on its own installation after a catalog install on another
  Agent` — the Preferences card stays usable and only the product installation
  changes state.
- `leaves an installed brief in place when the org default Agent changes` — the
  brief does not move, duplicate, or orphan; logging in does not re-provision; and
  ownership stays on the original Agent even after the new default Agent carries a
  brief of its own, which is the assertion that actually distinguishes recorded
  ownership from a "current default Agent" implementation.

The PR body records that all three fail with `409
MORNING_BRIEF_MULTIPLE_INSTALLATIONS` against the pre-fix service, which is the
right regression proof. `cron-official-workflow-catalog.test.ts` and
`email-subscription.test.ts` pass. Package lint, type checks and
`test-migration-consistency-schema.ts` pass.

Material gap: the migration backfill, covered in Findings.

Non-blocking observations:

- `EnsureMorningBriefDefaultEnabledResult.installationCount` can now report more
  than `1` for the `unchanged` outcome. It is only an informational log field with
  no in-repo consumer, and the true count is more useful than a constant, so this
  reads as intended rather than as a regression.
- Ownership is persisted only through `completeMorningBriefEnrollment`, which is
  gated on `state IN ('checking','pending')`. In practice the next auto-enrollment
  pass records it, and unrecorded ownership resolves through the same rule, so this
  is not a defect — just worth knowing when reading the column.

lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: ef6502273d6815d79e5f4c161617ffedee6e90a4
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `ef6502273d6815d79e5f4c161617ffedee6e90a4`
Practice revision: `5f1906df048cab1473ee37b55c193e78db649da4` (`main`)

### Summary

`morning_brief_enrollments` gains a nullable `workflow_id`, `completeMorningBriefEnrollment`
binds the enrollment to the installation it created, and all five former
`loadMorningBriefWorkflowIds` call sites resolve through a single
`loadMorningBriefOwnership` helper instead of an agent-agnostic lookup. Migration
`1117` backfills existing rows with the adoption rule and is now pinned by a
dedicated validator. Per-Agent installation and
`idx_workflows_private_owner_agent_name_unique` are unchanged.

Both P1 findings from the review of `b2d57ee3` are resolved by `ef650227`.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | All five call sites route through `loadMorningBriefOwnership`; remaining consumers (`user-preferences.ts`, `morning-brief-preference.ts`, `morning-brief-enrollment-worker.service.ts`, `test-workflow-automation-execution.ts`) use the unchanged exported commands. No other caller of `completeMorningBriefEnrollment` exists. |
| `REVIEW.md` | Races, cancellation, ordering, permissions, failure outcomes | Pass | All four `completeMorningBriefEnrollment` writers stay under `pg_try_advisory_xact_lock`; `signal.throwIfAborted()` follows every new await; resolution is scoped to `orgId` + `ownerUserId` + `visibility = 'private'`, so it cannot adopt another member's installation. |
| `REVIEW.md` | Meaningful coverage for new behavior and regression fixes | Pass | Three endpoint regression tests for the runtime rule, plus `scripts/test-morning-brief-enrollment-installation.ts` for the migration backfill. |
| `REVIEW.md` | Tests verify externally observable state, not internals or logs | Pass | API assertions go through `GET/PUT /api/preferences/morning-brief`, the installation detail route and the workflows collection. No log, service-return, or DB-row assertion in the API suite. The new validator is a migration script under `packages/db/scripts`, not an API test. |
| `REVIEW.md` | Regression coverage and containment for a rerouted existing flow | Pass | The retired conflict case is replaced rather than kept alongside. `officialWorkflows` stays staff-only, `morningBrief` stays `enabled: true`; no switch value changes. |
| `REVIEW.md` | Review schema, API contract, and response validation together | Pass | Schema, migration, snapshot and contract reviewed together; the response schema is unchanged and the error-code enum only gains a comment. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | `pnpm -F api check-types`, `pnpm -F api lint`, `pnpm -F @okouai/db check-types`, `pnpm -F @okouai/db lint` clean; no suppression comment added. |
| `docs/bad-smell.md` | No fabricated default that hides a broken invariant | Pass | Resolution returns `undefined` when the member owns no installation and only ever selects among installations that member owns. |
| `docs/bad-smell.md` | No code path nobody executes | Pass | `ef650227` removes the `workflow_id IS NULL` predicate on the UPDATE target that the CTE filter already made unreachable; mutation checks confirm the surviving guard is the one that carries idempotency. |
| `docs/bad-smell.md` | Document changed public contracts and assess consumers | Pass | `MORNING_BRIEF_MULTIPLE_INSTALLATIONS` is retained in the contract with a comment and still handled in `morning-brief-settings.tsx`, which is what an App served by an older API needs. |
| `docs/fallback.md` | No fallback for a state the contract, schema, or single writer prevents | Pass | The null/stale `workflow_id` branch is not a fallback: `qualifyMorningBriefMembership$` and `recordMorningBriefChoice` legitimately insert enrollments with no installation, and uninstall does not clear the column. That is §6(3) "genuinely nullable persisted columns", permanent with no removal gate, so §9 requires no `Fallbacks` section. |
| `docs/fallback.md` | No negative tests pinning removed behavior | Pass | The old conflict test was rewritten to the replacement behavior; no tombstone remains. |
| `docs/fallback.md` | Non-GA switch needs no compatibility work | N/A | `FeatureSwitchKey.MorningBrief` is `enabled: true` (`feature-switch.ts:242-247`), so the feature is GA and the §2 exemption does not apply; the backfill is required. |
| `docs/testing.md` | Test at the API boundary through production endpoints | Pass | Every new API case drives installs through the catalog route or the preference toggle and asserts HTTP responses. |
| `docs/testing.md` | Persisted-state changes cover the old/new combinations | Pass | Runtime: rows in the previous shape (enrollment present, no recorded owner) are exercised by both conflicting-pair tests. Migration: the validator seeds the pre-`1117` schema, upgrades it, and asserts the resulting ownership. |
| `docs/testing/testing-external-behavior.md` | Leave the boundary only when unavoidable, and state why | Pass | `setOrgDefaultAgentFixture` now carries the reason at the call site as well as in the fixture docblock: only the Clerk org-creation bootstrap writes `org_metadata.default_agent_id`, so no production endpoint can repoint an existing org. |
| `docs/testing/api-testing.md` | `setupApp` + contract, uniquely owned identities under shared persistent state | Pass | New tests reuse the existing contract clients and per-test orgs from `workflowBdd.setupWorkflowOrg()`; the default Agent is restored and the extra Agent deleted in `onTestFinished`. The validator runs in a per-run temporary schema inside a transaction that always rolls back. |
| `docs/deployment-compatibility.md` | Old code after migration stays legal | Pass | `1117` only adds a nullable column with no default; the previous API's Drizzle-generated `SELECT` / `INSERT ... ON CONFLICT` / `UPDATE` column lists on `morning_brief_enrollments` remain valid. |
| `docs/deployment-compatibility.md` | New code before migration must not require the new column | Pass | `loadMorningBriefEnrollment` uses `db.select()`, so the new API requires `1117`. `promote-api-production` migrates before promoting the artifact, closing this direction for a normal release — the same ordering `1112` already relies on for this table. |
| `docs/deployment-compatibility.md` | Populate the pre-migration schema, upgrade it, assert the outcome | Pass | That is exactly what `scripts/test-morning-brief-enrollment-installation.ts` does, wired into `test:migration-consistency`. |
| `.claude/skills/database-development/SKILL.md` | Use Drizzle to generate migration metadata; never hand-edit journal or snapshot | Pass | `1117` and `1117_snapshot.json` come from `drizzle-kit generate --name=...`; only the migration body was extended, which is the documented custom-SQL workflow. `test-migration-consistency-schema.ts` passes against a real PostgreSQL. |
| `.claude/skills/database-development/references/query-contracts.md` | Structured selections use schema columns or reviewed decoders; builder-first SQL | Pass | `loadMorningBriefInstallations` selects schema columns and orders with `asc(...)`. `setOrgDefaultAgentFixture` uses `sql\`now()\`` only as a write value and returns a schema column, matching `setOnboardingPaymentPendingFixture`. The validator is a migration script using `pg` directly, outside the Drizzle result boundary. |
| `turbo/packages/db/MIGRATIONS.md` | Numbering, default timeouts, no table lock, unchanged trigger/function inventory | Pass | `1117` is the next free number on `main` (highest is `1116`), adds no trigger or function, takes no `LOCK TABLE`, and its `ALTER` plus small `UPDATE` fit the default `1s`/`10s` transactional limits. It is a one-shot backfill validator like `test-reusable-ssh-credentials.ts`, not an expand→contract transition validator, so it is correctly absent from the "Active transition validators" list. |
| `.claude/skills/ccstate/SKILL.md` | Async ownership and cancellation inside commands | Pass | Derivation stays in plain helpers, semantic actions stay in `command`, every new await inside a command is followed by `signal.throwIfAborted()`, and no `detach()`/`void` is introduced. |
| `docs/externally-managed-references.md` | Preference resolution must be explicit and authorized, never a silent identity substitution | Pass | The adoption rule is written down in the schema comment, the resolver docblock and the migration comment, and can only select installations already scoped to the same org and owner. A stale `workflow_id` degrades to that documented rule, not to an arbitrary row. |

### Findings

None.

Both P1 findings against `b2d57ee3` are resolved:

- The migration backfill now has a validator that replays `1112` and `1117` in a
  temporary schema. It was mutation-checked: inverting the default-Agent
  preference, dropping `created_at` from the ordering, and removing the
  idempotency guard each make it fail.
- The external-behavior exception for `setOrgDefaultAgentFixture` is now stated at
  the test call site.

### Testing

- Runtime rule, through the API: `adopts the default Agent installation when
  installations exist across Agents`, `keeps the preference on its own
  installation after a catalog install on another Agent`, and `leaves an installed
  brief in place when the org default Agent changes`. The third one asserts that
  ownership stays on the original Agent even after the new default Agent carries a
  brief of its own, which is what distinguishes recorded ownership from a "current
  default Agent" implementation. All three fail with `409
  MORNING_BRIEF_MULTIPLE_INSTALLATIONS` against the pre-fix service.
- Backfill: `test:morning-brief-enrollment-installation` pins default-Agent
  preference over an older install, oldest-wins with no install on the default
  Agent and with no `org_metadata` row at all, an enrollment without an
  installation left unowned, a private non-Morning-Brief workflow and a public one
  excluded, another member's installation never adopted and no row created for
  them, an already-owned enrollment left untouched, and a zero-row re-run.
- `official-workflows.test.ts -t "Morning Brief"` 19 passed;
  `cron-official-workflow-catalog.test.ts` and `email-subscription.test.ts` pass;
  package lint and type checks pass; `db:migrate` and
  `test-migration-consistency-schema.ts` pass against a real PostgreSQL.
- The `Official Workflow Run admission` block of `official-workflows.test.ts`
  fails identically (33 tests) on unmodified `main` in the author's sandbox for
  lack of queue infrastructure; that block is unrelated to this diff and is left
  to CI.

Non-blocking observations, unchanged from the previous review:

- `EnsureMorningBriefDefaultEnabledResult.installationCount` can now report more
  than `1` for the `unchanged` outcome. It is an informational log field with no
  in-repo consumer, and the true count is more useful than a constant.
- Ownership is persisted only through `completeMorningBriefEnrollment`, which is
  gated on `state IN ('checking','pending')`. The next auto-enrollment pass records
  it, and unrecorded ownership resolves through the same rule, so this is not a
  defect — just worth knowing when reading the column.

PR #33942, opened 2026-09-14
perf(guest-agent): bound cascading json key masking

### LINKED ISSUE #33924: perf(rust): avoid quadratic rescanning of cascading masked JSON keys
## Summary

`SecretMasker::fully_masked_key` can rescan and copy an almost unchanged JSON key once per two input bytes. One valid configured secret and one event key are enough to produce quadratic CPU/copying work during synchronous CLI event preparation.

Audited `main`: `0f2e7c8e3147cc1d10da7e57f24c2d9007523ce4`.

## Location and cause

- [`crates/guest-agent/src/masker.rs:371-378`](https://github.com/vm0-ai/vm0/blob/0f2e7c8e3147cc1d10da7e57f24c2d9007523ce4/crates/guest-agent/src/masker.rs#L371-L378): the cascading-redaction loop repeatedly calls `masked_string` until there are no more matches. Strictly decreasing length guarantees termination, but does not bound total work linearly.
- [`masker.rs:336-342`](https://github.com/vm0-ai/vm0/blob/0f2e7c8e3147cc1d10da7e57f24c2d9007523ce4/crates/guest-agent/src/masker.rs#L336-L342) and [`630-648`](https://github.com/vm0-ai/vm0/blob/0f2e7c8e3147cc1d10da7e57f24c2d9007523ce4/crates/guest-agent/src/masker.rs#L630-L648): each successful pass searches the current key and allocates/copies a replacement.

With synthetic secret `a***b` and key `"a".repeat(n) + "***" + "b".repeat(n)`, exactly one central `a***b` matches each pass. Replacing it with `***` recreates the same pattern one position farther outward. There are n passes over lengths approximately 2n, 2n-2, ..., 5: O(n²) scanned/copied bytes. No collision with another object key is required.

Use the unchanged public masker:

```rust
use base64::Engine;
use guest_agent::masker::SecretMasker;
use serde_json::{Map, Value};
use std::time::Instant;

fn main() {
    let n: usize = std::env::args().nth(1).unwrap().parse().unwrap();
    let raw = base64::engine::general_purpose::STANDARD.encode("a***b");
    let masker = SecretMasker::from_raw(&raw);
    let mut object = Map::new();
    object.insert(format!("{}***{}", "a".repeat(n), "b".repeat(n)), Value::Null);
    let mut event = Value::Object(object);
    let started = Instant::now();
    masker.mask_value(&mut event);
    println!("{:?}", started.elapsed());
    assert_eq!(event, serde_json::json!({"***": null}));
}
```

The audit measured the **exact unmodified `masker.rs` loaded by path** in a separate release-mode harness, with the repository's dependency versions/features (`aho-corasick 1.1.5`, `base64 0.23.1`, `serde_json 1.0.151`) and Rust 1.98.1 on x86_64 Linux. Only the unused `from_config` constructor's config type was stubbed; `from_raw`, matching, JSON traversal, and redaction were the actual source. Input construction and matcher compilation were excluded from timing. This was a synthetic local benchmark, not an observed production incident.

| n | Serialized input bytes | Masking time |
| ---: | ---: | ---: |
| 4,096 | 8,204 | 17.128 ms |
| 8,192 | 16,396 | 66.113 ms |
| 16,384 | 32,780 | 266.182 ms |
| 32,768 | 65,548 | 1,039.398 ms |

Doubling the key length approximately quadruples masking time. Run larger reproductions under an external process watchdog; a Tokio timeout cannot preempt this synchronous loop.

## Reachability

- `a***b` meets the five-byte minimum and is accepted by `SecretMasker::from_raw`; [`runner/src/executor/env.rs:1091-1102`](https://github.com/vm0-ai/vm0/blob/0f2e7c8e3147cc1d10da7e57f24c2d9007523ce4/crates/runner/src/executor/env.rs#L1091-L1102) encodes supplied secret values without excluding this configuration.
- [`guest-agent/src/events.rs:92-105`](https://github.com/vm0-ai/vm0/blob/0f2e7c8e3147cc1d10da7e57f24c2d9007523ce4/crates/guest-agent/src/events.rs#L92-L105) recursively masks event-controlled keys. The key can be nested inside an otherwise ordinary event.
- [`guest-agent/src/cli/mod.rs:797-818`](https://github.com/vm0-ai/vm0/blob/0f2e7c8e3147cc1d10da7e57f24c2d9007523ce4/crates/guest-agent/src/cli/mod.rs#L797-L818) does this synchronously before delivery admission. The ordinary stdout record cap is 16 MiB ([`stdout_framing.rs:11`](https://github.com/vm0-ai/vm0/blob/0f2e7c8e3147cc1d10da7e57f24c2d9007523ce4/crates/guest-contracts/src/stdout_framing.rs#L11)); the measured input is only about 64 KiB. Delivery request limits are applied after masking, so they do not prevent the CPU stall.

The affected event cannot enter delivery until this work finishes, and its executor thread has no await/cancellation point inside the loop. This finding is specifically about cascades in keys under accepted secret configurations, not ordinary string-value masking.

## Acceptance criteria

- Bound total cascading-key work as a function of input size; use a progress-aware algorithm or a bounded-work, secret-free fallback rather than rescanning the entire key once per tiny reduction.
- Preserve all values, unmatched keys, deterministic key uniqueness, and the guarantee that generated keys do not reintroduce configured secrets. Do not remove masking or silently overwrite entries.
- Add an externally bounded regression through event preparation for the single-key family above, plus nested and existing cascade/collision behavior. Demonstrate bounded scaling with operation-count evidence or a controlled benchmark; do not rely only on a flaky elapsed-time assertion.

All 47 existing masker unit tests passed on the audited commit. No repository source changes were made.

seven332:
## Summary

A valid secret such as `a***b` can make JSON key masking repeatedly scan and copy an almost unchanged key before event delivery. Charge cascading scans against four times the original key length, and redact the whole matched key when the next scan cannot fit. The existing collision allocator preserves every entry and keeps generated keys secret-free.

Add subprocess-bounded regressions through public event preparation for the single-key family up to a 512 KiB key, nested/reserved-key collisions, deterministic output, Unicode and lowercase percent-encoded cascades, and readable short cascades. Include an opt-in public-API benchmark.

Closes #33924

## Performance

With a fixed matcher, cumulative cascade haystack bytes are at most `4 * original_key_bytes`; the existing discovery scan adds at most one original-key pass. Collision-heavy map complexity and pattern-set cardinality remain separate dimensions.

Controlled local-profile measurements on Rust 1.98.1 (unoptimized, five repetitions, input allocation/matcher construction excluded):

| Key bytes | Median event preparation |
| ---: | ---: |
| 8,195 | 0.421 ms |
| 16,387 | 0.825 ms |
| 32,771 | 1.650 ms |
| 65,539 | 3.328 ms |
| 131,075 | 6.546 ms |
| 262,147 | 13.207 ms |

Input doubling gives approximately 2x runtime. These are synthetic local measurements, not production or release-profile timings. The benchmark validates every payload and runs under an external watchdog.

## Fallbacks

- `crates/guest-agent/src/masker.rs::fully_masked_key`: replace an already-matched key with `***` when its scan budget is exhausted, instead of returning an unchecked intermediate key. Accepted secrets are at least five bytes, so the three-byte marker is intrinsically secret-free; the existing uniqueness allocator handles occupied/forbidden suffixes. This is permanent fail-closed handling of reachable input, not cross-version compatibility. Surface: synchronous guest event preparation; no rollout expiry or old/new interaction. It can remove readable context from an expensive matched key. Remove only with an alternative proving bounded work, secret-free generated keys, and entry preservation; no temporary rollout follow-up is applicable.

## Validation

- `cargo fmt --manifest-path crates/Cargo.toml --all -- --check` passed.
- `cargo clippy --manifest-path crates/Cargo.toml --locked --profile local -j 1 -p guest-agent --all-targets --all-features` passed.
- `cargo doc --manifest-path crates/Cargo.toml --locked --profile local -j 1 -p guest-agent --all-features --no-deps` passed.
- The identical new single-key regression failed on original `main` (`cbc43c760551064838b6505e9dabbe1c61f1ab44`) at its five-second subprocess watchdog. The full original guest-agent crate was rebuilt in an independent target directory; the fixed test passes in about 0.04 seconds.
- 47 existing masker unit tests passed.
- Six subprocess-bounded event-preparation tests and the UTF-8 minimum-secret-length integration test passed.
- `cargo build --manifest-path crates/Cargo.toml --locked --profile local -j 1 -p guest-agent --example masked_key_cascade_bench` and `timeout 30s crates/target/local/examples/masked_key_cascade_bench` passed.

Scope is guest-agent key masking, regression coverage, and the opt-in benchmark. Full guest/runner VM integration is left to the PR pipeline.


seven332 (commented):
LGTM

**Reviewed head:** `e2e8800a55d9cee0b83157f8bfe61e7fbf2e73e5`
**Practice revision:** `ad7ef4c6f4653213e59d3324e6b1d9e407faefeb`

## Summary

Bounds cascading JSON-key scans by the original key length while preserving ordinary key text, all entries, and secret-free generated names. The submitted head completed all six self-review passes without findings.

## Findings

No actionable findings. Each subsequent scan is charged before execution; return paths provide either a verified key or the intrinsically safe three-byte marker. The existing collision allocator continues to reserve unmatched keys and validate suffixes.

## Fallbacks

`masker.rs::fully_masked_key` now redacts the whole matched key on budget exhaustion. **Justified and declared**: this handles reachable accepted input and preserves the fail-closed security boundary. Surface: synchronous guest event preparation. It is a permanent input-resource policy with no rollout expiry or old/new protocol interaction. Removing it requires an equivalent bounded, secret-free, entry-preserving algorithm; no temporary rollout follow-up applies. The PR explicitly discloses loss of readable context for expensive keys.

## Testing

47 existing masker tests and seven integration cases passed, including externally bounded single-key/nested/encoding regressions. The identical new regression failed at the five-second watchdog on the independently rebuilt original-main crate and passed on this head in about 0.04 seconds. Formatting, all-target/all-feature guest-agent Clippy, rustdoc, and the checked public-API benchmark passed. Local unoptimized measurements show approximately 2x runtime per input doubling; collision-heavy maps and pattern cardinality are separate unchanged costs.

Full guest/runner VM integration and production execution were not run locally. CI and required GitHub review/merge gates remain separate from this code-review verdict.

seven332:
## Code Review: PR #33942

**Reviewed head:** `e2e8800a55d9cee0b83157f8bfe61e7fbf2e73e5`
**Verdict:** LGTM. No actionable findings; all six self-review passes completed on this unchanged head.

The [commit-bound review](https://github.com/vm0-ai/vm0/pull/33942#pullrequestreview-5194258342) records the correctness/security analysis, declared whole-key redaction policy, 54 passing affected tests, independent original-main watchdog failure, and bounded-scaling evidence. Formatting, Clippy, and rustdoc passed. Full guest/runner VM integration is left to CI. This records the workflow's code-review verdict; it does not replace required GitHub approval or CI checks.

PR #33943, opened 2026-09-14
perf(guest-agent): share masked key collision allocation progress

### LINKED ISSUE #33925: perf(rust): avoid quadratic suffix probing for masked JSON key collisions
## Summary

`SecretMasker::unique_masked_key` restarts its suffix search at 2 for every colliding JSON key. A single ordinary configured secret can produce thousands of different input keys that all redact to the same key. Event preparation then performs O(n²) candidate allocations and occupied-key probes, despite the current per-call termination bound.

Audited `main`: `0f2e7c8e3147cc1d10da7e57f24c2d9007523ce4`.

## Location and cause

- [`crates/guest-agent/src/masker.rs:365-368`](https://github.com/vm0-ai/vm0/blob/0f2e7c8e3147cc1d10da7e57f24c2d9007523ce4/crates/guest-agent/src/masker.rs#L365-L368) allocates a unique key independently for each masked entry.
- [`masker.rs:382-405`](https://github.com/vm0-ai/vm0/blob/0f2e7c8e3147cc1d10da7e57f24c2d9007523ce4/crates/guest-agent/src/masker.rs#L382-L405) starts decimal probes at 2 and fallback probes at 0 on every call. Earlier occupied candidates are retried repeatedly.

Configure only the synthetic secret `secret`. Both its plaintext `secret` and base64 `c2VjcmV0` are automatically registered. For k positions, construct every binary choice of those two strings. This gives 2^k distinct keys, each redacting to exactly k adjacent `***` markers. There are no cascading matches and no configured patterns that reject numeric suffixes. Inserting n such keys performs about n(n-1)/2 suffix probes, plus their allocations and map lookups.

```rust
use base64::Engine;
use guest_agent::masker::SecretMasker;
use serde_json::{Map, Value};
use std::time::Instant;

fn main() {
    let k: u32 = std::env::args().nth(1).unwrap().parse().unwrap();
    let n = 1usize << k;
    let raw = base64::engine::general_purpose::STANDARD.encode("secret");
    let masker = SecretMasker::from_raw(&raw);
    let mut object = Map::new();
    for index in 0..n {
        let mut key = String::new();
        for bit in 0..k {
            key.push_str(if index & (1 << bit) == 0 { "secret" } else { "c2VjcmV0" });
        }
        object.insert(key, Value::Null);
    }
    let mut event = Value::Object(object);
    let started = Instant::now();
    masker.mask_value(&mut event);
    println!("{:?}", started.elapsed());
    assert_eq!(event.as_object().unwrap().len(), n);
}
```

The audit measured the **exact unmodified `masker.rs` loaded by path** in a separate release-mode harness using the repository's dependency versions/features (`aho-corasick 1.1.5`, `base64 0.23.1`, `serde_json 1.0.151`) and Rust 1.98.1 on x86_64 Linux. Only the unused `from_config` constructor's config type was stubbed; masking used the actual source. Input construction and matcher compilation were excluded from timing. This is a synthetic local benchmark, not an observed production incident.

| k | Input keys | Serialized input bytes | Masking time |
| ---: | ---: | ---: | ---: |
| 10 | 1,024 | 79,873 | 68.375 ms |
| 11 | 2,048 | 174,081 | 301.712 ms |
| 12 | 4,096 | 376,833 | 1,284.650 ms |
| 13 | 8,192 | 811,009 | 5,357.768 ms |

The final case makes 33,550,336 numeric candidate probes. Doubling the key count roughly quadruples time. Use an external process watchdog for larger cases.

## Reachability and distinction from existing work

[`guest-agent/src/events.rs:92-105`](https://github.com/vm0-ai/vm0/blob/0f2e7c8e3147cc1d10da7e57f24c2d9007523ce4/crates/guest-agent/src/events.rs#L92-L105) recursively masks event-controlled keys, and [`cli/mod.rs:797-818`](https://github.com/vm0-ai/vm0/blob/0f2e7c8e3147cc1d10da7e57f24c2d9007523ce4/crates/guest-agent/src/cli/mod.rs#L797-L818) invokes it synchronously before delivery admission. The object can be nested in a normal event. The 811 KB input fits comfortably below both the 16 MiB ordinary stdout record cap and the 4 MiB delivery request cap; the latter is checked after masking anyway. This stalls the affected event pipeline and occupies its executor thread for seconds.

Closed #32594 fixed nontermination when configured secrets exclude the entire positive-suffix family. That fix is present and its small-input termination guarantee remains valid. This is a **distinct scaling follow-up**: one ordinary secret, always-available decimal candidates, and repeated rescanning of keys already assigned to earlier entries. Existing `tests/masked_key_collisions.rs` covers termination, safety, and at most 64 deliberately colliding entries, not this complexity. Open issues were searched for masking, masked/JSON keys, collisions, quadratic work, and the function name before filing.

## Acceptance criteria

- Share suffix-allocation progress across entries in one object (or use an equivalent deterministic strategy) so a family of colliding keys does not revisit all earlier candidates. Bound both decimal and fallback allocation work without weakening #32594's secret-free termination guarantee.
- Preserve every value, unmatched/reserved original keys, deterministic output, ordinary suffix behavior, and secret-free generated keys.
- Add coverage through event preparation for a large collision family derived from plaintext/base64 variants of one secret, as well as reserved decimal/fallback keys and nested objects. Show near-linear or O(n log n) aggregate collision bookkeeping with operation-count evidence or a controlled benchmark.

All 47 existing masker unit tests passed on the audited commit. No repository source changes were made.

seven332:
## Summary

Many event-controlled JSON keys can redact to the same key, making the old allocator rescan every earlier suffix. Share decimal cursors per masked base, one Unicode cursor, and a decimal probe budget across each object. For N input entries, allocation now makes at most N decimal and N Unicode candidate probes while preserving original-key reservations and secret-free generated keys.

Closes #33925.

Ordinary decimal suffixes and deterministic output remain stable, including interleaved bases that themselves contain suffixes. Objects with secret-rejected decimal candidates may use the existing Unicode identifiers earlier. The Unicode alphabet, cascading redaction, event envelope, public API, and dependencies are unchanged.

## Validation

- 47 masker unit tests passed.
- 5 collision integration tests passed through real event preparation, covering 4,096 plaintext/base64 variants, large blocked families, multiple bases, reserved decimal/Unicode keys, and nested/repeated objects.
- The large ordinary-collision regression failed on unchanged baseline production code at the existing five-second watchdog, then completed in about 0.13 seconds with the fix.
- Passed `cargo fmt --manifest-path crates/Cargo.toml --all -- --check`.
- Passed `cargo clippy --manifest-path crates/Cargo.toml --locked --profile local -p guest-agent --all-targets --all-features`.
- Passed `cargo doc --manifest-path crates/Cargo.toml --locked --profile local -p guest-agent --all-features --no-deps`.
- Broader repository validation runs in the PR pipeline.

## Performance evidence

Synthetic local measurements through `prepare_event_payload_for_run_id`, using Rust 1.98.1 on x86_64 Linux, locked dependencies, and the same unoptimized `local` profile. Each entry is the median of three sequential samples; input construction/cloning and matcher compilation are excluded. Baseline production code is `cbc43c760551064838b6505e9dabbe1c61f1ab44`, with this benchmark fixture added.

| Keys | Decimal before / after (ms) | Unicode before / after (ms) | Multiple families before / after (ms) |
| ---: | ---: | ---: | ---: |
| 256 | 11.147 / 1.180 | 81.885 / 2.029 | 32.825 / 1.101 |
| 512 | 45.544 / 2.593 | 347.962 / 4.507 | 135.506 / 2.330 |
| 1,024 | 211.386 / 5.620 | 1,490.958 / 9.765 | 560.827 / 4.730 |
| 2,048 | 850.448 / 11.817 | 6,443.785 / 21.174 | 2,378.247 / 9.817 |

Run the opt-in benchmark explicitly:

```sh
CARGO_BUILD_JOBS=1 cargo test --manifest-path crates/Cargo.toml --locked --profile local \
  -p guest-agent --test masked_key_collisions event_preparation_collision_scaling_benchmark \
  -- --exact --ignored --nocapture
```

The ordinary test suite uses the existing subprocess deadline only as a liveness guard; it contains no timing-ratio assertions.


seven332 (commented):
LGTM

**Reviewed head:** `e78166a280ba35e0bbd855b162fe1ce0667dd5dd`
**Practice revision:** `ad7ef4c6f4653213e59d3324e6b1d9e407faefeb`

### Summary
Object-local cursors and a shared decimal allowance remove repeated suffix scanning while retaining reserved originals, deterministic ordinary suffixes, and secret-free generated keys. The existing Unicode allocation path remains bounded across all collision families.

### Findings
No findings. Reviewed map growth, disjoint decimal families, budget/counter bounds, interleaved suffix-shaped bases, cascading redaction, Unicode safety, nested/repeated calls, and subprocess cleanup. A complete six-pass self-review also finished cleanly on this unchanged head.

### Testing
47 masker unit tests and all 5 collision integration tests pass, including the large-family regression that failed against baseline production code. Formatting, affected-crate Clippy across all targets/features, and rustdoc pass. The explicitly run three-sample local benchmark shows roughly doubling fixed-path work when inputs double; the PR includes raw-size comparisons and measurement limits. Broader CI and deployed-runner validation are separate from this review.

seven332:
## Code Review: PR #33943

**Reviewed head:** `e78166a280ba35e0bbd855b162fe1ce0667dd5dd`

### Summary
Shared per-object collision progress bounds aggregate suffix allocation while retaining original-key reservations and secret-free output. [Commit-bound review](https://github.com/vm0-ai/vm0/pull/33943#pullrequestreview-5194247624).

### Findings
No findings. All six self-review passes completed cleanly on this unchanged head.

### Testing
47 masker unit tests, 5 collision integration tests, formatting, Clippy, and rustdoc passed. The opt-in benchmark was explicitly run on baseline and fixed code; the PR records its controlled local results. Broader CI remains a separate gate.

### Verdict
LGTM

PR #33979, opened 2026-09-14
refactor(platform): replace mermaid diagram selectors with utilities

bingjiez666:
Part of #32402.

Removes six legacy class selectors from the Mermaid diagram block and replaces them with Tailwind utilities on the elements that already belong to first-party components.

## Drained tokens

`mermaid-block`, `mermaid-diagram-expand`, `mermaid-diagram-image`, `mermaid-diagram-pending`, `mermaid-diagram-source`, `icon-tooltip-trigger`.

| Inventory | Before | After |
| --- | --- | --- |
| Legacy tokens | 59 | 53 |
| CSS declarations | 396 | 365 |
| Production usages | 136 | 130 |

(Counts are from `pnpm style:migration:inventory` on this branch's merge base and head; main moved during the batch, so the absolute numbers differ from an earlier snapshot while the delta is the same six tokens.)

Nine CSS rules and 31 declarations are gone from `apps/platform/src/views/css/index.css`. Six consumption sites move into `mermaid-diagram.tsx`; one test reference moves to a documented slot.

## Replacement contract

| Retired declaration | Replacement |
| --- | --- |
| `.mermaid-block { margin-bottom: 16px }` | `mb-4` |
| `.mermaid-diagram-expand` (12 declarations) | `relative block w-full max-w-[420px] aspect-4/3 my-1 p-2 border-[1px] border-[hsl(var(--foreground)/0.1)] rounded-lg bg-[hsl(var(--muted)/0.3)] overflow-hidden cursor-zoom-in` |
| `.mermaid-diagram-expand:disabled { cursor: default }` | `disabled:cursor-default` |
| `.mermaid-block > .icon-tooltip-trigger` (3 declarations) | `wrapperClassName="block w-full max-w-[420px]"` |
| `.mermaid-diagram-image` (4 + 2 shared) | `absolute inset-2 w-[calc(100%-16px)] h-[calc(100%-16px)] object-contain bg-transparent!` |
| `.mermaid-diagram-pending` (4 + 2 shared) | `absolute inset-2 flex items-center justify-center text-muted-foreground` |
| `.mermaid-diagram-source > summary` (3 declarations) | `cursor-pointer text-muted-foreground text-[0.8125rem]` |
| `.mermaid-diagram-source > pre { margin-top: 8px }` | `mt-2!` |

Three choices need stating, because each is a deliberate deviation from the ergonomic spelling:

- **`border-[1px]` names its width.** The retired rule drew a 1px edge, while the shared bare `border` reads `--default-border-width`, which is 0.5px. Preserving the current rendering was chosen over adopting the hairline token; moving this box onto the token is a visual decision for a separate change.
- **Two values are important.** `bg-transparent!` on the image and `mt-2!` on the source block are the only faithful translations available: their competitors are unlayered rules (github-markdown-css's opaque canvas colour on `.wmde-markdown img`, and the Markdown adapter's 14px block margin on `.wmde-markdown pre`) that a utility in `@layer utilities` cannot outrank at any specificity. The retired rules won by being unlayered and late; importance reproduces that, and nothing else does.
- **Exact colour forms.** `border-[hsl(var(--foreground)/0.1)]` and `bg-[hsl(var(--muted)/0.3)]` rather than `border-foreground/10` and `bg-muted/30`, because Tailwind's alpha utilities interpolate in oklab and do not reproduce a literal `hsl()` value — the behaviour `docs/styles.md` already records for the mic meter.

`IconTooltip` gains `wrapperClassName` for the span it wraps a disabled child in. That span becomes the diagram box's layout parent while a diagram is still rendering, and it is the one element a caller cannot otherwise reach. It composes through `cn()`, so a caller's display utility replaces the wrapper's own `inline-flex` instead of racing it inside the same Tailwind layer, where class-attribute order decides nothing. The merge is checked directly: `cn("inline-flex", "block w-full max-w-[420px]")` returns `"block w-full max-w-[420px]"`.

`rehype-mermaid` emits an empty placeholder whose Mermaid source rides on the node's `data`; it never set `class="mermaid-block"`. Two comments that still described the old markup are corrected.

`data-slot="mermaid-diagram-source"` identifies the collapsible source block for page tests and carries no styles.

## Evidence

Measured against main with the App's own Tailwind compiler (`@tailwindcss/node`, the package `@tailwindcss/vite` drives) in system Chromium over CDP. `before` compiles `git show origin/main:` of the stylesheet with main's class strings; `after` compiles this branch with this branch's class strings. The fixture reproduces the ancestor chain captured by rendering the real chat thread page, not a hand-written approximation: `.okou-app` shell, workspace inset, chat thread section, scroll container, message container, assistant bubble, and the Markdown frame, with both Mermaid shapes (`rendered`, where Base UI merges the trigger and button onto one element, and `rendering`, where the disabled path splits them across the wrapper span).

Vendor and App stylesheet order matches the real bundle: in `apps/platform/dist/assets/index-*.css` the vendored sheet occupies bytes 0–27,227, the App entry's `@layer` statement starts at 27,228, and the App's own unlayered Markdown overrides sit at ~287k.

**All 18 theme states — the default palette plus all eight gradient palettes, each in Light and Dark — report 0 changed pixels and 0 computed-style or geometry differences.**

`GradientColorThemes` is `enabled: false` with no user, email or org overrides, so the default palette is the online-visible result and the eight palette states are a superset.

| | Light | Dark | All 18 states |
| --- | --- | --- | --- |
| This change | 0 px / 0 obs | 0 px / 0 obs | 0 px / 0 obs |
| Negative control | 493,605 px / 25 obs | 493,578 px / 25 obs | 8,836,436 px / 450 obs |

The negative control drops `aspect-4/3` from the migrated box. A diff channel that cannot fail proves nothing, so the zeros above rest on it.

Before (default Light): https://a.okou.io/c5k6bniq08.png
After (default Light): https://a.okou.io/toot579l4w.png

Per-pixel diff, default Light: no changed pixels

Negative control diff, default Light: 493,605 changed pixels

Headless Chromium reports `(hover: hover)` as false and Tailwind wraps `hover:` in `@media (hover: hover)`, so the run sets `--blink-settings=primaryHoverType=2,availableHoverTypes=2,primaryPointerType=4,availablePointerTypes=4`. A coarse-pointer run was not performed; no `hover:` utility is introduced or retired by this change.

## Checks

`pnpm lint:style`, `pnpm knip`, `pnpm -F @okouai/app check-types`, `pnpm -F @okouai/app lint` (including `oxlint --type-aware`), Prettier on every changed file, and 28 focused Vitest tests with `--maxWorkers=1`: the 16 Markdown/Mermaid tests plus 12 tests covering three other `IconTooltipButton` consumers. All pass.

`packages/ui` is unchanged.

## Not in this batch

- **No manifest entry.** `turbo/style-migration-manifest.json` requires a non-empty `cases` array whose IDs are registered in a case file, and those runners need a deployed preview plus private TEST storage state. No case ID was invented; the evidence above is a local equivalence harness, which is why this batch is not claimed as `baselined` or `verified`.
- **`language-mermaid` is misclassified in the baseline.** It reads as a first-party styling token, but the element carrying it in the DOM is produced by the Markdown pipeline, and `rehype-mermaid.ts` keys fence detection on exactly that class. It has zero CSS declarations anywhere — the vendored sheet only styles `.language-css`. It belongs to the external-contract audit, not to a first-party migration, and is left untouched here.
- **`.wmde-markdown pre .copied.active` is dead.** `CopyButton` never adds an `active` class; it swaps icons from React state. Narrowing that selector list was attempted and reverted: the ratchet keys an atom on its full selector text, so removing one branch registers as growth even though it is strictly a reduction. Draining it needs the batch that retires the whole `copied` contract.
- The chat-bubble and `okou-markdown-card` group is a separate contract: it needs a `p` renderer in the plain Markdown component map and its own analysis of the desktop-shell `user-select` rule.


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: bbf479368e3e0600693b2cc9ca73800fb098f6a6
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `bbf479368e3e0600693b2cc9ca73800fb098f6a6`. Practice documents read from `main` at `7887b755d4150ef0a974c5f8319dcdb64811a174`.

### Summary

Retires six legacy class selectors belonging to the Mermaid diagram block (`mermaid-block`, `mermaid-diagram-expand`, `mermaid-diagram-image`, `mermaid-diagram-pending`, `mermaid-diagram-source`, `icon-tooltip-trigger`) and spells their 31 declarations as Tailwind utilities on the first-party elements that already render them. `IconTooltip` gains a `wrapperClassName` prop for the span it wraps a disabled child in, which is the diagram box's layout parent while a diagram is still rendering. Two comments in `rehype-mermaid.ts` that described markup the plugin no longer emits are corrected.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | `MermaidDiagramView` is reached only through `MarkdownDivRenderer`'s `data.mermaidSignals` branch inside `MarkdownFrame`, so every retired `.wmde-markdown …` scope still holds at the new call sites. `IconTooltip`'s other 10 consumers were checked: on `main` the only rule matching `.icon-tooltip-trigger` is `index.css:1072`, which requires `.wmde-markdown .mermaid-block >` ancestry, so dropping the class is inert for them. |
| `REVIEW.md` | Meaningful coverage for new behavior; existing coverage still protects refactors | Pass | 50 focused tests run green at this HEAD: 16 markdown/mermaid, 22 in `diagram-rendering`/`chat-navigation-artifact-sidebar`/`shared-thread-presentation`, 12 across three other `IconTooltipButton` consumers. See Testing for why the new prop is not asserted directly. |
| `REVIEW.md` | Report concrete practice violations with rule and source location; do not infer defects from length or preference | Pass | No P0/P1 found. Non-blocking observations are listed separately and marked as such. |
| `REVIEW.md` | Persisted data, API contracts, response validation, compatibility windows | N/A | No schema, request, protocol, queue payload, persisted state, or service-worker surface is touched; the diff is CSS, two view components, one comment-only file, and the style baseline. |
| `REVIEW.md` | Every new fallback declared in a `Fallbacks` PR-body section | N/A | The diff introduces no default, no removed-path reader, and no rollout switch, so no `Fallbacks` section is required. |
| `docs/styles.md` | Business UI composes Tailwind utilities; no first-party CSS class selectors | Pass | Nine rules and 31 declarations leave `index.css`; the replacements are utilities on the consuming elements. `pnpm lint:style` passes, including `no-unknown-classes` against the real App Tailwind entry, which is what validates the bare-fraction `aspect-4/3`. |
| `docs/styles.md` | Legacy baseline shrinks only; prune rather than expand | Pass | `style-legacy-baseline.json` is regenerated with `lint:style:prune` plus Prettier. Measured at the merge base `394a3c36` and at this HEAD: 59 → 53 tokens, 396 → 365 declarations, 136 → 130 production usages, matching the PR body exactly. |
| `docs/styles.md` | A hand-written border width is a second registry for `--default-border-width` | Finding (non-blocking, see below) | `border-[1px]` on the diagram box names 1px while the shared hairline token is 0.5px. Deliberate and stated in code and PR body: the retired rule drew 1px, and adopting the token is a visual change. Recorded as a knowing deviation, not a defect. |
| `docs/styles.md` | Tailwind colour/gradient utilities interpolate in oklab and do not reproduce a literal value | Pass | `border-[hsl(var(--foreground)/0.1)]` and `bg-[hsl(var(--muted)/0.3)]` use the exact forms rather than `border-foreground/10` / `bg-muted/30`, which is the behaviour this guide already records for the mic meter. |
| `docs/styles.md` | Tests and runtime use roles, names, refs, `data-*` hooks, or documented slots rather than styling classes | Pass | `data-slot="mermaid-diagram-source"` replaces the class the test used to query, and it carries no styles. |
| `docs/styles.md` | Shared component APIs do not export class strings | Pass | `wrapperClassName` is an ordinary optional React `className` prop with a default of `undefined`; no class-name constant is exported or imported across modules. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | `wrapperClassName?: string` on both components; no suppression comment anywhere in the diff. `pnpm -F @okouai/app check-types` and `lint` (including `oxlint --type-aware`) pass. |
| `docs/bad-smell.md` | Document changed public contracts and assess their consumers | Pass | The new prop carries a doc comment explaining why the wrapper is the one element a caller cannot reach; all 10 existing `IconTooltipButton` consumers are unaffected because the prop is optional and the removed class had no rule matching them. |
| `docs/bad-smell.md` | Static imports in production code | Pass | `cn` is added as a static named import from `@okouai/ui`. |
| `docs/bad-smell.md` | No error-handling or fallback slop | N/A | No `catch`, default value, or branch over a contract-excluded state is added or removed. |
| `docs/testing.md` | Prefer integration coverage through real entry points; avoid cases that pin incidental implementation | Pass | The one changed assertion stays inside the existing `setupPage`-driven page test and still reads the rendered diagram source text. |
| `docs/testing.md` | Changes to requests, protocols, queue payloads, or persisted state cover old/new rollout combinations | N/A | None of those surfaces changes. |
| `docs/testing/testing-external-behavior.md` | Do not assert on component state, query cache, CSS classes, or internal callbacks | Pass | This is why no test asserts that `wrapperClassName` lands on the wrapper span: such a test would assert a CSS class, which this guide forbids. The layout contract is carried by the measured evidence in the PR body instead. |
| `docs/testing/testing-external-behavior.md` | Construct cases through page interactions and verify through the page | Pass | `markdown-mermaid.test.tsx` continues to drive the real chat page and read rendered content; the selector change swaps one DOM hook for a documented slot and does not move the boundary. |

### Findings

No P0 or P1 findings.

Non-blocking observations, none of which should hold the merge:

1. **`border-[1px]` is a knowing deviation from the hairline token** — `turbo/apps/platform/src/views/components/mermaid-diagram.tsx:66`. `docs/styles.md` says components must not hand-write a border width because `--default-border-width` owns that decision. The retired rule drew `1px solid` while the token is 0.5px, so the bare `border` utility would change the stroke on high-density displays. The PR keeps 1px to preserve the current rendering and says so both in code and in the body. Adopting the token later is a visual decision for its own review.
2. **Two important utilities** — `bg-transparent!` (`mermaid-diagram.tsx:102`) and `mt-2!` (`mermaid-diagram.tsx:122`). Both competitors are unlayered rules that no utility can outrank at any specificity: github-markdown-css's `.wmde-markdown img { background-color: var(--color-canvas-default) }` and the allowlisted adapter's `.wmde-markdown blockquote, table, pre, details { margin: 14px }` at `index.css:876`. Importance reproduces the precedence the retired rules had by being unlayered and late; `markdown-frame.tsx:23` already uses the same device. Worth remembering that these two now outrank any future deliberate override.
3. **`background: none` narrowed to `bg-transparent!`** — the retired shorthand also reset `background-image`, while the utility sets only `background-color`. Nothing sets a background image on `.wmde-markdown img`, and `background-image` was compared explicitly across all 18 measured states with no difference, so this is safe today; it is a slightly narrower guarantee than the rule it replaces.
4. **`wrapperClassName` widens a shared component's API for one consumer** — `turbo/apps/platform/src/views/components/icon-tooltip.tsx:12`. The alternative is for the Mermaid box to compose `Tooltip` directly and own its own disabled-trigger wrapper, which is a larger change. The prop is optional and inert for the other 10 consumers, so the cost is small, but the seam is worth revisiting if a second consumer ever needs different wrapper styling.
5. **A pre-existing test fixture is now slightly stale** — `turbo/apps/platform/src/views/okou-page/__tests__/markdown-content.test.tsx:273` feeds raw HTML `<div class="mermaid-block" data-mermaid-code="…">` to prove a quoted marker renders as a plain div. The real placeholder never carried that class (`mermaidBlockNode` sets `properties: {}`), and after this PR the rendered block does not either, so the fixture no longer resembles current markup. The protection it exercises — `rehype-raw` cannot produce `data` — is unchanged and the test passes, so this is cosmetic. Outside this PR's scope.

### Testing

Coverage at this HEAD, all with `--maxWorkers=1`:

- `markdown-mermaid.test.tsx` + `markdown-content.test.tsx`: 16 passed. These drive the real chat page through `setupPage` and cover the rendered diagram, the collapsible source block, the streaming/partial fence, and the counterfeit-marker case.
- `diagram-rendering.test.tsx`, `chat-navigation-artifact-sidebar.test.tsx`, `shared-thread-presentation.test.tsx`: 22 passed. These are the other page tests that render a diagram through `IconTooltipButton`, including the shared-thread surface.
- `chat-attachment-composer.test.tsx`, `chat-composer-layout.test.tsx`, `agentphone-connect-page.test.tsx`: 12 passed. Three unrelated `IconTooltipButton` consumers, exercised to confirm the shared-component change is inert for them.

Static checks: `pnpm lint:style`, `pnpm knip`, `pnpm -F @okouai/app check-types`, `pnpm -F @okouai/app lint`, and Prettier on every changed file. `packages/ui` is unchanged, so its checks are not implicated.

The one changed assertion moves from a styling class to `data-slot="mermaid-diagram-source"`, which `docs/styles.md` documents as the supported hook and which does not change the test's boundary.

**Deliberate coverage gap.** No test asserts that `wrapperClassName` reaches the wrapper span, and none should: `docs/testing/testing-external-behavior.md` explicitly forbids asserting on CSS classes, and jsdom has no layout engine, so the contract this prop protects — the reserved diagram box keeping its size while the diagram renders — is not observable there. The PR carries that contract as measured evidence instead: the App's own Tailwind compiler in Chromium over CDP, across the default palette plus all eight gradient palettes in Light and Dark, reporting 0 changed pixels and 0 computed-style or geometry differences in all 18 states, with a negative control that drops `aspect-4/3` reporting 493,605 changed pixels. That evidence is a local equivalence harness rather than a registered migration case, which the PR body states plainly, and it is why the batch is not claimed as `baselined` or `verified`. This residual gap is disclosed rather than closed, and it is consistent with how the other `#32402` batches without a deployed runner have been handled.

PR #33999, opened 2026-09-14
fix(runner): reject overlong integers in flush markers

### LINKED ISSUE #33994: test(python): cover overlong-integer runner flush markers
The malformed-marker tests omit a decoder failure that terminates the JSONL flush watcher on the repository's pinned runtime. A 5,018-byte marker containing a 5,000-digit integer raises an uncaught `ValueError` on CPython 3.14.0 with mitmproxy 12.2.3. After the watcher terminates, replacing the file with a valid marker does not restore processing.

Verified against `main` at `ef1c689b4bae1ec5a3f4e2b89516819419d4b882`, using the committed `.python-version` and `uv.lock`.

Relevant code and coverage:

- `crates/runner/mitm-addon/tests/test_runner_flush_request.py:89–155` covers missing files, invalid UTF-8/JSON, non-object input, stale generations, and invalid request IDs for both consumers. It has no integer-conversion-limit case or subsequent-valid-marker recovery assertion.
- `crates/runner/mitm-addon/src/runner_flush_request.py:24–32` bounds the input at 65,536 bytes but catches only `OSError`, `UnicodeDecodeError`, and `json.JSONDecodeError`. The integer conversion limit raises plain `ValueError` before envelope validation.
- `crates/runner/mitm-addon/src/runner_flush_lifecycle.py:136–142` lets this exception leave the watcher loop. The usage consumer also reads the marker before its delivery-flush `try/finally` at lines 209–224.
- `crates/runner/mitm-addon/tests/test_runner_jsonl_flush.py:262–284` covers an exception from `flush_log_path()`, after marker parsing, so that recovery test does not exercise this failure.

Minimal reproduction from `crates/runner/mitm-addon`, after `uv sync --locked`:

```bash
uv run --no-sync python - <<'PY'
import platform
import sys
import tempfile
from pathlib import Path

sys.path.insert(0, str(Path("src").resolve()))
from runner_flush_request import MAX_RUNNER_FLUSH_REQUEST_BYTES, read_runner_flush_request

print(platform.python_version(), sys.get_int_max_str_digits())  # 3.14.0 4300
payload = b'{"requestedAtMs":' + b"1" * 5000 + b"}"
assert len(payload) == 5018 < MAX_RUNNER_FLUSH_REQUEST_BYTES
with tempfile.TemporaryDirectory() as directory:
    path = Path(directory) / "usage-flush-request"
    path.write_bytes(payload)
    print(read_runner_flush_request(path, get_usage_state_id=lambda: "audit-state"))
PY
```

Actual result: `ValueError: Exceeds the limit (4300 digits) for integer string conversion: value has 5000 digits`. The invalid envelope should be ignored without terminating its consumer.

I also exercised the unchanged public usage consumer and real JSONL watcher in an isolated temporary directory. The usage consumer raised `ValueError`; the `runner-jsonl-flush` thread terminated with that exception before any flush or acknowledgement. A later atomically published valid marker had no live watcher to consume it. Explicitly restarting the watcher provided a positive control: the valid request was then flushed and acknowledged with `pending=0`. Synchronization used an exception event and thread join, with the real decoder and flush operation unchanged.

This is a malformed/corrupt local state-file recovery scenario. The normal Rust producer writes a flat typed envelope; this report does not claim that ordinary requests generate such integers. The observable impact is loss of subsequent JSONL flush acknowledgements; the Rust timeout path proceeds with network-log upload without that acknowledgement (`crates/runner/src/proxy/flush.rs:795–829`).

Suggested acceptance criteria:

- Add a byte-bounded overlong-integer case for both marker consumers, asserting rejection without an exception, flush, or acknowledgement.
- Exercise the real watcher observing that invalid marker, then a valid current-generation marker, and assert that the same watcher flushes and publishes the matching acknowledgement. Use explicit synchronization and fail on background-thread exceptions.
- Classify the actual decoder rejection at the shared parsing boundary while retaining the integer-conversion guard and existing valid-marker behavior.
- Run the regression on pinned CPython 3.14.0 without changing the interpreter's digit or recursion limits.

The related nested-JSON reports #29842 and #31426 were closed because their `RecursionError` was not reproducible on Python 3.14. This integer-conversion failure reproduces on that exact runtime; the nested-array probes up to 64 KiB still pass. Open-issue searches for flush, marker, parser, integer, and the shared parser path found no matching issue. The 530 selected and related existing tests pass on Python 3.14.0.

seven332:
## Summary

A corrupt flush marker containing a 5,000-digit integer raises plain `ValueError` on the pinned Python runtime, terminating the JSONL watcher and preventing later requests from being acknowledged. Classify that rejection at the shared decoder boundary so both consumers ignore the malformed marker and the same watcher can flush and acknowledge a subsequent valid request.

The regression uses real files, decoding, flushing, and atomic acknowledgement publication, with explicit bad-file observation and fatal background-thread warnings. Existing byte/digit limits and request/acknowledgement formats are preserved; this change is limited to parser recovery and its tests.

Closes #33994.

## Validation

Verified with committed uv 0.11.32 / CPython 3.14.0 / mitmproxy 12.2.3, without changing interpreter limits.

- All 3 new cases fail before the production fix, including watcher termination and the missing subsequent acknowledgement.
- 177 related tests passed:
  ```sh
  uv run --no-sync python -m pytest tests/test_runner_flush_request.py tests/test_runner_jsonl_flush.py tests/test_runner_usage_flush_signal.py tests/test_done_hook.py tests/test_counters.py tests/test_state_file.py tests/test_jsonl_writer.py tests/test_jsonl_writer_recovery.py tests/test_jsonl_writer_lifecycle.py tests/test_logging_utils.py -q -W error::pytest.PytestUnhandledThreadExceptionWarning
  ```
- `uv run --no-sync ruff format --check .` and `uv run --no-sync ruff check .` passed.
- `uv run --no-sync basedpyright -p .`: 0 errors, warnings, or notes.
- `git diff --check` passed. The complete addon suite runs in the PR pipeline.


seven332 (commented):
LGTM

**Reviewed head:** `56df2e69837902979350b7cd0b0dbedb20496583`
**Practice revision:** `9047ac1b5739f9de0dc64c9abc4534b92ef81648` (main).

**Summary:** The shared marker reader now classifies integer-limit ValueError at the UTF-8/JSON decoding boundary, preserving the existing invalid-marker outcome for both consumers and allowing the same JSONL watcher to acknowledge later valid requests.

**Findings:** No findings. The byte and integer guards, safe file opening, generation/request/path validation, flush failure ownership, and acknowledgement format remain intact. The new handling is meaningful input rejection and recovery, with no fabricated data or rollout fallback.

**Testing:** All three new cases fail against the pre-fix code on pinned CPython 3.14.0. The final head passes 177 related tests with fatal background-thread warnings, locked Ruff formatting/lint, and BasedPyright (0 errors or warnings). The recovery test observes actual malformed bytes before atomic replacement, keeps the real reader/decoder/flush behavior, verifies the same live watcher, and checks the matching pending=0 acknowledgement and network-log output. Six focused self-review passes completed without findings on this unchanged head.

The complete addon suite and other required checks remain separate PR CI gates. This author-owned PR receives a COMMENT review; it does not satisfy a required independent approval.

seven332:
## Code Review: PR #33999

**Reviewed head:** `56df2e69837902979350b7cd0b0dbedb20496583`

### Summary

Shared decoder rejection now preserves flush-marker recovery for both consumers. The six-pass self-review completed cleanly on this head.

### Findings

No findings. [Commit-bound review and rationale](https://github.com/vm0-ai/vm0/pull/33999#pullrequestreview-5195205006).

### Testing

Three regressions failed before the fix; 177 related tests, locked Ruff formatting/lint, and BasedPyright passed after it. The real watcher recovers after observed invalid bytes and publishes a matching acknowledgement for a valid replacement. Full CI and any required independent approval remain separate gates.

### Verdict

LGTM.

PR #33831, opened 2026-09-14
refactor: remove markdown time feature switch

e7h4n:
`markdownTime` has reached full rollout, but Markdown rendering still reads the switch and retains its disabled branch. Make valid offset-bearing `<time>` tags always display in the browser's locale and timezone, and remove the retired switch from configuration and tests.

## Terminal decision and retained behavior

- Terminal state: `fully-rolled-out`, explicitly confirmed by Ethan in this cleanup request and corroborated by merged rollout PR #33704 (`a8970b8e481c29ebb4d0b7277820c2dd179f1c2b`). Latest starting `main`: `826d131351049b7f35f45cad577618e01b231544`.
- Keep the existing datetime schema, browser locale selection, timezone conversion, original `datetime` attribute, and renderer registration for plain, media, and linked-media Markdown.
- Keep original text for missing, invalid, or offset-free datetimes, and keep time tags literal inside code. These are retained input-handling contracts, not the retired switch-off behavior.

## History and cleanup

- Reviewed introduction commit `98d2ec14f6f36170b904ebfd40204d58050e252d` (#32899) and its parent. Before that commit, time elements rendered their original content; the commit introduced the gated renderer, enum entry, registry entry, and page tests. Preserve its enabled path and remove only the gate and disabled behavior.
- The same introduction also added CLI `formatRunTimestamp` / `printRunTimeGuidance` behavior and command tests. Those were never gated and remain applicable to the permanent feature, so they are retained.
- Reviewed later changes: #33047 already removed the cached-switch fixture; #33704 globally enabled the switch and added rollout assertions; #33062 copied the switch into the icon-migration fixture. Later rich-Markdown layout and thumbnail changes do not alter time rendering and are preserved.
- Remove the enum and registry entry, renderer switch import/subscription/conditional, five enabled test overrides, the disabled-localization test, two rollout assertions, and the stale E2E fixture key. Keep the existing nine positive page cases without switch setup. No tests were added.
- Existing registry filtering removes retired overrides from API reads and future writes; the updated renderer no longer consults them. The existing App hydration starts from local defaults and applies only known keys, so the globally enabled prior App tolerates the omitted key and the updated App ignores the prior API's extra key. No schema migration is needed.

## Residual search

[...]

PR #32930, opened 2026-09-09
feat: support fuzzy workflow command matching

bingjiez666:
Typing `/pdau` currently returns no workflow suggestions even when `pr-design-acceptance-url` is available. Add a small, dependency-free matcher that accepts characters in order with gaps, highlights the actual matched ranges, and inserts the workflow's canonical command when selected.

Results rank by exact name, prefix, substring, then abbreviation. Queries shorter than three characters retain literal matching, and numeric runs must remain contiguous so PR identifiers cannot match across unrelated digits. The `composerWorkflowFuzzySearch` switch enables the new behavior for staff; existing prefix and substring behavior remains available when it is off.

## Verification

- All 16 workflow composer page tests pass, including abbreviation highlighting and insertion, ranking, numeric identifiers, short queries, and the disabled switch.
- Formatting, affected-file Oxlint/ESLint, Platform and Core type checks, scoped Knip, and style policy.
- Browser verification on the App/API previews for `57bec4398517f1c7df2068ec6b30f8525fc1cd04`: mixed-case abbreviation highlighting, full command insertion with preceding text preserved, exact/prefix/substring/abbreviation ordering, contiguous numeric matching, and the short-query limit all pass. Chromium 152, 1280 × 900, disposable test account with seven private workflow fixtures; onboarding bypassed and the Lab switch enabled.
- All applicable PR CI checks pass, including App/API tests, deployment, and the Turbo/security gates.

## Acceptance

1. Enable `composerWorkflowFuzzySearch` in Lab and open a chat with the relevant workflows available.
2. Type `/pdau` to find `pr-design-acceptance-url`; confirm the complete name is displayed with the matching letters highlighted and Enter inserts the full command.
3. Type `/pr-auto`; confirm the exact workflow precedes longer prefixes, substring matches, and abbreviations.
4. Type `/pr26809`; confirm `pr-26809-ship-watch` can match while `pr-26-809-ship-watch` cannot.

PR #34033, opened 2026-09-14
refactor: remove composer workflow fuzzy search feature switch

bingjiez666:
## Terminal state

`fully-rolled-out`. The enabled behavior is now the permanent product behavior.

**Evidence:** PR #33680 (`57b7f9b726`, 2026-09-12) removed the staff-org gating
and set `enabled: true` globally for `composerWorkflowFuzzySearch`. No
consumer, registry, or API entry restricted it afterwards.

## History

- Introduced by #32930 (`550b15a978`): added the `ComposerWorkflowFuzzySearch`
  key, its registry entry, the `fuzzy` parameter on `findWorkflowQueryMatches`,
  the `ComposerSlashWorkflowMatch` / `matchRanges` model, the
  `SlashWorkflowName` renderer, and `useComposerWorkflowSuggestions`.
- Refined by #33186 (`27d2b3e050`): compact-abbreviation ranking
  (`skippedCharacters`, `start`) inside the already gated fuzzy branch.
- Fully rolled out by #33680 (`57b7f9b726`).

Comparing the introducing commit against its parent, the pre-switch behavior
was prefix-then-substring matching with a single highlighted range. Everything
else added by #32930 and #33186 belongs to the enabled path and is kept.

## Kept behavior (switch on)

- Exact workflow-name matches rank before prefix, substring, and abbreviation
  matches.
- Subsequence (abbreviation) matching for queries of at least three characters,
  with contiguous numeric identifiers.
- Compact-abbreviation ranking, multi-range `matchRanges` highlighting, and the
  `SlashWorkflowName` renderer.

## Removed

- `FeatureSwitchKey.ComposerWorkflowFuzzySearch` and its registry entry.
- The `fuzzy` parameter of `findWorkflowQueryMatches`, and the two conditions
  it guarded (`fuzzy && normalizedName === normalizedQuery`,
  `fuzzy && normalizedQuery.length >= 3`).
- The switch read in `useComposerWorkflowSuggestions`.
  `FeatureSwitchKey` / `featureSwitch$` stay imported in that file for
  `ComposerSlashTemplatePanel`.
- The recorded switch value in the frozen `icon-mono-fixtures.json` snapshot,
  matching the precedent in #33831.

## Tests

- Dropped the `featureSwitches` overrides from the four fuzzy-matching page
  tests; they now exercise the permanent behavior directly.
- Replaced the `enabled: false / enabled: true` matrix with a single test. The
  `enabled: false` case asserted the discarded strict-matching behavior and was
  deleted. The `enabled: true`, `/pd` case is permanent behavior (queries
  shorter than three characters never abbreviate-match) and is kept as
  "Keep strict workflow matching for queries shorter than three characters".
- No negative or existence-assertion tests were added.

## Second-pass keyword sweep

`composerWorkflowFuzzySearch`, `ComposerWorkflowFuzzySearch`, `fuzzy`/`Fuzzy`,
`findWorkflowQueryMatches`, `findWorkflowSubsequence`, `matchRanges`,
`ComposerSlashWorkflowMatch`, `workflow-query-match`, `SlashWorkflowName`,
`skippedCharacters`.

Remaining hits are all unrelated or intentional: `CHANGELOG.md` release history
(immutable), `e2e/playwright/lib/clerk-api.test.ts` ("fuzzy lookalikes"),
`run-reads.bdd.test.ts` (API log search), and the
`1083_fuzzy_dragon_lord` migration tag. The `matchRanges` /
`findWorkflowSubsequence` / `SlashWorkflowName` symbols are permanent product
code, now decoupled from any switch.

## Verification run

- `prettier --check` on all changed files — pass
- `pnpm -F @okouai/core check-types` — pass
- `pnpm -F @okouai/app check-types` — pass
- `pnpm -F @okouai/core lint` — pass
- `pnpm -F @okouai/app lint` — pass
- `pnpm knip` — no unused-code issues
- `vitest --run packages/core/src/__tests__/feature-switch.test.ts` — 33/33 pass
- `vitest --run apps/platform/src/views/okou-page/__tests__/chat-composer-workflows.test.tsx`
  — 20/20 pass

Left to the pipeline: the full Vitest suite, Rust crate checks, e2e/Playwright,
style lint, and the remaining workspace builds.


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 3d6c89901c21828e1b92228c6d22d63465ec4909
PR_AUTO_REVIEW_VERDICT: LGTM

Reviewed HEAD: `3d6c89901c21828e1b92228c6d22d63465ec4909`
Practice revision: `ba0fc4da5dac1c87b94310a236c2bce42f4cb0b7` (`main`)

### Summary

Removes the terminal `composerWorkflowFuzzySearch` feature switch and makes its
enabled behavior unconditional: `findWorkflowQueryMatches` loses its `fuzzy`
parameter, the composer stops reading the switch, and the enum entry, registry
entry, and frozen fixture value are deleted. The four gated page tests now run
against the production default, and the switch-matrix test is reduced to the one
case that is permanent behavior.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | `findWorkflowQueryMatches` has exactly one caller (`tiptap-workflow-composer.tsx:456`). With `fuzzy = true` the retained expressions are token-identical to the old enabled path: `normalizedName === normalizedQuery ? 0 : start === 0 ? 1 : 2` and `normalizedQuery.length >= 3`. |
| `REVIEW.md` | Confirm existing coverage still protects a refactor | Pass | The four abbreviation/ranking page tests are retained unchanged apart from dropping the override, and they exercise exact-match rank 0, prefix/substring ranks, compact-abbreviation ordering, contiguous numeric identifiers, and highlight ranges. |
| `REVIEW.md` | For a removed/rerouted flow, require regression coverage of the affected path and correct containment | Pass | The permanent path keeps full coverage; the only removed path is the disabled branch, which no longer exists. No capability or permission surface is involved — the switch was never an authorization boundary (`feature-switch.ts` header). |
| `REVIEW.md` | Report practice violations with rule and source; do not infer defects from search matches | Pass | No P0/P1 found. Residual `fuzzy` matches in the repo (`clerk-api.test.ts`, `run-reads.bdd.test.ts`, migration tag `1083_fuzzy_dragon_lord`, CHANGELOGs) are unrelated symbols, correctly left alone. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | Diff adds no type assertions or suppression comments; it only deletes a parameter and two conditions. |
| `docs/bad-smell.md` | Interface changes: document the changed public contract and assess consumers | Pass | `findWorkflowQueryMatches` is module-internal to Platform (single caller); `FeatureSwitchKey` is a cross-package enum whose removal is covered in the compatibility row below. `tsc` exhaustiveness on `Record<FeatureSwitchKey, FeatureSwitch>` guarantees the registry and enum stay in sync. |
| `docs/bad-smell.md` | Preserve meaningful recovery, cleanup, and security checks when removing code | Pass | Nothing removed here is recovery, cleanup, or a security check; the deleted branch was a rollout gate only. |
| `docs/fallback.md` §2 | When a switch is terminal, delete the switch, the disabled branch, and the disabled branch's tests together | Pass | Enum + registry + consumer read + disabled branch + the `enabled: false` test case are all removed in one commit. |
| `docs/fallback.md` §1 | No negative/tombstone tests asserting removed behavior stays removed | Pass | The `enabled: false` matrix case was deleted rather than rewritten into an existence assertion; no new negative test is added. |
| `docs/fallback.md` §10 | A removal PR must carry type, single-writer, production, or rollout evidence | Pass | Rollout evidence: #33680 (`57b7f9b726`) set `enabled: true` with no hash gating, so every identity resolved `true`. The PR body records the introducing commit, the refinement (#33186), and the rollout commit. |
| `docs/fallback.md` §9 | Declare newly introduced fallbacks in the PR summary | N/A | The diff introduces no fallback; per §9 a removal-only PR needs neither a `Fallbacks` section nor `Fallbacks: none`. |
| `.claude/skills/feature-switch/SKILL.md` | The enum key and the `FEATURE_SWITCHES` registry entry are the two required registration points | Pass | Both removed (`feature-switch-key.ts:73`, `feature-switch.ts:73`). |
| `.claude/skills/feature-switch/SKILL.md` | Clear every auxiliary association: connector auth-method map, token capabilities, org-scoped keys, sidebar nav | Pass | Repo-wide search for `ComposerWorkflowFuzzySearch` / `composerWorkflowFuzzySearch` returns zero hits after the diff; `FEATURE_SWITCH_BY_AUTH_METHOD`, `CONDITIONAL_CAPABILITIES`, and `ORG_SCOPED_FEATURE_SWITCH_KEYS` never referenced this key. |
| `.claude/skills/ccstate/SKILL.md` | Keep derivation in computed values; do not create signals or mutate the store during render | Pass | `useComposerWorkflowSuggestions` only drops one unconditional `useGet(featureSwitch$)` read; hook order stays unconditional and no signal is created during render. |
| `.claude/skills/ccstate/SKILL.md` | Every async operation keeps an owner and cancellation path | Pass | No async work changes; `useLastLoadable(composer.workflow.workflows$)` and its loading state are untouched. `featureSwitch$` / `FeatureSwitchKey` imports are correctly retained for `ComposerSlashTemplatePanel` at line 417. |
| `docs/deployment-compatibility.md` | Frontend → backend responses must survive old-frontend/new-backend and new-frontend/old-backend pairings | Pass | `featureSwitchesResponseSchema` is `z.record(z.string(), z.boolean())`, so neither direction is a strict-schema break. An old bundle computes its base state from its own bundled `getAllFeatureStates` and merges API overrides by iterating **its own** `FeatureSwitchKey` values, skipping `undefined` — so it keeps the `enabled: true` default. A new bundle ignores a key an old API still returns. |
| `docs/deployment-compatibility.md` | Do not remove a response field while old frontend code may still read it | Pass | The key is not a required response field: it only ever appeared in the overrides record when a user had stored an override, and the old client already tolerates its absence through the `value !== undefined` guard in `applySwitches`. |
| `docs/deployment-compatibility.md` | Persisted state written by one version must stay readable by another | Pass | Stale `user_feature_switches` rows holding the retired key are dropped by `filterFeatureSwitchOverrides` (`key in FEATURE_SWITCHES`), on both the read and the `POST` write path. No migration or backfill is required, matching prior removals such as `chatTouchSelection`. |
| `docs/testing.md` | Test the contract an external caller relies on; prefer integration coverage through real entry points | Pass | All retained cases boot the Router with `setupPage`, type into the composer, and assert the rendered slash menu and editor text. |
| `docs/testing.md` | Avoid duplicate cases that restate static configuration | Pass | Collapsing the two-case matrix into one test removes the case that only restated the switch's off state. |
| `docs/testing/testing-external-behavior.md` | Construct cases through page interaction and verify through the page; never assert internal state | Pass | No store mutation, hook call, or signal assertion; assertions are on `slashWorkflowNames()`, highlighted `workflow-query-match` text, editor content, and the "No matching workflows" empty state. |
| `docs/testing/app-testing.md` | `setupPage`'s `featureSwitches` configures the mocked authoritative API response; defaults come from the repository registry | Pass | Removing the override is the correct expression of the post-removal default; the suite passing locally (20/20) confirms the registry default now drives the behavior. |
| `docs/testing/app-testing.md` | Default to `toBeInTheDocument()`; use `toBeVisible()` only when visibility is the contract | Pass | See non-blocking observation below — the `toBeVisible()` call is carried over verbatim from the case being restructured, not newly introduced behavior. |

### Findings

No P0 or P1 findings.

**Non-blocking observations**

- `chat-composer-workflows.test.tsx:703` — the retained empty-state assertion uses
  `.resolves.toBeVisible()`, while `docs/testing/app-testing.md#assertions` asks for
  `toBeInTheDocument()` unless visibility itself is the contract. The line is carried
  over unchanged from the matrix case this PR restructures, so the diff neither
  introduces nor worsens the deviation. Worth aligning opportunistically, not here.
- The new test name, "Keep strict workflow matching for queries shorter than three
  characters", now states the real invariant (the `normalizedQuery.length >= 3`
  threshold) rather than the switch state. Good naming for the permanent contract.

### Testing

- Retained coverage for the permanent behavior: abbreviation match + insertion,
  exact-before-prefix-before-substring-before-abbreviation ranking, compact-abbreviation
  ordering with highlight ranges, contiguous numeric identifiers, and the sub-three-character
  strict-matching boundary. Each runs through `setupPage` with no switch override.
- Removed exactly one case — the `enabled: false` arm asserting `/pdau` finds nothing —
  which covered the discarded behavior and would have become a tombstone if rewritten.
- Author-reported verification: `prettier --check` on the changed files, `check-types` and
  `lint` for `@okouai/core` and `@okouai/app`, `pnpm knip` with no unused-code issues,
  `feature-switch.test.ts` 33/33, and `chat-composer-workflows.test.tsx` 20/20.
- `turbo/packages/core/src/__tests__/feature-switch.test.ts` needed no edit because it
  never asserted on this key, unlike the `MarkdownTime` removal in #33831.
- No coverage gap identified. Cross-version coverage is not required here: the response
  contract is an open-keyed record and the retired key is filtered out of persisted
  overrides by existing registry-membership logic, so there is no old/new request or
  payload combination whose behavior this diff changes.
- CI note: the `cli-e2e-03-runner` `run-t10-real-claude-smoke.bats` failure is an upstream
  provider refusal ("This request triggered cyber-related safeguards") on a real-Claude
  smoke run. It is unrelated to this Platform-only diff and is being retried, not repaired.

PR #34043, opened 2026-09-14
refactor(app): drop the category description in the slash template pane

tongx-blip:
Ming's review of the `/` panel: 感觉字儿有点儿多 — the template detail pane opens with a line of prose above the covers.

The description repeated what the row you just highlighted, the template count and the thumbnails already say, so it is removed. The header (icon, category name, `N templates`) now sits directly above the cover grid with the same 12px gap the paragraph used.

- `SlashTemplateDetailPane` loses the `<p>` and `categoryDescription()`.
- The five now-unused `chat.composer.slashPanel.*Description` strings are deleted from all 10 locales.

No other surface referenced those keys (`onboarding.workflowPicker.categoryDescription` is unrelated), and the panel tests assert on `data-category`, not the prose.


tongx-blip:
LGTM


PR_AUTO_REVIEW_COMMIT: fddc542d2749ebd7e34599db5808d919e00c5e6c
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `fddc542d2749ebd7e34599db5808d919e00c5e6c`
Practice revision: `af933c2c2e506f8e377b07411776c77ee99e0a85` (`main`)

### Summary

Removes the one-line category description from the slash panel's template detail pane (`SlashTemplateDetailPane`), deletes the now-unreferenced `categoryDescription()` helper, and drops the five `chat.composer.slashPanel.*Description` strings from all ten locale files. The pane header (icon, category label, `N templates`) now sits directly above the cover grid; the covers scroller keeps its existing `mt-3` offset, so the only geometry change is the removed paragraph. `chat.composer.create.videoDescription` — a different key still consumed by `signals/okou-page/composer-create.ts:87` — is untouched, and `onboarding.workflowPicker.categoryDescription` is unrelated.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | `categoryDescription()` was module-local with a single call site in `SlashTemplateDetailPane`; the pane is reached only from `tiptap-workflow-composer.tsx:820`. `category` is still consumed for the icon, `slashTemplateCategoryLabel()` and `slashTemplatePreviewGroup()`, so no other branch loses input. |
| `REVIEW.md` | Removed keys must have no remaining consumers | Pass | Repo-wide search for `slidesDescription`/`illustrationDescription`/`videoDescription`/`websiteDescription`/`workflowDescription` under `chat.composer.slashPanel` returns no code references after the diff; all ten locales were edited symmetrically, leaving `slashPanel` = `browseAll`, `make`, `templateCount_*`, `useTemplate` in every file. |
| `REVIEW.md` | Existing coverage still protects a refactor; identify the genuinely missing scenario | Pass | `views/okou-page/__tests__/chat-composer-slash-panel.test.tsx` asserts the pane via `[data-slot="slash-template-detail"]` and `data-category`, plus cover selection and keyboard preview — none of it asserted the deleted prose, so the suite still covers the pane after the removal. Removing static copy adds no behavior that needs a new case. |
| `REVIEW.md` | Persisted data and deployment compatibility | N/A | The diff touches only client-bundled i18n JSON and one presentational component; no schema, request, response, queue payload, or persisted state is involved. |
| `REVIEW.md` | Fallbacks section required for new fallbacks | N/A | No fallback, default, or tolerance branch is introduced; the deletion removes rendered copy, not a code path that handled an unexpected state. |
| `docs/bad-smell.md` | No `any`, no lint/type suppressions | Pass | Deletion-only diff; no type annotations, casts, or suppression comments added. |
| `docs/bad-smell.md` | Error handling: no redundant wrappers or fabricated defaults | N/A | No error path, catch, or default value is touched. |
| `docs/bad-smell.md` | Interface changes must document contracts and assess consumers | Pass | `categoryDescription` was not exported; the exported `slashTemplateCategoryLabel` and the component props are unchanged. Locale JSON is compiled into the same bundle as its reader, so removing a key with its only reader is not a cross-deployment contract change. |
| `docs/bad-smell.md` | Static imports / centralized configuration | N/A | No imports or configuration values change; the `i18n` import is still used by `slashTemplateCategoryLabel`. |
| `docs/bad-smell.md` | Fallback patterns: removal needs evidence of the owning contract | Pass | The evidence is the switch state plus the absent references: `ComposerSlashTemplatePanel` is `enabled: false` with `STAFF_ORG_ID_HASHES` (`turbo/packages/core/src/feature-switch.ts:486`), and the strings had exactly one reader, now deleted with them. |
| `docs/fallback.md` | Do not write negative tests against removed code | Pass | No test was added asserting the description is gone; the existing pane tests were left as-is. |
| `docs/fallback.md` | A non-GA feature switch needs no compatibility work | Pass | The panel is staff-only and pre-GA, so no dual-read, transitional string, or deprecation window is required for the removed keys. |
| `docs/fallback.md` | Do not fall back on states the contract cannot produce | N/A | No `??`/`||` chain or tolerance branch is added or removed. |
| `docs/testing.md` | Test the external contract through the production entry point | Pass | The affected suite boots the composer and queries rendered DOM slots rather than component state, so it keeps verifying the pane through the production surface after the copy removal. |
| `docs/testing.md` | Add coverage for new behavior and regressions where it adds confidence | Pass | The change deletes static presentation with no new state, branch, or user-triggerable path; a test asserting the paragraph's absence would be exactly the tombstone `docs/fallback.md` §1 rejects. |
| `docs/styles.md` | One styling API: Tailwind utilities, no first-party selectors | Pass | The removed `<p>` used only utilities (`mt-3 shrink-0 text-[13px] leading-6 text-muted-foreground`); nothing introduces a CSS module, `<style>`, or class selector, so the shrink-only baseline is unaffected. |
| `docs/styles.md` | Token and variant governance | N/A | No token, variant, or theme mapping is added, renamed, or removed; the surrounding `text-muted-foreground` / `rounded-lg` usages are untouched. |

### Findings

No P0 or P1 findings.

Non-blocking observation: `workflowDescription` is deleted along with the other four, which is correct but worth noting — the `workflow` category never reached the detail pane (`isSlashTemplatePreviewCategory` narrows it out in `SlashTemplatePanel`), so that string was already unreachable before this PR.

### Testing

- No test changes, and none are required: the deleted paragraph was static copy with no branch or interaction.
- Existing coverage that still exercises the pane: `turbo/apps/platform/src/views/okou-page/__tests__/chat-composer-slash-panel.test.tsx` (pane presence/absence, `data-category` per highlighted row, cover selection, keyboard preview) — all assert DOM slots and accessible names rather than the removed prose, so they continue to protect the refactor.
- No material gap identified. Visual confirmation belongs on the PR preview rather than a unit assertion.

PR #34051, opened 2026-09-14
refactor: remove agent message math feature switch

bingjiez666:
## Terminal state

`agentMessageMath` is **fully rolled out**. It was introduced staff-only in
e227a458 (#32362, `enabled: false` + `STAFF_ORG_ID_HASHES`) and flipped to
`enabled: true` with no allowlist in #33680 ("enable agent math and workflow
fuzzy search for all users", merged 2026-09-12). This PR permanently keeps the
switch-on behavior and removes the switch, mirroring #33831, which retired
`markdownTime` the same way after its rollout.

## Commit archaeology

The introducing commit's gating was compared against its parent to establish the
pre-switch baseline:

- Parent behavior: `createPlainMarkdownTree(..., { mathEnabled: false })` and
  `parseMarkdownTree(..., { mermaid: true })` with no `math` option.
- Switch-on behavior: the same call sites with `mathEnabled` / `math` set from
  `agentMessageMathEnabled$`.
- `EventTree.mathEnabled` and `RichEventTreePlan.mathEnabled` were added by the
  same commit purely as a cache identity, so a tree parsed under one switch
  value would be reparsed when the value changed during bootstrap hydration.
- The commit's `initialFeatureSwitchHydration$` boundary was already removed
  from `main` by later work, so nothing remained to clean up there. The
  vendored `katex-0.18.7` asset and its `.prettierignore` entry were likewise
  already replaced by the `katex` npm dependency.

## Behavior kept

- Chat events and shared threads parse Markdown with math recognition on:
  `parseMarkdownTree(..., { math: true, ... })`.
- `createPlainMarkdownTree(..., { mathEnabled: true })` on both surfaces, so
  `$$` bodies keep falling through to the full parser instead of taking the
  plain single-paragraph fast path.
- All rendering code stays: `lib/markdown/math.ts`, `signals/math-formula.ts`,
  `views/components/math-formula.tsx`, and the `rich-markdown.tsx` renderers.

## Behavior removed

- `mathEnabled` is no longer part of the `EventTree` / `RichEventTreePlan`
  identity, and the three comparisons that guarded against a mid-session switch
  change are gone; the event-content comparison returns to its pre-switch form.
- `planEventTreeUpdates` and `createSharedThreadRichContentSignals` lose their
  `mathEnabled` parameters.
- `agentMessageMathEnabled$`, the registry entry, and the `FeatureSwitchKey`
  member are deleted.

`mathEnabled` on `createPlainMarkdownTree` and `math` on `parseMarkdownTree`
are kept: both predate or outlive this switch and still distinguish Agent
message surfaces from `markdown.tsx` and `markdown-preview-tree.ts`, which
pass `false` independently of any rollout.

## Tests

- Deleted `The disabled rollout does not render formulas` — it only covered the
  discarded off behavior.
- Removed the three `AgentMessageMath: true` overrides from `markdown-math.test.tsx`
  and the `AgentMessageMath: false` override from `markdown-parser-isolation.test.tsx`
  (that file's fixtures contain no math delimiters).
- Retitled two tests that named the rollout rather than the behavior.
- Removed the three `agentMessageMath` assertions from the core feature-switch
  tests and the fixture key from `icon-mono-fixtures.json`.
- No negative tests were added.

## Second-round search

Swept for `agentMessageMath`, `AgentMessageMath`, `agentMessageMathEnabled`,
`agent-message-math`, `agent_message_math`, `AGENT_MESSAGE_MATH`, `mathEnabled`,
`initialFeatureSwitchHydration`, `syncHydratedEventTrees`, `featureSwitchHydration`,
and `katex-0.18.7` across the repository, including JSON fixtures and docs. All
switch terms return zero hits; the remaining `mathEnabled` hits are the retained
option and its two non-Agent callers.

## Verification

- Passed: `turbo run check-types --filter=@okouai/core --filter=@okouai/app`.
- Passed: `turbo run lint --filter=@okouai/core --filter=@okouai/app`.
- Passed: `vitest --run packages/core/src/__tests__/feature-switch.test.ts` (33 tests).
- Passed: `vitest --run` for `markdown-math.test.tsx`, `markdown-parser-isolation.test.tsx`,
  `markdown-interactions.test.tsx`, and `views/shared-thread-page/__tests__` (17 tests).
- Passed: `knip` (exit 0) and Prettier on the changed files.
- Core type check and the core feature-switch tests were rerun after rebasing
  onto current `main`.
- Two `markdown-interactions.test.tsx` copy-confirmation assertions failed once
  under two Vitest workers and passed on every single-worker run; they are
  timing-sensitive and unrelated to this diff, which touches no clipboard code.
- Full local Vitest and a local dev server were not run, per the requested
  verification policy; delegated to PR CI.


bingjiez666:
LGTM


PR_AUTO_REVIEW_COMMIT: 03baeb372e7163ce957aa1c2dee5d2a564f8848c
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `03baeb372e7163ce957aa1c2dee5d2a564f8848c`. Practice revision: `4a1be6501c472f58a7ec628429ef07478f416e03`.

### Summary

Retires the terminal `agentMessageMath` feature switch. The switch shipped staff-only in e227a458 (#32362) and went GA in #33680 with `enabled: true` and no allowlist, so this PR makes the LaTeX-delimiter path unconditional (`math: true`, `mathEnabled: true`) at the two Agent-message surfaces and deletes the registry entry, the enum member, `agentMessageMathEnabled$`, the per-event `mathEnabled` cache identity, the switch overrides in page tests, the disabled-branch test, and the e2e fixture key.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers | Pass | All five `parseMarkdownTree` / `createPlainMarkdownTree` production callers enumerated; the two Agent surfaces (`create-chat-thread.ts:1897`, `shared-thread-rich-content.ts:57`) now pass `true`, and `rich-markdown.tsx:512` + `markdown-preview-tree.ts:40` keep math off independently of any rollout. |
| `REVIEW.md` | Confirm existing coverage still protects a refactor; identify the missing scenario rather than inferring from a filename | Pass | `markdown-math.test.tsx` retains chat rendering, currency/`$$`-prose non-interference, KaTeX error states, and the shared-thread path, all now exercised without a switch override. |
| `REVIEW.md` | Features that change/remove/reroute an existing flow need regression coverage plus correct containment | Pass | The rerouted flow is switch-read removal only; rendered output is unchanged for GA users and is still asserted through the page (4 `<math>` nodes, 2 block-display, `data-math-status` sequence). |
| `REVIEW.md` | Report concrete practice violations with rule and source location; do not infer defects from search matches | Pass | No violation found; residual `mathEnabled` matches were each resolved to a pre-existing non-Agent caller. |
| `docs/fallback.md` | §2 — when a switch is terminal, delete the switch, the disabled branch, and the disabled-branch tests together | Pass | Registry entry, `FeatureSwitchKey.AgentMessageMath`, `agentMessageMathEnabled$`, the off-path plumbing, and `The disabled rollout does not render formulas` are removed in one diff. |
| `docs/fallback.md` | §1 — no negative/tombstone tests asserting removed behavior stays removed | Pass | The diff adds no test; it only deletes the disabled-branch case and four switch overrides. No assertion pins the key's absence. |
| `docs/fallback.md` | §9 — every fallback the diff introduces must be declared and listed by the reviewer | N/A | The diff introduces no fallback: no new `??`/`||` chain, tolerance branch, or dual-read. `math: true` is an unconditional literal, so no `Fallbacks` section is required. |
| `docs/fallback.md` | §10 — removal needs type/schema, single-writer, production, or rollout evidence | Pass | Rollout evidence: #33680 set `enabled: true` with the staff allowlist removed (merged 2026-09-12), recorded in the PR body. |
| `docs/bad-smell.md` | §1/§2 — no `any`, no lint/type suppressions | Pass | Diff adds no type assertion or suppression comment; `@okouai/core` and `@okouai/app` `check-types` and `lint` pass. |
| `docs/bad-smell.md` | §3 — preserve meaningful recovery, cleanup, and security checks when removing code | Pass | `signal.throwIfAborted()` guards, the `settle`/`markPendingEventTreesFailed` error path, `accept(..., [200, 404], signal)`, and `setPageSignal$` ownership are all retained; only the switch read is removed. |
| `docs/bad-smell.md` | §4 — document changed public contracts and assess consumers | Pass | `createSharedThreadRichContentSignals` loses a parameter; its single caller (`shared-thread-page-setup.ts:60`) is updated in the same diff and Knip reports no unused export. |
| `docs/deployment-compatibility.md` | Account for an old open frontend running against the new backend | Pass | `featureSwitches$` seeds `combined` from the client's own bundled registry and `applySwitches` only overwrites keys present in the response, so an already-open pre-deploy tab keeps `agentMessageMath: true` from its own bundle and continues rendering math. |
| `docs/deployment-compatibility.md` | Persisted state must stay readable across versions | Pass | Stale `agentMessageMath` override rows are dropped by `filterFeatureSwitchOverrides` (`key in FEATURE_SWITCHES`) on both read and write paths, and the contract is `z.record(z.string(), z.boolean())`, so a retired key cannot fail response validation. |
| `.claude/skills/ccstate/SKILL.md` | Keep derivation in `computed` and semantic actions in commands; no store mutation during render | Pass | Only a `get()` read is deleted from `ensureEventTrees$` and `setupSharedThreadPage$`; no signal is created or mutated in render, and `trees$` stays a `computed`. |
| `.claude/skills/ccstate/SKILL.md` | Every async operation keeps an owner and cancellation path; `accept` keeps a non-empty status list | Pass | No `detach`/`void` added, no `throwIfAborted` removed, and `accept(client.get(...), [200, 404], signal)` is unchanged in `shared-thread-page-setup.ts`. |
| `.claude/skills/feature-switch/SKILL.md` | A feature switch is not required for internal refactors or code cleanup | Pass | This is terminal-switch cleanup of an already-GA behavior, so no replacement gate is needed. |
| `.claude/skills/feature-switch/SKILL.md` | Registry, key, client reads, and overrides are removed together | Pass | `feature-switch.ts`, `feature-switch-key.ts`, `signals/external/feature-switch.ts`, the core rollout-matrix assertions, and the e2e fixture key are all updated in this diff; no `FEATURE_SWITCH_BY_AUTH_METHOD` or `CONDITIONAL_CAPABILITIES` entry referenced this key. |
| `docs/testing.md` | Delete tests for removed behavior instead of rewriting them into tombstones; avoid cases that restate static configuration | Pass | The disabled-branch test is deleted; the three surviving cases keep their original product assertions with only the override removed. |
| `docs/testing.md` | Platform boots the real Router with `setupPage` and asserts rendered content | Pass | All retained cases still go through `setupPage` with `chat.install(...)` / `context.mocks.api(sharedThreadsContract.get, ...)` and assert DOM output. |
| `docs/testing/testing-external-behavior.md` | Construct cases through page interactions and verify through the page; no internal state or mock-call assertions | Pass | Assertions read `document.querySelectorAll("math")`, `data-math-status`, and `screen.findByText`; removing the `featureSwitches` override moves the tests closer to the real user context rather than further from it. |

### Findings

No P0 or P1 findings.

Non-blocking observations:

- `e2e/playwright/style-migration/icon-mono-fixtures.json` participates in `fixtureSha256`, which `run-icon-mono.ts` asserts as `Frozen fixture changed` when invoked with `--baseline`. That tool is not referenced by any workflow in `.github/workflows` and is documented as a manual run in `docs/style-migration.md`, so this does not affect CI; an operator holding a frozen baseline will need to recapture it. Keeping the retired key instead would have been a small tombstone, so removing it is the right call.
- Users who had explicitly set an `agentMessageMath: false` override lose that opt-out, since `filterFeatureSwitchOverrides` now drops the row. That is the intended consequence of retiring a GA switch, matching #33831, and needs no action.
- `mathEnabled` on `createPlainMarkdownTree` and `math` on `parseMarkdownTree` are correctly retained: both have live `false` callers (`markdown.tsx:140`, `markdown-preview-tree.ts:36`, `rich-markdown.tsx:512`, `markdown-preview-tree.ts:40`) that predate this switch, so collapsing either option would change behavior for attachment previews and one-off Markdown documents.

### Testing

Coverage is appropriate for a terminal-switch cleanup. The retained `markdown-math.test.tsx` cases still prove the permanent behavior through the page: explicit `\(...\)`, `\[...\]`, and `$$...$$` delimiters render as MathML while `$2,499`, `$ABC$`, and inline `$$not math$$` stay literal, malformed TeX stays readable with `data-math-status="hasError"`, code and fenced blocks stay literal, and the shared-thread route renders one formula. The deleted case covered only the switch-off path, so it dies with the switch rather than becoming a tombstone, and no negative test replaces it.

Removing the `AgentMessageMath: false` override from `markdown-parser-isolation.test.tsx` is safe: none of that file's rendered bodies (`++Plain message++`, `**Streaming:** ++Reply++`, `` `++Code++` and C++ / i++ ``, the `++[Reference](...)++` link, the checklist, and the Activity log text) contain `$`, `\(`, or `\[`, so math recognition cannot alter those trees — confirmed by the suite passing unchanged.

Verification observed in the PR body and reproduced here: `check-types` and `lint` for `@okouai/core` and `@okouai/app`, the core feature-switch suite (33 tests), and the Platform suites for `markdown-math`, `markdown-parser-isolation`, `markdown-interactions`, and `views/shared-thread-page` (17 tests), plus Knip and Prettier. The author's disclosure that two `markdown-interactions.test.tsx` copy-confirmation assertions failed once under two Vitest workers and passed on single-worker runs is consistent with the diff touching no clipboard code. Full-suite confirmation is correctly delegated to PR CI.

No material coverage gap identified.

Issue #33822, opened 2026-09-14
Map deleted Clerk user during command authorization to a controlled auth outcome


seven332:
## Summary

A command-status read failed with Clerk 404 resource_not_found / No user was found. The SDK exception escaped the auth boundary as an unhandled request error.

## Impact

Classification: **bug; medium priority**. The evidence establishes the component outcome described above, not a service-wide outage or an unmeasured affected-user total. The named operation can fail or degrade, and incomplete diagnostics can prevent reliable recovery. Scope is bounded to the recorded operations.

### Coverage

Reporting window: **2026-09-11 00:00 to 2026-09-14 00:00 Asia/Shanghai**, end exclusive. Exact UTC interval: **[2026-09-10T16:00:00Z, 2026-09-13T16:00:00Z)**.

Axiom: dataset `vm0-web-logs-prod`, predicate `level == "error" or level == "warn"`; **1 raw events** (1 error, 0 warn). First/last: 2026-09-12T18:20:25.725Z — 2026-09-12T18:20:25.725Z.

Sentry: projects **api**, explicit `environment:production level:error`, all issue statuses; **1 raw events / 1 issue groups** (API-5A). First/last: 2026-09-12T18:20:25Z — 2026-09-12T18:20:25Z. Native level is retained in the tags when the event has no top-level level field. Private Sentry links are intentionally omitted.

Correlation and counting: One matching request/Clerk exception in Axiom and API-5A.

Privacy: identifiers, network addresses, private URLs, and SQL parameters are explicitly redacted. Samples are field-selected original records; absent fields are not invented. Native exception/stack wording is preserved except these redactions and explicitly marked length truncation. Counts and variant metadata are derived separately.

### Variant 1 — Unhandled request error: Not Found (1 Axiom events)

Source: Axiom. First/last: 2026-09-12T18:20:25.725Z — 2026-09-12T18:20:25.725Z.

Original message path: `message`; causal fields are shown below with their exact native names. `fields.context`: App (1); `source`: api (1); `route`: /api/computer-use/commands/:commandId (1); `fields.route`: /api/computer-use/commands/:commandId (1); `fields.method`: GET (1); `fields.x_client_type`: CLI (1); `fields.x_client_version`: 9.324.4 (1).

```json
{
  "_time": "2026-09-12T18:20:25.725Z",
  "errorCode": "api_response_error",
  "errorSummary": "Not Found",
  "fields.context": "App",
  "fields.error.clerkError": true,
[...]

PR #31615, opened 2026-09-04
fix(api): retry transient clerk read failures

seven332:
## Summary

- retry Clerk Backend API 500-599 failures only inside the existing idempotent-read gateway, using the current three-attempt/shared-delay budget and a stable terminal error
- propagate the active API signal into CLI PAT membership reads, then map exhausted reads on `POST /api/scrape` to the existing non-cacheable 503 contract
- emit one sanitized, queryable Clerk provider-unavailable event and cover recovery, exhaustion, cancellation, authorization misses, 4xx behavior, and mutation non-retry through API entry points

## Scope and compatibility

- Clerk 429 `Retry-After` behavior is unchanged
- Clerk mutations remain single-attempt
- expired membership roles are never used as an availability fallback
- direct Clerk session failures and unrelated exceptions remain on their existing unhandled paths
- no schema, contract, frontend, CLI, runner, queue, or persisted-state change is required

## Validation

- `pnpm format`
- `pnpm -F api lint`
- `pnpm check-types --filter=api`
- `pnpm -F api check-types:tests`
- `pnpm knip --workspace api`
- `pnpm -F api exec vitest run src/lib/__tests__/log.test.ts src/signals/routes/__tests__/scrape.test.ts src/signals/routes/__tests__/org-team.bdd.test.ts --maxWorkers=1` (89 passed)
- pre-commit hook: full-repository Knip, full-repository type checking (15 tasks), staged formatting, and file-size checks

Closes #31576

PR #34052, opened 2026-09-14
fix(api): map deleted clerk identity to a controlled auth outcome

lancy:
## Summary

Classify the two unclassified Clerk failures at the authentication boundary so they become controlled auth outcomes instead of escaping as `unhandled_request_error` / HTTP 500.

- **`resource_not_found` (404)** — the recorded production incident (#33822). The Clerk identity no longer exists.
- **`ClerkReadUnavailableError` (5xx retries exhausted)** — the adjacent gap in the same function at the same boundary.

Both are classification at one call site. This is not a general Clerk error policy.

### The defect

`getMemberRoleAndUpdateCache$` returned `{ role } | null`, and `null` already means "the user exists but is not a member of this org", which legitimately degrades to a user-only PAT context. "The identity is gone" had no way to express itself in that type, so the raw `ClerkAPIResponseError` escaped `cliAuth$` and `agentAuth$` and surfaced as HTTP 500.

### The change

1. **Three-valued result.** `MemberRoleResult` is `member | not_member | identity_not_found`, so a deleted identity can no longer collapse into the `not_member` degraded context.
2. **Allowlist classification at the call site**, not in `retryClerkRead`. Written as a negated guard so anything unclassified keeps its current path by construction:
   ```ts
   if (!isClerkResourceNotFound(read.error)) {
     throw read.error;
   }
   ```
   Exactly one predicate produces `identity_not_found`, and exactly one (`clerkReadUnavailable`) produces the 503. Every other Clerk or database error is rethrown unchanged.
3. **Fail closed.** `identity_not_found` drops the stale `org_members_cache` row and returns `null` from `cliAuth$` / `agentAuth$`, reaching the existing `401 UNAUTHORIZED` result. No new response shape.
4. **Controlled provider-unavailable.** `requiredAuthContext$` maps an exhausted Clerk read to a non-cacheable `503 PROVIDER_UNAVAILABLE`, following the existing `routes/scrape.ts:38-58` precedent, and keeps its actionable error-level `provider_unavailable` record.

`retryClerkRead`, `clerkReadRetry`, all retry/backoff timing in `clerk.ts`, and `MEMBER_ROLE_CACHE_TTL_MS` are unchanged. No webhook, background job, reconciliation path, token/session revocation, or change to `not_member` behavior.

## Residual risk (accepted, not addressed here)

A deleted identity keeps its organization authority for **at most 60 seconds** while a fresh `org_members_cache` row exists: the cached branch returns before Clerk is consulted. Eliminating that window requires Clerk-deletion-event-driven invalidation, which belongs to **#33745 slice B2**. `MEMBER_ROLE_CACHE_TTL_MS` is deliberately not shortened or invalidated here.

## Deviations from the implementation contract

Each deviation is forced by current `main` or by a repository gate, and none changes an acceptance criterion's observable outcome.

1. **Eight consumer call sites, not four.** The contract listed the 4 OAuth call sites. `getMemberRoleAndUpdateCache$` actually has 8 consumers outside the auth boundary — the 4 listed plus `run-error-format.service.ts`, `integration-run-errors.service.ts`, and two in `slack-connector-oauth.service.ts`. The return-type change forces all of them. Every one maps the old `null` branch to `kind !== "member"`, so `not_member` behaves exactly as before and `identity_not_found` denies (criterion 7).
2. **The new record is `debug`, not `info`** (criterion 8). `api/no-logger-info` is an error-level ESLint rule forbidding `logger.info()` in API source, and its own guidance prescribes debug for a routine diagnostic. The rule has an `allowedMessages` escape hatch, deliberately not used here rather than weaken a project-wide guardrail for a handled, expected condition. The issue's desired outcome explicitly sanctions "info/debug". The record is otherwise exactly as specified: stable `type: "clerk_identity_not_found"`, and no user, org, or Clerk trace identifier.
3. **`MemberRoleResult` is module-local, not exported.** `pnpm knip --workspace api` fails on unused exports; every consumer narrows through inference.
4. **Criterion 7 is covered in `teams-oauth.test.ts`, not `org-team.bdd.test.ts`.** `org-team.bdd.test.ts` contains no OAuth call-site coverage — its Clerk reads go through `organizations.getOrganizationMembershipList`, a different call site. `teams-oauth.test.ts` already covers `teams-oauth.ts` `not_member` denial directly, so the new `identity_not_found` test sits beside it and asserts the identical denial.
5. **Criterion 2 is verified through route behavior, not a direct DB assertion.** ESLint `no-restricted-imports` forbids API tests from importing DB schema or handles. The cache probe re-issues the request inside `MEMBER_ROLE_CACHE_TTL_MS` of the warm-up write while Clerk still reports the identity as gone: a surviving row answers from cache and reaches the handler (404), a dropped row must consult Clerk and fails closed (401).
6. **Criterion 5 is observed on `GET /api/auth/me`.** `GET /api/computer-use/commands/:commandId` uses `requireOrganization: true, missingOrganizationStatus: 401`, so on that route `not_member` and `identity_not_found` both yield 401 and cannot be told apart. The degraded user-only context is only observable on a route that does not require an org, so the `not_member` control uses the existing PAT `me` surface in the same suite.

## Scope and compatibility notes

- **`scrape.ts` is intentionally untouched.** Its route-level `clerkReadUnavailable` wrapper still guards a Clerk exhaustion raised inside the scrape handler itself, which remains reachable. The auth path is now classified upstream with the identical body and `Cache-Control: no-store`, so `scrape.test.ts` is unchanged and still asserts the same 503 contract.
- **503 is not declared in most route `responses` maps.** `validateResponse` returns a response untouched when the contract has no schema for its status, so an auth-boundary 503 is returned as-is. No route or contract needed a change; `authRoute` passes `AuthErrorResponse` straight through.
- **Clerk session failures are unaffected.** The allowlist keys on the `ClerkReadUnavailableError` instance, not a status code, so a direct `authenticateRequest` 5xx stays on its existing unhandled path (`scrape.test.ts` "does not retry direct Clerk session failures" still expects 500).
- Clerk 429 keeps its current behavior; it is neither retried nor classified here.

## Validation

Rebased onto `origin/main` `ba0fc4da5dac1c87b94310a236c2bce42f4cb0b7`; every contract anchor was re-read first at `d14e0764` and re-confirmed after the rebase. None of the 10 intervening commits touch any file in this diff.

- `pnpm format`
- `pnpm check-types --filter=api`
- `pnpm -F api check-types:tests`
- `pnpm -F api lint`
- `pnpm knip --workspace api`

Targeted tests only, per the contract; no local full Vitest run and no dev server. Test execution is left to PR CI, which also covers the expected-affected `scrape.test.ts` and `org-team.bdd.test.ts` Clerk-boundary suites.

New coverage, keyed to the acceptance criteria:

- `computer-use.bdd.test.ts` — `AUTH-05`, on the route that failed in production:
  - deleted identity → 401 `UNAUTHORIZED`, no retry, no Sentry capture, stale cache row dropped (criteria 1, 2)
  - exhausted 5xx → 503 `PROVIDER_UNAVAILABLE` + `Cache-Control: no-store`, 3 attempts / 2 delays, cached role survives (criterion 3)
  - Clerk 429 → unchanged 500, single attempt, no delay (criterion 4)
  - Clerk 403 and a non-Clerk error → unchanged 500, cached authority untouched (criterion 6)
  - non-member → degraded user-only context on `me`, contrasted with the deleted identity failing closed (criterion 5)
- `teams-oauth.test.ts` — OAuth call site denies `identity_not_found` exactly as it denies a non-member today (criterion 7)

`git diff` shows no change to retry or backoff behavior in `clerk.ts` and no change to `MEMBER_ROLE_CACHE_TTL_MS` (criterion 9).

Refs #33822


lancy:
LGTM


PR_AUTO_REVIEW_COMMIT: 92f8de2e2b8f46b9e960479b27bdaf8cff1abf9f
PR_AUTO_REVIEW_VERDICT: LGTM

Inspected HEAD: `92f8de2e2b8f46b9e960479b27bdaf8cff1abf9f`. Practice revision: `main` `ba0fc4da5dac1c87b94310a236c2bce42f4cb0b7`.

### Summary

Turns two unclassified Clerk failures at the authentication boundary into controlled auth outcomes: a deleted Clerk identity (404 `resource_not_found`) now fails closed with 401 and drops its stale `org_members_cache` row, and an exhausted Clerk read (`ClerkReadUnavailableError`) returns a non-cacheable 503 `PROVIDER_UNAVAILABLE`. `getMemberRoleAndUpdateCache$` gains a three-valued `MemberRoleResult` so "the identity is gone" can no longer collapse into the `not_member` degraded PAT context, and its eight consumers are updated to the new discriminant.

### Review Criteria

| --- | --- | --- | --- |
| `REVIEW.md` | Trace changed behavior through its real entry point and consumers; check auth, permissions, and failure outcomes | Pass | All 8 `getMemberRoleAndUpdateCache$` consumers updated; `agentAuth$` returning `null` cannot fall through to a sandbox context because `sandboxTokenPayloadSchema`/`okouTokenPayloadSchema` are exclusive `z.literal` scopes (`auth/tokens.ts:53,75`), so `resolveSandboxAuth` rejects an okou-scoped token. |
| `REVIEW.md` | Meaningful coverage for new behavior and regressions | Pass | `computer-use.bdd.test.ts` AUTH-05 covers 401/cache-drop, 503, 429, Clerk 403, non-Clerk error and `not_member`; `teams-oauth.test.ts` covers the OAuth denial. |
| `REVIEW.md` | Tests verify externally observable state or HTTP responses, not internals | Pass | All assertions are HTTP status, body, headers, or a later request's outcome. The cache assertion is a follow-up request, not a DB read. |
| `REVIEW.md` | API tests must not assert log output | Pass | No `axiomLogging`/logger assertions in the diff. `sentry.captureException` is used only as the observable unhandled-error outcome, matching the precedent merged in #31615 (`scrape.test.ts`, `org-team.bdd.test.ts`). |
| `REVIEW.md` | Changed/replaced flows need regression coverage for the affected path | Pass | The pre-existing `not_member` degraded context and the 429/unrelated-error paths each have an explicit negative control, so the replacement is proven scoped. |
| `REVIEW.md` | Report practice violations with rule and source; do not infer bugs from shape | Pass | No inferred findings raised. |
| `REVIEW.md` | Persisted data / API contract and response validation reviewed together | Pass | No schema, migration, or `jsonb().$type<>()` change. `AuthErrorResponse` widens to include 503; `validateResponse` (`trpc-contract.ts:334-337`) returns a response untouched when the contract declares no schema for its status, so routes without a declared 503 still serialize correctly. |
| `REVIEW.md` | Deployment-compatibility windows for changed requests/persisted state | N/A | No request, queue payload, runner protocol, or persisted-state shape changes. A 500 becoming 401/503 on an already-failing request needs no cross-version window. |
| `docs/bad-smell.md` §1-2 | No `any`, no lint/type suppressions | Pass | No `any` and no suppression comments added; `read.error` stays `unknown` and is narrowed by predicates. |
| `docs/bad-smell.md` §3 | Catch only at a boundary that meaningfully handles the error; no code whose only output is a log level | Pass | Both `settle` regions are the narrowest that contains the classified read, and each produces a real outcome (401 + cache deletion; 503 + `no-store`), not just a log. See the observation below on the new debug record. |
| `docs/bad-smell.md` §3 / `docs/externally-managed-references.md` | An expected external miss differs from invalid input, a failed required operation, or a local invariant | Pass | 404 → `identity_not_found`; exhausted 5xx → `PROVIDER_UNAVAILABLE`; 429, Clerk 403 and database errors keep propagating. |
| `docs/bad-smell.md` §4 | Document changed public contracts and assess consumers | Pass | The new 503 and the `MemberRoleResult` consumer migration are both described in the PR body; all consumers verified by `pnpm check-types --filter=api`. |
| `docs/bad-smell.md` §5-6 | Static imports; centralized configuration | Pass | Only static imports added; no URLs or credentials introduced. |
| `docs/bad-smell.md` §7 / `docs/fallback.md` | No fallback for a state the contract rules out; declare every new fallback | Pass | The diff introduces no fallback — see **Fallbacks** below. |
| `docs/fallback.md` §1 | No negative tests pinning removed behavior | Pass | Every new test asserts currently-reachable behavior; the 429/403/unrelated controls are fail-closed boundary assertions, which §1 explicitly preserves. |
| `docs/externally-managed-references.md` | Represent an expected resolution miss as a normal domain result, not a 500 | Pass | `auth.service.ts` returns `{ kind: "identity_not_found" }` instead of letting `ClerkAPIResponseError` escape. |
| `docs/externally-managed-references.md` | Security must fail closed; an unresolved reference grants nothing | Pass | `cliAuth$`/`agentAuth$` return `null` → existing 401. No degraded context, no cached role, no sandbox upgrade. |
| `docs/externally-managed-references.md` | Handle only the authority's explicit outcomes; do not wrap a route and call every exception "unavailable" | Pass | Both guards are negated allowlists (`if (!isClerkResourceNotFound(...)) throw`, `if (!unavailable) throw`); the 503 keys on the `ClerkReadUnavailableError` instance, so a direct `authenticateRequest` 5xx stays on its unhandled path. |
| `docs/externally-managed-references.md` | Record diagnostic context without credentials or sensitive provider data | Pass | The new record carries only `type: "clerk_identity_not_found"`; no user ID, org ID, token, or Clerk trace ID. |
| `docs/externally-managed-references.md` | Test at a real entry point: well-formed reference, unavailable result not a 500, no capability granted | Pass | AUTH-05 drives `GET /api/computer-use/commands/:commandId` over HTTP with a real PAT and asserts 401 rather than 500. |
| `docs/testing.md` | Prefer integration coverage through real entry points; assert status/headers/body/effects | Pass | `createCliToken` builds the PAT through the production device-auth flow; assertions are HTTP-level. |
| `docs/testing.md` | Mock external services at their boundary; use real internal code and the real database | Pass | Only `context.mocks.clerk`, `signalTimers`, and `sentry` are used; no `vi.mock()` of internal modules. |
| `docs/testing.md` | Avoid fake timers; wait for the observable result, not an internal cache update | Pass | Uses the sanctioned `mockNow`/`clearMockNow` already used throughout this suite, and asserts an HTTP status rather than inspecting the cache. |
| `docs/testing/testing-external-behavior.md` | API tests must not import DB schema, services, or the logger | Pass | No DB, service, or logger imports; cache state is observed through a follow-up request. |
| `docs/testing/testing-external-behavior.md` | Exceptions must be rare and justified | N/A | No external-behavior exception was needed. |
| `docs/testing/api-testing.md` | Cover auth and membership behavior through the endpoint | Pass | Auth classification is exercised entirely through `GET /api/computer-use/commands/:commandId` and `GET /api/auth/me`. |
| `docs/testing/api-testing.md` | Uniquely owned identities for shared persistent state | Pass | Each test allocates a fresh `bdd.user()`, so no two tests share an `(orgId, userId)` cache key. |
| `docs/testing/api-testing.md` | Cache assertions own their key; set and advance mocked time inside the test that exercises the TTL | Pass | `patWithWarmMembershipCache` and `statusAfterCacheProbe` set the clock within the same test; the probe instant is derived from the warm-up instant, so the TTL window does not depend on suite wall time. |
| `.claude/skills/ccstate/SKILL.md` | Every async operation needs an owner and cancellation path; do not silence work | Pass | Both new `settle(p, signal)` calls propagate `AbortError` (`signals/utils.ts`), preserving request cancellation; no `void`, no `detach()`, no new signal created. |
| `.claude/skills/ccstate/SKILL.md` | Keep derivation in `computed` and semantic actions in commands | Pass | Classification stays inside the existing commands; `ccstate/no-command-in-command` and `api/signal-check-await` pass. |

### Findings

No P0 or P1 findings.

Non-blocking observations:

1. `auth.service.ts` — the new `L.debug("Clerk identity no longer exists during membership read", { type: "clerk_identity_not_found" })` never reaches Axiom, because `lib/log.ts` sets the transport threshold to `info` and documents that debug records are dropped. `REVIEW.md` asks a PR that answers a log-noise issue to justify why a new record is needed at all, and `docs/bad-smell.md` §3 calls a new outcome classifier whose only output is a log level a smell. Here the classifier's output is the 401/503 status and the cache deletion, and the record is a local-diagnosis aid with zero production signal, so it is not the smell that rule targets. The production-observable outcome of this PR is the disappearance of the `unhandled_request_error` record plus the new status codes, which is what the issue's post-release check queries. Keeping or deleting the debug line would both be defensible; not blocking.
2. The 503 is returned from a shared boundary, so any authenticated route can now emit a status its ts-rest contract does not declare. This is safe today (`validateResponse` passes undeclared statuses through) and is strictly better than the previous 500, but it means the OpenAPI/contract surface under-describes the real responses. Worth a follow-up if contract completeness becomes a requirement; not a defect in this diff.

### Fallbacks

None introduced. Both new branches resolve an externally managed Clerk reference against its current authority and return that authority's outcome, which `docs/externally-managed-references.md` ("Relationship to Fallbacks") classifies as normal reference resolution rather than rolling-deployment compatibility. No action is taken after the unavailable result other than denying, so no separate fallback decision arises. The pre-existing 60-second `MEMBER_ROLE_CACHE_TTL_MS` read-through cache is unchanged by this diff; the PR body records its residual window and assigns removal to #33745 slice B2. Per §9, a PR that neither adds nor removes a fallback needs no `Fallbacks` section.

### Testing

Coverage is targeted and keyed to the behavior that failed in production.

- `computer-use.bdd.test.ts` — new `AUTH-05` suite on `GET /api/computer-use/commands/:commandId`, the exact route from the issue's Axiom record, authenticated with a real CLI PAT:
  - deleted identity → 401 `UNAUTHORIZED`, single Clerk attempt, no delay, no Sentry capture, and a follow-up request proving the stale cache row cannot re-authorize the deleted identity;
  - exhausted 5xx → 503 `PROVIDER_UNAVAILABLE` with `Cache-Control: no-store`, 3 attempts / 2 delays, cached role intact;
  - 429 → unchanged 500 with a single attempt and no delay, pinning the untouched rate-limit path;
  - Clerk 403 and a plain `Error` → unchanged 500 with the original error reaching Sentry, and cached authority untouched — these are the controls that prove the allowlist did not widen;
  - `not_member` → degraded user-only context on `GET /api/auth/me` (`orgId: null`), contrasted with the same PAT failing closed once the identity is gone. Using `me` rather than the computer-use route is necessary: `computerUseAuthOptions` sets `requireOrganization: true, missingOrganizationStatus: 401`, so both outcomes would be 401 there and the degraded context would not be observable.
- `teams-oauth.test.ts` — a deleted identity denies the OAuth callback with the same `Invalid connect state.` redirect as the existing non-member test, and captures nothing to Sentry, proving the call-site change is behavior-preserving.

Verification recorded in the PR body: `pnpm format`, `pnpm check-types --filter=api`, `pnpm -F api check-types:tests`, `pnpm -F api lint`, `pnpm knip --workspace api`. No local full Vitest run, per the repository guidance; the new suites and the expected-affected `scrape.test.ts` / `org-team.bdd.test.ts` Clerk-boundary suites run in PR CI.

Gaps considered and judged acceptable: no test asserts the `provider_unavailable` or `clerk_identity_not_found` log records, which is required rather than optional — `docs/testing/testing-external-behavior.md` forbids API tests from asserting logger calls, and the `provider_unavailable` field shape already has coverage in `lib/__tests__/log.test.ts` from #31615.

